PR #91, opened 2024-09-05
fix local cpu & add local async put (cpu and disk)

### LINKED ISSUE #48: [v0.1.x] (Refactor) Decouple `CacheEngine` and `device`
# Problem
Current `CacheEngine` has messy tensor device movement code and the semantics are unclear.

# Ideal design
All the device-related operations (i.e., `tensor.to(device)`) should be done in the storage backend layer. 
The `put` operation of the storage backend should take in a GPU tensor, and the `get` operation should return a GPU tensor

# Potential improvements
- Clearer semantics --> less bugs and better evolvability in the future

### LINKED ISSUE #81: [v0.1.x] (Performance, bugfix) Huge slowdown in TTFT and throughput when using local disk storage backend
**When testing with a 16K reused context on mistral-7b model:**

**The yellow curve is vLLM w/o LMCache. The blue curve is vLLM with LMCache local disk storage**

Image

Image

YaoJiayi:
This PR is related to the following issues:
**1. Add support for async put in local store (both cpu and gpu)** (resolves #81 )
Performance comparison when using local disk:

Blocking put (qps=1/8):
361980515-bfaf29a5-04ff-4b52-a6ee-c4bbb4d3aa53

Async put (qps=1/8):
图片

The few requests after the first requests are still being slowed doen significantly. The reason could be bandwidth contention (the process of saving the first req is not completed). We might need a rate controller for this (created a new issue on this #92 ).

Async put (qps=1/32):
图片

**2. The cache was wrongly stored in gpu instead of cpu when using cpu as local backend!!!**
Following are latest e2e results (qps=1/8). 
Block put w.o. pin memory:
图片
Block put w. pin memory:
图片

Async put w.o. pin memory:
图片

Async put w. pin memory:
图片

The performance should be improved with better cuda kernel and less memory copy.

**4. Decouple `device` and `CacheEngine put` (partly resolves #48 ). Also see pull request in lmcache-vllm**

**5. Fix a minor bug in unit test.**

### DIFF
--- a/lmcache/cache_engine.py
+++ b/lmcache/cache_engine.py
@@ -66,7 +66,6 @@ class LMCacheEngine:
     def _chunk_tokens(
             self, 
             tokens: torch.Tensor, 
-            #device
         ) -> Iterator[torch.Tensor]:
         """
         Chunk the tokens into chunks of size self.chunk_size.
@@ -260,7 +259,6 @@ class LMCacheEngine:
     @torch.no_grad()
     def retrive(self,
                 tokens: torch.Tensor,
-                #device: str = 'cuda'
         ) -> Tuple[KVCache, int]:
         """
         Retrive the KV cache of the tokens from the cache engine. The retrived KV cache 
--- a/lmcache/cache_engine.py
+++ b/lmcache/cache_engine.py
@@ -260,7 +260,7 @@ class LMCacheEngine:
     @torch.no_grad()
     def retrive(self,
                 tokens: torch.Tensor,
-                device: str = 'cuda'
+                #device: str = 'cuda'
         ) -> Tuple[KVCache, int]:
         """
         Retrive the KV cache of the tokens from the cache engine. The retrived KV cache 
@@ -289,7 +289,7 @@ class LMCacheEngine:
         for chunk in retrival_iterator:
             if chunk is None:
                 break
-            retrived_kv_chunks.append(chunk.to(device))
+            retrived_kv_chunks.append(chunk)#.to(device))
 
         ''' concatenate the kv cache '''
         dim = None
@@ -306,7 +306,7 @@ class LMCacheEngine:
             return (), 0
 
         st2 = time.perf_counter()
-        ret = self._blob_to_tuple_kv(torch.cat(retrived_kv_chunks, dim=dim + 2))#.to(device))
+        ret = self._blob_to_tuple_kv(torch.cat(retrived_kv_chunks, dim=dim + 2))
         ed2 = time.perf_counter()
         logger.info(f"Concatenated {len(retrived_kv_chunks)} chunks -- elapsed time {ed2 - st2}")
         retrived_token_count = 0 if len(ret) == 0 else ret[0][0].shape[dim]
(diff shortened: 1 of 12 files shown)


tests/data/test_creation_from_file/remote.yaml
```diff
@@ -1,6 +1,6 @@
 chunk_size: 256
 local_device: null
-remote_url: "lm://localhost:65000"
+remote_url: "lm://localhost:65432"
```
ApostaC: The port should be 65000. During the test, a `lmcache_server` process will be initialized at localhost:65000. Please refer to [conftest.py](https://github.com/LMCache/LMCache/blob/1f3ecc1d5949a7c8bfc973df1cbfaa56c1cdc0f0/tests/conftest.py#L34C1-L34C31)

lmcache/cache_engine.py
```diff
@@ -74,7 +66,7 @@ def _hash(
     def _chunk_tokens(
             self, 
             tokens: torch.Tensor, 
-            device
+            #device
```
ApostaC: A general comment: just feel free to remove the useless code instead of commenting it out -- git can be used to track history anyway.

lmcache/storage_backend/local_backend.py
```diff
@@ -32,6 +38,23 @@ def __init__(
...
+        self.device = config.local_device
+        
+        self.put_queue = queue.Queue()
+        self.put_thread = threading.Thread(
+                target=self.put_worker, args=()
+            ) 
+        self.put_thread.start()
+        self.update_lock = threading.Lock()
+        
+        # FIXME(Jiayi): `use_pin_memory` and `dst_device` should be configged dynamically
```
ApostaC, line 50: Maybe we should discuss this. There are a few alternatives for dst_device:
- Pass `dst_device` through vllm --> lmcache_vllm_driver --> lmcache engine --> lmcache backend
- Just assume dst_device is cuda, because a lot of lmcache functionalities depends on GPU, such as CacheGen.

lmcache/storage_backend/local_backend.py
```diff
@@ -15,6 +18,9 @@
 
 logger = init_logger(__name__)
 
+class LocalBackendEndSignal:
+    pass
+
 class LMCLocalBackend(LMCBackendInterface):
```
ApostaC, line 24: Seems like LMLocalBackend and LMRemoteBackend share a lot of codes. Maybe we should consider merging it back. 
This PR doesn't need to solve this, let's create a new issue after this PR is merged.

In my opinion, there should be 2 different scenarios:
- KV cache is in fast memory and serialization overhead is 0 or negligible --> don't need async put
- KV cache is in slower storages or have high serialization overhead --> need async put, and need pipelined get if storage is slow IO and deserialization is slow computation

lmcache/storage_backend/serde/cachegen_decoder.py
ApostaC, line 1: This PR seems to revert the changes from PR #86.

Did you pull the latest dev branch?

ApostaC (changes requested):
Thanks, Jiayi! Great stuff! Just had a few comments.

ApostaC (approved):

PR #102, opened 2024-09-20
Using `frombuffer` instead of load for deserialization 

xenshinu:
Default serde is single threaded and is too slow which takes about 800ms to deserialize kv cache of 1000 tokens in a 7B model. This code use `frombuffer` to deser KV cache which only takes 0.06ms

### DIFF
--- a/lmcache/storage_backend/serde/__init__.py
+++ b/lmcache/storage_backend/serde/__init__.py
@@ -3,6 +3,7 @@ from lmcache.storage_backend.serde.torch_serde import TorchSerializer, TorchDese
 from lmcache.storage_backend.serde.safe_serde import SafeSerializer, SafeDeserializer
 from lmcache.storage_backend.serde.cachegen_decoder import CacheGenDeserializer
 from lmcache.storage_backend.serde.cachegen_encoder import CacheGenSerializer
+from lmcache.storage_backend.serde.fast_serde import FastSerializer, FastDeserializer
 from lmcache.config import LMCacheEngineConfig, LMCacheEngineMetadata, GlobalConfig
 
 def CreateSerde(serde_type: str, config: LMCacheEngineConfig, metadata: LMCacheEngineMetadata):
@@ -12,6 +13,8 @@ def CreateSerde(serde_type: str, config: LMCacheEngineConfig, metadata: LMCacheE
         s, d = SafeSerializer(), SafeDeserializer()
     elif serde_type == "cachegen":
         s, d = CacheGenSerializer(config, metadata), CacheGenDeserializer(config, metadata)
+    elif serde_type == "fast":
+        s, d = FastSerializer(), FastDeserializer()
     else:
         raise ValueError(f"Invalid serde type: {serde_type}")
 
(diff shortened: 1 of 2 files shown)


lmcache/storage_backend/serde/fast_serde.py
```diff
@@ -0,0 +1,28 @@
...
+
+logger = init_logger(__name__)
+
+class FastSerializer(Serializer):
+    def __init__(self):
+        super().__init__()
+
+    def to_bytes(self, t: torch.Tensor) -> bytes:
+        # FIXME: only support fp16 for now
+        assert t.dtype == torch.float16
```
ApostaC: vLLM will use bfloat16 on devices supporting compute compatibility >= 8.0 (including A-series, H-series, L40, etc).
Therefore, it would be very helpful to support flexible data types. 
For instance, maybe we can add a separate tag in the byte stream for the data type?

lmcache/storage_backend/serde/__init__.py
```diff
@@ -12,6 +13,8 @@ def CreateSerde(serde_type: str, config: LMCacheEngineConfig, metadata: LMCacheE
         s, d = SafeSerializer(), SafeDeserializer()
     elif serde_type == "cachegen":
         s, d = CacheGenSerializer(config, metadata), CacheGenDeserializer(config, metadata)
+    elif serde_type == "fast":
```
ApostaC, line 16: I would like to directly replace the current "torch serializer/deserializer". But depends on your preference.

ApostaC (commented):
Hi Luke, thank you very much for contributing to the LMCache project!
This PR looks good to me in general. Just some minor comments.

Also, feel free to let me know if you would like to contribute more or have further collaboration. My email is yihua98@uchicago.edu

lmcache/storage_backend/serde/fast_serde.py
xenshinu: Totally agree. Just to give a rough example here. Let me think how to make it more flexible.

lmcache/storage_backend/serde/fast_serde.py
```diff
@@ -0,0 +1,42 @@
...
+class FastSerializer(Serializer):
+    def __init__(self):
+        super().__init__()
+
+    def to_bytes(self, t: torch.Tensor) -> bytes:
+        # make dtype into bit stream
+        tag = DTYPE_TO_TAG[t.dtype]
+        # make tensor into bit stream
+        buf = t.contiguous().cpu().view(torch.uint8).numpy().tobytes()
+        return tag.to_bytes(1, byteorder='big') + buf
```
xenshinu: Add `dtype` tag. But this will trigger memory copy on deser, maybe attach the dtype to meta data is a better solution.

ApostaC (approved):
Looks good to me! Thanks!
@YaoJiayi @KuntaiDu Feel free to take a look and let us know if you have any last-minute comments. Otherwise, let's merge it next Monday

xenshinu:
> Looks good to me! Thanks! @YaoJiayi @KuntaiDu Feel free to take a look and let us know if you have any last-minute comments. Otherwise, let's merge it next Monday

Hey, sorry about the confusion. I mean adding the tag will require an additional slicing when rebuilding the tensor and will slow down the program. Instead of attaching the tag to the bitstream, adding an additional entry to meta data could be a better solution.

ApostaC:
> Hey, sorry about the confusion. I mean adding the tag will require an additional slicing and will slow down the program. Instead of attaching the tag to the bitstream, adding an additional entry to meta data could be a better solution.

Oh, let me clarify. We already have the plan to add the dtype in metadata and it's in progress. We can first merge your PR and then update this code again once the code change about metadata is ready.

PR #106, opened 2024-09-23
Fix issue 104 -- `store` is slow on CPU

### LINKED ISSUE #104: [v0.1.x] `store` is slow when using CPU

ApostaC:
This PR resolves #104 
There are no important changes on the LMCache side. The major changes can be found in `lmcache-vllm` repo (commit [eebfe3](https://github.com/LMCache/lmcache-vllm/commit/eebfe331c9dee88dab4bbe8f7195d8a6e9821a41))

### DIFF
--- a/lmcache/storage_backend/local_backend.py
+++ b/lmcache/storage_backend/local_backend.py
@@ -48,7 +48,7 @@ class LMCLocalBackend(LMCBackendInterface):
         self.update_lock = threading.Lock()
         
         # FIXME(Jiayi): `use_pin_memory` and `dst_device` should be configged dynamically
-        self.use_pin_memory = True
+        self.use_pin_memory = False
         logger.info(f"Using pinned cpu memory: {self.use_pin_memory}")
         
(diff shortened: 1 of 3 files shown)


lmcache/storage_backend/local_backend.py
```diff
@@ -48,7 +48,7 @@ def __init__(
         self.update_lock = threading.Lock()
         
         # FIXME(Jiayi): `use_pin_memory` and `dst_device` should be configged dynamically
-        self.use_pin_memory = True
+        self.use_pin_memory = False
```
KuntaiDu, line 51: Why `False` here

ApostaC: Did some local tests and found that `pin_memory = True` will lead to some extra overhead. Maybe there are some performance bugs. 

Created another ticket #107 for this. Let's investigate that later

KuntaiDu (approved):
LGTM

PR #115, opened 2024-09-24
Add python & CUDA requirement to README

qyy2003:


ApostaC (approved):
Thanks, looks good to me!

PR #126, opened 2024-09-27
Add format checking github workflow file

KuntaiDu:
Add a format checking github workflow. Need to further modify the repo's github action so that these files can take effect


ApostaC (approved):
LGTM!

PR #127, opened 2024-09-27
update mypy to exclude checking types in test

KuntaiDu:
RT


ApostaC (approved):
LGTM! Thanks, Kuntai! Let me merge it

PR #128, opened 2024-09-28
Add and organize examples

### LINKED ISSUE #124: [v0.1.x] (Documentation) Create new examples for LMCache and readmes for the examples

XbzOnGit:
This PR resolves #124  

The main components:  
1. offline inference  
offline inference on vLLM + LMCache.  
2. online session  
Online chat session on vLLM + LMCache.  
3. frontend  
Online chat on vLLM + LMCache with a QA frontend.  
Default context is ffmpeg man page.  
4. share_across_instances  
Show LMCache can share KV cache across serving instances.  
Ask the same question in the same context(ffmpeg man page) on two serving instances.  
5. storage_backends  
How to write yamls to configure storage backend.


YaoJiayi:
Thanks, Sixian! Quick question: have you seen improvement in the `online_session` example?
The reason why I'm asking this is https://github.com/LMCache/lmcache-tests/pull/11 .

XbzOnGit:
> Thanks, Sixian! Quick question: have you seen improvement in the `online_session` example? The reason why I'm asking this is [LMCache/lmcache-tests#11](https://github.com/LMCache/lmcache-tests/pull/11) .

In online_session, I didn't feed a long context so it is hard to see improvements.  
But in a long-context version of it(frontend), I can see improvements.  

So let's look at frontend example instead.  
The only differences between online_session and frontend are 
1. frontend has a long context of 20000+ tokens.  
2. frontend has a UI.  

And in frontend example, I CAN actually see an obvious improvement.  
(TTFT from 7+ secons to 1- seconds.)  

And from the loggings, lmcache server is also accessed.  

I once encountered a similar problem(LMCache not enabled at all), and it finally turned out to be a python import problem.  
The monkey patching in lmcache_vllm vllm_injection.py, InitLMCacheEnvironment fails because of importing the same module twice(due to a python environment path problem).  
I think this is exactly the problem you mentioned.  
I worked around it by pip install ./ (without -e) everytime. This can build from source, while solving this problem.

YaoJiayi:
Noted. Thanks!

YaoJiayi (approved):

YaoJiayi:
About the performance:
Based on my observation, on Mistral-7b, prefilling 24k tokens takes ~4s and prfilling w. local cpu cache takes ~1.2s.

XbzOnGit:
> About the performance: Based on my observation, on Mistral-7b, prefilling 24k tokens takes ~4s and prfilling w. local cpu cache takes ~1.2s.

Thanks, the data I provided above is not on nature server. It can be a hardware difference.

PR #138, opened 2024-10-04
Update examples and change launch methods

XbzOnGit:
1. Update examples with long contexts.  
2. Change launch methods.


YaoJiayi:
LGTM. Thanks @XbzOnGit ! Btw, just to make sure, we can now see the performance improvement for all the examples right?

XbzOnGit:
> LGTM. Thanks @XbzOnGit ! Btw, just to make sure, we can now see the performance improvement for all the examples right?

Yes, they are all expected to show improvements now.

YaoJiayi (approved):

Issue #135, opened 2024-10-03
Llama-3.1 LMCacheEngine Store Problem


Hanchenli:
While adding CacheGen support for Llama-3.1-8B-Instruct, the lmcache_vllm server will encounter a bug when the the cache is being retrieved and reused.

Initial debugging seems to suggest the issue is with the store function inside LMCacheEngine class (when cache was first put inside storage server). 

With the context in example/frontend, the log looks like this:
```
INFO LMCache: Stored/updated 2 chunks, total time 0.11s, make chunks time 0.00s [2024-10-03 16:45:58,087.087]                                                                                                                
INFO LMCache: Using default batched implementation of the get() method [2024-10-03 16:45:58,109.109]
```
I did some initial debugging and printing the "len(tokens)" in the store function gives 512 (just one chunk). According to @YaoJiayi , this should not be correct and we suspect there might be something wrong with how "store" is called in this case. 

The original mistralai/Mistral-7B-Instruct-v0.2 works normally and logs look something like:
```
INFO LMCache: Stored/updated 100 chunks, total time 8.23s, make chunks time 0.01s [2024-10-03 17:30:47,160.160]
DEBUG LMCache: Serialization took 1.09 seconds [2024-10-03 17:30:47,421.421]
DEBUG LMCache: Call to set()! [2024-10-03 17:30:47,421.421]
```

YaoJiayi:
Related to #80

PR #143, opened 2024-10-04
Add support for llama-3.1-8b-instruct and qwen-2-7b

### LINKED ISSUE #80: [v0.1.x] (Functionality) CacheGen support for different models
# Description

Currently, CacheGen only supports a few models. 

CacheGen will perform best if its configuration is tuned for each model. But we should also have a "default fallback" for un-tuned models. 

This may require a refactor of the current CacheGen configuration code

# How to test

This should be covered by some end-to-end tests. When passing some CacheGen-untuned models, it should still be able to reduce TTFT compared to using safetensor compression

YaoJiayi:
This request resolves #80 #135 .
`Qwen/Qwen-7B` and `meta-llama/Llama-3.1-8B-Instruct` are added.

Note that for models that has context window > 32768, we need to use `enable_chunked_prefill=False` and `max_model_len=32768`. Chunked prefill is automatically enabled in vllm when context window >32768.

@Siddhant-Ray Can we add corresponding document for the models we currently support?
@Shaoting-Feng Can we add tests for more models?


Siddhant-Ray:
Will add documentation for other models by this weekend.

Shaoting-Feng:
Will add tests for other models by this weekend.

examples/offline_inference/offline_inference.py
```diff
@@ -7,6 +7,8 @@
 from transformers import AutoTokenizer
 
 model_name = "mistralai/Mistral-7B-Instruct-v0.2"
+#model_name = "Qwen/Qwen-7B"
+#model_name = "meta-llama/Llama-3.1-8B-Instruct"
```
KuntaiDu, line 11: Would be great if you can remove these 2 line

KuntaiDu (approved):
LGTM!

ApostaC:
As a side note, we still want to do some refactoring for the CacheGenConfig and have a "default configuration" for different models.

Just created a new ticket for this: #148

Issue #142, opened 2024-10-04
Performance improvement might not be genuine due to warm up issues


YaoJiayi:
In the examples, the first run could be slow due to cuda warmup issues.

YaoJiayi:
Also we should specify the expected improvement in each readme.

PR #149, opened 2024-10-08
Support saving decode KV cache to boost multi-turn conversation

### LINKED ISSUE #129: Support saving KV cache after decoding
Currently, KV cache is saved after prefilling. Saving KV cache after decoding may be advantageous for applications such as multi-turn conversation.

YaoJiayi:
This PR resolves #129 .

Currently, decode cache is saved if seqlen (prompt tokens+generated tokens) % chunk_size == 0
By default, this feature is disabled.
To enable this feature, we need to specify "save_decode_cache: True" in config yaml. 
@Siddhant-Ray Please update the doc as well.

Currently, we only support mistral-7b.
There should be a patch per model. For example, in mistral-7b, assume the first run is text 1 -> text 2. Then input of the second run should take in text 1 + text 2. However, there's always an empty padding token 28705 between text 1 and text 2, making the cache for text 2 not reusable.
See https://github.com/LMCache/lmcache-vllm/pull/17/files#diff-ebeb6ecf30c81dd3fac7c2a4ec1b9ed188a07dd005a6fa94b1d5eabfcea68d1aR175 for solution.
Created an issue to support more models #151 .

@Shaoting-Feng There should be more tests for this feature.

@XbzOnGit There's a naive example at https://github.com/LMCache/LMCache/pull/149/files#diff-3fb66e5f1f5c6e784cf25f543303d240b9130723481a3ea0481d3e38dcfbcb1dR1 .
Please refine this example and also other examples (see #142).

Created an issue to improve the performance #150.

### DIFF
--- a/examples/save_decode_cache/offline_inference.py
+++ b/examples/save_decode_cache/offline_inference.py
@@ -1,0 +1,121 @@
+import copy
+import json
+import os
+import time
+
+from lmcache_vllm.vllm import LLM, SamplingParams
+from transformers import AutoTokenizer
+
+model_name = "mistralai/Mistral-7B-Instruct-v0.2"
+#model_name = "Qwen/Qwen-7B"
+#model_name = "meta-llama/Llama-3.1-8B-Instruct"
+context_file = os.path.join(os.pardir, 'ffmpeg.txt')
+output_file = "offline_inference_outputs.jsonl"
+
+context_text = None
+with open(context_file, 'r') as f:
+    context_text = f.read()
+assert context_text is not None
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+
+context_messages = [
+    {
+        "role":
+        "user",
+        "content":
+        "You are a helpgul assistant."
+    },
+    {
+        "role": "assistant",
+        "content": "Got it."
+    },
+]
+user_inputs_batch = [
+    "What is FFmpeg?"
+    "Please include some details."
+    "Your answer should be around 5k words",
+]
+
+
+def get_context_length(tokenizer, context_messages):
+    return len(tokenizer.apply_chat_template(context_messages, tokenize=False))
+
+
+def gen_prompts(tokenizer, context_messages, user_inputs_of_batch):
+    generated_prompts = []
+    for user_input in user_inputs_of_batch:
+        copyed_context_messages = copy.deepcopy(context_messages)
+        copyed_context_messages.append({"role": "user", "content": user_input})
+        generated_prompts.append(
+            tokenizer.apply_chat_template(copyed_context_messages,
+                                          tokenize=False))
+    return generated_prompts
+
+
+def append_outputs(output_file_name, outputs, context_length, time_taken):
+    user_inputs = []
+    generated_texts = []
+    for output in outputs:
+        prompt = output.prompt
+        user_input = prompt[context_length:]
+        user_inputs.append(user_input)
+        generated_text = output.outputs[0].text
+        generated_texts.append(f"{generated_text!r}")
+    json_dict = {
+        "user_inputs": user_inputs,
+        "generated_texts": generated_texts,
+        "time in seconds": time_taken
+    }
+    with open(output_file_name, "a") as f:
+        f.write(json.dumps(json_dict) + '\n')
+
+
+context_length = get_context_length(tokenizer, context_messages)
+# Create a sampling params object.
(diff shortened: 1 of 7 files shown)


examples/save_decode_cache/offline_inference.py
```diff
@@ -0,0 +1,112 @@
...
+import json
+import os
+import time
+
+from lmcache_vllm.vllm import LLM, SamplingParams
+from transformers import AutoTokenizer
+
+model_name = "mistralai/Mistral-7B-Instruct-v0.2"
+#model_name = "Qwen/Qwen-7B"
+#model_name = "meta-llama/Llama-3.1-8B-Instruct"
```
KuntaiDu: Maybe clean comments

examples/save_decode_cache/offline_inference.py
```diff
@@ -0,0 +1,112 @@
...
+context_text = None
+with open(context_file, 'r') as f:
+    context_text = f.read()
+assert context_text is not None
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+
+context_messages = [
+    {
+        "role": "user",
+        "content": "You are a helpgul assistant."
```
KuntaiDu: helpgul --> helpful

examples/save_decode_cache/offline_inference.py
```diff
@@ -0,0 +1,112 @@
...
+user_inputs_batch = [
+    "Score your answer from 1-10",
+]
+context_length = get_context_length(tokenizer, context_messages)
+prompts = gen_prompts(tokenizer, context_messages, user_inputs_batch)
+t3 = time.perf_counter()
+second_outputs = llm.generate(prompts, sampling_params)
+t4 = time.perf_counter()
+print(f"\n\nSecond request Time: {t4 - t3} seconds\n\n")
+append_outputs(output_file, second_outputs, context_length, t4 - t3)
```
KuntaiDu, line 111: I feel like this file is a bit long -- but let's fix that later.

KuntaiDu (changes requested):

ApostaC (commented):
Looks good to me! Please don't forget to let Siddhant update the documentation about the new configuration item.
Let's merge it after Kuntai approves

KuntaiDu (approved):
LGTM

Shaoting-Feng:
This has been E2E tested. **TTFT doesn't reduce; retrieved number of chunks looks good; `prompt_token_ids` confusing.** @YaoJiayi 
* E2E Test Setting
  * Suppose text1 is the prompt (context + question) of request 1 and text2 is the output message of request 1. Then text1 + text2 is the context of request 2. In this logic, I ran five-turn tests. The number of tokens goes from <100 to >15000.
* Performance: 
  * TTFT doesn't reduce
    |      | TTFT1 | TTFT2 | TTFT3 | TTFT4 | TTFT5 |
    |------|-------|-------|-------|-------|-------|
    | save_decode_cache: True  | 0.196 | 0.281 | 0.672 | 1.432 | 3.570 |
    | save_decode_cache: False | 0.094 | 0.558 | 0.684 | 1.529 | 3.769 |
  * From `stderr.log`, retrieved number of chunks: 
    * `save_decode_cache = True`: 0 0 2 10 25
    * `save_decode_cache = False`: 0 0 2 2 2
  * From `stdout.log`, last four IDs (after 28705) in `prompt_token_ids` are different in the next turn of request

Shaoting-Feng:
* About the number of retrieved chunks: 
  * Only the fifth round contain 15000 tokens. The fourth round contain 6538 tokens, so the test result of the number of retrieved chunks is correct.
* About the TTFT:
Screen Shot 2024-10-21 at 03 57 18 AM

LifeIsHardBruh:
> * About the number of retrieved chunks:
>   
>   * Only the fifth round contain 15000 tokens. The fourth round contain 6538 tokens, so the test result of the number of retrieved chunks is correct.
> * About the TTFT:
>   Screen Shot 2024-10-21 at 03 57 18 AM

Hi, I have a question regarding this E2E test: I am wondering why the TTFTs without LMCache are lower than the TTFTs with LMCache in later turns. Isn't reusing KV caches going to reduce the context-processing delay and TTFTs?

ApostaC:
@LifeIsHardBruh Hey, the above figure is extremely outdated. Current lmcache performance should not be worse than no-lmcache

Issue #145, opened 2024-10-07
LMCache and VLLM's internal prefix caching need to be compatible 


YaoJiayi:

Shaoting-Feng:
This has been E2E tested.
* Function
  * `test_cache_compatibility` in `tests/tests.py`
* Configuration
  * Two servers with LMCache
  * Different `gpu_memory_utilization` for vllm
* Results
  * stored and retrieved same amount of chunks
  * larger `gpu_memory_utilization` provides better TTFT performance (engine0: `gpu_memory_utilization` = 0.5; engine1: `gpu_memory_utilization` = 0.8)
    <img src="https://github.com/user-attachments/assets/65a2d934-3416-4040-b07b-7db49c3f3fd8" width="50%">
    <img src="https://github.com/user-attachments/assets/76b76e90-2a79-4ced-95c4-319c57c36485" width="50%">

Issue #144, opened 2024-10-04
Support chunked prefill


YaoJiayi:

YaoJiayi:
A more efficicient implementation should be in a better injection place. Trying to have a less efficeint version now.

Shaoting-Feng:
engine 0: enable_chunked_prefill = False
engine 1: enable_chunked_prefill = True

<img src="https://github.com/user-attachments/assets/66e41a90-10c0-46d6-9f3e-4773c34a1b0f" alt="Screen Shot 2024-10-21 at 05 10 00 AM" width="50%"/>
<img src="https://github.com/user-attachments/assets/96e195fb-9ecd-4554-bb2b-27aec2221057" alt="Screen Shot 2024-10-21 at 05 10 47 AM" width="50%"/>
<img src="https://github.com/user-attachments/assets/51aceb4c-b329-4c6b-9aaf-ef7e4f7e0d55" alt="Screen Shot 2024-10-21 at 05 11 03 AM" width="50%"/>

PR #153, opened 2024-10-11
Adding a mask for retrieve to avoid repetitive data loading

YaoJiayi:
This PR is needed for #144 (chunk prefill) and #145 (vllm's internal prefix caching).


lmcache/cache_engine.py
```diff
@@ -301,17 +308,29 @@ def retrieve(
...
 
-            num_tokens: the number of tokens in the kv cache
+            ret_mask: indicate which tokens are retrieved
         """
+        num_skip_chunk = 0
+        num_skip_tok = 0
+        ret_mask = torch.ones_like(tokens, dtype=torch.bool)
+        if mask is not None:
+            num_skip_tok = (len(mask) - torch.sum(mask))
+            num_skip_chunk = num_skip_tok // self.chunk_size
```
ApostaC, line 328: What if the `mask` does not align with the chunk boundary? Will the `num_skip_chunk` still make sense?
Please clarify if you had any extra assumption on `mask`

ApostaC (changes requested):
Thanks, Jiayi! Please see my question below:

lmcache/cache_engine.py
YaoJiayi: `num_skip_chunk` will still make sense. For example, if the mask is xxxvvv (only want to retrieve the last three tokens) and the chunk_size is 2, then we will skip the first chunk (the first two tokens) and still retrieve the third token even though it's x. This is because we save the cache chunk by chunk.

ApostaC: I see, so it should work for any "suffix mask", right? Do you support (or plan to support) arbitrary masks (like xxvvxvxv)?

YaoJiayi: Yes. But I think we should have a separate pr for this. Currently, this is only needed for (1) chunk prefill and (2) compatibility with vllm's internal prefix caching. Suffix mask would suffice for both cases.

ApostaC: Sounds great! Can you maybe add a few sentences in the docstring about this? (I mean currently the mask should be "suffix")

YaoJiayi: Done.

ApostaC (approved):
LGTM

PR #156, opened 2024-10-15
cachegen config refactor

### LINKED ISSUE #148: [v0.1.x] Refactor the CacheGen configuration part and add a "default configuration" for all models
# Description

The current `CacheGenConfig` has a bunch of fields that are difficult to understand and extend.

```python
@dataclass
class CacheGenConfig:
    key_first_layers: int
    key_second_layers: int
    key_third_layers: int
    key_first_bins: int
    key_second_bins: int
    key_third_bins: int
    value_first_layers: int
    value_first_bins: int
    value_second_bins: int
```

We'd better refactor it for better readability and extendability.
Here's a brief proposal:
```python
@dataclass
class QuantizationSpec:
    start_layer: int
    end_layer: int    
    bins: int

class CacheGenConfig:
    nlayers: int
    kspecs: List[QuantizationSpec]
    vspecs: List[QuantizationSpec]
```

A quick problem with this proposal is that it does not force `kspecs[i].end_layer == kspecs[i+1].start_layer`. One potential solution is to remove the `end_layer` and assume that the next element's `start_layer` is the current element's `end_layer`.

----

Also, the current implementation only supports a limited number of models and does not have a "default" configuration for all models. Maybe it will be better to have one default configuration.
In the future, we should have a automatic tool to profile those configurations and select the best configuration for different models.

Oasis-Git:
This PR resolves https://github.com/LMCache/LMCache/issues/148

Modification on CacheGenConfig.
No default is provided unless the default model type is fixed.


Oasis-Git:
Format Mismatch. Close the PR temporarily.

YaoJiayi:
bash format.sh

Oasis-Git:
Test Pass. Please Review

lmcache/storage_backend/serde/cachegen_basics.py
```diff
@@ -39,40 +43,43 @@ def from_model_name(model_name: str) -> "CacheGenConfig":
...
+                nlayers=40,
+                kspecs=[
+                    QuantizationSpec(start_layer=0, end_layer=10, bins=32),
+                    QuantizationSpec(start_layer=10, end_layer=20, bins=16),
+                    QuantizationSpec(start_layer=20, end_layer=40, bins=16),
+                ],
+                vspecs=[
+                    QuantizationSpec(start_layer=0, end_layer=2, bins=32),
+                    QuantizationSpec(start_layer=2, end_layer=40, bins=16),
+                ],
```
YaoJiayi, line 80: Is there any support for the default setting (for models other than these five models)? For exmple, we can pass the model_config in vllm (which contains info about the number of layers) or the number of layers directly as the function param, and configure the `CacheGenConfig` dynamically based on this input.

YaoJiayi (commented):
Thanks @Oasis-Git! I left a comment:)

ApostaC (dismissed):
Looks good to me! Thanks @Oasis-Git 
I will merge it soon

ApostaC (approved):
Let's merge it now

PR #158, opened 2024-10-15
Examples update

### LINKED ISSUE #142: Performance improvement might not be genuine due to warm up issues

XbzOnGit:
This resolves #142 

Update examples with prewarm if we are comparing the first with the second.  
Add some numbers in README.md  
Add vllm prefix cache example

### DIFF
--- a/lmcache/cache_engine.py
+++ b/lmcache/cache_engine.py
@@ -83,11 +83,20 @@ class LMCacheEngine:
         for i in range(0, len(tokens), self.chunk_size):
             yield tokens[i:i + self.chunk_size]
 
-    def _prefix_hash(self, token_chunks: Iterable[torch.Tensor]) -> List[str]:
+    def _prefix_hash(
+        self,
+        token_chunks: Iterable[torch.Tensor],
+        chunked_skip_len: Optional[int] = None,
+    ) -> List[str]:
         prefix_hash = self._get_init_hash()
         prefix_hashes = []
+        tok_idx = 0
         for token_chunk in token_chunks:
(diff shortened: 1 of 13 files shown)


lmcache/cache_engine.py
```diff
@@ -301,17 +313,30 @@ def retrieve(
...
 
         Output:
             kv_tensors: the kv cache of the tokens, in the format of nested 
                         tuples. Will be an empty tuple if no kv cache is 
                         retrieved.
 
-            num_tokens: the number of tokens in the kv cache
+            ret_mask: indicate which tokens are retrieved
         """
+        chunked_skip_len = 0
```
ApostaC, line 327: The changes here seem to conflict with #153. Did you pulled/picked some old codes from there?

ApostaC (commented):
Thanks Sixian! One quick question about the changes in the `cache_engine.py`:

lmcache/cache_engine.py
YaoJiayi: Another side note. @KuntaiDu and I have talked about these examples. They contain too much repetitive code. I'll refactor them later.

YaoJiayi (commented):
Another side note. @KuntaiDu and I have talked about these examples. They contain too much repetitive code. I'll refactor them later.

lmcache/cache_engine.py
XbzOnGit: There are new commits after I wrote this PR. I think simply keeping the code in #153 should work.

PR #161, opened 2024-10-16
Hanchen dev ("get_locations" and "remove" function)

### LINKED ISSUE #147: LMCacheEngine exists and delete functionality
Add two functions to current LMCacheEngine and tests
1. exists(key), takes key hash as input, returns a list of locations that this chunk is in (ex. local backend, remote backend).
2. remove(key, location), remove entry with key equal to key from location, returns 1 when is successful and 0 when is not successful.

Hanchenli:
This PR resolves https://github.com/LMCache/LMCache/issues/147

Added get_locations and remove for cache_engine and "where_is" & "remove" in storage_backend level. The cache_engine level functions uses the storage_backend level functions. I added initial unit tests in test_cache_eviction.py. 

get_locations iterates through the locations that the tokens are stored and return their entries. With get_locations, users can have an overview of how the current prefix is cached inside the system. It is implemented by visiting storage backends for each kv cache.

remove removes the KV cache of specified blocks in a prefix. This allows program to manipulate KV cache from CacheEngine level: ex. user deletes a multi-turn conversation history and we can manually delete the KV cache for them.

Notice that this pr also contains "where_is: and "remove" in the storage_backend level, which can help with writing the evictor (using them in a way similar to how "get_locations" and "remove" uses them in cache_engine level). The where_is searches for all the storage locations of a CacheKey and remove removes one entry with the CacheKey from the specified locations.


examples/frontend/chat_session.py
```diff
@@ -59,9 +60,10 @@ def chat(self, question):
         server_message = []
         for chunk in chat_completion:
             chunk_message = chunk.choices[0].delta.content
-            if chunk_message is not None:
+            if chunk_message is not None and chunk_message != "":
```
KuntaiDu, line 63: Maybe `if chunk_message`?

examples/frontend/chat_session.py
```diff
@@ -59,9 +60,10 @@ def chat(self, question):
...
                 if end is None:
-                    end = time.perf_counter()
+                    end = time.time()
+                    print(end)
```
KuntaiDu: `print(end)` is a bit not user-friendly -- they have to manually do the difference to get how many time is spent on running the code.

lmcache/cache_engine.py
```diff
@@ -294,7 +294,7 @@ def retrieve(
         Input:
             tokens: the input tokens, with shape [seq_len]
 
-            format: either 'huggingface' or 'vllm'
+            format is either 'huggingface' or 'vllm'
```
KuntaiDu: Maybe still use ':', or change the above ':' also to `is` for consistency

lmcache/cache_engine.py
```diff
@@ -348,6 +348,93 @@ def retrieve(
...
+    ) -> Tuple[bool, List[bool]]:
+        """
+        Remove the KV cache entry from the locations specified. 
+        Tokens has to be one block. 
+        Returns validity of token_block and success of deletions
+
+        Input:
+            token_block: the input tokens block, with shape [chunk_size]
+
+            format is either 'huggingface' or 'vllm'
```
KuntaiDu, line 402: Ditto

lmcache/cache_engine.py
```diff
@@ -348,6 +348,93 @@ def retrieve(
...
+            or the last block is not full (ex. 13 token / (16 token per block)) 
+        """
+
+        chunk_hashes = self._prefix_hash(self._chunk_tokens(tokens))
+
+        ret_locations_list: list[Optional[list[str]]] = []
+        for chunk_hash in chunk_hashes:
+            list_locations = self.engine_.where_is(
+                self._make_key(chunk_hash, self.metadata.fmt))
+            if list_locations[0] == 'NOT IN CACHE':
```
KuntaiDu, line 378: A better design is to use `enum` class, but this will be a large-scale change so feel free to ignore it for now.

lmcache/storage_backend/abstract_backend.py
```diff
@@ -45,6 +45,26 @@ def contains(
...
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def remove(self, key: CacheEngineKey, location: str) -> bool:
+        """
+        Remove the key from the location
+        If backend is not the location, return False automatically.
+        """
+        #TODO NO Lock right now
```
KuntaiDu: Remove

examples/frontend/README.md
```diff
@@ -9,7 +9,9 @@ This will use the port 8000 (for vLLM), 8501 (for the frontend) and port 65432(f
 1.  ```lmcache_server localhost 65432```  
 And wait until it's ready.  
 2. In one terminal,  
-```LMCACHE_CONFIG_FILE=example.yaml CUDA_VISIBLE_DEVICES=0 python3 -m lmcache_vllm.vllm.entrypoints.openai.api_server --model mistralai/Mistral-7B-Instruct-v0.2 --gpu-memory-utilization 0.8 --port 8000```  
+```LMCACHE_CONFIG_FILE=example.yaml CUDA_VISIBLE_DEVICES=0 python3 -m lmcache_vllm.vllm.entrypoints.openai.api_server --model meta-llama/Llama-3.1-8B-Instruct --gpu-memory-utilization 0.8 --port 8000```
```
ApostaC, line 14: Can you elaborate why making the change here?

lmcache/cache_engine.py
ApostaC: +1. Please define an enum class for the locations, maybe in `utils.py`. Both this function and the `where_is` function should use this enum class.

lmcache/storage_backend/serde/cachegen_basics.py
```diff
@@ -74,6 +73,30 @@ def from_model_name(model_name: str) -> "CacheGenConfig":
                 value_first_bins=32,
                 value_second_bins=16,
             )
+        elif model_name == "meta-llama/Llama-3.1-8B-Instruct":
```
ApostaC, line 76: Seems like this does not belong to this PR?

ApostaC (changes requested):
Thanks, Hanchen! Two high-level comments + some detailed comments below:
1. Can you explain the design of `get_locations` and `remove`? Why having these two functions in the `LMCacheEngine` class? Who will call those functions? 
2. You added a few new functions in the connectors and backends. Can you add unit tests for those functions in the corresponding test files?

examples/frontend/README.md
Hanchenli: We can revert back. This is simply changing mistral to 3.1B for another task

lmcache/storage_backend/serde/cachegen_basics.py
Hanchenli: No

YaoJiayi:
> Thanks, Hanchen! Two high-level comments + some detailed comments below:
> 
>     1. Can you explain the design of `get_locations` and `remove`? Why having these two functions in the `LMCacheEngine` class? Who will call those functions?
> 
>     2. You added a few new functions in the connectors and backends. Can you add unit tests for those functions in the corresponding test files?

Side note: I need the "remove" functionality in my LRU Eviction (but only need it in every storage backend). Haven't created a PR yet. Will create a draft one tonight.

Hanchenli:
> > Thanks, Hanchen! Two high-level comments + some detailed comments below:
> > ```
> > 1. Can you explain the design of `get_locations` and `remove`? Why having these two functions in the `LMCacheEngine` class? Who will call those functions?
> > 
> > 2. You added a few new functions in the connectors and backends. Can you add unit tests for those functions in the corresponding test files?
> > ```
> 
> Side note: I need the "remove" functionality in my LRU Eviction (but only need it in every storage backend). Haven't created a PR yet. Will create a draft one tonight.



> Thanks, Hanchen! Two high-level comments + some detailed comments below:
> 
> 1. Can you explain the design of `get_locations` and `remove`? Why having these two functions in the `LMCacheEngine` class? Who will call those functions?
> 2. You added a few new functions in the connectors and backends. Can you add unit tests for those functions in the corresponding test files?

Hey Yihua, please check the newest explanation of PR. The "remove" that Jiayi needs on storage_backend level is also part of the PR .

Hanchenli:
> > Thanks, Hanchen! Two high-level comments + some detailed comments below:
> > ```
> > 1. Can you explain the design of `get_locations` and `remove`? Why having these two functions in the `LMCacheEngine` class? Who will call those functions?
> > 
> > 2. You added a few new functions in the connectors and backends. Can you add unit tests for those functions in the corresponding test files?
> > ```
> 
> Side note: I need the "remove" functionality in my LRU Eviction (but only need it in every storage backend). Haven't created a PR yet. Will create a draft one tonight.

Notice that this PR also has "remove" function implemented inside storage backend.

lmcache/cache_engine.py
Hanchenli: Let's fix this in the next version? This requires a lot of code change but has no impact on functionality.

XbzOnGit:
I'm kind of confused about how remove works with existing_keys in remote_backend.  
If there is a remote server A, a serving instance B(client), and a serving instance C(client).  
There is a KV cache chunk called E, which is put into A by C. This makes E inside C's local backend && C's existing_keys in remote backend.  
Then some requests to C makes C evict E from its local backend by calling remove.  
Then B removes E from remote server A.  
Then when C calls contains, it will return True because it is inside C's existing keys. This is not true because it has been removed by B.  

The problem is, currently the only self.existing_keys.remove(key) call happens on a serving instance calling remove on its remote backend. But the actual remote server is shared. It can be removed by other instances. But existing_keys is not changed by now.

Hanchenli:
> to



> I'm kind of confused about how remove works with existing_keys in remote_backend. If there is a remote server A, a serving instance B(client), and a serving instance C(client). There is a KV cache chunk called E, which is put into A by C. This makes E inside C's local backend && C's existing_keys in remote backend. Then some requests to C makes C evict E from its local backend by calling remove. Then B removes E from remote server A. Then when C calls contains, it will return True because it is inside C's existing keys. This is not true because it has been removed by B.
> 
> The problem is, currently the only self.existing_keys.remove(key) call happens on a serving instance calling remove on its remote backend. But the actual remote server is shared. It can be removed by other instances. But existing_keys is not changed by now.

Hey Sixian, I think you are correct. We may need to discuss whether to modify remove or contains for this case, thanks for pointing it out!

Hanchenli:
Changing this to #166

PR #164, opened 2024-10-17
Add lookup and allow passing a part of KV cache to store

### LINKED ISSUE #150: Reduce the overhead of extracting KV cache from paged memory if the cache is already in LMCache
For example, if we save KV cache for every 256 tokens generated, we should only extract the KV Cache of the latest 256 tokens.

XbzOnGit:
This resolves  #150.  
By calling lookup with align_to_chunk set to true, and invert the bool tensor, we can get a tensor mask with missing tokens set to true, and False existing tokens a multiple of chunk_size.  
Only copy those out of computation memory, then call store with this mask. store function will only store these KV cache.


YaoJiayi (commented):
Thanks @XbzOnGit ! The unit tests are failing?

XbzOnGit:
> Thanks @XbzOnGit ! The unit tests are failing?

Fixed

lmcache/cache_engine.py
```diff
@@ -380,6 +406,49 @@ def retrieve(
...
+    @torch.no_grad()
+    def lookup(
+        self,
+        tokens: torch.Tensor,
+        align_to_chunk: bool,
+    ) -> torch.Tensor:
+        """
+        Checks the existence of KV cache of the tokens from the cache engine.
+        The return should be a boolean tensor indicating the existence of 
+        the KV cache of each token.
```
ApostaC: Is the returned mask always a prefix (e.g., like TTTTTTFFF)? If so, maybe explicitly say that in the docstring.

lmcache/cache_engine.py
```diff
@@ -380,6 +406,49 @@ def retrieve(
 
         return ret, ret_mask
 
+    @_lmcache_nvtx_annotate
+    @torch.no_grad()
+    def lookup(
+        self,
+        tokens: torch.Tensor,
+        align_to_chunk: bool,
```
ApostaC: When will `align_to_chunk = True` be used?

lmcache/cache_engine.py
```diff
@@ -380,6 +406,49 @@ def retrieve(
...
+            interval_len = next_token_idx - current_token_idx
+            if interval_len < self.chunk_size and align_to_chunk:
+                break
+            if not self.engine_.contains(self._make_key(chunk_hash, fmt)):
+                break
+            else:
+                ret[current_token_idx:next_token_idx] = True
+            current_token_idx = next_token_idx
+            next_token_idx = min(next_token_idx + self.chunk_size,
+                                 total_token_cnt)
```
ApostaC: The logic in this for-loop seems a bit complicated. Can you explain it? Maybe we can try making it easier to understand

lmcache/cache_engine.py
```diff
@@ -231,6 +238,7 @@ def store(
         self,
         tokens: torch.Tensor,
         kv_tensors_raw: KVCache,
+        kv_tensors_mask: Optional[torch.Tensor] = None,
```
ApostaC, line 241: What's the assumption you made for kv_tensors_mask? I can see 2 different designs here:
1. `kv_tensors_mask`'s length is the same as number of tokens in in the `kv_tensors_raw`
2. number of `True`s in `kv_tensors_mask` is the same as number of tokens in the `kv_tensors_raw`

Can you also make it explicit in the docstring?

ApostaC: And will the `kv_tensors_mask` have arbitrary True/False in it? Or the Falses/Trues can only be the prefix?

lmcache/cache_engine.py
```diff
@@ -257,19 +267,35 @@ def store(
...
         assert (len(
             tokens.shape) == 1), f"Invalid shape of tokens: {tokens.shape}"
+        assert (len(
+            kv_tensors_mask.shape) == 1), \
+        f"Invalid shape of mask: {kv_tensors_mask.shape}"
+        assert len(tokens) == len(kv_tensors_mask), \
+            "token length does not match mask length"
+        num_skip_chunk = 0
+        num_skip_tok = 0
+        if kv_tensors_mask is not None:
```
ApostaC: Why having this if? `kv_tensors_mask` is already initialized in line 271

lmcache/cache_engine.py
```diff
@@ -257,19 +267,35 @@ def store(
...
+            # NOTE(Sixian): Now kv_tensors_mask always a suffix mask.
+            num_skip_tok = (len(kv_tensors_mask) - torch.sum(kv_tensors_mask))
+            num_skip_chunk = num_skip_tok // self.chunk_size
+        assert num_skip_tok == num_skip_chunk * self.chunk_size, \
+            "Only support chunk skip in store."
```
ApostaC: What do you mean by "Only support chunk skip in store"? Does it mean the `True`s in `kv_tensors_mask` should align to the chunk boundaries?

lmcache/cache_engine.py
```diff
@@ -257,19 +267,35 @@ def store(
...
+        assert num_skip_tok < len(tokens), "No tokens to store"
```
ApostaC: Do we need to throw a runtime error if there is no token to store? Why not just log a warning?

lmcache/cache_engine.py
```diff
@@ -185,11 +185,15 @@ def _make_chunks_skip_existing(
         tokens: torch.Tensor,
         kv_tensors: torch.Tensor,
         fmt: str,
+        num_skip_chunk=0,
```
ApostaC: The "physical meaning" of `num_skip_chunk` does not match its name. IIUC, the physical meaning here is "number of prefix chunks to skip". Can we have a better variable name?
@YaoJiayi feel free to chime in if you have any thoughts on this.

ApostaC (changes requested):
Thanks Sixian! In general, I feel like this PR introduces a lot of assumptions in the arguments, and those assumptions are not clearly stated in the docstring. 

Please see the detailed comments below.

As a side note, since you added a new function `lookup`, can you add a unit test for this function?

lmcache/cache_engine.py
XbzOnGit: When skipping some tokens in store(used together with kv_tensor_mask in store), it should skip a chunk multiple tokens. Here align_to_chunk should be True.

ApostaC (commented):
Note:
- unit test for lookup
- unit test for "kv_tensors_mask"

tests/test_cache_engine.py
```diff
@@ -287,6 +287,85 @@ def test_skipping(fmt, autorelease):
...
+    engine.store(tokens, kv_cache)
+    prefix_length = engine.lookup(tokens)
+    assert prefix_length == num_tokens, \
+        f"Expected {num_tokens} prefix tokens, but got {prefix_length}"
+    kv_tensor_mask = torch.ones_like(final_tokens, dtype=torch.bool)
+    align_to_chunk_cnt = (num_tokens // chunk_size) * chunk_size
+    kv_tensor_mask[:align_to_chunk_cnt] = False
+    more_cache_tokens = num_tokens + new_num_tokens - align_to_chunk_cnt
+    more_kv_cache = generate_kv_cache(more_cache_tokens, fmt, device)
+    engine.store(final_tokens, more_kv_cache, kv_tensor_mask)
```
ApostaC, line 362: Quick question: what should be the shape of `kv_tensors_raw` when calling `engine.store`? Should it have the same # of tokens as the tokens and `kv_tensors_mask`, or it should have the same # of tokens as the # of trues in the `kv_tensors_mask`?
My previous understanding is the former one, but here looks like `more_kv_cache` has less tokens than `final_tokens`

ApostaC (commented):
Thanks Sixian! Looks good to me in general. Just one question:

tests/test_cache_engine.py
XbzOnGit: The second. That's the way we reduce memory copy. If it is the first one, then no copy is reduced.

ApostaC: Okay, maybe the latter one also makes sense. I realized only a "short" KV tensor will be extracted on the lmcache-vllm side. 

However, I feel like this could create some inconsistencies:
- If there is no `kv_tensors_mask`, the KV tensor and the tokens should be the same "length" (meaning the same number of tokens)
- If there is `kv_tensors_mask`, the KV tensor will have a different length from the input tokens.

In that case, I have another interface proposal: `store(prefix_tokens, new_tokens, KV_tensors)`, and the KV tensors have the same number of tokens in `new_tokens`. But maybe we can do that in the future.

@YaoJiayi Feel free to let us know if you have any thoughts on this

XbzOnGit: > Quick question: what should be the shape of `kv_tensors_raw` when calling `engine.store`? Should it have the same # of tokens as the tokens and `kv_tensors_mask`, or it should have the same # of tokens as the # of trues in the `kv_tensors_mask`? My previous understanding is the former one, but here looks like `more_kv_cache` has less tokens than `final_tokens`

kv_tensors_mask means, if this position is true, then the corresponding token has its KV cache in kv_cache_raw passed into store.

ApostaC: > The second. That's the way we reduce memory copy. If it is the first one, then no copy is reduced.

Can you add this to the docstring? Specifically, documenting the shape/device requirements explicitly will be useful.

XbzOnGit: > > The second. That's the way we reduce memory copy. If it is the first one, then no copy is reduced.
> 
> Can you add this to the docstring? Specifically, documenting the shape/device requirements explicitly will be useful.

The docstring already has this:
```
            kv_tensors_mask: a boolean mask of tokens indicating which tokens'
            KV Cache should be stored. Only support suffix mask.
            len(kv_tensors_mask) should be the same as len(tokens)
            number of true should be the same as kv_tensors_raw token number
```
Do I need to write something else?

ApostaC: Oh! I missed the last sentence, sorry about that!
Personally, I feel like it would be very useful to say something in the `kv_tensors_raw`'s description. For example "When kv_tensors_mask is None, the number of tokens in kv_tensors_raw should be the same as the number of Trues in kv_tensors_mask. Otherwise, it should be the same as len(tokens)".

ApostaC (approved):
LGTM!

PR #196, opened 2024-10-31
Add Dockerfile and its Documentation

qyy2003:
Create Dockerfle, docker image and docker documentation for version `0.1.3`.

Fixed the Docker build's dependency on `vllm`.


Siddhant-Ray:
This PR is breaking other docs, don't merge. Let's do a review @qyy2003

Siddhant-Ray (changes requested):
These changes are breaking, the build needs to be fixed.
We don't have a directory called developer, instead we have advanced.

PR #197, opened 2024-10-31
Add Dockerfile and its Documentation

qyy2003:
Create Dockerfle, docker image and docker documentation for version ` 0.1.3`.

Fixed the Docker build's dependency on `vllm`.

Fixed Error in developer documentation's picture.

PR #199, opened 2024-10-31
Revert "Add Dockerfile and its Documentation"

ApostaC:
Reverts LMCache/LMCache#197

Let's don't introduce dependency on vLLM for now


Siddhant-Ray (approved):
lgtm

PR #209, opened 2024-11-06
[Enhancement] Drop special tokens in kv blending

XbzOnGit:
Drop special tokens, making them only used to separate prompt into segments but not seen by underlying serving system. 

This PR resolves #194 

drop_spt_and_get_indices should be used by tokenizers to remove the special tokens, and record the indices of segments.

new_request now accepts full prompts instead of input_ids(which can cause problem if the prefill of this request does not start on 0, like a vllm internal prefix cache hit or chunked prefill).
It also accepts an external indices because the special tokens should have been removed.

PR #215, opened 2024-11-08
[Misc]Add cacheblend hyper-parameter configurations

### LINKED ISSUE #190: [KV Blending] Make hyper-parameters configurable (e.g., recompute ratio)
The current KV blending module can only be controlled by `enable_blending` option in the configuration yaml.

However, multiple internal hyper-parameters, such as the recomputation ratio, could also impact the KV blending module.

Specifically, we need at least 2 more hyper-parameters:
- Recomputation ratio -- controls how much tokens' KV will be re-computed by the KV blending module.
- Minimum # of tokens to enable blending -- if the request is too short, we don't want to trigger KV blending logic.

This feature also requires changing the `blend_adapter.py` in the lmcache_vllm repo. See the code here:
https://github.com/LMCache/lmcache-vllm/blob/b890544e7efbaa2738902889c6f4ac56cc22ecd5/lmcache_vllm/blend_adapter.py#L19

------------
**Update:**

We also need to support changing special tokens

XbzOnGit:
Add cacheblend hyper-parameter configurations on recompute ratio(blend_recompute_ratio) and minimum token number to trigger cacheblend(blend_min_tokens).

This PR resolves #190


XbzOnGit:
Closing because changes included in this [PR](https://github.com/LMCache/LMCache/pull/209)

Issue #206, opened 2024-11-06
Flexible dtype support for old GPUs


ApostaC:
**Is your feature request related to a problem? Please describe.**
Currently, LMCache assumes vLLM uses bfloat16 and hardcoded type conversions everywhere. However, on older GPUs, such as V100, bfloat16 is not supported.
Therefore, to use LMCache on old GPUs, we need to eliminate all those hardcoded conversions.

**Describe the solution you'd like**
Based on PR #103 , we can read the types from the `LMCacheMetadata`.

**Additional context**
This is expected to be fixed together with #94

PR #217, opened 2024-11-08
Fixed Hardcoded Dtype

qyy2003:
Solved #206 

It seems nearly all the hardcoded types are in test file. (Ignored right now.)


qyy2003:
I have verified that the local CPU + remote disk setup works correctly with float16 on all four different serde methods.

The only remaining hardcoded element in the functional code is in the following snippet within the `validate_reverse_correctness` function, located in `lmcache-vllm/lmcache_vllm/utils/positional_encoding.py`:

```python
dumb_q = torch.rand((num_tokens, hidden_dim), device="cuda", dtype=torch.bfloat16)
dumb_k = torch.rand((num_tokens, hidden_dim), device="cuda", dtype=torch.bfloat16)
```

It seems to be test code, so I decided not to modify it.

YaoJiayi (commented):
Thanks @qyy2003 ! But the unit tests are failing. Could you fix it first?
Another side note: could you compare the performance improvement from bfloat16 to datatypes ike fp8?

qyy2003:
@YaoJiayi  Fixed unitest problem. Its system dtype is half but it rands float32 tensor.

qyy2003:
Locally, the test is ok. I guess the reason it fails in github current is that it uses the original unitest code.

lmcache/storage_backend/serde/cachegen_decoder.py
```diff
@@ -17,7 +17,7 @@
 @_lmcache_nvtx_annotate
 def quant(bins: int, xq: torch.Tensor, max1: float):
     C = bins // 2 - 1
-    x = xq / C * max1  # .to(torch.float16)
+    x = xq / C * max1  # .to(self.dtype)
```
ApostaC: Maybe just remove the comment in this line?

ApostaC (approved):
LGTM, thanks! Just one minor comment, feel free to fix that if you want.

PR #239, opened 2024-11-25
refactor(connector): support connector storing tensor directly without serde

### LINKED ISSUE #237: [refactor] move serde into connector
Currently, before we call `set()` or `get()` for remote backend storage, one of the serializer/deserializer must be called to convert between `tensor` and `bytes`.

I am going to introduce a new storage backend, which is based on RDMA, for LMCache. In such cases, serializer/deserializer is not necessary for 3 following reasons:

1. RDMA needs to register a memory in client side before communicating. In `set()`, it is not easy to get the accuracy memory size for `bytes`; similarly, it is also a puzzle to register a `bytes` memory in `get()`.
2. As mentioned in first point, we need to register a memory region before `get()`. That is to say, we must know the value length before calling `get()`, which is quite different from general KV database like `Redis`. As a result, we need another way to store metadata, for example store an additional meta-tensor in remote backend storage.
3. For higher performance, RDMA could communicate between GPU and remote directly, which means that we **DO NOT** want to transfer bytes in CPU. Instead, transferring tensor directly between GPU device and remote would save time.

For those reasons, I hope a simple refactor for `storage_backend` module. That is, move `serde` variable into connector such as `redis` and `lm`:
```
def CreateConnector(url: str, s: Serializer = None, d: Deserializer = None, device = None) -> RemoteConnector:
    # ...

    match parsed_url.connector_type:
        case "redis":
            if num_hosts == 1:
                host, port = parsed_url.hosts[0], parsed_url.ports[0]
                connector = RedisConnector(host, port, s, d)
            else:
                raise ValueError(
                    f"Redis connector only supports a single host, but got url:"
                    f" {url}")
        case "RDMAKV":
            if num_hosts == 1:
                host, port = parsed_url.hosts[0], parsed_url.ports[0]
                connector = RDMAKVConnector(host, port, device)
            else:
                raise ValueError(
                    f"RDMAKV connector only supports a single host, but got url:"
                    f" {url}")
    # ...
```
By doing so, `storage_backend` could let `connector` decide whether to use `serde` or not.

DellCurry:
A simply refactor to remote backend storage, just move `serde` into `connection`.

This resolves #237

### DIFF
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -110,7 +110,7 @@ class LMCacheEngineConfig:
         chunk_size = config.get("chunk_size", 256)
         local_device = config.get("local_device", None)
         remote_url = config.get("remote_url", None)
-        remote_serde = config.get("remote_serde", "torch")
+        remote_serde = config.get("remote_serde", None)
         pipelined_backend = config.get("pipelined_backend", False)
         save_decode_cache = config.get("save_decode_cache", False)
         enable_blending = config.get("enable_blending", False)
--- a/lmcache/storage_backend/connector/__init__.py
+++ b/lmcache/storage_backend/connector/__init__.py
@@ -57,7 +57,7 @@ def parse_remote_url(url: str) -> ParsedRemoteURL:
     return ParsedRemoteURL(connector_type, hosts, ports)
 
 
-def CreateConnector(url: str) -> RemoteConnector:
+def CreateConnector(url: str, device=None) -> RemoteConnector:
     """
     Creates the corresponding remote connector from the given URL.
     """
--- a/lmcache/storage_backend/connector/base_connector.py
+++ b/lmcache/storage_backend/connector/base_connector.py
@@ -1,11 +1,19 @@
 import abc
 import time
+from enum import Enum
 from typing import List, Optional
+
+import torch
 
 from lmcache.logging import init_logger
 from lmcache.utils import _lmcache_nvtx_annotate
 
 logger = init_logger(__name__)
+
+
+class ConnectorType(Enum):
+    BYTES = 1
+    TENSOR = 2
 
 
 class RemoteConnector(metaclass=abc.ABCMeta):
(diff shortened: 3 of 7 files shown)


YaoJiayi:
@DellCurry Thanks for the PR! Overall it looks good to me. Could you try fixing the format checker (run bash format.sh) and also the unit tests (run pytest under tests/ folder)?

DellCurry:
@YaoJiayi Of course! This pr is just a draft, I will fix syntax format and fill unit test later.

But besides, I also find some other problems and have no clear idea how to resolve it properly:

1. for unit tests, since we want to move serde into connector, the original test_connector, which just `get`/`put` bytes, is no longer valid. We need to call 'CreateSerde' and pass serializer/deserializer to the connector. Obviously, it is somehow not a single unit test, but a test combining `test_serde` and `test_connector`.
2. how to revise LMCPipelinedRemoteBackend, which has a single thread to do deserializer works. What if we just remove deserializer threads? Will it affect the performance?

Hopping some advice or opinion on it!

YaoJiayi:
For 1, feel free to make changes to unit tests:)
For 2, currently we are pielining deserialization and data transmission (from remote storage). So the performance will be affected. And why are we removing the deserialization thread in PipelinedBackend? In RDMA case (if KV cache doesn't have to be serialized/deserialized), then maybe there's no need to set up a PipelinedBackend when RDMA is used.

DellCurry:
> For 1, feel free to make changes to unit tests:) For 2, currently we are pielining deserialization and data transmission (from remote storage). So the performance will be affected. And why are we removing the deserialization thread in PipelinedBackend? In RDMA case (if KV cache doesn't have to be serialized/deserialized), then maybe there's no need to set up a PipelinedBackend when RDMA is used.

For 2, the question is, this pr aims to move `serde` into connector. As a result, we serialize/deserialize tensor inside connector, `get()`/`set()` would operate `Tensor` directly. That is to say, we can not do serialization/deserialization outside connector, not only RDMA but all backends.

To fix this, I come up with another way instead of moving `serde` inside connector. We can let `set()` take `Union[bytes, torch.Tensor]` as parameter, and `get()` return `Optional[bytes | torch.Tensor]`, and still do serialization/deserialization in upper layer if necessary. Maybe a rename for this pr and issue is needed.

DellCurry:
hi @YaoJiayi , sry for delay last week. I have rewrite this pr, to support both `Tensor` and `bytes` storage backend. Hoping for your advice!

YaoJiayi:
The current code for supporting both types of remote backends look good to me:)

DellCurry:
> The current code for supporting both types of remote backends look good to me:)

hi @YaoJiayi ! Wondering if some other steps to be approved?

ApostaC (approved):
Thanks for contributing to LMCache @DellCurry . This PR looks good to me!

PR #251, opened 2024-12-04
[Refactoring, Enhancement] Major refactor to support prefetch and finegrained pipelining

YaoJiayi:
This is a refactor PR to better support prefetching and layer-wise piplining.
Note that this PR only supports local cpu + local disk backend (w. prefetching).

TODOs:
- [ ] Support zero copy store/retrieve with optimized kernels
- [ ] Support prefetch/retrieve from vllm scheduler/ochestrator?
- [ ] Support more fine-grained pipelining (e.g., layer-wise pipelining)
- [ ] Flush in original features
- [ ] Testing & Bug fixing & Performance tuning
- [ ] Stress testing (see how long the entire system can run)
  

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>


lmcache/experimental/cache_engine.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+        for start, end, key in self.token_database.process_tokens(
+                tokens, mask):
+            if self.storage_manager.contains(key):
+                continue
+            # Allocate the memory object
+            num_tokens = end - start
+            kv_shape = self.gpu_connector.get_shape(num_tokens)
+            kv_dtype = self.metadata.kv_dtype
+            memory_obj = self.memory_allocator.allocate(kv_shape, kv_dtype)
```
ApostaC, line 109: Currently, `memory_allocator` is directly owned by `CacheEngine`. 
However, I'm thinking about an alternative design where `memory_allocator` is owned by `storage_manager` (which means storage_manager should provide an extra interface called `allocate`).

This is because
- `stroage_manager` needs to free the memory objects, so it is potentially better to handle the allocation and free in one class.
- `storage_manager` may need to call the eviction logic when memory allocation fails, otherwise, the eviction logic will be called by `CacheEngine`
- I think `CacheEngine` is more like a top-level "orchestrator" that calls the underlying components without too much complicated logic.

lmcache/experimental/cache_engine.py
```diff
@@ -0,0 +1,289 @@
...
+            ret_mask[start:end] = True
+
+            # NOTE(Jiayi): memory_obj doesn't have to be a pinned
+            # cpu tensor for the sake of performance.
+            # For example, disk->gpu is faster than disk->cpu->gpu.
+            # RDMA is another example.
+
+            # TODO(Jiayi): Move the memory object directly to
+            # paged memory to avoid data copies
+            self.gpu_connector.to_gpu(memory_obj, start, end, **kwargs)
```
ApostaC, line 170: When will the `memory_obj` be freed? (Just leaving a note here in case I forget this. Will check it in the storage manager's code)

lmcache/experimental/config.py
ApostaC, line 1: It looks like most configurations are the same as the current LMCache. Is it possible for us to reuse the current `config.py` instead of having a new `experimental/config.py`?

lmcache/experimental/storage_backend/abstract_backend.py
```diff
@@ -0,0 +1,91 @@
...
+    def contains(self, key: CacheEngineKey) -> bool:
+        """
+        Check whether key is in the storage backend. 
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def insert_key(self, key: CacheEngineKey, size: int):
+        """
+        Insert the keu after data is put to storage backend.
```
ApostaC: keu -> key

lmcache/experimental/storage_backend/abstract_backend.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @abc.abstractmethod
+    def contains(self, key: CacheEngineKey) -> bool:
+        """
+        Check whether key is in the storage backend. 
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def insert_key(self, key: CacheEngineKey, size: int):
```
ApostaC, line 40: When will this function be called and who will call this function? Is this more like a callback after "put_task" is done?

lmcache/experimental/storage_backend/local_disk_backend.py
```diff
@@ -0,0 +1,159 @@
...
+    ) -> None:
+        """
+        Blocking put function.
+        This is for debugging and testing purposes.
+        """
+        path = self._key_to_path(key)
+        kv_chunk = memory_obj.tensor
+        fmt_str = str(memory_obj.metadata.fmt.value)
+        save_disk(path, kv_chunk, fmt_str)
+        self.insert_key(key, memory_obj.get_size())
```
ApostaC, line 119: Seems like this function is not being called? Can we delete it?

lmcache/experimental/storage_backend/abstract_backend.py
```diff
@@ -0,0 +1,91 @@
...
+    def submit_prefetch_task(
+        self,
+        key: CacheEngineKey,
+    ) -> Optional[Future]:
+        """
+        An async function to get the MemoryObj from the storage backend.
+
+        :param CacheEngineKey key: The key of the MemoryObj.
+
+        :return: a future object
```
ApostaC: In what case it will return None? Please add a sentence to explain it a bit in the docstring

lmcache/experimental/storage_backend/local_disk_backend.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Blocking get function.
+        """
+        self.disk_lock.acquire()
+        if key not in self.dict:
+            self.disk_lock.release()
+            return None
+        path = self.dict[key].path
+        buffer_memory_obj = load_disk(path, self.dst_device)
+        self.disk_lock.release()
```
ApostaC, line 149: Should we do `self.disk_lock.release()` before calling the `load_disk`? Otherwise, this function may hold the lock for a very long time when reading the tensor from the disk.

lmcache/experimental/storage_backend/storage_manager.py
```diff
@@ -0,0 +1,250 @@
...
+        self.use_hot = config.local_cpu
+
+        #TODO: remove hardcode
+        dst_device = "cuda"
+        self.storage_backends: Dict[str, StorageBackendInterface] =\
+            CreateStorageBackends(config, metadata, dst_device)
+        self.prefetch_tasks: Dict[CacheEngineKey, Future] = {}
+        self.put_tasks: Dict[str, Dict[CacheEngineKey, Tuple[Future,
+                                                             MemoryObj]]] = {}
+        self.callback_executor = ThreadPoolExecutor(max_workers=1)
```
ApostaC: We can remove this executor as callbacks are non-blocking.

ApostaC (commented):
Thanks, Jiayi! Left a few comments as well as some thoughts, otherwise LGTM.

lmcache/experimental/config.py
YaoJiayi: Not sure if it's a good idea. The major difference between the new and the old config is that the new config allows the user to compose more flexible storage heirarchies (e.g., `local cpu` AND `local disk` instead of `local cpu` OR `local cpu`)

lmcache/experimental/storage_backend/abstract_backend.py
YaoJiayi: It's always called in the callback of `put`. Adding an abstract method here since all storage backends need to implement this function.

lmcache/experimental/storage_backend/local_disk_backend.py
YaoJiayi: It's very helpful for debugging. We can drop it once the refactored version becomes stable

lmcache/experimental/storage_backend/storage_manager.py
YaoJiayi: Done

lmcache/experimental/storage_backend/local_disk_backend.py
YaoJiayi: Not sure if it's a good idea. I intentionally lock this since I want to prioritize `load_disk` over `save_disk`. Multiple concurrent/parallel `load_disk` could increase throughput but it's a blocking function for now anyway.

lmcache/experimental/cache_engine.py
YaoJiayi: Makes sense to me. I can try to reduce the complexity in `CacheEngine` in the next version (i.e., 0-copy store/retrieve).

ApostaC (approved):
LGTM, give an approve here.

Do we have any benchmark results on this version? I guess we still need some changes on lmcache-vllm side?

YaoJiayi:
> LGTM, give an approve here.
> 
> Do we have any benchmark results on this version? I guess we still need some changes on lmcache-vllm side?

Will have e2e results after implementing 0-copy store/retrieve.

PR #256, opened 2024-12-05
Add Disk_Address_Manager

qyy2003:
# Storage Address Manager

### Overview

TL;DR; propose an **independent process** to globally manage the disk files in different LMCache Engines through file path.

This pull request introduces an independent process for managing disk file paths across different LMCache Engines globally. The proposed design revolves around two key components:

1. **`disk_address_manager`**
2. **`LMCDiskBackend`**

These components interact seamlessly to handle file storage and retrieval. Additionally, the eviction logic is moved from the LMCache Backend to the Address Manager, ensuring centralized management.

### Key Features

- **Centralized Disk File Management:** The `disk_address_manager` provides file paths for all read/write operations.
- **Integrated Eviction Logic:** Address Manager now handles eviction policies, simplifying backend responsibilities.

### How to Launch

To start the Disk Address Manager:

```bash
LMCACHE_CONFIG_FILE=example.yaml lmcache_disk_address_manager
```

The `LMCACHE_CONFIG_FILE` is shared between the Address Manager and LMCache Backend. A sample configuration might include:

```yaml
local_device: "disk_url://http://localhost:4322"
disk_path: "/local/local_disk/"
```

### Core Logic

1. **Chunk Storage:**
    When the LMCache Engine needs to store a batch of chunks:
   - The backend uses `batched_contains` to determine if chunks are stored/ storing.
   - This method interacts with the Address Manager’s `batched_write_check`, which provides paths for new chunks to be written.
   - This method cached the return path for `put` operation
2. **Write Workflow:**
    During a `put` operation:
   - The backend fetched the cached paths and writes the chunks asynchronously.
   - Once the written buffer reaches a predefined size (`written_buffer_size`), the backend notifies the Address Manager, marking the cache as ready for reading.
3. **Chunk Retrieval:**
    When retrieving chunks, the LMCache Engine:
   - Calls `batched_read_check` to obtain file paths for reading.

### Interface Functions

**`write_check(self, key_str: str, kv_size: float) -> str`**
 Determines the write path for a given key:

- Returns `""` if the key is already cached to prevent redundant writes.
- Returns a file path if the key is not cached 
- Initializes the cache entry with an empty placeholder.

**`write_ready(self, key_str: str, kv_size: float)`**
 Updates cache state after the backend completes a write:

- Transitions the cache entry from an empty placeholder to its actual file path and size.

**`read_check(self, key_str: str) -> str`**
 Checks if a key is ready for reading:

- Returns the file path if the key is in the cache and ready.
- Returns `""` otherwise, including when the cache entry is still writing.

<html><body>
<html><head></head><body><h3>Performance Insights</h3>
<p><strong>Test Methodology:</strong><br>
The performance was evaluated by running <code>examples/disk_offline_inference</code> twice. The first test was executed on GPU 0, followed by the second test on GPU 1. The baseline comparison was performed using a local backend on GPU 0.</p>
<p><strong>Results Summary:</strong></p>
<ul>
<li><strong>Communication Overhead:</strong> Retrieving file paths for chunks (25,514 tokens) takes approximately <strong>5–7 milliseconds</strong>.</li>
<li><strong>Overall Performance:</strong> The results are as expected, with performance influenced by cold starts and disk speed.</li>
<li><strong>Functionality:</strong> The KV Cache was successfully reused between Test 0 and Test 1.</li>
</ul>

Test Configuration | Network Get Time (s) | Disk Retrieve Time (s) | Request Time (s)
-- | -- | -- | --
Test 1, Request 1 | 0.0044 | 0.0047 | 9.2938
Test 1, Request 2 | 0.0058 | 1.0544 | 2.7525
Test 2, Request 1 | 0.0053 | 4.3302 | 7.7243
Test 2, Request 2 | 0.0064 | 0.7659 | 2.4600
Baseline, Request 1 | - | 0.0001 | 9.7393
Baseline, Request 2 | - | 0.9931 | 2.7357

</ul></body></html>
</body>
</html>

Below is the raw material:
[Disk_Address_Manager_log.txt](https://github.com/user-attachments/files/18030720/Disk_Address_Manager_log.txt)


examples/disk_offline_inference/README.md
YaoJiayi, line 1: How does this readme show the benefit of local disk sharing?

examples/disk_offline_inference/offline_inference_outputs.jsonl
YaoJiayi, line 1: Please remove

lmcache/cache_engine.py
```diff
@@ -371,6 +376,9 @@ def retrieve(
             if chunk is None:
                 break
             retrieved_kv_chunks.append(chunk)
+
+        logger.debug(
+            f"Overall disk retrieve time: {time.perf_counter() - time0}")
```
YaoJiayi: This log is confusing. Other backends share this code path

lmcache/storage_backend/evictor/base_evictor.py
```diff
@@ -62,24 +64,7 @@ def get_size(self, kv_obj: Union[torch.Tensor, bytes, KVObj]) -> float:
...
-        elif isinstance(kv_obj, KVObj):
-            size_in_gb = kv_obj.size
-        elif isinstance(kv_obj, DiskCacheMetadata):
-            size_in_gb = kv_obj.size
-        else:
-            raise Exception(
-                f"Encountered unknown kv data type {type(kv_obj)}!")
-
-        return size_in_gb
+        return _get_size_in_gb(kv_obj)
```
YaoJiayi: What's the point of this?

lmcache/utils.py
```diff
@@ -78,3 +80,29 @@ def _lmcache_nvtx_annotate(func, domain="lmcache"):
...
+
+    elif isinstance(kv_obj, KVObj):
+        size_in_gb = kv_obj.size
+
+    elif isinstance(kv_obj, DiskCacheMetadata):
+        size_in_gb = kv_obj.size
+    else:
+        raise Exception(f"Encountered unknown kv data type {type(kv_obj)}!")
+
+    return size_in_gb
```
YaoJiayi: Seems redundant? Isn't this the same function as in evictor?

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,278 @@
...
+    Cache engine for storing the KV cache of the tokens in the local disk.
+    """
+
+    def __init__(self, config: LMCAddressManagerConfig):
+        """
+        Throws:
+            RuntimeError if the loaded configuration does not match the current
+                configuration
+        """
+        super().__init__()
```
YaoJiayi: What is this super?

examples/disk_offline_inference/offline_inference_outputs.jsonl
YaoJiayi, line 1: Please remove

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        Load the config from a yaml file
+        """
+        with open(file_path, "r") as fin:
+            config = yaml.safe_load(fin)
+
+        local_device = config.get("local_device",
+                                  "disk_url://http://localhost:4322")
+        disk_url = local_device[11:]
+        disk_path = config.get("disk_path", "/local/local_disk/")
```
YaoJiayi: Maybe use relative path?

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,280 @@
...
+        LMCDiskAddressManager(config))  # Register your class instance
+
+    print("Server is running on port {}...".format(port))
+    server.serve_forever()
+
+
+def main():
+    if "LMCACHE_CONFIG_FILE" not in os.environ:
+        config = LMCAddressManagerConfig("http://localhost:4322",
+                                         "/local/local_disk/")
```
YaoJiayi: Use relative path

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,280 @@
...
+        return True
+
+    # def Info(self):
+    #     return self.info
+
+    def close(self):
+        pass
+
+    def __del__(self):
+        self.close()
```
YaoJiayi: The last three functions seem redundant

lmcache/storage_backend/local_backend.py
```diff
@@ -563,3 +564,211 @@ def close(self):
...
+        self,
+        key: CacheEngineKey,
+    ) -> bool:
+        anws: str = self.proxy.contains(key.to_string)  # type: ignore
+        self.cached_paths[key] = anws
+        return anws != ""
+
+    def batched_contains(self,
+                         keys: Iterable[CacheEngineKey],
+                         total_size: float = 0) -> Iterable[bool]:
```
YaoJiayi: List[bool]

lmcache/storage_backend/local_backend.py
```diff
@@ -563,3 +564,211 @@ def close(self):
...
+            # Write check
+            paths: Iterable[str] = self.proxy.batched_write_check(
+                [key.to_string() for key in keys], total_size)  # type: ignore
+
+            an = []
+            for path, key in zip(paths, keys):
+                if path == "":
+                    an.append(True)
+                else:
+                    self.cached_paths[key] = path
```
YaoJiayi, line 643: Just curious. Why do we need a `cached_paths` here?

lmcache/storage_backend/local_backend.py
```diff
@@ -563,3 +564,211 @@ def close(self):
...
+                    an.append(False)
+            return an
+
+    def save_end_callback(self, future: Future):
```
YaoJiayi, line 648: I'm a bit confused about the use of lock in this function. Why is `batched_write_ready` the only operation locked in this callback function?

lmcache/storage_backend/local_backend.py
```diff
@@ -563,3 +564,211 @@ def close(self):
...
+        if len(self.written_key
+               ) >= self.written_buffer_size or self.remain_writing == 0:
+            self.update_lock.acquire()
+            logger.debug(
+                f"Betched Read Signal for {len(self.written_key)} keys:")
+            # print("Write Ready",self.written_key,self.written_size)
+            signal = self.proxy.batched_write_ready(self.written_key,
+                                                    self.written_size)
+            if not signal:
+                logger.error("Write Ready Signal Failed")
```
YaoJiayi: When does it fail?

lmcache/storage_backend/local_backend.py
```diff
@@ -563,3 +564,211 @@ def close(self):
...
+        Output:
+            the kv cache of the token chunk, in the format of nested tuples
+            None if the key is not found
+        """
+        path = self.proxy.read_check(key.to_string())  # type: ignore
+
+        if path == "":
+            return None
+
+        self.update_lock.acquire()
```
YaoJiayi: Why lock here in `get` but no lock in the following `batched_get` function?

YaoJiayi (commented):
Thanks @qyy2003 ! I left some comments. About the benchmarking result, sorry I'm a little confused. Where is Test 0?
Also, shouldn't the result be something like (using fake numbers here) the following?
Test 1:
(1) first req: 10s
(2) second req: 1s
Test 2:
(1) first req: 1s

qyy2003:
Good question.That's what I've expected at first. 
I guess there are two reasons: disk buffer and LLM cold start. 
- When you write/read it last time, it will be faster to retrieve this time.
- The first request of LLM is slow. (I guess this is the main reason since the non-blocking first request in test 0 takes 9s)

If you see the log, you will find that LMCache fetched all the KV Cache correctly. So the code itself is correct.

examples/disk_offline_inference/README.md
qyy2003: Maybe I should change the test code.

But I encountered a strange problem:
When I tried to add a warm-up request before the original request, its KV Cache can't be saved to disk. I'm not sure how to fix it.

lmcache/cache_engine.py
qyy2003: Good point~

lmcache/storage_backend/evictor/base_evictor.py
qyy2003: OK, there is the thing.
I moved the evictor from backend to address manager. Evictor requires tensor size but I don't want to transmit tensor between them. So the file size needs to be calculated outside Evictor.

lmcache/utils.py
qyy2003: So I moved the one in evictor

lmcache/address_manager/disk_address_manager.py
qyy2003: good point~

lmcache/address_manager/disk_address_manager.py
qyy2003: Info is useful for future usage. It broadcast the manager's config to all the backends so that every KV Cache is stored in an uniformed way.
For the other two function, I could remove them. (I'm not sure about it, please give some advises.)

lmcache/storage_backend/local_backend.py
qyy2003: Since you don't want to query the Address Manager Twice. Otherwise you could first contain then ask_path.

lmcache/storage_backend/local_backend.py
qyy2003: We could talk about lock tomorrow. I'm not sure where I should use them.

lmcache/storage_backend/local_backend.py
qyy2003: Only because by default the rpcxml can't return None

qyy2003:
By the way, @YaoJiayi I've replied all your comments

lmcache/storage_backend/local_backend.py
YaoJiayi: How do you handle lock in this case?

YaoJiayi:
> Good question.That's what I've expected at first. I guess there are two reasons: disk buffer and LLM cold start.
> 
>     * When you write/read it last time, it will be faster to retrieve this time.
> 
>     * The first request of LLM is slow. (I guess this is the main reason since the non-blocking first request in test 0 takes 9s)
> 
> 
> If you see the log, you will find that LMCache fetched all the KV Cache correctly. So the code itself is correct.

@qyy2003  Could you add a random warmup request? 
Also, could you elaborate on disk buffer? I haven't observed this issue in my previous benchmarking.

@Shaoting-Feng Feel free to weigh in here. What's the expected performance for context len~ 25K? Also, please add an e2e test case for this:)

Shaoting-Feng:
This is the result of TTFT local disk storage with 24K context length (without the disk manager).
描述

Shaoting-Feng:
> @Shaoting-Feng Feel free to weigh in here. What's the expected performance for context len~ 25K? Also, please add an e2e test case for this:)

The new test case:

1. launch the disk manager with `LMCACHE_CONFIG_FILE=example.yaml lmcache_disk_address_manager`.
2. In the yaml file, apart from the original configs, specify `local_device: "disk_url://http://localhost:xxxx"`.

Is my understanding correctly?

ApostaC:
Thanks @qyy2003 !! I have a question about the format of the configuration:

In the configuration:
```
local_device: "disk_url://http://localhost:4322"
```

This is not a standard format of a URI. Usually, people only have a single `://` with a schema identifier. 

Also, what's the meaning of this `local_device` configuration? Why don't we just use `http://`?

examples/disk_offline_inference/example.yaml
```diff
@@ -0,0 +1,7 @@
+chunk_size: 256
+local_device: "file:///local/qinyuyang2003/local_disk/"
+# local_device: "disk_url://localhost:4322"
+# disk_path: "/local/qinyuyang2003/local_disk/"
+# disk_path: "/dataheart/qinyuyang2003/local_disk/"
```
ApostaC: Can we remove those lines?

examples/disk_offline_inference/example.yaml
```diff
@@ -0,0 +1,7 @@
+chunk_size: 256
+local_device: "file:///local/qinyuyang2003/local_disk/"
```
ApostaC: Can we change this to "file://local_folder", which means writing to the folder named `local_folder` under the current directory? (Usually, there won't be a folder named `/local/qingyuyang2003/local_disk` on user's machine)

And add the instruction of `mkdir -p local_folder` in the README.md to let users create a folder at the beginning.

examples/disk_offline_inference/my_example.yaml
ApostaC, line 1: Maybe we want to exclude this file from this PR? I guess this is only for your local test?

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,285 @@
...
+from lmcache.logging import init_logger
+from lmcache.storage_backend.evictor import LRUEvictor
+from lmcache.storage_backend.evictor.base_evictor import PutStatus
+from lmcache.utils import CacheEngineKey, DiskCacheMetadata
+
+logger = init_logger(__name__)
+
+
+@dataclass
+class LMCAddressManagerConfig:
```
ApostaC: Instead of having a new configuration class defined here, can we reuse the `LMCacheEngineConfig` (i.e., pass `LMCacheEngineConfig` to `LMCDiskAddressManager.__init__()`?

This is because if someone wants to add other ways to initialize the configuration (for instance, initialize from environment variables), he/she is likely to only look at the LMCacheEngineConfig and may not know this class needs to be updated as well.

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,285 @@
...
+        The logic here is that:
+            return "" if the key is already in the cache, 
+                so that backend will not write it again
+            return path if the key is not in the cache, 
+                so that backend will write it to the path
+        if path is not in the cache, initialize the cache with ("",0)
+        """
+        key = CacheEngineKey.from_string(key_str)
+        with self.update_lock:
+            if key in self.dict:
```
ApostaC, line 129: IIUC, the code here has a bug:
process A calling `write_check` with `key_str` --> `self.dict[key]` is assigned with `DiskCacheMetadata("", 0)`
process B calling `write_check` or `read_check` with  the same `key_str` --> return the path.

However, it's possible that process A hasn't written any thing into the disk yet, and race condition will happen (process B will try read / write a non-exist file)

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,285 @@
...
+    def batched_write_check(self, key_strs: Iterable[str],
+                            kv_size: float) -> Iterable[str]:
+        paths = []
+        kv_size = int(kv_size * (1024 * 1024))
+
+        self.update_lock.acquire()
+        for key_str in key_strs:
+            key = CacheEngineKey.from_string(key_str)
+            if key in self.dict:
+                kv_size = kv_size - self.dict[key].size
```
ApostaC, line 158: Why doing `kv_size = kv_size - self.dict[key].size` in `batched_write_check` but not in `write_check`?

lmcache/utils.py
```diff
@@ -78,3 +80,25 @@ def _lmcache_nvtx_annotate(func, domain="lmcache"):
         color=_get_color_for_nvtx(func.__qualname__),
         domain=domain,
     )(func)
+
+
+def _get_size(
```
ApostaC: Can we change the name of this function and add some detailed docstring for it?

Usually, the function starting with `_` is a private function that cannot be seen and called by other modules. However, this function is designed as a utility function that will be called by multiple different modules.

lmcache/cache_engine.py
```diff
@@ -194,14 +195,17 @@ def _make_chunks_skip_existing(
...
-                                   range(0, num_tokens, self.chunk_size)):
-            if not self.engine_.contains(self._make_key(chunk_hash, fmt)):
-                start_token_idx = idx
+        keys = []
+        for chunk_hash in chunk_hashes:
+            keys.append(self._make_key(chunk_hash, fmt))
+        anws = self.engine_.batched_contains(keys,
+                                             total_size=_get_size(kv_tensors))
+        for anw in anws:
+            if not anw:
```
ApostaC: The loop is easy to understand. But what does `anw` and `anws` mean? 

Note that the code might be changed by other people in the future, it's better to have a clear variable name, thanks!

lmcache/cache_engine.py
```diff
@@ -439,8 +446,12 @@ def lookup(
         total_token_cnt = len(tokens)
         current_token_idx = 0
         chunk_hashes = self._prefix_hash(self._chunk_tokens(tokens), 0)
+        keys = []
         for chunk_hash in chunk_hashes:
-            if not self.engine_.contains(self._make_key(chunk_hash, fmt)):
+            keys.append(self._make_key(chunk_hash, fmt))
+        anws = self.engine_.batched_contains(keys, total_size=0)
+        for anw in anws:
+            if not anw:
```
ApostaC: Same here. Better to have a better name for `anw` and `anws`

lmcache/config.py
```diff
@@ -53,6 +54,7 @@ def from_defaults(
         chunk_size: int = 256,
         local_device: str = "cuda",
         max_local_cache_size: int = 5,
+        disk_url: str = "",
```
ApostaC: A few quick comments about `disk_url` here.
- The name of `disk_url` itself is a bit ambiguous. Is it the URL of the disk manager process? Then maybe we should use something like `disk_manager_url`
- If it's an optional configuration, it's clearer to use `None` as the default value, instead of `""`
- In the `from_file` code, `disk_url` is first set by `disk_url = config.get("disk_url", None)`, and then be changed again when `if re.match(r"disk_url://(.*)", local_device)`. This is pretty confusing. Why changing the value of `local_device` can overwrite the `disk_url`? Does this mean `disk_url` in the yaml config is useless?

lmcache/storage_backend/local_backend.py
```diff
@@ -562,3 +563,214 @@ def close(self):
...
+
+@_lmcache_nvtx_annotate
+@torch.inference_mode()
+def save_disk_with_callback(
+    path: str,
+    kv_obj: KVObj,
+    key: CacheEngineKey,
+) -> Tuple[CacheEngineKey, float, int]:
+    save_file({"kv_chunk": kv_obj.data.contiguous()}, path)
+    return key, _get_size(kv_obj.data), kv_obj.chunk_idx
```
ApostaC: Why this function is called `save_disk_with_callback`, where's the callback?

lmcache/storage_backend/abstract_backend.py
```diff
@@ -124,6 +124,22 @@ def batched_get(
...
+                         total_size: float) -> Iterable[bool]:
+        """
+        Query if the keys are in the cache or not in a batched manner
+
+        :param keys: the iterator of keys of the token chunks, including prefix 
+                hash and format
+
+        :return: the iterator of boolean values, indicating if the key is in 
+                the cache or not
+        """
```
ApostaC: What does `total_size` mean here? It's not in the docstring. Also, why querying the existence of the keys requires `total_size`?

ApostaC (changes requested):
Thanks @qyy2003 ! This is awesome! Before the detailed comments, here are some general comments:
- Have more docstrings to explain the semantics of the key functions (like `write_check` and `read_check`. Currently, it explains the logic but not semantics.
- Write some unittests to make sure there is no logical bug under corner cases.
- Be more clear about the user interface and avoid using ambiguous names for the variable and configuration items.

I haven't taken a detailed look at the new codes in `local_backend.py`. Will take a look again once you have a clearer explaination of the core interfaces it is calling.

examples/disk_offline_inference/example.yaml
qyy2003: we will create the folder for them

examples/disk_offline_inference/my_example.yaml
qyy2003: OK~

lmcache/address_manager/disk_address_manager.py
qyy2003: Because write_check will return if it's in the dict

lmcache/address_manager/disk_address_manager.py
qyy2003: I didn't get it.

lmcache/config.py
qyy2003: Actually, the `disk_url`  in yaml config means `disk folder path` in disk_address_manager and `address manager url` in LMCache backend.

lmcache/storage_backend/local_backend.py
qyy2003:
```
future = self.proc_pool_executor.submit(save_disk_with_callback, path,
                                                kv_obj, key)
        future.add_done_callback(self.save_end_callback)
```
in `put_nonblocking` function

lmcache/storage_backend/abstract_backend.py
qyy2003: Because here the `contains` function combines existence check and path retrieve. If you call `contains` with a total size other than 0, it means you will `put` these KV Chunks later: LMCache backend will cache the path for `put` function and the address manager will incur the evictor.

qyy2003:
**New Test Results**

---

**Conclusion:**  
The unexpected slowdown of the first request is caused by the Disk Cache.


**Test Setup:**  
1. Start the Address Manager.  
2. Perform the following steps for each test case:  
   - **TEST1 (GPU0):**  
     a. Initialize vLLM on GPU0.  
     b. Prefill one request.  
     c. Wait 30 seconds for the KV Chunked data to be stored.  
     d. Prefill the same request again.  
     e. Close the vLLM.  
   - **TEST2 (GPU1):**  
     a. Initialize vLLM on GPU1.  
     b. Prefill one request.  
     c. Wait 30 seconds for the KV Chunked data to be stored.  
     d. Prefill the same request again.  
     e. Close the vLLM.  
   - **TEST3 (GPU1):**  
     a. Repeat the same steps as in TEST2.  
3. Before prefilling requests in TEST1 and TEST3, manually preload the Disk Folder using `cat`.  
4. Close the Address Manager.


**Experiment Results:**  

<html><body>

Test Case | GPU | Request 1 Time (s) | Request 2 Time (s)
-- | -- | -- | --
TEST1 | GPU0 | 6.171 | 1.145
TEST2 | GPU1 | 2.543 | 1.121
TEST3 | GPU1 | 1.662 | 1.205

</body></html>
</body>
</html>

**Observations:**  
The retrieval time on the backend side for all tests is approximately 6ms, which is negligible.

qyy2003:
@ApostaC @YaoJiayi  I've added doc string, unitest and test examples, please review.

examples/disk_offline_inference/preload_disk.sh
```diff
@@ -0,0 +1,4 @@
+DISK_PATH=local_disk/
+for file in $DISK_PATH/*; do
+  cat "$file" > /dev/null
+done
```
YaoJiayi, line 4: Delete

lmcache/address_manager/disk_address_manager.py
YaoJiayi: Race condition might happen

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,286 @@
...
+            when other backend try to read the same KV Cache chunks, 
+                it will be blocked by the "" value
+        """
+        key = CacheEngineKey.from_string(key_str)
+        with self.update_lock:
+            if key in self.dict:
+                return ""
+
+            evict_keys, put_status = self.evictor.update_on_put(
+                self.dict, int(kv_size * (1024 * 1024)))
```
YaoJiayi, line 133: What is 1024 * 1024?

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,286 @@
...
+
+        with self.update_lock:
+            self.dict[key] = DiskCacheMetadata("", 0)
+
+        return self._key_to_path(key)
+
+    def batched_write_check(self, key_strs: Iterable[str],
+                            kv_size: float) -> Iterable[str]:
+        paths = []
+        kv_size = int(kv_size * (1024 * 1024))
```
YaoJiayi, line 151: Same as above.

lmcache/address_manager/disk_address_manager.py
YaoJiayi: Could you elaborate?

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,286 @@
...
+        logger.error(f"Cannot parse disk url {config.disk_url} in the config")
+        raise ValueError(f"Invalid disk url {config.disk_url}")
+
+    connector_type, host, port = m.group(1), m.group(2), int(m.group(3))
+    print(connector_type, host, port)
+    server = SimpleXMLRPCServer((host, port), requestHandler=RequestHandler)
+    server.register_introspection_functions(
+    )  # Optional: provides a list of registered functions
+    server.register_instance(
+        LMCDiskAddressManager(config))  # Register your class instance
```
YaoJiayi, line 275: Is this thing multi threaded or mutil processed or just single threaded/processes?

lmcache/storage_backend/__init__.py
```diff
@@ -40,11 +41,19 @@ def CreateStorageBackend(config: LMCacheEngineConfig,
                         f" backend")
 
                     return LMCLocalBackend(config, mpool_metadata, dst_device)
-                case _:
+                case "global_disk":
```
YaoJiayi, line 44: What does this "global disk" mean? Is it remote shared disk?

lmcache/storage_backend/local_backend.py
YaoJiayi: Is this resolved?

lmcache/storage_backend/local_backend.py
YaoJiayi: I'm a bit confused about why you are using lock in the disk manager instead of in the backends here.

lmcache/storage_backend/local_backend.py
```diff
@@ -562,3 +574,220 @@ def close(self):
...
+        paths: Iterable[str] = self.proxy.batched_read_check(
+            keys_strs)  # type: ignore
+        logger.debug(
+            f"Network Time taken for batched_get {time.perf_counter() - time0}s"
+        )
+        for path in paths:
+            if path == "":
+                yield None
+            else:
+                with safe_open(path, framework="pt",
```
YaoJiayi, line 777: How do you handle read-write contention here? I understand you queried the disk manager before. But that doesn't gurantee anything here.

examples/disk_offline_inference/preload_disk.sh
qyy2003: It is used to test whether the disk cache is the primary cause of the first request slowdown.

lmcache/address_manager/disk_address_manager.py
qyy2003: transform into GB.  The number in int is too large for xmlrpc to transmit.

lmcache/address_manager/disk_address_manager.py
qyy2003: Nop. This will not happens. process B calling write_check or read_check with the same key_str will get the path "", which also means not exist

lmcache/address_manager/disk_address_manager.py
qyy2003:
```
kv_size: the size of the KV Cache chunk group in GB, 
            free up the space if the existing cache is too big
            
# Do not free up space for existed KV Cache chunks
```
Just see the comment

lmcache/address_manager/disk_address_manager.py
qyy2003: The SimpleXMLRPCServer in the code is single-threaded by default. This means it processes one request at a time and does not handle multiple client requests concurrently.

lmcache/storage_backend/__init__.py
qyy2003: No, it simply means a disk that can be shared among different LM Engines. If you can use a virtual disk like `/dataheart`, it works the same way. The disk doesn’t need to be physically present; as long as you can interact with it like a normal disk, it can be utilized.

Additionally, it complements the LocalDisk in both name and functionality.

For a remote server, you'll serialize it using our own protocol. So, I believe it’s independent of the disk backend.

lmcache/storage_backend/local_backend.py
qyy2003: You will always first call contains than call put right?

lmcache/storage_backend/local_backend.py
qyy2003: There is no lock here. You just inform the address manager that you have finished. You don't even update dict.

lmcache/storage_backend/local_backend.py
qyy2003: This conflict will be handled by address manager.

lmcache/storage_backend/local_backend.py
qyy2003: It's guaranteed by the address manager

qyy2003 (commented):
Reply to @YaoJiayi's question.

lmcache/address_manager/disk_address_manager.py
qyy2003: In MB actually

lmcache/address_manager/disk_address_manager.py
qyy2003: There are two options

## 1. Use single-thread mode and Delete lock 
If multiple clients send requests concurrently to the SimpleXMLRPCServer in its default configuration (single-threaded), the following behavior occurs:
Sequential Processing

    The server will process one request at a time in the order it receives them.
    While handling a request, any other client requests will be blocked (queued) until the current request is completed.
    As a result, clients might experience delays, especially if the handling of one request takes a long time.
    
## 2. Change to multi-thread mode with 4 lines of code.

from
```
server = SimpleXMLRPCServer((host, port), requestHandler=RequestHandler)
```

to

```
from socketserver import ThreadingMixIn
class ThreadedXMLRPCServer(ThreadingMixIn, SimpleXMLRPCServer):
    pass
server = ThreadedXMLRPCServer((host, port), requestHandler=RequestHandler)
```

lmcache/storage_backend/local_backend.py
qyy2003: Fixed

lmcache/address_manager/disk_address_manager.py
qyy2003: For write_check, if key in self.dict, the function will return "" and not incur evictor.

qyy2003:
In summary, all the bugs have been fixed, and the issues have been resolved.

---

The only potential concern is that `SimpleXMLRPCServer` is single-threaded by default. However, I have added a lock to ensure thread safety, which does not impact functionality. 

While this can be improved, I believe it’s sufficient for this PR. A solution can be addressed in a future PR if needed.There are two possible approaches to address the single-threaded limitation:

### 1. **Keep Single-Thread Mode and Remove the Lock**
If multiple clients send requests concurrently to the `SimpleXMLRPCServer` in its default single-threaded mode, the following behavior will occur:
- **Sequential Processing**: The server processes one request at a time in the order it receives them.
- **Blocking Requests**: While handling one request, any other client requests will be queued until the current request is completed.
- **Impact on Performance**: Clients might experience delays, especially if a request takes significant time to process.

### 2. **Switch to Multi-Threaded Mode (Minimal Changes)**
By modifying just 4 lines of code, we can enable multi-threaded support:

Change the following:

```python
server = SimpleXMLRPCServer((host, port), requestHandler=RequestHandler)
```

To:

```python
from socketserver import ThreadingMixIn

class ThreadedXMLRPCServer(ThreadingMixIn, SimpleXMLRPCServer):
    pass

server = ThreadedXMLRPCServer((host, port), requestHandler=RequestHandler)
```

This approach allows the server to handle multiple client requests concurrently, significantly improving performance in scenarios with concurrent clients.

---

As it may requests lots of testing. I thought it's better to fix it with other PR

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        # evict caches
+        for evict_key in evict_keys:
+            self.remove(evict_key)
+
+        with self.update_lock:
+            self.dict[key] = DiskCacheMetadata("", 0)
+
+        return self._key_to_path(key)
+
+    def batched_write_check(self, key_strs: Iterable[str],
```
qyy2003, line 148: total KV size

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,292 @@
...
+
+        return paths
+
+    def write_ready(self, key_str: str, kv_size: float):
+        """
+        When backend has finished writing the cache, 
+            it will call this function to update the cache 
+            from ("",0) to the actual path and size
+        
+        kv_size: the size of the KV Cache chunk in GB,
```
qyy2003, line 185: GB

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,292 @@
...
+            # print("Illegal")
+            return [""] * len(key_strs)  # type: ignore
+
+        # evict caches
+        for evict_key in evict_keys:
+            self.remove(evict_key)
+
+        return paths
+
+    def write_ready(self, key_str: str, kv_size: float):
```
qyy2003, line 179: on_write_finish

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,292 @@
...
+
+        assert self.path is not None, ("Need to specify local path if when "
+                                       "using  Local Disk")
+
+        if not os.path.exists(self.path):
+            os.makedirs(self.path)
+
+        self.update_lock = threading.Lock()
+
+        self.evictor = LRUEvictor()
```
qyy2003, line 46: param

lmcache/address_manager/disk_address_manager.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class LMCDiskAddressManager():
+    """
+    The address managers centralize the key-path dictionary and 
+        the evictor of each disk engine into an independent process, 
+        enabling multiple LMCache Disk Backends to share KV Cache files.
+    """
+
+    def __init__(self, config: LMCAddressManagerConfig):
```
qyy2003, line 25: global config

toilaluan:
@qyy2003 
It looks great
I assume with this future, i can manually create kv cache tensors: (n_layers, 2, batch_size, seq_len, n_head, head_dim) then save it to disk with random hash then opt in this hash to llm engine when run generate?

My purpose is to use custom kv-compress method for long prefix prompt.

hickeyma:
@qyy2003 Do you mind rebasing?

@ApostaC @YaoJiayi Is this ready to review again once rebased?

Shaoting-Feng:
> @qyy2003 Do you mind rebasing?
> 
> @ApostaC @YaoJiayi Is this ready to review again once rebased?

This PR is deprecated and we can ignore this.

hickeyma:
> This PR is deprecated and we can ignore this.

Closing as per @Shaoting-Feng comment in https://github.com/LMCache/LMCache/pull/256#issuecomment-3045753229.

@qyy2003 Let us know if you want this re-opened.

PR #312, opened 2025-01-16
[Misc] RAG benchmarks

XbzOnGit:
Add RAG benchmarks, including CacheBlend.
Update some documentations on benchmarking and blend examples.


lmcache/blend/retriever.py
```diff
@@ -245,7 +245,7 @@ def new_request(
                     tokens,  # tokens
                     None,  # mask
                     False,  # return_tuple
-                ) for tokens in splitted_tokens if len(tokens) > 0
```
ApostaC, line 248: Is it because there is a for-loop over `tasks` and the number of tasks should match something?

benchmarks/README.md
```diff
@@ -118,3 +116,117 @@ The `multi-round-qa.py` script works by:
...
+1. Install the required dependencies:
+   ```bash
+   pip install -r requirements.txt
+   ```
+## Running the RAG Benchmark
+
+To run the RAG benchmark, use the following command:
+
+```bash
+python3 rag.py --qps 3.5 --model mistralai/Mistral-7B-Instruct-v0.2 --dataset ~/CacheBlend/inputs/musique_s.json --system-prompt "You will be asked a question after reading several passages. Please directly answer the question based on the given passages. Do NOT repeat the question. The answer should be within 5 words..\nPassages:\n" --query-prompt "\n\nAnswer the question directly based on the given passages. Do NOT repeat the question. The answer should be within 5 words. \nQuestion:" --separator "" --prompt-build-method QA --base-url "http://localhost:8000/v1" --kv-storage-size 30GB --kv-storage-token-unit 256 --max-tokens 32 --output summary.csv
```
ApostaC: Can we simplify the example command by the following:
- Remove the system prompt and query prompt text (maybe have a default prompt in the script when users don't provide them)
- Skip the command line args if there is a default value 
- Split the command into multiple lines with `\`

benchmarks/README.md
```diff
@@ -118,3 +116,117 @@ The `multi-round-qa.py` script works by:
...
+*Note:* the above command requires there is a serving engine with the `mistralai/Mistral-7B-Instruct-v0.2` model served locally at `http://localhost:8000/v1`. Here's an example command to launch the serving engine:
+
+```bash
+vllm serve mistralai/Mistral-7B-Instruct-v0.2 --disable-log-requests
+```
+
+## Running the RAG benchmark on LMCache
+For benchmarking LMCache with CacheBlend:  
+```bash
+LMCACHE_CONFIG_FILE=example_blending.yaml python3 rag.py --qps 3.5 --model mistralai/Mistral-7B-Instruct-v0.2 --dataset ~/CacheBlend/inputs/musique_s.json --system-prompt "You will be asked a question after reading several passages. Please directly answer the question based on the given passages. Do NOT repeat the question. The answer should be within 5 words..\nPassages:\n" --query-prompt "\n\nAnswer the question directly based on the given passages. Do NOT repeat the question. The answer should be within 5 words. \nQuestion:" --separator "[BLEND_SEP]" --prompt-build-method QA --base-url "http://localhost:8000/v1" --kv-storage-size 30GB --kv-storage-token-unit 256 --max-tokens 32 --output summary.csv
```
ApostaC: Simplify this command as well

benchmarks/README.md
```diff
@@ -118,3 +116,117 @@ The `multi-round-qa.py` script works by:
...
+
+
+
+### What does precompute.py do
+--skip-precompute will skip the precomputation, while by default to will call precompute functions on documents.  
+If no --end-index provided, it will check kv-storage-size and try to precompute the documents that can be held in this size.  
+Later RAG benchmark will only run these requests.  
+
+### Arguments
+#### Configure the workload
```
ApostaC: It's great to have such flexible configurations, but as a user, I found it a bit difficult to understand all the arguments. Here are few items to potentially improve it:
- Try to hide the LMCache-related details. IIUC, there are quite a few args related to LMCache, including `-separator`, `--model`, `--skip-precompute`, `--kv-storage-size`, `--kv-storage-token-bit`, `--kv-precision-bit`. 
  - One potential solution is to use another command to run all the LMCache precompute related stuff. And rag.py only provides a single argument called `--enable-lmcache` (so the script knows it needs to add separators). 
- Try to avoid ambiguous naming. For example, at first glance, I thought `--model` is for specifying the model's name, but it seems like the correct configuration is `--model-api-name`.
- It's okay to hard-code some configurations that won't have too much impact on end-to-end results. E.g., `--step-interval`, `--separator`.
- Better clarification of the arg's meaning. I didn't quite get the meaning of `--doc-token-cnt`

I'm happy to have a chat to discuss the details of those configurations. Just let me know if you want.

ApostaC (dismissed):
Thanks @XbzOnGit for the great work!

The benchmarking scripts look good to me, but the user interface can be improved (see the comments below).

Also, it would be great if we can put the current multi-round QA benchmark and the RAG benchmark into 2 sub-folders. In this case, user's who want to run some rag benchmarking won't be "noised" by multi-round QA benchmark.

lmcache/blend/retriever.py
XbzOnGit: The number of tasks should match the number of token_segments.
It does not matter if every segment is not empty.

benchmarks/README.md
XbzOnGit: We can decide those prompts by prompt-build-method if not provided. That makes sense.

benchmarks/README.md
XbzOnGit: We can remove --doc-token-cnt. It is used for getting a sense that how much of text is documents(something like reuse ratio).
--separator cannot be hard-coded because there's a difference between blending and non-blending/vllm ones.
--step-interval is just in case someone wants a higher QPS. We can keep it default most of the time.
The precompute part in README is related with LMCache. Previously I used this design just for convenience. I can run one command to launch one benchmark. It would be logically better to divide them into two commands.

XbzOnGit:
I will create some new pushes addressing the command line arguments soon.

YaoJiayi (approved):
LGTM in general. Just one minor comment for the README: could you explicitly state how to run the benchmark? It's still a bit confusing to me. Maybe start with instructions to run the benchmark. Then explain the arguments later.

PR #325, opened 2025-01-21
[Bugfix] Update local_backend.py, use thread pool instead of process pool in localdisk backend

xenshinu:
FIX #xxxx
```
future = self.proc_pool_executor.submit(save_disk, path, kv_obj.data)
```
Multi-process won't share memory space, if you simply open a process to save the chunk to file, it will pass an unaccessible pointer and stuck. A simple solution is to use threadpool instead. If you want a higher speed, consider build-in multi-thread functions of torch.

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

### DIFF
--- a/lmcache/storage_backend/local_backend.py
+++ b/lmcache/storage_backend/local_backend.py
@@ -3,7 +3,7 @@ import queue
 import threading
 import time
 from collections import OrderedDict
-from concurrent.futures import Future, ProcessPoolExecutor
+from concurrent.futures import Future, ThreadPoolExecutor
 from typing import Dict, Optional, Tuple, Union
 
 import torch
@@ -306,7 +306,7 @@ class LMCLocalDiskBackend(LMCBackendInterface):
         # TODO(Jiayi): share the buffer if both cpu and disk backend are enabled
         self.cpu_mbufferpool = LocalCPUBufferPool(metadata)
 
-        self.proc_pool_executor = ProcessPoolExecutor(max_workers=4)
+        self.proc_pool_executor = ThreadPoolExecutor(max_workers=4)
 
     def contains(
         self,


YaoJiayi:
> FIX #xxxx
> 
> ```
> future = self.proc_pool_executor.submit(save_disk, path, kv_obj.data)
> ```
> 
> Multi-process won't share memory space, if you simply open a process to save the chunk to file, it will pass an unaccessible pointer and stuck. A simple solution is to use threadpool instead. If you want a higher speed, consider build-in multi-thread functions of torch.
> 
> **BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**
> PR Checklist (Click to Expand)

Hi, thanks for the suggestion! Could you share some numbers (process pool vs. thread pool) if possible?
Btw, asyncio is used in experimental version (https://github.com/LMCache/LMCache/blob/dev/lmcache/experimental/storage_backend/local_disk_backend.py#L64), and will become the default path soon.

yanok:
> Hi, thanks for the suggestion! Could you share some numbers (process pool vs. thread pool) if possible?

Please note that this is **NOT** an improvement, it's a **FIX**. Without this change writing to disk hangs indefinitely. It's super funny that it _sometimes_ work. I'd guess that's because of COW for forked process memory: probably if there were no writes, memory space is still shared, so forked process can read from it and write succeeds.

Please accept this trivial change to fix the hangs.

yanok:
@YaoJiayi ping. The code at HEAD is broken, local disk backend is pretty much unusable, could you please accept this.

I guess the description could be a bit clearer: I don't think it hangs because of "unaccessible pointer". What happens is:
1. ProcessPool `fork`s the process
2. `fork` clones the address space of the current process using COW. The two are initially shared, but if one of the processes write to memory, the corresponding page will be copied and processes will have _different_ memory contents.
3. I don't know for sure, but most likely hangs happen if `fork` clones the address space while some mutex is held. This way, when the mutex is freed, **only one** process will see it, the other one will have the mutex locked forever.
4. That also explains some non-determinism of hangs: they happen on some examples, but not others.
5. But even if there are no hangs, ProcessPool won't work correctly: processes in the pool will _never_ see the changes to the main process memory done after `fork`, so even if there are no hangs, you are probably saving an outdated copy now.

So, please, please accept this PR.

xenshinu:
> @YaoJiayi ping. The code at HEAD is broken, local disk backend is pretty much unusable, could you please accept this.
> 
> I guess the description could be a bit clearer: I don't think it hangs because of "unaccessible pointer". What happens is:
> 
> 1. ProcessPool `fork`s the process
> 2. `fork` clones the address space of the current process using COW. The two are initially shared, but if one of the processes write to memory, the corresponding page will be copied and processes will have _different_ memory contents.
> 3. I don't know for sure, but most likely hangs happen if `fork` clones the address space while some mutex is held. This way, when the mutex is freed, **only one** process will see it, the other one will have the mutex locked forever.
> 4. That also explains some non-determinism of hangs: they happen on some examples, but not others.
> 5. But even if there are no hangs, ProcessPool won't work correctly: processes in the pool will _never_ see the changes to the main process memory done after `fork`, so even if there are no hangs, you are probably saving an outdated copy now.
> 
> So, please, please accept this PR.

@yanok Thanks for your explanation. But I suppose the `ProcessPool` will launch a few worker processes at the beginning of `init`, if it just forks the main process on demand, the overhead could be high. So if the processes are created ahead of time, they won't share any memory space. But tbh I have no idea what is the exact impl of `ProcessPool` in Python, so correct me if I am wrong.

YaoJiayi (approved):
Sorry for the late response. Thanks for the reminder! The PR is merged. But please use experimental version for the sake of performance:)

yanok:
> @yanok Thanks for your explanation. But I suppose the `ProcessPool` will launch a few worker processes at the beginning of `init`, if it just forks the main process on demand, the overhead could be high. So if the processes are created ahead of time, they won't share any memory space. But tbh I have no idea what is the exact impl of `ProcessPool` in Python, so correct me if I am wrong.

yes, I'm pretty sure it does `fork`s from `init`, but it only makes things worse: this way workers have even less chance to have up to date data in their address space, so they are not saving the final result (but the contents of corresponding vars at the time of `fork`). Anyway, with this PR merged, this should be fixed now. Thanks for your PR!

@YaoJiayi, thanks for accepting! I took a quick look at the experimental version, but it doesn't currently write metadata, which we need, so I continued playing with the non-experimental one. Also, I was not sure how experimental is experimental :)

PR #363, opened 2025-02-18
[Feature]Add new storage backend infinistore

thesues:
Dear list
   [infinstore](https://github.com/bd-iaas-us/infiniStore) is a storage backend which supports RDMA and in-place download and upload.  this PR is to add a bridge between  lmcache and infinistore. We have been using InfiniStore internally with vLLM for the past three months and have found it to be highly effective. you could check the benchmark results: 

https://github.com/bd-iaas-us/InfiniStore/issues/55

This is how it works:

Install infinistore

```
pip install infinistore
```

Start infinistore

```
python -m infinistore.server --service-port 12345 --dev-name mlx5_0 --link-type Ethernet  --manage-port 8080 --minimal-allocate-size 32
```

How to start LMCache

```
LMCACHE_CONFIG_FILE=backend_type.yaml LMCACHE_USE_EXPERIMENTAL=True python -m lmcache_vllm.vllm.entrypoints.openai.api_server --model mistralai/Mistral-7B-Instruct-v0.2  --max-model-len 8192
```

backend_type.yaml

```
chunk_size: 256
remote_url: "infinistore://127.0.0.1:12345"
```

We appreciate any feedback or suggestions. Thank you for your time and consideration

Best regards,


ApostaC:
Hey @thesues , can you run `format.sh` locally so that the format checker can be passed? Thanks!

Btw, really appreciate the contribution!

YaoJiayi (commented):
Quick comment: could you put inifinistore dependencies in `setup.py` and `requirements.txt`?
Another side question: can we run infinistore without rdma?

thesues:
> Quick comment: could you put inifinistore dependencies in `setup.py` and `requirements.txt`?

sure, updated.

> Another side question: can we run infinistore without rdma?


no, RDMA is mandatory.  But I think we could add TCP  transport as well in the future.

YaoJiayi (commented):
Hey, thanks for fixing the requirement! I was thinking of putting rdma into unit tests but our server, where buildkite is run on, does not support rdma. But let's skip this for now:)

The code in general looks good to me. 
However, correct me if I'm wrong, this code is not runnable even with rdma? Don't we also need a storage backend file? Maybe it's better to provide an example under `example/` people with rdma can run it?

thesues:
> The code in general looks good to me. However, correct me if I'm wrong, this code is not runnable even with rdma? Don't we also need a storage backend file? Maybe it's better to provide an example under `example/` people with rdma can run it?

From my test, the code is OK with RDMA driver installed.

infinistore is a memory storage for now, so no backfile needed. And what kind of example do you need ?

YaoJiayi:
> > The code in general looks good to me. However, correct me if I'm wrong, this code is not runnable even with rdma? Don't we also need a storage backend file? Maybe it's better to provide an example under `example/` people with rdma can run it?
> 
> From my test, the code is OK with RDMA driver installed.
> 
> infinistore is a memory storage for now, so no backfile needed. And what kind of example do you need ?

https://github.com/LMCache/LMCache/tree/dev/examples/experimental. Sth like this? You can create a new folder.

thesues:
examples/experimental/storage_backends/infinistore/README.md was added to help user to use infinistore

YaoJiayi:
> examples/experimental/storage_backends/infinistore/README.md was added to help user to use infinistore

Wonderful! One last comment, could you pass the format checker again? I'll merge it afterwards.

thesues:
> > examples/experimental/storage_backends/infinistore/README.md was added to help user to use infinistore
> 
> Wonderful! One last comment, could you pass the format checker again? I'll merge it afterwards.

lint fixed

YaoJiayi (approved):
LGTM!

lmcache/experimental/storage_backend/connector/infinistore_connector.py
```diff
@@ -0,0 +1,142 @@
...
+                   ) <= self.buffer_size, "metadata size exceeds buffer size"
+
+        # copy metadata to self.buffer
+        self.buffer[:len(metadata_bytes)] = metadata_bytes
+
+        await self.rdma_conn.rdma_write_cache_single_async(
+            key.to_string() + "metadata", _get_ptr(self.buffer),
+            len(self.buffer))
+
+        pointer = ctypes.cast(ctypes.c_char_p(memory_obj.byte_array),
```
maobaolong, line 124: this raise a `bytes or integer address expected instead of memoryview instance` exception, and swallowed. @thesues @YaoJiayi 

It means that infinistore can only support `BINARY_BUFFER` MemoryFormat?

PR #370, opened 2025-02-27
[Misc] add two simple utility functions

yanok:
1. `is_envvar_enabled` checks is the specified variable is set to one of "enable" values: `1`/`true`/`enable`/`enabled`
2. `timing` lets to collect time measurements easily, here is an example:
    ```python
    @dataclass
    class Stats:
      times = datasketches.kll_floats_sketch(100)
      total_time = 0
    stats = Stats()
    with timing(stats, 'times', 'total_time'):
      an_operation_to_time()
    ```

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -1,6 +1,8 @@
 import hashlib
 import os
 import threading
+import time
+from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


ApostaC:
@yanok Thanks for the PR! Can you fix the format checker by running "format.sh" locally? Thanks!

hickeyma (commented):
@yanok Do you mind rebasing and run the formatting?

yanok:
> @yanok Do you mind rebasing and run the formatting?

I think let's just close this PR. For checking env vars I think there is another helper function already, and I was not entirely happy with the timing implementation myself, so we ended up not using it ourselves.

hickeyma:
> I think let's just close this PR. For checking env vars I think there is another helper function already, and I was not entirely happy with the timing implementation myself, so we ended up not using it ourselves.

Ok, thanks @yanok

PR #373, opened 2025-02-28
[CI/Build] Bump vllm version to v0.7.3.

maobaolong:
As vllm release v0.7.3 recently, we need to bump the docker file also


YaoJiayi (approved):
Thanks for the update. Appreciate!

YaoJiayi:
@maobaolong Btw, the docker patch (at the end of the dockerfile) is no longer needed since our pr on vllm side is already merged.

maobaolong:
@YaoJiayi I see, I will watch the docker hub of `vllm`, I'll submit another PR to remove the extra content once `vllm` push new docker.

PR #382, opened 2025-03-04
[Enhancement] Copy the LMSLocalDiskBackend implementation to the experimental version

maobaolong:
The experimental version do not support store to local disk through LMServer for now, although the original implementation needs some future work, but it is still necessary to have this feature for now, after this, we can do some efforts to improve it.

### DIFF
--- a/lmcache/experimental/server/storage_backend/local_disk_backend.py
+++ b/lmcache/experimental/server/storage_backend/local_disk_backend.py
@@ -1,0 +1,171 @@
+import os
+import threading
+from collections import OrderedDict
+from typing import List, Optional
+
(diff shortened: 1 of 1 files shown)


YaoJiayi:
> The experimental version do not support store to local disk through LMServer for now, although the original implementation needs some future work, but it is still necessary to have this feature for now, after this, we can do some efforts to improve it.

Totally agreed the this should be added. But it seems the current code is not runnable?

maobaolong:
@YaoJiayi Thanks for your reply.  it need some adopting works to do, not only copy code to the experimental folder.

PR #394, opened 2025-03-10
Add use_existing_torch for users who don't want to reinstall torch

JiamingMai:
For supporting NVIDIA 50 series GPU which uses CUDA 12.8, we can't reinstall torch. This PR adds use_existing_torch.py for user who don't want to reinstall torch.

### DIFF
--- a/use_existing_torch.py
+++ b/use_existing_torch.py
@@ -1,0 +1,36 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import glob
+import re
+
+requires_files = glob.glob('requirements.txt')
+requires_files += ["pyproject.toml"]
+for file in requires_files:
+    print(f">>> cleaning {file}")
+    with open(file) as f:
+        lines = f.readlines()
+    if "torch" in "".join(lines).lower():
+        print("removed:")
+        with open(file, 'w') as f:
+            for line in lines:
+                if 'torch' not in line.lower():
+                    f.write(line)
+                else:
+                    print(line.strip())
+    print(f"<<< done cleaning {file}")
+    print()
+
+setup_file_path = "setup.py"
+with open(setup_file_path, "r") as file:
+    content = file.read()
+modified_content = re.sub(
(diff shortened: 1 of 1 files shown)


YaoJiayi (commented):
@JiamingMai Thanks for the PR! Quick question: how will this use_existing_torch.py called?

JiamingMai:
> @JiamingMai Thanks for the PR! Quick question: how will this use_existing_torch.py called?

@YaoJiayi When users want to install lmcache without reinstalling torch, run `python use_existing_torch.py` first. This will update `requirements.txt` and `setup.py`. Just like vllm https://github.com/vllm-project/vllm/blob/main/use_existing_torch.py https://docs.vllm.ai/en/latest/getting_started/installation/gpu.html

```shell
git clone https://github.com/LMCache/LMCache.git
cd LMCache
python use_existing_torch.py
pip install -r requirements.txt
pip install -e . --no-build-isolation
```

YaoJiayi:
> > @JiamingMai Thanks for the PR! Quick question: how will this use_existing_torch.py called?
> 
> @YaoJiayi When users want to install lmcache without reinstalling torch, run `python use_existing_torch.py` first. This will update `requirements.txt` and `setup.py`. Just like vllm https://github.com/vllm-project/vllm/blob/main/use_existing_torch.py https://docs.vllm.ai/en/latest/getting_started/installation/gpu.html
> 
> ```shell
> git clone https://github.com/LMCache/LMCache.git
> cd LMCache
> python use_existing_torch.py
> pip install -r requirements.txt
> pip install -e . --no-build-isolation
> ```

@JiamingMai Great! Could you fix the format checker with `bash format.sh` locally? After this, I can merge this pr. Thanks!
@Siddhant-Ray Could you help update the documentation accordingly once this is merged?

hickeyma (changes requested):
@JiamingMai Do you mind rebasing an syncing with current code base?

PR #413, opened 2025-03-19
update infinistore connector

lyppg:
fix several infinistore connector issue:
1. support kv_blob
2. queue the get/put requests

PR #440, opened 2025-04-04
Dongmao/refactory infinistore connector

thesues:
hi, upstream

   this is a PR which deprecate #413, for the following reasons:

1. we  found that register_memory is not thread-safe, and should not execute in threadpool. so in the new connector, we pre alllocated buffer to recv and send tensor.
2. CopyLessMemoryObj is introduced here to avoid memcopy in Get() API. But memcpy still happens on Put() API 
3. Known issues: we have preallocated 40MB buffer.  But if chunk size or model is too big. it will raise an excepiton.
4. it includes a new unit-test.

this is my benchmark result comparing all backend using this branch : 
https://github.com/bytedance-iaas/LMCache/tree/release which added additional thread to download kvcache BEFORE decoding.

bench command:

```
python3 vllm/benchmarks/benchmark_serving.py --backend openai-chat --model /data/models/QwQ-32B --base-url http://127.0.0.1:8080 --endpoint /v1/chat/completions --num-prompts 16  --request-rate 1 --metric_percentiles '50,90,95,99' --goodput ttft:5000 tpot:50 --max-concurrency 16 --random-input-len 1000 --random-output-len 250 --dataset-name random --ignore-eos --seed 114
```

TTFT/TPOP results:

|   ----         | PD infinistore      | PD lmcache  | PD redis| origin vllm|
|----------|-------------------|----------------|------|---------|
| TTFT  |1872ms  | 3700ms     |2700ms| 1712ms |
| TPOT   | 50ms   |  59ms  |86ms| 84ms|

compared to origin vllm, infinistore in PD shows very stable and shorter TPOT.  while all TTFT still have to improve.

### DIFF
--- a/lmcache/experimental/memory_management.py
+++ b/lmcache/experimental/memory_management.py
@@ -203,6 +203,16 @@ class TensorMemoryObj(MemoryObj):
         return memoryview(byte_array)
 
 
+class CopyLessMemoryObj(TensorMemoryObj):
+
+    def __init__(self, raw_data, metadata, callback):
+        super().__init__(raw_data, metadata)
+        self.callback = callback
+
+    def __del__(self):
+        self.callback()
+
+
 class BytesBufferMemoryObj(MemoryObj):
     """
     Wraps a raw flat tensor with some metadata
--- a/lmcache/experimental/protocol.py
+++ b/lmcache/experimental/protocol.py
@@ -54,6 +54,22 @@ class RedisMetadata:
     dtype: Optional[torch.dtype]
     fmt: MemoryFormat
 
+    def serialize_into(self, buffer):
+        assert (len(self.shape) == 4), "Shape dimension should be 4"
+
+        struct.pack_into(
+            "iiiiiii",
+            buffer,
+            0,
+            self.length,
+            int(self.fmt.value),
+            DTYPE_TO_INT[self.dtype],
+            self.shape[0],
+            self.shape[1],
+            self.shape[2],
+            self.shape[3],
+        )
+
     def serialize(self) -> bytes:
 
         # NOTE(Jiayi): 4 is the maximum dimension of memory object.
@@ -75,7 +91,7 @@ class RedisMetadata:
     @staticmethod
     def deserialize(s: bytes) -> "RedisMetadata":
         length, fmt, dtype, shape0, shape1, shape2, shape3 = \
-            struct.unpack("iiiiiii", s)
+            struct.unpack_from("iiiiiii", s)
         return RedisMetadata(length,
                              torch.Size([shape0, shape1, shape2, shape3]),
                              INT_TO_DTYPE[dtype], MemoryFormat(fmt))
--- a/lmcache/experimental/storage_backend/connector/infinistore_connector.py
+++ b/lmcache/experimental/storage_backend/connector/infinistore_connector.py
@@ -1,17 +1,15 @@
 import asyncio
 import ctypes
+import operator
+from functools import reduce
 from typing import List, Optional, Union, no_type_check
 
 import infinistore
 import torch
-import operator
-import numpy as np
 
-from functools import reduce
-
-from lmcache.experimental.memory_management import (MemoryAllocatorInterface,
-                                                    MemoryObj,
-                                                    CopyLessMemoryObj)
+from lmcache.experimental.memory_management import (CopyLessMemoryObj,
+                                                    MemoryAllocatorInterface,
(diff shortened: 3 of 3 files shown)


lmcache/experimental/protocol.py
```diff
@@ -54,6 +54,22 @@ class RedisMetadata:
...
+            0,
+            self.length,
+            int(self.fmt.value),
+            DTYPE_TO_INT[self.dtype],
+            self.shape[0],
+            self.shape[1],
+            self.shape[2],
+            self.shape[3],
+        )
+
```
YaoJiayi, line 72: This reduces memory copy, great!

lmcache/experimental/storage_backend/connector/test_infinistore_connector.py
```diff
@@ -0,0 +1,61 @@
+# type: ignore
```
YaoJiayi: Why are we putting the tests here?

lmcache/experimental/storage_backend/connector/infinistore_connector.py
```diff
@@ -39,98 +47,105 @@ def __init__(self, host: str, port: int, dev_name,
...
-        self.buffer = bytearray(self.buffer_size)
-        self.rdma_conn.register_mr(_get_ptr(self.buffer), self.buffer_size)
+        self.send_buffers = []
+        self.recv_buffers = []
+        self.send_queue: asyncio.Queue[int] = asyncio.Queue(
+            maxsize=MAX_BUFFER_CNT)
+        self.recv_queue: asyncio.Queue[int] = asyncio.Queue(
+            maxsize=MAX_BUFFER_CNT)
+
+        # 40MB
```
YaoJiayi: How is this 40MB decided? IIUC, this size should be related to the KV cache size of each chunk?

lmcache/experimental/memory_management.py
```diff
@@ -203,6 +203,16 @@ def byte_array(self) -> bytes:
...
 
 
+class CopyLessMemoryObj(TensorMemoryObj):
+
+    def __init__(self, raw_data, metadata, callback):
+        super().__init__(raw_data, metadata)
+        self.callback = callback
+
+    def __del__(self):
+        self.callback()
```
YaoJiayi, line 213: IIUC, the buffer will be freed during garbage collection?

YaoJiayi (commented):
@thesues Thanks for the effort! Just some minor comments. Btw, the format checks seem to be failing?

lmcache/experimental/storage_backend/connector/test_infinistore_connector.py
thesues: I am not sure if we need a unit test for infinistore here,  we can remove this.

lmcache/experimental/memory_management.py
thesues: Yes,  the  __del__ make sure the buffer is reclaimed.

lmcache/experimental/storage_backend/connector/infinistore_connector.py
thesues: this 40MB is ```hard coded``` for my test. and if any use found 40MB is not big enough(an exception will raise), they have to decrease chunk size. 
I was using  QwQ 7B mode and chunk size = 64. so the max chunk is around 33MB. 40MB is OK to hold all the data.

Right now, I do not know how to guess the MAX_SIZE based on the chunk size and Model.  any ideas?

lmcache/experimental/storage_backend/connector/test_infinistore_connector.py
YaoJiayi: Let's remove for now:)

lmcache/experimental/storage_backend/connector/infinistore_connector.py
YaoJiayi: I think this buffer size can be inferred from `model_config` and `chunk_size`. Basically, you can calculate the size of each KV cache chunk based on these two things. But let's push this to a future PR.

YaoJiayi (commented):
@thesues Let's remove the test and fix the format checker. Then we can merge it. Thanks!

lmcache/experimental/storage_backend/connector/test_infinistore_connector.py
thesues: 1. unit test removed
2. format fixed.

YaoJiayi (approved):
LGTM!

PR #449, opened 2025-04-09
add redis cluster feature

stone-wlg:
add redis cluster feature
example:
remote_url: "redis-cluster://localhost:26379,localhost:26380,localhost:26381"


noskill:
works for me

hickeyma (changes requested):
@stone-wlg Do you mind rebasing an syncing with the code base as there has been refactor in this area?

PR #451, opened 2025-04-09
[Doc] chore: add doc on how to use engine lookup method

rootfs:
I am reading LMCache and having trouble understanding the engine lookup info from the documentation, So I hope this contribution is useful.

### DIFF
--- a/docs/source/developer_tutorial/LMCache_Engine_Examples.rst
+++ b/docs/source/developer_tutorial/LMCache_Engine_Examples.rst
@@ -1,0 +1,167 @@
+.. _dev_doc_LMCache_Engine_Examples:
+
+LMCache Engine Examples
+-----------------------
+
+This page provides programming examples of using the LMCache Engine, demonstrating how to use the core functionality for storing and retrieving cache data.
+
+Store and Retrieve Operations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The LMCache Engine provides three primary operations:
+
+* ``store``: Caches key-value tensors from transformer attention layers
+* ``retrieve``: Fetches cached KV values for token sequences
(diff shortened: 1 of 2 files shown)


Siddhant-Ray:
Hi @rootfs 

Thanks for the contribution, this is really helpful. The documentation just went through a major rewrite, so could you please merge/rebase your patch on top of those?

As you will see, the current developer_documentation has an ``overview.rst`` which is currently empty, maybe you could add your content to that file? The rest of the files will be the interfaces for the developer..

Here's a hosted version of the new docs for the structure: https://docs.lmcache.ai/

rootfs:
@Siddhant-Ray sounds good! i'll move it over there, thanks

rootfs:
Actually, the redis connector config doc is also missing. I'll get that too. @Siddhant-Ray which file do you think is related to this?

Siddhant-Ray:
Thanks for helping out on this, I think the backends we ideally want to create a new section called ``Backend Connectors``
and then under that we can have ``redis.rst``. The ``index.rst`` structure can be like this. Does this make sense to you? Later I can add other backend connectors, once supported.


```
.. toctree::
   :maxdepth: 1
   :caption: Models

   models/models
   
  .. toctree::
   :maxdepth: 1
   :caption: Backend Connectors

   backends/redis

.. toctree::
   :maxdepth: 1
   :caption: Developer Documentation

   developer_tutorial/overview
   developer_tutorial/EngineInterface
   developer_tutorial/GPUConnectorInterface
   developer_tutorial/MemoryObject
   developer_tutorial/MemoryInterface
   developer_tutorial/BackendInterface
```

hickeyma (changes requested):
@rootfs Do you mind rebasing and re-syncing with code base as there has been some refactor in the docs area?

Issue #483, opened 2025-04-17
Refactor Storage Manager’s CPU Hot Cache as a StorageBackendInterface


sammshen:
**Is your feature request related to a problem? Please describe.**

The hot cache in the StorageManager is exposed as an ordered dictionary.

```python
class StorageManager:
        def __init__(…):
		self.hot_cache_: OrderedDict[CacheEngineKey, MemoryObj] = OrderedDict()
```

**Describe the solution you'd like**

Create new StorageBackendInterface subclass in `lmcache/experimental/storage_backend/local_cpu_backend.py` to encapsulate insertion and eviction logic.

```python
class LocalCPUBackend(StorageBackendInterface)
    def __init__(self, config, loop, memory_allocator, lookup_server, dst_device):
	self.dict: OrderedDict[CacheEngineKey, MemoryObj] = OrderedDict()
	self.evictor = LRUEvictor(max_cache_size = config.max_local_cpu_size)
	self.lookup_server = lookup_server
	self.loop = loop
	self.put_tasks: List[CacheEngineKey] = []
	self.cpu_lock = threading.Lock()
	self.memory_allocator = memory_allocator
	self.stream = torch.cuda.Stream() # moved here instead of in the StorageManager
		
    def exists_in_put_tasks(key) -> bool

    def submit_put_task(key, mem_obj) -> Optional[Future]

    def submit_prefetched_task(key) -> Optional[Future]

    def get_blocking(key) -> Optional[MemoryObj]

    def close() 
	
    def make_space_for(shape, dtype)
	“””
	evict keys in the hot cache until we can allocate (shape, dtype) or we run out of memory
	returns Nothing
	“””

    def _update_hot_cache(self, key: CacheEngineKey, memory_obj: MemoryObj):
	“””
	copy the same logic from the current StorageManager's _update_hot_cache()
	“””
```

update `CreateStorageBackends` and make the CPU backend the first in the hierarchy

```python
def CreateStorageBackends(…):
	…
	if config.local_cpu and config.max_local_cpu_size > 0:
        	local_cpu_backend = LocalCPUBackend(config, loop, memory_allocator, dst_device, lookup_server)
        	backend_name = str(local_cpu_backend)
       		storage_backends[backend_name] = local_cpu_backend
	if config.local_disk and config.max_local_disk_size > 0:
		…
	if config.remote_url is not None:
	…
```

update the `StorageManager`

```python
…
[...]

PR #496, opened 2025-04-19
[Core]: Hot Cache Refactor Initial Proposal

sammshen:
[New Proposal](https://github.com/LMCache/LMCache/pull/514)
[Refactor Hot Cache Original Issue/Design](https://github.com/LMCache/LMCache/issues/483)

Initial proposal for how to encapsulate the hot cache logic.

Tests ran so far:
1. new unit/integration test suite: tests/experimental/test_local_cpu_backend.py
- tests: allocation, eviction, ref count consistency, disk backend compatibility (e.g. prefetching)
- not tested: remote backend compatibility
2. tests/experimental/
- 69/69
3. .buildkite/pipeline.yml (executed locally)
- 195/195

I will run more tests/experiments to make sure I didn't break anything. I will update after doing so.

### DIFF
--- a/lmcache/experimental/storage_backend/local_cpu_backend.py
+++ b/lmcache/experimental/storage_backend/local_cpu_backend.py
@@ -1,7 +1,7 @@
 from collections import OrderedDict
 from concurrent.futures import Future
 import threading
-from typing import List, Optional
+from typing import List, Optional, cast
 
 import torch
 
@@ -197,6 +197,9 @@ class LocalCPUBackend(StorageBackendInterface):
         evict_keys = []
 
         self.hot_cache_lock.acquire()
+        if self.real_allocator:
+            assert isinstance(self.memory_allocator, MixedMemoryAllocator)
+            mixed_allocator = cast(MixedMemoryAllocator, self.memory_allocator)
         for evict_key in self.hot_cache_:
             # If the ref_count > 1, we cannot evict it as the hot cache
             # might be used as buffers by other storage backends
@@ -213,8 +216,7 @@ class LocalCPUBackend(StorageBackendInterface):
             # In this way, we don't need to do eviction for big objects
             # TODO(Jiayi): the following code is hacky, please refactor
             if self.real_allocator and \
-                isinstance(self.memory_allocator, MixedMemoryAllocator) and \
-                self.memory_allocator.pin_allocator.num_active_allocations == 0:
+                mixed_allocator.pin_allocator.num_active_allocations == 0:
                 break
         for evict_key in evict_keys:
             self.hot_cache_.pop(evict_key)
--- a/lmcache/experimental/storage_backend/local_cpu_backend.py
+++ b/lmcache/experimental/storage_backend/local_cpu_backend.py
@@ -213,6 +213,7 @@ class LocalCPUBackend(StorageBackendInterface):
             # In this way, we don't need to do eviction for big objects
             # TODO(Jiayi): the following code is hacky, please refactor
             if self.real_allocator and \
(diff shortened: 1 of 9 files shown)


sammshen:
Would it be possible to get permission to see the buildkite unit test results? (and to get the other workflows approved since I am first time contributor) Thanks!

ApostaC:
@sammshen Nice work! I will take a look.

In the meantime, can you fix the formatting issue? (You can run `format.sh` locally)

YaoJiayi:
@sammshen I think you need to do (1) `pip install requirements-lint.txt` and then (2) `bash format.sh` locally.

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,260 @@
...
+                "LocalCPUBackend must be used with a MixedMemoryAllocator"
+
+        # multiple threads can access the hot cache (protects self.hot_cache_)
+        self.hot_cache_lock = threading.Lock()
+
+    def exists_in_put_tasks(self, key: CacheEngineKey) -> bool:
+        """
+        please do not check asynchronous futures for cpu backend
+        """
+        raise NotImplementedError
```
ApostaC, line 56: We can do `return False` here. 
Let's avoid throwing an error because the storage manager may directly call it without checking the type of the StorageBackend

lmcache/experimental/storage_backend/storage_manager.py
ApostaC, line 1: Let's not make changes in `StorageManager` now. The idea is to completely get rid of `hot_cache` in the storage manager, and the CPU backend should be treated the same as other storage backends. We can change it in the followup PRs
(cc @YaoJiayi)

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def submit_put_task(self, key: CacheEngineKey,
+                        memory_obj: MemoryObj) -> Optional[Future]:
+        """
+        please do not run the asynchronous put for cpu backend
+        """
+        raise NotImplementedError
```
ApostaC, line 63: This function can call `put` since it does not have I/O operations

lmcache/experimental/storage_backend/storage_manager.py
ApostaC: I realized the CPU backend might play some extra role when doing prefetching. But it could be decoupled and implement outside of `StorageManager`

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,260 @@
...
+            assert isinstance(self.memory_allocator, MixedMemoryAllocator)
+            mixed_allocator = cast(MixedMemoryAllocator, self.memory_allocator)
+        for evict_key in self.hot_cache_:
+            # If the ref_count > 1, we cannot evict it as the hot cache
+            # might be used as buffers by other storage backends
+            if self.memory_allocator.get_ref_count(
+                    self.hot_cache_[evict_key]) > 1:
+                continue
+            evict_keys.append(evict_key)
+            self.memory_allocator.ref_count_down(self.hot_cache_[evict_key])
```
ApostaC, line 210: Why doing ref_count_down here? The object hasn't been evicted at this line.

ApostaC (changes requested):
Did some reviews. PTAL, thanks!

sammshen:
Closing because of https://github.com/LMCache/LMCache/pull/514

PR #501, opened 2025-04-19
[Build/CI] Read requirements.txt from setup.py

tlrmchlsmth:
Establish a single source of truth for requirements by reading `install_requires` from `requirements.txt`, using the same approach as in vLLM.

Also pin torch to 2.6.0, to match vLLM.


YaoJiayi (approved):
LGTM. Thanks! @tlrmchlsmth

PR #511, opened 2025-04-22
Refactor: rename RedisMetadata to RemoteMetadata

maobaolong:
As https://github.com/LMCache/LMCache/pull/506#discussion_r2053295670 motioned, do this refactor in this PR.

Since the RedisMetadata is used by many connectors already.

### DIFF
--- a/lmcache/experimental/protocol.py
+++ b/lmcache/experimental/protocol.py
@@ -48,7 +48,7 @@ INT_TO_DTYPE = {
 
 
 @dataclass
-class RedisMetadata:
+class CommonMetadata:
     length: int
     shape: torch.Size
     dtype: Optional[torch.dtype]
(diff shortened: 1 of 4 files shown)


lmcache/experimental/protocol.py
```diff
@@ -48,7 +48,7 @@ class Constants:
 
 
 @dataclass
-class RedisMetadata:
+class CommonMetadata:
```
YaoJiayi: Maybe `RemoteMetadata` is a better name?

YaoJiayi (commented):
LGTM in general. Just one minor comment

lmcache/experimental/protocol.py
maobaolong: done

maobaolong:
@YaoJiayi Thanks for your review, PTAL.

YaoJiayi (approved):
LGTM!

maobaolong:

PR #514, opened 2025-04-22
Local CPU Backend

sammshen:
Isolated LocalCPUBackend class, replacing previous PR [Initial Hot Cache Refactor Proposal](https://github.com/LMCache/LMCache/pull/496) <- can be closed if this one gets merged.

TODO after this PR:
- [ ] Integrate LocalCPUBackend into the backend hierarchy in `lmcache/experimental/storage_backend/__init__.py`
- [ ] Refactor special LocalCPUBackend behavior including eviction and prefetching
- [ ] Add pin functionality to cpu backend


format.sh
```diff
@@ -110,9 +110,9 @@ echo 'lmcache yapf: Done'
 
 # Run mypy
 echo 'lmcache mypy:'
-mypy
-mypy tests --follow-imports skip
-mypy lmcache --follow-imports skip
+mypy --cache-dir=/dev/null
+mypy tests --follow-imports skip --cache-dir=/dev/null
+mypy lmcache --follow-imports skip --cache-dir=/dev/null
```
ApostaC: What's the effect of adding `--cache-dir=/dev/null`?

ApostaC (commented):
Otherwise LGTM!
(Btw, are the changes in redis-related source codes made by the format checker?)

format.sh
sammshen: I think my local mypy was caching some type checking errors and this helped it to "forget" after fixing the type issues. I think it makes mypy slower and I just read that it is not best practice. I will delete it, sorry about that!

sammshen: My local format checker was angry about some of the types in the redis-related source code even though the CI/CD checker was not. I will look into making sure my local format checker behaves the same!

lmcache/experimental/storage_backend/connector/redis_connector.py
YaoJiayi, line 1: Why do we have these redis-related changes?

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,121 @@
...
+                return None
+            # ref count up for caller to avoid situation where the memory_obj
+            # is freed from local cpu backend before the caller unlocks the
+            # future
+            self.memory_allocator.ref_count_up(self.dict[key])
+            memory_obj = self.dict[key]
+
+            future: Future = Future()
+            future.set_result(memory_obj)
+            return future
```
YaoJiayi: Can you explain the logic here a bit?

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,121 @@
...
+            self.memory_allocator.ref_count_up(obj)
+
+    def submit_put_task(self, key: CacheEngineKey,
+                        obj: MemoryObj) -> Optional[Future]:
+        """
+        return a dummy future object (result is immediately available so
+        non blocking)
+        """
+        self._put(key, obj)
+        return Future()
```
YaoJiayi: Why are we returning a Future() here?

lmcache/experimental/storage_backend/local_cpu_backend.py
YaoJiayi: Why is prefetch needed for cpu backend for now?

lmcache/storage_backend/connector/redis_connector.py
```diff
@@ -88,14 +88,14 @@ def __init__(self, hosts_and_ports: List[Tuple[str, Union[str, int]]]):
...
-        self.sentinel = redis.Sentinel(hosts_and_ports, timeout)
+        self.sentinel = redis.Sentinel(hosts_and_ports, socket_timeout=timeout)
         self.master = self.sentinel.master_for(service_name,
                                                socket_timeout=timeout)
         self.slave = self.sentinel.slave_for(service_name,
                                              socket_timeout=timeout)
 
     def exists(self, key: str) -> bool:
-        return self.slave.exists(key)
+        return bool(self.slave.exists(key))
```
YaoJiayi, line 98: Same here. Why do we have these redis-related changes?

YaoJiayi (commented):
@sammshen LGTM in general! Left some minor comments:) Let's merge this as soon as possible.

lmcache/experimental/storage_backend/local_cpu_backend.py
sammshen: I think it is not a good design actually and maybe should just leave as `pass` for now until further refactor?

Originally, I wanted to return a future to trigger the `break` at the end of this section of code (when the local cpu is added to the front of the hierarchy) so we don't move onto the LocalDiskbackend.
```python
        for backend in self.storage_backends.values():
            prefetch_task = backend.submit_prefetch_task(key)
            if prefetch_task is None:
                continue
            lambda_callback = lambda f: \
                self.prefetch_callback(f, key)

            self.manager_lock.acquire()
            self.prefetch_tasks[key] = prefetch_task
            prefetch_task.add_done_callback(lambda_callback)
            self.manager_lock.release()
            break
```
I realize this is buggy because the prefetch_callback will actually allocate a *second* memory object and put it into the hot cache, which is a really bad bug. I am really sorry for not being more careful.

lmcache/experimental/storage_backend/local_cpu_backend.py
sammshen: Similar reasoning here. I want to trigger some type of `break` in storage_manager.put() when iterating through the backend hierarchy. I can change all asynchronous/submit methods to`pass` for now?
```python
        #ever_put = False
            put_task = backend.submit_put_task(key, memory_obj)

            if put_task is None:
                continue
                
            else:
               break
```

lmcache/experimental/storage_backend/connector/redis_connector.py
sammshen: Sorry, my local `format.sh` was complaining about some lines in the redis_connector (even though the workflow ones are not).

lmcache/experimental/storage_backend/local_cpu_backend.py
YaoJiayi: No worries:) Let's keep it an empty function then.

lmcache/experimental/storage_backend/local_cpu_backend.py
YaoJiayi: Yeah. Good idea.

YaoJiayi (approved):
@sammshen LGTM! Let's move to the next PR soon

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,191 @@
...
+                self.memory_allocator.ref_count_down(self.dict[evict_key])
+                memory_obj = self.memory_allocator.allocate(shape, dtype)
+                logger.debug("Evicting 1 chunk from cpu memory")
+                if memory_obj is not None:
+                    break
+        for evict_key in evict_keys:
+            self.remove(evict_key)
+        if self.lookup_server is not None:
+            self.lookup_server.batched_remove(evict_keys)
+        return memory_obj
```
ApostaC: IIUC, the eviction logic should only be triggered when the first allocation fails. The "first allocation" is missing in this function.
cc @YaoJiayi

ApostaC (changes requested):
Please see the comments below

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,211 @@
...
+                 memory_allocator: MemoryAllocatorInterface,
+                 lookup_server: Optional[LookupServerInterface] = None,
+                 lmcache_worker: Optional["LMCacheWorker"] = None,
+                 dst_device: str = "cpu"):
+        self.hot_cache: Optional[OrderedDict[CacheEngineKey, MemoryObj]] = None
+        if config.local_cpu:
+            self.hot_cache = OrderedDict()
+        self.lookup_server = lookup_server
+        self.memory_allocator = memory_allocator
+        self.dst_device = dst_device
```
ApostaC: This variable is not used, we can remove it.

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def __init__(self,
+                 config: LMCacheEngineConfig,
+                 memory_allocator: MemoryAllocatorInterface,
+                 lookup_server: Optional[LookupServerInterface] = None,
+                 lmcache_worker: Optional["LMCacheWorker"] = None,
+                 dst_device: str = "cpu"):
+        self.hot_cache: Optional[OrderedDict[CacheEngineKey, MemoryObj]] = None
+        if config.local_cpu:
+            self.hot_cache = OrderedDict()
```
ApostaC: Will the `config.local_cpu` always be true when using LocalCPUBackend?

ApostaC: In this case, we make self.hot_cache non-optimal, and thus can remove all the checks to `self.hot_cache is not None`

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,211 @@
...
+        self.lookup_server = lookup_server
+        self.memory_allocator = memory_allocator
+        self.dst_device = dst_device
+        self.lmcache_worker = lmcache_worker
```
ApostaC, line 46: Do we already have a global function to get the lmcache worker? @YaoJiayi

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,211 @@
...
+
+        evict_keys = []
+        with self.cpu_lock:
+            for evict_key in self.hot_cache:
+                # If the ref_count > 1, we cannot evict it as the cpu memory
+                # might be used as buffers by other storage backends
+                if self.memory_allocator.get_ref_count(
+                        self.hot_cache[evict_key]) > 1:
+                    continue
+                evict_keys.append(evict_key)
```
ApostaC, line 162: We also need to update the stats monitor here when evicting

lmcache/experimental/storage_backend/local_cpu_backend.py
```diff
@@ -0,0 +1,211 @@
...
+        with self.cpu_lock:
+            for key in self.hot_cache:
+                memory_obj = self.hot_cache[key]
+                if self.memory_allocator.get_ref_count(memory_obj) > 1:
+                    continue
+                clear_keys.append(key)
+                self.memory_allocator.ref_count_down(memory_obj)
+
+        for key in clear_keys:
+            self.remove(key)
```
ApostaC, line 200: Need to update the stats_monitor here

ApostaC (changes requested):
Some more comments about the neatness and stats monitor consistency

lmcache/experimental/storage_backend/local_cpu_backend.py
YaoJiayi: Not yet. I want to push all the controller-related PRs after CPU backend refactor

ApostaC: Okay

sammshen:
Sorry for requesting reviews. I saw that this PR is blocking other ones and I don't want to slow other important progress down.

ApostaC (approved):
Thanks @sammshen , great work! LGTM!

PR #524, opened 2025-04-24
[bugfix] fix counter type prometheus metrics error

chunxiaozheng:
In our grafana, VLLM is in an idle state, but we found that the following metrics are constantly increasing
- `lmcache:num_retrieve_requests`
- `lmcache:num_store_requests`
- `lmcache:num_requested_tokens`
- `lmcache:num_hit_tokens`

These four Counter type metrics update data through `_log_counter()`, which implementation is as follows:
```
def _log_counter(self, counter, data: Union[int, float]) -> None:
    # Convenience function for logging to counter.
    # Prevent ValueError from negative increment
    if data < 0:
        return
    counter.labels(**self.labels).inc(data)
```

`counter.labels(**self.labels).inc(data)` use `inc()` to update data, so we must use incremental data to update metrics instead of full data.

### DIFF
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -363,7 +363,8 @@ class PrometheusLogger:
 
         self._log_counter(self.counter_num_requested_tokens,
                           stats.interval_requested_tokens)
-        self._log_counter(self.counter_num_hit_tokens, stats.interval_hit_tokens)
+        self._log_counter(self.counter_num_hit_tokens,
+                          stats.interval_hit_tokens)
 
         self._log_gauge(self.gauge_cache_hit_rate, stats.cache_hit_rate)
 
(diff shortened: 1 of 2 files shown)


ApostaC:
Thanks! This is a good catch! I will take a look soon

Deeperfinder:
hi，how to import those metrics into grafana,  could you offer a grafana.json?

ApostaC (approved):
LGTM!

ApostaC:
@YaoJiayi @maobaolong PTAL, thanks!

chunxiaozheng:
> hi，how to import those metrics into grafana, could you offer a grafana.json?

Sorry, I am on vacation and will submit it when I return.

maobaolong (approved):
This is tested in our test cluster.

Before this PR.

image

After
image

PR #543, opened 2025-04-28
Fix p2p in vllm v1

orozery:
This PR fixes the lookup function (used by vllm v1) to support p2p mode.

### DIFF
--- a/lmcache/experimental/cache_engine.py
+++ b/lmcache/experimental/cache_engine.py
@@ -362,10 +362,32 @@ class LMCacheEngine:
         :return: An int indicating how many prefix tokens are cached.
         """
         end = 0
+        search_local = True  # we always lookup local storage_manager first
+        # secondary lookup on p2p (via lookup_server) if enabled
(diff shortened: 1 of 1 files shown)


maobaolong:
@orozery Is there a way to run p2p while TP>1?

orozery:
> @orozery Is there a way to run p2p while TP>1?

I never tried it, but looking at the LMCache code I do see it references parameters from vllm's ParallelConfig.
So maybe it will just work if you run with LMCACHE_ENABLE_P2P and vllm's tensor-parallel-size parameter.
Again, I did not try it, and specifically I don't think it matters in the context of this PR.

Siddhant-Ray:
@orozery Can you fix the format checker issues by running ``format.sh`` locally?

HuaizhengZhang (approved):
LGTM. Very detailed comments :)

KuntaiDu (approved):
LGTM!

ApostaC (approved):
LGTM!

PR #592, opened 2025-05-07
[CI/Build]: Add project metadata to pyproject toml file

hickeyma:
The [pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) is the more standard way now to handle the project metadata.  This PR moves the metadata from `setup.py` to the `pyproject.toml` file.
It also uses the new [pep 639](https://peps.python.org/pep-0639/) fields for [license and license-files](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files). This requires `setuptools >= 77.0.3`. 

Finally, [dynamic versioning](https://setuptools-scm.readthedocs.io/en/latest/) from `setuptools_scm` is added . This removes the need to manage a static version file as the version file is generated on the fly using [git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging).

~**Note:** Do not merge until `v0.2.11` tag is removed or updated. This tag is not being recognized by `setuptools-scm` because it is not verified.~ Resolved, thanks @sammshen

PR #644, opened 2025-05-14
[Fix]: Torch version for build

hickeyma:
Rollback change from #592

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
     "packaging>=24.2",
     "setuptools>=77.0.3,<80.0.0",
     "setuptools_scm>=8",
-    "torch==2.6.0",
+    "torch==2.7.0",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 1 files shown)


Siddhant-Ray:
@Shaoting-Feng @ApostaC Do we know why buildkite is failing here?

Shaoting-Feng:
`setup.py` or `[project.dependencies]` in `pyproject.toml` didn't specify torch version.

ApostaC:
The error itself is because the pytorch version in uv venv and in the temp pip build environment do not match. So we should either specify both venv installation and the pyproject.toml, or use `--no-build-isolation` when installing LMCache

hickeyma:
> setup.py or [project.dependencies] in pyproject.toml didn't specify torch version.

Requirements are loaded in https://github.com/LMCache/LMCache/blob/dev/setup.py#L61

hickeyma:
> The error itself is because the pytorch version in uv venv and in the temp pip build environment do not match.

@ApostaC How are they managing to have different versions of torch? The version in the build section and reqs file are the same versions.

Shaoting-Feng:
It seems that the latest build of your branch is successful now.
Screen Shot 2025-05-14 at 14 11 41 PM

Shaoting-Feng (approved):
LGTM

Siddhant-Ray (approved):
LGTM

PR #653, opened 2025-05-15
[Bugfix] Fix incorrect single-token saves in v1

orozery:
After saving a (256 token) chunk, `num_saved_tokens % chunk_size` is 0.
After the next decoded token, chunk_boundary will incorrectly be set to equal num_saved_tokens, instead of num_saved_tokens+chunk_size.
This will yield an additional unwanted 1 token save.
For example:
```
LMCache INFO: Storing KV cache for 1 out of 4097 tokens for request
```

This PR fixes `chunk_boundary` to avoid these incorrect writes.


orozery:
@ApostaC this is a fix to code introduced in #490

ApostaC (approved):
Good catch! LGTM!

Siddhant-Ray (approved):

Issue #660, opened 2025-05-16
[Bug]: redis_connector report NotImplementedError: memoryview: unsupported format <B


sydnash:
while use vllm 0.8.5.post1 with lmcache 0.2.1 with the redis_connector. and my python version is 3.12
it ok for save kv cache to redis, but while retrive kv cache from redis, the redis_connector report an error:
```
ERROR 05-16 15:25:00 [worker_base.py:620] Error executing method 'execute_model'. This might cause deadlock in distributed execution.
ERROR 05-16 15:25:00 [worker_base.py:620] Traceback (most recent call last):
ERROR 05-16 15:25:00 [worker_base.py:620]   File "/LocalRun/jun.dai/conda/envs/test_lmcache/lib/python3.12/site-packages/vllm/worker/worker_base.py", line 612, in execute_method
ERROR 05-16 15:25:00 [worker_base.py:620]     return run_method(self, method, args, kwargs)
ERROR 05-16 15:25:00 [worker_base.py:620]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 05-16 15:25:00 [worker_base.py:620]   File "/LocalRun/jun.dai/conda/envs/test_lmcache/lib/python3.12/site-packages/vllm/utils.py", line 2456, in run_method
ERROR 05-16 15:25:00 [worker_base.py:620]     return func(*args, **kwargs)
ERROR 05-16 15:25:00 [worker_base.py:620]            ^^^^^^^^^^^^^^^^^^^^^
ERROR 05-16 15:25:00 [worker_base.py:620]   File "/LocalRun/jun.dai/conda/envs/test_lmcache/lib/python3.12/site-packages/vllm/worker/worker_base.py", line 420, in execute_model
ERROR 05-16 15:25:00 [worker_base.py:620]     output = self.model_runner.execute_model(
ERROR 05-16 15:25:00 [worker_base.py:620]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 05-16 15:25:00 [worker_base.py:620]   File "/LocalRun/jun.dai/conda/envs/test_lmcache/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
ERROR 05-16 15:25:00 [worker_base.py:620]     return func(*args, **kwargs)
ERROR 05-16 15:25:00 [worker_base.py:620]            ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #662, opened 2025-05-16
fix the memoryview: unsupported format <B error at redis connector

sydnash:
issue: https://github.com/LMCache/LMCache/issues/660
change the view's format form <B to B while receive kv cache from redis.


KuntaiDu (approved):
LGTM!

KuntaiDu:
The mypy check seems to be broken. We will fix this later.

eryugey:
Thanks for the fix! I hit the same bug when playing with redis backend, and came out with a similar fix.

Just one minor question, why not cast memory_obj.byte_array to 'B' unconditionally? It's basically a no-op if it is already in 'B' format. So we can remove the format check.

sydnash:
> Thanks for the fix! I hit the same bug when playing with redis backend, and came out with a similar fix.
> 
> Just one minor question, why not cast memory_obj.byte_array to 'B' unconditionally? It's basically a no-op if it is already in 'B' format. So we can remove the format check.

if memory_obj.byte_array is a memoryview object and it's format is not '<B', we cannot safely change the format to 'B', it will case inconsistent binary format. 
'<B' format can change to 'B' is because Bytes are independent of endianness.

KuntaiDu:
@ApostaC @YaoJiayi @Siddhant-Ray could you please take a quick look at this PR?

ApostaC (approved):
LGTM!

Issue #594, opened 2025-05-07
[CD] Continuous Artifacts (Wheels and Images) of LMCache


sammshen:
LMCache should have a binary executable that is pip-installable after every new merged PR. 

There are two TODOs for generating nightly artifacts of LMCache:

1. TestPyPI

@hickeyma thanks for raising this issue and lending us your expertise :)

If you use `twine` to set up the CI workflow uploading like the current `publish.yml`,

```yaml
      - name: Publish to PyPI
        env:
          TWINE_USERNAME: __token__
          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
        run: twine upload dist/*
```

we will make sure to add the TEST_PYPI_API_TOKEN to LMCache GH secrets. 

**May 12th, 2025 Updates:**

Thanks for the PRs @hickeyma ! :)

https://github.com/LMCache/LMCache/pull/611
https://github.com/LMCache/LMCache/pull/592

TODOs for LMCache Team on Test PyPI:
a. Add the `.github/workflows/publish.yml` workflow in the LMCache repository as a "trusted publisher" to the LMCache PyPI and TestPyPI project via: https://docs.pypi.org/trusted-publishers/adding-a-publisher/
b. Add `pip install --index-url https://test.pypi.org/simple --extra-index-url https://pypi.org/simple lmcache` instructions to `installation.rst` documentation so that users know how to get latest lmcache wheels (not stable release)
c. At some point in the future, we need to change the egress-policy to block in `publish.yml` (not urgent)

2. Docker Image

The second artifact should be a Docker Image, which should update `lmcache/vllm-openai:latest` (contains vllm and lmcache with cpu offloading) as a cron job that runs before nightly benchmarking. 

https://hub.docker.com/r/lmcache/vllm-openai/tags

This is the dockerhub link to the image that needs to updated regularly (probably daily). The tag that should be updated is probably the `latest` tag. 

The Dockerfile that produces this image can be found in `docker/Dockerfile` along with all of its dependencies. The end goal will be to have a workflow that runs every night that will build and upload this image to dockerhub with latest tag!

**June 10th, 2025 Conclusion:**
Thank you @hickeyma for your incredibly fast work on automating continuous deployment! This will allow all future development of lmcache to iterate much faster now since we can easily run benchmarks on the latest version and don't have to do manual builds of wheels or images

High Level Retrospective:
[...]

PR #665, opened 2025-05-16
[CI/Build]: Update docker specification to be compatible with latest vLLM OpenAI server

hickeyma:
Updated the docker specification as follows:

- Removed redundant commands
- Updated requirements to latest packages
- Add build args to make build configurable for different vLLM images
- Added example build script

Partial #594

Tested by:
- Building image using latest vLLM image (`b8ed18f4b73f73f68fb238533a2ac1d61e471f50`) in AWS ECR registry: `public.ecr.aws/q9t5s3a7/vllm-ci-test-repo`
- Run the container which exposese the vLLM OpenAI server
- Test with `curl` command as per https://docs.lmcache.ai/getting_started/quickstart/offload_kv_cache.html

**Note to reviewers:** This is 1st step in providing pipeline for nightly build images as part of #594. There will be follow on PRs for centralizing requirement files and then pipeline file(s).


docker/example_run.sh
```diff
@@ -9,6 +19,5 @@ docker run --runtime nvidia --gpus all \
     --network host \
     --entrypoint "/usr/local/bin/vllm" \
     $IMAGE \
-    serve mistralai/Mistral-7B-Instruct-v0.2 --kv-transfer-config \
-    '{"kv_connector":"LMCacheConnector","kv_role":"kv_both"}' \
-    --enable-chunked-prefill false
+    serve $HF_MODEL_NAME --kv-transfer-config \
```
sammshen: Great catch on the deprecated --enable-chunked-prefill flag and new connector for vllm v1! Am I mistaken that the "serve" here will cause a redundancy since the ENTRYPOINT is already ["vllm", "serve"]?

docker/Dockerfile
```diff
@@ -1,16 +1,25 @@
-# The vLLM Dockerfile is used to construct vLLM image that can be directly used
-# to run the OpenAI compatible server.
+# The LMcache Dockerfile is used to build a LMCache image that is integrated
+# to run with vLLM OpenAI server.
 
-# Please update any changes made here to
-# docs/source/dev/dockerfile/dockerfile.rst and
-# docs/source/assets/dev/dockerfile-stages-dependency.png
+# Please update any changes made here to:
+# docs/source/developer_guide/docker_file.rst and
```
sammshen: for the recently revamped the docs so I think these paths no longer exist.

would you mind editing to:
```text
# Instructions on installing lmcache/vllm-openai will be in:
# docs/source/getting_started/installation.rst
```

thank you!

docker/requirements.txt
sammshen, line 1: Could I please just clarify why the new `requirements.txt` file is smaller than the old `requirements-build.txt`? 
e.g. 
```text
filelock >= 3.10.4 # filelock starts to support `mode` argument from 3.10.4
partial-json-parser # used for parsing partial JSON outputs
```
are removed? 

Also, sorry if this is a dumb question, but you mentioned in https://github.com/LMCache/LMCache/issues/594#issuecomment-2887109824 consolidating the requirements. Is this generally best practice? If so, should we just merge them all in this PR as well? :)

docker/example_build.sh
```diff
@@ -0,0 +1,19 @@
+# Example script to build the LMCache container image
+
+# Update the following variables accordingly
+# Note: latest (built from GitHub) images of vLLM are in AWS ECR registry:
+# public.ecr.aws/q9t5s3a7/vllm-ci-test-repo
```
sammshen, line 5: should lmcache also use ECR for latest images too or is DockerHub a perfectly viable option? I left a proposal about how to release images:
https://github.com/LMCache/LMCache/issues/594#issuecomment-2889322546

docker/requirements.txt
hickeyma: Good question. At this moment I am using [this requirements file](https://github.com/LMCache/LMCache/blob/dev/requirements.txt) as the file that defines the mandatory dependencies for LMCache.  I therefore replaced [requirements common](https://github.com/LMCache/LMCache/blob/dev/docker/requirements-common.txt) with this file. I have followed this approach to remove any unnecessary dependencies and from there identify if any dependencies are now missing.  I will add the diff below and maybe you can let me know if any dependencies need to be added back in and why?

hickeyma:
```command
$ diff requirements.txt requirements-common.txt

1,4d0
< # Copy of requirements.txt6 file in LMCache root
< # Remove this redundancy when add all requirement file to requirements directory in root
< 
< aiofiles
6c2,9
< infinistore
---
> einops # Required for Qwen2-VL.
> fastapi < 0.113.0; python_version < '3.9'
> fastapi >= 0.114.1; python_version >= '3.9'
> filelock >= 3.10.4 # filelock starts to support `mode` argument from 3.10.4
> gguf == 0.10.0
> importlib_metadata
> lm-format-enforcer == 0.10.6
> mistral_common >= 1.4.3
8,10c11,15
< nixl==0.2.0
< numpy
< nvtx
---
> numpy < 2.0.0
> openai >= 1.40.0 # Ensure modern openai package (ensure types module present)
> outlines >= 0.0.43, < 0.1
> partial-json-parser # used for parsing partial JSON outputs
> pillow  # Required for image processing
11a17,18
> prometheus-fastapi-instrumentator >= 7.0.0
> protobuf # Required by LlamaTokenizer.
12a20,21
> py-cpuinfo
> pydantic >= 2.9  # Required for fastapi >= 0.113.0
14,21c23,33
< pyzmq >= 25.0.0
< redis
< safetensors
< setuptools>=77.0.3,<80.0.0
< setuptools_scm>=8
< sortedcontainers
< torch==2.7.0 # Should correspond to the version used by latest vLLM package
< transformers >= 4.51.1
---
> pyzmq
> requests
> sentencepiece  # Required for LLaMA tokenizer.
> setuptools>=74.1.1; python_version > '3.11' # Setuptools is used by triton, we need to ensure a modern version is installed for 3.12+ so that it does not try to import distutils, which was removed in 3.12
> six>=1.16.0; python_version > '3.11' # transitive dependency of pandas that needs to be the latest version for python 3.12
> tiktoken >= 0.6.0  # Required for DBRX tokenizer
> tokenizers >= 0.19.1  # Required for Llama 3.
> tqdm
> transformers >= 4.45.0  # Required for Llama 3.2.
> typing_extensions >= 4.10
> uvicorn[standard]
```

hickeyma: > Also, sorry if this is a dumb question, but you mentioned in https://github.com/LMCache/LMCache/issues/594#issuecomment-2887109824 consolidating the requirements. Is this generally best practice? If so, should we just merge them all in this PR as well? :)

Good question as well. I would prefer to do this in a follow on PR as it will mean changes across the project and not just within the context of the container image changes for this PR.

docker/example_run.sh
hickeyma: Good catch, yes, its redundant. Updated `docker run` command to just pass the args to the `vllm serve` `entrypoint` in the image, i.e. `$HF_MODEL_NAME --kv-transfer-config`

docker/Dockerfile
hickeyma: Updated

docker/example_build.sh
hickeyma: @sammshen I think both are viable options. I'll do some investigation on this to understand this better.

hickeyma:
Thank you @sammshen for the review and feedback. Updated and questions responded to.

Ready for review again.

docker/requirements.txt
sammshen: makes sense now! Since you already did the entire building and testing of the image, this sounds great and I personally don't know why the extra dependencies are there so we could just add them in the future if they are needed (e.g. one of them looks Qwen-specific?)

sammshen (approved):

sammshen:
LGTM!!

Siddhant-Ray (approved):
LGTM!

docker/example_run.sh
```diff
@@ -1,14 +1,22 @@
...
+IMAGE=
+HF_MODEL_NAME='meta-llama/Llama-3.1-8B-Instruct'
+RUNTIME=nvidia
+
+docker run --runtime=$RUNTIME --gpus all \
+    --env "HF_TOKEN=<HF_TOKEN_ACCESS_MODEL>" \
     --env "LMCACHE_USE_EXPERIMENTAL=True" \
     --env "chunk_size=256" \
     --env "local_cpu=True" \
     --env "max_local_cpu_size=5" \
```
ApostaC, line 17: Just realized this command is not correct. 
@hickeyma Can you do us a favor to quickly change the `--env` commands to the following: 
```bash
    --env "LMCACHE_CHUNK_SIZE=256" \
    --env "LMCACHE_LOCAL_CPU=True" \
    --env "LMCACHE_MAX_LOCAL_CPU_SIZE=5" \
```

Thanks!

(The full command can be found here: https://docs.lmcache.ai/developer_guide/docker_file.html)

ApostaC (approved):
LGTM!

docker/example_run.sh
ApostaC: Actually NVM. I will create an issue for this

sammshen: I can do this small fix rn!

hickeyma: Good catch @ApostaC on the env variables. Thanks for raising issue #682

PR #666, opened 2025-05-16
[Misc] Add async sending based on async V1 connector API updates

njhill:
We've now extended the vLLM V1 connector API to support async saves/loads which can span multiple steps.

This PR adds optional support to the v1 vllm adapter to exploit this for saving. Similar could be done for loading but the tradeoff there is slightly difference since it could impact TTFT for individual requests (while hopefully still benefiting overall latency and throughput).

 I'm curious about whether this higher level async saving obviates the need for the recently-added layer-wise saving.
 
 
This also requires a couple of very minor (but backwards compatible) adjustments to the connector wrapper impl on the vLLM side. I've opened a draft to show that too: https://github.com/vllm-project/vllm/pull/18268

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -11,10 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import os
 import threading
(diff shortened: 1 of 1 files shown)


qiucy:
@njhill  hi njhill, Is the code for this branch ready?

njhill:
@qiucy sorry for missing your comment. Yes, I think the code was ready but now requires conflict resolution, and I'm not sure whether or not it would still be compatible/required given changes that have been made since.

PR #678, opened 2025-05-18
Micro benchmark for VLLMPagedMemGPUConnectorV2.toGpu and a performance fix

yanok:
While profiling, we've seen a ridiculous amount of time (well, proportional) spent in `VLLMPagedMemGPUConnectorV2.toGpu`, in particular in `_pointersAreGood` method.

This PR adds a micro benchmark and a fix that improves the outcome of this benchmark ~3.3x.

### DIFF
--- a/lmcache/experimental/gpu_connector.py
+++ b/lmcache/experimental/gpu_connector.py
@@ -280,9 +280,10 @@ class VLLMPagedMemGPUConnectorV2(GPUConnectorInterface):
         self.num_layers = num_layers
         self.kv_cache_pointers = torch.empty(num_layers,
                                              dtype=torch.int64,
-                                             device='cpu',
-                                             pin_memory=True)
-        self.pointers_initialized = False
+                                             device='cpu')
+        # Not sure we need a dict here. Maybe a single GPU connector always
+        # works with a single device?
+        self.kv_cache_pointers_on_gpu: dict[int, torch.Tensor] = {}
         self.page_buffer_size = 0
 
         self.gpu_buffer: Optional[torch.Tensor] = None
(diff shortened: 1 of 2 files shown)


YaoJiayi:
@yanok Thanks! Do you know the reason?

yanok:
I wrote some details in the commit message, tldr: get_item/set_item on
torch tensors are inefficient.

ApostaC (approved):
LGTM!

YaoJiayi:
@yanok In https://gist.github.com/yanok/685182fe15b46bf0d99797853a4f9f9a, why `0002_c01781c` is faster than `Now`? Is `Now` your current branch?

yanok:
> @yanok In https://gist.github.com/yanok/685182fe15b46bf0d99797853a4f9f9a, why `0002_c01781c` is faster than `Now`? Is `Now` your current branch?

No, vice versa. I should have named the measurements :) I first ran benchmark with --benchmark-autosave from the branch tip, so that's `0002_c01781c` and then did `git checkout HEAD^` and ran with --benchmark-compare to compare before the change to after the change.

lmcache/experimental/gpu_connector.py
```diff
@@ -298,29 +299,21 @@ def __init__(self,
...
-        for i in range(self.num_layers):
-            if self.kv_cache_pointers[i] != kv_caches[i].data_ptr():
-                return False
-        return True
-
-    def _initialize_pointers(self, kv_caches: List[torch.Tensor]):
-        for i in range(self.num_layers):
-            self.kv_cache_pointers[i] = kv_caches[i].data_ptr()
-        self.pointers_initialized = True
+    def _initialize_pointers(self,
```
YaoJiayi, line 302: `_initialize_pointers` here is executed every time, but `_initialize_pointers` in the old code is only called once during the first execution. IIUC, you are saying that the `_pointers_are_good` is slow only because of `if self.kv_cache_pointers[i] != kv_caches[i].data_ptr():` ?

YaoJiayi (commented):
Thanks @yanok ! Just one last question:)

lmcache/experimental/gpu_connector.py
yanok: Short answer: yes, the slowness was in this line.

Longer answer: there are multiple layers of this.
* First, we observed that most of `toGpu` time is spent inside some torch function with an obscure name, which we deciphered to be `!=` overload. So, making it `int(self.kv_cache_pointers[i]) != ...` helped a bit, but now a lot of time were spent in torch's `get_item`.
* At this point, I started thinking why we do `_pointers_are_good` at all? In the happy case in has to iterate over all elements, but so does `_initialize_pointers`. And in unhappy cases, we still have to `_initialize_pointers` after spending some time in `_pointers_are_good`.
* Finally, for `_initialize_pointers` we could use an overloaded array assignment, which is much more efficient than doing `set_item` from a Python loop.

Does that answer your question?

YaoJiayi (approved):
LGTM!

Issue #682, opened 2025-05-19
[Misc] Fix problem in the docker example script


ApostaC:
The docker run command in the [example script](https://github.com/LMCache/LMCache/blob/dev/docker/example_run.sh) is incorrect.

Please refer to https://docs.lmcache.ai/developer_guide/docker_file.html for the correct docker run command.

wwl2755:
Hi, @ApostaC 
I would like to work on this. Thanks!

PR #685, opened 2025-05-20
[Misc] Fix docker example script

wwl2755:
Fix docker example script as described in #682, and also some typo I met during the exploration.

### DIFF
--- a/docker/example_run.sh
+++ b/docker/example_run.sh
@@ -5,7 +5,7 @@
 # https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html
 
 # Set the following variables:
-IMAGE=lmcache/vllm-openai:2025-05-17-v1 # WARNING: Cannot use latest or omit the tag!!!
+IMAGE=<IMAGE_NAME>:<TAG>
 HF_MODEL_NAME='meta-llama/Llama-3.1-8B-Instruct'
 RUNTIME=nvidia
 
--- a/docker/example_run.sh
+++ b/docker/example_run.sh
@@ -5,16 +5,16 @@
 # https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html
 
 # Set the following variables:
-IMAGE=
+IMAGE=lmcache/vllm-openai:2025-05-17-v1 # WARNING: Cannot use latest or omit the tag!!!
 HF_MODEL_NAME='meta-llama/Llama-3.1-8B-Instruct'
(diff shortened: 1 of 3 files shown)


docker/example_run.sh
```diff
@@ -5,16 +5,16 @@
 # https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html
 
 # Set the following variables:
-IMAGE=
+IMAGE=lmcache/vllm-openai:2025-05-17-v1 # WARNING: Cannot use latest or omit the tag!!!
```
hickeyma: Lets not set that here. Want the image to be specified. ~Also, the `latest` tag seems to be working (see https://github.com/LMCache/LMCache/issues/686#issuecomment-2893607922).~ Thanks @wwl2755 for informing me of my misread in #686

hickeyma:
```suggestion
IMAGE=<IMAGE_NAME>:<TAG>
```

hickeyma (changes requested):
Thanks @wwl2755 for pushing this fix. Small suggestion inline.

wwl2755:
Hi @hickeyma 
Thanks for the comments. But I believe the docker under lmcache and vllm can sometimes be different from each other. And it would be better to "suggest" a reasonable place to find the image. 

For example, recently lmcache team made several api modification merged into main of vllm, but they are not included in the latest release (latest vllm docker). In such scenario, lmcache team could make a docker image on our own under lmcache/vllm-openai to provide immediate usage. 

I'm good with the suggestion (leaving it blank), but would love to bring my concern into discussion. What's your opinion? @YaoJiayi @ApostaC

wwl2755:
Probably I should leave a comment there, instead of specifying the version (as it is right now)

hickeyma:
> Thanks for the comments. But I believe the docker under lmcache and vllm can sometimes be different from each other. And it would be better to "suggest" a reasonable place to find the image.

This could be added as a comment and I would avoid using the specific tag (and go with `latest`). The `latest` tag should be fixed shortly hopefully. The reason I would rather not specify the image and tag by default is to avoid a user running the script which uses the latest image by default and they might not want or expect to be it. In this way, the user needs to be specific.

wwl2755:
Thanks @hickeyma It's updated. PTAL. Thank you!

hickeyma (approved):
LGTM, thanks @wwl2755

YaoJiayi (approved):
lgtm

ApostaC (approved):
LGTM!

PR #699, opened 2025-05-22
Introduce Weka Storage Backend

sdimitro:
This PR Involves basically two changes:
1. A new storage backend optimized for the Weka Filesystem
2. One minor change that makes the allocator accept a configurable byte alignment value - we've seen improved performance by using 4K alignment with Weka due to the filesystem's characteristics. There is a high chance other filesystems may benefit from this.

This is an introductory PR that implements the essential storage backend APIs. Once this gets accepted there are a few things that we'll add as follow ons involving but not limited to:
* Integration with LMCache Statistics
* Pin/Unpin
* Work to completely avoid ading the LocalCPU backend to the hierarchy
* Optional configuration for thread pools

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

`git tag vx.x.x`

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

`git tag -d vx.x.x`

For example:
`git tag -d v0.3.0`

`gh release create vx.x.x --repo LMCache/LMCache --title "vx.x.x" --notes "<Add decsription>"`

For example:
`gh release create v0.3.0 --repo LMCache/LMCache --title "v0.3.0" --notes "LMCache v0.3.0 is a feature realease. Users are encouraged to upgrade for the best experience."`

> [!TIP]

### DIFF
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -98,6 +98,12 @@ class LMCacheEngineConfig:
     # The url of the actual remote lmcache instance for auditing
     audit_actual_remote_url: Optional[str] = None
 
+    # (Optional) Weka/CuFile related configurations
+    # The path under the WekaFS mount that the cache will be stored
+    weka_path: Optional[str] = None
+    # Size of CuFile Buffer in MiB
+    cufile_buffer_size: Optional[int] = None
+
     @staticmethod
     def from_defaults(
         chunk_size: int = 256,
@@ -306,6 +312,9 @@ class LMCacheEngineConfig:
 
         audit_actual_remote_url = config.get("audit_actual_remote_url", None)
 
+        weka_path = config.get("weka_path", None)
+        cufile_buffer_size = config.get("cufile_buffer_size", None)
+
         local_disk_path = _parse_local_disk(local_disk)
 
         match remote_url:
@@ -346,6 +355,8 @@ class LMCacheEngineConfig:
                 nixl_buffer_device,
                 nixl_enable_gc,
                 audit_actual_remote_url,
+                weka_path,
+                cufile_buffer_size,
             )
             .validate()
             .log_config()
@@ -499,6 +510,15 @@ class LMCacheEngineConfig:
             get_env_name("audit_actual_remote_url"),
             config.audit_actual_remote_url,
         )
+
+        config.weka_path = parse_env(
+            get_env_name("weka_path"),
+            config.weka_path,
+        )
+        config.cufile_buffer_size = parse_env(
+            get_env_name("cufile_buffer_size"),
+            config.cufile_buffer_size,
+        )
         return config.validate().log_config()
 
     def to_original_config(self) -> orig_config.LMCacheEngineConfig:
@@ -552,6 +572,7 @@ class LMCacheEngineConfig:
         """log the configuration in LMCache"""
         config_dict = {
             "chunk_size": self.chunk_size,
+            "cufile_buffer_size": self.cufile_buffer_size,
             "local_cpu": self.local_cpu,
             "max_local_cpu_size": f"{self.max_local_cpu_size} GB",
             "local_disk": self.local_disk,
@@ -575,6 +596,7 @@ class LMCacheEngineConfig:
             "nixl_buffer_size": self.nixl_buffer_size,
             "nixl_buffer_device": self.nixl_buffer_device,
             "nixl_enable_gc": self.nixl_enable_gc,
+            "weka_path": self.weka_path,
         }
         logger.info(f"LMCache Configuration: {config_dict}")
 
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,5 @@ lmcache/experimental/tests
 # Setuptools scm version
 /lmcache/_version.py
 
+# Miscellaneous
+cufile.log
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -80,6 +80,7 @@ Documentation
    kv_cache/infinistore
    kv_cache/mooncake
    kv_cache/valkey
+   kv_cache/weka
 
 :raw-html:`<br />`
 
--- a/docs/source/kv_cache/weka.rst
+++ b/docs/source/kv_cache/weka.rst
@@ -1,0 +1,131 @@
+Weka
+====
+
+.. _weka-overview:
+
+Overview
+--------
+
+WekaFS is a high-performance, distributed filesystem and is a supported option for KV Cache offloading in
+LMCache. Even though the local filesystem backend can work with a WekaFS mount, this particular backend is
+optimized for Weka's characteristics. It leverages GPUDirect Storage for I/O and it allows data-sharing
+between multiple LMCache instances.
+
+Ways to configure LMCache WEKA Offloading
+-----------------------------------------
+
+**1. Environment Variables:**
+
+``LMCACHE_USE_EXPERIMENTAL`` MUST be set by environment variable directly.
+
+.. code-block:: bash
+
+    # Specify LMCache V1
+    export LMCACHE_USE_EXPERIMENTAL=True
+    # 256 Tokens per KV Chunk
+    export LMCACHE_CHUNK_SIZE=256
+    # Path to Weka Mount
+    export LMCACHE_WEKA_PATH="/mnt/weka/cache"
+    # CuFile Buffer Size in MiB
+    export LMCACHE_CUFILE_BUFFER_SIZE="8192"
+    # Disabling CPU RAM offload is sometimes recommended as the
+    # CPU can get in the way of GPUDirect operations
+    export LMCACHE_LOCAL_CPU=False
+
+**2. Configuration File**:
+
+Passed in through ``LMCACHE_CONFIG_FILE=your-lmcache-config.yaml``
+
+``LMCACHE_USE_EXPERIMENTAL`` MUST be set by environment variable directly.
+
+Example ``config.yaml``:
+
+.. code-block:: yaml
+
+    # 256 Tokens per KV Chunk
+    chunk_size: 256
+    # Disable local CPU
+    local_cpu: false
+    # Path to Weka Mount
+    weka_path: "/mnt/weka/cache"
+    # CuFile Buffer Size in MiB
+    cufile_buffer_size: 8192
+
+CuFile Buffer Size Explanation
+------------------------------
+
+The backend currently pre-registers buffer space to speed up cuFile operations. This buffer space
+is registered in VRAM so options like ``--gpu-memory-utilization`` from ``vllm`` should be considered
+when setting it. For example, a good rule of thumb for H100 which generally has 80GiBs of VRAM would
+be to start with 8GiB and set ``--gpu-memory-utilization 0.85`` and depending on your workflow fine-tune
+it from there.
+
+
+Setup Example
+-------------
+
+.. _weka-prerequisites:
+
+**Prerequisites:**
+
+- A Machine with at least one GPU. You can adjust the max model length of your vllm instance depending on your GPU memory.
+
+- Weka already installed and mounted.
+
+- vllm and lmcache installed (:doc:`Installation Guide <../getting_started/installation>`)
+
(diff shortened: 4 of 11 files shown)


lmcache/experimental/cache_engine.py
```diff
@@ -726,6 +726,11 @@ def _Create_memory_allocator(
             assert config.nixl_buffer_device is not None
             return AdHocMemoryAllocator(config.nixl_buffer_device)
 
+        if config.remote_url is not None and config.remote_url.startswith(
```
YaoJiayi: Why is this implemented as a remote backend? If so, should we but the integration under the connector folder?

lmcache/experimental/config.py
```diff
@@ -95,6 +95,10 @@ class LMCacheEngineConfig:
     # The url of the actual remote lmcache instance for auditing
     audit_actual_remote_url: Optional[str] = None
 
+    # (Optional) Weka/CuFile related configurations
+    # Size of CuFile Buffer in MiB
+    cufile_buffer_size: Optional[int] = None
```
YaoJiayi: A high-level question: Can the code run without weka hardware? If so, does using CuFile have performance improvement over normal file read?

lmcache/experimental/cache_engine.py
YaoJiayi: Currently, all remote backends are put under the connector folder

lmcache/experimental/storage_backend/__init__.py
```diff
@@ -71,7 +73,14 @@ def CreateStorageBackends(
         backend_name = str(local_disk_backend)
         storage_backends[backend_name] = local_disk_backend
 
-    if config.remote_url is not None:
+    if config.remote_url is not None and config.remote_url.startswith(
+            "weka://"):
+        weka_backend = WekaGdsBackend(config, loop, memory_allocator,
+                                      dst_device)
+        # TODO(Serapheim): there's a chance we don't want the local
+        # CPU cache in front of ours. Change
```
YaoJiayi: Agreed. The current design is very unfriendly to GPUDirect. Will try to resolve this in the future.

lmcache/experimental/storage_backend/weka_gds_backend.py
```diff
@@ -0,0 +1,460 @@
...
+    ) -> None:
+        """
+        Convert KV to bytes and async store bytes to disk.
+        """
+        kv_chunk = memory_obj.tensor
+        assert kv_chunk is not None
+        path = self._key_to_path(key)
+
+        fl = key.chunk_hash[:2]
+        sl = key.chunk_hash[2:4]
```
YaoJiayi: What are f1 and s1 here? Maybe adding some comments?

YaoJiayi (commented):
Thanks! @sdimitro The code looks goo to me in general:) Left some comments

lmcache/experimental/config.py
sdimitro: Great questions! The relevant piece of information here is that CuFIle will use GPUDirect under the covers if the underlying system supports it (see compat_mode here: https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html) - so to answer your question below

> Can the code run without weka hardware?

Yes, you don't need a Weka Filesystem for this to run. It will work with any type of filesystem. If you see I added a test case and it works regardless of whether Weka is installed or not. The thing is that if the underlying filesystem is not Weka then (1) it will use normal POSIX I/O instead of GPUDurect and (2) the performance won't be the same even if GPUDirect is enabled because in our design we take advantage of the fact that Weka can handle many small files in a single directory without a problem, especially with 4K alignment (i.e if it was another filesystem it could make sense to bundle the data together in less files and potentially use a different block/buffer size for reads/writes).

> If so, does using CuFile have performance improvement over normal file read?

It does because it uses GPUDirect under the covers. Initially we were planning on publishing a library that exposes a client API to hook into our GPUDirect but given that CuFile has the logic to fallback on POSIX I/O we thought it would make upstreaming easier.

Let me know if that answers your questions

lmcache/experimental/storage_backend/weka_gds_backend.py
yanok: First and Second Layer resp ;)

sdimitro: Lol yes I agree with Jiayi that this is not readable.

I'll write a block comment at the top explaining that our directory structure for our metadata has two levels and why, then I'll rename these variables all around the file to first_level and second_level.

lmcache/experimental/cache_engine.py
sdimitro: Per our conversation - I got rid of the use of `remote_url` and now there is a `weka_path` configuration field 👍

lmcache/experimental/storage_backend/weka_gds_backend.py
sdimitro: Updated - resolving this

lmcache/experimental/storage_backend/__init__.py
sdimitro: We can help tackling this when we see it getting too much in the way!

lmcache/experimental/config.py
sdimitro: Added the above notes as comments in the class block comment - feel free to open more issues in the PR if you have subsequent question or something doesn't make sense :)

lmcache/experimental/storage_backend/weka_gds_backend.py
```diff
@@ -0,0 +1,480 @@
...
+    def __str__(self):
+        return self.__class__.__name__
+
+    def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
+        # TODO(Serapheim): implement pin() semantics
+        with self.hot_lock:
+            res = key in self.hot_cache
+        if res:
+            return True
+        if self._try_to_read_metadata(key):
```
YaoJiayi: When will metadata be on disk?

lmcache/experimental/storage_backend/weka_gds_backend.py
```diff
@@ -0,0 +1,480 @@
...
+        """
+        kv_chunk = memory_obj.tensor
+        assert kv_chunk is not None
+        path = self._key_to_path(key)
+
+        l1_dir = key.chunk_hash[:2]
+        l2_dir = key.chunk_hash[2:4]
+        subdir_key = l1_dir + l2_dir
+        if subdir_key not in self.metadata_dirs:
+            os.makedirs(os.path.join(self.weka_path, l1_dir, l2_dir),
```
YaoJiayi: Why not use `_key_to_path` here?

lmcache/experimental/storage_backend/weka_gds_backend.py
```diff
@@ -0,0 +1,480 @@
...
+    else:
+        addr = ctypes.c_void_p(base_pointer)
+        dev_offset = device_offset
+    tmp_path = path + tmp
+    offset = _METADATA_MAX_SIZE
+    metadata = pack_metadata(kv_chunk.shape, kv_chunk.dtype, kv_chunk.nbytes)
+    try:
+        with open(tmp_path, "wb") as f:
+            f.write(metadata)
+        with cufile.CuFile(tmp_path, "r+") as f:
```
YaoJiayi: Are these two writes blocking functions?

YaoJiayi: IIUC, we need to use `aiofiles` want to make them async? Or is `cufile` non-blocking by itself?

lmcache/experimental/storage_backend/weka_gds_backend.py
```diff
@@ -0,0 +1,480 @@
...
+        tmp = '.tmp' + rand_suffix(self.rand, 8)
+        metadata = save_gds_cufile(path, tmp, kv_chunk,
+                                   self.cufile_base_pointer,
+                                   memory_obj.metadata.address)
+
+        self.insert_key(key, memory_obj)
+        memory_obj.ref_count_down()
+
+        task = asyncio.create_task(
+            save_metadata(path + _METADATA_FILE_SUFFIX, tmp, metadata))
```
YaoJiayi: Why are we saving metadata on disk? Besides, if we want to persist the data when the engine is down, will the metadata be enough?

YaoJiayi (commented):
Left some new comments:)

lmcache/experimental/storage_backend/weka_gds_backend.py
sdimitro: We actually do use `_key_to_path` in line 346 right above from your highlighted snippet, but it returns the whole path to the key. In the highlighted code we create `subdir_key` (which is very similar but it's not a path, just a plain string) and use it as the key to the `self.metadata_dirs`.

Admittedly, there's some duplicate code and I have cleaned it up in a private branch but wanted to get this in first as my other changes are bunching a together different changes. I'll take some time to cherry-pick them and update this.

lmcache/experimental/storage_backend/weka_gds_backend.py
sdimitro: Yes - I'll change to `aiofiles.open` 👍  Weirdly we didn't hit any hangs with it

lmcache/experimental/storage_backend/weka_gds_backend.py
sdimitro: Every time we write something on disk we write its metadata too. For every write we also add it to the `hot_cache`.

That said given that Weka is a distributed filesystem, some other host/node could have added the key that we are looking for (and thus the key is not in our `hot_cache`) so we need to check on the filesystem with `_try_to_read_metadata`.

lmcache/experimental/storage_backend/weka_gds_backend.py
sdimitro: Because of two reasons:
1. As you said for ourselves to come up quickly and serve keys when the engine restarts
2. For other engines that are sharing the filesystem

lmcache/experimental/storage_backend/weka_gds_backend.py
sdimitro: They are, I just updated the PR and cherry-picked some code that I have into a private branch that wraps this function into a `asyncio.to_thread`.

lmcache/experimental/storage_backend/weka_gds_backend.py
YaoJiayi: Got it:)

lmcache/experimental/storage_backend/weka_gds_backend.py
YaoJiayi: Okay. Then let's do this refactor later.

lmcache/experimental/storage_backend/weka_gds_backend.py
YaoJiayi: Okay. We can optimize the performance in the following PRs.

YaoJiayi:
@sdimitro LGTM in general. Could you resolve the conflicts before I merge it? Thanks!

sdimitro:
@YaoJiayi resolved merge conflicts and retested 👍

YaoJiayi (approved):
LGTM! Let's push for the follow-up improvement PRs!

lmcache/v1/config.py
ApostaC, line 1: Can we also add the config parsing support in `from_env` so that we can configure it via environment variables when the config file is not an available option?

ApostaC (commented):
LGTM in general! Just a minor comment.

ApostaC:
It would also be great if you could add a new doc page under the [docs/source/kv_cache](https://github.com/LMCache/LMCache/tree/dev/docs/source/kv_cache) folder talking about how to use the Weka backend.  The newly added page will be automatically built into the LMCache [doc website](https://docs.lmcache.ai). 

Really appreciate the contribution!

lmcache/v1/config.py
sdimitro: Good catch! I updated the PR with support for this 👍

sdimitro:
Updated PR - In order to test that the docs pages render well I used `sphinx-autobuild`

ApostaC (approved):
LGTM! Thanks!

PR #718, opened 2025-05-27
[performance] reduce the number of calls to remote backend exists rpc

chunxiaozheng:
The key logic of the `store` method in `cache_engine.py` is as follows,  the `contains` method of `storage_manager` has already been called, so there is no need to call it again in the `put` method of `storage_manager`. 
```
                tokens, mask):
            ...
            if self.storage_manager.contains(key):
                continue
            ...

            self.gpu_connector.from_gpu(memory_obj, start, end, **kwargs)
            self.storage_manager.put(key, memory_obj)

            ...
```

The `store_layer` method is the same as above.

### DIFF
--- a/lmcache/v1/storage_backend/storage_manager.py
+++ b/lmcache/v1/storage_backend/storage_manager.py
@@ -150,9 +150,7 @@ class StorageManager:
         # configure caching policies (e.g., write-through,
         # write-back, etc.)
         for storage_backend in self.storage_backends.values():
-            if storage_backend.exists_in_put_tasks(key) or storage_backend.contains(
-                key
-            ):
+            if storage_backend.exists_in_put_tasks(key):
                 memory_obj.ref_count_down()
                 return
 


chunxiaozheng:
hi~ @YaoJiayi could you help take a look? Thanks!

eryugey:
Makes sense, LGTM.

maobaolong:
@chunxiaozheng As this is discussed between us, it LGTM. @YaoJiayi Would you like to take another look?

@chunxiaozheng Would you like to rebase to main branch and resolve the conflict?

chunxiaozheng:
> @chunxiaozheng As this is discussed between us, it LGTM. @YaoJiayi Would you like to take another look?
> 
> @chunxiaozheng Would you like to rebase to main branch and resolve the conflict?

fixed

maobaolong (approved):

YaoJiayi (approved):
LGTM

PR #737, opened 2025-05-29
add username and password for redis connector

### LINKED ISSUE #667: Support username and password auth for Redis backends
I'd like to use LMCache with Redis as a backend. However, the Redis instance I'm trying to use uses username & password authentication and today that causes LMCache to fail when establishing the connection with this error:

```
LMCache WARNING: Failed to initialize/re-establish remote connection: Invalid host definition in URL:
```

I'd like to be able to pass a remove url with the username and password embedded in the url. E.g.
```
LMCACHE_REMOTE_URL: "redis://user:pass@localhost:6379"
```

And have that be passed down to the Redis client

Not sure if there are other alternative to get around this issue. Please let me know if I missed something

**Additional context**

I was able to get this working on my fork by updating [`parse_remote_url`](https://github.com/LMCache/LMCache/blob/98d4108d4c7120aab8cf4b4a579891b74fa822e5/lmcache/storage_backend/connector/__init__.py#L42 ) to extract the username and password and then integrate them into the `CreateConnector` logic. 

If this approach makes sense I'm happy to open a PR and contribute the fix

calvin0327:

FIX https://github.com/LMCache/LMCache/issues/667

### DIFF
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 # Standard
-from dataclasses import dataclass
-from typing import Dict, List, Optional
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Tuple
+from urllib.parse import parse_qs, urlparse
 import asyncio
-import re
 
 # First Party
 from lmcache.logging import init_logger
@@ -44,14 +45,15 @@ logger = init_logger(__name__)
 class ParsedRemoteURL:
     """
     The parsed URL of the format:
-    <connector_type>://<host>:<port>[/path][?query],<host2>:<port2>[/path2][?query2],...
+        <host>:<port>[/path][?query]
     """
 
-    connector_type: str
-    hosts: List[str]
-    ports: List[int]
-    paths: List[str]
-    query_params: List[Dict[str, str]]
+    host: str
+    port: int
+    username: Optional[str] = None
+    password: Optional[str] = None
+    path: Optional[str] = None
+    query_params: Dict[str, List[str]] = field(default_factory=dict)
 
 
 def parse_remote_url(url: str) -> ParsedRemoteURL:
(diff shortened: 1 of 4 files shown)


calvin0327:
@ApostaC There seems to be a problem with unittest. Why does it prompt "page not found" when I click the link?

ApostaC:
@calvin0327 I changed the buildkite permissions. Can you try it again?

lmcache/storage_backend/connector/__init__.py
```diff
@@ -49,19 +51,34 @@ def parse_remote_url(url: str) -> ParsedRemoteURL:
     """
     Parses the remote URL into its constituent parts.
 
+    Args:
+        url: The remote URL to parse. the URL format is as follows:
+            [connector_type]://[username:pass@]host1:port1[,host2:port2,...]
```
walterbm: could we support having different usernames and passwords for each host? would be useful for redis sentinel

calvin0327: It makes sence, it like: [connector_type]://[username:pass@]host1:port1[username:pass@host2:port2,...]

calvin0327: @walterbm It seems that the URL of Redis Sentinel cannot set different passwords for each node (e.g. `redis-sentinel://username:password@sentinel1:26379,sentinel2:26379/?service_name=mymaster&role=slave`). It can only set different passwords for the master and slave nodes. Considering compatibility with other URL types of different connector_type, designing a universal URL representation for this is indeed challenging.

Do you have any suggestions?

relyt0925:
@calvin0327 maybe this should be a separate PR: but I think we also need to be able to support rediss:// protocol beyond just authentication so the traffic is encrypted. Right now looking at this PR: I think it will add user + pass support: but it doesn't look to be able to parse the necessary configuration to enable a protocol with encryption

lmcache/storage_backend/connector/__init__.py
```diff
@@ -41,28 +41,52 @@ class ParsedRemoteURL:
...
+            [connector_type]://[username:password@]host1:port1[,host2:port2,...]
+
+        1. Connector_type(required), Valid values: redis, redis-sentinel, lm.
+        2. Credentials(optional): Includes username and password.
+        3. Host and Port(required): at least one pair required
 
     Raises:
         ValueError: If the URL is invalid.
     """
+    # Match connector type and the rest of the URL
```
relyt0925: couldn't a lot of this be simplified if we just use urlparse and the official redis python client?
```
import redis
import os
from urllib.parse import urlparse

# Example URL
redis_url = os.environ.get("REDIS_URL")

# Parse the URL
url = urlparse(redis_url)

# Create a Redis client
r = redis.Redis(host=url.hostname, port=url.port, password=url.password, ssl=(url.scheme == "rediss"), ssl_cert_reqs=None)
```

relyt0925: otherwise I think to fully cover redis configurations we would need
hostname
port
password
username
and a ssl flag or the protocol is passed and that is detected in the downstream class.

relyt0925: (this is just a question: does not mean this PR needs to change: I am also not a maintainer on the repo just someone who is interested in the functionality you are providing with the additional functionality of being able to support a secure protocol.)

calvin0327:
@ApostaC @relyt0925 @walterbm From this PR, it appears that there are numerous requirements for connecting to Redis. For example, username and password authentication are needed, and when using Redis Sentinel, passwords for both the master and slave nodes must be configured. Additionally, there's a need for a secure protocol. Should we separate the Redis URL configuration from the remote_url setting or create a dedicated class for it, like this:

```python
class RedisConfig:
    urls: List[str]         # host: port
    username: str
    password: str
    slave_username: str
    slave_password: str
    scheme: str
    protocol：str
```

relyt0925:
@calvin0327 I personally think 1 of 2 options is best:
1) We pass the full redis url to the class and the class handles parsing it and creating a client from it
2) We just pass the parameters parsed from the url to the client. The only key gotcha I believe if a user today only specifies `hostname:port` with no protocol prefix like `redis://` or `rediss://`: the code if we want to standardize on the url parsing needs to assume the scheme is `redis://` (tcp and unencrypted) to ensure no breakage

relyt0925:
Redis offers in other languages a `ParseURL` helper that is supposed to help automate the instantiation of everything from a redis url (an example in Go: https://pkg.go.dev/github.com/go-redis/redis#ParseURL). I couldn't find it in Python.

relyt0925:
I'm not sure on sentiel but I would think based off this strategy there is a way to encode all sentiel parameters in the url format as well and redis will ultimately have a way to parse and read that

relyt0925:
Maybe it is:
from_url ?
```
import redis

url = "redis://user:password@localhost:6379/0"
r = redis.Redis.from_url(url)
```

calvin0327:
@relyt0925 @ApostaC @walterbm I found that the previous `parse_remote_url` was difficult to parse different `connector_type` URL value, as their URLs vary widely. Handing these URLs in a single function would make the logic extremely complex and hart to maintain. Therefore, I have abstracted an adaptor for each different `connector_type` to handle the URL parsing of corresponding connector_type specific URLs. This approach can better adapt to various future requirement changes.

ApostaC:
@calvin0327 Got it, will take a look! Thanks!

relyt0925:
@calvin0327 amazing! the only other thing I would call out although we might be trying to keep some abstraction: I think the redis client directly provides a method where you wouldn't even have to do any parsing:
https://redis-py.readthedocs.io/en/stable/connections.html#redis.Redis.from_url

calvin0327:
> @calvin0327 amazing! the only other thing I would call out although we might be trying to keep some abstraction: I think the redis client directly provides a method where you wouldn't even have to do any parsing: https://redis-py.readthedocs.io/en/stable/connections.html#redis.Redis.from_url

@relyt0925  Ok, It is true that redis.Redis.from_url can automatically parse Redis URLs, so we no longer need to parse them manually.

ApostaC (commented):
This is great! The connector adapters are very clear. Can we add some unit tests (at least for redis & redis-sentinel connectors)? Thanks!

calvin0327:
> This is great! The connector adapters are very clear. Can we add some unit tests (at least for redis & redis-sentinel connectors)? Thanks!

@ApostaC Ok，I‘ve added unit tests for redis and redis sentinel first. later, I can create a separate PR to add other types of tests.

relyt0925:
amazing work!

ApostaC (commented):
Great work! LGTM!
I think this can be merged after fixing the tests and the conflicts.
@YaoJiayi PTAL, thanks!

calvin0327:
> Great work! LGTM! I think this can be merged after fixing the tests and the conflicts. @YaoJiayi PTAL, thanks!

@ApostaC @relyt0925 I have solved it，PTAL，thanks

YaoJiayi (approved):
LGTM!

calvin0327:
> LGTM!

Thanks

PR #739, opened 2025-05-29
[Misc] Rename Weka backend to GDS backend

da-x:
There are already several storage vendors supporting GDS, and there is really nothing Weka-specific in this implementation.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -80,7 +80,7 @@ Documentation
    kv_cache/infinistore
    kv_cache/mooncake
    kv_cache/valkey
-   kv_cache/weka
+   kv_cache/gds
 
 :raw-html:`<br />`
 
--- a/docs/source/kv_cache/gds.rst
+++ b/docs/source/kv_cache/gds.rst
@@ -1,17 +1,17 @@
-Weka
+GDS
 ====
 
-.. _weka-overview:
+.. _gds-overview:
 
 Overview
 --------
 
-WekaFS is a high-performance, distributed filesystem and is a supported option for KV Cache offloading in
-LMCache. Even though the local filesystem backend can work with a WekaFS mount, this particular backend is
-optimized for Weka's characteristics. It leverages GPUDirect Storage for I/O and it allows data-sharing
-between multiple LMCache instances.
+Networked, high-performance distributed filesystems are fully supported for KV-cache offloading in LMCache.
+While the local-filesystem backend will work when mounted on any such storage, this backend is specifically
+tuned for network-attached deployments. It takes advantage of GPU-Direct Storage for zero-copy I/O from GPU
+memory to storage systems, and enables multiple LMCache instances to share data seamlessly.
 
-Ways to configure LMCache WEKA Offloading
+Ways to configure LMCache GDS Offloading
 -----------------------------------------
 
 **1. Environment Variables:**
@@ -24,8 +24,8 @@ Ways to configure LMCache WEKA Offloading
     export LMCACHE_USE_EXPERIMENTAL=True
     # 256 Tokens per KV Chunk
     export LMCACHE_CHUNK_SIZE=256
-    # Path to Weka Mount
-    export LMCACHE_WEKA_PATH="/mnt/weka/cache"
+    # Path to GDS Mount
+    export LMCACHE_GDS_PATH="/mnt/gds/cache"
     # CuFile Buffer Size in MiB
     export LMCACHE_CUFILE_BUFFER_SIZE="8192"
     # Disabling CPU RAM offload is sometimes recommended as the
@@ -46,8 +46,8 @@ Example ``config.yaml``:
     chunk_size: 256
     # Disable local CPU
     local_cpu: false
-    # Path to Weka Mount
-    weka_path: "/mnt/weka/cache"
+    # Path to GDS-enabled mount Mount
+    gds_path: "/mnt/gds/cache"
     # CuFile Buffer Size in MiB
     cufile_buffer_size: 8192
 
@@ -64,13 +64,13 @@ it from there.
 Setup Example
 -------------
 
-.. _weka-prerequisites:
+.. _gds-prerequisites:
 
 **Prerequisites:**
 
 - A Machine with at least one GPU. You can adjust the max model length of your vllm instance depending on your GPU memory.
 
-- Weka already installed and mounted.
+- A remote file systems that is mounted and supports GDS.
 
 - vllm and lmcache installed (:doc:`Installation Guide <../getting_started/installation>`)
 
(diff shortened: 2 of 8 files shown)


YaoJiayi (commented):
Thanks @da-x ! The `gds_backend.py` is contributed by weka people, and they are adding weka-specific optimizations to it. Instead of renaming everything, maybe it's a better idea to have a separate file for gds backend?

da-x:
Understandable. Perhaps @YaoJiayi it would make more sense to add these weka-specific optimizations as backend flags to GDS, rather than to duplicate the code? We can have `LMCacheBlackedConfig` as an optional field under `LMCacheEngineConfig`, or direct keys under `LMCacheEngineConfig`.

da-x:
Also, it would be possible for the code to probe the Linux fstype in Backend init to see that it's `wekafs` and to apply the optimizations automatically.

KoyamaSohei:
I am an engineer at a storage vendor other than WEKA, and I would like to note that this patch is applicable to our parallel file system as well.
Of course, I understand the desire to differentiate from competitors, and I fully acknowledge the value of their contribution to GDS support.
However, it is puzzling that they have not disclosed any performance comparison between the local disk backend and the GDS backend.

sdimitro:
Hi folks! Apologies for the late reply I'm the engineer that wrote part of the code in that source file and I was just notified about this.

First of all, thank you for looking at the PR! Secondly, I'd like to make the case for the duplication of code if you can hear me for one minute. The code above may seem generic but it has makes many assumptions on parts that you cannot see - here are some examples:
* In `_async_save_bytes_to_disk` you'll notice that we save the metadata twice - once in the first 4K of the data file and then separately as a metadata file. This may seem random and unnecessary and would potentially degrade performance for some filesystems because we'd do 2 writes instead of 1. So if this source file went to become a generic GDS source file, someone would eventually remove that code because it doesn't make sense. That would hurt Weka though because we've designed logic that expects this metadata file to exist and optimize garbage collection speeds for detecting age of data/files (think LRU/ARC-like policies).
* The way that the directories and data are laid-out is put specifically to optimize fast startup & garbage collection (the code for fast startup is not in this file yet but there is a comment). Other filesystems that prefer more nested hierarchies to avoid having many files in a single directory would probably prefer to make the tradeoff of having a deeper hierarchy and/or bigger files so they can read/write more data at once and improve their throughput.
* The way that cuFile is used in this source file may also seem curious to the detailed reviewer as there are different APIs that can be used, like the async streaming, or a batched one depending on your FS's design. At Weka we made a deliberate effort to use this API to align operations in this file with the underlying FS.

My examples above are to convince you that at first glance, the code may seem generic in the surface but it is not. If you notice, we even do a couple of normal POSIX writes, not just GDS writes - so it doesn't really make sense to call this GDS as a backend because that would be incorrect. In retrospect, I could have spent more time documenting that on my first PR but I was motivated to merge it ASAP so we can start building more on top of it.

@da-x I think it's appropriate to take this code as a starting point and adjust it to a pure GDS backend BTW. I am saying that as an engineer to engineer (no marketing/company politics). I did consider the Weka-specific backend flag option to a pure GDS backend but then the codepaths ended up being completely separate and ended up not worth it. Moreover, we're considering plumbing some Rust code to get rid of some of the Python computational overheads on some paths. I'm more than happy to share more details on the Python overheads for the separate GDS backend if you plan on going down that path - here is an example flamegraph:

profile-256-upstream

You can drill down to `_async_save_bytes_to_disk` and see a bunch of low-hanging fruit for the Python code.

My final point is also have to admit that there will be minor issues with backwards-compatibility. We already have consumers that use `weka_path` in their LMCache cache config, and `weka1` extensions to their cache data paths. Even though, writing a translation layer is easy, actual migrations take time and that would split the attention of our contributions to LMCache forks than the actual upstream. We aim to be upstream first with our LMCache changes and my first PR paved the way to that.

Do the above points make sense @da-x @YaoJiayi ? I'd be more than happy to discuss this more. I can provide more flamegraphs too that could be helpful for a pure GDS backend.

sdimitro:
Apologies BTW I just realized that the flamegraph was not interactive and you can't analyze it well. Here is a link to download it and play with it: https://drive.google.com/file/d/1cvcl6C43Rp4WKOZKOJERhhb0aMOjnF3L/view

YaoJiayi:
@da-x @KoyamaSohei IIUC, the `weka_backend.py` and `non_weka_backend.py` will eventually diverge. So reusing the same code path or putting everthing in the same file might not be a good practice:) So feel free to duplicate or re-implement a general `gds_backend.py` or `non_weka_backend.py`.

@sdimitro Feel free to add more explanation:)

sdimitro:
Sure thing! Here are some steps one could take to approach  a pure `gds_backend.py`:
* You can start with `weka_backend.py`, and substitute `weka` with `gds` to start with
* Remove all the logic around `save_metadata` and `self.save_metadata_tasks` - you already save them once in `save_gds_cufile`, no need to save them twice. If you see Flamegraph above that should give an instant speedup in non-Weka systems.
* You also probably don't need `self.metadata_dir` and the `makedirs()` call - that's also a big bottleneck in the above Flamegraph
* `save_gds_cufile` uses blocking `open()` and a `cuFileWrite` call. You can change the former to have the asyncio/aiofiles version that's non-blocking and make that function truly async. You can also replace `cuFileWrite` and `cuFileRead` (in `load_gds_cufile`) with their respective async versions (https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-stream-apis) or batched ones (https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-batch-apis). That would make these truly async.
* Change the startup code (pre-heating) cache code in __init__ to a simpler structure, other less nested, or whatever else would help you parallelize it more in the future. This part can be very filesystem dependent and maybe different vendors can pass a flag as you initially suggested.

I would say, that the only thing that you probably want to keep from the Weka code in the pure GDS backend is the `CuFileMemoryAllocator` for buffer-preregistration for the GDS calls. It is not part of `weka_backends.py`, it is under `memory_management.py` and used in `cache_engine.py` when we register allocators. You want to make sure that is used instead of the default.

That should be a good start and the code should be a lot cleaner too. Then you can get more Flamegraphs with `py-spy` and drill down more to see what bottlenecks are left. Let me know if you want help with that.

da-x:
I really don't mind either way, if LMCache maintainers are fine with having us do upkeep of these two file-based backends.

This backend already _works really well on non-Weka_ file systems and a good starting point. Thanks @sdimitro I'll keep these suggestions in mind.

@YaoJiayi I'll open a separate PR, please suggest whether we should name it `GdsBackend` or more simply `FileBackend` (because it is easy to have it fallback for local or non-GDS remote file systems).

yanok:
@da-x You are right, technically it's a cuFile backend, and it _will_ work just fine without GDS. It just doesn't make much sense: if someone is going to use POSIX IO and a local filesystem, combination of the remote backend with the fs_connector is more standard and will likely work comparably well.

da-x:
@yanok Now that you mention it, there is also `local_disk`...

IIUC, to recap, for POSIX with varying GDS capabilities, we currently have the following:

- `remote_backend` with `remote_url` config e.g. `fs://<any-string>:<any-number>/<fs-path>` for local/remote file system, not sure if it is efficient enough for GDS-vendors due to lack of GDS.
- `local_disk_backend` with `local_disk` config being the path, and this would also work with remote file systems, possibly more efficient than `remote_backend` approach but it not GDS optimized, so also not efficient for large KV cache destaging.
- `weka_gds_backend` with is GDS optimized and can also work with local / remote and remote + GDS.

What would be the standard POSIX non-Weka option? It seems to me that to have the best of all worlds, and for the easiest user and most intuitive user interface and configuration, the `weka_gds_backend` fork should seek to replace `local_disk` long term and be named just `file`. I'd say long-term because we can first two approaches for layered testing.

yanok:
> @yanok Now that you mention it, there is also `local_disk`...

`local_disk` is like a swap actually. It's a poor man's version of `local_cpu`. There is no persistence or sharing support. So, it's not really a viable option for remote fs. But it's likely a bit faster than remote/fs due to lack of serialization/deserialization.

`remote_backend` + `fs` connector adds persistence for sure, and I think sharing as well (but I didn't check), but a at cost of serialization.

`weka_gds_backend` supports both persistence and sharing, and no serialization. And GDS of course. But if one doesn't have a GDS-capable FS, it relies on cuFile to fall back to POSIX.

da-x:
Thanks @yanok for completing the picture.
@YaoJiayi I'll prepare the PR with whatever name you guys prefer.

guymguym:
Is LMCache going to accept and maintain a separate backend with specific optimizations for every storage vendor?

YaoJiayi:
> Is LMCache going to accept and maintain a separate backend with specific optimizations for every storage vendor?

We can accept the PR. But the team submit the PR should maintain the backend in the future:)

YaoJiayi:
Closing this due to https://github.com/LMCache/LMCache/pull/773

PR #742, opened 2025-05-30
[feature] support set extra config in LMCacheEngineConfig

chunxiaozheng:
# Motivation

Without this PR, there are so many remote connector related configs spring up, the configs getting mess, this PR can keep the generic config clear and make it easy to extends lmcache without introduce any generic config, since the new coming feature related code can get the new non-generic config from `extra config` dynamicaly.

# How to use
Support set extra config in `LMCacheEngineConfig`, the format is as follows:
```
chunk_size: 256
local_cpu: False
extra_config:
  key1: value1
  key2: value2
```
when create remote connector in `CreateConnector` method, we can get `extra_config` from `LMCacheEngineConfig`. 

This PR is just a framework, each connector can refer to this conf to pass parameters later on.


maobaolong:
@chunxiaozheng Maybe you can remove the `connector_` prefix.

chunxiaozheng:
> @chunxiaozheng Maybe you can remove the `connector_` prefix.

done

maobaolong (approved):
@chunxiaozheng Thanks for your contribution! LGTM. @YaoJiayi Would you like to take  a look at this PR?

tests/v1/data/test_config.yaml
```diff
@@ -0,0 +1,5 @@
+chunk_size: 256
```
YaoJiayi, line 1: Remove this file?

YaoJiayi (commented):
LGTM! Just one minor comment

YaoJiayi (approved):
LGTM!

da-x:
Just to make sure, would `extra_config` be the place for storage-backend-specific config options, and not only to `remote_backend` connectors?

maobaolong:
@da-x Yeah, this is why I suggest to rename it from `connector_extra_config` to `extra_config`.

So you can use it for the place which is not belong to the LMCache core.

In word, if you think it doesn't have to be put in the generic config, it can be put in the `extra_config`.

PR #772, opened 2025-06-04
[Bugfix] make cache clean works

flycloud-hz:
LMCache the cache is not cleaned correctly.

In thread buffer_sweeper the clean cache's code does not run, stop_event should be stop_event.is_set().

Thanks.


YaoJiayi (commented):
Current development of LMCache is in `v1/`. Non-v1 code will be deprecated soon.

flycloud-hz:
Now this bug is only present in version v0. And in version v1 after the refactoring this code will be removed? Currently I use the version 0.1.4 in docker image of lmcache-vllm.

PR #740, opened 2025-05-30
[FSConnector] improve performance with asyncio and direct read to memory object

guymguym:
It was observed that fs_connector get() took ~200ms for a 130MB file while the actual filesystem speed is 30ms (with dd).

This PR implements fs_connector get() to use aiofiles (like in local_disk_backend), and also use file.readinto() to read the data directly into allocated memory. To allocate the memory it needs to first read the meta-data section from the beginning of the file, then parse the shape, dtype, fmt, so it can allocate the right amount of memory, and finally read directly into it. The 2 reads are worth it because with these changes get latency decreased from 200ms to 30ms (Python 3.12.3).

In addition, also changed fs_connector's put() and exists() to use aiofiles.

See small test output with time breakdowns of put and get:

```
$ LMCACHE_LOG_LEVEL=debug pytest -s tests/v1/test_connector.py
...
tests/v1/test_connector.py::test_fs_connector[cpu] [2025-05-30 02:05:05,060] LMCache INFO: Initialized FSConnector with base path /tmp/tmph8aximvi/ (fs_connector.py:62:lmcache.v1.storage_backend.connector.fs_connector)
[2025-05-30 02:05:05,061] LMCache INFO: Created connector <lmcache.v1.storage_backend.connector.fs_connector.FSConnector object at 0x71916d5bcd40> for fs (__init__.py:216:lmcache.v1.storage_backend.connector)
[2025-05-30 02:05:07,128] LMCache DEBUG: FSConnector put /tmp/tmph8aximvi/vllm@test_model@3@123@hash.data took 70.739375ms - open=0.462161ms md=0.122279ms write=69.588883ms replace=0.566052ms (fs_connector.py:149:lmcache.v1.storage_backend.connector.fs_connector)
[2025-05-30 02:05:07,129] LMCache DEBUG: Bytes offloaded: 131.0720 MBytes (instrumented_connector.py:51:lmcache.v1.storage_backend.connector.instrumented_connector)
[2025-05-30 02:05:07,161] LMCache DEBUG: FSConnector get /tmp/tmph8aximvi/vllm@test_model@3@123@hash.data took 30.869987ms - open=0.251681ms md=0.186751ms allocate=0.396347ms read=30.035209ms (fs_connector.py:109:lmcache.v1.storage_backend.connector.fs_connector)
[...]

PR #773, opened 2025-06-04
Add a generic GDS backend

da-x:
Following discussion in #739


lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -0,0 +1,535 @@
...
+    torch.uint8,
+    torch.float8_e4m3fn,
+    torch.float8_e5m2,
+]
+dtype_to_idx = {dtype: idx for idx, dtype in enumerate(torch_dtypes)}
+
+
+def get_fstype(path):
+    with open("/proc/mounts", "r") as f:
+        lines = f.readlines()
```
sdimitro, line 73: In Rust I could get away with using `statfs` from libc (because it has access to the struct that the syscall expects). Python unfortunately doesn't seem to have that - vide-coding a translation from Rust would only give this:
```
def get_fs_type(path: str) -> str:
    # Only define the first field of struct statfs that we care about: f_type
    # We need the "rest field" as statfs expects a whole statfs struct.
    class MinimalStatFs(ctypes.Structure):
        _fields_ = [
            ("f_type", ctypes.c_long),
            ("__other_fields__", ctypes.c_long * 15),
        ]
    # Known Linux FS magic numbers
    FS_TYPES = {
        0x9123683E: "BTRFS",
        0xEF53:     "EXT2/3/4",
        0x6969:     "NFS",
        0x5346544e: "NTFS",
        0x01021994: "TMPFS",
        0xFF534D42: "CIFS",
        0x65735546: "FUSE",
        0x4244:     "HFS",
        0x4d44:     "MSDOS",
        0x58465342: "XFS",
        # Add new ones here
    }
    libc = ctypes.CDLL("libc.so.6")
    stat = MinimalStatFs()
    res = libc.statfs(ctypes.c_char_p(path.encode()), ctypes.byref(stat))
    if res != 0:
        raise OSError("Unable to detect fs_type")
    return FS_TYPES.get(stat.f_type, f"UNKNOWN (0x{stat.f_type:X})")
```

Honestly parsing the text file as you do seems more straightforward at this point but I'll leave it up to you - so feel free to drop this 👍

sdimitro (approved):

lmcache/v1/storage_backend/gds_backend.py
da-x: Both interfaces are old and solid, but the string-based one seems to me more future compat. Anyway, if we end up resorting to `statfs` for some odd reason, we can use a package such as [python-statfs](https://github.com/Mortal/python-statfs).

YaoJiayi (approved):
LGTM!

ApostaC:
I will take a look tomorrow. Thanks @da-x

YaoJiayi:
@da-x The unit tests are failing?

da-x:
@YaoJiayi it's good now. Should squash in the fixes?

ApostaC (approved):
LGTM!

guymguym:
@da-x Thanks for pushing this generic GDS backend, but I have to wonder about this approach that keeps all meta-data in memory - why is that good? wouldn't it be much better to read each metadata file on demand from storage? This would allow sharing a large scale GDS storage across instances transparently, and will get rid of the non scalable scan_metadata on startup. This is what is done in fs_connector.py (see also this improvement [PR](https://github.com/LMCache/LMCache/pull/740)), and while this adds a small file read before reading the data file (to be able to determine the memory shape), the kernel will easily cache those small files and the overhead is generally low (depending on the case but I've observed around 0.1 ms). However this will be fully scalable, and will be able to use large scale storage systems shared across many lmcache instances. Thanks

PR #778, opened 2025-06-05
Improve KV cache storage performance and reliability

prashant182:
- Add exception handling around KV cache storage operations in vLLM adapter
- Implement batched put operations in local CPU backend to reduce lock contention
- Optimize storage manager to use batched operations and improve reference counting
- Add proper error logging for storage failures to aid debugging

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -618,19 +618,30 @@ class LMCacheConnectorV1Impl:
                     skip_leading_tokens,
                     request.req_id,
                 )
-                layerwise_storer = self.lmcache_engine.store_layer(
-                    token_ids,
-                    mask=store_mask,
-                    kvcaches=kvcaches,
-                    slot_mapping=slot_mapping,
-                    offset=skip_leading_tokens,
-                )
-                self.layerwise_storers.append(layerwise_storer)
+                try:
+                    layerwise_storer = self.lmcache_engine.store_layer(
+                        token_ids,
+                        mask=store_mask,
+                        kvcaches=kvcaches,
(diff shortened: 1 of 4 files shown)


prashant182:
@ApostaC  mind taking a look?

lmcache/integration/vllm/vllm_v1_adapter.py
ApostaC, line 1: Just wondering what kind of exception could occur when calling those functions? It would be helpful for us to think about whether to log the error or directly panic.

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -171,10 +171,40 @@ def batched_put(
...
+            if hasattr(backend, 'batched_submit_put_tasks'):
+                backend.batched_submit_put_tasks(filtered_keys, filtered_objs)
+            else:
+                # Fall back to individual puts for backends without batched support
+                for key, obj in zip(filtered_keys, filtered_objs, strict=False):
+                    backend.submit_put_task(key, obj)
+        
+        # Decrement reference counts
+        for obj in filtered_objs:
+            obj.ref_count_down()
```
ApostaC, line 208: Not sure if we should do `ref_count_down` here. After calling `submit_put_task`, the object may not be consumed directly, while counting down reference count here will trigger the free of the memory object.

IMO, the backend should call `ref_count_down` after making sure that the put has been finished. 

cc @YaoJiayi

ApostaC (commented):
Thanks for contributing to LMCache! I like the idea of coalescing the put tasks together. 
It would be better if you can provide some quick benchmarking numbers to show the performance benefits of this PR.

prashant182:
@ApostaC  and @YaoJiayi  Added the reference counter update from the storage backend.

PR #804, opened 2025-06-10
[optimize] support don't save unfull chunk

chunxiaozheng:
When dividing tokens into chunks, the last chunk is likely to be unfull. By default, all chunks are saved, the last unfull chunk is useful in PD disaggregation scenarios, but in some scenarios, such as reuse, the unfull chunk is useless.  

In reuse case, wo don't need to save the last unfull chunk, which can help us save space and time.


YaoJiayi (approved):
LGTM!

PR #825, opened 2025-06-12
[Fix] Fix race between test_gds/weka rmtree and asyncio thread loop

sdimitro:
@YaoJiayi let me know about a failure in the weka test: https://buildkite.com/lmcache/lmcache-unittests/builds/2097#0197659a-e33a-4d15-9072-a0384e616e86

The root cause is that we'd rmtree while the thread loop in the Weka backend can be issuing file operations in the same directory. The fix is to ensure that the thread loop is done before running rmtree. As an extra precaution I also made sure to ignore any errors as this is just cleanup and should not bother other tests in the CI.

Since the generic GDS backend is also susceptible I fixed the bug there too.


YaoJiayi (approved):
LGTM! Thanks!

PR #828, opened 2025-06-13
[feature] add exists sync interface

chunxiaozheng:
In our testing, we found that when the hit rate is around 15%, ttft will decrease. Testing has found that the `contains` interface takes up a lot of time, by changing it to synchronized calling, there can be significant optimization.


maobaolong (approved):
+1

YaoJiayi (commented):
Thanks @chunxiaozheng ! Could you give some numbers?

maobaolong:
@YaoJiayi We tested that an exist operation cost is 60~200+ us for mooncake store, but we found it can increase to 3~5ms after the `backend.contains` wrapped, the pain was made by the async cost.

chunxiaozheng:
> Thanks @chunxiaozheng ! Could you give some numbers?

Sure, as introduced by @maobaolong above. In our internal test, when the cache hit is about 16%, the performance of TTFT has decreased, which as follows:
| | ttft |
| ------ | ------ |
| not use lmcache | 6.9s |
| use lmcache | 7.3s |
| use lmcache and with this PR | 6.8s |

Moreover, the `backend.contains` method is synchronous, implementing `exists_sync` for remote connector can simplify the process of asyncio.

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -126,19 +126,7 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
...
-            res = future.result()
-            return res
-        except Exception as e:
-            with self.lock:
-                self.connection = None
-                self.failure_time = time.time()
-            logger.warning(f"Remote connection failed in contains: {e}")
-            logger.warning("Returning False")
-            return False
+        return self.connection.exists_sync(key)
```
YaoJiayi: Why do we remove the error handling here?

YaoJiayi (commented):
Just a minor comment:)

YaoJiayi:
> > Thanks @chunxiaozheng ! Could you give some numbers?
> 
> Sure, as introduced by @maobaolong above. In our internal test, when the cache hit is about 16%, the performance of TTFT has decreased, which as follows:
> 	ttft
> not use lmcache 	6.9s
> use lmcache 	7.3s
> use lmcache and with this PR 	6.8s
> 
> Moreover, the `backend.contains` method is synchronous, implementing `exists_sync` for remote connector can simplify the process of asyncio.

So the overhead of asyncio adds 500ms latency? The numbers look a bit wierd to me

lmcache/v1/storage_backend/remote_backend.py
maobaolong:
```suggestion
        try:
            return self.connection.exists_sync(key)
        except Exception as e:
            with self.lock:
                self.connection = None
                self.failure_time = time.time()
            logger.warning(f"Remote connection failed in contains: {e}")
            logger.warning("Returning False")
            return False
```

xiaguan:
Based on some of my performance tests, it looks like the scheduler in vLLM v1 is quite sensitive to latency in the exist phase—specifically, the time it takes to finish the `get_num_new_matched_tokens` function. If that part runs slowly, it can cause requests to pile up, which then increases TTFT (time to first token) for subsequent requests?

chunxiaozheng:
> > > Thanks @chunxiaozheng ! Could you give some numbers?
> > 
> > 
> > Sure, as introduced by @maobaolong above. In our internal test, when the cache hit is about 16%, the performance of TTFT has decreased, which as follows:
> > ttft
> > not use lmcache 	6.9s
> > use lmcache 	7.3s
> > use lmcache and with this PR 	6.8s
> > Moreover, the `backend.contains` method is synchronous, implementing `exists_sync` for remote connector can simplify the process of asyncio.
> 
> So the overhead of asyncio adds 500ms latency? The numbers look a bit wierd to me

We were also a bit surprised. We use `mooncake` as the remote backend, during the test, 202 prompts were send, with about 2w `contains` were called, all the `contains` cost approximately 80s in total. After changing to synchronized, this problem was significantly alleviated.

lmcache/v1/storage_backend/remote_backend.py
chunxiaozheng: Updated! Thanks.

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -130,12 +130,8 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
                 key.fmt, key.model_name, key.world_size, 0, key.chunk_hash
             )
 
-        future = asyncio.run_coroutine_threadsafe(
```
maobaolong, line 134: @chunxiaozheng Would you like to add an boolean type extra_config as a switch?

chunxiaozheng: Sure, I have updated.

YaoJiayi (approved):
LGTM. @chunxiaozheng @maobaolong Do we know why changing it to sync can alleviate this perf problem?

chunxiaozheng:
> LGTM. @chunxiaozheng @maobaolong Do we know why changing it to sync can alleviate this perf problem?

`exists` should be a very lightweight operation, but we use a event_loop to process `put`, `get` and `exists`, the time of `exists` includes the scheduling time and the execution time, when there are too many `put` requests, the time for the `exists` operation will increase.

PR #849, opened 2025-06-17
[optimize] support timeout when get blocking

chunxiaozheng:
This PR  support setting timeout when `get_blocking` is called, to improve `LMCache` stability.

### DIFF
--- a/lmcache/v1/storage_backend/remote_backend.py
+++ b/lmcache/v1/storage_backend/remote_backend.py
@@ -53,6 +53,7 @@ class RemoteBackend(StorageBackendInterface):
         assert config.remote_url is not None
 
         self.remote_url = config.remote_url
+        self.get_blocking_timeout_secs = config.get_blocking_timeout_secs
 
         self.local_cpu_backend = local_cpu_backend
 
@@ -237,7 +238,7 @@ class RemoteBackend(StorageBackendInterface):
         future = asyncio.run_coroutine_threadsafe(self.connection.get(key), self.loop)
 
         try:
-            memory_obj = future.result()
+            memory_obj = future.result(self.get_blocking_timeout_secs)
         except Exception as e:
             with self.lock:
                 self.connection = None
--- a/lmcache/v1/storage_backend/remote_backend.py
+++ b/lmcache/v1/storage_backend/remote_backend.py
@@ -53,7 +53,7 @@ class RemoteBackend(StorageBackendInterface):
         assert config.remote_url is not None
 
         self.remote_url = config.remote_url
-        self.get_blocking_timeout_secs = config.get_blocking_timeout_secs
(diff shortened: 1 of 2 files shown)


chunxiaozheng:
@maobaolong @YaoJiayi could you help take a look?

YaoJiayi (approved):
LGTM!

maobaolong (approved):
LGTM. This is discussed offline. Thanks for this improvement

guymguym:
Hi, I wonder what is the overall design here - as the original stuck request is not cancelled, wouldn't that timeout just cause leaked futures until the process will OOM crash?

chunxiaozheng:
> Hi, I wonder what is the overall design here - as the original stuck request is not cancelled, wouldn't that timeout just cause leaked futures until the process will OOM crash?

Thanks! I will fix later.

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -237,8 +238,11 @@ def get_blocking(
         future = asyncio.run_coroutine_threadsafe(self.connection.get(key), self.loop)
 
         try:
-            memory_obj = future.result()
+            memory_obj = future.result(self.blocking_timeout_secs)
         except Exception as e:
+            if isinstance(e, TimeoutError):
+                logger.warning("get blocking timeout, trigger cancel the future task")
+                future.cancel()
```
guymguym, line 245: [future.cancel()](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel) only works if the future has not started, so this will not help in most cases where the timeout occurs when a request is stuck. 

What was the specific case that this PR was trying to solve? Because a stuck network request can sometimes be released by closing the socket, but a stuck filesystem request (e.g from NFS) cannot be cancelled. I think that introducing this default timeout will cause any stuck case to eventually OOM kill because of this default. 

Perhaps a more holistic approach would be that every connector will have its own default timeout, and we could also introduce a self.connection.reset() method that will try to close existing network connections to try and recover.

chunxiaozheng: > [future.cancel()](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel) only works if the future has not started, so this will not help in most cases where the timeout occurs when a request is stuck.
> 
> What was the specific case that this PR was trying to solve? Because a stuck network request can sometimes be released by closing the socket, but a stuck filesystem request (e.g from NFS) cannot be cancelled. I think that introducing this default timeout will cause any stuck case to eventually OOM kill because of this default.
> 
> Perhaps a more holistic approach would be that every connector will have its own default timeout, and we could also introduce a self.connection.reset() method that will try to close existing network connections to try and recover.

I know, the reason for setting timeout is to prevent it from being blocked all the time, which will also affect VLLM's inference service.

PR #863, opened 2025-06-18
[RFC] Parallel get to improve TTFT a bit

da-x:
We saw a 10% improvement in TTFT by allowing `blocking_get` to be served from threadpool.

This adds the ability to enable this threadpool and to specify the size of the threadpool in the amount of threads.

I'm not expecting this to be merged as-is... I am putting this here more to open a discussion. 

Some points:

- I am aware there's the `LayerwiseLMCacheEngine` which is using the async APIs for fetching KV cache, so adding a thread pool may seem somewhat redundant. However after some experimentation I see it is either not fully stable or does not improve latency over original `LMCacheEngine`. Even after `LayerwiseLMCacheEngine` would work well, I think this change still has value, because it would allow to do a more robust comparison of the performance difference of the the two engines.
- We likely need to find a better name than the too long `parallel_blocking_get_thread_count`. Please suggest.

### DIFF
--- a/docs/source/kv_cache/storage_backends/gds.rst
+++ b/docs/source/kv_cache/storage_backends/gds.rst
@@ -133,18 +133,17 @@ and then comment out the ``LMCACHE_CONFIG_FILE`` below:
         '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
 
 
-POSIX fallback
---------------
+Further settings
+-----------------
 
-In some cases, libcufile implements its own internal POSIX fallback without `GdsBackend` being aware.
-In others, an error such as `RuntimeError: cuFileHandleRegister failed (cuFile err=5030, cuda_err=0)` may be throwned.
-Thus, backend can be configured to fallback to its own POSIX implementation when the usage of the libcufile APIs is not successful.
+The `extra_config` configuration variable can support the following:
 
-To force `GdsBackend` not use libcufile APIs for any reason, you can override its behavior via `extra_config`,
-e.g:
+- ``use_cufile`` (type: boolean): POSIX fallback. In some cases, NVIDIA `libcufile` implements its own internal POSIX fallback without `GdsBackend` being aware. In others, an error such as `RuntimeError: cuFileHandleRegister failed (cuFile err=5030, cuda_err=0)` may be throwned. Thus, backend can be configured to fallback to its own POSIX implementation when the usage of the libcufile APIs is not successful. Pass `false` here to enable the fallback. Note that under this mode it would still use CUDA APIs to map and do operations the pre-registered GPU memory. **Default**: true
+
+- ``batched_get_threads`` (type: integer): This is the number of threads that will be spawned for a dedicated blocking call threadpool to service backend ``blocking_get`` requests. This can improve latency on some cases. **Default:** none.
+
+Example config (via environment variable):
 
 .. code-block:: yaml
 
-    LMCACHE_EXTRA_CONFIG='{"use_cufile": false}'
-
-Note that under this mode it would still use CUDA APIs to map and do operations the pre-registered GPU memory.
+    LMCACHE_EXTRA_CONFIG='{"batched_get_threads": 16, "use_cufile": false}'
--- a/lmcache/v1/storage_backend/gds_backend.py
+++ b/lmcache/v1/storage_backend/gds_backend.py
@@ -35,6 +35,18 @@ _METADATA_MAX_SIZE = 4096  # reserve 4K for metadata.
 # TODO: It is possible to read this 4KB block without triggering read-ahead by
 # various means.
 
+THREAD_POOL = None
+
+
+def get_thread_pool(max_workers):
+    # Standard
+    from concurrent.futures import ThreadPoolExecutor
+
+    global THREAD_POOL
+    if not THREAD_POOL:
+        THREAD_POOL = ThreadPoolExecutor(max_workers=max_workers)
+    return THREAD_POOL
+
 
 class UnsupportedMetadataVersion(Exception):
     pass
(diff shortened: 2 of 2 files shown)


guymguym:
👍 I/O concurrency is extremely valuable to maximize the throughput of any storage because it hides roundtrip delays, and I agree this is notably missing from LMCache at the moment.

For filesystem I/O - aiofiles already uses a threadpool inside (used by LocalDiskBackend and FSConnector), so if we could make sure that the engine (both layerwise and blockwise) will use async this will be great.

For network I/O - it is typically considered most efficient to use asyncio - however this has a hidden throughput limit because a single thread running an async event loop will eventually be limited by ~3GB/s speed due to the overhead of actual data copying to/from the socket. In some cases, this can be sufficient, because there are multiple separate LMCache instances and each one will add another ~3GB to the total throughput, but when each GPU requires more KV throughput, a single thread will limit the ability to saturate the GPU.

For which backend did you see TTFT improvement?

da-x:
@guymguym I tested `GdsBackend` in cuFile POSIX compat mode with `use_direct_io=True`.

YaoJiayi:
@da-x Could you fix the code quality check first?

xiaguan:
For remote connectors, we’d love to have something like `batch_exist`, `batch_get`, and `batch_put` so we can fully leverage the performance of the underlying C++ network and storage stack.

YaoJiayi:
@guymguym @xiaguan 
I've been working on exposing the "batch" to the underlying backends so that the backend can operate on the batch more efficiently.
For examples:
https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/cache_engine.py#L207
https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/storage_manager.py#L186

guymguym:
@YaoJiayi @xiaguan Using also batched_get/exists would be great and allow the backends optimize with async/threads.

What will be the actual batch size on get - assuming no chunked prefill, will the batch be the full list of input tokens or smaller?

lmcache/v1/cache_engine.py
```diff
@@ -262,12 +274,29 @@ def retrieve(
...
+        # Get the memory objects from the storage backend
+        token_processing = list(self.token_database.process_tokens(tokens, mask))
+        if self.config.parallel_blocking_get_thread_count is not None:
+            futures = []
+            for start, end, key in token_processing:
+                assert isinstance(key, CacheEngineKey)
+                futures.append(
+                    get_thread_pool(
+                        self.config.parallel_blocking_get_thread_count
+                    ).submit(self.storage_manager.get, key)
```
guymguym: @da-x I wonder if it's needed to separate this threadpool from the default executor that asyncio/aiofiles already initialized lazy? see these docs -

awaitable loop.run_in_executor(executor, func, *args)[¶](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor)

> The executor argument should be an [concurrent.futures.Executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) instance. The default executor is used if executor is None. The default executor can be set by [loop.set_default_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor), otherwise, a [concurrent.futures.ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) will be lazy-initialized and used by [run_in_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) if needed.

class concurrent.futures.ThreadPoolExecutor(max_workers=None, thread_name_prefix='', initializer=None, initargs=())[¶](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)

> Changed in version 3.13: Default value of max_workers is changed to min(32, (os.process_cpu_count() or 1) + 4).

guymguym: we might still want to add a general config option to override the default executor number of workers...

YaoJiayi:
@da-x Could you resolve the conflict a bit?

lmcache/v1/cache_engine.py
da-x: > @da-x I wonder if it's needed to separate this threadpool from the default executor that asyncio/aiofiles already initialized lazy? see these docs -

@guymguym 
- The threads created for the threadpool are not connected to the current default loop. They have no loop, unless they decide to create it or bind it specifically. See example program.
- If 'blocking get' would use an existing loop to which we have an handle, say, `self.loop`, then we would have an issue here, but I think `self.loop` is only being used for put and prefetch. You can check this too.
- To avoid adding another config option, maybe we can allow `"default"` value for using the Python default instead of passing a number, in such that the type is `string | number | None`?

```python
import asyncio
from concurrent.futures import ThreadPoolExecutor


async def threaded_f():
    loop = asyncio.get_running_loop()
    # This is a different loop than in `main`.
    print("loop in threaded_f", id(loop))


def has_no_async_loop():
    # NOTE:
    #     asyncio.get_running_loop()
    # Throws: Throws RuntimeError: no running event loop
    asyncio.run(threaded_f())


async def main():
    loop = asyncio.get_running_loop()
    print("loop in main", id(loop))

    futures = []
    tp = ThreadPoolExecutor(max_workers=1)
    for x in range(1):
        futures.append(tp.submit(has_no_async_loop))
    _results = [f.result() for f in futures]


if __name__ == '__main__':
    asyncio.run(main())
```

chunxiaozheng:
hi, @da-x, I test this PR with `FSConnector`, run about 220 prompts, without this PR, the ttft is `2.45s`, and with this PR, the ttft is `1.12s`. This is a very good improvement! But I'm not sure if there will be such a good optimization effect in high concurrency situations.

da-x:
There is a new `batched_get_blocking` backend API that each backend can implement, and I am considering moving this to `GdsBackend`.

@chunxiaozheng what backend were you using?

PR #877, opened 2025-06-19
Fix double invoke ref_count_down issue while using audit connector

maobaolong:
This is a quick fix.

As audit connector is wrapped by `InstrumentedRemoteConnector`, but the `_connector` of `auditConnector` is wrapped by `InstrumentedRemoteConnector` too, so after a `put` operation called, the `ref_count_down ` invoked twice.

### DIFF
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -393,7 +393,8 @@ class AuditConnectorAdapter(ConnectorAdapter):
         connector = CreateConnector(
             real_url, context.loop, context.local_cpu_backend, context.config
         )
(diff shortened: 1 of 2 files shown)


maobaolong:
@YaoJiayi This is a obvious bug which is made by me, please take a look, thanks!

YaoJiayi (commented):
I don't see any `ref_count_down` in the diff?

maobaolong:
@YaoJiayi there are two instrumentconnector wrapped,

one wrap auditconnector, two is wrap real connector, each put finished will call ref_count_down

YaoJiayi (approved):
LGTM

maobaolong:

PR #881, opened 2025-06-20
[bugfix] fix skip_leading_tokens always 0

chunxiaozheng:
# Description
In out internal test, we found that there is hits in lmcache, but `skip_leading_tokens` is always 0.

image
image

By learning code,  `from_new_request` always return `RequestTracker` with `num_saved_tokens=0`,  which result in that no matter if there is token hit or not, `skip_leading_tokens` is always 0.

With this PR, the number of calls to `lookup` will increase(Actually, it's calling `contains`), but with `store_mask[:skip_leading_tokens] = False`,  when processing keys through `process_token`, the previous keys will also be filtered out, which will decrease the number of calls to `contains`. Therefore, I understand that this will have almost no impact on performance.

In addition, with `store_mask[:skip_leading_tokens] = False`, the existing keys have been skipped. Do we still need to call `self.storage_manager.contains(key)` in `store` or `store_layer` of `cache_engine`?

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -640,11 +640,10 @@ class LMCacheConnectorV1Impl:
                 # return 0 if there is no local storage configured.
                 # In this case, we should rely on the slip_leading_tokens in
                 # save_spec to avoid transmit the already saved tokens again.
-                # skip_leading_tokens = max(
-                #    self.lmcache_engine.lookup(token_ids),
-                #    save_spec.skip_leading_tokens,
-                # )
-                skip_leading_tokens = save_spec.skip_leading_tokens
+                skip_leading_tokens = max(
+                    self.lmcache_engine.lookup(token_ids),
+                    save_spec.skip_leading_tokens,
+                )
 
                 if skip_leading_tokens == len(token_ids):
                     continue  # skip this request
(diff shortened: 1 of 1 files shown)


chunxiaozheng:
With https://github.com/LMCache/LMCache/pull/794, could @YaoJiayi  help test in you environment? Thanks!

chunxiaozheng:
@YaoJiayi @ApostaC @maobaolong could you help take a look?

Shaoting-Feng (commented):
If `num_saved_tokens` in `RequestTracker` is always zero, should we update it when calling `update`?

chunxiaozheng:
> If `num_saved_tokens` in `RequestTracker` is always zero, should we update it when calling `update`?

When receive a prefill request, `num_saved_tokens` is set to 0 and updated with the following code.
```
if not skip_save:
        tracker.num_saved_tokens = num_tokens_to_save
```
For prefill requests, it must use `lookup` to find which chunks already exist and then skip them.

The `update` in `RequestTracker` only called when process `scheduler_output.scheduled_cached_reqs`, i think we don't need to update `num_saved_tokens` in `update`.

YaoJiayi:
@chunxiaozheng @Shaoting-Feng See https://github.com/LMCache/LMCache/pull/911

YaoJiayi:
@chunxiaozheng Is this still needed?

PR #887, opened 2025-06-22
[optimize] optimize pin/unpin log level

chunxiaozheng:
Reduce log printing.


maobaolong (approved):
LGTM @diabloneo Thanks for your improvement, this log seems a little verbose.

@YaoJiayi Would you like to take another look?

YaoJiayi (approved):
LGTM! But I feel we should support it soon:)

PR #902, opened 2025-06-25
[Bugfix] fix batched insert in lookup server

chunxiaozheng:
fix batched insert in lookup server.


YaoJiayi (approved):
Good catch! Thanks

PR #895, opened 2025-06-24
[Core] XPYD support

YaoJiayi:
XPYD support.

Examples can be found at:
(1) 1p1d: https://github.com/LMCache/LMCache/blob/3e981844519c24abf21c55e86836ff509be3e8ae/examples/disagg_prefill/1p1d_experimental/README.md
(2) 2p2d: https://github.com/LMCache/LMCache/blob/3e981844519c24abf21c55e86836ff509be3e8ae/examples/disagg_prefill/xpyd_experimental/README.md

Key next steps:
(1) TP + Hetero TP support
(2) Compatibility with prefix caching/cpu offloading
(3) Async and layerwise transfer
(4) Misc perf optimization 
(5) More unit testing

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

To create a new tag for lmcache (Note: `v` prefix is required):
`git tag vx.x.x`
`git push origin vx.x.x` (same version again)

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

In case the workflow fails, delete the tag and try again:
[...]

PR #915, opened 2025-06-26
[optimize] optimize alloc and evict performance by chunk allocator

llc-kc:
Optimize memocy alloc and evict performance by MixedChunkMemoryAllocator.
This can replace MixedMemoryAllocator when alloc and free memory of full chunk each time.


YaoJiayi:
See paged memory allocator in https://github.com/LMCache/LMCache/pull/895

PR #911, opened 2025-06-26
[Bugfix] Wrong token truncating

YaoJiayi:
(1) Wrong token truncating during full hit
(2) A more efficient fix than https://github.com/LMCache/LMCache/pull/881

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

To create a new tag for lmcache (Note: `v` prefix is required):
`git tag vx.x.x`
`git push origin vx.x.x` (same version again)

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

In case the workflow fails, delete the tag and try again:
`git tag -d vx.x.x`
`git push origin :refs/tags/vx.x.x`

For example:
`git tag -d v0.3.0`
`git push origin :refs/tags/v0.3.0`

To create a new release and publish `lmcache` Python package to PyPi:
`git remote add upstream git@github.com:LMCache/LMCache.git`
`gh release create vx.x.x --repo LMCache/LMCache --title "vx.x.x" --notes "<Add description>"`

For example:
[...]

PR #916, opened 2025-06-26
Enabled configuring save_decode_cache

TeenSpirit1107:
- Setting `save_decode_cache` as Ture or False in the config is enabled.
- Code for testing `save_decode_cache` and a README.md is added to `examples/kv_cache_reuse/save_decode_cache`.


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -282,12 +282,18 @@ def from_request_tracker(
         # For save operation: do not save if the following condition is met
         # 1. has already been saved before (num_saved_tokens > 0)
         # 2. number of unsaved tokens is not reached the chunk boundary
+        # 3. if save_decode_cache is False and we are in decode phase (num_saved_tokens > 0)
+        
+        # Skip save if we're in decode phase and save_decode_cache is False
+        config = lmcache_get_config()
```
maobaolong: This call is terrible, it caused `Loading LMCache config file` log for each decode.

image


Suggest to save the lmcache config as the member of `LMCacheConnectorV1Impl`, and pass the LMCacheConnectorV1Impl instance to the `ReqMeta` and  `from_request_tracker` can get lmcache from `LMCacheConnectorV1Impl` instance.

Something like 

```python
    @staticmethod
    def from_request_tracker(
        lmcache_connector : "LMCacheConnectorV1Impl",
        tracker: RequestTracker,
        block_size: int,
        lmcache_chunk_size: int = 256,
        load_spec: Optional[LoadSpec] = None,
        skip_save: bool = False,
        discard_partial_chunks: bool = True,
    ) -> Optional["ReqMeta"]:
....

class LMCacheConnectorV1Impl:
    def __init__(
        self,
        vllm_config: "VllmConfig",
        role: KVConnectorRole,
        parent: KVConnectorBase_V1,
    ):
        ...
        self.config = lmcache_get_config()
```

maobaolong:
@TeenSpirit1107 Thanks for the feature for v1. Left a suggestion.

And I tested this feature just now.

Found that it cannot work.

My test is 

- lmcache.yaml
```yaml
chunk_size: 256
local_device: "cpu"
remote_url: "fs://host:0/disc/data1/baoloongmao/fs_data/dir0"
remote_serde: "naive"
pipelined_backend: False
local_cpu: False
max_local_cpu_size: 10
save_unfull_chunk: False
save_decode_cache: False

```
- test_vllm.py
```python
import argparse
import json
import requests

def read_file(file_path: str) -> str:
    """Read the content of a file."""
    try:
        with open(file_path, 'r', encoding='utf-8') as file:
            return file.read().strip()
    except FileNotFoundError:
        raise SystemExit(f"Error: File '{file_path}' not found.")
    except Exception as e:
        raise SystemExit(f"Error reading file: {e}")

def send_request(content: str, model: str, api_url: str, max_tokens: int, 
                 top_k: int, top_p: float, temperature: float, 
                 repetition_penalty: float) -> None:
    """Send a request to the vLLM API and print the response."""
    headers = {"Content-Type": "application/json"}
    payload = {
        "model": model,
        "messages": [{"role": "user", "content": content}],
        "max_tokens": max_tokens,
        "top_k": top_k,
        "top_p": top_p,
        "temperature": temperature,
        "repetition_penalty": repetition_penalty,
        "stream": False
    }

    try:
        response = requests.post(api_url, headers=headers, json=payload)
        response.raise_for_status()
        result = response.json()
        print(json.dumps(result, indent=2))
    except requests.exceptions.RequestException as e:
        raise SystemExit(f"Request failed: {e}")

def main():
    parser = argparse.ArgumentParser(description="Send file content to vLLM API")
    parser.add_argument("--file", required=True, help="Path to the input file")
    parser.add_argument("--model", default="vllm_cpu_offload", help="Model name")
    parser.add_argument("--host", default="localhost", help="API host")
    parser.add_argument("--port", type=int, default=8000, help="API port")
    parser.add_argument("--max-tokens", type=int, default=1024, help="Max tokens to generate")
    parser.add_argument("--top-k", type=int, default=20, help="Top-K sampling")
    parser.add_argument("--top-p", type=float, default=0.6, help="Top-P sampling")
    parser.add_argument("--temperature", type=float, default=0.7, help="Temperature for sampling")
    parser.add_argument("--repetition-penalty", type=float, default=1.05, help="Repetition penalty")

    args = parser.parse_args()
    api_url = f"http://{args.host}:{args.port}/v1/chat/completions"

    content = read_file(args.file)
    send_request(
        content=content,
        model=args.model,
        api_url=api_url,
        max_tokens=args.max_tokens,
        top_k=args.top_k,
        top_p=args.top_p,
        temperature=args.temperature,
        repetition_penalty=args.repetition_penalty
    )

if __name__ == "__main__":
    main()  
```

- run_test.sh
```shell
python3 test_vllm.py --file qa.txt --host localhost --port 8000 --max-tokens 10000
```

- qa.txt
```
The following is a multiple-choice question with only one correct answer. Please give the correct answer choices directly, conclude the final answer by stating 'Answer: '. Just give the final answer, no extra information please.
You have an interesting drought-resistant cultivar of barley, which, unfortunately, contains an anti-nutritional compound. This compound makes it impossible to use this cultivar for food purposes. After an extensive investigation, you discover the gene responsible for the synthesis of this compound. This gene consists of five exons and four introns. You decide to use old-school approaches and produce a collection of mutants using EMS chemical mutagen. You sequence a target gene in all mutant lines and discover some changes at the beginning of its sequence in the first exon area. Which of the following mutations will most probably let you eliminate the anti-nutritional compound in the cultivar?

Intact gene:
5’-ATGTTTCTCGCTGGTACTTCTGTGGATGAACATATTTATTGTCGT…TGA-3’

Mutant 1:
5’-ATGTTCTACGCTGGTACTTCTGTGGATGAACATATTTATTGTCGC…TGA-3’
Mutant 2:
5’-ATGTTCTAAGCTGGTACTTCTGTGGATGAACATATTTATTGTCGC…TGA-3’
Mutant 3:
5’-ATGTTTTACGCTGGTGTCACTTCTGTGGATGAACATATTTATTGTCGT…TGA-3’
Mutant 4:
5’-ATGTTTTACGCTACTTCTGTGGATGAACATATTTATTGTCGT…TGA-3’<b>
A. Mutant 4
B. Mutant 3
C. Mutant 2
D. Mutant 1

Answer:
```

lmcache/integration/vllm/vllm_v1_adapter.py
TeenSpirit1107: Thank you so much for your insightful suggestions!

You're absolutely right about saving the config — I've now updated the code on this branch to reflect your recommendation and moved the `lmcache_config` into the `LMCacheConnectorV1Impl` as suggested.

I also appreciate you testing the `save_decode_cache` feature and pointing out the issue. I will prioritize fixing it as soon as possible.

Apologies for the trouble caused, and thanks again for your helpful and detailed feedback.

examples/kv_cache_reuse/save_decode_cache/test_save_decode_cache.py
```diff
@@ -0,0 +1,396 @@
...
+    
+    # Cleanup
+    try:
+        os.remove(config_enabled)
+        os.remove(config_disabled)
+    except:
+        pass
+
+if __name__ == "__main__":
+    main()
```
YaoJiayi: Let's drop the e2e test for now?
E2E test support will be in: https://github.com/LMCache/LMCache/issues/920

YaoJiayi (commented):
Left a minor comment

maobaolong:
@TeenSpirit1107 Would you like to fix the style issues?

TeenSpirit1107:
> @TeenSpirit1107 Would you like to fix the style issues?

Sure, I am going to meet with Jiayi soon and discuss relevant issues. Thank you.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -256,6 +256,7 @@ class ReqMeta:
 
     @staticmethod
     def from_request_tracker(
+        lmcache_connector: "LMCacheConnectorV1Impl",
```
yoo-kumaneko, line 259: Why add this parameter? It seems that it is unused.

TeenSpirit1107:
> @TeenSpirit1107 Would you like to fix the style issues?

Thanks a lot for your reminding! If you meant lint, I will remember to fix that in my next commit. :smile:

TeenSpirit1107:
> @TeenSpirit1107 Would you like to fix the style issues?

@maobaolong Hi, I believe the `save_decode_cache` is now successfully implemented. The problem in the previously committed version may be due to its failure to check whether the current phase is a decode phase, before using the `save_decode_cache` attribute. If the current phase is a prefill phase, not a decode phase, then the cache should be saved regardless of the value of `save_decode_cache`. The style issues you previously mentioned has also be fixed.

I have run the test in `examples/kv_cache_reuse/local_backends/offload.py`, as suggested by @YaoJiayi . `save_decode_cache` can be enabled or disabled by modifying this: 

```py
os.environ["LMCACHE_SAVE_DECODE_CACHE"] = "true"
```

When `save_decode_cache=False`, it gives this result, with only one line of log storing KV Cache, with skip_leading_tokens=0, indicating this is during prefill phase, meeting our expectation. 

```log
[32;20m[2025-07-01 04:45:47,915] LMCache INFO:[0m Storing KV cache for 606 out of 606 tokens (skip_leading_tokens=0) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
```

When `True`, it gives the following result. The lines from the second line on display the saving of decode cache.

```log
Processed prompts:   0%|                                         | 0/1 [00:00<?, ?it/s, est. speed input: 0.00 toks/s, output: 0.00 toks/s][32;20m[2025-07-01 04:41:18,461] LMCache INFO:[0m Reqid: 0, Total tokens 606, LMCache hit tokens: 0, need to load: 0 [3m(vllm_v1_adapter.py:845:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,709] LMCache INFO:[0m Storing KV cache for 606 out of 606 tokens (skip_leading_tokens=0) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,735] LMCache INFO:[0m Storing KV cache for 95 out of 607 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,745] LMCache INFO:[0m Storing KV cache for 96 out of 608 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,755] LMCache INFO:[0m Storing KV cache for 97 out of 609 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,764] LMCache INFO:[0m Storing KV cache for 98 out of 610 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,774] LMCache INFO:[0m Storing KV cache for 99 out of 611 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,783] LMCache INFO:[0m Storing KV cache for 100 out of 612 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,793] LMCache INFO:[0m Storing KV cache for 101 out of 613 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,802] LMCache INFO:[0m Storing KV cache for 102 out of 614 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
[32;20m[2025-07-01 04:41:18,812] LMCache INFO:[0m Storing KV cache for 103 out of 615 tokens (skip_leading_tokens=512) for request 0 [3m(vllm_v1_adapter.py:774:lmcache.integration.vllm.vllm_v1_adapter)[0m
```

Thank you again for pointing out an error!

examples/kv_cache_reuse/save_decode_cache/test_save_decode_cache.py
```diff
@@ -0,0 +1,455 @@
...
+    # Cleanup
+    try:
+        os.remove(config_enabled)
+        os.remove(config_disabled)
+    except Exception:
+        pass
+
+
+if __name__ == "__main__":
+    main()
```
YaoJiayi: @TeenSpirit1107 Maybe we don't need this in this PR?

YaoJiayi (commented):
LGTM! Just a minor comment:) @maobaolong Would you mind testing again?

maobaolong:
@YaoJiayi This has been tested yesterday after backport this PR

examples/kv_cache_reuse/save_decode_cache/test_save_decode_cache.py
TeenSpirit1107: @YaoJiayi Yes, you are absolutely right. This has been removed in the latest commit.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -170,18 +170,18 @@ class RequestTracker:
     #        preemption
     allocated_block_ids: list[int]
 
-    # The number of tokens that has been saved
+    # The number of tokens that has been savd
```
chunxiaozheng: why revert https://github.com/LMCache/LMCache/pull/911?

maobaolong: It looks strange, maybe by accident.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -283,12 +289,20 @@ def from_request_tracker(
...
+        save_decode_cache = lmcache_connector.config.save_decode_cache
+
         skip_leading_tokens = tracker.num_saved_tokens
         chunk_boundary = (
             cdiv(tracker.num_saved_tokens + 1, lmcache_chunk_size) * lmcache_chunk_size
         )
         skip_save = skip_save or (
-            tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary
+            tracker.num_saved_tokens > 0
+            and input_token_len < chunk_boundary
```
maobaolong, line 315: Can you explain the logic of these three conditions? @TeenSpirit1107

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -283,12 +289,20 @@ def from_request_tracker(
...
+            # check whether in decode phase
+            and (tracker.is_decode_phase and not save_decode_cache)
```
maobaolong: Maybe this should be `or`? @TeenSpirit1107

maobaolong:
@TeenSpirit1107 Hope this diff can help
```
Subject: [PATCH] Add external_adapter.py
---
Index: lmcache/integration/vllm/vllm_v1_adapter.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lmcache/integration/vllm/vllm_v1_adapter.py b/lmcache/integration/vllm/vllm_v1_adapter.py
--- a/lmcache/integration/vllm/vllm_v1_adapter.py	(revision 97590eb403e8bd857cd1fa688511ede82f86f3a4)
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py	(date 1751469694273)
@@ -176,6 +176,7 @@
     # Multimodal hashes and positions
     mm_hashes: Optional[list[str]] = None
     mm_positions: Optional[list["PlaceholderRange"]] = None
+    is_decode_phase = False
 
     @staticmethod
     def from_new_request(
@@ -240,6 +241,9 @@
             )
             new_block_ids = new_block_ids[0]
+        # When a request is scheduled again,
+        # it means that the request is in decode phase
+        self.is_decode_phase = True
 
 
 @dataclass
@@ -257,6 +261,7 @@
 
     @staticmethod
     def from_request_tracker(
+        lmcache_connector: "LMCacheConnectorV1Impl",
         tracker: RequestTracker,
         block_size: int,
         lmcache_chunk_size: int = 256,
@@ -267,6 +272,7 @@
         """Create the request metadata from a request tracker.
 
         Args:
+            lmcache_connector (LMCacheConnectorV1Impl): the LMCache connector instance.
             tracker (RequestTracker): the request tracker.
             block_size (int): the block size in vLLM.
             lmcache_chunk_size (int): the chunk size for LMCache.
@@ -284,12 +290,20 @@
         # For save operation: do not save if the following condition is met
         # 1. has already been saved before (num_saved_tokens > 0)
         # 2. number of unsaved tokens is not reached the chunk boundary
+        # 3. if save_decode_cache is False and we are in decode phase (num_saved_tokens > 0)
+
+        # Skip save if we're in decode phase and save_decode_cache is False
+        save_decode_cache = lmcache_connector.config.save_decode_cache
+
         skip_leading_tokens = tracker.num_saved_tokens
         chunk_boundary = (
             cdiv(tracker.num_saved_tokens + 1, lmcache_chunk_size) * lmcache_chunk_size
         )
         skip_save = skip_save or (
-            tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary
+            tracker.num_saved_tokens > 0
+            and input_token_len < chunk_boundary
+            # check whether in decode phase
+            or (tracker.is_decode_phase and not save_decode_cache)
         )
 
         if skip_save and load_spec is None:
@@ -392,6 +406,7 @@
         is_tp = vllm_config.parallel_config.tensor_parallel_size > 1
 
         config = lmcache_get_config()
+        self.config = config
         self.layerwise_retrievers = []
         if role == KVConnectorRole.SCHEDULER:
             self.lookup_client = LMCacheLookupClient(role, is_tp, vllm_config)
@@ -932,6 +947,7 @@
             self._request_trackers[request.req_id] = request_tracker
 
+                self,
                 request_tracker,
                 self._block_size,
                 self._lmcache_chunk_size,
@@ -951,6 +967,7 @@
             request_tracker.update(new_token_ids, new_block_ids)
 
+                self,
                 request_tracker,
                 self._block_size,
                 self._lmcache_chunk_size,

```

lmcache/integration/vllm/vllm_v1_adapter.py
TeenSpirit1107: @maobaolong  Thanks for correcting. I realized this should indeed be modified as:

```py
```

By the 3 conditions, I thought you mean this:

```py
        # For save operation: do not save if the following condition is met
        # 1. has already been saved before (num_saved_tokens > 0)
        # 2. number of unsaved tokens is not reached the chunk boundary
        # 3. if save_decode_cache is False in decode phase
```

I understood that, the first two conditions together ensure there is no redundant caching operations. Since requests are cached by chunks, if (2.) the next chunk boundary is not reached, then no new chunks need to be cached in addition to (1.) the ones already cached previously.

Given that the caching operation is not redundant, we consider condition 3, which is we save decode cache when the request is in decode phase and `save_decode_cache` is True. 

But my previous implementation doesn't avoid redundant saving when `(tracker.is_decode_phase and not save_decode_cache)` is True. So, you're totally right that the condition checking should be modified as:

```py
```

And the 3 conditions should be restated as:

```py
        # For save operation: do not save if the following condition is met
        # 1. has already been saved before (num_saved_tokens > 0)
        # 2. number of unsaved tokens is not reached the chunk boundary
        # Or this condition is met
        # 3. if save_decode_cache is False in decode phase
```

Sorry for my misunderstanding, and I truly appreciate your insights.

lmcache/integration/vllm/vllm_v1_adapter.py
TeenSpirit1107: Thanks for pointing out. This is an accident indeed.

TeenSpirit1107:
> @TeenSpirit1107 Hope this diff can help
> 
> ```
> Subject: [PATCH] Add external_adapter.py
> ---
> Index: lmcache/integration/vllm/vllm_v1_adapter.py
> IDEA additional info:
> Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
> <+>UTF-8
> ===================================================================
> diff --git a/lmcache/integration/vllm/vllm_v1_adapter.py b/lmcache/integration/vllm/vllm_v1_adapter.py
> --- a/lmcache/integration/vllm/vllm_v1_adapter.py	(revision 97590eb403e8bd857cd1fa688511ede82f86f3a4)
> +++ b/lmcache/integration/vllm/vllm_v1_adapter.py	(date 1751469694273)
> @@ -176,6 +176,7 @@
>      # Multimodal hashes and positions
>      mm_hashes: Optional[list[str]] = None
>      mm_positions: Optional[list["PlaceholderRange"]] = None
> +    is_decode_phase = False
>  
>      @staticmethod
>      def from_new_request(
> @@ -240,6 +241,9 @@
>              )
>              new_block_ids = new_block_ids[0]
>          self.allocated_block_ids.extend(new_block_ids)
> +        # When a request is scheduled again,
> +        # it means that the request is in decode phase
> +        self.is_decode_phase = True
>  
>  
>  @dataclass
> @@ -257,6 +261,7 @@
>  
>      @staticmethod
>      def from_request_tracker(
> +        lmcache_connector: "LMCacheConnectorV1Impl",
>          tracker: RequestTracker,
>          block_size: int,
>          lmcache_chunk_size: int = 256,
> @@ -267,6 +272,7 @@
>          """Create the request metadata from a request tracker.
>  
>          Args:
> +            lmcache_connector (LMCacheConnectorV1Impl): the LMCache connector instance.
>              tracker (RequestTracker): the request tracker.
>              block_size (int): the block size in vLLM.
>              lmcache_chunk_size (int): the chunk size for LMCache.
> @@ -284,12 +290,20 @@
>          # For save operation: do not save if the following condition is met
>          # 1. has already been saved before (num_saved_tokens > 0)
>          # 2. number of unsaved tokens is not reached the chunk boundary
> +        # 3. if save_decode_cache is False and we are in decode phase (num_saved_tokens > 0)
> +
> +        # Skip save if we're in decode phase and save_decode_cache is False
> +        save_decode_cache = lmcache_connector.config.save_decode_cache
> +
>          skip_leading_tokens = tracker.num_saved_tokens
>          chunk_boundary = (
>              cdiv(tracker.num_saved_tokens + 1, lmcache_chunk_size) * lmcache_chunk_size
>          )
>          skip_save = skip_save or (
> -            tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary
> +            tracker.num_saved_tokens > 0
> +            and input_token_len < chunk_boundary
> +            # check whether in decode phase
> +            or (tracker.is_decode_phase and not save_decode_cache)
>          )
>  
>          if skip_save and load_spec is None:
> @@ -392,6 +406,7 @@
>          is_tp = vllm_config.parallel_config.tensor_parallel_size > 1
>  
>          config = lmcache_get_config()
> +        self.config = config
>          self.layerwise_retrievers = []
>          if role == KVConnectorRole.SCHEDULER:
>              self.lookup_client = LMCacheLookupClient(role, is_tp, vllm_config)
> @@ -932,6 +947,7 @@
>              self._request_trackers[request.req_id] = request_tracker
>  
>              req_meta = ReqMeta.from_request_tracker(
> +                self,
>                  request_tracker,
>                  self._block_size,
>                  self._lmcache_chunk_size,
> @@ -951,6 +967,7 @@
>              request_tracker.update(new_token_ids, new_block_ids)
>  
>              req_meta = ReqMeta.from_request_tracker(
> +                self,
>                  request_tracker,
>                  self._block_size,
>                  self._lmcache_chunk_size,
> ```

@maobaolong Thank you for the suggestion. You are right, and it has been applied now.

lmcache/integration/vllm/vllm_v1_adapter.py
maobaolong: @TeenSpirit1107 But it seems you didn't revert the revert fully.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -532,14 +550,14 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
                     sync = False
                 # NOTE(Jiayi): Perform blending before layerwise prefix caching
                 if self.enable_blending:
-                    # TODO(Jiayi): Need to make prefix caching and blending compatible
                     self.blender.blend(
-                        tokens[:lmcache_cached_tokens],
-                        token_mask[:lmcache_cached_tokens],
+                        tokens[: request.load_spec.lmcache_cached_tokens],
```
maobaolong: revert this unrelated code change

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -873,22 +889,17 @@ def update_state_after_alloc(self, request: "Request", num_external_tokens: int)
...
-                and num_external_tokens
-                == self.load_specs[request.request_id].lmcache_cached_tokens
-                - self.load_specs[request.request_id].vllm_cached_tokens
-            ), (
-                f"Mismatch in number of tokens: {num_external_tokens} vs "
-                f"{self.load_specs[request.request_id].lmcache_cached_tokens} - "
-                f"{self.load_specs[request.request_id].vllm_cached_tokens}"
-                f" for request {request.request_id}"
-            )
+        assert (
```
maobaolong: this change is not related to this PR too

maobaolong (changes requested):
@TeenSpirit1107 Thanks for the updated, but the diff of this PR still look strange, if it is hard to rebase your changes to the head of dev, you can create a new branch based on dev branch and apply your changes to it. Whatever, this PR is useful, thanks for your efforts.

lmcache/integration/vllm/vllm_v1_adapter.py
TeenSpirit1107: Thanks for pointing out. The unrelated changes are made to meet the coding style suggestions indicated by the  code quality check, but some of them are wrong, as the ones indicated by you. I will try rebasing and fix these wrong modifications. Thank you for all your suggestions on this PR. I have learnt a lot from them.

maobaolong:
@TeenSpirit1107 Maybe you need fetch your remote upstream dev branch before rebase.

TeenSpirit1107:
> @TeenSpirit1107 Maybe you need fetch your remote upstream dev branch before rebase.

Thank you, I have adopted the method suggested by you. 
Previously, when updating with remote, I have tried merging upstream dev branch into my branch, and then push my branch to remote, which, I now realize, is not the standard practice. I will remember your suggestion the next time :smile:

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -241,6 +241,10 @@ def update(
             new_block_ids = new_block_ids[0]
         self.allocated_block_ids.extend(new_block_ids)
 
+        # When a request is scheduled again,
+        # it means that the request is in decode phase
```
maobaolong: how about chunked prefill? @TeenSpirit1107

TeenSpirit1107: @maobaolong Thank you. This is indeed an important point to make. In my previous commit, `is_decode_phase` is updated as `True` after the first `update()` call, assuming the function will be called **only once** during prefill phase, overlooking the case of chunked prefill. As a result, parts of the chunked prefill may not be cached.

Therefore, I added a function for determining `is_decode_phase`, instead of judging by `update()`. This function utilizes the facts that:

1. in decode phase, only one new token will be included each time;
2. if the number of  `new_token_ids` is the same as `token_ids`, then it is in prefill phase.

```py
def _update_decode_phase(self, new_token_ids: list[int]) -> None:
    """Update the decode phase based on the new tokens.
    
    Args:
        new_token_ids (list[int]): The new tokens being added to the request.
    """
    # Heuristic: Decode phase usually has only one new token
    if len(new_token_ids) == 1 and len(self.token_ids) > 1:
        self.is_decode_phase = True
    else:
        self.is_decode_phase = False
```

This implementation has a limitation: if a prefill chunk contains only one token, then its corresponding request will be mistakenly marked as in decode phase. 

However, I don't think this will have an impact on deciding `save_skip`. Since the `chunk_size` is typically around 256, a prefill chunk with one token will **not** be cached, even if correctly classified as prefill. 

```py
skip_leading_tokens = tracker.num_saved_tokens
chunk_boundary = (
    cdiv(tracker.num_saved_tokens + 1, lmcache_chunk_size) * lmcache_chunk_size
)
skip_save = skip_save or (
    tracker.num_saved_tokens > 0
    and input_token_len < chunk_boundary
    # check whether in decode phase
)
```

I have not yet figured out a way to deal with this corner case. Unlike `vllm_adapter.py` which uses `StoreStatus(Enum)` to keep track of prefill, decode and other phases, `vllm_v1_adapter.py` does not include this mechanism.

maobaolong:
@TeenSpirit1107 Thanks for your quick react and updated!

Would you like to revert the unrelated code first? So that the diff can be reviewed easily.

TeenSpirit1107:
> @TeenSpirit1107 Thanks for your quick react and updated!
> 
> Would you like to revert the unrelated code first? So that the diff can be reviewed easily.

@maobaolong Sure, and thank you for your patience. 

Now the changes only contains:

1. `is_decode_phase` and `save_decode_cache` related modifications;
2. storing config in a member variable (instead of calling functions when requiring it).

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -309,7 +334,11 @@ def from_request_tracker(
         save_spec = SaveSpec(skip_leading_tokens, not skip_save)
 
         # Calculate the token ids and slot mappings for load and save
-        # OPTIMIZATION: pre-allocate the buffer for token ids and block
+        # OPTIMIZATION: pre-allocate the buffer for token ids
+        # and blocktracker.num_saved_tokens > 0
+        # and input_token_len < chunk_boundary
+        # and not save_decode_cache
```
maobaolong: or

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -240,6 +241,19 @@ def update(
...
+        self._update_decode_phase(new_token_ids)
+
+    def _update_decode_phase(self, new_token_ids: list[int]) -> None:
+        """Update the decode phase based on the new tokens.
+
+        Args:
+            new_token_ids (list[int]): The new tokens being added to the request.
+        """
+        # Heuristic: Decode phase usually has only one new token
+        if len(new_token_ids) == 1 and len(self.token_ids) > 1:
```
maobaolong, line 253: @YaoJiayi @sammshen Would you like to take a look at this line? Thanks

lmcache/integration/vllm/vllm_v1_adapter.py
TeenSpirit1107: Thanks. This wrong comment is removed in the latest version. The latest version also fetch upstream and rebased. Sorry for the late reply -- I was on a flight and was unable to update during the last two days.

PR #921, opened 2025-06-29
[bugfix]: Fix nixl_peer_host and nixl_peer_port to nixl_receiver_host and nixl_receiver_port

### LINKED ISSUE #894: The examples of disagg_prefill's configs nixl_peer_host should be changed to  nixl_receiver_host
In the examples of disagg_prefill, 1p1d/configs lmcache-prefiller-config.yaml and lmcache-decoder-config.yaml,   

the config property nixl_peer_host and nixl_peer_port should be changed to nixl_receiver_host and nixl_receiver_port.

xp1d/configs also should be changed as above.

aztecher:

FIX #894

---

This PR addresses the above issue by applying a simple fix as reported there.

### DIFF
--- a/docs/source/api_reference/configurations.rst
+++ b/docs/source/api_reference/configurations.rst
@@ -158,12 +158,12 @@ Settings for Nixl-based disaggregated prefill functionality.
    * - nixl_role
      - LMCACHE_NIXL_ROLE
      - Nixl role. Values: "sender" or "receiver"
-   * - nixl_peer_host
-     - LMCACHE_NIXL_PEER_HOST
-     - Host of the Nixl peer
-   * - nixl_peer_port
-     - LMCACHE_NIXL_PEER_PORT
-     - Base port of the Nixl peer
+   * - nixl_receiver_host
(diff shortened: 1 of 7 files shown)


aztecher:
I have a couple of follow-up questions related to this PR:

1. Should I also update the documentation accordingly? (like [here](https://github.com/LMCache/LMCache/blob/dev/docs/source/disaggregated_prefill/nixl/1p1d.rst#prefiller-configuration))
If so, would it be acceptable to include those changes in this same PR?

2. I noticed that the implementation still supports an old environment variable related to this change. (like [here](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/config.py#L336-L351))
Should we keep this for backward compatibility, or would it be better to remove it now?

Thanks in advance for your feedback!

YaoJiayi:
@aztecher Thanks!
1. Please updatethe doc too:)
2. Let's keep the backward compat for now.

aztecher:
@YaoJiayi Thanks for your comment!
I've fixed the doc as well.

YaoJiayi (approved):
lgtm

Issue #891, opened 2025-06-23
LookupClient with Remote Backend Causes Latency Bottleneck


xiaguan:
The current implementation shows high latency (tens of milliseconds) that impacts throughput, even when the remote connector uses local data structures like `set()`. 

### Observed Performance Characteristics

My benchmark script runs each test twice - the first run typically experiences cache misses, while the second run benefits from cache hits.

**Baseline Performance (Before Optimization):**

*First Run (Cache Miss Scenario):*
- `get_num_new_matched_tokens` latency: **2.58-655ms** 
- Mean TTFT: **16,457ms**
- Request throughput: **0.86 req/s**
- Total token throughput: **8,669 tok/s**

*Second Run (Cache Hit Scenario):*
- `get_num_new_matched_tokens` latency: **5.60-116ms** 
- Mean TTFT: **1,204ms**
- Request throughput: **6.05 req/s**
- Total token throughput: **61,274 tok/s**

**Sample latency measurements from logs:**
```
# First run - cache miss scenario
INFO 06-23 16:04:16 [scheduler.py:377] get_num_new_matched_tokens took 117.90 ms
INFO 06-23 16:04:20 [scheduler.py:377] get_num_new_matched_tokens took 655.33 ms
INFO 06-23 16:04:23 [scheduler.py:377] get_num_new_matched_tokens took 16.84 ms

# Second run - cache hit scenario
INFO 06-23 16:05:06 [scheduler.py:377] get_num_new_matched_tokens took 15.16 ms
INFO 06-23 16:05:06 [scheduler.py:377] get_num_new_matched_tokens took 35.15 ms
INFO 06-23 16:05:07 [scheduler.py:377] get_num_new_matched_tokens took 116.24 ms
```

### Direct Mooncake Master Integration

We experimented with optimizing the scheduler to directly send RPC batch queries to the Mooncake Master, which helped us bypass some of the communication overhead:

### Results After Our Optimization

**Optimized Performance:**

*First Run (Cache Miss Scenario):*
- `get_num_new_matched_tokens` latency: **1.89-3.60ms** 
- Mean TTFT: **903ms** 
- Request throughput: **0.88 req/s** 
- Total token throughput: **8,871 tok/s** 

*Second Run (Cache Hit Scenario):*
- `get_num_new_matched_tokens` latency: **1.62-2.15ms** 
- Mean TTFT: **903ms** 
- Request throughput: **6.91 req/s** 
- Total token throughput: **69,980 tok/s** 

[...]

PR #936, opened 2025-07-01
perf: optimize get_num_new_matched_tokens latency in vLLM scheduler

xiaguan:
## Summary

This PR optimizes the latency of the `get_num_new_matched_tokens` function in the vLLM scheduler, stabilizing overall latency to around 1-2ms.

It checks the keys using the `batchexist` RPC interface from the `Mooncake master` (which handles global metadata in Mooncake). This way, it can get hit info from the whole cluster, and the latency tends to be more stable too

## Configuration Requirements

**Important**: LMCache must be configured to use **Mooncake-only storage** for this optimization to work correctly. Using other storage backends will result in inaccurate lookup results.

### Required Configuration Example

```yaml
chunk_size: 256
local_device: "cpu"
remote_url: "mooncakestore://127.0.0.1:50051/"
remote_serde: "naive"
pipelined_backend: False
local_cpu: False
max_local_cpu_size: 5
external_lookup_client: "mooncakestore://127.0.0.1:50051"
```


Shaoting-Feng:
Thanks for your PR. Could you provide the performance comparison beween two look up server?

xiaguan:
> Thanks for your PR. Could you provide the performance comparison beween two look up server?

Sure, details in this issue https://github.com/LMCache/LMCache/issues/891.

maobaolong:
@chunxiaozheng Would you like to take a look at this PR?

Shaoting-Feng (approved):
LGTM

xiaguan:
Conflicts resolved. Thanks for your review 👍

xiaguan:
Tested in stable vllm 0.9.2 both LMCacheLookupClient and MooncakeLookupClient.

lmcache/v1/config.py
```diff
@@ -157,6 +161,7 @@ def from_defaults(
         extra_config: Optional[dict] = None,
         save_unfull_chunk: bool = True,
         blocking_timeout_secs: int = 10,
+        mooncake_lookup_client: Optional[str] = None,
```
maobaolong: Can you naming this more abstract so that other backend can implement other `xxx_lookup_client` and can reuse this config?

lmcache/v1/lookup_client/factory.py
```diff
@@ -0,0 +1,109 @@
...
+            is_tp: Whether tensor parallelism is enabled
+            vllm_config: The vLLM configuration
+
+        Returns:
+            A lookup client instance
+        """
+        config = lmcache_get_config()
+
+        # Check if mooncake_lookup_client is configured
+        if config.mooncake_lookup_client is not None:
```
maobaolong: If this is `config.external_lookup_client`, it could be designed to be used by variety of implementations, for example, we can use `mooncakestore://<MASTER_ADDRESS>` or `infinistore://<INFINISTORE_ADDRESS>` as the content of `external_lookup_client`.

This idea is referent of remote backend framework.

maobaolong (commented):
@xiaguan Left a refactor suggestion, otherwise, it look wonderful!

lmcache/v1/lookup_client/factory.py
xiaguan: Thanks for the suggestion — it's been fixed!

lmcache/v1/lookup_client/factory.py
```diff
@@ -0,0 +1,162 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Standard
+from typing import TYPE_CHECKING, Optional
+
+# First Party
+from lmcache.integration.vllm.utils import lmcache_get_config
+from LMCache.lmcache.v1.lookup_client.mooncake_lookup_client import MooncakeLookupClient
```
Shaoting-Feng: No module named 'LMCache'

xiaguan: Thks. Fixed

maobaolong (approved):
LGTM @xiaguan Thanks for this great work!

PR #959, opened 2025-07-02
Bugfix: Pin Unpin notImplementedError

4D0R:
Making Weka backend just return false for pin unpin so that engine isn't stopped with notimplementederror for now. We will implement these methods soon.


ApostaC (approved):
LGTM!

sammshen (approved):

PR #973, opened 2025-07-04
Allow skipping cache saving in decode phase

yoo-kumaneko:
Credit to @TeenSpirit1107

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -101,6 +101,9 @@ class RequestTracker:
     mm_hashes: Optional[list[str]] = None
     mm_positions: Optional[list["PlaceholderRange"]] = None
 
+    # Whether the request is in decode phase
+    is_decode_phase = False
+
     @staticmethod
     def from_new_request(
         new_request: "NewRequestData",
@@ -170,6 +173,12 @@ class RequestTracker:
             raise ValueError(f"Unsupported new_block_ids type {type(new_block_ids)}")
         self.allocated_block_ids.extend(new_block_ids)
 
+        # When a request is scheduled again, and the number of new tokens
+        # is 1 (excluding chunked prefill), the request is in decode phase.
+        # TODO: Need to further exclude the case of chunked prefill with 1 token.
+        if len(new_token_ids) == 1:
+            self.is_decode_phase = True
+
 
 @dataclass
 class ReqMeta:
@@ -198,6 +207,7 @@ class ReqMeta:
         load_spec: Optional[LoadSpec] = None,
         skip_save: bool = False,
         discard_partial_chunks: bool = True,
+        save_decode_cache: bool = False,
     ) -> Optional["ReqMeta"]:
         """Create the request metadata from a request tracker.
 
@@ -208,6 +218,7 @@ class ReqMeta:
             load_spec (Optional[LoadSpec]): the load spec for KV cache loading.
             skip_save (bool): whether to skip the save operation.
             discard_partial_chunks (bool): whether to discard partial chunks.
+            save_decode_cache (bool): whether to save the cache in decode phase.
 
         Returns:
             the request metadata if we need to perform load/save
@@ -223,12 +234,16 @@ class ReqMeta:
         # For save operation: do not save if the following condition is met
         # 1. has already been saved before (num_saved_tokens > 0)
         # 2. number of unsaved tokens is not reached the chunk boundary
+        # 3. if save_decode_cache is False and it is in decode phase
+
         skip_leading_tokens = tracker.num_saved_tokens
         chunk_boundary = (
             cdiv(tracker.num_saved_tokens + 1, lmcache_chunk_size) * lmcache_chunk_size
         )
-        skip_save = skip_save or (
-            tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary
+        skip_save = (
+            skip_save
+            or (tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary)
+            or (tracker.is_decode_phase and not save_decode_cache)
         )
 
         if skip_save and load_spec is None:
@@ -389,6 +404,7 @@ class LMCacheConnectorV1Impl:
         )
 
         self._lmcache_chunk_size = config.chunk_size
+        self._save_decode_cache = config.save_decode_cache
 
         self.skip_last_n_tokens = vllm_config.kv_transfer_config.get_from_extra_config(
             "skip_last_n_tokens", 0
@@ -933,6 +949,7 @@ class LMCacheConnectorV1Impl:
                 load_spec=load_spec,
                 skip_save=force_skip_save,
                 discard_partial_chunks=self._discard_partial_chunks,
+                save_decode_cache=self._save_decode_cache,
             )
             if req_meta is not None:
                 meta.add_request(req_meta)
@@ -983,6 +1000,7 @@ class LMCacheConnectorV1Impl:
                 load_spec=None,
                 skip_save=force_skip_save,
                 discard_partial_chunks=self._discard_partial_chunks,
+                save_decode_cache=self._save_decode_cache,
             )
             if req_meta is not None:
                 meta.add_request(req_meta)


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -241,6 +244,11 @@ def update(
             new_block_ids = new_block_ids[0]
         self.allocated_block_ids.extend(new_block_ids)
 
+        # When a request is scheduled again, and the number of new tokens is 1 (exluding chunked prefill)
```
maobaolong: You can mark this as TODO or FIXME comment

maobaolong (commented):
Left a minor comment @yoo-kumaneko  otherwise, it looks good. @YaoJiayi @ApostaC Would you like to take another look?

lmcache/integration/vllm/vllm_v1_adapter.py
yoo-kumaneko: done

yoo-kumaneko:
This is a cleaner implementation. Needs reviewing.

ApostaC:
Will take a look today

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -284,12 +295,19 @@ def from_request_tracker(
         # For save operation: do not save if the following condition is met
         # 1. has already been saved before (num_saved_tokens > 0)
         # 2. number of unsaved tokens is not reached the chunk boundary
+        # 3. if save_decode_cache is False in decode phase
+
+        # Skip save if we're in decode phase and save_decode_cache is False
+        save_decode_cache = lmcache_connector.config.save_decode_cache
```
ApostaC: Since we are introducing `save_decode_cache` back, one quick check: is the default config for save_decode_cache True or False?

My proposal is make it `True` by default and users can turn it off explicitly. What do you guys think? @yoo-kumaneko @maobaolong @YaoJiayi

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -257,6 +266,7 @@ class ReqMeta:
 
     @staticmethod
     def from_request_tracker(
+        lmcache_connector: "LMCacheConnectorV1Impl",
```
ApostaC: I think it's a bit heavy to pass the whole `lmcache_connector` in. And the code assumes the existence of `lmcache_connector.config.save_decode_cache`. 

Would be better if we only pass in a flag of `save_decode_cache`.

ApostaC (commented):
LGTM in general. Just some minor comments.

Also, would be great to share some experiment results to make sure it works. We have introduced the vllm integration test into our CI (#1000), so maybe it's a good place to add some basic/quick test there.

lmcache/integration/vllm/vllm_v1_adapter.py
maobaolong: @ApostaC I'm not sure whether `True` or `False` is the better default value, but there are some consideration items:
- The original behavior of v0 is `False`, it means we do not support save decoded code by default before. 
- There are more cost if we save decoded kvcaches, and maybe the decoded kvcaches can not be reused for some scenarios.

yoo-kumaneko: @ApostaC  If the current default behavior is True, let's keep it this way. We don't want to surprise users with inconsistent behavior.

lmcache/integration/vllm/vllm_v1_adapter.py
yoo-kumaneko: Makes sense. Will do.

lmcache/integration/vllm/vllm_v1_adapter.py
ApostaC: Yeah, I think the current behavior is `True`, and that's the reason of my proposal.

yoo-kumaneko: @ApostaC Got it done. The default value is now True in `lmcache/v1/config.py`, while the default value in `lmcache/config.py` stays unchanged (which is False). Do you think we should modify `lmcache/config.py`?

ApostaC: We don't have to modify `lmcache/config.py`. That code is deprecated and we should remove it soon.

yoo-kumaneko: @ApostaC Could you give me some hints on adding a test to the CI? It seems like the current CI test is only a basic functional test. Should I add the save_decode_cache argument to the config and check that it doesn’t crash, or is there something else I should focus on?

ApostaC: I think #1000 introduced end-to-end CI test with vLLM. 
@Shaoting-Feng Can you give some instructions about how to add a new vLLM integration test? Thanks!

Shaoting-Feng: You can add the vLLM and LMCache configurations to `scripts/vllm-integration-tests.sh` and start a new test.

yoo-kumaneko: I've added a new test to the `vllm-integration-tests.sh` file. Could you please review it and let me know if it looks good?

maobaolong:
@ApostaC As discussed offline, the existing save_decode_cache maybe cannot be reused, besides, the `save_decode_cache` is not the expected behavior of `v1`, do you think it would be a little better if we make it default to False?

ApostaC:
> @ApostaC As discussed offline, the existing save_decode_cache maybe cannot be reused, besides, the `save_decode_cache` is not the expected behavior of `v1`, do you think it would be a little better if we make it default to False?

Sounds good to me!

Also, is this PR ready for review again?

yoo-kumaneko:
> > @ApostaC As discussed offline, the existing save_decode_cache maybe cannot be reused, besides, the `save_decode_cache` is not the expected behavior of `v1`, do you think it would be a little better if we make it default to False?
> 
> Sounds good to me!
> 
> Also, is this PR ready for review again?

Yes. I've changed the default option for `save_decode_cache` to False, and added a CI test for it. It's now ready for review again.

.buildkite/scripts/vllm-integration-tests.sh
```diff
@@ -29,6 +29,15 @@ SERVER_WAIT_TIMEOUT=180
 # UTILITIES #
 #############
 
+cleanup_container() {
```
Shaoting-Feng: Is it possible to utilize `.buildkite/scripts/docker-cleanup.sh`?

.buildkite/scripts/vllm-integration-tests.sh
```diff
@@ -113,6 +122,72 @@ run_lmcache_vllmopenai_container() {
...
+            --env "LMCACHE_SAVE_DECODE_CACHE=True" \
+            --volume ~/.cache/huggingface:/root/.cache/huggingface \
+            --network host \
+            'lmcache/vllm-openai:build-latest' \
+            'meta-llama/Llama-3.2-1B-Instruct' --kv-transfer-config \
+            '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}' \
+            --gpu-memory-utilization '0.5' \
+            --enforce-eager)
+    fi
+    buildkite-agent meta-data set "docker-CID" "$CID"
```
Shaoting-Feng: This may overwrite the `docker-CID` meta-data.

.buildkite/scripts/vllm-integration-tests.sh
```diff
@@ -150,6 +225,32 @@ test_vllmopenai_server_with_lmcache_integrated() {
...
+    )
+
+    if [ "$http_status_code" -ne 200 ]; then
+        echo "Model prompt request from OpenAI API server with save decode cache failed, HTTP status code: ${http_status_code}."
+        cat response-file-save-decode.txt
+        docker logs -n 20 $CID
+        cleanup 1
+        exit 1
+    else
+         echo "Model prompt request from OpenAI API server with save decode cache succeeded"
```
Shaoting-Feng: This still doesn’t verify whether saving the decode cache was executed. But I do not have any good idea in mind.

yoo-kumaneko: Me neither. It seems that the main challenge here is the lack of test infrastructure. The current tests we have are limited to launch tests, which don’t cover function-level verification. Would it be possible to raise a separate issue specifically for adding tests, so we can keep this PR focused and move forward? I’ve tested this locally, and it works as expected.

maobaolong (approved):
@yoo-kumaneko Sorry for ignoring this PR the past weeks, it looks good to me, and I have used this for my single node test already.

.buildkite/scripts/vllm-integration-tests.sh
Shaoting-Feng: > Would it be possible to raise a separate issue specifically for adding tests

That makes sense. Thank you for your PR!

Shaoting-Feng (approved):

maobaolong:
@TeenSpirit1107 Thank you very much for your great effort on https://github.com/LMCache/LMCache/pull/916 , and thanks @yoo-kumaneko for the follow-up work by this PR.

Merged it.

PR #974, opened 2025-07-04
Remove redundant and incorrect truncation of the last n tokens.

yoo-kumaneko:
It errors if we don't delete these lines.

At line 314 in vllm_v1_adapter.py, we truncate the tokens.

[token_ids = torch.tensor(input_token_ids)[:num_tokens_to_save]](https://github.com/LMCache/LMCache/blob/f62585dea788e8c4f8bc98b7183edbe2afba5701/lmcache/integration/vllm/vllm_v1_adapter.py#L314)

Suppose the input length=300, chunk size=256, skip_last_n_tokens=3. If we discard partial chunks, we will truncate the tokens tensor to 256. Then, if we truncate the last 3 tokens again before retrieving, we will be retrieving 253 tokens, which is incorrect.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -521,10 +521,6 @@ class LMCacheConnectorV1Impl:
             )
             token_mask[:masked_token_count] = False
 
-            if self.skip_last_n_tokens > 0:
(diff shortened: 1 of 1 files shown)


maobaolong (approved):
@yoo-kumaneko Thanks for your fix, it LGTM @YaoJiayi Would you like to take another look?

YaoJiayi (approved):

PR #976, opened 2025-07-04
[ADAPT-VLLM] Support vllm 0.8.5 lmcache connector

maobaolong:
vLLM 0.8.5 cannot use LMCache connector https://github.com/LMCache/LMCache/blob/dev/lmcache/integration/vllm/lmcache_connector_v1.py,

Since there are many api changes for the newest lmcache.

The state of vLLM 0.8.5 api of `KVConnectorBase_V1` is https://github.com/vllm-project/vllm/blob/v0.8.5/vllm/distributed/kv_transfer/kv_connector/v1/base.py.

After this PR. we can let vllm v0.8.5 use Lmcache connector by the following config

```bash
LMCACHE_USE_EXPERIMENTAL=True LMCACHE_TRACK_USAGE=false \
VLLM_MLA_DISABLE=0 VLLM_USE_V1=1 LMCACHE_CONFIG_FILE=./lmcache.yaml \
vllm serve /disc/data1/deepseek/DeepSeek-V2-Lite-Chat/ \
           --trust-remote-code \
           --served-model-name vllm_cpu_offload \
           --max-model-len 32768 \
           --max-seq-len-to-capture 10000 \
           --max-num-seqs 64 \
           --host 0.0.0.0 \
           -tp 1 \
           --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1Dynamic","kv_role":"kv_both","kv_connector_module_path":"lmcache.integration.vllm.lmcache_connector_v1_085"}'
```

This is a way to support variety of vLLM version by specified relevant `kv_connector_module_path`.

### DIFF
--- a/lmcache/integration/vllm/lmcache_connector_v1_085.py
+++ b/lmcache/integration/vllm/lmcache_connector_v1_085.py
@@ -1,0 +1,142 @@
+# SPDX-License-Identifier: Apache-2.0
+# Standard
+from typing import TYPE_CHECKING
+
+# Third Party
+from vllm.config import VllmConfig
+from vllm.distributed.kv_transfer.kv_connector.v1.base import (
+    KVConnectorBase_V1,
+    KVConnectorMetadata,
+    KVConnectorRole,
+)
+from vllm.logger import init_logger
+from vllm.v1.core.sched.output import SchedulerOutput
+import torch
+
+# First Party
+from lmcache.integration.vllm.vllm_v1_adapter import LMCacheConnectorV1Impl
+
+if TYPE_CHECKING:
+    # Third Party
+    from vllm.attention.backends.abstract import AttentionMetadata
+    from vllm.forward_context import ForwardContext
+    from vllm.v1.request import Request
+
+logger = init_logger(__name__)
+
+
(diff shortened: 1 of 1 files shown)


maobaolong:
@ApostaC Would you like to take a look at this PR? Thanks.

YaoJiayi:
@maobaolong Why do we need this? Why not simply have a table that matches the vllm and lmcache version?

maobaolong:
@YaoJiayi Thanks for your reply, for the released vllm version, it could find a `lmcache_connector.py` inside `vllm`, just like vllm-0.8.5, but unluckily, some company like ours forked it just before `vllm-0.8.5` release, it didn't contains v1 version `lmcache_connector.py`. So if we maintains `lmcache_connector.py` inside `lmcache` whl, so we could use lmcache with the relevant version `lmcache_connector.py`.

ApostaC:
This makes sense to me! Maybe we need to have separate page in the documentation to talk about the compatibility @maobaolong

ApostaC (approved):
LGTM!

YaoJiayi (approved):

PR #990, opened 2025-07-07
[example] Run 1P1D with K8S LWS

panpan0000:
Provide a Kubernetes example for 1P1D

I've another code with sts, but too many duplicated code. so I choose LWS instead.

the xPxD will require subgroup from LWS. I will raise PR later.

This example will NOT limit running P/D in the same machine , but can run in diff nodes .

### DIFF
--- a/examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
+++ b/examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
@@ -1,0 +1,89 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
+kind: LeaderWorkerSet
+metadata:
+  name: vllm-pd-lws
+spec:
+  replicas: 1
+  leaderWorkerTemplate:
+    size: 3  # leader + 2 workers
+    restartPolicy: RecreateGroupOnPodRestart
+    leaderTemplate:
+      metadata:
+        labels:
+          role: leader
+          app: pd-proxy
+      spec:
+        containers:
+        - name: pd-proxy
+          image: lmcache/vllm-openai:latest
+          # Pod-0 is the proxy
+          command: ["/bin/bash", "-c"]
+          args:
+          - |
+            source /opt/venv/bin/activate
+            python3 /workspace/LMCache/examples/disagg_prefill/1p1d/disagg_proxy_server_first_token_from_decoder.py \
+              --host 0.0.0.0 \
+              --port 9000 \
+              --decoder-host  vllm-pd-lws-0-1.vllm-pd-lws \
+              --decoder-port 8200 \
+              --prefiller-host vllm-pd-lws-0-2.vllm-pd-lws \
+              --prefiller-port 8200
+          ports:
+          - containerPort: 9000
+    workerTemplate:
+      metadata:
+        labels:
+          role: worker
+          app: vllm-pd-instance
+      spec:
+        containers:
+        - name: vllm-pd-instance
+          image: lmcache/vllm-openai:latest
+          env:
+          - name: MODEL
+            value: "Qwen/Qwen2.5-0.5B"
+          - name: PORT
+            value: "8200"
+          # Pod-0-1 is the Decoder
+          # Pod-0-2 is the Prefiller
+          command: ["/bin/bash", "-c"]
+          args:
+          - |
+            source /opt/venv/bin/activate
+            # Processing config files
+            decoder_svc="vllm-pd-lws-0-1.vllm-pd-lws" # worker-1 svc created by LWS
+            cfg_path="/workspace/LMCache/examples/disagg_prefill/1p1d/configs"
+            prefill_conf="${cfg_path}/lmcache-prefiller-config.yaml"
+            decoder_conf="${cfg_path}/lmcache-decoder-config.yaml"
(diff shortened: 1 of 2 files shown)


panpan0000:
@nicole-lihui can you help to review my k8s/lws part yaml ?

nicole-lihui:
Sure thing! 🚀

examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
```diff
@@ -0,0 +1,85 @@
...
+          image: lmcache/vllm-openai:latest
+          # Pod-0 is the proxy
+          command: ["/bin/bash", "-c"]
+          args:
+          - |
+            source /opt/venv/bin/activate
+            python3 /workspace/LMCache/examples/disagg_prefill/1p1d/disagg_proxy_server_first_token_from_decoder.py \
+              --host 0.0.0.0 \
+              --port 9000 \
+              --decoder-host  vllm-pd-lws-0-1.vllm-pd-lws \
```
nicole-lihui, line 27: **My guess is that the current LWS controller is intended to auto-create a headless Service. However, `networkConfig.subdomainPolicy` only supports `Shared` and `UniquePerReplica` modes.**

* In **Shared** mode, there's only one Service: `vllm-pd-lws`
* In **UniquePerReplica** mode, there's one Service per replica, e.g., `vllm-pd-lws-0`

**In this case, we need a custom Service definition that can distinguish between P (prefiller) and D (decoder) components, ideally using Pod labels.**
For example:

* `decoder-svc`: `apps.kubernetes.io/pod-index=1`
* `prefiller-svc`: `apps.kubernetes.io/pod-index=0`

examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
```diff
@@ -0,0 +1,85 @@
+apiVersion: leaderworkerset.x-k8s.io/v1
+kind: LeaderWorkerSet
+metadata:
+  name: vllm-pd-lws
+spec:
+  replicas: 1
+  leaderWorkerTemplate:
+    size: 3  # leader + 2 workers
+    restartPolicy: RecreateGroupOnPodRestart
+
```
nicole-lihui: blank line？I don't see any meaning here

examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
panpan0000: 1) this is build-in K8S headless feature `vllm-pd-lws-0-1.vllm-pd-lws` : access individual pod via pod-name + svc-name.   (the `vllm-pd-lws` svc has been auto created by LWS as your said.)

2) "UniquePerReplica" is to provide extra service for each replicas (not workers in same group)
for example, 
when set `networkConfig.subdomainPolicy` = `UniquePerReplica` (default is `Shared`)
if LWS.replicas = 2, there will be 2 svc by `kubectl get svc`:      `vllm-pd-lws-0`, `vllm-pd-lws-1`

examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
panpan0000: fixed .thanks

examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
nicole-lihui: You're right. Here, I forgot about the headless feature

nicole-lihui:
YAML looks clean and correct. Well done! 👍

examples/disagg_prefill/1p1d/k8s/1p1d-lws.yaml
```diff
@@ -0,0 +1,86 @@
...
+          - |
+            source /opt/venv/bin/activate
+            # Processing config files
+            decoder_svc="vllm-pd-lws-0-1.vllm-pd-lws" # worker-1 svc created by LWS
+            cfg_path="/workspace/LMCache/examples/disagg_prefill/1p1d/configs"
+            prefill_conf="${cfg_path}/lmcache-prefiller-config.yaml"
+            decoder_conf="${cfg_path}/lmcache-decoder-config.yaml"
+            sed -i s/localhost/${decoder_svc}/ $prefill_conf
+            sed -i s/localhost/0.0.0.0/ $decoder_conf
+            MODEL="Qwen/Qwen2.5-0.5B"
```
nicole-lihui: MODEL via ENV would be more flexible ？

nicole-lihui: This is just a brainstorm and doesn’t affect the current functionality.

panpan0000: done, thanks

nicole-lihui:
LGTM

mgw2168-1:
@panpan0000 Hi, I deployed vllm for xpyd and used it https://github.com/LMCache/LMCache/blob/dev/examples/disagg_prefill/disagg_proxy_server.py The problem with the proxy that started is that the streaming return always returns the first token, and the ones after it are gone. Do you know the reason for this?It is normal in Docker environment.
image

PR #993, opened 2025-07-07
[Bug]: Double Free

### LINKED ISSUE #987: A memory leak bug in cpu offload of version 0.3.1
## Description

After upgrading to v0.3.1.post1, we tested the CPU offload feature in our production environment to accelerate inference using the memory buffer.

**YAML spec:**
```yaml
- name: LMCACHE_LOG_LEVEL
  value: DEBUG
  value: /tmp
- name: VLLM_USE_V1
  value: '1'
  value: 'True'
- name: VLLM_RPC_TIMEOUT
  value: '1000000'
- name: LMCACHE_LOCAL_CPU
  value: 'True'
  value: '100'
- name: LMCACHE_CHUNK_SIZE
  value: '128'
```

When first launched, everything worked well according to the logs. However, after some time, we observed that the hit rate gradually decreased, and eventually, the logs showed that the hit tokens remained at 0, for example:

```
[2025-07-07 04:36:10,801] LMCache INFO: Reqid: cmpl-57823e79a4e3497ab79f238db64bc102-0, Total tokens 7283, LMCache hit tokens: 0, need to load: -400 (vllm_v1_adapter.py:836:lmcache.integration.vllm.vllm_v1_adapter)
[2025-07-07 04:36:11,258] LMCache INFO: Reqid: cmpl-0a067a58d03d492cae752427bba11b03-0, Total tokens 7325, LMCache hit tokens: 0, need to load: -336 (vllm_v1_adapter.py:836:lmcache.integration.vllm.vllm_v1_adapter)
```

This is unreasonable given a 100GB memory buffer.

---

## Analysis

After investigating the code, we found a bug in the management of memory objects:

1. The `lookup` method adds a parameter `pin=True`, which keeps the hit memory object in the request from being evicted.

    ```python
    frames = self.socket.recv_multipart(copy=False)
    token_ids = self.decoder.decode(frames)
    result = self.lmcache_engine.lookup(token_ids, pin=True)
    response = result.to_bytes(4, "big")
    self.socket.send(response)
    ```

2. The pinned memory object is supposed to be unpinned when the kvcache is retrieved to the GPU, which works most of the time. However, sometimes the next request is not a retrieval of the previously selected tokens, but a store request. If a store request arrives and the cache is full, it triggers eviction during allocation, which iterates over the cache dict and picks memory objects one by one.

It is possible to pick a pinned memory object. In the `ref_count_down` method, it prevents the pinned memory object from being freed:

```python
evict_keys = []
with self.cpu_lock:
    for evict_key in self.hot_cache:
        old_mem_obj = self.hot_cache[evict_key]
        # If the ref_count > 1, we cannot evict it as the cpu memory
        # might be used as buffers by other storage backends
        if old_mem_obj.get_ref_count() > 1:
            continue
        evict_keys.append(evict_key)

        old_mem_obj.ref_count_down()
        memory_obj = self.memory_allocator.allocate(shape, dtype, fmt)
        logger.debug("Evicting 1 chunk from cpu memory")
        if memory_obj is not None:
            break
for evict_key in evict_keys:
    self.remove(evict_key)
```

However, the key of the pinned memory object is still added to `evict_keys` at the end, whether it is freed or not. The pinned memory object can skip the `free` method, but its `ref_count` becomes 0, and then `hot_cache` deletes it. In this case, its memory is never released, and the `hot_cache` usage keeps decreasing because it cannot allocate new memory objects.

---
/cc @YaoJiayi @ApostaC

sammshen:
Fix https://github.com/LMCache/LMCache/issues/987

This issue identifies issue of LMCache CPU hot cache running out of space. PTAL: @wangqia0309

1. fix double free in allocate() and clear()

2. skip over evicting pinned memory objects from both hot cache and the memory allocator

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -328,6 +328,14 @@ class TensorMemoryObj(MemoryObj):
     def ref_count_down(self):
         with self.lock:
             self.meta.ref_count -= 1
+            if self.meta.ref_count < 0:
+                logger.warning(
+                    f"Ref count of MemoryObj {self.meta.address}"
+                    f"is negative: {self.meta.ref_count}."
+                    "Double free occurred somewhere."
+                    "Setting ref count back to 0 as a hack but please find the bug."
+                )
+                self.meta.ref_count = 0
             if (
                 self.meta.ref_count == 0
                 and self.parent_allocator is not None
--- a/lmcache/v1/storage_backend/local_cpu_backend.py
+++ b/lmcache/v1/storage_backend/local_cpu_backend.py
@@ -246,7 +246,8 @@ class LocalCPUBackend(StorageBackendInterface):
                 old_mem_obj = self.hot_cache[evict_key]
                 # If the ref_count > 1, we cannot evict it as the cpu memory
                 # might be used as buffers by other storage backends
-                if old_mem_obj.get_ref_count() > 1:
+                # Also, don't evict pinned objects
+                if old_mem_obj.get_ref_count() > 1 or old_mem_obj.is_pinned:
                     continue
                 evict_keys.append(evict_key)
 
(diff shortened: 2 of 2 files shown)


YaoJiayi (approved):
LGTM

PR #1018, opened 2025-07-10
FIX Runtime error & Refactor: Optimize cache_controller for robustness and style

skyloevil:
# FIX Runtime error & Refactor: Improve Robustness and Style in cache_controller

This pull request introduces several minor improvements to the `cache_controller` module to enhance its robustness and code quality without altering the existing logic.

# Key Changes

## kv_controller.py
Here's the content formatted in Markdown:

### Description

This change refactors the `deregister` method within the `KVController` to prevent a critical runtime error. The modification ensures that keys are removed from the `self.kv_pool` dictionary in a safe, predictable, and robust manner by avoiding mutation during iteration.

### The Problem

The previous logic for deregistering a worker could be implemented in a way that attempts to delete keys from the `self.kv_pool` dictionary while iterating over it.

An implementation structured like the one below is unsafe in Python:

```python
# Unsafe Implementation
for key in self.kv_pool:
    # Filter out metadata related to the deregistered worker
    self.kv_pool[key] = [
        m for m in self.kv_pool[key]
        if not (m.instance_id == instance_id and m.worker_id == worker_id)
    ]
    
    # If the list of metadata becomes empty, delete the key
    if not self.kv_pool[key]:
        # This operation is unsafe inside the loop
        del self.kv_pool[key]
```

Modifying a dictionary's size (i.e., adding or deleting keys) while iterating over it is a forbidden operation in Python. It breaks the iterator's internal state and immediately raises a `RuntimeError: dictionary changed size during iteration`.

This error would not only crash the process but also leave the system in an inconsistent state, as the deregister operation would be only partially completed. Some keys might be left in the pool erroneously, leading to unpredictable behavior later on.

### The Solution

This PR adopts a standard and safe two-phase "mark and sweep" pattern to handle the key removal:

1. **Mark Phase**: The first loop iterates through the dictionary to perform the filtering. Instead of deleting a key immediately when its associated list becomes empty, the key is "marked" for deletion by appending it to a separate list, `empty_keys`. During this phase, the dictionary's size remains constant, so the iteration is safe.

2. **Sweep Phase**: A second, separate loop then iterates over the `empty_keys` list. For each key collected in the first phase, the corresponding entry is now safely deleted from `self.kv_pool`. Since this loop does not iterate over the dictionary itself, these deletions are safe.

The corrected, robust implementation is as follows:

```python
# Safe Implementation
empty_keys = []
for key in self.kv_pool:
    self.kv_pool[key] = [
        m
        for m in self.kv_pool[key]
        if not (m.instance_id == instance_id and m.worker_id == worker_id)
    ]
    if not self.kv_pool[key]:
        empty_keys.append(key)

# Now, safely delete the marked keys
for key in empty_keys:
    del self.kv_pool[key]
```

This change is crucial for the stability and correctness of the KV Controller, guaranteeing that the deregister operation completes fully and without raising runtime errors.

## registration_controller.py
- Improved robustness of the `deregister` method
- Now handles cases where a `worker_id` is not found in `self.worker_mapping[instance_id]`
- Prevents `ValueError` by logging a warning instead

## executor.py
- Refactored the `execute_workers` method for better code quality and safety
- Moved the `send_and_receive` async function outside the loop to avoid re-creation on each iteration
- Now uses `zip` with `strict=True` to ensure lists of sockets and messages have the same length
- Prevents potential bugs from mismatched list lengths

These changes are intended as small-scale refactorings to make the codebase more stable and maintainable.


lmcache/v1/cache_controller/controllers/registration_controller.py
```diff
@@ -124,9 +124,15 @@ async def deregister(self, msg: DeRegisterMsg) -> None:
...
         if instance_id in self.worker_mapping:
-            self.worker_mapping[instance_id].remove(worker_id)
-            if not self.worker_mapping[instance_id]:
-                del self.worker_mapping[instance_id]
+            try:
+                self.worker_mapping[instance_id].remove(worker_id)
+                if not self.worker_mapping[instance_id]:
+                    del self.worker_mapping[instance_id]
+            except ValueError:
+                logger.warning(
```
Shaoting-Feng, line 132: When will this happen?

PR #1024, opened 2025-07-11
[kv-cache-caculator] add Qwen3 models into caculator

panpan0000:
add some more models and adjust the orders

Test snapshot

image


panpan0000:
the https://lmcache.ai/kv_cache_calculator.html is out of date, we can use CI build or manually update it

PR #1026, opened 2025-07-11
[Bugfix] Metadata file path missing suffix in GdsBackend and WekaGdsB…

YurianStormrage:


lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -366,6 +366,7 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
 
     def _try_to_read_metadata(self, key: CacheEngineKey) -> Optional[DiskCacheMetadata]:
         path, subdir_key, _, _ = self._key_to_path(key)
+        path += _METADATA_FILE_SUFFIX
```
Shaoting-Feng, line 369: I see the path already has `_DATA_FILE_SUFFIX`. Do we still need `_METADATA_FILE_SUFFIX`?

YurianStormrage: `_try_to_read_metadata`:
```python
    def _try_to_read_metadata(self, key: CacheEngineKey) -> Optional[DiskCacheMetadata]:
        path, subdir_key, _, _ = self._key_to_path(key)
        if os.path.exists(path):
            try:
                return self._read_metadata(key, path, subdir_key)
        ...
```
meanwhile, `_read_metadata`:
```python
    def _read_metadata(self, key, filename, subdir_key):
        with open(filename, "rb") as f:
            buf = f.read(_METADATA_MAX_SIZE)
        ...
        metadata = DiskCacheMetadata(
            filename.removesuffix(_METADATA_FILE_SUFFIX), size, shape, dtype
        )
        ...
```
In my opinion, `_read_metadata` needs `_METADATA_FILE_SUFFIX`.

Shaoting-Feng:
Thanks for your PR! It would be best if @sdimitro can take a look. Thanks!

sdimitro (approved):
For the Weka backend I think that's a fine change. It currently works without it because our data file also includes metadata in the beginning. So for the Weka backend LGTM.

Not sure if the GDS backend is the same way.

@YurianStormrage how are you testing this?

YurianStormrage:
> For the Weka backend I think that's a fine change. It currently works without it because our data file also includes metadata in the beginning. So for the Weka backend LGTM.
> 
> Not sure if the GDS backend is the same way.
> 
> @YurianStormrage how are you testing this?

I'm currently implementing a backend for a particular file system, which requires me to handle additional metadata operations. In doing so, I've observed some problems with the metadata operations in `gds_backend`.

From what I've observed, `gds_backend`, just like `weka_gds_backend`, places 4KB of metadata at the beginning of data files, so it's currently working well. However, when handling metadata files separately, some issues arise. I've added a simple test in `test_gds.py` to illustrate this.
```python
def test_gds_backend_metadata_operations():
    # The omitted part is the same as in test_gds_backend_sanity
    try:
        ...
        from lmcache.v1.storage_backend.gds_backend \
            import save_metadata, pack_metadata, \
            _METADATA_FILE_SUFFIX, DiskCacheMetadata

        # Generate metadata
        memory_obj = gds_backend.memory_allocator.allocate(
            [2048, 2048], dtype=torch.uint8
        )
        metadata = pack_metadata(memory_obj.tensor) # ignore the version for simplicity

        path = getattr(gds_backend, '_key_to_path')(TEST_KEY)[0]
        tmp = ".tmp" + "_hard_coded_suffix"

        # Same as the corresponding code in _async_save_bytes_to_disk,
        # but without coroutine.
        save_metadata(path + _METADATA_FILE_SUFFIX, tmp, metadata)

        # If the metadata is not in the hot cache,
        # we should load it from the file.
        disk_metadata = getattr(gds_backend, '_try_to_read_metadata')(TEST_KEY)
        assert isinstance(disk_metadata, DiskCacheMetadata)
    finally:
        ...
```

Shaoting-Feng (approved):
LGTM

YurianStormrage:
@Shaoting-Feng Sorry to bother you, but I was wondering if there will be any more reviews, or should I just close this PR?

sammshen (approved):

PR #1027, opened 2025-07-11
[feature] support batched_get_blocking in remote backend

chunxiaozheng:
support `batched_get_blocking` in remote backend to improve get-blocking performance.

In my test, I deployed with `1p1d`, use `FSConnector`, in our internal test system, run 1000s for every test, the result is as follows:
| | TTFT | QPM | The average length of input tokens | Number of runs |
| ------ | ------ | ------ | ------ | ------|
| baseline(not use lmcache) | 20.57s | 25.56 | 1.55w | 460 |
| use lmcache without this PR | 20.06s | 22.28 | 1.38w | 401 |
| use lmcache with this PR | 2.76s | 28.61 | 1.68w | 515 |

The results indicate that `batched_get_blocking` is highly effective in improving TTFT.

### DIFF
--- a/lmcache/v1/storage_backend/remote_backend.py
+++ b/lmcache/v1/storage_backend/remote_backend.py
@@ -271,7 +271,7 @@ class RemoteBackend(StorageBackendInterface):
 
         t1 = time.perf_counter()
         # batched get
-        if hasattr(self.connection, "batched_get"):
+        if self.connection.support_batched_get():
             future = asyncio.run_coroutine_threadsafe(
                 self.connection.batched_get(new_keys), self.loop
             )
--- a/lmcache/v1/storage_backend/remote_backend.py
+++ b/lmcache/v1/storage_backend/remote_backend.py
@@ -260,32 +260,41 @@ class RemoteBackend(StorageBackendInterface):
 
         # For MLA worker id as 0 mode, use worker_id 0
         if self._mla_worker_id_as0_mode:
-            new_keys = []
-            for key in keys:
-                new_key = CacheEngineKey(
+            new_keys = [
+                CacheEngineKey(
                     key.fmt, key.model_name, key.world_size, 0, key.chunk_hash
                 )
-                new_keys.append(new_key)
+                for key in keys
+            ]
         else:
             new_keys = keys
 
         t1 = time.perf_counter()
         # batched get
         if hasattr(self.connection, "batched_get"):
-            future = asyncio.run_coroutine_threadsafe(self.connection.batched_get(new_keys), self.loop)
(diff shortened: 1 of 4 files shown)


maobaolong (commented):
@chunxiaozheng The test result looks amazing!

guymguym:
Hi @chunxiaozheng @maobaolong @YaoJiayi this looks very promising. Any concerns about merging it?

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -268,6 +268,93 @@ def get_non_blocking(
     ) -> Optional[Future]:
         raise NotImplementedError
 
+    def batched_get_blocking(
+        self,
+        keys: List[CacheEngineKey],
+    ) -> List[Optional[MemoryObj]]:
+        if self.connection is None:
+            logger.warning("Connection is None in batched_get_blocking, returning None")
```
YaoJiayi, line 258: If the connection is `None`, are we planning to re-establish the connection?

YaoJiayi (approved):
LGTM. Just a minor question. And I approved it first:)

YaoJiayi:
@chunxiaozheng Just curious, do we have results for a real remote connector (i.e., not just a file system)?

maobaolong:
> @chunxiaozheng Just curious, do we have results for a real remote connector (i.e., not just a file system)?

@YaoJiayi The `FS` here is not a local fs, it is a distributed file system.

lmcache/v1/storage_backend/remote_backend.py
maobaolong: @YaoJiayi As there is a background RemoteMonitor, the `self.connection is None` would be handled by `RemoteMonitor`, so for this working thread, return None can trigger fast fail aware.

chunxiaozheng:
> @chunxiaozheng Just curious, do we have results for a real remote connector (i.e., not just a file system)?

I will test with another remote connector.

chunxiaozheng:
> @chunxiaozheng Just curious, do we have results for a real remote connector (i.e., not just a file system)?

We use a distributed file system, mount to local directory through fuse.

ziruiliu:
@chunxiaozheng how did your test avoid hitting page cache? In fs connector, files are opened with aiofiles.open() so file could be read from page cache? thanks

chunxiaozheng:
> @chunxiaozheng how did your test avoid hitting page cache? In fs connector, files are opened with aiofiles.open() so file could be read from page cache? thanks

we use our internal distributed file system, mount to prefill node through fuse, and turn off  the page cache  when mounting.

PR #1036, opened 2025-07-14
optimize lock usage and scope

llc-kc:
This PR optimizes lock usage and scope:
1, use "with lock" to replace explicit lock acquire and release calling.
2, only keep minumum lock scope to avoid bug and reduce risk of dead lock. For example, lock useage in StorageManager can cause dead lock, since prefetch callback also use manager_lock:
```
            self.manager_lock.acquire()
            self.prefetch_tasks[key] = prefetch_task
            prefetch_task.add_done_callback(lambda_callback)
            self.manager_lock.release()
```

### DIFF
--- a/lmcache/v1/storage_backend/local_cpu_backend.py
+++ b/lmcache/v1/storage_backend/local_cpu_backend.py
@@ -107,15 +107,15 @@ class LocalCPUBackend(StorageBackendInterface):
             self.hot_cache[key] = memory_obj
             memory_obj.ref_count_up()
 
-            self.usage += memory_obj.get_size()
-            self.stats_monitor.update_local_cache_usage(self.usage)
+        self.usage += memory_obj.get_size()
+        self.stats_monitor.update_local_cache_usage(self.usage)
 
-            # TODO(Jiayi): optimize this with batching?
-            # push kv admit msg
-            if self.lmcache_worker is not None:
-                self.lmcache_worker.put_msg(
-                    KVAdmitMsg(self.instance_id, key.worker_id, key.chunk_hash, "cpu")
-                )
+        # TODO(Jiayi): optimize this with batching?
+        # push kv admit msg
+        if self.lmcache_worker is not None:
+            self.lmcache_worker.put_msg(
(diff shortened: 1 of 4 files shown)


lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -387,24 +384,21 @@ def write_back(self, key: CacheEngineKey, memory_obj: MemoryObj):
...
-                            key.chunk_hash,
-                            "cpu",
-                        )
+            with self.cpu_lock:
+                if self.use_hot and key not in self.hot_cache:
+                    self.hot_cache[key] = memory_obj
+                    memory_obj.ref_count_up()
+
+            # Push kv msg
+            if self.lmcache_worker is not None:
```
Shaoting-Feng: `if` logic has changed here

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -233,22 +228,20 @@ def get_blocking(
...
+            shape = self.dict[key].shape
+            fmt = self.dict[key].fmt
 
-        path = self.dict[key].path
-        dtype = self.dict[key].dtype
-        shape = self.dict[key].shape
-        fmt = self.dict[key].fmt
         assert dtype is not None
         assert shape is not None
         memory_obj = self.load_bytes_from_disk(path, dtype=dtype, shape=shape, fmt=fmt)
```
Shaoting-Feng, line 247: if `self.dict` is changed before here, there will be an error

llc-kc: @Shaoting-Feng fixed

lmcache/v1/storage_backend/local_cpu_backend.py
llc-kc: you mean here should not be modified?

llc-kc: Also fixed

Shaoting-Feng (approved):
LGTM

PR #1052, opened 2025-07-15
[Opt]: Speed up V1 token hashing

sammshen:
https://github.com/LMCache/LMCache/commit/b9e8b81a6c61147a15e7a3f60ccb38ff8e180654 by @zhouwfang updated teh V0 codepath. This PR updates V1 codepath with xx64 hasher as well (thanks @Siddhant-Ray!)

Some analysis:

With the rolling hasher we hash each byte exactly once.
The old scheme concatenated the entire cumulative prefix into a new buffer every chunk; total bytes processed grows like  n^2

xxh64 is less collision resistant but we need to store 30M chunks to reach 0.027% collision rate, which is 80k GB of KV cache even on a Qwen 4B with int8 quantization so we should still be safe.

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -16,11 +16,11 @@
 from typing import Iterable, List, Optional, Tuple, Union
 import abc
 import array
-import hashlib
 
 # Third Party
 from transformers import AutoTokenizer
 import torch
+import xxhash
 
 # First Party
 from lmcache.config import LMCacheEngineMetadata
@@ -85,20 +85,21 @@ class ChunkedTokenDatabase(TokenDatabase):
             chunk_hash,
         )
 
-    def _get_init_hash(self) -> str:
-        return ""
+    def _get_init_hasher(self) -> xxhash.xxh64:
+        # seed defaults to 0
+        return xxhash.xxh64()
 
-    def _hash(
+    def _hash_incremental(
         self,
         tokens: Union[torch.Tensor, List[int]],
-        prefix_hash: str,
+        hasher: xxhash.xxh64,
     ) -> str:
-        # TODO: change it to a more efficient hash function
         if isinstance(tokens, torch.Tensor):
             tokens_bytes = tokens.cpu().to(torch.uint32).numpy().tobytes()
         elif isinstance(tokens, list):
             tokens_bytes = array.array("I", tokens).tobytes()
-        return hashlib.sha256(prefix_hash.encode("ascii") + tokens_bytes).hexdigest()
+        hasher.update(tokens_bytes)
+        return hasher.hexdigest()
 
     def _chunk_tokens(
         self,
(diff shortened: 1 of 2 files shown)


sammshen:
Some tiny experiments show ~2x speedup (which is not a lot and only around 10-15 nano seconds per 2048 tokens):

```python
#!/usr/bin/env python
"""
Compare hashing cost:

1. OLD  – hashlib.sha256(prev_hex + chunk_bytes).hexdigest()
2. NEW  – xxhash.xxh64().update(chunk_bytes); hexdigest() per chunk

Results printed in ms per prompt.
"""
import argparse, statistics, time, hashlib, xxhash, torch
from typing import Callable

# ---------- hash path implementations ----------

def old_sha256(tokens: torch.Tensor, chunk: int) -> None:
    prev = ""                       # prefix hash string
    view = memoryview(tokens.numpy())
    step = chunk * 4                # int32 → 4 B
    for i in range(0, len(tokens), chunk):
        chunk_bytes = view[i*4 : i*4 + step]
        prev = hashlib.sha256(prev.encode() + chunk_bytes).hexdigest()

def new_xxh64(tokens: torch.Tensor, chunk: int) -> None:
    h = xxhash.xxh64()
    view = memoryview(tokens.numpy())
    step = chunk * 4
    for i in range(0, len(tokens), chunk):
        h.update(view[i*4 : i*4 + step])
        _ = h.hexdigest()           # 16-char hex, like old path

# ---------- benchmark driver ----------

def time_once(fn: Callable, tokens, chunk) -> float:
    t0 = time.perf_counter()
    fn(tokens, chunk)
    return time.perf_counter() - t0

def bench(seq_len: int, chunk: int, iters: int) -> tuple[float, float]:
    tokens = torch.randint(0, 32000, (seq_len,), dtype=torch.int32)
    sha, xxh = [], []
    for _ in range(iters):
        sha.append(time_once(old_sha256, tokens, chunk))
        xxh.append(time_once(new_xxh64,  tokens, chunk))
    return statistics.mean(sha), statistics.mean(xxh)

if __name__ == "__main__":
    p = argparse.ArgumentParser()
    p.add_argument("--seq",  type=int, nargs="+", default=[1024, 2048, 4096])
    p.add_argument("--chunk", type=int, default=64)
    p.add_argument("--iters", type=int, default=30)
    args = p.parse_args()

    print(f"chunk={args.chunk}, iters={args.iters}")
    for L in args.seq:
        sha_m, xxh_m = bench(L, args.chunk, args.iters)
        speedup = sha_m / xxh_m
        print(f"  seq={L:5d}:  sha256 {sha_m*1e3:6.2f} ms   "
              f"xxh64 {xxh_m*1e3:6.2f} ms   ×{speedup:4.1f}")

```

```bash
python bench_hash_paths.py --seq 2048 4096 8192 --chunk 64 --iters 50
chunk=64, iters=50
  seq= 2048:  sha256   0.04 ms   xxh64   0.02 ms   × 2.1
  seq= 4096:  sha256   0.07 ms   xxh64   0.03 ms   × 2.2
  seq= 8192:  sha256   0.15 ms   xxh64   0.06 ms   × 2.3
  ```

Siddhant-Ray (approved):
LGTM

Siddhant-Ray:
@sammshen Wondering if it will be useful to have a watch to catch hash collision errors and see how much they  actually degrade performance (especially for large KV cache sizes)

Shaoting-Feng (approved):
LGTM

sammshen:
> @sammshen Wondering if it will be useful to have a watch to catch hash collision errors and see how much they actually degrade performance (especially for large KV cache sizes)

That would definitely be useful/interesting but I can't come up with a good way off the top of my head on how to catch collisions since the CacheEngineKey itself does not contain the tokens that produced it!

sammshen:
looks like Jiayi made a huge refactor in https://github.com/LMCache/LMCache/commit/2f039eb9c610bc60ab6cf4ea6e2d53269e75f995#diff-9c18f236bd4af44762c4fd9208d7169e851237363c69ce938ac0a7f6abbf025e, closing this one :)

PR #1056, opened 2025-07-15
[BUGFIX] self._maxes of sortedcontainers.SortedList Unchanged

yangxianpku:
1. _maxes stores the maximum value of each internal sublist in a SortedList.
2. It enables fast O(log n) searches and inserts by narrowing down the target sublist.

self._maxes of sortedcontainers.SortedList should automatically change with the elements of sortedcontainers.SortedList, and the original code directly operates on the elements without modifying self._maxes, resulting in errors when evicting elements in sortedcontainers.SortedList.

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -587,10 +587,14 @@ class TensorMemoryAllocator(MemoryAllocatorInterface):
             merge_succ = False
 
         if merge_prev and merge_succ:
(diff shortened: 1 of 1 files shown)


Shaoting-Feng:
Since `self.explicit_list` is initialized as

```python
sortedcontainers.SortedList(key=lambda x: x.start)
```

—sorting entries by their start address—I don’t think modifying the size will be a problem.

PR #1071, opened 2025-07-16
[FIX] Add type checking and conversion for rpc path generation

zejunchen-zejun:
Hi, @YaoJiayi 
I find for some existing examples, like vllm LMCache example (https://github.com/vllm-project/vllm/blob/main/examples/others/lmcache/disagg_prefill_lmcache_v1/disagg_vllm_launcher.sh#L34), the rpc port `lmcache_rpc_port` will be specified as a str variable instead of int, so the `rpc_port += tp_rank` could lead to the type error.
I add type checking and conversion for above situation. Could you help review?
Thank you.

### DIFF
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -96,7 +96,15 @@ def get_zmq_rpc_path_lmcache(
 
     base_url = envs.VLLM_RPC_BASE_PATH
 
-    rpc_port += tp_rank
+    if isinstance(rpc_port, int):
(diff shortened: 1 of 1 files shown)


zejunchen-zejun:
Close it because the issue has been fixed in https://github.com/LMCache/LMCache/commit/59f47bd203d67b5f8aab78bbe6b4a839494e2ea6#diff-051153d4478e66d21e8b274e292a36f13530568856207ddf8b8cab9098c32eccR99-R102

PR #1075, opened 2025-07-16
Bug/unpin lookup

### LINKED ISSUE #1034: [BUG] Pinned memory objects which are not retrieved are never unpinned, causing memory leak
**Describe the bug**
When a memory object is hit in lookup() but is never retrieved, the object stays pinned and is never unpinned.

**How to reproduce**
When the tokens are hit on GPU, the need_to_load can be zero negative if the number of matched tokens on CPU is equal or smaller than on GPU. In that case, the hit memory object are never retrieved and thus never unpinned.

This bug cannot be fixed by the fix to issue #987

sammshen:
Replaces: https://github.com/LMCache/LMCache/pull/1046 as Fix https://github.com/LMCache/LMCache/issues/1034#issuecomment-3079261730

@ApostaC thanks for the suggestion! this avoids:
1. the adapter having to call another method inside the cache engine (EDIT: STILL HAVE TO)
2. a second pass through the token database

There was also a bug with lookup layerwise where we potentially don't pin all layers that is patched here too


YaoJiayi (commented):
@sammshen IIUC, do we have the assumption that lookup and retrieve will be called one by one? What if there are two consecutive lookups and one retrieve? Can we handle those cases?

sammshen:
@YaoJiayi , you are right and the only way (?) to fix this is to have the adapter call unpin instead of retrieve() or retrieve_layer() since `start_load_kv()` is always called by worker

YaoJiayi:
> @YaoJiayi , you are right and the only way (?) to fix this is to have the adapter call unpin instead of retrieve() or retrieve_layer() since `start_load_kv()` is always called by worker

@sammshen My take is to unpin not only the retrieved tokens but also the tokens that are masked out? What do you think?

sammshen:
@YaoJiayi , yep! that is the current fix (I think)!

Everything pinned in the lookup (including masked keys) is added to the list of lookup_pins and unpinned at end of step

YaoJiayi:
> @YaoJiayi , yep! that is the current fix (I think)!
> 
> Everything pinned in the lookup (including masked keys) is added to the list of lookup_pins and unpinned at end of step

Will there be a problem in the following setting?

lookup(A), lookup(B), retrieve(A) (in which both A and B are unpinned)?

sammshen:
just decoupled the unpin from retrieve() and put it into the adapter instead.

previously there is no issue with 2x lookup and 1x retrieve because the lookup_pins is a list so even if there are shared tokens between A and B, they will still be unpinned twice. 

the danger with having unpin in retrieve/retrieve_layer was that there may be many many lookups without a retrieve (store heavy workload) and we should always still unpin at the end of every step

YaoJiayi:
> just decoupled the unpin from retrieve() and put it into the adapter instead.
> 
> previously there is no issue with 2x lookup and 1x retrieve because the lookup_pins is a list so even if there are shared tokens between A and B, they will still be unpinned twice.
> 
> the danger with having unpin in retrieve/retrieve_layer was that there may be many many lookups without a retrieve (store heavy workload) and we should always still unpin at the end of every step

Sorry I'm still a bit confused:) I mean:
lookup(A), lookup(B), retrieve(A) (in which both A and B are unpinned and B gets evicted), retrieve(B) will fail?

sammshen:
yes, that was a race condition! 😰

sorry I probably misunderstand the adapter but I'm not sure in what scenarios that kind of interleaving might happen? does the adapter / vllm loop not always do scheduler lookup(A, pin=True) -> worker maybe retrieve(A) -> scheduler lookup(B, pin=True) -> worker maybe retrieve(B)? If B is evicted by some retrieve(), it must be after we already checked for retrieve(B)?

let me double check

YaoJiayi:
@sammshen we can discuss bit tmrw:)

PR #1089, opened 2025-07-17
Add structure for pluggable hash libraries

### LINKED ISSUE #1079: Feature request: deterministic and consistently-typed hashing
`self.hash_func = hash` currently relies on Python’s built-in `hash`, which is not stable across interpreter runs.

Proposed enhancement
• Provide an optional, pluggable hashing backend (e.g., xxHash, murmur3, or SipHash-1-3) whose results are:
 1. deterministic across runs and platforms,
 2. uniformly emitted as either `str` or `int`
This would deprecate the opaque reliance on `hash` and guarantee predictable semantics in all code paths.

hickeyma:
Needs to handle vLLM hash algorithm support because vLLM  allows hashes in addition to tokens for querying LMCache.

Fix #1079

### DIFF
--- a/examples/cache_controller/example.yaml
+++ b/examples/cache_controller/example.yaml
@@ -6,4 +6,4 @@ max_local_cpu_size: 5
 enable_controller: True
 lmcache_instance_id: "lmcache_default_instance"
 controller_url: "localhost:9001"
-lmcache_worker_url: "localhost:8001"
+lmcache_worker_port: 8001
--- a/lmcache/v1/cache_controller/controllers/kv_controller.py
+++ b/lmcache/v1/cache_controller/controllers/kv_controller.py
@@ -61,7 +61,6 @@ class KVController:
 
         # TODO(Jiayi): remove this hardcode
         self.token_database = ChunkedTokenDatabase()
-        self.token_database.chunk_size = 256
 
     def post_init(self, cluster_executor):
         """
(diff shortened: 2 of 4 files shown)


hickeyma:
@YaoJiayi This is a starting point for discussion on how best to implement this. Stared with hard coding `xxhash` library for now in `TokenDatabase`.

YaoJiayi (commented):
Thanks @hickeyma ! We need to use the python internal `hash()` for now since that's what vllm does. Can we add a config to support different hashing algo?

hickeyma:
> Thanks @hickeyma ! We need to use the python internal hash() for now since that's what vllm does. Can we add a config to support different hashing algo?

@YaoJiayi I'll investigate this further.

hickeyma:
@YaoJiayi PR is ready for review now. Is it ok to just support the vLLM hashing functions for now or do you want me to wrap `xxhash` in this PR as well, as per @xiaguan comment in #1079?

examples/cache_controller/example.yaml
```diff
@@ -6,4 +6,4 @@ max_local_cpu_size: 5
 enable_controller: True
 lmcache_instance_id: "lmcache_default_instance"
 controller_url: "localhost:9001"
-lmcache_worker_url: "localhost:8001"
\ No newline at end of file
+lmcache_worker_port: 8001
```
YaoJiayi, line 9: Good catch!

YaoJiayi (approved):
LGTM!

PR #1092, opened 2025-07-17
[bug]: unpin unretrieved

sammshen:
@YaoJiayi thanks very much for your feedback. this replaces https://github.com/LMCache/LMCache/pull/1075

you are right that it is possible for the scheduler to be async with the worker when `max_concurrent_batches > 1` (e.g.
PP or async scheduling is enabled  with `scheduler_config.async_scheduling = True`).


YaoJiayi (approved):
lgtm

PR #1107, opened 2025-07-21
[optimize] support not save chunk meta in remote backend

chunxiaozheng:
# DESCRIPTION
In remote backend, connectors will save chunk meta, which includes `shape`, `dtype` and `format`, there are currently two ways to save, one is to save metadata and data as two keys like `RedisConnector`, and the other is to save them in one key like `FSConnector`. When reading, it must read metadata firstly and then read data, `RedisConnector` requires two RPC operations, and `FSConnector` requires two IO operations. The current PR provides a configuration option that does not save chunk meta, reducing one RPC or IO. 

# HOW TO USE
set `save_chunk_meta` in `extra_config`, default is `True`
```
chunk_size: 256
local_device: "cpu"
remote_url: "fs://host:0/data/test1"
remote_serde: "naive"
pipelined_backend: False
local_cpu: False
max_local_cpu_size: 10
save_unfull_chunk: False
extra_config:
  remote_enable_mla_worker_id_as0: True
  save_chunk_meta: False
```

# TODO
Other connectors can also be optimized using this method, the implementation of connectors is as follows: 
- [x] FSConnector
- [ ] RedisConnector
- [ ] RedisSentinelConnector
- [ ] InfinistoreConnector
- [ ] MooncakestoreConnector


lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -29,6 +34,86 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
+        if (
+            config is None
+            or metadata is None
+            or config.extra_config is None
+            or config.extra_config.get("save_chunk_meta", True)
+        ):
+            return
+
+        self.save_chunk_meta = False
+        self.meta_shape = torch.Size(
```
Shaoting-Feng, line 47: Will `self.meta_shape` be different with different `self.meta_fmt`?
From `VLLMPagedMemGPUConnectorV2`:
```python
if self.use_mla:
    # kv_caches[0].shape: [num_pages, page_size, head_size]
    assert kv_caches[0].dim() == 3
    self.page_buffer_size = kv_caches[0].shape[0] * kv_caches[0].shape[1]
else:
    # kv_caches[0].shape: [2, num_pages, page_size, num_heads, head_size]
    assert kv_caches[0].dim() == 5
    self.page_buffer_size = kv_caches[0].shape[1] * kv_caches[0].shape[2]
```

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -29,6 +34,86 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
+            [
+                metadata.kv_shape[1],
+                metadata.kv_shape[0],
+                metadata.kv_shape[2],
+                metadata.kv_shape[3] * metadata.kv_shape[4],
+            ]
+        )
+        self.meta_dtype = metadata.kv_dtype
+        self.meta_fmt = (
```
Shaoting-Feng, line 56: Do we also need to support the layerwise case?

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -29,6 +34,86 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
+                f"full_chunk_size: {self.full_chunk_size}"
+            )
+
+        if bytes_read == self.full_chunk_size:
+            # full chunk
+            return memory_obj
+
+        # NOTE: for unfull chunk, we have no way to verify
+        shape_list = list(memory_obj.meta.shape)
+        shape_list[2] = bytes_read // self.single_token_size
```
Shaoting-Feng, line 99: Is this related to the KV cache fmt?

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: > Will `self.meta_shape` be different with different `self.meta_fmt`? From `VLLMPagedMemGPUConnectorV2`:
> 
> ```python
> if self.use_mla:
>     # kv_caches[0].shape: [num_pages, page_size, head_size]
>     assert kv_caches[0].dim() == 3
>     self.page_buffer_size = kv_caches[0].shape[0] * kv_caches[0].shape[1]
> else:
>     # kv_caches[0].shape: [2, num_pages, page_size, num_heads, head_size]
>     assert kv_caches[0].dim() == 5
>     self.page_buffer_size = kv_caches[0].shape[1] * kv_caches[0].shape[2]
> ```

Sure, the content in `self.meta_shape` is related to whether `MLA` is enabled or not. But, when calculate `kv_shape` of `metadata`, `MLA` and other factors has already been taken into consideration, so here we can directly use the content of `kv_shape` to calculate `self.meta_shape`.

the `metadata` and `kv_shape` is calculated by the following code that in `init_lmcache_engine`:
```
    # construct kv shape (for mem pool)
    num_layer = model_config.get_num_layers(parallel_config)
    chunk_size = config.chunk_size
    num_kv_head = model_config.get_num_kv_heads(parallel_config)
    head_size = model_config.get_head_size()
    kv_shape = (num_layer, 1 if use_mla else 2, chunk_size, num_kv_head, head_size)
    logger.info(f"use mla: {use_mla}, kv shape: {kv_shape}")

    # Change current device.
    torch.cuda.device(parallel_config.rank)
    device = torch.device(f"cuda:{parallel_config.rank}")
    metadata = LMCacheEngineMetadata(
        model_config.model,
        parallel_config.world_size,
        parallel_config.rank,
        "vllm",
        kv_dtype,
        kv_shape,
        use_mla,
    )
```

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: We do not use layerwise case in our scenarios , so I will leave a todo here to test layerwise later. Currently, set layerwise case must save chunk meta.

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: Sure, when calculate `self.single_token_size`, these factors have already been taken into consideration, which also can refer to the reply above.

Shaoting-Feng (approved):
LGTM

maobaolong (approved):
@chunxiaozheng LGTM, thanks for your great improvement!

Issue #1064, opened 2025-07-15
gpu buffer memory reuse bug causing CacheBlend incorrect model outputs on subsequent prompts


amyluo1106:
**Describe the bug**
I have been running a modified version of the `blend.py` example, but attempting to make 3 prompt calls that are exactly the same. The expected behaviour is that we will see the same model output but generated faster for the second and third calls which should be reusing tokens from the KV Cache. 
The second call indeed demonstrates this behaviour, but in the third call (and any other subsequent calls), the model starts outputting nonsense such as 
`':>, the, and,,,,,,,,,,,,,,,,,,ce,ceecece,,'`

Investigating this issue, it seems that it comes from the way the gpu memory buffer is being allocated and used. In the `gpu_connector.py`, we allocate a gpu buffer [here](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/gpu_connector.py#L410-L412). It seems that this initializes a tensor of all 0s on the second prompt, then when we populate the gpu memory buffer in subsequent code, we noticed that where the separator tokens are, the tensor remains 0 entries for the second prompt. But on the third prompt the tensor is initialized as non-zero (likely due to memory being reused, and was populated previously from the second prompt? and was never cleared). But we do not explicitly set separator token positions to 0 in subsequent code so in the third prompt the tensor, takes whatever non-zero value is in that position for the separator token positions. This seems to be what is affecting the accuracy of the model output as the memory is not being cleaned up.

In debugging when I reinitialize the tensor to be all 0s before the third prompt, I am able to get correct model outputs. I am wondering what would be the best way to address this bug to maintain optimal run time? Should we be doing better clean-up or inserting 0s to the separator token positions? Appreciate any guidance from the team, thanks!

sammshen:
reproduced the bug and taking a closer look. thanks!

sammshen:
@amyluo1106  thanks for your analysis! simplest solution, like you mentioned would be:

Inside of `batched_to_gpu(...)` in `VLLMBufferLayerwiseGPUConnector`
```python
        assert compute_gpu_buffer_obj.tensor is not None
        assert load_gpu_buffer_obj.tensor is not None

        # Clear buffers to prevent stale data from previous allocations
        compute_gpu_buffer_obj.tensor.zero_()
        load_gpu_buffer_obj.tensor.zero_()
```

However, I think targeting the gaps between cached chunks might be a little more elegant. 

[...]

PR #1108, opened 2025-07-21
[fix]: zeroing out separator token positions

sammshen:
https://github.com/LMCache/LMCache/issues/1064 @amyluo1106 identifies that the separator tokens are prone to corruption because they are not written to when using buffers from the gpu memory allocator. 

a naive solution would be to always zero out the load and compute buffers (and `.zero_` in torch should be pretty optimized so maybe that might be the way to go?)

this is a proposal to find the gaps and zero them out. it uses a single write operation as well but has some overhead to compute the gaps. On tests with the `blend.py` example, this solution sometimes is 1ms better than the full zero (0.16ms vs 0.15ms).

### DIFF
--- a/lmcache/v1/gpu_connector.py
+++ b/lmcache/v1/gpu_connector.py
@@ -356,6 +356,9 @@ class VLLMBufferLayerwiseGPUConnector(GPUConnectorInterface):
 
         self.buffer_mapping = {}
 
+        # track gap positions between blended chunks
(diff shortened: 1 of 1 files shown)


Shaoting-Feng (approved):
LGTM

hickeyma (approved):
LGTM, thanks @sammshen for the fix.

PR #1112, opened 2025-07-21
[Core] Make PD and offloading compatible

YaoJiayi:
Make PD and offloading compatible

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

`git tag vx.x.x`

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

`git tag -d vx.x.x`

For example:
`git tag -d v0.3.0`

For example:

> [!TIP]

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -121,7 +121,7 @@ class LMCacheEngine:
         # HACK: remove this in the future
         # NOTE (Jiayi): This is currently used to support
         # dropping the kv cache in nixl backend at decoder.
-        self.remove_after_retrieve = config.enable_nixl
+        self.remove_after_retrieve = config.nixl_role == "receiver"
 
         if self.enable_p2p:
             self.distributed_loop = asyncio.get_event_loop()
@@ -503,13 +503,9 @@ class LMCacheEngine:
 
         # TODO(Jiayi): Remove the following for loop with batched operations
         for key, memory_obj in zip(reordered_keys, reordered_memory_objs, strict=False):
-            memory_obj.ref_count_down()
-
-            # NOTE (ApostaC): This is only for the current implementation:
-            # When the object is retrieved back to vLLM, the storage backend
-            # will immediately remove the object from itself
             if self.remove_after_retrieve:
                 self.storage_manager.remove(key)
+            memory_obj.ref_count_down()
 
         retrieved_tokens = torch.sum(ret_mask)
         self.stats_monitor.on_retrieve_finished(monitor_req_id, retrieved_tokens)
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -220,8 +220,6 @@ class LMCacheEngine:
             tokens, hashes, offsets, mask
         ):
             assert isinstance(key, CacheEngineKey)
-            if self.storage_manager.contains(key):
-                continue
             # Allocate the memory object
             num_tokens = end - start
             kv_shape = self.gpu_connector.get_shape(num_tokens)
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -121,7 +121,9 @@ class LMCacheEngine:
         # HACK: remove this in the future
         # NOTE (Jiayi): This is currently used to support
         # dropping the kv cache in nixl backend at decoder.
-        self.remove_after_retrieve = config.nixl_role == "receiver"
+        self.remove_after_retrieve = (
+            config.enable_nixl and config.nixl_role == "receiver"
+        )
 
         if self.enable_p2p:
             self.distributed_loop = asyncio.get_event_loop()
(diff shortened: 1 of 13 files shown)


lmcache/v1/cache_engine.py
```diff
@@ -119,7 +119,9 @@ def __init__(
         )
 
         # HACK: remove this in the future
-        self.remove_after_retrieve = config.enable_nixl
+        # NOTE (Jiayi): This is currently used to support
+        # dropping the kv cache in nixl backend at decoder.
+        self.remove_after_retrieve = config.nixl_role == "receiver"
```
vladnosiv: nitpick: it's safer to check both conditions here: config.enable_nixl and config.nixl_role == "receiver"

lmcache/v1/storage_backend/__init__.py
```diff
@@ -63,29 +63,27 @@ def CreateStorageBackends(
...
+        local_cpu_backend = LocalCPUBackend(
+            config,
+            memory_allocator,
+            lookup_server,
+            lmcache_worker,
+        )
+        backend_name = str(local_cpu_backend)
+        storage_backends[backend_name] = local_cpu_backend
 
     if config.local_disk and config.max_local_disk_size > 0:
```
vladnosiv, line 90: what are the restrictions here so as not to include all other features? I tested a similar code and did not see any problems when turning on offloading to disk too

lmcache/v1/storage_backend/__init__.py
```diff
@@ -63,29 +63,27 @@ def CreateStorageBackends(
...
     # other backends might need it as a buffer.
-    local_cpu_backend = LocalCPUBackend(
-        config,
-        memory_allocator,
-        lookup_server,
-        lmcache_worker,
-    )
-    backend_name = str(local_cpu_backend)
-    storage_backends[backend_name] = local_cpu_backend
+    if config.local_cpu:
```
vladnosiv: this is a bit unclear: previously, I could have omitted to specify the inclusion of local_cpu and the rest of the engines would have received it anyway as a buffer, now it requires explicit indication.

lmcache/v1/memory_management.py
```diff
@@ -1566,3 +1633,83 @@ def __init__(self, size: int, device=None):
 
     def __del__(self):
         self.cuFileBufDeregister(ctypes.c_void_p(self.base_pointer))
+
+
+class NixlCPUMemoryAllocator(MemoryAllocatorInterface):
+    """
+    NIXL + CPU Memory Allocator
```
vladnosiv, line 1640: The question is not so much about this pr as about the code in the project as a whole, are you considering making a separate interface for the transfer engine? 
According to the code, there are quite a lot of specifics about nixl in places where it seems to be enough to keep in mind that it is an abstract transfer engine. 
For example, if you want to add mooncake transfer now, it may involve quite a lot of changes instead of adding a new implementation of one interface and editing the factory.

lmcache/v1/storage_backend/__init__.py
Shaoting-Feng: Good catch. The point here is that when doing disaggregated prefill with NIXL, if CPU offloading is not selected, we do not initialise a `LocalCPUBackend`. However, in the previous logic, whether to use the CPU for storage was determined by `self.use_hot` in `LocalCPUBackend`. I’ll fix this judgement.

lmcache/v1/storage_backend/__init__.py
Shaoting-Feng: Thanks for your effort! It’s great to know this. Enabling disaggregated prefill and CPU offloading simultaneously is our top priority, so we haven’t worked on the slower backends yet. But please feel free to open a PR if you’re interested.

lmcache/v1/storage_backend/__init__.py
```diff
@@ -63,29 +63,29 @@ def CreateStorageBackends(
...
-        memory_allocator,
-        lookup_server,
-        lmcache_worker,
-    )
-    backend_name = str(local_cpu_backend)
-    storage_backends[backend_name] = local_cpu_backend
+    if config.enable_nixl and not config.local_cpu:
+        pass
+    else:
+        local_cpu_backend = LocalCPUBackend(
```
vladnosiv, line 81: by the way, it seems there is a place for funny behavior: now it is possible to create a decode node with a local CPU cache, since there is no more validation. however, due to deletion when retrieving, such a cache will work 1 time. maybe I don't see any other limiting factors right away

Shaoting-Feng: Right now only the prefill node will do cpu offloading and the `config.local_cpu` for the decode node is `False`.

vladnosiv: I agree on the meaning, but from the point of view of validation, it seems that I will be able to create a decode node with local_cpu and it may even work, but use the caches only once

PR #1113, opened 2025-07-21
[fix]: check remote put task

sammshen:
The new `batched_put()` interface of the storage manager relies on the backends themselves to check for existence in put tasks or in teh backend (unlike previous `put()`).


chunxiaozheng:
Hi, @sammshen, I think writing twice will not happen, because `skip_leading_tokens` indicates which tokens already exist. By setting the store mask, when calling `process_tokens`, the exist keys will be skipped.

sammshen:
HI @chunxiaozheng , thank you for pointing this out! I will make sure to read/think more carefully in the future before making PR

chunxiaozheng:
@sammshen I am sorry, I found some ERROR logs when using `FSConnector`. After consideration, I believe that this PR is still necessary. For example, if there are prompts with the same prefix in a batch, the prefix chunk may be the same, which can cause problems with async put.

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -179,6 +179,9 @@ def submit_put_task(
         if self._mla_worker_id_as0_mode:
             return None
 
+        if self.exists_in_put_tasks(key) or self.contains(key):
```
chunxiaozheng: I think we do not need to call `contains`, which will add one RPC or IO.
```
if self.exists_in_put_tasks(key):
    return None
```

sammshen:
@chunxiaozheng looks like @YaoJiayi already refactored the disk path in https://github.com/LMCache/LMCache/pull/1172!

chunxiaozheng:
> @chunxiaozheng looks like @YaoJiayi already refactored the disk path in #1172!

Okay

chunxiaozheng:
lgtm

maobaolong (approved):
@sammshen Thanks for this contribution! LGTM.

maobaolong:
@sammshen Would you mind to fix the DCO issue?

sammshen:
@chunxiaozheng would you like to approve so this can be merged? :)

Shaoting-Feng (approved):

chunxiaozheng:
> @chunxiaozheng would you like to approve so this can be merged? :)

@sammshen I have approved last week, but I am not yet a committer🥹

PR #1101, opened 2025-07-18
[Bugfix] Lookup compatibility with tp

Shaoting-Feng:
1. Create a lookup server on each TP worker. The client sends lookup requests to all servers.
2. Remove redundant unpin.

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

To create a new tag for lmcache (Note: `v` prefix is required):
`git tag vx.x.x`
`git push origin vx.x.x` (same version again)

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

In case the workflow fails, delete the tag and try again:
`git tag -d vx.x.x`
`git push origin :refs/tags/vx.x.x`

For example:
`git tag -d v0.3.0`
`git push origin :refs/tags/v0.3.0`

To create a new release and publish `lmcache` Python package to PyPi:
`git remote add upstream git@github.com:LMCache/LMCache.git`
[...]

PR #803, opened 2025-06-10
Reduce space cost to 1/TP while using MLA

maobaolong:
# Background

This PR do the improvement under the result while vLLM using MLA for TP > 0, we found that the kvcache for each TP are all the same for the same token list, so I submit this PR, aimed to reduced the space cost to 1/TP.

# How to test

- start vLLM + LMCache + FSConnector v0

Check the generated files and generated output by vLLM.

```shell
LMCACHE_USE_EXPERIMENTAL=True LMCACHE_TRACK_USAGE=false \
VLLM_MLA_DISABLE=0 VLLM_USE_V1=0 LMCACHE_CONFIG_FILE=./lmcache.yaml \
vllm serve /disc/data1/deepseek/DeepSeek-V2-Lite-Chat/ \
           --trust-remote-code \
           --served-model-name vllm_cpu_offload \
           --max-model-len 32768 \
           --max-seq-len-to-capture 10000 \
           --max-num-seqs 64 \
           --gpu-memory-utilization 0.9 \
           --host 0.0.0.0 \
           -tp 2 \
           --kv-transfer-config '{"kv_connector":"LMCacheConnector","kv_role":"kv_both","kv_parallel_size":2}'
```

- start vLLM + LMCache + FSConnector v0
```shell
LMCACHE_USE_EXPERIMENTAL=True LMCACHE_TRACK_USAGE=false \
VLLM_MLA_DISABLE=0 VLLM_USE_V1=1 LMCACHE_CONFIG_FILE=./lmcache.yaml \
vllm serve /disc/data1/deepseek/DeepSeek-V2-Lite-Chat/ \
           --trust-remote-code \
           --served-model-name vllm_cpu_offload \
           --max-model-len 32768 \
           --max-seq-len-to-capture 10000 \
           --max-num-seqs 64 \
           --gpu-memory-utilization 0.9 \
           --host 0.0.0.0 \
           -tp 2 \
           --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both","kv_parallel_size":2}'
```

- lmcache.yaml

```yaml
chunk_size: 256
local_device: "cpu"
remote_url: "fs://host:0/disc/data1/baoloongmao/fs_data/"
#remote_serde: "cachegen"
remote_serde: "naive"
# Whether retrieve() is pipelined or not
pipelined_backend: False
local_cpu: False
max_local_cpu_size: 10
extra_config:
  remote_enable_mla_worker_id_as0: True
```

- Without this PR
```
[...]

PR #1128, opened 2025-07-23
[VLLM|CacheEngine]support only save on rank 0 for DS MLA

maobaolong:
# Highlight

This feature is credit to @xshwu-ai, I help to do a refactor and submit this PR.

Please add the following great co-author of the following before merge.
- xshwu <xshwu@tencent.com>
- Yihua Cheng <yihua98@uchicago.edu

# Motivation

This is an enhance compare to #803 , it not only work for remote connector but also work to any other backend.

For deepseek MLA, this improvement can reduce transfer data from backend, instead, it transfer data between worker0 to workerN. 

# Description

This make the retrieve function broadcast data from the first rank to other ranks in distributed setups when "save_only_first_rank" is set, ensuring only one rank does the heavy lifting and others stay in sync via efficient communication. This design helps reduce storage and IO redundancy across multiple processes.

We added a "save_only_first_rank" mode to the LMCacheEngine, which changes how the retrieve function works in distributed environments:

- Conditional Retrieval and Broadcasting

When "save_only_first_rank" is enabled, only the first rank performs the actual retrieval and broadcasts the relevant metadata and tensors to other ranks.
Other ranks receive the starts/ends, metadata, and tensors via the broadcast functions instead of retrieving from storage themselves.

- Broadcast Logic

On the first rank:
After collecting the needed segments, it broadcasts the starts/ends list and then, for each segment, broadcasts metadata and tensor data to other ranks.
There are safety checks for list length consistency and warnings if mismatches occur.
On non-first ranks:
They receive the starts/ends, then in a loop, receive metadata and tensor data for each segment, reconstructing the memory objects accordingly.

- Removal and Prefetch Consistency

Removal of data after retrieval and prefetching now only happen on the first rank when "save_only_first_rank" is enabled, preventing redundant operations on other ranks since other rank have no data at all.

# How to test

- lmcache.yaml
```yaml
chunk_size: 256
local_cpu: False
max_local_cpu_size: 5
remote_serde: "naive"
pipelined_backend: False
remote_url: "fs://host:0/disc/data1/baoloongmao/fs_data/"
save_unfull_chunk: False
extra_config:
  save_only_first_rank: True
  first_rank_max_local_cpu_size: 50
```

It can be found from `top` that the used memory for the two worker process is different, this is expected.

image

- Start vllm
```bash
export PROMETHEUS_MULTIPROC_DIR="/tmp/lmcache_prometheus" 
CUDA_VISIBLE_DEVICES=6,7 \
LMCACHE_USE_EXPERIMENTAL=True LMCACHE_TRACK_USAGE=false \
VLLM_MLA_DISABLE=0 VLLM_USE_V1=1 LMCACHE_CONFIG_FILE=./lmcache.yaml \
vllm serve /disc/data1/deepseek/DeepSeek-V2-Lite-Chat/ \
           --trust-remote-code \
           --served-model-name vllm_cpu_offload \
           --max-model-len 32768 \
           --no-enable-prefix-caching \
           --max-seq-len-to-capture 10000 \
           --max-num-seqs 64 \
           --host 0.0.0.0 \
           -tp 2 \
           --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both","kv_parallel_size":2}'
```

- test 
Run the following command twice, one for trigger store, the other for trigger load.
```
curl http://localhost:8000/v1/chat/completions     -H "Content-Type: application/json"     -d '{
    "model": "vllm_cpu_offload",
    "messages": [{"role": "user", "content": "Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?"}],
    "max_tokens": 10,
    "temperature": 0,
    "top_p": 0.95
}'
{"id":"chatcmpl-f3a4870c42254680918a8c3844df230b","object":"chat.completion","created":1753340706,"model":"vllm_cpu_offload","choices":[{"index":0,"message":{"role":"assistant","content":" Hello! As an AI, I don't","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning_content":null},"logprobs":null,"finish_reason":"length","stop_reason":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":307,"total_tokens":317,"completion_tokens":10,"prompt_tokens_details":null},"prompt_logprobs":null,"kv_transfer_params":null}
```

- check chunk file
There is only 1 file belong to tp 0, without any chunk file belong to other tp, that is expected.
```
[root@TENCENT64 cpu]# ll -h  ../fs_data/
total 7.6M-rw-r--r-- 1 root root 7.6M Jul 24 15:05 vllm@-disc-data1-deepseek-DeepSeek-V2-Lite-Chat-@2@0@6888939885391290423.data
```

- check from metrics
```
[root@TENCENT64 cpu]# curl -s http://localhost:8000/metrics/ |grep -v \# | grep lmcache | grep num_remote_read_bytes_total
lmcache:num_remote_read_bytes_total{model_name="/disc/data1/deepseek/DeepSeek-V2-Lite-Chat/",worker_id="1"} 0.0
lmcache:num_remote_read_bytes_total{model_name="/disc/data1/deepseek/DeepSeek-V2-Lite-Chat/",worker_id="0"} 7.962624e+06
```

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1,7 +1,17 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
 from collections import defaultdict
-from typing import Dict, Generator, List, Optional, OrderedDict, Tuple, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    OrderedDict,
+    Tuple,
+    Union,
+)
 import asyncio
 import gc
 import multiprocessing
@@ -33,8 +43,10 @@ from lmcache.v1.memory_management import (  # noqa: E501
     MemoryAllocatorInterface,
     MemoryFormat,
     MemoryObj,
+    MemoryObjMetadata,
     MixedMemoryAllocator,
     NixlCPUMemoryAllocator,
+    TensorMemoryObj,
 )
 from lmcache.v1.storage_backend.storage_manager import StorageManager
 from lmcache.v1.token_database import (
@@ -74,6 +86,8 @@ class LMCacheEngine:
         memory_allocator: MemoryAllocatorInterface,
         token_database: TokenDatabase,
         gpu_connector: GPUConnectorInterface,
+        broadcast_fn: Callable[[torch.Tensor, int], None],
+        broadcast_object_fn: Callable[[Any, int], Any],
     ):
         logger.info(f"Creating LMCacheEngine with config: {config}")
         self.config = config
@@ -81,6 +95,16 @@ class LMCacheEngine:
         self.memory_allocator = memory_allocator
         self.token_database = token_database
         self.gpu_connector = gpu_connector
+        self.broadcast_fn = broadcast_fn
+        self.broadcast_object_fn = broadcast_object_fn
+        save_only_first_rank_default = True if metadata.use_mla else False
+        self.save_only_first_rank = (
+            self.config.extra_config.get(
+                "save_only_first_rank", save_only_first_rank_default
+            )
+            if self.config.extra_config
+            else save_only_first_rank_default
+        )
 
         self.enable_p2p = config.enable_p2p
 
@@ -184,6 +208,9 @@ class LMCacheEngine:
         :raises: ValueError if the number of Falses in the mask is not a
             multiple of the chunk size.
         """
+        if self._is_passive():
+            logger.debug(f"rank={self.metadata.worker_id} ignore store")
+            return
 
         if mask is not None:
             num_to_store_tokens = torch.sum(mask).item()
@@ -427,90 +454,19 @@ class LMCacheEngine:
 
         ret_mask = torch.zeros_like(tokens, dtype=torch.bool, device="cpu")
 
-        # location -> [(CacheEngineKey, start, end)]
-        block_mapping: Dict[str, List[Tuple[CacheEngineKey, int, int]]] = defaultdict(
-            list
-        )
-        # [(CacheEngineKey, MemoryObj, start, end)]
         reordered_chunks: List[Tuple[CacheEngineKey, MemoryObj, int, int]] = []
-
-        tags = kwargs.get("tags")
-        if tags is not None and len(tags) != 0:
-            assert isinstance(tags, OrderedDict)
-
-        for start, end, key in self.token_database.process_tokens(
-            tokens=tokens,
-            mask=mask,
-            tags=tags,
-        ):
-            assert isinstance(key, CacheEngineKey)
-
-            if key in self.lookup_cache:
-                # TODO(Jiayi): we can reduce the number of `contains` calls
-                # by checking the lookup cache first (should be updated in `lookup`)
-                pass
-            else:
-                # NOTE: key should always be in the lookup cache once
-                # we support it.
-                location = self.storage_manager.contains(key)
-                if location is None:
-                    # TODO(Jiayi): Need to refactor P2P as a storage backend to
-                    # clean up the following code.
-                    if self.enable_p2p:
-                        future_memory_obj = asyncio.run_coroutine_threadsafe(
-                            self.distributed_server.issue_get(key),
-                            self.distributed_loop,
-                        )
-                        memory_obj = future_memory_obj.result()
-                        if memory_obj:
-                            reordered_chunks.append((key, memory_obj, start, end))
-                            ret_mask[start:end] = True
-                        else:
-                            # NOTE: break for P2P retrieve KV because of no required
-                            # memory obj
-                            break
-                        continue
-                    break
-
-                # NOTE: Here we make the assumption that the underlying
-                # storage backend support pin operation, and the memory
-                # object is already pinned in the storage backend.
-                ret_mask[start:end] = True
-
-            assert location is not None
-
-            block_mapping[location].append((key, start, end))
-
-        # TODO(Jiayi): We can parallelize the retrieval from
-        # different storage backends.
-        last_failed_block_start = None
-        for location, blocks in block_mapping.items():
-            keys = [key for key, _, _ in blocks]
-            memory_objs = self.storage_manager.batched_get(
-                keys=keys,
-                location=location,
+        if not self._is_passive():
+            reordered_chunks = self._process_tokens_internal(
+                tokens,
+                mask,
+                ret_mask,
+                **kwargs,
             )
-            for (key, start, end), memory_obj in zip(blocks, memory_objs, strict=False):
-                if memory_obj is None:
-                    logger.warn(
-                        "The cache block is in the storage, but it can't be retrieved"
-                    )
-                    if (
-                        last_failed_block_start is None
-                        or last_failed_block_start < start
-                    ):
-                        last_failed_block_start = start
-                    break
-                reordered_chunks.append((key, memory_obj, start, end))
-
-        if last_failed_block_start is not None:
-            ret_mask[last_failed_block_start:] = False
-
-            reordered_chunks = [
-                (key, memory_obj, start, end)
-                for key, memory_obj, start, end in reordered_chunks
-                if end < last_failed_block_start
-            ]
+        if self.save_only_first_rank:
+            self._broadcast_or_receive_memory_objs(
+                reordered_chunks,
+                ret_mask,
+            )
 
         # NOTE(Jiayi): memory_obj doesn't have to be a pinned
         # cpu tensor for the sake of performance.
@@ -524,7 +480,7 @@ class LMCacheEngine:
 
         # TODO(Jiayi): Remove the following for loop with batched operations
         for key, memory_obj, _, _ in reordered_chunks:
-            if self.remove_after_retrieve:
+            if self.remove_after_retrieve and not self._is_passive():
                 self.storage_manager.remove(key)
             memory_obj.ref_count_down()
 
@@ -663,6 +619,8 @@ class LMCacheEngine:
         """Launch the prefetching process in the storage manager to load the
         KV to the local CPU memory
         """
+        if self._is_passive():
+            return
         for start, end, key in self.token_database.process_tokens(
             tokens=tokens, mask=mask, tags=tags
         ):
@@ -883,6 +841,22 @@ class LMCacheEngine:
         locations: Optional[List[str]] = None,
         tags: OrderedDict = None,  # TODO: need to clean by tags
     ) -> int:
+        if self.save_only_first_rank:
+            if self.metadata.is_first_rank():
+                num_removed = self._clear(tokens, locations)
+                self.broadcast_object_fn(num_removed, self.metadata.first_rank)
+                return num_removed
+            else:
+                num_removed = self.broadcast_object_fn(None, self.metadata.first_rank)
+                return int(num_removed)
+        return self._clear(tokens, locations)
+
+    def _clear(
+        self,
+        tokens: Optional[Union[torch.Tensor, List[int]]] = None,
+        locations: Optional[List[str]] = None,
+        tags: OrderedDict = None,  # TODO: need to clean by tags
+    ) -> int:
         assert isinstance(self.storage_manager, StorageManager)
         # Clear all caches if tokens is None
         if tokens is None or len(tokens) == 0:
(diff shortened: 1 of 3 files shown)


lmcache/v1/cache_engine.py
```diff
@@ -429,69 +441,143 @@ def retrieve(
...
-        # different storage backends.
-        for location, keys in key_mapping.items():
-            memory_objs = self.storage_manager.batched_get(
-                keys=keys,
-                storage_backend_name=location,
-            )
-            reordered_memory_objs.extend(memory_objs)
-            reordered_keys.extend(keys)
-            reordered_starts.extend(start_mapping[location])
-            reordered_ends.extend(end_mapping[location])
```
maobaolong: The git diff is not so smart, the true diff for this code block is that I just only add a `if not self.save_only_first_rank or self.metadata.is_first_rank():` outside, so i have to add a tab for each line.

lmcache/v1/cache_engine.py
```diff
@@ -119,11 +130,7 @@ def __init__(
         )
 
         # HACK: remove this in the future
-        # NOTE (Jiayi): This is currently used to support
-        # dropping the kv cache in nixl backend at decoder.
-        self.remove_after_retrieve = (
-            config.enable_nixl and config.nixl_role == "receiver"
-        )
+        self.remove_after_retrieve = config.enable_nixl
```
Shaoting-Feng: `config.nixl_role == "receiver"` is for making cpu offload and disaggregated prefill compatible.

lmcache/v1/cache_engine.py
```diff
@@ -807,6 +911,19 @@ def _Create_memory_allocator(
             return CuFileMemoryAllocator(config.cufile_buffer_size * 1024**2)
 
         max_local_cpu_size = config.max_local_cpu_size
+        if config.extra_config:
+            save_only_first_rank = config.extra_config.get(
+                "save_only_first_rank", False
```
Shaoting-Feng: I suggest we set this as default behavior for MLA models

lmcache/v1/cache_engine.py
```diff
@@ -807,6 +911,19 @@ def _Create_memory_allocator(
...
+            if save_only_first_rank and metadata.is_first_rank():
+                # Only the first rank will save the cache,
+                # so we need to set it lager than other ranks
+                first_rank_max_local_cpu_size = config.extra_config.get(
+                    "first_rank_max_local_cpu_size", max_local_cpu_size
+                )
+                return MixedMemoryAllocator(
+                    int(first_rank_max_local_cpu_size * 1024**3)
+                )
         return MixedMemoryAllocator(int(max_local_cpu_size * 1024**3))
```
Shaoting-Feng, line 1159: Do other rank workers still need to allocate a cpu buffer?

lmcache/v1/cache_engine.py
```diff
@@ -429,69 +439,143 @@ def retrieve(
...
+                    keys=keys,
+                    storage_backend_name=location,
+                )
+                reordered_memory_objs.extend(memory_objs)
+                reordered_keys.extend(keys)
+                reordered_starts.extend(start_mapping[location])
+                reordered_ends.extend(end_mapping[location])
+
+        if self.save_only_first_rank:
+            if self.metadata.is_first_rank():
```
Shaoting-Feng: I think `torch.distributed.broadcast` can be called the same on both the sender and the receiver, so do we still need to check if `self.metadata.is_first_rank()`?

lmcache/v1/cache_engine.py
```diff
@@ -429,69 +439,143 @@ def retrieve(
...
+                # broadcast memory_objs
+                for i in range(len(reordered_starts)):
+                    memory_obj = reordered_memory_objs[i]
+
+                    # broadcast metadata
+                    metadata_dict = memory_obj.metadata.to_dict()
+                    self.broadcast_object_fn(metadata_dict, self.metadata.first_rank)
+
+                    # broadcast tensor
+                    tensor_to_broadcast = memory_obj.tensor.to(
```
Shaoting-Feng: I wonder if it’s possible to broadcast after calling LMCache’s kernel to transfer `memory_obj` from CPU to GPU, instead of using `tensor.to`, since it’s faster.

lmcache/v1/cache_engine.py
maobaolong: I'm not sure exactly, but if we need to decrease it, we can set `max_local_cpu_size` to a smaller value.

But the `first_rank_max_local_cpu_size` can be set to a large value.

lmcache/v1/cache_engine.py
maobaolong: This is an accidental change, revert it. @Shaoting-Feng  Thanks for point it out.

lmcache/v1/cache_engine.py
maobaolong: I agree with your assessment that trying to eliminate all is_first_rank() checks would actually make the code more complex, not less. The current implementation represents the cleanest possible balance between symmetry and practicality.
Would you like me to suggest any specific improvements to make the distinction clearer in the code?

I feel hard to abandon the `is_first_rank check` base on the following point.
1. The first rank has access to the original storage data
2. Other ranks must reconstruct objects from transmitted metadata
3. The broadcast API itself is asymmetric by design:
  - Sender provides actual data
  - Receivers provide empty containers

lmcache/v1/cache_engine.py
maobaolong: Good suggestion, done

lmcache/v1/cache_engine.py
maobaolong: @xshwu-ai Would you like to help to reply this question?

maobaolong:
@Shaoting-Feng Thanks for your review, addressed most of your comments except the last question, I invite @xshwu-ai to answer that question.

lmcache/v1/cache_engine.py
maobaolong: @Shaoting-Feng After discussed with @xshwu-ai , this might be a good improvement point, this can be achieved by a followup PR.

maobaolong:
@Shaoting-Feng To make this PR works, I have to comment the https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/lookup_client/lmcache_lookup_client.py#L70 to workaround.

Would you like to take another look at this PR and submit another PR https://github.com/LMCache/LMCache/pull/1144 to fix #1101 . Thanks!

ApostaC:
@maobaolong Thanks for the implementation! 
Two high-level concerns:
- A lot of new complexities were introduced in `cache_engine.retrieve()`.
- Broadcasting non-tensor objects may incur non-negligible overheads.

After reading and understanding the implementation, I've got a new proposal that we can implement in LocalCPUBackend. The only thing we need to change is to add a few lines in `get_blocking` and `submit_put_task`, and it can avoid broadcasting metadata.

cc @YaoJiayi

maobaolong:
@ApostaC Thanks for your review, glad you have a new proposal to solve this issue, feel free to submit a PR.

Just curious about can be only modify the `LocalCPUBackend` to benefit all the backends?

maobaolong:
@ApostaC Sorry to ping you to active this PR.

lmcache/v1/cache_engine.py
```diff
@@ -178,6 +192,9 @@ def store(
         :raises: ValueError if the number of Falses in the mask is not a
             multiple of the chunk size.
         """
+        if self.save_only_first_rank and not self.metadata.is_first_rank():
```
ApostaC: Can we add a helper function in CacheEngine with a name like `is_passive`:
```
    def is_passive(self):
        """A 'passive' CacheEngine means that the node itself will not store/retrieve 
        the data directly, but from the "active" worker (i.e., rank 0 in MLA)
        """
        return self.save_only_first_rank and not self.metadata.is_first_rank()
```

Then the if condition can be more readable:

```suggestion
        if self.is_passive():
```

lmcache/v1/cache_engine.py
```diff
@@ -409,77 +426,29 @@ def retrieve(
...
-                    start_mapping[location] = [start]
-                    end_mapping[location] = [end]
-                    continue
-
-            assert location is not None
-
-            key_mapping[location].append(key)
-            start_mapping[location].append(start)
-            end_mapping[location].append(end)
+        if not self.save_only_first_rank or self.metadata.is_first_rank():
```
ApostaC: Similar to the above suggestion, we can do:

```suggestion
        if not self.is_passive():
```

lmcache/v1/cache_engine.py
```diff
@@ -491,7 +460,9 @@ def retrieve(
 
         # TODO(Jiayi): Remove the following for loop with batched operations
         for key, memory_obj in zip(reordered_keys, reordered_memory_objs, strict=False):
-            if self.remove_after_retrieve:
+            if self.remove_after_retrieve and (
+                not self.save_only_first_rank or self.metadata.is_first_rank()
+            ):
```
ApostaC:
```suggestion
            if self.remove_after_retrieve and not self.is_passive()
```

lmcache/v1/cache_engine.py
```diff
@@ -623,6 +594,9 @@ def prefetch(
         """Launch the prefetching process in the storage manager to load the
         KV to the local CPU memory
         """
+        if self.save_only_first_rank:
+            if not self.metadata.is_first_rank():
+                return
```
ApostaC:
```suggestion
        if self.is_passive():
            return
```

lmcache/v1/cache_engine.py
```diff
@@ -857,6 +846,163 @@ def close(self) -> None:
...
+    def _process_tokens_internal(
+        self,
+        tokens,
+        mask,
+        reordered_keys,
+        reordered_memory_objs,
+        reordered_starts,
+        reordered_ends,
+        ret_mask,
+    ):
```
ApostaC: Can we add a docstring for this function to briefly describe what this function does, and what side-effect this function will have on the input parameters (`reordered_keys`, `reordered_memory_objs`, ...)

lmcache/v1/cache_engine.py
```diff
@@ -857,6 +846,163 @@ def close(self) -> None:
...
+            reordered_ends.extend(end_mapping[location])
+
+    def _broadcast_or_receive_memory_objs(
+        self,
+        reordered_keys,
+        reordered_starts,
+        reordered_ends,
+        reordered_memory_objs,
+        ret_mask,
+    ):
```
ApostaC: Also, it would be great if we could have a docstring for this function.

ApostaC (commented):
Thanks @maobaolong . This version looks much better! Just left some high-level comments.

@YaoJiayi Please take a look, especially the details in `_process_tokens_internal` and `_broadcast_or_receive_memory_objs`

lmcache/v1/cache_engine.py
maobaolong: Done

lmcache/v1/cache_engine.py
maobaolong: Done

lmcache/v1/cache_engine.py
maobaolong: done

lmcache/v1/cache_engine.py
maobaolong: done

lmcache/v1/cache_engine.py
maobaolong: done

lmcache/v1/cache_engine.py
maobaolong: done

maobaolong:
@ApostaC Thanks for your code suggestion.
I did the following changes 
- abstracted a private method `_is_passive`, simplify code.
- add docs for the new added method.

The new added `_process_tokens_internal` is abstracted from `retrieve` directly without any line changed. @YaoJiayi Just easy to skip review the content of this method.

lmcache/v1/cache_engine.py
```diff
@@ -864,6 +850,215 @@ def close(self) -> None:
...
+                    metadata = MemoryObjMetadata.from_dict(metadata_dict)
+                    tensor = torch.empty(
+                        metadata.shape,
+                        dtype=metadata.dtype,
+                        device=f"cuda:{self.metadata.worker_id}",
+                    )
+                    self.broadcast_fn(tensor, self.metadata.first_rank)
+
+                    # create temp memory obj
+                    memory_obj = TensorMemoryObj(raw_data=tensor, metadata=metadata)
```
YaoJiayi: I'm not sure about this. Why are we not allocating from the memory allocator?

YaoJiayi (commented):
LGTM:) Just one comment. Also, what's the deployment of deepseek? TP=8 on one node? Also, this could be a temporary solution as we might want to separate LMCache to another process than vllm in the future

lmcache/v1/cache_engine.py
maobaolong: @YaoJiayi Thanks for point this out!

Since the tensor here is a gpu buffer to accelerate broadcast and reduce cpu->gpu copy, so we didn't use `local_cpu_backend` to allocate the memory object for now.

But we can improve this performance by using `GPUMemoryAllocator` or other way while we the this place become a hot path.

maobaolong:
> LGTM:) Just one comment. Also, what's the deployment of deepseek? TP=8 on one node? 

The deepseek R1 we used are 3 versions. all can run in one node.
1. DeepSeek R1 AWQ 
2. DeepSeek V2 lite
3. DeepSeek R1 int4 (Our Internal model)

> Also, this could be a temporary solution as we might want to separate LMCache to another process than vllm in the future

I heard about this from @ApostaC , but not very clear how can we benefit from it?

ApostaC:
> I heard about this from @ApostaC , but not very clear how can we benefit from it?

@maobaolong I'm building a prototype for this. Hopefully can get some numbers for this.

lmcache/v1/cache_engine.py
```diff
@@ -874,6 +842,186 @@ def close(self) -> None:
 
         logger.info("LMCacheEngine closed.")
 
+    def _process_tokens_internal(
```
maobaolong, line 901: @YaoJiayi  This method is abstracted from `retrieve` method without modify any content.

maobaolong:
Sorry to ping @YaoJiayi 🙏

lmcache/v1/cache_engine.py
```diff
@@ -74,13 +76,25 @@ def __init__(
         memory_allocator: MemoryAllocatorInterface,
         token_database: TokenDatabase,
         gpu_connector: GPUConnectorInterface,
+        broadcast_fn: Callable[[torch.Tensor, int], None],
```
YaoJiayi, line 89: Curious why we pass these communication related functions as init arguments.

lmcache/v1/cache_engine.py
```diff
@@ -879,6 +847,186 @@ def close(self) -> None:
...
+            # Broadcast each chunk's data
+            for key, memory_obj, start, end in reordered_chunks:
+                # Combine (start, end) and metadata into single broadcast
+                metadata_dict = memory_obj.metadata.to_dict()
+                combined_metadata = (start, end, metadata_dict)
+                self.broadcast_object_fn(combined_metadata, self.metadata.first_rank)
+
+                # Broadcast tensor data
+                tensor_to_broadcast = memory_obj.tensor.to(
+                    f"cuda:{self.metadata.worker_id}"
```
YaoJiayi, line 1041: Raising a performance concern here: are other ranks waiting for this pice transfer? If so, this on-gpu broadcast seems to be redundant. A better approach would be using an on-cpu shared memory broadcast or unifying all local lmcache workers into a single process.

We can postpone handling this though.

YaoJiayi (approved):
Approved it for now. But we need to change the code in the future.

lmcache/v1/cache_engine.py
maobaolong: Because I want to keep `cache_engine` inference engine independent.

maobaolong:
@YaoJiayi Thanks for your review! @ApostaC Would you like to take a double check?

lmcache/v1/cache_engine.py
ApostaC: Agreed. Otherwise, we have to `import vllm` in `cache_engine.py` (which we don't want for now)

ApostaC (approved):
Thanks! LGTM!

One side note: now we have the comprehensive test in CI. It would be nice if we can add a bunch of new tests in CI (not limited to this functionality).

PR #1133, opened 2025-07-23
[Bugfix] Decode in PD never makes a cache hit and causes OOM

vladnosiv:
After the merge of PR https://github.com/LMCache/LMCache/pull/1040, the PD-disaggregation broke down. 

Due to the replacement of deterministic hashing with the built-in hash function from python, prefill and decode now receive different hash values of the same blocks. This is mainly due to the use of NONE_HASH from vLLM, which differs from launch to launch. This leads to the fact that decode does not see the blocks that have arrived to it. In addition to recalculating anew, this entails OOM due to overflow of the NiXL buffer, from which nothing is deleted, since deletion is used during retrieving.

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -152,7 +152,7 @@ class ChunkedTokenDatabase(TokenDatabase):
             # Standard
             import os
 
-            if config.remote_url is not None:
+            if config.remote_url is not None or config.enable_nixl:
                 pythonhashseed = os.getenv("PYTHONHASHSEED")
                 if pythonhashseed is None:
                     logger.warning(
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -136,10 +136,14 @@ class ChunkedTokenDatabase(TokenDatabase):
         config: Optional[LMCacheEngineConfig] = None,
         metadata: Optional[LMCacheEngineMetadata] = None,
     ):
-        super().__init__(config, metadata)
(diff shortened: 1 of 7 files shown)


Shaoting-Feng (approved):
LGTM

vladnosiv:
@Shaoting-Feng thank you for your review

Unfortunately, in another PR https://github.com/LMCache/LMCache/pull/1140, a similar problem was solved with the help of validation and documentation. I'm still convinced that for a PD case, this should be solved automatically, it looks pretty crutchy. However, I have adjusted the PR to take into account the new introductory, I will be glad if I can see it.

@sammshen I'm also requesting a review, because you conducted a PR https://github.com/LMCache/LMCache/pull/1140 review.

lmcache/v1/token_database.py
```diff
@@ -128,7 +152,7 @@ def __init__(
             # Standard
             import os
 
-            if config.remote_url is not None:
+            if config.remote_url is not None or config.enable_nixl:
                 pythonhashseed = os.getenv("PYTHONHASHSEED")
                 if pythonhashseed is None:
                     logger.warning(
```
sammshen, line 144: can you add "Prefill Disaggregation is detected" or something like that to the warning?

sammshen:
LGTM! Thanks for noticing and adapting to https://github.com/LMCache/LMCache/pull/1140 ! tiny comment (please change) but approving first

sammshen (approved):

vladnosiv:
@sammshen 
it also looks like an environment error in unit tests, PR is ready for merging i think

Issue #1060, opened 2025-07-15
LMCacheEngine retrieve does not check None results of storage_manager.batched_get


llc-kc:
In older version, LMCacheEngine retrieve use storage_manager.get and check None result and break. 
Current version does not consider this condition. Actually this can happen with low probability when no memory can be allocated (memory are used by store, prefetch, etc.), and gpu_connector.to_gpu will raise exception "assert memory_obj.tensor is not None" and crash.
One possible solution is check none results of storage_manager.batched_get and another solution is limite the usage of store and prefetch when memory is not too much. Or just fallback to 0.3.1.post1 version.

sammshen:
can close if #1134 merged

PR #1134, opened 2025-07-23
[Core] Make PD and disk offloading compatible 

vladnosiv:
Nothing PD-specific was required for compatibility with Disk Storage. It simply unlocks the option in the config.

However, during load tests, I found that the server crashes when it sees blocks in DiskStorage through contains, but cannot extract them from there, because there is not enough space for them in CpuStorage due to pinned blocks. In this case, a crash occurs, because memory_obj = None is returned.

So I added more tolerance to falls by dropping the entire suffix that could not be extracted. 

In fact, it's worth making a more reliable mechanism for pinning blocks from disk during lookup, since there will now be quite a lot of warnings in the logs under load. However, it's worth leaving a more tolerant code in any case, because if the storage is remote, it can respond with an error even with strict pins too (due to a connection failure, for example)


vladnosiv:
Resolves https://github.com/LMCache/LMCache/issues/1060

Shaoting-Feng:
Thanks for your PR. Could you show an example with both PD and disk offloading enabled? Thanks!

vladnosiv:
@Shaoting-Feng 

### Configuration
if you apply patches https://github.com/LMCache/LMCache/pull/1133 and https://github.com/LMCache/LMCache/pull/1162, then it is enough to change the config in 1p1d_experimental to run:

```yaml
local_cpu: True
max_local_cpu_size: 10
 
local_disk: file:///path/to/your/LMCache/examples/disagg_prefill/1p1d_experimental/ssd_cache
max_local_disk_size: 100

enable_nixl: True
enable_xpyd: True
nixl_role: "sender"
nixl_proxy_host: "localhost"
nixl_proxy_port: 7500
nixl_buffer_size: 10737418240 # 10GB
nixl_buffer_device: "cuda"
```
(and similar nixl_buffer_size on decoder side)

I tested with the Qwen/Qwen3-32B-FP8 model with TP1 and gpu-memory-utilization 0.65 and this bench:
```shell
python -m sglang.bench_serving \
   --backend vllm \
   --port 9100 \
   --model Qwen/Qwen3-32B-FP8 \
   --max-concurrency 10  \
   --dataset-name generated-shared-prefix \
   --gsp-output-len 2 
```

### Without RAM, without SSD

```
Backend:                                 vllm      
Traffic request rate:                    inf       
Max request concurrency:                 10        
Successful requests:                     1024      
Benchmark duration (s):                  229.83    
Total input tokens:                      2348469   
Total generated tokens:                  2048      
Total generated tokens (retokenized):    1877      
Request throughput (req/s):              4.46      
Input token throughput (tok/s):          10218.16  
Output token throughput (tok/s):         8.91      
Total token throughput (tok/s):          10227.07  
Concurrency:                             9.97      
Mean TTFT (ms):                          2164.95   
Median TTFT (ms):                        2302.46   
P99 TTFT (ms):                           2718.67   
```

### With 10GB RAM, Without SSD

```
Backend:                                 vllm      
Traffic request rate:                    inf       
Max request concurrency:                 10        
Successful requests:                     1024      
Benchmark duration (s):                  193.67    
Total input tokens:                      2348469   
Total generated tokens:                  2048      
Total generated tokens (retokenized):    1874      
Request throughput (req/s):              5.29      
Input token throughput (tok/s):          12126.04  
Output token throughput (tok/s):         10.57     
Total token throughput (tok/s):          12136.62  
Concurrency:                             9.96      
Mean TTFT (ms):                          1800.84   
Median TTFT (ms):                        1612.13   
P99 TTFT (ms):                           2462.91   
```

### With 10GB RAM, With 100GB SSD

```
Backend:                                 vllm      
Traffic request rate:                    inf       
Max request concurrency:                 10        
Successful requests:                     1024      
Benchmark duration (s):                  68.85     
Total input tokens:                      2348469   
Total generated tokens:                  2048      
Total generated tokens (retokenized):    1809      
Request throughput (req/s):              14.87     
Input token throughput (tok/s):          34109.36  
Output token throughput (tok/s):         29.75     
Total token throughput (tok/s):          34139.10  
Concurrency:                             9.97    
Mean TTFT (ms):                          556.04    
Median TTFT (ms):                        424.87    
P99 TTFT (ms):                           2455.18   
```

### Important

There are a lot of  errors now that the expected number of tokens could not be extracted due to the features of contains(pin=True) for LocalDiskBackend, let's discuss this here: https://github.com/LMCache/LMCache/pull/1172#discussion_r2235731224

lmcache/v1/cache_engine.py
```diff
@@ -421,14 +422,9 @@ def retrieve(
 
         ret_mask = torch.zeros_like(tokens, dtype=torch.bool, device="cpu")
 
-        key_mapping: Dict[str, List[CacheEngineKey]] = {}
-        start_mapping: Dict[str, List[int]] = {}
-        end_mapping: Dict[str, List[int]] = {}
+        block_mapping: Dict[str, List[Tuple[CacheEngineKey, int, int]]] = {}
```
sammshen: let's use a `defaultdict(list)`

lmcache/v1/cache_engine.py
```diff
@@ -469,40 +462,59 @@ def retrieve(
                 # object is already pinned in the storage backend.
                 ret_mask[start:end] = True
 
-                if location not in key_mapping:
-                    key_mapping[location] = [key]
-                    start_mapping[location] = [start]
-                    end_mapping[location] = [end]
-                    continue
+                if location not in block_mapping:
```
sammshen: delete if choose to use `defaultdict(list)`

lmcache/v1/cache_engine.py
```diff
@@ -421,14 +422,9 @@ def retrieve(
 
         ret_mask = torch.zeros_like(tokens, dtype=torch.bool, device="cpu")
 
-        key_mapping: Dict[str, List[CacheEngineKey]] = {}
-        start_mapping: Dict[str, List[int]] = {}
-        end_mapping: Dict[str, List[int]] = {}
+        block_mapping: Dict[str, List[Tuple[CacheEngineKey, int, int]]] = {}
+        reordered_blocks: List[Tuple[CacheEngineKey, MemoryObj, int, int]] = []
```
sammshen: suggest rename `reordered_chunks` to avoid confusion between vllm blocks and lmcache chunks

lmcache/v1/cache_engine.py
sammshen: the types are obvious but maybe just for future readability
```python
        # location -> [CacheEngineKey, start, end]
        block_mapping: Dict[str, List[Tuple[CacheEngineKey, int, int]]] = {}
        # [(CacheEngineKey, MemoryObj, start, end)]
        reordered_blocks: List[Tuple[CacheEngineKey, MemoryObj, int, int]] = []
```

lmcache/v1/cache_engine.py
```diff
@@ -469,40 +462,59 @@ def retrieve(
...
+                    logger.warn(
+                        "The cache block is in the storage, but it can't be retrieved"
+                    )
+                    ret_mask[start:end] = False
+                    if (
+                        last_failed_block_start is None
+                        or last_failed_block_start < start
+                    ):
+                        last_failed_block_start = start
+                    continue
```
sammshen: we can break out here and move onto the next location since we have the assumption that the starts are monotonocially increasing

lmcache/v1/cache_engine.py
```diff
@@ -469,40 +462,59 @@ def retrieve(
...
-            reordered_memory_objs.extend(memory_objs)
-            reordered_keys.extend(keys)
-            reordered_starts.extend(start_mapping[location])
-            reordered_ends.extend(end_mapping[location])
+            for (key, start, end), memory_obj in zip(blocks, memory_objs, strict=False):
+                if memory_obj is None:
+                    logger.warn(
+                        "The cache block is in the storage, but it can't be retrieved"
+                    )
+                    ret_mask[start:end] = False
```
sammshen: delete since already do:
```python
            ret_mask[last_failed_block_start:] = False
```
later

sammshen:
this LGTM in general, just added some pedantic comments. My main concern is how many extra python collections are being made which may add some overhead. this is a good fix to https://github.com/LMCache/LMCache/issues/1060

lmcache/v1/cache_engine.py
```diff
@@ -469,40 +462,59 @@ def retrieve(
...
+            reordered_blocks = filtered_reordered_blocks
 
         # NOTE(Jiayi): memory_obj doesn't have to be a pinned
         # cpu tensor for the sake of performance.
         # For example, disk->gpu is faster than disk->cpu->gpu.
         # RDMA is another example.
-        self.gpu_connector.batched_to_gpu(
-            reordered_memory_objs, reordered_starts, reordered_ends, **kwargs
-        )
+        memory_objs, starts, ends = [], [], []
```
sammshen: can we use:
```python
if reordered_list: 
    _, memory_objs, starts, ends = zip(*reordered_blocks) 
    self.gpu_connector.batched_to_gpu(list(memory_objs), list(starts), list(ends), **kwargs)
```

1. check for empty reordered_blocks
2. avoid repeated calls to `.append()` which IIRC is not very fast

vladnosiv:
@sammshen Thanks for the review!
I have resolved the comments.

Regarding `batched_to_gpu`, I generally have a desire to make an object with four fields (key, mem_obj, start, end) and use it everywhere, in which case it will be safer and more readable than with a tuple and several arrays as arguments.
But since this requires changes to the problem-independent code, I can bring it back later as a refactoring.

sammshen (approved):

vladnosiv:
The failed test works with the `store`, although the PR changes affect the `retrieve`. In addition, the tests are currently taking place in my environment, it looks like a flaky test.

sammshen:
you can delete the bad unit tests

vladnosiv:
@sammshen can you merge PR? I don't think this is a problem with the test itself, but rather with the environment, so I don't want to delete it

sammshen:
@vladnosiv unit tests are fixed now, just waiting for another reviewer to see if this is a good change

vladnosiv:
@sammshen great !
that sounds like a great reason to watch the second PR #1162 :)

Shaoting-Feng (approved):

PR #1140, opened 2025-07-24
[Bugfix] Cross-process cache sharing

### LINKED ISSUE #1139: [Bug] Cross-process cache sharing fails due to non-deterministic NONE_HASH initialization
## **Describe the bug**

LMCache's cross-process cache sharing functionality fails when multiple vLLM instances are started as separate processes. Cache hits are always 0 between different processes, even when using deterministic hash algorithms like `sha256`. The root cause is that vLLM's `NONE_HASH` uses random initialization (`os.urandom(32)`) when `PYTHONHASHSEED` is not set, causing different initial prefix hash values across processes.

## **To Reproduce**

1. Follow the guide in `examples/kv_cache_reuse/share_across_instances/centralized_sharing`. Start `lmcache_server localhost 65432`

2. Configure `example.yaml` with `pre_caching_hash_algorithm: "sha256"`

3. Start first vLLM instance at port 8000

4. Start second vLLM instance at port 8001

5. Send identical requests to both instances

6. See unexpected behavior: Second instance shows `LMCache hit tokens: 0` instead of cache hit

## **Expected behavior**

The second vLLM instance should successfully retrieve cached KV data from the first instance, showing `LMCache hit tokens: X` where X > 0, demonstrating successful cross-process cache sharing.

### Problem Source

In vLLM's `/vllm/v1/core/kv_cache_utils.py` (lines 55-60):

```python
# Each process generates a different random [NONE_HASH] using os.urandom(32)
NONE_HASH = int.from_bytes(os.urandom(32), byteorder="big") if os.getenv(
    "PYTHONHASHSEED") is None else sha256(os.getenv("PYTHONHASHSEED"))
```

### Evidence from Logs

**Process 8000:**

```
Hash input - prefix_hash: 59570484063868875693259937106251691136066452055412420178591257427603749464067
Hash result: 78817873630196241307793693528316268995342581910346703718095370162059851014422
```

**Process 8001:**

```
Hash input - prefix_hash: 47338760816776868545644811062890465917290878847102770972024822303967100458324
Hash result: 22338767528131956234408931604220411660519024494279578697438002082001871849283
```

Different initial prefix_hash values → Different hash chains → Incompatible cache keys

###

junl666:

FIX #1139 

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

`git tag vx.x.x`

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

`git tag -d vx.x.x`

For example:
`git tag -d v0.3.0`

For example:

> [!TIP]

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -28,6 +28,9 @@ from lmcache.v1.config import LMCacheEngineConfig
 
 logger = init_logger(__name__)
 
+# NOTE: vLLM's NONE_HASH uses random values,
+# but it is not suitable for LMCache's cross-process cache sharing.
+# We import it for compatibility but override it in _get_init_hash().
 try:
     # Third Party
     from vllm.v1.core.kv_cache_utils import NONE_HASH
@@ -136,7 +139,7 @@ class ChunkedTokenDatabase(TokenDatabase):
         )
 
     def _get_init_hash(self) -> int:
-        return NONE_HASH
+        return 0  # Override vLLM's random NONE_HASH for cross-process cache sharing.
 
     def _hash(
         self,
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -28,9 +28,8 @@ from lmcache.v1.config import LMCacheEngineConfig
 
 logger = init_logger(__name__)
 
-# NOTE: vLLM's NONE_HASH uses random values,
-# but it is not suitable for LMCache's cross-process cache sharing.
-# We import it for compatibility but override it in _get_init_hash().
+# NOTE: For centralized cache sharing, ensure PYTHONHASHSEED is
+# set consistently across all processes (e.g., export PYTHONHASHSEED=0).
 try:
     # Third Party
     from vllm.v1.core.kv_cache_utils import NONE_HASH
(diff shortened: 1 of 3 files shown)


vladnosiv:
This solution does not help for python < 3.12, since up to this point the hash(None) depends on the address of the None object in memory, so it will be different from run to run. 
But I totally agree with the problem, it completely breaks PD now.

https://github.com/python/cpython/issues/99540
https://github.com/vllm-project/vllm/security/advisories/GHSA-rm76-4mrf-v9r8

lmcache/v1/token_database.py
```diff
@@ -136,7 +139,7 @@ def _make_key_by_hash(self, chunk_hash: int):
         )
 
     def _get_init_hash(self) -> int:
-        return NONE_HASH
+        return 0  # Override vLLM's random NONE_HASH for cross-process cache sharing.
```
vladnosiv: there is in the `_hash` function below `hash_func((prefix_hash, tokens_tuple, None))`

and if prefix_hash is not None, but 0 at the start, it does not help with the third tuple object

sammshen:
same comment as from https://github.com/LMCache/LMCache/issues/1139

using NONE_HASH is deliberate and part of some efforts to make lmcache more optimized with vllm prefix caching. I think the simplest solution is for centralized case, leave a comment (or some other validation technique) to set PYTHONHASHSEED for both engines to the same thing

junl666:
> same comment as from #1139
> 
> using NONE_HASH is deliberate and part of some efforts to make lmcache more optimized with vllm prefix caching. I think the simplest solution is for centralized case, leave a comment (or some other validation technique) to set PYTHONHASHSEED for both engines to the same thing

Thanks for the clarification!  I've updated the implementation to:

1. Keep using  `NONE_HASH`
2. **Update `README`** about setting `PYTHONHASHSEED` for centralized kv cache sharing scenarios
3. **Add warning logs** when it is detected without `PYTHONHASHSEED`

YaoJiayi (approved):
LGTM

PR #1144, opened 2025-07-24
Introduce create_lookup_server_only_on_worker_0 extra config

maobaolong:
This PR aim to and has been tested can make #1128 can work. 

Not sure the motivation or what bug #1101 trying to fix, but it really introduce the following downside.
1. Comparing to the previous behavior, after #1101, one lookup become TP lookup, it could bad for performance while we using a FSConnector as a remote backend which is slower than other backend.
2. For #1128, it is a enhance for DS MLA, as the chunk for each TP are all the same, so we can only let tp0 to save chunk, and let tp0 to read chunk but broadcast to other tp, so there are all chunk within tp0, but no chunk in other tp. After #1101, I got an exception while lookup, so I have to submit this PR to make #1128 can work.


maobaolong:
@Shaoting-Feng Would you like to take a look at this PR? Thanks.

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -38,13 +39,26 @@
...
+        if self.create_lookup_server_only_on_worker_0:
+            socket_path = get_zmq_rpc_path_lmcache(vllm_config, rpc_port)
+            self.socket = make_zmq_socket(
+                self.ctx,
+                socket_path,
+                zmq.REQ,  # type: ignore[attr-defined]
+                bind=False,
+            )
+            return
         for tp_rank in range(self.tensor_parallel_size):
```
ApostaC: We can simplify the code here to avoid copy-pasting:

```suggestion
        ranks = self.tensor_parallel_size
        if self.create_lookup_server_only_on_worker_0:
            ranks = 1
        for tp_rank in range(ranks):
```

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -38,13 +39,26 @@
...
+    def __init__(self, vllm_config: "VllmConfig", config: LMCacheEngineConfig):
         self.encoder = MsgpackEncoder()
         self.ctx = zmq.Context()  # type: ignore[attr-defined]
         rpc_port = vllm_config.kv_transfer_config.get_from_extra_config(
             "lmcache_rpc_port", 0
         )
         self.tensor_parallel_size = vllm_config.parallel_config.tensor_parallel_size
+        self.create_lookup_server_only_on_worker_0 = (
+            config.extra_config
+            and config.extra_config.get("create_lookup_server_only_on_worker_0", True)
```
ApostaC, line 44: A separate note: I find a bunch of usage of `extra_config`. We probably should have clear documentation for it.

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -60,6 +74,11 @@ def __init__(self, vllm_config: "VllmConfig"):
     def lookup(self, token_ids: torch.Tensor, request_id: Optional[str] = None) -> int:
         token_bufs = self.encoder.encode(token_ids)
         request_id_buf = request_id.encode("utf-8")
+        if self.create_lookup_server_only_on_worker_0:
+            self.socket.send_multipart(token_bufs + [request_id_buf], copy=False)
+            resp = self.socket.recv()
+            result = int.from_bytes(resp, "big")
+            return result
         results = []
         for i in range(self.tensor_parallel_size):
```
ApostaC: We can do similar thing as above:

```suggestion
        ranks = self.tensor_parallel_size
        if self.create_lookup_server_only_on_worker_0:
            ranks = 1
        results = []
        for i in range(ranks):
```

ApostaC (commented):
Thanks! Come comments about avoiding copy-pasting codes

lmcache/v1/lookup_client/lmcache_lookup_client.py
maobaolong: good suggestion!

lmcache/v1/lookup_client/lmcache_lookup_client.py
maobaolong: Doc added within this class.

maobaolong:
@ApostaC Thanks for your suggestion, it really reduce the duplicated code and make code simplify. And I add a simply doc about the new added extra config.  PTAL.

ApostaC (approved):
Thanks for the update. Looks good to me!

Shaoting-Feng (approved):

PR #1117, opened 2025-07-22
[Core] update cache recency on lookup instead of fetching

yangdsh:
FIX #1116 

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

To create a new tag for lmcache (Note: `v` prefix is required):
`git tag vx.x.x`
`git push origin vx.x.x` (same version again)

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

In case the workflow fails, delete the tag and try again:
`git tag -d vx.x.x`
`git push origin :refs/tags/vx.x.x`

For example:
`git tag -d v0.3.0`
`git push origin :refs/tags/v0.3.0`

To create a new release and publish `lmcache` Python package to PyPi:
`git remote add upstream git@github.com:LMCache/LMCache.git`
`gh release create vx.x.x --repo LMCache/LMCache --title "vx.x.x" --notes "<Add description>"`

For example:
[...]

PR #1154, opened 2025-07-25
[opt]: evict suffixes before prefixes

### LINKED ISSUE #1159: [CI/Build][Bug] Integration test cannot run concurrently
**Describe the bug**
Integration used a fixed port 8000 and therefore cannot run concurrently. Current there are two agents on a single server.

**To Reproduce**
Start two integration test and let each agent pick up one.

**Expected behavior**
Two tests should use two separate ports.

sammshen:
taking @yangdsh 's suggestion in https://github.com/LMCache/LMCache/pull/1117 to move "touch" to lookup. also always insert suffixes before prefixes in ordered dict per request. 

Heuristic:
we only have to choose to update the LRU eviction dict ordering once per vllm step, either in lookup, store, or retrieve. lookup tokens is always a superset of the other two so we choose lookup.

as @maobaolong notes, this is especially useful for when vllm prefix cache hits but lmcache has more.

### Extra CI related piggy back: 

I don't want to make another PR so I also added some last touches to unit test (mock the pin memory allocator in `test_gpu_connector.py` in this PR. Previously this will sometimes (but not always) cause unit test to crash.

FIX https://github.com/LMCache/LMCache/issues/1159 (integration test port collision)


Shaoting-Feng:
DCO can be easily fixed after review.

Shaoting-Feng (approved):
LGTM

maobaolong:
@sammshen Thanks for this improvement, can this PR solve the prefix-cache part hit and lmcache hit tail case?

maobaolong:
@sammshen Why there are two individual different evict policy, for `local_cpu` , we use `hot_cache` with ordered dict, but for `local_disk`, we use lru evictor, is it possible to unify these two?

sammshen:
@maobaolong cannot unify because we always know how much disk can be used for kv cache but we do not know how much of the max local cpu size can be used (some of it will be used by other backends and other gpu trasnfers). the dynamic size of the hot cache forces us to use the eviction mechanism and raw OrderedDict

it will solve the lmcache tail hit case because we switch to lookup and lookup happens for the entire prompt

sammshen:
Some performance benchmarks to see if this actually does anything:

```
LMCACHE_CHUNK_SIZE=256 \
LMCACHE_LOCAL_CPU=True \
LMCACHE_MAX_LOCAL_CPU_SIZE=5.0 \
LMCACHE_USE_EXPERIMENTAL=True \
vllm serve \
    meta-llama/Llama-3.1-8B-Instruct \
    --max-model-len 32000 \
    --kv-transfer-config \

# from LMCache/LMBenchmark
./long_input_short_output_run.sh meta-llama/Llama-3.1-8B-Instruct http://localhost:8000 test_suffix_eviction 1 2 3
```
Screenshot 2025-07-31 at 12 15 50 PM

yangdsh:
It is exciting to see the huge improvements. But why is there a mismatch between number annotations and plotted dots?

sammshen:
I think the last point the numbers are squished but top is blue and bottom is actually orange. i will benchmark again, not sure why the performance got worse at 3 QPS

yangdsh:
That makes sense. It is indeed confusing because TTFT should increase when QPS increase. And TTFT should not be negative.

sammshen:
Screenshot 2025-07-31 at 9 15 55 PM
Screenshot 2025-07-31 at 9 16 06 PM

not the most clean graph again 🤔

my understanding of these two graphs is that updating on lookup instead of load on average (median gets a little worse) makes lmcache a little bit slower but removes the outlier cases where TTFT is super high from lots of thrashing. I think on a real workload (not synthetic), this will be an even better improvement because there will be more eviction without cache hit.

also part of the weird behavior (decreasing TTFT in both experiments) is that in between runs there is prefix sharing on different QPS I suspect

sammshen:
@YaoJiayi if you decide this is ok to merge, it is ready!

YaoJiayi (approved):
lgtm

yangdsh:
@sammshen Thank you for the work! Could you share the daily benchmark results with me later if you see stable improvements? Many thanks!

PR #1156, opened 2025-07-25
[bug]: free on unpin

### LINKED ISSUE #1130: the order of unpin and ref_count_down
When ref_count_down to zero, the memory_obj will be freed, but pined will not be freed.
Currently LMCacheConnectorV1Impl first call retrieve, whch call ref_count_down , and then call lmcache_engine.lookup_unpin. Should we call lookup_unpin first and then ref_count_down to correctly free the memory_obj ?
If pined memory_obj is not freed in ref_count_down , when will it be freed? unpin does not call free.

Moreover, another bug is memory_obj .valid is set to false in free, but not set to true in alloc.

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1130

PTAL: @llc-kc 

Primary: add free on unpin

Also: add stats observability for prometheus (see grafana image below)

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -360,6 +360,12 @@ class TensorMemoryObj(MemoryObj):
                 "Setting pin count back to 0 as a hack but please find the bug."
             )
             self.metadata.pin_count = 0
+        if (
+            self.meta.ref_count == 0
+            and self.parent_allocator is not None
+            and self.meta.pin_count == 0
+        ):
+            self.parent_allocator.free(self)
         return True
 
     @property
@@ -449,6 +455,12 @@ class BytesBufferMemoryObj(MemoryObj):
                 "Setting pin count back to 0 as a hack but please find the bug."
             )
             self.metadata.pin_count = 0
+        if (
+            self.meta.ref_count == 0
+            and self.parent_allocator is not None
+            and self.meta.pin_count == 0
+        ):
+            self.parent_allocator.free(self)
         return True
 
     def ref_count_up(self):


lmcache/v1/memory_management.py
```diff
@@ -449,6 +455,12 @@ def unpin(self) -> bool:
                 "Setting pin count back to 0 as a hack but please find the bug."
             )
             self.metadata.pin_count = 0
+        if (
+            self.meta.ref_count == 0
+            and self.parent_allocator is not None
+            and self.meta.pin_count == 0
```
maobaolong, line 461: As maybe there are something wrong with pin/unpin, I worry about this part too, would you like to add a metric to monitor the pinned memory object count, so that we can make sure there is none pined memory object if there is no request at all. @sammshen

sammshen:
@maobaolong good suggestion!

I added some prometheus observability.

test:
```bash

mkdir -p local/disk_test/local_disk/

LMCACHE_LOCAL_DISK="file://local/disk_test/local_disk/" vllm serve meta-llama/Llama-3.1-8B-Instruct     --kv-transfer-config     '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'

python benchmark_serving.py \
        --model meta-llama/Llama-3.1-8B-Instruct \
        --dataset-name sharegpt \
        --request-rate 5 \
        --dataset-path ShareGPT_V3_unfiltered_cleaned_split.json \
        --num-prompts 1000
```

monitoring local cpu and disk pinned bytes and usage

Screenshot 2025-07-25 at 9 05 07 PM

sammshen:
pin bytes fluctuates up and comes right back down, and stays at 0, which is currently correct but this will allow easier debugging on pinning semantics in the future

lmcache/v1/memory_management.py
sammshen: agreed, this has been a source of a few bugs in the last two weeks

sammshen:
note: since pin is a frequently done operation, this could cause more overhead for the stat logging

maobaolong:
@sammshen Thanks for this fix, this looks like a necessary bug fix, without this, the memory will leak for a moment, correct me if there something misunderstand of me.

> note: since pin is a frequently done operation, this could cause more overhead for the stat logging

Sine more overhead for stat logging, we can just only fix the critical bug fix, and introduce a good performance metrics next.

But it look like there is a thread `LMCacheStatsLogger` call `log_prometheus` every `10s`, it looks okay for the performance, WDYT?

sammshen:
I believe currently there is no bug but it is possible in the future for the order of unpin to come after ref count down, in that case, we still want to free memory. even though prometheus logging is every 10 seconds, the overhead of collecting stats is still there so I agree unpin is probably not needed for prometheus because it is a mostly internal mechanism:

https://github.com/LMCache/LMCache/pull/1173

PR #1162, opened 2025-07-26
[Core] Fix decoder memory leaks and chunked prefill compatability in PD

vladnosiv:
Fixed issues leading to hung queries (with CPP) and errors in NiXL Buffer on the Decode side.

The following script was used for testing:

```shell
python -m sglang.bench_serving \
    --backend vllm \
    --port 9100 \
    --model Qwen/Qwen3-32B-FP8 \
    --max-concurrency 10 \
    --dataset-name generated-shared-prefix \
    --gsp-output-len 2
```


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -240,18 +241,24 @@ def from_request_tracker(
         chunk_boundary = (
             cdiv(tracker.num_saved_tokens + 1, lmcache_chunk_size) * lmcache_chunk_size
         )
-        skip_save = skip_save or (
-            tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary
+        # NOTE(vladnosiv): for disagg, you cannot skip saving, as saving is a transfer
+        skip_save = tracker.disagg_spec is None and (
```
vladnosiv, line 232: setting `skip_save` to true can lead to bugs if you accidentally skip sending it to decode -- it will have to recalculate some of the blocks on its own

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -695,24 +702,21 @@ def wait_for_save(self):
...
-
-                if skip_leading_tokens == len(token_ids):
-                    continue  # skip this request
-                # Align to lmcache chunk size
-                skip_leading_tokens = (
-                    skip_leading_tokens
-                    // self._lmcache_chunk_size
-                    * self._lmcache_chunk_size
+                skip_leading_tokens = min(
+                    skip_leading_tokens, request.disagg_spec.num_transferred_tokens
```
vladnosiv, line 698: there were the following problems at this point:
* in the case of CPP, there was a duplicate sending of chunks to decode with only one deletion. This caused chunks to leak during decoding. 
* at the same time, you can not rely on skip_leading_chunks as is, because if the skip is caused by hitting the cache on the prefill, this does not mean that the necessary chunks have already been sent for decoding.

lmcache/v1/storage_backend/connector/nixl_connector_v3.py
```diff
@@ -286,7 +286,17 @@ def prepare_send(
...
         )
         remote_indexes = alloc_response.remote_indexes
-        self._blocking_send(req_id, receiver_id, local_indexes, remote_indexes)
+
+        # NOTE (vladnosiv): len(local_indexes) may be zero
+        # if the requests in the batch have a large common prefix
+        if len(local_indexes) == 0:
+            logger.debug(
+                "Sending objs with request ID: %s is not required: "
+                "all indexes already sent",
```
vladnosiv, line 282: if two requests with the same long prefix get into one chunked prefill batch, but the second one ends up on the border of the chunk, all its blocks could have already been sent, since they are shared

lmcache/v1/storage_backend/connector/nixl_connector_v3.py
```diff
@@ -559,7 +569,8 @@ def _allocate_and_put(self, alloc_request: NixlAllocRequest) -> NixlAllocRespons
         alloc_indexes = []
         already_send_indexes = []
 
-        for idx, key in enumerate(alloc_request.keys):
+        for idx, key_str in enumerate(alloc_request.keys):
+            key = CacheEngineKey.from_string(key_str)
             if self._backend.contains(key, pin=True):
```
vladnosiv, line 561: here contains always returned False, because the key was supplied as a string, so I added assertions in those places

this led to repeated puts on the same key and leakage of the past mem-obj, respectively

lmcache/v1/storage_backend/nixl_backend_v3.py
```diff
@@ -168,10 +172,11 @@ def get_blocking(self, key: CacheEngineKey) -> Optional[MemoryObj]:
         :return: MemoryObj. None if the key does not exist.
         """
 
+        assert isinstance(key, CacheEngineKey)
         with self._data_lock:
             # NOTE(Jiayi): we assume that the key must be in local data
             # because we are using a push-based transfer
-            mem_obj = self._data.pop(key, None)
+            mem_obj = self._data.get(key, None)
```
vladnosiv, line 166: you can't do a pop here, since the same chunk can be shared between two running requests

in the case of pop, the first (idx=0 i mean) prefix block will not be found for the second running query, and in addition to performing a prefill on the decode node, non-common blocks of the second query will leak

vladnosiv:
Hi @YaoJiayi ! 
I'd be very happy if you could take a look at PR, thanks !

sammshen:
```bash
LMBenchmark/synthetic-multi-round-qa$ ./long_input_short_output_run.sh meta-llama/Llama-3.1-8B-Instruct http://localhost:9100 pd_test_1 0.5 
```

Still encounter the infinite spin loop from: 
`LMBenchmark/synthetic-multi-round-qa$ ./long_input_short_output_run.sh meta-llama/Llama-3.1-8B-Instruct http://localhost:9100 pd_test_1 0.5 `

https://github.com/LMCache/LMBenchmark


If you can resolve this, this solution should be a full patch since this is a high stress workload (concurrent users and context length)

vladnosiv:
@sammshen Thank you for the comment. This PR was primarily intended to fix the bugs that were identified during testing in the first comment. I will be able to update the behavior based on your example on Sunday or Monday.

In any case, according to my tests, these fixes were necessary for the gsp benchmark to work. 

I'm interested in clean PD load tests in LMCache without leaks, so I'll try to complete the PR as soon as possible.

sammshen:
LGTM! @Shaoting-Feng @YaoJiayi @ApostaC should we use this PR as the base change? And PTAL at https://github.com/LMCache/LMCache/pull/1134 asw!

sammshen (approved):

sammshen:
Unit tests for PD and potential refactors and especially fixing the load heavy workload (high QPS high reuse) is high on the priority list. @vladnosiv thanks for working on this!! :)

Shaoting-Feng:
This is absolutely great.

lmcache/v1/storage_backend/connector/nixl_connector_v3.py
Shaoting-Feng: I am still confused here.

Shaoting-Feng (commented):
Generally great, but I'm a bit confused about `if len(local_indexes) == 0`.

lmcache/v1/storage_backend/connector/nixl_connector_v3.py
vladnosiv: checking through len == 0 is the consequence of my recent transition to python, I'll correct it)

If the question is in essence, then local indexes are built with already sent chunks, which are compared by key, so if all the request chunks are already sent - they got into already sent and an empty list will return

vladnosiv:
To be honest, in the end I did not understand the question about len == 0, because in the LMCache codebase there are 19 occurrences of just such a check for the emptiness of the sequence, but just in case I corrected it to a more idiomatic style for python.

If the question is still about the essence of the fix in this place, then we can discuss it in more detail. 

Just in case, I'll briefly describe the problem once again: if a request that is a prefix of an already sent request gets into the batch, the local_indexes list becomes empty, you can look at how it is built. This causes a NIXL transfer error, because there is a contract for non-empty sending. This happens not only in the very extreme case of a prefix request, but also when one of the requests ends up in the batch in a truncated format due to CPP and, for example, it has the same system prompt with the rest of the variable.

YaoJiayi (approved):
lgtm!

PR #1164, opened 2025-07-27
[CD]: nightly build patch

sammshen:
harden builder should not block sudo or else setup docker buildx will fail


YaoJiayi (approved):
lgtm

PR #1166, opened 2025-07-27
[Bugfix] Concurrent LMCache instances on the same machine conflict due to a hardcoded IPC socket path, causing a race condition.

### LINKED ISSUE #1165: [Bug] When multiple LMCache instances run concurrently on the same machine, they may experience communication conflicts and race conditions.
**Describe the bug**
This issue is based on #1122 .

When multiple, independent LMCache instances are run concurrently on the same machine, they conflict with each other due to using a shared, hardcoded Inter-Process Communication (IPC) socket path. 

This leads to a race condition where one process's LookupClient can mistakenly connect to another process's LookupServer. 
The related codes are:
```
# lmcache/v1/rpc_utils.py

def get_zmq_rpc_path_lmcache(
    vllm_config: Optional["VllmConfig"] = None,
    rpc_port: int = 0,
    tp_rank: int = 0,
) -> str:
    """Get the ZMQ RPC path for LMCache lookup communication."""
    # Third Party
    import vllm.envs as envs

    base_url = envs.VLLM_RPC_BASE_PATH

    if isinstance(rpc_port, str):
        rpc_port = rpc_port + str(tp_rank)
    else:
        rpc_port += tp_rank

    logger.debug("Base URL: %s, RPC Port: %s", base_url, rpc_port)
    return f"ipc://{base_url}/lmcache_rpc_port_{rpc_port}"
```
The socket_path `f"ipc://{base_url}/lmcache_rpc_port_{rpc_port}"` is the a constant path for different instances. And this function `get_zmq_rpc_path_lmcache` is called in the initialization of `LMCacheLookupClient`.
```
# lmcache/v1/lookup_client/lmcache_lookup_client.py

class LMCacheLookupClient(LookupClientInterface):
    """ZMQ-based lookup client that communicates with a lookup server."""

    def __init__(self, vllm_config: "VllmConfig"):
        self.encoder = MsgpackEncoder()
        self.ctx = zmq.Context()  # type: ignore[attr-defined]
        rpc_port = vllm_config.kv_transfer_config.get_from_extra_config(
            "lmcache_rpc_port", 0
        )
        self.tensor_parallel_size = vllm_config.parallel_config.tensor_parallel_size
        for tp_rank in range(self.tensor_parallel_size):
            socket_path = get_zmq_rpc_path_lmcache(vllm_config, rpc_port, tp_rank)
            if tp_rank == 0:
                self.socket = make_zmq_socket(
                    self.ctx,
                    socket_path,
                    zmq.REQ,  # type: ignore[attr-defined]
                    bind=False,
                )
            else:
                self.socket.connect(socket_path)
```
So, for different LMCache instances, socket_path is always the same. This leads to a race condition.

Now I am taking an example for clarify this bug :
I start 2 vllm instances on the same machine, both of which are deployed with separate. My LMCache config is `chunk_size: 256`, `local_cpu: true`, `max_local_cpu_size: 5.0` (only use local_cpu, not to use a remote storage backend). For the 2 vllm instances, the same prompt is submitted. And the second vllm instance processes the prompt until the first vllm instance processes and stored the KVs.
In theory, the second vllm instance CANNOT hit the prefix. However, sometimes it does hit the prefix. (This is a matter of probability: I tried 5 times, where 2 hit the prefix.)

The full test codes are: (The code is slightly different from the demo code in my YouTube video, but the core logic is the same.)
```
import os
import subprocess
import time
from multiprocessing import Event, Process

from vllm import LLM, SamplingParams
from vllm.config import KVTransferConfig

prompts = [
    "Hello, how are you?" * 100
]

'''
lmcache_config.yaml:

# Basic configurations
chunk_size: 256

# CPU offloading configurations
local_cpu: true
max_local_cpu_size: 5.0  # 5GB CPU memory limit
'''

def run_vllm1(store_done, prompts):
    # We use GPU 0 for KV cache store process.
    os.environ["CUDA_VISIBLE_DEVICES"] = "0"
    os.environ["LMCACHE_CONFIG_FILE"] = "~/lmcache_config.yaml"

    sampling_params = SamplingParams(temperature=0, top_p=0.95, max_tokens=10)

    ktc = KVTransferConfig(kv_connector="LMCacheConnectorV1", kv_role="kv_both")
    llm = LLM(
        model="~/models/Mistral-7B-Instruct-v0.3",
        kv_transfer_config=ktc,
        max_model_len=8000,
        gpu_memory_utilization=0.7,
        enforce_eager=True,
    )

    outputs = llm.generate(prompts, sampling_params)
    for output in outputs:
        generated_text = output.outputs[0].text
        print(f"Generated text: {generated_text!r}")
    print("KV cache store is finished.")
    store_done.set()


def run_vllm2(store_done, prompts, timeout=1):
    # We use GPU 1 for KV cache retrieve process.
    os.environ["CUDA_VISIBLE_DEVICES"] = "1"
    os.environ["LMCACHE_CONFIG_FILE"] = "~/lmcache_config.yaml"

    sampling_params = SamplingParams(temperature=0, top_p=0.95, max_tokens=10)

    ktc = KVTransferConfig(kv_connector="LMCacheConnectorV1", kv_role="kv_both")
    llm = LLM(
        model="/home/lpy/models/Mistral-7B-Instruct-v0.3",
        kv_transfer_config=ktc,
        max_model_len=8000,
        gpu_memory_utilization=0.7,
        enforce_eager=True,
    )

    print("Waiting for KV cache store to finish...")
    store_done.wait()
    time.sleep(timeout)

    outputs = llm.generate(prompts, sampling_params)
    for output in outputs:
        generated_text = output.outputs[0].text
        print(f"Generated text: {generated_text!r}")


def main():
    store_done = Event()
    vllm1_process = Process(target=run_vllm1, args=(store_done, prompts))
    vllm2_process = Process(target=run_vllm2, args=(store_done, prompts))


    # Start KV cache store process
    vllm1_process.start()

    # Start KV cache retrieve process
    vllm2_process.start()

    # Clean up the processes
    vllm1_process.join()
    vllm2_process.join()


if __name__ == "__main__":
    main()
```

**To Reproduce**
The video url is "https://youtu.be/EqH-M-Z4U0A".
1. Create a python file named `test.py`, and copy my test code into it.
2. Create a yaml file named `lmcache_config.yaml`, and copy my yaml content into it. (My yaml content is `chunk_size: 256`, `local_cpu: true`, `max_local_cpu_size: 5.0`)
3. Change the path in the test code (the model path and the config file path) to the right path.
4. Run `python test.py` for several times (May need 3-5 times, even more).
5. Check the logs. We may see the second vllm instance hits the prefix. It looks like: 
```[2025-07-27 08:11:38,168] LMCache INFO: Reqid: 0, Total tokens 601, LMCache hit tokens: 601, need to load: 600 (vllm_v1_adapter.py:814:lmcache.integration.vllm.vllm_v1_adapter)
[2025-07-27 08:11:38,173] LMCache INFO: Post-initializing LMCacheEngine (cache_engine.py:160:lmcache.v1.cache_engine)
[2025-07-27 08:11:38,174] LMCache INFO: Retrieved 0 out of 601 out of total 601 tokens (cache_engine.py:512:lmcache.v1.cache_engine)
[2025-07-27 08:11:38,174] LMCache ERROR: The number of retrieved tokens is less than the expected number of tokens! This should not happen! (vllm_v1_adapter.py:524:lmcache.integration.vllm.vllm_v1_adapter)
[2025-07-27 08:11:38,174] LMCache ERROR: Num retrieved tokens: 0, num expected tokens: 601 (vllm_v1_adapter.py:528:lmcache.integration.vllm.vllm_v1_adapter
```
This seems like a matter of probability. I have already recorded a video to show this error in detail (I tried 5 times, where 2 hit prefix and 3 not). The video url is "https://youtu.be/EqH-M-Z4U0A"

**Screenshots**

Image

**How to fix**
The root cause of this issue lies in the get_zmq_rpc_path_lmcache utility function. This function constructs the IPC socket path using components that are not unique to the process instance:

A base path from VLLM_RPC_BASE_PATH (/tmp).

A configurable RPC port (lmcache_rpc_port) that defaults to 0.

Since both concurrent processes share the same environment and default configuration, they both generate the exact same socket path (ipc:///tmp/lmcache_rpc_port_0). This causes the first instance's server to bind to the path, and the second instance's client to erroneously connect to the first instance's server, breaking process isolation. The same issue applies to the ZMQOffloadServer.

Therefore we should not use a fixed socket_path, which is fatal for multiple LMCache Engines on the same machine.

 - OS: macOS
 - Browser safari
 - Version 18.5

**vllm:version 0.9.2.dev241+g8a5c226a1.d20250627**

**LMCache version: Using the codes after the commit 0f1593b**

@sammshen

Foreverythin:

FIX #1165  (*link existing issues this PR will resolve*)

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

`git tag vx.x.x`

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

`git tag -d vx.x.x`

For example:
`git tag -d v0.3.0`

For example:

> [!TIP]

### DIFF
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -123,7 +123,7 @@ def get_zmq_rpc_path_lmcache(
     )
 
     socket_path = (
-        f"{base_url}/engine_{engine_id}_service_{service_name}_"
+        f"ipc://{base_url}/engine_{engine_id}_service_{service_name}_"
         f"lmcache_rpc_port_{rpc_port}"
     )
 
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -15,7 +15,6 @@
 # Standard
 from typing import TYPE_CHECKING, Optional
 import socket
-import os
 
 # Third Party
 import zmq
@@ -95,13 +94,17 @@ def get_zmq_rpc_path_lmcache(
     # Third Party
     import vllm.envs as envs
 
+    if vllm_config is None or vllm_config.kv_transfer_config is None:
+        raise ValueError("vllm_config with kv_transfer_config is required to determine the engine_id.")
+
     base_url = envs.VLLM_RPC_BASE_PATH
-    pid = os.getpid()
+
+    engine_id = vllm_config.kv_transfer_config.engine_id
 
     if isinstance(rpc_port, str):
         rpc_port = rpc_port + str(tp_rank)
     else:
         rpc_port += tp_rank
 
-    logger.debug("Base URL: %s, PID: %d, RPC Port: %s", base_url, pid, rpc_port)
-    return f"ipc://{base_url}/pid_{pid}_lmcache_rpc_port_{rpc_port}"
+    logger.debug("Base URL: %s, Engine: %d, RPC Port: %s", base_url, engine_id, rpc_port)
+    return f"ipc://{base_url}/engine_{engine_id}_lmcache_rpc_port_{rpc_port}"
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Standard
-from typing import TYPE_CHECKING, Optional, Literal
+from typing import TYPE_CHECKING, Literal, Optional
 import socket
 
 # Third Party
@@ -90,6 +90,7 @@ def get_ip():
 def get_zmq_rpc_path_lmcache(
     vllm_config: Optional["VllmConfig"] = None,
     service_name: ServiceKind = "lookup",
+    rpc_port: int = 0,
     tp_rank: int = 0,
 ) -> str:
     """Get the ZMQ RPC path for LMCache lookup and offload communication."""
@@ -110,8 +111,19 @@ def get_zmq_rpc_path_lmcache(
 
     engine_id = vllm_config.kv_transfer_config.engine_id
 
+    if isinstance(rpc_port, str):
+        rpc_port = rpc_port + str(tp_rank)
+    else:
+        rpc_port += tp_rank
+
     logger.debug(
-        "Base URL: %s, Engine: %d, Service Name: %s, TP Rank: %s", 
-        base_url, engine_id, service_name, tp_rank
+        "Base URL: %s, Engine: %s, Service Name: %s, RPC Port: %s", 
+        base_url, 
+        engine_id, 
+        service_name, 
+        rpc_port
     )
-    return f"ipc://{base_url}/engine_{engine_id}_service_{service_name}_tp_rank_{tp_rank}"
+
+    return (
+        f"ipc://{base_url}/engine_{engine_id}_service_{service_name}_lmcache_rpc_port_{rpc_port}"
+    )
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -15,6 +15,7 @@
 # Standard
 from typing import TYPE_CHECKING, Optional
 import socket
+import os
 
 # Third Party
 import zmq
@@ -95,11 +96,12 @@ def get_zmq_rpc_path_lmcache(
     import vllm.envs as envs
 
     base_url = envs.VLLM_RPC_BASE_PATH
+    pid = os.getpid()
 
     if isinstance(rpc_port, str):
         rpc_port = rpc_port + str(tp_rank)
     else:
         rpc_port += tp_rank
 
-    logger.debug("Base URL: %s, RPC Port: %s", base_url, rpc_port)
-    return f"ipc://{base_url}/lmcache_rpc_port_{rpc_port}"
+    logger.debug("Base URL: %s, PID: %d, RPC Port: %s", base_url, pid, rpc_port)
+    return f"ipc://{base_url}/pid_{pid}_lmcache_rpc_port_{rpc_port}"
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -98,9 +98,7 @@ def get_zmq_rpc_path_lmcache(
     import vllm.envs as envs
 
     if vllm_config is None or vllm_config.kv_transfer_config is None:
-        raise ValueError(
-            "vllm_config with kv_transfer_config is required to determine the engine_id."
-        )
+        raise ValueError("A valid kv_transfer_config with engine_id is required.")
 
     if service_name not in {"lookup", "offload"}:
         raise ValueError(
@@ -117,13 +115,16 @@ def get_zmq_rpc_path_lmcache(
         rpc_port += tp_rank
 
     logger.debug(
-        "Base URL: %s, Engine: %s, Service Name: %s, RPC Port: %s", 
-        base_url, 
-        engine_id, 
-        service_name, 
-        rpc_port
+        "Base URL: %s, Engine: %s, Service Name: %s, RPC Port: %s",
+        base_url,
+        engine_id,
+        service_name,
+        rpc_port,
     )
 
-    return (
-        f"ipc://{base_url}/engine_{engine_id}_service_{service_name}_lmcache_rpc_port_{rpc_port}"
+    socket_path = (
+        f"{base_url}/engine_{engine_id}_service_{service_name}_"
+        f"lmcache_rpc_port_{rpc_port}"
     )
+
+    return socket_path
--- a/lmcache/v1/rpc_utils.py
+++ b/lmcache/v1/rpc_utils.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Standard
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Literal
 import socket
 
 # Third Party
@@ -28,6 +28,8 @@ if TYPE_CHECKING:
     from vllm.config import VllmConfig
 
 logger = init_logger(__name__)
+
+ServiceKind = Literal["lookup", "offload"]
 
 
 def get_zmq_context():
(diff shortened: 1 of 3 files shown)


lmcache/v1/rpc_utils.py
```diff
@@ -95,11 +96,12 @@ def get_zmq_rpc_path_lmcache(
     import vllm.envs as envs
 
     base_url = envs.VLLM_RPC_BASE_PATH
+    pid = os.getpid()
```
maobaolong: The  pid of each vllm process are all different, so how to communicate each other? @Foreverythin

Foreverythin: Thank you for your correction. Using os.getpid() will indeed destroy the internal communication of multiple ranks in a single instance.

After consideration, I improved the solution: use engine_id (vllm_config.kv_transfer_config.engine_id). This ensures that multiple independent vLLM instances running on the same machine will be completely isolated (there was a conflict in the previous implementation), and also ensures that all ranks in the same instance can share this ID, thereby building the correct internal communication path.

Is this version acceptable?

Foreverythin:
I tested the code (deploying two vllm instances on the same machine using two different methods: 1. with separate; 2. TP=2). After testing, the program did not encounter the issue described in Issue#1165 (conflict between different instances due to the same socket_path), and everything worked well.

The test codes are:
```
import os
import subprocess
import time
from multiprocessing import Event, Process

from vllm import LLM, SamplingParams
from vllm.config import KVTransferConfig

prompts = [
    "Hello, how are you?" * 100
]

'''
lmcache_config.yaml:

# Basic configurations
chunk_size: 256

# CPU offloading configurations
local_cpu: true
max_local_cpu_size: 5.0  # 5GB CPU memory limit
'''

def run_vllm1(store_done, prompts):
    # We use GPU 0 for KV cache store process.
    os.environ["CUDA_VISIBLE_DEVICES"] = "2,3"
    os.environ["LMCACHE_CONFIG_FILE"] = "/home/lpy/dist-rag/examples/lmcache_config.yaml"

    sampling_params = SamplingParams(temperature=0, top_p=0.95, max_tokens=10)

    ktc = KVTransferConfig(kv_connector="LMCacheConnectorV1", kv_role="kv_both")
    llm = LLM(
        model="/home/lpy/models/Mistral-7B-Instruct-v0.3",
        kv_transfer_config=ktc,
        max_model_len=8000,
        gpu_memory_utilization=0.95,
        enforce_eager=True,
        tensor_parallel_size=2,
    )

    outputs = llm.generate(prompts, sampling_params)
    for output in outputs:
        generated_text = output.outputs[0].text
        print(f"Generated text: {generated_text!r}")
    print("KV cache store is finished.")
    store_done.set()


def run_vllm2(store_done, prompts, timeout=1):
    # We use GPU 1 for KV cache retrieve process.
    os.environ["CUDA_VISIBLE_DEVICES"] = "4,5"
    os.environ["LMCACHE_CONFIG_FILE"] = "/home/lpy/dist-rag/examples/lmcache_config.yaml"

    sampling_params = SamplingParams(temperature=0, top_p=0.95, max_tokens=10)

    ktc = KVTransferConfig(kv_connector="LMCacheConnectorV1", kv_role="kv_both")
    llm = LLM(
        model="/home/lpy/models/Mistral-7B-Instruct-v0.3",
        kv_transfer_config=ktc,
        max_model_len=8000,
        gpu_memory_utilization=0.95,
        enforce_eager=True,
        tensor_parallel_size=2,
    )

    print("Waiting for KV cache store to finish...")
    store_done.wait()
    time.sleep(timeout)

    outputs = llm.generate(prompts, sampling_params)
    for output in outputs:
        generated_text = output.outputs[0].text
        print(f"Generated text: {generated_text!r}")


def main():
    store_done = Event()
    vllm1_process = Process(target=run_vllm1, args=(store_done, prompts))
    vllm2_process = Process(target=run_vllm2, args=(store_done, prompts))


    # Start KV cache store process
    vllm1_process.start()

    # Start KV cache retrieve process
    vllm2_process.start()

    # Clean up the processes
    vllm1_process.join()
    vllm2_process.join()


if __name__ == "__main__":
    main()

```

The screenshots of results:
**1. TP=2**
image

**2. With seperate:**
image


@YaoJiayi  Could you please take a look?

lmcache/v1/rpc_utils.py
```diff
@@ -94,12 +94,17 @@ def get_zmq_rpc_path_lmcache(
...
+    engine_id = vllm_config.kv_transfer_config.engine_id
+
     if isinstance(rpc_port, str):
         rpc_port = rpc_port + str(tp_rank)
     else:
         rpc_port += tp_rank
 
-    logger.debug("Base URL: %s, RPC Port: %s", base_url, rpc_port)
-    return f"ipc://{base_url}/lmcache_rpc_port_{rpc_port}"
+    logger.debug("Base URL: %s, Engine: %d, RPC Port: %s", base_url, engine_id, rpc_port)
```
maobaolong: @Foreverythin  The root cause of this issue is that you didn't set the `rpc_port`, so each instance use the same `rpc_port`.

I know it is not easy to use to set the `lmcache_rpc_port` while starting vLLM. So It indeed need a better way to instead of.

So I think your improvement is better than the original `lmcache_rpc_port`, so let's wait for the confirm from @ApostaC .

Apart from that, this LGTM.

sammshen: would it be correct that we can get rid of `rpc_port` if we are using the `engine_id` now?

Foreverythin: > would it be correct that we can get rid of `rpc_port` if we are using the `engine_id` now?

Yes, we can get rid of using `rpc_port`, but we need modify more codes.

The function of `get_zmq_rpc_path_lmcache` is called by `LMCacheLookupClient`, `LMCacheLookupServer` and `ZMQOffloadServer`.  To expand on this, the core issue is that `LMCacheLookupServer` (for handling lookups) and `ZMQOffloadServer` (for handling offloading) are two distinct services. They must use separate, isolated communication channels to avoid conflicts, as they handle different message types.

Previously, this separation was achieved by having them use different `rpc_port` values (e.g., a default of 0 for lookup and 100 for offload). If we just remove `port` in the socket_path, then there will be a conflict between `LMCacheLookupServer` and `ZMQOffloadServer` .

My proposed solution is to replace the numeric `rpc_port` with a descriptive string identifier for each service. This makes the distinction explicit within the code and completely removes the need for any user-facing port configuration.

1. Remove the parameter `rpc_port`;
2. Add a new parameter `service_name` ("lookup", "offload");
3. Change socket_path to `f"ipc://{base_url}/engine_{engine_id}_service_{service_name}_tp_rank_{tp_rank}"`;
4. Modify the way function `get_zmq_rpc_path_lmcache` is called.

The latest code has been updated (48e3198)  and I have tested it with no issues. So, should we avoid using rpc_port? I believe that avoiding the use of rpc_port will make it easier for users and reduce user configuration. @maobaolong @sammshen

Foreverythin: > would it be correct that we can get rid of `rpc_port` if we are using the `engine_id` now?

Yes, we can get rid of using `rpc_port`, but we need modify more codes.

The function of `get_zmq_rpc_path_lmcache` is called by `LMCacheLookupClient`, `LMCacheLookupServer` and `ZMQOffloadServer`.  To expand on this, the core issue is that `LMCacheLookupServer` (for handling lookups) and `ZMQOffloadServer` (for handling offloading) are two distinct services. They must use separate, isolated communication channels to avoid conflicts, as they handle different message types.

Previously, this separation was achieved by having them use different `rpc_port` values (e.g., a default of 0 for lookup and 100 for offload). If we just remove `port` in the socket_path, then there will be a conflict between `LMCacheLookupServer` and `ZMQOffloadServer` .

My proposed solution is to replace the numeric `rpc_port` with a descriptive string identifier for each service. This makes the distinction explicit within the code and completely removes the need for any user-facing port configuration.

1. Remove the parameter `rpc_port`;
2. Add a new parameter `service_name` ("lookup", "offload");
3. Change socket_path to `f"ipc://{base_url}/engine_{engine_id}_service_{service_name}_tp_rank_{tp_rank}"`;
4. Modify the way function `get_zmq_rpc_path_lmcache` is called.

I have given a solution and updated the codes (48e3198). And it LGTM. So, should we avoid using rpc_port? I believe that avoiding the use of rpc_port will make it easier for users and reduce user configuration. @maobaolong @sammshen

maobaolong: @Foreverythin I'd prefer to remove the `lmcache_rpc_port` initially, but after search the existing usage, i'm ok you keep it to keep consistency.

image

sammshen:
LGTM!

sammshen (approved):

Foreverythin:
Considering there are usages of `lmcache_rpc_port` in some current documents, I agree with @maobaolong to keep `lmcache_rpc_port`. I have updated the codes. 

The new implementation now follows a hybrid approach:
1. If `lmcache_rpc_port` is not provided by the user, the connector now automatically generates a unique socket path using the engine_id and an internal service identifier ("lookup" or "offload"). This means new users don't have to worry about the `lmcache_rpc_port` setting, which makes LMCache more convenient and robust to use.
2. If a user does explicitly provide a value for `lmcache_rpc_port` in their configuration, the connector will honor it and use that value to construct the socket path. This ensures that users can continue to follow existing documentation and that current scripts will work without any changes. 

The codes LGTM.
@sammshen @maobaolong

maobaolong (approved):
LGTM

YaoJiayi (approved):
LGTM

PR #1170, opened 2025-07-28
[bugfix] fix save unfull chunk in vllm_v1_adapter

chunxiaozheng:
Support setting not save unfull chunks through lmcache config file or vllm config.

- lmcache config file:
```
...
save_unfull_chunk: False
```

- vllm config:
```
{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both","kv_parallel_size":2,"kv_connector_extra_config":{"discard_partial_chunks":true}}
```


maobaolong (approved):
@chunxiaozheng Thanks for this improvement, now the two source configs are clear and align together.

maobaolong:
@YaoJiayi Would you like to take another look? Thank you.

YaoJiayi (approved):
Good catch. Thanks!

Issue #962, opened 2025-07-03
[RFC] Page cache control knob


1saac-k:
In a Linux kernel environment, page cache is used when using disk (`LMCACHE_LOCAL_DISK`). Using the page cache can lead to duplicate data and memory waste. This RFC propose an experimental patch that allows bypassing the page cache. Please share your comments.

### What is the page cache?

> The physical memory is volatile and the common case for getting data into the memory is to read it from files. Whenever a file is read, the data is put into the page cache to avoid expensive disk access on the subsequent reads. Similarly, when one writes to a file, the data is placed in the page cache and eventually gets into the backing storage device. The written pages are marked as dirty and when Linux decides to reuse them for other purposes, it makes sure to synchronize the file contents on the device with the updated data.

https://www.kernel.org/doc/html/v6.15/admin-guide/mm/concepts.html#page-cache

> The page cache is the primary way that the user and the rest of the kernel interact with filesystems. It can be bypassed (e.g. with O_DIRECT), but normal reads, writes and mmaps go through the page cache.

https://www.kernel.org/doc/html/v6.15/mm/page_cache.html#page-cache

### Page cache usage check

0. Discard page caches (with root privilege): `echo 1 > /proc/sys/vm/drop_caches`
1. Configure LMCache to use only disk and run it.
    - Local memory is used as a buffer with a capacity of 5GB by default.
2. Check the capacity of `used` and `buff/cache` (page cache) items using the `free -h` command.
3. Send requests to store tens of GB of KV Cache.
4. Using the `free -h` command, observe that `used` increases by about 5GB, while `buff/cache` increases by tens of GB!

When using both memory and disk, page cache is also used.

### Problems with using page cache

The roles of LMCache's internal buffer (`StorageManager`) and page cache overlap. Using page cache results in duplicate data in both LMCache and page cache, leading to memory waste.

Users may assume that they are only using the memory capacity set by `LMCACHE_MAX_LOCAL_CPU_SIZE`. This can lead to misunderstandings in terms of capacity and performance. Due to kernel policy, page cache uses the free memory (CPU memory) without capacity limits. For example, consider a system with 1TB of memory, where LMCache's memory size is set to 10GB and disk size is set to 500GB. All 500GB of data can be stored in page cache. Users may encounter a situation where much more than the expected 10GB of memory (500GB) is used, and since this data is in memory, disk access does not occur, leading to excellent performance when using disk configuration. 

[...]

PR #1172, opened 2025-07-28
[Refactor] refactor disk and prefetch code paths

YaoJiayi:
---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

`git tag vx.x.x`

For example:
`git tag v0.3.0`
`git push origin v0.3.0`

`git tag -d vx.x.x`

For example:
`git tag -d v0.3.0`

For example:

> [!TIP]


lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -96,8 +224,7 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
             return True
```
vladnosiv, line 224: Hi @YaoJiayi ! 
It seems to me that contains should work differently: a pin for contains should mean moving to RAM right here, only in this case can we assume that a pin has occurred.

As it is now: pin does not allow deleting a block from disk, but there is no guarantee that during get_blocking it will be possible to move the block to RAM, since the entire RAM cache may be occupied by blocks with pin=True (under load). 

At the same time, the cache engine rightly believes that if contains(pin=True) returned True, then there are no obstacles to extracting this block.

I will be glad to discuss this as part of this PR, since it largely concerns the disk backend. I regularly encounter this situation in my SSD cache load tests.

YaoJiayi: The reason you see RAM cache occupied with pin is probably a bug. But I do agree the entire RAM could be pinned in certain extreme cases especially we allow users to explicitly pin a cache. However, I don't think handling such cases is in the scope of this PR. We have other ongoing efforts to support error handling and limiting the number of pinned caches.

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -284,32 +454,59 @@ async def async_save_bytes_to_disk(
...
 
-        async with aiofiles.open(path, "wb") as f:
-            await f.write(byte_array)
+        # FIXME(Jiayi): need to add ref count in disk memory object
+        if size % self.os_disk_bs != 0 or not self.use_odirect:
+            with open(path, "wb") as f:
+                f.write(byte_array)
+        else:
+            fd = os.open(path, os.O_CREAT | os.O_WRONLY | os.O_DIRECT)
+            os.write(fd, byte_array)
```
1saac-k, line 463: `os.close(fd)` required.

YaoJiayi: Hey, thanks for the reminder. Could you submit a small pr to fix this?

1saac-k: I'm happy to do so as well. Due to our company's security policies, it may take a day or two (or possibly a few days).
I've found a few more issues, which I'll detail in the comments of the issue I posted (#962).
- os.close()
- 512-byte alignment of buffer address and size is required
- File permissions to 664 (`os.open(.., .., 0o664)`)

1saac-k: alignment & permission issue fixed in #1233

PR #1173, opened 2025-07-28
[bug]: free on unpin

sammshen:
https://github.com/LMCache/LMCache/pull/1156
cleaned up version w/o prometheus logging (which shows no bugs currently with pin)

in case a memory object is unpineed after ref counted down

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -360,6 +360,12 @@ class TensorMemoryObj(MemoryObj):
                 "Setting pin count back to 0 as a hack but please find the bug."
             )
             self.metadata.pin_count = 0
+        if (
+            self.meta.ref_count == 0
+            and self.parent_allocator is not None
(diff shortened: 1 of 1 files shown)


maobaolong (approved):
@sammshen Thanks for the explain within the #1156 . This PR LGTM.

Just some question and suggestion about pin/unpin, not related to this PR.
- We assume the `contains` call only 1 time and retrieve 1 time, if not, the pin_count will leak or underflow.
- As the ref_count approach is used by all the existing subclass of `MemoryObj`, so suggest to do a refactor to reduce the duplicate code, something like to introduce a generic parent class, or put the metadata and pin logic into the `MemoryObj`.

sammshen:
closing after refactor in https://github.com/LMCache/LMCache/pull/1172/files#diff-8987071aae9ca6f53325e3094b21d00d81e2e3bd728008cfa352a31d4b7e24dd

PR #1193, opened 2025-07-30
[WIP] Support Async Loading

yoo-kumaneko:
This is still a work-in-progress PR. We aim to support the async loading feature by creating a new LMCacheConnectorV1AsyncImpl that inherits the original adapter.


ApostaC:
Thanks! I just did a very quick pass, and the high-level structure looks good. Will do a detailed review later!

PR #1197, opened 2025-07-30
[Bugfix] Get cache size from memory when removing a cache without accessing disk

YichuanSun:
When invoking `LocalDiskBackend.remove()`, it gets the cache size from disk by calling `os.path.getsize(path)`, but the size already saved in `self.dict`. By invoking `self.dict[key].size`, one costly disk access is avoided.


Shaoting-Feng (approved):
LGTM

YaoJiayi (approved):
lgtm

YaoJiayi:
@YichuanSun Please fix the DCO

PR #1198, opened 2025-07-30
[CI]: Refactor/Fix CI

sammshen:
FIX currently broken unit/integration/correctness/e2e/codeql tests

1. unit tests

mock the pin and mixed memory allocators to still use `pin_memory=True` instead of manual `cudaHostRegister`

rewrite the package lock

2. integration tests

disable flash infer because not supported on L4
add timeout

3. correctness

simplify script to be able to run on 2x L4s to offline

4. e2e

fix the buildkite id for two agents

rewrite

5. codeql

make the harden runner not crash from changing github CDN

misc:

add buildkite deletion for `/var/cache/buildkite` so that buildkite runner won't crash in teh future from out of disk space


Shaoting-Feng (approved):
Thank you so much for the fix!

YaoJiayi (approved):

PR #1199, opened 2025-07-30
[CI]: isolate versioning fix

### LINKED ISSUE #1184: LMCache is causing torch versioning issue in vLLM docker build pipeline
# Background

In vllm-project/vllm#18193, LMCache is introduced as a dependency in vLLM's Docker image at the end of the vLLM Docker build pipeline. 

# Problem

When installing LMCache, it forces torch == 2.7.0 (in those requirement txt files), which will overwrite the torch version inside vLLM's docker image during their docker build pipeline.

# Proposed solution

Remove the torch version spec in the requirement files

# Other considerations

We build the kernels using torch 2.7.0 in the `pyproject.toml`, and it may cause some version-incompatible issues.
We might need to be able to detect if the current environment has torch installed and use that version to build.

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1184

SOLUTION: as long as lmcache does not pin a torch version, when the vllm dockerfile does pip install lmcache, it will reuse vllm's own torch version.

Downside: since torch version is no longer deterministic, it may be harder to debug some issues in the future but prioritization is compatibility with vllm first

TODO (asap): create a new LMCache release after this PR is merged

### DIFF
--- a/requirements/cuda.txt
+++ b/requirements/cuda.txt
@@ -5,7 +5,9 @@
 ray >= 2.9
 nvidia-ml-py # for pynvml package
 
-# These must be updated alongside torch to correspond to vLLM versions
-torch == 2.7.0
-torchvision == 0.22.0   # Required for phi3v processor. See https://github.com/pytorch/vision?tab=readme-ov-file#installation for corresponding version
-xformers == 0.0.30; platform_system == 'Linux' and platform_machine == 'x86_64'  # Requires PyTorch 2.7.0
+# These dependencies also exist in vLLM (so we don't pin the version, even though this may cause some hard-to-debug issues)
+torch
+# Required for phi3v processor. See https://github.com/pytorch/vision?tab=readme-ov-file#installation for corresponding version
+torchvision
+# platform_system == 'Linux' and platform_machine == 'x86_64' 
+xformer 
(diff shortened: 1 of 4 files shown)


requirements/cuda.txt
```diff
@@ -5,7 +5,9 @@
...
-# These must be updated alongside torch to correspond to vLLM versions
-torch == 2.7.0
-torchvision == 0.22.0   # Required for phi3v processor. See https://github.com/pytorch/vision?tab=readme-ov-file#installation for corresponding version
-xformers == 0.0.30; platform_system == 'Linux' and platform_machine == 'x86_64'  # Requires PyTorch 2.7.0
+# These dependencies also exist in vLLM (so we don't pin the version, even though this may cause some hard-to-debug issues)
+torch
+# Required for phi3v processor. See https://github.com/pytorch/vision?tab=readme-ov-file#installation for corresponding version
+torchvision
+# platform_system == 'Linux' and platform_machine == 'x86_64' 
+xformer
```
YaoJiayi: Is it xformers or xformer?

sammshen: thank you!!

YaoJiayi (approved):
lgtm

PR #1200, opened 2025-07-31
[CacheEngine] Support separate cache key by user through lmcache tags

xshwu-ai:
In our case, we should to separate cache data by `user` field，then I add `tags: OrderedDict` for the extention

How to use:
add `user` field to `kv_transfer_params`

```
    -d '{
    "model": "deepseek-r1",
    "messages": [
        {
            "role": "user",
            "content": "Hello, how are you?"
        }
    ],
    "max_tokens": 1000,
    "temperature": 0,
    "top_p": 0.95,
    "stream": true,
    "stream_options": {
         "include_usage": true
    },
    "kv_transfer_params":{
            "user":"example_user_id"
        }
    }'
```


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -777,6 +791,12 @@ def get_num_new_matched_tokens(
                 token_ids, request.mm_hashes, request.mm_positions
             )
 
+        tags = None
+        if request.sampling_params.extra_args is not None:
```
maobaolong: As there are two place similar, suggest to abstract a method

lmcache/utils.py
```diff
@@ -78,10 +79,14 @@ def __hash__(self):
         )
 
     def to_string(self):
-        return (
+        s = (
             f"{self.fmt}@{self.model_name}@{self.world_size}"
             f"@{self.worker_id}@{self.chunk_hash}"
         )
+        if self.tags is not None:
```
maobaolong: Is it necessary to add an addition check `and len(tags) == 0`

maobaolong (commented):
@xshwu-ai Thanks for this feature, left some minor comments, PTAL.

lmcache/utils.py
xshwu-ai: `and len(tags) != 0`

maobaolong (approved):
The code LGTM. @YaoJiayi Would you like to take another look at this PR? Thanks!

YaoJiayi (commented):
@xshwu-ai Can we add an example under examples/ ? Otherwise, it might be hard for people to maintain or use it

xshwu-ai:
> @xshwu-ai Can we add an example under examples/ ? Otherwise, it might be hard for people to maintain or use it

I added the example under examples/cache_by_tags

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -101,6 +112,8 @@ class RequestTracker:
     # Multimodal hashes and positions
     mm_hashes: Optional[list[str]] = None
     mm_positions: Optional[list["PlaceholderRange"]] = None
+    # The request tags
+    tags: OrderedDict = None
```
YaoJiayi: Optional[OrderDict]? @hickeyma do you know why linter cannot detect this?

YaoJiayi: Also, why do we need OrderedDict here? Why not string?

YaoJiayi (commented):
Why not make tags a string? I think it could make the code a lot simpler:)

lmcache/integration/vllm/vllm_v1_adapter.py
maobaolong: @YaoJiayi Because the tags can be multiply key-value pair, for example
```json
{
  "user":"mbl",
  "ns":"test_ns"
}
```

As the above use case, maybe use tags can be more clear. WDYT?

YaoJiayi (commented):
LGTM! Please fix the conflicts

maobaolong:
@YaoJiayi Conflict has been resolved, PTAL.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -115,6 +126,8 @@ class RequestTracker:
...
```
chunxiaozheng: use `Optional[OrderedDict]`

lmcache/utils.py
```diff
@@ -72,6 +72,7 @@ class CacheEngineKey:
     world_size: int
     worker_id: int
     chunk_hash: int
+    tags: OrderedDict = None
```
chunxiaozheng: use `Optional[OrderedDict]`

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -89,6 +90,16 @@ class DisaggSpec:
 tmp_disagg_tracker: dict[str, DisaggSpec] = {}
 
 
+def extract_tags(sampling_params: SamplingParams) -> OrderedDict:
+    tags = None
+    if sampling_params.extra_args is not None:
+        if kv_transfer_params := sampling_params.extra_args.get("kv_transfer_params"):
+            if user := kv_transfer_params.get("user"):
+                tags = OrderedDict()
```
chunxiaozheng: if we have more than one tags, init in here is inflexible

xshwu-ai:
@YaoJiayi all comments are resolved, please merge the PR, thank you.

chunxiaozheng:
lgtm

ApostaC (approved):
LGTM!

ApostaC:
@xshwu-ai Thanks for the great effort! It would also be great if you could contribute to our docs in the `docs/` folder. There is a section called `KV Cache offloading and sharing` and we can add a new page there.

image

maobaolong:
@ApostaC https://github.com/LMCache/LMCache/issues/1322 Created an issue to track this followup task, @xshwu-ai Would you like to take task?

xshwu-ai:
> @ApostaC #1322 Created an issue to track this followup task, @xshwu-ai Would you like to take task?

yes

PR #1201, opened 2025-07-31
[CI]: follow up fix gpu contention

sammshen:
avoid integration adn unit test try to choose same gpu.

add daemon on ci runner to kill processes using gpu for > 1 hour


ApostaC (approved):
LGTM!

Issue #807, opened 2025-06-10
Add support for ascend npu


fhygh:
**Is your feature request related to a problem? Please describe.**

**Add support for ascend npu**

Describe the solution you'd like

**Describe alternatives you've considered**

**Additional context**

YaoJiayi:
@fhygh Thanks for raising this issue! Do you want to contribute to this?

fhygh:
> [@fhygh](https://github.com/fhygh) Thanks for raising this issue! Do you want to contribute to this?

@YaoJiayi  hm.I think this is a big task, and I am very willing to do it, but I am worried that I may not have enough time and energy. If there are experts in Ascend and Npu who can collaborate, that would be even better.

YaoJiayi:
Yeah, the problem is that we do not have any hardwares:)

matthewygf:
> > [@fhygh](https://github.com/fhygh) Thanks for raising this issue! Do you want to contribute to this?
> 
> [@YaoJiayi](https://github.com/YaoJiayi) hm.I think this is a big task, and I am very willing to do it, but I am worried that I may not have enough time and energy. If there are experts in Ascend and Npu who can collaborate, that would be even better.

Hey ! We are in the process of porting, and already have a version ready. Just finalizing ! Happy to discuss more !

matthewygf:
> Yeah, the problem is that we do not have any hardwares:)

Actually I would like to discuss whether we can do something similar to vllm hardware plugin, currently we are simply omitting the cuda code path. Probably need a larger refactoring for different hardware...

YaoJiayi:
@matthewygf  Great! Feel free to talk to me here or on slack!

fhygh:
> Yeah, the problem is that we do not have any hardwares:)

@YaoJiayi  Our team is working on supporting the Ascend NPU card with lmcache. We have obtained some hardware and are conducting some experiments. Further discussions can be held later.

fhygh:
> > > [@fhygh](https://github.com/fhygh) Thanks for raising this issue! Do you want to contribute to this?
> > 
> > 
> > [@YaoJiayi](https://github.com/YaoJiayi) hm.I think this is a big task, and I am very willing to do it, but I am worried that I may not have enough time and energy. If there are experts in Ascend and Npu who can collaborate, that would be even better.
> 
> Hey ! We are in the process of porting, and already have a version ready. Just finalizing ! Happy to discuss more !

[...]

PR #1209, opened 2025-08-01
[Enhancement] Ascend NPU Support on LMCache.

matthewygf:
Acciendentally closed previous [PR](https://github.com/LMCache/LMCache/pull/1191)

This PR adds Ascend NPU 910b series initial support to LMCache :

- support for multi_layer and single_layer transfer for vLLM connector

Changes Made

1. Added custom CMakeBuild system support based on Ascend CANN Toolkit 8.2
2. Instead of BUILD_WITH environment variable to switch between CUDA and HIP builds, replaced with LMCACHE_TARGET_DEVICE, similar to VLLM_TARGET_DEVICE
3. Kernels support for multi_layer and single_layer transfer
4. Added a host register API in the pybind11 interface for Ascend only build due to current python API limitation
5. Require current_platform API in the vllm_adapter to use NPU specific import
6. Made changes to the test_*.py that uses the necessary NPU APIs to run successfully, all unit tests passed on our platform except the skipped ones, and require non-supported kernels.

[related issue](https://github.com/LMCache/LMCache/issues/807)

## Pre-requisite

NPU Driver version >= 24.1

CANN Tool-kit version >= 8.2

Installed compatible version of vllm-ascend, torch-npu

We currently support the following vLLM-ascend versions:
- v0.9.2

## Build Instructions

```yaml
LMCACHE_TARGET_DEVICE="ASCEND" \
pip install --no-build-isolation -v -e .
```

## Test Coverage

We ran the following tests:
1) unit tests successfully except skipped, and non-supported kernels, i.e. cachegen, gds, weka.
2) lm_eval: 

### Results without LMCache on Qwen3-8B

| Tasks | Version | Filter | n-shot | Metric | | Value | Stderr |
| :--- | :--- | :--- | :--- | :--- | :-: | :--- | :--- |
| gsm8k | 3 | flexible-extract | 5 | exact_match | ↑ | 0.8400 | ± 0.0101 |
| gsm8k | 3 | strict-match | 5 | exact_match | ↑ | 0.8355 | ± 0.0102 |

<br>

### Results with LMCache Layerwise

| Tasks | Version | Filter | n-shot | Metric | | Value | Stderr |
| :--- | :--- | :--- | :--- | :--- | :-: | :--- | :--- |
| gsm8k | 3 | flexible-extract | 5 | exact_match | ↑ | 0.8385 | ± 0.0101 |
| gsm8k | 3 | strict-match | 5 | exact_match | ↑ | 0.8332 | ± 0.0103 |

## Related PR

This following PR in vllm-ascend is related to support LMCache
- [support for KVConnector v1](https://github.com/vllm-project/vllm-ascend/pull/2039)

## Roadmap

We currently only support eagermode execution. We plan to have the following features in the near future:
- [ ] Dockerfile-npu-cann
- [ ] Cachegen
- [ ] Cacheblend
- [ ] Graph Mode with torch_npu torchair
- [ ] P/D with NPU Transport


matthewygf:
@YaoJiayi here is the new pr.

csrc/ascend/kernels/load_and_reshape_flash.cpp
```diff
@@ -0,0 +1,264 @@
+/*
```
YaoJiayi, line 1: A very dumb question since I'm not familiar with Ascend. Are those cpp files the same as cuda on nv devices?

YaoJiayi (commented):
LGTM in general. Thanks for the huge effort! 
High-level idea on how we move forward: can we have a separate repo (e.g., sth like LMCache-Ascend under our project)? I think it would be easier for everyone to maintain the code:)
@matthewygf What do you think?

csrc/ascend/kernels/load_and_reshape_flash.cpp
matthewygf: No worries at all. Yep these are basically the same as cu files for nv

matthewygf:
Agreed. Would be a lot easier for sure. I suppose we don't have to do the platform plugin style registration similar to vLLM, since other hardware vendors can easily use LMCache. Wdyt? @YaoJiayi

hijeffwu:
@YaoJiayi  Is there a more suitable way to support different device backends, such as a plugin-based approach? Besides NPU, there will be AI accelerator cards from other manufacturers in the future. How can we extend it conveniently?

YaoJiayi:
@matthewygf No need to be that complicated for now:)

Shichang-Zhang:
Hello Matthew, thanks for the great enhancement of Ascend support on LMCache! 
Recently I'm working on running vllm+LMCache on NPU. I built an image of your PR code based on the **vllm-ascend v0.9.1-dev image**( quay.io/ascend/vllm-ascend:v0.9.1-dev), and then wrote the K8s deployment yaml with command:
```
vllm serve deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B \
                  --enable-prefix-caching \
                  --block-size 128 \
                  --kv-events-config '{"enable_kv_cache_events": true, "publisher":"zmq", "topic":"kv-events"}' \
```
The environment has 8 chips of 910B4 NPU, and the resource limit in the K8s deployment yaml is:
```
resources:
            limits:
              cpu: 24
              memory: 64G
              huawei.com/Ascend910: "1"
            requests:
              cpu: 24
              memory: 64G
              huawei.com/Ascend910: "1"
```
When the container was started, I received a mlock call error failing to pin host memory. Did you encounter similar questions before? Thanks!
vllm-lmcache-bug1

matthewygf:
@Shichang-Zhang Hi, thanks for trying out the PR. 
The issue is likely related to the driver being at 24.1, where the host registered API does not work with the torchnpu malloc. Therefore we have to add mlock to the memory pages. In the container environment, if permission allows, you could add IPC_LOCK Capabilities to the container and configure the ulimit.

Otherwise I would advice updating the driver to >= 25, so the host registered API would work with torchnpu malloc memory.

Shichang-Zhang:
@matthewygf Thanks Matthew! The problem was solved by enabling IPC_LOCK.

matthewygf:
Closing, as we now have LMCache-Ascend repo.

PR #1211, opened 2025-08-01
[bugfix] fix local cache usage error

chunxiaozheng:
We should update `local_cache_usage_bytes` only in `Allocator`, includes `TensorMemoryAllocator` and `PagedTensorMemoryAllocator`, increase when `allocate` or `batched_allocate` is called, and decrease when `free` or `batched_free` is called.


Shaoting-Feng (approved):
LGTM

maobaolong (approved):
@chunxiaozheng LGTM thanks for this improvement, it make code much clear.

PR #1215, opened 2025-08-01
[WIP] 1P1D mem leak fix

sammshen:
The issue is with chunked prefill, because there is a dependency where the prefiller allocates more and more during the chunked prefill on the decoder's nixl buffer. however, the prefiller does not report back to teh proxy untill the entire prefill is finished. and only then can the proxy feed the request to the decoder. Thus, the issue is that if the decoder's buffer fills before chunked prefill finishes on the prefiller side, nobody can progress anymore since:
1. the prefiller is blocked by the decoder to respond that it allocated on its buffer
2. the decoder is blocked by the proxy not allowing it to free its buffer by sending a request
3. the proxy is blocked by the prefiller not having finished the request at hand and reporting to the proxy

This creates a three-way deadlock between prefiller, decoder, and proxy.

# Proposed Fix:

make decoder use cpu buffer during chunked prefill to offload nixl buffer

### DIFF
--- a/examples/disagg_prefill/1p1d_experimental/configs/lmcache-decoder-config.yaml
+++ b/examples/disagg_prefill/1p1d_experimental/configs/lmcache-decoder-config.yaml
@@ -1,5 +1,6 @@
-local_cpu: False
-max_local_cpu_size: 0
+local_cpu: True
+max_local_cpu_size: 10
+max_local_disk_size: 0
 
 enable_nixl: True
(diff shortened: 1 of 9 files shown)


sammshen:
This is a build off of https://github.com/LMCache/LMCache/pull/1207

I just saw https://github.com/LMCache/LMCache/pull/1162. Taking a look now and apologies for the delay @vladnosiv

PR #1223, opened 2025-08-04
[Core] Nixl storage backend

tshmilnvidia:
Adding support for new nixl backend, NixlStorageBackend. This backend supports storage operations via NIXL storage plugins.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -45,6 +45,7 @@ from lmcache.v1.memory_management import (  # noqa: E501
     MemoryObjMetadata,
     MixedMemoryAllocator,
     NixlCPUMemoryAllocator,
+    PagedTensorMemoryAllocator,
     TensorMemoryObj,
 )
 from lmcache.v1.storage_backend.storage_manager import StorageManager
@@ -1167,7 +1168,11 @@ class LMCacheEngineBuilder:
         metadata: LMCacheEngineMetadata,
         numa_mapping: Optional[NUMAMapping] = None,
     ) -> MemoryAllocatorInterface:
-        if config.enable_nixl:
+        extra_config = config.extra_config
+        enable_nixl_storage = extra_config is not None and extra_config.get(
+            "enable_nixl_storage"
+        )
+        if config.enable_nixl and not enable_nixl_storage:
             assert config.nixl_buffer_device is not None
             # TODO (Jiayi): make this less hacky
             if config.enable_xpyd:
@@ -1203,6 +1208,38 @@ class LMCacheEngineBuilder:
                     )
                 return nixl_cpu_mem_allocator
             return AdHocMemoryAllocator(config.nixl_buffer_device)
+
+        if enable_nixl_storage:
+            # First Party
+            from lmcache.v1.storage_backend.connector.nixl_utils import (
+                get_correct_nixl_device,
+            )
+
+            corrected_device = get_correct_nixl_device(
+                config.nixl_buffer_device,
+                metadata.worker_id,
+            )
+
+            buffer = torch.empty(
+                config.nixl_buffer_size,
+                dtype=torch.uint8,
+                device=corrected_device,
+            )
+
+            if corrected_device == "cpu":
+                torch.cuda.cudart().cudaHostRegister(
+                    buffer.data_ptr(), config.nixl_buffer_size, 0
+                )
+            else:
+                logger.info(f"Setting cuda device to {corrected_device} ")
+                torch.cuda.set_device(corrected_device)
+
+            return PagedTensorMemoryAllocator(
+                buffer,
+                torch.Size(metadata.kv_shape),
+                metadata.kv_dtype,
+                MemoryFormat.KV_2LTD,
+            )
 
         if config.weka_path is not None or config.gds_path is not None:
             assert config.cufile_buffer_size is not None
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -374,6 +374,9 @@ def _validate_config(self):
         assert self.lookup_url is not None
         assert self.distributed_url is not None
 
+    enable_nixl_storage = self.extra_config is not None and self.extra_config.get(
+        "enable_nixl_storage"
+    )
     if self.enable_nixl:
         assert self.nixl_role is not None
         assert self.nixl_buffer_size is not None
@@ -384,6 +387,13 @@ def _validate_config(self):
             "Nixl only supports save_decode_cache=False"
         )
         assert self.enable_p2p is False, "Nixl only supports enable_p2p=False"
+
+    if enable_nixl_storage:
+        assert self.extra_config.get("nixl_backend") is not None
+        assert self.extra_config.get("nixl_path") is not None
+        assert self.extra_config.get("nixl_file_pool_size") is not None
+        assert self.nixl_buffer_size is not None
+        assert self.nixl_buffer_device is not None
 
     return self
 
--- a/lmcache/v1/storage_backend/nixl_storage_backend.py
+++ b/lmcache/v1/storage_backend/nixl_storage_backend.py
@@ -57,9 +57,13 @@ class NixlStorageConfig:
     backends: list[str]
 
     @staticmethod
-    def validate_nixl_backend(backend: str):
-        supported_backends = ["GDS", "GDS_MT"]
-        return backend in supported_backends
+    def validate_nixl_backend(backend: str, device: str):
+        if backend in ("GDS", "GDS_MT"):
+            return device == "cpu" or device == "cuda"
+        elif backend in ("POSIX", "HF3FS"):
+            return device == "cpu"
+        else:
+            return False
 
     @staticmethod
     def from_cache_engine_config(
@@ -76,7 +80,7 @@ class NixlStorageConfig:
         assert extra_config.get("nixl_file_pool_size") is not None
 
         assert NixlStorageConfig.validate_nixl_backend(
-            extra_config.get("nixl_backend")
+            extra_config.get("nixl_backend"), config.nixl_buffer_device
         ), "Invalid NIXL backend & device combination"
 
         corrected_device = get_correct_nixl_device(
@@ -101,7 +105,7 @@ class NixlFilePool:
 
         for i in reversed(range(size)):
             tmp_path = path + f"obj_{i}_{uuid.uuid4().hex[0:4]}.bin"
-            fd = os.open(tmp_path, os.O_CREAT | os.O_RDWR | os.O_DIRECT)
+            fd = os.open(tmp_path, os.O_CREAT | os.O_RDWR)
             self.fds.append(fd)
             self.indices.append(i)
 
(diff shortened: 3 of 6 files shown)


lmcache/v1/config.py
```diff
@@ -86,6 +86,10 @@ class LMCacheEngineConfig:
     # (Optional) Nixl configurations
     # whether to enable Nixl
     enable_nixl: Optional[bool] = False
+    # whether to enable Nixl storage backend
+    enable_nixl_storage: Optional[bool] = False
+    #Select which NIXL backend to use for NIXL storage backend
+    nixl_backend: Optional[str] = None
```
YaoJiayi: Can we put these new configs to `extra_config`?

YaoJiayi: Also, can we have a runnable example?

lmcache/v1/config.py
```diff
@@ -700,7 +735,13 @@ def validate(self) -> "LMCacheEngineConfig":
             assert self.distributed_url is not None
 
         if self.enable_nixl:
-            assert self.nixl_role is not None
+            if self.enable_nixl_storage:
+                assert self.nixl_gds_path is not None
+                assert self.nixl_file_pool_size is not None
+                assert self.nixl_role is None
+                assert not self.local_cpu
```
YaoJiayi: Why do we need to disable cpu?

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -0,0 +1,402 @@
...
+            raise RuntimeError("NIXL transfer failed")
+
+    def close(self):
+        self.nixl_agent.release_dlist_handle(self.file_xfer_handler)
+        self.nixl_agent.release_dlist_handle(self.vram_xfer_handler)
+        self.nixl_agent.deregister_memory(self.file_reg_descs)
+        self.nixl_agent.deregister_memory(self.vram_reg_descs)
+
+class NixlStorageBackend(StorageBackendInterface):
+
```
YaoJiayi, line 227: Is this essentially a gpu direct backend? In other words, does the buffer has to be on GPU?

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -0,0 +1,402 @@
...
+        with self.key_lock:
+            metadata = self.key_dict.pop(key.chunk_hash, None)
+            if metadata is None:
+                return False
+
+        self.file_pool.push(metadata.address)
+        return True
+
+    def pin(self, key: CacheEngineKey) -> bool:
+        return False
```
YaoJiayi, line 416: The interaction with vLLM could be buggy without pin. Because vLLM will first lookup the cache and pin it and then retrieve it. There might be some time gap between lookup and retrieve.

YaoJiayi (commented):
Thanks for the great effort! I left some comments:)

lmcache/v1/config.py
tshmilnvidia: Yes, I will move those new configs to extra_config.

Regarding a runnable example: we did not currently add a runnable example as this is an initial draft version - we will have a runnable example for future versions. We can go into more detail into how we tested it so far if you think it's neccessary.

lmcache/v1/config.py
tshmilnvidia: This is only a limitation of this initial draft version. We are currently working on making this work without this limitation.

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: It is limited to GPU for this initial draft version - however we do plan to be able to support all NIXL storage backends (POSIX, 3FS, etc..) - which will in turn allow CPU buffers as well.

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: We would like to discuss this in more detail with you.

lmcache/v1/config.py
tshmilnvidia: Limitation removed.

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: CPU buffer support added.

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: As discussed, will be handled in a future PR.

lmcache/v1/config.py
tshmilnvidia: As discussed, will be handled in a future PR.

YaoJiayi:
@tshmilnvidia Sorry I might need to review it tmrw. Any idea on why unit testing is failing?

tshmilnvidia:
@YaoJiayi Added some fixes to pre-commit errors, CI now passes.

lmcache/v1/storage_backend/connector/nixl_utils.py
```diff
@@ -87,6 +87,58 @@ def from_cache_engine_config(
...
+        )
+
+        return NixlStorageConfig(
+            buffer_size=config.nixl_buffer_size,
+            file_pool_size=extra_config.get("nixl_file_pool_size"),
+            buffer_device=corrected_device,
+            path=extra_config.get("nixl_path"),
+            backends=[extra_config.get("nixl_backend")],
+        )
+
```
YaoJiayi: Any reason we put the nixl_utils under connector/ ?

lmcache/v1/storage_backend/nixl_storage_backend.py
YaoJiayi: Is it possible to reuse the cpu buffer in the offloading code path? (if cpu is the buffer device)

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -156,7 +162,9 @@ def batched_put(
         storage manager) or has been stored (handled by storage backend).
         """
 
-        if self.enable_nixl or (location and location == "NixlBackend"):
+        if self.enable_nixl or (
+            location and (location == "NixlBackend" or location == "NixlStorageBackend")
+        ):
```
YaoJiayi: I'm a bit confused about the logic here?
See the below comment.

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -191,7 +199,7 @@ def batched_put(
                 keys = cpu_keys
 
         for backend_name, backend in self.storage_backends.items():
-            if backend_name == "NixlBackend":
+            if backend_name == "NixlBackend" or backend_name == "NixlStorageBackend":
                 continue
```
YaoJiayi: I think nixl storage backend should be here?

YaoJiayi: What's the reason not to include nixl storage backend in this loop?

YaoJiayi (commented):
Thanks! Sorry for the delay. A few comments left:)

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: I'm not sure I understand what you mean here. Are you asking if we can use the buffer allocated for the local_cpu_backend (if exists) as the cpu buffer of the nixl_storage_backend?

lmcache/v1/storage_backend/connector/nixl_utils.py
tshmilnvidia: Moved to nixl_storage_backed.py

lmcache/v1/storage_backend/storage_manager.py
tshmilnvidia: Fixed.

lmcache/v1/storage_backend/storage_manager.py
tshmilnvidia: Fixed.

lmcache/v1/storage_backend/nixl_storage_backend.py
YaoJiayi: Yes, why do we need an extra cpu buffer here?

tshmilnvidia: I see your point. Yes, we should be able to use the already allocated buffer for local_cpu_backend if it exists.
However, we need the memory allocator to be PagedTensorMemoryAllocator, as it fits best for NIXL.
 
The way I see it is this way:
if local_cpu == false:
allocate a new buffer and make a PagedTensorMemoryAllocator (same as now)
 
if local_cpu == true:
configure the memory allocator (in advance) with use_paging=true. that way, the memory allocator will be MixedMemoryAllocator with the "pin_allocator" being a PagedTensorMemoryAllocator. Then we should be able to use this buffer for NIXL as well.
 
WDYT?

Shaoting-Feng: When `local_cpu == false`, shall we still use the local_cpu_backend, which aligns with all other backends? Code path: https://github.com/LMCache/LMCache/blob/99bcf07e3eeaf6890f28060fad90d22f80471bfe/lmcache/v1/storage_backend/storage_manager.py#L87.

Shaoting-Feng (commented):
General LGTM. One comment same as @YaoJiayi

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: Thanks for your comments! I have given it more thought and decided to seperate the "enable_nixl_storage" flag completely from "enable_nixl" - this now makes it behave a lot closer to the other backends, including also always using local_cpu_backend and not allocating an extra buffer in the CPU case.

lmcache/v1/cache_engine.py
```diff
@@ -1204,6 +1205,42 @@ def _Create_memory_allocator(
...
+
+            if corrected_device == "cpu":
+                torch.cuda.cudart().cudaHostRegister(
+                    buffer.data_ptr(), config.nixl_buffer_size, 0
+                )
+            else:
+                logger.info(f"Setting cuda device to {corrected_device} ")
+                torch.cuda.set_device(corrected_device)
+
+            return PagedTensorMemoryAllocator(
```
Shaoting-Feng, line 1237: From `if config.enable_nixl` above, will this function return `AdHocMemoryAllocator` instead of `PagedTensorMemoryAllocator` as expected?

tshmilnvidia: this would be true for all other backends as well, enable_nixl's allocator takes priority.

I would prefer to not allow enable_nixl and enable_nixl_storage togerther in order to avoid this, but I ultimately chose not to do this, in order to align with the behavior of other storage backends.

Shaoting-Feng: Sorry I am a bit confused. Which type of memory allocator does `NixlStorageBackend` need?

tshmilnvidia: PagedTensorMemoryAllocator

Shaoting-Feng: Then shall we change line 1171 from `if config.enable_nixl:` to `if config.enable_nixl and not enable_nixl_storage:`?

tshmilnvidia: Done

Shaoting-Feng (approved):
LGTM

YaoJiayi (approved):
lgtm

ziruiliu:
if `enable_nixl: True` is set in yaml file, lmcache/v1/config.py then requires 'nixl_role' to be either 'sender' or 'receiver' which does not make sense. I suggest removing line `enable_nixl: True` from example yaml file
see https://github.com/LMCache/LMCache/pull/1450

tshmilnvidia:
> if `enable_nixl: True` is set in yaml file, lmcache/v1/config.py then requires 'nixl_role' to be either 'sender' or 'receiver' which does not make sense. I suggest removing line `enable_nixl: True` from example yaml file see #1450

@ziruiliu Thank you, this is correct, I changed how enable_nixl_storage works together with enable_nixl during the review and didn't update the docs in accordance.

PR #1226, opened 2025-08-04
[Docs] Multimodality

sammshen:


YaoJiayi (approved):
@sammshen LGTM. Would you mind add a similar example under examples/ ?

hickeyma (changes requested):
Thanks @sammshen for pushing this PR.

Some feedback from reading it first off:

- Its not obvious from the explanation what is the value of LMCache here
- Maybe change title from "vLLM Multimodality Support" to "How to show reduced inferencing time using MultiModal models"
- Change "Explanation" to "Overview"
- Simplify the vLLM explanation to read simply as vLLM support multi-modal models and points to vLLM doc, for example:  https://docs.vllm.ai/en/stable/contributing/model/multimodal.html
- Then can mention that LMCache provides caching for multi-modal models in vLLM
- Maybe change "Basic Examples for each modality:" to "Example to demonstrate caching with multi-modal models
- In the example section, maybe start with explaining what the user will be doing. In other words, going to be running 4 types of inferencing: Audio Inference with Ultravox, Single Image Inference with Llava, Multi-image Inference with Phi-3.5-vision-instruct & Video Inference with Llava-OneVision 
- Now add sub-section for each type of inferencing
- In each sub-section, show command to run vLLM with LMCache for that modality
- Then explain that you can test caching by running the command (e.g. `python openai_chat_completion_client_for_multimodal.py --chat-type audio`) twice. The first time it will take x amount of time and 2nd time it will be greatly reduced due to the caching
- Repeat for other inference types
- There is also no need to show the test code from vLLM, just add link to the vLLM code base

sammshen:
Thanks so much for the comments: 
 I am a little afraid that the vllm codepath may change in the future and the link will break and also I made some small edits to the example (just adding the time measurements). Would like to make it copy pastable for documentation reader

sammshen:
@hickeyma are there any more changes you would like? can we merge this documentation? :)

hickeyma:
> @hickeyma are there any more changes you would like? can we merge this documentation? :)

I'lll get to it today. Looking at unit test issue at the moment.

docs/source/getting_started/installation.rst
```diff
@@ -100,8 +100,8 @@ LMCache with vLLM v0
 Setup using Docker
 ------------------
 
-Prerequisites
-~~~~~~~~~~~~~
+Prerequisites for vLLM v0
```
hickeyma: Why is the vLLM v0 prereq?

requirements/docs.txt
```diff
@@ -1,4 +1,4 @@
-Sphinx==8.2.3
+Sphinx==8.1.3
```
hickeyma: Why are we downgrading the sphinx version?

hickeyma (changes requested):
@sammshen I pushed updates to the api reference doc. Hope that is ok? It was easier than trying to add comments because of the changes. Let me know what you think.

I also have some questions inline on small changes.

docs/source/getting_started/installation.rst
sammshen: sphinx will complain about more than one: 
```text
Prerequisites
~~~~~~~~~~
```

and there is already one above. feel free to rename!

requirements/docs.txt
sammshen: the build failed with sphinx 8.2.3 for me but we can change it back if it doesn't break on your side (probably my local env specific issue)

sammshen:
> @sammshen I pushed updates to the api reference doc. Hope that is ok? It was easier than trying to add comments because of the changes. Let me know what you think.
> 
> I also have some questions inline on small changes.

LGTM, thanks so much for cleaning up these docs!!

YaoJiayi:
@hickeyma Is this PR ready to be merged?

docs/source/getting_started/installation.rst
hickeyma: Sphinx warns about duplicate ids but its works as if you copy the link for the 2nd pre-req section in docker, it will give a random id instead. This is not an issue for the user as it works as expected. Going to roll it back as section title is confusing for the user.

requirements/docs.txt
hickeyma: Sphinx `8.2.3` works for me, so going to roll back. We can change it if we notice any issues again.

sammshen: great!

hickeyma (approved):
LGTM, thanks @sammshen.

Unit test failing are not related to this PR, so merging as its doc only PR.

PR #1235, opened 2025-08-05
[docker] remove unused patch files

panpan0000:
Just remove old patch files from docker build


panpan0000:
I'm a bit confused.

in theory, vLLM v0 implementation(like `vllm/worker/model_runner.py` ) still requires some function like `recv_kv_caches_and_hidden_states` in this patch folder.

but there's no any code in LMCache dev branch using this patch folder.

I'm not sure if this patch folder deleted, will it break vLLM v0 ?

hickeyma (commented):
@panpan0000 It might be best wait until the v0 support is deprecated before removing this code.

What do you think @YaoJiayi?

panpan0000:
> @panpan0000 It might be best wait until the v0 support is deprecated before removing this code.
> 
> What do you think @YaoJiayi?

agree @hickeyma , 
but I didn't find anywhere to use this patch files in current code (either docker or wheel). 
So what if they are deleted, what will happen ?

hickeyma:
> but I didn't find anywhere to use this patch files in current code (either docker or wheel).
So what if they are deleted, what will happen ?

I can't see either, just want to get confirmation

hickeyma (approved):

YaoJiayi (approved):
LGTM. We can remove them for now

PR #1242, opened 2025-08-05
[Core] move os.remove from submit_put_task thread to worker thread

RangerCD:
While submitting a put task to `LocalDiskBackend`, some keys will be evicted if needed, so all `os.remove` are synchronously called in current thread(`submit_put_task`->`self.remove`->`os.remove`).

`os.remove` is a POSIX system call, which means it should be considered as a slow operation, and the whole `store` function in `LMCacheEngine` will have to wait every single system call.

Since there is already a `LocalDiskWorker` for I/O operations, it's more suitable to submit remove operation to this worker and execute it on another thread.


Shaoting-Feng (approved):
LGTM. Could you fix the failing unit test?

RangerCD:
@Shaoting-Feng unit tests fixed.

Shaoting-Feng (approved):
LGTM

ApostaC (commented):
LGTM!

RangerCD:
The priority of local disk tasks adjusted to:

prefetch: 0
delete: 0->1
put: 1->2

In order to make sure delete operation doesn't block any prefetch

RangerCD:
@YaoJiayi please have a look

YaoJiayi (approved):
lgtm

PR #1254, opened 2025-08-06
[Bugfix]: dont pin on second lookup

sammshen:
1. avoid memory leak when lookup happens twice in a single vllm schedule step for a single request

2. move unpin to wait_for_save instead of start_load_kv to avoid race conditions with layerwise (some deeper layers getting evicted because cache_engine.retriee() only gets the first layer on first call)

long document qa brief benchmark:
```text
=== BENCHMARK RESULTS ===
Warmup round mean TTFT: 1.213s
Warmup round time: 3.094s
Warmup round prompt count: 24
Query round mean TTFT: 0.957s
Query round time: 3.027s
Query round prompt count: 24
```


YaoJiayi (approved):
lgtm

PR #1256, opened 2025-08-06
Fix the config logic of `create_lookup_server_only_on_worker_0`

maobaolong:
The bug this PR will fix is introduced by #1144

As we intend to set the default value of `create_lookup_server_only_on_worker_0` to `true`, so we need to fix by this PR.

### DIFF
--- a/lmcache/v1/lookup_client/factory.py
+++ b/lmcache/v1/lookup_client/factory.py
@@ -72,8 +72,8 @@ class LookupClientFactory:
         # Only create the KV lookup API server on worker rank 0
         # when there are multiple workers and when not using external lookup client
         create_lookup_server_only_on_worker_0 = (
-            config.extra_config
-            and config.extra_config.get("create_lookup_server_only_on_worker_0", True)
+            not config.extra_config
+            or config.extra_config.get("create_lookup_server_only_on_worker_0", True)
         )
         if config.external_lookup_client is None and (
             not create_lookup_server_only_on_worker_0
--- a/lmcache/v1/lookup_client/factory.py
+++ b/lmcache/v1/lookup_client/factory.py
@@ -71,12 +71,14 @@ class LookupClientFactory:
 
         # Only create the KV lookup API server on worker rank 0
         # when there are multiple workers and when not using external lookup client
-        create_lookup_server_only_on_worker_0 = (
-            not config.extra_config
-            or config.extra_config.get("create_lookup_server_only_on_worker_0", True)
+        create_lookup_server_only_on_worker_0_for_mla = (
+            config.extra_config
+            and config.extra_config.get(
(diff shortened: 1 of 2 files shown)


maobaolong:
@Shaoting-Feng @ApostaC Would you like to take a look at this PR? This is a simple bugfix, thanks!

Shaoting-Feng (commented):
Will this cause a problem in the tp case?

maobaolong:
@Shaoting-Feng For ds mla， we only  use tp0 to store and  retrieve, so without  this PR,tp n>0 will return 0 when lookup

Shaoting-Feng (approved):
LGTM

ApostaC:
@maobaolong Just wanted to understand the behavior here.

Does that mean we will always only create lookup server on worker 0, unless `config.extra_config.create_lookup_server_only_on_worker_0 == False`?

I'm thinking about the behavior that the user may expect. I feel like most of the models are not MLA, so we may still want to have look-up servers on all TP workers?

Hadigan:
> @maobaolong Just wanted to understand the behavior here.
> 
> Does that mean we will always only create lookup server on worker 0, unless `config.extra_config.create_lookup_server_only_on_worker_0 == False`?
> 
> I'm thinking about the behavior that the user may expect. I feel like most of the models are not MLA, so we may still want to have look-up servers on all TP workers?

I agree with this perspective. Regarding this, I'd like to propose two suggestions:

Firstly, I think the default setting for `create_lookup_server_only_on_worker_0` should be False. This flag should only be explicitly set to True by users when using MLA models. The current default (True) causes tensor parallelism to fail, while setting False as default ensures compatibility for both TP scenarios and MLA models？

Secondly, I'd suggest considering either:
1. Renaming this flag to include an MLA-specific identifier (e.g., create_lookup_server_only_on_worker_0_for_mla) to make MLA users aware of its purpose, or
2. Removing this flag entirely and instead inferring MLA requirements automatically from the model metadata at the vLLM layer. This would simplify configuration while maintaining functionality.

maobaolong:
@Hadigan Thanks for your suggestion! I make it default to False and rename it, PTAL

maobaolong:
Sorry to ping @Hadigan , would you mind to take a look whether the new state of this PR ok for you?

Hadigan:
> Sorry to ping @Hadigan , would you mind to take a look whether the new state of this PR ok for you?

The changes look good to me. 😊

maobaolong:
@sammshen Would you like to take another look?

ApostaC (approved):
LGTM!

ApostaC:
Sorry, I was trying to do the comprehensive test, but it seems like the CI machine is offline. I will merge this PR now.

maobaolong:
@ApostaC Thanks a lot!

PR #1266, opened 2025-08-07
Refactor config to make it easy to extend

maobaolong:


ApostaC (approved):
LGTM!

ApostaC:
@YaoJiayi @Shaoting-Feng @sammshen Can one of you take a look at this?

sammshen:
LGTM but defer final decision to @YaoJiayi 

does any documentation need to be updated here? https://docs.lmcache.ai/api_reference/configurations.html

sammshen (approved):

maobaolong:
@sammshen If correct, i should just do a refactor only, do not remove or add configuration. 

image

image

And I checked the affected config is correct and same with my following config.

```yaml
chunk_size: 256
local_device: "cpu"
remote_url: "fs://host:0/apdcephfs_szgm/share_303468461/baoloongmao/fs_data/"
  #remote_url: "fs://host:0/disc/data1/baoloongmao/fs_data/dir0"
remote_serde: "naive"
pipelined_backend: False
local_cpu: False
max_local_cpu_size: 10
save_decode_cache: False
extra_config:
  remote_enable_mla_worker_id_as0: True
```

PR #1268, opened 2025-08-07
[fix]: race condition on multi-lookup then multi-forward-passes on the same request

sammshen:
this looks kind of weird
```python
lookup_hash = "".join(random.choices(string.ascii_letters + string.digits, k=8))
        lookup_id = request.request_id + "###" + lookup_hash
```

@YaoJiayi I will use whatever you suggest

I am leaning toward just using:
```python
lookup_id = "".join(random.choices(string.ascii_letters + string.digits, k=16))
```

to avoid string concatenation but looks a little weird as well.

alternative is to switch to ints using `uuid`s?  (this is actually my top choice)


YaoJiayi:
@sammshen Let's do uuid. Btw, this won't break the controller functionalities right?

sammshen:
This should not affect the controller functionalities at all. I will test the move and compress endpoints and update here!

sammshen:
@YaoJiayi tested long doc qa and the compress and move endpoints, sorry for delay!

```text
=== BENCHMARK RESULTS ===
Warmup round mean TTFT: 23.704s
Warmup round time: 47.737s
Warmup round prompt count: 50
Query round mean TTFT: 3.666s
Query round time: 14.031s
Query round prompt count: 100
```

YaoJiayi (commented):
Sorry for requesting another change here again. Can we do str(uuid) to make type hints a lot simpler. Sorry I forgot uuid will not return str directly

sammshen:
thanks for the feedback! changed types back to str and ran all 3 experiments again!

sammshen:
also added some unit test fixes

YaoJiayi (approved):
lgtm

PR #988, opened 2025-07-07
feat: integrate mooncake backend for enhanced performance

xiaguan:
This PR moves Mooncake up from the remote connector abstraction to the backend layer, which helps enable more efficient zero-copy batch_put and batch_get operations. It also allows Mooncake to function as a standalone backend, opening the door for future fast-path optimizations.

I'll run a more accurate performance comparison after PR (https://github.com/LMCache/LMCache/pull/936) is merged. In theory, combining these two PRs should get us to about 85% of the potential max performance. The remaining gap is mostly due to things like for-loop contains checks and some room for improvement in batched_to_gpu. Those are optimizations I plan to tackle down the line.

---
Both use `external_lookup_client: "mooncakestore://localhost:50052"`

### Mooncake Backend Performance （qwen-7b)

| Metric                               | baseline<br>(mooncake-connector) | optimized<br>(mooncake-store) | ideal<br>(gpu-kv-cache) |
| ------------------------------------ | ------------------------------- | ---------------------------- | ------------------------ |
| **Successful requests**              | 50                              | 50                           | 50                       |
| **Duration (s)**                     | 8.28                            | 5.37                         | 4.79                     |
| **Total input tokens**               | 409 494                         | 409 494                      | 409 494                  |
| **Total generated tokens**           | 6 400                           | 6 400                        | 6 400                    |
| **Request throughput (req/s)**       | 6.04                            | 9.32                         | 10.43                    |
| **Output token throughput (tok/s)**  | 772.85                          | 1 192.52                     | 1 334.87                 |
| **Total token throughput (tok/s)**   | 50 222.75                       | 77 494.35                    | 86 744.40                |
| **Mean TTFT (ms)**                   | 3 530.17                        | 925.00                       | 640.29                   |
| **Median TTFT (ms)**                 | 3 553.70                        | 947.27                       | 645.90                   |
| **P99 TTFT (ms)**                    | 4 573.48                        | 1 585.86                     | 1 214.33                 |
| **Mean TPOT (ms, excl. 1st)**        | 37.13                           | 34.21                        | 30.99                    |
| **Median TPOT (ms, excl. 1st)**      | 36.96                           | 34.14                        | 31.34                    |
| **P99 TPOT (ms)**                    | 58.73                           | 39.21                        | 32.62                    |
| **Mean ITL (ms)**                    | 37.13                           | 34.21                        | 31.43                    |
| **Median ITL (ms)**                  | 29.12                           | 29.19                        | 29.17                    |
[...]

PR #1269, opened 2025-08-07
 perf(mooncakestore connector): optimize get/put with zero-copy operations

xiaguan:
another version of #988 

Although I've run multiple benchmark, the performance doesn't match that of PR #988. This might be because the remote connector requires additional code paths to handle operations. However, it still provides significant performance improvements.

optimized connector
```
Successful requests:                     50        
Benchmark duration (s):                  5.72      
Total input tokens:                      404689    
Total generated tokens:                  6400      
Request throughput (req/s):              8.74      
Output token throughput (tok/s):         1118.59   
Total Token throughput (tok/s):          71849.95  
Mean TTFT (ms):                          1540.05   
Median TTFT (ms):                        1477.40   
P99 TTFT (ms):                           2029.28   
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          32.74     
Median TPOT (ms):                        33.24     
P99 TPOT (ms):                           43.12     
---------------Inter-token Latency----------------
Mean ITL (ms):                           32.74     
Median ITL (ms):                         29.06     
P99 ITL (ms):                            42.61     
```

compare to result in #988, throuput 77494 tok/s, mean ttft: 925ms, media ttft: 947ms

### DIFF
--- a/lmcache/v1/storage_backend/connector/mooncakestore_connector.py
+++ b/lmcache/v1/storage_backend/connector/mooncakestore_connector.py
@@ -12,6 +12,7 @@ import os
 import torch
 
 # First Party
+from lmcache.config import LMCacheEngineMetadata
 from lmcache.logging import init_logger
 from lmcache.utils import CacheEngineKey
 from lmcache.v1.config import LMCacheEngineConfig
@@ -36,6 +37,7 @@ class MooncakeStoreConfig:
     master_server_address: str
     transfer_timeout: int
     storage_root_dir: str
+    prefer_local_alloc: bool = False
 
     @staticmethod
     def from_file(file_path: str) -> "MooncakeStoreConfig":
@@ -52,6 +54,7 @@ class MooncakeStoreConfig:
             master_server_address=config.get("master_server_address"),
             transfer_timeout=config.get("transfer_timeout", 1),
             storage_root_dir=config.get("storage_root_dir", ""),
+            prefer_local_alloc=config.get("prefer_local_alloc", False),
         )
 
     @staticmethod
@@ -82,6 +85,7 @@ class MooncakeStoreConfig:
             master_server_address=extra_config["master_server_address"],
             transfer_timeout=extra_config.get("transfer_timeout", 1),
             storage_root_dir=extra_config.get("storage_root_dir", ""),
+            prefer_local_alloc=extra_config.get("prefer_local_alloc", False),
         )
 
 
@@ -97,7 +101,7 @@ class MooncakestoreConnector(RemoteConnector):
     ):
         try:
             # Third Party
-            from mooncake.store import MooncakeDistributedStore
+            from mooncake.store import MooncakeDistributedStore, ReplicateConfig
         except ImportError as e:
             raise ImportError(
                 "Please install mooncake by following the instructions at "
@@ -133,6 +137,15 @@ class MooncakestoreConnector(RemoteConnector):
                     "Set MOONCAKE_STORAGE_ROOT_DIR to: %s", self.config.storage_root_dir
                 )
 
+            logger.info("Setting up Mooncake store with parameters:")
+            logger.info(f"  local_hostname: {self.config.local_hostname}")
+            logger.info(f"  metadata_server: {self.config.metadata_server}")
+            logger.info(f"  global_segment_size: {self.config.global_segment_size}")
+            logger.info(f"  local_buffer_size: {self.config.local_buffer_size}")
+            logger.info(f"  protocol: {self.config.protocol}")
+            logger.info(f"  device_name: {self.config.device_name}")
+            logger.info(f"  master_server_address: {self.config.master_server_address}")
+
             self.store.setup(
                 self.config.local_hostname,
                 self.config.metadata_server,
@@ -142,6 +155,7 @@ class MooncakestoreConnector(RemoteConnector):
                 self.config.device_name,
                 self.config.master_server_address,
             )
+            logger.info("Mooncake store setup completed successfully")
 
         except ValueError as e:
             logger.error("Configuration loading failed: %s", e)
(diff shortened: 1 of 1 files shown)


xiaguan:
@maobaolong @Shaoting-Feng 

Are you available to review this? I've modified it to the connector level based on the feedback, which introduces a slight performance overhead, but overall it still delivers substantial performance improvements.

Hope we can merge either of these optimizations soon. Thanks!

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -130,6 +130,24 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
         """
         raise NotImplementedError
 
+    async def batched_get(
```
maobaolong: This seems duplicated with  L209

lmcache/v1/storage_backend/connector/instrumented_connector.py
```diff
@@ -59,6 +59,30 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
     async def exists(self, key: CacheEngineKey) -> bool:
         return await self._connector.exists(key)
 
+    async def batched_get(
```
maobaolong: This is duplicated to L104

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -139,11 +334,277 @@ def __init__(
...
+        """Unregister CPU buffer."""
+        if self.registered_buffer_ptr is not None:
+            result = self.store.unregister_buffer(self.registered_buffer_ptr)
+            if result == 0:
+                logger.info(f"Unregistered buffer: {hex(self.registered_buffer_ptr)}")
+            else:
+                logger.warning(f"Buffer unregistration failed: error={result}")
+            self.registered_buffer_ptr = None
+
+    async def batched_get(
```
maobaolong, line 249: You have to override this `support_batched_get` method and return true for mooncake

maobaolong (changes requested):
@xiaguan Thanks for this PR, left some comments.

xiaguan:
@maobaolong Thanks for your review. All issues have been addressed. Could you please take another look?

I've only modified the `mooncakestore_connector.py` file - could we expedite the merge process?

maobaolong:
@xiaguan Thanks for help me and @chunxiaozheng to go through the code by tencent metting, it really help us to know well to this PR quickly. Will give you some fead back this week. 

First of all this PR really supply some performance improvement points by batched operation and zero copy.

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -152,33 +166,232 @@ def __init__(
...
+        # Set preferred_segment based on configuration
+        if self.config.prefer_local_alloc:
+            self.replica_config.preferred_segment = self.store.get_hostname()
+
+        # Register CPU buffer for zero-copy operations
+        self._register_cpu_buffer()
+
+        logger.info("MooncakeConnector initialized successfully.")
+
+    def init_chunk_meta(
```
chunxiaozheng, line 184: The log has been printed in parent class `RemoteConnector`, is the `init_chunk_meta` method need to override here?

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -152,33 +166,232 @@ def __init__(
...
+        This is used when save_chunk_meta=True (metadata stored with data).
+        """
+        key_strs = [key.to_string() for key in keys]
+
+        try:
+            buffers = await asyncio.to_thread(self.store.batch_get_buffer, key_strs)
         except Exception as e:
-            logger.error(f"Failed to get key {key_str}. {e}")
+            logger.error(f"batch_get_buffer failed: {str(e)}")
+            return [None] * len(keys)
```
maobaolong, line 359: If we sallow this exception, the vllm will generate wrong output.

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -152,33 +166,232 @@ def __init__(
...
+        for i, buffer in enumerate(buffers):
+            if buffer is None:
+                logger.warning(f"Buffer {i} is None for key {key_strs[i]}")
+                results.append(None)
+                continue
+            try:
+                memory_obj = self._process_buffer_with_metadata(buffer)
+                results.append(memory_obj)
+            except Exception as e:
+                logger.error(
```
maobaolong, line 371: This is another place for silent error handling.

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -213,19 +426,69 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
+                    self.store.put_from,
+                    key_str,
+                    buffer_ptr,
+                    buffer_size,
+                    self.replica_config,
+                ),
+                timeout=self.config.transfer_timeout,
+            )
+        except asyncio.TimeoutError:
+            logger.warning(
```
maobaolong, line 461: I guess we need to raise the exception so that caller can aware and handling.

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -234,20 +497,23 @@ async def put(self, key: CacheEngineKey, memory_obj: MemoryObj):
             )
         except asyncio.TimeoutError:
             logger.warning(
-                f"Timeout when putting key {key_str} from mooncake store."
+                f"Timeout when putting key {key_str} using put_parts. "
```
maobaolong, line 497: same silent error handling.

maobaolong (approved):
@xiaguan This PR look good overal left some exception handling related comments inline, otherwise, it LGTM. thank for your improvement.

maobaolong:
@xiaguan After a discussion with @chunxiaozheng offline, I aware that we should not raise the exception, return the mem_object we read or None is enough, because there is another work is right doing  parallel by others. For this PR, it is ok for me to keep it not worse than before.

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
xiaguan: Thanks. Fixed

xiaguan:
Thank you both for your thorough reviews - all comments have been resolved.

chunxiaozheng (approved):
lgtm

byte-ss:
Could you please explain why the method batched_submit_put_task() was not included in this commit? I believe it would also bring performance improvements.

xiaguan:
> Could you please explain why the method batched_submit_put_task() was not included in this commit? I believe it would also bring performance improvements.

Write requests are executed asynchronously in the background and aren't on the critical path.

There might be some benefit to this approach - feel free to give it a try.

ANormalMan12:
Nice

PR #1274, opened 2025-08-07
[Fix]: Pin torch version again

### LINKED ISSUE #1273: [Bug]: Install editable torch/CUDA failed to link
Currently, `pip install -e .` causes a bug because the torch version is not pinned after https://github.com/LMCache/LMCache/commit/ea063de5ca187244890ca06bfa4c0db2f40bdd6f (purpose was to avoid versioning collision when upstream vllm pip install lmcache in their Dockerfile).

The error is: 
```text
ERROR 08-07 19:58:38 [core.py:632] ImportError: /home/ubuntu/LMCache/lmcache/c_ops.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN3c104cuda9SetDeviceEab
```

To reproduce:
```bash
pip install -e .
```

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1273

The issue arose when torch 2.8.0 was released. This caused the build stage torch from the runtime torch versions ABI to diverge and cause linking issues. 

The cause of unpinning torch version was: https://github.com/LMCache/LMCache/issues/1184

This means we need to quickly watch for when torch versions change in vllm and update lmcache quickly

WIP: looking for a way to insert torch version dynamically @maobaolong

### DIFF
--- a/requirements/build.txt
+++ b/requirements/build.txt
@@ -4,7 +4,7 @@ ninja
 packaging>=24.2
 setuptools>=77.0.3,<81.0.0
 setuptools_scm>=8
-# don't pin torch version to avoid breaking vllm dockerfile (contains lmcache wheels)
-torch
+# please update ASAP when vllm torch updates
+torch==2.7.1
 wheel
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,13 @@
 [build-system]
 # Requirements for package build
 # Should be mirrored in requirements/build.txt (for builds external to setuptools)
+# the torch version in requirements/common.txt MUST match
 requires = [
     "ninja",
     "packaging>=24.2",
     "setuptools>=77.0.3,<81.0.0",
     "setuptools_scm>=8",
-    "torch",
+    "torch==2.7.0",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"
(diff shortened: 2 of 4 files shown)


YaoJiayi (approved):
Lgtm

requirements/build.txt
```diff
@@ -4,7 +4,7 @@ ninja
 packaging>=24.2
 setuptools>=77.0.3,<81.0.0
 setuptools_scm>=8
-# don't pin torch version to avoid breaking vllm dockerfile (contains lmcache wheels)
-torch
+# please update ASAP when vllm torch updates
+torch==2.7.1
```
maobaolong, line 8: @sammshen As discussed  in slack, we can avoid  pin torch  version?

maobaolong (commented):
@sammshen Left a comment

requirements/build.txt
sammshen: sorry this is temporary patch for now so people can use: 
```python
pip install -e . 
```

will copy what you said from slack here for reference:
"if we can supply a way to make it parameterized, it could be best!"

Should try to meet the following goals:
1. the wheels build properly and always compatible with latest vllm
2. the dockerfile builds properly
3. pip install -e . can work

PR #1299, opened 2025-08-11
Misc Upstreaming

sdimitro:
This PR attempts to upstream a couple of commits:

1. A commit that takes advantage of the newer `batched_get_blocking()` interface to add some "I/O threads" to the weka backend. A configuration options has been added that's generic enough so other backends can use it.

2. A minor commit that fixes some hardcoded magic-number in the weka backend.

3. A commit that fixes something that seems to be a regression in `CacheEngineKey` introduced in https://github.com/LMCache/LMCache/commit/2f039eb9c610bc60ab6cf4ea6e2d53269e75f995 - the basic issue is that `chunk_hash`es are in general base-16 with no `0x` prefix and thus just doing `int(chunk_hash_string)` fails when importing and creating keys from disk-based backends.

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -269,7 +269,7 @@ class LayerCacheEngineKey(CacheEngineKey):
             parts[1],
             int(parts[2]),
             int(parts[3]),
-            int(parts[4]),
+            int(parts[4], 16),
             tags,
             int(parts[5]),
         )
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -157,7 +157,7 @@ class CacheEngineKey:
                     raise ValueError(f"Invalid key string: {s}")
                 tags[kvs[0]] = kvs[1]
         return CacheEngineKey(
-            parts[0], parts[1], int(parts[2]), int(parts[3]), int(parts[4]), tags
+            parts[0], parts[1], int(parts[2]), int(parts[3]), int(parts[4], 16), tags
         )
 
     def to_dict(self):
--- a/lmcache/v1/storage_backend/weka_gds_backend.py
+++ b/lmcache/v1/storage_backend/weka_gds_backend.py
@@ -208,7 +208,7 @@ class WekaGdsBackend(StorageBackendInterface):
                         if not fentry.name.endswith(target_suffix):
                             continue
                         filename = os.path.basename(fentry.name)
-                        key_str = filename[:-14].replace("_", "/")
+                        key_str = filename[: -len(target_suffix)].replace("_", "/")
                         try:
                             key = CacheEngineKey.from_string(key_str)
                         except ValueError as e:
--- a/lmcache/v1/storage_backend/weka_gds_backend.py
+++ b/lmcache/v1/storage_backend/weka_gds_backend.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
 from collections import OrderedDict
-from concurrent.futures import Future
+from concurrent.futures import Future, ThreadPoolExecutor
 from typing import List, Optional, Tuple
 import asyncio
 import ctypes
@@ -153,14 +153,14 @@ class WekaGdsBackend(StorageBackendInterface):
         self.put_tasks: set[CacheEngineKey] = set()
 
         self.rand = random.Random(self.dst_device)
+        thread_count = config.extra_config.get("gds_io_threads", 4)
+        self._thread_pool = ThreadPoolExecutor(
+            max_workers=thread_count, thread_name_prefix="weka-gds-io"
+        )
 
         self._cufile_driver = self.cufile.CuFileDriver()
-        if hasattr(self.memory_allocator, "base_pointer"):
-            logger.debug(f"Using base pointer {self.memory_allocator.base_pointer}")
-            self.cufile_base_pointer = self.memory_allocator.base_pointer
-        else:
-            logger.info("No base pointer found, cufile will use bounce buffers")
-            self.cufile_base_pointer = None
+        assert hasattr(self.memory_allocator, "base_pointer")
+        self.cufile_base_pointer = self.memory_allocator.base_pointer
         asyncio.run_coroutine_threadsafe(self._scan_metadata(), self.loop)
         self.save_metadata_tasks: set[asyncio.Task] = set()
 
@@ -381,7 +381,7 @@ class WekaGdsBackend(StorageBackendInterface):
         dtype: torch.dtype,
         shape: torch.Size,
     ) -> Optional[MemoryObj]:
-        return self._load_bytes_from_disk(key, path, dtype, shape)
+        return self._load_bytes_from_disk_with_allocation(key, path, dtype, shape)
 
     def get_blocking(
         self,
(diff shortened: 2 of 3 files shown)


lmcache/utils.py
```diff
@@ -124,7 +124,7 @@ def from_string(s):
         if len(parts) != 5:
             raise ValueError(f"Invalid key string: {s}")
         return CacheEngineKey(
-            parts[0], parts[1], int(parts[2]), int(parts[3]), int(parts[4])
+            parts[0], parts[1], int(parts[2]), int(parts[3]), int(parts[4], 16)
```
maobaolong: Why there is a 16 here? @sdimitro

YaoJiayi:
@sdimitro Sorry for the delay. Might need to review this tmrw:)

lmcache/utils.py
sdimitro: Hi @maobaolong!

If you read my PR description I talk about this:
> seems to be a regression in CacheEngineKey introduced in https://github.com/LMCache/LMCache/commit/2f039eb9c610bc60ab6cf4ea6e2d53269e75f995 - the basic issue is that chunk_hashes are in general base-16 with no 0x prefix and thus just doing int(chunk_hash_string) fails when importing and creating keys from disk-based backends.

Basically the `chunk_hash` of `CacheEngineKey` was changed from an `str` to an `int`. The problem is that most of these strings used to look like this -> `ac9e9f5fed8e440bd32d41ceb59142884f903725fda622d63df483146c1a97be`

Now if you try to just use plain `int()` for that you'll get this error:
```
$ python3
>>> int('ac9e9f5fed8e440bd32d41ceb59142884f903725fda622d63df483146c1a97be')
  File "<python-input-0>", line 1, in <module>
    int('ac9e9f5fed8e440bd32d41ceb59142884f903725fda622d63df483146c1a97be')
    ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'ac9e9f5fed8e440bd32d41ceb59142884f903725fda622d63df483146c1a97be'
```

But if you add the `16` part as I do above then it works correctly:
```
>>> int('ac9e9f5fed8e440bd32d41ceb59142884f903725fda622d63df483146c1a97be', 16)
78078071756438414486485023651038479058945413968205959365197822800621458855870
```

Hope that makes sense.

sdimitro:
> @sdimitro Sorry for the delay. Might need to review this tmrw:)

All good @YaoJiayi ! BTW the person above asked about the potential regression that I found and I provided a more thorough explaination. Let me know if that makes sense and if it is also the way to go about it. Happy to change it to whatever is better if you know of something.

lmcache/v1/storage_backend/weka_gds_backend.py
```diff
@@ -153,14 +153,14 @@ def __init__(
         self.put_tasks: set[CacheEngineKey] = set()
 
         self.rand = random.Random(self.dst_device)
+        thread_count = config.gds_io_threads or 4
+        self._thread_pool = ThreadPoolExecutor(
```
maobaolong, line 157: Can you explain why you use threadpool instead of `asyncio` ?

sdimitro: Sure - in order to use `asyncio` the `cufile` APIs that we use would also need to be asynchronous. When we did our initial perf testing with cufile we tried 3 different APIs - their sync, async, and batch. From all of those their sync  APIs worked best for Weka. Since the sync APIs are the best for us and don't work with `asyncio` we used a thread pool. I hope that makes sense.

Aside: We also suspect for most other filesystems the cuFile sync APIs work best because for NIXL they also preferred to use those APIs (disclaimer: we, as Weka, actually wrote that NIXL code too but it was under NVIDIA's guidance and the code there is generic unlike this weka backend in LMCache) - see https://github.com/ai-dynamo/nixl/pull/519

lmcache/v1/config.py
```diff
@@ -185,6 +185,11 @@ def _to_int_list(
         "default": None,
         "env_converter": int,
     },
+    "gds_io_threads": {
+        "type": Optional[int],
+        "default": None,
+        "env_converter": int,
+    },
```
YaoJiayi: It's better to make this as an extra config?

lmcache/v1/storage_backend/weka_gds_backend.py
```diff
@@ -445,6 +445,81 @@ def _load_bytes_from_disk(
...
+            key: Cache key for error handling
+            path: File path to load from
+            dtype: Data type for memory allocation
+            shape: Shape for memory allocation
+
+        Returns:
+            A new memory object with loaded data, or None if allocation or
+            loading failed
+        """
+        memory_obj = self.memory_allocator.allocate(shape, dtype)
```
YaoJiayi, line 469: @sdimitro Sorry, can you remind me how does this memory allocator behave?

YaoJiayi (commented):
LGTM. Just a few minor comments

lmcache/v1/storage_backend/weka_gds_backend.py
sdimitro: Sure thing!

Here's the meat of it:
```
class CuFileMemoryAllocator(GPUMemoryAllocator):
    def __init__(self, size: int, device=None):
        ...
        from cufile.bindings import cuFileBufDeregister, cuFileBufRegister

        self.cuFileBufDeregister = cuFileBufDeregister
        ...
        super().__init__(size, device, align_bytes=4096)
        self.base_pointer = self.tensor.data_ptr()
        cuFileBufRegister(ctypes.c_void_p(self.base_pointer), size, flags=0)

    def __del__(self):
        self.cuFileBufDeregister(ctypes.c_void_p(self.base_pointer))
```
It's basically a wrapper on top of `TensorMemoryAllocator/GPUMemoryAllocator` for now. It pre-allocates a GPU buffer that it registers with cuFile for GDS I/O (the buffer is the self.tensor). If we didn't pre-register the buffer then cuFile would use bounce-buffers in the CPU which would involve basically copying the data twice instead of once with pre-registration. The actual allocation behavior is the same as `TensorMemoryAllocator.allocate()`.

Does this make sense?

lmcache/v1/config.py
sdimitro: Will look into that and update the PR 👍

sdimitro: Done

ApostaC (approved):
LGTM In general! Just wondering will the optimizations be also useful in the current  `gds_backend`?

sdimitro:
> Just wondering will the optimizations be also useful in the current gds_backend?

It probably would but since we don't use it at Weka and it has diverged in a few ways (including their on-disk format) I'm hesitant to make changes to it from fear of introducing any regressions.

sdimitro:
@YaoJiayi cleaned up the conflict from https://github.com/LMCache/LMCache/commit/4517b07ca1dd8eb4f181ba2a812ba35b5783d287

I also found a small bug/omission there for the Weka backend where one replacement of `get_size()` to `get_physical_size()` was missed. Regardless, I really appreciate you playing well with all the backends and updating ours.

lmcache/v1/storage_backend/weka_gds_backend.py
YaoJiayi: Makes sense to me:)

YaoJiayi (approved):
lgtm

PR #1308, opened 2025-08-12
Use dedicated read-write functions and eliminate redundant codes in local disk backend

llc-kc:
Use dedicated read-write functions and eliminate redundant codes in local disk backend


YaoJiayi (commented):
LGTM. Can we move all the debug logs measuring read and write bandwidth to `read_file` and `write_file`?

llc-kc:
@YaoJiayi Ok, finished.

YaoJiayi (approved):
Thanks @llc-kc !

PR #1320, opened 2025-08-12
Support addon num_mtp_layers from num_nextn_predict_layers

maobaolong:


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -520,6 +537,7 @@ def __init__(
                 vllm_config.cache_config,
                 vllm_config.scheduler_config,
                 config,
+                vllm_config,
```
Shaoting-Feng, line 552: Shall we put `vllm_config.speculative_config` here like the first four arguments?

Shaoting-Feng (approved):
LGTM with one minor comment

lmcache/integration/vllm/vllm_v1_adapter.py
maobaolong: @Shaoting-Feng 
Would you mind I change the signature of `init_lmcache_engine` from 
```python
def init_lmcache_engine(
    model_config: ModelConfig,
    parallel_config: ParallelConfig,
    cache_config: CacheConfig,
    scheduler_config: SchedulerConfig,
    speculative_config: SpeculativeConfig,
    config: LMCacheEngineConfig,
) -> Optional[LMCacheEngine]:
```
to 

```python
def _init_lmcache_engine(vllm_config, lmcache_config) -> Optional[LMCacheEngine]:
```

Treat _init_lmcache_engine as the internal method, just pass through two composited configs, then we can easy to use new coming sub configs of vllm_config within this method without modify the signature.

maobaolong:
@sammshen @chunxiaozheng Would you like to take another look at this PR ? Thanks!

lmcache/integration/vllm/vllm_v1_adapter.py
Shaoting-Feng: I agree :)

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -412,6 +417,11 @@ def init_lmcache_engine(
     if LMCacheEngineBuilder.get(ENGINE_NAME) is not None:
         return None
 
+    model_config = (vllm_config.model_config,)
```
sammshen: sorry, I don't quite understand why are we turning these unto tuples?


for exaple, integration tests are failing right now because of:
```text

(EngineCore_0 pid=143)     kv_dtype = get_kv_cache_torch_dtype(cache_config.cache_dtype, model_config.dtype)
--
  | (EngineCore_0 pid=143)                                         ^^^^^^^^^^^^^^^^^^^^^^^^
  | (EngineCore_0 pid=143) AttributeError: 'tuple' object has no attribute 'cache_dtype'
  | [rank0]:[W818 03:09:15.129769420 ProcessGroupNCCL.cpp:1479] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
```

maobaolong: Sorry for confusing you, this is my fault . @sammshen

maobaolong:
@sammshen Thanks for point the code issue, fixed it just now, PTAL.

sammshen:
LGTM！

sammshen (approved):

maobaolong:
@Shaoting-Feng @sammshen Thanks for your review!

PR #1332, opened 2025-08-13
[bugfix] adapt old vllm version

chunxiaozheng:
set lookup requests when use old vllm version

our vllm version is old, the lookup requests will not set without this PR, which will result in memory leak.

our grafana can clearly show that pinned memory objects not released, and will rusult in allocate failed in `local_cpu_backend`.
image

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1131,6 +1131,10 @@ class LMCacheConnectorV1Impl:
 
         meta = LMCacheConnectorMetadata()
 
+        # set and update lookup requests for unpin
+        meta.lookup_requests_in_step = self._lookup_requests_in_step
+        self._lookup_requests_in_step = []
+
         for finished_req_id in scheduler_output.finished_req_ids:
             self._request_trackers.pop(finished_req_id, None)
             self._unfinished_requests.pop(finished_req_id, None)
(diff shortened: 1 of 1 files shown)


YaoJiayi:
@chunxiaozheng Why the old vllm doesn't set the lookup requests?

chunxiaozheng:
> @chunxiaozheng Why the old vllm doesn't set the lookup requests?

@YaoJiayi 
The old version will go to this `if` branch, only add requests here. 
```
        if isinstance(cached_reqs, list):
            for i, req in enumerate(cached_reqs):
                request_tracker = self._request_trackers[req.req_id]
                request_tracker.update(req.new_token_ids, req.new_block_ids)

                    self,
                    request_tracker,
                    self._block_size,
                    self._lmcache_chunk_size,
                    load_spec=None,
                    skip_save=force_skip_save,
                    discard_partial_chunks=self._discard_partial_chunks,
                )
                if req_meta is not None:
                    meta.add_request(req_meta)
            return meta
```

The new version set lookup requests by the follow code.
```
    meta.lookup_requests_in_step = self._lookup_requests_in_step
    self._lookup_requests_in_step = []
```

This code should be mentioned above, both the new and old versions can set lookup requests

sammshen (approved):
LGTM

maobaolong (approved):
@chunxiaozheng Thanks for this fix, it can adapt the vllm in our company which is < 0.9.2

PR #1340, opened 2025-08-14
Add `caching` to enable_cache for one request

xshwu-ai:
When we want to enable cache by request, we can set `kv_transfer_params.caching` field to control whether to cache this request.

How to use:
add caching field to `kv_transfer_params`
```
    -d '{
    "model": "deepseek-r1",
    "messages": [
        {
            "role": "user",
            "content": "Hello, how are you?"
        }
    ],
    "max_tokens": 1000,
    "temperature": 0,
    "top_p": 0.95,
    "stream": true,
    "stream_options": {
         "include_usage": true
    },
    "kv_transfer_params":{
            "caching": false
        }
    }'
```

### DIFF
--- a/examples/caching_switch/README.md
+++ b/examples/caching_switch/README.md
@@ -1,0 +1,55 @@
+# Cache by request
+This is an example to cache by request, use the `kv_transfer_params.caching` field to control whether to cache this request.
+
+## Prerequisites
+Your server should have at least 1 GPU.
+
+This will use the port 8000 for 1 vllm.
+
+## Steps
+1. Start the vllm engine at port 8000:
+
+```bash
+VLLM_USE_V1=1 \
+LMCACHE_USE_EXPERIMENTAL=True \
+LMCACHE_TRACK_USAGE=false \
+LMCACHE_CONFIG_FILE=example.yaml \
+vllm serve /disc/f/models/opt-125m/ \
+           --served-model-name "facebook/opt-125m" \
+           --enforce-eager  \
+           --port 8000 \
+           --gpu-memory-utilization 0.8 \
+           --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}' \
+           --trust-remote-code
+```
+
+3. Send a request to vllm engine with `kv_transfer_params: {caching: false}`:
+```bash
+curl -X POST http://localhost:8000/v1/completions \
+  -H "Content-Type: application/json" \
+  -d '{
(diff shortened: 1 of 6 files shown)


maobaolong:
@xshwu-ai Would you like to add a docs also?

xshwu-ai:
> @xshwu-ai Would you like to add a docs also?

done

examples/caching_switch/README.md
```diff
@@ -0,0 +1,55 @@
+# Cache by request
+This is an example to cache by request, use the `kv_transfer_params.caching` field to control whether to cache this request.
```
Shaoting-Feng, line 2: We shall mention that this field controls whether to save to LMCache but cannot control the prefix caching in vLLM.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -892,6 +906,9 @@ def wait_for_save(self):
             ) and self.kv_role != "kv_producer":
                 continue
 
+            if not request.enable_cache:
+                continue
```
Shaoting-Feng, line 910: This implementation has conflict with PD disaggregation. If we `continue` here, the prefiller then will not send the KV cache to the decoder.

xshwu-ai: PD disaggregation will use `MultiConnector`, `continue` does not affect PD disaggregation.
Clipboard_Screenshot_1755243281

xshwu-ai: After careful consideration, I realize this does affect the implementation of PD disaggregation within LMCache. I'll reconsider how to approach the caching implementation. Thank you for your valuable input.

Shaoting-Feng: Sounds good to me :)

sammshen:
We can just ignore enable/disable caching flag in PD case?

sammshen:
can you integrate the `enable_cache` into the `skip_save` parameter? 

We already have this check by @vladnosiv from before: 
```python
        # NOTE(vladnosiv): for disagg, you cannot skip saving, as saving is a transfer
        skip_save = tracker.disagg_spec is None and (
            skip_save
            or (tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary)
        )
```


Also how does this differ from `examples/cache_interface`? That seems to be store_cache as well? will this one make that one redundant? 

THanks for the PR!

PR #1344, opened 2025-08-14
Add metrics to LocalCPUBackend by lambda

maobaolong:
A simple way to add metrics

image

### DIFF
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -780,6 +780,21 @@ class PrometheusLogger:
             labelnames=labelnames,
             multiprocess_mode="livemostrecent",
         )
+        # Dynamically get value while capture
+        self.local_cpu_hot_cache_count = self._gauge_cls(
+            name="lmcache:remote_ping_error_code",
+            documentation="The size of the hot_cache",
+            labelnames=labelnames,
+            _labelvalues=self.labels,
+            multiprocess_mode="livemostrecent",
+        )
+        self.local_cpu_keys_in_request_count = self._gauge_cls(
+            name="lmcache:local_cpu_keys_in_request_count",
+            documentation="The size of the keys_in_request",
+            labelnames=labelnames,
+            _labelvalues=self.labels,
+            multiprocess_mode="livemostrecent",
+        )
 
     def _log_gauge(self, gauge, data: Union[int, float]) -> None:
         # Convenience function for logging to gauge.
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
 from dataclasses import dataclass
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 import os
 import threading
 import time
@@ -780,21 +780,24 @@ class PrometheusLogger:
             labelnames=labelnames,
             multiprocess_mode="livemostrecent",
         )
-        # Dynamically get value while capture
+        self._dynamic_metrics(labelnames)
+
+    def _dynamic_metrics(self, labelnames):
+        """
+        Dynamically get value by lambda function while capture
+        """
         self.local_cpu_hot_cache_count = self._gauge_cls(
-            name="lmcache:remote_ping_error_code",
+            name="lmcache:local_cpu_hot_cache_count",
             documentation="The size of the hot_cache",
             labelnames=labelnames,
-            _labelvalues=self.labels,
             multiprocess_mode="livemostrecent",
-        )
+        ).labels(**self.labels)
         self.local_cpu_keys_in_request_count = self._gauge_cls(
             name="lmcache:local_cpu_keys_in_request_count",
             documentation="The size of the keys_in_request",
             labelnames=labelnames,
-            _labelvalues=self.labels,
             multiprocess_mode="livemostrecent",
-        )
+        ).labels(**self.labels)
 
     def _log_gauge(self, gauge, data: Union[int, float]) -> None:
         # Convenience function for logging to gauge.
@@ -933,6 +936,14 @@ class PrometheusLogger:
         assert PrometheusLogger._instance is not None, (
             "PrometheusLogger instance not created yet"
         )
+        return PrometheusLogger._instance
+
+    @staticmethod
+    def GetInstanceIfCreate() -> Optional["PrometheusLogger"]:
+        """
+        Returns the singleton instance of PrometheusLogger if it exists,
+        otherwise returns None.
+        """
         return PrometheusLogger._instance
 
 
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -939,7 +939,7 @@ class PrometheusLogger:
         return PrometheusLogger._instance
 
     @staticmethod
-    def GetInstanceIfCreate() -> Optional["PrometheusLogger"]:
+    def GetInstanceOrNone() -> Optional["PrometheusLogger"]:
         """
         Returns the singleton instance of PrometheusLogger if it exists,
         otherwise returns None.
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -7,6 +7,7 @@ import threading
 import time
 
 # Third Party
+from prometheus_client import REGISTRY
 import prometheus_client
 
 # First Party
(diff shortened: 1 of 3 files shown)


Shaoting-Feng (commented):
This is convenient, but I’m concerned that current metrics are set up in `observability.py`, so having two scripts perform the same function may be hard to maintain.

maobaolong:
@Shaoting-Feng Thanks for your reply. 

I noticed there are two kind of metrics.
1. Harmful for performance while computing value, something like compute a count of a composite collection without a inner counter, for this kind of metrics, we'd better maintain a stat variable and sink to metric system periodically.
2. There is a collection variable which supply an O(1)  len(xxx) way to compute value or a number variable already. For this kind of metrics, if we use the previous metrics framework(`observability.py`), there are following downside.
   - It make things going complex, we have to write more code to `LMCStatsMonitor` and convert to `LMCacheStats` periodically, then `log_prometheus` by a thread, it is so complex.
   - I found that we monitor some value but defined a sister metrics and try to keep the two consistent by `+` or `-` same time, but they are still two variable, there is a chance to be different for this two.

So, I still would like to suggest keep the two way both.
1. Previous way maintain the counter or some metrics Inefficient to compute while capture.
2. New way(lambda) maintain the gauge efficient to compute, especially for number variable or O(1) len(x) collection variable.

Any thoughts?

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -67,6 +69,21 @@ def __init__(
         # assumption: only one request is looked up at a time
         # (only one worker per cache engine)
         self.keys_in_request: List[CacheEngineKey] = []
+        self._setup_metrics()
+
+    def _setup_metrics(self):
+        Gauge(
+            "lmcache:local_cpu_hot_cache_count",
+            "The size of the hot_cache",
+            ["uuid"],
```
sammshen: this is to distinguish between workers right?

tests/v1/storage_backend/test_local_cpu_backend.py
```diff
@@ -87,6 +88,10 @@ def local_cpu_backend_disabled(memory_allocator):
 class TestLocalCPUBackend:
     """Test cases for LocalCPUBackend."""
 
+    def teardown_method(self, method):
```
sammshen, line 91: why this is unused?

lmcache/observability.py
```diff
@@ -396,6 +397,15 @@ def GetOrCreate() -> "LMCStatsMonitor":
     def DestroyInstance():
         LMCStatsMonitor._instance = None
 
+    @staticmethod
+    def unregister_all_metrics():
```
sammshen, line 425: sorry where is this supposed to be called?

sammshen:
Makes sense to me. this will reduce the overhead of quick query stats and we can sprinkle in more deubgging metrics in the future

sammshen:
@ApostaC not sure what you think about having the prometheus logic for O(1) query stats spill outside of `observability.py`? LGTM though!

sammshen (approved):

tests/v1/storage_backend/test_local_cpu_backend.py
maobaolong: @sammshen This is a reserved method name `teardown_method ` which invoked after each test ended.

lmcache/observability.py
maobaolong: it called by teardown_method of each test @sammshen

sammshen: I see it now thx!

tests/v1/storage_backend/test_local_cpu_backend.py
sammshen: I see, thanks I did not know this about pytest!

lmcache/v1/storage_backend/local_cpu_backend.py
maobaolong: No, this avoid start more than one instance of LocalCpuBackend within one Process, it don't happen in the production code, but it can happen for pytest, without this trick, pytest will failed, the root cause is that prometheus Registry cannot permit register an existing gauge again.

maobaolong:
@sammshen Thanks for your review! @ApostaC @Shaoting-Feng Would you like to take another look?

lmcache/v1/storage_backend/local_cpu_backend.py
sammshen: I see, that is a tricky bug nice fix! I think the extra benefit is that vllm/lmcache workers on different rank will get diff uuid and will be able to see separate them in prom stats?

sammshen: https://prometheus.github.io/client_python/multiprocess/

probably need to set `PROMETHEUS_MULTIPROC_DIR` by default for the cache_engine_internal_api_server in the future right?

Shaoting-Feng:
LGTM. cc @ApostaC for the design choice.

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -67,6 +69,21 @@ def __init__(
         # assumption: only one request is looked up at a time
         # (only one worker per cache engine)
         self.keys_in_request: List[CacheEngineKey] = []
+        self._setup_metrics()
+
+    def _setup_metrics(self):
+        Gauge(
```
ApostaC: Sorry, I'm not so familiar with the Prometheus python SDK. If we create gauges here, can they be automatically exposed to the prometheus metric endpoint?

ApostaC: Also, is it possible to add a helper function in `PrometheusLogger` for other modules to register metrics?

The goal of such a helper function is to make sure that the metric labels (right now include `model_id` and `worker_id`) can be correctly added. (See the function `PrometheusLogger._metadata_to_labels`)

ApostaC (commented):
@maobaolong Thanks, this is very helpful! Just left 1 comment and 1 quick question.

lmcache/v1/storage_backend/local_cpu_backend.py
maobaolong: Yes, it is verified by me, further more, I get into the constructor of the `Gauge`, and its parent class `MetricWrapperBase`, I found `REGISTRY` the global registry of prometheus metrics, register this new constructed `Gauge` while initing, so it can be exposed automically.

```python
class Gauge(Metric):
    def __init__(self,
                 name: str,
                 documentation: str,
                 labelnames: Iterable[str] = (),
                 namespace: str = '',
                 subsystem: str = '',
                 unit: str = '',
                 registry: Optional[CollectorRegistry] = REGISTRY,
                 _labelvalues: Optional[Sequence[str]] = None,
                 multiprocess_mode: Literal['all', 'liveall', 'min', 'livemin', 'max', 'livemax', 'sum', 'livesum', 'mostrecent', 'livemostrecent'] = 'all',
                 ):
        ...
        super().__init__(
            name=name,
            documentation=documentation,
            labelnames=labelnames,
            namespace=namespace,
            subsystem=subsystem,
            unit=unit,
            registry=registry,
            _labelvalues=_labelvalues,
        )
        ...


class MetricWrapperBase(Collector):
    def __init__(self: T,
                 name: str,
                 documentation: str,
                 labelnames: Iterable[str] = (),
                 namespace: str = '',
                 subsystem: str = '',
                 unit: str = '',
                 registry: Optional[CollectorRegistry] = REGISTRY,
                 _labelvalues: Optional[Sequence[str]] = None,
                 ) -> None:
        ....
        if not self._labelvalues:
            # Register the multi-wrapper parent metric, or if a label-less metric, the whole shebang.
            if registry:
                registry.register(self)
```

maobaolong: Thanks for this tip

maobaolong:
@ApostaC Would you like to take another? Thanks 

I tested again and checked the new added metrics with the `model_name` and `worker_id` label.

image

lmcache/observability.py
```diff
@@ -910,6 +938,14 @@ def GetInstance() -> "PrometheusLogger":
         )
         return PrometheusLogger._instance
 
+    @staticmethod
+    def GetInstanceIfCreate() -> Optional["PrometheusLogger"]:
```
ApostaC: `GetInstanceIfCreate` sounds a little bit weird (I'm okay with it though)

Proposing some alternatives:
- `GetInstanceIfExists`
- `GetInstanceOrNone`
- `MaybeGetInstance`

ApostaC (approved):
Just one minor comment about the function name. Otherwise LGTM!

ApostaC:
Tagging the `full` label to trigger the comprehensive test.

lmcache/observability.py
maobaolong: GetInstanceOrNone better

maobaolong:
@ApostaC Thanks for your review!

Issue #1346, opened 2025-08-14
LMCache seems to be using the wrong CUDA devices with Ray+PP


sdimitro:
## Setup Context

We got 2 nodes. Each node has 8 GPUs. Both nodes and all 16 GPUs are shown as resources under `ray status`.

## Reproducer & Error Message

Running the following `vllm` command:
```
VLLM_WORKER_MULTIPROC_METHOD=spawn  LMCACHE_USE_EXPERIMENTAL=True  vllm serve meta-llama/Llama-3.1-70B-Instruct  --gpu-memory-utilization 0.7  --tensor_parallel_size 8  --pipeline_parallel_size 2  --no-enable-chunked-prefill  --no-enable-prefix-caching --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both","kv_connector_extra_config": {}}' --distributed-executor-backend ray
```

causes the following error in 8 of the nodes:
```
RayWorkerWrapper pid=3182670, ip=10.192.207.175) ERROR 08-14 17:21:25 [worker_base.py:619]   File "/home/amg/vllm/.venv/lib/python3.12/site-packages/vllm/v1/worker/gpu_worker.py", line 181, in init_device                                             
(RayWorkerWrapper pid=3182670, ip=10.192.207.175) ERROR 08-14 17:21:25 [worker_base.py:619]     init_worker_distributed_environment(self.vllm_config, self.rank,                                                                                          
(RayWorkerWrapper pid=3182670, ip=10.192.207.175) ERROR 08-14 17:21:25 [worker_base.py:619]   File "/home/amg/vllm/.venv/lib/python3.12/site-packages/vllm/v1/worker/gpu_worker.py", line 584, in init_worker_distributed_environment                     
(RayWorkerWrapper pid=3182670, ip=10.192.207.175) ERROR 08-14 17:21:25 [worker_base.py:619]     ensure_kv_transfer_initialized(vllm_config)                                                                                                               
(RayWorkerWrapper pid=3182670, ip=10.192.207.175) ERROR 08-14 17:21:25 [worker_base.py:619]   File "/home/amg/vllm/.venv/lib/python3.12/site-packages/vllm/distributed/kv_transfer/kv_transfer_state.py", line 64, in ensure_kv_transfer_initialized      
[...]

PR #1356, opened 2025-08-15
Workaround for issue #1346

sdimitro:
The issue this PR is trying to solve is documented here -> https://github.com/LMCache/LMCache/issues/1346

This PR is by no means a suggestion to what the code should look like. It's goals are the following:
* Provide a starting point for discussion to design the final solution
* Provide a workaround for people that need it until we arrive to that solution

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -447,8 +447,10 @@ def init_lmcache_engine(
     logger.info(f"use mla: {use_mla}, kv shape: {kv_shape}")
 
     # Change current device.
-    torch.cuda.device(parallel_config.rank)
-    device = torch.device(f"cuda:{parallel_config.rank}")
+    num_gpus = torch.cuda.device_count()
+    local_rank = parallel_config.rank % num_gpus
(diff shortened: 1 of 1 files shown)


sdimitro:
cc @ApostaC @maobaolong

ApostaC (approved):
Looks like a clean fix, thanks!. Have you tested it @sdimitro ?

@maobaolong Please take a look also.

ApostaC:
Hey @sdimitro , would you also like to create a doc page for Ray + PP + LMCache? That would be very useful.

sdimitro:
@ApostaC replies in-line:

> Have you tested it @sdimitro ?

I've tested it only once and then I hit this issue https://github.com/LMCache/LMCache/issues/1357 .  I'd like to properly test it first in an internal system multiple times and hopefully fix the issue 1357 to really test inference before merging this. This may take some time (maybe late next week?) as most of our internal systems are being used for PoCs now :/

> Hey @sdimitro , would you also like to create a doc page for Ray + PP + LMCache? That would be very useful.

As long as I'm able to make it work I'd be happy to!

YaoJiayi (approved):
lgtm

PR #1363, opened 2025-08-16
[CD]: Dynamic Torch Versioning

sammshen:
1. make installation from source torch version agnostic and force users / serving engine dependencies to specify instead. 
2. fix the wheel installation instructions
3. make the docker image always compatible with lmcache and vllm

All of this was exposed when pytorch upgraded to 2.8.0, which changed the ABI compatibiltiy with 2.7.1. 

This is the proposed solution after https://github.com/LMCache/LMCache/pull/1274

### DIFF
--- a/docs/source/getting_started/installation.rst
+++ b/docs/source/getting_started/installation.rst
@@ -52,18 +52,41 @@ Confirm that you have the latest pre-release:
 Install Latest LMCache from Source
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To install from source, clone the repository and install in editable mode:
+To install from source, clone the repository and install in editable mode. 
+
+The reason that torch installation is separated is:
+- different serving engines and different versions of those serving engines 
+have different torch dependencies and we want to maintain compatibility. 
+- no build isolation bypasses `PEP 517 <https://peps.python.org/pep-0517/>`_ / `PEP 518 <https://peps.python.org/pep-0518/>`_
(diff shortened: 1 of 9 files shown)


sammshen:
@YaoJiayi unit tests work without the mock!

sammshen:
@hickeyma @maobaolong what are your thoughts on this current torch flexible version? to adjust for many serving engines (in the future) and serving engine versions, install from source is the solution. the release docker file will be built with the latest stable vllm torch version and teh nightly from nightly vllm

docs/source/getting_started/installation.rst
```diff
@@ -72,7 +112,23 @@ We recommend developers to use `uv` for a better package management:
...
+
+    # Option 1. 
+    # select the torch version that matches the dependency of your serving engine
+    # 2.7.1 is an example for vllm 0.10.0
+    uv pip install torch==2.7.1
+
+    # Option 2. 
+    # install your serving engine with its required torch version declared already
+    # example: vllm 0.10.0 will install torch 2.7.1
+    pip install vllm==0.10.0
```
Shaoting-Feng: Missing `uv` here?

Shaoting-Feng (approved):
LGTM. This shall fix the failing integration test.

docs/source/getting_started/installation.rst
sammshen: nice catch!! thanks!

YaoJiayi (approved):
lgtm

hickeyma:
Nice work @sammshen, thanks for finding a better solution.

Issue #1367, opened 2025-08-17
[new feature]Have an option to store only high priority requests


FerdinandZhong:
**Description**
For the inference server serving numerous downstream applications and users, some of the requests are not "necessary" to be cached. For example the requests from a batch engine (sent in jsonl file) don't typically have the following requests and can be bypassed for storing the kv cache for them.

**Describe the solution you'd like**
vLLM itself has the priority argument for every request, this feature can be utilized for the optional kv cache storing.

PR #1368, opened 2025-08-17
Priority based storing -- only store kv cache for high priority requests

FerdinandZhong:
This PR provides a simple solution for adding a new feature -- enabling the optional kv cache storage based on the priority of requests (vLLM has this "priority" argument for every request).

Through the configuration, users can define a priority_limit (LMCACHE_PRIORITY_LIMIT for env var) as the boundary for storing the kv cache. Only requests with priority value <= the limit can be stored (for vLLM, lower the priority value, higher the priority). However, the cache can still be retrieved for the low-priority request if there is any (the low-priority requests repeat the high-priority request).

If the configuration -- "priority_limit" remains unset (default is None) then there's no checking of the request's priority and all remain same as before.

[Unit Test]
The changes mainly implemented in the "vLLM_v1_adapter", and due to no unit test has been written for the vLLM_integration, therefore I didn't add the new unit test.

[Manual Testing]
* The feature has been manually tested with the different situations. Below is the screenshoot of vLLM logs when the priority_limit is set as 0, and the requests randomly have the priority as 0 or 10. When the priority is 10, the kv cache storing is bypassed
Screenshot 2025-08-17 at 12 28 08 AM

* When the server was restarted with the priority_limit set as 10, all requests sent from the same client had their kv cache stored (no request had the priority value > 10)
Screenshot 2025-08-17 at 12 30 25 AM

Issue: [1367](https://github.com/LMCache/LMCache/issues/1367) (*link existing issues this PR will resolve*)

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>


sammshen:
LGTM!

sammshen (approved):

maobaolong:
@Shaoting-Feng Reference to https://github.com/LMCache/LMCache/pull/1340#discussion_r2277715749

there is the same issue within this PR.

FerdinandZhong:
> @Shaoting-Feng Reference to [#1340 (comment)](https://github.com/LMCache/LMCache/pull/1340#discussion_r2277715749)
> 
> there is the same issue within this PR.

Hi @maobaolong could you explain the issue a bit, thanks?

maobaolong:
> > @Shaoting-Feng Reference to [#1340 (comment)](https://github.com/LMCache/LMCache/pull/1340#discussion_r2277715749)
> > there is the same issue within this PR.
> 
> Hi @maobaolong could you explain the issue a bit, thanks?

This implementation has conflict with PD disaggregation. If we skip save here, the prefiller then will not send the KV cache to the decoder ?

FerdinandZhong:
> > > @Shaoting-Feng Reference to [#1340 (comment)](https://github.com/LMCache/LMCache/pull/1340#discussion_r2277715749)
> > > there is the same issue within this PR.
> > 
> > 
> > Hi @maobaolong could you explain the issue a bit, thanks?
> 
> This implementation has conflict with PD disaggregation. If we skip save here, the prefiller then will not send the KV cache to the decoder ?

Understood thanks. Let me think about how to disable skip for PD disaggregation.

sammshen:
can ignore priority filtering for PD and add comment + documentation.

sammshen:
Actually @vladnosiv already put in a checker in `from_request_tracker`:
```python
        # NOTE(vladnosiv): for disagg, you cannot skip saving, as saving is a transfer
        skip_save = tracker.disagg_spec is None and (
            skip_save
            or (tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary)
        )
 ```

**Comment by FerdinandZhong** [comment]
> Actually @vladnosiv already put in a checker in `from_request_tracker`:
> 
> ```python
>         # NOTE(vladnosiv): for disagg, you cannot skip saving, as saving is a transfer
>         skip_save = tracker.disagg_spec is None and (
>             skip_save
>             or (tracker.num_saved_tokens > 0 and input_token_len < chunk_boundary)
>             or (tracker.is_decode_phase and not save_decode_cache)
>         )
> ```

I see, thanks @sammshen 
So seems I don't need to update my implementation as the skip_save will be checked again in `from_request_tracker`. I will mention this is not suitable for PD disaggregation id documentation

**Comment by FerdinandZhong** [comment]
Hi @sammshen @maobaolong I further updated the documentation to mention the "priority_limit" is not applicable for PD disaggregation. Thanks

**Review by Shaoting-Feng — COMMENTED** [review]
Can we add a field to `RequestTracker` instead of maintaining a separate list `_skip_save_reqs`? Since you get the priority from vLLM’s request, maybe we could move the priority comparison inside `from_new_request`.

**Comment by Shaoting-Feng** [comment]
In this way, we may also get to use `save_spec` in `ReqMeta`.

**Comment by FerdinandZhong** [comment]
Hi @Shaoting-Feng thank you for the comment.
In my latest commit, I removed `_skip_save_reqs` and add the new field `skip_save` to the `RequestTracker` as what you mentioned in your comment. The `ReqMeta` is then based on this field of `request_tracker` to create the SaveSpec object.

However, I still need to keep a `requests_priority` list as the `from_new_request` function is having vLLM's `NewRequestData` which doesn't have the priority field. But the vLLM's `Request` class has this `priority` field. So I use the `requests_priority` list to keep the `priority` of the request only and whether `skip_save` or not for this request is determined inside the from_new_request` function.

The `priority` information of each request might be possible to utilized for other features in the future as well.

**Comment by FerdinandZhong** [comment]
Hi @Shaoting-Feng @maobaolong could you review the new commits? Thanks

**Comment by sammshen** [comment]
@ApostaC @Shaoting-Feng running comprehensive test on this to be safe

File: lmcache/integration/vllm/vllm_v1_adapter.py [hunk 1]
```diff
@@ -1008,6 +1022,8 @@ def get_num_new_matched_tokens(
         ):
             return 0
 
+        self._requests_priority[request.request_id] = request.priority
```
ApostaC, line 1047: QQ: is `request.priority` already stable enough in vLLM?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -135,13 +135,18 @@ class RequestTracker:
...
+
     @_lmcache_nvtx_annotate
     @staticmethod
     def from_new_request(
         lmcache_config: LMCacheEngineConfig,
         new_request: "NewRequestData",
         num_tokens_to_compute: int,
         lmcache_cached_tokens: int,
+        request_priority: int,
+        force_skip_save: bool,
```
ApostaC: Why do we need to put the skip_save flag in the request tracker? This class is meant to save the request states, such as token ids/block ids/number of cached tokens.

An alternative way is to check the priority in `build_connector_meta` and directly calculate and pass the `skip_save` flag in the `build_connector_meta` function.

ApostaC (commented):
Otherwise LGTM

lmcache/integration/vllm/vllm_v1_adapter.py
FerdinandZhong: Hi @ApostaC yes I think so, this feature has been there since v0.7 and it's been enabled in v1 engine since v0.9

lmcache/integration/vllm/vllm_v1_adapter.py
FerdinandZhong: @ApostaC  Thanks for the feedback. 

I put the `skip_save` into the request_tracker as feedbacked by @Shaoting-Feng in the previous comment. However, I think it can make sense as whether the request should `skip_save` can be tracked and re-visit when looping over `scheduler's scheduled_cached_reqs`. And I put the calculation of the `skip_save` for new request into the `build_connector_meta` now.

Please let me know your thoughts thanks

ApostaC (approved):
LGTM!

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1036,6 +1044,8 @@ def get_num_new_matched_tokens(
         ):
             return 0
 
+        self._requests_priority[request.request_id] = request.priority
```
maobaolong, line 1047: The `request.priority` is not exists in vllm `0.8.5`, so this PR make LMCache break backward-compatible.

@ApostaC @FerdinandZhong @yoo-kumaneko

FerdinandZhong: Ok I will make a new PR to fix this

maobaolong: @FerdinandZhong This has been resolved by https://github.com/LMCache/LMCache/pull/1678 
Thanks anyway!

FerdinandZhong: Got it, thank you!

maobaolong: You are welcome

PR #1377, opened 2025-08-18
[refact] use request_configs replace tags

chunxiaozheng:
# DESCRIPTIOIN
This PR refact https://github.com/LMCache/LMCache/pull/1200, use `request_configs` to replace tags. By `request_configs`, we can set tags and the other configs.

# WHY DO THIS
In our scenario, we not only need to set tags to separate different requests, but also need to support setting different configs for different requests, such as setting ttl when puting the key, etc. This PR use `request_configs` to replace `tags`, if the key starts with `lmcache.tag.`, such as `lmcache.tag.tag1`, the `tag1` will be treated as the tag, if the key starts with `lmcache.`, such as `lmcache.config1`, the `configs1` will be treated as the config.

Besides, for the convenience of review, most of the changes is renaming `tags` to `request_configs`. The change is that add a `__post_init__` method in `CacheEngineKey` to extract `tags` from the `request_configs`.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -88,16 +88,16 @@ class DisaggSpec:
 tmp_disagg_tracker: dict[str, DisaggSpec] = {}
 
 
-def extract_request_extra_configs(sampling_params: SamplingParams) -> Optional[dict]:
-    extra_configs = None
+def extract_request_configs(sampling_params: SamplingParams) -> Optional[dict]:
+    request_configs = None
     if sampling_params.extra_args is not None:
         if kv_transfer_params := sampling_params.extra_args.get("kv_transfer_params"):
             for k, v in kv_transfer_params.items():
                 if k.startswith("lmcache."):
-                    if extra_configs is None:
-                        extra_configs = {}
-                    extra_configs[k] = v
-    return extra_configs
+                    if request_configs is None:
+                        request_configs = {}
+                    request_configs[k] = v
+    return request_configs
 
 
 @dataclass
@@ -127,8 +127,8 @@ class RequestTracker:
     mm_hashes: Optional[list[str]] = None
     mm_positions: Optional[list["PlaceholderRange"]] = None
 
-    # The extra configs of the request, includes tags and other configs
-    extra_configs: Optional[dict] = None
+    # The configs of the request, includes tags and other configs
+    request_configs: Optional[dict] = None
 
     # Whether the request is in decode phase
     is_decode_phase = False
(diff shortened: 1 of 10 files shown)


maobaolong:
@chunxiaozheng Would  you like to add a meaningful description? So reviewers can quick getting in this PR.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -91,21 +91,16 @@ class DisaggSpec:
...
+    if sampling_params.extra_args is not None:
         if kv_transfer_params := sampling_params.extra_args.get("kv_transfer_params"):
-            for k in tag_keys:
-                if v := kv_transfer_params.get(k):
-                    if tags is None:
-                        tags = OrderedDict()
-                    tags[k] = v
-    return tags
+            for k, v in kv_transfer_params.items():
+                if k.startswith("lmcache."):
```
maobaolong, line 96: This is better than set a  `tag_keys` into confiig.👍

chunxiaozheng:
> @chunxiaozheng Would you like to add a meaningful description? So reviewers can quick getting in this PR.

@maobaolong Sure, I will update the description when i complete the job.

chunxiaozheng:
hi @sammshen, could you help take a look?

maobaolong:
@chunxiaozheng Conflict need resolved.

chunxiaozheng:
> @chunxiaozheng Conflict need resolved.

@maobaolong thanks! the conflicts have fixed.

lmcache/integration/vllm/vllm_v1_adapter.py
maobaolong: Key changes 1

lmcache/utils.py
```diff
@@ -79,7 +79,17 @@ class CacheEngineKey:
     world_size: int
     worker_id: int
     chunk_hash: int
-    tags: Optional[OrderedDict] = None
+    request_configs: Optional[dict] = None
+
+    def __post_init__(self):
+        tags = None
```
maobaolong, line 85: key change 2

maobaolong (approved):
@chunxiaozheng I found there are only 2 key changes,  in addition, others are all rename and misc.

The 2 key changes LGTM, this is a nice abstraction.👍

sammshen:
LGTM, TTL will be next PR right?

sammshen (approved):

chunxiaozheng:
> LGTM, TTL will be next PR right?

@sammshen thanks for your review! TTL in our another PR.

maobaolong:
@chunxiaozheng thanking to https://github.com/LMCache/LMCache/pull/1400 is fixing the comprehensive test, we just ignore this ci failed.

Shaoting-Feng:
> @chunxiaozheng thanking to #1400 is fixing the comprehensive test, we just ignore this ci failed.

The comprehensive test failed because the server running it was down, and the new server still requires some setup. Everything should be fine now.

PR #1387, opened 2025-08-19
Improve internal api server to support multiple lmcache server

maobaolong:
# Description

This PR is a big Improvement base on internal api server, the major changes of this PR is
1.🔥Refactor to support introduce new api without any existing code change.
2.🔥Support flexible config
  - Support config whether enable or disable internal api server
  - Support config include list of which server should enabled internal api server.(0, instead of scheduler, 1 means worker0, 2 is worker1, ...)

# How to use

1. Config internal api server

```
internal_api_server_enabled: True
internal_api_server_port_start: 8080
internal_api_server_include_index_list: [1,2]
```

This is disable scheduler internal api server, enable worker 0 and worker 1 api server. If you do not config `internal_api_server_include_index_list`, it means start internal api server for all lmcache servers, and the port starts with 8080.


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -610,6 +612,9 @@ def __init__(
 
         self.force_skip_save = bool(os.environ.get("LMCACHE_FORCE_SKIP_SAVE", False))
 
+        self.api_server = InternalAPIServer(self)
+        self.api_server.start()
```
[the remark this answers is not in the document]
maobaolong: It is hard to add close method into vllm. Let's ignore this suggestion.

maobaolong:
@Shaoting-Feng @sammshen Would you like to take a look at this PR? Thanks

lmcache/v1/internal_api_server/loglevel_api.py
```diff
@@ -0,0 +1,54 @@
...
+
+# Third Party
+from fastapi import APIRouter
+from starlette.responses import PlainTextResponse
+
+router = APIRouter()
+
+
+@router.get("/loglevel")
+async def get_or_set_log_level(logger_name: str = None, level: str = None):
```
sammshen, line 13: nice! now we can only get more verbose logs for certain parts of the code!

lmcache/v1/internal_api_server/api_server.py
```diff
@@ -0,0 +1,75 @@
...
+                f"Internal API server disabled. internal_api_server_enabled="
+                f"{config.internal_api_server_enabled}, port_offset={port_offset}, "
+                f"port = {self.port}, include_index_list={include_index_list}"
+            )
+            self.enable = False
+            return
+
+        logger.info(f"Init internal API server on port {self.port}")
+        uvicorn_config = uvicorn.Config(
+            app, host="0.0.0.0", port=self.port, loop="uvloop", http="httptools"
```
sammshen, line 53: nice use of uvloop and httptools!

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -610,6 +612,9 @@ def __init__(
 
         self.force_skip_save = bool(os.environ.get("LMCACHE_FORCE_SKIP_SAVE", False))
 
+        self.api_server = InternalAPIServer(self)
```
sammshen, line 617: nice, now we can also debug the scheduler independently

sammshen (approved):
LGTM! Once you think this feature is stable, I think the community will be able to benefit from this with some examples or docs!

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -610,6 +612,9 @@ def __init__(
 
         self.force_skip_save = bool(os.environ.get("LMCACHE_FORCE_SKIP_SAVE", False))
```
sammshen, line 614: nit: can you write a small comment here that activation of API server for this worker / scheduler connector component is decided inside the constructor or else it looks like we always create the `api_server`

maobaolong: Done

maobaolong:
@sammshen Thanks for your continues review and the suggestion, will contribute more base on this, after stable, I'd love to write a introduction of this feature, hope this can help us to defeat more bugs and performance issue.

sammshen:
so if we create client, do we connect to all these servers? what if we consolidate all of them into one on api server on the scheduler using some RPC? will this be too complicated? just open the door for discussion

chunxiaozheng (approved):
lgtm

maobaolong:
> so if we create client, do we connect to all these servers? 
There are two follow-up tasks can make things simple.
1. Introduce a proxy_api endpoint, so we can access all internal api servers by one internal api server leverage the request forward ability.
2. Introduce a common frontend, which act as a UI/API user interface server, we can execute all operation to all nodes(scheduler and works) by it.

> what if we consolidate all of them into one on api server on the scheduler using some RPC? will this be too complicated? 

`consolidate all of them into one` is a good proposal, but I'm not sure if we let scheduler or one of works process act as a proxy may be introduce performance affect, so I create a `lmcache-frontend` process which is another process, but we still can make it be a sub-process of lmcache, so it can be startup one each node.

`using some RPC` can make things going complex i think, if so, we have to defined `message` and `encode\decode` codes when we add a new api.

BTW, discussing also going on slack.

PR #1380, opened 2025-08-18
[Core] Fix hash mismatch between LMCache and vLLM

Shaoting-Feng:
1. Add `extra_keys`
2. Correctly initialize `NONE_HASH`

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient unit tests to ensure the change is stay correct and robust. The unit and integration tests will always run and our comprehensive test will be triggered after the "full" label is tagged onto a PR.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

We aim to address all PRs in a timely manner. If no one reviews your PR within 5 days, please @-mention one of KuntaiDu, ApostaC or YaoJiayi.

PR #1403, opened 2025-08-21
[Bugfix] Fix hash value mismatch between Prefill and Decode

Hadigan:
Fix hash value mismatch between Prefill and Decode. This error was introduced by #1380 

There are two issues with the current version of the code:

1. Hashing None values yields inconsistent results across different runs, causing mismatches in hash values between the prefill and decode phases.
2. The code attempts to hash a list without first converting it to an immutable type such as a tuple, which leads to a TypeError.

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -137,7 +137,7 @@ class TokenDatabase(metaclass=abc.ABCMeta):
             raise ValueError(f"Unsupported tokens type: {type(tokens)}")
         
         # NOTE: Using the builtin hash to calculate the None object will give different results each time
-        prefix_hash = prefix_hash or prefix_hash
+        prefix_hash = prefix_hash or 0
         extra_keys_tuple = tuple(extra_keys or [])
 
         # Ignore extra keys for now
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -137,7 +137,7 @@ class TokenDatabase(metaclass=abc.ABCMeta):
             raise ValueError(f"Unsupported tokens type: {type(tokens)}")
         
         # NOTE: Using the builtin hash to calculate the None object will give different results each time
-        prefix_hash = prefix_hash or 0
+        prefix_hash = () if prefix_hash is None else prefix_hash
         extra_keys_tuple = tuple(extra_keys or [])
 
         # Ignore extra keys for now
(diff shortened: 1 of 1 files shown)


YaoJiayi:
@Hadigan Why not simply set the same python hash seed for prefill and decode nodes? We can change the vllm launchers (e.g., https://github.com/LMCache/LMCache/blob/dev/examples/disagg_prefill/1p1d_experimental/disagg_vllm_launcher.sh) with aligned python hash seed?

Hadigan:
> @Hadigan Why not simply set the same python hash seed for prefill and decode nodes? We can change the vllm launchers (e.g., https://github.com/LMCache/LMCache/blob/dev/examples/disagg_prefill/1p1d_experimental/disagg_vllm_launcher.sh) with aligned python hash seed?

Sorry for not clearly describing my test case earlier. In fact, I already set the same PYTHONHASHSEED on both the prefill and decode nodes. The root cause lies in the fact that hashing a None object in Python produces a random value—even when a fixed PYTHONHASHSEED is set. You can verify this by running:
```
PYTHONHASHSEED=0 python3 -c "print(hash(((1,2,3), None)))"
```
or
```
PYTHONHASHSEED=0 python3 -c "print(hash(None))"
```

You'll notice that the results are still non-deterministic.
In the current implementation of ```_hash_tokens```, the parameter ```extra_keys``` is not being used and thus defaults to None. This means that hash(None) is effectively being computed as part of the token hash, leading to inconsistent hash values for the same token between the prefill and decode—even with identical PYTHONHASHSEED.

Hadigan:
Hi @YaoJiayi @KuntaiDu @maobaolong, would you mind taking a look at this PR when you have a moment? Looking forward to your feedback, thanks！

panpan0000:
sounds reasonable,  and BTW, please fix per-commit lints

lmcache/v1/token_database.py
```diff
@@ -135,11 +135,15 @@ def _hash_tokens(
             tokens_tuple = tuple(tokens)
         else:
             raise ValueError(f"Unsupported tokens type: {type(tokens)}")
+        
+        # NOTE: Using the builtin hash to calculate the None object will give different results each time
```
panpan0000, line 139:
```suggestion
        """
        NOTE: Using the builtin hash to calculate the None object 
        will give different results each time
        """
```

PR #1396, opened 2025-08-20
[CI]: Add build support for Nvidia B200 GPU

hickeyma:
This also adds support for GB200 GPU.

PR #1408, opened 2025-08-21
[Docker]: Add build support for Nvidia B200 GPU to container image

hickeyma:
This is a follow on PR to #1396 to update the Dockerfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -80,8 +80,8 @@ RUN --mount=type=cache,target=/root/.cache/pip \
 ARG torch_cuda_arch_list='7.0 7.5 8.0 8.6 8.9 9.0 10.0+PTX'
 ENV TORCH_CUDA_ARCH_LIST=${torch_cuda_arch_list}
 
-# Override the vLLM arch list for flash-attn to reduce the binary size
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -132,17 +128,13 @@ WORKDIR /workspace
 ENTRYPOINT ["/opt/venv/bin/vllm", "serve"]
 
 #################### vLLM IMAGE & LMCache (Release) #######################
-# Integrate vLLM and LMCache stable releases and expose vLLM
-# OpenAI API
-
-# Will trigger from `.github/workflows/publish.yml`
```
sammshen, line 138: is it ok to keep these comments? the release right now is tricky where the next lmcache wheel will be with 2.8.0 and so will the official docker image so we have to patiently wait for vllm to upgrade to 2.8.0 as well

sammshen (approved):
LGTM thanks!

docker/Dockerfile
hickeyma: Good point but I only removed specific CI comments that should be on client side of Dockerfile and not in Dockerfile. `.github/workflows/publish.yml` is not applicable to user building the image from CLI.

YaoJiayi (approved):
lgtm

PR #1429, opened 2025-08-25
[patch]: busy loop on local cpu full

sammshen:
1. when LMCache is retrieving from disk or a remote backend, it needs to make an intermediary copy inside of the local cpu backend. it is possible that the local cpu backend is full. In this case, instead of exiting immediately, we should "busy wait" (other requests will clear up their intermediary KV caches). This only occurs with high concurrency but is an incorrect behavior to return no tokens when we guaranteed vllm tokens inside of lookup()

2. I tested this behavior with the redis backend. I was unable to achieve high concurrency until changing the redis client to async as well as implementing batched behavior. Overall, these are some optimizations to the redis backend but mainly to test the high concurrency behavior. Also, this combines redis metadata and payload under a single key for consistency

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -13,7 +13,7 @@ steps:
       uv pip install --upgrade pip setuptools wheel
       uv pip install -r requirements/common.txt
       uv pip install -r requirements/test.txt
-      uv pip install -e .
+      uv pip install -e . --no-build-isolation
       uv pip freeze
 
       source .buildkite/scripts/pick-free-gpu.sh 18000
--- a/.buildkite/pipelines/end-to-end-tests.yml
+++ b/.buildkite/pipelines/end-to-end-tests.yml
@@ -7,10 +7,10 @@ steps:
       uv venv --python 3.12 .venv-$BUILDKITE_BUILD_ID
       source .venv-$BUILDKITE_BUILD_ID/bin/activate
 
-      uv pip install --upgrade pip setuptools wheel
+      uv pip install --upgrade pip setuptools wheel 
       uv pip install -r requirements/common.txt
       uv pip install -r requirements/test.txt
-      uv pip install -e .
+      uv pip install -e . --no-build-isolation
       uv pip install matplotlib
       uv pip install pandas
       uv pip install -U vllm --pre --extra-index-url https://wheels.vllm.ai/nightly
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -254,12 +254,12 @@ class LMCacheEngine:
 
             # TODO (Jiayi): should be batched in the future
             memory_obj = self.storage_manager.allocate(
-                kv_shape, kv_dtype, is_store=True
+                kv_shape, kv_dtype, busy_loop=False
             )
             if memory_obj is None:
                 logger.warning(
-                    "Failed to allocate memory for the KV cache.\n"
-                    "The KV cache will not be stored."
+                    "Local cpu memory under pressure so"
+                    " choosing to not store the KV cache."
                 )
                 break
 
@@ -366,13 +366,13 @@ class LMCacheEngine:
                 kv_dtype,
                 batch_size=self.num_layers,
                 fmt=self.fmt,
-                is_store=True,
+                busy_loop=False,
             )
 
             if memory_objs_multi_layer is None:
                 logger.warning(
-                    "Failed to allocate memory for the KV cache.\n"
-                    "The KV cache will not be stored."
+                    "Local cpu memory under pressure so"
+                    " choosing to not store the KV cache."
                 )
                 break
 
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -253,7 +253,9 @@ class LMCacheEngine:
             kv_dtype = self.metadata.kv_dtype
 
             # TODO (Jiayi): should be batched in the future
-            memory_obj = self.storage_manager.allocate(kv_shape, kv_dtype)
+            memory_obj = self.storage_manager.allocate(
+                kv_shape, kv_dtype, is_store=True
+            )
             if memory_obj is None:
                 logger.warning(
                     "Failed to allocate memory for the KV cache.\n"
@@ -364,6 +366,7 @@ class LMCacheEngine:
                 kv_dtype,
                 batch_size=self.num_layers,
                 fmt=self.fmt,
+                is_store=True,
             )
 
             if memory_objs_multi_layer is None:
--- a/lmcache/v1/storage_backend/connector/base_connector.py
+++ b/lmcache/v1/storage_backend/connector/base_connector.py
@@ -223,5 +223,24 @@ class RemoteConnector(metaclass=abc.ABCMeta):
         """
         raise NotImplementedError
 
+    def support_batched_put(self) -> bool:
+        """
+        Check if the connector supports batched put
+        Returns:
+            True if batched put is supported, False otherwise
+        """
+        return False
+
+    async def batched_put(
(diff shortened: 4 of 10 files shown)


ApostaC:
Quick comment: we should try to decouple eviction logic and the backend itself. Let's probably find some time to discuss this.

lmcache/v1/cache_engine.py
```diff
@@ -253,11 +253,13 @@ def store(
             kv_dtype = self.metadata.kv_dtype
 
             # TODO (Jiayi): should be batched in the future
-            memory_obj = self.storage_manager.allocate(kv_shape, kv_dtype)
+            memory_obj = self.storage_manager.allocate(
+                kv_shape, kv_dtype, busy_loop=False
```
sammshen, line 257: Main change #1

avoid busy loop for store and choose not to store

lmcache/v1/storage_backend/connector/base_connector.py
sammshen, line 1: Main Change #2: I had to change redis to batched and async in order to encounter cpu memory pressure. this is not the main purpose of the PR

lmcache/v1/storage_backend/connector/redis_connector.py
sammshen, line 1: Main Change #2: I had to change redis to batched and async in order to encounter cpu memory pressure. this is not the main purpose of the PR

lmcache/v1/storage_backend/connector/base_connector.py
sammshen, line 1: Main Change #2: I had to change redis to batched and async in order to encounter cpu memory pressure. this is not the main purpose of the PR

lmcache/v1/storage_backend/connector/redis_connector.py
sammshen, line 1: Main Change #2: I had to change redis to batched and async in order to encounter cpu memory pressure. this is not the main purpose of the PR

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -258,33 +263,58 @@ def allocate(
...
+                f"Unable to allocate memory object after {num_attempts}"
+                " attempts of local cpu backend allocate()"
+            )
+            # whether or not this request needs to wait or other requests
+            wait_other_requests = True
+            if self.use_hot:
+                # TODO(Jiayi): optimize `num_candidates` with estimation.
+                # Accurate estimation is hard due to fragmentation
+                num_candidates = 1
+                with self.cpu_lock:
```
sammshen, line 274: limit scope of lock

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -161,9 +161,8 @@ def put_callback(self, future: Future, key: CacheEngineKey):
         """
         Callback function for put tasks.
         """
-        self.lock.acquire()
-        self.put_tasks.remove(key)
-        self.lock.release()
+        with self.lock:
```
sammshen, line 164: O(1) here instead of O(n)

pyproject.toml
```diff
@@ -5,12 +5,14 @@
 # to stay flexible for the torch version with `--no-build-isolation`. Thus, these packages
 # should be installed before `pip install -e . --no-build-isolation`. 
 # However, build isolation is still used for the cibuildwheel releases
+# Thus, we will still lock a torch version here because we can choose to release wheels
+# in sync with vllm and update our torch version accordingly
```
sammshen, line 9: need to release lmcache 0.3.5 becuase 0.3.4 is incompatible with current vllm after this

.buildkite/pipeline.yml
```diff
@@ -13,7 +13,7 @@ steps:
       uv pip install --upgrade pip setuptools wheel
       uv pip install -r requirements/common.txt
       uv pip install -r requirements/test.txt
-      uv pip install -e .
```
sammshen, line 16: becuase of pinning version in pyproject.toml

.buildkite/pipelines/end-to-end-tests.yml
```diff
@@ -7,10 +7,10 @@ steps:
       uv venv --python 3.12 .venv-$BUILDKITE_BUILD_ID
       source .venv-$BUILDKITE_BUILD_ID/bin/activate
 
-      uv pip install --upgrade pip setuptools wheel
+      uv pip install --upgrade pip setuptools wheel 
       uv pip install -r requirements/common.txt
       uv pip install -r requirements/test.txt
-      uv pip install -e .
```
sammshen, line 13: becuase of pinning version in pyproject.toml

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -325,51 +334,64 @@ def batched_allocate(
...
+                # do not hold the lock during sleep
+                time.sleep(time_to_wait)
 
-                if memory_objs:
-                    break
+            memory_objs = self.memory_allocator.batched_allocate(
+                shape, dtype, batch_size, fmt
+            )
+            if memory_objs:
+                break
```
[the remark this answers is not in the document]
sammshen: fixed, good catch

xleoken:
> _ No description provided. _

 hi @sammshen your patch is very good. It would be great if you could briefly describe the main modifications of this patch here.

sammshen:
@xleoken would you like to take a look

1. when LMCache is retrieving from disk or a remote backend, it needs to make an intermediary copy inside of the local cpu backend. it is possible that the local cpu backend is full. In this case, instead of exiting immediately, we should "busy wait" (other requests will clear up their intermediary KV caches). This only occurs with high concurrency but is an incorrect behavior to return no tokens when we guaranteed vllm tokens inside of `lookup()`

2. I tested this behavior with the redis backend. I was unable to achieve high concurrency until changing the redis client to async as well as implementing batched behavior. Overall, these are some optimizations to the redis backend but mainly to  test the high concurrency behavior.

xleoken:
> @xleoken would you like to take a look
> 
> 1. when LMCache is retrieving from disk or a remote backend, it needs to make an intermediary copy inside of the local cpu backend. it is possible that the local cpu backend is full. In this case, instead of exiting immediately, we should "busy wait" (other requests will clear up their intermediary KV caches). This only occurs with high concurrency but is an incorrect behavior to return no tokens when we guaranteed vllm tokens inside of `lookup()`
> 2. I tested this behavior with the redis backend. I was unable to achieve high concurrency until changing the redis client to async as well as implementing batched behavior. Overall, these are some optimizations to the redis backend but mainly to  test the high concurrency behavior.

hi @sammshen what I mean is to update here :)

image

sammshen:
thanks for the reminder, done!

sammshen:
@YaoJiayi removed the redis related refactors for now😅 PTAL again

YaoJiayi (approved):
lgtm

ApostaC:
@YaoJiayi @sammshen Probably we should trigger full tests for the PRs like this

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -103,14 +103,17 @@ def allocate(
...
         """
         Allocate memory object with memory allocator.
         Use LRU evictor if eviction is enabled.
         """
         # TODO (Jiayi): We might need to pre-allocate and management
         # disk in a similar way as CPU.
-        return self.allocator_backend.allocate(shape, dtype, fmt, eviction=eviction)
+        return self.allocator_backend.allocate(
+            shape, dtype, fmt, eviction=eviction, busy_loop=busy_loop
+        )
```
ApostaC, line 116: This line causes problems. Some allocator backends don't have `busy_loop` and will cause runtime error

sammshen: Sorry for this mistake and I will be more careful in the future. @Shaoting-Feng has fixed it in his latest PR. Looking forward to building more CI to prevent things like this

PR #1440, opened 2025-08-26
[Build]: Update Python lower and upper version support

hickeyma:
Change as follows:

- Lower: 3.9
- Upper: 3.13


sammshen (approved):
LGTM!

YaoJiayi (approved):
Thanks!

PR #1446, opened 2025-08-27
[Bug][PD] Avoid doing int16 conversion on hash

Shaoting-Feng:
This will break the PD chunk hash consistency.

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

</details>

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -171,7 +171,7 @@ class CacheEngineKey:
             parts[1],
             int(parts[2]),
             int(parts[3]),
-            int(parts[4], 16),
+            parts[4],
             request_configs,
         )
 
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -116,7 +116,7 @@ class CacheEngineKey:
     def to_string(self):
         s = (
             f"{self.fmt}@{self.model_name}@{self.world_size}"
-            f"@{self.worker_id}@{self.chunk_hash}"
+            f"@{self.worker_id}@{self.chunk_hash:x}"
         )
         if self.tags is not None and len(self.tags) != 0:
             tags = [f"{k}%{v}" for k, v in self.tags.items()]
@@ -171,7 +171,7 @@ class CacheEngineKey:
             parts[1],
             int(parts[2]),
             int(parts[3]),
-            int(parts[4]),
+            int(parts[4], 16),
             request_configs,
         )
 
@@ -244,7 +244,7 @@ class LayerCacheEngineKey(CacheEngineKey):
     def to_string(self):
         s = (
             f"{self.fmt}@{self.model_name}@{self.world_size}"
-            f"@{self.worker_id}@{self.chunk_hash}@{self.layer_id}"
+            f"@{self.worker_id}@{self.chunk_hash:x}@{self.layer_id}"
         )
         if self.tags is not None and len(self.tags) != 0:
             tags = [f"{k}%{v}" for k, v in self.tags.items()]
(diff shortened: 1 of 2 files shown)


YaoJiayi (commented):
Why didn't we have this issue before?

Shaoting-Feng:
> Why didn't we have this issue before?

Broken by https://github.com/LMCache/LMCache/pull/1377 last week.

Shaoting-Feng:
@chunxiaozheng Could you please review this change when you have time? I want to make sure it doesn’t break your feature. Thanks!

chunxiaozheng:
> @chunxiaozheng Could you please review this change when you have time? I want to make sure it doesn’t break your feature. Thanks!

@Shaoting-Feng I didn't update `int(parts[4], 16)`, I only updated the style, this code was updated by https://github.com/LMCache/LMCache/pull/1299

YaoJiayi:
@sdimitro Does the change look good to you?

sdimitro:
@Shaoting-Feng @YaoJiayi will take a look in a few hours! :)

That said, @Shaoting-Feng can you actually provide a trace and some info on how, why things break?

On my end I made that change because the weka and GDS backends actually use hexadecimal strings (see this comment on my PR: https://github.com/LMCache/LMCache/pull/1299#discussion_r2269672894). Is that not the case anymore? Aren't chunk-hashes base 10 now?

Shaoting-Feng:
> That said, @Shaoting-Feng can you actually provide a trace and some info on how, why things break?

Engine: 
- https://github.com/LMCache/LMCache/blob/dev/examples/disagg_prefill/1p1d_experimental/disagg_example_1p1d.sh

Workload:
- `bash long_input_short_output_run.sh meta-llama/Llama-3.1-8B-Instruct http://localhost:9992/ vllm_dynamo_lmcache_latest_version 1` from https://github.com/LMCache/LMBenchmark/tree/main/synthetic-multi-round-qa.

The reason it is breaking is that the chunk hash after the int16 conversion on the decoder side differs from that of vLLM.

Shaoting-Feng:
> @Shaoting-Feng I didn't update `int(parts[4], 16)`, I only updated the style, this code was updated by #1299

Got it. Thanks for pointing this out!

Shaoting-Feng:
> Aren't chunk-hashes base 10 now?

Chunk hashes are in base 10. I think another fix method is to also perform an int base-16 conversion in the `to_string` function.

sdimitro:
Apologies for the late reply, here are my thoughts

> The reason it is breaking is that the chunk hash after the int16 conversion on the decoder side differs from that of vLLM.

This seems a bit confusing to me. If the data is the same and the hash function is the same and static (e.g. doesn't have randomness), then the resulting hash should be same. Is this because we use a different hash function or something? 

> Chunk hashes are in base 10. I think another fix method is to also perform an int base-16 conversion in the to_string function.

This is also awkward because hashes are in general represented as hex in most places - e.g. when I see decimal numbers I expect counters, nanoseconds, etc.. , while for hex I expect pointer addresses, hashes, codes, etc...

I don't want to hold you guys in fixing your bug. I think if people are ok with making the `to_string()` formatting print hex, this sounds more correct IMHO. At the same time, I don't think I fully understand the underlying issue here so I'm also good with you merging what you have here and I'll try to adjust the weka-backend accordingly.

Shaoting-Feng:
> Is this because we use a different hash function or something?

The decoder receives the hash from the prefiller rather than computing it. If the prefiller’s `to_string` doesn’t match the decoder’s `from_string`, the decoder won’t get a cache hit.

Shaoting-Feng:
@sdimitro I’ve updated the `to_string` function. Please take a look when you have a moment—I hope this doesn’t affect the Weka backend.

sdimitro (approved):

sdimitro:
@Shaoting-Feng Thanks for being thorough in answering my questions and considering the weka/gds backends in this PR 🙏

YaoJiayi (approved):
lgtm

Issue #1449, opened 2025-08-27
[Doc] wrong configuration in docs of nixl storage backend


ziruiliu:
In docs/source/kv_cache/storage_backends/nixl.rst
The example yaml file, which looks like
```yaml
chunk_size: 256
enable_nixl: true
nixl_buffer_size: 1073741824 # 1GB
nixl_buffer_device: cpu
extra_config: {enable_nixl_storage: true, nixl_backend: POSIX, \
               nixl_file_pool_size: 64, nixl_path: /mnt/nixl/cache/}
```
is not correct. The option `enable_nixl` should not be false.
If the option `enable_nixl` is true, then code in config.py
```python
    if self.enable_nixl:
        assert self.nixl_role is not None
```
will report failure since neither 'sender' nor 'receiver' is set. 
But as a storage backend for offload/onboard, no role should be set.
So line `enable_nixl: true` should be removed

PR #1450, opened 2025-08-27
[Doc] update doc for nixl storage backend

ziruiliu:
remove enabel_nixl because nixl storage backend does not need this option

see issue: https://github.com/LMCache/LMCache/issues/1449


ApostaC (approved):
Good catch! Have you tested it with the new config? 

Happy to merge it if it now works on your side.

ziruiliu:
> Good catch! Have you tested it with the new config?
> 
> Happy to merge it if it now works on your side.

nixl storage backend works fine after removing this line. 
There might other issue if request is long but buffer size is small, but that's another case

PR #1466, opened 2025-08-29
support start lookup server on other rank

chunxiaozheng:
If mla is enabled, the read\write operations and lookup server will default to worker0, which will result in a higher workload for worker0 than other workers.

This PR support start lookup server on other worker, use `mla_lookup_server_worker_id` to replace 'create_lookup_server_only_on_worker_0_for_mla', the new config is as follows:
```
mla_lookup_server_worker_id: 1
```

if mla is enabled, the default value of `mla_lookup_server_worker_id` is 0, which means start lookup server on worker 0; if mla is not enabled, the default value of `mla_lookup_server_worker_id` is -1,  negative number means start lookup server on all workers.

### DIFF
--- a/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
@@ -36,12 +36,10 @@ class LMCacheAsyncLookupClient(LookupClientInterface):
     Related extra_config:
     - mla_lookup_server_worker_id:
         is a flag to control whether to create lookup server only on one worker.
-        if mla is not enabled(return -1):
-            start lookup server on all workers
-        if mla is enabled(default is 0):
-            - if mla_lookup_server_worker_id < 0, start lookup server on all workers
-            - if mla_lookup server_worker_id >= 0, start lookup server on the given
-              worker
+        if mla is not enabled, default is -1;
+        if mla is enabled, default is 0;
+        - if mla_lookup_server_worker_id < 0, start lookup server on all workers
+        - if mla_lookup server_worker_id >= 0, start lookup server on the given worker
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


chunxiaozheng:
@maobaolong @Shaoting-Feng could you help take a look?

hickeyma (changes requested):
Thanks @chunxiaozheng for the PR.

The name `create_lookup_server_only_on_one_worker_for_mla` seems a bit clunky andnmot very user friendly.

I wonder would the name `mla_lookup_server_on_worker_rank`,  read better? I'm not sure we need verb `create` in a config item either?

chunxiaozheng:
> Thanks @chunxiaozheng for the PR.
> 
> The name `create_lookup_server_only_on_one_worker_for_mla` seems a bit clunky andnmot very user friendly.
> 
> I wonder would the name `mla_lookup_server_on_worker_rank`, read better? I'm not sure we need verb `create` in a config item either?

@hickeyma Thanks for your review! I will update later.

chunxiaozheng:
@hickeyma I have discussed with @maobaolong offline and updated this PR, could you help take a look? cc @maobaolong

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -47,17 +51,17 @@ def __init__(
...
+        self.mla_lookup_server_worker_id = config.get_mla_lookup_server_worker_id(
+            metadata.use_mla
         )
+        assert self.mla_lookup_server_worker_id < metadata.world_size
         ranks = self.tensor_parallel_size
         self.push_sockets = []
-        if self.create_lookup_server_only_on_worker_0_for_mla:
+        if self.mla_lookup_server_worker_id >= 0:
             ranks = 1
         for tp_rank in range(ranks):
```
maobaolong: The loop for range is not suitable for `self.mla_lookup_server_worker_id >= 0` i guess,
Would you like to generate a list and add all rank into it? You can loop by use that list instead of the `range(rank)`, maybe this can make code clearer than before, WDYT?

chunxiaozheng: @maobaolong Thanks for your review, I have updated, PTAL.

maobaolong (approved):
@chunxiaozheng LGTM after this updated, it looks more flexible.

chunxiaozheng:
hi, @hickeyma , could you help take another look?

hickeyma (approved):

PR #1479, opened 2025-09-01
[fix] fix to workaround for dp !=1

maobaolong:
Without this PR, multi DP will encountered the port in use issue while use internal api server.

If we enabled the internal api server in lmcache process, we assumed it should be only one process for one tp rank before, but when we set DP > 1, there are more dp ranks in a tp rank, so it cannot started, but port conflict.


YaoJiayi (commented):
Can you explain this a bit?

maobaolong:
@YaoJiayi I add a explain to the description.

YaoJiayi (approved):
Lgtm

chunxiaozheng (approved):
lgtm

PR #1480, opened 2025-09-01
Fix device ordinal overflow bug in retrieve broadcast for TP > 8

yoo-kumaneko:

When we are using cross-node TP, like when TP = 16, the current code will attempt to create a tensor on cuda device id that is >= 8. In a word, we should use the local rank rather than the global rank.


YaoJiayi (approved):
lgtm

maobaolong (approved):
LGTM for this PR. It would be nice to submit a followup PR to add local_rank to the metadata to cover the partial gpu device visible situation..

PR #1486, opened 2025-09-02
Use a more robust approach to get local rank

yoo-kumaneko:
Use the global rank to mod device count implicitly assumes that all nodes have the same number of devices. We should use a more robust approach.


maobaolong:
@yoo-kumaneko There are style issues here.

maobaolong:
@yoo-kumaneko Would you like to do something to let this PR keep active?

yoo-kumaneko:
> @yoo-kumaneko Would you like to do something to let this PR keep active?

Yes. Let me update it.

PR #1493, opened 2025-09-02
[bugfix] Fix missing parent_allocator while init TensorMemoryObj for save_only_first_rank

maobaolong:
`TypeError: TensorMemoryObj.__init__() missing 1 required positional argument: 'parent_allocator'`

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1204,7 +1204,9 @@ class LMCacheEngine:
                 self.broadcast_fn(tensor, self.metadata.first_rank)
 
                 # Create temporary memory object (key not needed for other ranks)
-                memory_obj = TensorMemoryObj(raw_data=tensor, metadata=metadata)
+                memory_obj = TensorMemoryObj(
+                    raw_data=tensor, metadata=metadata, parent_allocator=None
+                )
                 reordered_chunks.append((None, memory_obj, start, end))
 
     def _is_passive(self):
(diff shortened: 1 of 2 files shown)


maobaolong:
@ApostaC Would you like to take a look at this PR? Thanks!

hickeyma (changes requested):
@maobaolong `TensorMemoryObj` is initialized with `parent_allocator` sometimes set to `None` and other times set to a value. I would think then that its best to not set a default value to the method as the method expects a value explicitly set. 

I am thinking therefore that it might be best to change the line `memory_obj = TensorMemoryObj(raw_data=tensor, metadata=metadata)` in https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/cache_engine.py#L1207 to `memory_obj = TensorMemoryObj(raw_data=tensor, metadata=metadata, parent_allocator=None)` instead.

@ApostaC You might know best however.

maobaolong:
@hickeyma Thanks for your suggestion. Updated, PTAL.

hickeyma (approved):
LGTM, thanks @maobaolong.

Waiting on @ApostaC opinion as well before merging.

ApostaC (approved):
LGTM! Not sure why this is not captured by the linter though.

ApostaC:
Btw, @maobaolong , since we have the comprehensive test ready, do you think it will be useful to include all those deepseek-related optimizations into the comprehensive test CI pipeline?

Right now it's running on a 2-GPU machine and should be able to run deepseek v2 lite with TP=2.

If you are interested, we can have a call together with Shaoting and let him introduce how to write a new comprehensive test to your team if needed. I believe it should be very straightforward.

maobaolong:
@ApostaC Thank you very much for remind me, this is on our way, and @yoo-kumaneko will do something about the comprehensive test.

And it would be very nice if you and @Shaoting-Feng can give a introduction by a call!

What time is nice for you?

PR #1504, opened 2025-09-03
[Improvement][PluginFramework] Add default interpreter from plugin script

maobaolong:
This PR add more default interpreter, and we can find existing one.

### DIFF
--- a/lmcache/v1/plugin/plugin_launcher.py
+++ b/lmcache/v1/plugin/plugin_launcher.py
@@ -110,30 +110,34 @@ class PluginLauncher:
 
     def _get_interpreter(self, file: Path) -> str:
         """Get interpreter from first line comment"""
+        interpreters = []
         try:
             with open(file, "r", encoding="utf-8") as f:
                 first_line = f.readline().strip()
                 if first_line.startswith("#!"):
                     # Extract interpreter paths by comma-separated
                     interpreters_str = first_line[2:].strip()
-                    # Try each interpreter until we find one that exists
-                    for interp in interpreters_str.split(","):
-                        interp = interp.strip()
-                        resolved_interp = shutil.which(interp)
-                        if resolved_interp:
-                            return resolved_interp
-
+                    interpreters = interpreters_str.split(",")
                     # If none found, fall through to the default behavior
-        except Exception:
+        except Exception as e:
+            logger.error(f"Error reading plugin file {file}: {e}")
             pass
 
         # Fallback to default interpreters
         if file.suffix == ".py":
-            return "python"
+            interpreters.append("python")
+            interpreters.append("python3")
         elif file.suffix == ".sh":
-            return "bash"
-        else:
-            raise ValueError(f"Unsupported plugin type: {file.suffix}")
+            interpreters.append("bash")
+
+        # Try each interpreter until we find one that exists
+        for interp in interpreters:
+            interp = interp.strip()
+            resolved_interp = shutil.which(interp)
+            if resolved_interp:
+                return resolved_interp
+
+        raise ValueError(f"No valid interpreter found for {file} from {interpreters}")
 
     def _capture_plugin_output(self, proc: subprocess.Popen, plugin_name: str):
         """Continuously capture and log plugin output"""
--- a/lmcache/v1/plugin/plugin_launcher.py
+++ b/lmcache/v1/plugin/plugin_launcher.py
@@ -126,7 +126,8 @@ class PluginLauncher:
             interpreters.append("python")
             interpreters.append("python3")
         elif file.suffix == ".sh":
-            interpreters.append("bash")
+            else:
+                raise ValueError(f"Plugin type {file.suffix} not supported ")
 
         # Try each interpreter until we find one that exists
         for interp in interpreters:
--- a/lmcache/v1/plugin/plugin_launcher.py
+++ b/lmcache/v1/plugin/plugin_launcher.py
@@ -3,6 +3,7 @@
 from pathlib import Path
 import atexit
 import os
+import shutil
 import subprocess
 import threading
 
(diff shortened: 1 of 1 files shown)


maobaolong:
@sammshen Hi, would you like to take a look at this PR? Thanks!

hickeyma (changes requested):
@maobaolong Do you mind providing a concrete example in the description for this PR?

maobaolong:
> @maobaolong Do you mind providing a concrete example in the description for this PR?

Done

hickeyma:
> #!/opt/venv/bin/python,python3

@maobaolong I'm not sure that this is good practice. Do we really want to support this?

sammshen:
LGTM, I am not sure about interpreter selection best practices so I will defer to @hickeyma here :)

sammshen (approved):

hickeyma (changes requested):
I'm slow to add any code unless its really needed.  I am not sure that this enhancement is required.

maobaolong:
@hickeyma yeah, this PR is necessary because we use python3 in our image, but python3 cannot be found if we don't use UV or do not source the activite file within the lmcache image.

maobaolong:
@hickeyma I add a new commit to support try each default interpreters, this can make our scenario pass too, if you don't like the comma-separated handling multi-interpreter approach, i am ok to revert the first commit of this PR. PTAL.

hickeyma:
> @hickeyma I add a new commit to support try each default interpreters, this can make our scenario pass too, if you don't like the comma-separated handling multi-interpreter approach, i am ok to revert the first commit of this PR. PTAL.

@maobaolong I appreciate that you might have certain things you need to do on your side but as an open source project we need to follow standard ways of doing things to provide best support to all. I cannot find anywhere that says you can have multiple interpreters in a script.  This is standard stock answer:

"In Python & Bash, a script can only be executed by one interpreter at a time, specified by the shebang (#!) at the top of the file. You can't have multiple interpreters simultaneously run the same script. Technically, only the first line matters for the shebang. Anything after that is treated as a comment."

If you can point me to where it recommends and uses multiple interpreters then I am happy to take a look.

maobaolong:
@hickeyma Thanks for your reply, i revert the multiple interpreter support, only left the default python interpreter. PTAL

lmcache/v1/plugin/plugin_launcher.py
```diff
@@ -109,22 +110,32 @@ def _launch_plugin(self, file: Path):
...
+        interpreters = []
         try:
             with open(file, "r", encoding="utf-8") as f:
                 first_line = f.readline().strip()
                 if first_line.startswith("#!"):
                     # Extract interpreter path from shebang
                     return first_line[2:].strip()
-        except Exception:
+        except Exception as e:
+            logger.error(f"Error reading plugin file {file}: {e}")
```
hickeyma: Error reading plugin file {file}: {e} --> Error reading interpreter from plugin file {file} - using default interpreters: {e}

lmcache/v1/plugin/plugin_launcher.py
```diff
@@ -109,22 +110,32 @@ def _launch_plugin(self, file: Path):
...
         # Fallback to default interpreters
         if file.suffix == ".py":
-            return "python"
+            interpreters.append("python")
+            interpreters.append("python3")
         elif file.suffix == ".sh":
-            return "bash"
-        else:
-            raise ValueError(f"Unsupported plugin type: {file.suffix}")
+            interpreters.append("bash")
```
hickeyma, line 132:
```suggestion
            else:
                raise ValueError(f"Plugin type {file.suffix} not supported ")
```

lmcache/v1/plugin/plugin_launcher.py
```diff
@@ -109,22 +110,32 @@ def _launch_plugin(self, file: Path):
...
+
+        # Try each interpreter until we find one that exists
+        for interp in interpreters:
```
hickeyma: interp -> interpreter

lmcache/v1/plugin/plugin_launcher.py
```diff
@@ -109,22 +110,32 @@ def _launch_plugin(self, file: Path):
...
+            interp = interp.strip()
+            resolved_interp = shutil.which(interp)
+            if resolved_interp:
```
hickeyma: resolved_interp --> resolved_interpreter

hickeyma (changes requested):
Thanks @maobaolong for updates. Some small feedback and suggestion inline.

lmcache/v1/plugin/plugin_launcher.py
maobaolong: Done

lmcache/v1/plugin/plugin_launcher.py
maobaolong: done

lmcache/v1/plugin/plugin_launcher.py
maobaolong: done

maobaolong:
@hickeyma All the suggestion has been addressed, PTAL.

lmcache/v1/plugin/plugin_launcher.py
hickeyma: This doesn't look done

lmcache/v1/plugin/plugin_launcher.py
hickeyma: This doesn't look done

lmcache/v1/plugin/plugin_launcher.py
hickeyma: This doesn't look done

hickeyma (changes requested):
@maobaolong I think you might have forgot to push the changes?

Also, there is a linting and formatting issue to address: https://github.com/LMCache/LMCache/actions/runs/17496722036/job/49699345896?pr=1504

maobaolong:
@hickeyma Sorry, it's my fault, my commit was prevented by pre-commit-hook, I didn't realize it...

Sorry again! Will be careful next.

hickeyma (approved):
LGTM, thanks @maobaolong

hickeyma:
> @hickeyma Sorry, it's my fault, my commit was prevented by pre-commit-hook, I didn't realize it...
Sorry again! Will be careful next.

Thats ok @maobaolong, these things happen.

sammshen:
Thanks for the detailed feedback @hickeyma !!

PR #1515, opened 2025-09-04
[CD]: Lightweight Dockerfile

sammshen:
Screenshot 2025-09-04 at 2 22 12 AM

Lightweight version of the `lmcache/vllm-openai` container image which does not contain NIXL installation.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -253,3 +253,14 @@ jobs:
               run: |
                 docker push lmcache/vllm-openai:latest
                 docker push lmcache/vllm-openai:${{ env.LATEST_TAG }}
+            
+            - name: Build lmcache/vllm-openai:light
+              run: | 
+                docker build \
+                --tag lmcache/vllm-openai:light --tag lmcache/vllm-openai:${{ env.LATEST_TAG }}-light \
+                --file docker/Dockerfile.light
+            
+            - name: Push lmcache/vllm-openai:light image to DockerHub
+              run: |
+                docker push lmcache/vllm-openai:light
+                docker push lmcache/vllm-openai:${{ env.LATEST_TAG }}-light
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -254,13 +254,13 @@ jobs:
                 docker push lmcache/vllm-openai:latest
                 docker push lmcache/vllm-openai:${{ env.LATEST_TAG }}
             
-            - name: Build lmcache/vllm-openai:light
+            - name: Build lmcache/vllm-openai:lightweight
               run: | 
                 docker build \
-                --tag lmcache/vllm-openai:light --tag lmcache/vllm-openai:${{ env.LATEST_TAG }}-light \
-                --file docker/Dockerfile.light
+                --tag lmcache/vllm-openai:lightweight --tag lmcache/vllm-openai:${{ env.LATEST_TAG }}-lightweight \
+                --file docker/Dockerfile.lightweight
             
-            - name: Push lmcache/vllm-openai:light image to DockerHub
+            - name: Push lmcache/vllm-openai:lightweight image to DockerHub
               run: |
-                docker push lmcache/vllm-openai:light
-                docker push lmcache/vllm-openai:${{ env.LATEST_TAG }}-light
+                docker push lmcache/vllm-openai:lightweight
+                docker push lmcache/vllm-openai:${{ env.LATEST_TAG }}-lightweight
--- a/docker/Dockerfile.light
+++ b/docker/Dockerfile.light
@@ -1,0 +1,22 @@
+# vllm/vllm-openai:latest currently has lmcache bundled
+# this is in case vllm does not have lmcache bundled in the future
+# this docker file will not be able to run PD (no nixl)
+FROM vllm/vllm-openai:latest
+
+RUN uv pip install --system --no-cache lmcache
+
+
+# Build: 
+# docker build -f Dockerfile.light -t 'lmcache/vllm-openai:light' .
+
+# Run: 
+# export HF_TOKEN=<your_huggingface_token>
+# docker run --runtime nvidia --gpus all \
+#    -v ~/.cache/huggingface:/root/.cache/huggingface \
+#    --env "HUGGING_FACE_HUB_TOKEN=$HF_TOKEN" \
+#    -p 8000:8000 \
+#    --ipc=host \
+#    lmcache/vllm-openai:light \
+#    --model Qwen/Qwen3-0.6B \
+#    --kv-transfer-config \
+#    '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
--- a/docs/source/production/docker_deployment.rst
+++ b/docs/source/production/docker_deployment.rst
@@ -28,5 +28,6 @@ See example run file in `docker <https://github.com/LMCache/LMCache/tree/dev/doc
 
 .. note::
     DockerHub contains the following image types:
-    - Nightly build images of LMCache and vLLM latest code
-    - Images of stable releases of LMCache and vLLM
+    - Nightly build images of LMCache and vLLM latest code (e.g. tagged with `latest-nightly` and `nightly-<date>`)
+    - Images of stable releases of LMCache and vLLM (tagged with `v0.x.x`, the exact version of vllm a version of lmcache was built with can be discovered by consulting the compatibility matrix inside of `installation <../installation.rst>`_)
+    - Light image without nixl (tagged with `light`)
--- a/docs/source/production/docker_deployment.rst
+++ b/docs/source/production/docker_deployment.rst
@@ -27,7 +27,10 @@ The image name and tag can be found on DockerHub - `LMCache/vllm-openai <https:/
 See example run file in `docker <https://github.com/LMCache/LMCache/tree/dev/docker>`_ for more details.
 
 .. note::
+
     DockerHub contains the following image types:
+
     - Nightly build images of LMCache and vLLM latest code (e.g. tagged with `latest-nightly` and `nightly-<date>`)
     - Images of stable releases of LMCache and vLLM (tagged with `v0.x.x`, the exact version of vllm a version of lmcache was built with can be discovered by consulting the compatibility matrix inside of `installation <../installation.rst>`_)
-    - Light image without nixl (tagged with `light`)
+    
+    - Lightweight image that cannot run PD disaggregation (tagged with `lightweight`)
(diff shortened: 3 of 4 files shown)


hickeyma (changes requested):
@sammshen Do you mind explaining in the description the reason for this light container file? 

Can you also explain what version of LMcache does "vllm/vllm-openai:latest currently has lmcache bundled" bundle and why then are we creating a image for the bundling?

sammshen:
Hi @hickeyma , thx for the clarifying questions :)

the problem is: 
1. in the future, we cannot guarantee that vllm will always bundle lmcache. for example, if they find that lmcache breaks their release pipeline, they may fallback to not importing it for a particular image
2. the version of lmcache inside of the vllm/vllm-openai file may not be latest lmcache. this helps us build our own on release with latest lmcache

.github/workflows/publish.yml
```diff
@@ -253,3 +253,14 @@ jobs:
               run: |
                 docker push lmcache/vllm-openai:latest
                 docker push lmcache/vllm-openai:${{ env.LATEST_TAG }}
+
```
sammshen, line 256: @hickeyma any thoughts on this new addition to the release pipeline?

hickeyma: This has been there for a good amount of time. It was added to create an image when we cut a release.

hickeyma (commented):
> Hi @hickeyma , thx for the clarifying questions :)
the problem is:
in the future, we cannot guarantee that vllm will always bundle lmcache. for example, if they find that lmcache breaks their release pipeline, they may fallback to not importing it for a particular image
    the version of lmcache inside of the vllm/vllm-openai file may not be latest lmcache. this helps us build our own on release with latest lmcache

LMCache already does 2 images builds integrated with vLLM already as follows:

1. Nightly which builds the latest LMCache and integrated with the latest nightly vLLM build
2. At release which build a image of the latest LMcache release integrated with the latest vLLM release

@sammshen Maybe I am wrong but this PR seems to be doing the same image build as pt 2 above.

sammshen:
we were hoping to create a smaller docker image. Ours is currently ~10GB larger than vllm's because of nixl dependencies and maybe inefficiencies (the screenshots are the compressed sizes on dockerhub, so the difference needs to be multipllied). 

Screenshot 2025-09-05 at 11 27 02 AM

Screenshot 2025-09-05 at 11 27 17 AM

Will work on optimizing the size and build speed of the docker image more

hickeyma (commented):
@sammshen Thanks for explanation on having an image (minus nixl), it makes sense now. Let me know when the pr is finished and ready to review.

sammshen:
@hickeyma  @ApostaC I think it is ready for review!

docs/source/production/docker_deployment.rst
```diff
@@ -28,5 +28,6 @@ See example run file in `docker <https://github.com/LMCache/LMCache/tree/dev/doc
 
 .. note::
     DockerHub contains the following image types:
-    - Nightly build images of LMCache and vLLM latest code
-    - Images of stable releases of LMCache and vLLM
\ No newline at end of file
+    - Nightly build images of LMCache and vLLM latest code (e.g. tagged with `latest-nightly` and `nightly-<date>`)
+    - Images of stable releases of LMCache and vLLM (tagged with `v0.x.x`, the exact version of vllm a version of lmcache was built with can be discovered by consulting the compatibility matrix inside of `installation <../installation.rst>`_)
+    - Light image without nixl (tagged with `light`)
```
ApostaC: Can we add another sentence here saying:"it's XXX GB smaller, but cannot run PD disaggregation."

docker/Dockerfile.light
```diff
@@ -0,0 +1,22 @@
...
+
+
+# Build: 
+# docker build -f Dockerfile.light -t 'lmcache/vllm-openai:light' .
+
+# Run: 
+# export HF_TOKEN=<your_huggingface_token>
+# docker run --runtime nvidia --gpus all \
+#    -v ~/.cache/huggingface:/root/.cache/huggingface \
+#    --env "HUGGING_FACE_HUB_TOKEN=$HF_TOKEN" \
```
ApostaC: The environment name should be `HF_TOKEN`.
Also, `Qwen3-0.6B` doesn't need HF token to download.

ApostaC (approved):
Otherwise LGTM

docker/Dockerfile.light
```diff
@@ -0,0 +1,22 @@
+# vllm/vllm-openai:latest currently has lmcache bundled
+# this is in case vllm does not have lmcache bundled in the future
+# this docker file will not be able to run PD (no nixl)
+FROM vllm/vllm-openai:latest
+
+RUN uv pip install --system --no-cache lmcache
```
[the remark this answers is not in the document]
hickeyma: Its ok, want to get the latest

docs/source/production/docker_deployment.rst
```diff
@@ -28,5 +28,6 @@ See example run file in `docker <https://github.com/LMCache/LMCache/tree/dev/doc
 
 .. note::
     DockerHub contains the following image types:
-    - Nightly build images of LMCache and vLLM latest code
-    - Images of stable releases of LMCache and vLLM
\ No newline at end of file
+    - Nightly build images of LMCache and vLLM latest code (e.g. tagged with `latest-nightly` and `nightly-<date>`)
+    - Images of stable releases of LMCache and vLLM (tagged with `v0.x.x`, the exact version of vllm a version of lmcache was built with can be discovered by consulting the compatibility matrix inside of `installation <../installation.rst>`_)
```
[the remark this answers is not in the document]
hickeyma: Its there

docker/Dockerfile.light
hickeyma: Agree

docs/source/production/docker_deployment.rst
hickeyma: Agree

docker/Dockerfile.light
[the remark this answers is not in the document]
hickeyma: Agree

docker/Dockerfile.light
```diff
@@ -0,0 +1,22 @@
+# vllm/vllm-openai:latest currently has lmcache bundled
+# this is in case vllm does not have lmcache bundled in the future
+# this docker file will not be able to run PD (no nixl)
```
hickeyma: For better clarity, I would change the above comments to say something like the following:

This is a lightweight image of the latest release of LMCache (without NIXL) bundled with the latest release of vLLM.
Note: You will not be able to run Prefill-Decode Disaggregation (PD) with this image because NIXL is not installed.

docker/Dockerfile.light
```diff
@@ -0,0 +1,22 @@
...
+# export HF_TOKEN=<your_huggingface_token>
+# docker run --runtime nvidia --gpus all \
+#    -v ~/.cache/huggingface:/root/.cache/huggingface \
+#    --env "HUGGING_FACE_HUB_TOKEN=$HF_TOKEN" \
+#    -p 8000:8000 \
+#    --ipc=host \
+#    lmcache/vllm-openai:light \
+#    --model Qwen/Qwen3-0.6B \
+#    --kv-transfer-config \
+#    '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
```
hickeyma: Would it be better pass the LMCache args? For example:

```
 --env "LMCACHE_CHUNK_SIZE=256" \
 --env "LMCACHE_LOCAL_CPU=True" \
 --env "LMCACHE_MAX_LOCAL_CPU_SIZE=5" \
```

hickeyma: Is `--ipc=host` required?

docs/source/production/docker_deployment.rst
hickeyma: Light image --> Lightweight image

hickeyma: @sammshen This is broken prior to this PR but do you mind also fixing this list as it down not line up as a list when rendered?

hickeyma (changes requested):
Getting there @sammshen, thanks for your patience.

Some comments line to address when you get a chance.

sammshen:
Thanks for the review @hickeyma 

Fixed the RST syntax issues:
Screenshot 2025-09-12 at 3 31 53 AM

hickeyma (changes requested):
Nearly there @sammshen.

Do you mind addressing the open comments:

- https://github.com/LMCache/LMCache/pull/1515#discussion_r2339304179
- https://github.com/LMCache/LMCache/pull/1515#discussion_r2339320084

docker/Dockerfile.light
sammshen: Sorry I meant to respond to this! I believe `--ipc=host` is for the vllm shared memory for tensor parallelism and not LMCache specific. I copied the deployment command from vllm.

sammshen: any reason to pass the LMCache args when they are the defaults?

sammshen:
Sorry about that, I forgot to respond to those! @hickeyma

docker/Dockerfile.light
hickeyma: > Sorry I meant to respond to this! I believe --ipc=host is for the vllm shared memory for tensor parallelism and not LMCache specific. I copied the deployment command from vllm.

Do you mind confirming if we need this flag as its open a potential attack vector? Also we recommend `--network host` flag in https://github.com/LMCache/LMCache/blob/dev/docker/example_run.sh. I think we should be fully aware of using the need for these flags.

hickeyma: > any reason to pass the LMCache args when they are the defaults?

I suppose to be explicit to the user and also provide the flags so if a user wants to change them.

hickeyma:
> any reason to pass the LMCache args when they are the defaults?

Not a bother @sammshen. I have responded inline.

docker/Dockerfile.light
sammshen: I think not since I just mimicked: 
https://docs.vllm.ai/en/latest/deployment/docker.html?h=

sammshen:
@hickeyma any blocking issues? :)

docker/Dockerfile.light
hickeyma: >I think not since I just mimicked:
https://docs.vllm.ai/en/latest/deployment/docker.html?h=

Ok, thanks for the clarification.

hickeyma (approved):
LGTM, thanks @sammshen for following up on all feedback. This PR provides a nice more streamlined image.

@sammshen I see the build and run commands are documented in the container file. As a follow on it might be beneficial to the user if the build and run commands had their own example scripts like the full image does?

sammshen:
@hickeyma I agree with you, will add those soon!

PR #1516, opened 2025-09-04
[doc] add UCX_NET_DEVICES as kind of reminder

panpan0000:
This just a tips to remind the users:

Nixl uses **UCX** instead of **NCCL**,
so Nixl requires UCX_** env var,
previous NCCL_IB_HCA, NCCL_SOCKET_IFNAME are useless to Nixl.

FIX #NA


panpan0000:
@YaoJiayi 

not eveybody are familar with NIXL so much, the story is we are on a system with multiple NIC cards there but some of them should  not be chosen due to some routing issue, so we need  `UCX_NET_DEVICES ` to pick specific cards .

unfortunately , Nvidia UCX document is poor....

so any tip will save lots of people's time.

YaoJiayi (commented):
so should we add sth in the readme?

panpan0000:
> so should we add sth in the readme?

I think there're too many things to explain in README beyond this , like I remembered somebody asked about `spawn` before .
So defer you expert to summarize all of them or kick off it , maybe when you have time.

PR #1524, opened 2025-09-05
[infinistore] Remove infinistore dependency

maobaolong:
Remnove infinistore dependency, and who can install it when need, for lmcache core, it should not add each remote storage backend.


YaoJiayi (approved):
lgtm

hickeyma (approved):
LGTM, thanks @maobaolong.

Waiting on CI to finish before merging.

PR #1545, opened 2025-09-09
[bugfix] fix not hit in LocalCpuBackend

chunxiaozheng:
if use request configs and `LocalCpuBackend`, the request configs is as follows:
```
"kv_transfer_params": {
      "lmcache.tag.xxx": 60
 }
```
if the type of value is `int` or other numeric type, when the `lookup` method is called, these params will be transfer to `str`, which will result in not hit in `LocalCpuBackend`.

This PR override the `__eq__` method, use tags_str to judge equals. In my tests, the change works good.

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -101,7 +101,7 @@ class CacheEngineKey:
                 if k.startswith("lmcache.tag."):
                     if tags is None:
                         tags = {}
-                    tags[k[len("lmcache.tag.") :]] = v
+                    tags[k[len("lmcache.tag.") :]] = str(v)
         self.tags = tags
 
     def __hash__(self):
@@ -125,6 +125,19 @@ class CacheEngineKey:
                 "%".join([f"{k}={v}" for k, v in self.tags.items()]),
             )
         )
+
+    def __eq__(self, other):
+        if type(self) == type(other):
+            return (
+                self.fmt == other.fmt
+                and self.model_name == other.model_name
+                and self.world_size == other.world_size
+                and self.worker_id == other.worker_id
+                and self.chunk_hash == other.chunk_hash
+                and self.tags == other.tags
+            )
+
+        return False
 
     def to_string(self):
         s = (
@@ -253,6 +266,12 @@ class LayerCacheEngineKey(CacheEngineKey):
                 self.layer_id,
             )
         )
+
+    def __eq__(self, other):
+        if super().__eq__(other):
+            return self.layer_id == other.layer_id
+
+        return False
 
     def to_string(self):
         s = (
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -127,7 +127,7 @@ class CacheEngineKey:
         )
 
     def __eq__(self, other):
-        if type(self) == type(other):
+        if type(self) is type(other):
             return (
                 self.fmt == other.fmt
                 and self.model_name == other.model_name


lmcache/utils.py
```diff
@@ -126,6 +126,32 @@ def __hash__(self):
             )
         )
 
+    def __eq__(self, other):
+        if not isinstance(other, CacheEngineKey):
+            return False
+
+        if type(self) is CacheEngineKey and type(other) is not CacheEngineKey:
```
maobaolong: Is this condition redundant?

lmcache/utils.py
```diff
@@ -126,6 +126,32 @@ def __hash__(self):
...
+            return False
+
+        self_tags_str = (
+            None
+            if self.tags is None
+            else "%".join([f"{k}={v}" for k, v in self.tags.items()])
```
maobaolong: Does this string concat can be avoid if we just use self.tags equals other.tags?

maobaolong (commented):
@chunxiaozheng Thanks for this fix!

Left some minor suggestion, otherwise, it looks good overall.

lmcache/utils.py
chunxiaozheng: @maobaolong Thanks for your review! This is not redundant, it is to prevent the situation where `other` is a subclass of `CacheEngineKey`.

lmcache/utils.py
chunxiaozheng: direct use `self.tags` and `other.tags`, if the type of value is int or other numeric type, when the lookup method is called, these params will be transfer to str, which will result in not hit.

maobaolong (approved):
LGTM

chunxiaozheng:
@sammshen could you help take another look?

sammshen:
Weird unit test fail: 
```text

________________________________________________________ ERROR collecting tests/v1/test_cache_engine.py ________________________________________________________
--
  | ImportError while importing test module '/var/lib/buildkite-agent/builds/lmcache-unit-test-build-1/lmcache/unit-tests/tests/v1/test_cache_engine.py'.
  | Hint: make sure your test modules/packages have valid Python names.
  | Traceback:
  | ../../../../.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py:90: in import_module
  | return _bootstrap._gcd_import(name[level:], package, level)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | tests/v1/test_cache_engine.py:12: in <module>
  | from utils import (
  | E   ImportError: cannot import name 'DummyLMCacheAsyncLookupServer' from 'utils' (/var/lib/buildkite-agent/builds/lmcache-unit-test-build-1/lmcache/unit-tests/tests/benchmarks/utils.py)


```

sammshen:
Retrying unit test

maobaolong:
@sammshen The failure looks persistent, but I guess it does not caused to this PR?
```

________________________________________________________ ERROR collecting tests/v1/test_cache_engine.py ________________________________________________________
--
  | ImportError while importing test module '/var/lib/buildkite-agent/builds/lmcache-unit-test-build-1/lmcache/unit-tests/tests/v1/test_cache_engine.py'.
  | Hint: make sure your test modules/packages have valid Python names.
  | Traceback:
  | ../../../../.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py:90: in import_module
  | return _bootstrap._gcd_import(name[level:], package, level)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | tests/v1/test_cache_engine.py:12: in <module>
  | from utils import (
  | E   ImportError: cannot import name 'DummyLMCacheAsyncLookupServer' from 'utils' (/var/lib/buildkite-agent/builds/lmcache-unit-test-build-1/lmcache/unit-tests/tests/benchmarks/utils.py)


```

chunxiaozheng:
> Weird unit test fail:
> 
> ```
> 
> ________________________________________________________ ERROR collecting tests/v1/test_cache_engine.py ________________________________________________________
> --
>   | ImportError while importing test module '/var/lib/buildkite-agent/builds/lmcache-unit-test-build-1/lmcache/unit-tests/tests/v1/test_cache_engine.py'.
>   | Hint: make sure your test modules/packages have valid Python names.
>   | Traceback:
>   | ../../../../.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/importlib/__init__.py:90: in import_module
>   | return _bootstrap._gcd_import(name[level:], package, level)
>   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   | tests/v1/test_cache_engine.py:12: in <module>
>   | from utils import (
>   | E   ImportError: cannot import name 'DummyLMCacheAsyncLookupServer' from 'utils' (/var/lib/buildkite-agent/builds/lmcache-unit-test-build-1/lmcache/unit-tests/tests/benchmarks/utils.py)
> ```

@sammshen this error may be fixed by #1536, and I have rebased the `dev` branch.

sammshen (approved):
LGTM!

Issue #1482, opened 2025-09-01
[Bug] nixl storage backend has conflicts with local cpu backend


ziruiliu:
In document docs/source/kv_cache/storage_backends/nixl.rst, example yaml is configured like
```yaml
chunk_size: 256
nixl_buffer_size: 1073741824 # 1GB
nixl_buffer_device: cpu
extra_config: {enable_nixl_storage: true, nixl_backend: POSIX, \
               nixl_file_pool_size: 64, nixl_path: /mnt/nixl/cache/}
```

with such configuration, local_cpu_backend is by default also enabled, which may causes assertion failure at eviction 

<details>
<summary>At this point, use_hot is True, number of items in hot_cache is full</summary>
[multiproc_executor.py:596] AssertionError<br>
[multiproc_executor.py:596]   File "/home/LMCache-nixl/lmcache/integration/vllm/vllm_v1_adapter.py", line 963, in wait_for_save<br>
[multiproc_executor.py:596]     self.lmcache_engine.store(<br>
[multiproc_executor.py:596]   File "/home/ve_cache_dev/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context<br>
[multiproc_executor.py:596]     return func(*args, **kwargs)<br>
[multiproc_executor.py:596]            ^^^^^^^^^^^^^^^^^^^^^<br>
[multiproc_executor.py:596]   File "/home/LMCache-nixl/lmcache/v1/cache_engine.py", line 263, in store<br>
[multiproc_executor.py:596]     memory_obj = self.storage_manager.allocate(<br>
[multiproc_executor.py:596]                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^<br>
[multiproc_executor.py:596]   File "/home/LMCache-nixl/lmcache/v1/storage_backend/storage_manager.py", line 118, in allocate<br>
[multiproc_executor.py:596]     return self.allocator_backend.allocate(<br>
[multiproc_executor.py:596]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^<br>
[multiproc_executor.py:596]   File "/home/LMCache-nixl/lmcache/v1/storage_backend/local_cpu_backend.py", line 264, in allocate<br>
[multiproc_executor.py:596]     assert isinstance(self.memory_allocator, MixedMemoryAllocator) or isinstance(<br>
[...]

PR #1558, opened 2025-09-10
[Core] fix nixl_storage_backend assert

tshmilnvidia:
an assert is hit when local_cpu_backend tries to allocate, because allocator type is not MixedMemoryAllocator.

In order to adapt to this assert, change nixl_storage_backend to use MixedMemoryAllocator with paging=true, which will create a PagedTensorMemoryAllocator internally. then use this internal allocator for NIXL instead of creating PagedTensorMemoryAllocator.

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -1464,24 +1464,29 @@ class MixedMemoryAllocator(MemoryAllocatorInterface):
         """
 
         self.numa_mapping = kwargs.get("numa_mapping", None)
+        self.device = kwargs.get("device", None)
 
         self.size = size
 
-        if self.numa_mapping:
-            current_device_id = torch.cuda.current_device()
-            gpu_to_numa_mapping = self.numa_mapping.gpu_to_numa_mapping
-            assert current_device_id in gpu_to_numa_mapping, (
-                f"Current device {current_device_id} is not in the GPU NUMA mapping."
-            )
-            numa_id = gpu_to_numa_mapping[current_device_id]
-            ptr = lmc_ops.alloc_pinned_numa_ptr(size, numa_id)
+        if self.device == "cpu":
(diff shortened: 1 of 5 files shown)


barneuman:
[https://github.com/LMCache/LMCache/issues/1482](url)

ziruiliu:
Yes. I also found this issue in https://github.com/LMCache/LMCache/issues/1482
I also posted a PR as https://github.com/LMCache/LMCache/pull/1506 to add PagedTensorMemoryAllocator check in cpu backend. It is good to avoid PagedTensorMemoryAllocator and use MixedMemoryAllocator instead

lmcache/v1/memory_management.py
```diff
@@ -1464,24 +1464,29 @@ def __init__(self, size: int, use_paging: bool = False, **kwargs):
         """
 
         self.numa_mapping = kwargs.get("numa_mapping", None)
+        self.device = kwargs.get("device", None)
```
Shaoting-Feng, line 1467: What does this variable mean?

tshmilnvidia: this is whether the main buffer should be allocated on CPU or GPU (as nixl storage backend allows both options)

ApostaC:
Hey @tshmilnvidia , thanks for the fix! We recently had a bigger refactor (#1578) to enable using local CPU backend and other GPU-related backends together.

This is done by having separate memory allocators for different backends. 

Although the fixes in the current PR are no longer needed, we truly appreciate your contribution to LMCache!

tshmilnvidia:
> Hey @tshmilnvidia , thanks for the fix! We recently had a bigger refactor (#1578) to enable using local CPU backend and other GPU-related backends together.
> 
> This is done by having separate memory allocators for different backends.
> 
> Although the fixes in the current PR are no longer needed, we truly appreciate your contribution to LMCache!

I went over this refactor and I agree it invalidates this PR. I will close it.

PR #1560, opened 2025-09-10
[Misc] Add MRU cache policy to v1

yunjiangster:
Implement Most Recent Used Cache policy in v1.

Logic is very similar to LRU but we reverse the traversal of cache_dict during eviction, to avoid adding cache_dict as an argument for the `update_on_put` method.


YaoJiayi (approved):
LGTM!

PR #1572, opened 2025-09-11
Support override config by env

maobaolong:
# Motivation

The existing logic of configuration of LMCache is that we could only get config from one of `lmcache.yaml` file or env vars, could not treat env vars as a supplement.

After this PR, LMCache will read config from `lmcache.yaml` first, then override the configs which configured by env var.

In another word, the `env var` config source should treat as higher priority  than file `lmcache.yaml`.

# Test

- lmcahe.yaml
```yaml
chunk_size: 64
remote_serde: "naive"
local_cpu: True
max_local_cpu_size: 0.1
extra_config:
  save_only_first_rank: True
  create_lookup_server_only_on_worker_0_for_mla: True
  audit_exclude_cmds: exists,get
  tag_keys: ["user"]
  plugin.frontend.port: 8080
  external_backend.log_external_backend.module_path: lmc_external_log_backend.lmc_external_log_backend
  external_backend.log_external_backend.class_name: ExternalLogBackend
internal_api_server_enabled: True
internal_api_server_port_start: 9090
internal_api_server_include_index_list: [0, 1]
plugin_locations: ["/opt/venv/lib/python3.12/site-packages/lmcache_frontend/lmcache_plugin/"]
internal_api_server_socket_path_prefix: "/tmp/lmcache_internal_api_server/socket"
# external_backends: "log_external_backend"

```

- env
```
export LMCACHE_CHUNK_SIZE=512
```

- Result
image

The blue rectangle comes from `lmcache.yaml`, the green rectangle comes from `env var`.

### DIFF
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -265,6 +265,10 @@ class LMCacheEngineConfig:
 
         return config.log_config()
 
+    def update_config_from_env(self):
+        """Update the configuration from the environment variables"""
+        pass
+
     def log_config(self) -> "LMCacheEngineConfig":
         """Log all configuration settings"""
         config_dict = {
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -267,6 +267,7 @@ class LMCacheEngineConfig:
 
     def update_config_from_env(self):
         """Update the configuration from the environment variables"""
+        # TODO(baoloongmao): implement this when needed
         pass
 
     def log_config(self) -> "LMCacheEngineConfig":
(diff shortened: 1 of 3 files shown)


sammshen:
Integration Test: https://github.com/LMCache/LMCache/issues/1573

maobaolong:
@sammshen Thank you for the help to pass the Integration tests. Would you like to take a look at this PR?

sammshen (approved):
LGTM!

chunxiaozheng:
@maobaolong quick question: when the config both set in config file and env, use env as the final value?

maobaolong:
> @maobaolong quick question: when the config both set in config file and env, use env as the final value?

@chunxiaozheng Thanks for your question, this should be clarified to the question to let others can aware `env var` is higher priority than file `lmcache.yaml`, and I've add this to the description.

chunxiaozheng (approved):
lgtm

PR #1578, opened 2025-09-12
[Core] Refactor the storage manager to support using CPU & GPU backends at the same time

### LINKED ISSUE #1577: [RFC] Refactor the memory object allocation logic to use local cpu backend with all GPU related backends
# Background and motivation

Recently, there have been a bunch of issues/PRs about the problem when trying to launch the local CPU backend together with other GPU-related backends (NIXL, GDS, etc).
For example, #1482, #1464, #1502, and #1558

The core reason is that the local CPU backend, being used by remote backends and the local disk backend, is not supposed to manage the memory objects on the GPU. 
However, when putting new memory objects into the storage manager, the memory object could be on the GPU if we are using backends like GDSBackend or NixlBackend -- which will trigger errors if there is also a local CPU backend initialized.

Another big problem is that currently, the local CPU backend may hold the GPU allocators (e.g., CuFileMemoryAllocator), which should never happen.

# Proposal

To address this problem, the proposed solution is to do an extra allocation and D2H copy if the memory objects are allocated on the GPU but the local CPU backend is also initialized.

At the code level, each storage backend will be explicitly assigned with an allocator backend (could be itself). More specifically, there will be a new interface:

```python
class StorageBackendInterface:
    # ... Existing codes
    def get_allocator_backend(self) -> AllocatorBackendInterface:
        """Return the allocator backend associated with the current storage backend
        """
        raise NotImplementedError
```

_Note: Such an interface makes sense because every backend needs to allocate the memory object during `get` operation._

With this new interface, the storage manager can check if the current backend's allocator backend is the same as the default allocator backend used in the storage manager during `put`.
- If it's the same, it means that the memory object can be directly put into the storage backend.
- Otherwise, we will call `current_backend.allocator_backend.allocate` and then copy the current memory object into the newly allocated memory object.

ApostaC:
This is a WIP PR to make it able to have CPU & GPU backends at the same time. 

Fixes #1577 

Now we are able to use local CPU and GDS/NixlStorage backends at the same time.

**Example config**:
```yaml
gds_path: "/tmp/gds_path",
cufile_buffer_size: 1024
local_cpu: True
max_local_cpu_size: 5
```

**Core changes**:
- Introduce new functions for `AllocatorBackendInterface`: `initialize_allocator` and `get_memory_allocator`
- Introduce new functions for `StorageBackendInterface`: `get_allocator_backend`
- Add the implementation of the new interfaces in the storage backends.
- Refactor the creation of the storage backends and memory allocators (now CacheEngine won't create memory allocators)
- Refactor `StorageManager.put` to support using CPU and GPU backends at the same time

**Follow ups**:
- [ ] Completely remove the `_Create_memory_allocator` function
- [ ] Change the unit tests to use the new interface
- [ ] New refactors:
  - [ ] extract the eviction logic from storage backends 
  - [ ] only keep two types of memory allocators -- CPU and GPU shared across all backends

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

</details>


YaoJiayi:
@ApostaC Is this ready to be reviewed?

ApostaC:
@YaoJiayi This is ready for review

YaoJiayi:
@ApostaC Will take a look asap today.
Btw, should we include cufile-related stuff in the unit test?

ApostaC:
> Btw, should we include cufile-related stuff in the unit test?

Yeah, we should. Right now seems like the unit test server does not support cufile. We can do it in comprehensive test

lmcache/v1/storage_backend/__init__.py
```diff
@@ -48,6 +42,11 @@ def create_dynamic_backends(
     Looks for backend configurations in config.extra_config and instantiates
     them using the specified module and class names.
     """
+    logger.warning(
+        "Memory allocator will never be passed from outside since"
```
YaoJiayi: Outside from storage_backend?

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -102,7 +151,7 @@ def __init__(
                 config,
                 metadata,
                 self.loop,
-                allocator,
+                # allocator,
```
YaoJiayi: remove

YaoJiayi (approved):
LGTM!

ApostaC:
Merging this to unblock #1579

PR #1586, opened 2025-09-14
Refactor lmcache_get_config to lmcache_get_or_create_config

maobaolong:
This refactor can make sure there is only one instance of `LMCacheEngineConfig` same time

### DIFF
--- a/examples/plugins/scheduler_foo_plugin.py
+++ b/examples/plugins/scheduler_foo_plugin.py
@@ -10,7 +10,7 @@ import signal
 import time
 
 # First Party
-from lmcache.integration.vllm.utils import lmcache_get_config
+from lmcache.integration.vllm.utils import lmcache_get_or_create_config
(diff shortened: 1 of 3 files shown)


maobaolong:
@YaoJiayi Would you like to take a look at this PR? Thanks!

YaoJiayi (commented):
Quick question: since scheduler and worker are in two processes, how is the singleton property maintained?

maobaolong:
@YaoJiayi Thanks for this question， we can only keep singlton within each process.

YaoJiayi:
@maobaolong So when updating the config externally, are we targeting at the scheduler process or worker process?

maobaolong:
@YaoJiayi We can update the frontend as target, and the frontend can update all the processes eventually.

YaoJiayi (approved):
lgtm

maobaolong:
@YaoJiayi Thanks for your review！

PR #1603, opened 2025-09-16
support tmp path in FSConnector

chunxiaozheng:
# DESCEPTION
Support set tmp path in FSConnector. 

There are the following benefits：
- The data is more intuitive. The data directory only contains the data files, and all temporary files are placed in a temporary directory.
- When there is a large amount of data, it is convenient to clean up temporary files

# HOW TO USE
set `fs_connector_relative_tmp_dir` in `extra_config` to enable this feature, such as
```
remote_url: "fs://host:0/data/test"
extra_config:
  fs_connector_relative_tmp_dir: "tmp"
```
the `fs_connector_relative_tmp_dir` must be a relative path, if `fs_connector_relative_tmp_dir` is set and use `FSConnector`, in the example above, tmp path `/data/test/tmp` will be created, the `put` operation will write data file in `/data/test/tmp`  first, and then move the data file to`/data/test` .

### DIFF
--- a/lmcache/v1/storage_backend/connector/fs_connector.py
+++ b/lmcache/v1/storage_backend/connector/fs_connector.py
@@ -155,6 +155,7 @@ class FSConnector(RemoteConnector):
     async def put(self, key: CacheEngineKey, memory_obj: MemoryObj):
         """Store data to file system"""
         final_path, temp_path = self._get_file_path(key, True)
+        assert temp_path is not None
 
         try:
             # Prepare metadata
--- a/lmcache/v1/storage_backend/connector/fs_connector.py
+++ b/lmcache/v1/storage_backend/connector/fs_connector.py
@@ -68,11 +68,8 @@ class FSConnector(RemoteConnector):
             if self.relative_tmp_dir is not None:
                 (path / self.relative_tmp_dir).mkdir(parents=False, exist_ok=True)
 
-    def _get_file_path(
-        self, key: CacheEngineKey, generate_tmp_path=False
-    ) -> Tuple[Path, Optional[Path]]:
-        """Get file path and tmp path for the given key"""
-        # If there's only one path, use it directly
+    def _get_base_path(self, key: CacheEngineKey) -> Path:
+        """Get file base path for the given key"""
         if len(self.base_paths) == 1:
             base_path = self.base_paths[0]
         else:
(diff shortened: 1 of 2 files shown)


lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -69,21 +82,28 @@ def _get_file_path(self, key: CacheEngineKey) -> Path:
...
+            if self.relative_tmp_dir is not None:
+                tmp_path = base_path / self.relative_tmp_dir / key_path
+            else:
+                tmp_path = data_path.with_suffix(".tmp")
+        return data_path, tmp_path
 
     async def exists(self, key: CacheEngineKey) -> bool:
         """Check if key exists in file system"""
-        file_path = self._get_file_path(key)
+        file_path, _ = self._get_file_path(key)
```
maobaolong: This method didn't need the second value of the tuple which returned by `_get_file_path `

lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -69,21 +82,28 @@ def _get_file_path(self, key: CacheEngineKey) -> Path:
...
         return await aiofiles.os.path.exists(file_path)
 
     def exists_sync(self, key: CacheEngineKey) -> bool:
         """Check if key exists in file system synchronized"""
-        file_path = self._get_file_path(key)
+        file_path, _ = self._get_file_path(key)
```
maobaolong: ditto

lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -69,21 +82,28 @@ def _get_file_path(self, key: CacheEngineKey) -> Path:
...
         return os.path.exists(file_path)
 
     async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
         """Get data from file system"""
-        file_path = self._get_file_path(key)
+        file_path, _ = self._get_file_path(key)
```
maobaolong: ditto

maobaolong (changes requested):

maobaolong:
@chunxiaozheng Thanks for this PR, left a minor improvement suggestion. Otherwise, it looks good. It would be nice if you add some motivation why add this `fs_connector_relative_tmp_dir` extra_config?

maobaolong (approved):
@chunxiaozheng Thanks for the quick updated, it LGMT! @sammshen Would you like to take a double check?

lmcache/v1/storage_backend/connector/fs_connector.py
chunxiaozheng: @maobaolong Thanks for your review! I hava updated!

sammshen (approved):
LGTM!

PR #1609, opened 2025-09-16
[Benchmark] configure user stacking at high qps in multi-round-qa

yuezhu1:
Fix issue on user sessions stacking in LMCache benchmark [multi_round_qa.py](https://github.com/LMCache/LMCache/blob/dev/benchmarks/multi_round_qa/multi-round-qa.py) at high QPS. 

In [_remove_finished_sessions()](https://github.com/LMCache/LMCache/blob/0f9ae55627a9e7273b49f2f911b7064dc7449a55/benchmarks/multi_round_qa/multi-round-qa.py#L411), we only remove sessions which are marked as finished in user session at [here](https://github.com/LMCache/LMCache/blob/cad4db10592e8487d1e0bbd02d1a6f92c37d0581/benchmarks/multi_round_qa/multi-round-qa.py#L310). This means that the user session is removed only when all pending requests have been finished and we have reached configured number of round. But in UserSessionManager.step() at [here](https://github.com/LMCache/LMCache/blob/cad4db10592e8487d1e0bbd02d1a6f92c37d0581/benchmarks/multi_round_qa/multi-round-qa.py#L429), a new user is added when the time interval is greater than the pre-calculated gap between users. If we have high QPS (like QPS=20) with active user count being set as 15, there are cases that the number of active user counts during the test is greater than 15, as new users keep joining but old users are still around due to outstanding requests. 

This PR fix this to maintain fixed number of active users by only adding new user sessions when the active user count is less than our defined number.

FIX #xxxx (*link existing issues this PR will resolve*)

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

</details>

### DIFF
--- a/benchmarks/multi_round_qa/multi-round-qa.py
+++ b/benchmarks/multi_round_qa/multi-round-qa.py
@@ -42,6 +42,9 @@ class WorkloadConfig:
     # Whether to include user id in request header
     enable_user_id: bool
 
+    # Whether strictly cap active sessions at num_users
+    enforce_strict_concurrent_users: bool = False
+
 
 @dataclass
 class UserConfig:
@@ -374,6 +377,10 @@ class UserSessionManager:
         if self.use_sharegpt:
             self._load_sharegpt_data()
 
+        self.enforce_strict_concurrent_users = (
+            workload_config.enforce_strict_concurrent_users
+        )
+
     def _load_sharegpt_data(self):
         with open("ShareGPT.json", "r", encoding="utf-8") as file:
             self.sharegpt_data = json.load(file)
@@ -419,6 +426,19 @@ class UserSessionManager:
                 self.session_summaries.append(session.summary())
         self.sessions = [s for s in self.sessions if not s.finished]
 
+    def _can_join_user(self, timestamp: float) -> bool:
+        # No new user session if gap_between_users time interval not meets
+        if timestamp - self.last_user_join <= self.gap_between_users:
+            return False
+
+        # No user seession if active user count is less than configured
+        if (
+            self.enforce_strict_concurrent_users
+            and len(self.sessions) >= self.workload_config.num_users
+        ):
+            return False
+        return True
+
     def step(self, timestamp: float, executor: RequestExecutor):
         if self.need_ramp_up:
             self._ramp_up(timestamp, self.ramp_up_time)
@@ -426,7 +446,8 @@ class UserSessionManager:
         if self.start_time is None:
             self.start_time = timestamp
 
-        if timestamp - self.last_user_join > self.gap_between_users:
+        # Check if can join new user session
+        if self._can_join_user(timestamp):
             self._create_user_session()
             self.last_user_join = timestamp
             logger.info(
@@ -635,6 +656,11 @@ def parse_arguments():
         action="store_true",
         help="Does not send requests to the endpoint (server)",
     )
+    parser.add_argument(
+        "--enforce-strict-concurrent-users",
+        action="store_true",
+        help="Strictly enforce concurrent users count to match --num-users",
+    )
     args = parser.parse_args()
     return args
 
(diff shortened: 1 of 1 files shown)


ApostaC:
@Shaoting-Feng @sammshen Since we use this benchmark a lot, can you help review this PR? Thanks!

sammshen:
Please fix the line too long :)
sorry for the trouble but please run the pre-commit
https://docs.lmcache.ai/developer_guide/contributing.html

as mentioned on slack, I think this (semantic not correctness) change makes sense since the risk of hurting the real QPS will not happen when the current users are struggling to finish their requests

sammshen (approved):

benchmarks/multi_round_qa/multi-round-qa.py
```diff
@@ -426,7 +426,8 @@ def step(self, timestamp: float, executor: RequestExecutor):
         if self.start_time is None:
             self.start_time = timestamp
 
-        if timestamp - self.last_user_join > self.gap_between_users:
+        # New user session only joins when meets time interval and active user count is less than configured
+        if timestamp - self.last_user_join > self.gap_between_users and len(self.sessions) < self.workload_config.num_users:
```
Shaoting-Feng: In my opinion, we should not add this additional condition. We use this script to benchmark different engines. If an engine performs poorly and requests finish slowly, adding this condition will cause the QPS to decrease, making the workload lighter than for an engine that performs well. This would result in an unfair comparison.

Shaoting-Feng (changes requested):

benchmarks/multi_round_qa/multi-round-qa.py
yuezhu1: without this condition, we would have more active user count than we configured in the workload.config. This differs the meaning of keeping fixed amount of concurrent users, as active user count could keep growing. I wonder if there is any reason behind this design?

sammshen: one reason could be that we are simulating users joining a serving engine and back-pressuring. it is a question of do we want to strictly enforce QPS (to the extent possible) or strictly enforce concurrent users. Both can make sense

ApostaC: How about we introduce a new cmdline arg (something like `--enforce-strict-qps`) to switch between two modes?

yuezhu1: sure. it is always good to give people options. my concern is that `--enforce-strict-qps` may not force the qps in real practice if we have high QPS even with high GPU count. I am testing on 16GPUs test env, but the users keep stacking when set QPS=14. So if we set `--enforce-strict-qps`, how do we actually enforce this? We may need to give warning message somewhere. maybe in the [data_preprocessing.py](https://github.com/LMCache/LMCache/blob/3439ab3c65fc2ce696f5bf711fc2e752ab0c7922/benchmarks/multi_round_qa/data_preprocessing.py), we can indicate that the target QPS is not reached?

sammshen: good point, let's use `--enforce-strict-concurrent-users` instead since like you said QPS is technically impossible to enforce under the assumption that users have to receive responses to send the next one. This is an example strict QPS enforcer but it breaks the assumpiton of receiving a response before sending a new request:
https://github.com/LMCache/LMBench/blob/main/3-workloads/strict-synthetic/strict-multi-round-qa.py

yuezhu1: For execution plan on `--enforce-strict-concurrent-users`, I would like to confirm if we want to keep the assumption that we receive a response before sending a new request. If yes,  the only additional change for me is to add `--enforce-strict-concurrent-users` option. If not, we have to tweak the request submission code. My preference is to add  `--enforce-strict-concurrent-users while keeping the assumption. If this is ok, i will push the changes.

sammshen: yes we should keep that assumption. the `--enforce-strict-concurrent-users` flag will suffice :)

yuezhu1:
added `--enforce-strict-concurrent-users` option as discussed above. @sammshen @Shaoting-Feng

Shaoting-Feng (approved):
LGTM

PR #1615, opened 2025-09-17
[Improve] Add timeout for lookup client 

maobaolong:
# issue

In some case, we make the socket path mismatch between `scheduler` process which act as `lmcache lookup client` and `worker` which act as `lmcache lookup server`.

But we cannot aware of this issue from the log, we found this root cause after several hours.

# Proposal

This PR aim to log the warn and error log to let user aware the connection error by a timeout and error handling. Eventually, the lookup function return 0 if exception occured.

# Log example

```
[2025-09-17 11:53:58,687] LMCache WARNING: Receive timeout (retry 1/3) for rank 0 (lmcache_lookup_client.py:181:lmcache.v1.lookup_client.lmcache_lookup_client)
[2025-09-17 11:54:03,692] LMCache WARNING: Receive timeout (retry 2/3) for rank 0 (lmcache_lookup_client.py:181:lmcache.v1.lookup_client.lmcache_lookup_client)
[2025-09-17 11:54:08,698] LMCache WARNING: Receive timeout (retry 3/3) for rank 0 (lmcache_lookup_client.py:181:lmcache.v1.lookup_client.lmcache_lookup_client)
[2025-09-17 11:54:08,698] LMCache ERROR: Failed to receive from rank 0 after 3 retries (lmcache_lookup_client.py:188:lmcache.v1.lookup_client.lmcache_lookup_client)
```

### DIFF
--- a/lmcache/v1/lookup_client/lmcache_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_lookup_client.py
@@ -56,6 +56,10 @@ class LMCacheLookupClient(LookupClientInterface):
         self.sockets = []
         if self.create_lookup_server_only_on_worker_0_for_mla:
             ranks = 1
+
+        # Set timeout values from config
+        timeout_ms = config.lookup_timeout_ms
+
         for tp_rank in range(ranks):
             socket_path = get_zmq_rpc_path_lmcache(
                 vllm_config, "lookup", rpc_port, tp_rank
@@ -70,6 +74,10 @@ class LMCacheLookupClient(LookupClientInterface):
                 zmq.REQ,  # type: ignore[attr-defined]
                 bind=False,
             )
+
+            # Set socket timeout during initialization
+            socket.setsockopt(zmq.RCVTIMEO, timeout_ms)
+            socket.setsockopt(zmq.SNDTIMEO, timeout_ms)
 
             self.sockets.append(socket)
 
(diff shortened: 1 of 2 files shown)


maobaolong:

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -128,16 +129,71 @@ def lookup(
...
-            result = int.from_bytes(resp, "big")
-            results.append(result)
+            socket = self.sockets[i]
+
+            # Check socket state
+            if socket.closed:
+                logger.error(f"Socket for rank {i} is closed")
+                return 0
 
+            # Set socket timeout
```
sammshen: why don't we set socket timeouts once at initiatialization?

sammshen (approved):
LGTM!

Why will the socket connection fail? Furthermore, why might the socket fail and THEN succeed after a few retries? If we do multiple retries this might hurt performance so maybe we should just return 0 (so we don't crash the inference engien) with a warning saying the socket broke? I think this is a safer semantics.

chunxiaozheng:
> LGTM!
> 
> Why will the socket connection fail? Furthermore, why might the socket fail and THEN succeed after a few retries? If we do multiple retries this might hurt performance so maybe we should just return 0 (so we don't crash the inference engien) with a warning saying the socket broke? I think this is a safer semantics.

@sammshen now, the socket path use `engine_id` from `vllm_config.kv_transfer_config`, our internal vllm version is old, when use `MultiConnector`, the `engine_id` between `lookup_client` and `lookup_server` is different, which will result in connection failed.

vllm has fixed the `engine_id` in `MultiConnector`(https://github.com/vllm-project/vllm/pull/19487)

sammshen:
but how will retry help? if the engine id is wrong, the first failure is enough. we should just say the lookup is 0 so we don't affect the performance or correctness

lmcache/v1/lookup_client/lmcache_lookup_client.py
maobaolong: Yeah, it will be better if we set the socket timeout at initiatialization

maobaolong:
@sammshen Thanks for your review, the suggestions from your side are all addressed, removed the useless retry logic, and moved the timeout set code to the initialization of the `LMCacheLookupClient`. PTAL.

sammshen:
LGTM! thx for the change

chunxiaozheng (approved):
lgtm

PR #1584, opened 2025-09-13
[Bugfix] Fix Python 3.9 compatibility by replacing match statements

NaikPrats:
Replace match-case statements with if-elif chains in config.py to support Python 3.9, which doesn't have pattern matching syntax.

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient unit tests to ensure the change is stay correct and robust. The unit and integration tests will always run and our comprehensive test will be triggered after the "full" label is tagged onto a PR.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

We aim to address all PRs in a timely manner. If no one reviews your PR within 5 days, please @-mention one of KuntaiDu, ApostaC or YaoJiayi.

</details>

PR #1618, opened 2025-09-17
[Chore]: Remove Python 3.9 as the minimum Python version

hickeyma:
LMCache was built on Python 3.10 and as such has keywords like match which are not compatible in Python 3.9. We therefore would need to make changes to the code base. Ref: https://github.com/LMCache/LMCache/pull/1584#pullrequestreview-3224555325

Python 3.9 will no longer be supported next month (https://en.wikipedia.org/wiki/History_of_Python#Support). Therefore, it is more practical to support 3.10 as the minimum version.

This is a revert of some of the changes in:
https://github.com/LMCache/LMCache/pull/1440

### DIFF
--- a/docs/source/developer_guide/contributing.rst
+++ b/docs/source/developer_guide/contributing.rst
@@ -90,7 +90,7 @@ The following prerequisites are required:
 The following tools are required:
 
 - `git <https://git-scm.com>`_
-- `python <https://www.python.org>`_ (v3.9 -- v3.13)
+- `python <https://www.python.org>`_ (v3.10 -- v3.13)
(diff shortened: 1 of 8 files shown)


sammshen (approved):
LGTM, thanks for digging up the deprecation date of 3.9! Are the strict=False's from a pre-commit or personal preference?

hickeyma:
> LGTM, thanks for digging up the deprecation date of 3.9! Are the strict=False's from a pre-commit or personal preference?

From tools in pre-commit

YaoJiayi (approved):
lgtm

PR #1619, opened 2025-09-17
Fix two bugs in async loading with MLA and TP > 1

yoo-kumaneko:
When we use MLA, we likely have lookup server only on worker process 0.
This causes two issues:
1. Other workers have self.lookup_server = None, and hence fails the assertion.
2. len(all_res) is only 1, and it will never enter the if branch. As a result, req_status will never be updated.

I’ve updated the code to properly handle MLA with TP > 1.

### DIFF
--- a/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
@@ -195,7 +195,10 @@ class LMCacheAsyncLookupClient(LookupClientInterface):
                     self.res_for_each_worker[lookup_id].append(res)
                 all_res = self.res_for_each_worker[lookup_id]
 
-                if len(all_res) == self.tensor_parallel_size:
+                if len(all_res) == self.tensor_parallel_size or (
+                    self.create_lookup_server_only_on_worker_0_for_mla
+                    and len(all_res) == 1
+                ):
                     self.res_for_each_worker.pop(lookup_id)
 
                     # NOTE: it is possible that the number of hit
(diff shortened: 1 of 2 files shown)


lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -195,7 +195,10 @@ def process_responses_from_workers(self):
                     self.res_for_each_worker[lookup_id].append(res)
                 all_res = self.res_for_each_worker[lookup_id]
 
-                if len(all_res) == self.tensor_parallel_size:
+                if len(all_res) == self.tensor_parallel_size or (
+                    self.create_lookup_server_only_on_worker_0_for_mla
+                    and len(all_res) == 1
+                ):
```
YaoJiayi, line 201: Do we need `len(all_res) == 1`? Is it possible that `all_res != 1`?

YaoJiayi (commented):
LGTM. Just a minor question.

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: `len(all_res) == 1` only matters if `self.create_lookup_server_only_on_worker_0_for_mla` is on. Note the `or` before them. It doesn't need to be 1 if `create_lookup_server_only_on_worker_0_for_mla` is off.

yoo-kumaneko: I'm not 100% sure if `len(all_res)` is always 1 when `create_lookup_server_only_on_worker_0_for_mla` is enabled. Although, like its name suggests, there's only one lookup server, so it should be one.

YaoJiayi (approved):

PR #1629, opened 2025-09-18
[Core]Heterogenous TP

novahow:

FIX #xxxx (*link existing issues this PR will resolve*)

This is for supporting heterogenous PD disaggregation, any comments are welcomed! 

Key Changes:
1. Identify Decoder TP size and provide in disagg_spec
2. Exchange metadata such as world_size between sender/receiver
3. Shard the shape of memallocrequest along the hidden dim dimension
4. Permute the memory_obj tensor from 2LTD -> TD2L if it is nixl in the reshape kernel, the new page_size for nixl transfer would be D*2*L. the sharding doesn't work if D is not the first dim of the page.  
  a. upd: 2LTD -> DT2L, and we're using the full chunk_size in `memory_obj.tensor`, which is usually 256 if we're doing transpose
5.  Mutate the cacheEngineKey before putting into nixlbackend, since `contains` checks attributes such as `world_size` and `worker_id`. 

TODOs:
[ ] 1. currently it's token-by-token sending and kinda slow, so maybe will consider adding a "block" dimension like vllm
[v] 2. add a flag like `enable_hetero_nixl` to turn on the permutation and modify nixl page size to separate hetero and homogenous tp code-path
[ ]3. MLA/flashinfer not yet handled
[] 4. disabled pin in decoder's pd_backend on `contains` call, since it somehow causes some issues, which needs some investigation. 
  a. upd: tried the dev branch instead of this new branch, tp=[2,2] also doesn't work under `pin=True`
[] 5. depends on https://github.com/LMCache/LMCache/pull/1818

## Benchmarking
```
python benchmark_serving.py --port {port} --seed $(date +%s) \
    --model meta-llama/Llama-3.1-8B-Instruct \
    --dataset-name hf --random-input-len 7500 --random-output-len 200 --dataset-path philschmid/mt-bench \    --num-prompts 30 --burstiness 100 --request-rate 1 --ignore-eos
 ```

#### [1,2]
```
Successful requests:                     30        
Benchmark duration (s):                  33.57     
Total input tokens:                      2873      
Total generated tokens:                  7650      
Request throughput (req/s):              0.89      
Output token throughput (tok/s):         227.87    
Total Token throughput (tok/s):          313.45    
Mean TTFT (ms):                          44.71     
Median TTFT (ms):                        43.68     
P99 TTFT (ms):                           54.44     
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          14.22     
Median TPOT (ms):                        14.25     
P99 TPOT (ms):                           14.29     
---------------Inter-token Latency----------------
Mean ITL (ms):                           14.16     
Median ITL (ms):                         14.09     
P99 ITL (ms):                            16.72     
```

#### [1,1]
```
Successful requests:                     30        
Benchmark duration (s):                  37.03     
Total input tokens:                      2815      
Total generated tokens:                  7650      
Request throughput (req/s):              0.81      
Output token throughput (tok/s):         206.57    
Total Token throughput (tok/s):          282.58    
Mean TTFT (ms):                          1043.76   
Median TTFT (ms):                        1042.97   
P99 TTFT (ms):                           1053.52   
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          23.88     
Median TPOT (ms):                        23.91     
P99 TPOT (ms):                           23.97     
---------------Inter-token Latency----------------
Mean ITL (ms):                           23.79     
Median ITL (ms):                         23.71     
P99 ITL (ms):                            24.96     

```

#### [2,2]
```
Successful requests:                     30        
Benchmark duration (s):                  33.80     
Total input tokens:                      3470      
Total generated tokens:                  7650      
Request throughput (req/s):              0.89      
Output token throughput (tok/s):         226.32    
Total Token throughput (tok/s):          328.98    
Mean TTFT (ms):                          73.96     
Median TTFT (ms):                        72.00     
P99 TTFT (ms):                           110.12    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          15.08     
Median TPOT (ms):                        15.01     
P99 TPOT (ms):                           15.74     
---------------Inter-token Latency----------------
Mean ITL (ms):                           15.02     
Median ITL (ms):                         14.31     
P99 ITL (ms):                            25.70     
```

#### [1,2]
```
Successful requests:                     30        
Benchmark duration (s):                  32.70     
Total input tokens:                      2755      
Total generated tokens:                  7650      
Request throughput (req/s):              0.92      
Output token throughput (tok/s):         233.92    
Total Token throughput (tok/s):          318.17    
Mean TTFT (ms):                          34.66     
Median TTFT (ms):                        32.27     
P99 TTFT (ms):                           51.47     
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          10.66     
Median TPOT (ms):                        10.66     
P99 TPOT (ms):                           10.79     
---------------Inter-token Latency----------------
Mean ITL (ms):                           10.64     
Median ITL (ms):                         10.53     
P99 ITL (ms):                            13.53     
```
#### [1,1]
```
Successful requests:                     30        
Benchmark duration (s):                  32.17     
Total input tokens:                      3164      
Total generated tokens:                  7650      
Request throughput (req/s):              0.93      
Output token throughput (tok/s):         237.80    
Total Token throughput (tok/s):          336.16    
Mean TTFT (ms):                          32.25     
Median TTFT (ms):                        30.57     
P99 TTFT (ms):                           39.74     
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          8.47      
Median TPOT (ms):                        8.48      
P99 TPOT (ms):                           8.51      
---------------Inter-token Latency----------------
Mean ITL (ms):                           8.43      
Median ITL (ms):                         8.38      
P99 ITL (ms):                            9.51      
```

#### [2,2]
```
Successful requests:                     30        
Benchmark duration (s):                  32.98     
Total input tokens:                      3345      
Total generated tokens:                  7650      
Request throughput (req/s):              0.91      
Output token throughput (tok/s):         231.93    
Total Token throughput (tok/s):          333.34    
Mean TTFT (ms):                          87.54     
Median TTFT (ms):                        83.92     
P99 TTFT (ms):                           129.05    
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          13.97     
Median TPOT (ms):                        13.32     
P99 TPOT (ms):                           18.38     
---------------Inter-token Latency----------------
Mean ITL (ms):                           13.91     
Median ITL (ms):                         10.92     
P99 ITL (ms):                            32.95     
```

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

</details>

### CODE REVIEW

**Comment by novahow** [comment]
Hi @sammshen 
Would you be able to take a quick look on this?
Before migrating my solution to the new PD code architecture, I would appreciate your feedback on my high-level design idea to ensure I’m on the right track. Thanks!

**Comment by YaoJiayi** [comment]
@novahow Thanks for the huge effort! The description in "Key Changes" makes sense to me. Looking forward to the PR (and some benchmarking numbers)!

**Comment by novahow** [comment]
Hi @YaoJiayi Besides code migration, I did some benchmarking on both L40S and H100. PTAL when you're available, thanks!

**Comment by YaoJiayi** [comment]
@novahow We will try to push the code in asap. This is an important feature:)

**Comment by sammshen** [comment]
@novahow @YaoJiayi will do a first pass now, so sorry for the delay

File: csrc/mem_kernels.cu [hunk 1]
```diff
@@ -162,10 +162,16 @@ __device__ __forceinline__ int64_t page_buffer_offset_unilateral(
   return token_idx * scalars_per_token + scalar_offset;
 }
 
+template <bool transpose = false>
 __device__ __forceinline__ int64_t
 key_value_offset(const int k_or_v, const int layer_idx, const int token_idx,
                  const int scalar_offset, const int scalars_per_token,
-                 const int num_tokens, const int num_layers) {
+                 const int num_tokens, const int num_layers, const int num_kv) {
+  if (transpose) {
```
sammshen, line 169: nit: reorder like: 

```cpp
  if (transpose) {
    // [2LTD->DT2L], T is full chunk size
    return scalar_offset * num_tokens * num_kv * num_layers  +
           token_idx * num_kv * num_layers + k_or_v * num_layers + layer_idx;
  }
```

csrc/mem_kernels.cu
```diff
@@ -202,9 +208,9 @@ __global__ void load_and_reshape_multi_layer_kernel(
 
   /** Copy the data from page buffer to key_value **/
   for (int i = tid; i < scalars_per_token; i += num_threads) {
-    const int64_t lmcache_offset =
-        key_value_offset(k_or_v, layer_id, token_id, i, scalars_per_token,
-                         num_tokens, num_layers);
+    const int64_t lmcache_offset = key_value_offset<transpose>(
```
sammshen: can we grab the gridDim.z and name it num_kv in the beginning of the func

csrc/mem_kernels.cu
```diff
@@ -247,7 +253,7 @@ __global__ void load_and_reshape_multi_layer_kernel_unilateral(
   for (int i = tid; i < scalars_per_token; i += num_threads) {
     const int64_t lmcache_offset =
         key_value_offset(k_or_v, layer_id, token_id, i, scalars_per_token,
-                         num_tokens, num_layers);
+                         num_tokens, num_layers, gridDim.z);
```
sammshen: same here

csrc/mem_kernels.cu
```diff
@@ -341,24 +347,42 @@ void multi_layer_kv_transfer(
...
   const at::cuda::OptionalCUDAGuard device_guard(paged_memory_device);
   const cudaStream_t stream = at::cuda::getCurrentCUDAStream();
 
-  if (not direction) {
-    lmc::load_and_reshape_multi_layer_kernel<int64_t, false>
-        <<<grid, block, 0, stream>>>(key_value_ptr, page_buffer_ptrs,
-                                     slot_mapping_ptr, num_qwords, num_tokens,
-                                     num_layers, page_buffer_size);
-    C10_CUDA_KERNEL_LAUNCH_CHECK();
+  if (transpose) {
```
sammshen: sorry if I am missing something here but can we not just pass in tranpose as teh third template argument? :)

sammshen: maybe we can put the conditional on top with: 
```cpp
if (transpose) 
    num_tokens = (int)key_value.size(1)
else
    num_tokens = (int)key_value.size(2)
assert(num_tokens==slot_mapping.size(0))
```

sammshen:
The examples and benchmarks are awesome! Thanks @novahow

csrc/mem_kernels.cu
novahow: I'm addressing the rest, but here the real `num_tokens` would be `slot_mapping.size(0)`. Since I enforced the `memory_obj` to use the full chunk size. And the format here would still be KV_2LTD, but T is full chunk_size now if it is transpose. That's why `key_value.size(2)` isn't necessarily same as `slot_mapping.size(0)`.  Let me know whether we should introduce a new fmt called `KV_DT2L` so that it might be less misleading about the layout of the `memory_obj.tensor`. 

Let's say chunk_size is 4, but only 3 tokens, the `memory_obj`'s raw_data buffer would look like below. It won't be 2LTD in the memory_obj when copying `from_gpu` to buffer, but it would transpose back when calling a `to_gpu` 

image

examples/disagg_prefill/disagg_proxy_server.py
```diff
@@ -273,7 +272,7 @@ async def stream_service_response(
             yield chunk
 
 
-def round_robin_pick_client(clients, idx):
+def round_robin_pick_client(clients, idx) -> ClientInfo:
```
sammshen, line 283: nice type hint

examples/disagg_prefill/disagg_proxy_server.py
```diff
@@ -494,6 +498,10 @@ async def generate_stream():
                 "data: " + json.dumps(head_chunk, separators=(",", ":")) + "\n\n"
             ).encode()
 
+            # TODO(novahow): this was originally num_tp_rank
```
sammshen: CRITICAL: the `num_tp_rank` definitely needs to be prefill side here (what you did above). the previous assumption was implicit that decode and prefill tp are the same

examples/disagg_prefill/hetero_tp/configs/lmcache-prefiller-config.yaml
```diff
@@ -0,0 +1,14 @@
+local_cpu: False
```
sammshen, line 1: sorry for the nit, but can we remove all the extraneous files inside of these new examples? like the decoder-2 and renaming the files to just prefiller and decoder and the launcher scripts. :)

examples/disagg_prefill/xpyd_tp2/README.md
```diff
@@ -0,0 +1,82 @@
...
+```bash
+python benchmark_serving.py --port 9100 --seed $(date +%s) \
+    --model meta-llama/Llama-3.1-8B-Instruct \
+    --dataset-name random --random-input-len 7500 --random-output-len 200 \
+    --num-prompts 30 --burstiness 100 --request-rate 1 --ignore-eos
+```
+
+Expected output from the benchmark script:
+
+```plaintext
```
sammshen: also can put your actual numbers in the READMEs? thanks!

examples/disagg_prefill/disagg_proxy_server.py
```diff
@@ -308,13 +306,15 @@ async def handle_completions(request: Request):
         # Pick decode client
         decode_client = round_robin_pick_client(app.state.decode_clients, counter)
 
+        num_tp_rank = len(decode_client.init_port or [])
```
sammshen: two more nits: 
let's initialize `decoder_tp_rank` and `prefiller_tp_rank` up front as two variables here. btw we should assert that the init_port is not None since 0 (`len([])`) should not be valid.

Also, this is good because we should be able to have like multiple prefillers of different TPs (weird but possible) and same for decoders

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -147,10 +148,8 @@ def __init__(
         self.transfer_channel = CreateTransferChannel(
             channel_type=config.transfer_channel,
             async_mode=True,
-            role="both",
-            buffer_ptr=self.memory_allocator.cpu_allocator.buffer_ptr,
-            buffer_size=self.memory_allocator.cpu_allocator.buffer_size,
-            align_bytes=self.memory_allocator.cpu_allocator.align_bytes,
+            role=PDRole.BOTH,
```
sammshen: these changes are really clean!

csrc/mem_kernels.cu
sammshen: a KV_DT2L is a good idea!

novahow: Hi, I introduced a new `DT2L` format and made it work, but during the process I found it's hard to integrate the format cleanly as the default "KV_2LTD" format is hardcoded or scattered across the codebase, introducing this format not only involves altering the format but also permuting the shape, and `2LTD` it's also the default shape of `VLLMPagedMemGPUConnectorV2`.  So from my point of view maybe using `2LTD` and the `transpose` attribute of `memory_obj` together fits more under the current scheme and brings less complexity. let me know what do you think, thanks

examples/disagg_prefill/disagg_proxy_server.py
novahow: Sure, should we introduce a new argument called `prefiller-init-port` or like a `prefiller-tp-size`, so that we could get the real tp size of the prefiller? My current implementation uses `global_args.prefiller_port` which i think is not correct.

vladnosiv:
Hi ! It looks impressive.
Can you tell me if this will work if PP is enabled on the decode host? If such tests are not currently in the plans, then no problem.

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -362,67 +419,137 @@ def _get_remote_alloc_request(
...
+        transfer_spec: Any = None,
+    ) -> None:
+        for mem_obj in memory_objs:
+            mem_obj.ref_count_up()
+
+        # Third Party
+        from vllm.distributed.utils import divide
+
+        # TODO(novahow): is there better way to obtain receiver tp_size?
+        self.dp_ratio = divide(
```
sammshen: (random suggestion) in a future PR, we can maybe also allow this to be fractions like 1/2^n so that we can do prefill TP > decode TP!

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -362,67 +419,137 @@ def _get_remote_alloc_request(
...
+            len(transfer_spec.receiver_init_port), self.tp_world_size
+        )
+        # NOTE(novahow), assume dp_ratio = 2
+        # rank 0 on P maps to rank [0,1] on D
+        receiver_init_port = transfer_spec.receiver_init_port[
```
sammshen: (random suggesiton) or we can just do extra branch logic since we know each prefill worker only have to do one set of mem alloc requests

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -464,7 +591,10 @@ def _allocate_and_put(self, alloc_request: AllocRequest) -> AllocResponse:
 
         for idx, key_str in enumerate(alloc_request.keys):
             key = CacheEngineKey.from_string(key_str)
```
sammshen, line 573: (random suggestion for future PR), for prefill TP > decode TP, maybe we can add smth like: 
```python
merge_ratio = prefill_tp // decode_tp
rank_within_decode = self.tp_rank % merge_ratio

for idx, key_str in enumerate(alloc_request.keys):
    ...
    mem_obj = self.allocate(torch.Size(shape), dtype, fmt)

    # Compute per-prefill offset into this decode buffer
    base_addr = mem_obj.meta.address
    offset = rank_within_decode * shape[fmt.token_dim()] * bytes_per_token
    alloc_indexes.append(base_addr + offset)

    self.put(key, mem_obj)
    
    ```

File: csrc/mem_kernels.cu (hunk 4 above)
Comment (sammshen) [reply csrc/mem_kernels.cu]: makes sense, I agree since we can just hide this in the GPU connector. sorry for the bad suggestion

**Comment by sammshen** [comment]
> Hi ! It looks impressive. Can you tell me if this will work if PP is enabled on the decode host? If such tests are not currently in the plans, then no problem.

Hi @vladnosiv! do you mind explaining what is the benefit of having PP on the decoder but not the prefiller? 

I think it is not supported yet but @novahow correct me if I'm wrong but in the future we can also transpose like :
2LTD -> LDT2

Then we can shard along L and also along D secondarily if PP and TP are both active.

**Comment by vladnosiv** [comment]
> Hi @vladnosiv! do you mind explaining what is the benefit of having PP on the decoder but not the prefiller?

A rather specific case, perhaps, but if we consider the Qwen3-235B-A22B-FP8 model and the H100 card, then the following situation occurs at the decode stage:

TP4 potentially works better for attention than TP8 (assumption), while setting TP4 on the decode stops being memory-bandwith-bound, and becomes memory-capacity-bound. In this case, it makes sense to make TP4 PP2, then there will be more space for kv and we will be able to become memory-bandwith-bound again (with large batches).

On the other hand, we won't have such a problem on Prefill, because there's no need for large batches.

This is not a very critical example, because the alternative here is TP8, but on larger models something similar may occur, where TP greater than 8 will lead to high communication overhead.

File: lmcache/v1/storage_backend/pd_backend.py (hunk 12 above)
Comment (novahow) [reply lmcache/v1/storage_backend/pd_backend.py]: Hi, sorry I didn't get what this means, would you mind elaborating more on this? thanks!

**Comment by novahow** [comment]
> > Hi ! It looks impressive. Can you tell me if this will work if PP is enabled on the decode host? If such tests are not currently in the plans, then no problem.
> 
> Hi @vladnosiv! do you mind explaining what is the benefit of having PP on the decoder but not the prefiller?
> 
> I think it is not supported yet but @novahow correct me if I'm wrong but in the future we can also transpose like : 2LTD -> LDT2
> 
> Then we can shard along L and also along D secondarily if PP and TP are both active.

@vladnosiv That's an interesting use case! I believe this is not supported yet. But in the future PR, I believe as @sammshen mentioned, we could also introduce new transpose scheme to make it compatible with PP+TP by LDT2. Probably some work on `disagg_spec` would need to be accomplished as well

File: lmcache/v1/storage_backend/pd_backend.py (hunk 13 above)
Comment (novahow) [reply lmcache/v1/storage_backend/pd_backend.py:573]: Hi, would you mind explaining the possible performance gain with P TP > D TP? I guess we're doing D TP > P TP because decode is memory-bound, so it could benefit from higher TP. I've found some discussions on this:
https://github.com/vllm-project/vllm/pull/18833#discussion_r2127375752

If there’s a concrete use case for this configuration, we could definitely consider supporting it in the future.

File: csrc/mem_kernels.cu (hunk 1 above)
Comment (YaoJiayi) [reply csrc/mem_kernels.cu:169]: Are we assuming full chunk size here?

Comment (YaoJiayi) [reply csrc/mem_kernels.cu:169]: @novahow

Comment (YaoJiayi) [reply csrc/mem_kernels.cu:169]: If so, why are we making this assumption?

File: csrc/mem_kernels.cu [hunk 14]
```diff
@@ -234,6 +240,7 @@ __global__ void load_and_reshape_multi_layer_kernel_unilateral(
   const int k_or_v = blockIdx.z;
   const int tid = threadIdx.x;
   const int num_threads = blockDim.x;
+  const int num_kv = gridDim.z;
```
YaoJiayi: Would you mind using `k_or_v_size` to align with the variable name in the launcher?

examples/disagg_prefill/hetero_tp/configs/lmcache-prefiller-config.yaml
YaoJiayi: @novahow Also, we do we create two folders in the examples? One is hetero_tp and the other one is xpyd_tp2? Can we use one?

examples/disagg_prefill/disagg_proxy_server.py
```diff
@@ -187,11 +196,13 @@ def parse_args():
     parser.add_argument("--host", type=str, default="localhost")
     parser.add_argument("--prefiller-host", type=csv_strs, default=["localhost"])
     parser.add_argument("--prefiller-port", type=csv_ints, default=[8100])
+    parser.add_argument("--prefiller-init-port", type=csv_ints, default=[8200])
+    parser.add_argument("--prefiller-alloc-port", type=csv_ints, default=[8300])
     parser.add_argument("--num-prefillers", type=int, default=1)
```
YaoJiayi, line 193: What are `profiler-init-port` and `profiler-alloc-port`?

examples/disagg_prefill/disagg_proxy_server.py
```diff
@@ -77,12 +77,21 @@ def pair_hosts_and_ports(hosts, ports, count=None):
...
     )
-    for host, port in prefill_pairs:
+
+    incremental_mode = (
+        len(pref_hosts) == 1 and len(pref_ports) == 1 and global_args.num_prefillers > 1
+    )
+    for i, (host, port) in enumerate(prefill_pairs):
         prefiller_base_url = f"http://{host}:{int(port)}"
         prefill_client = httpx.AsyncClient(timeout=None, base_url=prefiller_base_url)
+        if incremental_mode:
```
YaoJiayi: Why do me need this incremental mode?

YaoJiayi: For now, I think we should let users explicitly specify the correct ports. In the future, we will have a component in lmcache to handle all kind of port initializations

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -85,10 +85,11 @@ class SaveSpec:
 @dataclass
 class DisaggSpec:
     req_id: str
-    receiver_id: str
+    receiver_id: list[str]
     receiver_host: str
-    receiver_init_port: int
-    receiver_alloc_port: int
+    receiver_init_port: list[int]
+    receiver_alloc_port: list[int]
```
YaoJiayi: Make these variables plural

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -85,10 +85,11 @@ class SaveSpec:
...
+    tp_size: int
```
YaoJiayi: Why do we still need tp_size if we already have the length of the above three variables?

lmcache/v1/config.py
```diff
@@ -216,6 +216,11 @@ def _to_bool(
         "default": False,
         "env_converter": _to_bool,
     },
+    "enable_asym_tp": {
+        "type": bool,
+        "default": False,
+        "env_converter": _to_bool,
+    },
```
YaoJiayi: Why do we need a special init variable here?

lmcache/utils.py
```diff
@@ -243,6 +243,12 @@ def from_dict(d):
             request_configs=request_configs,
         )
 
+    def update_rank_info_from_nixl(
```
YaoJiayi: use name `from_pd` instead of `from_nixl`?

lmcache/v1/memory_management.py
```diff
@@ -33,22 +33,22 @@
 
 class MemoryFormat(Enum):
     UNDEFINED = 0
-    """[2, num_layers, num_tokens, hidden_dim]
-    """
+
     # KV_BLOB = 1
     KV_2LTD = auto()
-    """[num_tokens, 2, hidden_dim]
+    """[2, num_layers, num_tokens, hidden_dim]
```
YaoJiayi, line 39: Why changing these comments?

lmcache/v1/memory_management.py
```diff
@@ -334,13 +334,15 @@ def __init__(
         raw_data: torch.Tensor,
         metadata: MemoryObjMetadata,
         parent_allocator: Optional["MemoryAllocatorInterface"],
+        transpose: Optional[bool] = False,
```
YaoJiayi, line 357: Instead of adding a `transpose` argument here, cn we simply add another memory format?

lmcache/v1/storage_backend/__init__.py
```diff
@@ -115,12 +115,13 @@ def CreateStorageBackends(
...
         from lmcache.v1.storage_backend.pd_backend import PDBackend
 
-        storage_backends["PDBackend"] = PDBackend(config, metadata)
+        storage_backends["PDBackend"] = PDBackend(config, metadata, loop)
 
     # TODO(Jiayi): The hierarchy is fixed for now
     # NOTE(Jiayi): The local_cpu backend is always created because
     # other backends might need it as a buffer.
-    if not config.enable_pd or config.local_cpu:
+    nixl_skip_cpu_backend: bool = config.enable_pd and not config.local_cpu
```
YaoJiayi: change variable name from nixl to pd

lmcache/v1/storage_backend/__init__.py
```diff
@@ -168,7 +173,7 @@ def CreateStorageBackends(
     if config.gds_path is not None:
         gds_backend = GdsBackend(config, metadata, loop, dst_device)
         storage_backends[str(gds_backend)] = gds_backend
-    if config.remote_url is not None:
+    if config.remote_url is not None and not nixl_skip_cpu_backend:
         remote_backend = RemoteBackend(
```
YaoJiayi, line 195: Add some assertion or warning

lmcache/v1/storage_backend/__init__.py
```diff
@@ -179,8 +184,8 @@ def CreateStorageBackends(
         backend_name = str(remote_backend)
         storage_backends[backend_name] = remote_backend
 
-    if not config.enable_pd or config.local_cpu:
-        # Create dynamic backends from configuration
+    # Create dynamic backends from configuration
+    if not nixl_skip_cpu_backend:
```
YaoJiayi: Add some assertion or warning.

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -66,9 +68,44 @@ class ProxyNotif(PDMsgBase):
...
 
+@dataclass
+class TPRankRecvInfo:
+    group_tp_rank: int
+    receiver_id: str
+    receiver_init_url: str
+    receiver_mem_alloc_url: str
+
+
+class NixlReceiverInfo:
```
YaoJiayi: change var name from Nixl to PD

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -179,28 +228,29 @@ def __init__(
...
             self.mem_alloc_sockets: dict[str, zmq.Socket] = {}
-        elif self.pd_config.role == "receiver":
+        elif self.pd_config.role == PDRole.RECEIVER:
             self._init_receiver()
         else:
-            raise ValueError("Invalid PD role.")
+            raise ValueError(f"Invalid PD role: {self.pd_config.role}")
 
         self.full_chunk_size = config.chunk_size
+        self.dp_ratio = 1
```
YaoJiayi: What does dp_ratio mean?

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -362,67 +419,137 @@ def _get_remote_alloc_request(
...
-        receiver_id = transfer_spec.receiver_host + str(receiver_init_port)
-        receiver_host = transfer_spec.receiver_host
-
-        self._ensure_peer_connection(
-            receiver_id=receiver_id,
-            receiver_host=receiver_host,
-            receiver_init_port=receiver_init_port,
-            receiver_alloc_port=receiver_alloc_port,
-        )
+        executor: cf.ThreadPoolExecutor,
```
YaoJiayi: I suggest we make the code synchronous for now (w.o. using thread pool executor). Two reasons. First, we've been experiencing some crashes (on some machines) when running nixl in another thread. Second, nixl already uses separate threads in itself. We don't have to use another thread here.

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -362,67 +419,137 @@ def _get_remote_alloc_request(
...
+        self._ensure_peer_connection(
+            receiver_info,
+        )
+
+        # Allocate remote memory objects
+        alloc_request = self._get_remote_alloc_request(keys, memory_objs)
+        all_sent_indexes: list[set[int]] = [set() for _ in range(self.dp_ratio)]
+        any_remote_indexes: set[int] = set()
+
+        with cf.ThreadPoolExecutor(self.dp_ratio) as executor:
```
YaoJiayi: We have async primitives in transfer_channel: https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/transfer_channel/nixl_channel.py#L464

YaoJiayi: Probably no need to do in this PR. But we can have a clean version first (w.o. thread pool executor) and add asynchronous support in the future

lmcache/v1/transfer_channel/__init__.py
```diff
@@ -1,24 +1,53 @@
...
+    channel_type: str,
+    async_mode: bool,
+    role: PDRole,
+    allocator_meta: PagedTensorMemoryMetadata,
+    tp_rank: int,
+    peer_init_url: str,
+    tp_size: Optional[int] = None,
+    **kwargs,
+) -> BaseTransferChannel: ...
+
```
YaoJiayi, line 37: We do we have these two functions?

lmcache/v1/transfer_channel/nixl_channel.py
```diff
@@ -556,10 +573,37 @@ def close(self):
             thread.join()
         self.zmq_context.term()
         self.agent.deregister_memory(self.reg_descs)
-        self.agent.release_dlist_handle(self.xfer_handler)
+        for xfer_handler in self.nixl_wrapper.xfer_handlers.values():
+            self.agent.release_dlist_handle(xfer_handler)
         for remote_xfer_handler in self.remote_xfer_handlers_dict.values():
             self.agent.release_dlist_handle(remote_xfer_handler)
 
+    def get_block_descs(
```
YaoJiayi: Why is this not part of the `self.get_local_mem_indices`?

YaoJiayi (commented):
@novahow LGTM in general! I left some comments.

Two high-level questions:
(1) Why transpose is needed instead of adding a memory format?
(2) We need to make the logic (i.e., sharing the KV cache) less PD or nixl related since we want all backends and all transfer channels benefit from this.

We can postpone doing (2) in this PR and try to have a cleanest possible version first for PD w. nixl.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -29,6 +29,7 @@
 from lmcache.v1.transfer_channel.transfer_utils import (
     P2PInitSideMsg,
     P2PInitSideRetMsg,
+    PDRole,
```
YaoJiayi: Maybe we shouldn't use PDRole for p2p_backend?

examples/disagg_prefill/disagg_proxy_server.py
novahow: Hi @YaoJiayi , I think for 2 folders, the xpyd_tp2 is just an extension for xpyd, which tests it on 1p1d of tp2, just for benchmarking purposes so that we could better compare P tp=1, D tp=2 vs P tp=2, D tp=2. 

As for the `profiler-init-port`, since I saw some comments about switching roles in the future, so I thought  adding that might be helpful, maybe i was misunderstanding the comments. As for now that isn't required, but I think we may need a `--prefiller-tp-size` to replace `--prefiller-init-port` since we can't use `len(decoder_init_port)` to determine the prefill tp size anymore with asymmetric tp.

examples/disagg_prefill/disagg_proxy_server.py
novahow: sure, as mentioned above, we could simply use `prefiller-tp-size` instead

lmcache/integration/vllm/vllm_v1_adapter.py
novahow: I added it just in case in the future `len(receiver_init_port) == tp_size` isn't necessarily true, but I could remove that since they're equal for now.

lmcache/v1/config.py
novahow: since the variable controls whether we want to use asymmetric tp, so that we could initialize our memory allocators correctly. otherwise it seems that we could only know the info in  `batched_put `, however whether to use asymmetric tp needs be known earlier at least in `self.storage_manager.allocate`


```python
            tokens,
            hashes,
            offsets,
            mask,
        ):
            assert isinstance(key, CacheEngineKey)
            # Allocate the memory object
            num_tokens = end - start
            kv_shape = self.gpu_connector.get_shape(num_tokens)
            kv_dtype = self.metadata.kv_dtype

            # TODO (Jiayi): should be batched in the future
            memory_obj = self.storage_manager.allocate(
                kv_shape,
                kv_dtype,
                busy_loop=self.force_store_wait,
            )
            if memory_obj is None:
                logger.warning(
                    "Local cpu memory under pressure so"
                    " choosing to not store the KV cache."
                )
                break

            starts.append(start)
            ends.append(end)
            keys.append(key)
            memory_objs.append(memory_obj)
            tot_token_num += num_tokens

        # memory_objs might be empty, directly return to avoid sending tokens
        if not memory_objs:
            return
        offload_time += time.perf_counter() - t

        t = time.perf_counter()

        transfer_spec = kwargs.get("transfer_spec", None)
        self.storage_manager.batched_put(keys, memory_objs, transfer_spec=transfer_spec)
```

lmcache/v1/memory_management.py
novahow: hello, i only shifted the position of the comments, since the old one didn't show the right annotations on pylance

image

lmcache/v1/memory_management.py
novahow: I had a discussion with @sammshen on this before, 
https://github.com/LMCache/LMCache/pull/1629#discussion_r2396673653

please lmk if you have other ideas on this

lmcache/v1/storage_backend/pd_backend.py
novahow: sorry for the naming confusion, it implies decoder_tp / prefiller_tp ratio

lmcache/v1/transfer_channel/__init__.py
novahow: Hi, the overload function was due to the typing annotation difference in `CreateTransferChannel`between p2p and pd backends, without this the pre-commit check won't pass.

lmcache/v1/storage_backend/pd_backend.py
novahow: hi I'm launching threads because it's a fan-out for prefiller, so if `dp_ratio` is 2, it will send out the corresponding portion of kv cache twice. if `dp_ratio` is 4, we're sending it 4 times. so if it's done sequentially the performance might drop.

csrc/mem_kernels.cu
novahow: we assuming full chunk size here due to our format of KV_DT2L
https://github.com/LMCache/LMCache/pull/1629#discussion_r2393222107

i tried a KV_TD2L before, which didn't need full chunk size, but then one cache block would have len(tokens) descriptors, which I found not really efficient from what i observed in the logs

lmcache/v1/storage_backend/p2p_backend.py
novahow: Hi, I changed the naming to `TransferRole`

lmcache/v1/transfer_channel/nixl_channel.py
novahow: Hi, my initial thought was to keep them separate to adhere to the single-responsibility principle. combining them would mean `get_local_mem_indices` would have two very different responsibilities: extracting addresses from objects or performing buffer sharding and NIXL handler updates. This could make the function's purpose less obvious and harder to maintain.

lmcache/v1/transfer_channel/__init__.py
novahow: Hi, I tried using assert to avoid the mypy pre-commit check but seems to still be failing, so I didn't change that part. Besides, as we add more channel types in the future, we can add more `@overload` definitions to maintain this type safety as well as retain extensibility.

lmcache/v1/storage_backend/pd_backend.py
novahow: Hi, I used the `async_batch_write` to handle this part and removed the `threadpoolexecutor`.

lmcache/v1/config.py
novahow: @YaoJiayi  Hi, I removed the `enable_asym_tp` config and introduced a new memory format to handle the transpose.

lmcache/integration/vllm/vllm_v1_adapter.py
novahow: Removed. it's now sender_tp_size to help us decide transpose in `transfer_spec`

examples/disagg_prefill/disagg_proxy_server.py
novahow: removed xpyd_tp2 example

lmcache/v1/config.py
novahow: Upd: tested 1p2d, which the prefiller tp=1, 1st decoder tp=2, 2nd decoder tp=1 and was working.

sammshen:
@novahow is this ready for review again?

novahow:
> @novahow is this ready for review again?

Hi @sammshen @YaoJiayi . 
Yes, and I made some changes accordingly few days ago. 
Some changes include 
1. removed `transpose` attribute from allocators, now only `memoryObj` would hold the transpose attribute
2. decide `transpose` with transfer_spec and remove `dp_ratio` attribute from backend, enabling more dynamic scenarios such as 2 decoder, 1 with tp=1 and another with tp=2
3. introduced new `MemoryFormat`, and let `get_shape` in `gpu_connector` be aware of the `transpose`

To further incorporate more types of sharding spec, maybe in the future PRs we could directly pass the `memoryFormat` into the `lmc_ops.multilayer_transfer_kernel` instead of the `transpose` flag

YaoJiayi:
@novahow Thanks! Will take a look today!

lmcache/v1/cache_engine.py
```diff
@@ -243,6 +245,16 @@ def store(
...
+        transfer_spec = kwargs.get("transfer_spec", None)
+        transpose = maybe_transpose(transfer_spec) and self.config.enable_pd
+        fmt = self.fmt
+        if transpose:
+            assert isinstance(
+                self.storage_manager.allocator_backend.get_memory_allocator(),
+                PagedCpuGpuMemoryAllocator,
+            ), "Transpose is only supported when using PagedCpuGpuMemoryAllocator"
+            fmt = MemoryFormat.KV_DT2L
+
```
YaoJiayi, line 269: @novahow Why can't this format be detected during start up time?

lmcache/v1/gpu_connector.py
```diff
@@ -1490,5 +1503,5 @@ def batched_from_gpu(
             tmp_gpu_buffer_obj.ref_count_down()
         yield
 
-    def get_shape(self, num_tokens: int) -> torch.Size:
+    def get_shape(self, num_tokens: int, transpose: bool = False) -> torch.Size:
```
YaoJiayi, line 1506: I think the shapes can be detected at start-up time without transpose argument? Also, for these layerwise gpu connectors, PD is not compatible with them for now anyway?

lmcache/v1/storage_backend/__init__.py
```diff
@@ -115,12 +115,13 @@ def CreateStorageBackends(
         # First Party
         from lmcache.v1.storage_backend.pd_backend import PDBackend
 
-        storage_backends["PDBackend"] = PDBackend(config, metadata)
+        storage_backends["PDBackend"] = PDBackend(config, metadata, loop)
```
YaoJiayi, line 132: Why loop is needed here?

lmcache/v1/storage_backend/__init__.py
```diff
@@ -115,12 +115,13 @@ def CreateStorageBackends(
...
 
     # TODO(Jiayi): The hierarchy is fixed for now
     # NOTE(Jiayi): The local_cpu backend is always created because
     # other backends might need it as a buffer.
-    if not config.enable_pd or config.local_cpu:
+    pd_skip_cpu_backend: bool = config.enable_pd and not config.local_cpu
```
YaoJiayi: See my comment at line 121. LocalCPU backend might still be needed even if config.local_cpu is false

lmcache/v1/storage_backend/__init__.py
```diff
@@ -169,18 +178,26 @@ def CreateStorageBackends(
...
+                """
+                remote_backend is skipped because local_cpu_backend
+                is disabled when using PD backend.
+                """
+            )
 
-    if not config.enable_pd or config.local_cpu:
-        # Create dynamic backends from configuration
+    # Create dynamic backends from configuration
+    if not pd_skip_cpu_backend:
```
YaoJiayi: I don't quite get any of the changes in this file. Why are they needed?

lmcache/v1/transfer_channel/nixl_channel.py
YaoJiayi: The reason we have the transfer channel is to let different backends reuse/adhere to the same functions. I think it'd be better to use get_local_mem_indices.

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -376,56 +425,99 @@ def batched_submit_put_task(
         for mem_obj in memory_objs:
             mem_obj.ref_count_up()
 
-        receiver_init_port = transfer_spec.receiver_init_port[self.tp_rank]
-        receiver_alloc_port = transfer_spec.receiver_alloc_port[self.tp_rank]
-        receiver_id = transfer_spec.receiver_host + str(receiver_init_port)
-        receiver_host = transfer_spec.receiver_host
+        # Third Party
+        from vllm.distributed.utils import divide
```
YaoJiayi: Can we avoid importing from vllm unless necessary?

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -468,7 +560,10 @@ def _allocate_and_put(self, alloc_request: AllocRequest) -> AllocResponse:
 
         for idx, key_str in enumerate(alloc_request.keys):
             key = CacheEngineKey.from_string(key_str)
-            if self.contains(key, pin=True):
+            key.update_rank_info_from_pd(self.tp_world_size, self.tp_rank)
+            # NOTE(novahow): pin causes failed to allocate in L40S in TP=(2,2),
+            # disable for now
+            if self.contains(key, pin=False):
```
YaoJiayi: Why `pin` causes failed to allocate? There might be incorrect results without pin?

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -468,7 +560,10 @@ def _allocate_and_put(self, alloc_request: AllocRequest) -> AllocResponse:
 
         for idx, key_str in enumerate(alloc_request.keys):
             key = CacheEngineKey.from_string(key_str)
-            if self.contains(key, pin=True):
+            key.update_rank_info_from_pd(self.tp_world_size, self.tp_rank)
```
YaoJiayi: Please add a comment here explaining why update this key is needed in hetero tp

lmcache/v1/storage_backend/pd_backend.py
```diff
@@ -376,56 +425,99 @@ def batched_submit_put_task(
...
 
-        receiver_init_port = transfer_spec.receiver_init_port[self.tp_rank]
-        receiver_alloc_port = transfer_spec.receiver_alloc_port[self.tp_rank]
-        receiver_id = transfer_spec.receiver_host + str(receiver_init_port)
-        receiver_host = transfer_spec.receiver_host
+        # Third Party
+        from vllm.distributed.utils import divide
+
+        # TODO(novahow): is there better way to obtain receiver tp_size?
+        dp_ratio = divide(len(transfer_spec.receiver_init_ports), self.tp_world_size)
```
YaoJiayi: Please add a comment on explaining this dp_ratio variable.

YaoJiayi (commented):
LGTM in general! Just a few minor comments! I think this is close to be merged:)

lmcache/v1/storage_backend/__init__.py
novahow: I think these are needed before the PR https://github.com/LMCache/LMCache/pull/1733

Now i think we can remove these checks since that PR fixed it.

lmcache/v1/storage_backend/__init__.py
novahow: https://github.com/LMCache/LMCache/pull/1629/files#diff-673f8ab408b20276c52245ab9f0bc41eadcb5e550fcf09fa5f891551502ffc77R489

I use it to run nixl transfer coroutines.

lmcache/v1/cache_engine.py
novahow: Hi, I think it was detected during startup time in older version of my pr, but from our last discussion you mentioned that we should make the layout more dynamic to support multiple decoder with different TP, either symmetric or asymmetric. that's why I made the change to detect the format according to `transfer_spec`

lmcache/v1/gpu_connector.py
novahow: Hi, I think the shape depends on the transfer_spec as well. 

https://github.com/LMCache/LMCache/pull/1629#discussion_r2493370091

lmcache/v1/storage_backend/pd_backend.py
novahow: Hi, could you elaborate more on why pinning is crucial here? Since I got the `failed to allocate` error on TP=2 on the **_dev_** branch when using 4 L40S gpus. I resolved that with pin=False. I tried to lookup the logs for the root cause last month, but didn't gain much insight on why this is happening. 

```
python benchmark_serving.py --port 9487 --seed $(date +%s) \
    --model meta-llama/Llama-3.1-8B-Instruct \
    --dataset-name random --random-input-len 7500 --random-output-len 200 --dataset-path philschmid/mt-bench    --num-prompts 30 --burstiness 100 --request-rate 1 --ignore-eos
```

tp=2 setup:
https://github.com/LMCache/LMCache/tree/09b8889fde34e14eb885e9751ee73b2c38d9bb80/examples/disagg_prefill/xpyd_tp2

sammshen:
@novahow sorry might have to run the pre-commit: 
https://docs.lmcache.ai/developer_guide/contributing.html

sammshen:
This PR seemed to fail the local CPU tests:

```text
Warning: This log has exceeded the 2 MB display limit. Below we are only showing the last 2 MB of output.
See https://buildkite.com/docs/guides/managing-log-output for tips on managing your log output.
You can still download the full log using the "Download" button above.
...
[+] Building 1156.0s (17/17) FINISHED                                                                                                            docker:default
 => [internal] load build definition from Dockerfile                                                                                                       0.0s
 => => transferring dockerfile: 4.42kB                                                                                                                     0.0s
 => [internal] load metadata for nvcr.io/nvidia/cuda-dl-base:25.03-cuda12.8-devel-ubuntu24.04                                                              0.5s
 => [internal] load .dockerignore                                                                                                                          0.0s
 => => transferring context: 2B                                                                                                                            0.0s
 => CACHED [base 1/6] FROM nvcr.io/nvidia/cuda-dl-base:25.03-cuda12.8-devel-ubuntu24.04@sha256:25f3a1b813daf052e67f8b239f7cff9db01ec3fb3c567040b9fac2de32  0.0s
 => [internal] load build context                                                                                                                          0.0s
 => => transferring context: 843.98kB                                                                                                                      0.0s
 => [base 2/6] RUN echo 'tzdata tzdata/Areas select America' | debconf-set-selections     && echo 'tzdata tzdata/Zones/America select Los_Angeles' | deb  49.2s
 => [base 3/6] WORKDIR /workspace                                                                                                                          0.0s
 => [base 4/6] COPY ./requirements/common.txt common.txt                                                                                                   0.0s
 => [base 5/6] COPY ./requirements/cuda.txt cuda.txt                                                                                                       0.0s
 => [base 6/6] RUN --mount=type=cache,target=/root/.cache/pip     . /opt/venv/bin/activate &&     uv pip install -r cuda.txt                             886.5s
 => [image-build 1/6] COPY ./requirements/build.txt build.txt                                                                                              0.0s
 => [image-build 2/6] RUN --mount=type=cache,target=/root/.cache/pip     . /opt/venv/bin/activate &&     uv pip install -r build.txt                       0.5s
 => [image-build 3/6] COPY . /workspace/LMCache                                                                                                            0.1s
 => [image-build 4/6] WORKDIR /workspace/LMCache                                                                                                           0.0s
 => [image-build 5/6] RUN --mount=type=cache,target=/root/.cache/ccache,id=ccache     --mount=type=cache,target=/root/.cache/uv,id=uv-cache,sharing=loc  207.2s
 => [image-build 6/6] WORKDIR /workspace                                                                                                                   0.0s
 => exporting to image                                                                                                                                    11.9s
 => => exporting layers                                                                                                                                   11.9s
 => => writing image sha256:33aa0ccd38e04bd8ad40679827aa25a00ac8894ca26336b96f870d4fa0d261d4                                                               0.0s
 => => naming to docker.io/lmcache/vllm-openai:build-latest                                                                                                0.0s
===== Testing LMCache with local_cpu.yaml =====
✅ Selected GPU #1 (CUDA_VISIBLE_DEVICES=1)
Curl /v1/models endpoint
Model meta-llama/Llama-3.2-1B-Instruct is available on 8000
→ Running long_doc_qa with customed workload config:
{
  "max-inflight-requests": 20,
  "expected-latency-gain": 3.6,
  "model": "meta-llama/Llama-3.2-1B-Instruct"
}
Warmup round mean TTFT: 0.776s
Warmup round time: 1.594s
Warmup round prompt count: 8
=== BENCHMARK RESULTS ===
Query round mean TTFT: 0.470s
Query round time: 0.909s
Query round prompt count: 16
Actual latency gain: 3.51×
ERROR: latency gain 3.51× < expected 3.60×
→ Cleaning up Docker containers and ports...
🚨 Error: The command exited with status 1
user command error: exit status 1
```

I will run the comprehensive tests again. I don't think this PR should affect the kernel performance.

But maybe the new transpose conditional: 
- hurt some compiler optimizations
- hurt parallelization in some way

sammshen:
https://github.com/LMCache/LMCache/pull/1994 (a documentation PR) passes the CPU test (but fails the PD test).

sammshen:
same issue again: 
```text

===== Testing LMCache with local_cpu.yaml =====
--
  | ✅ Selected GPU #1 (CUDA_VISIBLE_DEVICES=1)
  | Curl /v1/models endpoint
  | Model meta-llama/Llama-3.2-1B-Instruct is available on 8000
  | → Running long_doc_qa with customed workload config:
  | {
  | "max-inflight-requests": 20,
  | "expected-latency-gain": 3.6,
  | "model": "meta-llama/Llama-3.2-1B-Instruct"
  | }
  | Warmup round mean TTFT: 0.776s
  | Warmup round time: 1.594s
  | Warmup round prompt count: 8
  |  
  | === BENCHMARK RESULTS ===
  | Query round mean TTFT: 0.470s
  | Query round time: 0.909s
  | Query round prompt count: 16
  | Actual latency gain: 3.51×
  | ERROR: latency gain 3.51× < expected 3.60×
  | → Cleaning up Docker containers and ports...
  | 🚨 Error: The command exited with status 1
  | user command error: exit status 1


```

sammshen:
@novahow 

We should not have runtime conditionals. 

One option is to template the transpose: 

```cpp
template<bool TRANSPOSE>
__device__ __forceinline__ int64_t key_value_offset(
    const int k_or_v, const int layer_idx, const int token_idx,
    const int scalar_offset, const int scalars_per_token,
    const int num_tokens, const int num_layers, const int k_or_v_size) {

  if constexpr (TRANSPOSE) {
    // Transposed layout
    return scalar_offset * num_tokens * k_or_v_size * num_layers +
           token_idx    * k_or_v_size * num_layers +
           k_or_v       * num_layers +
           layer_idx;
  } else {
    // Original layout
    return k_or_v * num_layers * num_tokens * scalars_per_token +
           layer_idx * num_tokens * scalars_per_token +
           token_idx * scalars_per_token +
           scalar_offset;
  }
}
```

sammshen:
and template all the way up: 

```cpp
template<typename scalar_t, bool DIRECTION, bool TRANSPOSE>
__global__ void load_and_reshape_multi_layer_kernel(...)
```

sammshen:
```cpp
if (transpose) {
  load_and_reshape_multi_layer_kernel<scalar_t, DIRECTION, true><<<...>>>(...);
} else {
  load_and_reshape_multi_layer_kernel<scalar_t, DIRECTION, false><<<...>>>(...);
}
```

novahow:
> and template all the way up:
> 
> ```c++
> template<typename scalar_t, bool DIRECTION, bool TRANSPOSE>
> __global__ void load_and_reshape_multi_layer_kernel(...)
> ```

Hi @sammshen , I think this is what I was trying to do before this [commit](https://github.com/LMCache/LMCache/pull/1629/commits/dd9a889bfb288c91d380b1b73b7b9f1f9bc7a0fc). 

But I think your [previous comment](https://github.com/LMCache/LMCache/pull/1629#discussion_r2389301656) is also valid cause that condition is true across all threads, so warp divergence would not happen. Therefore, the degradation shouldn't happen, and the [test](https://buildkite.com/lmcache/comprehensive-tests/builds/356/steps/canvas?sid=019a0a21-ca46-45e3-8103-f3faa4510c3a) for the previous commit shows passing the local_cpu test. 

I'll try to bisect my commits to see where the problem stems from.

ApostaC:
Hey @novahow , really appreciate the great effort you have made here!

Could you please break down this PR into multiple pieces to help the reviewers better review the PR (as well as help early identify the perf regression). WDYT @sammshen @YaoJiayi ?

sammshen:
I think that is a great suggestion.

@novahow Can we start with a PR (`1/N`) just introducing the modifications to the transfer kernels as well as their unit tests? :)

We can start with running the comprehensive tests on that PR and then integrate the remaining changes in later PRs.

novahow:
> I think that is a great suggestion.
> 
> @novahow Can we start with a PR (`1/N`) just introducing the modifications to the transfer kernels as well as their unit tests? :)
> 
> We can start with running the comprehensive tests on that PR and then integrate the remaining changes in later PRs.

Hi @sammshen , I was working on the transfer-kernel only PR [#2008 ], but the comprehensive tests didn't start. Is there any steps required to trigger that test?

sammshen:
@novahow yes, I just tagged the PR with the `full` tag

PR #1631, opened 2025-09-18
[Core] Fix NixlStorageBackend import

tshmilnvidia:
Function get_correct_nixl_device was both moved and renamed in a recent refactor, causing import issues for NixlStorageBackend. This commit fixes this import.

Also gets rid of an extra unneeded import.

### DIFF
--- a/lmcache/v1/storage_backend/nixl_storage_backend.py
+++ b/lmcache/v1/storage_backend/nixl_storage_backend.py
@@ -43,7 +43,7 @@ from lmcache.v1.memory_management import (
     PagedTensorMemoryAllocator,
 )
 from lmcache.v1.storage_backend.abstract_backend import AllocatorBackendInterface
-from lmcache.v1.storage_backend.connector.nixl_utils import get_correct_nixl_device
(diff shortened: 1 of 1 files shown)


panpan0000:
I've confirmed `get_correct_device()` has same implementation as previous code. 

/lgtm

```
-def get_correct_nixl_device(nixl_device: str, worker_id: int) -> str:
-    """
-    Get the correct Nixl device based on the given device string.
-
-    Args:
-        nixl_device (str): The device string, could be cpu or cuda
-
-    Returns:
-        str: The correct device string for Nixl -- with correct
-          device id.
-    """
-    if nixl_device == "cpu":
-        return "cpu"
-    elif nixl_device.startswith("cuda"):
-        return f"cuda:{worker_id}"
-    else:
-        raise ValueError(f"Invalid Nixl device: {nixl_device}")
```

hickeyma (approved):
LGTM, thanks @tshmilnvidia

YaoJiayi (approved):
@tshmilnvidia Thanks for spotting this!

PR #1636, opened 2025-09-18
[Backend][BugFix]: Refactor StorageBackendInterface constructor to enable dynamic loading

### LINKED ISSUE #1635: [Backend][Bug]: Constructor not clearly defined for storage backends that are loaded dynamically
**Describe the bug**
The constructor currently defined for `StorageBackendInterface` is as follows:

```Python
 def __init__(
        self,
        dst_device: str = "cuda",
    ):
```

This is a limited constructor definition and does not correspond with the constructor that the dynamic loading mechanism in LMCache expects which is:

```Python
def __init__(
    config: LMCacheEngineConfig,
    metadata: LMCacheEngineMetadata,
    local_cpu_backend: LocalCPUBackend,
    dst_device: str,
    lookup_server: Optional[LookupServerInterface],
```

Therefore, a user wishing to add a new storage backend dynamically will need to work through runtime errors and trace back to the code and find how the backend is instantiated.

**To Reproduce**
1. Create a new backend as per the documentation: https://docs.lmcache.ai/kv_cache/storage_backends/external_backend.html
2. There is a possibility that it may not be loaded when the LMCache engine starts

**Expected behavior**
A clear and concise constructor definition for `StorageBackendInterface`.

**Additional Comments**
- This feature was added in #1439 
- The definition will need to accommodate the existing OOOTB provided backend constructors
- The interface constructor can not be abstracted as it needs to initialize the torch type

/cc @ApostaC @maobaolong

hickeyma:
There is no clear definition of the constructor required so that a customized backend can be loaded properly during LMCache startup.

~This PR refactors the StorageBackendInterface constructor to make the interface required to be more concise for the implementer.~

This PR adds a new interface  `ConfigurableStorageBackendInterface` to define the interface required tto be implemented which includes the constructor signature.

FIX #1635

### DIFF
--- a/docs/source/kv_cache/storage_backends/external_backend.rst
+++ b/docs/source/kv_cache/storage_backends/external_backend.rst
@@ -11,7 +11,7 @@ Backend Definition Requirements
 4. Package as an installable Python module
 
 How to Integrate the Backend with LMCache
--------------------------------------
+-----------------------------------------
 1. Install your backend package in the LMCache environment
 2. Add ``external_backends`` and its related ``module_path`` and ``class_name`` to ``extra_config`` section of LMCache configuration as follows:
 
--- a/docs/source/kv_cache/storage_backends/external_backend.rst
+++ b/docs/source/kv_cache/storage_backends/external_backend.rst
@@ -22,8 +22,8 @@ How to Integrate the Backend with LMCache
     max_local_cpu_size: 5
     external_backends: <backend_name>
     extra_config:
-      external_backend.log_external_backend.module_path: <module_path>
-      external_backend.log_external_backend.class_name: <class_name>
+      external_backend.<backend_name>.module_path: <module_path>
+      external_backend.<backend_name>.class_name: <class_name>
 
 .. note::
 
--- a/docs/source/kv_cache/storage_backends/external_backend.rst
+++ b/docs/source/kv_cache/storage_backends/external_backend.rst
@@ -3,37 +3,34 @@ External Storage Backends
 
 LMCache supports integrating custom storage backends through dynamic loading. This allows extending cache storage capabilities without modifying core code.
 
-Configuration
--------------
-Add the following to your ``lmcache.yaml``:
+Backend Definition Requirements
+-------------------------------
+1. Inherit from ``StorageBackendInterface``
+2. Add constructor with the same signature as ``StorageBackendInterface``
+3. Implement all abstract methods
+4. Package as an installable Python module
+
+How to Integrate the Backend with LMCache
+-------------------------------------
+1. Install your backend package in the LMCache environment
+2. Add ``external_backends`` and its related ``module_path`` and ``class_name`` to ``extra_config`` section of LMCache configuration as follows:
 
 .. code-block:: yaml
 
     chunk_size: 64
     local_cpu: False
     max_local_cpu_size: 5
-    external_backends: "log_external_backend"
+    external_backends: <backend_name>
     extra_config:
-      external_backend.log_external_backend.module_path: lmc_external_log_backend.lmc_external_log_backend
-      external_backend.log_external_backend.class_name: ExternalLogBackend
-
-Implementation Example
-----------------------
-A sample backend implementation (e.g., https://github.com/opendataio/lmc_external_log_backend/ ):
-
-Key Requirements
-----------------
-1. Inherit from ``StorageBackendInterface``
-2. Implement all abstract methods
-3. Provide the constructor as this example
-4. Package as installable Python module
-
-Usage Notes
------------
-1. Install your backend package in LMCache environment
-2. Add ``external_backends`` configuration and its related ``module_path`` and  ``class_name`` to ``extra_config`` section
-3. Backends are initialized during LMCache startup
-4. Use ``external_backends`` list to enable specific backends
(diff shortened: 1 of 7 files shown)


lmcache/v1/storage_backend/abstract_backend.py
```diff
@@ -10,18 +11,28 @@
...
 
 class StorageBackendInterface(metaclass=abc.ABCMeta):
     def __init__(
         self,
         dst_device: str = "cuda",
+        config: Optional[LMCacheEngineConfig] = None,
+        metadata: Optional[LMCacheEngineMetadata] = None,
+        local_cpu_backend: Optional["LocalCPUBackend"] = None,
+        lookup_server: Optional[LookupServerInterface] = None,
+        loop: Optional[asyncio.AbstractEventLoop] = None,
```
ApostaC: Let's not include this for the default `StorageBackendInterface` since those parameters are pretty tightly coupled to each backend's implementation.

How about we create a base class for all the dynamic loading storage backend? Something like:
```python
class ExternalStorageBackend(StorageBackendInterface):  # I'm open to better class names.
    def __init__(
        self,
        dst_device: str = "cuda",
        config: Optional[LMCacheEngineConfig] = None,
        metadata: Optional[LMCacheEngineMetadata] = None,
        local_cpu_backend: Optional["LocalCPUBackend"] = None,
        lookup_server: Optional[LookupServerInterface] = None,
        loop: Optional[asyncio.AbstractEventLoop] = None,
    ):
        """
        All the external storage backends need to reimplement this function in order to
        override this function.
        """
```

ApostaC (commented):
Thanks @hickeyma . LGTM in general!

lmcache/v1/storage_backend/__init__.py
```diff
@@ -79,12 +79,12 @@ def create_dynamic_backends(
 
             # Create the backend instance
             backend_instance = backend_class(
```
maobaolong, line 80: @hickeyma As this PR modified this, and as the lmc_external_log_backend is a demonstrate project for `external backend framework`, would you like to modify the https://github.com/opendataio/lmc_external_log_backend/blob/main/lmc_external_log_backend/lmc_external_log_backend.py also?

docs/source/kv_cache/storage_backends/external_backend.rst
```diff
@@ -3,37 +3,34 @@ External Storage Backends
...
+1. Install your backend package in the LMCache environment
+2. Add ``external_backends`` and its related ``module_path`` and ``class_name`` to ``extra_config`` section of LMCache configuration as follows:
 
 .. code-block:: yaml
 
     chunk_size: 64
     local_cpu: False
     max_local_cpu_size: 5
-    external_backends: "log_external_backend"
+    external_backends: <backend_name>
```
maobaolong, line 27: As you updated "log_external_backend" to a more generic "<backend_name>", so the L25 and L26 should be modified to  `external_backend.<backend_name>.module_path: <module_path>` and `external_backend.<backend_name>.class_name: <class_name>`.

But if we put a concrete example, it will help some one new to python to know what is `module_path`

maobaolong (commented):
@hickeyma Thanks for the refactor, it look much clear. Left some minor comments

lmcache/v1/storage_backend/__init__.py
hickeyma: Sure. I'll do it once this PR is merged. That way I will be aware of all the changes.

docs/source/kv_cache/storage_backends/external_backend.rst
hickeyma: Good idea. Added example to doc in af0c9c65749e834f18646b03feb7119d89d74a1c

lmcache/v1/storage_backend/abstract_backend.py
hickeyma: Great suggestion. Added separate interface in 2f75bb2710cb41209f22ccd71e11ce5488b0e6b7.

hickeyma:
@ApostaC @maobaolong I have suggestion to rename external backends to configurable backends. 
My thinking is that an external backend usually refers to a server/system/service that resides outside of the system that core process or application is running on. This is not necessarily the case for the backends, is it? I also think the "configurable" better represents the capability being provided.

Interested in your feedback and of others also.

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -101,7 +101,7 @@ def __init__(
         dst_device: str = "cuda",
         lmcache_worker: Optional["LMCacheWorker"] = None,
     ):
-        super().__init__(dst_device)
```
maobaolong: Why these changes are necessary?  @hickeyma

hickeyma: Good observation. There were originally required when parameters were added to the `StorageBackendInterface`. Now that there is a standalone interface, they are no longer necessary.

However, some implementations specify the parameter name when calling the parent constructor and other don't. Therefore for consistency, maybe it reads better to be explicit across all the built-in backends?

maobaolong (commented):
@hickeyma LGTM thanks for this refactor.

maobaolong (approved):

ApostaC (approved):
LGTM!

PR #1647, opened 2025-09-19
[benchmark] add random question to long docs

panpan0000:
Provide random question appending each "long doc"

Background (Why)

 Append a "random" “question” after each “long document” to make the benchmark closer to a "simplified version" of real-world RAG;(previously in log-doc-qa, there's no question after each doc, in multi-round-qa, the question is fixed for each doc.)
 
 The real world is like : enterprise has M documents , diff users will ask different questions ,  the RAG will select some of the M documents( in this long-doc, assuming select 1 for each time),  then append the user's question, as a total prompt.

The content of the M docs will not change, but the N question from users varies for each time.

What's Changed:

- Change the document from fixed "hi hi ..."( ' '.join(['hi'] * N) ) to lightweight random pseudo-word-generator, keeping the total length unchanged.

- add  `--with-question` and `--question-length` parameters to append random questions to the end of each repeated document, thereby introducing controllable variation to each request.

This PR just use most simple `benchmarks/long_doc_qa/long_doc_qa.py` as an explanation of this idea.
although `benchmarks/rag` has more complex design.

Testul result  with my local model
```
 python3 benchmarks/long_doc_qa/long_doc_qa.py    --document-length 100   --num-documents 2   --output-len 8   --repeat-count 1   --max-inflight-requests 1   --with-question   --question-length 20   --port 8000   --model "public/glm-4-9b-chat"
```

```
--- Sending prompt 1/5 ---

Response of request 0:
The text you've provided appears to

--- Sending prompt 2/5 ---

Response of request 1:
The text you've provided appears to

--- Sending prompt 3/5 ---

Response of request 2:
It appears that you've shared a

--- Sending prompt 4/5 ---

Response of request 3:
The text you've provided appears to

--- Sending prompt 5/5 ---

Response of request 4:
The text you've provided appears to
Repeat mode:  random
------warm up round------

--- Sending prompt 1/2 ---

Response of request 0:
The text you've provided appears to

--- Sending prompt 2/2 ---

Response of request 1:
It seems like you've provided a
------query round------

--- Sending prompt 1/2 ---

Response of request 0:
It seems like you've provided a

--- Sending prompt 2/2 ---

Response of request 1:
The text you've provided appears to

=== BENCHMARK RESULTS ===
Warmup round mean TTFT: 0.052s
Warmup round time: 0.489s
Warmup round prompt count: 2
Query round mean TTFT: 0.054s
Query round time: 0.492s
Query round prompt count: 2
```

### DIFF
--- a/benchmarks/long_doc_qa/long_doc_qa.py
+++ b/benchmarks/long_doc_qa/long_doc_qa.py
@@ -54,6 +54,10 @@ Commandline arguments:
 
     --hit-miss-ratio: In query round, control how many of the prompts
     will miss the cache. For example, 3:1 means every fourth repeated prompt
+
+    --with-question: Whether to append a question to each prompt.
+
+    --question-length: Length of each random question appended.
 """
 
 # Standard
@@ -61,6 +65,7 @@ from dataclasses import dataclass
 import argparse
 import asyncio
 import random
+import string
 import sys
 import time
 
@@ -73,6 +78,53 @@ OUTPUT_FILE = None
 completions_mode = False
 visualize = False
 eos_token_id = None
+
+
+class RandomTokenGenerator:
+    def __init__(self, vocab_size: int = 2048, token_len_range=(2, 5)):
+        self._min_len = max(1, int(token_len_range[0]))
+        self._max_len = max(self._min_len, int(token_len_range[1]))
+        cap = max(1, int(vocab_size))
+        # Prebuild vocab to avoid rebuilding per call
+        # Each vocab word is guaranteed to be a single token when space-separated
+        self._vocab = [
+            "".join(
+                random.choices(
+                    string.ascii_lowercase,
+                    k=random.randint(self._min_len, self._max_len),
+                )
+            )
+            for _ in range(cap)
+        ]
+
+    def generate(self, num_tokens: int, chunk_size: int = 4096) -> str:
+        """
+        Generate exactly num_tokens tokens.
+
+        Each word from self._vocab is treated as exactly one token when
+        space-separated, ensuring accurate token count.
+
+        Args:
+            num_tokens: The exact number of tokens to generate
+            chunk_size: Process tokens in chunks to avoid memory issues
+
+        Returns:
+            A string containing exactly num_tokens space-separated tokens
+        """
+        if num_tokens <= 0:
+            return ""
+
+        remaining = int(num_tokens)
+        parts: list[str] = []
+
+        while remaining > 0:
+            k = min(chunk_size, remaining)
+            # Select exactly k tokens from vocab
+            tokens = random.choices(self._vocab, k=k)
+            parts.append(" ".join(tokens))
+            remaining -= k
+
+        return " ".join(parts)
 
 
 @dataclass
@@ -423,9 +475,10 @@ async def main(args):
         timeout=None,
     )
     model = args.model
+    # Instantiate once to avoid rebuilding vocab repeatedly
+    token_gen = RandomTokenGenerator()
 
-    pre_warmup_prompts = [str(i) + "xx" + " ".join(["hi"] * 1000) for i in range(5)]
-
+    pre_warmup_prompts = [str(i) + "xx" + token_gen.generate(1000) for i in range(5)]
     await test_long_document_qa(
         client=client,
         model=model,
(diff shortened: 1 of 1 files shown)


panpan0000:
@ApostaC , I used this PR to explain the idea what we talked.

previously, the prompts are fixed doc selection : [DocA, DocB, DocC...]  
with --repeat , it may be  [DocA,DocA, DocB,DocB, DocC,DocC...]
with random questions  : [Q1, Q2, Q3, Q4, Q5, Q6...]  
So the real requests will be: [ DocA+Q1, DocA+Q2, DocB+Q3, DocB+Q4, DocC+Q5,DocC+Q6]

benchmarks/long_doc_qa/long_doc_qa.py
```diff
@@ -71,6 +76,33 @@
...
+        # Prebuild vocab to avoid rebuilding per call
+        self._vocab = [
+            "".join(
+                random.choices(
+                    string.ascii_lowercase,
+                    k=random.randint(self._min_len, self._max_len),
+                )
+            )
+            for _ in range(cap)
+        ]
```
ApostaC: The key problem is that the random "words" generated by this part of the code may contain more than 1 tokens.

Previously, we were using repeated "hi " because it's guaranteed to be one token for each "hi".

benchmarks/long_doc_qa/long_doc_qa.py
```diff
@@ -426,11 +459,13 @@ async def main(args):
     # we append the document id at the beginning to avoid any of the document
     # being the prefix of other documents
     warmup_prompts = [
-        str(i) + " " + " ".join(["hi"] * args.document_length)
+        str(i) + " " + token_gen.generate(args.document_length)
```
ApostaC, line 486: Based on the above comments, can we keep using `"hi " * args.document_length` here?

ApostaC (changes requested):
Thanks @panpan0000 ! I like the idea of having the `--with-question`. 

Got some comments on how to generate the tokens. The current implementation may generate more tokens than expected.

My proposal is that we can generate the question using this template: 
`[p + " " + f"question {question_id}" + "hi " * question_length for question_id, p in enumerate(prompts)]`

benchmarks/long_doc_qa/long_doc_qa.py
panpan0000: @ApostaC , it seems my code does have same effect as  `hi` , but more randomly ?

----------------------------------------

```
Tokens requested: 5
  Old approach: 'hi hi hi hi hi' -> 5 tokens
  New approach: 'gwbu qmit gkh gwbu uwz...' -> 5 tokens

Tokens requested: 10
  Old approach: 'hi hi hi hi hi hi hi hi hi hi' -> 10 tokens
  New approach: 'uwz gkh gwbu gwbu gwbu qmit qmit qmit qmit gkh...' -> 10 tokens
  ```

Comment (panpan0000) [reply benchmarks/long_doc_qa/long_doc_qa.py:486]: hmm... if random generator create a word "cats", it will be two tokens....

Comment (panpan0000) [reply benchmarks/long_doc_qa/long_doc_qa.py:486]: I changed the vocabulary from "hi hi " to "t1 t2 t3 t4" ...


Tokens requested: 5
  Old approach: 'hi hi hi hi hi' -> 5 tokens
  New approach: 't4  t102  t33  t77  t1022' -> 5 tokens

File: benchmarks/long_doc_qa/long_doc_qa.py [hunk 3]
```diff
@@ -75,6 +79,46 @@
...
+        Returns:
+            A string containing exactly num_tokens space-separated tokens
+        """
+        if num_tokens <= 0:
+            return ""
+
+        remaining = int(num_tokens)
+        parts: list[str] = []
+
+        while remaining > 0:
```
KuntaiDu, line 112: Is there a benefit of chunking the tokens? If not maybe directly random.choice(self._vocab, num_tokens).

benchmarks/long_doc_qa/long_doc_qa.py
```diff
@@ -75,6 +79,46 @@
...
 
+class RandomTokenGenerator:
+    def __init__(self, vocab_size: int = 2048, token_len_range=(2, 5)):
+        self._min_len = max(1, int(token_len_range[0]))
+        self._max_len = max(self._min_len, int(token_len_range[1]))
+        cap = max(1, int(vocab_size))
+        # Use a simple vocabulary with format "t{i}"
+        # to ensure each word is exactly one token
+        # This guarantees accurate token count generation
+        self._vocab = [f"t{i}" for i in range(cap)]
```
KuntaiDu, line 90: Several questions to this method:
- Does `token_len_range` have effect on the generated token sequence?
- If `cap` is very large, say over 100000, will this way of constructing tokens still guarantee that "t{i}" is one single token?

KuntaiDu (commented):
Thanks for contributing! Left some comments.

PR #1656, opened 2025-09-22
[bugfix] fix memory leak in fs connector

chunxiaozheng:
# DES
When use `RemoteBackend`, the `get` method will try to allocate `memory_obj` first and then read data from remote storage to `memory_obj`, if `memory_obj` allocate succeed, but read from remote failed, the ref count of `memory_obj` will always be 1 and never be evicted. 

# TODO
fix other remote connector :) 
- [x] FSConnector
- [ ] InfinistoreConnector
- [ ] LMConnector
- [ ] MooncakeConnector
- [ ] RedisConnector
- [ ] S3Connector

besides, we should think of a more universal way to handle the reference count of `memory_obj`. Currently, this PR is a fast and temp fix.

### DIFF
--- a/lmcache/v1/storage_backend/connector/fs_connector.py
+++ b/lmcache/v1/storage_backend/connector/fs_connector.py
@@ -114,6 +114,7 @@ class FSConnector(RemoteConnector):
         """Get data from file system"""
         file_path = self._get_file_path(key)
 
+        memory_obj = None
         try:
             async with aiofiles.open(file_path, "rb") as f:
                 if self.save_chunk_meta:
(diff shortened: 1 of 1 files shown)


chunxiaozheng:
hi~ @maobaolong @YaoJiayi could you help take a look?

maobaolong (approved):
After an offline discussing with @chunxiaozheng , I was aware that we need a big refactor in future. But for this PR, it LGTM. @YaoJiayi Would you like to take a look at this PR too?

chunxiaozheng:
@sammshen could you help take a look?

sammshen (approved):
LGTM!

PR #1666, opened 2025-09-23
[core] support set limit of cache hit

chunxiaozheng:
This PR support set hit limit, which is used for test. We can enable it by the following config:
```
hit_miss_ratio: 0.7
```
the `hit_miss_ratio` is set to `0.7`, which means the max hit ratio is `30%`

In our scenario, we want to test the performance under different hit rates, but it is difficult to find such a dataset. Therefore, it can be achieved through above config and the following two steps:
1. Run the dataset and save the data to lmcache
2. Run the same dataset again

By step 1, we can save all kv cache chunk in lmcache(Note: even if the hit rate is set to 30%, all data will still be stored). Then in step 2, the `hit_miss_ratio` will work. 

We have tested in our env, the metrics in grafana are as follows:
- lookup total tokens
Clipboard_Screenshot_1758632829

- hit token tokens
Clipboard_Screenshot_1758632849

-hit rate
Clipboard_Screenshot_1758678923

### DIFF
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -356,7 +356,7 @@ _CONFIG_DEFINITIONS: dict[str, dict[str, Any]] = {
         "default": 3000,
         "env_converter": int,
     },
-    "hit_limit_upper": {
+    "hit_miss_ratio": {
         "type": Optional[float],
         "default": None,
         "env_converter": float,
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -356,6 +356,11 @@ _CONFIG_DEFINITIONS: dict[str, dict[str, Any]] = {
         "default": 3000,
         "env_converter": int,
     },
+    "hit_limit_upper": {
+        "type": Optional[float],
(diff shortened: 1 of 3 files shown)


chunxiaozheng:

maobaolong:
@chunxiaozheng This looks an useful approach to test the performance under a specified cache hit rate, will take a review tomorrow, thanks for this useful contribution first!

maobaolong (approved):
It LGTM. Thanks a lots!

sammshen:
@maobaolong  @chunxiaozheng  the long doc qa also has a `--hit-miss-ratio` parameter that we have been using! :)

sammshen (approved):

maobaolong:
@chunxiaozheng  So it would be nice if you make the name align to the parameter of long qa. 
Concretely, `hit_miss_ratio` should be the good name candidate.

chunxiaozheng:
@maobaolong @sammshen Thanks for your review, I have updated with `hit_miss_ratio`.

PR #1669, opened 2025-09-24
Remove redundancy. Also old version of vllm do not support size local.

yoo-kumaneko:
We don't need to have the first condition because in the case of TP=1, the local rank is 0, which satisfies the second condition.


yoo-kumaneko:
@maobaolong @chunxiaozheng Can you please review this PR?

maobaolong (approved):
LGTM

chunxiaozheng (approved):
lgtm!

PR #1671, opened 2025-09-24
Support remove around quotes in env config

maobaolong:
For some reason, there are quotes surround the config value from environment variable. 
This PR aims to detect this case and remove the extra heading and tailing quote.

### DIFF
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -99,24 +99,6 @@ def _parse_quoted_string(value: str) -> str:
     return value
 
 
-def _manual_unescape(s: str) -> str:
-    """Manually handle common escape sequences."""
-    replacements = {
-        "\\n": "\n",
-        "\\t": "\t",
-        "\\r": "\r",
-        "\\\\": "\\",
-        '\\"': '"',
-        "\\'": "'",
-    }
-
-    result = s
-    for escaped, unescaped in replacements.items():
-        result = result.replace(escaped, unescaped)
-
-    return result
-
-
 # Configuration aliases and deprecated mappings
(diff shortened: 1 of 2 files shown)


maobaolong:

lmcache/v1/config.py
```diff
@@ -71,6 +71,62 @@ def _to_bool(
...
+        "\\\\": "\\",
+        '\\"': '"',
+        "\\'": "'",
+    }
+
+    result = s
+    for escaped, unescaped in replacements.items():
+        result = result.replace(escaped, unescaped)
+
+    return result
```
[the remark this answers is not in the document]
maobaolong: looks good

lmcache/v1/config.py
```diff
@@ -71,6 +72,51 @@ def _to_bool(
...
+                return evaluated
+        except (ValueError, SyntaxError):
+            # If ast.literal_eval fails, it's not a valid Python literal.
+            # Fall back to simply stripping the outer quotes.
+            return value[1:-1]
+
+    return value
+
+
+def _manual_unescape(s: str) -> str:
```
chunxiaozheng: Where is this method used?

maobaolong: This is a legacy from the init commit of this PR, and it should be removed for now. @chunxiaozheng Thanks for your remind.

maobaolong:
@chunxiaozheng The extra method has been removed, PTAL.

chunxiaozheng (approved):
LGTM!

sammshen (approved):
LGTM!

PR #1674, opened 2025-09-24
[DOCS] Leverage LMCache KV calculator logic for memory projections

mcgrof:
Extend the existing KV cache calculator's proven formulas to generate memory projections for extreme context lengths. By reusing the calculator's model-specific logic, we ensure consistency between the web interface and projection visualizations.

The implementation directly imports and uses the calculator's three computation methods:
- Standard models: 2 × layers × tokens × kv_heads × (hidden_size/attention_heads)
- DeepSeek models: layers × tokens × (kv_lora_rank + qk_rope_head_dim)
- Qwen3 models: 2 × layers × tokens × kv_heads × head_dim

This approach guarantees that memory estimates remain synchronized across all LMCache tools, preventing discrepancies between the interactive calculator and static projections.

The projections validate the calculator's accuracy while revealing critical infrastructure requirements:
- DeepSeek-V3: 65.4 GiB at 1M context (7x reduction via KV-LoRA)
- Llama-405B: 480.7 GiB at 1M context (6x H100 GPUs minimum)
- Small models (1-3B): Viable for 1M context on single datacenter GPUs


ziruiliu:
totally agree that current calculator is way behind. can you also include latest MOE models? like Qwen3-Next and Qwen3-Omni series

mcgrof:
> totally agree that current calculator is way behind. can you also include latest MOE models? like Qwen3-Next and Qwen3-Omni series

This is getting unwieldy fast, ie, how do we scale? And so I decided to take this up a notch. I'd like to suggest what if we have a new project under LMCache for a full simulator based on all the math we have on LMCache?

Here's a PoC:

https://kvcache.io/

KevinCheung2259:
Coooooooooooool !

mcgrof:
Based on community discussions I'm folding  https://kvcache.io/ work under LMcache to scale visualizations.

PR #1675, opened 2025-09-24
[bugfix] fix broadcast and to_gpu error

chunxiaozheng:
When `save_only_first_rank` is enabled, the broadcast operation use default stream, this is a async operation in cuda. #1463 use `load_stream` to execute `batched_to_cpu` in `VLLMPagedMemGPUConnectorV2`,  the `load_stream` and default stream concurrent execution will result in some unknown error. 

This PR ensure that the broadcast operation and to_gpu operation are executed in sequence.

#1657 can also work, but use `torch.cuda.synchronize()` is a much heavier operation than the current implementation.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -97,6 +97,9 @@ class LMCacheEngine:
             and metadata.use_mla
         )
 
+        if self.save_only_first_rank:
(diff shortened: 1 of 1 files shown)


YaoJiayi (approved):

maobaolong (approved):
Nice catch! LGTM

PR #1575, opened 2025-09-11
[Test]: Enable unit tests to run on non-CUDA environment

hickeyma:
Enable unit tests where applicable to run on non-CUDA environment. The goal is to be able to run `pytest` and only the relevant tests run.

Unit tests:

- [x] Tests for storage backend: `pytest tests/v1/storage_backend/`
- [x] Skip NIXL tests if nixl not installed: `pytest tests/disagg tests/v1/test_nixl_storage.py`
- [x] Benchmark tests: `pytest tests/benchmarks/` (caveat: skipping `benchmark` tests for now due to CUDA complexity)
- [x] Tests in `tests`: `pytest tests/*.py` (caveat: skipping `tests/test_serde.py` for now due to CUDA complexity)
- [x] Tests in `tests/v1`: `pytest tests/v1*.py`

CI:

- [x] Add GH workflow to test the unit tests as non-CUDA

FIX #1503 

**Note:** 

- Used Mac M-series as the test system
- Changes made to core code was in a clean incisive fashion with minimal of change to enable some of the tests to run on a non-CUDA environment. There are not meant to enable LMCache to be able to run fully on non-CUDA envoronments.
- Opened PR #1664 for follow on work where finding an equivalent non CUDA implementation is more complex
- There are a numbers of warnings listed post unit test run. The warnings are inherited and not caused by this PR (https://buildkite.com/lmcache/unit-tests/builds/5196#0199516d-ea47-4f7a-b013-d578a66297b2)

PR #1677, opened 2025-09-24
[CI][BugFix]: Enable non-CUDA unit testing for all tests

hickeyma:
GitHub workflow are only running tests on `tests/v1/storage_backend/`. This PR enables it for all. It is a follow on from 
#1575 where we missed that test only ran on storage backends.


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #1678, opened 2025-09-24
Get a default value if old vllm doesn't have priority

yoo-kumaneko:
Since old versions of the vLLM do not support priority, we need to have a default fallback value.


yoo-kumaneko:
@maobaolong @chunxiaozheng Can you please take a look at this PR?

maobaolong:
@yoo-kumaneko it looks good, I will approve after you verified it work.

YaoJiayi (approved):
lgtm

PR #1685, opened 2025-09-25
[minor fix]fix pin_count type

chunxiaozheng:
the type of `pin_count` in `MemoryObjMetadata` is `int`, use `0` rather than `False` to init the object.


chunxiaozheng:

maobaolong (approved):
Good catch! LGMT

chunxiaozheng:
@sammshen could you help take another look?

sammshen:
Good catch! False is a subtype of int in python which is why the lints did not complain before

sammshen (approved):

Issue #1694, opened 2025-09-25
[RFC][Core]: Replace assert statements in project


hickeyma:
### Motivation

Using `assert` is a development time sanity check and **NOT** a runtime safeguard. Production code should have robust error handling, logging, and clear exception management.

https://www.inspiredpython.com/article/assertions-and-exceptions-are-not-the-same:
- Assertions are programmer aids, not error handlers for your business logic
- Assertions can be toggled on or off at will

[From Python docs](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement): "Assert statements are a convenient way to insert debugging assertions into a program 

Unfortunately, in the core code base (not including `tests/`) there is widespread use of `assert` and all contributions are using this pattern. We have logging but not robust error handling and clear exception management. 

This RFC is to recommend robust error handling and clear exception management into the project.

### Proposed Change:

Python provides a robust [error handling  and exception management](https://docs.python.org/3/tutorial/errors.html) through its `try` and `except` blocks which allows to gracefully manage unexpected situations.

Errors are generaded by raising (`raise`) exceptions manually when something goes wrong. This would replace the use of `assert` directly. 
Exception handling would then need to be added throughout the call stack to handle exceptions that are raised in that stack. Python provides the ability to handle this in a robust and informative fashion to help the user, using `try` and `except`.

### Additional concerns:

This is a large undertaking which requires updates to most of the code base. It is recommended to do this in a piece meal approach with each PR that adds exceptions raised in a file or code area, must also add the exception handling throughout the stack. It is probably best to leave this issue open with `help wanted` label and to welcome PRs created.

/cc @ApostaC @YaoJiayi @sammshen @maobaolong @Shaoting-Feng

hickeyma:
Created a draft PR to demonstrate potential pattern for the  work to be done: PR #1695

ApostaC:
I think this proposal makes sense. Adding a few more things:
- We should explicitly add the potential exception that a function will raise into the docstring
- What's the performance implications of using try-catch compared to assert? I have experienced performance degradation caused by try-catch in c++ but not sure about python
- Sometimes, the assert is used to suppress the linter's complains. Using try-catch for that could be an overkill?

sammshen:
https://stackoverflow.com/questions/2522005/cost-of-exception-handlers-in-python

When assert fails, is this the behavior we want to continue execution? would we rather crash if a key assumpiton is not met?

hickeyma:
[...]

PR #1695, opened 2025-09-25
[WIP][Core]: Refactor assert in cache engine with standard Python error handling

hickeyma:
Thsi PR usese the standard Python error handling instead of `assert` statements. It also updates files which use the cache engine functions that are updated.

Refs: #1694


hickeyma (commented):
Going to close this PR for now as the using `assert` seems to be preferred over error handling during inferencing of models in most cases primarily for seed and latency reasons.

PR #1708, opened 2025-09-27
Support adjust the effective memory by system available memory and reserve size

maobaolong:


maobaolong:

lmcache/v1/system_detection.py
```diff
@@ -21,6 +28,30 @@ class NUMAMapping:
...
+    def get_available_memory_gb() -> float:
+        """
+        Get system available memory in GB using psutil.
+        This method is cross-platform and doesn't require subprocess calls.
+
+        Returns:
+            Available memory in GB, or 0.0 if detection fails.
+        """
+        try:
+            # Use psutil to get virtual memory information
```
sammshen, line 42: not all of the memory is pinnable / page-lockable

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -266,17 +319,18 @@ def initialize_allocator(
...
+                # so we need to set it larger than other ranks
+                cpu_size = config.get_extra_config_value(
+                    "first_rank_max_local_cpu_size", cpu_size
                 )
 
         # Detect the numa mapping
         numa_mapping = NUMADetector.get_numa_mapping(config)
         logger.info(f"NUMA mapping {numa_mapping}")
 
+        # Calculate effective CPU memory size
```
sammshen, line 334: we should not always override the cpu size right, only in the save_only_first_rank case maybe

lmcache/v1/system_detection.py
sammshen: previous heuristic for setting cpu size: 
https://docs.lmcache.ai/getting_started/benchmarking.html

sammshen:
I think this PR is ok as long as it is extra config which it is not right now

sammshen:
LGTM, I think this is fine since it overestimates

sammshen (approved):

lmcache/v1/storage_backend/local_cpu_backend.py
maobaolong: done

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -250,6 +250,61 @@ def remove(self, key: CacheEngineKey, force: bool = True) -> bool:
...
+
+        # TODO(baoloongmao): For disable save_only_first_rank case,
+        #  we need to avoid multi-rank race condition in future.
+        #  But for enable save_only_first_rank case,
+        #  we can handle reserve memory simply since non-first ranks
+        #  do not allocate memory.
+        # Effective memory: min(configured_size, available_memory - reserve_size)
+        if system_available_memory_gb > 0:
+            max_usable_memory = max(0, system_available_memory_gb - reserve_cpu_size)
+            effective_cpu_size = min(configured_cpu_size, max_usable_memory)
```
chunxiaozheng, line 294: `effective_cpu_size` may be 0, should an exception or assert be thrown in this situation?

chunxiaozheng (approved):
lgtm

PR #1711, opened 2025-09-28
[core]Add cache/clear api to internal_api_server

maobaolong:
# Description

This PR add a new API `cache/clear`, it allows to clear the cache belongs specified backend or all backends if not specified.

# Test

Add some of UTs.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -980,11 +980,9 @@ class LMCacheEngine:
         if self.save_only_first_rank:
             if self.metadata.is_first_rank():
                 num_removed = self._clear(tokens, locations, request_configs)
-                self.broadcast_object_fn(num_removed, self.metadata.first_rank)
                 return num_removed
             else:
-                num_removed = self.broadcast_object_fn(None, self.metadata.first_rank)
-                return int(num_removed)
+                return 0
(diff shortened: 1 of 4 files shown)


lmcache/v1/cache_engine.py
```diff
@@ -980,11 +980,9 @@ def clear(
         if self.save_only_first_rank:
             if self.metadata.is_first_rank():
                 num_removed = self._clear(tokens, locations, request_configs)
-                self.broadcast_object_fn(num_removed, self.metadata.first_rank)
```
chunxiaozheng, line 983: hi @maobaolong, why delete the broadcast here?

maobaolong: Since the clear operation is not from vllm, but from `lmcache_controller` or others, it is not invoked like multiple tp style, so I think it is not necessary to broadcast.

chunxiaozheng: @maobaolong thanks for your reply, I see, if `save_only_first_rank` is enabled, we only need to clear in one worker.

chunxiaozheng (approved):
lgtm

maobaolong:
@sammshen Would you like to take another look?

sammshen (approved):
LGTM! thanks for the clarification @chunxiaozheng

PR #1733, opened 2025-10-01
Add a LMCacheBypassLookupClient to support lookup without communicate to any worker

maobaolong:
# Description

Reduce the `lookup` communicate cost between scheduler and worker for mla with remote backend.

Clipboard_Screenshot_1761032114

# Test

```
enable_scheduler_bypass_lookup: True
```


lmcache/v1/lookup_client/lmcache_lookup_client_bypass.py
```diff
@@ -0,0 +1,123 @@
...
+        self.lmcache_engine = lmcache_engine
+        self.config = config
+
+        # Initialize token database for processing tokens
+        self.enable_blending = config.enable_blending
+        self.token_database: TokenDatabase
+        if self.enable_blending:
+            self.token_database = SegmentTokenDatabase(config, metadata)
+        else:
+            self.token_database = ChunkedTokenDatabase(config, metadata)
```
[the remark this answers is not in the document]
maobaolong: Done

maobaolong:
@YaoJiayi Would you like to take a look at this PR? Thanks!

maobaolong:

chunxiaozheng:
@maobaolong This is an effective feature! Could you leave some comments explaining under what scenarios this feature can be enabled?

maobaolong:
@chunxiaozheng Add an assertion, the `enable_scheduler_bypass_lookup` is only supported one of `save_only_first_rank` or `remote_enable_mla_worker_id_as0` enabled.

chunxiaozheng (approved):
LGTM!

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -495,21 +498,24 @@ def _init_lmcache_engine(
...
+        role,
     )
 
     use_gpu = need_gpu_interm_buffer(lmcache_config)
-    vllm_gpu_connector: Union[
-        VLLMBufferLayerwiseGPUConnector,
-        VLLMPagedMemGPUConnectorV2,
-        VLLMPagedMemLayerwiseGPUConnector,
-    ]
+    vllm_gpu_connector: Optional[GPUConnectorInterface]
```
sammshen, line 510: this is a nice clean up

lmcache/config.py
```diff
@@ -33,6 +33,8 @@ class LMCacheEngineMetadata:
     kv_shape: tuple[int, int, int, int, int]
     """ whether use MLA"""
     use_mla: bool = False
+    """ the role of the current instance (e.g., 'scheduler', 'worker') """
+    role: Optional[str] = None
```
sammshen, line 37: where is this passed in?

sammshen: do we ever initialize it as worker?

lmcache/v1/lookup_client/lmcache_lookup_client_bypass.py
```diff
@@ -0,0 +1,101 @@
...
+                hashes = []
+                offsets = []
+                for start, end, key in self.token_database.process_tokens(
+                    token_ids, make_key=False
+                ):
+                    hashes.append(key)
+                    offsets.append(end - start)
+
+                # Call LMCacheEngine lookup with hashes and offsets
+                result = self.lmcache_engine.lookup(
```
sammshen, line 74: sorry if the gpu connector is none what is this lookup doing exactly? the backends will never be populated? can you help me resolve my understanding?

maobaolong: The `lmcache_engine` for `lookup_client` is used only for `lookup` purpose, and it live within `scheduler` role only, in another word, it will never execute `get` `put` such data relevent operation.

lmcache/config.py
maobaolong: Good catch, although it is not used for worker role, initialize it also for worker role can be better.

maobaolong:
@sammshen Thanks for the previous review, updated the code, PTAL.

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -195,7 +195,10 @@ def __init__(
         )
         self.thread.start()
 
-        if torch.cuda.is_available():
+        # For scheduler role, always use CPU device
+        if metadata.role == "scheduler":
```
sammshen: can we create a helper 
```python
is_cuda_worker(metadata): 
     return metadata.role != "scheduler" and torch.cuda.is_available()
```

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -230,7 +237,7 @@ def __init__(
         self.async_serializer: Optional[AsyncSerializer] = None
 
         # The cuda stream for internal copies during put
-        if torch.cuda.is_available():
+        if torch.cuda.is_available() and metadata.role != "scheduler":
```
sammshen: can also use the helper here

lmcache/v1/lookup_client/lmcache_lookup_client_bypass.py
sammshen: so lookup will always return False then can we just skip the lookup altogether

maobaolong: It's not like that, the `lmcache_engine` in the `scheduler` role can be responsible for lookup, so the lookup will do return a actual boolean value stand for existence of a specified key.

lmcache/v1/storage_backend/storage_manager.py
maobaolong: Good suggestion!

lmcache/v1/storage_backend/storage_manager.py
maobaolong: Done

maobaolong:
@sammshen I abstracted the helper method under your suggestion, PTAL.

lmcache/v1/storage_backend/connector/__init__.py
```diff
@@ -92,13 +111,19 @@ def __init__(
         self,
         url: str,
         loop: asyncio.AbstractEventLoop,
-        local_cpu_backend: LocalCPUBackend,
+        local_cpu_backend: Optional[LocalCPUBackend],
```
sammshen, line 114: why optional if using a stub?

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         config: LMCacheEngineConfig,
         metadata: LMCacheEngineMetadata,
         loop: asyncio.AbstractEventLoop,
-        local_cpu_backend: LocalCPUBackend,
+        local_cpu_backend: Optional[LocalCPUBackend],
```
sammshen, line 31: same nit here. we can remove all the changes in this file?

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -214,7 +215,11 @@ def __init__(
 
         self.enable_pd = config.enable_pd
 
-        self.allocator_backend = self._get_allocator_backend(config)
+        self.allocator_backend = (
```
sammshen: super nit (sorry): 

```python

self.allocator_backend = None
if metadata.role != "scheduler": 
    self.allocator_backend = self._get_allocator_backend(config)
```

lmcache/v1/lookup_client/lmcache_lookup_client_bypass.py
sammshen: why will there be any hits for the hashes? I still don't understand where the `lmcache_engine` with the `scheduler` role ever gets populated? 

sorry for lots of questions. I don't see the purpose of giving the scheduler an "empty" (no cpu backend and no gpu connector) lmcache engine. the lookup will always return 0?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -553,7 +560,19 @@ def _init_lmcache_engine(
         tpg.broadcast,
         tpg.broadcast_object,
     )
-
+    assert (
```
sammshen: super nit (sorry): 
```python

if role == "scheduler" and lmcache_config.enable_scheduler_bypass_lookup:
    assert (
            engine.save_only_first_rank
            or lmcache_config.get_extra_config_value(
                "remote_enable_mla_worker_id_as0", metadata.use_mla
            )
        )
```

sammshen:
hey @maobaolong , I left one main question. I am still really confused on what the point of this lmcache engine is if the lookup will always return 0? `store()` is never called. 

I also was annoying and left some tiny nits but only because the main question is still blocking approval (feel free to ignore)

lmcache/v1/storage_backend/storage_manager.py
maobaolong: done

lmcache/integration/vllm/vllm_v1_adapter.py
maobaolong: done

lmcache/v1/storage_backend/connector/__init__.py
maobaolong: Since we don't need `local_cpu_backend` while `role=scheduler`, the logic is put in the `lmcache/v1/storage_backend/__init__.py`

```python
    local_cpu_backend: Optional[LocalCPUBackend] = None
    if metadata.role == "scheduler":
        # For scheduler role, local_cpu_backend is None
        pass
```

sammshen:
thanks for the explanation @maobaolong !! LGTM

sammshen (approved):

maobaolong:
@chunxiaozheng @sammshen Thanks for your review!

PR #1771, opened 2025-10-11
[FSConnector] support read ahead in FSConnector

chunxiaozheng:
When `save_chunk_meta` is set to `False`,  `num_read = await f.readinto(buffer)` will read the contents of the file at once, the read ahead of DFS may not be effective. This PR supports read twice, the first time reading a small IO to trigger a read ahead, and the second time reading from memory.

We can enable this feature by the following config, default is `None`, means staying consistent with previous behavior.
```
extra_config:
  fs_connector_read_ahead_size: 1048576
```

### DIFF
--- a/lmcache/v1/storage_backend/connector/fs_connector.py
+++ b/lmcache/v1/storage_backend/connector/fs_connector.py
@@ -171,10 +171,16 @@ class FSConnector(RemoteConnector):
                         num_read_ahead = await f.readinto(
                             buffer[: self.read_ahead_size]
                         )
-                        assert num_read_ahead is not None
-                        num_read_tail = await f.readinto(buffer[self.read_ahead_size :])
-                        assert num_read_tail is not None
-                        num_read = num_read_ahead + num_read_tail
+                        assert num_read_ahead <= self.read_ahead_size
+
+                        if num_read_ahead == self.read_ahead_size:
+                            num_read_tail = await f.readinto(
+                                buffer[self.read_ahead_size :]
+                            )
+                            assert num_read_tail is not None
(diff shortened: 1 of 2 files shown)


chunxiaozheng:

lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -142,13 +156,25 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
                 buffer = memory_obj.byte_array
-                num_read = await f.readinto(buffer)
                 if self.save_chunk_meta:
+                    num_read = await f.readinto(buffer)
                     if num_read != len(buffer):
                         raise RuntimeError(
                             f"Partial read data {len(buffer)} got {num_read}"
                         )
                 else:
+                    if self.read_ahead_size is None:
```
maobaolong, line 168: It will be graceful if you could care about the tail chunk case.

lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -35,13 +36,14 @@ def __init__(
         base_paths_str: str,
         loop: asyncio.AbstractEventLoop,
         local_cpu_backend: LocalCPUBackend,
-        relative_tmp_dir: Optional[str],
+        config: Optional[LMCacheEngineConfig],
```
maobaolong, line 39: Nice update.

maobaolong (commented):
@chunxiaozheng Thanks for this improvement, left a comment.

maobaolong (approved):
LGTM

sammshen (approved):
LGTM!

PR #1778, opened 2025-10-13
[core] add batched_contains interface

chunxiaozheng:
add `batched_contains` interface, we need to implement this interface in backends and connectors.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -689,22 +689,23 @@ class LMCacheEngine:
             assert hashes is not None
             lookup_request_id = self.stats_monitor.on_lookup_request(sum(offsets))
 
+        res = 0
         try:
-            end = 0
-            prev_end = 0
-
             if pin:
                 assert lookup_id is not None, "lookup_id is required when pin is True"
 
-            for start, end, key in self.token_database.process_tokens(
+            chunk_infos = self.token_database.process_tokens(
                 tokens=tokens,
                 hashes=hashes,
                 offsets=offsets,
                 request_configs=request_configs,
-            ):
-                assert isinstance(key, CacheEngineKey)
+            )
 
-                if self.use_layerwise:
+            # TODO: support batched_contains when layerwise is enabled
+            if self.use_layerwise:
+                for start, end, key in chunk_infos:
+                    assert isinstance(key, CacheEngineKey)
+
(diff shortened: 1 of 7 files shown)


chunxiaozheng:

lmcache/v1/cache_engine.py
```diff
@@ -685,22 +685,20 @@ def lookup(
             assert hashes is not None
             self.stats_monitor.on_lookup_request(sum(offsets))
 
+        res = 0
         try:
-            end = 0
-            prev_end = 0
-
-            if pin:
```
sammshen, line 696: can we keep this assert? :)

sammshen: nvm I see it below

lmcache/v1/cache_engine.py
```diff
@@ -713,29 +711,44 @@ def lookup(
...
+                                "lookup_id is required when pin is True"
                             )
-                        prev_end = end
+                            self.lookup_pins[lookup_id].append(key)
+                        res = end
                         continue
-
-                    end = prev_end
-                    return prev_end
+                    return res
```
sammshen, line 750: thanks for cleaning up the prev_end logic

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -621,6 +621,67 @@ def contains(
 
         return None
 
+    def batched_contains(
+        self,
+        keys: List[CacheEngineKey],
+        search_range: Optional[List[str]] = None,
+        pin: bool = False,
+        stop_after_first_not_exits: bool = True,
```
sammshen, line 629: what's the use case for this being False?

sammshen (approved):
LGTM! this will be very useful

lmcache/v1/storage_backend/storage_manager.py
chunxiaozheng: > what's the use case for this being False?

@sammshen Thanks for your review! `stop_after_first_not_exits` must be set to `True` in the current scenario, when find the first `False`, all subsequent ones do not need to be judged and return False directly. 
This parameter is reserved for this situation, if we want to know if all the keys exist in the future, we can set it to `False`.

maobaolong (approved):
LGTM. Thanks!

Issue #1781, opened 2025-10-13
[Core][LocalCPU] NUMA-aware pinned pool when external memory is system RAM


DongDongJu:
Summary:

  - When external memory like CXL memory is exposed as ordinary system RAM, the LocalCPU backend still allocates its page-locked pool without regard to NUMA location. This causes large GPU↔CPU transfers to traverse slow links on multi-socket or CXL-heavy systems.
  - Enhance LocalCPU so its pinned pool honors NUMA placement policies per allocation: respect manual GPU→NUMA mappings, auto-detected mappings, or fallback nodes; support both bind and preferred policies; log decisions, and fall back cleanly if NUMA controls are unavailable.
  - Include a benchmark script that contrasts baseline vs NUMA-managed vs disk-backed throughput, highlighting why placement matters for CXL system-RAM scenarios.
  - Request focused review of lmcache/v1/storage_backend/local_cpu_backend.py, lmcache/v1/memory_management.py, lmcache/v1/system_detection.py, and supporting C++ helpers.

PR #1782, opened 2025-10-13
[Core][Doc] Add NUMA placement controls and benchmark for LocalCPU backend

DongDongJu:
implementation issue for #1781 

Key changes:

  1. Extend the C++ allocation helpers (alloc_pinned_numa_policy_ptr) and pybind exports so we can bind or prefer
     sets of NUMA nodes, falling back gracefully when NUMA APIs are absent.
  2. Teach LocalCPU to resolve a NUMAPlacement from config (numa_mode, gpu_to_numa_mapping, numa_policy,
     numa_nodes_fallback), apply it to both mixed and paged allocators, log the policy/source, and revert to the
     legacy behavior when multi-node or library support is missing.
  3. Document the new knobs beside numa_mode and introduce examples/benchmarks/numa_pinned_backend_demo.py, which
     compares baseline, per-node, and disk-backed throughput—useful for CXL system-RAM environments.
  4. Note in the PR description that contributors must reinstall the extension (pip install -e .) so the new
     c_ops symbol is available.

  Testing to mention:

  sudo python \
    -m examples.benchmarks.numa_pinned_backend_demo \
    --size-mb 512 --iterations 8 \
    --mode manual --policy bind \
    --nodes 0 1 2 --skip-mapping \
    --compare-baseline --per-node --compare-disk

Result in our three socket system with one external memory device compared with disk backed pool
```
LocalCPU NUMA Benchmark (size=512 MiB, iter=8)
baseline       elapsed=  0.082s throughput= 52.64 GiB/s
    Pinned pool without NUMA placement.
numa-enabled   elapsed=  0.049s throughput= 87.21 GiB/s
    Pinned pool with requested NUMA policy.
node-0         elapsed=  0.054s throughput= 79.84 GiB/s
    Pinned pool bound to NUMA node 0.
node-1         elapsed=  0.049s throughput= 87.06 GiB/s
    Pinned pool bound to NUMA node 1.
node-2         elapsed=  0.363s throughput= 11.83 GiB/s
    Pinned pool bound to NUMA node 2.
disk-backed    elapsed=  5.403s throughput=  0.79 GiB/s
    Sequential write to disk-backed storage.
    placement: disk
```

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -198,8 +198,10 @@ class LMCacheEngine:
         # NOTE(ApostaC): we haven't support lookup-cache yet
         self.lookup_cache: dict[CacheEngineKey, Any] = {}
 
-        # lookup_id -> [pinned keys]
-        self.lookup_pins: dict[str, list] = defaultdict(list)
+        # lookup_id -> {location -> [pinned keys]}
+        self.lookup_pins: dict[str, dict[str, list]] = defaultdict(
+            lambda: defaultdict(list)
+        )
 
         InitializeUsageContext(config.to_original_config(), metadata)
         self.stats_monitor = LMCStatsMonitor.GetOrCreate()
@@ -794,20 +796,20 @@ class LMCacheEngine:
                     # of one layer
                     key_all_layers = key.split_layers(self.num_layers)
 
-                    found = False
-                    for key_single_layer in key_all_layers:
-                        if self.storage_manager.contains(
-                            key_single_layer, search_range, pin
-                        ):
-                            found = True
-                    if found:
+                    hit_chunks, block_mapping = self.storage_manager.batched_contains(
+                        key_all_layers,  # type: ignore
+                        search_range,
+                        pin,
+                    )
+                    # Only all layers are hit and hit in one location,
+                    # we consider this key as a hit
+                    if hit_chunks == self.num_layers and len(block_mapping) == 1:
                         if pin:
                             assert lookup_id is not None, (
                                 "lookup_id is required when pin is True"
                             )
-                            self.lookup_pins[lookup_id].extend(  # type: ignore
-                                key_all_layers
-                            )
+                            location = next(iter(block_mapping.keys()))
+                            self.lookup_pins[lookup_id][location].extend(key_all_layers)
                         res = end
                         continue
                     return res
@@ -820,16 +822,16 @@ class LMCacheEngine:
                     keys.append(chunk_info[2])
 
                 # hit chunks by prefix matching
-                hit_chunks = self.storage_manager.batched_contains(
+                hit_chunks, block_mapping = self.storage_manager.batched_contains(
                     keys, search_range, pin
                 )
+                if pin and block_mapping:
+                    assert lookup_id is not None, (
+                        "lookup_id is required when pin is True"
+                    )
+                    self.lookup_pins[lookup_id] = block_mapping
                 for idx, (start, end, key) in enumerate(chunk_info_list):
                     if idx < hit_chunks:
-                        if pin:
-                            assert lookup_id is not None, (
-                                "lookup_id is required when pin is True"
-                            )
-                            self.lookup_pins[lookup_id].append(key)
                         res = end
                         continue
                     return res
@@ -858,7 +860,7 @@ class LMCacheEngine:
 
         num_tokens = self.lookup(
             tokens,
-            search_range=old_position,
+            search_range=[old_position],
             lookup_id=event_id,
             pin=True,
         )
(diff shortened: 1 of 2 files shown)


csrc/mem_alloc.cpp
```diff
@@ -68,6 +96,72 @@ uintptr_t alloc_pinned_numa_ptr(size_t size, int node) {
   return reinterpret_cast<uintptr_t>(ptr);
 }
 
+uintptr_t alloc_pinned_numa_policy_ptr(size_t size,
```
YaoJiayi: Can we add a doc string here?

YaoJiayi: Also, I think this function overlaps with `alloc_pinned_numa_ptr`? Can we merge these two together to make it cleaner?

docs/source/api_reference/configurations.rst
```diff
@@ -78,7 +78,24 @@ Basic cache settings that control the core functionality of LMCache.
...
+
+   .. code-block:: yaml
+
+      numa_mode: "manual"
+      extra_config:
+        gpu_to_numa_mapping:
+          "0": 0
+          "1": 1
+        numa_policy: "bind"
+        numa_nodes_fallback: [0]
```
YaoJiayi: Thanks for adding the doc:)

examples/benchmarks/numa_pinned_backend_demo.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+        print(f"    {desc}")
+        if result.placement:
+            print(f"    placement: {result.placement}")
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
YaoJiayi: Can you explain a bit what this Benchmark is doing?

lmcache/v1/memory_management.py
```diff
@@ -300,8 +307,16 @@ def can_evict(self) -> bool:
 def _allocate_cpu_memory(
     size: int,
     numa_mapping: Optional[NUMAMapping] = None,
+    numa_placement: Optional[NUMAPlacement] = None,
```
YaoJiayi: Why do we need two variables for `numa_mapping` and `numa_placement`? Can they be fit into a single variable?

lmcache/v1/system_detection.py
```diff
@@ -46,17 +46,31 @@ def _read_from_config(config) -> NUMAMapping:
...
+        gpu_to_numa_mapping = {}
+        try:
+            for gpu_id, node_id in raw_mapping.items():
+                gpu_key = int(gpu_id)
+                numa_val = int(node_id)
+                gpu_to_numa_mapping[gpu_key] = numa_val
+        except (TypeError, ValueError) as exc:
+            raise ValueError(
+                "gpu_to_numa_mapping keys and values must be integers"
+            ) from exc
```
YaoJiayi: The changes in this file seem a bit unnecessary? I think we want to have some assertions to crash the program if necessary.

YaoJiayi (commented):
Thanks @DongDongJu ! I left a few questions:)

sammshen:
@DongDongJu thanks for the PR! pre-commit guide here to pass code quality: 
https://docs.lmcache.ai/developer_guide/contributing.html :)

csrc/mem_alloc.cpp
DongDongJu: I agree with your opinion. changes made for this in https://github.com/DongDongJu/LMCache/commit/22f92ce0b52a17f1bfb155a9c2feef84933f402f

examples/benchmarks/numa_pinned_backend_demo.py
DongDongJu: This file simply times how fast LMCache can copy data into its pinned CPU memory while you turn NUMA placement on and off, so you can see whether binding memory to specific
  sockets (like nodes 0/1/2) helps.

lmcache/v1/memory_management.py
DongDongJu: No we need both for accurate numa setting.
numa_mapping is the GPU→NUMA lookup (NUMAMapping) detected earlier in the stack when _select_numa_placement receives it, the helper turns the current CUDA device’s mapped node into a bind-style placement if no explicit override is present.
numa_placement is already a resolved request (NUMAPlacement) describing which NUMA nodes to use and under which policy (e.g., bind vs preferred); it usually comes from higher-level config logic like the backend’s fallback list, and it’s passed through unchanged if provide.

lmcache/v1/system_detection.py
DongDongJu: Thats make sense. It resolved by https://github.com/LMCache/LMCache/pull/1782/commits/8e6bec7930fce81ccc87ae3b9100b1bfb15b552e

DongDongJu:
@YaoJiayi Could you check the comments in reviews?

Issue #1794, opened 2025-10-14
[Bug] StorageManager layerwise_batched_get fails in sync mode


yoo-kumaneko:
LMCache/lmcache/v1/storage_backend/storage_manager.py
```
class StorageManager:
    def post_init(self, **kwargs) -> None:
        if "async_lookup_server" in kwargs:
            assert not self.config.save_unfull_chunk, (
                "save_unfull_chunk should be automatically set to False when using "
                "async loading."
            )
            self.async_lookup_server = kwargs.pop("async_lookup_server")
            self.async_serializer = AsyncSerializer(self.allocator_backend, self.loop)
```

```
    def layerwise_batched_get(
        self,
        keys: List[List[CacheEngineKey]],
        location: Optional[str] = None,
    ) -> Generator[Future, None, None]:
        """
        Non-blocking function to get the memory objects into the storages
        in a layerwise manner.
        Do not store if the same object is being stored (handled here by
        storage manager) or has been stored (handled by storage backend).

        :param List[List[CacheEngineKey]] keys: The keys to get. The first
            dimension corresponds to the number of layers, and the second
            dimension corresponds to the number of chunks.

        :return: A generator that yields a future for each layer.
        """
        if location is None:
            location = "LocalCPUBackend"

        for keys_multi_chunk in keys:
            # Retrieve all chunks for one layer
            backend = self.storage_backends[location]
            # TODO(Jiayi): need to make async loading and layerwise compatible
            task = asyncio.run_coroutine_threadsafe(
                self.async_serializer.run(
                    backend.batched_get_non_blocking(
                        "fake_lookup_id", keys_multi_chunk
                    ),
                    len(keys_multi_chunk),
                ),
                self.loop,
            )
            yield task
```

async_serializer is not initialized when async loading isn't enabled, but layerwise_batched_get tries to use it unconditionally.

yoo-kumaneko:
@sammshen Hey Sam, could you please take a look at this issue?

sammshen:
let's use @DongDongJu's PR. I left some comments on it and I think it will fix the issue

yoo-kumaneko:
@sammshen Got it. Great to see someone has started to work on it.

Issue #1765, opened 2025-10-10
[Bug] AttributeError: 'StorageManager' object has no attribute 'async_serializer'


KevinCheung2259:
**Label**
Please label this issue as "bug" and add any other relevant labels for easier categorization under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/627)

**Describe the bug**
```
(EngineCore_DP0 pid=116) ERROR 10-10 01:56:48 [dump_input.py:79] Dumping scheduler stats: SchedulerStats(num_running_reqs=54, num_waiting_reqs=237, step_counter=0, current_wave=0, kv_cache_usage=0.2942583211948605, prefix_cache_stats=PrefixCacheStats(reset=False, requests=2, queries=5262, hits=2560), spec_decoding_stats=None, num_corrupted_reqs=0)
...
(EngineCore_DP0 pid=116) AttributeError: 'StorageManager' object has no attribute 'async_serializer'
```
LMCache crashes in multi-user multi-turn dialogue scenarios due to StorageManager missing async_serializer, causing EngineCore to fail.

**To Reproduce**
Steps to reproduce:
1. Configure LMCache as follows:
```
chunk_size: 256
local_cpu: True
max_local_cpu_size: 200
use_layerwise: true
```
2. Start vllm service (H100):
```
LMCACHE_CONFIG_FILE=example.yaml vllm serve xxx-10B-dense-model --max-model-len 10000 --port 8000 --quantization fp8 --max-num-batched-tokens 1024 --kv-cache-dtype fp8 --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
```
3. Run a benchmark script (such as https://github.com/vllm-project/production-stack/blob/main/benchmarks/multi-round-qa/multi-round-qa.py)

**Expected behavior**
LMCache should run stably in multi-user multi-turn scenarios without crashing or raising AttributeError.

**Screenshots**
If relevant, please attach screenshots or error logs.

**Desktop (please complete the following information):**
- OS: [e.g. Ubuntu 22.04]
- Python: [e.g. 3.12]
- Torch: [e.g. 2.1.0]

**Smartphone (please complete the following information):**

**Additional context**
Add any other context, related code, or commit links here.

DongDongJu:
Hopely it solved by this.
#1795

KevinCheung2259:
Thanks. It looks good to me; I’ll run the tests again tomorrow.

[...]

Issue #1780, opened 2025-10-13
[Bug]: Performance Degradation When enable_async_loading: True Under High CPU Memory Usage


KevinCheung2259:
## Description
When `enable_async_loading: True` is set, in scenarios with multiple users interacting in multi-turn conversations under high concurrency, if CPU memory usage becomes saturated, the GPU utilization drops sharply from 90% to 60%-70% and overall performance degrades significantly.

## To Reproduce
Steps to reproduce:
1. Configure LMCache as follows:
```
chunk_size: 256
local_cpu: True
max_local_cpu_size: 200
use_layerwise: true
```
2. Start vllm service (H100):
```
LMCACHE_CONFIG_FILE=example.yaml vllm serve xxx-10B-dense-model --max-model-len 10000 --port 8000 --quantization fp8 --max-num-batched-tokens 1024 --kv-cache-dtype fp8 --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
```
3. Run a benchmark script (such as https://github.com/vllm-project/production-stack/blob/main/benchmarks/multi-round-qa/multi-round-qa.py)

## Expected Behavior
GPU utilization should remain high and performance should not degrade even when CPU memory is fully utilized.

## Screenshots
<img alt="Performance dashboard" width="1585" src="https://github.com/user-attachments/assets/4e9f58c6-1a87-45de-9324-05f34f2cea2a" />
<img width="1585" height="617" alt="Image" src="https://github.com/user-attachments/assets/e34fa2cc-0ea8-4894-86e2-ca43a68882f5" />
<img alt="GPU metrics snapshot" width="1898" src="https://github.com/user-attachments/assets/0515f87e-8f24-44aa-9935-2b5f3d915f7b" />

## Environment
 - Device: Server/Workstation running LMCache
 - OS: Linux (please specify version)
 - GPU: NVIDIA H100 80GB HBM3
 - CUDA Driver: 12.2
 - LMCache version: (please specify)
 - Other info: Driver Version 535.129.03, concurrency level: high

## Additional Context
This issue only occurs when `enable_async_loading` is enabled. In other configurations, GPU utilization remains stable even under high CPU memory usage. See attached performance dashboards and GPU metrics for reference.

## Resolution
- [ ] I would be interested in resolving this bug.

sammshen:
Does the GPU utilization drop temporarily sometimes or is it consistently 60-70% versus 90% when not using async loading?

[...]

PR #1795, opened 2025-10-14
Guarded the async serializer usage so we only wrap the backend load c…

DongDongJu:
Bug fix for #1794 and #1765.


lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -436,20 +437,15 @@ def layerwise_batched_get(
...
-                    backend.batched_get_non_blocking(
-                        "fake_lookup_id", keys_multi_chunk
-                    ),
-                    len(keys_multi_chunk),
-                ),
-                self.loop,
-            )
+            coro = backend.batched_get_non_blocking("fake_lookup_id", keys_multi_chunk)
+            serializer = self.async_serializer
+            if serializer is not None:
```
sammshen: Thanks for the fix @DongDongJu . instead of checking for if the serializer is None here, can we always intialize it (not only in the async codepath). Layerwise codepath also needs async serializer because multiple layers loading at once shoudl ensure that we don't deadlock the cpu as well

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -224,6 +224,7 @@ def __init__(
         self.event_manager = event_manager
 
         self.async_lookup_server: Optional["LMCacheAsyncLookupServer"] = None
+        self.async_serializer: Optional[AsyncSerializer] = None
```
sammshen, line 227: instead of None here, we can change post_init to: 

```python
    def post_init(self, **kwargs) -> None:
        if "async_lookup_server" in kwargs:
            assert not self.config.save_unfull_chunk, (
                "save_unfull_chunk should be automatically set to False when using "
                "async loading."
            )
            self.async_lookup_server = kwargs.pop("async_lookup_server")
        self.async_serializer = AsyncSerializer(self.allocator_backend, self.loop)
        
```

DongDongJu: @sammshen Thanks for the review. I agree. this approach is better. also I added asserts for just in case for later.
By the way, is there any chance that post_init could be called twice at runtime?

sammshen: yes definitely, IIRC, every start_load_kv()

DongDongJu: Okay good to know. I will check the asyncserializer behavior more deeply after this bugfix. for #1780

KuntaiDu:
Good work! looking forward to the merge.

sammshen (approved):
LGTM!

sammshen:
sorry I just realized something, I think the storage_manager post_init will not be called right now in your code: 
```python
    def post_init(self, **kwargs) -> None:
        if "async_lookup_server" in kwargs:
            self.async_lookup_server = kwargs.pop("async_lookup_server")
            self.storage_manager.post_init(async_lookup_server=self.async_lookup_server)
        if not self.post_inited:
            logger.info("Post-initializing LMCacheEngine")
            self.gpu_connector.initialize_kvcaches_ptr(**kwargs)
            self.post_inited = True
            
 ```

**Comment by DongDongJu** [comment]
Thanks for the catch. Yes that was the reason why I asked the question about how many time post_init will call in runtime. 
I added into cache engine init then it always init even Layerwise codepath w/o async code path.

**Comment by sammshen** [comment]
@KevinCheung2259 is your use case resolved?

**Comment by KevinCheung2259** [comment]
Sorry for the late, it works nice on in my exp, thanks for your contribution !!! @sammshen

**Review by YaoJiayi — APPROVED** [review]
lgtm!

File: lmcache/v1/storage_backend/storage_manager.py [hunk 3]
```diff
@@ -238,7 +239,7 @@ def post_init(self, **kwargs) -> None:
                 "async loading."
             )
             self.async_lookup_server = kwargs.pop("async_lookup_server")
-            self.async_serializer = AsyncSerializer(self.allocator_backend, self.loop)
+        self.async_serializer = AsyncSerializer(self.allocator_backend, self.loop)
```
maobaolong, line 242: Clipboard_Screenshot_1761033530

This will introduce amount of logs for each start_load_kv, especially in the decode node.

maobaolong: Is this ident added by accident? @DongDongJu

DongDongJu: No we want to make sure that async AsyncSerializer always initialized in post_init. We need better whole flow I believe.

maobaolong: @DongDongJu But for now, there are 2 line logs for each prefill/decode step...

DongDongJu: @maobaolong I will hotfix this after the lmcache meetup.

DongDongJu: @maobaolong Let me open another pr for this issue.

sammshen: the post_init reptition is patched in: https://github.com/LMCache/LMCache/pull/1852

PR #1800, opened 2025-10-15
[metrics] add request_cache_hit_rate metric

chunxiaozheng:
add a Histogram metric `lmcache:request_cache_hit_rate`, indicates the distribution of hit rates for each request.

### DIFF
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -959,7 +959,7 @@ class PrometheusLogger:
         ]
         self.histogram_request_cache_hit_rate = self._histogram_cls(
             name="lmcache:request_cache_hit_rate",
-            documentation="Reqeust cache hit rate",
+            documentation="Request cache hit rate",
             labelnames=labelnames,
             buckets=request_cache_hit_rate,
         )
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -74,16 +74,15 @@ class LMCacheStats:
     p2p_time_to_transfer: List[float]
     p2p_transfer_speed: List[float]  # Tokens per second
 
-    # request cache hit rate
-    interval_request_cache_hit_rate: List[float]
+    # request lookup hit rates
+    interval_lookup_hit_rates: List[float]
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


maobaolong (approved):
LGMT, thanks for contribute for this useful metrics.

Another high-level issue discussed offline with @chunxiaozheng , but it is not belong to the scope of this PR, will raise a discussion future.

chunxiaozheng:
hi, @ApostaC could you help take a look?

lmcache/observability.py
```diff
@@ -74,11 +74,21 @@ class LMCacheStats:
     p2p_time_to_transfer: List[float]
     p2p_transfer_speed: List[float]  # Tokens per second
 
+    # request cache hit rate
+    interval_request_cache_hit_rate: List[float]
```
ApostaC: Can we rename it to `interval_lookup_hit_rates`? "lookup" indicates this metric is calculated by lookup results, and "hit_rates" indicates that this is a list of values.

lmcache/observability.py
```diff
@@ -74,11 +74,21 @@ class LMCacheStats:
...
+
 
 @dataclass
 class LookupRequestStats:
     num_tokens: int
     hit_tokens: int
+    start_time: float
+    end_time: float
```
ApostaC: Why do we add the start_time and end_time here? I didn't see where we used it. How about changing it to a boolean flag (something like `is_finished`).

If you want to add another histogram metric of lookup time, I'm also open to that.

ApostaC (commented):
Otherwise LGTM!

lmcache/observability.py
chunxiaozheng: done!

lmcache/observability.py
chunxiaozheng: done!

chunxiaozheng:
> Otherwise LGTM!

@ApostaC Thanks for your review! I have updated, could you help take a look again?

ApostaC (approved):
LGTM! Thanks!

Issue #1807, opened 2025-10-15
KV Cache Inconsistency in Mooncake with PP=2, TP=1 Configuration - Rank1 Cache Miss Due to skip_leading_tokens Issue


tianlang-wq:
Environment Details
Model Deployment: PP=2, TP=1

KV Cache: Mooncake

Versions:

lmcache=0.3.6

vllm=0.10.2

Problem Description
During inference with pipeline parallelism (PP=2), we're observing inconsistent KV cache behavior between ranks:

Rank0: Successfully hits cache and properly calls lmcache_engine.store

Rank1: Fails to hit cache and never calls lmcache_engine.store

Expected Behavior
With pipeline parallelism, both Rank0 and Rank1 should independently manage and store their respective KV cache segments. Both ranks should call lmcache_engine.store during the wait_for_save method.

Current Behavior & Root Cause Analysis
Through debugging, we identified the issue in the wait_for_save method:

Problem Location: wait_for_save method in Mooncake KV cache implementation

Root Cause:
In Rank1, the condition skip_leading_tokens == len(token_ids) always evaluates to True, causing the request to be skipped entirely. This prevents Rank1 from calling lmcache_engine.store.

Suspected Issue:
It appears that skip_leading_tokens might be using shared context between Rank0 and Rank1. When Rank0 sets skip_leading_tokens, this value seems to be affecting Rank1's execution context.

Evidence
Rank0: Properly executes cache storage logic

Rank1: Always hits the skip condition and bypasses cache storage

Cache Impact: Only Rank0's cache segments are stored, leading to incomplete cache entries

Questions
Is this intended behavior, or is it a bug in the pipeline parallelism implementation?

Should each rank maintain independent skip_leading_tokens context in pipeline parallelism?

If this is a bug, what's the recommended fix - should we ensure per-rank context isolation?

Reproduction Steps
Deploy model with PP=2, TP=1 configuration

Enable Mooncake KV cache with lmcache 0.3.6

Make inference requests

Observe that only Rank0 triggers cache storage while Rank1 skips

sammshen:
Pipeline parallelism had a small bug in lmcache, will be patched soon, thanks for your issue

tianlang-wq:
Okay, thank you very much

tianlang-wq:
@sammshen hi

I reviewed the changes in https://github.com/LMCache/LMCache/pull/1813
[...]

PR #1813, opened 2025-10-17
[Feat]: Pipeline Parallelism

### LINKED ISSUE #1770: [BUG] When Set TP=2 PP=2 lmcache meet  LMCache ERROR: The number of retrieved tokens is less than the expected number of tokens!
vllm lmcache meet issue like below:
(EngineCore_DP0 pid=1642) (RayWorkerWrapper pid=270, ip=172.30.7.195) [2025-10-10 19:24:43,546] LMCache ERROR: The number of retrieved tokens is less than the expected number of tokens! This should not happen! (vllm_v1_adapter.py:882:lmcache.integration.vllm.vllm_v1_adapter)
(EngineCore_DP0 pid=1642) (RayWorkerWrapper pid=270, ip=172.30.7.195) [2025-10-10 19:24:43,546] LMCache ERROR: Num retrieved tokens: 0, num expected tokens: 9 (vllm_v1_adapter.py:886:lmcache.integration.vllm.vllm_v1_adapter)

My env: 
i setup an env with 2nodes, each nodes with 2 gpus;

vllm serve /models/Qwen/Qwen3-1.7B \
                    --served-model-name tp-1p1d-pgs-2 \
                    --host "0.0.0.0" \
                    --port "8000" \
                    --uvicorn-log-level warning \
                    --gpu-memory-utilization 0.8 \
                    --swap-space 0 \
                    --trust-remote-code \
                    --distributed-executor-backend ray \
                    --tensor-parallel-size 2 \
                    --pipeline-parallel-size 2 \
                    --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}' \
                    --max-model-len 4096 \
                    --max-num-seqs 5

I use mooncake as remote backend; lmcache envs:
# env | grep LMCACHE
LMCACHE_LOCAL_CPU=False
LMCACHE_MAX_LOCAL_CPU_SIZE=5
LMCACHE_REMOTE_URL=mooncakestore://mooncake-master:50051/
LMCACHE_PRE_CACHING_HASH_ALGORITHM=sha256_cbor_64bit
LMCACHE_EXTRA_CONFIG={"local_hostname": "0.0.0.0","metadata_server": "etcd://etcd:2379","master_server_address": "etcd://etcd:2379","protocol": "rdma","transfer_timeout": 360,"global_segment_size": 0,"local_buffer_size": 1073741824,"save_chunk_meta": "False"}

2、when i use 1 nodes with 2gpus and vllm with 
                    --tensor-parallel-size 2 \
                    --pipeline-parallel-size 1 \
read and write  mooncake remote backend all right；

3、 i found in lmcache， 
```python
def _make_key(self, chunk_hash: int, fmt: str) -> CacheEngineKey:
        return CacheEngineKey(
            fmt,
            self.metadata.model_name,
            self.metadata.world_size,
            self.metadata.worker_id,
            chunk_hash,
        )
```
haven't contains PP rank info, only container worker_id with  "parallel_cfg.rank,",  i think maybe lmcache should contains TP and PP rank info when generate CacheEngineKey.

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1770#issuecomment-3405833185 https://github.com/LMCache/LMCache/issues/1807

This adds PP to the ranks of the lookup and async lookup clients/servers

@maobaolong @chunxiaozheng IIUC, MLA models should have one KV store per PP rank. This should probably be specified for multiple lookup servers

### DIFF
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -474,6 +474,15 @@ def _create_config_class():
 def _validate_config(self):
     """Validate configuration"""
 
+    # auto-adjust save_unfull_chunk for async loading to prevent CPU fragmentation
+    if self.enable_async_loading or self.use_layerwise:
+        logger.warning(
+            "Automatically setting save_unfull_chunk=False because "
+            "enable_async_loading=True or use_layerwise=True to prevent "
+            "CPU memory fragmentation"
+        )
+        self.save_unfull_chunk = False
+
     if self.enable_p2p:
         assert self.enable_controller
         assert self.controller_pull_url is not None
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -473,14 +473,6 @@ def _create_config_class():
 
 def _validate_config(self):
     """Validate configuration"""
-    # auto-adjust save_unfull_chunk for async loading to prevent CPU fragmentation
-    if self.enable_async_loading or self.use_layerwise:
-        logger.warning(
-            "Automatically setting save_unfull_chunk=False because "
-            "enable_async_loading=True or use_layerwise=True to prevent "
-            "CPU memory fragmentation"
-        )
-        self.save_unfull_chunk = False
 
     if self.enable_p2p:
         assert self.enable_controller
@@ -705,6 +697,15 @@ def _update_config_from_env(self):
                     f"Failed to parse {get_env_name(name)}={raw_value!r}: {e}"
                 )
                 # Keep existing value if conversion fails
+
+    # auto-adjust save_unfull_chunk for async loading to prevent CPU fragmentation
+    if self.enable_async_loading or self.use_layerwise:
+        logger.warning(
+            "Automatically setting save_unfull_chunk=False because "
+            "enable_async_loading=True or use_layerwise=True to prevent "
+            "CPU memory fragmentation"
+        )
+        self.save_unfull_chunk = False
 
     return self
 
--- a/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
@@ -53,7 +53,9 @@ class LMCacheAsyncLookupClient(LookupClientInterface):
         rpc_port = vllm_config.kv_transfer_config.get_from_extra_config(
             "lmcache_rpc_port", 0
         )
+        self.pipeline_parallel_size = vllm_config.parallel_config.pipeline_parallel_size
         self.tensor_parallel_size = vllm_config.parallel_config.tensor_parallel_size
+        self.num_ranks = self.tensor_parallel_size * self.pipeline_parallel_size
         self.mla_lookup_server_worker_id = config.get_mla_lookup_server_worker_id(
             metadata.use_mla
         )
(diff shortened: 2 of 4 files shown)


chunxiaozheng:
@sammshen from the vllm code, `self.world_size = self.pipeline_parallel_size * self.tensor_parallel_size`, I think your description is correct.

lmcache/v1/config.py
```diff
@@ -706,6 +707,15 @@ def get_env_name(attr_name: str) -> str:
                 )
                 # Keep existing value if conversion fails
 
+    # auto-adjust save_unfull_chunk for async loading to prevent CPU fragmentation
+    if self.enable_async_loading or self.use_layerwise:
```
chunxiaozheng, line 711: Is this also related to `pp` configuration?

sammshen: not related to pp but just to avoid >50% fragmentation in async and alyerwise code path. previously, it was only catching the configuration file but this is to also catch teh environment variable

sammshen:
@maobaolong @chunxiaozheng could you please help to review?

chunxiaozheng (approved):
lgtm

maobaolong:
@sammshen Thanks for this great feature, after that, we can support PP > 1 with LMCache.

But I have some question about this PR. Let's say tp4, pp2
- If PP implement parallelism by `layer`, which worker_id belongs to a group which have a full layer kvcache? `tp0,tp1` or `tp0,tp4`?
- If we enable PP = 2, and meanwhile we enable `save_only_first_rank` and `mla_lookup_server_worker_id` , what will happen?

Not sure this feature can works with the existing improvement optional configs related MLA.

chunxiaozheng:
@sammshen @maobaolong The relationship between `PP` and `TP` can be referred to in the following code comments in VLLM, from this relation, the  feature `save_only_first_rank` and `mla_lookup_server_worker_id` will make some error when `PP` is not 1. Therefore, I think we not only need to update the `lookup_server`, but also the feature related to these two parms.
Clipboard_Screenshot_1761225165

maobaolong:
@sammshen @chunxiaozheng As the above code,

For the existing mla save_only_first_rank, it would be much better if we call it `save_only_first_tp_rank`.

And if we set pp=4, then, we the first tp rank relevant worker id is [0, 8, 16, 24], so these four vllm worker(lmcache worker) should be treat as first tp rank and responsible for save kvcache.

sammshen:
> @sammshen Thanks for this great feature, after that, we can support PP > 1 with LMCache.
> 
> But I have some question about this PR. Let's say tp4, pp2
> 
> * If PP implement parallelism by `layer`, which worker_id belongs to a group which have a full layer kvcache? `tp0,tp1` or `tp0,tp4`?
> * If we enable PP = 2, and meanwhile we enable `save_only_first_rank` and `mla_lookup_server_worker_id` , what will happen?
> 
> Not sure this feature can works with the existing improvement optional configs related MLA.

PP is major and TP is minor so with tp4, pp2 it wil be worker 0 and worker 3. Thanks @chunxiaozheng for showign the vllm example

sammshen:
> @sammshen @chunxiaozheng As the above code,
> 
> For the existing mla save_only_first_rank, it would be much better if we call it `save_only_first_tp_rank`.
> 
> And if we set pp=4, then, we the first tp rank relevant worker id is [0, 8, 16, 24], so these four vllm worker(lmcache worker) should be treat as first tp rank and responsible for save kvcache.

agreed, do you want to do this automatically for MLA? I see there is already an interface to specify which workers will have lookup servers?

maobaolong:
@sammshen The interface it not fully suitable, since existing config can only set one workerid, but cannot set more than one.

And it is not very difficult to refactor existing config and interface to support config more one workerid, but it still hard to understand and set the correct worker ids for PP > 1. So it would be friendly to use if we have a composite/group concept can include a set of worker_id belongs to one TP rank.

sammshen:
right now there is no change to MLA yet (if PP is enabled it will just look at worker 0 which is technically not safe but same as before). we can do the composite/group idea in the next PR

chunxiaozheng:
@sammshen @maobaolong I have supported set list of lookup server worker ids(https://github.com/LMCache/LMCache/pull/1872), after current PR is merged, I will rebase and fix the conflict.

chunxiaozheng:
@sammshen @maobaolong What is the current status of PR? If there are no issues with this part, can we first merge the PR and then submit the PP related functions separately later? What do you think?

maobaolong (approved):
LGTM. @sammshen  Thanks for this feature! And @chunxiaozheng thank you for your review.

PR #1574, opened 2025-09-11
[feat] Add a request level skip_save parameter

maobaolong:
# Highlight

This PR is co-authored by @xshwu 

# Motivation

In our scenario, some of the request doesn't need to save kvcache, so we need to skip the kvcache saving for this kind of request.

# Test
Tested set `mcache.skip_save` to true can prevent save kvcache, otherwise, as normal, it save kvcache for the request.

```
curl http://localhost:8000/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
    "model": "vllm_cpu_offload",
    "messages": [
        {
            "role": "user",
            "content": "Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?Hello, how are you?"
        }
    ],
    "max_tokens": 10,
    "temperature": 0,
    "top_p": 0.95,
    "stream": true,
    "stream_options": {
         "include_usage": true
    },
    "kv_transfer_params":{
            "lmcache.skip_save":true
        }
    }'
```

PR #1816, opened 2025-10-17
[feat]: no store cache for v1 adapter

### LINKED ISSUE #1792: The following fields were present in the request but ignored: {'store_cache'}
I am trying to turn-off the cache storage, but I keep getting:
```
The following fields were present in the request but ignored: {'store_cache'}
```
Is it supported via `/v1/chat/completions` ? The official example uses `v1/completions`
```Python
def get_payload(text, img_data, store_cache=True):
    messages = [get_message(text, img_data)]
    payload = {
        "model": model_name,
        "messages": messages,
        "max_tokens": 128,
        "store_cache": store_cache,
    }
    return payload

response = client.request(
    method="POST",
    endpoint="/v1/chat/completions" ,
    body=json.dumps(payload),
    headers={"Content-Type": "application/json"}
)
```

Thank you!

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1792

updated example as well

### DIFF
--- a/examples/cache_interface/README.md
+++ b/examples/cache_interface/README.md
@@ -1,7 +1,8 @@
 # User Controllable Caching
 This is an example to demonstrate user controllable caching (e.g., specify whether to cache a request or not).
+
 ## Prerequisites
-Your server should have at least 1 GPU.  
+Your server should have at least 1 GPU.
 
 This will use the port 8000 for 1 vllm.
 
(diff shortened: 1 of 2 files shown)


maobaolong:
@sammshen Can you take a check whether this PR is duplicated to #1574 ?

sammshen:
@maobaolong good catch. @mobicham can you try the example that @maobaolong provided?

PR #1828, opened 2025-10-20
[optimize] reduce calls to contains for sync path

chunxiaozheng:
If lookup result is `True`, and there is only one actual storage backend, there is no need to check `contains` again.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1146,11 +1146,12 @@ class LMCacheEngine:
                 pass
             else:
                 # NOTE: key should always be in the lookup cache once we support it.
+                # TODO: use lookup_cache to skip the contains
                 if (
                     skip_contains_check
-                    and len(self.storage_manager.actual_storage_names) == 1
+                    and len(self.storage_manager.non_allocator_backends) == 1
                 ):
-                    location = self.storage_manager.actual_storage_names[0]
+                    location = self.storage_manager.non_allocator_backends[0]
                 else:
                     location = self.storage_manager.contains(key)
                 if location is None:
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1131,6 +1131,7 @@ class LMCacheEngine:
         ):
             assert isinstance(key, CacheEngineKey)
 
+            location = None
             if key in self.lookup_cache:
                 # TODO(Jiayi): we can reduce the number of `contains` calls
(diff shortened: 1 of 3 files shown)


chunxiaozheng:

maobaolong (approved):
LGTM. Thanks for your contribution

sammshen (dismissed):
LGTM

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -750,6 +753,23 @@ def memcheck(self) -> bool:
                 return False
         return True
 
+    def get_actual_storage_names(self) -> List[str]:
```
sammshen: can we change this to `get_non_allocator_backends()`?

lmcache/v1/cache_engine.py
```diff
@@ -1111,21 +1111,35 @@ def _process_tokens_internal(
...
                 # by checking the lookup cache first (should be updated in `lookup`)
                 pass
             else:
-                # NOTE: key should always be in the lookup cache once
-                # we support it.
-                location = self.storage_manager.contains(key)
+                # NOTE: key should always be in the lookup cache once we support it.
+                if (
+                    skip_contains_check
+                    and len(self.storage_manager.actual_storage_names) == 1
```
sammshen: in the future, to avoid this, we can store the location on the first contains in `get_num_new_matched_tokens()` (which should at least pin the local cpu and the disk) and then always skip this contains call. WDYT @maobaolong @chunxiaozheng

sammshen: maybe can keep this as a comment for now

chunxiaozheng: > in the future, to avoid this, we can store the location on the first contains in `get_num_new_matched_tokens()` (which should at least pin the local cpu and the disk) and then always skip this contains call. WDYT @maobaolong @chunxiaozheng

@sammshen That's right, we can use `lookup_cache` here, but `lookup_cache` is not implemented now, I will add comment here.

lmcache/v1/storage_backend/storage_manager.py
chunxiaozheng: > can we change this to `get_non_allocator_backends()`?

@sammshen Sure, I have updated! Thanks for your suggestion.

chunxiaozheng:
@sammshen could you help take a look again?

sammshen (approved):
LGTM

Issue #1830, opened 2025-10-20
[RFC][Core] Minimal MemoryPool Layer for engine-owned allocations


DongDongJu:
Follow up with #1713

  ### Summary
  Introduce a MemoryPool facade so every CPU‐side MemoryObj allocation runs through a single engine-owned path. Connectors and storage backends stop touching the allocator directly and instead call the new StorageManager.request_memory(...) / release_memory(...) helpers, which delegate to the existing MixedMemoryAllocator. (No allocator algorithm changes in this phase.)

  ### Background

  - #1713 asked that StorageBackend/RemoteConnector stop allocating so the CacheEngine can own frees—especially on exception/error paths.
  - We’ve chased repeated leaks and refcount bugs (FS #1656, S3 #1495, RemoteBackend #650, hierarchical storage #631, Redis memoryview #662).
  - FSConnector must keep its direct-into-MemoryObj read path (see #740); the pool only centralises construction, so the zero-copy path survives.
  - A long-standing footgun: creating LocalCPUBackend when max_local_cpu_size == 0 produced a 0-byte pinned buffer and torch.frombuffer crashes (#1768).

  ### Goals

  - Tighten ownership: all CPU allocations borrow/return via MemoryPool.
  - Keep the diff small (one new module + light wiring).

  ### Non-Goals (Phase 1)

  - New allocator strategies (buddy/slab), THP or NUMA policy tweaks, GPU/HBM changes, or new public APIs beyond the pool facade.

  ### Proposal

  1. New module lmcache/v1/memory_pool.py
      - MemoryPool.borrow(PoolRequest) delegates to the existing allocator.
      - MemoryPool.release(obj) funnels back through the allocator’s free.
      - lease(pool, req) context manager for exception-safe borrowing.
  2. Wire-up
      - LocalCPUBackend builds a pool alongside MixedMemoryAllocator and exposes it.
      - StorageManager attaches the pool and provides:
        request_memory(shape=..., dtype=..., fmt=..., tag="...") -> MemoryObj
        release_memory(mem_obj)
      - LocalCPUBackend now raises LocalCPUDisabledError when max_local_cpu_size <= 0; PD configs skip instantiating the CPU backend, while other configs fail fast instead of creating a zero-length buffer.

  Looking forward to feedback on the always-on pool and the new request/release surface.

sammshen:
sorry for the delay, will take a closer look today

DongDongJu:
[...]

Issue #1713, opened 2025-09-28
[RPC] StorageBackend and RemoteConnector should not allow allocate memory.


maobaolong:
# Background

The developers of connector and backend are not familiar with how memory `allocate` and `free`, it is dangerous to make it possible to allocate or free memory casually.

The following case are the reason why a introduce this RPC.
- https://github.com/LMCache/LMCache/pull/1656
- https://github.com/LMCache/LMCache/pull/1495
- https://github.com/LMCache/LMCache/pull/650
- #631 
- #662

# Proposal

We should invent a way to ensure memory should be freed properly even the backend or connector forget or cannot free it.

# Approach: high-level

The memory should be maintained by CacheEngine only, the `LocalCpuBackend` which is a memory maintainer should only be called by `CacheEngine` only, the other places, something like `FSConnector`, it is allowed to obtained a MemoryObj by request  the LocalCpuBackend of `CacheEngine`, but it is not allowed to allocate memoryObj by call `LocalCpuBackend` directly.

If we follow this way, all the memory objects allocated from the call chain from `CacheEngine->Backend->Connector` should be maintained by `CacheEngine.` and it should be freed immediately if there are any exception within the call chain, in short, `CacheEngine` is responsible for make sure memoryObj will never leak.

maobaolong:
@ApostaC @YaoJiayi @sammshen @chunxiaozheng and any others, what about your idea?

maobaolong:
Has been discussed with @chunxiaozheng , and sync the conclusion here.

There are two approach come out from our side.
- A: Aim to control the scope and lines of changes, only protect the remote connectors avoid memory leak. Just use a `with` and a wrapper of `LocalCpuBackend` which maintains the `MemoryObj` allocated and responsible for `ref_count_down` for the `orphan` MemoryObj after `with` block exit, note that we cannot make sure the third-party `Connector` always did the right thing, instead, we should think about the worse things, something like swallow exception, allocate memoryObj and lost reference...
- B: Aim to solve the memory leak issue thoroughly. Whatever the memory_obj is allocated within any backend or any connector or `StorageManager` or `CacheEngine`, we cannot make sure there is no bug this this long calling chain, but we can maintains the allocated memoryObj by a map `operation_id` -> `List<MemoryObj>` for each `retrieve` and `store` operation, and also use a `with` block to make sure before operation `retrieve` or `store` end, have a chance to `ref_count_down` the memory_objs allocated or `ref_count_up` by the current operation.

[...]

PR #1831, opened 2025-10-20
[RFC][Core] centralized memory pool layer phase 1

DongDongJu:

  - Refs #1830 and #1713.
  - Adds lmcache/v1/memory_pool.py, providing a MemoryPool/PoolRequest facade so every CPU MemoryObj is borrowed and released through the engine rather than through connectors/backends.
  - Wires the pool into LocalCPUBackend and StorageManager; introduces StorageManager.request_memory(...) / release_memory(...) as the sanctioned API for connectors.
  - Guards zero-capacity configurations via LocalCPUDisabledError, preventing creation of 0-byte pinned buffers while still allowing PD deployments to skip the CPU backend entirely.
  - Updates the FS connector sample path to use the pool and refreshes README highlights.
  - Adds tests/v1/test_memory_pool.py coverage for borrow/release accounting, exception safety, backend fallback, and the zero-size guard.

  Special notes:

  - MemoryPool is always-on: there is no feature flag. Please sanity-check the PD fallback path (allocate_from_pool vs. backend allocate), since that was the critical review finding.
  - Confirm that other backends relying on LocalCPUBackend (p2p, local disk, remote) still require a positive CPU budget; configuration errors now surface immediately.

  If applicable:

### DIFF
--- a/lmcache/v1/memory_pool.py
+++ b/lmcache/v1/memory_pool.py
@@ -1,13 +1,17 @@
 # SPDX-License-Identifier: Apache-2.0
+# Future
 from __future__ import annotations
 
+# Standard
 from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, Iterator, Optional, cast
 import threading
 
+# Third Party
 import torch
 
+# First Party
 from lmcache.logging import init_logger
 from lmcache.observability import LMCStatsMonitor, PrometheusLogger
 from lmcache.v1.memory_management import (
@@ -17,6 +21,7 @@ from lmcache.v1.memory_management import (
 )
 
 if TYPE_CHECKING:
+    # First Party
     from lmcache.v1.storage_backend.abstract_backend import AllocatorBackendInterface
 
 logger = init_logger(__name__)
@@ -41,15 +46,15 @@ class PoolRequest:
     def resolve_shape(self) -> torch.Size:
         if self.fmt == MemoryFormat.BINARY_BUFFER:
             if self.size is None and self.shape is None:
-                msg = "PoolRequest for BINARY_BUFFER expects size or shape"
-                raise ValueError(msg)
+                raise ValueError("PoolRequest for BINARY_BUFFER expects size or shape")
             if self.shape is not None:
                 return self.shape
             return torch.Size([self.size])  # type: ignore[arg-type]
 
         if self.shape is None or self.dtype is None:
-            msg = "PoolRequest requires shape and dtype for tensor allocations"
-            raise ValueError(msg)
+            raise ValueError(
+                "PoolRequest requires shape and dtype for tensor allocations"
+            )
         return self.shape
 
 
@@ -89,8 +94,9 @@ class MemoryPool:
                     req.fmt != MemoryFormat.BINARY_BUFFER
                 ):
                     if req.dtype is None:
-                        msg = "PoolRequest dtype must be set for backend allocations"
-                        raise ValueError(msg)
+                        raise ValueError(
+                            "PoolRequest dtype must be set for backend allocations"
+                        )
                     memory_obj = backend_allocate(
                         shape,
                         req.dtype,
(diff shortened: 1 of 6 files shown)


DongDongJu:
I tested it with what I have right now and works. but not sure for other case.
Can I know the normal flow for testing memory related?

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -445,13 +449,76 @@ def allocate(
...
+            dtype=dtype,
+            fmt=fmt_resolved,
+            tag=tag,
+            eviction=eviction,
+            busy_loop=busy_loop,
+        )
+        try:
+            return self.memory_pool.borrow(req)
+        except RuntimeError:
+            return None
```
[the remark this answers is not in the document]
DongDongJu: Resolved by this https://github.com/LMCache/LMCache/pull/1831/commits/9c1eb2edf0fa0791554ece234f49bcf8d716f0c2

lmcache/v1/memory_pool.py
```diff
@@ -0,0 +1,219 @@
...
+                        shape,
+                        req.dtype,
+                        req.fmt,
+                        eviction=req.eviction,
+                        busy_loop=req.busy_loop,
+                    )
+                else:
+                    memory_obj = self._allocator.allocate(shape, req.dtype, req.fmt)
+        else:
+            memory_obj = self._allocator.allocate(shape, req.dtype, req.fmt)
```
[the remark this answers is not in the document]
DongDongJu: resolved by https://github.com/LMCache/LMCache/pull/1831/commits/0d4b70064c2f63fb24c0a880fe4edca633275fe8

lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -143,12 +143,18 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
 
                     # Deserialize metadata and allocate memory
                     metadata = RemoteMetadata.deserialize(md_buffer)
-                    memory_obj = self.local_cpu_backend.allocate(
-                        metadata.shape, metadata.dtype, metadata.fmt
+                    memory_obj = self.local_cpu_backend.request_memory(
+                        metadata.shape,
+                        metadata.dtype,
+                        metadata.fmt,
+                        tag="fs/get",
```
sammshen, line 150: what's this tag for?

lmcache/v1/memory_pool.py
```diff
@@ -0,0 +1,216 @@
...
+        self._stats_monitor = LMCStatsMonitor.GetOrCreate()
+        self._prometheus_logger = PrometheusLogger.GetInstanceOrNone()
+
+    def borrow(self, req: PoolRequest) -> MemoryObj:
+        """Borrow a MemoryObj from the underlying allocator."""
+        shape = req.resolve_shape()
+        memory_obj: Optional[MemoryObj] = None
+        if self._backend is not None:
+            allocate_from_pool = getattr(self._backend, "allocate_from_pool", None)
+            if callable(allocate_from_pool):
```
sammshen, line 89: nit: checking None is probably better than checking callability

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -215,6 +214,14 @@ def __init__(
         if config.local_cpu:
             self.local_cpu_backend = self.storage_backends["LocalCPUBackend"]
 
+        backend_owner = cast(Any, self.allocator_backend)
```
sammshen, line 217: can we enforce a more elegant interface into `AllocatorBackendInterface`. Like what an allocator backend needs to support (let's encapsulate all the behavior instead of poking around into the backend) to be easily integrable with the Memory Pool

sammshen: for example, all of this code should not be cluttering the storage manager

sammshen:
The general idea behind this PR makes sense to me but I think we need to reconsider how exactly this is going to be designed. The pool facade is not nearly thin enough in my opinion. It should almost be a 1-1 wrapper around allocator backends. 

IIRC, the main use case was to have remote backend extensibility not have to manually manage memory. However, how does this PR solve this issue? we've centralized memory allocation from a single invocation point (the storage manager) but how does this simplify things going forward? my understanding from https://github.com/LMCache/LMCache/issues/1713 is that this allows us to have ref count up and down always be ina. context manager in a single component (the `lease` you wrote for example). right now "matching" ref count up and down are called in different components so integrating lease into the actual codebase will be a huge refactoring effort. 

Thanks for the RFC @DongDongJu

DongDongJu:
Thanks @sammshen. I will reopen this after revising it.

DongDongJu:
@sammshen Please check new one https://github.com/LMCache/LMCache/pull/1888. Thanks!

PR #1835, opened 2025-10-20
[core]add error handling in kv loading

ziruiliu:
**Why**:
In most implementation of backends/connectors, lookup() and get() are not atomic. There is a chance that lookup() returns OK and get_num_new_matched_tokens() returns full length of existing cache, but some of cached data is removed or become unreadable before get(). In such cases, lmcache returns shorter-than-expected KV cache to vLLM and may cause wrong output in following computation.

Fortunately vLLM just introduced feature (https://github.com/vllm-project/vllm/pull/19330) that recompute the failed blocks from KV connector. LMCache should have implement such function soon.

**What**:
Added method record_failed_blocks() in vllm_v1_adapter.py to compare `token_mask` and r`et_token_mask` to verify if there is any failure in loading kv. if so, generate a list of invalid block ID `_invalid_block_ids ` from `slot_mapping` 

@ApostaC please kindly review this since I found you are also the reviewer of PR https://github.com/vllm-project/vllm/pull/19330, thanks

**Test**:
to reproduce the issue and verify the change, see in the comment below

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -911,6 +911,11 @@ class LMCacheConnectorV1Impl:
                     )
                     self._invalid_block_ids.update(missing_blocks)
 
+            self._stats_monitor.update_interval_vllm_hit_tokens(
+                request.load_spec.vllm_cached_tokens
+            )
+            self._stats_monitor.update_interval_prompt_tokens(len(tokens))
+
     def record_failed_blocks(
         self,
         request_id: str,
@@ -984,11 +989,6 @@ class LMCacheConnectorV1Impl:
             len(missing_blocks),
         )
         return missing_blocks
-
-            self._stats_monitor.update_interval_vllm_hit_tokens(
-                request.load_spec.vllm_cached_tokens
-            )
-            self._stats_monitor.update_interval_prompt_tokens(len(tokens))
 
     @_lmcache_nvtx_annotate
     def wait_for_layer_load(self, layer_name: str) -> None:
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -934,9 +934,9 @@ class LMCacheConnectorV1Impl:
             ret_mask: Boolean tensor indicating which tokens were actually
                 successfully retrieved from LMCache. True means the token was
                 successfully loaded. For example, if 256 tokens are expected to be
-                loaded, but only 192 tokens are successfully loaded, then the 
+                loaded, but only 192 tokens are successfully loaded, then the
                 ret_mask will be a tensor of 256 items like [T, T, ..., F, F, ...]
-                where the first 192 elements are True and the last 64 elements 
+                where the first 192 elements are True and the last 64 elements
                 are False.
             slot_mapping: Tensor indicating slot IDs for each token. The block
                 ID is computed by dividing the slot ID by the block size.
@@ -947,7 +947,7 @@ class LMCacheConnectorV1Impl:
             missing_mask = expected_mask & ~ret_mask = [F, F, T, T]
             missing_indices = [2, 3]
             then missing_blocks is calculated from slot_mapping and missing_indices
-                
+
         Returns:
             set[int]: Set of block IDs that failed to load.
         """
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -671,6 +671,9 @@ class LMCacheConnectorV1Impl:
 
         self._requests_priority: dict[str, int] = {}
 
+        # Track block IDs associated with failed load attempts.
+        self._invalid_block_ids: set[int] = set()
+
         # TODO(baoloongmao): Internal api server & plugin framework support dp > 1
         if vllm_config.parallel_config.data_parallel_rank_local == 0:
             # Start internal API server if enabled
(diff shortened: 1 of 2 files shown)


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -885,6 +888,59 @@
                         num_retrieved_tokens,
                         num_expected_tokens,
                     )
```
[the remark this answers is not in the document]
ziruiliu: information is plenty in other places

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -885,6 +888,59 @@
...
+
+        if not missing_blocks:
+            return
+
+        logger.warning(
+            "Request %s failed to load %d token(s) across %d block(s); marking blocks invalid.",
+            request_id,
+            missing_indices.numel(),
+            len(missing_blocks),
+        )
```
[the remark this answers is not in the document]
ziruiliu: won't change. missing_blocks could be huge

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -885,6 +888,59 @@
...
+        if not torch.any(missing_mask):
+            return
+
+        missing_indices = torch.nonzero(missing_mask, as_tuple=False).view(-1)
+        if missing_indices.numel() == 0:
+            return
+
+        slot_mapping_cpu = slot_mapping.to(device="cpu", dtype=torch.long)
+        if slot_mapping_cpu.shape[0] > missing_mask.shape[0]:
+            slot_mapping_cpu = slot_mapping_cpu[: missing_mask.shape[0]]
```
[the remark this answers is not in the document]
ziruiliu: Actually there is no chance that incoming parameter `slot_mapping`, `token_mask` have different shape.
As long as `missing_mask` is calculated from `token_mask` and could only be shorter than `token_mask`, there is no chance that `missing_mask` is longer than `slot_mapping`

ApostaC:
Thanks! Will take a look!

@maobaolong You might also be interested in this PR.

maobaolong:
@ApostaC Maybe the approach of vLLM #19330 can only work if `get_num_new_matched_tokens` return `0,True`?

ziruiliu:
> @ApostaC Maybe the approach of vLLM #19330 can only work if `get_num_new_matched_tokens` return `0,True`?

invalid blocks are reported at stage of retrieve(), not at stage of lookup in lmcache, get_num_new_matched_tokens()

As I observed from log, get_num_new_matched_tokens() returns the number of existing tokens in storage, e.g. 256, but later on retrieve() only load part of those tokens, e.g. 192, then #19330 works as expected:

WARNING 10-20 16:12:58 [v1/core/sched/scheduler.py:1499] Recovered from KV load failure: 1 request(s) rescheduled (63 tokens affected).

maobaolong:
@yoo-kumaneko Would you please take a look at this PR? Thanks!

ziruiliu:
To reproduce and test
1. configuration: set chunk size to 64, enable fs connector (or local_disk_backend) in yaml file.
```yaml
chunk_size: 64
local_cpu: False
save_unfull_chunk: False
remote_url: "fs://nomatter:0/somecachedir"
remote_serde: "naive"
```

2. start vllm with --no-enable-prefix-caching to make it reproduced faster
```
VLLM_LOGGING_LEVEL=DEBUG PYTHONHASHSEED=0 \
LMCACHE_CONFIG_FILE=fs.yaml LMCACHE_LOG_LEVEL=DEBUG \
vllm serve Qwen/Qwen3-4B --served-model-name 'Qwen/Qwen3-4B' --enforce-eager  \
--kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}' \
--disable-log-stats --disable-log-requests -tp 1 \
--no-enable-prefix-caching
```

3. prepare a 256 token prompt and send
4. 4 chunk files will be created in configured directory
5. change the permission of the files, to make these file visible via lookup() but unreadable in later get()
```
chmod 000 some-cache-files
```
6. send request again

**With original codebase:**
Response will be incorrect and irrelevant
Errors will be found in log:
```
LMCache INFO: Reqid: cmpl-2cd53ce35ea345759274c5a2587a52f0-0, Total tokens 256, LMCache hit tokens: 256, need to load: 255
LMCache DEBUG: Scheduled to load 256 tokens for request cmpl-2cd53ce35ea345759274c5a2587a52f0-0
LMCache WARNING: The cache block is in the storage, but it can't be retrieved
LMCache INFO: Retrieved 128 out of 256 required tokens (from 256 total tokens)
LMCache ERROR: The number of retrieved tokens is less than the expected number of tokens! This should not happen!
LMCache ERROR: Num retrieved tokens: 128, num expected tokens: 256
```
Meaning only 128 token's cache is loaded properly and rest part is missing, causing reponse totally wrong.

**With the change:**
modify vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py, to enable this interface from vllm
```diff
--- a/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
+++ b/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py
@@ -110,6 +110,10 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
         """
         return self._lmcache_engine.get_finished(finished_req_ids)
 
+    def get_block_ids_with_load_errors(self) -> set[int]:
+        """Return block IDs that failed to load during the last interval."""
+        return self._lmcache_engine.get_block_ids_with_load_errors()
+

```
Also apply this change in LMCache, resend the request, you will see correc response and there will be recovery message in log:
```
[v1/core/sched/scheduler.py:1499] Recovered from KV load failure: 1 request(s) rescheduled (127 tokens affected).
```

ApostaC:
> @ApostaC Maybe the approach of vLLM #19330 can only work if `get_num_new_matched_tokens` return `0,True`?

IIRC, that PR supports both async and sync loading.

ApostaC:
~~Please don't merge it yet. I'm going to update it a bit more~~

ApostaC:
> Please don't merge it yet. I'm going to update it a bit more

Sorry folks, I realized I replied at a wrong PR! Sorry 🙏! Please ignore it!

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -886,6 +889,59 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
                         num_retrieved_tokens,
```
yoo-kumaneko, line 900: Why don’t we call `self.record_failed_blocks` inside this if branch?

ApostaC: Agreed. We should also do it here.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -886,6 +889,59 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
...
+                    slot_mapping[:lmcache_cached_tokens],
+                )
+
+    def record_failed_blocks(
+        self,
+        request_id: str,
+        expected_mask: torch.Tensor,
+        ret_mask: torch.Tensor,
+        slot_mapping: torch.Tensor,
+    ) -> None:
```
ApostaC: Can we have this function return an explicit list of `invalid_block_ids`? Otherwise, it's not clear to other developers what the "side effect" of this function is.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -886,6 +889,59 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
...
+        """Record block IDs associated with failed load attempts."""
```
ApostaC: It's better to add descriptions of the args, i.e., what each argument means. Would be even better to provide an example of values in the `expected mask` and `ret mask` and the corresponding return values.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -886,6 +889,59 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
                         num_retrieved_tokens,
                         num_expected_tokens,
                     )
+                self.record_failed_blocks(
+                    request.req_id,
+                    token_mask[:lmcache_cached_tokens],
+                    ret_token_mask,
+                    slot_mapping[:lmcache_cached_tokens],
+                )
```
ApostaC: As @yoo-kumaneko said, I think this function should only be called in the if branch above (`if num_retrieved_tokens < num_expected_tokens:`). It's to avoid the extra overhead caused by comparing tensors and doing D2H copy.

ApostaC (changes requested):
Thanks! Left some comments.

lmcache/integration/vllm/vllm_v1_adapter.py
ziruiliu: changed to return invalid block ids to caller star_load_kv(). star_load_kv() then update self._invalid_blocks_id

lmcache/integration/vllm/vllm_v1_adapter.py
ziruiliu: changed to call record_failed_blocks() only in failed condition

lmcache/integration/vllm/vllm_v1_adapter.py
ziruiliu: Added comments explaining input params and how missing blocks are calculated

lmcache/integration/vllm/vllm_v1_adapter.py
ziruiliu: Done as explained above. Call only in failed condition

ApostaC (approved):
Thanks! LGTM now!

@maobaolong @yoo-kumaneko Would you like to take a look again at this PR?

yoo-kumaneko:
LGTM!

maobaolong (approved):
LGTM

PR #1833, opened 2025-10-20
[FIX][Adapt_vllm] Fix ci failed by get_kv_cache_torch_dtype missed

maobaolong:
After merged https://github.com/vllm-project/vllm/pull/26908 the `get_kv_cache_torch_dtype` moved to new location, it caused LMCache failed.

PR #1822, opened 2025-10-18
[CI/Build] Add comprehensive test for layerwise KV transfer

Shaoting-Feng:
**What this PR does / why we need it**:
To ensure layerwise feature not broken by other PRs.

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #1836, opened 2025-10-20
[Patch] Update import for `get_kv_cache_torch_dtype` after vllm/utils/__init__.py refactoring

JZhou3083:
This patch updates LMCache’s vLLM integration to align with the recent refactor of vLLM that moved 
`get_kv_cache_torch_dtype` from `vllm/utils/__init__.py` to `vllm/utils/torch_utils.py`, which caused vLLM Integration Tests to fail:
image

 The update is for the latest vllm, but also is compatible to the older version.

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -133,7 +133,7 @@ def create_lmcache_metadata(
     try:
         # Third Party
         from vllm.utils.torch_utils import get_kv_cache_torch_dtype
-    except ImportError:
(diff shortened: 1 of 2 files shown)


Shaoting-Feng:
Thanks for your fix! This is also fixed in https://github.com/LMCache/LMCache/pull/1822.

JZhou3083:
Closing this PR as this is duplicated work with https://github.com/LMCache/LMCache/pull/1833

PR #1843, opened 2025-10-21
[WIP][Core] Add KV events

### LINKED ISSUE #1846: [RFC] Add KV events for vLLM
### Motivation

It is [proposed in vLLM to provide standardized notification events](https://github.com/vllm-project/vllm/issues/20492) called [KV Events](https://github.com/vllm-project/vllm/issues/16669) for GPU cache events and CPU offloading events. The [vLLM RFC](https://github.com/vllm-project/vllm/issues/20492) proposes _"formalizing KV Events as the public contract for any component emitting or consuming KV-Cache lifecycle events - including external indexers, routers, and engines"_.

This RFC is as a result of a discussion in the [community slack channel](https://lmcacheworkspace.slack.com/archives/C09GS9NQU56/p1758220352911439) about supporting KV Events in LMCache.

### Requirements

1. The event types are defined in [vLLM KV Events](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_events.py). LMCache id to follow these definitions.

2. Events published for the following operations:
- Cache stored
~~- Cache evicted~~

3. The event for cache stored contains the following information per chunk of a sequence:

- `token_ids`
- `block_size`
- `hashes` are prefix hashes of `token_ids` chunked
-  Event at the beginning of the sequence, has empty `parent_hash`

### Proposed Change 1:

1. Mirror the implementation in [vLLM](https://github.com/vllm-project/vllm/pull/16750):
- It is proposed to implement this as a publish-subscribe (pub-sub) messaging system. LMCache will publish messages/events to an endpoint and topic (which is a named channel or category) that consumers can subscribe to or listen on.
- It is proposed to use [ZeroMQ](https://zeromq.org/) which is a high-performance, asynchronous open source messaging library. It is also broker-less which means it doesn't require a dedicated message broker.

2. Reference the KV Event definitions from [vLLM](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_events.py)

3. Modify the publisher implementation from [vLLM](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_events.py) to work in LMCache

4.  Publish `BlockStored` events per sequence chunk in `store` and `store_layer` methods of [LMCacheEngine](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/cache_engine.py) for cache storage

5. Publish `BlockRemoved` events in <TODO>?? for cache evictions

6. Configuration will be added as follows:- To enable KV events - disabled by default 
- Messaging system configuration like:
   - Published endpoint
   - Replay endpoint
   - Topic name
   - .... 
 
### Proposed Change 2:

This proposed solution was suggested by @ApostaC in https://github.com/LMCache/LMCache/issues/1846#issuecomment-3482096943. It has the following advantages:

- Reuse of existing messaging publishing in vLLM
- Minimal change to vLLM - only need to generate the events and store then
- Most changes are to the connector where it is better placed to handle vLLM data structure required for publishing

The tasks are as follows:

1. Create events during storage operations and store then in the cache engine. The events custom LMCache data structure
2. Use the vLLM worker process to retrieve the events from the cache engine and for the scheduler process to publish them
3. The overall data flow is as follows:
- Model runner process calls connector `get_kv_connector_kv_events` method which calls cache_engine to return the events
- Model runner process sends the information to the scheduler process via `KVConnectorOutput` class
- Scheduler process calls connector `update_connector_output` method and the connector gets access to events in scheduler side now
- Scheduler calls connector `take_events` method, and the connector returns the latest events to be published by the scheduler

### Implementation

- [x] Events generation: Add cache storage event creation and passing to vLLM (LMCache) [#1959]
- [x] Event handling and publishing: Add KV events handling to KV Connectors and LMCache connector (vLLM) [https://github.com/vllm-project/vllm/pull/28309]
- [x] Documentation (LMCache) [#2241]

### Open Questions

1. The 2 RFCs for KV Events in vLLM do not seem to have received widespread agreement or acceptance by the community as of yet. 

[[RFC]: KVBlocks and Metrics Publishing In Inference Frameworks](https://github.com/vllm-project/vllm/issues/16669) is gone stale and now closed.  @alec-flowers What is the status of the RFC?

@vMaroon What is the status of: [[RFC]: KV-Cache Interoperability API Standardization](https://github.com/vllm-project/vllm/issues/20492) ?

Answer: Events to be used for KV cache aware routing by platforms like llm-d and Nvidia Dynamo. See response from @ApostaC in https://github.com/LMCache/LMCache/issues/1846#issuecomment-3482100663

2. Proposed changes to the event structure as follows:
- adding a version # (so that message structure can evolve over time)
- adding an owner field (so that ownership of the cache related to the event can be added if needed)

These would need to be added to vLLM where the structure is defined. @alec-flowers @ApostaC Can you elaborate as you discussed these AFIK?

### Additional Comments

- ~~PR #1843 implements `Proposed Change 1`~~
- PR #1959 implements `Proposed Change 2` and is the preferred choice.

@alec-flowers @YaoJiayi @junchenj @vMaroon @ApostaC

hickeyma:

This PR adds capability to publish events for cache storage and removal (eviction).  The underlying infrastructure is a publish-subscribe (pub-sub) messaging system using [ZeroMQ](https://zeromq.org/) for the publishing. [ZeroMQ](https://zeromq.org/) is a high-performance, asynchronous open source messaging library.

This PR is based on the vLLM RFC [[RFC]: KV-Cache Interoperability API Standardization ](https://github.com/vllm-project/vllm/issues/20492) which proposes "formalizing KV Events as the public contract for any component emitting or consuming KV-Cache lifecycle events - including external indexers, routers, and engines". 

Fixes #1846

- The configuration is defined in  `lmcache/v1/config.py`. Please note that KV events are disabled by default. Need to set `enable_kv_events` to `True` for events to be generated.
- Consumer code to test the events being generated can be referenced in: https://docs.vllm.ai/en/stable/examples/online_serving/kv_events_subscriber.html
- Currently, events only being generated for storage

**If applicable**:


lmcache/v1/cache_engine.py
```diff
@@ -128,6 +137,21 @@ def __init__(
...
+            if self.kv_events_enabled:
+                self.kv_event_publisher = EventPublisherFactory.create(
+                    config,
+                )
+                self.kv_event_queue: List[KVCacheEvent] = []
+                logger.info("KV events are enabled.")
+            else:
+                logger.info("KV events are disabled.")
+        else:
+            logger.info("KV events are unavailable as vLLM is required.")
```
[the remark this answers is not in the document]
hickeyma: Good catch

lmcache/v1/cache_engine.py
```diff
@@ -277,6 +302,32 @@ def store(
...
+                        if isinstance(tokens, torch.Tensor)
+                        else tokens[start : end + 1]
+                    )
+                elif hashes is not None:
+                    stored_block.token_ids = hashes[start : end + 1]
+                logger.info(f"Added block '{stored_block}' to kv events queue")
+                self.kv_event_queue.append(stored_block)
+                prev_key = prev_key = (
+                    hash(key) if isinstance(key, CacheEngineKey) else key
+                )
```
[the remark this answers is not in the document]
hickeyma: Good catch

lmcache/v1/cache_engine.py
```diff
@@ -417,6 +494,15 @@ def store_layer(
...
         logger.debug(f"Stored {tot_token_num} out of total {len(tokens)} tokens")
+
+        # Publish the KV events
+        if vllm_is_available and self.kv_events_enabled:
+            logger.info("Publish kv events")
+            batch = KVEventBatch(ts=time.time(), events=self.kv_event_queue)  # type: ignore[arg-type]
+            self.kv_event_queue = []
+            if self.kv_event_publisher:
+                self.kv_event_publisher.publish(batch)
+
```
YaoJiayi, line 518: Why do we add publish kv events in cache_engine?

YaoJiayi: Any reason we shouldn't do it here? https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/local_cpu_backend.py#L143

lmcache/v1/config.py
```diff
@@ -396,6 +396,47 @@ def _parse_quoted_string(value: str) -> str:
...
+    "kv_events_publisher_hwm": {
+        "type": int,
+        "default": 100_000,
+        "env_converter": int,
+    },
+    "kv_events_publisher_max_queue_size": {
+        "type": int,
+        "default": 100_000,
+        "env_converter": int,
+    },
```
YaoJiayi, line 460: Do we need this many configs? If so, can we add some comments here to explain them?

lmcache/v1/kv_events.py
```diff
@@ -0,0 +1,374 @@
...
+                raise ValueError(f"Unknown event publisher '{kind}'") from exc
+            return constructor(
+                data_parallel_rank=data_parallel_rank,
+                endpoint=config.kv_events_publisher_endpoint,
+                replay_endpoint=config.kv_events_publisher_replay_endpoint,
+                buffer_steps=config.kv_events_publisher_buffer_steps,
+                hwm=config.kv_events_publisher_hwm,
+                max_queue_size=config.kv_events_publisher_max_queue_size,
+                topic=config.kv_events_publisher_topic,
+            )
```
YaoJiayi: We already have the events publishing mechanism at https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/cache_controller/worker.py#L171 . Can we reuse the same code path here?

YaoJiayi (commented):
Thanks @hickeyma ! I left some comments

lmcache/v1/config.py
hickeyma: I think we do. I have added comments to each configuration setting as suggested.

lmcache/v1/cache_engine.py
hickeyma: Thanks @YaoJiayi, that makes more sense.

Here are potential updates to `local_cpu_backend`:

```python
def submit_put_task(
        self, key: CacheEngineKey, memory_obj: MemoryObj,
        parent_key: int,
    ) -> Optional[Future]:
        """
        Synchronously put the MemoryObj into the local cpu backend.
        """

        with self.cpu_lock:
            if key in self.hot_cache:
                return None

            memory_obj.ref_count_up()
            self.hot_cache[key] = memory_obj

            self.cache_policy.update_on_put(key)

            # TODO(Jiayi): optimize this with batching?
            # push kv admit msg
            if self.lmcache_worker is not None:
                self.lmcache_worker.put_msg(
                    KVAdmitMsg(
                        self.instance_id, key.worker_id, key.chunk_hash, str(self)
                    )
                )
                if vllm_is_available and self.kv_events_enabled:
                    logger.info(f"Added block '{stored_block}' to kv events queue")
                    self.kv_event_queue.append(stored_block)
                    stored_block = BlockStored(
                        block_hashes=[hash(key)],
                        parent_block_hash=None if parent_key == 0 else parent_key,
                        token_ids=<????>,
                        block_size=<????>,
                        lora_id=<????>,
                        medium="GPU",
                    )
        return None

    def batched_submit_put_task(
        self,
        keys: Sequence[CacheEngineKey],
        memory_objs: List[MemoryObj],
        transfer_spec: Any = None,
    ) -> None:
        """
        Synchronously put the MemoryObjs into the local cpu backend.
        """
        if not self.use_hot:
            return

        # TODO(Jiayi): optimize this with batching
        prev_key = 0
        for key, memory_obj in zip(keys, memory_objs, strict=False):
            self.submit_put_task(key, memory_obj, prev_key)
            if vllm_is_available and self.kv_events_enabled:
                prev_key = hash(key)
```

hickeyma: @YaoJiayi Have 2 questions:

1. How to fill the following fields: `token_ids`, `block_size`, `lora_id`?
2. When/Where to publish the events. Need to run the following:

```python
# Publish the KV events
        if vllm_is_available and self.kv_events_enabled:
            logger.info("Publish kv events")
            batch = KVEventBatch(ts=time.time(), events=self.kv_event_queue)  # type: ignore[arg-type]
            self.kv_event_queue = []
            if self.kv_event_publisher:
                self.kv_event_publisher.publish(batch)
```

lmcache/v1/kv_events.py
hickeyma: @YaoJiayi I updated the `kv_events` to `messaging` to be more generic rather than specific to kv events and therefore provide a class which abstracts the implementation the PUB/SUB messaging pattern using `ZeroMQ`.

It is essentially using the same underlying wrapper functions as in the `cache_controller`:

```python
from lmcache.v1.rpc_utils import (
    close_zmq_socket,
    get_zmq_context,
    get_zmq_socket,
)
```

The code in `messaging.py` however also abstracts the underlying code for messaging as follows:

- Provides the ability to use different messaging tools/libraries if we want. `ZeroMQ` is only supported at this moment
- Provides the`ZmqMessagePublisher` class which:
   - Implements the PUB/SUB messaging pattern for `ZeroMQ` 
   - Spawns a separate thread for publishing messages from an internal queue
   - Supports replay of messages 
   - Handles the socket configuration as follows:
      - Sets the socket connection type best suited based off the endpoint
      - Sets the role best suited for the messaging pattern

From looking at the code in `cache_controller`, there is a lot of duplicated code that could be abstracted as is done for the PUB/SUB messaging pattern in `messaging.py`. Therefore, I think that we should look at abstracting the messaging code in `cache_controller` at some stage in a separate PR for the different messaging patterns that the code is using. This way we could decrease the zmq code duplication across the code base and therefore minimize coding errors and bugs.

Let me know what you think.

hickeyma:
Thanks @YaoJiayi for your review and feedback. I have updated the code and also provided feedback to questions that you raised.

hickeyma:
Closing as superseded by #1959

PR #1848, opened 2025-10-21
[hotfix] Reduce async loading log verbosity

DongDongJu:
@maobaolong report that current post_init produce too many log caused by async serializer init.


DongDongJu:
I opened with wrong branch. Let me re open it

PR #1849, opened 2025-10-21
[hotfix] Reduce async loading log verbosity

DongDongJu:
@maobaolong report that current post_init produce too many log caused by async serializer init.

### DIFF
--- a/lmcache/v1/storage_backend/local_cpu_backend.py
+++ b/lmcache/v1/storage_backend/local_cpu_backend.py
@@ -613,7 +613,7 @@ class LocalCPUBackend(AllocatorBackendInterface):
         # calculate budget with safety margin
         max_chunks = total_memory // aligned_chunk_bytes
 
-        return int(max_chunks)
+        return max_chunks
 
     def get_keys(self) -> List[CacheEngineKey]:
         """
--- a/lmcache/v1/storage_backend/local_cpu_backend.py
+++ b/lmcache/v1/storage_backend/local_cpu_backend.py
@@ -573,7 +573,7 @@ class LocalCPUBackend(AllocatorBackendInterface):
         Returns:
             int: The estimated chunk budget for concurrent allocations
         """
-        logger.info("Attempting to calculate chunk budget for async loading")
+        logger.debug("Attempting to calculate chunk budget for async loading")
         assert self.metadata is not None, (
             "metadata required for chunk budget calculation"
         )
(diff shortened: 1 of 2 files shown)


DongDongJu:
@maobaolong @sammshen Could you take a look at this?

maobaolong:
@DongDongJu Thanks for this fix, it looks fix the too many log issue.

But, moreover, would you like to explain a bit about why we need to construct AsyncSerializer although async_lookup_server is None? It means that we start lmcache with `async_loading==False` mode.

DongDongJu:
Thanks @maobaolong 
Even when enable_async_loading is False, the layerwise retrieval path still leans on AsyncSerializer.
So if someone runs in “layerwise but no async loading” mode, we still have to instantiate the serializer or the assertion (and the semaphore protection that follows) breaks.
But, if both config.enable_async_loading and config.use_layerwise are false, we don’t need it. you are right. 
So I add guard for these condition in https://github.com/LMCache/LMCache/pull/1849/commits/ad01e46558724628e6a54c58bfdc075056d3e2c8

@sammshen Please correct me if I was wrong here.

maobaolong:
@DongDongJu Thanks for your clarification about `AsyncSerializer`.

I've know why we need to instantiate AsyncSerializer for each request, but is it necessary? Do you think it would be better if we do not instantiate AsyncSerializer for each request, instead, we only start one AsyncSerializer instance, but execute `calculate_chunk_budget` periodically or necessary?

maobaolong:
After a discussion between @sammshen , it could be another further work need to do to refactor the existing AsyncSerializer logic and behavior, but for this PR, it can be a hotfix enough, thanks @DongDongJu for this urgent fix!

maobaolong (approved):
LGMT

DongDongJu:
I agree. we need refactor for this initialization way. Let me try to working on it after centralized memory pool layer.
Anyone can merge this? @sammshen @YaoJiayi

maobaolong:
@DongDongJu  https://github.com/LMCache/LMCache/pull/1852/files  It looks @sammshen did a relevant modification within that PR.

DongDongJu:
@maobaolong LGTM. but still this one need to pick up. this condition I added is for do not allocating the async serializer under condition that its not need it.

sammshen (approved):
LGTM, thanks @DongDongJu

PR #1859, opened 2025-10-22
[PATCH]: add dtype to CEK

### LINKED ISSUE #1856: [BUG] Chunk name/CacheEngineKey should contains dtype to avoid multiple vllm cluster cache matched by accident
**Describe the bug**

There are two vllm cluster with same module, same tp count, but with different kvcache dtype. Under this assumption, there is a risk to have two different chunks belong to different vLLM cluster, but with save CacheEngineKey, if we use shared remote backend, vLLM0 cluster may get the vLLM0 cluster's chunk regard as cache hit, but the content is not correct.

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1856

We can support hetero KV dtype after vLLM does

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -185,7 +185,7 @@ class CacheEngineKey:
             f"@{self.worker_id}@{self.chunk_hash:x}@{self._dtype_str}"
         )
         if self.tags is not None and len(self.tags) != 0:
-            tags = [f"{k}%{v}" for k, v in sorted(self.tags)]
+            tags = [f"{k}%{v}" for k, v in self.tags]
             s += "@" + "@".join(tags)
         return s
 
@@ -314,7 +314,7 @@ class LayerCacheEngineKey(CacheEngineKey):
             f"@{self.worker_id}@{self.chunk_hash:x}@{self._dtype_str}@{self.layer_id}"
         )
         if self.tags is not None and len(self.tags) != 0:
-            tags = [f"{k}%{v}" for k, v in sorted(self.tags)]
+            tags = [f"{k}%{v}" for k, v in self.tags]
             s += "@" + "@".join(tags)
         return s
 
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -91,23 +91,49 @@ TORCH_DTYPE_TO_STR_DTYPE = {
     torch.bfloat16: "bfloat16",
     torch.float: "float",
     torch.float32: "float",
-    torch.float64: "double",
     torch.double: "double",
-    torch.uint8: "fp8",
-    torch.float8_e4m3fn: "fp8_e4m3",
-    torch.float8_e5m2: "fp8_e5m2",
+    torch.float64: "double",
+    torch.int8: "int8",
+    torch.uint8: "uint8",
+    torch.int16: "int16",
+    torch.int32: "int32",
+    torch.int64: "int64",
+    torch.bool: "bool",
 }
+
+# FP8 variants (PyTorch ≥2.1)
+if hasattr(torch, "float8_e4m3fn"):
+    TORCH_DTYPE_TO_STR_DTYPE[torch.float8_e4m3fn] = "fp8_e4m3"
+if hasattr(torch, "float8_e4m3fnuz"):
+    TORCH_DTYPE_TO_STR_DTYPE[torch.float8_e4m3fnuz] = "fp8_e4m3"
+if hasattr(torch, "float8_e5m2"):
+    TORCH_DTYPE_TO_STR_DTYPE[torch.float8_e5m2] = "fp8_e5m2"
+if hasattr(torch, "float8_e5m2fnuz"):
+    TORCH_DTYPE_TO_STR_DTYPE[torch.float8_e5m2fnuz] = "fp8_e5m2"
 
 STR_DTYPE_TO_TORCH_DTYPE = {v: k for k, v in TORCH_DTYPE_TO_STR_DTYPE.items()}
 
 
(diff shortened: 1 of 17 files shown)


lmcache/utils.py
```diff
@@ -91,24 +91,40 @@ def can_evict(self) -> bool:
...
-@dataclass(order=True)
+@dataclass(slots=True)
 class CacheEngineKey:
     fmt: str
     model_name: str
     world_size: int
     worker_id: int
     chunk_hash: int
-    request_configs: Optional[dict] = None
+    dtype: torch.dtype
```
maobaolong, line 152: Can we set a default value to `dtype`, so that the diff of this PR can reduce a lot?

lmcache/utils.py
```diff
@@ -91,24 +91,40 @@ def can_evict(self) -> bool:
...
+    request_configs: Optional[dict] = field(default_factory=dict)
+    tags: Optional[tuple] = field(init=False, default=None)
```
maobaolong, line 154: Why we need these changes?

lmcache/utils.py
```diff
@@ -272,10 +299,10 @@ def __eq__(self, other):
     def to_string(self):
         s = (
             f"{self.fmt}@{self.model_name}@{self.world_size}"
-            f"@{self.worker_id}@{self.chunk_hash:x}@{self.layer_id}"
+            f"@{self.worker_id}@{self.chunk_hash:x}@{self._dtype_str}@{self.layer_id}"
         )
         if self.tags is not None and len(self.tags) != 0:
-            tags = [f"{k}%{v}" for k, v in self.tags]
+            tags = [f"{k}%{v}" for k, v in sorted(self.tags)]
```
maobaolong: Why you execute `sort` for tags? @chunxiaozheng Please take a double check of this point.

maobaolong (commented):
Left some comments.

lmcache/utils.py
sammshen: this allows us to use the `@dataclass(sort=True)`, which makes the dataclass a struct with offsets instead of using internal `__dict__` in python, which is more space expensive

lmcache/utils.py
sammshen: then I think all of the dtype will be wrong then in the examples

lmcache/utils.py
sammshen: this is to keep the tag order deterministic regardless of insertion order

sammshen:
thanks for the review @maobaolong !

tests/v1/test_remote_mla_worker_id_as0.py
```diff
@@ -106,6 +106,7 @@ def test_remote_mla_worker_id_as0(mock_stream):
         world_size=4,
         worker_id=2,
         chunk_hash="test_hash",
+        dtype=torch.float32,
```
maobaolong, line 109: @sammshen 
Why here set to `torch.float32`, but other tests set to `torch.float16` or `torch.uint8,`?

If the it is same if we set any dtype for test code, can we set a default dtype for test purpose, so all of the test scope code will keep non-change, we only set the dtype for the production code.

maobaolong:
@sammshen Thanks for your reply. Left a new question inline.

tests/v1/test_remote_mla_worker_id_as0.py
sammshen: chose the dtype based on the dtype used in other places in that unit test file

sammshen: even though the dtype doesn't do anything in the tests, the test should show a good example of how to create CacheEngineKey so future tests also use it properly

maobaolong: @sammshen I found the `torch.float32` here referred to the `tests/v1/test_remote_mla_worker_id_as0.py#L132`

```
memory_obj = local_cpu_backend.allocate(torch.Size([10, 10]), torch.float32)
```

Which introduced by me.

maobaolong (approved):
@sammshen Sorry for the late review. This PR LGTM except the part about dataclass, let @chunxiaozheng do a double check.

lmcache/utils.py
chunxiaozheng: @sammshen Is there a problem with using sorted here?  When `from_string` is called to create `CacheEngineKey`,  `__post_init__` will construct tags in the order of strings, which will cause the order of tags to be inconsistent with the original order of tags.

lmcache/utils.py
```diff
@@ -149,10 +170,10 @@ def __eq__(self, other):
     def to_string(self):
         s = (
             f"{self.fmt}@{self.model_name}@{self.world_size}"
-            f"@{self.worker_id}@{self.chunk_hash:x}"
+            f"@{self.worker_id}@{self.chunk_hash:x}@{self._dtype_str}"
         )
         if self.tags is not None and len(self.tags) != 0:
-            tags = [f"{k}%{v}" for k, v in self.tags]
+            tags = [f"{k}%{v}" for k, v in sorted(self.tags)]
```
chunxiaozheng: The same problem as below.

lmcache/utils.py
chunxiaozheng: I don't think it's necessary to use `sorted` here. Without `sorted`, `__post_init__`, `to_string` and `from_string` will all be processed in the order of input.

sammshen:
addressed comments @chunxiaozheng !

chunxiaozheng (approved):
lgtm! Thanks @sammshen !

Issue #1866, opened 2025-10-23
[Bug] response is wrong due to async KV loading using wrong memory object


ziruiliu:
**Describe the bug**
When both vllm's prefix cache and LMCache are hit. `_async_process_tokens_internal()` in cache_engine.py appends wrong memory object to chunks and return to caller, causing vLLM generating wrong response

**To Reproduce**
1. Configure a simple cpu backend, enable async loading
```yaml
chunk_size: 64
local_cpu: True
max_local_cpu_size: 5
enable_async_loading: True
```
2. start vllm with such cpu.yaml
```
VLLM_LOGGING_LEVEL=DEBUG PYTHONHASHSEED=0 \
LMCACHE_CONFIG_FILE=cpu.yaml LMCACHE_LOG_LEVEL=DEBUG \
vllm serve Qwen/Qwen3-4B --served-model-name 'Qwen/Qwen3-4B' --enforce-eager  \
--kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}' \
--disable-log-stats --disable-log-requests -tp 1
``` 
3. prepare a 256 token prompt (for Qwen3-4B) and send
<details>
  <summary>256 token prompt and request.py</summary>
```
Stanford University, formally Leland Stanford Junior University, is a world-renowned private research institution located in the heart of Silicon Valley, California. Known colloquially as "The Farm" due to its original purpose as the Stanford family's stock farm, its sprawling 8,180-acre campus is one of the largest in the United States. Founded in 1885 by railroad magnate Leland Stanford and his wife, Jane, in memory of their only child, Leland Jr., the university officially opened its doors in 1891 with a commitment to educating both men and women.
The university's academic excellence is consistently recognized globally, fostering innovation across a vast range of disciplines, from its pioneering work in computer science and engineering to its highly respected schools of business, law, and medicine. Stanford's entrepreneurial spirit has profoundly influenced the surrounding technology ecosystem, with faculty and alumni playing pivotal roles in the creation of companies like Google Hewlett-Packard, and Yahoo.The university operates under the foundational creed, emphasizing curiosity.
Your task is to synthesize the above information and elaborate on the university's impact on innovation and entrepreneurship, focusing on its location in Silicon Valley. Here is the question: in what year did Stanford University officially open its doors to students?
```
</details>

4. for the first time, response is correct because kv cache is fully computed and written to vllm's GPU cache and LMCache's CPU cache
```
[...]

PR #1867, opened 2025-10-23
fix to use correct mem obj in async loading

ziruiliu:
**Why**
As described in issue #1866, when vllm's prefix cache and LMCache are both enabled and async load is enabled. In the test case, LMCache chose the wrong memory object from prefetched objects. It happens as:

When async_lookup_and_prefetch is called, it retrieves memory objects for all chunks (without considering the mask). Later, when _async_process_tokens_internal is called with a mask that has False values at the beginning:
1. The  process_tokens method skips chunks corresponding to the  False values in the mask
2. Current code used idx (iteration counter) to index into memory_objs, but idx only counted the iterations that weren't skipped
3. This caused the wrong memory objects to be selected 

Example: With 256 tokens total (4 chunks of 64 tokens each), if the mask has 192  False values followed by 64 True values:
- memory_objs contains 4 memory objects (for all chunks)
- process_tokens with mask only iterates once (for the 4th chunk at positions 192-256)
Current code: idx=0 → selects memory_objs[0] (1st chunk) ❌ WRONG
Expected: Maps (start=192, end=256) → memory_objs[3] (4th chunk) ✅ CORRECT

**What**
The fix creates a mapping from (start, end) positions to memory object indices by:
1. First iterating through all tokens with mask=None to get all chunk positions
2. Building a dictionary mapping (start, end) → memory object index
3. When iterating with the actual mask, using this mapping to select the correct memory object
4. Properly freeing unused memory objects by tracking which indices were used
This ensures that the correct memory objects are matched to their corresponding token ranges, regardless of how many chunks are skipped due to the mask.

**Test**
The prompt described in issue #1866 is tested. Response is as expected

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1089,7 +1089,7 @@ class LMCacheEngine:
             chunk_mask_list = [True] * num_chunks
 
         # Now chunk_mask_list is like [F, F, T, ..., T]
-        # where False means the chunk is in upper layer's cache and 
+        # where False means the chunk is in upper layer's cache and
         # doesn't need to be retrieved.
         # So we can use this list to filter out the chunks that are not needed.
         used_indices = set()
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1074,43 +1074,41 @@ class LMCacheEngine:
         # the same order as the lookup order.
         # TODO(Jiayi): hashing inside `process_tokens` can be skipped.
 
-        # First, iterate through all tokens with an all-true mask to get the
-        # correct mapping between chunk indices and memory objects
-        all_chunks_info = list(
+        # First, generate a list indicating which chunks are needed
+        num_chunks = len(tokens) // self.config.chunk_size
+        if mask is not None:
+            chunk_mask_list = []
+            for i in range(num_chunks):
+                start_idx = i * self.config.chunk_size
+                end_idx = start_idx + self.config.chunk_size
+                # Check if all values in this chunk are True
+                chunk_is_valid = all(mask[start_idx:end_idx])
+                chunk_mask_list.append(chunk_is_valid)
+        else:
+            # If no mask, all chunks are needed
+            chunk_mask_list = [True] * num_chunks
+
+        # Now chunk_mask_list is like [F, F, T, ..., T]
+        # where False means the chunk is in upper layer's cache and 
+        # doesn't need to be retrieved.
+        # So we can use this list to filter out the chunks that are not needed.
+        used_indices = set()
+        for idx, (start, end, key) in enumerate(
             self.token_database.process_tokens(
                 tokens=tokens,
-                mask=None,  # Use all-true mask to get all chunks
+                mask=None,
                 request_configs=request_configs,
             )
-        )
-
-        # Build a mapping from (start, end) to memory object index
-        chunk_to_idx = {
-            (chunk_start, chunk_end): idx
-            for idx, (chunk_start, chunk_end, _) in enumerate(all_chunks_info)
-        }
-
-        # Now iterate with the actual mask to determine which chunks to use
-        used_indices = set()
-        for start, end, key in self.token_database.process_tokens(
-            tokens=tokens,
-            mask=mask,
-            request_configs=request_configs,
         ):
+            if not chunk_mask_list[idx]:
+                continue
+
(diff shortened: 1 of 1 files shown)


sammshen:
sorry the integration test can be fixed with https://github.com/LMCache/LMCache/pull/1868

ApostaC:
@YaoJiayi Please take a look at this.

At a high level, I don't think we should rely on the index of the chunk to identify the memory objects. Ideally, those prefetched memory objects should also be indexed by the cache engine keys and managed by the storage manager. In this case, we can directly reuse the sync code path to get the objects, right?

ziruiliu:
> @YaoJiayi Please take a look at this.
> 
> At a high level, I don't think we should rely on the index of the chunk to identify the memory objects. Ideally, those prefetched memory objects should also be indexed by the cache engine keys and managed by the storage manager. In this case, we can directly reuse the sync code path to get the objects, right?

If there is only backend in use, the assumption that mem obj are in order, works fine. 
I think there is another potential issue in async_lookup_and_prefetch(), it issues lookup and get to all backends. If keys of certain tokens (for example 4 keys) exists in multiple backends (say 2 backends), there will be redundant retrieval from backends, and at here
```python
        memory_objs = future.result()
        memory_objs = [mm for m in memory_objs for mm in m]

        # NOTE(Jiayi): here we assume the retrieved memory_objs have
        # the same order as the lookup order.
```
memory_objs is finally a list of list, like [ [4 mem objs], [4 mem objs] ], and only first 4 will be consumed later, because process_tokens() will generate 4 iteration from original tokens

sammshen:
random suggestion: can we pass the `num_computed_tokens` into lookup to generate the mask? to avoid retrieving everything when prefix caching is enabled

```python
def get_num_new_matched_tokens(...):
        num_external_hit_tokens = self.lookup_client.lookup(
            token_ids,
            lookup_id=lookup_id,
            num_computed_tokens # PASS THIS IN
        )
```

ziruiliu:
> random suggestion: can we pass the `num_computed_tokens` into lookup to generate the mask? to avoid retrieving everything when prefix caching is enabled
> 
> ```python
> def get_num_new_matched_tokens(...):
>         num_external_hit_tokens = self.lookup_client.lookup(
>             token_ids,
>             lookup_id=lookup_id,
>             request_configs=request_configs,
>             num_computed_tokens # PASS THIS IN
>         )
> ```

That's another potential enhancement I am preparing. Currently the get_num_new_matched_tokens called lookup with too many tokens. Since key hash calculation depends on entire sequence of token ids, my future change would be pass  in full token_ids and exact number of tokens to be skipped, then skip object lookup and retrieval after key hash calculation

YaoJiayi (commented):
Thanks for the PR! The logic looks a bit complicated too me. Can't we use start/chunk_size to infer the index?

ziruiliu:
> Thanks for the PR! The logic looks a bit complicated too me. Can't we use start/chunk_size to infer the index?

Hi Jiayi, sorry for the previous version. I have refactored code, simply get the indice of chunks that are needed then append to caller. The change is more straightforward now

yoo-kumaneko:
I can reproduce the error.

Is it possible that LMCache does not load the tokens already in vLLM's GPU cache? (i.e. the "computed tokens") That could fix this bug naturally.

lmcache/v1/cache_engine.py
```diff
@@ -1073,22 +1073,47 @@ def _async_process_tokens_internal(
...
-                mask=mask,
+                mask=None,
                 request_configs=request_configs,
             )
         ):
+            if not chunk_mask_list[idx]:
+                continue
+
             assert isinstance(key, CacheEngineKey)
             memory_obj = memory_objs[idx]
```
yoo-kumaneko, line 1084: I think the main problem is here. In the example, we should get the 3rd element from the memory objs list. But the idx is from enumerate and will always be 0th. Let's find a way to use the correct indices then.

yoo-kumaneko: Can we simply use `idx = start // self.config.chunk_size`?

ziruiliu:
Thanks @yoo-kumaneko  for reminding!

Hi @YaoJiayi , I have updated changes to simply use start // chunk as index. Append the chunks that really needed to caller，and mark the used index in a set. Finally free unused memory objects according to recorded indices

YaoJiayi (approved):
lgtm!

PR #1868, opened 2025-10-24
[CI]: add new url triton dependency from latest vllm to fix integration tests

sammshen:
recent PRs are failing integration tests because uv cannot install implicit url dependencies which was just added to https://github.com/vllm-project/vllm/blob/main/requirements/cuda.txt

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -86,12 +86,15 @@ RUN --mount=type=cache,target=/root/.cache/ccache,id=ccache \
     --mount=type=cache,target=/root/.cache/uv,id=uv-cache,sharing=locked \
     . /opt/venv/bin/activate && \
     if [ "$VLLM_VERSION" = "nightly" ]; then \
-        uv pip install 'triton-kernels @ git+https://github.com/triton-lang/triton.git@v3.5.0#subdirectory=python/triton_kernels' && \
-        uv pip install --prerelease=allow 'vllm[runai,tensorizer,flashinfer]' \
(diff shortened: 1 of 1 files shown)


sammshen:
@hickeyma @maobaolong @chunxiaozheng this is pretty urgent, we should merge it asap or else CI will continue to fail

docker/Dockerfile
```diff
@@ -86,11 +86,15 @@ RUN --mount=type=cache,target=/root/.cache/ccache,id=ccache \
     --mount=type=cache,target=/root/.cache/uv,id=uv-cache,sharing=locked \
     . /opt/venv/bin/activate && \
     if [ "$VLLM_VERSION" = "nightly" ]; then \
-        uv pip install --prerelease=allow 'vllm[runai,tensorizer,flashinfer]' \
+        uv pip install --prerelease=allow \
```
ningziwen, line 89: (Non-blocking) If VLLM or any dependencies add other url dependency in the future, we would need to make the same change. Do we consider removing uv until uv supports url dependency support?

chunxiaozheng (approved):
LGTM!

ningziwen (approved):

docker/Dockerfile
sammshen: good point, I think we can probably include a link of url dependencies in a file and then install those. uv is still much faster than pip so I think we should keep uv for now

maobaolong (approved):
LGTM, thanks for this effort!

PR #1869, opened 2025-10-24
[optimize] add batched_get_blocking for nixl_storage_backend

penghb2025:
The current nixl_storage_backend implemented the get_blocking and batched_get_non_blocking. But the batched_get_blocking is inherited from StorageBackendInterface which is a for loop for each of the key. So this PR can help to submit multiple transfer in one nixl transfer request.

### DIFF
--- a/lmcache/v1/storage_backend/nixl_storage_backend.py
+++ b/lmcache/v1/storage_backend/nixl_storage_backend.py
@@ -564,14 +564,13 @@ class NixlStorageBackend(AllocatorBackendInterface):
         :return: MemoryObj. None if the key does not exist.
         """
 
-        # Fast metadata collection (with lock)
-        metadata_list = self._collect_metadata_with_lock([key])
+        future = asyncio.run_coroutine_threadsafe(self.storage_to_mem([key]), self.loop)
 
-        future = asyncio.run_coroutine_threadsafe(
-            self._nixl_transfer_async(metadata_list), self.loop
-        )
+        if future is None:
+            return None
+
         obj_list = future.result()
-        return obj_list[0] if obj_list else None
+        return obj_list[0]
 
     def batched_get_blocking(
         self,
@@ -588,12 +587,8 @@ class NixlStorageBackend(AllocatorBackendInterface):
         if not keys:
             return []
 
-        # Fast metadata collection (with lock)
-        metadata_list = self._collect_metadata_with_lock(keys)
+        future = asyncio.run_coroutine_threadsafe(self.storage_to_mem(keys), self.loop)
 
-        future = asyncio.run_coroutine_threadsafe(
-            self._nixl_transfer_async(metadata_list), self.loop
-        )
         obj_list = future.result()
         return obj_list
 
@@ -603,10 +598,7 @@ class NixlStorageBackend(AllocatorBackendInterface):
         keys: list[CacheEngineKey],
         transfer_spec: Any = None,
     ) -> list[MemoryObj]:
-        # Fast metadata collection (with lock)
-        metadata_list = self._collect_metadata_with_lock(keys)
-
-        obj_list = await self._nixl_transfer_async(metadata_list)
+        obj_list = await self.storage_to_mem(keys)
         assert None not in obj_list
         return cast(list[MemoryObj], obj_list)
 
--- a/lmcache/v1/storage_backend/nixl_storage_backend.py
+++ b/lmcache/v1/storage_backend/nixl_storage_backend.py
@@ -562,9 +562,6 @@ class NixlStorageBackend(AllocatorBackendInterface):
 
         future = asyncio.run_coroutine_threadsafe(self.storage_to_mem(keys), self.loop)
 
-        if future is None:
-            return []
-
         obj_list = future.result()
         return obj_list
 
--- a/lmcache/v1/storage_backend/nixl_storage_backend.py
+++ b/lmcache/v1/storage_backend/nixl_storage_backend.py
@@ -545,6 +545,29 @@ class NixlStorageBackend(AllocatorBackendInterface):
         obj_list = future.result()
         return obj_list[0]
 
+    def batched_get_blocking(
+        self,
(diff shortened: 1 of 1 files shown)


tshmilnvidia (changes requested):
LGTM except the small comment!

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -526,6 +526,29 @@ def get_blocking(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
+        :return: a list of memory objects.
+        """
+
+        if not keys:
+            return []
+
+        future = asyncio.run_coroutine_threadsafe(self.storage_to_mem(keys), self.loop)
+
+        if future is None:
+            return []
```
tshmilnvidia: I believe run_corutine_threadsafe never returns None, so this can be dropped. Might as well also remove the "future is None" case from the non-batched version too.

tshmilnvidia (changes requested):
LGTM, except 1 small comment!

lmcache/v1/storage_backend/nixl_storage_backend.py
penghb2025: Thank you. I updated the batched part.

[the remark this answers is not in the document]
penghb2025: This is the same as what tshmilnvidia suggested. And it was resolved.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -526,6 +526,26 @@ def get_blocking(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
+
+        :param List[CacheEngineKey] keys: The keys of the MemoryObjs.
+
+        :return: a list of memory objects.
+        """
+
+        if not keys:
+            return []
+
+        future = asyncio.run_coroutine_threadsafe(self.storage_to_mem(keys), self.loop)
```
jgoldsch12: Is it possible that calling storage_to_mem with a list of keys will expose a potential deadlock?

https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/nixl_storage_backend.py#L465

Performing the memory allocation while holding the key_lock seems risky - even if the allocation is not supposed to due to slab allocation.  I noticed that the weka backend takes a different approach to collect all the metadata values associated with key each in the list and then perform an allocation (outside the lock scope) for each metadata value.

vvenkates27: @penghb2025 are you able to move the allocation outside the lock? I don't believe this should be a problem. cc: @tshmilnvidia

penghb2025: There is another pr 2024 for this code. I would take a look with that and think whether we need this change.

tshmilnvidia: I'd prefer this to be merged first regardless of PR 2024. We need this function either way as it's quite a significant improvement.

jgoldsch12: I agree that batched_get_blocking shows a significant improvement for OBJ plugin.  For the file based plugins, I think more testing is needed.  Once this code is added, it becomes the default for all NIXL based plugins.  I have a feeling that the NIXL GDS plugin (which does its own batching) may actually perform worse with this change, but don't know for certain.

vvenkates27: Thanks @jgoldsch12 for your validation efforts—they’ve been really helpful. We’ve observed measurable gains on the GDS_MT plugin with this PR, and given the GDS plugin’s reliance on batch APIs, I’d be surprised if batched_get_blocking does not map well to GDS batch operations. We can do more testing, but moving forward with this PR makes sense since it benefits multiple NIXL plugins (files and object). If specific tests are needed for certain plugins, we can address them as separate targeted issues.

@penghb2025 as @tshmilnvidia mentioned, this PR adds significant improvements—please let us know if you’d like us to include it in PR 2024

jgoldsch12: If there's been a good amount of testing great.  My point is that once this goes in there is no optional way to prefer the previous mode of operation.  So if we are feeling confident, great.

I do believe the memory allocation I mentioned above needs to be moved prior to contributing this code.

sammshen (approved):
LGTM

penghb2025:
@jgoldsch12  Sorry for late. Now I split the storage_to_mem into 2 parts. The lock is only around collect the meta data but not with allocation. Please review.

penghb2025:
@vvenkates27 Please help to merge this PR (after review approval) with PR 2024. Thank you.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -537,21 +564,49 @@ def get_blocking(self, key: CacheEngineKey) -> Optional[MemoryObj]:
         :return: MemoryObj. None if the key does not exist.
         """
 
-        future = asyncio.run_coroutine_threadsafe(self.storage_to_mem([key]), self.loop)
+        # Fast metadata collection (with lock)
+        metadata_list = self._collect_metadata_with_lock([key])
+
+        future = asyncio.run_coroutine_threadsafe(
+            self._nixl_transfer_async(metadata_list), self.loop
+        )
```
tshmilnvidia: Can't we keep the call to storage_to_mem same as before, seeing as you did this change inside that function already?

Same for batched_get_blocking  & batched_get_non_blocking

penghb2025: thanks for the comments. Now I restored the code to call storage_to_mem.

tshmilnvidia (approved):
LGTM

sammshen:
@Shaoting-Feng can you help merge this? thank you!

Shaoting-Feng (approved):

PR #1870, opened 2025-10-24
[Improve|Core] Add default batched_get_non_blocking and batched_async_contains

maobaolong:
With this improvement, all the connector can use `async_loading` feature.

### DIFF
--- a/lmcache/v1/storage_backend/connector/base_connector.py
+++ b/lmcache/v1/storage_backend/connector/base_connector.py
@@ -278,25 +278,40 @@ class RemoteConnector(metaclass=abc.ABCMeta):
     ) -> List[MemoryObj]:
         """Batched get the memory_objs of the corresponding keys (non-blocking)
 
+        This method returns only the consecutive prefix of successfully retrieved
+        memory objects. Once a key is not found (None) or an exception occurs,
+        all subsequent memory objects (even if successfully retrieved) will be
+        released to avoid memory leaks, and only the prefix before the first
+        failure will be returned.
+
         Args:
             lookup_id: Identifier for this lookup operation
             keys: List of keys to get
 
         Returns:
-            List of memory objects that exist (excludes None values)
+            List of consecutive memory objects from the beginning until the first
+            failure (None or Exception). Empty list if the first key fails.
         """
         # Use asyncio.gather to fetch all keys concurrently
         results = await asyncio.gather(
             *(self.get(key) for key in keys), return_exceptions=True
         )
 
-        # Filter out None values and exceptions, return only valid memory objects
+        # Only return consecutive prefix of valid memory objects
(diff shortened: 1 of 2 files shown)


maobaolong:

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -243,37 +244,61 @@ async def batched_put(
...
+            keys: List of keys to check
+            pin: Whether to pin the keys (not used in FS connector)
+
+        Returns:
+            Number of consecutive keys that exist, starting from the first key
         """
-        Check if the remote server contains the keys
-        """
-        raise NotImplementedError
+        tasks = [self.exists(key) for key in keys]
```
YaoJiayi, line 265: We don't have batched_exists now?

YaoJiayi (approved):
lgtm

lmcache/v1/storage_backend/connector/base_connector.py
maobaolong: @YaoJiayi We have defined a sync `batched_contains` api, there are reasons why I didn't call `batched_contains` here:
1. batched_contains is not implemented by any connectors in the existing code base.
2. batched_contains is sync method, if we call that method in this async caller method, it will stuck the event loop.

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -243,37 +244,61 @@ async def batched_put(
...
-        raise NotImplementedError
+        # Use asyncio.gather to fetch all keys concurrently
+        results = await asyncio.gather(
+            *(self.get(key) for key in keys), return_exceptions=True
+        )
+
+        # Filter out None values and exceptions, return only valid memory objects
+        memory_objs = []
+        for result in results:
+            if isinstance(result, MemoryObj):
```
chunxiaozheng: if `result` is `None`, should we also add it to the list?

sammshen: furthermore, should we break early?

sammshen:
thanks for the default implementation, LGTM after @chunxiaozheng's comment is addressed

lmcache/v1/storage_backend/connector/base_connector.py
maobaolong: @sammshen Yes, I think this is a good approach and this can be the request implemented behavior of this method.

maobaolong:
@sammshen @chunxiaozheng  I push a new commit to make it early break and release the successor MemoryObj to avoid memory leak. PTAL.

sammshen (approved):
LGTM, this logic makes sense!

chunxiaozheng (approved):
lgtm

PR #1888, opened 2025-10-27
[RFC][Core] v2. memory pool facade phase1

DongDongJu:
Following v2 PR from #1831 https://github.com/LMCache/LMCache/issues/1830

  - Added a compact lmcache/v1/memory_pool.py that centralizes CPU MemoryObj borrowing/releasing, swaps itself in as the parent allocator for ref-count hygiene, keeps lightweight counters, and logs exceptions via a lease context manager.
  - Formalize backend integration with a new optional AllocatorBackendInterface.borrow_from_pool hook; 
  StorageManager instantiates the pool with that hook, and allocate/batched_allocate now call request_memory so every CPU allocation flows through the pool while preserving failure semantics.
  - tighten backend initialization: LocalCPUBackend raises LocalCPUDisabledError when the CPU budget is zero, and dependent backends (P2P/local disk/remote) now validate that a positive CPU budget is provided
  - update the FS connector to borrow through request_memory, ensuring file-system loads honor the pool without touching eviction logic
  - add a focused unit suite (tests/v1/test_memory_pool.py) covering borrow accounting, hook fallback, zero-capacity guard, and the exception-logging path

Class MemoryPool's detail:

- borrow(req) is the low-level, imperative API. You call it to grab a MemoryObj, use the object however you need, and you are responsible for calling release(memory_obj) (or letting ref-counting reach zero) when you’re done. It’s useful when your control flow doesn’t map neatly to a single scope.
- lease(req) wraps borrow and release in a context manager. You use it like with pool.lease(req) as mem: ..., and it guarantees the release happens in a finally block even if your code raises. That makes it the safer default when your usage can fit inside a with block.

  Under the hood, both paths call the same borrow/release logic; the only functional difference is that lease adds automatic cleanup and logs any exception that escapes the with block, while borrow leaves lifecycle management entirely up to the caller.

- this replaces the prior #1831 design with the slimmer façade reviewers requested; the diff shows new files because dev lacked the pool previously, but the runtime surface is just borrow/release plus a hook
- existing call sites that use LocalCPUBackend.allocate automatically route through the pool—only the FS connector needed a direct update

Addressing prior review feedback

  - tag="fs/get" clarity – The FS connector now calls request_memory(...) without a custom tag; the comment about unclear tagging is resolved (lmcache/v1/storage_backend/connector/fs_connector.py). It was originally for the future categorization of requests.
  - callability vs None check – The pool no longer probes backend methods with getattr/callable. Instead, allocator backends expose an optional borrow_from_pool hook; if it returns None, the pool falls back to the raw allocator. This satisfies the nit about avoiding callability checks and removes the brittle introspection (lmcache/v1/storage_backend/abstract_backend.py, storage_manager.py, local_cpu_backend.py).
  - “Thinner façade” concern – MemoryPool is now a ~120‑line wrapper that only handles borrow/release, ref-count bookkeeping, and logging. Metrics/Prometheus code and backend-specific logic have been dropped, so it’s nearly 1:1 with the allocator as requested (lmcache/v1/memory_pool.py).
  - Encapsulate backend integration – AllocatorBackendInterface gained borrow_from_pool, and StorageManager simply passes that hook into the pool; no more backend-specific code cluttering the manager (storage_manager.py:225). This is the “more elegant interface” the review asked for.
  - Zero-capacity guard – LocalCPUBackend now raises LocalCPUDisabledError when max_local_cpu_size <= 0, and dependent backends verify that a CPU pool exists before initializing, matching the earlier ask to fail fast (local_cpu_backend.py, storage_backend/__init__.py).
  - Exception logging coverage – Added a targeted test to assert that lease logs exceptions, preserving observability without relying on prior inline logging (tests/v1/test_memory_pool.py::test_lease_logs_exception).

Next steps toward the full rollout:
  1. Migrate remaining connectors/backends (Redis, S3, Valkey, etc.) to use StorageManager.request_memory/lease_memory directly so they no longer depend on LocalCPUBackend specifics.
  2. Expand usage of the lease_memory context manager where ref-count up/down currently split across components, reducing the chance of mismatched frees.
  3. If we want to keep the v1 drop-in API for compatibility, decide whether to deprecate StorageManager.allocate/batched_allocate now that they forward into the pool.

  Once those are done, we can widen the adoption to remote/plug-in backends and consider exposing the pool metrics externally if needed.

  If applicable:

### DIFF
--- a/lmcache/v1/memory_pool.py
+++ b/lmcache/v1/memory_pool.py
@@ -80,8 +80,6 @@ class MemoryPool:
         if self._borrow_hook is not None:
             memory_obj = self._borrow_hook(req)
         if memory_obj is None:
-            if req.dtype is None and req.fmt != MemoryFormat.BINARY_BUFFER:
-                raise ValueError("dtype must be specified for tensor allocations")
             memory_obj = self._allocator.allocate(shape, req.dtype, req.fmt)
         if memory_obj is None:
             raise RuntimeError(
@@ -183,4 +181,3 @@ class MemoryPool:
         with self._lock:
             self._live_leases = max(0, self._live_leases - 1)
             self._borrowed_bytes = max(0, self._borrowed_bytes - bytes_released)
-
--- a/lmcache/v1/memory_pool.py
+++ b/lmcache/v1/memory_pool.py
@@ -54,7 +54,12 @@ class PoolRequest:
 
 class MemoryPool:
     """
-    Thin facade that centralises engine-owned allocations and basic accounting.
+    Thin facade that centralises engine-owned allocations and ref-count hygiene.
+
+    Responsibilities:
+      * Delegate allocations to either the backend hook or the raw allocator.
+      * Swap in the pool as the parent allocator so ref_count_down() leads back here.
+      * Keep lightweight counters for observability/debug logging.
     """
 
     def __init__(
@@ -83,9 +88,9 @@ class MemoryPool:
                 f"MemoryPool failed to allocate object (fmt={req.fmt}, tag={req.tag})"
             )
 
-        self._mark_parent(memory_obj)
+        self._attach_parent_allocator(memory_obj)
         phy_size = memory_obj.meta.phy_size
-        self._increment_counters(phy_size)
+        self._record_borrow(phy_size)
         logger.debug(
             "MemoryPool borrowed %s bytes (tag=%s, fmt=%s, total_leases=%d)",
             phy_size,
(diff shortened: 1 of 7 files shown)


DongDongJu:
@sammshen Let me know what do you think

DongDongJu:
Let me re-open this after think more.

sammshen:
@DongDongJu did you mean to close this? 

I can take a look if it is ready

DongDongJu:
@sammshen Thanks for checking this. I will start to wiring up from storage backend initialization first.
Then I will re-open this. Thanks!

sammshen:
@DongDongJu looking forward to it

PR #1895, opened 2025-10-28
refactor: Use named tuple instead of tuple for ProcessTokensResult

maobaolong:
This is a demo for show how named tuple can make code clear than using tuple directly.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -7,8 +7,8 @@ from typing import (
     Dict,
     Generator,
     List,
+    NamedTuple,
     Optional,
-    Tuple,
     Union,
 )
 import asyncio
@@ -52,6 +52,26 @@ from lmcache.v1.token_database import (
 )
 
 logger = init_logger(__name__)
+
+
+class ProcessedChunk(NamedTuple):
+    """Represents a processed chunk with its key, memory object, and token range.
+
+    Note: key can be None for passive ranks in distributed scenarios where
+    the key is not needed for local operations.
+    """
+
+    key: Optional[CacheEngineKey]
+    memory_obj: MemoryObj
+    start: int
+    end: int
+
+
+class ProcessTokensResult(NamedTuple):
+    """Result of processing tokens, containing chunks and total KV size."""
+
+    chunks: List[ProcessedChunk]
+    total_kv_size: int
 
 
 class CacheEngineEndSignal:
@@ -457,22 +477,27 @@ class LMCacheEngine:
 
         ret_mask = torch.zeros(len(tokens), dtype=torch.bool, device="cpu")
 
-        reordered_chunks: List[Tuple[CacheEngineKey, MemoryObj, int, int]] = []
+        reordered_chunks: List[ProcessedChunk] = []
+        tot_kv_size = 0
         if not self._is_passive():
             if self.async_loading:
-                reordered_chunks, tot_kv_size = self._async_process_tokens_internal(  # noqa: E501
+                result = self._async_process_tokens_internal(  # noqa: E501
                     tokens,
                     mask,
                     ret_mask,
                     **kwargs,
                 )
+                reordered_chunks = result.chunks
+                tot_kv_size = result.total_kv_size
             else:
-                reordered_chunks, tot_kv_size = self._process_tokens_internal(
+                result = self._process_tokens_internal(
                     tokens,
                     mask,
                     ret_mask,
                     **kwargs,
                 )
+                reordered_chunks = result.chunks
+                tot_kv_size = result.total_kv_size
         if self.save_only_first_rank:
             with torch.cuda.stream(self.broadcast_stream):
                 self._broadcast_or_receive_memory_objs(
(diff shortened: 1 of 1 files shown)


maobaolong:
@ApostaC @DongDongJu Would you like to take a look at this PR? Thanks!

lmcache/v1/cache_engine.py
```diff
@@ -457,22 +477,27 @@ def retrieve(
...
             else:
-                reordered_chunks, tot_kv_size = self._process_tokens_internal(
+                result = self._process_tokens_internal(
                     tokens,
                     mask,
                     ret_mask,
                     **kwargs,
                 )
+                reordered_chunks = result.chunks
+                tot_kv_size = result.total_kv_size
```
ApostaC: Can we do something like the following to simplify the code a bit:
```python
if not self._is_passive():
    process_func = self.self._async_process_tokens_internal \
                    if self.async_loading else \
                    self.self._process_tokens_internal
    reordered_chunks, tot_kv_size = process_func(
                tokens,
                mask,
                ret_mask,
                **kwargs
    )
```

lmcache/v1/cache_engine.py
```diff
@@ -494,17 +519,21 @@ def retrieve(
...
-            _, memory_objs, starts, ends = zip(*reordered_chunks, strict=False)
-            self.gpu_connector.batched_to_gpu(
-                list(memory_objs), list(starts), list(ends), **kwargs
-            )
+            memory_objs, starts, ends = [], [], []
+            for chunk in reordered_chunks:
+                memory_objs.append(chunk.memory_obj)
+                starts.append(chunk.start)
+                ends.append(chunk.end)
+            self.gpu_connector.batched_to_gpu(memory_objs, starts, ends, **kwargs)
```
ApostaC: The original code should still work here.

Here's a quick example test:
```python
from typing import NamedTuple

class Point(NamedTuple):
    x: int
    y: int

list_of_points = [Point(1, 2), Point(3, 4), Point(5, 6)]
x, y = zip(*list_of_points)
print(x)
print(y)
```

The output is:
```
(1, 3, 5)
(2, 4, 6)
```

lmcache/v1/cache_engine.py
```diff
@@ -494,17 +519,21 @@ def retrieve(
...
         # TODO(Jiayi): Need to refactor the `remove_after_retrieve` logic.
-        for key, memory_obj, _, _ in reordered_chunks:
+        for chunk in reordered_chunks:
             if self.remove_after_retrieve and not self._is_passive():
-                self.storage_manager.remove(key)
-            memory_obj.ref_count_down()
+                # chunk.key can be None for passive ranks
+                if chunk.key is not None:
+                    self.storage_manager.remove(chunk.key)
+            chunk.memory_obj.ref_count_down()
```
ApostaC: The original code should also work here, since NamedTuple supports all the tuple operations.

lmcache/v1/cache_engine.py
```diff
@@ -1229,14 +1258,14 @@ def _broadcast_or_receive_memory_objs(
             self.broadcast_object_fn(chunk_count, self.metadata.first_rank)
 
             # Broadcast each chunk's data
-            for key, memory_obj, start, end in reordered_chunks:
+            for chunk in reordered_chunks:
```
ApostaC: The old code should also work here. Doing `chunk.xxx` may incur extra access overheads.

ApostaC (changes requested):
Thanks @maobaolong! I like the idea!

Just some nit comments.

lmcache/v1/cache_engine.py
maobaolong: Done

lmcache/v1/cache_engine.py
maobaolong: Reverted

lmcache/v1/cache_engine.py
maobaolong: Reverted

lmcache/v1/cache_engine.py
maobaolong: Reverted

maobaolong:
@ApostaC Thanks for your review and the suggested changes, PTAL.

ApostaC (approved):
LGTM! Let's use the comprehensive test to check whether there is any potential performance degradation.

ApostaC:
I found the comprehensive test is somehow impacted by the Docker image cache on the CI server. Trying to fix that...

DongDongJu:
@maobaolong This looks much clean.

sammshen (approved):
This is a nice clean up!

maobaolong:
@ApostaC @DongDongJu @sammshen Thanks for the approve and the previous review, hope we turn to the https://github.com/LMCache/LMCache/pull/1953

sammshen:
can we close this @maobaolong after #1953 is merged?

maobaolong:
> can we close this @maobaolong after #1953 is merged?

@sammshen Thanks for your remind

Issue #1901, opened 2025-10-28
[Bug] LMCacheConnectorV1 init fails with ValueError: both buffer length (0) and count (-1) must not be 0 if max_local_cpu_size: 0 is present when local_cpu: false


hyunnnchoi:
**Label**
Please label this issue with "**bug**" and "**vllm**" so that it can easily be categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/627).

**Describe the bug**
When configuring KV-Cache offloading using vLLM (`--tensor-parallel-size 4`) and `LMCacheConnectorV1`, setting `local_cpu: false` along with `max_local_cpu_size: 0` causes worker initialization to fail with a `ValueError`.

The `LMCache INFO` log confirms that the `lmcache` engine is configured with 'local_cpu: False` and 'max_local_cpu_size: 0.0`. However, the worker process proceeds to initialize the `LocalCPUBackend` with a size of zero, leading to a `ValueError` when `torch.frombuffer` is called with a buffer length of zero.

The initialization only succeeds if the `max_local_cpu_size` key is entirely removed/commented out from the `lmcache_config.yaml`.

**To Reproduce**
1. Configure vLLM (v0.11.0) to use LMCache V1 Connector:
``` Bash
# vLLM launch arguments (Example)
  ... other configurations...
  -e LMCACHE_CONFIG_FILE=/config_dir/lmcache_config.yaml \
  --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
```
2. Use a `lmcache_config.yaml` file with the following configuration:
``` YAML
# lmcache_config.yaml (Non-working configuration)
local_cpu: false        # CPU caching is disabled
max_local_cpu_size: 0   # But the size parameter is present (even set to 0)
local_disk: /lmcache
max_local_disk_size: 10737418240
# ... other settings
```
3. Launch the vLLM API server with multi-process workers (e.g., `--tensor-parallel-size 4`).
4. See the engine initialization fail with the error in the logs.

**Expected behavior**
When `local_cpu` is set to `false`, the underlying LMCache storage manager (`StorageManager`) should **completely skip** the creation of the `LocalCPUBackend` instance. This would prevent any subsequent logic (like allocator initialization) that results in the `ValueError`. The presence or value of `max_local_cpu_size` should be safely ignored.

**Screenshots**

[...]

PR #1905, opened 2025-10-28
[hotfix] guard local CPU backend creation

DongDongJu:
Bug fix for https://github.com/LMCache/LMCache/issues/1901

But need to think about app shutdown option in case that combination of parameters we are not considering.

### DIFF
--- a/lmcache/v1/storage_backend/__init__.py
+++ b/lmcache/v1/storage_backend/__init__.py
@@ -120,7 +120,7 @@ def CreateStorageBackends(
     # TODO(Jiayi): The hierarchy is fixed for now
     # NOTE(Jiayi): The local_cpu backend is always created because
     # other backends might need it as a buffer.
-    if not config.enable_pd or config.local_cpu:
+    if not config.enable_pd or (config.local_cpu and config.max_local_cpu_size > 0):
         local_cpu_backend = LocalCPUBackend(
             config,
             metadata,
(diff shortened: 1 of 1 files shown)


DongDongJu:
> ## Code Review
> This pull request aims to fix a bug where the `LocalCPUBackend` is created with a size of 0, causing a crash. The proposed change correctly guards the backend creation when `config.enable_pd` is true, but the bug persists when `config.enable_pd` is false. I've suggested a change to the condition to cover all cases and prevent the application from crashing.

Thanks for catching. My bad

lmcache/v1/storage_backend/__init__.py
```diff
@@ -120,7 +120,7 @@ def CreateStorageBackends(
     # TODO(Jiayi): The hierarchy is fixed for now
     # NOTE(Jiayi): The local_cpu backend is always created because
     # other backends might need it as a buffer.
-    if not config.enable_pd or config.local_cpu:
+    if (not config.enable_pd or config.local_cpu) and config.max_local_cpu_size > 0:
```
YaoJiayi: I think LMCache needs the cpu backend in call cases for now?

YaoJiayi (commented):
Thanks! Left a comment

lmcache/v1/storage_backend/__init__.py
DongDongJu: @YaoJiayi. you mean all cases? I think PD, GDS, NIXL? can run without it. Am I wrong?
If I understand it wrong way then please correct me @YaoJiayi @ApostaC

YaoJiayi: Oh you are right. In this case, can we print some logs like "no cpu memory is allocated as max_local_cpu_size <=0"?

DongDongJu: Applied!

YaoJiayi (approved):

PR #1907, opened 2025-10-29
[Core][2/N] Example LMCache independent cache server for multi-process mode

ApostaC:
This PR is a follow-up on #1853 

It provides an example to run LMCache as a separate process and be able to handle store/retrieve requests from different (vLLM) processes.

**In the follow-up PRs:**
- Pass finished CUDA events to vLLM
- Update the server to make sure that ZMQ usage is thread-safe
- Integrate with existing storage backends + thread safe management
- CI testing in LMCache
- Update the IPCCacheEngine key data structure
  - when registering, have model name, world rank, etc
  - when storing/retrieving, have block hashes (list of bytes) only

**If applicable**:


ApostaC:
WIP vLLM-side changes vllm-project/vllm#27902

lmcache/v1/multiprocess/server.py
```diff
@@ -0,0 +1,446 @@
...
+        """
+        Returns the shape of the KV buffer for the given number of tokens
+        """
+        if self.is_mla_:
+            return torch.Size((self.num_layers_, num_tokens, self.hidden_dim_size_))
+        else:
+            return torch.Size((2, self.num_layers_, num_tokens, self.hidden_dim_size_))
+
+
+class MPCacheEngine:
```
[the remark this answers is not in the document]
ApostaC: This will be fixed in follow-up PRs

sammshen:
taking a look!

YaoJiayi:
@ApostaC I've got several high-level questions:
1. What's the perf compared with our current single process connector?
2. Is there a reason for us to have a separate connector for multi process? maybe code maintainability is going to be an issue?
3. For current single-process lmcache, maybe I'm wrong, but I think we already have thread-safety across cache engine and storage backends?

Also, just curious, how did you solve the synchronization problem we encountered last time?

I'll take a closer look later!

ApostaC:
@YaoJiayi 
1. For perf, it has much smaller overhead during store. It will have a constant TTFT overhead (~100ms) during retrieve since it uses vLLM async code path, which will force retrieve and inference to be in 2 scheduler steps.
2. There is a core part of the code that is different -- the store and retrieve are now submitted as a job instead of executed as a function call. Therefore, it's quite different from the existing connector. Also, now it's pure block-based (which means it does not do token_ids + hashing)
3. While the current storage backend could be thread-safe, there are still some potential read-after-write race conditions. I can explain to you offline

ApostaC:
The synchronization problem is actually caused by a different problem... Anyway, it's solved now.

Right now, LMCache won't do any blocking wait for the kernel launches. Everything will be pure async and the call to `store` and `retrieve` could be non-blocking (there is one small technical issue but can be addressed soon).

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -66,16 +66,19 @@ class IPCCacheEngineKey:
```
sammshen, line 66: nit: in a future PR, we should add dtype to the IPCCacheEngineKey

lmcache/v1/multiprocess/protocol.py
```diff
@@ -45,6 +45,17 @@ class RequestType(enum.Enum):
     RETRIEVE = enum.auto()
     LOOKUP = enum.auto()
```
sammshen, line 46: how is unlock sent to LMCache server?

sammshen (approved):
LGTM!!

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -66,16 +66,19 @@ class IPCCacheEngineKey:
...
     @staticmethod
     def Bytes2IntHash(chunk_hash: bytes) -> int:
+        # NOTE: this is only used by tests
         return int.from_bytes(chunk_hash, byteorder="big") & ((1 << 64) - 1)
 
     @classmethod
     def from_int_hash(
         cls, model_name: str, world_size: int, worker_id: int, chunk_hash: int
     ) -> "IPCCacheEngineKey":
+        # NOTE: this is only used by tests
```
YaoJiayi, line 81: We can remove these test-only functions once the feature becomes stable.

lmcache/v1/multiprocess/mq.py
```diff
@@ -443,6 +396,10 @@ def _send_response(fut: Future):
             except Exception as e:
                 logger.error("Error in blocking handler: %s", e)
 
+        # TODO: HERE'S A BUG: WE CANNOT SEND RESPONSE IN THE FUTURE THREAD
+        # BECAUSE THE OUTPUT ZMQ SOCKET IS NOT THREAD-SAFE.
+        # WE SHOULD USE A ZMQ SOCKET TO NOTIFY THE MAIN THREAD TO SEND THE
+        # RESPONSE AND USE THE THREAD-QUEUE TO PASS THE RESPONSE DATA
```
YaoJiayi, line 402: Just curious: why is the concept of `BlockingRequest` needed at all?

lmcache/v1/multiprocess/protocol.py
```diff
@@ -45,6 +45,17 @@ class RequestType(enum.Enum):
...
+    # Controller operations
+    CLEAR = enum.auto()
+    # REMOVE = enum.auto()
+    # PERSIST = enum.auto()
+    # UNPERSIST = enum.auto()
+    # MOVE = enum.auto()
+    # COMPRESS = enum.auto()
+
+    # For configuration read commands (vllm integration)
+    GET_CHUNK_SIZE = enum.auto()
```
YaoJiayi, line 57: Are we assuming that vllm doesn't have any lmcache config? If so, probably only knowing the chunk size is not enough?

lmcache/v1/multiprocess/protocol.py
```diff
@@ -137,6 +150,19 @@ def get_handler_type(req_type: RequestType) -> HandlerType:
...
+        payload_classes=[],
+        response_class=None,
+        handler_type=HandlerType.BLOCKING,
+    ),
+    # Configuration read commands
+    RequestType.GET_CHUNK_SIZE: ProtocolDefinition(
+        payload_classes=[],
+        response_class=int,
+        handler_type=HandlerType.SYNC,
+    ),
```
YaoJiayi, line 165: So we only have `SYNC` and `BLOCKING` operation? Shouldn't they be sth like `BLOCKING` and `NONBLOCKING`?

lmcache/v1/multiprocess/server.py
```diff
@@ -0,0 +1,513 @@
...
+            torch.cuda.stream(gpu_context.stream),
+        ):
+            event = torch.cuda.Event(interprocess=True)
+            slot_mapping_tensor = gpu_context.get_slot_mapping_tensor(gpu_block_ids)
+
+            # Wait for vLLM to finish
+            vllm_event = torch.cuda.Event.from_ipc_handle(
+                gpu_context.device, event_ipc_handle
+            )
+            vllm_event.wait(stream=gpu_context.stream)
```
YaoJiayi, line 251: Is this implementation optimal here? We have to wait for all the vllm jobs to finish in order to start `store`?

YaoJiayi (approved):
LGTM! I feel like, in general, we need to align the multiprocess code (e.g., reuse the code) with the single-process code as much and as soon as possible to reduce the future engineering effort as we need to (1) always keep the single-process code for easy debugging and better perf in some cases (e.g., lower TTFT) and (2) maintain the compatibility with features like layerwise and CacheBlend. Also,

lmcache/v1/multiprocess/mq.py
ApostaC: To clarify, there are 3 types of requests:
- `sync`: small functions that can be run directly in the main loop
- `blocking`: a function that uses blocking I/O or cudaDeviceSynchronize (may block the whole thread and gives up GIL). This function needs to be launched in a thread pool worker. (And usually it may not be python-asyncio-compatible)
- `non-blocking`: a function that just launches a job non-blockingly (e.g., NIXL send). It requires another worker to periodically check whether the job is finished or not.

lmcache/v1/multiprocess/protocol.py
ApostaC: Right now, the vLLM-side integration only needs the lmcache chunk size. We can extend this in the future if vllm needs more configurations.

lmcache/v1/multiprocess/protocol.py
ApostaC: See my clarification above.

lmcache/v1/multiprocess/server.py
ApostaC: This wait does not block the CPU thread. It's actually a non-blocking call that tells the GPU to start the new kernels after `vllm_event`.

Issue #1826, opened 2025-10-20
[RFC] LMCache & Agentic Application/Benchmark/Workflow Traces


kobe0938:
## Goal

Collect a set of agentic applications and workflows to demonstrate that LMCache is suitable for running agentic applications/workflows in terms of context reuse, ultimately reducing TTFT (Time to First Token) and ITL (Inter-Token Latency).

## Motivation

We want to showcase that LMCache is not only limited to CPU offloading. The plan is to collect real traces to prove that LMCache can speed up inference in agentic settings. Agentic workloads typically involve multiple iterative LLM calls within a session; this pattern benefits from cache reuse. This issue is a long-lasting tracker and a call for help.

## Plan

A new folder will be added:

```
lmcache/examples/agents/
```

Each subfolder will correspond to a benchmark, application, or workflow:

```
examples/agents/<benchmark_or_application_name>/
 ├── README.md          # Brief intro to the benchmark/application/workflow and how to run it
 ├── trace.jsonl        # Example converted trace
 └── converter.py       # (Optional) script to convert raw logs/outputs to the unified trace format
```

### Unified Trace Format

Different repos log LLM calls, function calls, intermediate results, and faults in different ways. We only care about the LLM calls that reach the vLLM server, as those are relevant to LMCache.

**trace.jsonl**
```jsonl
{"timestamp": 0, "input": "input prompt in raw string", "output": "output response in raw string", "session_id": 0}
{"timestamp": 123, "input": "input prompt in raw string", "output": "output response in raw string", "session_id": 0}
```

- `timestamp`: relative timestamp from 0.
- `input` / `output`: raw strings for each LLM call.
- `session_id`: if one query to an agent executes N rounds and incurs N LLM calls, all entries share the same `session_id`. This enables overlapping different users/sessions to make the dataset more condensed. `session_id` is experimental and optional; the key fields are `timestamp`, `input`, and `output`.

Under each benchmark/application/workflow folder, include an example converted trace and, optionally, a `converter.py` to help others convert the repo’s raw outputs to the unified format.

## Analysis and Visualization

After obtaining unified traces, I will provide an analysis script to study the relationship between RAM pool size and hit rate for both:

1. Prefix string match (LMCache case).
2. Longest common substring match (CacheBlend case).

The physical meaning of the graphs is: given a unified RAM pool (GPU/CPU/Disk) for KV-cache storage, how much hit rate can be achieved.

For example:
[...]

PR #1912, opened 2025-10-29
[Agents] Improve LRU prefix matching with reverse iteration and early exit on full match

JZhou3083:
Part of https://github.com/LMCache/LMCache/issues/1826, following the work in this https://github.com/LMCache/LMCache/pull/1851. This PR improves the efficiency of prefix matching in the LRU token pool by:

1. Reversing the iteration order over the request pool to prioritise more recently used entries, increasing the likelihood of early hits.
2. Adding early stopping logic to exit the search loop immediately when a full prefix match is found, reducing unnecessary comparisons.
@kobe0938  Could you have a look please.

### DIFF
--- a/examples/agents/prefix_analysis.py
+++ b/examples/agents/prefix_analysis.py
@@ -68,7 +68,6 @@ class LRUTokenPool:
                 if common_len == len(tokens):
                     break
 
-
         # Update LRU ordering
         if best_id != -1:
             self.requests.move_to_end(best_id)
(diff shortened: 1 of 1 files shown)


kobe0938:
Thanks for the contribution, but I'm not sure it would work out because following the 'react' pattern(or even other agentic patterns), it's impossible to get a 'perfect match' - the incoming request will always be a superset of the previous ones. Therefore reversing it will also makes little sense.

JZhou3083:
> Thanks for the contribution, but I'm not sure it would work out because following the 'react' pattern(or even other agentic patterns), it's impossible to get a 'perfect match' - the incoming request will always be a superset of the previous ones. Therefore reversing it will also makes little sense.

Makes sense. Thanks for the reply.

PR #1918, opened 2025-10-29
Fix/token trim and fractional labels

yamazakihirofumi:
Part of #1826.
This PR fixes an issue in the analysis scripts where oversized requests could exceed the configured `max_tokens` limit, causing inflated hit rates and incorrect queue-size accounting.
Changes: 
1. Trim oversized requests before admission to ensure `current_tokens ≤ max_tokens` at all times.
2. Update x-axis labeling in the plotting script so fractional (<1 GB) pool sizes are displayed correctly.
3. Add an example plot generated by updated prefix_analysis.py


YaoJiayi:
@kobe0938

kobe0938 (commented):
Thanks for the fix @yamazakihirofumi :)
I also include those fixes (in a slightly different way, which covers some corner cases) here: https://github.com/LMCache/lmcache-agent-trace/blob/main/prefix_analysis.py. It also resolves two more issues: 1. LRU does not apply to the substring match case (the hit rate remains high even during smaller pool sizes) 2. log scale

Could you give them a look and check if they are proper? Thanks.

PR #1920, opened 2025-10-30
[optimize] catch exception in batched_contains

chunxiaozheng:
catch exception in `batched_contains`


chunxiaozheng:

maobaolong (approved):
Good catch. LGTM

sammshen (approved):
thanks for this fix and sorry for the late review!

PR #1921, opened 2025-10-30
[metrics]filter out 0 hit rate

chunxiaozheng:
0 as a completely missed situation, use `interval_lookup_0_hit_requests` to separately count lookup requests with a hit rate of 0.

### DIFF
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -76,7 +76,10 @@ class LMCacheStats:
     p2p_transfer_speed: List[float]  # Tokens per second
 
     # request lookup hit rates
+    # use bucket of interval_lookup_hit_rates to represents non-0 hit requests
+    # use interval_lookup_0_hit_requests to represents 0 hit requests
     interval_lookup_hit_rates: List[float]
+    interval_lookup_0_hit_requests: int
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


chunxiaozheng:

ApostaC (commented):
The changes look clear to me. Just wanted to understand more of the problem. Why should we filter out zeros when reporting the hit rates?

chunxiaozheng:
> The changes look clear to me. Just wanted to understand more of the problem. Why should we filter out zeros when reporting the hit rates?

@ApostaC Thanks for your review. In our scenario, it was found that the lookup hit rate of most requests falls within the 0-10 bucket, making it difficult for the business to distinguish between 0 hits and hits. Therefore, we hope to separately count requests with a hit rate of 0.

maobaolong (approved):
Thanks for this improvement, it LGTM.

lmcache/observability.py
```diff
@@ -486,11 +494,13 @@ def filter_out_invalid(stats: List[float]):
             [stats.transfer_speed() for stats in self.p2p_requests.values()]
         )
 
-        request_lookup_hit_rates = [
-            stats.hit_rate()
-            for stats in self.lookup_requests.values()
-            if stats.is_finished
-        ]
+        request_lookup_hit_rates = filter_out_invalid(
```
ApostaC: Can we change the `filter_out_invalid` to `filter_out_invalid` to make it clearer?

ApostaC (approved):
Otherwise LGTM!

lmcache/observability.py
chunxiaozheng: @ApostaC Thanks for your review! But I seem a bit confused about change the `filter_out_invalid` to `filter_out_invalid`.

ApostaC: Oh sorry! I mean change to `filter_out_zeros`

chunxiaozheng: Sure, I have updated!

PR #1361, opened 2025-08-15
[bug]: vllm preemption handling

sammshen:
Previously, Hao Chen (sorry I do not know your github at) identifies that we are not properly loading from LMCache for preempted requests even though we generate a new load spec for them inside of `get_num_new_matched_tokens()` when they are resumed. This PR attempts to fix the problem. 

1. Need to test on some workload that will generate preemption. Any recommendations? 

2. Need also to adapt for `vllm <0.9.2`
@maobaolong @chunxiaozheng any ideas based on the old vllm scheduler output format?

PR #1922, opened 2025-10-30
[feature] reduce redundant lookup [1/3]: implement in sync mode get_num_new_matched_tokens() 

ziruiliu:
**Why**
When vllm's prefix cache and LMCache are both enabled, vLLM is aware of how many tokens are there in GPU kv cache and pass `num_computed_tokens` to LMCache's `get_num_new_matched_tokens()` . However LMCache still lookup all the tokens in request, got number of hit tokens, then only uses `num_computed_tokens` to calculate the number of tokens that need to be retrieved from LMCache. That causes unnecessary lookup calls on those token existed in vLLM's cache, when lookup() is heavy for certain types of remote storage, the penalty in lookup latency is considerable.

It is even worse when async kv loading is enabled. Tokens existed in vLLM's prefix cache are lookup and prefetched in cache engine's `async_lookup_and_prefetch()`, meaning unnecessary read I/O are issued to remote storage. Though such I/O are done asynchronously and parallelly, they are still waste of network bandwidth

**What**
Simply truncating `token_ids` with`num_computed_tokens` in `get_num_new_matched_tokens()` is not the solution, because key hash calculation depends on entire sequence of `token_ids`. So in this change, `num_computed_tokens` is passed down through lookup client/server to cache engine. In cache engine's `async_lookup_and_prefetch()` and `lookup()`, keys are calculated with entire token_ids then the number of tokens to be skipped is calculated from `num_computed_tokens`. Doing so will reduce the calls to storage manager's contains().

To keep it simple and minimize the changes to vllm_v1_adapter.py, which will also be included in vllm's code. I added a new argument `num_computed_tokens` to `lookup_client.lookup()`. The returned value `num_external_hit_tokens` retains its original semantics and still counts the skipped tokens. I also  modified log to be like
```
"Scheduled to load %d tokens (%d cached in vLLM) for requst %s"
```

**Tests**
Tests are done with remote backend + fs connector, chunk size as 64. A 300-token prompt is used with save_unfull_chunk=True, and a 256-token prompt is used with save_unfull_chunk=False. Both are tested in both async KV loading mode and sync mode. Results are expected like:
```
# only lookup and retrive the last chunk
LMCache DEBUG: Scheduled to load 256 tokens (240 cached in vLLM) for request...
LMCache INFO: Retrieved 64 out of 64 required tokens (from 256 total tokens).
...
LMCache DEBUG: Scheduled to load 300 tokens (288 cached in vLLM) for request
LMCache INFO: Retrieved 44 out of 44 required tokens (from 300 total tokens).
```


ziruiliu:
@sammshen as you mentioned in another PR, #1867, this PR pass `num_computed_tokens` into lookup, to avoid unnecessary lookup and prefetch

ziruiliu:
@maobaolong @yoo-kumaneko please also take a look at this PR

YaoJiayi (changes requested):
@ziruiliu Can you elaborate on how preemption is handled in this case?

ziruiliu:
> @ziruiliu Can you elaborate on how preemption is handled in this case?

@YaoJiayi  if you are asking about the preemption in vLLM's kv-cache, I believe there is not much difference before and after this change:
in vLLM's scheduler.py, once a request is preempted around line 290 to 300, its preempted_req.num_computed_tokens will be set as 0
```python
                preempted_req.status = RequestStatus.PREEMPTED
                preempted_req.num_computed_tokens = 0
```
then this requests will be rescheduled later. Once the request is scheduled again, the number of gpu-cached tokens will be re-calculated in vllm kv_cache_manager, and then get_num_new_matched_tokens() will be invoked again.

To sum up, kv_cache_manager.get_computed_blocks() and connector.get_num_new_matched_tokens() are always invoked together in scheduler and preemption does not change any number which causes mismatch of cache hit numbers in vllm and lmcache.

lmcache/v1/cache_engine.py
```diff
@@ -701,6 +716,17 @@ def lookup(
             # TODO: support batched_contains when layerwise is enabled
             if self.use_layerwise:
                 for start, end, key in chunk_info_iterator:
+                    if end <= skip_n_tokens:
+                        continue
+                    if start < skip_n_tokens < end:
```
maobaolong: As you aligned num_computed_tokens to chunk size, so for skip_n_tokens, `start < skip_n_tokens < end` cannot be `True`, if you want ensure it, you can add an assertion here.

maobaolong (approved):
Thanks for this improvement, this PR LGTM overall, left a minor nit inline.

lmcache/v1/cache_engine.py
ziruiliu: changed to assert per reviewer's request. also added this param in new bypass lookup client

lmcache/v1/cache_engine.py
```diff
@@ -716,6 +731,9 @@ def lookup(
             # TODO: support batched_contains when layerwise is enabled
             if self.use_layerwise:
                 for start, end, key in chunk_info_iterator:
+                    if end <= skip_n_tokens:
+                        continue
+                    assert not (start < skip_n_tokens < end)
```
sammshen: this is because we aligned to chunk size?

lmcache/v1/cache_engine.py
```diff
@@ -694,17 +695,31 @@ def lookup(
...
+        skip_n_tokens = (
+            min(num_computed_tokens, total_length)
+            // self.config.chunk_size
+            * self.config.chunk_size
+            if num_computed_tokens > 0
+            else 0
+        )
 
-        res = 0
+        res = skip_n_tokens
```
sammshen, line 778: good catch

lmcache/v1/cache_engine.py
```diff
@@ -857,8 +883,23 @@ def async_lookup_and_prefetch(
...
+            min(num_computed_tokens, total_length)
+            // self.config.chunk_size
+            * self.config.chunk_size
+            if num_computed_tokens > 0
+            else 0
+        )
+
         keys: list[CacheEngineKey] = []
-        cum_chunk_lengths = [0]
+        cum_chunk_lengths = [skip_n_tokens]
```
sammshen: good catch

sammshen (approved):
thanks for explaining the preemption case, LGTM

yoo-kumaneko:
LGTM. If the rescheduled victim request triggers a new lookup entry, then that’s fine for preemption.

maobaolong:
Discussed with @yoo-kumaneko offline about this PR, for the preemption case, it will have a chance if the original prefix cache chunk has been evicted by vllm, this should be handled. @ziruiliu

maobaolong:
Clipboard_Screenshot_1762652658

This code is from `scheduler.py` of vllm, it shows that vllm freed the blocks after preemption.

ziruiliu:
This change is supposed to work properly in sync mode, however the vllm preemption does not work with current LMCache code, see PR #1361 and I will also prepare a detailed issue report soon. With that fix to handle preemption properly, this change works perfect. When a preempted request is rescheduled, with freed prefix cache, the number of hit tokens in prefix cache and in external storage will be calculated again.

[this is not a problem] However there could be a problem in async KV loading mode, because lookup client remembered results in reqs_status[lookup_id], so if a preempted request is rescheduled, async lookup() alway returns old number

ziruiliu:
As I posted in issue #1969 ,  preemption handling in lmcache has some fundamental problems in `build_connector_meta`, current code won't work as expected.

With changes to `build_connector_meta`, preemption handling works well,  no matter with or without my change.

In asyn loading mode, with my change, we do not to worry about the reqs_status[lookup_id] still holding stable information about external storage hit.  vLLM will trigger `update_state_after_alloc` in the connector, before actual retrieving data, and `update_state_after_alloc` will call `self.lookup_client.clear_lookup_status(request.request_id)` to clear the result in reqs_status[lookup_id]. When the request is rescheduled from preemption, the reqs_status[lookup_id] is cleared so lookup and prefetch wil called freshly

yoo-kumaneko:
@YaoJiayi We don't need to worry about the preemption of the requests being async looked up. Only running requests can be preempted. However, the prefix cache on the GPU can change during the async lookup and prefetch process. For example, if another request that shares a common prefix with the request being looked up finishes (or gets preempted), those GPU blocks lose the ref count and can be evicted. This is no problem because we will get to match the prefix again before we enter the get_num_new_matched_tokens. If some blocks are evicted, we get the new status and are able to launch a new async lookup.
```
# Get already-cached tokens.
if request.num_computed_tokens == 0:
    # Get locally-cached tokens.
    new_computed_blocks, num_new_local_computed_tokens = \
        self.kv_cache_manager.get_computed_blocks(
            request)

    # Get externally-cached tokens if using a KVConnector.
    if self.connector is not None:
        num_external_computed_tokens, load_kv_async = (
            self.connector.get_num_new_matched_tokens(
                request, num_new_local_computed_tokens))
```

ziruiliu:
Status Update:
There are few remaining issues with this PR
**1. Chunk Not Retrieved when request is once cached**
as descibed in issue #1969  and PR #1361, `build_connector_meta` has to be amended to pass proper load_spec if request is cached. Following code change and tests are done with a temp patch to this.

**2. Chunk Lost in Preemption / Race condition**
 As discussed with @yoo-kumaneko and @maobaolong , not only after preemption but also in other conditions there would be missing chunks, in async loading mode

for example
- request A is initialized when there are 2 block available in GPU after calling vllm kv_cache_manager get_computed_blocks().
- scheduler calls get_num_new_matched_tokens() and got `None` at first time, async lookup and prefetch issued, expecting hit 1 chunk (the 3rd block after those 2) and load the chunk to memory. 
- scheduler has 
- scheduler continue to schedule other requests. somehow the 2 blocks available for request A are used and written by other request ( becasue request A simply called get_computed_blocks but never pin these 2 blocks ). Preemption could cause. If other request share the same prefix cache with request A and are finished before A is rescheduled, blocks are freed too.
- request A is rescheduled, after calling `get_computed_blocks` there are 0 block in GPU cache. But calling to get_num_new_matched_tokens() will get a cached result in reqs_status(), no actual prefetch is triggered.
- when engine's 

**3. Chunk is not picked correctly if skipped**
Current async KV loading assumes all hit chunks are load into memory in order.
In `_async_process_tokens_internal`, the prefetched mem obj layout is like:
GPU   | Block 0 | Block 1 |
LMC.  | Block 0 | Block 1 | Block 2 | Block 3 |
Then `_async_process_tokens_internal` iterates block 0 to 3, picks block 2 and block 3, returning to engine's retrieve().

With the optimization in this PR, the prefetched mem obj layout is like:
GPU | Block 0 | Block 1 |
LMC | Block 2 | Block 3 | 
How current `_async_process_tokens_internal` uses start,end to locate block in the prefetched memory object list with a simple index. 
Therefore `_async_process_tokens_internal` tried to use index 2 to pick Block 2, however the list starts from 0, and index 2 is out of range.
This issue is also mentioned in https://github.com/LMCache/LMCache/pull/1867#issuecomment-3444784949
This issue has to be fixed by using cache engine key not list index, to locate prefetched memory object. This will introduce a large change to current code

**Conclusion**
I suggest to take step back. I will modify code to only enable optimization in SYNC lookup mode, and prepare a fix to the problem 3  then file a PR for async mode

ziruiliu:
@YaoJiayi @maobaolong @yoo-kumaneko @sammshen 
Since there are multiple remaining problems in async KV loading mode, with optimization that skips tokens already in GPU cache. I suggest to split these tasks to avoid a big bulk of change in one PR
1. Let's simply pass `num_computed_token` from adapter to lookup client/server to cache engine, and enable the skip in sync mode at first
2. In following tasks, let's fix `_async_process_tokens_internal` to identify memory objects not by list index but CacheEngineKey
3. Enable skip in async KV loading, by adding more information in reqs_status[] and trigger lookup and prefetch again if preemption happens

lmcache/v1/cache_engine.py
```diff
@@ -715,17 +716,31 @@ def lookup(
...
         """
 
         if tokens is not None:
             lookup_request_id = self.stats_monitor.on_lookup_request(len(tokens))
+            total_length = len(tokens)
         else:
             assert offsets is not None
             assert hashes is not None
             lookup_request_id = self.stats_monitor.on_lookup_request(sum(offsets))
+            total_length = sum(offsets)
```
ApostaC: Is `sum` here logically correct? I cannot remember the use case of `offsets` here but this should be introduced by @YaoJiayi IIRC.
@YaoJiayi Can you double check?

lmcache/v1/cache_engine.py
```diff
@@ -715,17 +716,31 @@ def lookup(
...
+
+        # Skip the number of tokens that are already computed, align to chunk size
+        skip_n_tokens = (
+            min(num_computed_tokens, total_length)
```
ApostaC: Is there any case that `num_computed_tokens` will be larger than `total_length`?

lmcache/v1/cache_engine.py
ApostaC: Same ask. Lines 735-741 already make sure that skip_n_tokens is aligned with the chunk size. We probably don't need this assertion?

lmcache/v1/cache_engine.py
ApostaC: Will this impact the lookup hit rate calculation? Say if LMCache doesn't have any match but vLLM has a match, will this cause over-estimation of the lookup cache hit rate?

ApostaC (commented):
Just a few quick comments. Mainly correctness related.

yoo-kumaneko (approved):

lmcache/v1/cache_engine.py
ziruiliu: No. If vllm works properly, num_computed_tokens is always equal or less than total_length

lmcache/v1/cache_engine.py
ziruiliu: offsets here are lengths of each chunks. for example, if chunk size = 64 and token length = 300, when this line is hit, offsets is [64,64,64,64,44]. sum() equals to total length

ziruiliu:
Hi @ApostaC , some of your comments above are not open to reply. Here is the summarize:
1.  total_length = sum(offsets) : the offsets here are lengths of each chunks
2.  and 3.  min() and assert are redundant, I can remove them later
4.  yes, there are impacts on lookup hit rate, since the lookup hit rate is calculated by
     `self.interval_lookup_hits / self.interval_lookup_tokens` while both include the number of tokens hit in vllm.
      Is it better to minus the num_computed_tokens from both?
Looking forward to you suggestion

ApostaC:
@ziruiliu 
For the hit rate calculation, there are a few cases
1. vLLM has first N tokens cached, and LMCache has first N + M tokens cached.
2. vLLM has first N tokens cached, and LMCache has n tokens (n < N)
3. vLLM has first N tokens cached, and LMCache has [N+1, N + M] tokens cached (first N tokens are evicted from LMCache) -- note: this shouldn't happen in most cases

Here's the hit rate calculation correctness in the above 3 cases
| Case | Old lookup hit tokens | new lookup hit tokens  |
|--------|--------|--------|
| 1. LMCache has N+M | N + M ✅ | N + M ✅ |
| 2. LMCache has n | n ✅ | N ❌ |
| 3. LMCache has [N+1, N+M] | 0 ❌ | M ✅ |

Since case 3 won't happen in most cases, what we have to do is try to make sure that case 2 has the correct numbers.

ziruiliu:
Hi @ApostaC , thanks for clarification. Let's focus on case 2.
In this optimization, if LMCache holds n tokens (n < N, N is in vLLM cache), there will be no lookup() calls to underlying storage backend, to save time spent on lookup() calls to remote storage, especially S3 which takes 10-100ms to check existence.
That means, LMCache will not be able to precisely estimate the number n, because the optimization skips N tokens. So the hit number could be either 0 or N.

My suggestion is to extend LookupRequestStats in observability.py's on_lookup_request() and on_lookup_finished(), to include both lmcache_hit_token and vllm_hit_token, other than a single hit_tokens. Recording more raw metric data instead of calculated or aggregated values improves observability and interpretability. And, can we do that in another PR?

> @ziruiliu For the hit rate calculation, there are a few cases
> 
> 1. vLLM has first N tokens cached, and LMCache has first N + M tokens cached.
> 2. vLLM has first N tokens cached, and LMCache has n tokens (n < N)
> 3. vLLM has first N tokens cached, and LMCache has [N+1, N + M] tokens cached (first N tokens are evicted from LMCache) -- note: this shouldn't happen in most cases
> 
> Here's the hit rate calculation correctness in the above 3 cases
> 
> Case	Old lookup hit tokens	new lookup hit tokens
> 1. LMCache has N+M	N + M ✅	N + M ✅
> 2. LMCache has n	n ✅	N ❌
> 3. LMCache has [N+1, N+M]	0 ❌	M ✅
> Since case 3 won't happen in most cases, what we have to do is try to make sure that case 2 has the correct numbers.

ApostaC:
> My suggestion is to extend LookupRequestStats in observability.py's on_lookup_request() and on_lookup_finished(), to include both lmcache_hit_token and vllm_hit_token, other than a single hit_tokens. Recording more raw metric data instead of calculated or aggregated values improves observability and interpretability. And, can we do that in another PR?

@ziruiliu This make sense. Actually, I think we have improved the metrics calculation during the `retrieve()`. Right now, it will record the number of prompt tokens, the number of hit tokens in vllm, and the number of hit tokens in LMCache during retrieve. So that might be a better place to monitor the "real" cache hit rate.
(cc @maobaolong )

Regarding the lookup hit rate, another proposal is that if we cannot accurately count the number of look up hit tokens, we should deprecate that metric (and remove it in the future). It's fine since we already have a pretty good monitoring during retrieve.

sammshen:
https://github.com/LMCache/LMCache/pull/2007 mainly addresses preeemption but also speeds up the synchronous lookup

ziruiliu:
> > My suggestion is to extend LookupRequestStats in observability.py's on_lookup_request() and on_lookup_finished(), to include both lmcache_hit_token and vllm_hit_token, other than a single hit_tokens. Recording more raw metric data instead of calculated or aggregated values improves observability and interpretability. And, can we do that in another PR?
> 
> @ziruiliu This make sense. Actually, I think we have improved the metrics calculation during the `retrieve()`. Right now, it will record the number of prompt tokens, the number of hit tokens in vllm, and the number of hit tokens in LMCache during retrieve. So that might be a better place to monitor the "real" cache hit rate. (cc @maobaolong )
> 
> Regarding the lookup hit rate, another proposal is that if we cannot accurately count the number of look up hit tokens, we should deprecate that metric (and remove it in the future). It's fine since we already have a pretty good monitoring during retrieve.

Hi @ApostaC  I have updated code according to you suggestions above: 
1. Removed unnecessary min() and assert.
2. Add comment and TODO in lookup hit statistic code. In future we will make decision if deprecating lookup hit rate metrics and use retrieve hit rate instead

ziruiliu:
Hi @YaoJiayi, please take a look and see if your "requested change" can be resolved.
To sum up, after such long discussion, I split this lookup optimization into smaller tasks.
In this one, I focused on passing `num_computed_tokens` down to cache_engine, via all possible interfaces.
However, only sync lookup() handle this `num_computed_tokens` and skip lookup in storage manager.
In sync mode, we do not have to worry about preemption because `self.kv_cache_manager.get_computed_blocks()` and 
`self.connector.get_num_new_matched_tokens()` are always called in pair and will always get most up-to-date number.
We will deal with lookup optimization in async mode later.

lmcache/v1/cache_engine.py
```diff
@@ -766,7 +770,12 @@ def lookup(
             assert hashes is not None
             lookup_request_id = self.stats_monitor.on_lookup_request(sum(offsets))
 
-        res = 0
+        # Skip the number of tokens that are already computed, align to chunk size
+        skip_n_tokens = (
+            num_computed_tokens // self.config.chunk_size * self.config.chunk_size
+        )
+
+        res = skip_n_tokens
```
sammshen, line 778: nit: can we leave a comment here addressing ApostaC's question (which I also had).

smth like: 
```python
even if we don't find anything in lmcache, we wil return the number of tokens already cached in vllm.
the caller must handle this properly
```

lmcache/v1/cache_engine.py
```diff
@@ -778,6 +787,8 @@ def lookup(
             # TODO: support batched_contains when layerwise is enabled
             if self.use_layerwise:
                 for start, end, key in chunk_info_iterator:
+                    if end <= skip_n_tokens:
+                        continue
```
sammshen, line 791: nit: 
add a comment: 
```python
# this continue optimizes lookup for serving engine's with native prefix caching
```

lmcache/v1/cache_engine.py
```diff
@@ -806,6 +817,9 @@ def lookup(
                 keys = []
                 for chunk_info in chunk_info_iterator:
                     assert isinstance(chunk_info[2], CacheEngineKey)
+                    start, end, _ = chunk_info
```
sammshen, line 820: nit: same comment here

lmcache/v1/cache_engine.py
```diff
@@ -827,6 +841,12 @@ def lookup(
             # all tokens where found, return the maximal end
             return res
         finally:
+            # When num_computed_tokens is greater than a chunk, we skip
+            # some tokens to reduce the number of lookup requests.
+            # It is possible that res = skip_n_tokens and no lookup is performed.
+            # In this case, using res as the number of hit tokens will overcount
```
sammshen, line 847: sorry I just saw that you already wrote the comment here

lmcache/v1/cache_engine.py
sammshen: sorry I see you already have a comment below

ziruiliu:
this issue has been revised several times and has too much context.
per discussion with Jiayi and other reviewers, I will make a new PR simply focusing on sync mode

Issue #1909, opened 2025-10-29
AssertionError: Buffer size 5368709120 must be a multiple of align bytes 17989632 in paged memory allocator.


beausoft:
lmcache 0.3.7
vllm 0.11.0

I am deploying deepseek-v3.1 using vllm. I attempted to configure P2P KV Cache Sharing, but I encountered some issues. I have two H200 machines, each with 8 GPUs. 
Environment variable is:

```
NCCL_IB_DISABLE=0
NCCL_IB_CUDA_SUPPORT=1
NCCL_NET_GDR_LEVEL=2
NCCL_IB_GID_INDEX=0
NCCL_IB_HCA=mlx5_0,mlx5_1,mlx5_2,mlx5_3,mlx5_4,mlx5_5,mlx5_6,mlx5_7,mlx5_8
NCCL_DEBUG=INFO
LMCACHE_LOG_LEVEL=DEBUG
VLLM_LOGGING_LEVEL=DEBUG
VLLM_HOST_IP = 215.2.0.129
LMCACHE_CONFIG_FILE=/vllm-workspace/lmcache_config.yaml
PYTHONHASHSEED=123
```

The configuration of lmcache_config.yaml is: 

```
chunk_size: 256
local_cpu: True
max_local_cpu_size: 5
enable_async_loading: True

# P2P configurations
enable_p2p: True
p2p_host: "215.2.0.129"
p2p_init_ports: 8200,8201,8202,8203,8204,8205,8206,8207
p2p_lookup_ports: 8300,8301,8302,8303,8304,8305,8306,8307
transfer_channel: "nixl"

# Controller configurations
enable_controller: True
lmcache_instance_id: "lmcache_instance_1"
controller_pull_url: "215.2.0.129:8300"
controller_reply_url: "215.2.0.129:8400"
lmcache_worker_ports: 8500,8501,8502,8503,8504,8505,8506,8507

extra_config:
  lookup_backoff_time: 0.001
```

Start the service on the first machine. The startup command is: 

```
vllm serve /data/DeepSeek-V3.1 \
        --served-model-name deepseek-v3.1 \
        --trust-remote-code \
        --tensor-parallel-size 8 \
        --port 8000 \
        --host 215.2.0.129 \
        --kv-transfer-config \
        '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
```

Error message:

```
[...]

PR #1924, opened 2025-10-30
[bugfix] memory alignment error for cpu backed memory

DongDongJu:
Motivated by https://github.com/LMCache/LMCache/issues/1909

Summary
  - add PagedTensorMemoryAllocator.required_alignment() and reuse it in _Create_memory_allocator and
    LocalCPUBackend.initialize_allocator so every CPU staging pool is rounded to the allocator’s KV alignment.

  - ensure the pool size is at least one aligned memorry by taking max(size_bytes, align_bytes) before rounding;
    this prevents the default 5 GiB (or any configured size) from falling below the model’s alignment
    requirement.

  Testing
  - check with local storage only.
  - passed the test cases exclude hardware needed.


DongDongJu:
Hello @beausoft, Could you try to reproduce bug with this changes? Thanks!

lmcache/v1/cache_engine.py
```diff
@@ -1364,12 +1367,16 @@ def _Create_memory_allocator(
                 if config.extra_config
                 else max_local_cpu_size
             )
+            size_bytes = int(first_rank_max_local_cpu_size * 1024**3)
+            size_bytes = ((size_bytes + align_bytes - 1) // align_bytes) * align_bytes
```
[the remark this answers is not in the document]
DongDongJu: Do we need to care this? it should normally larger than 1.

lmcache/v1/cache_engine.py
```diff
@@ -1364,12 +1367,16 @@ def _Create_memory_allocator(
...
             return MixedMemoryAllocator(
-                int(first_rank_max_local_cpu_size * 1024**3),
+                size_bytes,
                 numa_mapping=numa_mapping,
             )
+        size_bytes = int(max_local_cpu_size * 1024**3)
+        size_bytes = ((size_bytes + align_bytes - 1) // align_bytes) * align_bytes
```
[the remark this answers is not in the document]
DongDongJu: Same here.

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -331,8 +332,14 @@ def initialize_allocator(
...
 
-        # Calculate effective CPU memory size
+        # Calculate effective CPU memory size (GiB)
         cpu_size = self._calculate_effective_cpu_size(cpu_size, config, metadata)
+        assert metadata is not None
+        align_bytes = PagedTensorMemoryAllocator.required_alignment(
+            metadata, config.chunk_size
+        )
+        size_bytes = int(cpu_size * 1024**3)
+        size_bytes = ((size_bytes + align_bytes - 1) // align_bytes) * align_bytes
```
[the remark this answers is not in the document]
DongDongJu: Same here.

maobaolong:
Is `PagedTensorMemoryAllocator` only work for `PD` and `NixlBackend`? If so, do we need align bytes for other cases?

DongDongJu:
@maobaolong Good point. as I know, all DMA related(GDS,WEKA as well) and disk direct IO need memory alignment. Backend using OS page doesnt need it and most of them probably 4kb alignment is working fine.
but memory alignment is still need it per specific backend and future other backend too.

So, in next step, I think we need a refactor for the storagebackend initialization like categorized with (is_dma or not) which can apply memory alignment based on each backend's requirements.
Then, we can easily sub-categorized for the localCpuBackend's memory allocation problem as well.(I will discuss this in other issue).

Before that, This is good remedy reason why it just init part(overhead is not that high).
And we can reuse the calcuation function later too.
What do you think? @maobaolong @ApostaC @YaoJiayi

lmcache/v1/memory_management.py
```diff
@@ -1018,6 +1018,31 @@ class PagedTensorMemoryAllocator(MemoryAllocatorInterface):
...
+                kv_shape[2] = int(chunk_size)
+                element_size = torch.tensor(
+                    [], dtype=model_meta.kv_dtype
+                ).element_size()
+                alignment = math.prod(kv_shape) * element_size
+            else:
+                alignment = default_alignment
+        else:
+            alignment = default_alignment
+        if alignment <= 0:
```
maobaolong: Nit: As `chunk_size` can not be `<=0` ， so `default_alignment` is > 0 for sure, so this `if alignment <= 0` is redundant.

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -331,8 +332,14 @@ def initialize_allocator(
         numa_mapping = NUMADetector.get_numa_mapping(config)
         logger.info(f"NUMA mapping {numa_mapping}")
 
-        # Calculate effective CPU memory size
+        # Calculate effective CPU memory size (GiB)
         cpu_size = self._calculate_effective_cpu_size(cpu_size, config, metadata)
+        assert metadata is not None
+        align_bytes = PagedTensorMemoryAllocator.required_alignment(
```
maobaolong, line 338: If we need add an alignment logic here, will be more clear if abstract a method.

lmcache/v1/memory_management.py
```diff
@@ -1020,6 +1020,31 @@ class PagedTensorMemoryAllocator(MemoryAllocatorInterface):
...
+            if len(kv_shape) >= 3:
+                kv_shape[2] = int(chunk_size)
+                element_size = torch.tensor(
+                    [], dtype=model_meta.kv_dtype
+                ).element_size()
+                alignment = math.prod(kv_shape) * element_size
+            else:
+                alignment = default_alignment
+        else:
+            alignment = default_alignment
```
ApostaC:
```suggestion
        alignment = default_alignment
        if hasattr(model_meta, "kv_shape") and hasattr(model_meta, "kv_dtype"):
            kv_shape = list(model_meta.kv_shape)
            if len(kv_shape) >= 3:
                kv_shape[2] = int(chunk_size)
                element_size = torch.tensor(
                    [], dtype=model_meta.kv_dtype
                ).element_size()
                alignment = math.prod(kv_shape) * element_size
```

lmcache/v1/memory_management.py
```diff
@@ -1020,6 +1020,31 @@ class PagedTensorMemoryAllocator(MemoryAllocatorInterface):
     Implements a paged memory allocator.
     """
 
+    @staticmethod
+    def required_alignment(model_meta, chunk_size: int) -> int:
```
ApostaC: nit: Please have a type annotation for `model_meta`

lmcache/v1/memory_management.py
```diff
@@ -1020,6 +1020,31 @@ class PagedTensorMemoryAllocator(MemoryAllocatorInterface):
...
+            raise ValueError("chunk_size must be a positive integer")
+        default_alignment = chunk_size * 4096
+        if hasattr(model_meta, "kv_shape") and hasattr(model_meta, "kv_dtype"):
+            kv_shape = list(model_meta.kv_shape)
+            if len(kv_shape) >= 3:
+                kv_shape[2] = int(chunk_size)
+                element_size = torch.tensor(
+                    [], dtype=model_meta.kv_dtype
+                ).element_size()
+                alignment = math.prod(kv_shape) * element_size
```
ApostaC, line 1042: Is this part of the code also compatible with MLA models?

ApostaC (changes requested):
Thanks for the contribution! 
At a high level, I would like to put the size alignment inside the constructor of the corresponding memory allocator. This is because different memory allocators may have different alignment requirements in the future, and the callers (such as `LocalCPUBackend` and `_Create_memory_allocator`) should be agnostic to those implementation details.

lmcache/v1/memory_management.py
DongDongJu: Thanks. applied.

lmcache/v1/storage_backend/local_cpu_backend.py
DongDongJu: I will make the following changes after this by refactoring the initialization per storage backend

lmcache/v1/memory_management.py
DongDongJu: applied. Thanks!

lmcache/v1/memory_management.py
DongDongJu: Applied! thanks.

lmcache/v1/memory_management.py
DongDongJu: I made test for this in here (https://github.com/LMCache/LMCache/pull/1924/commits/c20ecc343990875bc0894cb3a82a2f3303cfa95a). please check it.

DongDongJu:
Hello @beausoft, Could you try to reproduce bug with this changes?

PR #1926, opened 2025-10-30
[Fix] Remove ref_count_up from naive serde serialize()

### LINKED ISSUE #1756: [Bug] The native serdes ref_count_up() is preventing MemoryObj from being unpinned and evicted
**Label**
Please label your issue with "bug" and any other relevant labels so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/627)

**Describe the bug**

The extra memory_obj.ref_count_up() is currently blocking the MemoryObj being released if the cache memory is filled up. We got the following WARNINGs which is not recoverable

```
(Worker_TP3 pid=430) [2025-10-05 20:03:21,359] LMCache WARNING: Local cpu memory under pressure so choosing to not store the KV cache. (cache_engine.py:282:
lmcache.v1.cache_engine)
(Worker_TP2 pid=429) [2025-10-05 20:03:21,360] LMCache DEBUG: Not busy looping because we are not immediately able to evict (local_cpu_backend.py:303:lmcach
e.v1.storage_backend.local_cpu_backend)
```

```
class NaiveSerializer(Serializer):
    def __init__(self):
        pass

    def serialize(self, memory_obj: MemoryObj) -> MemoryObj:
        memory_obj.ref_count_up()
        return memory_obj
```

**To Reproduce**
1. Setup a vLLM with both L1 and L2 Redis remote cache
2. Run a inference workload
3. After running for sometime, you will hit the above warning in model servers

**Expected behavior**
Cache eviction candidates should be found when unpinned and ref_count == 1

**Screenshots**
None

**Additional context**
When using cachegen as serdes, we got no such issue. Thus the issue primarily comes from the Serdes itself.

ningziwen:

This PR will fix the the local CPU memory leaking bug caused by redundant ref_count_up() in naive serde serialization.

Fixes #1756 

 Tested local CPU memory leaking no longer happens with this change.

Analysis for naive serde [serialize()](https://github.com/LMCache/LMCache/blob/b58a9086b96017bc21fe4da3cb88fec466c9042a/lmcache/v1/storage_backend/naive_serde/naive_serde.py#L11) flow.

There are only 3 places calling the method.

[remote_backend.submit_put_task()](https://github.com/LMCache/LMCache/blob/b58a9086b96017bc21fe4da3cb88fec466c9042a/lmcache/v1/storage_backend/remote_backend.py#L233)

[remote_backend.batched_submit_put_task()](https://github.com/LMCache/LMCache/blob/b58a9086b96017bc21fe4da3cb88fec466c9042a/lmcache/v1/storage_backend/remote_backend.py#L271)

Both 1 and 2 are already wrapped by ref_count_up() and ref_count_down() in remote_backend. No additinoal ref_count_up() is needed.

[cache_engine.compress()](https://github.com/LMCache/LMCache/blob/b58a9086b96017bc21fe4da3cb88fec466c9042a/lmcache/v1/cache_engine.py#L910)
The memory_obj here is already pinned so no ref_count change is needed. (Even needed, we should also wrap it similar to remote_backend).

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/naive_serde/naive_serde.py
+++ b/lmcache/v1/storage_backend/naive_serde/naive_serde.py
@@ -9,7 +9,6 @@ class NaiveSerializer(Serializer):
         pass
 
     def serialize(self, memory_obj: MemoryObj) -> MemoryObj:
-        memory_obj.ref_count_up()
         return memory_obj
 
 


maobaolong:
@ningziwen It seems duplicate to #1927 and that PR is more thorough

maobaolong:
@ningziwen Thanks for your contribution!

PR #1927, opened 2025-10-31
[core][bugfix] fix memory leak in batched_put

chunxiaozheng:
The following situation will result in incorrect reference or memory leak:
1. `submit_put_task` of `RemoteBackend`: if use `KIVISerializer` or `CacheGenSerializer`, the `memory_obj` did not call `ref_count_up`, but in `put` of `InstrumentedRemoteConnector`, the `finally` will call `ref_count_down`, which will result in a incorrect reference of `memory_obj`.
2.  `batched_submit_put_task` of `RemoteBackend`: current `NaiveSerializer` will call `ref_count_up` in `serialize`, while `batched_put` of `InstrumentedRemoteConnector` did not call `ref_count_down`, which will result in memory leak.

This PR fix these problems. When calling `submit_put_task` or `batched_submit_put_task`, uniformly applies `ref_count_up` to `memory_obj`, and after `put` or `batched_put` is done, uniformly applies `ref_count_down`.

### DIFF
--- a/lmcache/v1/storage_backend/connector/instrumented_connector.py
+++ b/lmcache/v1/storage_backend/connector/instrumented_connector.py
@@ -114,6 +114,8 @@ class InstrumentedRemoteConnector(RemoteConnector):
     ):
         try:
             await self._connector.batched_put(keys, memory_objs)
+        except Exception as e:
+            logger.warning(f"batched put error: {e}")
         finally:
             for memory_obj in memory_objs:
                 memory_obj.ref_count_down()
--- a/lmcache/v1/storage_backend/connector/instrumented_connector.py
+++ b/lmcache/v1/storage_backend/connector/instrumented_connector.py
@@ -112,7 +112,11 @@ class InstrumentedRemoteConnector(RemoteConnector):
     async def batched_put(
         self, keys: List[CacheEngineKey], memory_objs: List[MemoryObj]
     ):
-        return await self._connector.batched_put(keys, memory_objs)
+        try:
+            await self._connector.batched_put(keys, memory_objs)
+        finally:
+            for memory_obj in memory_objs:
+                memory_obj.ref_count_down()
 
     def remove_sync(self, key: CacheEngineKey) -> bool:
         return self._connector.remove_sync(key)
--- a/lmcache/v1/storage_backend/naive_serde/naive_serde.py
+++ b/lmcache/v1/storage_backend/naive_serde/naive_serde.py
@@ -9,6 +9,7 @@ class NaiveSerializer(Serializer):
         pass
 
     def serialize(self, memory_obj: MemoryObj) -> MemoryObj:
+        memory_obj.ref_count_up()
         return memory_obj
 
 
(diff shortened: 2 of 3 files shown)


maobaolong:
Discussed with @chunxiaozheng offline, this PR is a hotfix for the existing code base without big refactor.  I notice the final approach is doing by @DongDongJu .

maobaolong (approved):
For this PR, it LGTM.

ningziwen:
Fixes #1756

lmcache/v1/storage_backend/connector/instrumented_connector.py
```diff
@@ -112,7 +112,11 @@ async def batched_get(
     async def batched_put(
         self, keys: List[CacheEngineKey], memory_objs: List[MemoryObj]
     ):
-        return await self._connector.batched_put(keys, memory_objs)
+        try:
+            await self._connector.batched_put(keys, memory_objs)
+        finally:
```
DongDongJu, line 119: Can we add one log when it was failed by .debug?
it will helpful for debug it.

DongDongJu:
Hello, glad that this one fixed with more simple approach than what I made before.
I left one comment.

sammshen (approved):
thanks for this fix @chunxiaozheng !

lmcache/v1/storage_backend/connector/instrumented_connector.py
chunxiaozheng: @DongDongJu Thanks for your suggestion, I add a `warning` log to show the error info.

DongDongJu: @chunxiaozheng Thanks!

YaoJiayi (changes requested):
Is this a code refactor or bugfix? 
IIUC, for your first case, in cachegen serializer, the data is moved to the compressed memory object, so no ref count up is needed for the old memory object.
For the second case, there's one count up in naive serializer and one count down outside the naive serializer.

chunxiaozheng:
> Is this a code refactor or bugfix? IIUC, for your first case, in cachegen serializer, the data is moved to the compressed memory object, so no ref count up is needed for the old memory object. For the second case, there's one count up in naive serializer and one count down outside the naive serializer.

@YaoJiayi Thanks for your review! For case 2, this is a bugfix(#1756), if use `RedisConnector` and `NaiveSerializer`, which will result in memory leak. For case 1, after discussion with @maobaolong, I think the current implementation is simpler.

chunxiaozheng:
@YaoJiayi This Pr only fix #1756, PTAL.

YaoJiayi (approved):
LGTM! Thanks!

ningziwen:
@chunxiaozheng @YaoJiayi I thought we would remove ref_count_up() from naive serde but don't see it in final version. Are we addressing the memory leaking of naive serde through other changes?

maobaolong:
@ningziwen Thanks for raise this question, @chunxiaozheng will submit another PR to do a tiny refactor to remove the ref count logic from naive serializer and some place of the serialize caller.

ningziwen:
@maobaolong Thanks, though I see the issue is fixed in my test even with this change only.

maobaolong:
@ningziwen Yeah, indeed, as discussed, the previous content of this PR include two main changes.
- Bugfix 
- Refactor

And decided to do the bugfix in this PR, and submit another PR to do the refactor since bugfix is urgent.

PR #1929, opened 2025-10-31
[bugfix] Fix the lookup socket state error once timeout occurred

maobaolong:
Clipboard_Screenshot_1761955629

Clipboard_Screenshot_1761891526

This PR can solve the above issue which happened after lookup timeout, after that, the socket cannot work anymore.

The reason is "ZMQ REQ/REP" mode, the REQ socket sequence must be `send -> recv -> send -> recv`, once timeout occurred, socket will enter to the error state.

This PR will recreate the socket if error occurred.

### DIFF
--- a/lmcache/v1/lookup_client/lmcache_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_lookup_client.py
@@ -91,8 +91,9 @@ class LMCacheLookupClient(LookupClientInterface):
 
         for params in self.socket_params:
             logger.info(
-                f"lmcache lookup client connect to rank {params.rank} "
-                f"with socket path {params.socket_path}"
+                "lmcache lookup client connect to rank %s with socket path %s",
+                params.rank,
+                params.socket_path,
             )
             socket = get_zmq_socket(
                 self.ctx,
@@ -132,7 +133,9 @@ class LMCacheLookupClient(LookupClientInterface):
                     old_socket.close(linger=0)
                 except zmq.ZMQError as e:
                     logger.warning(
-                        f"ZMQ error closing old socket for rank {rank_idx}: {e}"
+                        "ZMQ error closing old socket for rank %s: %s",
+                        rank_idx,
+                        e,
                     )
                 except AttributeError:
                     # Socket already closed or invalid
@@ -141,8 +144,9 @@ class LMCacheLookupClient(LookupClientInterface):
             # Create new socket using stored parameters
             params = self.socket_params[rank_idx]
             logger.info(
-                f"Recreating socket for rank {params.rank} "
-                f"with path {params.socket_path}"
+                "Recreating socket for rank %s with path %s",
+                params.rank,
+                params.socket_path,
             )
 
             new_socket = get_zmq_socket(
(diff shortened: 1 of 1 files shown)


maobaolong:
@chunxiaozheng @YaoJiayi Would you like to take a look at this PR? Thanks!

ApostaC:
Thanks @maobaolong! 

2 high-level questions:
1. IIUC, the error `operation cannot be accomplished in the current state` is caused by calling 2 `send`s on the same `req` socket without calling `recv`. However, I'm not sure whether this is caused by the timeout. Therefore, could you maybe share a full log so that we can better understand the problem?
2. In the implementation, I saw you introduced a `socket_lock`. Since the lookup client it's single-threaded, why we need to have a lock here?

maobaolong:
@ApostaC I paste  another screenshot at the head of the description of this PR.

maobaolong:
@ApostaC The socket lock has been removed, thanks for your suggestion.

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -157,20 +200,30 @@ def lookup(
...
                 resp = self.sockets[i].recv()
                 result = int.from_bytes(resp, "big")
                 results.append(result)
-        except zmq.Again:
-            logger.error(f"Timeout occurred for rank {i}")
+        except zmq.Again as e:
+            logger.error(
+                f"Timeout occurred for rank {failed_rank}, "
+                f"recreating all sockets. Error: {str(e)}"
+            )
```
ApostaC, line 224: nit: it's better not to use f-string when using loggers.

```suggestion
            logger.error(
                "Timeout occurred for rank %d, "
                "recreating all sockets. Error: %s",
                failed_rank, e
            )
```

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -157,20 +200,30 @@ def lookup(
...
+            self._recreate_socket()
             return 0
         except zmq.ZMQError as e:
-            logger.error(f"ZMQ error for rank {i}: {str(e)}")
+            logger.error(
+                f"ZMQ error for rank {failed_rank}: {str(e)}, recreating all sockets"
+            )
```
ApostaC, line 232: nit: we can do similar updates here:
```suggestion
            logger.error(
                f"ZMQ error for rank %d: %s, recreating all sockets",
                failed_rank, e
            )
```

ApostaC (approved):
Nit & perf-related comment: It's better not to use f-string when using loggers. See: https://docs.python.org/3/howto/logging.html#optimization
(Left some example reviews below)

Otherwise LGTM!

chunxiaozheng:
@maobaolong the comprehensive test is failed, otherwise lgtm.

ApostaC:
Seems like we have got a little bit of performance degradation, which causes comprehensive tests to fail.
Not sure whether it's caused by this PR or not. 

@sammshen @Shaoting-Feng can you investigate it a bit when you got a chance?

Shaoting-Feng:
> Seems like we have got a little bit of performance degradation, which causes comprehensive tests to fail. Not sure whether it's caused by this PR or not.
> 
> @sammshen @Shaoting-Feng can you investigate it a bit when you got a chance?

There is actually no performance degradation. I reran four commits from the main branch up to the latest (including the one right after I added the layerwise test), and all of them failed to achieve the gain set in the layerwise test. I believe the minor runtime variations on the same idle server are due to nondeterministic factors such as GPU frequency scaling, PCIe transfer timing, and CPU scheduling jitter, etc. I’ll loosen the required gain threshold in a new PR.

Screenshot 2025-11-03 at 2 55 16 PM

maobaolong:
@Shaoting-Feng Thanks for the clarification about the comprehensive test. 

So, as conclusion, this PR did not get performance degradation. 

@chunxiaozheng WDYT?

Shaoting-Feng (approved):

chunxiaozheng (approved):
lgtm

maobaolong:
@ApostaC @chunxiaozheng @Shaoting-Feng Thanks for your review and suggestion. merge it.

PR #1937, opened 2025-11-01
[Core] Add SageMaker HyperPod remote connector

ningziwen:
Add SageMaker Hyperpod remote connector to support SageMaker Hyperpod ai-toolkit daemon (https://docs.aws.amazon.com/sagemaker/latest/dg/managed-tier-checkpointing.html) as L2 cache backend.

**If applicable**:

### DIFF
--- a/docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
+++ b/docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
@@ -28,9 +28,9 @@ SageMaker Hyperpod-Specific (in extra_config)
 
 * **sagemaker_hyperpod_bucket**: Bucket name for KV storage namespace (default: "lmcache")
 * **sagemaker_hyperpod_shared_memory_name**: Name of shared memory segment (default: "shared_memory"). Set to None to disable shared memory.
-* **sagemaker_hyperpod_max_concurrent_requests**: Maximum concurrent control plane requests (default: 100, minimum: 1)
-* **sagemaker_hyperpod_max_connections**: Maximum total HTTP connections in pool (default: 256, minimum: 1)
-* **sagemaker_hyperpod_max_connections_per_host**: Maximum HTTP connections per host (default: 128, minimum: 1)
+* **sagemaker_hyperpod_max_concurrent_requests**: Maximum concurrent HTTP requests allowed in-flight at any moment (application-level throttling, default: 100, minimum: 1). This limit is per LMCache engine instance. With multiple workers (e.g., high TP), each worker creates its own engine with separate limits.
+* **sagemaker_hyperpod_max_connections**: Maximum total TCP connections in the connection pool per LMCache engine across all daemons (default: 256, minimum: 1). For typical single-daemon setups, this effectively limits connections from one engine to one daemon. With N workers per node, total connections to the daemon = N × this value.
+* **sagemaker_hyperpod_max_connections_per_host**: Maximum TCP connections per LMCache engine to a single daemon address (IP:port) (default: 128, minimum: 1). "Host" refers to the daemon's network address, not the client machine. For today's typical single-daemon setup, this has similar effect as max_connections. This parameter enables future multi-daemon configurations where one engine connects to multiple daemons for load balancing. With N workers per node connecting to the same daemon, total connections = N × this value. Reduce proportionally for high TP setups (e.g., set to 16 for 8 workers to achieve ~128 total connections).
 * **sagemaker_hyperpod_timeout_ms**: Timeout for lease acquisition requests in milliseconds (default: 5000, minimum: 100)
 * **sagemaker_hyperpod_lease_ttl_s**: Server-side lease timeout in seconds (default: 30.0)
 * **sagemaker_hyperpod_put_stream_chunk_bytes**: Chunk size for streaming PUT requests in bytes (default: 65536, minimum: 1024)
--- a/docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
+++ b/docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
@@ -1,0 +1,71 @@
+SageMaker Hyperpod Backend
+===========================
+
+Prerequisites
+-------------
+
+Create an Amazon SageMaker HyperPod cluster with tiered storage enabled by following the instructions at:
+
+https://docs.aws.amazon.com/sagemaker/latest/dg/managed-tier-checkpointing-setup.html
+
+This enables the ai-toolkit daemon that provides shared memory access for LMCache.
+
+Example Configuration
+---------------------
+
+.. code-block:: yaml
+
+   chunk_size: 256
+   local_cpu: True
+   max_local_cpu_size: 5
+   remote_url: "sagemaker-hyperpod://$NODE_IP:9200"
+
+Configuration Parameters
+------------------------
+
+SageMaker Hyperpod-Specific (in extra_config)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* **sagemaker_hyperpod_bucket**: Bucket name for KV storage namespace (default: "lmcache")
+* **sagemaker_hyperpod_shared_memory_name**: Name of shared memory segment (default: "shared_memory"). Set to None to disable shared memory.
+* **sagemaker_hyperpod_max_concurrent_requests**: Maximum concurrent control plane requests (default: 100, minimum: 1)
+* **sagemaker_hyperpod_max_connections**: Maximum total HTTP connections in pool (default: 256, minimum: 1)
+* **sagemaker_hyperpod_max_connections_per_host**: Maximum HTTP connections per host (default: 128, minimum: 1)
+* **sagemaker_hyperpod_timeout_ms**: Timeout for lease acquisition requests in milliseconds (default: 5000, minimum: 100)
+* **sagemaker_hyperpod_lease_ttl_s**: Server-side lease timeout in seconds (default: 30.0)
+* **sagemaker_hyperpod_put_stream_chunk_bytes**: Chunk size for streaming PUT requests in bytes (default: 65536, minimum: 1024)
+* **sagemaker_hyperpod_use_https**: Enable HTTPS instead of HTTP (default: False). **Note**: Ignored if ``remote_url`` already contains ``http://`` or ``https://`` protocol.
+* **save_chunk_meta**: Whether to save chunk metadata with data (set False for performance)
+
+Kubernetes Deployment Requirements
+-----------------------------------
+
+Environment Variable for Node IP
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add the ``NODE_IP`` environment variable to resolve the local node's IP address:
+
+.. code-block:: yaml
+
+   env:
+     - name: NODE_IP
+       valueFrom:
+         fieldRef:
+           fieldPath: status.hostIP
+
+/dev/shm Volume Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SageMaker Hyperpod requires /dev/shm for high-performance shared memory operations:
+
+.. code-block:: yaml
+
+   volumeMounts:
+     - name: dshm
+       mountPath: /dev/shm/shared_memory
+       subPath: shared_memory
+
+   volumes:
+     - name: dshm
+       hostPath:
+         path: /dev/shm
--- a/lmcache/v1/storage_backend/connector/sagemaker_hyperpod_adapter.py
+++ b/lmcache/v1/storage_backend/connector/sagemaker_hyperpod_adapter.py
@@ -56,6 +56,7 @@ class SageMakerHyperPodConnectorAdapter(ConnectorAdapter):
             "lease_ttl_s": 30.0,
             "put_stream_chunk_bytes": 65536,
             "use_https": False,
+            "max_lease_size_mb": None,
         }
 
         # Extract and validate configuration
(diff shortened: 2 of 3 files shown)


lmcache/v1/storage_backend/connector/sagemaker_hyperpod_adapter.py
```diff
@@ -0,0 +1,218 @@
...
+        lease_ttl_s = self._get_positive_float(
+            extra_config,
+            "sagemaker_hyperpod_lease_ttl_s",
+            float(defaults["lease_ttl_s"]),  # Cast to float
+        )
+        put_stream_chunk_bytes = self._get_positive_int(
+            extra_config,
+            "sagemaker_hyperpod_put_stream_chunk_bytes",
+            int(defaults["put_stream_chunk_bytes"]),  # Cast to int
+        )
```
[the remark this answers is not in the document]
ningziwen: I run into formatting check error if removing it
```
lmcache/v1/storage_backend/connector/sagemaker_hyperpod_adapter.py:74: error: Argument 3 to "_get_positive_int" of "SageMakerHyperPodConnectorAdapter" has incompatible type "object"; expected "int"  [arg-type]
```

ApostaC:
Thanks @ningziwen . Will take a look today!

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_adapter.py
```diff
@@ -0,0 +1,218 @@
...
+        # Strip the sagemaker_hyperpod:// prefix
+        raw_url = url.removeprefix("sagemaker_hyperpod://")
+
+        assert raw_url, (
+            "SageMaker HyperPod URL must contain host information after 'sagemaker_hyperpod://'"
+        )
+
+        # If URL already has protocol, use it as-is
+        if raw_url.startswith("http://") or raw_url.startswith("https://"):
+            return raw_url
```
ApostaC, line 188: What if raw_url contains `http://` but the flag `use_https = True`?

ApostaC (commented):
In general looks good to me.

Can you also add a doc page to tell people how to use the new remote connector?

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_adapter.py
ningziwen: It would respect raw_url more. Let me know if you have better suggestions here.

ningziwen:
@ApostaC Added doc

ningziwen:
@ApostaC @YaoJiayi @maobaolong Could you please include this in next release (0.3.10)? Thanks.

ApostaC (approved):
LGTM in general. @YaoJiayi / @sammshen Can you guys take a second look?

In the meantime, @ningziwen is there a way to establish a unit test pipeline in our CI to verify the functionality of this SageMaker connector?

ningziwen:
@ApostaC Yeah I'm figuring out unit tests right now. I would add them as separated PR coming soon if it is ok since most of connectors today also don't have unit tests.

docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
```diff
@@ -0,0 +1,71 @@
...
+------------------------
+
+SageMaker Hyperpod-Specific (in extra_config)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* **sagemaker_hyperpod_bucket**: Bucket name for KV storage namespace (default: "lmcache")
+* **sagemaker_hyperpod_shared_memory_name**: Name of shared memory segment (default: "shared_memory"). Set to None to disable shared memory.
+* **sagemaker_hyperpod_max_concurrent_requests**: Maximum concurrent control plane requests (default: 100, minimum: 1)
+* **sagemaker_hyperpod_max_connections**: Maximum total HTTP connections in pool (default: 256, minimum: 1)
+* **sagemaker_hyperpod_max_connections_per_host**: Maximum HTTP connections per host (default: 128, minimum: 1)
```
sammshen: for high TP engine on a single host, this will create multiple lmcache engines (and hyperpod connectors). if this limit is per host, is this dangerous?

docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Configuration Parameters
+------------------------
+
+SageMaker Hyperpod-Specific (in extra_config)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* **sagemaker_hyperpod_bucket**: Bucket name for KV storage namespace (default: "lmcache")
+* **sagemaker_hyperpod_shared_memory_name**: Name of shared memory segment (default: "shared_memory"). Set to None to disable shared memory.
+* **sagemaker_hyperpod_max_concurrent_requests**: Maximum concurrent control plane requests (default: 100, minimum: 1)
```
sammshen: can you explain a bit more what's the difference between: 
- max_concurrent_requests
- max_connections
- max_connectiosn_per_host

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
```diff
@@ -0,0 +1,938 @@
...
+            logger.debug(f"Lease has no offsets: key={key_str}")
+            return None
+
+        lease_info = LeaseInfo(
+            lease_id=lease_data["id"],
+            offsets=offsets,
+        )
+
+        self.stats["lease_acquired"] += 1
+        total_size = sum(length for _, length in offsets)
```
sammshen, line 387: how are we controlling how large of a lease we are requesting here?

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
```diff
@@ -0,0 +1,938 @@
...
+
+    def post_init(self):
+        """Initialize shared memory connection after construction."""
+        if self.shared_memory_name:
+            self._init_shared_memory()
+
+    def _init_shared_memory(self):
+        """Initialize shared memory connection to ai-toolkit daemon."""
+        try:
+            self.shared_memory_obj = shared_memory.SharedMemory(
```
sammshen, line 163: what's the default size that this file in `/dev/shm` will be?

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
```diff
@@ -0,0 +1,938 @@
...
+                name=self.shared_memory_name, create=False
```
sammshen, line 164: just to clarify, it is intentional that all the parallel workers share the same file in `/dev/shm`?

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
```diff
@@ -0,0 +1,938 @@
...
+            self._acquire_lease,
+            key=key,
+            priority=Priorities.LEASE,
+        )
+
+    def _read_from_shared_memory(
+        self, key: CacheEngineKey, lease_info: LeaseInfo
+    ) -> Optional[MemoryObj]:
+        """
+        Read data from shared memory using lease offsets (zero-copy).
```
sammshen: this is not true zero copy right? we are going: 

hyperpod -> dev/shm -> pinned CPU RAM -> GPU

The S3 connector has the same issue though we should be able to call cudaHostRegister on `/dev/shm` so a true zero copy will be possible. Any thoughts on this @ningziwen ?

docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
ningziwen: The "host" refers to the daemon's address (IP:port). Today all LMCache engines on a node connect to the same local daemon (e.g., 127.0.0.1:9200).

Both max_connections and max_connections_per_host are per engine. So with 8 TP workers (8 engines), each engine has its own connection pool with these limits. If you want to control total connections to the daemon, you can reduce either max_connections or max_connections_per_host accordingly.

For today's typical single-daemon setup, these two parameters are pretty much the same. But sagemaker_hyperpod_max_connections_per_host is for future extensibility when users have multiple daemons and can properly shard traffic across them.

For example, if an engine connects to both daemon-1:9200 and daemon-2:9200 with max_connections=200 and max_connections_per_host=128, you get up to 128 connections to each daemon (256 total, but capped at 200 globally).

docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
ningziwen: I will update same content in doc. Let me know if you have further questions.

* **sagemaker_hyperpod_max_concurrent_requests**: Maximum concurrent HTTP requests allowed in-flight at any moment (application-level throttling, default: 100, minimum: 1). This limit is per LMCache engine instance. With multiple workers (e.g., high TP), each worker creates its own engine with separate limits.
* **sagemaker_hyperpod_max_connections**: Maximum total TCP connections in the connection pool per LMCache engine across all daemons (default: 256, minimum: 1). For typical single-daemon setups, this effectively limits connections from one engine to one daemon. With N workers per node, total connections to the daemon = N × this value.
* **sagemaker_hyperpod_max_connections_per_host**: Maximum TCP connections per LMCache engine to a single daemon address (IP:port) (default: 128, minimum: 1). "Host" refers to the daemon's network address, not the client machine. For today's typical single-daemon setup, this has similar effect as max_connections. This parameter enables future multi-daemon configurations where one engine connects to multiple daemons for load balancing. With N workers per node connecting to the same daemon, total connections = N × this value. Reduce proportionally for high TP setups (e.g., set to 16 for 8 workers to achieve ~128 total connections).

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ningziwen: 1GiB. This size is controlled by HyperPod/ai-toolkit. Connector does not have the control.

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ningziwen: Yes. /dev/shm/shared_memory is per hardware node.

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ningziwen: Good catch. I think cudaHostRegister would be a good direction to explore to optimize the performance. That would need larger change and validation. I'm removing the zero-copy statement here for now and will follow up later.

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ningziwen: Adding max_lease_size_mb field but not set it by default since I believe the lease size won't be a concern for major cases. If users want to restrict they can. I'm open to set to some number as well if you prefer.

ningziwen:
@sammshen Replied/Addressed all the comments

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ApostaC: I'm a bit confused here. If people run LMCahce with SageMaker connector in a pod, then the `/dev/shm` should also be configured by that user pod, right? Under what case it will be controlled by something else?

ApostaC:
@ningziwen It would be great if you could also share some performance numbers here. Also, we will be happy to collaborate on writing a blog.

ningziwen:
> @ningziwen It would be great if you could also share some performance numbers here. Also, we will be happy to collaborate on writing a blog.

@ApostaC Thanks. This PR will enable the initial integration. I will come back with unit tests, blog/performance number soon after it.

docs/source/kv_cache/storage_backends/sagemaker_hyperpod.rst
sammshen: thanks for updating the doc

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
sammshen: @ningziwen what configuration needs to be done before deployment. is adding this to the manifest enough?

```yaml
   volumeMounts:
     - name: dshm
       mountPath: /dev/shm/shared_memory
       subPath: shared_memory
   volumes:
     - name: dshm
       hostPath:
         path: /dev/shm
 ```

sammshen (approved):
no urgent blocking changes on my side for this initial integration. looking forward to any working results and further optimization. we can do parallel development on the S3 connector.

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ningziwen: @ApostaC @sammshen `/dev/shm/shared_memory` is created by Hyperpod's ai-toolkit daemonset. In model server pod, the existing `/dev/shm/shared_memory` just needs to be mounted to model server pod yaml as mentioned above.

sammshen: ok, let's add this nit to the doc in a follow up PR

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
ningziwen: Moving discussion to https://github.com/LMCache/LMCache/issues/2305.

PR #1557, opened 2025-09-10
[Core] Support NIXL storage obj backend

tshmilnvidia:
Support NIXL obj backend
    
    Adding support for NIXL obj backend.
    
    in order to use this backend, AWS parameters should be provided via
    extra_config.nixl_backend_params in the yaml file, for example:
    
    extra_config:
      enable_nixl_storage: true
      nixl_backend: OBJ
      nixl_pool_size: 64
      nixl_path: /mnt/nixl/cache/
      nixl_backend_params:
        access_key: <your_access_key>
        secret_key: <your_secret_key>
        bucket: <your_bucket>
        region: <your_region>

PR #1939, opened 2025-11-02
[Core] A nixl backend dedicated for object storage

ziruiliu:
## Why
- Comparing to existing s3 connector based on awscrt library, this nixl_object_backend  enabled self-signed certificates as ca_bundle option for private S3-compatible endpoints—that are difficult to supported with awscrt library.
- Nixl object backend leverages nixl transfer API to transfer data in a batch as nixl_storage_backend also did, but instead of the in-memory dictionary used by the original Nixl storage backend, each cache line is addressed by a deterministic key derived from the `CacheEngineKey`, letting us reuse objects after a reboot without rebuilding metadata.  Therefore the new nixl_object_backend is able to share KV cache data between multiple nodes, and survives process restart

see https://github.com/LMCache/LMCache/pull/1557, original Nixl storage backend needs nixl_pool_size to pre-allocate certain number of files or object names at initialization with random names. With the new backend, there is no nixl_pool_size value because objects are named with key's hash value. Exatr config looks like:
```yaml
extra_config:
  enable_nixl_object: true
  nixl_object_backend: OBJ
  nixl_object_presence_cache: False
  nixl_object_async_put: False
  nixl_object_backend_params:
    access_key: ACCESS_KEY
    secret_key: SECRET
    bucket: nixl-test-bucket
    region: any
    endpoint_override: https://url-to-object-storage
    ca_bundle: path to self-signed certificate
```

## What
- Implement a `NixlObjectBackend` built on top of `NixlAgent` helpers for memory registration and transfer so all reads/writes go through Nixl-provided batched descriptors. 
- Implement contains() via Nixl's query_mem() API (equivalent to HeadObject) and implement `batched_async_contains` by callin contains() parallely in multiple batches, fail fast if there is any failure in the small batchs, instead of issuing all contains() at once
- Introduce an optional presence cache to short-circuit remote existence checks and avoid repeated `query_memory` calls during lookups. The cache could be false-positive, which would trigger failure in following retrieval. However failure in retrieval could be recovered by async KV loading in LMCache or vllm's [#19330](https://github.com/vllm-project/vllm/pull/19330)
- Introduce an option that enable async put in write path to leverage Nixl's natively asynchronous API


mmgaggle:
I love that you're doing this. We will give it a spin with our all-flash Ceph cluster.

The CRT client does have upsides, namely distributing connections in the client connection pool across s3 endpoints returned via DNS. We played with a variation of nixl/obj connector that was modified to use S3CrtClient (to get multipathing), but there were enough other issues that we abandoned it in favor of the native s3 connector in lmcache. Making the nixl object content addressable is a huge step forward, and I also appreciate the lookup caching!

mmgaggle:
We are seeing significant improvement using this branch:

Screenshot 2025-11-12 at 12 07 30 PM

ziruiliu:
> I love that you're doing this. We will give it a spin with our all-flash Ceph cluster.
> 
> The CRT client does have upsides, namely distributing connections in the client connection pool across s3 endpoints returned via DNS. We played with a variation of nixl/obj connector that was modified to use S3CrtClient (to get multipathing), but there were enough other issues that we abandoned it in favor of the native s3 connector in lmcache. Making the nixl object content addressable is a huge step forward, and I also appreciate the lookup caching!

Thank you Kyle for testing with this branch. I am now working with Nvidia nixl team to get an unified version of nixl storage backend. Please stay tuned

leelavg:
if it helps in anyway, we also tested this branch with kserve/llminferenceservice observing good perf difference and also a streamlined config when compared to lmcache+s3 (shm keeps on filling up) and lmcache+nixl[storage]+s3 (inconsistent numbers intermittently).

This is a small test I've added to test the basic flow https://github.com/leelavg/LMCache/blob/nixl-test/tests/test_nixl_object.py

Thanks.

ziruiliu:
Close this one. This scalable nixl object backend is integrated into an unified nixl storage backend, see PR #2024 
Setting nixl_pool_size to 0 in nixl storage backend will enable dynamic mode, which is the same as this one

PR #1944, opened 2025-11-03
[feat] Support get the env of current process

maobaolong:
Support get env from lmcache.


maobaolong:

KuntaiDu (approved):
LGTM for the first version, but we may probably need some way to safeguard this in the future because envvars contain a lot of confidential information (e.g. `HF_TOKEN`)

maobaolong:
@KuntaiDu Thanks for your suggestion, we can add a config option which include some hidden regex list to specify which env vars should not be displayed or use star replace the original value, we can do this in the future.

chunxiaozheng (approved):
lgtm

PR #1951, opened 2025-11-05
[CI]: audit connector unit test patch

sammshen:
This is caused by: 

https://github.com/LMCache/LMCache/pull/1859

which merged without conflicts after more unit tests were added


YaoJiayi (approved):
lgtm

PR #1854, opened 2025-10-22
[bugfix] fix batched_contains bug when lookup server on other rank

chunxiaozheng:
Fix `batched_contains` bug in `RemoteBackend`.

PR #1955, opened 2025-11-06
[refactor] unified reconstruct cache engine key

chunxiaozheng:
unified reconstruct cache engine key, there are tow main benfits:
1. The code is clearer;
2. Prevent missing modifications, which will result in some bugs. such as the following case:

#1854 add the following code to reconstruct the `CacheEngineKey`
```
        if self._mla_worker_id_as0_mode:
            keys = [
                CacheEngineKey(
                    key.fmt,
                    key.model_name,
                    key.world_size,
                    0,
                    key.chunk_hash,
                    key.request_configs,
                )
                for key in keys
            ]
```
#1859 add `dtype` in `CacheEngineKey`, but misses the modification here.

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -298,6 +298,18 @@ class CacheEngineKey:
             request_configs=request_configs,
         )
 
+    def with_new_worker_id(self, new_worker_id: int) -> "CacheEngineKey":
+        # Reconstruct the cache engine key with new worker id
+        return CacheEngineKey(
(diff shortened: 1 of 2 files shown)


maobaolong (approved):
Thanks for this refactor, it looks better and this can reduce changes to forget modification while structure of CEK changing.

KuntaiDu (approved):
Looks cleaner and more intuitive. Thanks for contributing!

PR #1958, opened 2025-11-06
refactor: Abstract internal_api_server & plugin_launcher to LMCacheEngine

maobaolong:
This PR aim to abstract some content apart from vllm_v1_adapter into lmcache_engine module to make vllm adapter part as tiny as possible.

### DIFF
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -19,6 +19,9 @@ logger = init_logger(__name__)
 class LMCacheEngineMetadata:
     """name of the LLM model"""
 
+    ROLE_SCHEDULER = "scheduler"
+    ROLE_WORKER = "worker"
+
     model_name: str
     """ world size when running under a distributed setting """
     world_size: int
@@ -35,9 +38,10 @@ class LMCacheEngineMetadata:
     use_mla: bool = False
     """ the role of the current instance (e.g., 'scheduler', 'worker') """
     role: Optional[str] = None
+    """ the local dp rank """
+    dp_rank_local: int = 0
     """ the first rank of the distributed setting """
-    # TODO(baoloongmao): first_rank should be configurable
-    first_rank = 0
+    first_rank: int = 0
 
     def is_first_rank(self) -> bool:
         """Check if the current worker is the first rank"""
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
 from dataclasses import dataclass
+from enum import Enum
 from typing import Any, Optional, Tuple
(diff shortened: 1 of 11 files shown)


maobaolong:
@sammshen Would you please help to take a look at this refactor? Thanks!

maobaolong:
@ApostaC Would you like to take a look also?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -715,32 +578,143 @@ def __init__(
...
-            self.api_server = None  # type: ignore[assignment]
-            self.plugin_launcher = None  # type: ignore[assignment]
         logger.info(
             f"LMCache initialized for role {role} with version {utils.get_version()}, "
             f"vllm version {VLLM_VERSION}, "
             "lmcache cache_engine metadata: "
             f"{getattr(self.lmcache_engine, 'metadata', None)}"
         )
 
+    def _init_lmcache_engine(
```
maobaolong, line 589: This method is just moved from outside to here, the content is not changed. @sammshen @ApostaC

ApostaC:
I think @sammshen is also working on refactoring the vllm_v1_adapter. We should probably talk to avoid conflicts.

maobaolong:
@ApostaC Thanks for remind me, will discuss with @sammshen .

sammshen:
This change will not be compatible with the multi-process design right? 

Is it correct that we want every scheduler/connector to be able to have a plugin launcher / internal api server?

lmcache/config.py
```diff
@@ -19,6 +19,9 @@
 class LMCacheEngineMetadata:
     """name of the LLM model"""
 
+    ROLE_SCHEDULER = "scheduler"
```
sammshen: this is kind of a weird way to introduce an enumeration?

maobaolong: Thanks for point it out! Changed to use enum

maobaolong:
@sammshen Thanks for your review and suggested changes are addressed.

> This change will not be compatible with the multi-process design right?

I'm not sure about that, maybe @ApostaC can take a double check?

> Is it correct that we want every scheduler/connector to be able to have a plugin launcher / internal api server?

Yeah, the design of `PluginFramework` is working for both `scheduler` and `worker` role, and the script of plugin can distinguish the role of current process. see to https://docs.lmcache.ai/developer_guide/extending_lmcache/plugin.html

For Internal api server, it is yes too, it is a way we interact with each lmcache process by internal api server.

sammshen (approved):
LGTM, we can revisit the design in the multiprocess use case later. In that case, the adapter will not have an lmcache engine,=.

PR #1959, opened 2025-11-06
[Core][vLLM]: Add KV event generation

### LINKED ISSUE #1846: [RFC] Add KV events for vLLM

hickeyma:

This commit adds generation of kv events. It creates the events during LMCache storage operations and stores them in the cache engine. The vLLM model runner and scheduler processes then handle the events and the events are finally published by the vLLM scheduler process.

The overall data flow is as follows:
- Model runner process calls connector `get_kv_connector_kv_events` method which calls cache_engine to return the events
- Model runner process sends the information to the scheduler process via `KVConnectorOutput` class
- Scheduler process calls connector `update_connector_output` method and the connector gets access to events in scheduler side now
- Scheduler calls connector `take_events` method, and the connector returns the latest events to be published by the scheduler

This PR is based on the vLLM RFC [[RFC]: KV-Cache Interoperability API Standardization ](https://github.com/vllm-project/vllm/issues/20492) which proposes "formalizing KV Events as the public contract for any component emitting or consuming KV-Cache lifecycle events - including external indexers, routers, and engines". 

Fixes #1846

- The configuration is defined in  `lmcache/v1/config.py`. Please note that KV events are disabled by default. Need to set `enable_kv_events` to `True` for events to be generated.
- Example of the command for vLLM to publish events is as follows: `LMCACHE_CONFIG_FILE=lmcache_config.yaml vllm serve meta-llama/Llama-3.1-8B-Instruct -kv-transfer-config '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}' --disable-log-requests --no-enable-prefix-caching --kv-events-config '{"enable_kv_cache_events": "True", "publisher": "zmq", "topic": "kv-events"}'`. **Note:** You need to build vLLM PR https://github.com/vllm-project/vllm/pull/28309 and run that version of vLLM to pickup the changes so that the events work end to end. A good command to build vLLM Python only (and it only takes seconds) is as follows: `VLLM_USE_PRECOMPILED=1 uv pip install --editable . --extra-index-url https://download.pytorch.org/whl/cu129 --index-strategy unsafe-best-match --prerelease=allow`. 
- Consumer code to test the events being published can be referenced in: https://docs.vllm.ai/en/stable/examples/online_serving/kv_events_subscriber.html
- Currently, events only being generated for storage and not eviction

**If applicable**:

**Additional comment:**

A big thank you to @ApostaC for pointing me towards using the vLLM runner and scheduler flow to handle the event publishing and therefore simplify the solution as compared to #1843

### DIFF
--- a/lmcache/integration/vllm/lmcache_connector_v1.py
+++ b/lmcache/integration/vllm/lmcache_connector_v1.py
@@ -1,12 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
+
 # Standard
-from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, List, Optional
-import time
+from typing import TYPE_CHECKING, Any, Optional
 
 # Third Party
 from vllm.config import VllmConfig
-from vllm.distributed.kv_events import BlockStored, KVCacheEvent, KVEventBatch
 from vllm.distributed.kv_transfer.kv_connector.v1.base import (
     KVConnectorBase_V1,
     KVConnectorMetadata,
@@ -14,7 +12,6 @@ from vllm.distributed.kv_transfer.kv_connector.v1.base import (
 )
 from vllm.logger import init_logger
 from vllm.v1.core.sched.output import SchedulerOutput
-from vllm.v1.outputs import KVConnectorOutput
 import torch
 
 # First Party
@@ -27,25 +24,18 @@ if TYPE_CHECKING:
     from vllm.v1.core.kv_cache_manager import KVCacheBlocks
     from vllm.v1.request import Request
 
-
 logger = init_logger(__name__)
 
 
-class LMCacheConnectorV1(KVConnectorBase_V1):
-    def __init__(
-        self,
-        vllm_config: "VllmConfig",
-        role: KVConnectorRole,
-    ):
+class LMCacheConnectorV1Dynamic(KVConnectorBase_V1):
+    def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
         super().__init__(vllm_config=vllm_config, role=role)
         self._lmcache_engine = LMCacheConnectorV1Impl(vllm_config, role, self)
-
-        self._kv_events: List[KVCacheEvent] = []
 
     # ==============================
     # Worker-side methods
     # ==============================
-    def start_load_kv(self, forward_context: "ForwardContext", **kwargs: Any) -> None:
+    def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
         """
         Start loading the KV cache from the connector to vLLM's paged
         KV buffer. This is called from the forward context before the
@@ -80,7 +70,7 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
         layer_name: str,
         kv_layer: torch.Tensor,
         attn_metadata: "AttentionMetadata",
-        **kwargs: Any,
+        **kwargs,
     ) -> None:
         """
         Start saving the a layer of KV cache from vLLM's paged buffer
@@ -110,7 +100,7 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
 
     def get_finished(
         self, finished_req_ids: set[str]
-    ) -> tuple[set[str] | None, set[str] | None]:
+    ) -> tuple[Optional[set[str]], Optional[set[str]]]:
         """
         Notifies worker-side connector ids of requests that have
         finished generating tokens.
@@ -124,29 +114,9 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
         """
         return self._lmcache_engine.get_finished(finished_req_ids)
 
-    def get_kv_connector_kv_cache_events(self) -> Optional["KVEventBatch"]:
-        """
-        Get the KV connector kv cache events collected during the last interval.
-        """
-        events = self._lmcache_engine.get_kv_events()
-        if not events:
-            return None
-
-        lmcache_kv_events: KVEventBatch | None = None
-        for event in events:
-            if lmcache_kv_events is None:
-                lmcache_kv_events = KVEventBatch(ts=time.time(), events=[])
-            block = BlockStored(
-                block_hashes=event.block_hashes,
-                parent_block_hash=event.parent_block_hash,
-                token_ids=event.token_ids,
-                lora_id=event.lora_id,
-                block_size=event.block_size,
-                medium=event.medium,
-            )
-            lmcache_kv_events.events.append(block)
-
-        return lmcache_kv_events
+    def get_block_ids_with_load_errors(self) -> set[int]:
+        """Return block IDs that failed to load during the last interval."""
+        return self._lmcache_engine.get_block_ids_with_load_errors()
 
     def shutdown(self):
         """
@@ -163,7 +133,7 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
         self,
         request: "Request",
         num_computed_tokens: int,
-    ) -> tuple[int | None, bool]:
+    ) -> tuple[Optional[int], bool]:
         """
         Get number of new tokens that can be loaded from the
         external KV cache beyond the num_computed_tokens.
@@ -203,30 +173,11 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
         """
         return self._lmcache_engine.build_connector_meta(scheduler_output)
 
-    def update_connector_output(self, connector_output: KVConnectorOutput):
-        """
-        Update KVConnector state from worker-side connectors output.
-
-        Args:
-            connector_output (KVConnectorOutput): the worker-side
-                connectors output.
-        """
-        # Get the KV events
-        kv_events = connector_output.kv_cache_events
-        if (
-            not kv_events
-            or not isinstance(kv_events, KVEventBatch)
-            or not kv_events.events
-        ):
-            return
-        self._kv_events.extend(kv_events.events)
-        return
-
     def request_finished(
         self,
         request: "Request",
         block_ids: list[int],
-    ) -> tuple[bool, dict[str, Any] | None]:
+    ) -> tuple[bool, Optional[dict[str, Any]]]:
         """
         Called when a request has finished, before its blocks are freed.
 
@@ -238,14 +189,3 @@ class LMCacheConnectorV1(KVConnectorBase_V1):
             returned by the engine.
         """
         return self._lmcache_engine.request_finished(request, block_ids)
-
-    def take_events(self) -> Iterable["KVCacheEvent"]:
-        """
-        Take the KV cache events from the connector.
-
-        Yields:
-            New KV cache events since the last call.
-        """
-        if self._kv_events is not None:
-            yield from self._kv_events
-            self._kv_events.clear()
--- a/lmcache/integration/vllm/lmcache_connector_v1.py
+++ b/lmcache/integration/vllm/lmcache_connector_v1.py
@@ -1,18 +1,17 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
 from collections.abc import Iterable
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, List, Optional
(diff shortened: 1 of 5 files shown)


lmcache/integration/vllm/lmcache_connector_v1.py
```diff
@@ -27,15 +31,63 @@
...
+            self.data["kv_events"] = []
+        self.data["kv_events"].append(event)
+
+    def get_kv_events(self) -> Optional[List[BlockStored]]:
+        if self.is_empty():
+            return None
+        return self.data["kv_events"]
+
+    def is_empty(self) -> bool:
+        return not self.data or self.data.get("kv_events", 0) == 0
```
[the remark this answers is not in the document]
hickeyma: N/A

lmcache/integration/vllm/lmcache_connector_v1.py
```diff
@@ -27,15 +31,63 @@
...
 
+@dataclass
+class LMCacheKVEvents(KVConnectorStats):
+    """
+    Maintain a list of KV events from worker to scheduler
+    """
+
+    def aggregate(self, other: "KVConnectorStats") -> "LMCacheKVEvents":
+        if not other and not isinstance(other, LMCacheKVEvents):
+            raise TypeError("Can only aggregate with another LMCacheKVEvents")
```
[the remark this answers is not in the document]
hickeyma: N/A

lmcache/integration/vllm/lmcache_connector_v1.py
```diff
@@ -114,9 +166,29 @@ def get_finished(
...
+                block_hashes=event.block_hashes,
+                parent_block_hash=event.parent_block_hash,
+                token_ids=event.token_ids,
+                lora_id=event.lora_id,
+                block_size=event.block_size,
+                medium=event.medium,
+            )
+            lmcache_kv_events.add_kv_event(block)
+
+        return lmcache_kv_events
```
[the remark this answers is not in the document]
hickeyma: N/A

lmcache/integration/vllm/lmcache_connector_v1.py
```diff
@@ -165,11 +237,30 @@ def build_connector_meta(
...
+        # Get the KV events
+        kv_events = connector_output.kv_connector_stats
+        if (
+            not kv_events
+            or not isinstance(kv_events, LMCacheKVEvents)
+            or kv_events.is_empty()
+        ):
+            return
+        self._kv_events = kv_events.get_kv_events()
+        return
```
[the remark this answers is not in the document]
hickeyma: N/A

sammshen (approved):
LGTM!

lmcache/v1/cache_engine.py
```diff
@@ -335,6 +346,30 @@ def store(
...
 
+            # Create KV event
+            if self.kv_events_enabled:
+                stored_event = CacheStoreEvent(
+                    block_hashes=[key.chunk_hash],
+                    parent_block_hash=None if start == 0 else prev_key,
+                    token_ids=[],
+                    block_size=num_tokens,
+                    lora_id=None,
+                    medium="GPU",
```
ApostaC: Do we need to change the medium here?

lmcache/v1/cache_engine.py
```diff
@@ -447,6 +483,24 @@ def store_layer(
...
+            if self.kv_events_enabled and tokens is not None:
+                stored_event = CacheStoreEvent(
+                    block_hashes=[key.chunk_hash],
+                    parent_block_hash=None if start == 0 else prev_key,
+                    token_ids=tokens.tolist()[start : end + 1]
+                    if isinstance(tokens, torch.Tensor)
+                    else tokens[start : end + 1],
+                    block_size=num_tokens,
+                    lora_id=None,
+                    medium="GPU",
```
ApostaC: Same question to the `medium` as above.

lmcache/v1/cache_engine.py
```diff
@@ -447,6 +483,24 @@ def store_layer(
...
             tot_token_num += num_tokens
 
+            # Create KV event
+            if self.kv_events_enabled and tokens is not None:
+                stored_event = CacheStoreEvent(
+                    block_hashes=[key.chunk_hash],
+                    parent_block_hash=None if start == 0 else prev_key,
+                    token_ids=tokens.tolist()[start : end + 1]
+                    if isinstance(tokens, torch.Tensor)
+                    else tokens[start : end + 1],
```
ApostaC: QQ: When the `tokens` is a tensor, can we make sure that it's not on the GPU? Otherwise, calling `tolist` will force a d2h copy and cause potential stream sync.

ApostaC (commented):
Otherwise LGTM! Thanks!

lmcache/v1/cache_engine.py
```diff
@@ -1134,6 +1188,14 @@ def clear(
                 return 0
         return self._clear(tokens, locations, request_configs)
 
+    @_lmcache_nvtx_annotate
+    def get_kv_events(self) -> Iterable[CacheStoreEvent]:
+        if self.kv_events_enabled and self.kv_events:
```
sammshen: nit: 
```python
        if self.kv_events_enabled and events := self.kv_events:
            self.kv_events = []
            return events
```

lmcache/v1/cache_engine.py
hickeyma: Good catch.

hickeyma: Updated to get device where applicable, otherwise "cpu" by default.

lmcache/v1/cache_engine.py
hickeyma: Ditto

lmcache/v1/cache_engine.py
hickeyma: That's a good insight, thanks. Updated logic to check for device and if CUDA then do a `detach` to cpu for getting the list. If you have a better approach, I'd be happy to use it.

lmcache/v1/cache_engine.py
hickeyma: Done

hickeyma:
Thanks @ApostaC and @sammshen for the reviews and feedback. Updated and ready for review again.

lmcache/v1/cache_engine.py
ApostaC: Hey @hickeyma , I vaguely remember the token ids tensor is on CPU and will not be on GPU. Can you run some quick tests to see if it's true?
If so, we don't need to have this extra check.

ApostaC: Additionally, since we have the same logic (converting token ids to list) in 2 different places, it would be better if we could define a helper function that can be reused.

hickeyma: Added utility function for converting tokens to list

hickeyma: >  I vaguely remember the token ids tensor is on CPU and will not be on GPU. Can you run some quick tests to see if it's true?

It looks like that they are CPU only. I will remove the GPU optimization then.

hickeyma:
Thanks @ApostaC for the feedback. PR is updated and should be ready to go.

ApostaC (approved):
LGTM! @sammshen @KuntaiDu Can you take a quick look as well? Thanks!

sammshen (approved):
LGTM!

PR #1960, opened 2025-11-06
[hot fix]: add blocking_timeout_sec semantics to async codepath

sammshen:
Previously, `blocking_timeout_sec` only applies to synchronous KV loading codepath. 

This PR adds new semantics to `blocking_timeout_secs` for the async loading codepath, covering two main use cases: 
1. when `async_lookup()` or `batched_get_non_blocking()` take too long for a remote backend
2. when `get_num_new_matched_tokens()` has been called repeatedly on a request for too long, we return 0 and force recompute. 

how to test this fix:

`async.yaml`

```yaml
enable_async_loading: True
chunk_size: 256
local_cpu: False
max_local_cpu_size: 10
remote_url: "mock://100/?peeking_latency=30000&read_throughput=2&write_throughput=2"
```

```bash
LMCACHE_CONFIG_FILE=async.yaml \
vllm serve Qwen/Qwen3-8B \
    --port 8000 --kv-transfer-config \
```

```bash
    curl -X POST http://localhost:8000/v1/completions \
      -d '{
        "model": "Qwen/Qwen3-8B",
        "prompt": "'"$(printf 'Elaborate the significance of KV cache in language models. %.0s' {1..1000})"'",
        "max_tokens": 10
      }'
```

the error handling is graaceful: 
```text
[1;36m(EngineCore_DP0 pid=3595044)[0;0m [1m[2025-11-06 22:44:57,100] LMCache DEBUG:[0m In connector.start_load_kv, but the attn_metadata is None [3m(vllm_v1_adapter.py:848:lmcache.integration.vllm.vllm_v1_adapter)[0m
[1;36m(EngineCore_DP0 pid=3595044)[0;0m [33;20m[2025-11-06 22:44:57,106] LMCache WARNING:[0m batched_async_contains timed out [3m(remote_backend.py:482:lmcache.v1.storage_backend.remote_backend)[0m
[1;36m(EngineCore_DP0 pid=3595044)[0;0m [33;20m[2025-11-06 22:44:57,111] LMCache WARNING:[0m Request cmpl-c80b5d53924b442b9cf5845ef632e3f4-0 is still waiting for async lookup result after 10 seconds,returning 0 cached tokens so vllm can conduct recompute. [3m(vllm_v1_adapter.py:1315:lmcache.integration.vllm.vllm_v1_adapter)[0m
```


sammshen:
@chunxiaozheng you were the one to previously add the `blocking_timeout_secs`, would you like to take a look?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1306,6 +1308,20 @@ def get_num_new_matched_tokens(
         )
 
         if num_external_hit_tokens is None:
+            if request.request_id not in self.first_async_lookup:
```
maobaolong: Do you think it can be good to maintain if we put this logic into lmcache lookup client? @sammshen

maobaolong (commented):
@sammshen As we decouple the vllm_v1_adapter and lmcache core, so we'd better put changes in the lmcache repository as possible as we can. WDYT?

lmcache/integration/vllm/vllm_v1_adapter.py
ApostaC: Agreed. We should avoid putting too much vllm-agnostic logic into `vllm_v1_adapter` from now on.

ApostaC (commented):
Mostly LGTM! Also thanks for the doc update!

Can we change the lookup timeout monitoring to the lookup client? Let's avoid putting more logics into `vllm_v1_adapter.py`

lmcache/v1/internal_api_server/env_api.py
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: Apache-2.0
```
sammshen, line 1: this is not from this PR

lmcache/utils.py
```diff
@@ -298,6 +298,18 @@ def from_dict(d):
             request_configs=request_configs,
         )
 
+    def with_new_worker_id(self, new_worker_id: int) -> "CacheEngineKey":
```
sammshen, line 301: this is not from this PR

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: changed and tested again

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -452,7 +460,19 @@ async def batched_get_non_blocking(
...
             return []
-        return await self.connection.batched_get_non_blocking(lookup_id, keys)
+        try:
+            # warning, this timeout will not actually stop the
+            # scheduler from waiting for the result
+            return await asyncio.wait_for(
+                self.connection.batched_get_non_blocking(lookup_id, keys),
+                self.blocking_timeout_secs,
+            )
+        except asyncio.TimeoutError:
```
yoo-kumaneko, line 470: What will happen after the Timeout error? Will the self.connection.batched_get_non_blocking task be cancelled? What about the resources allocated in the task? Like memory objects. Do we need to free those resources?

yoo-kumaneko: We can skip cleaning up the memory resources in this PR and address it in a future one.

sammshen: agreed

yoo-kumaneko (approved):
LGTM

lmcache/v1/storage_backend/remote_backend.py
ApostaC: From what I read in Python documentation, it will cancel the underlying coroutine when the timeout occurs: https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for

ApostaC (approved):
LGTM!

sammshen:
@maobaolong would you like to take another look?

maobaolong (approved):
LGTM @sammshen Thanks for this improvement!

PR #1925, opened 2025-10-30
[CI/Build] Add /opt/venv/bin to PATH

ningziwen:
**What this PR does / why we need it**:

The latest VLLM requires ninja binary at runtime based on the error log.

```
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]   File "/opt/venv/lib/python3.12/site-packages/vllm/v1/sample/sampler.py", line 100, in forward
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]     sampled, processed_logprobs = self.sample(logits, sampling_metadata)
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]   File "/opt/venv/lib/python3.12/site-packages/vllm/v1/sample/sampler.py", line 180, in sample
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]     random_sampled, processed_logprobs = self.topk_topp_sampler(
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]                                          ^^^^^^^^^^^^^^^^^^^^^^^
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]   File "/opt/venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1773, in _wrapped_call_impl
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]     return self._call_impl(*args, **kwargs)
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]   File "/opt/venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1784, in _call_impl
(Worker_TP0 pid=428) ERROR 10-30 18:49:38 [multiproc_executor.py:671]     return forward_call(*args, **kwargs)
[...]

PR #1962, opened 2025-11-07
[CI/Build] Move /opt/venv/bin to PATH to BASE image

ningziwen:

The previous change only fixed for image-build target, which is used for nightly build and local. https://github.com/LMCache/LMCache/pull/1925

We also need to add env var change in image-release target. This issue is still not fixed in v0.3.9post1 without this change.

**If applicable**:


sammshen (approved):
LGTM

ApostaC (approved):
LGTM!

PR #1963, opened 2025-11-07
[controller] simplify config when p2p is not enabled

chunxiaozheng:
When `P2P` is enabled, the controller must be also enabled. But in some scenarios, we want to enable controller, but not use `P2P`, the `P2P` related configs is not necessary.

### DIFF
--- a/lmcache/v1/cache_controller/controllers/kv_controller.py
+++ b/lmcache/v1/cache_controller/controllers/kv_controller.py
@@ -219,6 +219,7 @@ class KVController:
                 peer_init_url = self.reg_controller.get_distributed_url(
                     instance_id, worker_id
                 )
+                assert peer_init_url is not None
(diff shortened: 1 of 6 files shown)


chunxiaozheng:
@YaoJiayi @maobaolong could you help take a look?

maobaolong (approved):
@chunxiaozheng Thanks for your improvement, indeed, using LMCache Controller may not need distributed server. LGTM

YaoJiayi (approved):
@chunxiaozheng LGTM. Do you have interest for adding some tests to the controller?

chunxiaozheng:
> @chunxiaozheng LGTM. Do you have interest for adding some tests to the controller?

@YaoJiayi Thanks for your review! I will add some ut for controller. 😄

PR #1973, opened 2025-11-10
[core] Support remote cache put and get metrics in sagemaker hyperpod connector

ningziwen:
… connector

The current remote put and get metrics are implemented in instrumented connector, but only works for get() and put() instead of batched functions.

https://github.com/LMCache/LMCache/blob/a64ca32cba3784fa2266b7343c7591ec88d6e277/lmcache/v1/storage_backend/connector/instrumented_connector.py#L37

batched_get() and batched_put() are called when load is high enough. Since batch functions call batch functions in each remote connector, the metric reporting needs to be in remote connector for now unless larger refactoring.

This PR only enables put and get metrics in sagemaker hyperpod connector.

There are two potential TODOs after this:

1. Enable get and put metrics in other remote connectors.
2. Explore larger refactoring to make the logic in instrumented connector.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
+++ b/lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
@@ -6,6 +6,7 @@ from multiprocessing import shared_memory
 from typing import AsyncIterator, List, Optional, Tuple
 import asyncio
 import json
+import time
 import urllib.parse
 
 # Third Party
@@ -14,6 +15,7 @@ import torch
 
 # First Party
 from lmcache.logging import init_logger
+from lmcache.observability import LMCStatsMonitor
 from lmcache.utils import CacheEngineKey
 from lmcache.v1.memory_management import MemoryObj
 from lmcache.v1.protocol import RemoteMetadata
@@ -134,6 +136,9 @@ class SageMakerHyperPodConnector(RemoteConnector):
         # Shared memory (lazy initialized)
         self.shared_memory_obj: Optional[shared_memory.SharedMemory] = None
         self.shared_memory_map: Optional[memoryview] = None
+
+        # Observability
+        self._stats_monitor = LMCStatsMonitor.GetOrCreate()
 
         # Statistics for monitoring
         self.stats = {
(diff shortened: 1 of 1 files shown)


sammshen:
I don't understand why we don't wrap timing around the instrumented connector methods? 

```python
    def support_batched_put(self) -> bool:
        return self._connector.support_batched_put()

    def support_batched_get(self) -> bool:
        return self._connector.support_batched_get()
 ```

**Comment by ningziwen** [comment]
@sammshen As mentioned in PR description, I think these metrics are for single put/get. Wrapping batch methods are not reflecting single put/get. Let me know if I miss anything.

**Comment by sammshen** [comment]
@ningziwen sorry I pasted the wrong code, I meant these two methods: 
 
 ```python
     def support_batched_async_contains(self) -> bool:
        return self._connector.support_batched_async_contains()

    async def batched_async_contains(
        self,
        lookup_id: str,
        keys: List[CacheEngineKey],
        pin: bool = False,
    ) -> int:
        return await self._connector.batched_async_contains(lookup_id, keys, pin)

    def support_batched_get_non_blocking(self) -> bool:
        return self._connector.support_batched_get_non_blocking()

    async def batched_get_non_blocking(
        self,
        lookup_id: str,
        keys: List[CacheEngineKey],
    ) -> List[MemoryObj]:
        return await self._connector.batched_get_non_blocking(lookup_id, keys)

```

ningziwen:
@sammshen They are calling batch methods as I mentioned. Let me know if I missed anything.

sammshen (approved):
LGTM, thanks for the explanation

ningziwen:
@maobaolong @YaoJiayi @ApostaC Need one more approval. Could you pls help?

maobaolong:
@ningziwen A high level question, we have supply audit_connector to measure the time cost and operation audit, is that enough? If not, do you think we can move the logic of get time costing calculate to `remote_backend`?

ningziwen:
@maobaolong We would like metrics written to Prometheus. Looks like audit connector only supports logging today. Since metrics are not supported out of box yet, could I look into how to integrate metrics to audit connector and move to audit connector as follow-up?

maobaolong (approved):
LGTM

PR #1979, opened 2025-11-11
[MINOR]Add metrics for scheduler role

maobaolong:
Introduce some metrics for scheduler role.

Clipboard_Screenshot_1762830365

Clipboard_Screenshot_1762833531

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -23,12 +23,10 @@ from vllm.sampling_params import SamplingParams
 
 # First Party
 # Use LMCache's own math utilities instead of vllm's
-# (avoids dependency on vllm internal changes like
-# https://github.com/vllm-project/vllm/pull/27188)
+# (avoids dependency on vllm internal changes like https://github.com/vllm-project/vllm/pull/27188)
 from lmcache.utils import cdiv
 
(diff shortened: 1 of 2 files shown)


maobaolong:
@sammshen @chunxiaozheng @yoo-kumaneko Would you like to take a look at this PR? thanks!

yoo-kumaneko:
LGTM!

yoo-kumaneko (approved):

chunxiaozheng (approved):
lgtm

sammshen (approved):
LGTM, these seem mostly useful for debugging? it will expose some pretty internal state to Prometheus. Is this best practice? Perhaps the internal API server would be a better use case for this debug type introspection?

maobaolong:
@sammshen Thanks for your review. 

And thanks for the point! I get your concern.

This PR is a first one adding metrics for scheduler process. The main reason we're exposing these metrics is for production monitoring, not just one-off debugging, we are trying to use `asyc_loading` feature, that feature make us worry about the internal state of LMCache scheduler process whether work as expected or get leaking. The metrics of scheduler exposed are essential for our alerts and dashboards to keep the service healthy. 

Whatever, your are right, we will be careful while adding new metrics from then now, thanks for your suggestion!

sammshen:
Thanks for the explanation! @maobaolong

PR #1983, opened 2025-11-11
[controller][core] support start lmcache worker on part ranks

chunxiaozheng:
Support start lmcache worker on part workers.
If use mla, lmcache worker only starts on rank0 can work.

### DIFF
--- a/lmcache/v1/cache_controller/worker.py
+++ b/lmcache/v1/cache_controller/worker.py
@@ -96,14 +96,15 @@ class LMCacheWorker:
         lmcache_worker_ids = config.get_lmcache_worker_ids(
             metadata.use_mla, metadata.world_size
         )
-        if len(lmcache_worker_ids) != len(config.lmcache_worker_ports):
-            raise ValueError(
-                f"lmcache_worker_ids and lmcache_worker_ports must have the same length, "
-                f"len of lmcache_worker_ids: {len(lmcache_worker_ids)}, "
-                f"len of lmcache_worker_ports: {len(config.lmcache_worker_ports)}"
-            )
-        index = lmcache_worker_ids.index(self.worker_id)
-        lmcache_worker_port = config.lmcache_worker_ports[index]
+        if len(lmcache_worker_ids) == 0:
+            # start lmcache worker on all ranks
+            assert len(config.lmcache_worker_ports) == metadata.world_size
+            lmcache_worker_port = config.lmcache_worker_ports[self.worker_id]
+        else:
+            # start lmcache worker on given worker ids
+            assert len(lmcache_worker_ids) == len(config.lmcache_worker_ports)
+            index = lmcache_worker_ids.index(self.worker_id)
+            lmcache_worker_port = config.lmcache_worker_ports[index]
 
         self.lmcache_worker_internal_url = f"*:{lmcache_worker_port}"
         self.lmcache_worker_ip = get_ip()
--- a/lmcache/v1/cache_controller/worker.py
+++ b/lmcache/v1/cache_controller/worker.py
@@ -93,7 +93,17 @@ class LMCacheWorker:
                 bind_or_connect="connect",
             )
 
-        lmcache_worker_port = config.lmcache_worker_ports[self.worker_id]
+        lmcache_worker_ids = config.get_lmcache_worker_ids(
+            metadata.use_mla, metadata.world_size
+        )
+        if len(lmcache_worker_ids) != len(config.lmcache_worker_ports):
+            raise ValueError(
+                f"lmcache_worker_ids and lmcache_worker_ports must have the same length, "
+                f"len of lmcache_worker_ids: {len(lmcache_worker_ids)}, "
+                f"len of lmcache_worker_ports: {len(config.lmcache_worker_ports)}"
+            )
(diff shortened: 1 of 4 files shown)


maobaolong (approved):
LGMT for this PR, discussed with @chunxiaozheng offline, maybe we can set the default port to 0 which stand for gain a serial of available ports automatically, hope some volunteer can finish this by another PR.

ApostaC:
@YaoJiayi Please take a look at this one.

lmcache/v1/config.py
```diff
@@ -571,6 +577,20 @@ def _get_extra_config_value(self, key, default_value=None):
         return default_value
 
 
+def _get_lmcache_worker_ids(self, use_mla, world_size):
+    if self.lmcache_worker_ids is None:
+        # if mla is not enabled, return [], which means start
+        # lmcache worker on all ranks as default;
+        # if mla is enabled, return [0], which means start lmcache
+        # worker on worker 0 as default.
+        return [0] if use_mla else []
```
YaoJiayi: Why are we returning empty `[]` when mla is not enabled? Shouldn't we return all worker ids?

lmcache/v1/cache_engine.py
```diff
@@ -120,7 +120,16 @@ def __init__(
...
         self.lmcache_worker: Optional[LMCacheWorker] = None
-        if self.enable_controller and self.metadata.role != "scheduler":
+        lmcache_worker_ids = config.get_lmcache_worker_ids(
+            metadata.use_mla, metadata.world_size
+        )
+        if (
+            self.enable_controller
+            and self.metadata.role != "scheduler"
+            and (
+                len(lmcache_worker_ids) == 0 or metadata.worker_id in lmcache_worker_ids
```
YaoJiayi: The condition `len(lmcache_worker_ids) == 0` is confusing. Should we set `len(lmcache_worker_ids) == number of total ranks` by default?

YaoJiayi (commented):
LGTM in general. Just a few minor comments.
Btw, I feel we have a lot of mla-specific configurations? Do we have any documentation for these?

chunxiaozheng:
> LGTM in general. Just a few minor comments. Btw, I feel we have a lot of mla-specific configurations? Do we have any documentation for these?

@YaoJiayi Thanks for your suggestion, I will add some docs in next independent PR.

lmcache/v1/config.py
chunxiaozheng: > Why are we returning empty `[]` when mla is not enabled? Shouldn't we return all worker ids?

In order to maintain consistency with `lookup_server_worker_ids`, if `world_size` is 8, empty `[]` or `[0, 1, 2, 3, 4, 5, 6, 7]` both represents all worker ids.

lmcache/v1/cache_engine.py
chunxiaozheng: > The condition `len(lmcache_worker_ids) == 0` is confusing. Should we set `len(lmcache_worker_ids) == number of total ranks` by default?

if `lmcache_worker_ids` includes all worker ids, `metadata.worker_id in lmcache_worker_ids` will return `True`, `len(lmcache_worker_ids) == number of total ranks` is redundant. 
I think use empty `[]` to represents all worker ids is reasonable, otherwise, if `lmcache_worker_ids` is empty and nothing starts, there will be problems.

chunxiaozheng:
@YaoJiayi I have updated, PTAL.

lmcache/v1/cache_engine.py
KuntaiDu: Maybe `not lmcache_worker_ids` is easier to understand compared to `len(lmcache_worker_ids) == 0`. Basically `[]` is more like an "unset" configuration and we should fall back to default behavior, which is launching all workers.

lmcache/v1/config.py
```diff
@@ -571,13 +577,27 @@ def _get_extra_config_value(self, key, default_value=None):
         return default_value
 
 
+def _get_lmcache_worker_ids(self, use_mla, world_size):
+    if self.lmcache_worker_ids is None:
```
KuntaiDu: Same here, prefer `if not self.lmcache_worker_ids`

chunxiaozheng: @KuntaiDu Thanks for your suggestion!

chunxiaozheng:
@KuntaiDu I have updated this PR, PTAL, Thanks!

lmcache/v1/cache_controller/worker.py
```diff
@@ -93,7 +93,18 @@ def __init__(
                 bind_or_connect="connect",
             )
 
-        lmcache_worker_port = config.lmcache_worker_ports[self.worker_id]
+        lmcache_worker_ids = config.get_lmcache_worker_ids(
+            metadata.use_mla, metadata.world_size
+        )
+        if not lmcache_worker_ids:
```
KuntaiDu, line 99: Is this branch redundant, since we set lmcache_worker_ids to `list(range(world_size))`? If so we can remove this branch.

KuntaiDu (approved):
LGTM, some nits in the comments.

YaoJiayi (approved):
lgtm!

PR #1986, opened 2025-11-12
[optimize] do not initialize storage manager in some rank

chunxiaozheng:
if `save_only_first_rank` is enabled, the first rank will execute put and get task, other ranks retrieve kv cache by broadcast, there is no need to init the storage manager of these ranks.

### DIFF
--- a/lmcache/v1/storage_backend/remote_backend.py
+++ b/lmcache/v1/storage_backend/remote_backend.py
@@ -51,10 +51,6 @@ class RemoteBackend(StorageBackendInterface):
         self.connection: Optional[RemoteConnector] = None
         self.min_reconnect_interval = 10
         self.failure_time = -1000000.0
-        self.save_only_first_rank = (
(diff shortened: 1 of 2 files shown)


chunxiaozheng:

lmcache/v1/storage_backend/remote_backend.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         from lmcache.v1.storage_backend.remote_monitor import RemoteMonitor
 
         self.remote_monitor = RemoteMonitor(self)
+
```
maobaolong, line 87: revert this

maobaolong (approved):
@chunxiaozheng left a minor comment, otherwise, LGTM.

lmcache/v1/storage_backend/remote_backend.py
chunxiaozheng: done

sammshen (approved):
LGTM

PR #2004, opened 2025-11-15
[Core] Add a timeout approach for last pin operation to avoid leak

maobaolong:

Add a timeout approach for last pin operation to avoid memory obj leaking

**If applicable**:

### DIFF
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -441,6 +441,12 @@ _CONFIG_DEFINITIONS: dict[str, dict[str, Any]] = {
         "default": False,
         "env_converter": _to_bool,
     },
+    # Memory management configurations
+    "pin_timeout_seconds": {
+        "type": int,
+        "default": 300,
+        "env_converter": int,
+    },
 }
 
 
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -442,9 +442,14 @@ _CONFIG_DEFINITIONS: dict[str, dict[str, Any]] = {
         "env_converter": _to_bool,
     },
     # Memory management configurations
-    "pin_timeout_seconds": {
+    "pin_timeout_sec": {
         "type": int,
         "default": 300,
+        "env_converter": int,
+    },
+    "pin_check_interval_sec": {
+        "type": int,
+        "default": 30,
         "env_converter": int,
     },
 }
--- a/lmcache/v1/config.py
+++ b/lmcache/v1/config.py
@@ -446,11 +446,21 @@ _CONFIG_DEFINITIONS: dict[str, dict[str, Any]] = {
         "type": int,
         "default": 300,
         "env_converter": int,
+        "description": (
+            "Maximum duration in seconds that a memory object can remain pinned. "
+            "If a pinned object exceeds this timeout, it will be forcibly unpinned "
+            "by the PinMonitor to prevent memory leaks. Default is 300 seconds."
+        ),
     },
     "pin_check_interval_sec": {
         "type": int,
         "default": 30,
         "env_converter": int,
+        "description": (
+            "Interval in seconds between PinMonitor timeout checks. "
+            "The background thread periodically scans all pinned objects at this "
+            "interval to detect and handle timeouts. Default is 30 seconds."
+        ),
     },
 }
 
--- a/lmcache/v1/pin_monitor.py
+++ b/lmcache/v1/pin_monitor.py
@@ -102,6 +102,8 @@ class PinMonitor:
                     elapsed_time = current_time - register_time
                     if elapsed_time > self._pin_timeout_sec:
                         timeout_objects.append((memory_obj, elapsed_time))
+                else:
+                    self.unregister_pinned_object(memory_obj)
 
         # Force unpin timeout objects outside the lock to avoid deadlocks
         for memory_obj, elapsed_time in timeout_objects:
@@ -116,33 +118,27 @@ class PinMonitor:
         self, memory_obj: "TensorMemoryObj", elapsed_time: float
     ):
         """Force unpin a timeout object and log the event."""
+        # Get current pin_count without holding the lock for unpin calls
         with memory_obj.lock:
-            if memory_obj.meta.pin_count > 0:
-                logger.warning(
-                    "Pin timeout detected for MemoryObj %s. "
-                    "Pin count: %s, Elapsed time: %.2fs. Forcing unpin to 0.",
-                    memory_obj.meta.address,
-                    memory_obj.meta.pin_count,
-                    elapsed_time,
-                )
-
-                # Update monitoring statistics
-                LMCStatsMonitor.GetOrCreate().update_forced_unpin_count(1)
-                if memory_obj.meta.pin_count > 0:
-                    LMCStatsMonitor.GetOrCreate().update_pinned_memory_objs_count(-1)
-
-                # Force unpin
-                memory_obj.meta.pin_count = 0
-
-                # Unregister from monitoring
+            current_pin_count = memory_obj.meta.pin_count
+            if current_pin_count <= 0:
                 self.unregister_pinned_object(memory_obj)
+                return
 
-                # Free the object if ref_count is also 0
-                if (
-                    memory_obj.meta.ref_count <= 0
-                    and memory_obj.parent_allocator is not None
-                ):
-                    memory_obj.parent_allocator.free(memory_obj)
+            logger.warning(
+                "Pin timeout detected for MemoryObj %s. "
+                "Pin count: %s, Elapsed time: %.2fs. Forcing unpin to 0.",
+                memory_obj.meta.address,
+                current_pin_count,
+                elapsed_time,
+            )
+
+        # Update forced unpin statistics
+        LMCStatsMonitor.GetOrCreate().update_forced_unpin_count(1)
+
+        # Call unpin() while pin_count > 0 to properly release resources
+        while memory_obj.meta.pin_count > 0:
+            memory_obj.unpin()
 
     def _monitor_loop(self):
         """Background thread loop for monitoring pinned objects."""
--- a/lmcache/v1/pin_monitor.py
+++ b/lmcache/v1/pin_monitor.py
@@ -1,0 +1,190 @@
+# SPDX-License-Identifier: Apache-2.0
+# Standard
+from typing import TYPE_CHECKING, Optional
+import threading
+import time
+
+# First Party
+from lmcache.logging import init_logger
+from lmcache.observability import LMCStatsMonitor
+
+if TYPE_CHECKING:
+    # First Party
+    from lmcache.v1.config import LMCacheEngineConfig
+    from lmcache.v1.memory_management import TensorMemoryObj
+
+
+logger = init_logger(__name__)
+
+
+class PinMonitor:
+    """
+    Global monitor for pinned TensorMemoryObj instances to handle timeout detection.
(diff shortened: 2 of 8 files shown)


ApostaC (commented):
In general LGTM!

Quick question: can this PR address the memory leak problem when `unpin` is never called?

maobaolong:
> Quick question: can this PR address the memory leak problem when `unpin` is never called?

@ApostaC This is a good question, how silly i am! You are right, if `unpin` is never called, this function will never work.

I will think another way to trigger the check. Thanks advance for your question!

ApostaC:
@maobaolong Thanks for the follow-up. I'm actually good with the current PR. Let me know if you would like to proceed.

My thoughts on how to address the "unpin never be called" problem: we can check the timestamp during eviction. If an object is pinned but the timestamp is too "old", we can still evict it and report the number to the observability module. This will work because the memory pin will only impact when we are doing the eviction.

maobaolong:
> @maobaolong Thanks for the follow-up. I'm actually good with the current PR. Let me know if you would like to proceed.
> 
> My thoughts on how to address the "unpin never be called" problem: we can check the timestamp during eviction. If an object is pinned but the timestamp is too "old", we can still evict it and report the number to the observability module. This will work because the memory pin will only impact when we are doing the eviction.

@ApostaC Thanks for sharing you idea, this is one of the possible approach, but I used another approach, let me summary the two approach here.
- Approach A, lazy check, check while evict.
    - Pros: Do not involve another background thread.
    - Cons: 1. Leak detected postponed until memory full. 2. Small affect to eviction.
- Approach B, background check monitor, check periodically.
    - Pros: 1. Find leak pin memory object within one sleep internal. 2. No affect to eviction. 3. Number of pin MemoryObj easy measurable.
    - Cons: Introduced a new pin monitor background thread.

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Args:
+            config: Required for first-time initialization.
+                Optional for subsequent calls.
+
+        Raises:
+            ValueError: If config is None when creating the instance
+                for the first time.
+        """
+        if PinMonitor._instance is None:
```
ApostaC, line 58: Can we remove the out-most `if` here?

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,190 @@
...
+
+                # Unregister from monitoring
+                self.unregister_pinned_object(memory_obj)
+
+                # Free the object if ref_count is also 0
+                if (
+                    memory_obj.meta.ref_count <= 0
+                    and memory_obj.parent_allocator is not None
+                ):
+                    memory_obj.parent_allocator.free(memory_obj)
```
ApostaC: Will this cause "double free"? When the ref_count reaches zero before unpin, will the memory object be freed?

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,190 @@
...
+            with PinMonitor._lock:
+                if PinMonitor._instance is None:
+                    assert config is not None, "config is required"
+                    PinMonitor._instance = PinMonitor(config)
+        return PinMonitor._instance
+
+    def register_pinned_object(self, memory_obj: "TensorMemoryObj"):
+        """Register a pinned memory object for timeout monitoring.
+        Updates the register time each time this is called.
+        """
```
ApostaC: Note that we could ping the same object multiple times, so this function might be called multiple times with the same memory_obj.

I think the current implementation is fine, but it would be useful to explicitly note it here.

ApostaC (commented):
One concern about this implementation is that it will do a "full-scan" on the internal dictionary every 30 seconds. It would be great if we could double check the performance.

Left some minor comments as well, otherwise LGTM.

lmcache/v1/pin_monitor.py
maobaolong: If we remove the out-most if, each caller will try to obtain `PinMonitor._lock` lock, but the _instance will never be none once assigned.

lmcache/v1/pin_monitor.py
maobaolong: It won't be double free, since memory_obj.invalidate() will be executed during free and it protected by lock, and if another free called, `memory_obj.is_valid()` is False, will prevent double free.

lmcache/v1/pin_monitor.py
maobaolong: Done, updated the comment.

maobaolong:
@ApostaC Thanks for your comments, all the comments addressed, would you like to help to review this PR again? Thank!

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,194 @@
...
+                logger.warning(
+                    "Pin timeout detected for MemoryObj %s. "
+                    "Pin count: %s, Elapsed time: %.2fs. Forcing unpin to 0.",
+                    memory_obj.meta.address,
+                    memory_obj.meta.pin_count,
+                    elapsed_time,
+                )
+
+                # Update monitoring statistics
+                LMCStatsMonitor.GetOrCreate().update_forced_unpin_count(1)
```
maobaolong: reuse the lock within memory object.

maobaolong:
@ApostaC Use the unpin to force unpin the overdue memory object. PTAL.

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,190 @@
...
+            for obj_id, (memory_obj, register_time) in list(
+                self._pinned_objects.items()
+            ):
+                # Check if object is still pinned and has exceeded timeout
+                if memory_obj.meta.pin_count > 0:
+                    elapsed_time = current_time - register_time
+                    if elapsed_time > self._pin_timeout_sec:
+                        timeout_objects.append((memory_obj, elapsed_time))
+                else:
+                    self.unregister_pinned_object(memory_obj)
```
ApostaC: Quick question: in what will we hit this branch? 
If I understand correctly, when `pin_count` reaches zero, it will always call `unregister_pinned_object` from `MemoryObj.unpin`. 

If we don't need to call `unregister_pinned_object` here, we could use `threading.Lock` instead of `threading.RLock`.

ApostaC (approved):
Just one quick question, otherwise LGTM!

ApostaC:
@sammshen Can you also do a review for this PR? Thanks!

maobaolong:
@ApostaC Thanks for your review and suggestion, removed unregister_pinned_object call and use threading.Lock instead of threading.RLock.

@sammshen Would you like to help to take another look? Thanks!

lmcache/v1/pin_monitor.py
sammshen: but this introduces race condition right? on the out-most call

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,205 @@
...
+    that have exceeded their timeout duration.
+    """
+
+    _instance = None
+    _lock = threading.Lock()
+
+    def __init__(self, config: "LMCacheEngineConfig"):
+        self._pinned_objects: dict[
+            int, tuple["TensorMemoryObj", float]
+        ] = {}  # {obj_id: (memory_obj, register_time)}
```
sammshen, line 36: can we also leave a comment here that the obj_id is the virtual memory address given by python?

lmcache/v1/config.py
```diff
@@ -441,6 +441,17 @@
         "default": False,
         "env_converter": _to_bool,
     },
+    # Memory management configurations
```
sammshen, line 444: the difference between these two should be described more clearly with the description attribute

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,205 @@
...
+    from lmcache.v1.config import LMCacheEngineConfig
+    from lmcache.v1.memory_management import TensorMemoryObj
+
+
+logger = init_logger(__name__)
+
+
+class PinMonitor:
+    """
+    Global monitor for pinned TensorMemoryObj instances to handle timeout detection.
```
sammshen: can you add a parentheses about what you mean by "Global" here? (like only one per vllm instance with multiple workers / cache engines or something like that). It should be clear

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,205 @@
...
+                for the first time.
+        """
+        if PinMonitor._instance is None:
+            with PinMonitor._lock:
+                if PinMonitor._instance is None:
+                    assert config is not None, "config is required"
+                    PinMonitor._instance = PinMonitor(config)
+        return PinMonitor._instance
+
+    def register_pinned_object(self, memory_obj: "TensorMemoryObj"):
```
sammshen: suggestion, rename to `on_pin` and `on_unpin` to make it clear that these are hooks that are meant to be called inside the memory object/manager

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,205 @@
...
+            self._pinned_objects[obj_id] = (memory_obj, current_time)
+            logger.debug(
+                "Registered pinned object %s for timeout monitoring at time %.2f",
+                obj_id,
+                current_time,
+            )
+
+    def unregister_pinned_object(self, memory_obj: "TensorMemoryObj"):
+        """Unregister a memory object from timeout monitoring."""
+        with self._objects_lock:
```
sammshen, line 86: nit: this only needs to protect `self._pinned_objects` and the obj_id = id(memory_obj) can go outside

lmcache/v1/pin_monitor.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def register_pinned_object(self, memory_obj: "TensorMemoryObj"):
+        """Register a pinned memory object for timeout monitoring.
+
+        Note: The same memory_obj can be pinned multiple times, so this
+        function may be called multiple times with the same object.
+        Each call updates the register time, effectively resetting the
+        timeout countdown.
+        """
+        with self._objects_lock:
```
sammshen, line 74: same comment about lock scope as below

lmcache/v1/pin_monitor.py
maobaolong: yeah, but if it is none, we can use lock for double check, if it is not none, it means other caller has been assigned a valid instance to that singleton variable.

lmcache/v1/pin_monitor.py
maobaolong: Sure

lmcache/v1/config.py
maobaolong: Done

lmcache/v1/pin_monitor.py
maobaolong: Done

lmcache/v1/pin_monitor.py
maobaolong: Good suggestion, done

lmcache/v1/pin_monitor.py
maobaolong: Great! Done.

lmcache/v1/pin_monitor.py
maobaolong: Done

maobaolong:
@sammshen Thanks for your review and helpful suggestion, i've addressed the comments from you, PTAL.

sammshen (approved):
LGTM, thanks for addressing the nits!

PR #2006, opened 2025-11-15
[Feat] fs_connector: adjust the fs tree to two-tier hash directory tree

hb-lee:
As the KV Cache files grow, the existing directory structure is prone to generating extremely large directories, which poses significant pressure on file lookup and traversal. Therefore, we adjust the single directory tree structure to a two-tier hash directory tree structure, allowing for a more even distribution of KV Cache files and avoiding the creation of large directories.
In addition, the new directory structure places KV Cache temporary files uniformly under base_path, enabling unified management of the temporary directory.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/fs_connector.py
+++ b/lmcache/v1/storage_backend/connector/fs_connector.py
@@ -71,16 +71,35 @@ class FSConnector(RemoteConnector):
             else config.get_extra_config_value("fs_connector_read_ahead_size", None)
         )
 
+        self.path_fanout_num = (
+            None
+            if config is None
+            else config.get_extra_config_value("fs_connector_path_fanout_num", None)
+        )
+
         logger.info(
             f"Initialized FSConnector with base paths {self.base_paths}, "
             f"relative tmp dir: {self.relative_tmp_dir}, "
-            f"read ahead size: {self.read_ahead_size}"
+            f"read ahead size: {self.read_ahead_size}, "
+            f"path fanout num: {self.path_fanout_num}"
         )
         # Create directories for all paths
         for path in self.base_paths:
             path.mkdir(parents=True, exist_ok=True)
             if self.relative_tmp_dir is not None:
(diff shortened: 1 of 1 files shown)


maobaolong:
@chunxiaozheng Would you like to take a look?

lmcache/v1/storage_backend/connector/fs_connector.py
```diff
@@ -33,24 +33,21 @@ class FSConnector(RemoteConnector):
...
             config: Lmcache engine config
         """
-        # Parse comma separated paths
-        self.base_paths = (
-            [Path(p.strip()) for p in base_paths_str.split(",")]
-            if "," in base_paths_str
-            else [Path(base_paths_str)]
-        )
+        # group_num is a power of 2
+        self.group_num = 256
```
chunxiaozheng: Could we modify this param to configurable?

maobaolong:
@hb-lee Thanks for this contribution, it really helps for most of FileSystem, but it would be nice if you:
1. Do not break the previous multiple base_path feature, that feature used for multi-mountpoints case.
2. To be configurable for the layout, the existing layout can be default one to keep backward-compatible, and user can define 256, 16x16, or other layout to adapt their different scenarios.

lmcache/v1/storage_backend/connector/fs_connector.py
hb-lee: ok, thanks for reviewing. 

And as for the tmp_path, what do you think about following the code assistant's recommendations?
`
        if self.relative_tmp_dir is not None:
            tmp_path = base_path / self.relative_tmp_dir / file_name
        else:
            tmp_path = file_path.with_suffix(".tmp")
        return file_path, tmp_path
`
If we want to delete the tmp files,  we just traverse the directory tree and delete them.

maobaolong:
Hey @hb-lee , Do you still work on this PR?

PR #2007, opened 2025-11-16
[Edge Case]: Preemption Loading

### LINKED ISSUE #1969: [Bug] cache is not retrieved when rescheduling a preempted request in vLLM
**To Reproduce**
Test script to reproduce this issue is attached at the bottom.
To run this script, activate the venv of vLLM + LMCache and run with cpu backend by default:
`PYTHONHASHSEED=0 VLLM_LOGGING_LEVEL=DEBUG LMCACHE_LOG_LEVEL=DEBUG pytest test_engine_lmcache_preemption.py -k preemption -s`
You can also run this script with other backends by setting LMCACHE_CONFIG_FILE.

**How it works**
The script simply uses a small model - `Qwen/Qwen3-0.6B-FP8` to generate response. `BLOCK_SIZE` is set to 16, aligning vllm's block size and lmcache's chunk size. `PROMPT_TOKEN_COUNT` is set to 2 BLOCKs plus 2 = 34 tokens, which essentially needs 3 blocks. `MAX_NEW_TOKENS` is set to 2 BLOCKs plus 3 = 35 tokens. All these parameters could be tuned in your environment.

There will be 2 requests running at the same time, so `NUM_BLOCKS` is set to 7, which limited vLLM's prefix cache size and guarantee the preemption. Set `NUM_BLOCKS` to greather than 10, no preemption will happen.

I ran this script with additional trace in vLLM and LMCache, found the bug happens as following:
* request 0: allocating blocks [1,2,3] to store prompt token cache around line 510 in vllm/schedule.py
```python
new_blocks = self.kv_cache_manager.allocate_slots()`
```
* request 1: allocating blocks [4,5,6] to store prompt token cache too
* data computation and stored to [1,2,3] and [4,5,6], respectively
* request 0 is generating more tokens and related cache data, filling up block pool
* Because total cache size is 7, so preemption starts
* request 1 is preempted. Log shows that blocks [4,5,6] are freed, ref_cnt is down to 0 but **data remains**
* request 0 keeps being scheduled, block [6] is allocated for request 0 to store cache, around line 263 in vllm/scheduler.py
* request 0's prompt and new tokens reaches 64, then block [5] is allocated for request 0
* request 0 finished, and free blocks [1,2,3,6,5]
* Now blocks[6] and [5] are filled with request 0's data
* request 1 is rescheduled, and at line 411 in scheduler.py, block [4] is found to be reused because block hash matches
```python
self.kv_cache_manager.get_computed_blocks(request)
```
At this point, LMCache is able to calculate the number of tokens to be load
```
LMCache INFO: Reqid: 1, Total tokens 49, LMCache hit tokens: 34, need to load: 18
```
Message shows that, vllm has req 1's first 16 tokens in block [4], althought the block is freed
LMCache has the entire cache of 34 computed tokens, first 16 are the same as block [4] in vllm,
rest 18 tokens have to be loaded from LMCache to GPU
* request 1 allocate blocks [5,6,3] for future token generation
At this point, cache data of the 18 tokens in LMCache are expected to be loaded to block [5] and [6], but it never happens
* At line 650 in vllm/scheduler.py, `build_connector_meta()` is called
```python
meta = self.connector.build_connector_meta(scheduler_output)
```
* However, in LMCache code, line 1509 is hit in build_connector_meta()'s cached_reqs code branch, with `load_spec=None`
```python
                request_tracker,
                self._block_size,
                self._lmcache_chunk_size,
                load_spec=None,
                discard_partial_chunks=self._discard_partial_chunks,
                save_decode_cache=self._save_decode_cache,
            )
```
* Retrieval is NOT scheduled because `load_spec=None` here. Blocks [5] and [6] are NOT filled with expected data from LMCache, and blocks [5,6] still contain stale data from request 0
* vllm then generate faulty response for request 1.

**Example of prompt and expected response**
The script is also attached at the bottom. Run the example, it shows the correct response of request 1
```
$ PYTHONHASHSEED=0 VLLM_LOGGING_LEVEL=DEBUG python3 ./example.py --gpu-memory-utilization=0.1 --enforce-eager --no-enable-prefix-caching --max_model_len=400
...
Request 1
  Prompt token IDs: [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2]
  Candidate 0 token IDs: [14582, 25, 16246, 264, 25, 715, 2, 9645, 264, 25, 715, 2, 256, 264, 729, 429, 4990, 264, 264, 1140, 315, 25780, 323, 4675, 279, 3082, 315, 279, 264, 21495, 448, 279, 2701, 4682, 510]
```
Run the test script with `NUM_BLOCKS = 7` to force preemption, the WRONG response is like:
```
$ PYTHONHASHSEED=0 VLLM_LOGGING_LEVEL=DEBUG LMCACHE_LOG_LEVEL=DEBUG LMCACHE_TRACK_USAGE=false pytest test_engine_lmcache_preemption.py -k preemption -s
...
FINISHED REQ 1: [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2]
FINISHED REQ 1: [14582, 25, 16246, 264, 25, 715, 2, 9645, 264, 25, 715, 2, 256, 264, 729, 311, 1477, 279, 279, 4226, 311, 279, 3405, 3118, 389, 279, 2661, 1946, 198, 2, 323, 279, 3405, 30, 715]
```

Then run the test script, with `NUM_BLOCKS = 11` to AVOID preemption, the response is like:
```
FINISHED REQ 1: [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2]
FINISHED REQ 1: [14582, 25, 16246, 264, 25, 715, 2, 9645, 264, 25, 715, 2, 256, 264, 729, 429, 4990, 264, 264, 1140, 315, 25780, 323, 4675, 279, 3082, 315, 279, 264, 21495, 448, 279, 2701, 4682, 510]
```
which is the same as result of example

I am not sure if this is exactly the issue mentiond in PR #1361 . But I found that the changes in PR #1361 fixes this issue by setting load_spec correctly.

**The Script**
Test script that force preemption
```python
# test_engine_lmcache_preemption.py 
import time
from typing import Iterable

import pytest

from vllm.config import KVTransferConfig
from vllm.engine.arg_utils import EngineArgs
from vllm.platforms import current_platform
from vllm.sampling_params import SamplingParams
from vllm.utils.torch_utils import set_default_torch_num_threads
from vllm.v1.engine import EngineCoreEventType, EngineCoreRequest
from vllm.v1.engine.core import EngineCore
from vllm.v1.executor.abstract import Executor

MODEL_NAME = "Qwen/Qwen3-0.6B-FP8"
BLOCK_SIZE = 16
NUM_BLOCKS = 7
PROMPT_TOKEN_COUNT = BLOCK_SIZE * 2 + 2
MAX_NEW_TOKENS = BLOCK_SIZE * 2 + 3

pytestmark = pytest.mark.skipif(  # type: ignore[var-annotated]
    not current_platform.is_cuda(),
    reason="LMCache preemption test requires CUDA support.",
)

def _iter_events(outputs: Iterable):
    for output in outputs:
        if not output.events:
            continue
        for event in output.events:
            yield event


def test_scheduler_handles_preemption_with_lmcache(monkeypatch: pytest.MonkeyPatch):
    """Launch a minimal vLLM engine with LMCache and trigger preemption.

    The test configures the engine to use the LMCache connector via the
    ``--kv-transfer-config`` equivalent API and constrains the number of GPU
    KV blocks. When multiple requests are executed concurrently the scheduler
    must preempt one of them in order to allocate a fresh block. We assert that
    a PREEMPTED event is observed in the engine outputs, ensuring that the
    LMCache-backed scheduler path handles preemption correctly.
    """

    pytest.importorskip("lmcache")

    from lmcache.integration.vllm.utils import ENGINE_NAME
    from lmcache.v1.cache_engine import LMCacheEngineBuilder

    env_vars = {
        "LMCACHE_USE_EXPERIMENTAL": "True",
        "LMCACHE_CHUNK_SIZE": str(BLOCK_SIZE),
        "LMCACHE_LOCAL_CPU": "True",
        "LMCACHE_MAX_LOCAL_CPU_SIZE": "1.0",
    }
    for key, value in env_vars.items():
        monkeypatch.setenv(key, value)

    kv_transfer_config = KVTransferConfig(
        kv_connector="LMCacheConnectorV1",
        kv_role="kv_both",
    )

    engine_args = EngineArgs(
        model=MODEL_NAME,
        max_model_len=BLOCK_SIZE * NUM_BLOCKS,
        block_size=BLOCK_SIZE,
        num_gpu_blocks_override=NUM_BLOCKS,
        kv_transfer_config=kv_transfer_config,
        enforce_eager=True,
        gpu_memory_utilization=0.1,
    )

    vllm_config = engine_args.create_engine_config()
    executor_cls = Executor.get_class(vllm_config)

    engine_core: EngineCore | None = None
    with set_default_torch_num_threads(1):
        engine_core = EngineCore(
            vllm_config=vllm_config,
            executor_class=executor_cls,
            log_stats=True,
        )

    def make_request(idx: int) -> EngineCoreRequest:
        prompt_tokens = [idx + 1] * PROMPT_TOKEN_COUNT
        sampling_params = SamplingParams(
            temperature=0.0,
            max_tokens=MAX_NEW_TOKENS,
        )
        return EngineCoreRequest(
            request_id=str(idx),
            prompt_token_ids=prompt_tokens,
            mm_features=None,
            sampling_params=sampling_params,
            pooling_params=None,
            eos_token_id=None,
            arrival_time=time.time(),
            lora_request=None,
            cache_salt=None,
            data_parallel_rank=None,
        )

    requests = [make_request(0), make_request(1)]

    prompt_token_ids = [requests[0].prompt_token_ids, requests[1].prompt_token_ids]
    output_token_ids = [[], []]

    try:
        for req in requests:
            engine_core.add_request(*engine_core.preprocess_add_request(req))

        seen_preemption = False
        finished = set()

        for _ in range(128):
            outputs_by_engine, model_executed = engine_core.step()

            if not outputs_by_engine:
                print("No outputs yet.")
                break

            engine_outputs = outputs_by_engine[0]

            for out in engine_outputs.outputs:
                output_token_ids[int(out.request_id)].extend(out.new_token_ids)
                if out.finished:
                    print(f"FINISHED REQ {out.request_id}: {prompt_token_ids[int(out.request_id)]}")
                    print(f"FINISHED REQ {out.request_id}: {output_token_ids[int(out.request_id)]}")
                    finished.add(out.request_id)

            for event in _iter_events(engine_outputs.outputs):
                print(f"Event: {event}")
                if event.type is EngineCoreEventType.PREEMPTED:
                    seen_preemption = True

            engine_core.post_step(model_executed)

            if len(finished) == len(requests):
                print("All requests finished.")
                break

        assert finished == {req.request_id for req in requests}
        assert seen_preemption, "Expected at least one preemption event."\
            " Check KV block configuration if this fails."

    finally:
        if engine_core is not None:
            engine_core.shutdown()
        LMCacheEngineBuilder.destroy(ENGINE_NAME)
```
and the example of promt and expected repsonse
run it as
```
PYTHONHASHSEED=0 VLLM_LOGGING_LEVEL=DEBUG python3 ./example.py --gpu-memory-utilization=0.1 --enforce-eager --no-enable-prefix-caching --max_model_len=400
```
script:
```python
# Example.py

from __future__ import annotations
from typing import Any
from vllm import EngineArgs, LLM
from vllm.inputs import TokensPrompt
from vllm.sampling_params import SamplingParams
from vllm.utils.argparse_utils import FlexibleArgumentParser

MODEL_NAME = "Qwen/Qwen3-0.6B-FP8"
BLOCK_SIZE = 16
PROMPT_TOKEN_COUNT = BLOCK_SIZE * 2 + 2
MAX_NEW_TOKENS = BLOCK_SIZE * 2 + 3


def create_parser() -> FlexibleArgumentParser:
    parser = FlexibleArgumentParser()
    EngineArgs.add_cli_args(parser)
    parser.set_defaults(model=MODEL_NAME)
    return parser


def build_prompts() -> list[TokensPrompt]:
    return [
        TokensPrompt(prompt_token_ids=[idx + 1] * PROMPT_TOKEN_COUNT)
        for idx in range(2)
    ]


def main(args: dict[str, Any]) -> None:
    llm = LLM(**args)

    sampling_params = SamplingParams(
        temperature=0.0,
        max_tokens=MAX_NEW_TOKENS,
    )

    prompts = build_prompts()
    outputs = llm.generate(prompts, sampling_params)

    print("-" * 80)
    for request_idx, output in enumerate(outputs):
        print(f"Request {request_idx}")
        print(f"  Prompt token IDs: {output.prompt_token_ids}")

        for candidate_idx, candidate in enumerate(output.outputs):
            print(
                f"  Candidate {candidate_idx} token IDs: {candidate.token_ids}"
            )
            print(f"  Candidate {candidate_idx} text: {candidate.text!r}")
        print("-" * 80)


if __name__ == "__main__":
    parser = create_parser()
    parsed_args: dict[str, Any] = vars(parser.parse_args())
    main(parsed_args)
```

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1969 https://github.com/LMCache/LMCache/pull/1361

This is the script I used to test preemption minimal case.

```python
"""
Intricacy of the vLLM KV Cache Management: 
- Usually 3 blocks cannot be used
- You should generally add 1 to your (# input tokens) + (# output tokens)

Server deployment: 
LMCACHE_CHUNK_SIZE=8 \
vllm serve meta-llama/Llama-3.1-8B-Instruct --num-gpu-blocks 10 --block-size 16 \
  --no-enable-prefix-caching \
  --kv-transfer-config \
    '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}' \
    > debug.log 2>&1

vllm serve meta-llama/Llama-3.1-8B-Instruct --num-gpu-blocks 10 --block-size 16 \
  --no-enable-prefix-caching \
    > debug.log 2>&1

We want both requests to saturate all the available blocks. This will force request 0 to to preempt request 1.
"""

import asyncio
from openai import AsyncOpenAI

client = AsyncOpenAI(
    base_url="http://localhost:8000/v1",
    api_key="none",
)

async def run_request(text: str):
    num_gpu_blocks = 10
    # n = num_gpu_blocks - 3
    # 5 + 16 * (n - 1)
    max_output_tokens = 5 + 16 * ((num_gpu_blocks - 3) - 1)
    resp = await client.chat.completions.create(
        model="meta-llama/Llama-3.1-8B-Instruct",
        messages=[{"role": "user", "content": text}],
        max_tokens=max_output_tokens,
    )
    return resp.choices[0].message.content

async def main():
    """
    "1Tell me a long story that never ends" and 
        "1Tell me a long story that never ends" 
        "2Tell me a long story that never ends" 
        are 10 tokens
    """
    prompts = [
        "1Tell me a long story that never ends",
        "2Tell me a long story that never ends",
    ]

    # run both requests at the same time
    results = await asyncio.gather(*(run_request(p) for p in prompts))

    for i, r in enumerate(results, 1):
        print(r)

if __name__ == "__main__":
    asyncio.run(main())
```

Logs before the fix (we see the tokens but don't load them): 
```text
(EngineCore_DP0 pid=3322770) [2025-11-18 19:21:06,890] LMCache INFO: Reqid: chatcmpl-75ed50a6af244d9981b15189a425de65, Total tokens 65, LMCache hit tokens: 44, need to load: 44 (vllm_v1_adapter.py:1330:lmcache.integration.vllm.vllm_v1_adapter)
(APIServer pid=3322482) INFO:     127.0.0.1:51520 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(APIServer pid=3322482) INFO:     127.0.0.1:51528 - "POST /v1/chat/completions HTTP/1.1" 200 OK
```

Logs after the fix: 
```text
[1;36m(EngineCore_DP0 pid=1538843)[0;0m INFO 11-17 02:55:14 [scheduler.py:428] Attempting to recover preempted request chatcmpl-0da54347fb1a4053b1b373a33ce88915
[1;36m(EngineCore_DP0 pid=1538843)[0;0m [32;20m[2025-11-17 02:55:14,985] LMCache INFO:[0m Reqid: chatcmpl-0da54347fb1a4053b1b373a33ce88915, Total tokens 65, LMCache hit tokens: 40, need to load: 40 [3m(vllm_v1_adapter.py:1354:lmcache.integration.vllm.vllm_v1_adapter)[0m
[1;36m(EngineCore_DP0 pid=1708755)[0;0m [32;20m[2025-11-17 07:23:12,194] LMCache INFO:[0m Request chatcmpl-b492c116d42a45e2bf5b8df13c501949 lmcache cached tokens: 40, vllm cached tokens: 0, can load: True [3m(vllm_v1_adapter.py:1579:lmcache.integration.vllm.vllm_v1_adapter)[0m
[1;36m(EngineCore_DP0 pid=1708755)[0;0m [32;20m[2025-11-17 07:23:12,195] LMCache INFO:[0m Retrieved 40 out of 40 required tokens (from 40 total tokens). size: 0.0049 gb, cost 0.4515 ms, throughput: 10.8152 GB/s; [3m(cache_engine.py:560:lmcache.v1.cache_engine)[0m

```


lmcache/v1/lookup_client/abstract_client.py
```diff
@@ -14,6 +14,15 @@
 class LookupClientInterface(metaclass=abc.ABCMeta):
     """Abstract interface for lookup clients."""
 
+    def lookup_cache(self, lookup_id: str) -> Optional[int]:
```
sammshen, line 17: the default does nothing

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1396,22 +1420,29 @@ def update_state_after_alloc(self, request: "Request", num_external_tokens: int)
...
-                num_external_tokens > 0
-                and num_external_tokens
-                == self.load_specs[request.request_id].lmcache_cached_tokens
-                - self.load_specs[request.request_id].vllm_cached_tokens
-            ), (
-                f"Mismatch in number of tokens: {num_external_tokens} vs "
-                f"{self.load_specs[request.request_id].lmcache_cached_tokens} - "
-                f"{self.load_specs[request.request_id].vllm_cached_tokens}"
-                f" for request {request.request_id}"
+        recalc_last = (
```
sammshen, line 1495: the logic is the same. we also check for prompt hit case with the `recalc_last`

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1437,8 +1468,12 @@ def build_connector_meta(
             self._request_trackers.pop(finished_req_id, None)
             self._unfinished_requests.pop(finished_req_id, None)
 
+        # We should load KV for:
```
sammshen, line 1543: updated comment

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -184,7 +184,7 @@ def from_new_request(
             skip_save (bool): whether the request cache should be saved
         """
         # vLLM 0.9.0 update: request.block_ids changed from list[int] to
-        # list[list[int]]
+        # tuple[list[int]]
```
sammshen, line 189: it is tuple now

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -246,7 +246,16 @@ def update(
...
         else:
             raise ValueError(f"Unsupported new_block_ids type {type(new_block_ids)}")
-        self.allocated_block_ids.extend(new_block_ids)
+
+        if preempted:
+            # the block ids will change after preemption
+            self.allocated_block_ids = new_block_ids
+            # reset the number of saved tokens
+            self.num_saved_tokens = lmcache_cached_tokens
+            # we don't need to extend the token ids in the preempted case
```
sammshen, line 257: because there is no previous step where we generated an extra token since the running request was preempted

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1280,14 +1289,29 @@ def get_num_new_matched_tokens(
             the number of tokens that can be loaded from the
             external KV cache beyond what is already computed.
         """
+        # to handle preempted requests, we want `get_num_new_matched_tokens` to be
```
sammshen, line 1366: this is the main logic for preemption

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1498,8 +1540,9 @@ def build_connector_meta(
         for i, req_id in enumerate(cached_reqs.req_ids):
             request_tracker = self._request_trackers[req_id]
             num_new_tokens = scheduler_output.num_scheduled_tokens[req_id]
+            # TODO: this is a dangerous reference to the request object inside vllm
             if request := self._unfinished_requests.get(req_id):
-                num_current_tokens = len(request_tracker.token_ids)
+                num_current_tokens = request.num_computed_tokens
```
sammshen, line 1619: if we use `len(request_tracker.token_ids)`, this is a serious correctness bug

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1498,8 +1540,9 @@ def build_connector_meta(
         for i, req_id in enumerate(cached_reqs.req_ids):
             request_tracker = self._request_trackers[req_id]
             num_new_tokens = scheduler_output.num_scheduled_tokens[req_id]
+            # TODO: this is a dangerous reference to the request object inside vllm
```
sammshen, line 1617: we can think of a better way to handle this in the future

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -303,7 +312,7 @@ def from_request_tracker(
         input_token_len = len(input_token_ids)
 
         is_last_prefill = False
-        if input_token_len == tracker.prompt_len:
+        if input_token_len >= tracker.prompt_len:
```
yoo-kumaneko, line 317: Is this a related bug or a separate one?

sammshen: this is the same bug. the input tokens may be larger than the prompt_len when a preempted request is recovered.

yoo-kumaneko (approved):
LGTM

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1280,14 +1289,29 @@ def get_num_new_matched_tokens(
...
         if self.kv_role == "kv_producer" and not hasattr(
             self.lookup_client, "supports_producer_reuse"
         ):
             return 0
 
-        self._requests_priority[request.request_id] = getattr(request, "priority", 0)
+        req_id = request.request_id
+
+        # consult the cache before any processing
+        if cached_num_hit_toks := self.lookup_client.lookup_cache(lookup_id=req_id):
```
maobaolong, line 1381: If we config to initialize a HitLimitLookupClient instance, it would be wrong.
Because the two above did not implement the new api you added. @sammshen

maobaolong: Maybe https://github.com/LMCache/LMCache/pull/2021 this can help to make sure we have implemented all the method without forget any one.

maobaolong: @sammshen You can add the following code to the tail of `lmcache/v1/lookup_client/hit_limit_lookup_client.py`

```python
    def lookup_cache(self, lookup_id: str) -> Optional[int]:
        return self.actual_lookup_client.lookup_cache(lookup_id)
```

sammshen: the default implementation will cover it right? the default implementation is to return None in `lmcache/v1/lookup_client/abstract_client.py`

sammshen:
@maobaolong @chunxiaozheng feel free to take a look again

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: thanks @maobaolong I did not read the hit limit client carefully enough before, but updated now!

maobaolong:
@sammshen Sorry, as `ChunkStatisticsLookupClient` merged to dev just today, so you have to implement the following method into it too.
```python
def lookup_cache(self, lookup_id: str) -> Optional[int]:
    return self.actual_lookup_client.lookup_cache(lookup_id)
```

maobaolong:
@sammshen I pressed the `update branch` button, so the UT check will failed.

sammshen:
The UT is very useful! updated @maobaolong

ApostaC (approved):
LGTM as we discussed offline.

maobaolong (approved):
LGTM

ApostaC:
@sammshen Do we want to create a special comprehensive test case if we know how to reproduce the preemption case?

maobaolong:
> @sammshen Do we want to create a special comprehensive test case if we know how to reproduce the preemption case?

This is a great idea!

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: why

PR #2010, opened 2025-11-17
[controller] support health check for lmcache worker

chunxiaozheng:
support heath check of lmcache worker


lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -206,6 +218,29 @@ async def handle_batched_req_request(self, socket) -> Optional[MsgBase]:
             except Exception as e:
                 logger.error(f"Controller Manager error: {e}")
 
+    async def health_check(self):
+        while True:
+            await asyncio.sleep(self.health_check_interval)
```
sammshen: should this be asyncio.sleep? this is a seaparte thread right?

sammshen (approved):
LGTM, just one comment

lmcache/v1/cache_controller/controller_manager.py
chunxiaozheng: > should this be asyncio.sleep? this is a seaparte thread right?

@sammshen Yeah, this is a separate thread, I have changed to sync sleep.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -206,6 +218,29 @@ async def handle_batched_req_request(self, socket) -> Optional[MsgBase]:
...
+        while True:
+            time.sleep(self.health_check_interval)
+            worker_infos = list(self.reg_controller.worker_info_mapping.values())
+            for worker_info in worker_infos:
+                if (
+                    time.time() - worker_info.last_heartbeat_time
+                    > self.lmcache_worker_timeout
+                ):
+                    logger.warning(
+                        f"Worker {worker_info.instance_id}_{worker_info.worker_id} last"
```
maobaolong: f-string is deprecated, please change to use formatted string. @chunxiaozheng

chunxiaozheng: @maobaolong Thanks for your review! I have updated.

maobaolong (approved):
Thanks for this great feature! LGTM

PR #2015, opened 2025-11-17
[Hot fix] Fix LMCacheConnector rank to use local GPU index instead of tp_rank

### LINKED ISSUE #1988: [Bug] When using lmcache with sglang in a two-node deployment, the LMCacheConnector on the second node incorrectly uses the rank value.
The second node uses GPU 3 and GPU 4：

```
device=device(type='cuda', index=3)
device=device(type='cuda', index=2) File "/sgl-workspace/sglang/LMCache/lmcache/integration/sglang/sglang_adapter.py", line 275, in start_load_kv
retrieve_token_num = self.global_min_tokens(
^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
return func(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^
File "/sgl-workspace/sglang/LMCache/lmcache/integration/sglang/sglang_adapter.py", line 233, in global_min_tokens
t = torch.tensor([local_tokens], dtype=torch.int32, device=device)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid device ordinal
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with TORCH_USE_CUDA_DSA to enable device-side assertions.
```

```
class LMCacheConnector:
    def __init__(
        self,
        sgl_config: ModelConfig,
        tp_size: int,
        rank: int,
        k_pool: List[torch.Tensor],
        v_pool: List[torch.Tensor],
    ):
        kv_dtype = k_pool[0].dtype
        self.lmcache_engine = init_lmcache_engine(
            sgl_config,
            tp_size,
            rank,
            kv_dtype,
        )
        self.sgl_config = sgl_config
        self.tp_size = tp_size
        self.rank = rank
        self.kvcaches = k_pool + v_pool
        self.num_layer = sgl_config.num_hidden_layers
```

In this case, the `rank `should use the `local rank` instead of the `tp_rank`.

DongDongJu:
Fix https://github.com/LMCache/LMCache/issues/1988.

use the local device index from k_pool[0].device.index when initializing LMCacheConnector instead of the tensor-parallel rank.
This prevents error on multi-node setups where tp_rank does not match the local GPU index.


sammshen:
So are you saying that in a multi-node setup, SGLang will pass in per-node rank instead of global rank?

DongDongJu:
> So are you saying that in a multi-node setup, SGLang will pass in per-node rank instead of global rank?

No. sglang sent the global rank(tp_rank) [reference link](https://github.com/sgl-project/sglang/blob/2bc7c5ebef9687d70ec2a05a148d52cecdcff314/python/sglang/srt/mem_cache/storage/lmcache/lmc_radix_cache.py#L62) and thats the reason why this patch finding local_rank for remote node.

sammshen:
Thanks, I see the call in https://github.com/sgl-project/sglang/blob/2bc7c5ebef9687d70ec2a05a148d52cecdcff314/python/sglang/srt/managers/scheduler.py

I see that it is important to update `self.rank` becvause `self.global_min_tokens()` will call `torch.device()`

Why do we also need update the `init_lmcache_engine()`? should we not use the global tp rank for that metadata?

DongDongJu:
> Thanks, I see the call in https://github.com/sgl-project/sglang/blob/2bc7c5ebef9687d70ec2a05a148d52cecdcff314/python/sglang/srt/managers/scheduler.py
> 
> I see that it is important to update `self.rank` becvause `self.global_min_tokens()` will call `torch.device()`
> 
> Why do we also need update the `init_lmcache_engine()`? should we not use the global tp rank for that metadata?

You are right. we still need tp_rank. Let me revise the patches.  Thanks @sammshen

lmcache/integration/sglang/sglang_adapter.py
```diff
@@ -129,16 +129,24 @@ def __init__(
...
+            local_rank = k_pool[0].device.index
+        else:
+            # Fallback for CPU / odd cases
+            local_rank = rank
+
         self.lmcache_engine = init_lmcache_engine(
             sgl_config,
             tp_size,
-            rank,
+            local_rank,
```
sammshen, line 155: the conclusion was to change back to global `rank` for the metadata right

DongDongJu: I applied our discussion by https://github.com/LMCache/LMCache/pull/2015/commits/a9c580eb068e6a8d3f5fbd4f69db6342cce2188e

Oasis-Git:
Does this pr introduce the modification of SGLang side implementation?

DongDongJu:
> Does this pr introduce the modification of SGLang side implementation?

Nope, sglang side [currently send the global rank](https://github.com/sgl-project/sglang/blob/2bc7c5ebef9687d70ec2a05a148d52cecdcff314/python/sglang/srt/mem_cache/storage/lmcache/lmc_radix_cache.py#L62). We can handle finding local rank in lmcache side.

Oasis-Git:
LGTM. Let's wait the CI pass and merge it.

sammshen (approved):

Oasis-Git (approved):

PR #2018, opened 2025-11-18
Refactor Async Lookup Client Message

yoo-kumaneko:
Let's use structured dataclasses instead of a raw list to pass the messages.

### DIFF
--- a/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
@@ -191,27 +191,31 @@ class LMCacheAsyncLookupClient(LookupClientInterface):
 
     def process_responses_from_workers(self):
         while self.running:
-            msg_buf = self.pull_socket.recv(copy=False)
-            # Deserialize message using msgspec
-            msg = msgspec.msgpack.decode(msg_buf, type=LookupResponseMsg)
-            lookup_id = msg.lookup_id
-            res = msg.num_hit_tokens
+            try:
+                msg_buf = self.pull_socket.recv(copy=False)
+                # Deserialize message using msgspec
+                msg = msgspec.msgpack.decode(msg_buf, type=LookupResponseMsg)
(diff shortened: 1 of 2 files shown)


yoo-kumaneko:
@ApostaC Hey, this is one of the split PR that refactors the messages in the async lookup client.

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -280,46 +281,40 @@ def __init__(self, lmcache_engine: LMCacheEngine, vllm_config: "VllmConfig"):
...
-                    lookup_id=lookup_id,
-                    hashes=hashes,
-                    offsets=offsets,
-                    pin=True,
-                    request_configs=request_configs,
+            try:
+                msg_buf = self.pull_socket.recv(copy=False)
+                msg = msgspec.msgpack.decode(
+                    msg_buf,
+                    type=Union[LookupRequestMsg],
```
ApostaC: Could `Union` be removed if we only have `LookupRequestMsg` here?

ApostaC (approved):
Otherwise LGTM!

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: Great catch. I did this because I plan to add another message type. Yeah, let's remove the union for now.

maobaolong (approved):
LGMT, it looks better!

yoo-kumaneko:
The comprehensive test failed.
ValueError: Free memory on device (11.71/139.8 GiB) on startup is less than desired GPU memory utilization (0.9, 125.82 GiB). Decrease GPU memory utilization or reduce GPU memory used by other processes.

ApostaC:
@yoo-kumaneko We also noticed the problem with the CI machine, and this should be fixed now
Please ping @Shaoting-Feng if you see similar issue again

PR #2022, opened 2025-11-18
auto align paged memory allocator size

chunxiaozheng:
The unit of `max_local_cpu_size` is `GB`, when transfer to bytes, it is difficult to be divided by `align_bytes` in `PagedTensorMemoryAllocator`, which will cause the following error:
Clipboard_Screenshot_1763468036


chunxiaozheng:
@sammshen @maobaolong could you help take a look?

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -350,8 +350,16 @@ def initialize_allocator(
                 ]
             )
             paged_mem_allocator = PagedCpuGpuMemoryAllocator()
+            chunk_size_bytes = (
```
maobaolong: You can abstract this to a method and place to `utils.py` to reduce redundant code, AFAIK, there is a same calculation in the `RemoteBackend`.

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -350,8 +350,16 @@ def initialize_allocator(
...
+                new_shape.numel()
+                * torch.tensor([], dtype=metadata.kv_dtype).element_size()
+            )
+            cpu_size_bytes = (
+                int(cpu_size * 1024**3) // chunk_size_bytes * chunk_size_bytes
+            )
+            logger.info(f"Align CPU size bytes: {cpu_size_bytes}")
```
maobaolong: We can log the size before aligned also.

BTW, we deprecated to use f-string already

maobaolong (approved):
@chunxiaozheng Create improvement, it LGTM overall, left a couple of nit comments inline.

lmcache/v1/storage_backend/local_cpu_backend.py
chunxiaozheng: done

lmcache/integration/vllm/utils.py
```diff
@@ -218,3 +218,7 @@ def extract_mm_features(
             return (request.mm_hashes, request.mm_positions)
     else:
         return ([], [])
+
+
+def get_chunk_size_bytes(shape: torch.Size, kv_dtype: torch.dtype):
```
sammshen: nit: this has nothing to do with chunk

sammshen (approved):
This is a great helper! just a nit about naming because the helper actually has nothing to do with chunk

PR #2024, opened 2025-11-18
[Core] Introduce NixlStorageBackend dynamic mode

tshmilnvidia:
This commit incorporates PR [#1939](https://github.com/LMCache/LMCache/pull/1939) into NixlStorageBackend, under the name of NixlDynamicStorageBackend, in order to prevent having 2 nixl storage backends as well as lessen code duplication.
    
In order to use this new dynamic storage mode, which creates object keys when needed instead of a limited pre-allocated amount, extra_config.nixl_pool_size should be set to 0.
    
This mode is currently only supported for nixl OBJ backend.


lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -289,59 +376,82 @@ def init_storage_handlers_object(self, page_size, keys):
...
-        return self.nixl_agent.make_prepped_xfer(
-            "READ",
-            self.mem_xfer_handler,
-            mem_indices,
-            self.storage_xfer_handler,
-            storage_indices,
+    def get_storage_to_mem_handle(self, mem_indices, storage_indices) -> NixlXferHandle:  # type: ignore[override]
+        return super().get_storage_to_mem_handle(
+            mem_indices, self.storage_xfer_handler, storage_indices
         )
```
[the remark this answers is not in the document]
tshmilnvidia: This is on purpose

tests/v1/data/nixl-dynamic.yaml
```diff
@@ -0,0 +1,20 @@
+chunk_size: 256
+local_cpu: False
+save_unfull_chunk: False
+enable_async_loading: False # set to True to test async loading
+# buffer size has to be divisible by chunk size
+# 2880MiB is divisible by 256 token chunk for Qwen3-4B/8B/32B
+nixl_buffer_size: 3019898880
+nixl_buffer_device: cpu
+extra_config:
+  nixl_backend: OBJ
```
ziruiliu: `enable_nixl_storage` has to be set true in this example

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -289,59 +376,82 @@ def init_storage_handlers_object(self, page_size, keys):
...
+            xfer_desc.append((0, page_size, i))
+
+        reg_descs = self.nixl_agent.register_memory(reg_list, mem_type="OBJ")
+        xfer_descs = self.nixl_agent.get_xfer_descs(xfer_desc, mem_type="OBJ")
+        xfer_handler = self.nixl_agent.prep_xfer_dlist(
+            self.agent_name, xfer_descs, mem_type="OBJ"
+        )
+        return reg_descs, xfer_handler
+
+    def post_blocking_async(self, handle: NixlXferHandle):
```
ziruiliu: This method name is vague (my fault in another PR), it should be simply post_async()

tests/v1/data/nixl-dynamic.yaml
tshmilnvidia: Correct. Also since I'm not using this file, I think this belongs better in the docs, WDYT?

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: Fixed

tests/v1/data/nixl-dynamic.yaml
ziruiliu: yes, we cann also update docs accordingly later

ziruiliu (approved):
LGTM now

jgoldsch12:
When using dynamic mode, mem_to_storage works fine.  I see the objects created with the serialized cache engine key names.  However, storage_to_mem instantly fails with the following:


(APIServer pid=575662) INFO:     127.0.0.1:36000 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(EngineCore_DP0 pid=575865) [2025-11-24 17:38:34,923] LMCache INFO: Reqid: chatcmpl-5805733a330149a68d81f49d3204606a, Total tokens 85, LMCache hit tokens: 85, need to load: 4 (vllm_v1_adapter.py:1426:lmcache.integration.vllm.vllm_v1_adapter)
(EngineCore_DP0 pid=575865) [2025-11-24 17:38:34,924] LMCache DEBUG: Scheduled to load 85 tokens for request chatcmpl-5805733a330149a68d81f49d3204606a (vllm_v1_adapter.py:403:lmcache.integration.vllm.vllm_v1_adapter)
(EngineCore_DP0 pid=575865) [2025-11-24 17:38:34,988] LMCache DEBUG: storage_to_mem for 1 objects size 58720256 took 0.062893 seconds (nixl_storage_backend.py:1013:lmcache.v1.storage_backend.nixl_storage_backend)
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [dump_input.py:72] Dumping input data for V1 LLM engine (v0.11.2) with config: model='meta-llama/Llama-3.2-3B-Instruct', speculative_config=None, tokenizer='meta-llama/Llama-3.2-3B-Instruct', skip_tokenizer_init=False, tokenizer_mode=auto, revision=None, tokenizer_revision=None, trust_remote_code=False, dtype=torch.bfloat16, max_seq_len=131000, download_dir=None, load_format=auto, tensor_parallel_size=1, pipeline_parallel_size=1, data_parallel_size=1, disable_custom_all_reduce=False, quantization=None, enforce_eager=False, kv_cache_dtype=auto, device_config=cuda, structured_outputs_config=StructuredOutputsConfig(backend='auto', disable_fallback=False, disable_any_whitespace=False, disable_additional_properties=False, reasoning_parser='', reasoning_parser_plugin='', enable_in_reasoning=False), observability_config=ObservabilityConfig(show_hidden_metrics_for_version=None, otlp_traces_endpoint=None, collect_detailed_traces=None), seed=0, served_model_name=meta-llama/Llama-3.2-3B-Instruct, enable_prefix_caching=True, enable_chunked_prefill=True, pooler_config=None, compilation_config={'level': None, 'mode': <CompilationMode.VLLM_COMPILE: 3>, 'debug_dump_path': None, 'cache_dir': '/home/jgoldsch/.cache/vllm/torch_compile_cache/fe9e360597', 'compile_cache_save_format': 'binary', 'backend': 'inductor', 'custom_ops': ['none'], 'splitting_ops': ['vllm::unified_attention', 'vllm::unified_attention_with_output', 'vllm::unified_mla_attention', 'vllm::unified_mla_attention_with_output', 'vllm::mamba_mixer2', 'vllm::mamba_mixer', 'vllm::short_conv', 'vllm::linear_attention', 'vllm::plamo2_mamba_mixer', 'vllm::gdn_attention_core', 'vllm::kda_attention', 'vllm::sparse_attn_indexer'], 'compile_mm_encoder': False, 'use_inductor': None, 'compile_sizes': [], 'inductor_compile_config': {'enable_auto_functionalized_v2': False, 'combo_kernels': True, 'benchmark_combo_kernel': True}, 'inductor_passes': {}, 'cudagraph_mode': <CUDAGraphMode.FULL_AND_PIECEWISE: (2, 1)>, 'cudagraph_num_of_warmups': 1, 'cudagraph_capture_sizes': [1, 2, 4, 8, 16, 24, 32, 40, 48, 56, 64, 72, 80, 88, 96, 104, 112, 120, 128, 136, 144, 152, 160, 168, 176, 184, 192, 200, 208, 216, 224, 232, 240, 248, 256, 272, 288, 304, 320, 336, 352, 368, 384, 400, 416, 432, 448, 464, 480, 496, 512], 'cudagraph_copy_inputs': False, 'cudagraph_specialize_lora': True, 'use_inductor_graph_partition': False, 'pass_config': {}, 'max_cudagraph_capture_size': 512, 'local_cache_dir': '/home/jgoldsch/.cache/vllm/torch_compile_cache/fe9e360597/rank_0_0/backbone'}, 
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [dump_input.py:79] Dumping scheduler output for model execution: SchedulerOutput(scheduled_new_reqs=[NewRequestData(req_id=chatcmpl-5805733a330149a68d81f49d3204606a,prompt_token_ids_len=85,mm_features=[],sampling_params=SamplingParams(n=1, presence_penalty=0.0, frequency_penalty=0.0, repetition_penalty=1.0, temperature=0.7, top_p=0.9, top_k=0, min_p=0.0, seed=None, stop=[], stop_token_ids=[128008, 128001], bad_words=[], include_stop_str_in_output=False, ignore_eos=False, max_tokens=1, min_tokens=0, logprobs=None, prompt_logprobs=None, skip_special_tokens=True, spaces_between_special_tokens=True, truncate_prompt_tokens=None, structured_outputs=None, extra_args=None),block_ids=([1, 2, 3, 4, 5, 18337],),num_computed_tokens=84,lora_request=None,prompt_embeds_shape=None)], scheduled_cached_reqs=CachedRequestData(req_ids=[], resumed_req_ids=[], new_token_ids=[], all_token_ids={}, new_block_ids=[], num_computed_tokens=[], num_output_tokens=[]), num_scheduled_tokens={chatcmpl-5805733a330149a68d81f49d3204606a: 1}, total_num_scheduled_tokens=1, scheduled_spec_decode_tokens={}, scheduled_encoder_inputs={}, num_common_prefix_blocks=[6], finished_req_ids=[], free_encoder_mm_hashes=[], pending_structured_output_tokens=false, kv_connector_metadata=LMCacheConnectorMetadata(requests=[ReqMeta(req_id='chatcmpl-5805733a330149a68d81f49d3204606a', token_ids=[128000, 128006, 9125, 128007, 271, 38766, 1303, 33025, 2696, 25, 6790, 220, 2366, 18, 198, 15724, 2696, 25, 220, 1187, 4723, 220, 2366, 20, 271, 128009, 128006, 882, 128007, 271, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 29474, 128009, 128006, 78191, 128007, 271], slot_mapping=Tensor(shape=torch.Size([85]), device=cpu,dtype=torch.int64), is_last_prefill=true, save_spec=SaveSpec(skip_leading_tokens=85, can_save=false), load_spec=LoadSpec(vllm_cached_tokens=80, lmcache_cached_tokens=85, can_load=true), disagg_spec=null, request_configs=null)]), ec_connector_metadata=null)
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [dump_input.py:81] Dumping scheduler stats: SchedulerStats(num_running_reqs=1, num_waiting_reqs=0, step_counter=0, current_wave=0, kv_cache_usage=0.00019550342130991716, prefix_cache_stats=PrefixCacheStats(reset=False, requests=1, queries=85, hits=80, preempted_requests=0, preempted_queries=0, preempted_hits=0), connector_prefix_cache_stats=PrefixCacheStats(reset=False, requests=1, queries=85, hits=4, preempted_requests=0, preempted_queries=0, preempted_hits=0), spec_decoding_stats=None, kv_connector_stats=None, waiting_lora_adapters={}, running_lora_adapters={})
.
.
.
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844]   File "/home/jgoldsch/LMCache-public/lmcache/v1/cache_engine.py", line 546, in retrieve
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844]     self.gpu_connector.batched_to_gpu(
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844]   File "/home/jgoldsch/LMCache-public/lmcache/v1/gpu_connector.py", line 316, in batched_to_gpu
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844]     self.load_stream.synchronize()
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844]   File "/home/jgoldsch/obs-dist/lib/python3.12/site-packages/torch/cuda/streams.py", line 102, in synchronize
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844]     super().synchronize()
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844] torch.AcceleratorError: CUDA error: an illegal memory access was encountered
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844] Search for `cudaErrorIllegalAddress' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844] CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844] For debugging consider passing CUDA_LAUNCH_BLOCKING=1
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844] Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
(EngineCore_DP0 pid=575865) ERROR 11-24 17:38:35 [core.py:844] 


This is my lmcache.yaml:

chunk_size: 512
local_cpu: false
nixl_buffer_size: 8573157376
nixl_buffer_device: cpu
extra_config:
  enable_nixl_storage: true
  nixl_backend: OBJ
  nixl_pool_size: 0
  nixl_presence_cache: False
  nixl_async_put: False
  nixl_backend_params:
    access_key:  <key>
    secret_key: <secret>
    bucket: <bucket>
    endpoint_override: http://<address>
    scheme: http
    use_virtual_addressing: "false"
    req_checksum: required


Same config, without dynamic mode (nixl_pool_size: 2000) works fine for both mem_to_storage and storage_to_mem

ziruiliu:
Hi @jgoldsch12 , from the first glance I think it could be nixl buffer address alignment issue. 
Does this issue always happen when dynamic is enabled ?

Please try this temp patch in cache_engine.py to verify if the buffer pointer is aligned?
```diff
 def _Create_memory_allocator(
...
-            buffer = torch.empty(
-                config.nixl_buffer_size,
+            orig_buf = torch.empty(
+                config.nixl_buffer_size + 4096,
                 dtype=torch.uint8,
                 device=corrected_device,
             )
+            orig_base = orig_buf.data_ptr()
+            aligned_base = (orig_base + (4096 - 1)) & ~(4096 - 1)
+            offset = aligned_base - orig_base
+            buffer = orig_buf.narrow(0, offset, config.nixl_buffer_size)
+            logger.debug(f"original buffer allocated at {orig_base:#018x} aligned to {aligned_base:#018x}")
 ```

**Comment by jgoldsch12** [comment]
> Hi @jgoldsch12 , from the first glance I think it could be nixl buffer address alignment issue. Does this issue always happen when dynamic is enabled ?
> 
> Please try this temp patch in cache_engine.py to verify if the buffer pointer is aligned?
> 
> ```diff
>  def _Create_memory_allocator(
> ...
> -            buffer = torch.empty(
> -                config.nixl_buffer_size,
> +            orig_buf = torch.empty(
> +                config.nixl_buffer_size + 4096,
>                  dtype=torch.uint8,
>                  device=corrected_device,
>              )
> +            orig_base = orig_buf.data_ptr()
> +            aligned_base = (orig_base + (4096 - 1)) & ~(4096 - 1)
> +            offset = aligned_base - orig_base
> +            buffer = orig_buf.narrow(0, offset, config.nixl_buffer_size)
> +            logger.debug(f"original buffer allocated at {orig_base:#018x} aligned to {aligned_base:#018x}")
> ```

Hi @ziruiliu, 

Yes this issue occurs each time dynamic mode is used.  The crash occurs also with the patch applied and only in the GET path.

**Comment by jgoldsch12** [comment]
Is it possible this PR is missing this change:

https://github.com/LMCache/LMCache/commit/2891a30a6050643a5e8441a30024adaf4c6c1e76

**Comment by vvenkates27** [comment]
@jgoldsch12  I believe the issue is with ref count leaks which is possibly causing this corruption
cc: @tshmilnvidia, @ziruiliu

File: lmcache/v1/storage_backend/nixl_storage_backend.py [hunk 4]
```diff
@@ -574,96 +842,427 @@ def close(self) -> None:
...
-    def get_memory_allocator(self):
-        return self.memory_allocator
+        handle = self.agent.get_mem_to_storage_handle(
+            mem_indices, storage_xfer_handler, storage_indices
+        )
 
-    def allocate(
+        if self.async_mode:
+            for mem_obj in mem_objs:
+                mem_obj.ref_count_up()
```
ziruiliu: It was my bad. The ref_count_up() should be moved to this async task is created, otherwise there is a chance that the mem_obj is freed before mem_to_storage() is actually executed.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -574,96 +842,427 @@ def close(self) -> None:
...
+        keys: Sequence[CacheEngineKey],
+        memory_objs: List[MemoryObj],
+        transfer_spec: Any = None,
+    ) -> None:
+        with self.progress_lock:
+            for key in keys:
+                self.progress_set.add(key)
 
-        return self.memory_allocator.batched_allocate(shape, dtype, batch_size, fmt)
+        if self.async_mode:
```
ziruiliu, line 1289: Move line 1082,1083 here:
+         for memory_obj in memory_objs:
+                memory_obj.ref_count_up()

ziruiliu:
> @jgoldsch12 I believe the issue is with ref count leaks which is possibly causing this corruption cc: @tshmilnvidia, @ziruiliu

I have updated my comments to address this ref count race condition.
Thanks

jgoldsch12:
> > @jgoldsch12 I believe the issue is with ref count leaks which is possibly causing this corruption cc: @tshmilnvidia, @ziruiliu
> 
> I have updated my comments to address this ref count race condition. Thanks

Adding save_unfull_chunk: False to my lmcache.yaml made all the difference.  I no longer see this crash when this value is set.  If it is not set (presumably setting the default, which may be True?) I consistently hit the crash described.  

We should consider forcing this to False if dynamic mode is used.  Something similar is done when setting enable_async_loading to True.

Here is my working YAML file:

``` yaml
chunk_size: 512
local_cpu: false
nixl_buffer_size: 8573157376
nixl_buffer_device: cpu
save_unfull_chunk: False
enable_async_loading: False
extra_config:
  enable_nixl_storage: true
  nixl_backend: OBJ
  nixl_pool_size: 0
  nixl_presence_cache: False
  nixl_async_put: False
  nixl_backend_params:
    access_key: test
    secret_key: xxx
    bucket: jg-kvcache2
    endpoint_override: http://a.b.c.d:9020
    scheme: http
    use_virtual_addressing: "false"
    req_checksum: required
```

jgoldsch12:
> > > @jgoldsch12 I believe the issue is with ref count leaks which is possibly causing this corruption cc: @tshmilnvidia, @ziruiliu
> > 
> > 
> > I have updated my comments to address this ref count race condition. Thanks
> 
> Adding save_unfull_chunk: False to my lmcache.yaml made all the difference. I no longer see this crash when this value is set. If it is not set (presumably setting the default, which may be True?) I consistently hit the crash described.
> 
> We should consider forcing this to False if dynamic mode is used. Something similar is done when setting enable_async_loading to True.
> 
> Here is my working YAML file:
> 
> ```yaml
> chunk_size: 512
> local_cpu: false
> nixl_buffer_size: 8573157376
> nixl_buffer_device: cpu
> save_unfull_chunk: False
> enable_async_loading: False
> extra_config:
>   enable_nixl_storage: true
>   nixl_backend: OBJ
>   nixl_pool_size: 0
>   nixl_presence_cache: False
>   nixl_async_put: False
>   nixl_backend_params:
>     access_key: test
>     secret_key: xxx
>     bucket: jg-kvcache2
>     endpoint_override: http://a.b.c.d:9020
>     scheme: http
>     use_virtual_addressing: "false"
>     req_checksum: required
> ```

Actually, scratch this comment.  When I set save_unfull_chunk to false, I do not see any GET operations, only PUT operations.  I do not see the crash, but also storage_to_mem is never called.

tshmilnvidia:
> > > > @jgoldsch12 I believe the issue is with ref count leaks which is possibly causing this corruption cc: @tshmilnvidia, @ziruiliu
> > > 
> > > 
> > > I have updated my comments to address this ref count race condition. Thanks
> > 
> > 
> > Adding save_unfull_chunk: False to my lmcache.yaml made all the difference. I no longer see this crash when this value is set. If it is not set (presumably setting the default, which may be True?) I consistently hit the crash described.
> > We should consider forcing this to False if dynamic mode is used. Something similar is done when setting enable_async_loading to True.
> > Here is my working YAML file:
> > ```yaml
> > chunk_size: 512
> > local_cpu: false
> > nixl_buffer_size: 8573157376
> > nixl_buffer_device: cpu
> > save_unfull_chunk: False
> > enable_async_loading: False
> > extra_config:
> >   enable_nixl_storage: true
> >   nixl_backend: OBJ
> >   nixl_pool_size: 0
> >   nixl_presence_cache: False
> >   nixl_async_put: False
> >   nixl_backend_params:
> >     access_key: test
> >     secret_key: xxx
> >     bucket: jg-kvcache2
> >     endpoint_override: http://a.b.c.d:9020
> >     scheme: http
> >     use_virtual_addressing: "false"
> >     req_checksum: required
> > ```
> 
> Actually, scratch this comment. When I set save_unfull_chunk to false, I do not see any GET operations, only PUT operations. I do not see the crash, but also storage_to_mem is never called.

Thanks @jgoldsch12. As of now I still haven't managed to reproduce the error you reported. In any case, I have rebased this PR to include the O_DIRECT fix, could you please test it with the rebased version?

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -593,96 +857,433 @@ def close(self) -> None:
...
+            assert self.meta_shape is not None
+            assert self.meta_dtype is not None
+            assert self.meta_fmt is not None
+            obj = self.memory_allocator.allocate(
+                self.meta_shape, self.meta_dtype, self.meta_fmt
             )
-            torch.cuda.set_device(corrected_device)
-            self.base_buffer = base_buffer  # Prevents early GC of the aligned tensor.
-            self.free_pinned_buffer = False
+            if obj is None:
```
jgoldsch12, line 1058: I don't believe this block is working as intended.  In the original storage_to_mem, all allocations are asserted to succeed.  In this case, we see to be attempting to allow the upper layers to handle this failure.  However, it seems when this case is hit, it causes an assertion failure in the VLLM connector code:

(EngineCore_DP0 pid=2814729)   File "/home/jgoldsch/LMCache-public/lmcache/integration/vllm/vllm_v1_adapter.py", line 1270, in wait_for_save
(EngineCore_DP0 pid=2814729)     assert len(slot_mapping) == len(token_ids)
(EngineCore_DP0 pid=2814729) 

This may need more testing and consideration.  I was able to hit this by undersizing the NIXL buffer.

jgoldsch12:
> > > > > @jgoldsch12 I believe the issue is with ref count leaks which is possibly causing this corruption cc: @tshmilnvidia, @ziruiliu
> > > > 
> > > > 
> > > > I have updated my comments to address this ref count race condition. Thanks
> > > 
> > > 
> > > Adding save_unfull_chunk: False to my lmcache.yaml made all the difference. I no longer see this crash when this value is set. If it is not set (presumably setting the default, which may be True?) I consistently hit the crash described.
> > > We should consider forcing this to False if dynamic mode is used. Something similar is done when setting enable_async_loading to True.
> > > Here is my working YAML file:
> > > ```yaml
> > > chunk_size: 512
> > > local_cpu: false
> > > nixl_buffer_size: 8573157376
> > > nixl_buffer_device: cpu
> > > save_unfull_chunk: False
> > > enable_async_loading: False
> > > extra_config:
> > >   enable_nixl_storage: true
> > >   nixl_backend: OBJ
> > >   nixl_pool_size: 0
> > >   nixl_presence_cache: False
> > >   nixl_async_put: False
> > >   nixl_backend_params:
> > >     access_key: test
> > >     secret_key: xxx
> > >     bucket: jg-kvcache2
> > >     endpoint_override: http://a.b.c.d:9020
> > >     scheme: http
> > >     use_virtual_addressing: "false"
> > >     req_checksum: required
> > > ```
> > 
> > 
> > Actually, scratch this comment. When I set save_unfull_chunk to false, I do not see any GET operations, only PUT operations. I do not see the crash, but also storage_to_mem is never called.
> 
> Thanks @jgoldsch12. As of now I still haven't managed to reproduce the error you reported. In any case, I have rebased this PR to include the O_DIRECT fix, could you please test it with the rebased version?

Thank you for rebasing.  I worked through a few issues and dynamic mode is now working ONLY if save_unfull_chunk is explicitly set to false (and, as I noted above, if the NIXL buffer is large enough for the number of concurrent requests.)  If save_unfull_chunk is left to the default value, I hit the failure when using the dynamic mode storage_to_mem.  I'm not sure why this is, but it bears more investigation as to why this setting is required to avoid a crash.

Regarding my setup, I am using vLLM 0.11.2, Torch 2.9.0.  I do not believe these were built with CUDA-13.  I am using NIXL build 0.8.0.  I'm using meta-llama/Llama-3.2-3B-Instruct and a single GPU.

I'm running VLLM as so:

vllm serve "$MODEL" --dtype bfloat16 --max-model-len 131000 --host 0.0.0.0 --port 8000 --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}' --gpu-memory-utilization 0.8 --no-enable-prefix-caching

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -305,59 +391,82 @@ def init_storage_handlers_object(self, page_size, keys):
...
+
+    def release_storage_handler(self, reg_descs, xfer_handler):
+        """Release storage handler resources"""
+        self.nixl_agent.release_dlist_handle(xfer_handler)
+        self.nixl_agent.deregister_memory(reg_descs)
+
+    def nixl_obj_exists(self, object_key: str) -> bool:
+        reg_list = [(0, 0, 0, object_key)]
+        try:
+            resp = self.nixl_agent.query_memory(reg_list, "OBJ", mem_type="OBJ")
```
jgoldsch12: My recommendation is to not hardcode the input here, but rather capture the backend provided when the object was instantiated.  This way we don't have to change this later to support other backends.  We already have a method which limits the backends for dynamic mode.

ziruiliu:
@jgoldsch12 good catch, to enable dynamic mode, save_unfull_chunk has to be false. Because the nixl storage backend doe snot persist cache's meta with cache file or object (refer to remote_backend/fs_connector's save_chunk_meta option),  it would be difficult to figure out the shape and the number of tokens in a partial object.
In static mode, every cache file/object's metadata is retained in memory so saving and loading partial chunk is feasible. In dynamic mode, chunk's meta is not stored anyway so it does not support unfull chunk.
Maybe we should mention it in markdown document

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: Yes of course, I will fix this. Thanks for this catch!

jgoldsch12:
> @jgoldsch12 good catch, to enable dynamic mode, save_unfull_chunk has to be false. Because the nixl storage backend doe snot persist cache's meta with cache file or object (refer to remote_backend/fs_connector's save_chunk_meta option), it would be difficult to figure out the shape and the number of tokens in a partial object. In static mode, every cache file/object's metadata is retained in memory so saving and loading partial chunk is feasible. In dynamic mode, chunk's meta is not stored anyway so it does not support unfull chunk. Maybe we should mention it in markdown document

Thanks for tracking this down.  What you described makes a lot of sense.
If we wanted to support unfull chunks, there may be some approaches (used by other backends) to store the chunk metadata when PutObject is performed.  This is probably a follow on effort.

Documenting is good, but maybe asserting is better like is done here for when async loading is enabled:

https://github.com/LMCache/LMCache/blob/571bbdc7196318c100d76c0a7ee4e56a2829f2de/lmcache/v1/storage_backend/storage_manager.py#L270

ziruiliu (changes requested):
move ref_count_up() from mem_to_storage() to batched_submit_task()

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: Interestingly, this memory allocation failure case is handled better when async loading is enabled.  As opposed to crashing vLLM, we see this:

(APIServer pid=3361626) INFO:     127.0.0.1:34354 - "POST /v1/chat/completions HTTP/1.1" 200 OK
(EngineCore_DP0 pid=3361823) [2025-12-04 18:09:12,480] LMCache WARNING: Failed to allocate memory (nixl_storage_backend.py:997:lmcache.v1.storage_backend.nixl_storage_backend)
(EngineCore_DP0 pid=3361823) [2025-12-04 18:09:12,481] LMCache ERROR: [asyncio] Exception in callback StorageManager.async_lookup_and_prefetch.<locals>.<lambda>(<_GatheringFu...ertionError()>) at /home/jgoldsch/LMCache-public/lmcache/v1/storage_backend/storage_manager.py:705 (utils.py:442:lmcache.utils)
(EngineCore_DP0 pid=3361823) Traceback (most recent call last):
(EngineCore_DP0 pid=3361823)   File "/usr/lib/python3.12/asyncio/events.py", line 88, in _run
(EngineCore_DP0 pid=3361823)     self._context.run(self._callback, *self._args)
(EngineCore_DP0 pid=3361823)   File "/home/jgoldsch/LMCache-public/lmcache/v1/storage_backend/storage_manager.py", line 705, in <lambda>
(EngineCore_DP0 pid=3361823)     lambda future: self.prefetch_all_done_callback(
(EngineCore_DP0 pid=3361823)                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3361823)   File "/home/jgoldsch/LMCache-public/lmcache/v1/storage_backend/storage_manager.py", line 531, in prefetch_all_done_callback
(EngineCore_DP0 pid=3361823)     res = task.result()
(EngineCore_DP0 pid=3361823)           ^^^^^^^^^^^^^
(EngineCore_DP0 pid=3361823)   File "/home/jgoldsch/LMCache-public/lmcache/v1/storage_backend/storage_manager.py", line 172, in run
(EngineCore_DP0 pid=3361823)     return await coro_fn
(EngineCore_DP0 pid=3361823)            ^^^^^^^^^^^^^
(EngineCore_DP0 pid=3361823)   File "/home/jgoldsch/LMCache-public/lmcache/v1/storage_backend/nixl_storage_backend.py", line 1259, in batched_get_non_blocking
(EngineCore_DP0 pid=3361823)     assert None not in obj_list
(EngineCore_DP0 pid=3361823)            ^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3361823) AssertionError
(EngineCore_DP0 pid=3361823) [2025-12-04 18:09:15,210] LMCache WARNING: Request chatcmpl-9209377d0dfd4b879eb9d9ad3f7e3572 is still waiting for async lookup after 3 seconds, returning 0 lmcache cached tokens so vllm can recompute (lmcache_async_lookup_client.py:183:lmcache.v1.lookup_client.lmcache_async_lookup_client)
(EngineCore_DP0 pid=3361823) [2025-12-04 18:09:15,211] LMCache INFO: Reqid: chatcmpl-9209377d0dfd4b879eb9d9ad3f7e3572, Total tokens 80044, LMCache hit tokens: 0, need to load: 0 (vllm_v1_adapter.py:1511:lmcache.integration.vllm.vllm_v1_adapter)
(EngineCore_DP0 pid=3361823) [2025-12-04 18:09:15,313] LMCache INFO: Storing KV cache for 8192 out of 8192 tokens (skip_leading_tokens=0) for request chatcmpl-9209377d0dfd4b879eb9d9ad3f7e3572 (vllm_v1_adapter.py:1293:lmcache.integration.vllm.vllm_v1_adapter)

In the case of async loading, the assertion failure in batched_get_non_blocking can be recovered and results in recompute/store within the external kvcache.

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: This worked much better.  Otherwise I see these warnings:

(EngineCore_DP0 pid=3373712) [2025-12-04 18:44:21,534] LMCache WARNING: Ref count of MemoryObj 43is negative: -1.Double free occurred somewhere.Setting ref count back to 0 as a hack but please find the bug. (memory_management.py:412:lmcache.v1.memory_management)

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -593,96 +857,433 @@ def close(self) -> None:
...
+                raise RuntimeError("NIXL transfer failed")
+
+            for key in keys:
+                with self.progress_lock:
+                    self.progress_set.discard(key)
+                self._cache_add(key.chunk_hash)
+        finally:
+            # Release the handle after transfer completes (success or failure)
+            self.agent.release_handle(handle)
+            self.agent.release_storage_handler(storage_reg_descs, storage_xfer_handler)
```
jgoldsch12, line 1140: This seems to create a race between PUTs and GETs when async PUT mode is enabled.  
The async PUTs complete successfully and then the key is removed from the in-progress list and optionally added to the presence cache.  A call to contains() will now return true for this key, so a GET may be performed.  If the GET is performed (storage_to_mem()) it will register the OBJ_SEG with NIXL using an object name and devID.  Given that the devID number space is shared and not unique to the request, this call to release_storage_handler could deregister a devID (let's use devID 0 as an example, since that's the issue that I saw).  When the storage_to_mem path goes to perform the transfer, NIXL will complain that the OBJ_SEG is not registered, since it cannot find the devID in its in-memory dictionary.  The ordering of deregistering and removal from the in-progress list/adding to the presence cache needs to be rethought when async PUT mode is used.

jgoldsch12:
Unclear if this crash is related, as it is upstack in vLLM.  I've hit this consistently running LMBenchmark long input/short reply test with NIXL OBJ plugin (dynamic mode, presence cache enabled, async loading disabled, save_unfull_chunk false, async put false)

(EngineCore_DP0 pid=3721220) [2025-12-05 18:45:27,831] LMCache INFO: Reqid: cmpl-b3eb2adf502c7925-0, Total tokens 5272, LMCache hit tokens: 4608, need to load: 3600 (vllm_v1_adapter.py:1511:lmcache.integration.vllm.vllm_v1_adapter)
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845] EngineCore encountered a fatal error.
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845] Traceback (most recent call last):
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 836, in run_engine_core
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]     engine_core.run_busy_loop()
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 863, in run_busy_loop
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]     self._process_engine_step()
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 892, in _process_engine_step
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]     outputs, model_executed = self.step_fn()
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]                               ^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 342, in step
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]     scheduler_output = self.scheduler.schedule()
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]                        ^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/core/sched/scheduler.py", line 531, in schedule
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]     assert num_new_tokens > 0
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845]            ^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=3721220) ERROR 12-05 18:45:27 [core.py:845] AssertionError
(Worker_TP2 pid=3721388) INFO 12-05 18:45:27 [multiproc_executor.py:709] Parent process exited, terminating worker
(Worker_TP3 pid=3721389) INFO 12-05 18:45:27 [multiproc_executor.py:709] Parent process exited, terminating worker
(Worker_TP0 pid=3721386) INFO 12-05 18:45:27 [multiproc_executor.py:709] Parent process exited, terminating worker
(Worker_TP1 pid=3721387) INFO 12-05 18:45:27 [multiproc_executor.py:709] Parent process exited, terminating worker
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546] AsyncLLM output_handler failed.
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546] Traceback (most recent call last):
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/async_llm.py", line 498, in output_handler
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546]     outputs = await engine_core.get_output_async()
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546]               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core_client.py", line 885, in get_output_async
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546]     raise self._format_exception(outputs) from None
(APIServer pid=3721008) ERROR 12-05 18:45:27 [async_llm.py:546] vllm.v1.engine.exceptions.EngineDeadError: EngineCore encountered an issue. See stack trace (above) for the root cause.
(APIServer pid=3721008) INFO:     127.0.0.1:36436 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     127.0.0.1:36448 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     127.0.0.1:36456 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     127.0.0.1:36466 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     127.0.0.1:36482 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     127.0.0.1:36482 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     127.0.0.1:52418 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
(APIServer pid=3721008) INFO:     Shutting down
(APIServer pid=3721008) INFO:     Waiting for application shutdown.
(APIServer pid=3721008) INFO:     Application shutdown complete.
(APIServer pid=3721008) INFO:     Finished server process [3721008]

I've not hit this issue before with vLLM when running this test with NIXL OBJ and my own changes to LMCache.

ziruiliu:
> v1/engine/core.py

Hi @jgoldsch12 . I tried benchmarks/long_doc_qa/long_doc_qa.py with Qwen3-4B on a single GPU node but did not see such error. Would you please list your model, hardware config and how you meet this bug?
Also, add a debug line right before the assertion would help a lot 
```diff
+                   logger.info(f"num_new_tokens is {num_new_tokens}, budget is {token_budget}")
                   num_new_tokens = min(num_new_tokens, token_budget)
                    assert num_new_tokens > 0
```
Thanks

jgoldsch12:
I'm using "meta-llama/Llama-3.3-70B-Instruct".  I'm running VLLM in
aggregated mode with 4 visible H100 GPUs and TP=4.

Prior testing with LMBenchmark was stable using vLLM 0.10.2.  I'm looking
at what changed.

My colleague also hit this issue, but they were not using this PR and did
not have save_unfull_chunk disabled.  They were testing GDS NIXL plugin
with LMCache.

ziruiliu:
Maybe it's vllm's version. I also update vllm to latest release so I tested with vllm 0.11.2.

jgoldsch12:
I rolled back to vllm 0.10.2, but continue to hit the issue.  When adding
debugging I see this:

(EngineCore_DP0 pid=212324) INFO 12-08 18:57:22 [scheduler.py:435]
num_new_tokens is -344, budget is 8134

Finding this very easy to hit.  I'll try to work backwards and see why this
number is negative.

Prior info messages show:

(APIServer pid=211989) INFO:     127.0.0.1:41304 - "POST /v1/completions
HTTP/1.1" 200 OK
(EngineCore_DP0 pid=212324) INFO 12-08 18:57:19 [scheduler.py:435]
num_new_tokens is 664, budget is 8137
(APIServer pid=211989) INFO:     127.0.0.1:41310 - "POST /v1/completions
HTTP/1.1" 200 OK
(EngineCore_DP0 pid=212324) INFO 12-08 18:57:20 [scheduler.py:435]
num_new_tokens is 664, budget is 8136
(APIServer pid=211989) INFO:     127.0.0.1:53278 - "POST /v1/completions
HTTP/1.1" 200 OK
(APIServer pid=211989) INFO 12-08 18:57:21 [loggers.py:123] Engine 000: Avg
prompt throughput: 6807.9 tokens/s, Avg generation throughput: 97.1
tokens/s, Running: 56 reqs, Waiting: 0 reqs, GPU KV cache usage: 95.7%,
Prefix cache hit rate: 21.8%
(EngineCore_DP0 pid=212324) INFO 12-08 18:57:21 [scheduler.py:435]
num_new_tokens is 152, budget is 8135
(APIServer pid=211989) INFO:     127.0.0.1:53286 - "POST /v1/completions
HTTP/1.1" 200 OK
(EngineCore_DP0 pid=212324) INFO 12-08 18:57:22 [scheduler.py:435]
num_new_tokens is 664, budget is 8134

jgoldsch12:
The issue seems to be rooted in the LMCache V1 connector which is returning 'num_external_computed_tokens' to the VLLM scheduler through the get_num_new_matched_tokens() interface.   

(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:377] request.num_computed_tokens is 0, request.num_tokens is 5272
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:400] num_external_computed_tokens is 4112
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:401] num_new_local_computed_tokens is 1008
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:413] num_computed_tokens is 5120
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:426] request.num_tokens is 5272, num_computed_tokens is 5120
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:428] After computing num_new_tokens: 152
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:442] num_new_tokens is 152, budget is 8135
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:377] request.num_computed_tokens is 0, request.num_tokens is 5272
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:400] num_external_computed_tokens is 5120
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:401] num_new_local_computed_tokens is 1008
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:413] num_computed_tokens is 6128
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:426] request.num_tokens is 5272, num_computed_tokens is 6128
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:428] After computing num_new_tokens: -856
(EngineCore_DP0 pid=369672) INFO 12-09 15:03:02 [scheduler.py:442] num_new_tokens is -856, budget is 8135

In normal cases, I see get_num_new_matched_tokens repond with 4112 as a maximum value.  num_new_local_computed_tokens seems to always be 1008 (which I don't fully understand.)  

In the failing case, the LMCache V1 connector returns 5120 from get_num_new_matched_tokens.  This value, plus the local tokens exceeds the number of tokens in the request (6128 vs 5272).  Why does the get_num_new_matched_tokens suddenly repond with this out of bound value?  And what has recently changed in LMCache in this space?  @ziruiliu, I noticed there has been some recent refactoring in vllm_v1_adapter.py

jgoldsch12:
I wonder if https://github.com/LMCache/LMCache/pull/2146 is what addresses this issue?

tshmilnvidia:
Updated this PR + Rebase.
Following issues addressed, each in a seperate commit:
* asserting on dynamic_mode + save_unfull_chunk not False
* Better generalization of create_batched_storage_handler (as suggested by @jgoldsch12 )
* Fixing mem_obj ref_count leak (fix from @ziruiliu)
* Fixing order of nixl storage handler & removal from progress_set in async put

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: This is appears resolved in the new code

jgoldsch12:
Syncing up to the latest code in this PR (which appears to rebase to head), I now see a different assertion failure than before when I'm running the multi-turn test:

(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:377] request.num_computed_tokens is 0, request.num_tokens is 5498
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:400] num_external_computed_tokens is 4112
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:401] num_new_local_computed_tokens is 1008
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:413] num_computed_tokens is 5120
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:426] request.num_tokens is 5498, num_computed_tokens is 5120
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:428] After computing num_new_tokens: 378
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:442] num_new_tokens is 378, budget is 8031
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:377] request.num_computed_tokens is 0, request.num_tokens is 5498
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:400] num_external_computed_tokens is 4112
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:401] num_new_local_computed_tokens is 1008
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:413] num_computed_tokens is 5120
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:426] request.num_tokens is 5498, num_computed_tokens is 5120
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:428] After computing num_new_tokens: 378
(EngineCore_DP0 pid=573786) INFO 12-10 21:04:40 [scheduler.py:442] num_new_tokens is 378, budget is 7653
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720] EngineCore encountered a fatal error.
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720] Traceback (most recent call last):
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 711, in run_engine_core
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     engine_core.run_busy_loop()
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 738, in run_busy_loop
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     self._process_engine_step()
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 764, in _process_engine_step
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     outputs, model_executed = self.step_fn()
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]                               ^^^^^^^^^^^^^^
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/engine/core.py", line 291, in step
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     scheduler_output = self.scheduler.schedule()
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]                        ^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/v1/core/sched/scheduler.py", line 615, in schedule
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     meta = self.connector.build_connector_meta(scheduler_output)
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/lmcache-object/lib/python3.12/site-packages/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 150, in build_connector_meta
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     return self._lmcache_engine.build_connector_meta(scheduler_output)
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]   File "/home/jgoldsch/LMCache-public/lmcache/integration/vllm/vllm_v1_adapter.py", line 1770, in build_connector_meta
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]     assert request.num_computed_tokens == lmcache_cached_tokens, (
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=573786) ERROR 12-10 21:04:40 [core.py:720] AssertionError: Preempted request cmpl-89a155d20bbc4a85a6f41dcdd714b50d-0 has num_computed_tokens 5408 but lmcache_cached_tokens 5120

This is the code block failing in vllm_v1_adapter.py:

if preempted:
                # num_computed_tokens should be reset to 0 during preemption
                # and then set to the number of external tokens (from vllm
                # scheduler's perspective)
                # this assumption is crucial for the update() call of RequestTracker
                assert request.num_computed_tokens == lmcache_cached_tokens, (
                    f"Preempted request {req_id} has "
                    f"num_computed_tokens {request.num_computed_tokens} "
                    f"but lmcache_cached_tokens {lmcache_cached_tokens}"
                )

There seems to still be instability which is impacting testing of this change.

ziruiliu:
@jgoldsch12 the assertion you showed about preemption above could be possibly introduced by the PR #2146 
You can simply verify this by switch from nixl storage backend to local disk backend or remote backend + fs_connector, to see if the same issue happened on local file system. If so, there could be a potential bug in lookup() mechanism not this nixl storage backend.
You test env might have higher pressure than my test env. The negative number, the mismatched number are both triggered when preemption happened. Thanks

jgoldsch12:
I also believe that PR is at fault (it also seems to have addressed the
previous issue I was seeing, but introduced this new one during
preemption.)  I also believe this has nothing to do with the
NixlStorageBackend, nor the changes in this PR (potentially, as you
suggest, the bug is easier to hit using this setup.)  I'm confident we
would see it with local CPU, as we also hit it with file backends.  My
point is more this issue is preventing testing of this change, which is
frustrating, as is the general stability of LMCache at this time.

jgoldsch12:
This PR seems to address the issue I was seeing.  https://github.com/LMCache/LMCache/pull/2189
I'm making progress now.

ziruiliu (approved):
all issues have been addressed

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: This appears to still be an issue in the latest code.

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: This request wasn't resolved.  Is it possible to make this change?

jgoldsch12 (changes requested):
A few reported issues to consider prior to submitting.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -305,59 +401,89 @@ def init_storage_handlers_object(self, page_size, keys):
...
+
+    def release_storage_handler(self, reg_descs, xfer_handler):
+        """Release storage handler resources"""
+        self.nixl_agent.release_dlist_handle(xfer_handler)
+        self.nixl_agent.deregister_memory(reg_descs)
+
+    def nixl_obj_exists(self, object_key: str) -> bool:
+        reg_list = [(0, 0, 0, object_key)]
+        try:
+            resp = self.nixl_agent.query_memory(reg_list, "OBJ", mem_type="OBJ")
```
jgoldsch12: The comment here appears to have been mistakenly marked as resolved, so I'm adding it again:

My recommendation is to not hardcode the input here, but rather capture the backend provided when the object was instantiated. This way we don't have to change this later to support other backends. We already have a method which limits the backends for dynamic mode.

jgoldsch12:
Folks, I've been doing multi-turn testing LMBenchmark comparing static and dynamic mode for OBJ.  I'm seeing significant differences between the two as shown on the graph below.  For static mode, I have added PR #2189 to enable batched_get_blocking().  For static, save_unful_chunks is the default (true) and the pool is set to 5000.  For dynamic mode, save_unfull_chunks is false, presence cache is true.  

image

This graph is showing how input tokens per second scales with the QPS range of 1->3).  ISL is 20k, OSL is 100, with 80 users performing 10 rounds over a 100 second run.  I'm unclear if this degradation is the result of some change in lmcache that got swept up in this PR, the disabling of save_unfull_chunks or something in dynamic mode itself.

DongDongJu:
> Folks, I've been doing multi-turn testing LMBenchmark comparing static and dynamic mode for OBJ. I'm seeing significant differences between the two as shown on the graph below. For static mode, I have added PR #2189 to enable batched_get_blocking(). For static, save_unful_chunks is the default (true) and the pool is set to 5000. For dynamic mode, save_unfull_chunks is false, presence cache is true.

Hello @jgoldsch12, We changed default value of `save_unfull_chunks` to false. Could you verify that this degradation is coming from `save_unfull_chunks` is false?

> 
> image
> This graph is showing how input tokens per second scales with the QPS range of 1->3). ISL is 20k, OSL is 100, with 80 users performing 10 rounds over a 100 second run. I'm unclear if this degradation is the result of some change in lmcache that got swept up in this PR, the disabling of save_unfull_chunks or something in dynamic mode itself.

DongDongJu:
@jgoldsch12 I couldn't review the all codes and test as well becuz dont have nixl access now.
but it seems when we do dynamic mode, it kept running `create_batched_storage_handler()->register_memory->get_xfer_descs->prep_xfer_dlist` in every put and get (`storage_to_mem, mem_to_stroage`).
then, when QPS is going up, means that RPC/metat data work takes more which possibly affect to the performance.
Let me know what you think.

jgoldsch12:
> @jgoldsch12 I couldn't review the all codes and test as well becuz dont have nixl access now. but it seems when we do dynamic mode, it kept running `create_batched_storage_handler()->register_memory->get_xfer_descs->prep_xfer_dlist` in every put and get (`storage_to_mem, mem_to_stroage`). then, when QPS is going up, means that RPC/metat data work takes more which possibly affect to the performance. Let me know what you think.

Thank you for this @DongDongJu.  I ran static mode with save_unfull_chunks set to true and got nearly identical results - so I don't believe save_unfull_chunks is at issue.  I'll keep digging.

ziruiliu:
Hi @jgoldsch12, in comments above, did you mean PR #1869 but not 2189 that enables batched_get_blocking().
I am trying to reproduce your test to find out difference of static mode and dynamic mode, would you also post the details of configuration you tested with? for example, the gpu mem utilization or kv cache size in vllm's config, the lmcache's config that includes nixl's param like nixl buffer size and etc.

jgoldsch12:
> Hi @jgoldsch12, in comments above, did you mean PR #1869 but not 2189 that enables batched_get_blocking(). I am trying to reproduce your test to find out difference of static mode and dynamic mode, would you also post the details of configuration you tested with? for example, the gpu mem utilization or kv cache size in vllm's config, the lmcache's config that includes nixl's param like nixl buffer size and etc.

I'm testing with 4xH100 (TP=4) and .7 memory utilization.


``` yaml
chunk_size: 512
local_cpu: false
nixl_buffer_size: 12037652480
nixl_buffer_device: cpu
enable_async_loading: False
save_unfull_chunk: False
extra_config:
  enable_nixl_storage: true
  nixl_backend: OBJ
  nixl_pool_size: 0
  nixl_presence_cache: True
  nixl_async_put: False
  nixl_backend_params:
    access_key: test
    secret_key: secret
    bucket: bucket
    endpoint_override: http://10.10.10.10:9020
    scheme: http
    use_virtual_addressing: "false"
    req_checksum: required
```

(Worker_TP1 pid=1980990) [2025-12-15 13:43:07,271] LMCache INFO: LMCache initialized for role KVConnectorRole.WORKER with version 0.3.11.dev13-gb32e6a232, vllm version 0.12.0, lmcache cache_engine metadata: LMCacheEngineMetadata(model_name='meta-llama/Llama-3.3-70B-Instruct', world_size=4, worker_id=1, fmt='vllm', kv_dtype=torch.bfloat16, kv_shape=(80, 2, 512, 2, 128), use_mla=False, role='worker', served_model_name='meta-llama/Llama-3.3-70B-Instruct') (vllm_v1_adapter.py:802:lmcache.integration.vllm.vllm_v1_adapter)

```bash
Mon Dec 15 13:48:16 2025       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.65.06              Driver Version: 580.65.06      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H100 80GB HBM3          On  |   00000000:19:00.0 Off |                    0 |
| N/A   44C    P0             76W /  700W |       0MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA H100 80GB HBM3          On  |   00000000:3B:00.0 Off |                    0 |
| N/A   26C    P0            117W /  700W |   60557MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA H100 80GB HBM3          On  |   00000000:4C:00.0 Off |                    0 |
| N/A   25C    P0            114W /  700W |   60557MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   3  NVIDIA H100 80GB HBM3          On  |   00000000:5D:00.0 Off |                    0 |
| N/A   24C    P0             71W /  700W |       0MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   4  NVIDIA H100 80GB HBM3          On  |   00000000:9B:00.0 Off |                    0 |
| N/A   24C    P0             69W /  700W |       0MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   5  NVIDIA H100 80GB HBM3          On  |   00000000:BB:00.0 Off |                    0 |
| N/A   26C    P0            113W /  700W |   60557MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   6  NVIDIA H100 80GB HBM3          On  |   00000000:CB:00.0 Off |                    0 |
| N/A   24C    P0             69W /  700W |       0MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   7  NVIDIA H100 80GB HBM3          On  |   00000000:DB:00.0 Off |                    0 |
| N/A   23C    P0            112W /  700W |   60557MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    1   N/A  N/A         1980989      C   VLLM::Worker_TP0                      60516MiB |
|    2   N/A  N/A         1980990      C   VLLM::Worker_TP1                      60516MiB |
|    5   N/A  N/A         1980991      C   VLLM::Worker_TP2                      60516MiB |
|    7   N/A  N/A         1980992      C   VLLM::Worker_TP3                      60516MiB |
+-----------------------------------------------------------------------------------------+
```

ziruiliu:
@jgoldsch12 , by 'ISL is 20k, OSL is 100, with 80 users performing 10 rounds over a 100 second run' do you set these in synthetic-multi-round-qa/long_input_long_output.sh? 
I also suggest that we end/merge this PR first to enable dynamic mode first, if all explicit leak/race condition issues are addressed (there are 2 more change requests pending), letting more users be able to try the dynamic deployment on s3-compatible storage product. 
After this dynamic mode is enabled, we can open new issues to trace new problems we found in nixl storage backend. Vllm and LMCache evolves very fast so performance degradation or other bugs may pop up any time. This discussion in this PR is too long and has gone off the topic somehow.

jgoldsch12:
> @jgoldsch12 , by 'ISL is 20k, OSL is 100, with 80 users performing 10 rounds over a 100 second run' do you set these in synthetic-multi-round-qa/long_input_long_output.sh? I also suggest that we end/merge this PR first to enable dynamic mode first, if all explicit leak/race condition issues are addressed (there are 2 more change requests pending), letting more users be able to try the dynamic deployment on s3-compatible storage product. After this dynamic mode is enabled, we can open new issues to trace new problems we found in nixl storage backend. Vllm and LMCache evolves very fast so performance degradation or other bugs may pop up any time. This discussion in this PR is too long and has gone off the topic somehow.

@ziruiliu Here the LMBenchmark settings:

#LMbenchmark Variables
export NUM_USERS=80           # Number of concurrent users
export NUM_ROUNDS=10          # Rounds per user
export SYSTEM_PROMPT=1000     # System prompt length
export CHAT_HISTORY=20000     # Chat history length per user
export ANSWER_LEN=100         # Generation length per round
export TEST_DURATION=100      # Test duration in seconds

I run it using:

./run_benchmarks.sh $MODEL http://localhost:8000 results-obj-static long-input 1 1.5 2.0 2.5 3.0 >> results-obj-static.log

I hear you that it has gone on long, but I don't think static vs dynamic performance comparison is off topic.  I want to make sure this change is of good quality and provides the expected benefit.  As of right now, something isn't right.  The change is fine from a functional standpoint (outstanding comments aside.)  If merged, the expectations should be low and an issue should be opened to determine why performance suffers so much with dynamic mode.

I had implemented my own OBJ plugin support in LMCache.  My approach also avoided using the static file pools.  It did, however, cache the tensor shapes and allowed for unfull chunks.  So far I haven't found this to be an issue, but it is something to investigate more.

DongDongJu:
Hello, @ziruiliu @jgoldsch12 
How about make the new issue with pended series of these work? I will follow the history and review them ASAP.
And, we can discuss current and expected issues in there. What do you guys think?

ziruiliu:
> Hello, @ziruiliu @jgoldsch12 How about make the new issue with pended series of these work? I will follow the history and review them ASAP. And, we can discuss current and expected issues in there. What do you guys think?

Agree.
I tried to reproduce LMBenchmark that Jason @jgoldsch12 have done. 
For the latest version, I saw the assertion error mentioned above -  num_computed_tokens and lmcache_cached_tokens don't match.
Rollback to the earlier version where 2146 was not introduced,  I got numbers like:
QPS Input tokens/s
1.0 21672
1.5 30851
2.0 32381
2.5 28099
3.0 23228
The numbers do not scale perfectly however the input tokens/s does not decline sharply as Jason posted. There could be too many factors impacting the perf number so I would like to see if we can discuss perf issues in a new issue

jgoldsch12:
It is possible that 2146 introduced this regression.

Sounds like a fine plan.

ziruiliu:
Hi @jgoldsch12 , I have set up environment with vllm==0.12.0 and lmcache==0.3.11 which includes the fix #2243 to the assertion introduced in #2146. Tests went well and no assertion occur again. Here are numbers I got:
- Dynamic + presence cache ON + async put OFF
0.5 3900
1.0 6283
1.5 10410
2.0 11071
2.5 11761
3.0 12362
- Static + batched_get_blocking #1869
0.5 3964
1.0 6402
1.5 10696
2.0 11310
2.5 11914
3.0 12600
- Static (vanilla)
0.5 3338
1.0 4107
1.5 4812
2.0 4672
2.5 4849
3.0 4886

The numbers scale as expected. I believe that we should move on with this PR and open a new issue to track if there is any corner case that causes the unexpected numbers you have encountered

BTW, there is one place to be updated to align with latest lmcache code. I will post a review comment soon.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -375,75 +501,202 @@ def __init__(
...
+            torch.cuda.set_device(corrected_device)
+            self.base_buffer = base_buffer  # Prevents early GC of the aligned tensor.
+            self.free_pinned_buffer = False
 
-        :param key: The key to check
-        :param pin: Whether to pin the object in the backend.
+        return PagedTensorMemoryAllocator(
+            self.buffer,
+            torch.Size(metadata.kv_shape),
+            metadata.kv_dtype,
```
ziruiliu: Please update line 538 and 539 as
```python
[torch.Size(metadata.kv_shape)],
[metadata.kv_dtype],
```
to be compatible with latest PagedTensorMemoryAllocator interface

ziruiliu (changes requested):

jgoldsch12:
Thanks for running this and sharing the results.  Is this with TP=4?  Can
you share your lmcache.yaml also for the dynamic test you ran?  Thanks.

ziruiliu:
I ran test with "-tp 4"
For dynamic:
```yaml
local_cpu: False
save_unfull_chunk: False
enable_async_loading: False # set to True to test async loading
nixl_buffer_size: 21474836480
nixl_buffer_device: cpu
extra_config:
  enable_nixl_storage: True
  nixl_backend: OBJ
  nixl_pool_size: 0
  nixl_presence_cache: True
  nixl_async_put: False
  nixl_backend_params:
...
```
In dynamic mode, I can see there 34,088 objects created after entire benchmark test completed

For static:
```yaml
chunk_size: 512
local_cpu: False
save_unfull_chunk: False
enable_async_loading: False # set to True to test async loading
nixl_buffer_size: 21474836480
nixl_buffer_device: cpu
extra_config:
  enable_nixl_storage: True
  nixl_backend: OBJ
  nixl_pool_size: 5000
  nixl_backend_params:
 ...
```
Only difference is the non-zero nixl_pool_size. In static mode, there will be 5000 * 4 = 20,000 objects created

lmcache/v1/storage_backend/nixl_storage_backend.py
ziruiliu: I can reproduce this in sync mode, if setting nixl_buffer_size to a small number.
screenshot:
image

This bug not only happen with nixl storage backend, but also local disk backend and fs_connector+remote backend, when `max_local_cpu_size` is set smaller than one request's kv cache size.
For example, in my test, a single request is generate 32G kv cache and max_local_cpu_size is set to 20G

The bug in sync mode looks like something went wrong in scheduler and upper layer. The dynamic backend itself failed to allocate enough memory (maybe cache to be loaded is larger than `nixl_buffer_size` in yaml), so the backend reported 0 objects is loaded, which is expected.

Shorter than expected cache retrieved is handled by lmcache and vllm, but somehow when request is rescheduled, the request's length is longer than expected, in the screenshot, the request's length is 139008 which should be 131072.

I am not sure how this is handled normally in async loading mode, but it is not a bug in nixl storage backend. I can dive in this bug and try to fix later but we can move forward without being blocked by this one

jgoldsch12:
Thanks for investing

lmcache/v1/storage_backend/nixl_storage_backend.py
tshmilnvidia: Thanks @ziruiliu, I'll skip this one.

tshmilnvidia:
Updated code:

* Address missing generalization in nixl_obj_exists as pointed out by @jgoldsch12 
* Rebase to take care of the PagedTensorMemoryAllocator pointed by @ziruiliu 
* Not addressing the undersized buffer error as @ziruiliu encountered it with different backends as well

I believe this should cover everything, please let me know if I missed anything

ziruiliu (approved):

vvenkates27 (approved):
Given that all open issues have been addressed and tested, I believe this PR is ready

sammshen (approved):
LGTM

sammshen:
@Shaoting-Feng can you help merge this? thank you!

Shaoting-Feng (approved):

jgoldsch12:
Ship it!  Looks good. Thanks for the effort.

tshmilnvidia:
Rebased due to conflict with PR #1869

PR #2033, opened 2025-11-19
[controller]prevent duplicate registration

chunxiaozheng:
prevent duplicate registration


lmcache/v1/cache_controller/controllers/registration_controller.py
```diff
@@ -104,16 +104,28 @@ async def register(self, msg: RegisterMsg) -> None:
...
         instance_id = msg.instance_id
         worker_id = msg.worker_id
+        key = (instance_id, worker_id)
+
+        # prevent duplicate registration
+        if (
+            key in self.distributed_url_mapping
+            or key in self.socket_mapping
+            or key in self.worker_info_mapping
+        ):
```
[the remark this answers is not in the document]
maobaolong: @chunxiaozheng Do you think what gemini-bot said make sense?

maobaolong (commented):
@chunxiaozheng Left acomment

lmcache/v1/cache_controller/controllers/registration_controller.py
```diff
@@ -104,16 +104,28 @@ async def register(self, msg: RegisterMsg) -> None:
...
+            logger.warning(
+                f"Instance-worker {key} already registered, skip registration"
+            )
+            return
+
```
ApostaC: Dumb question: Will this controller be called by multiple threads? If so, do we want to have any kind of thread-safe guarantee?

ApostaC (approved):
Otherwise LGTM. 

One quick thing is that if you happen to change exiting logging & add more logging, can we avoid using f-string? 
i.e., let's use `logger.info("log some variable %s", some_variable)` instead of `logger.info(f"log some variable {some_variable}")`

chunxiaozheng:
> Otherwise LGTM.
> 
> One quick thing is that if you happen to change exiting logging & add more logging, can we avoid using f-string? i.e., let's use `logger.info("log some variable %s", some_variable)` instead of `logger.info(f"log some variable {some_variable}")`

@ApostaC Thanks for your suggestion! I have updated.

lmcache/v1/cache_controller/controllers/registration_controller.py
chunxiaozheng: > Dumb question: Will this controller be called by multiple threads? If so, do we want to have any kind of thread-safe guarantee?

Yeah, there is currently no thread safety protection in controller, we can submit a separate PR later to handle it.

maobaolong (approved):
LGMT

PR #2034, opened 2025-11-19
[Controller] Setup internal_api_server to controller to make it observable

maobaolong:
# Motivation
- Make controller observable
- Support existing common api, e.g. /threads /env /metrics ...
- Support to extend more api.


maobaolong:

lmcache/v1/internal_api_server/api_registry.py
```diff
@@ -1,35 +1,59 @@
...
 # Third Party
 from fastapi import APIRouter
 
+APICategory = Literal["common", "vllm", "controller"]
+
 
 class APIRegistry:
     """
-    Automatically discovers and registers API routes
+    Automatically discovers and registers API routes by category
```
sammshen, line 16: the doc string is very helpful!

lmcache/v1/internal_api_server/api_registry.py
```diff
@@ -1,35 +1,59 @@
...
+
+    Categories:
+    - common: APIs that work for all components (metrics, logs, etc.)
```
sammshen, line 19: could we keep a list of all of the APIs in this doc?

sammshen (approved):
LGTM!

lmcache/v1/internal_api_server/api_registry.py
maobaolong: It would introduce extra changes to this files for every new api adding, and the initial motivation for the `api_registry` is to reduce unrelated changes as possible as I can.

But I get what you need, we can add all api into the document.

chunxiaozheng (approved):
lgtm

PR #2040, opened 2025-11-20
Adding chunk lifespan tracker to ChunkStatisticsLookupClient

YuhanLiu11:
**What this PR does / why we need it**: This PR adds a chunk lifespan tracker, such that the output json of "/chunk_statistics/status" endpoint outputs a "chunk_lifespan_buckets" too:

An example output of the status endpoint is:

```
{
  "enabled": true,
  "total_requests": 3,
  "timing": {
    "lookup_time_seconds": 0.0047152042388916016,
    "record_statistics_time_seconds": 7.796287536621094e-05,
    "check_exit_conditions_time_seconds": 5.4836273193359375e-06,
    "total_time_seconds": 0.0047986507415771484,
    "overhead_time_seconds": 8.344650268554688e-05,
    "overhead_percentage": 1.738957619118597
  },
  "total_chunks": 3,
  "unique_chunks": 2,
  "duplicate_chunks": 1,
  "reuse_rate": 0.3333333333333333,
  "chunk_lifespan_buckets": {
    "<= 0.5 hour": 2,
    "0.5 to 1 hour": 0,
    "1 to 2 hours": 0,
    "2 to 5 hours": 0,
    "5 to 10 hours": 0,
    "10 to 24 hours": 0,
    "24 to 48 hours": 0,
    "> 48 hours": 0
  },
  "bloom_filter": {
    "size_mb": 22.852559089660645,
    "hash_count": 6,
    "item_count": 2,
    "bits_set": 12,
    "fill_rate": 6.259742800626769e-08,
    "expected_elements": 20000000,
    "false_positive_rate": 0.01
  },
  "async_queue": {
    "capacity": 100000,
    "current_size": 0,
    "max_size_reached": 0,
    "full_blocks": 0,
    "utilization": 0.0
  },
  "timestamp": 1763599653.9581861,
  "auto_exit_enabled": true,
  "auto_exit_timeout_hours": 0.0,
  "auto_exit_target_unique_chunks": 0
}
```

Also did some quick benchmarking to see if it introduces extra overhead. I ran the following command with LMBench:
``bash run_synthetic.sh vllm_cpu_offload http://localhost:8000 naive-vllm 40 40  20 1000 20000 100  false naive-vllm 0 0-bench-specs/routing/4-spec.yaml 0.5 completions 3``
which runs multi-round workload with 40 concurrent users with 20K input contexts, at a QPS of 3. 

If we only run LMCache with CPU offloading (100 GB max CPU memory), the TTFT is 1.1s, and ITL is 0.057s.
If we run LMCache with CPU offloading, and enabling this chunk lifespan tracker, the TTFT is 1.06s, and ITL is 0.054s. 

**If applicable**:


lmcache/v1/lookup_client/record_strategies/base.py
```diff
@@ -30,6 +81,14 @@ def __init__(self, chunk_size: int):
         self.total_chunks = 0
         self.unique_chunks_count = 0
         self.lock = threading.RLock()
+        self.chunk_to_timestamps: dict[
```
maobaolong, line 84: This introduced unlimited dict can affect the memory cost and performance, WDYT?

lmcache/v1/lookup_client/record_strategies/base.py
```diff
@@ -30,6 +81,14 @@ def __init__(self, chunk_size: int):
...
+            Union[str, tuple[int, ...]], float
+        ] = {}  # chunk hash to list of bloom filter positions
+        # or chunk hash to timestamp of first seen
+        self.chunk_to_latest_timestamp: dict[
```
maobaolong, line 88: doubled mem cost

maobaolong (changes requested):
@YuhanLiu11 Thanks for this great feature, it is very useful, it can help us insight the chunk lifespan of our production use case.

Left a couple of comments, just worried about the memory cost increasing.

The main motivation I introduced bloom filter strategy is just reduce record all received chunk.

So there are two suggested option.
1. Add this feature to examples/chunk_statistics/analyze_chunk_hashes.py, so that we can parse offline recorded request files without affect production runtime LMCache.
2. Move the logic of this chunk lifespan into a new sub class, let's say memory_bf_lifespan which may inherit from memory_bloom_filter strategy.

YuhanLiu11:
> @YuhanLiu11 Thanks for this great feature, it is very useful, it can help us insight the chunk lifespan of our production use case.
> 
> Left a couple of comments, just worried about the memory cost increasing.
> 
> The main motivation I introduced bloom filter strategy is just reduce record all received chunk.
> 
> So there are two suggested option.
> 
> 1. Add this feature to examples/chunk_statistics/analyze_chunk_hashes.py, so that we can parse offline recorded request files without affect production runtime LMCache.
> 2. Move the logic of this chunk lifespan into a new sub class, let's say memory_bf_lifespan which may inherit from memory_bloom_filter strategy.

@maobaolong thanks for these comments! Can you elaborate a bit how to add this feature in the offline processing script analyze_chunk_hashes.py? I thought each chunk doesn't have the earliest access timestamp and latest timestamp right?

maobaolong:
@YuhanLiu11 file_hash strategy can record the chunks for each request with timestamp, so we are able to get the earliest access time and last access time through parse and handle the recorded file line by line. 

```python
data = {
                "timestamp": time.time(),
                "lookup_id": lookup_id,
                "chunk_hashes": chunk_hashes,
            }
```

YuhanLiu11:
Closing this PR now, will have a new one which has fewer changes in code.

PR #2044, opened 2025-11-20
[1/3]: Refactor S3 Connector Configuration and Initialization

### LINKED ISSUE #1961: [bug] s3_connector failed to initialize
**Describe the bug**
Failed to initialize/re-establish remote connection: 'NoneType' object cannot be interpreted as an integer (remote_backend.py:129:lmcache.v1.storage_backend.remote_backend)

https://github.com/LMCache/LMCache/blob/e5ab00abd69c8616c99f9e285de36df8227c2c72/lmcache/v1/storage_backend/connector/s3_connector.py#L128 always fails because self.full_chunk_size is never set.

Setting s3_part_size will fail the assertion and not setting gives the following error: 'NoneType' object cannot be interpreted as an integer 

**To Reproduce**
Follow the config in this doc: https://docs.lmcache.ai/kv_cache/storage_backends/s3.html

**Expected behavior**
Initialize remote connection with s3 connector

sammshen:
FIX https://github.com/LMCache/LMCache/issues/1961 @ksuma2109

This PR fixes the initialization logic and renames / adds some configs such as `num_io_threads` and `disable_tls` and enforcing the `s3_part_size` properly as well as doing type validation.

The follow up PRs are: 
[2/3]: Add proper error validation to the S3Connector
[3/3]: create zero copy semantics (currently there is a copy from `/dev/shm` to cudaHostRegistered memory but we should be able to remove this).

### DIFF
--- a/lmcache/v1/storage_backend/connector/s3_adapter.py
+++ b/lmcache/v1/storage_backend/connector/s3_adapter.py
@@ -23,21 +23,22 @@ class S3ConnectorAdapter(ConnectorAdapter):
         config = context.config
         assert config is not None
 
-        if config.extra_config is not None:
-            # Different parts can be transferred in parallel.
-            self.s3_part_size = config.extra_config.get("s3_part_size", None)
-            self.s3_max_io_concurrency = config.extra_config.get(
-                "s3_max_io_concurrency", 64
-            )
-            self.s3_max_inflight_reqs = config.extra_config.get(
-                "s3_max_inflight_reqs", 64
-            )
-            self.s3_prefer_http2 = config.extra_config.get("s3_prefer_http2", True)
-            self.s3_region = config.extra_config.get("s3_region", None)
-            self.s3_enable_s3express = config.extra_config.get(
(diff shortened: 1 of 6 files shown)


lmcache/v1/storage_backend/connector/s3_adapter.py
```diff
@@ -23,21 +23,22 @@ def create_connector(self, context: ConnectorContext) -> RemoteConnector:
...
-            self.s3_max_inflight_reqs = config.extra_config.get(
-                "s3_max_inflight_reqs", 64
-            )
-            self.s3_prefer_http2 = config.extra_config.get("s3_prefer_http2", True)
-            self.s3_region = config.extra_config.get("s3_region", None)
-            self.s3_enable_s3express = config.extra_config.get(
-                "s3_enable_s3express", True
-            )
-            self.s3_file_prefix = config.extra_config.get("s3_file_prefix", None)
+        # Get full_chunk_size from context
```
sammshen, line 26: we should do type validation here

PR #2045, opened 2025-11-20
Handle aborted requests in async lookup client and server.

yoo-kumaneko:
1. The adapter on the scheduler side receives finished requests from the scheduler, and passes the aborted requests to the async lookup client.
2. The async lookup client keeps a record of aborted requests.
3. The async lookup client constantly checks if aborted requests are done loading. Pick those that are done loading and send cleanup messages for them to the async lookup servers.
4. The async lookup servers get the aborted requests and safely count down their memory objects.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1662,8 +1662,9 @@ class LMCacheConnectorV1Impl:
         # Cleanup if request was aborted
         if request.status == RequestStatus.FINISHED_ABORTED and self.async_loading:
             # Cancel any ongoing async lookup and prefetch tasks on workers
+            lookup_id = request.request_id
             self.lookup_client.cancel_lookup(  # type: ignore[attr-defined]
-                request.request_id
+                lookup_id
             )
 
         params = (
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -20,6 +20,7 @@ from vllm.distributed.parallel_state import (
     get_tp_group,
 )
 from vllm.sampling_params import SamplingParams
+from vllm.v1.request import RequestStatus
 
 # First Party
 # Use LMCache's own math utilities instead of vllm's
@@ -1658,6 +1659,11 @@ class LMCacheConnectorV1Impl:
         request: "Request",
         block_ids: list[int],
     ) -> tuple[bool, Optional[dict[str, Any]]]:
+        # Cleanup if request was aborted
+        if request.status == RequestStatus.FINISHED_ABORTED and self.async_loading:
+            # Cancel any ongoing async lookup and prefetch tasks on workers
+            self.lookup_client.cancel_lookup(request.request_id)
+
         params = (
             request.kv_transfer_params
             if hasattr(request, "kv_transfer_params")
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1662,7 +1662,9 @@ class LMCacheConnectorV1Impl:
         # Cleanup if request was aborted
         if request.status == RequestStatus.FINISHED_ABORTED and self.async_loading:
             # Cancel any ongoing async lookup and prefetch tasks on workers
-            self.lookup_client.cancel_lookup(request.request_id)
+            self.lookup_client.cancel_lookup(  # type: ignore[attr-defined]
+                request.request_id
+            )
 
         params = (
             request.kv_transfer_params
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -26,7 +26,7 @@ from lmcache.observability import LMCacheStatsLogger, LMCStatsMonitor
 from lmcache.usage_context import InitializeUsageContext
 from lmcache.utils import CacheEngineKey, _lmcache_nvtx_annotate
 from lmcache.v1.config import LMCacheEngineConfig
-from lmcache.v1.event_manager import EventManager, EventType
+from lmcache.v1.event_manager import EventManager, EventStatus, EventType
 from lmcache.v1.gpu_connector import (
     GPUConnectorInterface,
     SGLangLayerwiseGPUConnector,
@@ -933,6 +933,42 @@ class LMCacheEngine:
             ),
             self.storage_manager.loop,
         )
+
+    def cleanup_memory_objs(self, lookup_id: str) -> None:
+        """
+        Cleanup memory objects allocated during prefetch for an aborted lookup.
+
+        Called by the scheduler when it determines that an aborted lookup
+        has finished its prefetch tasks.
+        """
+        try:
+            # Get the completed future from event_manager
+            if (
+                self.event_manager.get_event_status(EventType.LOADING, lookup_id)
+                != EventStatus.DONE
+            ):
+                logger.debug(
+                    f"No completed event found for lookup_id={lookup_id} to clean up."
+                )
+                return
+            future = self.event_manager.pop_event(EventType.LOADING, lookup_id)
+
+            # Get memory objects from the future result
+            memory_objs = future.result()
+            # Flatten nested lists (each backend returns a list of chunks)
+            memory_objs_flat = [mm for m in memory_objs for mm in m]
+
+            # Release each memory object
+            for memory_obj in memory_objs_flat:
+                try:
+                    logger.debug(f"Releasing memory object for lookup_id={lookup_id}")
+                    memory_obj.ref_count_down()
+                except Exception as e:
+                    logger.error(f"Error releasing memory object: {e}")
+        except Exception as e:
+            logger.error(
+                f"Error during cleanup_memory_objs for lookup_id={lookup_id}: {e}"
+            )
 
     # TODO(Jiayi): Need to handle the case where `tokens=None`.
     # In this case, we compress all tokens.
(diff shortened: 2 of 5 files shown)


lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -221,6 +228,34 @@ def clear_lookup_status(self, lookup_id: str) -> None:
         with self.lock:
             self.reqs_status.pop(lookup_id, None)
 
+    def cancel_lookup(self, lookup_id: str) -> None:
+        """Mark lookup as aborted. Cleanup will happen after task finishes."""
+        self.aborted_lookups.add(lookup_id)
```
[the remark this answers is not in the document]
yoo-kumaneko: Don't need a lock at all.

yoo-kumaneko:
@ApostaC Could you take a look at this PR? It adds cleanup handling for aborted requests that are being async looked up.

yoo-kumaneko:
@ziruiliu @sammshen Guys, would you like to take a look at this PR?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1658,6 +1659,13 @@ def request_finished(
         request: "Request",
         block_ids: list[int],
     ) -> tuple[bool, Optional[dict[str, Any]]]:
+        # Cleanup if request was aborted
+        if request.status == RequestStatus.FINISHED_ABORTED and self.async_loading:
+            # Cancel any ongoing async lookup and prefetch tasks on workers
+            self.lookup_client.cancel_lookup(  # type: ignore[attr-defined]
+                request.request_id
```
sammshen: nit: 
lookup_id = request.request_id

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -313,7 +349,7 @@ def process_requests_from_scheduler(self):
                 msg_buf = self.pull_socket.recv(copy=False)
                 msg = msgspec.msgpack.decode(
                     msg_buf,
-                    type=LookupRequestMsg,
+                    type=Union[LookupRequestMsg, LookupCleanupMsg],
```
sammshen, line 357: nit, add a small comment here that we are relying on # msgspec.msgpack.decode to discriminate between messages here

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -247,6 +254,35 @@ def clear_lookup_status(self, lookup_id: str) -> None:
             self.reqs_status.pop(lookup_id, None)
             self.first_lookup_time.pop(lookup_id, None)
 
+    def cancel_lookup(self, lookup_id: str) -> None:
+        """Mark lookup as aborted. Cleanup will happen after task finishes."""
+        self.aborted_lookups.add(lookup_id)
+
+    def _cleanup_finished_aborted_lookups(self) -> None:
```
sammshen, line 263: why not just move the logic for `_cleanup_finished_aborted_lookups(self)` into `cancel_lookup()`?

why do we need to maintain this extra aborted_lookups list and then call the actual abortion during lookup?

sammshen:
LGTM, just a few questions, please address

lmcache/v1/cache_engine.py
```diff
@@ -934,6 +934,42 @@ def async_lookup_and_prefetch(
...
+
+            # Get memory objects from the future result
+            memory_objs = future.result()
+            # Flatten nested lists (each backend returns a list of chunks)
+            memory_objs_flat = [mm for m in memory_objs for mm in m]
+
+            # Release each memory object
+            for memory_obj in memory_objs_flat:
+                try:
+                    logger.debug(f"Releasing memory object for lookup_id={lookup_id}")
```
KuntaiDu: Prefer `logger.debug("xxxx %d", lookup_id)`. This is because writing f-string forces Python interpreter to serialize lookup_id, even when this log won't show up (e.g. when the log level is info).

lmcache/v1/cache_engine.py
```diff
@@ -934,6 +934,42 @@ def async_lookup_and_prefetch(
...
+                != EventStatus.DONE
+            ):
+                logger.debug(
+                    f"No completed event found for lookup_id={lookup_id} to clean up."
+                )
+                return
+            future = self.event_manager.pop_event(EventType.LOADING, lookup_id)
+
+            # Get memory objects from the future result
+            memory_objs = future.result()
```
KuntaiDu, line 957: This seems to be a blocking wait. Will this slow down the cache-engine significantly?

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -247,6 +254,35 @@ def clear_lookup_status(self, lookup_id: str) -> None:
...
+            lookup_id
+            for lookup_id in self.aborted_lookups
+            if self.reqs_status.get(lookup_id) is not None
+        ]
+        if finished_lookups:
+            self.aborted_lookups.difference_update(finished_lookups)
+            with self.lock:
+                for lookup_id in finished_lookups:
+                    self.reqs_status.pop(lookup_id, None)
+
```
KuntaiDu, line 274: Maybe a comment here saying that the message aim to notify server to free reserved memory buffer?

KuntaiDu (commented):
See the comments.

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: done

lmcache/v1/cache_engine.py
yoo-kumaneko: done

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: done

lmcache/integration/vllm/vllm_v1_adapter.py
yoo-kumaneko: done

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: When `cancel_lookup()` is called, the request may still be in the async loading phase. We can’t halt that immediately, so we wait for the loading to finish before cleaning up resources. That’s why the cleanup is deferred.

lmcache/v1/cache_engine.py
yoo-kumaneko: The future here is at the DONE state. The result is ready and will be returned immediately.

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -249,6 +256,36 @@ def clear_lookup_status(self, lookup_id: str) -> None:
...
+    def cancel_lookup(self, lookup_id: str) -> None:
+        """Mark lookup as aborted. Cleanup will happen after task finishes."""
+        self.aborted_lookups.add(lookup_id)
+
+    def _cleanup_finished_aborted_lookups(self) -> None:
+        """Check for finished aborted lookups and send cleanup messages to workers."""
+        finished_lookups = [
+            lookup_id
+            for lookup_id in self.aborted_lookups
+            if self.reqs_status.get(lookup_id) is not None
```
sammshen, line 270: nit: can we add a comment that lookup_id is None means still loading

sammshen: also we do need to protect any accesses to req_status with a lock such as this here

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -249,6 +256,36 @@ def clear_lookup_status(self, lookup_id: str) -> None:
...
+        ]
+        if finished_lookups:
+            self.aborted_lookups.difference_update(finished_lookups)
```
sammshen, line 273: just a design question, do we want a configrmation from the lookup server / cache engine that the clear was completed? right now, for example, the loading event could not be DONE?

sammshen:
left another round of comment @yoo-kumaneko looks pretty good!

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: Good question. We need to protect the `req_status` dict because the `process_responses_from_workers` thread collects responses from workers and updates `req_status`. As you can see, the get operation isn’t protected by the lock, while pop is protected as it is in the `clear_lookup_status` method. Like below.
```
def clear_lookup_status(self, lookup_id: str) -> None:
    with self.lock:
        self.reqs_status.pop(lookup_id, None)
        self.first_lookup_time.pop(lookup_id, None)
```
In fact, I should call this `clear_lookup_status` as well, instead of directly popping the `req_status`.

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
yoo-kumaneko: The async lookup client waits until the loading event reaches DONE before performing cleanup. At that point, we can safely assume the loading phase has completed, so collecting the memory objects shouldn’t pose any issues. An explicit confirmation from the workers isn’t necessary.

lmcache/v1/cache_engine.py
KuntaiDu: Oh I see

KuntaiDu (approved):
LGTM!

sammshen (approved):
LGTM, this is a great fix!

PR #2047, opened 2025-11-20
[TEST|P2P] Add py_socket channel and P2PBackend test

maobaolong:

- Add py_socket channel.
- Add P2PBackend test.

```
┌─────────────┐         ┌─────────────┐
│   Peer 1    │         │   Peer 2    │
│  (Worker 0) │         │  (Worker 1) │
└──────┬──────┘         └──────┬──────┘
       │                       │
       │    ┌─────────────┐    │
       └────┤ Controller  ├────┘
            └─────────────┘
```

- Tests use available ports to avoid conflicts
- Memory allocator size must be multiple of align_bytes (1120 * 1024 * 1024 bytes)

**If applicable**:


lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,412 @@
...
+        self.side_channels: list[zmq.Socket] = []
+        self.running_threads: list[threading.Thread] = []
+
+        self.async_mode = async_mode
+        if self.async_mode:
+            self.zmq_context = get_zmq_context(use_asyncio=True)
+        else:
+            self.zmq_context = get_zmq_context(use_asyncio=False)
+        self.peer_init_url = kwargs["peer_init_url"]
+        self.event_loop = kwargs.get("event_loop", None)
```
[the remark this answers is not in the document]
maobaolong: Keep code consistent with nixl_channel

ApostaC:
Is this because we need to run the p2p backend unit test without Nixl?

maobaolong:
> Is this because we need to run the p2p backend unit test without Nixl?

@ApostaC Yeah

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,412 @@
...
+        py_socket_init_resp_bytes = await init_tmp_socket.recv()
+        py_socket_init_resp = msgspec.msgpack.decode(
+            py_socket_init_resp_bytes, type=PySocketMsg
+        )
+        remote_id = py_socket_init_resp.remote_id
+
+        # Register remote memory
+        py_socket_mem_reg_req = PySocketMemRegRequest(
+            remote_id=remote_id,
+            local_id=local_id,
```
ApostaC: Where else do we use `local_id` and `remote_id` other than during initialization?

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,412 @@
...
+
+        # Store connection info
+        self.remote_connections[peer_id] = {
+            "remote_id": remote_id,
+            "peer_init_url": peer_init_url,
+        }
+
+        # Initialize remote data store for this peer
+        if peer_id not in _GLOBAL_DATA_STORE:
+            _GLOBAL_DATA_STORE[peer_id] = {}
```
ApostaC: From the p2p backend, the peer_id is actually peer_init_url. Will it create any issues
cc @YaoJiayi 

https://github.com/LMCache/LMCache/blob/ef34e994d703d93b0f88a4d4c87f12749cec2441/lmcache/v1/storage_backend/p2p_backend.py#L331-L336

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,412 @@
...
+
+    def _init_loop(self):
+        # Initialize initialization side channels
+        self.init_side_channel = get_zmq_socket(
+            self.zmq_context,
+            self.peer_init_url,
+            "tcp",
+            zmq.REP,
+            "bind",
+        )
```
ApostaC: This part of the code is duplicated in `_async_init_loop`. We can move it to `_init_side_channels` function.

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,412 @@
...
+            resp = self.handle_init_side_msg(req)
+            logger.info("Replying P2P init side response")
+        else:
+            raise ValueError(f"Unsupported InitMsg type: {type(req)}")
+
+        return resp
+
+    def _init_loop(self):
+        # Initialize initialization side channels
+        self.init_side_channel = get_zmq_socket(
```
ApostaC: nit: `self.init_side_channel` sounds like a method (actually, we do have a method with a very similar name `self._init_side_channels`.
Let's change the field name? Do you think `self.side_channel` is good enough?

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,412 @@
...
+        for obj, remote_idx in zip(objects, remote_indexes, strict=True):
+            if isinstance(obj, MemoryObj) and obj.tensor is not None:
+                # Store a copy of the tensor data
+                if receiver_id not in _GLOBAL_DATA_STORE:
+                    _GLOBAL_DATA_STORE[receiver_id] = {}
+                _GLOBAL_DATA_STORE[receiver_id][remote_idx] = obj.tensor.clone()
+
+        # Simulate transfer delay
+        await asyncio.sleep(0.001)
+        return len(objects)
```
ApostaC: So we are not really send/recv data now?

ApostaC: If so, should we rename the class to something else?

ApostaC (commented):
Done a pass. Please see below

tests/v1/storage_backend/test_p2p_backend_with_controller.py
ApostaC, line 1: Haven't checked the details in the unit test implementation. But one high-level advice is that we should try to avoid using any implementation details in each class, since they could be changed in the future.

maobaolong:
@ApostaC Thanks for your valuable comments!
I did a small refactor, and abstract control pane to `py_socket_channel.py` which is created by jiayi, but I create another simple test purpose `MockMemoryChannel` extended from `PySocketChannel`, and put some data pane code there.

The data pane of `MockMemoryChannel` leveraged by a global dict `_GLOBAL_DATA_STORE`, so you are right, this didn't transfer data really, and I renamed it to `MockMemoryChannel` and add docstring to introduce this is a testing purpose implementation.

In fact I didn't implemented the data pane for `PySocketChannel`, since I don't need it yet, left a  TODO comment to the head of `py_socket_channel.py` 

PTAL.

maobaolong:
Hi @Shaoting-Feng Is it possible that this failed test related to this PR?
```

=========================================================================== FAILURES ===========================================================================
--
____________________________________________________________________ test_force_store_wait _____________________________________________________________________
 
autorelease_v1 = <function autorelease_v1.<locals>._factory at 0x7f2d1416b1a0>
 
@pytest.mark.no_shared_allocator
@pytest.mark.skipif(
not torch.cuda.is_available(),
reason="TODO: Add non-CUDA implementation to VLLMPagedMemGPUConnectorV2",
)
def test_force_store_wait(autorelease_v1):
device = "cuda"
fmt = "vllm"
num_tokens = 10000
num_blocks = 5000
block_size = 16
dtype = torch.bfloat16
 
chunk_size = 256
kv_shape = (32, 2, chunk_size, 8, 128)
 
connector = create_gpu_connector(1024, 32)
 
kv_cache = generate_kv_cache_paged_list_tensors(
num_blocks, device, block_size, dtype
)
 
num_requests = 8
 
def generate_random_slot_mapping(num_blocks, block_size, num_tokens, device):
slot_mapping = random.sample(range(0, num_blocks * block_size), num_tokens)
return torch.tensor(slot_mapping, device=device)
 
list_tokens = [generate_tokens(num_tokens, device) for _ in range(num_requests)]
list_slot_mappings = [
generate_random_slot_mapping(num_blocks, block_size, num_tokens, device)
for _ in range(num_requests)
]
 
homedir = os.environ.get("HOME", "/tmp")
with tempfile.TemporaryDirectory(
dir=homedir, ignore_cleanup_errors=True
) as temp_dir:
cfg = LMCacheEngineConfig.from_defaults(
local_cpu=False,
max_local_cpu_size=2,  # small cpu buffer
local_disk=temp_dir,
max_local_disk_size=20,
extra_config={"force_store_wait": True},
)
 
engine = autorelease_v1(
LMCacheEngineBuilder.get_or_create(
"test",
cfg,
dumb_metadata(fmt, kv_shape),
connector,
mock_up_broadcast_fn,
mock_up_broadcast_object_fn,
)
)
 
# Store kv cache into slow devices
for t, s in zip(list_tokens, list_slot_mappings, strict=False):
engine.store(t, kvcaches=kv_cache, slot_mapping=s)
 
# Sleep 10 seconds for the last request
time.sleep(10)
 
# No KV cache should be skipped
for t in list_tokens:
>               assert engine.lookup(t) == len(t)
E               AssertionError: assert 9216 == 10000
E                +  where 9216 = lookup(tensor([2523, 2726, 5953,  ..., 1978, 6406, 6317], device='cuda:0'))
E                +    where lookup = <lmcache.v1.cache_engine.LMCacheEngine object at 0x7f2d15db7830>.lookup
E                +  and   10000 = len(tensor([2523, 2726, 5953,  ..., 1978, 6406, 6317], device='cuda:0'))
 
tests/v1/test_cache_engine.py:1206: AssertionError


```

Shaoting-Feng:
> Hi @Shaoting-Feng Is it possible that this failed test related to this PR?
> 
> ```
> 
> =========================================================================== FAILURES ===========================================================================
> --
> ____________________________________________________________________ test_force_store_wait _____________________________________________________________________
>  
> autorelease_v1 = <function autorelease_v1.<locals>._factory at 0x7f2d1416b1a0>
>  
> @pytest.mark.no_shared_allocator
> @pytest.mark.skipif(
> not torch.cuda.is_available(),
> reason="TODO: Add non-CUDA implementation to VLLMPagedMemGPUConnectorV2",
> )
> def test_force_store_wait(autorelease_v1):
> device = "cuda"
> fmt = "vllm"
> num_tokens = 10000
> num_blocks = 5000
> block_size = 16
> dtype = torch.bfloat16
>  
> chunk_size = 256
> kv_shape = (32, 2, chunk_size, 8, 128)
>  
> connector = create_gpu_connector(1024, 32)
>  
> kv_cache = generate_kv_cache_paged_list_tensors(
> num_blocks, device, block_size, dtype
> )
>  
> num_requests = 8
>  
> def generate_random_slot_mapping(num_blocks, block_size, num_tokens, device):
> slot_mapping = random.sample(range(0, num_blocks * block_size), num_tokens)
> return torch.tensor(slot_mapping, device=device)
>  
> list_tokens = [generate_tokens(num_tokens, device) for _ in range(num_requests)]
> list_slot_mappings = [
> generate_random_slot_mapping(num_blocks, block_size, num_tokens, device)
> for _ in range(num_requests)
> ]
>  
> homedir = os.environ.get("HOME", "/tmp")
> with tempfile.TemporaryDirectory(
> dir=homedir, ignore_cleanup_errors=True
> ) as temp_dir:
> cfg = LMCacheEngineConfig.from_defaults(
> local_cpu=False,
> max_local_cpu_size=2,  # small cpu buffer
> local_disk=temp_dir,
> max_local_disk_size=20,
> extra_config={"force_store_wait": True},
> )
>  
> engine = autorelease_v1(
> LMCacheEngineBuilder.get_or_create(
> "test",
> cfg,
> dumb_metadata(fmt, kv_shape),
> connector,
> mock_up_broadcast_fn,
> mock_up_broadcast_object_fn,
> )
> )
>  
> # Store kv cache into slow devices
> for t, s in zip(list_tokens, list_slot_mappings, strict=False):
> engine.store(t, kvcaches=kv_cache, slot_mapping=s)
>  
> # Sleep 10 seconds for the last request
> time.sleep(10)
>  
> # No KV cache should be skipped
> for t in list_tokens:
> >               assert engine.lookup(t) == len(t)
> E               AssertionError: assert 9216 == 10000
> E                +  where 9216 = lookup(tensor([2523, 2726, 5953,  ..., 1978, 6406, 6317], device='cuda:0'))
> E                +    where lookup = <lmcache.v1.cache_engine.LMCacheEngine object at 0x7f2d15db7830>.lookup
> E                +  and   10000 = len(tensor([2523, 2726, 5953,  ..., 1978, 6406, 6317], device='cuda:0'))
>  
> tests/v1/test_cache_engine.py:1206: AssertionError
> ```

Yes

maobaolong:
@Shaoting-Feng Thanks a lot, will dig deeply.

maobaolong:
@Shaoting-Feng I ran the tests locally and passed. I skipped the `test_gds_backend.py` and `test_gpu_connector.py`
```
CUDA_LAUNCH_BLOCKING=1 LMCACHE_TRACK_USAGE="false" pytest --maxfail=1 --cov=lmcache   --cov-report term --cov-report=html:coverage-test   --cov-report=xml:coverage-test.xml --html=durations/test.html   --ignore=tests/disagg --ignore=tests/v1/test_pos_kernels.py   --ignore=tests/v1/test_nixl_storage.py --ignore=tests/v1/storage_backend/test_gds_backend.py --ignore=tests/v1/test_gpu_connector.py 

======================= 303 passed, 12 skipped, 64 warnings in 642.80s (0:10:42) =======================
buildkite-test-collector - WARNING - No BUILDKITE_ANALYTICS_TOKEN environment variable present
[2025-11-22 12:21:07,317] LMCache WARNING: original_connector is None, re-initializing connection. (remote_monitor.py:111:lmcache.v1.storage_backend.remote_monitor)
[2025-11-22 12:21:07,317] LMCache INFO: Connector <MockConnector> does not support ping, break RemoteMonitor thread. (remote_monitor.py:119:lmcache.v1.storage_backend.remote_monitor)
[2025-11-22 12:21:07,327] LMCache WARNING: original_connector is None, re-initializing connection. (remote_monitor.py:111:lmcache.v1.storage_backend.remote_monitor)
[2025-11-22 12:21:07,327] LMCache INFO: Connector <MockConnector> does not support ping, break RemoteMonitor thread. (remote_monitor.py:119:lmcache.v1.storage_backend.remote_monitor)
[W1122 12:21:58.535573666 CudaIPCTypes.cpp:16] Producer process has been terminated before all shared CUDA tensors released. See Note [Sharing CUDA tensors]
```

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,260 @@
...
+        assert "buffer_size" in kwargs
+        assert "align_bytes" in kwargs
+        assert "tp_rank" in kwargs
+        assert "peer_init_url" in kwargs
+
+        self.role = kwargs["role"]
+        self.buffer_ptr = kwargs["buffer_ptr"]
+        self.buffer_size = kwargs["buffer_size"]
+        self.align_bytes = kwargs["align_bytes"]
+        self.tp_rank = kwargs["tp_rank"]
```
ApostaC, line 81: Are these parameters needed by `PySocketChannel`? If not, we can remove them.
(The subclasses should do their own checks if they need those params)

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,260 @@
...
+
+        self.peer_lookup_url = kwargs.get("peer_lookup_url", None)
```
ApostaC, line 83: This one seems not being used as well

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,260 @@
...
+            "tcp",
+            zmq.REQ,
+            "connect",
+        )
+
+        mock_memory_init_req = PySocketInitRequest(peer_init_url=self.peer_init_url)
+        await init_tmp_socket.send(msgspec.msgpack.encode(mock_memory_init_req))
+
+        mock_memory_init_resp_bytes = await init_tmp_socket.recv()
+        _ = msgspec.msgpack.decode(mock_memory_init_resp_bytes, type=PySocketMsg)
```
ApostaC: nit: variables used in the base class should not use `mock_memory_xxx`

lmcache/v1/transfer_channel/py_socket_channel.py
```diff
@@ -1,2 +1,260 @@
...
+                "peer_init_url": peer_url,
+            }
+
+            self._on_peer_connected(peer_url)
+
+            resp = PySocketInitResponse(status="ok")
+            logger.info("Replying initialization response")
+
+        elif isinstance(req, InitSideMsgBase):
+            resp = self.handle_init_side_msg(req)
```
ApostaC, line 192: Where is `handle_init_side_msg` defined? If we want subclasses to extend this class, we need to define it as an abstract method in `PySocketChannel`

ApostaC (commented):
Thanks for the update!
Just some nit comments.

lmcache/v1/transfer_channel/py_socket_channel.py
maobaolong: Just copy it from nixl_channel.py, it is used by parent class `BaseTransferChannel`,  the `self.peer_lookup_url` of `NixlChannel` did the same thing too.

I refactored this

lmcache/v1/transfer_channel/py_socket_channel.py
maobaolong: Renamed

lmcache/v1/transfer_channel/py_socket_channel.py
maobaolong: I abstract this into lmcache/v1/transfer_channel/transfer_utils.py

maobaolong:
@ApostaC Thanks for your review again, i found there are some confusing code between `BaseTransferChannel` and `NixlChannel` and did a refactor, PTAL.

lmcache/v1/transfer_channel/abstract.py
```diff
@@ -56,28 +54,6 @@ async def async_lazy_init_peer_connection(
...
-        if isinstance(req, P2PInitSideMsg):
-            assert hasattr(self, "peer_lookup_url"), (
-                "P2PInitSideMsg requires `peer_lookup_url` attribute."
-            )
-            return P2PInitSideRetMsg(
-                peer_lookup_url=self.peer_lookup_url,
-            )
-        else:
-            raise ValueError(f"Unsupported InitSideMsg type: {type(req)}")
-
```
ApostaC, line 80: Sorry for the back and forth... I just realized that `handle_init_side_msg` was defined in the base class. 

I think the old design makes sense because different subclasses may have different operations on the "side msg". In this case, let subclasses override this function makes more sense.

Can you maybe revert the edits here? Sorry again for the confusion 🙏

ApostaC: One irrelevant follow-up: Does Python have any annotations for "override"? This will potentially make code more readable when we are overriding functions that have default implementations in the base class.

ApostaC (approved):
Thanks @maobaolong . Just one minor thing. Otherwise LGTM!

lmcache/v1/transfer_channel/abstract.py
maobaolong: Done

maobaolong:
> Sorry for the back and forth... I just realized that `handle_init_side_msg` was defined in the base class.
> 
> I think the old design makes sense because different subclasses may have different operations on the "side msg". In this case, let subclasses override this function makes more sense.
> 
> Can you maybe revert the edits here? Sorry again for the confusion 🙏

@ApostaC I revert the change related to `handle_init_side_msg`. 

For the second comment (about override annotation):
Yes! Python has `@override` decorator from typing module (Python 3.12+) or typing_extensions (for earlier versions). It helps type checkers verify that a method actually overrides a base class method. We could consider create a demonstrate PR , so that contributors can following this practice.

ApostaC:
> Yes! Python has `@override` decorator from typing module (Python 3.12+) or typing_extensions (for earlier versions). It helps type checkers verify that a method actually overrides a base class method. We could consider create a demonstrate PR , so that contributors can following this practice.

Thanks! I think we are still supporting python 3.10 and 3.11 (since vLLM also has the support). Maybe we can write a simple annotation function ourselves if early Python verions don't have this annotation.
Anyway, I think it's out of this PR's scope anyway.

ApostaC (approved):
LGTM!

chunxiaozheng (approved):
lgtm

maobaolong:
@ApostaC @Shaoting-Feng 

The full comprehensive test is failed previous 3 time retry.

```
Expected warmup latency: 3.6283979535102846
Actual warmup latency: 4.288712763786316

Expected warmup latency: 3.6283979535102846
Actual warmup latency: 4.291113018989563


Expected warmup latency: 3.6283979535102846
Actual warmup latency: 4.134437596797943

```

PR #2049, opened 2025-11-20
[2/3]: Secure error handling and resource leaks in S3 Connector

sammshen:
Follow up to:  https://github.com/LMCache/LMCache/pull/2044

[3/3]: Register Local CPU Backend with `/dev/shm`

UPDATE: the memmove from /dev/shm to local cpu backend only takes ~0.02 seconds, which is not worth the refactor

### DIFF
--- a/lmcache/v1/storage_backend/connector/s3_connector.py
+++ b/lmcache/v1/storage_backend/connector/s3_connector.py
@@ -446,7 +446,7 @@ class S3Connector(RemoteConnector):
     def on_get_done(
         self,
         obj_size: int,
-        memory_obj: Union[MemoryObj, None],
+        memory_obj: Optional[MemoryObj],
         shm: Union[int, None],
         recv_path: str,
         _: asyncio.Future,
--- a/lmcache/v1/storage_backend/connector/s3_connector.py
+++ b/lmcache/v1/storage_backend/connector/s3_connector.py
@@ -2,7 +2,7 @@
 # Standard
 from enum import IntEnum, auto
 from functools import partial
-from typing import List, Optional
+from typing import List, Optional, Union
 from urllib.parse import quote as url_quote
 import asyncio
 import ctypes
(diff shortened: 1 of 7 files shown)


lmcache/v1/storage_backend/connector/s3_connector.py
```diff
@@ -403,38 +416,44 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
 
-        await asyncio.wrap_future(s3_req.finished_future)
-
-        dst_ptr = memory_obj.data_ptr
-        ctypes.memmove(dst_ptr, shm, obj_size)
-
-        self.adhoc_shm_manager.free(recv_path, shm)
-
-        self.inflight_sema.release()
+        try:
```
sammshen, line 429: only this try except is associated with this PR, the other changes are associated with [1/3]

lmcache/v1/storage_backend/connector/s3_connector.py
```diff
@@ -403,38 +416,44 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
+            self.adhoc_shm_manager.free(recv_path, shm)
 
     # this callback allows us to safely have multiple calls to batched_get
     # since we release the semaphores 1-by-1
     def on_get_done(
         self,
         obj_size: int,
-        memory_obj: MemoryObj,
-        shm: int,
+        memory_obj: Union[MemoryObj, None],
```
sammshen: changes here are also from this PR

YaoJiayi: Why not use Optional type here?

lmcache/v1/storage_backend/connector/s3_connector.py
```diff
@@ -403,38 +426,43 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
+        shm: Union[int, None],
         recv_path: str,
-        fut: asyncio.Future,
+        _: asyncio.Future,
```
YaoJiayi, line 452: What is `_`?

YaoJiayi (commented):
LGTM in general. Just a few comments.

lmcache/v1/storage_backend/connector/s3_connector.py
sammshen: the future must be in the signature of the callback but is not used

lmcache/v1/storage_backend/connector/s3_connector.py
sammshen: good catch, thanks!

lmcache/v1/storage_backend/connector/s3_connector.py
YaoJiayi: Using `_` in as a argument is a bit weird:)

YaoJiayi (approved):
lgtm

PR #2055, opened 2025-11-21
[Controller]Add metrics system to controller

maobaolong:

Add PrometheusLogger for controller role, and add some metrics as example for other contributors to ease to contribute.

Clipboard_Screenshot_1763713435

### DIFF
--- a/lmcache/v1/cache_controller/controller_manager.py
+++ b/lmcache/v1/cache_controller/controller_manager.py
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
-from enum import Enum
 from typing import Optional
 import asyncio
 import json
@@ -15,7 +14,11 @@ import zmq
 from lmcache.logging import init_logger
 from lmcache.v1.cache_controller.controllers import KVController, RegistrationController
 from lmcache.v1.cache_controller.executor import LMCacheClusterExecutor
-from lmcache.v1.cache_controller.observability import PrometheusLogger
+from lmcache.v1.cache_controller.observability import (
+    PrometheusLogger,
+    SocketMetricsContext,
+    SocketType,
+)
 from lmcache.v1.rpc_utils import (
     get_zmq_context,
     get_zmq_socket,
@@ -53,13 +56,6 @@ logger = init_logger(__name__)
 # TODO(Jiayi): Need to align the message types. For example,
 # a controller should take in an control message and return
 # a control message.
-
-
-class SocketType(Enum):
-    """Enum for socket types to ensure type safety."""
-
-    PULL = "pull"
-    REPLY = "reply"
 
 
 class LMCacheControllerManager:
@@ -138,12 +134,6 @@ class LMCacheControllerManager:
         # Setup socket message count metrics
         self._setup_socket_metrics()
 
-        # self.loop = asyncio.new_event_loop()
-        # self.thread = threading.Thread(target=self.loop.run_forever,
-        #                               daemon=True)
-        # self.thread.start()
-        # asyncio.run_coroutine_threadsafe(self.start_all(), self.loop)
-
     async def handle_worker_message(self, msg: WorkerMsg) -> None:
         if isinstance(msg, HeartbeatMsg):
             await self.reg_controller.heartbeat(msg)
@@ -189,39 +179,6 @@ class LMCacheControllerManager:
         else:
             logger.error(f"Unknown orchestration message type: {msg}")
             raise RuntimeError(f"Unknown orchestration message type: {msg}")
-
-    class _SocketMetricsContext:
-        """Context manager for socket message counting and error handling."""
-
-        def __init__(self, manager, socket_type: SocketType, message_count: int = 1):
-            self.manager = manager
-            self.socket_type = socket_type
-            self.message_count = message_count
-            self.counter_attr = f"{socket_type.value}_socket_message_count"
-            self.active_attr = f"{socket_type.value}_socket_active_requests"
-
-        def __enter__(self):
-            setattr(
-                self.manager,
-                self.counter_attr,
-                getattr(self.manager, self.counter_attr) + self.message_count,
-            )
-            setattr(
-                self.manager,
-                self.active_attr,
-                getattr(self.manager, self.active_attr) + self.message_count,
-            )
-            return self
-
-        def __exit__(self, exc_type, exc_val, exc_tb):
-            setattr(
-                self.manager,
-                self.active_attr,
-                getattr(self.manager, self.active_attr) - self.message_count,
-            )
-            if exc_type is not None:
-                logger.error(f"Controller Manager error: {exc_val}")
-            return False
 
     def _setup_socket_metrics(self):
         """Setup metrics for socket message counts."""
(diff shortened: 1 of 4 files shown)


lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -84,6 +90,15 @@ def __init__(self, controller_urls: dict[str, str]):
...
                 bind_or_connect="bind",
             )
+
+        # Message counters for observability
+        self.pull_socket_message_count = 0
+        self.rep_socket_message_count = 0
+
+        # Active request counters
+        self.pull_socket_active_requests = 0
+        self.rep_socket_active_requests = 0
```
KuntaiDu: Prefer putting these definitions into `_setup_socket_metrics`. This is not required tho.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -154,12 +172,93 @@ async def handle_orchestration_message(self, msg: OrchMsg) -> OrchRetMsg:
...
+                    # Decrement active request counter
+                    self.pull_socket_active_requests -= part_count
+            except Exception as e:
+                logger.error(f"Controller Manager error: {e}")
+
+    async def handle_batched_req_request(self, socket) -> Optional[MsgBase]:
+        while True:
+            try:
+                part = await socket.recv()
+                self.rep_socket_message_count += 1
```
KuntaiDu: Are we going to expand the amount of metrics in the future? If so maybe it is better to wrap it into a python context, like
`with self.rep_socket_context:`
and we can also put the error handling logic there. Hopefully this will also reduce the code diff in github.

KuntaiDu (commented):
See the comments.

maobaolong:
s@KuntaiDu Thanks for your suggestion, the `with... _SocketMetricsContext` is a good practice, it look much clear than before. PTAL.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -120,6 +135,15 @@ def __init__(
...
             asyncio.run_coroutine_threadsafe(self.health_check(), self.loop)
 
+        # Setup socket message count metrics
+        self._setup_socket_metrics()
+
+        # self.loop = asyncio.new_event_loop()
+        # self.thread = threading.Thread(target=self.loop.run_forever,
+        #                               daemon=True)
+        # self.thread.start()
+        # asyncio.run_coroutine_threadsafe(self.start_all(), self.loop)
```
KuntaiDu: Maybe clean this part of code up.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -166,15 +190,100 @@ async def handle_orchestration_message(self, msg: OrchMsg) -> OrchRetMsg:
...
+            if self.controller_urls["reply"] is not None:
+                prometheus_logger.rep_socket_has_pending.set_function(
+                    lambda: self._check_socket_has_pending(self.controller_rep_socket)
+                )
+
+            # Active request metrics
+            prometheus_logger.pull_socket_active_requests.set_function(
+                lambda: self.pull_socket_active_requests
+            )
+            prometheus_logger.rep_socket_active_requests.set_function(
```
KuntaiDu: Ideally we should stick with `rep` or stick with `reply`. Switching between them is slightly confusing.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -166,15 +190,100 @@ async def handle_orchestration_message(self, msg: OrchMsg) -> OrchRetMsg:
             logger.error(f"Unknown orchestration message type: {msg}")
             raise RuntimeError(f"Unknown orchestration message type: {msg}")
 
+    class _SocketMetricsContext:
```
KuntaiDu: Maybe put this code into `observability.py` as this part of logic is tightly coupled with the variable namings in `observability.py`.

lmcache/v1/cache_controller/controller_manager.py
KuntaiDu: But it doesn't bother me that much so it is OK to not change it if running out of bandwidth.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -91,6 +105,7 @@ def __init__(
                 role=zmq.REP,  # type: ignore[attr-defined]
                 bind_or_connect="bind",
             )
+
```
KuntaiDu: Maybe remove this line, just to make the code diff cleaner.

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -53,13 +55,25 @@
 # a control message.
 
 
+class SocketType(Enum):
```
KuntaiDu: Maybe also move it into observability as it is only used by Prometheus metric for now.

KuntaiDu (commented):
See comments.

lmcache/v1/cache_controller/controller_manager.py
maobaolong: done

lmcache/v1/cache_controller/controller_manager.py
maobaolong: done

lmcache/v1/cache_controller/controller_manager.py
maobaolong: done

lmcache/v1/cache_controller/controller_manager.py
maobaolong: done

lmcache/v1/cache_controller/controller_manager.py
maobaolong: Done, thanks for explain this to me offline!

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -166,15 +179,67 @@ async def handle_orchestration_message(self, msg: OrchMsg) -> OrchRetMsg:
...
+                )
+
+            # Active request metrics
+            prometheus_logger.pull_socket_active_requests.set_function(
+                lambda: self.pull_socket_active_requests
+            )
+            prometheus_logger.reply_socket_active_requests.set_function(
+                lambda: self.reply_socket_active_requests
+            )
+
```
KuntaiDu, line 217: Ideally this should also be in the observability file because these variables are be owned and maintained just for observability. And ideal API would be:
```
during init:
self.socket_metric_logger = SocketMetricLogger(self)
self.socket_metric_context = self.socket_metricLogger.get_logging_context()

during work:
with self.socket_metric_context(SocketType.Reply)
```

KuntaiDu (approved):
Logic LGTM, the changes I listed are not required (but would be great to have if you have bandwidth).

maobaolong:
> Logic LGTM, the changes I listed are not required (but would be great to have if you have bandwidth).

@KuntaiDu Thank you for you review again, let me do the improvement after this PR.

chunxiaozheng (approved):
As discussed offline, this LGTM!

maobaolong:
@KuntaiDu @chunxiaozheng Thanks for your review!

PR #2056, opened 2025-11-21
fix format error and typo

yoo-kumaneko:
The multi-line string needs to be wrapped in parentheses if we use the % format.
Otherwise, it errors like this.
Also, the number in ms should be divided by 1000. Fix the typo.

(EngineCore_DP0 pid=607072)   File "/root/vllm-lmcache/LMCache/lmcache/v1/lookup_client/lmcache_async_lookup_client.py", line 176, in lookup
(EngineCore_DP0 pid=607072)     logger.warning(
(EngineCore_DP0 pid=607072) Message: 'Request %s is still waiting for async lookup'
(EngineCore_DP0 pid=607072) Arguments: ('after %d seconds, returning 0 lmcache cached tokens', 'so vllm can recompute', 'chatcmpl-55fb505003354be7a6d4185f2205235f', 3000)

### DIFF
--- a/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
+++ b/lmcache/v1/lookup_client/lmcache_async_lookup_client.py
@@ -174,11 +174,13 @@ class LMCacheAsyncLookupClient(LookupClientInterface):
                     time.time() - self.first_lookup_time[lookup_id]
                 ) * 1000 > self.timeout_ms:
                     logger.warning(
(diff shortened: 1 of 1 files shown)


yoo-kumaneko:
@sammshen @maobaolong would you like to take a look at this?

maobaolong (approved):
Good catch! LGTM

sammshen (approved):
Thanks for the fix @yoo-kumaneko !

PR #2073, opened 2025-11-25
[Observability] Adding lifespan tracker

YuhanLiu11:
**What this PR does / why we need it**: This PR adds the chunk-level lifespan tracker, whenever a chunk is added initially, it records the initial timestamp in LRUCachePolicy. When the chunk is reused, it calls on_chunk_reuse in the Prometheus logger to add the time interval of the reuse timestamp and the initial timestamp. 

**If applicable**:


YuhanLiu11:
BTW @ApostaC I also benchmarked the time to read an item from a 1250000-item dictionary, and the read time is 18.7 ns, and I believe this time is ok overall.

lmcache/observability.py
```diff
@@ -502,6 +518,13 @@ def filter_out_zeros(stats: List[float]):
             ]
         )
 
+        request_lifespan = filter_out_zeros(
+            [
+                self.interval_request_cache_lifespan[key]
+                for key in self.interval_request_cache_lifespan.keys()
+            ]
```
ApostaC: As suggested by gemini review, we can do `self.interval_request_cache_lifespan.values()`. And I guess we don't need `filter_out_zeros` for this, right?

lmcache/v1/storage_backend/cache_policy/lru.py
```diff
@@ -18,22 +20,38 @@ class LRUCachePolicy(BaseCachePolicy[OrderedDict[CacheEngineKey, Any]]):
 
     def __init__(self):
         logger.info("Initializing LRUCachePolicy")
+        self.chunk_hash_to_init_timestamp: Dict[Any, Any] = {}
+        self.stats_monitor = LMCStatsMonitor.GetOrCreate()
+        self.max_num_chunk_hash = 12500000
```
ApostaC, line 25: I remember we said ~1 million. Did we get this number from some perf bench?

lmcache/v1/storage_backend/cache_policy/lru.py
```diff
@@ -18,22 +20,38 @@ class LRUCachePolicy(BaseCachePolicy[OrderedDict[CacheEngineKey, Any]]):
...
+        if key.chunk_hash not in self.chunk_hash_to_init_timestamp:
+            if len(self.chunk_hash_to_init_timestamp) >= self.max_num_chunk_hash:
+                # Clear the dictionary to avoid memory leak
+                self.chunk_hash_to_init_timestamp = {}
+            self.chunk_hash_to_init_timestamp[key.chunk_hash] = time.time()
+        if key.chunk_hash in self.chunk_hash_to_init_timestamp:
+            time_interval = (
+                time.time() - self.chunk_hash_to_init_timestamp[key.chunk_hash]
+            )
+            self.stats_monitor.on_chunk_reuse(time_interval)
```
ApostaC: We can reduce the number of lookups in this function. Could be something like:
```suggestion
        curr_time = time.time()
        key_hash = key.chunk_hash
        if init_timestamp := self.chunk_hash_to_init_timestamp.get(key_hash, None):
            time_interval = curr_time - init_timestamp
            self.stats_monitor.on_chunk_reuse(time_interval)
        else:
            if len(self.chunk_hash_to_init_timestamp) >= self.max_num_chunk_hash:
                self.chunk_hash_to_init_timestamp.clear()
            self.chunk_hash_to_init_timestamp[key_hash] = curr_time
```

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/storage_backend/cache_policy/lru.py
ApostaC: Oh just saw your comment above. Nvm.

lmcache/v1/storage_backend/cache_policy/lru.py
```diff
@@ -18,22 +20,37 @@ class LRUCachePolicy(BaseCachePolicy[OrderedDict[CacheEngineKey, Any]]):
 
     def __init__(self):
         logger.info("Initializing LRUCachePolicy")
+        self.chunk_hash_to_init_timestamp: Dict[Any, Any] = {}
```
ApostaC: Genimi did a good catch here. We can have
```python3
self.chunk_hash_to_init_timestamp: dict[int, float] = {}
```

ApostaC (approved):
Otherwise LGTM!

lmcache/v1/storage_backend/cache_policy/lru.py
YuhanLiu11: Sure!

sammshen (approved):
LGTM！

PR #2079, opened 2025-11-25
[Bugfix|p2p] P2P Server side do not set time out while receiving

maobaolong:
Before this PR, I set timeout for both side of p2p, but it should not be set timeout for server role side, since server role receive request all the time, it there are no request, server should receive all the time.

@chunxiaozheng Thanks for raise this issue.


maobaolong:
@sammshen @chunxiaozheng Would you like to take a look at this PR?

sammshen (approved):
LGTM!

chunxiaozheng (approved):
lgtm

PR #2083, opened 2025-11-25
[P2PBackend] support p2p sync loading

chunxiaozheng:
# DESCRIPTION
support p2p sync loading, includes the following methods:
- `batched_contains`
- `batched_get_blocking`

The current implementation reuses async implementations.

# TODO
- implement real sync method

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1049,7 +1049,7 @@ class LMCacheEngine:
                     keys.append(chunk_info[2])
                 # hit chunks by prefix matching
                 hit_chunks, block_mapping = self.storage_manager.batched_contains(
-                    keys, search_range, pin
+                    keys, search_range, pin, lookup_id
                 )
                 if pin and block_mapping:
                     assert lookup_id is not None, (
@@ -1509,6 +1509,7 @@ class LMCacheEngine:
 
         tot_kv_size = 0
         reordered_chunks: List[ProcessedChunk] = []
+        lookup_id = kwargs["req_id"] if "req_id" in kwargs else None
         request_configs = kwargs.get("request_configs")
         if request_configs is not None and len(request_configs) != 0:
             assert isinstance(request_configs, dict)
@@ -1538,14 +1539,23 @@ class LMCacheEngine:
             location = self.storage_manager.non_allocator_backends[0]
             block_mapping = {location: chunk_infos}
         else:
-            block_mapping = self.storage_manager.get_block_mapping(chunk_infos)
+            block_mapping = self.storage_manager.get_block_mapping(
+                chunk_infos, lookup_id
+            )
 
         last_failed_block_start = None
         for location, blocks in block_mapping.items():
-            keys = [key for key, _, _ in blocks]
+            keys = []
+            cum_chunk_lengths = [blocks[0][1]]
+            for key, start, end in blocks:
+                keys.append(key)
+                cum_chunk_lengths.append(end)
+            transfer_spec = {"cum_chunk_lengths": cum_chunk_lengths}
             memory_objs = self.storage_manager.batched_get(
                 keys=keys,
                 location=location,
+                lookup_id=lookup_id,
+                transfer_spec=transfer_spec,
             )
             assert memory_objs is not None, (
                 "Failed to get memory objects from storage backend"
--- a/lmcache/v1/storage_backend/abstract_backend.py
+++ b/lmcache/v1/storage_backend/abstract_backend.py
@@ -165,11 +165,17 @@ class StorageBackendInterface(metaclass=abc.ABCMeta):
     def batched_get_blocking(
         self,
         keys: List[CacheEngineKey],
+        lookup_id: Optional[str] = None,
+        transfer_spec: Any = None,
     ) -> List[Optional[MemoryObj]]:
         """
         A blocking function to get the kv cache from the storage backend.
 
         :param List[CacheEngineKey] keys: The keys of the MemoryObjs.
+
+        :param Optional[str] lookup_id: The lookup id of the request.
+
+        :param Any transfer_spec: The transfer spec of the request.
 
         :return: a list of memory objects.
         """
@@ -260,6 +266,7 @@ class StorageBackendInterface(metaclass=abc.ABCMeta):
         self,
         keys: List[CacheEngineKey],
         pin: bool = False,
+        lookup_id: Optional[str] = None,
     ) -> int:
         """
         Check whether the keys are in the storage backend.
@@ -269,6 +276,8 @@ class StorageBackendInterface(metaclass=abc.ABCMeta):
         :param bool pin: Whether to pin the key.
             If True, the corresponding KV cache will be
             pinned in the storage backend.
+
+        :param Optional[str] lookup_id: The lookup id.
 
         :return: Return hit chunks by prefix match.
         """
--- a/lmcache/v1/storage_backend/p2p_backend.py
+++ b/lmcache/v1/storage_backend/p2p_backend.py
@@ -97,6 +97,21 @@ class BatchedLookupAndPutRetMsg(P2PMsgBase):
     num_read_chunks: int
 
 
+class BatchedP2PPinMsg(P2PMsgBase):
+    """P2P pin message"""
+
+    lookup_id: str
+    # CacheEngineKey in string form
+    keys: list[str]
+
+
+class BatchedP2PPinRetMsg(P2PMsgBase):
+    """P2P pin return message"""
+
+    # Number of pin chunks
+    pin_count: int
+
+
 class P2PErrorCode(enum.Enum):
     """P2P error codes enumeration"""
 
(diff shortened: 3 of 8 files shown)


lmcache/v1/storage_backend/abstract_backend.py
```diff
@@ -165,12 +165,18 @@ async def batched_get_non_blocking(
     def batched_get_blocking(
         self,
         keys: List[CacheEngineKey],
+        lookup_id: Optional[str] = None,
+        transfer_spec: Any = None,
```
KuntaiDu, line 169: The parameter ordering is not aligning with `batched_get_non_blocking`, which is
```
        lookup_id: str,
        keys: list[CacheEngineKey],
        transfer_spec: Any = None,
```
I would prefer aligning the parameter ordering, maybe by editing the codepath and pass down the `lookup_id` to existing `batched_get_blocking` call.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -269,6 +295,20 @@ async def batched_async_contains(
                     self.peer_id_to_lookup_url_mapping[peer_init_url],
                     location,
                 )
+                # pin for sync loading
+                if pin and not self.enable_async_loading:
```
KuntaiDu, line 346: This codepath only pins for async loading. Do we also need to pin for async loading?

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -335,26 +424,50 @@ async def _handle_batched_lookup_and_get(
...
-
-            # TODO(Jiayi): Optimally, there's no need to use async call
-            # for some backends (e.g., local cpu) as there's overhead for
-            # async function call.
-            num_hit_chunks = await self.local_cpu_backend.batched_async_contains(
-                lookup_id=lookup_id,
-                keys=keys,
-                pin=True,
-            )
+            if self.enable_async_loading:
```
KuntaiDu, line 478: I would prefer a way that unifies the logic between async loading, as sync loading is just a special case of async loading for me.

KuntaiDu (commented):
See comments. Basically I would prefer a way that maximally unifies sync loading and async loading code path.

lmcache/v1/storage_backend/abstract_backend.py
chunxiaozheng: @KuntaiDu Thanks for your suggestion! But the `lookup_id` is sync mode is unnecessary, the definition is `lookup_id: Optional[str] = None`, default value is `None`, and must be placed after `keys`, while `lookup_id` is necessary in async mode.

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: The implements of sync and async are very different. 
In sync mode, the hit chunk is queried through `lookup` and returned directly, so it must be pinned during lookup and retrieved later; 
In async mode, it is first queried through `lookup`, and then `prefetch`, returning the count of prefetched.
Therefore, in sync mode, it must be pinned during `lookup`, while in async mode, it is pinned during `prefetch`.

lmcache/v1/storage_backend/abstract_backend.py
KuntaiDu: @ApostaC What's your suggestion on this?
My preference is that, if `lookup_id` is unnecessary in sync mode, can we remove it in `batched_get_blocking`? Inside the implementation of `batched_get_blocking` I am ok with implementing some logic to fake a lookup id.

lmcache/v1/storage_backend/p2p_backend.py
KuntaiDu: Just to rephrase what you said: basically `pin` ties to `prefetch`, but in sync loading `prefetch` needs to be called **inside** `lookup`, while in async code path the `prefetch` can be called **after** lookup returns. Is this the case?

If this is the case, I would prefer making the call stack clear: in sync code path, `lookup` calls `prefetch`, and then `prefetch` calls pin.

YaoJiayi (commented):
@chunxiaozheng I have two questions.
(1) What was the goal of having a sync code path?
(2) Why can't we reuse the code of async in sync path? It doesn't have to follow the lookup/pin + retrieve pattern. We can still do everything in the lookup/pin part and retrieve part is simply an empty function call? I assume this will make everything a lot easier?

chunxiaozheng:
> @chunxiaozheng I have two questions. (1) What was the goal of having a sync code path? (2) Why can't we reuse the code of async in sync path? It doesn't have to follow the lookup/pin + retrieve pattern. We can still do everything in the lookup/pin part and retrieve part is simply an empty function call? I assume this will make everything a lot easier?

@YaoJiayi Thanks for your review. 
- The first question, we haven't tested a large number of async mode in our scenario, and even now we can't run the tests. This poses a certain risk, so we need to develop sync mode to ensure a minimum.
- The second question, I don't quite understand how to handle it?

In sync mode, it has already run smoothly according to the current implementation. 
In async mode, the current error is as follows, the `MultiConnector` does not adapt `get_num_new_matched_tokens` return `None`:
Clipboard_Screenshot_1765160835
(🥲 this question is fixed, our internal version need to cherry-pick https://github.com/vllm-project/vllm/pull/23620)
It is uncertain if there will be other issues during subsequent pressure testing.

lmcache/v1/cache_engine.py
```diff
@@ -989,7 +989,7 @@ def lookup(
                     return res
                 # hit chunks by prefix matching
                 hit_chunks, block_mapping = self.storage_manager.batched_contains(
-                    keys, search_range, pin
+                    keys, search_range, pin, lookup_id
```
sammshen: lookup_id doesn't exist anymore

sammshen: need to update branch and rebase

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -896,24 +905,30 @@ def batched_contains(
...
+        chunk_infos: List[Tuple[CacheEngineKey, int, int]],
+        lookup_id: Optional[str] = None,
     ) -> Dict[str, List[Tuple[CacheEngineKey, int, int]]]:
         """
         Get block mapping for the given chunk infos, works by prefix match.
 
         :param List[Tuple[CacheEngineKey, int, int]] chunk_infos:
         List of chunk infos, each tuple contains (key, begin, end)
 
+        :param Optional[str] lookup_id: The lookup id.
```
sammshen, line 918: can we add some comment on what cases the lookup id should be passed in?

sammshen: same comment to all the places the optional lookup id has been added

lmcache/v1/storage_backend/p2p_backend.py
sammshen: agreed

lmcache/v1/storage_backend/storage_manager.py
sammshen: adding the extra field without this explanation will make the storage manager harder to understand in the future

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -396,26 +475,50 @@ async def _handle_batched_lookup_and_get(
...
+                # async function call.
+                num_hit_chunks = await self.local_cpu_backend.batched_async_contains(
+                    lookup_id=lookup_id,
+                    keys=keys,
+                    pin=True,
+                )
+                keys = keys[:num_hit_chunks]
+                remote_mem_indexes = msg.mem_indexes[:num_hit_chunks]
+            else:
+                keys = self.pin_keys.pop(lookup_id)
```
sammshen, line 492: just to clarify, the synchronous case will also call `BatchedLookupAndGet`?

sammshen: and then will also call `batched_get_non_blocking()` at the end? so the synchronous case is using this for unpin + retrieve, is this correct?

lmcache/v1/storage_backend/storage_manager.py
chunxiaozheng: @sammshen Thanks for your suggestion! the `lookup_id` only used in `P2PBackend`, I will add some comments.

feixiangpeng:
What's the point of adding synchronous mode compared to existing asynchronous mode

sammshen:
I have the same question as @feixiangpeng

chunxiaozheng:
> I have the same question as @feixiangpeng

@sammshen @feixiangpeng Our internally use the sync mode, and the async mode hasn't been extensively tested in our scenario, so we prefer to use the sync mode.

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: @sammshen Yeah. The `pin` operation will save `pin_keys`.

sammshen (approved):
LGTM, I don't think adding a lookup_id to the sync case just for P2PBackend is great but we can use this as a starting point for async refactor planned in Q1

PR #2085, opened 2025-11-26
[Controller] Introduce batch send message to controller

maobaolong:
# Summary:
This PR introduces batch message sending capability to the Controller component, improving performance and efficiency when handling multiple messages.

# Changes:
- Implement batch message sending mechanism in Controller
- Optimize message processing for better performance
- Enhance system scalability for handling concurrent operations


lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -140,6 +142,37 @@ async def handle_worker_message(self, msg: WorkerMsg) -> None:
...
+                        key=op.key,
+                        location=location,
+                        seq_num=op.seq_num,
+                    )
+                    await self.kv_controller.evict(evict_msg)
+                    # Check sequence number if available
+                    self.kv_controller.check_sequence_number(evict_msg)
+                else:
+                    logger.error(f"Unknown operation type: {op.op_type}")
+        # TODO(baoloongmao): Use BatchedKVOperationMsg instead of these two
```
ApostaC: nit: is this `TODO` still needed?

lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -52,12 +56,55 @@ def __init__(self) -> None:
         self.kv_pool: dict[int, list[KVChunkMetadata]] = {}
         # TODO(Jiayi): remove this hardcode
         self.token_database = ChunkedTokenDatabase()
+
+        # Track sequence numbers for each (instance_id, worker_id, location) tuple
+        # Key: (instance_id, worker_id, location), Value: last_seq_num
+        self.seq_tracker: dict[tuple[str, int, str], int] = {}
```
ApostaC, line 62: Is it expected to have `location` in the key? 
I suppose the `location` here means something like `LocalCPU` or `LocalDisk`. Does this mean different storage backends need to track their own sequence number?

lmcache/v1/cache_controller/controller_manager.py
```diff
@@ -140,6 +142,37 @@ async def handle_worker_message(self, msg: WorkerMsg) -> None:
...
         elif isinstance(msg, KVAdmitMsg):
             await self.kv_controller.admit(msg)
         elif isinstance(msg, KVEvictMsg):
```
ApostaC, line 177: Why don't we check seq num for `KVAdmitMsg` or `KVEvictMsg`?  They are also inherited from `KVOperationMsg`.
Should the sequence number be a common logic across all messages? Or it's just specialized for the batched KV cache operation message?

lmcache/v1/cache_engine.py
```diff
@@ -130,6 +130,16 @@ def __init__(
             and (not lmcache_worker_ids or metadata.worker_id in lmcache_worker_ids)
         ):
             self.lmcache_worker = LMCacheWorker(config, metadata, self)
+        else:
+            self.lmcache_worker = None
+            logger.info(
+                "LMCacheWorker is not initialized. enable_controller: %s, "
+                "role: %s, worker_id: %s, worker_ids: %s",
```
ApostaC: nit: improve the log message's clarity. Something like:
```suggestion
            logger.info(
                "LMCacheWorker is not initialized (related configs: "
                "enable_controller: %s, role: %s, worker_id: %s, worker_ids: %s).",
```

lmcache/v1/storage_backend/batched_message_sender.py
```diff
@@ -0,0 +1,188 @@
...
+        if prometheus_logger is not None:
+            prometheus_logger.kv_msg_queue_size.set_function(
+                lambda: len(self.message_queue)
+            )
+
+    def _start_background_thread(self):
+        """Start background thread for periodic flushing."""
+        self.running = True
+        self.loop = asyncio.new_event_loop()
+        self.thread = threading.Thread(target=self._run_event_loop, daemon=True)
```
ApostaC: Since we already initialized a new thread here, why do we use async io loop? 
`time.sleep()` will work well in this case.

lmcache/v1/storage_backend/batched_message_sender.py
```diff
@@ -0,0 +1,188 @@
...
+        self.lmcache_worker = lmcache_worker
+
+        # Common fields shared by all operations in the batch
+        self.instance_id = config.lmcache_instance_id
+        self.worker_id = metadata.worker_id
+        self.location = location
+
+        # Use a single queue to maintain order consistency between admit and evict
+        # Store lightweight operations without redundant common fields
+        self.message_queue: List[KVOpEvent] = []
```
ApostaC: We can use `queue.Queue` which is a thread-safe producer-consumer queue (https://docs.python.org/3/library/queue.html)

ApostaC: At a high level, we can implement this without threading.lock if we use `queue.Queue`.

For the producer (`add_kv_op`), we simply call `put`. We can configure it to block if the queue is full, or use a conditional variable (https://docs.python.org/3/library/threading.html#condition-objects) to notify the consumer and trigger the flush.

For the consumer loop, we can use `cv.wait(timeout = 0.01)` to make sure it wakes up every 10ms. If the queue is full and the producer calls `notify`, it will wake up early. Every time it wakes up, it will try to drain the queue and send all the messages together.

lmcache/v1/storage_backend/batched_message_sender.py
ApostaC, line 1: Should the `batched_message_sender.py` be put together with `worker.py` rather than in `storage_backend/`?

tests/v1/storage_backend/test_local_cpu_backend.py
ApostaC, line 1: Can we add some tests for `batched_message_sender`? Since it involves multi-threading, it would be great if we have some stress tests for it.

ApostaC (changes requested):
Have some proposed changes for `BatchedMessageSender`

lmcache/v1/cache_controller/controller_manager.py
maobaolong: Yeah, Since I didn't handle the related code within `local_disk_backend.py`,  and local disk still keep to send original `AdmitMsg` and `EvictMsg`, i want focus on local cpu backend and P2P, and after the sprint during this period, i will handle the local_disk_backend, after that, the TODO can be removed.

lmcache/v1/cache_controller/controllers/kv_controller.py
maobaolong: Yeah, i think so, i intended to keep the last_seq_num for each location.

lmcache/v1/cache_controller/controller_manager.py
maobaolong: In fact, local_cpu backend should never send `KVAdmitMsg` or `KVEvictMsg` to KVController, instead, it will send `BatchedKVOperationMsg`, but `local_disk` backend will send `KVAdmitMsg` or `KVEvictMsg` as before, and I didn't want modify local_disk backend in this PR to limit the size of PR and focus on `local_cpu and p2p`.

So we cannot check seq num for KVAdmitMsg or KVEvictMsg for now.

lmcache/v1/storage_backend/batched_message_sender.py
maobaolong: Thank you for this suggestion. After careful analysis, I guess that batched_message_sender.py should be better to remain under the storage_backend/directory. 

This is because tts core responsibility is to assist the storage backend in batch-sending KV operation messages.

But whatever, I'm open to move it to `lmcache/v1/cache_controller/` if you think it would be better.

maobaolong:
@ApostaC Thanks for your suggestion about refactor. I've completely refactored it to use queue.Queue and threading.Condition as you suggested:
- Producer (add_kv_op): Simply calls queue.put() and notifies the consumer via threading.Condition.notify() when the batch size threshold is reached
- Consumer (_consumer_loop): Uses threading.Condition.wait(timeout=0.01) to wake up periodically or get notified early by the producer, then drains the queue and sends all messages together

This approach is much simpler, more efficient, and easier to understand. No more asyncio complexity!

About the suggestion of moving `batched_message_sender.py`, i replied below your comment. PTAL.

lmcache/v1/cache_controller/controller_manager.py
ApostaC: Sounds good! Can we add a sentence to the TODO here to indicate there are some pending tasks for LocakDiskBackend? Thanks!

lmcache/v1/cache_controller/controller_manager.py
ApostaC: I see what you mean. Then, can we make the seq_num a field of the batched message? (I suppose we will change to use the batched message in the future as well)

lmcache/v1/storage_backend/batched_message_sender.py
ApostaC: I see. Let's keep it as is for now. Renaming & moving files will always be easy if it's needed in the future

tests/v1/storage_backend/test_local_cpu_backend.py
```diff
@@ -485,3 +531,486 @@ def test_ref_count_management(self, local_cpu_backend):
         local_cpu_backend.remove(key)
         assert memory_obj.get_ref_count() == initial_ref_count + 1
         local_cpu_backend.memory_allocator.close()
+
+
+class TestBatchedMessageSender:
```
ApostaC: Quick comment: can we move the test to another standalone test file (something like `test_batched_message_sender.py`) instead of in `test_local_cpu_backend.py`? 
IIUC, the batched message sender will be used by multiple different backends in the future.

ApostaC (approved):
Otherwise LGTM!

lmcache/v1/cache_controller/controller_manager.py
maobaolong: Sure

lmcache/v1/cache_controller/controllers/kv_controller.py
chunxiaozheng, line 62: Current `kv_controller` is not thread safe, we can handle it uniformly in another PR.

chunxiaozheng (approved):
As discussed offline, this LGTM!

PR #2086, opened 2025-11-26
Fix bugs

crystalxyz:
Hashing is unstable with `None` value, and also built-in hash cannot handle `list`, so we convert it to `tuple`.

**If applicable**:

- [X] this PR contains unit tests


crystalxyz:
Opened to the wrong source

PR #2088, opened 2025-11-26
[bugfix] support re-establish peer connection when some peer restart

chunxiaozheng:
When the peers have build the connection, the peer will have cache in `self.peer_id_to_lookup_url_mapping` and other dict, if some peer crashed and restart, the new started peer does not have peer init information, but the other peers still have peer connection, which will result in the following error in the restarted peer
Clipboard_Screenshot_1764139703

To fix this problem, I add the registration_time as an identifier, which is used to check if the peer in the controller has been updated, if the peer is updated, such as restart, we must re-establish the peer connection.

### DIFF
--- a/lmcache/v1/storage_backend/p2p_backend.py
+++ b/lmcache/v1/storage_backend/p2p_backend.py
@@ -99,6 +99,9 @@ class P2PErrorMsg(P2PMsgBase):
     Error message, return error code to client.
 
     -1 represents unknown msg type;
+    -2 represents remote xfer handler not initialized,
+        call `_ensure_peer_connection` first;
+    -3 represents p2p peer_request_handler error;
     """
 
     error_code: int
@@ -138,6 +141,9 @@ class P2PBackend(StorageBackendInterface):
         self.socket_send_timeout_ms = config.get_extra_config_value(
             "p2p_socket_send_timeout_ms", DEFAULT_SOCKET_SEND_TIMEOUT_MS
         )
+
+        # Load max retry count from extra_config
+        self.max_retry_count = config.get_extra_config_value("p2p_max_retry_count", 3)
 
         # tp rank is worker id for now
         self.tp_rank = metadata.worker_id
@@ -190,8 +196,6 @@ class P2PBackend(StorageBackendInterface):
         self.running.set()
         self.lookup_url_to_socket_mapping: dict[str, zmq.asyncio.Socket] = {}
         self.lookup_url_to_lock_mapping: dict[str, asyncio.Lock] = {}
-        # Store socket parameters for recreation
-        self.lookup_url_to_socket_params: dict[str, dict] = {}
         self.async_context: Optional[zmq.asyncio.Context] = None
         self.async_peer_socket: Optional[zmq.asyncio.Socket] = None
         asyncio.run_coroutine_threadsafe(
@@ -327,7 +331,7 @@ class P2PBackend(StorageBackendInterface):
 
     async def _handle_batched_lookup_and_get(
         self, msg: BatchedLookupAndGetMsg
-    ) -> BatchedLookupAndGetRetMsg:
+    ) -> P2PMsgBase:
         lookup_id = msg.lookup_id
         mem_objs = None
         try:
@@ -335,6 +339,13 @@ class P2PBackend(StorageBackendInterface):
                 "Received P2P batched lookup and get msg, lookup_id: %s", lookup_id
             )
             receiver_id = msg.receiver_id
+            if not self.transfer_channel.remote_xfer_handler_exists(receiver_id):
+                logger.error(
+                    "Receiver %s does not exist in transfer channel",
+                    receiver_id,
+                )
+                return P2PErrorMsg(error_code=-2)
+
             remote_mem_indexes = msg.mem_indexes
             keys = [CacheEngineKey.from_string(key) for key in msg.keys]
 
@@ -370,7 +381,7 @@ class P2PBackend(StorageBackendInterface):
                 e,
                 exc_info=True,
             )
-            return BatchedLookupAndGetRetMsg(num_hit_chunks=0)
+            return P2PErrorMsg(error_code=-3)
         finally:
             if mem_objs is not None:
                 for mem_obj in mem_objs:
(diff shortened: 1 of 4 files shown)


YaoJiayi (commented):
Thanks @chunxiaozheng ! I'm not sure why registration time is needed here. Correct me if I'm wrong: my take is that we only need a simple try except to re-establish the connection? Why can't a simple try-except work here?

chunxiaozheng:
> Thanks @chunxiaozheng ! I'm not sure why registration time is needed here. Correct me if I'm wrong: my take is that we only need a simple try except to re-establish the connection? Why can't a simple try-except work here?

@YaoJiayi Thanks for your review! In my test, I start two peer(`peer1` and `peer2`), if `peer2` restart, the error happens in `peer2`, but the `ensure_peer_connection` need called by `peer1`.

chunxiaozheng:
> > Thanks @chunxiaozheng ! I'm not sure why registration time is needed here. Correct me if I'm wrong: my take is that we only need a simple try except to re-establish the connection? Why can't a simple try-except work here?
> 
> @YaoJiayi Thanks for your review! In my test, I start two peer(`peer1` and `peer2`), if `peer2` restart, the error happens in `peer2`, but the `ensure_peer_connection` need called by `peer1`.

Besides, If we add `try-except` in `peer1`, we must wait `recv` timeout.

YaoJiayi:
> > Thanks @chunxiaozheng ! I'm not sure why registration time is needed here. Correct me if I'm wrong: my take is that we only need a simple try except to re-establish the connection? Why can't a simple try-except work here?
> 
> @YaoJiayi Thanks for your review! In my test, I start two peer(`peer1` and `peer2`), if `peer2` restart, the error happens in `peer2`, but the `ensure_peer_connection` need called by `peer1`.

@chunxiaozheng Why not send an error message back to peer1? This will give two benefits: (1) make the solution simpler; (2) make fault tolerance more general (e.g., the same mechanism/code path can be used to handle disk failure in the future).

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -454,9 +458,17 @@ def _recreate_lookup_socket(self, peer_lookup_url: str) -> None:
...
         self,
         peer_init_url: str,
+        registration_time: float = 0.0,
     ) -> None:
         if peer_init_url in self.peer_id_to_lookup_url_mapping:
-            return
+            if (
+                registration_time > 0.0
+                and registration_time
+                != self.peer_id_to_lookup_url_mapping[peer_init_url][1]
```
maobaolong: Could you add a comment here to explain [1] is the registration_time in `self.peer_id_to_lookup_url_mapping: dict[str, tuple[str, float]]`

BTW, it's good for performance to use tuple, but hard to maintain and hard for readability. Whatever, let's keep to use tuple here and discuss whether to use `NamedTuple` or any other data structure in future.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -484,6 +499,28 @@ async def _ensure_peer_connection(
...
+            peer_init_url
+        )
+        logger.info(
+            "clearing peer connection cache for peer_init_url %s, "
+            "old registration time: %s, new registration time: %s",
+            peer_init_url,
+            old_registration_time,
+            registration_time,
+        )
+        self.lookup_url_to_lock_mapping.pop(peer_lookup_url, None)
```
maobaolong: It would be better to maintain a PeerStatus or PeerInfo data class in future, something like
```python
class PeerInfo:
  peer_lock: asyncio.Lock,
  peer_socket: zmq.asyncio.Socket,
  peer_socket_params: dict
```

After that, we can pop these related together.

maobaolong (commented):
@chunxiaozheng LGTM overall, left some minor comment and future idea.

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: Good idea! We can update here in another PR!

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: @maobaolong I have update the code, the implements changed, PTAL! Thanks!

chunxiaozheng:
> > > Thanks @chunxiaozheng ! I'm not sure why registration time is needed here. Correct me if I'm wrong: my take is that we only need a simple try except to re-establish the connection? Why can't a simple try-except work here?
> > 
> > 
> > @YaoJiayi Thanks for your review! In my test, I start two peer(`peer1` and `peer2`), if `peer2` restart, the error happens in `peer2`, but the `ensure_peer_connection` need called by `peer1`.
> 
> @chunxiaozheng Why not send an error message back to peer1? This will give two benefits: (1) make the solution simpler; (2) make fault tolerance more general (e.g., the same mechanism/code path can be used to handle disk failure in the future).

@YaoJiayi Thanks for your suggestion, I have updated the code by your comment, PTAL!

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -99,6 +99,9 @@ class P2PErrorMsg(P2PMsgBase):
     Error message, return error code to client.
 
     -1 represents unknown msg type;
+    -2 represents remote xfer handler not initialized,
```
KuntaiDu, line 111: Prefer using ENUM instead.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -426,76 +437,12 @@ async def _handle_batched_lookup_and_put(
...
-                e,
-                exc_info=True,
-            )
-            self.lookup_url_to_socket_mapping.pop(peer_lookup_url, None)
-            raise
-
     async def _ensure_peer_connection(
         self,
         peer_init_url: str,
+        force_update: bool = False,
```
KuntaiDu, line 454: I didn't find the code of setting `force_update = True` in this PR. Will this be in future PR?

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -505,27 +452,56 @@ async def _ensure_peer_connection(
             init_side_msg=init_side_msg,
         )
         assert isinstance(init_ret_msg, P2PInitSideRetMsg)
-        peer_lookup_url = init_ret_msg.peer_lookup_url
-        self.peer_id_to_lookup_url_mapping[peer_init_url] = peer_lookup_url
+        # update peer_lookup_url
```
KuntaiDu, line 466: The following code discusses the case of `old_peer_lookup_url is not None` in multiple places. To simplify logic, is it possible to perform only one
`if old_peer_lookup_url is not None:`
at the beginning, and recycle / remove all previous locks inside the code block?

KuntaiDu (commented):
See comments

lmcache/v1/storage_backend/p2p_backend.py
YaoJiayi: Yeah ENUM is better

YaoJiayi (approved):
LGTM! Please change the error code to enum:)

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: Thanks for your suggestion, I have updated!

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -568,40 +540,60 @@ async def batched_get_non_blocking(
...
+                    ret_msg_bytes = await lookup_socket.recv()
+                    ret_msg = msgspec.msgpack.decode(ret_msg_bytes, type=P2PMsg)
+                    if isinstance(ret_msg, P2PErrorMsg) and ret_msg.error_code == -2:
+                        logger.warning(
+                            "Peer connection not initialized for lookup_id %s, "
+                            "ensure peer connection first, retry count: %s",
+                            lookup_id,
+                            retry_count,
+                        )
+                        await self._ensure_peer_connection(peer_init_url, True)
```
chunxiaozheng, line 574: @KuntaiDu here use `force_update` is `True`

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -568,40 +540,60 @@ async def batched_get_non_blocking(
...
+                    else:
+                        break
+                except zmq.ZMQError as e:
+                    logger.error(
+                        "ZMQ error occurred for lookup_id %s. Error: %s",
+                        lookup_id,
+                        e,
+                    )
+                    await self._ensure_peer_connection(peer_init_url, True)
```
chunxiaozheng, line 583: @KuntaiDu here also use `force_update` is `True`

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: Marked below

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: done

maobaolong (approved):
LGTM, thanks for this enhancement contribution!

PR #2089, opened 2025-11-26
[bugfix] Remote Connector should raise exception while read bytes is 0

chunxiaozheng:
In our test scenario, the distributed fs use async write, when the fs process crashed, which may produce some error files, the size is 0, if read from these error files, `bytes_read` is `0`, which will result in torch error. 
Therefore, we must check the `bytes_read` is not `0`, and promptly throw out exceptions.


chunxiaozheng:

maobaolong (approved):
LGTM @chunxiaozheng  Thanks for this fix.

sammshen (approved):
LGTM!

PR #2097, opened 2025-11-27
[Core][5/N] Locking mechanism and eviction for LMCache multi-process mode

ApostaC:
# Summary

This PR added the following feature to the multi-process mode:
1. "lock" mechanism (equivalent to "pin") with a TTL (default 300 seconds)
2. LRU-based eviction

### 📢📢📢 Once this PR is merged, the multi-process mode will become fully functional for CPU offloading.

# Follow up tasks
- End-to-end test
- Refactor the implementation with `StorageBackend` 
- New docker build pipeline and cli improvements
- Better fault tolerance

### DIFF
--- a/lmcache/v1/multiprocess/mp_storage_manager.py
+++ b/lmcache/v1/multiprocess/mp_storage_manager.py
@@ -2,10 +2,14 @@
 # NOTE: this file will be moved and the class implementation
 # will be largely refactored in the future.
 # Standard
+from collections import OrderedDict
+from collections.abc import Hashable
 from contextlib import contextmanager
+from dataclasses import dataclass
 from itertools import compress
-from typing import Iterator, Union
+from typing import Any, Generic, Iterator, TypeVar, Union
 import threading
+import time
 
 # Third Party
 import torch
@@ -15,6 +19,7 @@ from lmcache.logging import init_logger
 from lmcache.utils import _lmcache_nvtx_annotate
 from lmcache.v1.memory_management import MemoryFormat, MemoryObj, MixedMemoryAllocator
 from lmcache.v1.multiprocess.custom_types import IPCCacheEngineKey
+from lmcache.v1.storage_backend.cache_policy.lru import LRUCachePolicy
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


lmcache/v1/multiprocess/mp_storage_manager.py
```diff
@@ -131,23 +325,57 @@ def reserve(
...
-        return handle, reserved_dict
+                logger.info(
+                    "Recycled %d committed memory objects to free up space.",
+                    len(candidates),
+                )
+
+                # Try to allocate again
+                objects = self._memory_allocator.batched_allocate(
+                    shape, dtype, num_objects_to_allocate, fmt
+                )
```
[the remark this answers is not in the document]
ApostaC: No, you are wrong. Both this part of the code, `retrieve`, and the `lookup` function are secured by `self._buffer_lock`.

lmcache/v1/multiprocess/server.py
```diff
@@ -463,6 +463,13 @@ def lookup(
         """
         # NOTE: we are doing per-request lookup, the caller need
         # to be aware of this! We need to add this to the doc!
+        if not lock:
+            logger.warning(
+                "MPCacheEngine.lookup called with lock=False, this is "
+                "not recommended and may cause memory object being pinned "
+                "for 5 minutes"
+            )
```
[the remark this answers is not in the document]
ApostaC: `lock == False` and `lock == None` are the same

lmcache/v1/multiprocess/mp_storage_manager.py
```diff
@@ -30,13 +35,166 @@ class MemoryExhaustedError(Exception):
...
+        3. A lock can expire. Every time an object is "locked", the TTL of the
+        lock will be refreshed. Once the TTL expires, the object will become
+        fully unlocked, no matter how many "locks" are accumulated.
+    """
+
+    DEFAULT_TTL = 300  # 5 minutes
+
+    def __init__(self):
+        self._locks: dict[LockKey, LockMetadata] = {}
+        self._mutex = threading.Lock()
```
KuntaiDu, line 77: Is the performance of single mutex enough, given that mp_storage_manager will be accessed by multiple vLLM instances?

An alternative: create `n` locks, where lock i corresponds to `hash(key) % n == i`

lmcache/v1/multiprocess/mp_storage_manager.py
```diff
@@ -131,23 +325,57 @@ def reserve(
...
+        if objects is not None:
+            return handle, _confirm_reserve_objects(keys, ret_mask, objects, handle)
+
+        # Failed to allocate, try to evict once
+        # NOTE: we are doing very aggressive eviction here: every time
+        # we will try to evict num_objects_to_allocate objects and try
+        # to allocate again, until we cannot evict any more objects.
+        # NOTE: we cannot directly recycle the allocated objects in
+        # multi-process mode, because there could be multiple different
+        # models connecting to the same storage manager
```
KuntaiDu, line 342: We can let different models connect to different storage manager?

KuntaiDu (commented):
See comments

lmcache/v1/multiprocess/mp_storage_manager.py
ApostaC: Make sense. 
Also, given that this class is relatively simple, I would say it would be very easy to implement it in C++ (with std::atomic) and wrap it with pybinding.

lmcache/v1/multiprocess/mp_storage_manager.py
ApostaC: The core problem is not the storage manager, but the memory allocation itself. 
In multi-process mode, LMCache will be used as a global memory pool for the KV cache, and memory needs to be dynamically allocated from this pool. 
Additionally, the eviction should occur in the "global scope", which means that when allocating model A's KV cache, model B's KV cache may get evicted, and this prohibits directly reusing memory objects that were allocated for model B.

lmcache/v1/multiprocess/mp_storage_manager.py
```diff
@@ -131,23 +325,57 @@ def reserve(
...
+        with self._allocator_lock, self._buffer_lock:
+            while objects is None:
+                candidates = [
+                    c
+                    for c in self._cache_policy.get_evict_candidates(
+                        self._commited_memory_objects,
+                        num_objects_to_allocate,
+                    )
+                    if not self._obj_lock_manager.is_locked(c)
```
ApostaC: This could be removed because it's checked in `get_evicted_candidates`

lmcache/v1/multiprocess/mp_storage_manager.py
```diff
@@ -131,23 +325,57 @@ def reserve(
...
+                ]
+
+                # If the candidates are not enough, break
+                if len(candidates) == 0:
```
ApostaC:
```suggestion
                if not candidates:
```

KuntaiDu (approved):
LGTM!

wz1qqx:
hello, when i try this feature on 8*RTX4090 with 8 vLLM Instance, each instance is deployed with tp1.
i run same benchmark repeatly, Subsequent benchmarks will likely exhibit a hang phenomenon. like below:
image
and benchmark like this:
image
There appears to be a memory leak or other issue.

ApostaC:
> hello, when i try this feature on 8*RTX4090 with 8 vLLM Instance, each instance is deployed with tp1. i run same benchmark repeatly, Subsequent benchmarks will likely exhibit a hang phenomenon. like below: image and benchmark like this: image There appears to be a memory leak or other issue.

@wz1qqx Hey, can you share the commands that launches vLLM and LMCache + the command for benchmarking? It helps to reproduce the issue. Thanks!

wz1qqx:
> > hello, when i try this feature on 8*RTX4090 with 8 vLLM Instance, each instance is deployed with tp1. i run same benchmark repeatly, Subsequent benchmarks will likely exhibit a hang phenomenon. like below: image and benchmark like this: image There appears to be a memory leak or other issue.
> 
> @wz1qqx Hey, can you share the commands that launches vLLM and LMCache + the command for benchmarking? It helps to reproduce the issue. Thanks!

thx your reply, here is the scripts i use:
docker images：lmcache/vllm-openai:nightly-2025-11-30
hardware: 8*RTX4090  w/ P2P
lmcache+vllm:
[lmcache-ipc-share.sh](https://github.com/user-attachments/files/23870519/lmcache-ipc-share.sh)
`bash lmcache-ipc-share.sh`

router cmd:
`python3 -m sglang_router.launch_router --policy random \
    --worker-urls http://localhost:8000 http://localhost:8001 http://localhost:8002 http://localhost:8003 http://localhost:8004 http://localhost:8005 http://localhost:8006 http://localhost:8007 \
    --host 127.0.0.1 --port 30000 \
    --max-concurrent-requests 80 \
    --queue-size 20`

benchmark: 
`vllm bench serve --model meta-llama/Llama-3.1-8B-Instruct --dataset-name random --random-input-len 3072 --random-output-len 200 --random-range-ratio 0.2 --num-prompts 200  --max-concurrency 80 --request-rate 10  --port 30000`

PR #2099, opened 2025-11-27
Ignore mock requests for DP attention

yoo-kumaneko:
When DP attention is used with MLA, mock requests are sent to some vLLM engines because the MoE components still require all-to-all communication. Engines with no real requests would otherwise block that communication, hence the need for mock requests.

On the LMCache side, however, these mock requests should be ignored. This PR adds the necessary handling.


yoo-kumaneko:
@maobaolong Would you like to take a look at this PR?

maobaolong (approved):
@yoo-kumaneko LGTM, as discussed and tested in our env.

maobaolong:
@chunxiaozheng Would you like to take a double check?

chunxiaozheng (approved):
lgtm

PR #2101, opened 2025-11-27
[P2PBackend] refactor handle peer requests

chunxiaozheng:
This PR has two functions:
1. close `self.async_peer_socket` when socket error happens, prevent `[2025-11-27 11:42:48,055] LMCache ERROR: Peer request handler crashed: Address already in use`
2. add `try-except` when handle requests, when error happens, we should send a msg rather than raise a exception. **I only separated the original code and added `try-except`, without changing the functionality**

### DIFF
--- a/lmcache/v1/storage_backend/p2p_backend.py
+++ b/lmcache/v1/storage_backend/p2p_backend.py
@@ -90,9 +90,13 @@ class BatchedLookupAndPutRetMsg(P2PMsgBase):
 
 
 class P2PErrorMsg(P2PMsgBase):
-    """Error message"""
+    """
+    Error message, return error code to client.
 
-    error: str
+    -1 represents unknown msg type;
+    """
+
+    error_code: int
 
 
 P2PMsg = Union[
@@ -330,7 +334,7 @@ class P2PBackend(StorageBackendInterface):
                 ret_msg = await self._handle_batched_lookup_and_put(msg)
             else:
                 logger.error("Unknown message type: %s", type(msg))
-                ret_msg = P2PErrorMsg(error="Unknown message type")
+                ret_msg = P2PErrorMsg(error_code=-1)
 
             logger.info(f"P2P transfer finished for request {monitor_req_id}")
             self.stats_monitor.on_p2p_transfer_finished(monitor_req_id)
--- a/lmcache/v1/storage_backend/p2p_backend.py
+++ b/lmcache/v1/storage_backend/p2p_backend.py
@@ -89,11 +89,18 @@ class BatchedLookupAndPutRetMsg(P2PMsgBase):
     num_read_chunks: int
 
 
+class P2PErrorMsg(P2PMsgBase):
+    """Error message"""
+
+    error: str
+
+
 P2PMsg = Union[
     BatchedLookupAndGetMsg,
     BatchedLookupAndGetRetMsg,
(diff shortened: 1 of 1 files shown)


maobaolong (commented):
@chunxiaozheng Thanks for fixing this! Would you like take a look at Gemini's comment, I think it make sense.

chunxiaozheng:
> @chunxiaozheng Thanks for fixing this! Would you like take a look at Gemini's comment, I think it make sense.

@maobaolong Thanks for your suggestion, I have updated!

maobaolong (approved):
Although the diffs of this PR is over one hundred, but look like you just move the original code into a method surrounded by `try-except`. It LGTM.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -308,89 +325,118 @@ async def _handle_peer_requests(self):
...
-                ret_msg = BatchedLookupAndGetRetMsg(
-                    num_hit_chunks=num_hit_chunks,
-                )
+    async def _handle_batched_lookup_and_get(
+        self, msg: BatchedLookupAndGetMsg
+    ) -> BatchedLookupAndGetRetMsg:
+        lookup_id = msg.lookup_id
+        mem_objs = None
+        try:
+            logger.info(
```
chunxiaozheng, line 350: I only add `try-except` here, the logic is not change

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -308,89 +325,118 @@ async def _handle_peer_requests(self):
...
-                    memory_objs=local_mem_objs,
-                )
 
-                ret_msg = BatchedLookupAndPutRetMsg(
-                    num_read_chunks=len(local_mem_objs),
+    async def _handle_batched_lookup_and_put(
+        self, msg: BatchedLookupAndPutMsg
+    ) -> BatchedLookupAndPutRetMsg:
+        try:
+            logger.info("Received P2P batched lookup and put msg")
```
chunxiaozheng, line 400: I only add try-except here, the logic is not change

chunxiaozheng:
@sammshen could you help take a look?

lmcache/v1/storage_backend/p2p_backend.py
maobaolong: Thanks for this comment, it really help to review your PR.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -89,11 +89,18 @@ class BatchedLookupAndPutRetMsg(P2PMsgBase):
     num_read_chunks: int
 
 
+class P2PErrorMsg(P2PMsgBase):
```
sammshen, line 92: nit: can we add what scenarios potentially trigger error in this doc string?

sammshen (approved):
LGTM!

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: @sammshen Thanks for your review! Under normal scenarios, this message would not be triggered. Currently, `peer handler` only supports `BatchedLookupAndGetMsg` and `BatchedLookupAndPutMsg`, if receive other type msgs, will return the `P2PErrorMsg`.

PR #2105, opened 2025-11-27
[BugFix] The variable name is incorrect. The prefiller configuration …

ENg-122:
The variable name is incorrect. The prefiller configuration for "Disaggregated prefill" should be "pd_buffer_device" instead of "nixl_buffer_device".

in v1/config.py 602-617 line:
    if self.enable_pd:
        assert self.pd_role is not None
        assert self.pd_buffer_size is not None
        assert self.pd_buffer_device is not None

        assert self.remote_url is None, "PD only supports remote_url=None"
        assert self.save_decode_cache is False, (
            "PD only supports save_decode_cache=False"
        )
        assert self.enable_p2p is False, "PD only supports enable_p2p=False"

**If applicable**:

- [√] this PR contains user facing changes - docs added
- [×] this PR contains unit tests


hickeyma (approved):
LGTM, thanks @ENg-122

maobaolong (approved):
@ENg-122 Good catch!

PR #2111, opened 2025-11-27
Optimize lookups in sync mode by skipping cached chunks.

ziruiliu:
**Why**
Please refer to #1922 , for those kv cache already computed in GPU memory, we do not need to issue lookup to lookup server or cache engine or storage manger.
However, async lookup and prefetch logic is more complex when evction, preemption happened if async taks in still going. So in this PR, we simply focus on optimizing lookup in sync mode

**What**
num_computed_tokens is passed to all types of lookup clients. But only `LMCacheLookupClient` and `LMCacheBypassLookupClient` actually handle this parameter, other types of clients simply ignore or pass down this parameter.
In these two lookup clients, when blending is not enabled, hashes will be generated by token database. Therefore we can simply skip these hashes in lookup client and send rest hashes to cache engine.
If blending is enabled, lookup client pass token ids and num_computed_tokens to cache engine and let cache engine skip chunks when hashes are calculated properly

**Test**
Tested with normal lookup client/server and also LMCacheBypassLookupClient.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1482,6 +1482,11 @@ class LMCacheConnectorV1Impl:
             logger.debug(f"Looking up cache for the first time for request {req_id}!")
             self._requests_priority[req_id] = getattr(request, "priority", 0)
 
+            # Align computed tokens once to avoid repeated chunk-size rounding downstream
+            aligned_num_computed_tokens = (
+                num_computed_tokens // self._lmcache_chunk_size * self._lmcache_chunk_size
+            )
+
             # token_ids = request.prompt_token_ids
             # all token ids covers the preemption case
             token_ids = request.all_token_ids
@@ -1502,7 +1507,7 @@ class LMCacheConnectorV1Impl:
                 token_ids,
                 lookup_id=req_id,
                 request_configs=request_configs,
-                num_computed_tokens=num_computed_tokens,
+                num_computed_tokens=aligned_num_computed_tokens,
             )
 
         if num_external_hit_tokens is None:
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1482,7 +1482,8 @@ class LMCacheConnectorV1Impl:
             logger.debug(f"Looking up cache for the first time for request {req_id}!")
             self._requests_priority[req_id] = getattr(request, "priority", 0)
 
-            # Align computed tokens once to avoid repeated chunk-size rounding downstream
+            # Align computed tokens once to avoid repeated
+            # chunk-size rounding downstream
             aligned_num_computed_tokens = (
                 num_computed_tokens
                 // self._lmcache_chunk_size
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1484,7 +1484,9 @@ class LMCacheConnectorV1Impl:
 
             # Align computed tokens once to avoid repeated chunk-size rounding downstream
             aligned_num_computed_tokens = (
-                num_computed_tokens // self._lmcache_chunk_size * self._lmcache_chunk_size
+                num_computed_tokens
+                // self._lmcache_chunk_size
+                * self._lmcache_chunk_size
             )
 
             # token_ids = request.prompt_token_ids
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -409,8 +409,9 @@ class ReqMeta:
         # For load operation: check whether the request is scheduled to load
         if load_spec is not None and load_spec.can_load:
             logger.debug(
-                "Scheduled to load %d tokens for request %s",
+                "Scheduled to load %d tokens (%d cached in vLLM) for request %s",
                 load_spec.lmcache_cached_tokens,
+                load_spec.vllm_cached_tokens,
                 tracker.req_id,
             )
         else:
@@ -1501,6 +1502,7 @@ class LMCacheConnectorV1Impl:
                 token_ids,
                 lookup_id=req_id,
                 request_configs=request_configs,
+                num_computed_tokens=num_computed_tokens,
             )
 
         if num_external_hit_tokens is None:
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -786,12 +786,10 @@ class LMCacheEngine:
             assert hashes is not None
             lookup_request_id = self.stats_monitor.on_lookup_request(sum(offsets))
 
-        # Skip the number of tokens that are already computed, align to chunk size
-        skip_n_tokens = (
-            num_computed_tokens // self.config.chunk_size * self.config.chunk_size
-        )
-
-        res = skip_n_tokens
+        # Skip the number of tokens that are already computed (aligned upstream to
+        # chunk size)
+        aligned_computed_tokens = num_computed_tokens
+        res = aligned_computed_tokens
         try:
             chunk_info_iterator = self.token_database.process_tokens(
                 tokens=tokens,
@@ -803,7 +801,7 @@ class LMCacheEngine:
             # TODO: support batched_contains when layerwise is enabled
             if self.use_layerwise:
                 for start, end, key in chunk_info_iterator:
-                    if end <= skip_n_tokens:
+                    if end <= aligned_computed_tokens:
                         continue
                     assert isinstance(key, CacheEngineKey)
 
@@ -834,7 +832,7 @@ class LMCacheEngine:
                 for chunk_info in chunk_info_iterator:
                     assert isinstance(chunk_info[2], CacheEngineKey)
                     start, end, _ = chunk_info
-                    if end <= skip_n_tokens:
+                    if end <= aligned_computed_tokens:
                         continue
                     chunk_info_list.append(chunk_info)
                     keys.append(chunk_info[2])
@@ -861,7 +859,8 @@ class LMCacheEngine:
         finally:
             # When num_computed_tokens is greater than a chunk, we skip
             # some tokens to reduce the number of lookup requests.
-            # It is possible that res = skip_n_tokens and no lookup is performed.
+            # It is possible that res equals aligned_computed_tokens and no lookup is
+            # performed.
             # In this case, using res as the number of hit tokens will overcount
             # the number of hit tokens.
             # TODO deprecate this metric and use retrieve metrics instead.
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -747,6 +747,7 @@ class LMCacheEngine:
         lookup_id: Optional[str] = None,
         pin: bool = False,
         request_configs: Optional[dict] = None,
+        num_computed_tokens: int = 0,
     ) -> int:
         """
         Checks the existence of KV cache of the tokens from the cache engine.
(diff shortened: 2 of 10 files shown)


ziruiliu:
Hi @YaoJiayi , I have created this new PR per our discussion earlier, please take a look. cc @maobaolong @yoo-kumaneko @sammshen

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -168,7 +168,12 @@ def lookup(
         token_ids: Union[torch.Tensor, list[int]],
         lookup_id: str,
         request_configs: Optional[dict] = None,
+        num_computed_tokens: int = 0,
     ) -> Optional[int]:
+        # TODO num_computed_tokens is passed in
```
maobaolong: It would be nice if you could add your id to the todo comment.
Something like `TODO(zirui): num_computed_tokens is passed in`

maobaolong (approved):
LGTM otherwise, only a nit left

maobaolong:
@YaoJiayi Would you like to take a double check?

lmcache/v1/cache_engine.py
```diff
@@ -766,7 +770,12 @@ def lookup(
             assert hashes is not None
             lookup_request_id = self.stats_monitor.on_lookup_request(sum(offsets))
 
-        res = 0
+        # Skip the number of tokens that are already computed, align to chunk size
+        skip_n_tokens = (
+            num_computed_tokens // self.config.chunk_size * self.config.chunk_size
+        )
```
YaoJiayi: Why is this needed?

lmcache/v1/cache_engine.py
```diff
@@ -766,7 +770,12 @@ def lookup(
...
+
+        res = skip_n_tokens
```
YaoJiayi: Why is needed?

YaoJiayi: I thought res should be initialized as 0? I thought you use `result + skip_n_tokens` in the lmcache_lookup_client?

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -169,23 +169,39 @@ def lookup(
...
         # because the blender need the input embedding.
         if not self.enable_blending:
             hashes = []
             offsets = []
+
+            # We already have hashes here so we can skip the chunks that are already
+            # in GPU cache. Don't pass num_computed_tokens to lookup server.
+            skip_n_tokens = (
+                num_computed_tokens // self.config.chunk_size * self.config.chunk_size
+            )
```
YaoJiayi: Shouldn't this be handled in vllm_v1_adapter?

YaoJiayi (commented):
Thanks @ziruiliu ! Left some comments

lmcache/v1/cache_engine.py
ziruiliu: res calculation in cache_engine, lookup client, and lookup_client_bypass are different.
lookup_client:
results list is initialized empty. Each result received from lookup server represents the additional tokens found in engine, therefore we have to add skip_n_tokens and number of found tokens, which is result + skip_n_tokens
image


lookup_client_bypass:
res is initialized as skip_n_token. If there are N more chunks to check, it calls engine's lookup, which returned the number of additional tokens found in engine. Return value should be the sum of skipped tokens + found tokens
image


cache_engine:
res is initialized as skip_n_token. If all tokens are in vllm gpu cache, no lookup needs to be issued and return value is skip_n_token. If there are N more chunks to be lookup, res is set to the last chunk's `end` value
image
The `end` value is the absolute value of the offset and is greater than skip_n_token, so return value is now `end`

The 3 calculation methods above are independent of each other. In any given situation, only one of the algorithms will be used — there will never be a case where a computation is performed in both lookup_client and cache_engine

lmcache/v1/lookup_client/lmcache_lookup_client.py
ziruiliu: Good idea. I will move all these skip_n_token calculation up to vllm_v1_adpater

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
ziruiliu: changed

lmcache/v1/cache_engine.py
ziruiliu: I will move all 3 calculation of skip_n_token to vllm_v1_adapter

ziruiliu:
Hi @YaoJiayi, I moved those "skip_n_token = ..." to vllm_v1_adapter.py, pass aligned number of tokens down to lookup client
and cache engine. I also replied about res calculation in your comment. Please check

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1387,6 +1388,11 @@ def get_num_new_matched_tokens(
 
         self._requests_priority[req_id] = getattr(request, "priority", 0)
 
+        # Align computed tokens once to avoid repeated chunk-size rounding downstream
```
sammshen: this is great!

lmcache/v1/cache_engine.py
```diff
@@ -806,9 +815,14 @@ def lookup(
...
                 for chunk_info in chunk_info_iterator:
                     assert isinstance(chunk_info[2], CacheEngineKey)
+                    start, end, _ = chunk_info
+                    if end <= aligned_computed_tokens:
+                        continue
                     chunk_info_list.append(chunk_info)
                     keys.append(chunk_info[2])
-
+                # If no tokens to lookup, return immediately
+                if len(keys) == 0:
```
sammshen: nit: 
`if not keys:`

lmcache/v1/cache_engine.py
```diff
@@ -827,6 +841,13 @@ def lookup(
             # all tokens where found, return the maximal end
             return res
         finally:
+            # When num_computed_tokens is greater than a chunk, we skip
+            # some tokens to reduce the number of lookup requests.
+            # It is possible that res equals aligned_computed_tokens and no lookup is
+            # performed.
+            # In this case, using res as the number of hit tokens will overcount
```
sammshen, line 866: this comment is useful

lmcache/v1/lookup_client/lmcache_lookup_client_bypass.py
```diff
@@ -58,25 +58,37 @@ def lookup(
...
                 for start, end, key in self.token_database.process_tokens(
                     token_ids, make_key=False
                 ):
+                    if end <= aligned_computed_tokens:
+                        continue
                     hashes.append(key)
                     offsets.append(end - start)
+                # Return aligned_computed_tokens immediately if there is no token to
+                # lookup
+                if len(hashes) == 0:
```
sammshen: nit: 
`if not hashes`

lmcache/v1/cache_engine.py
```diff
@@ -766,7 +770,10 @@ def lookup(
             assert hashes is not None
             lookup_request_id = self.stats_monitor.on_lookup_request(sum(offsets))
 
-        res = 0
+        # Skip the number of tokens that are already computed (aligned upstream to
+        # chunk size)
+        aligned_computed_tokens = num_computed_tokens
```
sammshen, line 791: this is a nice renaming

lmcache/v1/cache_engine.py
```diff
@@ -806,9 +815,14 @@ def lookup(
                 keys = []
                 for chunk_info in chunk_info_iterator:
                     assert isinstance(chunk_info[2], CacheEngineKey)
+                    start, end, _ = chunk_info
+                    if end <= aligned_computed_tokens:
+                        continue
                     chunk_info_list.append(chunk_info)
                     keys.append(chunk_info[2])
```
sammshen, line 840: nit:
I know this is not from your PR but could we add a comment on what `chunk_info[2]` is?

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -169,23 +169,37 @@ def lookup(
...
         request_configs: Optional[dict] = None,
+        num_computed_tokens: int = 0,
     ) -> Optional[int]:
         lookup_id_buf = lookup_id.encode("utf-8")
         request_configs_str = ""
         if request_configs is not None and len(request_configs) != 0:
             request_configs_str = json.dumps(request_configs)
         request_configs_buf = request_configs_str.encode("utf-8")
+        num_computed_buf = num_computed_tokens.to_bytes(8, "big", signed=False)
+        aligned_computed_tokens = num_computed_tokens  # pre-aligned in adapter
```
sammshen, line 186: thanks for rename and comment

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -169,23 +169,37 @@ def lookup(
...
             for start, end, key in self.token_database.process_tokens(
                 token_ids, make_key=False
             ):
+                if end <= aligned_computed_tokens:
+                    continue
                 hashes.append(key)
                 offsets.append(end - start)
+            # Return aligned_computed_tokens immediately if there is no token to
+            # lookup
+            if len(hashes) == 0:
```
sammshen: nit:
`if not hashes`

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -313,15 +328,18 @@ def process_request():
                         lookup_id=lookup_id,
                         pin=True,
                         request_configs=request_configs,
+                        num_computed_tokens=0,
```
sammshen, line 350: why we pass in 0 for regular (non-blending) case?

sammshen: oh, I see this work is done in client for non-blending and in cache engine for blending, is this right? @ziruiliu

sammshen: anyways, can we clarify this logic with a comment

lmcache/v1/cache_engine.py
YaoJiayi: Hmm. What's the problem of initializing res as 0 in cache engine? What will go wrong in this case?

lmcache/v1/lookup_client/lmcache_lookup_client.py
ziruiliu: because if blending is not enabled, we already have hashes in lookup() in lookup client.
So we can simply skip these tokens in gpu cache in lookup client, the hashes passed down to cache_engine are those not in gpu cache, so we do not need cache engine to skip any more tokens.

lmcache/v1/cache_engine.py
ziruiliu: will change in next commit

lmcache/v1/lookup_client/lmcache_lookup_client_bypass.py
ziruiliu: will change in next commit

lmcache/v1/cache_engine.py
ziruiliu: added comment about this `chunk_info[2]`, it's the key

sammshen (approved):
lgtm

maobaolong:
@ziruiliu It's time to merge this improvement, I pressed the button of `auto merge`.

lmcache/v1/cache_engine.py
```diff
@@ -771,6 +772,9 @@ def lookup(
 
         :param Optional[dict] request_configs: the configs of the request.
 
+        :param int num_computed_tokens: Number of leading tokens those are already
```
sammshen, line 775: @ziruiliu this definition is not consistent if we pass in the aligned value. this is a big problem that we need to fix now. I will create a PR shortly. this is a cause of correctness issues right now

lmcache/v1/cache_engine.py
sammshen: I made a mistake, this renaming is actually bad

PR #2128, opened 2025-12-01
Add k8s example for multi process mode

ningziwen:

Adds Kubernetes deployment example for LMCache multi-process mode with vLLM using a sidecar container pattern.

The existing documentation only covers local and Docker deployments. This PR adds:
- Working Kubernetes YAML configuration (`examples/multi_process/multi_process.yaml`)
- Comprehensive Kubernetes deployment section in the multi-process mode documentation
- Step-by-step deployment, testing, and troubleshooting instructions

**Key differences from Docker deployment:**

Kubernetes requires special configuration due to GPU allocation model differences:

1. **Pod-level GPU allocation**: Both containers must see all 4 GPUs (via `CUDA_VISIBLE_DEVICES=0,1,2,3`) to avoid device ID mismatches
2. **hostIPC: true**: Required for IPC-based GPU memory sharing between containers
3. **GPU requirement for LMCache server**: The server needs GPU access to perform IPC-based GPU-to-CPU memory transfers (not just CPU-only as the name "CPU offloading" might suggest)

**Configuration details:**
- Uses `lmcache/standalone:nightly` and `lmcache/vllm-openai:latest-nightly` images
- Allocates 4 GPUs per pod (shared between containers)
- LMCache server: 60GB CPU buffer, 4 worker threads
- vLLM: TP=4, max-model-len=32768, gpu-memory-utilization=0.85
- Tested with Llama-3.1-8B-Instruct model

**Testing performed:**
- Deployed on Kubernetes cluster with 4 GPUs per node
- Verified KV cache registration and unregistration in logs
- Confirmed KV cache offloading (store) and retrieval on repeated requests

**If applicable**:

### DIFF
--- a/docs/source/kv_cache/multiprocess_mode.rst
+++ b/docs/source/kv_cache/multiprocess_mode.rst
@@ -168,7 +168,7 @@ Prerequisites
     kubectl apply -f examples/multi_process/multi_process.yaml
 
 .. note::
-    The default model is ``Qwen/Qwen2.5-1.5B``, which does not require a Hugging Face token. If you want to use a gated model (like Llama), you need to:
+    The default model is ``Qwen/Qwen3-14B``, which does not require a Hugging Face token. If you want to use a gated model (like Llama), you need to:
     
     1. Add a Secret with your Hugging Face token:
     
@@ -228,9 +228,9 @@ Send a test request with repeated prompts:
     curl -X POST http://localhost:8000/v1/completions \
       -H "Content-Type: application/json" \
       -d "{
-        \"model\": \"Qwen/Qwen2.5-1.5B\",
+        \"model\": \"Qwen/Qwen3-14B\",
         \"prompt\": \"$(printf 'Explain the significance of KV cache in language models.%.0s' {1..100})\",
-        \"max_tokens\": 100
+        \"max_tokens\": 10
       }"
 
 On the first request, check LMCache server logs for:
--- a/docs/source/kv_cache/multiprocess_mode.rst
+++ b/docs/source/kv_cache/multiprocess_mode.rst
@@ -140,6 +140,138 @@ Once both containers are running, you can send requests to vLLM the same way as 
         \"max_tokens\": 10
     }"
 
+Kubernetes Deployment
+---------------------
+
+You can deploy LMCache and vLLM on Kubernetes using a sidecar container pattern. This approach is suitable for production deployments where you want to leverage Kubernetes orchestration.
+
+.. note::
+    The Kubernetes deployment requires special configuration due to GPU allocation differences from Docker. Both containers must see all GPUs to avoid device ID mismatches in IPC-based GPU memory operations.
+
+Prerequisites
+~~~~~~~~~~~~~
+
+- Kubernetes cluster with GPU support (NVIDIA GPU Operator installed)
+- At least 4 GPUs per node
+- ``kubectl`` configured to access your cluster
+- Hugging Face token (for gated models like Llama)
+
+**Step 1: Create Namespace**
+
+.. code-block:: bash
+
+    kubectl create namespace multi-process
+
+**Step 2: Configure Hugging Face Token**
+
+Edit ``examples/multi_process/multi_process.yaml`` and replace the placeholder token. The `data` field in the Secret requires the token to be base64 encoded. You can generate it using `echo -n "your_hf_token_here" | base64`.
+
+.. code-block:: yaml
+
+    data:
+      hf_token: YOUR_BASE64_ENCODED_HF_TOKEN_HERE
+
+Alternatively, create the secret separately:
+
+.. code-block:: bash
+
+    kubectl create secret generic vllm-secrets \
+      --from-literal=hf_token=your_hf_token_here \
+      -n multi-process
+
+**Step 3: Deploy the Application**
+
+.. code-block:: bash
+
+    kubectl apply -f examples/multi_process/multi_process.yaml
+
+**Step 4: Monitor Deployment**
+
+Check pod status:
+
+.. code-block:: bash
+
+    kubectl get pods -n multi-process -w
+
+Check LMCache server logs:
+
+.. code-block:: bash
+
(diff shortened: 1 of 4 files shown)


docs/source/kv_cache/multiprocess_mode.rst
```diff
@@ -140,6 +140,138 @@ Once both containers are running, you can send requests to vLLM the same way as
...
+- ``CUDA_VISIBLE_DEVICES`` is set to ``0,1,2,3`` for both containers
+- This avoids device ID mismatches that occur with per-container GPU allocation
+
+**Why LMCache Server Needs GPU Access:**
+
+The LMCache server requires GPU access because it:
+
+- Receives GPU tensor IPC handles from vLLM workers
+- Directly accesses vLLM's GPU memory to perform GPU-to-CPU transfers
+- Uses CUDA operations for efficient memory copying
```
sammshen: also potentially allocates intermediate gpu buffers

ningziwen: @sammshen Added.

sammshen (approved):
LGTM, thanks @ningziwen !

examples/multi_process/multi_process.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - "60"
+            - --max-workers
+            - "4"
+          resources:
+            requests:
+              memory: "10Gi"
+              cpu: "4"
+            limits:
+              memory: "120Gi"
+              cpu: "8"
```
ApostaC: Dumb question: it seems like we are not allocating GPUs here, and we did not specify `runtimeClassName` either. How does the Kubernetes engine know it needs to let the LMCache-server pod access the GPU?

docs/source/kv_cache/multiprocess_mode.rst
```diff
@@ -140,6 +140,139 @@ Once both containers are running, you can send requests to vLLM the same way as
...
+    data:
+      hf_token: YOUR_BASE64_ENCODED_HF_TOKEN_HERE
+
+Alternatively, create the secret separately:
+
+.. code-block:: bash
+
+    kubectl create secret generic vllm-secrets \
+      --from-literal=hf_token=your_hf_token_here \
+      -n multi-process
```
ApostaC: Can we make this step optional (probably needs some changes in the multi_process.yaml as well)? This makes the process simpler.

We can use qwen-series models which do not require hf tokens.

ApostaC (commented):
Otherwise LGTM! Appreciate the help!

docs/source/kv_cache/multiprocess_mode.rst
ningziwen: Changed.

examples/multi_process/multi_process.yaml
ningziwen: Good question. Looks like GPU is always allocated to whole pod even requested at container level. Lmcache-server is a sidecar container in same pod so have access to it.

Actually this fact advises sidecar container may be the only efficient approach until we remove GPU from LMCache server requirement. Since if we run LMCache server in separate pod, there will have to be dedicated GPU assigned to LMCache server, causing waste.

I added some explanation in comments as well.

ningziwen:
> Otherwise LGTM! Appreciate the help!

@ApostaC Addressed comments.

examples/multi_process/multi_process.yaml
ApostaC: Oh, I see what you mean! 
The problem is that if we have multiple vLLM pods on the same node, there should be only one LMCache server. In this case, how does the sidecar solution work?

ningziwen: I've been testing the DaemonSet approach where multiple vLLM pods share one LMCache server per node. Network connectivity works and requests reach LMCache successfully, but I'm hitting a CUDA error: mapping of buffer object failed (cudaErrorMapBufferObjectFailed) when vLLM tries to register KV caches. Both pods see the same GPUs with identical UUIDs and have hostIPC: true, but it seems CUDA IPC handles created in one pod can't be opened in another pod due to container isolation. Still investigating potential solutions.

```
[2025-12-04 10:58:59,571] LMCache ERROR: Error handling request RequestType.REGISTER_KV_CACHE: CUDA error: mapping of buffer object failed
Search for `cudaErrorMapBufferObjectFailed' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

ApostaC: Thanks! This is very useful! 
I'm not a k8s expert, but I think what we want here is to enable "cuda IPC" across different pods. Not sure if that will help

ApostaC: Will be happy to merge the PR once this is addressed!

ningziwen: @ApostaC Pushed new revision. Now LMCache server is running as a separated daemonset. So multiple vllm pods should be able to interact with one LMCache server now. The yamls are the minimum working configs. Fortunately we don't need more permissions in LMCache server like hostIPC, priviledged. Mounting /dev/shm is sufficient.

examples/multi_process/lmcache-daemonset.yaml
```diff
@@ -0,0 +1,49 @@
...
+            - "60"
+            - --max-workers
+            - "4"
+          volumeMounts:
+            - name: dev-shm
+              mountPath: /dev/shm
+          resources:
+            requests:
+              memory: "10Gi"
+              cpu: "4"
```
ApostaC, line 44: Dumb question: do we need to make the requested memory larger than 60GB in order for LMCache to grab 60GB cpu memory buffer?

ApostaC (approved):
Left a small comment, otherwise LGTM!

examples/multi_process/lmcache-daemonset.yaml
ningziwen: Good point. Changed.

PR #2139, opened 2025-12-02
[Core] add new batched copy cuda kernel

ApostaC:

This PR adds new memory copy kernels using cuda batched memcpy.

The new kernel requires cuda > 12.8

**If applicable**:


ApostaC:
This kernel does not work well under high concurrency. Closing this PR..

PR #2140, opened 2025-12-02
[bugfix] fix unpin error

chunxiaozheng:
When storage backends is more than 1, the `lookup` may pin keys in different backends, when `lookup_unpin` is called, we should unpin keys in the backend which the keys has pinned before, rather than unpin all keys in all backends.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -198,8 +198,10 @@ class LMCacheEngine:
         # NOTE(ApostaC): we haven't support lookup-cache yet
         self.lookup_cache: dict[CacheEngineKey, Any] = {}
 
-        # lookup_id -> [pinned keys]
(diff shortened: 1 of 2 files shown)


maobaolong (approved):
Good catch! Thanks for this bug fix!

YaoJiayi (approved):
lgtm

lmcache/v1/cache_engine.py
```diff
@@ -198,8 +198,10 @@ def __init__(
         # NOTE(ApostaC): we haven't support lookup-cache yet
         self.lookup_cache: dict[CacheEngineKey, Any] = {}
 
-        # lookup_id -> [pinned keys]
-        self.lookup_pins: dict[str, list] = defaultdict(list)
+        # lookup_id -> {location -> [pinned keys]}
```
sammshen, line 201: thanks for updating the comment

PR #2146, opened 2025-12-02
[critical patch]: refactor async and sync lookup cache

sammshen:
# Description
Previously, the preemption refactor (loading KV caches for preempted requests) https://github.com/LMCache/LMCache/pull/2007 broke the async robustness for high concurrency.

In this PR, we
1. refactor `lookup_cache` (THIS ALSO FIXES A BUG IN SYNC CASE WHERE WE PASS BACK A NEGATIVE NUMBER OF TOKENS)
2. fix the `len(tokens) == len(slot_mappings)` assertion error bug
3. change async multi serializer to async single serializer (performance gain is maintained) CRITICAL: releases the `save_unfull_chunk=False` requirement

# A question that we considered: 
should we remove the async serializer abstraction? 

Potential Motivation: we want to support `save_unfull_chunk` which is a forced assumption by the async serializer because it needs to make the strong assumption that 50% of the cpu memory pool is usable

Answer: No, but we should redesign it less greedily. We refactor it to the async single serializer

# Basic reasoning: 

Sync Case: 
- many stores happen at the same time. this is ok because `busy_loop` is false so they will just not store if there is no more space
- one retrieve at a time. this is ok because `busy_loop` will make the retrieve wait for stores to finish

Async Case: 
- same logic for stores
- many retrieves at the same time. we need some sort of serializer here to ensure that all the `busy_loop` True retrieves (we cannot set False or else we might not retrieve what we promised vllm in lookup) don't collide with each other. essentially, they have to lease how many memory chunks (full) that they want to take from the pool
 
 
# Benchmark ran to test the preservation of performance on the single serializer in high concurrency case: 

```yaml
# mock.yaml
chunk_size: 256
local_cpu: False
max_local_cpu_size: 40
save_unfull_chunk: False
enable_async_loading: True
remote_url: "mock://240/?peeking_latency=0&read_throughput=2&write_throughput=2"
remote_serde: "naive"
```

```bash
LMCACHE_LOG_LEVEL=DEBUG \
CUDA_VISIBLE_DEVICES=1 \
PYTHONHASHSEED=0 \
LMCACHE_CONFIG_FILE=mock.yaml \
  vllm serve meta-llama/Llama-3.1-8B-Instruct \
  --kv-transfer-config   '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}' \
  --no-enable-prefix-caching > output.log 2>&1
```

```bash
python benchmarks/long_doc_qa/long_doc_qa.py \
      --num-documents 15 \
      --document-length 100000 \
      --output-len 100 \
      --repeat-count 1 \
      --repeat-mode tile \
      --model  meta-llama/Llama-3.1-8B-Instruct \
      --max-inflight-requests 15 \
      --eos-token-id 128001
```

## Multi-Serializer (and had some retrieved less than lookup errors): 
```text
Warmup round mean TTFT: 61.303s
Warmup round time: 116.958s
Warmup round prompt count: 15
Warmup round successful prompt count: 15

=== BENCHMARK RESULTS ===
Query round mean TTFT: 56.426s
Query round time: 100.363s
Query round prompt count: 15
Query round successful prompt count: 15
```

## Single Serializer
```text
Warmup round mean TTFT: 61.002s
Warmup round time: 116.505s
Warmup round prompt count: 15
Warmup round successful prompt count: 15

=== BENCHMARK RESULTS ===
Query round mean TTFT: 47.969s
Query round time: 88.854s
Query round prompt count: 15
Query round successful prompt count: 15
```

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1462,33 +1462,38 @@ class LMCacheConnectorV1Impl:
 
         req_id = request.request_id
 
-        # consult the cache before any processing
-        if cached_num_hit_toks := self.lookup_client.lookup_cache(lookup_id=req_id):
-            return cached_num_hit_toks
+        if (
+            num_external_hit_tokens := self.lookup_client.lookup_cache(lookup_id=req_id)
+        ) != -1:
+            logger.debug(
+                f"Found {num_external_hit_tokens} hit tokens for request"
+                f" {req_id} in the lookup cache."
+            )
+        else:
+            logger.debug(f"Looking up cache for the first time for request {req_id}!")
+            self._requests_priority[req_id] = getattr(request, "priority", 0)
 
-        self._requests_priority[req_id] = getattr(request, "priority", 0)
+            # token_ids = request.prompt_token_ids
+            # all token ids covers the preemption case
+            token_ids = request.all_token_ids
 
-        # token_ids = request.prompt_token_ids
-        # all token ids covers the preemption case
-        token_ids = request.all_token_ids
+            # If the request has multimodal hashes, apply them to the token ids
+            mm_hashes, mm_positions = extract_mm_features(request)
+            if mm_hashes and mm_positions:
+                # TODO(Jiayi): Optimize this
+                token_ids = torch.tensor(request.prompt_token_ids)
+                apply_mm_hashes_to_token_ids(token_ids, mm_hashes, mm_positions)
+                token_ids = token_ids.tolist()
 
-        # If the request has multimodal hashes, apply them to the token ids
-        mm_hashes, mm_positions = extract_mm_features(request)
-        if mm_hashes and mm_positions:
-            # TODO(Jiayi): Optimize this
-            token_ids = torch.tensor(request.prompt_token_ids)
-            apply_mm_hashes_to_token_ids(token_ids, mm_hashes, mm_positions)
-            token_ids = token_ids.tolist()
+            request_configs = extract_request_configs(request.sampling_params)
+            if self.skip_last_n_tokens > 0:
+                token_ids = token_ids[: -self.skip_last_n_tokens]
 
-        request_configs = extract_request_configs(request.sampling_params)
-        if self.skip_last_n_tokens > 0:
-            token_ids = token_ids[: -self.skip_last_n_tokens]
-
-        num_external_hit_tokens = self.lookup_client.lookup(
-            token_ids,
-            lookup_id=req_id,
-            request_configs=request_configs,
-        )
+            num_external_hit_tokens = self.lookup_client.lookup(
+                token_ids,
+                lookup_id=req_id,
+                request_configs=request_configs,
+            )
 
         if num_external_hit_tokens is None:
             logger.debug(
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -141,8 +141,6 @@ class RequestTracker:
 
     # The block ids that has been allocated so far
     # NOTE: allocated blocks could be more than the number of tokens
-    # FIXME: need to check whether the block ids will be changed after
-    #        preemption
     allocated_block_ids: list[int]
 
     # The number of tokens that has been saved
@@ -256,6 +254,12 @@ class RequestTracker:
             # reset the number of saved tokens
             self.num_saved_tokens = lmcache_cached_tokens
             # we don't need to extend the token ids in the preempted case
+            # however, it is possible for the scheduled tokens of the request
+            # to be less than the total number of tokens (partial cache hit)
+            # so we may need to truncate
+            self.token_ids = self.token_ids[
+                : lmcache_cached_tokens + len(new_token_ids)
+            ]
         else:
             self.allocated_block_ids.extend(new_block_ids)
             self.token_ids.extend(new_token_ids)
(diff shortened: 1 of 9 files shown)


sammshen:
@maobaolong @chunxiaozheng I plan to fix the async codepath fully in this PR. I will let you know when it is fully ready

maobaolong:
@yoo-kumaneko Please track this PR and take a review after it is ready, thanks!

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -256,6 +254,12 @@ def update(
             # reset the number of saved tokens
             self.num_saved_tokens = lmcache_cached_tokens
             # we don't need to extend the token ids in the preempted case
+            # however, it is possible for the scheduled tokens of the request
```
sammshen, line 257: this is a bug in the preemption case when we have partial cache hit that was not caught before

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1462,33 +1470,38 @@ def get_num_new_matched_tokens(
...
-        request_configs = extract_request_configs(request.sampling_params)
-        if self.skip_last_n_tokens > 0:
-            token_ids = token_ids[: -self.skip_last_n_tokens]
-
-        num_external_hit_tokens = self.lookup_client.lookup(
-            token_ids,
-            lookup_id=req_id,
-            request_configs=request_configs,
-        )
+        if (
```
sammshen, line 1473: this branch logic fixes the passing back negative tokens

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1462,33 +1470,38 @@ def get_num_new_matched_tokens(
...
+            num_external_hit_tokens := self.lookup_client.lookup_cache(lookup_id=req_id)
+        ) != -1:
```
sammshen, line 1475: this -1 refactors the logic from lookup_cache

sammshen:
@maobaolong @chunxiaozheng @yoo-kumaneko 

This PR is ready for review. 

It is pretty critical and also changes the serializer so we no longer are forced to have `save_unfull_chunk=False`

lmcache/v1/lookup_client/lmcache_lookup_client.py
```diff
@@ -162,7 +162,12 @@ def _recreate_socket(self) -> None:
             self.sockets[rank_idx] = new_socket
 
     def lookup_cache(self, lookup_id: str) -> Optional[int]:
-        return self.reqs_status.get(lookup_id, None)
+        """
+        "-1 means not found;
```
ApostaC, line 166: What's the difference between -1 and 0?

lmcache/v1/config.py
```diff
@@ -569,14 +569,15 @@ def _post_init(self):
...
-        self.save_unfull_chunk = False
+    # needed for the old async serializer implementation
+    # # auto-adjust save_unfull_chunk for async loading to prevent CPU fragmentation
+    # if self.enable_async_loading:
+    #     logger.warning(
+    #         "Automatically setting save_unfull_chunk=False because "
+    #         "enable_async_loading=True or use_layerwise=True to prevent "
+    #         "CPU memory fragmentation"
+    #     )
+    #     self.save_unfull_chunk = False
```
ApostaC, line 580: Is it safe to remove this?

lmcache/v1/cache_engine.py
```diff
@@ -1207,8 +1211,12 @@ def _async_process_tokens_internal(
...
         future = self.event_manager.pop_event(EventType.LOADING, kwargs["req_id"])
 
-        memory_objs = future.result()
-        memory_objs = [mm for m in memory_objs for mm in m]
+        try:
+            memory_objs = future.result()
+            memory_objs = [mm for m in memory_objs for mm in m]
+        except Exception as e:
+            logger.error(f"Error popping event for request {kwargs['req_id']}: {e}")
+            return [], 0
```
ApostaC, line 1219: Do we need to change ref count for the existing memory objs when error happens?

ApostaC (commented):
The vllm_v1_adapter is becoming non-managable unless we spend a lot of time investigating it... Didn't spend too much time on that, so please make sure you have stress-tested it.

Otherwise LGTM in general with a few minor comments.

lmcache/v1/lookup_client/lmcache_lookup_client.py
sammshen: -1 is not found so we cannot use the lookup cache and 0 means we actually hit 0 tokens. this was defined by jiayi's async lookup client

lmcache/v1/config.py
sammshen: yes, finally refactored the async codepath and this was a big request from @maobaolong and @chunxiaozheng before to make async compatible with P2P

lmcache/v1/cache_engine.py
sammshen: we do not because the error happens if the future cannot be unpacked

sammshen:
@ApostaC thanks for the review!

I stress tested with the super high concurrency (100k token contexts) in the PR description

lmcache/v1/lookup_client/lmcache_lookup_client.py
yoo-kumaneko: So None is replaced by -1, right?

sammshen: yes, this matches the original semantics defined by Jiayi

yoo-kumaneko:
How did the previous refactor break the async robustness?

sammshen:
@yoo-kumaneko Previously, we did not handle preemption in the asynchronous case so there was no issue

lmcache/v1/lookup_client/lmcache_async_lookup_client.py
```diff
@@ -159,23 +159,19 @@ def __init__(
...
-
-    # TODO(Jiayi): Consider batching here
-    def lookup(
-        self,
-        token_ids: Union[torch.Tensor, list[int]],
-        lookup_id: str,
-        request_configs: Optional[dict] = None,
-    ) -> Optional[int]:
+        """
+        -1 means not found;
```
sammshen, line 163: be clear that this is lookup cache not found

ApostaC (approved):
LGTM!

Shaoting-Feng (approved):

PR #2148, opened 2025-12-02
[refactor] refactor p2p backend mapping

chunxiaozheng:
There are too many mapping `dict` in `P2PBackend`, the code looks quite difficult, by abstracting a class, a lot of code can be simplified.

### DIFF
--- a/lmcache/v1/storage_backend/p2p_backend.py
+++ b/lmcache/v1/storage_backend/p2p_backend.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, List, Optional, Sequence, Union
 import asyncio
 import enum
@@ -125,6 +126,16 @@ P2PMsg = Union[
 ]
 
 
+@dataclass
+class PeerInfo:
+    """Peer information"""
+
+    peer_init_url: str  # peer id
+    peer_lookup_url: str
+    lookup_lock: asyncio.Lock
+    lookup_socket: zmq.asyncio.Socket
+
+
 # TODO(Jiayi): handle asymmetric TP.
 class P2PBackend(StorageBackendInterface):
     def __init__(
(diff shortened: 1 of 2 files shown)


lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -169,11 +180,11 @@ def __init__(
...
         self.local_lookup_cache: dict[int, tuple[str, str, str]] = {}
-        # A set of peer_init_urls
-        self.peer_id_to_lookup_url_mapping: dict[str, str] = {}
-
-        # A lookup_id -> (peer_init_url, peer_lookup_url, location)
-        self.lookup_id_to_peer_mapping: dict[str, tuple[str, str, str]] = {}
+        # the peer info mapping
+        self.peer_info_mapping: dict[str, PeerInfo] = {}
+        # A lookup_id -> (peer_init_url, location)
+        # TODO: location is not used for now
```
maobaolong: If this is a TODO comment, it would be better if you add you name here.
e.g. TODO(chunxiaozheng) : location is not used for now

maobaolong: BTW, the peer info here and the L171~L176, make concept confusing, these two places are different, but we all call it peer.

maobaolong (commented):
@chunxiaozheng Left a couple of comments.

Would you like to forward a step to refactor the concept of `peer` here also?

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: @maobaolong Thanks for your suggestion, I have updated!

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -450,67 +458,59 @@ async def _handle_batched_lookup_and_put(
...
-            new_peer_lookup_url,
+            peer_lookup_url,
             "tcp",
             zmq.REQ,
             "connect",
             self.socket_recv_timeout_ms,
             self.socket_send_timeout_ms,
         )
-        self.lookup_url_to_socket_mapping[new_peer_lookup_url] = new_lookup_socket
+        if peer_info is not None:
```
maobaolong, line 503: We can put this operation into `PeerInfo`

chunxiaozheng: done

maobaolong (approved):
LGTM

KuntaiDu (approved):
The code is getting easier to understand. LGTM!

PR #2169, opened 2025-12-04
[#2168] Add freeze mode for cache_engine

maobaolong:
[#2168](https://github.com/LMCache/LMCache/issues/2168)

### DIFF
--- a/docs/source/controller/freeze_mode.rst
+++ b/docs/source/controller/freeze_mode.rst
@@ -1,0 +1,95 @@
+Freeze Mode
+==========
+
+Overview
+--------
+
+Freeze Mode is a safety mechanism in LMCache Controller designed to prevent data inconsistency and potential data loss during critical system events. When the controller detects severe state inconsistencies or is undergoing restart/recovery, it can activate Freeze Mode to temporarily restrict certain operations.
+
+Motivation
+----------
+
+The primary motivations for Freeze Mode are:
+
+1. **Controller Failure Recovery**
+   - When the controller crashes and restarts, it may need to perform a full synchronization with workers
+   - During this sync period, the system state may be incomplete or inconsistent
+
+2. **State Inconsistency Detection**
+   - When severe state mismatches are detected between the controller and workers
+   - This could happen due to network issues, worker failures, or unexpected system behavior
+
+3. **Full Synchronization Safety**
+   - During full sync operations triggered by the above events
+   - To prevent operations that could cause eviction or admission events while state is being rebuilt
+
+Purpose
+-------
+
+When Freeze Mode is activated:
+
+- **All operations that could generate evict/admit events are disabled**
+- Only read operations and basic queries are allowed
+- The system essentially enters a "read-only" state for cache management operations
+- This prevents data corruption or loss during the recovery/sync process
+
+Current Implementation Status
+------------------------------
+
+**Currently Implemented:**
+
+1. **Freeze Mode Mechanism**
+   - StorageManager supports freeze mode
+   - When frozen, only LocalCPUBackend is used for retrieval operations
+   - Other storage backends are temporarily excluded from active use
+
+2. **API Endpoints**
+   - Freeze mode can be toggled via controller API
+   - Status can be queried to check if system is in freeze mode
+
+**Not Yet Implemented:**
+
+1. **Automatic Triggering**
+   - Controller does not yet automatically trigger freeze mode
+   - Manual activation via API is currently required
+
+2. **Full Sync Integration**
+   - Full synchronization process that would trigger freeze mode is not implemented
+   - Controller restart detection and automatic freeze mode activation not yet developed
+
+3. **Recovery Completion**
+   - Automatic unfreeze after sync completion not yet implemented
+   - Manual intervention currently required to exit freeze mode
+
+Usage
+-----
+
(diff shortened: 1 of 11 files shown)


maobaolong:
@sammshen @ApostaC @chunxiaozheng Would you like to take a look at this PR? Thanks!

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -431,6 +434,10 @@ def get(
 
         # Search all backends for blocking get
         for backend_name, backend in self.storage_backends.items():
+            # In hot cache freeze mode, only check local_cpu backend
+            if self.hot_cache_freeze_mode and backend_name != "LocalCPUBackend":
+                continue
+
```
ApostaC: I feel like the code here is repeated quite a few times. Can we add a new function called something like `self.get_active_storage_backeds()`

tests/v1/lookup_client/test_lmcache_lookup_client.py
```diff
@@ -13,125 +11,25 @@
...
-        local_cpu=local_cpu,
-        max_local_cpu_size=max_local_cpu_size,
-        lmcache_instance_id=instance_id,
-    )
-    config.extra_config = extra_config.copy() if extra_config else {}
-    config.extra_config["lmcache_rpc_port"] = rpc_port
-    return config
-
-
-class MockGPUConnector(GPUConnectorInterface):
```
ApostaC: Is this class un-needed anymore?

lmcache/v1/cache_engine.py
```diff
@@ -228,6 +228,36 @@ def post_init(self, **kwargs) -> None:
                 self.gpu_connector.initialize_kvcaches_ptr(**kwargs)
             self.post_inited = True
 
+    def set_hot_cache_freeze_mode(self, enabled: bool) -> None:
```
ApostaC: What about directly naming the functions as `freeze()` and `is_freezed()`?

tests/v1/lookup_client/test_lmcache_lookup_client.py
ApostaC: Oh, never mind. I see it's been moved.

ApostaC (changes requested):
Thanks @maobaolong !

A high-level comment: can we rename it as "freeze" mode and give it a clear semantics? My proposal is that the freeze mode will disable all the store operations and retrieval from the lower-tier storages.

And we can have the documentation for it, WDYT?

lmcache/v1/cache_engine.py
```diff
@@ -243,6 +243,34 @@ def post_init(self, **kwargs) -> None:
...
+        - Only local_cpu backend will be used for retrieval
+        - No admit/evict messages will be generated
+        This protects the local_cpu hot cache from changes.
+
+        Args:
+            enabled (bool): Whether to enable freeze mode
+        """
+        if self.storage_manager is not None:
+            self.storage_manager.set_freeze(enabled)
+            logger.info("freeze mode %s", "enabled" if enabled else "disabled")
```
ApostaC, line 261: After a second thought, do we need to have any thread-safe guarantee here (or in the storage manager)?

I suppose the `freeze` will be called from the api server, which will be in another thread, right?

ApostaC (commented):
Looking forward to the doc update as well + a quick question about thread-safety. Otherwise LGTM!

lmcache/v1/cache_engine.py
maobaolong: @ApostaC Within current PR changes, the freeze will only be called by `internal api server`, but I will submit another PR to set freeze mode by received full_sync request from controller.

maobaolong:
@ApostaC Make thread safe and add document, PTAL

lmcache/v1/cache_engine.py
maobaolong: yeah

lmcache/v1/storage_backend/storage_manager.py
maobaolong: done

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -430,9 +428,7 @@ def get(
         """
 
         # Search all backends for blocking get
-        for backend_name, backend in self.storage_backends.items():
-            if location and backend_name != location:
-                continue
+        for backend_name, backend in self.get_active_storage_backends(location):
```
sammshen, line 431: this is a nice refactor to move the repeated logic

docs/source/controller/freeze_mode.rst
```diff
@@ -0,0 +1,95 @@
+Freeze Mode
```
sammshen, line 1: this documentation looks great.


just a nit: 

would `freeze_non_local` be a more intuitive naming? this is just a suggestion, feel free not to take it if it is too much trouble

sammshen: sorry I read the functionality more closely and I think freeze is a good naming actually!

sammshen (approved):
LGTM!

ApostaC (approved):
LGTM!

maobaolong:
@ApostaC @sammshen Thanks for your review!

Issue #2170, opened 2025-12-04
[Bug] part of the kv get operation fails, it will cause an OOM


Leafykn:
**Describe the bug**
release: v0.3.10
When part of the kv get operation fails, it will cause an OOM

<img width="3022" height="1178" alt="Image" src="https://github.com/user-attachments/assets/2fc65b2a-648b-431e-9b3d-502700dea79c" />

**To Reproduce**
using mock connector to randomly make some `_get` requests fail.

`lmcache/v1/storage_backend/connector/mock_connector.py
`
<img width="2838" height="1086" alt="Image" src="https://github.com/user-attachments/assets/884702b8-223c-40cb-bc14-70eaabffd66c" />

yaml config
```
chunk_size: 256
local_device: null
local_cpu: False
max_local_cpu_size: 2
remote_url: "mock://5/?peeking_latency=1&read_throughput=2&write_throughput=2"
```
server
```bash
export LMCACHE_CONFIG_FILE=debug_config.yaml \
export LMCACHE_USE_EXPERIMENTAL=True \
export VLLM_USE_V1=1

python3 -m vllm.entrypoints.openai.api_server \
  --port "8000" \
  --uvicorn-log-level warning \
  --model /models/DeepSeek-R1-Distill-Qwen-1.5B \
  --served-model-name DeepSeek-R1-Distill-Qwen-1.5B \
  --trust-remote-code \
  --max-model-len "16384" \
  --disable-log-requests \
  --disable-fastapi-docs \
  --swap-space "0" \
  --no-enable-chunked-prefill \
  --no-enable-prefix-caching \
  --gpu-memory-utilization="0.95" \
  --tensor-parallel-size=2 \
  --kv-transfer-config '{"kv_connector": "LMCacheConnectorV1","kv_role": "kv_both"}' \
  >test.log 2>&1 &
```

test
```
// warm
  python3 -m vllm.entrypoints.cli.main bench serve --backend vllm \
    --tokenizer /workspace/models/DeepSeek-R1-Distill-Qwen-1.5B \
    --model DeepSeek-R1-Distill-Qwen-1.5B  \
    --base-url http://0.0.0.0:8000 \
    --endpoint /v1/completions \
    --metric_percentiles '50,90,95,99' \
    --goodput ttft:5000 tpot:250 \
    --num-prompts 16 \
    --max-concurrency 8 \
    --random-input-len 8000 \
    --random-prefix-len 0 \
    --random-output-len 200 \
[...]

PR #2172, opened 2025-12-04
[Bugfix]: part of the kv get operation fails mem not free

Leafykn:
**Describe the bug**
release: v0.3.10
When part of the kv get operation fails, it will cause an OOM

for detail [#2170](https://github.com/LMCache/LMCache/issues/2170)

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1319,6 +1319,12 @@ class LMCacheEngine:
                 reordered_chunks.append((key, memory_obj, start, end))
                 tot_kv_size += memory_obj.get_size()
                 ret_mask[start:end] = True
+                if last_failed_block_start is not None:
+                    for (key, start, end), memory_obj in zip(
+                        blocks, memory_objs, strict=False
+                    ):
+                        if memory_obj is not None and start >= last_failed_block_start:
+                            memory_obj.ref_count_down()
 
         if last_failed_block_start is not None:
             ret_mask[last_failed_block_start:] = False
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1304,7 +1304,7 @@ class LMCacheEngine:
             assert memory_objs is not None, (
                 "Failed to get memory objects from storage backend"
             )
-
+            retrieval_failed = False
             for (key, start, end), memory_obj in zip(blocks, memory_objs, strict=False):
                 if memory_obj is None:
                     logger.warning(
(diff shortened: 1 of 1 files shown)


lmcache/v1/cache_engine.py
```diff
@@ -1319,6 +1319,12 @@ def _process_tokens_internal(
                 reordered_chunks.append((key, memory_obj, start, end))
                 tot_kv_size += memory_obj.get_size()
                 ret_mask[start:end] = True
+                if last_failed_block_start is not None:
+                    for (key, start, end), memory_obj in zip(
```
luotang-me: Not handling partial failure scenarios？

Leafykn:
@maobaolong  @ApostaC @sammshen  Could you please check if this MR is correct?

lmcache/v1/cache_engine.py
```diff
@@ -1319,6 +1319,12 @@ def _process_tokens_internal(
                 reordered_chunks.append((key, memory_obj, start, end))
                 tot_kv_size += memory_obj.get_size()
                 ret_mask[start:end] = True
+                if last_failed_block_start is not None:
```
ApostaC: There is a `break` in line 1318. IIUC, the new logic may not be executed when there are load failures? 

Why don't we process this at the end of this function (i.e., working on `reordered_chunks`)?

ApostaC (commented):
Thanks for the contribution! I'm not sure about the correctness here. See the comment below.

ApostaC:
@sammshen @maobaolong Can you help review this as well?

lmcache/v1/cache_engine.py
Leafykn: fixed

lmcache/v1/cache_engine.py
```diff
@@ -1315,10 +1315,14 @@ def _process_tokens_internal(
...
                     ):
                         last_failed_block_start = start
-                    break
-                reordered_chunks.append((key, memory_obj, start, end))
-                tot_kv_size += memory_obj.get_size()
-                ret_mask[start:end] = True
+                        retrieval_failed = True
+                if memory_obj:
+                    if retrieval_failed:
+                        memory_obj.ref_count_up()
```
sammshen, line 1321: should we break early?

Leafykn: There is no need to use break to exit the loop; the remaining memory is freed via the flag value.

Leafykn: @sammshen

ApostaC, line 1321: Can you explain why we need to do a `ref_count_up` here? Will it cause a memory leak on the retrieved objects?

ApostaC (commented):
Just a quick question.

lmcache/v1/cache_engine.py
Leafykn: @ApostaC `_process_tokens_internal` only returns useful memory_obj instances and frees memory after batched_to_gpu is executed. However, it appears that obsolete memory_obj instances are not being released — you may check the logic around line 576.
image

sammshen: how is it possible that memory_obj is not None and also retrieval_failed?

lmcache/v1/cache_engine.py
```diff
@@ -1315,10 +1315,14 @@ def _process_tokens_internal(
...
+                    elif memory_obj:
```
sammshen, line 1322: this elif memory_obj is inside of an if memory_obj block

sammshen:
even if part of the KV get operation fails, we are already RC down at the end of `cache_engine.retrieve()`

sorry I'm not sure I see the point of this PR

lmcache/v1/cache_engine.py
sammshen: please see the comments I left, these conditionals feel unnatural

Leafykn:
> sorry I'm not sure I see the point of this PR

reordered_chunks will not contain all of the memery_obj ; Only part of the memory is reclaimed at then endof cache_engine.retrieve() method.
@sammshen

luotang-me:
hao de ...哦，DEAR，已经收到

PR #2174, opened 2025-12-04
[Bugfix] Fix batched message sender race condition

maobaolong:
Fix flaky test.

### DIFF
--- a/lmcache/v1/storage_backend/batched_message_sender.py
+++ b/lmcache/v1/storage_backend/batched_message_sender.py
@@ -70,13 +70,6 @@ class BatchedMessageSender:
         self.sequence_number = 0
         self.sequence_lock = threading.Lock()
 
-        # Lock to ensure _drain_and_send is not called concurrently
-        self.drain_lock = threading.Lock()
-
-        # Flag to indicate flush is in progress
-        self.flushing = False
-        self.flush_lock = threading.Lock()
-
         # Condition variable for coordinating producer and consumer
         self.cv = threading.Condition()
         self.running = False
@@ -111,11 +104,6 @@ class BatchedMessageSender:
                 # This prevents race conditions but we'll release lock
                 # before blocking operations
                 if self.message_queue.empty():
-                    continue
-
-            # Skip processing if flush is in progress to avoid interference
-            with self.flush_lock:
-                if self.flushing:
                     continue
 
             # Drain the queue without holding the lock to avoid blocking producers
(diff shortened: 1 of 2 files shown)


KuntaiDu (approved):
The code is not introducing new functionality, just wrapping around existing code using locks so LGTM.

ApostaC:
Quick question: what's the root cause of the race condition in the previous implementation?

lmcache/v1/storage_backend/batched_message_sender.py
```diff
@@ -150,65 +162,81 @@ def _drain_and_send(self):
...
-        with self.cv:
-            self.cv.notify()
+        # Set flushing flag to prevent background thread from processing
+        with self.flush_lock:
+            self.flushing = True
+
+        try:
+            # Notify consumer thread (it will skip due to flushing flag)
+            with self.cv:
+                self.cv.notify()
```
ApostaC: I'm a bit confused here. Why we need to call `cv.notify()`? Is it supposed to be called in `add_kv_op` to flag that the number of messages is larger than the pre-defined threshold?

lmcache/v1/storage_backend/batched_message_sender.py
```diff
@@ -150,65 +162,81 @@ def _drain_and_send(self):
...
-                threading.Event().wait(0.001)
-            else:
-                break
-
-        # Final drain to ensure we got everything
-        self._drain_and_send()
+        finally:
+            # Clear flushing flag to allow background thread to resume
+            with self.flush_lock:
+                self.flushing = False
```
ApostaC: Sorry, I didn't pay too much attention to this part of the code last time.

At a high level, what should be the semantics of `flush`? 

My previous understanding is, when calling `flush`, it makes sure that all the events generated "before" this call should be flushed out. In this case, we can do it with a simple `cv.notify` to trigger the consumer thread, and then `_drain_and_send` will be called in the consumer thread.

My idea is that if we can have only one thread calling `_drain_and_send` (which is the consumer thread), it will be easier to avoid race conditions.

ApostaC (commented):
Have got some thoughts.

lmcache/v1/storage_backend/batched_message_sender.py
ApostaC: Okay, somehow this makes sense to me now

lmcache/v1/storage_backend/batched_message_sender.py
maobaolong: That's great, x producers -> one single consumer, we never need a lock to protect _drain_and_send

maobaolong:
@ApostaC Would you like to take another look, the suggestion has been addressed. And the failed test is not related to this PR.

lmcache/v1/storage_backend/batched_message_sender.py
```diff
@@ -159,59 +159,52 @@ def _drain_and_send(self):
...
-            self._drain_and_send()
-            # Small sleep to allow any in-flight messages to be queued
-            if not self.message_queue.empty():
-                threading.Event().wait(0.001)
-            else:
-                break
-
-        # Final drain to ensure we got everything
-        self._drain_and_send()
+        self.message_queue.join()
```
ApostaC, line 203: 👍

ApostaC (approved):
LGTM! Thanks!

PR #2176, opened 2025-12-04
fix: use served_model_name in LMCache metrics

mustafayildirim:
## Summary
This PR adds `served_model_name` as an **optional field** in `LMCacheEngineMetadata` and exposes it as an additional Prometheus metric label when provided by vLLM.

## Problem
Currently, LMCache Prometheus metrics show the full filesystem path in the `model_name` label:
```
lmcache:retrieve_hit_rate{model_name="/huggingface/cache/hub/Qwen/Qwen3-235B-A22B-Instruct-2507-FP8",...}
```

When users configure vLLM with `--served-model-name`, they expect to see this user-friendly name in metrics instead of the internal path.

## Solution
**Key Design Decision**: Keep `model_name` using the physical model path for cache key consistency, and add `served_model_name` as a separate optional field.

This approach ensures:
- **KV cache sharing works correctly**: Multiple vLLM instances serving the same physical model (e.g., `Qwen/Qwen3-0.6B`) with different `served_model_name` values (e.g., `qwen3-model-a`, `qwen3-model-b`) can still share the same KV cache
- **Metrics are user-friendly**: When `served_model_name` is provided, it appears as an additional label in Prometheus metrics

### 1. Add `served_model_name` field to `LMCacheEngineMetadata`
- Added as `Optional[str] = None` in [lmcache/config.py](lmcache/config.py#L41)
- Does not affect cache key generation

### 2. Update vLLM integration to pass `served_model_name`
- [lmcache/integration/vllm/vllm_v1_adapter.py](lmcache/integration/vllm/vllm_v1_adapter.py#L539): Pass `served_model_name` as kwarg
- [lmcache/integration/vllm/utils.py](lmcache/integration/vllm/utils.py#L181): Pass `served_model_name` as kwarg
- `model_name` field continues to use `model_config.model` (physical path)

### 3. Add `served_model_name` to Prometheus metric labels
- [lmcache/observability.py](lmcache/observability.py#L1315-1316): Add `served_model_name` label when available
- Metrics will show both labels:
  - `model_name`: physical model path (for cache key consistency)
  - `served_model_name`: user-configured name (for readability)

### 4. Fix test issues
- Added `served_model_name` to test mock
- Fixed `lookup_cache()` assertion (returns `-1` not `None` after clear, per API contract)

## Example Metrics Output

**Before:**
```
lmcache:retrieve_hit_rate{model_name="/huggingface/cache/hub/Qwen/...",role="worker",worker_id="0"}
```

**After:**
```
lmcache:retrieve_hit_rate{model_name="/huggingface/cache/hub/Qwen/...",served_model_name="qwen3-model-a",role="worker",worker_id="0"}
```

## Testing
- Unit tests pass with mock `served_model_name`
- vLLM integration tests pass
- Backward compatible: works with and without `served_model_name`

## Related
- vllm-project/vllm#30007


lmcache/integration/vllm/utils.py
```diff
@@ -170,7 +170,7 @@ def create_lmcache_metadata(
 
     # Create metadata
     metadata = LMCacheEngineMetadata(
-        model_cfg.model,
+        model_cfg.served_model_name,
```
ApostaC: Hi, thanks for the contribution!

I don't think we should use `served_model_name` here, because whether the KV cache can be reused is related to the "physical model" rather than the model name.

If you really want to have `served_model_name` in the Prometheus metrics, an alternative way is to introduce some extra optional field in the LMCacheEngineMetadata.

ApostaC (changes requested):
Thanks for your contribution!

lmcache/integration/vllm/utils.py
mustafayildirim: Hi,

If you provide `served_model_name` in vllm, that takes precedence. Otherwise fall back to `model_cfg.model` by default in `vllm` 
That is the default behavior https://github.com/vllm-project/vllm/blob/1f0d1845909877b3492d725cb63b386d80c39b47/vllm/entrypoints/openai/api_server.py#L1067

But again, if you want me to introduce a new field, I can do that.

mustafayildirim: https://github.com/vllm-project/vllm/blob/1f0d1845909877b3492d725cb63b386d80c39b47/vllm/config/model.py#L438

https://github.com/vllm-project/vllm/blob/1f0d1845909877b3492d725cb63b386d80c39b47/vllm/config/model.py#L1870

ApostaC: What I'm saying is: one can launch vllm instances the same model (let's say "Qwen/Qwen3-0.6B") but different `served_model_name` (e.g., "qwen3-model-a", "qwen3-model-b" for different instances).
In this case, LMCache should still be able to reuse the KV cache, which means they should have the same key.

Therefore, to avoid breaking existing logics, can you introduce a new field?

ziliangpeng: so the reason this matter in LmCache is that the model's name is part of the key for the cache?

mustafayildirim: updated the PR

tests/v1/lookup_client/test_lmcache_lookup_client.py
```diff
@@ -212,7 +213,8 @@ def test_basic_lookup_communication(self, lmcache_engine):
 
                 # Test clear lookup status
                 client.clear_lookup_status(lookup_id)
-                assert client.lookup_cache(lookup_id) is None
+                # After clearing, lookup_cache returns -1 (not found)
+                assert client.lookup_cache(lookup_id) == -1
```
maobaolong, line 216: This has been fixed in another PR and merged, would you like to rebase your branch? @mustafayildirim

maobaolong (approved):
Left a comment, otherwise LGTM.

tests/v1/lookup_client/test_lmcache_lookup_client.py
mustafayildirim: rebased the branch

ApostaC (approved):
LGTM!

mustafayildirim:
[buildkite/comprehensive-tests](https://buildkite.com/lmcache/comprehensive-tests/builds/716) tests are failing but I cant retry the job

```

→ Running long_doc_qa with customed workload config:
--
{
"max-inflight-requests": 2,
"repeat-count": 1,
"completion": true,
"num-documents": 20,
"model": "meta-llama/Llama-3.2-1B-Instruct"
}
Expected warmup latency: 1.3680354595184325
Actual warmup latency: 1.5318623542785645
Latency requirement not met

```

maobaolong:
> [buildkite/comprehensive-tests](https://buildkite.com/lmcache/comprehensive-tests/builds/716) tests are failing but I cant retry the job
> 
> ```
> 
> → Running long_doc_qa with customed workload config:
> --
> {
> "max-inflight-requests": 2,
> "repeat-count": 1,
> "completion": true,
> "num-documents": 20,
> "model": "meta-llama/Llama-3.2-1B-Instruct"
> }
> Expected warmup latency: 1.3680354595184325
> Actual warmup latency: 1.5318623542785645
> Latency requirement not met
> ```

I believe this is not related to this PR. Retried for you just now.

PR #2179, opened 2025-12-05
Fix: adjust test_lmcache_lookup_client.py to align the change of return val of lookup_cache

maobaolong:
Fix test failed after merge https://github.com/LMCache/LMCache/pull/2146


maobaolong:
@sammshen Would you like to take a look at this PR? Thanks.

sammshen:
will fix by https://github.com/LMCache/LMCache/pull/2175 which also fixes async pq shutdown which sometimes crashes UT also!

maobaolong:
@sammshen Close this as it is duplicated.

PR #2182, opened 2025-12-05
[P2PBackend] add lock to protect peer_info_mapping

chunxiaozheng:
add lock to protect `self.peer_info_mapping`


lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -480,36 +483,45 @@ async def _ensure_peer_connection(
...
+                peer_lookup_url,
+                "tcp",
+                zmq.REQ,
+                "connect",
+                self.socket_recv_timeout_ms,
+                self.socket_send_timeout_ms,
             )
+            if peer_info is not None:
+                peer_info.update_peer_lookup_url(peer_lookup_url)
+                peer_info.update_lookup_socket(lookup_socket)
```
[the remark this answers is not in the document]
chunxiaozheng: `batched_get_non_blocking` and `async_batched_submit_put_task` will try get `lookup_lock` first and then use `lookup_socket` to send or recv data, which will protest the `lookup_socket`. If modified by your comments, which will try to get 'lookup_lock' again, and result in dead lock.

lmcache/v1/storage_backend/p2p_backend.py
```diff
@@ -480,36 +483,45 @@ async def _ensure_peer_connection(
...
-            self.target_peer_info_mapping[target_peer_init_url] = PeerInfo(
+        async with self.update_peer_lock:
+            # double check
+            if (
+                not force_update
+                and target_peer_init_url in self.target_peer_info_mapping
+            ):
+                return
+
+            init_side_msg = P2PInitSideMsg()
```
maobaolong, line 495: The follow codes are not changed, just only surrounded by the lock, is that right?

If so, the change is extremely tiny, but github cannot show the diff clearly.

maobaolong: Clipboard_Screenshot_1765111738

This is much clear.

maobaolong (approved):
@chunxiaozheng Thanks for this fix. LGTM.

lmcache/v1/storage_backend/p2p_backend.py
chunxiaozheng: @maobaolong Yeah, the changes is only surrounded by the lock.

chunxiaozheng:
@KuntaiDu Could you help take another look?

KuntaiDu (approved):
LGTM!

PR #2184, opened 2025-12-05
[controller][optimize] catch exception when process WorkerReqMsg

chunxiaozheng:
when process `WorkerReqMsg`, we must add `try-except` to process exception, otherwise, if error happens, the server will not return any msg and the client will block until timeout occurs, and will re-create a new socket, but this is unnecessary.

**In this PR, I only added an exception handling in `batched_p2p_lookup`, and for the rest of the code, I only organized the classification of msg to make code more clear, the code has not changed.**


lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -223,55 +271,33 @@ async def batched_p2p_lookup(
...
+                num_hit_chunks += 1
+
+            return BatchedP2PLookupRetMsg(
+                layout_info=[
+                    (instance_id, location, num_hit_chunks, peer_init_url),
+                ]
+            )
+        except Exception as e:
+            logger.error("batched p2p lookup error", e)
+            return BatchedP2PLookupRetMsg(layout_info=[("", "", 0, "")])
```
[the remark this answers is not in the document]
chunxiaozheng: done

chunxiaozheng:

maobaolong (approved):
@chunxiaozheng Thanks for the introduction of this PR, in fact, the only meaningful change of this PR is adding a `try-except`, it is simple and clear, so this LGTM.

chunxiaozheng:
@sammshen could you help take another look? the only change is add `try-except` in `batched_p2p_lookup`, other changes is to organize the classification of msg and place them together.

lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -55,6 +55,10 @@ def __init__(self) -> None:
 
         self._setup_metrics()
 
+    ############################################################
+    # private and public methods
```
sammshen: what is the point of this comment? 

every method is private or public right? 

I think the implicit best-practice is already that _ prefix is already for internal use and should not be exposed

sammshen (approved):
LGTM, thanks!

lmcache/v1/cache_controller/controllers/kv_controller.py
chunxiaozheng: @sammshen Thanks for your review, I have deleted the comment.

PR #2131, opened 2025-12-01
Refactor: Replace Complex Mappings with Hierarchical Tree Structure for Instance-Worker Management

maobaolong:
**What this PR does / why we need it**:

This PR refactors the complex mapping relationships in the LMCache registration system by introducing a clear hierarchical tree structure. The previous implementation used multiple scattered dictionaries to manage instance-worker mappings, which was difficult to maintain and error-prone.

**Changes**
- Introduced RegistryTree class: Centralized management of all instance-worker relationships
- Hierarchical structure: RegistryTree → InstanceNode → WorkerNode
- Eliminated scattered mappings: Replaced 5+ separate mapping with a single unified tree structure
- Improved data encapsulation: Each node class manages its own data and operations

**Key Benefits**
- Simplified maintenance: Single source of truth for all instance-worker relationships
- Type safety: Full type hints with dataclasses
- Automatic cleanup: Empty instances are automatically removed

```mermaid
graph TB
    %% Old Structure - Complex Flat Mappings
    Old[Old: Complex Flat Mappings]
    Old --> O1[instance_id_mapping]
    Old --> O2[worker_id_mapping] 
    Old --> O3[worker_info_mapping]
    Old --> O4[kv_pool]
    Old --> O5[seq_tracker]
    
    %% New Structure - Hierarchical Tree
    New[New: Hierarchical Tree Structure]
    New --> RegistryTree[RegistryTree]
    RegistryTree --> Instance1[InstanceNode]
    RegistryTree --> Instance2[InstanceNode]
    RegistryTree --> InstanceN[InstanceNode]
    
    Instance1 --> Worker1[WorkerNode]
    Instance1 --> Worker2[WorkerNode]
    Instance2 --> Worker3[WorkerNode]
    InstanceN --> WorkerN[WorkerNode]
    
    Worker1 --> WI1[WorkerInfo]
    Worker2 --> WI2[WorkerInfo]
    Worker3 --> WI3[WorkerInfo]
    WorkerN --> WIN[WorkerInfo]
    
    WI1 --> KP[KV Pool]
    WI2 --> KP
    WI3 --> KP
    WIN --> KP

    %% Comparison
    Old -.->|replaced by| New


```

PR #2187, opened 2025-12-05
[Controller] Refactor: Combine kv_pool and worker_node

maobaolong:
This refactor is a follow-up of #2131

We can remove kv_pool and use the workerNode to keep the keys belong to.

### DIFF
--- a/lmcache/v1/cache_controller/utils.py
+++ b/lmcache/v1/cache_controller/utils.py
@@ -32,6 +32,40 @@ class WorkerNode:
     registration_time: float
     last_heartbeat_time: float
     seq_tracker: dict[str, int] = field(default_factory=dict)  # location -> seq_num
+    kv_store: dict[str, set[int]] = field(
+        default_factory=dict
+    )  # location -> set[chunk_hash]
+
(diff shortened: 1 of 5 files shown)


lmcache/v1/cache_controller/utils.py
```diff
@@ -223,3 +257,70 @@ def get_seq_num(
...
+        if worker_node is None:
+            return False
+        return worker_node.evict_kv(location, key)
+
+    def find_kv(
+        self,
+        key: int,
+        exclude_instance_id: Optional[str] = None,
+        exclude_location: Optional[str] = None,
+    ) -> Optional[tuple[str, int, str]]:
```
KuntaiDu: I would prefer wrapping `tuple[str, int, str]` into a class. But this can be done in future PR.

KuntaiDu (approved):
LGTM!

lmcache/v1/cache_controller/utils.py
maobaolong: Done, change to `class KVChunkInfo(NamedTuple)`

chunxiaozheng (approved):
lgtm

Issue #2181, opened 2025-12-05
[BUG] in pd disaggregation scenario, 1p1d, different node, mooncake transfer, the output is not correct


yangshanjun:
<img width="1067" height="855" alt="Image" src="https://github.com/user-attachments/assets/8f398abf-9606-4a4f-9bd2-1413aaf00a0d" />

in pd disa scen, memory_obj.tensor is cuda tensor, so it will not execute self.store_stream.synchronize, then may transfer invalid kvcache value to the decode side,  then i remove if condition, like this， it works, the output is correct

<img width="978" height="759" alt="Image" src="https://github.com/user-attachments/assets/c81fb7ed-d2be-414f-b9c9-e938b527e403" />

DongDongJu:
I will take a look

liwei-rk:
even if on same node, there is a need for synchronize. as the kv cache is stored on different process. the storage_manager.put function will not wait the memory obj to be filled. It may be common issue if the memory_obj will stored on storage backend other than LocalCPUBackend

liwei-rk:
Hi Shanjun, can you show your lmcache_config?  maybe you have  configured the enable_pd, if you remove it, the issue will be not trigged. For mooncake usage, no need to configure it.

yangshanjun:
> Hi Shanjun, can you show your lmcache_config? maybe you have configured the enable_pd, if you remove it, the issue will be not trigged. For mooncake usage, no need to configure it.


yes, I have configured the enable_pd=true, why in moocake usage, no need to configue enable_pd？

 cat lmcache-prefiller-config.yaml
local_cpu: True
max_local_cpu_size: 64
max_local_disk_size: 0

enable_pd: True
transfer_channel: "nixl"
pd_role: "sender"
pd_proxy_host: "localhost"
pd_proxy_port: 7600
pd_buffer_size: 4076863488 # 1GB
pd_buffer_device: "cuda"
nixl_backends: [Mooncake]

liwei-rk:
> > Hi Shanjun, can you show your lmcache_config? maybe you have configured the enable_pd, if you remove it, the issue will be not trigged. For mooncake usage, no need to configure it.

https://github.com/LMCache/LMCache/pull/2193
see mooncake settings below.
https://docs.lmcache.ai/kv_cache/storage_backends/mooncake.html
https://kvcache-ai.github.io/Mooncake/getting_started/examples/vllm-integration/vllmv1-lmcache-integration.html

[...]

PR #2193, opened 2025-12-06
[FIXBUG] in pd disaggregation scenario, 1p1d, different node, mooncake transfer, the output is not correct

KeeProMise:
see: https://github.com/LMCache/LMCache/issues/2181

**If applicable**:


KeeProMise:
@DongDongJu 
Hi, if you have time, please help to take a look at this PR, thanks!

DongDongJu (approved):
LGTM

KeeProMise:
> LGTM

@DongDongJu 
Hi, if you have no more comments or questions, please help merge this PR, thank!

liwei-rk (changes requested):
I'm not sure the code here is enough. In which cace, can the cache_engine skip the torch.xpu.synchronize()?
If there is a non LocalCpuBackend exist, not only the PD scenaria,  for example,  the mooncake backend (for non PD scenario, e.g. shared storage mode)  or shared DFS LocalDiskBacked,  it need a syncrhonize, otherwise, the StorageManager.put will use partitial memory_objs.

DongDongJu:
> I'm not sure the code here is enough. In which cace, can the cache_engine skip the torch.xpu.synchronize()? If there is a non LocalCpuBackend exist, not only the PD scenaria, for example, the mooncake backend (for non PD scenario, e.g. shared storage mode) or shared DFS LocalDiskBacked, it need a syncrhonize, otherwise, the StorageManager.put will use partitial memory_objs.

IMO, there are tons of configuration setup we can make it with inference engines and LMCache configs now. 
And, I believe good way to resolve these type of question we have in here is that we can make the code changes together for when you find the case.
This one is fair enough small changes to me.
But ping me when you make the changes for case you describe here.
I am happy to discuss.

liwei-rk:
Hi @DongDongJu 
There is a optimization introduced in [PR#522](https://github.com/LMCache/LMCache/pull/522), before that, every put will be blocked until the syncronize is done.

this issue happens as the belowing call stack 
```
LMCacheEngine.store():
    ...
    ...
    // Here is, if the memory_objs is not syncronized before batched_put, and there is a non LocalCPUBackend,
    // The memory_obj may be filled partly.
    self.storage_manager.batched_put(keys, memory_objs, transfer_spec=transfer_spec)
```
The LMCacheEngine.store() is called in
```
LMCacheConnectorV1Impl.wait_for_save()
    """Blocking until the KV cache is saved to the connector buffer."""
    ...
    for request in connector_metadata.requests:
         ...
        self.lmcache_engine.store(...)
        ...
```
since the wait_for_save is called in vllm, when a batch of requests have been done by gpu.
and it is be desinged as a blocking function.

So, If we can split the store function into more atomic operations like batched_from_gpu and batched_put (The names can be changed)  and adds a barrier sync between them. Then it may solve this issue, and keeps the  optimization as well.

```
LMCacheConnectorV1Impl.wait_for_save()
    """Blocking until the KV cache is saved to the connector buffer."""
    ...
    for request in connector_metadata.requests:
         ...
        self.lmcache_engine.batched_from_gpu(...)
        ...
    torch.cuda.syncronize()
    self.lmcache_engine.batched_put(...)
```
Add @YaoJiayi the orignal PR reviewer into this thread as well.

DongDongJu:
Thanks @liwei-rk for kind description.
I fully understand what you are meaning. 

but your approach is little bit off from this PR's contribution.
Could you open other PR for your idea?
I fully support your idea if we dont have that much performance degradation in high concurrency case.

liwei-rk:
add a new PR for to track futher optimization. https://github.com/LMCache/LMCache/issues/2286

KeeProMise:
Hi @DongDongJu @liwei-rk Any additional thoughts on this PR?

liwei-rk:
> Hi @DongDongJu @liwei-rk Any additional thoughts on this PR?

Hi KeeProMise, please check the original PR, https://docs.lmcache.ai/kv_cache/storage_backends/mooncake.html

For mooncake scenaria, there is no need to configure the enable_pd in your lmcache_config. If you remove it, the issue will be resloved. 

There are two branches in the from_batch function.

The enable_pd is not designed for mooncake.

liwei-rk:
Hi, @KeeProMise, It seems you configured enable_pd option in the previous failed tests. 

For, mooncake settings, there is no need to configure the enable_pd. If you remove it, the issue #2181  will not be trigged.
1. the enable_pd option is not designed for mooncake. It's designed for nixl. 
https://docs.lmcache.ai/disaggregated_prefill/nixl/1p1d.html
3.  in the from_gpu function, if you remove enable_pd, it will goes to the second branch, where it has a memory_obj.tensor.copy_ funtion call, and it's a blocking function, which already has the same semantic of syncronize().

Please have a try by removing the enable_pd  from your config settings.

see mooncake settings below.
https://docs.lmcache.ai/kv_cache/storage_backends/mooncake.html
https://kvcache-ai.github.io/Mooncake/getting_started/examples/vllm-integration/vllmv1-lmcache-integration.html

Issue #2206, opened 2025-12-08
[BUG] Incorrect output when LMCache loads partial chunks overlapping with vLLM prefix cache


DongDongJu:
**Describe the bug**
When LMCache is enabled with vLLM, sometimes requests with warm cache produce incorrect outputs even with deterministic sampling (`temperature=0`, `top_k=1`).

### Root Cause            
                                                            
The issue occurs when `vllm_cached_tokens` is not aligned to `chunk_size` boundaries. For example:
- `vllm_cached_tokens = 14816` (not chunk-aligned, chunk_size=256)                    
- `lmcache_cached_tokens = 14820`          
- `masked_token_count = 14816 // 256 * 256 = 14592` (rounded DOWN)

**The Problem:**
1. LMCache masks only up to `14592` tokens (chunk-aligned boundary)
2. LMCache attempts to load tokens `14592~14820` (228 tokens)                                                            
3. However, vLLM has already cached tokens `14592~14816` (224 tokens)
4. This creates a **partial chunk** (`14592~14848`) that mixes:
   - **vLLM's live cache** (`14592~14816`) - computed in current execution context
   - **LMCache's stored cache** (`14816~14820`) - loaded from previous execution
                                                                                                                         
**Why This Causes Wrong Outputs:**                                                                                       
Mixing data sources within a single chunk creates inconsistent Attention Context. The Attention mechanism relies on all previous KV cache values being consistent. When vLLM's live cache (computed with current context) is mixed with LMCache's stored cache (computed with potentially different context), the Attention computation produces incorrect results, leading to wrong token generation.

**To Reproduce**
1. Run `vllm serve Qwen/Qwen3-0.6B --port 8000 --kv-offloading-backend lmcache --kv-offloading-size 20 --disable-hybrid-kv-cache-manager --override-generation-config '{"temperature": 0, "top_k": 1}'`
2. Follow here example (https://docs.lmcache.ai/kv_cache/storage_backends/local_storage.html)
3. run `python query_two_times.py`
4.  got result like
```
Number of tokens in prompt: 14785                                                                                        
Querying vLLM server with cold LMCache Disk Offload                                                                      
<think>                                                     
Okay, let's see. The user wants a summary of bash in two sentences. First, I need to recall the main features of bash. It
's a shell interpreter, supports various options, handles commands, and has some advanced features like the --debugger an
d --version. Also, it's a POSIX-compliant shell.            
                                                                                                                         
[...]

Issue #2136, opened 2025-12-02
[BUG] LMCache unable to load stored KVCache back on another vLLM process because of hash value inconsistency


crystalxyz:
**Describe the bug**
The hash value for tokens is not constant across processes even though `PYTHONHASHSEED=0` is set.

**To Reproduce**
1. Start lmcache server
2. Start vLLM server, set `PYTHONHASHSEED=0`
3. Send request to vLLM
4. Stop vLLM and start another one
5. Send the same request to vLLM
6. Observed from log that the code is not reusing the saved entries in LMCache, but instead recomputing KV

**Expected behavior**
Hash key should match and we should load from LMCache instead of recomputing.

**Desktop (please complete the following information):**
 - OS: Linux

**Additional context**
Did some basic investigation and it seems that the `None` value in `extra_keys` is causing indeterministic hash values although hash seed is already set. The problem is very likely about the `TokenDatabase::_hash_tokens()` methods. Replacing None value with an empty tuple solves the problem.

DongDongJu:
I will take a look

DongDongJu:
I will make the PR for this issue after merging https://github.com/LMCache/LMCache/pull/2207.

DongDongJu:
@sammshen

sammshen:
are you starting the lmcache server with `PYTHONHASHSEED=0` as well?

JianDan0212:
I encountered this issue before and submitted a corresponding PR, but no one has responded. I’ve already fixed the problem, so you can check out the reason/details.
https://github.com/LMCache/LMCache/issues/1842

PR #2207, opened 2025-12-08
[hotfix/UT] Prevent cache corruption when vLLM prefix cache overlaps with LMCache

DongDongJu:
Updates:
After discussion with @ApostaC, we currently decided to set the default value of `save_unfull_chunk` as false.
I changed the code, doc and modify the ut cover for both case(True and False).

# history 
During investigation https://github.com/LMCache/LMCache/issues/2136 with local disk backend test,
but I found out that when vLLM prefix cache and LMCache align to the same chunk boundary, LMCache loads tokens already cached by vLLM, causing cache corruption.

**Example:** `vllm_cached_tokens=14816`, `lmcache_cached_tokens=14820` → both round to `14592` → previous logic overwrites vLLM cache

Fixes cache corruption when vLLM prefix cache and LMCache align to the same chunk boundary. 
Previously, LMCache would load tokens already cached by vLLM, overwriting vLLM's cache and causing incorrect outputs.

This PR skips LMCache retrieval when all chunk-aligned tokens are cached by vLLM, letting vLLM recompute partial chunks to avoid data inconsistency.

- Skip logic: `if retrieve_end_aligned <= masked_token_count: continue`
- When skipped, vLLM recomputes tokens from `vllm_cached_tokens` to `lmcache_cached_tokens`, which may cause slightly different output due to floating-point non-determinism, but prevents cache corruption
- Tested with same prompt four times (cold → warm), `temperature=0`, `top_k=1`

Compared to https://github.com/LMCache/LMCache/issues/2206's result.
we can get below result with 4 times same query.
image
image

**If applicable**:


DongDongJu:
I ran w/o layerwise cuz it has issue now.

DongDongJu:
After discussing with @ApostaC offline, currently we set save_unfull_chunk to false as a default variable.
result in same request return same result
```
<think>
Okay, let's see. The user wants a summary of bash in two sentences. First, I need to recall the main features of bash. It's a shell interpreter, supports various options, handles commands, and has some advanced features like the --debugger and --version. Also, it's a POSIX-compliant shell. 
Wait, the user provided a detailed description, but they want it in two sentences. So I should condense that into two concise points. Maybe start with its role as a shell interpreter and mention key features like options, command execution, and compatibility. That should cover it.
</think>
Bash is a powerful, POSIX-compliant shell interpreter that executes commands from standard input or files, supports various options, and handles advanced features like interactive mode and restricted shells.
Cold TTFT: 33.647 seconds
Querying vLLM server with warm LMCache Disk Offload
<think>
Okay, let's see. The user wants a summary of bash in two sentences. First, I need to recall the main features of bash. It's a shell interpreter, supports various options, handles commands, and has some advanced features like the --debugger and --version. Also, it's a POSIX-compliant shell. 
Wait, the user provided a detailed description, but they want it in two sentences. So I should condense that into two concise points. Maybe start with its role as a shell interpreter and mention key features like options, command execution, and compatibility. That should cover it.
</think>

Bash is a powerful, POSIX-compliant shell interpreter that executes commands from standard input or files, supports various options, and handles advanced features like interactive mode and restricted shells.
Warm TTFT: 0.147 seconds

TTFT Improvement: 33.500 seconds     (229.3x faster)
```

DongDongJu:
@ApostaC @sammshen Could you take a look?

tests/benchmarks/test_benchmark.py
```diff
@@ -152,6 +162,14 @@ def test_store_1GB(benchmark, backend, create_config, autorelease_v1):
         )
     )
 
+    # Calculate expected lookup count based on save_unfull_chunk setting
+    def get_expected_count(token_len):
+        if save_unfull_chunk:
+            return token_len
+        else:
+            # Round down to chunk boundary
+            return (token_len // chunk_size) * chunk_size
```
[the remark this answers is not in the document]
DongDongJu: You are right. I resolved it by https://github.com/LMCache/LMCache/pull/2207/commits/bcb86070573b4f72b3f526b5fd3f71028d80daa9

DongDongJu:
@sammshen @ApostaC ready to review it. I just made `save_unfull_chunk` to false and modify the docs and ut tests based on that changes. Tried to keep the original ut behavior.

ApostaC (approved):
LGTM! Thanks!

@sammshen @maobaolong @KuntaiDu Can one of you take a quick review at this PR? We are going to make `save_unfull_chunk` default to false for a while until we've got a clear sense about the potential correctness issue. Thanks!

YaoJiayi (commented):
@DongDongJu thanks! lgtm in general. Just one question: will pd backend be affected by this?

DongDongJu:
> @DongDongJu thanks! lgtm in general. Just one question: will pd backend be affected by this?

You are right. I added https://github.com/LMCache/LMCache/pull/2207/commits/25310af6351c70bb34fa1d88c4ae82e7690f550a to force sending full cache if pd is enabled.

YaoJiayi (approved):
Thanks for this quick fix!

DongDongJu:
I think there are some bugs related chunked prefill with mla mode in vllm https://github.com/vllm-project/vllm/issues/26042.
I will take a look after make next PR.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -501,6 +501,22 @@ def _init_lmcache_engine(
     ):
         raise ValueError("MLA only works with naive serde mode..")
 
+    # MLA requires save_unfull_chunk=True for correct KV cache storage and retrieval.
```
yoo-kumaneko, line 504: @DongDongJu I might be missing something. Could you explain why save_unfull_chunk affects the storage of the MLA cache? Also, does MLA refers to Multi-head Latent Attention or Multi Level Attention?

DongDongJu: Hello @yoo-kumaneko, Happy new year.
Sry mb, it was Deepseek's Multi-head Latent Attention.
I will change it soon.
I simply thought that it was buggy(https://github.com/vllm-project/vllm/issues/26042) before.
So maybe we can remove the entire block for now.
what do you think?

yoo-kumaneko: > Hello @yoo-kumaneko, Happy new year. Sry mb, it was Deepseek's Multi-head Latent Attention. I will change it soon. I simply thought that it was buggy([vllm-project/vllm#26042](https://github.com/vllm-project/vllm/issues/26042)) before. So maybe we can remove the entire block for now. what do you think?

Happy New Year. 
Yes, we can remove the entire block. I don't think it has any effect on the cache storage for MLA now.

DongDongJu: Let me make the hotfix for this.

DongDongJu: made PR for changes https://github.com/LMCache/LMCache/pull/2349

PR #2217, opened 2025-12-10
Support eviction/caching policies for GDS backend

MingqinHan:

This PR adds full caching and eviction policy support to the GDS backend, making its behavior consistent with the existing LocalDisk backend.

Concretely, this PR:

- Integrates the existing `cache_policy` framework into the GDS backend:
  - `update_on_put(key)`
  - `update_on_hit(key, cache_dict)`
  - `update_on_force_evict(key)`
  - `get_evict_candidates(cache_dict, num_candidates)`
- Introduces a configurable maximum GDS cache size, derived from:
  - `config.max_gds_size` (in GB), or
  - the `LMCACHE_MAX_GDS_SIZE` environment variable.
- Tracks current GDS disk usage via `self.current_cache_size` and enforces
  `max_cache_size` in `_maybe_evict_for(required_size)`.
- Implements policy-driven eviction that:
  - selects candidates via `cache_policy.get_evict_candidates`,
  - removes both `.kvcache.safetensors` and `.metadata` files,
  - updates `current_cache_size`,
  - notifies the policy via `update_on_force_evict`.
- Implements proper `pin` / `unpin` semantics for GDS:
  - `pin(key)` increments the pin count in `DiskCacheMetadata`,
  - `unpin(key)` decrements it,
  - pinned entries are not evicted because policies respect
    `DiskCacheMetadata.can_evict`.
- Extends `contains(key, pin=True)` to optionally pin an entry when it is
  found, matching LocalDisk behavior.

With these changes, the GDS backend now supports capacity-aware, policy-driven eviction and consistent pinning semantics, while preserving existing APIs.

---

- This PR adds eviction support to the GDS backend's disk cache, similar to LocalDiskBackend.
- The allocator initialization order is slightly refactored so that `use_cufile` is decided before constructing the allocator, but the external behavior is preserved.
- The tests are in `tests/v1/test_gds_evictor.py` and cover LRU, FIFO, and no-eviction scenarios.

---

**If applicable**:


lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -249,6 +255,24 @@ def __init__(
 
         self.stats = None  # TODO: plug into LMCache Statistics
 
+        # Cache policy and size tracking for GDS eviction
+        # Use max_cache_size as the GDS cache size limit (unified naming)
+        self.cache_policy = get_cache_policy(config.cache_policy)
+        max_gds_size: float = getattr(config, "max_gds_size", None) or 0
+        if max_gds_size == 0:
+            env_val = os.environ.get("LMCACHE_MAX_GDS_SIZE")
```
YaoJiayi: We probably don't need to read from env var here. The `config.py` has the logic to read from env var.

lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -354,20 +378,58 @@ def _read_metadata(self, key, filename, subdir_key):
...
+        if pin:
+            with self.hot_lock:
+                cached_meta = self.hot_cache.get(key)
+                if cached_meta is not None:
+                    cached_meta.pin()
+                else:
+                    # Race condition: entry was evicted after being read from disk
+                    # but before it could be pinned. We cannot honor pin=True.
+                    logger.warning(f"Key {key} was evicted before it could be pinned.")
+                    return False
```
YaoJiayi, line 417: good catch.

lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -481,6 +662,11 @@ def insert_key(self, key: CacheEngineKey, memory_obj: MemoryObj) -> None:
         with self.hot_lock:
             # TODO(Jiayi): need to support `cached_positions`.
             self.hot_cache[key] = DiskCacheMetadata(path, size, shape, dtype, None, fmt)
+            logger.info(
+                f"[GDS CACHE] Inserted key={key}, size={size / 1024 / 1024:.2f} MB. "
+                f"Total hot_cache entries: {len(self.hot_cache)}, "
+                f"current_cache_size={self.current_cache_size / 1024 / 1024:.2f} MB"
+            )
```
YaoJiayi, line 656: Maybe turn this into debug log.

lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -709,23 +897,42 @@ def _load_gds(
...
-    ) -> CuFileMemoryAllocator:
+    ) -> MemoryAllocatorInterface:
         assert config.cufile_buffer_size is not None
-        return CuFileMemoryAllocator(config.cufile_buffer_size * 1024**2)
+        size = config.cufile_buffer_size * 1024**2
+        if self.use_cufile:
+            return CuFileMemoryAllocator(size)
+        else:
+            logger.debug("Using GPUMemoryAllocator instead of CuFileMemoryAllocator")
+            return GPUMemoryAllocator(size, device=self.dst_device)
```
YaoJiayi: Could you elaborate a bit why this is needed?

YaoJiayi (commented):
LGTM in general! Just a few comments:)

lmcache/v1/storage_backend/gds_backend.py
MingqinHan: Thanks for the catch! I removed the redundant env var read here.

lmcache/v1/storage_backend/gds_backend.py
MingqinHan: Thanks for the catch! I just turned this into debug log.

lmcache/v1/storage_backend/gds_backend.py
MingqinHan: If we strictly required use_cufile=True, the allocator switch wouldn’t be needed. We keep the fallback so the backend can run when cuFile isn’t available, falling back to the non-cuFile I/O path.

YaoJiayi: I think non-cufile path should be in `local_disk_backend.py`?

YaoJiayi:
@MingqinHan There are code conflicts and the tests are failing:)

lmcache/v1/storage_backend/gds_backend.py
MingqinHan: Yes, I agree. The fallback here turned out to be redundant.
The earlier change was intended to keep the backend runnable without cuFile, but I agree it’s cleaner to keep that logic in local_disk_backend.py. I’ve reverted to the original initialize_allocator accordingly. Thanks for the feedback!

PR #2220, opened 2025-12-10
[refactor]use list to process shape and dtype

chunxiaozheng:
refactor `allocate` related methods, use `list[torch.Size]` and `list[torch.dtype]` rather than `torch.Size` or `torch.dtype`

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -228,9 +228,10 @@ def extract_mm_features(
 
 def get_size_bytes(shapes: list[torch.Size], kv_dtypes: list[torch.dtype]):
     """
-    Calculate the size in bytes with the given shape and dtype.
+    Calculate the size in bytes with the given shapes and dtypes.
     """
-    size_in_bytes = 0
-    for shape, kv_dtype in zip(shapes, kv_dtypes, strict=False):
-        size_in_bytes += shape.numel() * kv_dtype.itemsize
-    return size_in_bytes
+    assert len(shapes) == len(kv_dtypes)
+    return sum(
+        shape.numel() * kv_dtype.itemsize
+        for shape, kv_dtype in zip(shapes, kv_dtypes, strict=False)
+    )
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -226,8 +226,11 @@ def extract_mm_features(
         return ([], [])
 
 
-def get_size_bytes(shape: torch.Size, kv_dtype: torch.dtype):
+def get_size_bytes(shapes: list[torch.Size], kv_dtypes: list[torch.dtype]):
     """
     Calculate the size in bytes with the given shape and dtype.
     """
-    return shape.numel() * kv_dtype.itemsize
+    size_in_bytes = 0
+    for shape, kv_dtype in zip(shapes, kv_dtypes, strict=False):
+        size_in_bytes += shape.numel() * kv_dtype.itemsize
+    return size_in_bytes
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -751,10 +751,10 @@ class TensorMemoryAllocator(MemoryAllocatorInterface):
     @_lmcache_nvtx_annotate
     def _Compute_raw_size(shapes: list[torch.Size], dtypes: list[torch.dtype]) -> int:
         assert len(shapes) == len(dtypes)
-        raw_size = 0
-        for shape, dtype in zip(shapes, dtypes, strict=False):
-            raw_size += shape.numel() * dtype.itemsize
-        return raw_size
+        return sum(
+            shape.numel() * dtype.itemsize
+            for shape, dtype in zip(shapes, dtypes, strict=False)
+        )
 
     @staticmethod
     @_lmcache_nvtx_annotate
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -725,6 +725,27 @@ class MemoryAllocatorInterface(metaclass=abc.ABCMeta):
         """
         return True
 
+    # TODO(chunxiaozheng): remove if after all params replaced by shapes/dtypes
+    def _adapt_shapes_and_dtypes(
+        self,
+        shapes: Union[torch.Size, Tuple[int, ...], list[torch.Size]],
+        dtypes: Union[torch.dtype, list[torch.dtype]],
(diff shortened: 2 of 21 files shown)


maobaolong:
The log showed the test broke and need to fix. @chunxiaozheng  
```python
    def create_mock_memory_obj(backend: LocalCPUBackend, data: bytes) -> MemoryObj:
        """Helper to create a mock MemoryObj with proper structure"""
        tensor = torch.tensor(
            [ord(c) for c in data.decode("latin1") if ord(c) < 256], dtype=torch.uint8
        )
        if len(tensor) == 0:
            tensor = torch.tensor([0], dtype=torch.uint8)
    
>       memory_obj = backend.allocate(
            shape=tensor.shape,
            dtype=tensor.dtype,
            fmt=MemoryFormat.KV_2LTD,
        )
E       TypeError: LocalCPUBackend.allocate() got an unexpected keyword argument 'shape'. Did you mean 'shapes'?
```

lmcache/integration/vllm/utils.py
```diff
@@ -226,8 +226,11 @@ def extract_mm_features(
         return ([], [])
 
 
-def get_size_bytes(shape: torch.Size, kv_dtype: torch.dtype):
+def get_size_bytes(shapes: list[torch.Size], kv_dtypes: list[torch.dtype]):
     """
     Calculate the size in bytes with the given shape and dtype.
```
ApostaC: nit: we can be clearer here 
1. Saying it expects len(shapes) == len(kv_dtypes)
2. It's summing up the size for all (shape, dtype) pair.

lmcache/v1/memory_management.py
```diff
@@ -112,30 +112,46 @@ class MemoryObjMetadata:
     # Positions when the cache is stored
     cached_positions: Optional[torch.Tensor] = None
 
+    # shapes and dtypes should be used in the future
+    shapes: Optional[list[torch.Size]] = None
+    dtypes: Optional[list[torch.dtype]] = None
```
ApostaC, line 117: Seems like we are changing almost all the related places from shape/dtype to shapes/dtypes.
How about let's directly **replace** the shape/dtype with **shapes/dtypes**. Therefore, the `allocate()` function's type annotation don't need to include non-list type anymore.

I understand there may need to be more changes in the unit test if we do this, but I'm good with that. WDYT?

ApostaC (commented):
Thanks for the refactoring!

Is it possible to completely replace the `shape/dtype` with `shapes/dtypes`?

chunxiaozheng:
> The log showed the test broke and need to fix. @chunxiaozheng
> 
> ```python
>     def create_mock_memory_obj(backend: LocalCPUBackend, data: bytes) -> MemoryObj:
>         """Helper to create a mock MemoryObj with proper structure"""
>         tensor = torch.tensor(
>             [ord(c) for c in data.decode("latin1") if ord(c) < 256], dtype=torch.uint8
>         )
>         if len(tensor) == 0:
>             tensor = torch.tensor([0], dtype=torch.uint8)
>     
> >       memory_obj = backend.allocate(
>             shape=tensor.shape,
>             dtype=tensor.dtype,
>             fmt=MemoryFormat.KV_2LTD,
>         )
> E       TypeError: LocalCPUBackend.allocate() got an unexpected keyword argument 'shape'. Did you mean 'shapes'?
> ```

@maobaolong Thanks! fixed.

lmcache/integration/vllm/utils.py
chunxiaozheng: @ApostaC Thanks for your suggestion, I have updated, PTAL!

lmcache/v1/memory_management.py
chunxiaozheng: @ApostaC Good idea, this is consistent with my first idea, but there are too many changes 😭, the unit tests need to adapt, and some logics in the code also need to adapt.

ApostaC:
We just fixed the comprehensive test. Triggering it again.

lmcache/v1/memory_management.py
ApostaC: Got it. No worries, we can do it later

lmcache/v1/storage_backend/connector/base_connector.py
ApostaC, line 1: In the remote storage connectors, we will serialize the memory object with `get_shape` and `get_dtype`. For example: https://github.com/LMCache/LMCache/blob/bf74b5e27a0b203ae1cca91c11f9a333cb93ea26/lmcache/v1/storage_backend/connector/fs_connector.py#L210-L216

Such wrong metadatas will be loaded and reused during `retrieve`, causing issues when using remote connectors.

ApostaC (commented):
LGTM in general!

Two things:
1. In remote connectors, there are codes that call `get_shape` and `get_dtype` to generate some metadata for each memory object when saving. Those should also be changed to `get_shapes` and `get_dtypes` as well.
2. We should eventually remove the `dtype` and `shape` and replace them with shapes and dtypes. I'm actually thinking about having a special wrapper class for them so it will be much more flexible if we need to do something similar in the future.

We don't need to change 1 and 2 in this PR (changes for 1 are preferred if they are not too complicated), but let's at least create an issue to track it, and let's do it soon once we've got cycles @chunxiaozheng ?

chunxiaozheng:
> LGTM in general!
> 
> Two things:
> 
> 1. In remote connectors, there are codes that call `get_shape` and `get_dtype` to generate some metadata for each memory object when saving. Those should also be changed to `get_shapes` and `get_dtypes` as well.
> 2. We should eventually remove the `dtype` and `shape` and replace them with shapes and dtypes. I'm actually thinking about having a special wrapper class for them so it will be much more flexible if we need to do something similar in the future.
> 
> We don't need to change 1 and 2 in this PR (changes for 1 are preferred if they are not too complicated), but let's at least create an issue to track it, and let's do it soon once we've got cycles @chunxiaozheng ?

@ApostaC Thanks for your suggestion, I have add `get_shapes` and `get_dtypes` interface in `MemoryObj`, and `get_shape` and `get_dtype` can be gradually replaced in the future.

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: Yeah, In current refactor, there won't be any problems here temporarily. #2230 implements `dsa` only in `LocalCPUBackend`, I will gradually adapt to other backends and replace 'get_shape' and 'get_dtype' here.

ApostaC (approved):
LGTM!

maobaolong (approved):
LGTM! Thanks!

PR #2223, opened 2025-12-10
[Controller] Refactor to remove the ip->instance relationship

maobaolong:

Remove the ip->instanceNode mapping, as it is actually useless almost, the original  `ip->instanceNode` make we hard to handle whether we should support same instanceNode but their IPs are different, such as xPyD, maybe the instanceId of them can be same, but the IPs of them are different.

So remove it make the RegisterTree tidy and clear.

- RegistryTree
  - instances: dict[str, InstanceNode] -- (instance_id -> InstanceNode)
      - workers: dict[int, WorkerNode] -- (worker_id -> WorkerNode)
          - kv_store: dict[str, set[int]] -- location -> set[chunk_hash]


maobaolong:
@KuntaiDu @ApostaC Would you like to help to review this PR? Thanks!

lmcache/v1/cache_controller/utils.py
```diff
@@ -182,62 +175,49 @@ def deregister_worker(
...
         """Get an instance by instance_id."""
-        return self.instance_id_index.get(instance_id)
+        return self.instances.get(instance_id)
 
     def get_instance_by_ip(self, ip: str) -> Optional[InstanceNode]:
-        """Get an instance by IP address. Returns first instance if multiple exist."""
-        ip_instances = self.instances.get(ip)
-        if ip_instances:
-            return next(iter(ip_instances.values()), None)
+        """Get an instance by IP address."""
```
sammshen, line 202: do we still need this method? 

it has become O(n) now so maybe we could still maintain an ip -> [worker node] mapping for `get_instances_by_ip`

lmcache/v1/cache_controller/utils.py
```diff
@@ -293,37 +273,45 @@ def find_kv(
...
-            for instance_id, instance_node in ip_instances.items():
+        for instance_id, instance_node in self.instances.items():
+            for worker_id, worker_node in instance_node.workers.items():
+                # Only exclude the specific worker identified by
+                # BOTH instance_id and worker_id
                 if (
                     exclude_instance_id is not None
+                    and exclude_worker_id is not None
                     and instance_id == exclude_instance_id
+                    and worker_id == exclude_worker_id
```
sammshen: then we will still look at the sibling workers on the same instance here, is this desired?

sammshen:
LGTM in general. left some  comments.

Also, I think the xPyD (or any PD case) may not be the best example of one instance id, multiple ip since the P and D are separate instances (correct me if I'm wrong). a simpler case might just be pipeline parallelism.

lmcache/v1/cache_controller/utils.py
maobaolong: This is a rest api `/query_instance` , although it become O(n) , but if we have only dozens of workers, its impact is minimal, and I found no caller in this repository.

lmcache/v1/cache_controller/utils.py
maobaolong: Done. After a think twice, it would be better to exclude whole instance, so we should only need pass a exclude_instance_id.

maobaolong:
@sammshen Thanks for your review, I correct the exclude logic, PTAL.

sammshen (approved):
LGTM! thanks for address comments

chunxiaozheng (approved):
lgtm

PR #2224, opened 2025-12-10
[Controller] Add RWLockWithTimeout and FastLockWithTimeout, setup to RegistryTree and workerNode

maobaolong:

Use RWLock to help protect thread safe of controller register tree.

**If applicable**:

### DIFF
--- a/lmcache/v1/cache_controller/controllers/kv_controller.py
+++ b/lmcache/v1/cache_controller/controllers/kv_controller.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any
 # First Party
 from lmcache.logging import init_logger
 from lmcache.v1.cache_controller.message import (
+    BatchedKVOperationMsg,
     BatchedP2PLookupMsg,
     BatchedP2PLookupRetMsg,
     CheckFinishMsg,
@@ -15,9 +16,6 @@ from lmcache.v1.cache_controller.message import (
     CompressRetMsg,
     DecompressMsg,
     DecompressRetMsg,
-    KVAdmitMsg,
-    KVEvictMsg,
-    KVOperationMsg,
     LookupMsg,
     LookupRetMsg,
     MoveMsg,
@@ -49,59 +47,18 @@ class KVController:
     def __init__(self, registry: RegistryTree) -> None:
         # TODO(Jiayi): remove this hardcode
         self.token_database = ChunkedTokenDatabase()
-
-        # Track sequence discontinuity count for metrics
-        self.seq_discontinuity_count = 0
-
         self.registry = registry
         self.cluster_executor: Any = None
 
     def _setup_metrics(self) -> None:
         prometheus_logger = PrometheusLogger.GetInstanceOrNone()
         if prometheus_logger is not None:
-            # TODO(baoloongmao): Cache values for better performance
-            prometheus_logger.kv_pool_keys_count.set_function(self._get_kv_pool_size)
+            prometheus_logger.kv_pool_keys_count.set_function(
+                self.registry.get_total_kv_count
+            )
             prometheus_logger.kv_op_seq_discontinuity_count.set_function(
-                lambda: self.seq_discontinuity_count
+                self.registry.get_seq_discontinuity_count
             )
-
-    def _get_kv_pool_size(self) -> int:
-        return self.registry.get_total_kv_count()
-
-    def check_sequence_number(self, msg: KVOperationMsg) -> None:
-        """
-        Check if the sequence number is continuous for the given source.
-
-        Args:
-            msg: KVOperationMsg
-        """
-        instance_id = msg.instance_id
-        worker_id = msg.worker_id
-        location = msg.location
-        seq_num = msg.seq_num
-
-        last_seq_num = self.registry.get_seq_num(instance_id, worker_id, location)
-
-        if last_seq_num is None:
-            # First message from this source
-            self.registry.update_seq_num(instance_id, worker_id, location, seq_num)
-            return
-
-        expected_seq = last_seq_num + 1
-        if seq_num != expected_seq:
-            # Sequence number discontinuity detected
-            self.seq_discontinuity_count += 1
-            logger.warning(
-                "KV operation sequence discontinuity detected: "
-                "key=%s, expected_seq=%s, actual_seq=%s, gap=%s",
-                (instance_id, worker_id, location),
-                expected_seq,
-                seq_num,
-                seq_num - expected_seq,
-            )
-
-        # Update tracker with current sequence number
-        self.registry.update_seq_num(instance_id, worker_id, location, seq_num)
 
     def post_init(
         self, reg_controller: "RegistrationController", cluster_executor: Any
@@ -155,17 +112,17 @@ class KVController:
         assert self.cluster_executor is not None
         return await self.cluster_executor.execute("check_finish", msg)
 
-    async def admit(self, msg: KVAdmitMsg) -> None:
-        """
-        Admit a new kv chunk.
-        """
-        self.registry.admit_kv(msg.instance_id, msg.worker_id, msg.location, msg.key)
+    async def handle_batched_kv_operations(self, msg: BatchedKVOperationMsg) -> None:
+        """Handle batched KV operations by forwarding to registry."""
+        if not msg.operations:
+            return
 
-    async def evict(self, msg: KVEvictMsg) -> None:
-        """
-        Evict a kv chunk.
-        """
-        self.registry.evict_kv(msg.instance_id, msg.worker_id, msg.location, msg.key)
+        if not self.registry.handle_batched_kv_operations(msg):
+            logger.warning(
+                "Failed to handle batched KV operations, instance: %s, worker: %d",
+                msg.instance_id,
+                msg.worker_id,
+            )
 
     # TODO(Jiayi): The current implementation does not handle
     # the case where the prefix chunks are evicted while the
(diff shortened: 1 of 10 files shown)


lmcache/v1/cache_controller/utils.py
```diff
@@ -154,140 +181,176 @@ def register_worker(
...
         self, instance_id: str, worker_id: int, timestamp: float
     ) -> bool:
         """Update worker heartbeat timestamp. Returns True if successful."""
-        worker_node = self.get_worker(instance_id, worker_id)
-        if worker_node is None:
-            return False
-        worker_node.last_heartbeat_time = timestamp
-        return True
+        # TODO(baoloongmao): Move timeout values to configuration
+        with self._rwlock.read_lock(timeout=0.1):  # Write operation
```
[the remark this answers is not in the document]
maobaolong: I think read_lock for heartbeat is enough.

maobaolong:
@sammshen @ApostaC @chunxiaozheng Would you please help to take a look at this PR? Thanks

lmcache/v1/cache_controller/utils.py
```diff
@@ -42,44 +39,138 @@ class WorkerNode:
...
-        if location not in self.kv_store or key not in self.kv_store[location]:
-            return False
-        self.kv_store[location].remove(key)
-        if not self.kv_store[location]:
-            del self.kv_store[location]
-        return True
+    def handle_batched_kv_operations(
+        self,
+        msg: BatchedKVOperationMsg,
+        on_seq_discontinuity: Optional[Callable[[], None]] = None,
```
KuntaiDu: Naming-wise maybe `on_seq_num_out_of_order` is better.

KuntaiDu: Initially I misread this func as KV cache sequence is missing something in the middle

lmcache/v1/cache_controller/utils.py
```diff
@@ -96,49 +187,160 @@ class InstanceNode:
...
+    ) -> Optional[tuple[KVChunkInfo, Optional[str], set[int]]]:
+        """
+        Find a key in any worker within this instance.
+        Returns: (KVChunkInfo, peer_init_url, keys) if found, None otherwise.
+        """
+        # Snapshot workers to avoid RuntimeError if dict changes during iteration
+        workers_snapshot = list(self.workers.items())
+        for worker_id, worker_node in workers_snapshot:
+            result = worker_node.find_key(key)
+            if result is not None:
```
KuntaiDu: `if result := worker_node.find_key(key) is not None:`

lmcache/v1/cache_controller/utils.py
```diff
@@ -96,49 +187,160 @@ class InstanceNode:
...
+        Find a key in any worker within this instance, returning only KVChunkInfo.
+        """
+        # Snapshot workers to avoid RuntimeError if dict changes during iteration
+        workers_snapshot = list(self.workers.items())
+        for worker_id, worker_node in workers_snapshot:
+            kv_info = worker_node.find_key_simple(key)
+            if kv_info is not None:
+                return KVChunkInfo(self.instance_id, worker_id, kv_info.location)
+        return None
+
```
KuntaiDu, line 264: For the following two function, maybe:

```
    def has_worker_with_ip(self, ip: str) -> bool:
        return any(worker.ip == ip for worker in list(self.worker.values()))

    def get_total_kv_count(self) -> int:
        return sum(worker.get_kv_count() for worker in list(self.worker.values()))
```

lmcache/v1/cache_controller/utils.py
```diff
@@ -96,49 +187,160 @@ class InstanceNode:
...
+    def find_key_simple(self, key: int) -> Optional[KVChunkInfo]:
+        """
+        Find a key in any worker within this instance, returning only KVChunkInfo.
+        """
+        # Snapshot workers to avoid RuntimeError if dict changes during iteration
+        workers_snapshot = list(self.workers.items())
+        for worker_id, worker_node in workers_snapshot:
+            kv_info = worker_node.find_key_simple(key)
+            if kv_info is not None:
```
KuntaiDu: `if kv_info := worker_node.find_key_simple(key) is not None`

KuntaiDu: This is not required though --- just feel like the code is slightly shorter if using `:=`

lmcache/v1/cache_controller/utils.py
KuntaiDu: Oh I see --- you are intentionally avoiding using the iterator to handle the code. Personally I feel like iterator is cleaner but if you prefer the for loop way I am also fine.

KuntaiDu (approved):
This PR does two things:
- Add locks
- Attribute the action corresponding to the message to `WorkerNode`, more object-oriented programming so yes yes.

I like both changes and the logic seems to be straightforward. So approve but with minor comments.

lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -200,33 +157,30 @@ async def batched_p2p_lookup(
...
-        location = result.location
-        assert self.reg_controller is not None
-        peer_init_url = self.reg_controller.get_peer_init_url(instance_id, worker_id)
-        assert peer_init_url is not None
-        current_instance_keys = self.registry.get_worker_kv_keys(
-            instance_id, worker_id, location
-        )
+        kv_info, peer_init_url, current_keys = result
+        if peer_init_url is None:
+            return BatchedP2PLookupRetMsg(layout_info=[("", "", 0, "")])
```
chunxiaozheng, line 173: could we add a `warning` log here, the `peer_init_url` should not be `None` in normal case.

lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -200,33 +157,30 @@ async def batched_p2p_lookup(
...
+        # Count hits efficiently
         num_hit_chunks = 0
-        for key in msg.hashes:
-            if key not in current_instance_keys:
+        for key in hashes:
+            if key not in current_keys:
                 break
             num_hit_chunks += 1
 
         return BatchedP2PLookupRetMsg(
```
chunxiaozheng, line 182: we should add `try-except` here, if server occurs errors, we should return a default mst to client.

lmcache/v1/cache_controller/locks.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self):
+        self._readers = 0
+        self._writers_waiting = 0
+        self._writer_active = False
+        self._condition = threading.Condition(threading.Lock())
+
+    def acquire_read(self, timeout: Optional[float] = None) -> bool:
+        """Acquire a read lock with optional timeout."""
+        deadline = time.time() + timeout if timeout is not None else None
```
sammshen: nit:
change all time.time() to time.monotonic()
used internally by threading.Condition

lmcache/v1/cache_controller/locks.py
```diff
@@ -0,0 +1,139 @@
...
+
+        with self._condition:
+            while self._writer_active or self._writers_waiting > 0:
```
sammshen, line 42: nit: 
maybe add a small comment that sustained writes may starve reads?

lmcache/v1/cache_controller/locks.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class RWLockTimeoutError(Exception):
+    """Exception raised when a lock acquisition times out."""
+
+    pass
+
+
+class RWLockWithTimeout:
+    """
```
sammshen, line 22: timeout=None has different semantics between acquire_* and the context managers (infinite wait vs default timeout)

acquire_read(None) -> wait forever

read_lock(None) -> 50ms

should we make this consistent?

lmcache/v1/cache_controller/locks.py
```diff
@@ -0,0 +1,139 @@
...
+        if timeout is None:
+            return self._lock.acquire()
+        return self._lock.acquire(timeout=timeout)
+
+    def release(self):
+        """Release the lock."""
+        self._lock.release()
+
+    def __enter__(self):
+        # Fast path: try non-blocking acquire first (no syscall overhead)
```
sammshen: I think blocking=False still does syscall just avoids context switch

lmcache/v1/cache_controller/locks.py
```diff
@@ -0,0 +1,139 @@
...
+import threading
+import time
+
+
+class RWLockTimeoutError(Exception):
+    """Exception raised when a lock acquisition times out."""
+
+    pass
+
+
```
sammshen, line 20: nit: 
add comment that both of these locks are not reentrant

lmcache/v1/cache_controller/locks.py
```diff
@@ -0,0 +1,139 @@
...
+    def release(self):
+        """Release the lock."""
+        self._lock.release()
+
+    def __enter__(self):
+        # Fast path: try non-blocking acquire first (no syscall overhead)
+        if self._lock.acquire(blocking=False):
+            return self
+        # Slow path: reduced timeout for faster failure detection
```
sammshen, line 141: I think the todo is right and we should not set a default hidden 10 second timeout for a lock that is supposed to be fast

sammshen (approved):
LGTM just some semantic nits on the new locks module

lmcache/v1/cache_controller/controllers/kv_controller.py
maobaolong: Add some error handling

lmcache/v1/cache_controller/locks.py
maobaolong: I updated code and change to default `wait forever`

maobaolong:
@KuntaiDu @sammshen Thanks for your review and suggestion!

@chunxiaozheng Would you like to take another look?

chunxiaozheng (approved):
lgtm

PR #2225, opened 2025-12-10
[Core]: Enable remote connectors to be loaded on the fly

### LINKED ISSUE #2161: Investigate how usable the Remote Connector Framework is for extending
From https://github.com/LMCache/LMCache/pull/2118#discussion_r2582155410, @sammshen asked if `Remote Connector Framework` could also be aligned with storage and runtime plugins for more consistency and better usability.

Looking at it initially, it would seem to need some work to make it easier for users to extend and add a remote connector.

A clear and concise means to add remote connectors.

**Implementation**

- [ ] Add hooks to load remote connectors on the fly (#2225)
- [ ] Add unit tests (#2225)
- [ ] Update documentation accordingly (#2225)

hickeyma:

Enables remote connectors to be added on the fly similar to storage and runtime plugins.

The current mechanism uses the [external adapter](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/connector/external_adapter.py) where you write the connector and the URL has to be of the form: `external://host:port/module_path/?connector_name=ConnectorName`.

This PR adds a more plug and play approach which follows the pattern used for storage and runtime plugins. The current external adapter will not be removed for now for backwards compatibility. 

Fixes #2161  

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -270,10 +270,13 @@ class ConnectorManager:
                     continue
                 self.adapters.append(adapter_instance)
                 logger.info(f"Discovered adapter: {adapter_class.__name__}")
+            except (ImportError, AttributeError) as e:
+                logger.error(
+                    f"Failed to load remote connector {remote_connector_plugin} due to import/attribute error: {e}"
+                )
             except Exception as e:
                 logger.error(
-                    f"Failed to create remote connector {remote_connector_plugin} "
-                    f"adapter: {str(e)}"
+                    f"Failed to create remote connector {remote_connector_plugin} adapter: {str(e)}"
                 )
 
     def create_connector(self) -> RemoteConnector:
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -224,6 +224,7 @@ class ConnectorManager:
             logger.warning(
                 "Configuration not available to parse remote connector adapters."
             )
+            return
 
         # Get the list of allowed remote connector adapters if configured
         remote_storage_plugins = (
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -234,11 +234,19 @@ class ConnectorManager:
 
         for remote_connector_plugin in remote_connector_plugins:
             try:
-                module_path = config.extra_config.get(
+                extra_config = config.extra_config
+                if extra_config is None:
+                    logger.warning(
+                        f"Remote connector {remote_connector_plugin} configuration is missing 'extra_config'."
+                    )
+                    continue
+
+                module_path = extra_config.get(
                     f"remote_connector.{remote_connector_plugin}.module_path"
                 )
-                class_name = config.extra_config.get(
+                class_name = extra_config.get(
                     f"remote_connector.{remote_connector_plugin}.class_name"
+                )
                 )
 
                 if not module_path or not class_name:
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -226,32 +226,32 @@ class ConnectorManager:
             )
 
         # Get the list of allowed remote connector adapters if configured
-        remote_connector_plugins = (
-            set(config.remote_connector_plugins)
-            if config.remote_connector_plugins
+        external_storage_plugins = (
+            set(config.external_storage_plugins)
+            if config.external_storage_plugins
             else set()
         )
 
-        for remote_connector_plugin in remote_connector_plugins:
+        for external_storage_plugin in external_storage_plugins:
             try:
                 extra_config = config.extra_config
                 if extra_config is None:
                     logger.warning(
-                        f"Remote connector {remote_connector_plugin} configuration is "
+                        f"Remote connector {external_storage_plugin} configuration is "
                         f"missing 'extra_config'."
                     )
                     continue
 
                 module_path = extra_config.get(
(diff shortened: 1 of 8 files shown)


lmcache/v1/storage_backend/connector/__init__.py
```diff
@@ -178,10 +178,11 @@ def __init__(
             metadata=metadata,
         )
         self.adapters: List[ConnectorAdapter] = []
-        self._discover_adapters()
+        self._remote_adapters_builtin_launcher()
```
sammshen, line 181: this was probably in documentation already but I think having a comment in the code defining "builtin" and "plugin" would be great

sammshen: or maybe it's already sufficiently clear, like there's a comment "Discover all modules in the connector package" below

maobaolong:
@hickeyma Could you add a config example to show how this can improve usable?

maobaolong:
> @hickeyma Could you add a config example to show how this can improve usable?

Since we have a internal connector running base on this framework, it would be nice to keep the backward compatible.

hickeyma:
> @hickeyma Could you add a config example to show how this can improve usable?

@maobaolong Similar to the following:

```yaml
external_storage_plugins: ["test"]
extra_config:
    external_storage_plugin.test.module_path: test_remote_connector.test_adapter
    external_storage_plugin.test.class_name: TestConnectorAdapter
```

hickeyma:
> Since we have a internal connector running base on this framework, it would be nice to keep the backward compatible.

@maobaolong Can you explain more what you mean by this? Sorry I don't fully understand.

lmcache/v1/storage_backend/connector/__init__.py
hickeyma: Yes, there is a comment at top of `_remote_adapters_builtin_launcher` definition as follows:

```python
"""Automatically load all builtin remote connector adapters."""
```

@sammshen Is this sufficient?

sammshen (approved):
LGTM, the documentation is great

hickeyma:
Ping @maobaolong

maobaolong (commented):
If i understand correct, i found that you did a concept refactor, which is rename external remote connector to a new concept `external_storage_plugins`.

And i'm not sure whether both the existing user and new user of lmcache can distinguish the `storage_plugin` and `external_storage_plugin`?

The original naming `remote connector` can let us know the `connector` is the extension of `remote backend`, but not other storage backend.

Besides, without this refactor, the existing state can support dynamic seek and load third-party remote connectors, so i'm sorry for didn't catch the motivation of this PR.

hickeyma:
Thanks @maobaolong for you feedback above. I will try and tackle your questions below.

> If i understand correct, i found that you did a concept refactor, which is rename external remote connector to a new concept external_storage_plugins.

Yes. The [external adapter](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/connector/external_adapter.py) approach is where you write the connector and the URL has to be of the `form: external://host:port/module_path/?connector_name=ConnectorName.` The idea of the refactor is to:

- Make it easier to extend using plugin approach
- Follow the pattern used for runtimes and storage plugins
- Follow the pattern used for the built-in `RemoteConnectors`

> And i'm not sure whether both the existing user and new user of lmcache can distinguish the storage_plugin and external_storage_plugin?

`external` can be replaced with `remote` but to me its the same as the  [external adapter](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/connector/external_adapter.py) in that you extend from `RemoteConnector`  and you access the storage through a URL.

> The original naming remote connector can let us know the connector is the extension of remote backend, but not other storage backend.

As I said above: `external` can be replaced with `remote` if we think it makes more sense.

> Besides, without this refactor, the existing state can support dynamic seek and load third-party remote connectors, so i'm sorry for didn't catch the motivation of this PR.

Thats ok. This PR also does the same but its more standard plugin and play like we did for the storage plugins.

maobaolong:
@hickeyma Sorry for the late reply. I prefer `remote` compare to `external`, as we have a `external_backends` concept before, otherwise, both are good for me.

After a double think, i think i've got your motivation. And the `remote_url` is a tricky and complex legacy, if this refactor can remove the remote_url, it would be a great job, instead, we can set `remote=True` to enable remote_backend, and use `extra_config` for each remote_connector, maybe that will introduce many changes if we remove the `remote_url`, to keep compatible, we can keep the `remote_url` alive for several months, and mark it as a deprecated `config`.

hickeyma:
Thanks @maobaolong  for the feedback. Answers below to the points raised.

> @hickeyma Sorry for the late reply. I prefer remote compare to external, as we have a external_backends concept before, otherwise, both are good for me.

I agree. Updated to use "remote" instead of "external" in 9e2c5031dc60e79fb06c2c560483efc5740aa9f7.

> After a double think, i think i've got your motivation. And the remote_url is a tricky and complex legacy, if this refactor can remove the remote_url, it would be a great job, instead, we can set remote=True to enable remote_backend, and use extra_config for each remote_connector, maybe that will introduce many changes if we remove the remote_url, to keep compatible, we can keep the remote_url alive for several months, and mark it as a deprecated config.

So, I have kept the external connector for backwards compatibility and added deprecation messages in docs and code in 40c0d323a182b6172d4eb012d1693dcbe1c08436. The remote storage and external connector  both use `remote_url` but the remote storage lacks the complexity of the former as shown in the docs.

maobaolong:
Sounds good, will take a look tomorrow

hickeyma:
> Sounds good, will take a look tomorrow

Not a bother @maobaolong; whenever you get a chance, thanks.

maobaolong (approved):
@hickeyma  It LGTM, thanks for this refactor.

PR #2230, opened 2025-12-11
[feature][core]support dsa in LocalCPUBackend

chunxiaozheng:
support dsa in `LocalCPUBackend`

### DIFF
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -42,6 +42,8 @@ class LMCacheEngineMetadata:
     # TODO(baoloongmao): first_rank should be configurable
     first_rank = 0
     served_model_name: Optional[str] = None
+    """chunk size"""
+    chunk_size: int = 256
     """ Manager for groups of layers with identical KV cache structure """
     kv_layer_groups_manager: KVLayerGroupsManager = field(
         default_factory=KVLayerGroupsManager
@@ -50,6 +52,14 @@ class LMCacheEngineMetadata:
     def is_first_rank(self) -> bool:
         """Check if the current worker is the first rank"""
         return self.worker_id == self.first_rank
+
+    # TODO(chunxiaozheng): some uts do not `build_kv_layer_groups`
+    def get_dtypes(self) -> list[torch.dtype]:
+        if self.kv_layer_groups_manager.kv_layer_groups:
+            return [
+                group.dtype for group in self.kv_layer_groups_manager.kv_layer_groups
+            ]
+        return [self.kv_dtype]
 
 
 @dataclass
--- a/lmcache/v1/gpu_connector.py
+++ b/lmcache/v1/gpu_connector.py
@@ -103,10 +103,18 @@ class GPUConnectorInterface(metaclass=abc.ABCMeta):
         """Get the shape of the data given the number of tokens."""
         raise NotImplementedError
 
+    def get_shapes(self, num_tokens: int) -> list[torch.Size]:
+        """Get the shape of the data given the number of tokens."""
+        return [self.get_shape(num_tokens)]
+
     def initialize_kvcaches_ptr(self, **kwargs):
         """Initialize the kvcaches pointers if not already initialized."""
         if "kvcaches" in kwargs:
             self.kvcaches = kwargs["kvcaches"]
+
+    def init_group_info(self):  # noqa: B027
+        """Initialize the group info."""
+        pass
 
 
 class VLLMPagedMemGPUConnectorV2(GPUConnectorInterface):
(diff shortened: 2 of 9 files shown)


lmcache/config.py
```diff
@@ -42,6 +42,8 @@ class LMCacheEngineMetadata:
     # TODO(baoloongmao): first_rank should be configurable
     first_rank = 0
     served_model_name: Optional[str] = None
+    """chunk size"""
```
sammshen, line 45: nit:

this is a config so should we add it to the metadata?

lmcache/v1/gpu_connector.py
```diff
@@ -361,6 +369,226 @@ def get_shape(self, num_tokens: int) -> torch.Size:
...
+
+    def init_group_info(self):
+        assert self.metadata.kv_layer_groups_manager.kv_layer_groups
+        self.group_layers = []
+        self.group_hidden_dim_sizes = []
+        if self.use_gpu:
+            self.group_tmp_buffer = []
+        for group in self.metadata.kv_layer_groups_manager.kv_layer_groups:
+            # init layers
+            num_layers = len(group.layer_indices)
```
sammshen: isn't there a num_layers property?

lmcache/v1/gpu_connector.py
```diff
@@ -361,6 +369,226 @@ def get_shape(self, num_tokens: int) -> torch.Size:
...
+                    ]
+                )
+                tmp_buf = torch.empty(
+                    tmp_buf_shape, dtype=group.dtype, device=self.device
+                )
+                self.group_tmp_buffer.append(tmp_buf)
+        logger.info("init group info success in VLLMPagedMemGPUConnectorV3")
+
+    def _initialize_kv_cache_pointers(self):
+        if self.group_kv_cache_pointers_on_gpu is None:
```
sammshen: nit: 
return if self.group_kv_cache_pointers_on_gpu is not None to avoid massive indentation

lmcache/v1/gpu_connector.py
```diff
@@ -361,6 +369,226 @@ def get_shape(self, num_tokens: int) -> torch.Size:
...
+        metadata: LMCacheEngineMetadata,
+        device: torch.device,
+        use_gpu: bool = False,
+    ):
+        assert device.type == "cuda", "The device should be CUDA."
+        self.metadata = metadata
+        self.device = device
+        self.use_mla = metadata.use_mla
+        self.chunk_size = metadata.chunk_size
+        self.use_gpu = use_gpu
```
sammshen, line 393: do we need this if asserting devce.type == "cuda"

sammshen: or is this temp

lmcache/config.py
ApostaC: I talked to Baolong offline. The reason here is that the chunk size is needed by the GPUConnector during initialization. Only LMCacheEngineMetadata is passed by then.

lmcache/v1/gpu_connector.py
```diff
@@ -103,11 +103,19 @@ def get_shape(self, num_tokens: int) -> torch.Size:
...
+        """Get the shape of the data given the number of tokens."""
+        return [self.get_shape(num_tokens)]
+
     def initialize_kvcaches_ptr(self, **kwargs):
         """Initialize the kvcaches pointers if not already initialized."""
         if "kvcaches" in kwargs:
             self.kvcaches = kwargs["kvcaches"]
 
+    def init_group_info(self):  # noqa: B027
+        """Initialize the group info."""
```
ApostaC: nit: can we make the docstring more informative here?

lmcache/v1/gpu_connector.py
ApostaC: Same ask here. What's the meaning of the `device`?

lmcache/v1/gpu_connector.py
```diff
@@ -361,6 +369,226 @@ def get_shape(self, num_tokens: int) -> torch.Size:
...
+                        1 if self.use_mla else 2,
+                        num_layers,
+                        self.chunk_size,
+                        hidden_dim_size,
+                    ]
+                )
+                tmp_buf = torch.empty(
+                    tmp_buf_shape, dtype=group.dtype, device=self.device
+                )
+                self.group_tmp_buffer.append(tmp_buf)
```
ApostaC: Can we do something to reuse the `get_shapes` and `get_dtypes` function here? Just like

```suggestion
        # outside the loop
        if self.use_gpu:
            dtypes = sellf.metadata.get_dtypes()
            shapes = self.get_shapes(self.chunk_size)
            self.group_tmp_buffer = [
                 torch.empty(shape, dtype=dtype, device=self.device)
                 for shape, dtype in zip(shapes, dtypes)
             ]
```

lmcache/v1/memory_management.py
```diff
@@ -405,6 +418,14 @@ def __init__(
         self.valid = True
         self.lock = threading.Lock()
         self.parent_allocator = parent_allocator
+        self.group_length = [0]
```
ApostaC: nit: can we change the variable name to make it more explicit that this list is something like the "prefix sum" of the group lengths instead of the length of each group?

lmcache/v1/memory_management.py
```diff
@@ -548,6 +566,27 @@ def can_evict(self) -> bool:
...
+            return None
+        assert self.meta.shapes is not None
+        assert self.meta.dtypes is not None
+        begin = self.group_length[index]
+        end = self.group_length[index + 1]
+        return (
+            self.raw_data[begin:end]
+            .view(self.meta.dtypes[index])
+            .view(self.meta.shapes[index])
+        )
```
ApostaC, line 594: Can we leave a `TODO` comment here? I would like to explicity say that in the future we will force the use of `shapes` and `dtypes`?

ApostaC (commented):
Otherwise LGTM!

lmcache/config.py
chunxiaozheng: Yeah, thanks for your explanation.

lmcache/v1/gpu_connector.py
chunxiaozheng: I am sorry, I overlooked it before and I have updated.

lmcache/v1/gpu_connector.py
chunxiaozheng: Thanks for your suggestion, I have updated!

lmcache/v1/gpu_connector.py
chunxiaozheng: done!

lmcache/v1/gpu_connector.py
chunxiaozheng: done!

lmcache/v1/memory_management.py
chunxiaozheng: done!

lmcache/v1/memory_management.py
chunxiaozheng: done!

lmcache/v1/gpu_connector.py
chunxiaozheng: @sammshen @ApostaC `use_gpu` means whether create tmp buffer, and `device` is used in `torch.empty(tmp_buf_shape, dtype=tmp_buf_dtype, device=self.device)`, this is the implementation of reference `VLLMPagedMemGPUConnectorV2`

chunxiaozheng:
@ApostaC @sammshen I have updated, PTAL! Thanks!

sammshen (approved):
LGTM!

maobaolong (approved):
LGTM! Thanks for this great work!

ApostaC (approved):
LGTM!

PR #2251, opened 2025-12-15
[Controller] Refactor heartbeat from PUSH to REQ-REP mode and supply heartbeat command framework

maobaolong:

- HeartbeatMsg now inherits from WorkerReqMsg instead of RegisterMsg
- Added HeartbeatCommand struct for extensible commands
- HeartbeatRetMsg now uses command pattern
- Worker heartbeat handler dispatches commands via _handle_heartbeat_command
- Updated registration_controller heartbeat to return HeartbeatCommand
- Updated benchmark heartbeat handler for command-based response

The command abstraction allows controller to send various commands to workers through the heartbeat mechanism. New commands can be added by extending HeartbeatCommandType enum without changing the message structure.

**If applicable**:

### DIFF
--- a/tests/tools/test_controller_zmq_benchmark.py
+++ b/tests/tools/test_controller_zmq_benchmark.py
@@ -158,7 +158,7 @@ class TestOperationHandlers:
 
         assert isinstance(msg, HeartbeatMsg)
         assert handler.get_message_count(controller_zmq_benchmark) == 1
-        assert not handler.use_req_socket()
+        assert handler.use_req_socket()
 
     def test_register_handler(self, controller_zmq_benchmark, test_data):
         """Test RegisterHandler functionality"""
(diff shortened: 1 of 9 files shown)


maobaolong:
@ApostaC @sammshen @chunxiaozheng Would you please help to take a look at this PR? Thanks!

tests/tools/test_controller_zmq_benchmark.py
```diff
@@ -158,7 +158,7 @@ def test_heartbeat_handler(self, controller_zmq_benchmark, test_data):
 
         assert isinstance(msg, HeartbeatMsg)
         assert handler.get_message_count(controller_zmq_benchmark) == 1
-        assert not handler.use_req_socket()
+        assert handler.use_req_socket()
```
KuntaiDu, line 161: Wondering why we remove the `not` here.

KuntaiDu (approved):
Otherwise LGTM

tests/tools/test_controller_zmq_benchmark.py
maobaolong: since i change to use request response type socket for heartbeat

sammshen (approved):
LGTM!

PR #2259, opened 2025-12-16
fix(s3_connector): fix memory leaks in get/batched_get

drivebyer:

**If applicable**:


drivebyer:
PTAL @sammshen @kobe0938   @YaoJiayi

sammshen:
the remote connector is not responsible for memory_count_down. the cache_engine does this after the gpu_connector transfer has been done

drivebyer:
> the remote connector is not responsible for memory_count_down. the cache_engine does this after the gpu_connector transfer has been done

IMO, if caller got `None` and has no reference to the memory_obj, the remote connector should resiponsible for calling `ref_count_down()`

sammshen (approved):
I see, thanks for the explanation, I missed that the RC down was in the Exception case

KuntaiDu (approved):
LGTM.

PR #2260, opened 2025-12-16
code readability and lock improvements in storage backend

ankit-sam:
1. Minor typos and comment description updates in storage backend.
2. Use with block for locks in place of manually acquiring and releasing locks.

### DIFF
--- a/lmcache/v1/storage_backend/abstract_backend.py
+++ b/lmcache/v1/storage_backend/abstract_backend.py
@@ -77,8 +77,9 @@ class StorageBackendInterface(metaclass=abc.ABCMeta):
         """
         An async function to put the MemoryObj into the storage backend.
 
-        :param List[CacheEngineKey] keys: The keys of the MemoryObjs.
+        :param Sequence[CacheEngineKey] keys: The keys of the MemoryObjs.
         :param List[MemoryObj] objs: The MemoryObjs to be stored.
+        :param transfer_spec: Additional specifications for the transfer.
 
         :return:  Union[List[Future], None]: A list of `Future` objects if the
         storage persistence operation is asynchronous and is successful.
@@ -135,6 +136,7 @@ class StorageBackendInterface(metaclass=abc.ABCMeta):
         """
         Check whether keys are in the storage backend.
 
+        :param str lookup_id: The unique id (e.g., request id) for the request.
         :param List[CacheEngineKey] keys: The keys of the MemoryObjs.
 
         :param bool pin: Whether to pin the keys.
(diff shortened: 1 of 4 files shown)


sammshen (approved):
this LGTM. the lock scope is covering some unecessary lines but I think this is negligble. Not sure if there was previously a concrete reason not to lock?

lmcache/v1/storage_backend/abstract_backend.py
```diff
@@ -77,8 +77,9 @@ def batched_submit_put_task(
         """
         An async function to put the MemoryObj into the storage backend.
 
-        :param List[CacheEngineKey] keys: The keys of the MemoryObjs.
+        :param Sequence[CacheEngineKey] keys: The keys of the MemoryObjs.
```
DongDongJu, line 80: Any reason changed list to sequence? for the string performance?

lmcache/v1/storage_backend/local_cpu_backend.py
```diff
@@ -244,19 +244,13 @@ def unpin(self, key: CacheEngineKey) -> bool:
             return True
 
     def remove(self, key: CacheEngineKey, force: bool = True) -> bool:
-        if force:
-            self.cpu_lock.acquire()
-        if key not in self.hot_cache:
+        with self.cpu_lock:
```
DongDongJu, line 247: Good point, But need to make sure that it is not affecting to the layerwise mode because it seems previously, we can remove the object w/o locking. @YaoJiayi, Could you confirm that this changes? it seems coming from your refactor before (https://github.com/LMCache/LMCache/commit/451501601d79181ab5d177bff08949f7459b39d3) and previously it is similar like this changes.

And maybe it serializes the async behavior with multi-core and high concurrency requests.

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -224,32 +224,20 @@ def remove(
         key: CacheEngineKey,
         force: bool = True,
     ) -> bool:
-        if force:
-            self.disk_lock.acquire()
+        path = None
```
DongDongJu, line 227: same as follows with previous comment.

DongDongJu (commented):
Hello @ankit-sam, it seems we changed the behavior that optional needed lock to must have lock in operation.
I bliv this one possibly affect to the performance in high concurrency scenario.
Need to confirm with @YaoJiayi why we reafactor into this design.

lmcache/v1/storage_backend/abstract_backend.py
ankit-sam: No actually its sequence only, the description was wrong here which I just updated.

PR #2261, opened 2025-12-16
[Controller] Support full sync controller-side and update all 

maobaolong:

**How to test**
- Start Controller and a LMCache with Worker.
- Restart Controller
- Log in Controller
```
[2025-12-16 20:39:02,125] LMCache INFO: Registered instance-worker ('lmcache_instance_1', 0) with URL 192.168.255.10:8500 (registration_controller.py:148:lmcache.v1.cache_controller.controllers.registration_controller)
INFO:     Uvicorn running on http://0.0.0.0:9999 (Press CTRL+C to quit)
[2025-12-16 20:39:02,572] LMCache INFO: Started full sync for worker ('lmcache_instance_1', 0): sync_id=lmcache_instance_1_0_98459711, expected_keys=3906, expected_batches=2 (full_sync_tracker.py:173:lmcache.v1.cache_controller.controllers.full_sync_tracker)
[2025-12-16 20:39:02,573] LMCache INFO: Accepted full sync start: worker=('lmcache_instance_1', 0), sync_id=lmcache_instance_1_0_98459711, total_keys=3906, batch_count=2 (kv_controller.py:226:lmcache.v1.cache_controller.controllers.kv_controller)
[2025-12-16 20:39:02,581] LMCache WARNING: KV batch sequence out of order detected: key=('lmcache_instance_1', 0, 'LocalCPUBackend'), expected_seq=2000, actual_seq=0, gap=-2000 (utils.py:126:lmcache.v1.cache_controller.utils)
[2025-12-16 20:39:02,582] LMCache INFO: Completed full sync for worker ('lmcache_instance_1', 0): sync_id=lmcache_instance_1_0_98459711, received_keys=3906, batches=2 (full_sync_tracker.py:281:lmcache.v1.cache_controller.controllers.full_sync_tracker)
[2025-12-16 20:39:02,582] LMCache INFO: Full sync completed for worker ('lmcache_instance_1', 0): sync_id=lmcache_instance_1_0_98459711, reported_keys=3906, keys_in_pool=3906 (kv_controller.py:326:lmcache.v1.cache_controller.controllers.kv_controller)
[2025-12-16 20:39:07,583] LMCache INFO: Full sync completion threshold reached (100.0%), disabling need_full_sync_all (full_sync_tracker.py:368:lmcache.v1.cache_controller.controllers.full_sync_tracker)

```
- LMCache worker side log
```
[2025-12-16 20:39:02,126] LMCache INFO: Executing heartbeat command: FullSyncCommand(reason=worker_re_registered, args=None) (worker.py:350:lmcache.v1.cache_controller.worker)
[2025-12-16 20:39:02,126] LMCache INFO: Received full sync command with reason: worker_re_registered, args: None (full_sync.py:35:lmcache.v1.cache_controller.commands.full_sync)
[2025-12-16 20:39:02,126] LMCache INFO: Starting full sync for worker lmcache_instance_1:0, reason: worker_re_registered (full_sync_sender.py:189:lmcache.v1.storage_backend.full_sync_sender)
[2025-12-16 20:39:02,126] LMCache INFO: Full sync startup delay: 0.44s (full_sync_sender.py:199:lmcache.v1.storage_backend.full_sync_sender)
[2025-12-16 20:39:02,565] LMCache INFO: StorageManager freeze mode set to True (storage_manager.py:750:lmcache.v1.storage_backend.storage_manager)
[2025-12-16 20:39:02,570] LMCache INFO: Full sync: total_keys=3906, batch_size=2000, batch_count=2 (full_sync_sender.py:211:lmcache.v1.storage_backend.full_sync_sender)
[2025-12-16 20:39:02,580] LMCache INFO: Full sync batches sent, total_keys=3906 (full_sync_sender.py:263:lmcache.v1.storage_backend.full_sync_sender)
[2025-12-16 20:39:07,587] LMCache INFO: Sync status: is_complete=True, global_progress=100.0%, can_exit_freeze=True (full_sync_sender.py:278:lmcache.v1.storage_backend.full_sync_sender)
[2025-12-16 20:39:07,587] LMCache INFO: Full sync completed successfully (full_sync_sender.py:295:lmcache.v1.storage_backend.full_sync_sender)
[2025-12-16 20:39:07,587] LMCache INFO: StorageManager freeze mode set to False (storage_manager.py:750:lmcache.v1.storage_backend.storage_manager)
[2025-12-16 20:39:07,587] LMCache INFO: Full sync completed successfully (full_sync.py:58:lmcache.v1.cache_controller.commands.full_sync)
```

**If applicable**:


lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -144,44 +185,131 @@ async def handle_full_sync_start(
...
-        return FullSyncStartRetMsg(sync_id=sync_id, accepted=True)
+        if not success:
+            logger.warning(
+                "Failed to start sync for worker %s: sync_id=%s", report_id, sync_id
+            )
+            return FullSyncStartRetMsg(
+                sync_id=sync_id,
+                accepted=False,
+                error_msg="Worker already syncing with different sync_id",
+            )
```
ApostaC, line 211: IIUC, there are two cases where start_sync returns false:
1. there is already a sync.
2. worker not found.

Therefore, should we change the `error_msg` here to avoid potential confusion?

lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -144,44 +185,131 @@ async def handle_full_sync_start(
...
+            return FullSyncStartRetMsg(sync_id=sync_id, accepted=True)
+        else:
+            logger.warning(
+                "Rejected full sync start: worker=%s, sync_id=%s", report_id, sync_id
+            )
+            return FullSyncStartRetMsg(
+                sync_id=sync_id,
+                accepted=False,
+                error_msg="Worker already syncing with different sync_id",
+            )
```
ApostaC: This part of the code and lines 202-210 are repetitive

lmcache/v1/cache_controller/controllers/full_sync_tracker.py
ApostaC, line 1: Quick question about `global_progress`:
When the controller starts, it will do the full sync with all the workers, and the global progress makes sense in this case. 
However, will there be any cases where full sync happens with just one worker (e.g., the worker's state becomes inconsistent with the controller's state)? If so, what's the meaning of `global_progress` in such cases?

lmcache/v1/cache_controller/utils.py
```diff
@@ -53,30 +83,52 @@ def handle_batched_kv_operations(
...
-                    if op.seq_num != expected_seq and seq_warning is None:
-                        seq_warning = (
-                            expected_seq,
-                            op.seq_num,
-                            op.seq_num - expected_seq,
-                        )
-                self.seq_tracker[location] = op.seq_num
+                # Skip sequence check during full sync
+                if not is_full_sync:
+                    # Sequence check
```
ApostaC: Can we do the following to avoid another indentation?
```python
if is_full_sync:
    continue
```

lmcache/v1/utils/cache_utils.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
ApostaC: Do we need a shebang here?

lmcache/v1/cache_controller/controllers/full_sync_tracker.py
```diff
@@ -0,0 +1,461 @@
...
+        self._need_full_sync_all = need
+        logger.info("Set need_full_sync_all to %s", need)
+
+    def _get_all_workers_cached(
+        self, timeout_seconds: Optional[float] = None
+    ) -> list[tuple[str, WorkerNode]]:
+        """Get all registered workers from the registry tree."""
+        return self.registry_tree.get_all_worker_nodes_cached(timeout_seconds)
+
+    def should_request_full_sync(
```
ApostaC, line 82: Is this function being called anywhere?

ApostaC (commented):
Got a few comments. Otherwise LGTM!

lmcache/v1/cache_controller/controllers/full_sync_tracker.py
maobaolong: Clipboard_Screenshot_1766577789

Here is the caller.

lmcache/v1/utils/cache_utils.py
maobaolong: No need, removed.

lmcache/v1/cache_controller/utils.py
maobaolong: Done

lmcache/v1/cache_controller/controllers/kv_controller.py
maobaolong: Done

lmcache/v1/cache_controller/controllers/full_sync_tracker.py
maobaolong: Great question!

Yes, there are some reasons will trigger one worker or part of workers full sync, and `global_progress` act as a service safety guard, if there are many workers execute full sync, it means that the whole controller/cluster unhealthy, all the worker should get into the freeze mode. Not a special case, if there is exact one worker execute full sync, since the `global_progress` = `full_sync_worker_num / len(all_workers)`, so `global_progress >= self.completion_threshold` is true, all the other worker should stay unfrozen mode.

maobaolong:
@ApostaC Would you like to take another look? Thanks!

lmcache/v1/cache_controller/controllers/full_sync_tracker.py
ApostaC: Just wanted to double check: let's say I have 10 workers, and only one of them wants to due a full sync because that worker has some failures. In this case, what will be the `full_sync_worker_num`? Is it 10 or 1?
If it's 1, then the global progress will always be `0.1` and will be less than the `completion_threshold` and as a result cause that worker to freeze forever. 

Maybe I have some misunderstanding here?

lmcache/v1/cache_controller/controllers/kv_controller.py
ApostaC: Hey Baolong @maobaolong, a follow-up question on this:
IIUC, the `success` flag hasn't been changed since line 199. We did a check on `success` between lines 202-210, sothe `else` branch between lines 248-256 will never get executed?

ApostaC:
@maobaolong Thanks! Please see my new messages:
- https://github.com/LMCache/LMCache/pull/2261#discussion_r2650159671
- https://github.com/LMCache/LMCache/pull/2261#discussion_r2650159671

lmcache/v1/cache_controller/controllers/full_sync_tracker.py
maobaolong: Updated the calculate logic.

lmcache/v1/cache_controller/controllers/kv_controller.py
maobaolong: Remove this dead code branch.

maobaolong:
@ApostaC If you have time, please help to take another look, thanks!

lmcache/v1/cache_controller/controllers/kv_controller.py
```diff
@@ -144,22 +185,65 @@ async def handle_full_sync_start(
...
+                    old_count,
+                    report_id,
+                    msg.location,
+                )
+            else:
+                logger.warning(
+                    "Failed to clear keys for worker %s location %s",
+                    report_id,
+                    msg.location,
+                )
```
ApostaC, line 237: Does this mean we are still going to trigger full sync even if we failed to clear the existing keys? Just wanted to understand if it's the expected behavior here.

ApostaC (commented):
Mostly LGTM! One question: where do we use `can_exit_freeze`? Is that being used in the worker process? Maybe I didn't find it.

maobaolong:
> can_exit_freeze

Clipboard_Screenshot_1767490072


@ApostaC  It is a little hard to find it indeed, the worker process will handle `FullSyncCommand` once get the heartbeat return message with  `FullSyncCommand` and handle it, the last step of it is that query status from controller and decide whether permit to exit freeze mode.

lmcache/v1/cache_controller/controllers/kv_controller.py
maobaolong: Yeah, it shouldn't fail to clear kv while full sync starting, so I left a warning log, but if it happened, it still should not break the full sync flow, WDYT?

ApostaC: I think it's okay. Just wanted to have a better understanding here, when doing full sync, do we need to make sure the old states in the controller are fully cleared? (Basically, what will happen if the controller processed the full sync batches with some un-cleared states at the beginning?)

maobaolong: Good question! Let me clarify the behavior:
When clear fails but full sync proceeds, the bad case is `keys that were removed from the worker but not cleared from controller will remain as stale entries`.

Impact of stale entries:
- Lookup may return a stale location, but the actual fetch from worker would fail/miss
- This is a degraded but not broken state.

If we want be stricter, we can cancel full sync and marked sync failed for this worker and devops engineer​ will aware this failure and do something to recover health state then will trigger a full sync after.

ApostaC (approved):
LGTM!

chunxiaozheng (approved):
lgtm

PR #2263, opened 2025-12-17
[correctness]: fix early alignment of num_computed_tokens

sammshen:
Correctness issues introduced from https://github.com/LMCache/LMCache/pull/2111

Follow-up: PR that will ensure APC + LMCache compatibility

Testing manually: 
```bash
VLLM_SERVER_DEV_MODE=1 \
VLLM_BATCH_INVARIANT=1 \
VLLM_ATTENTION_BACKEND=FLASH_ATTN \
vllm serve meta-llama/Llama-3.2-1B-Instruct \
  --kv-transfer-config \
```

```bash
CONTEXT="$(
  man bash \
  | col -b \
  | tr -s '[:space:]' ' ' \
  | awk '{
      for (i = 1; i <= NF; i++) {
        printf "%s ", $i
        if (++c == 5000) exit
      }
    }'
)"

HALF_CONTEXT="$(
  man bash \
  | col -b \
  | tr -s '[:space:]' ' ' \
  | awk '{
      for (i = 1; i <= NF; i++) {
        printf "%s ", $i
        if (++c == 2500) exit
      }
    }'
)"

# 1. send curl request with CONTEXT (populates APC and LMCache)
  -d "$(jq -n \
    --arg model "meta-llama/Llama-3.2-1B-Instruct" \
    --arg content "$CONTEXT" \
    --argjson max_tokens 100 \
    '{
      model: $model,
      temperature: 0,
      max_tokens: $max_tokens,
      messages: [
        { role: "user", content: $content }
      ]
    }'
  )" | jq -r '.choices[0].message.content'
  
# 2. send clear APC request
curl -X POST "http://localhost:8000/reset_prefix_cache"

# 3. send curl request with HALF_CONTEXT (will populate APC)
  -d "$(jq -n \
    --arg model "meta-llama/Llama-3.2-1B-Instruct" \
    --arg content "$HALF_CONTEXT" \
    --argjson max_tokens 100 \
    '{
      model: $model,
      temperature: 0,
      max_tokens: $max_tokens,
      messages: [
        { role: "user", content: $content }
      ]
    }'
  )" | jq -r '.choices[0].message.content'
  
# 4. send curl request with CONTEXT (will half hit APC and half hit LMCache)
  -d "$(jq -n \
    --arg model "meta-llama/Llama-3.2-1B-Instruct" \
    --arg content "$CONTEXT" \
    --argjson max_tokens 100 \
    '{
      model: $model,
      temperature: 0,
      max_tokens: $max_tokens,
      messages: [
        { role: "user", content: $content }
      ]
    }'
  )" | jq -r '.choices[0].message.content'
```

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1518,14 +1518,6 @@ class LMCacheConnectorV1Impl:
             logger.debug(f"Looking up cache for the first time for request {req_id}!")
             self._requests_priority[req_id] = getattr(request, "priority", 0)
 
-            # Align computed tokens once to avoid repeated
-            # chunk-size rounding downstream
-            aligned_num_computed_tokens = (
-                num_computed_tokens
-                // self._lmcache_chunk_size
-                * self._lmcache_chunk_size
-            )
-
             # token_ids = request.prompt_token_ids
             # all token ids covers the preemption case
             token_ids = request.all_token_ids
@@ -1542,11 +1534,21 @@ class LMCacheConnectorV1Impl:
             if self.skip_last_n_tokens > 0:
                 token_ids = token_ids[: -self.skip_last_n_tokens]
 
+            # this rounds down
+            # example diagram: vllm page size = 16, lmcache chunk size = 256 (brackets)
+            # [ vllm page 1 | ... | vllm page 16 ] [ vllm page 17 | vllm page 18 ]
+            # lmcache chunk 1                                     lmcache chunk 2
+            # total tokens = 512 (2 lmcache chunks)
+            # vllm num_computed_tokens = 288 (18 vllm pages)
+            # we can completely skip lmcache chunk 1 but we will still retrieve chunk 2
+            # and overwrite vllm page 17 and 18
+            # the need_to_allocate is 512 - 288 = 224
+            # and not 256 even though we will retrieve 256 tokens from lmcache
             num_external_hit_tokens = self.lookup_client.lookup(
                 token_ids,
                 lookup_id=req_id,
                 request_configs=request_configs,
-                num_computed_tokens=aligned_num_computed_tokens,
+                num_computed_tokens=num_computed_tokens,
             )
 
         if num_external_hit_tokens is None:
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1035,9 +1035,10 @@ class LMCacheConnectorV1Impl:
 
                 # Check the result
                 num_retrieved_tokens = ret_token_mask.sum().item()
-                num_expected_tokens = (
-                    lmcache_cached_tokens - request.load_spec.vllm_cached_tokens
-                )
+                # use token_mask to calculate expected tokens, matching what retrieve()
+                # calculates internally as num_required_tokens
+                token_mask_slice = token_mask[:lmcache_cached_tokens]
+                num_expected_tokens = token_mask_slice.sum().item()
                 if num_retrieved_tokens < num_expected_tokens:
                     logger.error(
                         "Request %s"
@@ -1060,7 +1061,6 @@ class LMCacheConnectorV1Impl:
                         slot_mapping[:lmcache_cached_tokens],
                     )
                     self._invalid_block_ids.update(missing_blocks)
-
             self._stats_monitor.update_interval_vllm_hit_tokens(
                 request.load_spec.vllm_cached_tokens
             )
@@ -1109,7 +1109,6 @@ class LMCacheConnectorV1Impl:
         ret_mask_cpu = ret_mask.to(device="cpu", dtype=torch.bool)
 
         if ret_mask_cpu.shape[0] != expected_mask_cpu.shape[0]:
-            logger.debug("expected_mask_cpu.shape[0] != ret_mask_cpu.shape[0]")
             return set()
 
         missing_mask = expected_mask_cpu & ~ret_mask_cpu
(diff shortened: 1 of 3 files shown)


sammshen:
I think we should consider reverting: 
https://github.com/LMCache/LMCache/pull/2111

the problem is that we should be pinning the vllm computed tokens during lookup which we are not doing right now. this is causing serious correctness issue in the preemption case

sammshen:
to summarize the issues introduced by https://github.com/LMCache/LMCache/pull/2111
- first we return the wrong number in `get_num_new_matched_tokens`
- then we are failing to pin the prefix cached tokens

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1109,7 +1109,6 @@ def record_failed_blocks(
         ret_mask_cpu = ret_mask.to(device="cpu", dtype=torch.bool)
 
         if ret_mask_cpu.shape[0] != expected_mask_cpu.shape[0]:
-            logger.debug("expected_mask_cpu.shape[0] != ret_mask_cpu.shape[0]")
```
DongDongJu, line 1112: why it removed?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1542,11 +1533,20 @@ def get_num_new_matched_tokens(
...
 
+            # example diagram: vllm page size = 16, lmcache chunk size = 256 (brackets)
+            # [ vllm page 1 | ... | vllm page 16 ] [ vllm page 17 | vllm page 18 ]
+            # lmcache chunk 1                                     lmcache chunk 2
+            # total tokens = 512 (2 lmcache chunks)
+            # vllm num_computed_tokens = 288 (18 vllm pages)
+            # we can completely skip lmcache chunk 1 but we will still retrieve chunk 2
+            # and overwrite vllm page 17 and 18
+            # the need_to_allocate is 512 - 288 = 224
+            # and not 256 even though we will retrieve 256 tokens from lmcache
```
DongDongJu: How about describe like this?
```
# Example (page_size=16, chunk_size=256):
#
# chunks:  [0..255]                [256..511]
# pages:   [0..15]...[240..255]    [256..271][272..287] ...
#
# num_computed_tokens = 288 => vLLM already has [0..287] (18 pages)
# LMCache hit_prefix_tokens = 512 => cache covers [0..511] (2 chunks)
#
# We skip chunk 1, retrieve chunk 2, and overwrite [256..287] (32-token overlap).
# Allocate only the truly new tail: 512 - 288 = 224 tokens.
```

lmcache/v1/cache_engine.py
DongDongJu, line 926: I think gemini confused caused by this docstring.
Could you update this docstring too?

DongDongJu (changes requested):
Hello @sammshen, Thanks for what a fast fix.
I left few comments. otherwise looks promising.
Can we make ut in this case for the future? like overlap scenario.

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: this is great, will take this!

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: adding it back, was causing some errors during testing. thanks for the catch

lmcache/integration/vllm/vllm_v1_adapter.py
ziruiliu: I am not sure why there is `res = aligned_res - overlap`.
In this example, the expected result from cache_engine's lookup() should be 512, which is size of 2 chunks. If adding `res = aligned_res - overlap`,  aligned_res will be 512 and res will be 480?

ziruiliu: My understanding is similar with Gemini's. Please correct me/us where the calculation goes wrong

chunxiaozheng:
> to summarize the issues introduced by #2111
> 
> * first we return the wrong number in `get_num_new_matched_tokens`
> * then we are failing to pin the prefix cached tokens

@sammshen Does the second point refer to the `pins` in VLLM prefix cache?

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: you're right, the overlap should be handled in the adapter instead of the cache engine

sammshen: after some more thought, I have a couple of realizations: 
1. we don't need alignment at all: 
```python
end <= aligned_computed_tokens:
```
statements such as the above will still hold even if we don't align to lmcache chunk size
2. due to the need to still pin the vllm cached tokens, I am considering reverting the skip vllm contains optimization. because for example if a remote backend implements PIN, we still need to call contains in the storage manager to trigger this. alternatively, we can decouple pin and contains

sammshen: I'm going to revert the PR. 

We have three choices: 
1. revert this optimization. reasoning: in the preemption case, when vllm number of computed tokens changes between lookup and retrieve, there are serious correctness issues. motivation: lookup is relatively cheap and we already optimized lookup to only trigger once
2. finish this optimization. difficulty: forces us to decouple pin and contains. this is dangerous because under a single acquisition of the lock, we need to check whether a key exists AND pin. decoupling is extremely difficult
3. no longer load under the preemption case.

sammshen:
reverting with https://github.com/LMCache/LMCache/pull/2317

lmcache/integration/vllm/vllm_v1_adapter.py
ziruiliu: just notice this message. agree with reverting the optimization

PR #2266, opened 2025-12-17
[feature][core] support dsa in RemoteBackend

chunxiaozheng:
support dsa in `RemoteBackend`

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -751,9 +751,15 @@ class LMCacheConnectorV1Impl:
             )
 
             # In case of MLA, the lookup server is only created on worker 0
-            if self.async_loading and self.lookup_server is not None:
+            if (
+                self.async_loading
+                and self.lookup_server is not None
+                and self.lmcache_engine.storage_manager is not None
+            ):
                 assert isinstance(self.lookup_server, LMCacheAsyncLookupServer)
-                self.lmcache_engine.post_init(async_lookup_server=self.lookup_server)
+                self.lmcache_engine.storage_manager.set_async_lookup_server(
+                    self.lookup_server
+                )
 
         self.kv_caches: dict[str, torch.Tensor] = {}
 
@@ -949,10 +955,13 @@ class LMCacheConnectorV1Impl:
             The number of elements in kv_caches and layer_names should be
             the same.
         """
+        assert self.lmcache_engine is not None
         self.current_layer = 0
 
         if len(self.kv_caches) == 0:
+            logger.info("Initializing kv_caches")
             self._init_kv_caches_from_forward_context(forward_context)
+            self.lmcache_engine.post_init(kvcaches=list(self.kv_caches.values()))
 
         metadata = self._parent._get_connector_metadata()
         assert isinstance(metadata, LMCacheConnectorMetadata)
@@ -964,10 +973,6 @@ class LMCacheConnectorV1Impl:
         if attn_metadata is None:
             logger.debug("In connector.start_load_kv, but the attn_metadata is None")
             return
-
-        assert self.lmcache_engine is not None
-
-        self.lmcache_engine.post_init(kvcaches=kvcaches)
 
         self.layerwise_retrievers = []
 
--- a/lmcache/v1/protocol.py
+++ b/lmcache/v1/protocol.py
@@ -9,8 +9,11 @@ import struct
 import torch
 
 # First Party
+from lmcache.logging import init_logger
 from lmcache.utils import CacheEngineKey, LayerCacheEngineKey, parse_cache_key
 from lmcache.v1.memory_management import MemoryFormat
+
+logger = init_logger(__name__)
 
 MAX_KEY_LENGTH = 150
 
@@ -71,53 +74,50 @@ INT_TO_LOCATION = {
 @dataclass
 class RemoteMetadata:
     length: int
-    shape: torch.Size
-    dtype: Optional[torch.dtype]
+    shapes: list[torch.Size]
+    dtypes: list[torch.dtype]
     fmt: MemoryFormat
 
-    def serialize_into(self, buffer):
-        assert len(self.shape) == 4, "Shape dimension should be 4"
+    def _prepare_params(self):
+        params = [self.length, int(self.fmt.value)]
+        for shape, dtype in zip(self.shapes, self.dtypes, strict=True):
+            assert len(shape) == 4, "Shape dimension should be 4"
+            params.append(DTYPE_TO_INT[dtype])
+            params.append(shape[0])
+            params.append(shape[1])
+            params.append(shape[2])
+            params.append(shape[3])
+        return params
 
-        struct.pack_into(
-            "iiiiiii",
-            buffer,
-            0,
-            self.length,
-            int(self.fmt.value),
-            DTYPE_TO_INT[self.dtype],
-            self.shape[0],
-            self.shape[1],
-            self.shape[2],
-            self.shape[3],
-        )
+    def serialize_into(self, buffer, fmt: str):
+        params = self._prepare_params()
+        assert len(fmt) == len(params)
+        struct.pack_into(fmt, buffer, 0, *params)
 
-    def serialize(self) -> bytes:
-        # NOTE(Jiayi): 4 is the maximum dimension of memory object.
-        # Pass in shape [x, 0, 0, 0] if it is a bytes memory object
-        assert len(self.shape) == 4, "Shape dimension should be 4"
-
-        packed_bytes = struct.pack(
-            "iiiiiii",
-            self.length,
-            int(self.fmt.value),
-            DTYPE_TO_INT[self.dtype],
-            self.shape[0],
-            self.shape[1],
-            self.shape[2],
-            self.shape[3],
-        )
+    def serialize(self, fmt: str) -> bytes:
+        params = self._prepare_params()
+        assert len(fmt) == len(params)
+        packed_bytes = struct.pack(fmt, *params)
         return packed_bytes
 
     @staticmethod
-    def deserialize(s: bytes) -> "RemoteMetadata":
-        length, fmt, dtype, shape0, shape1, shape2, shape3 = struct.unpack_from(
-            "iiiiiii", s
-        )
+    def deserialize(s: bytes, fmt: str) -> "RemoteMetadata":
+        # length, fmt, dtype, shape0, shape1, shape2, shape3
+        result = struct.unpack_from(fmt, s)
+        assert len(fmt) == len(result)
+        length = result[0]
+        memory_fmt = MemoryFormat(result[1])
+        shapes = []
+        dtypes = []
+        for i in range(2, len(result), 5):
+            shapes.append(torch.Size(result[i + 1 : i + 5]))
+            dtypes.append(INT_TO_DTYPE[result[i]])
+
         return RemoteMetadata(
             length,
-            torch.Size([shape0, shape1, shape2, shape3]),
-            INT_TO_DTYPE[dtype],
-            MemoryFormat(fmt),
+            shapes,
+            dtypes,
+            memory_fmt,
         )
 
 
--- a/lmcache/v1/storage_backend/abstract_backend.py
+++ b/lmcache/v1/storage_backend/abstract_backend.py
@@ -43,6 +43,10 @@ class StorageBackendInterface(metaclass=abc.ABCMeta):
 
         self.dst_device = dst_device
 
+    def post_init(self):  # noqa: B027
+        """Post initialize the storage backend."""
+        pass
+
     @abc.abstractmethod
     def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
         """
--- a/lmcache/v1/storage_backend/connector/base_connector.py
+++ b/lmcache/v1/storage_backend/connector/base_connector.py
@@ -32,54 +32,49 @@ class RemoteConnector(metaclass=abc.ABCMeta):
     Interface for remote connector
     """
 
-    save_chunk_meta: bool = True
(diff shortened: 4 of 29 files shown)


ApostaC:
Thanks @chunxiaozheng ! This is a relatively large change. I would prefer to have some comprehensive tests that enable `use_gpu_conennctor_v3` to make sure that it won't break.
Probably we can do it in a separate PR and merge that first?

chunxiaozheng:
> Thanks @chunxiaozheng ! This is a relatively large change. I would prefer to have some comprehensive tests that enable `use_gpu_conennctor_v3` to make sure that it won't break. Probably we can do it in a separate PR and merge that first?

@ApostaC Sure, I will add some comprehensive test in another PR.（#2279）

chunxiaozheng:
@ApostaC @sammshen could you help take a look?

lmcache/v1/storage_backend/abstract_backend.py
```diff
@@ -43,6 +43,10 @@ def __init__(
 
         self.dst_device = dst_device
 
+    def post_init(self):  # noqa: B027
+        """Post initialize the storage backend."""
```
ApostaC, line 47: Can we add a more detailed clarification here?

If I understand correctly, the `post_init` itself does not take any arguments. However, it assumes a workflow that "some other modules update the global LMCacheEngineMetadata", and the `post_init` functions use the updated KV cache group information to do some extra initialization. 

It's important to make the above workflow clear in the docstring so that other people can easily understand and maintain this part of the code.

ApostaC: In the meantime, I realized that although we are passing the metadata over all the modules, there is no explicit guarantee to make sure that the metadata object is **_singular_**.

In this case, the whole process might be broken if other contributors (who are not familiar with this part of the code) change the metadata's mechanism.

ApostaC:
@chunxiaozheng Thanks for this great effort! I'm reviewing this PR but it may take some time since it's a big one.

In the meantime, I want to quickly check one thing: now we do `post_init` after we see the shape of KV cache. I'm wondering will the workflow be much simpler if we can get the KV cache shape information before the vLLM worker initializes the KV cache (e.g., from vLLM config)? 
@KuntaiDu I remember you have some knowledge about vLLM's KV cache config. Can you give us some insight here?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -949,10 +955,13 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
             The number of elements in kv_caches and layer_names should be
             the same.
         """
+        assert self.lmcache_engine is not None
         self.current_layer = 0
 
         if len(self.kv_caches) == 0:
+            logger.info("Initializing kv_caches")
             self._init_kv_caches_from_forward_context(forward_context)
+            self.lmcache_engine.post_init(kvcaches=list(self.kv_caches.values()))
```
ApostaC, line 964: If we only pass `kvcaches` into `LMCacheEngine.post_init`, should we make it an explicit argument rather than `kwargs`?

lmcache/v1/protocol.py
```diff
@@ -71,53 +74,50 @@ class ServerReturnCode(IntEnum):
...
-            DTYPE_TO_INT[self.dtype],
-            self.shape[0],
-            self.shape[1],
-            self.shape[2],
-            self.shape[3],
-        )
+    def _prepare_params(self):
+        params = [self.length, int(self.fmt.value)]
+        for shape, dtype in zip(self.shapes, self.dtypes, strict=True):
+            assert len(shape) == 4, "Shape dimension should be 4"
```
ApostaC, line 84: Is it always true? 
An alternative is to use `msgspec` to do the encoding and decoding. It might be easier than using `struct` and can give more flexibility.

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -32,54 +32,49 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
+            "full chunk size: %s, single token size: %s, "
+            "remote metadata fmt: %s, bytes: %s",
+            self.save_chunk_meta,
+            self.meta_shapes,
+            self.meta_dtypes,
+            self.meta_fmt,
+            self.full_chunk_size,
+            self.single_token_size,
+            self.remote_metadata_fmt,
+            self.remote_metadata_bytes,
```
ApostaC, line 77: The following variables are not used in this file. Should we consider removing them?
- `self.save_chunk_meta`
- `self.meta_fmt`
- `self.remote_metadata_fmt`
- `self.remote_metadata_bytes`

Additionally, `meta_shapes` and `meta_dtypes` are only used within this function. Do we need to make them the field of the BaseConnector class?

ApostaC: Oh, I see some of them are used in the sub-classes. 
In this case, is there a way to explicitly mark the existence of the fields in the base class?

lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -535,27 +539,30 @@ async def _batched_put(
...
             # Get memory object data
+            # TODO(chunxiaozheng): support dsa
             kv_bytes = memory_obj.byte_array
             kv_tensor = memory_obj.tensor
-            kv_shape = memory_obj.get_shape()
-            kv_dtype = memory_obj.get_dtype()
+            kv_shapes = memory_obj.get_shapes()
+            kv_dtypes = memory_obj.get_dtypes()
+            assert len(kv_shapes) == 1
+            assert len(kv_dtypes) == 1
```
ApostaC, line 548: Can we add a string saying something like `EICConnector does not support more than 1 kv groups" for better clarity?

lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -535,27 +539,30 @@ async def _batched_put(
...
+            meta_bytes = remote_meta.serialize(self.remote_metadata_fmt)
             meta_list.append(meta_bytes)
             meta_ptr = self.bytes_get_ptr(meta_bytes)
             meta_size = len(meta_bytes)
             data_ptr = kv_tensor.data_ptr()
             data_size = len(kv_bytes)
 
             logger.info(
-                f"eic batched_put {key_str} shape {kv_shape} dtype {kv_dtype}"
+                f"eic batched_put {key_str} shapes {kv_shapes} dtypes {kv_dtypes}"
```
ApostaC, line 565: nit: we can %s rather than f-strings here.

lmcache/v1/storage_backend/connector/infinistore_connector.py
ApostaC, line 1: Do we need to call `post_init` anywhere in this class?

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
```diff
@@ -329,11 +309,11 @@ async def _batch_get_into(
...
         This is used when save_chunk_meta=False (metadata not stored remotely).
         """
-        if not self.meta_shape or not self.meta_dtype or not self.meta_fmt:
+        if not self.meta_shapes or not self.meta_dtypes or not self.meta_fmt:
             logger.error(
                 f"Metadata required for batch_get_into but not available: "
-                f"meta_shape={self.meta_shape}, "
-                f"meta_dtype={self.meta_dtype}, "
+                f"meta_shapes={self.meta_shapes}, "
+                f"meta_dtypes={self.meta_dtypes}, "
```
ApostaC, line 316: nit: let's change to use "%s" rather than f-string

lmcache/v1/storage_backend/connector/mooncakestore_connector.py
ApostaC, line 1: Double check on updating `post_init` for this class

lmcache/v1/storage_backend/connector/redis_connector.py
ApostaC, line 1: Double check on updating `post_init` for this class

lmcache/v1/storage_backend/connector/s3_connector.py
```diff
@@ -164,17 +166,22 @@ def __init__(
...
         if self.s3_part_size is None:
             # Default to chunk size
             self.s3_part_size = self.full_chunk_size
         assert self.s3_part_size == self.full_chunk_size, (
             "S3 part size must be equal to chunk size in S3Connector"
         )
-        logger.info(f"s3 connector meta_shape: {self.meta_shape}")
-        logger.info(f"s3 connector meta_dtype: {self.meta_dtype}")
+        logger.info(f"s3 connector meta_shapes: {self.meta_shapes}")
+        logger.info(f"s3 connector meta_dtypes: {self.meta_dtypes}")
```
ApostaC, line 184: nit: use %s to replace f-string.

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
```diff
@@ -458,31 +464,33 @@ def _read_from_shared_memory(
...
 
             # Parse metadata
-            metadata = RemoteMetadata.deserialize(header)
+            metadata = RemoteMetadata.deserialize(header, self.remote_metadata_fmt)
             if metadata.length <= 0:
                 logger.error(f"Invalid payload length: {metadata.length}")
                 return None
 
+            assert len(metadata.shapes) == 1
+            assert len(metadata.dtypes) == 1
```
ApostaC, line 486: nit: add a string to explain the assertion error here.

lmcache/v1/storage_backend/connector/sagemaker_hyperpod_connector.py
```diff
@@ -825,19 +833,24 @@ def _build_put_stream(self, memory_obj: MemoryObj) -> Tuple[int, AsyncIterator]:
         kv_len = len(kv_view)
 
         # Prepare metadata
-        shape = list(memory_obj.get_shape())
+        # TODO(chunxiaozheng): support dsa
+        shapes = memory_obj.get_shapes()
+        dtypes = memory_obj.get_dtypes()
+        assert len(shapes) == 1
+        assert len(dtypes) == 1
```
ApostaC, line 840: nit: add a string to explain the assertion error here.

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -296,13 +301,9 @@ def _setup_metrics(self):
                 )
             )
 
-    def post_init(self, **kwargs) -> None:
-        if "async_lookup_server" in kwargs:
-            self.async_lookup_server = kwargs.pop("async_lookup_server")
```
ApostaC, line 301: Just wanted to double check, the `async_look_server` is not used anymore in the code, right?

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -271,6 +271,11 @@ def __init__(
         self._freeze = False
         self._freeze_lock = threading.RLock()
 
+        # PDBackend hasn't supported calculate_chunk_budget
+        if not self.enable_pd and self.config.enable_async_loading:
+            assert self.allocator_backend is not None
+            self.async_serializer = AsyncSingleSerializer(self.loop)
```
ApostaC, line 277: @sammshen We need to check whether this change impacts the async loading.

ApostaC (commented):
Did a pass. Mostly LGTM! Almost all the comments are about nit and clarity.

ApostaC:
@sammshen @Shaoting-Feng Can you help with some follow-up offline tests after this PR is merged? It should at least cover async-offloading and S3 connector. Thanks!

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: agreed and also a comment in post_init doc string on when kvcaches is passed in

sammshen: could we add clear doc strings to the responsibilites of cache engine init() and post_init() as well as for storage manager?

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -32,54 +32,49 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
-    save_chunk_meta: bool = True
-    meta_shape: Optional[torch.Size] = None
-    meta_dtype: Optional[torch.dtype] = None
-    meta_fmt: Optional[MemoryFormat] = None
-    full_chunk_size: Optional[int] = None
-    single_token_size: Optional[int] = None
-
     @NotAudit
-    def init_chunk_meta(
+    def post_init(
```
sammshen, line 36: same comment as from https://github.com/LMCache/LMCache/pull/2314/files#diff-d2ecc263bdac26c42fabd0bf879f8858c51fd43b1ce62f47d43443c90ee18a16

and as above. could we clearly define what init() and post_init() are responsible for?

sammshen: btw what's the order of merging

lmcache/v1/storage_backend/storage_manager.py
sammshen: it looks like it is being set in a new `set_async_lookup_server` helper

lmcache/v1/storage_backend/storage_manager.py
sammshen: this looks like it just changed the location of initialization. you are right that we should add some async tests though. I am thinking of a high concurrency workload to compare the degradation / check for deadlocks

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: @sammshen Thanks for your review, I split this PR into #2312, #2314, #2315, #2316, and I will close current PR.

Issue #2137, opened 2025-12-02
[BUG] Layerwise mode incompatible with remote offloading


crystalxyz:
**Describe the bug**
After applying some temporary fixes at #2135 and #2136, the layer mode is still not working correctly and raises errors when storing parameters.

Glad to provide more context to this problem if needed!

**To Reproduce**
1. Start LMCache server
2. Start vLLM with LMCache config (use_layerwise=True, remote_backend="...", remote_serde="cachegen")
3. Send a request
```
RuntimeError: permute(sparse_coo): number of dimensions in the tensor input does not match the length of the desired ordering of dimensions i.e. input.dim() = 4 is not equal to len(dims)
```

**Expected behavior**
vLLM should serve requests correctly.

**Desktop (please complete the following information):**
 - OS: Linux
 - Version: VLLM v0.11.0, LMCache latest `dev` branch

**Additional context**

crystalxyz:
Attempted to patch it but turned out to be a big fix. I'm dropping some findings during my investigation, but please feel free to point me towards the right direction if I'm missing anything.
- There is no proper deserialization from the remote backend for layerwise mode, so we need to call deserialization method per-layer per-chunk somewhere.
- Tensor shape mismatch occurred in multiple locations of the codenase, so we might need a more systematic way to fix it cleanly.

sammshen:
Thanks for the issue, I will bring it up in our community meeting

hickeyma:
@sammshen I will start looking into this bug.

DongDongJu:
even local mode, I had problem for layerwise option, So I made fix for it(https://github.com/LMCache/LMCache/pull/2151).

DongDongJu:
I will update this issue today.

DongDongJu:
it still happened like https://github.com/LMCache/LMCache/issues/2135. I will check https://github.com/LMCache/LMCache/issues/2135 first and moving to here.

Issue #2135, opened 2025-12-02
[BUG] Layerwise mode format mismatch with vLLM CacheEngineKey


crystalxyz:
**Describe the bug**
When starting vLLM server with LMCache layerwise mode, the LMCache process crashes on a mismatched key format.

**To Reproduce**
1. Start LMCache server using `lmcache-server localhost 123`
2. Start vLLM server (version 0.11.0) and connect to LMCache
```
chunk_size: 256
local_cpu: true
use_layerwise: true
remote_url: "lm://localhost:123"
remote_serde: "cachegen"
```
3. Send a request to vLLM
4. LMCache server will crash with error message

```
Exception in thread Thread-1 (handle_client):
Traceback (most recent call last):
    raise ValueError(f"Invalid key string: {s}")
ValueError: Invalid key string: vllm@Qwen/Qwen3-4B@1@0@-5d1b8de6da1b6d09@bfloat16@0
```

**Expected behavior**
vLLM server should start correctly and serve requests.

**Desktop (please complete the following information):**
 - OS: Linux

**Additional context**

DongDongJu:
I confirmed that it still happen when I tested with multi_round_qa.py  
```
python3 multi-round-qa.py --num-users 10 --num-rounds 5 --qps 0.3 --shared-system-prompt 1000 --user-history-prompt 2000 --answer-len 100 --model Qwen/Qwen3-0.6B --base-url http://localhost:8000/v1 --sharegpt
```
within `vllm==0.11.2` and latest dev lmcache dev branch.
I will take a look

lmcache_server
```
[2025-12-16 16:27:08,251] LMCache INFO: Initializing cpu-only cache server (__init__.py:14:lmcache.v1.server.storage_backend)
[2025-12-16 16:27:08,251] LMCache INFO: Server started at localhost:8080 (__main__.py:138:lmcache.v1.server.__main__)
[2025-12-16 16:27:57,994] LMCache INFO: Connected by ('127.0.0.1', 39578) (__main__.py:142:lmcache.v1.server.__main__)
[2025-12-16 16:27:57,995] LMCache DEBUG: Waiting for command (__main__.py:46:lmcache.v1.server.__main__)
[2025-12-16 16:35:40,436] LMCache INFO: Client disconnected (__main__.py:134:lmcache.v1.server.__main__)
Exception in thread Thread-1 (handle_client):
Traceback (most recent call last):
[...]

PR #2268, opened 2025-12-17
[Bugfix] vllm layerwise save stopiteration

DongDongJu:
bug fix for https://github.com/LMCache/LMCache/issues/2135 and https://github.com/LMCache/LMCache/issues/2137
For https://github.com/LMCache/LMCache/issues/2135, motivated by @crystalxyz's fix but little bit different.

In the vLLM v1 integration, LMCache performs layerwise KV saving using a Python generator returned by LMCacheEngine.store_layer().
vLLM pumps this generator during attention execution (save_kv_layer) and finalizes it at the end of a 
forward step (wait_for_save).
                                                         
Under some scheduling / concurrency patterns, vLLM can call wait_for_save() when there is
nothing to save (e.g., empty connector metadata) or can call it multiple times.
In the old implementation, LMCache's worker-side adapter could call next() on an already exhausted 
store_layer generator, raising StopIteration. 
vLLM treats that exception as fatal and kills EngineCore (EngineDeadError). 
In generator context-manager paths, Python may also convert it into RuntimeError, making it even more disruptive.

What exactly crashed
--------------------
The crash chain looked like:

- vLLM worker forward exits
- kv_connector.wait_for_save() is invoked
- LMCache vLLM adapter calls next(layerwise_storer)
- layerwise_storer is already exhausted -> StopIteration
- vLLM propagates the exception as fatal -> EngineCore terminates

Why it did NOT happen often under low concurrency
------------------------------------------------
This bug requires a specific sequence of events that is less common with low concurrency:

1) A previous forward step creates layerwise storers (store_layer generators) and advances                         
   them.

2) A later forward step has nothing to save, for example:
   - connector metadata becomes empty (requests=[]), or
   - all requests are not savable (can_save=False), or
   - all tokens are effectively skipped (skip_leading_tokens causes a full skip), or
   - the step is a warmup/compile/graph path where save_kv_layer is not triggered.
   
3) Even in step (2), vLLM still calls wait_for_save() at forward exit.

If the adapter still holds stale layerwise storers from step (1), then calling next() again
hits StopIteration. 

With low concurrency, batches tend to be more stable: many steps still have at least one
savable request, and the save/finalize pattern is more consistent. As concurrency increases,
more no-save steps occur (due to scheduling, batching, prefix caching, chunked prefill,
compile/warmup behavior, and request composition). This makes the required sequence more
frequent and exposes the bug.

Root cause (fundamental issue)
------------------------------
The fundamental issue is a contract mismatch / fragility around using a generator as a
multi-stage async pipeline:

- LMCache’s layerwise save is a multi-step generator.
- vLLM calls into the adapter in a pattern that vary by step.
- The old adapter assumed that calling next() once in wait_for_save() is always valid.
  This is not safe if:
  - wait_for_save() is called when there are no new storers,
  - wait_for_save() is called multiple times,
  - the generator finishes earlier than expected for any reason.

Additionally, LMCacheEngine.store_layer() had a freeze-mode branch that yielded a different
number of steps than the normal path, which could make expected number of next() calls
assumptions even more brittle.

Fix / solution approach
-----------------------
The fix makes layerwise saving robust and idempotent.

1) Drain generators to completion in wait_for_save()
   - Instead of calling next() exactly once, we keep calling next() until StopIteration.
   - This guarantees all remaining pipeline stages are completed.
   - It also makes wait_for_save() safe if called multiple times: an already completed
     generator is simply already done.

2) Clear layerwise storers after finalization
   - After draining, the adapter clears the list to avoid reusing stale generators in later
     forward steps.

3) Ensure unpin happens even when metadata is empty
   - Under some steps, connector metadata can be empty at wait_for_save() time.
   - We track req_ids used for layerwise storing in the current forward and use them as a
     fallback for lookup_unpin(). This avoids leaked pins when metadata is missing.

4) Make freeze-mode yield count consistent
   - In freeze mode, store_layer() still yields the expected number of steps.
   - This prevents subtle off-by-one behavior where integrations assume a specific step count.

Operational note
----------------
The fix is not intended to hide real errors. It treats StopIteration from the layerwise
pipeline generator as normal completion, which is the correct generator semantics.
All other unexpected exceptions still surface.

Files changed (implementation)
------------------------------
- lmcache/integration/vllm/vllm_v1_adapter.py
  - Drain store_layer generators in wait_for_save() and clear them.
  - Track req_ids for unpin even if connector metadata becomes empty.

- lmcache/v1/cache_engine.py
  - In freeze mode, yield an additional finalize step to match expected step count.

- lmcache/v1/offload_server/zmq_server.py
  - Add an INFO log on server start to help debug

Related fix discovered while debugging
-------------------------------------
Separately, we also fixed lmcache/utils.py:parse_cache_key().

The cache key string format now includes an explicit dtype segment:

  fmt@model@world_size@worker_id@chunk_hash@dtype[@layer_id][@tag%value...]

parse_cache_key() decides whether a key contains a layer_id by checking whether the
corresponding '@' segment is numeric. After adding the dtype segment, the layer_id
index shifts by one, so the old detection logic could mis-classify layer keys.

This is not the direct cause of the StopIteration crash above, but it is a correctness
fix for key parsing that was required to keep layer-key detection consistent with the
new format.


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1279,12 +1292,23 @@ def wait_for_save(self):
...
 
         if self.use_layerwise:
+            # Drain all remaining steps to guarantee completion.
             for layerwise_storer in self.layerwise_storers:
-                next(layerwise_storer)
+                while True:
+                    try:
+                        next(layerwise_storer)
+                    except StopIteration:
+                        break
```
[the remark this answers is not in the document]
DongDongJu: resolved by https://github.com/LMCache/LMCache/pull/2268/changes/ab29c4af755b9fb5b3aba5b2c2f4b1fbc3c6f821.

lmcache/v1/cache_engine.py
```diff
@@ -514,9 +514,13 @@ def store_layer(
                 "Freeze mode enabled, skipping store_layer for %d tokens",
                 num_to_store_tokens,
             )
-            # Still need to yield to avoid StopIteration
+            # Still need to yield to match the expected step count in
+            # layerwise integrations (e.g., vLLM calls next() once per layer
+            # plus one extra in wait_for_save()).
             for layer_id in range(self.num_layers):
                 yield
+            # One extra yield for the "finalize" step.
```
sammshen, line 516: what's the finalize step? wait_for_save?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1279,12 +1292,20 @@ def wait_for_save(self):
             return
 
         if self.use_layerwise:
+            # Drain all remaining steps to guarantee completion.
             for layerwise_storer in self.layerwise_storers:
-                next(layerwise_storer)
+                for _ in layerwise_storer:
+                    pass
```
sammshen, line 1339: what is this pass? what's the point of iterating if we don't do anything?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1279,12 +1292,20 @@ def wait_for_save(self):
...
+
+            # Prevent double-finalization if wait_for_save() is called twice.
```
sammshen, line 1341: why will wait_for_save() be called twice?

lmcache/integration/vllm/vllm_v1_adapter.py
DongDongJu: The loop is intentionally exhausting the generator. 
its work same as below code
```
                while True:
                    try:
                        next(layerwise_storer)
                    except StopIteration:
                        break
```
store_layer() uses yield as step barriers iterating consumes the remaining steps so all async offload stages are completed before returning.

lmcache/integration/vllm/vllm_v1_adapter.py
DongDongJu: from my understanding, vLLM can invoke it in multiple forward steps even when a later step has nothing to save (e.g., empty connector metadata), and without clearing, stale storers from a previous step could be finalized again.
Clearing makes the method idempotent/safe under such call patterns.
But, let me know if I am wrong at here.

lmcache/v1/cache_engine.py
DongDongJu: Yes it's the final stage of the store_layer() generator (the extra progress after the last layer) so the pipeline can finish the last pending put/cleanup. 
vLLM reaches it by exhausting the generator in wait_for_save().
but same here, please correct me if im wrong.

sammshen:
apologies for the delay, will review this tomorrow!

lmcache/integration/vllm/vllm_v1_adapter.py
sammshen: thanks for the reminder on iterator behavior. I forgot that for loop is sugar over next() in python

lmcache/v1/cache_engine.py
sammshen: could you leave this as a comment here?

sammshen (approved):
This PR makes sense to me. Thanks for the extremely detailed write up. 

Do you think the generator code needs refactoring or is most of the complexity unavoidable due to vllm scheduler behavior? how did you find out most of these unusual behaviors?

Did you run any tests?

lmcache/utils.py
```diff
@@ -318,13 +318,13 @@ def parse_cache_key(key_str: str) -> Union[CacheEngineKey, LayerCacheEngineKey]:
...
         key_str: String in format:
-            fmt@model@world_size@worker_id@chunk_hash[@layer_id][@tag%value...]
+            fmt@model@world_size@worker_id@chunk_hash@dtype[@layer_id][@tag%value...]
 
     Returns:
         CacheEngineKey if no layer_id, LayerCacheEngineKey if valid layer_id
     """
     parts = key_str.strip().split("@")
-    if len(parts) >= 6 and parts[5].isdigit():
+    if len(parts) >= 7 and parts[6].isdigit():
```
YaoJiayi, line 327: Why is this needed?

YaoJiayi (commented):
@DongDongJu I'm still a bit confused about why this PR is needed. maybe it's more helpful to point out the related vllm code as well? Do you mean that vllm can execute multiple forward steps at the same time?

lmcache/utils.py
DongDongJu: This is not the direct cause of the StopIteration crash above, but it is a correctness fix for key parsing that was required to keep layerwise mode test.

DongDongJu:
Hello @YaoJiayi , Happy holidays.

> @DongDongJu I'm still a bit confused about why this PR is needed. maybe it's more helpful to point out the related vllm code as well? Do you mean that vllm can execute multiple forward steps at the same time?

Sure, Let me try to clarify what I understood about layerwise mode's design first.

`save_kv_layer()` advances a storage generator(per layer) via next().
`wait_for_save()` finalizes that Storage Generator with a last next() call ( and, completes the batched puts/ cleanup).

So what exactly solving here is that,
- layerwise saving uses a generator returned by LMCacheEngine.store_layer()
- vLLM pumps it during attention (save_kv_layer) and “finalizes” it at forward-exit (wait_for_save) but vLLM can call wait_for_save() on a forward step that has nothing to save (metadata empty- (current case is not None, empty list) / not savable / warmup, etc.) or call it multiple times.
few examples : [metadata empty](https://github.com/vllm-project/vllm/blob/decc244767f276fdb4ab147ef227835c3a91cd13/vllm/attention/utils/kv_transfer_utils.py#L46), [dummy runs](https://github.com/vllm-project/vllm/blob/b5d2d71d26b28ad4f9cee8698fa20030835170ad/vllm/v1/worker/gpu_worker.py#L451), [request not savable](https://github.com/vllm-project/vllm/blob/b5d2d71d26b28ad4f9cee8698fa20030835170ad/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_integration/vllm_v1_adapter.py#L1298).
[“Wait-for-save” happens even if metadata is empty (default behavior)](https://github.com/vllm-project/vllm/blob/b5d2d71d26b28ad4f9cee8698fa20030835170ad/vllm/v1/worker/kv_connector_model_runner_mixin.py#L131)
[LMCache connector start_load_kv also no-ops when attn_metadata is None](https://github.com/vllm-project/vllm/blob/b5d2d71d26b28ad4f9cee8698fa20030835170ad/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_integration/vllm_v1_adapter.py#L810)
- if LMCache’s adapter still holds stale layerwise storers from a prior step, the old code would call next() on an already exhausted generator → StopIteration → vLLM treats it as fatal

Let me indicate more details with vllm codes.

vLLM v1 calls `kv_connector.wait_for_save()` at the end of every forward step via
```
KVConnectorModelRunnerMixin._get_kv_connector_output(...): 
finally: 
    if wait_for_save: 
        kv_connector.wait_for_save()
```
But the LMCache save hook is *conditional*: `maybe_transfer_kv_layer` is a no-op when `attn_metadata is None` **or** `not connector.has_connector_metadata()`, so some steps legitimately never call `save_kv_layer()` (e.g. empty/unsavable connector metadata, or dummy/warmup/graph runs without attn metadata). [reference](https://github.com/vllm-project/vllm/blob/decc244767f276fdb4ab147ef227835c3a91cd13/vllm/attention/utils/kv_transfer_utils.py#L46)

So, actually for metadata, most easy fix is change [this lines](https://github.com/vllm-project/vllm/blob/decc244767f276fdb4ab147ef227835c3a91cd13/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_integration/vllm_v1_adapter.py#L554) which makes metadata to none if nothing in there or change [this lines](https://github.com/vllm-project/vllm/blob/decc244767f276fdb4ab147ef227835c3a91cd13/vllm/attention/utils/kv_transfer_utils.py#L46) check that length of meta is not 0. I thought that might be good to have a guard in LMCache side for the future too.

Additionally, why it was not happened during sync one is when use_layerwise is off, LMCache isn’t driving a multi-yield generator that needs “finalization by next()”, so we don’t get the specific “StopIteration → EngineCore dies” failure mode. The “weird” behavior (vLLM calling wait_for_save() even on no-save steps) can still happen, but without the layerwise generator in the middle, there’s nothing to accidentally advance past completion. 

@ApostaC you might also need to check.

lmcache/v1/cache_engine.py
DongDongJu: Sure will do. Thanks!

DongDongJu:
> This PR makes sense to me. Thanks for the extremely detailed write up.
> 
> Do you think the generator code needs refactoring or is most of the complexity unavoidable due to vllm scheduler behavior? how did you find out most of these unusual behaviors?

Hello @sammshen. Happy holidays.

I believe vllm's behavior is complex enough and like I mentioned reply on jiayi's thread there are multiple case this one can happened and in the future with vllm code changes as well.
So, good to have it as a guard for breakdown because this patch's logic do not break the KVCache itself.

> 
> Did you run any tests?

Yes, like recent issues https://github.com/LMCache/LMCache/issues/2327, we can easily reproduce it with example workloads. I used multi turn chat with 10 user concurrencies.

ningziwen:
> > This PR makes sense to me. Thanks for the extremely detailed write up.
> > Do you think the generator code needs refactoring or is most of the complexity unavoidable due to vllm scheduler behavior? how did you find out most of these unusual behaviors?
> 
> Hello @sammshen. Happy holidays.
> 
> I believe vllm's behavior is complex enough and like I mentioned reply on jiayi's thread there are multiple case this one can happened and in the future with vllm code changes as well. So, good to have it as a guard for breakdown because this patch's logic do not break the KVCache itself.
> 
> > Did you run any tests?
> 
> Yes, like recent issues #2327, we can easily reproduce it with example workloads. I used multi turn chat with 10 user concurrencies.

I pulled the commits in this PR to test and confirmed the issue in #2327 still exists.

DongDongJu:
> > > This PR makes sense to me. Thanks for the extremely detailed write up.
> > > Do you think the generator code needs refactoring or is most of the complexity unavoidable due to vllm scheduler behavior? how did you find out most of these unusual behaviors?
> > 
> > 
> > Hello @sammshen. Happy holidays.
> > I believe vllm's behavior is complex enough and like I mentioned reply on jiayi's thread there are multiple case this one can happened and in the future with vllm code changes as well. So, good to have it as a guard for breakdown because this patch's logic do not break the KVCache itself.
> > > Did you run any tests?
> > 
> > 
> > Yes, like recent issues #2327, we can easily reproduce it with example workloads. I used multi turn chat with 10 user concurrencies.
> 
> I pulled the commits in this PR to test and confirmed the issue in #2327 still exists.

Hello @ningziwen, Sorry for my wrong indication.
I was looking over quickly and this one is other issue.
I will continue the discuss it in #2327.

DongDongJu:
IDW which patch resolve this but it seems problem disappear.

PR #2284, opened 2025-12-19
[refactor] move get_shapes to LMCacheEngineMetadata

chunxiaozheng:
move `get_shapes` from `gpu_connector` to `metadata`, it is convenient for us to use in other places, such as `LocalCPUBackend` and `RemoteConnector`.
In addition, this can help us remove some methods in `gpu_connector`, such as `init_group_info`.

### DIFF
--- a/lmcache/config.py
+++ b/lmcache/config.py
@@ -61,6 +61,45 @@ class LMCacheEngineMetadata:
             ]
         return [self.kv_dtype]
 
+    def get_shapes(self, num_tokens: Optional[int] = None) -> list[torch.Size]:
+        """Get the shapes of the KV cache in LMCache"""
+        if num_tokens is None:
+            num_tokens = self.chunk_size
+        if self.kv_layer_groups_manager.kv_layer_groups:
+            shapes = []
+            kv_size = 1 if self.use_mla else 2
+            for group in self.kv_layer_groups_manager.kv_layer_groups:
+                layers = group.num_layers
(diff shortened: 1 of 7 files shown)


lmcache/config.py
```diff
@@ -61,6 +61,45 @@ def get_dtypes(self) -> list[torch.dtype]:
...
+        """Get the shapes of the KV cache in LMCache"""
+        if num_tokens is None:
+            num_tokens = self.chunk_size
+        if self.kv_layer_groups_manager.kv_layer_groups:
+            shapes = []
+            kv_size = 1 if self.use_mla else 2
+            for group in self.kv_layer_groups_manager.kv_layer_groups:
+                layers = group.num_layers
+                num_heads = 1 if self.use_mla else group.shape[3]
+                hidden_dim_size = num_heads * group.shape[-1]
```
ApostaC: Dumb question: Will similar logic be used multiple times when we use `kv_layer_groups` in different places? If so, we can probably implement them as some methods in the `kv_layer_group` class.

ApostaC (approved):
One quick question. Otherwise LGTM

lmcache/config.py
chunxiaozheng: @ApostaC Good idea, this is a common attribute of `kv_layer_group`, I have updated!

maobaolong (approved):
lgtm

Issue #2291, opened 2025-12-22
[Bug] Wrong response with FlashInfer or Titron backend


ziruiliu:
**Description**
Start vLLM with flashinfer or triton, ask a simple 16 token questions twice to test retrieving cache from lmcache. The answer is not related which means cache is corrupted

**To Reproduce**
Start vLLM with a small model, chunk size 16
```
VLLM_ATTENTION_BACKEND=FLASHINFER \
LMCACHE_CHUNK_SIZE=16 \
LMCACHE_LOG_LEVEL=DEBUG \
vllm serve Qwen/Qwen3-0.6B \
--no-enable-prefix-caching --enforce-eager \
--kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
```

Query with a 16-token prompt REPEATEDLY
```
curl -X POST http://localhost:8000/v1/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen/Qwen3-0.6B",
    "prompt": "Here is the question: in what year did Stanford University officially open to students?",
    "max_tokens": 10
  }'
```

**Expected behavior**
Correct Answer: anything about the year.

With FLASH_ATTN (which is default), the response is correct
With TRITON_ATTN or FLASHINFER, the response is full of random tokens

**Cause**
flashinfer and triton has shape (num_blocks, 2, block_size, num_kv_heads, head_size)
flash attention's shape (2, num_blocks, block_size, num_kv_heads, head_size)
gpu_connector may handle flashinfer and triton attention's kv cache wrong? 
In gpu_connector and mem_kernels.cu's `multi_layer_kv_transfer` I only see flash attention's shape

ziruiliu:
kv cache shape [num_blocks, 2, block_size, num_kv_heads, head_size] is well handled by single_layer_xxx functions in mem_kernel.cu and layerwised GPU connectors in gpu_connector.py. However such kv cache shape has to be handled by multi layer functions and gpu connectors in a similar way

DongDongJu:
It seems hard question. it should be agnostic in connector side between vllm and LMCache not dealing the shape inside of LMCache.
@ApostaC what do you think?

PR #2308, opened 2025-12-24
[Bug][core] make multi_layer_kv_transfer() deal with flashinfer and triton

ziruiliu:
**Why**
As described in #2291 ,  `multi_layer_kv_transfer()` in mem_kernels.cu assumed a single vLLM layout and produced incorrect offsets when vLLM uses [num_blocks, 2, block_size, ...]  causing data copy errors with flashinfer or triton
Single-layer transfers `single_layer_kv_transfer()` already had a vllm_two_major flag,  so the multi-layer path and GPU connectors needed the same awareness to handle both layouts correctly.

**What**
- Added `vllm_two_major` and `block_size` parameters to the `multi_layer_kv_transfer` declaration in `csrc/mem_kernels.cuh` and threaded them through the kernel launches in `csrc/mem_kernels.cu`.
- Made `page_buffer_offset` layout-aware by updating its signature and logic to compute offsets for both two-major and block-major layouts, and passed `block_size`/`vllm_two_major` into the `load_and_reshape_multi_layer_kernel`.
- Updated GPU connectors (`VLLMPagedMemGPUConnectorV2`, `VLLMPagedMemGPUConnectorV3`) to detect the vLLM KV cache layout and compute `page_buffer_size` and `block_size`, and pass `self.vllm_two_major`/`self.block_size` into `lmc_ops.multi_layer_kv_transfer` calls.
- Updated `GPUCacheContext` in `lmcache/v1/multiprocess/server.py` to detect layout (`vllm_two_major_`) and pass `vllm_two_major`/`block_size` through the places that call `lmc_ops.multi_layer_kv_transfer`.
- Updated `test_mem_kernels.py` accordingly

### DIFF
--- a/csrc/mem_kernels.cu
+++ b/csrc/mem_kernels.cu
@@ -171,17 +171,9 @@ __global__ void single_layer_kv_transfer_kernel(
 
 __device__ __forceinline__ int64_t page_buffer_offset(
     const int k_or_v, const int token_idx, const int scalar_offset,
-    const int scalars_per_token, const int page_buffer_size,
-    const int block_size, const bool vllm_two_major) {
-  if (vllm_two_major) {
-    return k_or_v * page_buffer_size * scalars_per_token +
-           token_idx * scalars_per_token + scalar_offset;
-  }
-  const int block_idx = token_idx / block_size;
-  const int block_offset = token_idx % block_size;
-  return block_idx * (2 * block_size * scalars_per_token) +
-         k_or_v * (block_size * scalars_per_token) +
-         block_offset * scalars_per_token + scalar_offset;
+    const int scalars_per_token, const int page_buffer_size) {
+  return k_or_v * page_buffer_size * scalars_per_token +
+         token_idx * scalars_per_token + scalar_offset;
 }
 
 __device__ __forceinline__ int64_t page_buffer_offset_unilateral(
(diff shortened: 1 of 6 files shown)


ApostaC:
Hey @ziruiliu , appreciate the contribution!

I do have a concern about the code. As the arguments become more and more, it becomes very hard to track the meaning of each argument (e.g., it's not clear what `vllm_two_major` means if we only read the variable name), and it creates a lot of confusion in the Python layer when seeing a list of `True/False` arguments.

Therefore, I would like to see another non-functionality refactoring to export enum definition from c-level to python-level via pybind.

I'm open to offline discussions to move this forward. Please let me know!

ziruiliu:
> Hey @ziruiliu , appreciate the contribution!
> 
> I do have a concern about the code. As the arguments become more and more, it becomes very hard to track the meaning of each argument (e.g., it's not clear what `vllm_two_major` means if we only read the variable name), and it creates a lot of confusion in the Python layer when seeing a list of `True/False` arguments.
> 
> Therefore, I would like to see another non-functionality refactoring to export enum definition from c-level to python-level via pybind.
> 
> I'm open to offline discussions to move this forward. Please let me know!

Totally agree. My PR is simply following the style of single layer functions those using vllm_two_major. Do you want me to add some enum definition and also apply them to existing single layer transfer functions?

ziruiliu:
> Hey @ziruiliu , appreciate the contribution!
> 
> I do have a concern about the code. As the arguments become more and more, it becomes very hard to track the meaning of each argument (e.g., it's not clear what `vllm_two_major` means if we only read the variable name), and it creates a lot of confusion in the Python layer when seeing a list of `True/False` arguments.
> 
> Therefore, I would like to see another non-functionality refactoring to export enum definition from c-level to python-level via pybind.
> 
> I'm open to offline discussions to move this forward. Please let me know!

Let's pass the flags and options this way:
1. Use enum Direction::D2H and Direction::H2D to replace current True/False direction.
2. Use KVLayout::MLA, KVLayout::BlockFirst (shape starts with num_blocks) and KVLayout::KVFirst (shape starts with 2) to replace current True/False flags.

@ApostaC I will soon prepare an update. If you have any other idea please let me know. Thanks

ziruiliu:
updated to enum, but still working on unit test failures

ziruiliu:
Hi @ApostaC , I have updated this fix, introduced two enums:
1. TransferDirection::D2H and H2D to replace True/False for read/write
2. KVLayout::MLA, KVFirst and BlockFirst to represent different kv cache shapes, to replace bool value vllm_two_major

Please take a look at it, thanks.

ApostaC:
@ziruiliu Hey Zirui, thanks for the update! I will take a look at it soon!

ApostaC:
Hey @ziruiliu , just wanted to let you know that I have picked a few lines of your changes in this PR into #2413

sammshen:
Hey @ziruiliu , just wanted to let you know that I will pick a few lines of your changes in this PR into https://github.com/LMCache/LMCache/pull/2518

ziruiliu:
> Hey @ziruiliu , just wanted to let you know that I will pick a few lines of your changes in this PR into #2518

Got it. I understand that the community has its own priorities to refactor current code. Please feel free to use code from this PR

ziruiliu:
I believe the issue this PR trying to address is already fixed in latest LMCache

PR #2309, opened 2025-12-24
[Core] Add multi-layer transfer kernel for ds v3.2

ApostaC:

For ds v3.2, the hidden dim size is 132 * int8 for a single token. However, the current multi-layer transfer requires 8-byte alignment, which causes a correctness issue.

**In this PR**:

We introduced the following things:
1. A unittest to test the case for dsv3.2 (132-byte) copy
2. Make the multi layer kv transfer a templated function that accepts different copy types.

**In the future**:

For dsv32, since a single token's hidden size is very small (132 byte), using a thread-block to copy a single token single layer's KV cache could be inefficient.

We probably want to let a thread block copy the whole "page" to improve the efficiency.

**If applicable**:

### DIFF
--- a/csrc/mem_kernels.cu
+++ b/csrc/mem_kernels.cu
@@ -263,7 +263,7 @@ __global__ void load_and_reshape_multi_layer_kernel(
   const int num_threads = blockDim.x;
 
   const int64_t slot_idx = slot_mapping[token_id];
-  int64_t* paged_buffer_ptr = paged_buffer_ptrs[layer_id];
+  scalar_t* paged_buffer_ptr = paged_buffer_ptrs[layer_id];
 
   if (slot_idx < 0) {
     return;
@@ -305,8 +305,8 @@ __global__ void load_and_reshape_multi_layer_kernel_unilateral(
   const int num_threads = blockDim.x;
 
   const int64_t slot_idx = slot_mapping[token_id];
-  int64_t* key_ptr = paged_buffer_ptrs[layer_id];
-  int64_t* value_ptr = paged_buffer_ptrs[layer_id + num_layers];
+  scalar_t* key_ptr = paged_buffer_ptrs[layer_id];
+  scalar_t* value_ptr = paged_buffer_ptrs[layer_id + num_layers];
 
   if (slot_idx < 0) {
     return;
@@ -381,7 +381,8 @@ T* get_kernel_ptr(TENSOR_TYPE& tensor) {
  *  - direction: false  means LMCache to PagedBuffer, true  means PagedBuffer to
  * LMCache
  */
-void multi_layer_kv_transfer(
+template <typename T>
+void multi_layer_kv_transfer_templated(
     torch::Tensor&
         key_value,  // key/value must be on gpu/pinned cpu.
                     // [2, num_layer, num_tokens, num_heads*head_size] for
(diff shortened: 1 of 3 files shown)


ApostaC:
cc @chunxiaozheng @sammshen @YaoJiayi

csrc/mem_kernels.cu
```diff
@@ -393,44 +394,75 @@ void multi_layer_kv_transfer(
...
+  if (copy_size % 8 == 0) {
+    LAUNCH_MULTI_LAYER_KV_TRANSFER(int64_t);
+  } else if (copy_size % 4 == 0) {
+    LAUNCH_MULTI_LAYER_KV_TRANSFER(int32_t);
+  } else if (copy_size % 2 == 0) {
+    LAUNCH_MULTI_LAYER_KV_TRANSFER(int16_t);
+  } else {
+    LAUNCH_MULTI_LAYER_KV_TRANSFER(int8_t);
+  }
+#undef LAUNCH_MULTI_LAYER_KV_TRANSFER
```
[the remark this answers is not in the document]
ApostaC: Er.. I didn't see any `siez` in the code. Are you still okay😅?

csrc/mem_kernels.cu
```diff
@@ -393,44 +394,75 @@ void multi_layer_kv_transfer(
...
+      get_kernel_ptr<T*, const torch::Tensor>(key_value_ptrs);
   const int64_t* slot_mapping_ptr =
       get_kernel_ptr<const int64_t, const torch::Tensor>(slot_mapping);
 
   int num_layers = key_value.size(1);
   int num_tokens = slot_mapping.size(0);
   int num_origin_elements = key_value.size(3);
-  int elements_per_qword = 8 / key_value.element_size();
-  int num_qwords = num_origin_elements / elements_per_qword;
+  int elements_per_xword = sizeof(T) / key_value.element_size();
```
sammshen, line 406: what if sizeof(T) < key_value.element_size()?

tests/v1/test_mem_kernels.py
```diff
@@ -292,17 +286,17 @@ def test_multi_layer_kernel(num_tokens):
 
 
 @pytest.mark.parametrize("num_tokens", [256, 500, 1024, 8000])
-def test_multi_layer_kernel_use_mla(num_tokens):
+@pytest.mark.parametrize("head_size", [576, 66])  # Use 68 for dsv32 (132x int8)
```
chunxiaozheng, line 289: A minor question, what is the `66` means?

chunxiaozheng:
@ApostaC Thanks for your quick fix! I have tested in our env, this makes sense!

csrc/mem_kernels.cu
chunxiaozheng: I think `sizeof(T) < key_value.element_size()` will not occur in current situations, the following code will set `T`.
```
if (copy_size % 8 == 0) {
    LAUNCH_MULTI_LAYER_KV_TRANSFER(int64_t);
  } else if (copy_size % 4 == 0) {
    LAUNCH_MULTI_LAYER_KV_TRANSFER(int32_t);
  } else if (copy_size % 2 == 0) {
    LAUNCH_MULTI_LAYER_KV_TRANSFER(int16_t);
  } else {
    LAUNCH_MULTI_LAYER_KV_TRANSFER(int8_t);
  }
```

There is an example, if `key_value.element_size()` is `8`, the `copy_size = num_origin_elements * key_value.element_size()` and `copy_size % 8 == 0`, the `T` is set to `int64_t`, and the `sizeof(T) >= key_value.element_size()`.

chunxiaozheng: Only the `key_value.element_size()` is `16`, which will cause ` sizeof(T) < key_value.element_size()`, but I don't think we will use kv cache dtypes that exceed 8 bytes? @ApostaC what do you think?

csrc/mem_kernels.cu
```diff
@@ -393,44 +394,75 @@ void multi_layer_kv_transfer(
...
+ * @see multi_layer_kv_transfer_templated
+ */
+void multi_layer_kv_transfer(torch::Tensor& key_value,
+                             const torch::Tensor& key_value_ptrs,
+                             const torch::Tensor& slot_mapping,
+                             const torch::Device& paged_memory_device,
+                             const int page_buffer_size, const bool direction,
+                             const bool use_mla) {
+  int num_origin_elements = key_value.size(3);
+  int copy_size = num_origin_elements * key_value.element_size();
```
sammshen, line 445: should we rename `copy_size` to `single_token_copy_bytes` or `single_token_copy_size`?

csrc/mem_kernels.cu
sammshen: thanks for the explanation

have not seen any models using 128 bit kv cache so ignoring seems safe but leave the decision to @ApostaC

ApostaC: Good point. It's unlikely to have element size = 16 as @chunxiaozheng explained

chunxiaozheng (approved):
lgtm

tests/v1/test_mem_kernels.py
chunxiaozheng: 66 * 2(`sizeof(bfloat16)`) = 132

sammshen (approved):
LGTM

Issue #2166, opened 2025-12-04
[BUG]: lmcache v0.3.10 mooncake 0.3.7.post2 batched_input failed


elfisworking:
<img width="1199" height="412" alt="Image" src="https://github.com/user-attachments/assets/d4bb5851-3a03-4da5-9cce-7df0c9703f62" />
error log 

```bash
LMCache WARN : batched put error: unsupported format string passed to bytes.__format__ (instrumented_connector.py:118:lmcache.v1.storage_backend.connector.instrumented_connector)
```

Using lmcache v0.3.10 mooncake 0.3.2.post2 failed to  store tokens to mooncake store.
configurations is 

```yaml
chunk_size: 256
local_cpu: false
max_local_cpu_size: 5.0
numa_mode: auto
pre_caching_hash_algorithm: sha256_cbor_64bit
remote_url: mooncakestore://127.0.0.1:50051/
remote_serde: naive
extra_config:
  local_hostname: localhost
  metadata_server: etcd://localhost:2379
  protocol: tcp
  device_name: ''
  master_server_address: localhost:50051
  global_segment_size: 21474836480
  local_buffer_size: 0
  transfer_timeout: 5
  save_chunk_meta: false
  storage_root_dir: ''
  use_exists_sync: true
  mooncacke_prefer_local_alloc: true
```
mooncake strore log show that no strorage have been used

```bash
All=0/0,  | Batch Requests (Req=Success/PartialSuccess/Total, Item=Success/Total): PutStart:(Req=0/0/0, Item=0/0), PutEnd:(Req=0/0/0, Item=0/0), PutRevoke:(Req=0/0/0, Item=0/0), Get:(Req=0/0/0, Item=0/0), ExistKey:(Req=0/0/0, Item=0/0),  | Eviction: Success/Attempts=0/0, keys=0, size=0 B
[...]

PR #2313, opened 2025-12-25
[Bugfix] fix CacheEngineKey.to_string() str fmt type err

ljwh:

Class `CacheEngineKey` formats key with str template which specifies `chunk_hash` to hex in the `to_string()` function. This may cause an error if `chunk_hash` is a normal string like generated by `sha256_cbor`. Related issue: #2166

tiny change

**If applicable**:

### DIFF
--- a/lmcache/utils.py
+++ b/lmcache/utils.py
@@ -385,7 +385,7 @@ class CacheEngineKey:
     def to_string(self):
         s = (
             f"{self.fmt}@{self.model_name}@{self.world_size}"
-            f"@{self.worker_id}@{self.chunk_hash:x}@{self._dtype_str}"
+            f"@{self.worker_id}@{self.chunk_hash}@{self._dtype_str}"
         )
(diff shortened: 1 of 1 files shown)


DongDongJu:
Hello @ljwh, Happy holidays,
If you dont mind could you check once with this changes? https://github.com/LMCache/LMCache/pull/2268#issuecomment-3696977534
And, also please check the DCO.

xeonliu (changes requested):
```
@dataclass(slots=True)
class LayerCacheEngineKey(CacheEngineKey):
    def to_string(self):
        s = (
            f"@{self.worker_id}@{self.chunk_hash:x}@{self._dtype_str}@{self.layer_id}"
        )
        if self.tags is not None and len(self.tags) != 0:
            tags = [f"{k}%{v}" for k, v in self.tags]
            s += "@" + "@".join(tags)
        return s
```

This one also needs to change. Line 532 in lmcache/utils.py

PR #2314, opened 2025-12-26
[refactor] refactor RemoteConnector

chunxiaozheng:
refactor `RemoteConnector`, reference to https://github.com/LMCache/LMCache/pull/2266#discussion_r2644457897, I first refactor the `RemoteConnector`, move the attribute to `__init__` method, and rename `shape`, `dtype` to `shapes`, `dtypes`

### DIFF
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -221,8 +221,6 @@ class ConnectorManager:
             if adapter.can_parse(self.context.url):
                 logger.info(f"Creating connector for URL: {self.context.url}")
                 connector = adapter.create_connector(self.context)
-                logger.info(f"initializing chunk meta for connector: {connector}")
-                connector.init_chunk_meta(self.context.config, self.context.metadata)
                 logger.info(f"post-initializing connector: {connector}")
                 connector.post_init()
                 return connector
--- a/lmcache/v1/storage_backend/connector/__init__.py
+++ b/lmcache/v1/storage_backend/connector/__init__.py
@@ -221,8 +221,6 @@ class ConnectorManager:
             if adapter.can_parse(self.context.url):
                 logger.info(f"Creating connector for URL: {self.context.url}")
                 connector = adapter.create_connector(self.context)
-                logger.info(f"post-initializing connector: {connector}")
-                connector.post_init()
                 return connector
 
         raise ValueError(f"No adapter found for URL: {self.context.url}")
--- a/lmcache/v1/storage_backend/connector/base_connector.py
+++ b/lmcache/v1/storage_backend/connector/base_connector.py
@@ -35,6 +35,19 @@ class RemoteConnector(metaclass=abc.ABCMeta):
     def __init__(
         self, config: LMCacheEngineConfig, metadata: Optional[LMCacheEngineMetadata]
     ):
+        """
+        Initialize some common attributes, which will be used in the subclasses.
+        - `save_chunk_meta` is a flag to indicate whether to save the chunk meta info.
+        - `meta_shapes` is a list of shapes of lmcache full chunk.
+        - `meta_dtypes` is a list of dtypes of lmcache chunk.
+        - `meta_fmt` is the memory format of the lmcache chunk.
+        - `full_chunk_size` is the size of the lmcache full chunk.
+        - `single_token_size` is the size of a single token.`
+
+        Input:
+            config: the lmcache engine config
+            metadata: the lmcache engine metadata
+        """
         # TODO(chunxiaozheng): support layerwise here
         assert metadata is not None
         self.save_chunk_meta: bool = (
--- a/lmcache/v1/storage_backend/connector/base_connector.py
+++ b/lmcache/v1/storage_backend/connector/base_connector.py
@@ -32,54 +32,32 @@ class RemoteConnector(metaclass=abc.ABCMeta):
     Interface for remote connector
     """
 
-    save_chunk_meta: bool = True
-    meta_shape: Optional[torch.Size] = None
-    meta_dtype: Optional[torch.dtype] = None
-    meta_fmt: Optional[MemoryFormat] = None
-    full_chunk_size: Optional[int] = None
-    single_token_size: Optional[int] = None
-
-    @NotAudit
-    def init_chunk_meta(
-        self,
-        config: Optional[LMCacheEngineConfig],
-        metadata: Optional[LMCacheEngineMetadata],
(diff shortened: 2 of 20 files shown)


lmcache/v1/storage_backend/connector/s3_connector.py
```diff
@@ -83,11 +83,12 @@ def __init__(
...
         aws_secret_access_key: Optional[str] = None,
     ):
+        super().__init__(local_cpu_backend.config, local_cpu_backend.metadata)
+
         if not s3_endpoint.startswith("s3://"):
             raise ValueError("S3 url must start with 's3://'")
 
-        # post initialized
-        self.s3_part_size = None
+        self.s3_part_size = self.full_chunk_size
```
sammshen, line 92: this is nice

lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -118,6 +118,8 @@ def __init__(
         loop: asyncio.AbstractEventLoop,
         memory_allocator: LocalCPUBackend,
     ):
+        super().__init__(memory_allocator.config, memory_allocator.metadata)
```
sammshen, line 122: can we add a comment in each connector on what this super init is doing? like variables / data will be accessible in the child connector. I remember this was a little confusing during refactoring the S3 connector

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -32,54 +32,32 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
-        config: Optional[LMCacheEngineConfig],
-        metadata: Optional[LMCacheEngineMetadata],
-    ) -> None:
-        logger.info("initializing chunk meta for remote connector")
-        # TODO: support layerwise later
-        if (
-            config is None
-            or metadata is None
-            or config.extra_config is None
+    def __init__(
```
sammshen, line 35: can we add a clear doc string to this method as well as post_init clearly defining their responsibilites?

sammshen (approved):
LGTM! Some nits on clearer doc strings so that future developers can understand!

lmcache/v1/storage_backend/connector/base_connector.py
ApostaC: +1

ApostaC: Additionally, it would be great to have some clarity (i.e., docstrings) about what kind of fields this class has, because the sub-classes need to use those fields.

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -32,54 +32,32 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
-                metadata.kv_shape[1],
-                metadata.kv_shape[0],
-                metadata.kv_shape[2],
-                metadata.kv_shape[3] * metadata.kv_shape[4],
-            ]
         )
-        self.meta_dtype = metadata.kv_dtype
-        self.meta_fmt = (
+        self.meta_shapes: list[torch.Size] = metadata.get_shapes()
+        self.meta_dtypes: list[torch.dtype] = metadata.get_dtypes()
```
ApostaC, line 59: Right now, since `__init__` is called before `register_kv_caches`, I want to know what the behavior of `get_shapes` and `get_dtypes` here is, and whether it will impact the behavior when using normal models

ApostaC (commented):
One quick question: IIUC, we moved `init_chunk_meta` into `__init__` for `BaseConnector`. I noticed that some remote connectors previously have `init_chunk_meta()` but some don't have. Who will call `init_chunk_meta` before? And if a connect doesn't call `init_chunk_meta`, will it have any problems as it calls `super().__init__`? 

Additionally, it would be very helpful if we could have a developer doc about the initialization process of the modules (CacheEngine, StorageManager, StorageBackend, RemoteConnector). It can be done by AI, but I believe it will help onboard new contributors a lot.

Otherwise LGTM!

lmcache/v1/storage_backend/connector/eic_connector.py
chunxiaozheng: done

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: done

chunxiaozheng:
****

> One quick question: IIUC, we moved `init_chunk_meta` into `__init__` for `BaseConnector`. I noticed that some remote connectors previously have `init_chunk_meta()` but some don't have. Who will call `init_chunk_meta` before? And if a connect doesn't call `init_chunk_meta`, will it have any problems as it calls `super().__init__`?
> 
> Additionally, it would be very helpful if we could have a developer doc about the initialization process of the modules (CacheEngine, StorageManager, StorageBackend, RemoteConnector). It can be done by AI, but I believe it will help onboard new contributors a lot.
> 
> Otherwise LGTM!

@ApostaC Thanks for your review! `init_chunk_meta()` is  to initialize `meta_dtypes`, `meta_shapes` and other attributes, the current refactoring only puts the content of method `init_chunk_meta()` into method `__init__`. In the original implementation, each connector will call `init_chunk_meta()`, now we call through `super().__init__`, so there won't be any problems.

lmcache/v1/storage_backend/connector/__init__.py
```diff
@@ -221,10 +221,6 @@ def create_connector(self) -> RemoteConnector:
             if adapter.can_parse(self.context.url):
                 logger.info(f"Creating connector for URL: {self.context.url}")
                 connector = adapter.create_connector(self.context)
-                logger.info(f"initializing chunk meta for connector: {connector}")
```
chunxiaozheng, line 224: @ApostaC In the original implementation, each connector would call `init_chunk_meta` and `post_init` here.

By the way, previously, `post_init` needs to use some attributes with initialization in `init_chunk_meta()`, we must call `post_init` at the end, after move `init_chunk_meta` to `__init__`, the `post_init` is not need here, it can directly implements in `__init__` of every connector.

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: @ApostaC Current PR only refactor the origin implements, it does not support `DSA`,  the `get_shapes` and `get_dtypes` will get the shape and dtype from vllm config(the `dtype` and `shape` of LMCacheEngineMetadata).

ApostaC (approved):
LGTM!

Issue #2149, opened 2025-12-02
[RFC / Call for Help]: Backend / Connector Tests


sammshen:
# Motivation

Currently LMCache has many backends and connectors, any of which could be functional / non-functional at any given time. Furthermore, refactors to the Storage Manager or abstract base classes could break all of these backends/connectors at the same time. 

The main problem is that deploying each backend is configuration and resource intensive (e.g. we might need specific hardware or to set up an object store on a cloud console). It may not be possible to run all of these tests on a single CI machine. Furthermore, it is likely not desirable to tests all of the backends / connectors at the same time.

# Proposal

We add a tagging system similar to the current `full` tag that triggers the comprehensive test. We can add one per backend and those who have the resources to help build CI for a backend can reach out to hook up their machines via our buildkite etc. 

Easy backends / connectors to test: 
- Local CPU (already included in comprehensive test)
- Local Disk (already included in comprehensive test)
- P2P (already included in comprehensive test)
- PD (already included in comprehensive test)
- Nixl Storage Backend (the basic features)
- Redis Connector
- Mock Connector

Difficult backends / connectors to test: 
- Mooncake connector
- S3 Connector
- GDS backend and Weka GDS backend (to be merged)
- Nixl Storage Backend (the advanced RDMA features)

Zheng-Lu:
Good proposal! I’d like to start with a small first step for this RFC:

For a first step, I can add pytest markers for a few “easy” backends (e.g. local CPU / local disk / mock / redis), so that we can run tests like `pytest -m 'local_cpu'` or `pytest -m 'mock'` locally, I will also add documentation on how to run those tagged tests via pytest.

Once the markers are in place, CI changes to selectively run these tags can be a follow-up

If that sounds reasonable, I’ll send a PR that just adds the markers and a short note on how to use them.

sammshen:
Hey @Zheng-Lu , ping me on your PR once it's ready! :)

Zheng-Lu:
@sammshen PR created, feel free to take a look when you have time

skprasadu:
Hi LMCache team — I’m @skprasadu.
I see #2149 is assigned to @Zheng-Lu and PR #2216 is already tackling the pytest marker groundwork. To avoid overlap, I’d like to take the next chunk that makes the “easy connector” tests reproducible + CI-friendly:

Scope I’ll own (small, mergeable PRs):
[...]

PR #2315, opened 2025-12-26
[refactor] use shapes and dtypes in RemoteMetadata

chunxiaozheng:
use shapes and dtypes in RemoteMetadata

### DIFF
--- a/lmcache/v1/storage_backend/connector/base_connector.py
+++ b/lmcache/v1/storage_backend/connector/base_connector.py
@@ -43,6 +43,8 @@ class RemoteConnector(metaclass=abc.ABCMeta):
         - `meta_fmt` is the memory format of the lmcache chunk.
         - `full_chunk_size` is the size of the lmcache full chunk.
         - `single_token_size` is the size of a single token.`
+        - `remote_metadata_fmt` is the format of the remote metadata.
+        - `remote_metadata_bytes` is the size of the remote metadata.
 
         Input:
             config: the lmcache engine config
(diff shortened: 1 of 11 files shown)


lmcache/v1/storage_backend/connector/infinistore_connector.py
```diff
@@ -98,20 +100,20 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
-        ).reshape(metadata.shape)
+        ).reshape(metadata.shapes[0])
 
         memory_obj = self.memory_allocator.allocate(
-            metadata.shape,
-            metadata.dtype,
+            metadata.shapes[0],
+            metadata.dtypes[0],
             metadata.fmt,
         )
```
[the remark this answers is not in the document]
chunxiaozheng: `METADATA_BYTES_LEN` is defined in this file, are you ok?

lmcache/v1/storage_backend/connector/base_connector.py
```diff
@@ -32,54 +32,39 @@ class RemoteConnector(metaclass=abc.ABCMeta):
...
-        self.full_chunk_size = get_size_bytes([self.meta_shape], [self.meta_dtype])
-        assert self.full_chunk_size is not None
-        assert self.full_chunk_size % metadata.kv_shape[2] == 0
-        self.single_token_size = self.full_chunk_size // metadata.kv_shape[2]
+        self.full_chunk_size: int = get_size_bytes(self.meta_shapes, self.meta_dtypes)
+        assert self.full_chunk_size % metadata.chunk_size == 0
+        self.single_token_size = self.full_chunk_size // metadata.chunk_size
+
+        fmt_length = 2 + metadata.get_num_groups() * 5
+        self.remote_metadata_fmt = "i" * fmt_length
```
ApostaC: Since the `RemoteMetadata` class controls how to serialize and deserialize the metadata, should we make the format a private field inside the `RemoteMetadata` class?

Let's say in the future, someone wants to change the serialize/deserialize to use `msgpack`, such a `remote_data_fmt` may not be needed anymore. Then, it may require changing all the remote connector implementations if we use the current implementation.

For the current implementation, the only thing we may want to pass to the `deserialize` function is the value of `metadata.get_num_groups()`. And for the `serialize` function, we don't need to pass any extra arguments.

ApostaC (commented):
I suppose we need to do a rebase after #2314 is merged?

Left some comments about what we should pass to `serialize` and `deserialize`. Otherwise it's pretty clear to me! Also thanks for breaking it down!

lmcache/v1/storage_backend/connector/base_connector.py
chunxiaozheng: @ApostaC Thanks for your suggestion, I have updated.

I used global var to represents `remote_metadata_fmt`, if used private field, which will frequently creating strings by `"i" * fmt_length`.

ApostaC: I see. That also make sense. One quick question is that if the remote storage has the KV cache from different models (e.g., both dsv32 and dsv31), will the current code read corrupted data since the REMOTE_METADATA_FMT is now hard-coded?

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM, ideally we can protect this kind of PR in the future with https://github.com/LMCache/LMCache/issues/2149

chunxiaozheng:
> LGTM, ideally we can protect this kind of PR in the future with #2149

@sammshen Good idea, Thanks!

PR #2321, opened 2025-12-28
Add basic Kubernetes integration tests for LMCache

ningziwen:

This PR adds basic Kubernetes integration tests for LMCache to validate deployments in K8s environments. Currently includes two fundamental test scenarios with containerized execution for local development.

**Local execution:** These tests can be used for local development before LMCache has k8s cluster. They can be easily integrated into CI/CD when a K8s cluster with GPU nodes is available.

More k8s tests can come later following same style.

### DIFF
--- a/tests/k8s/manifests/sagemaker-hyperpod.yaml
+++ b/tests/k8s/manifests/sagemaker-hyperpod.yaml
@@ -1,0 +1,115 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: lmcache-sagemaker-hyperpod-test
+
+---
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: lmcache-config
+  namespace: lmcache-sagemaker-hyperpod-test
+data:
+  lmcache.yaml: |
+    chunk_size: 256
+    local_cpu: false
+    remote_url: "sagemaker-hyperpod://$NODE_IP:9200"
+    extra_config:
+      sagemaker_hyperpod_bucket: "lmcache"
+      sagemaker_hyperpod_shared_memory_name: "ai_toolkit_cache"
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: lmcache-sagemaker-hyperpod
+  namespace: lmcache-sagemaker-hyperpod-test
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: lmcache-sagemaker-hyperpod
+  template:
+    metadata:
+      namespace: lmcache-sagemaker-hyperpod-test
+      labels:
+        app: lmcache-sagemaker-hyperpod
+    spec:
+      containers:
+        - name: vllm
+          image: ${LMCACHE_IMAGE}
+          command:
+            - /opt/venv/bin/vllm
+            - serve
+            - Qwen/Qwen3-14B
+            - --host
+            - 0.0.0.0
+            - --port
+            - "8000"
+            - --no-enable-prefix-caching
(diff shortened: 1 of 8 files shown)


DongDongJu:
Hello @ningziwen, Happy holidays.
I think this one should go to the https://github.com/vllm-project/production-stack.
what do you think?

ningziwen:
> Hello @ningziwen, Happy holidays. I think this one should go to the https://github.com/vllm-project/production-stack. what do you think?

@DongDongJu 
Hi, thanks for the feedback and happy holidays! 🎉

That's a fair point to consider. Let me share my thinking:

The alternative would be to consolidate all K8s tests in vllm-project/production-stack.

My initial concern is that K8s environments can surface issues that non-K8s tests simply can't catch—and K8s is how most users actually run LMCache in production. This is particularly relevant for L2 cache remote connectors, many of which are inherently K8s-native and hard to properly validate outside of that environment. Additionally, we'll eventually need large-scale distributed testing to catch performance regressions, which requires K8s.

That said, this concern could be addressed by integrating production-stack's K8s tests into LMCache's CI pipeline. If we commit to doing that, it would be a reasonable solution for vLLM-related coverage.

The remaining gap, though, is that production-stack is scoped to vLLM. LMCache also needs to validate:
- **SGLang integration**
- **Multi-process standalone mode**

These use cases fall outside production-stack's scope, so there's no natural home for them there. Given this, I believe LMCache will eventually need its own K8s tests regardless.

---

## Summary

| Concern | Addressable via production-stack? |
|---------|-----------------------------------|
| K8s-native connector testing | ✅ Yes, via CI integration |
| Performance regression / load testing | ✅ Yes, via CI integration |
| SGLang integration | ❌ No, out of scope |
| Multi-process standalone mode | ❌ No, out of scope |



Would love to hear your thoughts!

ningziwen:
@ApostaC @sammshen Could you pls take a look?

sammshen (approved):
LGTM

tests/k8s/manifests/sagemaker-hyperpod.yaml
ApostaC, line 1: Quick question: does this test really need to connect to a sagemaker hyperpod bucket? Or how does it work?

ApostaC (commented):
Thanks for the great contribution! 

After taking a look at this PR, I feel like the most important thing in this PR is it's able to test the SageMaker HyperPod connector.

However, I would say we should need to get real sagemaker hyperpod to test this before merging this PR. Otherwise we don't have a good reason for introducing such a big part to the code without using it.

WDYT? @sammshen @DongDongJu

DongDongJu:
I do agree with @ApostaC. 
IMO, this one should add the test case to https://github.com/vllm-project/production-stack not here.

tests/k8s/manifests/sagemaker-hyperpod.yaml
ningziwen: Yes it needs to connect to a daemonset created by Sagemaker Hyperpod cluster.

ningziwen:
@ApostaC I could remove the sagemaker hyperpod tests from this PR for now and only keep basic tests if there is friction for everyone to test it. We can figure out how to add it later.

@DongDongJu This PR would be the base change of LMCache multi process k8s tests, which is not specific to VLLM.

tests/k8s/manifests/sagemaker-hyperpod.yaml
ningziwen: The test README in the PR has all the instructions of setting up and running the tests in SageMaker HyperPod.

PR #2326, opened 2025-12-29
[mp-mode]perform GPU transfers concurrently

wz1qqx:

**If applicable**:


DongDongJu (changes requested):
What I understand is that
currently, in this code, self.lock is only using for the clear KVCache which global(multi gpu) lock needed.
Can we change the name of self.lock to like self.global_lock?
Otherwise, looks good to me.

PR #2329, opened 2025-12-29
Fix LMCache layerwise/blending edge cases and add vLLM patch helper

tunglinwood:
# Problem Description

In LMCache with vLLM v1 and layerwise KV transfer + blending:

- ❌ Layerwise save can throw StopIteration during chunked prefill.
- ❌ Zero-token KV transfers can trigger CUDA invalid configuration errors.
- ❌ Blending can fail when a layer KV buffer is missing or the hit is too small.
- ❌ vLLM API changes (model wrappers, rope signature, attention import path) break
LMCache compute.
- ❌ There is no repo-provided, repeatable way to apply the vLLM GPU worker patch.

### 1. Workflow

```text
Request → vLLM Scheduler → LMCache Integration → Layerwise storer/save → GPU transfer → Model compute → Blending
```

### Layerwise Save 
`lmcache/integration/vllm/vllm_v1_adapter.py`
  The adapter advances a layerwise generator once per layer. If the generator is already
  exhausted, `next()` raises `StopIteration`.

```python
next(layerwise_storer)
```

### Kernel Launch 
`csrc/mem_kernels.cu`
The CUDA grid size is derived from `num_tokens`. If `num_tokens == 0`, the kernel launch
should be skipped to avoid invalid configuration errors.

```c++
const int num_tokens = slot_mapping.size(0);
// launch grid uses num_tokens
```
### Blending 
`lmcache/v1/gpu_connector.py` + integration
Blending paths assume the layer KV buffer is present and the hit is large enough; missing
buffers or tiny hits can cause invalid accesses or unstable behavior.

### Model / RoPE Handling
`lmcache/v1/compute/models/base.py`, `lmcache/v1/compute/positional_encoding.py`
Embedding lookup and RoPE invocation assume a specific vLLM wrapper and function
signature, which changed across vLLM versions.

### Attention Import
`lmcache/v1/compute/attention/*.py`
vLLM moved attention modules; a direct import can break across versions.

```python
from vllm.attention import Attention
```

## 3. Why These Commits Are Needed
- Layerwise saving is a multi‑stage generator; without explicit finalization it can desync under chunked prefill and crash vLLM.
- Zero‑token save/load is a valid path under skip logic and needs safe no‑op behavior.
- Blending needs to handle missing GPU buffers or tiny hits to avoid invalid accesses and broken outputs.
- vLLM refactors changed wrapper/rope/attention APIs; LMCache must adapt to keep layerwise compute working.
- The vLLM GPU worker patch is required for model registration/kv transfer init and need a reproducible script.

## 4. Fix Solution
- Finalize/reset layerwise storers for chunked prefill and guard wait_for_save.
- Add early returns to CUDA kernels when `num_tokens == 0`.
- Isolate GPU buffer mapping per request; skip blending when KV buffers are missing or hits are too small.
- Unwrap vLLM models before embedding, support new rope signature, and add attention import fallbacks.
- Add `scripts/patch_vllm_gpu_worker.py` plus `README` guidance.

### DIFF
--- a/notes/review-2025-12-27.md
+++ b/notes/review-2025-12-27.md
@@ -1,0 +1,12 @@
+# Review Notes - 2025-12-27
+
+Context: LMCache blending changes, positional encoding updates, and related tooling tweaks.
+
+Findings:
+- HIGH: `get_fused_rope` drops `rope_scaling` when using the new `rope_parameters` API. This can produce incorrect RoPE for scaled models. Add `rope_scaling` (and any other required fields) to `rope_params` or keep the old path when scaling is present.
+- MED: `inspect.signature(vllm_get_rope)` is unguarded; if `get_rope` becomes a builtin/pybind function, `inspect.signature` can raise `ValueError` and break blending. Add a try/except and fall back to the legacy path.
+- MED: `LMCBlender.process_qkv` assumes `common_metadata.check_layers` is iterable. Config-only fixes (like `blend_check_layers`) leave other configs vulnerable to `TypeError`. Default to an empty list in code.
+- LOW: `examples/kv_cache_calculator/modelconfig.json` removed GLM entries; the calculator no longer supports those models unless the UI/docs are updated.
+
+Notes:
+- I did not review the full HTML changes in `examples/kv_cache_calculator/kv_cache_calculator.html` beyond the model config update.
--- a/README.md
+++ b/README.md
@@ -75,6 +75,19 @@ Works on Linux NVIDIA GPU platform.
 
 More [detailed installation instructions](https://docs.lmcache.ai/getting_started/installation) are available in the docs, particularly if you are not using the latest stable version of vllm or using another serving engine with different dependencies. Any "undefined symbol" or torch mismatch versions can be resolved in the documentation. 
 
+## Optional vLLM GPU worker patch
+
+Some setups require a small vLLM GPU worker patch for LMCache model tracking.
+After installing vLLM, run this script with the same Python interpreter/venv
+used to launch vLLM:
+
+```bash
+python scripts/patch_vllm_gpu_worker.py
+```
+
+The script creates a `.bak` backup next to the original file and is safe to run
+multiple times. Restart vLLM workers after patching.
+
 ## Getting started
 
 The best way to get started is to checkout the [Quickstart Examples](https://docs.lmcache.ai/getting_started/quickstart/) in the docs.
--- a/csrc/mem_kernels.cu
+++ b/csrc/mem_kernels.cu
@@ -416,6 +416,9 @@ void multi_layer_kv_transfer_templated(
 
   int num_layers = key_value.size(1);
   int num_tokens = slot_mapping.size(0);
+  if (num_tokens == 0) {
+    return;
+  }
   int num_origin_elements = key_value.size(3);
   int elements_per_xword = sizeof(T) / key_value.element_size();
   int num_xwords = num_origin_elements / elements_per_xword;
@@ -526,6 +529,9 @@ void multi_layer_kv_transfer_unilateral(
 
   int num_layers = key_value.size(1);
   int num_tokens = slot_mapping.size(0);
+  if (num_tokens == 0) {
+    return;
+  }
   int num_origin_elements = key_value.size(3);
   int elements_per_qword = 8 / key_value.element_size();
   int num_qwords = num_origin_elements / elements_per_qword;
@@ -607,6 +613,9 @@ void single_layer_kv_transfer(
   int elements_per_entry = 8 / vllm_key_value_cache.element_size();
 
   int num_tokens = slot_mapping.size(0);
+  if (num_tokens == 0) {
+    return;
+  }
   int num_heads;
   int head_size_in_64bit;
   int block_size;
@@ -703,6 +712,9 @@ void load_and_reshape_flash(
   int elements_per_entry = 8 / key_cache.element_size();
 
   int num_tokens = slot_mapping.size(0);
+  if (num_tokens == 0) {
+    return;
+  }
   int num_heads = key_cache.size(2);
   int head_size_in_64bit = key_cache.size(3) / elements_per_entry;
 
@@ -752,6 +764,9 @@ void reshape_and_cache_back_flash(
   int elements_per_entry = 8 / key_cache.element_size();
 
   int num_tokens = slot_mapping.size(0);
+  if (num_tokens == 0) {
+    return;
+  }
   int num_heads = key_cache.size(2);
   int head_size_in_64bit = key_cache.size(3) / elements_per_entry;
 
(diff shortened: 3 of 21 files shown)


DongDongJu:
Hello @tunglinwood, we recognized this issue and worked on it.
If you dont mind, Could you double check [this](https://github.com/DongDongJu/LMCache/tree/bugfix/vllm-layerwise-save-stopiteration) one can solve your issue?

tunglinwood:
> Hello @tunglinwood, we recognized this issue and worked on it. If you dont mind, Could you double check [this](https://github.com/DongDongJu/LMCache/tree/bugfix/vllm-layerwise-save-stopiteration) one can solve your issue?

Hey @DongDongJu, thanks for sharing this — and sorry I should have said upfront: I see that your PR was submitted and reviewed earlier. Appreciate you catching and fixing the layerwise save issue early on.

I’ve taken a closer look at your branch, and it indeed focuses on the layerwise save finalization path (handling StopIteration, draining generators, freeze-mode yields, and req_id tracking), which is an important fix on its own 👍

The PR I’m working on is scoped a bit differently and builds on top of that area: it’s mainly about CacheBlend v1 serving stability (blending guards, per-request GPU buffer mapping, and compatibility with newer vLLM execution / wrapping behavior). It includes a small layerwise safeguard only because CacheBlend depends on the layerwise connector pipeline, but it doesn’t try to replace or supersede the more complete finalization logic you’ve added.

I think the two changes are complementary rather than conflicting. If your PR lands first, I’m very happy to rebase or adapt my changes on top of it, or split things cleanly so:
 - your layerwise save fix stands on its own, and
 - the CacheBlend stability changes follow as a separate PR.

Happy to help test or adjust things either way 😃  — would be great to get both pieces merged cleanly.

DongDongJu:
> > Hello @tunglinwood, we recognized this issue and worked on it. If you dont mind, Could you double check [this](https://github.com/DongDongJu/LMCache/tree/bugfix/vllm-layerwise-save-stopiteration) one can solve your issue?
> 
> Hey @DongDongJu, thanks for sharing this — and sorry I should have said upfront: I see that your PR was submitted and reviewed earlier. Appreciate you catching and fixing the layerwise save issue early on.
> 
> I’ve taken a closer look at your branch, and it indeed focuses on the layerwise save finalization path (handling StopIteration, draining generators, freeze-mode yields, and req_id tracking), which is an important fix on its own 👍
> 
> The PR I’m working on is scoped a bit differently and builds on top of that area: it’s mainly about CacheBlend v1 serving stability (blending guards, per-request GPU buffer mapping, and compatibility with newer vLLM execution / wrapping behavior). It includes a small layerwise safeguard only because CacheBlend depends on the layerwise connector pipeline, but it doesn’t try to replace or supersede the more complete finalization logic you’ve added.
> 
> I think the two changes are complementary rather than conflicting. If your PR lands first, I’m very happy to rebase or adapt my changes on top of it, or split things cleanly so:
> 
> * your layerwise save fix stands on its own, and
> * the CacheBlend stability changes follow as a separate PR.
> 
> Happy to help test or adjust things either way 😃 — would be great to get both pieces merged cleanly.

Sounds great! I will take a look this tomorrow. Thanks!

notes/review-2025-12-27.md
```diff
@@ -0,0 +1,12 @@
+# Review Notes - 2025-12-27
```
DongDongJu, line 1: Please remove this file.

DongDongJu (changes requested):
Hello @tunglinwood,
Could you make the UT for case this PR want to describe?
like below cases, all-True mask, all-False mask, empty tokens, mask length mismatches.

notes/review-2025-12-27.md
tunglinwood: Hello, @DongDongJu Happy New Year 😄  , I will work on it tomorrow. Still on 2026 New Year Holidays.

tunglinwood:
> Hello @tunglinwood, Could you make the UT for case this PR want to describe? like below cases, all-True mask, all-False mask, empty tokens, mask length mismatches.

## Unit Tests Added

- `tests/v1/test_blender_masks.py`: mask handling for blending (all-true, all-false, empty
  tokens, and mask-length mismatch).
- `tests/tools/test_patch_vllm_gpu_worker.py`: vLLM patch script is idempotent and creates a
  backup on first patch.
- `tests/v1/test_layerwise_storers.py`: layerwise storer finalization is safe when
  generators are exhausted (StopIteration guard).
- `tests/v1/test_rope_cache_device.py`: RoPE cache device/dtype matching, including
  `_match_cos_sin_cache_dtype` hook path.
- `tests/v1/test_mem_kernels.py`: empty slot_mapping path in `load_and_reshape_flash` returns
  without error.
- `tests/v1/test_gpu_connector.py`: unshared buffer mapping path in
  `VLLMBufferLayerwiseGPUConnector.batched_to_gpu` (no state leaks, allocator consistent).

### Notes

- Some tests are gated by CUDA and/or vLLM availability and will skip when not present.

### Reviewer Ping

- cc @DongDongJu — would appreciate your review, especially around the layerwise save +
  blending paths.

DongDongJu:
Hello @tunglinwood, it has conflict. please check it.
I will take a look this today afternoon.

tunglinwood:
> Hello @tunglinwood, it has conflict. please check it.
> 
> I will take a look this today afternoon.
> 
> 

Hi， @DongDongJu Have the layerwise problem solved yet? Otherwise I think it is better to break down this PR into some follow-up PRs that can solve individual problems.

rick-heig:
Hello @tunglinwood,

I have been running LMCache from this PR quite extensively to benchmark blending through QA datasets. I came across different responses on the same query when data is cached even as the model temperature is set to 0. Without caching I always get the same answer.

So I tried several different scenarios and made a minimal configuration (no blending) that was

```
max_local_cpu_size: 10
use_layerwise: True
```

And tested again, in some cases when I run the exact same query with temperature=0, I get a different response once the data is cached, so the first answer will be for example "A" and then all subsequent responses will be "B"

If I relaunch LMCache+vLLM and query again I get "A" and then once data is in cache always "B" (and "B" is often garbage output). Without LMCache (vLLM alone) I always get "A".

Then I tried the minimal configuration

```
max_local_cpu_size: 10
```

And I always get "A" as expected.

So it seems the layerwise implementation seems to corrupt the KV cache data somehow.

I use the model [Qwen3-32B-FP8](https://huggingface.co/Qwen/Qwen3-32B-FP8)

vLLM (v.0.13.0) options:
```
LMCACHE_CONFIG_FILE="${CONFIG_FILE}" \
VLLM_SERVER_DEV_MODE=1
vllm serve Qwen3-32B-FP8 \
    --no-enable-prefix-caching \
    --max-model-len 8192 \
    --max-num-batched-tokens=8192 \
    --tensor-parallel-size 2 \
    --gpu-memory-utilization 0.85 \
    --kv-transfer-config \
```

Any tips on how to debug this ?

Best regards,
Rick

PR #2322, opened 2025-12-29
[Controller] Improve heartbeat to controller logic

maobaolong:
**What this PR does / why we need it**:

Before this PR, recent changes to the LMCache Controller introduced a critical issue where heartbeat and p2P lookup operations were executed in different threads/async tasks while sharing the same ZMQ REQ socket. This violated ZMQ's REQ-REP pattern requirement that mandates strict `send() -> recv()` ordering within a single thread.

This PR added dedicated heartbeat socket to solve this issue.

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #2330, opened 2025-12-30
[Controller] Use dealer mode for worker->Controller zmq communication

maobaolong:

As discussed with @ApostaC and @KuntaiDu , I change to use ROUTER/DEALER mode for zmq socket between worker to controller, instead of the REQ/REPLY mode to support asynchronous communication.

If we merged this PR, #2322 can be closed.

**If applicable**:

### DIFF
--- a/lmcache/v1/cache_controller/controllers/registration_controller.py
+++ b/lmcache/v1/cache_controller/controllers/registration_controller.py
@@ -126,7 +126,11 @@ class RegistrationController:
                 "Instance-worker %s already registered, skip registration",
                 (instance_id, worker_id),
             )
-            return RegisterRetMsg(extra_config=extra_config or {})
+            return (
+                RegisterRetMsg()
+                if extra_config is None
+                else RegisterRetMsg(extra_config=extra_config)
+            )
 
         peer_init_url = msg.peer_init_url
         if peer_init_url is None:
@@ -158,7 +162,11 @@ class RegistrationController:
         logger.info(
             "Registered instance-worker %s with URL %s", (instance_id, worker_id), url
         )
-        return RegisterRetMsg(extra_config=extra_config or {})
+        return (
+            RegisterRetMsg()
+            if extra_config is None
+            else RegisterRetMsg(extra_config=extra_config)
+        )
 
     async def deregister(self, msg: DeRegisterMsg) -> None:
         """
--- a/lmcache/v1/cache_controller/controllers/registration_controller.py
+++ b/lmcache/v1/cache_controller/controllers/registration_controller.py
@@ -21,6 +21,7 @@ from lmcache.v1.cache_controller.message import (
     QueryWorkerInfoMsg,
     QueryWorkerInfoRetMsg,
     RegisterMsg,
+    RegisterRetMsg,
 )
 from lmcache.v1.cache_controller.observability import PrometheusLogger
 from lmcache.v1.cache_controller.utils import RegistryTree
@@ -98,9 +99,19 @@ class RegistrationController:
             return QueryInstRetMsg(instance_id=None, event_id=event_id)
         return QueryInstRetMsg(instance_id=instance_node.instance_id, event_id=event_id)
 
-    async def register(self, msg: RegisterMsg) -> None:
+    async def register(
+        self, msg: RegisterMsg, extra_config: Optional[dict[str, str]] = None
+    ) -> RegisterRetMsg:
         """
         Register a new instance-worker connection mapping.
+
+        Args:
+            msg: RegisterMsg from worker
+            extra_config: Optional extra configuration to return to worker,
+                          e.g., {"heartbeat_url": "tcp://...:8082"}
+
+        Returns:
+            RegisterRetMsg with extra_config for worker initialization
         """
         instance_id = msg.instance_id
         worker_id = msg.worker_id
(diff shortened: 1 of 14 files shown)


lmcache/v1/cache_controller/worker.py
```diff
@@ -149,27 +156,58 @@ def __init__(
...
+            worker_id=self.worker_id,
+            ip=self.lmcache_worker_ip,
+            port=self.lmcache_worker_port,
+            peer_init_url=self.p2p_init_url,
         )
 
+        # Send via DEALER socket (empty frame + payload) and wait for response
+        try:
+            await self.req_socket.send_multipart(
+                [b"", msgspec.msgpack.encode(register_msg)], flags=zmq.NOBLOCK
```
DongDongJu: I think timeout based is better for safety than NOBLOCK.

lmcache/v1/cache_controller/controllers/registration_controller.py
```diff
@@ -115,7 +126,7 @@ async def register(self, msg: RegisterMsg) -> None:
                 "Instance-worker %s already registered, skip registration",
                 (instance_id, worker_id),
             )
-            return
+            return RegisterRetMsg(extra_config=extra_config or {})
```
DongDongJu: I recommend to using default_factory=dict then just {}.

DongDongJu (changes requested):
Hello @maobaolong,
IIUC, heartbeat_url looks optional today, but the worker treats it as required (otherwise heartbeats are disabled). Please either make heartbeat_url mandatory (always return it on register) or add a sensible default/fallback to avoid silently turning off heartbeats.
And, I left few questions.

lmcache/v1/cache_controller/controllers/registration_controller.py
maobaolong: Done

lmcache/v1/cache_controller/worker.py
maobaolong: Done

maobaolong:
> IIUC, heartbeat_url looks optional today, but the worker treats it as required (otherwise heartbeats are disabled). Please either make heartbeat_url mandatory (always return it on register) or add a sensible default/fallback to avoid silently turning off heartbeats.
> And, I left few questions.

This is by design. heartbeat_url is intentionally optional because the controller may not always have a dedicated heartbeat port configured.
Here's how it works:
- Controller side (registration_controller.py): Only returns heartbeat_url in extra_config when the controller has controller_urls["heartbeat"] configured.
- Worker side (worker.py:202-209):
```python
heartbeat_url = extra_config.get("heartbeat_url")
if heartbeat_url:
    logger.info("Received heartbeat_url from controller: %s", heartbeat_url)
    self.controller_heartbeat_url = heartbeat_url
    self._create_heartbeat_socket()
else:
    logger.info("No dedicated heartbeat_url provided by controller")
```
When heartbeat_url is not provided, the worker explicitly logs "No dedicated heartbeat_url provided by controller" — this is not silent.
- Heartbeat enable condition (worker.py:412-423):
```python
enable_heartbeat = (
    self.config.lmcache_worker_heartbeat_time is not None
    and self.config.lmcache_worker_heartbeat_time > 0
    and self.heartbeat_socket is not None  # <-- key: socket must exist
)
```
Heartbeat is only enabled when all three conditions are met. If no heartbeat_url is provided, heartbeat_socket remains None, and heartbeat is naturally disabled.


This design allows flexible deployment where heartbeat functionality can be enabled or disabled at the controller level by simply configuring (or not configuring) the heartbeat port.

DongDongJu:
> > IIUC, heartbeat_url looks optional today, but the worker treats it as required (otherwise heartbeats are disabled). Please either make heartbeat_url mandatory (always return it on register) or add a sensible default/fallback to avoid silently turning off heartbeats.
> > And, I left few questions.
> 
> This is by design. heartbeat_url is intentionally optional because the controller may not always have a dedicated heartbeat port configured. Here's how it works:
> 
> * Controller side (registration_controller.py): Only returns heartbeat_url in extra_config when the controller has controller_urls["heartbeat"] configured.
> * Worker side (worker.py:202-209):
> 
> ```python
> heartbeat_url = extra_config.get("heartbeat_url")
> if heartbeat_url:
>     logger.info("Received heartbeat_url from controller: %s", heartbeat_url)
>     self.controller_heartbeat_url = heartbeat_url
>     self._create_heartbeat_socket()
> else:
>     logger.info("No dedicated heartbeat_url provided by controller")
> ```
> 
> When heartbeat_url is not provided, the worker explicitly logs "No dedicated heartbeat_url provided by controller" — this is not silent.
> 
> * Heartbeat enable condition (worker.py:412-423):
> 
> ```python
> enable_heartbeat = (
>     self.config.lmcache_worker_heartbeat_time is not None
>     and self.config.lmcache_worker_heartbeat_time > 0
>     and self.heartbeat_socket is not None  # <-- key: socket must exist
> )
> ```
> 
> Heartbeat is only enabled when all three conditions are met. If no heartbeat_url is provided, heartbeat_socket remains None, and heartbeat is naturally disabled.
> 
> This design allows flexible deployment where heartbeat functionality can be enabled or disabled at the controller level by simply configuring (or not configuring) the heartbeat port.

Thanks for the clarification. 
Basically, avoid the blocking operation in controller side.

DongDongJu (approved):
LGTM

chunxiaozheng (approved):
lgtm

PR #2332, opened 2025-12-30
Introduce a remote config supplier service

maobaolong:

With this PR, we can obtain config from remote config server, thus, we can change backend or remote connector without modify lmcache.

A simple config.
```yaml
chunk_size: 256
local_device: "cpu"
local_cpu: True
max_local_cpu_size: 10
remote_config_url: "http://<LMCache_CONFIG_ADDR>/hcfs_adapter/get_plugin/"
lmcache_app_id: "lmcache_config_001"
```

The config server should reply config json.

E.g.
```json
{"configs":[{"override":true,"key":"lmcache_worker_heartbeat_time","value":"14"},{"override":false,"key":"extra_config","value":"{\"internal_api_server_access_log\": true, \"internal_api_server_log_level\":\"info\", \"save_only_first_rank\": true, \"first_rank_max_local_cpu_size\": 1.1}"}],"createUser":"","maintainer":""}
```

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -1,8 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
-from typing import TYPE_CHECKING, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Optional, Tuple
+import json
 import os
 import threading
+import urllib.error
+import urllib.parse
+import urllib.request
 
 if TYPE_CHECKING:
     from vllm.config import ModelConfig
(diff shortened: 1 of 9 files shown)


lmcache/v1/config.py
```diff
@@ -440,6 +440,26 @@
...
+    "remote_config_url": {
+        "type": Optional[str],
+        "default": None,
+        "env_converter": str,
+        "description": (
+            "URL of the remote configuration service. When set, LMCache will "
+            "fetch additional configuration from this URL at startup."
+        ),
+    },
+    "lmcache_app_id": {
```
KuntaiDu: Since LMCache environment variable already uses `LMCACHE` as prefix, maybe just call it `app_id`?

lmcache/integration/vllm/utils.py
```diff
@@ -29,13 +33,138 @@ def is_false(value: str) -> bool:
     return value.lower() in ("false", "0", "no", "n", "off")
```
KuntaiDu, line 32: Can we put the code change below in a separate file? Feels like functionally this code does not belongs to `lmcache/integration/vllm`.

KuntaiDu (commented):
Functionality LGTM. Some small nits.

lmcache/v1/config.py
maobaolong: Done

lmcache/integration/vllm/utils.py
maobaolong: Done

maobaolong:
@KuntaiDu Thanks for you review, I renamed the lmcache_app_id to `app_id` and move the added methods into `config_base.py`, PTAL.

KuntaiDu (approved):
LGTM!

lmcache/v1/config_base.py
```diff
@@ -578,3 +580,148 @@ def parse_command_line_extra_params(extra_args: list[str]) -> dict[str, Any]:
...
+    try:
+        # Build request payload with current config and env variables
+        payload: dict[str, Any] = {
+            "current_config": config.to_dict(),
+            "env_variables": {},
+        }
+
+        # Add lmcache_appId if provided
+        if app_id:
+            remote_config_url = "%s?appId=%s" % (remote_config_url, app_id)
```
ApostaC: Is there any documentation saying we will try to parse `app_id` from the `remote_config_url` when it follows a certain format?

lmcache/v1/config_base.py
```diff
@@ -578,3 +580,148 @@ def parse_command_line_extra_params(extra_args: list[str]) -> dict[str, Any]:
...
+        logger.warning(
+            "Failed to fetch remote config from %s: %s", remote_config_url, e
+        )
+        return None
+    except json.JSONDecodeError as e:
+        logger.warning("Failed to parse remote config response: %s", e)
+        return None
+    except Exception as e:
+        logger.warning("Unexpected error fetching remote config: %s", e)
+        return None
```
ApostaC: Should we consider using `requests` library to make this part simpler?

ApostaC (commented):
LGTM in general. Can we have some documentation about the newly introduced config fields and the protocol between the config server and the LMCache worker? 

If possible, it would be really helpful if we could have a reference implementation of the config server so that people know how to customize their own ones. We can put that in the `examples` folder.

lmcache/v1/config_base.py
maobaolong: Added a demo

maobaolong:
@ApostaC Thanks for your suggestion, I improved the code and added a demo with relevant `README.md`. PTAL.

lmcache/v1/config_base.py
```diff
@@ -578,3 +579,148 @@ def parse_command_line_extra_params(extra_args: list[str]) -> dict[str, Any]:
...
+        if not override and current_value is not None:
+            logger.info(
+                "Skipping remote config %s (override=False, current value=%s)",
+                key,
+                current_value,
+            )
+            continue
+
+        # Try to convert value to appropriate type
+        if validate_and_set_config_value(config, key, value):
```
chunxiaozheng: if set `extra_config`, this is a full replacement, right?

maobaolong: yes, will support particial replacement
 next.

maobaolong: @chunxiaozheng Done.

maobaolong:
@chunxiaozheng Thanks for your question, now, it supports partial replacement for extra_config. PTAL.

lmcache/v1/config_base.py
```diff
@@ -578,3 +579,176 @@ def parse_command_line_extra_params(extra_args: list[str]) -> dict[str, Any]:
...
+                    merged_value = {**current_value, **value}
+                    setattr(config, config_key, merged_value)
+                    return True
+                # If new value is None or not a dict, keep current value
+                return True
+            # If current value is None or not a dict, just set the new value
+            setattr(config, config_key, value)
+            return True
+
+        setattr(config, config_key, value)
```
chunxiaozheng, line 672: @maobaolong Thanks for your update, i have a minor question, is the `override` only works when `config_key == "extra_config"`, other keys does not care `override`?

maobaolong: Fixed

maobaolong:
@chunxiaozheng Updated the logic, PTAL.

chunxiaozheng (approved):
lgtm

Issue #2294, opened 2025-12-22
[Bug] vllm crashes if retrieve() fails


ziruiliu:
**Symptoms**
For any reason triggered load failure, LMCache will crash at following:
```
(EngineCore_DP0 pid=923955)   File ".../vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 113, in wait_for_save
(EngineCore_DP0 pid=923955)     self._lmcache_engine.wait_for_save()
(EngineCore_DP0 pid=923955)   File ".../LMCache/lmcache/integration/vllm/vllm_v1_adapter.py", line 1309, in wait_for_save
(EngineCore_DP0 pid=923955)     assert len(slot_mapping) == len(token_ids)
(EngineCore_DP0 pid=923955)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=923955) AssertionError
```
As I added some log here, `token_ids` is larger than expected, it was somehow mis-calculated.
The possible fix is treating failed request specially do not go to code in `vllm_v1_adapter.py`
```python
class RequestTracker:
    def update()
...
         self.token_ids.extend(new_token_ids) # This line extends token_ids to unexpected size
```

**To Reproduce**
There are several ways to trigger load failure to trigger this issue:
1. use remote + fs connector, change file's permission to 000, to make LMCache able to lookup but unable to read from cache files.
```yaml
chunk_size: 16
local_cpu: False
max_local_cpu_size: 1
save_unfull_chunk: False
remote_url: "fs://nomatter:0/path/to/cachedir"
remote_serde: "naive"
```
Start vllm
`LMCACHE_CONFIG_FILE=/path/to/fs.yaml vllm serve Qwen/Qwen3-0.6B --no-enable-prefix-caching --enforce-eager --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'`
- Query with a 20-30 word question to generate cache
- Use chmod 000 to cache files
- Repeat the query
- vLLM crashed as retrieve failed. (In earlier version, maybe 0.3.10, vLLM will report failure for this request but will not crash)

2. use remote_backend + fs connector, try to load a larger than max_local_cpu_size cache file to trigger load failure.
- Ensure PYTHONHASHSEED=0 is set
- Query with a 20K token to generate 2GB cache file
- restart vLLM
- Repeat the query to trigger retrieval. You will see assertion failure.

[...]

PR #2336, opened 2026-01-02
[bugfix] Fix slot_mapping assertion error in v1 adapter for preempted requests

### LINKED ISSUE #2327: [bug] v0.3.11 Regression：vLLM Scheduler Allocates Insufficient Blocks, Crashes with AssertionError
## Describe the bug

vLLM workers crash with `AssertionError` in `vllm_v1_adapter.py:1298` during `wait_for_save()` when using LMCache with vLLM v1 API under concurrent load. The vLLM scheduler allocates insufficient blocks for requests, causing a mismatch between `slot_mapping` length (based on allocated blocks) and `token_ids` length (actual tokens).

## To Reproduce

1. Deploy 40 replicas of vLLM with LMCache using the configuration below
2. Send concurrent requests with 100 concurrent users using ~64K context length
3. Observe workers crash within approximately 15 minutes under load

### Environment

- **Image**: `lmcache/vllm-openai:latest` and `lmcache/vllm-openai:v0.3.10` (both affected)
- **Model**: `meta-llama/Llama-3.1-8B-Instruct`
- **vLLM config**: `--enable-prefix-caching --max-model-len 70000 --tensor-parallel-size 4`
- **LMCache config**: Local CPU cache only (`chunk_size: 256, local_cpu: true, max_local_cpu_size: 5.0`)
- **Deployment**: Kubernetes with 40 replicas on ml.g5.24xlarge instances

### Deployment YAML

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: lmcache-benchmark-config
  namespace: lmcache-benchmark
data:
  lmcache.yaml: |
    chunk_size: 256
    local_cpu: true
    max_local_cpu_size: 5.0

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: lmcache-benchmark
  namespace: lmcache-benchmark
spec:
  replicas: 40
  selector:
    matchLabels:
      app: lmcache-benchmark
  template:
    metadata:
      namespace: lmcache-benchmark
      labels:
        app: lmcache-benchmark
    spec:
      nodeSelector:
        node.kubernetes.io/instance-type: ml.g5.24xlarge
      containers:
        - name: vllm
          image: lmcache/vllm-openai:latest
          command:
            - /opt/venv/bin/vllm
            - serve
            - meta-llama/Llama-3.1-8B-Instruct
            - --host
            - 0.0.0.0
            - --port
            - "8000"
            - --enable-prefix-caching
            - --max-model-len
            - "70000"
            - --tensor-parallel-size
            - "4"
            - --kv-transfer-config
            - '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
          resources:
            limits:
              nvidia.com/gpu: "4"
            requests:
              nvidia.com/gpu: "4"
          env:
            - name: LMCACHE_CONFIG_FILE
              value: /etc/lmcache/lmcache.yaml
            - name: LMCACHE_LOG_LEVEL
              value: "DEBUG"
          volumeMounts:
            - name: lmcache-config
              mountPath: /etc/lmcache
      volumes:
        - name: lmcache-config
          configMap:
            name: lmcache-benchmark-config
```

## Expected behavior

vLLM workers should handle requests without crashing. The scheduler should allocate sufficient blocks to match the number of tokens in the request.

## Error Logs

```
(EngineCore_DP0 pid=370) [2025-12-29 02:36:05,912] LMCache ERROR: The number of tokens is more than the number of blocks for request chatcmpl-8c8575c46eae36fe. Something might be wrong in scheduling logic!
(EngineCore_DP0 pid=370) [2025-12-29 02:36:05,912] LMCache ERROR: Num tokens: 46336, num blocks: 2784, block size: 16

(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822] WorkerProc hit an exception.
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822] Traceback (most recent call last):
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822]   File "/opt/venv/lib/python3.12/site-packages/vllm/v1/worker/kv_connector_model_runner_mixin.py", line 134, in _get_kv_connector_output
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822]     kv_connector.wait_for_save()
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822]   File "/opt/venv/lib/python3.12/site-packages/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 121, in wait_for_save
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822]     self._lmcache_engine.wait_for_save()
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822]   File "/opt/venv/lib/python3.12/site-packages/lmcache/integration/vllm/vllm_v1_adapter.py", line 1298, in wait_for_save
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822]     assert len(slot_mapping) == len(token_ids)
(Worker_TP0 pid=504) ERROR 12-29 02:36:05 [multiproc_executor.py:822] AssertionError
```

## Root Cause Analysis

The vLLM scheduler allocates fewer blocks than needed:
- Request has 46,336 tokens
- Allocated blocks: 2,784 × 16 = 44,544 tokens
- Missing: 1,792 tokens

When `wait_for_save()` processes the KV cache, `slot_mapping` (based on allocated blocks) is shorter than `token_ids`, causing the assertion `assert len(slot_mapping) == len(token_ids)` to fail at line 1298 in `vllm_v1_adapter.py`.

## Additional context

**Workaround**: Image `lmcache/vllm-openai:v0.3.9post2` does not exhibit this issue and runs stably under the same load conditions.

**Affected versions**: Both `lmcache/vllm-openai:latest` and `lmcache/vllm-openai:v0.3.10` are affected.

**Working version**: `lmcache/vllm-openai:v0.3.9post2`

ningziwen:
Fixes #2327 

This PR fixes a critical bug in the vLLM v1 adapter that causes worker crashes with `AssertionError: assert len(slot_mapping) == len(token_ids)` when handling large context requests (~64K tokens) under concurrent load. The bug was introduced in v0.3.11 by commit 0a1e001f and manifests as pod crashes in production deployments.

## Root Cause Analysis

The assertion error occurs when `slot_mapping` (derived from allocated blocks) has a different length than `token_ids`. Through extensive debugging and binary search, we identified two root causes:

### Issue 1: Mismatch Between LMCache and vLLM Cached Token Counts

When a request is preempted and resumed, vLLM and LMCache may have cached different amounts of tokens:
- `vllm_cached_tokens`: Tokens in vLLM's local KV cache
- `lmcache_cached_tokens`: Tokens stored in LMCache's cache

By some reason, these values can differ, with `lmcache_cached_tokens > vllm_cached_tokens` in some cases.

**The Problem:**

vLLM's scheduler allocates memory blocks based on `vllm_cached_tokens`:
```
allocated_blocks = (vllm_cached_tokens + new_tokens) / block_size
```

The original code calculated `num_computed_tokens` using the maximum:
```python
num_computed_tokens = max(lmcache_cached_tokens, vllm_cached_tokens)
```

This creates a mismatch when `lmcache_cached_tokens > vllm_cached_tokens`:

**Example:**
- `vllm_cached_tokens = 44,944` → vLLM allocates blocks for 44,944 + 2,042 = 46,986 tokens
- `lmcache_cached_tokens = 45,000` (larger by some reason)
- `new_token_ids = 2,042`

Using `max()`:
- `num_computed_tokens = max(45,000, 44,944) = 45,000`
- `num_tokens_needed = 45,000 + 2,042 = 47,042`
- But vLLM only allocated blocks for 46,986 tokens
- When creating `slot_mapping` from allocated blocks: 46,986 slots
- But `token_ids` has 47,042 tokens
- **Assertion fails**: `assert len(slot_mapping) == len(token_ids)`

**The Fix:**

Since vLLM's scheduler has already made the block allocation decision based on `vllm_cached_tokens`, the adapter must use the same value:

```python
# vLLM allocates blocks based on vllm_cached_tokens
num_computed_tokens = vllm_cached_tokens
```

This ensures `num_tokens_needed` always matches vLLM's actual block allocation.

### Issue 2: Token Count Exceeding Allocated Block Capacity

In `from_request_tracker()`, `num_tokens_to_save` is calculated from `input_token_len`:
```python
if not is_last_prefill or discard_partial_chunks:
    num_tokens_to_save = (input_token_len // chunk_size) * chunk_size
else:
    num_tokens_to_save = input_token_len
```

This value can exceed the allocated block capacity (`num_blocks * block_size`), causing the assertion error.

**When does this happen?**

During preemption with reduced block allocation:
1. A request initially has many tokens (e.g., 43,008 tokens)
2. The request is preempted
3. Upon resumption, vLLM allocates fewer blocks (e.g., 2,576 blocks × 16 = 41,216 tokens)
4. But `input_token_len` is still 43,008 (the full token count)
5. `num_tokens_to_save = 43,008` exceeds allocated capacity of 41,216

**The Fix:**

Cap `num_tokens_to_save` to the allocated block capacity:

```python
# Cap to allocated block capacity to prevent token_ids/slot_mapping mismatch
num_blocks = len(tracker.allocated_block_ids)
max_tokens = num_blocks * block_size
if num_tokens_to_save > max_tokens:
    logger.warning(
        f"Request {tracker.req_id}: num_tokens_to_save ({num_tokens_to_save}) "
        f"exceeds allocated capacity ({max_tokens}). Capping to {max_tokens}."
    )
    num_tokens_to_save = max_tokens
```

This ensures we never try to save more tokens than we have allocated space for.

## Changes Made

1. **`RequestTracker.update()`**: Use `vllm_cached_tokens` instead of `max(lmcache_cached_tokens, vllm_cached_tokens)` to match vLLM's block allocation. Add debug logging for preempted requests.

2. **`ReqMeta.from_request_tracker()`**: Cap `num_tokens_to_save` to allocated block capacity to prevent exceeding available space.

3. **`LMCacheConnectorV1Impl`**: Remove incorrect assertion about `request.num_computed_tokens`.

## Testing

**Validation Environment**:
- 40 replicas on ml.g5.24xlarge (4 GPUs each)
- Model: `meta-llama/Llama-3.1-8B-Instruct`
- Config: `--enable-prefix-caching --max-model-len 70000 --tensor-parallel-size 4`
- Load: 100 concurrent requests with ~64K context

**Results**:
- **Before fix**: Multiple pods crashed within 5-10 minutes with assertion errors
- **After fix**: All 40 pods ran successfully for 6+ hours with no assertion errors or crashes
- **Baseline validation**: Deployed latest dev branch without fixes and confirmed the bug is reproducible

## If applicable

Note: This is a critical bug fix for production stability. Unit tests would require mocking vLLM's scheduler behavior, which is complex. The fix has been validated in production with 40 replicas under heavy load.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -266,17 +266,21 @@ class RequestTracker:
             self.allocated_block_ids = new_block_ids
             # reset the number of saved tokens
             self.num_saved_tokens = lmcache_cached_tokens
-            num_computed_tokens = max(lmcache_cached_tokens, vllm_cached_tokens)
-
-            # FIX: For preempted requests, restore token_ids from the full
-            # token list to ensure chunk keys match what was used during
-            # lookup. The lookup uses request.all_token_ids, so we need the
-            # same tokens for retrieve.
-            num_tokens_needed = max(
-                num_computed_tokens + len(new_token_ids),
-                lmcache_cached_tokens,
-            )
+            # vLLM allocates blocks based on vllm_cached_tokens
+            num_computed_tokens = vllm_cached_tokens
+            # Restore token_ids from full token list for preempted requests
+            num_tokens_needed = num_computed_tokens + len(new_token_ids)
             self.token_ids = all_token_ids[:num_tokens_needed]
+
+            logger.debug(
+                f"Preempted request {self.req_id}: "
+                f"vllm_cached={vllm_cached_tokens}, "
+                f"lmcache_cached={lmcache_cached_tokens}, "
+                f"new_tokens={len(new_token_ids)}, "
+                f"num_tokens_needed={num_tokens_needed}, "
+                f"allocated_blocks={len(new_block_ids)}, "
+                f"token_ids_len={len(self.token_ids)}"
+            )
         else:
             self.allocated_block_ids.extend(new_block_ids)
             self.token_ids.extend(new_token_ids)
@@ -376,6 +380,16 @@ class ReqMeta:
         else:
             num_tokens_to_save = input_token_len
 
+        # Cap to allocated block capacity to prevent token_ids/slot_mapping mismatch
+        num_blocks = len(tracker.allocated_block_ids)
+        max_tokens = num_blocks * block_size
+        if num_tokens_to_save > max_tokens:
+            logger.warning(
+                f"Request {tracker.req_id}: num_tokens_to_save ({num_tokens_to_save}) "
+                f"exceeds allocated capacity ({max_tokens}). Capping to {max_tokens}."
+            )
+            num_tokens_to_save = max_tokens
+
         # If we need to save, update the number of saved tokens
         if not skip_save:
             tracker.num_saved_tokens = num_tokens_to_save
@@ -396,9 +410,7 @@ class ReqMeta:
             )
             token_ids = token_ids.tolist()
 
-        num_blocks = len(tracker.allocated_block_ids)
-
-        if len(token_ids) > num_blocks * block_size:
+        if len(token_ids) > max_tokens:
             logger.error(
                 "The number of tokens is more than the number of blocks"
                 " for request %s. "
@@ -1859,18 +1871,6 @@ class LMCacheConnectorV1Impl:
             if preempted:
                 assert load_spec is not None, (
                     f"Request {req_id} is preempted but was not given a load spec"
-                )
-                # num_computed_tokens should be reset to 0 during preemption
-                # and then set to the number of already cached tokens (maxxing
-                # prefix caching and lmcache)
-                # this assumption is crucial for the update() call of RequestTracker
-                assert request.num_computed_tokens == max(
-                    lmcache_cached_tokens, load_spec.vllm_cached_tokens
-                ), (
-                    f"Preempted request {req_id} has "
-                    f"num_computed_tokens {request.num_computed_tokens} "
-                    "but max(lmcache_cached_tokens, vllm_cached_tokens) = "
-                    f"{max(lmcache_cached_tokens, vllm_cached_tokens)}"
                 )
 
             # Pass all_token_ids for preempted requests to restore


sammshen:
Taking another look tomorrow. Need to revisit the vllm scheduler logic. Mainly checking if the allocated_block_ids includes space allocated for LMCache tokens in `build_connector_meta()`. If so, then that was the previous reasoning for setting `max(vllm_cached_tokens, lmcache_cached_tokens)`

sammshen:
I don't think this PR is right. the block ids passed in should contain what LMCache contains as well (which could be more or less than what is in vllm).

if your PR/fix is correct, the line that should have triggered would be:
```python
assert request.num_computed_tokens == max(lmcache_cached_tokens, load_spec.vllm_cached_tokens)
```

there are two cases that the `max` covers: 
Case 1 (LMCache hit > Local hit): get_num_new_matched_tokens returns the difference. request.num_computed_tokens is updated to local_hits + difference = lmcache_cached_tokens. The assertion holds.
Case 2 (Local hit >= LMCache hit): get_num_new_matched_tokens returns 0. request.num_computed_tokens is updated to local_hits. The assertion holds (since local_hits is the max).

Furthermore, in your test case, the LMCache size is set so small that it seems unlikely that the LMCache hit tokens is larger than what vllm contains (unless the empty GPU HBM space is < 5 GB)

ziruiliu:
I did not dive deep in this case #2327 but I also met the same assertion failure in issue #2294 
In my case, the assertion failure is not because vllm allocates fewer blocks (shorter slot_mapping) but because lmcache wrongly extends token_ids (longer token_ids)
```python
    def update(...)
    ...
            self.token_ids.extend(new_token_ids)
```
Hope this helps for finding out root cause

PR #2337, opened 2026-01-02
[Bugfix] Wrap server kv_chunk read in try/except

mjmac:
Handle missing chunk as cache miss instead
of crashing.

### DIFF
--- a/lmcache/server/server_storage_backend/local_backend.py
+++ b/lmcache/server/server_storage_backend/local_backend.py
@@ -287,11 +287,19 @@ class LMSLocalDiskBackend(LMSBackendInterface):
             return None
 
         path = self.dict[key].path
-        self.evictor.update_on_get(key, self.dict)
 
-        with open(path, "rb") as binary_file:
-            kv_chunk = binary_file.read()
-        self.update_lock.release()
+        kv_chunk = None
+        try:
+            with open(path, "rb") as binary_file:
+                kv_chunk = binary_file.read()
+        except FileNotFoundError:
+            logger.warning(f"File {path} not found. Removing from index.")
+            self.dict.pop(key)
+        else:
+            self.evictor.update_on_get(key, self.dict)
+        finally:
+            self.update_lock.release()
+
         return kv_chunk
 
         # return torch.load(self._key_to_path(key))
--- a/tests/v1/storage_backend/test_local_disk_backend.py
+++ b/tests/v1/storage_backend/test_local_disk_backend.py
@@ -19,6 +19,7 @@ from lmcache.v1.memory_management import (
 )
 from lmcache.v1.storage_backend.local_cpu_backend import LocalCPUBackend
 from lmcache.v1.storage_backend.local_disk_backend import LocalDiskBackend
+from lmcache.server.server_storage_backend.local_backend import LMSLocalDiskBackend
 
 
 class MockLookupServer:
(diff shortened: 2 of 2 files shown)


DongDongJu:
Hello @mjmac, Happy new year.
Thanks for the contribution.
currently we are even not saving meta data as for retrieving.
Could you describe which scenario makes this bug happen?

mjmac:
> Hello @mjmac, Happy new year. Thanks for the contribution. currently we are even not saving meta data as for retrieving. Could you describe which scenario makes this bug happen?

Hi @DongDongJu, happy new year to you, too! I am experimenting with the multi-process server and using an external eviction process to manage the on-disk chunks. The idea behind the patch is that if we delete a chunk from under the server, it should just treat it as a miss instead of crashing. My testing with a mock vllm script seems to validate the approach.

DongDongJu:
@mjmac make sense to me.
Could you make the UT for this case in `tests/v1/storage_backend/test_local_disk_backend.py`?

mjmac:
> Could you make the UT for this case in tests/v1/storage_backend/test_local_disk_backend.py?

@DongDongJu: Sure, done.

lmcache/server/server_storage_backend/local_backend.py
```diff
@@ -289,9 +289,15 @@ def get(
         path = self.dict[key].path
         self.evictor.update_on_get(key, self.dict)
```
DongDongJu: Can we move this line into `finally:` block?
We need to sync with the evictor(which is decision logic for kvcache eviction) side too.

DongDongJu (changes requested):
Hello @mjmac, Happy new year.
Thanks for your contribution.
Thanks for your fast response and I letf one comment.
Otherwise, LGTM

lmcache/server/server_storage_backend/local_backend.py
mjmac: Hmm. Should it be in the finally block or the try block after the open? Seems like we would probably only want to update the evictor info if the file still exists, right? I don't know this codebase very well, though.

DongDongJu: yes, my bad. it should go to within `else:` not `finally`.

mjmac: Sorry, I'm still not clear on what you're asking for. Do you mean an `else:` block paired with the `if key not in self.dict:` block? If so, I think the current code works the same way, right? In other words, the `else:` is not necessary, since the code is executed anyhow. Or do you mean it should go in the `try:` block after the successful open?

I would think that either leaving it as-is or putting it in the `try:` block makes the most sense. Presumably the intent is to inform the evictor of access for the purpose of updating the LRU map. If the chunk doesn't exist anymore, then maybe it should be removed from the LRU entirely? Looking at the current implementations of the evictors, though, I don't see any method for removing an entry.

DongDongJu: Hello @mjmac,
I meant that python has `try, except, else, finally` grammar which is else block execute when try is succeeded.

mjmac: Oh! Huh. I've written a lot of Python over the years, but I don't think I've ever used `else:` with `try:`. OK, moved it into a new `else:` block.

DongDongJu (approved):
LGTM

mjmac:
@DongDongJu Hi, thanks for the LGTM. Do I need to request a review from someone else with merge privileges?

DongDongJu:
Hello @YaoJiayi, Could you take a look this PR when you have a time?

DongDongJu:
Hello @mjmac, after second look, this is the code piece not using that much at this moment.
Most of the stack currently working under lmcache/v1 folder structure.
I will working on remove the depreciated code by refactor the folder structure changes.
Thanks for your contribution!

mjmac:
Hi @DongDongJu. I understand that the server is still experimental, but it will eventually become production-ready, right? Or are you saying that this code is all deprecated and that the server will use the other code?

DongDongJu:
Hello @mjmac. most recent works are here. https://github.com/LMCache/LMCache/tree/dev/lmcache/v1/storage_backend.
I will suggest the refactor cleanup in next community meeting.

Issue #2257, opened 2025-12-16
A united way to maintain the states of LMCache


maobaolong:
**Label**
Please label your issue so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/1882)

**Summary**

As proposed in https://github.com/LMCache/LMCache/pull/2245#discussion_r2621207011 , we need a class to maintains some status or boolean type variable to alternate the behavior of LMCache in runtime via `internal_api_server`.

maobaolong:
<img width="1314" height="662" alt="Image" src="https://github.com/user-attachments/assets/768de65b-ad92-4d91-95d7-afb0d577d125" />

sammshen:
This is great.

to summarize the two components involved in the adapter refactor:

1. refactor the lmcache side interface for the adapter into a LMCacheManager
2. refactor the vllm side interface for the adapter (e.g. encapsulate state exposure from the `Request` data structure from vllm to be resilient to changes in the future)

maobaolong:
@sammshen Would you like to draw a digram to show the whole picture looks after your refactor?

maobaolong:
```mermaid
graph TB
    subgraph InferenceSide [Inference side]
        subgraph vLLM [vLLMServices]
            vLLM_Service[vLLM Service]
            vLLM_Lookup[LookupClient/Server]
        end
        
        subgraph SGLang [SGLangServices]
            SGLang_Service[SGLang Service]
        end
        
        Common[CommonServices]
    end

    subgraph LMCacheSide [LMCache Side]
        LMCacheEngine[LMCacheEngine]
        GpuConnector[GpuConnector]
        LMCacheManager[LMCacheManager]
        PluginLauncher[PluginLauncher]
        InternalApiServ[InternalApiServ]
    end

    %% 连接关系
    vLLM --> Common
    SGLang --> Common
    Common --> LMCacheEngine
    
    LMCacheEngine --> GpuConnector
    LMCacheEngine --> LMCacheManager
    LMCacheEngine --> PluginLauncher
    LMCacheEngine --> InternalApiServ
```
Something like this.

PR #2348, opened 2026-01-05
Refactor to abstract lmcacheManager from vllm_v1_adapter.py

maobaolong:

This PR is lmcache part of #2257

**If applicable**:

### DIFF
--- a/lmcache/v1/manager.py
+++ b/lmcache/v1/manager.py
@@ -410,7 +410,15 @@ class LMCacheManager:
         return not self._config.enable_pd
 
     def start_services(self) -> None:
-        """Start all managed services."""
+        """
+        Start all managed services.
+
+        Managed services include:
+        - InternalAPIServer: HTTP server exposing internal APIs for
+          monitoring and management (e.g., cache stats, flush operations).
+        - RuntimePluginLauncher: Launches external plugin processes defined
+          in the configuration (e.g., custom telemetry, cache warming scripts).
+        """
         if self._api_server is not None:
             self._api_server.start()
 
--- a/lmcache/v1/manager.py
+++ b/lmcache/v1/manager.py
@@ -310,19 +310,6 @@ class LMCacheManager:
         ):
             raise ValueError("MLA only works with naive serde mode..")
 
-        if use_mla and not self._config.save_unfull_chunk:
-            logger.warning(
-                "MLA (Multi-Level Attention) requires save_unfull_chunk=True "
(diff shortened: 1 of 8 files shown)


maobaolong:

lmcache/v1/manager.py
```diff
@@ -0,0 +1,649 @@
...
+        """Get the API server instance."""
+        return self._api_server
+
+    @property
+    def config(self) -> LMCacheEngineConfig:
+        """Get the LMCache engine configuration."""
+        return self._config
+
+
+class LMCacheManagerBuilder:
```
sammshen: the builder pattern doesn't seem like a good choice for the construction of an `LMCacheManager` because: 
1. the builder is stateless (so what's the point?)
2. the argument order is not relevant

lmcache/v1/manager.py
```diff
@@ -0,0 +1,649 @@
...
+
+        # Initialize components based on role
+        self._init_components()
+
+    @classmethod
+    def builder(cls) -> "LMCacheManagerBuilder":
+        """Create a builder for constructing LMCacheManager instances."""
+        return LMCacheManagerBuilder()
+
+    def _init_components(self) -> None:
```
sammshen, line 96: this function is fantastic

lmcache/v1/manager.py
```diff
@@ -0,0 +1,649 @@
...
+            return VLLMPagedMemXPUConnectorV2.from_metadata(metadata, use_gpu, device)
+        else:
+            raise RuntimeError("No supported connector found for the current platform.")
+
+    def _need_gpu_interm_buffer(self) -> bool:
+        """Check if GPU intermediate buffer is needed."""
+        return not self._config.enable_pd
+
+    def start(self) -> None:
+        """Start all managed services."""
```
sammshen: nit: rename to `start_plugins` and add some clearer definition to the doc string on what a managed service / plugin is

sammshen (approved):
LGTM! I will work on the vllm side encapsulation after this

lmcache/v1/manager.py
maobaolong: Revert builder mode

lmcache/v1/manager.py
maobaolong: Done

maobaolong:
@ApostaC Would you like to take a double check? Thanks!

chunxiaozheng (approved):
lgtm

PR #2354, opened 2026-01-05
[Nit]: update byte_array property of TensorMemoryObj

sammshen:
Correctly type the byte_array property and leave two comments about alternative implementations, which can be substituted once all dependencies on the property are checked.

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -555,7 +555,18 @@ class TensorMemoryObj(MemoryObj):
         )
 
     @property
-    def byte_array(self) -> bytes:
+    def byte_array(self) -> memoryview:
+        # TODO: consider using one of the alternatives
+
+        # Alternative 1:
+        # # PyTorch tensors support buffer protocol directly for CPU tensors
+        # return memoryview(self.raw_data)
+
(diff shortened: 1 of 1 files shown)


lmcache/v1/memory_management.py
```diff
@@ -555,7 +555,18 @@ def tensor(self) -> Optional[torch.Tensor]:
         )
 
     @property
-    def byte_array(self) -> bytes:
+    def byte_array(self) -> memoryview:
+        # TODO: consider using one of the alternatives
+
+        # Alternative 1:
+        # # PyTorch tensors support buffer protocol directly for CPU tensors
+        # return memoryview(self.raw_data)
```
DongDongJu, line 563: Is every torch version can support this?

DongDongJu (approved):
LGTM at this point.
just left one question about alternative.

kobe0938 (approved):

lmcache/v1/memory_management.py
sammshen: Good question. at least as of pytorch 1.9, it is supported according to: https://pearu.github.io/array_interface_pytorch.html?utm_source=chatgpt.com

PR #2369, opened 2026-01-08
fix: tensor shape mismatch while layerwise store

blossomin:

**If applicable**:


blossomin:
the bug was previously discussed here:
https://github.com/LMCache/LMCache/issues/832#issuecomment-3722897415

blossomin:
@YaoJiayi @sammshen  , could you have a look at this? thank!

PR #2370, opened 2026-01-08
[Feat] staging buffer mode when local_cpu is false for diskbackend

DongDongJu:

This PR introduces a new `use_only_staging_buffer` configuration flag that enables disk-only caching mode in LMCache. 
When enabled with `local_cpu=false`, CPU memory is used **only** as a temporary staging buffer for GPU-to-disk transfers, and all cache lookups go directly to disk.

### Problem

Currently, even with `local_cpu=false`, CPU memory still participates in cache lookups. This means:
- Data retrieved from disk is written back to CPU cache
- Subsequent lookups hit CPU cache instead of disk
- CPU memory usage grows over time with cached data

This behavior is problematic when users want pure disk-only caching for scenarios like:
- Limited CPU memory environments
- Testing disk backend performance in isolation
- Ensuring data persistence on disk without CPU cache interference

### Solution

Add `use_only_staging_buffer` flag that, when combined with `local_cpu=false`:
1. Skips CPU backend in lookups - Cache lookups go directly to disk
2. Disables write-back - Data retrieved from disk is not cached in CPU
3. Releases staging buffer after disk write - CPU memory is freed after GPU→Disk transfer completes

### Data Flow Comparison

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                    BEFORE (local_cpu=false only)                            │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│  PUT (Store KV Cache):                                                      │
│  ┌─────┐    ┌─────────────┐    ┌──────┐                                     │
│  │ GPU │───►│ CPU (stage) │───►│ Disk │                                     │
│  └─────┘    └─────────────┘    └──────┘                                     │
│                   │                                                         │
│                   └── CPU keeps data (memory grows)                         │
│                                                                             │
│  GET (Lookup & Retrieve):                                                   │
│  ┌──────┐    ┌─────────────┐    ┌─────┐                                     │
│  │ Disk │───►│ CPU (cache) │───►│ GPU │                                     │
│  └──────┘    └─────────────┘    └─────┘                                     │
│                   │                                                         │
│                   └── Write-back to CPU (unexpected CPU hits later)         │
│                                                                             │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────────────┐
│           AFTER (local_cpu=false + use_only_staging_buffer=true)            │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│  PUT (Store KV Cache):                                                      │
│  ┌─────┐    ┌─────────────┐    ┌──────┐                                     │
│  │ GPU │───►│ CPU (stage) │───►│ Disk │                                     │
│  └─────┘    └─────────────┘    └──────┘                                     │
│                   │                                                         │
│                   └──────────────────── CPU buffer released after write     │
│                                                                             │
│  GET (Lookup & Retrieve):                                                   │
│  ┌──────┐    ┌─────────────┐    ┌─────┐                                     │
│  │ Disk │───►│ CPU (stage) │───►│ GPU │                                     │
│  └──────┘    └─────────────┘    └─────┘                                     │
│                   │                                                         │
│                   └── No write-back, staging only                           │
│                                                                             │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

### Configuration Comparison

| Configuration | CPU in Lookups | Behavior |
|---------------|----------------|----------|
| `local_cpu: true` | Yes | Normal CPU+Disk tiered caching |
| `local_cpu: false` | Yes | CPU still participates in lookups (legacy) |
| `local_cpu: false` + `use_only_staging_buffer: true` | **No** | **Disk-only caching** (CPU is staging buffer only) |

### Usage

**Environment Variables:**
```bash
export LMCACHE_LOCAL_CPU=false
export LMCACHE_USE_ONLY_STAGING_BUFFER=true
```

**Configuration File:**
```yaml
local_cpu: false
use_only_staging_buffer: true
local_disk: "file:///path/to/disk/cache/"
max_local_disk_size: 100.0
```

1. Default value is `false`, so existing behavior is unchanged
2. Used completion callback pattern to release staging buffer after async disk write completes, avoiding race conditions

**If applicable**:

### DIFF
--- a/docs/source/kv_cache/storage_backends/local_storage.rst
+++ b/docs/source/kv_cache/storage_backends/local_storage.rst
@@ -64,6 +64,61 @@ backends have asynchronous put() operations so that the IO latency will not slow
 The local disk backend also has a prefetch() operation that will preemptively move KV caches from the disk to CPU RAM offloading storage
 (i.e. ``LMCACHE_LOCAL_CPU=True`` should be set, see :doc:`CPU RAM <./cpu_ram>`) for specified tokens (these KV caches are also still kept in the disk).
 
+.. _staging-buffer-mode:
+
+Staging Buffer Mode (Disk-Only Caching)
+---------------------------------------
+
+When you want to use **only disk** for KV cache storage (without CPU caching), you need to enable
+the ``use_only_staging_buffer`` flag along with ``local_cpu: false``.
+
+**Why is this needed?**
+
+Even with ``local_cpu: false``, CPU memory is still required as a **staging buffer** for data transfers
+between GPU and disk. By default, this staging buffer also participates in cache lookups, which means
+data might be served from CPU memory instead of disk.
+
+With ``use_only_staging_buffer: true``, the CPU memory is used **only** as a temporary staging buffer:
+
+- Cache lookups skip the CPU backend entirely
+- Data is always retrieved from disk
+- No write-back to CPU after disk retrieval
+- CPU memory usage is minimized
+
+**Configuration comparison:**
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 25 50
+
+   * - Configuration
+     - CPU in Lookups
+     - Behavior
+   * - ``local_cpu: true``
+     - Yes
+     - Normal CPU+Disk tiered caching
+   * - ``local_cpu: false``
+     - Yes
+     - CPU still participates in lookups (default behavior)
+   * - | ``local_cpu: false``
+       | ``use_only_staging_buffer: true``
+     - No
+     - **Disk-only caching** (CPU is staging buffer only)
+
+**Environment Variables:**
+
+.. code-block:: bash
+
+    export LMCACHE_LOCAL_CPU=false
+    export LMCACHE_USE_ONLY_STAGING_BUFFER=true
+
+**Configuration File:**
+
+.. code-block:: yaml
+
+    local_cpu: false
+    use_only_staging_buffer: true
+
 .. _local-storage-online-inference-example:
 
 Online Inference Example
@@ -114,7 +169,7 @@ GPU memory and LMCache is necessary to keep KV caches in non-GPU memory:
 
 **Step 2. Start a vLLM server with Disk offloading enabled:**
 
-*Generally, it is not recommended but we will disable CPU offloading to feel just the disk offloading latency.*
+*Generally, it is not recommended but we will disable CPU caching to feel just the disk offloading latency.*
 
 Create a an lmcache configuration file called: ``disk-offload.yaml``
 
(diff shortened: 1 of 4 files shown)


docs/source/kv_cache/storage_backends/local_storage.rst
ApostaC, line 1: Nice documentation 👍

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -292,6 +292,7 @@ def submit_put_task(
         self,
         key: CacheEngineKey,
         memory_obj: MemoryObj,
+        on_complete_callback: Optional[Callable[[CacheEngineKey], None]] = None,
```
ApostaC, line 295: I think this function is inherited from the base class. Do we want to add the argument into the base class as well?

ApostaC: Additionally, we probably need some clarification in the doc string about when the callback will be triggered (e.g., after each object has finished putting, or after all the objects have finished putting).

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -412,7 +429,21 @@ def batched_put(
...
+            ):
+                disk_backend = cast("LocalDiskBackend", backend)
+                disk_backend.batched_submit_put_task(
+                    ks,
+                    objs,
+                    transfer_spec=transfer_spec,
+                    on_complete_callback=staging_buffer_callback,
+                )
+            else:
+                backend.batched_submit_put_task(ks, objs, transfer_spec=transfer_spec)
```
ApostaC, line 446: Will the same callback abstraction work for other L2 storage backends?

lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -433,9 +464,14 @@ def get(
...
             memory_obj = backend.get_blocking(key)
             if memory_obj:
+                # Write-back to CPU cache (skip if use_only_staging_buffer is enabled)
                 if (
                     backend_name not in ["LocalCPUBackend", "PDBackend"]
                     and "LocalCPUBackend" in self.storage_backends
+                    and not (
+                        not self.config.local_cpu
+                        and self.config.use_only_staging_buffer
+                    )
```
ApostaC, line 474: This part of the code seems to appear multiple times. Maybe consider having a helper function like `self._should_write_back_to_cpu()`

ApostaC (changes requested):
Did a quick review. Please see the details, thanks!

ApostaC:
Actually, after a second thought, I started to think about whether we need to fully skip the CPU buffer.

At a high level, we should recommend people to use async loading when the KV cache is from L2 storage, like disk or remote. The async loading will first load the KV cache from L2 to L1 (CPU buffer), and then vLLM will load the KV cache to the GPU.

In general, I feel like `use_only_staging_buffer` is not compatible with the async mode (I could be wrong as well). That said, I'm still okay to proceed if the goal of this mode is to help with debugging and development.

lmcache/v1/storage_backend/local_disk_backend.py
DongDongJu: I do agree. will do

lmcache/v1/storage_backend/storage_manager.py
DongDongJu: lemme check

lmcache/v1/storage_backend/storage_manager.py
DongDongJu: will do

DongDongJu:
> Actually, after a second thought, I started to think about whether we need to fully skip the CPU buffer.
> 
> At a high level, we should recommend people to use async loading when the KV cache is from L2 storage, like disk or remote. The async loading will first load the KV cache from L2 to L1 (CPU buffer), and then vLLM will load the KV cache to the GPU.
> 
> In general, I feel like `use_only_staging_buffer` is not compatible with the async mode (I could be wrong as well). That said, I'm still okay to proceed if the goal of this mode is to help with debugging and development.

Hello @ApostaC, Thanks for the quick review.
I didnt thought about async mode. Let me complete to including with that one's behavior as well.

DongDongJu:
After talking with @YaoJiayi and @ApostaC, I will start the make abstraction for put completion for every backend fisrt. And I will revisit to here.

PR #2372, opened 2026-01-08
[Test] Skip test_eic.py when eic module import fails

DongDongJu:

This PR adds a conditional skip for `tests/v1/storage_backend/test_eic.py` when the `eic` module is not available.

- Wrap the `eic` import in a try-except block to handle ImportError
- Use `pytest.mark.skipif` to skip all tests in the module when `eic` is not installed

The `eic` module is an optional dependency that may not be available in all test environments. 
Without this change, pytest would always fail with an ImportError when collecting tests.

**If applicable**:


maobaolong (commented):
LGTM

maobaolong (approved):

sammshen (approved):
LGTM!

Issue #2374, opened 2026-01-09
[BUG] The multi_layer_kv_transfer function has an incorrect addressing calculation


zhangrongguo:
**Describe the bug**
I tested the mistralai/Mistral-7B-Instruct-v0.2 model and printed the following logs:
layer_name: model.layers.0.self_attn.attn
self.kv_caches[layer_name].shape: torch.Size([2, 7286, 64, 8, 128]), stride: (477495296, 65536, 128, 8192, 1)

For linear addressing in the data copy function (multi_layer_kv_transfer), a contiguous memory layout is required. From the above shape and strides, we can infer that the dimension order of the contiguous data should be shape[0] -> shape[1] -> shape[3] -> shape[2] -> shape[4].
Therefore, the current implementation of the multi_layer_kv_transfer function is incorrect, and I have fixed this bug accordingly.
In addition, I have also conducted tests on the following models:
Qwen/Qwen3-32B
deepseek-ai/DeepSeek-V2-Lite-Chat

**To Reproduce**
1. run vLLM prefill worker with --kv-transfer-config {"kv_connector":"MultiConnector","kv_role":"kv_both","kv_connector_extra_config":{"connectors":[{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}, {"kv_connector":"NixlConnector","kv_role":"kv_producer"}]}}
2. run vLLM decode worker with --kv-transfer-config {"kv_connector":"NixlConnector","kv_role":"kv_consumer"}
4. send chat completion request to prefill worker
5. check the shape and strides of self.kv_caches

**Environments (please complete the following information):**
 - OS: Ubuntu 24.04.1 LTS
 - LMCache: v0.3.7 or dev branch
 - vLLM v0.11.0

ApostaC:
Can you try if the issue is still there without the NIXL connector?

PR #2375, opened 2026-01-09
fix: 'the multi_layer_kv_transfer function has an incorrect addressing calculation' issue

zhangrongguo:
Fix issue: https://github.com/LMCache/LMCache/issues/2374

I tested the mistralai/Mistral-7B-Instruct-v0.2 model and printed the following logs:
layer_name: model.layers.0.self_attn.attn
self.kv_caches[layer_name].shape: torch.Size([2, 7286, 64, 8, 128]), stride: (477495296, 65536, 128, 8192, 1)

For linear addressing in the data copy function (multi_layer_kv_transfer), a contiguous memory layout is required. From the above shape and strides, we can infer that the dimension order of the contiguous data should be shape[0] -> shape[1] -> shape[3] -> shape[2] -> shape[4].
Therefore, the current implementation of the multi_layer_kv_transfer function is incorrect, and I have fixed this bug accordingly.
In addition, I have also conducted tests on the following models:
Qwen/Qwen3-32B
deepseek-ai/DeepSeek-V2-Lite-Chat

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Harden Runner"
-        uses: step-security/harden-runner@95d9a5deda9de15063e7595e9719c11c38c90ae2 # v2.13.2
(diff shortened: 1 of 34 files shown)


zhangrongguo:
I will open a new PR for this.

PR #2376, opened 2026-01-09
Remove vllm deps from lookup client/server

maobaolong:

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -204,14 +204,19 @@ def create_lmcache_metadata(
     head_size = model_cfg.get_head_size()
     kv_shape = (num_layer, 1 if use_mla else 2, chunk_size, num_kv_head, head_size)
 
-    # Extract engine_id from vllm_config if available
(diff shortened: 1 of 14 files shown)


maobaolong:

lmcache/integration/vllm/utils.py
```diff
@@ -204,6 +204,13 @@ def create_lmcache_metadata(
     head_size = model_cfg.get_head_size()
     kv_shape = (num_layer, 1 if use_mla else 2, chunk_size, num_kv_head, head_size)
 
+    # Extract engine_id from vllm_config if available
```
sammshen: what cases is the `engine_id` unavailable?

sammshen (approved):
LGTM! This is great!

lmcache/integration/vllm/utils.py
maobaolong: engine_id is introduced by https://github.com/vllm-project/vllm/pull/17751 , before this PR, there is no `engine_id` within `KVTransferConfig`

chunxiaozheng (approved):
lgtm

PR #2391, opened 2026-01-12
Introduce RECOMPUTE and LOCAL_CPU FallbackPolicy

maobaolong:

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/storage_manager.py
+++ b/lmcache/v1/storage_backend/storage_manager.py
@@ -275,6 +275,11 @@ class StorageManager:
         self._freeze = False
         self._freeze_lock = threading.RLock()
 
+        # Backend bypass mode: skip specific backends during health check failures
+        # Key: backend_name, Value: True if bypassed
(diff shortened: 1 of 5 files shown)


lmcache/v1/storage_backend/storage_manager.py
```diff
@@ -275,6 +275,11 @@ def __init__(
         self._freeze = False
         self._freeze_lock = threading.RLock()
 
+        # Backend bypass mode: skip specific backends during health check failures
+        # Key: backend_name, Value: True if bypassed
+        self._bypassed_backends: Dict[str, bool] = {}
```
sammshen: a dict with bool values in this case can be replaced by set (based on teh byass semantics below). correct me if I'm wrong

sammshen (approved):
lgtm

lmcache/v1/storage_backend/storage_manager.py
maobaolong: Done. Convert to set

chunxiaozheng (approved):
lgtm

PR #2393, opened 2026-01-12
[Refactor] adds a unified on_complete_callback abstraction to all storage backend interfaces.

DongDongJu:

This PR adds a unified `on_complete_callback` abstraction to all storage backend interfaces.
This enables callers to perform cleanup or notification actions after a put(write) operation completes, 
uniformly across all backend types.

### Problem

Currently, there is no standard way to get notified when a storage backend finishes persisting data.
This makes it difficult to implement features like:
- Releasing staging buffers after disk write completes
- Tracking completion of async operations
- Implementing write-through caching patterns

### Solution

Add `on_complete_callback: Optional[Callable[[CacheEngineKey], None]] = None` parameter to:
- `batched_submit_put_task()` in base interface
- `async_batched_submit_put_task()` in base interface
- All backend implementations

### Callback Contract

```
┌─────────────────────────────────────────────────────────────────┐
│                    Callback Invocation Flow                     │
├─────────────────────────────────────────────────────────────────┤
│                                                                 │
│  batched_submit_put_task([key1, key2, key3], [...], callback)   │
│                           │                                     │
│                           ▼                                     │
│              ┌────────────────────────┐                         │
│              │   Backend Processing   │                         │
│              └────────────────────────┘                         │
│                           │                                     │
│           ┌───────────────┼───────────────┐                     │
│           ▼               ▼               ▼                     │
│     key1 done       key2 done       key3 done                   │
│         │               │               │                       │
│         ▼               ▼               ▼                       │
│   callback(key1)  callback(key2)  callback(key3)                │
│                                                                 │
│   * Each callback is invoked AFTER that key's write completes   │
│   * Exceptions in callbacks are caught and logged               │
│   * Callbacks run in the backend's IO context                   │
│                                                                 │
└─────────────────────────────────────────────────────────────────┘

### Example Usage

```python
def release_staging_buffer(key: CacheEngineKey) -> None:
    """Called after each key is persisted to disk."""
    cpu_backend.remove(key)

# Pass callback to disk backend
disk_backend.batched_submit_put_task(
    keys=keys,
    objs=memory_objs,
    on_complete_callback=release_staging_buffer,
)
```

- Default value is `None`, so existing code works unchanged
- it all passed storage backend tests.
- Backends that cannot use callback simply ignore it
- This PR is a prerequisite for the staging buffer mode feature (#2370)
- The callback is invoked **per key**, not per batch, as clarified in docstrings
- All callbacks are wrapped in try/except to prevent backend crashes
- NIXL async mode callback support is bit hacky now. So it is marked as TODO for future work

**If applicable**:


ApostaC (approved):
LGTM! Thanks for adding this.

A side note, in my understanding, one of the goals of the async programming framework is to avoid the "callback hell". Right now, I feel like we are mixing callbacks, asyncio, python threads (concurrent.futures) together, which may create confusion and performance degradation sometimes. I'm trying to think about a cleaner abstraction and will try to implement that in the new multiprocessor storage manager.

sammshen:
LGTM too!

sammshen (approved):

sammshen:
correctness flaky will be fixed by https://github.com/LMCache/LMCache/pull/2474. might need to rebase sorry!

Issue #2379, opened 2026-01-10
[Core] Reimplement lazy memory allcator


ApostaC:
**Summary**
Right now, the lazy memory allocator has a few problems. One major problem is that it will stop the world when allocating pin memory in the background.

This issue is used to track the effort of reimplementing the lazy memory allocator.

- [x] Refactor the current TensorMemoryAllocator to support `sbrk()` (#2380)
- [x] Reimplement lazy memory allocator using cudaHostRegister (#2407)
- [x] Add kernel support for lazy memory allocator (#2413 )
- [ ] Integrate lazy memory allocator with normal mode and multi-process mode
  - [ ] Integrate with multi-process mode (#2463)
  - [ ] Integrate with normal mode
- [ ] End-to-end test to verify the performance

PR #2407, opened 2026-01-13
[Core] Introduce lazy memory allocator

ApostaC:

This PR is part of #2379 

This PR introduces a new class: `LazyMemoryAllocator`. It will first allocate 

A few follow-up PRs:
- [ ] Integrate the new LazyMemoryAllocator with normal mode and MP mode
- [ ] End-to-end test to make sure the performance is good

**If applicable**:


lmcache/v1/lazy_memory_allocator.py
```diff
@@ -0,0 +1,233 @@
...
+        self._expand_thread.join()
+
+        # Unpin all pinned memory chunks
+        for ptr, size in self._pin_record:
+            self._cudart.cudaHostUnregister(ptr)
+        self._pin_record.clear()
+
+        # Free the underlying buffer if using NUMA allocation
+        if self._use_numa:
+            lmc_ops.free_numa_ptr(self._buffer.data_ptr())
```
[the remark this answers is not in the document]
ApostaC: good catch

lmcache/v1/memory_management.py
```diff
@@ -1100,7 +1100,20 @@ def __init__(
         self,
         tensor: torch.Tensor,
         align_bytes: int = AddressManager.ALIGN_BYTES,
+        init_address_space: int | None = None,
     ):
```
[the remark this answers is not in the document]
ApostaC: good catch

lmcache/v1/lazy_memory_allocator.py
```diff
@@ -0,0 +1,233 @@
...
+        numa_mapping (NUMAMapping): The NUMA mapping object.
+
+    Returns:
+        int: The NUMA ID for the current GPU.
+
+    Raises:
+        ValueError: If GPU id is not detected in the numa mapping.
+    """
+    gpu_id = torch.cuda.current_device() if torch.cuda.is_available() else 0
+    return numa_mapping.gpu_to_numa_mapping[gpu_id]
```
maobaolong, line 40: Nit: It would raise `KeyError` but not `valueError` which is inconsistent to the doc string motioned.

lmcache/v1/lazy_memory_allocator.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def align_to(size: int, align_size: int) -> int:
```
maobaolong, line 43: Nit: We can put these two above generic tool methods into a utils class.

tests/skipped/test_lazy_allocator_integration.py
```diff
@@ -1,94 +0,0 @@
-# SPDX-License-Identifier: Apache-2.0
-"""Integration tests for LazyMixedMemoryAllocator with LocalCPUBackend"""
-
```
maobaolong, line 3: Why these two test file renamed and removed? It would be help if we have a non-gpu tests about lazy_memory_allocator.

maobaolong (commented):
@ApostaC Thanks for the improvement about the lazy memory allocator, left some nit and a question about removed test files.

lmcache/v1/lazy_memory_allocator.py
ApostaC: Good catch, it's a typo here. Will fix it!

lmcache/v1/lazy_memory_allocator.py
ApostaC: I would like to keep them like "helper functions" that will only be used in this source file. It's more like "static functions" in C.
Having a class for them seems to be overkill. What do you think?

tests/skipped/test_lazy_allocator_integration.py
ApostaC: The problem is `LazyMemoryAllocator` needs to call cudaHostRegister to allocate pinned memory.
"Allocating pinned memory" is not well-defined when there is no GPU. 

Additionally, we don't need LazyMemoryAllocator when there is no GPU, because the problem that "allocating pinned memory is very slow" does not exist in this case.

That's why I skipped the CPU-only UT for `LazyMemoryAllocator`.

ApostaC: The new UT to test the integration between `LazyMemoryAllocator` and `LocalCPUBackend` will be added in the next PR that does the real integration.

The reason why I removed this file here is that it makes the linter unhappy... And we will reimplement the UT anyway because the previous one has a bunch of assertions on the internal fields of the old LazyMemoryAllocator (where the new one does not have)

lmcache/v1/lazy_memory_allocator.py
ApostaC: I do realize that there are some other places in the code has very similar functions. But in order to keep the changes minimal in this PR, I just did this separately.
We can have a refactor after this PR to refactor and move functions to utils module if needed.

maobaolong (approved):
Thanks for your explain, it LGTM.

sammshen (approved):

lmcache/v1/lazy_memory_allocator.py
```diff
@@ -0,0 +1,233 @@
...
+        assert offset & (self.PIN_CHUNK_SIZE - 1) == 0, (
+            "Offset must be aligned to PIN_CHUNK_SIZE"
+        )
+        assert size & (self.PIN_CHUNK_SIZE - 1) == 0, (
+            "Size must be aligned to PIN_CHUNK_SIZE"
+        )
+        assert offset + size <= self._final_size, "Pinning exceeds buffer size"
+
+        ptr = self._buffer.data_ptr() + offset
+        self._cudart.cudaHostRegister(ptr, size, 0)
```
DongDongJu, line 204: I think this function can fail under memory pressure or cgroup mem limit.  What do you think?

lmcache/v1/lazy_memory_allocator.py
```diff
@@ -0,0 +1,233 @@
...
+        while self._curr_size < self._final_size and not self._stop_expand.is_set():
+            # Expand chunk by chunk and commit
+            for i in range(self.COMMIT_SIZE // self.PIN_CHUNK_SIZE):
+                if self._curr_size >= self._final_size:
+                    break
+                self._pin_memory_chunk(self._curr_size, self.PIN_CHUNK_SIZE)
+                self._curr_size += self.PIN_CHUNK_SIZE
+
+            expand_size = self._curr_size - last_commit_size
+            self._commit_expansion(expand_size)
```
DongDongJu, line 232: what happened addressmanager call the allocation in same address during this call happened without GIL?
IMO, we need to make AddressManager state.

DongDongJu (approved):
wow. Thanks for the this contribution.
It will be helpful for other memory type's backend.

From my understanding is that allocate a big buffer once, expose allocate/free via TensorMemoryAllocator, and grow the heap via AddressManager.sbrk.

I left few comments but its edge cases and my personal concern.
generally LGTM.

PR #2412, opened 2026-01-14
[Fix] Handle shape/dtype length mismatch with vLLM nightly builds (Fix zip error)

Young-TW:
Re-submitting because my previous PR was accidentally wiped out by my auto-sync bot before merge.

This PR fixes a crash occurring in `lmcache/v1/protocol.py` when using LMCache with **vLLM nightly builds** (e.g., `20260112`).

It adds a fallback mechanism to auto-fill self.dtypes when it is shorter than self.shapes (common in vLLM nightly builds). It retains strict=True in zip() to ensure that the padding logic functions correctly and that data integrity is maintained.

**Changes:**

1. Added a fallback mechanism in `_prepare_params` to auto-fill `self.dtypes` if its length is shorter than `self.shapes`. It uses the first available dtype (or defaults to `torch.bfloat16`) to pad the list.
2. Removed `strict=True` from the `zip` iteration to make the protocol more robust against such length mismatches in the future.

* **Hardware**: AMD MI300X *8 (ROCm environment).
* **Docker Image**: `rocm/vllm-dev:nightly_main_20260112` (This crash is reproducible on this specific nightly build).
* **LMCache Version**: Latest source.

* **Traceback**:
```text
File "/app/LMCache_Src/LMCache/lmcache/v1/protocol.py", line 108, in _prepare_params
  for shape, dtype in zip(self.shapes, self.dtypes, strict=True):
ValueError: zip() argument 2 is shorter than argument 1
```

* **Context**: The issue stems from a mismatch in the `shapes` and `dtypes` lists returned by the vLLM adapter in this nightly build. The patch assumes that if `dtypes` are missing but `shapes` are present, the dtype is uniform across layers (standard for models like Llama 3.1).

* **Logic**: The patch assumes that if `dtypes` are missing but `shapes` are present, the dtype is uniform across layers (standard for models like Llama 3.1).

**If applicable**:

* [ ] this PR contains user facing changes - docs added
* [ ] this PR contains unit tests


Young-TW:
Apologies for the notification noise. My automated bot accidentally closed/overwrote the previous PR (#2408). 😔

I have reopened the request here with the same changes.

@maobaolong  Could you please take a look when you have a moment? Thanks!

maobaolong:
@chunxiaozheng Would you like to take a look at this?

chunxiaozheng:
@Young-TW Thanks for your fix, some quality check failed.

lmcache/v1/protocol.py
```diff
@@ -104,8 +104,24 @@ class RemoteMetadata:
     fmt: MemoryFormat
 
     def _prepare_params(self):
+        # vLLM might return fewer dtypes than shapes (or empty dtypes).
+        # We auto-fill the missing dtypes with the first available one or a default.
+        final_dtypes = self.dtypes
+        if len(self.dtypes) < len(self.shapes):
```
chunxiaozheng, line 110: @Young-TW I have a question, if your vllm version with https://github.com/vllm-project/vllm/pull/31397, the `shapes` and `dtypes` are extract from `kv_caches` from vllm,  if without, the length of `shapes` and `dtypes` should be 1. why is it not equal here?

Young-TW: Hi @chunxiaozheng. thanks for the review!

I encountered this issue while testing with the latest vllm nightly builds (specifically rocm/vllm-dev:nightly_main_20260120).

In my observation, when running with these versions, self.shapes correctly reflects multiple layer groups (e.g., len(shapes) > 1), likely due to the changes in vLLM's V1 engine support. However, self.dtypes currently only contains a single element (e.g., len(dtypes) == 1).

This creates a mismatch (e.g., shapes=4, dtypes=1) that triggers the strict=True error in zip. My fix assumes that if a single dtype is provided, it should apply to all corresponding shape groups, so I pad the list to ensure compatibility without removing the strict check.

chunxiaozheng: @Young-TW `self.shapes` and `self.dtypes` are get from `metadata`, the code is as follows:
```
    def get_dtypes(self) -> list[torch.dtype]:
        if self.kv_layer_groups_manager.kv_layer_groups:
            return [
                group.dtype for group in self.kv_layer_groups_manager.kv_layer_groups
            ]
        return [self.kv_dtype]

    def get_shapes(self, num_tokens: Optional[int] = None) -> list[torch.Size]:
        """Get the shapes of the KV cache in LMCache"""
        if num_tokens is None:
            num_tokens = self.chunk_size
        if self.kv_layer_groups_manager.kv_layer_groups:
            shapes = []
            kv_size = 1 if self.use_mla else 2
            for group in self.kv_layer_groups_manager.kv_layer_groups:
                shapes.append(
                    torch.Size(
                        [
                            kv_size,
                            group.num_layers,
                            num_tokens,
                            group.hidden_dim_size,
                        ]
                    )
                )
            return shapes
        else:
            return [
                torch.Size(
                    [
                        self.kv_shape[1],
                        self.kv_shape[0],
                        num_tokens,
                        self.kv_shape[3] * self.kv_shape[4],
                    ]
                )
            ]
```
if the length of `self.shapes` and `self.dtypes` is not match, i suspect has bug here? I think we should find the reason that result in the mismatch, rather than fill the `dtypes`.

Young-TW: You are right that PR #31397 is merged. I double-checked my environment, and I am running a [vLLM image](https://hub.docker.com/layers/rocm/vllm-dev/nightly_main_20260120/images/sha256-6c90a3aaaf1dd878424ab72a731026de235bed2947776a7cb401f7744fc97685) from 3 days ago (which definitely includes that PR).

This implies a deeper root cause: Even with the PR present, self.kv_layer_groups_manager.kv_layer_groups is evaluating to Empty/False at the time LMCache initializes, yet the underlying engine is effectively producing multiple tensor chunks at runtime.

This suggests a race condition or initialization order issue: The LMCacheConnector might be initializing before the vLLM ModelRunner has fully populated the layer grouping information.

So, we have a "Split-Brain" state:

    Control Flow (Init): Reads empty groups -> Returns 1 dtype.

    Data Flow (Runtime): Engine executes with split groups -> Yields N chunks.

Young-TW:
I'm closing this PR for now. If the issue persists in the future or if the maintainers decide this fix is necessary, please let me know, and I'll be happy to reopen it or discuss further.

PR #2413, opened 2026-01-15
[Core] Add new memcpy c-extensions to support LazyMemoryAllocator

ApostaC:
**What this PR does / why we need it**:

It's part of #2379 

Need to rebase after #2407 is merged

The new `LazyMemoryAllocator` will register the pin memory segment by segment. However, when the memory copy happens across the segment boundary, it will cause a CUDA error of "invalid argument".

This PR introduces a new wrapper around the CUDA memory copy to make sure that the copies never cross the boundary.

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #2414, opened 2026-01-15
[Feat] Introduce progressive pinned memory allocators to reduce startup memory allocation time

AgainstEntropy:

This PR introduces `ProgressivePinnedMemoryAllocator`and `ProgressivePinnedPagedMemoryAllocator`, which is similar to original `lazy_memory_allocator.py`, but:
- Reserve a single contiguous host memory to avoid fragmentation
- Register only a ratio of memory on startup to reduce initialization time
- Expand asynchronously once usage crosses a threshold

Test cases are also added

**If applicable**:


YaoJiayi (commented):
Thanks @AgainstEntropy ! I actually think this is complementary to https://github.com/LMCache/LMCache/pull/2413 . 

The goal here (for this PR) is to unify the code path between the normal memory allocator and lazy memory allocator.

I think we can (1) allocate a non-pinned cpu buffer at the start (like you did), and (2) then gradually put the pinned memory pages/chunks into the existing memory allocator. I feel the code could be much simpler in this way:)

lmcache/v1/memory_management.py
```diff
@@ -1977,6 +1978,511 @@ def __str__(self):
...
+
+        with self._lock:
+            self._register_more_locked(0, initial_size)
+            self._on_init(initial_size)
+
+    def _register_more_locked(self, start: int, size: int) -> None:
+        """Register a subrange [base+start, base+start+size). Caller holds lock."""
+        if size <= 0:
+            return
+        lmc_ops.cuda_host_register(self._base_ptr + start, size, 0)
```
YaoJiayi, line 2086: I feel we can release the GIL for this function?

PR #2415, opened 2026-01-15
[Fix]: remove duplicated ref count decrement calls of Mooncake Connector

duhaode520:

Both `InstrumentedRemoteConnector.batched_put()`  and `MooncakestoreConnector._batched_put_zero_copy()`/`_batched_put_with_metadata()` call `ref_count_down()` in their `finally` blocks, resulting in double free.

Other connectors like `RedisConnector`do NOT call `ref_count_down()` in their `batched_put()` methods because `InstrumentedRemoteConnector` is designed to handle this uniformly.

**Fix**:

Remove the redundant `ref_count_down()` calls in `mooncakestore_connector.py`:
- `_batched_put_zero_copy()` method (line 536-538)
- `_batched_put_with_metadata()` method (line 548-549)

This aligns with the behavior of other connectors and the design intent of `InstrumentedRemoteConnector`.

- This issue only occurs when using mooncakestore backend; `local_cpu` backend is not affected since it doesn't go through `RemoteBackend` and `InstrumentedRemoteConnector`.
- The fix follows the existing pattern used by `RedisConnector` .

### DIFF
--- a/lmcache/v1/storage_backend/connector/mooncakestore_connector.py
+++ b/lmcache/v1/storage_backend/connector/mooncakestore_connector.py
@@ -533,9 +533,6 @@ class MooncakestoreConnector(RemoteConnector):
             logger.warning(
                 "Timeout during batch_put_from; some decoders may redo prefill."
             )
-        finally:
-            for obj in memory_objs:
-                obj.ref_count_down()
 
     async def _batched_put_with_metadata(
         self,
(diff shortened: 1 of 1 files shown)


duhaode520:
@maobaolong @chunxiaozheng ,could you help review this?

chunxiaozheng:
@duhaode520 Good catch! Could you add your `DCO` information?

duhaode520:
> @duhaode520 Good catch! Could you add your `DCO` information?

Sign-off information is added.

chunxiaozheng (approved):
lgtm

maobaolong (approved):
Good catch!

maobaolong:
BTW, we need a tool to verify the ref-count correct for all the backends.

PR #2399, opened 2026-01-13
[Refactor]: Move northbound source of truth to LMCacheEngineMetadata

sammshen:
Step 1 of https://github.com/LMCache/LMCache/issues/2384

Follow-up small action items (open to discussion): 
[ ] Move the LMCacheEngineMetadata to `v1/config.py` from `config.py`
[ ] Rename the LMCacheEngineConfig and LMCacheEngineMetadata to LMCacheConfig and LMCacheMetadata

# Testing

## SGLang Integration Testing (from https://docs.lmcache.ai/getting_started/quickstart.html):

@DongDongJu I cannot get SGLang to store and load KVCache (even without this PR). Were you able to test it recently?
```bash
cat > lmc_config.yaml <<'EOF'
chunk_size: 8  # demo only; use 256 for production
local_cpu: true
use_layerwise: true
max_local_cpu_size: 10  # GB
EOF

export LMCACHE_CONFIG_FILE=$PWD/lmc_config.yaml

python -m sglang.launch_server \
  --model-path Qwen/Qwen3-14B \
  --host 0.0.0.0 \
  --port 30000 \
  --enable-lmcache \
  --disable-radix-cache

curl http://localhost:30000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen/Qwen3-14B",
    "messages": [{"role": "user", "content": "Qwen3 is the latest generation of large language models in Qwen series, offering a comprehensive suite of dense and mixture-of-experts"}],
    "max_tokens": 100,
    "temperature": 0.7
  }'
```

## Standalone Mode Testing:
```bash
python -m lmcache.v1.standalone
```

PR #2444, opened 2026-01-19
[bugfix] fix LMCacheManager get kv_caches error

chunxiaozheng:
The `lmcache_adapter` is instance of `LMCacheManager`, and the `/cache/kvcache/check` in `InternalApiServer` will cause some errors in following position: 
Clipboard_Screenshot_1768822317

### DIFF
--- a/lmcache/v1/manager.py
+++ b/lmcache/v1/manager.py
@@ -644,6 +644,12 @@ class LMCacheManager:
         """Get the role of this manager (scheduler or worker)."""
         return self._role
 
+    @property
+    def kv_caches(self) -> dict[str, torch.Tensor]:
+        if self._connector is not None and hasattr(self._connector, "kv_caches"):
+            return self._connector.kv_caches
+        return {}
+
     def is_healthy(self) -> bool:
         """
         Check if the LMCacheManager is healthy.


maobaolong (approved):
LGTM

lmcache/v1/manager.py
```diff
@@ -644,6 +644,12 @@ def role(self) -> str:
         """Get the role of this manager (scheduler or worker)."""
         return self._role
 
+    @property
+    def kv_caches(self) -> dict[str, torch.Tensor]:
+        if self._connector is not None and hasattr(self._connector, "kv_caches"):
+            return self._connector.kv_caches
+        return {}
```
ApostaC, line 651: Actually, just realized that: will this create problems when doing integration with other serving engines like SGLang?

Here we assume the manager holds a reference to the connector, which is "vllm-specific". What's a good way to keep `LMCacheManager` serving engine agnostic?
 
cc @maobaolong

ApostaC (commented):
left 1 quick question

lmcache/v1/manager.py
sammshen: For this property specifically, I think we should hide/expose the connector inside of the LMCache Engine

I am trying to work on the serving engine agnostic version of LMCacheManager (rewriting this PR currently: https://github.com/LMCache/LMCache/pull/2399) and I realize that so much of the LMCacheManager is very tightly coupled with vllm or at least feels like it belongs inside of the adapter (especially related to roles). 

Some examples include: 
- any branching logic with "role" and the lookup client/server components
- various components like the internal api server and runtime plugin launcher require various elements from vllm config to create
- the adapter extracts these components from the lmcache manager anyways

My current thoughts are that: 
- lookup client, lookup server, cache engine are core services. furthermore lookup client/server are only used by vllm and not sglang right now.

Proposal: 
- revert LMCacheManager. Have every adapter manage the services that they need. If we need, we can bundle the auxiliary components into `LMCacheServiceManager`
- move toward making the LMCacheEngine serving engine agnostic (through LMCacheMetadata as source of truth) and also make the LMCacheEngineBuilder the main component responsible for branching logic (e.g. the gpu connector)

cc @chunxiaozheng @maobaolong

sammshen: one simple way to resolve this is to literally have LMCacheManager be thin wrapper around the components that are manually decided whether or not to set to True or False

maobaolong: > Actually, just realized that: will this create problems when doing integration with other serving engines like SGLang?
>
> Here we assume the manager holds a reference to the connector, which is "vllm-specific". What's a good way to keep LMCacheManager serving engine agnostic?
>

How about do an abstraction for the LMCacheConnector? Something like this?

```mermaid
classDiagram
    class AbstractLMCacheConnector {
        <<abstract>>
        +get_kv_caches() List
        +get_inference_info() Dict
    }
    
    class VLLMLMCacheConnector {
        +get_kv_caches() List
        +get_inference_info() Dict
    }
    
    class SGLangLMCacheConnector {
        +get_kv_caches() List
        +get_inference_info() Dict
    }
    
    class TrtLLMLMCacheConnector {
        +get_kv_caches() List
        +get_inference_info() Dict
    }
    
    AbstractLMCacheConnector <|-- VLLMLMCacheConnector
    AbstractLMCacheConnector <|-- SGLangLMCacheConnector
    AbstractLMCacheConnector <|-- TrtLLMLMCacheConnector
```

maobaolong: @sammshen 
> we can bundle the auxiliary components into LMCacheServiceManager
This is a good idea. 

And I got what you said, i'm happy if you could do a second refactor on this part, as the pervious refactor didn't care about SGLang and other inference engines.

chunxiaozheng:
@ApostaC @sammshen @maobaolong Thanks for your review! I'm not quite sure about the previous refactor, but now if this PR is missing, it will directly report an error in `/cache/kvcache/check` of `InternalApiServer`. Could we merge this small bug fix first and refactor it together with the subsequent PR?

sammshen (approved):
LGTM, but `_connector` has been renamed `northbound` in https://github.com/LMCache/LMCache/pull/2457

ideally first merge the refactor (easier to rebase this branch) but if it is urgent, we can also merge this first

chunxiaozheng:
> LGTM, but `_connector` has been renamed `northbound` in #2457
> 
> ideally first merge the refactor (easier to rebase this branch) but if it is urgent, we can also merge this first

@sammshen Thanks! I want to merge this PR and create a new package. Could you please rebase it? Thanks very much!

PR #2447, opened 2026-01-19
[Bugfix] Fix vLLM connector init signature ordering

### LINKED ISSUE #2446: [Bug] LMCache vLLM dynamic connector init signature mismatch causes role=None crash
**Label**

Label
bug, LMCache Onboarding

Describe the bug
LMCache’s dynamic vLLM connector constructor signature/order is incompatible with vLLM’s factory call. vLLM 0.13.x instantiates connectors as [`KVConnectorBase(config, role, kv_cache_config)`](https://docs.vllm.ai/en/latest/api/vllm/distributed/kv_transfer/kv_connector/factory/#vllm.distributed.kv_transfer.kv_connector.factory.KVConnectorFactory), but `LMCacheConnectorV1Dynamic` previously accepted `(vllm_config, kv_cache_config, role)`. This causes role to be `None` and crashes with `AttributeError: 'NoneType' object has no attribute 'name'` during worker initialization.

To Reproduce

1. Install LMCache from source and vLLM 0.13.x.
2. Launch vLLM with LMCache dynamic connector:

```bash
vllm serve /path/to/model \
--kv-transfer-config '{"kv_connector":"LMCacheConnectorV1Dynamic","kv_role":"kv_both","kv_connector_module_path":"lmcache.integration.vllm.lmcache_connector_v1"}'
```

3. Observe worker initialization error in vLLM logs.

Expected behavior
vLLM should initialize the LMCache connector without crashing, and workers should start normally.

Screenshots
N/A

Desktop:

- OS: Linux
- Browser: N/A
- Version: N/A

Additional context
Error snippet:

```bash
AttributeError: 'NoneType' object has no attribute 'name'
  File "lmcache/integration/vllm/vllm_v1_adapter.py", line 459, in __init__
    role=role.name.lower()
```

Fix: update `LMCacheConnectorV1Dynamic.__init__` signature to `(vllm_config, role, kv_cache_config=None)` and pass through to`KVConnectorBase_V1`.

tunglinwood:
Fix #2446

- Align `LMCacheConnectorV1Dynamic` constructor signature with vLLM's factory call order `(config, role, kv_cache_config)` to prevent `role` becoming `None` and crashing worker init.
- Pass `kv_cache_config` through to `KVConnectorBase_V1` for compatibility with new vLLM API.
- Note: `LMCacheConnectorV1Impl` does not accept `kv_cache_config`, so no changes are needed there.

- Single-file change in `lmcache/integration/vllm/lmcache_connector_v1.py`.
- Testing: `ruff check lmcache/integration/vllm/lmcache_connector_v1.py`, `pre-commit run --files lmcache/integration/vllm/lmcache_connector_v1.py`.

**If applicable**:


ApostaC (approved):
LGTM.
@maobaolong PTAL because I suppose you guys are using this?

tunglinwood:
@maobaolong PTAL. Thank you :)

maobaolong:
@tunglinwood This lgtm overall, the only concern is about the compatibility with old
 vllm

lmcache/integration/vllm/lmcache_connector_v1.py
```diff
@@ -28,8 +28,41 @@
...
+                sorted(kwargs.keys()),
+            )
+        if not isinstance(role, KVConnectorRole) and isinstance(
+            kv_cache_config, KVConnectorRole
+        ):
+            logger.warning(
+                "Detected swapped role/kv_cache_config in connector init; "
+                "auto-correcting."
+            )
+            role, kv_cache_config = kv_cache_config, role
```
ApostaC: Why do we have this? IIUC, this would not happen, right?

lmcache/integration/vllm/lmcache_connector_v1.py
```diff
@@ -28,8 +28,41 @@
 
 
 class LMCacheConnectorV1Dynamic(KVConnectorBase_V1):
-    def __init__(self, vllm_config: "VllmConfig", role: KVConnectorRole):
-        super().__init__(vllm_config=vllm_config, role=role)
+    def __init__(
+        self,
+        vllm_config: "VllmConfig",
+        role: Optional[KVConnectorRole] = None,
```
ApostaC: If I remember correctly, "role" argument is always there in all the versions. Don't need to make it optional. (I could be wrong, just let me know)

ApostaC (changes requested):
Thanks for the update. I think the new version is a bit over-complicated.

Previously I was thinking about having something like
```python3
    def __init__(
        self,
        vllm_config: "VllmConfig",
        role: KVConnectorRole,
        kv_cache_config: Optional[Any] = None,
    ):
        if kv_cache_config is not None:
            super().__init__(vllm_config=vllm_config, role=role, kv_cache_config=kv_cache_config)
        else:
            super().__init__(vllm_config=vllm_config, role=role)
```

Will this work for old versions as well? cc @maobaolong

maobaolong:
@ApostaC i think this can work from Python Interpretor vision

tunglinwood:
@maobaolong @ApostaC Updated PTAL.

ApostaC (approved):
lgtm!

maobaolong (approved):
LGTM

PR #2457, opened 2026-01-20
[Refactor]: Serving Engine Agnostic Manager

sammshen:
To Merge this PR, we need:
- [ ] CI to pass (helps verify vLLM integration under most cases) 
- [ ] manually show that SGLang quick start still works
- [ ] manually show that the standalone still works

## Manager Refactor PR: Serving-Engine-Agnostic Architecture

**Goal:** Decouple LMCacheManager from vLLM-specific logic to support multiple serving engines (vLLM, SGLang, standalone).

**Target:** LMCacheManager is now completely serving-engine-agnostic. All engine-specific logic moved to adapter-level metadata builders.

### Summary of Changes:
- Created adapter-level metadata builders (vLLM, SGLang, standalone)
- Moved GPU connector construction from manager → LMCacheEngineBuilder
- Simplified LMCacheEngineBuilder.get_or_create() to 3 parameters
- LMCacheManager now uses metadata "needs" methods instead of role-based branching
- Shrunk manager.py significantly and no vllm dependencies
- metadata refinement: Deprecated `fmt` field (use `use_case`) and `num_ranks` field (use `world_size`)

---

Details:
1. Metadata Responsibility Increase
   - Added property wrappers in LMCacheEngineMetadata for all vLLM-specific accesses:
     * `is_cuda_alike` - Platform detection (CUDA, ROCm)
     * `is_xpu` - Intel XPU detection
     * `compute_device` - Dynamically computes device based on use_case
     * `engine_name` - Returns appropriate ENGINE_NAME constant
     * `broadcast_fn` - Returns tensor broadcast function
     * `broadcast_object_fn` - Returns object broadcast function
     * `tensor_model_parallel_rank` - TP rank for offload server
     * `tensor_parallel_size` - TP size for plugin launcher
     * `is_dp_rank0` - Data parallel rank check
   - Added "needs" methods for component toggles:
     * `needs_cache_engine(config)` - With bypass lookup logic
     * `needs_lookup_client` - Scheduler only (vLLM)
     * `needs_lookup_server` - Worker only (vLLM)
     * `needs_offload_server` - Worker only (vLLM)
     * `needs_api_server` - DP rank 0 or standalone
     * `needs_runtime_plugin_launcher` - DP rank 0 (vLLM)
     * `needs_health_monitor` - vLLM only
   - Deprecated fields converted to properties:
     * `num_ranks` → property that computes TP*PP for vLLM, world_size for others
     * `fmt` → property that returns `use_case`

2. Adapter-Level Metadata Builders
   - Created `lmcache/integration/vllm/metadata.py`:
     * `VLLMMetadataBuilder.build()` - Extracts metadata from VllmConfig
     * Includes draft layer calculation (EAGLE, DeepSeek MTP)
     * Includes device detection (CUDA, XPU)
     * Includes MLA validation
   - Created `lmcache/integration/sglang/metadata.py`:
     * `SGLangMetadataBuilder.build()` - Extracts metadata from SGLang ModelConfig
     * Sets device based on local_rank
   - Created `lmcache/integration/standalone/metadata.py`:
     * `StandaloneMetadataBuilder.build()` - Creates metadata for standalone mode
     * Supports layer groups configuration
   - Moved standalone to `lmcache/integration/standalone/` (parallel to vllm/sglang)
   - Created ENGINE_NAME constants:
     * `lmcache/integration/vllm/utils.py` - "vllm-instance"
     * `lmcache/integration/sglang/utils.py` - "sglang-instance"  
     * `lmcache/integration/standalone/utils.py` - "standalone-instance"

3. LMCacheEngine and LMCacheEngineBuilder Refactoring
   - Moved GPU connector creation to `LMCacheEngineBuilder`:
     * `_Create_gpu_connector()` - Dispatcher based on use_case
     * `_create_gpu_connector_for_vllm()` - vLLM connector logic (from manager.py)
     * `_create_gpu_connector_for_sglang()` - SGLang connector logic (from adapter)
     * `_create_gpu_connector_for_standalone()` - Always uses MockGPUConnector
   - Moved broadcast function creation to metadata properties
   - Simplified `get_or_create()` signature to only require:
     * `instance_id` (from metadata.engine_name)
     * `config`
     * `metadata`
   - Builder now auto-creates GPU connector and broadcast functions internally
   - Removed 150+ lines of duplication from manager.py and adapters

4. Changed LMCacheManager Semantics
   - Removed role-based component initialization (_init_scheduler_components, _init_worker_components)
   - New unified initialization using metadata "needs" methods:
     * Components only created if metadata.needs_XXX returns True
     * No more hardcoded scheduler/worker branching in manager
   - Changed parameter from `connector` to `northbound` (clearer naming)
   - Removed `vllm_config` and `role` parameters from __init__
   - New signature: `LMCacheManager(config, metadata, northbound)`
   - Removed direct vLLM dependencies from manager.py
   - Manager is now a "loose bundler" - metadata controls what gets created

5. Additional Improvements
   - Standalone mode always uses MockGPUConnector (no real GPU interaction)
   - Simplified StandaloneLMCacheManager to be a thin wrapper
   - Fixed bypass lookup documentation (queries remote DB for key existence)
   - All three adapters now follow consistent pattern:
     1. Build metadata (adapter-specific)
     2. Create manager (generic)
     3. Manager creates components based on metadata


tests/v1/test_cache_engine.py
```diff
@@ -895,11 +896,12 @@ def test_paged_prefetch_retrieve(
     )
 
     async_lookup_server = DummyLMCacheAsyncLookupServer()
+    memory_format = MemoryFormat.KV_2LTD
     engine = autorelease_v1(
         LMCacheEngineBuilder.get_or_create(
             "test",
             cfg,
-            dumb_metadata(fmt, kv_shape),
+            dumb_metadata(memory_format, kv_shape),
```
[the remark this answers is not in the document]
sammshen: fixed

tests/v1/test_cache_engine.py
```diff
@@ -1279,11 +1281,12 @@ def generate_random_slot_mapping(num_blocks, block_size, num_tokens, device):
             extra_config={"force_store_wait": True},
         )
 
+        memory_format = MemoryFormat.KV_2LTD
         engine = autorelease_v1(
             LMCacheEngineBuilder.get_or_create(
                 "test",
                 cfg,
-                dumb_metadata(fmt, kv_shape),
+                dumb_metadata(memory_format, kv_shape),
```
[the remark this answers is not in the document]
sammshen: fixed

maobaolong:
@gemini-code-assist  Would you like to draw a diagram for us to clarify the relationship among these PR motioned components between inference side and lmcache side? Notes: Use mermaid.

maobaolong:
> needs_api_server - DP rank 0 or standalone

This is a trick, scheduler can be treat as DP rank 0 also actually.

lmcache/config.py
```diff
@@ -100,6 +119,428 @@ def get_num_groups(self) -> int:
...
+            return (
+                self.serving_engine_config.parallel_config.data_parallel_rank_local == 0
+            )
+        else:
+            # TODO: add the way to query data parallel rank for other use cases
+            logger.warning(
+                "data parallel rank checking is not supported "
+                f"for use case {self.use_case} yet"
+            )
+            return False
```
ApostaC, line 212: Probably let's move it down so that it stays together with the `is_xxx` functions?

lmcache/config.py
```diff
@@ -100,6 +119,428 @@ def get_num_groups(self) -> int:
...
+                logger.warning("vllm not available, XPU check unavailable")
+                return False
+        elif self.use_case == "sglang":
+            # SGLang doesn't support XPU currently
+            return False
+        else:
+            return False
+
+    @property
+    def compute_device(self) -> Any:  # torch.device
```
ApostaC, line 290: Should the return value be `torch.device`?

lmcache/config.py
```diff
@@ -100,6 +119,428 @@ def get_num_groups(self) -> int:
...
+                    return mock_up_broadcast_object_fn
+        else:
+            # Default no-op for other use cases
+            # First Party
+            from lmcache.utils import mock_up_broadcast_object_fn
+
+            return mock_up_broadcast_object_fn
+
+    # -- Serving engine specific property accessors --
+    def save_only_first_rank(self, lmcache_config) -> bool:
```
ApostaC, line 435: Why do we need to have this in `LMCacheEngineMetadata`? It should be from `LMCacheEngineConfig` IIUC

lmcache/config.py
```diff
@@ -100,6 +119,428 @@ def get_num_groups(self) -> int:
...
+                    )
+        else:
+            logger.warning(
+                "default None for kv connector extra config "
+                f"for use case {self.use_case}"
+            )
+        return kv_connector_extra_config
+
+    @property
+    def _calculate_draft_layers(self) -> int:
```
ApostaC, line 505: If it's a property that needs to be used by other functions, let's remove the preceding `_` from the function name

lmcache/config.py
ApostaC, line 1: I like the standard property interface defined in this file.
However, I personally don't want to do `import vllm` in `config.py`. 

IMO, we should make this a "standard interface that needs to be provided by the serving engines". And the implementation should be on the serving engine side (i.e., in vLLM/SGL repo) and be secured by their CI/CD.

LMCacheEngineMetadata should take that from the serving engine during initialization, checking whether it provides enough interfaces/properties, and export them to other LMCache modules.

ApostaC: It's more like

On LMCache side:
```python3
class LMCacheEngineMetadata:
    # other codes ...
    """ serving engine integration module """
    engine_adapter: Any

    def __post_init__(self):
        expected_fields = [...]
        expected_functions = [...]
        for field in expected_fields:
            assert ....
        for function in expected_functions:
            assert ....
        
        # Add the fields and functions into self
```

On serving engine side (e.g., vLLM)
```python3
class VLLMIntegrationAdapter:
    @property
    def need_xxx():
        # some implementations...

### Other codes

# When creating LMCacheEngineMetadata
    vllm_integration_adapter = VLLMIntegrationAdapter(...)
    metadata = LMCacheEngineMetadata(..., vllm_integration_adapter)
    # other intializations
```

ApostaC (commented):
Thanks for having this PR! I like the idea of the refactoring.

However, can we split the PR into multiple pieces to make reviewing easier?
1. Change the "fmt" to "use_case"
2. Define the standard interface in LMCacheEngineMetadata, and let other adapters to use those interfaces for initializations
3. Move the gpu_connector initialization to CacheEngineBuilder
4. Move the standalone module to `integration` folder

Let me know if the above split does not work. Maybe there are some extra coupling/dependency I'm not aware of

lmcache/config.py
ApostaC: or have a util function to take in both LMCacheEngineMetadata and LMCacheEngineConfig.
It's a bit weird to make it belong to LMCacheEngineMetadata

sammshen:
thanks for the feedback! will break this PR into 4 smaller PRs

maobaolong:
Discussed with @sammshen @chunxiaozheng in a short meeting.

We discussed on component ownership and placement logic and how to refactor the LMCache again.
1. Treat Lookup Client/Server as a general capability, although it belongs to vLLM for now. We can create and hold `Lookup Client/Server` in LMCacheManager, but decided whether to create them in inference side.
2. Samuel has been introduced some better conditional components methods, it is a good idea, and we can also use `Polymorphism and inheritance` to implement the need_xxx methods in each individual inference side.
3. Move Standalone to integration folder.
4. We discussed the component ownership by a diagram 
https://app.diagrams.net/#Hmaobaolong%2Fdraw%2Fmaster%2Flmcache.drawio#%7B%22pageId%22%3A%22eNNvAO4dPXWR6fiodOs1%22%7D

Clipboard_Screenshot_1768981894

Issue #2460, opened 2026-01-20
RFC: Recipe examples and best practices (vLLM/sglang, backends)


tunglinwood:
**Label**
LMCache Onboarding, RFC, Doc, Recipes

**Summary**
RFC: request committer feedback on a recipe template and propose several options for future recipe examples (vLLM/sglang, multiple backends), aligned with the 2026 Q1 roadmap.

**Details**
The goal is to build a clear, repeatable recipe library that looks like: "Do X and Y, then get Z" with measurable outcomes (TTFT, hit rate, throughput). I am requesting comments from committers/maintainers on scope, structure, and prioritization.

**RFC options (please comment on preference or suggest alternatives)**
Option A: "Core recipes" (MVP)
- vLLM + LMCache: CPU hot cache
- vLLM + LMCache: CPU + disk tier
- sglang + LMCache: CPU hot cache
- Each recipe includes validation + a small benchmark

Option B: "Backend matrix" focus
- Same frontend (vLLM/sglang), cover backends: CPU hot cache, local disk, Redis, Mooncake, GDS
- Emphasize configuration deltas and tradeoffs

Option C: "Performance playbook" focus
- A smaller set of recipes but deep benchmarks
- Show how to tune chunk size, layerwise, save_unfull_chunk, and measure hit rates

Option D: "Use-case guided"
- Multi-turn chat
- RAG with repeated context
- Long-context summary
- Each maps to a known LMCache KV reuse pattern

**Steps / Reproduction (if applicable)**
N/A

**Expected Outcome / Goal**
- Agreement on the recipe template and review criteria
- Guidance on which RFC option(s) to execute first
- Direction on how this fits the 2026 Q1 roadmap

**Actual Outcome (if applicable)**
N/A

**Additional Context**
Related roadmap issue: https://github.com/LMCache/LMCache/issues/2350
A draft recipe PR will follow, targeting `LMCache/LMCache:dev`, with:
- `recipes/dense_instruct_cpu_hot_cache.md`
- `recipes/dense_instruct_cpu_hot_cache.yaml`

PR #2461, opened 2026-01-20
[Doc] add dense instruct CPU hot cache recipe

tunglinwood:
## Summary
- add dense instruct CPU hot cache recipe (markdown + YAML config)

## Context
- RFC issue: https://github.com/LMCache/LMCache/issues/2460

## Testing
- Not run (docs/config only)


recipes/dense_instruct_cpu_hot_cache.md
```diff
@@ -0,0 +1,330 @@
...
+## 3. LMCache Configuration
+
+Create `recipes/dense_instruct_cpu_hot_cache.yaml`:
+
+```yaml
+chunk_size: 256           # Default chunk size
+local_cpu: true
+max_local_cpu_size: 8     # GB of pinned CPU memory
+use_layerwise: true       # Overlap KV load with forward pass
+```
```
ApostaC, line 66: What's the GPU we are using here? I'm asking this because `max_local_cpu_size: 8` seems to be very small. In practice, we need to make sure that  LMCache's CPU offload buffer is larger than vLLM's GPU KV cache buffer in order to be effective.

ApostaC: I saw you are using `--no-enable-prefix-caching` below. But let's try not to give people a impression that "a small local cpu size can lead to good performance improvements".

Therefore, how about let's make it clear with something like "in practice, we want to make sure that this number is 1.5x bigger than the GPU memory size"

tunglinwood: > What's the GPU we are using here? I'm asking this because `max_local_cpu_size: 8` seems to be very small. In practice, we need to make sure that LMCache's CPU offload buffer is larger than vLLM's GPU KV cache buffer in order to be effective.

I am on H20\*8 and L20\*8

> I saw you are using `--no-enable-prefix-caching` below. But let's try not to give people a impression that "a small local cpu size can lead to good performance improvements".
> 
> Therefore, how about let's make it clear with something like "in practice, we want to make sure that this number is 1.5x bigger than the GPU memory size"

Agreed. I will put that into description. Is x1.5 a ballpark figure? Or do we have supporting reference?

ApostaC: 1.5x is a ballpark.

ApostaC:
image

Here is a figure to illustrate the configuration recommendations (i.e., LMCache's CPU buffer should be larger than vllm's GPU buffer). Might be useful and feel free to put that into the doc if you want

tunglinwood:
@ApostaC This illustration looks really great!!!! If I were beginner, I would definitely like to see more graphic information in the recipes. So yes, it is really useful!!! Please provide me more so I can put them into the recipe. And I think I will come up more recipes this week and coming weeks until the CNY. Let's see how it goes and we can add them on the fly.

tunglinwood:
Hi @ApostaC,

Verification is done. Please review the first recipe.

PR #2463, opened 2026-01-20
[Core] enable lazy memory allocator by default in multi-process mode

ApostaC:

Part of #2379 
This PR enables lazy memory allocator by default in multi-process mode, and also introduces a new flag to use the "non-lazy" allocator.

**This PR will be directly tested by multi-process mode testing**


sammshen (approved):
LGTM!

KuntaiDu (approved):
LGTM

PR #2466, opened 2026-01-21
Update metrics system

yoo-kumaneko:
1. Add granular profiling on the remote backend and remote connectors.
2. Fix a bug that can make the retrieval success rate > 100%.
3. Improve the robustness of the observability module.

### DIFF
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -2,7 +2,7 @@
 # Standard
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, Union
 import os
 import threading
 import time
@@ -674,79 +674,69 @@ class LMCStatsMonitor:
             else sum_finished_lookup_hit / sum_finished_lookup_requested
         )
 
-        def filter_out_zeros(stats: List[float]):
+        def filter_out_zeros(stats: Iterable[float]) -> List[float]:
             return [x for x in stats if x != 0]
 
         time_to_retrieve = filter_out_zeros(
-            [stats.time_to_retrieve() for stats in self.retrieve_requests.values()]
+            stats.time_to_retrieve() for stats in self.retrieve_requests.values()
         )
 
         time_to_store = filter_out_zeros(
-            [stats.time_to_store() for stats in self.store_requests.values()]
+            stats.time_to_store() for stats in self.store_requests.values()
         )
 
         time_to_lookup = filter_out_zeros(
-            [stats.time_to_lookup() for stats in self.lookup_requests.values()]
+            stats.time_to_lookup() for stats in self.lookup_requests.values()
         )
 
         retrieve_speed = filter_out_zeros(
-            [stats.retrieve_speed() for stats in self.retrieve_requests.values()]
+            stats.retrieve_speed() for stats in self.retrieve_requests.values()
         )
 
         store_speed = filter_out_zeros(
-            [stats.store_speed() for stats in self.store_requests.values()]
+            stats.store_speed() for stats in self.store_requests.values()
         )
 
         # Granular profiling measurements
         retrieve_process_tokens_time = filter_out_zeros(
-            [stats.process_tokens_time for stats in self.retrieve_requests.values()]
+            stats.process_tokens_time for stats in self.retrieve_requests.values()
         )
         retrieve_broadcast_time = filter_out_zeros(
-            [stats.broadcast_time for stats in self.retrieve_requests.values()]
+            stats.broadcast_time for stats in self.retrieve_requests.values()
         )
         retrieve_to_gpu_time = filter_out_zeros(
-            [stats.to_gpu_time for stats in self.retrieve_requests.values()]
+            stats.to_gpu_time for stats in self.retrieve_requests.values()
         )
         remote_backend_batched_get_blocking_time = filter_out_zeros(
-            [
-                stats.detailed_metrics.get(
-                    "remote_backend_batched_get_blocking_time", 0.0
-                )
-                for stats in self.retrieve_requests.values()
-            ]
+            stats.detailed_metrics.get("remote_backend_batched_get_blocking_time", 0.0)
+            for stats in self.retrieve_requests.values()
         )
         instrumented_connector_batched_get_time = filter_out_zeros(
-            [
-                stats.detailed_metrics.get(
-                    "instrumented_connector_batched_get_time", 0.0
-                )
-                for stats in self.retrieve_requests.values()
-            ]
+            stats.detailed_metrics.get("instrumented_connector_batched_get_time", 0.0)
+            for stats in self.retrieve_requests.values()
         )
         store_process_tokens_time = filter_out_zeros(
-            [stats.process_tokens_time for stats in self.store_requests.values()]
+            stats.process_tokens_time for stats in self.store_requests.values()
         )
         store_from_gpu_time = filter_out_zeros(
-            [stats.from_gpu_time for stats in self.store_requests.values()]
+            stats.from_gpu_time for stats in self.store_requests.values()
         )
         store_put_time = filter_out_zeros(
-            [stats.put_time for stats in self.store_requests.values()]
+            stats.put_time for stats in self.store_requests.values()
         )
 
         p2p_time_to_transfer = filter_out_zeros(
-            [stats.time_to_transfer() for stats in self.p2p_requests.values()]
+            stats.time_to_transfer() for stats in self.p2p_requests.values()
         )
 
         p2p_transfer_speed = filter_out_zeros(
-            [stats.transfer_speed() for stats in self.p2p_requests.values()]
+            stats.transfer_speed() for stats in self.p2p_requests.values()
         )
 
         request_lookup_hit_rates = filter_out_zeros(
-            [
-                stats.hit_rate()
-                for stats in self.lookup_requests.values()
-                if stats.is_finished
-            ]
+            stats.hit_rate()
+            for stats in self.lookup_requests.values()
+            if stats.is_finished
         )
 
         request_lifespan = list(self.interval_request_cache_lifespan.values())
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -302,7 +302,7 @@ class LMCStatsMonitor:
 
         self._current_retrieve_stats: Optional[RetrieveRequestStats] = None
         self.retrieve_time_threshold: float = 1e9
-        self.retrieve_speed_threshold: float = -1.0
+        self.retrieve_token_speed_threshold: float = -1.0
         self.last_retrieve_warning_time: float = 0.0
 
     def set_current_retrieve_stats(self, stats: RetrieveRequestStats):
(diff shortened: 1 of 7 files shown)


lmcache/v1/cache_engine.py
```diff
@@ -388,102 +394,104 @@ def store(
...
-        t = time.perf_counter()
+        with store_stats.profile_from_gpu():
+            self.gpu_connector.batched_from_gpu(memory_objs, starts, ends, **kwargs)
 
-        transfer_spec = kwargs.get("transfer_spec", None)
-        # TODO: we implicitly rely on batched_put to call ref_count_down
-        # this management should be done in a cleaner way
-        self.storage_manager.batched_put(keys, memory_objs, transfer_spec=transfer_spec)
-        put_time += time.perf_counter() - t
+        with store_stats.profile_put():
```
maobaolong, line 480: if the put is asynchronous, this put will always fast but it is not what you need.

lmcache/observability.py
```diff
@@ -266,40 +371,85 @@ def on_retrieve_request(self, num_tokens: int) -> int:
...
+        self.interval_hit_tokens += num_retrieved_tokens
+        self.clear_current_retrieve_stats()
+
+        time_to_retrieve = retrieve_stats.time_to_retrieve()
+        retrieve_speed = retrieve_stats.retrieve_speed()
+        if (
+            time_to_retrieve > self.retrieve_time_threshold
+            or 0 < retrieve_speed < self.retrieve_speed_threshold
+        ):
+            logger.warning(
```
maobaolong: Can you also add a pair of metrics with counter type to record these two cases exceed the threshold.

lmcache/observability.py
```diff
@@ -266,40 +371,85 @@ def on_retrieve_request(self, num_tokens: int) -> int:
...
+                "Retrieve request %d surpassed threshold: "
+                "time_to_retrieve=%.4f s (threshold=%.4f s), "
+                "retrieve_speed=%.2f tokens/s (threshold=%.2f tokens/s). "
+                "Detailed metrics: "
+                "num_tokens=%d, local_hit_tokens=%d, remote_hit_tokens=%d, "
+                "process_tokens_time=%.5f s, "
+                "broadcast_time=%.5f s, "
+                "to_gpu_time=%.5f s, "
+                "remote_backend_batched_get_blocking_time=%.5f s, "
+                "instrumented_connector_batched_get_time=%.5f s",
```
maobaolong: It is great we have a log which can show detailed log for the worse cases, but it presented as flatten mode, which is difficult to understand the caller and callee relationship, if you could store and present it as a call stack, it will be easy to figure out the bottleneck, WDYT?

lmcache/observability.py
yoo-kumaneko: Sure. Will do.

lmcache/observability.py
```diff
@@ -481,16 +631,34 @@ def get_stats_and_clear(self) -> LMCacheStats:
...
+            s.num_tokens for s in finished_retrieve_stats
+        )
+        sum_finished_retrieve_hit_tokens = sum(
+            s.local_hit_tokens + s.remote_hit_tokens for s in finished_retrieve_stats
+        )
         retrieve_hit_rate = (
-            0
-            if self.interval_requested_tokens == 0
-            else self.interval_hit_tokens / self.interval_requested_tokens
+            1
```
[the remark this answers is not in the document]
yoo-kumaneko: No. This is exactly what I want. The success rate should be 1 when there is no retrieval requests.

lmcache/v1/cache_engine.py
maobaolong: Discussed offline, the current state is correct, what we need is getting a time costing detail which affect inference request.

lmcache/observability.py
yoo-kumaneko: done. Now it logs the dictionary.

lmcache/v1/cache_engine.py
```diff
@@ -208,6 +208,12 @@ def __init__(
         self.stats_monitor = LMCStatsMonitor.GetOrCreate()
         # Initialize PinMonitor singleton with config
         PinMonitor.GetOrCreate(config)
+        self.stats_monitor.retrieve_time_threshold = config.get_extra_config_value(
```
maobaolong: Add a comment about unit will helps

yoo-kumaneko: Renamed the variable to retrieve_token_speed_threshold

tests/test_observability.py
```diff
@@ -16,17 +16,17 @@ def test_on_retrieve_request(stats_monitor):
     stats_monitor.on_retrieve_request(num_tokens=100)
     stats = stats_monitor.get_stats_and_clear()
     assert stats.interval_retrieve_requests == 1
-    assert stats.retrieve_hit_rate == 0
+    assert stats.retrieve_hit_rate == 1.0
```
maobaolong, line 19: Can you explain why you did this change?

lmcache/v1/cache_engine.py
```diff
@@ -388,102 +388,104 @@ def store(
...
-                busy_loop=self.force_store_wait,
-                fmt=self.fmt,
-            )
-            if memory_obj is None:
-                logger.warning(
-                    "Local cpu memory under pressure so"
-                    " choosing to store only "
-                    f" {len(memory_objs)}"
-                    " total chunks of KV cache."
+        with store_stats.profile_process_tokens():
```
maobaolong, line 405: Is this the only thing you did?  If so, reviewer can easily skip the inner code changes.

lmcache/observability.py
```diff
@@ -266,40 +377,89 @@ def on_retrieve_request(self, num_tokens: int) -> int:
...
+        retrieve_speed = retrieve_stats.retrieve_speed()
+        if time_to_retrieve > self.retrieve_time_threshold:
+            self.interval_num_slow_retrieval_by_time += 1
+        if 0 < retrieve_speed < self.retrieve_token_speed_threshold:
+            self.interval_num_slow_retrieval_by_speed += 1
+
+        if (
+            time_to_retrieve > self.retrieve_time_threshold
+            or 0 < retrieve_speed < self.retrieve_token_speed_threshold
+        ) and curr_time - self.last_retrieve_warning_time > 10.0:
```
maobaolong: It will help for the following developer who can easily get this condition if you could add a comment.

maobaolong (commented):
@yoo-kumaneko Left some comment.

lmcache/observability.py
maobaolong: You'd better to add the skip log count of those exceed threshold log to the following log.

tests/test_observability.py
yoo-kumaneko: Retrive hit rate (success rate) should be 1 (instead of 0) when there are no finished requests in the interval. Note there's only a on_retrieve_reques call, no on_retrieve_finished call.

lmcache/v1/cache_engine.py
yoo-kumaneko: Yes. This is the only new line. Inner lines are just indented.

lmcache/observability.py
yoo-kumaneko: Sure. will do

lmcache/observability.py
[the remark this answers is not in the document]
ApostaC: Quick follow-up here, what's the correct way to understand the "1" hit rate when there is no retrieve requests? Could we use a special number like -1 to differentiate it from the full hit?

lmcache/observability.py
```diff
@@ -504,6 +685,10 @@ def filter_out_zeros(stats: List[float]):
             [stats.time_to_store() for stats in self.store_requests.values()]
         )
 
+        time_to_lookup = filter_out_zeros(
+            [stats.time_to_lookup() for stats in self.lookup_requests.values()]
```
ApostaC: Just realized, here we are doing another construction of a temporary list, which will have extra overheads.
What we can do is:
```python3
def filter_out_zeros(stats: Iterable[float]) -> List[float]:
    return [x for x in stats if x != 0]

time_to_lookup = filter_out_zeros(
    stats.time_to_lookup() for stats in self.lookup_requests.values()
) # no need to have `[` and `]`
```

ApostaC (commented):
Otherwise LGTM. 
@yoo-kumaneko Since we are getting more and more metrics, it would be great if we can use some AI tools to generate a doc page describing the observability metrics. There is a section called "Use LMCache in production" and we can put it there. WDYT?

lmcache/observability.py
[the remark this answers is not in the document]
yoo-kumaneko: By design, the retrieve hit rate is expected to be 1 when the system is healthy. Any value below 1 indicates a retrieval failure and leads to incorrect inference results. Therefore, a hit rate of 1 signifies the absence of failures. When there are no retrieve requests within a 15-second interval, we also consider the system to be in a healthy state.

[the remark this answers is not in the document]
yoo-kumaneko: While -1 would also work, I think 1 might be a slightly better choice.

lmcache/observability.py
yoo-kumaneko: That makes sense. will do

lmcache/observability.py
yoo-kumaneko: done

yoo-kumaneko:
> Otherwise LGTM. @yoo-kumaneko Since we are getting more and more metrics, it would be great if we can use some AI tools to generate a doc page describing the observability metrics. There is a section called "Use LMCache in production" and we can put it there. WDYT?

Sure. I've added a page to the "Use LMCache in production" section.

maobaolong (approved):
LGTM @yoo-kumaneko Thanks for this good feature. 
BTW, would you like to update the title to be a little more concrete?

ApostaC (approved):
LGTM!

lmcache/v1/cache_engine.py
```diff
@@ -492,12 +500,15 @@ def store(
...
-            tot_kv_size / tot_time / 1024**3,
-            offload_time * 1000,
-            put_time * 1000,
+            tot_kv_size / tot_time / 1024**3 if tot_time > 0 else 0,
+            (store_stats.process_tokens_time + store_stats.from_gpu_time) * 1000,
+            store_stats.put_time * 1000,
         )
 
-        self.stats_monitor.on_store_finished(monitor_req_id, tot_token_num)
+        self.stats_monitor.on_store_finished(
```
FeiGSSS, line 502: BUG here! Should first call `self.stats_monitor.on_store_finished` and then compute `tot_time = store_stats.time_to_store()`, otherwise, `self.end_time` in `self.stats_monitor` is default to 0, results `tot_time =0`  @yoo-kumaneko

lmcache/v1/cache_engine.py
```diff
@@ -792,10 +806,13 @@ def retrieve(
                 self.storage_manager.remove(key)
             memory_obj.ref_count_down()
 
-        onload_time = time.perf_counter() - t
+        onload_time = retrieve_stats.time_to_retrieve()
```
FeiGSSS, line 803: `onload_time = retrieve_stats.time_to_retrieve()` is always 0 since `self.stats_monitor.on_retrieve_finished` has not been called. @yoo-kumaneko

lmcache/v1/cache_engine.py
yoo-kumaneko: Got it. Thanks so much!

lmcache/v1/cache_engine.py
yoo-kumaneko: Got it. Thanks for the sharp eye!

yoo-kumaneko: Bugfix PR here.
https://github.com/LMCache/LMCache/pull/2485

PR #2468, opened 2026-01-21
[CI/Build] P2P verbose log and add trim to long_doc_qa

Shaoting-Feng:
Only minor changes.

1. Make the P2P first engine logs more verbose.

Previously, the logs looked like this:

image

This is not correct. In the P2P test case, only the second engine is used for performance measurement. Therefore, the first engine is not expected to have any baselines to compare against, and its logs should not raise a warning.

2. Add a trim option for calculating the mean TTFT.

Currently, we compute the round time as the total elapsed time divided by the number of requests. This is not equivalent to taking the mean of per-round times. As a result, trimming does not work for the round-time-per-prompt metric.

**If applicable**:


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

Issue #2458, opened 2026-01-20
[Refactor]: Northbound


sammshen:
The "refactor" of the "northbound" can be split into a couple of steps that targets: 
- code cleanliness (modular design)
- correctness (different KV cache formats and integrations)

## 1. Clearly define the format of `CacheEngineKey` and `TensorMemoryObj` going forward (key discussion: https://github.com/LMCache/LMCache/pull/2472#issuecomment-3803669896 finalizing `2LTD` and removing `fmt`)
- [x] Renaming + Deprecating attributes in LMCacheMetadata and CacheEngineKey https://github.com/LMCache/LMCache/pull/2472
- [ ] Documentation Example of cross KV sharing between vLLM and SGLang (through a shared local remote or disk backend)

## 2. Extensible GPUConnector Interface for more integrations in the future
- [x] Create a new folder for GPU Connectors and centralized `CreateGPUConnector` + logic to find local vs global worker ids / world sizes (https://github.com/LMCache/LMCache/pull/2530)
- [x] Introduce GPUKVFormat and handle flash infer and triton attention properly (https://github.com/LMCache/LMCache/pull/2567)
- [x] Add CacheEngineType Enum: https://github.com/LMCache/LMCache/pull/2555
- [ ] vLLM cross layer mode
    - Original PR: https://github.com/LMCache/LMCache/pull/2498
- [x] Add MLA to SGLangLayerwise (https://github.com/LMCache/LMCache/pull/2629)
- [x] Support HND (https://github.com/LMCache/LMCache/pull/2826)

## 3. Make the `LMCacheManager` usable by all adapters (first "failed" attempt for reference: https://github.com/LMCache/LMCache/pull/2457)
- [ ] Introduce `ServiceFactory` and make `LMCacheManager` a container for `LMCacheEngine` and `ServiceFactory` (https://github.com/LMCache/LMCache/pull/2683)
- [ ] Construct documentation for the LMCacheManager that clearly highlights how to use OffloadServer, InternalAPIServer, RuntimePluginLauncher, HealthMonitor, PeriodicThreadLauncher

## 4. (vllm-specific) Create a compatibility layer for `vllm_v1_adapter.py` and protect the compatibility layer with CI on the vllm side
- can we merge some modules in the MP mode and the non-MP mode in this compatibility layer? 

## 5. General Code Cleanliness: 
- [ ] Remove unused v0 code
- [ ] Move multiprocess and standalone to `lmcache/integration` folder

[...]

PR #2008, opened 2025-11-17
[Core][1/N]  transpose-kernel only for asym TP

novahow:
**What this PR does / why we need it**:

Transpose kernel dependency-only PR. 
Sub-PR of [#1629 ]

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #2472, opened 2026-01-22
[Northbound 1/N]: Clean up LMCacheMetadata

sammshen:
From discussion in https://github.com/LMCache/LMCache/pull/2457, we have decided to split the serving engine agnostic manager into 4 PRs (could be more). 

This first PR is the dirty PR to deprecate the `num_ranks` field (`world_size` already has TP x PP information) and to change the previous `fmt` attribute to `use_case`, which will allow us to make the Metadata data structure more useful in the future when increasing the northbound integrations (e.g. changing field exposure based on `use_case`).

This PR touches many files but most of them are unit tests.

Finally, this PR: 
1. renames `LMCacheEngineMetadata` to `LMCacheMetadata` in preparation of the metadata primarily being consumed by the new `LMCacheManager` class
2. moves the `LMCacheMetadata` from `lmcache/config.py` to `lmcache/v1/metadata.py` in preparation to deprecate v0 code in the future.

**EDIT**
From @ApostaC's suggestion, also added `local_world_size` and `local_worker_id` fields to the metadata. In single-node cases, these will be identical to `world_size` and `worker_id`

### DIFF
--- a/docs/source/kv_cache/storage_backends/redis.rst
+++ b/docs/source/kv_cache/storage_backends/redis.rst
@@ -297,9 +297,9 @@ LMCache stores data in Redis using a structured key format. Each key contains th
 
 .. code-block:: text
 
-    format@model_name@world_size@worker_id@chunk_hash
+    use_case@model_name@world_size@worker_id@chunk_hash
 
-- `format`: The model format (e.g., "vllm" or "huggingface")
+- `use_case`: The use case of the model (e.g., "vllm" or "huggingface" or "sglang" or "standalone")
 - `model_name`: Name of the language model
 - `world_size`: Total number of workers in distributed deployment
 - `worker_id`: ID of the worker that created this cache entry
--- a/docs/source/kv_cache/storage_backends/redis.rst
+++ b/docs/source/kv_cache/storage_backends/redis.rst
@@ -297,12 +297,11 @@ LMCache stores data in Redis using a structured key format. Each key contains th
 
 .. code-block:: text
 
-    use_case@model_name@world_size@worker_id@chunk_hash
+    model_name@world_size@worker_id@chunk_hash
 
-- `use_case`: The use case of the model (e.g., "vllm" or "huggingface" or "sglang" or "standalone")
 - `model_name`: Name of the language model
 - `world_size`: Total number of workers in distributed deployment
-- `worker_id`: ID of the worker that created this cache entry
+- `worker_id`: ID of the worker that created this cache entry, in the range of [0, world_size - 1]
 - `chunk_hash`: Hash of the token chunk (SHA-256 based)
 
 For example, a typical key might look like:
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -216,7 +216,6 @@ def create_lmcache_metadata(
             )
 
     # Create metadata
-    num_ranks = parallel_cfg.tensor_parallel_size * parallel_cfg.pipeline_parallel_size
     metadata = LMCacheEngineMetadata(
         model_cfg.model,
         parallel_cfg.world_size,
@@ -228,7 +227,6 @@ def create_lmcache_metadata(
         role,
         served_model_name=model_cfg.served_model_name,
         engine_id=engine_id,
-        num_ranks=num_ranks,
         kv_connector_extra_config=kv_connector_extra_config,
     )
 
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -217,14 +217,14 @@ def create_lmcache_metadata(
 
     # Create metadata
     metadata = LMCacheMetadata(
-        model_cfg.model,
-        parallel_cfg.world_size,
-        parallel_cfg.rank,
-        "vllm",
-        kv_dtype,
-        kv_shape,
-        use_mla,
-        role,
+        model_name=model_cfg.model,
+        world_size=parallel_cfg.world_size,
+        worker_id=parallel_cfg.rank,
+        use_case="vllm",
+        kv_dtype=kv_dtype,
+        kv_shape=kv_shape,
+        use_mla=use_mla,
+        role=role,
         served_model_name=model_cfg.served_model_name,
         engine_id=engine_id,
         kv_connector_extra_config=kv_connector_extra_config,
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -222,7 +222,6 @@ def create_lmcache_metadata(
         local_world_size=parallel_cfg.world_size,
         worker_id=parallel_cfg.rank,
         local_worker_id=parallel_cfg.rank,
-        use_case="vllm",
         kv_dtype=kv_dtype,
         kv_shape=kv_shape,
         use_mla=use_mla,
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -154,7 +154,7 @@ def create_lmcache_metadata(
     role=None,
 ):
     """
-    Create LMCacheEngineMetadata from vLLM configuration.
+    Create LMCacheMetadata from vLLM configuration.
 
     This function extracts common metadata creation logic that was duplicated
     across multiple files.
@@ -167,7 +167,7 @@ def create_lmcache_metadata(
         cache_config: Cache configuration (alternative to vllm_config)
 
     Returns:
-        tuple: (LMCacheEngineMetadata, LMCacheEngineConfig)
+        tuple: (LMCacheMetadata, LMCacheEngineConfig)
     """
     # Third Party
     # Try to import from old location before merged https://github.com/vllm-project/vllm/pull/26908
@@ -178,7 +178,7 @@ def create_lmcache_metadata(
         # Third Party
         from vllm.utils import get_kv_cache_torch_dtype
     # First Party
-    from lmcache.config import LMCacheEngineMetadata
+    from lmcache.config import LMCacheMetadata
 
     config = lmcache_get_or_create_config()
     # Support both vllm_config object and individual config parameters
@@ -216,7 +216,7 @@ def create_lmcache_metadata(
             )
 
     # Create metadata
-    metadata = LMCacheEngineMetadata(
+    metadata = LMCacheMetadata(
         model_cfg.model,
         parallel_cfg.world_size,
         parallel_cfg.rank,
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -178,7 +178,7 @@ def create_lmcache_metadata(
         # Third Party
         from vllm.utils import get_kv_cache_torch_dtype
     # First Party
-    from lmcache.config import LMCacheMetadata
+    from lmcache.v1.metadata import LMCacheMetadata
 
     config = lmcache_get_or_create_config()
     # Support both vllm_config object and individual config parameters
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -219,7 +219,9 @@ def create_lmcache_metadata(
     metadata = LMCacheMetadata(
         model_name=model_cfg.model,
         world_size=parallel_cfg.world_size,
+        local_world_size=parallel_cfg.world_size,
         worker_id=parallel_cfg.rank,
+        local_worker_id=parallel_cfg.rank,
         use_case="vllm",
         kv_dtype=kv_dtype,
         kv_shape=kv_shape,
--- a/lmcache/storage_backend/serde/cachegen_decoder.py
+++ b/lmcache/storage_backend/serde/cachegen_decoder.py
@@ -128,7 +128,6 @@ class CacheGenDeserializer(Deserializer):
         self.cachegen_config = CacheGenConfig.from_model_name(metadata.model_name)
         self.chunk_size = config.chunk_size
         self.output_buffer: Optional[torch.Tensor] = None
-        self.use_case = metadata.use_case
         self.key_bins = self.make_key_bins(self.cachegen_config)
         self.value_bins = self.make_value_bins(self.cachegen_config)
 
@@ -202,14 +201,11 @@ class CacheGenDeserializer(Deserializer):
                 encoder_output.head_size,
             )
         )
-        match self.use_case:
-            case "vllm":
-                return blob.permute((1, 0, 2, 3, 4)).to(
-                    self.dtype
-                )  # [nlayers, 2, ntokens, num_heads, head_size]
-            case "huggingface":
-                return blob.permute((1, 0, 3, 2, 4)).to(
-                    self.dtype
-                )  # [nlayers, 2, num_heads, ntokens, head_size]
-            case _:
-                raise RuntimeError("Unknown use case %s" % self.use_case)
+
+        return blob.permute((1, 0, 2, 3, 4)).to(
+            self.dtype
+        )  # [nlayers, 2, ntokens, num_heads, head_size]
+        # huggingface
+        # return blob.permute((1, 0, 3, 2, 4)).to(
+        #     self.dtype
+        # )  # [nlayers, 2, num_heads, ntokens, head_size]
--- a/lmcache/storage_backend/serde/cachegen_decoder.py
+++ b/lmcache/storage_backend/serde/cachegen_decoder.py
@@ -6,7 +6,7 @@ from typing import List, Optional
 import torch
 
 # First Party
-from lmcache.config import LMCacheEngineMetadata
+from lmcache.config import LMCacheMetadata
 from lmcache.logging import init_logger
 from lmcache.storage_backend.serde.cachegen_basics import (
     CacheGenConfig,
@@ -121,7 +121,7 @@ class CacheGenDeserializer(Deserializer):
     def __init__(
         self,
         config: LMCacheEngineConfig,
-        metadata: LMCacheEngineMetadata,
+        metadata: LMCacheMetadata,
         dtype,
     ):
         self.dtype = dtype
--- a/lmcache/storage_backend/serde/cachegen_decoder.py
+++ b/lmcache/storage_backend/serde/cachegen_decoder.py
@@ -6,7 +6,6 @@ from typing import List, Optional
 import torch
 
 # First Party
-from lmcache.config import LMCacheMetadata
 from lmcache.logging import init_logger
 from lmcache.storage_backend.serde.cachegen_basics import (
     CacheGenConfig,
@@ -16,6 +15,7 @@ from lmcache.storage_backend.serde.cachegen_basics import (
 from lmcache.storage_backend.serde.serde import Deserializer
 from lmcache.utils import _lmcache_nvtx_annotate
 from lmcache.v1.config import LMCacheEngineConfig
+from lmcache.v1.metadata import LMCacheMetadata
 
 if torch.cuda.is_available():
     import lmcache.c_ops as lmc_ops
--- a/lmcache/storage_backend/serde/cachegen_decoder.py
+++ b/lmcache/storage_backend/serde/cachegen_decoder.py
@@ -128,7 +128,7 @@ class CacheGenDeserializer(Deserializer):
         self.cachegen_config = CacheGenConfig.from_model_name(metadata.model_name)
         self.chunk_size = config.chunk_size
         self.output_buffer: Optional[torch.Tensor] = None
-        self.fmt = metadata.fmt
+        self.use_case = metadata.use_case
         self.key_bins = self.make_key_bins(self.cachegen_config)
         self.value_bins = self.make_value_bins(self.cachegen_config)
 
@@ -202,7 +202,7 @@ class CacheGenDeserializer(Deserializer):
                 encoder_output.head_size,
             )
         )
-        match self.fmt:
+        match self.use_case:
             case "vllm":
                 return blob.permute((1, 0, 2, 3, 4)).to(
                     self.dtype
@@ -212,4 +212,4 @@ class CacheGenDeserializer(Deserializer):
                     self.dtype
                 )  # [nlayers, 2, num_heads, ntokens, head_size]
             case _:
-                raise RuntimeError("Unknown format %s" % self.fmt)
+                raise RuntimeError("Unknown use case %s" % self.use_case)
--- a/lmcache/storage_backend/serde/cachegen_encoder.py
+++ b/lmcache/storage_backend/serde/cachegen_encoder.py
@@ -6,7 +6,6 @@ from typing import Dict, Tuple
 import torch
 
 # First Party
-from lmcache.config import LMCacheMetadata
 from lmcache.logging import init_logger
 from lmcache.storage_backend.serde.cachegen_basics import (
     CacheGenConfig,
@@ -16,6 +15,7 @@ from lmcache.storage_backend.serde.cachegen_basics import (
 from lmcache.storage_backend.serde.serde import Serializer
 from lmcache.utils import _lmcache_nvtx_annotate
 from lmcache.v1.config import LMCacheEngineConfig
+from lmcache.v1.metadata import LMCacheMetadata
 
 if torch.cuda.is_available():
     import lmcache.c_ops as lmc_ops
--- a/lmcache/storage_backend/serde/cachegen_encoder.py
+++ b/lmcache/storage_backend/serde/cachegen_encoder.py
@@ -6,7 +6,7 @@ from typing import Dict, Tuple
 import torch
 
 # First Party
-from lmcache.config import LMCacheEngineMetadata
+from lmcache.config import LMCacheMetadata
 from lmcache.logging import init_logger
 from lmcache.storage_backend.serde.cachegen_basics import (
     CacheGenConfig,
@@ -340,7 +340,7 @@ def encode_function(
 
 
 class CacheGenSerializer(Serializer):
-    def __init__(self, config: LMCacheEngineConfig, metadata: LMCacheEngineMetadata):
+    def __init__(self, config: LMCacheEngineConfig, metadata: LMCacheMetadata):
         self.cachegen_config = CacheGenConfig.from_model_name(metadata.model_name)
         self.chunk_size = config.chunk_size
         self.use_case = metadata.use_case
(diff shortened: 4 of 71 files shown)


tests/v1/test_freeze_mode_integration.py
```diff
@@ -63,9 +63,7 @@ def test_freeze_with_real_cache_engine(autorelease_v1):
     # Create mock GPU connector that works on CPU
     connector = MockGPUConnector(kv_shape=kv_shape)
     # Use explicit metadata with worker_id=0 to control port offset
-    metadata = LMCacheEngineMetadata(
-        "test_model", 1, 0, "vllm", torch.bfloat16, kv_shape
-    )
+    metadata = LMCacheMetadata("test_model", 1, 0, "vllm", torch.bfloat16, kv_shape)
```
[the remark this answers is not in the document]
sammshen: addressed, thanks!

tests/v1/utils.py
```diff
@@ -29,14 +29,14 @@ def recover_gpu_connector_states(gpu_connector):
     gpu_connector.kv_cache_pointers_on_gpu = {}
 
 
-def dumb_metadata(fmt="vllm", kv_shape=(32, 2, 256, 8, 128)):
-    return LMCacheEngineMetadata("test_model", 3, 123, fmt, torch.bfloat16, kv_shape)
+def dumb_metadata(use_case="vllm", kv_shape=(32, 2, 256, 8, 128)):
+    return LMCacheMetadata("test_model", 3, 123, use_case, torch.bfloat16, kv_shape)
```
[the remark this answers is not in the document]
sammshen: addressed, thanks!

tests/v1/utils.py
```diff
@@ -29,14 +29,14 @@ def recover_gpu_connector_states(gpu_connector):
...
 
 
 def dumb_metadata_with_model_name(
-    model_name: str, fmt="vllm", kv_shape=(32, 2, 256, 8, 128)
+    model_name: str, use_case="vllm", kv_shape=(32, 2, 256, 8, 128)
 ):
-    return LMCacheEngineMetadata(model_name, 3, 123, fmt, torch.bfloat16, kv_shape)
+    return LMCacheMetadata(model_name, 3, 123, use_case, torch.bfloat16, kv_shape)
```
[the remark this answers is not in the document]
sammshen: addressed thanks!

ApostaC:
Will there be any special cases for PP?
If we use PP=2, TP=8, each node (PP instance) will have 8 TP worker. And within each PP instance, the rank for communication is 0-7. 

We cannot use the global rank (0-15) unless we use something like TP=16 on Ray.

sammshen:
@ApostaC There is already an edge case in the `SGLangAdapter` for this:

```python
    # Change current device using local GPU index
    torch.cuda.device(local_rank)
    device = torch.device(f"cuda:{local_rank}")
    # Use global rank for metadata (tensor parallel rank)
    metadata = LMCacheMetadata(
        model_name=model_config.model_path,
        world_size=tp_size,
        local_world_size=tp_size, # Adding in this PR now
        worker_id=global_rank,
        local_worker_id=local_rank, # Adding in this PR now
        use_case="sglang",
        kv_dtype=kv_dtype,
        kv_shape=kv_shape,
    )
```

Thanks for the great suggestion!

Will add this to the `LMCacheMetadata` for future multi-node logic handling:
```python
    """ global world size when running under a distributed setting (total number of workers)"""
    world_size: int
    """ host world size (workers on active localhost)
    This information can be useful for multi-node
    deployment. Will be the same as world_size in single-node deployments.
    """
    local_world_size: int
    """ worker id when running under a distributed setting """
    worker_id: int
    """ host worker id (a gpu bound worker id on active localhost)
    This information can be useful for multi-node deployment. 
    Will be the same as worker_id in single-node deployments.
    """
    local_worker_id: int
 ```
 
 The main use case of `num_ranks` was in lookup client and lookup server and that semantics **does** need TP x PP even cross-node because lookup results need to match for **all** workers.

File: lmcache/integration/vllm/utils.py [hunk 4]
```diff
@@ -216,19 +216,19 @@ def create_lmcache_metadata(
...
-        kv_dtype,
-        kv_shape,
-        use_mla,
-        role,
+    metadata = LMCacheMetadata(
+        model_name=model_cfg.model,
+        world_size=parallel_cfg.world_size,
+        local_world_size=parallel_cfg.world_size,
+        worker_id=parallel_cfg.rank,
+        local_worker_id=parallel_cfg.rank,
```
ApostaC, line 224: What will be the case for PP here?

lmcache/utils.py
```diff
@@ -318,7 +318,7 @@ def parse_cache_key(key_str: str) -> Union[CacheEngineKey, LayerCacheEngineKey]:
 
     Args:
         key_str: String in format:
-            fmt@model@world_size@worker_id@chunk_hash[@layer_id][@tag%value...]
+            use_case@model@world_size@worker_id@chunk_hash[@layer_id][@tag%value...]
```
ApostaC: Not requesting changes here, but if we want to share KV caches between vLLM and SGLang in the future, should we not put the use case in the key?

lmcache/v1/manager.py
```diff
@@ -308,23 +308,20 @@ def _create_lmcache_engine(self, role: str) -> LMCacheEngine:
...
-            parallel_config.rank,
-            "vllm",
-            kv_dtype,
-            kv_shape,
-            use_mla,
-            role,
+        metadata = LMCacheMetadata(
+            model_name=model_config.model,
+            world_size=parallel_config.world_size,
+            local_world_size=parallel_config.world_size,
```
ApostaC, line 351: Should we use tp_size * pp_size and tp_size for these two?

Also need some logic for worker_id

ApostaC (changes requested):
Thanks for the huge effort! Some small issue about local world size and local rank. Otherwise LGTM!

lmcache/integration/vllm/utils.py
sammshen: in multi-node case, we should pass in the TP rank for `local_worker_id` and the global rank for `worker_id` and tensor_parallel_size for `local_world_size` (assuming nodes are by PP and TP is intra-node)

lmcache/utils.py
sammshen: that's a really good point. I vote to replace it with memory_fmt? but then every constructor of CacheEngineKeys needs to remember the memory format of its context. this feels unnatural. 

also, there is a very unlikely edge case where the CEK between a TP 2 PP 8 and a TP 8 PP 2 engine share cache key when they shoudl not

lmcache/v1/manager.py
sammshen: is it always guaranteed that TP is intra-node? is it always guaranteed that PP is inter-node? can TP be across nodes and can PP be within a node?

sammshen:
@ApostaC I left both of your comments as **HELP NEEDED** in the northbound tracker: https://github.com/LMCache/LMCache/issues/2458 . I think these can be separate PRs that can actually have active discussion while other threads in Northbound refactor progress!

DongDongJu:
Hello @sammshen, I will take a look this today.

lmcache/utils.py
sammshen: I see this is basically what @DongDongJu has discussed in https://docs.google.com/document/d/14GimHIhxH8IRryonjyUi0i6plO9PIuKoqpWDc_u23pM/edit?tab=t.0#heading=h.omhujaenleiq

DongDongJu (commented):
Hello @sammshen,

From my understanding,
this one related with 
- rename LMCacheEngineMetadata -> LMCacheMetadata.
- better multi instance support (local_world_size, local_worker_id).

One big thing unclear to me is that
`fmt` can be equivalent with `use_cases`?
If we want to say fmt is KVCache format then we need both('fmt' and 'use_cases') because fmt can change based on model not use_case.
Please correct me if Im wrong in here.

We can talk in offline for this.

sammshen:
@DongDongJu @ApostaC 

Before continuing with this PR (and northbound refactor in general), I just want to clarify a couple of points. I think these two points will also be relevant to the multiprocess mode progress. 

## Discussion Point 1. Main proposal for this PR (removing `use_case`)

I think it is reasonable to remove the `use_case` (previously `fmt`) from the default (non multiprocess) use case entirely (which unifies it with the multiprocess mode). the reasoning is as follows: 

- the northbound use case is implicitly tied to how LMCache is started up (vllm, sglang, standalone). any logic that belongs to a specific northbound interface can live inside of adapters (or at least under the `lmcache/integration/` folder). In the follow up PRs belonging to https://github.com/LMCache/LMCache/issues/2458, I will add `Adapter`s and `MetadataBuilder`s that will not need `use_case`.
- two serving engines should be able to retrieve the same KV Cache from remote backend (shared KV pool) as long as the CacheEngineKey is the same. How the equality is implemented for CacheEngineKey will be the Discussion Point 2. 

```
    def __eq__(self, other):
               self.model_name == other.model_name
                and self.world_size == other.world_size
                and self.worker_id == other.worker_id
                and self.chunk_hash == other.chunk_hash
                and self.dtype == other.dtype
                and self.tags == other.tags
```


## Discussion Point 2. Asking 4 Critical Questions (in 2 different ways)

Regarding the big picture, I think there are two main directions. For each direction, there are two considerations (same for each):

### Direction 1. the KV Cache layout of the northbound serving engine. Let's call this `northbound_layout`
- (QUESTION 1, Ask Method 1) Consideration 1: does `northbound_layout` need to be inside of the CacheEngineKey?
- (QUESTION 2, Ask Method 1) Consideration 2: does `northbound_layout` need to be inside of the MemoryObjMetadata?

### Direction 2. the KV Cache layout of a memory object (the striding of `2`, `L`, `T`, and `D` and MLA cases) . Let's call this `mem_obj_layout`
- (QUESTION 3, Ask Method 1) Consideration 1: does `mem_obj_layout` need to be inside of the CacheEngineKey?
- (QUESTION 4, Ask Method 1) Consideration 2: does `mem_obj_layout` need to be inside of the MemoryObjMetadata? 

Currently, we only have one GPU connector per LMCacheEngine. Which means we do not support routing logic based on the `northbound_layout` and the `mem_obj_layout` (we assume that the same `northbound_layout` and `mem_obj_layout` will be used for one cache engine). 

## Summary / TL;DR:

We have to answer four questions for both multiprocess and regular LMCache mode.  These are the same four questions being answered above:

- (QUESTION 1, Ask Method 2). do we want to prevent serving engines with different KV Cache Layouts to be able to consume the same LMCache KV Cache? 

- (QUESTION 2, Ask Method 2). do we want a single LMCache Engine to be able to talk to multiple serving engine KV Cache Layouts (multiple GPU Connectors / transfer kernel paths)? 

- (QUESTION 3, Ask Method 2). do we want KV Cache of different formats (e.g. `2LTD` `D2LT`) to be considered different?

- (QUESTION 4, Ask Method 2). If so, do we need to change back to a given format (e.g. `2LTD`) before sending them to GPU connector? (if not, we can add a transpose like https://github.com/LMCache/LMCache/pull/2008/)


## My proposal for these four questions for normal / non-multiprocess mode: 

- (QUESTION 1, Ask Method 1 and 2). do we want to prevent serving engines with different KV Cache Layouts to be able to consume the same LMCache KV Cache? Same question as: does `northbound_layout` need to be inside of the CacheEngineKey?

**Proposed Answer**: **NO (current impl is YES)**  <- This is why I propose we remove `use_case` entirely!

- (QUESTION 2, Ask Method 1 and 2). do we want a single LMCache Engine to be able to talk to multiple serving engine KV Cache Layouts (multiple GPU Connectors / transfer kernel paths)? Same question as: does `northbound_layout` need to be inside of the MemoryObjMetadata? 

**Proposed Answer**: **NO (current impl is NO)** <- However, multiprocess mode should add this in `GPUCacheContext` on KV Cache Registration (talked to @ApostaC offline about this)

- (QUESTION 3, Ask Method 1 and 2). do we want KV Cache of different formats (e.g. `2LTD` `D2LT`) to be considered different? Same question as: does `mem_obj_layout` need to be inside of the CacheEngineKey?  

**Proposed Answer:** **POTENTIALLY YES (current impl is NO)** <- can be NO for the normal case (there will be bugs if we retrieve differently layed out KV Cache)

- (QUESTION 4, Ask Method 1 and 2). do we need to change back to a given format (e.g. `2LTD`) before sending them to GPU connector? Same question as: does `mem_obj_layout` need to be inside of the MemoryObjMetadata? 
- 
**Proposed Answer:** **YES (current impl is basically YES since GPU Connectors force consumption of one `mem_obj_layout`, namely `2LTD` and `2TD` for non-layerwise and layerwise)** <- However, if we want to have flexible formats, we probably need something similar to a transpose kernel as mentioned in https://github.com/LMCache/LMCache/pull/2008/)

### Final Design Proposal / Changes Needed for this specific PR: 
- remove `use_case` and it's main downstream use in `CacheEngineKey`

DongDongJu:
> @DongDongJu @ApostaC
> 
> Before continuing with this PR (and northbound refactor in general), I just want to clarify a couple of points. I think these two points will also be relevant to the multiprocess mode progress.
> 
> ## Discussion Point 1. Main proposal for this PR (removing `use_case`)
> I think it is reasonable to remove the `use_case` (previously `fmt`) from the default (non multiprocess) use case entirely (which unifies it with the multiprocess mode). the reasoning is as follows:
> 

I do agree.

> * the northbound use case is implicitly tied to how LMCache is started up (vllm, sglang, standalone). any logic that belongs to a specific northbound interface can live inside of adapters (or at least under the `lmcache/integration/` folder). In the follow up PRs belonging to [[Core]: Refactor Northbound Interface #2458](https://github.com/LMCache/LMCache/issues/2458), I will add `Adapter`s and `MetadataBuilder`s that will not need `use_case`.

👍 

> * two serving engines should be able to retrieve the same KV Cache from remote backend (shared KV pool) as long as the CacheEngineKey is the same. How the equality is implemented for CacheEngineKey will be the Discussion Point 2.
> 
> ```
>     def __eq__(self, other):
>                self.model_name == other.model_name
>                 and self.world_size == other.world_size
>                 and self.worker_id == other.worker_id
>                 and self.chunk_hash == other.chunk_hash
>                 and self.dtype == other.dtype
>                 and self.tags == other.tags
> ```
> 
> ## Discussion Point 2. Asking 4 Critical Questions (in 2 different ways)
> Regarding the big picture, I think there are two main directions. For each direction, there are two considerations (same for each):
> 
> ### Direction 1. the KV Cache layout of the northbound serving engine. Let's call this `northbound_layout`
> * (QUESTION 1, Ask Method 1) Consideration 1: does `northbound_layout` need to be inside of the CacheEngineKey?
> * (QUESTION 2, Ask Method 1) Consideration 2: does `northbound_layout` need to be inside of the MemoryObjMetadata?
> 
> ### Direction 2. the KV Cache layout of a memory object (the striding of `2`, `L`, `T`, and `D` and MLA cases) . Let's call this `mem_obj_layout`
> * (QUESTION 3, Ask Method 1) Consideration 1: does `mem_obj_layout` need to be inside of the CacheEngineKey?
> * (QUESTION 4, Ask Method 1) Consideration 2: does `mem_obj_layout` need to be inside of the MemoryObjMetadata?
> 
> Currently, we only have one GPU connector per LMCacheEngine. Which means we do not support routing logic based on the `northbound_layout` and the `mem_obj_layout` (we assume that the same `northbound_layout` and `mem_obj_layout` will be used for one cache engine).
> 
> ## Summary / TL;DR:
> We have to answer four questions for both multiprocess and regular LMCache mode. These are the same four questions being answered above:
> 
> * (QUESTION 1, Ask Method 2). do we want to prevent serving engines with different KV Cache Layouts to be able to consume the same LMCache KV Cache?
> * (QUESTION 2, Ask Method 2). do we want a single LMCache Engine to be able to talk to multiple serving engine KV Cache Layouts (multiple GPU Connectors / transfer kernel paths)?
> * (QUESTION 3, Ask Method 2). do we want KV Cache of different formats (e.g. `2LTD` `D2LT`) to be considered different?
> * (QUESTION 4, Ask Method 2). If so, do we need to change back to a given format (e.g. `2LTD`) before sending them to GPU connector? (if not, we can add a transpose like [[Core][1/N]  transpose-kernel only for asym TP #2008](https://github.com/LMCache/LMCache/pull/2008))
> 
> ## My proposal for these four questions for normal / non-multiprocess mode:
> * (QUESTION 1, Ask Method 1 and 2). do we want to prevent serving engines with different KV Cache Layouts to be able to consume the same LMCache KV Cache? Same question as: does `northbound_layout` need to be inside of the CacheEngineKey?

IMO no, CacheEngineKey should represent the logical cache identity. northbound_layout is an interface detail (producer/consumer layout). 
Putting it in the key prevents cross-engine reuse even when conversion is possible. 
This matches your proposed direction.

> 
> **Proposed Answer**: **NO (current impl is YES)** <- This is why I propose we remove `use_case` entirely!
> 
> * (QUESTION 2, Ask Method 1 and 2). do we want a single LMCache Engine to be able to talk to multiple serving engine KV Cache Layouts (multiple GPU Connectors / transfer kernel paths)? Same question as: does `northbound_layout` need to be inside of the MemoryObjMetadata?

Hmm, only if LMCache needs it to correctly convert stored data back to the caller layout.
If the storage representation is canonical, the consumer’s adapter/connector can produce its own layout, 
and we don’t need to store producer northbound_layout in metadata.
If storage representation is not canonical (or we support multiple producer layouts), 
then we must carry a layout id somewhere in metadata so decode is correct.

> 
> **Proposed Answer**: **NO (current impl is NO)** <- However, multiprocess mode should add this in `GPUCacheContext` on KV Cache Registration (talked to @ApostaC offline about this)
> 
> * (QUESTION 3, Ask Method 1 and 2). do we want KV Cache of different formats (e.g. `2LTD` `D2LT`) to be considered different? Same question as: does `mem_obj_layout` need to be inside of the CacheEngineKey?
> 

Not if we enforce a single canonical mem-object layout per backend, yes if we allow multiple.

If LMCache guarantees the stored memory objects are always in one canonical mem_obj_layout (e.g., the connector enforces it), then it does not belong in the key.

If multiple mem_obj layouts could be present in the same backend namespace, we either:
include mem_obj_layout in the key (hard partition), or
implement conversion/transpose during load/store and keep it in metadata.


> **Proposed Answer:** **POTENTIALLY YES (current impl is NO)** <- can be NO for the normal case (there will be bugs if we retrieve differently layed out KV Cache)
> 
> * (QUESTION 4, Ask Method 1 and 2). do we need to change back to a given format (e.g. `2LTD`) before sending them to GPU connector? Same question as: does `mem_obj_layout` need to be inside of the MemoryObjMetadata?
> 

IMO, Yes, 
even if we don’t use it for routing today, including it in MemoryObjMetadata lets us validate compatibility at load time,
evolve toward transpose/conversion later without changing storage semantics and avoid consumption bugs.

> **Proposed Answer:** **YES (current impl is basically YES since GPU Connectors force consumption of one `mem_obj_layout`, namely `2LTD` and `2TD` for non-layerwise and layerwise)** <- However, if we want to have flexible formats, we probably need something similar to a transpose kernel as mentioned in #2008)
> 
> ### Final Design Proposal / Changes Needed for this specific PR:
> * remove `use_case` and it's main downstream use in `CacheEngineKey`

It seems all the direction promising to me. I am currently testing the KVCacheFormat by multiple models with tp and pp changes vllm and sglang. I will make the PR soon.

ApostaC:
Some additional notes:
In the long run, I think we should keep only one format when storing KV caches as chunks in LMCache storage manager. I see the possibility that we only use `2LTD` for the memory objects in the storage manager.

| Case | Current status | Proposal for future |
|--------|--------|--------|
| Layer-wise  | Each layer is stored as a different object with a different key | Data of all the layers should be merged as the same object and only has one key. Layerwise logic should be processed by GPU connector. |
| Hybrid model (e.g., DS v32) | One object stores all the layers with multiple shapes/dtypes | One object stores all the layers with multiple shapes/dtypes, **plus a clear spec of how do we store those layers**  |


In this case, we no-longer need `memory_obj_layout` in cache engine key. Probably still need that in memory object metadata because a lot of modules need to use that to calculate the size of the memory object.


cc @DongDongJu @sammshen

sammshen:
@ApostaC @DongDongJu 

Final Summary of PR: 

| Change | Purpose |
|--------|--------|
| Rename `LMCacheEngineMetadata` to `LMCacheMetadata` and `LMCacheEngineConfig` to `LMCacheConfig` | `LMCacheManager` will be the interface for adapters instead of `LMCacheEngine` |
| moves the LMCacheMetadata from lmcache/config.py to lmcache/v1/metadata.py | Prepare to deprecate v0 code |
| remove num_ranks from LMCacheMetadata and introduce `local_worker_id` and `local_world_size` | future logic belonging to vllm / sglang can decide when local indices diverge from global ones | 
| Remove `fmt` from Metadata, CacheEngineKey, and CacheGen codepath | different serving engines should be able to consume the same KV Cache |

Concrete next steps for northbound refactor are all listed in https://github.com/LMCache/LMCache/issues/2458

sammshen:
mp tests will be fixed by https://github.com/LMCache/LMCache/pull/2499

ApostaC (approved):
LGTM as we discussed offline

docs/source/kv_cache/storage_backends/redis.rst
```diff
@@ -297,12 +297,11 @@ LMCache stores data in Redis using a structured key format. Each key contains th
...
 .. code-block:: text
 
-    format@model_name@world_size@worker_id@chunk_hash
+    model_name@world_size@worker_id@chunk_hash
 
-- `format`: The model format (e.g., "vllm" or "huggingface")
 - `model_name`: Name of the language model
 - `world_size`: Total number of workers in distributed deployment
-- `worker_id`: ID of the worker that created this cache entry
+- `worker_id`: ID of the worker that created this cache entry, in the range of [0, world_size - 1]
```
DongDongJu, line 304: 👍

lmcache/storage_backend/serde/cachegen_decoder.py
```diff
@@ -202,14 +201,11 @@ def from_bytes(self, bs: bytes) -> torch.Tensor:
...
-                return blob.permute((1, 0, 3, 2, 4)).to(
-                    self.dtype
-                )  # [nlayers, 2, num_heads, ntokens, head_size]
-            case _:
-                raise RuntimeError("Unknown format %s" % self.fmt)
+
+        return blob.permute((1, 0, 2, 3, 4)).to(
+            self.dtype
+        )  # [nlayers, 2, ntokens, num_heads, head_size]
+        # huggingface
```
DongDongJu, line 208: good to mention that cachegen currently only support vllm at this point in here.

lmcache/storage_backend/serde/cachegen_encoder.py
```diff
@@ -384,8 +383,8 @@ def to_bytes(self, tensor: torch.Tensor) -> bytes:
 
         # TODO: permute is expensive here, need a better way to do it at lower
         # level
-        if self.fmt == "huggingface":
-            tensor = tensor.permute(0, 1, 3, 2, 4)
+        # huggingface:
```
DongDongJu, line 386: Same. not sure even cachegen use case for huggingface is existing now.

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -101,6 +101,8 @@ def Deserialize(data: bytes) -> "CudaIPCWrapper":
         return pickle.loads(data)
 
 
+# TODO: consider adding local_world_size and local_worker_id
+# for multi-node use cases
```
DongDongJu, line 105: 👍

lmcache/v1/storage_backend/naive_serde/cachegen_decoder.py
```diff
@@ -114,14 +113,11 @@ def deserialize(self, buffer_memory_obj: BytesBufferMemoryObj) -> MemoryObj:
...
-                kv_chunk = blob.reshape(*blob.shape[:-2], hidden_dim).to(
-                    self.dtype
-                )  # [nlayers, 2, ntokens, num_heads, head_size]
-            case _:
-                raise RuntimeError("Unknown format %s" % self.fmt)
+
+        hidden_dim = blob.shape[-1] * blob.shape[-2]
+        kv_chunk = blob.reshape(*blob.shape[:-2], hidden_dim).to(
+            self.dtype
+        )  # [nlayers, 2, ntokens, num_heads, head_size]
```
DongDongJu, line 120: even here huggingface fmt was not considered :).

lmcache/v1/metadata.py
```diff
@@ -0,0 +1,114 @@
...
+    LMCacheMetadata should be extracted from the northbound
+    serving engine configuration and wrap the extraction of
+    attributes (e.g. model name, tp rank, etc.)
+    """
+
+    """name of the LLM model"""
+    model_name: str
+    """ global world size when running under a distributed setting 
+    (total number of workers)"""
+    world_size: int
```
DongDongJu, line 28: num_of_total_workers?

lmcache/v1/metadata.py
```diff
@@ -0,0 +1,114 @@
...
+    """ host world size (workers on active localhost)
+    This information can be useful for multi-node
+    deployment. Will be the same as world_size 
+    in single-node deployments.
+    """
+    local_world_size: int
```
DongDongJu, line 34: num_of_local_workers?

DongDongJu (approved):
Generally LGTM.
Its same to me that worker_id and world_size logic is unclear for multi gpu/node cases.
we can add up after this one. 
its already big enough.

Last question is that now we can sharing the kvcache when model_name and other meta data is same even not properly working now.
Do we have any exploration result about this? for example, try to check whether vllm produced kvcache can work with sglang?

nit) 
var name `world_size` and worker didn't match that much to me.
I suggested the new variable names.

sammshen:
@DongDongJu thanks for your review! 

Agreed we can further discuss the multi node world sizes and worker ids when building the `MetadataBuilder`s in (3/N). The reason this PR did not add this logic is because the `LMCacheManager` is constructing `LMCacheMetadata` and this is not desired (this will be removed). Once the Metadata construction logic is in one single place, this will be the PR to introduce new logic. 

I think adding an example for cross engine sharing would be really great and I added a follow up to the roadmap for (1/N)

hickeyma:
Sorry that I am late responding. 

This looks good, thanks @sammshen. Some small nits as follows:

- Does `world_size` mean `num_workers`?
- Do you mind updating the description to show what changed in the PR?  Things like `LMCacheEngineConfig` changes and `use_case` seems out of date?

sammshen:
@hickeyma yep world_size is TP x PP (num_workers potentially across nodes)

hickeyma:
> @hickeyma yep world_size is TP x PP (num_workers potentially across nodes)

@sammshen Would it make more sense to call it `num_workers` then? `world_size` doesn't describe what the value means.

PR #2476, opened 2026-01-23
Http Server

Oasis-Git:
Adds an experimental option for the HTTP server to start and manage the MP (ZMQ) server internally. The run_cache_server() now supports a return_engine parameter that returns the server and engine without blocking. This enables a unified entry point (python -m        
  lmcache.v1.multiprocess.http_server) for testing HTTP API and ZMQ cache operations together. 

**If applicable**:


lmcache/v1/multiprocess/http_server.py
```diff
@@ -0,0 +1,784 @@
...
+logger = init_logger(__name__)
+
+app = FastAPI(title="LMCache HTTP API", version="1.0.0")
+
+# ----------------------------
+# Global server lifecycle state
+# ----------------------------
+_server_instance: Optional[uvicorn.Server] = None
+_server_lock = threading.Lock()
+_zmq_server = None  # ZMQ MessageQueueServer reference for cleanup
```
ApostaC: We can use FastAPI lifespan to manage the global variables' initialization and closure.
See https://fastapi.tiangolo.com/advanced/events/#lifespan

ApostaC: Example:
```python3
from fastapi import FastAPI, Request
from contextlib import asynccontextmanager

# Define a type hint for the state
# This helps with type checking, as app.state is a generic namespace
class AppState:
    config: dict

# Use asynccontextmanager to define the lifespan
@asynccontextmanager
async def lifespan(app: FastAPI):
    # Load configuration during startup
    app.state.config = {"database_url": "sqlite:///./test.db", "api_key": "some_key"}
    print("Application startup: Configuration loaded")
    yield
    # Clean up resources during shutdown (optional)
    print("Application shutdown: Cleaning up resources")

# Pass the lifespan context manager to the FastAPI app
app = FastAPI(lifespan=lifespan)

@app.get("/config")
def get_config(request: Request):
    # Access the state in a path operation
    config = request.app.state.config
    return {"database_url": config["database_url"]}
```

Cursor should be able to do this

lmcache/v1/multiprocess/http_server.py
```diff
@@ -0,0 +1,784 @@
...
+        logger.error(
+            "Error retrieving KV cache metadata for hash %s: %s",
+            hash_str,
+            e,
+            exc_info=True,
+        )
+        raise HTTPException(status_code=500, detail="Internal server error") from e
+
+
+@app.post("/api/kv-cache/download")
```
ApostaC: We have the endpoints both with `/api/xxx` and without the `api`. Should we consider unifying them?

ApostaC (changes requested):
Some minor comment. Otherwise LGTM! Will be definitely helpful for debugging!

lmcache/v1/multiprocess/http_server.py
KuntaiDu: Would be great if we can rename the API name to something like
/api/v1/xxx

lmcache/v1/multiprocess/http_server.py
```diff
@@ -0,0 +1,764 @@
...
+    app.state.zmq_server = zmq_server
+    app.state.engine = engine
+    logger.info("LMCache HTTP server initialized")
+
+    yield
+
+    # Shutdown
+    logger.info("Shutting down LMCache HTTP server...")
+    if hasattr(app.state, "zmq_server") and app.state.zmq_server is not None:
+        app.state.zmq_server.close()
```
ApostaC, line 76: Add `app.state.engine.close()`

ApostaC (approved):
LGTM!

KuntaiDu (approved):
LGTM

PR #2478, opened 2026-01-23
Refactor: sequential key polling for distributed key synchronization

hlin99:
Modified the cache lookup logic to handle synchronization delays in distributed environments. By introducing a polling mechanism, slow devices or lagging storage nodes are given a grace period to synchronize missing keys.

Key changes:
- Introduced `lookup_poll_key_timeout_ms` and `lookup_poll_key_intervals_ms` to configure retry behavior (defaults to 0, maintaining legacy one-shot behavior).
- Implemented a nested polling loop that ensures all layers of a specific chunk are synchronized (`hit_chunks == self.num_layers`) before advancing.
- Integrated `time.monotonic()` for robust timeout tracking against distributed clock shifts.

This ensures data consistency and prefix-matching integrity in scenarios where storage layers exhibit eventual consistency.


hlin99:
hi @maobaolong  @sammshen  would you like to have a review and comment is this change can help increase the robustness and scalability of remote backends?  I found such issues on mooncake backends where the receiver may have some latency to successfully poll the key. that's why adding polling timeout may help. Many thanks!

maobaolong (commented):
@hlin99  Would you like to add configuration introduce into the `docs/source/api_reference/configurations.rst` also?

hlin99:
> @hlin99 Would you like to add configuration introduce into the `docs/source/api_reference/configurations.rst` also?

Sure. Will update. Thanks for the comments!

hlin99:
> > @hlin99 Would you like to add configuration introduce into the `docs/source/api_reference/configurations.rst` also?
> 
> Sure. Will update. Thanks for the comments!

doc has been updated. please help review if anything else is needed. thanks. @maobaolong

sammshen:
do we want this or do we want lookup to fail fast?

hlin99:
> do we want this or do we want lookup to fail fast?

hi @sammshen, in some cases of remote backends + proxy server as PD system, the proxy server is not aware of when prefill put kv is done, then to notify decode.  It relies on decode to poll kv from the remote. The latency is not high however current implementation w/o poll timeout can fail the get kv call. I encountered such an issue on mooncake + vLLM proxy server. And this PR is to offer a mechanism in LMCache to overcome the issue if others encounter the same.

lmcache/v1/cache_engine.py
```diff
@@ -1119,22 +1145,46 @@ def lookup(
...
+                        elapsed_ms = (time.monotonic() - start_time) * 1000
+                        if elapsed_ms > self.lookup_poll_key_timeout_ms:
+                            logger.warning(
+                                f"Lookup poll timeout after {elapsed_ms:.2f}ms"
+                            )
+                            break
+                    else:
+                        break
+
+                    time.sleep(self.lookup_poll_key_intervals_ms / 1000)
```
sammshen, line 1187: lookup is synchronous so this could delay the scheduler

sammshen:
can you show the concrete example sorry I don't fully understand

hlin99:
> can you show the concrete example sorry I don't fully understand

hi @sammshen  whether the issue exists depend on how the PD proxy server is implemented. a diagram is shown as below. if the proxy server gets notified at the same time of store KV, get KV from decode may happen earlier than KV is ready in the backend. hopefully, this clarifies the issue.

[
截屏2026-02-25 20 18 21
](url)

hlin99:
> > can you show the concrete example sorry I don't fully understand
> 
> hi @sammshen whether the issue exists depend on how the PD proxy server is implemented. a diagram is shown as below. if the proxy server gets notified at the same time of store KV, get KV from decode may happen earlier than KV is ready in the backend. hopefully, this clarifies the issue.
> 
> [ 截屏2026-02-25 20 18 21 ](url)

the issue is not caused by LMCache. and it seems the proxy server in LMCache example is notified after store KV, so the issue won't be there, but proxy server can be implemented in diff ways... so this PR offers a configuration to fit any implementation by tuning configs.

sammshen (changes requested):
I'm hesitant to approve this PR due to the while loop on the critical path `lookup`. I see the motivation but I don't see any issues directly otivating the need for this change. @maobaolong WDYT?

PR #2479, opened 2026-01-23
Mark unhealthy while init or post_init failed

maobaolong:

This PR improves the robustness of the LMCache system by introducing explicit error handling during its initialization and post-initialization phases. It ensures that if any critical component fails to initialize, the system is correctly marked as unhealthy, preventing further operations and allowing it to gracefully fall back to a degraded mode (recomputation). This change enhances system reliability and provides clearer health status reporting.

**If applicable**:

### DIFF
--- a/tests/v1/test_manager.py
+++ b/tests/v1/test_manager.py
@@ -477,3 +477,192 @@ class TestLMCacheManagerCalculateDraftLayers:
             )
 
         assert manager._calculate_draft_layers() == 3
+
+
+class TestLMCacheManagerInitFailure:
+    """Tests for LMCacheManager initialization failure handling."""
+
+    def test_init_components_exception_makes_unhealthy(self):
+        """Test that exception during _init_components makes manager unhealthy."""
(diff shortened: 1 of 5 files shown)


chunxiaozheng (approved):
lgtm

maobaolong:
@sammshen Would you like to take another look? Thanks!

tests/v1/test_manager.py
ApostaC, line 1: High-level comment: it seems like we are using a lot of fields that starts with `_` in the unit tests. 
IIUC, those fields are supposed to be "private" in the class implementation, which means they might be changed in the future during code refactoring. Therefore, it would be a good practice to avoid using those fields in the unit tests.

I'm okay with not changing the current PR if it's too complicated. But let's keep this in mind in the future, WDYT?

ApostaC (approved):
LGTM!

tests/v1/test_manager.py
maobaolong: Thanks for your suggestion, I will reduce access the private method/field.

Issue #2394, opened 2026-01-12
[RFC] Rust based raw block device support


DongDongJu:
## Overview

The Rust Raw Block Backend is a new high-performance storage plugin for LMCache 
that enables direct NVMe device I/O for KV cache storage. 
Built with Rust for maximum performance, it bypasses the OS page cache using O_DIRECT 
and provides persistent cache storage across server restarts.

1. Simple read/write support for raw rust backend : https://github.com/LMCache/LMCache/pull/2482
2. zero-copy btw rust and python side for decreasing memcpy : https://github.com/LMCache/LMCache/pull/2573
3. block based meta data support : https://github.com/LMCache/LMCache/pull/2614
4. add io_worker
5. async mode support.
6. posix io_uring support @ankit-sam https://github.com/LMCache/LMCache/pull/2635
7. io_uring raw cmd support @ankit-sam 
8.  mp mode support. @djkimpp
9. scheduler support for higher TP.

I made draft MP suppport based on what we have and based on https://github.com/LMCache/LMCache/pull/2667 changes.
I made the comparison two case ( 6TP, 6 vllm instance with mp mode with rustrawblockbackend vs localdiskbackend )

testing model used : bigcode/starcoder2-3b

  - TP6 populate: 17.631 vs 11.056 req/s (1.595x), mean TTFT 821.8 vs 1263.8 ms
  - TP6 retrieve: 18.900 vs 16.109 req/s (1.173x), mean TTFT 775.8 vs 901.6 ms
  - 6xTP1 populate (agg): 370.662 vs 355.981 req/s (1.041x), mean TTFT 270.1 vs 371.9 ms
  - 6xTP1 retrieve (agg): 376.988 vs 141.636 req/s (2.662x), mean TTFT 274.6 vs 644.0 ms

Feel free to test [my local branch](https://github.com/DongDongJu/LMCache/tree/feature/mp-rustraw-combined-meta-io-20260302)
its chimera branch of https://github.com/LMCache/LMCache/pull/2614 + add io_worker feature + https://github.com/LMCache/LMCache/pull/2667 and mp mode support for rawblockbackend.

DongDongJu:
Hello Martin, I do understand that your concern.
But, if we are managing detached one with current status, I bliv it's really easy to break up the integration.
[...]

PR #2482, opened 2026-01-25
[1/N][Feat] Simplified rust raw block backend

DongDongJu:

Adds a new Rust-based raw block device storage backend for LMCache with improved error handling.
I removed all optimization for the first PR.

**Changes**:
- its series for https://github.com/LMCache/LMCache/issues/2394.
- Add `RustRawBlockBackend` storage plugin using Rust extension for high-performance I/O
- Implement error handling with proper logging for I/O operations
- Add manifest persistence for crash recovery
- Support O_DIRECT for bypassing page cache

### `BLKGETSIZE64`
- Linux `ioctl` command used to query the size of a block device in bytes.
- We use it when the opened file descriptor is an actual block device (e.g., `/dev/nvmeXnY`).

### `PYBUF_*` flags
These are CPython C-API buffer flags.
- `PYBUF_WRITABLE`: requires the buffer to be writable.
- `PYBUF_ND`: requests an N-dimensional buffer (standard flag for structured buffers).
- `PYBUF_STRIDES`: requests stride info; used by CPython to validate contiguity.
- `PYBUF_ANY_CONTIGUOUS`: accepts any contiguous buffer (C or Fortran order). This is important because raw pointer I/O assumes contiguous memory.

### `round_up(x, align)`
- **Purpose**: Align a byte count up to the next multiple of `align`.
- **Why**: O_DIRECT requires I/O sizes to be aligned to the device block size (commonly 4096 bytes).
- **Example**: `round_up(4100, 4096)` → `8192`.

### `errno()`
- **Purpose**: Reads the thread-local `errno` after a failed libc syscall.
- **Why**: We convert this into Python `OSError` so callers get a meaningful error code.

### `os_err(msg)`
- **Purpose**: Creates a Python `OSError` from the current `errno` and a custom message.
- **Why**: Python-side error handling should see real OS error codes.

### `pwrite_from_ptr(fd, offset, ptr, len)`
- **Purpose**: Low-level loop that writes exactly `len` bytes to `fd` at `offset`.
- **How it works**:
  - Calls `libc::pwrite` repeatedly until all bytes are written.
  - Advances pointer and offset after each partial write.
- **Why**: `pwrite` can return partial writes, especially under load or with O_DIRECT.

### `pread_into(fd, offset, dst, size)`
- **Purpose**: Low-level loop that reads exactly `size` bytes into `dst`.
- **How it works**:
  - Calls `libc::pread` repeatedly until all bytes are read.
  - Treats EOF as an error (unexpected for block device reads).
- **Why**: Matches the semantics expected by the Python caller (all-or-nothing read).

### `fd_size_bytes(fd)`
- **Purpose**: Determine the size of the underlying file/device in bytes.
- **How it works**:
  - First tries `ioctl(fd, BLKGETSIZE64)` for block devices.
  - Falls back to `fstat` for regular files.
- **Why**: Tests use a regular file, but production uses block devices.

### Struct Overview
- Holds a pointer to aligned memory allocated with `posix_memalign`.
- Needed for **O_DIRECT**, which requires aligned buffers.

### `AlignedBuf::new(len, align)`
- Allocates an aligned buffer of size `len` with alignment `align`.
- Uses `posix_memalign`; returns error if allocation fails.

### `as_mut_ptr()` / `as_ptr()`
- Provide raw pointers to the allocated buffer for `pread`/`pwrite`.

### `Drop` implementation
- Frees the buffer with `libc::free` when the struct goes out of scope.

### `get_pybuffer(py, obj, writable)`
- **Purpose**: Acquire a raw buffer view from a Python object.
- **Why**: We need raw memory access to read/write bytes directly.
- **Behavior**:
  - If `writable=true`, the buffer must be mutable.
  - Enforces contiguity via `PYBUF_ANY_CONTIGUOUS`.

### `release_pybuffer(view)`
- Releases the buffer view obtained by `get_pybuffer`.
- Required to avoid Python-side resource leaks.

### Overview
Represents a single file descriptor opened on a block device (or file). Exposed to Python.

Fields:
- `fd`: raw OS file descriptor
- `size`: cached device size
- `closed`: prevents double-close
- `use_odirect`: whether O_DIRECT is enabled
- `alignment`: required alignment for offsets and I/O sizes

### `new(path, writable, use_odirect=false, alignment=4096)`
- Opens the path with read/write or read-only flags.
- Adds `O_DIRECT` if requested.
- Queries device size using `fd_size_bytes`.

### `size_bytes()`
- Returns the cached device size in bytes.

### `pwrite_from_buffer(offset, data, payload_len=None, total_len=None)`
- **Purpose**: Write data from a Python buffer into the device.
- **Key steps**:
  1. Acquire a Python buffer view.
  2. Validate lengths.
  3. If O_DIRECT: enforce alignment of offset/size.
  4. If `total_len > payload_len` or O_DIRECT: use a bounce buffer.
  5. Write using `pwrite_from_ptr`.
- **Why bounce buffer?** Python buffers are not guaranteed to be aligned.

### `pread_into(offset, out, payload_len, total_len=None)`
- **Purpose**: Read data from the device into a Python buffer.
- **Key steps**:
  1. Acquire a writable Python buffer.
  2. Validate size and alignment.
  3. If O_DIRECT or padded read: use bounce buffer.
  4. Read using `pread_into` and copy into Python buffer.

### `close()`
- Explicitly closes the file descriptor.
- Safe to call multiple times (no double-close).

### `Drop` for `RawBlockDevice`
- Ensures file descriptor is closed if Python forgets to call `close()`.

### Summary 
This crate is a thin Rust wrapper for Linux block I/O. 
It does **not** handle caching logic or metadata (next TODO with async support).
those are implemented in Python. 
The main complexity here is making raw block I/O safe and compatible with Python buffers, especially under O_DIRECT alignment constraints.

**If applicable**:


DongDongJu:
@sammshen @ApostaC I resolved the ci issue. PTAL

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -0,0 +1,727 @@
...
+            return False
+
+    def unpin(self, key: CacheEngineKey) -> bool:
+        with self._lock:
+            if key in self._pinned:
+                self._pinned.remove(key)
+                return True
+            return key in self._index
+
+    def remove(self, key: CacheEngineKey, force: bool = True) -> bool:
```
sammshen, line 315: remove() currently reclaims slots for keys in _inflight, but the async write may still be running and will write to offset even after the slot is reallocated to another key. This can corrupt data. Consider not returning inflight slots to _free_slots until the write finishes (e.g., block remove, or mark “pending free” and free in completion, or add slot generation and prevent reuse until completion).

DongDongJu: Hello @sammshen, Thanks for the catch.
fixed by https://github.com/LMCache/LMCache/pull/2482/changes/b47eabb46d72007198830530c517a3398db60b62.

sammshen:
CQ failing now: 
```text
@@ -453,9 +453,7 @@ class RustRawBlockBackend(StoragePluginInterface):
                 inflight = self._inflight.pop(key, None)
                 if inflight is not None:
                     if inflight.canceled:
-                        self._free_slots.append(
-                            int(inflight.offset // self.slot_bytes)
-                        )
+                        self._free_slots.append(int(inflight.offset // self.slot_bytes))
                     else:
                         self._index[key] = _Entry(
                             offset=inflight.offset,
```

DongDongJu:
@sammshen fix it!

DongDongJu:
@sammshen @ApostaC fixed matching with new metadata. PTAL

DongDongJu:
@sammshen @ankit-sam I updated pr description and code comment too. PTAL

rust/raw_block/README.md
```diff
@@ -0,0 +1,34 @@
...
+- Batch read/write operations (preadv/pwritev)
+- Optional asynchronous I/O via multi-process mode
+
+## Usage
+
+```python
+from lmcache_rust_raw_block_io import RawBlockDevice
+
+# Open device
+dev = RawBlockDevice("/dev/nvme0n1", use_odirect=True)
```
sammshen: now needs: 
```python
RawBlockDevice(path, writable, use_odirect=False, alignment=4096)
```

rust/raw_block/README.md
```diff
@@ -0,0 +1,34 @@
...
+maturin develop --release
+```
+
+## Features
+
+- Direct block device access with O_DIRECT support
+- Batch read/write operations (preadv/pwritev)
+- Optional asynchronous I/O via multi-process mode
+
+## Usage
```
sammshen, line 19: rec for these docs.

I like that they are very minimal to start with. 

can we leave/update: 
1. The code only uses synchronous pread / pwrite (no preadv/pwritev).
2. There is no async I/O; py.allow_threads releases the GIL but still blocks the OS thread.

rust/raw_block/src/lib.rs
```diff
@@ -0,0 +1,444 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Raw block device I/O extension for LMCache.
+//! Provides direct block device access with optional O_DIRECT support.
+//!
+//! Design notes (for reviewers unfamiliar with Rust / Linux I/O):
+//! - This module exposes a very small surface to Python via PyO3.
+//! - We wrap Linux `pread` / `pwrite` on a file descriptor opened from a
+//!   block device (e.g., /dev/nvmeXnY) or a regular file (for tests).
+//! - When O_DIRECT is enabled, Linux requires aligned offsets and I/O sizes.
```
sammshen, line 10: then IIUC, `O_DIRECT` will not be that effective right? 

can we potentially move the memory allocation to the Rust side (and expose to python) to guarantee alignment and avoid bounce buffer? 

expose something like:

```python
AlignedBuffer(size, alignment=4096) from Rust

It stores ptr/len/alignment

It implements the Python buffer protocol so Python can do:

buf = AlignedBuffer(4096, alignment=4096)
dev.pwrite_from_buffer(offset=0, data=buf, total_len=4096)
```

the problem is we also need to cudaHostRegister all of the LMCache CPU Memory

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -0,0 +1,736 @@
...
+                    "TP > 1 support will be added in a future release.\n"
+                    "For now, please use TP=1 or choose a different storage backend."
+                )
+
+        extra = self.config.extra_config or {}
+        self.device_path: str = extra.get("rust_raw_block.device_path", "")
+        if not self.device_path:
+            raise ValueError("extra_config['rust_raw_block.device_path'] is required")
+
+        self.manifest_path: Optional[str] = extra.get("rust_raw_block.manifest_path")
```
sammshen, line 117: what do you think about separating manifest related methods into a separate class?

sammshen (changes requested):
LGTM! left some comments. 

Can we get some initial documentation for planned future optimizations? I think it might be great to track the perf improvements over time. Could we benchmark this PR as an initial starting point? And then we can update the improvements over time. 

If my understanding is correct, the main benefit of using Rust (or running away from python in general) is multi-threading. If this is the case, then we should not just provide a BlockDevice handle but should actually provide an interface that hides the multi-threading underneath in the Rust? 

Looking forward to unblocking this soon!

rust/raw_block/README.md
DongDongJu: Yes I want to make the incremental PR for performance too. Thanks.

rust/raw_block/src/lib.rs
DongDongJu: Agreed, that O_DIRECT is not fully effective with bounce. 
It still avoids page cache but does not eliminate copies. 
An AlignedBuffer that implements the Python buffer protocol would help, but LMCache would need allocator changes (and likely pinned memory registration) to actually use it.
We can track this as follow‑up.

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
DongDongJu: Make sense. 
A small manifest helper class will simplify backend logic and make the upcoming Rust-side metadata move easier. 
I can keep it minimal in this PR (load/save/serialize only) and avoid touching I/O/eviction logic.
The next PR will include this with moving meta saving into block device too.

rust/raw_block/README.md
DongDongJu: Added in https://github.com/LMCache/LMCache/pull/2482/changes/f73631e3f8bbc5732f1b816ad0e0b20329ecbdc7

rust/raw_block/README.md
DongDongJu: Added in here https://github.com/LMCache/LMCache/pull/2482/changes/fc61b85977bfa914031feffeea9172b35a231f07

DongDongJu:
> LGTM! left some comments.
> 
> Can we get some initial documentation for planned future optimizations? I think it might be great to track the perf improvements over time. Could we benchmark this PR as an initial starting point? And then we can update the improvements over time.
> 
> If my understanding is correct, the main benefit of using Rust (or running away from python in general) is multi-threading. If this is the case, then we should not just provide a BlockDevice handle but should actually provide an interface that hides the multi-threading underneath in the Rust?
> 
> Looking forward to unblocking this soon!

Hello @sammshen, Thanks for the review.

Sure. I added a baseline perf section and keep a small benchmark record in benchmark folder so we can track improvements over time.
You can run like below,
```
 python benchmarks/storage_backend_io/storage_backend_io_benchmark.py     --num-ops 512 --concurrency 32 --backend both     --local-disk-dir /tmp/lmcache_local_disk_bench     --max-local-disk-gb 2     --output-json /tmp/storage_backend_io_same_fs.json
```

Also agree the real win is hiding threading inside Rust; current PR only exposes raw pread/pwrite (GIL released but still blocks the OS thread).
A follow-up PRs can introduce a Rust-side worker pool + async queue interface to unlock true parallel I/O.)
I updated issue for this series too (https://github.com/LMCache/LMCache/issues/2394).

sammshen (approved):
LGTM! Thank you for the really great work on this series!!

rust/raw_block/README.md
ApostaC, line 1: Do we want to put the wheel building process into LMCache CI/CD and release the pypi?

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
ApostaC, line 1: IIUC, the current backend uses multiple Python threads for non-blocking writes, and a single thread (the main thread) for blocking reads. Is it correct?

In the future, do we want to spawn the threads in the rust backend?

ApostaC (approved):
Like this a lot!

There could be a lot of changes in the python level once we want to integrate it in MP mode, but really looking forward to it.

Giving a approve to this PR, but just with a few questions:

rust/raw_block/README.md
DongDongJu: I think after mp mode integration is good timing for this.
WDYT?

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
DongDongJu: Yes.
Yes will spawn multiple process with shared memory.

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
```diff
@@ -0,0 +1,366 @@
...
+        output_path = args.output_json
+        if output_path.endswith(os.sep) or os.path.isdir(output_path):
+            ts = time.strftime("%Y%m%d_%H%M%S")
+            output_path = os.path.join(output_path, f"storage_backend_io_{ts}.json")
+        with open(output_path, "w") as f:
+            json.dump(results, f, indent=2)
+        print(f"Wrote results to {output_path}")
+
+
+if __name__ == "__main__":
```
maobaolong, line 365: I guess you can reference and add the benchmark logic to the generic tool `lmcache/v1/check/check_mode_test_storage_manager.py`.

rust/raw_block/README.md
maobaolong: @ApostaC @DongDongJu Do you think we can create a project `lmcache_raw_block_backend` in the lmcache group?

rust/raw_block/README.md
```diff
@@ -0,0 +1,33 @@
+# LMCache Rust Raw Block I/O
+
+This crate provides raw block device I/O operations for LMCache using Rust and PyO3.
```
maobaolong, line 3: We'd better add a limitation, for example, we have to use raw black device?

rust/raw_block/README.md
```diff
@@ -0,0 +1,33 @@
...
+cd rust/raw_block
+pip install maturin
+maturin develop --release
+```
+
+## Features
+
+- Direct block device access with O_DIRECT support
+- Synchronous `pread` / `pwrite` only (no `preadv`/`pwritev`)
+- No async I/O; `py.allow_threads` releases the GIL but still blocks the OS thread
```
maobaolong, line 17: This is a good point. But it can also apply to other devices and other remote storages? WDYT?
E.g. FileSystem based storage, third-party remote storages

maobaolong (commented):
Thanks for invite me to take a look, this is a good startup which can inspire us to add more logic to the native code.

rust/raw_block/README.md
DongDongJu: Yes. I agree we need better abstraction to adapt these into others. Let me share the design for this soon.

rust/raw_block/README.md
DongDongJu: I will include this in the following one.

rust/raw_block/README.md
maobaolong: @DongDongJu Look forward to take a look how your native remote backend looks.

rust/raw_block/README.md
DongDongJu: Done!

PR #2487, opened 2026-01-26
[CI]: Add stress test pipeline with real-multi-round-qa benchmark

ningziwen:
⚠️ **DRAFT PR - Early feedback requested**

This PR depends on [LMBenchmark PR](https://github.com/LMCache/LMBenchmark/pull/36) being merged first.

---

Adds a stress test pipeline using the real-multi-round-qa benchmark from LMBenchmark. This enables continuous stress testing of LMCache with realistic multi-turn conversation workloads using 64k context novels. This tests should capture the issues related to memory leaking that require long running.

The stress test:
- Maintains 5 concurrent users with constant load (--num-users mode)
- Runs 5 rounds of Q&A per session
- Uses 64k context from Gutenberg novels
- Runs for 10 minutes (600 seconds)
- Collects TTFT, ITL, and throughput metrics

**Files added:**
- `.buildkite/pipelines/stress-tests.yml` - Buildkite pipeline definition
- `.buildkite/scripts/real-multi-round-qa.sh` - Script to run the benchmark

Requires LMBenchmark multiround branch to be merged to main first. The script clones from `https://github.com/ningziwen/LMBenchmark.git` branch `multiround`.

**If applicable**:

### DIFF
--- a/.buildkite/pipelines/stress-tests.yml
+++ b/.buildkite/pipelines/stress-tests.yml
@@ -1,0 +1,32 @@
+steps:
+  - label: ":books: Real Multi-Round QA"
+    key: "real-multi-round-qa"
+    command: |
+      uv venv --python 3.12 .venv-$BUILDKITE_BUILD_ID
+      source .venv-$BUILDKITE_BUILD_ID/bin/activate
+
+      uv pip install --upgrade pip setuptools wheel
+      uv pip install -r requirements/common.txt
+      uv pip install -e . --no-build-isolation
+      uv pip install -U vllm --pre --extra-index-url https://wheels.vllm.ai/nightly
+
+      source .buildkite/scripts/pick-free-gpu.sh 18000
+
+      bash .buildkite/scripts/real-multi-round-qa.sh
+    artifact_paths:
+      - "real-multi-round-qa-64k-results.json"
+      - "lmcache_vllm.log"
+
+  - label: ":wastebasket: Kill all"
+    key: "clean"
+    depends_on: "real-multi-round-qa"
+    allow_dependency_failure: true
+    command: bash .buildkite/scripts/bare-machine-cleanup.sh
+
+  - label: "Clean up"
+    depends_on: "clean"
+    command: |
+      export TARGET="$PWD"
+      echo "Deleting current workspace $TARGET"
+      cd /
+      sudo rm -rf "$TARGET"
--- a/.buildkite/scripts/real-multi-round-qa.sh
+++ b/.buildkite/scripts/real-multi-round-qa.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -ex
+
+MODEL="meta-llama/Llama-3.2-1B-Instruct"
+
+# Clone LMBenchmark repository
+git clone --depth 1 -b multiround https://github.com/ningziwen/LMBenchmark.git /tmp/LMBenchmark
+
+# Install requirements
+pip install -r /tmp/LMBenchmark/real-multi-round-qa/requirements.txt
+
+# Download novel data (64k)
+python3 /tmp/LMBenchmark/real-multi-round-qa/prepare.py \
+    --output /tmp/novels \
+    --model $MODEL \
+    --start 0 \
+    --end 100
+
+# Start the vLLM server with LMCache
+LMCACHE_TRACK_USAGE="false" lmcache_vllm serve $MODEL \
+    --disable-log-requests > lmcache_vllm.log 2>&1 &
+
+echo "Waiting for service to start..."
+timeout=120
+elapsed=0
+until grep -q "Uvicorn running on" lmcache_vllm.log; do
+    if [ $elapsed -ge $timeout ]; then
+        echo "Timeout: Service did not start within $timeout seconds."
+        cat lmcache_vllm.log
+        exit 1
+    fi
+    sleep 10
+    elapsed=$((elapsed + 10))
+    echo "Waiting... ($elapsed seconds elapsed)"
+done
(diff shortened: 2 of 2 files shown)


ningziwen:
@ApostaC @Shaoting-Feng Getting some early high level feedback for this draft and attention to https://github.com/LMCache/LMBenchmark/pull/36.

Another option is not depending on LMBenchmark and moving everything in LMCache repo.

ningziwen:
@ApostaC Follow up on why cannnot use existing shareGPT based multi-round QA. Looking forward to your thoughts.

The target context length I want to test is 20k tokens or even 64k tokens. The current ShareGPT-based multi-round-qa benchmark has two issues that may not match actual multi-round qa scenarios.

### Issue 1: Padding Creates Shared Prefix

The current benchmark uses `--shared-system-prompt` and `--user-history-prompt` to reach longer context lengths. Looking at the CI configuration, we use 1000 + 2000 = 3000 tokens of padding. The problem is this padding is generated as dummy text (`"hi hi hi..."`) that is nearly identical across all users - the only difference is the user ID number embedded in the text.

When all users share the same prefix, the KV cache can trivially deduplicate and reuse that prefix across all sessions. This makes the benchmark artificially easy and not representative of real-world workloads where each user brings their own unique document or context. We are essentially testing prefix caching efficiency rather than KV cache memory pressure under diverse workloads.

### Issue 2: ShareGPT Turns Are Short

ShareGPT contains 53k+ conversations, but each individual conversation turn is typically only 100-500 tokens. The benchmark sends one turn at a time and accumulates history across rounds. With 5 rounds (as in CI), the total context is roughly:

- Turn 1: 3000 (padding) + ~200 (ShareGPT) = ~3200 tokens
- Turn 5: ~4000-5000 tokens total

To reach 20k or 64k tokens using pure ShareGPT content (without padding), we would need 40-100+ rounds. This is impractical and still doesn't solve the shared prefix problem if we keep any padding.

### Issue 3: ShareGPT Lacks Long Conversations

Even if we wanted to increase rounds to accumulate more context, we hit another limitation. The benchmark filters ShareGPT data to only include conversations with enough turns:

```python
self.sharegpt_data = [
    d for d in self.sharegpt_data
    if d["num_round"] > 2 * self.workload_config.num_rounds
]
```

For N rounds, we need conversations with > 2N turns. Testing showed:
- 5 rounds: need >10 turns (many conversations qualify)
- 15 rounds: need >30 turns (102 conversations qualify based on benchmark log)
- 35 rounds: need >70 turns (benchmark failed with `IndexError` - too few conversations)

This means we simply cannot use more rounds to reach longer context - ShareGPT doesn't have enough long conversations. The dataset is fundamentally limited by how humans interact with ChatGPT (typically short back-and-forth exchanges, not 70+ turn marathons).

Even if we set padding to zero and rely purely on ShareGPT, each user does get assigned a different conversation from the dataset, so the content would be unique. But the context length would still be limited by ShareGPT's natural conversation lengths, and we cannot increase rounds beyond what the dataset supports.

### The Practical Limit

I attempted to run the benchmark with 35 rounds and zero padding to reach 7000 tokens through pure ShareGPT content. The job failed immediately with `IndexError: list index out of range` because only 5 conversations in the entire ShareGPT dataset have more than 70 turns.

The practical maximum is around 15-20 rounds. With 15 rounds and ~200 tokens per round from ShareGPT content, we can only reach about 3000 tokens of unique context. To reach the target of 7000 tokens, we would need to add ~4000 tokens of padding - but that padding would be shared across all users, defeating the purpose of testing unique context workloads.

## How real-multi-round-qa Solves This

The real-multi-round-qa benchmark from LMBenchmark takes a fundamentally different approach. It uses actual long documents (novels from Project Gutenberg) that are pre-categorized by token count: 8k, 16k, 32k, 64k, 128k, 256k, 512k, and 1M tokens.

The key difference is in how the first turn works:

```python
FIRST_PROMPT = "Read and summarize this novel.\n\n{}"

def _load_random_file(self):
    files = [f for f in os.listdir(self.src_dir) if os.path.isfile(os.path.join(self.src_dir, f))]
    with open(os.path.join(self.src_dir, random.choice(files)), encoding='utf-8') as f:
        return FIRST_PROMPT.format(f.read())
```

Each user session picks a random novel from the source directory. User A might get "Pride and Prejudice", User B gets "Moby Dick", User C gets "War and Peace" - completely different documents with no shared prefix whatsoever.

## Summary of Differences

| Aspect | ShareGPT + Padding | real-multi-round-qa |
|--------|-------------------|---------------------|
| Context source | Dummy padding + short conversation turns | Full novels (8k-1M tokens) |
| Prefix uniqueness | Nearly identical across users | Completely unique per user |
| Context length control | Requires many rounds or shared padding | Direct control via `--src-dir` (8k, 16k, 32k, etc.) |
| KV cache behavior | Tests prefix caching (shared prefix) | Tests true memory pressure (unique contexts) |
| Realism | Synthetic padding | Real document Q&A workload |

## Benefits of Adding real-multi-round-qa

1. Tests the scenario we actually care about: multiple concurrent users each with their own large, unique context competing for KV cache space.

2. Provides direct control over context length without needing to run dozens of rounds or add artificial padding.

3. Represents realistic long-context workloads like document Q&A, where users upload their own documents rather than sharing a common system prompt.

4. Complements the existing ShareGPT benchmark - ShareGPT tests prefix caching efficiency (which is valuable), while real-multi-round-qa tests KV cache capacity under diverse workloads.

The existing ShareGPT benchmark is good for what it tests, but I think it cannot simulate the memory and compute pressure pattern of unique long-context multi-round workloads. That's what real-multi-round-qa provides.

PR #2497, opened 2026-01-27
[1/N][Northbound] Introduce KVCacheFormat

DongDongJu:
- Adds a KVCacheFormat descriptor contract in LMCache core for integration mode.
- Validates `kv_format` against legacy metadata early to fail on mismatches.
- Makes GPU connectors honor declared layout when present, without engine-name branching.
- Wires vLLM and SGLang integrations to declare the format, while keeping cache keys unchanged in this PR.

- Scope is intentionally limited to integration mode (no MP protocol changes here).
- `kv_format` is not added to cache keys in this PR.
- manual A/B correctness runs were done for vLLM and SGLang with `chunk_size=8`.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -18,6 +18,7 @@ import torch
 from lmcache.logging import init_logger
 from lmcache.v1.config import LMCacheEngineConfig
 from lmcache.v1.config_base import apply_remote_configs, fetch_remote_config
+from lmcache.v1.kvcache_format import build_dense_format_single_group
 
 logger = init_logger(__name__)
 ENGINE_NAME = "vllm-instance"
@@ -203,6 +204,17 @@ def create_lmcache_metadata(
     num_kv_head = model_cfg.get_num_kv_heads(parallel_cfg)
     head_size = model_cfg.get_head_size()
     kv_shape = (num_layer, 1 if use_mla else 2, chunk_size, num_kv_head, head_size)
+    hidden_dim = head_size if use_mla else num_kv_head * head_size
+
+    kv_format = build_dense_format_single_group(
+        num_layers=num_layer,
+        dtype=kv_dtype,
+        hidden_dim=hidden_dim,
+        block_size=chunk_size,
+        use_mla=use_mla,
+        separation="packed",
+        format_version="v1",
+    )
 
     # Extract engine_id and kv_connector_extra_config from vllm_config if available
     engine_id = None
@@ -229,6 +241,7 @@ def create_lmcache_metadata(
         served_model_name=model_cfg.served_model_name,
         engine_id=engine_id,
         kv_connector_extra_config=kv_connector_extra_config,
+        kv_format=kv_format,
     )
 
     return metadata, config
--- a/lmcache/v1/gpu_connector.py
+++ b/lmcache/v1/gpu_connector.py
@@ -78,6 +78,16 @@ def lmcache_memcpy_async_d2h(
         memory_obj.tensor.copy_(gpu_buffer, non_blocking=True)
 
 
+def _kv_layout_from_metadata(metadata: LMCacheMetadata) -> dict:
+    """Extract a uniform layout description, honoring kv_format when present.
+
+    Returns dict with keys: block_size, hidden_dim, num_layers, use_mla, dtype.
+    Raises ValueError on inconsistencies between kv_format and legacy fields.
+    """
+    layout, _ = metadata.get_kv_layout("for connector use")
+    return layout
+
+
 class GPUConnectorInterface(metaclass=abc.ABCMeta):
     @abc.abstractmethod
     def to_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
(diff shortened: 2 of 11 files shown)


lmcache/v1/gpu_connector.py
```diff
@@ -78,6 +79,87 @@ def lmcache_memcpy_async_d2h(
...
+        )
+
+    return {
+        "block_size": block_size,
+        "hidden_dim": hidden_dim,
+        "num_layers": total_layers,
+        "use_mla": expected_use_mla,
+        "dtype": metadata.kv_dtype,
+    }
+
```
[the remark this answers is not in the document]
DongDongJu: resolved by https://github.com/LMCache/LMCache/pull/2497/commits/1a806c99324eeee4f91b93c4785610f32b04d44d.

lmcache/integration/vllm/utils.py
```diff
@@ -230,6 +242,7 @@ def create_lmcache_metadata(
         engine_id=engine_id,
         num_ranks=num_ranks,
         kv_connector_extra_config=kv_connector_extra_config,
+        kv_format=kv_format,
```
sammshen, line 244: this is not the only place where metadata is constructed

sammshen: the manager also has two places where it creates `LMCacheEngineMetadata`

lmcache/v1/gpu_connector.py
```diff
@@ -78,6 +79,18 @@ def lmcache_memcpy_async_d2h(
         memory_obj.tensor.copy_(gpu_buffer, non_blocking=True)
 
 
+def _kv_layout_from_metadata(metadata: LMCacheEngineMetadata) -> dict:
```
sammshen: ideally this is inside of the LMCacheMetadata class

lmcache/v1/kvcache_format.py
```diff
@@ -0,0 +1,282 @@
...
+
+# Third Party
+import msgspec
+
+# Public schema/version identifier for the format descriptor.
+KV_FORMAT_SCHEMA_VERSION = 1
+
+Family = Literal["MHA_DENSE", "MLA_LATENT", "SPARSE"]
+Canonical = Literal["KV_2LTD", "KV_MLA_FMT", "KV_SPARSE_FMT"]
+Separation = Literal["packed", "separated"]
```
sammshen, line 39: Separation is for layers?

sammshen: one tensor or multiple?

lmcache/v1/kvcache_format.py
```diff
@@ -0,0 +1,282 @@
...
+from typing import Literal, Optional, Protocol, Sequence
+import re
+
+# Third Party
+import msgspec
+
+# Public schema/version identifier for the format descriptor.
+KV_FORMAT_SCHEMA_VERSION = 1
+
+Family = Literal["MHA_DENSE", "MLA_LATENT", "SPARSE"]
```
sammshen: sparse is for k and v separated?

sammshen: these all need to be documented

lmcache/v1/kvcache_format.py
```diff
@@ -0,0 +1,282 @@
+# SPDX-License-Identifier: Apache-2.0
```
sammshen, line 1: @DongDongJu my general thoughts on this file are that we should stick having a simple Enum for the northbound KV Cache Layout instead of breaking down the layout's elements and having to account for a cartesian product. 

For example:
`vLLMPLTD` or `vLLMLPTD_SEP`

this is simpler to understand (maybe less elegant but much simpler) and allows us to route multiple enums to the same path.

My understanding based on conversation in https://github.com/LMCache/LMCache/pull/2472 is that:
1. either the northbound layout is only consumed by gpu connector on initialization
2. the northbound layout is consumed on a routing basis for each MemoryObjMetadata. 

For example, @Shaoting-Feng is building a new GPU connector for a different vLLM layout: https://github.com/LMCache/LMCache/pull/2498

Generally, for the non-multiprocess mode (and the multiprocess mode actually), we only need to consume the GPUConnector / transfer path once on initialization. Thus, I think we can add this to (2/N) of the northbound path in the GPU Connector factory initialization on the northbound refactor roadmap: https://github.com/LMCache/LMCache/issues/2458

lmcache/v1/kvcache_format.py
DongDongJu: Let me remove SPARSE for now.
I was intended NSA like architecture for this.
but lets add it later.

lmcache/v1/kvcache_format.py
DongDongJu: will documenting this.

lmcache/v1/kvcache_format.py
DongDongJu: Why @ApostaC and me discussed about this one before is we want to design it in agnostic way.
if we maintain the enum like `vLLMPLTD` then no point to make this.
what do you think?

lmcache/integration/vllm/utils.py
DongDongJu: I’ll add a focused unit test to ensure `kv_format` doesn’t get dropped when metadata is created via the manager.

lmcache/v1/gpu_connector.py
DongDongJu: IMO, Metadata is handshaking information btw engine and lmcache not kvcache format itself.
what do you think?

lmcache/v1/kvcache_format.py
DongDongJu: Added. PTAL!

lmcache/v1/kvcache_format.py
sammshen: agree the enum is too simple. 

currently, the `LMCacheMetadata` has `use_mla` and     `kv_layer_groups_manager` (assumption is that the layer groups of LMCache metadata are the same as northbound serving engine) already. We can add `kv_packed` and `layer_packed` as well. 

Since the primary consumer is GPU Connector initialization (correct me if I'm wrong and you actaully want this information to be in each `MemoryObjMetadata` eventually), let's do it in (2/N) on the roadmap: https://github.com/LMCache/LMCache/issues/2458

What do you think?

DongDongJu: Sounds great. let me know if other stuffs need for 1/N. I can dive in.

DongDongJu:
It already merged into little bit different form by @sammshen but functionally same.

PR #2505, opened 2026-01-28
[MP][2/N] Add native implementation for ttl lock for mp storage manager

ApostaC:

This PR adds a native implementation for "lock/pin" mechanism that can be used for the MP storage manager.  

Under extreme load tests (4 threads, 100K locks), it can achieve 1-2 us average latency, <20 us P99 latency, 1M+ operations per second. Will be more than enough for LMCache to use.

**If applicable**:


ApostaC:
cc @maobaolong This might also be useful for the current LMCache pin/unpin handler.

Oasis-Git (approved):
LGTM

sammshen (approved):
LGTM

Issue #2204, opened 2025-12-08
[Bug] Cross-server LMCache KV load causes AssertionError in vllm_v1_adapter when network is slow


hamid370:
## Summary

I'm using LMCache with two vLLM servers (each on a separate GPU machine), both connected to a single shared `lmcache_server` on another server.

When I send a long prompt (~6000 tokens) to **server A**, the KV cache is saved correctly to LMCache.
However, when I send **the same prompt** to **server B**, the engine crashes with LMCache-related errors and an `AssertionError` inside `vllm_v1_adapter.py`.

After improving the network between the two servers (higher bandwidth / lower latency / without drop ), the problem stopped happening. This suggests a timing / scheduling / partial KV retrieval bug that appears when the network is slow or unstable and cause crash and restarting image.

---

## Environment

* Docker image: `lmcache/vllm-openai:v0.3.10` and `lmcache/vllm-openai:v0.3.10.post1`

* Python: 3.12 (inside the image)

* Model: `Qwen/Qwen3-8B`

* vLLM:

  * engine launched from the docker image with:

    * `--tensor-parallel-size 2`
    * `--max_model_len 10000`
    * `--trust-remote-code`
    * `--max-num-seqs 16`
    * `--disable-custom-all-reduce`
    * `--kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'`

* LMCache config:

  * via env in vLLM containers:

    * `LMCACHE_CONFIG_FILE=/app/lmcache_config.yaml`
    * `LMCACHE_CHUNK_SIZE=256`
    * `LMCACHE_LOCAL_CPU=False`
  * `lmcache_config.yaml` contents:

    ```yaml
    chunk_size: 128
    local_cpu: true
    remote_url: "lm://192.168.41.201:65432"
    remote_serde: "cachegen"
    ```

* LMCache server:

  * docker image: `lmcache/vllm-openai:v0.3.10` and `lmcache/vllm-openai:v0.3.10.post1`
  * launched as:

    ```bash
    lmcache_server 0.0.0.0 65432
    ```

* Topology:

  * Server A: vLLM + LMCache client
  * Server B: vLLM + LMCache client
  * Both connect to the same `lmcache_server` instance over the network.

---

## Steps to Reproduce

1. Start `lmcache_server`:

   ```bash
   lmcache_server 0.0.0.0 65432
   ```

2. Start **vLLM server A** with:

   ```bash
[...]

Issue #1732, opened 2025-10-01
[Bug] `len(slot_mapping) == len(token_ids)` Assertion failure with Decode Context Parallelism


josephrocca:
**Describe the bug**
When running DeepSeek V3/R1 with `-tp 4 -dcp 4` on 4xH200, LMCache fails the assertion `assert len(slot_mapping) == len(token_ids)`.

That said, I think this may actually be a bug with vLLM's DCP, regarding its interaction with the KV cache storage/retrieval? I'm not sure.

**To Reproduce**
I've tried `0.3.7` and `0.3.8.dev6` - the issue is the same.
```sh
uv pip install vllm[flashinfer]==0.10.2 --torch-backend=auto
uv pip install lmcache==0.3.7
cat > /root/lmcache_config.yaml << EOF
chunk_size: 256
local_cpu: true
max_local_cpu_size: 80
EOF
LMCACHE_USE_EXPERIMENTAL=True LMCACHE_CONFIG_FILE="/root/lmcache_config.yaml" CUDA_VISIBLE_DEVICES='0,1,2,3' vllm serve RedHatAI/DeepSeek-R1-0528-quantized.w4a16 --port 4000 --tensor-parallel-size 4 -dcp 4 --served-model-name default --max-model-len 8192 --host 0.0.0.0 --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
```
Then run this:
```sh
[...]

PR #2509, opened 2026-01-29
feat(kv_cache): enable asymmetric store/retrieve storages in PD backend

hlin99:
Remove the restriction that prevented using `save_decode_cache` and `remote_backend` simultaneously in Prefill-Decode (PD) separation scenarios.

This change introduces `pd_retrieve_locations` and `pd_store_location` parameters to decouple the KV cache retrieval and storage logic. This enables an asymmetric cache flow:
1. Prefill nodes transmit KV cache to Decode nodes via the PDBackend.
2. Decode nodes write back their generated KV cache to a remote backend for subsequent prefill reuse.
3. In multi-turn dialogue scenarios, subsequent prefill requests retrieve historical KV cache from the remote backend, significantly increasing Prefix Cache hit rates and reducing TTFT

This decoupling provides greater flexibility for cross-instance cache management and improves overall pipeline efficiency in distributed inference.

Gemini_Generated_Image_ue7nxmue7nxmue7n

Workflow:
1. Prefill -> Decode (PDBackend): Initial KV transfer for the current turn.
2. Decode -> Remote (Store): Decode saves updated KV to NFS for persistence.
3. Remote -> Prefill (Retrieve): Next-turn prefill pulls from Remote, drastically increasing Prefix Cache hit rate for multi-turn dialogues.

PR #2516, opened 2026-01-30
[Bugfix] fix crash in wait_for_save when retrieve fail from lmcache_engine

### LINKED ISSUE #2294: [Bug] vllm crashes if retrieve() fails

liubj77:
Fixes #2294 #2327 #1732 #2204 

@DongDongJu @sammshen  Please take a look

In LMCacheConnectorV1Impl, if retrieval from the lmcache_engine fails, vLLM will call _handle_invalid_blocks to process the computed tokens of the request. However, this scenario is not currently handled by LMCacheConnectorV1Impl.

The crash stack show below:
```
(EngineCore_DP0 pid=85772)     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=85772)   File "/usr/lib/python3.12/contextlib.py", line 144, in __exit__
(EngineCore_DP0 pid=85772)     next(self.gen)
(EngineCore_DP0 pid=85772)   File "/home/baojun.lbj/pycharm/vllm/vllm/vllm/v1/worker/kv_connector_model_runner_mixin.py", line 131, in _get_kv_connector_output
(EngineCore_DP0 pid=85772)     kv_connector.wait_for_save()
(EngineCore_DP0 pid=85772)   File "/home/baojun.lbj/pycharm/vllm/vllm/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 187, in wait_for_save
(EngineCore_DP0 pid=85772)     self._lmcache_engine.wait_for_save()
(EngineCore_DP0 pid=85772)   File "/usr/local/lib/python3.12/dist-packages/lmcache/integration/vllm/vllm_v1_adapter.py", line 1105, in wait_for_save
(EngineCore_DP0 pid=85772)     assert len(slot_mapping) == len(token_ids)
(EngineCore_DP0 pid=85772)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=85772) AssertionError
```

It can be easily reproduce:

- the token length of the request is greater than 1 block,
- enforce schedule the request in multiple steps
-  `get_num_new_matched_tokens` return is greater than 0,
    - we can simply hack the code in `get_num_new_matched_tokens`
    ```
    # num_external_hit_tokens = self.lookup_client.lookup(
    #     token_ids,
    #     lookup_id=req_id,
    #     request_configs=request_configs,
    # )
    num_external_hit_tokens = 2
    ```
    
- `self.lmcache_engine.retrieve` return fail in `start_load_kv`
    - just comment `_process_tokens_internal` in `LMCacheEngine.retrieve` to simulate retrieve fail

The vLLM start command like below:

```
LMCACHE_CHUNK_SIZE=1 vllm serve facebook/opt-125m --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}' --chat-template=/home/baojun.lbj/work/chat-tmpl.jinja --max-num-batched-tokens=4 --no-async-scheduling
#  set LMCACHE_CHUNK_SIZE=1 
#  set max-num-batched-tokens=4, so the request can be easily scheduled in multiple steps
```

The prompt like below:

```
curl -X POST "http://localhost:8000/v1/completions" -H "Content-Type: application/json" -d '{
    "model": "facebook/opt-125m",
    "prompt": "Where is the capital of China? And tell me what the whether of Hangzhou, please as detail as possible",
    "temperature": 0.0,
    "best_of": 1,
    "max_tokens": 100
}'
```

The same crash will appear when the above request is received.

After this change, I compared the request.num_computed_tokens in self._unfinished_requests and len(self._request_trackers[req_id].token_ids), which are exactly the same, and there is no way to guarantee it before.

I have been running this modification online for more than 4 day until now. Before that, the same pressure tests would crash in up to 3 to 4 hours.

**If applicable**:


DongDongJu:
Hello @liubj77,
Thank you for your hard work! 
Do you know when retrieve failed in vllm?

hlin99:
> Hello @hlin99, Thank you for your hard work! Do you know when retrieve failed in vllm?

hi @DongDongJu , you probably input comments in wrong context? this PR is not from me :)

ps, my pending PRs are https://github.com/LMCache/LMCache/pull/2467, https://github.com/LMCache/LMCache/pull/2478, https://github.com/LMCache/LMCache/pull/2509, please also help review. thanks!  👍

hlin99:
btw, for retrieve fail... I'm really not aware such handing in vllm. maybe we're not running the exact version or cases.... so i can't tell more about this patch.

DongDongJu:
> > Hello @hlin99, Thank you for your hard work! Do you know when retrieve failed in vllm?
> 
> hi @DongDongJu , you probably input comments in wrong context? this PR is not from me :)
> 
> ps, my pending PRs are #2467, #2478, #2509, please also help review. thanks! 👍

My bad. Sorry! I will take a look too.

liubj77:
> Hello @liubj77, Thank you for your hard work! Do you know when retrieve failed in vllm?

@DongDongJu  I discovered this issue while using eic_connector to proxy requests to our backend. The hack code above is to reproduce this scenario.

- In lmcache, if `retrieve` in `start_load_kv` fails, `_invalid_block_ids` will be updated. 
- vllm calls `output.invalid_block_ids = kv_connector.get_block_ids_with_load_errors()` in function `KVConnectorModelRunnerMixin._get_kv_connector_output` to obtain these invalid_block_ids for processing. 
- Then the processing function is `_handle_invalid_blocks` called by `update_from_output` in scheduler, which modifies `request.num_computed_tokens`, but lmcache does not handle this scenario.

This logic has been present in vllm from version 0.12 to 0.14; I haven't checked earlier versions.

DongDongJu (commented):
This makes sense to me. 

When KV retrieval fails, vLLM can roll a request back by adjusting `request.num_computed_tokens` to the longest valid prefix (invalid-block handling). 

Before this change, `request_tracker.token_ids` could remain ahead of `num_computed_tokens`, which matches the reported `wait_for_save` crash `len(slot_mapping) != len(token_ids)`.

With this patch rollback `num_current_tokens < len(token_ids)` and truncate tracker state so `token_ids` stays aligned with vLLM’s rolled-back progress, preventing the assertion.

Nit: Should `num_saved_tokens` be clamped to `tokens_to_keep` to keep it <= len(token_ids) after truncation?

liubj77:
> This makes sense to me.
> 
> When KV retrieval fails, vLLM can roll a request back by adjusting `request.num_computed_tokens` to the longest valid prefix (invalid-block handling).
> 
> Before this change, `request_tracker.token_ids` could remain ahead of `num_computed_tokens`, which matches the reported `wait_for_save` crash `len(slot_mapping) != len(token_ids)`.
> 
> With this patch rollback `num_current_tokens < len(token_ids)` and truncate tracker state so `token_ids` stays aligned with vLLM’s rolled-back progress, preventing the assertion.
> 
> Nit: Should `num_saved_tokens` be clamped to `tokens_to_keep` to keep it <= len(token_ids) after truncation?

@DongDongJu I don't think these two are the same concept. `num_saved_tokens` simply indicates the number of tokens already saved to lmcache.  And both new requests and preempted requests will retrieve this information from the last lookup result, and this length will not exceed the length of `num_computed_tokens`.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1543,6 +1543,35 @@ def build_connector_meta(
...
+                        num_token_slots,
+                        num_current_tokens,
+                    )
+                    tokens_to_keep = num_token_slots
+
+                request_tracker.token_ids = list(
+                    request.all_token_ids[:tokens_to_keep]
+                )
+                request_tracker.num_saved_tokens = min(
+                    request_tracker.num_saved_tokens, num_current_tokens
```
DongDongJu: My concern is here.
IIUC, vllm using `num_saved_tokens` as `skip_leading_tokens`.
please correct me if im wrong.
Can we change `num_current_tokens` to `tokens_to_keep`?

DongDongJu (commented):
I indicated the line I mentioned earlier.

lmcache/integration/vllm/vllm_v1_adapter.py
liubj77: I've reviewed the logic again,  as you said, using `tokens_to_keep` instead of `num_current_tokens` would be more robust, and I've updated the pr.

When retrieval fails, the computed number of tokens must be less than both the previous `num_saved_tokens` and `last_allocated_block_ids * block_size`. Therefore, the condition `num_token_slots < num_current_tokens` will never be triggered in this case, here is just a double check.

DongDongJu: Yes exactly. Thanks!

DongDongJu:
please check the DCO

liubj77:
> please check the DCO

@DongDongJu  done

ziruiliu:
I applied this change to my test in #2294, this change does not fix the problem. The assertion disappeared but the next line failed:
```
(EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/vllm_exp/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 187, in wait_for_save
(EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     self._lmcache_engine.wait_for_save()
(EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/LMCache-0.3.12/lmcache/integration/vllm/vllm_v1_adapter.py", line 1123, in wait_for_save
(EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     slot_mapping = slot_mapping.to(self.device)
(EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982] torch.AcceleratorError: CUDA error: device-side assert triggered
```

I believe other lines should be also modified to completely fix this issue, like this
```python
    def update(...)
    ...
            self.token_ids.extend(new_token_ids)
```

I added logs here and found it still mistakenly extend self.token_ids

Hope this helps

liubj77:
> I applied this change to my test in #2294, this change does not fix the problem. The assertion disappeared but the next line failed:
> 
> ```
> (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/vllm_exp/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 187, in wait_for_save
> (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     self._lmcache_engine.wait_for_save()
> (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/LMCache-0.3.12/lmcache/integration/vllm/vllm_v1_adapter.py", line 1123, in wait_for_save
> (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     slot_mapping = slot_mapping.to(self.device)
> (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982] torch.AcceleratorError: CUDA error: device-side assert triggered
> ```
> 
> I believe other lines should be also modified to completely fix this issue, like this
> 
> ```python
>     def update(...)
>     ...
>             self.allocated_block_ids.extend(new_block_ids)
>             self.token_ids.extend(new_token_ids)
> ```
> 
> I added logs here and found it still mistakenly extend self.token_ids
> 
> Hope this helps

Can you confirm that these are the same issue? This fixes the `assert len(slot_mapping) == len(token_ids)` error. `allocated_block_ids` cannot be guaranteed to align with `token_ids` because blocks allocated in the last vLLM allocation were not reclaimed, and `slot_mapping` is pruned based on `len(token_ids)`.

ziruiliu:
> > I applied this change to my test in #2294, this change does not fix the problem. The assertion disappeared but the next line failed:
> > ```
> > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/vllm_exp/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 187, in wait_for_save
> > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     self._lmcache_engine.wait_for_save()
> > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/LMCache-0.3.12/lmcache/integration/vllm/vllm_v1_adapter.py", line 1123, in wait_for_save
> > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     slot_mapping = slot_mapping.to(self.device)
> > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982] torch.AcceleratorError: CUDA error: device-side assert triggered
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I believe other lines should be also modified to completely fix this issue, like this
> > ```python
> >     def update(...)
> >     ...
> >             self.allocated_block_ids.extend(new_block_ids)
> >             self.token_ids.extend(new_token_ids)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I added logs here and found it still mistakenly extend self.token_ids
> > Hope this helps
> 
> Can you confirm that these are the same issue? This fixes the `assert len(slot_mapping) == len(token_ids)` error. `allocated_block_ids` cannot be guaranteed to align with `token_ids` because blocks allocated in the last vLLM allocation were not reclaimed, and `slot_mapping` is pruned based on `len(token_ids)`.

I guess so. Your fixs does get rid of `assert len(slot_mapping) == len(token_ids)` , but there might be something else remaining so the next line `slot_mapping = slot_mapping.to(self.device)` throws an exception in my test.
As you included in the begining, issue #2294  is similar to your issue that vllm crashes when lmcache found but failed to retrieve the cache. Both hit the same place: `assert len(slot_mapping) == len(token_ids)` because request is rescheduled but slot_mapping is not aligned with token_ids.
However, these 2 issues may have difference root causes. I did not have much time to dive deep into it.

DongDongJu:
> > > I applied this change to my test in #2294, this change does not fix the problem. The assertion disappeared but the next line failed:
> > > ```
> > > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/vllm_exp/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 187, in wait_for_save
> > > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     self._lmcache_engine.wait_for_save()
> > > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]   File "/work/ziliu/LMCache-0.3.12/lmcache/integration/vllm/vllm_v1_adapter.py", line 1123, in wait_for_save
> > > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]     slot_mapping = slot_mapping.to(self.device)
> > > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982]                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> > > (EngineCore_DP0 pid=2978907) ERROR 02-10 09:33:27 [core.py:982] torch.AcceleratorError: CUDA error: device-side assert triggered
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I believe other lines should be also modified to completely fix this issue, like this
> > > ```python
> > >     def update(...)
> > >     ...
> > >             self.allocated_block_ids.extend(new_block_ids)
> > >             self.token_ids.extend(new_token_ids)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I added logs here and found it still mistakenly extend self.token_ids
> > > Hope this helps
> > 
> > 
> > Can you confirm that these are the same issue? This fixes the `assert len(slot_mapping) == len(token_ids)` error. `allocated_block_ids` cannot be guaranteed to align with `token_ids` because blocks allocated in the last vLLM allocation were not reclaimed, and `slot_mapping` is pruned based on `len(token_ids)`.
> 
> I guess so. Your fixs does get rid of `assert len(slot_mapping) == len(token_ids)` , but there might be something else remaining so the next line `slot_mapping = slot_mapping.to(self.device)` throws an exception in my test. As you included in the begining, issue #2294 is similar to your issue that vllm crashes when lmcache found but failed to retrieve the cache. Both hit the same place: `assert len(slot_mapping) == len(token_ids)` because request is rescheduled but slot_mapping is not aligned with token_ids. However, these 2 issues may have difference root causes. I did not have much time to dive deep into it.

Hello @ziruiliu, Thanks for the call out. let me take a look more detail today.

deng451e:
Thanks for the fix, @liubj77! 🙏
Just wanted to add a bit more context on the code flow behind this bug.

When retrieval fails for some blocks, lmcache reports invalid blocks and the vllm scheduler rolls back request.num_computed_tokens in Scheduler._update_requests_with_invalid_blocks, but the allocated block_ids aren’t evicted for request. However, lmcache does not roll back RequestTracker.token_ids accordingly.

Later, in lmache ReqMeta.from_request_tracker, slot_mapping is derived as:
`slot_mapping = (
    block_offsets.reshape((1, block_size))
    + block_ids.reshape((num_blocks, 1)) * block_size
).flatten()[: len(token_ids)]`
When:
`Rounded up(len(token_ids) / block_size) > Rounded up(request.num_computed_tokens / block_size)`
no new blocks are assigned from scheduler because the num_blocks are sufficient for request.num_computed_tokens, causing:
`len(slot_mapping) < len(token_ids)` and triggering the assertion failure.
With your fix, I think simply rolling back RequestsTracker.token_ids to match request.num_computed_tokens should already be sufficient as .
`request_tracker.token_ids = list(request.all_token_ids[:num_current_tokens])`
num_saved_tokens can probably be ignored for now since it mainly affects KV saving (lmcache already has the tokens but retrieval failed), and those cases could be handled separately — e.g., removing the failed keys or verifying their existence (please correct me if I’m mistaken).

sammshen:
@deng451e would you like to create a PR for the fix discussed offline?

liubj77:
> Thanks for the fix, @liubj77! 🙏 Just wanted to add a bit more context on the code flow behind this bug.
> 
> When retrieval fails for some blocks, lmcache reports invalid blocks and the vllm scheduler rolls back request.num_computed_tokens in Scheduler._update_requests_with_invalid_blocks, but the allocated block_ids aren’t evicted for request. However, lmcache does not roll back RequestTracker.token_ids accordingly.
> 
> Later, in lmache ReqMeta.from_request_tracker, slot_mapping is derived as: `slot_mapping = ( block_offsets.reshape((1, block_size)) + block_ids.reshape((num_blocks, 1)) * block_size ).flatten()[: len(token_ids)]` When: `Rounded up(len(token_ids) / block_size) > Rounded up(request.num_computed_tokens / block_size)` no new blocks are assigned from scheduler because the num_blocks are sufficient for request.num_computed_tokens, causing: `len(slot_mapping) < len(token_ids)` and triggering the assertion failure. With your fix, I think simply rolling back RequestsTracker.token_ids to match request.num_computed_tokens should already be sufficient as . `request_tracker.token_ids = list(request.all_token_ids[:num_current_tokens])` num_saved_tokens can probably be ignored for now since it mainly affects KV saving (lmcache already has the tokens but retrieval failed), and those cases could be handled separately — e.g., removing the failed keys or verifying their existence (please correct me if I’m mistaken).

@deng451e Modifying only `request_tracker.token_ids` is fine, but I prefer to modify `num_saved_tokens` synchronously, as this is used to handle `skip_leading_tokens`. 
If a previous retrieve failed, it's highly likely because the token no longer exists in the backend storage, resaving it is better.  On the other hand, some people might wonder why `num_saved_tokens` is longer than `token_ids` when tracking logs,  modifying `num_saved_tokens` synchronously is preferable.

lmcache/integration/vllm/vllm_v1_adapter.py
deng451e, line 1619: Maybe we should also update load_spec.lmcache_cached_tokens here? Otherwise this change could overwrite num_saved_tokens when updating the request tracker.

liubj77: I think this is unnecessary. The `num_saved_tokens` in `RequestTracker` is only updated in preempted scenarios, which is not the case here.

deng451e: I see — LGTM

deng451e (approved):

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1543,6 +1543,31 @@ def build_connector_meta(
...
+                    logger.warning(
+                        "Request %s tracker has %d token slots but %d tokens; "
+                        "capping token_ids to slot capacity.",
+                        req_id,
+                        num_token_slots,
+                        num_current_tokens,
+                    )
+                    tokens_to_keep = num_token_slots
+
+                request_tracker.token_ids = list(request.all_token_ids[:tokens_to_keep])
```
sammshen, line 1606: can we take the difference and slice away with a negative index here instead? it will be slightly more memory efficient IIUC

sammshen (approved):
LGTM! This is a great fix

lmcache/integration/vllm/vllm_v1_adapter.py
liubj77: After consulting some resources and Gemini, using negative indexes did not prove to be more efficient?  It might be more appropriate to retain these changes

DongDongJu (approved):
Thanks for revising. Good to go.

DongDongJu:
@liubj77 lots of CI failed. PTAL

liubj77:
> @liubj77 lots of CI failed. PTAL

@DongDongJu The code formatting has already been handled. Is there anything else I need to do? The workflow requires approval to continue running, and links like buildkite/k3-comprehensive-test are showing "page not found".

sammshen:
please do one more pre-commit @liubj77

sammshen:
updating the branch to address the not found but the k3 tests are non-blocking for now

yanok:
Thanks! That's solves the issues with unreliable backends for us indeed.

PR #2470, opened 2026-01-22
Docs/local disk backend benchmarking

benyebai:
**What this PR does / why we need it**:

Adds a **Benchmarking** section to the Local Storage documentation and fixes incorrect configuration examples.

**Changes:**
1. **New Benchmarking section** - Step-by-step guide using Long Doc QA workload generator to benchmark disk offloading performance
2. **Fixed disk backend config** - The disk backend requires CPU backend to be enabled (`local_cpu: true`). Updated docs to reflect this code requirement.
3. **Added explanatory notes** - Clarifies that CPU RAM is used as a staging layer for disk transfers, and how to set a small `max_local_cpu_size` to isolate disk performance

**Why this is needed:**
- The previous config (`local_cpu: false`) causes an `AssertionError` because the code requires CPU backend for disk operations
- Users need clear instructions on how to benchmark and validate disk offloading is working
- The relationship between CPU and disk backends was not documented

**Special notes for your reviewers**:

This PR stacks on top of #2429 (docs/local-disk-backend-diagrams)

**If applicable**:

- [x] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #2526, opened 2026-02-02
[Docs] Local disk backend benchmarking

benyebai:
## Summary

Adds a **Benchmarking** section to the Local Storage documentation and fixes incorrect configuration examples.

**Changes:**
- **New Benchmarking section** - Step-by-step guide using Long Doc QA workload generator to benchmark disk offloading performance
- **Fixed disk backend config** - The disk backend requires CPU backend to be enabled (`local_cpu: true`). Updated docs to reflect this code requirement.
- **Added explanatory notes** - Clarifies that CPU RAM is used as a staging layer for disk transfers, and how to set a small `max_local_cpu_size` to isolate disk performance

**Why this is needed:**
- The previous config (`local_cpu: false`) causes an `AssertionError` because the code requires CPU backend for disk operations
- Users need clear instructions on how to benchmark and validate disk offloading is working
- The relationship between CPU and disk backends was not documented

---

*This is a clean version of #2470 with only the doc changes (the original branch got polluted with other commits after merging dev).*

**If applicable**:

### DIFF
--- a/docs/source/kv_cache/storage_backends/local_storage.rst
+++ b/docs/source/kv_cache/storage_backends/local_storage.rst
@@ -396,3 +396,104 @@ Tips:
 - If you want to run the ``query-twice.py`` script multiple times, you'll need to either restart the vLLM LMCache server or change the prefix of the context you pass in since you've already warmed LMCache.
 
 - The max model length here was decided by running an L4 with only 23GB of GPU memory. If you have more memory, you can increase the max model length and modify ``query-twice.py`` to use more of the long context. LMCache TTFT improvement becomes more pronounced as the context length increases!
+
+
+.. _local-storage-benchmarking:
+
+Benchmarking
+------------
+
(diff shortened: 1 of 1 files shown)


docs/source/kv_cache/storage_backends/local_storage.rst
```diff
@@ -396,3 +396,104 @@ Tips:
...
+**Step 2. Deploy vLLM with Disk Offloading**
+
+Create a config file ``disk-benchmark.yaml``:
+
+.. code-block:: yaml
+
+    chunk_size: 256
+    local_cpu: false
+    max_local_cpu_size: 10.0
+    local_disk: "file:///tmp/lmcache_disk/"
```
maobaolong, line 425: If we use local_disk, we have to introduce `use_odirect` extra config.

maobaolong (commented):
@benyebai It would be nice if you could add `use_odirect` config to this document.

PR #2529, opened 2026-02-02
[Bugfix] Fix local disk put collisions via per-key locks and atomic replace

DongDongJu:
- Serialize per-key filesystem puts with flock to prevent temp-name collisions and inconsistent cache state.
- Make temp filenames unique and commits atomic via temp+replace, with idempotent put behavior.
- Add a short negative-cache TTL on local disk misses to avoid permanent miss metadata.
- Add unit tests for flock blocking and concurrent puts.

- Lock files are now derived from `key.chunk_hash` (e.g., `<chunk_hash>.lock`).
- New extra_config: `local_disk_missing_recheck_ttl_sec` (default 1.0s, 0 disables negative caching).
- Tests run: `pytest -k fs_connector_flock`, `pre-commit run --all-files`.

**If applicable**:


DongDongJu:
I will reopen after #2734

PR #2538, opened 2026-02-04
feat: add batch operation for eic connector

liubj77:

Currently, the store/load/query operations in eic are performed key by key. I haven't understood the reason for this approach; perhaps it's related to storing meta and data separately?

This pr decides whether to use batch operations based on `self.save_chunk_meta`. This maintains compatibility with the old logic while allowing batch operations when `self.save_chunk_meta` is false. This pr implements these changes.

It also fixes a memory leak issue in `memory_obj` when `get_data` fails.

 @DongDongJu @ApostaC @KuntaiDu  Please take a look.

**If applicable**:


lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -689,8 +714,88 @@ async def _batched_get(
...
+                return [None] * len(keys)
+            else:
+                for i, data_key in enumerate(data_keys):
+                    status_code = get_outcome.status_codes[i]
+                    if status_code != eic.StatusCode.SUCCESS:
+                        logger.error(f"eic batched_get_without_meta {data_key} failed, err_code {status_code}")
+                        memory_objs[i].ref_count_down()
+                    else:
+                        logger.debug(f"eic batched_get_without_meta {data_key} success")
+                        results[i] = memory_objs[i]
```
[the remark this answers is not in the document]
liubj77: The first issue is not a problem, since it's prefix matching, if the first key fails, the other keys will also fail.

lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -689,8 +714,88 @@ async def _batched_get(
...
+                memory_obj.get_memory_format()
+            )
+
+            memory_objs.append(memory_obj)
+            data_keys.append(key.to_string())
+            if self.trans_type == eic.TransportType.TRANSPORT_GDR:
+                data_vals.append(data_ptr, obj_size, True)
+            else:
+                data_vals.append(data_ptr, obj_size, False)
+
```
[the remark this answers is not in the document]
liubj77: By design, this section of the logic will not throw an exception.

Leafykn:
@liubj77 Thanks for the PR for EIC! Just wanted to check if you’re an EIC customer – this’ll help us follow up on the adaptation work smoothly.

liubj77:
> @liubj77 Thanks for the PR for EIC! Just wanted to check if you’re an EIC customer – this’ll help us follow up on the adaptation work smoothly.

@Leafykn We are using the EIC connector to adapt it to our internal storage system, and we found that the EIC connector has a good interface design. Could you please take a look at this PR?

Leafykn:
@liubj77 Is the storage backend your internal storage system, or is it EIC?

liubj77:
> @liubj77 Is the storage backend your internal storage system, or is it EIC?

@Leafykn We use our own internal storage, but it uses the eic interface

Leafykn:
@liubj77 Will you also implement APIs like eic.iobuffer? The API evolution roadmaps of the two different libraries are definitely not aligned, and this approach seems a bit odd to me.

liubj77:
> @liubj77 Will you also implement APIs like eic.iobuffer? The API evolution roadmaps of the two different libraries are definitely not aligned, and this approach seems a bit odd to me.

@Leafykn We can adapt to lmcache implementations, so you don't need to worry about compatibility. We have also adapted to existing open-source interface implementations. I've only implemented the batch interface for EIC here, which will also benefit other EIC users.

lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -689,8 +711,87 @@ async def _batched_get(
...
+            return [None] * len(keys)
+
+        logger.debug(f"Using batch_get_without_meta for {len(keys)} keys")
+
+        data_keys = eic.StringVector()
+        data_vals = eic.IOBuffers()
+        memory_objs: List[Optional[MemoryObj]] = []
+        for key in keys:
+            memory_obj = self.memory_allocator.allocate(
+                self.meta_shapes, self.meta_dtypes, self.meta_fmt
```
Leafykn, line 737: self.meta_shapes, self.meta_dtypes, self.meta_fmt has not init

lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -689,8 +711,87 @@ async def _batched_get(
         self, keys: List[CacheEngineKey]
     ) -> List[Optional[MemoryObj]]:
         # calling self.get will create a circular dependency
-        results = await asyncio.gather(*(self._get(key) for key in keys))
-        return results
+        if self.save_chunk_meta:
```
Leafykn, line 714: Eic also plans to support batch get, and needs to use chunk meta to allocate space, which is not easy to expand here.

Leafykn:
@liubj77 I've noticed that some implementations are not very extensible, and creating a new Connector to adapt to your storage system would likely be a more elegant approach.

lmcache/v1/storage_backend/connector/eic_connector.py
liubj77: this is inited in RemoteConnector. I referred to mooncake's implementation.

lmcache/v1/storage_backend/connector/eic_connector.py
liubj77: This code path is executed when `chunk_meta` is not used, and it wouldn't affect future extensions.

PR #2541, opened 2026-02-04
[Remote Connector]: cpp multi-threaded RESP

sammshen:
@ApostaC as discussed, the interface is agnostic to `asyncio.Future` versus `concurrent.future.Future` so can be folded in as a Controller inside MP Mode. 

This cpp integration has brought the raw client throughput to 10 GB / s (see the `examples/` folder addition) from previous 4-5 GB/s limit with python (optimizations are thoroughly documented in `/csrc/redis/resp.cpp`). LMCache integration now stands at ~7GB/s (compared to previous 0.5 GB/s)


sammshen:
@hickeyma 

do you have any ideas why the github runners cannot run the free disk space action? 

```text
Prepare all required actions
Run ./.github/actions/free-disk-space
Run df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       145G   53G   92G  37% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G 1008K  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
efivarfs        128M   32K  128M   1% /sys/firmware/efi/efivars
/dev/sda16      881M   62M  758M   8% /boot
/dev/sda15      105M  6.2M   99M   6% /boot/efi
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
Run sudo docker rmi "$(docker image ls -aq)" >/dev/null 2>&1 || true
Error: Process completed with exit code 100.
```

Update: should be fixed by https://github.com/LMCache/LMCache/pull/2543

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+  // 2. parse the payload (KV Cache)
+  conn.recv_exactly(buf, len);
+
+  // 3. parse the trailer and validate
+  char trailer[WorkerConn::crlf_len];
+  conn.recv_exactly(trailer, WorkerConn::crlf_len);
+  if (std::memcmp(trailer, conn.crlf, WorkerConn::crlf_len) != 0) {
+    throw std::runtime_error("GET: trailer mismatch");
+  }
+}
```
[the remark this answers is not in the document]
sammshen: I think this suggestion is reasonable. The no parsing optimization might be a little aggressive. Parsing only the first byte avoid O(n) equality check.

csrc/redis/pybind.cpp
sammshen, line 1: follow up: whether to have "pin"/"lock" semantics

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+        throw std::runtime_error("buffer must be byte addressable");
+      if ((size_t)info.size != chunk_bytes_)
+        throw std::runtime_error("buffer size != chunk_bytes");
+
+      tile_req.keys.push_back(keys[i]);
+      tile_req.buf_ptrs.push_back(info.ptr);
+      tile_req.buf_owners.push_back(mv);
+    }
+
+    enqueue_request(std::move(tile_req));
```
sammshen, line 564: double check whether std::move and auto move constructor can work on all the fields inside Request struct (especially py::object)

lmcache/v1/cache_engine.py
```diff
@@ -826,6 +827,7 @@ def retrieve(
                 self.storage_manager.remove(key)
             memory_obj.ref_count_down()
 
+        retrieve_stats.end_time = time.perf_counter()
```
ApostaC: Why do we have this in this PR?

lmcache/v1/storage_backend/connector/base_connector.py
ApostaC, line 1: Can we rename it as "full_chunk_size_bytes"? It's clearer because `full_chunk_bytes` sounds like a data chunk of bytes.

Btw, this name change can be separated to another PR to keep this PR clean.

csrc/redis/resp.h
ApostaC, line 1: Is it possible to avoid having py-related stuff in `resp.h` and `resp.cpp`? Let's try to put all the python-related types and operations into `pybind.cpp`.

FYI, py::list can be directly (and automatically) converted to std::vector if we include  `pybind11/stl.h` (https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html)

We can also manually convert py::memoryview to std::vector as well (or maybe there are better practices).

csrc/redis/pybind.cpp
```diff
@@ -0,0 +1,27 @@
...
+           py::arg("memoryview"))
+      .def("submit_exists", &MultiRESPClient::submit_exists, py::arg("key"))
+      .def("submit_batch_get", &MultiRESPClient::submit_batch_get,
+           py::arg("keys"), py::arg("memoryviews"))
+      .def("submit_batch_set", &MultiRESPClient::submit_batch_set,
+           py::arg("keys"), py::arg("memoryviews"))
+      .def("submit_batch_exists", &MultiRESPClient::submit_batch_exists,
+           py::arg("keys"))
+      .def("drain_completions", &MultiRESPClient::drain_completions)
+      .def("close", &MultiRESPClient::close);
```
ApostaC: If we removed all the py-related stuff from `resp.h` and `resp.cpp`, we could have a few very simple lambda expressions to do the type conversion here.

csrc/redis/resp.h
```diff
@@ -0,0 +1,89 @@
...
+  ~MultiRESPClient();
+
+  MultiRESPClient(const MultiRESPClient&) = delete;
+  MultiRESPClient& operator=(const MultiRESPClient&) = delete;
+
+  int event_fd() const;
+
+  uint64_t submit_get(const std::string& key, py::memoryview mv);
+  uint64_t submit_set(const std::string& key, py::memoryview mv);
+  uint64_t submit_exists(const std::string& key);
```
ApostaC: We can force all the operations to be "batched" and remove these three functions. It can simplify the cpp implementation.

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+  // the number of tiles left to finish
+  // we can only send a completion when the last tile is finished
+  std::atomic<uint32_t> remaining_tiles{0};
+  std::atomic<bool> any_failed{false};
+
+  std::mutex err_mu;
+  std::string first_error;
+
+  // for batch exists, store the boolean results (0/1)
+  // IMPORTANT: not vector<bool> due to concurrent write data race
```
ApostaC: Why not vector<bool> here? What kind of concurrent write data race could occur if we switch to vector<bool>?

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+  py::object buf_owner;
+
+  /*
+  for tiling batched operations
+  */
+
+  // these will be the keys and buffers that this Request is responsible for
+  std::vector<std::string> keys;
+  std::vector<void*> buf_ptrs;
+  std::vector<py::object> buf_owners;
```
ApostaC: It's weird to have two totally different data structures for batched and non-batched requests. Let's just keep batched operations.

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+3. zero copy (no bounce bufferse)
+*/
+struct WorkerConn {
+  int fd = -1;
+  std::string host;
+  int port;
+  size_t chunk_bytes;
+
+  // pre-computed headers
+  std::string size_header;
```
ApostaC: In MP mode, we will have different sizes for different models. We should make it able to deal with variable sizes.

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+  std::string get_prefix;
+  std::string set_prefix;
+  std::string exists_prefix;
+
+  // pre-computed constants (for comparisons)
+  const char* crlf = "\r\n";
```
ApostaC: We can use `constexpr std::string_view` to make it more "C++".

Note that this will need C++17; we need to double-check whether the current compilation supports this.

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+  static constexpr size_t crlf_len = 2;
```
ApostaC: If we use string_view, we can have `static constexpr size_t crlf_len = crlf.size()` here, which is better than hard-coding a "2".

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+
+    {
+      std::ostringstream oss;
+      oss << "$" << chunk_bytes << "\r\n";
+      size_header = oss.str();
+    }
+
+    get_prefix = "*2\r\n$3\r\nGET\r\n";
+    set_prefix = "*3\r\n$3\r\nSET\r\n";
+    exists_prefix = "*2\r\n$6\r\nEXISTS\r\n";
```
ApostaC: Generally, the code here can be put into the constructor. 

Also, the `get_prefix`, `set_prefix`, and `exists_prefix` can be `constexpr std::string_view` IIUC?

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+      }
+      recv_so_far += n;
+    }
+  }
+
+  std::string make_key_header(const std::string& key) {
+    std::ostringstream oss;
+    oss << "$" << key.size() << "\r\n" << key << "\r\n";
+    return oss.str();
+  }
```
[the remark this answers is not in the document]
ApostaC: > The make_key_header function uses std::ostringstream to construct the key header string for each request. While convenient, this can introduce performance overhead due to repeated dynamic memory allocations in a performance-critical path.
>
> For a high-throughput client, consider optimizing this by using a pre-allocated buffer (e.g., thread-local) with snprintf, or a more efficient string building mechanism to avoid allocations on each call.


I agree with Gemini's opinion here. (But only when we can have an upper-bound on the key's size)

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+    throw std::runtime_error(
+        "EXISTS returned invalid response that wasn't :0\r\n or :1\r\n");
+  }
+}
+
+// MultiRESP means multi-threaded RESP (multiple workers)
+// constructor
+MultiRESPClient::MultiRESPClient(std::string host, int port, size_t chunk_bytes,
+                                 int num_workers)
+    : host_(std::move(host)),
```
ApostaC: What's the reason to have `std::string host` and `host_(std::move(host))` instead of using `const std::string &host)` with `host_(host)`?

csrc/redis/resp.cpp
ApostaC: I'm not an expert here, but I do want to get rid of `buf_owners` (and `buf_owner`) here.

From pybind11's doc, I do see we can do `inc_ref` and `dec_ref` on the python objects, including py::memoryview (see https://pybind11.readthedocs.io/en/stable/reference.html#without-reference-counting)

Not sure what the best practice is, but let's investigate and learn.

csrc/redis/resp.h
ApostaC: By doing this, we can also have more freedom to release GIL. See: https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil

> The Python C API dictates that the Global Interpreter Lock (GIL) must always be held by the current thread to safely access Python objects. As a result, when Python calls into C++ via pybind11 the GIL must be held, and pybind11 will never implicitly release the GIL.

csrc/redis/resp.cpp
ApostaC: One thing we need to be careful about is that we need to do dec_ref when the failure happens. Since the tasks are executed asynchronously, I'm not sure what's the best place to do the `dec_ref`.

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+      }
+    }
+  }
+}
+
+// private methods
+void MultiRESPClient::enqueue_request(Request&& req) {
+  {
+    std::lock_guard<std::mutex> lk(req_mu_);
+    requests_.push(std::move(req));
```
ApostaC, line 697: we can probably do `requests_.emplace(req)` here (see https://en.cppreference.com/w/cpp/container/queue/emplace.html)

ApostaC (dismissed):
Did a pass. The major things
1. Try to avoid having python-related stuff in `resp.h` and `resp.cpp`
2. only keep the "batched" primitives

csrc/redis/resp.cpp
DongDongJu: @ApostaC From my understanding,
vector<bool> is [b0,b1,b2 ...,b7] in 1 byte.
and vector<int8> can save like [b0],[b1],[b2].
So, we can not guarantee the atomic byte access for 8 continuous bool value.

Is this what you mean? @sammshen

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+    size_t sent_so_far = 0;
+    size_t iov_idx = 0;
+
+    while (sent_so_far < total_to_send) {
+      ssize_t n = ::writev(fd, &iov[iov_idx], iov.size() - iov_idx);
+      if (n < 0) {
+        if (errno == EINTR) {
+          continue;  // retry on EINTR
+        }
+        throw std::runtime_error("socket writev failed");
```
DongDongJu, line 191: adding retry logic is too aggressive in here?
Eg., if we just failed last vector then retry will be meaningful.

csrc/redis/resp.cpp
DongDongJu: IMO, the best place is the destructor.

DongDongJu (commented):
Hello @sammshen, Thanks for the hard work.
It seems very similar like what I currently do right now.
Left few comments and will revisit tomorrow afternoon.

csrc/redis/resp.cpp
sammshen: yes @DongDongJu's explanation is right. many booleans are compacted into a single vector

lmcache/v1/cache_engine.py
sammshen: needed to check the throughput. this has been broken for a while now

csrc/redis/resp.cpp
sammshen: I think we can get rid of `buf_owners` since we can make the assumption in LMCache use case that the memory objects will not be garbage collected

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+  /*
+  for tiling batched operations
+  */
+
+  // these will be the keys and buffers that this Request is responsible for
+  std::vector<std::string> keys;
+  std::vector<void*> buf_ptrs;
+  std::vector<py::object> buf_owners;
+
+  uint64_t batch_future_id = 0;
```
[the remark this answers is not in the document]
sammshen: wrong

lmcache/v1/cache_engine.py
sammshen: I can separate a new PR for this but this PR needs it as well to test so it will create redundant changes

csrc/redis/resp.h
sammshen: Yes GIL only needed for py::object and py::handle so this will clarify where GIL can be released (do scoped release ASAP in every cpp function)

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+      }
+    }
+  } catch (const std::exception& e) {
+    // catch any exception from connect() or other unexpected errors
+    // log error and exit thread gracefully
+    // we can't throw from here as it would call std::terminate
+    (void)e;  // suppress unused warning
+  } catch (...) {
+    // catch any non-standard exception
+  }
```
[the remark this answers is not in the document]
sammshen: done

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+                 chunk_bytes_);  // one RESP session per worker/thread
+
+    // register socket fd so close() can shutdown this socket
+    {
+      std::lock_guard<std::mutex> lk(worker_fds_mu_);
+      worker_fds_.push_back(conn.fd);
+    }
+
+    for (;;) {
+      Request req;  // make sure req is scoped
```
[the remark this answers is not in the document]
sammshen: removed buf_owner and buf_owners under assumption that python side will not release.

csrc/redis/resp.cpp
sammshen: the req struct already exists (and is populated dynamically) so this does nothing? unless the cpp compiler is a super genius and can somehow figure this out.

csrc/redis/resp.cpp
sammshen: Since host will always be stored in the object (it's a "sink"), this pattern lets the compiler choose the optimal path based on what the caller provides, without needing overloads.
https://stackoverflow.com/questions/21035417/is-the-pass-by-value-and-then-move-construct-a-bad-idiom

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+    throw std::runtime_error("keys and memviews size mismatch");
+  }
+
+  // divide work evenly between workers into tiles (round up, the last tile
+  // will be clipped)
+  size_t num_items = keys.size();
+  size_t num_tiles =
+      std::min<size_t>(num_workers_, num_items);  // avoid empty tiles
+  size_t tile_size = (num_items + num_tiles - 1) / num_tiles;  // round up
+
```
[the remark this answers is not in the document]
sammshen: fixed thanks

csrc/redis/resp.cpp
```diff
@@ -0,0 +1,984 @@
...
+        }
+      }
+    }
+  }
+
+  void recv_exactly(void* buf, size_t len) {
+    size_t recv_so_far = 0;
+    char* ptr = static_cast<char*>(buf);
+    while (recv_so_far < len) {
+      ssize_t n = ::recv(fd, ptr + recv_so_far, len - recv_so_far, 0);
```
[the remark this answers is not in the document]
sammshen: this is handled by timeouts on the lmcache side

csrc/redis/resp.cpp
sammshen: will add cpp17 flag to setup.py

csrc/redis/resp.cpp
sammshen: done

csrc/redis/resp.cpp
sammshen: done

csrc/redis/pybind.cpp
sammshen: done

lmcache/v1/storage_backend/resp_client.py
```diff
@@ -0,0 +1,272 @@
...
+from lmcache.lmcache_redis import LMCacheRedisClient
+
+# we need to import torch since LMCacheRedisClient is built
+# with torch.utils.cpp_extension.*Extension
+# so we are linking against libc10.so and libtorch.so
+
+
+# sync and asyncio wrapper around the LMCacheRedisClient
+# the pybinding interface allows us to work with both sync and async code
+class RESPClient:
```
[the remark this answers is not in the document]
sammshen: done

csrc/redis/resp.h
sammshen: done

csrc/redis/resp.cpp
sammshen: Done.

The optimization is per batch (not per connection) so this change is very easy

sammshen: still requires the assumption that `save_unfull_chunk` = false

sammshen: the chunk_bytes will automatically be detected in the pybind layer. the forced assumption is that all chunks in a batched operation must be the same payload size

csrc/redis/resp.cpp
sammshen: the loop track sent_so_far / recv_so_far so we still make progress with no risk

csrc/redis/pybind.cpp
sammshen: doesn't seem to be in RESP right now

csrc/redis/pybind.cpp
```diff
@@ -0,0 +1,227 @@
...
+              if (info.ndim != 1) throw std::runtime_error("buffer must be 1D");
+              if (info.itemsize != 1)
+                throw std::runtime_error("buffer must be byte addressable");
+
+              bufs.push_back(info.ptr);
+              lens.push_back(static_cast<size_t>(info.size));
+            }
+
+            // use the first buffer's size as chunk_bytes (all must match)
+            size_t chunk_bytes = lens[0];
```
sammshen: save_unfull_chunk must be off

csrc/pybind.cpp
```diff
@@ -39,4 +39,4 @@ PYBIND11_MODULE(c_ops, m) {
         py::call_guard<py::gil_scoped_release>());
   m.def("free_numa_ptr", &free_numa_ptr);
   m.def("get_gpu_pci_bus_id", &get_gpu_pci_bus_id);
-}
+}
```
sammshen, line 42: not sure what this is

csrc/redis/pybind.cpp
```diff
@@ -0,0 +1,227 @@
...
+            py::buffer_info info = py::buffer(mv).request();
+            if (info.ndim != 1)
+              throw std::runtime_error("memoryview must be 1D");
+            if (info.itemsize != 1)
+              throw std::runtime_error("memoryview must be byte addressable");
+
+            std::vector<std::string> keys = {key};
+            std::vector<void*> bufs = {info.ptr};
+            std::vector<size_t> lens = {static_cast<size_t>(info.size)};
+            size_t chunk_bytes = static_cast<size_t>(info.size);
```
DongDongJu: here's info.size is meaning that # of elements(https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html).
I am bit confused that what here's chunk_bytes meaning.

examples/kv_cache_reuse/remote_backends/resp/resp.yaml
```diff
@@ -0,0 +1,10 @@
+chunk_size: 256
+local_cpu: false
+max_local_cpu_size: 20.0
+remote_url: "resp://localhost:6379"
+async_loading: false # LMCache RESP also support async loading
+remote_serde: "naive"
+blocking_timeout_secs: 120
+extra_config:
+  resp_num_threads: 8 # default is 8 if not specified
+  save_chunk_meta: False # make sure we have fixed size payloads
```
DongDongJu: Good to mention this in upper README.md.

lmcache/v1/storage_backend/resp_client.py
```diff
@@ -0,0 +1,219 @@
...
+                "Build with: pip install -e ."
+            )
+        self.loop = loop or asyncio.get_running_loop()
+        self._client = LMCacheRedisClient(host, port, num_workers)
+        self._fd = int(self._client.event_fd())
+        self._closed = False
+
+        # future_id -> (Future, op_name)
+        # we support both types of futures since we only their basic interface
+        self._pending: Dict[
```
DongDongJu, line 60: Is this for preventing the python memory gc?

DongDongJu (commented):
Overall looks good to me.
Could you help me to understand the info.size issue question I left?
Let me test it tomorrow.

csrc/redis/pybind.cpp
sammshen: previously assert itemsize is 1 so number of elements is number of bytes

examples/kv_cache_reuse/remote_backends/resp/resp.yaml
sammshen: done

lmcache/v1/storage_backend/resp_client.py
sammshen: this is to find the python future to which a completion belongs to from the cpp side.

cpp wakes up python via eventfd and outputs (future_id, ok, result_bool, error, result_bools) and python side needs to populate the future indexed by the future_id

csrc/redis/pybind.cpp
sammshen: changed to `batch_chunk_num_bytes`

DongDongJu: Sounds clear to me. Thanks

lmcache/v1/storage_backend/resp_client.py
DongDongJu: Okay then it does prevent GC of python future object not memoryview.

DongDongJu: Is there any possibility that completion drained before _pending insert?
If then we need another types of structure that Tuple[pending_future structure, str, Tuple[memoryview, ...]] types of thing for safe resolving. but It seems this case didnt happened in my test case.
So, just left as a concern in here.

DongDongJu (approved):
Generally LGTM.
I checked that functionally working.
but I left on concern for memoryview gc issue.
We can take care it when it happened.

lmcache/v1/storage_backend/resp_client.py
sammshen: this will not happen because the callback is not preemptive at the python level and the submission is non-blocking!

sammshen:
Addressed correctness concern @DongDongJu . Thanks for your review!

sammshen:
Potential future ideas:
autodiscovery tool for ideal chunk size (based on model and running tests on local redis)

sammshen:
can follow https://github.com/LMCache/LMCache/pull/2607 to extract out some interfaces (batching, GIL release patterns etc.)

maobaolong:
@sammshen Would you like to implement a fs connector also?

maobaolong:
BTW, as discussed with @sammshen and @DongDongJu , https://github.com/LMCache/LMCache/pull/2607 may be a reference.

We can do an abstraction to `native remote backend`.

Clipboard_Screenshot_1771899931

maobaolong (approved):
LGTM, thanks for this great work!

PR #2544, opened 2026-02-05
[Bugfix] fix some memory leak in cache_engine and eic connector

### LINKED ISSUE #2017: [bug] Local CPU memory leaking when using remote connector
**Label**

**Describe the bug**
When using Redis remote connector, I noticed local memory leaking issue and TTFT P50 turns from ~0.4s to 3s after several hours of tests using https://github.com/LMCache/LMBenchmark/blob/main/real-multi-round-qa/multi-round-qa.py#L104. The hours taken to increase seems linear to the max_local_cpu_size config. This issue should not be specific to Redis connector since other connectors are interacting local cpu with similar approach (such as SageMaker HyperPod connector). Using Redis connector as example here since it exists for long time.

In logs, I'm seeing lots of `No eviction candidates found in local cpu backend.`.

With enabling logs in this PR, https://github.com/LMCache/LMCache/pull/1972

I'm seeing multiple cases of local cpu memory leaking.

1. All of items are pinned and never unpinned. `Local CPU backend state: total_items=18, pinned_count=18, ref_count_distribution={1: 18}.` I added some pin/unpin logs locally and noticed it is from VLLM. By some reason, the wait_for_save() which has lookup_pin() is not called after lookup().

2. No items are in hot cache. `Local CPU backend state: total_items=0, pinned_count=0, ref_count_distribution={}.` I suspect it is caused by the items are not registered in hot cache for batched_get() https://github.com/LMCache/LMCache/blob/8353b671b192d1db6bff6a4f96b3068aa3b27ff2/lmcache/v1/storage_backend/storage_manager.py#L450. In get(), it is registered. https://github.com/LMCache/LMCache/blob/8353b671b192d1db6bff6a4f96b3068aa3b27ff2/lmcache/v1/storage_backend/storage_manager.py#L424

**To Reproduce**
```
---

apiVersion: v1
kind: ConfigMap
metadata:
  name: lmcache-config
  namespace: lmcache
data:
  lmcache.yaml: |
    local_cpu: true  # Set to 'true' in production to enable both LMCache native CPU offload AND distributed caching
    chunk_size: 6082
    max_local_cpu_size: 5
    remote_url: "redis://redis.lmcache.svc.cluster.local:6379"

---

apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
  namespace: lmcache
spec:
  replicas: 40
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      namespace: lmcache
      labels:
        app: test
    spec:
      containers:
        - name: vllm
          image: lmcache/vllm-openai:v0.3.9post2
          command:
            - /opt/venv/bin/vllm
            - serve
            - meta-llama/Llama-3.1-8B-Instruct
            - --host
            - 0.0.0.0
            - --port
            - "8000"
            - --enable-prefix-caching
            - --max-model-len
            - "70000"
            - --tensor-parallel-size
            - "4"
            - --kv-transfer-config
            - '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
          resources:
            limits:
              nvidia.com/gpu: "4"
            requests:
              nvidia.com/gpu: "4"
          startupProbe:
            failureThreshold: 60
            httpGet:
              path: /health
              port: 8000
              scheme: HTTP
            initialDelaySeconds: 15
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 1
          env:
            - name: HF_TOKEN
              valueFrom:
                secretKeyRef:
                  key: hf_token_llama
                  name: vllm-secrets
            - name: LMCACHE_CONFIG_FILE
              value: /etc/lmcache/lmcache.yaml
            - name: PYTHONHASHSEED
              value: "0"
              value: "/tmp"
            - name: LMCACHE_LOG_LEVEL
              value: "DEBUG"
          volumeMounts:
            - name: lmcache-config
              mountPath: /etc/lmcache
      volumes:
        - name: lmcache-config
          configMap:
            name: lmcache-config

```

**Expected behavior**
TTFT should be consistent and items should flow in local memory hot cache and be evicted as needed.

**Screenshots**

EKS GPU instances

**Additional context**

liubj77:
Fixes #2017  

@DongDongJu @sammshen @ApostaC Please take a look.

This might be related to this bug, because although I didn't use RedisConnector, I looked at the code and it seems that RedisConnector would also have the same memory leak.

After running for a long period of time, this log will appear.
```
No eviction candidates found in local cpu backend
```

When `async_loading` is not used, memory leaks can occur due to differences in the implementation of some `RemoteConnector` classes, such as whether they support `batched_get`, and differences in the implementation of `batched_get` itself.

For example:
`RedisConnector` does not support batched_get `support_batched_get=False`, it will call the below logic in `batched_get_blocking`, which queries multiple keys simultaneously. This can lead to a problem: the previous key may fail, but the next key will succeed.
```
futures = [
      asyncio.run_coroutine_threadsafe(self.connection.get(key), self.loop)
      for key in keys
  ]
```

Similarly, although `EicConnector` supports batched_get `support_batched_get=True`, but its `batched_get` implementation also queries each key independently in parallel, leading to the same problem as above.

However, function `_process_tokens_internal` in `LMCacheEngine` does not handle this scenario, it only performs `break` and `truncate`, as shown in the code below, this will lead to memory leak of `memory_obj`.

```
for (key, start, end), memory_obj in zip(blocks, memory_objs, strict=False):
      if memory_obj is None:
          logger.warning(
              "The cache block is in the storage, but it can't be retrieved"
          )
          if (
              last_failed_block_start is None
              or last_failed_block_start < start
          ):
              last_failed_block_start = start
          break
```

1.  verify

As shown in the PR's new code, I added logging for unreleased memory. After running it in the test environment for a long time, I confirmed that some `memory_obj` instances are not being released.

image

image

2. other leak

This PR also fixes the memory_obj leak in EICConnector when `get_data` fails.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Touches retrieval control-flow and reference counting, so mistakes could cause premature frees or missing cache hits; changes are bounded and covered by new unit tests for failure/truncation cases.
> 
> **Overview**
> Fixes a KV-cache *memory leak during retrieval* when `batched_get` returns non-contiguous results (e.g., a later key succeeds after an earlier key fails).
> 
> `LMCacheEngine._process_tokens_internal` now (1) tracks the earliest failed block across locations, (2) immediately `ref_count_down()`s any successfully-fetched but unused blocks after a per-location failure, and (3) truncates returned chunks using `end <= last_failed_block_start` while freeing any already-added chunks that extend past the gap.
> 
> `EICConnector.get_data` now guarantees allocated `memory_obj`s are released via `ref_count_down()` on both error status codes and unexpected exceptions during `mget`.
> 
> Adds focused unit tests covering boundary truncation, multi-location failure ordering, freeing of unused successful blocks, and avoiding double-free scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05c404350c4cf7d3abba1056453941bf89becdb2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/storage_backend/connector/eic_connector.py
```diff
@@ -391,6 +391,7 @@ async def get_data(self, key_str: str, meta: RemoteMetadata) -> Optional[MemoryO
                 f"eic mget data {key_str} failed, status_code {status_code}"
                 " err_code {err_code}"
             )
+            memory_obj.ref_count_down()
```
maobaolong, line 406: How about the exception path? In another word, memory_obj should ref_count_down although exception occured.

maobaolong (commented):
@liubj77 Left a comment.

lmcache/v1/storage_backend/connector/eic_connector.py
liubj77: > How about the exception path? In another word, memory_obj should ref_count_down although exception occured.

According to the code design, EIC returns values ​​based on the code and will not encounter any exceptions.

maobaolong: That's ok. There is another trick, we can surround the code after `self.memory_allocator.allocate` and catch all exception, then, we can ensure whatever happened, we can call ref_count_down. The memory object will never leak.

liubj77: done, possible exceptions have been handled in try-catch

sammshen (approved):
LGTM

lmcache/v1/storage_backend/connector/eic_connector.py
liubj77: @maobaolong Please review this again and see if it's acceptable to merge.

liubj77:
@maobaolong @sammshen @deng451e Could you please review this and see if it's acceptable to merge.

deng451e:
LGTM

deng451e (approved):

sammshen (changes requested):
need to take another look sorry

lmcache/v1/cache_engine.py
```diff
@@ -1664,10 +1665,30 @@ def _process_tokens_internal(
                 reordered_chunks.append((key, memory_obj, start, end))
                 tot_kv_size += memory_obj.get_size()
                 ret_mask[start:end] = True
+                used_keys.add(key)
+
+            for (key, _, _), memory_obj in zip(blocks, memory_objs, strict=False):
```
sammshen, line 1680: dumb question: 

why do we need to do this check twice? 

@deng451e has a great question which is: how do you know there is no overlap bewteen these two sets (complement of `used_keys` and `end >= last_failed_block_start`)

liubj77: > dumb question:
> 
> why do we need to do this check twice?
> 
> @deng451e has a great question which is: how do you know there is no overlap bewteen these two sets (complement of `used_keys` and `end >= last_failed_block_start`)

@sammshen  I've modified some logic and added unit tests. please take a look again.

lmcache/v1/cache_engine.py
```diff
@@ -1650,28 +1650,51 @@ def _process_tokens_internal(
...
                         last_failed_block_start is None
-                        or last_failed_block_start < start
+                        # The minimum value should be taken here to ensure that
+                        # the prefix keys are all consecutive successful.
+                        or last_failed_block_start > start
                     ):
                         last_failed_block_start = start
                     break
                 reordered_chunks.append((key, memory_obj, start, end))
                 tot_kv_size += memory_obj.get_size()
```
DongDongJu, line 1676: `tot_kv_size` is accumulated before truncation, but after `last_failed_block_start` some chunks are dropped from `reordered_chunks` without adjusting `tot_kv_size`. This one may over-report bytes and throughput. but not a critical blocker IMO.

DongDongJu (commented):
LGTM, but I am not that familiar with eic connector.
So, just left one my comment from code reading.

lmcache/v1/cache_engine.py
liubj77: @DongDongJu  `tot_kv_size` is currently used only for logging, I have updated the code to ensure that `tot_kv_size` is adjusted synchronously.

liubj77:
> LGTM, but I am not that familiar with eic connector. So, just left one my comment from code reading.

@DongDongJu @sammshen @maobaolong @deng451e  Can this PR be merged?  And the EIC modify didn't change the original logic, it just releases memory on failure. 

I've tested it for about two weeks, and with this patch, lmcache memory leaks have decreased significantly. Of course, there are some other memory leaks that haven't been discovered yet.

image

sammshen (approved):
LGTM

PR #2545, opened 2026-02-05
[Bugfix] Ensure validate() is called in config factory methods

### LINKED ISSUE #2540: [BUG] For PD disaggregation setup, prefill instance should save all prompt tokens despite not meeting chunk boundary for its tail section
**Describe the bug**
Running a PD Disaggregation setup with vLLM v0.9.2 + LMCache v0.3.9 + Mooncake. There is a decent chance the last batch of scheduled tokens onthe prefill instance don't meet the chunk boundary requirement, therefore gotten thrown away even though they were just computed. And the decode instance has to redo the compute before getting into decode phase for the request.

**To Reproduce**
A minimal setup would be having one prefill and one decode instance running vLLM v0.9.2 + LMCache v0.3.9 + Mooncake stack. Devise a request with a particular number prompt tokens, so that after broken up into batches by vLLM, the last batch failed to meet chunk boundary, so the `skip_save` flag in LMCache is set to 'True' wrongly.

e.g. the prompt has 2049 tokens, with vLLM's `max_num_batched_tokens` set to 2048, the 2nd batch would be of size 1, assume the chunk size is sth like 256, the last batch would not be saved.  

**Expected behavior**
All prompt tokens should have their tensors saved by LMCache

**Screenshots**

**Additional context**

RGB-loop:

Configuration validation (`validate()`) was not being called when creating config 
instances via `from_env()`, `from_file()`, or `from_defaults()`. This caused 
auto-correction logic in `validate()` to not take effect, including:

- PD mode: forcing `save_unfull_chunk=True` for complete KV cache transfer
- Blending mode: forcing `save_unfull_chunk=True`

Without this fix, in PD disaggregation setups, tokens that don't meet chunk 
boundaries would be discarded by the prefill instance, causing the decode 
instance to recompute them unnecessarily.

This PR wraps `__post_init__` in `create_config_class()` to automatically call 
`validate()` after instance creation, ensuring validation runs regardless of 
how the config is created.

Fixes [#2540](https://github.com/LMCache/LMCache/issues/2540)

- The fix is minimal: only 13 lines added in one file
- `validate()` is idempotent, so existing explicit calls remain safe
- Tested with both `LMCacheEngineConfig` and `ControllerConfig`


maobaolong (approved):
@RGB-loop Thanks for this fix, LGTM.

PR #2548, opened 2026-02-05
[Chore][Linter] Force SLF001 rule to avoid accessing private member (mp mode only)

ApostaC:

With this PR, we can have a better code quality by enforcing no private member access.

**If applicable**:

- this PR should not need tests


sammshen (approved):
LGTM!

KuntaiDu (approved):
LGTM

PR #2557, opened 2026-02-06
Fixes #2556: Assertion when remote backend is enabled without local CPU backend

### LINKED ISSUE #2556: [Bug] Remote backend fails to initialize when local cpu is off
**Label**
commit 09d3270e84cd7f580739d7f283b205f5e7c5c7df (HEAD -> remote_cpu)
Author: Samuel Shen <102553648+sammshen@users.noreply.github.com>
Date:   Thu Feb 5 14:49:38 2026 -0800

    [docs]: permissions for root files (#2547)

**Describe the bug**
remote backend will fail to initialize if local cpu is turned off

**To Reproduce**
yaml content

local_cpu: False
max_local_cpu_size: 0

remote_serde: naive
remote_url: "lm://localhost:8888"

**Expected behavior**
cmd:
LMCACHE_CONFIG_FILE=lmcache.yaml vllm serve /workspace/Meta-Llama-3-8B-Instruct/   --host 0.0.0.0   --port 8868   --block-size 128   --dtype bfloat16   -tp 1   --max-model-len 8192   --max-num-batched-tokens 8192   --max-num-seqs 16   --gpu_memory_utilization 0.9   --no-enable-prefix-caching   --no-enable-chunked-prefill   --async-scheduling --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both","kv_connector_extra_config": {"discard_partial_chunks": false, "lmcache_rpc_port": "producer1"}}'

**Screenshots**

Image

hlin99:
#2556 includes two underneath issues.
1. SafeLocalCPUBackend config and metadata are missing, which throw exception when connector is initialized
refer to code https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/connector/lm_connector.py#L45

2. Local CPU backend is must have for remote backend, otherwise remote backend won't work properly. so raise an assertion during creating backend is a decent handling way
refer to code: https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/storage_backend/storage_manager.py#L327


maobaolong (approved):
@hlin99 LGTM. thanks for this fix.

hlin99:
Hi @sammshen could u take 5mins to take a look? This is a small safety check. Thx!

lmcache/v1/storage_backend/__init__.py
```diff
@@ -192,6 +192,11 @@ def CreateStorageBackends(
         gds_backend = GdsBackend(config, metadata, loop, dst_device)
         storage_backends[str(gds_backend)] = gds_backend
     if config.remote_url is not None:
+        if metadata.role != "scheduler":
```
sammshen: do we need this? will the scheduler LMCacheEngine create a CacheEngine and then a StorageManager?

sammshen: part of refactor effort will be to move all role related things to adapter. ideally we don't introduce role in lower modules

hlin99: you're right, @sammshen . thanks for pointing this out, scheduler will never get into this code path. PR has been updated. please help double check, thanks!

sammshen (approved):
lgtm

hickeyma (approved):

hlin99:
hi @maobaolong  @sammshen  @hickeyma  just resolved the conflicts and rebased to the latest...   could you help manually merge this PR if no more comment? auto-merge seems does not work...... thanks.

Issue #2562, opened 2026-02-07
Support L2 storage in MP mode


ApostaC:
# Design concept: controller and L2 adapters

<img width="1440" height="431" alt="Image" src="https://github.com/user-attachments/assets/fdcfd2e5-32a7-4532-8c31-c6997ee226ee" />

**2 ROLES: controller and L2 adapter**
- The controller needs to handle the "control logic" (i.e., acquiring/deleting the L1 objects, emitting observability data, etc)
- The L2 adapter focuses on I/O and has standard interfaces that can be easily implemented in native languages.
- The L2 adapter uses an event fd to notify the controller whether there are new IO operations finished. This can avoid the controller having a busy loop.

# Execution plan (list of PRs)

- [x] Simplest L2 adapter implementation: use host memory to mock an "L2 device"
  - [x] Bitmap implementation (#2563)
  - [x] L2 adapter implementation (#2569)
- [x] Store controller implementation (#2646)
- [x] Prefetch controller implementation
  - [x] Prefetch controller foundation: `PrefetchPolicy`, `PrefetchHandle` redesign, `L1Manager.finish_write_and_reserve_read()` (#2658 — needs rebase after #2646)
  - [ ] Prefetch controller core: multi-phase state machine, event-driven background thread
  - [ ] StorageManager integration: wire prefetch controller, update `submit_prefetch_task` / `query_prefetch_status`

ApostaC:
Closed by the final PR #2667

PR #2563, opened 2026-02-07
[1/4] Bitmap for L2 storage in MP mode 

ApostaC:

Part of #2562 

**If applicable**:


csrc/storage_manager/bitmap.cpp
```diff
@@ -0,0 +1,144 @@
...
+    uint8_t mask = static_cast<uint8_t>((1u << remaining_bits) - 1);
+    last_byte &= mask;
+    count += static_cast<size_t>(
+        __builtin_popcount(static_cast<unsigned>(last_byte)));
+  }
+  return count;
+}
+
+size_t Bitmap::clz() const {
+  if (size_ == 0) return 0;
```
KuntaiDu, line 61: What does `size_` mean? Is it defined?

csrc/storage_manager/bitmap.cpp
```diff
@@ -0,0 +1,144 @@
+// SPDX-License-Identifier: Apache-2.0
```
KuntaiDu, line 1: Why not std::set?

csrc/storage_manager/bitmap.cpp
ApostaC: It's a private member of this class defined in bitmap.h

csrc/storage_manager/bitmap.cpp
ApostaC: It's for a totally different purpose. Probably a similar one is `std::vector<bool>` but it doesn't provide the functionality of "count leading ones", which will be used by the python's lookup function

KuntaiDu (approved):
Code is straightforward for me.

ApostaC:
Going to merge this as we discussed offline.

PR #2575, opened 2026-02-09
[Doc]: Fix to update the Move API doc.

### LINKED ISSUE #2571: [BUG]: LMCache docs for moving the kv block contains mistakes.
**Doc for KV block move API demonstration contains a mistake**

**Describe the bug**

- https://docs.lmcache.ai/kv_cache_management/move.html The doc includes a small mistake in labelling the instance_id name for second instance2.yaml thus it was failing to move the sed.

- Please cor,rsome if I am wrong, kv_events should be enabled for the kv wrongongm thevement to happen. While we didn't add this in the config, thus towas facing 0 locks repeatedly seeingrepeatedly until I aconfig to register this., I think we should add the same as well.
- Also, as the UCX_TLS mode taken as the base example is in rc mode, it comes with a base assumption that it would only work with Infiniband, I think we should have an example maybe or some helper doc to suggest that how we can condigure the same to accept CUDA ipc mode.

**To Reproduce**
1. Go to https://docs.lmcache.ai/kv_cache_management/move.html 
2. Please follow the doc, you should have "0" num blocks reported.
`{"event_id":"Move492fba71-649f-42bc-a3f5-d42f5b587825","num_tokens":0}%`

**Expected behavior**

`{"event_id":"Movecad79ccd-21a0-45f3-81ee-6cef434698f5","num_tokens":12}%  `

**Screenshots**
Image

Please look at the name for thof the issues. Rest other issue are betteabove.

 - OS: LinuxOS

**Additional context**
I pretty much got to a debuggin state where I was able to edit it with right config and solve the issue, would open a PR soon to get this fixed.

Thanks
Sneh

atalhens:
This PR corrects critical configuration errors in the Move API documentation that would prevent the example from working correctly.
- Added missing enable_kv_events: True to both instance configurations required for KV cache event tracking during move operations. Without this, the move functionality cannot properly monitor the transfer status.
- Fixed incorrect instance ID in instance2.yaml
Changed from lmcache_instance_1 to lmcache_instance_2
This was a copy-paste error that would cause both instances to have the same ID, breaking P2P transfer

Closes https://github.com/LMCache/LMCache/issues/2571
**If applicable**:


docs/source/kv_cache_management/move.rst
```diff
@@ -33,6 +33,7 @@ configure two lmcache instances:
 
     # cache controller configurations
     enable_controller: True
+    enable_kv_events: True
```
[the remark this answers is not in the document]
DongDongJu: @atalhens I agreed that good to have a comment in here for whom not familiar with kv_event.

DongDongJu (changes requested):
Generally LGTM, please add comment for kv_events.

sammshen (approved):
LGTM!

DongDongJu (approved):
Thanks for the contribution. LGTM.

docs/source/kv_cache_management/move.rst
```diff
@@ -33,6 +33,7 @@ configure two lmcache instances:
 
     # cache controller configurations
     enable_controller: True
+    enable_kv_events: True  # Required to track KV cache events for move operations by controller.
```
maobaolong, line 36: why this is a necessary config for move example?

maobaolong (commented):
left a question @atalhens .

docs/source/kv_cache_management/move.rst
atalhens: So I tried the provided config in this example, apparently the move operation was not causing any operation post that I looked a little in deep, and IIUC for the move operation to work, the controller must need the information about the kv block stored in each of the instances, which is only provided if we have the kv_events emitted in place. Thus, as soon as I realized it was not enabled by default, I tried enabling it, and it worked.

Please let me know if I am wrong here. Also, if this is the case, I was curious why we are not enabling it by default.

maobaolong: If `enable_controller` set, `LMCacheWorker` should be created while `LMCacheEngine` init, then lmcache controller can retrieve `admit` and `evict` `KVOpEvent`, this is my understanding of `LMCacheController` and `KVOpEvent` syncing. 

But I'm not very familiar with `KVEvent` framework, so I looked a little code too, now I'm confusing where invoke the `LMCacheConnectorV1Impl#get_kv_events`?

atalhens: Thank you so much for questioning. I got confused due to my naivety about the codebase. 

Apparently, while I was setting my own config locally while testing, I added two variables: `enable_kv_events` and `save_unfull_chunk`, and used a very big prompt. I assumed that it was save_unfull_chunk that was helping me save the complete chunk, including the incomplete partial blocks. At the same time, it was kv_events that was actually helping create the `CacheStoreEvent`, which was apparently my initial thought.

While digging deep now, I understood the example was not working before because the request prompt given in the LmCache doc was very short (< 256 tokens), and in config we did not set the `save_unfull_chunks` flag, which defaults to false. 

```
    "save_unfull_chunk": {
        "type": bool,
        "default": False,
        "env_converter": _to_bool,
    },
```
Thus, it was always showing me 0 tokens. And as soon as I set the other variable, it started working. But apparently, I assumed kv_event got that working.

Poligies for it, Now we should either update the config to include 'save_unfull_chunks' or change the prompt to a larger one. Let me know what is the suggested way I can make a change as per that.

hickeyma: @atalhens KV Cache events are not generated for move operation. Events are generated for only store operations at the moment. They also are not necessary for operations and are used usually for KV-Cache-Aware routing in districuted inference platforms. 

Refer to following doc for more details: https://docs.lmcache.ai/production/kv_cache_events.html

hickeyma (changes requested):

docs/source/kv_cache_management/move.rst
atalhens: I totally understand. I added to the comment as well. Apparently, the example given didn't work as-is because save_unfull_chunk is set to false by default while the example given has much less token than 256 which is the minimum token length required.

PR #2611, opened 2026-02-17
[HW: XPU] Enable Layerwise XPU Connector

slokesha:
This PR adds full Layerwise KV cache support for XPU connectors, bringing feature parity with the CUDA connector implementation.

It enables LMCache’s use_layerwise=True workflow for XPU devices, supporting:
- Layerwise KV retrieve (batched_from_gpu)
- Layerwise KV restore (_batched_to_gpu_gen)
- Optional GPU staging (use_gpu=True)

**If applicable**:


lmcache/v1/gpu_connector/xpu_connectors.py
```diff
@@ -241,3 +241,528 @@ def from_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
     def batched_to_gpu(self, memory_objs, starts, ends, **kwargs):
         for memory_obj, start, end in zip(memory_objs, starts, ends, strict=False):
             self.to_gpu(memory_obj, start, end, **kwargs)
+
+def _split_token2d_kv(token2d: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
sammshen: can you move this to gpu_connector/utils.py

lmcache/v1/gpu_connector/xpu_connectors.py
```diff
@@ -241,3 +241,528 @@ def from_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
...
+    if token2d.dim() != 3:
+        raise ValueError(f"Expected token2d dim=3, got {token2d.shape}")
+    if token2d.shape[0] == 2:  # [2, T, D]
+        return token2d[0], token2d[1]
+    if token2d.shape[1] == 2:  # [T, 2, D]
+        return token2d[:, 0, :], token2d[:, 1, :]
+    raise ValueError(f"Unrecognized token2d layout: {token2d.shape}")
+
+
+def _get_paged_kv_views(
```
sammshen: can you rename this,` _get_head_size_view()` or something similar and move it to utils.py as well. all the helpers on kv cache should be generalizable and put in one place.

this method (since it deals with GPU side kv cache) should take in the GPUKVFormat enum.

lmcache/v1/gpu_connector/xpu_connectors.py
slokesha: Moved to gpu_connector/utils.py.

lmcache/v1/gpu_connector/xpu_connectors.py
slokesha: renamed the function to _get_head_size_view() and added gpu_kv_format: lmc_ops.GPUKVFormat as an optional argument.

lmcache/v1/gpu_connector/xpu_connectors.py
```diff
@@ -241,3 +242,468 @@ def from_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
...
+            use_mla=metadata.use_mla,
+        )
+
+    def _lazy_initialize_buffer(self, kv_caches: List[torch.Tensor]) -> None:
+        # Buffer allocator only needed when use_gpu=True (device staging)
+        if self.use_gpu and self.gpu_buffer_allocator is None:
+            # Import here to avoid circulars
+            from lmcache.v1.memory_management import GPUMemoryAllocator
+            staging_bytes = 256 * 1024 * 1024
+            self.gpu_buffer_allocator = GPUMemoryAllocator(size=staging_bytes)
```
DongDongJu: Can we add device=self.device to param?

lmcache/v1/gpu_connector/xpu_connectors.py
```diff
@@ -241,3 +242,468 @@ def from_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
...
+                                mem.tensor.copy_(gathered.to(mem.tensor.device), non_blocking=True)
+                    else:
+                        for s, e, mem in zip(starts, ends, mem_layer, strict=False):
+                            assert mem.tensor is not None
+                            sl = slot_mapping_on_device[s:e]
+                            gathered = src_flat.index_select(0, sl)
+                            mem.tensor.copy_(gathered.to(mem.tensor.device), non_blocking=True)
+
+                    # keep fmt update
+                    mem.metadata.fmt = MemoryFormat.KV_MLA_FMT
```
DongDongJu: Can we make the format validation flow in here?

lmcache/v1/gpu_connector/xpu_connectors.py
```diff
@@ -241,3 +242,468 @@ def from_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
...
+            device=device,
+            use_mla=metadata.use_mla,
+        )
+
+    def _lazy_initialize_buffer(self, kv_caches: List[torch.Tensor]) -> None:
+        # Buffer allocator only needed when use_gpu=True (device staging)
+        if self.use_gpu and self.gpu_buffer_allocator is None:
+            # Import here to avoid circulars
+            from lmcache.v1.memory_management import GPUMemoryAllocator
+            staging_bytes = 256 * 1024 * 1024
```
DongDongJu: Why its hard coded?

DongDongJu (commented):
Hello @slokesha , Thanks for the great work!
I left few questions.
Please add following tests: 
- Multi-chunk starts/ends (e.g., two or three segments) for both layerwise and non-layerwise
- `use_gpu=True ` test or forcing use_gpu to false when xpu enabled..

lmcache/v1/gpu_connector/xpu_connectors.py
slokesha: Now the staging_bytes is calculated using kv_cache or can be set using env variable LMCACHE_GPU_STAGING_BUFFER_BYTES

lmcache/v1/gpu_connector/xpu_connectors.py
slokesha: Added function _validate_format_transition(), to valdidate if the fmt is in ALLOWED_FORMAT_TRANSITIONS

slokesha:
@DongDongJu @sammshen , 
Thank you for the comments. I’ve made the requested changes and added additional test coverage.

The new tests cover use_gpu=true and multi-chunk scenarios in test_xpu_connector.py.

I’m not completely sure whether the benchmark test is necessary for this PR, but I’ve kept it for now. Happy to remove it if you think it doesn’t belong in the test suite.

lmcache/v1/gpu_connector/__init__.py
```diff
@@ -71,6 +71,18 @@ def CreateGPUConnector(
         torch_dev.set_device(local_worker_id)
         device = torch.device(f"{dev_name}:{local_worker_id}")
 
+        if dev_name == "xpu":
+            # First Party
+            from lmcache.v1.gpu_connector.xpu_connectors import (
+                VLLMPagedMemXPUConnectorV2,
+                VLLMPagedMemLayerwiseXPUConnector,
+            )
+            if config.use_layerwise:
```
DongDongJu, line 81: what happened if enable_blending is true in here?

lmcache/v1/gpu_connector/utils.py
```diff
@@ -443,3 +445,109 @@ def get_dtype(kv_caches: Any, gpu_kv_format: "lmc_ops.GPUKVFormat") -> torch.dty
...
+    kv_cache_layer: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
+    *,
+    use_mla: bool,
+    gpu_kv_format: Optional["lmc_ops.GPUKVFormat"] = None,
+) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
+    """
+    Returns flattened views for index_copy/index_select.
+
+    If gpu_kv_format is provided, use it to interpret tensor layout explicitly.
+    If not provided, fall back to current structural behavior:
```
DongDongJu, line 522: When we need fallback?

lmcache/v1/gpu_connector/xpu_connectors.py
```diff
@@ -241,3 +248,586 @@ def from_gpu(self, memory_obj: MemoryObj, start: int, end: int, **kwargs):
...
+            for _ in range(self.num_layers):
+                _ = yield
+            yield
+            if sync:
+                torch.xpu.current_stream().wait_stream(self.load_stream)
+            yield
+            return
+
+        tmp_gpu_buffer_obj: Optional[MemoryObj] = None
+        if self.use_gpu:
```
DongDongJu: IMO its so confused that use_gpu and xpu settings. can we do move the xpu logic under use_xpu in later? not use_gpu?

DongDongJu (commented):
Hello @slokesha,
I left few comments.

lmcache/v1/gpu_connector/xpu_connectors.py
slokesha: Yeah, I was a bit confused too. In the current
implementation, --use_gpu actually controls whether we create a device
staging buffer. In this case ends up being XPU staging
buffer, so I used the same flag to be consistent with the existing
`VLLMPagedMemXPUConnectorV2`.

I agree the naming is not ideal. We could clean this up by
introducing something like `use_xpu` and refactoring the connectors in a follow-up PR.

If you’d prefer, I can also remove `use_gpu` and introduce a `use_xpu` in this PR.

lmcache/v1/gpu_connector/__init__.py
slokesha: This PR doesn’t change the blending behavior — it just enables the
layerwise XPU connector so it can be used in the same scenarios as the
existing paged XPU connector.

lmcache/v1/gpu_connector/utils.py
slokesha: Right now the fallback path can still detect and handle the two current
stacked KV layouts from the tensor shape:
- `[2, NB, BS, NH, HS]`
- `[NB, 2, BS, NH, HS]`

When `gpu_kv_format` is passed, we use it explicitly to interpret the
layout. At the moment that still covers the same two formats, but it gives
us a cleaner path to support more KV layouts in the future without relying
only on shape-based fallback logic.

lmcache/v1/gpu_connector/xpu_connectors.py
DongDongJu: That sounds great idea 👍. cleanup with use_xpu makes easy to read.

lmcache/v1/memory_management.py
```diff
@@ -2416,3 +2416,84 @@ def batched_free(
...
+
+    def __init__(
+        self,
+        size: int,
+        device="xpu",
+        align_bytes: Optional[int] = None,
+        use_paging: bool = False,
+        **kwargs,
+    ):
+        self.tensor = torch.empty((size,), dtype=torch.uint8, device=device)
```
sammshen, line 2523: is there no page locking on XPU?

sammshen (approved):
LGTM

DongDongJu (approved):
Please make the folllowing changes for the use_xpu cleanup!
LGTM

sammshen:
also please run the pre-commit

lmcache/v1/memory_management.py
slokesha: Good catch! XPU does support page locking via SYCL USM host allocation.

The device tensor here :
`self.tensor = torch.empty((size,), dtype=torch.uint8, device=device)` - doesn't need pinning just like GPUMemoryAllocator, page locking is used on CPU tensors. The fix is in `non_cuda_equivalents.py`, which is used on the non-CUDA code path to allocate CPU staging buffers. Previously those buffers hardcoded `pin_memory=False`. Now I've added a module-level flag: 
`_XPU_PIN_MEMORY = hasattr(torch, "xpu")`
and changed alloc_pinned_ptr and alloc_pinned_numa_ptr to use `pin_memory=_XPU_PIN_MEMORY`, so CPU staging buffers are properly page-locked when running on XPU.

lmcache/v1/gpu_connector/xpu_connectors.py
slokesha: Done 👍

slokesha:
@sammshen and @DongDongJu, Can we merge this PR?

sammshen:
rerunning CI

PR #2614, opened 2026-02-18
[3/N][Feat]Persist metadata on device and fix raw-device benchmark setup

DongDongJu:

  - Replaces rust raw block backend metadata persistence path to use on-device checkpointing
  - Adds metadata format/version config support and robust load/apply validation behavior.
  - Updates rust raw block backend unit tests to match the new metadata persistence/recovery semantics.
  - Fixes storage backend microbenchmark behavior for real raw devices by skipping truncate() on /dev/* paths.

This is needed to keep metadata colocated with raw block storage, simplify recovery semantics, and avoid invalid benchmark setup on real block devices.

  If applicable:


lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -743,52 +874,168 @@ def _save_manifest(self, path: str) -> None:
...
+            self.meta_magic,
+            self.meta_version,
+            int(next_seq),
+            int(payload_len),
+            int(crc),
+        )
+
+        raw = self._rawdev()
+
+        raw.pwrite_from_buffer(payload_off, payload, payload_len, payload_total_len)
```
sammshen, line 940: checkpoint payload is written in-place (payload_off) before header commit. if the process/device crashes during that payload write, the previous payload is partially overwritten while the old header is still present. on restart CRC check fails and the loader drops metadata entirely so restart recovery can be lost after a single torn write.

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -743,52 +874,168 @@ def _save_manifest(self, path: str) -> None:
...
+                            k_str,
+                            e,
+                        )
                         continue
-                    offset = int(e.get("offset", 0))
-                    size = int(e.get("size", 0))
-                    shape_list = e.get("shape")
-                    fmt_name = e.get("fmt")
+
+                    offset = int(entry.get("offset", 0))
```
sammshen, line 1074: offset/size from checkpoint entries are accepted without checking:

offset >= data_base_offset
slot alignment ((offset - data_base_offset) % slot_bytes == 0)
size > 0 and size <= slot_bytes - header_bytes

DongDongJu:
@maobaolong @sammshen Could you take a look by any chance?

sammshen:
@DongDongJu I left two comments previously :)

DongDongJu:
> @DongDongJu I left two comments previously :)

I bliv that one was io_uring one. haha. isnt it?

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
sammshen: @DongDongJu

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
sammshen: @DongDongJu

sammshen:
Screenshot 2026-03-09 at 2 10 22 PM
do these not show up?

DongDongJu:
> Screenshot 2026-03-09 at 2 10 22 PM do these not show up?

Thanks for the review! It seems pended. Could you click submit the review?

sammshen (changes requested):
LGTM!

sammshen:
my bad, I forgot to click "submit review", thx for the reminder!

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
DongDongJu: You are right. Thanks! I added the fix for this in here https://github.com/LMCache/LMCache/pull/2614/commits/e76aa4697e19f3aff0b85546c6208cfe6806effe

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
DongDongJu: Thanks! I added _is_valid_checkpoint_entry func.

DongDongJu:
@sammshen @ApostaC PTAL!

DongDongJu:
It seems weird. wha is the latest patch touch cache engine? 
UT for cache_engine broke `tests/v1/test_cache_engine.py::test_force_store_wait - AssertionError: assert 9472 == 9984`.

sammshen (approved):
LGTM!

DongDongJu:
@deng451e Could you take a look by any chance?

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -743,52 +905,188 @@ def _save_manifest(self, path: str) -> None:
...
-            return
-        logger.info("RustRawBlockBackend: loading manifest from %s", path)
-        with open(path, "r", encoding="utf-8") as f:
-            data = json.load(f)
-        if not isinstance(data, dict) or data.get("version") != 1:
-            logger.warning("Ignoring incompatible rust_raw_block manifest")
-            return
+        return snapshot, dirty_total
+
+    def _write_checkpoint(self, payload: bytes, dirty_total_snapshot: int) -> bool:
```
ApostaC, line 910: Should we make this function being thread-safe? This will be called by _checkpoint_once, which will be called in both `_checkpoint_loop` and `close`

ApostaC (approved):
Some quick question, otherwise LGTM!

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
DongDongJu: Good catch. 
The intended flow here is single-writer: `close()` stops the periodic checkpoint thread and joins it before issuing the final `_checkpoint_once(force=True)`, so under the normal shutdown path `_checkpoint_loop` and `close()` should not overlap.

That said, relying on `join(timeout=5)` alone is a bit brittle, since it does not strictly guarantee the background thread has exited. 
I agree it’s safer to serialize the checkpoint path explicitly. I’ll tighten this by guarding the checkpoint flow itself (likely at `_checkpoint_once()`, not only `_write_checkpoint()`), so the snapshot/write/seq update stays single-writer even during shutdown.

Next one is target to multi io worker. I will dealing this sync issue in next series.
Thanks for the review!

PR #2623, opened 2026-02-19
Rename some of the variables in observability.py

yoo-kumaneko:
The variables in `observability.py` are poorly named. Some incorrect, some inaccurate. Let's improve it!

Also, I've done a minor improvement on `on_store_finished`, removing a redundant input parameter.
And I replaced the counter increment from the beginning of the operations (retrieve or store) to the finish of them. That aligns with how we define an interval. We only view requests that finish in an interval as requests in that interval.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -516,7 +516,6 @@ class LMCacheEngine:
 
         self.stats_monitor.on_store_finished(
             store_stats,
-            tot_token_num,
         )
         tot_time = store_stats.time_to_store()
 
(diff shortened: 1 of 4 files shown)


maobaolong:
Clipboard_Screenshot_1771638315

@yoo-kumaneko Thanks for this improve, but maybe you can care about the failed UT.

lmcache/v1/cache_engine.py
```diff
@@ -540,7 +540,6 @@ def store(
 
         self.stats_monitor.on_store_finished(
             store_stats,
-            tot_token_num,
```
maobaolong, line 543: Is there anything else other than `renaming`?  @yoo-kumaneko

yoo-kumaneko: Yes. I've removed `tot_token_num` from `on_store_finished`. Because we already have the number of tokens to store recorded at `on_store_request`. There is no need to record the same stat once again. (Note that we don't know how many tokens are successfully stored. We only know how many tokens we want to store, and this number is the same at the beginning and the finish.)

yoo-kumaneko: Let me also add this to the PR description. Thanks for bringing this up.

PR #2635, opened 2026-02-24
[6/N][Feat] This PR adds io_uring support to the Rust raw block backend.

ankit-sam:

This PR adds io_uring support to the Rust raw block backend. This PR uses io_uring's fixed buffer registration process for true zero copy support. Thus removing any user to kernel copy and vice versa.

Main changes:
- io_uring backend with dedicated worker thread for submission/completion loop (```--use-uring```)
- Implicit write batching to reduce io_uring_enter syscalls
- Synchonous Read path
- Use PagedTensorMemoryAllocator if io_uring is enabled and fixed buffer support is requested (```--fixed-buffer```)
- Update docs with io_uring section with data path diagram and benchmark summary tables
- Add reviewer-focused comments in Rust I/O code paths for maintainability

- No significant changes to the rust raw block plugin i.e. no async path, no multi io-workers. Major performance benefits observed because of batching and fixed buffers.
- Fixed buffer consumes time and there is a limit on number of iovecs that can be registered. These are added as notes in the README section of benchmark script. 

**If applicable**:

- this PR contains user facing changes - docs added, and updated the benchmark script for comparison.

**Performance comparison posix pwrite vs uring_write with fixed buffer**

chunk_size = 4 (64 KB write), concurrency=4
| num_ops=16384  | posix  | uring, fixed_buffer |
| ------------- | ------------- | ------------- |
| ops / sec  | 3144.8 | 11245.8 |
| Elapsed time (sec) | 5.21  | 1.45  |

chunk_size = 256 (4 MB write), concurrency=4
| num_ops=4096  | posix  | uring, fixed_buffer |
| ------------- | ------------- | ------------- |
| ops / sec  | 742 | 748.5 |
| Elapsed time (sec) | 5.52  | 5.47  |

- For small chunks (64KB write) io_uring improves **~3X** compared to the posix implementation
- For large chunks (4 MB write) io_uring performs similar compared to the posix implementation.

---

> [!NOTE]
> **High Risk**
> Adds a new io_uring-based async I/O path with fixed-buffer registration and modifies allocator selection/alignment; failures could cause data corruption, deadlocks, or resource leaks due to low-level concurrency and unsafe buffer handling across Python/Rust.
> 
> **Overview**
> Enables an optional **io_uring** fast path for `RustRawBlockBackend`, including fixed-buffer registration for *true zero-copy* and new batched read/write operations that run through a dedicated Rust worker thread.
> 
> Updates `LocalCPUBackend`/allocators to support io_uring by using paged pinned buffers (and alignment auto-selection) and exposes paged-buffer accessors for registration.
> 
> Extends the benchmark harness with `--use-uring`, configurable `--chunk-size`, and a new `rust_raw_block_read` mode (write+read with optional `--verify-integrity`), adds a vLLM offload example for the raw-block backend, and adds io_uring-focused tests plus updated docs/README guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bd363127436adaeb3df91a48686d5a4092f90d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/benchmarks/storage_backend_io/README.md
+++ b/benchmarks/storage_backend_io/README.md
@@ -1,11 +1,19 @@
 # Storage Backend I/O Benchmark
 
 This microbenchmark compares **LocalDiskBackend** vs **RustRawBlockBackend** under high write-concurrency.
+For **RustRawBlockBackend** this also supports write and read performance testing for **posix** and **io_uring** backends.
 
 ## What It Measures
 
+### Write Benchmark
 - Total time to submit and complete `num_ops` write (put) operations
 - Effective ops/sec under concurrent submission
+
+### Read Benchmark (`--backend rust_raw_block_read`)
+- Write phase: Time to write `num_ops` memory objects to the raw block device
+- Read phase: Time to read back all `num_ops` objects using concurrent batched blocking reads
+- Data integrity verification (optional with `--verify-integrity`)
+- Separate metrics for write and read performance
 
 ## Usage
 
@@ -30,7 +38,8 @@ python benchmarks/storage_backend_io/storage_backend_io_benchmark.py \
 - `--raw-odirect` should only be used with a real block device that supports O_DIRECT.
 - When `--local-disk-odirect` is enabled, the benchmark allocates **page-aligned** buffers to avoid EINVAL from O_DIRECT.
 - Local disk backend uses its internal worker pool; completion is tracked via callbacks.
-- Rust raw block benchmark uses a unique manifest path per run to avoid stale-index reuse between runs.
+- Read benchmark (`rust_raw_block_read`) performs a write phase followed by a read phase, measuring both separately.
+- Use `--verify-integrity` with read benchmark to ensure data correctness (compares read data with original written data).
 
 ## How To Compare On Real NVMe
 
@@ -94,6 +103,58 @@ print(f"rust vs local: {(rust / local - 1.0) * 100.0:+.2f}%")
 PY
 ```
 
+## Use io_uring on Real NVMe
+
+### Compare posix vs io_uring write and read performance
+
+```bash
+# Run posix write & read benchmark. To enable data integrity `--verify-integrity`
+python benchmarks/storage_backend_io/storage_backend_io_benchmark.py \
+  --num-ops 1024 \
+  --concurrency 4 \
+  --backend rust_raw_block_read \
+  --raw-device /dev/nvme1n1 \
+  --chunk-size 256 \
+  --alignement 4096 \
+  --raw-odirect \
+  --output-json /tmp/rust_raw_block_read_posix.json
+
+# Run io_uring write & read benchmark. To enable data integrity `--verify-integrity`
+python benchmarks/storage_backend_io/storage_backend_io_benchmark.py \
+  --num-ops 1024 \
+  --concurrency 4 \
+  --backend rust_raw_block_read \
+  --raw-device /dev/nvme1n1 \
+  --raw-odirect \
+  --chunk-size 256 \
+  --alignement 4096 \
+  --use-uring \
+  --output-json /tmp/rust_raw_block_read_uring.json
+
+# Compute comparison
+python - <<'PY'
+import json
+
+with open("/tmp/rust_raw_block_read_posix.json") as f:
+    posix = json.load(f)[0]
+with open("/tmp/rust_raw_block_read_uring.json") as f:
+    uring = json.load(f)[0]
+
+print(f"posix read ops/sec: {posix['read_ops_per_sec']:.2f}")
+print(f"uring read ops/sec: {uring['read_ops_per_sec']:.2f}")
+print(f"uring vs posix read: {(uring['read_ops_per_sec'] / posix['read_ops_per_sec'] - 1.0) * 100.0:+.2f}%")
+print(f"posix write ops/sec: {posix['write_ops_per_sec']:.2f}")
+print(f"uring write ops/sec: {uring['write_ops_per_sec']:.2f}")
+print(f"uring vs posix write: {(uring['write_ops_per_sec'] / posix['write_ops_per_sec'] - 1.0) * 100.0:+.2f}%")
+PY
+```
+### Notes
+
+- There is a limit on the number of fixed buffers that can be registered. For unprivileged users its 16384.
+- Buffer registration and de-registration is time consuming.
+
 ## Output
 
 The script prints a summary and optionally writes JSON results if `--output-json` is provided.
+```
+
--- a/benchmarks/storage_backend_io/storage_backend_io_benchmark.py
+++ b/benchmarks/storage_backend_io/storage_backend_io_benchmark.py
@@ -331,9 +331,10 @@ def _bench_rust_raw_block(
     keys = _make_keys(num_ops)
     # Use the memory allocator from LocalCPUBackend
     shapes = metadata.get_shapes()
+    use_aligned = use_odirect and not use_uring
     objs = _make_memory_objs(
         num_ops,
-        False,
+        use_aligned,
         alignment,
         [],
         memory_allocator=local_cpu.memory_allocator,
@@ -506,9 +507,10 @@ def _bench_rust_raw_block_read(
 
     keys = _make_keys(num_ops)
     shapes = metadata.get_shapes()
+    use_aligned = use_odirect and not use_uring
     objs = _make_memory_objs(
         num_ops,
-        False,
+        use_aligned,
         alignment,
         [],
         memory_allocator=local_cpu.memory_allocator,
--- a/benchmarks/storage_backend_io/storage_backend_io_benchmark.py
+++ b/benchmarks/storage_backend_io/storage_backend_io_benchmark.py
@@ -24,7 +24,9 @@ from lmcache.utils import CacheEngineKey
 from lmcache.v1.config import LMCacheEngineConfig
 from lmcache.v1.memory_management import (
     AdHocMemoryAllocator,
+    MemoryAllocatorInterface,
     MemoryFormat,
+    MemoryObj,
     MemoryObjMetadata,
     TensorMemoryObj,
 )
@@ -37,6 +39,7 @@ from lmcache.v1.storage_backend.plugins.rust_raw_block_backend import (
 
 DEFAULT_SHAPE = torch.Size([2, 16, 8, 128])
 DEFAULT_DTYPE = torch.bfloat16
+DEFAULT_CHUNK_SIZE = 4
 
 
 def _start_loop() -> tuple[asyncio.AbstractEventLoop, threading.Thread]:
@@ -60,7 +63,8 @@ def _build_metadata() -> LMCacheMetadata:
         worker_id=0,
         local_worker_id=0,
         kv_dtype=DEFAULT_DTYPE,
-        kv_shape=(4, 2, 256, 8, 128),
+        kv_shape=(4, 2, DEFAULT_CHUNK_SIZE, 8, 128),
+        chunk_size=DEFAULT_CHUNK_SIZE,
     )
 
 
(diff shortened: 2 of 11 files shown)


lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -612,20 +793,36 @@ def get_blocking(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
-                read_payload_len = (
-                    total_len if len(direct_view) >= total_len else payload_len
-                )
-                self._rawdev().pread_into(
-                    entry.offset + self.header_bytes,
-                    direct_view,
-                    read_payload_len,
-                    total_len,
-                )
+            if self.use_uring:
```
DongDongJu: Maybe we can moving read_payload_len definition before the `if self.use_uring` condition?

rust/raw_block/src/lib.rs
```diff
@@ -254,20 +352,464 @@ impl RawBlockDevice {
...
+            self.enqueue_write(offsets[i], buffer_ptrs[i] as *const u8, total_lens[i])?;
+        }
+
+        Ok(())
+    }
+
+    /// Flush: wait for all in-flight I/O to complete.
+    /// The method waits on a condition variable that gets signaled
+    /// when in_flight_count reaches 0.
+    fn flush_iouring(&self) -> PyResult<()> {
```
DongDongJu: it seems only waits only while `in_flight_count >0`.
So there is a race where queued but not yet submitted writes are invisible to flush_iouring(), which can return immediately. 
After that, Python promotes keys from _inflight to _index, call callbacks, and ref_count_down()s the payload objects as though the batch were durable.
What do you think? Please correct me if im wrong.

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -468,6 +520,135 @@ def batched_submit_put_task(
             futures.append(fut)
         return futures or None
 
+    def _batched_submit_put_task_uring(
```
DongDongJu: Is there any case for partial completed?

DongDongJu (commented):
Hello @ankit-sam,
I left few questions. I will test it by myself and leave the review again.
Please add the test that 
1. submit is failed,
2. close with queued/in-flight I/O case, and
3.  unaligned payload size.

sammshen:
why is small POSIX faster than large POSIX?

sammshen:
can you add an example to the `examples/kv_cache_reuse/` folder?

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
ankit-sam: Ok

rust/raw_block/src/lib.rs
ankit-sam: Yes you are right about the in_flight_count > 0 check, gemini gave the same comment. I moved it to enque_read and write.

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
ankit-sam: For io_uring case usually block layer splits and returns only when the entire I/O is completed so no issues with partial I/O

ankit-sam:
> why is small POSIX faster than large POSIX?

No small POSIX is not faster in comparison to large POSIX, in terms of actual throughput. If you see for small POSIX 16384 ops of size 64K (chunk size = 4) gets completed in 5.21 seconds. That's 196.5 MiB/s
For large POSIX 4096 ops of size 4M (chunk size = 256) gets completed in 5.52 seconds, which is 2968 MiB/s

For io_uring case with smaller chunk size 4 (64 K writes), we see 3-4X improvement compared to POSIX, as it is able to do batched asynchronous submissions. But for larger chunk sizes 256 (4 M writes), both io_uring and POSIX hit the storage / NVMe bandwidth limit for a single write request, and even with batching it can't scale any further,

ankit-sam:
> can you add an example to the `examples/kv_cache_reuse/` folder?

Sure, will add it.

DongDongJu (commented):
@ankit-sam Thanks for addressing the reviews.
I have general question, what happened if write was failed in batched requests?
It seems currently only checking the # of I/O.
Otherwise, LGTM.

benchmarks/storage_backend_io/README.md
```diff
@@ -94,6 +96,36 @@ print(f"rust vs local: {(rust / local - 1.0) * 100.0:+.2f}%")
...
 ```
 
+## How To use io_uring on Real NVMe
+
+Use the rust_raw_block on the raw block device (unmounted)
+
+Example parameters:
+- `num_ops=16384`
+- `concurrency=4`
+- `--raw-odirect`
```
sammshen: @ankit-sam  dumb question: is there any benefit to not using odirect, especially with io_uring

DongDongJu (approved):
LGTM as discussed offline.

benchmarks/storage_backend_io/README.md
ankit-sam: Without o-direct i.e. in buffered mode, data is first copied from user buffer into kernel page cache, and later written to the disk in background. So usually it will be beneficial for smaller I/Os as kernel will coalesce them and write larger chunks.

ankit-sam:
> @ankit-sam Thanks for addressing the reviews. I have general question, what happened if write was failed in batched requests? It seems currently only checking the # of I/O. Otherwise, LGTM.

Yes thanks for pointing it, I missed it initially, now fixed it. I also made some changes to the rust backend, with python GIL release, which was not there initially.

sammshen:
@cursor review

DongDongJu:
@ankit-sam Could you resolve the conflict?

sammshen (approved):
LGTM

examples/kv_cache_reuse/local_backends/rust_backend_offload.py
```diff
@@ -0,0 +1,132 @@
...
+from vllm import LLM, SamplingParams
+from vllm.config import KVTransferConfig
+from vllm.engine.arg_utils import EngineArgs
+
+# First Party
+from lmcache.integration.vllm.utils import ENGINE_NAME
+from lmcache.v1.cache_engine import LMCacheEngineBuilder
+
+
+def setup_environment_variables(raw_block_path: str, use_uring: bool = False):
```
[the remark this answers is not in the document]
DongDongJu: @ankit-sam Could you resolve this comment?

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -589,6 +647,154 @@ def batched_submit_put_task(
...
+                header_bytes = bytearray(header)
+                if self.use_odirect and len(header_bytes) < hdr_total:
+                    header_bytes.extend(b"\x00" * (hdr_total - len(header_bytes)))
+                offsets.append(offset)
+                buffers.append(header_bytes)
+                total_lens.append(hdr_total)
+
+                # Prepare payload with proper O_DIRECT alignment and zero-copy handling.
+                try:
+                    buf, payload_len, total_len = self._prepare_write_payload(obj)
```
DongDongJu: Here batched io_uring write path drops payload_len. 
Python computes (buf, payload_len, total_len), but this one only accepts total_lens and then copies/submits total_len bytes from the source pointer. This can over read the source buffer whenever total_len > payload_len. 
Please either add payload_lens to batched_write() or guarantee every submitted payload buffer is already padded to total_len.

rust/raw_block/src/lib.rs
```diff
@@ -254,20 +405,1096 @@ impl RawBlockDevice {
...
+    }
+
+    /// Batched write: submit multiple writes at once via io_uring.
+    /// All writes are queued to the worker thread, which processes them
+    /// in batches to maximize throughput.
+    ///
+    /// Returns a batch_id that must be passed to wait_iouring() to wait
+    /// for completions for that batch.
+    #[pyo3(signature = (offsets, buffers, total_lens))]
+    fn batched_write(
```
DongDongJu, line 943: I’m worried about the partial-failure behavior here. 
This method can queue part of a batch and then error before returning batch_id, which means the caller cannot wait on the already-submitted prefix. Python then goes through the failure cleanup path and frees slots. 
Please either make submission atomic or make partial submission cleanup deterministic.

DongDongJu, line 943: Also entry points don’t check closed, while the sync entry points do. 
A concurrent/post-close batch submission can still enter while shutdown is in progress or after the worker is gone. 
Please add the same closed/shutdown guard here that read_uring() / pwrite_from_buffer() already use.

DongDongJu (changes requested):
It was quite changed a lot after I last time reviewed. Please check the comments and revise it.
IMO, this is too big changes as a single PR, Lets minimize in next one.

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
ankit-sam: Actually I should have mentioned in a comment, and commit log. For io_uring it will always use aligned paged_buffers. So in that case payload_len should always be same as total_len. I will add comments, assert for this.

rust/raw_block/src/lib.rs
ankit-sam: Thanks for pointing it out, will add that

ankit-sam: Ah, I see the issue. I will do the I/O request preparation part first and then submit all of them at once.

ankit-sam:
> It was quite changed a lot after I last time reviewed. Please check the comments and revise it. IMO, this is too big changes as a single PR, Lets minimize in next one.

Yeah, I should have anticipated that the changes would be large. I split the first commit into 2.
First with changes to allocator init which uses PagedTensorMemoryAllocator for io_uring. It also makes sure to use alignment in this case.
Second commit handles all the rust side io_uring changes and the python plugin. This change is still pretty big

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
```diff
@@ -24,7 +24,9 @@
 from lmcache.v1.config import LMCacheEngineConfig
 from lmcache.v1.memory_management import (
     AdHocMemoryAllocator,
+    MemoryAllocatorInterface,
```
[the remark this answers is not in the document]
ankit-sam: benchmarks/storage_backend_io/storage_backend_io_benchmark.py already has # SPDX-License-Identifier: Apache-2.0 on the first line

DongDongJu:
@ankit-sam Could you check ut? and bot's request?

rust/raw_block/src/lib.rs
```diff
@@ -254,20 +405,1129 @@ impl RawBlockDevice {
...
+
+        // Extract pointers as usize before releasing GIL (raw pointers are not Send)
+        let mut ptrs = Vec::with_capacity(n);
+        for view in &views {
+            ptrs.push(view.buf as usize);
+        }
+
+        for view in views {
+            release_pybuffer(view);
+        }
```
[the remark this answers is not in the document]
ankit-sam: The underlying memory objects won't be relallocated as they are
1. registered with io_uring for fixed buffers,
2. the reference count has been increased for them, which won't be decreased till the I/O gets completed

DongDongJu (approved):
Its too old thread now. For iouring_cmd support, Let me refactor/clean up after merging this.

DongDongJu:
@ankit-sam Could you add the dependency requirements in readme? eg., io_uring lib, kernel version, Kernel configs..

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
```diff
@@ -285,8 +329,16 @@ def _bench_rust_raw_block(
...
+    # Use the memory allocator from LocalCPUBackend
+    shapes = metadata.get_shapes()
+    objs = _make_memory_objs(
+        num_ops,
+        False,
+        alignment,
+        [],
+        memory_allocator=local_cpu.memory_allocator,
+        shapes=shapes,
+    )
```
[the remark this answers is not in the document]
DongDongJu: Hello @ankit-sam , Could you check this issue?

Issue #2632, opened 2026-02-24
[Refactor]: Southbound


sammshen:
Proposed PR for this RFC: #2642 

Many "native" (non python backends are appearing): 

Redis: https://github.com/LMCache/LMCache/pull/2541
3FS: https://github.com/LMCache/LMCache/issues/2437
Disk: https://github.com/LMCache/LMCache/issues/2394

Based on https://github.com/LMCache/LMCache/pull/2607 and following the empirically tested strategies in https://github.com/LMCache/LMCache/pull/2541, plan to create an interface for implementing native backends that will include: 
1. GIL release best practice
2. eventfd based non-polling and coalesced completions
3. non-blocking submission
4. batched only operations
5. data structures for tiling a single batched request across threads

Estimated Time of Delivery: 1-1.5 weeks

cc @DongDongJu @maobaolong

DongDongJu:
FYI: https://github.com/LMCache/LMCache/issues/2437 is just need to call python binding from 3fs side. dosent need to make native code support.

DongDongJu:
Hello @sammshen, 
Sounds great! If you plan to make the same behavior in c_ops side, it will more easy to avoid the same efforts.
Then I will merge the rust side ops to c_ops side when its done after comparing the result.
what do you think?

maobaolong:
@sammshen hope fs connector could be considered to be supported to check whether the abstraction is enough

PR #2642, opened 2026-02-26
[Southbound]: Create a Native Protocol for MP and non-MP

sammshen:
Implementation of https://github.com/LMCache/LMCache/issues/2632 as a generalization of the strategies employed in https://github.com/LMCache/LMCache/pull/2541


DongDongJu:
Impressive! I will review in this night. Thanks @sammshen

sammshen:
try to unify with MP mode: https://github.com/LMCache/LMCache/pull/2569

csrc/storage_backends/connector_interface.h
```diff
@@ -0,0 +1,112 @@
...
+#include <vector>
+
+namespace lmcache {
+namespace connector {
+
+/*
+instantiated in connector_base.h and further overridden by custom storage
+connectors
+*/
+class IStorageConnector {
```
DongDongJu, line 16: What is the I meaning in this class name?

csrc/storage_backends/connector_pybind_utils.h
```diff
@@ -0,0 +1,148 @@
...
+    std::vector<void*> bufs;
+    std::vector<size_t> lens;
+    bufs.reserve(keys.size());
+    lens.reserve(keys.size());
+
+    for (size_t i = 0; i < keys.size(); ++i) {
+      py::memoryview mv = memviews[i].cast<py::memoryview>();
+      py::buffer_info info = py::buffer(mv).request();
+      bufs.push_back(info.ptr);
+      lens.push_back(static_cast<size_t>(info.size));
```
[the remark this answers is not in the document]
DongDongJu: I agree with this opinion. Good to have for future debugging.
Since, `info.size` is number of elements, not bytes.

lmcache/v1/storage_backend/native_clients/connector_client_base.py
```diff
@@ -1,76 +1,34 @@
...
         self.loop = loop or asyncio.get_running_loop()
-        self._client = LMCacheRedisClient(host, port, num_workers, username, password)
-        self._fd = int(self._client.event_fd())
+        self._client: NativeClientT = native_client
+        self._fd = int(self._client.event_fd())  # type: ignore[attr-defined]
         self._closed = False
-
-        # future_id -> (Future, op_name)
-        # we support both types of futures since we only their basic interface
         self._pending: Dict[
```
DongDongJu, line 22: It seems does not hold references to the submitted memoryviews or their backing objects.
when a task awaiting a Future is cancelled, the awaited Future can be cancelled too, and the coroutine can unwind, dropping the last Python references while C++ worker threads still hold raw pointers.

Can we store keepalive refs per future_id, e.g. _pending[fid] = (fut, op, tuple(bufs))?
And we should only drop those refs after completion is handled or native shutdown is fully done.

csrc/storage_backends/redis/connector.cpp
```diff
@@ -0,0 +1,351 @@
...
+      port_(port),
+      username_(std::move(username)),
+      password_(std::move(password)) {
+  worker_fds_.reserve(num_workers);
+  start_workers();  // start after derived class is fully constructed
+}
+
+RedisConnector::~RedisConnector() { close(); }
+
+WorkerConn RedisConnector::create_connection() {
```
DongDongJu, line 231: what happened if this func was failed?

DongDongJu (commented):
@sammshen Great work. it seems you made the gem.
I left few comments.

maobaolong (commented):
@sammshen Could you also add a pure c++ implemented fs connector?

lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
```diff
@@ -0,0 +1,297 @@
+# SPDX-License-Identifier: Apache-2.0
```
sammshen, line 1: @ApostaC the new adapter for native MP mode adapters

sammshen:
Thanks for review @maobaolong. I think this should be a follow up PR

csrc/storage_backends/redis/connector.cpp
sammshen: good point, right now we just stay dead but we can add reconnection logic

csrc/storage_backends/connector_interface.h
sammshen: InterfaceStorageConnector, sorry if it is unclear

lmcache/v1/storage_backend/native_clients/connector_client_base.py
sammshen: great point, added extra lifetime to the pending operations that will be freed either on failure or on completion

lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def query_load_result(self, task_id: L2TaskId) -> Bitmap | None:
+        with self._lock:
+            return self._completed_loads.pop(task_id, None)
+
+    # -------------------------------------------------------------------
+    # Cleanup
+    # -------------------------------------------------------------------
+
+    def close(self) -> None:
```
maobaolong, line 252:
```python
    if self._closed:
        return
    self._closed = True
```

Suggest to add the closed flag to avoid double close.

lmcache/v1/distributed/l2_adapters/config.py
```diff
@@ -159,6 +159,72 @@ def help(cls) -> str:
...
+            "RESP L2 adapter config fields:\n"
+            "- host (str): Redis/Valkey server hostname or IP (required)\n"
+            "- port (int): server port (required, >0)\n"
+            "- num_workers (int): C++ worker threads for I/O (default 8, >0)\n"
+            "- username (str): auth username (default empty)\n"
+            "- password (str): auth password (default empty)"
+        )
+
+
+register_l2_adapter_type("resp", RESPL2AdapterConfig)
```
maobaolong: https://github.com/LMCache/LMCache/pull/2704

I submitted a PR to do a refactor, after that, add a new coming l2_adapter, we never need to change any existing file.

maobaolong (commented):
Left a minor nit, otherwise, this PR looks good.

ApostaC (approved):
Please make the fixes following our offline discussions. Give an approve here.

deng451e (approved):

maobaolong (approved):
LGTM @sammshen Thanks for this great feature, will extend some fs native adapter base on your effort!

PR #2647, opened 2026-02-27
Add Mooncake dummy client mode for MooncakeStore

duhaode520:
## Summary

This PR adds support for Mooncake's dummy client mode in lmcache, enabling a dual-process architecture where RDMA transfers run in a separate real client process, free from Python GIL contention. In dummy mode, lmcache's internal pinned memory is replaced with Mooncake-managed shared memory, allowing both GPU DMA and cross-process RDMA to operate on the same buffer without extra copies.

## Motivation

Mooncake provides a dummy client mode (`setup_dummy`) with a dual-process architecture:
- **Dummy client** (embedded in the vLLM worker): communicates via shared memory + IPC
- **Real client** (separate process): performs actual RDMA/TCP transfers, free from Python GIL

This avoids GIL blocking on RDMA operations, allows the real client to be independently restarted for better availability, and eliminates the need to launch a heavy real client process per vLLM worker in tensor-parallel setups — multiple dummy clients can share a single real client process.

### Configuration

Add `use_dummy_client` and `dummy_server_address` fields to `MooncakeStoreConfig`:

```yaml
extra_config:
  use_dummy_client: true
  dummy_server_address: "127.0.0.1:50052"
```

All three config loading paths (JSON file, lmcache extra_config, environment variables) support the new fields. Default is `False` for full backward compatibility.

### Initialization

When `use_dummy_client=True`, the connector calls `store.setup_dummy()` instead of `store.setup()`, and skips NUMA binding (RDMA runs in the real client process). The init logic is refactored into `_setup_dummy_client()` and `_setup_real_client()` for clarity.

### Shared memory integration

Dummy client mode requires `register_buffer` to accept a pointer from Mooncake's shared memory pool (`alloc_from_mem_pool`). Since lmcache's original pinned memory (`cudaHostAlloc`) is not in Mooncake's `ShmHelper::shms_` registry, the connector replaces lmcache's internal CPU buffer with Mooncake shared memory during initialization.

This is implemented across three layers with strict dependency direction (`Connector → MixedMemoryAllocator → lmc_ops → C++`):

**C++ layer** — New `register_pinned_ptr` / `unregister_pinned_ptr` functions wrap `cudaHostRegister` / `cudaHostUnregister` to pin externally allocated memory for GPU DMA. HIP compatible via auto-mapping. Non-CUDA no-op fallbacks provided.

**Memory layer** — New `MixedMemoryAllocator.replace_buffer(external_ptr, size)` method encapsulates all low-level details: frees the original buffer, pins the external memory, wraps it as `torch.Tensor`, rebuilds `pin_allocator`, and registers a cleanup callback so `close()` calls `unregister_pinned_ptr` instead of the default free path.

**Connector layer** — In `_setup_dummy_client()`, after `store.setup_dummy`:
1. `store.alloc_from_mem_pool(buffer_size)` → shared memory pointer
2. `allocator.replace_buffer(shm_ptr, buffer_size)` → memory layer handles the rest
3. `store.register_buffer(ptr, size)` → succeeds because the pointer is in Mooncake's shm registry

### Data flow

```
GPU → cudaMemcpy (DMA) → Mooncake SHM (cudaHostRegistered as pinned)
                           ↑ lmcache pin_allocator allocates from this memory
                           ↓ register_buffer succeeds (pointer found in shms_)
                       Real Client process (zero-copy RDMA)
```

## Performance

Tested with DeepSeek V3.2 (MLA) under TP=8 configuration with a representative workload. No noticeable performance degradation compared to the original real client mode.

`cudaHostRegister(flags=0)` and `cudaHostAlloc(flags=0)` produce equivalent DMA performance — both page-lock the memory via the same PCIe DMA channel. Mooncake's `mmap(MAP_POPULATE)` pre-allocates physical pages, covering the same role as lmcache's `first_touch`. The only theoretical gap is NUMA affinity (Mooncake SHM does not call `mbind`), but this only matters when `numa_mode` is explicitly configured (defaults to `None`).

### Real client mode (baseline)

Clipboard_Screenshot_1772190642

### Dummy client mode

Clipboard_Screenshot_1772190661

## Changed files

| File | Change |
|------|--------|
| `csrc/mem_alloc.h` | Add `register_pinned_ptr` / `unregister_pinned_ptr` declarations |
| `csrc/mem_alloc.cpp` | Add implementations wrapping `cudaHostRegister` / `cudaHostUnregister` |
| `csrc/pybind.cpp` | Add pybind11 bindings with GIL release |
| `lmcache/non_cuda_equivalents.py` | Add non-CUDA no-op fallbacks |
| `lmcache/v1/memory_management.py` | Add `replace_buffer()` method; modify `close()` to support cleanup callback |
| `mooncakestore_connector.py` | Add dummy client init branch; shm allocation + buffer replacement; refactor into `_from_dict` / `_setup_dummy_client` / `_setup_real_client` |
| `tests/v1/test_memory_management.py` | Add `TestMixedMemoryAllocatorReplaceBuffer` (7 tests) |
| `tests/v1/storage_backend/test_mooncakestore_connector.py` | Add dummy client + shm buffer replacement tests (21 tests total) |

## Backward compatibility

- Config files without `use_dummy_client` load normally (defaults to `False`)
- Setting `use_dummy_client=True` with a Mooncake version that lacks `setup_dummy` raises a clear error
- All existing data operations (`batched_get`, `batched_put`, `exists`) are unchanged

## Testing

- 21 connector tests + 7 memory management tests all pass
- Pre-commit checks (isort, ruff, ruff-format, mypy, clang-format) all pass

### DIFF
--- a/csrc/mem_alloc.cpp
+++ b/csrc/mem_alloc.cpp
@@ -145,3 +145,19 @@ void free_shm_pinned_ptr(uintptr_t ptr, size_t size,
   }
   shm_unlink(shm_name.c_str());
 }
+
+void register_pinned_ptr(uintptr_t ptr, size_t size) {
+  cudaError_t st = cudaHostRegister(reinterpret_cast<void*>(ptr), size, 0);
+  if (st != cudaSuccess) {
+    throw std::runtime_error(std::string("cudaHostRegister failed: ") +
+                             cudaGetErrorString(st));
+  }
+}
+
+void unregister_pinned_ptr(uintptr_t ptr, size_t size) {
+  cudaError_t st = cudaHostUnregister(reinterpret_cast<void*>(ptr));
+  if (st != cudaSuccess) {
+    throw std::runtime_error(std::string("cudaHostUnregister failed: ") +
+                             cudaGetErrorString(st));
+  }
+}
--- a/csrc/mem_alloc.h
+++ b/csrc/mem_alloc.h
@@ -11,3 +11,6 @@ void free_numa_ptr(uintptr_t ptr, size_t size);
 void free_pinned_numa_ptr(uintptr_t ptr, size_t size);
 void free_shm_pinned_ptr(uintptr_t ptr, size_t size,
                          const std::string& shm_name);
+
+void register_pinned_ptr(uintptr_t ptr, size_t size);
+void unregister_pinned_ptr(uintptr_t ptr, size_t size);
--- a/csrc/pybind.cpp
+++ b/csrc/pybind.cpp
@@ -49,6 +49,10 @@ PYBIND11_MODULE(c_ops, m) {
   m.def("alloc_pinned_numa_ptr", &alloc_pinned_numa_ptr,
         py::call_guard<py::gil_scoped_release>());
   m.def("free_pinned_numa_ptr", &free_pinned_numa_ptr);
+  m.def("register_pinned_ptr", &register_pinned_ptr,
+        py::call_guard<py::gil_scoped_release>());
+  m.def("unregister_pinned_ptr", &unregister_pinned_ptr,
+        py::call_guard<py::gil_scoped_release>());
   m.def("alloc_numa_ptr", &alloc_numa_ptr,
         py::call_guard<py::gil_scoped_release>());
   m.def("free_numa_ptr", &free_numa_ptr);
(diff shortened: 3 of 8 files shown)


duhaode520:
@maobaolong, could you help review this? And help to add other reviewers that you think proper.

PR #2653, opened 2026-02-27
[UX] Log full stack trace upon exception in MP mode

KuntaiDu:

Currently LMCache MP mode will not log full stack trace. This PR fixes that.

**If applicable**:


ApostaC (commented):
The fix looks good to me.
We also want to use traceback and use `TracebackException` to print out the  stack trace when an exception happens (https://docs.python.org/3/library/traceback.html#traceback.TracebackException)

royyhuang (approved):
LGTM!

KuntaiDu:
@ApostaC We do not need to put the exception inside `logger.exception`. Also, I have did a quick test by manually injected a `NotImplementedError` and the stack trace looks good.

image

PR #2656, opened 2026-02-28
[feat] add free_locks api to MP mode

royyhuang:
Add free_locks API to MP mode. This is needed to explicitly free the locked blocks for vLLM hit tokens which will never be released by vLLM since these will never be read by vLLM.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -211,6 +211,39 @@ class LMCacheMPSchedulerAdapter:
         """
         self.lookup_futures.pop(request_id, None)
 
+    def free_locks(
+        self,
+        token_ids: list[int],
+        start: int,
+        end: int,
+        request_id: str,
+    ) -> None:
+        """Release read locks acquired during lookup without a full retrieve.
+
+        Use this when a request is cancelled or aborted after lookup but
+        before retrieve to avoid holding read locks until TTL expiry.
+
+        When ``start`` or ``end`` is not aligned to the chunk size, the
+        entire chunk containing that boundary is freed.  If the caller
+        needs to preserve locks on part of a boundary chunk, it should
+        align ``start`` / ``end`` to chunk boundaries before calling
+        this method.
+
+        Args:
+            token_ids: Token IDs for the key (same as used in lookup).
+            start: Start token index.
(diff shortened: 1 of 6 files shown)


lmcache/v1/multiprocess/server.py
```diff
@@ -448,6 +448,39 @@ def lookup(
         found_count = found_count // ipc_keys[0].world_size
         return found_count
 
+    def free_locks(
```
ApostaC: Can we call it "free_lookup_locks" to have more clarity?

lmcache/v1/multiprocess/server.py
```diff
@@ -448,6 +448,39 @@ def lookup(
...
+        self,
+        keys: list[IPCCacheEngineKey],
```
ApostaC: Do we want to take in a list of keys instead of a single key here? It can probably simplify the code a bit

ApostaC: I saw in the multiprocess adapter, we only have one key at a time.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -211,6 +211,39 @@ def cleanup_lookup_result(self, request_id: str) -> None:
...
+        self,
+        token_ids: list[int],
+        start: int,
+        end: int,
+        request_id: str,
+    ) -> None:
+        """Release read locks acquired during lookup without a full retrieve.
+
+        Use this when a request is cancelled or aborted after lookup but
+        before retrieve to avoid holding read locks until TTL expiry.
```
ApostaC: Another use case of this function: when vLLM holds the prefix, we want to unlock the vLLM's prefix.

tests/v1/multiprocess/test_free_locks.py
ApostaC, line 1: We can actually put this in the `tests/v1/multiprocess/test_cache_server.py` since what we did here is adding a new method to the cache server.

ApostaC (commented):
Let's catchup offline quickly to talk about the design

royyhuang:
@ApostaC Comments addressed PTAL.

ApostaC (approved):
LGTM!

Oasis-Git (approved):

PR #2646, opened 2026-02-27
[3/4][MP] L2 Store controller for MP mode

ApostaC:
**What this PR does / why we need it**:
Part of #2562 

It supports storing to the L2 adapter

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [x] this PR contains unit tests

PR #2658, opened 2026-02-28
[4/4][MP] L2 Prefetch controller foundation

ApostaC:
Part of #2562

This is the foundation PR for the L2 prefetch controller. It adds the building blocks needed before implementing the full controller (next PR):

1. **`L1Manager.finish_write_and_reserve_read()`** — Atomic transition from write-locked to read-locked state. After loading L2 data into write-reserved L1 buffers, this prevents a race window where eviction could remove the key between separate `finish_write()` + `reserve_read()` calls.

2. **`PrefetchPolicy` ABC + `DefaultPrefetchPolicy`** — Policy interface for deciding which L2 adapter loads which key when multiple adapters have the same data. The default policy assigns each key to the lowest-indexed adapter that has it.

3. **`PrefetchHandle` redesign** — Extended from a single `prefix_hit_chunks` field to include `request_id` (for async L2 tracking), `l1_prefix_hit_count`, and `total_requested_keys`.

4. **More interfaces in `Bitmap`** — Expose more interfaces in Bitmap for the prefetch controller to use.

**Note:** This PR needs to be rebased after #2646 is merged. And it requires #2662 to be merged first.

**If applicable**:

### DIFF
--- a/tests/v1/distributed/test_l1_manager.py
+++ b/tests/v1/distributed/test_l1_manager.py
@@ -877,6 +877,167 @@ class TestFinishWrite:
 
 
 # =============================================================================
+# Tests for L1Manager.finish_write_and_reserve_read()
+# =============================================================================
+
+
+class TestFinishWriteAndReserveRead:
+    """
+    Tests for L1Manager.finish_write_and_reserve_read() method.
+
+    This method atomically finishes write and acquires read lock,
+    preventing a race window where eviction could interfere.
+
+    Per the docstring:
+    - KEY_NOT_EXIST: The key does not exist.
+    - KEY_IN_WRONG_STATE: Not write-locked, or already read-locked.
+    - SUCCESS: Write unlocked and read lock acquired atomically.
+    """
+
+    def test_normal_transition(self, basic_l1_config, basic_layout):
+        """Test normal write-locked -> read-locked transition."""
+        manager = L1Manager(basic_l1_config)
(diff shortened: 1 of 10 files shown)


tests/v1/distributed/test_l1_manager.py
```diff
@@ -876,6 +876,167 @@ def test_finish_write_non_write_locked_returns_wrong_state(
...
+    """
+
+    def test_normal_transition(self, basic_l1_config, basic_layout):
+        """Test normal write-locked -> read-locked transition."""
+        manager = L1Manager(basic_l1_config)
+        key = make_object_key(12345)
+
+        # Reserve write (key is now write-locked)
+        write_result = manager.reserve_write([key], [False], basic_layout)
+        assert write_result[key][0] == L1Error.SUCCESS
```
KuntaiDu, line 904: Naming-wise `L1Error.SUCCESS` sounds weird. L1 Error is an error and it should not be with `SUCCESS`. I'm fine to not change it in this PR tho.

KuntaiDu (approved):
LGTM

tests/v1/distributed/test_l1_manager.py
ApostaC: Here I followed the naming convention of CUDA errors -- i.e., `cudaSuccess` is the first value of `cudaError_t` 
image

YaoJiayi (approved):

PR #2664, opened 2026-03-02
[1/N] Support NIXL-based L2 storage in MP mode

YaoJiayi:
Initial support of Nixl-based L2 adapter.

TODOs:

- [ ] Eviction is not handled
- [ ] E2E testing
- [ ] E2E testing with CB
- [ ] Unit test for different nixl backends (currently only have tests for POSIX backend)
- [ ] Documentation on how to enable l2 storage with NIXL in mp mode
- [ ] KV cache persistence

**If applicable**:


lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+    The object stored in Nixl L2 cache.
+    Can be used for both file and object.
+    """
+
+    page_index: int
+
+    size: int  # in bytes
+
+    shape: Optional[torch.Size] = None
+    dtype: Optional[torch.dtype] = None
```
ApostaC: Can we use layout_desc (`MemoryLayoutDesc`), which was introduced for MP mode here?

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+
+    fmt: Optional[MemoryFormat] = None
```
ApostaC: In the future, we assume all the LMCache-managed object will have the same fmt (2LTD) so fmt is not needed.

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -55,6 +55,23 @@ class L2AdapterInterface(ABC):
     and prefetch controller), therefore, it needs to be thread-safe.
     """
 
+    #####################
+    # Memory Registration Interface
+    #####################
+    def lazy_init_memory(self, **kwargs) -> None:  # noqa: B027
```
ApostaC: Let's chat and see if we can get rid of this top-level new API

ApostaC: By design, the L1 memory manager should already be initialized before creating L2 managers. So we should be able to get enough information there

ApostaC: Another big problem in this API is that in MP mode, LMCache will connect to different engines with different models, and the shape/dtype will be different.

lmcache/v1/distributed/memory_manager.py
```diff
@@ -30,9 +32,17 @@ def create_memory_allocator(config: L1MemoryManagerConfig) -> MemoryAllocatorInt
         MemoryAllocatorInterface: An instance of a memory allocator.
     """
     if config.use_lazy:
+        assert not config.use_page, (
+            "Page mem allocator is not supported with lazy initialization."
+        )
```
ApostaC: Probably we don't need to use the paged memory allocator. The current memory allocator guarantees the memory object's virtual address alignment, and therefore we may pre-register with the page size = alignment size.

We can discuss this offline

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+        )
+
+        self.pool = NixlObjPool(num_total_objs=self.pool_size)
+        if self.backend in ["GDS", "GDS_MT", "POSIX", "HF3FS"]:
+            assert "file_path" in self.backend_params, (
+                f"file_path is required for backend {self.backend}"
+            )
+            assert "use_direct_io" in self.backend_params, (
+                f"use_direct_io is required for backend {self.backend}"
+            )
```
ApostaC: Probably a better place to put these is in the config class.

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+            else:
+                logger.warning(
+                    "use_direct_io is True, but O_DIRECT is not available on "
+                    "this system. Falling back to buffered I/O."
+                )
+        for i in range(num_pages):
+            filename = f"obj_{i}_{uuid.uuid4().hex[0:4]}.bin"
+            tmp_path = os.path.join(file_path, filename)
+            fd = os.open(tmp_path, flags)
+            fds.append(fd)
```
ApostaC, line 223: The IBM storage team complained about this this morning. Let's see how we can improve this.

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+        """Signal the lookup event fd to notify completion."""
+        os.eventfd_write(self._lookup_efd, 1)
+
+    def _execute_lookup_in_the_loop(
+        self, keys: list[ObjectKey], task_id: L2TaskId
+    ) -> None:
+        bitmap = Bitmap(len(keys))
+        with self._lock:
+            for i, key in enumerate(keys):
+                if (obj := self._memory_objects.get(key)) is None:
```
ApostaC, line 666: Does this mean we will never have a "real" lookup in the storage, but just rely on in-proc states?

ApostaC (commented):
1

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Done

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Done

lmcache/v1/distributed/l2_adapters/base.py
YaoJiayi: Done

lmcache/v1/distributed/memory_manager.py
YaoJiayi: done

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Currently we are doing in-process lookup

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Persistence will be handled in followup prs

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: done

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+        self.nixl_agent.release_xfer_handle(handle)
+
+    def close(self):
+        self.nixl_agent.release_dlist_handle(self.storage_xfer_handler)
+        self.nixl_agent.release_dlist_handle(self.mem_xfer_handler)
+        self.nixl_agent.deregister_memory(self.storage_reg_descs)
+        self.nixl_agent.deregister_memory(self.mem_reg_descs)
+
+
+class NixlStoreL2Adapter(L2AdapterInterface):
```
ApostaC, line 354: Can we add the intialization code into `l2_adapters/__init__.py`?

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+                raise
+
+            # TODO(Jiayi): Tune this for better perf
+            await asyncio.sleep(0.01)
+
+        if state == "ERR":
+            raise RuntimeError("NIXL transfer failed")
+
+    def get_storage_indices(self, num_objs: int) -> list[int]:
+        return self.pool.batched_allocate(num_objs)
```
ApostaC, line 325: We do allocate here, but where do we free?

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+            self.nixl_agent.release_handle(handle)
+
+        except Exception:
+            success = False
+
+        with self._lock:
+            for key, storage_obj in zip(keys, storage_objs, strict=False):
+                self._memory_objects[key] = storage_obj
+
+            self._completed_store_tasks[task_id] = success
```
ApostaC, line 640: If the success = False, do we want to update the `self._memory_obj`? Will this cause an incorrect lookup/load?

lmcache/v1/distributed/l2_adapters/config.py
```diff
@@ -159,6 +159,83 @@ def help(cls) -> str:
...
+        )
+
+    @classmethod
+    def help(cls) -> str:
+        return (
+            "Nixl store L2 adapter config fields:\n"
+            "- backend (str): Nixl storage backend, one of "
+            f"{_VALID_NIXL_BACKENDS} (required)\n"
+            "- backend_params (dict): backend-specific string key-value pairs "
+            "(optional, default {}). File-based backends require 'path'. "
```
ApostaC: file-based backends require "file-path" rather than path

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
ApostaC, line 1: Can we have a design doc for this so that other people can modify/maintain this in the future? The doc can be put into l2_adapters/ folder (a subfolder inside it should be fine)

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -0,0 +1,630 @@
...
+    mem_reg_descs: nixlBind.nixlRegDList
+    mem_xfer_handler: NixlDlistHandle
+
+    def __init__(
+        self,
+        device: str,
+        backend: str,
+        backend_params: dict[str, str],
+        pool_size: int,
+        l1_memory: L1MemoryDesc,
```
ApostaC: It's better to provide docstrings for those fields (as well as other functions)

lmcache/v1/distributed/api.py
```diff
@@ -98,6 +98,15 @@ def ComputeKVRank(
...
 
+@dataclass(frozen=True)
+class L1MemoryDesc:
+    """Describes the L1 memory buffer registered with Nixl."""
+
+    ptr: int
+    size: int
+    align_bytes: int
+
+
```
ApostaC: Since we do not expect external classes uses this clas, can we put it into `internal_api.py`? 
We can implement a function in `l1_memory_manager.py` to get this desc.

lmcache/v1/memory_management.py
ApostaC, line 1: I'm not against this change, but just wanted to make sure that it's not needed by this PR, right?

ApostaC (commented):
Some minor comments. Otherwise LGTM.

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Oh right, the code for init is already merged. Updated

lmcache/v1/distributed/api.py
YaoJiayi: Done

lmcache/v1/memory_management.py
YaoJiayi: It's needed for initialization as we need to register the memory

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: done

lmcache/v1/distributed/l2_adapters/config.py
YaoJiayi: Fixed

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Yeah the logic here is a bit confused. Transfer errors and allocation failures are handled in the same way. Updated the logic here.

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: Will be handled in the eviction pr. Feels we might need to reuse the eviction code across different adapters and also l1 memory.

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
YaoJiayi: done

KuntaiDu (approved):
LGTM

lmcache/v1/distributed/memory_manager.py
ApostaC: Let's remove this old function to avoid confusion

ApostaC (approved):
LGTM!

lmcache/v1/distributed/memory_manager.py
YaoJiayi: ok

PR #2663, opened 2026-03-01
K3 CI Refactor

sammshen:
Rewrites four tests (`.buildkite/k3_tests/` with a new k3 based infra `.buildkite/k3_harness/`).

Benefits: 
1. better resource scheduling (multiple concurrent tasks in comprehensive can run at once) 
2. unified env setup (install nightly vllm and lmcache from source with wheels and no docker builds)
3. k3 pods make resource cleanup guaranteed

PR #2671, opened 2026-03-03
[Correctness]: Avoid overwriting APC overlap

sammshen:
NEEDS: https://github.com/vllm-project/vllm/pull/35831

Tested with the correctness test in the new experimental CI which caught a bug in MP mode: https://github.com/LMCache/LMCache/pull/2663 

running lm_eval` gsm8k (300 samples, 50 concurrent, Qwen3-14B) samples identical across two runs. Previously failed consistently on doc_id 114 and sometimes on doc_id 57) which is now passing with this PR and the vllm side adapter PR

we want to fix CUDA stream race during MP retrieve that caused incorrect outputs under high concurrency. when `GetRetrieveMetadata` aligns the retrieve start down to the chunk boundary, APC-shared blocks get included in the write target. The LMCache server then writes to these blocks on its own CUDA stream while concurrent requests read them on the vLLM stream.

do this by adding a `skip_prefix_n_tokens` parameter to the multi-layer KV transfer kernel so the server skips writing to APC-overlapping positions entirely. The kernel handles the offset natively (no Python-side tensor slicing or `.contiguous()` copy needed which would be the case because T is not the last dimension in 2LTD).


csrc/mem_kernels.cu
```diff
@@ -305,14 +305,16 @@ __global__ void load_and_reshape_multi_layer_kernel(
...
+    const int page_buffer_size, const int block_size,
+    const int skip_prefix_n_tokens) {
   const int token_id = blockIdx.x;
   const int layer_id = blockIdx.y;
   const int k_or_v = blockIdx.z;
   const int tid = threadIdx.x;
   const int num_threads = blockDim.x;
 
-  const int64_t slot_idx = slot_mapping[token_id];
+  const int kv_token_id = token_id + skip_prefix_n_tokens;
```
sammshen, line 316: we add here to avoid having to slice the slot_mapping

csrc/mem_kernels.cu
```diff
@@ -322,7 +324,7 @@ __global__ void load_and_reshape_multi_layer_kernel(
   /** Copy the data from page buffer to key_value **/
   for (int i = tid; i < scalars_per_token; i += num_threads) {
     const int64_t lmcache_offset =
-        key_value_offset(k_or_v, layer_id, token_id, i, scalars_per_token,
+        key_value_offset(k_or_v, layer_id, kv_token_id, i, scalars_per_token,
```
sammshen, line 327: we do this to avoid ahving to slice the LMCache side memory object (2LTD cannot be contiguously sliced along T dimension)

csrc/mem_kernels.cu
```diff
@@ -452,22 +455,23 @@ void multi_layer_kv_transfer_templated(
...
+    const int block_size, const int skip_prefix_n_tokens) {
   T* key_value_ptr = get_kernel_ptr<T, torch::Tensor>(key_value);
   T** page_buffer_ptrs =
       get_kernel_ptr<T*, const torch::Tensor>(key_value_ptrs);
   const int64_t* slot_mapping_ptr =
       get_kernel_ptr<const int64_t, const torch::Tensor>(slot_mapping);
 
   int num_layers = key_value.size(1);
-  int num_tokens = slot_mapping.size(0);
+  int num_tokens = key_value.size(2);
```
sammshen, line 466: there's a potentially interesting semantic discussion about whether `key_value.size(2)` or `slot_mapping.size(0)` should be used in the future as the transfer size (in tokens) but for now they are the same

csrc/mem_kernels.cu
```diff
@@ -452,22 +455,23 @@ void multi_layer_kv_transfer_templated(
...
+  int num_transfer_tokens = num_tokens - skip_prefix_n_tokens;
   int num_origin_elements = key_value.size(3);
   int elements_per_xword = sizeof(T) / key_value.element_size();
   int num_xwords = num_origin_elements / elements_per_xword;
 
   int k_or_v_size = lmc::is_mla(gpu_kv_format) ? 1 : 2;
 
-  dim3 grid(key_value.size(2), num_layers, k_or_v_size);
+  dim3 grid(num_transfer_tokens, num_layers, k_or_v_size);
```
sammshen, line 474: we launch fewer blocks in the grid :)

lmcache/v1/multiprocess/server.py
ApostaC, line 1: Probably need to assert `skip_first_n_tokens < self.chunk_size` (we can put it into adapter as well)

ApostaC (approved):
LGTM!

Oasis-Git (approved):

PR #2673, opened 2026-03-03
Fix race condition in LocalDiskBackend initialization

joaquinescalante23:
## Summary

- Fixed race condition in `local_disk_backend.py` where `os.makedirs` was called without `exist_ok=True`
- This could cause initialization failures when multiple processes try to create the cache directory simultaneously
- Added `exist_ok=True` parameter to prevent `FileExistsError`

## Fixes

Fixes #2486

## Testing

This is a minimal fix that changes one line to handle concurrent directory creation safely.

PR #2674, opened 2026-03-03
Fix race condition in FSConnector tmp file naming

### LINKED ISSUE #2517: [BUG] LMCache remote file connector breaks with concurrency
**Describe the bug**
Parallel cache puts with fs connector produce inconsistent cache state

**To Reproduce**

Start > 2 instances of vllm with LMCACHE_REMOTE_URL pointing to the same file system folder.
From a set of fixed prompts (I used 32), submit these prompts in parallel, with a queue depth of 1 per vllm instance. I did this using async python, but this could easily be done with threads/processes.

It seems the issue stems from the fs_connector.py in the `_get_file_and_tmp_path` function.
The process of populating a block cache file involves writing to a tmp file, and then using a `rename` system call to finish the operation. From a POSIX point of view, `rename` (python `os.replace`)  is acquired to be atomic.
This atomic operation is seen in the `async def put` call in the same file.
The issue occurs if two instances of the file connector are in the process of a `put`, and the following interleaving is possible:

1. Process A opens the temp file and write the data.
2. Process A closes the file (leaves the with statement)
3. Process A calls `os.replace()` which goes to a `rename` system call on Linux
4. Process B goes to open the file, but gets an error because the file does not exist
5. Process B fails to write the cache

I used a distributed POSIX filesystem for this, but I suspect this can reproduced locally on a single node as well.

**Expected behavior**
So ideally, if this interleaving happens, you would expect just an error message from B, and if the same prompt were sent again to B, it should be able to see the completed entry by A.

**Instead**, it seems that something happens inside of B's metadata, and if you continue to send prompts to only B, it will continue to send offload requests instead of reading the file produced by A.
The only way I've been able to recover this is to restart all of my lmcache instances.
This indicates that the problem with inconsistent cache state lies in memory with lmcache and not with the underlying storage.

**Proposed fix**
https://github.com/LMCache/LMCache/blob/ab97cfe47fa6cb6bf31a206d476e35448e10bbd3/lmcache/v1/storage_backend/connector/fs_connector.py#L134-L137 should probably use a random value in the tmp file name to help prevent collisions

**Screenshots**
N/A

 - OS: Ubuntu 22.04
 - Browser: firefox

**Additional context**
LMCache variables in k8s env:
```yaml
          - name: PYTHONHASHSEED
            value: "0"
          - name: VLLM_LOGGING_LEVEL
            value: "DEBUG"
            value: "50"
          - name: LMCACHE_LOCAL_CPU
            value: "false"
          - name: LMCACHE_CHUNK_SIZE
            value: "256"
            value: "true"
          - name: LMCACHE_LOG_LEVEL
            value: "DEBUG"
          - name: LMCACHE_REMOTE_URL
            value: "fs://host:0/kvc"
```
lmcache image version: 0.3.9

joaquinescalante23:
## Summary

- Added UUID to temporary filename in `_get_file_and_tmp_path` to prevent collisions when multiple processes try to write the same cache key concurrently
- Changed from `.tmp` suffix to `.{uuid}.tmp` format

## Problem (Issue #2517)

When multiple vLLM instances write to the same cache key simultaneously:
1. Process A opens temp file and writes data
2. Process A calls `os.replace()` (atomic rename)
3. Process B tries to open the same temp file but it no longer exists
4. Process B fails, and worse, enters an inconsistent state in memory

## Fix

Generate a unique temporary filename using UUID to ensure each process has its own temp file:

```python
# Before
tmp_path = file_path.with_suffix(".tmp")

# After  
tmp_file_name = f"{file_name}.{uuid.uuid4().hex}.tmp"
tmp_path = file_path.with_name(tmp_file_name)
```

## Fixes

Fixes #2517

## Testing

This is a targeted fix that adds randomness to temp file naming to prevent collisions.

### DIFF
--- a/lmcache/v1/storage_backend/connector/fs_connector.py
+++ b/lmcache/v1/storage_backend/connector/fs_connector.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import List, Optional, Tuple, no_type_check
 import asyncio
 import os
+import uuid
 
 # Third Party
 import aiofiles
@@ -134,7 +135,8 @@ class FSConnector(RemoteConnector):
         if self.relative_tmp_dir is not None:
             tmp_path = base_path / self.relative_tmp_dir / file_name
         else:
-            tmp_path = file_path.with_suffix(".tmp")
+            tmp_file_name = f"{file_name}.{uuid.uuid4().hex}.tmp"
+            tmp_path = file_path.with_name(tmp_file_name)
         return file_path, tmp_path
 
     async def exists(self, key: CacheEngineKey) -> bool:
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -121,7 +121,7 @@ class LocalDiskBackend(StorageBackendInterface):
         assert config.local_disk is not None
         self.path: str = config.local_disk
         if not os.path.exists(self.path):
-            os.makedirs(self.path)
+            os.makedirs(self.path, exist_ok=True)
             logger.info(f"Created local disk cache directory: {self.path}")
 
         self.loop = loop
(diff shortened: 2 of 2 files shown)


deng451e:
Is this PR the same as #2673

Issue #2510, opened 2026-01-29
PD Separation hangs at Decode node in v0.3.12 (Works in v0.3.11)


JianDan0212:
# Description
I am reporting a regression issue found in lmcache version 0.3.12. When running a PD (Prefill-Decode) separation setup, the request hangs/stalls at the Decode node.

I have performed a regression test and confirmed:

LMCache v0.3.12: Fails (The Decode node gets stuck).

LMCache v0.3.11: Works (PD separation runs normally).

This issue is reproducible regardless of the vLLM version used (tested on both v0.13.0 and v0.14.0).

# Environment
Hardware: NVIDIA A100

Model: Qwen3 series

LMCache Version: 0.3.12 (Problematic), 0.3.11 (Working)

vLLM Version: Verified on 0.13.0 and 0.14.0

# Reproduction Steps
Set up a PD separation cluster (Prefill node + Decode node).

Install lmcache==0.3.12.

Send a request to the cluster.

Observation: The process hangs when the request is handed over to the Decode node.

Downgrade to lmcache==0.3.11 and repeat -> The process works as expected.

<img width="2190" height="629" alt="Image" src="https://github.com/user-attachments/assets/85e7ddee-60af-4536-966f-16252511b918" />

deng451e:
Could you please share the commands and the LMCache config file you used for the PD deployment?

JianDan0212:
Below are the commands, configuration files I used, as well as the details of the error status. I am using the combination of vLLM 0.15.0 + LMCache 0.3.14, and I noticed that the same issue also exists in version 0.3.13. My hardware is the H20 GPU. I hope this information is helpful to you. Please feel free to reach out if you need any further details or assistance.
<img width="2772" height="1122" alt="Image" src="https://github.com/user-attachments/assets/3e51c350-5c5f-4a0c-8bc8-ec1590272321" />

<img width="1284" height="1008" alt="Image" src="https://github.com/user-attachments/assets/ce8fa144-b4be-4240-a90a-fa73c2d094f2" />

<img width="1233" height="644" alt="Image" src="https://github.com/user-attachments/assets/3f78f738-6833-4061-8f1d-2624afbea068" />

<img width="2535" height="207" alt="Image" src="https://github.com/user-attachments/assets/dd3b742d-92c4-4230-bb1b-4045357a8392" />

[...]

PR #2690, opened 2026-03-04
Fix regression: restore config validate() call in config.py

hlin99:
Fix regression(https://github.com/LMCache/LMCache/issues/2510): Add _update_config_from_env to namespace_extras The _validate_config function at line 504 was not being called after commit 3d88cee because the custom _update_config_from_env method was removed from namespace_extras, causing the config to use the base implementation which doesn't call validate().

This commit:
- Adds _update_config_from_env back to namespace_extras
- Updates the method to use the new _resolve_config_aliases signature
- Adds a regression test to verify validate() is called

The fix ensures that validation logic (like auto-setting save_unfull_chunk for PD mode) is properly executed when configs are updated from env vars.

- [ Y ] this PR contains unit tests


hlin99:
hi @maobaolong  would you like to take a look? the regression seems to be introduced by 3d88cee,  but i'm not 100% sure if this is intentional. please double confirm. thanks.

sammshen (approved):
lgtm

hlin99:
hi @maobaolong , just make sure you're aware of this PR... if you have other solutions, please comments. thx.

maobaolong (approved):
LGTM. Thanks for your fix.

PR #2692, opened 2026-03-04
Fault Tolerance Check

Oasis-Git:

**If applicable**:


lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -120,12 +182,35 @@ def __init__(
...
         # Read chunk size from lmcache
-        self.chunk_size = get_lmcache_chunk_size(self.mq_client)
+        try:
+            self.chunk_size = get_lmcache_chunk_size(self.mq_client)
+        except TimeoutError:
+            self.mq_client.close()
+            raise ConnectionError(
+                f"LMCache server did not respond within {LMCACHE_MQ_TIMEOUT}s. "
+                "Is the server running?"
+            ) from None
```
ApostaC, line 222: Do we want to raise the error here or mark self as "unhealthy" so that vLLM can still work?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -19,9 +20,13 @@
 )
 from lmcache.v1.multiprocess.mq import MessageQueueClient, MessagingFuture
 from lmcache.v1.multiprocess.protocol import RequestType, get_response_class
+from lmcache.v1.periodic_thread import PeriodicThread, ThreadLevel, ThreadRunSummary
 
 logger = init_logger(__name__)
 
+LMCACHE_MQ_TIMEOUT = float(os.getenv("LMCACHE_MQ_TIMEOUT", "30"))
+LMCACHE_HEARTBEAT_INTERVAL = float(os.getenv("LMCACHE_HEARTBEAT_INTERVAL", "10"))
```
ApostaC: Can we also use vLLM's `--kv-transfer-config` to pass them in? Please check the docs here: https://docs.lmcache.ai/mp/configuration.html#vllm-client-configuration

Additionaly, we should update the docs to explain the new configs.

lmcache/v1/multiprocess/protocols/controller.py
```diff
@@ -41,4 +42,12 @@ def get_protocol_definitions() -> dict[str, ProtocolDefinition]:
             response_class=int,
             handler_type=HandlerType.SYNC,
         ),
+        # Health check
+        # Payload: None
+        # Returns: bool - True if server is healthy
+        "HEALTH_CHECK": ProtocolDefinition(
```
ApostaC: Can we rename it to "PING"? 

In general, HEALTH_CHECK sounds like checking a more detailed status for all the components, but PING sounds like checking whether the server is responsive or not. 

I feel like what we are doing here is the latter.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -64,10 +69,67 @@ def get_lmcache_chunk_size(
...
+    Manages a threading.Event that adapters use to gate operations.
+    When unhealthy, the adapter enters degraded mode; if the server
+    recovers, the adapter automatically resumes normal operation.
+    """
+
+    def __init__(
+        self,
+        mq_client: MessageQueueClient,
+        health_event: threading.Event,
+    ):
```
ApostaC, line 111: Add the docstring for the args. We should explain what it does to `health_event`

lmcache/integration/vllm/vllm_multi_process_adapter.py
ApostaC: Actually, I don't think we should have a timeout here. A simpler solution is just to let vLLM block here if the LMCache is not ready. In this case, we don't need to maintain complicated states (i.e., whether the chunk size is initialized) if we want to implement "reconnect" in the future.

Another reason is that in some cases, it may take a while for LMCache to be ready.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -354,7 +480,13 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
             ],
         )
-        future.result()
+        try:
+            future.result(timeout=LMCACHE_MQ_TIMEOUT)
+        except TimeoutError:
+            raise ConnectionError(
+                "LMCache server did not respond to register_kv_caches "
+                f"within {LMCACHE_MQ_TIMEOUT}s. Is the server running?"
+            ) from None
```
ApostaC, line 565: Similar to what I said above, we can just let vLLM block here during the initialization if LMCache is not ready.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -184,12 +272,16 @@ def check_lookup_result(self, request_id: str) -> int | None:
...
+        if request_id not in self.lookup_futures:
+            # No future — either unhealthy at submit time or already cleaned up
+            return 0
 
         future = self.lookup_futures[request_id]
         if not future.query():
+            if not self.is_healthy:
+                # Server went down — give up on this lookup
+                self.lookup_futures.pop(request_id)
+                return 0
```
ApostaC: Can we move this if-branch out side? We can check it before checking the future.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -314,12 +417,30 @@ def __init__(
...
         # Read chunk size from lmcache
-        chunk_size = get_lmcache_chunk_size(self.mq_client)
+        try:
+            chunk_size = get_lmcache_chunk_size(self.mq_client)
+        except TimeoutError:
+            self.mq_client.close()
+            raise ConnectionError(
+                f"LMCache server did not respond within {LMCACHE_MQ_TIMEOUT}s. "
+                "Is the server running?"
+            ) from None
```
ApostaC, line 502: Similar to what I said above

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -369,6 +501,9 @@ def submit_store_request(
             event: The CUDA event that is recorded after the current
                 model inference step
         """
+        if not self.is_healthy:
+            return
```
ApostaC, line 581: If we don't add the store request here, will it cause a problem if the request ID is never returned from `get_finished_request`, and thus block vLLM from releasing the request's GPU pages?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -464,6 +603,29 @@ def get_finished(
...
+            self.retrieve_futures.clear()
+
+            self.finished_stores.update(finished_stores)
+            ret_stores = set()
+            for req_id in finished_req_ids_from_engine:
+                if req_id in self.finished_stores:
+                    self.previously_finished.add(req_id)
+                else:
+                    ret_stores.add(req_id)
+            ret_stores.update(self._update_and_get_finished_store())
```
ApostaC: This piece of code is repetitive. Let's have a helper function
```python
    def _process_internal_and_external_finished_stores(
        self, finished_req_ids_from_lmcache, finished_req_ids_from_engine
    ):
        """
            ... doc string here ...
        """ 
        self.finished_stores.update(finished_req_ids_from_lmcache)
        ret_stores = set()
        for req_id in finished_req_ids_from_engine:
            if req_id in self.finished_stores:
                self.previously_finished.add(req_id)
            else:
                ret_stores.add(req_id)
        ret_stores.update(self._update_and_get_finished_store())
        return ret_stores
```

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -474,22 +636,20 @@ def get_finished(
...
             if not s_result:
-                # TODO: add error handling here
                 logger.error(
                     "Something went wrong when processing the "
                     "store request for request_id=%s",
                     request_id,
                 )
 
-        for request_id, r_future in self.retrieve_futures.items():
+        for request_id, (r_future, r_block_ids) in self.retrieve_futures.items():
```
ApostaC: nit: 
```suggestion
        for request_id, (r_future, _) in self.retrieve_futures.items():
```

ApostaC (changes requested):
Some changes are needed. See the details below

ApostaC:
cc @maobaolong for visibility.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -19,9 +20,13 @@
 )
 from lmcache.v1.multiprocess.mq import MessageQueueClient, MessagingFuture
 from lmcache.v1.multiprocess.protocol import RequestType, get_response_class
+from lmcache.v1.periodic_thread import PeriodicThread, ThreadLevel, ThreadRunSummary
 
 logger = init_logger(__name__)
 
+LMCACHE_MQ_TIMEOUT = float(os.getenv("LMCACHE_MQ_TIMEOUT", "30"))
```
ApostaC: Add some docs about when this timeout is used

lmcache/integration/vllm/vllm_multi_process_adapter.py
ApostaC: Actually let vLLM die is also a good choice

lmcache/v1/multiprocess/protocols/controller.py
```diff
@@ -41,4 +42,12 @@ def get_protocol_definitions() -> dict[str, ProtocolDefinition]:
...
+            payload_classes=[],
+            response_class=bool,
+            handler_type=HandlerType.SYNC,
```
ApostaC: We can use BLOCKING as the handler type

lmcache/v1/multiprocess/server.py
```diff
@@ -487,6 +487,15 @@ def free_lookup_locks(
...
     # =========================================================================
 
+    def health_check(self) -> bool:
+        """
+        Check if the server is healthy by running a memory check.
+
+        Returns:
+            bool: True if the server memory is healthy.
+        """
+        return self.storage_manager.memcheck()
```
ApostaC: return True

lmcache/v1/multiprocess/server.py
```diff
@@ -487,6 +487,15 @@ def free_lookup_locks(
     # Utility methods
     # =========================================================================
 
+    def health_check(self) -> bool:
```
ApostaC: rename to ping

ApostaC (commented):
Something more

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -19,9 +20,16 @@
 )
 from lmcache.v1.multiprocess.mq import MessageQueueClient, MessagingFuture
 from lmcache.v1.multiprocess.protocol import RequestType, get_response_class
+from lmcache.v1.periodic_thread import PeriodicThread, ThreadLevel, ThreadRunSummary
 
 logger = init_logger(__name__)
 
+# Timeout (seconds) for blocking MQ requests: initial chunk-size query,
+# KV cache registration/unregistration, and other synchronous operations.
+DEFAULT_MQ_TIMEOUT: float = 30.0
```
ApostaC: Let's use 300 here

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -64,10 +72,78 @@ def get_lmcache_chunk_size(
         An integer representing the LMCache chunk size
     """
     future = send_lmcache_request(mq_client, RequestType.GET_CHUNK_SIZE, [])
-    chunk_size = future.result()
+    chunk_size = future.result(timeout=DEFAULT_MQ_TIMEOUT)
```
ApostaC, line 75: We are hard-coding the timeout here. Does this mean that people cannot overwrite this from vLLM's config?

ApostaC (approved):
LGTM!

KuntaiDu (approved):
LGTM!

PR #2695, opened 2026-03-05
[MP][UX] Unified config + argparse for multiprocess mode

ApostaC:

The multiprocess module (`server.py`, `blend_server.py`, `http_server.py`) has fragmented config and argument parsing:

- `server.py` defines 5 inline `add_argument` calls, and `run_cache_server()` takes 8 individual parameters.
- `http_server.py` has a standalone `ServerConfig` dataclass with a `to_storage_manager_config()` method that hardcodes L1 settings and eviction policy. It does not support L2 adapters, hash algorithm, or Prometheus from CLI. It also uses `--zmq-host`/`--zmq-port` instead of the `--host`/`--port` convention used everywhere else.
- `blend_server.py` reuses `parse_args` from `server.py` (good), but still passes 8 individual params to its own `run_cache_server()`.

This PR creates `lmcache/v1/multiprocess/config.py` following the established composable `(DataclassConfig, add_*_args, parse_args_to_*_config)` triple pattern from `lmcache/v1/distributed/config.py`, unifying the scattered config.

**User-facing changes:**

- `--zmq-host` / `--zmq-port` (http_server.py only) are replaced by `--host` / `--port` (consistent across all servers). HTTP-specific args are now `--http-host` / `--http-port`.
- `--cpu-buffer-size` is removed from http_server.py. Users now use `--l1-size-gb` (from the storage manager arg group) instead.
- `http_server.py` now supports the full set of CLI args: L2 adapters, hash algorithm, Prometheus, eviction tuning — previously hardcoded or missing.
- All three servers (`server.py`, `blend_server.py`, `http_server.py`) now show organized argument groups in `--help`: MP Server, L1 Memory, Eviction, L2 Adapters, Prometheus (and HTTP Frontend for http_server.py).

- `MPCacheEngine` constructor is unchanged — it doesn't need host/port/max_workers (those are ZMQ transport concerns). `run_cache_server()` extracts the relevant fields from `MPServerConfig`.
- The `_configs` dict pattern in `http_server.py` replaces the old `_server_config` global; it avoids the need for uninitialized `StorageManagerConfig` sentinels since `run_http_server()` always populates it before FastAPI's lifespan starts.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/blend_server.py
+++ b/lmcache/v1/multiprocess/blend_server.py
@@ -34,6 +34,10 @@ from lmcache.v1.mp_observability.prometheus_controller import (
     get_prometheus_controller,
     init_prometheus_controller,
 )
+from lmcache.v1.multiprocess.config import (
+    MPServerConfig,
+    parse_args_to_mp_server_config,
+)
 from lmcache.v1.multiprocess.custom_types import (
     IPCCacheEngineKey,
     KVCache,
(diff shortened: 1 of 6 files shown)


Oasis-Git (approved):

KuntaiDu (approved):
LGTM!

PR #2696, opened 2026-03-05
[MP][Observability] Add telemetry subsystem for multiprocess mode

ApostaC:

Adds an event-based telemetry framework for MP mode observability, complementing the existing Prometheus metrics (aggregated counters/histograms) with per-operation event tracing.

**Design:** Uses a START/END event model with session IDs. Call sites emit a `START` event when an operation begins and an `END` event when it completes, sharing the same `session_id` to correlate them into spans. This two-event design is necessary because start and end happen at different call sites in LMCache's async submit/check/complete patterns. See `DESIGN.md` for full architecture details and how to implement new processors.

**Key components:**
- `TelemetryEvent` dataclass (name, event_type, session_id, metadata)
- `TelemetryProcessor` / `TelemetryProcessorConfig` ABCs with a registry pattern (same as L2 adapters)
- `LoggingProcessor` built-in processor with configurable log level
- `TelemetryController` with lock-free deque, daemon drain thread, tail-drop backpressure
- Disabled-by-default singleton — no thread or overhead until explicitly enabled
- `make_start_event()` / `make_end_event()` convenience functions for concise call sites

**User-facing changes:**
- New CLI args: `--enable-telemetry`, `--telemetry-max-queue-size`, `--telemetry-processor JSON`
- Example: `--enable-telemetry --telemetry-processor '{"type": "logging", "log_level": "INFO"}'`
- Telemetry events emitted for `store`, `retrieve`, and `lookup` operations in `server.py`
- Wired into both `server.py` and `blend_server.py` startup/shutdown lifecycle

- The processor registry + factory pattern mirrors `lmcache/v1/distributed/l2_adapters/config.py`
- `collections.deque` is used without a lock — CPython GIL guarantees atomic `append()`/`popleft()` with a single consumer thread
- GPU-stream call sites use `launch_host_func` with `is_enabled()` guards to avoid registering unnecessary CUDA callbacks when telemetry is disabled
- This PR implements the framework only. Adding more call sites (e.g., blend operations, L2 store/retrieve) is a follow-up

**If applicable**:

### DIFF
--- a/lmcache/v1/mp_observability/telemetry/controller.py
+++ b/lmcache/v1/mp_observability/telemetry/controller.py
@@ -79,8 +79,11 @@ class TelemetryController:
             self._processors.append(processor)
 
     def start(self) -> None:
-        """Start the drain thread. No-op when disabled."""
+        """Start the drain thread. No-op when disabled or already started."""
         if not self._config.enabled:
+            return
+
+        if self._thread is not None and self._thread.is_alive():
             return
 
         self._thread = threading.Thread(
--- a/lmcache/v1/mp_observability/telemetry/controller.py
+++ b/lmcache/v1/mp_observability/telemetry/controller.py
@@ -69,6 +69,7 @@ class TelemetryController:
                 self._last_discard_warning = now
             return
 
+        event.timestamp = time.time()
         self._queue.append(event)
         self._wake.set()
 
--- a/lmcache/v1/mp_observability/telemetry/controller.py
+++ b/lmcache/v1/mp_observability/telemetry/controller.py
@@ -1,0 +1,196 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Telemetry controller: queue, drain thread, singleton."""
+
+# Future
+from __future__ import annotations
+
+# Standard
+import collections
+import threading
+import time
+
+# First Party
+from lmcache.logging import init_logger
+from lmcache.v1.mp_observability.telemetry.config import TelemetryConfig
+from lmcache.v1.mp_observability.telemetry.event import TelemetryEvent
+from lmcache.v1.mp_observability.telemetry.processors.base import (
+    TelemetryProcessor,
+)
+from lmcache.v1.mp_observability.telemetry.processors.logging_processor import (
+    LoggingProcessor,
+    LoggingProcessorConfig,
+)
+
+logger = init_logger(__name__)
+
+
+class TelemetryController:
+    """Manages telemetry event ingestion, queueing, and dispatch to processors.
+
+    Events are appended to a lock-free deque on the hot path and drained by a
+    background thread that dispatches to registered processors.
+    """
+
+    def __init__(self, config: TelemetryConfig):
+        self._config = config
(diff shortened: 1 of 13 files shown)


ApostaC:
Example usage command (log the telemetry to terminal)
```bash
python3 -m lmcache.v1.multiprocess.server --max-workers 4 --port 6555 \
    --l1-size 70 --eviction-policy LRU \
    --l2-adapter '{"type": "mock", "max_size_gb": 200, "mock_bandwidth_gb": 10}' \
    --enable-telemetry --telemetry-processor '{"type": "logging", "log_level":"warning"}'
```

Expected output:
image

KuntaiDu (commented):
Some quick comments:
1. previously IIRC you prefer to keep the `lmcache/v1` for pure storage rather than request-level telemetry. Is that still true?
2. Please also re-implement the request_finished telemetry under the integration folder using this framework, and deprecate the one under integration folder.
3. When request fully ends, it would be really nice if we can have a "summary" of this request logged out, to avoid people manually align the time. Example:
```
Summary of request chat-cmpl-xxx:
Received at (time)
Preprocessing:  ??ms (st->ed: 2:04.025 -> 2.04.124)
Retrieve ?? tokens: ?? ms (st->ed: ???)
LLM computation runs: ??ms (st->ed: ???)
Store ?? tokens: ?? ms (st->ed: ???)
```
This might be convenient for debugging.

ApostaC:
@KuntaiDu 

> previously IIRC you prefer to keep the lmcache/v1 for pure storage rather than request-level telemetry. Is that still true?

This one is actually designed for request-level telemetry. The core design here is to have a very fast, non-blocking API to submit the telemetry event (i.e., push it into a queue), and then there is a background thread processing all the telemetry events.

> Please also re-implement the request_finished telemetry under the integration folder using this framework, and deprecate the one under integration folder.

Unfortunately, this could be pretty difficult, since we need to track the TP information at the LMCache-level, which is quite hard. Additionally, though both modules are named as "telemetry", this one is more for the debuggability and visualization (e.g., OpenTelemetry support)

> When request fully ends, it would be really nice if we can have a "summary" of this request logged out, to avoid people manually align the time. Example:

Will have a follow-up PR for it. There should also be some tools to visualize the spans.

lmcache/v1/mp_observability/telemetry/controller.py
```diff
@@ -0,0 +1,225 @@
...
+        event.timestamp = time.time()
+        self._queue.append(event)
+        self._wake.set()
+
+    def register_processor(self, processor: TelemetryProcessor) -> None:
+        """Register a processor for event dispatch (thread-safe)."""
+        with self._lock:
+            self._processors.append(processor)
+
+    def start(self) -> None:
```
sammshen, line 81: should this be idempotent? or doesn't matter

ApostaC: Good point! Let me fix it

KuntaiDu (approved):
LGTM!

sammshen (approved):
LGTM!

PR #2697, opened 2026-03-05
Fix to support mla multiple tp failed to read issue

maobaolong:

Fix the issue when run vllm and lmcache with mp mode for deepseek mla tp=2

Clipboard_Screenshot_1772696677

- After fixed, send two same request to verify the store and retrieve flow.
```
[2026-03-05 15:32:00,163] LMCache INFO: Stored 256 tokens in 0.006 seconds (server.py:294:__main__)
(APIServer pid=5070) INFO 03-05-2026-15:32:00.192 [serving.py:1354] Finished request: model=vllm_cpu_offload, request_id=896ea0c7d96c9b31, input_tokens=343, output_tokens=10, cached_tokens=0, total_tokens=353, ttft=0.099s, total_time=0.099s, decode_gen_speed=0.00tokens/s, finish_reason=length

[2026-03-05 15:32:20,788] LMCache INFO: Prefetch request completed (L1+L2): 1/1 prefix hits (1 L1, 0 L2) in 0.0 ms (request_id=-1) (storage_manager.py:357:lmcache.v1.distributed.storage_manager)
[2026-03-05 15:32:20,791] LMCache INFO: Retrieved 256 tokens in 0.001 seconds (server.py:393:__main__)
[2026-03-05 15:32:20,792] LMCache INFO: Retrieved 256 tokens in 0.000 seconds (server.py:393:__main__)
```

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -104,6 +104,7 @@ class LMCacheMPSchedulerAdapter:
         world_size: int,
         kv_rank: int,
         vllm_block_size: int,
+        tp_size: int = 1,
     ):
         """
         Args:
@@ -114,6 +115,8 @@ class LMCacheMPSchedulerAdapter:
             world_size: The world size used for LMCache keys
             kv_rank: The kv rank used for LMCache keys
             vllm_block_size: The block size used in vLLM
+            tp_size: Tensor-parallel size for MLA
+                multi-reader locking (default 1).
         """
         self.mq_client = MessageQueueClient(server_url, context)
 
@@ -122,6 +125,7 @@ class LMCacheMPSchedulerAdapter:
 
         self.model_name = model_name
         self.world_size = world_size
+        self.tp_size = tp_size
 
         # Read chunk size from lmcache
         self.chunk_size = get_lmcache_chunk_size(self.mq_client)
@@ -174,7 +178,7 @@ class LMCacheMPSchedulerAdapter:
         future = send_lmcache_request(
             self.mq_client,
             RequestType.LOOKUP,
-            [key],
+            [key, self.tp_size],
         )
         job_id = future.result()
         self._lookup_job_ids[request_id] = job_id
@@ -260,7 +264,7 @@ class LMCacheMPSchedulerAdapter:
         send_lmcache_request(
             self.mq_client,
             RequestType.FREE_LOOKUP_LOCKS,
-            [key],
+            [key, self.tp_size],
         )
 
     def end_session(self, request_id: str) -> None:
--- a/lmcache/v1/distributed/l1_manager.py
+++ b/lmcache/v1/distributed/l1_manager.py
@@ -83,6 +83,38 @@ def l1_mgr_synchronized(func):
 
 L1OperationResult = tuple[L1Error, MemoryObj | None]
 
+# Upper bound for the count parameter in reserve_read / finish_read
+# to prevent a single call from holding the global lock for too long.
+MAX_READ_LOCK_COUNT = 128
+
+
+def _validate_extra_count(extra_count: int) -> int:
+    """Validate and clamp extra_count.
+
+    Args:
+        extra_count: Extra lock count on top of the
+            default 1 lock.
+
+    Returns:
+        Clamped value in [0, MAX_READ_LOCK_COUNT - 1].
+    """
+    if extra_count < 0:
+        logger.warning(
+            "L1Manager: extra_count=%d is invalid, clamping to 0",
+            extra_count,
+        )
+        return 0
+    upper = MAX_READ_LOCK_COUNT - 1
+    if extra_count > upper:
+        logger.warning(
+            "L1Manager: extra_count=%d exceeds limit=%d, clamping",
+            extra_count,
+            upper,
+        )
+        return upper
+    return extra_count
+
+
 # Main classes
 
 
@@ -159,20 +191,30 @@ class L1Manager:
     def reserve_read(
         self,
         keys: list[ObjectKey],
+        extra_count: int = 0,
     ) -> dict[ObjectKey, L1OperationResult]:
         """Reserve read access for the given keys.
 
         Args:
-            keys: The list of object keys to reserve read access for.
+            keys: The list of object keys to reserve
+                read access for.
+            extra_count: Extra read locks on top of the
+                default 1 lock.  Total locks acquired per
+                key = 1 + extra_count.  Useful when multiple
+                workers each consume one read lock for the
+                same key (e.g. MLA models with TP > 1).
 
         Returns:
-            A dictionary mapping each object key to a tuple of
-            (L1Error, Optional[MemoryObj]).
+            A dictionary mapping each object key to a tuple
+            of (L1Error, Optional[MemoryObj]).
 
         Errors:
             KEY_NOT_EXIST: The key does not exist.
-            KEY_NOT_READABLE: The key exists but is not readable.
+            KEY_NOT_READABLE: The key exists but is not
+                readable.
         """
+        extra_count = _validate_extra_count(extra_count)
+        total = 1 + extra_count
         ret: dict[ObjectKey, L1OperationResult] = {}
         successful_keys: list[ObjectKey] = []
         for key in keys:
@@ -185,7 +227,11 @@ class L1Manager:
                 ret[key] = (L1Error.KEY_NOT_READABLE, None)
                 continue
 
-            entry.read_lock.lock()
+            # TODO(perf): support a count argument in
+            # TTLLock.lock() to avoid Python for-loop
+            # overhead (TTLLock is C++ std::atomic).
+            for _ in range(total):
+                entry.read_lock.lock()
             ret[key] = (L1Error.SUCCESS, entry.memory_obj)
             successful_keys.append(key)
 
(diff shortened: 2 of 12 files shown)


lmcache/v1/distributed/l1_manager.py
```diff
@@ -159,11 +159,16 @@ def register_listener(self, listener: L1ManagerListener) -> None:
     def reserve_read(
         self,
         keys: list[ObjectKey],
+        count: int = 1,
```
ApostaC: Can we make this become `extra_count = 0`, so that there won't be any misuse of this interface as `reserve_read(keys, count = 0)`?

lmcache/v1/distributed/l1_manager.py
```diff
@@ -232,13 +238,21 @@ def unsafe_read(
         return ret
 
     @l1_mgr_synchronized
-    def finish_read(self, keys: list[ObjectKey]) -> dict[ObjectKey, L1Error]:
+    def finish_read(
+        self,
+        keys: list[ObjectKey],
+        count: int = 1,
```
ApostaC: Do we need to have a configurable count for finish_read?

ApostaC: Oh, I see the need! Ignore my comment above.
But I would like to have "extra_count" similar to above

lmcache/v1/multiprocess/server.py
```diff
@@ -416,12 +416,15 @@ def lookup(
...
-                layout_desc = get_layout_desc(
-                    self.gpu_contexts[gpu_id], self.chunk_size
-                )
-                break
+                num_readers += 1
+                if layout_desc is None:
+                    layout_desc = get_layout_desc(
+                        self.gpu_contexts[gpu_id],
+                        self.chunk_size,
+                    )
```
ApostaC: We cannot count num_readers this way here. The example is if the LMCache is connecting to 4 vLLM instances, where each vLLM instance runs TP=2, the current implementation will count num_reader to 8.

The best way to do it is to update the protocol and pass the TP size when calling lookup. @maobaolong It may need to create a PR on the vLLM side, modifying the `lmcache_mp_connector.py` (need to initialize the `LMCacheMPSchedulerAdapter` with the TP information). I can help review the PR there, should be a very simple one.

lmcache/v1/multiprocess/server.py
```diff
@@ -476,7 +481,18 @@ def free_lookup_locks(
...
+
+        # Count how many GPU contexts share the same
+        # (model_name, world_size) to match lookup's num_readers.
+        model_name = key.model_name
+        world_size = key.world_size
+        num_readers = sum(
+            1
+            for m, w in self.gpu_context_meta.values()
+            if m == model_name and w == world_size
+        )
```
ApostaC: Here we have the same issue as `lookup`

ApostaC (changes requested):
The big problem is the `num_reader` calculation. A minor issue is whether to use `count` or `extra_count`. 
Otherwise LGTM!

lmcache/v1/distributed/l1_manager.py
maobaolong: Thanks for this suggestion, it is better.

lmcache/v1/multiprocess/server.py
maobaolong: https://github.com/vllm-project/vllm/pull/36129 PR created

lmcache/v1/multiprocess/server.py
maobaolong: done

maobaolong:
@ApostaC Is this failure related to this PR?

It is from https://buildkite.com/lmcache/k3-correctness-test/builds/66
```
[INFO] Using GPU(s): 0
[INFO] Converting ShareGPT dataset to OpenAI format...
Converted /root/correctness/.ShareGPT_V3_unfiltered_cleaned_split.json -> ./shareGPT_dataset.json
[INFO] Starting BASE vLLM server on port 8000...
[INFO] Waiting for Base server readiness...
Waiting for vLLM on port 8000 (timeout=180s)...
vLLM failed to start on port 8000 within 180s
[ERROR] Base vLLM failed to start
[INFO] Stopping vLLM process (PID: 1011)
[INFO] Collecting logs into build_019cbe12-8c8c-4d0d-87bc-c52db6134f5c.log
usage: vllm serve [model_tag] [options]
vllm serve: error: argument --compilation-config/-cc: 1 validation error for CompilationConfig
level
  Unexpected keyword argument [type=unexpected_keyword_argument, input_value=0, input_type=int]
    For further information visit https://errors.pydantic.dev/2.12/v/unexpected_keyword_argument
```

ApostaC:
@maobaolong The K3 test is the new environment we just setup. The failure should not block the merge of the PR.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -277,6 +281,7 @@ def _create_key(
             start=start,
             end=end,
             request_id=request_id,
+            tp_size=self.tp_size,
```
ApostaC: I would suggest we have the tp_size in the lookup interface instead of putting it here. The reason is that the IPCCacheEngineKey already has `world_size`, and having a `tp_size` here may confuse people.

My proposal:
1. Edit `lmcache/v1/multiprocess/protocols/engine.py` and add the tp_size to the argument list of `LOOKUP` operation.
2. In the adapter (this file), pass the tp_size when calling `send_lmcache_request` for LOOKUP and FREE_LOOKUP_LOCKS. E.g., 
```python3
        send_lmcache_request(
            self.mq_client,
            RequestType.LOOKUP,
            [key, self.tp_size],
        )
```
3. In server.py (`lookup` and `free_lookup_locks`), check the tp_size argument and MLA flag, and compute the `extra_count` accordingly.

lmcache/v1/distributed/l1_manager.py
```diff
@@ -83,6 +83,38 @@ def wrapper(self: "L1Manager", *args, **kwargs):
 
 L1OperationResult = tuple[L1Error, MemoryObj | None]
 
+# Upper bound for the count parameter in reserve_read / finish_read
+# to prevent a single call from holding the global lock for too long.
+MAX_READ_LOCK_COUNT = 128
+
+
+def _validate_extra_count(extra_count: int) -> int:
```
ApostaC, line 91: 👍 I like this

lmcache/v1/distributed/l1_manager.py
```diff
@@ -185,7 +227,8 @@ def reserve_read(
                 ret[key] = (L1Error.KEY_NOT_READABLE, None)
                 continue
 
-            entry.read_lock.lock()
+            for _ in range(total):
+                entry.read_lock.lock()
```
ApostaC, line 234: Can we have a TODO/a issue here? It will be more performant if we support a `count` argument in `TTLLock.lock()` and `TTLLock.unlock()`. Since `TTLLock` is implemented in C++ with `std::atomic`, having a python for loop here may impact the performance.

lmcache/v1/multiprocess/server.py
```diff
@@ -431,41 +448,60 @@ def lookup(
...
+        # populate tp_size (default 1) before
+        # https://github.com/vllm-project/vllm/pull/36129.
+        # Non-MLA: world_size includes TP dimension, so
+        #   tp <= world_size → extra_readers = 0 (each
+        #   ObjectKey is retrieved by exactly 1 worker).
+        # MLA: world_size excludes TP (divided by tp), so
+        #   tp > world_size → extra_readers = tp - 1 (all
+        #   TP workers retrieve the same ObjectKey).
+        tp = key.tp_size if key.tp_size > 1 else world_size
+        extra_readers = tp - 1 if tp > world_size else 0
```
ApostaC: The TP problem will only occur for MLA, this is different TP workers will share the same object only in MLA.
For non-MLA models, different TP workers will have their own "object", and therefore, only need to lock once.

ApostaC: One thing we may need to double-check: there are some codes on the vLLM side, which will pass "world_size = 1" for MLA models even with TP, and all the LMCache logic actually relies on this to do the correct MLA + TP handling (lookup, store, and retrieve) (for example, see the implementation of `ipc_keys_to_object_keys`, which will explode the keys based on the world_size, and it's expected to be 1 for MLA case)

We probably want to have a better semantics for `world_size` in the future

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/multiprocess/server.py
maobaolong: @ApostaC Updated the comments.

lmcache/v1/distributed/l1_manager.py
maobaolong: TODO comment added.

lmcache/integration/vllm/vllm_multi_process_adapter.py
maobaolong: Great idea, addressed. 

But I did nothing for `free_lookup_locks` operation, since we use `lookup_lock_counts` to get the `extra_readers` other than the argument.

maobaolong:
@ApostaC Thanks for the previous review, addressed you suggestions, PTAL.

lmcache/v1/multiprocess/server.py
```diff
@@ -455,41 +448,73 @@ def lookup(
...
-        # 2. the lookup sort the keys in prefix order and breaks at the first failure
+        # 2. the lookup sort the keys in prefix order
+        #    and breaks at the first failure
         found_count = found_count // ipc_keys[0].world_size
         return found_count
 
     def free_lookup_locks(
         self,
         key: IPCCacheEngineKey,
     ) -> None:
```
ApostaC, line 692: Let's have tp_size passed here so we don't need lookup_lock_counts

ApostaC (approved):
Otherwise LGTM as we discussed offline. Giving an approve here

lmcache/v1/multiprocess/server.py
maobaolong: done

ApostaC (commented):
Sorry for the back and forth, but can we do the following things:
- Unify the naming of `extra_readers` (used in `finished_read_prefetched`) vs `extra_count` (used in other functions). Let's have `extra_count` in all the places.

Another thing is that there are unit tests using the `lookup` and `free_lookup_locks` interface. We need to update them otherwise the UTs will fail. These include:
- `test_mq_handler_helpers.py`
- `test_mq.py`
- `test_cache_server.py`
- `test_blend_server.py`

lmcache/v1/multiprocess/server.py
```diff
@@ -132,6 +132,46 @@ def resolve_key(
     ]
 
 
+def compute_extra_count(
+    tp_size: int,
+    world_size: int,
+) -> int:
+    """Compute extra count for MLA multi-reader locking.
```
chunxiaozheng, line 139: Thanks for the comments!

lmcache/v1/multiprocess/server.py
```diff
@@ -517,15 +577,7 @@ def lookup(
...
-        # Find the gpu context and calculate the layout desc
-        layout_desc: MemoryLayoutDesc | None = None
-        for gpu_id, (m_name, w_size) in self.gpu_context_meta.items():
-            if m_name == model_name and w_size == world_size:
-                layout_desc = get_layout_desc(
-                    self.gpu_contexts[gpu_id], self.chunk_size
-                )
-                break
-
+        layout_desc = self._find_layout_desc(model_name, world_size)
```
chunxiaozheng, line 580: This is only a refactor.

maobaolong: yeah

chunxiaozheng (approved):
After discuss offline, LGTM!

maobaolong:
@ApostaC @chunxiaozheng Thanks for your review!

PR #2706, opened 2026-03-06
[Correctness]: Fix the overlapping race condition for non-MP as well

sammshen:
essentially the same issue as https://github.com/LMCache/LMCache/pull/2671/ where the read and write stream have a race condition on shared prefix blocks across requests.

Why was this not caught by the previous correctnes tests. 

the high concurrency had the same lmcache chunk size as the vllm block size which avoids this error. 
the single request only had one request which was not enough concurrency to tirgger the race condition.

this PR also changes the chunk size of teh high concucrrency correctness test back to 256

### DIFF
--- a/.buildkite/k3_tests/correctness/scripts/run-correctness.sh
+++ b/.buildkite/k3_tests/correctness/scripts/run-correctness.sh
@@ -126,7 +126,7 @@ stop_vllm
 
 echo "[INFO] Preparing LMCache config (cpu.yaml)..."
 cat <<EOF > cpu.yaml
-chunk_size: 16
+chunk_size: 256
 local_cpu: true
 max_local_cpu_size: 50
 EOF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -805,13 +805,6 @@ class LMCacheConnectorV1Impl:
             )
             token_mask[:masked_token_count] = False
 
-            # prevent lmcache from overwriting APC-shared blocks:
-            # setting slot_mapping=-1 makes the xfer kernel skip
-            # those tokens (checks slot_idx < 0).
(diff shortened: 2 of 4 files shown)


.buildkite/k3_tests/correctness/scripts/run-correctness.sh
```diff
@@ -126,7 +126,7 @@ stop_vllm
 
 echo "[INFO] Preparing LMCache config (cpu.yaml)..."
 cat <<EOF > cpu.yaml
-chunk_size: 16
```
sammshen, line 129: this is why the correctness tests could not catch the error before

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -819,13 +819,15 @@ def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
                         token_mask[:lmcache_cached_tokens],
                         kvcaches=kvcaches,
                         slot_mapping=slot_mapping[:lmcache_cached_tokens],
+                        vllm_cached_tokens=request.load_spec.vllm_cached_tokens,
```
sammshen, line 822: we pass this as kwargs which eventually makes its way to gpu connector

DongDongJu (approved):
LGTM.
So, my understanding is that layerwise is enough to small size which not happening this problem.

sammshen:
@DongDongJu layerwise should have the same issue but has other bugs that prevent revealign this one

DongDongJu:
> @DongDongJu layerwise should have the same issue but has other bugs that prevent revealign this one

LoL. I see. do you know what is the left bug for layerwise by any chance?

sammshen:
@DongDongJu not sure, did your StopIteration fix go through?

sammshen:
I see that you mention the problem disappeared: https://github.com/LMCache/LMCache/pull/2268

ApostaC (approved):
LGTM!

PR #2707, opened 2026-03-06
[MP][Telemetry] Hot-fix to enable the telemetry logging for store

ApostaC:

Re-enables telemetry logging for store operations in the CuPy stream, which was previously disabled due to a hanging issue in certain environments.

The fix introduces a `_can_log_store` flag that gates store telemetry logging until after the first retrieve operation completes. This ensures a retrieve operation is always logged in the CuPy stream before any store operation, which avoids the hanging behavior.

The root cause of the hang appears to be related to logging a store telemetry event in the CuPy stream before any retrieve event has been logged. The flag-based ordering constraint is a minimal fix that preserves the telemetry data while avoiding the problematic code path.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/server.py
+++ b/lmcache/v1/multiprocess/server.py
@@ -178,6 +178,11 @@ class MPCacheEngine:
         )
         self.session_manager = SessionManager(self.token_hasher)
 
+        # FIX: fix the problem of telemetry logging in cupy stream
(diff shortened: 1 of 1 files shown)


YaoJiayi (approved):

KuntaiDu (approved):
LGTM

sammshen (approved):
LGTM

PR #2710, opened 2026-03-07
[MP][Core] Update the workflow for lookup to avoid busy loop

ApostaC:

Splits the MP lookup operation into a two-phase non-blocking protocol to eliminate the busy-loop in `MPCacheEngine.lookup()`.

Previously, `lookup()` called `StorageManager.submit_prefetch_task()` then busy-looped on `query_prefetch_status()` until the result was ready. With `max_workers=1` (the default), this blocked the sole thread pool worker, preventing all other BLOCKING requests (STORE, RETRIEVE, FREE_LOOKUP_LOCKS, END_SESSION) from executing concurrently.

Now:
- **LOOKUP** (BLOCKING handler): hashes keys, submits the prefetch task, stores a `PrefetchHandle` server-side, and returns a prefetch job ID immediately — no busy loop.
- **QUERY_PREFETCH_STATUS** (new SYNC handler): takes a job ID, polls the prefetch status, and returns `int | None`. Runs in the main ZMQ loop so it never consumes a thread pool worker. Auto-cleans the job entry when a non-None result is returned (exactly-once semantics).

The client adapter (`LMCacheMPSchedulerAdapter`) blocks on `LOOKUP` to get the job ID in `maybe_submit_lookup_request()`, then sends blocking `QUERY_PREFETCH_STATUS` polls in `check_lookup_result()`.

- Telemetry event name changed from `"lookup"` to `"lookup_and_prefetch"` since the span now covers both phases.
- Fault tolerance: if a client crashes after `LOOKUP` but never calls `QUERY_PREFETCH_STATUS`, the `_prefetch_jobs` entry leaks. A TODO is left for periodic TTL-based cleanup in a future PR.
- The blend server's `cb_lookup_pre_computed` still has its own internal busy-loop — out of scope for this PR.

**If applicable**:


KuntaiDu (approved):
LGTM!

YaoJiayi (approved):

PR #2713, opened 2026-03-07
[MP] Support MP Server restart

maobaolong:

introduces an automatic re-registration capability for KV caches on the client side and enhances error handling on the server side, ensuring that operations can resume seamlessly after a server interruption. These changes make the system more resilient and reliable in dynamic environments.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -16,6 +16,7 @@ from lmcache.v1.multiprocess.custom_types import (
     CudaIPCWrapper,
     IPCCacheEngineKey,
     KVCache,
+    OperationStatus,
 )
 from lmcache.v1.multiprocess.mq import MessageQueueClient, MessagingFuture
 from lmcache.v1.multiprocess.protocol import RequestType, get_response_class
@@ -90,8 +91,8 @@ class LoadStoreOp:
         return len(self.block_ids)
 
 
-StoreResult = bool
-RetrieveResult = bool
+StoreResult = int
+RetrieveResult = int
 LookupResult = int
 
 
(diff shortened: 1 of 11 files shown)


DongDongJu (approved):
Hello @maobaolong, thanks for the work.
Generally, LGTM.
But I have one concern that need_reregister decide logic is too simple now since retrieve return other failure too. (currently only two return True/False).
But we can handle it next pr.

maobaolong:
@DongDongJu Thanks for your review and good suggestion, i addressed it. 
@chunxiaozheng Would you like to take a look at this PR.

chunxiaozheng (approved):
LGTM!

lmcache/v1/multiprocess/futures.py
```diff
@@ -128,14 +135,19 @@ def wait(self, timeout: Optional[float] = None) -> bool:
             self.event_.synchronize()
             return True
 
+        if self.is_done_.is_set():
+            # Completed without a CUDA event (e.g. server
+            # returned failure with empty handle).
+            return True
+
```
ApostaC, line 142: There is a race condition here.

Thread A (`_on_raw_future_complete`):  run `self.is_done_.set()`
Thread B (`wait`): run `if self.event_:` and found `self.event_ is None`
Thread B (`wait`): run `if self.is_done_.set():` return True
Thread A (`_on_raw_future_complete`): run `self.event_ = torch.cuda.Event.from_ipc_handle(self.device_, event_bytes)` that initializes the real cuda event

This race condition will cause correctness issues or un-needed re-registration which have undefined behavior.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -364,18 +366,28 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
             [
                 self.instance_id,
-                wrap_kv_caches(kv_caches),
+                wrap_kv_caches(self.kv_caches),
                 self.model_name,
                 self.world_size,
             ],
         )
-        future.result()
+        timeout = float(os.getenv("LMCACHE_REGISTER_TIMEOUT", "5.0"))
```
ApostaC, line 388: This default timeout is too short. The LMCache side need to initialize the cuda stream and cupy, which may take ~2 seconds for a single TP worker.

ApostaC (commented):
@maobaolong this one actually incurs a few problems... The default value of LMCACHE_REGISTER_TIMEOUT is too small. And right now, this may have some design conflict with #2692 

Can we revert this PR temporarily? To support server restart, a better solution is to make the mq_client be able to reconnect.

cc @chunxiaozheng @DongDongJu

PR #2722, opened 2026-03-09
[MP][UX][Docs] Enhance http server and its docs for MP mode

ApostaC:

Adds a `POST /api/clear-cache` endpoint to the MP mode HTTP server that force-clears all KV cache data stored in L1 (CPU) memory. Also simplifies the `/api/healthcheck` endpoint by removing the expensive `memcheck()` call — it now only checks that the engine is initialized.

Updates the MP mode documentation (quickstart, architecture, deployment) to cover all HTTP API endpoints including the new `clear-cache`, the upstream `status` endpoint, and the simplified healthcheck.

The `/api/status` endpoint was added upstream — this PR only adds its documentation.

**If applicable**:


YaoJiayi (approved):

Oasis-Git (approved):

ApostaC:
Need to merge #2723 first

PR #2725, opened 2026-03-10
fix(l1_manager): propagate extra_count through prefetch path to prevent premature eviction

liuyumoye:
fix(l1_manager): propagate extra_count through prefetch path to prevent premature eviction

When a prefetch request completes and loaded keys transition from write-locked to read-locked, finish_write_and_reserve_read was always acquiring exactly 1 read lock per key, regardless of the extra_count passed by the caller. This caused premature eviction in Tensor Parallelism (TP > 1) scenarios (e.g. MLA models), where each TP worker needs to consume its own read lock — but only 1 lock was held, so the first worker's finish_read would drop the ref-count to 0 and trigger eviction before the remaining workers could access the data.

This PR propagates extra_count through the entire prefetch path:

StorageManager.submit_prefetch_request → PrefetchController.submit_prefetch_request → _submission_queue / _pending_queue → _start_lookup_phase → InFlightPrefetchRequest.extra_count

On prefetch completion: finish_write_and_reserve_read(loaded_keys, extra_count=request.extra_count) acquires 1 + extra_count read locks per key

Non-prefix loaded keys that are immediately released also pass extra_count to finish_read

L1Manager.finish_write_and_reserve_read now accepts extra_count and calls entry.read_lock.lock() a total of 1 + extra_count times per key

The extra_count semantics are unchanged from the existing submit_prefetch_task / finish_read contract — this PR only ensures the value is correctly threaded through the prefetch pipeline rather than silently dropped.

_pending_queue and _submission_queue tuple types are widened from (..., MemoryLayoutDesc) to (..., MemoryLayoutDesc, int)

No behavior change when extra_count=0 (the default), so existing single-TP deployments are unaffected.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l1_manager.py
+++ b/lmcache/v1/distributed/l1_manager.py
@@ -512,6 +512,7 @@ class L1Manager:
     def finish_write_and_reserve_read(
         self,
         keys: list[ObjectKey],
+        extra_count: int = 0,
     ) -> dict[ObjectKey, L1OperationResult]:
         """Atomically finish write and acquire read lock for the given keys.
 
@@ -522,6 +523,10 @@ class L1Manager:
 
         Args:
             keys: Keys to transition from write-locked to read-locked.
+            extra_count: Extra read locks on top of the default 1 lock.
+                Total locks acquired per key = 1 + extra_count.  Useful
+                when multiple TP workers each consume one read lock for
+                the same key (e.g. MLA models with TP > 1).
 
         Returns:
             A dictionary mapping each object key to a tuple of
@@ -532,6 +537,8 @@ class L1Manager:
             KEY_IN_WRONG_STATE: The key is not write-locked, or it already
                 has read locks.
         """
+        extra_count = _validate_extra_count(extra_count)
+        total = 1 + extra_count
         ret: dict[ObjectKey, L1OperationResult] = {}
         successful_keys: list[ObjectKey] = []
 
@@ -559,7 +566,8 @@ class L1Manager:
                 continue
 
             entry.write_lock.unlock()
-            entry.read_lock.lock()
+            for _ in range(total):
+                entry.read_lock.lock()
             ret[key] = (L1Error.SUCCESS, entry.memory_obj)
             successful_keys.append(key)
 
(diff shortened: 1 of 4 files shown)


ApostaC:
Hey @liuyumoye , can you fix the DCO issue by doing?
```
git commit --amend --signoff
git push -f
```

Thanks!

liuyumoye:
> Hey @liuyumoye , can you fix the DCO issue by doing?
> 
> ```
> git commit --amend --signoff
> git push -f
> ```
> 
> Thanks!

Thanks for the reminder! Fixed the DCO issue and force-pushed.

ApostaC (commented):
Can we add some unit test to test the extra_count path for the prefetch controller?

Otherwise LGTM!

liuyumoye:
> Can we add some unit test to test the extra_count path for the prefetch controller?
> 
> Otherwise LGTM!

Thanks for the review! I've added a `TestExtraCountPrefetch` test class in `tests/v1/distributed/test_prefetch_controller.py` covering the `extra_count` path:

- `test_extra_count_zero_default_behavior` – verifies `extra_count=0` degrades to the original single-lock behavior: 1 lock acquired, 1 `finish_read` fully releases it

- `test_extra_count_one_requires_two_finish_reads` – verifies `extra_count=1` acquires 2 locks; the key remains readable after the 1st `finish_read` and is only released after the 2nd

- `test_extra_count_three_requires_four_finish_reads` – same logic with `extra_count=3` (4 locks total); the key stays readable through 3 `finish_read` calls and is released on the 4th

- `test_extra_count_with_prefix_trim` – verifies that when L2 has a gap, only the contiguous prefix keys get `1 + extra_count` locks; keys beyond the gap are never loaded into L1

- `test_extra_count_non_prefix_loaded_keys_fully_released` – verifies that `_finalize_load` correctly releases all 1 + extra_count locks for keys that were loaded but fall outside the prefix, preventing lock leaks

maobaolong:
@liuyumoye Maybe you have to update your branch to the dev, otherwise, the UT will not passed.

ApostaC (approved):
LGTM!

maobaolong (approved):
LGTM @liuyumoye Thanks for this fix.

PR #2733, opened 2026-03-11
[MP][Bugfix] fix vllm-side lookup logical issue and cuda stream deadlock problem

ApostaC:

Fixes two bugs in the multi-process (MP) mode that together cause hangs:

1. **vLLM-side lookup result caching**: After a prefetch job completes, the server removes it from its job map. Subsequent `QUERY_PREFETCH_STATUS` calls for that job ID return `None` (previously `0`), causing the scheduler to think the lookup is still in progress and poll forever. This PR:
   - Caches finished lookup results on the vLLM side (`_finished_lookup_jobs`) so repeated queries return the cached count instead of re-querying the server.
   - Changes the server to return `None` (instead of `0`) for unknown job IDs, making the semantics consistent: `None` = not ready / unknown, `int` = done.
   - Cleans up cached results in `cleanup_lookup_result`.

2. **CUDA stream deadlock from `launch_host_func`**: Telemetry logging callbacks scheduled via `cupy_stream.launch_host_func` need the GIL, but `paged_kv_transfer` and `lmcache_memcpy_async` hold the GIL while waiting on the CUDA stream — creating a deadlock. This PR adds `py::call_guard<py::gil_scoped_release>()` to both C++ bindings so the GIL is released during execution, allowing host callbacks to acquire it. This also removes the `_can_log_store` workaround that was masking the issue.

- The GIL release guard is safe for these two functions because they operate on GPU memory via CUDA streams and do not touch Python objects.
- The `_can_log_store` flag was a partial workaround that prevented store telemetry until after the first retrieve; it is no longer needed with the proper GIL fix.

**If applicable**:

### DIFF
--- a/.buildkite/scripts/multiprocessing-test/launch-containers.sh
+++ b/.buildkite/scripts/multiprocessing-test/launch-containers.sh
@@ -11,6 +11,7 @@ export VLLM_BASELINE_CONTAINER_NAME="${VLLM_BASELINE_CONTAINER_NAME:-vllm-baseli
 
 # Configuration
 LMCACHE_PORT="${LMCACHE_PORT:-6555}"
+LMCACHE_HTTP_PORT="${LMCACHE_HTTP_PORT:-6556}"
 VLLM_PORT="${VLLM_PORT:-8000}"
 VLLM_BASELINE_PORT="${VLLM_BASELINE_PORT:-9000}"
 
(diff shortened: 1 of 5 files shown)


sammshen (approved):
lgtm

KuntaiDu (commented):
LGTM

KuntaiDu (approved):
L.  G.  T.  M.

PR #2735, opened 2026-03-11
[MP] feat: add LMCacheMPPollingSchedulerAdapter for synchronous lookup and prefetch  in MP mode

liuyumoye:
Currently, the MP-mode prefetch lookup is asynchronous and two-step: the vllm scheduler first calls maybe_submit_lookup_request() to submit a prefetch job, then polls check_lookup_result() in subsequent scheduling rounds to retrieve the result. This introduces scheduling latency when the prefetch result is needed immediately (e.g., for short requests or when the cache hit rate is high).

This PR adds LMCacheMPPollingSchedulerAdapter, a subclass of LMCacheMPSchedulerAdapter that provides a synchronous (blocking) prefetch lookup in MP mode.

The base class check_lookup_result returns None immediately when the prefetch job is not yet complete, requiring the caller to retry across multiple scheduling steps. This new subclass overrides check_lookup_result to poll the server in a loop until the prefetch job finishes, so the caller always gets a definitive matched-token count in a single call.

Key behaviors:

- Polls QUERY_PREFETCH_STATUS at a configurable interval (poll_interval, default 5ms) until the job is done.

- Enforces a configurable timeout (lookup_timeout, default 5s); if exceeded, logs a warning and returns 0 matched tokens so the request falls back to normal inference without KV cache hit.

- Fully backward-compatible: the base class behavior is unchanged; callers opt in by using LMCacheMPPollingSchedulerAdapter instead.

- The polling loop uses time.monotonic() for timeout tracking to avoid wall-clock drift issues.

- poll_interval=0.005 (5ms) and lookup_timeout=5.0 (5s) are chosen as sensible defaults but are fully configurable via constructor arguments.

- On timeout, _lookup_job_ids entry is cleaned up to avoid memory leaks.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Introduces a blocking polling loop around `QUERY_PREFETCH_STATUS`, which can affect scheduler latency and load on the MQ server if misconfigured. Timeout/health handling reduces worst-case impact but correctness depends on exactly-once response semantics and proper timeout tuning.
> 
> **Overview**
> Adds `LMCacheMPPollingSchedulerAdapter`, an opt-in subclass of `LMCacheMPSchedulerAdapter` that makes MP-mode lookup *synchronous*: `check_lookup_result` now blocks and polls until the LMCache prefetch job completes, instead of returning `None` and requiring callers to re-check across scheduling rounds.
> 
> The polling implementation uses a configurable `poll_interval` and overall `lookup_timeout`, cleans up per-request lookup state on timeout/unhealthy server, and carefully reuses a single in-flight MQ future to avoid losing results under the server’s exactly-once semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1a226d9846a283a1d90cfc6a74fd73ab38ac5ad8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from typing import Any
 import os
 import threading
+import time
 
 # Third Party
 import torch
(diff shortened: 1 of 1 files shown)


ApostaC (commented):
Hey @liuyumoye , `lookup_and_prefetch_sync` will block the LMCache server's worker and delay the other `store` and `retrieve` operations. 

Therefore, can we introduce the periodic polling thread in the vllm_multi_process_adapter.py? This can avoid all the code changes in `v1/multiprocess`.

ApostaC:
Additionally, if we are going to have a special `lmcache_mp_connector.py` anyway, it might be a good idea to have the new scheduler adapter (with lookup polling) also implemented in a new class. The new class can inherit from the existing `LMCacheMPSchedulerAdapter` and only overrides the `check_lookup_result` function, IIUC.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -291,17 +184,46 @@ def maybe_submit_lookup_request(
...
         self._lookup_job_ids[request_id] = job_id
 
+    @_lmcache_nvtx_annotate
+    def maybe_submit_lookup_and_prefetch_request(
+        self,
+        request_id: str,
+        token_ids: list[int],
+    ) -> int:
+        """
+        同步版 lookup：提交 prefetch 并阻塞等待结果，直接返回匹配的 token 数。
```
maobaolong: Convent Chinese to English

liuyumoye:
> Additionally, if we are going to have a special `lmcache_mp_connector.py` anyway, it might be a good idea to have the new scheduler adapter (with lookup polling) also implemented in a new class. The new class can inherit from the existing `LMCacheMPSchedulerAdapter` and only overrides the `check_lookup_result` function, IIUC.

Thanks for the suggestion! The current implementation already avoids blocking the server worker — `check_lookup_result` polls `QUERY_PREFETCH_STATUS` on the scheduler adapter side, so the server's worker thread remains free for other operations.

liuyumoye:
> Additionally, if we are going to have a special `lmcache_mp_connector.py` anyway, it might be a good idea to have the new scheduler adapter (with lookup polling) also implemented in a new class. The new class can inherit from the existing `LMCacheMPSchedulerAdapter` and only overrides the `check_lookup_result` function, IIUC.

Great point, and this is exactly what the current implementation does! `LMCacheMPPollingSchedulerAdapter` in `vllm_multi_process_adapter.py` inherits from `LMCacheMPSchedulerAdapter` and only overrides `check_lookup_result` to add the polling-until-done behavior. The `lmcache_mp_connector.py` then uses this subclass on the scheduler side. Please let me know if you'd like any further adjustments to the class structure!

PR #2736, opened 2026-03-11
fix(server): guard finish_read_prefetched behind retrieve_succeeded flag

maobaolong:
When _retrieve_loop raises an exception inside the with-block, the read_prefetched_results context manager already releases the read locks in its finally clause. The retrieve() finally block was unconditionally scheduling a GPU callback to call
finish_read_prefetched, causing a double-free of read locks.

Introduce retrieve_succeeded flag that is only set to True when the with-block exits normally. The GPU callback is now conditional on this flag, forming a defense-in-depth pair with the context manager re-raise fix.

In MPCacheEngine.retrieve(), the finally block unconditionally schedules a GPU callback to call finish_read_prefetched, which releases the read lock. When _retrieve_loop raises an exception, the read_prefetched_results context manager already releases the lock in its own finally clause. However, Python's finally block executes even after a return in an except block, so the GPU callback is still scheduled, causing a read lock double-free.
Fix: Introduce a retrieve_succeeded flag initialized to False, set to True only after the with block exits normally. The GPU callback in finally is now conditional on this flag. This forms a defense-in-depth pair with the context manager re-raise fix (PR #2737) — either fix alone prevents the double-free, but together they provide two independent safety layers.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -218,11 +218,11 @@ class StorageManager:
         try:
             yield good_objs if all_good else None
             successfully_yielded = True
-        except Exception as e:
-            logger.warning(
-                "Exception occurred while processing read prefetched results: %s",
-                str(e),
+        except Exception:
+            logger.exception(
+                "Exception occurred while processing read prefetched results",
             )
+            raise
         finally:
             # Decrease the read lock for all successfully read memory objects
             # if None is yielded or exception occurs during caller's processing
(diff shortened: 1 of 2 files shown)


maobaolong:
@ApostaC Thanks for the review at #2737, I closed it and merge it into this PR.

ApostaC (approved):
LGTM!

ApostaC:
@KuntaiDu Can you take another look?

chunxiaozheng (approved):
LGTM

PR #2737, opened 2026-03-11
fix(storage_manager): re-raise exception in read_prefetched_results

maobaolong:
The context manager was swallowing exceptions raised inside the with-block (e.g. TypeError from _retrieve_loop). This caused the caller to continue as if the retrieve succeeded, leading to a double-free of read locks: once by the context manager's finally clause, and again by the caller's GPU callback.

Add 'raise' to propagate the exception so callers correctly enter their except branch and skip the redundant finish_read_prefetched.

The read_prefetched_results context manager catches exceptions raised inside the with block (e.g., TypeError from _retrieve_loop) but does not re-raise them. Per Python's @contextmanager protocol, this causes the exception to be silently swallowed — the caller continues execution as if the with block succeeded normally.
This leads to a read lock double-free: the context manager's finally clause releases the read lock (correctly), but since the caller doesn't know an exception occurred, it proceeds to schedule a GPU callback that calls finish_read_prefetched again, releasing the same lock a second time. For temporary objects, this drives the lock count to zero and triggers premature deletion.
Fix: Add raise after the warning log in the except block so the exception propagates correctly. The caller then enters its own except branch and skips the redundant finish_read_prefetched call.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -218,11 +218,12 @@ class StorageManager:
         try:
             yield good_objs if all_good else None
             successfully_yielded = True
-        except Exception as e:
+        except Exception:
             logger.warning(
-                "Exception occurred while processing read prefetched results: %s",
-                str(e),
+                "Exception occurred while processing read prefetched results",
+                exc_info=True,
             )
+            raise
         finally:
             # Decrease the read lock for all successfully read memory objects
             # if None is yielded or exception occurs during caller's processing


lmcache/v1/distributed/storage_manager.py
```diff
@@ -218,11 +218,12 @@ def read_prefetched_results(
         try:
             yield good_objs if all_good else None
             successfully_yielded = True
-        except Exception as e:
+        except Exception:
             logger.warning(
```
ApostaC, line 222: We can use `logger.exception` here. I think it will also include the traceback.

lmcache/v1/distributed/storage_manager.py
```diff
@@ -218,11 +218,12 @@ def read_prefetched_results(
...
-                "Exception occurred while processing read prefetched results: %s",
-                str(e),
+                "Exception occurred while processing read prefetched results",
+                exc_info=True,
             )
+            raise
```
ApostaC, line 226: If we raise here, will the `finally` block below still be executed?

ApostaC (commented):
I double-checked the logic here. It's quite complicated due to cuda async execution.

1. For the normal path (i.e., no error happening), the `finish_read_prefetched` MUST be called by `server.py` via `cupy_stream.launch_host_func`, because we don't want the memory objects to be evicted before the CUDA kernels are finished.
2. For the `retrieve` failure path, i.e., memory objects are None or the expected count does not match the requested keys in `retrieve`, the `_retrieve_loop` won't be executed. In this case, we need to have a clear way to define "who should call the `finish_read`.

**There are two potential designs**:
1. Always call `finish_read_prefetched` in `retrieve` via `launch_host_func`, no matter whether the retrieve loop runs or not. In this case, the `finally` block in the context manager can be removed.
2. Only call the `finished_read_prefetched` in`retrieve` via `launch_host_func` if everything runs successfully. Otherwise, we rely on the context manager to free the objects.

Personally, I prefer the second design because when there is something wrong in CUDA (i.e., during retrieve loop), the `cupy_stream.launch_host_func` may have errors as well.


Let me know what you think @maobaolong

ApostaC:
Ah, I see #2736. So if we combine this one and #2736 together, it's basically the "second design" in my above description, right? 

If that's the case, can we do it in the same PR to streamline the merging process?

lmcache/v1/distributed/storage_manager.py
maobaolong: Yeah, finally block will still be executed. but the caller cannot aware this exception.

lmcache/v1/distributed/storage_manager.py
maobaolong: Sure

maobaolong:
@ApostaC Thanks for the review, i'll close this PR and push a new commit to the #2376

PR #2740, opened 2026-03-11
[MP][optimize] optimize evict in lru policy

chunxiaozheng:
For one request, the later keys should be evicted first, for example, the request has `(key1, key2, key3)`, if we first evict `key1`, due to prefix match, `key2` and `key3` will not be hit.

My test result is as follows:
Clipboard_Screenshot_1773227384


maobaolong (approved):
@chunxiaozheng Thanks for this fix, LGTM.
@ApostaC @sammshen Would you like to take a look at this PR? Thanks!

ApostaC (approved):
LGTM!

PR #2741, opened 2026-03-11
[MP] introduce server auditlog

maobaolong:

```
[2026-03-11 20:33:23,829] LMCache INFO: AUDIT {'request_type': 'NOOP', 'source': 'worker-0-12346', 'success': True, 'elapsed_ms': 0.002, 'result': "'OK'"} (audit.py:153:lmcache.audit)
[2026-03-11 20:33:23,830] LMCache INFO: AUDIT {'request_type': 'GET_CHUNK_SIZE', 'source': 'worker-0-12346', 'success': True, 'elapsed_ms': 0.003, 'result': '256'} (audit.py:153:lmcache.audit)
```

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -118,7 +118,8 @@ class LMCacheMPSchedulerAdapter:
             tp_size: Tensor-parallel size for MLA
                 multi-reader locking (default 1).
         """
-        self.mq_client = MessageQueueClient(server_url, context)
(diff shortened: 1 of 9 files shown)


ApostaC (commented):
The zmq identity part looks good at a high level.

We don't need to introduce the new audit ctx, instead, we can reuse the existing telemetry subsystem (introduced #2696 )
Also, I feel like we don't need `SET_AUDIT_ENABLED` command. It should be the enabled by the config (i.e., from the terminal) rather than something sent by the vLLM engine.

PR #2744, opened 2026-03-11
[MP][Bugfix] introducing new l1 listener to prevent re-storing prefetched object

ApostaC:
Inspired by PR: #2738

Previously, `L1Manager.finish_write_and_reserve_read()` (used by prefetch) fired two separate listener callbacks: `on_l1_keys_write_finished` + `on_l1_keys_reserved_read`. The `on_l1_keys_write_finished` callback caused `StoreListener` to enqueue prefetched keys for L2 store — re-storing objects that were just fetched *from* L2.

This PR introduces a dedicated `on_l1_keys_finish_write_and_reserve_read` callback in `L1ManagerListener` so that each subclass can handle the prefetch transition appropriately:

- **StoreListener**: no-op (don't re-store prefetched data back to L2)
- **EvictionPolicy**: registers keys as created (tracks them for eviction)
- **L1ManagerStatsLogger**: counts as write completion (for Prometheus metrics)

The L1Manager now fires a single `on_l1_keys_finish_write_and_reserve_read` callback instead of separate `on_l1_keys_write_finished` + `on_l1_keys_reserved_read` in `finish_write_and_reserve_read()`.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/eviction.py
+++ b/lmcache/v1/distributed/eviction.py
@@ -108,3 +108,6 @@ class EvictionPolicy(L1ManagerListener):
 
     def on_l1_keys_deleted_by_manager(self, keys: list[ObjectKey]):
         self.on_keys_removed(keys)
+
+    def on_l1_keys_finish_write_and_reserve_read(self, keys: list[ObjectKey]):
+        self.on_keys_created(keys)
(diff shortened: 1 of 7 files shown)


ApostaC:
@maobaolong Can you take a look at this? Thanks!

maobaolong (approved):
@ApostaC This is a better fix than #2738 . LGTM and hope to merge this.

maobaolong:
@chunxiaozheng Would you like to take another look?

chunxiaozheng (approved):
Thanks for the fix, LGTM

PR #2749, opened 2026-03-11
[Fix][Operator] Add privileged mode and nvidia runtime for GPU visibility

royyhuang:
## Summary
- Add `runtimeClassName: nvidia` and `privileged: true` to the DaemonSet pod spec so LMCache pods get GPU visibility via the NVIDIA container runtime without claiming GPUs through `nvidia.com/gpu` resource requests
- Add `NVIDIA_DRIVER_CAPABILITIES=all` env var alongside existing `NVIDIA_VISIBLE_DEVICES=all`
- Update DESIGN.md and README.md with the new GPU visibility requirements and security implications
- **[CI] Use job-level path filtering to unblock operator-only PRs**: Move path filtering from workflow-level `paths:`/`paths-ignore:` triggers to job-level `if:` conditions using `dorny/paths-filter`. When jobs are skipped via `if:`, GitHub reports them as "Success" — unlike workflow-level `paths:` which causes required checks to never report, permanently blocking PRs that don't touch relevant files.

## Why
LMCache needs access to all GPUs on the node for CUDA IPC and custom data transfer kernels. However, requesting GPUs via the device plugin (`nvidia.com/gpu`) would exclusively claim them, making them unavailable to the serving engine (e.g., vLLM). The combination of `runtimeClassName: nvidia` + `privileged: true` + NVIDIA env vars allows full GPU access without consuming device plugin resources.

## CI workflow changes
- `test.yml`: Always triggers on PRs, but skips the test matrix when no Python files (`**.py`, `pyproject.toml`, `requirements/**.txt`) changed
- `code_quality_checks.yml`: Always triggers on PRs, but skips pre-commit checks when only `operator/` files changed
- Uses `dorny/paths-filter@v3` for path detection at the job level
- Jobs skipped via `if:` report as "Success" to GitHub, so required status checks pass without running unnecessary CI

## Test plan
- [x] Verified `nvidia-smi` shows all 8 GPUs in a test pod with this configuration (no `nvidia.com/gpu` request)
- [x] Verified the operator builds cleanly (`go build ./...`)
- [x] Deploy LMCacheEngine CR on a cluster with NVIDIA GPU Operator and verify pods start successfully
- [x] Verify vLLM serving engine can still claim all GPUs on the same node
- [ ] Verify CI workflows trigger correctly on this PR and required checks pass


ruizhang0101 (approved):
LGTM :)))

sammshen (approved):
LGTM!

ApostaC (approved):
LGTM!

PR #2759, opened 2026-03-12
[MP][bugfix] fix --l1-use-lazy is always True

chunxiaozheng:
currently, if we set `--l1-use-lazy`, `l1_use_lazy` is `True`, if we not set, the default value still is `True`.


chunxiaozheng:

PR #2761, opened 2026-03-12
Fix[config]: replace store_true with BooleanOptionalAction for --l1-use-lazy

liuyumoye:
fix(config): replace store_true with BooleanOptionalAction for --l1-use-lazy

Use argparse.BooleanOptionalAction instead of action='store_true' so that --no-l1-use-lazy can be passed on the command line to explicitly disable lazy loading for L1 memory.

### DIFF
--- a/docs/source/mp/configuration.rst
+++ b/docs/source/mp/configuration.rst
@@ -79,9 +79,11 @@ Source: ``lmcache/v1/distributed/config.py``
    * - ``--l1-size-gb``
      - *required*
      - Size of L1 memory in GB.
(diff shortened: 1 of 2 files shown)


ApostaC (commented):
Can we update the docs as well? There is a configuration page for mp mode that talks about the cli args: https://docs.lmcache.ai/mp/configuration.html

maobaolong (approved):
LGTM. thanks for this improvement.

chunxiaozheng (approved):
LGTM!

PR #2763, opened 2026-03-13
[MP][Feat] Support dedicated thread pool for MP callbacks

ApostaC:

Adds `MessageQueueServer.add_dedicated_thread_pool(request_types, max_workers)` so that specific blocking request types can run on their own `ThreadPoolExecutor` instead of sharing the single default pool. This prevents slow/blocking handlers (e.g., STORE, RETRIEVE) from starving other request types when the default pool is exhausted.

Usage (after handler registration, before `server.start()`):
```python
server.add_dedicated_thread_pool([RequestType.STORE, RequestType.RETRIEVE], max_workers=4)
```

Multiple dedicated pools are supported. The method validates that all specified request types are registered `BlockingRequestHandler`s before creating the pool (two-pass for atomicity).

Minimal changes to existing code — only added one field to `__init__`, one guard in `close()`, and pool shutdown logic. The new method reassigns the `executor` attribute on already-registered `BlockingRequestHandler` instances. No existing methods were modified.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/mq.py
+++ b/lmcache/v1/multiprocess/mq.py
@@ -641,6 +641,9 @@ class MessageQueueServer:
                 )
 
         # Pass 2: create pool and assign
+        if not request_types:
+            return
+
         pool = ThreadPoolExecutor(
             max_workers=max_workers,
             thread_name_prefix=f"dedicated-pool-{len(self.dedicated_pools)}",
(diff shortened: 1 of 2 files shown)


KuntaiDu (approved):
LGTM!

Oasis-Git (approved):

Issue #1882, opened 2025-10-26
[Onboarding][Q4] Welcoming contributors with good first issues!


kobe0938:
# Welcome to Q4, LMCache Contributors!

We’re thrilled to have you here!
LMCache is growing fast — and your ideas, code, and insights keep this community thriving.
Pick a task, comment to claim it, and dive in.

ps. this board is a successor of https://github.com/LMCache/LMCache/issues/627

Make sure to claim & later link the pr to the issue so that we know that you are working on this.

---

### Documentation

* [ ] FAQ + Troubleshoot https://docs.lmcache.ai/getting_started/faq.html (slack in the past 2 months); WIP: https://github.com/LMCache/LMCache/pull/1879
* [ ] Cookbook for differnt models https://docs.sglang.ai/basic_usage/gpt_oss.html (Qwen, Llama, GPT-OSS) WIP: https://github.com/LMCache/LMCache/pull/1825
* [ ] Chinese doc -> add pipeline automation for auto translation; WIP: https://github.com/LMCache/LMCache/pull/1772
* [ ] Rerun `more examples` section; Owned by Loki
* [ ] Hardware section for AMD, XPU(PR + Jiayi), Acsend(Huawei Martin); Owned by Loki
* [x] Modify bug report template https://github.com/LMCache/LMCache/pull/1906
* [ ] Fix kv cache calculator, add more models like qwen3.5 next oss; Owned by Kobe
* [x] Infinistore backend; WIP: https://github.com/LMCache/LMCache/pull/1783#pullrequestreview-3406137550
* [x] https://github.com/LMCache/LMCache/issues/1940; Owned by Kobe

---

## Bugs in the past month (including replying/closing issues so that bug is fixed/issue is closed) 

* [x] https://github.com/LMCache/LMCache/issues/1901
* [ ] https://github.com/LMCache/LMCache/issues/1889
* [ ] https://github.com/LMCache/LMCache/issues/1881
* [ ] https://github.com/LMCache/LMCache/issues/1865
* [ ] https://github.com/LMCache/LMCache/issues/1864
* [ ] https://github.com/LMCache/LMCache/issues/1861
* [ ] https://github.com/LMCache/LMCache/issues/1858 WIP by: @Martin-HZK
* [x] https://github.com/LMCache/LMCache/issues/1856
* [ ] https://github.com/LMCache/LMCache/issues/1802
* [ ] https://github.com/LMCache/LMCache/issues/1799
[...]

PR #2766, opened 2026-03-13
refactor: move busy loop from LocalCPUBackend to MixedMemoryAllocator

Dayuxiaoshui:

- Move the busy loop (wait + retry on allocation failure) from `LocalCPUBackend` into `MixedMemoryAllocator`: add `allocate_with_retry` and `batched_allocate_with_retry` so the allocator owns the wait/retry policy when `busy_loop=True`.
- `LocalCPUBackend.allocate` / `batched_allocate` now call the allocator retry API and no longer use `time.sleep`; the eviction loop (get evict candidates, evict, then allocate) stays in the backend.

Refs #1882 (Good First Issue: move busy loop from backend to allocator).

- Behaviour unchanged: one failed allocate → sleep(0.1) → one retry is now inside the allocator; backend still does eviction then one “allocate attempt” (which may include that retry).

## If applicable

- [ ] This PR contains user facing changes - docs added
- [x] This PR contains unit tests (existing `tests/v1/storage_backend/test_local_cpu_backend.py` covers the refactor; no new tests added)

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -10,6 +10,7 @@ import abc
 import ctypes
 import os
 import threading
+import time
 
 # Third Party
 from sortedcontainers import SortedList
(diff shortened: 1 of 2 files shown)


Dayuxiaoshui:
@kobe0938

PR #2767, opened 2026-03-13
Support non-contiguous alloc in MemoryAllocator

chunxiaozheng:
Support non-contiguous alloc in MemoryAllocator.

As the running time increases, the last block of each request will first be evicted each time, which will result in the inability to allocate large contiguous space in the future.

### DIFF
--- a/lmcache/v1/distributed/memory_manager.py
+++ b/lmcache/v1/distributed/memory_manager.py
@@ -28,13 +28,32 @@ def create_memory_allocator(config: L1MemoryManagerConfig) -> MemoryAllocatorInt
         MemoryAllocatorInterface: An instance of a memory allocator.
     """
     if config.use_lazy:
+        logger.info(
+            "use lazy memory allocator, init size is %d bytes, "
+            "final size is %d bytes, align bytes is %d bytes, contiguous alloc is %s",
+            config.init_size_in_bytes,
+            config.size_in_bytes,
+            config.align_bytes,
+            config.contiguous_alloc,
+        )
         return LazyMemoryAllocator(
-            config.init_size_in_bytes, config.size_in_bytes, config.align_bytes
+            init_size=config.init_size_in_bytes,
+            final_size=config.size_in_bytes,
+            align_bytes=config.align_bytes,
+            contiguous_alloc=config.contiguous_alloc,
         )
     else:
+        logger.info(
+            "use mixed memory allocator, total size is %d bytes, "
+            "align bytes is %d bytes, contiguous alloc is %s",
+            config.size_in_bytes,
+            config.align_bytes,
+            config.contiguous_alloc,
+        )
         return MixedMemoryAllocator(
             config.size_in_bytes,
             align_bytes=config.align_bytes,
+            contiguous_alloc=config.contiguous_alloc,
         )
 
 
--- a/lmcache/v1/distributed/memory_manager.py
+++ b/lmcache/v1/distributed/memory_manager.py
@@ -34,13 +34,11 @@ def create_memory_allocator(config: L1MemoryManagerConfig) -> MemoryAllocatorInt
             config.init_size_in_bytes,
             config.size_in_bytes,
             config.align_bytes,
-            config.contiguous_alloc,
         )
         return LazyMemoryAllocator(
             init_size=config.init_size_in_bytes,
             final_size=config.size_in_bytes,
             align_bytes=config.align_bytes,
-            contiguous_alloc=config.contiguous_alloc,
         )
     else:
         logger.info(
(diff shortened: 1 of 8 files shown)


maobaolong (approved):
LGTM, Thanks for the fix.

chunxiaozheng:
@ApostaC could you help take a look?

lmcache/v1/memory_management.py
```diff
@@ -1277,6 +1283,21 @@ def batched_allocate(
...
+            for i in range(batch_size):
+                tensor_mem_obj = self.allocate(shapes, dtypes, fmt, allocator_type)
+                if tensor_mem_obj is None:
+                    # if any of the allocation failed,
+                    # free all the allocated memory objs and return None
+                    self.batched_free(tensor_mem_objs)
+                    return None
+                else:
+                    tensor_mem_objs.append(tensor_mem_obj)
+            return tensor_mem_objs
```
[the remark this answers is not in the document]
chunxiaozheng: Good idea! I have updated!

lmcache/v1/memory_management.py
```diff
@@ -1090,6 +1090,83 @@ def allocate(self, size: int) -> tuple[int, int]:
...
+
+        Args:
+            size: The requested size of the memory block. Should be greater
+                than 0.
+            batch_size: The number of memory blocks to allocate.
+
+        Returns:
+            A list of tuple (address, allocated_size) where address is the starting
+            address of the allocated block and allocated_size is the actual
+            size of the allocated block.
```
ApostaC: nit: we need to mention in the docstring that the length of the return list is the same as the `batch_size` argument.

lmcache/v1/memory_management.py
```diff
@@ -1277,18 +1360,41 @@ def batched_allocate(
...
+            try:
+                alloc_results = self.address_manager.batched_allocate(
+                    unit_aligned_size, batch_size
+                )
+            except RuntimeError:
+                return None
+            addresses = [addr for addr, _ in alloc_results]
+            raw_datas = [
+                self._get_buffer_slice(addr, unit_aligned_size) for addr in addresses
+            ]
```
ApostaC: Let's just use this, and we can discard the old "contiguous" mode so that the code will be much simpler.

ApostaC (changes requested):
I like the implementation in the AddressManager. My proposal is to have it become the default and completely discard the previous "contiguous" mode. In this case, we don't need to change the config and CLI, and don't need to pass down the new arguments all the way down to the address manager.

Additionally, please add UTs for the `AddressManager.batched_allocate`.

lmcache/v1/distributed/memory_manager.py
```diff
@@ -28,10 +28,23 @@ def create_memory_allocator(config: L1MemoryManagerConfig) -> MemoryAllocatorInt
         MemoryAllocatorInterface: An instance of a memory allocator.
     """
     if config.use_lazy:
+        logger.info(
+            "use lazy memory allocator, init size is %d bytes, "
+            "final size is %d bytes, align bytes is %d bytes, contiguous alloc is %s",
+            config.init_size_in_bytes,
+            config.size_in_bytes,
+            config.align_bytes,
+        )
```
ApostaC:
```suggestion
        logger.debug(
            "use lazy memory allocator, init size is %d bytes, "
            "final size is %d bytes, align bytes is %d bytes",
            config.init_size_in_bytes,
            config.size_in_bytes,
            config.align_bytes,
        )
```

lmcache/v1/distributed/memory_manager.py
```diff
@@ -28,10 +28,23 @@ def create_memory_allocator(config: L1MemoryManagerConfig) -> MemoryAllocatorInt
...
         return LazyMemoryAllocator(
             config.init_size_in_bytes, config.size_in_bytes, config.align_bytes
         )
     else:
+        logger.info(
+            "use mixed memory allocator, total size is %d bytes, "
+            "align bytes is %d bytes, contiguous alloc is %s",
+            config.size_in_bytes,
+            config.align_bytes,
+        )
```
ApostaC:
```suggestion
        logger.debug(
            "use mixed memory allocator, total size is %d bytes, "
            "align bytes is %d bytes",
            config.size_in_bytes,
            config.align_bytes,
        )
```

ApostaC (commented):
Please fix the issue in the logger output. Otherwise LGTM!

chunxiaozheng:
@ApostaC Thanks for your review, I have updated, could you help take a look again?

ApostaC (approved):
LGTM!

PR #2768, opened 2026-03-13
[MP][Bugfix] fix free error when memory_objs is empty

chunxiaozheng:
fix this error when not use `lazy_memory_allocator` in mp mode.
Clipboard_Screenshot_1773392685

In mp mode, if use `MixedMemoryAllocator`, when trigger write to L2, will first `reserve_read` L1 objects and then `finish_read`, `finish_read` will call `self._memory_manager.free(need_to_free)`, but all `mem_objs` is not `temp`, and `need_to_allocate` is empty, this will trigger this bug.


maobaolong (approved):
LGTM. This is an obvious bug, thanks for this fix.

ApostaC (approved):
LGTM!

PR #2771, opened 2026-03-13
fix the torch ABI mismatch in docker image build 

deng451e:
Ensure lmcache installation uses the locally built wheel during image build

**If applicable**:


sammshen (approved):
LGTM!!

ApostaC (changes requested):
Let's not modify the release docker image. The "release" is expected to use the pypi image.

We should make sure that the cibuildwheel uses the correct torch version, and people won't have errors when they do `pip install lmache`  from their bare metal environments 

cc @sammshen

PR #2772, opened 2026-03-14
[BugFix] Remove unused memory cleanup logic to prevent premature free

matthewygf:
- Currently, the async lookup path will call ref_count_down() during process_tokens_internal for the unused mem obj, then lookup_unpin() will call ref_count_down() during the cleanup.
- This lead to the mem_obj in the local_cpu_backend being invalidated but still remain in hotcache.
- The **next request** that hits the hot cache will obtain this invalidated mem_obj and hit the assertion error in the gpu_connector.
- Removing the unused only should be fine, as the clean_up will bulk remove the unused anyway.

- I can consistently reproduce this after running the vllm benchmark, and this only manifest after the benchmark has progressed for a while.

**If applicable**:

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1553,7 +1553,6 @@ class LMCacheEngine:
                 memory_obj_map[key] = memory_obj
 
         # TODO(Jiayi): hashing inside `process_tokens` can be skipped.
-        used_keys: set[CacheEngineKey] = set()
         for start, end, key in self.token_database.process_tokens(
             tokens=tokens,
             mask=mask,
@@ -1568,7 +1567,6 @@ class LMCacheEngine:
             chunks.append((key, memory_obj, start, end))
             tot_kv_size += memory_obj.get_size()
             ret_mask[start:end] = True
-            used_keys.add(key)
 
         return chunks, tot_kv_size
 
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -1570,11 +1570,6 @@ class LMCacheEngine:
             ret_mask[start:end] = True
             used_keys.add(key)
 
-        # NOTE: free the memory objects that are not hit.
-        for key, mem_obj in memory_obj_map.items():
-            if key not in used_keys:
-                mem_obj.ref_count_down()
-
         return chunks, tot_kv_size
 
     def _process_tokens_internal(


matthewygf:
cc @deng451e

deng451e:
Thanks for the fix! Could you clarify which assertion in gpu_connector is being triggered? It would also be helpful to share a minimal error log when the issue occurs. Theoretically, we should not see a cache hit on an invalidated memory_obj.

matthewygf:
> Thanks for the fix! Could you clarify which assertion in gpu_connector is being triggered? It would also be helpful to share a minimal error log when the issue occurs. Theoretically, we should not see a cache hit on an invalidated memory_obj.

Thanks for the msg @deng451e  ! So the assertion error is triggered during the multi-layer retrieve.

The ref_counts are as follow:
- batched_async_contains(pin=True)  →  pin()      →  ref=1, pin=1
- batched_get_non_blocking()        →  ref_count_up() →  ref=2, pin=1
- For unused keys: mem_obj.ref_count_down()   →  ref=1, pin=1
- For ALL objects (used AND unused):
- unpin()         →  pin: 1→0
- ref_count_down()
Then:                  
- USED objects:   ref 2→1, pin 0  →  alive in hot_cache 
- UNUSED objects: ref 1→0, pin 0  →  BOTH ZERO → free() → invalidate() , but also in hot_cache.

For next request:
- prefix-cache hit again for unused objects
- trigger the invalidate 
assertion_error

Apologies, there are also a few things I should clarify that might help identifying the issues:
1. I am using an older version of LMCache (v0.3.12) with back ported classes -i.e. CacheEngine , MemoryManagement, StorageManager, LocalCPUBackend. (However, no async process nor memory management logic is being changed except some observabilities functionality).
2. I am using the async processing mode with remote backend like Mooncake.
3. I am running on Ascend Platform - which have no changes to the kvconnector behavior.

I believe the current fix would result in no change for the ref counts.

deng451e:
> > Thanks for the fix! Could you clarify which assertion in gpu_connector is being triggered? It would also be helpful to share a minimal error log when the issue occurs. Theoretically, we should not see a cache hit on an invalidated memory_obj.
> 
> Thanks for the msg @deng451e ! So the assertion error is triggered during the multi-layer retrieve.
> 
> The ref_counts are as follow:
> 
> * batched_async_contains(pin=True)  →  pin()      →  ref=1, pin=1
> * batched_get_non_blocking()        →  ref_count_up() →  ref=2, pin=1
> * For unused keys: mem_obj.ref_count_down()   →  ref=1, pin=1
> * For ALL objects (used AND unused):
> * unpin()         →  pin: 1→0
> * ref_count_down()
>   Then:
> * USED objects:   ref 2→1, pin 0  →  alive in hot_cache
> * UNUSED objects: ref 1→0, pin 0  →  BOTH ZERO → free() → invalidate() , but also in hot_cache.
> 
> For next request:
> 
> * prefix-cache hit again for unused objects
> * trigger the invalidate
> 
> assertion_error
> Apologies, there are also a few things I should clarify that might help identifying the issues:
> 
> 1. I am using an older version of LMCache (v0.3.12) with back ported classes -i.e. CacheEngine , MemoryManagement, StorageManager, LocalCPUBackend. (However, no async process nor memory management logic is being changed except some observabilities functionality).
> 2. I am using the async processing mode with remote backend like Mooncake.
> 3. I am running on Ascend Platform - which have no changes to the kvconnector behavior.
> 
> I believe the current fix would result in no change for the ref counts.

To my understanding, a subsequent lookup should not hit self.hot_cache in local_cpu_backend, since the key and memory_obj should be released together.
In LMCache v0.3.12, remote connectors such as Mooncake use the CPU allocator only for staging buffers, and those keys are not tracked by hot_cache.
Could you add some logs to pinpoint which backend is returning the invalidated memory_obj in your case — local_cpu_backend or mooncake_store?

deng451e:
@sammshen PTAL — correct me if I’m wrong.

PR #2784, opened 2026-03-15
Add Python fallback for native storage ops

darsh7807:
## Summary
- provide a pure-Python implementation of TTLLock/Bitmap/pattern matchers when the native extension is absent, enabling CPU-only/distributed tests to import lmcache.native_storage_ops
- keep the compiled extension as the preferred path; fallback is for correctness in non-CUDA environments

## Testing
- not run (CPU-only mac; change is import fallback)

### DIFF
--- a/lmcache/native_storage_ops.py
+++ b/lmcache/native_storage_ops.py
@@ -103,13 +103,17 @@ class Bitmap:
     def __and__(self, other: "Bitmap") -> "Bitmap":
         size = min(len(self._bits), len(other._bits))
         out = Bitmap(size)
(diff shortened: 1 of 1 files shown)


maobaolong (commented):
@darsh7807 Thanks for this work!

Please take a look at the gemini's comments.

darsh7807:
Applied Gemini feedback: TTLLock now uses monotonic time and avoids decrement after expiry; RangePatternMatcher.match now returns (start, end) tuples (end-exclusive) per spec. Updated branch pushed.

maobaolong:
@darsh7807 Hey Thanks for the contribution again! This PR is useful for some case! Would you like to fix the DCO issue ?

Issue #842, opened 2025-06-16
Supply LMCacheConnectorV1 to vLLM by dynamic loading to avoid code conflict


maobaolong:
**Is your feature request related to a problem? Please describe.**

Powered by https://github.com/vllm-project/vllm/pull/18142, supply LMCacheConnector to vLLM by dynamic loading to avoid code conflict.

**Describe the solution you'd like**

- Move LMCacheConnectorV1 from vLLM to LMCache repository, instead, write a document to show how to use LMCacheConnectorV1 by dynamic loading way.
- Or copy LMCacheConnectorV1 from vLLM to LMCache repository.
- Write a document to use LMCacheConnectorV1 by dynamic loading way.

# How to test
- start vllm
```console
LMCACHE_USE_EXPERIMENTAL=True LMCACHE_TRACK_USAGE=false \
VLLM_MLA_DISABLE=0 VLLM_USE_V1=1 LMCACHE_CONFIG_FILE=./lmcache.yaml \
vllm serve /disc/data1/deepseek/DeepSeek-V2-Lite-Chat/ \
           --trust-remote-code \
           --served-model-name vllm_cpu_offload \
           --max-model-len 32768 \
           --max-seq-len-to-capture 10000 \
           --max-num-seqs 64 \
           --gpu-memory-utilization 0.9 \
           --host 0.0.0.0 \
           -tp 2 \
           --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1Dy","kv_role":"kv_both","kv_parallel_size":2, "kv_connector_module_path":"lmcache.integration.vllm.lmcache_connector_v1"}'
```
- vllm log
```
VllmWorker rank=0 pid=146787) INFO 06-16 15:49:51 [factory.py:74] Creating v1 connector with name: LMCacheConnectorV1Dy and engine_id: f1f81d1e-c6f6-4dd5-a3ed-2fc1eb8489c5
(VllmWorker rank=0 pid=146787) WARNING 06-16 15:49:51 [base.py:71] Initializing KVConnectorBase_V1. This API is experimental and subject to change in the future as we iterate the design.
(VllmWorker rank=0 pid=146787) [2025-06-16 15:49:51,111] LMCache INFO: Loading LMCache config file ./lmcache.yaml (utils.py:64:lmcache.integration.vllm.utils)
[...]

PR #2786, opened 2026-03-16
[Bugfix]: Fix pin count balancing in PD Disaggregation mode

lisiG9:
# Summary
This PR fixes a pin count balancing issue in PD (Prefill-Decode) Disaggregation mode. When using LMCache with vLLM's PD Disaggregation (via `kv_producer`/`kv_consumer` roles), memory objects in `LocalCPUBackend` were being pinned indefinitely on the Decoder (D) side, causing pin timeout warnings and potential memory leaks. The Prefill (P) side previously handled pin count balancing correctly, but the Decode (D) side lacked the necessary unpin operation.

# Problem Description
In PD Disaggregation mode:
- **Prefill (P) side (`kv_producer`)**: Performs lookup with `pin=True`, saves KV cache, and calls `lookup_unpin()` in `wait_for_save()` to balance the pin count.
- **Decode (D) side (`kv_consumer`)**: Performs lookup with `pin=True`, but `wait_for_save()` was returning early without calling `lookup_unpin()`, leaving the pin count unbalanced.

This caused memory objects to remain pinned indefinitely on the D-side, leading to:
- Pin timeout warnings from the pin monitor
- Reduced effectiveness of the cache eviction policy
- Increased memory pressure and potential out-of-memory situations
- Performance degradation due to suboptimal cache management

# Solution
Modified `LMCacheConnectorV1Impl.wait_for_save()` in `vllm_v1_adapter.py` to call `lookup_unpin()` for both P-side and D-side:

```python
if self.kv_role == "kv_consumer":
    # Don't do save if the role is kv_consumer
    # But still need to unpin the kv caches according to req_id
    # to balance the pin count from contains()
    for request in connector_metadata.requests:
        self.lmcache_engine.lookup_unpin(request.req_id)
    return
```

This ensures that both P-side and D-side properly balance the pin count incremented during `contains(pin=True)` calls.

# Impact Analysis
- **Scope**: This fix affects PD Disaggregation mode using vLLM's `kv_producer`/`kv_consumer` roles
- **Backward Compatibility**: No breaking changes - the fix only adds the missing `lookup_unpin()` call on D-side
- **Performance**: Minimal overhead - just an additional loop to call `lookup_unpin()` for each request

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1660,11 +1660,16 @@ class LMCacheConnectorV1Impl:
             self._layerwise_save_storers.pop(request.request_id, None)
 
         # Cleanup if request was aborted
-        if request.status == RequestStatus.FINISHED_ABORTED and self.async_loading:
-            # Cancel any ongoing async lookup and prefetch tasks on workers
-            lookup_id = request.request_id
-            assert self.lookup_client is not None
-            self.lookup_client.cancel_lookup(lookup_id)  # type: ignore[attr-defined]
+        if request.status == RequestStatus.FINISHED_ABORTED:
+            if self.async_loading:
+                # Cancel any ongoing async lookup and prefetch tasks on workers
+                lookup_id = request.request_id
+                assert self.lookup_client is not None
+                self.lookup_client.cancel_lookup(lookup_id)  # type: ignore[attr-defined]
+
+            # Unpin to balance pin count from lookup
+            if self.lmcache_engine is not None:
+                self.lmcache_engine.lookup_unpin(request.request_id)
 
         params = (
             request.kv_transfer_params
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1091,8 +1091,7 @@ class LMCacheConnectorV1Impl:
                 "LMCacheEngine must be initialized to unpin requests."
             )
             for request in connector_metadata.requests:
-                if request.status != RequestStatus.FINISHED_ABORTED:
-                    self.lmcache_engine.lookup_unpin(request.req_id)
+                self.lmcache_engine.lookup_unpin(request.req_id)
 
             return
 
@@ -1661,16 +1660,11 @@ class LMCacheConnectorV1Impl:
             self._layerwise_save_storers.pop(request.request_id, None)
 
         # Cleanup if request was aborted
-        if request.status == RequestStatus.FINISHED_ABORTED:
-            if self.async_loading:
-                # Cancel any ongoing async lookup and prefetch tasks on workers
-                lookup_id = request.request_id
-                assert self.lookup_client is not None
-                self.lookup_client.cancel_lookup(lookup_id)  # type: ignore[attr-defined]
-
-            # Unpin to balance pin count from lookup
-            if self.lmcache_engine is not None:
-                self.lmcache_engine.lookup_unpin(request.request_id)
+        if request.status == RequestStatus.FINISHED_ABORTED and self.async_loading:
+            # Cancel any ongoing async lookup and prefetch tasks on workers
+            lookup_id = request.request_id
+            assert self.lookup_client is not None
+            self.lookup_client.cancel_lookup(lookup_id)  # type: ignore[attr-defined]
 
         params = (
             request.kv_transfer_params
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1085,6 +1085,10 @@ class LMCacheConnectorV1Impl:
 
         if self.kv_role == "kv_consumer":
             # Don't do save if the role is kv_consumer
+            # But still need to unpin the kv caches according to req_id
+            # to balance the pin count from contains()
+            for request in connector_metadata.requests:
+                self.lmcache_engine.lookup_unpin(request.req_id)
             return
 
         if self.use_layerwise:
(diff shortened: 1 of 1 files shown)


lisiG9:
@sammshen  @YaoJiayi  please take a look, thanks!

deng451e:
In LMCache PD in-process mode, the prefiller writes KV cache directly to the decoder GPU buffer without going through local_cpu_backend, so it shouldn’t require pinning memory_obj on the decoder host side.
Could you share the config and script you used to launch PD? That would help us better reproduce the memory leak you observed.

lisiG9:
> In LMCache PD in-process mode, the prefiller writes KV cache directly to the decoder GPU buffer without going through local_cpu_backend, so it shouldn’t require pinning memory_obj on the decoder host side. Could you share the config and script you used to launch PD? That would help us better reproduce the memory leak you observed.

Thanks for the question! I'm using **MooncakeStore** as the remote storage backend, not the in-process mode.

In MooncakeStore mode, the decoder side calls `contains()` with `pin=True` which triggers `local_cpu_backend.pin()`, but `wait_for_save()` returns early for `kv_consumer` without calling `lookup_unpin()`.

**Config (`LMCACHE_CONFIG_FILE`):**
```yaml
chunk_size: 256
remote_url: "mooncakestore://80.11.128.72:50011/"
local_cpu: True
max_local_cpu_size: 10
use_layerwise: False
save_decode_cache: False
enable_async_loading: False
pre_caching_hash_algorithm: sha256_cbor_64bit
extra_config:
  local_hostname: "80.11.128.72"
  protocol: "tcp"
  metadata_server: "http://80.11.128.72:8082/metadata"
  master_server_address: "80.11.128.72:50011"
  global_segment_size: 53687091200
  local_buffer_size: 0
  save_chunk_meta: False
  use_exists_sync: True
```

**Launch commands:**
```bash
export LMCACHE_CONFIG_FILE="$LMCACHE_CONFIG_FILE"
KV_TRANSFER_CONFIG=$(cat <<EOF
{"kv_connector":"LMCacheConnectorV1","kv_role":"$KV_ROLE_VALUE","kv_connector_extra_config":{"discard_partial_chunks":false,"lmcache_rpc_port":"$LMCACHE_RPC_PORT_NAME"}}
EOF
)
# --- 启动服务 ---
echo "--- 正在启动 $ROLE 服务 (TP=$TENSOR_PARALLEL_SIZE) ---"
python3 -m vllm.entrypoints.openai.api_server \
--model "$MODEL_PATH" \
--port "$PORT" \
--tensor-parallel-size "$TENSOR_PARALLEL_SIZE" \
"${COMMON_ARGS[@]}" \
--kv-transfer-config "$KV_TRANSFER_CONFIG"
```

deng451e:
According to your description, the KV cache flow seems to be prefiller → mooncakestore → decoder — hope I got that right.
A similar setup exists in LMCache multi-process mode:
https://github.com/LMCache/LMCache/tree/dev/examples/disagg_prefill_mp — feel free to try it. It should also help address the memory leak you observed.

lisiG9:
> According to your description, the KV cache flow seems to be prefiller → mooncakestore → decoder — hope I got that right. A similar setup exists in LMCache multi-process mode: https://github.com/LMCache/LMCache/tree/dev/examples/disagg_prefill_mp — feel free to try it. It should also help address the memory leak you observed.

Thanks for the suggestion! My setup uses `LMCacheConnectorV1` with MooncakeStore, which is different from the `LMCacheMPConnector` example.

MooncakeStore is required in my case as I rely on its specific features (RDMA, distributed memory pooling).

The pin count leak occurs in this specific call path for `kv_consumer`:

**1. Lookup phase** (`vllm_v1_adapter.py`):
```python
# get_num_new_matched_tokens() for kv_consumer
num_external_hit_tokens = self.lookup_client.lookup(
    token_ids,
    lookup_id=req_id,
    ...
)
```

**2. Contains with pin** (`local_cpu_backend.py`):
```python
def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
    with self.cpu_lock:
        if key not in self.hot_cache:
            return False
        if pin:
            self.hot_cache[key].pin()  # ← Pin happens here
            self.keys_in_request.append(key)
        return True
```

**3. wait_for_save() returns early** (`vllm_v1_adapter.py`):
```python
def wait_for_save(self):
    if self.kv_role == "kv_consumer":
        return  # ← Returns without calling lookup_unpin()!
```

The fix ensures `lookup_unpin()` is called for `kv_consumer` to balance the pin count.

deng451e:
Make sense to me. PTAL @sammshen

ApostaC:
@sammshen

sammshen (approved):
LGTM! Thanks for this fix!

Quick question: do we also have to add `lookup_unpin` to `request_finished` for aborted requests?

deng451e (approved):

lisiG9:
> LGTM! Thanks for this fix!
> 
> Quick question: do we also have to add `lookup_unpin` to `request_finished` for aborted requests?

Good catch! For aborted requests, `wait_for_save()` may not be called, leaving the pin count unbalanced.

The fix should be:

```python
def request_finished(self, request: "Request", block_ids: list[int]):
    # ... existing code ...
    
    if request.status == RequestStatus.FINISHED_ABORTED:
        if self.async_loading:
            self.lookup_client.cancel_lookup(lookup_id)
        
        # Unpin to balance pin count from lookup
        if self.lmcache_engine is not None:
            self.lmcache_engine.lookup_unpin(request.request_id)
```

I'll verify this and update the PR when I get a chance. Thanks!

lisiG9:
@sammshen @deng451e I've updated the PR to add `lookup_unpin()` for aborted requests in `request_finished()`. Could you please trigger the auto-merge?

lisiG9:
buildkite/k3-comprehensive-test — Build #842 failed
Rebased to latest dev branch. Please re-enable auto-merge when CI passes. Thanks!@sammshen @deng451e

lisiG9:
Hi @sammshen @ApostaC , could you help check why buildkite/k3-multiprocess-test Build #827 failed? The logs show "Page not found" for me. Thanks!

lisiG9:
Skip lookup_unpin for FINISHED_ABORTED requests in wait_for_save() since
they are already unpinned in request_finished(). This prevents pin count
from becoming negative and fixes k3-multiprocess-test failures.

PR #2714, opened 2026-03-09
Add /dev/dax (Device-DAX) backend for KV cache storage

jayhpark530:
Fixes #2572

## What this PR does / why we need it

This PR implements the `/dev/dax` (Device-DAX) backend discussed in **RFC issue #2572**.

The goal is to enable KV cache storage on **byte-addressable memory devices** exposed through `/dev/dax`, such as:

* Persistent memory
* CXL-based memory expanders
* Other byte-addressable memory tiers

Since LMCache already supports multiple storage backends, adding a DAX backend enables experimentation with KV cache tiering on emerging heterogeneous memory systems.

The backend maps a `/dev/dax` device into userspace and uses the mapped region as a KV cache arena managed by LMCache.

Two usage modes are supported:

### 1. Tiered backend mode

- `/dev/dax` is used as a storage tier after `LocalCPUBackend`, similar to other LMCache storage backends
- KV cache chunks are stored in the DAX arena when evicted from DRAM
- Reads copy cached chunks back into CPU memory

This is the recommended configuration.

### 2. Primary backend mode

Unlike traditional storage backends, `/dev/dax` exposes **byte-addressable memory** that can be directly mapped into userspace.
Because of this property, the mapped DAX arena can also serve as an allocator backend in LMCache.

- `/dev/dax` replaces the DRAM tier and acts as both the allocator and storage backend
- KV cache chunks are stored directly inside the DAX arena
- Returned `MemoryObj`s are backed by the mapped region

When supported by the platform, the backend registers the mapped address range using `cudaHostRegister`, allowing GPU ↔ DAX transfers to bypass an additional DRAM staging copy.
As discussed in the RFC thread, this mode is platform-dependent, so it is implemented as an opt-in fast path.

Potential benefits include:

- Extending KV cache capacity beyond DRAM using byte-addressable memory devices
- Potentially reducing DRAM staging by using a `/dev/dax` mmap arena as KV cache storage
- Enabling KV cache tiering experiments on emerging memory tiers such as PMem and CXL memory expanders

---

## Key features

* `/dev/dax` storage backend implemented as an LMCache storage plugin
* Fixed-size **slot allocator** on top of a mapped DAX arena
* LRU eviction with pin support
* Safe lifetime management using **arena leases and slot generations**
* Zero-copy reads in **primary mode**
* Graceful shutdown with in-flight operation tracking
* Integration with LMCache `AllocatorBackendInterface`

---

## Documentation

Documentation for the DAX backend configuration and usage is included:

```
[...]

PR #2788, opened 2026-03-16
Add Device-DAX (/dev/dax) storage backend for KV cache (follow-up to #2714)

### LINKED ISSUE #2572: [RFC] Add /dev/dax (Device-DAX) backend support for KV cache storage
- I’m currently working on KV cache offloading experiments using **byte-addressable memory devices exposed via `/dev/dax`** (e.g., persistent memory or emerging CXL-based memory tiers).
- Since LMCache already supports multiple storage backends, I wanted to ask whether adding a Device-DAX–backed KV cache tier would be a useful extension for the project.

- **This could also be helpful for others interested in evaluating KV cache tiering with `/dev/dax`-accessible memory (PMem, CXL, etc.) in modern heterogeneous inference setups.**

- I would like to propose adding a  `/dev/dax` **(Device-DAX) backend** for KV cache storage in LMCache.
-  `/dev/dax` enables direct userspace  `mmap` access to byte-addressable memory without involving the page cache, which seems like a good fit for KV cache tiering.
- Two possible usage models:
  - Tiered backend
    - Using `/dev/dax` as an additional capacity tier after host DRAM.
  - DRAM-bypass primary tier
    - Treating a  `/dev/dax` device as the main KV cache tier.

- At the moment, there is no direct alternative within LMCache for using byte-addressable `/dev/dax` devices as a KV cache storage tier.

**Additional context**
- The backend would be implemented as a storage plugin (storage_plugins, not deprecated external_backends).
- Memory management approach: reuse LMCache’s existing metadata/storage-manager integration, while the DAX plugin owns DAX arena allocation, indexing, and eviction.
- It would be a standalone mmap-based backend following existing backend integration patterns.
- No additional dependencies are required.
- CI-friendly unit tests can be supported via file-backed mmap emulation (no real /dev/dax device required).

I already have a working prototype and basic test coverage, and I’m currently preparing it for a PR.
Any feedback on expected design patterns, testing strategy, or integration approach would be greatly appreciated.

Thanks!

jayhpark530:
Fixes #2572

## History

This PR implements the `/dev/dax` (Device-DAX) backend proposed in **RFC issue #2572**.

The implementation is based on a previous attempt: **PR https://github.com/LMCache/LMCache/pull/2714**

Compared to the earlier version, this PR simplifies the design and provides a clean initial implementation of a DAX storage backend used as a KV cache tier alongside `LocalCPUBackend`.

The goal is to enable KV cache storage on **byte-addressable memory devices** exposed through `/dev/dax`, such as:

* Persistent memory
* CXL-based memory expanders
* Other byte-addressable memory tiers

Since LMCache already supports multiple storage backends, adding a DAX backend enables experimentation with KV cache tiering on emerging heterogeneous memory systems.

The backend maps a `/dev/dax` device into userspace and uses the mapped region as a KV cache arena managed by LMCache.

### DAX Backend

- `/dev/dax` is used as a storage tier after `LocalCPUBackend`, similar to other LMCache storage backends
- KV cache chunks are stored in the DAX arena when evicted from DRAM
- Reads copy cached chunks back into CPU memory

---

## Key features

* Device-DAX (/dev/dax) storage backend implemented as an LMCache storage plugin
* Fixed-size slot allocator on top of a memory-mapped DAX arena
* LRU-based eviction with pin support

---

## Documentation

Documentation for the DAX backend configuration and usage is included:

```
docs/source/kv_cache/storage_backends/dax.rst
```

---

## If applicable

* [x] this PR contains user facing changes - docs added
* [x] this PR contains unit tests


lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def _allocate_slot_locked(self) -> int:
+        if self._free_slots:
+            slot = self._free_slots.pop()
+            return slot * self.slot_bytes
+        if self._next_slot < self._max_slots:
+            slot = self._next_slot
+            self._next_slot += 1
+            return slot * self.slot_bytes
+        raise RuntimeError("No free slots available; eviction required")
```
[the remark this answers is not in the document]
jayhpark530: Fixed. Also updated the caller in batched_submit_put_task accordingly.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def _free_slot_locked(self, slot_id: int) -> None:
+        if slot_id < 0:
+            return
+        if slot_id not in self._free_slots:
+            self._free_slots.append(slot_id)
```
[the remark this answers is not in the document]
jayhpark530: Partially adopted. Replaced `_free_slots` with a set to get O(1) membership checks while preserving the double-free guard. deque would not improve the O(N) check.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,709 @@
...
+                        )
+                        continue
+                    while True:
+                        try:
+                            offset = self._allocate_slot_locked()
+                            break
+                        except RuntimeError:
+                            if not self._evict_one_locked():
+                                raise
+                    slot_id = int(offset // self.slot_bytes)
```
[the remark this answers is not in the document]
jayhpark530: Fixed as part of the _allocate_slot_locked refactoring.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,707 @@
...
+        extra = self.config.extra_config or {}
+        self.device_path = str(extra.get("dax.device_path", "")).strip()
+        if not self.device_path:
+            raise ValueError("extra_config['dax.device_path'] is required")
+
+        self.async_put = _to_bool(extra.get("dax.async_put", False))
+        if self.async_put and self.loop is None:
+            raise ValueError("DaxBackend async_put=true requires an asyncio event loop")
+
+        self.arena_size_gb = float(extra.get("dax.arena_size_gb", 0))
```
DongDongJu: what is meaning of arena overall? memory mapped space in devdax side?

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,707 @@
...
+            on_complete_callback(key)
+        except Exception as e:
+            logger.warning("on_complete_callback failed for key %s: %s", key, e)
+
+    def _open_arena(self) -> None:
+        fd: Optional[int] = None
+        mmap_obj: Optional[mmap.mmap] = None
+        arena_view: Optional[memoryview] = None
+        try:
+            fd = os.open(self.device_path, os.O_RDWR)
```
DongDongJu: Do we need any device permission to using this?

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,707 @@
...
+        try:
+            try:
+                capacity_bytes = os.fstat(fd).st_size
+                if capacity_bytes > 0 and self._arena_bytes > capacity_bytes:
+                    raise RuntimeError(
+                        f"dax.arena_size_gb ({self._arena_bytes} bytes) exceeds "
+                        f"device capacity ({capacity_bytes} bytes)"
+                    )
+            except OSError:
+                # Some dax devices may not report size via fstat.
```
DongDongJu: Please raise the warning here with this comment.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,707 @@
...
+
+                    if size > self.slot_bytes:
+                        logger.warning(
+                            "Skipping DAX put for key %s: object size %d exceeds "
+                            "slot size %d",
+                            key,
+                            size,
+                            self.slot_bytes,
+                        )
+                        continue
```
DongDongJu: IMO, raise error in here is better idea for decrease the confusion between cache engine and backend.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,707 @@
...
+        return self.local_cpu_backend
+
+    def close(self) -> None:
+        """Quiesce outstanding operations and release the mapped DAX arena."""
+        with self._state_lock:
+            if self._closed:
+                return
+            self._closing = True
+            while self._active_puts > 0 or self._active_ops > 0:
+                self._state_condition.wait()
```
DongDongJu: async write path * close() possibly to have deadlock here.

docs/source/kv_cache/storage_backends/dax.rst
```diff
@@ -0,0 +1,46 @@
+# SPDX-License-Identifier: Apache-2.0
+
+Device-DAX (/dev/dax)
+=====================
+
+Overview
+--------
+
+The DAX storage plugin maps a ``/dev/dax`` device with ``mmap(MAP_SHARED)``
+and uses the mapped region as a fixed-size arena for persisted KV cache
```
DongDongJu: Is is persisted?

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -0,0 +1,707 @@
...
+            raise ValueError(
+                "keys and objs must have the same length, "
+                f"got {len(keys)} and {len(objs)}"
+            )
+        futures: List[Future] = []
+
+        for key, obj in zip(keys, objs, strict=True):
+            should_finish_put = False
+            try:
+                # Reject multi-tensor objects explicitly
```
DongDongJu: Please document this as well.

DongDongJu (changes requested):
Hello @jayhpark530, Thanks for decrease the coverage for first PR.

I have few comments but generally looks good.
I tested with h100 and devdax memory and confirmed this patch works well.

One concern is that
In here, pinned state is a plain set: contains(..., pin=True) and pin() just add the key, unpin() removes it, and eviction only checks membership in that set. 

From my understanding, normal lookup flow, pins are tracked per `lookup_id` and later released per `lookup_id`, and StorageManager.batched_unpin() calls backend.unpin(key) for each pinned key. 
LocalCPUBackend handles the same path with counted memory_obj.pin()/unpin() semantics. 

So if request A and request B both pin the same key, DAX collapses both pins into one bit when A unpins, B’s protection is gone and `_evict_one_locked()` can evict the entry early. 
I recommend to change _pinned from set[CacheEngineKey] to a refcounted map, or reuse the existing pin-count model, and add a regression test for overlapping lookup IDs.

DongDongJu:
Please check the DCO and code quality as well! Thanks @jayhpark530

jayhpark530:
> Hello @jayhpark530, Thanks for decrease the coverage for first PR.
> 
> I have few comments but generally looks good.
> I tested with h100 and devdax memory and confirmed this patch works well.
> 
> One concern is that
> In here, pinned state is a plain set: contains(..., pin=True) and pin() just add the key, unpin() removes it, and eviction only checks membership in that set.
> 
> From my understanding, normal lookup flow, pins are tracked per lookup_id and later released per lookup_id, and StorageManager.batched_unpin() calls backend.unpin(key) for each pinned key.
> LocalCPUBackend handles the same path with counted memory_obj.pin()/unpin() semantics.
> 
> So if request A and request B both pin the same key, DAX collapses both pins into one bit when A unpins, B’s protection is gone and _evict_one_locked() can evict the entry early.
> I recommend to change _pinned from set[CacheEngineKey] to a refcounted map, or reuse the existing pin-count model, and add a regression test for overlapping lookup IDs.


Hello @DongDongJu,  Thanks for the careful review!

You are right that a plain set for `_pinned` is not sufficient when overlapping lookups pin the same key. In that case, one unpin could remove protection too early and allow premature eviction.

I changed `_pinned: set[CacheEngineKey]` to `_pin_counts: dict[CacheEngineKey, int]` so pinning is now ref-counted. This matches the counted pin/unpin semantics used in other backends. I also added a backend-level regression test for overlapping pin/unpin behavior, which aligns with the scope of this change and the existing backend tests.

I will address the remaining comments in follow-up replies.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: Yes. "arena" refers to the contiguous memory-mapped region on the `/dev/dax` device. I've kept it as an internal variable name, but renamed the user-facing config key to `dax.max_dax_size` for consistency with `max_local_cpu_size` / `max_local_disk_size`.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: Yes. The process needs read-write access to the `/dev/dax` device. I added a note in the docs to make that requirement explicit.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: Added a warning when fstat() cannot determine device capacity, so size validation is skipped in that case.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: Agreed. Changed this path to raise `ValueError` instead of silently skipping the put.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: I do not see a lock cycle on `_state_lock` here because `Condition.wait()` releases the lock while waiting, and the active put/op counters are decremented in the completion paths before notifying the condition. That said, `close()` can still stall if async writes do not make progress, so I added timeout-based warnings to make such stalls visible during shutdown.

docs/source/kv_cache/storage_backends/dax.rst
jayhpark530: Updated the docs to avoid the word “persisted.” Data in the mapped DAX region may survive process restarts depending on the device, but the backend does not provide durability guarantees or crash consistency.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: Added a note in `dax.rst` and clarified the inline code comment.

jayhpark530:
> Please check the DCO and code quality as well! Thanks @jayhpark530

Thanks for the reminder! I fixed the DCO and addressed the code quality checks as well, including the issues flagged by ruff / pre-commit.
I also made the AGENTS.md-related cleanups, including adding missing public docstrings and reorganizing helper methods for clarity.
I’d appreciate another look when you have a chance.

lmcache/v1/storage_backend/plugins/dax_backend.py
DongDongJu: Yes please add comment about what is area in the code.

DongDongJu:
> > Hello @jayhpark530, Thanks for decrease the coverage for first PR.
> > I have few comments but generally looks good.
> > I tested with h100 and devdax memory and confirmed this patch works well.
> > One concern is that
> > In here, pinned state is a plain set: contains(..., pin=True) and pin() just add the key, unpin() removes it, and eviction only checks membership in that set.
> > From my understanding, normal lookup flow, pins are tracked per lookup_id and later released per lookup_id, and StorageManager.batched_unpin() calls backend.unpin(key) for each pinned key.
> > LocalCPUBackend handles the same path with counted memory_obj.pin()/unpin() semantics.
> > So if request A and request B both pin the same key, DAX collapses both pins into one bit when A unpins, B’s protection is gone and _evict_one_locked() can evict the entry early.
> > I recommend to change _pinned from set[CacheEngineKey] to a refcounted map, or reuse the existing pin-count model, and add a regression test for overlapping lookup IDs.
> 
> Hello @DongDongJu, Thanks for the careful review!
> 
> You are right that a plain set for `_pinned` is not sufficient when overlapping lookups pin the same key. In that case, one unpin could remove protection too early and allow premature eviction.
> 
> I changed `_pinned: set[CacheEngineKey]` to `_pin_counts: dict[CacheEngineKey, int]` so pinning is now ref-counted. This matches the counted pin/unpin semantics used in other backends. I also added a backend-level regression test for overlapping pin/unpin behavior, which aligns with the scope of this change and the existing backend tests.
> 
> I will address the remaining comments in follow-up replies.

Please address one left comment but generally looks good to me.

lmcache/v1/storage_backend/plugins/dax_backend.py
jayhpark530: Added a comment clarifying that "arena" refers to the mapped /dev/dax region.

jayhpark530:
> Please address one left comment but generally looks good to me.

@DongDongJu hanks again for the careful review! Addressed the remaining comment as well.

DongDongJu (approved):
LGTM,
I checked this one works in my env.

jayhpark530:
Hello @sammshen, @deng451e.

This PR adds a /dev/dax storage backend for KV cache offloading.
Since it introduces a new backend, I thought it might be relevant to your area.
I'd appreciate your feedback when you have time. Thanks!

deng451e:
Thank you for the great work. I have tested this on a server config with a dax device, and it works as expected.

deng451e (approved):

PR #2803, opened 2026-03-17
[Core] feat: add SemanticLookupProvider for approximate KV cache matching

zbennett10:
## Motivation

When the standard exact chunk-hash lookup returns zero hit tokens, there is currently no way for an external system to supply an alternate token sequence whose KV cache *is* already stored.  This blocks several important use-cases:

* **Semantic KV sharing** — reuse cached KV for semantically similar but lexically distinct contexts (e.g. different phrasings of the same document).
* **Fuzzy prefix matching** — tolerate small edits at prefix boundaries without a full cold prefill.
* **RAG-aware caching** — reuse cached KV for the same retrieved context served to different queries.
* **Topic-based KV sharing** — share prefill computation across requests with the same subject matter.

This PR adds a minimal, opt-in `SemanticLookupProvider` interface that covers all of these.

---

### New file: `lmcache/v1/lookup_client/semantic_provider.py`

```
SemanticLookupResult   — dataclass returned by the provider
SemanticLookupProvider — ABC with two required hooks:
    on_lookup_miss(request_id, token_ids, num_computed_tokens)
        -> SemanticLookupResult | None
    on_request_finished(request_id, token_ids, num_prompt_tokens) -> None
  and two optional lifecycle hooks (on_init / on_shutdown).
```

`SemanticLookupResult` carries:
| Field | Purpose |
|---|---|
| `alternate_token_ids` | Donor tokens whose KV is already in the store |
| `num_cached_tokens` | Expected hit count (for logging) |
| `skip_save` | Prevent cache pollution (default `True`) |
| `provider_metadata` | Picklable opaque data forwarded to `PostLoadHook` (PR 2) |
| `source_id` | Optional label for logs/metrics |

### Modified: `lmcache/integration/vllm/vllm_v1_adapter.py`

Integration in `LMCacheConnectorV1Impl`:

| Addition | Description |
|---|---|
| `set_semantic_lookup_provider(provider)` | Public registration method |
| `_try_semantic_lookup(...)` | Called when standard lookup returns 0; re-looks up donor tokens |
| `_apply_semantic_substitution(req_meta)` | Swaps `token_ids` in `ReqMeta`; called from all 3 `build_connector_meta` paths |
| `provider_metadata` field on `ReqMeta` | Carries opaque data to `start_load_kv` / `PostLoadHook` |

**Zero overhead when no provider is registered** — all new code paths are guarded by `if self._semantic_provider is not None`.

---

## Behaviour

1. Standard chunk-hash lookup runs as normal.
2. **Only if** the result is zero hit tokens **and** a provider is registered, `on_lookup_miss` is called.
3. If the provider returns `None`, the request proceeds as a normal cold prefill (no change).
4. If the provider returns a `SemanticLookupResult`, the donor tokens are looked up in the chunk store.
5. On a donor hit: a `LoadSpec` is created, the substitution is staged, and the token-to-allocate count is returned.
6. `build_connector_meta` applies the substitution (swaps `token_ids`, sets `can_save=False` when `skip_save=True`).
7. `request_finished` notifies the provider via `on_request_finished` for bookkeeping.

---

## Test coverage

New file: `lmcache/integration/vllm/tests/test_semantic_provider.py` — 26 tests:

* `SemanticLookupResult` defaults and fields
* `SemanticLookupProvider` ABC contract (cannot instantiate without implementing abstract methods)
* `set_semantic_lookup_provider` registration
* `on_lookup_miss` not called when standard lookup returns hits
* `on_lookup_miss` not called for mock/metadata-only requests
* `on_lookup_miss` called when standard lookup returns zero hits
* Semantic hit flow: correct token count, substitution stored, `LoadSpec` created, falls back gracefully when donor not in store
* `_apply_semantic_substitution`: swaps tokens, disables save on `skip_save=True`, sets `provider_metadata`, consumes substitution exactly once, no-op when no substitution, **short donor skipped safely** (guard against `slot_mapping` length mismatch)
* `on_request_finished` notification

---

## Backward compatibility

`_semantic_provider` defaults to `None`.  All existing behaviour is completely unchanged when no provider is registered.  The new `provider_metadata` field on `ReqMeta` defaults to `None`.

---

## Note on PR ordering

**https://github.com/LMCache/LMCache/pull/2804** ([Core] feat: add PostLoadHook for KV cache transformations after retrieval) builds on top of this PR.  Specifically, the `provider_metadata` field added to `ReqMeta` here flows through to `PostLoadContext`.  Please review and merge this PR before [PR-2804](https://github.com/LMCache/LMCache/pull/2804).

### DIFF
--- a/lmcache/integration/vllm/tests/conftest.py
+++ b/lmcache/integration/vllm/tests/conftest.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Pytest configuration for vllm integration tests.
+
+Patches torch attributes that are not available in older torch versions
+so that imports of lmcache.v1.storage_backend (which references torch.uint16)
+don't fail during test collection on development machines without GPU
+dependencies installed.
+
+Also stubs out vllm when it is not installed so that tests which only use
+LMCache-internal code (and import the adapter via __new__) can still be
+collected and executed.
+"""
+# Standard
+import enum
+import sys
+from unittest.mock import MagicMock
+
+# Third Party
+import torch
+
+# ---------------------------------------------------------------------------
+# Patch torch unsigned integer dtypes missing in older torch versions
+# ---------------------------------------------------------------------------
+for _attr, _fallback in [("uint16", "int16"), ("uint32", "int32"), ("uint64", "int64")]:
+    if not hasattr(torch, _attr):
+        setattr(torch, _attr, getattr(torch, _fallback))
+
+# ---------------------------------------------------------------------------
+# Stub out vllm if not installed
+# ---------------------------------------------------------------------------
+if "vllm" not in sys.modules:
+    # Real base classes needed so that inheritance and @dataclass work
+    class _KVConnectorMetadata:
+        pass
+
+    class _KVConnectorBase_V1:
+        pass
+
+    class _KVConnectorRole(enum.Enum):
+        SCHEDULER = "scheduler"
+        WORKER = "worker"
+
+    class _RequestStatus(enum.Enum):
+        FINISHED_STOPPED = "FINISHED_STOPPED"
+        FINISHED_ABORTED = "FINISHED_ABORTED"
+        FINISHED_IGNORED = "FINISHED_IGNORED"
+
+    class _VllmConfig:
+        pass
+
+    class _SchedulerOutput:
+        pass
+
+    class _SamplingParams:
+        pass
+
+    # Build the stub hierarchy
+    vllm_stub = MagicMock()
+
+    # config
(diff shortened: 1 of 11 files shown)


KuntaiDu:
Thank you for your contribution! Same hesitation here: adding extra control logic in between vLLM and LMCache tends to be difficult to maintain because it depends on both vLLM team and LMCache team to make it functional. We would prefer this logic to be added in a way that does not touch this part of control logic --- ideally it can be purely inside LMCache.

zbennett10:
Linking this PR as well - related: https://github.com/vllm-project/vllm/pull/37339/

PR #2804, opened 2026-03-17
[Core] feat: add PostLoadHook for KV cache transformations after retrieval

zbennett10:
## Motivation

After LMCache retrieves KV tensors from the store, there is no way for external code to perform in-place transformations on those tensors before the forward pass begins.  This is required for:

* **RoPE position correction** — when a KV block was stored at position *p₁* but is being reused at position *p₂*, the rotary embeddings embedded in the K tensors must be corrected in-place.
* **Attention-sink adjustments** — modify sink tokens after retrieval.
* **Custom KV normalization** — post-process KV for quantization-aware inference.
* **Model migration** — adapt KV tensors from one model checkpoint to another.

This PR adds a minimal `PostLoadHook` interface that covers all of these.

---

## Depends on

**PR 1** — [Core] feat: add SemanticLookupProvider for approximate KV cache matching (#2803)

This PR builds on top of PR 1.  The `provider_metadata` field added to `ReqMeta` in PR 1 is what flows through to `PostLoadContext.provider_metadata` here, allowing a `SemanticLookupProvider` to pass arbitrary data to the hook that fires after its donor KV is loaded.  **Please merge PR 1 before this PR.**

---

### New file: `lmcache/v1/hooks.py`

```
PostLoadContext  — dataclass passed to every hook:
    request_id       str
    kv_caches        dict[str, torch.Tensor]   (layer_name -> paged tensor)
    slot_mapping     torch.Tensor              (physical block slots)
    num_loaded_tokens int
    provider_metadata Any | None               (from SemanticLookupProvider)

PostLoadHook     — ABC with one required method:
    after_kv_load(ctx: PostLoadContext) -> None
```

### Modified: `lmcache/integration/vllm/vllm_v1_adapter.py`

Integration in `LMCacheConnectorV1Impl`:

| Addition | Description |
|---|---|
| `add_post_load_hook(hook)` | Append a hook; hooks fire in registration order |
| `_fire_post_load_hooks(...)` | Called after `retrieve()` completes; exceptions in individual hooks are logged and suppressed |

Hooks fire **in the non-layerwise retrieval path only**.  Layerwise retrieval is async layer-by-layer and does not currently support post-load hooks.

**Zero overhead when no hooks are registered** — the hook list is checked with `if self._post_load_hooks` before any allocation.

---

## Behaviour

1. After `self.engine.retrieve(...)` completes for a request in the non-layerwise path, `_fire_post_load_hooks` is called.
2. Each registered hook receives a `PostLoadContext` with the paged KV caches, slot mapping, token count, and optional `provider_metadata`.
3. Hooks may modify `kv_caches` tensors in-place.
4. An exception in one hook is logged (with traceback) and suppressed — subsequent hooks and the forward pass are unaffected.

---

## Test coverage

New file: `lmcache/integration/vllm/tests/test_post_load_hook.py` — 14 tests:

* `PostLoadContext` field defaults and typing
* `PostLoadHook` ABC contract (cannot instantiate without implementing `after_kv_load`)
* `add_post_load_hook` registration
* `_fire_post_load_hooks` — no hooks: no-op; single hook: correct context fields; kv_caches reference equality; in-place mutation persists; multiple hooks fire in registration order; exception in one hook does not prevent others; fast-path guard (no call when list empty)
* `provider_metadata` from PR 1 (`ReqMeta.provider_metadata`) flows correctly to `PostLoadContext`

---

## Backward compatibility

`_post_load_hooks` defaults to `[]`.  All existing behaviour is completely unchanged when no hooks are registered.

### DIFF
--- a/lmcache/integration/vllm/tests/conftest.py
+++ b/lmcache/integration/vllm/tests/conftest.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Pytest configuration for vllm integration tests.
+
+Patches torch attributes that are not available in older torch versions
+so that imports of lmcache.v1.storage_backend (which references torch.uint16)
+don't fail during test collection on development machines without GPU
+dependencies installed.
+
+Also stubs out vllm when it is not installed so that tests which only use
+LMCache-internal code (and import the adapter via __new__) can still be
+collected and executed.
+"""
+# Standard
+import enum
+import sys
+from unittest.mock import MagicMock
+
+# Third Party
+import torch
+
+# ---------------------------------------------------------------------------
+# Patch torch unsigned integer dtypes missing in older torch versions
+# ---------------------------------------------------------------------------
+for _attr, _fallback in [("uint16", "int16"), ("uint32", "int32"), ("uint64", "int64")]:
+    if not hasattr(torch, _attr):
+        setattr(torch, _attr, getattr(torch, _fallback))
+
+# ---------------------------------------------------------------------------
+# Stub out vllm if not installed
+# ---------------------------------------------------------------------------
+if "vllm" not in sys.modules:
+    # Real base classes needed so that inheritance and @dataclass work
+    class _KVConnectorMetadata:
+        pass
+
+    class _KVConnectorBase_V1:
+        pass
+
+    class _KVConnectorRole(enum.Enum):
+        SCHEDULER = "scheduler"
+        WORKER = "worker"
+
+    class _RequestStatus(enum.Enum):
+        FINISHED_STOPPED = "FINISHED_STOPPED"
+        FINISHED_ABORTED = "FINISHED_ABORTED"
+        FINISHED_IGNORED = "FINISHED_IGNORED"
+
+    class _VllmConfig:
+        pass
+
+    class _SchedulerOutput:
+        pass
+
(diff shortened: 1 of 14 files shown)


KuntaiDu:
Thank you for your contribution! We are also thinking alongside this direction. I like hook-based design, but I am a bit hesitating in placing the hook in `vllm_adapter.py` --- this file is the integration file with vllm, and is already difficult to maintain. Adding new hooks inside this file introduces much more effort compared to adding hooks in lmcache-internal functions.

zbennett10:
@KuntaiDu Thank you for your feedback. It is good to hear you have been exploring this as well. We (WorldFlow AI) have been researching this for quite some time and would love to help contribute the interface to LMCache that would allow for this kind of functionality. We have similar work that we have submitted to SGLang to start to get the ball rolling there as well. We are designing an interface with them and would love to potentially create a shared interface if possible. I adjusted parts of the PR according to your feedback but it might require a more refined interface discussion. Happy to have that/adjust according to your guidance. Thanks :)

zbennett10:
We are also looking to work with the dynamo team to help define a similar interface within their own KV block manager. Eagerly awaiting any feedback you guys have as we just want to get the conversation/design of the generic semantic kv-cache interface started. Thanks team :)

PR #2805, opened 2026-03-17
Fix cachegen BytesBuffer dtype for remote metadata

darsh7807:
reproduced the failure locally with Redis and confirmed the root cause:
  BytesBufferMemoryObj.get_dtypes() returned an empty list, while get_shapes()
  returned one entry.
  RemoteMetadata.serialize() uses zip(..., strict=True) and raised:

  ValueError: zip() argument 2 is shorter than argument 1

  Fix: return a concrete dtype for byte buffers.
  BytesBufferMemoryObj.get_dtypes() now returns [torch.uint8], keeping shapes/dtypes
  aligned.


darsh7807:
Rebased onto latest dev and force-pushed the branch.

PR #2806, opened 2026-03-18
[MP][Observability][3/3] Migrate MP server telemetry to EventBus, unify config

royyhuang:
## Summary

- Migrate all 6 MP server telemetry call sites (store/retrieve/lookup START/END) from the old `TelemetryController` to `EventBus.publish(Event(...))`
- Add `MPServerLoggingSubscriber` (debug logs) and `MPServerTracingSubscriber` (OTel spans from START/END pairs)
- Replace `PrometheusConfig` + `TelemetryConfig` with unified `ObservabilityConfig` controlling metrics, logging, and tracing via CLI flags
- Move OTLP endpoint selection from `OTEL_EXPORTER_OTLP_ENDPOINT` env var to `--otlp-endpoint` config option
- Delete entire `telemetry/` subdirectory (controller, processors, config, event model) and its tests
- Clean up docs: `REFACT_DESIGN.md` → `DESIGN.md` (architecture only), new `EVENTS.md` for metadata contracts

### New files
- `subscribers/logging/mp_server.py` — `MPServerLoggingSubscriber`
- `subscribers/tracing/mp_server.py` — `MPServerTracingSubscriber`
- `DESIGN.md` — architecture design doc (replaces `REFACT_DESIGN.md`)
- `EVENTS.md` — event metadata contracts

### Modified files
- `server.py`, `blend_server.py`, `blend_server_v2.py`, `http_server.py` — use `ObservabilityConfig`, conditional subscriber registration, OTLP endpoint passthrough
- `config.py` — `ObservabilityConfig` with `add_observability_args()` / `parse_args_to_observability_config()`
- `otel_init.py` — remove env var fallback, config-driven OTLP endpoint
- `README.md` — full config reference (CLI flags, config fields, export modes, tracing)
- `METRICS.md` — add MP server events, link to `EVENTS.md`
- All subscriber `__init__.py` — add `__all__` exports

### Deleted files
- `telemetry/` directory (7 source files + design doc)
- `tests/v1/mp_observability/telemetry/` (3 test files)

## Test plan
- [x] `uv run python -m pytest tests/v1/mp_observability/ --noconftest -v` — 55 tests pass
- [x] Pre-commit hooks pass
- [x] OTLP collector verified: all 9 metrics (`l1_read_keys`, `l1_write_keys`, `l1_evicted_keys`, `sm_read_requests`, `sm_read_succeed_keys`, `sm_read_failed_keys`, `sm_write_requests`, `sm_write_succeed_keys`, `sm_write_failed_keys`) received with real vLLM traffic
- [x] All 3 span types verified: `mp.store` (115), `mp.lookup_prefetch` (28), `mp.retrieve` (3) with correct attributes and sub-ms GPU-accurate timestamps


sammshen:
@codex review

lmcache/v1/mp_observability/subscribers/logging/mp_server.py
```diff
@@ -0,0 +1,87 @@
...
+logger = init_logger(__name__)
+
+try:
+    # Third Party
+    from opentelemetry.sdk._logs import LoggingHandler
+
+    _otel_handler = LoggingHandler(level=logging.DEBUG)
+    logger.addHandler(_otel_handler)
+except ImportError:
+    pass
```
KuntaiDu: UX-wise, ideally I would prefer to put
```
try:
    # Third Party
    from opentelemetry.sdk._logs import LoggingHandler

    _otel_handler = LoggingHandler(level=logging.DEBUG)
    logger.addHandler(_otel_handler)
except ImportError:
    pass
```
inside `init_logger` function, so that in the future, those who calls `init_logger` will automatically try to add OTel SDK handler without writing extra code.

KuntaiDu (approved):
Otherwise LGTM

lmcache/v1/mp_observability/subscribers/logging/mp_server.py
ApostaC, line 1: Previously, we could configure the level for the telemetry logging. Now seems like everything is hard-coded to "debug". Is it  possible to make the log level configurable?

lmcache/v1/mp_observability/config.py
ApostaC, line 1: Since the CLI args are drastically changed, can we update the user docs to reflect the changes as well?

ApostaC: IIUC, we now have 3 modes: prometheus, logging, and tracing. It would be great if we could update https://docs.lmcache.ai/mp/observability.html with details about how to use them and what to expect

lmcache/v1/mp_observability/EVENTS.md
ApostaC, line 1: Should we create AGENTS.override.md in the `mp_observability` folder to let the AI agents respect the EVENTS.md and remember to update it when introducing new events?

lmcache/v1/mp_observability/config.py
ApostaC, line 1: For the new configs, will there by any conflicting arguments, or will one argument require another to be set? 
For example, if we have `--tracing-enabled`, should the argparser assert `--otlp-endpoint` to be set?

lmcache/v1/mp_observability/subscribers/tracing/mp_server.py
ApostaC, line 1: Can we have some "sweeping" logic to prevent memory leak? E.g., when the number of active spans becomes more than XXX (for example, 10K), we remove the spans whose start time is YYY seconds ago (for example, 300 seconds)

lmcache/v1/multiprocess/server.py
```diff
@@ -382,15 +374,14 @@ def retrieve(
...
-                ),
-            )
+        gpu_context.cupy_stream.launch_host_func(
+            self._event_bus.publish,
+            Event(
+                event_type=EventType.MP_RETRIEVE_START,
+                session_id=key.request_id,
+                metadata={"device": str(gpu_context.device)},
+            ),
+        )
```
ApostaC: Can we still have the global checking flag? Although `self._event_bus.publish` will become a noop, we still don't want to call `launch_host_func` in the cuda stream.

ApostaC (commented):
Please see the comments below.

Regarding the user-facing changes, here's the summary from cc:

Image

lmcache/v1/mp_observability/subscribers/tracing/mp_server.py
royyhuang: iirc, otel will either directly export it out the span (w/ SimpleSpanProcessor) or batch up to a configurable number of spans, 2048 by default, then export. Either way, there will be no memory leak.

ApostaC (approved):
LGTM!

PR #2808, opened 2026-03-18
[MP][Bugfix] fixing race condition for zmq output notifier

ApostaC:

Fixes a race condition in `MessageQueueServer._call_blocking_handler` where the `output_notifier` zmq PUSH socket was called from `ThreadPoolExecutor` callback threads. ZMQ sockets are not thread-safe, so concurrent `send()` calls from multiple worker threads could corrupt state.

Replaces the zmq PUSH/PULL inproc socket pair with a single `os.eventfd`. `eventfd_write()` is an atomic kernel syscall, making it inherently thread-safe for cross-thread notification. This also simplifies the drain logic — `eventfd_read()` atomically resets the counter in one call vs. a recv loop.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/mq.py
+++ b/lmcache/v1/multiprocess/mq.py
@@ -337,8 +337,9 @@ class MessageQueueServer:
         self.ctx = context
         self.socket = self.ctx.socket(zmq.ROUTER)
         self.socket.bind(bind_url)
-        # Output task notifier socket and output queue
-
+        # Use eventfd instead of zmq PUSH/PULL sockets because blocking
+        # handler callbacks run on ThreadPoolExecutor threads, and zmq
+        # sockets are not thread-safe. eventfd_write() is atomic.
         self._output_efd = os.eventfd(0, os.EFD_NONBLOCK | os.EFD_CLOEXEC)
         self.output_queue: queue.Queue = queue.Queue()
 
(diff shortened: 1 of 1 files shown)


lmcache/v1/multiprocess/mq.py
```diff
@@ -338,15 +339,13 @@ def __init__(self, bind_url: str, context: zmq.Context, max_workers: int = 4):
         self.socket.bind(bind_url)
         # Output task notifier socket and output queue
 
-        self.output_notifier, self.output_waiter = prepare_internal_push_pull_sockets(
-            self.ctx
-        )
+        self._output_efd = os.eventfd(0, os.EFD_NONBLOCK | os.EFD_CLOEXEC)
```
maobaolong, line 343: nit: Worth adding a brief comment explaining why eventfd is used instead of the previous ZMQ socket, for future readers who won't have the PR context.

maobaolong (commented):
Thanks for this fix, LGTM. left nit only.

sammshen:
@cursor review

sammshen:
@cursor review

sammshen (approved):
LGTM

ApostaC:
@codex review

YaoJiayi (approved):

PR #2813, opened 2026-03-18
[MP] Add a new argument to specify whether retain_in_l1

maobaolong:

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Adds a new prefetch policy that changes L1 lifetime semantics for prefetched data and can increase memory usage if enabled. The default behavior remains temporary prefetch, so impact is limited to users selecting the new policy.
> 
> **Overview**
> Adds an L2 prefetch retention mechanism by extending `PrefetchPolicy` with `select_l1_retentions()` and wiring it into `PrefetchController` so `reserve_write()` can mark prefetched entries as temporary vs permanent.
> 
> Introduces a new `retain` prefetch policy (`RetainPrefetchPolicy`) that keeps all prefetched keys permanently in L1, updates MP docs for the new `--l2-prefetch-policy retain` option, and adds unit tests covering both default (temporary) and retain behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 529032ca4a1b1ea41624d043ff329e4ff6b5a83d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/distributed/storage_controllers/prefetch_policy.py
+++ b/lmcache/v1/distributed/storage_controllers/prefetch_policy.py
@@ -169,13 +169,6 @@ class DefaultPrefetchPolicy(PrefetchPolicy):
 
         return plan
 
-    def select_l1_retentions(
-        self,
-        keys: list[ObjectKey],
-    ) -> list[bool]:
-        """All prefetched keys are temporary (not retained)."""
-        return [False] * len(keys)
-
 
 class RetainPrefetchPolicy(DefaultPrefetchPolicy):
     """Prefetch policy that retains all prefetched keys in L1.
--- a/lmcache/v1/distributed/storage_controllers/prefetch_policy.py
+++ b/lmcache/v1/distributed/storage_controllers/prefetch_policy.py
@@ -50,6 +50,32 @@ class PrefetchPolicy(ABC):
             of the union of the input bitmaps.
         """
 
+    def select_l1_retentions(
+        self,
+        keys: list[ObjectKey],
+    ) -> list[bool]:
+        """Determine which keys to retain in L1 after prefetched
+        objects are consumed.
+
+        Called by PrefetchController just before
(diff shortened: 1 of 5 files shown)


ApostaC (commented):
Can we do it as part of the prefetch policy? 
We can introduce a new interface in the prefetch policy, for example
```python
def select_l1_retentions(
    self,
    keys: list[ObjectKey],
) -> list[bool]
    """
    Determine which keys need to be retained/deleted after the prefetched objects are consumed.
    Args: ......
    Returns: ...... (the number of bools needs to be the same as the number of input keys)
    """
```

Then, we can call it just before the `l1_mgr.reserve_write` in the prefetch controller

maobaolong:
@ApostaC Thanks for the previous review, addressed the comment, PTAL.

ApostaC (approved):
LGTM!

lmcache/v1/distributed/storage_controllers/prefetch_policy.py
```diff
@@ -50,6 +50,32 @@ def select_load_plan(
...
+        hot-cache or selective-retention strategies.
+
+        Args:
+            keys: Keys about to be written into L1.
+
+        Returns:
+            A list of bools with the same length as *keys*.
+            ``True`` = retain (permanent), ``False`` = temporary.
+        """
+        return [False] * len(keys)
```
chunxiaozheng, line 77: @maobaolong Good catch, but I have a little question, all the implementations directly return `false`, when will it return True?

maobaolong: Lost a `RetainPrefetchPolicy` implementation, and now added.

maobaolong:
@chunxiaozheng Thanks for the remind, PTAL.

chunxiaozheng (approved):
lgtm!

PR #2816, opened 2026-03-18
fix: replace global lock with per-device transfer_lock to prevent deadlock

maobaolong:

The global `self.lock` in MPCacheEngine was acquired inside
`torch.cuda.device()` context, creating a circular dependency
with the implicit CUDA driver lock (ABBA deadlock).

Replace it with a per-device `transfer_lock` on GPUCacheContext
 so that GPU↔CPU transfers on the same
device are serialised without cross-lock contention. 

**If applicable**:


ApostaC (approved):
LGTM!

chunxiaozheng (approved):
lgtm

PR #2821, opened 2026-03-19
fix: save_decode_cache is broken on vllm >= 0.9.2

hlin99:
In newer vLLM versions (>= 0.9.2), request.num_computed_tokens is already updated before build_connector_meta is called. This causes the slice request.all_token_ids[num_computed_tokens:...] to exceed the length of all_token_ids during decode, silently returning an empty list instead of the new token.

As a result, tracker.token_ids never grows, input_token_len stays at prompt length, skip_save is always True, and KV cache is never saved during decode phase.

Fix by using len(request_tracker.token_ids) as the slice base, which is only updated after the slice and is always in sync with LMCache's own state.

---

> [!NOTE]
> **Medium Risk**
> Adjusts how new tokens are sliced during decode/cached-request scheduling; incorrect indexing could still impact cache save behavior, but the change is localized to metadata construction.
> 
> **Overview**
> Fixes `build_connector_meta` token slicing for scheduled cached requests by basing the slice start on LMCache’s tracked token count (and taking `min(tracker_len, request.num_computed_tokens)`), instead of relying solely on `request.num_computed_tokens`.
> 
> This prevents empty `new_token_ids` slices on vLLM >= 0.9.2 (and remains safer under preemption/rollback), allowing `RequestTracker.token_ids` to grow correctly so decode-phase KV cache can be saved as intended.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c608c668893058a438b06b1aad65c636811b3edb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1537,8 +1537,14 @@ def build_connector_meta(
             # TODO: this is a dangerous reference to the request object inside vllm
             if request := self._unfinished_requests.get(req_id):
                 num_current_tokens = request.num_computed_tokens
+                # Use tracker's token_ids length as the slice base instead of
```
sammshen: if it's not computed, how can we have saved the decode cache previously?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1537,8 +1537,14 @@ def build_connector_meta(
...
+                # request.num_computed_tokens, because in newer vLLM versions
```
sammshen: why will num_computed_tokens be updated before the forward pass which computes them is finished?

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1537,8 +1537,14 @@ def build_connector_meta(
...
+                # num_computed_tokens may already be updated before
+                # build_connector_meta is called, causing the slice to be empty
+                # during decode phase.
+                tracker_len = len(request_tracker.token_ids)
```
sammshen, line 1543: I think the upstream issue is can we keep request_tracker.token_ids semantics concistent (is it the number of computed tokens, if so, why will it be mismatched with num_computed_tokens)

sammshen (changes requested):

hlin99:
thanks for your input @sammshen . your points make sense to me. give me some time to deep dive the issue. it's weird that seems only occurs at some condition.

hlin99:
Hi @sammshen   I cleaned up my env and on cuda device I can easily repro the issue with below configs. (so it's a common issue across platforms)

LMCache: 3137ce0ba3c0875faa1d583ad7115fc411733a3d
vllm: 2e67fa756d05518823196815a30bc633031ddb60 (today's latest)

lmcache.yaml
    local_cpu: True
    max_local_cpu_size: 10
    save_unfull_chunk: true
    save_decode_cache: true

The root cause is that, num_computed_tokens is not updated together with request.all_token_ids. num_computed_tokens is updated in scheduler, while the fwd path is not updated request.all_token_ids yet when LMCache tries to get the new_token_ids from the list. https://github.com/LMCache/LMCache/blob/dev/lmcache/integration/vllm/vllm_v1_adapter.py#L1536

the list is Out of bounds at this point, so input_token_len won't get increamented here (https://github.com/LMCache/LMCache/blob/dev/lmcache/integration/vllm/vllm_v1_adapter.py#L317) and failed to hit the chunk boundary and get stored.... 

Let me know if you have any other question. thanks.

lmcache/integration/vllm/vllm_v1_adapter.py
hlin99: the vllm flow is:  engine->schedule+++ -> call LMCache this function -> schedule --- -> model executor. so forward path is not being called, and it's the vllm design flow, not vllm bug...

lmcache/integration/vllm/vllm_v1_adapter.py
hlin99: same as above... i think LMCache can't take num_computed_tokens to index token_ids. it's unsafe per vllm design.

lmcache/integration/vllm/vllm_v1_adapter.py
hlin99: it worked either vllm < 0.9.2 where vllm uses the legacy way. or before the change of https://github.com/LMCache/LMCache/pull/2007 ....  in https://github.com/LMCache/LMCache/pull/2007 num_current_tokens is changed to take from request.num_computed_tokens, before that it was from len(request_tracker.token_ids).   @sammshen  do you still remember, why we have such change in the PR?

sammshen:
does this have to do with async scheduling (which is the default in the latest vllm versions)

sammshen:
@hlin99 could you share your testing procedure?

sammshen:
would the correct thing to do be to update lmcache's request tracker to be num_computed_tokens

hlin99:
> @hlin99 could you share your testing procedure?

hi @sammshen

LMCache: https://github.com/LMCache/LMCache/commit/3137ce0ba3c0875faa1d583ad7115fc411733a3d
vllm: 2e67fa756d05518823196815a30bc633031ddb60 (today's latest)

lmcache.yaml
local_cpu: True
max_local_cpu_size: 10
save_unfull_chunk: true
save_decode_cache: true

server launch cmd:
CUDA_VISIBLE_DEVICES=4 PYTHONHASHSEED=0 LMCACHE_CONFIG_FILE=lmcache.yaml vllm serve --model /workspace/Meta-Llama-3-8B-Instruct/ --dtype bfloat16 --gpu-memory-utilization 0.95 --max-num-seqs 512 --max-model-len 4096 --enforce-eager --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both","kv_connector_extra_config": {"discard_partial_chunks": false, "lmcache_rpc_port": "producer1"}}'

vllm bench:
vllm bench serve --host 127.0.0.1 --port 8000 --model /workspace/Meta-Llama-3-8B-Instruct/ --dataset-name random --random-input-len 100 --random-output-len 2000 --num-prompts 20 --burstiness 100 --request-rate 3.6

you'll see if decode save cache is triggered https://github.com/256 chunk boundary w & w/o this PR

hlin99:
> does this have to do with async scheduling (which is the default in the latest vllm versions)

actually not...
 request.num_computed_tokens is updated before new decode token is appended on request.all_token_ids
so, at this code point, if lmcache uses request.num_computed_tokens to index request.all_token_ids, it must be out of bound. and new_token_ids is actually None.
and later on the code request_tracker.update(new_token_ids, ...) ~line 1630 actually do nothing update. so decode number of tokens never gets increased to hit chunk boundary.

               num_current_tokens = request.num_computed_tokens
               new_token_ids = request.all_token_ids[
                    num_current_tokens : num_current_tokens + num_new_tokens
                ]

sammshen:
conceptually, consulting request_tracker instead of num_computed_tokens is right

sammshen:
I think this solves save decode cache but regresses preemptino case. the correct approach is probably to take a min: 

```python
num_current_tokens = request.num_computed_tokens
  tracker_len = len(request_tracker.token_ids)                                                                                                                                                                               
  slice_base = min(num_current_tokens, tracker_len)                                                                                                                                                                          
  new_token_ids = request.all_token_ids[                                                                                                                                                                                     
      slice_base : slice_base + num_new_tokens                                                                                                                                                                               
  ] 
```

neither one should be smaller than we expect. I made the change earlier becasue there were soem cases (e.g. preemption) where request_tracker.token_ids <= request_tracker.token_ids but it seems from your PR the opposite can be true too

sammshen:
@cursor review

hlin99:
> I think this solves save decode cache but regresses preemptino case. the correct approach is probably to take a min:
> 
> ```python
> num_current_tokens = request.num_computed_tokens
>   tracker_len = len(request_tracker.token_ids)                                                                                                                                                                               
>   slice_base = min(num_current_tokens, tracker_len)                                                                                                                                                                          
>   new_token_ids = request.all_token_ids[                                                                                                                                                                                     
>       slice_base : slice_base + num_new_tokens                                                                                                                                                                               
>   ] 
> ```
> 
> neither one should be smaller than we expect. I made the change earlier becasue there were soem cases (e.g. preemption) where request_tracker.token_ids <= request_tracker.token_ids but it seems from your PR the opposite can be true too

taking a min is a good idea! fix has been updated. please take another look, thanks. @sammshen

sammshen:
there are two bugs with save decode cache, this PR should do both: https://github.com/LMCache/LMCache/pull/2929

thanks for the catch @hlin99

hlin99:
> there are two bugs with save decode cache, this PR should do both: #2929
> 
> thanks for the catch @hlin99

nice to have both issue fixed! thx.

PR #2794, opened 2026-03-17
[MP][Observability][2/3] Migrate L1 + SM to EventBus + OTel, remove old Prometheus pipeline

royyhuang:
## Summary
- Migrate L1Manager and StorageManager observability from old Listener/Stats/PrometheusController system to EventBus + OpenTelemetry
- L1Manager publishes events directly to EventBus alongside listener iteration (listeners stay for StoreListener/EvictionPolicy business logic)
- StorageManager fully migrated: listener iteration replaced with `bus.publish()`
- New OTel metrics subscribers (`L1MetricsSubscriber`, `SMMetricsSubscriber`) and logging subscribers with OTel `LoggingHandler` bridge
- Dual metrics export mode: OTLP push (`OTEL_EXPORTER_OTLP_ENDPOINT`) or Prometheus pull fallback (default)
- Removed old Prometheus pipeline from MP mode: `PrometheusController`, stats dataclasses, stats loggers
- Restructured subscribers into `metrics/`, `logging/`, `telemetry/` subdirs
- Updated docs: `README.md` (replaces `LOGGER_GUIDE.md`), `METRICS.md`, `REFACT_DESIGN.md`

Depends on #2792. **This is PR 2 of 3.**

## Test plan
- [x] 78 tests pass (`uv run python -m pytest tests/v1/mp_observability/ --noconftest`)
- [x] All pre-commit hooks pass
- [x] Metrics verified on Prometheus fallback (`curl localhost:9090/metrics`)
- [x] Metrics verified on OTLP push (via test receiver on :4317)
- [x] End-to-end: run LMCache MP server, do store/retrieve, verify metrics appear

PR #2824, opened 2026-03-19
[Core] Add L2 eviction in mp mode

YaoJiayi:

**If applicable**:


lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -193,7 +193,7 @@ def __init__(
         policy: PrefetchPolicy,
         max_in_flight: int = 8,
     ) -> None:
-        super().__init__(l1_manager)
+        self._l1_manager = l1_manager
```
ApostaC, line 196: Why do we need to make the change here?

lmcache/v1/distributed/storage_controllers/store_controller.py
```diff
@@ -170,13 +170,13 @@ def __init__(
         adapter_descriptors: list[AdapterDescriptor],
         policy: StorePolicy,
     ) -> None:
-        super().__init__(l1_manager)
+        self._l1_manager = l1_manager
```
ApostaC, line 173: Same question, what's the issue with using `super().__init__`?

docs/design/l2_adapters/l2_eviction.md
```diff
@@ -0,0 +1,230 @@
...
+  │     ▲                  │
+  │     │ events           │ delete(keys)
+  │     └──────────────────┘
+  │
+  ├─ L2Adapter[1]  (no eviction config → no controller)
+  │
+  └─ L2Adapter[2]  ──► L2EvictionController[2]  (if eviction_config set)
+        ▲                  │
+        │ events           │ delete(keys)
+        └──────────────────┘
```
ApostaC: Quick question: can we have a unified L2EvictionController (i.e., only a single controller thread).
Having a new thread per L2 adapter will be a bit too expensive (especially in python).

The unified controller need to have a map between the L2 adapter (there should be a small desc class that can be used as the dict key) and the registered eviction policy. And it should go over all the adapters in the loop.

In the future, a better design is to let the adapter notify the watermark information via in-process signals (eventfd or threading.event) so that the controller doesn't need to pull every one second.

lmcache/v1/mp_observability/logger/l2_stats_logger.py
ApostaC, line 1: This class will no longer be needed (should be deleted in #2794).

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -16,12 +16,16 @@
...
 
 L2TaskId = int
 
 
 class L2AdapterInterface(ABC):
+    def __init__(self):
+        self._listeners: list[L2AdapterListener] = []
+
     """
     The abstracted interface for L2 I/O adapters.
```
ApostaC: nit: Please put the new __init__ function below the class-level docstring

lmcache/v1/distributed/storage_controllers/eviction_controller.py
```diff
@@ -6,78 +6,128 @@
...
+        self._thread.join()
+
+    def report_status(self) -> dict:
+        raise NotImplementedError
+
+    def _eviction_loop(self):
+        raise NotImplementedError
+
+    def _execute_eviction_action(self, action: EvictionAction):
+        raise NotImplementedError
```
ApostaC: Since the L1 and L2 eviction controllers may be very different. We don't need to have a unified base class for 2 eviction controllers.

We can just do:
- rename `EvictionController` to `L1EvictionController`, and 
- create a new class called `L2EvictionController`

lmcache/v1/distributed/eviction.py
ApostaC, line 1: Similar thing here: let's avoid multi-inheritance by doing:
- rename `EvictionPolicy` to `L1EvictionPolicy` (extending L1ManagerListener interface only)
- create a new class called `L2EvictionPolicy` (extending the L2AdapterListener interface only)
- no shared base class between L1 and L2 eviction policy

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -266,6 +270,66 @@ def query_load_result(self, task_id: L2TaskId) -> Bitmap | None:
...
+                - ``current_usage``: fraction of total L2 capacity currently
+                  occupied (bytes used / total bytes).
+                - ``usage_after_ongoing_eviction``: estimated fraction once all
+                  in-flight deletes/evictions complete
+                  ((bytes used - bytes being deleted) / total bytes).
+
+            The default implementation returns ``(0.0, 0.0)``. Subclasses that
+            support eviction should override this method.
+        """
+        return (0.0, 0.0)
```
ApostaC: Return 0 is a bit ambiguous here: it could mean that there is no data stored in L2, or the L2 adapter does not support `get_usage`. 

Can we make it return (-1, -1) by default, to differentiate zero usage and non-implemented?

lmcache/v1/distributed/l2_adapters/mock_l2_adapter.py
ApostaC, line 1: do we need to call `_notify_keys_accessed`?

lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
```diff
@@ -245,6 +245,18 @@ def query_load_result(self, task_id: L2TaskId) -> Bitmap | None:
...
+    # ---------------------------------------------------------------
+
+    def delete(self, keys: list[ObjectKey]) -> None:
+        # Not implemented for the native connector adapter.
+        pass
+
+    def get_usage(self) -> tuple[float, float]:
+        # Not implemented for the native connector adapter.
+        return (0.0, 0.0)
+
```
ApostaC, line 259: cc @sammshen , we may need to consider supporting eviction in the native L2 adapter

lmcache/v1/distributed/storage_manager.py
```diff
@@ -78,6 +79,20 @@ def __init__(self, config: StorageManagerConfig):
...
+                eviction_config=ac.eviction_config,
+            )
+            for adapter, ac in zip(
+                self._l2_adapters, config.l2_adapter_config.adapters, strict=False
+            )
+            if ac.eviction_config is not None
+        ]
+        for ctrl in self._l2_eviction_controllers:
+            ctrl.start()
+
```
ApostaC: If we implemented the unified L2 eviction controller interface, here will be:
1. initialize L2 adapters
2. initialize L2 eviction policies based on the config, and register them to L2 adapters
3. create the L2 eviction controller
4. register (L2 adapter, L2 eviction policy) pairs into the eviction controller, and start it

Note that only L2 eviction policy need to extend L2AdapterListener, the eviction controller doesn't need to

lmcache/v1/distributed/storage_controllers/eviction_controller.py
```diff
@@ -6,78 +6,128 @@
...
+
+
+class L1EvictionController(EvictionController, L1ManagerListener):
```
ApostaC: Just wondering why L1EvictionController need to extend the L1ManagerListener? This is the job of L1EvictionPolicy

ApostaC (changes requested):
Main comment on the overall design: let's have a unified L2 eviction controller, not one per L2 adapter

Other things:
- Let's avoid multi inheritance
- The eviction controller don't need to extend the "Listener" interface, because they do not need to be aware of the exact key access/insertion/deletion, but only the memory usage information. Only the eviction policy need to extend the listener interface to monitor the accesses of keys.

lmcache/v1/distributed/storage_controllers/store_controller.py
YaoJiayi: Because l2 eviction controller does not need to hold a reference to l1_manager for now.

lmcache/v1/distributed/storage_controllers/eviction_controller.py
YaoJiayi: done

lmcache/v1/distributed/eviction.py
YaoJiayi: done

lmcache/v1/distributed/l2_adapters/base.py
YaoJiayi: done

docs/design/l2_adapters/l2_eviction.md
YaoJiayi: done

lmcache/v1/distributed/l2_adapters/base.py
YaoJiayi: done

lmcache/v1/distributed/storage_manager.py
YaoJiayi: Done

lmcache/v1/distributed/l2_adapters/mock_l2_adapter.py
YaoJiayi: good catch. fixed

lmcache/v1/distributed/storage_controllers/eviction_controller.py
YaoJiayi: Fixed multi-inheritance

lmcache/v1/distributed/storage_controllers/eviction_controller.py
```diff
@@ -7,77 +7,101 @@
...
-        self._eviction_policy = CreateEvictionPolicy(self._eviction_config)
-        self.get_l1_manager().register_listener(self._eviction_policy)
+    Provides the shared eviction loop structure: background thread, stop flag,
+    and eviction policy. Subclasses implement _eviction_loop and
+    _execute_eviction_action for their specific tier (L1 or L2).
+    """
 
+    def __init__(self, eviction_config: EvictionConfig):
+        self._eviction_config = eviction_config
+        self._eviction_policy = CreateEvictionPolicy(eviction_config)
```
ApostaC: We can move this into L1EvictionController, because L2 eviction controller does not have a single eviction policy.

The shared part of the controller is just the thread.

lmcache/v1/distributed/storage_controllers/eviction_controller.py
```diff
@@ -7,77 +7,101 @@
...
         self._thread.join()
 
+    def report_status(self) -> dict:
+        raise NotImplementedError
+
+    def _eviction_loop(self):
+        raise NotImplementedError
+
+    def _execute_eviction_action(self, action: EvictionAction):
+        raise NotImplementedError
```
ApostaC: nit: can we define those functions as public functions (i.e., not starting with `_`) and make them "abstract". 
Additionally, it would be great if we could state "the child class needs to override this function to implement internal eviction controlling logic" in the docstring.

ApostaC (commented):
just nit comments, otherwise LGTM!

lmcache/v1/distributed/storage_controllers/eviction_controller.py
YaoJiayi: done

lmcache/v1/distributed/storage_controllers/eviction_controller.py
YaoJiayi: done

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #2825, opened 2026-03-19
[CLI] Implement `lmcache describe kvcache` subcommand

royyhuang:
## Summary

- Add `lmcache describe kvcache --url <url>` command (Phase 1 of the CLI roadmap)
- Fetches `/api/status` from the HTTP server and displays health, engine type, chunk size, L1 memory usage, eviction policy, cached objects, and active sessions
- Supports `--format json` and `--output <path>` via the existing Metrics framework
- Change default HTTP port from 8000 to 8080
- Use `__class__.__name__` for `engine_type` in `report_status()` so subclasses report their own name
- Design doc at `docs/design/cli/describe.md`
- 18 unit tests covering helpers, field extraction, error handling, and a real HTTP server round-trip

## Test plan

- [x] `pytest tests/cli/test_describe.py` — 18 tests pass
- [x] Manual: `lmcache describe kvcache --url http://localhost:8080` against a running HTTP server
- [x] Manual: `--format json` output has clean snake_case keys and raw values
- [x] Manual: connection refused / 503 errors print to stderr and exit 1


docs/design/cli/describe.md
```diff
@@ -0,0 +1,286 @@
...
+ZMQ endpoint:                            tcp://localhost:5555
+HTTP endpoint:                           http://localhost:8000
+Engine type:                             blend
+Chunk size:                              256
+L1 capacity (GB):                        60.00
+L1 used (GB):                            42.30 (70.5%)
+Eviction policy:                         LRU
+Cached objects:                          1024
+Active sessions:                         3
+Uptime:                                  2h 14m 32s
```
KuntaiDu, line 32: kv cache shape & layout

KuntaiDu: ========== DeepSeek =========
Registered KV caches:     [2, T, Head, Hidden]
World size:                        4

KuntaiDu: Open up a section

lmcache/v1/multiprocess/server.py
```diff
@@ -698,7 +698,7 @@ def report_status(self) -> dict:
         sm = self.storage_manager.report_status()
         return {
             "is_healthy": sm["is_healthy"],
-            "engine_type": "MPCacheEngine",
+            "engine_type": self.__class__.__name__,
```
KuntaiDu, line 720: Let's expose more information to `lmcache describe`

tests/cli/test_describe.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def test_unhealthy(self):
+        """Verify health shows UNHEALTHY when is_healthy is False."""
+        # First Party
+        from lmcache.cli.commands.describe import DescribeCommand
+
+        unhealthy_data = {**SAMPLE_STATUS, "is_healthy": False}
+        cmd = DescribeCommand()
+
+        class FakeArgs:
```
KuntaiDu, line 196: I don't fully understand this test -- what is it for?

tests/cli/test_describe.py
```diff
@@ -0,0 +1,332 @@
...
+
+            output = json.loads(buf.getvalue())
+
+        m = output["metrics"]
+        assert m["health"] == "OK"
+        assert m["url"] == "http://localhost:8000"
+        assert m["engine_type"] == "MPCacheEngine"
+        assert m["chunk_size"] == 256
+        assert m["l1_capacity_gb"] == 60.0
+        assert m["l1_used_gb"] == "42.30 (70.5%)"
```
KuntaiDu, line 170: Wondering why this `l1_used_gb` is a fixed value --- is it coded somewhere in the test?

KuntaiDu: Oh I see it is coded in the sampled trace. In this case I feel like the current way does not protect the semantics of `report_status`. Maybe we can add test around that.

tests/cli/test_describe.py
KuntaiDu: Oh I see --- it is testing the case where response_status returns nothing

docs/design/cli/describe.md
```diff
@@ -0,0 +1,286 @@
...
+==================================================
```
ApostaC, line 51: What about L2 information? Are we going to display them as different sections?

ApostaC (approved):
When thinking about `lmcache bench`, one of the feature requirements is to let `lmcache describe` return the KV cache shape or size information.
Ideally, something like the following will be enough for the benchmark tool to compute the KV cache size information: 
```json
{
  ### Other data in the json output ...
  "kv_cache_information": [
    {
      "model_name": "model A",
      "world_size": 2,
      "local_rank": 0,
      "num_kv_tensors": 32,
      "kv_tensor_shape": [2, 1000, 16, 4, 128],
      "dtype": "torch.bfloat16"
    },
    {
      "model_name": "model A",
      "world_size": 2,
      "local_rank": 1,
      "num_kv_tensors": 32,
      "kv_tensor_shape": [2, 1000, 16, 4, 128],
      "dtype": "torch.bfloat16"
    },
    {
      "model_name": "model B",
      "world_size": 1,
      "local_rank": 0,
      "num_kv_tensors": 64,
      "kv_tensor_shape": [2, 2000, 16, 8, 128],
      "dtype": "torch.bfloat16"
    }
  ]
}
```

No need to do that for now, but it's needed before the `lmcache bench`

docs/design/cli/describe.md
royyhuang: Yes I just added them to another section.

royyhuang: The model specific information (e.g., GPU KV shape etc) is added as well.

KuntaiDu (approved):
Workflow in general LGTM. Small nits listed in comments.

KuntaiDu:
Also add user-facing doc in docs/source

PR #2832, opened 2026-03-20
[MP][1/2] Create new cuda kernel for MP mode with a kernel testing tool

ApostaC:
Please see #2838

---

Adds a standalone CUDA kernel test harness (`lmcache/tools/kernel_harness/`) for the new `multi_layer_block_kv_transfer` kernel, which transfers KV cache between vLLM/SGLang paged buffers and LMCache memory objects at block granularity (vs token-level `slot_mapping`).

The harness includes:
- CUDA kernel supporting 6 GPU KV formats: Normal, Cross-layer, Flash Infer, MLA, SGLang MHA, SGLang MLA (bf16 + fp8)
- Pure Python reference implementation for correctness oracle
- D2H ↔ H2D roundtrip correctness tests (24 configs: 6 formats × 2 dtypes × 2 skip modes)
- CUDA-event-based benchmarking with throughput reporting
- PTX `ld.global.cs` / `st.global.cs` streaming copy to bypass L2 cache
- CLI with `--format`, `--dtype`, `--mem-device` (gpu/cpu), `--skip-prefix-n-blocks` flags

- The kernel uses a pointer-based interface (`paged_buffer_ptrs_tensor` + `lmcache_objects_ptrs`) rather than passing tensor objects directly
- For SGLang MHA, K/V distinction is handled by pointer selection (`ptrs[k_or_v * nl + layer_idx]`) rather than offset arithmetic, since K and V reside in separate tensors
- The harness builds independently via `python setup.py build_ext --inplace` in the kernel_harness directory
- Memory objects default to GPU; `--mem-device cpu` uses pinned CPU memory

**If applicable**:

### DIFF
--- a/lmcache/tools/kernel_harness/DESIGN.md
+++ b/lmcache/tools/kernel_harness/DESIGN.md
@@ -1,0 +1,76 @@
+# Design: Block Transfer Kernel Test Harness
+
+## Motivation
+
+The existing `multi_layer_kv_transfer` kernel operates at token granularity via `slot_mapping`. The new `multi_layer_block_kv_transfer` kernel operates at **block granularity** via `block_ids`, which is more natural for the MP (multi-process) mode where entire blocks of KV cache are transferred between GPU and CPU.
+
+## Architecture
+
+```
+__main__.py          CLI entry point, dispatches to correctness/benchmark
+    |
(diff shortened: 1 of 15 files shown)


ApostaC:
Please see #2838

PR #2833, opened 2026-03-20
[Core] Add S3 over RDMA channel in storage backend

### LINKED ISSUE #2875: [Feature request] Add S3-over-RDMA data plane via NVIDIA cuObject
 LMCache already supports S3 as a remote storage backend via S3Connector, which uses the AWS CRT library to transfer KV cache data over standard HTTP PUT/GET. However, for deployments with RDMA-capable networking (InfiniBand, RoCE) and a cuObject-enabled S3 endpoint (e.g. Dell ECS with GPUDirect Storage), the HTTP body transfer becomes the bottleneck — it incurs TCP/IP stack overhead, kernel buffer copies, and cannot exploit the network's native RDMA capability. • We need to add an RDMA-accelerated data plane to the S3 connector that uses NVIDIA's cuObject library (libcuobjclient.so) to transfer data directly between the client's pinned memory and the S3 server via RDMA, bypassing the HTTP body entirely.

NVIDIA's cuObject library (shipping with CUDA Toolkit >= 13.1.1) enables S3-over-RDMA by injecting RDMA tokens into standard S3 HTTP requests. The server reads from (or writes to) the client's registered memory region directly over RDMA — no data travels through the HTTP body. The S3 control plane (auth, signing, TLS) is unchanged; only the data plane changes. The cuObjClient API Specification v1.0.0 ([docs](https://docs.nvidia.com/gpudirect-storage/cuobject/cuObjClient-api/index.html)) documents two usage patterns: 
• Pattern A — Callback-based I/O (spec sections 1.2, 1.6, 1.10): cuObjPut()/cuObjGet() orchestrate the full I/O lifecycle via user-supplied callbacks. The callback receives an RDMA descriptor and must perform the entire server round-trip synchronously. 
• Pattern B — Manual RDMA Token Management (spec section 1.12.4): cuMemObjGetRDMAToken() generates a token string with caller-managed lifetime. The caller injects it into its own HTTP request as an x-amz-rdma-token header. 

Pattern B is the correct fit for LMCache because: (1) CRT's async event-loop model is incompatible with Pattern A's blocking callback requirement, (2) Pattern A's descriptor lifetime is limited to the callback scope (spec section 1.15), and (3) LMCache already has a full-featured S3 control plane — we only need the RDMA token, not the I/O orchestration.

Proposed design

┌──────────────────────────────────────────────────┐
│  CuObjectS3ConnectorAdapter                                                                               │  URL routing
│  Matches "cuobj+s3://" URLs                                                                                   │
│  Strips prefix, creates CuObjectS3Connector                                                           │
└───────────────────────┬──────────────────────────┘
                        │
┌───────────────────────▼──────────────────────────┐
│  CuObjectS3Connector (extends S3Connector)                                                       │  Data-plane override
│  Overrides _s3_upload / _s3_download only                                                            │
│  Injects x-amz-rdma-token into HTTP headers                                                        │
│  Falls back to parent HTTP body on any failure                                                       │
└───────────────────────┬──────────────────────────┘
                        │
┌───────────────────────▼──────────────────────────┐
│  CuObjClientWrapper (cuobject_bindings.py)                                                          │  Python wrapper
│  Pool registration, RDMA token generation,                                                            │
│  RDMA reply parsing (JSON / numeric / keyword)                                                  │
└───────────────────────┬──────────────────────────┘
                        │
┌───────────────────────▼──────────────────────────┐
│  CuObjectClient (C++ pybind11 extension)                                                             │  Build-time linked
│  Links against libcuobjclient.so at build time                                                           │
│  Pool tracking (base + size for offset calc)                                                               │
│  cuMemObjGetRDMAToken / cuMemObjPutRDMAToken                                       │
│  All methods release GIL                                                                                          │
└──────────────────────────────────────────────────┘

NA

**Additional context**
Performance expectations

RDMA eliminates three overheads present in HTTP body transfer:

  • TCP/IP stack processing (kernel → userspace copies)
  • HTTP chunked encoding/decoding
  • CRT's multipart upload/parallel download machinery (unnecessary for pinned memory)

For KV cache workloads with 256 KB–32 MB chunk sizes over InfiniBand/RoCE networks, RDMA should provide
significantly lower latency and higher throughput for PUT/GET operations compared to HTTP body transfer.

realAaronWu:
Fixes #2875 
Please see the issue description: https://github.com/LMCache/LMCache/issues/2875

**If applicable**:


sammshen:
do you wnat to try using the new native interface to build in cpp

sammshen:
you might have to undraft

realAaronWu:
Open PR: https://github.com/LMCache/LMCache/pull/2877

PR #2841, opened 2026-03-21
Clean up on-disk cache files in LocalDiskBackend.close()

### LINKED ISSUE #2840: LocalDiskBackend: clean up or persist on-disk cache files on shutdown

`LocalDiskBackend` stores KV cache chunks as individual files under the configured disk path. The key-to-file mapping is maintained in an in-memory dict (`self.dict`), which is not persisted to disk.

When the service shuts down, `close()` only stops the worker executor and message sender — it does not delete the cached files. Since the mapping dict is lost on exit, these files become orphaned and silently leak disk space.

On shutdown, either:

1. **Clean up**: Delete cached files under `self.path` in `close()` — simple fix, matches current design where the mapping is ephemeral.
2. **Persist index**: Write the mapping dict to an index file so a future instance can reload and reuse the cached data — higher value, enables cross-restart cache reuse.

- Require users to manually `rm -rf` the disk cache directory between restarts — current workaround but error-prone and undocumented.
- Add a startup-time scan that removes stale files not present in the mapping.

FeiGSSS:

`LocalDiskBackend` stores KV cache chunks as `.pt` files on disk, but the key-to-file mapping (`self.dict`) is in-memory only. When the service shuts down, `close()` stops the worker executor without deleting the cached files. Since the mapping is lost on exit, these files become orphaned and silently leak disk space.

This PR adds a `batched_remove()` call in `close()` to delete all cached files before stopping the executor. The `batched_msg_sender` is also set to `None` after closing to prevent stale eviction messages during cleanup.

Fixes #2840

Minimal change — two lines added to `close()`. No new APIs or behavioral changes at runtime.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -632,4 +632,6 @@ class LocalDiskBackend(StorageBackendInterface):
     def close(self) -> None:
         if self.batched_msg_sender is not None:
             self.batched_msg_sender.close()
+            self.batched_msg_sender = None  # prevent stale sends during cleanup
+        self.batched_remove(list(self.dict.keys()))
         self.disk_worker.close()


ianliuy:
Hi @FeiGSSS  great initiative filing the issue and jumping right into a fix!  I've been looking at this area and wanted to flag one concern before this gets merged.

**Ordering bug: `disk_worker` must be shut down *before* the file cleanup**

`LocalDiskWorker` holds up to 4 async worker threads that execute pending `put` tasks. At the time `batched_remove(list(self.dict.keys()))` runs, those threads may still be completing in-flight writes  each of which calls `async_save_bytes_to_disk`  `insert_key`, adding new entries to `self.dict` and writing new files to disk.

Because `batched_remove` snapshots `self.dict.keys()` *before* the worker drains, any files written by late-finishing workers will:
1. Not be in the snapshot  **never deleted** (the leak persists for those files)
2. Race with deletions of already-snapshotted files

The fix is to reverse the order  call `self.disk_worker.close()` first (which calls `executor.shutdown(wait=True)`, blocking until all queued tasks finish), then snapshot and delete:

```python
def close(self) -> None:
    if self.batched_msg_sender is not None:
        self.batched_msg_sender.close()
        self.batched_msg_sender = None

    self.disk_worker.close()            # drain in-flight writes first
    keys = list(self.dict.keys())       # now self.dict is fully up-to-date
    for key in keys:
        meta = self.dict.pop(key, None)
        if meta is None:
            continue
        try:
            os.remove(meta.path)
        except OSError as e:
            logger.warning("Failed to remove %s: %s", meta.path, e)
```

A few other things worth addressing before merge:
- **No error handling**: if `os.remove()` raises (file already deleted externally, permission error, etc.), the exception propagates out of `close()` and `disk_worker.close()` is never reached  leaking the thread pool.
- **No tests**: without tests this path has no regression protection.
- **DCO check failing**: the commit needs a `Signed-off-by:` trailer (use `git commit -s`).

Happy to collaborate or submit a follow-up PR if helpful. Wanted to flag the race condition before it ships. Thanks for finding this!

PR #2842, opened 2026-03-21
[MP][Feat] support worker-affinity in the MQ thread pool

ApostaC:

Introduces thread affinity routing for GPU-bound request handlers in the multiprocess message queue server. All STORE/RETRIEVE requests from the same vLLM instance (identified by zmq identity) are always dispatched to the same worker thread. This eliminates the need for `gpu_context.transfer_lock` since same-instance GPU transfers are now inherently serialized.

Also cleans up pool naming:
- **Affinity pool** — for GPU-bound handlers (STORE/RETRIEVE). Routes by `hash(zmq_identity) % N` so same-client work is serialized on one thread.
- **Normal pool** — for non-GPU blocking handlers (LOOKUP, END_SESSION, etc.). Standard `ThreadPoolExecutor`.
- The old default thread pool is removed; all blocking handlers must be explicitly assigned to a pool.

New CLI args (backward compatible):
- `--max-workers` sets both pools (existing behavior preserved)
- `--max-gpu-workers` overrides the affinity pool size
- `--max-cpu-workers` overrides the normal pool size

- `blend_server.py` is deprecated but updated minimally to avoid breakage
- `MPCacheEngine.lock` (used only in `clear()`) is unrelated and left as-is

**If applicable**:


sammshen (approved):
LGTM!

deng451e (approved):

PR #2826, opened 2026-03-19
[Core]: Support HND KV Format

sammshen:
## Test plan                                                                              
                                                                                          
  - [x] E2E test with `VLLM_KV_CACHE_LAYOUT=HND` (Qwen2.5-3B): basic completion, KV cache   
  reuse, deterministic output after cache reset                                           
  - [x] NHD regression test: deterministic output matches baseline                          
  - [x] E2E with flash infer backend (`NL_X_NB_TWO_NH_BS_HS`)                               
  - [x] Multi-process (tensor parallel) HND test      

---

> [!NOTE]
> **Medium Risk**
> Touches CUDA KV-cache transfer kernels and the multi-process registration protocol; incorrect layout detection or offset math could corrupt KV data or crash, but changes are localized and guarded by new checks/tests.
> 
> **Overview**
> Adds end-to-end support for vLLM’s **HND KV cache layout** (heads-before-block) alongside existing NHD/MLA formats.
> 
> Updates the CUDA transfer path (`multi_layer_kv_transfer`/`single_layer_kv_transfer`) to compute correct offsets for new HND `GPUKVFormat`s, adds required `head_size` plumbing/validation, and exposes the new formats + `head_size` arg via pybind.
> 
> Extends the vLLM integration and multi-process server protocol to pass `layout_hints` (e.g. `{"kv_layout": "HND"}`), auto-detect layout at runtime, and permute non-contiguous vLLM HND tensors back to a contiguous physical view before pointer capture/format discovery. Adds unit tests covering HND round-trips and updates existing registration tests for the new payload shape.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9325b08c6cd08e65671a19af953298e0ddee6f29. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #2857, opened 2026-03-23
[Feat] add mp support for sglang

DongDongJu:

Sglang side PR : https://github.com/sgl-project/sglang/pull/21229
  This PR adds LMCache-side support for SGLang multi-process (MP) mode while keeping the MP design generic and keeping SGLang-side changes minimal.

  Before this change, LMCache MP was still effectively vLLM-shaped:
  - `REGISTER_KV_CACHE` used a positional payload
  - `GPUCacheContext` still assumed vLLM layout
  - MP `RETRIEVE` could not express real layerwise progress for SGLang
  - SGLang integration was still using the older in-process layerwise path

  This PR fixes that by extending the existing LMCache MP path instead of adding a SGLang-only backend path.

  Main changes:
  - replace the positional `REGISTER_KV_CACHE` payload with a structured `KVCacheRegistration`
  - make `GPUCacheContext` discover and handle SGLang KV layout through GPUKVFormat
  - extend MP `RETRIEVE` with a generic layer window so SGLang can drive true `load_kv_layerwise()` behavior
  - add an LMCache-side SGLang MP connector that reuses existing MP client pieces:
    - MQ client setup
    - chunk-size fetch
    - lookup / prefetch polling
    - lookup lock release
    - session end
    - store / retrieve request flow
  - keep the in-process SGLang path unchanged
  - keep the protocol generic rather than adding SGLang-only request types

  Scope:
  - validated on the SGLang MHA path
  - in-process behavior remains unchanged
  - vLLM MP adapter is updated to use the new registration / retrieve protocol

  Example flow that this PR enables:

  1. Start LMCache MP daemon.
  2. Start SGLang with:
     - `--enable-lmcache`
     - `--lmcache-mp-host <host>`
     - `--lmcache-mp-port <port>`
  3. First request stores KV into LMCache MP.
  4. Second identical request retrieves KV from LMCache MP through SGLang’s existing layerwise connector contract.

  Example test setup used during validation:
  ```bash
  # LMCache MP daemon
  python -m lmcache.v1.multiprocess.server \
    --host 127.0.0.1 \
    --port 65000 \
    --l1-size-gb 10 \
    --eviction-policy LRU

  # SGLang server
  python -m sglang.launch_server \
    --model-path Qwen/Qwen3-0.6B \
    --enable-lmcache \
    --lmcache-mp-host 127.0.0.1 \
    --lmcache-mp-port 65000

  Observed behavior in the tested setup:

  - first request: cached_tokens=0
  - second identical request: non-zero cached_tokens
  - deterministic outputs matched in-process vs MP at temperature=0

  - This PR is intended to pair with the companion [SGLang PR](https://github.com/sgl-project/sglang/pull/21229).
  - The goal is to keep SGLang-specific translation at the connector boundary and keep LMCache MP generic.
  - I explicitly did not use the “transfer all layers at once and leave load_kv_layerwise() as a no-op” approach.
  - The protocol-level layer-window retrieve extension is generic, but the validated end-to-end SGLang path in this
    branch is the SGLang MHA layout.
  - No LMCache user docs were added in this PR; this PR focuses on the runtime / protocol / test path.

  Tested in this branch:

  - LMCache MP unit tests:
      - tests/v1/multiprocess/test_mq.py
      - tests/v1/multiprocess/test_cache_server.py
      - selected blend-server MP tests
  - deterministic in-process vs MP parity with temperature=0
  - end-to-end SGLang MP cache-hit validation on:
      - Qwen3-0.6B
      - facebook/opt-125m
      - Qwen3-8B
      - Qwen2.5-14B-Instruct
  - two-server validation:
      - two independent SGLang servers connected to the same LMCache MP daemon
      - verified cross-server KV reuse
      - verified a server can flush local cache and still retrieve from LMCache MP

  Representative results:

  - Qwen3-0.6B
      - second request: cached_tokens=768 in-process and MP
      - outputs matched byte-for-byte
  - facebook/opt-125m
      - second request: cached_tokens=1024 in-process and MP
      - outputs matched byte-for-byte
  - Qwen3-8B
      - second request: cached_tokens=768 in-process and MP
      - outputs matched byte-for-byte
  - Qwen2.5-14B-Instruct
      - prompt tokens: 811
      - second request: cached_tokens=800 in-process and MP
      - outputs matched byte-for-byte
```
  If applicable:

### DIFF
--- a/lmcache/v1/multiprocess/custom_types.py
+++ b/lmcache/v1/multiprocess/custom_types.py
@@ -238,6 +238,8 @@ class KVCacheRegistration:
             raise ValueError(f"Unsupported engine type: {self.engine_type}")
         if self.block_size <= 0:
             raise ValueError(f"block_size must be positive, got {self.block_size}")
+        if not _has_kv_cache_payload(self.kv_caches):
+            raise ValueError("kv_caches must contain at least one KV cache tensor")
 
 
 @dataclass
@@ -254,6 +256,14 @@ _CUSTOMERIZED_SERIALIZERS = {
         code=1,
     ),
 }
+
+
+def _has_kv_cache_payload(kv_caches: Any) -> bool:
+    if kv_caches is None:
+        return False
+    if isinstance(kv_caches, list | tuple):
+        return any(_has_kv_cache_payload(kv_cache) for kv_cache in kv_caches)
+    return True
 
 
 def get_customized_encoder(type: Any) -> msgspec.msgpack.Encoder:
--- a/lmcache/v1/multiprocess/custom_types.py
+++ b/lmcache/v1/multiprocess/custom_types.py
@@ -223,6 +223,24 @@ KVCache = list[CudaIPCWrapper]
 
 
 @dataclass
+class KVCacheRegistration:
+    """Structured MP registration payload for GPU KV cache workers."""
+
+    instance_id: int
+    model_name: str
+    world_size: int
+    engine_type: str
+    block_size: int
+    kv_caches: Any
+
+    def __post_init__(self) -> None:
+        if self.engine_type not in {"vllm", "sglang"}:
+            raise ValueError(f"Unsupported engine type: {self.engine_type}")
+        if self.block_size <= 0:
+            raise ValueError(f"block_size must be positive, got {self.block_size}")
+
+
+@dataclass
 class CustomizedSerdeConfig:
     serializer: Callable[[Any], bytes]
     deserializer: Callable[[bytes], Any]
--- a/lmcache/v1/multiprocess/gpu_context.py
+++ b/lmcache/v1/multiprocess/gpu_context.py
@@ -44,7 +44,15 @@ def unwrap_kv_cache_tensors(kv_caches: Any) -> Any:
         return [unwrap_kv_cache_tensors(kv_cache) for kv_cache in kv_caches]
     if isinstance(kv_caches, tuple):
         return tuple(unwrap_kv_cache_tensors(kv_cache) for kv_cache in kv_caches)
-    return kv_caches.to_tensor()
+    if isinstance(kv_caches, torch.Tensor):
+        return kv_caches
+    to_tensor = getattr(kv_caches, "to_tensor", None)
+    if callable(to_tensor):
+        return to_tensor()
+    raise TypeError(
+        "Expected KV cache leaf to be a torch.Tensor or CUDA IPC wrapper, "
+        f"got {type(kv_caches)}"
+    )
 
 
 def flatten_kv_cache_tensors(kv_caches: Any) -> list[torch.Tensor]:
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

File: lmcache/v1/multiprocess/custom_types.py [hunk 1]
```diff
@@ -222,6 +222,26 @@ def no_worker_id_version(self) -> "IPCCacheEngineKey":
...
 
+@dataclass
+class KVCacheRegistration:
+    """Structured MP registration payload for GPU KV cache workers."""
+
+    instance_id: int
+    model_name: str
+    world_size: int
+    engine_type: str
+    block_size: int
```
sammshen, line 233: Hi @DongDongJu I actually ran into a similar dilemma here on registration. Basically, how to clean the REGISTER protocol clean. 

A solution that I used in https://github.com/LMCache/LMCache/pull/2826 is to have a new field called `LayoutHints` which is a general `TypedDict` so that this protocol stays minimal

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -81,9 +81,11 @@ def get_protocol_definitions() -> dict[str, ProtocolDefinition]:
         #   - event_ipc_handle: bytes - CUDA event IPC handle for synchronization
         #   - skip_first_n_tokens: int - Number of tokens to skip writing at the
         #     start of the retrieve range (to avoid overwriting APC-shared blocks)
+        #   - layer_begin: int - Inclusive layer index to retrieve, or -1 for all
```
sammshen, line 84: IIRC, layerewise is deprioritized for MP mode

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -121,12 +178,16 @@ def __init__(self, kv_caches: KVCache, lmcache_chunk_size: int = 256):
...
+    def gpu_kv_format(self):
+        return self.gpu_kv_format_
+
     @property
     def device(self) -> torch.device:
         return self.device_
 
     @property
-    def kv_tensors(self) -> list[torch.Tensor]:
+    def kv_tensors(self) -> Any:
```
sammshen, line 190: is this to support `list[list[torch.Tensor]]`?

lmcache/v1/multiprocess/custom_types.py
DongDongJu: @sammshen Lets discuss which way is better and priority for this.

lmcache/v1/multiprocess/protocols/engine.py
DongDongJu: Good to know.
Cuz current sglang's in-process mode integration only support layerwise mode unless someone made the changes.
Lets just using normal mode for mp mode.
@ApostaC WDYT?

lmcache/v1/multiprocess/gpu_context.py
DongDongJu: Yes cuz unwrap_kv_cache_tensors cannot guarantee it anymore.
But if we have better design for this inconsistency (vllm, sglang ...) then I will follow that.

DongDongJu:
Since @Shaoting-Feng working in https://github.com/LMCache/LMCache/pull/3166, Let me close the thread.

PR #2799, opened 2026-03-17
Add hipFile support for AIS (AMD Infinity Storage) storage

glimchb:
This commit adds comprehensive support for AMD hipFile, providing the ROCm equivalent to NVIDIA cuFile for GPU-direct storage operations.

- **HipFileMemoryAllocator**: New memory allocator for AMD GPUs
- **GDS Backend Integration**: Added use_hipfile configuration option
- **ROCm Device Detection**: Properly detects HIP devices via torch.version.hip
- **Automatic Allocator Selection**: Backend chooses between cuFile/hipFile

- Seamless integration with existing GDS backend
- Configuration via extra_config: {"use_hipfile": true}
- Automatic fallback to cuFile for NVIDIA systems
- Full test coverage with mock-based tests
- Complete documentation with examples

- lmcache/v1/memory_management.py: Added HipFileMemoryAllocator
- lmcache/v1/storage_backend/gds_backend.py: Added hipFile support
- lmcache/v1/config.py: Added use_hipfile config option
- lmcache/v1/cache_engine.py: Updated allocator selection logic
- tests/v1/utils.py: Added has_hipfile() function
- tests/v1/*: Updated tests to support both backends
- docs/*: Updated documentation for hipFile support
- requirements/common.txt: Added hipfile-python dependency

extra_config:
  use_hipfile: true

export LMCACHE_EXTRA_CONFIG='{"use_hipfile": true}'

The implementation maintains full backward compatibility while enabling GPU-direct storage on AMD ROCm platforms.

Resolves: Support for AMD GPU-direct storage in LMCache

**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #2858, opened 2026-03-23
[refactor]: Replace use_cufile with use_gds/gds_backend config flags

glimchb:
Follow up work post #2799 (AMD hipFile AIS support).

Refactor the GDS backend configuration to be backend-agnostic:
- Replace extra_config["use_cufile"] with top-level use_gds (bool) config
- Add gds_backend config field ("cufile" default) to select GDS library
- Rename cufile_buffer_size to gds_buffer_size
- Rename internal attributes: self.cufile -> self.gds_module, self._cufile_driver -> self._gds_driver, self.cufile_base_pointer -> self.gds_base_pointer
- Update tests and documentation accordingly

New env vars: LMCACHE_USE_GDS, LMCACHE_GDS_BACKEND, LMCACHE_GDS_BUFFER_SIZE

Generated with [Devin](https://cli.devin.ai/docs)

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Touches the `GdsBackend` initialization and allocator selection paths, which are performance/IO-critical and could change runtime behavior (e.g., auto-disabling GDS on certain filesystems or selecting the wrong backend). Overall scope is contained to configuration and GDS backend wiring, with tests/docs updated accordingly.
> 
> **Overview**
> Refactors GDS configuration to be backend-agnostic: replaces `extra_config` flags (`use_cufile`/`use_hipfile`) with top-level `use_gds` plus a `gds_backend` selector (`cufile`/`hipfile`), and renames `cufile_buffer_size` to `gds_buffer_size` (including new env vars like `LMCACHE_GDS_BUFFER_SIZE`, `LMCACHE_USE_GDS`, `LMCACHE_GDS_BACKEND`).
> 
> Updates `GdsBackend` to route imports/driver setup through the chosen backend module, standardizes internal naming (`gds_module`, `_gds_driver`, `gds_base_pointer`), and tweaks filesystem-based auto-disable logic to only override defaults when `use_gds` wasn’t explicitly set. Tests, sample configs, Buildkite config, and docs are updated to the new keys and behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 217b6b949931fde108c32df154b30eb780cdcd9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/storage_backend/gds_backend.py
+++ b/lmcache/v1/storage_backend/gds_backend.py
@@ -148,6 +148,18 @@ async def save_metadata(path: str, tmp: str, metadata: bytes):
 
 
 def get_extra_config_bool(key, config: LMCacheEngineConfig) -> bool | None:
+    """Extract a boolean value from the config's extra_config dict.
+
+    Args:
+        key: The key to look up in extra_config.
+        config: The LMCacheEngineConfig instance.
+
+    Returns:
+        The boolean value if present, or None if not set.
+
+    Raises:
+        RuntimeError: If the value is not a valid boolean representation.
+    """
     value = config.extra_config.get(key, None)
     if value is None:
         return None
@@ -166,17 +178,21 @@ def get_extra_config_bool(key, config: LMCacheEngineConfig) -> bool | None:
 class GdsBackend(AllocatorBackendInterface):
     """
     Originally based on the open sourced WekaGdsBackend, this is a backend that
-    leverages NVIDIA's cuFile API to issue GDS requests directly to the
+    leverages GPU Direct Storage APIs to issue GDS requests directly to the
     GDS-supported remote filesystem.  In order to use it, users need to specify
-    `gds_path` and `cufile_buffer_size` in their LMCache config.
+    `gds_path` and `gds_buffer_size` in their LMCache config.
+
+    The GDS library to use is controlled by the `gds_backend` config field
+    (default: ``"cufile"``). Setting ``use_gds=False`` disables the GDS API
+    and falls back to POSIX I/O via cudart.
 
     Cache Directory Structure created by this Backend:
     /{gds_path}/{first_level}/{second_level}/{data & metadata} This structure
     is semi-arbitrary. We create two levels in the directory hierarchy to
     parallelize loading the data during initialization in the Python code.
 
-    NOTE: If GPUDirect is not supported on that other filesystem, then CuFile will
-    fall back to POSIX I/O.
+    NOTE: If GPUDirect is not supported on that other filesystem, then the GDS
+    library will fall back to POSIX I/O.
     """
 
     def __init__(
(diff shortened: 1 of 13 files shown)


lmcache/v1/storage_backend/gds_backend.py
```diff
@@ -217,50 +233,41 @@ def __init__(
...
+        self.gds_backend = config.gds_backend
+        # _user_set_keys is populated by LMCacheEngineConfig when a field is
+        # explicitly provided via a config file, environment variable, or
+        # keyword argument — as opposed to falling back to its default value.
+        # We check it here so that the tmpfs/overlayfs auto-disable logic
+        # below can distinguish "the user never mentioned use_gds (default
+        # True)" from "the user explicitly wrote use_gds: true".  In the
+        # first case we auto-disable GDS on unsupported filesystems; in the
+        # second we respect the user's explicit intent and leave it enabled.
+        user_set_keys: set[str] = getattr(config, "_user_set_keys", set())
```
DongDongJu, line 240: I didnt get it that well. what is this variable for? is this for k8s?
Why dont we just pass this by config?

DongDongJu (commented):
Thanks for this refactoring. I left one question.

lmcache/v1/storage_backend/gds_backend.py
glimchb: The old approach used extra_config (a dict), where get_extra_config_bool("use_cufile", config) returned None when the key was absent — so
distinguishing "not set" from "set to true" was trivial.

The new approach promotes use_gds to a top-level config field, which always has a value (default True). There's no None sentinel, so we rely on
_user_set_keys instead.

This is better because:

  • Discoverability — top-level fields appear in --help, config validation, env var listing (LMCACHE_USE_GDS), and documentation tables automatically.
  extra_config keys are invisible to all of that.
  • Validation — the config system validates types and applies env_converter for top-level fields. extra_config is a raw dict with no schema — a typo
  like use_cufile: tru silently does nothing.
  • Consistency — every other GDS setting (gds_path, gds_buffer_size, gds_backend) is top-level. Having use_gds buried in extra_config would be the
  odd one out.

The _user_set_keys check is slightly more indirect than the old is not None check, but the config infrastructure already tracks it — we're just using
what's there.

sammshen (approved):
LGTM! please address @DongDongJu's comments

lmcache/v1/storage_backend/gds_backend.py
DongDongJu: oh I got what it means in here.
But here's  `use_gds_explicitly_set ` and  `self.use_gds` eventually same value. why we need both?

glimchb: - for `tmpfs/overlayfs`, GDS APIs don't work
- if user didn't specify anything in the config, then `self.use_gds=True` (default) but we apply **automatic disabling** and set `self.use_gds = False`
- if user did specify `self.use_gds=True` then **automatic disabling** is not working and `self.use_gds=True` remains

see this code
```
        if self.fstype in ["tmpfs", "overlayfs"]:
            # TODO: we can replace the auto-detection of unsupported GDS
            # file systems by doing a small GDS API test on them. If a
            # read/write test fails, we can fallback to not using GDS APIs.
            if use_gds_explicitly_set:
                logger.warning("No automatic disabling of GDS usage due to fstype")
            else:
                logger.info("Automatic disabling of GDS usage due to fstype")
                self.use_gds = False
        elif self.fstype == "wekafs":
            logger.info("Weka filesystem detected, GDS usage is enforced")
            assert self.use_gds
```

I agree this is not easy logic, but it was there before my change. If you think we should get rid of it - let me know. we can simplify this a lot by saying for `tmpfs/overlayfs` there is no GDS....

I can do like this
```
if self.fstype in ["tmpfs", "overlayfs"]:
    logger.info(f"GDS is not supported on {self.fstype}. ")
    assert not self.use_gds
elif self.fstype == "wekafs":
    logger.info("Weka filesystem detected, GDS usage is enforced")
    assert self.use_gds
```

wdyt? happy to change according to feedback

DongDongJu: I see. so basically, overall logic is like this.
```
if config.use_gds is None:
    self.use_gds = self.fstype not in ["tmpfs", "overlayfs"]
else:
    self.use_gds = config.use_gds
```

But I bliv good to having a comments for each case.
LGTM

DongDongJu (approved):

lmcache/v1/storage_backend/gds_backend.py
glimchb: @DongDongJu do you want this change ?
```
if self.fstype in ["tmpfs", "overlayfs"]:
    logger.info(f"GDS is not supported on {self.fstype}. ")
    assert not self.use_gds
elif self.fstype == "wekafs":
    logger.info("Weka filesystem detected, GDS usage is enforced")
    assert self.use_gds
```
or this change 
```
if config.use_gds is None:
    self.use_gds = self.fstype not in ["tmpfs", "overlayfs"]
else:
    self.use_gds = config.use_gds
``` 
or leave the code as-is ?

Issue #2221, opened 2025-12-10
Make lmcache kvcache shape agnostic


maobaolong:
**Summary**

This is a high level proposal, maybe this need something change not only in lmcache but also in vllm or vllm_v1_adapter,
but if we achieved this goal, we can never need extra efforts while new attention or new model publish.

**Details**

- Never record or try to parse the shape of kvcache, we need an abstraction to make it easy to get the thing we need, something like, layer count, kv count(1 or 2), head count, head size dtype for each layer.
- CacheEngine GpuConnector MemoryAllocator Backends should not care the concrete shape of kvcache, just use a shape to allocate buffer or exchange kvcache.

**Expected Outcome / Goal**

Reduce adaptive efforts while new attention or model publish

DongDongJu:
@maobaolong I felt similar sense before. but what you mean is like abstract by bytes-level? then we can not apply any KVCache optimization in LMCache layer.
@ApostaC what do you think?

DongDongJu:
I agreed with @maobaolong's idea.
So, I tried to think about this design.
How about like format with bytes?
```
KVFormat {
format_type: "vllm...", "sglang...",
plans_per_layer:[
{ name: "k", dtype: "", axes: [ "token","kv_head","head_dim"...]},
{ name: "v" ...},
]
chunk_tokens: 256,
sharding: { axis: "kv_head", world_size: .., rank: ..}
}
```
and bytes data or page based?

KuntaiDu:
Totally agree and this is important.

Some high-level goals:
- Unifying code path: use the same path of code to handle
  - normal KV cache
  - MLA
  - deepseek V3.2
  - (better to have) sliding window attention, mamba
- (better to have) Improving performance
  - only load the KV cache for the tokens inside sliding window, or only load the last token for Mamba layer
  - only save the KV cache for Mamba layers for the last token in the chunk


On this front, vLLM-side has an abstraction called `KVCacheGroups`, which basically groups and unifies the processing logic for layers with the same tensor shape and same  has some high-level interfaces, maybe we can also inherent some of its design.

Retrieve example:
```
The full KV caches descriptor of the model 
--chunker--> The chunked KV caches descriptor of the model 
--sharder--> A list of hashes, where each hash only corresponds to one contiguous tensor 
--key-value-store--> tensors
--reverse_sharder--> chunked kv hash
--reverse_chunker--> full KV cache
```

[...]

PR #2863, opened 2026-03-25
Support hybrid KV cache models (Mamba + attention) in GPU connector V3

### LINKED ISSUE #2845: Hybrid models like qwen3.5 9B are not supported
**Label**

I am trying to deploy the Model Qwen3.5 9B in vLLM vith LMcache but I hit this senario where Hybrid models are not suggested.
The quickstart(https://docs.lmcache.ai/getting_started/quickstart.html) explicitly uses Qwen/Qwen3-8B-Instruct (pure transformer) and even stated that --disable-hybrid-kv-cache-manager is mandatory but its not called out clearly in the docs

**Additional context**

oceanplexian:
## Summary
Adds support for hybrid KV cache models (Mamba/GDN + attention) in the V3 GPU connector. Models like Qwen3.5, Falcon-H1, and Jamba store multiple state tensors per recurrent layer, which crashes `build_kv_layer_groups` and `VLLMPagedMemGPUConnectorV3`. Fixes #2845. Related: vllm-project/vllm#36771, #2221. 

**Changes**

  - Import `SupportsHMA` and add it to `LMCacheConnectorV1` class bases
  - Implement `request_finished_all_groups()` which combines block IDs from all KV cache groups and delegates to the existing `request_finished` handler

  ## Testing

  - [x] `pytest tests/v1/test_kv_layer_groups_manager.py` — 10/10 pass (9 existing + 1 new)
  - [x] `ruff check` + `ruff format` clean
  - [x] Qwen3.5-35B-A3B-GPTQ-Int4 (GDN + attention, 30 recurrent + 10 attn layers) on 2x RTX 3090, TP=2, 256K context, LMCache V3 + vllm + prefix caching, Tested 1-8 parallel requests
  - [x] Falcon-H1-7B-Instruct (Mamba-2 + attention, 44 recurrent + 44 attn layers). Same setup, Tested 1-8 parallel requests

---

> [!NOTE]
> **Medium Risk**
> Touches GPU KV transfer and cache-hit scheduling logic; incorrect layer filtering or hybrid detection could silently disable cache loads or corrupt transfers for some models.
> 
> **Overview**
> Improves support for *hybrid models* (attention + recurrent/Mamba-style layers) by teaching `KVLayerGroupsManager` to **skip unsupported multi-tensor layer states** instead of raising, and updating `VLLMPagedMemGPUConnectorV3` to **only build pointers/format info from attention KV tensors**.
> 
> On the vLLM integration side (`vllm_v1_adapter.py`), the connector now **detects non-attention layers** (via `model_config.is_hybrid` and a worker-side layer-group count check) and **disables external KV cache loading** (`get_num_new_matched_tokens` returns `0`) when such layers are present. A new unit test validates recurrent-layer skipping in layer group construction.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d4c672290ae1cef2c2822f10c676b7cf0b26193e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -538,6 +538,9 @@ class LMCacheConnectorV1Impl:
         self._check_legacy_register_kv_caches()
 
         self.kv_caches: dict[str, torch.Tensor] = {}
+        self._has_non_attention_layers = getattr(
+            vllm_config.model_config, "is_hybrid", False
+        )
         self._block_size = vllm_config.cache_config.block_size
         self.load_specs: dict[str, LoadSpec] = {}
         self.kv_cache_manager: Optional["KVCacheManager"] = None
@@ -705,6 +708,20 @@ class LMCacheConnectorV1Impl:
                 self.lmcache_engine.metadata.kv_layer_groups_manager
             )
             kv_layer_groups_manager.build_kv_layer_groups(self.kv_caches)
+
+            # Also detect from layer groups (worker-side confirmation)
+            total_grouped = sum(
+                len(g.layer_indices)
+                for g in kv_layer_groups_manager.kv_layer_groups
+            )
+            if total_grouped < len(self.kv_caches):
+                self._has_non_attention_layers = True
+                logger.warning(
+                    "Hybrid model: %d/%d non-attention layers, "
+                    "cache load disabled",
+                    len(self.kv_caches) - total_grouped,
+                    len(self.kv_caches),
+                )
 
     # TODO(chunxiaozheng): in the latest lmcache_connector, we use `register_kv_caches`
     #  to init self.kv_caches, we keep it in order to be compatible with old versions
@@ -1225,6 +1242,11 @@ class LMCacheConnectorV1Impl:
         # Ignore DP attention mock requests
         if request.request_id.startswith("mock_req"):
             return 0
+
+        # Hybrid models: can't restore non-attention layers from cache
+        if self._has_non_attention_layers:
+            return 0
+
         # to handle preempted requests, we want `get_num_new_matched_tokens` to be
         # idempotent under the condition that `update_state_after_alloc` is NOT called
         # then the two side-effects that must be idempotent are:
--- a/lmcache/v1/gpu_connector/gpu_connectors.py
+++ b/lmcache/v1/gpu_connector/gpu_connectors.py
@@ -425,6 +425,10 @@ class VLLMPagedMemGPUConnectorV3(GPUConnectorInterface):
         if self.init:
             return
         assert self.metadata.kv_layer_groups_manager.kv_layer_groups
+
+        # Skip recurrent layers (multi-tensor state per layer)
+        attn_kvcaches = [kv for kv in self.kvcaches if isinstance(kv, torch.Tensor)]
+
         if self.use_gpu:
             # init tmp buffer
             tmp_buf_shapes = self.metadata.get_shapes(self.chunk_size)
@@ -437,14 +441,21 @@ class VLLMPagedMemGPUConnectorV3(GPUConnectorInterface):
                 )
             ]
         self.group_kv_cache_pointers_on_gpu = []
+
+        # Map original indices to filtered indices
+        attn_idx_map = {}
+        attn_i = 0
+        for i, kv in enumerate(self.kvcaches):
+            if isinstance(kv, torch.Tensor):
+                attn_idx_map[i] = attn_i
+                attn_i += 1
+
         for group in self.metadata.kv_layer_groups_manager.kv_layer_groups:
             # init kv cache pointers
             num_layers = group.num_layers
             kv_cache_pointers = torch.empty(num_layers, dtype=torch.int64, device="cpu")
             kv_cache_pointers.numpy()[:] = [
-                t.data_ptr()
-                for i, t in enumerate(self.kvcaches)
-                if i in group.layer_indices
+                attn_kvcaches[attn_idx_map[i]].data_ptr() for i in group.layer_indices
             ]
             kv_cache_pointers_on_gpu = torch.empty(
                 num_layers, dtype=torch.int64, device=self.device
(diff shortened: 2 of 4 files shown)


Syan0o0:
Does this only store the attention layer? Do we need to store the Mamba layer?

oceanplexian:
Yes, there's an unfortunate trade-off in this approach. 

Mamba layers store multiple tensors with different shapes per layer, so they can't be represented as a single address per layer. Supporting them would require a much larger change.  This only offloads the Attention KV Cache but keeps a small amount of the Mamba state on the GPU (a few hundred KB per layer). 

I'm glad to look into it but didn't want to come in cold with a massive PR could impact a lot of other models. Thanks!

shengyan-chen:
If a new request arrives and hits the lmcache cache across requests, but the lmcache only stores the attention layer, won't directly reusing it cause accuracy issues?

malaiwah:
Hey @oceanplexian — nice work on this, we've been hitting the exact same problem with Qwen3.5-4B-FP8 (hybrid mamba+attention, 24 recurrent + 8 attention layers) on our end.

To answer @shengyan-chen's question directly — **yes, this causes accuracy issues as-is.** We've been running extensive A/B tests and can confirm:

When LMCache reports N cached tokens via `get_num_new_matched_tokens()`, the vLLM scheduler sets `request.num_computed_tokens = N` for **all** layer types. The forward pass then skips the first N tokens entirely — which is correct for attention layers (KV restored from cache), but the mamba recurrent state for those N tokens was never loaded. The mamba layers start from uninitialized state, producing garbled output.

The code path:
1. `gpu_connectors.py` filters to attention-only → correct, prevents crash ✓
2. `get_num_new_matched_tokens()` still reports the full token count → scheduler skips mamba computation
3. `mamba_utils.py:186`: `prev_state_idx = (num_computed_tokens - 1) // block_size` → points to a mamba block that was never populated
4. Forward pass produces garbage reasoning / word salad / repetition loops

We measured ~40% garbled output rate on identical requests with LMCache load enabled vs 0% on vanilla vLLM (no caching) with the same model and params. The fix we applied on our side:

```python
# In vllm_v1_adapter.py, _build_kv_layer_groups():
if skipped > 0:
    self._has_non_attention_layers = True
    # ... existing filtering ...

# In get_num_new_matched_tokens():
if self._has_non_attention_layers:
    return 0  # Can't claim a hit when mamba state is missing
```

This way LMCache still saves attention KV (useful for future full mamba support), but doesn't report cache hits for hybrid models where it can't restore the complete state. After the fix: 0/15 garbled, matching vanilla vLLM.

For context, we're running this in a trifecta setup (LMCache CPU + llm-d disk offload via MultiConnector). llm-d's `OffloadingConnector` saves all layer types including mamba state, so it handles hybrid models correctly. The corruption was specifically in the LMCache load path.

shengyan-chen:
> Hey @oceanplexian — nice work on this, we've been hitting the exact same problem with Qwen3.5-4B-FP8 (hybrid mamba+attention, 24 recurrent + 8 attention layers) on our end.
> 
> To answer @shengyan-chen's question directly — **yes, this causes accuracy issues as-is.** We've been running extensive A/B tests and can confirm:
> 
> When LMCache reports N cached tokens via `get_num_new_matched_tokens()`, the vLLM scheduler sets `request.num_computed_tokens = N` for **all** layer types. The forward pass then skips the first N tokens entirely — which is correct for attention layers (KV restored from cache), but the mamba recurrent state for those N tokens was never loaded. The mamba layers start from uninitialized state, producing garbled output.
> 
> The code path:
> 
> 1. `gpu_connectors.py` filters to attention-only → correct, prevents crash ✓
> 2. `get_num_new_matched_tokens()` still reports the full token count → scheduler skips mamba computation
> 3. `mamba_utils.py:186`: `prev_state_idx = (num_computed_tokens - 1) // block_size` → points to a mamba block that was never populated
> 4. Forward pass produces garbage reasoning / word salad / repetition loops
> 
> We measured ~40% garbled output rate on identical requests with LMCache load enabled vs 0% on vanilla vLLM (no caching) with the same model and params. The fix we applied on our side:
> 
> ```python
> # In vllm_v1_adapter.py, _build_kv_layer_groups():
> if skipped > 0:
>     self._has_non_attention_layers = True
>     # ... existing filtering ...
> 
> # In get_num_new_matched_tokens():
> if self._has_non_attention_layers:
>     return 0  # Can't claim a hit when mamba state is missing
> ```
> 
> This way LMCache still saves attention KV (useful for future full mamba support), but doesn't report cache hits for hybrid models where it can't restore the complete state. After the fix: 0/15 garbled, matching vanilla vLLM.
> 
> For context, we're running this in a trifecta setup (LMCache CPU + llm-d disk offload via MultiConnector). llm-d's `OffloadingConnector` saves all layer types including mamba state, so it handles hybrid models correctly. The corruption was specifically in the LMCache load path.
> 
> Happy to contribute the `get_num_new_matched_tokens` guard if useful — it's a small change but critical for correctness on hybrid models.

That is indeed a solid approach to circumventing accuracy issues. Furthermore, I'm curious about the TTFT (Time to First Token) performance gains of the adaptation in LMCache compared to vLLM's prefix matching.

malaiwah:
@shengyan-chen updated with LMCache CPU TTFT numbers. All measurements on Creativity (RTX 4080 Super), Qwen3.5-4B-FP8, same machine:

**TTFT by cache tier (apples-to-apples):**

| Prompt Size | Cold (no cache) | LMCache CPU | APC (in-GPU) | Speedup (LMCache vs cold) |
|---|---|---|---|---|
| ~2k tokens | 0.55s | 0.076s | 0.052s | 7× |
| ~6k tokens | 0.63s | 0.078s | 0.071s | 8× |
| ~21k tokens | 1.88s | 0.175s | 0.101s | 11× |

LMCache CPU consistently delivers 7-11× TTFT improvement over cold prefill, with 95% cache hit rate. APC is slightly faster (in-GPU, no transfer), but both APC and LMCache CPU are lost on restart — they're session-only. The key advantage of LMCache CPU is offloading KV from VRAM to host RAM, freeing GPU memory for more concurrent requests while still serving repeat prefixes fast.

For cross-restart persistence, we use llm-d (OffloadingConnector) with shared NFS storage. llm-d saves all layer types including mamba state to disk, getting 97% cache hit on cold boot. A future LMCache disk tier (`local_disk` config) would add another persistence option within LMCache itself.

**Important caveat for hybrid models:** these LMCache CPU numbers only restore attention KV (8/32 layers). The remaining 24 mamba layers are recomputed. A full mamba-aware LMCache that could restore all 32 layers would bring TTFT even closer to APC levels.

For the full multi-tier stack and test methodology: #2879 (LMCache fix), [vllm-project/vllm#38230](https://github.com/vllm-project/vllm/issues/38230), [llm-d/llm-d-kv-cache#465](https://github.com/llm-d/llm-d-kv-cache/issues/465).

sammshen:
@cursor review

hcyz33:
Do you have an estimated timeline for when this PR will be merged?

WenbinHou:
Looking forward to merging this PR. 👀

PR #2864, opened 2026-03-25
[CI]: Fix the LMCache random throughput being higher than native vllm

sammshen:
The MP throughput benchmark consistently shows LMCache ~46% faster than baseline on random workloads — where it should show a slight slowdown

Root cause: lm-eval sends 600 requests to the LMCache server but none to the baseline. With VLLM_ENABLE_V1_MULTIPROCESSING=0, the first batch of requests to a cold server incurs ~25s of tokenizer/template compilation overhead that gets included in the benchmark duration.

Verified on this machine: same server, same GPU, cold=81.50s vs warm=55.57s — delta of exactly 25.93s.

Fix:
  - Send 3 warmup chat completion requests to both servers before benchmarking
  - Add upper-bound check that fails if LMCache is >10% faster than baseline on
  random workload (catches future measurement asymmetries)


ApostaC (approved):
LGTM!

deng451e (approved):

PR #2867, opened 2026-03-25
fix: relax worker port count assertion

can-sun:
## Problem

The strict `==` assertions on `lmcache_worker_ports` length in `LMCacheWorker` reject valid configurations where more ports are provided than the effective `world_size` or worker count.

This is problematic in two scenarios:
1. **MLA models** reduce the effective `world_size` (vLLM divides by `tp_size`), so users providing ports based on total GPU count (TP × PP) hit the assertion. For example, on a p5 instance with 8 GPUs, providing 8 ports fails when MLA reduces `world_size` to 1.
2. **Config reuse across instance types** — a single config with 8 ports (for p5) fails on `g5.24xlarge` which only has 4 GPUs.

## Fix

Change `==` to `>=` in both assertions in `worker.py`:
- `len(config.lmcache_worker_ports) >= metadata.world_size`
- `len(config.lmcache_worker_ports) >= len(lmcache_worker_ids)`

Extra ports are simply unused — they are never bound to sockets or registered with the controller. The port selection logic indexes by `worker_id` or `lmcache_worker_ids.index()`, so extra entries are harmless. This is consistent with how P2P ports (`p2p_init_ports`, `p2p_lookup_ports`) already work.

Split out from #2732 per reviewer feedback.

---

> [!NOTE]
> **Low Risk**
> Low risk: only loosens config validation for `lmcache_worker_ports` without changing runtime socket binding logic; main risk is allowing misconfigured port lists that are too short for the selected workers to slip through until index errors occur elsewhere.
> 
> **Overview**
> Relaxes `LMCacheWorker` startup validation so `lmcache_worker_ports` can contain *more* entries than the effective `world_size` or explicit `lmcache_worker_ids` list.
> 
> The strict `==` length checks are replaced with `>=`, with updated inline documentation explaining why extra ports are safe (especially for MLA where vLLM may report a reduced `world_size`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 617c8655162fe683e8e8ebf9eddb5023f13f9dd7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sammshen:
I think this makes sense but could you add some documentation in the code on exactly how the number of worker ports is decided and the semantics of the world_size for different models above the conditional?

DongDongJu (commented):
So IIUC, this is user error/fault which is that user configured more than actual # of gpu needed at workload.
Is this true? then not sure that why we strictly design this at the first time.
@ApostaC Do you have any idea on this?

sammshen:
@cursor review

sammshen:
@cursor review

sammshen (approved):
LGTM

DongDongJu (approved):
LGTM

PR #2877, opened 2026-03-26
[Core] Add S3 over RDMA channel in storage backend

### LINKED ISSUE #2875: [Feature request] Add S3-over-RDMA data plane via NVIDIA cuObject

realAaronWu:
Fixes #2875 
Please see the issue description: https://github.com/LMCache/LMCache/issues/2875

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Adds a new RDMA data path for S3 PUT/GET using cuObject token headers plus optional native extension build logic; failures could impact data transfer correctness or connector selection/fallback behavior for S3-compatible endpoints.
> 
> **Overview**
> Adds **RDMA-accelerated S3 transfers** via NVIDIA cuObject by introducing a new optional `lmcache_cuobject` pybind11 extension (C++ `CuObjectClient` for pool registration and RDMA token generation) and Python wrappers that generate/validate `x-amz-rdma-token`/`x-amz-rdma-reply` headers.
> 
> Introduces a new `cuobj+s3://` connector adapter/connector that registers the local pinned-memory pool once, performs PUT/GET with empty/ignored HTTP bodies while verifying RDMA completion, and **falls back to standard HTTP S3** on init or per-request RDMA failures; also adds `s3_bucket` support for path-style S3 URLs and documents Dell ObjectScale RDMA setup.
> 
> Updates `setup.py` to auto-detect CUDA>=13.1 cuObjClient SDK and conditionally build/link the cuObject extension, and adds extensive unit tests for the bindings, adapter discovery, and RDMA upload/download behaviors (plus minor formatting-only tweaks elsewhere).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84400686c946db89b3b9738df24fb36c0f84de0a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/lookup_client/record_strategies/base.py
```diff
@@ -142,9 +142,11 @@ def setup_metrics(self, prometheus_logger) -> None:
             lambda: self.unique_chunks_count
         )
         prometheus_logger.chunk_statistics_reuse_rate.set_function(
-            lambda: (self.total_chunks - self.unique_chunks_count) / self.total_chunks
-            if self.total_chunks > 0
-            else 0.0
+            lambda: (
```
sammshen, line 145: note: pure formatting

sammshen:
this looks great. could you add some user facing docs?

realAaronWu:
> this looks great. could you add some user facing docs?

OK, will do.

csrc/storage_backends/cuobject/cuobject_client.cpp
```diff
@@ -0,0 +1,144 @@
...
+    oss << "Data region [0x" << std::hex << data_ptr << ", 0x"
+        << (data_ptr + size) << ") is outside registered pool [0x"
+        << pool_base_ << ", 0x" << (pool_base_ + pool_size_) << ")";
+    throw std::runtime_error(oss.str());
+  }
+
+  size_t buffer_offset = data_ptr - pool_base_;
+  char *desc_str = nullptr;
+
+  cuObjErr_t rc = client_->cuMemObjGetRDMAToken(
```
jgoldsch12, line 122: Using this approach of using one memory region and getting descriptors for each sub-region has a limitation where if the size of the sub-region is larger than per_buffer_cache_size_kb (for GPU memory) will result in a CU_FILE_MEMORY_NOT_REGISTERED error.  The default is is 10MiB I think, with a maximum value of 16MiB.  I believe CPU memory has different requirements for memory region registration.

realAaronWu: Thanks for raising this! Per my understanding, the per_buffer_cache_size_kb only applies to the unregistered GPU buffers, but in our case, we register the entire CPU pinned memory pool at init via cuMemObjGetDescriptor(), we do not tranfer data directly from GPU to S3, due to the scattered KV cache pages and the scarce GPU memory. If later we need to extend the direct GPU to S3 capability, we of cource need a regestered GPU memory. I have not done the testing for this scenrario. Have you experienced this error on CPU registered memory?

sammshen:
@cursor review

csrc/storage_backends/cuobject/cuobject_api.h
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// cuObject API header.
```
jgoldsch12, line 3: Putting a blocking comment in place, so we can discuss this PR's approach prior to merging.  Please do not merge until this review is complete.

jgoldsch12 (changes requested):

realAaronWu:
Need more work on the testing, dockerfile and doc things, convert to draft.

setup.py
```diff
@@ -94,6 +123,10 @@ def cuda_extension() -> tuple[list, dict]:
     fs_sources = [
         "csrc/storage_backends/fs/pybind.cpp",
         "csrc/storage_backends/fs/connector.cpp",
+        ]
+    cuobject_sources = [
+        "csrc/storage_backends/cuobject/pybind.cpp",
+        "csrc/storage_backends/cuobject/cuobject_client.cpp",
```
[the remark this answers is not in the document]
realAaronWu: The indentation is correct — cuobject_sources, ext_modules, and the cuObject if/else block are all at the same 4-space indent level inside def cuda_extension() (line 135). There's no inner 6-space block;ext_modules is always in scope when ext_modules.append(...) runs. For sdist builds,_find_cuobject_paths() returns (None, None) so the extension is skipped cleanly.

DongDongJu (commented):
Hello @realAaronWu,
Thanks for the amazing work.
I would suggest that split the generic S3/ObjectScale path-style support and Docker packaging out unless they are strictly required for the first usable cuObject path.

realAaronWu:
> Hello @realAaronWu, Thanks for the amazing work. I would suggest that split the generic S3/ObjectScale path-style support and Docker packaging out unless they are strictly required for the first usable cuObject path.

@DongDongJu  thanks for your suggestion! The docker packaging is needed because it patches the libcufile [here](https://github.com/realAaronWu/LMCache/blob/0d25d105d27cf1ee8695f5eabc1d0c5005be4326/docker/Dockerfile.cuobject#L119-L124)  so that the 2.10 pytorch with CUDA 12.8 will not missing the cuobject RDMA symbols during runtime. I can definitely remove it, but users would have no easy way to packaging it by themselves. In my opinion, we could wait for the pytorch and vLLM supporting CUDA 13.1 and remove the docker packaging completely.

realAaronWu:
@sammshen @DongDongJu would you mind to review the PR, all comments are resolved, thanks!

PR #467, opened 2025-04-14
Update installation.rst  'ERROR: Could not find a version that satisfies the requirement vllm==0.6.2.3'

princepride:
屏幕截图 2025-04-14 162517

**[Doc] Fix incorrect `vllm` version in installation guide**

This PR updates the installation command in the documentation to use the correct version of the `vllm` package.  
The version `0.6.2.3` does not exist on PyPI and causes an installation failure.  
We change it to `vllm==0.6.2`, which is a valid release and compatible with the project setup.

### Changes
- Update [installation.rst](https://github.com/LMCache/LMCache/blob/dev/docs/source/getting_started/installation.rst) to fix `pip install vllm==0.6.2.3` → `pip install vllm==0.6.2`

### Related Issues
None officially reported yet.

---

<details>
<summary><b> PR Checklist (Click to Expand) </b></summary>

### PR Title and Classification
- [x] **[Doc]** for documentation fixes and improvements

### Code Quality
- [x] No code changes involved
- [x] No linting or formatting issues
- [x] No tests required for doc-only PR

### Review Notes
This is a documentation-only fix. Should be quick to review and merge.

</details>

PR #2879, opened 2026-03-26
Fix garbled output on hybrid models (mamba + attention)

### LINKED ISSUE #2845: Hybrid models like qwen3.5 9B are not supported

malaiwah:
## Summary

Hybrid models like Qwen3.5 interleave attention and mamba/recurrent layers. LMCache correctly filters to attention-only KV caches for storage, but still reports the full token count as cached via `get_num_new_matched_tokens()`. This causes the vLLM scheduler to skip mamba computation for those tokens, leaving recurrent state uninitialized and producing garbled output on cross-request cache hits.

**Root cause**: when LMCache reports N cached tokens, the scheduler sets `request.num_computed_tokens = N` for all layer types. The forward pass skips the first N tokens — correct for attention (KV restored), but mamba layers start from uninitialized state. This produces word salad, repetition loops (`** **`), or empty output ~40% of the time.

## Changes

- **vllm_v1_adapter.py**: Set `_has_non_attention_layers` flag when hybrid model detected; return 0 from `get_num_new_matched_tokens()` when flag is set. Saves remain active (attention KV still cached for future full mamba support).
- **kv_layer_groups.py**: Gracefully skip recurrent/Mamba layers with multi-tensor state instead of crashing with `ValueError`. Logs skipped layers at INFO.
- **gpu_connectors.py**: Filter non-attention tensors in `_lazy_initialize_buffer` for both `VLLMBufferLayerwiseGPUConnector` instances.
- **test_kv_layer_groups_manager.py**: Add test for hybrid model layer group building with Mamba layers skipped.

## Test plan

- [x] Validated on Qwen3.5-4B-FP8 (24 mamba + 8 attention layers, RTX 4080 Super)
- [x] A/B comparison: 0/15 garbled with fix vs 3/20 without (vanilla vLLM baseline: 0/25)
- [x] Unit tests: 10/10 pass (`pytest tests/v1/test_kv_layer_groups_manager.py`)
- [x] No regressions in attention-only model caching (save path unchanged)

Reproduction: send identical streaming requests with tools to a hybrid model. Without fix, ~40% produce garbled reasoning. With fix, output matches vanilla vLLM.

Fixes #2845
Related: #2863 (crash prevention — this PR adds the correctness fix that #2863 is missing)

> AI-assisted: this work was developed with Claude (Anthropic). All changes reviewed and tested by a human.

### DIFF
--- a/lmcache/integration/vllm/lmcache_connector_v1.py
+++ b/lmcache/integration/vllm/lmcache_connector_v1.py
@@ -10,6 +10,20 @@ from vllm.distributed.kv_transfer.kv_connector.v1.base import (
     KVConnectorMetadata,
     KVConnectorRole,
 )
+
+try:
+    # Third Party
+    from vllm.distributed.kv_transfer.kv_connector.v1.base import (
+        SupportsHMA as SupportsHMABase,
+    )
+except ImportError:
+
+    class SupportsHMABase:
+        """Compatibility fallback for vLLM versions without HMA support."""
+
+        pass
+
+
 from vllm.logger import init_logger
 from vllm.v1.core.sched.output import SchedulerOutput
 import torch
@@ -27,7 +41,7 @@ if TYPE_CHECKING:
 logger = init_logger(__name__)
 
 
-class LMCacheConnectorV1Dynamic(KVConnectorBase_V1):
+class LMCacheConnectorV1Dynamic(KVConnectorBase_V1, SupportsHMABase):
     def __init__(
         self,
         vllm_config: "VllmConfig",
@@ -206,3 +220,21 @@ class LMCacheConnectorV1Dynamic(KVConnectorBase_V1):
             returned by the engine.
         """
         return self._lmcache_engine.request_finished(request, block_ids)
+
+    def request_finished_all_groups(
+        self,
+        request: "Request",
+        block_ids: tuple[list[int], ...],
+    ) -> tuple[bool, Optional[dict[str, Any]]]:
+        """
+        Called when an HMA request has finished for all KV cache groups.
+
+        Args:
+            request (Request): the finished request.
+            block_ids (tuple[list[int], ...]): block IDs for each KV cache group.
+
+        Returns:
+            The same async-transfer ownership flag and optional transfer params
+            as request_finished().
+        """
+        return self._lmcache_engine.request_finished_all_groups(request, block_ids)
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1928,6 +1928,31 @@ class LMCacheConnectorV1Impl:
         return False, return_params
 
     @_lmcache_nvtx_annotate
+    def request_finished_all_groups(
+        self,
+        request: "Request",
+        block_ids: tuple[list[int], ...],
+    ) -> tuple[bool, Optional[dict[str, Any]]]:
+        """Handle HMA request completion across all KV cache groups.
+
+        Args:
+            request (Request): the finished request.
+            block_ids (tuple[list[int], ...]): block IDs for each KV cache group.
+
+        Returns:
+            The same async-transfer ownership flag and optional transfer params
+            as request_finished().
+        """
+        if len(block_ids) == 0:
+            return self.request_finished(request, [])
+
+        selected_block_ids, _ = _select_largest_block_group(
+            block_ids,
+            request.request_id,
+        )
+        return self.request_finished(request, selected_block_ids)
+
+    @_lmcache_nvtx_annotate
     def get_kv_events(self) -> Iterable[CacheStoreEvent]:
         if self.lmcache_engine is not None:
             return self.lmcache_engine.get_kv_events()
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Generator, Optional, Union
 import math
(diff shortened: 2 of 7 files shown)


malaiwah (commented):
Thanks for the review notes, both are fair points:

1. **GPU connector reconfigure**: agree, the manual attribute reset is fragile. I'll look at adding a `reconfigure_layers()` method in a follow-up — want to keep this PR focused on the correctness fix.

2. **DRY filter**: the duplicated filter blocks are identical and should be a shared helper. Will extract to `utils.py` alongside the reconfigure method.

Keeping these as follow-up items rather than expanding this PR's scope.

malaiwah:
Addressed both review items in the latest push:

1. **Encapsulation**: Added `reconfigure_for_layers(num_layers)` method to both `VLLMBufferLayerwiseGPUConnector` classes. The adapter now calls `gc.reconfigure_for_layers(n_attn)` instead of manually resetting attributes.

2. **DRY**: Extracted the duplicated hybrid filter into `filter_attention_kv_caches()` in `lmcache/v1/gpu_connector/utils.py`. Both `_lazy_initialize_buffer` methods now call the shared helper.

Tests: 10/10 pass.

malaiwah:
Added commit `2d3ee54` — **fix: detect hybrid models at `__init__` time on both scheduler and worker**

**Root cause of the original bug resurfacing in multiprocess vLLM v1:**

vLLM v1 runs the scheduler (EngineCore) and GPU worker in separate processes. The original fix set `_has_non_attention_layers` in `register_kv_caches()` which only executes in the Worker process (pid for GPU). When `get_num_new_matched_tokens()` runs in the EngineCore/scheduler process, `_has_non_attention_layers` was still `False` — causing the scheduler to report cached tokens even for hybrid models.

**Fix:** Set `_has_non_attention_layers` at `__init__` time from `vllm_config.model_config.is_hybrid` and `is_attention_free`. This runs in BOTH processes at startup, so both see the correct flag from the first request.

**Validation:** Both EngineCore (pid=169) and Worker (pid=232) log on AIBoss:
```
Hybrid/attention-free model detected at init (is_hybrid=True, is_attention_free=False): LMCache load disabled to avoid uninitialized mamba state.
```

malaiwah:
Full cross-node + cross-restart test suite: **10/10 PASS**

The multiprocess init fix (`2d3ee54`) is the key enabler for phases I and G/H — without it, the EngineCore process (running the scheduler) still had `_has_non_attention_layers=False` and would report cached tokens for mamba layers, poisoning the APC cache after a restart-warm-load.

With this PR + llm-d #467 + vllm #38261, both processes log hybrid detection at startup:
```
[Worker    pid=233] Hybrid/attention-free model detected at init (is_hybrid=True, is_attention_free=False): LMCache load disabled to avoid uninitialized mamba state.
[EngineCore pid=170] Hybrid/attention-free model detected at init (is_hybrid=True, is_attention_free=False): LMCache load disabled to avoid uninitialized mamba state.
```

Test setup: Creativity (RTX 4080 Super SM 8.9) ↔ AIBoss (RTX 5090 SM 12.0), model `lovedheart/Qwen3.5-4B-FP8`, same `multiarch` container on both hosts.

| Phase | Description | Result |
|-------|-------------|--------|
| A | Creativity cold start, correct output | ✅ |
| B | APC warm: 8448 cached tokens (attention KV hits LMCache CPU) | ✅ |
| D | AIBoss cold: no garbage (mamba state properly recomputed) | ✅ |
| E | AIBoss APC warm: not poisoned — mamba state was never incorrectly loaded | ✅ |
| G | Creativity restart: warm NFS hit, APC repopulated, correct output | ✅ |
| H | AIBoss restart: warm NFS hit, APC repopulated, correct output | ✅ |
| I | LMCache hybrid detection: **both Worker AND EngineCore** log it correctly | ✅ |
| J | Concurrent cross-host: 97% word overlap, no mamba state interference | ✅ |

Also requesting review from @Jiayi-Pan and @LMCacheBot — this touches the core `vllm_v1_adapter.py` hybrid detection path.

malaiwah:
Friendly ping — all CI checks are green (DCO ✅, k3-correctness ✅, k3-integration ✅, unit-tests ✅, unit-tests-amd ✅) and the PR is mergeable.

The fix has two layers:
1. **Correctness** (`vllm_v1_adapter.py`): return 0 from `get_num_new_matched_tokens()` for hybrid models to prevent mamba state skipping
2. **Multiprocess init** (same file): detect hybrid models at `__init__` time so both EngineCore and Worker processes see the flag from the first request — without this, the scheduler process silently re-enables mamba skipping after container restart

Both layers are needed; without #2 the fix regresses on every vLLM container restart.

@Jiayi-Pan — would appreciate your review on the `vllm_v1_adapter.py` changes in particular.

malaiwah:
## Update: closing this PR as historical

After one more pass over the current vLLM behavior, I think the right conclusion is different from the earlier framing.

Current vLLM already protects the original unsafe condition by failing closed:

- The bundled vLLM `LMCacheConnectorV1` does **not** advertise `SupportsHMA`.
- With HMA enabled, vLLM refuses to start that connector:

```text
ValueError: Connector LMCacheConnectorV1 does not support HMA but HMA is enabled.
Please set `--disable-hybrid-kv-cache-manager`.
```

That means the original corruption path is no longer a live unsafe default on current vLLM. The service will not silently report attention-only LMCache hits as globally reusable hybrid prefix state.

This PR can make the LMCache dynamic connector start under HMA, but the safe behavior is still to return `0` external hits for hybrid/attention-free models. In other words, it prevents corruption, but it does not make LMCache useful for hybrid-model prefix reuse. That is not a strong enough reason to keep this PR open as an HMA support patch.

The cleaner path is:

1. Keep current vLLM's fail-closed behavior for connectors that do not explicitly support HMA.
2. Treat any future `SupportsHMA` declaration as an explicit correctness claim by that connector.
3. Implement real hybrid support separately, including recurrent/Mamba state persistence and restore, before reporting positive external hits.
4. Use this PR and its test history as background for that future work, not as the patch to merge.

Validation from this rebase remains useful historically:

```text
Focused CUDA/vLLM tests: 38 passed
Qwen3.6 correctness probe pass 1: 20 passed, 0 failed
Qwen3.6 correctness probe pass 2: 20 passed, 0 failed
```

But those results validate the conservative guard, not useful hybrid LMCache acceleration.

Closing this PR as historical. The issue it explored is real, but current vLLM already prevents the unsafe default path, and full HMA/Mamba LMCache support should be done as a new design/implementation rather than merging this guard.

malaiwah:
Closing as historical. Current vLLM already fails closed for the bundled LMCacheConnectorV1 under HMA because it does not advertise SupportsHMA, so the original unsafe default path is protected. Full hybrid/Mamba LMCache support should be implemented separately with recurrent state persistence/restore before reporting positive external hits. See the updated comment above for details and latest test results.

PR #2883, opened 2026-03-26
[MP] Improve the stability for controllers and improve log clarity

ApostaC:

Improves MP mode resilience and debuggability:

1. **Main loop protection**: Wraps internal dispatch calls in `_prefetch_loop` (PrefetchController) and `_store_loop` (StoreController) with try-except blocks so that an unexpected exception in one iteration does not crash the background thread.

2. **Stack trace logging**: Switches `logger.error`/`logger.warning` to `logger.exception` at exception catch sites in `multiprocess/` and `distributed/` so that full stack traces are logged instead of just the exception message. Also adds missing logging for the silent `except Exception` in `nixl_store_l2_adapter.py` store path.

The catch-all in the main loops intentionally uses bare `except Exception` (not specific types) because the goal is to prevent *any* internal bug from killing the background thread. The exception is still logged with full traceback via `logger.exception`.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
@@ -696,6 +696,7 @@ class NixlStoreL2Adapter(L2AdapterInterface):
 
         # success is only set to false for transfer failures
         except Exception:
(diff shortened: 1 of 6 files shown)


KuntaiDu (approved):
LGTM!

sammshen (approved):
LGTM!

PR #2885, opened 2026-03-26
[Bench]: Add Blend V2 Stress Test script

sammshen:
Add long_doc_permutator.py and README for stress testing the blend v2 server across 5 axes: context boundaries, eviction, chunk homogeneity, prefix domination, and concurrency.

---

> [!NOTE]
> **Medium Risk**
> Adds a new async benchmark script that can generate high request volume/concurrency against a Blend v2 server and write artifacts; misuse or missing optional dependencies (e.g., LMCache log parser/module) could cause runtime failures or heavy load.
> 
> **Overview**
> Adds a new `benchmarks/blend_v2` permutation-based stress test to exercise Blend v2 KV reuse across *context boundary orderings, eviction pressure, chunk-hash collision risk, prefix-dominated prompts, and concurrency*.
> 
> Introduces `long_doc_permutator.py`, which generates synthetic system prompts/contexts, enumerates or samples context permutations, drives async streaming chat-completion requests with configurable concurrency, and writes results/plots plus a combined `summary.txt` (optionally attempting to parse an `--lmcache-log`). A new README documents the 5 stress axes and provides runnable example configurations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 968a94ea659a8a359dfa4bfd40465ca1dc976538. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


JiwaniZakir (commented):
In `generate_vocab_pool`, the `pool` variable is typed as `set[str]` to avoid duplicates, but the uniqueness is already guaranteed structurally: the suffix `f"{word}{len(pool)}"` uses the current pool size as a counter, so every generated word is inherently unique regardless of the random base. Using a `set` here adds unnecessary overhead for large `--vocab-size` values; a `list` with a simple counter would be both cleaner and more efficient.

The sentinel value `--max-inflight-requests 0` meaning "flood all requests" is a subtle footgun — zero conventionally reads as "no concurrency allowed" rather than "unlimited." A value of `-1` (or a dedicated `--flood` flag) would align better with common CLI conventions and avoid confusion when users scan the argument help text.

One missing stress axis worth considering: the README documents five axes but there's no test scenario combining a very small `--vocab-size` (e.g., 6) with a high `--num-permutations` to simultaneously stress both chunk collision and eviction, which seems like the most adversarial real-world case for the rolling-hash logic.

sammshen:
now in https://github.com/LMCache/LMCache/pull/2937

PR #2888, opened 2026-03-27
[chore]Fix rust Clippy check error on macos

maobaolong:

**If applicable**:

### DIFF
--- a/rust/raw_block/src/lib.rs
+++ b/rust/raw_block/src/lib.rs
@@ -327,7 +327,7 @@ impl RawBlockDevice {
         let ptr_usize = ptr as usize;
         let res = py.allow_threads(move || {
             let src = ptr_usize as *const u8;
-            let src_aligned = (src as usize).is_multiple_of(align);
+            let src_aligned = (src as usize) % align == 0;
             if total_len == payload_len && !self.use_odirect {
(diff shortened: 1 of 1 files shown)


JiwaniZakir (commented):
The fix in `lib.rs` correctly replaces `.is_multiple_of(align)` with `% align == 0` to avoid using an unstable Rust nightly API that wasn't available on the macOS CI toolchain. Both the write path (line 330) and read path (line 470) are updated symmetrically, which is good consistency.

One minor concern: if `align` is ever `0`, the modulo operation will panic with a division-by-zero at runtime, whereas `.is_multiple_of(0)` in the nightly API is defined to return `true` (any number is considered a multiple of zero by that convention). It's worth verifying that `align` is guaranteed non-zero at the call sites—likely it is, since it presumably comes from a block alignment value, but an explicit assertion or comment would make that invariant clear.

Also worth noting: this is a purely mechanical fix for compiler compatibility, but if the codebase targets stable Rust going forward, a `debug_assert!(align > 0)` guard before each modulo would be a cheap safeguard against a silent panic in future refactors.

maobaolong:
@JiwaniZakir Thanks for your review, upgrade my rustc version to 1.94.1 can passed all checks.

PR #2896, opened 2026-03-28
fix: skip vLLM lookup when hash_algorithm is 'builtin'

### LINKED ISSUE #2891: [BUG] LMCache incorrectly queries vLLM with builtin hash algorithm, causing fallback and misleading warning
**Describe the bug**
LMCache uses [`"builtin"` as the default value](https://github.com/LMCache/LMCache/blob/8d8c908fd7b7e1af9dcf858a0173ebae5f1ea574/lmcache/v1/config.py#L97) for `pre_caching_hash_algorithm`, but incorrectly attempts to resolve it via vLLM’s [`get_hash_fn_by_name`](https://github.com/LMCache/LMCache/blob/8d8c908fd7b7e1af9dcf858a0173ebae5f1ea574/lmcache/v1/token_database.py#L97).

Since `"builtin"` is not a valid hash algorithm in vLLM, this lookup fails and LMCache falls back to Python’s built-in `hash()` function, emitting the misleading warning:

```bash
LMCache WARNING: Could not load 'builtin' from vLLM. Using builtin hash.
```

So "builtin" is not a valid vLLM hash algorithm and the warning should not suggest that this kind of algorithm like "builtin" exists in vLLM.

**To Reproduce**
Run `vllm serve` without `LMCACHE_PRE_CACHING_HASH_ALGORITHM` parameter:
```bash
vllm serve $MODEL \
```

**Expected behavior**
- LMCache should NOT attempt to resolve "builtin" via vLLM
- "builtin" should be treated as an internal fallback mode and skip vLLM lookup entirely
If "builtin" is used, LMCache should:
- clearly communicate that this is a local fallback (not a vLLM algorithm)
- provide actionable guidance [similar to the existing PYTHONHASHSEED warning](https://github.com/LMCache/LMCache/blob/8d8c908fd7b7e1af9dcf858a0173ebae5f1ea574/lmcache/v1/token_database.py#L144)

New log suggestion:
>>Falling back to LMCache builtin hash. For production environments (non-testing scenarios), you should set a stable hash algorithm to ensure consistent hashing across processes and nodes.

OS: Linux (Ubuntu)
Environment: Docker / CUDA
vLLM version: 0.18.0
LMCache version: 0.4.2

majiayu000:
Fixes #2891

When `hash_algorithm` is set to `"builtin"`, LMCache was passing that string to vLLM's `get_hash_fn_by_name`, which doesn't recognize it. This caused an unnecessary fallback path and a misleading warning like "Hash function builtin not found in vLLM".

The fix adds an early return in `_get_vllm_hash_func` (`token_database.py`) and `_get_hash_func` (`token_hasher.py`) when the algorithm is `"builtin"`, returning Python's built-in `hash()` directly without ever calling into vLLM. Also logs an INFO message instead of the misleading WARNING, and checks for `PYTHONHASHSEED` in the builtin path.

Both `token_database.py` and `token_hasher.py` have the same logic (the multiprocess version is adapted from the single-process one), so both got the same fix.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/token_hasher.py
+++ b/lmcache/v1/multiprocess/token_hasher.py
@@ -81,12 +81,7 @@ class TokenHasher:
 
         # Short-circuit: 'builtin' means Python's built-in hash(), no vLLM lookup needed
         if hash_algorithm == "builtin":
-            logger.info(
-                "Using LMCache builtin hash (Python hash). "
-                "For production environments (non-testing scenarios), "
-                "consider setting a stable hash algorithm (e.g., sha256_cbor) "
-                "to ensure consistent hashing across processes and nodes."
-            )
+            logger.info("Using LMCache builtin hash (Python hash)")
             if os.getenv("PYTHONHASHSEED") is None:
                 logger.warning(
                     "Using builtin hash without PYTHONHASHSEED set. "
--- a/lmcache/v1/multiprocess/token_hasher.py
+++ b/lmcache/v1/multiprocess/token_hasher.py
@@ -79,6 +79,22 @@ class TokenHasher:
             logger.info("Using blake3 hash function")
             return _make_blake3_hash_func()
 
+        # Short-circuit: 'builtin' means Python's built-in hash(), no vLLM lookup needed
+        if hash_algorithm == "builtin":
+            logger.info(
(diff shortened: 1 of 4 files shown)


rogowski-piotr (approved):
lgtm

Issue #2865, opened 2026-03-25
[MP] LMCache MP mode lack of error handling related method impl


maobaolong:
**Label**
Please label your issue with "new feature" and any other relevant labels so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/1882)

**Is your feature request related to a problem? Please describe.**

If we support sync lookup but async prefetch, there is a window could allocate blocks but didn't read successfully.

**Describe the solution you'd like**
Implement the related apis, refer to https://github.com/LMCache/LMCache/pull/1835 which is contributed by @ziruiliu 

**Describe alternatives you've considered**

**Additional context**

danielhumanmod:
Hey @maobaolong, are this issue open to OSS contribution? If yes I would like to work on this, sounds an interesting problem to solve!

maobaolong:
@danielhumanmod I'm so glad you'd like to take this, thanks for your contribution first. And @yoo-kumaneko and @sammshen @ziruiliu are familiar to this part, they can help to review, you can invite them to review.

maobaolong:
@danielhumanmod It seems #2692 and https://github.com/vllm-project/vllm/pull/36586 are merged, so we already have supported error handling feature for MP mode.

danielhumanmod:
Thanks for the context @maobaolong! Based on my understanding, we need to address a different scenario in this issue than #2692 did.

In #2692, it handles "Server Down" (via `HeartbeatThread`), but in the "Server Healthy" path, it still uses **All-or-Nothing** semantics — if 1 out of 10 chunks is missing, it drops all 10 and only returns `False`. vLLM never receives the specific `error_block_ids`.

But based on my understanding - what we want to handle Partial Failure? — it rescues the 9 good chunks and upgrades exact bad block IDs, so that vLLM can retrieve them via `get_block_ids_with_load_errors`. Just like [#1835](https://github.com/LMCache/LMCache/pull/1835) did for single-process mode.

Happy to know your thought here, and feel free to let me know if my understanding is not correct!

maobaolong:
> Thanks for the context [@maobaolong](https://github.com/maobaolong)! Based on my understanding, we need to address a different scenario in this issue than [#2692](https://github.com/LMCache/LMCache/pull/2692) did.
> 
[...]

PR #2898, opened 2026-03-28
feat: Handle partial retrive failure in LMCache MP mode

danielhumanmod:

Solve #2865 

### Why we need it?
In #1835 , we introduced partial retrieval failure handling for the single-process mode. This PR extends that robustness to the MP flow.

### What this PR does? 
1. **Storage Layer:** Updated to yield both successful and failed chunks (previously only yielded successful ones).
2. **Server Layer:** A new partial retrieval - filters and stores only the "good" blocks to the GPU one by one
3. **Client Layer:** Now captures failed_block_ids and updates the error ID set
4. **Protocol:** Refactored the response format from `tuple[bytes, bool]` to `tuple[bytes, tuple[bool, list[int]]]`

Except new added unit tests, I also update related tests since we change the protocol in this PR.

At the same time, I am working on a e2e flow to verify this change too

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Changes the MP retrieve protocol and server-side GPU copy path to allow partial cache hits and surface failed GPU block IDs; this affects inter-process API compatibility and correctness of GPU writes under partial-miss scenarios.
> 
> **Overview**
> Enables **partial-recovery** for multiprocess `RETRIEVE` so missing chunks no longer fail the entire retrieve; the server now copies only the available chunks to GPU and reports which GPU block IDs could not be loaded.
> 
> This introduces a new storage-manager context manager `read_prefetched_results_partial` (returning `PartialReadResult` with hit/miss indices), updates the MP protocol/handlers to return `tuple[bytes, (success, failed_block_ids)]`, and updates the vLLM MP adapter + tests to consume the new result and track failed blocks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3038ae1e258dee2d3beddd2075e6dc56e255abae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -1125,8 +1125,7 @@ class LMCacheMPWorkerAdapter:
             success, failed_block_ids = r_result
             if failed_block_ids:
                 logger.warning(
-                    "Retrieve for request_id=%s had %d failed block(s) "
-                    "out of %d total",
+                    "Retrieve for request_id=%s had %d failed block(s) out of %d total",
                     request_id,
                     len(failed_block_ids),
                     len(r_block_ids),
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -291,7 +291,8 @@ class LoadStoreOp:
 
 
 StoreResult = bool
-RetrieveResult = bool
+# (success, failed_block_ids): failed_block_ids is empty on full success.
+RetrieveResult = tuple[bool, list[int]]
 LookupResult = int
 
 
@@ -1114,20 +1115,32 @@ class LMCacheMPWorkerAdapter:
                     request_id,
                 )
 
-        for request_id, (r_future, _) in self.retrieve_futures.items():
+        for request_id, (r_future, r_block_ids) in self.retrieve_futures.items():
             if not r_future.query():
                 continue
 
             r_result = r_future.result()
             finished_retrieves.add(request_id)
 
-            if not r_result:
-                logger.error(
-                    "Something went wrong when processing the "
-                    "retrieve request for request_id=%s, result=%s",
+            success, failed_block_ids = r_result
+            if failed_block_ids:
+                logger.warning(
+                    "Retrieve for request_id=%s had %d failed block(s) "
+                    "out of %d total",
                     request_id,
-                    r_result,
+                    len(failed_block_ids),
+                    len(r_block_ids),
                 )
+                self.error_block_ids.update(failed_block_ids)
+            elif not success:
+                # No block-level detail available; mark all blocks as failed.
+                logger.error(
+                    "Retrieve for request_id=%s failed with no block details; "
+                    "marking all %d block(s) as failed",
+                    request_id,
+                    len(r_block_ids),
+                )
+                self.error_block_ids.update(r_block_ids)
 
         # Remove the finished requests from the tracking dicts
         for request_id in finished_stores:
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -1124,22 +1124,21 @@ class LMCacheMPWorkerAdapter:
 
             success, failed_block_ids = r_result
             if failed_block_ids:
-                logger.warning(
-                    "Retrieve for request_id=%s had %d failed block(s) out of %d total",
-                    request_id,
-                    len(failed_block_ids),
-                    len(r_block_ids),
-                )
+                if not success:
+                    logger.error(
+                        "Retrieve for request_id=%s failed entirely; "
+                        "%d block(s) marked as failed",
+                        request_id,
+                        len(failed_block_ids),
+                    )
+                else:
+                    logger.warning(
+                        "Retrieve for request_id=%s: %d/%d block(s) failed",
+                        request_id,
+                        len(failed_block_ids),
+                        len(r_block_ids),
+                    )
                 self.error_block_ids.update(failed_block_ids)
-            elif not success:
-                # No block-level detail available; mark all blocks as failed.
-                logger.error(
-                    "Retrieve for request_id=%s failed with no block details; "
-                    "marking all %d block(s) as failed",
-                    request_id,
-                    len(r_block_ids),
-                )
-                self.error_block_ids.update(r_block_ids)
 
         # Remove the finished requests from the tracking dicts
         for request_id in finished_stores:
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -5,6 +5,7 @@ Distributed multi-tier storage manager for MP mode
 
 # Standard
 from contextlib import contextmanager
+from dataclasses import dataclass
 from typing import Iterator, Literal
 import time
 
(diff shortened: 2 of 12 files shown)


danielhumanmod:
Hey @maobaolong @ziruiliu @sammshen @yoo-kumaneko,

I have the initial version ready for review. I’d appreciate your feedback when you have a chance :) 

From the CI I see UTs/ITs passed, can you help to launch k3-multiprocess-test + k3-comprehensive-test as I dont have. permission, thanks! I’m currently working on an e2e flow to further verify this change at the same time

sammshen:
run the pre-commit please

sammshen:
@cursor review

lmcache/v1/multiprocess/server.py
```diff
@@ -468,35 +454,81 @@ def _retrieve_loop(keys: list[ObjectKey], memory_objs: list[MemoryObj]) -> None:
...
+                    chunk_block_ids_gpu,
+                    gpu_context.device,
+                    lmc_ops.TransferDirection.H2D,
+                    gpu_context.shape_desc,
+                    self.chunk_size,
+                    gpu_context.gpu_kv_format_,
+                    skip_blocks_in_chunk,
+                )
+
+        def _partial_retrieve_loop(
```
sammshen, line 531: can we add a doc string here?

lmcache/v1/distributed/storage_manager.py
```diff
@@ -41,6 +39,20 @@
 logger = init_logger(__name__)
 
 
+@dataclass(frozen=True)
```
sammshen, line 44: can we use `list[MemoryObj | None]` here instead?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -821,20 +799,31 @@ def get_finished(
...
                 continue
 
             r_result = r_future.result()
             finished_retrieves.add(request_id)
 
-            if not r_result:
+            success, failed_block_ids = r_result
+            if failed_block_ids:
+                logger.warning(
+                    "Retrieve for request_id=%s had %d failed block(s) out of %d total",
```
sammshen: nit: Partial retrieve for request ...

sammshen: can we split two cases here as well?

something like: 
```python
  if failed_block_ids:                                                          
      if not success:                                                           
          logger.error("Retrieve for %s failed entirely; %d block(s) marked",
                       request_id, len(failed_block_ids))                       
      else:                                                                     
          logger.warning("Retrieve for %s: %d/%d block(s) failed",
                         request_id, len(failed_block_ids), len(r_block_ids)) 
```

danielhumanmod: Thanks for all the feedback, will fix today or tomorrow

lmcache/v1/distributed/storage_manager.py
danielhumanmod: Thanks for the feedback! I was wondering if we could stick with `ist[MemoryObj]` here. Looking at the logic in L291-304, we ensure that only non-None items are added to `good_objs`. Keeping the current type would help us avoid adding extra null checks in downstream code. But happy to know your further thought here!
```
  for idx, k in enumerate(keys):
            e, o = read_results[k]
            if o is None:
                logger.error(
                    "Failed to read prefetched object %s from L1 storage: %s",
                    k,
                    strerror(e),
                )
                bad_keys.append(k)
                bad_indices.append(idx)
            else:
                good_keys.append(k)
                good_objs.append(o)
                good_indices.append(idx)
```

lmcache/v1/multiprocess/server.py
danielhumanmod: Added, thanks! 

Additionally, regarding a few critical sections like `retrieve_loop` and `read_prefetched_results`, I’ve implemented a new version specifically for partial fetching rather than modifying the existing logic. My main reason was to avoid new change cause potential regressions in the current flow as it's my first time contribute here.

I’d love to hear your thoughts on this—would you prefer this decoupled approach, or would you rather have them integrated in-place?

yoo-kumaneko:
Hey, one thing on the docs side before digging into the code:

This PR adds a new public context manager (`read_prefetched_results_partial`) and a new public type (`PartialReadResult`), but I don't see any doc updates to go with them. Per the coding standards (§1.2 and §5.1), new public APIs should come with doc updates on both surfaces:

- **Design doc** — there's no doc under `docs/design/v1/distributed/` covering the storage manager's read path today, and nothing in this PR adds one for the new partial API or the `PartialReadResult` contract (good/bad indices, lock-release semantics on normal vs. exception exit, caller's responsibility to call `finish_read_prefetched` with the good keys).
- **User-facing doc** — `docs/source/mp/architecture.rst` (around line 169) still lists only `read_prefetched_results()` under `StorageManager`. Worth mentioning the partial variant there too.
- **Server-side partial retrieve** — the MP `retrieve()` protocol change (`tuple[bytes, bool]` → `tuple[bytes, tuple[bool, list[int]]]`) and the new `_partial_retrieve_loop` behavior would be nice to capture under `docs/design/v1/multiprocess/` so the next person touching this code understands why there are two loops and what the `failed_block_ids` contract is.

Could you add the design doc + update the architecture.rst entry in this PR? Happy to review a draft.

yoo-kumaneko:
@danielhumanmod 
Also, it seems the branch has conflicts with the dev branch. Would you like to fix them?

danielhumanmod:
> @danielhumanmod Also, it seems the branch has conflicts with the dev branch. Would you like to fix them?

Thanks for the feedback, will fix this and add doc today or tomorrow!

yoo-kumaneko:
Overall, it looks good to me. The logic is correct.

yoo-kumaneko:
@danielhumanmod Would you like to join the LMCache WeChat group?
Clipboard_Screenshot_1776927530

danielhumanmod:
> @danielhumanmod Would you like to join the LMCache WeChat group?
> 
> Clipboard_Screenshot_1776927530
> 
> 

Yes very happy to join, thanks for the invite!

yoo-kumaneko:
Did a quick skim and it looks good to me. Code is clean.

maobaolong:
@danielhumanmod Would you like to resolve the conflict?

lazariv:
we're hitting #3388 in production and this PR is the fix we need!

**Setup:** vLLM 0.25.0 + LMCache 0.5.1, `LMCacheMPConnector` (`kv_role=kv_both`, `ptr`/lmcache_driven), serving DSA models (GLM-5.2-FP8, MiniMax-M3) on H100 / TP16.

**Why this is high severity:** under heavy-overflow load (working set > GPU KV pool) we saw `result=False` from the MP retrieve on a large fraction of requests — **~57% in our stress runs**. On that path, `get_finished` in `vllm_multi_process_adapter.py` currently only *logs* the failure and marks the request finished; the failed blocks are never added to `error_block_ids`, so `get_block_ids_with_load_errors()` returns empty and vLLM proceeds to run attention over **uninitialized KV**. The result is **silently corrupted output returned as HTTP 200** — no error surfaces to the caller, which makes it a particularly dangerous failure mode in production (we only caught it via a downstream correctness check, not from any error signal).

This PR fixes exactly that by propagating the failed `block_ids` so vLLM recomputes them. Would really like to see it land — happy to help test against our MP + DSA deployment under load. Thanks @danielhumanmod! 🙏

PR #2904, opened 2026-03-30
[Feat][DAX] Optimize staged batched restore path and document modification

DongDongJu:
This PR optimizes the DAX backend's staged restore path.
  Right next plan is support Higher TP/DP/EP and support MP mode like raw block backend.

Before this change, DAX retrieval restored cached chunks through serialized per-chunk restaging, which became the main bottleneck on long-context cache-hit workloads. This PR keeps the existing staged retrieve model (`DAX -> CPU -> normal GPU connector`) but makes the DAX restore path batched and reusable.

  Main changes:
  - add a shared batched restore pipeline for blocking and async-prefetch DAX retrieval
  - group restore output allocation from `LocalCPUBackend`
  - coalesce adjacent DAX spans and execute restore work in regions/waves
  - add a reusable pinned retrieve staging slab and persistent restore executors
  - add a native `batched_memcpy` helper with non-CUDA fallback
  - document the new DAX restore tuning knobs

  In local `long_doc_qa` validation on `Qwen/Qwen3-14B`, the optimized staged DAX path reduced mean query TTFT by about
  `61.6%` and mean query-round time by about `33.3%` versus the pre-optimization DAX backend.

  - Scope is intentionally limited to the DAX staged retrieve path and the small native copy helper it depends on.
  - This PR does not change the basic store path; stores are still CPU-staged before being written to DAX.
  - Sync and async-prefetch retrieval semantics are preserved; the backend only replaces serialized per-chunk restore
  with batched restore.
  - New DAX knobs documented in this PR:
    - `dax.restore_workers`
    - `dax.restore_max_regions`
    - `dax.retrieve_staging_slab_bytes`
  - Local validation beyond unit tests:
    -`long_doc_qa` comparison vs pre-optimization DAX: TTFT `-61.6%`, query-round `-33.3%`
    - reversed-order control (`baseline -> optimized`): TTFT `-58.4%`, query-round `-32.5%`
    - longer-running control with query round > 1 minute: TTFT `-62.0%`, query-round `-36.0%`

```
sample test configs

  chunk_size: 256
  local_cpu: false
  max_local_cpu_size: 16
  cache_policy: LRU
  min_retrieve_tokens: 0
  storage_plugins: ["dax"]
  store_location: dax
  retrieve_locations: ["dax"]
  extra_config:
    storage_plugin.dax.module_path: lmcache.v1.storage_backend.plugins.dax_backend
    storage_plugin.dax.class_name: DaxBackend
    dax.device_path: "/dev/daxxxx"
    dax.max_dax_size: 64
    dax.async_put: false

  vllm serve Qwen/Qwen3-14B \
    --host 0.0.0.0 \
    --port 8000 \
    --reasoning-parser qwen3 \
    --attention-backend FLASH_ATTN \
    --gpu-memory-utilization 0.5 \
    --max-model-len 32768 \
    --no-enable-prefix-caching \

  python benchmarks/long_doc_qa/long_doc_qa.py \
    --host 127.0.0.1 \
    --port 8000 \
    --model Qwen/Qwen3-14B \
    --document-length 10000 \
    --num-documents 30 \
    --output-len 200 \
    --repeat-count 2 \
    --repeat-mode tile \
    --max-inflight-requests 5 \
    --sleep-time-after-warmup 0 \
    --trim-fraction 0.0 \
    --json-output
```

  **If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Reworks the DAX retrieval path to use new threaded batched restore logic and a shared pinned staging slab, which can affect concurrency, lifecycle/cleanup, and memory-safety if mis-tuned or if edge cases slip through.
> 
> **Overview**
> **DAX retrieval is refactored from per-key restores into a staged *batched restore pipeline*.** Both blocking and async-prefetch reads now go through a shared dispatcher that reserves readable entries, batches CPU output allocation, coalesces adjacent DAX spans into region/wave copy plans, and restores via persistent thread pools while preserving existing semantics (*blocking keeps positional `None` holes; async returns only the consecutive hit prefix*).
> 
> Adds new DAX tuning knobs (`dax.restore_workers`, `dax.restore_max_regions`, `dax.retrieve_staging_slab_bytes`) and manages a backend-owned pinned staging slab with explicit shutdown/cleanup on `close()` and init failures.
> 
> Introduces a native `batched_memcpy` (`csrc/mem_alloc.*` + `pybind`) with a Python non-CUDA fallback, and expands DAX unit tests to cover batched restore behavior, locking/close interactions, and the memcpy helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c5c53af74ab012a9a5ad5bd04b832d83138fd6e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -502,13 +616,29 @@ async def batched_get_non_blocking(
...
+        )
+        return cast(list[MemoryObj], result)
+
+    def batched_get_blocking(
+        self,
+        keys: List[CacheEngineKey],
+    ) -> List[Optional[MemoryObj]]:
+        if not keys:
+            return []
+        return cast(List[Optional[MemoryObj]], self._restore_batch(list(keys), False))
```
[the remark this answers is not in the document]
DongDongJu: resolved in 141f7631be9495ff59c215d30b8b71b05db5d405.

csrc/mem_alloc.cpp
```diff
@@ -26,6 +26,23 @@ void free_pinned_ptr(uintptr_t ptr) {
   }
 }
 
+void batched_memcpy(const std::vector<uintptr_t>& src_ptrs,
```
sammshen, line 29: is the main benefit just avoiding GIL?

sammshen (approved):
LGTM!

jayhpark530:
Nice optimization! Looking forward to MP mode support as well.
How does `local_cpu: false` work in the sample test config?

csrc/mem_alloc.cpp
DongDongJu: Yes one of reason is that GIL avoidance. That helps, but it’s only one part.
The bigger gain here is that the restore path is now batched end-to-end.
`batched_memcpy` mainly removes Python per-copy overhead on that path and lets the copy loop run in native code with the GIL released, but by itself it would not explain all of the TTFT improvement.

DongDongJu:
> Nice optimization! Looking forward to MP mode support as well. How does `local_cpu: false` work in the sample test config?

The DAX backend still needs LocalCPUBackend as its allocator backend for restore buffers/staging outputs. 

So in this config, max_local_cpu_size: 16 is still providing the CPU-side allocation budget used by DAX restore, but retrieved chunks are not being admitted into the local CPU hot cache.

So the sample path is effectively:

DAX -> CPU restore buffers from LocalCPUBackend -> normal GPU connector

rather than

DAX -> keep result in local CPU hot cache.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -769,6 +981,301 @@ def _evict_one_locked(self) -> bool:
...
+                touched_keys.add(item.key)
+        except Exception:
+            self._cleanup_restore_outputs(reserved)
+            self._finalize_reserved_items(reserved)
+            raise
+
+        self._finalize_reserved_items(reserved, touched_keys=touched_keys)
+        if prefix_only:
+            return [cast(MemoryObj, results[item.result_index]) for item in reserved]
+        return results
```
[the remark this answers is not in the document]
DongDongJu: Yes this is exactly what I need to targeting for next one to support higher parallelism.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -595,12 +753,66 @@ def close(self) -> None:
...
+        if ptr:
+            try:
+                lmc_ops.free_pinned_ptr(ptr)
+            except Exception as e:
+                logger.warning("Failed to free DAX retrieve slab: %s", e)
+
+        if restore_slab_ptr is None:
+            self._retrieve_staging_slab_ptr = 0
+            self._retrieve_staging_slab_bytes = 0
+            self._restore_region_bytes = 0
```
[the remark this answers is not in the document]
DongDongJu: Resolved in f95213e

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -769,6 +977,301 @@ def _evict_one_locked(self) -> bool:
...
+                    state.pending_free = False
+                    self._free_slot_locked(item.slot_id)
+
+            self._state_condition.notify_all()
+
+    def _restore_batch(
+        self,
+        keys: list[CacheEngineKey],
+        prefix_only: bool,
+    ) -> list[Optional[MemoryObj]] | list[MemoryObj]:
```
ApostaC: dumb question: why we have `list[Optional[MemoryObj]] | list[MemoryObj]`? I think it's effectively the same as `list[Optional[MemoryObj]]`.

lmcache/v1/storage_backend/plugins/dax_backend.py
```diff
@@ -595,12 +751,64 @@ def close(self) -> None:
...
+        except (TypeError, ValueError) as e:
+            raise ValueError(f"extra_config['{key}'] must be a positive integer") from e
+        if parsed <= 0:
+            raise ValueError(f"extra_config['{key}'] must be a positive integer")
+        return parsed
+
+    def _release_restore_resources(
+        self,
+        restore_slab_ptr: Optional[int] = None,
+    ) -> None:
```
ApostaC, line 802: nit: regarding those newly introduced long helper functions, it would be better if we can add some docstring about describing the args, expected return values and expected behavior. It will help other maintainers (and AI agents) better understand the codes.

lmcache/v1/storage_backend/plugins/dax_backend.py
ApostaC: It's a bit weird to have `list[Optional[MemoryObj]] | list[MemoryObj]` here. Isn't it the same as `list[Optional[MemoryObj]]`?

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/storage_backend/plugins/dax_backend.py
DongDongJu: You are right. thanks for catching it. it removed.

DongDongJu: You are right. thanks for catching it. it removed.

lmcache/v1/storage_backend/plugins/dax_backend.py
DongDongJu: addressed here https://github.com/LMCache/LMCache/pull/2904/commits/92b7fa67c9b03154756900d749e40f83e5928dc1

ApostaC (approved):
LGTM!

PR #2915, opened 2026-03-31
[Feature] A two-stage design that aims to reduce scheduling-induced latency

yoo-kumaneko:
Add SYNC_LOOKUP request type that returns hit count in a single blocking round-trip by composing existing lookup() + query_prefetch_lookup_hits() internally. L2-to-L1 data transfer runs in the existing background prefetch loop — no new synchronous code paths in PrefetchController or StorageManager.

RETRIEVE coordination via extra_count on _PrefetchJob:
- query_prefetch_status() caches result on first successful query; subsequent TP workers read the cached value
- remaining_consumers countdown (set to tp_size by sync_lookup) controls when the _PrefetchJob entry is cleaned up
- _PendingLookupState(job_id) saved by sync_lookup so retrieve() knows which job to poll

Also adds LMCacheMPSyncLookUPSchedulerAdapter that sends SYNC_LOOKUP instead of the two-phase LOOKUP + QUERY_PREFETCH_STATUS.

---

> [!NOTE]
> **Medium Risk**
> Adds a new blocking RPC and dedicated server thread pool, plus new background-threaded worker behavior; concurrency and cleanup paths could introduce deadlocks, starvation, or leaked prefetch jobs if incorrect.
> 
> **Overview**
> Introduces a new `SYNC_LOOKUP` request type that combines `LOOKUP` + lookup-hit polling into a single blocking round-trip, returning the hit chunk count immediately.
> 
> Updates the multiprocess server to implement `MPCacheEngine.sync_lookup`, register it as a handler, run it in a **dedicated normal thread pool** (new `max_sync_lookup_workers` config/CLI knob), and adds extra `end_session` cleanup to drain any lingering prefetch jobs.
> 
> Adds vLLM integration adapters: `LMCacheMPSyncLookupSchedulerAdapter` (single `SYNC_LOOKUP` call with cached results) and `LMCacheMPSyncLookupWorkerAdapter` (queues `RETRIEVE` and uses a background poll of `QUERY_PREFETCH_STATUS` to avoid blocking GPU-affinity threads). Comprehensive tests cover protocol wiring, server/job lifecycle, thread-pool assignment, and adapter behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1682a656dfd7d48ef9795a5a887a8eb8961f121d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -551,6 +551,96 @@ class LMCacheMPSchedulerAdapter:
         )
 
 
+class LMCacheMPSyncLookupSchedulerAdapter(LMCacheMPSchedulerAdapter):
+    """Scheduler adapter that resolves lookups via a single SYNC_LOOKUP RPC.
+
+    The default :class:`LMCacheMPSchedulerAdapter` performs a two-phase
+    lookup (LOOKUP then repeated QUERY_PREFETCH_STATUS polls from the
+    scheduler loop).  This subclass replaces that with a single blocking
+    SYNC_LOOKUP request: the server completes the L2 lookup phase before
+    returning, and the scheduler learns the hit count immediately.
+
+    The prefetch job remains alive on the server so that the matching
+    :class:`LMCacheMPSyncLookupWorkerAdapter` can poll
+    QUERY_PREFETCH_STATUS to gate the eventual RETRIEVE.
+    """
+
+    @_lmcache_nvtx_annotate
+    def maybe_submit_lookup_request(
+        self,
+        request_id: str,
+        token_ids: list[int],
+        cache_salt: str = "",
+    ) -> None:
+        """Submit a SYNC_LOOKUP request if there is no ongoing lookup.
+
+        Blocks until the server returns the hit chunk count and caches
+        the result for :meth:`check_lookup_result`.
+
+        Args:
+            request_id: The ID of the lookup request.
+            token_ids: Token IDs to look up in LMCache.
+            cache_salt: Per-user isolation salt. Requests with different
+                cache_salt values produce separate cache entries.
+        """
+        self._ensure_heartbeat_started()
+
+        if not self.is_healthy:
+            return
+
+        if (
+            request_id in self._pending_lookups
+            or request_id in self._finished_lookup_results
+        ):
+            return
+
+        aligned_end = (len(token_ids) // self.chunk_size) * self.chunk_size
+
+        key = self._create_key(
+            token_ids,
+            start=0,
+            end=aligned_end,
+            request_id=request_id,
+            cache_salt=cache_salt,
+        ).no_worker_id_version()
+
+        self._pending_lookups.add(request_id)
+        try:
+            hit_chunks = send_lmcache_request(
+                self.mq_client,
+                RequestType.SYNC_LOOKUP,
+                [key, self.tp_size],
+            ).result(timeout=self._mq_timeout)
+        except TimeoutError:
+            logger.warning(
+                "SYNC_LOOKUP request timed out after %ss. Marking server as unhealthy.",
+                self._mq_timeout,
+            )
+            self._health_event.clear()
+            self._pending_lookups.discard(request_id)
+            return
+
+        self._finished_lookup_results[request_id] = hit_chunks * self.chunk_size
+
+    @_lmcache_nvtx_annotate
+    def check_lookup_result(self, request_id: str) -> int | None:
+        """Return the cached SYNC_LOOKUP result for ``request_id``.
+
+        SYNC_LOOKUP completes the lookup phase synchronously, so the
+        result is always available the moment
+        :meth:`maybe_submit_lookup_request` returns.  This method simply
+        reads the cached token count and never returns ``None``.
+
+        Args:
+            request_id: The ID of the lookup request.
+
+        Returns:
+            Number of tokens matched in LMCache (prefix matching). Zero
+            when the server was unhealthy or the lookup timed out.
+        """
+        return self._finished_lookup_results.get(request_id, 0)
+
+
 class LMCacheMPWorkerAdapter:
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


sammshen:
@cursor review

sammshen:
@cursor review

lmcache/v1/multiprocess/server.py
```diff
@@ -416,6 +448,23 @@ def retrieve(
...
+        # background prefetch loop to finish L2→L1 data transfer.
+        with self._pending_lookups_lock:
+            pending = self._pending_lookups.get(key.request_id)
+
+        if pending is not None:
+            # Poll until the prefetch completes. The first worker
+            # triggers the real query; others get the cached result
+            # via the extra_count mechanism in query_prefetch_status.
+            while self.query_prefetch_status(pending.job_id) is None:
+                time.sleep(0.001)  # 1 ms
```
ApostaC: We should not do busy sleeping in `retrieve`, because it will block the gpu worker and slow down all the other GPU-related operations (like `store`, or other `retrieve`s). The key thing here is we need to make sure `retrieve` is non-blocking.

Instead, we need to have a separate "busy sleeping" thread in vLLM worker process looping over each prefetch request to check whether we can call retrieve:
1. The prefetch has finished (by calling `query_prefetch_status`)
2. GPU IDs have been seen by the vLLM worker process.

lmcache/v1/multiprocess/server.py
```diff
@@ -696,6 +749,14 @@ def query_prefetch_status(
             )
             return None
 
+        # Fast path: result already cached by an earlier consumer
+        if job.cached_result is not None:
+            with self._prefetch_job_lock:
+                job.remaining_consumers -= 1
+                if job.remaining_consumers <= 0:
+                    self._prefetch_jobs.pop(prefetch_job_id, None)
+            return job.cached_result
```
ApostaC: Agreed with gemini, we need to have the lock when we are going to modify the job status

lmcache/v1/multiprocess/server.py
```diff
@@ -706,6 +767,9 @@ def query_prefetch_status(
         #    first failure
         found_count = found_count // job.world_size
 
+        # Cache the result for other consumers
+        job.cached_result = found_count
```
ApostaC: Same thread-safety issue here

lmcache/v1/multiprocess/server.py
```diff
@@ -715,7 +779,80 @@ def query_prefetch_status(
...
+
+        Returns:
+            Number of matched chunks (prefix hit count).
+        """
+        self._event_bus.publish(
+            Event(
+                event_type=EventType.MP_LOOKUP_PREFETCH_START,
+                session_id=key.request_id,
+            )
+        )
```
ApostaC: We don't need to emit this event if we are going to call `self.lookup`

lmcache/v1/multiprocess/server.py
```diff
@@ -715,7 +779,80 @@ def query_prefetch_status(
...
+
+        # Step 1: submit to the existing background prefetch loop
+        job_id = self.lookup(key, tp_size)
+
+        # Step 2: poll until the lookup phase finishes
+        while True:
+            found_count = self.query_prefetch_lookup_hits(job_id)
+            if found_count is not None:
+                break
+            time.sleep(0.001)  # 1 ms
```
ApostaC: Let's probably have a 5ms sleep interval here? 1ms may incur quite amount of overhead here.

ApostaC (changes requested):
Key thing: let's not have the while loop in `retrieve` (therefore, `_PendingLookupState` is not needed). Instead, we should consider having a while loop thread in the vLLM worker process to track the prefetch status for each request.

lmcache/v1/multiprocess/server.py
maobaolong: @ApostaC `N`ms means the worst case we have to wait extra more `N`ms even the lookup completed, but indeed, 1ms may make `cpu` busy to loop.
So actually, I've discussed with @yoo-kumaneko there are some further improve ideas, e.g.
- Callback or threading.Condition notify the caller immediately when lookup completed.
- Increase sleep interval from init value to maximum value.

Do you have any other thoughts?

lmcache/v1/multiprocess/server.py
yoo-kumaneko: I see. There's an obstacle to querying the prefetch status on the worker side. The workers don't have the prefetch job ID. How do we tackle this problem?

yoo-kumaneko: Why do we use the job ID in the first place? Can we reuse the request ID, which is broadly used in the vLLM?

yoo-kumaneko:
@ApostaC I’ve refactored the design. The synchronization issues around the query_prefetch_status handler stem from the fact that it actually _pops_ the query status rather than simply _querying_ it.

I added a pop parameter to the query method in both the storage manager and the prefetch controller, with pop=true as the default behavior. I also introduced a new handler that truly _queries_ the prefetch status (i.e., no popping). Cleanup is now deferred to the end_session handler, which pops the job from the prefetched jobs dictionary.

I tried to reuse the existing query_prefetch_status method, but couldn’t make it work. The method takes job_id as input rather than request_id. The workers don't have it. I’m curious why we chose job_id over request_id (which is a more universal option IMO) in the first place.

lmcache/v1/multiprocess/server.py
yoo-kumaneko: @ApostaC done. I moved the sleeping loop to the worker adapter, in its own background thread. I also did a refactor on the prefetch status query part. Please also see my newest comment down below.

lmcache/v1/multiprocess/server.py
yoo-kumaneko: removed

lmcache/v1/multiprocess/server.py
```diff
@@ -714,8 +735,101 @@ def query_prefetch_status(
...
+                self._pending_lookups[key.request_id] = _PendingLookupState(
+                    job_id=job_id,
+                )
+        else:
+            # No hits — no RETRIEVE will be issued. Clean up the
+            # prefetch job now; end_session will handle the pop of
+            # the prefetch controller result.
+            with self._prefetch_job_lock:
+                self._prefetch_jobs.pop(job_id, None)
+                self._request_to_job_id.pop(key.request_id, None)
```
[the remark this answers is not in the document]
yoo-kumaneko: Good catch. Fixed in c7d935e0 — the zero-hits branch now calls `query_prefetch_status(job.handle, pop=True)` directly before discarding the job reference, so the prefetch controller's `_completed_results` and `_completed_lookups` entries are properly cleaned up.

lmcache/v1/multiprocess/server.py
```diff
@@ -774,13 +888,25 @@ def get_chunk_size(self) -> int:
...
         """
         self.session_manager.remove(request_id)
 
+        # Clean up prefetch job (pop=True to consume the result from
+        # the prefetch controller, preventing memory leaks).
+        with self._prefetch_job_lock:
+            job_id = self._request_to_job_id.pop(request_id, None)
+            job = self._prefetch_jobs.pop(job_id, None) if job_id is not None else None
+        if job is not None and job.handle.prefetch_request_id != -1:
+            self.storage_manager.query_prefetch_status(job.handle, pop=True)
```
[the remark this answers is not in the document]
yoo-kumaneko: That's a real problem, but this PR does not bring it. It should be fixed in another PR.

lmcache/v1/multiprocess/server.py
```diff
@@ -147,6 +147,20 @@ class _PrefetchJob:
...
+    """State saved between SYNC_LOOKUP and RETRIEVE.
+
+    SYNC_LOOKUP stores the ``job_id`` here so that the worker adapter
+    can poll ``PREFETCH_READY(request_id)`` to wait for L2-to-L1 data
+    transfer to complete before sending RETRIEVE.
+    """
+
+    job_id: int
+    """Prefetch job ID returned by ``lookup()``."""
+    """Countdown of RETRIEVE workers; last one removes the entry."""
```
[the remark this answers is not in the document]
yoo-kumaneko: fixed

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -469,6 +469,95 @@ def _create_key(
...
+
+    The SYNC_LOOKUP RPC performs L1 prefix scan and L2 existence check
+    in a single blocking round-trip, returning the hit count directly.
+    L2-to-L1 data movement is deferred to the RETRIEVE call, overlapping
+    it with the forward pass of other scheduled requests.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._lookup_hit_counts: dict[str, int] = {}
```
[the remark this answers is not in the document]
yoo-kumaneko: doesn't really matter. ignored

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -27,8 +27,10 @@
     "STORE",
     "RETRIEVE",
     "LOOKUP",
+    "SYNC_LOOKUP",
     "QUERY_PREFETCH_STATUS",
     "QUERY_PREFETCH_LOOKUP_HITS",
```
ApostaC, line 32: In the new implementation, since we are using SYNC_LOOKUP and vLLM will never call `QUERY_PREFETCH_LOOKUP_HITS`, we can remove `QUERY_PREFETCH_LOOKUP_HITS` from the protocol.

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -27,8 +27,10 @@
...
+    "QUERY_PREFETCH_STATUS_WITH_REQ_ID",
```
ApostaC, line 33: I agree with you on removing the `job ID` and reuse request ID. I think that will make the code cleaner. Therefore, we can modify and reuse `QUERY_PREFETCH_STATUS` without having a new `QUERY_PREFETCH_STATUS_WITH_REQ_ID`. WDYT?

lmcache/v1/multiprocess/server.py
```diff
@@ -714,8 +735,105 @@ def query_prefetch_status(
...
+
+        # Step 1: submit to the existing background prefetch loop
+        job_id = self.lookup(key, tp_size)
+
+        # Step 2: poll until the lookup phase finishes
+        while True:
+            found_count = self.query_prefetch_lookup_hits(job_id)
+            if found_count is not None:
+                break
+            time.sleep(0.005)  # 5 ms
```
ApostaC, line 810: Here we probably need to update the docs somewhere -- saying if we are using sync lookup, we need to make sure `--cpu-workers` is larger than the number of scheduler instances running on the node. Otherwise, it may block other functionalities like heartbeat

lmcache/v1/multiprocess/server.py
```diff
@@ -714,8 +735,105 @@ def query_prefetch_status(
...
+            request_id: The external request ID.
+
+        Returns:
+            Chunk count (int) when done, ``None`` if still in progress
+            or no SYNC_LOOKUP was issued for this request.
+        """
+        with self._pending_lookups_lock:
+            pending = self._pending_lookups.get(request_id)
+        if pending is None:
+            return None
```
ApostaC, line 761: Why do we need to check `_pending_lookups` here? Conceptually, whether we call `sync_lookup` should not impact the behavior of `query_prefetch_status`.

ApostaC (commented):
@yoo-kumaneko I like your proposal of using the request ID in the prefetch-related operations. I believe that will make the code cleaner. Please take a look at my other comments/questions. Thanks!

yoo-kumaneko:
Addressed the two unresolved Cursor Bugbot comments in `1682a656`:

**1. `SYNC_LOOKUP blocking loop starves shared thread pool` (Medium)**

`SYNC_LOOKUP` now gets its own dedicated `ThreadPoolExecutor` via a second `add_normal_thread_pool` call in `run_cache_server`, keyed off a new `MPServerConfig.max_sync_lookup_workers` knob (default 16, `--max-sync-lookup-workers` CLI arg). The shared normal pool no longer hosts any blocking-poll handlers, so heartbeat `PING`, `END_SESSION`, `QUERY_PREFETCH_STATUS`, etc. can't be starved by concurrent `sync_lookup` calls.

`test_sync_lookup_uses_dedicated_thread_pool` asserts that `SYNC_LOOKUP` is on its own pool (and not with `LOOKUP`/`PING`/`END_SESSION`/`QUERY_PREFETCH_STATUS`) and that the pool size tracks the config knob.

**2. `New feature classes lack any test coverage` (Medium)**

Added `tests/v1/multiprocess/test_sync_lookup.py` with 23 tests covering:

- Protocol registration (enum, payload, response, handler type) — 4 tests
- Message-queue round-trip with a stub handler — 1 test
- Server `sync_lookup`: non-zero hits keep the job, zero hits drain it, `None` polls wait, missing job returns 0, handler registered — 5 tests
- Dedicated thread pool wiring — 1 test
- `end_session` fallback drain of a lingering `_prefetch_jobs` entry — 2 tests
- Scheduler adapter: single `SYNC_LOOKUP` RPC, cached `check_lookup_result`, unknown-id → 0, unhealthy short-circuit, timeout clears pending + marks unhealthy — 5 tests
- Worker adapter: queues retrieve without RPC, unhealthy path routes block_ids to errors, background loop dispatches RETRIEVE after `QUERY_PREFETCH_STATUS` goes non-`None`, unhealthy `get_finished` drains queued ops, `error_block_ids` lock demonstrably held across copy+clear — 5 tests

All 23 new tests pass; no regressions in the rest of `tests/v1/multiprocess/` (70 pre-existing MP tests still pass alongside).

PR #2926, opened 2026-04-01
[MP]feat: support different kv cache shape and dtype across layers

liuyumoye:
support different kv cache shape and dtype across layers

This PR adds support for heterogeneous KV cache shapes and dtypes across layers (e.g., models where different layers have different KV head dimensions or data types).

Previously, GPUCacheContext assumed all layers share the same shape and dtype. This PR introduces KVLayerGroupsManager to group layers by (shape, dtype), and updates the D2H/H2D transfer logic in server.py to iterate over each group independently, using per-group tmp_gpu_buffer, kv_pointers, and tensor views.

Key changes:

`kv_layer_groups.py`: Add `build_kv_layer_groups_from_list()` to build layer groups from a raw list of KV cache tensors (no layer names required), grouping by (shape, dtype).

`gpu_context.py`: Replace single hidden_dim_size_ / tmp_gpu_buffer_ with per-group lists (hidden_dim_sizes_, tmp_gpu_buffers_); expose get_tmp_gpu_buffer(num_tokens, group_idx), get_kv_buffer_shape(num_tokens, group_idx), and kv_layer_groups_manager property.

`server.py`: Update get_layout_desc() to produce per-group shapes/dtypes; refactor D2H (store) and H2D (retrieve) loops to iterate over all groups.

`memory_management.py`: Improve error handling in tensor property and get_tensor() — replace bare assert with descriptive ValueError; make tensor fall back to get_tensor(0) in multi-group scenarios.

`mock_l2_adapter.py`: Replace bare assert with ValueError for cleaner error messages.

The grouping logic in `build_kv_layer_groups_from_list()` is order-preserving: groups are sorted by the first layer index they contain.

get_tmp_gpu_buffer and get_kv_buffer_shape are backward-compatible — group_idx defaults to 0, so single-group models are unaffected.

The tensor property on TensorMemoryObj now delegates to get_tensor(0) when per-group metadata is present, maintaining backward compatibility with callers that only use the single-tensor interface.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Updates CUDA transfer kernels and the MP server store/retrieve path to handle per-layer-group shapes/dtypes, which can affect correctness and performance of GPU<->CPU KV transfers. Added tests mitigate risk but changes touch core memory copy and layout logic.
> 
> **Overview**
> Enables multiprocessing cache store/retrieve to support **heterogeneous KV cache shapes and dtypes across layers** by grouping layers with identical `(shape, dtype)` and transferring each group independently.
> 
> `GPUCacheContext` now builds `KVLayerGroupsManager`, maintains per-group `PageBufferShapeDesc` and pointer arrays, and replaces the old single temporary KV buffer with a **flat `uint8` chunk buffer** that concatenates all groups (with helpers to view per-group/per-batch slices). `server.py`’s layout description, store (D2H), and retrieve (H2D) paths are updated to iterate over groups and copy via the new flat buffer.
> 
> CUDA `multi_layer_block_kv_transfer` is generalized to dispatch over vector widths (`uint4`/`uint32_t`/`uint16_t`) based on alignment instead of requiring `uint4`, and Python GPU memcpy helpers now validate sizes and copy raw bytes for non-lazy allocators. New unit tests cover multi-group temp-buffer layout and non-4-byte-aligned `lmcache_memcpy_async` copies (e.g. `int8` hidden size 132).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89ece3ce74bbc4bcedd026f51e039b95786777a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/csrc/mp_mem_kernels.cu
+++ b/csrc/mp_mem_kernels.cu
@@ -217,17 +217,41 @@ __global__ void multi_layer_block_transfer_kernel(
       shape_desc, lmcache_chunk_size);
 }
 
-#define LAUNCH_BLOCK_KERNEL_WITH_FORMAT(DIRECTION, FORMAT)               \
-  multi_layer_block_transfer_kernel<uint4, DIRECTION, FORMAT>            \
+#define LAUNCH_KERNEL(DIRECTION, FORMAT)                                 \
+  multi_layer_block_transfer_kernel<ScalarType, DIRECTION, FORMAT>       \
       <<<grid, block, 0, stream>>>(lmcache_obj4, paged_buffer_ptrs,      \
                                    block_ids_ptr, num_blocks_per_object, \
                                    shape_desc, lmcache_chunk_size,       \
                                    skip_prefix_n_blocks);                \
   C10_CUDA_KERNEL_LAUNCH_CHECK();
 
-}  // namespace
+#define DISPATCH_FORMAT(DIRECTION)                                  \
+  switch (gpu_kv_format) {                                          \
+    case GPUKVFormat::NB_NL_TWO_BS_NH_HS:                           \
+      LAUNCH_KERNEL(DIRECTION, GPUKVFormat::NB_NL_TWO_BS_NH_HS);    \
+      break;                                                        \
+    case GPUKVFormat::NL_X_TWO_NB_BS_NH_HS:                         \
+      LAUNCH_KERNEL(DIRECTION, GPUKVFormat::NL_X_TWO_NB_BS_NH_HS);  \
+      break;                                                        \
+    case GPUKVFormat::NL_X_NB_TWO_BS_NH_HS:                         \
+      LAUNCH_KERNEL(DIRECTION, GPUKVFormat::NL_X_NB_TWO_BS_NH_HS);  \
+      break;                                                        \
+    case GPUKVFormat::NL_X_NB_BS_HS:                                \
+      LAUNCH_KERNEL(DIRECTION, GPUKVFormat::NL_X_NB_BS_HS);         \
+      break;                                                        \
+    case GPUKVFormat::TWO_X_NL_X_NBBS_NH_HS:                        \
+      LAUNCH_KERNEL(DIRECTION, GPUKVFormat::TWO_X_NL_X_NBBS_NH_HS); \
+      break;                                                        \
+    case GPUKVFormat::NL_X_NBBS_ONE_HS:                             \
+      LAUNCH_KERNEL(DIRECTION, GPUKVFormat::NL_X_NBBS_ONE_HS);      \
+      break;                                                        \
+    default:                                                        \
+      TORCH_CHECK(false, "Unsupported GPUKVFormat: ",               \
+                  static_cast<int>(gpu_kv_format));                 \
+  }
 
-void multi_layer_block_kv_transfer(
+template <typename ScalarType>
+void multi_layer_block_kv_transfer_templated(
     const torch::Tensor& paged_buffer_ptrs_tensor,
     std::vector<int64_t> lmcache_objects_ptrs, const torch::Tensor& block_ids,
     const torch::Device& device, TransferDirection direction,
@@ -249,23 +273,19 @@ void multi_layer_block_kv_transfer(
               num_blocks_per_object * shape_desc.bs,
               ") must equal lmcache_chunk_size (", lmcache_chunk_size, ")");
 
-  TORCH_CHECK(
-      shape_desc.hs * shape_desc.element_size % sizeof(uint4) == 0,
-      "head_size * element_size (", shape_desc.hs * shape_desc.element_size,
-      ") must be divisible by ", sizeof(uint4), " for uint4 vectorized access");
-
   // --- Build MemoryObj4 ---
-  MemoryObj4<uint4> lmcache_obj4;
+  MemoryObj4<ScalarType> lmcache_obj4;
   lmcache_obj4.num_objects = num_objects;
   for (int i = 0; i < 4; ++i) {
     lmcache_obj4.objects[i] =
-        (i < num_objects) ? reinterpret_cast<uint4*>(lmcache_objects_ptrs[i])
-                          : nullptr;
+        (i < num_objects)
+            ? reinterpret_cast<ScalarType*>(lmcache_objects_ptrs[i])
+            : nullptr;
   }
 
   // --- Build paged buffer pointer array ---
-  uint4** paged_buffer_ptrs =
-      reinterpret_cast<uint4**>(paged_buffer_ptrs_tensor.data_ptr());
+  ScalarType** paged_buffer_ptrs =
+      reinterpret_cast<ScalarType**>(paged_buffer_ptrs_tensor.data_ptr());
 
   const at::cuda::OptionalCUDAGuard device_guard(device);
   const cudaStream_t stream = at::cuda::getCurrentCUDAStream();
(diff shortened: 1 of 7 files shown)


maobaolong:
@liuyumoye It seems getting conflict with dev now, would you like to resolve the conflict first? Hope to merge this PR so that MP mode can support DSA.

csrc/mp_mem_kernels.cu
ApostaC, line 1: nit: to avoid using if conditions and `use_xxx` flags, we can make the  `multi_layer_block_kv_transfer` a templated function and then have a very simple wrapper to determine the data type.
You can take a look at `mem_kernels.cu` where we did the same thing.

lmcache/v1/gpu_connector/gpu_ops.py
ApostaC, line 1: I don't think we need to touch this file.
`lmcache_memcpy_async_h2d/d2h` doesn't need to know the layout inside the memory object, and it should be called outside the `for group in kv_groups:` loop.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -86,21 +87,55 @@ def __init__(
...
+                    [self.kv_caches_[i].data_ptr() for i in group.layer_indices],
+                    self.device_,
+                )
+            )
+
+        # Backward-compat scalar aliases (group 0)
+        self.hidden_dim_size_ = self.hidden_dim_sizes_[0]
+        self.num_heads_ = self.group_num_heads_[0]
+        self.head_size_ = self.group_head_sizes_[0]
+        self.shape_desc_ = self.shape_descs_[0]
```
ApostaC: Do we really need to keep this backward compatibility? I feel like we can force all the codes in `server.py` to use new interfaces

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -119,17 +154,27 @@ def __init__(
             0, self.block_size_, dtype=torch.long, device=self.device_
         ).unsqueeze(0)
         self.slot_mapping_tensor_ = (offsets + block_ids * self.block_size_).reshape(
-            (self.num_blocks, self.block_size_)
+            (self.num_blocks_, self.block_size_)
         )
```
ApostaC: Actually, this can be dropped.

ApostaC: And also the old `slot_mapping` related apis

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -119,17 +154,27 @@ def __init__(
...
+                ),
+                dtype=group.dtype,
+                device=self.device_,
+            )
+            for group_idx, group in enumerate(
+                self.kv_layer_groups_manager_.kv_layer_groups
+            )
+        ]
+        # Single-group alias for backward compatibility
+        self.tmp_gpu_buffer_ = self.tmp_gpu_buffers_[0]
```
ApostaC: As I mentioned above, for tmp_gpu_buffer, we don't need to create `tmp_gpu_buffer` for each group, but just a "flat" one for all the groups.

ApostaC: We can have a helper function called something like `_get_kv_buffer_shape_unified_group()` to get the shapes.

lmcache/v1/multiprocess/server.py
```diff
@@ -450,35 +466,35 @@ def _retrieve_loop(keys: list[ObjectKey], memory_objs: list[MemoryObj]) -> None:
...
+                        group_kv_pointers,
+                        [tb.data_ptr() for tb in tmp_buffers],
+                        chunk_block_ids_gpu,
+                        gpu_context.device,
+                        lmc_ops.TransferDirection.H2D,
+                        gpu_context.get_shape_desc(group_idx),
+                        self.chunk_size,
+                        gpu_context.gpu_kv_format_,
+                        skip_blocks_in_chunk,
+                    )
```
ApostaC, line 499: With my proposal above, the code will be something like this:
```suggestion
                # H2D copy for all chunks in the batch
                tmp_buffers = gpu_context.get_tmp_gpu_buffer_batched(
                    self.chunk_size, batch_len
                )
                lmcache_memcpy_async_h2d(memory_obj, tmp_buffer, group_idx)
                
                for group_idx in range(num_groups):
                    group_kv_pointers = gpu_context.get_group_kv_pointers(group_idx)
                
                    ### New code to get buffer offset from gpu_context by group_idx    
                    tmp_buffer_offsets = gpu_context.get_tmp_gpu_buffer_offset(group_idx)
                
                    lmc_ops.multi_layer_block_kv_transfer(
                        group_kv_pointers,
                        [tb.data_ptr() + tmp_buffer_offsets for tb in tmp_buffers],
                        chunk_block_ids_gpu,
                        gpu_context.device,
                        lmc_ops.TransferDirection.H2D,
                        gpu_context.get_shape_desc(group_idx),
                        self.chunk_size,
                        gpu_context.gpu_kv_format_,
                        skip_blocks_in_chunk,
                    )
```

ApostaC (changes requested):
High-level comments:
- Tmp buffer and `lmcache_async_memcpy_h2d/d2h` should not be aware of kv cache group information. This can reduce the number of code changes by a lot
- Please add some unit test for 132 int8s to test the kernel support.

Please see the detailed comments below

lmcache/v1/multiprocess/gpu_context.py
liuyumoye: Done — removed the old slot_mapping related APIs as they are no longer needed.

lmcache/v1/multiprocess/gpu_context.py
liuyumoye: Thanks for the suggestion! We've already addressed this in the current implementation.

The `tmp_gpu_buffer_` in `GPUCacheContext` is now **a single flat uint8 buffer** allocated for all groups together, laid out in chunk-major order:

> [ chunk_0: group_0_bytes | group_1_bytes | ... ]
> [ chunk_1: group_0_bytes | group_1_bytes | ... ]
> ...

- `get_tmp_gpu_buffer_flat(chunk_idx)` returns the entire flat slice for a given chunk index (all groups concatenated), which matches the layout of `MemoryObj.raw_data` exactly — enabling a single memcpy for the whole chunk without any group awareness.

- `get_tmp_chunk_gpu_buffer(group_idx) `and `get_tmp_chunk_gpu_buffer_batched(batch_size, group_idx)` are only used internally when the kernel needs a per-group shaped view, but they are just views into the same flat buffer — no separate allocation per group.

The `lmcache_async_memcpy_h2d / d2h` calls in `server.py` only use `get_tmp_gpu_buffer_flat`, so they have no awareness of KV cache group information at all.

liuyumoye:
> @liuyumoye It seems getting conflict with dev now, would you like to resolve the conflict first? Hope to merge this PR so that MP mode can support DSA.

Thanks for pointing that out! I've resolved the merge conflict with the latest dev branch. The PR is ready for review again. Please let me know if there are any other issues.

csrc/mp_mem_kernels.cu
ApostaC, line 1: nit note: usually we put `#define` and `#undef` outside the function body.

lmcache/v1/gpu_connector/gpu_ops.py
```diff
@@ -24,19 +24,30 @@ def lmcache_memcpy_async_h2d(
     :param MemoryObj memory_obj: The memory object to be copied.
     :param torch.Tensor gpu_buffer: The GPU buffer to copy the data to.
     """
-    assert memory_obj.tensor is not None
-    assert memory_obj.tensor.numel() == gpu_buffer.numel()
+    src_tensor = memory_obj.raw_data
```
ApostaC: nit: not sure MemoryObj.raw_data is a public&stable API or not. But I do see there is `data_ptr()` property define in the MemoryObj base class. We can use that directly when calling `lmc_ops.lmcache_memcpy_async` instead.

lmcache/v1/gpu_connector/gpu_ops.py
```diff
@@ -51,16 +62,27 @@ def lmcache_memcpy_async_d2h(
     :param torch.Tensor gpu_buffer: The GPU buffer to copy the data from.
     :param MemoryObj memory_obj: The memory object to be copied to.
     """
-    assert memory_obj.tensor is not None
-    assert memory_obj.tensor.numel() == gpu_buffer.numel()
+    dst_tensor = memory_obj.raw_data
```
ApostaC: same nit comment as above

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -261,39 +286,70 @@ def concrete_gpu_kv_shape(self) -> str:
         """Returns the GPU KV shape with actual numeric values substituted."""
         return get_concrete_gpu_kv_shape(self.kv_caches_, self.gpu_kv_format_)
 
-    def get_tmp_gpu_buffer(self, num_tokens: int) -> torch.Tensor:
+    def get_group_kv_pointers(self, group_idx: int) -> torch.Tensor:
+        """Returns the pre-computed GPU tensor of KV cache pointers for the
+        given group."""
+        return self.group_kv_pointers_[group_idx]
+
+    def get_tmp_gpu_buffer_flat(self, chunk_idx: int = 0) -> torch.Tensor:
```
ApostaC: nit: let's avoid using a default parameter for `chunk_idx`. We should make sure that the caller understands it needs to pass in `chunk_idx` because it directly relates to the batching logic.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -261,39 +286,70 @@ def concrete_gpu_kv_shape(self) -> str:
...
+        """Returns the flat uint8 view of the temporary GPU buffer for the
+        given chunk index, covering all KV layer groups.
+
+        The returned slice has exactly ``tmp_chunk_bytes_`` bytes and its
+        layout matches ``MemoryObj.raw_data`` (groups concatenated in order),
+        so it can be copied to/from a MemoryObj with a single memcpy.
```
ApostaC: nit: Let's avoid using `tmp_chunk_bytes_` and `MemoryObj.raw_data` in docstring to avoid confusion for other developers. Ideally, we can say something like:
```
The returned tensor will fit a memory full object corresponding ``self.chunk_size`` tokens.
```

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -261,39 +286,70 @@ def concrete_gpu_kv_shape(self) -> str:
...
 
-        Note:
-            The returned buffer is guaranteed to be contiguous
+    def get_tmp_chunk_gpu_buffer(self, group_idx: int = 0) -> torch.Tensor:
         """
-        shape = self.get_kv_buffer_shape(num_tokens)
-        num_elems = shape.numel()
-        return self.tmp_gpu_buffer_.flatten()[:num_elems].view(shape)
+        Returns a view of the temporary GPU buffer for the given group,
+        sized for a single request of ``num_tokens`` tokens (chunk 0).
```
ApostaC: nit: `num_tokens` --> `lmcache_chunk_size`. Also, the `(chunk 0)` at the end is a bit confusing.

lmcache/v1/multiprocess/server.py
```diff
@@ -797,7 +814,7 @@ def report_status(self) -> dict:
                 entry["kv_cache_layout"] = {
                     "num_layers": ctx.num_layers,
                     "block_size": ctx.block_size,
-                    "hidden_dim_size": ctx.hidden_dim_size,
+                    "hidden_dim_sizes": str(ctx.hidden_dim_sizes_),
```
ApostaC: nit: we should not use private members here. Let's have a property defined as `hidden_dim_sizes` in `GPUCacheContext`

ApostaC (approved):
All are nit comments. Otherwise LGTM!

liuyumoye:
> All are nit comments. Otherwise LGTM!

Thanks for the review! All nit comments have been addressed.

maobaolong (approved):
@liuyumoye Thanks for this feature, LGTM.

csrc/mp_mem_kernels.cu
```diff
@@ -274,68 +294,51 @@ void multi_layer_block_kv_transfer(
...
+    const torch::Tensor& paged_buffer_ptrs_tensor,
+    std::vector<int64_t> lmcache_objects_ptrs, const torch::Tensor& block_ids,
+    const torch::Device& device, TransferDirection direction,
+    PageBufferShapeDesc shape_desc, int lmcache_chunk_size,
+    GPUKVFormat gpu_kv_format, int skip_prefix_n_blocks) {
+  int head_bytes = shape_desc.hs * shape_desc.element_size;
+  TORCH_CHECK(head_bytes % sizeof(uint16_t) == 0, "head_size * element_size (",
+              head_bytes, ") must be divisible by 2 for vectorized access");
+
+  if (head_bytes % sizeof(uint4) == 0) {
```
chunxiaozheng, line 335: @liuyumoye could we add some comments to indicate how many bytes?

csrc/mp_mem_kernels.cu
```diff
@@ -274,68 +294,51 @@ void multi_layer_block_kv_transfer(
...
+    LAUNCH_TEMPLATED(uint4);
```
chunxiaozheng: Besides, could we add some 8 bytes or 1 bytes copy?

chunxiaozheng: After discuss offline, there is no need to add 8 bytes or 1 bytes.

chunxiaozheng (approved):
LGTM!

princepride:
So can I use mp LMCache when I deploy GLM-5 now?

maobaolong:
@princepride yes.

princepride:
> @princepride yes.

Thank you! Have you seen I joined the latest `月球大叔` live streaming? I still remember about 1 year ago, we left comments under his channel😊, It's not very long ago. BTW, I want add your wechat.

princepride:
> @princepride yes.

I left comments on your slack

PR #2932, opened 2026-04-02
fix(security): expand multimodal hash entropy for cache keys

### LINKED ISSUE #3301: [Security] 16-bit multimodal hash collision can poison KV cache entries
**Describe the bug**
LMCache's multimodal caching path reduces multimodal identifiers, including image hashes or OpenAI-style image identifiers, to a 16-bit value before using them to modify token IDs that participate in cache key computation.

The affected function is `hex_hash_to_int16()` in `lmcache/integration/vllm/utils.py`. It either masks parsed hexadecimal input with `0xFFFF` or hashes non-hex input with SHA-256 and keeps only the first 2 bytes. This creates only 65,536 possible multimodal hash values.

As a result, distinct images can produce the same multimodal token replacement value. When those token IDs are used to compute cache keys, LMCache can incorrectly return KV cache data generated for a different image. This can cause incorrect model output and may expose cross-user visual context through cache poisoning.

**To Reproduce**

1. Use LMCache with vLLM multimodal requests where image identifiers are passed through `hex_hash_to_int16()` and `apply_mm_hashes_to_token_ids()`.
2. Generate many distinct multimodal image identifiers, for example OpenAI-style identifiers such as `chatcmpl-<uuid>-image-<n>`.
3. Convert each identifier with the same 16-bit logic used by `hex_hash_to_int16()`.
4. Observe that collisions typically occur after only a few hundred inputs due to the birthday paradox.
5. Submit two requests whose distinct image identifiers collide in the 16-bit multimodal hash value and share the relevant prompt/token structure.
6. Observe that the modified token IDs can produce the same cache key for different image contexts, allowing a cache hit for the wrong KV cache.

Minimal standalone collision demonstration:

```python
import hashlib
import string
import uuid


def hex_hash_to_int16(s):
    s = "" if s is None else str(s)
    s_stripped = s.strip()
    hex_part = s_stripped[2:] if s_stripped.lower().startswith("0x") else s_stripped
    if hex_part and all(c in string.hexdigits for c in hex_part):
        try:
            return int(hex_part, 16) & 0xFFFF
        except ValueError:
            pass
    digest = hashlib.sha256(s_stripped.encode("utf-8")).digest()
    return int.from_bytes(digest[:2], byteorder="big", signed=False)


seen = {}
for i in range(100_000):
    ident = f"chatcmpl-{uuid.uuid4().hex}-image-{i}"
    h = hex_hash_to_int16(ident)
    if h in seen:
        print(f"COLLISION at input #{i}:")
        print(f"  A: {seen[h]}")
        print(f"  B: {ident}")
        print(f"  hash: {h:#06x}")
        break
    seen[h] = ident
```

**Expected behavior**
Different multimodal inputs should not be collapsed into a 16-bit identifier for cache key generation. The multimodal cache key material should preserve enough entropy to make cross-image collisions computationally infeasible in normal production use.

A practical fix is to expand the derived value from 16 bits to at least 64 bits, because `token_ids` already uses `torch.long` / int64:

```python
return int(hex_part, 16) & 0xFFFFFFFFFFFFFFFF
return int.from_bytes(digest[:8], byteorder="big", signed=False)
```

The function should also be renamed from `hex_hash_to_int16()` to something like `hex_hash_to_int64()` if the implementation changes to 64 bits.

**Screenshots**
Not applicable.

 - OS: Not OS-specific
 - Browser: Not applicable
 - Version: Current `dev` branch, observed at commit `729ff73` according to the disclosure draft

 - Device: Not applicable
 - OS: Not applicable
 - Browser: Not applicable
 - Version: Not applicable

**Additional context**
Security impact:

- Integrity: High. The model can answer using KV cache data derived from a different image than the one submitted by the user.
- Confidentiality: Medium. A user may receive behavior influenced by another user's private image context if cache keys collide.
- Availability: Low. The service remains operational.
- Attack vector: Network-accessible in deployments accepting multimodal requests.
- Exploitability: Low complexity. Brute-force collision is guaranteed within 65,536 attempts, and natural collisions are expected after roughly 300 distinct images.

3em0:

- Replaces the old 16-bit multimodal hash key material with a signed-int64-safe 63-bit value before writing it into token IDs used for cache key computation.
- Adds `hex_hash_to_int64()` and keeps `hex_hash_to_int16()` as a deprecated compatibility wrapper that delegates to the safer implementation.
- Updates multimodal hash tests for signed int64 bounds, deterministic non-hex handling, large-hex masking, deprecated alias compatibility, and collision resistance across a realistic working set.

Fixes LMCache/LMCache#3301

The implementation masks to `0x7FFFFFFFFFFFFFFF` rather than the full unsigned 64-bit range so values remain safe for `torch.long` / signed int64 tensors.

This branch has been refreshed so GitHub reports it as mergeable with the current `dev` branch.

**If applicable**:

**Test Plan**:

- [x] `python -m compileall -q lmcache/integration/vllm/utils.py lmcache/integration/vllm/tests/test_mm_hash_utils.py`
- [x] `git diff --check`
- [x] `ruff check lmcache/integration/vllm/utils.py lmcache/integration/vllm/tests/test_mm_hash_utils.py`
- [ ] `pytest -q lmcache/integration/vllm/tests/test_mm_hash_utils.py` (blocked locally: current Python environment is missing `torch`)

### DIFF
--- a/lmcache/integration/vllm/tests/test_mm_hash_utils.py
+++ b/lmcache/integration/vllm/tests/test_mm_hash_utils.py
@@ -8,8 +8,10 @@ import torch
 # First Party
 from lmcache.integration.vllm.utils import (
     apply_mm_hashes_to_token_ids,
-    hex_hash_to_int16,
+    hex_hash_to_int64,
 )
+
+INT64_MAX = 0xFFFFFFFFFFFFFFFF
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


3em0:
@da-x @yanok @thesues @lalith-b

PR #2887, opened 2026-03-27
Copy a snapshot of lmcache_mp_connector.py for vllm 0.18.0

maobaolong:
**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

---

> [!NOTE]
> **Medium Risk**
> Adds a large new vLLM integration module that implements async KV cache lookup/retrieve/store orchestration; while mostly additive, it touches performance- and correctness-sensitive cache-transfer logic (GPU/ZMQ/stream synchronization).
> 
> **Overview**
> Adds a version-pinned `LMCacheMPConnector` implementation (`lmcache_mp_connector_0180.py`) for vLLM `0.18.0`, including request tracking/state transitions, lookup vs. vLLM-hit reconciliation, batched async retrieve/store submission, and MLA-aware rank/world-size handling.
> 
> Updates tooling configs to *exclude* `lmcache_mp_connector_*.py` snapshots from `pre-commit`, `ruff`, `mypy`, and `codespell`, reducing lint/typecheck noise for these vendored compatibility files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8589997f3e0e6f4986679d3f7eb2b5b22ecc2e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2935, opened 2026-04-02
[MP][optimize] optimize save when mla enabled

chunxiaozheng:
When `MLA` is enabled, `store` or `retrieve` requests only need to be sent once in multi workers, which can greatly reduce the number of requests in the server. 
The current PR only modifies `store` requests and will modify `retrieve` requests in the next request.

---

> [!NOTE]
> **Medium Risk**
> Changes how `world_size`, `tp_size`, and `worker_id` are derived for LMCache multiprocess requests, which can affect cache-keying and cross-worker coordination in distributed runs. Backward-compat defaults exist, but misconfigured `parallel_strategy` could lead to incorrect sharing or runtime errors.
> 
> **Overview**
> Adds a new `ParallelStrategy` dataclass and wires it into `LMCacheMPSchedulerAdapter` and `LMCacheMPWorkerAdapter` so cache-key fields (`world_size`, `worker_id`) and `tp_size` can be computed from KV-parallel topology (e.g., MLA excluding TP) rather than the raw process rank.
> 
> Updates adapter constructors with defaults and new properties (`world_size`, `tp_size`, `worker_id`, plus `use_mla`/`is_first_rank_of_pp_group` on the worker) and adjusts `FREE_LOOKUP_LOCKS` tests to construct adapters via `parallel_strategy` instead of directly setting `world_size`/`tp_size`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac60e130e96eca575f186b34c5d0851ec6d749f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ApostaC (commented):
Quick comment: I think right now we have `world_size` and `kv_rank` in the initialization args, though their meaning is not super clear. How about we define a new dataclass to describe the parallel strategy and local rank information (and include mla flag there)?

chunxiaozheng:
> Quick comment: I think right now we have `world_size` and `kv_rank` in the initialization args, though their meaning is not super clear. How about we define a new dataclass to describe the parallel strategy and local rank information (and include mla flag there)?

@ApostaC Sorry, I just saw the message and I think it's okay, thanks for your review.

chunxiaozheng:
@ApostaC I have updated, could you help take a look?

(the change of `LMCacheMPConnector` is https://github.com/vllm-project/vllm/pull/38810)

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -96,6 +96,36 @@ def send_ping(
...
 
+@dataclass
+class ParallelStrategy:
+    use_mla: bool
+    """Whether to use the MLA."""
+
+    world_size: int
+    """
+    The world size, world_size may not be equal to the actual_world_size, 
+    in the case of mla, it will 'exclude' the effect of TP.
```
ApostaC: What does it mean by `"exclude" the effect of TP`? Does that mean for MLA TP 8 case, we will have:
- `world_size` = 1, `actual_world_size` = 8
- `worker_id` = 1, `actualy_worker_id` = <TP rank in the world>

In this case, can we name it to `kv_world_size` and `kv_worker_id`, which means how do we divide the KV cache. WDYT?

ApostaC (commented):
I like this PR! One thing on my mind is how to introduce the PRs on both the LMCache & vLLM side without breaking the CI/CD.

**Potential idea:**
- Step 1: introduce `ParallelStrategy` as an optional argument with default value None, and keep the old world_size and kv_rank parameter with default value None as well. (Probably we need some logic to handle both the old version and new version). -- This will not break LMCache CI .
- Step 2: update the vLLM side code to use `ParallelStrategy` and remove the old world_size and kv_rank arguments. Since in step 1, we made world_size and kv_rank optional with a default value None, it will not break LMCache CI either.
- Step 3: clean up the LMCache-side code, to completely remove the old world_size and kv_rank related arguments.

I hate this complicated process, so let me know if you have better ideas @chunxiaozheng

lmcache/integration/vllm/vllm_multi_process_adapter.py
chunxiaozheng: @ApostaC Good idea, I will update here.

chunxiaozheng:
> I like this PR! One thing on my mind is how to introduce the PRs on both the LMCache & vLLM side without breaking the CI/CD.
> 
> **Potential idea:**
> 
> * Step 1: introduce `ParallelStrategy` as an optional argument with default value None, and keep the old world_size and kv_rank parameter with default value None as well. (Probably we need some logic to handle both the old version and new version). -- This will not break LMCache CI .
> * Step 2: update the vLLM side code to use `ParallelStrategy` and remove the old world_size and kv_rank arguments. Since in step 1, we made world_size and kv_rank optional with a default value None, it will not break LMCache CI either.
> * Step 3: clean up the LMCache-side code, to completely remove the old world_size and kv_rank related arguments.
> 
> I hate this complicated process, so let me know if you have better ideas @chunxiaozheng

@ApostaC Thanks for your suggestion, Iet me think about this.

lmcache/integration/vllm/vllm_multi_process_adapter.py
chunxiaozheng: @ApostaC  The meaning of `"exclude" the effect of TP` can refer to the following annotations in `LMCacheMPConnector ` 
```
def extract_world_size_and_kv_rank(
    world_size: int,
    rank: int,
    vllm_config: VllmConfig,
) -> tuple[int, int]:
    """
    Convert the rank for the MLA.
    """
    use_mla = mla_enabled(vllm_config.model_config)
    if not use_mla:
        return world_size, rank
    else:
        # Tensor parallel does not change the KV caches for MLA models.
        # So we need to "exclude" the effect of TP on rank and world size
        tp_size = vllm_config.parallel_config.tensor_parallel_size
        # vLLM constructs TP groups first, and then construct other
        # parallel groups on top of TP groups.
        # for example, TP=4, PP=2,
        # PP group: [0, 1, 2, 3], [4, 5, 6, 7]
        # TP group: [0, 4], [1, 5], [2, 6], [3, 7]
        # So we can "exclude" the effect of TP by rank // tp_size.
        return world_size // tp_size, rank // tp_size
```

chunxiaozheng:
> > I like this PR! One thing on my mind is how to introduce the PRs on both the LMCache & vLLM side without breaking the CI/CD.
> > **Potential idea:**
> > 
> > * Step 1: introduce `ParallelStrategy` as an optional argument with default value None, and keep the old world_size and kv_rank parameter with default value None as well. (Probably we need some logic to handle both the old version and new version). -- This will not break LMCache CI .
> > * Step 2: update the vLLM side code to use `ParallelStrategy` and remove the old world_size and kv_rank arguments. Since in step 1, we made world_size and kv_rank optional with a default value None, it will not break LMCache CI either.
> > * Step 3: clean up the LMCache-side code, to completely remove the old world_size and kv_rank related arguments.
> > 
> > I hate this complicated process, so let me know if you have better ideas @chunxiaozheng
> 
> @ApostaC Thanks for your suggestion, Iet me think about this.

@ApostaC If the `LMCacheMPConnector` is also maintained in the `LMCache` repository, would it simplify a lot? we only need to update the code in `LMCache`, @maobaolong has already submitted some PRs about this.
1. https://github.com/LMCache/LMCache/pull/2887  
2. https://github.com/vllm-project/vllm/pull/38301

ApostaC:
@chunxiaozheng I think the problem is the CI/CD. Right now, we are using vllm-side connector implementation to do the CI/CD. In that case, the CI/CD may still become red.

chunxiaozheng:
> @chunxiaozheng I think the problem is the CI/CD. Right now, we are using vllm-side connector implementation to do the CI/CD. In that case, the CI/CD may still become red.

@ApostaC Okay, I think the suggestion you gave above is a good choice.

chunxiaozheng:
@ApostaC I have updated this PR and it will be fully compatible with the previous `lmcache_mp_connector`, could you help take a look again.

ApostaC (approved):
LGTM!

maobaolong (approved):
LGTM.

PR #2952, opened 2026-04-03
[MP][Bugfix] Fix deadlock caused by cuda launch host func

ApostaC:

`EventBus.publish_on_stream()` schedules Python callbacks on CUDA streams via
`cupy_stream.launch_host_func(self.publish, event)`. This causes a CUDA
driver / GIL deadlock: the host-function callback holds the CUDA driver lock
and tries to acquire the GIL, while another thread holds the GIL and waits on
the CUDA driver lock.

This PR introduces a C++ `EventRecorder` that calls `cudaLaunchHostFunc`
directly. The callback runs entirely in native code (no GIL), stamps the
wall-clock timestamp, and buffers events in a thread-safe C++ vector. The
EventBus drain thread periodically pulls these buffered events into the Python
dispatch queue via `drain_recorded_events()`.

Changes:
- New `csrc/event_recorder.{h,cpp}`: `EventRecorder` singleton with
  `cudaLaunchHostFunc` callback and `drain()` API
- `csrc/pybind.cpp`: bind `record_event_on_stream` (GIL released) and
  `drain_recorded_events`
- `event_bus.py`: `publish_on_stream` routes through C++ when available;
  `_drain_all` drains the C++ buffer before processing the Python queue
- Fallback to the old Python path when `lmcache.c_ops` is unavailable
- New end to end CI test to prevent deadlock happen again

- `record_event_on_stream` releases the GIL via `py::call_guard` to avoid the
  reverse deadlock (Python thread → CUDA driver lock)
- ROCm/HIP supported via `#ifdef USE_ROCM` (maps to `hipLaunchHostFunc`)
- The 4 affected callsites (`MP_STORE_START/END`, `MP_RETRIEVE_START/END`) in
  `server.py` are unchanged — the fix is internal to `EventBus`

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Adds new CUDA/HIP host-callback C++ code and pybind APIs on the MP observability hot path; mistakes could cause crashes, leaks, or missing events, but behavior is gated with a Python fallback and covered by new tests.
> 
> **Overview**
> Fixes a multiprocessing observability deadlock by switching `EventBus.publish_on_stream` from scheduling a Python callback on the CUDA stream to scheduling a native `cudaLaunchHostFunc`/`hipLaunchHostFunc` callback that records events without touching the GIL.
> 
> Introduces a C++ `EventRecorder` buffer (`record_event_on_stream` + `drain_recorded_events`) exposed via `lmcache.c_ops`, and updates the EventBus drain loop to pull native-recorded events into the normal Python dispatch queue (falling back to the old Python path when the native API isn’t available).
> 
> Adds CUDA unit/integration tests for the recorder + EventBus routing, updates build configuration to compile the new source, and adds a dedicated Buildkite multiprocess `deadlock` regression step (while skipping the new CUDA-only test on ROCm CI).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7b6bfd44f8591554225238e0d53c86d3c587f64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.buildkite/k3_tests/multiprocess/scripts/run-deadlock.sh
+++ b/.buildkite/k3_tests/multiprocess/scripts/run-deadlock.sh
@@ -23,6 +23,29 @@ BUILD_ID="${BUILD_ID:-local_$$}"
 PID_FILE="/tmp/lmcache_mp_pids_${BUILD_ID}"
 TIMEOUT_SECONDS=180   # 3 minutes
 
+# ── Install py-spy for deadlock diagnosis ──────────────────
+echo "=== Installing py-spy ==="
+uv pip install py-spy
+PY_SPY="$(which py-spy)"
+echo "py-spy installed at: $PY_SPY"
+
+PYSPY_LOG="/tmp/build_${BUILD_ID}_pyspy.log"
+
+# ── Helper: dump stacks of server processes via py-spy ─────
+dump_stacks() {
+    echo "" | tee -a "$PYSPY_LOG"
+    echo "=== py-spy stack dump (native + Python) ===" | tee -a "$PYSPY_LOG"
+
+    if kill -0 "$LMCACHE_PID" 2>/dev/null; then
+        echo "" | tee -a "$PYSPY_LOG"
+        echo "--- LMCache server (PID=$LMCACHE_PID) ---" | tee -a "$PYSPY_LOG"
+        sudo "$PY_SPY" dump --pid "$LMCACHE_PID" --native 2>&1 | tee -a "$PYSPY_LOG" || true
+    fi
+
+    # Copy to repo root so cleanup.sh collects it as a Buildkite artifact
+    cp "$PYSPY_LOG" "${REPO_ROOT}/build_${BUILD_ID}_pyspy.log" 2>/dev/null || true
+}
+
 # ── 1. Launch LMCache server ───────────────────────────────
 echo "=== Launching LMCache server ==="
 echo "Port: $LMCACHE_PORT"
@@ -68,6 +91,7 @@ vllm serve "$MODEL" \
     --max-num-batched-tokens 16000 \
     --scheduling-policy fcfs \
     --port "$SAVED_VLLM_PORT" \
+    --enforce-eager \
     --kv-transfer-config "{\"kv_connector\":\"LMCacheMPConnector\", \"kv_role\":\"kv_both\", \"kv_load_failure_policy\": \"recompute\", \"kv_connector_extra_config\": {\"lmcache.mp.port\": $LMCACHE_PORT, \"lmcache.mp.mq_timeout\": 60}}" \
     > "/tmp/build_${BUILD_ID}_vllm.log" 2>&1 &
 
--- a/.buildkite/k3_tests/multiprocess/scripts/run-deadlock.sh
+++ b/.buildkite/k3_tests/multiprocess/scripts/run-deadlock.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+# Self-contained deadlock regression test.
+#
+# Launches DeepSeek-V2-Lite-Chat with TP=2 (both GPUs) + LMCache server,
+# sends 50 requests with ~30K token prefixes, and verifies they all
+# complete within 3 minutes.  A CUDA-driver/GIL deadlock would cause
+# requests to hang indefinitely, failing the timeout.
+#
+# This test is self-contained: it handles its own server lifecycle
+# instead of using the standard launch-processes.sh / wait-for-servers.sh.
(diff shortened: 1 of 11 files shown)


ApostaC:
@maobaolong I've tested it locally. This should fix the deadlock problem (at least for me)

royyhuang (approved):
LGTM!

.buildkite/k3_tests/multiprocess/scripts/run-deadlock.sh
```diff
@@ -0,0 +1,113 @@
...
+if ! timeout "$TIMEOUT_SECONDS" lmcache bench engine \
+        --engine-url "http://localhost:${VLLM_PORT}" \
+        --workload random-prefill \
+        --tokens-per-gb-kvcache 6000 \
+        --rp-request-length 30000 \
+        --rp-num-requests 50 \
+        --no-interactive \
+        --no-csv \
+        -q; then
+    echo "FAIL: Benchmark failed or timed out (possible deadlock)"
```
maobaolong, line 125: Maybe we need capture several thread stack and update to a place so the PR author can find the root cause of deadlook.

lmcache/v1/mp_observability/event_bus.py
```diff
@@ -120,14 +131,33 @@ def register_subscriber(self, subscriber: EventSubscriber) -> None:
...
+                    str_metadata[k] = str(v)
+            _lmc_ops.record_event_on_stream(
+                stream.ptr,
+                event.event_type.value,
+                event.session_id,
+                str_metadata,
+                int_metadata,
+            )
+        else:
+            stream.launch_host_func(self.publish, event)
```
maobaolong, line 160: When the lmc_ops can not be imported? UT?

csrc/pybind.cpp
```diff
@@ -84,4 +85,9 @@ PYBIND11_MODULE(c_ops, m) {
       .def_readwrite("nh", &PageBufferShapeDesc::nh)
       .def_readwrite("hs", &PageBufferShapeDesc::hs)
       .def_readwrite("element_size", &PageBufferShapeDesc::element_size);
+  m.def("record_event_on_stream", &record_event_on_stream,
+        py::arg("cuda_stream_ptr"), py::arg("event_type_name"),
+        py::arg("session_id"), py::arg("str_metadata"), py::arg("int_metadata"),
+        py::call_guard<py::gil_scoped_release>());
```
maobaolong, line 91: 👍 This release the gil so there is no chance hold gil then require cuda driver lock.

csrc/event_recorder.cpp
```diff
@@ -0,0 +1,83 @@
...
+    const std::string& session_id,
+    const std::unordered_map<std::string, std::string>& str_metadata,
+    const std::unordered_map<std::string, int64_t>& int_metadata) {
+  auto* event = new PendingEvent{
+      event_type_name, session_id, 0.0, str_metadata, int_metadata,
+  };
+
+  auto stream = reinterpret_cast<lmcache_stream_t>(
+      static_cast<uintptr_t>(cuda_stream_ptr));
+  LMCACHE_LAUNCH_HOST_FUNC(stream, event_host_callback, event);
```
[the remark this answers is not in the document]
maobaolong: @ApostaC How about this suggestion?

maobaolong (commented):
@ApostaC Great to see this PR, and thanks for help to resolve the deadlock issue as quick as possible.  This fix is more thorough. It's a fundamental fix, which is greater than #2823 .

And i like the work on `New end to end CI test to prevent deadlock happen again`.

maobaolong:
> @maobaolong I've tested it locally. This should fix the deadlock problem (at least for me)

And me, I've tested it in my env just now. @ApostaC

lmcache/v1/mp_observability/event_bus.py
ApostaC: If there is no cuda environment, some of the lmc_ops cannot work. (For example in this PR, it needs cudaLaunchHostFunc, which cannot be called without a cuda environment)

.buildkite/k3_tests/multiprocess/scripts/run-deadlock.sh
ApostaC: Good point, let me see if I can add this

ApostaC: Added 👌

maobaolong (approved):
@ApostaC LGTM

sammshen (approved):
LGTM!

PR #2966, opened 2026-04-06
feat: Add batched_contains() override for NixlDynamicStorageBackend

jgoldsch12:
Introduce a synchronous batched_contains() that replaces the sequential per-key contains() loop with a single batched query_memory call via a new batched_nixl_desc_exists() agent method.  Keys resolved from local data structures (put-task set and presence cache) are checked first; only the remainder goes to the remote NIXL agent.

Also extract _exists_in_put_tasks_or_cache() helper shared by both contains() and batched_contains(), and replace the magic 16 in batched_async_contains with the _CONTAINS_BATCH_SIZE constant.  This PR does not change the batching behavior in batched_async_contains.  Async_loading feature is not particularly stable, so this method was left in place and can be changed at another time.

Using this batched_contains resulted in up to 65% lower TTFT at 127K context compared to prior to this fix.  This was tested using the OBJ plugin with the presence cache disabled and paired with a change in NIXL to utilize HeadObjectAsync.

Note: To get the full benefit, this needs to be combined with NIXL PR: https://github.com/ai-dynamo/nixl/pull/1496

---

> [!NOTE]
> **Medium Risk**
> Changes the NIXL dynamic storage existence-check path to use a new batched `query_memory` call and refactors `contains()` logic, which can affect cache hit/miss behavior and remote I/O patterns. Adds unit tests for the new semantics but CI currently skips the new test file.
> 
> **Overview**
> Adds a synchronous `batched_contains()` override for `NixlDynamicStorageBackend` that resolves prefix hits via local put-task/presence-cache checks, then issues a single batched NIXL `query_memory` call for the remaining keys and caches remote hits.
> 
> Introduces `NixlDynamicStorageAgent.batched_nixl_desc_exists()` (returns consecutive-exists count), factors local-only logic into `_exists_in_put_tasks_or_cache()` used by both `contains()` and `batched_contains()`, and replaces the hardcoded async batch size with `_CONTAINS_BATCH_SIZE`.
> 
> Adds mock-based unit tests covering batched query semantics and key-formatting, and updates Buildkite to `--ignore` the new test file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32633b1969b189fe53c5873ab94053e2b98d48e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -1241,21 +1298,67 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
...
+        reg_list = [(0, 0, 0, self._format_object_key(key)) for key in remaining_keys]
+
+        # Use the agent's batched_nixl_desc_exists method
+        consecutive_hits = self.agent.batched_nixl_desc_exists(reg_list)
+
+        # Update cache for the hits and return total count
+        for i in range(consecutive_hits):
+            self._cache_add(remaining_keys[i].chunk_hash)
+
+        return true_count + consecutive_hits
```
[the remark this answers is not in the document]
jgoldsch12: This is a prefix-based match.  The keys list represents a contiguous prefix of chunks for a single request.
The batched_contains contract is to count consecutive hits from the start. If key at position i is a miss, everything after it is irrelevant -- the caller won't use anything past the first gap.  This scenario cannot happen.

tests/v1/test_nixl_batched_contains.py
```diff
@@ -0,0 +1,314 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for ``batched_contains()`` and ``batched_nixl_desc_exists()``.
+
+All tests use lightweight mocks so they run without NIXL or CUDA hardware.
+The mock approach binds the *real* unbound method onto a ``Mock(spec=…)``
+object, so the actual implementation is exercised while external
+dependencies (NIXL agent, memory allocators) are replaced by mocks.
+"""
+
```
DongDongJu, line 9: Could you add pytest.importorskip("nixl")?

DongDongJu (changes requested):
I requested one nit.
Otherwise LGTM.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -1241,21 +1301,70 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
...
         xfer_state = self.key_exists(key)
         if xfer_state:
             self._cache_add(key.chunk_hash)
 
         return xfer_state
 
+    def batched_contains(
+        self,
+        keys: List[CacheEngineKey],
+        pin: bool = False,
```
vvenkates27, line 1318: Is this argument (pin) used?

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -54,6 +54,9 @@
 
 logger = init_logger(__name__)
 
+# Max concurrency for parallel S3 HEAD requests in batched_contains().
```
vvenkates27, line 57: Is this needed as HEAD requests are parallelized in NIXL?

vvenkates27 (approved):
Overall LGTM, couple of minor comments.

vvenkates27:
cc: @yanok

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -479,6 +482,40 @@ def nixl_desc_exists(self, meta_info: str) -> bool:
...
+            ``query_memory`` call are caught internally and logged as
+            warnings; the method returns ``0`` in that case.
+        """
+        if not reg_list:
+            return 0
+
+        try:
+            resp = self.nixl_agent.query_memory(
+                reg_list, self.backend, mem_type=self.mem_type
+            )
```
yanok, line 504: There is a down-side of doing it in a single queryMem call: it will block until it has all the results. But here we only care about the prefix, so if the first block is missing, we will just waste time waiting to get results for all other blocks. But this is a limitation of NIXL API. Just something to be aware of.

lmcache/v1/storage_backend/nixl_storage_backend.py
yanok: It's part of the API, but AFAICT we never supported it.

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -1241,21 +1301,70 @@ def contains(self, key: CacheEngineKey, pin: bool = False) -> bool:
...
+
+        # First, do fast sequential check of local data structures
+        true_count = 0
+        for key in keys:
+            found, result = self._exists_in_put_tasks_or_cache(key)
+            if found:
+                if result:
+                    true_count += 1
+                else:
+                    # Found in put tasks (False), stop the loop
```
yanok, line 1345: well, I have a question here, but it's actually unrelated to this PR (since you just preserved the existing behavior).

So, the question is: if a key is in put tasks, yes, sure, it's not yet on the storage, so *technically* it's correct to return false from contains. But if you think about it, `contains` is used in two cases:
- when we need to load/compute a KV: but in this case, if the corresponding key is in put tasks, we don't even need to load it: it's already in DRAM/VRAM somewhere. So, today this situation is possible: one request offloads a KV, while another one looks for this exact key. Contains will return false and it will be recomputed, but we likely want to just re-use the KV from the offload buffer instead.
- when we check whether we need to save a computed KV: in this case, contains again says false, so the KV will be saved the second time (probably will be deduped in the backend, but it's the job of a backend now).

So, I think there is virtually no scenario where returning False for keys in the put tasks is desired. But maybe I'm missing something.

yanok (approved):

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: I think in real world situations it is likely that another "tier" of KVCache will have the KV corresponding to the key while put operation is in progress.  In some of the targeted testing we do where prefix caching and CPU caching are disabled, yes what you describe is likely.

lmcache/v1/storage_backend/nixl_storage_backend.py
jgoldsch12: It is not.  However, (as I stated in the PR description) I specifically did not want to mess with the async loading batch contains.  I've had a lot of trouble testing async loading, so I didn't want to try and bundle a change to both paths in this one PR.

tests/v1/test_nixl_batched_contains.py
jgoldsch12: done!  Thank you.

lmcache/v1/storage_backend/nixl_storage_backend.py
yanok: I think there is an idea of making query just another op type and then you would be able to launch multiple queries asynchronously like we do with data transfer. Together with the new status list API, this would allow solving the problem with having to wait for all results even if the first block is already missing. But that's just something to have in mind, nothing needs to be done in this PR.

yanok (approved):

jgoldsch12:
@DongDongJu could you help get this change merged.  I believe all comments have been addressed.  Thanks.

DongDongJu (approved):
LGTM.

sammshen (approved):
LGTM

PR #2996, opened 2026-04-10
refactor(mp): replace job_id with request_id in query_prefetch_status

yoo-kumaneko:
The server-internal integer job_id was unnecessarily leaked to the vLLM adapter via _lookup_job_ids / _finished_lookup_jobs, adding a redundant layer of indirection. The external request_id (str) is the natural key and is already stored in every _PrefetchJob.

No changes to storage_manager.py or prefetch_controller.py; the internal prefetch_request_id (int) inside PrefetchHandle is unaffected.

Signed-off-by: crclq2018@gmail.com

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Changes the multiprocess wire protocol for `LOOKUP`/`QUERY_PREFETCH_STATUS` (and related client/server state tracking) to use `request_id`, which could break older clients/servers and affects lookup/prefetch correctness under retries.
> 
> **Overview**
> Refactors the multiprocess lookup/prefetch flow to **stop leaking server-internal `job_id`s** and instead track/poll prefetch jobs purely by external `request_id`.
> 
> `LOOKUP` now returns `None` and the server stores `_prefetch_jobs` keyed by `request_id`; `QUERY_PREFETCH_STATUS`/`QUERY_PREFETCH_LOOKUP_HITS` payloads switch from `int` job IDs to `str` request IDs, with unknown requests returning `0` to avoid infinite polling. The vLLM scheduler adapter updates its lookup bookkeeping to cache results by `request_id` (supporting repeated checks after server pop) and adjusts cleanup accordingly, with tests updated to match the new protocol.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3c13a30f2e37f2df010a38b1b13736ccd36596a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


KuntaiDu (approved):
LGTM

lmcache/v1/multiprocess/server.py
```diff
@@ -643,74 +643,42 @@ def lookup(
...
-            job = self._prefetch_jobs.get(prefetch_job_id)
+            job = self._prefetch_jobs.get(request_id)
         if job is None:
             logger.warning(
-                "Prefetch job %d not found (already completed or invalid)",
-                prefetch_job_id,
+                "Prefetch job for request %s not found (already completed or invalid)",
+                request_id,
             )
             return None
```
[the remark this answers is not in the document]
yoo-kumaneko: This is a pre-existing issue. It's unrelated to this PR and should be fixed in another one.

lmcache/v1/multiprocess/server.py
```diff
@@ -643,74 +643,42 @@ def lookup(
...
-            )
-            return None
-
-        found_count = self.storage_manager.query_prefetch_lookup_hits(job.handle)
-        if found_count is None:
-            return None
-
-        found_count = found_count // job.world_size
-        return found_count
+            self._prefetch_jobs[job.request_id] = job
```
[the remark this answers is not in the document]
yoo-kumaneko: ignored

lmcache/v1/multiprocess/server.py
```diff
@@ -647,40 +643,37 @@ def lookup(
...
 
         Returns:
-            The number of hits for the prefetched keys if the lookup is all done.
-            None if the lookup phase is still in progress, or the prefetch is
-            already completed and consumed by query_prefetch_status, or the
-            prefetch_job_id is invalid.
+            The number of hits for the prefetched keys if the lookup phase is
+            done. None if the lookup phase is still in progress, or the prefetch
+            is already completed and consumed by query_prefetch_status, or the
+            request_id is invalid.
```
ApostaC: One potential problem here: 
Previously, since we have the job id, the caller should be able to avoid calling this function with an unknown job id.
However, since we changed to using an external request id, there could be a case where the caller calls this function with an unknown request id.  If we still return None here, it may cause the caller to spin on this request forever because it may interpret the returned `None` as "lookup is in progress".

Therefore, I suggest we return 0 if the request ID is not found. WDYT?

ApostaC (commented):
One potential problem in the `query_prefetch_status`'s return value. Please see the details below.

lmcache/v1/multiprocess/server.py
yoo-kumaneko: Agreed. We should never return None in this case. Let's return 0 and emit a warning.

yoo-kumaneko: updated

ApostaC (approved):
LGTM!

PR #2999, opened 2026-04-10
[BugFix] fix sha256_cbor + async_loading type mismatch in _hash_tokens

### LINKED ISSUE #2997: [Bug] Type mismatch when using sha256_cbor with enable_async_loading=True
**Label**
bug

**Summary**
When `pre_caching_hash_algorithm` is set to `sha256_cbor` and `enable_async_loading` is `True`, LMCache crashes with a type error during async lookup serialization.

**Details**
`sha256_cbor` (and `sha256_cbor_64bit`) returns `bytes`, but two places in the codebase declare `int` as the expected type:

1. `LookupRequestMsg.hashes: list[int]` in `lmcache/v1/lookup_client/async_lookup_message.py`  
2. `CacheEngineKey.chunk_hash: int` in `lmcache/utils.py`

When the async lookup client calls `process_tokens(..., make_key=False)`, it collects hash values (which are `bytes` for sha256_cbor) into `hashes: list[int]`. Then `msgspec.msgpack.encode(LookupRequestMsg(hashes=...))` raises an `OverflowError` or `TypeError` because msgpack cannot serialize `bytes` as an integer field.

Even if serialization were bypassed, the server-side path builds `CacheEngineKey` from `hashes=[int,...]` (after msgpack decode), while the retrieve-side path builds the key from raw `tokens` via sha256_cbor (returning `bytes`). Because `bytes != int`, `memory_obj_map.get(key)` in `_async_process_tokens_internal` always returns `None`, causing all cache lookups to miss.

**Steps / Reproduction**
1. Configure LMCache with:
   ```yaml
   pre_caching_hash_algorithm: sha256_cbor
   enable_async_loading: true
   ```
2. Run any inference request through vLLM with LMCacheConnectorV1
3. Observe `OverflowError: can't serialize ints < -2\*\*63 or > 2\*\*64 - 1` from `msgspec.msgpack.encode`

**Expected Outcome**
sha256_cbor should work correctly with async loading — cache hits should succeed.

**Actual Outcome**
`OverflowError` during serialization of the async lookup message, or silent 0% cache hit rate if the error is caught.

**Proposed Fix**
Change the type annotations to accept both `int` and `bytes`:
- `LookupRequestMsg.hashes: list[int]` → `list[Union[int, bytes]]`
- `CacheEngineKey.chunk_hash: int` → `Union[int, bytes]`

I am willing to submit a PR with the fix.

GentleCold:
## Summary

Fixes #2997

When `LMCACHE_PRE_CACHING_HASH_ALGORITHM=sha256_cbor` is combined with
`LMCACHE_ENABLE_ASYNC_LOADING=True`, LMCache raises:

```
OverflowError: can't serialize ints < -2**63 or > 2**64 - 1
```

**Root cause**: `sha256_cbor` (vLLM PR#23673) returns a 32-byte `bytes`
digest.  `_hash_tokens` returns this value directly, which then lands in
`LookupRequestMsg.hashes` — a field that msgpack encodes as integer.
A 256-bit value cannot fit in any integer msgpack type, causing the overflow.

**Fix**: one-line guard in `_hash_tokens` — after calling `hash_func`, if
the result is `bytes`, fold the first 8 bytes into a `uint64` int:

```python
result = self.hash_func((canon_prefix, canon_tokens, canon_extra))
if isinstance(result, bytes):
    result = int.from_bytes(result[:8], "big")
return result
```

Both the store path (`store(token_ids, ...)`) and the async lookup path
(`process_tokens(token_ids, make_key=False)`) go through `_hash_tokens`,
so the int keys they produce are always consistent with each other.
No type annotations elsewhere need to change.

## Testing

Added `test_hash_tokens_returns_int_for_bytes_hash_func` to
`tests/v1/test_token_database.py`: patches `hash_func` with a
bytes-returning sha256 stub and asserts that `_hash_tokens` returns a
`uint64` int.

End-to-end: ran `lmcache_bench.py` with sha256_cbor + async_loading; warm
phase shows 100% cache hits with no errors.

---

> [!NOTE]
> **Medium Risk**
> Changes cache key hashing behavior by folding `bytes` digests into `uint64` ints, which can affect cache key compatibility/hit rates for users on bytes-returning hash functions. Scope is small and guarded, but it touches core hashing used in cache lookup/serialization paths.
> 
> **Overview**
> Fixes a type/serialization mismatch when using bytes-returning hash algorithms (e.g. vLLM `sha256_cbor`) by **normalizing all hash outputs to `uint64` ints**.
> 
> `TokenDatabase` now folds `NONE_HASH` and `_hash_tokens()` results from `bytes` to an 8-byte `int`, ensuring prefix chaining and downstream msgpack/`CacheEngineKey` usage always sees plain integers. Adds a regression test asserting `process_tokens(..., make_key=False)` returns `int` hashes when `hash_func` returns `bytes`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 539a8eba5131666278daaa851939eb67359bc3ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -30,6 +30,32 @@ logger = init_logger(__name__)
 
 NONE_HASH = 0
 
+
+def _normalize_hash_to_int(hash_value: Union[int, bytes]) -> int:
+    """Normalize hash outputs to LMCache's int chunk-hash representation.
+
+    This function is triggered when vLLM's ``sha256_cbor`` hash function is
+    used because it returns a 32-byte digest. vLLM's
+    ``kv_cache_utils.init_none_hash`` can therefore initialize ``NONE_HASH`` as
+    bytes, and direct hash calls for token chunks can also return bytes.
+
+    LMCache stores chunk hashes in ``CacheEngineKey`` and serializes them with
+    msgpack, so byte digests must be folded into uint64-compatible ints before
+    they enter the prefix hash chain. This also keeps ``NONE_HASH`` and later
+    prefix hashes using the same structural type for CBOR hashing.
+
+    Args:
+        hash_value: Hash output from vLLM or Python's builtin hash.
+
+    Returns:
+        The original int hash value, or the first eight bytes of a digest as a
+        big-endian int.
+    """
+    if isinstance(hash_value, bytes):
+        return int.from_bytes(hash_value[:8], "big")
+    return hash_value
+
+
 # Type alias for process_tokens return value
 # (start_index, end_index, cache_engine_key｜hash)
 ProcessTokensResult = Tuple[int, int, Union[CacheEngineKey, int]]
@@ -70,11 +96,7 @@ class TokenDatabase(metaclass=abc.ABCMeta):
 
             if hasattr(kv_cache_utils, "init_none_hash"):
                 kv_cache_utils.init_none_hash(self.hash_func)
-                NONE_HASH = kv_cache_utils.NONE_HASH
-                # sha256_cbor returns bytes; fold to uint64 so the entire
-                # prefix chain uses a consistent int type.
-                if isinstance(NONE_HASH, bytes):
-                    NONE_HASH = int.from_bytes(NONE_HASH[:8], "big")
+                NONE_HASH = _normalize_hash_to_int(kv_cache_utils.NONE_HASH)
                 logger.info(
                     f"Initialized NONE_HASH={NONE_HASH} from vLLM (>= PR#20511)"
                 )
@@ -267,13 +289,9 @@ class TokenDatabase(metaclass=abc.ABCMeta):
             prefix_hash, tokens_tuple, extra_keys
         )
 
-        result = self.hash_func((canon_prefix, canon_tokens, canon_extra))
-        # sha256_cbor returns a 32-byte digest; fold it into a uint64 so that
-        # downstream code (msgpack serialisation, CacheEngineKey) always works
-        # with plain ints regardless of the configured hash algorithm.
-        if isinstance(result, bytes):
-            result = int.from_bytes(result[:8], "big")
-        return result
+        return _normalize_hash_to_int(
+            self.hash_func((canon_prefix, canon_tokens, canon_extra))
+        )
 
 
 class ChunkedTokenDatabase(TokenDatabase):
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -263,7 +263,13 @@ class TokenDatabase(metaclass=abc.ABCMeta):
             prefix_hash, tokens_tuple, extra_keys
         )
 
-        return self.hash_func((canon_prefix, canon_tokens, canon_extra))
+        result = self.hash_func((canon_prefix, canon_tokens, canon_extra))
+        # sha256_cbor returns a 32-byte digest; fold it into a uint64 so that
+        # downstream code (msgpack serialisation, CacheEngineKey) always works
+        # with plain ints regardless of the configured hash algorithm.
+        if isinstance(result, bytes):
+            result = int.from_bytes(result[:8], "big")
+        return result
 
 
 class ChunkedTokenDatabase(TokenDatabase):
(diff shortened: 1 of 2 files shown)


sammshen (approved):
LGTM!

GentleCold:
hi I need one more review, could I get merged? @hickeyma @ApostaC @deng451e

lmcache/v1/token_database.py
```diff
@@ -71,6 +71,10 @@ def __init__(
             if hasattr(kv_cache_utils, "init_none_hash"):
                 kv_cache_utils.init_none_hash(self.hash_func)
                 NONE_HASH = kv_cache_utils.NONE_HASH
+                # sha256_cbor returns bytes; fold to uint64 so the entire
+                # prefix chain uses a consistent int type.
+                if isinstance(NONE_HASH, bytes):
```
KuntaiDu: I would prefer wrapping this logic into an external function, and in the docstring of that function document why this function will be triggered and why this is necessary.

KuntaiDu (commented):
See comment.

GentleCold:
Would you like to take a look at this PR? Thanks! @hickeyma @ApostaC @deng451e @KuntaiDu  @hlin99 @chunxiaozheng

lmcache/v1/token_database.py
```diff
@@ -70,7 +96,7 @@ def __init__(
 
             if hasattr(kv_cache_utils, "init_none_hash"):
                 kv_cache_utils.init_none_hash(self.hash_func)
-                NONE_HASH = kv_cache_utils.NONE_HASH
```
chunxiaozheng, line 73: @GentleCold Thanks for your contribution, a minor question, the `NONE_HASH` is also a `bytes`, why not   occur errors before?

GentleCold: Good question. `NONE_HASH` itself did not trigger the serialization error because it is only used as the initial prefix-hash seed and is not sent through the async lookup message. The actual serialized values are the chunk hashes returned by `_hash_tokens`, and with `sha256_cbor` those are bytes.

I still normalize `NONE_HASH` because otherwise the prefix hash chain would mix input types: the first hash input would use a bytes prefix, while later hash inputs use normalized int prefixes. For CBOR-based hashing, bytes and int are encoded differently, so normalizing `NONE_HASH` keeps the chain structurally consistent and keeps LMCache’s public chunk-hash representation as int throughout. @chunxiaozheng

chunxiaozheng (approved):
LGTM!

lmcache/v1/token_database.py
chunxiaozheng: @GentleCold the `NONE_HASH` is also used in sync mode, `NONE_HASH` is just an intermediate calculated variable, and the final error is due to the fact that in `msgspec.msgpack.encode`, `int` cannot exceed 8 bytes, but the result of `sha256_cbor` exceed 8 bytes.

So, even if we don't handle `NONE_HASH`, I don't think there will be any problems, but if we handle it, it can still be done. In a word, thans for fixing it!

GentleCold:
Could you please re-enable auto-merge or add the full label, and approve the pending GitHub Actions workflows for this fork PR? Thank you! @chunxiaozheng

chunxiaozheng:
> Could you please re-enable auto-merge or add the full label, and approve the pending GitHub Actions workflows for this fork PR? Thank you! @chunxiaozheng

@GentleCold My bad, I think I've already clicked it. 😭

chunxiaozheng:
@GentleCold the `Code Quality Check` failed, could you help fix it?

GentleCold:
hi I fix it, could you please re-add the full label? Thanks! @chunxiaozheng

GentleCold:
hi, it seems the ci might be stuck? @chunxiaozheng

chunxiaozheng:
> hi, it seems the ci might be stuck? @chunxiaozheng

yeah, I have retried.

chunxiaozheng:
@GentleCold This PR has merged! Thanks for your contribution again, and we also look forward to the future contributions!

PR #3006, opened 2026-04-12
[BugFix] Fix deadlock + silent data corruption in storage backends (fixes #2942, fixes #3022)

### LINKED ISSUE #2942: [Bug]: Deadlock in LocalCPUBackend.allocate() when use_hot=False and staging buffer is full
**Label**
bug

**Describe the bug**

When `local_cpu=False` (i.e., `use_hot=False`), `LocalCPUBackend.allocate(busy_loop=True)` enters an infinite busy-wait loop if the CPU staging buffer is full. The eviction path is entirely gated by `if self.use_hot:`, so when `use_hot=False`, no eviction is ever attempted — the loop just calls `time.sleep(0.1)` and retries `memory_allocator.allocate()` forever.

This is triggered in production by `LocalDiskBackend.load_bytes_from_disk()` (`local_disk_backend.py:579`), which calls `allocate()` with the default `busy_loop=True`. The symptom is the engine becoming completely unresponsive with this warning flooding the logs:

```
LMCache WARNING: No eviction candidates found in local cpu backend.
Local cpu memory is under pressure. Waiting for 0.1 seconds before retrying.
```

**Configuration that triggers this:**
```yaml
chunk_size: 256
local_cpu: false
max_local_cpu_size: 5        # GB — staging buffer for disk offload
local_disk: "file:///data/lmcache/"
max_local_disk_size: 200
```

Per the documentation, `max_local_cpu_size` should be set even when `local_cpu=false`, as the CPU memory pool serves as a staging buffer for disk I/O. This is a valid and intended configuration.

**Root Cause Analysis**

The core issue is in `LocalCPUBackend.allocate()` (`local_cpu_backend.py:533-580`):

```python
while True:
    wait_other_requests = True
    if self.use_hot:              # ← skipped when use_hot=False
        evict_keys = self.cache_policy.get_evict_candidates(...)
        if evict_keys:
            wait_other_requests = False
            self.batched_remove(evict_keys, force=False)

    if wait_other_requests:       # ← always True when use_hot=False
        if not busy_loop:
            break                 # busy_loop=False: safe, returns None
        time.sleep(0.1)           # busy_loop=True: spins forever!

    memory_obj = self.memory_allocator.allocate(...)
    if memory_obj is not None:
        break
```

The same pattern exists in `batched_allocate()` (`local_cpu_backend.py:643-706`).

**Contributing factors that form a deadlock chain:**

1. **No active eviction path** — When `use_hot=False`, `hot_cache` is empty (because `batched_submit_put_task` returns early at line 193-194). The eviction logic only looks at `hot_cache`, so there is nothing to evict even though the staging buffer is full of in-flight objects.

2. **Staging buffer occupied by in-flight put tasks** — `submit_put_task()` calls `ref_count_up()` (line 344) before submitting an async disk write. The buffer is held until `async_save_bytes_to_disk()` completes and calls `ref_count_down()` (line 526). Slow disk I/O prolongs buffer occupation.

3. **Put tasks starved by priority scheduling** — `async_save_bytes_to_disk` is submitted with priority 2 (lowest), while prefetch tasks have priority 0 (highest). Under load, put completions (and their `ref_count_down()` calls) are delayed.

4. **New stores fail silently** — `allocate_and_copy_objects()` (storage_manager.py:89-94) uses `busy_loop=False`. When the buffer is full, store allocation returns `None` and breaks — no new data is written, so no new `ref_count_down()` calls will free buffer space.

5. **Cascading failure:**
```
Executor busy (prefetch priority > put)
  → put tasks delayed → ref_count_down() delayed
  → staging buffer stays full
  → new put allocations fail (busy_loop=False, break)
  → no new writes → no new buffer releases
  → get_blocking's busy_loop=True spins forever
  → engine thread blocked → all requests stall
```

**Why `use_hot=True` is not affected:**

With `use_hot=True`, `submit_put_task()` inserts `memory_obj` into `hot_cache` (line 159). When `allocate()` needs space, it synchronously evicts LRU entries from `hot_cache` via `batched_remove()` — this is a local, lock-protected operation that does not depend on the async executor, disk I/O completion, or any external scheduling. It breaks the entire deadlock chain.

**To Reproduce**

```python
import threading
import torch

from lmcache.utils import CacheEngineKey
from lmcache.v1.memory_management import MemoryFormat, MixedMemoryAllocator
from lmcache.v1.metadata import LMCacheMetadata
from lmcache.v1.pin_monitor import PinMonitor
from lmcache.v1.storage_backend.local_cpu_backend import LocalCPUBackend

CHUNK_SHAPE = torch.Size([2, 256, 8, 128])  # ~1 MB at bf16
CHUNK_DTYPE = torch.bfloat16
CHUNK_FMT = MemoryFormat.KV_2LTD
POOL_BYTES = 3 * 1024 * 1024  # 3 MB, fits ~3 chunks


def make_backend(use_hot: bool):
    config = LMCacheEngineConfig.from_defaults(
        chunk_size=256, local_cpu=use_hot, lmcache_instance_id="deadlock_test",
    )
    metadata = LMCacheMetadata(
        model_name="test", world_size=1, local_world_size=1,
        worker_id=0, local_worker_id=0,
        kv_dtype=CHUNK_DTYPE, kv_shape=(4, 2, 256, 8, 128),
    )
    PinMonitor.GetOrCreate(config)
    allocator = MixedMemoryAllocator(POOL_BYTES)
    backend = LocalCPUBackend(
        config=config, metadata=metadata,
        memory_allocator=allocator, dst_device="cpu",
    )
    return backend, allocator


def test_scenario(use_hot: bool):
    backend, allocator = make_backend(use_hot)
    try:
        # Fill pool
        held = []
        for i in range(100):
            obj = backend.allocate(CHUNK_SHAPE, CHUNK_DTYPE, CHUNK_FMT,
                                   eviction=False, busy_loop=False)
            if obj is None:
                break
            if use_hot:
                key = CacheEngineKey(model_name="test", world_size=1,
                                     worker_id=0, chunk_hash=i, dtype=CHUNK_DTYPE)
                backend.submit_put_task(key, obj)
                obj.ref_count_down()
            held.append(obj)

        # Try allocate like load_bytes_from_disk does
        result_box = [None]
        def try_alloc():
            result_box[0] = backend.allocate(CHUNK_SHAPE, CHUNK_DTYPE, CHUNK_FMT,
                                              eviction=True, busy_loop=True)
        t = threading.Thread(target=try_alloc, daemon=True)
        t.start()
        t.join(timeout=3.0)

        label = f"use_hot={use_hot}"
        if t.is_alive():
            print(f"  {label}: DEADLOCK (thread stuck after 3s)")
        else:
            print(f"  {label}: OK (eviction worked)")
            if result_box[0] is not None:
                result_box[0].ref_count_down()

        if not use_hot:
            for obj in held:
                obj.ref_count_down()
    finally:
        allocator.close()
        PinMonitor.DestroyInstance()


if __name__ == "__main__":
    test_scenario(use_hot=False)  # DEADLOCK
    test_scenario(use_hot=True)   # OK
```

**Expected output:**
```
  use_hot=False: OK
  use_hot=True: OK
```

**Actual output:**
```
  use_hot=False: DEADLOCK (thread stuck after 3s)
  use_hot=True: OK (eviction worked)
```

**Expected behavior**

When `use_hot=False` and the staging buffer is full, `allocate()` should not spin indefinitely. It should either:
- Return `None` after a bounded number of retries (and callers like `load_bytes_from_disk` should handle this gracefully instead of asserting), or
- Track staging buffer objects in an evictable structure so the eviction path works regardless of `use_hot`, or
- Force `busy_loop=False` when `use_hot=False`, since there is no active reclamation mechanism to make busy-looping productive.

**Affected code:**
- `lmcache/v1/storage_backend/local_cpu_backend.py`: `allocate()` L533-580, `batched_allocate()` L643-706
- `lmcache/v1/storage_backend/local_disk_backend.py`: `load_bytes_from_disk()` L579 (calls allocate with default `busy_loop=True`), `get_blocking()` L404

**Environment:**
- LMCache version: v0.4.2 (commit 996da472)
- OS: Linux
- Context length: 80K tokens
- Model: serving with vLLM

**Additional context**

The design comment in `allocate()` (lines 503-513) states:
> *busy_loop should only be used for retrieve ... one retrieve at a time (okay to busy loop because stores will clear)*

This assumption holds when `use_hot=True` (eviction can synchronously free space), but breaks completely when `use_hot=False` — stores may not clear because they also fail to allocate, and there is no eviction mechanism to break the cycle.

### LINKED ISSUE #3022: [Bug]: batched_get_non_blocking returns unloaded MemoryObjs on partial alloc failure  silent data corruption
## What's broken?

When `batched_get_non_blocking()` partially fails mid-batch (CPU staging pool exhausted at key N), the already-allocated `MemoryObj`s for keys 0..N-1 are returned to callers **without loading any data from disk**. The `disk_worker.submit_task` (which calls `batched_async_load_bytes_from_disk`  `read_file`) is completely skipped. Callers receive objects containing stale/uninitialized pool memory and treat them as valid KV cache hits, causing **silent model inference corruption**.

## Who is affected?

Any configuration using the disk backend under memory pressure:
- `local_disk: "file:///..."` with a CPU staging buffer (`max_local_cpu_size`) that can become full
- Triggered when prefetch load exceeds staging buffer capacity
- Both `local_cpu: true` and `local_cpu: false` configurations

## When does it trigger?

1. Multiple keys are requested via `batched_get_non_blocking`
2. CPU staging pool becomes exhausted mid-batch (allocation succeeds for first N keys, fails at key N+1)
3. The early return at L476 fires, returning allocated-but-unloaded objects

This became reachable after PR #2611 (`b7d025a0`, 2026-03-25) changed `busy_loop` from `True` to `False` in this method. Before that, allocation would block forever or succeed  the early return was unreachable.

## Where is the bug?

`lmcache/v1/storage_backend/local_disk_backend.py`, `batched_get_non_blocking()` L468-476:

```python
if memory_obj is None:
    logger.error(...)
    self.disk_lock.release()
    return mem_objs  #  these objects have NO data loaded from disk!
```

The normal path (L490-496) calls `disk_worker.submit_task("prefetch", self.batched_async_load_bytes_from_disk, ...)` which reads actual disk data into the memory buffers. The early return skips this entirely.

## Why does it happen?

Three cascading issues:

1. **Silent data corruption**: `MemoryObj.byte_array` contains whatever was in the pool memory region (not zeroed on allocation). Downstream code (`prefetch_all_done_callback`  `_async_process_tokens_internal`  `gpu_connector.batched_to_gpu`) treats these as valid cache hits and sends stale data to GPU.

2. **DiskCacheMetadata pin leak**: `self.dict[key].pin()` (L478) is called for each successful key, but the matching `unpin()` in `batched_async_load_bytes_from_disk` (L589) is never reached. Entries become permanently unevictable.

3. **MemoryObj pin leak**: `memory_obj.pin()` (L486) is called with no matching `unpin()` on the early return path.

## How did we fix it?

The early return now undoes all side effects and returns an empty list:

```python
if memory_obj is None:
    logger.error(...)
    for prev_obj in mem_objs:
        prev_obj.unpin()
        prev_obj.ref_count_down()
    for prev_key in keys[:len(mem_objs)]:
        if prev_key in self.dict:
            self.dict[prev_key].unpin()
    self.disk_lock.release()
    return []
```

Returning `[]` instead of partial unloaded objects ensures no stale data reaches GPU, and all resource leaks are cleaned up.

Fix is in PR #3006 (commit `dda46af`), discovered during review by @DongDongJu.

## How do we know it works?

Existing regression test `test_batched_get_non_blocking_releases_lock_on_alloc_failure` verifies `result == []` and `pin_count == 0` after allocation failure. Additional multi-key partial failure test to be added.

ianliuy:
## Summary

Fixes #2942 (deadlock when `use_hot=False` and staging buffer full)
Fixes #3022 (partial alloc failure returns unloaded MemoryObjs  silent data corruption, found by @DongDongJu during review)

### Root Cause
`LocalCPUBackend.allocate()` and `batched_allocate()` enter a `while True:` loop whose only exit (other than success) is inside `if self.use_hot:`. When `use_hot=False`, eviction is structurally impossible  `hot_cache` is always empty. With `busy_loop=True`, the loop spins forever.

### Changes in `local_cpu_backend.py`
Added guard before `while True:` in both `allocate()` and `batched_allocate()`:
```python
if not self.use_hot:
    busy_loop = False
```
Zero impact on `use_hot=True` path.

### Changes in `local_disk_backend.py`
- `load_bytes_from_disk()`: replaced `assert memory_obj is not None` with graceful `return None` + warning log
- `batched_get_non_blocking()`: fixed pre-existing `disk_lock` leak on early return

### Root Cause
`batched_get_non_blocking()` returns `mem_objs` on allocation failure mid-batch, but `disk_worker.submit_task` (which reads data from disk into the memory buffers) is completely skipped. The returned `MemoryObj`s contain stale pool memory, not valid KV cache data. No layer in the call chain validates data content  stale memory flows directly to GPU via `batched_to_gpu()`.

Additionally: `DiskCacheMetadata.pin()` and `MemoryObj.pin()` were called but never undone on this path.

### Changes in `local_disk_backend.py`
Early return now cleans up all side effects in a `try/finally` block:
```python
if memory_obj is None:
    try:
        for prev_obj in mem_objs:
            prev_obj.unpin()
            prev_obj.ref_count_down()
        for prev_key in keys[:len(mem_objs)]:
            if prev_key in self.dict:
                self.dict[prev_key].unpin()
    finally:
        self.disk_lock.release()
    return []
```

## Testing

9 regression tests added across both files:
- `test_allocate_no_deadlock_use_hot_false`  core deadlock prevention
- `test_load_bytes_from_disk_returns_none_on_pool_exhaustion`  graceful None
- `test_batched_get_non_blocking_releases_lock_on_alloc_failure`  lock release
- `test_get_blocking_returns_none_on_pool_exhaustion`  sync path
- `test_partial_alloc_cleanup_on_multi_key_failure`  multi-key rollback (#3022)
- `test_partial_alloc_no_stale_data_returned`  **critical**: no stale objects (#3022)
- `test_first_key_alloc_failure_multi_key_batch`  edge case
- `test_last_key_in_large_batch_fails`  5-key edge case
- `test_all_keys_succeed_returns_full_batch`  happy path baseline

All `pre-commit` hooks pass (ruff, mypy, isort, codespell, SPDX).

## Follow-up Issues
- #3015: `get_blocking()` records phantom cache hit when `load_bytes_from_disk()` fails

---

> [!NOTE]
> **Medium Risk**
> Touches core storage/memory-allocation paths and changes failure semantics (returning `None`/empty batch and rolling back pins), which could impact cache hit behavior under memory pressure despite being well-covered by new regression tests.
> 
> **Overview**
> Prevents a potential infinite busy-loop in `LocalCPUBackend.allocate()`/`batched_allocate()` when `use_hot=False` by forcing `busy_loop=False` in that configuration so allocation fails fast with `None` instead of hanging.
> 
> Hardens `LocalDiskBackend` disk reads/prefetch when the CPU staging pool is exhausted: `load_bytes_from_disk()` now returns `None` with a warning (instead of asserting), and `batched_get_non_blocking()` now drops the entire batch on mid-batch allocation failure while rolling back pins/refcounts, releasing `disk_lock`, and deferring cache-policy hit updates until the whole batch is successfully staged to avoid phantom hits and stale/uninitialized data exposure.
> 
> Adds regression tests covering the deadlock case, lock-release behavior, synchronous/async pool-exhaustion handling, and multi-key partial-allocation rollback to prevent silent data corruption.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5f719182582650a18b63e7bdc0c077d1479f040. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -463,13 +463,15 @@ class LocalDiskBackend(StorageBackendInterface):
                 # NOT run, so any returned mem_objs would contain stale
                 # memory — not valid KV cache data.  Clean up and return
                 # empty to avoid silent data corruption.
-                for prev_obj in mem_objs:
-                    prev_obj.unpin()
-                    prev_obj.ref_count_down()
-                for prev_key in keys[: len(mem_objs)]:
-                    if prev_key in self.dict:
-                        self.dict[prev_key].unpin()
-                self.disk_lock.release()
+                try:
+                    for prev_obj in mem_objs:
+                        prev_obj.unpin()
+                        prev_obj.ref_count_down()
+                    for prev_key in keys[: len(mem_objs)]:
+                        if prev_key in self.dict:
+                            self.dict[prev_key].unpin()
+                finally:
+                    self.disk_lock.release()
                 return []
 
             self.dict[key].pin()
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -589,7 +589,7 @@ class LocalDiskBackend(StorageBackendInterface):
         Load bytearray from disk.
         """
 
-        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt, busy_loop=False)
+        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt)
         if memory_obj is None:
             logger.warning(
                 "LocalDiskBackend: CPU staging pool exhausted while loading "
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -589,7 +589,13 @@ class LocalDiskBackend(StorageBackendInterface):
         """
 
         memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt)
-        assert memory_obj is not None, "Memory allocation failed during disk load."
+        if memory_obj is None:
+            logger.warning(
+                "LocalDiskBackend: CPU staging pool exhausted while loading "
+                "key %s from disk. Returning cache miss.",
+                key,
+            )
+            return None
 
         buffer = memory_obj.byte_array
         self.read_file(key, buffer, path)
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -427,6 +427,7 @@ class LocalDiskBackend(StorageBackendInterface):
     ) -> list[MemoryObj]:
         mem_objs: list[MemoryObj] = []
         paths: list[str] = []
+        hit_keys: list[CacheEngineKey] = []
 
         logger.debug(f"lookup_id: {lookup_id}; Prefetching {len(keys)} keys from disk.")
         for key in keys:
@@ -475,16 +476,20 @@ class LocalDiskBackend(StorageBackendInterface):
                 return []
 
             self.dict[key].pin()
-
-            # NOTE(Jiayi): Currently, we consider prefetch as cache hit.
-            # Update cache recency
-            self.cache_policy.update_on_hit(key, self.dict)
+            hit_keys.append(key)
 
             self.disk_lock.release()
             logger.debug(f"Prefetching {key} from disk.")
             memory_obj.pin()
             mem_objs.append(memory_obj)
             paths.append(path)
+
+        # NOTE(Jiayi): Currently, we consider prefetch as cache hit.
+        # Only record those hits after the whole batch is staged to avoid
+        # phantom cache-policy updates on partial-allocation failure.
+        with self.disk_lock:
+            for key in hit_keys:
+                self.cache_policy.update_on_hit(key, self.dict)
 
         return await self.disk_worker.submit_task(
             "prefetch",
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -458,6 +458,7 @@ class LocalDiskBackend(StorageBackendInterface):
                     "a previous retrieve). Returning partial results.",
                     key,
                 )
+                self.disk_lock.release()
                 return mem_objs
 
             self.dict[key].pin()
@@ -588,7 +589,7 @@ class LocalDiskBackend(StorageBackendInterface):
         Load bytearray from disk.
         """
 
-        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt)
+        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt, busy_loop=False)
         if memory_obj is None:
             logger.warning(
                 "LocalDiskBackend: CPU staging pool exhausted while loading "
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -455,11 +455,22 @@ class LocalDiskBackend(StorageBackendInterface):
                 logger.error(
                     "Memory allocation failed during async disk load for key %s. "
                     "CPU staging pool may be exhausted (unpin() not called after "
-                    "a previous retrieve). Returning partial results.",
+                    "a previous retrieve). Dropping all partial results.",
                     key,
                 )
+                # Undo side effects for already-processed keys: the
+                # disk_worker.submit_task (which loads data from disk) will
+                # NOT run, so any returned mem_objs would contain stale
+                # memory — not valid KV cache data.  Clean up and return
+                # empty to avoid silent data corruption.
+                for prev_obj in mem_objs:
+                    prev_obj.unpin()
+                    prev_obj.ref_count_down()
+                for prev_key in keys[: len(mem_objs)]:
+                    if prev_key in self.dict:
+                        self.dict[prev_key].unpin()
                 self.disk_lock.release()
-                return mem_objs
+                return []
 
             self.dict[key].pin()
 
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -603,7 +603,7 @@ class LocalDiskBackend(StorageBackendInterface):
         Load bytearray from disk.
         """
 
-        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt, busy_loop=False)
+        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt)
         if memory_obj is None:
             logger.warning(
                 "LocalDiskBackend: CPU staging pool exhausted while loading "
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -477,13 +477,15 @@ class LocalDiskBackend(StorageBackendInterface):
                 # NOT run, so any returned mem_objs would contain stale
                 # memory — not valid KV cache data.  Clean up and return
                 # empty to avoid silent data corruption.
-                for prev_obj in mem_objs:
-                    prev_obj.unpin()
-                    prev_obj.ref_count_down()
-                for prev_key in keys[: len(mem_objs)]:
-                    if prev_key in self.dict:
-                        self.dict[prev_key].unpin()
-                self.disk_lock.release()
+                try:
+                    for prev_obj in mem_objs:
+                        prev_obj.unpin()
+                        prev_obj.ref_count_down()
+                    for prev_key in keys[: len(mem_objs)]:
+                        if prev_key in self.dict:
+                            self.dict[prev_key].unpin()
+                finally:
+                    self.disk_lock.release()
                 return []
 
             self.dict[key].pin()
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -472,6 +472,7 @@ class LocalDiskBackend(StorageBackendInterface):
                     "a previous retrieve). Returning partial results.",
                     key,
                 )
+                self.disk_lock.release()
                 return mem_objs
 
             self.dict[key].pin()
@@ -602,7 +603,7 @@ class LocalDiskBackend(StorageBackendInterface):
         Load bytearray from disk.
         """
 
-        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt)
+        memory_obj = self.local_cpu_backend.allocate(shape, dtype, fmt, busy_loop=False)
         if memory_obj is None:
             logger.warning(
                 "LocalDiskBackend: CPU staging pool exhausted while loading "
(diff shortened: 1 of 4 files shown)


ianliuy:
Good catch  you're absolutely right. The explicit `busy_loop=False` at the callsite regresses the `use_hot=True` path by causing false cache misses under transient memory pressure.

**Reverted in commit `9b0a45c5`**: `load_bytes_from_disk()` now uses the default `busy_loop=True` again. The internal guard inside `allocate()` (`if not self.use_hot: busy_loop = False`) already handles the `use_hot=False` case correctly, so the callsite override was both redundant and harmful.

Note: the `busy_loop=False` in `batched_get_non_blocking()` (L461) is intentionally kept  that runs on the async event loop thread where blocking would deadlock the entire loop.

ianliuy:
Added 4 regression tests in commit `9b0a45c5`:

| Test | File | What it covers |
|------|------|----------------|
| `test_allocate_no_deadlock_use_hot_false` | `test_local_cpu_backend.py` | Core bug: `allocate(busy_loop=True)` returns `None` within bounded time when `use_hot=False` + pool exhausted (not deadlock) |
| `test_load_bytes_from_disk_returns_none_on_pool_exhaustion` | `test_local_disk_backend.py` | Graceful `None` return + warning log instead of `assert` crash |
| `test_batched_get_non_blocking_releases_lock_on_alloc_failure` | `test_local_disk_backend.py` | `disk_lock` properly released on early return (no lock leak) |
| `test_get_blocking_returns_none_on_pool_exhaustion` | `test_local_disk_backend.py` | Synchronous retrieval path returns `None` when staging pool full |

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -472,6 +472,7 @@ async def batched_get_non_blocking(
                     "a previous retrieve). Returning partial results.",
                     key,
                 )
+                self.disk_lock.release()
```
DongDongJu: What happened if we got partial success for multiple key at here?

DongDongJu (commented):
Hello @ianliuy, I left one nit comments.

lmcache/v1/storage_backend/local_disk_backend.py
ianliuy: Great catch  you're right, partial results here are **not safe**.

The early return at L476 returns `mem_objs` that were allocated and pinned, but `disk_worker.submit_task` (L490, which calls `batched_async_load_bytes_from_disk`  `read_file`) is completely skipped. The returned `MemoryObj`s contain stale pool memory, not valid KV cache data. Callers treat them as cache hits  silent data corruption.

Additionally, `self.dict[key].pin()` (L478) and `memory_obj.pin()` (L486) were called but never undone  permanent pin leaks.

Fixed in `dda46af`: the early return now undoes all side effects (unpin + ref_count_down for already-processed objects) and returns `[]` instead of partial unloaded objects.

```python
if memory_obj is None:
    for prev_obj in mem_objs:
        prev_obj.unpin()
        prev_obj.ref_count_down()
    for prev_key in keys[:len(mem_objs)]:
        if prev_key in self.dict:
            self.dict[prev_key].unpin()
    self.disk_lock.release()
    return []
```

This is a pre-existing bug (the early return with `busy_loop=False` was introduced in PR #2611), but fixing it here since we're already on this code path. Thanks for pointing this out!

PR #3007, opened 2026-04-12
Fix LocalDiskBackend.close() to drain async writes before disk cleanup

### LINKED ISSUE #2840: LocalDiskBackend: clean up or persist on-disk cache files on shutdown

ianliuy:

`LocalDiskBackend` stores KV cache chunks as `.pt` files under `self.path`, but the key-to-file mapping (`self.dict`) lives in memory only. On every graceful shutdown, `close()` stopped the worker executor without deleting any files  leaving all cached `.pt` files orphaned and silently leaking disk space (issue #2840).

### Root cause

The original `close()` had no file cleanup at all. PR #2841 added cleanup, but in the wrong order: it called `batched_remove()` before `disk_worker.close()`. The disk worker runs up to 4 async threads that may still be executing `put` tasks at that point  each in-flight put writes a new file and adds a new entry to `self.dict` *after* the key snapshot is taken, so those files are never deleted.

### This fix

1. **Drain first**  call `disk_worker.close()` (`executor.shutdown(wait=True)`) before snapshotting keys, ensuring `self.dict` is fully up-to-date.
2. **Thread-safe cleanup**  acquire `disk_lock`, snapshot `dict.items()`, `dict.clear()` atomically, then delete files outside the lock. Prevents races with concurrent `get_blocking()` calls.
3. **Per-file error handling**  each `os.remove()` is wrapped in `try/except OSError` so a single missing or locked file cannot abort the rest of the cleanup.
4. **Exception safety**  `batched_msg_sender.close()` and `disk_worker.close()` are each wrapped in `try/except Exception`; `batched_msg_sender = None` is in a `finally` block so `close()` is properly idempotent even if the sender's `close()` raises.
5. **Logging**  INFO log with count of files deleted; WARNING per file removal failure.

**Limitation**: graceful shutdown only. On a hard crash (SIGKILL, OOM), files still orphan. Full cross-restart recovery via a persistent index is tracked separately in PR #2734, which is complementary to this fix.

## Special notes for reviewers

- This PR supersedes PR #2841 (same goal, fixes the ordering race, adds tests, adds error handling). A comment has been left on that PR explaining the issue.
- PR #2734 (persistent index / crash recovery) is orthogonal  both PRs should eventually merge.
- All pre-commit hooks pass (isort, ruff, ruff-format, mypy, codespell).

## Tests

New class `TestLocalDiskBackendClose` with 8 cases:

| Test | Verifies |
|------|----------|
| `test_close_deletes_tracked_files` | 3 injected files deleted, `dict` empty |
| `test_close_preserves_untracked_files` | File not in `dict` untouched |
| `test_close_missing_file_no_raise` | Dict entry with non-existent path  no exception |
| `test_close_idempotent` | Two `close()` calls  no exception |
| `test_close_empty_backend` | Empty `dict`  `close()` is safe no-op |
| `test_close_drains_disk_worker_before_file_removal` | Records call order; asserts `disk_worker.close()` precedes `os.remove()`  **would have caught PR #2841's bug** |
| `test_close_calls_batched_msg_sender_close` | Mock sender's `close()` called once, attribute set to `None` |
| `test_close_sender_raises_still_cleared` | Sender `close()` raises `RuntimeError`  attribute still set to `None` (verifies `finally` block) |

- [x] This PR contains unit tests

Fixes #2840

---

> [!NOTE]
> **Medium Risk**
> Touches shutdown and file-deletion logic in a storage backend; mistakes could leak disk or delete the wrong files, but deletion is limited to tracked paths and is covered by targeted regression tests.
> 
> **Overview**
> Fixes graceful shutdown of `LocalDiskBackend` to **prevent orphaned on-disk `.pt` cache files** and ordering races with in-flight async `put` operations.
> 
> `close()` now drains `disk_worker` *before* closing the controller sender and snapshotting `dict`, then clears cache bookkeeping under `disk_lock`, deletes each tracked file with per-file error handling, and logs deletion stats; sender shutdown is made exception-safe and idempotent.
> 
> Adds a new `TestLocalDiskBackendClose` suite covering deletion behavior, call ordering (drain-before-remove), sender closing semantics, idempotency, missing files, and metrics/usage reset (including a Windows-friendly `os.statvfs` mock).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb73cc2267180cb5dab68a39db6ce59f271d97a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -670,13 +670,31 @@ class LocalDiskBackend(StorageBackendInterface):
         Ordering is important: ``disk_worker.close()`` must be called before
         the file cleanup so that any in-flight async ``put`` tasks complete and
         all entries are present in ``self.dict`` before we snapshot the keys.
+
+        Note: this only handles graceful shutdown.  On a hard crash (SIGKILL,
+        OOM) files will still be orphaned; cross-restart recovery requires a
+        separate persistent index (see issue #1175 / PR #2734).
         """
-        if self.batched_msg_sender is not None:
-            self.batched_msg_sender.close()
-            self.batched_msg_sender = None
+        try:
+            if self.batched_msg_sender is not None:
+                self.batched_msg_sender.close()
+                self.batched_msg_sender = None
+        except Exception:
+            logger.warning(
+                "LocalDiskBackend.close(): error closing batched_msg_sender",
+                exc_info=True,
+            )
 
         # Drain all queued async disk I/O before snapshotting keys.
-        self.disk_worker.close()
+        # This MUST complete before we iterate self.dict so that any in-flight
+        # put tasks finish writing files and inserting entries.
+        try:
+            self.disk_worker.close()
+        except Exception:
+            logger.warning(
+                "LocalDiskBackend.close(): error draining disk_worker",
+                exc_info=True,
+            )
 
         deleted = 0
         keys = list(self.dict.keys())
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -668,13 +668,25 @@ class LocalDiskBackend(StorageBackendInterface):
         """Shut down the backend and delete all tracked on-disk cache files.
 
         Ordering is important: ``disk_worker.close()`` must be called before
-        the file cleanup so that any in-flight async ``put`` tasks complete and
-        all entries are present in ``self.dict`` before we snapshot the keys.
+        the controller sender is closed or file cleanup begins so that any
+        in-flight async ``put`` tasks complete and all entries are present in
+        ``self.dict`` before we snapshot the keys.
 
         Note: this only handles graceful shutdown.  On a hard crash (SIGKILL,
         OOM) files will still be orphaned; cross-restart recovery requires a
         separate persistent index (see issue #1175 / PR #2734).
         """
+        # Drain all queued async disk I/O before closing the sender or
+        # snapshotting keys. This MUST complete first so that in-flight put
+        # tasks can still report admissions without racing on a cleared sender.
+        try:
+            self.disk_worker.close()
+        except Exception:
+            logger.warning(
+                "LocalDiskBackend.close(): error draining disk_worker",
+                exc_info=True,
+            )
+
         if self.batched_msg_sender is not None:
             try:
                 self.batched_msg_sender.close()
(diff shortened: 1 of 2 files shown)


ianliuy:
## Addressed review feedback in commit b7ca225

Thanks to both @gemini-code-assist and @cursor for the thorough reviews! Here's what I changed:

### Cursor Bugbot  `batched_msg_sender = None` in `finally`  Adopted

Good catch. Moved the `None` assignment into a `finally` block so it's always cleared even if `close()` raises, making `close()` properly idempotent:

```python
try:
    self.batched_msg_sender.close()
except Exception:
    logger.warning(...)
finally:
    self.batched_msg_sender = None  # always runs
```

Also added `test_close_sender_raises_still_cleared` (8th test) to directly verify this behavior.

### Gemini Code Assist  `disk_lock` + `dict.clear()`  Partially adopted

Adopted the `disk_lock` + `dict.clear()` pattern  `get_blocking()` can race with cleanup since inference threads aren't stopped by `close()`, and this is consistent with how `self.dict` is accessed elsewhere:

```python
with self.disk_lock:
    items = list(self.dict.items())
    self.dict.clear()

for _, meta in items:
    os.remove(meta.path)
```

**Skipped metrics resets** (`self.usage = 0`, `self.current_cache_size = 0`, `stats_monitor.update_local_storage_usage(0)`)  `close()` is a terminal operation and no other `close()` in the codebase resets metrics at shutdown. The process is going away; nobody will read these values.

Note: the lock narrows but doesn't fully close the race  a `get_blocking()` thread that already extracted `meta.path` before the lock can still hit `FileNotFoundError`. Fully fixing that requires a `_closed` flag in `get_blocking()`, which is a separate concern.

### Also rebased onto latest `dev` (3 new upstream commits).

sammshen (approved):
LGTM

ianliuy:
Hi @ApostaC and @DongDongJu, just a polite follow-up.

I fixed the DCO sign-off issue and the refreshed checks are now green. This PR previously had LGTM from @sammshen; would you be willing to take another look for merge when you have a chance?

PR #3020, opened 2026-04-13
[MP][optimize] unified touch all keys in end session request 

chunxiaozheng:
> [!NOTE]
> **Medium Risk**
> Changes L1 eviction recency tracking by introducing an explicit `on_l1_keys_accessed` event and by touching keys only when read locks fully release, which can alter eviction order and cache hit rates across MP requests.
> 
> **Overview**
> Implements a **unified “touch” path** for L1 eviction recency: adds `L1ManagerListener.on_l1_keys_accessed`, wires it through `L1Manager.touch_keys`/`StorageManager.touch_l1_keys`, and bridges it in `L1EvictionPolicy` to `EvictionPolicy.on_keys_touched`.
> 
> Updates MP request lifecycle so `lookup` records the session’s `lookup_ipc_key`, and `end_session` computes all chunk `ObjectKey`s from the session and touches them in one call (covering both retrieved and stored chunks).
> 
> Adjusts `L1Manager.finish_read` to only report “touched” keys once the final read lock is released and the key is non-temporary, and hardens `LRUEvictionPolicy` against empty key lists. Adds comprehensive tests for the new touch semantics and session API changes (including `Session.get_hashes(start)` and `SessionManager.remove` returning the removed session).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3863688018d4e439822c7173b6f8d82f045ed3d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/distributed/eviction.py
+++ b/lmcache/v1/distributed/eviction.py
@@ -144,7 +144,7 @@ class L1EvictionPolicy(L1ManagerListener):
         pass
 
     def on_l1_keys_read_finished(self, keys: list[ObjectKey]):
-        pass
+        self._policy.on_keys_touched(keys)
 
     def on_l1_keys_reserved_write(self, keys: list[ObjectKey]):
         # No-op
--- a/lmcache/v1/distributed/eviction.py
+++ b/lmcache/v1/distributed/eviction.py
@@ -144,7 +144,7 @@ class L1EvictionPolicy(L1ManagerListener):
         pass
 
     def on_l1_keys_read_finished(self, keys: list[ObjectKey]):
-        self._policy.on_keys_touched(keys)
+        pass
 
     def on_l1_keys_reserved_write(self, keys: list[ObjectKey]):
         # No-op
@@ -162,6 +162,9 @@ class L1EvictionPolicy(L1ManagerListener):
 
     def on_l1_keys_finish_write_and_reserve_read(self, keys: list[ObjectKey]):
         self._policy.on_keys_created(keys)
+
+    def on_l1_keys_accessed(self, keys: list[ObjectKey]):
+        self._policy.on_keys_touched(keys)
 
 
 class L2EvictionPolicy(L2AdapterListener):
--- a/lmcache/v1/distributed/eviction_policy/lru.py
+++ b/lmcache/v1/distributed/eviction_policy/lru.py
@@ -77,6 +77,8 @@ class LRUEvictionPolicy(EvictionPolicy):
         Args:
             keys (list[ObjectKey]): The keys that have been created
         """
+        if not keys:
+            return
         with self._lock:
             # NOTE: for the request, the later keys should be evicted first.
             # For example, the request has (key1, key2, key3), if we first
@@ -97,6 +99,8 @@ class LRUEvictionPolicy(EvictionPolicy):
         Args:
             keys (list[ObjectKey]): The keys that have been accessed
         """
+        if not keys:
+            return
         with self._lock:
             # NOTE: for the request, the later keys should be evicted first.
             # The example is the same as `on_keys_created`.
@@ -113,6 +117,8 @@ class LRUEvictionPolicy(EvictionPolicy):
         Args:
             keys (list[ObjectKey]): The keys that have been deleted
         """
+        if not keys:
+            return
         with self._lock:
             for key in keys:
                 # Remove from LRU order tracking
--- a/lmcache/v1/distributed/l1_manager.py
+++ b/lmcache/v1/distributed/l1_manager.py
@@ -650,6 +650,15 @@ class L1Manager:
         )
         return ret
 
+    def touch_keys(self, keys: list[ObjectKey]):
+        """Touch the given keys to update their access time.
+
+        Args:
+            keys: The list of object keys to touch.
+        """
+        for listener in self._registered_listeners:
+            listener.on_l1_keys_accessed(keys)
+
     @l1_mgr_synchronized
     def clear(self, force: bool = False) -> None:
         """Clear objects from L1 cache.
--- a/lmcache/v1/distributed/l1_manager.py
+++ b/lmcache/v1/distributed/l1_manager.py
@@ -651,7 +651,7 @@ class L1Manager:
         return ret
 
     def touch_keys(self, keys: list[ObjectKey]):
-        """Touch the given keys to update their access time.
+        """Touch the given keys, marking the keys as accessed(retrieved or stored).
 
         Args:
             keys: The list of object keys to touch.
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -457,6 +457,15 @@ class StorageManager:
             )
         return total_hits
 
+    def touch_l1_keys(self, keys: list[ObjectKey]):
+        """
+        Touch the keys in L1 storage.
+
+        Args:
+            keys (list[ObjectKey]): List of object keys to touch.
+        """
+        self._l1_manager.touch_keys(keys)
+
     def clear(self, force: bool = False):
         """
         Clear data in the storage manager.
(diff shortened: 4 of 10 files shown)


lmcache/v1/distributed/l1_manager.py
```diff
@@ -650,6 +650,15 @@ def delete(self, keys: list[ObjectKey]) -> dict[ObjectKey, L1Error]:
         )
         return ret
 
+    def touch_keys(self, keys: list[ObjectKey]):
```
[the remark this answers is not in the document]
chunxiaozheng: `self._objects` is not used in `touch_keys`, there is no need to get lock.

chunxiaozheng:

lmcache/v1/distributed/l1_manager.py
```diff
@@ -650,6 +650,15 @@ def delete(self, keys: list[ObjectKey]) -> dict[ObjectKey, L1Error]:
...
+        """Touch the given keys to update their access time.
```
ApostaC: nit: "access time" in the docstring is a bit misleading. Maybe change to "mark them as used by"

lmcache/v1/distributed/eviction.py
```diff
@@ -114,7 +114,7 @@ def on_l1_keys_reserved_read(self, keys: list[ObjectKey]):
         pass
 
     def on_l1_keys_read_finished(self, keys: list[ObjectKey]):
-        self._policy.on_keys_touched(keys)
```
ApostaC, line 147: We probably want to keep this as well. In the P2P scenario, the keys might be read by a remote peer.
(A potential design concern here is whether the P2P peer will cause the same problem as vLLM APC).

lmcache/v1/distributed/storage_manager.py
```diff
@@ -457,6 +457,15 @@ def query_prefetch_status(
             )
         return total_hits
 
+    def touch_l1_keys(self, keys: list[ObjectKey]):
+        """
+        Touch the keys in L1 storage.
```
ApostaC: nit: let's add a better description for "touch". (Maybe "mark the keys as accessed" or something like that?)

lmcache/v1/multiprocess/session.py
ApostaC: For better clarity, can we use `@overload` annotation in standard python ([doc](https://typing.python.org/en/latest/spec/overload.html)) to define two variants of the `get_hashes` function:
- get_hashes(self, start, end)
- get_hashes(self, start)

This one is more "pythonic" than passing -1.

ApostaC (commented):
Thanks! Please see the comments.

lmcache/v1/distributed/l1_manager.py
chunxiaozheng: updated!

lmcache/v1/distributed/storage_manager.py
chunxiaozheng: updated!

lmcache/v1/multiprocess/session.py
chunxiaozheng: updated!

lmcache/v1/distributed/eviction.py
chunxiaozheng: @ApostaC Thanks for your reminder, I have kept this place. (
The current implements will trigger `touch` both in `retrieve` and `store`, and then touch all keys at the `end_session`, this may have some redundancy, but it doesn't seem like a big problem.) WDYT.

chunxiaozheng:
@ApostaC Thanks for your review, I have updated! PTAL.

lmcache/v1/distributed/eviction.py
chunxiaozheng: @ApostaC By the way, as show in the picture, when `mla` is enabled and `tp > 1`, the keys in all worker is same but will trigger multi `finished_read`, the keys will be touched multiple times.
Clipboard_Screenshot_1776151455

Is it more reasonable to only touch keys that non-temp and non-read-locked, I also updated this in current PR. 
```
    for key in keys:
            ...
            if not entry.read_lock.is_locked():
                if entry.is_temporary:
                    # NOTE: temporary objects shouldn't have write-locks
                    need_to_free.append(entry.memory_obj)
                    need_to_free_keys.append(key)
                    del self._objects[key]
                else:
                    # only the following two conditions are met,
                    # trigger `on_keys_touched` to update the LRU order.
                    # 1. non-temporary: temporary keys will be deleted;
                    # 2. non-read-locked: read-locked keys will not be evicted.
                    touch_keys.append(key)

        ...
        for listener in self._registered_listeners:
            listener.on_l1_keys_read_finished(touch_keys)
            ...
        ...
```

lmcache/v1/distributed/eviction_policy/lru.py
```diff
@@ -76,6 +76,8 @@ def on_keys_created(self, keys: list[ObjectKey]):
         Args:
             keys (list[ObjectKey]): The keys that have been created
         """
+        if not keys:
```
maobaolong, line 80: This check can avoid hold lock and release

maobaolong (approved):
LGTM

lmcache/v1/distributed/l1_manager.py
```diff
@@ -346,19 +347,26 @@ def finish_read(
...
+                    # NOTE: temporary objects shouldn't have write-locks
+                    need_to_free.append(entry.memory_obj)
+                    need_to_free_keys.append(key)
+                    del self._objects[key]
+                else:
+                    # only the following two conditions are met,
+                    # trigger `on_keys_touched` to update the LRU order.
+                    # 1. non-temporary: temporary keys will be deleted;
+                    # 2. non-read-locked: read-locked keys will not be evicted.
+                    touch_keys.append(key)
```
ApostaC: Quick question: can you explain why we don't touch the read-locked keys? The touch operation simply changes the "priority" of the keys in the eviction policy. I'm not sure why the key being locked or not impacts the touch operation.

lmcache/v1/distributed/l1_manager.py
```diff
@@ -346,19 +347,26 @@ def finish_read(
...
 
             ret[key] = L1Error.SUCCESS
             successful_keys.append(key)
 
         self._memory_manager.free(need_to_free)
 
         for listener in self._registered_listeners:
-            listener.on_l1_keys_read_finished(successful_keys)
+            listener.on_l1_keys_read_finished(touch_keys)
```
ApostaC: A small nit here: naming the variable `touch_key` but actually calling `on_l1_keys_read_finished` is a bit weird. 
(Still trying to understand why we need to introduce touch_keys here)

ApostaC (commented):
Oh I just realized I haven't submitted my review. My bad...

lmcache/v1/distributed/l1_manager.py
chunxiaozheng: @ApostaC The `touch`is to update the order of lru queue, but only keys without locks will be evicted. Therefore, for read-locaked keys, just wait for the last touch.

Without this change,  trigger `touch` in each `finish_read`, when `mla` is enabled and `tp > 1`, the same keys will be touched `tp` times.
In my tests, `tp=2`, and the log is as follows:
image

lmcache/v1/distributed/l1_manager.py
chunxiaozheng: @ApostaC thanks for your suggestion, how about naming `read_finished_keys`?

lmcache/v1/distributed/l1_manager.py
ApostaC: @chunxiaozheng thanks for the explanation! I would suggest we keep the old code since it does not create foreseeable correctness issues, and the old code is quite easier to understand -- unless we see a huge performance penalty of touching the chunk multiple times. WDYT?

chunxiaozheng: okay

ApostaC (approved):
LGTM!

PR #1610, opened 2025-09-16
[Core] P2P Backend

YaoJiayi:
Add p2p transfer.

Needs to do some debugging and benchmarking.

Done in this PR:
- [x] Implement P2P Backend with NIXL transfer channel and lookup service based on the controller
- [x] Remove old P2P related code
- [x] Update move operation in controller using the new transfer channel
- [x] Update related documentation and examples and maintain backward compatibility
- [x] Have some real benchmark numbers

Workload:
```bash
python3 long_doc_qa.py     --num-documents 20     --model meta-llama/Meta-Llama-3.1-8B-Instruct   --max-inflight-requests 2 --port 8011 --repeat-count 1
```

Local CPU offloading:
```bash
=== BENCHMARK RESULTS ===
Warmup round mean TTFT: 0.911s
Warmup round time: 17.471s
Warmup round prompt count: 20
Query round mean TTFT: 0.145s
Query round time: 6.115s
Query round prompt count: 20
```

Remote CPU offloading:
```bash
=== BENCHMARK RESULTS ===
Warmup round mean TTFT: 0.246s
Warmup round time: 6.668s
Warmup round prompt count: 20
Query round mean TTFT: 0.250s
Query round time: 6.773s
Query round prompt count: 20
```

TODO:
- [x] Docs related to P2P transfer @kobe0938 
- [ ] Add back tcp socket path as a reference path 
- [ ] Asymmetric TP for P2P transfer (and transfer channel in general) (related to https://github.com/LMCache/LMCache/pull/1629)
- [ ] Compatibility with vLLM prefix caching (related to https://github.com/LMCache/LMCache/pull/1600)
- [ ] Add e2e test
- [ ] Support loading the same request from different peers
- [ ] Add nixl (i.e., rdma) to remote cpu/disk server with transfer channel

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
[...]

PR #3036, opened 2026-04-15
[Doc] Fix instance_id typo, add kv_events config and UCX_TLS note in move example

### LINKED ISSUE #2571: [BUG]: LMCache docs for moving the kv block contains mistakes.

ianliuy:

Fixes #2571

The [KV cache move documentation](https://docs.lmcache.ai/kv_cache_management/move.html) has three mistakes that cause the move example to silently fail (returning `num_tokens: 0`).

## What's broken?

Any user following the KV cache move tutorial gets `num_tokens: 0` instead of the expected token count.

## Where is the bug?

`docs/source/kv_cache_management/move.rst`, introduced in PR #1610 (`[Core] P2P Backend`).

## Root Cause

1. **Wrong instance_id** — `instance2.yaml` sets `lmcache_instance_id: "lmcache_instance_1"` (copy-paste from instance1). The controller can't route the move to the correct destination.
2. **Missing `enable_kv_events`** — neither config block has `enable_kv_events: True`, needed for KV event tracking.
3. **No CUDA IPC example** — only `UCX_TLS=rc` (InfiniBand) is shown. Users without InfiniBand have no guidance.

## How we fixed it

Three surgical edits to `move.rst`:

| Fix | Before | After |
|-----|--------|-------|
| instance_id | `lmcache_instance_id: "lmcache_instance_1"` | `lmcache_instance_id: "lmcache_instance_2"` |
| kv_events | (missing) | `enable_kv_events: True` in both YAML blocks |
| UCX_TLS | no note | `.. note::` with copy-pasteable `export` commands for CUDA IPC and TCP fallback |

## Testing

- RST parsed cleanly with docutils (0 warnings, 0 errors)
- codespell: no spelling errors
- All 3 `UCX_TLS` values cross-referenced against CI scripts, test READMEs, and other docs pages

cc @YaoJiayi — the instance_id typo was introduced in your PR #1610 (copy-paste from instance1 to instance2 config). This is a minimal fix.

Note: `examples/cache_controller/move/` README and YAML files have the same issues but are left for a separate follow-up to keep this PR scoped.

**If applicable**:

---

> [!NOTE]
> **Low Risk**
> Low risk documentation-only change that adjusts example configuration and adds environment guidance; no runtime code paths are modified.
> 
> **Overview**
> Updates the KV cache `move` documentation example to avoid silent failures by **fixing `instance2.yaml` to use `lmcache_instance_id: "lmcache_instance_2"`** and adding `enable_kv_events: True` to both example configs.
> 
> Adds a `.. note::` explaining `UCX_TLS` options, including CUDA IPC for same-node GPUs and a TCP fallback when RDMA/CUDA IPC aren’t available.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd71823665a1d97113a38b4a2dfc7d3e369452ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/kv_cache_management/move.rst
+++ b/docs/source/kv_cache_management/move.rst
@@ -30,6 +30,7 @@ configure two lmcache instances:
     chunk_size: 256
     local_cpu: True
     max_local_cpu_size: 5
+    enable_kv_events: True
 
     # cache controller configurations
     enable_controller: True
@@ -52,10 +53,11 @@ configure two lmcache instances:
     chunk_size: 256
     local_cpu: True
     max_local_cpu_size: 5
+    enable_kv_events: True
 
     # cache controller configurations
     enable_controller: True
-    lmcache_instance_id: "lmcache_instance_1"
+    lmcache_instance_id: "lmcache_instance_2"
     controller_pull_url: "localhost:8300"
     controller_reply_url: "localhost:8400"
     lmcache_worker_ports: 8501
@@ -78,6 +80,21 @@ Start two vllm engines:
 
     PYTHONHASHSEED=123 UCX_TLS=rc CUDA_VISIBLE_DEVICES=1 LMCACHE_CONFIG_FILE=instance2.yaml vllm serve meta-llama/Llama-3.1-8B-Instruct --max-model-len 4096 \
       --gpu-memory-utilization 0.8 --port 8001 --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}'
+
+.. note::
+
+   The examples above use ``UCX_TLS=rc`` (InfiniBand RDMA). If your GPUs are
+   on the **same node**, replace it with CUDA IPC:
+
+   .. code-block:: bash
+
+      export UCX_TLS=cuda_ipc,cuda_copy,tcp
+
+   For environments **without RDMA or CUDA IPC**, use TCP as a fallback:
+
+   .. code-block:: bash
+
+      export UCX_TLS=tcp
 
 Start the lmcache controller at port 9000 and the monitor at port 9001:
 


atalhens:
I agree with the first and third points, but there is no need to set kv events to true; the way it works doesn't require it.
I had previously created a PR to fix this https://github.com/LMCache/LMCache/pull/2575 
You may take a look at the conversation section. I think you would find that helpful on why it doesn't need to be enabled. From my 2 cents, I believe what needs to be changed is the request that we are sending, if that is big enough >256 tokens, then only you would see a positive value for num_tokens; else it would always be 0. Otherwise, for request of any size < 256 tokens then`save_unfull_chunk` needs to be set to true.

PR #3032, opened 2026-04-14
[Feat] Add is_user_level property and cache_salt param to EvictionPolicy

royyhuang:
## Summary

- Add `is_user_level` property to `EvictionPolicy` base class (default `False`)
- Add `cache_salt: str | None = None` parameter to `get_eviction_actions()`
- `LRUEvictionPolicy` and `NoOpEvictionPolicy` accept and ignore `cache_salt`

## Why

The eviction controller needs a way to distinguish aggregate-level policies (LRU) from per-user-level policies (UserLRU, coming in a follow-up PR) without `isinstance` checks. `is_user_level` is the extension point — future policies override it to `True` and use the `cache_salt` parameter for scoped eviction.

## Changes

| File | Change |
|------|--------|
| `eviction.py` | `is_user_level` property (default `False`); `cache_salt` param on `get_eviction_actions()` |
| `lru.py` | Accept (ignore) `cache_salt` |
| `noop.py` | Accept (ignore) `cache_salt` |

## Test plan

- [x] 441 distributed tests pass
- [x] pre-commit clean
- [x] No behavioral change (all defaults are None, existing callers unaffected)

## Context

PR4 of the per-user L2 quota feature. See design doc: `docs/design/l2_adapters/l2_per_user_quota.md`

---

> [!NOTE]
> **Low Risk**
> Low risk: this is a backward-compatible interface extension (new property + optional parameter) with no behavior change in existing policies, which simply accept/ignore `cache_salt`.
> 
> **Overview**
> Adds an isolation-capability extension point to `EvictionPolicy` via a new `support_isolation` property (default `False`) and an optional `cache_salt` argument on `get_eviction_actions()` for scoping evictions.
> 
> Updates `LRUEvictionPolicy` and `NoOpEvictionPolicy` to accept the new `cache_salt` parameter (documented as ignored), keeping current eviction behavior unchanged while enabling future per-salt/user eviction policies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 762966e3d46cce3c8eb45cdb93df99f7fcb40f08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3029, opened 2026-04-14
[Feat] Add cache_salt parameter to MP adapter interfaces

royyhuang:
## Summary

- Add `cache_salt=""` default parameter to all scheduler and worker adapter methods in the MP multi-process adapter
- This is a **no-op change** — all defaults are empty string, preserving existing behavior. The actual data model changes will be made after interface changes are merged on vLLM side vllm-project/vllm#39837.
- Prepares the adapter interface for per-user cache isolation (follow-up PRs will populate `cache_salt` from vLLM's `Request.cache_salt`)

## Changed methods

**Scheduler adapter** (`LMCacheMPSchedulerAdapter`):
- `maybe_submit_lookup_request(request_id, token_ids, cache_salt="")`
- `free_lookup_locks(..., cache_salt="")`
- `_create_key(..., cache_salt="")`

**Worker adapter** (`LMCacheMPWorkerAdapter`):
- `submit_store_request(request_id, op, event, cache_salt="")`
- `submit_retrieve_request(request_id, op, event, cache_salt="")`
- `batched_submit_store_requests(request_ids, ops, event, cache_salts=None)`
- `batched_submit_retrieve_requests(request_ids, ops, event, cache_salts=None)`
- `_create_key(..., cache_salt="")`

`cache_salt` is accepted but not yet forwarded to `IPCCacheEngineKey` — that field will be added in a follow-up PR.

## Test plan

- [x] All 441 distributed tests pass
- [x] pre-commit clean
- [x] No behavioral change (all defaults are empty string)

## Context

PR1a of the per-user L2 quota feature. See design doc: `docs/design/l2_adapters/l2_per_user_quota.md`

---

> [!NOTE]
> **Low Risk**
> Low risk interface expansion with default values, intended to be behavior-preserving; main risk is downstream callers needing to pass/ignore the new optional args correctly.
> 
> **Overview**
> Introduces a **behavior-preserving interface extension** to the vLLM multi-process adapter by adding optional `cache_salt` parameters to scheduler lookup/free-lock calls and worker store/retrieve APIs (including batched variants), threading the value through to key construction for future per-user cache isolation.
> 
> Adds a detailed design doc (`docs/design/l2_adapters/l2_per_user_quota.md`) describing the upcoming per-user quota + eviction approach and the planned cross-component changes.
> 
[...]

PR #3040, opened 2026-04-15
[Docs] Mirror lmcache/ layout in docs/design/ for discoverability

ApostaC:
**What this PR does / why we need it**:

Reorganizes `docs/design/` to mirror the `lmcache/` package tree so developers
and coding agents can easily locate the design doc for a given module:
a module at `lmcache/<path>/` has design docs at `docs/design/<path>/`.

Concretely:
- Moved existing design docs into the mirrored paths (e.g.,
  `docs/design/l2_adapters/` → `docs/design/v1/distributed/l2_adapters/`,
  `docs/design/cli/describe.md` → `docs/design/cli/commands/describe.md`).
- Moved design-flavored `.md` files from inside `lmcache/` into the
  corresponding `docs/design/<path>/` location (`DESIGN.md`, `EVENTS.md`,
  `METRICS.md` from `lmcache/v1/mp_observability/`).
- Kept user-entry-point `README.md` files co-located with code; added
  relative symlinks from the mirrored `docs/design/<path>/README.md` so the
  mirror tree is complete.
- Added `docs/design/README.md` explaining the convention.
- Updated `CLAUDE.md`, `AGENTS.md`, `docs/coding_standards.md` §5.1, and
  `.claude/skills/pr-review/SKILL.md` so both humans and coding agents are
  directed to the mirrored location when looking for or adding design docs.
- Fixed relative links inside moved docs.

**Special notes for your reviewers**:

This branch is forked off `chore/coding-standard` (PR #3039). Please rebase
this PR onto `dev` after #3039 is merged so the PR shows only the docs-layout
changes.

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

---

> [!NOTE]
> **Low Risk**
> Docs-only reorganization plus link/reference updates; primary risk is broken doc links or tooling that assumes the old `docs/design/` paths.
> 
> **Overview**
> Standardizes design documentation by making `docs/design/` **mirror the `lmcache/` package tree**, so design docs for `lmcache/<path>/` live under `docs/design/<path>/`.
> 
[...]

PR #3042, opened 2026-04-15
[MP][Feat] Add cache_salt to ObjectKey for cache isolation

royyhuang:
## Summary

Adds a `cache_salt: str = ""` field to `ObjectKey` and `IPCCacheEngineKey` so same-content / different-user cache entries produce distinct keys. This is the data-model piece of the per-user L2 cache quota feature (PR2 of a 6-PR series). #3029 wired `cache_salt` through the MP adapter surface but left it a no-op until the data model landed; this PR makes it load-bearing.

Stacks on top of #3029 and #3032 (both already merged). **No migration is required** — un-salted traffic (the default) produces bit-identical wire keys and filenames to the pre-cache_salt format.

### Data model (`lmcache/v1/distributed/api.py`, `.../multiprocess/custom_types.py`)

- **`ObjectKey`** gains `cache_salt: str = ""` as an identity field (participates in `eq`/`hash`). `__post_init__` validates:
  - `model_name` must not contain `@` (used as field separator in serialized keys)
  - `cache_salt` must not contain `@`, `/`, `\`, or NUL
  - `cache_salt` length ≤ 128 (leaves headroom for `NAME_MAX=255` on FS filenames)
- **`IPCCacheEngineKey`** gains `cache_salt` at the end (msgspec map-mode wire compat; old 7-field payloads still decode).
- **`ipc_key_to_object_keys(ipc_key, chunk_hashes)`** reads `ipc_key.cache_salt` directly — no separate parameter, so callers can't silently drop the salt.
- **`vllm_multi_process_adapter.py`**: scheduler + worker `_create_key()` now forward `cache_salt` to `IPCCacheEngineKey`.

### Serialization format (trailing-salt, no marker)

Both the native connector wire format and the FS adapter filename format use `@` as the field separator with `cache_salt` as an **optional trailing field**:

    Unsalted: <model>@<rank>@<hash>             (== pre-cache_salt format)
    Salted  : <model>@<rank>@<hash>@<salt>

- The unsalted 3-field shape is **bit-identical** to what older LMCache builds wrote, so existing cache directories, Redis keys, Mooncake keys, etc. remain valid — no migration script needed.
- The parser uses a single `split('@')` + dispatch on field count (3 or 4). No `@@` prefix, no rsplit, no marker byte — relies on the `model_name` / `cache_salt` "no `@`" invariants instead.
- Applies to: `_object_key_to_string` (native connector), `_object_key_to_filename` / `_filename_to_object_key` (FS adapter), `FSConnector::key_to_filename` (C++ in-tree), `ExampleFSConnector::safe_filename` (external connector template).

### Backward compatibility

- **Stored data** (FS / Redis / Mooncake): unchanged for un-salted traffic — same bytes on disk / wire.
- **msgspec wire** (`IPCCacheEngineKey`): map-mode encoding means old 7-field payloads decode on new code with `cache_salt` defaulting to `""`. Explicit wire-compat test.
- **No deprecation warnings, no migration tooling** — the no-legacy-format design means there's nothing to migrate.

## Test plan

- [x] **48 new unit tests**:
  - `test_fs_l2_adapter_keys.py` (new): filename roundtrip parametrized over 2 model_name values × 3 cache_salt values; explicit 3-field and 4-field format tests; rejection tests for `.txt`, too-few / too-many field counts
  - `test_native_connector_l2_adapter.py`: `_object_key_to_string` format tests for unsalted / salted shape; salt-inequality + hash separation; model_name `@` rejection; salt `@` / `/` / `\` / NUL rejection + length cap
  - `test_custom_types.py`: msgspec wire roundtrip with + without salt
- [x] Full `tests/v1/distributed/` + `tests/v1/multiprocess/` suites: **698 passed, 26 skipped** (GPU-heavy tests excluded)
- [x] Pre-commit clean (ruff, isort, codespell, clang-format, mypy, SPDX)
- [x] End-to-end C++ verified: `lmcache_fs.so` rebuilt; unsalted key produces `<safe_model>@0xrank@hash.data`, salted key produces `<safe_model>@0xrank@hash@salt.data`

## Related

- Part of per-user L2 cache quota feature; previous merged pieces: #3029, #3032, vllm-project/vllm#39837
- Design doc: `docs/design/v1/distributed/l2_adapters/l2_per_user_quota.md` (updated in this PR)
- PR3 (#3045) stacks on this one and refactors the adapter interface for per-cache_salt accounting

---

> [!NOTE]
> **Medium Risk**
> Touches cache-key identity, serialization, and filesystem/IPC parsing paths; bugs could cause cache misses, collisions, or inability to read existing entries, though unsalted backward-compat is explicitly preserved and covered by tests.
> 
> **Overview**
> Adds `cache_salt` as a first-class identity field on both `ObjectKey` and `IPCCacheEngineKey`, with validation and msgspec wire-compat defaults, so same-content entries from different users no longer collide.
> 
> Updates key serialization across the native connector wire format, FS adapter filenames, and the in-tree/external C++ FS connectors to support an optional **trailing** `@<cache_salt>` field while keeping the unsalted 3-field format byte-identical for backward compatibility.
> 
> Plumbs `cache_salt` through the vLLM MP adapter when constructing `IPCCacheEngineKey`, changes `ipc_key_to_object_keys()` to read the salt directly from the IPC key (removing the risk of callers dropping it), updates the design doc accordingly, and adds unit tests covering roundtrips, validation, and wire compatibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1f9ef17004ac255282b5419b251798427cde5f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
```diff
@@ -39,10 +39,38 @@
...
+    """
+    if key.cache_salt:
+        return (
+            f"{_SALT_MARKER}{key.cache_salt}{_KEY_SEP}"
+            f"{key.model_name}{_KEY_SEP}"
+            f"{key.kv_rank:08x}{_KEY_SEP}{key.chunk_hash.hex()}"
+        )
+    return (
+        f"{key.model_name}{_KEY_SEP}{key.kv_rank:08x}{_KEY_SEP}{key.chunk_hash.hex()}"
+    )
```
ApostaC: Using two totally different formats is a bit weird to me. 
An alternative proposal:
- Legacy: `<model_name>@<kv_rank_hex>@<chunk_hash_hex>`
- With salt: `<model_name>@<kv_rank_hex>@<chunk_hash_hex>@<cache_salt>`

If we need to reconstruct, we just need to split by `@`, and set the cache_salt if `len(splitted_string) >=4`

lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py
```diff
@@ -92,15 +97,31 @@ async def _async_readinto_full(
...
     """
     safe_model = key.model_name.replace("/", _PATH_SLASH_REPLACEMENT)
+    if key.cache_salt:
+        return (
+            f"{_SALT_MARKER}{key.cache_salt}"
+            f"{_KEY_SEP}{safe_model}"
+            f"{_KEY_SEP}{key.kv_rank:#010x}"
+            f"{_KEY_SEP}{key.chunk_hash.hex()}"
+            f"{_FILE_EXT}"
+        )
```
ApostaC: **[bug] `cache_salt` is not sanitized for path separators.**

`safe_model` (line 116) runs `model_name.replace("/", _PATH_SLASH_REPLACEMENT)`, but `cache_salt` is inlined verbatim on line 119. If a caller passes `cache_salt="tenant/alice"` (vLLM's `cache_salt` is a free-form OpenAI-API string with no documented `/` restriction), the resulting filename embeds an unescaped path separator. `os.open` will then treat it as a missing subdirectory and fail at write time.

The `__post_init__` invariant in `api.py` only catches `@`. The C++ side at `csrc/storage_backends/fs/connector.cpp:89-93` has the same asymmetry — `cache_salt` is appended without any character substitution.

Suggested fix: extend `ObjectKey.__post_init__` (and `IPCCacheEngineKey.__post_init__`) to also reject `/` (and probably `\`, NUL, newline). Rejecting at construction time is easier to reason about than per-serializer transformations, and keeps the C++ and Python sides in lockstep without two parallel sanitization paths.

The roundtrip test at `tests/v1/distributed/test_fs_l2_adapter_keys.py` parametrizes over `["", "alice", "user-abc_123.xyz:42"]` — adding `"a/b"` would surface this immediately.

csrc/storage_backends/fs/connector.cpp
```diff
@@ -27,27 +27,55 @@ std::string FSConnector::replace_all(const std::string& str,
...
+  if (work_key.size() >= 2 && work_key[0] == KEY_SEP &&
+      work_key[1] == KEY_SEP) {
+    // Salted format: @@<cache_salt>@<model>@<kv_rank>@<hash>
+    size_t salt_end = work_key.find(KEY_SEP, 2);
+    if (salt_end == std::string::npos) {
+      return key + FILE_EXT;
+    }
+    cache_salt = work_key.substr(2, salt_end - 2);
+    work_key = work_key.substr(salt_end + 1);
+  }
```
ApostaC: **[warning] The example external connector template is not updated.**

`examples/lmc_external_native_connector/csrc/connector.cpp::safe_filename()` (lines 66-104) still uses the legacy 3-field `rfind`/`rfind` parser, unaware of the `@@` prefix. Third-party connectors cloned from that template will misparse new `@@`-prefixed salted keys: `rfind` from the right still succeeds, but `@@<salt>@<model>` ends up treated as the "model" portion, producing non-canonical filenames (and silent collisions if two different salts share a model_name structure that aliases under that mistreatment).

Either update the example alongside this in-tree change, or add a comment at the top of `safe_filename()` in the example pointing out that salted keys require additional handling and linking to this implementation.

lmcache/v1/distributed/api.py
```diff
@@ -34,6 +34,25 @@ class ObjectKey:
...
+    forms and rely on this invariant for unambiguous parsing.
+    """
+
+    def __post_init__(self) -> None:
+        # Enforce the "no @ in cache_salt" invariant at the choke point
+        # so the serialization formats in the L2 adapters stay parseable.
+        if "@" in self.cache_salt:
+            raise ValueError(
+                f"cache_salt must not contain '@' (got {self.cache_salt!r})"
+            )
```
ApostaC, line 77: **[info] Worth adding a length cap on `cache_salt` alongside the `@`-rejection.**

The FS adapter inlines `cache_salt` directly into a filename, and a sufficiently long salt will push the filename past `NAME_MAX` (255 bytes on most Linux filesystems). Not a blocker — vLLM-supplied salts will be short in practice — but the constraint belongs with the other invariants enforced here, so it's caught at construction rather than at write time.

A simple `if len(self.cache_salt) > N: raise ValueError(...)` with `N` chosen to leave headroom for `model_name`, `kv_rank_hex`, `chunk_hash_hex`, and the `.data` extension would be sufficient.

docs/design/v1/distributed/l2_adapters/l2_per_user_quota.md
ApostaC, line 1: We probably need to move this file a bit after #3040 is merged.

csrc/storage_backends/fs/connector.cpp
ApostaC, line 1: Will it be simpler if we change the obj_key <-- --> str mapping?

ApostaC (commented):
3 of the comments are generated by claude code. Remainings are my minor comments.

csrc/storage_backends/fs/connector.cpp
royyhuang: Modified the example in `examples/`

lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py
royyhuang: fixed

lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
royyhuang: Will add deprecation warning when encounter legacy format as discussed offline.

examples/lmc_external_native_connector/csrc/connector.cpp
ApostaC, line 1: @sammshen Can you take a quick look at this file?

ApostaC (approved):
LGTM!

lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py
```diff
@@ -142,6 +151,7 @@ def _filename_to_object_key(
         chunk_hash=chunk_hash,
         model_name=model_name,
         kv_rank=kv_rank,
+        cache_salt=cache_salt,
```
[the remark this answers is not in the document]
royyhuang: fixed.

sammshen (approved):
LGTM!

sammshen (changes requested):
one sec

sammshen (approved):
LTGTM

PR #3045, opened 2026-04-15
[MP][Refactor] Centralize L2 adapter byte accounting via AdapterUsage

royyhuang:
## Summary

Replaces `L2AdapterInterface.get_usage() -> tuple[float, float]` with a structured `AdapterUsage` dataclass (aggregate + per-user byte counts) and centralizes byte accounting in the base class. Adapters pass `max_capacity_bytes` to `super().__init__()` and call `_notify_keys_stored(keys, sizes)` / `_notify_keys_deleted(keys, sizes)`; the base class maintains both `_total_bytes_used` and `_per_user_size_bytes` (keyed by `ObjectKey.cache_salt` from #3042) under a dedicated `_usage_lock`.

**Existing LRU eviction behavior is unchanged** — this is purely an internal refactor that sets up the per-user quota policy in PR5. PR3 of a 6-PR series.

> ⚠️ **Stacks on #3042** — the data-model PR. While #3042 is in review, the diff in this PR includes both #3042 + this refactor. Once #3042 merges, this diff will collapse to only the refactor changes.

### Base class (`l2_adapters/base.py`)

- New **`AdapterUsage`** dataclass: `total_bytes_used`, `total_capacity_bytes`, `per_user_bytes` (read-only `MappingProxyType` snapshot), and a `usage_fraction` property that preserves the legacy `-1.0` "no eviction signal" sentinel.
- New **`supports_eviction`** property: `True` only when `max_capacity_bytes > 0`.
- **`_notify_keys_stored(keys, sizes)`** and **`_notify_keys_deleted(keys, sizes)`** drive aggregate + per-user accounting under `_usage_lock`. Listener callbacks fire outside the lock so a slow listener can't stall further notifies.
- **Underflow clamp**: `_notify_keys_deleted` clamps `_total_bytes_used` to 0 with a warning if a buggy caller would drive it negative — without the clamp the `-1` sentinel would silently disable eviction forever.
- **Default `get_usage()`** returns `AdapterUsage` derived from base counters; adapters no longer override it.

### Adapters

- **mock**: passes `max_capacity_bytes` to super, passes sizes to notifies. Keeps local `_current_size_bytes` for the synchronous within-batch capacity check (clearly documented as not the source-of-truth for external reporting).
- **nixl**: passes `pool.total_objs * align_bytes` to super, passes sizes. Only notifies keys actually stored when the pool runs out of slots mid-batch (fixes a pre-existing partial-batch over-notify). `NixlObjPool.get_usage` renamed to `get_slot_usage` to avoid colliding with the byte-based adapter API.
- **native_connector**: passes `max_capacity_gb*1024^3` to super; demux thread collects sizes alongside keys and notifies after the lock. The synchronous `delete()` no longer notifies — the demux thread does it with accurate per-key sizes drawn from `_key_sizes`. Re-stores notify with `size=0` so LRU policies still `move_to_end` without double-counting bytes.
- **fs**: removes the `get_usage` override (defaults to base, `supports_eviction == False`).

### Storage manager / eviction controller

- `StorageManager` skips `L2AdapterEvictionState` for adapters where `supports_eviction == False` even if `eviction_config` was set (logs a warning).
- `eviction_controller.report_status` and `_check_and_evict` consume `AdapterUsage.usage_fraction` instead of unpacking the old tuple. `report_status` now exposes `total_bytes_used`, `total_capacity_bytes`, and `per_user_bytes`.

## Test plan

- [x] **24 new unit tests**:
  - `test_l2_adapter_base.py` (new): `AdapterUsage` semantics, `supports_eviction`, base-class accounting (aggregate + per-user, bucket cleanup at zero, underflow clamp, `MappingProxyType` immutability, snapshot detachment, listener-only `size=0` notify, concurrent thread safety)
  - `test_mock_l2_adapter.py`: end-to-end test that `cache_salt` flows from a real adapter store through to `AdapterUsage.per_user_bytes`
  - Updated existing `get_usage` tests in `test_mock_l2_adapter.py` and `test_native_connector_l2_adapter.py` for the new dataclass shape
- [x] Full `tests/v1/distributed` + `tests/v1/multiprocess` suites: **734 passed, 26 skipped**
- [x] Pre-commit clean (ruff, isort, codespell, mypy, clang-format, SPDX)

## Related

- Stacks on #3042 (`cache_salt` on `ObjectKey`) — required because per-user accounting keys by `cache_salt`
- Design doc: `docs/design/l2_adapters/l2_per_user_quota.md` (unchanged in this PR)
- Sets up PR5 (per-user LRU policy + `QuotaManager`)

---

> [!NOTE]
> **Medium Risk**
> Touches core L2 adapter interfaces and eviction plumbing; while intended as an internal refactor, incorrect size reporting or missed notify paths could skew usage/eviction behavior across adapters.
> 
> **Overview**
> Replaces `L2AdapterInterface.get_usage()`’s tuple return with an `AdapterUsage` dataclass that reports aggregate bytes, capacity, `usage_fraction`, and per-`cache_salt` byte buckets (`bytes_by_cache_salt`) maintained centrally in `l2_adapters/base.py` via `_notify_keys_stored(keys, sizes)` / `_notify_keys_deleted(keys, sizes)`.
> 
> Updates L2 adapter implementations (mock, native connector, Nixl store, dynamic Nixl store, FS) to pass `max_capacity_bytes` to `super().__init__`, forward per-key sizes into the new notify helpers, and drop per-adapter `get_usage()` overrides; adds `supports_global_eviction` and filters eviction setup in `StorageManager` accordingly, while `L2EvictionController` status/eviction logic now consumes `AdapterUsage` fields.
> 
> Adds focused tests covering base-class accounting semantics (immutability, underflow clamp, concurrency) and updates existing adapter usage tests to the new API, plus an end-to-end check that `cache_salt` populates `bytes_by_cache_salt`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee624118298f3f963c5a4237774a2f26937fca88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -8,20 +8,69 @@
...
+
+    Replaces the old ``tuple[float, float]`` return shape with a structured
+    record that exposes both aggregate and per-user byte counts. Per-user
+    bookkeeping is keyed by ``ObjectKey.cache_salt``; an empty salt (the
+    default for un-isolated traffic) is just another bucket.
+
+    Instances are returned as immutable snapshots: ``per_user_bytes`` is a
+    read-only ``Mapping`` (``MappingProxyType``) so callers cannot mutate
+    the snapshot after the fact. Each ``get_usage()`` call returns a fresh
+    snapshot so a held reference will never reflect later state.
```
ApostaC, line 46: Just a quick question: do we still want to call it "per-user" or use something like the term "isolated"?

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -8,20 +8,69 @@
...
+    """
+
+    total_bytes_used: int
+    """Aggregate bytes across all users (cache_salt buckets)."""
+
+    total_capacity_bytes: int
+    """Adapter's maximum capacity. ``0`` means unknown / unlimited; the
+    adapter does not support usage-based eviction."""
+
+    per_user_bytes: Mapping[str, int] = field(default_factory=lambda: _EMPTY_PER_USER)
```
ApostaC, line 56: Just wondering why we use mapping and `_EMPTY_PER_USER` here to initialize the per_user_bytes default value. 

I understand that we don't want the "user" of the `AdapterUsage` instance to modify the internal "raw" byte calculation in the L2 adapter class. I do see we created a new snapshot in `L2AdapterInterface.get_usage()`, so I guess it will be okay?

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -63,9 +112,25 @@ class L2AdapterInterface(ABC):
...
     """
 
-    def __init__(self):
+    def __init__(self, max_capacity_bytes: int = 0):
+        """
+        Args:
+            max_capacity_bytes: Adapter's maximum byte capacity. ``0``
+                (default) marks the adapter as not supporting eviction —
+                ``supports_eviction`` returns ``False`` and ``get_usage``
+                returns ``usage_fraction == -1.0``.
```
ApostaC, line 121: IIUC, whether we want to do an eviction on the L2 adapter does not just depend on whether it's unlimited or not.
For example, with an unlimited NFS, we still want to have a per-user eviction policy running. 

Unless we say the per-user eviction is not triggered/controlled by `support_eviction`, I think it would be a problem. Maybe we need to tweak the naming a bit? (e.g., `support_global_eviction`?)

ApostaC (commented):
Did a pass. Will do another pass once #3042 is merged.

lmcache/v1/distributed/l2_adapters/base.py
aeon-x, line 56: is per_cache_salt_bytes a better name?   

As cache salt can be extended to `per vllm deployment`,  `per vllm deployment per user` or whatever granularity in the future.

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -278,22 +350,91 @@ def register_listener(self, listener: L2AdapterListener) -> None:
...
+                        size,
+                    )
+                    self._total_bytes_used = 0
+                new_total = (
+                    self._per_cache_salt_size_bytes.get(key.cache_salt, 0) - size
+                )
+                if new_total <= 0:
+                    self._per_cache_salt_size_bytes.pop(key.cache_salt, None)
+                else:
+                    self._per_cache_salt_size_bytes[key.cache_salt] = new_total
```
ApostaC: small potential optimization to avoid a lot of reads and writes to `per_cache_salt_bytes` (which could be a potentially very large dict).
```python
delta = {
    salt: sum(group) 
    for salt, group in itertools.groupby(
        keys, lambda key: key.cache_salt
    )
}
delta_on_total_bytes = sum(delta.values())

# Then update `self._total_bytes_used` and `self._per_cache_salt_size_bytes`
# ...
```

Additionally, itertools may have better performance (despite of working on a smaller dict) because it has an underlying C implementation.

We can have the same optimization for `_notify_keys_stored` as well

lmcache/v1/distributed/storage_controllers/eviction_controller.py
```diff
@@ -185,14 +185,16 @@ def stop(self):
...
                 {
                     "eviction_policy": state.eviction_config.eviction_policy,
                     "trigger_watermark": state.eviction_config.trigger_watermark,
                     "eviction_ratio": state.eviction_config.eviction_ratio,
-                    "current_usage": current_usage,
-                    "usage_after_ongoing_eviction": usage_after_eviction,
+                    "current_usage": usage.usage_fraction,
+                    "total_bytes_used": usage.total_bytes_used,
+                    "total_capacity_bytes": usage.total_capacity_bytes,
+                    "per_cache_salt_bytes": dict(usage.per_cache_salt_bytes),
```
ApostaC: I'm a bit hesitant to put the per_cache_salt_bytes in `report_status` because it could potentially be very big (i.e., 10K+ keys).

ApostaC (commented):
Please the comments below, thanks!

ApostaC:
@YaoJiayi please take a review on the NIXL-related part.
@sammshen please take a review on the native connector related part

lmcache/v1/distributed/l2_adapters/base.py
royyhuang: Done. Although I did not use `itertool.groupby` since it requires key to be sorted otherwise I believe it basically is the same as we iterate through `keys`.

lmcache/v1/distributed/storage_controllers/eviction_controller.py
royyhuang: Fixed.

ApostaC (approved):
LGTM now. Let's wait for @sammshen and @YaoJiayi to give a look at the native adapter and NIXL adapter.

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -660,16 +662,25 @@ def _execute_lookup_in_the_loop(
...
+                    if obj is not None:
+                        recovered_keys.append(key)
+                        recovered_sizes.append(obj.size)
                 if obj is None:
                     continue
                 bitmap.set(i)
                 obj.increase_pin_count()
             self._completed_lookup_tasks[task_id] = bitmap
+        if recovered_keys:
+            self._notify_keys_stored(recovered_keys, recovered_sizes)
```
YaoJiayi, line 683: Why do we do `_notify_keys_stored` in lookup?

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -116,9 +121,12 @@ def batched_free(self, obj_indices: list[int]) -> None:
...
             self.indices.extend(obj_indices)
 
-    def get_usage(self) -> tuple[float, float]:
+    def get_slot_usage(self) -> tuple[float, float]:
         """
-        Return (current_usage, usage_after_ongoing_eviction) in [0, 1].
+        Return (current_usage, usage_after_ongoing_eviction) in [0, 1] for
+        the slot pool. Renamed from ``get_usage`` to avoid colliding with
+        the byte-based ``L2AdapterInterface.get_usage`` shape — this is
+        an internal pool helper, not the adapter-level usage report.
```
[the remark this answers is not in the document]
YaoJiayi: is this true?

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -746,11 +762,19 @@ async def _execute_store_in_the_loop(
...
                     self._memory_objects[key] = storage_obj
                     storage_obj.decrease_pin_count()
-            self._notify_keys_stored(keys)
+                    stored_keys.append(key)
+                    stored_sizes.append(storage_obj.size)
+            # Note: ``storage_objs`` may be shorter than ``keys`` if the
+            # pool ran out of slots mid-batch (the loop above ``break``s
+            # in that case), so we notify only the keys we actually stored.
+            if stored_keys:
+                self._notify_keys_stored(stored_keys, stored_sizes)
```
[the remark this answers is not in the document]
YaoJiayi: Will we have double count here?

YaoJiayi (commented):
LGTM in general. Left a few minor comments

sammshen (approved):
native adapter changes LGTM!

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
royyhuang: This is used to report usage that is persistent from previous runs. Otherwise, we will lost these "sizes" if LMCache engine restarts (e.g., report 0 bytes used even if user already store some cache from previous LMCache run).

PR #2981, opened 2026-04-08
[Chore][CI] Upgrade CI base image to CUDA 13.0

sammshen:
vLLM nightly now requires PyTorch 2.11.0 which is built against CUDA 13.0. Update the CI base image to match.

**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the CUDA base image and runtime libraries used by CI pods, which can break GPU-dependent builds/tests if the new CUDA 13 environment differs from previous images.
> 
> **Overview**
> Updates the K3s Buildkite harness CI base image to CUDA 13 by switching the Docker base from NVIDIA’s CUDA DL image to `nvidia/cuda:13.0.2-devel-ubuntu24.04`.
> 
> Adjusts image setup to install `libcudart12` and run a generic `ldconfig` (removing the prior CUDA 12.9 compat path), aligning the CI environment with newer CUDA/PyTorch requirements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dcf764cf05d8b8a1e628e9c91ed826fe254f51d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3061, opened 2026-04-16
[Hotfix][CI] Pin vLLM nightly to cu130 index to match CUDA 13 base image

ApostaC:

Every k3 build on `dev` since 10fd636e (2026-04-16 09:55 UTC, merge of #3055) has failed at the LMCache editable install with:

```
/opt/venv/lib/python3.12/site-packages/torch/include/ATen/cuda/CUDAContextLight.h:10:10:
fatal error: cusparse.h: No such file or directory
```

Root cause is a two-hop CUDA-version mismatch that the alignment path added in #3055 can't paper over:

1. The k3 base image was bumped to `nvidia/cuda:13.0.2-devel-ubuntu24.04` in #2981 (system `nvcc` = 13).
2. `setup-env.sh` installs vLLM from the generic nightly index (`wheels.vllm.ai/nightly/vllm/`). That index serves both `cu128` and `cu130` torch wheels on different days — today it's publishing `cu128`.
3. When `torch.version.cuda="12.8"` meets system `nvcc 13`, `torch.utils.cpp_extension._check_cuda_version` aborts with `The detected CUDA version (13.0) mismatches the version that was used to compile PyTorch (12.8)`.
4. The #3055 fallback apt-installs `cuda-compiler-12-8` and sets `CUDA_HOME=/usr/local/cuda-12.8`. But `cuda-compiler-*-*` only ships `nvcc` — it does **not** ship `libcusparse-dev`, `libcublas-dev`, etc., which `torch/include/ATen/cuda/CUDAContextLight.h` pulls in via `#include <cusparse.h>`. So the build still fails, just later.

Extending the apt install list is a band-aid; the real fix is to stop resolving torch non-deterministically. This PR pins the install to vLLM's per-CUDA-major **cu130** sub-index per the official vllm.ai install instructions:

```
uv pip install -U "vllm[...]" --pre \
    --extra-index-url https://wheels.vllm.ai/nightly/cu130 \
    --extra-index-url https://download.pytorch.org/whl/cu130 \
    --index-strategy unsafe-best-match
```

`torch.version.cuda` is now deterministically `"13.0"` and matches the base image's system `nvcc 13`. `_check_cuda_version` passes with the system toolchain — no `CUDA_HOME` override, no runtime apt install, no HTML scraping.

**What gets removed**:

- HTML-scraping of `wheels.vllm.ai/nightly/vllm/` to locate a wheel URL (no longer needed now that `--extra-index-url` + `--index-strategy unsafe-best-match` resolves properly).
- The entire `:wrench: Aligning nvcc with torch's reported CUDA version` block from #3055 (~40 lines: `torch.version.cuda` detection, `nvcc --version` parsing, apt `cuda-compiler-*-*` install, `CUDA_HOME` switcheroo).

**What gets kept**:

- Pandas auto-heal loop from #3055 — independent of the CUDA mismatch, still needed because vLLM nightlies periodically add undeclared `import pandas`.
- `wait_for_server` log-dump improvement from #3055 (untouched by this PR).
- `[runai,tensorizer,flashinfer]` extras.

**What gets added**:

- A small Python sanity check that compares `torch.version.cuda` against `nvcc --version` and fails fast with a clear message if the cu130 pin ever drifts. The previous failure surfaced deep inside ninja as `cusparse.h: No such file or directory`, which took a while to track back to the real cause.

- If vLLM's cu130 nightly index is briefly empty on some day (e.g. their nightly CI hiccuped), the install will fail loudly here rather than silently drifting into a broken state elsewhere — which I think is the better failure mode.
- If/when we need to support a non-Blackwell, pre-CUDA-13 target again, the pin can be parameterized by an env var (e.g. `CI_CUDA_TAG=${CI_CUDA_TAG:-cu130}`). Not doing that now to keep the PR focused.
- Tested the install command structure locally by dry-running `uv pip install --dry-run` against the cu130 indexes; the resolver picks `vllm-*.dev*+cu130` and `torch-*+cu130` as expected.

**If applicable**:

Refs #3055, #2981.

---

> [!NOTE]
> **Low Risk**
> CI-only changes that primarily adjust dependency resolution and add a sanity check; main risk is install breakage if the `cu130` nightly index is unavailable or changes.
> 
> **Overview**
> Fixes flaky k3 CI environment setup by pinning `uv pip install` of vLLM nightlies to the CUDA 13.0 (`cu130`) indexes (vLLM + PyTorch), avoiding non-deterministic resolver picks that could pull a `cu128` torch wheel.
> 
> Removes the prior HTML wheel scraping and the runtime apt-based `nvcc` alignment/`CUDA_HOME` override logic, and replaces it with a small fail-fast check that verifies `torch.version.cuda` matches the system `nvcc` major version before installing LMCache from source.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86abd3507b60a31b61b0879d3c7e20e5f0906d8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.buildkite/k3_harness/setup-env.sh
+++ b/.buildkite/k3_harness/setup-env.sh
@@ -12,30 +12,24 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 source "${REPO_ROOT}/.buildkite/k3_tests/common_scripts/helpers.sh"
 check_gpu_health 80
 
-echo "--- :python: Installing vLLM nightly"
-# Resolve the latest nightly wheel URL directly from the nightly index.
-# PEP 440 ranks stable releases (0.17.0) above pre-release nightlies
-# (0.17.0rc1.devN), so pip/uv always picks the stable version when both
-# indexes are available. We work around this by parsing the nightly index
-# page and installing the wheel by URL.
-ARCH=$(uname -m)  # x86_64 or aarch64
-VLLM_NIGHTLY_INDEX="https://wheels.vllm.ai/nightly/vllm/"
-INDEX_HTML=$(curl -sfL "$VLLM_NIGHTLY_INDEX" 2>&1) || true
-VLLM_NIGHTLY_URL=$(echo "$INDEX_HTML" \
-    | grep -oP 'href="\K[^"]+'"${ARCH}"'\.whl' \
-    | head -1) || true
-if [[ -z "$VLLM_NIGHTLY_URL" ]]; then
-    echo "WARNING: Could not find vLLM nightly wheel for ${ARCH} — falling back to latest stable" >&2
-    uv pip install "vllm[runai,tensorizer,flashinfer]"
-else
-    # href is relative (../../<commit>/vllm-....whl), resolve to absolute URL
-    VLLM_WHEEL_URL="https://wheels.vllm.ai/nightly/vllm/${VLLM_NIGHTLY_URL}"
-    echo "Resolved nightly wheel: $VLLM_WHEEL_URL"
-    uv pip install --prerelease=allow \
-        "${VLLM_WHEEL_URL}[runai,tensorizer,flashinfer]" \
-        --extra-index-url https://pypi.org/simple \
-        --index-strategy unsafe-best-match
-fi
+echo "--- :python: Installing vLLM nightly (pinned to cu130 index)"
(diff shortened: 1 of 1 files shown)


sammshen (approved):
LGTM!

deng451e (approved):

PR #3063, opened 2026-04-16
[MP] Add tracing functionality for storage manager

ApostaC:

Adds a trace-recording subsystem under `lmcache/v1/mp_observability/trace/` that captures StorageManager public-API calls to a binary trace file. This enables offline inspection and replay of cache workloads for debugging and benchmarking.

User-facing additions:
- `--trace-level storage`: enable trace recording at the storage-manager level (only `storage` is supported today).
- `--trace-output <path>`: target trace file. Defaults to a timestamped file under `$TMPDIR` when `--trace-level` is set without an explicit path.

Public API under `lmcache.v1.mp_observability.trace`:
- `enable_tracing` decorator — publishes a `TRACE_CALL` event on each call (entry-only, inputs only). When tracing is disabled (the default), this adds a single boolean attribute load per call.
- `StorageTraceRecorder` — `EventSubscriber` that writes `TRACE_CALL` events to the binary trace file.
- `TraceReader` — streaming reader for inspection.

- Trace payloads are serialized via custom codecs (`trace/codecs.py`) to keep the binary format compact and stable across runs.
- The recorder is wired into MP servers (`server.py`, `blend_server_v2.py`) via `maybe_register_trace_recorder` called from the observability lifecycle.
- Replay lives in a separate package and is not part of this PR.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Adds new file I/O and event publication in core `StorageManager` call paths when enabled, plus new CLI/config surface; disabled-by-default gating limits impact but regressions could affect performance or shutdown/flush behavior.
> 
> **Overview**
> Adds a new **binary trace-recording** path for MP mode that captures `StorageManager` public-API calls as unified `EventType.TRACE_CALL` events and persists them to a length-prefixed msgpack trace file (new `trace/` package with decorator, codec registry, file format, recorder, and streaming reader).
> 
> Wires the feature into `lmcache server` via new observability flags `--trace-level storage` and `--trace-output`, initializes the recorder in MP server entry points, and instruments `StorageManager` methods (including manual enter/exit events for the `read_prefetched_results` context manager). Also moves `PrefetchHandle` into `distributed/api.py`, updates docs/contracts, and adds unit tests covering gating/overhead, codec round-trips, and end-to-end record/write/read behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61cf5cb46c025714a1323cae5b3f3460894de611. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/distributed/storage_manager.py
```diff
@@ -37,6 +37,14 @@
 from lmcache.v1.memory_management import MemoryObj
```
KuntaiDu, line 37: Desing doc plz

KuntaiDu (changes requested):

lmcache/v1/distributed/storage_manager.py
ApostaC: updated

lmcache/v1/distributed/storage_manager.py
```diff
@@ -512,3 +542,9 @@ def memcheck(self) -> bool:
             True if memory is consistent, False otherwise.
         """
         return self._l1_manager.memcheck()
+
+
+# Now that ``PrefetchHandle`` exists in this module's namespace, finish
+# wiring its trace codec.  Done here (rather than at codec import time)
+# to avoid a ``storage_manager → trace.codecs → storage_manager`` cycle.
+register_prefetch_handle_codec()
```
ApostaC: This is not good. The correct way to break the circular import is to move PrefetchHandle into `lmcache/v1/distributed/api.py`, and this makes more sense because it's a module-level API data class.

lmcache/v1/multiprocess/blend_server_v2.py
```diff
@@ -770,6 +771,9 @@ def run_cache_server(
     """
     event_bus = init_observability(obs_config)
 
+    # Wire up the trace recorder (no-op when --trace-level is unset).
+    maybe_register_trace_recorder(event_bus, obs_config, storage_manager_config)
```
ApostaC: Can we rename it to `maybe_initialize_trace_recorder`?

lmcache/v1/mp_observability/trace/format.py
```diff
@@ -0,0 +1,121 @@
...
+
+    format_version: int
+    """File format version; readers reject unknown values."""
+
+    level: str
+    """Trace level — currently ``"storage"``.  Future levels (``"mq"``,
+    ``"gpu"``) will share this format."""
+
+    lmcache_version: str
+    """``lmcache.__version__`` of the recording process."""
```
ApostaC: Can you confirm whether LMCache has `lmcache.__version__`? I don't think so for now. Maybe maintain a separate version in this file? One potential problem is that it cannot capture the interface changes in the storage manager and other functions that is going to be captured.

lmcache/v1/mp_observability/trace/lifecycle.py
ApostaC, line 1: Will the recorder have a `close` function? (Not sure if it's needed). 
The real thing behind this: how to make sure that the output trace file is closed correctly?

ApostaC (commented):
PTAL

sammshen (approved):
LGTM!

docs/design/v1/mp_observability/trace.md
```diff
@@ -0,0 +1,376 @@
+# Trace Recording (`lmcache trace`)
```
KuntaiDu, line 1: High-level question: should we trace MP server IPC calls, or just trace L1 calls?

KuntaiDu: For me L1 calls are easiest to impl, no GPU required and useful --- I am OK with keeping it. But IPC calls seems to be most relevant in terms of evaluating GPU-related racing conditions.

docs/design/v1/mp_observability/trace.md
```diff
@@ -0,0 +1,376 @@
...
+
+---
+
+## 3. Capture: the `@enable_tracing` decorator
+
+### Single unified event
+
+All captured calls publish the same `EventType.TRACE_CALL` event:
+
+```python
```
KuntaiDu, line 86: I'm not familiar with EventBus impl. Will the `lmcache trace` invokes too many events that causes EventBut to overflow?

docs/design/v1/mp_observability/trace.md
```diff
@@ -0,0 +1,376 @@
...
+  arg capture; `capture` / `redact` filters; entry-only on exception.
+- `test_codecs.py` — round-trip every registered type, including
+  primitives, tuples, `torch.Size`, `torch.dtype`,
+  `MemoryLayoutDesc`, `ObjectKey`, `PrefetchHandle`. Unknown-type and
+  unknown-tag error paths.
+- `test_recorder.py` — header round-trip with and without
+  `attach_storage_config`; gate flip on init / off on close;
+  publish-via-EventBus end-to-end (codec encode → file → reader → codec
+  decode); truncated-tail tolerance; bad-magic rejection;
+  `dropped_count` increments on unencodable args.
```
KuntaiDu, line 376: NIT: ideally the doc should be shorter --- this doc is for both human and LLMs.

docs/source/mp/observability.rst
```diff
@@ -349,3 +360,105 @@ View traces in any OTel-compatible backend such as **Jaeger** or
...
+    #   /tmp/lmcache-trace-<pid>-<UTC>.lct"
+
+The trace file is closed cleanly on shutdown (SIGTERM is handled by
+the EventBus stop path).
+
+Replay
+~~~~~~
+
+Replaying a recorded trace is delivered separately via the
+``lmcache trace`` and ``lmcache bench trace-replay`` CLIs.
```
KuntaiDu, line 417: English-wise `lmcache traces` sounds like the same thing as `lmcache bench trace-replay` and it can be a bit confusing.

Maybe `lmcache trace capture` and `lmcache trace replay` and leave `lmcache bench` for just OpenAI API call benchmarks.

lmcache/v1/distributed/api.py
```diff
@@ -114,6 +114,31 @@ def __post_init__(self):
             )
 
 
+@dataclass(frozen=True)
+class PrefetchHandle:
```
KuntaiDu, line 118: QQ: is this inside the scope of this PR, or it is just a handy refactor?

lmcache/v1/mp_observability/trace/codecs.py
```diff
@@ -0,0 +1,255 @@
...
+    ),
+)
+register_codec(
+    torch.Size,
+    TypeCodec(tag="torch.Size", encode=_enc_torch_size, decode=_dec_torch_size),
+)
+register_codec(
+    torch.dtype,
+    TypeCodec(tag="torch.dtype", encode=_enc_torch_dtype, decode=_dec_torch_dtype),
+)
```
KuntaiDu, line 255: QQ: should we use Protobuf here so that we can easily ensure reader / writer has the same version, or is it too heavy?

KuntaiDu: Asking this because the codec feels like redoing things in protobuf.

KuntaiDu (commented):
Left several quick design-related questions. PTAL.

docs/design/v1/mp_observability/trace.md
ApostaC: We will do MPCacheEngine and MQ (IPC) level tracing as the next step. For this version, let's just do storage-level tracing

docs/design/v1/mp_observability/trace.md
ApostaC: Not too many. Usually, a call to StorageManager's top-level function will emit multiple events. So this won't overflow event bus.

In the worst case, event bus has some overflow handling so that the memory usage won't explode

docs/source/mp/observability.rst
ApostaC: Agree, will update in the next PR where `lmcache trace` is really implemented.

lmcache/v1/distributed/api.py
ApostaC: It's a small refactor. Otherwise there will be circular import: previously the trace's module needs to import PrefetchHandle from StorageManager to do serde, and StorageManager needs to import the trace module to generate the trace.

This refactor actually should have been done eariler -- the PrefetchHandle is the public API data type, so it should stay in api.py not storage_manager.py

lmcache/v1/mp_observability/trace/codecs.py
ApostaC: Actually protobuf is more complicated. Here we just want a safe way to serialize/deserialize the function arguments when tracing.
Pickle is the easiest one to use, but it has big security problem -- attackers can embed executable code in the trace and execute it when user use `lmcache trace replay`

KuntaiDu (approved):
LGTM

PR #3074, opened 2026-04-17
[Core] Make sure `submit_store_task` sees key from the same model when handling multiple model.

KuntaiDu:
** Why

 L1 is a shared pool, so keys from different models can end up in one finish_write. The StoreListener coalesces them, and StoreController then hands the mixed-shape batch to submit_store_task. This works today because L2 adapters iterate per-key, but it breaks any future consumer that treats a batch as uniform-shape (e.g. a serde processor that sizes a temp buffer from the first object).

** What

In StoreController._process_new_keys, group incoming keys by ObjectKey.model_name before running the per-adapter flow. Each submit_store_task call now sees a single model's keys (same shape). Assumes same model_name ⇒ same shape, which holds within a server.

PrefetchController is already structurally single-model (one layout_desc per submit_prefetch_request), so no change.

** Testing

New end-to-end test in tests/v1/distributed/test_multimodel.py writes two models with different shapes in one finish_write and asserts every submit_store_task sees uniform (shape, dtype). Fails pre-fix, passes post-fix.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Changes StoreController batching semantics by splitting one write-completion batch into multiple per-shape store submissions, which can affect L2 adapter load and task tracking in the async store loop. Adds a CUDA-gated end-to-end test that may be sensitive to timing/races.
> 
> **Overview**
> Ensures `StoreController` never calls `submit_store_task` with a mixed-shape batch when a single `finish_write()` contains keys from multiple models/parallelism configs.
> 
> `_process_new_keys` now **groups keys by KV shape** (currently `(model_name, kv_rank)`) and runs the existing policy/`reserve_read`/submit flow per group via a new helper, and the design doc is updated to reflect this invariant. Adds an end-to-end multi-model test that writes two different layouts in one `finish_write` and asserts each `submit_store_task` sees a uniform `(shape, dtype)`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8845c085da7b3ad19f1dcb858c23260bf0c3e1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/v1/distributed/storage_controllers/store_controller.py
```diff
@@ -291,6 +292,15 @@ def _process_new_keys(self, keys: list[ObjectKey]) -> None:
         Args:
             keys (list[ObjectKey]): Keys that finished writing to L1.
         """
+        # Split by model so each submit_store_task sees uniform (shape, dtype).
+        by_model = defaultdict(list)
+        for key in keys:
+            by_model[key.model_name].append(key)
```
ApostaC: quick suggestion: we can have a helper function called `_group_keys_by_shape` inside the `_process_new_keys` function.

The reason is that with the same model name, the objects may still have different shapes when the world size is different (and may (though less likely) have more complicated logic in the future). Therefore, we may need to use both `model_name` and `world_size` to group the keys.

ApostaC (approved):
Left a comment, but gave an approval.

lmcache/v1/distributed/storage_controllers/store_controller.py
KuntaiDu: Fixed.

Oasis-Git (approved):

docs/design/v1/distributed/l2_adapters/overall.md
```diff
@@ -157,20 +157,25 @@ _store_loop: poll wakes up
   ▼
 _process_new_keys(keys)
   │
-  ├─ 1. StorePolicy.select_store_targets(keys, adapters)
-  │     → dict[adapter_index, list[ObjectKey]]
+  ├─ 1. Group keys by shape (today: (model_name, kv_rank); L1 is a shared
```
DongDongJu, line 160: Here's shared pool means that shared inside node?

DongDongJu (approved):
I just left one my personal question.

docs/design/v1/distributed/l2_adapters/overall.md
KuntaiDu: Yes. On the same node, different vLLM instances (that runs different models / same model with different parallel strategy) share the same LMCache L1 memory pool on that node.

PR #3075, opened 2026-04-18
[CLI] Introduce lmcache trace CLI

ApostaC:
**Screenshot**

image

Introduces a new `lmcache trace` CLI for working with storage-level trace files produced by `lmcache server --trace-level storage`. This enables offline analysis and replay of recorded workloads, which is useful for:

- **Regression hunting** — reproduce production bugs offline by replaying captured traces against a build under investigation.
- **Performance characterization** — measure L1/L2 latency distributions under realistic access patterns without vLLM or a GPU.
- **Configuration tuning** — compare L1 sizes, eviction policies, and L2 adapter choices against identical input.

Subcommands:

- `lmcache trace info FILE` — print header metadata and per-qualname record counts.
- `lmcache trace replay FILE ...` — reissue recorded calls against a fresh `StorageManager`, honoring recorded inter-call timings. Accepts the standard storage-manager flags plus the full observability flag surface (shared with `lmcache server`), `--verbose`/`--jsonl-out` for per-record output, `--output-dir`/`--no-csv`/`--json` for aggregated summaries, and `-q` to suppress the terminal metrics table.

Trace *capture* is intentionally not a `trace` subcommand — recording is bound to the live process via `lmcache server --trace-level storage [--trace-output …]`. Surfacing a CLI stub here would only duplicate that flag while suggesting a runtime-capture CLI that does not yet exist.

The CLI, replay driver, dispatcher, and stats collector are colocated under `lmcache/cli/commands/trace/`.

**If applicable**:

- [x] this PR contains user facing changes - docs added (`docs/source/mp/tracing_and_debugging.rst`, updated `docs/design/v1/mp_observability/trace.md`)
- [x] this PR contains unit tests (`tests/cli/commands/test_trace_command.py`, `tests/cli/commands/trace/`)

---

> [!NOTE]
> **Medium Risk**
> Adds a new CLI entry point that constructs and drives a real `StorageManager` during trace replay, plus new stats/export paths; failures could impact user workflows and produce misleading replay results if edge cases are missed. Also adjusts trace-config digest computation shared with the recorder, which affects mismatch detection semantics.
> 
> **Overview**
> Adds a new top-level `lmcache trace` command with `info` (header + op counts) and `replay` (timing-paced reissue of recorded `StorageManager` calls) including optional per-record streaming (`--verbose`, `--jsonl-out`), aggregated CSV/JSON exports, and a terminal metrics table.
> 
> Implements replay internals (`StorageReplayDriver`, `CallDispatcher`, `ReplayStatsCollector`) to decode trace records, map recorded `qualname`s to live `StorageManager` methods (including FIFO pairing for `read_prefetched_results` enter/exit), honor recorded inter-call timing, and report replay/skip/failure totals plus config-digest mismatch.
> 
> Refactors trace recording to expose `safe_storage_config_dict()` so replay can compute the same header digest algorithm, adds extensive docs for tracing/replay, and updates CI to skip the new `tests/cli/commands/trace` suite in the non-CUDA test job.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c693525bbe9fdbd8ff405dfc38c2a83c8aee2ec1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/cli/commands/bench/__init__.py
ApostaC, line 1: This change is not needed.

lmcache/tools/trace_replay/__init__.py
ApostaC, line 1: All the trace replay-related functionality should be moved to under `lmcache/cli/commands/trace`.

ApostaC (commented):
Comments are below.

KuntaiDu (approved):
LGTM!

lmcache/cli/commands/trace/driver.py
```diff
@@ -0,0 +1,372 @@
...
+                    "trace replay: no handler for qualname %r; skipping",
+                    record.qualname,
+                )
+                if on_record is not None:
+                    on_record(record.qualname, 0.0, True)
+                continue
+
+            t0 = time.monotonic()
+            try:
+                self._dispatcher.dispatch(record.qualname, context, decoded_args)
```
royyhuang, line 311: IIUC, even though the underline dispatched SM method is non-blocking but `dispatch` itself is blocking. I would recommend put this on separate thread to reduce the chance the replayer machine falling behind the collected trace.

lmcache/cli/commands/trace/driver.py
```diff
@@ -0,0 +1,372 @@
...
+                latency = time.monotonic() - t0
```
royyhuang, line 312: We are recording the elapsed time on the non-blocking methods not the actual IO time (even though it may reflect the actual performance to some extent). I feel this could be more useful if we report the actual performance metrics.

royyhuang (approved):
Otherwise, LGTM!

lmcache/cli/commands/trace/driver.py
royyhuang: Even better if we can have dispatcher in C/C++ since later we could have the case that the trace is collected and aggregated from multiple nodes and replayed on a local dev machine. Python w/ GIL can easily fall behind the trace. (Also this can be in the long run)

lmcache/cli/commands/trace/driver.py
ApostaC: The actual I/O time performance needs to be monitored via Prometheus metrics / OTel events. The event bus is enabled in the replayer, so it will generate the observability events.

lmcache/cli/commands/trace/driver.py
ApostaC: Oh this makes sense! Will do it in a follow-up PR.

PR #3079, opened 2026-04-19
[ROCm][MP] Fix HIP invalid-argument on lazy host buffer past 2 GB

Shaoting-Feng:

On AMD/ROCm in MP mode, the second long request fails with
`HIP error: invalid argument` once the `LazyMemoryAllocator` starts
handing out addresses at or past the 2 GB virtual-offset boundary.

Root cause: `lmcache_memcpy_async` in `csrc/mem_kernels.cu` uses an
unqualified `min(size_t, size_t)`. HIP's overload set resolves this to
an `int`-returning variant, silently narrowing the arguments. Once
`host_buffer_offset + nbytes >= 2 GB`, the intermediate value crosses
`INT_MIN` as a signed int, `real_end` is cast back to `size_t` as a
huge garbage number, and `hipMemcpyAsync` rejects the call. CUDA's
resolution picks a wider overload, which is why CUDA users never hit
this.

Fix: `csrc/mem_kernels.cu` is updated to use `std::min<size_t>` so a
future rebuild of the kernel is correct on HIP.

**If applicable**:


Shaoting-Feng:
Tested by sending two long requests consecutively, and it did not fail this time.

ApostaC (approved):
Good catch!

PR #3086, opened 2026-04-20
[CLI][fix] lazy torch import in __init__.py to unblock CLI-only installs

deng451e:
lmcache/__init__.py imported torch at the top level, causing any import of a lmcache subpackage — including lmcache.cli — to
  unconditionally require torch. This broke the lightweight lmcache-cli wheel, which ships without torch.

  Changes

  - Move import torch inside _get_backend() so it only loads when a CUDA backend is being selected.
  - Wrap the _get_backend() call in try/except ImportError so CLI-only environments (no torch/numba) import cleanly; the compute
  backend is simply skipped.
  - Make the cache-simulator subcommand optional: skip registration if its extra dependencies (sortedcontainers, matplotlib) aren't
  installed.
  - Add a CI smoke test that installs the CLI wheel without torch and verifies import lmcache and lmcache --help succeed.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -171,6 +171,12 @@ jobs:
                 cp pyproject_cli.toml pyproject.toml
                 NO_CUDA_EXT=1 python -m build --wheel
 
+            - name: Smoke-test CLI wheel (no torch)
+              run: |
+                pip install dist/lmcache_cli-*.whl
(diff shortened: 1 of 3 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3093, opened 2026-04-21
[Hotfix][CI] Fail-fast when vLLM CLI import chain is broken post-install

sammshen:
The k3 integration tests have been red since 2026-04-21 ~04:00 UTC with:

    ImportError: cannot import name 'GenerationConfig' from 'transformers'
    (/opt/venv/lib/python3.12/site-packages/transformers/__init__.py)

at vllm/transformers_utils/config.py line 18. The failure surfaces 180s after the test starts as a generic "vLLM failed to start on port 8000 within 180s" in wait_for_server, and only then does the harness tail vllm.log to show the real traceback.

Root cause is that setup-env.sh declared the environment "ready" without exercising the CLI import chain that `vllm serve` runs at startup. The existing sequence was:

  1. Install vLLM nightly
  2. Probe `from vllm.entrypoints.cli.main import main` (auto-heal)
  3. `uv pip install -e . --no-build-isolation` (LMCache install)
  4. `python -c "import vllm; import lmcache"` (final probe)

Step 3 silently downgrades 9 transitive packages (opentelemetry-* 1.41->1.40, prometheus-client 0.25->0.24.1) to honor the caps in requirements/common.txt. Step 4 is the only post-install check, but plain `import vllm` doesn't pull vllm.entrypoints.cli.main -> vllm.config -> vllm.transformers_utils.config, so any CLI-chain breakage introduced by the downgrades slips through until the first `vllm serve` subprocess fails 180s later.

Fixes:

- Extract the CLI import probe into a `probe_vllm_cli` function so the same check runs both during the auto-heal loop (pre-install) and as a hard probe after the LMCache install.
- Add a post-install CLI probe that fails fast with the actual traceback and a full `uv pip freeze` if the env is broken, instead of letting the 180s test-harness timeout hide the real failure.
- Snapshot `uv pip freeze` before and after `uv pip install -e .` and diff them, so the silent downgrades done by LMCache's pins are visible in the build log instead of having to be reconstructed from package-install stderr.

With this change, the current k3 failure mode surfaces in ~10s at setup time with a clear ImportError traceback and the exact package versions at fault, instead of a 180s port-wait timeout.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> CI-only changes but they directly mutate the installed `vllm` package and alter dependency installation behavior, which could introduce new breakages or mask upstream issues if the patch/reinstall assumptions drift.
> 
> **Overview**
> Improves K3 CI environment setup to *fail fast* when the `vllm serve` import chain is broken, instead of surfacing later as a 180s server-start timeout.
> 
> `setup-env.sh` now clears Python/uv caches before installs, forces a reinstall of key `vllm`/`transformers`-stack packages, patches the installed `vllm` CLI entrypoint to pre-import `transformers` (avoiding a background-thread race), and replaces the previous light import check with a `vllm --help` probe used both during auto-heal and after the LMCache editable install. It also snapshots/diffs `uv pip freeze` around the LMCache install and emits a diagnostic dump / full package list on failures to aid debugging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b700632f5597f0f42cb679c88be77d93d20aa849. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.buildkite/k3_harness/setup-env.sh
+++ b/.buildkite/k3_harness/setup-env.sh
@@ -31,14 +31,25 @@ uv pip install -U "vllm[runai,tensorizer,flashinfer]" --pre \
     --extra-index-url https://download.pytorch.org/whl/cu130 \
     --index-strategy unsafe-best-match
 
+# Probe the vLLM CLI import chain. This is the same chain that `vllm serve`
+# runs at startup, so anything that fails here would otherwise surface 180s
+# later as a silent "vLLM failed to start on port 8000" timeout in the test
+# harness. Shared between the pre-install auto-heal loop below and the
+# post-install hard probe at the end of this script.
+probe_vllm_cli() {
+    python -c "from vllm.entrypoints.cli.main import main" 2>&1
+}
+
 # vLLM nightlies periodically add eager imports of packages that aren't in
-# their declared deps (e.g. `pandas` from vllm/_aiter_ops.py). Probe-import
-# vllm's CLI entry point and auto-install any ModuleNotFoundError modules
-# so the job keeps going. Capped to avoid infinite loops; every auto-install
-# is logged so the drift is visible in the build output.
+# their declared deps (e.g. `pandas` from vllm/_aiter_ops.py). Auto-install
+# any ModuleNotFoundError modules so the job keeps going. Capped to avoid
+# infinite loops; every auto-install is logged so the drift is visible in
+# the build output. ImportError with a missing top-level name (e.g. a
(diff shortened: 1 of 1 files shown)


chunxiaozheng:
@maobaolong could you help take a look?

royyhuang (approved):

maobaolong (approved):
lgtm

sammshen:
once this PR passes CI, it will be fixed

PR #3100, opened 2026-04-22
[Hotfix][CI] Replace vllm main.py patch with sitecustomize.py

sammshen:
So the issue is that vllm is now importing `torch` and `transformers` in a background thread to speed up the start. However, this causes a race condition in python's import system since python will first statically resolve all imports recursively throughout the codebase (establishing one module per import) and then populate these modules by running the correct `__init__.py` for each. Usually this is not problem since both of these steps execute before any real code using the modules is executed but with two threads, it is possible for vllm to reach an `transformeres` related code (e.g. `GenerationConfig`) that will think the symbol is resolved but it actually is the other thread still populating the module object. the solution in #3093  was a monkeypatch to modify vllm code to import transformers before the background thread launches but the correct solution is just to import transformers before starting vllm.

PR #3093 defused a race between vllm's BG-thread transformers preload and the main thread's `from transformers import GenerationConfig` by patching /opt/venv/.../vllm/entrypoints/cli/main.py to add `import transformers` above the thread spawn. That worked for the specific vllm nightly it was written against, but it was brittle: the patch script used a fixed text needle for the thread-spawn line and failed hard the moment the next vllm nightly restructured the file. CI started reporting:

    vllm/entrypoints/cli/main.py layout unexpected;
    BG-thread race patch could not be applied.
    ERROR: setup-env.sh failed at line 90 (exit code 1)

Swap the source-patch for a sitecustomize.py written into /opt/venv/lib/python3.12/site-packages. Python auto-imports sitecustomize during interpreter startup -- before any user code in the target script, including /opt/venv/bin/vllm. The shim does `import transformers` on the main thread, which drives the _LazyModule init to completion (registers in sys.modules, builds _import_structure, builds _class_to_module, installs __getattr__) before vllm's CLI even begins to run.

Whatever vllm nightly is doing now to preload transformers in a background thread -- or not doing, if upstream removed it -- the second `import transformers` finds a fully-wired module in sys.modules and returns immediately. Attribute lookups for GenerationConfig / PretrainedConfig on the main thread then succeed every time.

The new approach depends only on `import transformers` working, which is a prerequisite of any vllm install, so it is immune to vllm internal restructuring. When upstream vllm eventually stops running the BG preload, this file can be removed with no behavioral change.

sitecustomize swallows exceptions and falls through on any failure: it must never crash the interpreter, because every Python invocation in the venv runs it on startup. If transformers is genuinely broken, we want the real import site in vllm to fail with its own specific traceback, not have the whole interpreter die on startup.

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Touches CI environment bootstrapping and vLLM integration adapter initialization, which can affect test reliability and runtime configuration across distributed setups; changes are targeted but impact all K3 jobs.
> 
> **Overview**
> Replaces the brittle CI monkeypatch of `vllm/entrypoints/cli/main.py` with a `sitecustomize.py` shim that pre-imports `transformers` early, preventing a vLLM background-thread import race in both the standard K3 env and the blend (two-venv) setup.
> 
> Hardens K3 pipelines by (1) forcing LMCache editable installs to bypass `setuptools_scm` tag parsing via `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LMCACHE`, and (2) retrying the correctness test up to 3 times to mitigate non-deterministic output diffs across vLLM process restarts.
> 
> Updates the vLLM multiprocess adapter (`vllm_multi_process_adapter.py`) constructor signatures to be backward-compatible with vLLM’s bundled caller by allowing legacy positional args and synthesizing a `ParallelStrategy` when not provided.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b27920408f75e751e8a960f86ecc4b13bb4a530. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.buildkite/k3_harness/setup-blend-env.sh
+++ b/.buildkite/k3_harness/setup-blend-env.sh
@@ -67,7 +67,28 @@ export FLASHINFER_DISABLE_VERSION_CHECK=1
 "${DEFAULT_VENV_BIN}/python" -c 'import torch; print(f"default venv torch={torch.__version__}, torch.version.cuda={torch.version.cuda}")' 
 "${TEST_VENV_BIN}/python" -c 'import torch; print(f"test venv torch={torch.__version__}, torch.version.cuda={torch.version.cuda}")'
 
+echo "--- :wrench: Install sitecustomize.py in both venvs (transformers pre-import)"
+# The blend test runs prefiller in /opt/venv and decoder in
+# /workspace/.venv. Both launch vllm, which spawns a BG thread that
+# calls `import transformers` concurrently with the main thread's
+# later `from transformers import GenerationConfig, ...`. Pre-import
+# on the main thread via sitecustomize.py so the race is closed in
+# both venvs (see setup-env.sh for the full write-up).
+for site in \
+    /opt/venv/lib/python3.12/site-packages \
+    /workspace/.venv/lib/python3.12/site-packages; do
+    cat > "${site}/sitecustomize.py" <<'PY'
+try:
+    import transformers  # noqa: F401
+except Exception:
+    pass
+PY
+done
+
 echo "--- :python: Installing LMCache from source"
+# Skip setuptools_scm git describe; the repo carries non-PEP-440 tags
+# (nightly, nightly-cu13) that crash the newer vcs_versioning backend.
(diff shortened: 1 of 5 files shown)


maobaolong (approved):
@sammshen Thanks for fixing this CI issue, look forward to merge this PR.

DongDongJu (approved):
LGTM

PR #3026, opened 2026-04-14
[MP] Add extra_config to construct lmcache mp adapter

maobaolong:
**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

---

> [!NOTE]
> **Medium Risk**
> Touches MP adapter initialization and timeout/heartbeat configuration resolution; mis-typed `extra_config` values or missing prefix handling could change runtime behavior or cause connection failures.
> 
> **Overview**
> Wires vLLM’s `kv_connector_extra_config` through `lmcache_mp_connector_0180` into the LMCache MP scheduler/worker adapters, replacing the previous hardcoded/locally-parsed `mq_timeout` and `heartbeat_interval` constructor args.
> 
> Adds centralized `ExtraConfigDefault` + `_resolve_extra_config()` in `vllm_multi_process_adapter.py` to parse `lmcache.mp.*` settings (type-cast + log defaults/overrides) and uses the resolved values for chunk-size queries, MQ request timeouts, and heartbeat intervals, including updated connection-timeout error messages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d69c417e296c3b90bdbd29f9122a5dce689c7903. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3105, opened 2026-04-22
feat(mp): fully synchronous mode (sync lookup + sync retrieve)

yoo-kumaneko:
Add an opt-in "fully synchronous" mode to the MP connector that mirrors the native (non-MP) LMCache connector's blocking flow:

- Scheduler uses SYNC_LOOKUP to get the hit count in a single round trip, instead of the two-phase LOOKUP + QUERY_PREFETCH_STATUS poll.
- Worker blocks in start_load_kv() until all retrieves complete, so KV data is in L1 before the vLLM forward pass runs.

With this, requests go directly from WAITING to RUNNING (no WAITING_FOR_REMOTE_KVS / get_finished promote-back), eliminating the scheduling latency of the async path.

Activated by setting lmcache.mp.sync_mode=true in
kv_connector_extra_config. Default (async) behavior is unchanged.

Changes
-------
Protocol (lmcache/v1/multiprocess/protocols/):
- New RequestType.SYNC_LOOKUP ([KeyType, int] -> int) served by the normal (non-affinity) thread pool.

Server (lmcache/v1/multiprocess/server.py):
- sync_lookup(key, tp_size) -> int: runs lookup(), polls query_prefetch_lookup_hits until the lookup phase completes, and returns the hit chunk count.  On non-zero hits the prefetch job stays registered under request_id so the worker can poll QUERY_PREFETCH_STATUS to gate RETRIEVE on the L2->L1 transfer. Zero-hits eagerly drain the job; end_session drains any lingering job as a fallback for worker crashes / early cancellation.

Scheduler adapter (LMCacheMPSyncLookUpSchedulerAdapter):
- Subclasses LMCacheMPSchedulerAdapter and exposes sync_lookup() as a single blocking SYNC_LOOKUP call returning matched token count.

Worker adapter (LMCacheMPWorkerAdapter):
- New sync_mode kwarg.  In sync_mode, submit_retrieve_request polls QUERY_PREFETCH_STATUS per request_id (treating a non-None response, including 0 from a TP peer that already consumed the job, as "ready") before dispatching RETRIEVE.  batched_submit_retrieve_requests blocks on all RETRIEVE futures before returning.

Connector (lmcache/integration/vllm/lmcache_mp_connector_0180.py):
- Parse lmcache.mp.sync_mode from kv_connector_extra_config, plumb through create_scheduler_adapter / create_worker_adapter.  In sync_mode, get_num_new_matched_tokens() issues sync_lookup and returns load_kv_async=False so the request bypasses WAITING_FOR_REMOTE_KVS and goes straight to RUNNING.

---

> [!NOTE]
> **Medium Risk**
> Changes MP connector control flow and server protocol behavior; while opt-in, the new blocking/polling paths could impact scheduling latency or hang on timeouts/edge cases if prefetch job lifecycle handling is incorrect.
> 
> **Overview**
> Adds an opt-in **fully synchronous** multi-process (MP) connector flow (`lmcache.mp.sync_mode`) that replaces the scheduler’s async `LOOKUP`+polling with a single blocking `SYNC_LOOKUP`, and makes workers block until all `RETRIEVE` operations complete before the forward pass proceeds.
> 
> This introduces a new MP RPC `RequestType.SYNC_LOOKUP` (protocol + server handler + thread-pool wiring), implements `MPCacheEngine.sync_lookup()` with polling/drain logic to avoid leaking prefetch jobs, and updates the vLLM connector/adapters to select the sync scheduler adapter and gate `RETRIEVE` on prefetch completion in sync mode while keeping the default async behavior unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f70fa24c2c2e46bc4ac0b111b8098fa36ce448b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector_0180.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0180.py
@@ -26,12 +26,14 @@ from vllm.v1.utils import ConstantList
 try:
     from lmcache.integration.vllm.vllm_multi_process_adapter import (
         LMCacheMPSchedulerAdapter,
+        LMCacheMPSyncLookUpSchedulerAdapter,
         LMCacheMPWorkerAdapter,
         LoadStoreOp,
     )
 except ImportError:
     from vllm.distributed.kv_transfer.kv_connector.v1.lmcache_integration import (
         LMCacheMPSchedulerAdapter,
+        LMCacheMPSyncLookUpSchedulerAdapter,
         LMCacheMPWorkerAdapter,
         LoadStoreOp,
     )
@@ -106,6 +108,7 @@ def create_scheduler_adapter(
     vllm_config: VllmConfig,
     mq_timeout: float,
     heartbeat_interval: float,
+    sync_mode: bool = False,
 ) -> LMCacheMPSchedulerAdapter:
     world_size, kv_rank = extract_world_size_and_kv_rank(
         vllm_config.parallel_config.world_size,
@@ -120,7 +123,12 @@ def create_scheduler_adapter(
     if _adapter_accepts_tp_size():
         kwargs["tp_size"] = tp_size
 
-    return LMCacheMPSchedulerAdapter(
+    adapter_cls = (
+        LMCacheMPSyncLookUpSchedulerAdapter
+        if sync_mode
+        else LMCacheMPSchedulerAdapter
+    )
+    return adapter_cls(
         server_url,
         zmq_context,
         vllm_config.model_config.model,
@@ -139,6 +147,7 @@ def create_worker_adapter(
     vllm_config: VllmConfig,
     mq_timeout: float,
     heartbeat_interval: float,
+    sync_mode: bool = False,
 ) -> LMCacheMPWorkerAdapter:
     world_size, kv_rank = extract_world_size_and_kv_rank(
         vllm_config.parallel_config.world_size,
@@ -154,6 +163,7 @@ def create_worker_adapter(
         vllm_config.cache_config.block_size,
         mq_timeout=mq_timeout,
         heartbeat_interval=heartbeat_interval,
+        sync_mode=sync_mode,
     )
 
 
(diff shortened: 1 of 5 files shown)


lmcache/v1/multiprocess/server.py
```diff
@@ -1046,6 +1113,7 @@ def run_cache_server(
     )
     add_handler_helper(server, RequestType.STORE, engine.store)
     add_handler_helper(server, RequestType.LOOKUP, engine.lookup)
+    add_handler_helper(server, RequestType.SYNC_LOOKUP, engine.sync_lookup)
```
ApostaC, line 1116: Can you add the same lines to `blend_server_v2.py` (for here and the `add_normal_thread_pool` line below)? Thanks!

lmcache/integration/vllm/lmcache_mp_connector_0180.py
ApostaC, line 1: @maobaolong Just wondering what's the plan for `lmcache_mp_connector_0180.py`? Do we want to keep this or make this our "internal dev version"?

lmcache/integration/vllm/lmcache_mp_connector_0180.py
```diff
@@ -106,6 +108,7 @@ def create_scheduler_adapter(
     vllm_config: VllmConfig,
     mq_timeout: float,
     heartbeat_interval: float,
+    sync_mode: bool = False,
```
ApostaC, line 111: In #3026, we are adding `extra_configs` for argument passing, and I think we can reuse that here.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -813,6 +855,23 @@ def submit_retrieve_request(
...
+        if self.sync_mode:
+            while True:
+                result = send_lmcache_request(
+                    self.mq_client,
+                    RequestType.QUERY_PREFETCH_STATUS,
+                    [request_id],
+                ).result(timeout=self._mq_timeout)
+                if result is not None:
+                    break
+                time.sleep(0.005)
```
ApostaC, line 873: This will block the vLLM worker process and may have performance impact. Ideally, the polling should happen in a background "job queue", and the real submission of retrieve could be something like a "callback". 
In this case, we may need to have another MPWorkerAdapter to avoid changing to much on the current worker adapter.

ApostaC (changes requested):
Logic is pretty clear to me now! Please take a look at the comments, thanks!

PR #3114, opened 2026-04-22
[MP][Feat] Real-reuse gap metrics

ruizhang0101:
## Summary

Adds `SMLifecycleSubscriber` emitting two new per-`cache_salt` histograms:

- **`lmcache_mp.real_reuse_gap_seconds`** — wall-clock gap between a chunk's last access (read or write) and its next read. Captures **storage cost** (how long a stored chunk sat unread before its next read).
- **`lmcache_mp.real_reuse_gap_chunks`** — gap measured in a per-`cache_salt` access counter that bumps on every read AND write of every chunk. Captures **storage volume** (chunk-accesses elapsed while waiting for the next read).

Both are emitted **only on read events** (`read → read` and `write → read`); writes update the anchor but never emit a sample.

### Sources

Subscribes to caller-facing StorageManager events:

- `SM_READ_PREFETCHED_FINISHED` — caller signals prefetched chunks consumed by GPU.
- `SM_WRITE_FINISHED` — caller signals chunk store complete.

Deliberately not `L1_READ_FINISHED` / `L1_WRITE_FINISHED` — those also fire from internal store-controller / prefetch-controller lock releases that aren't user-driven cache uses.

### Sampling and bounds

- Per-`(cache_salt, chunk_hash)` deterministic hash sampling — same chunk always gets the same decision.
- Per-salt access counter advances on **every** access (sampled or not, read or write) so chunks-gap reflects true volume.
- Tracking dict bounded by `_REUSE_TRACK_CAP = 100_000`. On overflow, random eviction (entries already passed sampling gate, so equal priority).
- `cache_salt` cardinality is naturally bounded — operator-set tenant identifier with `≤ 128` char limit. Counter dict has no internal cap.

### Isolation

`cache_salt` tag on both histograms, distinct `(cache_salt, chunk_hash)` keys, per-salt counter. Tenants cannot pollute each other's metrics.

### TP fanout

A single logical chunk fans out to one ObjectKey per `kv_rank`. Both handlers dedupe per logical chunk so the counter advances once per logical-chunk access regardless of TP size.

### Known caveat — CB blend error path

`blend_server_v2` calls `finish_read_prefetched` unconditionally in `finally` regardless of whether the `read_prefetched_results` context exited cleanly. On CB error, both publish sites fire and the metric records an extra near-zero-gap sample. Existing bug acknowledged in `blend_server_v2.py` source comment; out of scope here. Effect: minor histogram-tail pollution toward zero on CB errors only.

## Changes

- `lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py` — new `SMLifecycleSubscriber`.
- `lmcache/v1/mp_observability/subscribers/metrics/__init__.py` — export.
- `lmcache/v1/mp_observability/config.py` — register in `init_observability` with `sample_rate` from observability config.
- `docs/design/v1/mp_observability/METRICS.md` — design-doc table entries.
- `docs/source/mp/observability.rst` — user-facing "StorageManager Real-Reuse Metrics" section.

`L1LifecycleSubscriber` is unchanged.

## Test plan

- [x] `tests/v1/mp_observability/subscribers/metrics/test_sm_lifecycle.py` — 15 tests across:
  - `read → read` and `write → read` emit gap.
  - `write → write` and `read → write` do not emit; writes update anchor.
  - Counter advances for read AND write of all chunks under salt.
  - `cache_salt` isolates both counter and chunk identity.
  - Cold sighting under near-zero sample rate is dropped.
  - TP fanout deduped on both reads and writes.
  - `_reuse_track` cap enforced via random eviction.
- [x] Full `tests/v1/mp_observability/` suite: **272 passed, 1 skipped**.
- [x] `L1LifecycleSubscriber` tests still pass (untouched).

## If applicable

- [x] User-facing changes — docs added in `observability.rst`.
- [x] Unit tests added.

---

> [!NOTE]
> **Medium Risk**
> Adds new event-driven metric state tracking (including an in-memory per-chunk map and per-tenant counters), which could affect memory/CPU overhead or metric accuracy under high event rates, but is isolated to observability and covered by unit tests.
> 
> **Overview**
> Adds a new MP-mode metrics subscriber, `SMLifecycleSubscriber`, that emits two per-tenant (`cache_salt`) “real reuse” histograms from `SM_READ_PREFETCHED_FINISHED`/`SM_WRITE_FINISHED` events: wall-clock reuse gap (`lmcache_mp.real_reuse_gap_seconds`) and access-count reuse gap (`lmcache_mp.real_reuse_gap_chunks`).
> 
> The subscriber is registered in observability initialization, includes deterministic sampling plus a bounded in-memory tracking map with overflow eviction, and has a new unit test suite covering read/write transitions, sampling behavior, TP fanout dedupe, and cap enforcement. Docs are updated in `METRICS.md` and the user-facing observability page to describe the new metrics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f93e9bf4e855142e286f5c2001473cf76eaec20a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/design/v1/mp_observability/METRICS.md
+++ b/docs/design/v1/mp_observability/METRICS.md
@@ -139,6 +139,8 @@ contribute to histograms; counters above always count all events.
 | `lmcache_mp.l1_chunk_idle_before_evict_seconds` | `lmcache_mp_l1_chunk_idle_before_evict_seconds` | Histogram | `L1_KEYS_EVICTED` | `eviction_time - last_access_time` per sampled chunk |
 | `lmcache_mp.l1_chunk_reuse_gap_seconds` | `lmcache_mp_l1_chunk_reuse_gap_seconds` | Histogram | `L1_READ_FINISHED`, `L1_WRITE_FINISHED`, `L1_WRITE_FINISHED_AND_READ_RESERVED` | Time gap between consecutive touches of the same chunk |
 | `lmcache_mp.l1_chunk_evict_reuse_gap_seconds` | `lmcache_mp_l1_chunk_evict_reuse_gap_seconds` | Histogram | `L1_KEYS_EVICTED` → `L1_WRITE_FINISHED` | Time from eviction to next reuse (capped at 300 s) |
+| `lmcache_mp.real_reuse_gap_seconds` | `lmcache_mp_real_reuse_gap_seconds` | Histogram (tagged `cache_salt`) | `SM_READ_PREFETCHED_FINISHED`, `SM_WRITE_FINISHED` | Time gap between a chunk's last access (read or write) and the next read.  Captures **storage cost**.  Emitted only on read events. |
+| `lmcache_mp.real_reuse_gap_chunks` | `lmcache_mp_real_reuse_gap_chunks` | Histogram (tagged `cache_salt`) | `SM_READ_PREFETCHED_FINISHED`, `SM_WRITE_FINISHED` | Per-`cache_salt` access-counter gap between two reads of the same chunk.  Counter bumps on every read and write of every chunk; histogram emitted only on read events for sampled chunks.  Captures **storage volume**. |
 
 **What it answers:** How long do L1 chunks live? How idle are they before eviction? How quickly are evicted chunks reused?
 
--- a/lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
+++ b/lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
@@ -63,9 +63,7 @@ class SMLifecycleSubscriber(EventSubscriber):
 
     def __init__(self, sample_rate: float = 0.01) -> None:
         if not 0 < sample_rate <= 1.0:
-            raise ValueError(
-                f"sample_rate must be in (0, 1.0], got {sample_rate}"
-            )
+            raise ValueError(f"sample_rate must be in (0, 1.0], got {sample_rate}")
         self._sample_rate = sample_rate
         self._sample_prime = 1_000_003
         self._sample_threshold = int(sample_rate * self._sample_prime)
@@ -144,9 +142,7 @@ class SMLifecycleSubscriber(EventSubscriber):
                 last_time, last_counter = prior
                 attrs = {"cache_salt": cache_salt}
                 self._real_reuse_gap_seconds_hist.record(now - last_time, attrs)
-                self._real_reuse_gap_chunks_hist.record(
-                    counter - last_counter, attrs
-                )
+                self._real_reuse_gap_chunks_hist.record(counter - last_counter, attrs)
                 self._reuse_track[track_key] = (now, counter)
             elif self._should_sample(track_key):
                 self._admit_track(track_key, now, counter)
--- a/lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
+++ b/lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
@@ -24,7 +24,6 @@ access, sampled or not, so the chunks-gap reflects true volume.
 from __future__ import annotations
 
 # Standard
-from typing import Any
 import random
 import time
 
@@ -89,7 +88,7 @@ class SMLifecycleSubscriber(EventSubscriber):
         # is naturally bounded.
         self._salt_chunk_counter: dict[str, int] = {}
         # (cache_salt, chunk_hash) -> (last_access_time, counter_at_last_access).
-        self._reuse_track: dict[tuple[str, Any], tuple[float, int]] = {}
+        self._reuse_track: dict[tuple[str, bytes], tuple[float, int]] = {}
 
     def get_subscriptions(self) -> dict[EventType, EventCallback]:
         return {
@@ -105,7 +104,7 @@ class SMLifecycleSubscriber(EventSubscriber):
 
     def _admit_track(
         self,
-        track_key: tuple[str, Any],
+        track_key: tuple[str, bytes],
         when: float,
         counter: int,
     ) -> None:
@@ -118,7 +117,7 @@ class SMLifecycleSubscriber(EventSubscriber):
             self._reuse_track.pop(victim, None)
         self._reuse_track[track_key] = (when, counter)
 
-    def _should_sample(self, key: object) -> bool:
+    def _should_sample(self, key: tuple[str, bytes]) -> bool:
         return hash(key) % self._sample_prime < self._sample_threshold
 
     def _on_sm_read_finished(self, event: Event) -> None:
@@ -126,7 +125,7 @@ class SMLifecycleSubscriber(EventSubscriber):
         now = event.timestamp or time.time()
         # Dedupe TP fanout: one ObjectKey per kv_rank for the same logical
         # chunk should bump the per-salt counter once.
-        seen: set[tuple[str, Any]] = set()
+        seen: set[tuple[str, bytes]] = set()
         for key in event.metadata.get("succeeded_keys", []):
             chunk_hash = getattr(key, "chunk_hash", None)
             if chunk_hash is None:
@@ -150,7 +149,7 @@ class SMLifecycleSubscriber(EventSubscriber):
     def _on_sm_write_finished(self, event: Event) -> None:
         """Bump counter and update anchor; never emit a sample on write."""
         now = event.timestamp or time.time()
-        seen: set[tuple[str, Any]] = set()
+        seen: set[tuple[str, bytes]] = set()
         for key in event.metadata.get("succeeded_keys", []):
             chunk_hash = getattr(key, "chunk_hash", None)
             if chunk_hash is None:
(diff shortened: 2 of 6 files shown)


docs/design/v1/mp_observability/METRICS.md
```diff
@@ -117,8 +117,10 @@ contribute to histograms; counters above always count all events.
...
+| `lmcache_mp.l1_num_chunk_reuse` | `lmcache_mp_l1_num_chunk_reuse` | Histogram (tagged `cache_salt`) | `L1_KEYS_EVICTED` (incremented on `SM_READ_PREFETCHED.succeeded_keys`) | Actual reuse count per chunk — number of lookups whose `succeeded_keys` (contiguous L1-hit prefix) contained this chunk during its L1 residency.  Subject to per-chunk sampling (`metrics_sample_rate`). |
+| `lmcache_mp.chunk_ideal_reuse_distance` | `lmcache_mp_chunk_ideal_reuse_distance` | Histogram | `SM_READ_PREFETCHED` | Mattson LRU stack reuse distance per sampled lookup access.  SHARDS-style sampled exact stack: hash-sampled at `metrics_sample_rate`, exact LRU within the sampled set, bounded by `reuse_distance_stack_size`.  Buckets `[0, cap)` are exact distances; the `cap` bucket is a censored sample emitted when a key that had been reused at least once is evicted from the stack (next reuse, if any, fell outside the observation window).  Cold-only keys (admitted but never reused) are not recorded.  Workload-level metric — independent of any cache tier. |
```
ApostaC: Why is the source event for `chunk_ideal_reuse_distance` from `SM_READ_PREFETCHED`? IIUC, it should be calculated based on the lookup requests (or maybe chunk allocation event, in order to avoid the impact of vllm's prefix caching).

docs/source/mp/configuration.rst
```diff
@@ -343,6 +343,11 @@ logging, tracing).
        Resource attribute ``service.instance.id`` on every metric and
        span. When the flag is not passed, defaults to a random UUID v4.
        Pass ``--service-instance-id=""`` to force an empty value.
+   * - ``--reuse-distance-stack-size``
+     - ``10000``
+     - Capacity of the bounded LRU stack used to compute
+       ``lmcache_mp.chunk_ideal_reuse_distance``. Set this to
+       approximate the L1 chunk capacity.
```
ApostaC: Another proposal: we don't need to expose this to users. 
My thoughts:
- This argument is mainly for the internal module to avoid memory leak and have a cap on the max memory usage.
- The effect of changing this value could be a bit confusing to users.
- We can have an internal constant (like 100K), which probably just takes <1MB of memory.

lmcache/v1/mp_observability/subscribers/metrics/l1_lifecycle.py
```diff
@@ -91,20 +127,112 @@ def __init__(
...
+                "lookup access (Mattson stack distance over the "
+                "sampled-key subset, bounded by "
+                "reuse_distance_stack_size).  Distances in [0, cap) "
+                "are exact; the cap bucket is a censored sample for "
+                "keys reused at least once whose later access fell "
+                "outside the observation window.  Cold-only keys "
+                "(never reused before eviction) are not recorded.  "
+                "Workload-level metric — independent of any cache tier."
+            ),
+        )
```
ApostaC: For `num_chunk_reuse`, we should use `SM_READ_PREFETCHED` event, but for `chunk_ideal_reuse_distance`, let's use `MP_LOOKUP` (see https://github.com/LMCache/LMCache/blob/df43922fe5e329be63f5d24f25395ce5252d5d92/docs/design/v1/mp_observability/EVENTS.md?plain=1#L83)

lmcache/v1/mp_observability/subscribers/metrics/l1_lifecycle.py
```diff
@@ -91,20 +127,112 @@ def __init__(
...
+          - If the key passes the sample decision, update the bounded
+            LRU stack and emit a reuse-distance sample (skipping cold
+            misses).  Workload signal — independent of L1 state.
+        """
+        for key in event.metadata.get("succeeded_keys", []):
+            self._record_access(key, succeeded=True)
+        for key in event.metadata.get("failed_keys", []):
+            self._record_access(key, succeeded=False)
+
+    def _record_access(self, key: Any, succeeded: bool) -> None:
```
ApostaC: let's avoid using `Any` here. We should have a concrete type definition

lmcache/v1/mp_observability/subscribers/metrics/l1_lifecycle.py
```diff
@@ -91,20 +127,112 @@ def __init__(
...
+    def _stack_distance(self, key: Any) -> int:
+        """Return the LRU stack distance of *key* (number of distinct
+        keys above it in the stack).  Caller must ensure *key* is in
+        the stack.  Top of stack = most-recently used = last item.
+        """
+        distance = 0
+        for k in reversed(self._lru_stack):
+            if k == key:
+                return distance
+            distance += 1
```
ApostaC: This is an O(n) operation. Can we optimize it to O(1)?
Actually, do we need to have LRU at all? I had an impression that we can use random eviction on the internal dictionary.

ApostaC (changes requested):
Thanks for the PR! Please see the comment below.

docs/design/v1/mp_observability/METRICS.md
ruizhang0101: The main reason of using `SM_READ_PREFETCHED` is that it get object keys from IPC key and the object keys has cache_salt, the `MP_LOOKUP` only has original chunk hash which can't differentiate between tenants. Meanwhile, they both got emitted from lookup func so they should both before vllm cache hit IIUC. I can modify the `MP_LOOKUP` event metadata to include cache_salt.

lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+        when: float,
+        counter: int,
+    ) -> None:
+        """Admit a newly-sampled entry; random-evict on overflow."""
+        if (
+            len(self._reuse_track) >= _REUSE_TRACK_CAP
+            and track_key not in self._reuse_track
+        ):
+            victim = random.choice(list(self._reuse_track))
+            self._reuse_track.pop(victim, None)
```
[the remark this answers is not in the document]
ruizhang0101: We want random choice to prevent distribution drift

ruizhang0101:
@ApostaC PTAL when you have time.

lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+            counter = self._bump_salt_counter(cache_salt)
+            prior = self._reuse_track.get(track_key)
+            if prior is not None:
+                last_time, last_counter = prior
+                attrs = {"cache_salt": cache_salt}
+                self._real_reuse_gap_seconds_hist.record(now - last_time, attrs)
+                self._real_reuse_gap_chunks_hist.record(counter - last_counter, attrs)
+                self._reuse_track[track_key] = (now, counter)
+            elif self._should_sample(track_key):
+                self._admit_track(track_key, now, counter)
```
ApostaC, line 147: We do have a clear type for the chunk hash which is `bytes` (see https://github.com/LMCache/LMCache/blob/71e05949ca4535168854d2e3926b31252af67479/lmcache/v1/distributed/api.py#L28)

Let's avoid using `Any` in the type annotation.

lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
```diff
@@ -0,0 +1,167 @@
...
+        """Admit a newly-sampled entry; random-evict on overflow."""
+        if (
+            len(self._reuse_track) >= _REUSE_TRACK_CAP
+            and track_key not in self._reuse_track
+        ):
+            victim = random.choice(list(self._reuse_track))
+            self._reuse_track.pop(victim, None)
+        self._reuse_track[track_key] = (when, counter)
+
+    def _should_sample(self, key: object) -> bool:
```
ApostaC: Let's also avoid using `object` in type annotation, since it does not provide any extra information. Let's use the concrete type here.

ApostaC (commented):
Just nit comments, otherwise LGTM. Thanks!

lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
ruizhang0101: Done

lmcache/v1/mp_observability/subscribers/metrics/sm_lifecycle.py
ruizhang0101: Done

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3123, opened 2026-04-23
[CI] fix egress block and invalid PEP 440 version from variant tags 

deng451e:
Two CI failures on the v0.4.4 release branch:

  1. Docker build blocked — curl installing uv redirects from astral.sh to releases.astral.sh, which was missing from the harden-runner egress
  allowlist in publish.yml.
  2. Invalid PEP 440 version — tags like v0.4.4-cu13 were surviving the "Remove non-release tags" step because the old filter (grep -v '^v') only
  stripped non-v tags. setuptools_scm then tried to parse v0.4.4-cu13 as a version and crashed. Fixed by tightening the filter to grep -vE
  '^v[0-9]+\.[0-9]+\.[0-9]+$', which strips anything that isn't a clean vX.Y.Z.


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3137, opened 2026-04-24
[MP][Feat] Add IsolatedLRU eviction policy + per-cache_salt quotas

royyhuang:
## Summary

Adds multi-tenant quota support to the L2 eviction path. A single `cache_salt` (i.e. a user) cannot evict another salt's cached data under the new `IsolatedLRU` policy; quotas are configured dynamically via HTTP, with no restart needed.

Stacks on #3045 (already merged) which added `AdapterUsage.bytes_by_cache_salt` byte accounting.

## What it adds

- **`IsolatedLRUEvictionPolicy`** — one LRU list per `cache_salt`. `support_isolation=True` routes the `L2EvictionController` to the per-bucket eviction branch; `get_eviction_actions(cache_salt=...)` scopes victim selection to a single bucket.
- **`QuotaManager`** — thread-safe `{cache_salt: limit_bytes}` registry. Shared by the controller (reader) and the HTTP endpoints (CRUD). Allowlist semantics: unregistered salts have effective limit 0 and are wiped on the next eviction cycle.
- **`L2EvictionController`** — dispatches on `policy.support_isolation`. Per-salt branch compares `user_bytes` vs `watermark * limit` for every salt with non-zero bytes; unregistered salts always land over-budget (`limit=0`) and get a full-ratio eviction.
- **HTTP quota endpoints** (`http_apis/quota_api.py`):
  - `PUT/GET/DELETE /api/quota/{cache_salt}`
  - `GET /api/quota`
  - `_default` URL sentinel maps to the empty-string salt.
  - `math.isfinite` rejects NaN / Infinity before they propagate to `int()`.
- **`StorageManager`** exposes `quota_manager` and `get_usage_bytes_by_cache_salt()` (aggregated across adapters). Adapter filter allows `IsolatedLRU` on adapters without `supports_global_eviction` since the per-salt branch operates on the base class's byte counts regardless of adapter capacity.

## Usage

```bash
# Enable the policy:
--eviction-policy IsolatedLRU

# Configure a 2GB quota for user 'alice':
curl -X PUT http://HOST:PORT/api/quota/alice \
     -H 'Content-Type: application/json' \
     -d '{"limit_gb": 2.0}'

# List all quotas:
curl http://HOST:PORT/api/quota
```

## Tests

- `test_quota_manager.py` — CRUD + thread-safety (concurrent writes, read-write race, set/delete race on same key).
- `test_isolated_lru_eviction_policy.py` — per-salt bucketing, scoped vs global eviction, locked-key filter.
- `test_cache_salt_l2_eviction.py` — end-to-end with a real adapter: over-budget salt evicted, unregistered salt wiped, other salts untouched.
- `test_http_quota_endpoints.py` — TestClient-driven CRUD + validation paths (non-numeric / negative / NaN / inf / missing body) + `_default` sentinel round-trip.

All 880 tests in `tests/v1/distributed/` + `tests/v1/multiprocess/` pass locally; pre-commit clean.

## Drive-by

Two test names / docstrings in the merged PR #3045 test files renamed from `per_user` → `by_cache_salt` / `cache_salt` for terminology consistency with this feature. Happy to split out if preferred.

## Test plan

- [ ] CI green on `tests/v1/distributed/` + `tests/v1/multiprocess/`
- [ ] Smoke test: start MP server with `--eviction-policy IsolatedLRU`, set quota for a salt via HTTP, exceed it, observe eviction scoped to that salt
- [ ] Verify `_default` sentinel round-trips for un-salted traffic

---

> [!NOTE]
> **Medium Risk**
> Changes L2 eviction behavior by adding quota-driven, per-`cache_salt` eviction (including allowlist semantics where unregistered salts are fully evicted), which can impact cache retention and eviction rates in production. Also introduces new HTTP endpoints for runtime quota CRUD that must be deployed/secured correctly.
> 
> **Overview**
> Adds a new `IsolatedLRU` eviction mode that keeps separate LRU lists per `cache_salt` and updates config/CLI parsing and the eviction-policy factory to support selecting it.
> 
> Extends `L2EvictionController` to dispatch to a new quota-driven per-salt eviction path when the policy supports isolation, using a new thread-safe `QuotaManager` registry (unregistered salts default to an effective quota of 0 and are wiped on the next eviction cycle).
> 
> Exposes quota administration via new FastAPI endpoints under `/api/quota` (PUT/GET/DELETE + list) and wires `StorageManager` to always create/expose a shared `quota_manager` plus an aggregate `get_usage_bytes_by_cache_salt()` helper. Adds unit/integration tests covering the new policy, quota manager semantics/thread-safety, controller behavior, and HTTP validation paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c37b549754b4076b29c61765f41dea3df13cb59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/distributed/eviction_policy/isolated_lru.py
+++ b/lmcache/v1/distributed/eviction_policy/isolated_lru.py
@@ -109,7 +109,13 @@ class IsolatedLRUEvictionPolicy(EvictionPolicy):
         key_eligible_filter: Callable[[ObjectKey], bool] | None = None,
         cache_salt: str | None = None,
     ) -> list[EvictionAction]:
-        """Select victims in LRU order, optionally scoped to one salt.
+        """Select victims in LRU order, scoped to a single ``cache_salt``.
+
+        IsolatedLRU is "isolated only" by contract — callers must pass a
+        concrete ``cache_salt``. The base interface keeps ``cache_salt``
+        optional for compatibility with non-isolated policies; passing
+        ``None`` here raises ``ValueError`` rather than silently
+        falling back to a global pool.
 
         Args:
             expected_ratio: Fraction of the candidate pool to evict,
@@ -118,30 +124,26 @@ class IsolatedLRUEvictionPolicy(EvictionPolicy):
                 returned (matches ``LRUEvictionPolicy``).
             key_eligible_filter: Optional predicate — keys failing the
                 filter (e.g. locked) are skipped.
-            cache_salt: When set, only this salt's bucket is considered.
-                When ``None``, victims are taken across all buckets.
+            cache_salt: The salt whose bucket to evict from. Required.
 
         Returns:
-            A single ``EvictionAction`` covering all selected keys, or
-            an empty list if nothing is available to evict.
+            A list with at most one ``EvictionAction`` (one per
+            destination — IsolatedLRU has a single destination, so the
+            list is either empty or length 1). Empty when nothing is
+            available to evict under the current filter / ratio.
+
+        Raises:
+            ValueError: If ``cache_salt`` is ``None``.
         """
+        if cache_salt is None:
+            raise ValueError(
+                "IsolatedLRUEvictionPolicy.get_eviction_actions requires "
+                "cache_salt; this policy is per-bucket and has no global "
+                "eviction path."
+            )
         with self._lock:
-            if cache_salt is not None:
-                order = self._per_salt_order.get(cache_salt)
-                pool = list(order.keys()) if order else []
-            else:
-                # Global eviction: concatenate every bucket in
-                # dict-insertion order (i.e. first-seen-salt first).
-                # This is deterministic but NOT "oldest key first across
-                # all salts" — salts that happened to see traffic early
-                # get evicted first even if their current front is
-                # newer than another salt's. The isolated controller
-                # branch doesn't hit this path (always passes an
-                # explicit salt); global eviction here is a fallback
-                # for callers that explicitly opt in.
-                pool = []
-                for order in self._per_salt_order.values():
-                    pool.extend(order.keys())
+            order = self._per_salt_order.get(cache_salt)
+            pool = list(order.keys()) if order else []
 
             if not pool:
                 return []
--- a/lmcache/v1/distributed/eviction_policy/isolated_lru.py
+++ b/lmcache/v1/distributed/eviction_policy/isolated_lru.py
@@ -1,0 +1,189 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+IsolatedLRU (per-``cache_salt`` LRU) eviction policy.
+
+Unlike :class:`LRUEvictionPolicy` which maintains a single global LRU
+list, this policy keeps one LRU list per ``cache_salt`` and uses
+``support_isolation=True`` so the L2 eviction controller can scope
+eviction to a specific salt via the ``cache_salt`` argument to
+``get_eviction_actions``.
+
+Combined with :class:`QuotaManager`, this lets the controller evict
+only the over-quota ``cache_salt``'s keys while leaving other salts'
+cached data untouched.
+"""
+
+# Future
+from __future__ import annotations
+
+# Standard
+from collections import OrderedDict
+from collections.abc import Callable
+import threading
+
+# First Party
+from lmcache.v1.distributed.api import ObjectKey
+from lmcache.v1.distributed.eviction import EvictionPolicy
+from lmcache.v1.distributed.internal_api import (
+    EvictionAction,
+    EvictionDestination,
+)
+
+
+class IsolatedLRUEvictionPolicy(EvictionPolicy):
+    """Per-``cache_salt`` LRU eviction policy.
+
+    Maintains one ``OrderedDict`` per ``cache_salt`` (keyed by
+    ``ObjectKey``). New and touched keys move to the end of their
+    bucket's ordering; least-recently-used keys stay at the front and
+    are evicted first.
+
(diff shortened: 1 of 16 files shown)


sammshen (approved):
LGTM

lmcache/v1/distributed/eviction_policy/isolated_lru.py
```diff
@@ -0,0 +1,189 @@
...
+                zero and the pool is non-empty, at least one key is
+                returned (matches ``LRUEvictionPolicy``).
+            key_eligible_filter: Optional predicate — keys failing the
+                filter (e.g. locked) are skipped.
+            cache_salt: When set, only this salt's bucket is considered.
+                When ``None``, victims are taken across all buckets.
+
+        Returns:
+            A single ``EvictionAction`` covering all selected keys, or
+            an empty list if nothing is available to evict.
```
ApostaC: nit: the docstring is not consistent with the real return value (`list[EvictionAction]`)

lmcache/v1/distributed/eviction_policy/isolated_lru.py
```diff
@@ -0,0 +1,189 @@
...
+                    # Drop the empty bucket so ``list``/iteration
+                    # snapshots stay compact.
+                    del self._per_salt_order[key.cache_salt]
+
+    def get_eviction_actions(
+        self,
+        expected_ratio: float,
+        key_eligible_filter: Callable[[ObjectKey], bool] | None = None,
+        cache_salt: str | None = None,
+    ) -> list[EvictionAction]:
```
ApostaC, line 111: Maybe I'm wrong, but it seems like we are mixing the global eviction and the isolated eviction, right?
Do we want to make it purely isolated instead of sharing the same `get_eviction_actions` interface? (e.g., `get_eviction_actions_by_cache_salt`)

lmcache/v1/distributed/eviction_policy/isolated_lru.py
ApostaC, line 1: Quick question: is it possible to implement this as a wrapper around non-isolated eviction logics?

lmcache/v1/distributed/eviction_policy/isolated_lru.py
ApostaC: After checking the base class's interface, each element in this list will correspond to a specific destination.

lmcache/v1/distributed/quota_manager.py
```diff
@@ -0,0 +1,100 @@
...
+from dataclasses import dataclass
+import threading
+
+
+@dataclass(frozen=True)
+class QuotaEntry:
+    """Snapshot of a single quota registration."""
+
+    cache_salt: str
+    limit_bytes: int
```
ApostaC: This can be put into `internal_api.py` to avoid future circular import.

lmcache/v1/distributed/storage_controllers/eviction_controller.py
```diff
@@ -241,6 +267,49 @@ def _check_and_evict(self, state: L2AdapterEvictionState):
...
+                user_bytes,
+                limit,
+                watermark,
+                effective_ratio,
+            )
+            actions = state.eviction_policy.get_eviction_actions(
+                effective_ratio, cache_salt=cache_salt
+            )
+            for action in actions:
+                self._execute_eviction_action(state.adapter, action)
```
ApostaC: A potential optimization: we can merge the actions for all the cache salts together to reduce the potential number of I/O operations.
Conceptually, a single eviction action is a mapping from a single eviction destination.
After running the loop, we should have the list of
```
cache_salt -> [eviction_destination -> keys]
```
And we can do shuffule & merge to get the list of
```
eviction_destination -> keys_for_all_the_cache_salts
```

lmcache/v1/distributed/storage_manager.py
```diff
@@ -65,26 +66,42 @@ def __init__(self, config: StorageManagerConfig):
...
+        # policies (``IsolatedLRU``) operate on per cache_salt byte
+        # counts which the base class tracks regardless of capacity,
+        # so they are wired up unconditionally.
+        #
+        # CAVEAT: IsolatedLRU on an adapter whose ``_notify_keys_stored``
+        # never fires (or whose ``delete()`` is a no-op) will log
+        # "triggering eviction" each cycle but have no observable
+        # effect. The FS adapter in particular is both unbounded and
+        # delete-less; configuring IsolatedLRU there is allowed but
+        # produces zero eviction activity.
```
ApostaC: Probably we don't need this -- just to keep the code clean

ApostaC (changes requested):
Comments are mainly about `isolated_lru.py`'s implementation. Please take a look, thanks!

lmcache/v1/distributed/eviction_policy/isolated_lru.py
royyhuang: I feel changing to `get_eviction_actions_by_cache_salt` does not really helping anything. We will still use `support_isloation` to check which function to call. Since it won't help, I personally would prefer to have a more general interface.

lmcache/v1/distributed/eviction_policy/isolated_lru.py
royyhuang: I feel maybe we can defer this until we have another isolated policy?

lmcache/v1/distributed/quota_manager.py
royyhuang: done.

ApostaC (approved):
LGTM! We can refactor the isolated eviction policy later.

PR #3142, opened 2026-04-25
[MP]: Use unified c_ops backend import and fix gpu_kv_format_name property

maobaolong:

This PR applies two small fixes to `lmcache/v1/multiprocess/gpu_context.py`:

1. Use the unified backend selection from `lmcache/__init__.py` by importing
   `lmcache.c_ops` unconditionally. The CUDA vs pure-Python fallback is
   already aliased into `sys.modules` by `_get_backend`, so the previous
   `if torch.cuda.is_available():` guard is redundant and makes the CPU-only
   path diverge from the rest of the codebase.
2. Decorate `GPUCacheContext.gpu_kv_format_name` with `@property` so callers
   (e.g. server metadata accessors) receive the enum name string directly,
   consistent with the other metadata accessors on `GPUCacheContext`.

⚠️ **Breaking change**: `GPUCacheContext.gpu_kv_format_name` is changed from a
method to a `@property`. Any existing caller that invokes it as
`ctx.gpu_kv_format_name()` must be updated to access it as
`ctx.gpu_kv_format_name`. A quick search in this repo shows no such callers
today, but downstream users of this public interface should update
accordingly.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/gpu_context.py
+++ b/lmcache/v1/multiprocess/gpu_context.py
@@ -41,7 +41,7 @@ from lmcache.v1.multiprocess.custom_types import (
 # fallback) is handled once in ``lmcache/__init__.py`` via ``_get_backend``,
 # which aliases the chosen module as ``lmcache.c_ops`` in ``sys.modules``.
 # Importing it here transparently works in both CUDA and CPU-only envs.
-import lmcache.c_ops as lmc_ops  # noqa: E402
+import lmcache.c_ops as lmc_ops
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


maobaolong:
@hlin99 @ApostaC @sammshen How about this simple fix? This PR just fix the gpu_context.py to use unified backend selection.

hlin99:
> @hlin99 @ApostaC @sammshen How about this simple fix? This PR just fix the gpu_context.py to use unified backend selection.

hi @maobaolong  this LGTM！thanks！

hlin99 (approved):

ApostaC (approved):
LGTM!

chunxiaozheng (approved):
LGTM

PR #3145, opened 2026-04-25
[Core] Add hidden state support

Vivo50E:
### Problem

When LMCache restores KV cache from CPU, the serving engine skips prefill for cached tokens. However, some downstream consumers (e.g., multi-stage pipelines, cross-attention layers) need the **full-sequence hidden states** from the prefill, not just KV. Without HS restore, these consumers receive partial hidden states and produce incorrect output.

### Solution

Add general-purpose hidden state store/retrieve to LMCache, using the same chunk-based token hash mechanism as KV:

- **`CacheEngineKey.to_hs_key()`** — derives a parallel key with `:hs` model_name suffix so HS entries don't collide with KV entries but share the same `chunk_hash`
- **`MemoryFormat.HS_TD`** — new memory format for `[num_tokens, hidden_dim]` tensors, routed through the pinned memory allocator like existing KV formats
- **`LMCacheEngine.store_hidden_states()`** — stores HS chunks to CPU storage using the same `token_database.process_tokens()` chunking as KV (guarantees identical chunk boundaries)
- **`LMCacheEngine.retrieve_hidden_states()`** — retrieves HS chunks and assembles into a contiguous CPU tensor
- **`_enable_hs_offload` config flag** — opt-in via `kv_connector_extra_config["lmcache.enable_hidden_state_offload"]` so vanilla vLLM is completely unaffected

### Key Design Decisions

- **HS bypasses GPU connector entirely** — HS tensors are already on CPU, so no `multi_layer_kv_transfer` or CUDA kernel needed
- **Parallel keys, not modified MemoryObj** — HS uses separate `CacheEngineKey` entries (`:hs` suffix) rather than extending the KV MemoryObj format, keeping backward compatibility
- **Same chunk boundaries as KV** — `store_hidden_states` reuses `token_database.process_tokens()`, so HS and KV chunks are always aligned
- **Opt-in flag** — `_enable_hs_offload` defaults to `False`. No impact on existing users

### Changes

| File | Change |
|---|---|
| `lmcache/utils.py` | `CacheEngineKey.to_hs_key()` method |
| `lmcache/v1/memory_management.py` | `MemoryFormat.HS_TD` enum + allocator/free support |
| `lmcache/v1/cache_engine.py` | `store_hidden_states()` and `retrieve_hidden_states()` |
| `lmcache/integration/vllm/vllm_v1_adapter.py` | `_enable_hs_offload` config flag |
| `tests/v1/test_hidden_state_offload.py` | Unit tests for key derivation and store/retrieve roundtrip |

### Related

- Companion PR: vllm-project/vllm-omni#1330 (wires store/restore from the model runner side)

### Testing

- Unit tests: key derivation (distinct, idempotent, preserves configs, no KV collision), store/retrieve roundtrip (single chunk, multi-chunk), missing returns None
- E2E verified with vllm-omni (Qwen3-Omni-30B, 2×A100, 20K input tokens): both `prefix_caching=true` and `prefix_caching=false` paths produce correct full-sequence HS after KV restore

---

> [!NOTE]
> **Medium Risk**
> Adds new storage paths and memory formats (hidden-state offload and configurable multimodal save masking) that affect cache persistence and retrieval behavior; mistakes could cause cache misses or increased memory pressure but changes are largely additive and opt-in.
> 
> **Overview**
> **Adds hidden-state offload support alongside KV cache.** Introduces `CacheEngineKey.to_hs_key()` and a new `MemoryFormat.HS_TD`, plus `LMCacheEngine.store_hidden_states()`/`retrieve_hidden_states()` to chunk, store, and reconstruct per-token hidden states using the same token-hash boundaries as KV while keeping HS entries separate from KV keys.
> 
> **Improves vLLM integration controls around what gets stored.** Adds `skip_mm_storage` config and plumbing through `ReqMeta`/`SaveSpec` to cap KV storage to the text-only prefix before multimodal tokens (without shrinking the load path), and extends `StorageManager.get()` to accept a backend `search_range` for retrieve-location-aware HS fetches. Includes new unit tests covering HS roundtrips/key separation and `skip_mm_storage` masking behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 943b49312e7c839836350641c8ebaa9b0c192642. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -569,7 +569,7 @@ class LMCacheEngine:
         tokens: Union[torch.Tensor, list[int]],
         mask: Optional[torch.Tensor] = None,
         hidden_states: Optional[torch.Tensor] = None,
-        **kwargs,
+        request_configs: Optional[dict] = None,
     ) -> None:
         """Store hidden states alongside KV cache using the same chunk keys.
 
@@ -582,7 +582,8 @@ class LMCacheEngine:
             mask: Boolean mask (same semantics as ``store()``).
             hidden_states: CPU tensor of shape ``[num_tokens, hidden_dim]``.
                 If ``None``, this is a no-op.
-            **kwargs: Passed to ``token_database.process_tokens()``.
+            request_configs: Forwarded to ``token_database.process_tokens()``
+                for tag-aware chunk hashing.
         """
         if hidden_states is None:
             return
@@ -594,8 +595,6 @@ class LMCacheEngine:
         if self._is_passive():
             logger.debug("rank=%s ignore HS store", self.metadata.worker_id)
             return
-
-        request_configs = kwargs.get("request_configs")
         hs_keys: list[CacheEngineKey] = []
         hs_objs: list[MemoryObj] = []
         partial = False
@@ -606,6 +605,12 @@ class LMCacheEngine:
             request_configs=request_configs,
         ):
             hs_key = key.to_hs_key()
+            # Skip chunks already in storage (mirrors KV store()).  Concurrent
+            # requests sharing a token prefix would otherwise redundantly
+            # allocate, copy, and put identical entries.
+            if self.storage_manager.contains(hs_key, self.retrieve_locations):
+                continue
+
             hs_chunk = hidden_states[start:end].contiguous()
             hs_shape = torch.Size(list(hs_chunk.shape))
             hs_dtype = hs_chunk.dtype
@@ -646,14 +651,15 @@ class LMCacheEngine:
         self,
         tokens: Union[torch.Tensor, list[int]],
         mask: Optional[torch.Tensor] = None,
-        **kwargs,
+        request_configs: Optional[dict] = None,
     ) -> Optional[torch.Tensor]:
         """Retrieve hidden states stored alongside KV cache.
 
         Args:
             tokens: Token IDs for chunk boundary computation.
             mask: Boolean mask (same semantics as ``retrieve()``).
-            **kwargs: Passed to ``token_database.process_tokens()``.
+            request_configs: Forwarded to ``token_database.process_tokens()``
+                for tag-aware chunk hashing.
 
         Returns:
             A contiguous CPU tensor of shape ``[num_tokens, hidden_dim]``,
@@ -662,7 +668,6 @@ class LMCacheEngine:
         if not self.is_healthy() or self.storage_manager is None:
             return None
 
-        request_configs = kwargs.get("request_configs")
         chunks: list[tuple[int, int, torch.Tensor]] = []
 
         for start, end, key in self.token_database.process_tokens(
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -563,6 +563,106 @@ class LMCacheEngine:
             store_stats.put_time * 1000,
         )
 
+    @torch.inference_mode()
+    def store_hidden_states(
+        self,
+        tokens: Union[torch.Tensor, list[int]],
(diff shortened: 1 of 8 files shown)


tests/v1/test_hidden_state_offload.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    hs_key = key.to_hs_key()
+
+    # Must not be equal
+    assert hs_key != key
+    # Must not collide in dict/set
+    d = {key: "kv", hs_key: "hs"}
+    assert len(d) == 2
+    assert d[key] == "kv"
+    assert d[hs_key] == "hs"
```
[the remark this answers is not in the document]
Vivo50E: store_hidden_states / retrieve_hidden_states require a fully initialized LMCacheEngine (with GPU connector and storage manager), which is not feasible in unit tests.  Those methods are verified via E2E serving benchmarks in the companion vllm-omni PR.

allytotheson:
Looks great overall; just two suggestions to keep the KV path and hidden-state path more clearly separated:

Dedicated hidden-state API: Consider introducing a small class (e.g., hs_store) to own hidden-state put/get, rather than folding it into the main cache engine. This keeps the abstraction cleaner (e.g., engine.hs_store.store_hidden_states(...)).

Separate memory pool: Mixing KV and hidden states in the same pool could lead to fragmentation. It might be worth allocating a separate pool (lazy or pinned, similar to LMCache) specifically for hidden states.

lmcache/v1/cache_engine.py
```diff
@@ -563,6 +563,141 @@ def store(
...
+            request_configs=request_configs,
+        ):
+            hs_key = key.to_hs_key()
+            # Skip chunks already in storage (mirrors KV store()).  Concurrent
+            # requests sharing a token prefix would otherwise redundantly
+            # allocate, copy, and put identical entries.
+            if self.storage_manager.contains(hs_key, self.retrieve_locations):
+                continue
+
+            hs_chunk = hidden_states[start:end].contiguous()
```
sammshen, line 614: @Vivo50E  quick question, why are we calling `.contiguous()` here? so if the slice is happening along the token dimesion, this does not cause non-contiguity so that means the `D` is non contiguous? is `D` here the hidden states across all the layers or only the last layer? 

We want to minimize unecessary copies on LMCache side

lmcache/v1/cache_engine.py
```diff
@@ -563,6 +563,141 @@ def store(
...
+                partial = True
+                break
+            memory_obj.tensor[:] = hs_chunk
+            hs_keys.append(hs_key)
+            hs_objs.append(memory_obj)
+
+        # retrieve_hidden_states is all-or-nothing: any missing chunk causes
+        # the whole retrieve to return None.  Drop a partial allocation
+        # instead of writing orphaned, unreachable entries to storage.
+        if partial:
```
sammshen, line 636: @allytotheson prefix hit

lmcache/v1/cache_engine.py
Vivo50E: @sammshen The .contiguous() here is defensive — while the current caller (gpu_model_runner.py:155) already ensures contiguity via .to("cpu").contiguous(), a future caller could pass a strided/transposed tensor. Slicing along dim 0 alone doesn't cause non-contiguity.

Re D: each call receives one layer's hidden states — shape [num_tokens, hidden_dim]. Layers "0", "24", and "hidden" are stored as separate LMCache entries with distinct keys, not stacked into one tensor.

DongDongJu:
Will iterate this after revision based on sam's suggestion.

Issue #3090, opened 2026-04-21
[MP] LMCache MultiProcess multi platform support


maobaolong:
**Label**
Please label your issue so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/1882)

**Summary**

As motioned in the community sync meeting, i create this PR to track and settle a discussion about this topic.

**Initial Motivation**

- Run LMCache MP UTs on macos successfully.
- Run LMCache MP Server and execute `test-cache cli tools` on macos successfully.
- Run vLLM + LMCache MP Server on cpu only (Macos and ubuntu)

So, developer can easy to iterate the develop state and improve the working code locally without gpu dependencies.

**Further gain**
- Run vLLM + LMCache MP mode on intel NPU/XPU and any other platform or new devices.

**Related PRs**
- https://github.com/LMCache/LMCache/pull/3009
- https://github.com/LMCache/LMCache/pull/3081
- https://github.com/LMCache/LMCache/pull/2908
- https://github.com/LMCache/LMCache/pull/3288/
- https://github.com/LMCache/LMCache/pull/3013
- https://github.com/LMCache/LMCache/pull/2591
- https://github.com/LMCache/LMCache/pull/2822
- https://github.com/LMCache/LMCache/pull/2848
- https://github.com/LMCache/LMCache/pull/3144
- https://github.com/LMCache/LMCache/pull/3091
- https://github.com/LMCache/LMCache/pull/3294 
- https://github.com/LMCache/LMCache/pull/3308
- https://github.com/LMCache/LMCache/pull/3259
- https://github.com/LMCache/LMCache/pull/3280
- https://github.com/vllm-project/vllm/pull/42740
- https://github.com/LMCache/LMCache/pull/3302
- https://github.com/LMCache/LMCache/pull/3148 
- https://github.com/LMCache/LMCache/pull/3359
- https://github.com/LMCache/LMCache/pull/3590
- https://github.com/LMCache/LMCache/pull/3514
- https://github.com/LMCache/LMCache/pull/3528
- https://github.com/LMCache/LMCache/pull/3513
- https://github.com/LMCache/LMCache/pull/3531
- https://github.com/LMCache/LMCache/pull/3563
- https://github.com/LMCache/LMCache/pull/3352
- https://github.com/LMCache/LMCache/pull/3590
- https://github.com/LMCache/LMCache/pull/3642
- https://github.com/LMCache/LMCache/pull/3672
[...]

PR #3148, opened 2026-04-26
Add macos ci check

maobaolong:

The macos ci test passed.

Clipboard_Screenshot_1777202031

**If applicable**:

---

> [!NOTE]
> **Medium Risk**
> Adds early-import monkey patches for `torch.cuda`, `cupy`, and `os.eventfd`, which can affect runtime behavior across platforms if any code relies on real CUDA/eventfd semantics. Also updates packaging/build selection logic, so mis-detection could change which native extensions are built/installed.
> 
> **Overview**
> Adds a new GitHub Actions *macOS compatibility* workflow that installs LMCache in CPU-only mode on both Intel and Apple Silicon (Python 3.11/3.12) and runs a smoke script that starts `lmcache server` and validates the HTTP endpoints.
> 
> Introduces a new `lmcache.v1.platform` shim layer that is imported at `lmcache` package import time to provide cross-platform fallbacks: a pipe-based `os.eventfd` implementation, a fake `cupy` module for CPU-only environments, and no-op stubs for key `torch.cuda` APIs.
> 
> Tightens macOS/CPU-only robustness by catching `AssertionError` during vLLM `NONE_HASH` initialization, and updates `setup.py` so `NO_CUDA_EXT=1` builds CPU-only C++ extensions (instead of treating it like an sdist/no-extension build).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f02c68a105f86ce4d83c17363ea30eea54eb9a3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/scripts/macos_smoke_test.sh
+++ b/.github/scripts/macos_smoke_test.sh
@@ -4,8 +4,11 @@
 # macOS basic-compatibility smoke test for the lmcache multiprocess
 # server. Verifies that:
 #   1) `lmcache --help` works (CLI entry point is importable)
-#   2) `lmcache server` can launch the ZMQ + HTTP server on CPU
-#   3) the HTTP server answers GET / and GET /healthcheck
+#   2) common C++ extensions (native_storage_ops / lmcache_redis /
+#      lmcache_fs) load on macOS, exercising the PipeNotifier fallback
+#      in csrc/storage_backends/event_notifier.h
+#   3) `lmcache server` can launch the ZMQ + HTTP server on CPU
+#   4) the HTTP server answers GET / and GET /healthcheck
 #
 # This script is intentionally minimal — it does not exercise any
 # GPU / CUDA / vLLM code paths. It is meant to be a fast regression
@@ -31,6 +34,21 @@ python -c "import torch; print('torch', torch.__version__, 'cuda', torch.cuda.is
 
 echo "==> Step 1: lmcache CLI help"
 lmcache --help >/dev/null
+
+echo "==> Step 1.5: import common C++ extensions (CPU-only build)"
+# Mirrors the verify step in .github/workflows/build_cpu_artifacts.yml
+# on the macOS axis: makes sure NO_GPU_EXT=1 produced loadable .so's
+# and that c_ops resolves to the python_ops_fallback shim.
+python -c "
+import sys
+import lmcache
+import lmcache.native_storage_ops  # noqa: F401
+import lmcache.lmcache_redis  # noqa: F401
+import lmcache.lmcache_fs  # noqa: F401
+import lmcache.c_ops  # noqa: F401
+assert lmcache.torch_device_type == 'cpu', lmcache.torch_device_type
+assert sys.modules['lmcache.c_ops'].__name__ == 'lmcache.python_ops_fallback'
+"
 
 echo "==> Step 2: launch 'lmcache server' on ${HTTP_HOST}:${HTTP_PORT} (zmq ${ZMQ_PORT})"
 rm -f "${LOG_FILE}"
--- a/.github/scripts/macos_smoke_test.sh
+++ b/.github/scripts/macos_smoke_test.sh
@@ -5,7 +5,7 @@
 # server. Verifies that:
 #   1) `lmcache --help` works (CLI entry point is importable)
 #   2) `lmcache server` can launch the ZMQ + HTTP server on CPU
-#   3) the HTTP server answers GET / and GET /api/healthcheck
+#   3) the HTTP server answers GET / and GET /healthcheck
 #
 # This script is intentionally minimal — it does not exercise any
 # GPU / CUDA / vLLM code paths. It is meant to be a fast regression
--- a/.github/scripts/macos_smoke_test.sh
+++ b/.github/scripts/macos_smoke_test.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# macOS basic-compatibility smoke test for the lmcache multiprocess
+# server. Verifies that:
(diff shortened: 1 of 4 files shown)


hlin99:
hi @maobaolong thanks for the work! do you think it's a good way to add subfolder "mac-os" under platform? so moving forward, we may have diff platforms across OS and HWs, and each can have a unified place to monkey patching functions.

maobaolong:
> hi @maobaolong thanks for the work! do you think it's a good way to add subfolder "mac-os" under platform? so moving forward, we may have diff platforms across OS and HWs, and each can have a unified place to monkey patching functions.

This is a good idea, ideally, we can supply a hardware-independent api layer, meanwhile, we can also supply some implementations for each hardware, E.g. cpu and others.

BTW, we can discuss this topic in the #3090. This PR is just for show that macos ci can work for now.

.github/scripts/macos_smoke_test.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#
+# macOS basic-compatibility smoke test for the lmcache multiprocess
+# server. Verifies that:
+#   1) `lmcache --help` works (CLI entry point is importable)
+#   2) `lmcache server` can launch the ZMQ + HTTP server on CPU
+#   3) the HTTP server answers GET / and GET /api/healthcheck
```
yoo-kumaneko: Do we still have the /api endpoint now?

yoo-kumaneko (approved):
otherwise LGTM

.github/scripts/macos_smoke_test.sh
maobaolong: Good catch!

maobaolong: Done

yoo-kumaneko:
One small thought on test coverage — completely optional, your call whether to act on it.

The new macOS smoke test boots `lmcache server` with the default L1-only config and curls `/` + `/healthcheck`, which verifies imports and startup don't break on macOS. But `ConnectorBase` (and therefore the new `PipeNotifier`) is on the *storage-backend* connector path, which the smoke test never exercises. In other words, if `PipeNotifier::notify()` were silently a no-op, this lane would still go green.

If you wanted to close that gap, two cheap options:

1. A tiny gtest that does `notify()` → `poll(fileno, POLLIN, 100)` → `consume()` round-trip on both notifier impls, or
2. Extending the smoke script to configure a file-backed connector and round-trip a real put/get so the wakeup path actually fires.

Neither is necessary to land this PR — the refactor is small and the Linux path is byte-equivalent to the original. Just flagging it as a follow-up worth considering. Up to you whether to address it here, in a follow-up, or not at all.

maobaolong:
> One small thought on test coverage — completely optional, your call whether to act on it.
> 
> The new macOS smoke test boots `lmcache server` with the default L1-only config and curls `/` + `/healthcheck`, which verifies imports and startup don't break on macOS. But `ConnectorBase` (and therefore the new `PipeNotifier`) is on the _storage-backend_ connector path, which the smoke test never exercises. In other words, if `PipeNotifier::notify()` were silently a no-op, this lane would still go green.
> 
> If you wanted to close that gap, two cheap options:
> 
> 1. A tiny gtest that does `notify()` → `poll(fileno, POLLIN, 100)` → `consume()` round-trip on both notifier impls, or
> 2. Extending the smoke script to configure a file-backed connector and round-trip a real put/get so the wakeup path actually fires.
> 
> Neither is necessary to land this PR — the refactor is small and the Linux path is byte-equivalent to the original. Just flagging it as a follow-up worth considering. Up to you whether to address it here, in a follow-up, or not at all.

@yoo-kumaneko Thanks for this great suggestion, it worth to do for sure, but it cannot be done for this stage, because there are some PRs on the fly. So for this PR, i have to create a smoke test as a baseline for now, but you are right and I will improve this smoke test after everything is done and it can works for lmcache on macos.

hlin99:
hi @maobaolong , overall LGTM. but i have a small suggestion, if the .h file is a mirror to event_notifier.py, can we use pybind11, so that only maintaining the py file is sufficient for both sides? is it doable?

maobaolong:
> hi @maobaolong , overall LGTM. but i have a small suggestion, if the .h file is a mirror to event_notifier.py, can we use pybind11, so that only maintaining the py file is sufficient for both sides? is it doable?

@hlin99 Thanks for your suggestion! I thought about this. The two files look like mirrors but serve independent call sites: the C++ one is invoked from connector worker threads that don't hold the GIL, so going through pybind11 would force a gil_scoped_acquire on every notify() / consume() in the hot path, which defeats the purpose of an eventfd-based wakeup. The header is also a pretty thin RAII wrapper over eventfd(2) / pipe(2) (~100 lines, no business logic), so I'd rather keep them as two small language-native shims of the same OS concept.

hlin99:
> > hi @maobaolong , overall LGTM. but i have a small suggestion, if the .h file is a mirror to event_notifier.py, can we use pybind11, so that only maintaining the py file is sufficient for both sides? is it doable?
> 
> @hlin99 Thanks for your suggestion! I thought about this. The two files look like mirrors but serve independent call sites: the C++ one is invoked from connector worker threads that don't hold the GIL, so going through pybind11 would force a gil_scoped_acquire on every notify() / consume() in the hot path, which defeats the purpose of an eventfd-based wakeup. The header is also a pretty thin RAII wrapper over eventfd(2) / pipe(2) (~100 lines, no business logic), so I'd rather keep them as two small language-native shims of the same OS concept.

thanks for the explanation. it makes sense and LGTM!

hlin99 (approved):

PR #3149, opened 2026-04-26
fix(disk): defer cache-policy hit update until load succeeds in get_blocking()

### LINKED ISSUE #3015: [Bug]: get_blocking() records phantom cache hit when load_bytes_from_disk() fails
## Summary

`LocalDiskBackend.get_blocking()` calls `cache_policy.update_on_hit()` **before** `load_bytes_from_disk()` confirms the data is actually loaded. When `load_bytes_from_disk()` returns `None` (e.g., due to CPU staging pool exhaustion), the cache policy has already recorded a "hit" that never delivered data  a **phantom hit**.

## Affected Code

`lmcache/v1/storage_backend/local_disk_backend.py` lines 418-434:

```python
def get_blocking(self, key):
    self.disk_lock.acquire()
    if key not in self.dict:
        self.disk_lock.release()
        return None

    self.cache_policy.update_on_hit(key, self.dict)  #  hit recorded HERE

    disk_meta = self.dict[key]
    # ...
    self.disk_lock.release()
    memory_obj = self.load_bytes_from_disk(...)       #  but data load may FAIL here

    return memory_obj  # returns None on failure
```

## Impact

- The phantom hit artificially boosts the key's recency/frequency in the cache policy
- This could protect the key from eviction even though it failed to serve
- Under repeated pool exhaustion, popular keys accumulate phantom hits, distorting eviction decisions

## Context

This is a **pre-existing issue** that became more reachable after #2942 / PR #3006. Before that fix, `load_bytes_from_disk()` would either succeed or deadlock (never return `None`), so this code path was effectively unreachable.

## Suggested Fix

Move `update_on_hit()` after `load_bytes_from_disk()`, only if it succeeds:

```python
self.disk_lock.release()
memory_obj = self.load_bytes_from_disk(...)
if memory_obj is not None:
    self.disk_lock.acquire()
    self.cache_policy.update_on_hit(key, self.dict)
    self.disk_lock.release()
return memory_obj
```

Or alternatively, add a rollback on failure.

## Related

- Issue: #2942
- PR: #3006

SuperMarioYL:
## Summary

Fixes #3015.

`LocalDiskBackend.get_blocking()` called `cache_policy.update_on_hit()` **before** `load_bytes_from_disk()` confirmed that data was actually loaded. When `load_bytes_from_disk()` returns `None` (e.g. after the CPU staging pool is exhausted — the graceful-failure path introduced by PR #3006), the policy has already recorded a hit that never delivered data: a **phantom hit**.

The phantom hit inflates the key's recency/frequency score, shielding it from eviction even though the load failed. Under repeated pool exhaustion the same keys accumulate phantom hits and silently distort eviction order.

## Root Cause

```python
# Before this fix (local_disk_backend.py)
def get_blocking(self, key):
    self.disk_lock.acquire()
    ...
    self.cache_policy.update_on_hit(key, self.dict)  # ← recorded BEFORE load
    ...
    self.disk_lock.release()
    memory_obj = self.load_bytes_from_disk(...)       # ← may return None
    return memory_obj
```

## Fix

Move `update_on_hit()` to after `load_bytes_from_disk()` returns, gated on `memory_obj is not None`. Re-acquire `disk_lock` for the update (so concurrent eviction threads cannot observe a partially-updated `OrderedDict`), and guard with `key in self.dict` in case the entry was evicted concurrently during the lock-free disk I/O window.

```python
# After this fix
self.disk_lock.release()
memory_obj = self.load_bytes_from_disk(key, path, dtype=dtype, shape=shape, fmt=fmt)

if memory_obj is not None:
    self.disk_lock.acquire()
    if key in self.dict:
        self.cache_policy.update_on_hit(key, self.dict)
    self.disk_lock.release()

return memory_obj
```

## Tests

Three new regression tests in `tests/v1/storage_backend/test_local_disk_backend.py` (`TestGetBlockingCachePolicyUpdate`):

| Test | Scenario | Expected |
|------|----------|----------|
| `test_no_phantom_hit_when_load_fails` | `load_bytes_from_disk` returns `None` | `update_on_hit` NOT called |
| `test_updates_cache_policy_on_successful_load` | load returns a valid `MemoryObj` | `update_on_hit` called exactly once |
| `test_key_absent_returns_none_without_policy_update` | key not in cache | `update_on_hit` NOT called |

## Scope

This PR fixes only `get_blocking()`. `batched_get_non_blocking()` has a similar pattern (line 467) but the async path delivers data via a `disk_worker` callback, making the fix more involved; that can be addressed in a follow-up.

---

> [!NOTE]
> **Low Risk**
> Small behavioral change scoped to synchronous disk reads plus new unit tests; minimal risk beyond potentially altering eviction ordering to match actual successful loads.
> 
> **Overview**
> Fixes `LocalDiskBackend.get_blocking()` to **defer `cache_policy.update_on_hit()` until after `load_bytes_from_disk()` succeeds**, preventing eviction policy “phantom hits” when a disk load returns `None`.
> 
> The method now releases `disk_lock` during the potentially slow disk/CPU-staging load and then re-acquires it conditionally to update recency only if the key still exists. Adds targeted regression tests asserting no policy update on missing keys or failed loads, and exactly one update on successful loads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa9d401434d33433aab0b48fa6283a4e87b77686. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -404,33 +404,34 @@ class LocalDiskBackend(StorageBackendInterface):
         :returns: A ``MemoryObj`` containing the loaded KV data, or ``None``
             if the key is not present or the load fails.
         """
-        self.disk_lock.acquire()
-        if key not in self.dict:
-            self.disk_lock.release()
-            return None
+        with self.disk_lock:
+            if key not in self.dict:
+                return None
 
-        disk_meta = self.dict[key]
-        path = disk_meta.path
-        dtype = disk_meta.dtype
-        shape = disk_meta.shape
-        fmt = disk_meta.fmt
-        assert dtype is not None
-        assert shape is not None
+            disk_meta = self.dict[key]
+            path = disk_meta.path
+            dtype = disk_meta.dtype
+            shape = disk_meta.shape
+            fmt = disk_meta.fmt
+            assert dtype is not None
+            assert shape is not None
 
-        self.disk_lock.release()
+        # Load is performed outside the lock: it can block for a non-trivial
+        # amount of time (CPU staging pool allocation + memcpy from disk) and
+        # must not hold disk_lock while waiting, or concurrent insert/evict
+        # operations would deadlock.
         memory_obj = self.load_bytes_from_disk(
             key, path, dtype=dtype, shape=shape, fmt=fmt
         )
 
         if memory_obj is not None:
-            # Update cache recency only after the load succeeds; calling
-            # update_on_hit() before confirming success would record a phantom
-            # hit and skew eviction order when load_bytes_from_disk() returns
-            # None (e.g. CPU staging pool exhausted after PR #3006).
-            self.disk_lock.acquire()
-            if key in self.dict:
-                self.cache_policy.update_on_hit(key, self.dict)
-            self.disk_lock.release()
+            # Re-acquire the lock to update the eviction policy.  The key
+            # membership check guards against the entry being evicted between
+            # the two lock regions — in that case the policy state is already
+            # consistent and no update is needed.
+            with self.disk_lock:
+                if key in self.dict:
+                    self.cache_policy.update_on_hit(key, self.dict)
 
         return memory_obj
 
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -394,15 +394,20 @@ class LocalDiskBackend(StorageBackendInterface):
         key: CacheEngineKey,
     ) -> Optional[MemoryObj]:
         """
-        Blocking get function.
+        Load a cached KV chunk from disk synchronously.
+
+        The cache policy is updated only after a successful load so that a
+        failed load (``load_bytes_from_disk`` returning ``None``) does not
+        record a phantom cache hit and skew future eviction decisions.
+
+        :param key: The cache key identifying the KV chunk.
+        :returns: A ``MemoryObj`` containing the loaded KV data, or ``None``
+            if the key is not present or the load fails.
         """
         self.disk_lock.acquire()
         if key not in self.dict:
             self.disk_lock.release()
             return None
-
-        # Update cache recency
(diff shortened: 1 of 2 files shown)


lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -417,6 +422,16 @@ def get_blocking(
...
 
+        if memory_obj is not None:
+            # Update cache recency only after the load succeeds; calling
+            # update_on_hit() before confirming success would record a phantom
+            # hit and skew eviction order when load_bytes_from_disk() returns
+            # None (e.g. CPU staging pool exhausted after PR #3006).
+            self.disk_lock.acquire()
+            if key in self.dict:
+                self.cache_policy.update_on_hit(key, self.dict)
+            self.disk_lock.release()
```
ApostaC: nit: we can use `with self.disk_lock:` rather than calling acquire and relesase.

Additionally, can you explain why we need to have a lock for this operation?

ApostaC (commented):
Otherwise LGTM

SuperMarioYL:
Good nit — updated to use `with self.disk_lock:` throughout, consistent with the rest of the class.

On **why the lock is needed for `update_on_hit`**: between the first lock region (where we read the metadata) and the second (where we update the eviction policy), another goroutine could evict the key. The `if key in self.dict` guard inside the second `with` block is what makes this safe — if the key was concurrently removed, `cache_policy` is already consistent and we skip the update. Without any lock, a concurrent eviction happening between the `if key in self.dict` check and the `update_on_hit` call would corrupt the policy's internal data structures.

The disk load itself deliberately runs between the two lock regions: it can block for a non-trivial amount of time (CPU staging pool allocation + disk read) and holding `disk_lock` across that would block concurrent `insert`/`evict` calls on other keys.

DongDongJu (approved):
LGTM, It recall me this PR(https://github.com/LMCache/LMCache/pull/2734). I will rebase after merging this.

ApostaC (approved):
lgtm!

PR #3168, opened 2026-04-30
[experimental] torch dev for cpu only

hlin99:
> [!NOTE]
> **Medium Risk**
> Medium risk because it refactors device selection/stream/event handling across cache engine, storage backends, and multiprocess IPC; incorrect device routing or missing backend capabilities could cause runtime failures on non-CUDA platforms.
> 
> **Overview**
> Introduces a unified `torch_dev`/`torch_device_type` entrypoint (detected once in `lmcache/__init__.py`) and updates LMCache to use it instead of hard-coded `torch.cuda` calls, enabling broader accelerator support (CUDA/XPU/HPU) and safer CPU/CLI fallback.
> 
> Refactors key paths (SGLang/vLLM integrations, cache engine, memory allocators, storage backends, transfer channels, and multiprocess servers) to route tensor moves, streams, synchronization, and device selection through `torch_dev`, adding explicit runtime guards/errors where CUDA-only features are required (IPC `Event`, `Event.from_ipc_handle`, `cudart`, `_share_cuda_`). Also adds an architecture design doc and skips CUDA-dependent tests when CUDA is unavailable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6be30ae4dcd828e919b9224c554bf668a14fe17f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lmcache/__init__.py
```diff
@@ -15,7 +15,55 @@
...
+            e.g. ``(torch.cuda, "cuda")`` or ``(torch.xpu, "xpu")``.
+
+    Raises:
+        RuntimeError: If no supported accelerator is found (checked CUDA, XPU, HPU).
+    """
+    try:
+        # Third Party
+        import torch
+    except ImportError:
+        return None, "cpu"  # fallback，CLI-only
```
maobaolong, line 40: For cpu, my env(macos), it can import torch successfully.

DongDongJu:
@hlin99 @maobaolong  I was thought about this approach. But it seems too hard to maintain it. 
Just my opinion, How about add the CPUAdapter and connect to them not to pretend the CUDA existing? WDYT?

hlin99:
> @hlin99 @maobaolong I was thought about this approach. But it seems too hard to maintain it. Just my opinion, How about add the CPUAdapter and connect to them not to pretend the CUDA existing? WDYT?

yes. agree.

there're two ways for cpu only device after https://github.com/LMCache/LMCache/pull/3091 is merged.

1. Pretend to be cuda, and mock functions as needed. short term easy, but long term hard.
2. consider cpu same as other gpu device, and implement an adapter for the torch_dev.functions, I would prefer this more than opt#1.

DongDongJu:
@hlin99 Sounds like a plan. Let me go with https://github.com/LMCache/LMCache/pull/3091 first then.

Issue #3156, opened 2026-04-28
[Feature Request] DeepSeek V4 Support


yoo-kumaneko:
**Description**
LMCache does not support DeepSeek-V4 now. Let's support it!

**Describe the solution you'd like**
We need to support the hybrid attention architectures of CSA (Compressed Sparse Attention) and HCA (Heavily Compressed Attention)

tianlang-wq:
@liuyumoye  Does the system currently support lmcache + mooncake to implement L3 kvcache storage?

d3nb:
Still relevant — please keep this open. Some field notes from running LMCache against DeepSeek-V4 (Flash and Pro) on Blackwell, in case they help scope what "V4 support" actually means.

**It is partially usable today, but only by steering around a few sharp edges**, and they all trace back to the same structural fact: V4's hybrid attention produces *multiple KV kernel groups* (full-attention plus sliding-window), while several LMCache paths assume a single shape view over `MemoryObj.tensor`.

Concretely, what we hit:

- **#3106 (multi-group KV layouts)** is the root cause, not a Gemma-specific issue. On V4-Pro with `separate_object_groups=True` (the default), the 8 kernel groups get split into 2 object groups by sliding-window size, and retrieve fails 100% with a shape mismatch (`memory_obj=… vs gpu_buffer=…`). `--no-separate-object-groups` is a working workaround — details and a clean A/B are in [my comment there](https://github.com/LMCache/LMCache/issues/3106#issuecomment-5005769386).
- **#4217** (PR, mine) fixes what that workaround then exposes: with object-group separation off, the sliding-window H2D load plan needs trimming per kernel group.
- **#3931 / #3957** cover the MPConnector side — a V4-Flash load error, and `lmcache_mp_connector_0201.py` drifting from the refactored adapter.

So this umbrella issue decomposes fairly cleanly into those, which is an argument for keeping it open rather than letting it age out: the sub-issues each fix one edge, but none of them declares V4 supported, and someone arriving at LMCache with a V4 checkpoint has no single place that says which combination currently works.

If it helps, the two things that would most move "V4 support" from *workaround-able* to *supported*, from where we sit:

1. Make the multi-group KV layout a first-class case rather than something `separate_object_groups` has to be turned off for (#3106 + #4217).
2. State the supported matrix explicitly — which connector, which vLLM version, and whether CSA/HCA are expected to work — since the V4 support path currently reads differently depending on which issue you land on.

[...]

PR #3171, opened 2026-04-30
[MP][Feat] Support DeepSeek V4

liuyumoye:

Enable fast-path support for DeepSeek V4 in the LMCache MP (multi-process) runtime. DeepSeek V4 introduces a mixed-compression KV layout where some layer groups (e.g. the Compressor / Indexer caches) compress multiple logical tokens into a single physical slot while sharing a unified KV pool with the larger attention groups. The existing code assumed a single global `block_size` and a strictly contiguous per-layer KV tensor, both of which break on V4. This PR adapts the MP path end-to-end so that V4 workloads can store/retrieve KV caches correctly without touching the hot-path kernels' launch overhead.

Concrete adaptations:

1. **Per-group `block_size` (compression-aware layer grouping).**
   - `LayerGroupIdentity` is extended from `(kv_size, num_heads, head_size, dtype)` to `(kv_size, num_heads, head_size, block_size, dtype)`, so compressed and non-compressed groups that happen to share `(nh, hs, dtype)` no longer collapse into the same dispatch group.
   - `KVLayerGroupsManager` discovers each group's `block_size` per-layer via `get_block_size(...)` and stamps it into `PageBufferShapeDesc.bs`; the old `block_size=` constructor argument is removed.
   - `GPUCacheContext` drops the single `self.block_size_` field. A new `REGISTER_KV_CACHE` payload field `vllm_block_size` is forwarded from the adapter so the server can derive per-group `compress_ratio = vllm_block_size // group.shape_desc.bs` and `chunk_slots = lmcache_chunk_size // compress_ratio`. This is required because when *every* layer is compressed, `max(shape_desc.bs)` is strictly less than the true vLLM block size and cannot be recovered from GPU tensor shapes alone.
   - `get_layout_desc` and the chunk transfer path in `MPCacheEngine` now size each group's MemoryObj shape in *physical* slots (`num_tokens // compress_ratio`); `blocks_per_chunk` is computed against `vllm_block_size` since block-id indexing is shared with vLLM.
   - `GPUCacheContext.__init__` is split into six focused phases (`_init_kv_layout`, `_init_layer_groups`, `_init_compression_metadata`, buffer alloc, stream init) to keep the compression-metadata derivation readable.

2. **Dim-0-padded KV tensors from vLLM's shared KV pool.**
   In vLLM's mixed-compression KV pool, every group's per-block row is padded up to the pool's maximum row width, i.e. `tensor.stride(0) > prod(shape[1:])`. The tensor is *not* non-contiguous in the slicing sense; it is a periodic-dim-0 strided view over a larger storage.
   - `attempt_permute_to_contiguous_view` now accepts a tensor as a valid zero-copy view iff only dim-0 carries the extra stride (inner dims tightly packed, `storage_offset == 0`, no dim overlap). Any other pattern (interior-dim padding, slice, `as_strided`, etc.) is rejected loudly with full `shape/stride/storage_offset/dtype/storage_nbytes` context — we refuse to silently `.contiguous()`-copy.
   - `PageBufferShapeDesc` gains a new field `block_stride_elems` (physical per-block stride in source-dtype elements). `make_page_buffer_shape_desc` accepts it as an optional argument and validates it against the tight stride.
   - The CUDA kernel offset math (`calculate_engine_global_offset`) in `csrc/mp_mem_kernels.cu` is updated for the two block-axis formats `NL_X_NB_TWO_BS_NH_HS` and `NL_X_NB_BS_HS` to step by `block_stride_elems` instead of the tight stride when set; the SGL token-row formats and the layer-packed `NB_NL_*` format explicitly ignore it (dim-0 is not a block axis there). Both `multi_layer_block_kv_transfer` and `lmcache_memcpy_async_d2h` share this code path.
   - `KVLayerGroupsManager._resolve_block_stride` is the format-aware second line of defence: for block-axis formats it propagates `stride(0)` into the shape desc; for non-block-axis formats it raises if dim-0 padding is present (since the kernel cannot honour it).
   - The representative per-group tensor layout (shape / stride / `is_contiguous` / `padding_per_block` / storage offset / dtype) is logged at init so operators can diagnose unexpected layouts without re-running with extra instrumentation.

3. **Misc robustness & adapter plumbing.**
   - `get_device` now raises `ValueError` with the full descent path when the KV cache structure is empty or malformed (previously it died with a bare `IndexError` when the adapter's skip filter dropped every layer).
   - The vLLM adapter (`lmcache_mp_connector_0180.py`, `vllm_multi_process_adapter.py`) is refactored to build a single `ParallelStrategy` object and forward `vllm_block_size` to both the scheduler and worker adapter; existing callers that only cared about `world_size/kv_rank` keep working.
   - `LMCacheMPConnector` is renamed to `LMCacheMPConnectorDynamic` to match the dynamic compression-aware code path.
   - `wrap_kv_caches` now logs a per-layer `(name, shape, dtype)` summary so operators can verify the exact layer set shipped to the server.

- The `block_stride_elems` field on `PageBufferShapeDesc` defaults to `0` ("unset"), and the kernel falls back to the format-specific tight stride — so every existing caller (non-V4, non-padded pools) is bit-for-bit unchanged. Only callers that pass a non-zero value take the new path.
- The tight/padded stride conversion happens in `block_stride_or<ScalarType>()` on the C++ side and converts from *source KV dtype* element units to the kernel's scalar type, so uint8 staging of fp16 KV still lands at the right offsets.
- The non-contiguous-view acceptance in `attempt_permute_to_contiguous_view` is intentionally narrow (dim-0 only, `storage_offset == 0`, inner dims tight). If you see cases in the wild that should be accepted but aren't, please flag them — the failure path deliberately logs every piece of info needed to diagnose.
- `REGISTER_KV_CACHE`'s protocol signature changed (extra `int` payload). Older servers talking to a newer adapter (or vice-versa) will need to be upgraded together; no compat shim is provided because this is an MP-internal RPC.
- We removed the `is_mla` branch from the layer-grouping loop's identity (MLA already forces `kv_size=1, nh=1`, so adding `block_size` is strictly more discriminative and preserves the previous partitioning for non-V4 MLA models).

**If applicable**:

- [x] this PR contains user facing changes - docs added *(API-level: `LMCacheMPConnector` → `LMCacheMPConnectorDynamic`; `REGISTER_KV_CACHE` payload gains `vllm_block_size`; `PageBufferShapeDesc` gains `block_stride_elems`. Inline docstrings updated; no external user-facing docs changes beyond the renamed connector.)*
- [x] this PR contains unit tests *(updated: `test_gpu_context.py`, `test_gpu_connector.py`, `test_kv_layer_groups_manager.py`, `test_mq.py`, `test_mq_handler_helpers.py`, `test_blend_server_v2.py`, `test_cache_server.py` for the new `vllm_block_size` payload and per-group `block_size`.)*

---

> [!NOTE]
> **High Risk**
> Touches CUDA kernel addressing and the MP wire protocol (`REGISTER_KV_CACHE` payload), so mismatches between client/server builds or stride/slot miscalculation could corrupt KV transfers or break MP compatibility.
> 
> **Overview**
> Enables DeepSeek V4-style *mixed-compression* KV layouts in the multi-process (MP) path by making KV transfer grouping and chunk sizing **per-layer-group** instead of assuming a single global `block_size`.
> 
> Adds `vllm_block_size` to the `REGISTER_KV_CACHE` RPC and reworks `GPUCacheContext`/server store+retrieve to compute per-group `compress_ratio` and pass *physical slots per chunk* into `multi_layer_block_kv_transfer`; also extends `KVLayerGroupsManager` identity to include per-group `bs` and tightens layout validation/logging.
> 
> Supports vLLM unified-pool *dim-0 padded* strided views by introducing `PageBufferShapeDesc.block_stride_elems` (pybind + Python stand-in), teaching the CUDA MP kernel to honor it for block-axis formats, and relaxing `attempt_permute_to_contiguous_view` to accept dim-0-only padding while rejecting other unsupported strided views. The vLLM connector plumbing is updated to pass a `ParallelStrategy`, forward `vllm_block_size`, and the connector class is renamed to `LMCacheMPConnectorDynamic` (tests updated accordingly).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4832a4000250db528098ce65d8caaf748a688ebc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/csrc/mp_mem_kernels.cuh
+++ b/csrc/mp_mem_kernels.cuh
@@ -15,6 +15,23 @@ struct PageBufferShapeDesc {
   int nh;            // num heads
   int hs;            // head size
   int element_size;  // bytes (1 or 2)
+  // Physical per-block stride in source-dtype element units, used by
+  // formats whose dim-0 is the block axis to step over padding bytes
+  // (e.g. DeepSeek V4 compressor / indexer caches sharing a vLLM KV
+  // pool with larger attn groups, whose rows are padded up to the
+  // pool's max row width). 0 means "unset — fall back to the
+  // format-specific tight stride".
+  //
+  // CONTRACT: pass ``tensor.stride(0)`` verbatim. PyTorch stride
+  // semantics already absorb every inner-dim extent (including
+  // ``kv_size``), so DO NOT pre-multiply by any inner dim.
+  //
+  // Honoured today only by NL_X_NB_BS_HS (per-layer [NB, BS, HS],
+  // MLA). NL_X_NB_TWO_BS_NH_HS is restricted to the tight form
+  // upstream and leaves this field at 0; all other formats either
+  // pack non-block info into dim-0 or do not support dim-0 padding,
+  // and ignore this field.
+  int block_stride_elems;
 
   template <typename ScalarType>
   __host__ __device__ inline size_t scalars_per_head() const {
@@ -26,9 +43,20 @@ struct PageBufferShapeDesc {
     return nh * hs * element_size / sizeof(ScalarType);
   }
 
+  // Per (K or V) block step along dim-0, expressed in ``ScalarType``
+  // element units (the kernel's working dtype, e.g. uint4 / uint32_t /
+  // uint16_t). Returns the tight ``bs * nh * hs`` by default, or the
+  // physical ``block_stride_elems`` when dim-0 carries padding (today
+  // only NL_X_NB_BS_HS, see ``block_stride_elems`` above). Every
+  // ``calculate_engine_global_offset`` branch uses this as the dim-0
+  // step, so honouring padding here propagates to all formats without
+  // per-branch changes.
   template <typename ScalarType>
   __host__ __device__ inline size_t scalars_per_block() const {
-    return bs * nh * hs * element_size / sizeof(ScalarType);
+    const size_t elems = block_stride_elems > 0
+                             ? static_cast<size_t>(block_stride_elems)
+                             : static_cast<size_t>(bs) * nh * hs;
+    return elems * element_size / sizeof(ScalarType);
   }
 };
 
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -33,7 +33,22 @@ DEFAULT_HEARTBEAT_INTERVAL: float = 10.0
 
 
 def wrap_kv_caches(kv_caches: dict[str, torch.Tensor]) -> KVCache:
-    logger.info("KV caches keys are %s", list(kv_caches.keys()))
+    # Emit a per-layer (name, shape, dtype) summary so the operator can
+    # verify the exact layer set & tensor geometry being shipped to the
+    # LMCache server, then the low-noise count of handles being wrapped.
+    kept_summary = [
+        (name, tuple(tensor.shape), str(tensor.dtype))
+        for name, tensor in kv_caches.items()
+    ]
+    logger.debug(
+        "KV cache transfer keeping %d layer(s) (name, shape, dtype):\n%s",
+        len(kept_summary),
+        "\n".join(
+            f"  [{i}] {name}  shape={shape}  dtype={dtype}"
+            for i, (name, shape, dtype) in enumerate(kept_summary)
+        ),
+    )
+    logger.info("Wrapping %d KV cache tensors for IPC", len(kv_caches))
     return [CudaIPCWrapper(tensor) for tensor in kv_caches.values()]
 
 
@@ -729,6 +744,14 @@ class LMCacheMPWorkerAdapter:
             "LMCache chunk size should be a multiple of vLLM block size"
         )
         self.blocks_in_chunk = chunk_size // vllm_block_size
+        # Retain the vLLM logical block size so we can ship it to the
+        # LMCache server in ``register_kv_caches`` — the server uses it
+        # (as ``layout_hints["inference_engine_logical_block_size"]``)
+        # to derive per-group compression ratios when some KV layer
+        # groups compress multiple logical tokens into a single physical
+        # slot (``shape_desc.bs <
+        # inference_engine_logical_block_size``).
+        self.vllm_logical_block_size = vllm_block_size
 
         # Health state (shared with heartbeat thread)
         self._health_event = threading.Event()
@@ -824,6 +847,9 @@ class LMCacheMPWorkerAdapter:
         from lmcache.integration.vllm.utils import vllm_layout_hints
 
         layout_hints = vllm_layout_hints()
+        layout_hints["inference_engine_logical_block_size"] = (
+            self.vllm_logical_block_size
+        )
         future = send_lmcache_request(
             self.mq_client,
             RequestType.REGISTER_KV_CACHE,
--- a/lmcache/v1/gpu_connector/utils.py
+++ b/lmcache/v1/gpu_connector/utils.py
@@ -70,12 +70,23 @@ class LayoutHints(TypedDict, total=False):
             reshape its 4-D pool tensor into the canonical 6-D form.
         tokens_per_block: Tokens per paged block. Used by TRT-LLM (same).
         head_dim: Per-head dimension. Used by TRT-LLM (same).
+        inference_engine_logical_block_size: Inference-engine-side block
+            size (logical tokens per engine block; for vLLM this is
+            ``cache_config.block_size``). Carried inside
+            ``LayoutHints`` (instead of as a standalone
+            ``REGISTER_KV_CACHE`` argument) so that engines without a
+            logical block-size concept can simply omit it. The server
+            uses it to derive per-group compression ratios when some
+            KV layer groups compress multiple logical tokens into a
+            single physical slot
+            (``shape_desc.bs < inference_engine_logical_block_size``).
     """
 
     kv_layout: Literal["NHD", "HND"]
     num_kv_heads: int
     tokens_per_block: int
     head_dim: int
+    inference_engine_logical_block_size: int
 
 
 def attempt_permute_to_contiguous_view(
@@ -94,11 +105,21 @@ def attempt_permute_to_contiguous_view(
     ``[2, NB, BS, NH, HS]`` via a dim permute. Sorting dims by stride
     undoes the permute without touching storage.
 
-    Raises:
-        ValueError: If a tensor leaf is non-contiguous for a reason
-            other than dim permutation (e.g. slicing, ``as_strided``).
-            We refuse to fall back to ``.contiguous()`` (which would
-            copy) so the caller's invariant is never silently violated.
+    For tensors that remain non-contiguous even after dim-permute
+    recovery (e.g. vLLM unified KV pool views where dim-0 has an
+    inflated periodic stride because every block slot is padded to a
+    model-wide maximum), this function returns the tensor unchanged.
+    Rationale: :class:`CudaIPCWrapper` transports ``(shape, stride,
+    storage_offset)`` verbatim and the receiver rebuilds the view via
+    ``torch.Tensor.set_(storage, offset, shape, stride)``, which
+    supports arbitrary strided views (including periodic-dim-0 and
+    ``as_strided``-produced layouts) and yields a bit-identical view.
+    Downstream consumers that rely on ``shape`` alone to infer the
+    physical layout must therefore also consult ``stride``.
+
+    We deliberately never fall back to ``.contiguous()`` (which would
+    allocate and copy), so the caller's zero-copy invariant is
+    preserved.
     """
     if isinstance(kv_caches, torch.Tensor):
         if kv_caches.is_contiguous():
@@ -106,13 +127,105 @@ def attempt_permute_to_contiguous_view(
         strides = kv_caches.stride()
         perm = sorted(range(kv_caches.ndim), key=lambda i: strides[i], reverse=True)
         result = kv_caches.permute(perm)
-        if not result.is_contiguous():
-            raise ValueError(
-                "tensor is non-contiguous for reasons other than permutation "
-                "(e.g. slicing or as_strided). Cannot recover contiguous view."
-            )
+        if result.is_contiguous():
+            return result
+        # Non-permute non-contiguity: only the strict dim-0-padding pattern
+        # is recoverable downstream. Delegate validation + diagnostics to
+        # the helper; on success we keep the stride-sorted view as-is and
+        # rely on ``PageBufferShapeDesc.block_stride_elems`` to honour the
+        # padding.
+        padding_per_block = _validate_dim0_padded_layout(result)
+        logger.debug(
+            "attempt_permute_to_contiguous_view: accepting dim-0-padded "
+            "view; downstream kernels must honour block_stride_elems. "
+            "shape=%s, stride=%s, padding_per_block_elems=%d, "
+            "storage_nbytes=%s, dtype=%s",
+            tuple(result.shape),
+            tuple(result.stride()),
+            padding_per_block,
+            int(result.untyped_storage().nbytes()),
+            result.dtype,
+        )
         return result
     return [attempt_permute_to_contiguous_view(sub) for sub in kv_caches]
+
+
+def _validate_dim0_padded_layout(tensor: torch.Tensor) -> int:
+    """Validate that *tensor* matches the dim-0-padding-only strided layout.
+
+    Mainly used for DeepSeek V4 integration, where compressor / indexer
+    KV groups share a pool with larger attn groups and end up with
+    per-block dim-0 padding. The downstream KV transfer kernels only
+    honour this single non-contiguous shape (via
+    :class:`PageBufferShapeDesc.block_stride_elems`); any other strided
+    view would cause wrong reads/writes and is rejected here.
+
+    The accepted layout requires:
+
+    * ``stride[-1] == 1`` and ``stride[-2] == shape[-1]`` — each block
+      row is internally tightly packed.
+    * Every interior dim ``i`` satisfies
+      ``stride[i] == prod(shape[i+1:])`` — only dim-0 may carry
+      padding, with ``stride[0] >= prod(shape[1:])``.
+    * ``storage_offset == 0`` — no slice/narrow base shift.
+
+    Callers must pass the stride-sorted permuted view (not the original
+    tensor): for tensors that are both permuted and dim-0-padded, the
+    original's unsorted inner strides would falsely trip the tight-
+    packing check. ``permute`` shares storage and preserves
+    ``storage_offset``/``numel``/storage bytes, so those checks are
+    equivalent on either view.
+
+    Returns:
+        ``padding_per_block_elems`` (= ``stride[0] - prod(shape[1:])``).
+
+    Raises:
+        ValueError: *tensor* violates any of the invariants above.
+    """
+    shape = tuple(tensor.shape)
+    stride = tuple(tensor.stride())
+    ndim = tensor.ndim
+    storage_offset = int(tensor.storage_offset())
+
+    def _fail(reason: str) -> None:
+        raise ValueError(
+            "attempt_permute_to_contiguous_view: tensor is non-contiguous "
+            f"and not a supported (dim-0 padding only) layout — {reason}. "
+            f"shape={shape}, stride={stride}, "
+            f"storage_offset={storage_offset}, numel={int(tensor.numel())}, "
+            f"storage_nbytes={int(tensor.untyped_storage().nbytes())}, "
+            f"dtype={tensor.dtype}. "
+            "Downstream KV transfer kernels only understand dim-0 "
+            "block-row padding; other strided views would produce "
+            "wrong reads/writes and are rejected."
+        )
+
+    if ndim < 2:
+        _fail("ndim < 2")
+    if stride[-1] != 1:
+        _fail("stride[-1] != 1 (inner dim not contiguous)")
+    if stride[-2] != shape[-1]:
+        _fail("stride[-2] != shape[-1] (last-two dims not tightly packed)")
+    if storage_offset != 0:
+        _fail("storage_offset != 0 (slice/narrow view, base address shifted)")
+    # Interior dims (1 .. ndim-2 exclusive) must be tightly packed with
+    # respect to the dims to their right. Only dim-0's stride is allowed
+    # to exceed the tight value.
+    inner_tight = 1
+    for i in range(ndim - 1, 0, -1):
+        if i < ndim - 1 and stride[i] != inner_tight:
+            _fail(
+                f"dim {i} stride {stride[i]} != tight {inner_tight} "
+                "(interior-dim padding is not supported)"
+            )
+        inner_tight *= shape[i]
+    # Now ``inner_tight == prod(shape[1:])``; dim-0 must be >= that.
+    if stride[0] < inner_tight:
+        _fail(
+            f"dim-0 stride {stride[0]} < prod(shape[1:])={inner_tight} "
+            "(overlapping blocks)"
+        )
+    return stride[0] - inner_tight
 
 
 def assert_contiguous(tensor: torch.Tensor) -> None:
@@ -527,10 +640,17 @@ def get_num_blocks(
 
 
 def get_block_size(
-    kv_caches: DiscoverableKVCache, gpu_kv_format: "lmc_ops.GPUKVFormat"
+    kv_caches: DiscoverableKVCache,
+    gpu_kv_format: "lmc_ops.GPUKVFormat",
+    layer_idx: int = 0,
 ) -> int:
-    """
-    Get the block size from the kv_caches
+    """Return the block size (tokens per block) for layer ``layer_idx``.
+
+    ``layer_idx`` is honoured only for per-layer formats where BS may
+    differ across layers (e.g. mixed-compression MLA pools). For
+    cross-layer formats BS is shared across layers and ``layer_idx``
+    is ignored. Raises ``ValueError`` for NBBS-fused formats, which
+    have no separate BS dim.
     """
     if gpu_kv_format == lmc_ops.GPUKVFormat.NB_NL_TWO_BS_NH_HS:
         return kv_caches.shape[3]
@@ -542,15 +662,15 @@ def get_block_size(
         lmc_ops.GPUKVFormat.NL_X_NB_TWO_BS_NH_HS,
     ):
         # NHD: [..., BS, NH, HS] — block_size at shape[2]
-        return kv_caches[0].shape[2]
+        return kv_caches[layer_idx].shape[2]
     elif gpu_kv_format in (
         lmc_ops.GPUKVFormat.NL_X_TWO_NB_NH_BS_HS,
         lmc_ops.GPUKVFormat.NL_X_NB_TWO_NH_BS_HS,
     ):
         # HND: [..., NH, BS, HS] — block_size at shape[3]
-        return kv_caches[0].shape[3]
+        return kv_caches[layer_idx].shape[3]
     elif gpu_kv_format == lmc_ops.GPUKVFormat.NL_X_NB_BS_HS:
-        return kv_caches[0].shape[1]
+        return kv_caches[layer_idx].shape[1]
     elif gpu_kv_format == lmc_ops.GPUKVFormat.TWO_X_NL_X_NBBS_NH_HS:
         raise ValueError(_ATTRIBUTE_NOT_EXIST_ERROR.format(format=gpu_kv_format))
     elif gpu_kv_format == lmc_ops.GPUKVFormat.NL_X_NBBS_ONE_HS:
@@ -950,6 +1070,153 @@ def get_device(kv_caches: DiscoverableKVCache) -> torch.device:
     return probe.device
 
 
+# Formats whose per-layer tensor dim-0 is the *block* axis AND for
+# which we currently support dim-0 padding (e.g. DeepSeek V4
+# compressor / indexer caches sharing a KV pool with larger attn
+# groups). Today only the MLA layout (``NL_X_NB_BS_HS``, kv_size==1)
+# is exercised by real mixed-compression workloads.
+#
+# ``NL_X_NB_TWO_BS_NH_HS`` *could* in principle also be the block
+# axis on dim-0, but no real serving engine emits a padded layout of
+# that format yet, and supporting it would require: (a) deciding
+# (without a ground-truth example) which axis carries the padding
+# — NB boundary vs K↔V offset — and (b) a coordinated change in
+# ``attempt_permute_to_contiguous_view`` to let interior-dim padding
+# through for that one format. Rather than ship an unverifiable code
+# path, we keep ``NL_X_NB_TWO_BS_NH_HS`` out of this set, which means
+# any padded tensor of that format will fail loudly via the
+# non-block-axis dim-0-padding check below. Revisit and add a
+# properly-tested branch when a concrete use case lands.
+_BLOCK_AXIS_FORMATS: frozenset = frozenset(
+    {
+        lmc_ops.GPUKVFormat.NL_X_NB_BS_HS,
+    }
+)
+
+
+def resolve_block_stride_and_log_layout(
+    kv_caches: DiscoverableKVCache,
+    gpu_kv_format: "lmc_ops.GPUKVFormat",
+    layer_idx: int,
+    group_idx: int,
+) -> Optional[int]:
+    """Resolve the per-block stride for a KV layer group and log its layout.
+
+    Single entry point for :class:`KVLayerGroupsManager` to obtain the
+    ``block_stride_elems`` value for :class:`PageBufferShapeDesc` and emit
+    a one-shot layout audit line. All ``GPUKVFormat``-aware reasoning is
+    kept here so callers never touch a "representative KV cache" tensor.
+
+    * Block-axis formats (:data:`_BLOCK_AXIS_FORMATS`): ``stride(0)`` is
+      the per-block step and is returned as-is. A value larger than the
+      tight stride indicates dim-0 padding (e.g. DeepSeek V4 compressor
+      caches sharing a KV pool with larger attn groups).
+    * Other formats: dim-0 is not the block axis, so ``None`` is
+      returned and ``shape_desc`` falls back to the tight stride. Any
+      dim-0 padding in such formats is rejected with ``ValueError``
+      since downstream kernels cannot honour it.
+
+    Args:
+        kv_caches: Full KV cache structure (already normalised).
+        gpu_kv_format: Format of ``kv_caches``.
+        layer_idx: 0-based layer index used as the layout probe.
+        group_idx: 0-based group index, used only for logging.
+
+    Returns:
+        ``stride(0)`` for block-axis formats; ``None`` otherwise.
+
+    Raises:
+        ValueError: Non-block-axis format carries dim-0 padding.
+    """
+
+    def _pick_layout_probe_tensor() -> torch.Tensor:
+        # Layout probe only (shape/stride/storage_offset/dtype); not
+        # the K/V slice fed to the transfer kernel.
+        # - Cross-layer formats: ``kv_caches`` is the single backing
+        #   tensor packing all layers along dim-1; indexing dim-0
+        #   (= NB) would yield a per-block slice, so return the whole
+        #   tensor (its ``stride(0)`` is the authoritative per-NB step).
+        # - SGL MHA: outer list is K/V (length 2), inner list is
+        #   per-layer; K & V share shape/stride by construction.
+        # - Other formats: ``kv_caches`` is already a per-layer list.
+        if gpu_kv_format in (
+            lmc_ops.GPUKVFormat.NB_NL_TWO_BS_NH_HS,
+            lmc_ops.GPUKVFormat.NB_NL_TWO_NH_BS_HS,
+        ):
+            if not isinstance(kv_caches, torch.Tensor):
+                raise TypeError(
+                    "Cross-layer GPUKVFormat expects a single backing "
+                    f"torch.Tensor, got {type(kv_caches).__name__}."
+                )
+            return kv_caches
+        if gpu_kv_format == lmc_ops.GPUKVFormat.TWO_X_NL_X_NBBS_NH_HS:
+            return kv_caches[0][layer_idx]  # type: ignore[index]
+        return kv_caches[layer_idx]  # type: ignore[index]
+
+    rep = _pick_layout_probe_tensor()
+
+    block_stride_elems: Optional[int]
+    if gpu_kv_format in _BLOCK_AXIS_FORMATS and rep.ndim > 0:
+        block_stride_elems = int(rep.stride(0))
+    else:
+        # Non-block-axis format: detect forbidden dim-0 padding.
+        if rep.ndim >= 2:
+            tight_dim0 = 1
+            for d in range(1, rep.ndim):
+                tight_dim0 *= int(rep.shape[d])
+            padding = int(rep.stride(0)) - tight_dim0
+            if padding > 0:
+                raise ValueError(
+                    "resolve_block_stride_and_log_layout: group's probe "
+                    f"tensor has dim-0 padding ({padding} elements per "
+                    f"block) but gpu_kv_format={gpu_kv_format!r} is not "
+                    "a supported dim-0-padded format (only "
+                    "NL_X_NB_BS_HS is); downstream transfer kernels "
+                    "cannot honour this padding and would read/write "
+                    "wrong bytes. "
+                    f"layer_idx={layer_idx}, shape={tuple(rep.shape)}, "
+                    f"stride={tuple(rep.stride())}, "
+                    f"tight_stride0={tight_dim0}, "
+                    f"storage_offset={int(rep.storage_offset())}, "
+                    f"dtype={rep.dtype}."
+                )
+        block_stride_elems = None
+
+    # Best-effort layout audit log; the log line itself must not raise.
+    shape = tuple(rep.shape)
+    stride = tuple(rep.stride())
+    try:
+        inner = 1
+        for s in shape[1:]:
+            inner *= int(s)
+        padding_per_block = stride[0] - inner if stride else 0
+    except Exception:
+        padding_per_block = -1
+    try:
+        storage_nbytes = rep.untyped_storage().nbytes()
+    except Exception:
+        storage_nbytes = -1
+    logger.info(
+        "Group %d first-layer tensor: layer_idx=%d shape=%s "
+        "stride=%s is_contiguous=%s dtype=%s device=%s "
+        "storage_offset=%d numel=%d storage_nbytes=%d "
+        "padding_per_block=%d",
+        group_idx,
+        layer_idx,
+        shape,
+        stride,
+        rep.is_contiguous(),
+        rep.dtype,
+        rep.device,
+        rep.storage_offset(),
+        rep.numel(),
+        storage_nbytes,
+        padding_per_block,
+    )
+
+    return block_stride_elems
+
+
 def make_page_buffer_shape_desc(
     kv_caches: DiscoverableKVCache,
     gpu_kv_format: "lmc_ops.GPUKVFormat",
@@ -957,6 +1224,7 @@ def make_page_buffer_shape_desc(
     num_layers_in_group: int,
     num_blocks: int,
     block_size: int,
+    block_stride_elems: Optional[int] = None,
 ) -> "lmc_ops.PageBufferShapeDesc":
     """Build a :class:`PageBufferShapeDesc` from a representative layer.
 
@@ -967,6 +1235,17 @@ def make_page_buffer_shape_desc(
         num_layers_in_group: Number of layers in the group (``nl``).
         num_blocks: Number of paged blocks (``nb``).
         block_size: Tokens per block (``bs``).
+        block_stride_elems: Physical per-block stride in *elements*
+            (= ``tensor.stride(0)`` of the representative layer). Pass
+            the real value whenever the group's KV pool may be
+            dim-0-padded (e.g. DeepSeek V4 compressor/indexer caches
+            sharing a row width with a larger group in the same pool);
+            otherwise downstream transfer kernels will skip into
+            padding and corrupt data. Leave as ``None`` for unpadded
+            pools — the kernel's ``per_block_stride()`` fallback
+            (block_stride_elems <= 0) will reconstruct the tight
+            stride from ``kv_size`` and ``scalars_per_block`` itself,
+            so we don't duplicate that arithmetic on the Python side.
 
     Returns:
         A populated ``PageBufferShapeDesc``.
@@ -983,6 +1262,9 @@ def make_page_buffer_shape_desc(
     )
     desc.hs = get_head_size(kv_caches, gpu_kv_format, layer_idx)
     desc.element_size = get_dtype(kv_caches, gpu_kv_format, layer_idx).itemsize
+
+    resolved_stride = int(block_stride_elems) if block_stride_elems else 0
+    desc.block_stride_elems = resolved_stride
     return desc
 
 
--- a/lmcache/v1/kv_layer_groups.py
+++ b/lmcache/v1/kv_layer_groups.py
@@ -13,7 +13,7 @@ import lmcache.c_ops as lmc_ops
 
 if TYPE_CHECKING:
     # First Party
-    from lmcache.v1.gpu_connector.utils import DiscoverableKVCache
+    from lmcache.v1.gpu_connector.utils import DiscoverableKVCache, LayoutHints
 
 logger = init_logger(__name__)
 
@@ -36,10 +36,10 @@ DTYPE_MAP: dict[str, torch.dtype] = {
 
 
 # The 4-tuple that uniquely identifies a set of kernel-equivalent layers:
-# ``(kv_size, num_heads, head_size, dtype)``. Two layers share a transfer-
+# ``(kv_size, num_heads, head_size, block_size, dtype)``. Two layers share a transfer-
 # kernel launch iff they share this identity — see the grouping loop in
 # :meth:`KVLayerGroupsManager.__init__` for the derivation.
-LayerGroupIdentity = tuple[int, int, int, torch.dtype]
+LayerGroupIdentity = tuple[int, int, int, int, torch.dtype]
 
 
 @dataclass
@@ -49,11 +49,12 @@ class KVLayerGroupInfo:
 
     Membership is decided by :class:`KVLayerGroupsManager` according to
     :data:`LayerGroupIdentity`; every layer referenced by
-    ``layer_indices`` shares the same ``(kv_size, num_heads, head_size,
-    dtype)`` signature. Consumers use ``layer_indices`` to pull the
-    matching device pointers out of ``kv_caches`` (via
-    :func:`~lmcache.v1.gpu_connector.utils.get_group_data_ptrs`) and feed
-    them to the kernel alongside ``shape_desc``.
+    ``layer_indices`` shares the same
+    ``(kv_size, num_heads, head_size, block_size, dtype)`` signature.
+    Consumers use ``layer_indices`` to pull the matching device pointers
+    out of ``kv_caches`` (via
+    :func:`~lmcache.v1.gpu_connector.utils.get_group_data_ptrs`) and
+    feed them to the kernel alongside ``shape_desc``.
 
     ``dtype`` is carried alongside ``shape_desc`` because
     ``PageBufferShapeDesc.element_size`` is a byte width, which cannot
@@ -71,12 +72,28 @@ class KVLayerGroupInfo:
     the per-group pointer array."""
     shape_desc: "lmc_ops.PageBufferShapeDesc"
     """Kernel-facing shape descriptor shared by every layer in the group.
-    All seven fields (``kv_size, nl, nb, bs, nh, hs, element_size``) are
-    stamped once at construction."""
+    All eight fields (``kv_size, nl, nb, bs, nh, hs, element_size,
+    block_stride_elems``) are stamped once at construction."""
     dtype: torch.dtype
     """Torch dtype of the KV cache tensors for this group. Used for
     kernel template instantiation; see class docstring for why we keep
     this alongside ``shape_desc.element_size``."""
+    compress_ratio: int = 1
+    """Logical-tokens-per-physical-slot for this group. ``1`` for
+    non-compressed groups (one logical token per physical slot);
+    greater than ``1`` for compressed groups where each physical slot
+    packs ``compress_ratio`` logical tokens (e.g. DeepSeek V4
+    compressor / indexer caches). Derived from
+    ``inference_engine_logical_block_size`` carried in ``layout_hints``
+    at :class:`KVLayerGroupsManager` construction time."""
+    physical_chunk_size: int = 0
+    """Number of *physical* slots in one LMCache chunk for this group
+    (= ``lmcache_logical_chunk_size // compress_ratio``). This is what
+    the block-level transfer kernel must be told, not the logical
+    ``lmcache_logical_chunk_size`` which counts vLLM tokens. ``0``
+    means the field has not been populated yet; ``GPUCacheContext``
+    fills it in after construction once ``lmcache_logical_chunk_size``
+    is known."""
 
     def __repr__(self) -> str:
         if not self.layer_indices:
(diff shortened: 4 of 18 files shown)


liuyumoye:
Related to #3156

biondizzle:
Gave this a test on a 8x B200 node 

I get an OOM. 
```
RuntimeError: Worker failed with error 'CUDA out of memory. Tried to allocate 258.00 MiB. GPU 0 has a total capacity of 178.35 GiB of which 23.81 MiB is free. Including non-PyTorch memory, this process has 178.31 GiB memory in use. Of the allocated memory 173.23 GiB is allocated by PyTorch, and 453.39 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation. 
```

Env vars:
```
- name: NCCL_TOPO_FILE
  value: "/etc/nccl/virtualTopology.xml"
- name: PYTORCH_CUDA_ALLOC_CONF
   value: "expandable_segments:True"
- name: LMCACHE_CONFIG_FILE
   value: "/opt/lmcache-config-dsv4.yaml"
```

Vllm args:
```
spec:
  containers:
    - command:
        - vllm
        - serve
        - deepseek-ai/DeepSeek-V4-Pro
        - '--host'
        - 0.0.0.0
        - '--port'
        - '8000'
        - '--no-enable-prefix-caching'
        - '--tensor-parallel-size'
        - '8'
        - '--trust-remote-code'
        - '--kv-cache-dtype=fp8'
        - '--block-size=256'
        - '--no-enable-flashinfer-autotune'
        - '--attention_config.use_fp4_indexer_cache=True'
        - '--tokenizer-mode=deepseek_v4'
        - '--tool-call-parser=deepseek_v4'
        - '--enable-auto-tool-choice'
        - '--reasoning-parser=deepseek_v4'
        - '--max-model-len=524288'
        - '--kv-transfer-config'
        - '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both"}'
        - '--safetensors-load-strategy=prefetch'
```

LM Cache config:
```
chunk_size: 256
local_cpu: true
max_local_cpu_size: 512.0
enable_lazy_memory_allocator: true
lazy_memory_initial_ratio: 0.2
save_decode_cache: true
use_layerwise: true
save_unfull_chunk: true
blocking_timeout_secs: 60
cache_policy: LRU
```

Running nightly build of vLLM
I reduced context size down to 512K
I tried every gpu memory utilization from .65 to .96

No idea. I will try DP next instead of TP

liuyumoye:
> Gave this a test on a 8x B200 node
> 
> I get an OOM.

Thanks for testing!

A couple of quick questions to help debug:

1. Did you already test that the vanilla vLLM service runs **without LMCache** on this setup? Does the OOM happen at vLLM startup? If it's a startup OOM, please try lowering `--gpu-memory-utilization`, while ensuring the value is still sufficient to fit both model weights and a proper KV cache budget.

2. This PR only supports LMCache **MP (multi-process) mode** for DeepSeek V4. From your description, it looks like you are using in-process mode, which I haven't tested and is not supported yet.

Let me know your setup and I can help further.

ApostaC (commented):
Thanks for the terrific work! I tried to run this PR and it's working! I also learned a lot regarding deepseek v4 kv cache layout from running this PR.

To me, this PR includes 2 main parts:
1. Dealing with the padding in the KV cache
2. Parsing the compression ratio so that we can compute the correct number of slots when launching the kernel.

For 1, I've got another idea (we can discuss further) -- instead of changing the kernel, we recover the "real" shape of the KV cache (i.e., the shape that is with the padding).
In this case, we don't need to introduce the concept of `block stride` into the kernel.

For 2, I've got a few suggestions on how we keep the code cleaner. The high-level idea is that we keep the compression ratio related things inside the `kv_cache_layer_group`. The GPUCacheContext and MPCacheEngine don't need to be aware of this.

H100-H200-B200:
Tried to host with a 8*H20 node, vllm v0.20.1, cuda 13.0
```
lmcache server \
    --host 127.0.0.1 \
    --port 57771 \
    --http-host 127.0.0.1 \
    --http-port 58881 \
    --chunk-size 512 \
    --max-workers 64 \
    --hash-algorithm blake3 \
    --engine-type default \
    --lookup-hash-log-dir /data7/lmcache/lookup_hashes \
    --lookup-hash-log-rotation-interval 21600 \
    --lookup-hash-log-rotation-max-size 104857600 \
    --lookup-hash-log-max-files 100 \
    --l1-size-gb 1200 \
    --l1-use-lazy \
    --l1-init-size-gb 20 \
    --l1-align-bytes 4096 \
    --l1-write-ttl-seconds 600 \
    --l1-read-ttl-seconds 300 \
    --eviction-policy noop \
    --l2-store-policy skip_l1 \
    --eviction-trigger-watermark 0.9 \
    --eviction-ratio 0.1 \
    --l2-prefetch-policy default \
    --l2-prefetch-max-in-flight 8 \
    --l2-adapter '{"type": "nixl_store", "backend": "POSIX", "backend_params": {"file_path": "/data7/lmcache/l2", "use_direct_io": "false"}, "pool_size": 3000}' \
    --prometheus-port 9090 \
    --metrics-sample-rate 0.01
```
The vllm serve command is 
```
vllm serve \
  /data1/public/models/DeepSeek-V4-Flash \
  --served-model-name deepseek-v4-flash \
  --host 127.0.0.1 \
  --port 10081 \
  --trust-remote-code \
  --kv-cache-dtype fp8 \
  --block-size 256 \
  --enable-expert-parallel \
  --tensor-parallel-size 8 \
  --tokenizer-mode deepseek_v4 \
  --tool-call-parser deepseek_v4 \
  --enable-auto-tool-choice \
  --reasoning-parser deepseek_v4 \
  --gpu-memory-utilization 0.92 \
  --kv-transfer-config \
    '{"kv_connector":"LMCacheMPConnector", "kv_role":"kv_both", "kv_connector_extra_config": {"lmcache.mp.host": "127.0.0.1", "lmcache.mp.port": 57771}}'
```
The vllm failed with a seemingly wrongly computed kv cache budget
```
ValueError: To serve at least one request with the models's max seq len (1048576), (318.45 GiB KV cache is needed, which is larger than the available KV cache memory (57.56 GiB). Based on the available memory, the estimated maximum model length is 189440. Try increasing `gpu_memory_utilization` or decreasing `max_model_len` when initializing the engine. See https://docs.vllm.ai/en/latest/configuration/conserving_memory/ for more details.
```
However, dsv4 flash can be served with 1M context without lmcache:
```
vllm serve \
  /data1/public/models/DeepSeek-V4-Flash \
  --served-model-name deepseek-v4-flash \
  --host 127.0.0.1 \
  --port 10081 \
  --trust-remote-code \
  --kv-cache-dtype fp8 \
  --block-size 256 \
  --enable-expert-parallel \
  --tensor-parallel-size 8 \
  --tokenizer-mode deepseek_v4 \
  --tool-call-parser deepseek_v4 \
  --enable-auto-tool-choice \
  --reasoning-parser deepseek_v4 \
  --gpu-memory-utilization 0.92
```
as it estimated
```
(EngineCore pid=976115) INFO 05-08 05:18:20 [kv_cache_utils.py:1708] GPU KV cache size: 8,333,880 tokens
(EngineCore pid=976115) INFO 05-08 05:18:20 [kv_cache_utils.py:1709] Maximum concurrency for 1,048,576 tokens per request: 7.95x
```
Therefore, is there something wrong in kv compute procedure?

liuyumoye:
> Thanks for the terrific work! I tried to run this PR and it's working! I also learned a lot regarding deepseek v4 kv cache layout from running this PR.
> 
> To me, this PR includes 2 main parts:
> 
> 1. Dealing with the padding in the KV cache
> 2. Parsing the compression ratio so that we can compute the correct number of slots when launching the kernel.
> 
> For 1, I've got another idea (we can discuss further) -- instead of changing the kernel, we recover the "real" shape of the KV cache (i.e., the shape that is with the padding). In this case, we don't need to introduce the concept of `block stride` into the kernel.
> 
> For 2, I've got a few suggestions on how we keep the code cleaner. The high-level idea is that we keep the compression ratio related things inside the `kv_cache_layer_group`. The GPUCacheContext and MPCacheEngine don't need to be aware of this.

Thank you for the review and kind words! I really appreciate you taking the time to test and provide detailed feedback.

For point 1, as we discussed in the meeting, after introducing padding, the actual KV cache shape may not be evenly divisible. Therefore, I think keeping a `stride-like` parameter in the kernel would be the most straightforward way to handle this correctly. Of course, I’m also open to any alternative implementations you might suggest.

For point 2, I agree that the current implementation introduces quite a few extra changes for the compression ratio. I will look into whether we can extend the `layout_hints` fields to directly pass the compression ratio and vllm_block_size. This way, we can reduce modifications to `GPUCacheContext` and make the code cleaner.
I appreciate your further suggestions and am happy to discuss more to refine this PR.

liuyumoye:
> Tried to host with a 8*H20 node, vllm v0.20.1, cuda 13.0
> 
> ```
> lmcache server \
>     --host 127.0.0.1 \
>     --port 57771 \
>     --http-host 127.0.0.1 \
>     --http-port 58881 \
>     --chunk-size 512 \
>     --max-workers 64 \
>     --hash-algorithm blake3 \
>     --engine-type default \
>     --lookup-hash-log-dir /data7/lmcache/lookup_hashes \
>     --lookup-hash-log-rotation-interval 21600 \
>     --lookup-hash-log-rotation-max-size 104857600 \
>     --lookup-hash-log-max-files 100 \
>     --l1-size-gb 1200 \
>     --l1-use-lazy \
>     --l1-init-size-gb 20 \
>     --l1-align-bytes 4096 \
>     --l1-write-ttl-seconds 600 \
>     --l1-read-ttl-seconds 300 \
>     --eviction-policy noop \
>     --l2-store-policy skip_l1 \
>     --eviction-trigger-watermark 0.9 \
>     --eviction-ratio 0.1 \
>     --l2-prefetch-policy default \
>     --l2-prefetch-max-in-flight 8 \
>     --l2-adapter '{"type": "nixl_store", "backend": "POSIX", "backend_params": {"file_path": "/data7/lmcache/l2", "use_direct_io": "false"}, "pool_size": 3000}' \
>     --prometheus-port 9090 \
>     --metrics-sample-rate 0.01
> ```
> 
> The vllm serve command is
> 
> ```
> vllm serve \
>   /data1/public/models/DeepSeek-V4-Flash \
>   --served-model-name deepseek-v4-flash \
>   --host 127.0.0.1 \
>   --port 10081 \
>   --trust-remote-code \
>   --kv-cache-dtype fp8 \
>   --block-size 256 \
>   --enable-expert-parallel \
>   --tensor-parallel-size 8 \
>   --tokenizer-mode deepseek_v4 \
>   --tool-call-parser deepseek_v4 \
>   --enable-auto-tool-choice \
>   --reasoning-parser deepseek_v4 \
>   --enable-prompt-tokens-details \
>   --gpu-memory-utilization 0.92 \
>   --kv-transfer-config \
>     '{"kv_connector":"LMCacheMPConnector", "kv_role":"kv_both", "kv_connector_extra_config": {"lmcache.mp.host": "127.0.0.1", "lmcache.mp.port": 57771}}'
> ```
> 
> The vllm failed with a seemingly wrongly computed kv cache budget
> 
> ```
> ValueError: To serve at least one request with the models's max seq len (1048576), (318.45 GiB KV cache is needed, which is larger than the available KV cache memory (57.56 GiB). Based on the available memory, the estimated maximum model length is 189440. Try increasing `gpu_memory_utilization` or decreasing `max_model_len` when initializing the engine. See https://docs.vllm.ai/en/latest/configuration/conserving_memory/ for more details.
> ```
> 
> However, dsv4 flash can be served with 1M context without lmcache:
> 
> ```
> vllm serve \
>   /data1/public/models/DeepSeek-V4-Flash \
>   --served-model-name deepseek-v4-flash \
>   --host 127.0.0.1 \
>   --port 10081 \
>   --trust-remote-code \
>   --kv-cache-dtype fp8 \
>   --block-size 256 \
>   --enable-expert-parallel \
>   --tensor-parallel-size 8 \
>   --tokenizer-mode deepseek_v4 \
>   --tool-call-parser deepseek_v4 \
>   --enable-auto-tool-choice \
>   --reasoning-parser deepseek_v4 \
>   --enable-prompt-tokens-details \
>   --gpu-memory-utilization 0.92
> ```
> 
> as it estimated
> 
> ```
> (EngineCore pid=976115) INFO 05-08 05:18:20 [kv_cache_utils.py:1708] GPU KV cache size: 8,333,880 tokens
> (EngineCore pid=976115) INFO 05-08 05:18:20 [kv_cache_utils.py:1709] Maximum concurrency for 1,048,576 tokens per request: 7.95x
> ```
> 
> Therefore, is there something wrong in kv compute procedure?

Thanks a lot for your testing and feedback!

vLLM provides a `hybrid-kv-cache-manager `option for KV cache memory management. You can check the runtime logs for reference:

This option is enabled by default when LMCache is not used. However, it will be forcibly disabled when enabling LMCache, because LMCache does not support the hybrid KV cache manager for now.

Enabling `hybrid-kv-cache-manager` can effectively reduce KV cache memory consumption, which is probably the reason why you are seeing insufficient GPU memory during testing.

Once this PR gets approved and merged, we will extend the implementation to support `hybrid-kv-cache-manager `with LMCache enabled.

csrc/mp_mem_kernels.cuh
```diff
@@ -30,6 +46,21 @@ struct PageBufferShapeDesc {
...
+  __host__ __device__ inline size_t block_stride_or(
+      size_t tight_in_scalar_type) const {
+    if (block_stride_elems <= 0) {
+      return tight_in_scalar_type;
+    }
+    // block_stride_elems counts elements of the source KV dtype
+    // (element_size bytes each); convert to ScalarType units.
+    return static_cast<size_t>(block_stride_elems) * element_size /
+           sizeof(ScalarType);
+  }
```
ApostaC: ~~Is it possible to get rid of the input argument for this function. I.e., if `block_stride_elems == 0`, this function will return `this->scalars_per_block<ScalarType>() * this->kv_size`~~

ApostaC: Actually, after reading the code again, how about directly changing the implementation of `scalars_per_block`? Then, we can probably minimize the changes in `mp_mem_kernes.cu`. WDYT?

csrc/mp_mem_kernels.cu
```diff
@@ -39,16 +39,29 @@ __device__ inline size_t calculate_engine_global_offset(
...
+    // dim-0 is the block axis → honour block_stride_elems if the caller
+    // populated it (needed for dim-0-padded vLLM KV pools where the
+    // group's row is padded up to the pool's max row width, e.g.
+    // DeepSeek V4 compressor / indexer caches). Fallback to the tight
+    // stride ``kv_size * scalars_per_block`` preserves legacy behaviour
+    // for untouched callers.
+    const size_t tight_block_stride = shape_desc.kv_size * scalars_per_block;
+    const size_t per_block =
+        shape_desc.block_stride_or<ScalarType>(tight_block_stride);
+    return engine_block_idx * per_block + k_or_v * scalars_per_block;
```
ApostaC: In the previous code, we have `kv_size` when we calculate the block stride (see `tight_block_stride`).
But in the new code (the implementation of `block_stride_or`) does not use the `kv_size` at all. Why is that?

It may work because for deepseek v4, the kv_size is 1. But will it create problems in the future when some models have`kv_size` not 1?

lmcache/integration/vllm/lmcache_mp_connector_0180.py
ApostaC, line 1: @maobaolong @chunxiaozheng can you help double-check whether we need to update the `connector_0180` here?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -33,7 +33,22 @@
...
+        for name, tensor in kv_caches.items()
+    ]
+    logger.info(
+        "KV cache transfer keeping %d layer(s) (name, shape, dtype):\n%s",
+        len(kept_summary),
+        "\n".join(
+            f"  [{i}] {name}  shape={shape}  dtype={dtype}"
+            for i, (name, shape, dtype) in enumerate(kept_summary)
+        ),
+    )
```
ApostaC: Do you think it's better to change this to the `debug` loglevel?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -676,7 +697,7 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
 
         layout_hints = vllm_layout_hints()
         self.kv_caches = kv_caches
-
+        logger.info(f"model_name: {self.model_name}")
```
ApostaC: nit: we probably don't need this line of log?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -686,6 +707,7 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
                 self.model_name,
                 self.world_size,
                 EngineType.VLLM,
+                self.vllm_block_size,
```
ApostaC: As we discussed offline, we can put the `vllm_block_size` into the layout hint dictionary. In this case, we don't need to change the protocol

lmcache/v1/gpu_connector/utils.py
```diff
@@ -94,23 +94,115 @@ def attempt_permute_to_contiguous_view(
...
+        # permuted view), not ``kv_caches``. For tensors that are both
+        # permuted (e.g. vLLM HND physical layout exposed as NHD logical
+        # shape) and dim-0-padded, the original's inner-dim strides are
+        # unsorted and would falsely trip the ``stride[-2] == shape[-1]``
+        # tight-packing check. After stride-sorting by ``permute``, the
+        # inner dims of ``result`` are guaranteed to be in descending
+        # stride order, so only dim-0 (outermost) can legitimately carry
+        # padding. ``permute`` shares storage and preserves
+        # ``storage_offset`` / ``numel`` / ``storage_nbytes``, so those
+        # checks are equivalent on either view.
```
ApostaC: I'm good with the implementation of the padding detection and tracking. 
For better code clarity, can we make this an individual helper function?

lmcache/v1/gpu_connector/utils.py
```diff
@@ -943,10 +1042,34 @@ def get_device(kv_caches: DiscoverableKVCache) -> torch.device:
...
+                "layer survives filtering before calling GPUCacheContext."
+            )
+        path.append(0)
         probe = probe[0]
+    if not isinstance(probe, torch.Tensor):
+        raise ValueError(
+            "get_device: descent reached a non-tensor leaf of type "
+            f"{type(probe).__name__} at path {path}; expected a "
+            "torch.Tensor. KV cache structure is malformed."
+        )
```
ApostaC: I understand this part of the code. Just wanted to check is this related to dsv4 support? Or a generic failure message improvement?

lmcache/v1/gpu_connector/utils.py
```diff
@@ -967,6 +1091,15 @@ def make_page_buffer_shape_desc(
...
         block_size: Tokens per block (``bs``).
+        block_stride_elems: Physical per-block stride in *elements*
+            (= ``tensor.stride(0)`` of the representative layer). If
+            ``None``, defaults to the tight stride derived from
+            ``block_size``, ``num_heads``, ``head_size``, and
+            ``kv_size``. Pass the real value whenever the group's KV
+            pool may be dim-0-padded (e.g. DeepSeek V4
+            compressor/indexer caches sharing a row width with a
+            larger group in the same pool); otherwise downstream
+            transfer kernels will skip into padding and corrupt data.
```
ApostaC: IIRC, in `mp_mem_kernels.cuh`, we assume that the `block_stride_elems` should be 0 when the tensor is not padded. Does that mean the expected behavior should set the value of `block_stride_elems` to 0 instead of calculating the tight stride here?

In this case, we assume all the complexities are handled at the same level (i.e., the kernel implementation), and we can get rid of the changes in this function.

lmcache/v1/multiprocess/protocols/engine.py
ApostaC, line 1: As mentioned above, if we use layout hints to pass the vllm block size, we don't need to change the protocol here.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -64,33 +63,87 @@ class GPUCacheContext:
...
+    _MAX_BLOCK_IDS = 1_000_000
+    # Max number of chunks processed concurrently on ``tmp_gpu_buffer_``.
+    _MAX_BATCH_SIZE = 4
+
     def __init__(
         self,
         kv_caches: KVCache,
-        lmcache_chunk_size: int = 256,
+        lmcache_chunk_size: int,
+        vllm_block_size: int,
```
ApostaC: I'm wondering whether we can avoid `GPUCacheContext` to have `vllm_block_size`.
The major reason is that the `GPUCacheContext` should be serving-engine-agnostic, which means that it will be super weird to have a "vllm-specific" variable in the `__init__` function..

Implementation-wise, I feel like the logic related to the "real block size" can be all encapsulated into the `KVLayerGroupManager`.

I may have a few related comments below.

lmcache/v1/multiprocess/server.py
```diff
@@ -110,22 +110,37 @@ def compute_extra_count(
...
+    shapes = []
+    for group_idx in range(num_groups):
+        compress_ratio = gpu_context.get_group_compress_ratio(group_idx)
+        if num_tokens % compress_ratio != 0:
+            raise ValueError(
+                f"num_tokens ({num_tokens}) is not a multiple of "
+                f"compress_ratio ({compress_ratio}) for group {group_idx}"
+            )
+        group_num_slots = num_tokens // compress_ratio
+        shapes.append(gpu_context.get_kv_buffer_shape(group_num_slots, group_idx))
```
ApostaC: `GPUCacheContext.get_kv_buffer_shape` calls `KVLayerGroupManager` to get the shape information.
Therefore, if we move the real block size and the compression ratio calculation into `KVLayerGroupManager`, we can directly call the `get_kv_buffer_shape` without dealing with the compression ratio here (i.e., no change is needed in this function)

lmcache/v1/multiprocess/gpu_context.py
ApostaC: Regarding the argument passing, the `vllm_block_size` should be contained in `layout_hints` -- which contains engine-specific implementations.
`GPUCacheContext` class does not use the layout hints directly. Instead, it passes it to the `KVLayerGroupManager` during the `register_kv_cache`.

The KVLayerGroupManager will further call the utility function defined in `gpu_connector/utils.py` to check the real KV cache size, and also extract the compression ratio. Those utility functions may have specialized logics to parse the serving-engine-related information in `layout_hints`.

This workflow was done by @sammshen . @sammshen can you help check if what I said above is correct?

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -100,48 +153,85 @@ def __init__(
...
-        self.max_batch_size = 4
-        self.lmcache_chunk_size = lmcache_chunk_size
-        # Byte size of one chunk entry (= one chunk across all groups).
-        # tmp_chunk_group_offsets_[g] is the byte offset of group g within
-        # a single chunk; tmp_chunk_group_offsets_[num_groups] ==
-        # tmp_chunk_bytes_.
+    def _init_compression_metadata(
+        self,
+        lmcache_chunk_size: int,
+        vllm_block_size: int,
```
ApostaC: Logically, the compression metadata can be encapsulated into `KVLayerGroupManager`, and this function can be moved there.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -64,33 +63,87 @@ class GPUCacheContext:
...
+        # chunks laid out in chunk-major order.
+        self.tmp_gpu_buffer_ = torch.empty(
+            self.tmp_chunk_bytes_ * self.max_batch_size,
+            dtype=torch.uint8,
+            device=self.device_,
+        )
+
+        # Phase 6: create the CUDA / CuPy streams (regular + high prio)
+        # and emit the init log from within the stream.
+        self._init_streams()
```
ApostaC: If we move `_init_compression_metadata` to `KVLayerGroupManager`, the remaining changes seem to be just to reorder the code a bit. 
I guess we don't need to introduce those changes in this PR; we should keep it as close to the original code as possible.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -195,11 +284,53 @@ def high_priority_cupy_stream(self) -> cupy.cuda.Stream:
...
-    def block_size(self) -> int:
+    def vllm_block_size(self) -> int:
+        """
+        Returns the vLLM-side block size (number of *logical* tokens per
+        vLLM block). For non-compressed groups this matches each group's
+        physical slot count; for compressed groups, each vLLM block maps
+        to ``vllm_block_size // compress_ratio`` physical slots.
         """
-        Returns the block size (number of tokens per block)
+        return self.vllm_block_size_
```
ApostaC: Here, let's try keeping the old `block_size` property instead of introducing the new `vllm_block_size`.

I see `server.py` uses this function, and the current changes seem to be just a simple rename.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -195,11 +284,53 @@ def high_priority_cupy_stream(self) -> cupy.cuda.Stream:
...
+
+    def get_group_compress_ratio(self, group_idx: int) -> int:
+        """Returns the compression ratio (logical tokens / physical slot)
+        for the given KV layer group. Equal to 1 for non-compressed groups.
+        """
+        return self.group_compress_ratios_[group_idx]
```
ApostaC: The functionality of this method can be moved to `KVLayerGroupsManager`

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -195,11 +284,53 @@ def high_priority_cupy_stream(self) -> cupy.cuda.Stream:
...
+
+    def get_group_lmcache_chunk_slots(self, group_idx: int) -> int:
+        """Returns the number of *physical* slots per chunk for the given
+        KV layer group (= ``lmcache_chunk_size // compress_ratio``).
+        This is what the block-level kernel must be told, not the logical
+        ``lmcache_chunk_size`` which represents vLLM tokens.
+        """
+        return self.group_lmcache_chunk_slots_[group_idx]
```
ApostaC: I agree we need this function here. But the implementation can become something like `lmcache_chunk_size // self.kv_layer_groups_manager_.get_compression_ratio(group_idx)`

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -195,11 +284,53 @@ def high_priority_cupy_stream(self) -> cupy.cuda.Stream:
...
+
+    def get_group_block_size(self, group_idx: int) -> int:
+        """Returns the *physical* per-block slot count for the given KV
+        layer group (i.e. ``shape_desc.bs``).
+        """
+        return self.kv_layer_groups_manager_.kv_layer_groups[group_idx].shape_desc.bs
```
ApostaC: Just realized that now we have 2 different semantics of block sizes. Here's my understanding
1. The logical block size (`self.block_size`/`self.vllm_block_size`): it captures how many logical tokens are in each block of the KV caches.
2. The physical block size: it captures the size of the "token dimension" in the underlying kv cache (e.g., if logical block size = 256, the physical block size will be 64 for C4 layers, and 2 for C128 layers)

If my understanding is correct, we might have to make the naming super clear in the code with docstrings. In addition, we need to be careful that when to use which block size in `server.py`

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -195,11 +284,53 @@ def high_priority_cupy_stream(self) -> cupy.cuda.Stream:
...
+    def group_block_sizes(self) -> list[int]:
+        """Per-group physical slot count (``shape_desc.bs``) in group
+        order. For non-compressed groups this equals ``vllm_block_size``;
+        for compressed groups it equals ``vllm_block_size //
+        compress_ratio``.
+        """
+        return [
+            group.shape_desc.bs
+            for group in self.kv_layer_groups_manager_.kv_layer_groups
+        ]
```
ApostaC: Is this function physical block size or logical block size?

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -195,11 +284,53 @@ def high_priority_cupy_stream(self) -> cupy.cuda.Stream:
...
+
+    @property
+    def group_compress_ratios(self) -> list[int]:
+        """Per-group compression ratio (= ``vllm_block_size //
+        shape_desc.bs``) in group order. ``1`` for non-compressed groups.
         """
-        return self.block_size_
+        return list(self.group_compress_ratios_)
```
ApostaC: We can have our definition of the `compress_ratio` to be `logical_block_size // physical_block_size`. This definition avoids any serving engine-related details -- it's an internal attribute of the KV cache.

lmcache/v1/multiprocess/server.py
```diff
@@ -304,7 +327,12 @@ def store(
         gpu_context = self.gpu_contexts[instance_id]
         model_name = self.gpu_context_meta[instance_id][0]
 
-        blocks_per_chunk = self.chunk_size // gpu_context.block_size
+        # ``blocks_per_chunk`` is counted in vLLM-side blocks (each block
+        # addresses ``vllm_block_size`` *logical* tokens). For compressed
+        # groups the per-group physical slot count differs, but the
+        # block-id indexing is shared with vLLM and therefore uses the
+        # vLLM block size here.
+        blocks_per_chunk = self.chunk_size // gpu_context.vllm_block_size
```
ApostaC: Quick question: `block_per_chunk` is going to be used below to slice on the `block_ids` list. Should we calculate this using physical block size or logical block size? (I'm a bit not sure here)

lmcache/v1/multiprocess/server.py
```diff
@@ -932,12 +976,19 @@ def end_session(self, request_id: str) -> None:
...
-            logger.warning(
-                "Session %s has no lookup ipc key, skipping touch", request_id
-            )
+            # Normal: the request's ``lookup()`` RPC took an early-return
+            # branch before it could stamp ``lookup_ipc_key`` on the
+            # session. Typical triggers:
+            #   - prompt shorter than ``chunk_size`` → empty chunk_hashes;
+            #   - model/world_size mismatch (layout_desc is None).
+            # There is nothing to touch in L1 in any of these cases.
+            logger.debug("Session %s has no lookup ipc key, skipping touch", request_id)
```
ApostaC: Is this change related to the dsv4 integration? If not, we can remove it. (There is a similar change below)

lmcache/v1/kv_layer_groups.py
ApostaC, line 1: Haven't done a detailed pass on this file yet. But if we are going to move the compression metadata into this module, I expect there are a lot of incoming changes -- so I will come back to this file after the updates.

ApostaC (changes requested):
Thanks again for the great effort. I think the major recommendation is to move the compression ratio related logic into `KVLayerGroupsManager` and keep GPUCacheContext "serving-engine-agnostic".

A minor thing is to clarify on the "physical block size" vs "logical block size".

Please take a look on the detailed comments below, thanks!

csrc/mp_mem_kernels.cu
liuyumoye: Great catch, this is worth clarifying.
1. `block_stride_elems` is defined as `tensor.stride(0)`, which by PyTorch stride semantics already absorbs the `kv_size` factor for `NL_X_NB_TWO_BS_NH_HS` (the `dim‑1 `step lives inside `dim‑0`'s stride). Multiplying by kv_size again would double‑count and skip into the next block.
2.  `kv_size` only appears in the fallback because that branch reconstructs the tight stride from shape; once a real stride(0) is supplied, it's authoritative — so this works for any future model with kv_size != 1 as long as K/V slices stay tightly packed within a block (currently enforced by `attempt_permute_to_contiguous_view`, which rejects dim‑1 padding). I've tightened the block_stride_elems docstring to make this contract explicit.

>  tight_block = 2 * BS * NH * HS
padded_block = stride(0)
pad = padded_block - tight_block
block 0  ──┬─ K cache (BSxNHxHS) ─┬─ V cache (BSxNHxHS) ─┬─ pad ─┐

csrc/mp_mem_kernels.cuh
liuyumoye: Totally agree we should minimize the` mp_mem_kernes.cu` diff — but I think folding into `scalars_per_block` itself would actually break things.
The problem is that `scalars_per_block` plays two distinct roles in the same expression — outer block step and the K↔V switch within a block. 
```
return engine_block_idx * per_block + k_or_v * scalars_per_block;
//                       ↑                       ↑
//                outer block step          K↔V switch (must stay tight) 
```
The K↔V use must stay tight (LMCache's layout protocol forbids dim‑1 padding), so making scalars_per_block padding‑aware would land V offsets inside the trailing pad. Instead I added a small `per_block_stride<ScalarType>(scalars_per_block_tight)` accessor that reuses the already‑computed tight value, so the call sites collapse to two symmetric one‑liners and there's no duplicate arithmetic — same .cu minimisation goal, cleaner separation. PTAL.

lmcache/integration/vllm/lmcache_mp_connector_0180.py
liuyumoye: I share the same concern. I'm wondering if this issue has already been resolved in the latest vLLM mp_connector snapshot.

lmcache/integration/vllm/vllm_multi_process_adapter.py
liuyumoye: I think so too.

lmcache/integration/vllm/vllm_multi_process_adapter.py
liuyumoye: Yes, you are right

lmcache/integration/vllm/vllm_multi_process_adapter.py
liuyumoye: Done.

lmcache/v1/gpu_connector/utils.py
liuyumoye: Done.

lmcache/v1/gpu_connector/utils.py
liuyumoye: Good point — this isn't DSv4-specific, just a fail-fast cleanup that happened to surface during bring-up. I've reverted it from this PR to keep the diff focused, and will send it separately as a small standalone improvement. ✅

lmcache/v1/gpu_connector/utils.py
liuyumoye: Done, you are right.

lmcache/v1/multiprocess/protocols/engine.py
liuyumoye: I have added a new parameter `inference_engine_logical_block_size` in the layout hints to pass the block size.Please let me know if you have any concerns or suggestions about the parameter name.

lmcache/v1/multiprocess/server.py
liuyumoye: Perfect, I've removed this unnecessary change.

lmcache/v1/multiprocess/server.py
liuyumoye: Good question. The current implementation uses logical block size, here's why:

`self.chunk_size` in LMCache is the logical chunk size (counted in logical tokens), and `inference_engine_logical_block_size`, as its name suggests, is also a quantity in the logical domain (how many logical tokens each engine block corresponds to). Dividing one by the other gives `blocks_per_chunk`, which is exactly what's needed to slice on `gpu_block_ids` — and `gpu_block_ids` itself is the list of logical block ids on the engine side, so the units line up.

lmcache/v1/multiprocess/gpu_context.py
liuyumoye: I've updated it. The value obtained here is the physical block size.

liuyumoye:
> Thanks again for the great effort. I think the major recommendation is to move the compression ratio related logic into `KVLayerGroupsManager` and keep GPUCacheContext "serving-engine-agnostic".
> 
> A minor thing is to clarify on the "physical block size" vs "logical block size".
> 
> Please take a look on the detailed comments below, thanks!

Thanks for the thorough review! I've addressed both points:

1. Moving `compression-ratio` logic into KVLayerGroupsManager

I've reworked the implementation so that `GPUCacheContext` stays `serving-engine-agnostic` — it no longer needs to know anything about compression ratios. All compression-ratio handling now lives inside` KVLayerGroupsManager`, and`GPUCacheContext.get_kv_buffer_shape` simply delegates to it without any extra adjustment on its side.

2. Clarifying "physical block size" vs "logical block size"

To make the terminology consistent across the two layers, the naming now follows this convention:

From the inference engine (vLLM) side:

`LayoutHints.inference_engine_logical_block_size` — the logical block size exposed by the engine.

`KVLayerGroupInfo.shape_desc.bs`— the physical block size, per group (i.e. after applying that group's compression ratio).

From the LMCache side:

`lmcache_logical_chunk_size` — the logical chunk size used by LMCache.

`KVLayerGroupInfo.physical_chunk_size` — the physical chunk size, per group.

So on each side we have a single logical quantity shared across groups, and a per-group physical quantity derived from it via the group's compression ratio. Hopefully this makes the boundary between the two concepts clearer.

Let me know if anything still looks off — happy to iterate further!

csrc/mp_mem_kernels.cu
ApostaC: Thanks for the explanation!

After revisiting the code, I think `scalars_per_block()` is the one that actually confuses people here. To better explain:

From what we have seen right now, all the different GPU KV cache layouts have 2 "parts" of dimensions:
1. `[num layers, k_v size, num_blocks]` -- this 3 dimensions could permute internally
2. `[block size, num heads, head size]` -- this 3 dimensions could also permute internally
The MP memcpy kernel is also designed based on this assumption: a single GPU thread block takes care of the `[block size, num heads, head size]` chunk of elements, and we have `num layers * k_v size * num_blocks` thread blocks launched at the same time.

Based on this, the actual behavior of `scalars_per_block()` is to compute how many elements are in the `[block size, num heads, head size]` chunk. The better name should be something like `scalars_per_layer_kv_block()`.

Going back to the current implementation, is it true that the `block_stride_elems` is a concept on the `BS x NH x HS` chunk? Or it potentially depends on the order of `[num layers, k_v size, num_blocks]`?
- If it's a concept on `BS x NH x HS` chunk, we can safely merge the `per_block_stride()` function with the current `scalars_per_block()` function.
- If not, then we need to calculate the real size based on the `GPUKVFormat` -- because different GPU KV formats may need different calculations for the "tight stride".

lmcache/v1/gpu_connector/utils.py
```diff
@@ -94,27 +105,138 @@ def attempt_permute_to_contiguous_view(
...
+            padding_per_block,
+            int(result.untyped_storage().nbytes()),
+            result.dtype,
+        )
         return result
     return [attempt_permute_to_contiguous_view(sub) for sub in kv_caches]
 
 
+def _validate_dim0_padded_layout(tensor: torch.Tensor) -> int:
+    """Validate that *tensor* matches the dim-0-padding-only strided layout.
```
ApostaC, line 154: nit: it's better to add a sentence in the first paragraph saying "this is mainly used for deepseek v4 integration"

lmcache/v1/kv_layer_groups.py
```diff
@@ -148,66 +167,352 @@ def __init__(
...
                 )
             )
 
         logger.info("KV layer groups: %s", self.kv_layer_groups)
 
+    @staticmethod
+    def _derive_compression_metadata(
+        group_idx: int,
+        bs: int,
+        ie_logical_block_size: "int | None",
```
ApostaC, line 272: Just wondering, in what case, `is_logical_block_size` will be `None`?

lmcache/v1/kv_layer_groups.py
```diff
@@ -148,66 +167,352 @@ def __init__(
...
+        # indices remain deterministic across runs.
+        for group_idx, ((_, _, _, bs, dt), indices) in enumerate(
+            sorted(groups_dict.items(), key=lambda kv: kv[1][0])
         ):
+            rep = self._resolve_representative_tensor(
+                kv_caches, gpu_kv_format, indices[0]
+            )
+            block_stride_elems = self._resolve_block_stride(
+                rep, gpu_kv_format, indices[0]
+            )
```
ApostaC: Let's merge the implementation of these two functions (plus the `_log_representative_tensor_layout` below) into a single one and put it into `gpu_connector/utils.py`.

Here's the reason
- We wanted to put all the `GPUKVFormat`-related resolution logic into the same place, which is `gpu_connector/utils.py`. In this case, other upper-level modules don't need to do manual parsing of the KV cache format.
- We don't want any module to hold anything like "representative kv cache". This is because there are a lot of potential KV cache formats, and it's hard to track the semantics and expected shape of the "representative KV cache".

ApostaC (commented):
Almost good to go! Please see the comments below.

lmcache/v1/kv_layer_groups.py
liuyumoye: As we discussed offline, is_logical_block_size can be None when the upper layer uses other inference engines, such as TRT or SGLang.

csrc/mp_mem_kernels.cu
liuyumoye: As we discussed offline, `block_stride_elems` is defined **only for the BS × NH × HS** dimension chunk.Therefore, padding exists on the K/V dimension under the` NL_X_NB_TWO_BS_NH_HS `GPUKVFormat.
During my refactoring, I found that if we make `block_stride_elems` select its dimension calculation dynamically according to `GPUKVFormat`, we will also need to modify `attempt_permute_to_contiguous_view` accordingly.Currently this function only assumes padding exists only on `stride(0)`. If we support more formats, we have to restrict padding to different dimensions based on each `GPUKVFormat`.
At present, only **DeepSeek V4 (NL_X_NB_BS_HS)** hits this padding scenario, and we cannot anticipate which dimension will have padding for other unknown `GPUKVFormat` cases.
For this reason, I prefer to narrow the scope of this PR:we only handle the `NL_X_NB_BS_HS` case for now. We can add support and adapt other `GPUKVFormat` later once new padding scenarios actually appear.

lmcache/v1/kv_layer_groups.py
liuyumoye: Done.

lmcache/v1/gpu_connector/utils.py
liuyumoye: Done

liuyumoye:
> Almost good to go! Please see the comments below.

As we discussed offline,` block_stride_elems` is defined only for the **BS × NH × HS** dimension chunk.

Therefore, padding exists on the K/V dimension under the **NL_X_NB_TWO_BS_NH_HS** GPUKVFormat.

During my refactoring, I found that if we make `block_stride_elems` select its dimension calculation dynamically according to **GPUKVFormat**, we will also need to modify `attempt_permute_to_contiguous_view` accordingly. 

Currently `attempt_permute_to_contiguous_view` function only assumes padding exists only on stride(0). If we support more formats, we have to restrict padding to different dimensions based on each GPUKVFormat.

At present, only **DeepSeek V4 (NL_X_NB_BS_HS)** hits this padding scenario, and we cannot anticipate which dimension will have padding for other unknown GPUKVFormat cases.

For this reason, I prefer to narrow the scope of this PR:we only handle the **NL_X_NB_BS_HS** case for now. We can add support and adapt other GPUKVFormat later once new padding scenarios actually appear.

lmcache/v1/gpu_connector/utils.py
```diff
@@ -950,13 +1070,178 @@ def get_device(kv_caches: DiscoverableKVCache) -> torch.device:
...
+        lmc_ops.GPUKVFormat.NL_X_NB_BS_HS,
+    }
+)
+
+
+def _pick_layout_probe_tensor(
+    kv_caches: DiscoverableKVCache,
+    gpu_kv_format: "lmc_ops.GPUKVFormat",
+    layer_idx: int,
+) -> torch.Tensor:
```
ApostaC: nit: since this function is only used in `resolve_block_stride_and_log_layout`, we can put it inside that function.

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -55,6 +55,14 @@ def get_protocol_definitions() -> dict[str, ProtocolDefinition]:
...
         #     caches (vLLM, SGLang, ...). Drives format detection.
         #   - layout_hints: LayoutHints - See custom_types.LayoutHints.
+        #     For vLLM,
+        #     ``layout_hints["inference_engine_logical_block_size"]``
+        #     carries the logical tokens-per-engine-block (previously a
+        #     standalone argument); the server reads it from here to
+        #     derive per-group compression ratios for compressed KV
+        #     layer groups (where
+        #     ``shape_desc.bs <
+        #     inference_engine_logical_block_size``).
```
ApostaC: nit: proably we don't need this because we have enough docstrings in LayoutHints class definition already

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -83,16 +82,19 @@ def __init__(
...
+        self.lmcache_logical_chunk_size = lmcache_logical_chunk_size
 
-        # Build per-layer KV groups. The manager owns each group's
-        # PageBufferShapeDesc (kernel-facing shape); this context only
-        # retains per-group GPU resources (pointer tensors, tmp buffers).
+        # ``layout_hints`` carries (among other things) the inference
+        # engine's logical block size, which the manager needs to
+        # derive per-group compression metadata. Forwarded as-is so
+        # that engines without that concept (TRT-LLM / SGLang) can
+        # simply omit the key.
```
ApostaC: nit: same as above, this comment is not necessary.

lmcache/v1/multiprocess/server.py
```diff
@@ -236,8 +236,18 @@ def register_kv_cache(
...
+                pulling out ``inference_engine_logical_block_size`` and
+                using the rest for GPU KV format detection. For vLLM,
+                ``layout_hints["inference_engine_logical_block_size"]``
+                must be present and carries the logical
+                tokens-per-engine-block; without it ``GPUCacheContext``
+                cannot derive per-group compression ratios when some
+                KV layer groups compress multiple logical tokens into a
+                single physical slot
+                (``shape_desc.bs <
+                inference_engine_logical_block_size``).
```
ApostaC: nit: let's keep the old docstrings here.

lmcache/v1/multiprocess/server.py
```diff
@@ -985,7 +1021,18 @@ def report_status(self) -> dict:
...
                 entry["kv_cache_layout"] = {
                     "num_layers": ctx.num_layers,
-                    "block_size": ctx.block_size,
+                    # ``inference_engine_logical_block_size`` is the
+                    # scalar "logical tokens per engine block"
+                    # addressing unit. ``group_physical_block_sizes``
+                    # is the per-KV-layer-group physical slot count;
+                    # for compressed groups it is smaller than
+                    # ``inference_engine_logical_block_size`` by
+                    # ``group_compress_ratios``.
```
ApostaC: nit: this docstring is duplicated and we can remove it -- there is one above.

ApostaC (approved):
LGTM! Left some nit comment (mainly about docstrings) but giving an approve here!

lmcache/v1/multiprocess/server.py
liuyumoye: Done.

lmcache/v1/multiprocess/server.py
liuyumoye: Done.

lmcache/v1/multiprocess/gpu_context.py
liuyumoye: Done.

lmcache/v1/multiprocess/protocols/engine.py
liuyumoye: Done.

lmcache/v1/gpu_connector/utils.py
liuyumoye: Done.

lmcache/v1/multiprocess/protocols/engine.py
maobaolong: @liuyumoye How about this nit?

liuyumoye: > [@liuyumoye](https://github.com/liuyumoye) How about this nit?


Sorry for the duplicate code here. I missed this spot and have already removed it.

lmcache/v1/gpu_connector/utils.py
chunxiaozheng: @liuyumoye could you add the sentence?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -33,7 +33,22 @@
...
 def wrap_kv_caches(kv_caches: dict[str, torch.Tensor]) -> KVCache:
-    logger.info("KV caches keys are %s", list(kv_caches.keys()))
+    # Emit a per-layer (name, shape, dtype) summary so the operator can
+    # verify the exact layer set & tensor geometry being shipped to the
+    # LMCache server, then the low-noise count of handles being wrapped.
+    kept_summary = [
+        (name, tuple(tensor.shape), str(tensor.dtype))
+        for name, tensor in kv_caches.items()
+    ]
+    logger.debug(
```
maobaolong, line 43: @liuyumoye This log is really helpful, do you think it is a good idea to set the log level to `info`? Since the `wrap_kv_caches` only be invoked on startup stage, but i'm not sure the exactly for the real log size for this line, can you paste it from your runtime logger file? Thx.

liuyumoye: The log level was initially set to info,  but @ApostaC suggested changing it to debug because the number of log lines is equal to the number of attention layers.For DS V4 Flash models, this would print 167 lines, so setting it to info is not ideal.

lmcache/v1/gpu_connector/utils.py
```diff
@@ -94,27 +105,129 @@ def attempt_permute_to_contiguous_view(
...
+            result.dtype,
+        )
         return result
     return [attempt_permute_to_contiguous_view(sub) for sub in kv_caches]
 
 
+def _validate_dim0_padded_layout(tensor: torch.Tensor) -> int:
+    """Validate that *tensor* matches the dim-0-padding-only strided layout.
+
+    Mainly used for DeepSeek V4 integration, where compressor / indexer
```
liuyumoye, line 156: @chunxiaozheng add here

lmcache/v1/kv_layer_groups.py
```diff
@@ -148,66 +170,183 @@ def __init__(
...
         )
 
+        # Pull the inference-engine logical block size out of
+        # ``layout_hints`` once; ``None`` means no compression info
+        # available and every group is treated as non-compressed below.
+        # The attribute is finalised after the group-building loop
+        # below, where ``None`` is replaced by the first group's
+        # physical ``bs`` so the public ``int`` contract holds.
+        self.inference_engine_logical_block_size_: "int | None" = (
+            layout_hints.get("inference_engine_logical_block_size")
```
maobaolong, line 208: We'd better to abstract a constant for these new defined name of layout_hint, so that we can put/get by a same constant, this can avoid partial update while renaming

liuyumoye: Thanks for the suggestion! That’s a good point to avoid potential errors during renaming.

Actually, LayoutHints is a TypedDict, so the field name itself already serves as the single source of truth — type checkers (pyright/mypy) will flag any mistyped key at the call site, and IDE rename refactors can propagate across all use sites automatically.

Wrapping each field in a separate constant would actually defeat TypedDict's literal-key value-type inference (PEP 589) and force us to keep two names in sync per field.

Happy to revisit this if we ever migrate to plain `dict[str, Any]`in the future.

maobaolong (approved):
Left some minor comments, otherwise, lgtm

chunxiaozheng (approved):
LGTM!

H100-H200-B200:
A small query, as stated in https://github.com/vllm-project/vllm/releases/tag/v0.21.0
> Engine Core
KV offloading + HMA: Scheduler-side sliding window groups (https://github.com/vllm-project/vllm/pull/41228), full HMA enablement (https://github.com/vllm-project/vllm/pull/41445), multi-connector HMA (https://github.com/vllm-project/vllm/pull/39571), per-job store completion (https://github.com/vllm-project/vllm/pull/39186), DCP/PCP support in OffloadingConnector (https://github.com/vllm-project/vllm/pull/41549), MooncakeStoreConnector for distributed KV offloading (https://github.com/vllm-project/vllm/pull/40900).

I wonder if within this PR, DSV4 can be served with vllm 0.21.0 + `--no-disable-hybrid-kv-cache-manager ` enable?

PR #3175, opened 2026-04-30
[Observability] Per-request hit rate attributes on root OTel spans

deng451e:
Description:

Summary

Adds hit_tokens, requested_tokens, and hit_rate (float, 0.0–1.0) as attributes on the root OTel span for both the MP server and Cache Blending code paths.

- MP path ("request" span): attributes are set when MP_LOOKUP_PREFETCH_END is processed by MPServerTracingSubscriber. Data was already in the event — no
server changes needed.
- CB path ("cb.request" span): attributes are set when CB_LOOKUP_END is processed by BlendTracingSubscriber. All three CB_LOOKUP_END emit sites in
blend_server_v2.py are extended to carry hit_tokens and requested_tokens (chunk-aligned, matching the MP server convention).

hit_rate is stored as a precomputed float so it can be filtered directly in Tempo/Jaeger without derived-metric support (e.g. { name = "request" &&
span.hit_rate < 0.5 }).
[
Screenshot 2026-04-30 at 12 05 26 PM
](url)

---

> [!NOTE]
> **Low Risk**
> Low risk observability-only change that adds additional OTel span attributes and event metadata; main risk is mismatched/missing metadata causing incorrect hit-rate values rather than functional impact.
> 
> **Overview**
> Adds per-request cache hit rate visibility to tracing by attaching `hit_tokens`, `requested_tokens`, and precomputed `hit_rate` attributes to the MP root `request` span (on `MP_LOOKUP_PREFETCH_END`) and the CB root `cb.request` span (on `CB_LOOKUP_END`).
> 
> Extends CB `CB_LOOKUP_END` emission in `blend_server_v2.py` to always include `hit_tokens` and chunk-aligned `requested_tokens`, and updates docs/examples with Tempo/Jaeger query patterns for filtering by `span.hit_rate`.
> 
> Adds new unit tests using an in-memory OTel exporter to verify attributes are recorded correctly (including zero-denominator and store-only/no-lookup cases).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d97d379b66bce22a3a2146a7c96535aa0ccc5fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3179, opened 2026-05-01
[CB] Fix CB lookup correctness, thread safety, and store-complete race

deng451e:
## Summary

  - Fix 0% CB hit rate on re-requests caused by `cb_store_final` never registering chunks in the fingerprint table (omission since initial `blend_server_v2` commit)
  - Add thread safety to `BlendTokenRangeMatcher` (had a `TODO: Needs thread-safety` comment since inception)
  - Fix store-complete race in `_cb_store_gpu_copy` causing lookups to miss freshly-stored chunks
  - Add `hit_tokens` / `requested_tokens` metrics to `CB_LOOKUP_END` to feed per-request hit rate on root OTel spans

  ## Root Cause

  `cb_store_final` was introduced with the docstring *"The stored chunk should be accessible for normal mode LLMs"* — it stored
  data in the storage backend but never called `on_new_token_hashes`. The CB re-lookup path (`match_sub_sequence` against the
  fingerprint table) was simply never considered. Any re-request for a sequence that went through a final store got 0% CB hit
  rate silently.

  ## Changes

  **`cb_store_final`: register chunks in fingerprint table**
  Calls `on_new_token_hashes` after storing, mirroring `cb_store_pre_computed`. Without this, `match_sub_sequence` finds nothing
  on re-requests even when the data is in storage.

  **`BlendTokenRangeMatcher`: thread safety**
  Added `threading.Lock` to `on_new_token_hashes`, `match_sub_sequence`, and `remove_chunks`. Numba computation
  (`chunk_hash_windows_numba`, `rolling_hash_windows_numba`) is moved outside the lock since it accesses no shared state.

  **`on_new_token_hashes`: duplicate compact-ID dedup**
  Now that both `cb_store_pre_computed` and `cb_store_final` register the same sequence, a second call would overwrite
  `_token_hash_to_compact_id[th]` and orphan the first compact ID permanently. Fixed by filtering already-registered hashes
  before allocating IDs.

  **`_cb_store_gpu_copy`: store-complete race**
  `event.record()` was enqueued before `launch_host_func(finish_write)` on the same CUDA stream. The client's
  `event.synchronize()` could unblock while data was not yet committed to the storage index, causing an immediately-following CB
  lookup to report a miss. Fixed by enqueuing `finish_write` before `event.record()`.

  **`CB_LOOKUP_END` metrics**
  Added `hit_tokens`, `requested_tokens`, and corrected `fingerprint_hits` (was the pre-storage candidate count; now reflects
  actual storage hits). These feed the per-request `hit_rate` attribute on root OTel spans added in #3175.

  ## Test Plan

  - [ ] All 52 tests in `tests/v1/multiprocess/test_blend_server_v2.py` pass (6 were failing before this PR)
  - [ ] `tests/v1/mp_observability/subscribers/tracing/test_cb_server.py` passes

### DIFF
--- a/docs/design/observability/request-event-span.md
+++ b/docs/design/observability/request-event-span.md
@@ -95,6 +95,50 @@ All three `CB_LOOKUP_END` emit sites (no-fingerprint-match, no-GPU-context,
 happy path) populate these fields, so `hit_rate` is always present on the
 `cb.request` span.
 
+A fourth attribute is also set on `"cb.request"` at `CB_LOOKUP_END` time:
+
+| Attribute | OTel type | Value |
+|-----------|-----------|-------|
+| `prefix_hits` | `int` | chunks found via the prefix probe (not fingerprint matching) |
+
+#### Prefix probe
+
+`cb_lookup_pre_computed` has two lookup paths:
+
+1. **Fingerprint path** — `BlendTokenRangeMatcher.match_sub_sequence` finds
+   sub-sequence matches using polynomial rolling hashes.  Covers arbitrary
+   (non-prefix) positions in the token sequence.
+2. **Prefix probe** — a fallback that runs after the fingerprint path and fills
+   in chunks at contiguous prefix positions not already covered by fingerprint
+   results.  It calls `token_hasher.compute_chunk_hashes(token_ids)` to derive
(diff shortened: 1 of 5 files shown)


ApostaC (approved):
LGTM as we discussed offline.

sammshen (approved):
LGTM!

Issue #3187, opened 2026-05-03
[Bug] LazyMemoryAllocator exhausts under sustained load with no recovery (17k+ "Failed to allocate" warnings/4h on a single H200 server)


ccgibson:
## Your current environment

- **LMCache**: 0.4.5 (NeuralWatt internal build with the multi-group MLA fix from PR #2951; functionally equivalent to LMCache `dev` HEAD with the LazyMemoryAllocator code from #2407 + #2413 + #2449 + #2463 + #2767 + #3068 all included — verified by git-ancestry against the merge commits)
- **vLLM**: 0.19.0, image digest `sha256:12b3d2089d9d…` (`vllm/vllm-openai:latest-cu130`)
- **Model**: `zai-org/GLM-5.1-FP8` (MLA + MoE + DSA)
- **Hardware**: 8× NVIDIA H200 (143 GB each), bare metal, 2 TB host RAM
- **Single-process LMCache** (not MP mode), `LMCacheConnectorV1`
- **Configuration** (`/opt/lmcache/config.yaml`):
  ```yaml
  chunk_size: 256
  local_cpu: true
  max_local_cpu_size: 500.0
  local_disk: /mnt/lmcache
  max_local_disk_size: 13000.0
  save_decode_cache: false
  enable_lazy_memory_allocator: true
  lazy_memory_initial_ratio: 1.0
  lazy_memory_expand_trigger_ratio: 0.5
  lazy_memory_step_ratio: 0.1
  use_gpu_connector_v3: true
  enable_kv_events: true
  ```

## 🐛 Describe the bug

After ~7.5 hours of sustained production traffic, `LazyMemoryAllocator` (via `PinnedTensorMemoryAllocator.allocate()`) starts failing every chunk-store request with:

```
LMCache WARNING: Failed to allocate memory block of size 15734784 because no memory is available
  (memory_management.py:1076:lmcache.v1.memory_management)
```

The size is always exactly 15 734 784 bytes (15 MiB) — one MLA-multi-group chunk for one TP=8 worker (`align_bytes` for `KV_MLA_FMT` shapes `[1, 78, 256, 132]` + `[1, 78, 256, 656]`, chunk_size=256).

Once the allocator enters this state, it does **not recover**. Every subsequent chunk store fails for the rest of the container's lifetime, until vLLM is restarted.

### Quantitative scope

Single H200 server, last 4 hours of production traffic, vLLM uptime 16 h:

| metric | value |
|---|---|
[...]

PR #3189, opened 2026-05-04
[Bugfix] Fix LocalCPUBackend recovery when pinned CPU chunks block eviction

DongDongJu:
This PR fixes CPU cache allocation recovery https://github.com/LMCache/LMCache/issues/3187 when cached chunks are pinned or still in use.

Before this fix, batched CPU eviction could remove a layer group even when one layer was pinned. 
That could make allocation recovery behave incorrectly under memory pressure.

This PR makes batched eviction skip pinned or still-used layer groups. 
It also makes MixedMemoryAllocator-owned tensor objects free through the MixedMemoryAllocator wrapper.

  **If applicable**:


lmcache/v1/memory_management.py
```diff
@@ -2153,9 +2156,14 @@ def batched_allocate(
...
         ]:
             with self.host_mem_lock:
-                return self.pin_allocator.batched_allocate(
+                objs = self.pin_allocator.batched_allocate(
                     shapes, dtypes, batch_size, fmt, str(self)
                 )
+                if objs is not None:
+                    for obj in objs:
+                        if isinstance(obj, TensorMemoryObj):
+                            obj.parent_allocator = self
```
ApostaC, line 2275: Could you quickly explain why we need to do this?

ApostaC (approved):
LGTM. Just one clarification question.

lmcache/v1/memory_management.py
DongDongJu: The exact difference is the automatic free path.

Without this change, the object returned by `MixedMemoryAllocator.allocate()` still has `pin_allocator` as its parent, so `ref_count_down()` frees it directly through `pin_allocator.free()`.

With this change, the parent becomes `MixedMemoryAllocator`, so `ref_count_down()` enters `MixedMemoryAllocator.free()`, and the wrapper can route the free to the correct inner allocator. 
This keeps the allocation and free ownership symmetric.

Let me know If I understand the wrong about the mm semantic.

DongDongJu: like this 

Allocation path before this PR
------------------------------

Caller
  |
  v
MixedMemoryAllocator.allocate()
  |
  v
pin_allocator.allocate()
  |
  v
TensorMemoryObj(parent_allocator = pin_allocator)
  |
  v
Returned to caller

Free path before this PR
------------------------

TensorMemoryObj.ref_count_down()
  |
  v
obj.parent_allocator.free(obj)
  |
  v
pin_allocator.free(obj)

MixedMemoryAllocator is skipped.

-----

Allocation path after this PR
-----------------------------

Caller
  |
  v
MixedMemoryAllocator.allocate()
  |
  v
pin_allocator.allocate()
  |
  v
TensorMemoryObj(parent_allocator = pin_allocator)
  |
  v
MixedMemoryAllocator rewrites:
obj.parent_allocator = MixedMemoryAllocator
  |
  v
Returned to caller

Free path after this PR
-----------------------

TensorMemoryObj.ref_count_down()
  |
  v
obj.parent_allocator.free(obj)
  |
  v
MixedMemoryAllocator.free(obj)
  |
  v
pin_allocator.free(obj)

DongDongJu:
@sammshen @YaoJiayi Could you take a look? it can not solve whole thing in https://github.com/LMCache/LMCache/issues/3187, but we can partially solve it as a remedy for some cases.
Current problem is existing allocator architecture can not handle exhaustion case in https://github.com/LMCache/LMCache/issues/3187. I will introduce solution for this in next PR.

koopatroopa787:
Hey @DongDongJu, thanks for the ping! I took a look at #3189, the fix makes sense. Skipping layer groups where any layer is still pinned during batched eviction is the right call; letting eviction blow through pinned chunks is exactly the kind of thing that would wedge the allocator under sustained load.

One thing I noticed that's loosely related to my PR, on my side I'm resetting all pin counts to 0 on metadata reload after a restart (since no requests are in-flight at startup). Your fix and mine touch pin count logic from different angles, but I don't see an interaction issue between them since mine only applies at cold start.

I saw in your comment that this is a partial fix and the full exhaustion case needs a follow-up PR. Curious what direction you're thinking for that, are you planning to introduce a way to forcibly reclaim memory from long-pinned chunks, or more of an architectural change to the allocator itself?

Overall LGTM on the fix, the regression tests cover the pinned-block eviction case well.

ApostaC:
@YaoJiayi Can you help take a look at this PR? It's been here for a while.

DongDongJu:
> Hey @DongDongJu, thanks for the ping! I took a look at #3189, the fix makes sense. Skipping layer groups where any layer is still pinned during batched eviction is the right call; letting eviction blow through pinned chunks is exactly the kind of thing that would wedge the allocator under sustained load.
> 
> One thing I noticed that's loosely related to my PR, on my side I'm resetting all pin counts to 0 on metadata reload after a restart (since no requests are in-flight at startup). Your fix and mine touch pin count logic from different angles, but I don't see an interaction issue between them since mine only applies at cold start.
> 
> I saw in your comment that this is a partial fix and the full exhaustion case needs a follow-up PR. Curious what direction you're thinking for that, are you planning to introduce a way to forcibly reclaim memory from long-pinned chunks, or more of an architectural change to the allocator itself?
> 
> Overall LGTM on the fix, the regression tests cover the pinned-block eviction case well.

Hello @koopatroopa787, Sorry I was meant this pr https://github.com/LMCache/LMCache/pull/2734.

YaoJiayi (approved):

PR #3191, opened 2026-05-04
[Fix] Change pinned pointer allocations in non-CUDA equivalents to be page-aligned for O_DIRECT

zhengfeihe:
  
IO uring tests will fail on a non-cuda environment, while the test limit doesn't ban it.
 
I ran the io_uring tests on a fresh GCP instance and forgot to install the CUDA C++ extension, which caused the tests to fail. Looking into  it, the fallback `alloc_pinned_numa_ptr` for non-CUDA callers returns  a cache-line-aligned pointer instead of a page-aligned one, so this PR adjusts it to return page-aligned pointers.                                                                                           
                                                                                                                                           
 A second issue I hit was in `test_rust_raw_block_backend_close_with_inflight`,  where the test sometimes hangs forever. It looked like a deadlock at first, but gdb showed that pytest's `assert` uses `difflib` to render the diff on failure, and computing a diff between two 64KiB objects  takes forever. I changed those checks to raise exceptions directly so pytest skips the diff step. 
  
  similar issue in pytest been raised before : https://github.com/pytest-dev/pytest/issues/8998
              

**If applicable**:

### DIFF
--- a/lmcache/python_ops_fallback.py
+++ b/lmcache/python_ops_fallback.py
@@ -10,6 +10,7 @@ from multiprocessing import shared_memory
 from typing import Optional, Tuple
 import ctypes
 import ctypes.util
+import os
 
 # Third Party
 from numba import njit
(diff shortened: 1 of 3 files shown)


lmcache/python_ops_fallback.py
```diff
@@ -301,25 +302,49 @@ def __init__(self) -> None:
...
+    Allocate a pinned CPU buffer whose first usable byte is page-aligned,
+    and return a torch view of ``size`` bytes plus its base pointer.
 
-    # Get a pointer to the start of the tensor object as this is what is
-    # returned by the CUDA equivalent function
-    ptr = tensor.data_ptr()
+    Internally over-allocates one extra page on a backing tensor, then
+    slices the aligned region out. The slice shares storage with the
+    backing tensor, so keeping the slice alive keeps the underlying
+    memory alive (no need to track the backing tensor separately).
```
ApostaC, line 345: Could you explain why we need to allocate one more page?

ApostaC (commented):
Thanks for the catch! Please see my question below

lmcache/python_ops_fallback.py
zhengfeihe: Hi! We need to guarantee the pointer is page-aligned.                                                                                                                                                              
                                                                                                                                                                                                                     
  Detailed explanation:                                                                                                                                                                                              
                                                                                                                                                                                                                     
  torch.empty ends up calling posix_memalign with an alignment constant of 64, so the returned address is only cache-line aligned — not page-aligned.                                       
   
  Code references:                                                                                                                                                                                                   
  - Linux allocator path: https://github.com/pytorch/pytorch/blob/902b0400bb778167bec3e8a402f92943cbb91f8d/c10/core/impl/alloc_cpu.cpp#L126
  - Alignment constant (gAlignment = 64): https://github.com/pytorch/pytorch/blob/3fa3dc7e3dd54e6e1a8c5a2bd647796e34a086f8/c10/core/alignment.h#L15                                                                  
                                                                                                                                                   
  cudaHostAlloc returns page-aligned memory because the OS locks the buffer at a 4 KiB physical-page boundary. The non-CUDA fallback needs to mirror that guarantee, since downstream consumers (the   
  Rust raw-block backend with O_DIRECT) require page-aligned pointers.                                                                                                                                               
                                                                                                                                                                                                                                       
  We don't know up front how far the pytorch-returned base address sits from the next page boundary, as offset can be anywhere in [0, PAGE_SIZE - 1]. By allocating size + PAGE_SIZE bytes, we're guaranteed that  
  somewhere inside the over-allocated region there is a page-aligned address with size bytes of usable memory after it.

zhengfeihe:
@ApostaC 

Thank you for helping review this.~                                                                                                                                                                                
                                    
Also, I was probably smoking when I made the PR, so there was a misleading part in the original PR title and comment: page-aligned allocation is only needed when O_DIRECT is enabled in the storage backend, since under O_DIRECT disk I/O is strictly page-based. io_uring itself doesn't care — the reason the failing tests are all io_uring ones is that O_DIRECT is enabled by default for io_uring in our tests. I've already fixed them by force-pushing the branch.

lmcache/python_ops_fallback.py
ApostaC: Got it, this is clear!

ApostaC (approved):
LGTM! Please fix the merge conflicts

ApostaC:
@YaoJiayi Can you take a quick review on this? Thanks!

zhengfeihe:
> LGTM! Please fix the merge conflicts

@ApostaC  Thank you~. Fixed the merge conflicts

YaoJiayi (approved):
lgtm!

zhengfeihe:
@ApostaC @YaoJiayi 

Hi! I've rebased the branch onto the latest master. Could you take another look and merge it when you have a moment? Thanks!

sammshen:
@hlin99 would you be able to give this a quick test before we merge it :)

PR #3194, opened 2026-05-04
[CI] Verify full MP observability surface in long_doc_qa_l2

Oasis-Git:
Reuses the existing L2 long_doc_qa step instead of paying for a second model load. Two changes to that script:

1. Bump --metrics-sample-rate to 1.0 on the L2 relaunch so the histograms record on every event. The default 0.01 would leave them empty in this short workload and flake the assertions.

2. After the existing L2 data-flow checks, add a "Step 5" block that asserts every metric we publish from MP mode actually advances:
   - newer counters with label dimensions advance > 0 and carry the expected label (l2_store_completed/l2_load_completed by l2_name, lookup_requested_tokens/lookup_hit_tokens by model_name, num_chunks_loaded by worker_id)
   - the four throughput histograms record at least one observation (lmcache_mp_l0_l1_*_throughput_gbs and lmcache_mp_l2_*_throughput_gbs)

The label-presence check catches the case where a counter fires but the attribute plumbing broke — e.g. a future refactor that drops the attribute at emit time but still ticks the counter.

**If applicable**:

### DIFF
--- a/.buildkite/k3_tests/multiprocess/scripts/run-long-doc-qa-l2.sh
+++ b/.buildkite/k3_tests/multiprocess/scripts/run-long-doc-qa-l2.sh
@@ -484,17 +484,34 @@ def counter_total(name: str) -> float:
     return total
 
 
-def histogram_count(name: str) -> float:
+def histogram_count(base_name: str) -> float:
     """Sum the ``_count`` series across all label combinations.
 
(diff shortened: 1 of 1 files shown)


sammshen (approved):
LGTM!

ApostaC:
image

ApostaC (approved):
LGTM!

PR #3195, opened 2026-05-05
[CLI[[Bench] Add prefix-suffix-tuner workload for tiered + Blending KV-cache

sammshen:
Adds a sequential two-pass workload that exercises the L0/L1/L2 tiered KV cache hierarchy with a single benchmark configurable via context-length, prefix-ratio, and a thrash multiplier. Each request has the layout [prefix][random breaker][shared suffix]; pass 1 populates the cache and pass 2 (measured) repeats the same prefix order, so a 1.05x overflow of the targeted tier suffices to drive every measured request to the next tier down. Designed to be run unchanged across vanilla vLLM (L0 only), vLLM + LMCache L1+L2, and vLLM + LMCache + CacheBlend baselines.

Includes factory wiring, CLI args, interactive-mode schema entries, 33 new unit tests, factory-test coverage, and documentation updates across the design doc and user-facing rST references.

**If applicable**:

### DIFF
--- a/docs/design/cli/commands/bench/engine_bench/bench-engine.md
+++ b/docs/design/cli/commands/bench/engine_bench/bench-engine.md
@@ -1,14 +1,21 @@
 # `lmcache bench engine` — Design & Extension Guide
 
-**Status:** Implemented  |  **Date:** 2026-03-27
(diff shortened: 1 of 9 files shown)


ApostaC (approved):
LGTM!

PR #3199, opened 2026-05-05
[Docs] Add Recipes section with MiniMax-M2 as first entry

sammshen:
Introduce a Recipes section under docs/source/recipes/ that catalogs model architectures validated end-to-end with LMCache. Each recipe page uses synced engine tabs (vLLM / SGLang / TRT-LLM) so the format scales as more architectures are validated. MiniMax-M2 is the first entry.

**If applicable**:

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -85,6 +85,14 @@ Documentation
 
 .. toctree::
    :maxdepth: 2
+   :caption: Recipes
+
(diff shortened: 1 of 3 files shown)


Shaoting-Feng (approved):

ApostaC (approved):
LGTM! Good starting point!

PR #3203, opened 2026-05-05
[CI/CD] Add CI-safe raw-block temp-file tests

DongDongJu:

Adds focused CI coverage for the raw-block backend without requiring a real raw block device, 
NVMe device, loop device, sudo, or root-only setup.

The new tests use pytest temp files as truncated regular files and exercise:
  - the Rust `RawBlockDevice` extension with POSIX I/O
  - shared `RawBlockCore` store/load/delete/duplicate/recovery behavior
  - the MP `RawBlockL2Adapter` store/lookup/load/delete path

This also adds a dedicated raw-block GitHub Actions job that builds the Rust extension with maturin 
and runs only the raw-block tests on Ubuntu/Python 3.11. 
The job is path-filtered so Rust raw-block changes and raw-block MP/test changes trigger it without adding Rust builds to the full Python matrix.

Default PR CI uses `use_odirect=False` and `io_engine="posix"` against truncated regular files. O_DIRECT coverage is present only as an opt-in smoke test and is skipped by default. The required pytest temp directory is rooted under the repo workspace with `--basetemp=.pytest-tmp/raw-block`, so the workflow does not depend on `/tmp` access.

  `io_uring` has a best-effort Linux smoke test that skips on unsupported-kernel/container errors.

  **If applicable**:


DongDongJu:
Hello @ApostaC @sammshen , I made raw block test by git (https://github.com/LMCache/LMCache/actions/runs/25403110100/job/74507477198?pr=3203).
Do you guys think that we should make it via k3? WDYT?

ApostaC (approved):
LGTM!

DongDongJu:
@sammshen @deng451e @hickeyma Could you take a look by any chance? its raw block ci

DongDongJu:
@sammshen Could you check this once?

DongDongJu:
@KuntaiDu @sammshen Could you take look this by any chance?

DongDongJu:
Hello @sammshen, Could you check this one?

KuntaiDu (approved):
LGTM --- let's see if CI passes.

PR #3210, opened 2026-05-06
[Feature][RawBlock] Add multi-device sharding to the MP L2 adapter

2xdevv:

The multiprocessing raw-block L2 adapter currently accepts only one device path, preventing a single adapter from distributing local-rank KV objects across multiple raw devices. This PR adds deterministic single-node device sharding while preserving request ordering and raw-block recovery behavior.

- Replace `device_path` with `device_paths`, accepting a single path, a list, or a comma-separated string.
- Validate that configured device paths are non-empty and unique.
- Create one `RawBlockCore` per device and route objects using `kv_rank.local_rank % len(device_paths)`.
- Fan out store, lookup, and load operations across devices while preserving input-order results and FDP placement IDs.
- Size the shared shard pool using `len(device_paths) * num_shard_workers_per_device`, with a default multiplier of `4`.
- Apply unlock and delete operations to each key's owning device.
- Aggregate capacity, health, recovery, status, and lifecycle handling across all cores.
- Add configuration, routing, recovery, ordering, and deletion-isolation tests.
- Document multi-device configuration, routing, worker sizing, and restart requirements.

- This changes the raw-block configuration key from `device_path` to `device_paths`; existing configurations must be updated.
- Device-path order must remain stable across restarts because local ranks map deterministically to positions in `device_paths`.
- `capacity_bytes` is applied independently to each configured device.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
@@ -11,7 +11,8 @@ lookup, and load.
 from __future__ import annotations
 
 # Standard
-from concurrent.futures import Future, ThreadPoolExecutor
+from collections.abc import Callable, Sequence
+from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from functools import partial
 from typing import TYPE_CHECKING, Any, Optional, cast
 import threading
@@ -56,6 +57,7 @@ RawBlockStoreTaskResult = tuple[
     list[ObjectKey],
     list[int],
 ]
+RawBlockDeviceSelector = Callable[[ObjectKey, int], int]
 
 _FDP_DATA_PLACEMENT_POLICY_NONE = "none"
 _FDP_DATA_PLACEMENT_POLICY_CACHE_SALT_PREFIX = "cache_salt_prefix"
@@ -210,13 +212,40 @@ def _make_bitmap(size: int) -> "Bitmap":
     return Bitmap(size)
 
 
+def _select_device_by_local_rank(key: ObjectKey, device_count: int) -> int:
+    """Select a raw-block device by the key's packed local rank."""
+    return _local_rank_from_kv_rank(key.kv_rank) % device_count
+
+
+def _group_key_indices_by_device(
+    keys: Sequence[ObjectKey],
+    cores: Sequence[RawBlockCore],
+    select_device: RawBlockDeviceSelector,
+) -> dict[RawBlockCore, list[int]]:
+    """Group submitted key indices by selected raw-block core."""
+    device_count = len(cores)
+    if device_count <= 0:
+        raise ValueError("raw_block requires at least one device")
+
+    groups: dict[RawBlockCore, list[int]] = {}
+    for i, key in enumerate(keys):
+        device_index = select_device(key, device_count)
+        if device_index < 0 or device_index >= device_count:
+            raise ValueError(
+                "device selector returned out-of-range index "
+                f"{device_index} for {device_count} raw-block devices"
+            )
+        groups.setdefault(cores[device_index], []).append(i)
+    return groups
+
+
 class RawBlockL2AdapterConfig(L2AdapterConfigBase):
     """Configuration object for the built-in raw-block MP L2 adapter."""
 
     def __init__(
         self,
         *,
-        device_path: str,
+        device_paths: Sequence[str] | str,
         slot_bytes: int,
         capacity_bytes: int = 0,
         use_odirect: bool = True,
@@ -243,11 +272,15 @@ class RawBlockL2AdapterConfig(L2AdapterConfigBase):
         num_store_workers: int = 2,
         num_lookup_workers: int = 1,
         num_load_workers: int = 4,
-    ):
+        num_shard_workers_per_device: int = 4,
+    ) -> None:
         """Initialize raw-block MP adapter configuration.
 
         Args:
-            device_path: Raw device path or pre-sized file path used for L2.
+            device_paths: Raw device path(s) or pre-sized file path(s). A
+                single string selects one device; a list of strings or a
+                comma-separated string enables multi-device sharding. Objects
+                are routed by ``kv_rank.local_rank % len(device_paths)``.
             slot_bytes: Fixed data-slot size in bytes.
             capacity_bytes: Optional cap on usable bytes; zero uses device size.
             use_odirect: Whether to open the raw path with O_DIRECT.
@@ -286,9 +319,12 @@ class RawBlockL2AdapterConfig(L2AdapterConfigBase):
             num_store_workers: Number of store worker threads.
             num_lookup_workers: Number of lookup worker threads.
             num_load_workers: Number of load worker threads.
+            num_shard_workers_per_device: Worker-count multiplier per configured
+                device for the shared shard operation pool.
         """
         super().__init__()
-        self.device_path = device_path
+        paths = self._normalize_device_paths(device_paths)
+        self.device_paths = tuple(paths)
         self.slot_bytes = int(slot_bytes)
         self.capacity_bytes = int(capacity_bytes)
         self.use_odirect = bool(use_odirect)
@@ -349,19 +385,19 @@ class RawBlockL2AdapterConfig(L2AdapterConfigBase):
         self.num_store_workers = int(num_store_workers)
         self.num_lookup_workers = int(num_lookup_workers)
         self.num_load_workers = int(num_load_workers)
+        self.num_shard_workers_per_device = int(num_shard_workers_per_device)
 
     @classmethod
     def from_dict(cls, d: dict) -> "RawBlockL2AdapterConfig":
         """Build and validate a raw-block config from ``--l2-adapter`` JSON."""
-        device_path = d.get("device_path")
-        if not isinstance(device_path, str) or not device_path:
-            raise ValueError("device_path must be a non-empty string")
         if "per_tp_device_paths" in d:
             raise ValueError(
                 "per_tp_device_paths is not supported in MP raw_block mode"
             )
         if not bool(d.get("persist_enabled", True)):
             raise ValueError("raw_block requires persist_enabled=true")
+
+        device_paths = cls._parse_device_paths(d)
 
         slot_bytes = d.get("slot_bytes")
         if not isinstance(slot_bytes, int) or slot_bytes <= 0:
@@ -425,6 +461,7 @@ class RawBlockL2AdapterConfig(L2AdapterConfigBase):
             "num_store_workers": 2,
             "num_lookup_workers": 1,
             "num_load_workers": 4,
+            "num_shard_workers_per_device": 4,
         }
         worker_counts: dict[str, int] = {}
         for field_name, default in worker_defaults.items():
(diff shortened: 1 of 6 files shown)


ApostaC:
@DongDongJu Would you like to take a look? Thanks!

DongDongJu:
> @DongDongJu Would you like to take a look? Thanks!

Could you add me into codeowners for this folder?

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
```diff
@@ -62,13 +64,46 @@ def _make_bitmap(size: int) -> "Bitmap":
...
+
+
 class RawBlockL2AdapterConfig(L2AdapterConfigBase):
     """Configuration object for the built-in raw-block MP L2 adapter."""
 
     def __init__(
         self,
         *,
-        device_path: str,
+        device_path: str | None = None,
```
DongDongJu: Can we avoid non-optional canonical representation if possible?

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
```diff
@@ -694,15 +834,26 @@ def _run_load_task(
...
-                accessed_keys.append(keys[i])
+        for core, indices in self._group_key_indices(keys).items():
+            results = core.load_many_into(
+                [specs[i].encoded for i in indices],
+                [objects[i] for i in indices],
+            )
+            for local_idx, ok in zip(indices, results, strict=True):
+                if ok:
+                    bitmap.set(local_idx)
+                    accessed_keys.append(keys[local_idx])
```
DongDongJu: Please fix the notification order issue.
listener notification from `_finish_load_task` receives the reordered accessed_keys.

Lets assume that 
- nvme devices: 0, 1, 2
- request order:
keys = [k_dev2_a, k_dev0, k_dev2_b]

- current grouping order:
group dev2 -> [k_dev2_a, k_dev2_b]
group dev0 -> [k_dev0]

- current accessed_keys:
[k_dev2_a, k_dev2_b, k_dev0]

- expected accessed_keys:
[k_dev2_a, k_dev0, k_dev2_b]

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
```diff
@@ -138,16 +180,18 @@ def __init__(
...
         if "per_tp_device_paths" in d:
             raise ValueError(
                 "per_tp_device_paths is not supported in MP raw_block mode"
             )
         if not bool(d.get("persist_enabled", True)):
             raise ValueError("raw_block requires persist_enabled=true")
 
+        multi_device_paths = cls._parse_multi_device_paths(d)
+        device_count = d.get("device_count")
+        if device_count is not None and int(device_count) != len(multi_device_paths):
```
DongDongJu: `device_count` validation is too permissive. Could you add exact reason per condition?

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
```diff
@@ -138,16 +180,18 @@ def __init__(
...
         """Build and validate a raw-block config from ``--l2-adapter`` JSON."""
-        device_path = d.get("device_path")
-        if not isinstance(device_path, str) or not device_path:
-            raise ValueError("device_path must be a non-empty string")
         if "per_tp_device_paths" in d:
             raise ValueError(
                 "per_tp_device_paths is not supported in MP raw_block mode"
             )
         if not bool(d.get("persist_enabled", True)):
             raise ValueError("raw_block requires persist_enabled=true")
```
DongDongJu, line 399: Also, if both `device_path` and `multi_device_paths` is set then what happen?

DongDongJu (changes requested):
Hello @2xdevv, Thanks for your contribution!
I left few feedbacks.
Also, Currently, multi-device operations are sharded but still serial within a task.
Can we add per-worker support for this sharding?

2xdevv:
thank you for the feedback!

I've added worker pool sharding and fixed notification ordering issue.

New worker pool is added because using already existing one would have a problem with completion tracking, requiring some reconstructuring.

For `multi_device_paths`, I referenced the raw_block_backend, which has both `device_path` and `per_tp_device_paths`, so leaving original `device_path` would be fine.

I first thought of changing the argument to `device_paths`, which accepts both single and multiple device cases, but changed to leave `device_path`. Which one do you think is the better way?

Also, `device_path` is is ignored if `per_tp_device_paths` is a proper one and I don't think device count is really needed.

DongDongJu:
> thank you for the feedback!
> 
> I've added worker pool sharding and fixed notification ordering issue.
> 
> New worker pool is added because using already existing one would have a problem with completion tracking, requiring some reconstructuring.
> 
> For `multi_device_paths`, I referenced the raw_block_backend, which has both `device_path` and `per_tp_device_paths`, so leaving original `device_path` would be fine.
> 
> I first thought of changing the argument to `device_paths`, which accepts both single and multiple device cases, but changed to leave `device_path`. Which one do you think is the better way?
> 
> Also, `device_path` is is ignored if `per_tp_device_paths` is a proper one and I don't think device count is really needed.

Hello @2xdevv, lets remove all `device_path` related and keep one canonical `device_paths` to take care of all of them.
And, please modify the docs as follows.
I am not preferring to diverge behavior based on vague params.

2xdevv:
> Hello @2xdevv, lets remove all `device_path` related and keep one canonical `device_paths` to take care of all of them. And, please modify the docs as follows. I am not preferring to diverge behavior based on vague params.

Hello, @DongDongJu
I've update everything requested, including the docs. Thanks!

DongDongJu (changes requested):
Please update the design doc, not only `docs/source/mp/l2_storage.rst`.
Otherwise, LGTM!

DongDongJu (approved):
Hello @2xdevv, LGTM!
Could you leave the performance improve number for this?
I can expect close linear improvement before ceiling by PCIe.
If not we can work on linux kernel side too.

2xdevv:
Ran the requested scaling numbers on real hardware: two `SAMSUNG_PM1753` NVMe drives (PCIe Gen4 x4 each), driving the actual `RawBlockL2Adapter` multi-device sharding path (`kv_rank.local_rank % len(device_paths)` routing), io_uring + O_DIRECT, 64MB chunks, ~10.7GB/device per run.

| | 1 device | 2 devices (sharded) | Scaling factor | Per-device efficiency retained |
|---|---|---|---|---|
| Write | 5.64 GB/s | 10.00 GB/s (5.00/device) | **1.77x** | 89% |
| Read | 4.33 GB/s | 8.39 GB/s (4.19/device) | **1.94x** | 97% |

So it is close to linear, matching your expectation. Read is nearly perfect (94-97% efficiency across two independent shard-pool workers); write retains ~89%.

One caveat worth flagging: per-device throughput here (5.6 GB/s write / 4.3 GB/s read) sits below what the same PM1753 hardware does through the legacy single-device `RustRawBlockBackend` path (~7.0-7.1 GB/s on both), and below the raw PCIe Gen4 x4 ceiling (~7.9 GB/s theoretical). It seems like it's from the missing zero-copy fixed-buffer registration in the MP path, not a PCIe or kernel-side limit — the scaling itself is healthy, closing that per-device gap is a separate follow-up.

DongDongJu:
@zhengfeihe PTAL!

zhengfeihe:
Multi-device support for `raw_block` is worth pursuing, and I don't think we have existing work covering it. Before going into line-level review, I have one design concern: this PR makes the adapter responsible for managing multiple raw block devices, and I'm not sure that responsibility belongs inside the adapter.

We already manage adapters individually at the controller layer, including eviction watermarks, health, add/remove, drain, and status. So another option is one `raw_block` adapter per device, with a small routing policy to select the adapter for each key. `StorePolicy` / `PrefetchPolicy` already support per-key adapter targeting. This would give us per-device accounting, failure isolation, and drain without reimplementing them inside `raw_block`.

The current approach also has a few concrete issues:

1. `local_rank % N` makes existing data depend on the device list staying unchanged. If a device is added, removed, or replaced, the modulo changes and recovered keys may route to the wrong device. Since each device already rebuilds its on-disk index at startup, we could instead rebuild a key-to-device map and only use rank affinity for write placement.

2. FDP configuration currently uses placement IDs from `self._cores[0]` for every device. Placement IDs are namespace-specific, so this can break with heterogeneous devices. I think the simplest option for this PR is to reject `fdp_enabled` with multiple `device_paths`, then add per-device FDP discovery separately.

3. Capacity/usage is aggregated, but writes are pinned by rank. One device can become full while aggregate usage is still below the eviction watermark, so eviction does not run and writes to that device can fail.

I'm open to either direction. If we keep multi-device management inside the adapter, I think we need per-device usage, FDP handling, eviction, and eventually drain. Otherwise, one adapter per device plus a routing policy seems simpler and reuses more of the existing controller machinery.

zhengfeihe:
I realize that when this PR was opened back in early May, many of these concerns didn't exist yet. Since then, we've added the coordinator, per-`cache_salt` quotas, and FDP support to `raw_block`, so I think multi-device support now needs to account for these pieces as well. Otherwise, we risk introducing quite a few edge cases and subtle bugs.

Sorry for raising this so late and potentially blocking the PR. I should have reviewed it earlier, and I'll try to be quicker next time.

Thanks again for the contribution!

2xdevv:
Thanks for the detailed review, @zhengfeihe. I mostly agree with your points.

When this PR was originally written, keeping the multi-device handling inside `RawBlockL2Adapter` seemed like the simplest way to add device sharding without changing the controller path. But with the controller, per-adapter eviction/accounting, runtime add/remove/drain, quotas, and FDP support that have been added since then, one adapter per device looks like a better fit for the current architecture.

Using one `raw_block` adapter per device would avoid most of these special cases and reuse the existing adapter lifecycle and accounting machinery.

No worries about the timing — the surrounding architecture changed quite a bit while this PR was open. Thanks again for taking a careful look at it.

BTW, if you have some time, could you also take a look at [#3812](https://github.com/LMCache/LMCache/pull/3812)? It's an important piece for `raw_block` performance, since it batches `load_many` reads instead of issuing them one by one. I'd really appreciate your review on that one as well.

zhengfeihe:
> Thanks for the detailed review, @zhengfeihe. I mostly agree with your points.
> 
> When this PR was originally written, keeping the multi-device handling inside `RawBlockL2Adapter` seemed like the simplest way to add device sharding without changing the controller path. But with the controller, per-adapter eviction/accounting, runtime add/remove/drain, quotas, and FDP support that have been added since then, one adapter per device looks like a better fit for the current architecture.
> 
> Using one `raw_block` adapter per device would avoid most of these special cases and reuse the existing adapter lifecycle and accounting machinery.
> 
> No worries about the timing — the surrounding architecture changed quite a bit while this PR was open. Thanks again for taking a careful look at it.
> 
> BTW, if you have some time, could you also take a look at [#3812](https://github.com/LMCache/LMCache/pull/3812)? It's an important piece for `raw_block` performance, since it batches `load_many` reads instead of issuing them one by one. I'd really appreciate your review on that one as well.

Sure～ I assigned myself. Will take a look today or tomorrow

PR #3215, opened 2026-05-07
[Bench] Add per-request timeout

sammshen:

**If applicable**:


deng451e (approved):
LGTM

ApostaC:
Maybe let's don't fix the problem like this? A better way is to fix the PD problem

sammshen:
@ApostaC maybe the timeout behavior is good in general?

PR #3220, opened 2026-05-07
  [CLI] count requests as successful when stream has no content but usage reports tokens

deng451e:
Root cause

  _extract_content() only inspects delta.content / reasoning_content / reasoning. For gpt-oss-20b on
  vLLM with the harmony chat template + --tool-call-parser openai --enable-auto-tool-choice (esp.
  with --load-format dummy), the chat-completion stream emits only an opening content="" chunk and a
  final finish_reason chunk with empty delta — even though usage.completion_tokens > 0.
  first_token_time is never set, so every request is logged as successful=False, ttft=-1.0.

  Fix

  After the stream ends, if first_token_time == 0.0 but num_output_tokens > 0, set first_token_time =
   finish_time so the request is marked successful.

### DIFF
--- a/lmcache/cli/commands/bench/engine_bench/request_sender.py
+++ b/lmcache/cli/commands/bench/engine_bench/request_sender.py
@@ -136,6 +136,8 @@ class RequestSender:
                     tokens.append(content)
 
             finish_time = time.time()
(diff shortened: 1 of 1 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM

Issue #1330, opened 2025-08-13
Crash when RAM offload is full


panpan0000:
**Describe the bug**
When the RAM + GDS offload together, when RAM is full, whole vLLM crash instead of rotate gc.

lmcache docker image  v0.3.3

**To Reproduce**
```
       LMCACHE_GDS_PATH="/mnt/afs/cache/Qwen-2.5-0.5B"\
        LMCACHE_CUFILE_BUFFER_SIZE=500 \
        LMCACHE_LOCAL_CPU=True \
	LMCACHE_MAX_LOCAL_CPU_SIZE=1 \
        VLLM_ENABLE_V1_MULTIPROCESSING=1 \
        VLLM_LOGGING_LEVEL=DEBUG   LMCACHE_LOG_LEVEL=DEBUG  \
        vllm serve /models/Qwen-2.5-0.5B/ \
       --gpu-memory-utilization 0.05 \
        --port 9100 \
       --max_model_len 10000 \
        --disable-log-requests \
        --enforce-eager \
        --no-enable-prefix-caching \
        --kv-transfer-config \
        '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
```

```
python benchmark_serving.py --port 9100 --model "/models/Qwen-2.5-0.5B/"     --dataset-path /models/sonnets.txt  --dataset-name sonnet     --sonnet-input-len 9300     --num-prompts 10 --max-concurrency 10    --sonnet-output-len 5
```

**Expected behavior**

process should alive 

**Screenshots**
```
[2025-08-13 08:25:56,383] LMCache DEBUG: Failed to allocate memory for tensor(torch.Size([2, 24, 256, 128]), torch.bfloat16) because no memory is available (memory_management.py:672:lmcache.v1.memory_management)
[...]

PR #3221, opened 2026-05-07
[Feat]: HiddenStateStore for hidden-state caching

allytotheson:
# Hidden state store / retrieve with KV cache (LMCache + vLLM-Omni)

## Problem

When LMCache restores KV cache, the engine can skip full prefill for cached tokens. Downstream stages (e.g. thinker → talker) still need **full-prefix hidden states** for those tokens, not only KV. Without storing and restoring hidden states alongside KV, consumers see truncated or wrong activations.

## Solution (LMCache)

Add `**HiddenStateStore`**: a dedicated pinned-CPU pool and chunk cache keyed like KV (`TokenDatabase` → same `CacheEngineKey` per chunk).

- `**LMCacheEngine`** exposes `**engine.hidden_state_store**`: a `**HiddenStateStore**` instance when `enable_hidden_state_cache` is true, otherwise `**None**` (integrators must guard before calling HS APIs).

**Public API** (on `hidden_state_store`, not on the engine):

| Method                                                          | Role                                                                                                                         |
| --------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------- |
| `store_hidden_states(token_ids, hidden_states, *, layer_idx=0)` | Store one logical tensor per layer index; chunking matches KV.                                                               |
| `retrieve_hidden_states(token_ids, *, layer_idx=0)`             | Return the longest prefix-strict CPU tensor for that layer, or `None` if the first chunk is missing KV or HS for that layer. |

**LMCache YAML / `LMCacheEngineConfig`:**

- `enable_hidden_state_cache` — master switch for constructing the store.
- `max_hidden_state_cpu_size` — HS-only pinned pool size (GiB); must be positive when HS is enabled.
- `hidden_state_layers` — optional allowlist of storage `layer_idx` values accepted on store. Prefer leaving unset unless you align indices with the worker that calls store/retrieve.

**Eviction:** HS has its own LRU under pressure; KV eviction is observed lazily on retrieve (`storage_manager.contains(key)`). If KV for a chunk is gone, paired HS for that key is dropped.

## Solution (vLLM-Omni)

`**OmniGPUModelRunner`** (`vllm-omni/vllm_omni/worker/gpu_model_runner.py`) talks to LMCache through the KV connector’s `**lmcache_engine`**, then `**engine.hidden_state_store**`, after checking it is not `None`.

**Hybrid multimodal hook order**

- **Default:** multimodal output keys `("0", "24")` (Qwen-style thinker hooks), then the synthetic key `**hidden`** for the main text hidden-state tensor. List order maps to LMCache `**layer_idx`** 0, 1, 2, …
- **Forks / other models:** subclass `**OmniGPUModelRunner**` and set the class attribute `**_LMCACHE_OMNI_MM_HS_KEYS**` to the hook names you need. There is no `omni_kv_config` / YAML key for this yet; one could be added later without changing the baked-in default.

## Repository locations

| Area                                  | Path                                                                             |
| ------------------------------------- | -------------------------------------------------------------------------------- |
| Store implementation                  | `LMCache/lmcache/v1/hidden_state_store.py`                                       |
| Engine wiring                         | `LMCache/lmcache/v1/cache_engine.py`                                             |
| Config definitions                    | `LMCache/lmcache/v1/config.py`                                                   |
| Design doc                            | `LMCache/docs/design/v1/hidden_state_store.md`                                   |
| LMCache tests                         | `LMCache/tests/v1/test_hidden_state_store.py`                                    |

## Tests (LMCache)

`LMCache/tests/v1/test_hidden_state_store.py` covers chunk/KV alignment, round-trip store/retrieve, KV-driven truncation, HS-only eviction, multi-layer behavior, layer allowlist filtering, independent pools, and callers skipping HS when `hidden_state_store` is `None`.

## Attribution
This write-up started from **[LMCache #3145 — Add hidden state support](https://github.com/LMCache/LMCache/pull/3145)** ([LMCache/LMCache](https://github.com/LMCache/LMCache)). The original PR description and framing were used as a starting point, then revised and expanded here.
It is meant to be read together with the companion integration work in **[vLLM-Omni #1330 — Support KV Cache CPU Offloading](https://github.com/vllm-project/vllm-omni/pull/1330)** ([vllm-project/vllm-omni](https://github.com/vllm-project/vllm-omni)), which covers the Omni-side wiring that consumes LMCache’s KV (and related) paths alongside this stack.
**Links**
| PR | Repository | Title |
|----|------------|-------------------|
| [#3145](https://github.com/LMCache/LMCache/pull/3145) | `LMCache/LMCache` | Add hidden state support |
| [#1330](https://github.com/vllm-project/vllm-omni/pull/1330) | `vllm-project/vllm-omni` | Support KV Cache CPU Offloading |


lmcache/v1/hidden_state_store.py
```diff
@@ -0,0 +1,305 @@
...
+        if hidden_states.dim() != 2:
+            raise ValueError(
+                f"hidden_states must be 2-D [num_tokens, hidden_dim], "
+                f"got shape {tuple(hidden_states.shape)}"
+            )
+
+        n_toks = (
+            len(token_ids) if isinstance(token_ids, list) else int(token_ids.shape[0])
+        )
+        if hidden_states.shape[0] != n_toks:
```
Vivo50E: Would it make sense to add a `token_offset` parameter here, changing this check to `hidden_states.shape[0] == n_toks - token_offset`? In vllm-omni, `store_hidden_states` is called incrementally — each call only has the newly computed hidden states (sched tokens), not the full prefix history. We still need to pass the full token_ids for correct chunk keys, but without token_offset we'd have to zero-pad a [num_computed+sched, hidden_dim] tensor just for the already-cached rows.

Vivo50E:
Thanks for the change! This looks good to me overall. Have you end-to-end tested the path with any omni model?

lmcache/v1/hidden_state_store.py
allytotheson: That makes sense. Thanks for the suggestion!

allytotheson:
> Thanks for the change! This looks good to me overall. Have you end-to-end tested the path with any omni model?
Yes! Here is a summary.

**Model:** Qwen2.5-Omni-3B (thinker -> talker pipeline)  
**GPU:** NVIDIA H100 80GB
**Prefix caching:** disabled, LMCache connector is the only cache path  

**Prompt:** a 2,510-token numbered-word sequence shared across two requests. Two requests were sent back-to-back
with the same prefix but different tails. Run 1 stores KV+HS to CPU; run 2 restores them.

| Metric | Run 1 (cold) | Run 2 (cached) | Delta | Speedup |
|--------|-------------|----------------|-------|---------|
| Thinker latency | 443 ms | 305 ms | −138 ms | **31%** |
| Wall-clock | 0.555 s | 0.309 s | −0.246 s | **44%** |

Run 2 matched 2,304 tokens (9 × 256) from LMCache and only recomputed the
206-token tail.

ApostaC:
Ping @sammshen

Vivo50E:
The latest update LGTM — I've already adapted vllm-omni: `store_hidden_states` now passes `token_offset=num_computed`, so the shape mismatch is gone.

lmcache/v1/hidden_state_store.py
```diff
@@ -0,0 +1,346 @@
...
+                f"({n_toks})]"
+            )
+        expected_rows = n_toks - token_offset
+        if hidden_states.shape[0] != expected_rows:
+            raise ValueError(
+                f"hidden_states first dim ({hidden_states.shape[0]}) must equal "
+                f"len(token_ids) - token_offset ({expected_rows})"
+            )
+
+        hs_cpu = hidden_states.detach().to("cpu", dtype=torch.float32).contiguous()
```
sammshen: this mostly looks good. however, there are a couple of issues: 
1. can we confirm that we actuaally need to call `.contiguous()`
2. we are creating an extra copy on the CPU right now because of `.to("cpu)` in addition to the call to `.copy_()`. could we also remokve the `to("cpu")`, especially since we don't have to do any shape transformation on the hidden states

docs/design/v1/hidden_state_store.md
```diff
@@ -0,0 +1,102 @@
+# HiddenStateStore
+
+`HiddenStateStore` is an LMCache-internal store dedicated to caching per-token
```
sammshen, line 3: nit can we have here that "it exists to support any system where we may need to draw out intermediate activations from an inference forward pass (e.g. vLLM-Omni thinker -> talker) where ..."

docs/design/v1/hidden_state_store.md
```diff
@@ -0,0 +1,102 @@
...
+pipelines (e.g. vLLM-Omni thinker -> talker) where the talker needs the
+thinker's intermediate activations for the cached prefix.
+
+## Goals
+
+- Keep KV cache logic and hidden-state cache logic in **separate classes**.
+  `LMCacheEngine` owns one instance and exposes it via
+  `engine.hidden_state_store`, but does not embed hidden-state state in
+  itself.
+- Use a **separate pinned CPU memory pool** so that hidden-state allocations
```
sammshen, line 16: nit: mention the heterogeneous shape of KV and other intermediate activations

docs/design/v1/hidden_state_store.md
```diff
@@ -0,0 +1,102 @@
...
+  cannot fragment the KV pool.
+- **Reuse** existing chunking and key generation
+  (`ChunkedTokenDatabase.process_tokens` -> `CacheEngineKey`) so each hidden
+  chunk is keyed identically to its KV counterpart.
+- Implement the **coupled-but-asymmetric** eviction rule:
```
sammshen, line 26: this is great

deng451e (approved):

lmcache/v1/config.py
```diff
@@ -592,6 +592,43 @@
...
+        "env_converter": float,
+        "description": (
+            "Maximum size in GiB of pinned CPU memory for the hidden-state cache. "
+            "Each chunk-layer tensor is shaped [chunk_size, hidden_dim] float32. "
+            "If the budget is too small, hidden entries may be evicted before "
+            "their KV counterparts. Required to be > 0 when "
+            "enable_hidden_state_cache=True."
+        ),
+    },
+    "hidden_state_layers": {
```
sammshen, line 649: this is great

lmcache/v1/hidden_state_store.py
```diff
@@ -0,0 +1,346 @@
...
+                )
+                break
+            tensor = obj.tensor
+            if tensor is None:
+                obj.ref_count_down()
+                raise RuntimeError(
+                    "HiddenStateStore: allocator returned MemoryObj with no "
+                    "backing tensor"
+                )
+            tensor.copy_(hs_cpu[start - token_offset : end - token_offset])
```
sammshen: tied to comment above

lmcache/v1/hidden_state_store.py
```diff
@@ -0,0 +1,346 @@
...
+            # Without an SM we cannot distinguish "KV evicted" from "no KV";
+            # treat as present so the prefix walk doesn't truncate spuriously.
+            return True
+        return self._storage_manager.contains(key) is not None
+
+    def _alloc_chunk(self, n_tokens: int, hidden_dim: int) -> Optional[MemoryObj]:
+        shape = torch.Size([n_tokens, hidden_dim])
+        dtype = torch.float32
+        for _ in range(_HS_ALLOC_MAX_RETRIES):
+            obj = self._allocator.allocate(shape, dtype, MemoryFormat.EC_TD)
```
sammshen: nit, can we introduced a new fmt `MemoryFormat.HS_TD`. should be the same as encoder caching in shape in the memory allocator

lmcache/v1/hidden_state_store.py
```diff
@@ -0,0 +1,346 @@
...
+            obj = layer_map[layer_idx]
+            tensor = obj.tensor
+            if tensor is None:
+                raise RuntimeError(
+                    "HiddenStateStore: cached MemoryObj has no backing tensor"
+                )
+            out_rows.append(tensor)
+            # Update LRU on hit.
+            self._lru.pop(key, None)
+            self._lru[key] = None
```
sammshen, line 274: small nit: if you look at how the existing code updates LRU, we should "touch" the LRU in reverse order to ensure that suffixes are evicted before prefixes

lmcache/v1/hidden_state_store.py
```diff
@@ -0,0 +1,346 @@
...
+
+        if not out_rows:
+            return None
+        return torch.cat(out_rows, dim=0)
```
sammshen, line 278: nit: add comment here that `torch.cat` will create a new buffer. I think this is unavoidable right now. we could keep a stable buffer of a long size (the max model length).

sammshen: reduce allocations

sammshen:
oh last thing, could we add soem docs to LMCache side, preferrably under Non KV Caching here: https://docs.lmcache.ai/non_kv_cache/encoder_cache.html with a new "Hidden States" section

sammshen (approved):
LGTM!

sammshen:
fix DCO and merge conflicts please

PR #3223, opened 2026-05-07
[Core] KV cache store & retrieve using HTTP interface

KuntaiDu:

This PR implements KV cache store and retrieval using HTTP interface. 

**If applicable**:


YaoJiayi (commented):
IIUC, we are retrieving and storing the actual KV cache instead of some pointers over http? Will this be slow? Why are we not keeping the actual data (i.e., KV cache) inside LMCache and let users' customized operations submitted through http so that all data-heavy operations happen locally inside LMCache?

KuntaiDu:
> Will this be slow?

It will be slow and it should be optimized in the future. But for async KV cache token dropping (one of the major planned use case of this PR), it's fine if the KV cache transmission is slow, because we can do token dropping while the original request continue its decoding and thus HTTP-based KV cache transmission is not on critical path.

> Why are we not keeping the actual data (i.e., KV cache) inside LMCache and let users' customized operations submitted through http so that all data-heavy operations happen locally inside LMCache?

For me, this is an ideal implementation performance-wise. But:
* Technically
  * P2P backend is missing
  * LMCache's CPU-only support is missing (likely token dropping algorithms need to be run on CPU)
  * Exporting LMCache data from its internal memory to an external CPU buffer is missing (this is necessary --- machine learning developer likes to handle a contiguous KV cache instead of chunked KV cache).
 
* User-experience wise
  * Using LMCache to transfer data requires user to launch MP server first. This seems to be a bit heavy for machine learning developer.
  * User may want to inject some external KV cache that are not generated by LMCache. In this case, KV cache cannot live inside LMCache anyway (though ur right the transmission could be handled by LMCache).

@ApostaC @YaoJiayi I want to hear from you on whether you think it is feasible, or preferred, to handle all user-experience issues, while still using LMCache for underlying transfer.

KuntaiDu:
Based on discussion with @ApostaC  and @YaoJiayi , the best technical route is:
* First PoC: put inference engine and kv optimization on same node, transfer path: engine --> LMCache --> CPUconnector --> contiguous CPU buffer.
* After P2P is done --> we will be able to put inference engine and kv optimization on different nodes.

KuntaiDu:
Based on discussion with @ApostaC and @YaoJiayi , the best technical route is:

First PoC: put inference engine and kv optimization on same node, transfer path: engine --> LMCache --> CPUconnector --> contiguous CPU buffer.
After P2P is done --> we will be able to put inference engine and kv optimization on different nodes.

PR #3151, opened 2026-04-27
[MP]: Add batch operations and per-op worker pools to Mooncake L2 adapter  

fcczzz:
**What this PR does / why we need it**:

Implements batch operations (`do_batch_get/do_batch_set/do_batch_delete`) in the `MooncakeConnector`, replacing per-key RPC calls with single batched calls. This reduces RPC overhead and improves throughput for L2 cache lookups, retrievals, and stores.
                                                                                                                                                                                                       
Also adds per-operation worker pool configuration (`lookup_workers, retrieve_workers, store_workers`) to the MooncakeStore L2 adapter, allowing each operation type to use an independently-sized worker pool to avoid head-of-line blocking.              
 <details>                                                                                                                                                                                            
  <summary>Benchmark: per-operation vs shared worker pool (mixed pressure)</summary>                                                                                                                   

<details>                                                                                                                                                                                                       
<summary>Script</summary>                                                                                                                                                               
                                                                                                                                                                                                       
  ```python
#!/usr/bin/env python3
"""Benchmark Mooncake L2 adapter latency with mixed operation pressure.

Run from the LMCache repo root, or pass --repo-root.  The config can be a JSON
string or a JSON file path.  Example:

  python bench_mooncake_multiworkers.py \
    --config-json '{"type":"mooncake_store","local_hostname":"192.168.0.139",...}' \
    --num-keys 1024 \
    --batch-size 16 \
    --value-size-bytes 1048576 \
    --warmup 20 \
    --iters 200 \
    --stores-per-round 4
"""

from __future__ import annotations

import argparse
import json
import os
import select
import statistics
import sys
import time
from pathlib import Path
from typing import Any


def _add_repo_to_path(repo_root: str) -> None:
    if repo_root not in sys.path:
        sys.path.insert(0, repo_root)


def _load_config(config_json: str | None, config_file: str | None) -> dict[str, Any]:
    if config_file:
        with open(config_file, encoding="utf-8") as f:
            return json.load(f)
    if config_json:
        return json.loads(config_json)
    raise ValueError("pass either --config-json or --config-file")


def _wait_any_event(fds: dict[int, str], timeout_s: float) -> list[str]:
    poll = select.poll()
    for fd in fds:
        poll.register(fd, select.POLLIN)
    events = poll.poll(int(timeout_s * 1000))
    ready: list[str] = []
    for fd, _event in events:
        try:
            os.eventfd_read(fd)
        except BlockingIOError:
            pass
        ready.append(fds[fd])
    return ready


[...]

PR #3227, opened 2026-05-08
[MP]: Add per-operation dedicated worker pools to Mooncake L2 connector

fcczzz:

> This is the second part of the split from #3151 .
> That PR was too large, so it was broken into:
> - Part 1: #3172 , already merged
> - Part 2 (this PR): per-operation dedicated worker pools

Different cache operations (lookup, retrieve, store, delete) have different latency profiles and load characteristics. Previously, all operations shared a single worker pool, which meant a burst of store operations could starve latency-sensitive lookups. This PR introduces per-operation worker pools, allowing each operation type to have its own dedicated C++ worker threads.

Key changes:
- Adds a `WorkerPoolConfig` struct to `ConnectorBase` that maps lane keys (`"lookup"`, `"retrieve"`, `"store"`, `"delete"`) to dedicated worker counts
- Requests are routed to the appropriate lane's queue based on the operation type — if a lane is not configured, the request falls back to the shared pool
- Python config class exposes `per_op_workers` as an optional `dict[str, int]`, with full validation

Benchmark results (Mooncake RDMA, 1024 keys, 1 MiB values, 4 stores/round, read after store):

| Config | Lookup avg | Lookup p99 | Load avg | Load p99 | Store avg | Store p99 |
|---|---|---|---|---|---|---|
| shared 4 workers | 0.947ms | 16.832ms | 3.399ms | 41.165ms | 5.292ms | 51.390ms |
| per-op (1/1/2) | 0.266ms | 0.483ms | 2.590ms | 3.506ms | 3.216ms | 67.434ms |

With per-op pools, lookup p99 drops ~35x (16.8ms → 0.48ms) and load p99 drops ~12x (41.2ms → 3.5ms) because reads no longer compete with store operations for the same workers. 

- The lane registry (`lookup`/`retrieve`/`store`/`delete`) is defined statically in `ConnectorBase` and validated at construction time — invalid lane keys in `per_op_workers` will throw immediately
- Partial `per_op_workers` is valid: unmentioned lanes use the shared `num_workers` pool
- The lane mechanism lives in `ConnectorBase` and is backend-agnostic; this PR only wires it up for Mooncake. Other L2 backends (Redis, S3, etc.) can adopt it by accepting `WorkerPoolConfig` in their constructors and pybind
- Tests no longer require a real Mooncake installation — a fake native extension is installed via monkeypatch for factory unit tests

**If applicable**:


csrc/storage_backends/connector_base.h
```diff
@@ -423,7 +497,64 @@ class ConnectorBase : public IStorageConnector {
...
+          reverse[o] = key;
+        }
+      }
+    }
+    auto it = reverse.find(op);
+    if (it == reverse.end()) {
+      throw std::runtime_error("unknown Op type");
+    }
+    return it->second;
+  }
```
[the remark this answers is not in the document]
fcczzz: done

csrc/storage_backends/connector_base.h
```diff
@@ -547,6 +679,9 @@ class ConnectorBase : public IStorageConnector {
   std::queue<Completion> completions_;
 
   std::vector<std::thread> workers_;
+  // Populated only during start_workers() (single-threaded construction).
+  // All subsequent accesses are reads — no lock needed in enqueue_request.
+  std::unordered_map<std::string, WorkerLane> lanes_;
```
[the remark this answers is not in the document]
fcczzz: done

csrc/storage_backends/connector_base.h
```diff
@@ -423,7 +497,64 @@ class ConnectorBase : public IStorageConnector {
...
+  }
+
+  void start_lane_workers(WorkerLane& lane, int num_workers) {
+    lane.workers.reserve(static_cast<size_t>(num_workers));
+    WorkerLane* lane_ptr = &lane;
+    for (int i = 0; i < num_workers; i++) {
+      lane.workers.emplace_back([this, lane_ptr]() {
+        this->worker_loop_for_queue(lane_ptr->mu, lane_ptr->cv,
+                                    lane_ptr->requests);
+      });
```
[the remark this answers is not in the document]
fcczzz: done

csrc/storage_backends/connector_base.h
```diff
@@ -423,7 +501,66 @@ class ConnectorBase : public IStorageConnector {
     }
   }
 
-  void worker_loop() {
+  static const std::unordered_map<std::string, std::vector<Op>>&
+  lane_registry() {
+    static const std::unordered_map<std::string, std::vector<Op>> registry = {
```
maobaolong, line 506: Is there a way to avoid introduce this map, or you can construct this map dynamically rather than construct it static like this?

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
```diff
@@ -153,14 +236,20 @@ def _create_mooncake_store_l2_adapter(
...
-    )
+    native_client_kwargs = {
+        "config": config.setup_config,
+        "num_workers": config.num_workers,
+        "l1_registration": l1_registration,
+    }
+    if config.per_op_workers is not None:
+        native_client_kwargs["per_op_workers"] = config.per_op_workers
+
+    native_client = LMCacheMooncakeClient(**native_client_kwargs)
```
maobaolong: Why we use a dict here?

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
```diff
@@ -78,8 +121,41 @@ def from_dict(cls, d: dict) -> "MooncakeStoreL2AdapterConfig":
...
+            ValueError: If ``raw`` is not a positive integer.
+        """
+        # bool is a subclass of int, so isinstance(True, int) is True.
+        # Explicitly reject bool to prevent True/False from being accepted.
+        if isinstance(raw, bool) or not isinstance(raw, int) or raw <= 0:
+            raise ValueError("num_workers must be a positive integer")
+        return raw
+
+    @staticmethod
+    def _validate_per_op_workers(
```
maobaolong: I guess this method is not only use for mooncake?

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
fcczzz: Fixed.

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
fcczzz: Currently only mooncake's pybind passes per_op_workers through to the C++ side at the moment — fs and redis backends don't accept it, so the feature is effectively mooncake-only for now. If we add `per_op_workers` support to other backends in the future, we can move this method up to a shared config base.

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
```diff
@@ -153,14 +236,19 @@ def _create_mooncake_store_l2_adapter(
             l1_registration.base = l1_memory_desc.ptr
             l1_registration.size = l1_memory_desc.size
 
-    native_client = LMCacheMooncakeClient(
-        config=config.setup_config,
-        num_workers=config.num_workers,
-        l1_registration=l1_registration,
-    )
+    native_client_kwargs = {
```
chunxiaozheng: Could we directly pass parameters here?

fcczzz: Done.

csrc/storage_backends/mooncake/pybind.cpp
```diff
@@ -1,11 +1,38 @@
...
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 #include "../connector_pybind_utils.h"
 #include "connector.h"
 
 namespace py = pybind11;
 
+namespace {
+
+lmcache::connector::WorkerPoolConfig parse_per_op_workers(
```
maobaolong: Is this method exclusive to mooncake only?

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
```diff
@@ -34,41 +33,85 @@
...
+        Returns:
+            A validated ``MooncakeStoreL2AdapterConfig``.
+
+        Raises:
+            ValueError: If ``num_workers`` or ``per_op_workers`` values
+                are invalid.
+        """
+        num_workers = cast(int, d.get("num_workers", 4))  # validated in __init__
 
+        per_op_workers: dict[str, int] | None = None
```
maobaolong: Is it possible to put the logic related to `per_op_workers` and `num_workers` to the L2AdapterConfigBase class?

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
```diff
@@ -78,8 +121,41 @@ def from_dict(cls, d: dict) -> "MooncakeStoreL2AdapterConfig":
         return cls(
             setup_config=setup,
             num_workers=num_workers,
+            per_op_workers=per_op_workers,
         )
 
+    @staticmethod
+    def _validate_num_workers(raw: object) -> int:
```
maobaolong: And this can be placed to L2AdapterConfigBase too

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
maobaolong: ditto

maobaolong (commented):
@fcczzz Thanks for this PR, Left some comments.

csrc/storage_backends/connector_base.h
fcczzz: Based on our offline discussion, we looked into constructing this map dynamically. The most viable option we found is `magic_enum`, which could generate the mapping at compile time without a hand-rolled static map. However, introducing a third-party dependency just for this seems like overkill given the map is small and only used in a non-hot path. We think the current static approach is the pragmatic choice for now.

csrc/storage_backends/mooncake/pybind.cpp
fcczzz: Moved the parsing logic to L2AdapterConfigBase._parse_per_op_workers_from_dict, and the C++ side parse_per_op_workers in pybind is just a thin conversion helper now.

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
fcczzz: Done.

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
fcczzz: Done.

fcczzz:
@maobaolong Thanks for the review. Now all the validation and parsing logic is now in L2AdapterConfigBase as shared static methods. The mooncake adapter and pybind layer just call into them. PTAL.

lmcache/v1/distributed/l2_adapters/mooncake_store_l2_adapter.py
fcczzz: Done - Moved alongside the other validation methods.

maobaolong (approved):
@fcczzz Thanks for this big improvement! LGTM. @chunxiaozheng Would you like to take another look? Thanks!

chunxiaozheng (approved):
Thanks for your contribution, LGTM!

PR #3217, opened 2026-05-07
[Bugfix][CB] Align k with old_k in process_qkv when retrieve_layer skips masked-prefix chunks

intelc:
**What this PR does / why we need it**:

Fixes the long-standing `RuntimeError: The size of tensor a (X) must match the size of tensor b (Y) at non-singleton dimension 0` in `lmcache/v1/compute/blend/blender.py:process_qkv`. This kills `EngineCore` mid-request and was reported in:

- #1875 — `[Bug] The tensor size of k and old_k mismatch on dimension 0 in blender.process_qkv`
- #854 — `CacheBlend fails on sending exact same prompt a second time`
- #938 — `RuntimeError in Blender When Handling 100% Cache Hit with Reordered Chunks` (acknowledged by @YaoJiayi as known but unfixed)

All three were closed as stale-without-fix.

## Root cause

Inside `blend_layer` two parallel iterators run over the same `tokens`:

| | what it sees | uses mask? |
|---|---|---|
| `compute_layer(tokens)` (in `LMCBaseModel`) | full forward over all `len(tokens)` positions | **no** |
| `retrieve_layer(tokens, mask, ...)` (in `LMCacheEngine`) | breaks at first content-hash miss; `process_tokens` skips chunks whose `start_idx < num_falses` | **yes** |

When `vllm_v1_adapter.start_load_kv` constructs the `FFFFFTTTTT` mask to skip vllm's prefix-cached prefix (`masked_token_count = vllm_cached_tokens // chunk_size * chunk_size`), `retrieve_layer` honors it and skips the leading chunks entirely. `gpu_connector.batched_to_gpu` then allocates `num_all_tokens = ends[-1] - starts[0]` of buffer (less than `lmcache_cached_tokens`). Result:

- `k.shape[0]` = `lmcache_cached_tokens` (e.g. 1292)
- `old_k.shape[0]` = `ends[-1] - starts[0]` (e.g. 801)
- `k.shape[0] - old_k.shape[0]` = `starts[0]` = position of first chunk past the masked-prefix region

The trailing `old_k.shape[0]` rows of `k` correspond to the same token positions as `old_k`.

## Why slice inside `check_layer` (not at top of `process_qkv`)

[...]

PR #3228, opened 2026-05-08
[Bugfix][CB] Fix sep_tokens BPE-context bug in SegmentTokenDatabase

intelc:

Fixes a tokenizer-context-sensitive bug in `SegmentTokenDatabase.__init__` that silently breaks CacheBlend retrieval to 0% on tokenizers with default-BOS prepending (Llama-2/3, Mistral, OPT). Symptom: zero segments detected → entire prompt stored as a single chunk → no cross-request reuse.

The bug has been latent because Qwen-family tokenizers don't prepend BOS by default, so `encode(s)[1:]` happens to leave a valid mid-text token, masking the issue in CacheBlend's primary published examples. Verified affected: Llama-3, OPT-125m. Verified working-by-accident: Qwen2.5.

## Root cause

Inside `SegmentTokenDatabase.__init__`:

```python
# TODO (Jiayi): figure out how to decide when
# to use `1:` (whether there's a special starting token
# in the beginning)
self.sep_tokens = self.tokenizer.encode(config.blend_special_str)[1:]
```

The `[1:]` slice strips a single leading token, presumably BOS. But:

1. On default-BOS tokenizers (Llama-3, OPT), `encode("# #")` produces e.g. `[BOS, '#', ' #']`. Slicing to `['#', ' #']` keeps a **start-of-string** `#` token (id 2 on Llama-3, 10431 on OPT) that **never appears mid-text** — every in-prompt occurrence of `#` is preceded by a word/space and merges into a different BPE token (id 674 / 849 respectively).
2. On no-BOS tokenizers (Qwen), the slice mis-strips a real content token, leaving a single token that happens to match space-prefixed `#` mid-text — explaining why Qwen-based examples work.
3. The common production configuration `blend_special_str = " # # "` (with surrounding spaces, used by the public CacheBlend RAG examples) makes it worse: even the bare encode-with-BOS-stripped path produces a 3-token sequence ending in a trailing-space token (id 220) that doesn't appear immediately after the second `#` in mid-text either.

Empirical confirmation across three tokenizers on a production-shape RAG prompt with 4 ` # # ` separators:

| Tokenizer | `blend=" # # "` old | `blend="# #"` old | `blend=" # # "` fix | `blend="# #"` fix |
|---|---:|---:|---:|---:|
| Llama-3.2-1B | 0/4 | 0/4 | **4/4** | **4/4** |
| Qwen2.5-0.5B | 0/4 | 8/4* | **4/4** | **4/4** |
| OPT-125m | 0/4 | 0/4 | **4/4** | **4/4** |

*Qwen `[671]` over-matches because the single-token marker hits both `#` tokens in each `# #` independently — over-segmentation, not the intended chunk boundaries.

## Fix

Replace the length-dependent slice with a tokenizer-agnostic encoding that:
- **strips surrounding whitespace** (`config.blend_special_str.strip()`) so a common production setting `" # # "` produces the same sep_tokens as the un-padded `"# #"`,
- **prepends exactly one leading space** so BPE merges anchor on the space-prefixed `' #'` form that occurs at every mid-text boundary,
- passes **`add_special_tokens=False`** so BOS is never inserted in the first place — no length-dependent slice needed.

```python
self.sep_tokens = self.tokenizer.encode(
    " " + config.blend_special_str.strip(),
    add_special_tokens=False,
)
```

This produces the right sep_tokens deterministically across the three tokenizers tested:

- Llama-3: `[674, 674]` (=`' #'`, `' #'`)
- Qwen2.5: `[671, 671]` (same)
- OPT-125m: `[849, 849]` (same)

…each matching exactly the 4 separator boundaries in the production RAG prompt.

## Verification

GPU end-to-end on `lmcache/vllm-openai:v0.4.3-lightweight` + Llama-3.3-70B-Instruct fp8 + 1×H100, with this fix and PR #3217 (the dim-fix) both applied. PR #3217 is required separately to avoid a downstream tensor-mismatch when retrieval engages on Llama; this PR is what makes retrieval engage in the first place on default-BOS tokenizers.

| Req | Total tokens | LMCache hit tokens | Hit rate |
|----:|-------------:|-------------------:|---------:|
| warmup | 36 | 0 | — |
| 1 | 1854 | 0 | 0% (cold) |
| 2 | 1865 | 936 | **50.2%** |
| 3 | 1855 | 1839 | **99.1%** |
| 4 | 1863 | 1839 | **98.7%** |
| 5 | 1862 | 1839 | **98.8%** |

Hit-rate progression matches the published Qwen3-8B baseline (0%, 49%, 99%, 99%, 99%) — the sep_tokens fix is the load-bearing piece that makes Llama behave like the working baseline.

CPU-only repro (parametric regression test included): \`tests/v1/test_token_database.py::test_segment_token_database_sep_tokens_match_real_text\` — parametrized over `(facebook/opt-125m, Qwen/Qwen2.5-0.5B-Instruct) × (" # # ", "# #")`. Asserts the resulting sep_tokens match all 4 separator boundaries in a production-shape RAG prompt. Runs in <1s without GPU.

- The fix is tokenizer-agnostic by construction — `add_special_tokens=False` prevents BOS injection regardless of tokenizer defaults; `.strip()` normalizes whitespace-padded blend strings; the leading space ensures BPE merges that occur in mid-text are produced.
- This bug has likely been present since `SegmentTokenDatabase` landed; the TODO above the line dates from the same era and called out exactly this uncertainty. The fix closes that TODO.
- Worth a maintainer pass on whether existing benchmark suites were running against tokenizers in the Qwen-family-by-accident bucket; their numbers may shift slightly if rerun (the over-segmentation case noted above goes away).
- Filed in parallel with PR #3217. They're independent fixes for separate bugs; #3217 fixes the `process_qkv` tensor-mismatch when retrieval engages, this one makes retrieval engage in the first place on default-BOS tokenizers.
- We've been carrying this as a downstream patch in our oneshot-pod boot script for several weeks of internal benching against Llama-3.3-70B; happy to add a third tokenizer to the test or any other shape change reviewers prefer.

**If applicable**:

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -431,10 +431,22 @@ class SegmentTokenDatabase(TokenDatabase):
 
         self.tokenizer = AutoTokenizer.from_pretrained(metadata.model_name)
 
-        # TODO (Jiayi): figure out how to decide when
-        # to use `1:` (whether there's a special starting token
-        # in the beginning)
-        self.sep_tokens = self.tokenizer.encode(config.blend_special_str)[1:]
+        # Encode blend_special_str so the resulting tokens match the BPE
+        # merges that occur when it appears mid-text. We strip surrounding
+        # whitespace from the configured separator, prepend exactly one
+        # leading space to anchor space-prefixed BPE merges, and pass
+        # add_special_tokens=False so no BOS is added. The prior length-
+        # dependent slice (encode(...)[1:]) was tokenizer-specific: it
+        # stripped a leading BOS that some tokenizers prepend (Llama-2/3,
+        # Mistral, OPT) but produced start-of-string tokens that never
+        # appear mid-text on those tokenizers. Symptom: zero segments
+        # detected → entire prompt stored as one chunk → 0% retrieval.
+        # Verified on Llama-3, Qwen2.5, OPT-125m for both `"# #"` and
+        # `" # # "` separator forms.
+        self.sep_tokens = self.tokenizer.encode(
+            " " + config.blend_special_str.strip(),
+            add_special_tokens=False,
+        )
         self.sep_tokens = torch.tensor(self.sep_tokens, device="cpu")
         self.sep_len = len(self.sep_tokens)
 
--- a/tests/v1/test_token_database.py
+++ b/tests/v1/test_token_database.py
@@ -132,3 +132,67 @@ def test_segment_token_database(prefix_length, chunk_lengths):
         assert key.chunk_hash == hashes[i]
         # print(st, starts[i])
         # print(ed, ends[i])
+
+
+@pytest.mark.parametrize(
+    "model_name", ["facebook/opt-125m", "Qwen/Qwen2.5-0.5B-Instruct"]
+)
+@pytest.mark.parametrize("blend_special_str", [" # # ", "# #"])
+@pytest.mark.skipif(
+    not hf_credentials_available(), reason="No Hugging Face credentials found"
+)
+def test_segment_token_database_sep_tokens_match_real_text(
+    model_name, blend_special_str
+):
+    """Regression test: ``SegmentTokenDatabase.sep_tokens`` must equal the
+    BPE token sequence produced when ``blend_special_str`` appears inside
+    a real prompt. The prior ``tokenizer.encode(blend_special_str)[1:]``
+    produced start-of-string tokens that never appear mid-text on tokenizers
+    with default-BOS prepending (Llama-2/3, Mistral, OPT) — silently
+    breaking CacheBlend retrieval to 0% on those models. Symptom: zero
+    segments detected → entire prompt stored as one chunk.
+
+    The fix encodes ``" " + blend_special_str.strip()`` with
+    ``add_special_tokens=False``: stripping surrounding whitespace
+    normalizes the common-but-fragile production setting ``" # # "``,
+    the leading space anchors the same BPE merge that occurs at every
+    mid-text occurrence, and ``add_special_tokens=False`` prevents BOS
+    being prepended in the first place.
+    """
+    cfg = LMCacheEngineConfig.from_legacy(blend_special_str=blend_special_str)
+    metadata = dumb_metadata_with_model_name(model_name)
+    db = SegmentTokenDatabase(cfg, metadata)
+    sep_tokens = db.sep_tokens.tolist()
+
+    # Production-shape user_content: passages joined by " # # " (matches the
+    # real CacheBlend RAG pattern). The configured blend_special_str may
+    # vary in whitespace (" # # " vs "# #"), but the literal text in mid-
+    # prompt is always the space-padded form — which the fix must handle.
+    sep_in_text = " # # "
+    user_content = (
+        "Reference passages:"
+        + sep_in_text
+        + sep_in_text.join(
+            [
+                "[Passage A]\nFirst passage text.",
+                "[Passage B]\nSecond passage text.",
+                "[Passage C]\nThird passage text.",
+            ]
+        )
+        + sep_in_text
+        + "Question: q?"
+    )
+    prompt_ids = db.tokenizer.encode(user_content, add_special_tokens=False)
+    n_seps_in_text = user_content.count(sep_in_text)
+
+    n = len(sep_tokens)
+    matches = sum(
+        1
+        for i in range(len(prompt_ids) - n + 1)
+        if prompt_ids[i : i + n] == sep_tokens
+    )
+    assert matches == n_seps_in_text, (
+        f"{model_name} blend_special_str={blend_special_str!r}: "
+        f"sep_tokens {sep_tokens} matched {matches} times in user_content "
+        f"(expected {n_seps_in_text}). prompt_ids[:30]={prompt_ids[:30]}"
+    )


ApostaC:
Hey @intelc , thanks for your interest in LMCache and Cacheblend! we are working on a cleaner Cacheblend implementation, which should be available around June.

Let me know if you are interested in trying that. Feel free to reach out in LMCache Slack as well (My name in Slack is `Yihua Cheng`)

VinayakVig:
Independent confirmation of this bug, plus two tokenizers where the fix as written still misses.

We hit this from the other direction — building a cross-node routing digest keyed on blend segments, and finding that `SegmentTokenDatabase` returned one segment covering the whole prompt. Same root cause you identified. Reproduced on **0.5.3**, and `token_database.py:466` is unchanged on `dev` as of today, so this is still live.

## Confirming the current behaviour

Driving the real class on CPU (`enable_blending: true`, default `blend_special_str`):

```
sep_tokens=[671, 220]  ->  1 segment(s), wanted 4
  [0, 76)  'The Eiffel Tower was completed in 1889 for the World...'
```

Worth stating plainly for anyone triaging: this is not "no speedup", it is **worse than the `ChunkedTokenDatabase` it replaces**, since one whole-prompt key is exact-match-only while prefix chunks at least share a common head.

## Where the proposed fix still misses

I ran three derivations over five tokenizers, with `blend_special_str = " # # "` and a 3-passage RAG-shaped prompt. The `+Nf` column counts **spurious** matches across six prose samples that mark no boundary at all (`issue #42`, a markdown `---` rule, `#include`, an index `a[SEP]`) — a pattern that fires there splits prompts nobody asked to split, which a boundary count alone hides.

| tokenizer | `encode(s)[1:]` (current) | `encode(" "+s.strip(), add_special_tokens=False)` (this PR) | derived in context |
|---|---|---|---|
| facebook/opt-125m | 0/3 ❌ | 3/3 ✅ | 3/3 ✅ |
| TinyLlama-1.1B (Llama-2 SP) | 0/3 ❌ | **0/3 ❌** | 3/3 ✅ |
| Qwen2.5-0.5B-Instruct | 0/3 ❌ | 3/3 ✅ | 3/3 ✅ |
| deberta-v3-base | 0/3 ❌ | 3/3 ✅ | 3/3 ✅ |
| Llama-Prompt-Guard-2-86M | 0/3 ❌ | **0/3 +1f ❌** | 3/3 ✅ |

Your OPT-125m and Qwen2.5 results reproduce exactly. The two failures are both cases where prepending a space produces a token that does not occur mid-text:

- **TinyLlama / Llama-2 (SentencePiece)** — `" # #"` encodes to `[29871, 396, 396]`, and `29871` is the standalone `▁` piece. Mid-text the space is absorbed into the following piece, so the pattern never matches. Note this is the *Llama-2* SentencePiece tokenizer, not the Llama-3 tiktoken BPE you verified on — I'd expect your Llama-3 result to stand, which makes this a gap in coverage rather than a contradiction.
- **Llama-Prompt-Guard-2-86M** — `" # #"` encodes to a single merged token `[124106]`, while in context the separator becomes `[73193]`. Different merge entirely, and `124106` also fires once in unmarked prose.

So `add_special_tokens=False` correctly removes the BOS/EOS half of the problem, and the `.strip()` + prepended space is a BPE-shaped assumption that doesn't hold across tokenizer families.

## The part that generalises

The failure mode both derivations share is deriving the pattern by encoding the separator **out of context**. Any such encoding can pick up a leading or trailing piece the tokenizer would have merged away in a real prompt. Recovering it in context is cheap, needs no per-tokenizer knowledge, and is a one-time cost at `__init__`:

```python
def _sep_tokens_in_context(tok, sep):
    # Longest common PREFIX of j1,j3 spans A1 then the separator; longest common
    # SUFFIX of j1,j2 spans the separator then B1. The slice inside both is the
    # separator. A1/A2 differ in their last character and B1/B2 in their first,
    # so anything that depends on the neighbouring text is excluded and what
    # survives is the context-invariant core -- which is what we want to match on.
    a1, a2 = "Passage one ends here.", "Second lead-in without punctuation"
    b1, b2 = "Passage two starts here.", "Mount Everest is a mountain."
    j1, j2, j3 = (tok.encode(x) for x in (a1+sep+b1, a2+sep+b1, a1+sep+b2))
    pre = 0
    while pre < min(len(j1), len(j3)) and j1[pre] == j3[pre]:
        pre += 1
    suf = 0
    while suf < min(len(j1), len(j2)) and j1[-1-suf] == j2[-1-suf]:
        suf += 1
    lo, hi = len(j1) - suf, pre
    return j1[lo:hi] if lo < hi else []
```

That is the third column above: 3/3 with zero false splits on all five tokenizers, including the two this PR still misses. Three extra `encode()` calls at init.

Two things it does not solve, worth naming so it isn't oversold: a separator whose token image genuinely varies with the preceding character has no context-invariant core and comes back empty, which should be a loud startup error rather than a silent zero-segment fallback; and it still can't stop a *legitimate* pattern from occurring in ordinary prose, so a warning when the derived pattern is a single common token would catch the `[671]`-style over-segmentation you flagged in your footnote.

Happy to send this as a PR against your branch, or as a test-only addition if you'd rather keep the fix as is — say which is more useful. All of the above runs on CPU in a few seconds, no GPU needed, since `SegmentTokenDatabase` pins to `device="cpu"`.

Unrelated to the derivation but adjacent, in case it helps whoever picks this up: `pre_caching_hash_algorithm` defaults to `"builtin"`, and when that can't be loaded from vLLM the fallback is Python's `hash()`. With `PYTHONHASHSEED` unset the same prompt keyed differently across two processes here (`0f3e9afcd1abd967` / `d1eb62a0b5ad6da0`), identical at `PYTHONHASHSEED=0`. The existing warning says exactly this; it just costs more than it looks like it does in a multi-node deployment.

PR #3231, opened 2026-05-08
[Feat] Server-side worker liveness tracking and reaping

royyhuang:

Today the MP server only has one-way health monitoring: vLLM adapters PING the server, but the server doesn't monitor adapters. When a vLLM worker dies silently, `gpu_contexts[instance_id]` and its KV-cache CUDA IPC handles linger until something triggers `unregister_kv_cache`, which never happens on a crash. Re-registration with the same `instance_id` is also unsafe because the old context isn't cleaned up.

This PR adds server-side liveness tracking for workers and a periodic reaper that evicts stale entries.

Highlights:

- **Random `instance_id`** — workers now use `random.getrandbits(63)` instead of `os.getpid()`. PIDs collide across containerized vLLM replicas that share PID namespaces; opaque random ids do not.
- **PING wire bump** — payload becomes `[instance_id]`; response stays `bool`. Workers send their `instance_id`; the scheduler-side adapter sends sentinel `0` (server-health probe only). Mismatched binaries fail loudly via msgspec.
- **Eager-start heartbeat** on the worker adapter (in `register_kv_caches` after the REGISTER ack), so PINGs flow within one heartbeat interval of registration regardless of warmup or first-traffic timing.
- **Tri-state `send_ping`** (`Optional[bool]`) — `True` healthy, `False` server rejected our id (terminal — heartbeat stops, operator restarts vLLM, no auto-recovery), `None` transient (cold-start grace tolerates 2 consecutive transient failures).
- **Server-side `_liveness` table + `_InstanceReaperThread`** — reaps stale entries past `reap_after_seconds` (default 30s, configurable via `--reap-after-seconds` / `LMCACHE_MP_REAP_AFTER_SECONDS`). CAS re-check inside `_reap_worker` handles the snapshot/refresh race.
- **`engine.close()`** stops the reaper before closing the storage manager.

Operator tuning rule (documented): keep `reap_after_seconds >= 3 * adapter_heartbeat_interval` so a single missed PING never reaps. The defaults satisfy this.

Out of scope, intentionally:

- **Scheduler liveness tracking and prefetch-job reap.** L1 read-lock TTL (default 300s) already bounds the chunk-pinning window. The remaining `_prefetch_jobs` Python dict-entry leak (the `server.py:208` TODO) is addressed in a separate PR; the choices here keep the door open without depending on it.
- **`assert` → `raise ValueError` in `store`/`retrieve`.** The MQ layer doesn't propagate handler exceptions today (`mq.py:419-482`); converting would silently hang the adapter's future. Tracked as a follow-up alongside MQ exception plumbing.
- **vLLM connector changes** (lives in the vLLM repo).
- **Observability** (`MP_INSTANCE_REAPED` event, gauges/counters, `report_status` live-worker count) — logging-only in this PR; metrics PR to follow.
- **`REPORT_BLOCK_ALLOCATION`** still uses `os.getpid()` server-side as event metadata. Trivial follow-up for consistency.

Design doc: `docs/design/v1/multiprocess/instance_liveness.md` — went through 7 rounds of expert review (vLLM-side and LMCache-side) before implementation.

Operator-visible behavior change to call out at release time: a server restart now degrades every connected vLLM worker until each is itself restarted (PING returns False → heartbeat stops → no auto-recovery). Sized for sidecar deployments where server lifecycle == vLLM lifecycle. If multi-tenant becomes painful, soft-recover-on-PING-False is the documented next-iteration choice.

**If applicable**:


lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -688,8 +742,12 @@ def __init__(
...
         self._mq_timeout = mq_timeout
 
-        # Instance id for GPU worker
-        self.instance_id = os.getpid()
+        # Instance id for GPU worker — opaque random 63-bit int generated
+        # once at adapter construction. Replaces the previous `os.getpid()`,
+        # which collided across containerized vLLM replicas that share PID
+        # namespaces (e.g. PID 1 in every replica). The server treats this
+        # value as opaque; no code on either side parses it as a PID.
+        self.instance_id = random.getrandbits(63)
```
ApostaC, line 750: We can use uuid.uuid4().int() to get a random 128-bit integer. (see https://docs.python.org/3/library/uuid.html#uuid.UUID.int)
Potential problem of using random: if the worker processes are initialized with the same random seed, they may have the same `instance_id`.

ApostaC: For the scheduler adapter, a potential solution is to use None instead of 0 for the instance id.

lmcache/v1/multiprocess/server.py
```diff
@@ -235,13 +329,12 @@ def register_kv_cache(
...
+        # old `GPUCacheContext` (and its CUDA IPC handles) before installing
+        # a new one. Cheap insurance.
         if instance_id in self.gpu_contexts:
-            logger.warning(
-                "Instance %s's KV cache is already registered, "
-                "skipping the new registration",
-                instance_id,
-            )
-            return
+            self.unregister_kv_cache(instance_id)
```
ApostaC, line 337: In PR #3208 (support LMCache restart), the semantics have been set to "noop" if the instance ID already exists.

This is because the vLLM-side may have duplicated registration if there is a transient PING check failure. In this case, we don't want to re-register the kv cache again since the creation of GPUCacheContext may take a long time.

lmcache/v1/multiprocess/server.py
```diff
@@ -262,15 +360,68 @@ def unregister_kv_cache(self, instance_id: int) -> None:
...
+            entry = self._liveness.get(instance_id)
+            if entry is None:
+                return
+            if snapshot_now - entry.last_seen <= self.reap_after_seconds:
+                return  # raced with a refresh
+            staleness = snapshot_now - entry.last_seen
+            del self._liveness[instance_id]
+        # `unregister_kv_cache` is idempotent; benign if `gpu_contexts` was
+        # already cleared via a concurrent UNREGISTER_KV_CACHE handler.
+        self.unregister_kv_cache(instance_id)
```
ApostaC, line 419: Since we are calling `unregister_kv_cache` from multiple threads, it's important to make sure that this function is thread-safe.

docs/design/v1/multiprocess/instance_liveness.md
ApostaC, line 1: After reading this doc, I found there are quite a few intermediate AI-generated planning details. It would be better if we only keep the "description of the final design" in the design doc instead of having all the "thought processes" here.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -31,6 +33,13 @@
 # Interval (seconds) between periodic heartbeat pings to the server.
 DEFAULT_HEARTBEAT_INTERVAL: float = 10.0
 
+# Number of consecutive failed pings tolerated before the heartbeat thread
+# clears `_health_event` for the *first* time on a fresh adapter. After
+# the first successful ping latches `_first_success_seen`, any single
+# steady-state failure flips immediately. Absorbs a one-off GIL stall
+# during warmup without weakening steady-state detection.
+COLD_START_FAILURE_THRESHOLD: int = 2
+
```
ApostaC, line 42: Can you explain why we are introducing this? I feel like this could be an over design

ApostaC: If we simply treat the "None" as "False" in `send_ping`'s response, a lot of changes are not needed.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -205,67 +245,74 @@ def __init__(
...
-        """
-        self._recover_callback = callback
+        # Cold-start grace state. `_first_success_seen` latches True on
+        # the first healthy ping and is never reset, so steady-state
+        # detection (any single failure flips immediately) is restored
+        # for the rest of the thread's lifetime.
+        self._consecutive_failures: int = 0
+        self._first_success_seen: bool = False
 
     def _execute(self) -> ThreadRunSummary:
```
ApostaC, line 257: This function is updated by #3208. Did this PR updated to the latest dev or not?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -205,67 +245,74 @@ def __init__(
...
+            # before flipping unhealthy on a fresh adapter that hasn't
+            # yet observed a single success.
+            return ThreadRunSummary(
+                success=True,
+                message=(
+                    f"cold-start-fail "
+                    f"({self._consecutive_failures}/"
+                    f"{COLD_START_FAILURE_THRESHOLD})"
+                ),
+            )
```
ApostaC, line 311: I feel like this part is unnecessarily complicated. Maybe let's think about what the clear logic should be here without using AI. (My Opus 4.7 has the same problem of getting everything super complicated)

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -845,8 +885,17 @@ def _send_register_kv_caches_request(
...
+        self._ensure_heartbeat_started()
+
     def _ensure_heartbeat_started(self) -> None:
-        """Lazily start the heartbeat thread on first use."""
+        """Start the heartbeat thread (idempotent).
+
+        Called from `register_kv_caches` after the REGISTER ack. The
+        previous lazy-start on first store/retrieve has been removed: by
+        starting here we get bounded "register-to-first-PING" latency
+        regardless of when (or whether) traffic flows.
```
ApostaC, line 897: nit: No need to have this docstring here.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1171,8 +1180,14 @@ def get_block_ids_with_load_errors(self) -> set[int]:
 
     def shutdown(self):
         """
-        Shutdown the LMCache MP worker adapter
+        Shutdown the LMCache MP worker adapter.
+
+        Stops the heartbeat thread first so it can't issue a PING on a
+        closing mq_client, then sends UNREGISTER_KV_CACHE and closes
+        the client / telemetry.
```
ApostaC, line 1187: nit: this docstring is not necessary.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -789,41 +841,29 @@ def is_first_rank_of_pp_group(self) -> bool:
...
-    def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]) -> None:
+    def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         """
         Register the kv caches with LMCache server.
 
+        After the REGISTER ack succeeds, the heartbeat thread is started.
+        Doing it here (rather than lazily on first store/retrieve) means
+        the server starts receiving PINGs within one heartbeat interval,
+        independent of warmup duration or whether traffic has begun. On
+        REGISTER failure, no heartbeat thread is created.
```
ApostaC, line 852: nit: we can simplify this to 
```plaintext
Also starts the heartbeat thread.
```

lmcache/v1/multiprocess/config.py
```diff
@@ -167,6 +177,22 @@ def add_mp_server_args(
         'Example: \'{"plugin.frontend.heartbeat_url": '
         '"http://localhost:5000/heartbeat"}\'',
     )
+    mp_group.add_argument(
+        "--reap-after-seconds",
+        type=float,
+        default=30.0,
```
ApostaC, line 183: I feel like a safer option is to have 1 or 2 minutes?

lmcache/integration/vllm/vllm_multi_process_adapter.py
ApostaC, line 852: @maobaolong I remember you previously moved the start of the heartbeat thread to `store`/`retrieve`. Can you remind us what the problem was back then?

lmcache/v1/multiprocess/server.py
```diff
@@ -169,13 +175,82 @@ class _PrefetchJob:
...
+                iid
+                for iid, st in self._engine._liveness.items()
+                if now - st.last_seen > deadline
+            ]
+        for iid in stale:
+            self._engine._reap_worker(iid, now)
+        return ThreadRunSummary(
+            success=True,
+            message=(f"reaped={len(stale)} alive={len(self._engine._liveness)}"),
+        )
```
ApostaC, line 242: We should avoid using the private members or methods of a class. Consider implementing a public method in `MPCacheEngine` that does the liveness check and stale kv cache reaping.

ApostaC (changes requested):
Please see the comments below. Mostly related to AI-generated bad code.

lmcache/v1/multiprocess/server.py
```diff
@@ -169,13 +175,82 @@ class _PrefetchJob:
...
+        # `unregister_kv_cache` (which can trigger `torch.cuda.empty_cache`)
+        # and we don't want to hold a fast lock across GPU work.
+        with self._engine._liveness_lock:
+            stale = [
+                iid
+                for iid, st in self._engine._liveness.items()
+                if now - st.last_seen > deadline
+            ]
+        for iid in stale:
+            self._engine._reap_worker(iid, now)
```
chunxiaozheng, line 238: `_reap_worker` should be a private method?

lmcache/v1/multiprocess/server.py
```diff
@@ -169,13 +175,82 @@ class _PrefetchJob:
...
+
+    def _execute(self) -> ThreadRunSummary:
+        now = time.monotonic()
+        deadline = self._engine.reap_after_seconds
+        # Snapshot stale candidates under the lock so the iteration can't
+        # race a concurrent PING insert/refresh ("dictionary changed size
+        # during iteration"). Cleanup runs outside the lock — it calls
+        # `unregister_kv_cache` (which can trigger `torch.cuda.empty_cache`)
+        # and we don't want to hold a fast lock across GPU work.
+        with self._engine._liveness_lock:
```
chunxiaozheng, line 231: the variable begin with `_`, such as `_liveness_lock`, should be a private variable?

chunxiaozheng:
@royyhuang There are some conflicts, could you help resolve it?

lmcache/integration/vllm/vllm_multi_process_adapter.py
maobaolong: @ApostaC Yeah, 
the https://github.com/LMCache/LMCache/pull/2943 and 
https://github.com/LMCache/LMCache/pull/2798 are the related PRs. The system might be marked as unhealthy prematurely during long model warmup phases without the two PRs.

PR #3233, opened 2026-05-09
 [Observability] Surface CB-registered GPU contexts in /api/statusreport cb server status

deng451e:

  `BlendEngineV2`'s inherited `report_status()` only sees normal vLLM-connector
  registrations, so CB-only deployments (`engine_type="blend"` with no
  `register_kv_cache` call) report empty `registered_gpu_ids` /
  `gpu_context_meta` even when a CB engine is actively connected —
  indistinguishable from "no engine connected at all" to `/api/status` clients.

  This PR overrides `BlendEngineV2.report_status()` to add two parallel fields
  populated from contexts registered via `cb_register_kv_cache`:

  - `registered_cb_gpu_ids`: list of CB instance IDs (disjoint from the base
    `registered_gpu_ids` namespace).
  - `cb_gpu_context_meta`: per-instance `model_name`, `world_size`, and a
    `kv_cache_layout` reflecting `PlainGPUCacheContext`'s
    `[2, num_layers, num_tokens, hidden_dim_size]` shape.

### DIFF
--- a/lmcache/cli/commands/bench/engine_bench/config.py
+++ b/lmcache/cli/commands/bench/engine_bench/config.py
@@ -173,6 +173,10 @@ def resolve_tokens_per_gb(lmcache_url: str, model_name: str) -> int:
 
     gpu_meta = data.get("gpu_context_meta", {})
     if not gpu_meta:
+        # CB-only deployments (engine_type="blend") populate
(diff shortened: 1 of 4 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3235, opened 2026-05-09
[MP]add the lmcache_mp_connector for dev

chunxiaozheng:
Maintain the latest `lmcache_mp_connector` in `LMCache`.

The file is copied from the latest VLLM, nothing changed.

### DIFF
--- a/docs/source/getting_started/quickstart.rst
+++ b/docs/source/getting_started/quickstart.rst
@@ -58,6 +58,30 @@ This guide helps you get LMCache running end-to-end in a couple of minutes. Use 
                    --port 8000 --kv-transfer-config \
                    '{"kv_connector":"LMCacheMPConnector", "kv_role":"kv_both"}'
 
+            .. note::
+               **Where does** ``LMCacheMPConnector`` **resolve to?** This depends on your vLLM version:
+
+               - **vLLM < 0.20.0** -- ``"kv_connector":"LMCacheMPConnector"`` always
+                 resolves to vLLM's built-in
+                 ``vllm.distributed.kv_transfer.kv_connector.v1.LMCacheMPConnector``;
+                 there is no way to redirect it to the LMCache-shipped implementation.
+
+               - **vLLM >= 0.20.0** -- ``"kv_connector":"LMCacheMPConnector"`` still
+                 defaults to vLLM's built-in connector, but you can opt in to the
+                 LMCache-shipped implementation
+                 (:mod:`lmcache.integration.vllm.lmcache_mp_connector`) by adding
+                 ``kv_connector_module_path``:
+
+                 .. code-block:: bash
+
+                    vllm serve Qwen/Qwen3-8B \
+                        --port 8000 --kv-transfer-config \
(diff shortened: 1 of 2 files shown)


maobaolong (approved):
lgtm

chunxiaozheng:
@ApostaC I have looked at the implementation code of get connector in VLLM, and if we only pass `kv_comnector:  LMCacheMPConnector`, it will still go to the connector in VLLM. We need to set `kv_comnector_madule_path` to point to the moudle in LMCache in order to go to the connector in LMcache.
```
    def _get_connector_class_with_compat(
        cls, kv_transfer_config: "KVTransferConfig"
    ) -> tuple[type[KVConnectorBaseType], bool]:
        connector_name = kv_transfer_config.kv_connector
        if connector_name is None:
            raise ValueError("Connector name is not set in KVTransferConfig")
        compat_sig = False
        connector_module_path = kv_transfer_config.kv_connector_module_path
        if connector_module_path is not None and not connector_module_path:
            raise ValueError("kv_connector_module_path cannot be an empty string.")
        if connector_module_path:
            # External module path takes priority over internal registry.
            connector_module = importlib.import_module(connector_module_path)
            try:
                connector_cls = getattr(connector_module, connector_name)
            except AttributeError as e:
                raise AttributeError(
                    f"Class {connector_name} not found in {connector_module_path}"
                ) from e
            connector_cls = cast(type[KVConnectorBaseType], connector_cls)
            if not supports_kw(connector_cls, "kv_cache_config"):
                compat_sig = True
                logger.warning(
                    "Connector %s uses deprecated signature with 2 required arguments. "
                    "Please update to include kv_cache_config as the second argument.",
                    connector_cls.__name__,
                )
        elif connector_name in cls._registry:
            connector_cls = cls._registry[connector_name]()
        else:
            raise ValueError(f"Unsupported connector type: {connector_name}")
        return connector_cls, compat_sig
```

chunxiaozheng:
@ApostaC could you help take another look?

ApostaC:
@chunxiaozheng follow-up question: for the `kv_comnector_module_path`, do we need a physical absolute path (e.g., something like `/opt/venv/lib/python3.12/site-packages/lmcache`)?

ApostaC (approved):
Giving an approval here. But please update the user-facing docs as we discussed offline. Could be a "note" section in https://docs.lmcache.ai/getting_started/quickstart.html

chunxiaozheng:
> @chunxiaozheng follow-up question: for the `kv_comnector_module_path`, do we need a physical absolute path (e.g., something like `/opt/venv/lib/python3.12/site-packages/lmcache`)?



> @chunxiaozheng follow-up question: for the `kv_comnector_module_path`, do we need a physical absolute path (e.g., something like `/opt/venv/lib/python3.12/site-packages/lmcache`)?

@ApostaC the `kv_comnector_madule_path` is the module path, such as `lmcache.integration.vllm.lmcache_mp_connector`

chunxiaozheng:
> Giving an approval here. But please update the user-facing docs as we discussed offline. Could be a "note" section in https://docs.lmcache.ai/getting_started/quickstart.html

done!

PR #3237, opened 2026-05-09
Refactor: abstract discover_subclasses util method

maobaolong:

**If applicable**:


DongDongJu (commented):
Thanks @maobaolong , this refactor looks good overall.

One thing, before this PR, CLI commands were explicitly imported, 
so if one command module was broken, `lmcache` failed immediately. 

After this PR, `discover_subclasses()` catches import errors and continues, 
which means a broken command could disappear from the CLI.

For CLI commands, I think we should keep the old behavior: import errors should raise instead of being only logged/skipped.  WDYT?

Otherwise, LGTM

maobaolong:
@DongDongJu Thanks for your review. The new commit addressed the comment, PTAL.

DongDongJu (approved):
Thanks for addressing the issue.
LGTM.
one nit, `test_cache` was intended to include to all_commands in top-level CLI?
maybe we can do `module_filter=lambda name: name not in {"base", "test_cache"}` like things.

maobaolong:
@DongDongJu good catch — `test_cache` is the impl behind `bench kvcache`, not a top-level verb. Added it to the `module_filter` excludes and a small regression test so it can't sneak back in. PTAL.

maobaolong:
Updated: instead of blacklisting `test_cache` in the top-level filter, just moved the module to where it belongs — `lmcache/cli/commands/bench/test_cache.py`, sitting next to its `engine_bench` peer. Top-level discovery now naturally won't see it. Thanks again for the nudge!

chunxiaozheng (approved):
As discussed offline, LGTM, and I will refactor the bench cli later.

PR #3240, opened 2026-05-10
[CI]: expose prefiller/decoder/proxy logs as artifacts

sammshen:

Server-side logs from the comprehensive PD test (prefiller, decoder, disagg proxy) are not reliably showing up in the Buildkite Artifacts tab. They were being written to `/tmp` and then `cp`'d to `REPO_ROOT` inside the script's `EXIT` trap so the existing `artifact_paths: ["*.log", ...]` glob would catch them. That `cp` is silent (`|| true`), races against pod teardown, and PD failures often leave us without the very logs we need to debug.

This PR writes those three logs directly to `REPO_ROOT` under stable names so they are unconditionally captured by the existing `*.log` artifact glob — no new globs, no trap dependency.

After merge, every comprehensive PD build will have these in Artifacts:
- `pd-prefiller.log`
- `pd-decoder.log`
- `pd-proxy.log`

- Single-server and P2P log paths are unchanged on purpose; they hit a different code path and aren't currently a debugging pain point.
- The runner's tee'd stdout (`pd.log`) and the workload `response.txt` continue to be uploaded as before.

**If applicable**:


maobaolong (approved):
LGTM

deng451e (approved):

PR #3245, opened 2026-05-10
Retain CUDA IPC events in MP adapter

### LINKED ISSUE #3236: [Bug][MP] LMCacheMPWorkerAdapter does not retain producer-side CUDA Event after exporting ipc_handle
**Label**
bug, MP, vLLM integration

**Describe the bug**

When using LMCache MP mode with a standalone LMCache daemon, `LMCacheMPWorkerAdapter` sends CUDA event IPC handles to the daemon but does not appear to retain the producer-side `torch.cuda.Event` object.

In our environment, the daemon may fail when opening the event IPC handle:

```text
RuntimeError: CUDA error: invalid argument
torch.cuda.Event.from_ipc_handle(...)
```

The relevant path is:

1. vLLM `LMCacheMPConnector` creates and records an interprocess CUDA event.
2. The event is passed into `LMCacheMPWorkerAdapter`.
3. `LMCacheMPWorkerAdapter.submit_store_request()` / `submit_retrieve_request()` sends `event.ipc_handle()` to the LMCache MP daemon.
4. The adapter stores the `MessagingFuture`, but not the original `torch.cuda.Event` object.
5. If the producer-side event object is garbage-collected before the daemon opens the handle, the daemon may fail in `torch.cuda.Event.from_ipc_handle()`.

This looks like a CUDA event lifetime issue in the LMCache MP adapter.

**To Reproduce**

Our deployment topology:

- vLLM runs in Kubernetes pods.
- LMCache runs as a standalone MP daemon pod on the same node.
- vLLM connects to the daemon through `LMCacheMPConnector`.
- vLLM and the daemon use host IPC for CUDA IPC.
- The daemon receives `event.ipc_handle()` from vLLM and opens it with `torch.cuda.Event.from_ipc_handle(...)`.

vLLM is launched with:

```bash
--kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both","kv_connector_extra_config":{"lmcache.mp.host":"tcp://${NODE_IP}","lmcache.mp.port":6555}}'
```

The issue is observed during MP store/retrieve when the daemon opens the event IPC handle.

**Expected behavior**

The producer-side `torch.cuda.Event` should remain alive until the LMCache MP daemon has opened the IPC handle and the corresponding store/retrieve operation has completed.

The MP adapter should manage the lifetime of the `torch.cuda.Event` object, not only the messaging future.

**Actual behavior**

The daemon may fail with:

```text
RuntimeError: CUDA error: invalid argument
torch.cuda.Event.from_ipc_handle(...)
```

**Relevant code path**

In the vLLM connector path, an interprocess event is created and recorded:

```python
event = torch.cuda.Event(interprocess=True)
event.record()
self.worker_adapter.batched_submit_store_requests(request_ids, ops, event)
```

Then `LMCacheMPWorkerAdapter.submit_store_request()` sends only the IPC handle:

```python
future = send_lmcache_request(
    self.mq_client,
    RequestType.STORE,
    [key, self.instance_id, op.block_ids, event.ipc_handle()],
).to_cuda_future()
self.store_futures[request_id] = future
```

Similarly, `submit_retrieve_request()` sends:

```python
future = send_lmcache_request(
    self.mq_client,
    RequestType.RETRIEVE,
    [
        key,
        self.instance_id,
        op.block_ids,
        event.ipc_handle(),
        op.skip_first_n_tokens,
    ],
).to_cuda_future()
self.retrieve_futures[request_id] = (future, list(op.block_ids))
```

On the daemon side, the handle is opened with:

```python
vllm_event = torch.cuda.Event.from_ipc_handle(
    gpu_context.device, event_ipc_handle
)
```

The adapter stores `store_futures` / `retrieve_futures`, but it does not retain the original producer-side `event` object.

**Workaround tested locally**

We tested a local workaround that keeps event references inside `LMCacheMPWorkerAdapter` until the corresponding future completes.

Conceptually:

```python
self._mp_event_refs[("store", request_id)] = event
self._mp_event_refs[("retrieve", request_id)] = event
```

Then after `get_finished()` removes completed entries from `store_futures` / `retrieve_futures`, the corresponding event references are removed.

With this workaround, the `Event.from_ipc_handle(...): CUDA error: invalid argument` failure no longer occurs in our MP deployment.

**Environment**

- LMCache: dev commit `06981d6a3`
- vLLM: 0.17-based build
- Model: `Qwen/Qwen3-32B`
- Deployment: Kubernetes
- LMCache mode: MP mode, standalone daemon

he-yufeng:
﻿**What this PR does / why we need it**:

Fixes #3236.

`LMCacheMPWorkerAdapter` now keeps the producer-side CUDA event objects alive for pending MP store and retrieve requests. The adapter already tracks the matching futures; this adds matching event references and drops them when the future completes or when pending work is drained after the server becomes unhealthy.

Without retaining the event object, the daemon can receive an IPC handle whose producer-side event has already been collected, which can make `torch.cuda.Event.from_ipc_handle(...)` fail with `CUDA error: invalid argument`.

The unit tests use a fake CUDA event plus `weakref` to check both sides of the lifetime: pending requests keep the event alive, and `get_finished()` releases it once the matching future completes.

Validation run locally on Windows with a CPU/source-only test environment:

```text
python -m ruff check lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
python -m ruff format --check lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
.\.venv\Scripts\python.exe -m pytest tests\v1\test_vllm_mp_adapter.py -q
.\.venv\Scripts\python.exe -m py_compile lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
git diff --check
git fsck --no-progress
```

**If applicable**:


KuntaiDu (approved):
LGTM! @ApostaC @YaoJiayi can you also check?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -700,6 +700,10 @@ def __init__(
         self.retrieve_futures: dict[
             str, tuple[MessagingFuture[RetrieveResult], list[int]]
         ] = {}
+        # The IPC handle is not enough by itself; CUDA needs the exporting
+        # event object to stay alive until the consumer is done with it.
+        self.store_events: dict[str, torch.cuda.Event] = {}
```
maobaolong: @hlin99 After the torch_dev abstraction, we can not ues torch cuda directly, please take a confirm.

maobaolong (changes requested):

lmcache/integration/vllm/vllm_multi_process_adapter.py
hlin99: yes cq check fails

he-yufeng:
﻿Follow-up pushed in 6dd0685.

Changes:
- replaced direct torch.cuda.Event annotations with a small structural ipc_handle() protocol, so this adapter API no longer couples the type surface to torch.cuda;
- annotated the fake heartbeat callback slot to clear the mypy assignment error.

Validation on Windows:
- python -m py_compile lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
- git diff --check
- python -m ruff check lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
- python -m mypy --config-file pyproject.toml lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py

Note: python -m pytest tests\v1\test_vllm_mp_adapter.py -q is blocked locally before collection by the missing lmcache.c_ops native extension in this Windows checkout.

he-yufeng:
﻿DCO is fixed now. I rebased the two PR commits with Signed-off-by footers and force-pushed the same code diff; current head is 530fd55. The temporary bad squash was immediately reverted, and the PR diff is back to the intended two files.

he-yufeng:
Rebased onto the latest dev and resolved the test conflict; current head is 383d485. The code diff is still limited to the MP adapter and its regression tests. Validation on Windows: python -m py_compile lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py; python -m ruff check lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py; python -m mypy --config-file pyproject.toml lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py; git diff --check upstream/dev...HEAD. Targeted pytest is still blocked before collection by missing lmcache.c_ops in this Windows checkout.

he-yufeng:
﻿Updated again in 3cd6b98.

Changes:
- rebased the PR onto current dev;
- fixed the CI failure in the new event-retention tests by stubbing the current TransferContext path instead of the older send_lmcache_request/to_cuda_future path;
- the earlier torch_dev review concern is addressed in 872da20: the adapter retains IPC-capable events through a small _IpcEvent protocol and no longer exposes a torch.cuda.Event type dependency in this adapter API.

Validation on Windows:
- python -m py_compile lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
- python -m ruff check lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
- python -m mypy --config-file pyproject.toml lmcache\integration\vllm\vllm_multi_process_adapter.py tests\v1\test_vllm_mp_adapter.py
- git diff --check

The targeted pytest command is still blocked locally before collection in this Windows checkout by the missing native lmcache.c_ops extension; the previous CI failure itself should be covered by the Linux test job after this push.

he-yufeng:
Pushed 195edaa to fix the Linux CI failures in the new event-retention tests. The adapter was already clearing store_events / retrieve_events; the test mock's call_args was still holding the FakeCudaEvent reference, so the weakref assertion was testing the fixture rather than adapter cleanup. Validation on Windows: py_compile for the adapter and test, ruff check on touched files, mypy on touched files, and git diff --check passed. The targeted pytest still cannot collect locally on this Windows host because lmcache.c_ops is not built.

he-yufeng:
﻿Follow-up for the CI failure: the test was still retaining the fake CUDA event through the parent `transfer_ctx` mock's call history. I switched the cleanup to reset the parent mock, so the weakref check now verifies the adapter no longer owns the event after `get_finished()` removes it.

Validated locally:

- `python -m py_compile tests\v1\test_vllm_mp_adapter.py`
- `ruff check tests\v1\test_vllm_mp_adapter.py`
- `mypy tests\v1\test_vllm_mp_adapter.py`
- `git diff --check`

`python -m pytest tests\v1\test_vllm_mp_adapter.py -q` is still blocked in my Windows checkout before test collection by `ModuleNotFoundError: No module named 'lmcache.c_ops'` from `tests/conftest.py`.

maobaolong (approved):
LGTM.

maobaolong:
@hlin99 Would you like to take a double check?

hlin99 (approved):
LGTM

PR #3246, opened 2026-05-10
[Feat][MP] Use io_uring fixed buffers for MP raw_block payload I/O

DongDongJu:

This PR makes MP `raw_block` faster when it is configured to use io_uring.

When `io_engine="io_uring"` and `enable_zero_copy=true`, LMCache now registers the MP L1 CPU memory area with the raw-block io_uring backend. If a KV cache buffer is inside that registered memory area, raw_block can use io_uring fixed-buffer reads and writes for the payload.

If fixed-buffer I/O cannot be used, LMCache keeps using the existing raw-block path.

This PR also updates the raw-block docs to describe the current behavior.

  **If applicable**:


DongDongJu:
@ankit-sam will working on this thread.

PR #3248, opened 2026-05-10
[MP][feat] Support multi LMCache servers deployment with vllm

Lyj1007:
### What this PR does / why we need it

This PR extends the LMCache vLLM multi-process connector to support multi-node deployments with multiple LMCache servers. The vLLM scheduler connects to **all** servers simultaneously for cache lookup, while each worker process connects only to the **local** server on its own node. The primary target is large TP deployments (e.g. TP=16) split across two physical machines:

```
Node 0: GPU 0-7  ←→  LMCache Server 0
Node 1: GPU 8-15 ←→  LMCache Server 1
```

Both MLA models (e.g. DeepSeek-V2) and non-MLA models (e.g. LLaMA) are supported.

### Highlights

- **Per-node ObjectKey space** — each server manages an independent key space scoped to its local workers; `world_size`, `tp_size`, and `worker_id` embedded in every cache key are adjusted to reflect the per-node view rather than the global TP world.
- **Scheduler fan-out with `min()` aggregation** — `LOOKUP` / `QUERY_PREFETCH_STATUS` are submitted to all servers via non-blocking `MessagingFuture`s and then awaited sequentially with a per-server `mq_timeout`, so the servers' processing overlaps in time without an explicit thread pool; a chunk counts as a hit only when **every** server reports it, ensuring cross-node cache consistency.
- **Worker-side local routing** — each worker connects only to its local server, determined by `server_urls[vllm_worker_id // ranks_per_node]`.
- **Per-node STORE gating for MLA** — under MLA, only the first rank on each node is a KV writer (`vllm_worker_id % (tp_size // n_servers) == 0`), since all ranks on the same node hold identical KV data.
- **Per-server health tracking** — one `HeartbeatThread` and one health `Event` per server; `is_healthy` is the AND of every per-server event, and lock-release / cleanup paths are skipped when any server is unhealthy.
- **Tail-lock cleanup on hit mismatch** — when servers disagree on hit counts, an internal helper releases the tail range `[min_hit_end, this_server_hit_end)` on every server whose hit count exceeded the global min, so over-hitting servers do not retain read locks on chunks vLLM never retrieves.
- **Full backward compatibility** — when `lmcache.mp.server_urls` is absent the connector falls back to legacy `lmcache.mp.host` / `lmcache.mp.port` keys; `n_servers` defaults to 1 and all multi-server paths degenerate to the original single-server behavior.

---

## Key design invariant: per-node ObjectKey space
Each LMCache server manages an independent ObjectKey space, isolated by the `(model_name, world_size, worker_id)` triple embedded in every cache key. The `world_size` / `worker_id` carried in the key are the **per-node** view (`ParallelStrategy.kv_world_size` / `kv_worker_id`), not the global TP world. Let `N = n_servers`:

| Scenario | kv_world_size | kv_tp_size | kv_worker_id | is_kv_writer |
|---|---|---|---|---|
| Single-server, MLA              | `vllm_world_size // tp_size = pp_size` | `tp_size`     | `vllm_worker_id // tp_size`              | `vllm_worker_id % tp_size == 0` |
| Single-server, non-MLA          | `vllm_world_size`                      | `tp_size`     | `vllm_worker_id`                         | always True |
| Multi-server, MLA, only TP      | `vllm_world_size // tp_size = pp_size` | `tp_size / N` | `vllm_worker_id // tp_size`              | `vllm_worker_id % (tp_size / N) == 0` |
| Multi-server, non-MLA, only TP  | `vllm_world_size / N`                  | `tp_size / N` | `vllm_worker_id % (vllm_world_size / N)` | always True |
| Multi-server, MLA, TP and PP    | not supported yet                      |               |                                          | |
| Multi-server, non-MLA, TP and PP| `vllm_world_size / N`                  | `tp_size / N` | `vllm_worker_id % (vllm_world_size / N)` | always True |

- **`kv_world_size`** — number of ObjectKey shards a single chunk is split into. A LOOKUP key is expanded into `kv_world_size` ObjectKeys (one per `worker_id ∈ [0, kv_world_size)`); STORE/RETRIEVE address one specific shard. The Scheduler (LOOKUP) and Workers (STORE/RETRIEVE) must agree on this value, otherwise LOOKUP probes a different shard count than what is actually written and hits become impossible.
- **`kv_tp_size`** — TP fan-in seen by **one** server. The Worker passes it as the `tp_size` argument of LOOKUP; the server then computes `extra_count = compute_extra_count(tp_size, world_size)` (= `tp_size − 1` for MLA, `0` for non-MLA), reserving exactly one read-lock slot per local TP reader so locks fully drain after RETRIEVE.
- **`kv_worker_id`** — per-shard index in `[0, kv_world_size)`. For non-MLA multi-server, `vllm_worker_id % kv_world_size` makes workers on different nodes map into the same `[0, kv_world_size)` range, so each node addresses the right slot in its local server's ObjectKey space. For MLA, `vllm_worker_id // tp_size` collapses TP workers onto a shared id, since TP does not shard the KV cache.
- **`is_kv_writer`** — whether this rank actually issues STOREs. Non-MLA: every rank writes its own shard. MLA: only one rank per server writes, the rest are read-only.

### Lock balance example (TP=16, 2 servers, MLA)

```
world_size       = kv_world_size         = 1   (MLA: TP does not shard KV)
tp_size_per_node = kv_tp_size            = 8
extra_count      = compute_extra_count(8, 1)   = 7

LOOKUP   → reserve_read: 1 + 7 = 8 locks per ObjectKey
RETRIEVE → 8 workers × finish_read_prefetched(extra_count=0)
         = 8 locks released
         → 0 residual locks, eviction unblocked
```

---

## Code changes vs. the single-server connector

The multi-server changes live in two files.

### `lmcache_mp_connector.py`

- Adds parsing for `lmcache.mp.server_urls` (list or comma-separated), with fallback to legacy `host` / `port` when `server_urls` is absent.
- Builds a single `ParallelStrategy` via `build_parallel_strategy_from_vllm_config(vllm_config, n_servers)` and passes it to the scheduler/worker adapters; the worker pins itself to its local server via `server_urls[vllm_worker_id // ranks_per_node]`.
- Init-time guards: when `n_servers > 1`, `dp_size > 1` is rejected outright; `pp_size > 1` is rejected only under MLA (non-MLA PP is unaffected).
- MLA STORE gating in `wait_for_save` switches to the per-node `worker_adapter.is_kv_writer`.

### `vllm_multi_process_adapter.py`

- **`ParallelStrategy`** centralises the per-node KV geometry: `kv_world_size`, `kv_tp_size`, `kv_worker_id`, and `is_kv_writer` are all derived from `(use_mla, vllm_world_size, vllm_worker_id, tp_size, pp_size, n_servers)`.
- **`LMCacheMPSchedulerAdapter`** now takes `server_urls: list[str]` and holds one `MessageQueueClient` + one `HeartbeatThread` + one health `Event` per server (heartbeats start lazily on first lookup); it asserts a single unique `chunk_size` across all servers at init. `LOOKUP` / `check_lookup_result` fan out to all servers as non-blocking `MessagingFuture`s and then await each with a per-server `mq_timeout`. `check_lookup_result` records each server's final hit count exactly once into `_per_server_hits`, returns `None` until every server has reported, and then returns `min(per_server_hits) * chunk_size` cached until `cleanup_lookup_result`; on a `min != max` mismatch it logs and internally frees the per-server tail `[min_hit_end, this_server_hit_end)` so over-hitting servers do not retain stale read locks. `free_lookup_locks` / `end_session` / `report_block_allocations` iterate every server and short-circuit when `is_healthy` (AND of all per-server events) is false. `world_size` / `tp_size` properties expose the per-node values.
- **`LMCacheMPWorkerAdapter`** still takes a single `server_url` (each worker only ever talks to its local server); `world_size` / `worker_id` / `is_kv_writer` are read straight from `ParallelStrategy` and naturally collapse to the legacy single-server values when `n_servers == 1`.

---

## Current limitations

- **DP is not supported.** For MLA models, **PP is also not supported** (only TP). Non-MLA models work with both TP and PP. DP and MLA + PP support will be pursued in follow-up PRs.

---

### Step 1 — Bring up the Ray cluster across nodes

`--distributed-executor-backend ray` in Step 3 requires a Ray cluster that spans both nodes. Run `ray stop` first on every node to clean up any stale processes, then start the head on one node and join the rest as workers. Pass `--node-ip-address` explicitly on every node so Ray binds to the routable NIC (not the loopback), and pass `--num-gpus` so all local GPUs are registered.

```bash
# Head node — clean up, then start the head:
ray stop  # clear any stale processes
ray start --head \
    --node-ip-address=<head_node_ip> \
    --port=<head_port> \
    --num-gpus=<gpus_per_node>

# Worker node(s) — clean up, then join the head:
ray stop
ray start \
    --address='<head_node_ip>:<head_port>' \
    --node-ip-address=<this_node_ip> \
    --num-gpus=<gpus_per_node>
```

Once all nodes have joined, run `ray status` on the head node and confirm the total GPU count equals `--tensor-parallel-size`.

### Step 2 — Start one LMCache server on every node

Run before launching vLLM:

```bash
# Run on each node (both Ray head and worker nodes):
lmcache server --host 0.0.0.0 --port <port> --l1-size-gb <size> --eviction-policy LRU
```

> **Important:** Every node's server must bind to `0.0.0.0` (not `localhost` or `127.0.0.1`). Ray worker processes communicate with servers on remote nodes via external IPs; `0.0.0.0` ensures the server accepts both local and cross-node connections.

### Step 3 — Launch vLLM

```bash
vllm serve <model_path> \
    --served-model-name <model_name> \
    --tensor-parallel-size 16 \
    --distributed-executor-backend ray \
    --trust-remote-code \
    --kv-transfer-config '{
        "kv_connector": "LMCacheMPConnector",
        "kv_role": "kv_both",
        "kv_connector_module_path": "lmcache.integration.vllm.lmcache_mp_connector",
        "kv_connector_extra_config": {
            "lmcache.mp.server_urls": [
                "tcp://<node0_ip>:<port>",
                "tcp://<node1_ip>:<port>"
            ],
        }
    }'
```

The new config key in `kv_connector_extra_config`:

- **`lmcache.mp.server_urls`** (`list[str]` or comma-separated string, e.g. `"tcp://host0:6667,tcp://host1:6667"`): URL of every LMCache server, one per node. Takes priority over the legacy `lmcache.mp.host` / `lmcache.mp.port` keys.

---

## Tested configuration

- Hardware: 2 nodes × 8 × H20
- Model: DeepSeek-V2-Lite (MLA), TP=16, 2 LMCache servers


maobaolong:
@Lyj1007 Will take a look this week.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -649,11 +714,13 @@ def wait_for_save(self):
...
-        # In MLA scenario, only the first rank of the pipeline group
-        # needs to save the KV cache.
+        # In MLA scenario, KV cache is not sharded across TP ranks, so only
+        # one rank per node needs to STORE to the local LMCache server.
+        # For single-server deployments this degenerates to rank 0 only.
+        # is_first_rank_of_node handles both cases internally.
         if (
             self.worker_adapter.use_mla
-            and not self.worker_adapter.is_first_rank_of_pp_group
+            and not self.worker_adapter.is_first_rank_of_node
```
chunxiaozheng: @Lyj1007 Thanks for your contribution! I have a question about here: when `pp > 1`, the id of the worker may be not the first rank of the node, but should save the kv cache, in your changes, this may be not work?

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -1199,4 +1268,4 @@ def _cleanup_request_tracker(self, request_id: str) -> None:
             logger.debug(
                 "[KVConnector] Cleaned up request_tracker for request %s",
                 request_id,
-            )
```
chunxiaozheng, line 1284: It seems like nothing has been changed this line, could we go back.

chunxiaozheng:
@Lyj1007 the code quality check run failed, the `format.sh` can help you fix this.

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Thanks a lot for the careful review and for raising this concern! 🙏

The multi-server path here is **intentionally TP-only** — `TP=16` across two nodes is the main use case we're targeting. Properly supporting **PP / DP** would also require a custom **per-GPU rank-grouping API**, so I'd prefer to address it thoroughly in a follow-up PR rather than rush it into this one. Hope that sounds reasonable.

### Safety on the current PR

- `is_first_rank_of_node` falls back to `is_first_rank_of_pp_group` when **`n_servers <= 1`**, so **single-server deployments behave exactly as before**.
- I've also just pushed a **new commit** that adds an explicit guard in `__init__`:
  - raises a `ValueError` when **`n_servers > 1`** is combined with **`pp_size > 1`** or **`dp_size > 1`**
  - so any such misconfiguration **fails fast at startup**, instead of silently storing the wrong KV.

Clipboard_Screenshot_1778828712

Clipboard_Screenshot_1778830193

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Good catch — reverted in the latest commit, thanks!

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -344,73 +356,126 @@ def __init__(
...
+        #   kept for debugging and consistency reporting.
+        #
+        # Single-server compatibility: when len(server_urls) == 1, min() over
+        # a single value is a no-op, so the aggregated result is identical to
+        # the legacy single-server behavior -- this class is a drop-in
+        # replacement for the original single-server lookup.
+
+        # One worker thread per server so all lookups can be fired off at once.
+        # Standard
+        from concurrent.futures import ThreadPoolExecutor
```
maobaolong: This import can be placed to the head of this file.

maobaolong:
@Lyj1007 I'm not sure the whether the following case this PR can handle well.

We assume that there are two node, let's say node0, node1.
- node0 contains chunk[0..4]
- node1 contains chunk[0..2]

Firstly, the scheduler lookup the two nodes and collect the hit result.

But, scheduler only use `min(node0_hit_count, node1_hit_count)` as the hit count eventually.

So, the vllm worker can only retrieve the chunk[0..2] from the two nodes, but never touch the chunk[3..4] in the node0.

Do you found there is an issue of leak, the chunk[3..4] of node0 are locked during handling `lookup` request, but never unlock.


There are two possible approaches,
- A, when hit count from two nodes are mismatch, send a `FREE_LOOKUP_LOCKS` to the target node to reduce lock count at once.
- B(recommend) `free_lookup_locks` always pass the actual hit count rather than `min_chunks`.

Lyj1007:
Hi @maobaolong, thanks for the careful review — you're right, this is a real leak. I've fixed `free_lookup_locks` along the lines of your option B:

1. The key sent to each server now uses the same `start` but a per-server `end = start + hit_chunks * self.chunk_size`, where `hit_chunks` comes from `self._per_server_hits[request_id]`, instead of one shared `min`-length range. So in your node0/node1 example, node0 frees 5 chunks worth and node1 frees 3, and locks balance per node.
2. The iteration target is `_per_server_hits` keys rather than `healthy_urls()`. A server that was healthy at LOOKUP time has already taken locks, so we must still try to free them even if heartbeat marks it unhealthy in between.
3. Removed the top-level `if not self.is_healthy: return` early exit. Each server is tried individually; a single failure is logged and left to server-side GC, instead of dropping the whole free.

**Fallback:** when `_per_server_hits` has no entry (e.g. LOOKUP didn't fully register), we still send to every known server using `(end - start) // self.chunk_size` as a best-effort length.

PTAL, thanks!

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks for the suggestion — `ThreadPoolExecutor` is now imported at the module header with the other stdlib imports.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -119,29 +119,42 @@ def extract_world_size_and_kv_rank(
...
-    server_url: str,
+    server_urls: list[str],
     zmq_context: zmq.Context,
     vllm_config: VllmConfig,
     mq_timeout: float,
     heartbeat_interval: float,
 ) -> LMCacheMPSchedulerAdapter:
-    world_size, kv_rank = extract_world_size_and_kv_rank(
+    n_servers = len(server_urls)
+    actual_world_size = vllm_config.parallel_config.world_size
```
sammshen: nit: can we use just `world_size` or `global_world_size` to distinguish here? and then `local_tp_size` instead of sub

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -119,29 +119,42 @@ def extract_world_size_and_kv_rank(
...
+    actual_rank = vllm_config.parallel_config.rank
+    tp_size = vllm_config.parallel_config.tensor_parallel_size
+    sub_world_size = actual_world_size // n_servers
+    tp_size_per_node = tp_size // n_servers
+    assert tp_size % n_servers == 0, (
+        f"tp_size ({tp_size}) must be divisible by n_servers ({n_servers})"
+    )
+
+    kv_world_size, kv_rank = extract_world_size_and_kv_rank(
```
sammshen: is this rank global or local? please be explicit

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Thanks for the suggestion! Renamed in [`a573f8b`](https://github.com/LMCache/LMCache/pull/3248/commits/a573f8b):

- `actual_world_size` → `global_world_size`
- `actual_rank` → `global_rank`
- `tp_size` → `global_tp_size`
- `sub_world_size` → `local_world_size`
- `tp_size_per_node` → `local_tp_size`

The convention used in this file: `global_*` is the cluster-wide value
(e.g. `global_tp_size = 16` for TP=16 across 2 nodes), `local_*` is the
per-node value (`local_tp_size = 8`).

A note on scope: the rename above only touches the **local variables**
inside `create_scheduler_adapter` / `create_worker_adapter`. The
`ParallelStrategy.tp_size` **dataclass field** is intentionally left
unchanged — it already represents the cluster-wide TP size, and renaming
the field would ripple into every `self.parallel_strategy.tp_size` reader
across the codebase, which is outside the scope of this review.

Let me know if any of the names still feel off.

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Thanks for pointing this out. `parallel_config.rank` is the **global**
rank (0 .. `world_size - 1` across all workers, not per-node). I've
renamed the variable to `global_rank` and added a short comment above
the `extract_world_size_and_kv_rank(...)` call noting that it expects
the global `world_size` / `rank`. Updated in [`a573f8b`](https://github.com/LMCache/LMCache/pull/3248/commits/a573f8b).

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -119,28 +119,50 @@ def extract_world_size_and_kv_rank(
...
+    global_rank = vllm_config.parallel_config.rank
+    global_tp_size = vllm_config.parallel_config.tensor_parallel_size
+
+    # Per-node (local) derivations: each node hosts one LMCache server.
+    if global_tp_size % n_servers != 0:
+        raise ValueError(
+            f"tp_size ({global_tp_size}) must be divisible by n_servers ({n_servers})"
+        )
+    local_world_size = global_world_size // n_servers
+    local_tp_size = global_tp_size // n_servers
```
ApostaC: Quick question: does this only for TP? What if someone is running PP?

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -150,28 +172,53 @@ def create_scheduler_adapter(
...
+        use_mla=mla_enabled(vllm_config.model_config),
+        kv_world_size=kv_world_size,
+        kv_worker_id=kv_rank,
+        global_world_size=global_world_size,
+        global_rank=global_rank,
+        tp_size=global_tp_size,
+        pp_size=vllm_config.parallel_config.pipeline_parallel_size,
+        kv_local_world_size=local_world_size,
+        local_tp_size=local_tp_size,
+        n_servers=n_servers,
```
ApostaC: This piece of the code is duplicated. Consider have a helper function that takes in `vllm_config` and produces `ParallelStrategy`, and reuse the function

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -225,31 +226,41 @@ class ParallelStrategy:
...
 
     kv_worker_id: int
     """
-    The kv worker id of the sub-process, kv_worker_id may not be equal to the 
-    actual_worker_id, in the case of mla, it will 'exclude' the effect of TP, 
-    the value is calculated by `extract_world_size_and_kv_rank` in 
-    `lmcache_mp_connector.py`.
+    The kv worker id of the sub-process. May differ from ``global_rank``: in
+    MLA mode it 'excludes' the effect of TP. Computed by
+    ``extract_world_size_and_kv_rank`` in ``lmcache_mp_connector.py``.
```
ApostaC: nit: since we are updating the docstring here, let's make it docstring clearer by saying something like:
```
The kv_world_size indicates how many different pieces the KV cache is split into.
The kv_worker_id indicates which piece of the KV cache the current worker has.
```

In the meantime, I want to know what's the difference between `kv_world_size` and `kv_local_world_size`. In what case, `kv_local_world_size` will be used?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -457,76 +471,129 @@ def __init__(
...
+        if not self.parallel_strategy.use_mla and self.parallel_strategy.n_servers > 1:
+            return self.parallel_strategy.kv_local_world_size
         return self.parallel_strategy.kv_world_size
 
     @property
     def tp_size(self) -> int:
-        """The tensor parallel size."""
+        if self.parallel_strategy.n_servers > 1:
+            return self.parallel_strategy.local_tp_size
         return self.parallel_strategy.tp_size
```
ApostaC: Maybe I'm missing something, but where did we use these two functions?

If they are not used, maybe consider removing them (and the corresponding class members) to keep the code cleaner?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -457,76 +471,129 @@ def __init__(
...
         # - _pending_lookups: request_ids submitted but not yet resolved
-        # - _finished_lookup_results: cached chunk count keyed by request_id,
-        #   so that repeated calls to check_lookup_result return the same value
-        #   even after the server has already popped the job (exactly-once).
+        #   (resolved == results from ALL servers have been merged).
+        # - _finished_lookup_results: aggregated chunk count keyed by
+        #   request_id, computed as the min hit across servers so a chunk
+        #   only counts as "hit" when every server has it. Cached so that
+        #   repeated calls to check_lookup_result return the same value
+        #   even after the servers have popped the job (exactly-once).
```
ApostaC: nit: not sure why it's not removed, but the current semantics are not "exactly once" anymore.

```suggestion
        #   even after the servers have popped the job. The cached result 
        #   will be popped by `cleanup_lookup_result`.
```

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -457,76 +471,129 @@ def __init__(
...
+        # - _per_server_hits: per-server raw hit counts keyed by request_id,
+        #   kept for debugging and consistency reporting.
```
ApostaC: nit: What's the key and the value of `_per_server_hits`? Please make it clear here

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -457,76 +471,129 @@ def __init__(
...
+        #
+        # Single-server compatibility: when len(server_urls) == 1, min() over
+        # a single value is a no-op, so the aggregated result is identical to
+        # the legacy single-server behavior -- this class is a drop-in
+        # replacement for the original single-server lookup.
```
ApostaC: nit: the comment here is not needed

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -562,6 +629,11 @@ def maybe_submit_lookup_request(
         self._ensure_heartbeat_started()
 
         if not self.is_healthy:
+            logger.warning(
+                "Skip LOOKUP for req=%s because not all servers are healthy: %s",
+                request_id,
+                {u: ev.is_set() for u, ev in self._health_events.items()},
+            )
```
ApostaC: Let's not have this log here. This is because this function will be called many times, and we don't want the log to be flooded by this message.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -578,21 +650,43 @@ def maybe_submit_lookup_request(
...
+                self._health_events[url].clear()
+                return url, False
+            except Exception as e:
+                logger.error("LOOKUP to %s failed: %s", url, e, exc_info=True)
+                self._health_events[url].clear()
+                return url, False
+
+        # Fan out in parallel; total latency ~= slowest server, not sum.
+        results = list(self._executor.map(_submit_one, self._server_urls))
+
```
ApostaC: The code is introducing unnecessary complexity here. We don't need a thread pool to submit the lookup at all, because
1. the submission itself is a non-blocking operation. Once the `LOOKUP` task is submitted, it will run simultaneously on 2 LMCache servers.
2. The execution `LOOKUP` itself is non-blocking as well, which means `future.result` here will not block for more than 1 or 2 milliseconds. And when we get the result from one LMCache server, another result will be back very soon.

Please remove the thread pool, thanks!

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -457,76 +471,129 @@ def __init__(
...
+        # a single value is a no-op, so the aggregated result is identical to
+        # the legacy single-server behavior -- this class is a drop-in
+        # replacement for the original single-server lookup.
+
+        # One worker thread per server so all lookups can be fired off at once.
+
+        self._executor = ThreadPoolExecutor(
+            max_workers=len(self._server_urls),
+            thread_name_prefix="lmcache-mp-lookup",
+        )
```
ApostaC: As I stated in another comment, we don't need a thread pool here.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -625,26 +719,50 @@ def check_lookup_result(self, request_id: str) -> int | None:
...
+
+        results = list(self._executor.map(_query_one, self._server_urls))
 
-        if result is None:
-            return None
+        per_server: dict[str, int] = {}
+        for url, r in results:
+            if r is None:
+                return None
+            per_server[url] = int(r)
```
ApostaC: Same as the above comment, we don't need a thread pool here since `QUERY_PREFETCH_STATUS` is also a non-blocking operation.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -691,21 +819,45 @@ def free_lookup_locks(
             request_id: The request ID.
             cache_salt: Per-user isolation salt.
         """
-        if not self.is_healthy:
-            return
```
ApostaC, line 809: Why is this removed?

lmcache/integration/vllm/vllm_multi_process_adapter.py
ApostaC: Additionally, if I remember correctly, `QUERY_PREFETCH_STATUS` has "exactly-once" semantics. Which means what we should have here is:
```suggestion
        if request_id is not in self._per_server_hits:
            self._per_server_hits[request_id] = {}
        per_server = self._per_server_hits.get(request_id)
        for url, r in results:
            if r is None:
                continue
            if per_server.get(url) is not None:
                continue
            per_server[url] = int(r)
        all_finished = ... # checks whether all the servers have returned the result
        if not all_finished:
            return None    
```

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -691,21 +819,45 @@ def free_lookup_locks(
...
+            per_server_end = start + hit_chunks * self.chunk_size
+            per_server_end = min(per_server_end, len(token_ids))
+
+            key = self._create_key(
+                token_ids=token_ids,
+                start=start,
+                end=per_server_end,
+                request_id=request_id,
+                cache_salt=cache_salt,
+            ).no_worker_id_version()
```
ApostaC: The logic here seems to be incorrect.

The `free_lookup_locks` was designed to free the locks on the vLLM's GPU-cached part, which means we always have `start = 0` and `end = aligned_vllm_gpu_cached_tokens`.
We should submit the `FREE_LOOKUP_LOCKS` to all servers with this start and end.

Additionally, in the case that some servers looked up more KV cache than others, we should submit additional `FREE_LOOKUP_LOCKS` with `start = min_num_lookuped_tokens` and `end = lookuped_tokens_for_current_server`.

ApostaC (changes requested):
Thanks for the contribution! 

High-level points:
1. The thread pool is not needed
2. Logic issue in free_lookup_locks

Please see the detailed comments below

ApostaC:
Additionally, is it possible to have some CI tests to emulate the 2-node case on a single machine with 2 GPUs?

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you. Updated in `aa0a2f0`. The `Lookup state tracking` block now describes `_per_server_hits` as the cross-poll accumulator (matching how `check_lookup_result` uses it after this commit), drops the obsolete "exactly-once" wording, and notes that it's cleared by `cleanup_lookup_result`.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks for the suggestion — applied in `aa0a2f0`. `_per_server_hits` is now filled in incrementally:

- Each call to `check_lookup_result` only queries servers not yet present in `_per_server_hits[request_id]`, so we never hit the server-side job-already-popped branch.
- `None` from `_query_one` (still in progress or a transient error) is skipped and retried on the next poll; the lookup completes only when every server has contributed an entry.
- `min` aggregation, the cached token count, and the `_pending_lookups` discard now happen exactly once on that transition.

Also changed the timeout/exception branches in `_query_one` from `return url, 0` to `return url, None`, so an unreachable server doesn't falsely advance the request to "complete with zero hits".

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you. Removed in `aa0a2f0`

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -691,11 +684,14 @@ def _submit_one(url: str) -> tuple[str, bool]:
...
+        or ``None`` if the lookup is still in progress on at least one
+        server.
 
-        Sends a QUERY_PREFETCH_STATUS request to the server and blocks
-        until the server responds.  Returns the matched token count
-        when the prefetch is complete, or None if still in progress.
+        On the first poll where every server has reported, the aggregate
+        is written to ``_finished_lookup_results`` and returned. Repeat
+        calls return the cached aggregate verbatim until
+        ``cleanup_lookup_result`` is invoked.
```
ApostaC: nit: at a high level, the docstring needs to describe the behavior with "public interfaces". We don't need to describe implementation details in the docstring because it may cause inconsistencies in the future.

I recommend keeping something similar to the old docstring here

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you for the careful explanation — you're absolutely right, and
I appreciate you taking the time to spell out the reasoning. Pushed
the fix in `ae4fd0df`.

`maybe_submit_lookup_request` now submits to all servers in a plain
loop and drains the futures in a second pass. Same for
`check_lookup_result`. `self._executor`, the `concurrent.futures`
import, and `executor.shutdown(wait=True)` in `shutdown()` are gone.

Behavior unchanged: per-server timeout/error still clears the health
event, `_per_server_hits` accumulation and the `min()` aggregation
are the same.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks, removed in `ae4fd0df`.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you — addressed together with the LOOKUP path in `ae4fd0df`.
`check_lookup_result` now uses the same submit-then-wait loop, and
the `_per_server_hits` accumulator across polls is unchanged.

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Thanks for raising this! Yes, the multi-server path in this PR is intentionally TP-only. 

I've added an explicit safeguard (raising a `ValueError` when `n_servers > 1` and `pp_size > 1` or `dp_size > 1`) so users won't accidentally run into unsupported configurations. Single-server mode continues to support PP/DP exactly as before.
Clipboard_Screenshot_1780045463

The reason for deferring PP/DP is that while the current logic naturally handles cases where PP/DP boundaries align with node boundaries, it gets tricky when a PP stage or DP replica straddles across nodes. Handling this correctly requires a generalized rank-grouping API and coordinated changes with the MLA storage logic to ensure a coherent keyspace. 

I'd like to land PP/DP support in a dedicated follow-up PR so we can design the abstraction carefully and keep this PR focused. Happy to file a tracking issue for this if you think it would be helpful!

Lyj1007:
> Thanks for the contribution!
> 
> High-level points:
> 
> 1. The thread pool is not needed
> 2. Logic issue in free_lookup_locks
> 
> Please see the detailed comments below

Thanks for the review! Both top-level points have been addressed in the latest revision:

1. **Thread pool removed**: `LOOKUP` and `QUERY_PREFETCH_STATUS` now submit to all servers inline and only block on the resulting futures. Since the submission is non-blocking, the wait time is bounded by the slowest server anyway.
2. **`free_lookup_locks` logic fixed**: It now strictly follows the original docstring, emitting one `FREE_LOOKUP_LOCKS` per server with the caller-supplied `[start, end)` without implicit trimming. The logic for releasing per-server over-hit tails has been moved to a dedicated `free_per_server_overhit_locks` method, which is invoked from `update_state_after_alloc` before `_per_server_hits` is discarded.

Detailed responses inline.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you for the suggestion! I have adopted the wording you proposed for `kv_world_size` and `kv_worker_id`.

To clarify the naming convention and address your question regarding `local_kv_world_size`, I have organized the prefixes to better reflect their specific scopes:
*   **`global_`**: Refers to the overall vLLM deployment (e.g., the total number of worker processes across all nodes).
*   **`local_`**: Refers to the scope of a single LMCache server (since LMCache servers are deployed per-node when `n_servers > 1`).
*   **`kv_`**: Refers to how the KV cache chunks are split and handled among workers.

The `local_*` fields (such as `local_kv_world_size`) are only consulted when `n_servers > 1`. They describe the same quantity as their global counterparts but are scoped to a single LMCache server. This ensures that each node's keyspace remains self-contained and that two nodes never collide on the same key.

Additionally, I have renamed `kv_local_world_size` to `local_kv_world_size` for symmetry with `local_tp_size`, and reordered the dataclass fields (deployment shape → model property → global coords → KV coords → local coords) so the relationships are obvious top-to-bottom.

Here is the updated dataclass:

```python
class ParallelStrategy:
    """vLLM parallelism layout snapshot consumed by the LMCache adapters."""

    use_mla: bool
    """Whether the model uses MLA."""

    global_world_size: int
    """Total number of vLLM worker processes."""

    global_rank: int
    """The current worker's rank in ``[0, global_world_size)``."""

    tp_size: int
    """Tensor-parallel size of the model."""

    pp_size: int
    """Pipeline-parallel size of the model."""

    n_servers: int
    """Number of LMCache servers backing this deployment
    (1 = single shared server, >1 = one server per node)."""

    kv_world_size: int
    """How many different pieces the KV cache is split into across the
    whole deployment."""

    kv_worker_id: int
    """Which piece of the KV cache the current worker has, in
    ``[0, kv_world_size)``."""

    local_kv_world_size: int
    """Same as ``kv_world_size`` but scoped to a single LMCache server;
    consulted only when ``n_servers > 1``."""

    local_tp_size: int
    """Same as ``tp_size`` but scoped to a single LMCache server;
    consulted only when ``n_servers > 1``."""
```

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you for the careful review! Both properties are indeed actively used, and they play an important role in keeping the scheduler and worker sides aligned. Let me walk through the details.

**Where they are used (in `lmcache/integration/vllm/vllm_multi_process_adapter.py`):**

- **`tp_size`** — included in the payload of every `LOOKUP` and `FREE_LOOKUP_LOCKS` request, so the server knows how many worker shards a given key spans (consumed by `compute_extra_count` on the server side for MLA multi-reader locking). Read sites: `maybe_submit_lookup_request`, `free_lookup_locks`, `free_per_server_overhit_locks`.
- **`world_size`** — passed into every `IPCCacheEngineKey` via `_create_key` on both the scheduler and the worker side, so lookup keys and store/retrieve keys mint from exactly the same keyspace. The worker also forwards it through `_send_register_kv_caches_request` at registration time.

**Why properties rather than direct field access:** the names `world_size` / `tp_size` were chosen to stay consistent with the variable names already used on the server side and in the original single-server adapter (which read these fields directly). The `kv_` prefix was introduced earlier to disambiguate the MLA case, and the `local_` prefix in this PR to handle the multi-server case. Wrapping them in properties keeps all call sites untouched and confines both prefix-aware switches to one place — `local_*` is consulted only when `n_servers > 1`, so callers never need to scatter `if n_servers > 1` / `if use_mla` checks.

**One follow-up I'd appreciate your thoughts on:** these four fields (`kv_world_size`, `kv_worker_id`, `local_kv_world_size`, `local_tp_size`) are only consumed by the `world_size` / `worker_id` / `tp_size` properties on the adapters — there are no other readers in the codebase. So in principle we could fold the `extract_world_size_and_kv_rank` MLA adjustment and the `local_world_size = global_world_size // n_servers` derivation directly into those properties, and slim `ParallelStrategy` down to a pure vLLM-layout snapshot below.
```python
@dataclass
class ParallelStrategy:
    use_mla: bool
    global_world_size: int
    global_rank: int
    tp_size: int   
    pp_size: int
    n_servers: int
```
I'd be happy to do this in a follow-up if you feel the slimmer struct would be cleaner — please let me know which direction you would prefer.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you for catching this! The comment now spells out the shape explicitly:

```python
# - _per_server_hits: {request_id: {server_url: hit_chunks}}.
```

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you for pointing this out! The `logger.warning` call has been removed from the unhealthy short-circuit in `maybe_submit_lookup_request`, so the log will no longer be flooded when the server is unhealthy.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Apologies for the confusion — that removal stemmed from a misunderstanding of `free_lookup_locks`'s contract on my side in the previous revision (I had conflated its semantics with the per-server tail cleanup, which led me to drop the `is_healthy` short-circuit by mistake). The check has been restored, so `free_lookup_locks` once again returns early with `if not self.is_healthy: return` before issuing any `FREE_LOOKUP_LOCKS` requests, matching the original behavior. Thank you for catching this!

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Apologies for the broken contract in the previous revision — I had conflated the GPU-cached cleanup path with the per-server tail cleanup, which caused `free_lookup_locks` to drift from the single-server semantics you originally designed. This has now been fixed exactly along the lines you suggested, split into two methods so each one carries a single, crisp contract:

- **`free_lookup_locks(token_ids, start, end, request_id, cache_salt)`** — restored to its original meaning. It just emits one `FREE_LOOKUP_LOCKS` per server with the caller-supplied `[start, end)`, with no implicit per-server trimming. The connector keeps calling this with `start = 0, end = aligned_vllm_gpu_cached_tokens` to release locks on the GPU-cached part, exactly as before.

- **`free_per_server_overhit_locks(token_ids, request_id, cache_salt)`** — a new, dedicated method for the "some servers hit more than others" case. It reads `_per_server_hits[request_id]`, computes `min_hit_end = min(hit_chunks) * chunk_size`, and for every server whose hit exceeds this minimum, sends a `FREE_LOOKUP_LOCKS` on its own tail `[min_hit_end, this_server_hit_end)`.

The connector now invokes `free_per_server_overhit_locks` from `update_state_after_alloc` **before** `cleanup_lookup_result` discards `_per_server_hits`, then proceeds with the regular `free_lookup_locks(start=0, end=aligned_vllm_gpu_cached_tokens)` call. This way the over-hit tails are released on every request — no leak even when servers disagree — and `free_lookup_locks` itself stays a clean single-contract primitive. Thank you for the careful guidance on the correct semantics here!

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thank you for the guidance — you are absolutely right that the docstring should describe the public contract rather than implementation details. The `check_lookup_result` docstring has been reverted to a contract-first style close to the original wording, with the `_finished_lookup_results` / per-server-polling paragraphs removed.

I also took this as a broader principle and went through every other docstring in `vllm_multi_process_adapter.py`.

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Thanks for the catch! You're right — the `ParallelStrategy` construction was duplicated between `create_scheduler_adapter` and `create_worker_adapter`.

I've extracted it into a private helper `_build_parallel_strategy(server_urls, vllm_config)` and reused it in both factories. Pure refactor, no behavioral change. Pushed in the latest commit — please take another look. 🙏

lmcache/integration/vllm/lmcache_mp_connector.py
ApostaC: Please revert the part about DP. Each DP is naturally supported by MP mode because each DP instance will have its own scheduler, which will unlikely be cross-node shared.

lmcache/integration/vllm/vllm_multi_process_adapter.py
ApostaC: `Same as kv_world_size` and `Same as tp_size` is a bit confusing here. Does it mean always the same value? Or you are saying the values are equal only when `n_servers = 1`. 

For example when tp = 16 and n_servers = 2, what would we got for `local_kv_world_size` and `local_tp_size`?

ApostaC: To make it clear, it would be great if the docstring could let people have an expectation of what the value of this field will be without looking at the real logic in the code.

lmcache/integration/vllm/vllm_multi_process_adapter.py
ApostaC: 1. If the `world_size` is going to be passed to `IPCCacheEngineKey`. In this case, we can distinguish between the KV cache from a `tp 16 + 2 servers` instance and a `tp 8 + 1 server` instance.
2. I think the proposal for the new parallel strategy makes sense. But I would love to have it in a separate PR.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -578,30 +624,60 @@ def maybe_submit_lookup_request(
...
+        for url in self._server_urls:
+            try:
+                futures[url] = send_lmcache_request(
+                    self.mq_clients[url],
+                    RequestType.LOOKUP,
+                    [key, self.tp_size],
+                )
+            except Exception as e:
+                logger.error("LOOKUP submit to %s failed: %s", url, e, exc_info=True)
+                self._health_events[url].clear()
```
ApostaC: Why do we have a try-catch block for send_lmcache_result? In what case will it raise an exception?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -578,30 +624,60 @@ def maybe_submit_lookup_request(
...
+                    "LOOKUP to %s timed out after %ss; marking unhealthy.",
+                    url,
+                    self._mq_timeout,
+                )
+                self._health_events[url].clear()
+                results.append((url, False))
+            except Exception as e:
+                logger.error("LOOKUP to %s failed: %s", url, e, exc_info=True)
+                self._health_events[url].clear()
+                results.append((url, False))
```
ApostaC: What are the other exceptions raised by `fut.result`?
If no other exceptions, we don't need an extra except block here.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -578,30 +624,60 @@ def maybe_submit_lookup_request(
...
+
+        for url in self._server_urls:
+            if url not in futures:
+                results.append((url, False))
```
ApostaC: Seems like this block is not needed.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -578,30 +624,60 @@ def maybe_submit_lookup_request(
...
+
+        # Only track as pending when every server accepted the job.
+        if all(ok for _, ok in results):
+            self._pending_lookups.add(request_id)
+        else:
+            failed = [u for u, ok in results if not ok]
+            logger.error(
+                "[req=%s] LOOKUP failed on servers %s -- fall back to no-hit",
+                request_id,
+                failed,
```
ApostaC: In a "partial failure" case, we need to remember which servers have successful lookups, and we need to call `FREE_LOOKUP_LOCKS` for those servers later (or immediately) to prevent memory leaks. 
Just a reminder to check whether we have done this.

ApostaC (commented):
We need to be super careful with new logics related to vLLM integrations. Additionally, the logic in the scheduler & worker adapters should be super simple so that people can tell the high-level stuff without looking at the code.

I feel like the main part of this PR is okay, but it has some AI-generated noises inside -- therefore, I'm a little bit hesitant to merge this.

My suggestion is that we break down this PR into two pieces:
1. A refactor on the `ParallelStrategy` to prepare for support of multi-node tp case
2. The real implementation of sending the lookup and free locks to both servers. It would be great if you could write down the logic changes **_yourself_** before writing the code.

Lyj1007:
> We need to be super careful with new logics related to vLLM integrations. Additionally, the logic in the scheduler & worker adapters should be super simple so that people can tell the high-level stuff without looking at the code.
> 
> I feel like the main part of this PR is okay, but it has some AI-generated noises inside -- therefore, I'm a little bit hesitant to merge this.
> 
> My suggestion is that we break down this PR into two pieces:
> 
> 1. A refactor on the `ParallelStrategy` to prepare for support of multi-node tp case
> 2. The real implementation of sending the lookup and free locks to both servers. It would be great if you could write down the logic changes **_yourself_** before writing the code.

Thanks for the review and the split suggestion. Part A (`ParallelStrategy` refactor for multi-node TP) is done in #3478 — refactor only, no behavior change for existing paths.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks! I've removed the unnecessary `except Exception as e` to align with the original code.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks! I've removed the unnecessary `except Exception as e` to align with the original code.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks! I've removed the unnecessary code.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: This has been updated in the latest version. For partial hit cases, we release the locks on the hitting servers via `FREE_LOOKUP_LOCKS` immediately upon detecting inconsistency. Those entries are then considered as not hit globally, which triggers `store` and follow-up steps.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -511,23 +522,81 @@ def __init__(
...
+        # reader counts to vary per (server, pp_stage) pair and break
+        # the single-``tp_size`` LOOKUP / FREE_LOOKUP_LOCKS protocol.
+        if n_servers > 1 and mla_enabled(vllm_config.model_config):
+            pp_size = vllm_config.parallel_config.pipeline_parallel_size
+            if pp_size > 1:
+                raise ValueError(
+                    "LMCacheMPConnector multi-server MLA mode only supports "
+                    "tensor parallelism (TP), not pipeline parallelism (PP). "
+                    f"Got pp_size={pp_size}."
+                )
```
ApostaC, line 577: Regarding this, is `mla_enabled` a hard requirement? IIUC, no matter whether it's MLA, we don't support PP right now, do we?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -728,36 +775,82 @@ def check_lookup_result(self, request_id: str) -> int | None:
...
+                        start=min_chunks * self.lmcache_tokens_per_chunk,
+                        end=tail_end,
+                        request_id=request_id,
+                        cache_salt=cs or "",
+                    ).no_worker_id_version()
+                    send_lmcache_request(
+                        self.mq_clients[url],
+                        RequestType.FREE_LOOKUP_LOCKS,
+                        [tail_key, self.tp_size],
+                    )
```
ApostaC: nit: we can have a helper function for this part (something like `_free_inconsistent_lookup_locks`)

ApostaC (commented):
Much clearer this time. Please take a look at my comments below.

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: The guard is intentionally scoped to MLA — non-MLA already supports PP.

For non-MLA, both TP and PP shard the KV cache, so every worker still maps to a unique slot and the multi-server case behaves the same as single-server — no extra handling needed.

For MLA, PP splits layers across nodes, causing per-piece reader counts to vary per (server, pp_stage) pair and breaking the single-`tp_size` LOOKUP / FREE_LOOKUP_LOCKS protocol. Support for the case where PP groups align with server node boundaries will be added as a follow-up.

The updated comment in the code now also clarifies that non-MLA mode is not affected by this restriction.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Done — extracted into `_free_inconsistent_lookup_locks(request_id, per_server, min_chunks)` and replaced the inline block with a single call.

lmcache/integration/vllm/lmcache_mp_connector.py
ApostaC: Sorry I don't quite get it. My understanding is for both MLA and non-MLA, the following are true:
- A different server will have its own pp_stage
- Different pp_stage will have different KV caches

I'm not sure i get the difference you said here

Lyj1007: Thanks for the question — I've updated the PR description with the following table to make the model explicit. Let me re-state the reasoning on top of it:

| Scenario | kv_world_size | kv_tp_size | kv_worker_id | is_kv_writer |
|---|---|---|---|---|
| Single-server, MLA              | `vllm_world_size // tp_size = pp_size` | `tp_size`     | `vllm_worker_id // tp_size`              | `vllm_worker_id % tp_size == 0` |
| Single-server, non-MLA          | `vllm_world_size`                      | `tp_size`     | `vllm_worker_id`                         | always True |
| Multi-server, MLA, only TP      | `vllm_world_size // tp_size = pp_size` | `tp_size / N` | `vllm_worker_id // tp_size`              | `vllm_worker_id % (tp_size / N) == 0` |
| Multi-server, non-MLA, only TP  | `vllm_world_size / N`                  | `tp_size / N` | `vllm_worker_id % (vllm_world_size / N)` | always True |
| Multi-server, MLA, TP and PP    | not supported yet                      |               |                                          | |
| Multi-server, non-MLA, TP and PP| `vllm_world_size / N`                  | `tp_size / N` | `vllm_worker_id % (vllm_world_size / N)` | always True |

You're right that in **both** MLA and non-MLA cases, different PP stages own different KV caches. The asymmetry I'm describing is not about "whether PP shards KV" — it's about **how PP and the server-grouping interact when projecting `(vllm_world_size, vllm_worker_id)` into the per-server `(kv_world_size, kv_worker_id)` view**.

### Non-MLA: single-server is just the `N = 1` degenerate case

For non-MLA, the formulas are *literally* the same in every row, with `N = n_servers`:

- `kv_world_size = vllm_world_size / N`
- `kv_worker_id  = vllm_worker_id % (vllm_world_size / N)`

When `N = 1` this collapses to `kv_world_size = vllm_world_size`, `kv_worker_id = vllm_worker_id`, which is exactly the single-server row. So single-server and multi-server **share one code path** — there is no special-case logic, regardless of whether PP is on or off. Each server simply handles a contiguous slice of `vllm_worker_id` space, and within a slice every rank is its own KV writer.

### MLA: only the "PP-aligned-with-servers" topology generalizes cleanly

For MLA, TP ranks **share** a KV shard (TP doesn't shard KV), so the natural shard count is `pp_size`, not `vllm_world_size`. The clean multi-server generalization of the single-server MLA row would be: **one server per PP stage**, i.e. `n_servers = pp_size`. In that topology the same projection rules carry over and everything stays symmetric with the non-MLA story.

What this PR actually supports for MLA, however, is **TP-only across servers** (`pp_size = 1`, TP fanned out to N nodes). That is *not* the PP-aligned topology — it's the opposite: a single PP "group" gets split across servers, so each server only sees a TP-fragment of one shared KV shard. That's why:

- `kv_world_size` stays `pp_size` (= 1 here) — the shard count doesn't change just because we split TP across nodes;
- `kv_tp_size` becomes `tp_size / N` — each server only sees its local TP fan-in, which is what `compute_extra_count` uses for the MLA multi-reader lock accounting;
- `is_kv_writer` becomes `vllm_worker_id % (tp_size / N) == 0` — exactly one writer **per server**, instead of one writer globally, because each server holds an independent ObjectKey space and needs its own local writer.

So the MLA single-server row and the MLA multi-server-TP-only row are not the same template with `N` plugged in — the multi-server-TP-only case is fundamentally a "PP group spanning multiple servers" topology and needs the extra per-server-writer logic.

### Why MLA + (TP and PP) is left for a follow-up

The TP-only case I'm landing here is the practically-needed-first variant but is *less* general than that, not *more*. Mixing TP+PP across servers for MLA would require a third projection (PP-stage routing on top of TP-fan-in splitting), which is why it's deferred to a follow-up PR — I want to keep this PR focused on the topology our users are actually deploying today.

TL;DR: for non-MLA, single-server is genuinely just `N = 1` of the multi-server formula, so I claim "unified behavior". For MLA, "unified behavior" would require PP-aligned servers; the TP-only multi-server case in this PR is a deliberately narrower variant, and full TP+PP MLA support is a follow-up.

ApostaC (approved):
LGTM!

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -511,23 +522,82 @@ def __init__(
...
+                "lmcache.mp.host", "tcp://localhost"
+            )
+            server_port = vllm_config.kv_transfer_config.get_from_extra_config(
+                "lmcache.mp.port", 5555
+            )
+            server_urls = [f"{server_host}:{server_port}"]
+
+        # lmcache.mp.n_servers is the authoritative server count.
+        # If provided it must match the length of server_urls.
+        n_servers_cfg = vllm_config.kv_transfer_config.get_from_extra_config(
```
maobaolong: We can remove this redundant config.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -306,13 +306,16 @@ class ParallelStrategy:
...
 
+    n_servers: int
+    """Number of LMCache servers backing this deployment"""
+
     @property
     def kv_world_size(self) -> int:
         """Number of pieces a single token chunk's KV cache is split into
         on the LMCache server storage."""
         if self.use_mla:
             return self.vllm_world_size // self.tp_size
```
maobaolong, line 320: As discussed, although we plan to do not support `pp + tp + mla` within this pr, and you have added a pre-condition check already, but add an extra comment here is essential.

maobaolong (approved):
@Lyj1007 This PR looks good for me overall, just left some nits.

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Thanks for the suggestion — addressed. I removed the redundant `lmcache.mp.n_servers` check and now derive server count directly from `lmcache.mp.server_urls`.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks for the note — addressed. I added an explicit comment here to clarify that this PR does not support `PP + TP + MLA`, and this branch relies on the existing precondition (`pp_size == 1`).

maobaolong:
> Step 3 — Launch vLLM
@Lyj1007 Please update the comment you left here, otherwise, guys may be confusing about the removed `lmcache.mp.n_servers`

PR #3256, opened 2026-05-11
 [CI] Temporary Pin blend test vLLM to cu129 channel to avoid CUDA-13 PyPI wheel

deng451e:
Description

  vLLM's stable PyPI release (0.20.2) recently moved to a CUDA 13 toolchain, while the tensormesh/cacheblend container still ships CUDA 12. Under the previous
  install command, --extra-index-url https://wheels.vllm.ai/nightly/cu129 combined with --index-strategy unsafe-best-match left PyPI as the implicit primary
  index. uv then picked the highest version across indices — PyPI's 0.20.2 (final) outranks the cu129 channel's 0.20.2rc1.devN+...cu129 (pre-release) per PEP
  440 — and installed a wheel linked against libcudart.so.13, which fails to import in the container with ImportError: libcudart.so.13: cannot open shared
  object file.

  This change promotes the cu129 channel to --index-url so vLLM resolves there exclusively, and re-adds PyPI as --extra-index-url so transitive deps still
  resolve. --pre continues to let the nightly rc/dev builds on cu129 qualify.


sammshen (approved):

ApostaC (approved):
LGTM

ApostaC:
adding a full tag to trigger the blend test

PR #3257, opened 2026-05-11
[MP][Observability] Drop inflated L2 store throughput on fast-path

Oasis-Git:
Some L2 adapters (mock, fs, nixl_store) skip the write when a key is already present in the backend.  That fast-path collapses the time between L2_STORE_SUBMITTED and L2_STORE_COMPLETED to near-zero, while the submitted-bytes count carried on SUBMITTED stays unchanged.  The L2 throughput subscriber computes ``bytes / dt`` and the result blows up -- locally I saw mean store throughput at 52.9 GB/s on a mock adapter pinned to 4 GB/s, with bimodal samples in the 100-250 GB/s bucket from duplicate-store batches.

Plumb the bytes actually transferred per task from the adapter through to the subscriber so the histogram reflects real work:

  * ``L2AdapterInterface.pop_completed_store_task_bytes`` (default ``{}``) lets adapters opt-in to per-task transfer reporting.  Adapters without a fast-path keep the current behavior.
  * mock / fs / nixl_store override it and accumulate ``total_bytes`` over actually-written keys in their existing store coroutines.
  * ``SerdeWrapper`` forwards inner reports, preserving "unknown" when the inner adapter doesn't track.
  * ``StoreController`` queries both maps when draining and surfaces ``bytes_transferred`` on ``L2_STORE_COMPLETED`` when known.
  * ``L2ThroughputSubscriber._on_store_completed`` prefers the override when present: ``>0`` records ``bytes_transferred / dt``, ``0`` drops the sample (no useful throughput from a no-op), absent falls back to submitted-bytes / dt.

After the change the local script shows mean store throughput 3.98 GB/s (matching the mock's ``--mock-bandwidth-gb 4``) and the 100-250 GB/s outlier bucket is empty.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/base.py
+++ b/lmcache/v1/distributed/l2_adapters/base.py
@@ -231,6 +231,27 @@ class L2AdapterInterface(ABC):
         """
         pass
 
+    def pop_completed_store_task_bytes(self) -> dict[L2TaskId, int]:
+        """Report bytes actually transferred per completed store task.
+
+        Optional. Default returns ``{}``, which leaves the L2 throughput
+        subscriber to fall back on submitted-bytes accounting.
+
+        Adapters that fast-path duplicate keys (e.g. skip the write when
+        the key already exists in the backend) SHOULD override this so
+        the throughput histogram reflects real work, not skipped no-ops.
+        A task with all keys fast-pathed reports ``0`` here, which the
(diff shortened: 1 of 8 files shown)


deng451e (approved):
LGTM

royyhuang (approved):
LGTM

PR #3259, opened 2026-05-12
feat(mp): Non-GPU Context by pickle

hlin99:
## Motivation

Multiprocess (MP) mode brings significant advantages to LMCache:

- **Process-level decoupling** between vLLM and LMCache — each runs in its own process, improving fault isolation
- **Cleaner error handling** — crashes in the cache layer do not bring down the serving engine
- **Richer observability** — independent metrics collection, health monitoring, and status reporting per component
- **Reduced memory copies** — GPU IPC enables zero-copy KV cache sharing between processes
- **Independent scaling** — cache server resources can be tuned separately from model serving
- **Simplified lifecycle management** — cache server can be restarted without disrupting in-flight inference

However, **MP mode currently only works on CUDA devices**. The existing implementation relies entirely on CUDA IPC (`CudaIPCWrapper`, `_share_cuda_()`) for cross-process KV cache transfer, which means non-CUDA devices (CPU, XPU, HPU, etc.) cannot benefit from any of these MP mode advantages.

## What this PR does

This PR introduces a `CPUContext` abstract base class with a **two-phase prepare/commit interface** and provides the first concrete implementation — `CPUContextPickle` — to bring MP mode support to non-CUDA devices.

The design anticipates **two transfer paths** for non-CUDA MP mode:

| Path | Mechanism | When to use | Trade-off |
|------|-----------|-------------|-----------|
| **Pickle** (this PR) | `pickle.dumps/loads` over MQ | When `/dev/shm` is smaller than L1 pool (e.g., K8s pods with default 64 MiB shm) | Universal fallback; involves 2 extra memory copies (serialize + deserialize) |
| **SHM** (follow-up PR) | Direct shared-memory read/write | When `/dev/shm` ≥ L1 pool size | Default mode; eliminates 2 memory copies vs pickle for near zero-copy transfer |

SHM will be the **default mode**. At initialization, if SHM allocation fails (insufficient `/dev/shm`), the system automatically falls back to the pickle path.

### Key changes

- **New `CPUContext` ABC** (`cpu_context.py`) with unified `prepare_store` / `commit_store` / `prepare_retrieve` / `commit_retrieve` interface — both pickle and future SHM implementations conform to this contract
- **`CPUContextPickle`** (`cpu_context_pickle.py`) — pickle-based implementation of the two-phase interface
- **`CPUContextMetadata`** dataclass — stores KV layout metadata (layout descriptor, block size, MLA flag) shared across implementations
- **Factory function** `create_cpu_context()` — currently returns `CPUContextPickle`; future SHM PR will add probe + fallback logic here
- adapter to use the unified `prepare/commit` interface — no `if pickle / if shm` branching needed in the adapter layer


lmcache/v1/multiprocess/cpu_context.py
```diff
@@ -0,0 +1,300 @@
...
+        chunk_device = chunk_cpu.to(device)
+
+        if use_mla:
+            for layer_idx, layer in enumerate(layer_tensors):
+                mla_src = chunk_device[layer_idx, skip_tokens:]
+                hidden_size = layer.shape[-1]
+                mla_src_3d = mla_src.reshape(
+                    len(effective_block_ids), block_size, hidden_size
+                )
+                layer[effective_block_ids] = mla_src_3d
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/cpu_context.py
```diff
@@ -0,0 +1,300 @@
...
+
+    chunks: list[torch.Tensor] = []
+    for chunk_idx in range(num_chunks):
+        chunk_block_ids = block_ids[
+            chunk_idx * blocks_per_chunk : (chunk_idx + 1) * blocks_per_chunk
+        ]
+        if use_mla:
+            mla_layers: list[torch.Tensor] = []
+            for layer in layer_tensors:
+                layer_blocks = layer[torch.tensor(chunk_block_ids, dtype=torch.long)]
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -824,18 +846,90 @@ def _send_register_kv_caches_request(
...
+
+        if not kv_caches:
+            raise ValueError("kv_caches is empty")
+        device_types = {tensor.device.type for tensor in kv_caches.values()}
+        if len(device_types) != 1:
+            raise ValueError(
+                f"All KV cache tensors must share one device type, got {device_types}"
+            )
+        self._device_type = next(iter(device_types))
+        self._use_cpu_context = self._device_type != "cuda"
```
maobaolong: This made cpu context special, maybe you could refer to the `lmcache/v1/platform/cache_context.py` of https://github.com/LMCache/LMCache/pull/3242/changes#diff-ab58e80f5b2348daf6570e272210379f3810b65c5ea3c66a44fefa436c935281

hlin99: good suggestion. i'll check how to implement it. thanks.

hlin99: hi @maobaolong, thanks for your advice. I did some investigation today and made some changes as well. On the adapter side, I abstracted a TransferContext for CPU/GPU, so the code change is minimized — even when we later introduce a CPU SHM path.

On the server side, here's my current thinking. Happy to discuss if there's a better way, but so far not come to my mind:

_SHM seems not able to directly reuse CUDA's store()/retrieve(). In the CUDA path, the server can see the worker's paged KV cache through CUDA IPC, so the server does scatter/gather using block_ids by itself. But in SHM, the worker's paged KV cache is not in shared memory — only a transfer buffer is. So the worker has to do scatter/gather on its side first, then write the contiguous chunks into SHM. The server just reads the ready data. That's why we need the two-phase prepare/commit — the worker must finish writing before the server can read. So SHM is actually closer to Pickle in terms of "who does the work" (the worker does), but it avoids the pickle serialization cost by using zero-copy shared memory._

sammshen:
can we start with a design doc for this?

hlin99:
> can we start with a design doc for this?

hi @sammshen  i embedded design doc in this PR. maybe a start from there before deep dive into code? docs/design/v1/multiprocess/cpu_context_design.md

docs/design/v1/multiprocess/non_gpu_context_design.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Transport comparison
+
+**Store (worker → server storage):**
+
+| Transport | Copies | Data flow |
+|---|---|---|
+| CUDA IPC | 2 | GPU KV → GPU staging buffer → CPU memory obj |
+| Pickle | 4 | GPU KV → CPU chunk → pickle.dumps → pickle.loads → CPU memory obj |
+| SHM (TODO) | 1 | GPU KV → CPU memory obj (SHM mapped) |
```
maobaolong: You can refer to the file lmcache/v1/platform/cpu/shm.py of https://github.com/LMCache/LMCache/pull/3242/changes#diff-ace6588444b2a438509e3c529e8618297c1ecdce1f0f3a64272b3885ef50018c

docs/design/v1/multiprocess/cpu_context_design.md
```diff
@@ -0,0 +1,250 @@
...
+
+## 2. Architecture Overview
+
+### 2.1 Layered architecture
+
+```
+vllm_multi_process_adapter.py    ← Engine adapter, device-agnostic
+  └── TransferContext             ← Worker-side transport abstraction (§3)
+        ├── CudaTransferContext    ← CUDA IPC + MQ future path
+        └── CPUTransferContext     ← Synchronous gather/scatter path
```
maobaolong: Hope to place these to the platform module WDYT?

lmcache/v1/multiprocess/non_gpu_context.py
```diff
@@ -0,0 +1,435 @@
...
+def create_cpu_context(
+    metadata: CPUContextMetadata,
+    mq_client: Any,
+    mq_timeout: float,
+) -> CPUContext:
+    """Factory that returns the appropriate :class:`CPUContext` implementation.
+
+    Currently always returns a :class:`~lmcache.v1.multiprocess.\
+cpu_context_pickle.CPUContextPickle` instance.  A future SHM-capable PR
+    may probe for shared-memory availability and fall back to pickle.
```
maobaolong, line 108: Great

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -146,4 +149,28 @@ def get_protocol_definitions() -> dict[str, ProtocolDefinition]:
             response_class=None,
             handler_type=HandlerType.BLOCKING,
         ),
+        "REGISTER_KV_CACHE_CPU_CONTEXT": ProtocolDefinition(
+            payload_classes=[
+                int,
```
maobaolong: dataclass

lmcache/v1/multiprocess/transfer_context.py
```diff
@@ -0,0 +1,404 @@
...
+        mq_timeout: float,
+        send_request: SendRequest,
+    ) -> None:
+        # First Party
+        from lmcache.integration.vllm.utils import vllm_layout_hints
+        from lmcache.integration.vllm.vllm_multi_process_adapter import wrap_kv_caches
+
+        self._mq_client = mq_client
+        self._send_request = send_request
+        layout_hints = vllm_layout_hints()
```
maobaolong: It would be nice if you can pass the layout_hints to register method from caller.

lmcache/v1/multiprocess/server.py
```diff
@@ -189,6 +192,8 @@ def __init__(
         # We assume that if the (model name, world size) is the same, then
         # the layout desc returned by the gpu context is the same.
         self.gpu_context_meta: dict[int, tuple[str, int]] = {}
+        self.cpu_contexts: dict[int, CPUContextMetadata] = {}
```
maobaolong: Suggest to reuse the `gpu_contexts`, and feel free to rename `gpu_contexts` to a more generic name.

lmcache/v1/multiprocess/transfer_context.py
```diff
@@ -0,0 +1,404 @@
...
+            RequestType.RETRIEVE,
+            [key, instance_id, block_ids, event.ipc_handle(), skip_first_n_tokens],
+        ).to_cuda_future()
+
+    def close(self) -> None:
+        self._mq_client = None
+        self._send_request = None
+
+
+class CPUTransferContext(TransferContext):
```
maobaolong: Suggest to rename to Noncuda TransferContext?

lmcache/v1/multiprocess/cpu_context.py
```diff
@@ -0,0 +1,435 @@
...
+        block_size: Number of tokens per paged block.
+        use_mla: Whether the worker KV format is MLA.
+    """
+
+    layout_desc: MemoryLayoutDesc
+    block_size: int
+    use_mla: bool
+
+
+class CPUContext(ABC):
```
maobaolong: Suggest to rename to NoneGpuContext?

hlin99: done. thx for the suggestion. semantic is much clearer

lmcache/v1/multiprocess/transfer_context.py
hlin99: done. thx for the suggestion. semantic is much clearer

lmcache/v1/multiprocess/server.py
hlin99: thanks for the suggestion. fixed.

lmcache/v1/multiprocess/protocols/engine.py
hlin99: resolved. thx

lmcache/v1/multiprocess/transfer_context.py
hlin99: is this the comment for https://github.com/LMCache/LMCache/pull/3280 ?

docs/design/v1/multiprocess/non_gpu_context_design.md
hlin99: hi @maobaolong  discussed offline already. Do we still have any open for this one?

docs/design/v1/multiprocess/cpu_context_design.md
hlin99: hi @maobaolong  to make semantic clearer, following your suggestion, all the cpu_xxx are changed to non_cuda_xxx.  do you still have any concern about this topic? thx.

hlin99:
hi @maobaolong suppose all the issues are addressed now. would you like to take a look? really appreciate for your suggestions.

lmcache/v1/distributed/config.py
```diff
@@ -38,6 +42,15 @@ class L1MemoryManagerConfig:
     def __post_init__(self):
         self.init_size_in_bytes = min(self.init_size_in_bytes, self.size_in_bytes)
 
+        # LazyMemoryAllocator requires cudart (CUDA host-pinned memory).
+        # Auto-disable on non-CUDA backends to avoid a RuntimeError.
+        if self.use_lazy and not hasattr(torch_dev, "cudart"):
+            logger.warning(
```
maobaolong, line 48: I like this fallback

lmcache/v1/multiprocess/transfer_context.py
hlin99: fixed.

lmcache/v1/multiprocess/server.py
```diff
@@ -181,14 +224,8 @@ def __init__(
...
-        # GPU ID -> KV cache tensors
-        self.gpu_contexts: dict[int, GPUCacheContext] = {}
-
-        # GPU ID -> (model name, world size) as metadata
-        # NOTE: This is mainly for determining the layout desc during prefetch
-        # We assume that if the (model name, world size) is the same, then
-        # the layout desc returned by the gpu context is the same.
-        self.gpu_context_meta: dict[int, tuple[str, int]] = {}
+        # Worker instance ID -> registered context metadata
+        self.contexts: dict[int, RegisteredContext] = {}
```
maobaolong, line 232: This abstraction looks pretty good!

docs/design/v1/multiprocess/cpu_context_design.md
maobaolong: Not yet

docs/design/v1/multiprocess/non_gpu_context_design.md
maobaolong: @hlin99 As the discussion offline, it would be nice to have another new PR to finish the shm implementation with two way(worker side or server side memory)

maobaolong (commented):
@hlin99 Big thanks for this PR, it is a key part of our goal. LGTM

maobaolong (approved):

maobaolong:
Hi @ApostaC , would you like to take a double check for this PR? Thanks!

lmcache/v1/multiprocess/transfer_context.py
```diff
@@ -0,0 +1,408 @@
...
+        Raises:
+            RuntimeError: If register() was not called first.
+        """
+
+    @abstractmethod
+    def close(self) -> None:
+        """Release resources held by this context."""
+
+
+class CudaTransferContext(TransferContext):
```
maobaolong: Maybe we can have another name since non-Cuda can use this class also, I'm going to implement a server-side shm copy approach.

For this class, it is responsible for send store/retrieve request to the server and let server-side handle this request, so it need another name.

Maybe `ServerSideTransferContext`, maybe others? @hlin99

hlin99: Rename to HandleTransferContext for CUDA and DataTransferContext for CPU based on their implementations.

maobaolong:
@hlin99 Maybe you forget to rename `tests/v1/multiprocess/test_non_cuda_context.py` ?

hlin99:
> @hlin99 Maybe you forget to rename `tests/v1/multiprocess/test_non_cuda_context.py` ?

good catch @maobaolong . fixed in latest commit

KuntaiDu (approved):
LGTM!

PR #3261, opened 2026-05-12
[MP][DSV4] Support hybrid KV cache manager (SupportsHMA)

yoo-kumaneko:

Enables LMCache MP to work with vLLM's hybrid KV cache manager
(`SupportsHMA`) — required for DeepSeek-V4-Flash, which exposes 5
`KVCacheGroupSpec`s with mixed scheduler block sizes (256 / 64 / 4 / 8),
two pairs of physically identical SWA layer groups in disjoint
block-ID namespaces, and an `as_strided` `page_size_padded` MLA
layout.

Rebased on top of the day-0 V4 support landed in #3171
(`block_stride_elems` kernel + `compress_ratio` /
`physical_chunk_size` plumbing); that PR works for V4 with
`--disable-hybrid-kv-cache-manager`. This PR adds the missing pieces
for `--no-disable-hybrid-kv-cache-manager`.

The work is split into eight small commits for review:

1. **Per-group `logical_block_size` in `LayerGroupIdentity`** —
   `LayerGroupIdentity` extends from a 5-tuple to a 6-tuple, adding
   the engine's *scheduler-side* block size (`KVCacheSpec.block_size`).
   Layers with same physical layout but different scheduler grids
   (V4 hybrid: dense MLA = 256-token grid, SWA = 64-token grid) now
   land in distinct LMCache groups. `_derive_compression_metadata`
   keys `compress_ratio` on the per-group `logical_bs / physical_bs`
   instead of one global `inference_engine_logical_block_size`.

2. **Per-group `kv_cache_group_id` namespace in `LayerGroupIdentity`** —
   identity extends to a 7-tuple, adding the engine's per-layer
   block-ID *namespace* handle. Two physically-identical layer sets
   that allocate from disjoint vLLM `BlockPool` namespaces (V4: gid 1
   even+MTP-SWA vs gid 2 odd-SWA) now land in distinct LMCache groups
   so a STORE op carrying gid 1's block IDs cannot mis-index gid 2's
   layers.

3. **Per-gid `LoadStoreOp` wire format + per-LMCache-group dispatch** —
   `LoadStoreOp.block_ids` changes from `list[int]` to
   `list[list[int]]` indexed by engine-side gid in registration
   order. New `GPUCacheContext.stage_block_ids_per_namespace` packs
   all namespaces into the shared GPU buffer at non-overlapping
   offsets. `MPCacheEngine.store` / `retrieve` loop per LMCache group
   with each group's own `blocks_per_chunk(g)`, kernel chunk size
   (= `get_physical_chunk_size(g)`), and namespace slice
   (`staged[group.kv_cache_group_id]`). This is a **breaking RPC
   change**: old servers cannot accept the new ops or vice versa.

4. **SWA-suffix-only store/retrieve** — for SWA groups
   (`sliding_window > 0`), store and retrieve only the trailing
   `ceil(window / logical_bs)` blocks of each chunk instead of the
   full chunk. The kernel-time SWA mask in the serving engine never
   reads positions older than `sliding_window`, so the truncated
   payload is sufficient. `LayerGroupIdentity` extends to an 8-tuple
   including `sliding_window`. Originally introduced behind an
   `LMCACHE_SWA_SUFFIX_ONLY=1` env-var gate; promoted to default in
   commit 8 after end-to-end verification. On V4-Flash
   (chunk_size=1024, windows of 8/128/256), this collapses gid 3's
   per-chunk SWA payload from ~236 MiB to ~1.85 MiB.

5. **Fix tmp buffer shape for SWA-truncated groups** — discovered
   during end-to-end smoke testing: `get_tmp_chunk_gpu_buffer`
   computed shape from `lmcache_logical_chunk_size` but the per-group
   tmp slot was sized at construction to the SWA-truncated payload.
   Mismatch caused `RuntimeError: shape '[1, 22, 1024, 584]' is
   invalid for input of size 1644544` on store. Fixed via a single
   `storage_logical_tokens_per_chunk(g)` helper used everywhere.

6. **Lint fixes** — `ruff-format` reformatted a few short ternaries;
   one mypy fix in `_build_manager`'s test helper signature.

7. **Port hybrid-on (`SupportsHMA`) to LMCache-side
   `lmcache_mp_connector`** — LMCache PR #3235 added an in-tree copy
   of the vLLM-side `LMCacheMPConnector` at
   `lmcache/integration/vllm/lmcache_mp_connector.py`, intended as
   the LMCache-tracked entrypoint users pin via
   `kv_connector_module_path`. That copy was forked off the
   pre-hybrid vLLM-side connector, so it still asserted a single
   `KVCacheGroupSpec` and was incompatible with this PR's per-gid
   wire format. This commit ports the full hybrid-on path to the
   LMCache-side copy: `SupportsHMA` inheritance, per-gid request
   tracker, per-gid `LoadStoreOp.block_ids`, `_per_gid_slice` with
   the divide-or-multiply rule for the GCD-collapsed
   `cache_config.block_size`, `register_kv_caches` forwarding
   `extra_layout_hints`, `request_finished_all_groups`, and the L0
   "primary namespace" reporting on gid 0. **No corresponding vLLM
   PR is required** — users pin via:

   ```
   --kv-transfer-config '{"kv_connector":"LMCacheMPConnector",
                          "kv_connector_module_path":
                            "lmcache.integration.vllm.lmcache_mp_connector",
                          ...}'
   ```

   The static `lmcache_mp_connector_0180.py` /
   `lmcache_mp_connector_0201.py` snapshots are intentionally NOT
   touched — those older vLLM revisions cannot drive HMA at all and
   keep their flat-list semantics; they hard-fail the moment
   `len(kv_cache_groups) > 1`, which is the right behavior.

8. **Make SWA-suffix-only the default** — drop the
   `LMCACHE_SWA_SUFFIX_ONLY=1` env-var gate introduced in commit 4
   and unconditionally publish `per_layer_sliding_window` via
   `extra_layout_hints`. Verified end-to-end (token IDs identical,
   max|Δlogprob| = 5.04e-04, well below the #13 noise floor) so the
   optimization is safe to enable by default. Full-attention groups
   (`sliding_window == 0`) are byte-for-byte unaffected. No off-switch
   env var — users who need the prior behavior can pin to a previous
   LMCache release.

Hybrid-OFF behavior is preserved by single-group/single-namespace
defaults: when no per-layer hint is supplied, every layer's
`logical_block_size` defaults to its physical `bs`, every namespace
to 0, every `sliding_window` to 0, so the 8-tuple identity collapses
to the prior 5-tuple grouping.

- **No companion vLLM PR is required.** Commit 7 adds the full
  hybrid-on path to the LMCache-tree connector at
  `lmcache/integration/vllm/lmcache_mp_connector.py`. Users opt into
  the LMCache-tracked copy via `kv_connector_module_path`, so the
  in-tree vLLM `LMCacheMPConnector` does not need any patches.
  Verified end-to-end: with a vanilla upstream vLLM (no local
  changes), DeepSeek-V4-Flash + `--no-disable-hybrid-kv-cache-manager`
  + LMCacheMPConnector gives 2.31×–16.47× speedups (see table
  below).
- **Backward compat**: every non-V4 / single-namespace caller hits
  a length-1 `gpu_block_ids` list, lands on namespace 0, has
  `logical == physical`, has `sliding_window == 0`, and sees
  byte-for-byte identical buffer offsets and kernel args. V3.2
  behavior is preserved (verified locally).
- **Wire format break**: protocol payload type for `STORE` /
  `RETRIEVE` updated from `list[int]` to `list[list[int]]`. Old
  LMCache servers cannot accept new ops, and new servers cannot
  accept old ops. No compat shim because the MP RPC is internal —
  same versioning approach as #3171's `vllm_block_size`-payload
  addition.

**Verified end-to-end on DeepSeek-V4-Flash, tp=4, fp8, block_size=256,
`--no-disable-hybrid-kv-cache-manager`, lmcache `chunk_size=1024`,
default-on SWA-suffix**:

Probe-confirmed layer grouping (logged at `register_kv_caches`):
- 8 LMCache groups with the expected 8-tuple split:
  - gid 0 splits into 3 groups (3 dense-MLA layer flavors at
    physical_bs ∈ {64, 64, 2}, logical_bs=256, namespace=0,
    sliding_window=0)
  - gid 1 (22 even+MTP-SWA layers) and gid 2 (21 odd-SWA layers)
    split despite identical specs; logical_bs=64, namespace ∈ {1, 2},
    sliding_window=128
  - gid 3 (compressor states for CSA) splits into 2 groups by hidden
    dim (hs ∈ {512, 2048}); physical_bs=4, logical_bs=4, namespace=3,
    sliding_window=8
  - gid 4 (compressor states for HCA): physical_bs=8, logical_bs=8,
    namespace=4, sliding_window=128

TTFT measurements via `deterministic_ttft.py` (single cold/warm pair
on identical token IDs to bypass `vllm bench serve`'s seed
non-determinism):

| Mode (LMCache-side connector, vanilla upstream vLLM) | Prompt    | Cold TTFT  | Warm TTFT  | Saved          | Speedup |
|------------------------------------------------------|-----------|-----------:|-----------:|---------------:|--------:|
| Hybrid OFF                                           | 9216 tok  | 1337.8 ms  |  520.5 ms  |  817 ms (61.1%) | **2.57×**  |
| Hybrid ON, default-on SWA-suffix                     | 9216 tok  | 1321.6 ms  |  573.2 ms  |  748 ms (56.6%) | **2.31×**  |
| Hybrid ON, default-on SWA-suffix                     | 30000 tok | 2862.0 ms  |  173.7 ms  | 2688 ms (93.9%) | **16.47×** |

The speedup ratio scales with prompt length — prefill is roughly
O(L²) in attention, retrieve is roughly O(L) in H2D bandwidth. The
SWA-suffix optimization compounds at long prompts where the
bandwidth savings (especially gid 3's ~128× per-chunk reduction
from 236 MiB to 1.85 MiB) exceed the per-call gather overhead.

Correctness via `partial_hit_harness.py` (cold-pass reference vs
LMCache-hit replay, comparing token IDs and per-token logprobs):
- token IDs identical (32/32), max|Δlogprob| = 5.04e-04 (well under
  the 4e-1 threshold from the documented #13 noise floor)

Zero `Cannot store keys` / `Cannot retrieve keys` /
`chunk_block_ids_gpu underflow` exceptions in any of the verification
runs.

**If applicable**:

**Tests**:

- `tests/v1/test_kv_layer_groups_manager.py` — extended with
  `TestPerLayerLogicalBlockSize` (5 tests), `TestPerLayerKvCacheGroupId`
  (4 tests), `TestPerLayerSlidingWindow` (4 tests). All 34 tests
  pass.
- `tests/v1/multiprocess/test_cache_server.py`,
  `tests/v1/multiprocess/test_mq.py`,
  `tests/v1/multiprocess/test_mq_handler_helpers.py`,
  `tests/v1/multiprocess/test_blend_server_v2.py` — updated to wrap
  flat block-ID lists in length-1 nested form for the new wire
  protocol.
- All 403 tests in `tests/v1/multiprocess/` +
  `test_kv_layer_groups_manager` + `test_vllm_mp_adapter` pass.

### DIFF
--- a/lmcache/v1/kv_layer_groups.py
+++ b/lmcache/v1/kv_layer_groups.py
@@ -35,10 +35,10 @@ DTYPE_MAP: dict[str, torch.dtype] = {
 }
 
 
-# The 5-tuple that uniquely identifies a set of kernel-equivalent layers:
+# The 7-tuple that uniquely identifies a set of kernel-equivalent layers:
 # ``(kv_size, num_heads, head_size, block_size, logical_block_size,
-# dtype)``. Two layers share a transfer-kernel launch iff they share
-# this identity — see the grouping loop in
+# kv_cache_group_id, dtype)``. Two layers share a transfer-kernel
+# launch iff they share this identity — see the grouping loop in
 # :meth:`KVLayerGroupsManager.__init__` for the derivation.
 #
 # The fifth slot (``logical_block_size``) is the *scheduler-side*
@@ -52,11 +52,27 @@ DTYPE_MAP: dict[str, torch.dtype] = {
 # grid while the SWA layers use a 64-token grid). Two layers with the
 # same physical layout but different scheduler grids are
 # kernel-incompatible — one ``LoadStoreOp`` cannot index both — so the
-# logical block size must participate in the identity tuple. When the
-# hint is absent, every layer's logical block size defaults to its
-# physical ``bs`` and the 6-tuple collapses to the prior 5-tuple
-# behavior.
-LayerGroupIdentity = tuple[int, int, int, int, int, torch.dtype]
+# logical block size must participate in the identity tuple.
+#
+# The sixth slot (``kv_cache_group_id``) is the engine's per-layer
+# block-ID namespace handle. Two layers may share an LMCache transfer-
+# kernel group only if a single ``LoadStoreOp``'s ``block_ids``
+# correctly indexes both — which is true iff they share a namespace.
+# It comes from
+# :data:`~lmcache.v1.gpu_connector.utils.LayoutHints.per_layer_kv_cache_group_id`
+# when the engine supplies it, otherwise every layer is treated as
+# namespace 0. On DeepSeek-V4 with the hybrid manager active, the
+# even-indexed-plus-MTP SWA layers (vLLM gid 1) and the odd-indexed
+# SWA layers (vLLM gid 2) have identical ``KVCacheSpec`` field values
+# — same ``block_size``, ``sliding_window``, ``head_size``, dtype, and
+# physical ``shape[1]`` — but pull block IDs from disjoint pools.
+# Without this slot they merge into one LMCache group and a STORE op
+# carrying gid 1's block IDs would silently mis-index gid 2's layers.
+#
+# When neither hint is provided, every layer's logical block size
+# defaults to its physical ``bs`` and every layer's namespace defaults
+# to 0, so the 7-tuple collapses to the prior 5-tuple grouping.
+LayerGroupIdentity = tuple[int, int, int, int, int, int, torch.dtype]
 
 
 @dataclass
@@ -131,6 +147,18 @@ class KVLayerGroupInfo:
     indicates the field has not been populated yet (left in for
     backwards compatibility with no-arg constructors); the manager
     always sets a positive value at construction."""
+    kv_cache_group_id: int = 0
+    """Engine-side block-ID namespace handle for this group. Stamped
+    from
+    :data:`~lmcache.v1.gpu_connector.utils.LayoutHints.per_layer_kv_cache_group_id`
+    at registration. Two LMCache groups with different
+    ``kv_cache_group_id`` values reflect layers whose
+    ``LoadStoreOp.block_ids`` come from disjoint engine-side
+    ``BlockPool``-allocated namespaces — they cannot be merged into one
+    transfer-kernel launch even when every other field matches.
+    Defaults to 0 when the engine does not provide the hint, in which
+    case all layers share namespace 0 and grouping behaves identically
+    to the prior 6-tuple identity."""
 
     def __repr__(self) -> str:
         if not self.layer_indices:
@@ -148,7 +176,8 @@ class KVLayerGroupInfo:
             f"dtype={self.dtype}, "
             f"compress_ratio={self.compress_ratio}, "
             f"physical_chunk_size={self.physical_chunk_size}, "
-            f"logical_bs={self.logical_block_size})"
+            f"logical_bs={self.logical_block_size}, "
+            f"kv_cache_group_id={self.kv_cache_group_id})"
         )
 
     @property
@@ -167,9 +196,10 @@ class KVLayerGroupsManager:
 
     At construction time, every layer in ``kv_caches`` is bucketed by its
     :data:`LayerGroupIdentity` (``(kv_size, num_heads, head_size,
-    block_size, logical_block_size, dtype)``). Each bucket becomes one
-    :class:`KVLayerGroupInfo` holding the layer indices, a shared
-    :class:`PageBufferShapeDesc`, and the group's torch dtype.
+    block_size, logical_block_size, kv_cache_group_id, dtype)``). Each
+    bucket becomes one :class:`KVLayerGroupInfo` holding the layer
+    indices, a shared :class:`PageBufferShapeDesc`, and the group's
+    torch dtype.
 
     Downstream consumers (``VLLMPagedMemGPUConnectorV3``,
     ``GPUCacheContext``, the multiprocess server) iterate
@@ -228,11 +258,19 @@ class KVLayerGroupsManager:
                   Required when the engine has KV layer groups with
                   mixed scheduler block sizes (vLLM hybrid manager
                   active on DeepSeek-V4).
+                * ``per_layer_kv_cache_group_id`` (optional list of
+                  length ``num_layers``). When present, each layer's
+                  engine-side block-ID *namespace* handle joins the
+                  identity tuple, so layers whose ``KVCacheSpec``
+                  field values fully match but pull block IDs from
+                  disjoint engine-side namespaces (e.g.
+                  DeepSeek-V4's vLLM gids 1 and 2: even+MTP vs odd
+                  SWA layers) end up in different LMCache groups.
 
-                ``None`` (or a hints dict without either key) means
-                every group is treated as non-compressed
+                ``None`` (or a hints dict without any of the above)
+                means every group is treated as non-compressed
                 (``compress_ratio == 1``, ``logical_block_size ==
-                shape_desc.bs``).
+                shape_desc.bs``, ``kv_cache_group_id == 0``).
             lmcache_logical_chunk_size: Logical tokens per LMCache chunk
                 (one logical token = one inference-engine token).
                 Together with ``compress_ratio`` it determines each
@@ -277,6 +315,7 @@ class KVLayerGroupsManager:
         # shape but different scheduler grids land in distinct LMCache
         # groups (DeepSeek-V4 hybrid manager active).
         per_layer_logical_bs: "list[int] | None" = None
+        per_layer_namespace: "list[int] | None" = None
         if layout_hints is not None:
             logical_hint = layout_hints.get("per_layer_logical_block_size")
             if logical_hint is not None:
@@ -299,14 +338,37 @@ class KVLayerGroupsManager:
                     )
                 per_layer_logical_bs = list(logical_hint)
 
+            ns_hint = layout_hints.get("per_layer_kv_cache_group_id")
+            if ns_hint is not None:
+                if len(ns_hint) != num_layers:
+                    raise ValueError(
+                        "per_layer_kv_cache_group_id length "
+                        f"({len(ns_hint)}) does not match num_layers "
+                        f"({num_layers}); each registered layer must "
+                        "have exactly one block-ID namespace entry"
+                    )
+                bad = [i for i, ns in enumerate(ns_hint) if ns < 0]
+                if bad:
+                    raise ValueError(
+                        "per_layer_kv_cache_group_id has invalid "
+                        f"(negative) entries at positions {bad[:8]}"
+                        + ("..." if len(bad) > 8 else "")
+                        + "; namespace IDs must be non-negative"
+                    )
+                per_layer_namespace = list(ns_hint)
+
         groups_dict = self._group_layers_by_identity(
-            kv_caches, gpu_kv_format, num_layers, per_layer_logical_bs
+            kv_caches,
+            gpu_kv_format,
+            num_layers,
+            per_layer_logical_bs,
+            per_layer_namespace,
         )
 
         # Emit groups in order of their first-appearing layer so that group
         # indices remain deterministic across runs.
         for group_idx, (
-            (_, _, _, bs, group_logical_bs, dt),
+            (_, _, _, bs, group_logical_bs, group_namespace, dt),
             indices,
         ) in enumerate(sorted(groups_dict.items(), key=lambda kv: kv[1][0])):
             block_stride_elems = resolve_block_stride_and_log_layout(
@@ -341,6 +403,7 @@ class KVLayerGroupsManager:
                     compress_ratio=compress_ratio,
                     physical_chunk_size=physical_chunk_size,
                     logical_block_size=group_logical_bs,
+                    kv_cache_group_id=group_namespace,
                 )
             )
 
@@ -440,20 +503,22 @@ class KVLayerGroupsManager:
         gpu_kv_format: "lmc_ops.GPUKVFormat",
         num_layers: int,
         per_layer_logical_bs: "list[int] | None" = None,
+        per_layer_namespace: "list[int] | None" = None,
     ) -> dict[LayerGroupIdentity, list[int]]:
         """Partition layer indices by :data:`LayerGroupIdentity`.
 
         Linear single pass over ``kv_caches``; layers sharing the same
         ``(kv_size, num_heads, head_size, block_size,
-        logical_block_size, dtype)`` signature land in the same bucket.
-        When ``per_layer_logical_bs`` is None, each layer's
-        ``logical_block_size`` defaults to its physical block size,
-        making the 6-tuple effectively a 5-tuple (preserving prior
-        grouping behavior on engines without the hint). The returned
-        dict's value lists are later passed by reference into
-        :class:`KVLayerGroupInfo` instances, so the dict itself is
-        garbage-collected after ``__init__`` returns while the lists
-        stay alive on each group.
+        logical_block_size, kv_cache_group_id, dtype)`` signature land
+        in the same bucket. When ``per_layer_logical_bs`` is None, each
+        layer's ``logical_block_size`` defaults to its physical block
+        size; when ``per_layer_namespace`` is None, each layer's
+        namespace defaults to 0. With both defaults, the 7-tuple is
+        effectively a 5-tuple (preserving prior grouping behavior on
+        engines without these hints). The returned dict's value lists
+        are later passed by reference into :class:`KVLayerGroupInfo`
+        instances, so the dict itself is garbage-collected after
+        ``__init__`` returns while the lists stay alive on each group.
         """
         # First Party
         from lmcache.v1.gpu_connector.utils import (
@@ -477,7 +542,14 @@ class KVLayerGroupsManager:
                 if per_layer_logical_bs is not None
                 else bs
             )
-            groups_dict[(kv_size, nh, hs, bs, logical_bs, dt)].append(idx)
+            namespace = (
+                per_layer_namespace[idx]
+                if per_layer_namespace is not None
+                else 0
+            )
+            groups_dict[
+                (kv_size, nh, hs, bs, logical_bs, namespace, dt)
+            ].append(idx)
         return groups_dict
 
     @property
--- a/lmcache/v1/kv_layer_groups.py
+++ b/lmcache/v1/kv_layer_groups.py
@@ -606,14 +606,10 @@ class KVLayerGroupsManager:
             dt = get_dtype(kv_caches, gpu_kv_format, idx)
             bs = get_block_size(kv_caches, gpu_kv_format, idx)
             logical_bs = (
-                per_layer_logical_bs[idx]
-                if per_layer_logical_bs is not None
-                else bs
+                per_layer_logical_bs[idx] if per_layer_logical_bs is not None else bs
             )
             namespace = (
-                per_layer_namespace[idx]
-                if per_layer_namespace is not None
-                else 0
+                per_layer_namespace[idx] if per_layer_namespace is not None else 0
             )
             sliding_window = (
                 per_layer_sliding_window[idx]
--- a/lmcache/v1/kv_layer_groups.py
+++ b/lmcache/v1/kv_layer_groups.py
@@ -35,154 +35,68 @@ DTYPE_MAP: dict[str, torch.dtype] = {
 }
 
 
-# The 8-tuple that uniquely identifies a set of kernel-equivalent layers:
-# ``(kv_size, num_heads, head_size, block_size, logical_block_size,
-# kv_cache_group_id, sliding_window, dtype)``. Two layers share a
-# transfer-kernel launch iff they share this identity — see the
-# grouping loop in :meth:`KVLayerGroupsManager.__init__` for the
-# derivation.
+# 8-tuple identity for kernel-equivalent layers. Two layers share a
+# transfer-kernel launch iff their identities match. The first five fields
+# (``kv_size, num_heads, head_size, block_size, dtype``) are the legacy
+# kernel-shape signature. The other three came from the DeepSeek-V4
+# hybrid-KV-cache work and default to safe values that collapse the tuple
+# to the legacy 5-tuple when the engine doesn't supply hints:
 #
-# The fifth slot (``logical_block_size``) is the *scheduler-side*
-# block size — i.e. the granularity at which the serving engine hands
-# out block IDs for this layer. It comes from
-# :data:`~lmcache.v1.gpu_connector.utils.LayoutHints.per_layer_logical_block_size`
-# when the engine supplies it, otherwise it defaults to the physical
-# block size (``bs``). On DeepSeek-V4 with vLLM's *hybrid manager
-# active*, several physical-equivalent layer sets receive block IDs at
-# different scheduler grids (e.g. dense-MLA layers use a 256-token
-# grid while the SWA layers use a 64-token grid). Two layers with the
-# same physical layout but different scheduler grids are
-# kernel-incompatible — one ``LoadStoreOp`` cannot index both — so the
-# logical block size must participate in the identity tuple.
-#
-# The sixth slot (``kv_cache_group_id``) is the engine's per-layer
-# block-ID namespace handle. Two layers may share an LMCache transfer-
-# kernel group only if a single ``LoadStoreOp``'s ``block_ids``
-# correctly indexes both — which is true iff they share a namespace.
-# It comes from
-# :data:`~lmcache.v1.gpu_connector.utils.LayoutHints.per_layer_kv_cache_group_id`
-# when the engine supplies it, otherwise every layer is treated as
-# namespace 0. On DeepSeek-V4 with the hybrid manager active, the
-# even-indexed-plus-MTP SWA layers (vLLM gid 1) and the odd-indexed
-# SWA layers (vLLM gid 2) have identical ``KVCacheSpec`` field values
-# — same ``block_size``, ``sliding_window``, ``head_size``, dtype, and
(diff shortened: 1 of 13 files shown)


KuntaiDu:
Thank you for your contribution! We are discussing this design and will get back to you.

lmcache/v1/multiprocess/server.py
```diff
@@ -356,19 +389,58 @@ def store(
...
+                                idx,
+                                bpc_g,
+                                chunk_block_ids_gpu.shape[0],
+                                ns_block_ids_gpu.shape[0],
+                                group.logical_block_size,
+                                group.physical_block_size,
+                                self.chunk_size,
+                                idx * bpc_g,
+                                (idx + 1) * bpc_g,
+                            )
```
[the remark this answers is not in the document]
yoo-kumaneko: We don't want the server to crash. So, we will bear the error.

lmcache/v1/kv_layer_groups.py
```diff
@@ -180,29 +339,51 @@ def __init__(
         mla = is_mla(gpu_kv_format)
         kv_size = 1 if mla else 2
         groups_dict: dict[LayerGroupIdentity, list[int]] = defaultdict(list)
```
maobaolong: There are so many variables named to `group` `layer`, it could make sense before we introduce `DeepSeekV4`, but now, the similar concept make me a little confusing, suggested to start a new individual PR to do a single renaming. @yoo-kumaneko

H100-H200-B200:
Tried to host with a 8*H20 node, vllm https://github.com/yoo-kumaneko/vllm/tree/feat/lmcache-mp-connector-supports-hma, cuda 13.0, lmcache https://github.com/yoo-kumaneko/LMCache/tree/feat/dsv4-hma-support
```
export PYTHONHASHSEED=0
unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
export GRPC_ENABLE_HTTP_PROXY=0
export no_proxy=localhost,127.0.0.1
export NO_PROXY=localhost,127.0.0.1
numactl --interleave=all lmcache server \
    --host 127.0.0.1 \
    --port 57771 \
    --http-host 127.0.0.1 \
    --http-port 58881 \
    --chunk-size 512 \
    --max-workers 64 \
    --hash-algorithm blake3 \
    --engine-type default \
    --lookup-hash-log-dir /data7/lmcache/lookup_hashes \
    --lookup-hash-log-rotation-interval 21600 \
    --lookup-hash-log-rotation-max-size 104857600 \
    --lookup-hash-log-max-files 100 \
    --l1-size-gb 1200 \
    --l1-use-lazy \
    --l1-init-size-gb 20 \
    --l1-align-bytes 4096 \
    --l1-write-ttl-seconds 600 \
    --l1-read-ttl-seconds 300 \
    --eviction-policy noop \
    --l2-store-policy skip_l1 \
    --eviction-trigger-watermark 0.9 \
    --eviction-ratio 0.1 \
    --l2-prefetch-policy default \
    --l2-prefetch-max-in-flight 8 \
    --l2-adapter '{"type": "nixl_store", "backend": "POSIX", "backend_params": {"file_path": "/data7/lmcache/l2", "use_direct_io": "false"}, "pool_size": 3000}' \
    --prometheus-port 9090 \
    --metrics-sample-rate 0.01 
```
The vllm serve command is 
```
export PYTHONHASHSEED=0
export VLLM_ENGINE_READY_TIMEOUT_S=3600
export VLLM_RPC_TIMEOUT=300000
numactl --interleave=all vllm serve \
  /data1/public/models/DeepSeek-V4-Flash \
  --served-model-name deepseek-v4-flash \
  --host 127.0.0.1 \
  --port 10081 \
  --trust-remote-code \
  --kv-cache-dtype fp8 \
  --block-size 256 \
  --enable-expert-parallel \
  --tensor-parallel-size 8 \
  --tokenizer-mode deepseek_v4 \
  --tool-call-parser deepseek_v4 \
  --enable-auto-tool-choice \
  --reasoning-parser deepseek_v4 \
  --max-num-seqs 128 \
  --kv-transfer-config \
    '{"kv_connector":"LMCacheMPConnector", "kv_role":"kv_both", "kv_connector_extra_config": {"lmcache.mp.host": "tcp://127.0.0.1", "lmcache.mp.port": 57771}}' \
  --no-disable-hybrid-kv-cache-manager 
```

But external cache hit rate always remained zero in production enviroment like
```
(APIServer pid=1962054) INFO 05-15 09:36:42 [loggers.py:271] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 18.0 tokens/s, Running: 19 reqs, Waiting: 2 reqs, GPU KV cache usage: 37.3%, Prefix cache hit rate: 61.6%, External prefix cache hit rate: 0.0%
(APIServer pid=1962054) INFO 05-15 09:36:52 [loggers.py:271] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 12.6 tokens/s, Running: 19 reqs, Waiting: 2 reqs, GPU KV cache usage: 37.7%, Prefix cache hit rate: 61.6%, External prefix cache hit rate: 0.0%
(APIServer pid=1962054) INFO 05-15 09:37:02 [loggers.py:271] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 12.6 tokens/s, Running: 19 reqs, Waiting: 2 reqs, GPU KV cache usage: 38.1%, Prefix cache hit rate: 61.6%, External prefix cache hit rate: 0.0%
```
Requests shown in client side (new api) also indicates that kv cache is not working even there's enough ddr/ssd to offload (with --enable-prompt-tokens-details serve command, it should show information of cache read if cache works):
image
If the cache works, it should show cache read information in small words like: 
image

MP server got lots of peculiar `session not found` warning messages like 
```
[2026-05-15 01:30:27,103] LMCache WARNING: Session chatcmpl-9e5f7cfd4e1821c1-86e547b4 not found, skipping touch
[2026-05-15 01:30:34,076] LMCache INFO: Stored 512 tokens in 0.003 seconds (server.py:515:lmcache.v1.multiprocess.server
```

Is there's something wrong with my host command?

yoo-kumaneko:
**Rebased on top of #3171 (currently `origin/dev`); broke the original squashed commit into 7 small reviewable commits.** Branch summary:

- `2aafb7e` Per-group `logical_block_size` (Step 1)
- `768a2dc` Per-group `kv_cache_group_id` namespace (Step 2)
- `b4d6500` Per-gid `LoadStoreOp` wire format + per-LMCache-group dispatch (Step 3)
- `58e1701` SWA-suffix-only optimization behind `LMCACHE_SWA_SUFFIX_ONLY=1` (Step 4 — new vs prior PR revision)
- `c8250c4` Wrap `block_ids` for legacy LMCache-side vLLM connectors
- `13fb7cb` Fix tmp-buffer shape for SWA-truncated groups (caught during e2e verify)
- `01dc992` Lint fixes

PR description rewritten end-to-end. Hybrid-OFF behavior preserved by per-layer hint defaults; the 8-tuple `LayerGroupIdentity` collapses to the prior 5-tuple when no hints are provided.

End-to-end verified on DeepSeek-V4-Flash tp=4 hybrid-on:

| Mode | Prompt | Cold | Warm | Speedup |
|---|---|---:|---:|---:|
| Non-SWA | 30000 tok | 4087 ms | 447 ms | **9.15×** |
| SWA-suffix | 30000 tok | 4088 ms | 403 ms | **10.14×** |

Token IDs identical and max\|Δlogprob\| ≤ 1.35e-3 against cold-pass reference (well under the 4e-1 noise-floor threshold from the documented partial-hit harness). Zero exceptions in store/retrieve loops on either mode.

yoo-kumaneko:
I have rebased the changes on top of the current dev branch.

H100-H200-B200:
> Tried to host with a 8*H20 node, vllm https://github.com/yoo-kumaneko/vllm/tree/feat/lmcache-mp-connector-supports-hma, cuda 13.0, lmcache https://github.com/yoo-kumaneko/LMCache/tree/feat/dsv4-hma-support
> 
> ```
> export PYTHONHASHSEED=0
> unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
> export GRPC_ENABLE_HTTP_PROXY=0
> export no_proxy=localhost,127.0.0.1
> export NO_PROXY=localhost,127.0.0.1
> numactl --interleave=all lmcache server \
>     --host 127.0.0.1 \
>     --port 57771 \
>     --http-host 127.0.0.1 \
>     --http-port 58881 \
>     --chunk-size 512 \
>     --max-workers 64 \
>     --hash-algorithm blake3 \
>     --engine-type default \
>     --lookup-hash-log-dir /data7/lmcache/lookup_hashes \
>     --lookup-hash-log-rotation-interval 21600 \
>     --lookup-hash-log-rotation-max-size 104857600 \
>     --lookup-hash-log-max-files 100 \
>     --l1-size-gb 1200 \
>     --l1-use-lazy \
>     --l1-init-size-gb 20 \
>     --l1-align-bytes 4096 \
>     --l1-write-ttl-seconds 600 \
>     --l1-read-ttl-seconds 300 \
>     --eviction-policy noop \
>     --l2-store-policy skip_l1 \
>     --eviction-trigger-watermark 0.9 \
>     --eviction-ratio 0.1 \
>     --l2-prefetch-policy default \
>     --l2-prefetch-max-in-flight 8 \
>     --l2-adapter '{"type": "nixl_store", "backend": "POSIX", "backend_params": {"file_path": "/data7/lmcache/l2", "use_direct_io": "false"}, "pool_size": 3000}' \
>     --prometheus-port 9090 \
>     --metrics-sample-rate 0.01 
> ```
> 
> The vllm serve command is
> 
> ```
> export PYTHONHASHSEED=0
> export VLLM_ENGINE_READY_TIMEOUT_S=3600
> export VLLM_RPC_TIMEOUT=300000
> numactl --interleave=all vllm serve \
>   /data1/public/models/DeepSeek-V4-Flash \
>   --served-model-name deepseek-v4-flash \
>   --host 127.0.0.1 \
>   --port 10081 \
>   --trust-remote-code \
>   --kv-cache-dtype fp8 \
>   --block-size 256 \
>   --enable-expert-parallel \
>   --tensor-parallel-size 8 \
>   --tokenizer-mode deepseek_v4 \
>   --tool-call-parser deepseek_v4 \
>   --enable-auto-tool-choice \
>   --reasoning-parser deepseek_v4 \
>   --enable-prompt-tokens-details \
>   --gpu-memory-utilization 0.9 \
>   --max-num-seqs 128 \
>   --kv-transfer-config \
>     '{"kv_connector":"LMCacheMPConnector", "kv_role":"kv_both", "kv_connector_extra_config": {"lmcache.mp.host": "tcp://127.0.0.1", "lmcache.mp.port": 57771}}' \
>   --no-disable-hybrid-kv-cache-manager 
> ```
> 
> But external cache hit rate always remained zero in production enviroment like
> 
> ```
> (APIServer pid=1962054) INFO 05-15 09:36:42 [loggers.py:271] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 18.0 tokens/s, Running: 19 reqs, Waiting: 2 reqs, GPU KV cache usage: 37.3%, Prefix cache hit rate: 61.6%, External prefix cache hit rate: 0.0%
> (APIServer pid=1962054) INFO 05-15 09:36:52 [loggers.py:271] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 12.6 tokens/s, Running: 19 reqs, Waiting: 2 reqs, GPU KV cache usage: 37.7%, Prefix cache hit rate: 61.6%, External prefix cache hit rate: 0.0%
> (APIServer pid=1962054) INFO 05-15 09:37:02 [loggers.py:271] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 12.6 tokens/s, Running: 19 reqs, Waiting: 2 reqs, GPU KV cache usage: 38.1%, Prefix cache hit rate: 61.6%, External prefix cache hit rate: 0.0%
> ```
> 
> Requests shown in client side (new api) also indicates that kv cache is not working even there's enough ddr/ssd to offload (with --enable-prompt-tokens-details serve command, it should show information of cache read if cache works): image If the cache works, it should show cache read information in small words like: image
> 
> MP server got lots of peculiar `session not found` warning messages like
> 
> ```
> [2026-05-15 01:30:27,103] LMCache WARNING: Session chatcmpl-9e5f7cfd4e1821c1-86e547b4 not found, skipping touch
> [2026-05-15 01:30:34,076] LMCache INFO: Stored 512 tokens in 0.003 seconds (server.py:515:lmcache.v1.multiprocess.server
> ```
> 
> Is there's something wrong with my host command?

Follow up, this may due to the cache key bug reported in https://github.com/vllm-project/vllm/issues/42948

yoo-kumaneko:
> Follow up, this may due to the cache key bug reported in [vllm-project/vllm#42948](https://github.com/vllm-project/vllm/issues/42948)

Let me see. Is that a problem related to a specific vLLM version range? I'm using 0.19.2rc1.dev220+g7b1bc0a3e and it seems to work fine

yoo-kumaneko:
**Update (commit 7).** Replaced the bare wrap-fix commit with a full hybrid-on port to the LMCache-side `lmcache_mp_connector.py` — the in-tree copy added by #3235.

That copy was forked off the pre-hybrid vLLM connector (no `SupportsHMA`, flat `block_ids`), so it was incompatible with this PR's per-gid wire format. Commit `298e4382` ports the full hybrid-on path: `SupportsHMA` inheritance, per-gid request tracker, per-gid `LoadStoreOp` slicing with the divide-or-multiply rule for the GCD-collapsed `cache_config.block_size`, `register_kv_caches` forwarding `extra_layout_hints`, and `request_finished_all_groups`. The new file is byte-for-byte equivalent to the vLLM-side connector modulo LMCache import-grouping conventions and `torch_dev`.

Use via:
```
--kv-transfer-config '{"kv_connector":"LMCacheMPConnector",
                       "kv_connector_module_path":
                         "lmcache.integration.vllm.lmcache_mp_connector",
                       ...}'
```

The `lmcache_mp_connector_0180.py` / `_0201.py` snapshots are intentionally untouched — those vLLM revisions can't drive HMA at all.

DCO green; 369 multiprocess + 34 layer-groups tests pass.

yoo-kumaneko:
**Update (commit 8).** Promoted SWA-suffix-only to default. Commit `d0f9940d` drops the `LMCACHE_SWA_SUFFIX_ONLY=1` env-var gate and unconditionally publishes `per_layer_sliding_window`.

Re-verified at HEAD:
- `partial_hit_harness`: token IDs identical, max\|Δlogprob\| = 5.04e-04 (well within #13 noise floor)
- 9216-tok cold/warm: 1321.6 → 573.2 ms (**2.31×**)
- 30000-tok cold/warm: 2862.0 → 173.7 ms (**16.47×**)

Full-attention groups (`sliding_window == 0`) are byte-for-byte unaffected. No off-switch env var to keep the surface area small. Final 8-commit shape; 403 tests pass.

yoo-kumaneko:
Screenshot 2026-05-20 at 00 33 57

Tested and it runs fine — results shown above. There are quite a few changes, so I'd suggest we begin with a quick high-level review and dive into details in the second or third round.

yoo-kumaneko:
The TTFT speedup comparison against APC is shown below.
Clipboard_Screenshot_1779258844

Note that the test was conducted on Nvidia H20, which is notoriously weak at prefill compute.

lmcache/v1/multiprocess/server.py
```diff
@@ -386,35 +402,83 @@ def store(
...
                         lmc_ops.multi_layer_block_kv_transfer(
                             group_kv_pointers,
                             [tmp_buffer.data_ptr()],
                             chunk_block_ids_gpu,
                             gpu_context.device,
                             lmc_ops.TransferDirection.D2H,
                             gpu_context.get_shape_desc(group_idx),
-                            group_lmcache_chunk_size,
+                            kernel_chunk_tokens,
                             gpu_context.gpu_kv_format_,
```
KuntaiDu, line 709: Will this work if different attention layers have different `gpu_kv_format_`?

lmcache/v1/kv_layer_groups.py
```diff
@@ -216,15 +338,102 @@ def __init__(
...
 
+        # Resolve the per-layer logical block size hint, if present.
+        # When absent, every layer's logical block size will default to
+        # its physical ``bs`` (effectively reproducing the prior
+        # 5-tuple grouping behavior on engines that don't supply the
+        # hint). When present, the per-layer values participate in
+        # :data:`LayerGroupIdentity` so layers with same physical
+        # shape but different scheduler grids land in distinct LMCache
+        # groups (DeepSeek-V4 hybrid manager active).
+        per_layer_logical_bs: "list[int] | None" = None
```
KuntaiDu, line 251: Naming-wise I am not a fan of `logical` vs `physical`, because it is a bit hand-wavy. I don't have a good alternative and you don't have to fix it in this PR (but if you have a fix that would be great).

Maybe `raw_token_bs` v.s. `comp_token_bs`?

yoo-kumaneko: Hmm, I’m not very good at variable naming either. I’m okay with raw_token_bs vs. comp_token_bs.
What do you guys think? @ApostaC @sammshen  @maobaolong

yoo-kumaneko: Claude comes up with three alternatives:

  - scheduler_block_size vs kernel_block_size
  - scheduler_block_size vs storage_block_size
  - addr_block_size vs layout_block_size

yoo-kumaneko: I'd lean toward updating the variable names in a follow-up PR.

lmcache/v1/multiprocess/server.py
yoo-kumaneko: No, it will not work if attention layers have different `gpu_kv_format_`. 

For the DeepSeek v4 case, this is fine because all layers share the same MLA format. We currently assume and check that gpu_kv_format_ is invariant across all layers.

normalize_kv_and_discover_format reads the first tensor it finds and returns a single gpu_kv_format for the entire engine. There is currently no per-layer format detection in LMCache. Supporting that would require fairly broad changes across the codebase, so I’d recommend handling it in a separate PR.

ApostaC:
@yoo-kumaneko Thanks for the PR! I also wanted to run some experiments locally and get a better sense of the implementation.
In the meantime, is it possible to simplify the docstrings and comments? Right now it's a bit too much and noisy.

yoo-kumaneko:
> @yoo-kumaneko Thanks for the PR! I also wanted to run some experiments locally and get a better sense of the implementation. In the meantime, is it possible to simplify the docstrings and comments? Right now it's a bit too much and noisy.

Definitely. I'll clean them up

lmcache/v1/kv_layer_groups.py
```diff
@@ -35,11 +35,56 @@
...
-# :meth:`KVLayerGroupsManager.__init__` for the derivation.
-LayerGroupIdentity = tuple[int, int, int, int, torch.dtype]
+# The 8-tuple that uniquely identifies a set of kernel-equivalent layers:
+# ``(kv_size, num_heads, head_size, block_size, logical_block_size,
+# kv_cache_group_id, sliding_window, dtype)``. Two layers share a
+# transfer-kernel launch iff they share this identity — see the
+# grouping loop in :meth:`KVLayerGroupsManager.__init__` for the
+# derivation.
+#
+# The fifth slot (``logical_block_size``) is the *scheduler-side*
```
sammshen: so both the old block size and the new logical_block_size are needed? if only one of them is used to index the transfer, can we just keep that one?

sammshen:
This looks great, thanks for using the hints interface! Will provide more specific feedback tomorrow :)

yoo-kumaneko:
docstrings and comments are trimmed.
Rebased. conflicts resolved.

lmcache/v1/kv_layer_groups.py
yoo-kumaneko: Yes, both are needed — they only differ when compress_ratio > 1.

`shape_desc.bs` is the physical count: tokens actually stored per block. For a block of size
256 and compress_ratio=4, that's 64 stored tokens per block, which is what the transfer
kernel strides on.

`logical_block_size` (or maybe a better name like block size before compression) is the scheduler count: the engine's KVCacheSpec.block_size, which is what `LoadStoreOp.block_ids` is keyed on. We need it for the chunking math — `blocks_per_chunk` = `lmcache_chunk_size` // `logical_block_size` — because the chunk boundary is defined in the number of tokens before compression, not after it. 

So the kernel uses bs directly, and chunk-level dispatch uses logical_block_size.

yoo-kumaneko:
Closing in favor of #3550 (`feat/connector-side-swa-trim`).

This PR predates the upstream hybrid-memory-allocator framework (#3419, merged), which redesigned the HMA interface and made this PR's hand-rolled 8-tuple identity / wire format obsolete. The contiguity blocker and `block_stride_bytes` issue this PR worked around are now fixed upstream.

The work has been rebased onto #3419's framework in #3550, which adds DeepSeek-V4 per-group logical-block-size handling plus connector-side SWA-suffix trimming. Thanks to everyone who reviewed here — the discussion carried directly into the new PR.

Validated end-to-end on DeepSeek-V4-Flash (tp=4, fp8); cold≈warm semantically identical, ~30% external prefix-cache hit rate.

PR #3268, opened 2026-05-13
feat: add vllm DP support

zhuangqh:

Under DP, vLLM's gpu_worker shifts each worker's local device by `dp_local_rank * tp_size * pp_size` so DP replica k binds to GPUs [k*W, (k+1)*W). LMCache however used parallel_config.rank directly as local_worker_id and then called torch.cuda.set_device, which left LMCache on cuda:0 while the DP>0 worker was on cuda:1+. The mismatched device for CUDA events/streams produced:

  RuntimeError: Event device index does not match recording stream's
  device index

Mirror vLLM's adjustment by adding compute_dp_device_offset(), gated on the same conditions vLLM uses (single-node-per-DP, non-Ray / non-external_launcher executor backend). For ray / external_launcher the runtime already isolates each worker to a single visible device, so no offset is applied. Both call sites (create_lmcache_metadata, calculate_local_rank_and_world_size) now share this helper.

Add unit tests covering DP enabled/disabled, TP-only, TP+PP, Ray, external_launcher, Ray DP backend, multi-node DP, and missing optional attributes on older vLLM versions.

Verified end-to-end with vLLM data parallelism (DP); works as expected.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -372,10 +372,10 @@ def compute_dp_device_offset(parallel_config: "ParallelConfig") -> int:
     index by ``dp_local_rank * tp_size * pp_size`` so that DP replica ``k``
     on a single node binds to GPUs ``[k*W, (k+1)*W)`` where
     ``W = tp_size * pp_size``. LMCache must apply the **same** offset when
-    it calls ``torch.cuda.set_device`` so its buffers / events / streams
-    live on the same physical GPU as the corresponding vLLM worker —
-    otherwise CUDA operations fail with ``Event device index does not
-    match recording stream's device index``.
+    it sets the active device so its buffers / events / streams live on
+    the same physical GPU as the corresponding vLLM worker — otherwise
+    CUDA operations fail with ``Event device index does not match
+    recording stream's device index``.
 
     However, vLLM only performs this shift in a specific configuration:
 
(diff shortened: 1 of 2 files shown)


lmcache/integration/vllm/utils.py
```diff
@@ -264,13 +264,17 @@ def create_lmcache_metadata(
                 kv_transfer_config, "kv_connector_extra_config", None
             )
 
+    # Compute local_worker_id with DP offset so the metadata reflects the
+    # actual local GPU device (matches vLLM's gpu_worker DP adjustment).
+    local_worker_id = parallel_cfg.rank + compute_dp_device_offset(parallel_cfg)
```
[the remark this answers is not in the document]
zhuangqh: it's intended.

KuntaiDu, line 269: Will this cause a DP worker not being able to hit another DP worker's KV cache?

sammshen: I think it is not an issue since this PR only modifies `local_worker_id` and it is the global `worker_id` that is infused into the CacheEngineKey @zhuangqh

sammshen (approved):
LGTM, great question by kuntai but it is safe as described above

lmcache/integration/vllm/utils.py
zhuangqh: @sammshen yes. agree with you. Thanks for clarification.

zhuangqh:
@sammshen UT now should be passing. PTAL.

zhuangqh:
Friendly ping — PTAL when you have a moment 🙏 cc @ApostaC  @YaoJiayi  @sammshen

PR #3119, opened 2026-04-23
[MP] Add raw_block MP L2 adapter support via shared RawBlockCore

DongDongJu:
This PR adds multiprocess `raw_block` L2 adapter support without changing the MP request flow.

It extracts a shared `RawBlockCore` from the original `RustRawBlockBackend` and reuses the existing raw-device Rust I/O path and on-device metadata/checkpoint format, adds a built-in `RawBlockL2Adapter` for `--l2-adapter
  '{"type":"raw_block", ...}'`, and rewires the non-MP backend to the shared core while preserving the existing non-MP prefix semantics.

  Concretely, this PR:
  - adds `RawBlockCore` and shared key codec under `lmcache/v1/storage_backend/raw_block/`
  - adds `RawBlockL2Adapter` plus config/factory registration for MP mode
  - keeps restart recovery semantics and locked-entry handling inside the shared core
  - preserves plain raw-block behavior only: no FDP, no placement hints, no NVMe control path
  - updates MP/raw-block/operator docs and adds targeted tests for the adapter, factory, and non-MP wrapper

  **Notes for reviewers**:

  This is intentionally the minimum MP integration cut for raw-block:

  Validation:
  - Focused regression/unit tests:
    `python -m pytest -q tests/v1/distributed/test_mock_l2_adapter.py tests/v1/distributed/test_l2_adapter_factory.py tests/v1/distributed/test_raw_block_l2_adapter.py tests/v1/distributed/test_native_connector_l2_adapter.py`
    Result: `170 passed, 67 warnings`
  - MP e2e on raw block:
    Model: `Qwen/Qwen2.5-14B-Instruct`
    vLLM: `0.19.1`
    TP: `2`
    L2: `raw_block` on `/dev/nvme5n1p1`
    Store policy: `skip_l1`
    Benchmark: `benchmarks/long_doc_qa/long_doc_qa.py`
    Result: warmup `6/6` success, query `6/6` success
    Warmup mean TTFT: `0.555s`
    Warmup round time: `5.317s`
    Query mean TTFT: `1.215s`
    Query round time: `7.051s` ( It seems we have bunch of room to optimize this path)
    L2 evidence: LMCache logged `78/78 prefix hits (0 L1, 78 L2)` on the query-round prefetches, and final adapter status reported `indexed_key_count=508` on `/dev/nvme5n1p1`
  - For the local vLLM `0.19.1` validation above, the connector was launched with `kv_connector_extra_config={"lmcache.mp.port":6555}`

[...]

PR #3274, opened 2026-05-13
Missing io_uring changes and introducing nvme io_uring_cmd (passthrough) changes

ankit-sam:
The io_uring changes during rebase were omitted during the MP mode integration
. This commit partially adds them back.
Fixes: https://github.com/LMCache/LMCache/pull/3119

This series also adds nvme io_uring_cmd support, which enables asynchronous, low-latency passthrough of NVMe commands directly from application, bypassing file system and most of block layer overhead.

Introduced in Linux kernel 5.19, it allows using `IORING_OP_URING_CMD` for raw NVMe commands, featuring big SQE (128 bytes) for NVMe command structures, and / big CQE (32 bytes) for custom completions.

Unlike block device interface it requires nvme namespace character device (/dev/ngXnY)

1. The write request batching is only done for headers and payloads, since for MP mode we need to order the requests as they may be sent together. This will be fixed later.
2. Fixed buffer support has not been enabled with MP mode, and will be added soon.
3. The NVMe io_uring_cmd support only works for the kernel specified `/sys/block/nvmeXnY/queue/max_hw_sectors_kb` limit. This may be smaller than the device reported `max_data_transfer_size` limit, and can be increased once hugepage support is merged.

**If applicable**:


rust/raw_block/src/lib.rs
```diff
@@ -712,7 +764,12 @@ impl RawBlockDevice {
             next_batch_id_opt,
             batch_in_flight_opt,
         ) = if use_iouring {
-            let ring = IoUring::new(iouring_queue_depth as u32)
+            // Create IoUring with Entry32 CQE type when using io_uring_cmd
+            // to support big CQE entries (32 bytes instead of 16 bytes)
+            // Create IoUring with big SQE entries to support Entry128 (128-bytes SQE)
+            // for NVMe passthrough commands
+            let ring = IoUring::<Entry128, Entry32>::builder()
```
daegyu94: Big SQE/CQE is now used for plain io_uring even when io_uring_cmd is disabled. Do we still need to consider compatibility with older kernels?

lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -232,6 +255,28 @@ def __init__(
...
+                raise ValueError(
+                    "use_uring_cmd requires an existing NVMe namespace "
+                    f"character device path, got {self.device_path!r}"
+                ) from e
+            if not stat.S_ISCHR(mode):
+                raise ValueError(
+                    "use_uring_cmd requires an NVMe namespace character device "
+                    f"(for example /dev/ng0n1), got {self.device_path!r}"
+                )
+
```
daegyu94, line 279: would it make sense to validate the NVMe generic namespace naming pattern in addition to checking whether the path is a character device (e.g., basename `ng*` checking)?
My understanding is that the `ng*` prefix is fairly stable and unlikely to change unless the kernel naming convention itself changes.

lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -287,6 +332,40 @@ def __init__(
             self._cleanup_after_init_failure()
             raise
 
+    def _resolve_max_data_transfer_size(self, configured_size: int) -> int:
+        """Resolve transfer split size from config or NVMe sysfs queue limits."""
+        if configured_size > 0:
```
daegyu94, line 369: can we validate `configured_size` to ensure it is block-aligned?

lmcache/v1/storage_backend/raw_block/core.py
ankit-sam: Done, thanks for the suggestion

lmcache/v1/storage_backend/raw_block/core.py
ankit-sam: Done

rust/raw_block/src/lib.rs
ankit-sam: Yes, thanks for pointing out modified it

ankit-sam:
Hello @DongDongJu can you review the changes.

rust/raw_block/src/lib.rs
```diff
@@ -396,32 +723,74 @@ struct RawBlockDevice {
...
+        }
+        let cpath =
+            CString::new(path.clone()).map_err(|_| PyValueError::new_err("path contains NUL"))?;
         let mut flags = if writable {
             libc::O_RDWR
         } else {
             libc::O_RDONLY
         };
-        if use_odirect {
+        if use_odirect && !use_uring_cmd {
```
DongDongJu, line 990: Can non-4096-byte payload work with uring_cmd now? it seems we depends on lot of raw block logic with o_direct. WDYT?

rust/raw_block/src/lib.rs
```diff
@@ -459,6 +860,213 @@ impl RawBlockDevice {
...
+            }
+
+            // Helper function to decrement in-flight counts and notify condition variables
+            fn decrement_in_flight(
+                in_flight_count: &Arc<AtomicU64>,
+                in_flight_cvar: &Arc<Condvar>,
+                batch_in_flight: &Arc<Mutex<HashMap<u64, BatchTracking>>>,
+                batch_id: u64,
+            ) {
+                let prev = in_flight_count.fetch_sub(1, Ordering::Relaxed);
```
DongDongJu, line 1217: This refactor should make each submitted request decrement the global count exactly once.

tests/v1/storage_backend/test_raw_block_uring_cmd.py
```diff
@@ -0,0 +1,278 @@
...
+            local_cpu_backend=local_cpu_backend,
+            loop=loop_in_thread,
+        )
+
+
+def test_uring_cmd_get_nvme_info(loop_in_thread):
+    """Test getting NVMe namespace ID and LBA size from character device."""
+    # This test requires a block device device /dev/nvme0n1
+    # Skip if this doesn't exist
+    device_path = os.environ.get("LMCACHE_TEST_BLOCK_DEVICE", "/dev/ng0n1")
```
DongDongJu: Tests are too hard-coded with this one (/dev/ng0n1). please change it to param

DongDongJu (changes requested):
Hello @ankit-sam,

Thanks for bringing the missing MP io_uring changes back and for adding the NVMe io_uring_cmd path. This is aligned with the raw-block direction and should be useful for reducing overhead on NVMe devices.
Please check my few comments.

rust/raw_block/src/lib.rs
```diff
@@ -164,6 +306,170 @@ fn fd_size_bytes(fd: RawFd) -> Result<u64, PyErr> {
...
+    nsid: u32,
+    offset: u64,
+    len: usize,
+    lba_shift: u32,
+    ptr: *const u8,
+    dtype: u8,
+    dspec: u16,
+) -> Result<(), PyErr> {
+    // Calculate SLBA (Starting LBA) and NLB (Number of LBAs)
+    let slba = offset >> lba_shift;
```
DongDongJu, line 465: This needs to reject unaligned byte ranges before converting to SLBA/NLB. I reproduced this on `/dev/ng1n1`: `read_uring(1, ...)` returned the same first-block data as `read_uring(0, ...)` because `offset >> lba_shift` silently rounds the byte offset down to the previous LBA. For writes, that can target the wrong LBA and corrupt data. Please validate `offset % lba_size == 0`, `len % lba_size == 0`, nonzero length, and that the computed NLB fits the NVMe field before filling the command.

lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -197,6 +221,13 @@ def __init__(
         self.meta_verify_on_load = bool(config.meta_verify_on_load)
         self.io_engine = normalize_raw_block_io_engine(config.io_engine)
         self.iouring_queue_depth = int(config.iouring_queue_depth)
+        self.use_uring_cmd = bool(config.use_uring_cmd)
+        if self.use_uring_cmd and self.use_odirect:
+            logger.warning(
+                "RawBlockCore: use_odirect is ignored for NVMe namespace "
+                "character devices when use_uring_cmd=true"
+            )
+            self.use_odirect = False
```
DongDongJu, line 230: Disabling `use_odirect` here also disables the existing block-size padding logic, but `io_uring_cmd` still requires block-aligned transfer lengths. I reproduced this with a 3-byte payload: after init `use_odirect` is `False`, `_prepare_write_payload()` returns `total_len=3`, and `_write_uring_cmd_buffers()` raises `io_uring_cmd requires aligned transfer lengths`. Please keep a separate transfer-alignment condition for `use_odirect or use_uring_cmd`, so payload/header/meta reads and writes are padded even when O_DIRECT itself is ignored for the character device.

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
```diff
@@ -669,13 +687,31 @@ def main() -> None:
         action="store_true",
         help="Enable io_uring for raw block backend",
     )
+    parser.add_argument(
+        "--use-uring-cmd",
+        action="store_true",
+        help=(
+            "Enable io_uring for raw block backend. "
```
DongDongJu: The help says `--use-uring-cmd` enables io_uring, but the config still passes `use_uring=args.use_uring`. Running the benchmark with only `--use-uring-cmd` fails with `ValueError: use_uring_cmd requires io_uring as io_engine`. Please either make `--use-uring-cmd` imply `--use-uring` after parsing, or update the help/error path so users know both flags are required.

DongDongJu (changes requested):
I tested this PR locally from the exact head SHA with a clean `uv` install and `/dev/ng1n1`. The install/build path works and read-only NVMe io_uring_cmd opens/reads successfully, but I found a few correctness issues that should be fixed before merge.

rust/raw_block/src/lib.rs
ankit-sam: Done

tests/v1/storage_backend/test_raw_block_uring_cmd.py
ankit-sam: I moved all the block, character and null devices to a global location which can be changed by setting environment variable.

rust/raw_block/src/lib.rs
ankit-sam: Thanks changed

lmcache/v1/storage_backend/raw_block/core.py
ankit-sam: Yes added an alignment check

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
ankit-sam: Done

ankit-sam:
https://github.com/cursor review

rust/raw_block/src/lib.rs
ankit-sam: Hmm, it seems a little bit tricky. We also have the issue of non-aligned data buffers. With fixed buffers it is ok as it will be page aligned. Let me think about that

DongDongJu (approved):
Hello @ankit-sam, Thanks for pushing this forward.
The Python core path looks clear now because `use_uring_cmd` has its own transfer-alignment condition, and device/transfer-limit validation is much clearer. 
LGTM

DongDongJu:
@ApostaC Could you take a look?

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
ApostaC, line 1: Should we update the docs (at least the README.md in the folder)?

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
```diff
@@ -110,6 +112,8 @@ def __init__(
             enable_zero_copy: Whether to use aligned direct-buffer I/O.
             io_engine: Raw-block I/O engine: ``"posix"`` or ``"io_uring"``.
             iouring_queue_depth: Queue depth for the Rust io_uring engine.
+            use_uring_cmd: Whether to use NVMe io_uring_cmd passthrough.
+            max_data_transfer_size: Max data transfer size for a single request.
```
ApostaC, line 116: Since we are introducing new configuration items for the L2 adapter, can we update the docs in `docs/source/mp` (see https://docs.lmcache.ai/mp/l2_storage.html#raw-block-raw-block-device-backed-persistent-storage)

ApostaC (commented):
Thanks for the contribution! Left 2 comments about the doc consistency. PTAL

benchmarks/storage_backend_io/storage_backend_io_benchmark.py
ankit-sam: Updated thanks

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
ankit-sam: Done thanks

sammshen (approved):
LGTM!

lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
```diff
@@ -406,6 +448,53 @@ async def _submit_put_one(
             with self._put_lock:
                 self._put_tasks.discard(key)
 
+    async def _submit_put_many(
```
sammshen, line 451: nit: can we fire callbacks for succeeded keys before surfacing the failure

PR #3277, opened 2026-05-14
[Feat][MP][1/N] Asymmetric Serde Support (eg., FP16 Key , FP8 Value)

mcgrof:
This pull request extends the serdes to add support where the KV cache may requires different data types or you may only want to cache for example values instead of both keys and values. The use case for both is the asymmetric KV cache quantization FP16 keys, and FP8 values, built on extending flashinfer as documented on knlp [0]. It also allows folks to test and leverage asymmetric quantization by keeping keys CPU pinned and only moving to L2 values as a new type of caching tier. This work rebased that work on top of the serdes implementation by Kuntai.

I'll be working on each upstream component separately.

[0] https://knlp.io/decode

- [x ] this PR contains user facing changes - docs added


lmcache/v1/distributed/serde/multi.py
```diff
@@ -0,0 +1,289 @@
...
+  V at another (e.g. FP8) cannot carry both -- one of them would
+  have to be reinterpreted bytewise, losing the typed-tensor
+  contract every other consumer relies on.
+
+* **One side absent.** For tier-split placements where K or V is
+  held outside this serde's data path -- e.g. K kept in L1 (CPU
+  pinned host memory) while V flows to L2 (durable storage) -- the
+  serialize input has no tensor for the absent slot and the
+  deserialize output has no destination for it.
+
```
KuntaiDu, line 23: Just want to further understand the design goal of multiserializer/multideserializer
Regarding these two cases:
* If we want to quantize K and V to different dtype --- we can serialize them to bytestreams and then write to the target `MemObj`. Will the multi-output design provides extra benefit than this?
* Regarding one-side absent: just wondering if allowing K and V of the same set of tokens to be stored in different places provide benefit.

KuntaiDu: Oh I see -- does this multiserializer/deserializer makes the code simpler when the dtype or the processing scheme of K and V diverges?

mcgrof: > Just want to further understand the design goal of multiserializer/multideserializer Regarding these two cases:
> 
> * If we want to quantize K and V to different dtype --- we can serialize them to bytestreams and then write to the target `MemObj`. Will the multi-output design provides extra benefit than this?

Serializing can surely works at the byte level but loses the typed-tensor contract everywhere else, and so  you would give up typed-tensor guarantees on the API surface. So there are two problems:

1) A MemoryObj carries one dtype + one shape. To shove K (fp16) and V (fp8) into one MemoryObj we'd either lie about the dtype (treat it as opaque uint8 and have the codec slice internally) or invent a two-dtype MemoryObj subclass. The first breaks the factory/wrapper/buffer-allocator machinery downstream — they think they're working with a typed tensor. The second is basically multi-output but with extra wrapping.

2) In my prior work before rebasing onto the new SerDes I did work on a solution like what you suggest  with a .asym_view side-channel attribute on the input MemoryObj. It worked but it's an undocumented sneak-channel that bypasses the type system. Multi-output is the typed, contracted version of the same thing.

> * Regarding one-side absent: just wondering if allowing K and V of the same set of tokens to be stored in different places provide benefit.

Yes, for the one-side-absent case I specifically mean the split-tier / V-only path, that has a real measured benefit. The benefit is reducing L2 traffic and capacity pressure. For K16/V8, writing only V8 means the L2 payload is 1/3 of the full asymmetric K16+V8 payload, or 1/4 compared with regular FP16 K+V. As to when you'd enable it, that would be useful if L2 bandwidth is the bottleneck and you're willing to pay L1 capacity for the K side. As for a full set of profiles of when this is useful / likely, I'm still working on those evaluations, but so far cases I am evaluating are high-concurrency short-output workloads, bursty admission pressure, prefix/offload cases where K can stay hot in L1, and speculative/draft-cache paths where writing full K+V may be wasted.

In the future I do expect this will be also useful to enable future memory tiers as well. For example, a future placement policy could keep K in a faster memory tier, such as L1 / CPU / CXL, while storing V in a colder tier, such as NVMe or another L2 backend. That gives the policy a way to choose different placement for the two slots instead of forcing K and V for the same token range to move as one opaque object.

So the design goal for one-side-absent is to make that split-tier placement explicit: one side stays in a higher tier, while the other side flows through the L2 serde.

mcgrof: > Oh I see -- does this multiserializer/deserializer makes the code simpler when the dtype or the processing scheme of K and V diverges?

yes, multi-output makes the code simpler for the consumers of the serde subsystem (factory, wrapper, allocator), not just the serde itself. But you're the maintainer so happy to yield to your preference / advice!

KuntaiDu: Thank you for the concrete details. It surfaces the typing up to API level and that is really useful. This part of design LGTM now.

lmcache/non_cuda_equivalents.py
```diff
@@ -278,35 +275,23 @@ class GPUKVFormat(IntEnum):
     # used by: TRT-LLM cross-layer (HND layout)
     NB_NL_TWO_NH_BS_HS = 8
 
+    # used by: vLLM asymmetric K/V (V-only split-tier).  Per-layer K and
+    # V live in separate 4D tensors at potentially different dtypes
+    # (e.g. K=bf16, V=fp8_e4m3); wrap_kv_caches flattens
+    # [(K0, V0), (K1, V1), ...] into a single list and the (K, V)
+    # pairing is implicit in K-then-V ordering.
+    NL_X_TWO_PER_LAYER_NB_BS_NH_HS_ASYM = 9
```
KuntaiDu: Regarding this part ---- this part of code is specifically for GPU KV cache format detection. Will the current serde path push serialized KV cache into GPU?

KuntaiDu (changes requested):
The scope of current serde interface makes 2 assumptions:
* All data that will be moved into storage must be serialized
* All data that will be moved to GPU must be deserialized.
Current PR seems to push serialized data into GPU. Can you remove that part of logic?

We have some design of how to push the serialization benefit further into GPU, but that will be through a different set of API.

lmcache/non_cuda_equivalents.py
mcgrof: Sorry this was pushed prematurely. I will drop it!

mcgrof:
> The scope of current serde interface makes 2 assumptions:
> 
> * All data that will be moved into storage must be serialized
> * All data that will be moved to GPU must be deserialized.
>   Current PR seems to push serialized data into GPU. Can you remove that part of logic?
> 
> We have some design of how to push the serialization benefit further into GPU, but that will be through a different set of API.

Sorry about that, that extra stuff was pushed by mistake. The pull request is slimmed to the 3 CPU-side serde commits.

lmcache/v1/distributed/serde/asym_k16_v8.py
```diff
@@ -0,0 +1,538 @@
...
+                f"scale_scope {scope} requires per-head/per-page metadata "
+                f"that MemoryLayoutDesc does not carry; only PER_TENSOR "
+                f"and EXTERNAL are supported here"
+            )
+        k_bytes = sum(
+            math.prod(s) * d.itemsize
+            for s, d in zip(k_layout.shapes, k_layout.dtypes, strict=True)
+        )
+        v_bytes = sum(math.prod(s) for s in v_layout.shapes)
+        header_allowance = 1024
```
DongDongJu, line 173: I know its temp value. but L2 will store whole this value.
For small chunks, the 1 KB allowance can materially reduce or erase the storage savings.
Id recommend either compute the header size exactly, or add a wrapper/interface change that propagates the actual n into the L2 store length.

DongDongJu (commented):
Hello @mcgrof,

Thanks for pushing this. The multi-output API direction LGTM conceptually, especially for surfacing K/V asymmetry at the typed API level instead of hiding it in an opaque byte side-channel.

Before review it, I want to test it first with current MP.
How should I build the test for this one?

KuntaiDu (approved):
LGTM!

KuntaiDu:
I'll also run some perf test on my side, but the current design and impl LGTM.

lmcache/v1/distributed/serde/asym_k16_v8.py
mcgrof: Good eye, I've fixed this here:

https://github.com/mcgrof/LMCache/commit/29207ec42d475f133c6bbee0c835ccb7b484a5a5

But I currently put it out as a separate commit given the review of this current branch, I didn't want to re-trigger a full review loop again. Let me know you guys want me to squash it.

mcgrof:
> Hello @mcgrof,
> 
> Thanks for pushing this. The multi-output API direction LGTM conceptually, especially for surfacing K/V asymmetry at the typed API level instead of hiding it in an opaque byte side-channel.
> 
> Before review it, I want to test it first with current MP. How should I build the test for this one?

#3277 is intentionally the **CPU-side typed building blocks** — the `MultiSerializer` / `MultiDeserializer` extension plus the `AsymK16V8` storage-path and V-only serdes. The right test for it is a **contract test**, not a full vLLM/MP model run. It's on the branch:                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
```                                                                                                                                                                                                                                          
https://github.com/mcgrof/LMCache/tree/serde-multi-output-mp-contract                                                                                                                                                                        
pytest tests/v1/distributed/serde/test_multi_output_mp_contract.py -q -rxX                                                                                                                                                                   
```
A small fake multi-output codec emits two named outputs (`k` / `v`, distinct sizes + sentinels so ordering/aliasing bugs can't hide; `group_size == 2` to match the upstream shape, scales packed in V's header) and the test pins: typed round-trip, the absent-K (split-tier) slot semantics, single-output backward compat through the length-one bridge, and the split-tier byte accounting (V-only L2 = 1/3 of K16/V8 = 1/4 of FP16). Worth noting `AsyncSerdeProcessor` already round-trips a multi-output group unchanged — the only piece that needed wiring was L2 tier-placement of the separate child outputs.                                                                                                              
                                                                                                                                                                                                                                             
The rest of `serde-multi-output-mp-contract` is my current work-in-progress on the MP integration — the split-tier wiring (multi-output dispatch, component-groups L1 layout, the split-tier store/load/manifest/eviction stack) stacked on top of #3277, each commit scoped and tested. The validated win there is storage: V-only writes 1/3 the L2 bytes (a byte/endurance result, not a throughput claim), with K kept bit-exact in L1. It's not part of this PR -- that full branch is just current work in progress of what that could look like. You can take a sneak peak and if you see anything you'd like to be done differently let me know! Otherwise I'll chug on and try to finalize that for a follow up  pull request after this one.

DongDongJu (approved):
LGTM, Could you open the issue about write up the PR plan to eventually support mp mode?

lmcache/v1/distributed/serde/asym_k16_v8.py
mcgrof: Since I had to fix a coding formatting issue, I decided to just fold the fix from https://github.com/mcgrof/LMCache/commit/29207ec42d475f133c6bbee0c835ccb7b484a5a5 as applied to the 3 commit only, this is folded onto commit 2 ("AsymK16V8 multi-output serde (L2 storage path)"), it added these 4 files but the MP-contract test was dropped:
                                                                                                                      
  - lmcache/v1/memory_management.py (+65) — the set_used_size fix
  - lmcache/v1/distributed/serde/async_processor.py (+14/−1) — propagates the real length
  - tests/v1/distributed/serde/test_async_processor.py (+89)                                                                                                                                                                                 
  - tests/v1/test_memory_management.py (+143)

mcgrof:
The [buildkite/k3-multiprocess-test](https://buildkite.com/lmcache/k3-multiprocess-test/builds/2777) failure was due to the new folded fixes addressing @DongDongJu 's feedback, I'll fix this and force push soon.

mcgrof:
> LGTM, Could you open the issue about write up the PR plan to eventually support mp mode?

Done ! https://github.com/LMCache/LMCache/issues/3710

PR #3278, opened 2026-05-14
fix: PD restore pin=True in PD sync backend dedup path

XuanCS:

Reverts `pin=False` back to `pin=True` in `PDBackend._allocate_and_put` (the dedup branch on the receiver side). This fixes the "Key not found" race introduced by #2847.

**The problem**:
- #2847 changed `pin=False` in `_allocate_and_put` to fix a cumulative buffer issue. But it introduced a "Key not found" race under concurrent shared-prefix workloads: when Request B hits the dedup path (`contains(key, pin=False)`), it does not bump the refcount, so Request A's later `remove(key)` deletes the entry. When Request B then calls `get_blocking(key)`, it asserts with `Key not found in local data.`

**The fix**:
- Restore `pin=True` so the deduplicated consumer increments the refcount on the dedup path. Sync `remove()` already does refcount-aware deletion (`if get_ref_count() == 1: del`), so a second consumer with `pin=True` correctly defers the delete until they've retrieved.
- The buffer-exhaustion behavior under sustained shared-prefix load with `pin=True` is the model's known limitation, not a leak — chunks are released at end-of-request via the existing `ref_count_down()` in `cache_engine.py`. Working set must fit within `pd_buffer_size` for sustained throughput.

**Scope**:
- This PR is revert in the sync backend.
- The async backend has the same shared-prefix race; that is being addressed separately in #3444 (which applies the same `pin=True` semantics on the async path).

**Reproduction**:
- Race: send 8+ concurrent diverse-prompt requests at C=8 with `pin=False` → 7/8 crash with "Key not found".
- Leak: send 250+ shared-prefix requests at C=8 with `pin=True` and 4GB PD buffer → "Failed to allocate memory object" warnings around request ~50-80.
- With this PR: both scenarios pass at 4GB buffer.

**Testing done**:
- Llama-3.3-70B-Instruct, 1P:1D on p5.48xlarge (8×H100), TP=8, InterCom shared-prefix dataset (12K-char system prompt)
- PD_BUFFER_SIZE: 4GB (intentionally small to expose the leak)

| Test | Requests | Concurrency | Result | TTFT P50/P99 |
|------|----------|-------------|--------|--------------|
| Leak | 250 | C=8 | 250/250 ✅ | 1.10s / 2.18s |
| Stress | 500 | C=8 | 500/500 ✅ | 1.09s / 2.31s |
| High-C | 200 | C=16 | 200/200 ✅ | 1.66s / 3.03s |
| Stress | 200 | C=32 | 200/200 ✅ | 2.15s / 4.84s |
| Race | 320 (10 rounds × 32 concurrent) | C=32 | 320/320 ✅ | 2.93s / 3.66s |

- Single 1-line change; net production code diff is `pin=False → pin=True`.
- Per discussion with @hlin99, this PR was trimmed from the earlier `_pending_consumers` approach down to just the revert. The async fix lives in #3444.

**If applicable**:
- [ ] this PR contains user facing changes


DongDongJu:
@XuanCS Thanks for the contribution. Could you check the DCO?

XuanCS:
@DongDongJu Updated commit message, pass DCO, thanks

DongDongJu (changes requested):
Hello @XuanCS, thanks for bring up.

This fixes the concurrent-consumer race, 
but it also increments ref_count for every already-existing key. 

How can we matching `ref_count_down()` for this additional consumer? 
Without that, this reintroduces the #2847 leak where remove() skips deletion because ref_count > 1. 

I think we need either an explicit reused-consumer release path or a test demonstrating that the current retrieve path balances this reference for all concurrent requests.

XuanCS:
Hi @DongDongJu   Thanks for the careful review you're right that the simple revert (the first commit) reintroduces the leak. For both concerns:

**Where the matching `ref_count_down()` lives:**
The existing `cache_engine.py:865` is the matching release path. After every `get_blocking()`, the cache engine calls `ref_count_down()` (via `for key, memory_obj, _, _ in reordered_chunks: ... memory_obj.ref_count_down()`). So every consumer (the original allocator AND each "already_sent" pinner) produces exactly one `ref_count_up` (in `_allocate_and_put`) paired with one `ref_count_down` (in cache_engine after `batched_to_gpu`). They balance 1:1.

**Why the simple revert still leaked even with that balance:**
The leak isn't a ref_count imbalance — it's that `remove()` checked `ref_count == 1`, but under sustained concurrent load `ref_count` oscillates between 1 and 2 and never hits 1 *at the moment `remove()` is called*. So the chunk stays in `self.data` forever even though the count is balanced.

**The fix:**
Decouple "deletion guard from `self.data`" from "memory lifecycle (free at rc=0)". A new `_pending_consumers: dict[Key, int]` counter tracks consumers between `_allocate_and_put` and `get_blocking`. `remove()` defers deletion while `_pending_consumers > 0`. ref_count keeps its existing role for memory lifecycle.

Trace under sustained load (e.g., new pin always before previous remove):
```
A allocates P:    rc=1, _pending_consumers[P]=1
B "already_sent": rc=2, _pending_consumers[P]=2
A's get_blocking: _pending_consumers[P]=1, returns mem_obj
A's remove:       _pending_consumers[P]=1 > 0 → skip delete
A's ref_count_down (cache_engine.py:865): rc=1
C "already_sent": rc=2, _pending_consumers[P]=2
B's get_blocking: _pending_consumers[P]=1
B's remove:       _pending_consumers[P]=1 > 0 → skip delete
B's ref_count_down: rc=1
... pattern continues under sustained load ...
[load drops]
last consumer's get_blocking: _pending_consumers[P]=0, popped from dict
last consumer's remove:       _pending_consumers[]=0, rc==1 → DELETE ✓
last consumer's ref_count_down: rc=0 → free ✓
```

**Test demonstrating the balance** (added in commit 3):
`test_eight_concurrent_consumers_balance` constructs 8 concurrent consumers for the same key, then runs the full retrieve cycle for each. Asserts that after all 8 retrieves: `ref_count == 0`, key removed from `self.data`, memory freed. This proves the existing retrieve path balances the reference for all concurrent requests.

Other tests in commit 3 cover: single-consumer (no regression), out-of-order retrieves, sustained load eventually cleaning up, and edge cases.

**Empirical verification on 70B + InterCom shared-prefix at 4GB PD_BUFFER:**
- 250 requests @ C=8: 250/250 ✓
- 500 requests @ C=8: 500/500 ✓
- 200 requests @ C=16: 200/200 ✓
- 200 requests @ C=32: 200/200 ✓
- 320-request race test: 320/320 ✓

0 "Failed to allocate" warnings, 0 "Key not found" assertions across 1470 total requests.

XuanCS:
Hi @DongDongJu @sammshen @hlin99 would you take a look when have a chance, thanks!

ApostaC:
Hey @hlin99 , can you also help review this PR? This is a fix to PD

hlin99:
hi @XuanCS  thanks for the analysis!

To help me better understand to issue
1. could you share your prefill.yaml and decode.yaml ? 
2. and i see your branch is 3 weeks old, do you mind to test on latest code base? make sure you set pd_backend_mode="sync", as latest codebase will go through async mode by default. （if you have bandwidth, testing on pd_backend_mode="async" is also welcomed and i would suppose the issue won't be there)
3. if you have a simple ut which can repro the issue is also very helpful. i can do some investigation locally. thanks.

I would acknowledge there might be some issue in the ref counter, but just gut feeling... we can fix the issue by properly handing the ref_counter.  For a key, ref counter is how many consumers are using it, and when it comes to 0, it can be deleted from dict, as well as memory free.

XuanCS:
Hi @hlin99 , this is  thanks for the review! I've rebased onto latest dev, and Here's the additional test coverage and info 

**1. Configs (prefill.yaml / decode.yaml)** Since we run this in operator, configMap we set:

Decoder config:
```yaml
enable_pd: True
transfer_channel: "nixl"
pd_role: "receiver"
pd_backend_mode: "sync"
pd_peer_host: "0.0.0.0"
pd_peer_init_port: [7300, 7301, 7302, 7303, 7304, 7305, 7306, 7307]
pd_peer_alloc_port: [7400, 7401, 7402, 7403, 7404, 7405, 7406, 7407]
pd_peer_query_port: [7500, 7501, 7502, 7503, 7504, 7505, 7506, 7507]
pd_buffer_size: 4194304000
pd_buffer_device: "cuda"
nixl_backends: [LIBFABRIC]
blocking_timeout_secs: 60
chunk_size: 256
local_cpu: True
max_local_cpu_size: 100.0
save_decode_cache: False
```

Prefiller config:
```yaml
enable_pd: True
transfer_channel: "nixl"
pd_role: "sender"
pd_backend_mode: "sync"
pd_peer_host: "<decoder-host>"
pd_peer_init_port: [7300, 7301, 7302, 7303, 7304, 7305, 7306, 7307]
pd_peer_alloc_port: [7400, 7401, 7402, 7403, 7404, 7405, 7406, 7407]
pd_skip_proxy_notification: true
pd_buffer_size: 4194304000
pd_buffer_device: "cuda"
nixl_backends: [LIBFABRIC]
blocking_timeout_secs: 60
chunk_size: 256
local_cpu: True
max_local_cpu_size: 100.0
save_decode_cache: False
```

vLLM args (both prefill and decode):
```
vllm serve <model> --tensor-parallel-size 8 --max-model-len 16384 --gpu-memory-utilization 0.75 --max-num-seqs 32 --no-enable-prefix-caching --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_consumer","kv_connector_extra_config":{"discard_partial_chunks":false}}'
```
(Use `kv_role: "kv_producer"` for prefiller, `"kv_consumer"` for decoder)

**2. Tested on latest codebase**

Rebased onto latest `origin/dev` (commit `5b71110a`, May 18). Tested with Llama-3.3-70B-Instruct, 1P:1D on p5.48xlarge (8×H100), TP=8, 4GB PD buffer, shared-prefix workload (~11K chars, 8+ concurrent requests).

| Image | pd_backend_mode | 250@C=8 | 500@C=8 | 200@C=16 | 200@C=32 |
|-------|----------------|---------|---------|----------|----------|
| origin/dev (no fix) | sync | ❌ **Crashed at 8/250** | — | — | — |
| origin/dev (no fix) | async | 250/250 ✅ | 500/500 ✅ | 200/200 ✅ | 200/200 ✅ |
| This PR (fix) | sync | 250/250 ✅ | 500/500 ✅ | 200/200 ✅ | 200/200 ✅ |

The crash with `pd_backend_mode="sync"` is: `RuntimeError: Key CacheEngineKey(...) not found in local data.` — triggered after just 8 concurrent requests with shared prefix.

As you suggested, `pd_backend_mode="async"` doesn't have this issue because `PDBackendAsync` doesn't use the dedup/already_sent logic. The fix is needed for the sync path.

Performance is identical between async and fix+sync (P50 ~1.01s at C=8, ~3.05s at C=32).

**3. Standalone UT reproducing the issue**

Added in commit `407db8af` (`tests/v1/storage_backend/test_pd_backend_race_repro.py`). Runs in <1s, pure Python, no GPU/cluster/vLLM needed:

```
$ python tests/v1/storage_backend/test_pd_backend_race_repro.py

Test 1: Buggy backend (pin=False) with 8 concurrent consumers...
  Results: 2 success, 6 crashes
  First crash: CRASH: Key shared_prefix_chunk_0 not found in local data.
  ❌ Bug reproduced: 'Key not found' race with pin=False

Test 2: Fixed backend (_pending_consumers) with 8 concurrent consumers...
  Results: 8 success, 0 crashes
  ✅ All consumers succeeded, no leak, no crash
```

**Re: fixing via ref_counter alone**

I agree ideally ref_count alone should track consumers and trigger cleanup at 0. The challenge is that in the cache_engine flow, `remove()` is called *before* `ref_count_down()` (line 860-865 in `cache_engine.py`). So at `remove()` time, ref_count is always ≥ 1 for the current consumer — we can't use "ref_count == 0 → delete" because that moment only arrives after `remove()` has already been called.

The `_pending_consumers` counter is essentially a "pre-remove consumer count" that tells `remove()` whether it's safe to delete. It's decoupled from ref_count so the memory lifecycle (free at rc=0) stays unchanged.

hlin99:
> Hi @DongDongJu Thanks for the careful review you're right that the simple revert (the first commit) reintroduces the leak. For both concerns:
> 
> **Where the matching `ref_count_down()` lives:** The existing `cache_engine.py:865` is the matching release path. After every `get_blocking()`, the cache engine calls `ref_count_down()` (via `for key, memory_obj, _, _ in reordered_chunks: ... memory_obj.ref_count_down()`). So every consumer (the original allocator AND each "already_sent" pinner) produces exactly one `ref_count_up` (in `_allocate_and_put`) paired with one `ref_count_down` (in cache_engine after `batched_to_gpu`). They balance 1:1.
> 
> **Why the simple revert still leaked even with that balance:** The leak isn't a ref_count imbalance — it's that `remove()` checked `ref_count == 1`, but under sustained concurrent load `ref_count` oscillates between 1 and 2 and never hits 1 _at the moment `remove()` is called_. So the chunk stays in `self.data` forever even though the count is balanced.
> 

hi @XuanCS thanks for raising this issue and detailed data!

I revisit the current code....and some thoughts to discuss

1. I think if a key is frequenctly hit by requests, the key is not released until those requests are done. this is expected behavior, do you agree?

2. there might be some logic changes after https://github.com/LMCache/LMCache/pull/2847, but looks like reverting https://github.com/LMCache/LMCache/pull/2847 can resolve both leak issue and key not found issue. can you help check? and if you still observe leak issue, can you help elaborate the scenario? thanks a lot

3. For the UT in your PR, suggest to test real implementation in lmcache. now, the UT defines implementation in the UT itself, and testing itself...  I did a UT in my repo, just for your reference. https://github.com/hlin99/LMCache/commit/769aaed1ebff3a65a795b844c5dbf1fc9649cd3e

XuanCS:
Hi @hlin99 thanks for review, I did more test using with the async as backend, using shared prefix like Our test uses the same long system prompt repeated across

  PREFIX = "You are a helpful AI assistant working at a technology company. Your job is to help debug distributed
  systems. " * 300  # ~6300 tokens shared
  SUFFIX = f" Question {idx}: Tell me about topic number {idx}
  in detail."  # unique per request

All requests share ~6300 tokens of prefix (same chunk_hash → same CacheEngineKey), with only the last few tokens
  differing.

We're hitting this crash **in the async backend** (`pd_backend_async.py`) which is the default in v0.4.5. The crash is:
```
AssertionError: Key CacheEngineKey(...) not found in local data.
```
at `pd_backend_async.py:1457` in `get_blocking()`.

**Reproduction:** Any shared-prefix workload at C≥2 on `lmcache/vllm-openai:v0.4.5-cu129` (tested on P5/H100, 70B model). C=2 crashes. So the changes covers the async path now


**1. Key not released until requests are done — yes, agreed.** That's the intended behavior. The key stays alive only until all concurrent consumers call `get_blocking()`, then `remove()` proceeds normally.

**2. Regarding reverting #2847:**

Reverting #2847 would only help the sync backend (via `pin=True` in `contains()`). The async backend has **no `contains()` call and no dedup** — `_async_allocate_and_put()` always allocates fresh memory and `put()` unconditionally overwrites. So reverting #2847 does not fix the async path.

Additionally, with pin=True we observed buffer exhaustion (16GB buffer fills) under sustained shared-prefix load because ref_count never decrements for reused keys.

`_pending_consumers` fixes both backends without leaking. **I've updated this PR to also patch `pd_backend_async.py`** with the same 4 changes.



**3. UT:** Added `test_pd_async_shared_prefix.py` which uses real `PDBackendAsync` methods (`_async_allocate_and_put`, `get_blocking`, `remove`, `put`) — only `__init__` and `allocate()` are mocked.

hlin99:
> Hi @hlin99 thanks for review, I did more test using with the async as backend, using shared prefix like Our test uses the same long system prompt repeated across
> 
> PREFIX = "You are a helpful AI assistant working at a technology company. Your job is to help debug distributed systems. " * 300 # ~6300 tokens shared SUFFIX = f" Question {idx}: Tell me about topic number {idx} in detail." # unique per request
> 
> All requests share ~6300 tokens of prefix (same chunk_hash → same CacheEngineKey), with only the last few tokens differing.
> 
> We're hitting this crash **in the async backend** (`pd_backend_async.py`) which is the default in v0.4.5. The crash is:
> 
> ```
> AssertionError: Key CacheEngineKey(...) not found in local data.
> ```
> 
> at `pd_backend_async.py:1457` in `get_blocking()`.

hi @XuanCS  thanks for the great deep dive data

let me try to understand the issue as first step. I did some investigation, and to me, below scenario may have issues in async mode. can you confirm if that's also what you've observed? thanks.

```
Decoder: Req A arrives, Req B arrives (both lookup K → hit)

Req A: get_blocking(K) → obj_B   (gets B's data, not A's — silent corruption)
Req A: remove(K) → data.pop(K), obj_B freed

Req B: get_blocking(K) → None → AssertionError CRASH
```

XuanCS:
Hi @hlin99 that's exactly the scenario we observed.

We add `_pending_consumers` because `remove()` needs to know if other requests are still waiting to consume the key. Without it, the first consumer's `remove()` deletes the entry while the second consumer hasn't called `get_blocking()` yet. `_pending_consumers` defers the deletion until all consumers have read the data.

hlin99:
> #2847

okay. Thanks for the confirmation.

Then let go back to sync mode.  just curious what issue does this PR resolve on top of reverting https://github.com/LMCache/LMCache/pull/2847

can you elaborate the scenario? thx.

XuanCS:
Hi @hlin99, this is scenarios,  

- **#2847**: Fixed `pin=True` causing ref_count leak (buffer slots never freed)
  - **#3278 (this PR)**: Fixes the race in scenario we discussed:

  > Req A: get_blocking(K) → obj_B
  > Req A: remove(K) → data.pop(K), obj_B freed
  > Req B: get_blocking(K) → AssertionError CRASH

  The crash in step 3 happens because `remove()` has no awareness that Req B is still a pending consumer. `_pending_consumers` tracks this:

  - When B's `_allocate_and_put` finds K exists → `_pending_consumers[K] += 1`
  - A's `remove(K)` sees `_pending_consumers[K] > 0` → defers deletion
  - B's `get_blocking(K)` succeeds, then decrements `_pending_consumers[K]`
  - After all consumers read, the next `remove(K)` actually deletes

  This race exists regardless of `pin=True` or `pin=False`, thanks!

hlin99:
> Hi @hlin99, this is scenarios,
> 
> * **[fix: use pin=False in _allocate_and_put to prevent pd_buffer leak #2847](https://github.com/LMCache/LMCache/pull/2847)**: Fixed `pin=True` causing ref_count leak (buffer slots never freed)
>   
>   * **[fix: PD backend deletion guard via pending_consumers  counter #3278](https://github.com/LMCache/LMCache/pull/3278) (this PR)**: Fixes the race in scenario we discussed:
>   
>   > Req A: get_blocking(K) → obj_B
>   > Req A: remove(K) → data.pop(K), obj_B freed
>   > Req B: get_blocking(K) → AssertionError CRASH
>   
>   
>   The crash in step 3 happens because `remove()` has no awareness that Req B is still a pending consumer. `_pending_consumers` tracks this:
>   
>   * When B's `_allocate_and_put` finds K exists → `_pending_consumers[K] += 1`
>   * A's `remove(K)` sees `_pending_consumers[K] > 0` → defers deletion
>   * B's `get_blocking(K)` succeeds, then decrements `_pending_consumers[K]`
>   * After all consumers read, the next `remove(K)` actually deletes
>   
>   This race exists regardless of `pin=True` or `pin=False`, thanks!

hi @XuanCS  thanks for the details. here's my understanding of the issue after compile all the stuff we have right now. let me know if anything we're not on the same page.

1. personally, I wouldn't call ref_count leak when set pin=True. it's true that a buffer can't be released when pin=True, but fundamentally it's due to the buffer is repeated reused by the data pattern.  as long as when all the requests are finished, the buffer can be freed, it's not a leak.
2. I would agree pd sync backend has some limitation when handling pd buffers, especially when the requested buffers > pd buffer size, the handling has some issue. and that's why pd async is introduced. pd async not only makes the pipeline async, but also resolve some pd sync backend native limitation.
3. in the scenario you listed above, I would still consider the new "_pending_consumers" counter does similar logic as pin=True. please correct me if i'm wrong. but I don't understand in pd sync backend, how this new counter can resolve the issue? suppose turn pin=True is sufficient. did I miss any important info here?

so the solution in my mind is like this:

pd sync mode: get pin=True back, and live with the known limitations unless we have a simple way to overcome it.
pd async mode: introduce pin=True logic from sync mode to async mode. I did a quick PR for your reference. if possible, you can have a quick check? https://github.com/LMCache/LMCache/pull/3444

I would still encourage you to pick up 3444 can continue the work if 3444 looks promising, as I now don't have sufficient bandwidth, and no real scenario that could well test the issue.

XuanCS:
Hi @hlin99 Thanks for the detailed analysis and the alternative PR — that helps us converge.

## On "leak" vs "limitation"

Agreed, dropping the "leak" framing. We performed load testing with concurrency 16/32/64/128 with request as 50/100. The user-visible failure under sustained load (buffer exhaustion → busy-loop / `_remote_allocate` timeout). With `pin=True`, the buffer is held until end-of-request release.

### On `_pending_consumers` vs `pin=True` for sync

You're right. I traced through the sync code again to confirm: dedup happens at allocate-time via `contains(key, pin=True)` which `ref_count_up`s, and sync `remove()` already does refcount-aware deletion (`if get_ref_count() == 1: del`). The combination is exactly the deferred-delete-while-consumers-pending semantics needed to fix the "Key not found" race. `_pending_consumers` is functionally equivalent bookkeeping that doesn't fix anything additional on the sync path. 

### For this PR
I'll trim #3278 down to what you suggested:

- **Keep:** the `pin=False → pin=True` revert on sync (`_allocate_and_put`). So #3278 becomes the minimal sync revert and #3444 is the async fix. Let me know if that matches your expectation.


### On testing #3444

I tested #3444 end-to-end on a real DPD setup (operator-managed, Llama-3.3-70B, TP=8, P5/H100, async path, NIXL/LIBFABRIC, shared-prefix prompts ~6300 tokens):

- C=1, C=8 (warm cache), C=16: 25/25 requests succeeded, 0 errors, 0 pod restarts
- KV transfer confirmed in decoder logs: 6350/6350 tokens retrieved per worker, External prefix cache hit rate 100%
- No `AssertionError`, no `Key not found`, no `TimeoutError`

hlin99:
> Hi @hlin99 Thanks for the detailed analysis and the alternative PR — that helps us converge.
> 
> ## On "leak" vs "limitation"
> Agreed, dropping the "leak" framing. We performed load testing with concurrency 16/32/64/128 with request as 50/100. The user-visible failure under sustained load (buffer exhaustion → busy-loop / `_remote_allocate` timeout). With `pin=True`, the buffer is held until end-of-request release.
> 
> ### On `_pending_consumers` vs `pin=True` for sync
> You're right. I traced through the sync code again to confirm: dedup happens at allocate-time via `contains(key, pin=True)` which `ref_count_up`s, and sync `remove()` already does refcount-aware deletion (`if get_ref_count() == 1: del`). The combination is exactly the deferred-delete-while-consumers-pending semantics needed to fix the "Key not found" race. `_pending_consumers` is functionally equivalent bookkeeping that doesn't fix anything additional on the sync path.
> 
> ### For this PR
> I'll trim #3278 down to what you suggested:
> 
> * **Keep:** the `pin=False → pin=True` revert on sync (`_allocate_and_put`). So [fix: PD backend deletion guard via pending_consumers  counter #3278](https://github.com/LMCache/LMCache/pull/3278) becomes the minimal sync revert and [Fix PDBackendAsync shared-key races in allocation, transfer, and removal paths #3444](https://github.com/LMCache/LMCache/pull/3444) is the async fix. Let me know if that matches your expectation.
> 
> ### On testing #3444
> I tested #3444 end-to-end on a real DPD setup (operator-managed, Llama-3.3-70B, TP=8, P5/H100, async path, NIXL/LIBFABRIC, shared-prefix prompts ~6300 tokens):
> 
> * C=1, C=8 (warm cache), C=16: 25/25 requests succeeded, 0 errors, 0 pod restarts
> * KV transfer confirmed in decoder logs: 6350/6350 tokens retrieved per worker, External prefix cache hit rate 100%
> * No `AssertionError`, no `Key not found`, no `TimeoutError`

sounds great!

suggest to have two PRs, one for sync and the other for async. thanks again for your great work and effort spending on this!

XuanCS:
Hi @hlin99, sounds good, I updated PR
- Trimmed down to just the `pin=False → pin=True` revert in `_allocate_and_put`
- Dropped the `_pending_consumers` changes

hlin99 (approved):
LGTM

XuanCS:
Hi @DongDongJu @sammshen @ApostaC would you take a look when have a chance, thanks

ApostaC (approved):
Following the discussion above, I think it looks good to me. 
@DongDongJu, gently ping since you have requested changes on this.

DongDongJu (approved):
LGTM. looking forward to see new version of #3444

PR #3289, opened 2026-05-14
[Test] Operator end-to-end smoke suite

royyhuang:

Adds an end-to-end smoke test suite for the operator, with two tiers:

- **no-GPU**: 9 specs covering CR shape, lifecycle (port update, GC, admission rejection), field coverage (ServiceMonitor, extraArgs override, resourceOverrides), and cross-namespace authSecretRef mirroring. Verifies the operator reconciles every K8s artifact it owns; deliberately does not wait for pods to become Ready. Here by no GPU, it does not mean the tests require no GPU since LMCache itself requires CUDA dependency. These tests would still be run on a GPU-equipped cluster.

- **GPU, build tag `e2e_gpu`**: 2 specs that need a live LMCache pod.
  1. `runtime_smoke_test.go`: `GET /api/status` reports the CR's `chunk_size` + `hash_algorithm` so we know argparse propagated CR fields into the running server, not just into the K8s objects.
  2. `vllm_integration_smoke_test.go`: spins up a vLLM Deployment configured against the operator's `<engine>-connection` ConfigMap with `--no-enable-prefix-caching` + `--load-format dummy`, sends the same long prompt twice, asserts the LMCache pod logs `Stored N tokens` after request 1 + `Retrieved N tokens` after request 2.

Four Make targets, naming consistent on `-kind` (self-contained Kind cluster) vs `-cluster` (existing cluster):

- `test-e2e-kind`, `test-e2e-cluster` — no-GPU
- `test-e2e-gpu-kind`, `test-e2e-gpu-cluster` — GPU

The Kind GPU variant uses an inline Kind config + the NVIDIA GPU Operator to make pods scheduled by Kind's inner containerd find NVML / libcuda. The Makefile is split into per-concern files under `operator/make/*.mk` so each file stays under ~175 lines; `make help` still groups targets correctly.

- CI: `.github/workflows/operator_test_e2e.yml` removed. The M1 specs only assert on reconciliation (not pod readiness) and in principle work on CPU, but `lmcache/vllm-openai` is CUDA-dependent and ~14 GB, which exceeds `ubuntu-latest`'s disk budget. E2E will move to Buildkite once that harness is set up. The envtest unit tests in `operator_ci.yml` still run on every PR.
- For run-it-yourself dependency info, see the new "End-to-End Tests" section in `operator/README.md`.
- For internal design rationale (why GPU Operator over the bare device plugin or `nvkind`, the volume-mount marker mechanism, etc.), see `operator/AGENTS.md` "GPU tier (M2)" + the helm-flag comment block at the top of `make/e2e-gpu.mk`.
- Verified on a single-machine Kind + 8×H100 setup: 11/11 specs pass end-to-end.

**If applicable**:

- [x] this PR contains user facing changes - docs added (README.md "End-to-End Tests" section)


sammshen (approved):
LGTM!

ApostaC (approved):
LGTM!

PR #3298, opened 2026-05-15
 [CI] pin vLLM CUDA wheel variant

deng451e:
Description:
  ## Summary
  - Add `VLLM_PRECOMPILED_WHEEL_VARIANT=${CUDA_TAG}` to all `uv pip install vllm`
    calls in `docker/Dockerfile` (3 stages) so vLLM matches the `CUDA_VERSION`
    build-arg instead of falling back to PyPI's cu13 default.
  - Drop the hardcoded cu129 nightly wheel URL in `setup-blend-env.sh`; use the
    variant env + nightly cu129 index instead. No more stale-wheel bumps.

  PyPI `vllm` is now cu13 (`libcudart.so.13`), which won't load in cu12.9
  containers — the env var pin prevents that mismatch.

  ## Test plan
  - [ ] `docker build` for cu13.0 and cu12.9 targets
  - [ ] k3 blend job picks up dynamic cu129 install


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3302, opened 2026-05-17
Unconditionally compile common c++ extensions in setup.py

hlin99:
`setup.py` treated `NO_CUDA_EXT=1` as `BUILDING_SDIST`, which skipped *all* extensions, including pure C++ backends that do not depend on CUDA/ROCm/SYCL. It also duplicated the same pure C++ extension definitions across CUDA/ROCm/SYCL builders.

- **Build-mode semantics cleanup**
  - `BUILDING_SDIST` now represents only real source distribution builds (`"sdist" in sys.argv`).
  - Added explicit `NO_CUDA_EXT` handling for “skip GPU extensions only”.

- **Pure C++ extensions centralized**
  - Introduced `_common_cpp_extensions(extra_cxx_flags)` to define:
    - `lmcache.native_storage_ops`
    - `lmcache.lmcache_redis`
    - `lmcache.lmcache_fs`
    - optional `lmcache.lmcache_mooncake` (env-gated)

- **GPU builders reduced to GPU-only scope**
  - `cuda_extension()`, `rocm_extension()`, `sycl_extension()` now only return:
    - CUDA/ROCm: `lmcache.c_ops`
    - SYCL: `lmcache.xpu_ops`
  - Removed duplicated pure C++ extension definitions from these functions.

- **Top-level extension composition updated**
  - Added `_collect_extensions()` to compose:
    - `sdist`: no extensions
    - `NO_CUDA_EXT=1`: common pure C++ extensions only
    - default backend build: common pure C++ + selected GPU extensions

- **Docs alignment**
  - Updated storage backend build docs to register native backends in `_common_cpp_extensions()`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,16 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
 from pathlib import Path
+from typing import TYPE_CHECKING
 import os
 import sys
 
 # Third Party
 from setuptools import find_packages, setup
+
+if TYPE_CHECKING:
+    # Third Party
+    from setuptools.extension import Extension
 
 ROOT_DIR = Path(__file__).parent
 HIPIFY_DIR = os.path.join(ROOT_DIR, "csrc/")
@@ -13,7 +18,8 @@ HIPIFY_OUT_DIR = os.path.join(ROOT_DIR, "csrc_hip/")
 
 # python -m build --sdist
 # will run python setup.py sdist --dist-dir dist
-BUILDING_SDIST = "sdist" in sys.argv or os.environ.get("NO_CUDA_EXT", "0") == "1"
+BUILDING_SDIST = "sdist" in sys.argv
+NO_CUDA_EXT = os.environ.get("NO_CUDA_EXT", "0") == "1"
 
 # New environment variable to choose between CUDA, HIP, and SYCL
 BUILD_WITH_HIP = os.environ.get("BUILD_WITH_HIP", "0") == "1"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -158,47 +168,75 @@ def cuda_extension() -> tuple[list, dict]:
...
             "lmcache.lmcache_fs",
             sources=fs_sources,
             include_dirs=["csrc/storage_backends", "csrc/storage_backends/fs"],
             extra_compile_args={
-                "cxx": [flag_cxx_abi, "-O3", "-std=c++17"],
+                "cxx": fs_extra_cxx_flags + ["-O3", "-std=c++17"],
             },
         ),
     ]
-    # Mooncake extension is optional.
```
maobaolong, line 194: Why you remove this comment, is it caused by accident?

setup.py
```diff
@@ -158,47 +168,75 @@ def cuda_extension() -> tuple[list, dict]:
...
+        "csrc/pybind.cpp",
+        "csrc/mem_kernels.cu",
+        "csrc/mp_mem_kernels.cu",
+        "csrc/cal_cdf.cu",
+        "csrc/ac_enc.cu",
+        "csrc/ac_dec.cu",
+        "csrc/pos_kernels.cu",
+        "csrc/mem_alloc.cpp",
+        "csrc/utils.cpp",
+        "csrc/event_recorder.cpp",
```
maobaolong, line 225: @ApostaC Do you still remember the dead-lock issue we encountered? The way you fixed that issue is by introduced native `event_recorder`, but how about non-cuda case?

maobaolong (commented):
@hlin99 Thanks for this contribution, left a minor comment and a question to @ApostaC .

maobaolong (approved):
This is essential, LGTM

chunxiaozheng:
@hlin99 Thanks for the improvement, I have a small question, with this PR, when releasing packages in the future, we need to distinguish between different platforms, right?

hlin99:
> @hlin99 Thanks for the improvement, I have a small question, with this PR, when releasing packages in the future, we need to distinguish between different platforms, right?

hi @chunxiaozheng . Good question.  to my best knowledge, this change does not affect anything in place today.
Before the change, several common c++ compilation is binded to cuda/rocm/xpu
After the change, these common c++ compilaltion is triggered regardless of cpu/rocm/xpu/cpu/anything else in future.
only when sdist is set, which means source code packaging, the common c++ compilation will be skipped.

already offline talked with @chunxiaozheng and his question has been addressed

chunxiaozheng (approved):
LGTM!

PR #3305, opened 2026-05-18
Add raw-block payload checksum validation

daegyu94:

This PR adds an optional `payload_checksum` consistency mode for the raw-block backend.

In the default mode, recovery validates only `slot_identity` and `payload_len` from the slot header, which does not detect stale or partially written payload contents after restart. 
In `payload_checksum` mode, raw-block additionally stores a payload checksum in the slot header and verifies recovered entries on the first read after recovery.

The current implementation uses `xxh3_64` as a lightweight integrity check while keeping the write path and on-device metadata format simple. 
The implementation intentionally keeps a single checksum per KV chunk because the goal is lightweight validation of recovered entries rather than a more complex chunk-based integrity or recovery mechanism.

The hash choice itself is not fixed, and the design remains open to adopting different hashes in the future if needed for different performance or reliability tradeoffs.

This PR also makes rollout behavior explicit by storing `consistency_level` in checkpoint metadata and rejecting recovery when the stored consistency level does not match the current configuration.

`payload_checksum` focuses on recovery-time validation of recovered entries, not continuous checksum verification on every read.

Key implementation choices:

- Supported consistency levels are `default` and `payload_checksum`.
- `payload_checksum` stores a payload checksum in the slot header.
- Recovery avoids full payload scans during startup and instead validates recovered entries lazily on the first read.
- On checksum mismatch, the recovered entry is dropped and treated as a cache miss.
- `payload_checksum` requires `meta_verify_on_load=True`, because header validation is responsible for loading checksum metadata into recovered entries.
- Checkpoint recovery is rejected if the stored `consistency_level` differs from the current configuration.
- The current implementation uses `xxh3_64`, but the design remains open to alternative hash implementations in the future.

Performance observations:

| Consistency Model | Backend | Write (MB/sec) | Read (MB/sec) |
|---|---|---:|---:|
| default | posix | 9632.46 | 7792.64 |
| payload_checksum | posix | 9653.50 | 6944.27 |
| default | uring | 9721.88 | 13782.36 |
| payload_checksum | uring | 9663.00 | 13018.09 |

- Write throughput remains effectively unchanged despite the additional payload checksum computation.
- Read throughput shows some regression on the posix path due to recovery-time checksum validation on the first recovered read, while the io_uring backend appears to amortize the additional validation overhead more effectively.
- The current implementation performs lazy validation for all recovered entries. Future optimizations could selectively pre-validate hot or frequently accessed entries during startup to further reduce critical-path recovery overhead.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
@@ -42,6 +42,7 @@ from lmcache.v1.storage_backend.raw_block import (
     RawBlockCoreConfig,
     decode_object_key,
     encode_object_key,
+    normalize_raw_block_consistency_level,
     normalize_raw_block_io_engine,
     validate_raw_block_io_options,
 )
@@ -82,6 +83,7 @@ class RawBlockL2AdapterConfig(L2AdapterConfigBase):
         meta_enable_periodic: bool = True,
         load_checkpoint_on_init: bool = True,
         meta_verify_on_load: bool = True,
+        consistency_level: str = "default",
         enable_zero_copy: bool = True,
         io_engine: str = "posix",
         iouring_queue_depth: int = DEFAULT_IOURING_QUEUE_DEPTH,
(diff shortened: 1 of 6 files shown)


daegyu94:
Hello @DongDongJu ,  I would appreciate your opinion.

Checksum computation adds minimal overhead on the write path. 
While read performance may degrade during recovery due to validation, an open question remains whether checksum validation after restart should be performed eagerly during recovery or deferred lazily until first reuse.

PR #3307, opened 2026-05-18
Improve code quality: simplify boolean return, replace assert, narrow exception

yoo-kumaneko:
## Summary

Small code quality improvements to make the codebase more robust and idiomatic.

### 1. Simplify boolean return in `gpu_connector/utils.py`
```python
# Before
if lmcache_config.enable_pd:
    return False
else:
    return True

# After
return not lmcache_config.enable_pd
```

### 2. Replace `assert` with `RuntimeError` in `protocol.py`
`assert` statements are stripped when Python runs with `-O` optimization flag, which would cause a silent failure. Replaced with an explicit `RuntimeError` to ensure the check always runs.

### 3. Narrow `except Exception` to `except (ValueError, TypeError)` in `s3_connector.py`
The `int(value)` conversion in the S3 connector's header callback was catching all exceptions silently. Since `int()` can only raise `ValueError` or `TypeError`, narrowing the catch avoids swallowing unexpected errors that should propagate.

## Test

All pre-commit hooks pass (codespell, ruff, mypy, etc.).

### DIFF
--- a/lmcache/v1/gpu_connector/utils.py
+++ b/lmcache/v1/gpu_connector/utils.py
@@ -265,10 +265,7 @@ def need_gpu_interm_buffer(lmcache_config: LMCacheEngineConfig):
     Check if the GPU Connector needs to create an intermediate
     buffer on the GPU
     """
-    if lmcache_config.enable_pd:
-        return False
-    else:
-        return True
+    return not lmcache_config.enable_pd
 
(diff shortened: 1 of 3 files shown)


yoo-kumaneko:
This AI-generated PR is too trivial. Let me close it.

PR #3322, opened 2026-05-19
fix: handle full-prompt-hit in MP connector to avoid scheduler assert failure

yoo-kumaneko:
## Problem

In MP mode, when LMCache hits all prompt tokens (full-prompt-hit), the `get_num_new_matched_tokens` method in `lmcache_mp_connector.py` returns `need_to_load` equal to the total number of tokens. This causes the vLLM scheduler to compute:

```
num_computed_tokens = num_new_local_computed + num_external_computed = total_tokens
num_new_tokens = request.num_tokens - num_computed_tokens = 0
```

Which then triggers the assertion failure:
```python
num_new_tokens = min(num_new_tokens, token_budget)
assert num_new_tokens > 0
```

This is because vLLM requires at least one token to be recomputed for logits generation, even when all KV cache is available.

## Why does the last token need to be recomputed when we already have the full KV Cache?

At first glance, it may seem unnecessary to recompute any token when the KV Cache for the entire prompt is already available. However, this is required because of how autoregressive language models work:

1. **KV Cache only stores Keys and Values, not the final hidden state.** During the attention computation in each Transformer layer, the Key and Value projections of all previous tokens are cached so they don't need to be recomputed in future forward passes. However, the **Query projection and the output of the attention layer** for the last token are not stored in the KV Cache — they are computed on the fly.

2. **To generate the next token, we need the hidden state (activation) of the last token.** The generation process works by passing the last token through all Transformer layers to obtain its final hidden state, which is then fed into the language model head (lm_head) to produce logits over the vocabulary. From these logits, the next token is sampled.

3. **The KV Cache saves recomputation of past tokens, but the last token still needs a forward pass.** Even with a full KV Cache hit, we still need to compute the forward pass for the last token — using the cached Keys and Values from all preceding tokens — to obtain its hidden state and subsequently the logits for the next token prediction.

In summary: the KV Cache avoids recomputing attention Keys/Values for previously seen tokens, but it does **not** eliminate the need to compute the forward pass through the model for the last token to get its hidden state for next-token prediction. This is why vLLM always requires `num_new_tokens >= 1`.

## Fix

Align the MP connector with `vllm_v1_adapter.py` by subtracting 1 from `need_to_load` in the full-prompt-hit case (`ret == len(request.all_token_ids)`). This ensures the scheduler always allocates at least one token for recomputation.

The same logic already exists in `vllm_v1_adapter.py`:
```python
if num_external_hit_tokens == request.num_tokens:
    need_to_allocate -= 1
```

## Why not in the adapter layer (`multi_process_adapter.py`)?

This fix belongs in the connector rather than the adapter for the following reasons:

1. **Lack of context**: The adapter (`multi_process_adapter.py`) only communicates with the LMCache server and returns raw lookup results. It does not have access to `request.num_tokens`, which is required to determine whether a full-prompt-hit has occurred.

2. **Separation of concerns**: The adapter is a transport/communication layer responsible for sending lookup/retrieve/store requests. The `-1` adjustment is a vLLM scheduling policy requirement (always recompute the last token for logits), not a concern of the communication layer.

3. **Interface complexity**: Forcing this logic into the adapter would require either passing `num_total_tokens` as an additional parameter to `check_lookup_result` or storing it during `maybe_submit_lookup_request`, both of which would unnecessarily complicate the adapter interface and blur its responsibilities.


lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -788,6 +788,14 @@ def get_num_new_matched_tokens(
         tracker.num_lmcache_hit_blocks = num_lmcache_blocks
 
         need_to_load = max(0, ret - num_computed_tokens)
+
+        # In full-prompt-hit case, we need to recompute the last token.
+        # Without this, num_computed_tokens would equal request.num_tokens,
+        # causing num_new_tokens to be 0 and triggering the
+        # `assert num_new_tokens > 0` in the scheduler.
+        if ret == len(request.all_token_ids):
+            need_to_load = max(0, need_to_load - 1)
```
maobaolong, line 804: It would be helps to maintain this part if you add a comment to describe the motivation of this parts.

yoo-kumaneko: done

ApostaC (approved):
LGTM!

hlin99 (approved):
LGTM

PR #3323, opened 2026-05-19
Refactor GPU KV format dispatch logic

maobaolong:

**If applicable**:


maobaolong:
@sammshen @ApostaC @chunxiaozheng Would you like to help to review this PR? Thanks!

lmcache/v1/gpu_connector/kv_format/kv_format_spec_families.py
chunxiaozheng, line 1: the name of `families.py` is confusing, the `kv_format_spec_families.py` is better than `families.py`, what do you think? @maobaolong

chunxiaozheng:
Otherwise LGTM, the refactor make the code more clear(remove some `if...else...` branch), and make a good abstract, thanks for the work!

ApostaC:
@sammshen Please take a look at this PR.

chunxiaozheng (approved):
LGTM!

sammshen:
@maobaolong quick point of discussion: part of the motivation of constructing the format was to become inference engine agnostic, is it a good idea to couple vllm, slgang, tensorrtllm back into the format?

maobaolong:
Good catch @sammshen — you're right that engine: ClassVar[str] and the engine-named classes leak the abstraction.
The only real runtime use of spec.engine today is in two assert_is_vllm_* helpers in utils.py, and those are really checking layout geometry (per-layer 5-D, non-MLA, non-cross-layer) — they just happen to spell it as "vllm". Detection itself stays per-EngineDetector, which I think is the right place to be engine-aware.
I'd like to keep this PR focused on the strategy/registry split, and follow up with a small separate PR that:
drops engine: ClassVar[str] from all specs,
rewrites the assert_is_vllm_* helpers to assert on layout properties instead,
updates the format-engine table in the doc to mark the engine column as informative.
Happy to fold it into this PR instead if you'd prefer — whichever is easier to review. Thanks!

lmcache/v1/gpu_connector/kv_format/kv_format_spec_families.py
```diff
@@ -0,0 +1,174 @@
...
+from typing import ClassVar
+
+# Third Party
+import torch
+
+# First Party
+from lmcache.v1.gpu_connector.kv_format.base import KVFormatSpec
+
+
+class PerLayer5DSpec(KVFormatSpec):
```
sammshen, line 23: the classification of families doesn't make that much sense to me. i think i see what the idea is with different families having the same way of accessing attributes of the gpu kv buffer. 

however,
1. it is really weird that thte 6D family is a superset of the attributes of the 5D famliy
2. it is really weird that 2 of the families are based on number of dimensions and the last one is called SGLangPBS (by serving engine)

sammshen: KV formats are orthogonal on _at leaset_ 5 axes:

1. engine
2. per layer vs cross layer
3. MLA vs MHA
4. NHD vs HND
5. fused / separate PBS

this can really only be captured by a large enumeration and is extremely difficult to create subdivisions of families.

sammshen:
one design decision that was strong motivation of the previous desing is **signle source of truth**

even though the `utils.py` looks super ugly and is a ton of if statements, it is actually very easy to maintain (by humans or by AI) because the format _implies_ each of the conditions (what the number of layers of, or if it is invalid) and all of this logic is consolidated into a single place.

sammshen:
fundamentally though, I think that the _Strategy pattern_ here will make the code cleanlier if it is implemented in isolation (the structure from `get_attribute` -> switch on the format to putting all of a single formats logic into its own _spec_). it is just that trying to partition by families or serving engine specific dispatch is doing more than one thing in the PR

sammshen:
the `GPUKVFormat` is a purely geometrical description enum (describe physical layout) so it should only ever be owned by higher level descriptors (serving engine, attention backend) and coupled to other geometric descriptors

sammshen:
this is based off the strategy pattern in this PR (which I think is the right direction) but creating purely geometric descriptors: https://github.com/LMCache/LMCache/pull/3485 

@maobaolong @chunxiaozheng @ApostaC PTAL

PR #3325, opened 2026-05-19
fix(v1): graceful skip on slot_mapping/token_ids desync in wait_for_save (fixes #3318)

### LINKED ISSUE #3318: [Bug] vLLM v1 adapter crashes with AssertionError on state desync (slot_mapping and token_ids mismatch)
**Label**

**Describe the bug**
A critical engine crash occurs in the vLLM v1 integration adapter (LMCacheConnectorV1Impl / RequestTracker) due to a strict assert statement checking the state consistency between allocated slot mappings and current token IDs. When a state desync happens (e.g., due to unexpected LMCache allocation failures, engine preemption, or remote cache synchronization lags), the lengths of slot_mapping and token_ids diverge. Currently, the code directly evaluates assert len(slot_mapping) == len(token_ids). When this assertion fails, it raises an AssertionError which abruptly terminates the entire EngineCore process, killing the serving instance for all connected users instead of dropping only the affected request.

**To Reproduce**
1. Initialize the vLLM engine with the v1 experimental backend and LMCache enabled.
2. Trigger a condition where a state desync occurs (e.g., heavy request preemption or a delay/failure in LMCache allocation).
3. The internal loop reaches LMCache/lmcache/integration/vllm/vllm_v1_adapter.py around line 1105.
4. See error: The entire EngineCore is terminated with an AssertionError: Slot mapping and token IDs length mismatch!.

**Expected behavior**
Instead of terminating the entire EngineCore via an unhandled AssertionError, the adapter should handle this mismatch gracefully. It should log an error/warning message and either raise a proper RuntimeError/ValueError that the upstream scheduler can catch to safely abort only that specific request, or executing a fallback mechanism to keep the core engine alive for other active users.

**Screenshots**
  File "lmcache/integration/vllm/vllm_v1_adapter.py", line 1124, in ...
    assert len(slot_mapping) == len(token_ids)
AssertionError: Slot mapping and token IDs length mismatch!
[INFO] vLLM EngineCore terminated unexpectedly (Killed).

 - OS:  Ubuntu 22.04 (NVIDIA DGX H100)

**Additional context**
Inside the wait_for_save(self) function, we should replace the strict assert statement with an explicit conditional check (if len(slot_mapping) != len(token_ids):). This allows the system to log the error diagnostics and handle the desync gracefully (e.g., skipping the save or raising a controlled exception) without crashing the entire engine.

# Before:
Inside wait_for_save(self)
assert len(slot_mapping) == len(token_ids), "Slot mapping and token IDs length mismatch!"

# After (Proposed):
Inside wait_for_save(self)
if len(slot_mapping) != len(token_ids):
    logger.warning(
        "Skipping KV save for request %s due to slot_mapping/token_ids "
        "length mismatch (slot_mapping=%d, token_ids=%d). "
        "Likely caused by upstream allocation failure under load.",
        request.req_id, len(slot_mapping), len(token_ids),
    )
    continue

abinggo:
## Summary

Fixes #3318. The vLLM v1 adapter's `wait_for_save` previously asserted `len(slot_mapping) == len(token_ids)` on the non-layerwise save path. When the engine hit an upstream state desync (allocation failure, preemption-induced mismatch, remote cache lag), the `AssertionError` was unhandled and killed the entire `EngineCore`, taking down every connected user.

## Repro (from #3318)

> 1. Initialize the vLLM engine with the v1 experimental backend and LMCache enabled.
> 2. Trigger a state desync (heavy preemption, LMCache allocation delay/failure).
> 3. Internal loop hits `vllm_v1_adapter.py` line ~1124.
> 4. Whole `EngineCore` is killed:
>
> ```
> AssertionError: Slot mapping and token IDs length mismatch!
> [INFO] vLLM EngineCore terminated unexpectedly (Killed).
> ```

## Fix

Replace the strict assert with a conditional `continue` that logs a warning naming the request id and both observed lengths. The engine stays alive and only the desynced request's save is dropped:

```python
if len(slot_mapping) != len(token_ids):
    logger.warning(
        "Skipping KV save for request %s: slot_mapping/token_ids "
        "length mismatch (slot_mapping=%d, token_ids=%d). Likely "
        "an upstream allocation/preemption desync; the engine "
        "stays alive and only this request's save is dropped.",
        request.req_id,
        len(slot_mapping),
        len(token_ids),
    )
    continue
```

This matches `docs/coding_standards.md` (\"No `assert` for validation\") — applied here as drop-the-save rather than raise, since the save path should be fail-soft toward the connected stream consumers (the alternative — raising — propagates back to `EngineCore` and reproduces the original kill).

## Test

`tests/v1/test_v1_adapter_state_desync.py` is a CPU-only regression test that:

- Feeds a length-mismatched request (`token_ids` len 4, `slot_mapping` len 3) into `wait_for_save` via fake parent / engine fixtures (same shape as the existing `tests/v1/test_vllm_layerwise_wait_for_save.py`).
- Asserts:
  1. `wait_for_save()` returns normally (no `AssertionError`).
  2. A warning is emitted naming the request id (`req-desync`) and both lengths.
  3. `lookup_unpin` still ran (pin balance preserved).
  4. `lmcache_engine.store` was **not** called for the desynced request (the save is dropped, not silently corrupted).

## Out of scope (intentional)

The same pattern (`assert len(slot_mapping) == len(token_ids)`) also appears at line 1023 inside `save_kv_layer` (the layerwise path). That one is **not** changed in this PR to keep the scope tight to #3318's reported failure mode (`wait_for_save`). A follow-up for the layerwise path can mirror this fix once we confirm the same desync class can hit it on the layerwise side.

## Local verification

The CPU regression test was authored against the existing layerwise test fixtures. It cannot run on this host (Mac without `vllm` and the `lmcache.c_ops` C extension), so it relies on the LMCache CI environment for validation. The fix itself was verified by:

- Reading the diff against the surrounding `wait_for_save` flow (`for request in connector_metadata.requests:` loop, with `lookup_unpin` already called and `save_spec.can_save` already gated above the changed line).
- Confirming the test mirrors `test_vllm_layerwise_wait_for_save.py`'s parent / engine / metadata fake pattern.

Closes #3318.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1121,7 +1121,17 @@ class LMCacheConnectorV1Impl:
 
             slot_mapping = request.slot_mapping
             assert isinstance(slot_mapping, torch.Tensor)
-            assert len(slot_mapping) == len(token_ids)
+            if len(slot_mapping) != len(token_ids):
+                logger.warning(
+                    "Skipping KV save for request %s: slot_mapping/token_ids "
+                    "length mismatch (slot_mapping=%d, token_ids=%d). Likely "
+                    "an upstream allocation/preemption desync; the engine "
+                    "stays alive and only this request's save is dropped.",
+                    request.req_id,
+                    len(slot_mapping),
+                    len(token_ids),
+                )
+                continue
 
             # TODO: have a pre-allocated buffer to hold the slot_mappings
             slot_mapping = slot_mapping.to(self.device)
--- a/tests/v1/test_v1_adapter_state_desync.py
+++ b/tests/v1/test_v1_adapter_state_desync.py
@@ -79,11 +79,15 @@ def _make_desync_request(
 def _make_connector(
     requests: list[SimpleNamespace],
 ) -> tuple[LMCacheConnectorV1Impl, _FakeEngine]:
-    metadata = LMCacheConnectorMetadata(requests=requests)
+    metadata = LMCacheConnectorMetadata(requests=requests)  # type: ignore[arg-type]
     engine = _FakeEngine()
     connector = LMCacheConnectorV1Impl.__new__(LMCacheConnectorV1Impl)
     connector._parent = _FakeParent(metadata)
-    connector.lmcache_engine = engine
+    # ``lmcache_engine`` is a read-only property backed by ``self._manager``;
+    # inject the fake engine through the manager so the property resolves to it.
+    connector._manager = SimpleNamespace(  # type: ignore[assignment]
+        lmcache_engine=engine
+    )
     connector.kv_role = "kv_producer"
     connector.use_layerwise = False
     connector.enable_blending = False
@@ -102,9 +106,7 @@ def test_wait_for_save_skips_desynced_request_and_keeps_engine_alive(
 
     Regression for https://github.com/LMCache/LMCache/issues/3318.
     """
-    desync_req = _make_desync_request(
-        "req-desync", token_ids_len=4, slot_mapping_len=3
-    )
+    desync_req = _make_desync_request("req-desync", token_ids_len=4, slot_mapping_len=3)
     connector, engine = _make_connector([desync_req])
 
     with caplog.at_level(logging.WARNING):
@@ -123,4 +125,7 @@ def test_wait_for_save_skips_desynced_request_and_keeps_engine_alive(
         and "slot_mapping=3" in r.getMessage()
         and "token_ids=4" in r.getMessage()
         for r in warnings
-    ), f"Expected desync warning naming req-desync; got {[r.getMessage() for r in warnings]}"
+    ), (
+        "Expected desync warning naming req-desync; "
+        f"got {[r.getMessage() for r in warnings]}"
+    )
--- a/tests/v1/test_v1_adapter_state_desync.py
+++ b/tests/v1/test_v1_adapter_state_desync.py
@@ -98,45 +98,58 @@ def _make_connector(
     return connector, engine
 
 
-def test_wait_for_save_skips_desynced_request_and_keeps_engine_alive(
-    caplog: pytest.LogCaptureFixture,
-    monkeypatch: pytest.MonkeyPatch,
-) -> None:
+def test_wait_for_save_skips_desynced_request_and_keeps_engine_alive() -> None:
     """Length mismatch must drop only the affected request's save, log a
     warning, and let ``wait_for_save`` return normally.
 
     Regression for https://github.com/LMCache/LMCache/issues/3318.
     """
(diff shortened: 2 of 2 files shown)


ApostaC (approved):
LGTM!

ApostaC:
@maobaolong @DongDongJu do you want to take a second quick look at this? It's a simple one.

DongDongJu (approved):
LGTM

ApostaC:
Hey @abinggo , could you fix the linter issue? Thanks!

abinggo:
Done — pushed 744c05b, "Check code quality" is green now.

The lint failures were all in the new regression test. The dev refactor turned `lmcache_engine` into a read-only property backed by `self._manager`, so the test's old direct `connector.lmcache_engine = engine` assignment broke both mypy (read-only property) and the runtime. Fixed by injecting the fake engine through `connector._manager` so the property resolves to it, plus a `type: ignore` for the `SimpleNamespace` mocks and wrapping the long assert message for E501. Verified locally with `ruff` + `mypy` against the repo config. Thanks for the review!

DongDongJu:
@abinggo IDK why CI failed but could you check again whether it caused by this or not?

abinggo:
@DongDongJu Investigated -- CI failure is caused by my own test, not unrelated. Pushed f579ebf to fix.

Root cause: my round-2 fix used `caplog` + `monkeypatch.setattr(propagate=True)` to bypass lmcache's `init_logger(propagate=False)`. CI logs show the warning *did* fire (visible twice in stderr via lmcache's own StreamHandler) but `caplog.records` was empty -- pytest's root-attached caplog handler never saw the record. The propagate flip is unreliable: anything that re-invokes `init_logger` for the same name resets handlers + `propagate=False`.

Switched to a local `_ListHandler` attached directly to the adapter logger. Records are captured regardless of how lmcache configures propagation, with cleanup in `finally`. Codex review caught a follow-up: `init_logger` reads logger level from `LMCACHE_LOG_LEVEL` (default INFO) -- if a prior import or CI env sets it above WARNING, `logger.warning` would be filtered before reaching the handler. Force-set logger level to WARNING for the test scope and restore in `finally`.

Per-file pre-commit (isort / ruff / ruff-format / codespell / mypy / SPDX) green. Waiting for CI to confirm.

abinggo:
@ApostaC this one has your approval + @DongDongJu's and CI is green — it's just waiting on a `/lmcache/integration/vllm/` codeowner to satisfy the merge gate. The sibling fix #3340 (Scheduler-role abort, same file) is in the same state: your approval, green CI, codeowner-gated. Both are small, self-contained exception-safety fixes for the v1 adapter (and #3499 was just filed as a fresh duplicate of #3340 — there's demand). Could you help route the codeowner step or land them? Happy to rebase if anything's gone stale.

PR #3327, opened 2026-05-19
fix(gds): use parse_cache_key to handle LayerCacheEngineKey on restart

JinuJeong:
## Problem
GDS backend stores KV cache files with keys formatted as `LayerCacheEngineKey.to_string()`, which appends a layer_id suffix (e.g., `@bfloat16@4`). On server restart, `_scan_metadata_subdir` scans these files and attempts to reconstruct the cache key from the filename.
However, the code was calling `CacheEngineKey.from_string()` unconditionally, which expects the 6th part of the string to be in `tag%value` format. When encountering a numeric layer_id like `@4`, it fails with:
Filename ... can't be converted back into cache key: Invalid key string: ...
This prevents `LayerCacheEngineKey` entries from being loaded into `hot_cache` on restart, causing cache misses for layer-wise stored KV caches.
## Fix
Use the existing `parse_cache_key()` utility instead of `CacheEngineKey.from_string()` directly. This utility correctly identifies when the 6th part is a numeric layer_id and dispatches to `LayerCacheEngineKey.from_string()`.
## Code Changes
- `lmcache/v1/storage_backend/gds_backend.py`: import `parse_cache_key` and use it in `_scan_metadata_subdir`
- `tests/v1/storage_backend/test_gds_backend.py`: add `create_test_layer_key()` helper and `test_scan_metadata_layer_key` test
## Test
```bash
pytest -xvs tests/v1/storage_backend/test_gds_backend.py::TestGdsBackend::test_scan_metadata_layer_key
- Writes a LayerCacheEngineKey via GDS
- Closes and recreates the backend to trigger _scan_metadata
- Verifies hot_cache correctly reconstructs the LayerCacheEngineKey

### DIFF
--- a/lmcache/v1/storage_backend/gds_backend.py
+++ b/lmcache/v1/storage_backend/gds_backend.py
@@ -21,7 +21,12 @@ import torch
 
 # First Party
 from lmcache.logging import init_logger
-from lmcache.utils import CacheEngineKey, DiskCacheMetadata, _lmcache_nvtx_annotate
+from lmcache.utils import (
+    CacheEngineKey,
+    DiskCacheMetadata,
+    _lmcache_nvtx_annotate,
+    parse_cache_key,
+)
 from lmcache.v1.config import LMCacheEngineConfig
 from lmcache.v1.memory_management import (
     CuFileMemoryAllocator,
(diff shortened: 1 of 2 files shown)


KuntaiDu (commented):
LGTM! @YaoJiayi @sammshen @deng451e @maobaolong  maybe also take a look?

DongDongJu (approved):
LGTM

ApostaC (approved):
LGTM!

PR #3330, opened 2026-05-19
[Core][MP] Add native C++ MP server

DongDongJu:

Adds an experimental native C++ implementation of the LMCache multiprocess server.
I know we need a discussion for these stuffs but open it for discussion as well.

The native path keeps the existing token-key MP protocol, supports the CLI server launcher, and defaults to no-CUDA unless CUDA mode is explicitly requested. This gives us a native MP server path for lower protocol/control overhead while preserving the Python MP path.

  Validation:
  - `uv run pytest tests/v1/multiprocess/test_custom_types.py tests/v1/multiprocess/test_free_locks.py tests/cli/commands/test_server.py tests/v1/test_vllm_mp_adapter.py`
    - `50 passed`
  - Native no-CUDA CMake build completed for `lmcache-mp-server-native`.
  - CLI-launched native token-key e2e passed.
  - Python MP vs native no-CUDA MP comparison matched correctness for common control and negative lookup operations.

  Local protocol performance comparison:
  - Native was faster on `CLEAR`, `END_SESSION`, `GET_CHUNK_SIZE`, and `LOOKUP + QUERY_PREFETCH_STATUS`.
  - `PING` was roughly comparable.
  - `NOOP` was slower in native in this local run.

  **If applicable**:


DongDongJu:
Before talking about the review, I want to know the opinion about this approach first.
@ApostaC @sammshen @maobaolong @hlin99 Please leave any opinion.
It will helpful for revising it. I forgot to add the AI DCO. will add it next rebase.

DongDongJu:
Like discussed in offline, I will make this one as standalone repo for AI to maintaining it.

PR #3334, opened 2026-05-19
Fix the translation icon together with the screen layout adjustment

glbyktjys:
Fix the language switcher (中文 | Eng) overlapping with the GitHub / profile / theme-toggle icons when the browser window is shrunk.
The switcher was appended to `document.body` and positioned with hard-coded viewport coordinates (`top: 10px; right: 72px`). As the window narrowed, the nav-bar icons shift leftward but the switcher stayed pinned at its fixed coordinates, causing overlap.
Fixed by injecting the switcher into the top nav bar (the same flex container that holds the GitHub / profile / theme-toggle icons), so it flows naturally with them at every window size.

**If applicable**:


deng451e (approved):
LGTM

sammshen (approved):
LGTM!

PR #169, opened 2024-10-21
[CI] Format checker fix

KuntaiDu:
Fix format checker.

Now it will prompt
image
upon success, and prompt
image
when format checker passed but there are still files that need to be commited, and prompt
image
when some packages are missing.

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

<details>

<summary><b> PR Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to LMCache! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Model]</code> for adding a new model or improving an existing model. Model name should appear in the title.</li>
    <li><code>[Core]</code> for changes in the core LMCache logic (e.g., <code>LMCacheEngine</code>, <code>Backend</code> etc.)</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <a href="https://github.com/LMCache/LMCache/blob/dev/format.sh"><code>format.sh</code></a> to format your code.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>What to Expect for the Reviews</h3>

We aim to address all PRs in a timely manner. If no one reviews your PR within 5 days, please @-mention one of KuntaiDu, ApostaC or YaoJiayi.

Issue #189, opened 2024-10-30
[Core] Upcoming features for KV blending module


ApostaC:
- [x] #190
- [x] #191
- [x] #192
- [x] #193
- [ ] [KV Blending] Compatible with LMCache's prefix caching functionality
- [x] #194
- [x] [KV Blending] Performance testing
- [ ] [KV Blending] (Optional) compatible with chunked prefill and vLLM's prefix caching

PR #3336, opened 2026-05-19
[MP][Bugfix] Fix store/retrieve deadlock by routing completion callbacks through C++ host callback

elliotz-ai:
## Problem Description

- PR #2952 fixed a CUDA-driver × GIL deadlock in `EventBus.publish_on_stream` by moving the on-stream host callback to native C++.
- Two operational callsites in `Server.store` and `Server.retrieve` still schedule Python bound-method callbacks via `cupy_stream.launch_host_func(...)`:
  ```python
  gpu_context.cupy_stream.launch_host_func(
      self.storage_manager.finish_write, list(reserved_dict.keys()))
  gpu_context.cupy_stream.launch_host_func(
      self.storage_manager.finish_read_prefetched, prefetched_keys)
  ```
- These hit the same AB-BA deadlock: the caller holds the GIL; when the temporary IPC `torch.cuda.Event` hits refcount-zero, `hipEventDestroy` (AMD) calls `SyncAllStreams` which waits for the host callback to drain — but the host callback is parked acquiring the GIL the destructing thread still holds.
- Reproduced reliably on MI350X (ROCm 7.0.2 / torch 2.10.0+rocm7.0): the engine hangs within ~30 s of the first store at 0.3 QPS prompt-replay, and `rocgdb` shows one R-state thread spinning in:
  ```
  sched_yield → amd::Event::awaitCompletion → HostQueue::finish → Device::SyncAllStreams →
    ihipHostUnregister → IPCEvent::~IPCEvent → hipEventDestroy → THCPEvent_dealloc → Py_DECREF
  ```

## Solution

- Extend PR #2952's `EventRecorder` pattern to a new `CompletionRecorder`:
  - C++ buffer of `std::unique_ptr<PendingCompletion>` pushed from a `cudaLaunchHostFunc` / `hipLaunchHostFunc` callback that never touches the GIL (`py::call_guard<py::gil_scoped_release>` on the pybind entry).
  - Python-side `DeviceHostFuncDispatcher` runs a background drain thread that pulls buffered records and dispatches each to the handler registered for its `kind`.
- `MPCacheEngine.__init__` constructs the dispatcher and registers `storage_manager.finish_write` and `storage_manager.finish_read_prefetched` with `item_type=ObjectKey`; `close()` stops it before tearing down the storage manager.
- The two `launch_host_func` calls in `Server.store` and `Server.retrieve` are replaced with `submit_callback_to_stream(...)`. The native recorder is always required (no fallback): `lmcache.c_ops` is built whenever `MPCacheEngine` is used.
- Payload (`list[ObjectKey]`) is `msgspec.msgpack`-encoded per-item before the C++ call so the buffer can hold opaque bytes without holding the GIL; the dispatcher decodes back via a per-kind `msgspec.msgpack.Decoder(type=item_type)` before invoking the handler.
- The C++ entry point checks the launch return code and reclaims the heap allocation on failure. Buffer ownership is `std::unique_ptr` end-to-end: `record_completion_on_stream` makes a `unique_ptr`, releases it as a raw pointer through the driver, the host callback re-adopts into a `unique_ptr`; on launch failure the raw pointer is `delete`d.
- `drain_recorded_completions` returns each payload item as `py::bytes` rather than `std::string`. Pybind11 utf-8-decodes `std::string` by default, which corrupted binary `msgpack`/`pickle` payloads on the way back to Python and produced `UnicodeDecodeError` in the dispatcher (this was the original "stored objects can't be retrieved" end-to-end CI failure).

## Architecture Diagram

```mermaid
graph TD
    StoreCall[Server store finally block] --> SubmitCb[submit callback to stream msgpack encode payload]
    RetrieveCall[Server retrieve finally block] --> SubmitCb
    SubmitCb --> CppRecorder[CompletionRecorder C plus plus buffer]
    CudaCallback[cudaLaunchHostFunc or hipLaunchHostFunc callback no GIL] --> CppRecorder
    CppRecorder --> DrainThread[DeviceHostFuncDispatcher drain thread]
    DrainThread --> Handlers[storage manager finish write or finish read prefetched]
```

## Code Flow Diagram

```mermaid
sequenceDiagram
    participant Caller as Caller thread Server store
    participant Stream as cupy stream
    participant Driver as CUDA HIP driver thread
    participant Recorder as CompletionRecorder
    participant Drain as DeviceHostFuncDispatcher thread
    participant Handler as storage manager finish write
    Caller->>Stream: enqueue D2H copies
    Caller->>Recorder: submit callback to stream kind finish write payload keys
    Recorder->>Stream: launch host callback no GIL
    Caller->>Caller: event record then function returns
    Stream-->>Driver: previous work completes
    Driver->>Recorder: host callback pushes pending completion
    loop drain interval 5ms
        Drain->>Recorder: drain recorded completions returns py bytes per item
        Drain->>Handler: msgpack decode then dispatch payload
    end
```

## Change Summary

| File | Change Type | Description |
|------|-------------|-------------|
| `csrc/completion_recorder.h` | Added | Public C++ API: `PendingCompletion`, `CompletionRecorder` singleton (buffer of `std::unique_ptr`), `record_completion_on_stream`, `drain_recorded_completions`. |
| `csrc/completion_recorder.cpp` | Added | Implementation: thread-safe buffer, `hipLaunchHostFunc`/`cudaLaunchHostFunc` callback that runs without the GIL, `unique_ptr` ownership end-to-end, error-check + free-on-failure on the launch call. |
| `csrc/pybind.cpp` | Modified | Bind `record_completion_on_stream` (GIL-released via `py::call_guard`) and `drain_recorded_completions` (lambda wrapping that returns `py::bytes` for each payload item to keep binary payloads intact). |
| `setup.py` | Modified | Add `csrc/completion_recorder.cpp` to both `cuda_sources` and `hip_sources`. |
| `lmcache/v1/multiprocess/native_completion.py` | Added | `DeviceHostFuncDispatcher` (drain thread + per-kind handler/`msgspec.msgpack.Decoder` registry) and `submit_callback_to_stream` helper. |
| `lmcache/v1/multiprocess/server.py` | Modified | Construct/start dispatcher in `MPCacheEngine.__init__`, register `finish_write` and `finish_read_prefetched` handlers with `item_type=ObjectKey`, swap the two `launch_host_func` calls to `submit_callback_to_stream`, stop dispatcher in `close()`. |
| `tests/v1/multiprocess/test_completion_recorder.py` | Added | Unit + integration tests covering the recorder, dispatcher, and a concurrent-records deadlock regression. |

## Test Case Coverage

| Category | Test Cases |
|----------|-----------|
| Unit (native, GPU-only) | `drain_recorded_completions` empty; single completion roundtrip with `py::bytes` payload assertion (catches the original utf-8 decode regression); ordered drain of 50 completions. |
| Integration (native, GPU-only) | Handler dispatched after stream sync (decoded via per-kind `msgspec.msgpack.Decoder`), unknown kind dropped with warning, handler exception does not kill drain thread. |
| Regression (native, GPU-only) | 200 concurrent `submit_callback_to_stream` calls complete within a 10 s timeout — the legacy `launch_host_func(python_callback)` path deadlocks here. |
| End-to-end (manual, AMD MI350X) | DeepSeek-V3 dummy weights, TP=8, ROCm 7.0.2 + torch 2.10.0+rocm7.0, `block-size=1`, AITER MLA, `LMCacheMPConnector` over loopback. Before: deadlock within ~30 s of first store at 0.3 QPS prompt-replay (5.4k-token prompts). After: 552 stores, 75% L1 prefix hit rate, 0 `LMCache server is unhealthy` warnings over a full 20-min replay run; TTFT p99 dropped from 17 s to 366 ms. Re-validated at higher load (4 QPS, 30 min) — clean. |

## Reproduction Results

Mistral-Small-24B-Instruct-2501 with `--load-format=dummy`, `LMCacheMPConnector`, `--enable-prefix-caching`, identical lmcache args (`--l1-size-gb 5.0`, `--chunk-size 256`). 5k-token prefix per request.

| Platform | TP | QPS | Duration | Result |
|---|---|---|---|---|
| AMD MI350X | 2 | 2 | 15 min | clean |
| AMD MI350X | 4 | 2 | hung at ~110s | deadlock fires (unpatched) |
| AMD MI350X | 8 | 2 | hung at ~70s | deadlock fires (unpatched) |
| NVIDIA H100 80GB | 4 | 2 | 5 min | clean (600 reqs, 0 errors) |
| NVIDIA H100 80GB | 8 | 2 | 5 min | clean (600 reqs, 0 errors) |
| NVIDIA H100 80GB | 8 | 8 | 15 min | clean (7200 reqs, 0 errors, ~73k lmcache store/retrieves) |

The deadlock reproduces on AMD/HIP only. Destructor wall-time on AMD scales with the mapped IPC handle count because `Device::SyncAllStreams` iterates them — higher TP yields more concurrent IPC events, longer destructor window, and a wider GIL × driver-lock collision interval (TP=4 hangs after ~110s, TP=8 after ~70s). The CUDA driver's analogue does not have this property; the same workload (and 4× heavier) ran clean on H100. With this PR's fix the GIL is no longer required on the driver thread, so the AB-BA cycle is broken regardless of how long the destructor runs.

## CI Status

- `Check code quality`, all 4 `test:` ubuntu matrix jobs (3.10–3.13), and `buildkite/k3-multiprocess-test` pass on the latest commit.
- `buildkite/k3-sglang-test` fails in ~5 s. The same job also fails on `upstream/dev` HEAD `d9f3438` (Build #169 — 4 s failure) and there is no in-repo pipeline configuration for `sglang` (no matches in `.buildkite/` for it). This is a pre-existing pipeline-side issue, not introduced by this PR.
- All 8 inline review comments from prior reviewers were addressed in commit `6f941130` (renames `CompletionDispatcher`→`DeviceHostFuncDispatcher` / `CompletionHandler`→`DeviceHostFunc` / `record_on_stream`→`submit_callback_to_stream`; swap pickle→`msgspec.msgpack`; drop the legacy `launch_host_func` fallback; switch buffer to `std::unique_ptr`; expand the module-level call-graph docstring).


csrc/completion_recorder.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+CompletionRecorder& CompletionRecorder::instance() {
+  static CompletionRecorder recorder;
+  return recorder;
+}
+
+void CompletionRecorder::push(PendingCompletion* completion) {
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    buffer_.push_back(std::move(*completion));
```
ApostaC: I'm concerned about whether std::move(completion) will free the original memory on the heap. It will move the underlying strings to the new object, but the reference to the original `CompletionRecorder` becomes dangling there. We need to free the original pointer to avoid a memory leak.

An alternative is to directly keep the pointers, and call `delete` during `drain()`.
Or we can use `std::unique_ptr` to better handle the "move" semantics.

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,158 @@
...
+through a C++ host callback so the CUDA driver thread never touches
+the GIL.
+
+The legacy path used ``stream.launch_host_func(finish_write, keys)``,
+which scheduled a Python callback on the CUDA stream. The driver thread
+running the callback would block on the GIL while the calling thread
+held both the GIL and the CUDA driver lock — the same deadlock that
+PR #2952 fixed for ``EventBus.publish_on_stream``. This module extends
+that pattern to the two remaining ``launch_host_func`` callsites in
+``Server``.
```
ApostaC: It's better to have a diagram showing the "call graph" (as in the PR description) than the current description.

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,158 @@
...
+    _has_native = False
+
+logger = init_logger(__name__)
+
+# Handler receives the same Python objects the caller passed to
+# ``record_on_stream``. Runs on the dispatcher thread under the GIL.
+CompletionHandler = Callable[[Any], None]
+
+
+class CompletionDispatcher:
```
ApostaC: Alternative proposal for the class name: How about we use `DeviceHostFunctDispatcher` to have better clarity? Therefore, it's not limited to the "completions".

And the `CompletionHandler` can be renamed to `DeviceHostFunc` in this case.

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,158 @@
...
+
+def is_native_available() -> bool:
+    return _has_native
+
+
+def record_on_stream(
+    stream: Any,
+    kind: str,
+    payload: Iterable[Any],
+    fallback_handler: CompletionHandler | None = None,
```
ApostaC: I don't think we need this fallback mechanism. The `lmcache.c_ops` should almost always be compiled when the MPCacheEngine is used (otherwise the GPU kernels won't work)

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,158 @@
...
+    """Schedule a completion record on *stream* without taking the GIL.
+
+    Falls back to ``stream.launch_host_func(fallback_handler, payload)``
+    when the native recorder isn't compiled into ``lmcache.c_ops``.
+    """
+    payload_list = list(payload)
+    if _has_native:
+        encoded = [
+            pickle.dumps(p, protocol=pickle.HIGHEST_PROTOCOL) for p in payload_list
+        ]
```
ApostaC: I'm a bit not sure about the performance of `pickle`. How about using `msgspec.msgpack` here?

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,158 @@
...
+                        self._exception_counts.get(kind, 0) + 1
+                    )
+                logger.exception("CompletionDispatcher: handler for %r raised", kind)
+
+
+def is_native_available() -> bool:
+    return _has_native
+
+
+def record_on_stream(
```
ApostaC: nit: maybe rename it to `submit_callback_to_stream`?

ApostaC (commented):
Thanks for the fix. Please see my comments below!

csrc/completion_recorder.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+void record_completion_on_stream(int64_t cuda_stream_ptr,
+                                 const std::string& kind,
+                                 std::vector<std::string> payload) {
+  auto* completion = new PendingCompletion{kind, std::move(payload)};
+  auto stream = reinterpret_cast<lmcache_completion_stream_t>(
+      static_cast<uintptr_t>(cuda_stream_ptr));
+  LMCACHE_COMPLETION_LAUNCH_HOST_FUNC(stream, completion_host_callback,
+                                      completion);
```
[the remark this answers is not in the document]
elliotz-ai: Addressed in 5b91c12c (now refactored further in 6f941130). The launch return code is checked and `raw` is `delete`d on failure.

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,164 @@
...
+
+    if fallback_handler is None:
+        raise RuntimeError(
+            "record_on_stream: native recorder unavailable and no fallback_handler"
+        )
+    stream.launch_host_func(fallback_handler, payload_list)
+
+
+def now_seconds() -> float:
+    return time.time()
```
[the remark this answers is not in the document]
elliotz-ai: Addressed in 5b91c12c. `now_seconds` removed.

csrc/completion_recorder.cpp
elliotz-ai: Switched the buffer to `std::vector<std::unique_ptr<PendingCompletion>>` in 6f941130. Ownership is now explicit end-to-end: `record_completion_on_stream` makes a `unique_ptr`, releases it as a raw pointer through the driver, and the host callback re-adopts into a `unique_ptr`; on launch failure the raw pointer is `delete`d. Drain transfers `unique_ptr`s out by swap.

lmcache/v1/multiprocess/native_completion.py
elliotz-ai: Added an ASCII call-graph diagram to the module docstring in 6f941130.

lmcache/v1/multiprocess/native_completion.py
elliotz-ai: Renamed in 6f941130: `CompletionDispatcher` -> `DeviceHostFuncDispatcher`, `CompletionHandler` -> `DeviceHostFunc`.

lmcache/v1/multiprocess/native_completion.py
elliotz-ai: Dropped the legacy `launch_host_func` fallback path entirely in 6f941130. `submit_callback_to_stream` now unconditionally goes through the native recorder.

lmcache/v1/multiprocess/native_completion.py
elliotz-ai: Switched to `msgspec.msgpack` in 6f941130. The dispatcher's `register(kind, handler, item_type)` now takes a per-kind decode type, used to construct a `msgspec.msgpack.Decoder` once per registration.

lmcache/v1/multiprocess/native_completion.py
elliotz-ai: Renamed to `submit_callback_to_stream` in 6f941130.

lmcache/v1/multiprocess/server.py
```diff
@@ -249,6 +253,21 @@ def __init__(
         # EventBus for observability
         self._event_bus = get_event_bus()
 
+        # Route finish_write / finish_read_prefetched through a C++ host
+        # callback so the driver thread doesn't acquire the GIL.
+        self._device_host_func_dispatcher = DeviceHostFuncDispatcher()
+        self._device_host_func_dispatcher.register(
+            "finish_write",
+            self.storage_manager.finish_write,
+            item_type=ObjectKey,
```
ApostaC: It's a bit confusing here because `finish_write` and `finish_read_prefetched` takes `list[ObjectKey]` as input, but we pass just `ObjectKey` here.

lmcache/v1/multiprocess/native_completion.py
```diff
@@ -0,0 +1,147 @@
...
+                    "DeviceHostFuncDispatcher: handler for %r raised", kind
+                )
+
+
+def submit_callback_to_stream(stream: Any, kind: str, payload: Iterable[Any]) -> None:
+    """Schedule a stream-ordered host callback for *kind* with *payload*.
+    Runs on the dispatcher worker registered for *kind*; never acquires the
+    GIL on the driver thread."""
+    encoded = [msgspec.msgpack.encode(p) for p in payload]
+    _lmc_ops.record_completion_on_stream(stream.ptr, kind, encoded)
```
ApostaC: The problem with the current implementation is that it assumes the callback function has the signature of `callback_function(arguments: list[ItemType])`. 
This is correct under the current implementation, but may create confusion for future use.

**My proposal:**
Let's assume the callback function just takes one single argument (with potentially complicated types). So the `payload`'s type will be `Any` rather than `Iterable[Any]` here.
- In the current case, the real payload type is `list[ObjectKey]`
- If we have a callback function that has 2 or more arguments in the future, we need to wrap it into a lambda/helper function that takes in a single tuple of all the arguments.

WDYT?

ApostaC (commented):
Thanks! A few comments on the callback function's clarity.

lmcache/v1/multiprocess/server.py
elliotz-ai: Done in 4c0da89e — the dispatcher now decodes the whole payload once and hands it to the handler unchanged. Registered as `payload_type=list[ObjectKey]` so the handler signature lines up.

lmcache/v1/multiprocess/native_completion.py
elliotz-ai: Agreed — adopted in 4c0da89e. `submit_callback_to_stream` now takes a single `payload: Any`, encoded once with `msgspec.msgpack.encode(payload)`. The dispatcher decodes once and calls `handler(decoded)`. For the current use case the registered `payload_type` is `list[ObjectKey]`; multi-arg handlers can wrap their args in a tuple (test added to cover that pattern).

ApostaC (approved):
LGTM!

elliotz-ai:
The two failing Buildkite checks (`k3-correctness-test` #3395 and `k3-sglang-test` #186) are not regressions from this PR:

- **`k3-sglang-test`** is failing on plain `upstream/dev` HEAD (`9793d8b0`) → [build #189](https://buildkite.com/lmcache/k3-sglang-test/builds/189). It was already failing on this PR's previous commit `d20d8c95` ([build #175](https://buildkite.com/lmcache/k3-sglang-test/builds/175)) before this round of changes.
- **`k3-correctness-test`** is also failing on `upstream/dev` HEAD → [build #3397](https://buildkite.com/lmcache/k3-correctness-test/builds/3397). It wasn't wired up to PR runs at the time of `d20d8c95` (only ran on a subset of pipelines back then), so we don't have a direct before/after comparison from this branch — but the dev-branch failure shows the issue exists independently.

The correctness-test failure mode in #3395 is `[FAIL] man bash output mismatch!` in the `reset_prefix_cache → half_context → full_context` sequence — an end-to-end output-equality check that depends on vLLM + LMCache + the model used by the test harness. Since the same sequence fails on `upstream/dev` without any of this PR's changes, it points at something in the harness or environment rather than the completion-callback refactor.

Happy to dig into the harness if you can grant access to the Buildkite logs / re-run on a clean dev to confirm.

sammshen (approved):
LGTM

PR #3338, opened 2026-05-20
[optimize] rename non_cuda_equivalents to python_ops_fallback

chunxiaozheng:
The name `non_cuda_equivalents.py` is very confusing. In fact, the function of this module is to implement fallback in `Python`. It can work normally when a native `c++` implementation cannot be found, and `CUDA` will be used.


hlin99:
thanks @chunxiaozheng for pointing this out. yes, it's confusing name to the content. it's time to make it clear. LGTM

hlin99 (approved):

maobaolong (approved):
@chunxiaozheng This name refactor make more sense to me! LGMT, thanks for this effort!

PR #3340, opened 2026-05-20
fix(v1): graceful skip on Scheduler-role abort in request_finished (fixes #3337)

### LINKED ISSUE #3337: [Bug] AssertionError in vllm_v1_adapter.py on request abort in vLLM V1 Scheduler
### Label

   bug 
  ──────
  ### Describe the bug

  Under high concurrent load in a vLLM V1 setup (using  LMCacheConnectorV1  and  LMCache 0.4.5 ), the vLLM  EngineCore  process fatally crashes with an  AssertionError  in  vllm_v1_adapter.py:request_finished  during
  request abort cleanups.

  This happens because in vLLM V1's split-process architecture, the Scheduler runs its own connector/adapter instance but does not initialize the underlying LMCache storage engine ( self.lmcache_engine  is  None  for
  the
  Scheduler role, as it is only needed on the Worker processes). When a request is aborted, the scheduler calls  request_finished() , which hits a hard assertion  assert self.lmcache_engine is not None  and crashes the
  entire engine.
  ──────
  ### To Reproduce

  1. Deploy vLLM with in-process LMCache enabled on a GKE cluster running vLLM V1 architecture (vLLM >=  0.21.0  and LMCache  0.4.5 ).
  2. Start a distributed load test (e.g., using Locust swarming) to send highly concurrent prompts.
  3. Trigger request abortions (e.g., by disconnecting clients, stopping the load test, or triggering network timeouts).
  4. Observe the  EngineCore  process crash immediately with an  AssertionError  from  vllm_v1_adapter.py  line 1779.
  ──────
  ### Expected behavior

  The Scheduler's adapter should safely bypass the storage manager cleanup if  self.lmcache_engine  is  None , rather than throwing a fatal  AssertionError  and crashing the server process.
  ──────
  ### Screenshots

  N/A
  ──────
  ### Desktop (please complete the following information):

  • OS: Linux (GKE container environment)
  • Python Version:  3.12 
  • vLLM Version:  0.21.0  (V1 Engine Core)
  • LMCache Version:  0.4.5 
  ──────
  ### Smartphone (please complete the following information):

  N/A
  ──────
  ### Additional context

  #### Stack Trace of the Crash:

    (EngineCore pid=147) Process EngineCore:
    (EngineCore pid=147) Traceback (most recent call last):
    (EngineCore pid=147)   ...
    (EngineCore pid=147)   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/core/sched/scheduler.py", line 1755, in _free_request
    (EngineCore pid=147)     connector_delay_free_blocks, kv_xfer_params = self._connector_finished(request)
    (EngineCore pid=147)   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/core/sched/scheduler.py", line 1960, in _connector_finished
    (EngineCore pid=147)     return self.connector.request_finished(request, block_ids[0])
    (EngineCore pid=147)   File "/usr/local/lib/python3.12/dist-packages/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 340, in request_finished
    (EngineCore pid=147)     return self._lmcache_engine.request_finished(request, block_ids)
    (EngineCore pid=147)   File "/usr/local/lib/python3.12/dist-packages/lmcache/integration/vllm/vllm_v1_adapter.py", line 1779, in request_finished
    (EngineCore pid=147)     assert self.lmcache_engine is not None
    (EngineCore pid=147) AssertionError

  #### Proposed Patch (Git Diff):

 verified that replacing the assertion with a safe null check completely resolves the crash and allows successful continuous swarming under heavy concurrent load:

    diff --git a/lmcache/integration/vllm/vllm_v1_adapter.py b/lmcache/integration/vllm/vllm_v1_adapter.py
    index 7a7934b..ga7934b 100644
    --- a/lmcache/integration/vllm/vllm_v1_adapter.py
    +++ b/lmcache/integration/vllm/vllm_v1_adapter.py
    @@ -1776,7 +1776,7 @@ class LMCacheV1Adapter:
             # Cleanup if request was aborted
             if request.status == RequestStatus.FINISHED_ABORTED:
		 # Notify storage backends of aborted requests
    -            assert self.lmcache_engine is not None
    -            sm = self.lmcache_engine.storage_manager
    -            if sm is not None:
    -                sm.cancel_request(request.request_id)
    +            if self.lmcache_engine is not None:
    +                sm = self.lmcache_engine.storage_manager
    +                if sm is not None:
    +                    sm.cancel_request(request.request_id)

abinggo:
## Summary

`request_finished` is a Scheduler-side connector API. The Scheduler typically does **not** initialize the storage engine — only the Worker role does so by default; the Scheduler also builds it when `enable_scheduler_bypass_lookup` is set. Under concurrent abort cleanups in the Scheduler the `FINISHED_ABORTED` branch hit `assert self.lmcache_engine is not None` as an unhandled `AssertionError` and killed the entire `EngineCore` process — every connected user's stream got dropped.

This PR replaces both asserts in that block with `if-warn-skip` so the engine stays alive and only the affected request's cleanup is dropped:

- `lmcache_engine is None` → warn + skip storage-backend cleanup (the path the bug report hits)
- `lookup_client is None` while `async_loading=True` → warn + skip the lookup cancel (sibling assert in the same block — same fail-soft contract)

Matches `docs/coding_standards.md`: *"No `assert` for validation — use `if/raise`"*.

## Fixes

- #3337

## Acknowledgement

Thanks to @annguy3n for the precise root-cause analysis (Scheduler vs. Worker role split in vLLM v1's split-process architecture), the reproduction recipe (Locust swarm against vLLM V1 + LMCache 0.4.5 on GKE), and the patch sketch in the issue body. This PR keeps the same fix shape and extends it to the sibling `lookup_client` assert in the same abort block.

## Why `if-else` instead of the reporter's nested `if`

The bug report's diff guards `storage_manager.cancel_request` with `if engine is not None`. That fixes the reported crash, but if the same Scheduler-role configuration ever runs with `async_loading=True`, the next abort would crash again at the sibling `assert self.lookup_client is not None` 8 lines below. Wrapping the whole abort block with `if engine is None: warn / else:` lets both paths fail-soft on the same root cause, then defends the inner sibling assert independently.

## Test plan

New CPU-only regression test: `tests/v1/test_v1_adapter_request_finished_scheduler_role.py`. Mirrors the fake-fixture style of `tests/v1/test_vllm_layerwise_wait_for_save.py` — bypasses `__init__` via `__new__` and pins only the attributes `request_finished` reads. Four cases:

1. **Scheduler default** (`engine=None`): aborts no longer crash; warning names the request id and `lmcache_engine`.
2. **Engine initialized** (Worker role, or Scheduler with `enable_scheduler_bypass_lookup`): `storage_manager.cancel_request` still runs — guards against the fix accidentally short-circuiting the happy path.
3. **`async_loading=True` + `lookup_client` present**: sibling `lookup_client.cancel_lookup` still runs — regression guard for the second assert in the same block.
4. **`async_loading=True` + `lookup_client` missing**: warns and skips instead of crashing — matches the new fail-soft contract.

I cannot run this test locally (this dev Mac has no `vllm` wheel and no `lmcache.c_ops` extension, so the file `pytest.importorskip`'s out). Relying on LMCache CI.

## Out of scope (intentional)

The same `assert self.lmcache_engine is not None` anti-pattern still appears in 5 other call sites in this file (lines 535, 766, 988, 1083, 1105) and `assert self.lookup_client is not None` in 2 more (1350, 1462). They sit on different code paths than the abort cleanup the bug report hits — keeping the diff scoped to the reported failure mode keeps the review surface tight. Happy to follow up with a sweep PR if maintainers prefer to scrub the rest of the file in one go.

Adjacent: companion PR #3325 covers the `wait_for_save` (line ~1124) instance of the same anti-pattern, on the save path rather than the abort path.


abinggo:
Thanks @gemini-code-assist — both points were correct, force-pushed `9507ba07` addressing them:

**[high] test fixture `AttributeError` on read-only properties** — `lmcache_engine` and `lookup_client` delegate to `self._manager` (line 601-614). Switched the fixture to install `connector._manager = SimpleNamespace(lmcache_engine=engine, lookup_client=lookup_client)` instead of assigning to the properties directly.

**[medium] async lookup cancel skipped on Scheduler-side aborts** — correct catch. `vllm_service_factory.maybe_create_lookup_client` (line 250) only creates the lookup client on the Scheduler, so on a Scheduler default abort `engine=None` while `lookup_client` is still live and there may be in-flight async lookups. Moved the `if self.async_loading:` block out of the engine-not-None branch so storage cleanup and lookup cancel are now two parallel paths under the abort block. Added a regression test case (Scheduler default + `async_loading=True` + `lookup_client` present → `cancel_lookup` must still fire) to lock that in.

The diff is now `+270 / -8`; commit message updated to reflect the parallel-path layout. DCO passes.

ApostaC (commented):
I doubt this is really needed.
I vaguely remember that at some point, we introduced a dummy "lmcache engine" on the scheduler side to do some special work, and it's enabled by default. @maobaolong do you remember it?

milesial:
Ran into an issue with vLLM `stop` word crashing the whole server because of request cancellation with LMCache and this PR solved it

abinggo:
Thanks for confirming this in production, @milesial — an independent repro on the vLLM `stop`-word cancellation path is a useful data point on top of #3337's explicit-abort case.

ApostaC (approved):
Took a look again. I think it looks good. Nice to hear that it fixes a problem in prod.

ApostaC:
@sammshen @maobaolong can you also help take a look?

abinggo:
@sammshen Friendly ping for codeowner review on this small fix. ApostaC already approved (#pullrequestreview-4422875663) and confirmed it fixes a real prod issue, but the merge is blocked on `/lmcache/integration/vllm/` codeowner approval. PR is +1-line core fix + a CPU-only regression test; CI (DCO + buildkite k3/amd) is green. Happy to address any feedback. Thanks!

sammshen (approved):
LGTM

EdHasNoLife:
**Independent repro — still present on `dev` as of 2026-06-14.** Thanks @abinggo for the fix.

- **Env:** vLLM 0.22.1 (v1, multiprocess `EngineCore`) + LMCache `0.4.7-g307439a1`, `LMCacheConnectorV1` (read+write, remote `lm://` backend), dense Qwen3-4B, single A10.
- **Minimal trigger — no load test needed:** start one streaming `/v1/chat/completions`, then close the client socket mid-stream. vLLM aborts the request → `request_finished` → `vllm_v1_adapter.py:1849 assert self.lmcache_engine is not None` → `AssertionError` → `EngineDeadError`; container `restartCount` +1. Deterministic: **1 abort = 1 crash**. Steady-state serving (no abort) is unaffected.
- **Root cause confirmed:** `self.lmcache_engine is None` is correct-by-design on the **scheduler-role** connector — `_init_connector_state` sets only `_unfinished_requests` for `KVConnectorRole.SCHEDULER`, while the engine is a worker-role object. The `FINISHED_ABORTED` branch is the only site that hard-asserts instead of guarding (cf. the `if self.lmcache_engine is None` guard on the legacy-register path). This PR's sibling guard on `assert self.lookup_client is not None` (async-loading path, ~line 1857) is also needed — same scheduler-role `None`.

Approved + CI-green + mergeable since May 20; +1 for codeowner merge on `/lmcache/integration/vllm/`. A routine client disconnect crashing the entire `EngineCore` (and every co-scheduled request) is a serious availability issue for any LMCache + vLLM-v1 deployment.

abinggo:
Thanks @EdHasNoLife for the independent repro — that's now two separate prod reports of the same 1-abort-=-1-`EngineCore`-crash.

@sammshen @maobaolong this is approved by @ApostaC and CI-green / mergeable since May 20. It's a one-line guard on the scheduler-role connector (mirroring the existing `if self.lmcache_engine is None` guard on the legacy-register path) plus a CPU-only regression test. Blocked only on `/lmcache/integration/vllm/` codeowner merge. Could one of you help land it? A routine client disconnect taking down the whole `EngineCore` (and every co-scheduled request) is a real availability hit. Happy to rebase if anything's needed.

lazariv:
Confirming this fixes a real, reproducible crash on the released 0.4.5 too.

Setup: vLLM 0.22.0 + LMCache 0.4.5, `LMCacheConnectorV1` (`kv_role=kv_both`), TP=4, GDS(cufile) + `local_cpu` backends, `enable_async_loading=False`.

Symptom: any aborted request (client disconnects mid-stream / cancel / timeout) reaches `request_finished()` on the `FINISHED_ABORTED` path, where assert `self.lmcache_engine is not None` fails - `lmcache_engine is None` in the scheduler/EngineCore process (it only lives in the workers). That raises AssertionError → EngineCore encountered a fatal error → the whole server dies, and every later request returns HTTP 500 / EngineDeadError. A single mid-stream disconnect is enough to take the server down, which is fatal in production where clients disconnect constantly.

After cherry-picking this PR onto 0.4.5: fired 8 consecutive abrupt mid-stream disconnects → engine stays healthy, no AssertionError/traceback in the logs. 

Thanks for the fix, would be great to get this merged.

PR #3341, opened 2026-05-20
[refactor] refactor bench engine cli

chunxiaozheng:
# DESCRIPTION
Refactor `lmcache bench engine` cli, current implementation put all methods in `__init__.py`, which makes `__init__.py` is very fat😿, the refactor move the related implementation to a single file(`command.py` in `bench_engine` module), which can make `__init__.py` more clear.   

# TODO
1. `lmcache bench kvcache` also need to refactor, and I will submit a PR later.

PR #3349, opened 2026-05-20
 [CI] Restore NO_CUDA_EXT=1 skip-all-extensions semantics

deng451e:
## Summary
  - `setup.py`: short-circuit `_collect_extensions()` on `NO_CUDA_EXT=1` before calling `_common_cpp_extensions()`, so the CLI wheel builds  without torch and ships zero compiled extensions.
  - `server.py`: replace the server-import `print` with a `logger.warning` that says this is the lightweight `lmcache-cli` install and the full package is needed for `server` / `bench`.

  ## Context
  #3302 redefined `NO_CUDA_EXT=1` to "skip GPU only", but pure C++ extensions still require `torch.utils.cpp_extension` at build time — which  the CLI wheel env doesn't have. Every PR opened after #3302 (incl. #3341) started failing `Build CLI Artifacts` with `ModuleNotFoundError: No module named 'torch'`. This restores the prior "skip all extensions" meaning that the CLI smoke test (`grep '\.(so|pyd)$'` → fail) requires. 
 #3302's `_common_cpp_extensions()` refactor is preserved.

  ## Follow-up
 The name `NO_CUDA_EXT` is misleading — it has always controlled *all* native extensions, not just CUDA ones. A follow-up PR will rename it to `NO_NATIVE_EXT` with a backwards-compatible alias (`NO_CUDA_EXT` keeps working, emits a deprecation warning) to reduce the chance of this regression recurring.

  ## Test plan
  - [x] CLI wheel builds with `NO_CUDA_EXT=1` in a torch-less venv; `unzip -l` shows no `.so`/`.pyd`
  - [x] `lmcache --help` and all 8 slim subcommand `--help`s pass on both `dev` and #3341 + this fix
  - [x] `NO_CUDA_EXT=1 ... --sdist` still works; full CUDA/HIP/SYCL paths byte-identical when `NO_CUDA_EXT` is unset

### DIFF
--- a/lmcache/cli/commands/server.py
+++ b/lmcache/cli/commands/server.py
@@ -6,6 +6,9 @@ import argparse
 
 # First Party
 from lmcache.cli.commands.base import BaseCommand
+from lmcache.logging import init_logger
+
+logger = init_logger(__name__)
 
 
 class ServerCommand(BaseCommand):
@@ -53,9 +56,10 @@ class ServerCommand(BaseCommand):
             add_http_frontend_args(parser)
             add_observability_args(parser)
         except ImportError as e:
-            print(
-                f"Failed to import server dependencies: {e}. "
-                "Install the full lmcache package to use 'lmcache server'."
+            logger.warning(
+                "lmcache-cli (lightweight) detected (%s); install the full "
+                "lmcache package to use 'server' and 'bench'.",
+                e,
             )
             return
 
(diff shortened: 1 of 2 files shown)


ApostaC (approved):
LGTM!

tensormesh (approved):
LGTM

sammshen (approved):
LGTM!

PR #3356, opened 2026-05-21
fix: Prevent Lock Resource Leakage and Potential Deadlocks in L2 Lookup Benchmark

weizhoublue:

 Trigger Scenario: 
Running the lmcache bench l2 CLI command to benchmark lookup performance (e.g., using --lookup-max-hit-rate > 0) with an L2 adapter  that enforces actual locking mechanisms (like Redis, DAX, or Raw Block adapters).

Fault Manifestation: 
Locks acquired on the cached objects during the benchmark are never released. This lock leakage can cause the adapter to exhaust  its lock resources or cause subsequent operations (which attempt to lock the same keys) to block indefinitely or fail during the process lifecycle.

**If applicable**:


ApostaC (approved):
Makes a lot of sense to me.

@chunxiaozheng Can you take a second look?

PR #3359, opened 2026-05-21
feat(mp): SHM-based data transfer path for GPUs/CPU/Accelerators

hlin99:
## Summary

Add a shared-memory (SHM) data transfer path for non-CUDA (XPU/CPU) workers
in multiprocess mode, complementing the existing pickle-based path introduced
in [#3259.](https://github.com/LMCache/LMCache/pull/3259)

When the LMCache server has a POSIX SHM pool available, non-GPU workers now
use zero-copy memory-mapped buffers instead of pickle serialization, reducing
store/retrieve latency and CPU overhead.

Pickle Path：4 copy
Shm Path：1 copy

## Changes

- **`non_gpu_context_shm.py`** — `NonGpuContextShm` implementation: maps the
  server's SHM pool into the worker process, provides pre-allocated tensor
  views for store (zero-copy gather) and retrieve (zero-copy scatter).
- **`server.py`** — `register_kv_cache_non_gpu_context` returns SHM pool info
  (`shm_name`, `pool_size`) when available; `prepare_store` reserves slots and
  returns per-chunk SHM offsets; partial-reservation support (only gather
  chunks not already cached).
- **`non_gpu_context.py`** — `create_non_gpu_context()` factory: selects SHM
  vs pickle based on server response; `gather_paged_kv_to_cpu` gains
  `chunk_indices` parameter for partial gather.
- **`transfer_context.py`** — `DataTransferContext.submit_store` handles the
  `(tensors, chunk_indices)` tuple from `prepare_store`, short-circuits when
  all chunks are already cached.
- **`protocols/engine.py`** — `RegisterNonGpuContextResponse` carries
  `shm_name` and `pool_size`.
- **Design doc** — `docs/design/v1/multiprocess/non_gpu_context_design.md`
- **Tests** — unit tests for SHM context, partial gather, server slot
  reservation, and unregister cleanup.

## How it works

```
Worker                          Server
  │                               │
  ├─ register() ─────────────────►│ returns shm_name + pool_size
  │  mmap(shm_name)               │
  │                               │
  ├─ prepare_store(key) ─────────►│ reserve_write → slots + chunk_indices
  │  gather only chunk_indices     │
  │  into pre-allocated SHM views  │
  ├─ commit_store(key) ──────────►│ finish_write (zero-copy, data already in SHM)
  │                               │
  ├─ prepare_retrieve(key) ──────►│ read_prefetched → SHM offsets
  │  scatter from SHM views        │
  ├─ commit_retrieve(key) ───────►│ finish_read
```

## Backward compatibility

- Workers connecting to a server **without** SHM pool automatically fall back
  to pickle (unchanged behavior from #3259).
- No changes to the CUDA IPC path.

Signed-off-by: Tony Lin <tony.lin@intel.com>


chunxiaozheng:
@hlin99 hi, Tony, https://github.com/LMCache/LMCache/pull/3352 also implements SHM-based data transfer, I feel that there are some conflicts between these two PRs?

hlin99:
> @hlin99 hi, Tony, https://github.com/LMCache/LMCache/pull/3352 also implements SHM-based data transfer, I feel that there are some conflicts between these two PRs?
> 
> 

hi @chunxiaozheng thanks for pointing this out. I guess this is the question for majority of people. actually, this was discussed with @maobaolong before
😀

this pr is aiming the function working pipeline for GPU(cuda or none cuda) and CPU as well. worker side copy is generally doable for all the devices

while, 3352 is aiming on CPU pipeline that simulate the cuda path using stream/event… @maobaolong would like to have such pipeline so developers can just use CPU for coding cuda alike workflow. functionalities are enabled on 3352 for CPU only, not for GPU

the purposes of PRs are different and have their own reasons.

@maobaolong feel free to correct me if anything is wrong

lmcache/v1/multiprocess/non_gpu_context_shm.py
```diff
@@ -0,0 +1,157 @@
...
+        pool_size: int,
+    ) -> None:
+        super().__init__(metadata, mq_client, mq_timeout)
+        if not shm_name or pool_size <= 0:
+            raise ValueError("shm_name must be non-empty and pool_size must be > 0")
+
+        self._shm_name = shm_name
+        self._pool_size = pool_size
+        self._shm_fd = INVALID_SHM_FD
+        shm_path = os.path.join("/dev/shm", shm_name.lstrip("/"))
```
maobaolong: This can only work for linux, right?

lmcache/v1/multiprocess/non_gpu_context_shm.py
```diff
@@ -0,0 +1,157 @@
...
+    ) -> tuple[list[torch.Tensor], list[int]] | None:
+        future = self.mq_client.submit_request(
+            RequestType.PREPARE_STORE,
+            [key, instance_id],
+            get_response_class(RequestType.PREPARE_STORE),
+        )
+        try:
+            response = future.result(timeout=self.mq_timeout)
+        except TimeoutError:
+            return None
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/server.py
```diff
@@ -348,13 +385,22 @@ def register_kv_cache_non_gpu_context(
...
+                    payload.instance_id,
+                )
+                if existing_context.shm_active:
+                    pool_info = self.storage_manager.get_shm_pool_info()
+                    if isinstance(pool_info, dict):
+                        return RegisterNonGpuContextResponse(
+                            shm_name=str(pool_info.get("shm_name", "")),
+                            pool_size=int(pool_info.get("pool_size", 0)),
+                        )
+                return RegisterNonGpuContextResponse()
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -317,18 +329,25 @@ def submit_store(
...
         cpu_chunks = gather_paged_kv_to_cpu(
             kv_caches,
             block_ids,
             blocks_in_chunk,
             layout_hints=self._layout_hints,
             gpu_kv_format=self._gpu_kv_format,
             out=out_buffers,
+            chunk_indices=chunk_indices,
         )
         ok = self._non_gpu_context.commit_store(key, instance_id, cpu_chunks)
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/distributed/memory_manager.py
```diff
@@ -16,7 +20,23 @@
...
 
-# HELPER FUNCTIONS
+def _unlink_stale_shm(shm_name: str) -> None:
+    """Remove a stale LMCache shm segment if it exists."""
+    normalized = shm_name.lstrip("/")
+    if "/" in normalized or "\\" in normalized:
+        logger.warning("Refusing to unlink invalid shm name %s", shm_name)
+        return
+    if not normalized.startswith("lmcache_l1_pool_"):
+        return
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/non_gpu_context_shm.py
hlin99: hi @maobaolong  can you try the latest? issue is supposed to be fixed.

maobaolong:
> > @hlin99 hi, Tony, #3352 also implements SHM-based data transfer, I feel that there are some conflicts between these two PRs?
> 
> hi @chunxiaozheng thanks for pointing this out. I guess this is the question for majority of people. actually, this was discussed with @maobaolong before 😀
> 
> this pr is aiming the function working pipeline for GPU(cuda or none cuda) and CPU as well. worker side copy is generally doable for all the devices
> 
> while, 3352 is aiming on CPU pipeline that simulate the cuda path using stream/event… @maobaolong would like to have such pipeline so developers can just use CPU for coding cuda alike workflow. functionalities are enabled on 3352 for CPU only, not for GPU
> 
> the purposes of PRs are different and have their own reasons.
> 
> @maobaolong feel free to correct me if anything is wrong

@hlin99 Thanks for the introduction of these two PRs. Both the two PRs are useful, as conclusion,
- #3359(This PR): This PR is a SHM-based data transfer approach, compare to `pickle approach`, `SHM-based` approach should reduce serialization and deserialization, it will have better performance. And mostly, is can be used for production case. But this is different with previous CUDA path.
- #3352(My PR): That PR is a SHM-based approach too, but the data copy is taken place to server-side, as @hlin99 motioned , in a sense, it is a cuda simulator, all the register_kvcache/retrieve/store path are same with CUDA above the hardware layer, so new contributor can develop and verify the features or bugfix without any hardware.

@chunxiaozheng Hope @hlin99 and me explain clearly.

chunxiaozheng:
> > > @hlin99 hi, Tony, #3352 also implements SHM-based data transfer, I feel that there are some conflicts between these two PRs?
> > 
> > 
> > hi @chunxiaozheng thanks for pointing this out. I guess this is the question for majority of people. actually, this was discussed with @maobaolong before 😀
> > this pr is aiming the function working pipeline for GPU(cuda or none cuda) and CPU as well. worker side copy is generally doable for all the devices
> > while, 3352 is aiming on CPU pipeline that simulate the cuda path using stream/event… @maobaolong would like to have such pipeline so developers can just use CPU for coding cuda alike workflow. functionalities are enabled on 3352 for CPU only, not for GPU
> > the purposes of PRs are different and have their own reasons.
> > @maobaolong feel free to correct me if anything is wrong
> 
> @hlin99 Thanks for the introduction of these two PRs. Both the two PRs are useful, as conclusion,
> 
> * [feat(mp): SHM-based data transfer path for GPGPUs/CPU #3359](https://github.com/LMCache/LMCache/pull/3359)(This PR): This PR is a SHM-based data transfer approach, compare to `pickle approach`, `SHM-based` approach should reduce serialization and deserialization, it will have better performance. And mostly, is can be used for production case. But this is different with previous CUDA path.
> * [feat(mp): add SHM-based NonGpuContext (server-side copy)  #3352](https://github.com/LMCache/LMCache/pull/3352)(My PR): That PR is a SHM-based approach too, but the data copy is taken place to server-side, as @hlin99 motioned , in a sense, it is a cuda simulator, all the register_kvcache/retrieve/store path are same with CUDA above the hardware layer, so new contributor can develop and verify the features or bugfix without any hardware.
> 
> @chunxiaozheng Hope @hlin99 and me explain clearly.

@hlin99 @maobaolong Thanks for your explanation, I already understand! 👍

maobaolong:
@hlin99 Good news to you, i've add macos smoke test to the dev branch, and the ci workflow show that all the checks of macos are passed.

hlin99:
> @hlin99 Good news to you, i've add macos smoke test to the dev branch, and the ci workflow show that all the checks of macos are passed.

nice👍👍👍

lmcache/v1/distributed/memory_manager.py
```diff
@@ -65,6 +119,13 @@ def __init__(self, config: L1MemoryManagerConfig):
         self._allocator = create_memory_allocator(config)
         self._size_in_bytes = config.size_in_bytes
         self._align_bytes = config.align_bytes
+        self._shm_pool_info = {"shm_name": "", "pool_size": 0}
+        if isinstance(self._allocator, MixedMemoryAllocator):
+            if self._allocator.shm_name:
+                self._shm_pool_info = {
+                    "shm_name": self._allocator.shm_name,
+                    "pool_size": self._size_in_bytes,
+                }
```
ApostaC: Does this mean the current shm solution only works with `MixedMemoryAllocator`? In this case, we need to make explicitly check the exclusiveness with `--l1-use-lazy` (which is True by default).

ApostaC: Additionally, is it possible to keep the `shm_pool_info` at the transfer-context or MPCacheEngine level? They may be able to see the shm name and the shm size as well? In this case, we don't need to introduce `get_shm_pool_info` here.

lmcache/v1/multiprocess/server.py
```diff
@@ -419,7 +532,50 @@ def prepare_store(
             PrepareStoreResponse with empty slots for pickle mode.
         """
```
ApostaC: The docstring of `prepare_store`, `prepare_retrieve`, `commit_store`, and `commit_retrieve` needs to be updated. 
I think it's important to put not just the behavior but the high-level semantics in the docstring, for example (based on my understanding):
- `prepare_store`: prepare the memory for the store operation on the server side and return a list of writeable memory "references" which can be understood by the transfer context. After calling this function, the serving engine can start writing data into LMCache.
- `commit_store`: mark the data copying on the serving engine side is fully finished.
(And for `retrieve` functions as well)

lmcache/v1/multiprocess/server.py
ApostaC, line 1: I feel like right now there are a lot of if-branches just for shm. Is it possible to avoid those if-branches?
(I'm asking this because I want to avoid having more if branches if we want to support different non-GPU transfer modes in the future.)

lmcache/v1/multiprocess/transfer_context/shm.py
```diff
@@ -0,0 +1,165 @@
...
+            [key, instance_id],
+            get_response_class(RequestType.PREPARE_RETRIEVE),
+        )
+        try:
+            response = future.result(timeout=self.mq_timeout)
+        except TimeoutError:
+            return None
+        if not response.success:
+            return None
+        slots = response.context.get("slots", [])
```
ApostaC, line 197: Is it possible to clearly define the format of `slots`? It's for better future maintainability.

lmcache/v1/multiprocess/transfer_context/shm.py
ApostaC, line 1: Since we are having multiple different transfer contexts, why don't we have a separate folder to keep the code layout cleaner?

ApostaC (commented):
Did a quick look at the code structure and got a few quick comments.

At a high-level, I feel like it is possible to have a better separation of the functions in `MPCacheEngine`. I'm trying to do some refactoring for this part as well (since it's also needed by the p2p module).
The idea is to have separate modules that can be "registered" into MPCacheEngine (and enabled/disabled separately), where every module keeps its own metadata and control flow and shares the same storage manager.

KuntaiDu:
@hlin99 This PR seems to put all kv caches in `/dev/shm`, but `/dev/shm` capacity is typically limited in docker-based deployment. Can we only use `/dev/shm` as a staging buffer, and host the KV caches in non-shared CPU memory?

cc @ApostaC

hlin99:
> @hlin99 This PR seems to put all kv caches in `/dev/shm`, but `/dev/shm` capacity is typically limited in docker-based deployment. Can we only use `/dev/shm` as a staging buffer, and host the KV caches in non-shared CPU memory?
> 
> cc @ApostaC

hi, @KuntaiDu  Thanks for the concern! The /dev/shm usage here is intentional and was discussed with @ApostaC beforehand — the design places L1 cache in SHM so that data transfer between worker and server only requires a single copy (zero-copy view). In contrast, the pickle fallback path requires 4 copies (GPU → CPU tensor → serialize → deserialize → CPU tensor), which is why SHM is the preferred mode when capacity allows.

You're right that there are cases where `/dev/shm` capacity may be insufficient for L1. A fallback path (pickle over ZMQ) is already implemented in #3259 (which you reviewed last week). The fallback can be:

- **Manually triggered**: `--shm-name=""` routes to pickle transport
- **Auto-detected**: `_resolve_shm_config` checks `/dev/shm` free space at startup and disables SHM automatically if capacity is insufficient

So in Docker environments with limited `/dev/shm`, the system gracefully degrades without user intervention.

Hope this background info addresses your concern! Let me know if there are other aspects I might not have considered.

hlin99:
> Did a quick look at the code structure and got a few quick comments.
> 
> At a high-level, I feel like it is possible to have a better separation of the functions in `MPCacheEngine`. I'm trying to do some refactoring for this part as well (since it's also needed by the p2p module). The idea is to have separate modules that can be "registered" into MPCacheEngine (and enabled/disabled separately), where every module keeps its own metadata and control flow and shares the same storage manager.

thanks @ApostaC for the good suggestion and refactor by https://github.com/LMCache/LMCache/pull/3391  will rebase after https://github.com/LMCache/LMCache/pull/3391 gets merged.

lmcache/v1/distributed/memory_manager.py
hlin99: > Does this mean the current shm solution only works with `MixedMemoryAllocator`? In this case, we need to make explicitly check the exclusiveness with `--l1-use-lazy` (which is True by default).

agree. actually, there's an auto fallback logic already implemented in last PR(#3259).  https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/distributed/config.py#L47

and the doc has been updated, need your approval @ApostaC  https://github.com/LMCache/LMCache/pull/3361

hlin99: > Additionally, is it possible to keep the `shm_pool_info` at the transfer-context or MPCacheEngine level? They may be able to see the shm name and the shm size as well? In this case, we don't need to introduce `get_shm_pool_info` here.

fixed in 09e81e7a95215c33b8b81ecb4611e142a8c9c2d1

lmcache/v1/multiprocess/transfer_context/shm.py
hlin99: fixed in 7809a3e4a2095efcd9199865885f79e22c4c3d67

lmcache/v1/multiprocess/transfer_context/shm.py
hlin99: fixed. please check the latest code structure. thanks

lmcache/v1/multiprocess/server.py
hlin99: fixed in 8492a75c701fbff7204131943a25cfe92f21da57

lmcache/v1/multiprocess/server.py
hlin99: fixed in 7e87a7a4005b9af4b0f845ad4de0da0c8a2729e8

hlin99:
FYi. rebasing in progress.. not fully done yet

DongDongJu:
FYI we can use cxl switched devdax for this purpose later.

hlin99:
Hi @ApostaC 

I've finished rebasing onto the latest refactored codebase from #3391, with additional code changes to address your earlier comments on this PR. Would you mind taking another look?

## Note on Transfer Mode

PR #3391 introduced the `transfer-mode` option (`gpu` / `non_gpu`), which is great. In this PR, I added an **`auto`** mode as the default. Here's the rationale:

- The transfer mode must match between the **worker** and the **server**.
- The `transfer-mode` config controls the server side but cannot manage the worker side.
- On the worker side today, only CUDA devices use the GPU path; all other devices use the non-GPU path.
- If `transfer-mode` is set to `gpu` on the server for any reason, the pipeline will break for non-CUDA workers.

The **`auto`** mode enables the server to support both GPU and non-GPU paths simultaneously. This means that if a node has two types of devices, each can use its appropriate path. The use case is uncommon, but it is architecturally supported.

For the **worker** side to control the transfer mode, we would need to introduce a new KV connector argument. I'm not sure that's warranted — please share your thoughts on the best approach. Thanks!


below is a code layout to help you quickly understand code structure.


## Code Layout

```text
Worker Adapter (vLLM MP Adapter)
  └─ TransferContext (adapter_connector/worker_transfer.py)
      ├─ HandleTransferContext  (CUDA IPC path)
      └─ DataTransferContext    (non-CUDA data path)
          └─ NonGpuContext (adapter_connector/base.py)
              ├─ NonGpuContextPickle (adapter_connector/pickle.py)
              └─ NonGpuContextShm    (adapter_connector/shm.py)

MPCacheEngine (Server)
  └─ MPCacheEngineContext (engine_context.py)
      ├─ StorageManager
      ├─ TokenHasher
      ├─ SessionManager
      ├─ EventBus
      ├─ LayoutDescRegistry
      └─ shm_pool_info (pre-computed once)
  └─ NonGPUTransferModule (modules/non_gpu_transfer.py)
      └─ TransferStrategy (modules/server_transfer.py)
          ├─ PickleTransferStrategy
          └─ ShmTransferStrategy
```

lmcache/v1/distributed/memory_manager.py
```diff
@@ -16,7 +19,26 @@
 logger = init_logger(__name__)
 
 
-# HELPER FUNCTIONS
```
ApostaC, line 19: nit: I think we can still keep this line

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -15,6 +15,11 @@
 from lmcache.integration.request_telemetry.factory import RequestTelemetryFactory
 from lmcache.integration.vllm.utils import vllm_layout_hints
 from lmcache.utils import _lmcache_nvtx_annotate, init_logger
+from lmcache.v1.multiprocess.adapter_connector.worker_transfer import (
+    DataTransferContext,
+    TransferContext,
+    create_transfer_context,
```
ApostaC: These symbols are exported in `lmcache/v1/multiprocess/adapter_connector/__init__.py`, so we can shorten the import here.

In the meantime, I think changing the naming from `adapter_connector` to `transfer_context` sounds more direct. WDYT?

lmcache/v1/multiprocess/adapter_connector/base.py
```diff
@@ -69,8 +72,18 @@ def layout_desc(self) -> MemoryLayoutDesc:
         return self.metadata.layout_desc
 
     @abstractmethod
-    def prepare_store(self, key: Any, instance_id: int) -> list[torch.Tensor] | None:
-        """Prepare store. Returns pre-allocated out buffers (shm) or None (pickle)."""
+    def prepare_store(
+        self, key: Any, instance_id: int
```
ApostaC: quick question: should the type of `key` be `IPCCacheEngineKey`?

Let's avoid using `Any` for better clarity.

lmcache/v1/multiprocess/adapter_connector/base.py
```diff
@@ -69,8 +72,18 @@ def layout_desc(self) -> MemoryLayoutDesc:
...
+    ) -> tuple[list[torch.Tensor], list[int]] | None:
+        """Prepare store.
+
+        Returns a ``(tensors, chunk_indices)`` tuple for SHM mode or ``None``
+        for pickle mode / when no new chunks need writing.
+
+        The ``chunk_indices`` list contains the position of each reserved chunk
+        within the full chunk sequence so that the caller only gathers data for
+        chunks that were actually reserved (i.e. not already present in cache).
```
ApostaC: This is more informative, but I still cannot quickly understand the expected behavior of this function by reading the docstring here.
Is the tensor used for write? What's the meaning of `list` here? (for example, every element in the list corresponding to a key?) How should the caller interpret the `chunk_indices`?

lmcache/v1/multiprocess/transfer_context/base.py
```diff
@@ -100,24 +113,51 @@ def create_non_gpu_context(
     metadata: NonGpuContextMetadata,
     mq_client: Any,
     mq_timeout: float,
+    shm_name: str = "",
+    pool_size: int = 0,
```
ApostaC: nit: it's better to have an "explicit" argument (rather than relying on the default value) to mark whether to use shm mode or pickle mode.

lmcache/v1/multiprocess/adapter_connector/shm.py
```diff
@@ -0,0 +1,220 @@
...
+            return None
+        if not slots:
+            # Server explicitly signals all chunks are already cached.
+            return [], []
+        # chunk_indices is always present when the server has this fix applied.
+        # The fallback to range(len(slots)) preserves correctness with an older
+        # server: if all returned slots happen to be contiguous from index 0 the
+        # result is the same; if not, the IndexError we are fixing would still
+        # occur.  In practice the server is always updated together with the
+        # client, so the fallback is a defensive guard only.
```
ApostaC: I don't think we need to handle "backward compatibility" here. This is because both client and server side logic are now fully controlled by LMCache. There should not be a case where client (on the serving engine side) and the LMCache server have 2 different versions.

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -317,18 +335,28 @@ def submit_store(
...
             gpu_kv_format=self._gpu_kv_format,
             out=out_buffers,
+            chunk_indices=chunk_indices,
         )
+        if out_buffers is not None:
+            # SHM path uses async device->CPU copies; complete them before commit.
+            torch_dev.synchronize()
         ok = self._non_gpu_context.commit_store(key, instance_id, cpu_chunks)
 
-        future: MessagingFuture[bool] = MessagingFuture()
```
ApostaC, line 331: Why do we remove the "bool" type flag here?

ApostaC, line 331: Why do we remove the "bool" type flag here? Seems like `ok`'s type will always be `bool`

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -427,10 +427,10 @@ def end_session(self, request_id: str) -> None:
             )
         )
         if session is None:
-            logger.warning("Session %s not found, skipping touch", request_id)
+            logger.debug("Session %s not found, skipping touch", request_id)
             return
         if session.lookup_ipc_key is None:
-            logger.warning(
+            logger.debug(
```
ApostaC: We still want the log level to be warning here. This is a rare case and indicates some potential errors.

lmcache/v1/multiprocess/modules/non_gpu_transfer.py
```diff
@@ -187,10 +213,28 @@ def register_kv_cache_non_gpu_context(
...
         )
+        strategy: TransferStrategy = create_transfer_strategy(
+            self._ctx.storage_manager,
+            shm_name=shm_name,
+            pool_size=pool_size,
+            pending_writes=self._pending_shm_writes,
+            pending_reads=self._pending_shm_reads,
+            pending_lock=self._pending_shm_lock,
+            transfer_key_factory=self._make_transfer_key,
+        )
```
ApostaC, line 224: Maybe a dumb question: when we enable the shm, will there be any case we still want to use pickle at the same time during the runtime?

lmcache/v1/multiprocess/modules/server_transfer.py
ApostaC, line 1: The interface of this module looks good. I haven't gone through the implementation details yet but I trust it's working.

lmcache/v1/multiprocess/mq.py
```diff
@@ -103,6 +103,8 @@ def msgspec_decode(b_obj: bytes, cls: Any) -> Any:
     if cls in _SPECIAL_ENCODER_DECODERS:
         _, decoder = _SPECIAL_ENCODER_DECODERS[cls]
         return decoder.decode(b_obj)
+    if cls in (bool, int):
+        return cls(msgspec.msgpack.decode(b_obj))
```
ApostaC, line 99: Why do we need this change?

lmcache/v1/multiprocess/server.py
```diff
@@ -169,13 +171,20 @@ def _build_modules(
 
     if mp_config.transfer_mode == "gpu":
         modules.append(GPUTransferModule(ctx))
-    else:
+    elif mp_config.transfer_mode == "non_gpu":
+        modules.append(NonGPUTransferModule(ctx))
+    elif mp_config.transfer_mode == "auto":
+        modules.append(GPUTransferModule(ctx))
         modules.append(NonGPUTransferModule(ctx))
```
ApostaC: Design question: Should we support both non-GPU and GPU transfer at the same time?

ApostaC (commented):
Did a pass. PTAL!

ApostaC:
Can we add something in `k3 multiprocess test` to test a cpu-only vLLM with non-GPU lmcache? Would be great if we have end-to-end CI coverage. 

This could be a follow-up PR.

hlin99:
> Can we add something in `k3 multiprocess test` to test a cpu-only vLLM with non-GPU lmcache? Would be great if we have end-to-end CI coverage.
> 
> This could be a follow-up PR.

yes. agree. that was in my plan too. just putting all the things in one PR makes the PR even more difficult to review :)

lmcache/v1/multiprocess/server.py
hlin99: The answer could be yes or no, depending on whether we want to introduce another extra config in the adapter side. 

Now, the flow would be:

### **Worker Side**
1. Register KV `gpu context` (for CUDA devices) or `non_gpu context` (for non-CUDA devices).
2. The response for the GPU context is straightforward, but the non-GPU context would receive a response to determine if it's the **pickle** path or the **shm** path.

---

### **Server Side**
1. Launch both GPU/non-GPU modules in **auto transfer mode**, as the server is unsure which "register" request would come from the worker.
2. The GPU context will go to the **CUDA IPC** path. For the non-GPU context, the server will determine if it's **pickle** or **shm** (depending on whether the system `shm` is larger than `L1`).

If we want to pick up a single way, we should explicitly set the paired transfer mode on both sides. we now have transfer-mode on server side, but on worker side, we need a similar extra config which is TBD.

this "auto" mode one side avoid making paired configurations on both sides, one the other hand can support both gpu/non_gpu paths in the same time. (i.e. one node running cuda+cpu, or cuda+other devices)

lmcache/v1/multiprocess/modules/non_gpu_transfer.py
hlin99: Runtime no. 
server will determine pickle way or shm way during init. and right now there's no runtime switch in design.

lmcache/v1/multiprocess/mq.py
hlin99: This is a defensive guard. I saw the following crash, so I added the fix.
actually, the fix is only on decode side, but nice to add it on encode side as well. WDYT?

reason: 
This is a known potential issue in the current msgspec encode/decode path. It happend here because `COMMIT_STORE` and `COMMIT_RETRIEVE` use `response_class=bool`, but the same issue can happen anywhere else we decode a primitive `bool`.

`COMMIT_STORE` and `COMMIT_RETRIEVE` declare `response_class=bool`. The server handler returns `True`/`False` to indicate success. However, at the msgpack wire level, `bool` and `int` use distinct encodings:

- `True` → `0xc3` (msgpack bool format)
- `1` → `0x01` (msgpack int format)

`msgspec.msgpack.decode(data, type=bool)` enforces strict format matching. If the response bytes arrive in int format (`0x01`) instead of bool format (`0xc3`), it raises:

```text
(EngineCore pid=88025) Exception in thread mq-client-thread:
(EngineCore pid=88025) Traceback (most recent call last):
(EngineCore pid=88025)   File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
(EngineCore pid=88025)     self.run()
(EngineCore pid=88025)   File "/usr/lib/python3.12/threading.py", line 1010, in run
(EngineCore pid=88025)     self._target(*self._args, **self._kwargs)
(EngineCore pid=88025)   File "/root/hlin76/LMCache/lmcache/v1/multiprocess/mq.py", line 208, in _main_loop
(EngineCore pid=88025)     response = msgspec_decode(b_response[0], cls=response_cls)
(EngineCore pid=88025)                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=88025)   File "/root/hlin76/LMCache/lmcache/v1/multiprocess/mq.py", line 106, in msgspec_decode
(EngineCore pid=88025)     return msgspec.msgpack.decode(b_obj, type=cls)
(EngineCore pid=88025)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=88025) msgspec.ValidationError: Expected `bool`, got `int`
```

lmcache/v1/multiprocess/modules/lookup.py
hlin99: sure, if this is intentional. i'll restore

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
hlin99: Good catch — it's just because future: MessagingFuture[bool] is already declared in the early-return path above (line 342). Python doesn't allow the same type annotation twice in one function scope. I can move the annotation here instead if you prefer, either way works.

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -317,18 +335,28 @@ def submit_store(
             )
 
         torch_dev.synchronize()
-        out_buffers = self._non_gpu_context.prepare_store(key, instance_id)
+        result = self._non_gpu_context.prepare_store(key, instance_id)
+        out_buffers, chunk_indices = result if result is not None else (None, None)
+        # All chunks already in cache — nothing to gather or commit.
+        if chunk_indices is not None and len(chunk_indices) == 0:
+            future: MessagingFuture[bool] = MessagingFuture()
```
hlin99, line 342: here @ApostaC

lmcache/v1/multiprocess/adapter_connector/shm.py
hlin99: agree. Will fix

lmcache/v1/distributed/memory_manager.py
hlin99: yes. overwrite by mistake. will restore

lmcache/v1/multiprocess/adapter_connector/base.py
hlin99: agree. add more details, please check

lmcache/integration/vllm/vllm_multi_process_adapter.py
hlin99: import path has been fixed. thanks

@ApostaC 
regarding to adapter_connector to transfer_context, i'm not sure i got the point. do you mean the folder name adapter_connector to transfer_context ?
I once used transfer_context, but later on, I felt it's a bit confusing as both worker side and server side have transfer context, and had to distinguish each time. that's the reason I rename to adaptor_connector, meaning it's the connection to adapters.... now vllm, and moving forward sglang, trt-llm, etc.

But i'm open to any name of it. actually, i'm not good at naming ..... 😅

lmcache/v1/multiprocess/adapter_connector/base.py
hlin99: good suggestion. removed Any type

lmcache/v1/multiprocess/transfer_context/base.py
hlin99: hi @ApostaC 

The design philosophy here is that the fallback to pickle mode is primarily intended for debugging purposes. End users shouldn't need to worry about whether pickle or SHM is being used — the code automatically switches between them based on environment conditions.

Because of this, I'd prefer to avoid introducing a new user-facing config to toggle between pickle and SHM. Instead, I'm leveraging the shm_name field as a lightweight escape hatch for developers who need to force pickle mode for any reason.

That said, if you see value in exposing this as an explicit user-facing config option, I'm happy to add it. Let me know what you think — thanks!

lmcache/v1/multiprocess/mq.py
ApostaC: I see, then it makes sense to fix on both encode and decode sides. Maybe in a different PR? WDYT?

lmcache/v1/multiprocess/server.py
ApostaC: I see. I'm good with keeping the "auto" mode.

docs/design/v1/multiprocess/non_gpu_context_design.md
```diff
@@ -20,12 +20,25 @@ across non-CUDA backends.
 
 ```text
 Worker adapter (vLLM MP adapter)
-  └─ TransferContext
+  └─ TransferContext (adapter_connector/worker_transfer.py)
       ├─ HandleTransferContext  (CUDA IPC path)
```
ApostaC: not necessarily bined to CUDA (see #3352)

lmcache/v1/multiprocess/server.py
ApostaC: Can rename to "supported_transfer_mode"

ApostaC (approved):
LGTM! Thanks for the great work!

maobaolong (approved):
LGTM! I can't wait to merge this anymore!

PR #3360, opened 2026-05-21
[XPU] Add SYCL CacheGen + RoPE kernels and in-process blender XPU tests

ftian1:
Implements native SYCL/DPC++ kernels for the four CacheGen + RoPE entry points on Intel XPU, exposed through lmcache.xpu_ops and transparently picked up by _get_backend():

- cal_cdf_sycl.cpp        : calculate_cdf
- ac_enc_sycl.cpp         : encode_fast_new
- ac_dec_sycl.cpp         : decode_fast_new / decode_fast_prefsum
- pos_kernels_sycl.cpp    : rotary_embedding_k_fused

This PR is to support kv cache compression and blending on Intel XPU

1)   examples/kv_cache_reuse/local_backends/offload.py (CacheGen only,
      no blending) works fully on XPU. Qwen3-8B/bf16, 541 prompt tokens,
      8 generated tokens: 
        5.80s cold -> 0.43s warm            **13.5x speedup**. 
2)   examples/blend_kv_v1/blend.py runs end-to-end on Intel XPU (Arc Pro
      B60, vLLM 0.17.2rc1, Qwen3-8B/bf16):
      - warmup: 1.19s
      - first (cold blend, ~10k tokens):    2.84s
      - second (blend warming up):        2.28s
      - third (warm blend):                      0.80s   **3.5x speedup**.


csrc/sycl/cal_cdf_sycl.cpp
```diff
@@ -0,0 +1,219 @@
...
+              }
+
+              if (in_range) {
+                // Histogram pass: counts at hist[v+1][tx].  hist[0][tx] stays 0.
+                for (int i = 0; i < ntokens; ++i) {
+                  const uint8_t v =
+                      input[(static_cast<int64_t>(layer_id) * ntokens + i) *
+                                nchannels +
+                            channel_id];
+                  hist[v + 1][tx] += 1;
```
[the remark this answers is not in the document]
ftian1: it's equal to cuda implementation and will never be triggered from CacheGen

csrc/sycl/ac_enc_sycl.cpp
```diff
@@ -0,0 +1,371 @@
...
+}
+
+template <typename SLMRef>
+inline void add_bits_to_output(uint32_t bit, int num, uint32_t& output_reg,
+                               int& output_reg_len, SLMRef output_shared,
+                               int& output_shared_offset) {
+  do {
+    const int remaining = sycl::min<int>(num, 32 - output_reg_len);
+    output_reg <<= remaining;
+    output_reg |= (bit << remaining) - bit;
```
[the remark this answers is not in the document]
ftian1: not a real issue, same with cuda's implementation and will not be triggered

csrc/sycl/ac_enc_sycl.cpp
```diff
@@ -0,0 +1,371 @@
...
+  // guaranteed to be 4-byte aligned).
+  output_shared[output_shared_offset] =
+      static_cast<uint8_t>(output_reg >> 24);
+  output_shared[output_shared_offset + 1] =
+      static_cast<uint8_t>((output_reg >> 16) & 0xFFu);
+  output_shared[output_shared_offset + 2] =
+      static_cast<uint8_t>((output_reg >> 8) & 0xFFu);
+  output_shared[output_shared_offset + 3] =
+      static_cast<uint8_t>(output_reg & 0xFFu);
+  output_shared_offset += 4;
```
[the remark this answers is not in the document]
ftian1: not a issue

csrc/sycl/ac_dec_sycl.cpp
```diff
@@ -0,0 +1,514 @@
...
+                    break;
+                  }
+
+                  if (bit_idx == 9) {
+                    bit_idx = 1;
+                    byte_buffer_offset += 1;
+                    byte_buffer =
+                        bs_shared[row_base + byte_buffer_offset];
+                  }
+                }
```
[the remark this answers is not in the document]
ftian1: not an issue

ftian1:
@hlin99 @sammshen @DongDongJu could you pls approve the CI test flow and take a look at this PR if having time:)

lmcache/v1/compute/attention/flash_attn.py
```diff
@@ -2,10 +2,28 @@
 # Standard
 from typing import TYPE_CHECKING
 
+try:
+    # Third Party
+    from vllm.attention import Attention
+except ImportError:
+    # vLLM >= 0.17 moved the Attention class.
+    from vllm.model_executor.layers.attention.attention import Attention
```
hlin99, line 10: is this a compatibility issue relating to vllm, not xpu sycl, right?

lmcache/v1/compute/attention/flash_attn.py
```diff
@@ -2,10 +2,28 @@
...
+    # Third Party
+    from vllm.vllm_flash_attn import flash_attn_varlen_func, get_scheduler_metadata
+except ImportError:
+    # CUDA flash-attention extensions are unavailable on Intel XPU; use the
+    # vLLM-shipped XPU flash-attention ops instead.
+    # Third Party
+    from vllm._xpu_ops import xpu_ops
+
+    flash_attn_varlen_func = xpu_ops.flash_attn_varlen_func
+    get_scheduler_metadata = xpu_ops.get_scheduler_metadata
```
hlin99: is there a way to handle this device specifics under lmcache/__init__.py or lmcache/platform/xpu, etc? keeping this file unchanged as common logic would be better.

hlin99:
> @hlin99 @sammshen @DongDongJu could you pls approve the CI test flow and take a look at this PR if having time:)

hi @ftian1  thanks for this PR. some nits inline. and in addition, there're several cuda or cu keywords in the sycl kernel files. can you check if they're intentional or should be changed to xpu/sycl, etc?

lmcache/v1/compute/attention/flash_attn.py
ftian1: yes, vllm compatibility issue. this try catch is trying to deal with two versions of vLLM

lmcache/v1/compute/attention/flash_attn.py
ftian1: I thought the same way before. but why I didn't adopt it is because this is vllm flashatt related import issue rather than lmcache self.

under lmcache/v1/compute/attention/, there are different attention backends. XPU path only uses flashatt. so this file is the only one shared by CUDA and XPU for attention.

that's why I think it's the best place to add such logic. just maybe the logic added looks not pretty...

if we want to enhance this part, I would suggest below solution

1. vllm different version symbole change, keep ths line 5-10 as is
2. move the device specific logic to utils.py under v1/compute/attention. 
```
      ## utils.py
      def get_vllm_device_func(device):
          if device == 'cuda':
               from vllm.vllm_flash_attn import flash_attn_varlen_func, get_scheduler_metadata
          elif device == 'xpu':
               from vllm._xpu_ops import xpu_ops
               flash_attn_varlen_func = xpu_ops.flash_attn_varlen_func
               get_scheduler_metadata = xpu_ops.get_scheduler_metadata
          return flash_attn_varlen_func, get_scheduler_metadata

      ## flash_attn.py
      class LMCFlashAttnBackend():
           def __init__():
                 self.flash_attn_varlen_func, self.get_scheduler_metadata = get_vllm_device_func(device)

```

hlin99:
hi @ftian1 overall LGTM.

I am just aware that tests/v1/test_python_ops_fallback.py is still only testing xpu python fallback(test exists before sycl kernel is ready). Can you turn it on by func “_build_backend_params” to test real sycl implementation and compare with python fallback? just like what we did on cuda. thanks

csrc/sycl/ac_enc_sycl.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+// Encoder caps lp at 48 to keep cdf_shared SLM at 48*BLOCK_SIZE*2B = 12KB
+// for BLOCK_SIZE=128.  Combined with output_shared (32KB) this fits one WG
+// per Xe sub-slice; bumping to 64 pushes total SLM past the occupancy
+// sweet spot and measurably slows the kernel.
+constexpr int MAX_LP = 48;
+constexpr int OUTPUT_BUFFER_LENGTH_PER_THREAD = 256;
+constexpr int MAX_TOKENS_PER_THREAD = 256;
+constexpr int PRECISION = 16;
+constexpr int INTEL_SUB_GROUP_SIZE = 16;
```
DongDongJu, line 55: I have a question for better understanding.
So basically, its hard-coded work-groups.
Do we have any possibility that it will change based on generations? or just fixed one for this hardware architecture?

csrc/sycl/ac_dec_sycl.cpp
```diff
@@ -0,0 +1,494 @@
...
+                value -= 0x40000000u;
+                value = (value << 1) | ((byte_buffer >> (8 - bit_idx)) & 1u);
+                bit_idx += 1;
+              } else {
+                break;
+              }
+              if (bit_idx == 9) {
+                bit_idx = 1;
+                byte_buffer_offset += 1;
+                byte_buffer = bs_shared[row_base + byte_buffer_offset];
```
DongDongJu, line 354: Is this access safe for general?

csrc/sycl/ac_enc_sycl.cpp
```diff
@@ -0,0 +1,357 @@
...
+      !output_buffer.device().is_xpu() || !output_lengths.device().is_xpu()) {
+    throw std::runtime_error("encode_fast_new_xpu: all tensors must be on XPU");
+  }
+
+  const auto cdf_shape = cdf.sizes();
+  const auto input_shape = input_sym.sizes();
+  const auto output_shape = output_buffer.sizes();
+
+  const int nlayers = static_cast<int>(cdf_shape[0]);
+  const int nchannels = static_cast<int>(cdf_shape[1]);
```
DongDongJu, line 303: one nit. it seems directly access with too many hard coded value.
How about adding TORCH_CHECK for each?

tests/test_serde.py
```diff
@@ -4,12 +4,17 @@
...
+from lmcache import torch_dev, torch_device_type
 from lmcache.storage_backend.serde.cachegen_basics import CacheGenEncoderOutput
 from lmcache.storage_backend.serde.cachegen_decoder import CacheGenDeserializer
 from lmcache.storage_backend.serde.cachegen_encoder import CacheGenSerializer
 from lmcache.v1.config import LMCacheEngineConfig
 from lmcache.v1.metadata import LMCacheMetadata
 
+# CacheGen has hand-written CUDA and SYCL kernels.  Skip when neither GPU
+# backend is available (e.g. CPU-only CI).
+_GPU_AVAILABLE = torch_dev.is_available()
```
DongDongJu: Please make sure that HPU case can avoid this.

DongDongJu (changes requested):
Hello @ftian1, Thanks for the contribution!
native SYCL CacheGen/RoPE support is valuable, and the reported XPU speedups look promising.
Please check my few comments and nit.

csrc/sycl/ac_enc_sycl.cpp
ftian1: good findings. yes, 16 is a common value which is suitable for all Intel GPUs (PVC, BMG, and future Xe3). we can query device's supported subgroup size by manual, but it will cause the kernel generation multi-times (for example, BMG supports 16, 32) if we introduce a template to deal with all sizes, the kernel has to be compile twice and make .so bigger. 

And 16 is the best value for ac_enc and ac_dec per real measurement & tuning. that's why I put it here. In the future, for example, on the Intel next generation GPU, we can revise this part later

csrc/sycl/ac_dec_sycl.cpp
ftian1: This is same logic as how CUDA enc/dec kernel works. CUDA supports UINT16 & UINT8 at the same time. but only UINT8 case is triggered in the whole path. so I didn't implement it as a template. but the algo is same

tests/test_serde.py
ftian1: you are right, this part has problem. let me fix it

csrc/sycl/ac_enc_sycl.cpp
ftian1: good suggestions. I will add more strict checks on that

tests/test_serde.py
ftian1: fixed, pls review

csrc/sycl/ac_enc_sycl.cpp
ftian1: done

ftian1:
> hi @ftian1 overall LGTM.
> 
> I am just aware that tests/v1/test_python_ops_fallback.py is still only testing xpu python fallback(test exists before sycl kernel is ready). Can you turn it on by func “_build_backend_params” to test real sycl implementation and compare with python fallback? just like what we did on cuda. thanks

comments solved. pls review. thanks

ftian1:
@hlin99  I further improved current code to make tests/v1/test_python_ops_fallback.py pass on XPU, pls check the latest PR

DongDongJu (commented):
I can not review 14k changes. Sorry.

tests/v1/test_python_ops_fallback.py
```diff
@@ -80,6 +81,16 @@ def _build_backend_params() -> list:
             )
 
     if hasattr(torch, "xpu") and torch.xpu.is_available():
+        try:
+            # First Party
+            import lmcache.xpu_ops as xpu_sycl_ops
```
hlin99: import lmcache.c_ops as xpu_sycl_ops
c_ops is the single entrypoint for code across backends to rely on

tests/v1/test_python_ops_fallback.py
```diff
@@ -80,6 +81,16 @@ def _build_backend_params() -> list:
...
+
+            params.append(
+                pytest.param(("xpu_sycl_ops", xpu_sycl_ops, "xpu"), id="xpu_sycl_ops")
+            )
+        except ImportError:
+            pass
+
         params.append(pytest.param(("xpu_py_ops", _py_ops, "xpu"), id="xpu_py_ops"))
```
hlin99: as xpu has real c_ops, this fake py ops can be removed.

tests/v1/test_python_ops_fallback.py
```diff
@@ -1767,7 +1791,11 @@ def test_1_scenario(
         by test_2_compare.
         """
         backend_id, ops, device = backend
-        result = fn(ops, device)
+        # Skip scenarios whose required op is not exported by this backend
```
hlin99: this logic should not be here.  the logic is to test each c_ops has full functions with same signature, so when c_ops is using in code, people don't need to aware of the differences of each device.

ftian1:
> I can not review 14k changes. Sorry.

my fault. have solved it by real merge. I also revise the python fallback related code per comments. pls review  @DongDongJu @hlin99

hlin99 (approved):
thanks @ftian1 for the great work. 
Now, this PR LGTM, and the relevant UTs have been passed on my local xpu machine.

hi @DongDongJu would you like to have another looks? thx.

DongDongJu (approved):
Results are looks promising.
LGTM

PR #3363, opened 2026-05-22
[MP] Update the L2 adapter interface to force measuring the real transferred bytes

ApostaC:

1. Introduces `L2StoreResult(int)` — an immutable result type
  a. `>= 0` means success with that many bytes transferred, 
  b. `-1` means failure.
2. Removes the `pop_completed_store_task_bytes()` interface.
3. `pop_completed_store_tasks()` now returns `dict[L2TaskId, L2StoreResult]` instead of `dict[L2TaskId, bool]`. All the related code is updated.

2 and 3 effectively force every L2 adapter to report actual bytes transferred, which avoids the potential for wrong metrics being reported.

**If applicable**:


Oasis-Git (approved):

chunxiaozheng (approved):
@ApostaC LGTM! There is a small question that not about this PR, should we add the similar interfaces for `load` later?

PR #3364, opened 2026-05-22
[CB] Blend v3

deng451e:

**If applicable**:


lmcache/v1/multiprocess/server.py
```diff
@@ -194,6 +194,19 @@ def _build_modules(
 
         modules.append(BlendModule(ctx))
 
+    if mp_config.engine_type == "blend_v3":
```
ApostaC: We can overwrite the `blend` engine type

ApostaC: In this case, we may remove the old CI/CD for blend and blend_v2.

ApostaC (commented):
Otherwise LGTM as we discussed offline

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM

Issue #3372, opened 2026-05-23
[Onboarding 2026] Good first issues — start contributing to LMCache in one PR


maobaolong:
👋 Welcome to LMCache! This umbrella issue is the **single entry point** for first-time
contributors. If you've never sent a PR to LMCache before, start here.

> If you're looking for the older onboarding thread, see #627. This 2026 version
> focuses on smaller, well-scoped tasks with a friendly step-by-step PR workflow.

---

## 0. Before you start

- Read [`CONTRIBUTING.md`](../blob/dev/CONTRIBUTING.md). The most important parts:
  - Base branch is `dev`. Open PRs against `dev`, not `main`.
  - Every Python file starts with `# SPDX-License-Identifier: Apache-2.0`.
  - Imports are grouped as `# Standard / # Third Party / # First Party / # Local`.
  - Line length is **88**, formatter is `ruff format`, lint is `ruff check`.
  - Never access `_private` members of other classes (SLF rule).
  - Public functions/methods need **type hints** + **docstring**.
- Set up the environment:
  ```bash
  uv venv --python 3.12 && source .venv/bin/activate
  uv pip install -e . --no-build-isolation
  uv pip install -r requirements/test.txt
  pre-commit install
  ```

## 1. How to claim a task

1. Pick **one** sub-issue below that has no assignee and no `🔒 claimed` comment.
2. Comment exactly: `/claim` (or `I'd like to work on this 🙋`).
   A maintainer will assign it to you within ~2 business days.
3. If you go silent for 7 days we'll un-assign so others can pick it up — totally fine,
   just leave a comment if you need more time.

## 2. Branch & commit

```bash
git checkout dev && git pull
git checkout -b gfi/<issue-number>-short-slug
# ... do the work ...
git commit -s -m "short imperative summary"   # -s is required (DCO)
```

- **Sign-off is required** (`-s`). PRs without DCO sign-off cannot be merged.
- Keep the PR small: one issue = one PR.

## 3. Local checks (must pass before pushing)

```bash
pre-commit run --all-files
pytest -xvs tests/<the_file_you_touched>.py    # if applicable
```

## 4. Open the PR

**PR title format** (please follow):

```
[good-first-issue] <area>: <imperative summary>  (#<sub-issue-number>)
```

Examples:
- `[good-first-issue] docs: fix broken link in quickstart (#1801)`
- `[good-first-issue] tests: add unit test for FooBar.flush (#1802)`
[...]

PR #3375, opened 2026-05-23
[good-first-issue] local-disk: persist metadata across restarts (#1175)

### LINKED ISSUE #1175: KV-Cache Persistence Issue with SSD Offloading
**Description:**  
follow the SSD offload guide in :

> https://docs.lmcache.ai/kv_cache/local_storage.html

experimentation with vLLM  and LMCache, we observed that the KV-Cache offloading to SSD works partially but lacks persistence across server restarts.  

**Steps to Reproduce:**  
1. Configure vLLM server with SSD offloading for KV-Cache  
2. Run inference scripts with a specific prompt 
3. Observe 2.4GB `.pt` file generated in the SSD cache directory  
4. Restart the vLLM server process  
5. Query the same prompt again  
6. the kv-cache regenerated in SSD

**Expected Behavior:**  
The existing KV-Cache files should be reused, avoiding redundant computation and disk I/O.  

**Actual Behavior:**  
New identical KV-Cache files are regenerated instead of reusing the existing ones

This optimization would significantly improve latency for recurring queries in production scenarios. I would appreciate guidance on recommended methods to achieve proper KV-Cache persistence across server restarts in this workflow.

koopatroopa787:
Fixes #1175
Refs #3372
- Persist LocalDiskBackend metadata index on shutdown and reload on startup so cached entries survive restarts.
- Restore usage/current_cache_size and reset pin counts on load.

- pre-commit run --all-files failed: mypy hook requires python3.12 which is unavailable in this environment.
- pytest -xvs tests/v1/storage_backend/test_local_disk_backend.py failed on Windows: os.statvfs is not available.

**If applicable**:


DongDongJu:
Hello @koopatroopa787, Thanks for the work! Could you check this once? https://github.com/LMCache/LMCache/pull/3189.

koopatroopa787:
Thanks for pointing me to #2734 @DongDongJu — I had a look at the right PR this time. The JSON manifest approach there is clearly the right direction, and it aligns with the feedback from the automated reviewers here too.

I've updated this PR to address the open review comments:

**Copilot (security):** Replaced `pickle` with `json` throughout — the index file is now `.lmcache_local_disk_metadata.json` written with `json.dump` and read with `json.load`. No more arbitrary-code-execution risk from a tampered index file.

**Gemini (performance — load path):** Replaced the per-entry `os.path.getsize()` call in `_load_metadata_index` with `os.path.exists()` and reuse the size already stored in the deserialized `DiskCacheMetadata`. Since KV cache chunks are immutable once written, the stored size is always accurate.

**Gemini (performance — persist path):** Same change in `_persist_metadata_index` — `os.path.exists()` instead of `os.path.getsize()`, and we no longer mutate `meta.size` on the live object at shutdown.

**Copilot (over-quota on load):** Already handled in the prior version — the load path evicts oldest entries until `current_cache_size <= max_cache_size`.

**Copilot (`cached_positions` size):** Already handled — `_serialize_cached_positions` calls `.tolist()` so tensors are stored as plain int arrays, not pickled tensor bytes.

I also switched to `tempfile.NamedTemporaryFile` for the atomic temp file write so the cleanup is always handled in the `finally` block regardless of what fails.

The overall structure here differs from #2734 (list of entry dicts vs. key-string-keyed dict, load-only-on-close vs. per-put), but functionally covers the same ground. Happy to consolidate further if that's the direction you want to take.

DongDongJu:
> Thanks for pointing me to #2734 @DongDongJu — I had a look at the right PR this time. The JSON manifest approach there is clearly the right direction, and it aligns with the feedback from the automated reviewers here too.
> 
> I've updated this PR to address the open review comments:
> 
> **Copilot (security):** Replaced `pickle` with `json` throughout — the index file is now `.lmcache_local_disk_metadata.json` written with `json.dump` and read with `json.load`. No more arbitrary-code-execution risk from a tampered index file.
> 
> **Gemini (performance — load path):** Replaced the per-entry `os.path.getsize()` call in `_load_metadata_index` with `os.path.exists()` and reuse the size already stored in the deserialized `DiskCacheMetadata`. Since KV cache chunks are immutable once written, the stored size is always accurate.
> 
> **Gemini (performance — persist path):** Same change in `_persist_metadata_index` — `os.path.exists()` instead of `os.path.getsize()`, and we no longer mutate `meta.size` on the live object at shutdown.
> 
> **Copilot (over-quota on load):** Already handled in the prior version — the load path evicts oldest entries until `current_cache_size <= max_cache_size`.
> 
> **Copilot (`cached_positions` size):** Already handled — `_serialize_cached_positions` calls `.tolist()` so tensors are stored as plain int arrays, not pickled tensor bytes.
> 
> I also switched to `tempfile.NamedTemporaryFile` for the atomic temp file write so the cleanup is always handled in the `finally` block regardless of what fails.
> 
> The overall structure here differs from #2734 (list of entry dicts vs. key-string-keyed dict, load-only-on-close vs. per-put), but functionally covers the same ground. Happy to consolidate further if that's the direction you want to take.

Hello @koopatroopa787, Thanks for your contribution!
I had an offline talk with @ApostaC and we decided not to save manifest like metadata into fs directly.
I will update the #2734 for the next direction today.
Agan, Thanks for your time and contribution! Please ping me in next other PR, I will take a look promptly.

PR #3382, opened 2026-05-24
[MP] Fix GPU block exhaustion deadlock at high concurrency with chunked KV loading

andyluo7:
## Problem

When using `LMCacheConnector` (v1) with high concurrency (e.g., 32+ concurrent
requests with 100K+ token contexts), `get_num_new_matched_tokens()` reports
ALL externally-cached tokens at once. The vLLM scheduler then allocates GPU
blocks for all concurrent requests simultaneously, which can exceed the total
GPU block pool capacity and cause a scheduling deadlock.

Example: 32 concurrent requests × 115K tokens × block_size=1 = 3.7M blocks
needed, but only 1.8M blocks available on 4x MI355X GPUs.

## Root Cause

The scheduler allocates GPU blocks for `num_external_computed_tokens` before
the async KV transfer, putting the request in `WAITING_FOR_REMOTE_KVS` state.
With many concurrent requests all needing large block allocations, the block
pool is exhausted and no new requests can be scheduled.

### 1. Chunked KV loading cap (`vllm_v1_adapter.py`)

Cap the number of external tokens reported per scheduling step via
`lmcache.max_tokens_per_load` (configurable, default 131072 = 128K tokens).
When a request has more cached tokens than this limit, only the capped amount
is reported to the scheduler. The remaining tokens beyond the cap are computed
locally via chunked prefill rather than loaded from external cache.

Key properties:
- The cap is aligned to LMCache chunk boundaries (`chunk_size`)
- `LoadSpec.lmcache_cached_tokens` is adjusted to match the capped amount,
  keeping the `update_state_after_alloc` assertion consistent
- Set to `0` to disable chunking (original behavior)

### 2. Defensive alignment in `Session.get_hashes()` (`session.py`)

Replace a hard assertion (`assert end % chunk_size == 0`) with a graceful
round-down and warning log. Non-chunk-aligned end values can legitimately occur
with `block_size=1` (MLA models like DeepSeek-V3, Kimi-K2.5) during chunked
loading.

## Configuration

```json
{
  "kv_connector_extra_config": {
    "lmcache.max_tokens_per_load": 131072
  }
}
```

Set to `0` to disable chunking (original behavior, all matched tokens reported
at once).

## Testing

Validated on MI355X (4x 288GB VRAM) with Kimi-K2.5-MXFP4 (block_size=1):
- ✅ c=32 (100K+ context): 120 requests completed (was 28 → deadlock)
- ✅ c=48: No deadlock under extreme pressure (KV 98.5%)
- ✅ c=8: No overhead when block pressure is low
- ✅ TP=8: Adapts to different block pool sizes
- ✅ Short context (8K): Chunking correctly bypassed
- ✅ Correctness: Greedy decode output matches across configurations

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -572,6 +572,23 @@ class LMCacheConnectorV1Impl:
 
         self.force_skip_save = bool(os.environ.get("LMCACHE_FORCE_SKIP_SAVE", False))
         self._requests_priority: dict[str, int] = {}
+
+        # Chunked KV loading: cap the number of external tokens
+        # reported per scheduling step to prevent GPU block pool
+        # exhaustion at high concurrency with long contexts.
+        # Configurable via kv_connector_extra_config:
+        #   "lmcache.max_tokens_per_load": <int>
+        # Default: max_model_len // 2, capped at 128K tokens.
+        # Set 0 to disable (original behavior, all matched tokens
+        # reported at once).
+        default_cap = min(
+            vllm_config.model_config.max_model_len // 2, 131072
+        )
+        self._max_tokens_per_load: int = int(
+            vllm_config.kv_transfer_config.get_from_extra_config(
+                "lmcache.max_tokens_per_load", default_cap
+            )
+        )
         self._invalid_block_ids: set[int] = set()
 
     def _check_legacy_register_kv_caches(self) -> None:
@@ -1437,9 +1454,44 @@ class LMCacheConnectorV1Impl:
                 max(need_to_allocate, 0),
             )
 
+        # Chunked KV loading: cap the number of tokens reported
+        # to the scheduler to avoid exhausting the GPU block pool
+        # when many concurrent requests each need large allocations.
+        # Remaining tokens beyond the cap will be computed locally
+        # via chunked prefill rather than loaded from external cache.
+        capped_lmcache_tokens = num_external_hit_tokens
+        if (
+            self._max_tokens_per_load > 0
+            and need_to_allocate > self._max_tokens_per_load
+        ):
+            # Align cap to LMCache chunk boundaries so that the
+            # retrieve path receives chunk-aligned token ranges.
+            cap = (
+                self._max_tokens_per_load
+                // self._lmcache_chunk_size
+                * self._lmcache_chunk_size
+            )
+            need_to_allocate = cap
+            # Align capped_lmcache_tokens to chunk boundary so that
+            # session.get_hashes() receives a chunk-aligned end value.
+            capped_lmcache_tokens = (
+                (num_computed_tokens + cap)
+                // self._lmcache_chunk_size
+                * self._lmcache_chunk_size
+            )
+            logger.debug(
+                "Reqid: %s, Chunked KV loading: capped from "
+                "%d to %d external tokens "
+                "(max_tokens_per_load=%d)",
+                req_id,
+                num_external_hit_tokens - num_computed_tokens,
+                need_to_allocate,
+                self._max_tokens_per_load,
+            )
+
         self.load_specs[req_id] = LoadSpec(
             vllm_cached_tokens=num_computed_tokens,
-            lmcache_cached_tokens=num_external_hit_tokens,
+            lmcache_cached_tokens=capped_lmcache_tokens,
             can_load=False,
         )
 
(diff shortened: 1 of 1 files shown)


lmcache/v1/multiprocess/session.py
```diff
@@ -83,9 +83,16 @@ def get_hashes(self, start: int, end: int | None = None) -> list:
...
+            if end % chunk_size != 0:
+                original_end = end
+                end = (end // chunk_size) * chunk_size
+                logger.debug(
+                    "end (%d) is not a multiple of chunk_size (%d). "
+                    "Rounding down to %d.",
+                    original_end,
+                    chunk_size,
+                    end,
+                )
```
ApostaC: As written in the docstring of this function, the `end` needs to be aligned to lmcache chunk size. Let's keep the old behavior and check the caller to see what the real issue is.

lmcache/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1433,9 +1347,38 @@ def get_num_new_matched_tokens(
                 max(need_to_allocate, 0),
             )
 
+        # Chunked KV loading: cap the number of tokens reported
+        # to the scheduler to avoid exhausting the GPU block pool
+        # when many concurrent requests each need large allocations.
+        # Remaining tokens beyond the cap will be computed locally
+        # via chunked prefill rather than loaded from external cache.
```
ApostaC, line 1458: I think the vLLM scheduler should be able to handle the case that connector reports more tokens than the current available capacity. The result will be available from the `udpate_state_after_alloc` function. 

Ideally, LMCache doesn't need to have a cap for this. If we really need the cap, instead of having a hard-coded default, I would recommend detecting the `max_model_len` from the vLLM config.

ApostaC (changes requested):
Thanks for the contribution. Please see my comments.

lmcache/integration/vllm/vllm_v1_adapter.py
andyluo7, line 1: Thanks for the review @ApostaC!

**Re: session.py — keeping the assertion:**

You're right, I'll revert the round-down change. Looking at the callers:

- `vllm_multi_process_adapter.py:571` already aligns end: `aligned_end = (len(token_ids) // self.chunk_size) * self.chunk_size`
- The non-aligned end comes from our chunked connector patch where the capped token count may not be chunk-aligned. The fix should be in the caller (align the capped value to chunk boundaries before it reaches `get_hashes`), not in `session.py`.

I'll revert session.py and fix the alignment in the chunked connector patch's capped value calculation instead.

**Re: vllm_v1_adapter.py — cap approach:**

The cap is needed because vLLM's v1 scheduler allocates GPU blocks for `need_to_allocate` tokens **before** the KV transfer happens. At high concurrency (c≥32) with long contexts (100K+ tokens), if all requests report their full external hit count, the scheduler tries to allocate blocks for all of them simultaneously, exhausting the GPU block pool. No request can make progress → deadlock.

The scheduler's `update_state_after_alloc` handles the case where fewer blocks are available than requested, but only for a single request — it can't resolve the circular dependency when *all* requests are waiting for blocks that other requests hold.

That said, I agree the hardcoded 128K default is not ideal. I'll change it to derive from `vllm_config`:
```python
# Default: max_model_len / 2, capped at 128K
default_cap = min(vllm_config.model_config.max_model_len // 2, 131072)
```

This scales with the model's context window and avoids the magic number. Users can still override via `kv_connector_extra_config["lmcache.max_tokens_per_load"]`.

Will push the updated PR shortly.

ApostaC: Thanks for the update. Can we make "no cap" the default behavior (to keep the same default behavior as existing versions)?

We can set the cap only when we see `lmcache.max_tokens_per_load` in the config

andyluo7:
Done — default is now 0 (no cap, preserving existing behavior). Chunked loading only activates when `lmcache.max_tokens_per_load` is explicitly set in the config. See 8f11bab.

ApostaC (approved):
lgtm

andyluo7:
Hi @ApostaC — thanks for the approval! 🙏 

All CI is green (the only red is `buildkite/k3-unit-tests` which appears to be an unrelated infra flake — `buildkite/unit-tests-amd` passed). Could you (or another maintainer) help merge this when you have a moment? Happy to rebase if needed.

cc @YaoJiayi @sammshen

andyluo7:
Friendly bump 🙏 — this is fully approved by @ApostaC and all relevant CI is green. The 3 red `buildkite/k3-*` checks look like the known infra flake (the AMD unit tests and all comprehensive/integration/multiprocess/operator suites passed).

@KuntaiDu @maobaolong — could one of you help merge when you have a moment? Happy to rebase if anything has drifted. Thanks!

sammshen (approved):
LGTM

Issue #3381, opened 2026-05-24
enable enable_async_load then suffer unpin -1 issue


gaowayne:
**Label**
Please label your issue with "bug" and any other relevant labels so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/1882)

**Describe the bug**
I enable_async_load, after kvcache ingestion, I try to run cache hit read from vllm bench. then I saw couples error
1. double unpin occur
2. async lookup timeout
3. batched_async_contains timeout

**To Reproduce**
here is my confige
```
chunk_size: 256

local_cpu: True
max_local_cpu_size: 256.0

remote_url: "s3://lmcache-kv.storage.local:9000"
remote_serde: "naive"

extra_config:
  aws_access_key_id: "minioadmin"
  aws_secret_access_key: "minioadmin123"
  s3_region: "us-east-1"
  disable_tls: True
  s3_enable_s3express: False
  s3_prefer_http2: False
  save_chunk_meta: False


LMCACHE_ENABLE_ASYNC_LOADING=True
enable_async_loading: true
```
here is error log
```
ck to 0 as a hack but please find the bug. (memory_management.py:619:lmcache.v1.memory_management)
(EngineCore pid=98264) [2026-05-24 10:21:29,867] LMCache WARNING: Pin count of MemoryObj 199179108352is negative: -1.Double unpin occurred somewhere.Setting pin count back to 0 as a hack but please find the bug. (memory_management.py:619:lmcache.v1.memory_management)
(EngineCore pid=98264) [2026-05-24 10:21:29,867] LMCache WARNING: Pin count of MemoryObj 199246217216is negative: -1.Double unpin occurred somewhere.Setting pin count back to 0 as a hack but please find the bug. (memory_management.py:619:lmcache.v1.memory_management)
(EngineCore pid=98264) [2026-05-24 10:21:29,867] LMCache WARNING: Pin count of MemoryObj 199313326080is negative: -1.Double unpin occurred somewhere.Setting pin count back to 0 as a hack but please find the bug. (memory_management.py:619:lmcache.v1.memory_management)
(EngineCore pid=98264) [2026-05-24 10:21:29,867] LMCache WARNING: Pin count of MemoryObj 199380434944is negative: -1.Double unpin occurred somewhere.Setting pin count back to 0 as a hack but please find the bug. (memory_management.py:619:lmcache.v1.memory_management)
[...]

PR #3385, opened 2026-05-25
[Fix]: Skip unpin for non-pinned objects in cleanup_memory_objs

zhengfeihe:
Fix unpin issue https://github.com/LMCache/LMCache/issues/3381

Cause: 

 cleanup_memory_objs unpins every chunk from async_lookup_and_prefetch, but not every backend pins them in the first place. LocalCPU pins in batched_async_contains when pin=True; Remote's pin is a no-op, so S3 chunks come back unpinned and the unpin drops pin_count below zero.
    
 Add a check before doing unpin.

**If applicable**:


zhengfeihe:
addressed Gemini comments

sammshen (approved):
LGTM

deng451e (approved):
LGTM

PR #3389, opened 2026-05-25
Fix SGLangLayerwiseGPUConnector generator and layer indexing

JinuJeong:

Three bugs were occurring in sequence during the SGLang layerwise KV cache load process.

### 1. Layer completion check off-by-one (sglang_adapter.py)

The condition `== num_hidden_layers` was evaluated immediately after incrementing `layer_load_layer[i]`, causing it to match before the final layer was actually processed. As a result, `layerwise_retrievers[i]` and associated resources were released prematurely. When load_kv_layerwise was called for the final layer, the retriever had already been deleted, so the KV cache load for that layer was silently skipped. Fixed by changing `==` to `>` so cleanup only happens after the final layer has been fully processed.

### 2. Missing final yield in SGLangLayerwiseGPUConnector.batched_from_gpu (gpu_connectors.py)

The generator exited without a yield after processing the last layer, causing the GPU buffer's ref_count to drop immediately before the caller could advance the generator to completion. This resulted in abnormal memory release while the buffer was still in use. Fixed by adding a yield after the last layer is processed, before the buffer cleanup.

### 3. Double unpin (sglang_adapter.py)

The bugs above caused `lookup_unpin` in `cache_engine` to be missed. When this was corrected, `lookup_unpin` was being called at the layer completion site in `sglang_adapter.py`, while `load_kv_layerwise` was already calling `unpin` — resulting in a double unpin. Removing `lookup_unpin` avoids the double unpin, but completely removing it causes a memory leak because the `lookup_id` is never popped from `LMCacheEngine.lookup_pins`, so the dictionary grows indefinitely with every layerwise retrieve request. Fixed by popping the `lookup_id` directly from `lookup_pins` without calling `batched_unpin`, which cleans up the dictionary entry while avoiding the double unpin.

### Changes

  - `lmcache/v1/gpu_connector/gpu_connectors.py`: Add missing final yield in batched_from_gpu generator before buffer cleanup
  - `lmcache/integration/sglang/sglang_adapter.py`: Fix layer completion check; replace `lookup_unpin` call with `lookup_pins.pop` to prevent memory leak without triggering double unpin.
  - `tests/v1/test_gpu_connector.py`: Add `test_layerwise_sglang_connector_with_gpu` for `SGLangLayerwiseGPUConnector`

#### Test plan
  - [ ] `pytest tests/v1/test_gpu_connector.py::test_layerwise_sglang_connector_with_gpu` passes on a CUDA-enabled machine
  - [ ] Existing test_sglang_connector_* tests unaffected

**If applicable**:

### DIFF
--- a/lmcache/integration/sglang/sglang_adapter.py
+++ b/lmcache/integration/sglang/sglang_adapter.py
@@ -243,13 +243,13 @@ class LMCacheLayerwiseConnector(LMCacheConnector):
             if self.layer_load_layer[i] == layer_id + 1:
                 next(self.layerwise_retrievers[i])
                 self.layer_load_layer[i] += 1
-                if self.layer_load_layer[i] == self.sgl_config.num_hidden_layers:
+                if self.layer_load_layer[i] > self.sgl_config.num_hidden_layers:
                     indices_to_remove.append(i)
 
         for i in sorted(indices_to_remove, reverse=True):
             del self.layerwise_retrievers[i]
             del self.layer_load_layer[i]
-            self.lmcache_engine.lookup_unpin(self.lookup_id_list[i])
+            self.lmcache_engine.lookup_pins.pop(self.lookup_id_list[i], None)
             del self.lookup_id_list[i]
 
         return
--- a/lmcache/v1/gpu_connector/gpu_connectors.py
+++ b/lmcache/v1/gpu_connector/gpu_connectors.py
@@ -1796,6 +1796,8 @@ class SGLangLayerwiseGPUConnector(GPUConnectorInterface):
                     token_major=True,
                 )
 
+        yield
+
         # free the buffer memory
         if self.use_gpu:
             tmp_gpu_buffer_obj.ref_count_down()
--- a/tests/v1/test_gpu_connector.py
+++ b/tests/v1/test_gpu_connector.py
@@ -12,6 +12,7 @@ import torch
 # First Party
 from lmcache.v1.gpu_connector.gpu_connectors import (
     SGLangGPUConnector,
+    SGLangLayerwiseGPUConnector,
     VLLMBufferLayerwiseGPUConnector,
     VLLMPagedMemGPUConnectorV2,
     VLLMPagedMemGPUConnectorV3,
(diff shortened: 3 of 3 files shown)


ApostaC:
@DongDongJu @Shaoting-Feng @Oasis-Git Can you take a look at this PR? Thanks!

Shaoting-Feng (commented):
The final layer is not silently skipped. Layer 0 is loaded before any `load_kv_layerwise` call, and `layer_load_layer` starts at 1.

Shaoting-Feng:
Thanks for your PR! The final layer is not silently skipped. Layer 0 is loaded before any load_kv_layerwise call, and layer_load_layer starts at 1. Feel free to reopen it if more bugs appear (workload to replay it will be largely appreciated!).

JinuJeong:
Sorry, there were some incorrect points in my previous explanation.

To clarify, the final layer's load is not actually skipped. The issue is that, in SGLang, when the `final - 1` layer calls `load_kv_layerwise`, it issues the load for the final layer. After that, `layer_load_layer` is incremented, and its value becomes equal to `num_hidden_layers`, so the retriever is appended to `indices_to_remove`. As a result, the layer retriever is removed and the corresponding memory objects are unpinned.

Later, when the final layer calls `load_kv_layerwise`, its corresponding layer retriever can no longer be processed because it has already been removed. More specifically, in `cache_engine.py`, inside `retrieve_layer`, execution is suspended at the `yield` at line 1037, but the layer retriever is removed before the generator is resumed, so the code after that `yield` cannot be executed.

I do not think this matches the intended behavior. Also, since the memory objects for the final layer are unpinned immediately after the load is issued, they could be unpinned before the copy has completed if the load uses an asynchronous copy.

Shaoting-Feng:
Thanks @JinuJeong for the clarification. Since v1 in LMCache will be deprecated soon and MP mode is going to be the default solution, I suggest we hold off on the v1 solution for now, as it is already workable using the original layer logic.

JinuJeong:
@Shaoting-Feng Got it. Thanks for the explanation!

PR #3391, opened 2026-05-26
[MP][Core] Refactor MPCacheEngine for better extendability

ApostaC:

Decomposes the monolithic `MPCacheEngine` (~1350 lines) and `BlendEngineV2` (inheritance-based) into composable, pluggable modules assembled by a thin compositor at startup.

The old design made it impossible to mix-and-match capabilities (e.g. blend + GPU transfer without non-GPU transfer) because `BlendEngineV2` inherited from `MPCacheEngine` and overrode specific methods. The new design uses a compositor pattern: `MPCacheEngine` holds a shared `MPCacheEngineContext` and a list of `EngineModule` instances, each contributing handler specs that are registered with the ZMQ message queue server.

**Architecture:**
- `EngineModule` protocol (`get_handlers()`, `report_status()`, `close()`) with `HandlerSpec` and `ThreadPoolType`
- `MPCacheEngineContext` — shared infrastructure (storage manager, token hasher, session manager, event bus, `LayoutDescRegistry`)
- 5 modules: `LookupModule`, `ManagementModule`, `GPUTransferModule`, `NonGPUTransferModule`, `BlendModule`
- GPU and non-GPU transfer are mutually exclusive at runtime, controlled by new `--transfer-mode` config
- Single unified `run_cache_server()` assembles modules based on `transfer_mode` and `engine_type`

**User-facing changes:**
- New CLI arg `--transfer-mode {gpu,non_gpu}` (default: `gpu`) — explicitly selects the transfer module
- `http_server.py` no longer conditionally imports from `blend_server_v2.py`
- `blend_server_v2.py` is now a thin entry point that delegates to the unified server

- Each module owns its own locks (no shared lock on context except `LayoutDescRegistry`'s internal lock)
- `BlendModule` uses a lazy import in `_build_modules()` to avoid loading heavy blend dependencies when blend is disabled
- The `LayoutDescRegistry` is a thread-safe shared registry on the context — modules write on KV cache registration, consumers read via `find()`
- 3 pre-existing test failures exist unrelated to this PR (`PageBufferShapeDesc.block_stride_elems` C++ extension mismatch, `test_http_api_registry` fixture issue)

**If applicable**:


lmcache/v1/multiprocess/blend_server_v2.py
ApostaC, line 1: This file can be completely removed.

lmcache/v1/multiprocess/config.py
```diff
@@ -39,9 +39,13 @@ class MPServerConfig:
 
     engine_type: str = "default"
     """Cache engine backend type
-    ('default' for MPCacheEngine, 'blend' for BlendEngineV2).
+    ('default' for standard prefix caching, 'blend' for cross-request KV reuse).
```
ApostaC:
```suggestion
    ('default' for standard prefix caching, 'blend' when cacheblend is enabled).
```

lmcache/v1/multiprocess/config.py
```diff
@@ -39,9 +39,13 @@ class MPServerConfig:
...
     """
 
+    transfer_mode: str = "gpu"
+    """Transfer mode: 'gpu' for GPU-based IPC transfer (STORE/RETRIEVE),
+    'non_gpu' for pickle-based CPU transfer (PREPARE/COMMIT)."""
```
ApostaC: nit: `non_gpu for non-GPU-based transfer (PREPARE/COMMIT).` Don't mention pickle here.

lmcache/v1/multiprocess/config.py
```diff
@@ -146,10 +150,19 @@ def add_mp_server_args(
         type=str,
         default="default",
         choices=["default", "blend"],
-        help="Cache engine backend type. 'default' uses MPCacheEngine, "
-        "'blend' uses BlendEngineV2 for cross-request KV reuse. "
+        help="Cache engine backend type. 'default' uses standard prefix caching, "
+        "'blend' enables cross-request KV reuse via BlendModule. "
```
ApostaC: nit: update according to my suggestion above.

lmcache/v1/multiprocess/config.py
```diff
@@ -198,6 +211,7 @@ def parse_args_to_mp_server_config(
         max_cpu_workers=max_cpu,
         hash_algorithm=args.hash_algorithm,
         engine_type=args.engine_type,
+        transfer_mode=getattr(args, "transfer_mode", "gpu"),
```
ApostaC: why use `getattr`? `args.transfer_mode` should exist.

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -0,0 +1,126 @@
...
+class LayoutDescRegistry:
+    """Thread-safe registry mapping (model_name, world_size) to MemoryLayoutDesc.
+
+    Modules write to this registry when KV caches are registered.
+    Consumers (e.g. LookupModule) read from it to find layout descriptors
+    for prefetch tasks.
+    """
+
+    def __init__(self) -> None:
+        self._registry: dict[tuple[str, int], MemoryLayoutDesc] = {}
```
ApostaC, line 34: Need a comment clarifying the key type. What's expected in `tuple[str, int]`?

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -0,0 +1,126 @@
...
+        hash_algorithm: str = "blake3",
+    ) -> None:
+        self.chunk_size = chunk_size
+        self.storage_manager = StorageManager(storage_manager_config)
+        self.token_hasher = TokenHasher(
+            chunk_size=chunk_size, hash_algorithm=hash_algorithm
+        )
+        self.session_manager = SessionManager(self.token_hasher)
+        self.event_bus = get_event_bus()
+        self.layout_desc_registry = LayoutDescRegistry()
```
ApostaC: Directly access the members defined in `self.__init__` of a class is not a good practice.
Either we define public properties to access the members (and make the members private), or we make the class become a dataclass.

lmcache/v1/multiprocess/http_server.py
```diff
@@ -61,12 +61,9 @@ async def lifespan(app: FastAPI):
...
     mp_config = _configs["mp"]
-    if mp_config.engine_type == "blend":
-        # First Party
-        from lmcache.v1.multiprocess.blend_server_v2 import run_cache_server
-    else:
-        # First Party
-        from lmcache.v1.multiprocess.server import run_cache_server
+
+    # First Party
+    from lmcache.v1.multiprocess.server import run_cache_server
```
ApostaC: This now can go to the top of the file.

tests/v1/multiprocess/test_free_locks.py
```diff
@@ -88,42 +88,42 @@ def test_mq_free_locks():
...
+    from lmcache.v1.multiprocess.modules.lookup import LookupModule
 
-    engine = MagicMock()
-    engine.token_hasher = MagicMock()
-    engine.token_hasher.chunk_size = 256
-    engine.token_hasher.compute_chunk_hashes.return_value = [b"hash0"]
+    module = MagicMock()
+    module._ctx.token_hasher = MagicMock()
+    module._ctx.token_hasher.chunk_size = 256
+    module._ctx.token_hasher.compute_chunk_hashes.return_value = [b"hash0"]
```
ApostaC: Directly access `_ctx` (a private field) in the unit test is not recommended.
Please add a public method shared by all the modules to get the context. Also, clarify in the docstring that this function is only used for testing.

ApostaC (commented):
Please see the comments

sammshen (approved):
LGTM!

lmcache/v1/multiprocess/server.py
```diff
@@ -1403,124 +121,119 @@ def add_handler_helper(
...
+
+    Raises:
+        ValueError: If blend engine is requested with non-GPU transfer mode.
+    """
+    modules: list[EngineModule] = [
+        LookupModule(ctx),
+        ManagementModule(ctx),
+    ]
+
+    if mp_config.transfer_mode == "gpu":
```
maobaolong, line 145: As the code path for server-side shm based cpu transfer mode is as the same as `gpu` mode, so maybe `gpu` or `non-gpu` can make user confused, how about named it `TransferDataModule` and `TransferHandleModule` which are named by @hlin99 , or how about named it `ServerSideCopyTransferModule` and `ClientSideCopyTransferModule` ?  Or any other better names?

hlin99 (approved):
LGTM!

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -0,0 +1,670 @@
...
+        """
+        if instance_id in self._gpu_contexts:
+            logger.warning(
+                "Instance %s's KV cache is already registered, "
+                "skipping the new registration",
+                instance_id,
+            )
+            return
+
+        gpu_context = GPUCacheContext(
```
maobaolong, line 254: Ref to https://github.com/LMCache/LMCache/pull/3352/changes#diff-e02d00dbea27ffec36f8fa559186d06b4bf8bcbd9c82be453f208d9ab7db2832L300

for server-side shm based copy Cpu only case, it may not be `GPUCacheContext`, it can be `CpuCacheContext` too.

maobaolong:
@ApostaC Left some questions, but hope it don't block the merge of this PR.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -0,0 +1,467 @@
...
+    requested_tokens: int
+    # Captured at lookup time so the ``MP_LOOKUP_PREFETCH_END`` event can
+    # carry them as labels.  ``model_name`` lets dashboards slice hit rate
+    # per model in multi-model deployments; ``cache_salt`` slices per
+    # tenant / isolation domain (an empty string means no salt set).
+    model_name: str = ""
+    cache_salt: str = ""
+
+
+class LookupModule:
```
maobaolong, line 89: Why these sub classes didn't inherit `EngineModule` class?

maobaolong (approved):

lmcache/v1/multiprocess/modules/lookup.py
ApostaC: The EngineModule is a `Protocol`, not a base class. I would like to give it more flexibility, but could be changed to AbstractClass + Subclasses as well.

lmcache/v1/multiprocess/modules/gpu_transfer.py
ApostaC: Good catch. We can do the update in `gpu_transfer.py` once this PR is merged.

ApostaC:
@maobaolong @hlin99 Thanks for your review. I'm automerging the current PR now, and your PRs (#3359 , #3352 ) may need a rebase, thanks 🙏 !

PR #3393, opened 2026-05-26
[Operator]: Force external LMCache MP connector path

sammshen:
## Summary
- add `kv_connector_module_path` to the operator-generated `kv-transfer-config.json`
- update operator tests and golden snapshot to enforce the new connection contract
- document that the operator path expects a vLLM build new enough to honor explicit external connector loading

## Why
The operator currently emits only `kv_connector=LMCacheMPConnector`, which still leaves room for vLLM to resolve a vendored builtin connector path instead of the external LMCache MP connector. Making `kv_connector_module_path` explicit forces the intended external path and makes missing/outdated LMCache integration fail more clearly instead of silently drifting into MQ protocol mismatches.

## Notes
- PR base: `dev`
- I could not run Go formatting/tests in this environment because the Go toolchain is not installed here.


royyhuang (approved):
LGTM

ApostaC (approved):
LGTM!

PR #3413, opened 2026-05-27
[Perf] Avoid redundant PCIe transfer on leader rank during retrieve

yoo-kumaneko:
# TLDR
The rank 0 transfers the data from CPU to GPU during broadcast, then mistakenly discards the GPU tensor and lets `batched_to_gpu` re-transfer the same data from CPU — paying the CPU→GPU cost twice per retrieve. This PR removes the redundancy.

When `save_only_first_rank` is enabled (the MLA path), the leader rank does a redundant PCIe transfer of every cached chunk on every retrieve — a structural ~9 ms tax on the critical path that also makes the leader trail the passive ranks by the same amount.

**Root cause** (verified via timing instrumentation, see "Measurements" below):

In `_broadcast_or_receive_memory_objs`, the leader path uploads each chunk to GPU once for the NCCL send buffer:

```python
tensor_to_broadcast = raw_tensor.to(f"{torch_device_type}:{worker_id}")
self.broadcast_fn(tensor_to_broadcast, ...)
```

But it never substitutes this GPU tensor back into `reordered_chunks`. The original CPU-resident `memory_obj` (from L1 backend) stays there. The subsequent `batched_to_gpu(memory_objs, ...)` then walks the same CPU bytes again — this time through `cudaHostGetDevicePointer` zero-copy mapping inside `multi_layer_kv_transfer`, where each kernel thread's load triggers a PCIe transaction.

So the same L1 bytes traverse PCIe **twice** on the leader: once into the broadcast send buffer, once into the paged KV cache via the kernel. Passive ranks pay neither cost — they receive the broadcast directly into a fresh GPU buffer and then scatter into the paged KV from HBM.

**The fix:** keep the broadcast's GPU copy alive on the leader (`self._leader_gpu_substitute_objs`) and substitute it as the source for `batched_to_gpu`. The kernel then reads from HBM (~3 TB/s) instead of pinned host memory over PCIe (~32 GB/s). The original CPU `memory_obj` remains in `reordered_chunks` so the existing cleanup / ref-counting paths are unchanged. Passive ranks are unaffected.

## Measurements

Setup: tp=8 on 8× H20 + PCIe Gen4, DeepSeek-R1-style MLA model, in-process connector (`LMCacheConnectorV1`), warm retrieve of 8192 tokens (8 chunks at chunk_size=1024), c=1.

Per-rank `batched_to_gpu` GPU-stream time (CUDA events on `load_stream`):

| rank          | before | after  |
|---------------|-------:|-------:|
| 0  (leader)   |  9.42 ms |  0.70 ms |
| 1 (passive)   |  0.48 ms |  0.48 ms |
| 2 (passive)   |  0.48 ms |  0.48 ms |
| 3 (passive)   |  0.47 ms |  0.48 ms |
| 4 (passive)   |  0.48 ms |  0.48 ms |
| 5 (passive)   |  0.48 ms |  0.48 ms |
| 6 (passive)   |  0.48 ms |  0.48 ms |
| 7 (passive)   |  0.48 ms |  0.48 ms |

Total CPU wall-clock for the leader's `retrieve()` drops from **~21 ms** to **~12 ms** (matching the passive ranks). End-to-end TTFT at c=1 improves by ~10 ms (~133 ms → ~123 ms in our measurements).

After the fix, all 8 ranks finish `retrieve()` within ~0.2 ms of each other — the structural asymmetry is gone.

Independently verified by inspecting the device of `memory_objs` passed into `batched_to_gpu`:

```
Before:  rank=0 leader=True  devices=['cpu', 'cpu', 'cpu']
After:   rank=0 leader=True  devices=['cuda:0', 'cuda:0', 'cuda:0']
```

- The fix is gated by `save_only_first_rank and is_first_rank()` so it only affects the MLA leader path. Passive ranks and the non-MLA path are completely unchanged.
- `reordered_chunks` is intentionally left untouched. The cleanup / ref-count loop further down in `retrieve()` still operates on the original CPU `memory_obj` references, preserving existing semantics for `remove_after_retrieve` and PD-backend ref counting.
- After `batched_to_gpu` returns, the substitute list is cleared so the temporary GPU tensors can be freed by Python's GC. The `load_stream.synchronize()` inside `batched_to_gpu` guarantees the kernel has finished consuming them before we drop the references.
- The substitute list is a simple Python list of `TensorMemoryObj(parent_allocator=None)`, so dropping it does not interact with any allocator.

## If applicable

No user-facing API changes; this is a transparent perf fix in an internal code path. Existing tests for `LMCacheEngine.retrieve()` cover the changed path (correctness of retrieved tokens / `ret_mask`).

---

This change was developed with AI assistance (Anthropic Claude, Opus 4.7). The submitter has reviewed every changed line and reproduced the measurements above on local hardware.


lmcache/v1/cache_engine.py
```diff
@@ -852,9 +852,35 @@ def retrieve(
...
                 )
+                # Release GPU substitute references so the temporary buffers
+                # can be freed; original memory_objs in reordered_chunks are
+                # still tracked for the cleanup loop below.
+                if (
+                    self.save_only_first_rank
+                    and self.metadata.is_first_rank()
+                    and hasattr(self, "_leader_gpu_substitute_objs")
+                ):
+                    self._leader_gpu_substitute_objs = []
```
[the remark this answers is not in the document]
yoo-kumaneko: Thanks — applied in 8fe21db6 with the suggested `try/finally` shape.

One small nuance worth recording for any human reviewer reading later: this is defensive cleanup rather than fixing an unbounded leak. `self._leader_gpu_substitute_objs` is reassigned to `[]` at the top of `_broadcast_or_receive_memory_objs` on every leader-side retrieve (in this same patch), so even without the `try/finally` the worst case is "tensors held until the *next* successful retrieve on rank 0" — bounded to one retrieve cycle, not persistent. Still, the `try/finally` is the right semantic and the cost is negligible, so it's in.

ApostaC:
@maobaolong @chunxiaozheng Can you take a look at this PR? You guys are more familiar with Deepseek first rank optimizations.

chunxiaozheng:
> @maobaolong @chunxiaozheng Can you take a look at this PR? You guys are more familiar with Deepseek first rank optimizations.

@ApostaC Sure, I will take a look later.

lmcache/v1/cache_engine.py
```diff
@@ -852,9 +858,36 @@ def retrieve(
...
+                # When save_only_first_rank is enabled, the leader rank's
+                # memory_objs from L1 are CPU-resident. The broadcast above
+                # already created GPU-resident copies on the leader to use as
+                # the NCCL send buffer. Substitute those here so this kernel
+                # reads from HBM rather than re-reading the same L1 bytes
+                # over PCIe via zero-copy mapped pinned memory. Without this
+                # swap, batched_to_gpu on the leader takes ~9 ms (PCIe-bound)
+                # while passive ranks take ~0.5 ms (HBM-bound) — a structural
+                # asymmetry on the critical path of every retrieve.
+                memory_objs_for_togpu = list(memory_objs)
```
maobaolong: Can you put this line inside the else block? I think it might be better that way, what do you think?

lmcache/v1/cache_engine.py
```diff
@@ -852,9 +858,36 @@ def retrieve(
...
+                if (
+                    self.save_only_first_rank
+                    and self.metadata.is_first_rank()
+                    and len(self._leader_gpu_substitute_objs)
```
chunxiaozheng: if `len(self._leader_gpu_substitute_objs)` not equals to `len(memory_objs_for_togpu)`, which represents some error happens, could we add some error or warning log here?

lmcache/v1/cache_engine.py
yoo-kumaneko: Done in 59eed356 — `memory_objs_for_togpu = list(memory_objs)` is now only constructed on the non-leader path (and on the leader fallback path when the substitute list doesn't match, which should not happen in practice). Thanks!

lmcache/v1/cache_engine.py
yoo-kumaneko: Good catch — added in 59eed356. On the leader rank, when the substitute list count does not match the memory_objs count we now log a `logger.warning(...)` describing the anomaly and fall back to the CPU L1 source so retrieval is still correct (just paying the ~9 ms PCIe cost rather than failing). The broadcast always runs before the togpu block when `save_only_first_rank` is enabled, so a mismatch indicates a bug or stale state worth surfacing. Thanks!

chunxiaozheng (approved):
LGTM!

maobaolong (approved):
+1

PR #3419, opened 2026-05-27
[Core] add support for hybrid memory allocator

KuntaiDu:

Add support for hybrid memory allocator

**If applicable**:


lmcache/v1/kv_cache_groups.py
```diff
@@ -0,0 +1,179 @@
...
+
+# Standard
+import json
+from collections.abc import Iterable, Sequence
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True)
+class LMCKVCacheGroup:
```
ApostaC: If this class (and `LMCKVCacheGroup`) are going to be used in IPC, we can put it in `custom_types.py` and use `msgspec` to do the serialize and deserialize.

lmcache/v1/kv_cache_groups.py
```diff
@@ -0,0 +1,179 @@
...
+    """One LMCache view of an engine-side KV cache group."""
+
+    engine_kv_cache_group_id: int
+    """Engine-side KV cache group ID used to choose the request block-id list."""
+
+    layer_names: tuple[str, ...]
+    """Layer names assigned to this engine KV cache group."""
```
ApostaC: How will the server use the layer_names? I'm thinking whether we need to do a lot of transfers for the strings.

lmcache/v1/kv_cache_groups.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def expand_engine_block_ids_to_lmc_groups(
+        self,
+        block_ids_per_engine_group: Sequence[Sequence[int]],
+    ) -> list[list[int]]:
+        """Expand engine-group block IDs to LMCache-group block IDs.
+
+        vLLM reports block IDs indexed by its engine KV cache groups. LMCache
+        transfer requests are indexed by inflated LMCache KV layer groups, so
+        each LMCache group reuses the block IDs from its source engine group.
```
ApostaC: nit: let's have an "Args" section and a "Returns" section in the docstring

lmcache/v1/kv_cache_groups.py
ApostaC: I realized that this is not just pure data, but also includes some useful methods. Maybe we can move the serialize & deserialize logic to `custom_types.py` and keep the class and methods definition here.

Another way is to define them as data-only class, and have helper functions that takes it as input in this file

lmcache/v1/kv_cache_groups.py
```diff
@@ -0,0 +1,179 @@
...
+            block_ids_per_lmc_group.append(
+                list(block_ids_per_engine_group[engine_group_id])
+            )
+        return block_ids_per_lmc_group
+
+    def per_layer_engine_group_indices(
+        self,
+        num_registered_layers: int,
+    ) -> list[int] | None:
+        """Return engine group index per registered KV tensor."""
```
ApostaC: nit:  let's have an "Args" section and a "Returns" section in this docstring as well

Also, since it's a method, maybe rename it to `get_per_layer_engine_group_indices` so that it starts with an verb.

lmcache/v1/kv_layer_groups.py
```diff
@@ -35,11 +40,101 @@
...
+    for group in lmc_kv_cache_groups.groups:
+        for layer_idx, layer_name in zip(
+            group.layer_indices, group.layer_names, strict=False
+        ):
+            if 0 <= layer_idx < num_layers:
+                names[layer_idx] = layer_name
+    return names
+
+
+def group_layers_by_identity(
```
ApostaC, line 53: nit: better to have `Args` and `Returns` in the docstring for this function and `inflate_lmc_kv_cache_groups`

lmcache/v1/kv_layer_groups.py
```diff
@@ -35,11 +40,101 @@
...
+        engine_group_idx = (
+            per_layer_engine_group_idx[idx]
+            if per_layer_engine_group_idx is not None
+            else 0
+        )
+        groups_dict[(kv_size, nh, hs, bs, engine_group_idx, dt)].append(idx)
+    return sorted(groups_dict.items(), key=lambda kv: kv[1][0])
+
+
+def inflate_lmc_kv_cache_groups(
```
ApostaC: nit: instead of calling `inflate`, maybe use `create`? It may not always be `inflate` in the future (we will probably do some merging)

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -360,22 +363,57 @@ def get_tmp_chunk_gpu_buffer_batched(
...
+            len(block_ids_per_lmc_group) == 1
+            and num_groups > 1
+            and {group.engine_group_idx for group in groups} == {0}
+        ):
+            return [list(block_ids_per_lmc_group[0]) for _ in range(num_groups)]
+
+        raise ValueError(
+            "Expected block IDs for "
+            f"{num_groups} LMCache KV groups, got {len(block_ids_per_lmc_group)}"
+        )
```
ApostaC: Why do we need this?

IIUC, after this PR, the caller of `store` and `retrieve` should make sure that the length of the `list[list[block_ids]]` should match the length of lmcache kv groups.
In what case will this piece of logic be useful?

lmcache/integration/vllm/kv_cache_groups.py
```diff
@@ -0,0 +1,83 @@
...
+                if name in layer_to_idx
+            ),
+        )
+        for engine_kv_cache_group_id, group in enumerate(
+            _vllm_kv_cache_groups(kv_cache_config)
+        )
+    )
+
+
+def inflated_lmcache_kv_cache_groups_from_vllm(
```
ApostaC: nit: similar to above, maybe use `create` rather than `inflate`

lmcache/integration/vllm/kv_cache_groups.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _vllm_kv_cache_groups(kv_cache_config: Any) -> Sequence[Any]:
+    """Return vLLM KV cache groups, or an empty sequence when unavailable."""
+    if kv_cache_config is None:
+        return ()
+    return getattr(kv_cache_config, "kv_cache_groups", ()) or ()
+
+
+def lmcache_kv_cache_groups_from_vllm(
```
ApostaC: Will this function be called by any external modules? Or this is just a internal helper function which is used by `inflated_lmcache_kv_cache_groups_from_vllm`?

For the latter case, we can have a `_` at the beginning of this function's name

lmcache/cli/commands/bench/server_bench/helpers.py
ApostaC, line 1: This reminds me: do we want to add the kv layer group description in `lmcache describe kvcache` cli? (Or it's already there?)

ApostaC (commented):
Done a pass. Looking forward to some testing results!

lmcache/v1/kv_layer_groups.py
```diff
@@ -35,11 +40,142 @@
...
 
 
-# The 4-tuple that uniquely identifies a set of kernel-equivalent layers:
-# ``(kv_size, num_heads, head_size, block_size, dtype)``. Two layers share a transfer-
-# kernel launch iff they share this identity — see the grouping loop in
-# :meth:`KVLayerGroupsManager.__init__` for the derivation.
-LayerGroupIdentity = tuple[int, int, int, int, torch.dtype]
+# The tuple that uniquely identifies a set of kernel-equivalent layers:
+# ``(kv_size, num_heads, head_size, block_size, engine_group_idx, dtype)``.
+# The engine group slot keeps layers with disjoint engine block-ID spaces from
```
KuntaiDu: This should probably be (..., lmcache_group_idx, ...)

lmcache/v1/kv_cache_groups.py
```diff
@@ -0,0 +1,282 @@
...
+        engine_kv_cache_group_id, layer_names, layer_indices = payload
+        return cls(
+            engine_kv_cache_group_id=int(engine_kv_cache_group_id),
+            layer_names=tuple(str(name) for name in layer_names),
+            layer_indices=tuple(int(idx) for idx in layer_indices),
+        )
+
+
+@dataclass(frozen=True)
+class LMCKVCacheGroups:
```
KuntaiDu: Change this to `LMCacheKVSpec`

lmcache/v1/kv_cache_groups.py
KuntaiDu: Fixed.

lmcache/v1/kv_cache_groups.py
KuntaiDu: Done

lmcache/v1/kv_layer_groups.py
KuntaiDu: Done.

lmcache/cli/commands/bench/server_bench/helpers.py
KuntaiDu: We should. We can do this after this PR.

lmcache/integration/vllm/kv_cache_groups.py
KuntaiDu: Fixed.

lmcache/v1/kv_layer_groups.py
KuntaiDu: Fixed.

lmcache/v1/kv_layer_groups.py
KuntaiDu: Fxied.

lmcache/v1/kv_cache_groups.py
KuntaiDu: Fixed.

lmcache/integration/vllm/kv_cache_groups.py
KuntaiDu: Fixed. I reduced the # of function from vllm group to lmcache spec to just 1. It is a public function that will be imported by mp connector now.

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -395,34 +404,48 @@ def store(
...
+                        if chunk_block_ids_gpu.shape[0] != blocks_per_chunk:
+                            # Fail closed: a short block-id slice would make the
+                            # transfer kernel read out-of-bounds GPU memory.
+                            raise ValueError(
+                                "STORE block ID underflow: "
+                                f"lmc_group_idx={lmc_group_idx} "
+                                f"hybrid_group={group.hybrid_block_group_idx} "
+                                f"chunk={idx} expected={blocks_per_chunk} "
+                                f"got={chunk_block_ids_gpu.shape[0]}"
+                            )
```
ApostaC: A potential alternative: we log the error and continue processing other keys.
At the end of this function, we need to remove the memory objects that have failed.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -360,22 +363,81 @@ def get_tmp_chunk_gpu_buffer_batched(
...
+
+        return [
+            self.block_ids_buffer_[offsets[i] : offsets[i + 1]]
+            for i in range(len(block_ids_per_lmc_group))
+        ]
+
+    def normalize_lmc_group_block_ids(
+        self,
+        block_ids_per_lmc_group: list[list[int]],
+    ) -> list[list[int]]:
```
ApostaC: I still feel like this function is not needed.

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,198 @@ class CBMatchResult:
...
+
+class LMCacheKVGroup(msgspec.Struct, frozen=True):
+    """One LMCache hybrid block group: layers sharing one block-id space.
+
+    A ``msgspec.Struct`` so it can be encoded/decoded directly as part of a
+    multiprocess IPC payload.
+    """
+
+    hybrid_block_group_id: int
+    """Hybrid block group ID; selects which request block-id list applies."""
```
ApostaC: I think the previous name `engine_block_group_id` sounds clearer. WDYT?

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,198 @@ class CBMatchResult:
...
+    def num_hybrid_block_groups(self) -> int:
+        """Number of hybrid block groups (block-id lists per transfer request).
+
+        Returns:
+            ``max(hybrid_block_group_id) + 1`` (ids are dense from 0), or ``1``
+            for an empty spec (single-group fallback).
+        """
+        if not self.groups:
+            return 1
+        return max(group.hybrid_block_group_id for group in self.groups) + 1
```
ApostaC: Looks like this function is only used internally. Consider make it private?

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,198 @@ class CBMatchResult:
...
+
+        Returns:
+            The number of LMCache KV groups, or ``1`` for an empty spec
+            (single-group fallback).
+        """
+        if not self.groups:
+            return 1
+        return len(self.groups)
+
+    def hybrid_block_group_ids_by_lmc_group(self) -> tuple[int, ...]:
```
ApostaC: Same for this function -- it's only used inside this class

Also, it would be super helpful to clarify the length of the return value. Is it the same as the LMcache-side groups or the engine-side groups?

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,198 @@ class CBMatchResult:
...
+            hybrid block group id that LMCache group ``i`` draws block IDs from.
+            ``(0,)`` for an empty spec (single-group fallback).
+        """
+        if not self.groups:
+            return (0,)
+        return tuple(group.hybrid_block_group_id for group in self.groups)
+
+    def expand_block_ids_to_lmc_groups(
+        self,
+        block_ids_per_hybrid_block_group: Sequence[Sequence[int]],
```
ApostaC: nit: although this variable name sounds more neutral, it's a bit hard to grasp the meaning. What about `engine_side_block_ids`? And we specify the expectation (list of list of int, each list corresponds to the engine-side group) in the docstring.

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,198 @@ class CBMatchResult:
...
+                    "Missing block IDs for hybrid block group "
+                    f"{hybrid_block_group_id}; got "
+                    f"{len(block_ids_per_hybrid_block_group)} groups"
+                )
+            block_ids_per_lmc_group.append(
+                list(block_ids_per_hybrid_block_group[hybrid_block_group_id])
+            )
+        return block_ids_per_lmc_group
+
+    def get_per_layer_hybrid_block_group_indices(
```
ApostaC: Maybe rename it to `get_engine_group_indices`? (remove `per_layer` and rename `hybrid_block_group` -> `engine_group`)

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -395,8 +396,14 @@ class LoadStoreOp:
     token_ids: list[int]
     """Token IDs for the load/store operation"""
 
-    block_ids: list[int]
-    """Block ids for the load/store operation"""
+    block_ids: list[int] | list[list[int]]
```
ApostaC: Is it to force block_ids to be `list[list[int]]`?
I remember in the old code, vLLM already passes `list[list[int]]` and we manually do the flatten. We can remove the flatten since this PR.

If we do this, we can simplify the docstring as well as `block_ids_per_engine_group` function a lot.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -408,8 +415,27 @@ class LoadStoreOp:
...
+        """Return all block IDs flattened for engine-group-blind error paths."""
+        return [
+            block_id
+            for engine_group_block_ids in self.block_ids_per_engine_group
+            for block_id in engine_group_block_ids
+        ]
+
     def __len__(self) -> int:
-        return len(self.block_ids)
+        return len(self.block_ids_per_engine_group[0])
```
ApostaC: What's the semantics of `LoadStoreOp.__len__` for hybrid models? Maybe we need to consider removing this function.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -940,22 +967,34 @@ def is_first_rank_of_pp_group(self) -> bool:
...
         self.kv_caches = kv_caches
+        self.lmc_kv_cache_groups = lmc_kv_cache_groups
         self._send_register_kv_caches_request(kv_caches)
 
+    def _block_ids_per_lmc_group(self, op: LoadStoreOp) -> list[list[int]]:
+        lmc_kv_cache_groups = self.lmc_kv_cache_groups or LMCacheKVSpec()
+        return lmc_kv_cache_groups.expand_block_ids_to_lmc_groups(
+            op.block_ids_per_engine_group
+        )
+
```
ApostaC: Since this function will be called every time during store and retrieve, can we avoid creating a new `LMCacheKVSpec` object every time?

lmcache/integration/vllm/kv_cache_groups.py
```diff
@@ -0,0 +1,102 @@
...
+        LMCacheKVGroup(
+            hybrid_block_group_id=hybrid_block_group_id,
+            layer_indices=tuple(
+                layer_to_idx[name]
+                for name in getattr(group, "layer_names", ())
+                if name in layer_to_idx
+            ),
+        )
+        for hybrid_block_group_id, group in enumerate(vllm_groups)
+    )
```
ApostaC: Can we avoid creating a `LMCacheKVSpec` here? It's a bit confusing.

Looks like we only use the `get_per_layer_hybrid_block_group_indices` (which is a function only called here). Why don't we make that function a helper function inside this module instead of in `LMCacheKVSpec`?

lmcache/integration/vllm/kv_cache_groups.py
```diff
@@ -0,0 +1,102 @@
...
+    # the real registered tensors. ``group_layers_by_identity`` is the shared,
+    # engine-neutral grouping primitive (the server reuses it too).
+    gpu_kv_format, normalized_kv_caches = normalize_kv_and_discover_format(
+        list(kv_caches.values()),
+        EngineType.VLLM,
+        layout_hints=layout_hints,
+    )
+    num_layers = get_num_layers(normalized_kv_caches, gpu_kv_format)
+    if num_layers == 0:
+        return LMCacheKVSpec()
```
ApostaC: We probably want to raise an exception or log some error here?

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -78,20 +94,32 @@
 
 
 # Helper functions
-def reformat_block_ids(block_ids: tuple[list[int], ...] | None) -> list[int]:
-    if block_ids is None:
-        return []
-    assert isinstance(block_ids, tuple), (
-        f"Expected block_ids to be a tuple of lists, but got {type(block_ids)}"
-    )
+def normalize_block_ids_per_engine_group(
```
ApostaC: Let's still call it `reformat_block_ids`. 

I think having too many `per_engine_group` or `per_group` in the code will hurt the readability. We should treat the `block_id` as an object that already captures information from all the groups. In the code, we don't break the `block_ids` then use a subset of the block ids, so emphasizing `per_group` is not needed.

ApostaC (commented):
Mostly nit comment.  PTAL!

lmcache/integration/vllm/lmcache_mp_connector.py
yoo-kumaneko: Agree. The long function name is difficult to read and understand, so we should shorten it. That said, `reformat_block_ids` may need a second thought, since the function doesn't actually "reformat" the block IDs.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -267,14 +295,26 @@ def increase_num_stored_blocks(self, num_new_blocks: int):
         """
         self.num_stored_blocks += num_new_blocks
 
-    def append_block_ids(
+    def append_block_ids_per_engine_group(
```
yoo-kumaneko: The new function name sounds like it operates on the block IDs of a single engine group. How about keeping the name `append_block_ids`?

Groups of block IDs are still fundamentally a collection of block IDs, so omitting the "per engine group" suffix shouldn't cause much ambiguity. The simpler name also feels more general and easier to read, and is also used in other places like the dict `allocated_block_ids`.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -302,11 +343,30 @@ class LMCacheMPRequestMetadata:
     op: LoadStoreOp
     cache_salt: str = ""
 
+    @staticmethod
+    def _per_engine_group_slice(
```
yoo-kumaneko: It's not immediately clear whether start/end refer to token indices or block indices. It might be helpful if we briefly docstring it here, though it's a private function.

lmcache/integration/vllm/kv_cache_groups.py
```diff
@@ -0,0 +1,102 @@
...
+
+    per_layer_group_idx = engine_kv_spec.get_per_layer_hybrid_block_group_indices(
+        num_layers
+    )
+    return LMCacheKVSpec.from_groups(
+        LMCacheKVGroup(
+            hybrid_block_group_id=identity[4],
+            layer_indices=tuple(indices),
+        )
```
yoo-kumaneko, line 93: what do these code do?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -940,22 +967,34 @@ def is_first_rank_of_pp_group(self) -> bool:
             == 0
         )
 
-    def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]) -> None:
+    def register_kv_caches(
```
yoo-kumaneko, line 967: Why do we want to have `register_kv_caches` and `_send_register_kv_caches_request` seperately when the first one does almost nothing but call the second one?

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: Not sure, because LMCache does not see engine groups at server side, it only sees `LMCacheKVSpec` and block ids corresponding to that spec. But yeah, `hybrid_block_group_id` is also not a good name. Maybe `kv_spec_idx` or something?

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: This might be used for future observability purpose so I keep it public, and it is a `property` method so keeping it public might be fine.

lmcache/integration/vllm/vllm_multi_process_adapter.py
KuntaiDu: Main concern: do we maintain compatibility with older vLLM that only has `list[int]`? If we don't have to, then I fully support this!

lmcache/integration/vllm/kv_cache_groups.py
KuntaiDu: In each vLLM group, the tensors in the same group might have different hidden dimension and thus requires different copy kernel. This code separates them into different groups. After this code, each `LMCacheKVSpec` can simply use the same GPU copy kernel.

Let me docstring this.

lmcache/integration/vllm/vllm_multi_process_adapter.py
KuntaiDu: lol yeah. But this design is an existing design in LMCache, so this PR keeps it in order to make it easier to review. We can change it in the future.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: Maybe just `engine_block_group_id`, and call the LMCacheKVSpec something like `EngineGroup` because it seems to be easier to digest.

lmcache/integration/vllm/vllm_multi_process_adapter.py
KuntaiDu: Yeah we should.

lmcache/integration/vllm/vllm_multi_process_adapter.py
KuntaiDu: I'll assume we don't need to support ancient version of vLLM for now. @ApostaC if you think we need to, please let me know.

lmcache/integration/vllm/vllm_multi_process_adapter.py
KuntaiDu: Fixed.

lmcache/v1/multiprocess/modules/gpu_transfer.py
KuntaiDu: Fixed.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: Fixed.

lmcache/integration/vllm/kv_cache_groups.py
KuntaiDu: done.

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,161 @@ class CBMatchResult:
...
+# package, not here.
+#
+# LMCache's neutral KV cache spec is simply a ``list[EngineGroup]`` (passed as a
+# ``Sequence[EngineGroup]`` where only order matters). The group order is the
+# protocol-visible LMCache group order used by store/retrieve block IDs. An
+# empty list means a single non-hybrid group (the default for engines that do
+# not report KV cache group metadata).
+
+
+class EngineGroup(msgspec.Struct, frozen=True):
```
ApostaC: `class LMCacheGroupView`

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,161 @@ class CBMatchResult:
...
+    """One LMCache KV group: layers that share a single GPU copy kernel.
+
+    Several ``EngineGroup`` instances may share the same
+    ``engine_block_group_id`` when one engine block group is split by physical
+    transfer identity (e.g. differing hidden dims). A ``list[EngineGroup]`` is
+    carried verbatim in the ``REGISTER_KV_CACHE`` IPC payload; the message queue
+    handles encoding/decoding.
+    """
+
+    engine_block_group_id: int
```
ApostaC: `engine_group_id`

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,161 @@ class CBMatchResult:
...
+    """Engine block group this group's layers live in (one distinct paged-block
+    address space). Selects which request block-id list applies."""
+
+    layer_indices: tuple[int, ...] = ()
+    """Registered KV tensor indices assigned to this group."""
+
+
+def num_engine_block_groups(groups: Sequence[EngineGroup]) -> int:
```
ApostaC: `num_engine_groups`

lmcache/v1/multiprocess/custom_types.py
```diff
@@ -403,3 +404,161 @@ class CBMatchResult:
...
+    """Return the number of engine block groups (block-id lists per request).
+
+    Args:
+        groups: The LMCache KV groups, in protocol order.
+
+    Returns:
+        ``max(engine_block_group_id) + 1`` (ids are dense from 0), or ``1`` for
+        an empty ``groups`` (single non-hybrid group).
```
ApostaC: nit: add a note here saying "we assume engine group id starts from 0 and it's consecutive"

lmcache/v1/multiprocess/custom_types.py
ApostaC, line 1: helper functions can be put in a different file called `group_view.py` (or somthing similar)

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -308,17 +313,29 @@ def store(
...
             RuntimeError: If the backend does not support IPC event handles.
+
+        Notes:
+            Store is incremental. If ``gpu_block_ids`` cover fewer chunks than
+            ``key`` resolves to (e.g. a caller/protocol bug), only the longest
+            chunk prefix that every LMCache group can fully cover is stored; the
+            remainder is skipped (logged at WARNING) and may be stored by a
+            later request for the same prefix. Skipped chunks are never
+            committed, so a subsequent retrieve simply misses them and the
+            engine recomputes — there is no partially written cache entry.
```
ApostaC:
```suggestion
            Store is incremental. If ``gpu_block_ids`` cover fewer chunks than
            ``key`` resolves to (e.g. a caller/protocol bug), only the longest
            chunk prefix that every LMCache group can fully cover is stored; the
            remainder is skipped (logged at WARNING) and may be stored by a
            later request for the same prefix.
```

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -347,7 +364,32 @@ def store(
...
+            # fully present. ``storable_chunks`` is the longest chunk prefix that
+            # every LMCache group can fully cover; a shorter block-id list (e.g.
+            # a caller/protocol bug) would otherwise drive the transfer kernel to
+            # read out-of-bounds GPU memory. Capping ``obj_keys`` here means only
+            # storable chunks are reserved and finalized below — the rest are
+            # never committed and may be stored by a later request.
+            storable_chunks = min(
+                group_block_ids.shape[0] // blocks_per_chunk
+                for group_block_ids in block_ids_per_lmc_group_gpu
+            )
```
ApostaC: skip this logic.
We just skip the fault objects

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -395,34 +437,40 @@ def store(
...
-                    ]
-
-                    # Copy from GPU paged buffer to tmp buffer, then to CPU — per group
-                    for group_idx in range(num_groups):
-                        tmp_buffer = gpu_context.get_tmp_chunk_gpu_buffer(group_idx)
-                        group_kv_pointers = gpu_context.get_group_kv_pointers(group_idx)
+                    # Copy from GPU paged buffer to tmp buffer, then to CPU — per
+                    # group. The block-id slice is always full: ``obj_keys`` was
+                    # capped to the prefix every group can cover (storable_chunks).
+                    for lmc_group_idx in range(len(groups)):
```
ApostaC: kuntai says this is unnecessary diff

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -79,18 +95,9 @@
 
 # Helper functions
 def reformat_block_ids(block_ids: tuple[list[int], ...] | None) -> list[int]:
+    """Flatten vLLM's per-engine-group block IDs to the primary group's list."""
```
ApostaC: kuntai says this will be removed.

ApostaC (commented):
Almost good to go.

lmcache/integration/vllm/lmcache_mp_connector.py
KuntaiDu: Fixed.

lmcache/integration/vllm/vllm_multi_process_adapter.py
yoo-kumaneko: Sure. that makes sense.

lmcache/v1/kv_cache_groups.py
KuntaiDu: fixed.

lmcache/v1/kv_cache_groups.py
KuntaiDu: fixed.

lmcache/v1/multiprocess/gpu_context.py
KuntaiDu: fixed.

lmcache/v1/multiprocess/gpu_context.py
KuntaiDu: fixed.

lmcache/integration/vllm/kv_cache_groups.py
KuntaiDu: fixed.

lmcache/integration/vllm/kv_cache_groups.py
KuntaiDu: fixed.

lmcache/integration/vllm/lmcache_mp_connector.py
KuntaiDu: Removed this function.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: Fixed.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: Fixed.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: Fixed.

lmcache/v1/multiprocess/modules/gpu_transfer.py
KuntaiDu: fixed

lmcache/v1/multiprocess/modules/gpu_transfer.py
KuntaiDu: fixed.

lmcache/v1/multiprocess/modules/gpu_transfer.py
KuntaiDu: fixed.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: fixed.

lmcache/v1/multiprocess/custom_types.py
KuntaiDu: fixed.

ApostaC (approved):
LGTM as we discussed offline

sammshen (approved):

deng451e (approved):

ApostaC:
Merging this to unblock futher development

DongDongJu:
@KuntaiDu Thanks for great work!

Issue #3108, opened 2026-04-22
[Proposal] LMCache MP mode & Operator observability metrics


ApostaC:
## Overview

This issue tracks the proposed observability metrics for LMCache MP mode and LMCache Operator. Metrics are grouped into three categories: health monitoring, performance monitoring, and production insights.

## Generic tags

All metrics below should include the following generic tags:

- `lmcache_instance_id`
- `model_name`

## Health monitoring

| Category | Metric name | Metric description | Metric Type | Tags |
| --- | --- | --- | --- | --- |
| L1/L2 failure monitoring | `lmcache_l1_allocation_failure` | Memory allocation failures in L1 (only captures OOM) | counter (on function calls) | `during`: l1_store \| l2_prefetch |
| | `lmcache_l1_read_failure` | Failures when trying to read an object in L1 | counter (on function calls) | `during`: l2_store \| l1_retrieve; `reason`: not_found \| write_locked |
| | `lmcache_l2_prefetch_failure` | Counts the number of objects that are reported to exist in L2, but failed during L2 → L1 load | counter (on # of objects) | `reason`: l1_oom \| not_found \| serde_failure |
| | `lmcache_l2_serde_failure` | Counts the number of objects that have errors during serde (Note: need to wait for serde PR to get in) | counter (on # of objects or # of function calls) | `during`: l2_store \| l2_load |
| | `lmcache_l1_ttl_lock_expire` | Counts the number of TTLLock's unlocks that are triggered by TTL expiration | counter (on # of expires) | `type`: read \| write *(Note: hard to achieve this in the current code)* |
| L1 usage monitoring | `lmcache_l1_memory_usage` | Memory usage for L1 | gauge | |
| | `lmcache_num_inflight_l2_stores` | Monitor the number of in-flight L2 stores | gauge | `l2_name` |
| | `lmcache_num_inflight_l2_loads` | Monitor the number of in-flight L2 loads | gauge | `l2_name` |
| | `lmcache_inflight_load_memory_usage` | Monitors the L1 memory usage introduced by L2→L1 prefetch | gauge | `l2_name` |
| Event bus monitoring | `lmcache_event_bus_queue_depth` | Number of events queued in the event bus | gauge | |
| | `lmcache_event_bus_drain_lag_seconds` | The latency between the event enqueue and the processing of the event | gauge | |
| | `lmcache_event_bus_dropped_events_total` | Number of events that are dropped from the event bus | counter | |
[...]

PR #3428, opened 2026-05-28
fix: register EventBus health metrics via subscriber

RitwijParmar:

Refs #3108. The EventBus self-health docs and `setup_mp_observability()` wiring say queue depth, drain lag, dropped events, and subscriber exceptions are owned by `EventBusSelfMetricsSubscriber`. Two of those metrics (`queue_depth` and `drain_lag_seconds`) were still registered directly from `EventBus.__init__`, so constructing any EventBus, including disabled/test buses, could create OTel instruments even when metrics subscribers were not enabled.

This moves the queue-depth and drain-lag observable gauges into `EventBusSelfMetricsSubscriber`, keeping all four EventBus health metrics behind the metrics subscriber path. The `EventBus` still exposes the same public accessors (`queue_depth()` and `oldest_event_lag_seconds()`), so dashboard/query semantics are unchanged.

I read `CONTRIBUTING.md` and used a signed-off commit. This contribution was prepared with Codex assistance; I reviewed the diff and ran the checks below locally.

Checks run:
- `PYTHONPATH=. .venv/bin/python -m py_compile lmcache/v1/mp_observability/event_bus.py lmcache/v1/mp_observability/subscribers/metrics/event_bus.py tests/v1/mp_observability/subscribers/metrics/test_event_bus_self_metrics.py tests/v1/mp_observability/test_event_bus.py`
- `.venv/bin/python -m ruff check lmcache/v1/mp_observability/event_bus.py lmcache/v1/mp_observability/subscribers/metrics/event_bus.py tests/v1/mp_observability/subscribers/metrics/test_event_bus_self_metrics.py tests/v1/mp_observability/test_event_bus.py`
- Direct subscriber smoke check exercising the new gauge callbacks against a mocked OTel meter.

I also tried the focused pytest files, but the local macOS environment cannot load `lmcache.c_ops` from the repository test `conftest.py` without building native extensions.

**If applicable**:

### DIFF
--- a/lmcache/v1/mp_observability/event_bus.py
+++ b/lmcache/v1/mp_observability/event_bus.py
@@ -17,7 +17,6 @@ import time
 # First Party
 from lmcache.logging import init_logger
 from lmcache.v1.mp_observability.event import Event, EventType
-from lmcache.v1.mp_observability.otel_init import register_gauge
 
 try:
     # Third Party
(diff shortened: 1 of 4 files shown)


RitwijParmar:
Thanks, fixed in 995dce6c by avoiding direct _queue access in the drain-lag test. It now uses EventBus.publish() with patched time.time so the test exercises the public API.

ApostaC:
Can you explain what's wrong with the existing implementation in a simpler language?

RitwijParmar:
Yeah, the values themselves were fine. The issue was the side effect.

Before this PR, just creating an EventBus registered two OTel gauges. So even a disabled or test bus could create metrics instruments.

Since the other EventBus health metrics already live in `EventBusSelfMetricsSubscriber`, I moved these two there as well. That keeps the metric output the same, but makes registration happen only through the metrics subscriber path.

PR #3440, opened 2026-05-28
[fix] bench close event loop

Shaoting-Feng:
Anything that opens loop-bound I/O, `RequestSender._client._client` (the `httpx.AsyncClient` inside `AsyncOpenAI`), the `_SelectorSocketTransport` instances inside its `_pool`, the `workload._pending_tasks` (set of `asyncio.Task`), `workload._semaphore` (`asyncio.Semaphore`), must be closed inside the same event loop it was opened in.

If you let the loop die while the object still exists, the object becomes an orphan pointing at a dead loop, and any later operation that touches it will hit `RuntimeError: Event loop is closed`.

Concretely:

1. We run `asyncio.run(self._run_async())` in `run(self)` in `base.py`. We create BaseEventLoop A in `self._loop = events.new_event_loop()` in `_lazy_init` in `runners.py`. Then we reach `transport = self._make_socket_transport(sock, protocol, waiter)` in `base_events.py`. And it will call `self._loop = loop` in `_FlowControlMixin` in `transports.py`, which binds the BaseEventLoop A to `_SelectorSocketTransport`. A.close is called when this asyncio finishes, setting `A._closed = True`.

2. We run `asyncio.run(request_sender.close())` in `command.py`, that same `_SelectorSocketTransport` that lives inside `self._request_sender` still calls `A.call_soon(self._call_connection_lost, None)` at `selector_events.py:875`. `call_soon` internally runs `A._check_closed()`, which sees `A._closed == True` and raises `RuntimeError('Event loop is closed')`.

Why it's environment-dependent: before reaching the bad `transport.close()`, anyio checks `if not self._transport.is_closing()`. If the server closes the keep-alive socket before A dies, asyncio processes the FIN, marks the transport as closing, and step 2's check skips the crash. On CI the server doesn't close in time, so the check passes and the crash fires.

Concrete log: https://buildkite.com/organizations/lmcache/pipelines/k3-sglang-test/builds/327/jobs/019e6fd7-37e7-464e-96ee-a17fb656720a/artifacts/019e6fdd-7b62-4512-8e88-c454c8e98ef3.

**If applicable**:

### DIFF
--- a/lmcache/cli/commands/bench/engine_bench/command.py
+++ b/lmcache/cli/commands/bench/engine_bench/command.py
@@ -13,7 +13,6 @@ from __future__ import annotations
 # Standard
 from typing import TYPE_CHECKING
 import argparse
-import asyncio
 import os
 import sys
 
@@ -577,7 +576,6 @@ def run_engine_bench(command: "BaseCommand", args: argparse.Namespace) -> None:
         workload.run()
     finally:
         progress_monitor.stop()
-        asyncio.run(request_sender.close())
 
     # 7. Final metrics
     final = stats_collector.get_final_stats()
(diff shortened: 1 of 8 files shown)


ApostaC (approved):
LGTM!

deng451e (approved):

Issue #3438, opened 2026-05-28
[Bug][Examples] offload.py fails with pydantic validation error on vLLM 0.20.1+


yangyonggit:
**Label**: `bug`, `examples`

---

**Summary**

`examples/kv_cache_reuse/local_backends/offload.py` fails to run on vLLM 0.20.1 and 0.21.0 due to a pydantic validation error when constructing the LLM instance.

---

**Details**

The script uses `EngineArgs` + `asdict()` to build kwargs for `LLM()`. `asdict()` recursively serializes all dataclass fields including `None` values, which are then passed into `CompilationConfig`. Newer vLLM versions use pydantic v2 with strict validation, causing `None` to be rejected for fields like `cudagraph_capture_sizes` (expects `list`) and `pass_config.fuse_minimax_qk_norm` (expects `bool`).

Additionally, the v0 path has a separate issue: `VLLM_USE_V1=0` is set inside `setup_environment_variables()` which runs after `import vllm`, so the env var is ignored and the v1 engine is always used. The v1 engine does not recognize `LMCacheConnector`, causing a second validation error.

---

**Steps / Reproduction**

Environment:
- LMCache: nightly 

[error.log](https://github.com/user-attachments/files/28365655/error.log)

- vLLM: 0.20.1 / 0.21.0
- Python: 3.12.12

1. Install vLLM 0.20.1 or 0.21.0 with LMCache nightly
2. Run `python offload.py -v v1`
3. Observe pydantic `ValidationError` on `CompilationConfig`

For v0:
4. Run `VLLM_USE_V1=0 python offload.py -v v0`
5. Observe `Unsupported connector type: LMCacheConnector`

---

**Expected Outcome**

Both commands run without error and demonstrate KV cache reuse.

---

**Actual Outcome**

v1 crashes with:
```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for CompilationConfig
cudagraph_capture_sizes
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
pass_config.fuse_minimax_qk_norm
  Input should be a valid boolean [type=bool_type, input_value=None, input_type=NoneType]
```

v0 crashes with:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for VllmConfig
  Value error, Unsupported connector type: LMCacheConnector
```

---

**Additional Context**

Fix for v1: replace `EngineArgs + asdict()` with direct kwargs passed to `LLM()`. 

[...]

PR #3441, opened 2026-05-29
[Fix][Example]: Drop EngineArgs+asdict to fix vLLM 0.20+ pydantic error

zhengfeihe:
for issue: https://github.com/LMCache/LMCache/issues/3438
to make `offload.py` example run with latest vllm

I ran it on an instance, confirmed working. 

Also add a comment in the code explaining that V0 is supported only up to vLLM 0.10.x, as per https://github.com/vllm-project/vllm/issues/18571

**If applicable**:


zhengfeihe:
addressed comments

ApostaC (approved):
LGTM! Thanks for the quick fix!

sammshen (approved):
LGTM

PR #3437, opened 2026-05-28
  [MP][Bugfix] Restore MPCacheEngine HTTP-layer passthroughs

deng451e:
Description:

  ## Summary

  PR #3391 split the monolithic `MPCacheEngine` into a compositor + per-module handlers, but dropped three engine attributes that `http_apis/` reads directly (bypassing ZMQ). All three are silently broken on `upstream/dev`:

  | HTTP endpoint | Reads | Symptom on `dev` |
  |---|---|---|
  | `POST /clear-cache` (+ `lmcache kvcache clear` CLI) | `engine.clear()` | HTTP 500 `AttributeError` |
  | `GET /quota/*` | `engine.storage_manager` | HTTP 500 `AttributeError` |
  | `GET /kvcache/check` (+ `lmcache bench server` CLI) | `engine.gpu_contexts` | HTTP 501 silent "not supported" |

  This restores all three on `MPCacheEngine` as thin passthroughs:

  - `storage_manager` → `self._context.storage_manager`
  - `gpu_contexts` → finds `GPUTransferModule` and unwraps `GPUContextEntry → GPUCacheContext` so the pre-refactor contract holds
  - `clear()` → delegates to `ManagementModule.clear` (which owns the lock)

  Also fixes a stale `/api/` URL prefix in `lmcache bench server`'s checksum helper (path was renamed in an earlier PR; helper + its test were missed).

  ## Test plan

  - [x] New `tests/v1/multiprocess/test_engine_passthroughs.py` — 5 regression  tests covering each passthrough + the missing-module guards.
  - [x] Existing `tests/cli/commands/bench/test_server_bench.py` updated to match the corrected URL; all 23 tests pass.
  - [x] Live verification: started an HTTP server off this branch, hit each codes (200 / 200 / 404-without-instance) instead of 500 / 500 / 501.

PR #3442, opened 2026-05-29
fix(mp): restore HTTP-API surface on MPCacheEngine after compositor refactor (fixes #3431)

### LINKED ISSUE #3431: [Regression][MP][Core] MP HTTP APIs fail after `MPCacheEngine` compositor refactor
After the `MPCacheEngine` compositor refactor in #3391 , some MP HTTP endpoints still appear to depend on the old monolithic engine surface.

The server starts successfully, and `/healthcheck` plus `/status` work, but a few management/debug endpoints fail because the refactored `MPCacheEngine` no longer exposes `clear`, `storage_manager`, or `gpu_contexts`.

---

**Env**

- Branch/commit: dev at 3e0e3ff0
- Install: source install with CUDA extensions
- Platform: WSL2 Ubuntu

---

**Repro**

Start the server:

```bash
LMCACHE_LOG_LEVEL=DEBUG lmcache server \
  --host 127.0.0.1 \
  --port 15555 \
  --http-host 127.0.0.1 \
  --http-port 18080 \
  --l1-size-gb 0.25 \
  --eviction-policy LRU \
  2>&1 | tee /tmp/lmcache-mp-http-repro.log
```

Then run:

```bash
curl -i http://127.0.0.1:18080/healthcheck
curl -i http://127.0.0.1:18080/status
curl -i -X POST http://127.0.0.1:18080/clear-cache
curl -i http://127.0.0.1:18080/quota
curl -i "http://127.0.0.1:18080/kvcache/check?block_ids=0&chunk_size=1"
```

**Actual behavior**

`/healthcheck` works:

```text
HTTP/1.1 200 OK
{"status":"healthy"}
```

`/status works` and reports `engine_type: "MPCacheEngine"`

But:

```text
POST /clear-cache -> HTTP/1.1 500 Internal Server Error
GET /quota       -> HTTP/1.1 500 Internal Server Error
GET /kvcache/check?block_ids=0&chunk_size=1
                 -> HTTP/1.1 501 Not Implemented
{"error":"checksum not supported for this engine type"}
```

**Expected behavior**

These HTTP APIs should continue working with the refactored MP engine:

- POST `/clear-cache` should clear L1/cache storage
- GET `/quota` should access the shared `StorageManager`
- GET `/kvcache/check` should be able to inspect GPU contexts when the server is running in GPU transfer mode

---

**Likely Cause**

The HTTP APIs still use the old `MPCacheEngine` surface:

- `lmcache/v1/multiprocess/http_apis/cache_api.py` calls `engine.clear()`
- `lmcache/v1/multiprocess/http_apis/quota_api.py` calls `engine.storage_manager`
- `lmcache/v1/multiprocess/http_apis/cache_api.py` reads `engine.gpu_contexts`

After the refactor, `MPCacheEngine` only exposes `context`, `report_status()`, and `close()`. The relevant state/behavior moved into modules:

- `ManagementModule.clear()`
- `MPCacheEngineContext.storage_manager`
- `GPUTransferModule._gpu_contexts`

**Why this appears to pass existing tests**

The current HTTP tests use mock/stub engines that still have the old monolithic engine shape:

- `tests/v1/multiprocess/test_http_server.py` uses a mock engine with `clear()` and `gpu_contexts`.
- `tests/v1/multiprocess/test_http_quota_endpoints.py` uses a stub engine with `storage_manager`.

Those tests validate the HTTP handlers against the pre-refactor engine interface, so they do not exercise the real `MPCacheEngine` compositor returned by `run_cache_server()`. As a result, the endpoint-level tests pass even though the live `lmcache server` path fails

**Possible Fix**

Either preserve a compatibility facade on `MPCacheEngine`:

- `storage_manager` property forwarding to `context.storage_manager`
- `clear()` forwarding to `ManagementModule.clear()`
- `gpu_contexts` exposing GPU transfer module contexts

or update the HTTP APIs to use the new context/module APIs directly.

A regression test should instantiate the real refactored `MPCacheEngine` through the HTTP server path and assert that `/clear-cache`, `/quota`, and `/kvcache/check` do not depend on removed monolithic-engine attributes.

---

I’m happy to work on a PR for this once the preferred fix direction is clear.

Would maintainers prefer:
1. preserving a small compatibility facade on `MPCacheEngine` for these HTTP APIs, or
2. migrating the HTTP APIs to access the new `context` / module APIs directly?

My initial read is that the facade is the smaller compatibility fix, while direct module/context access may be cleaner long-term. I can implement either direction and add regression coverage for the live refactored engine path.

abinggo:
## Summary

After #3391 split engine state across modules, the live `lmcache server` HTTP endpoints crash because the handlers in `lmcache/v1/multiprocess/http_apis/` reach for `engine.clear`, `engine.storage_manager`, and `engine.gpu_contexts` that no longer exist on the compositor. This restores those surfaces as a thin compatibility facade so HTTP handlers stay decoupled from module internals.

Fixes #3431.

## Reproduction (no GPU required)

```bash
LMCACHE_LOG_LEVEL=DEBUG lmcache server \
    --host 127.0.0.1 --port 15555 \
    --http-host 127.0.0.1 --http-port 18080 \
    --l1-size-gb 0.25 --eviction-policy LRU

curl -i -X POST http://127.0.0.1:18080/clear-cache  # 500 before this PR, 200 after
curl -i http://127.0.0.1:18080/quota                # 500 before, 200 after
curl -i 'http://127.0.0.1:18080/kvcache/check?block_ids=0&chunk_size=1'
                                                    # 501 (clean) for non-GPU engines
```

## Approach

The reporter (@vmanvs) outlined two options in the issue: (1) compatibility facade on `MPCacheEngine`, or (2) migrate HTTP handlers to access the new context/module APIs directly. This PR goes with (1) for the smaller diff and to keep the HTTP surface stable across future module changes — happy to flip to (2) (or close in @vmanvs's favor if they'd rather drive the fix) on maintainer preference.

### Facade additions on `MPCacheEngine`

* `storage_manager` (property) — forwards to `_context.storage_manager`. Used by `/quota` and `/status`.
* `clear()` (method) — looks up the registered `ManagementModule` via a typed `_find_module(type[ModuleT]) -> ModuleT | None` helper and delegates. Locking and the `memcheck → clear(force=True) → memcheck` sequence are owned by `ManagementModule`, matching the pre-refactor behavior. Raises `RuntimeError` if a custom loadout drops `ManagementModule` so a silent no-op can't mask the same kind of regression.
* `gpu_contexts` (property) — matches the pre-refactor contract (always returns a `dict[int, GPUCacheContext]`, empty when the engine has no `GPUTransferModule`). Backed by a new public `GPUTransferModule.gpu_contexts` that snapshots `_gpu_contexts` before iterating, so a concurrent ZMQ-thread `register_kv_cache` cannot raise `dictionary changed size during iteration` in the comprehension.
* `supports_gpu_kvcache_check` (property) — capability flag the `/kvcache/check` endpoint reads to return a clean `501 Not Implemented` in non-GPU mode without overloading the `gpu_contexts` data contract with `None` (avoids `Optional` mapping return per `docs/coding_standards.md`).

### HTTP handler change

`http_apis/cache_api.py::kvcache_check` now reads `supports_gpu_kvcache_check` instead of testing `gpu_contexts is None`, since `gpu_contexts` always returns a dict after the fix. Behavior matrix:

| engine state | `supports_gpu_kvcache_check` | endpoint result |
|---|---|---|
| no `GPUTransferModule` (transfer_mode != gpu) | `False` | `501 Not Implemented` |
| GPU module present, `instance_id` not registered | `True` | `404 not registered` |
| GPU module present, `instance_id` registered | `True` | `200` + checksums |

## Why existing tests didn't catch this

Pre-existing endpoint tests (`tests/v1/multiprocess/http_apis/test_http_server.py`, `test_http_quota_endpoints.py`) hand the FastAPI app a `MagicMock` with the **pre-refactor** monolithic engine shape (`engine.clear()`, `engine.storage_manager`, `engine.gpu_contexts` all present), so they keep passing even though the live `lmcache server` path is broken. New regression test instantiates the real refactored `MPCacheEngine` compositor.

## Test plan

`tests/v1/multiprocess/test_mp_engine_http_facade.py` — 14 new tests:

**Facade unit (real compositor):**
- `test_storage_manager_forwards_to_context`
- `test_clear_routes_through_management_module` (verifies `memcheck → clear(force=True) → memcheck` call sequence)
- `test_clear_raises_without_management_module`
- `test_gpu_contexts_returns_empty_dict_in_non_gpu_mode`
- `test_supports_gpu_kvcache_check_false_in_non_gpu_mode`
- `test_gpu_contexts_returns_empty_snapshot_when_no_registrations`
- `test_supports_gpu_kvcache_check_true_in_gpu_mode`
- `test_gpu_contexts_unwraps_entry_to_gpu_context` (snapshot independence)
- `test_gpu_contexts_snapshot_survives_concurrent_dict_mutation` (race-condition pin)
- `test_find_module_returns_first_match`
- `test_find_module_returns_none_when_absent`

**HTTP integration (FastAPI `TestClient` + real compositor):**
- `test_http_clear_cache_with_real_engine_returns_200`
- `test_http_quota_list_with_real_engine_returns_200`
- `test_http_kvcache_check_non_gpu_engine_returns_501`

```bash
$ pytest tests/v1/multiprocess/test_mp_engine_http_facade.py -v
======================= 14 passed in 1.62s =======================

$ pre-commit run --files \
    lmcache/v1/multiprocess/server.py \
    lmcache/v1/multiprocess/modules/gpu_transfer.py \
    lmcache/v1/multiprocess/http_apis/cache_api.py \
    tests/v1/multiprocess/test_mp_engine_http_facade.py
ruff..............................Passed
ruff-format.......................Passed
isort.............................Passed
mypy..............................Passed
codespell.........................Passed
spdx..............................Passed
```

AI assistance was used in drafting this fix and tests; the human contributor reviewed every line and ran the test plan locally.

### DIFF
--- a/lmcache/v1/multiprocess/http_apis/cache_api.py
+++ b/lmcache/v1/multiprocess/http_apis/cache_api.py
@@ -117,13 +117,18 @@ async def kvcache_check(
             content={"error": "engine not initialized"},
         )
 
-    gpu_ctxs = getattr(engine, "gpu_contexts", None)
-    if gpu_ctxs is None:
+    # ``supports_gpu_kvcache_check`` distinguishes "engine has no GPU
+    # transfer module" (501 Not Implemented — clean capability signal)
+    # from "GPU module present but instance_id not registered" (404
+    # below). ``gpu_contexts`` itself always returns a dict, even when
+    # empty, so the previous ``is None`` probe would no longer fire.
+    if not getattr(engine, "supports_gpu_kvcache_check", False):
         return JSONResponse(
             status_code=501,
             content={"error": "checksum not supported for this engine type"},
         )
 
+    gpu_ctxs = engine.gpu_contexts
     ctx = gpu_ctxs.get(instance_id)
     if ctx is None:
         return JSONResponse(
--- a/lmcache/v1/multiprocess/modules/gpu_transfer.py
+++ b/lmcache/v1/multiprocess/modules/gpu_transfer.py
@@ -140,6 +140,35 @@ class GPUTransferModule:
         """Return the shared engine context. Exposed for testing only."""
         return self._ctx
 
+    @property
+    def gpu_contexts(self) -> dict[int, GPUCacheContext]:
+        """Snapshot of registered GPU contexts keyed by ``instance_id``.
+
+        Returns the underlying :class:`GPUCacheContext` per instance,
+        not the internal :class:`GPUContextEntry`. Diagnostic callers
+        (HTTP ``/kvcache/check``, ``status``) want shape/format access,
+        not registration metadata, so the entry is unwrapped here.
+
+        Note: ``_gpu_contexts`` is not lock-protected, mirroring the
+        pre-refactor ``MPCacheEngine`` behaviour. A concurrent
+        ``register_kv_cache`` / ``unregister_kv_cache`` from a ZMQ
+        worker thread while this comprehension is iterating can still
+        raise ``RuntimeError: dictionary changed size during
+        iteration``. That race predates this PR and is tracked
+        separately; properly fixing it requires a registry-wide lock
+        (covering register / unregister / close / iterate) that is
+        out of scope for the HTTP-API regression fix this property
+        enables.
+
+        Returns:
+            A fresh ``dict[instance_id, GPUCacheContext]``. Mutating
+            the returned dict has no effect on registration state.
+        """
+        return {
+            instance_id: entry.gpu_context
+            for instance_id, entry in self._gpu_contexts.items()
+        }
+
     def get_handlers(self) -> list[HandlerSpec]:
         """Return handler specs for all request types this module serves.
 
--- a/lmcache/v1/multiprocess/server.py
+++ b/lmcache/v1/multiprocess/server.py
@@ -2,6 +2,7 @@
 """MPCacheEngine compositor and unified cache server entry point."""
 
 # Standard
+from typing import TypeVar
 import argparse
 import time
 
@@ -16,6 +17,7 @@ from lmcache.v1.distributed.config import (
     add_storage_manager_args,
     parse_args_to_config,
 )
+from lmcache.v1.distributed.storage_manager import StorageManager
 from lmcache.v1.mp_observability.config import (
     ObservabilityConfig,
     add_observability_args,
@@ -34,6 +36,7 @@ from lmcache.v1.multiprocess.engine_module import (
     HandlerSpec,
     ThreadPoolType,
 )
+from lmcache.v1.multiprocess.gpu_context import GPUCacheContext
 from lmcache.v1.multiprocess.modules.gpu_transfer import GPUTransferModule
 from lmcache.v1.multiprocess.modules.lookup import LookupModule
 from lmcache.v1.multiprocess.modules.management import ManagementModule
@@ -46,6 +49,13 @@ from lmcache.v1.multiprocess.protocol import (
 )
 
 logger = init_logger(__name__)
+
+# Type variable for ``MPCacheEngine._find_module``: the lookup is invariant
+# in the requested module class, so callers receive the precise subclass
+# back rather than a bare ``EngineModule`` whose protocol-level surface
+# would not include the methods (``clear``, ``gpu_contexts``) the facade
+# delegates to.
+ModuleT = TypeVar("ModuleT", bound=EngineModule)
 
 
 class MPCacheEngine:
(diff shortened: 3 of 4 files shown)


abinggo:
Thanks for the careful read, @gemini-code-assist. You're right on both:

1. **`dict(self._gpu_contexts)` does not prevent the race.** CPython's `PyDict_Copy` is atomic in practice (single C call under the GIL), but Python's language spec doesn't promise this, so relying on it is brittle.
2. **The test never invokes `_ConcurrentMutatingDict.items()`.** `dict(d)` uses `PyDict_Next` / `PyDict_Copy` in C and doesn't call the source dict's Python `.items()`. The test was decorative.

Just pushed a cleanup that drops both the false-snapshot pattern and the misleading test (force-push `41555d43 → 121d27d7`). The underlying race is pre-existing — pre-refactor `MPCacheEngine.gpu_contexts` (before #3391) iterated `self.contexts` without a lock either — so I'd like to keep this PR focused on the HTTP-API regression and address the registry lock in a separate `[MP][Refactor]` follow-up. Happy to flip to in-PR locking if @ApostaC prefers it bundled.

ApostaC:
Hi @abinggo , thanks for your contribution and your interest! We have merged another fix #3437 yesterday.

PR #3445, opened 2026-05-29
cleanup: simplify redundant conditions in RawBlockCore

nayeonikim:

  Pure refactor cleanups in `raw_block/core.py`:
  - `delete_many`: drop `existed` (equivalent to `removed_entry is not None or inflight is not None` under
  `_lock`).
  - `_is_valid_checkpoint_entry`: drop unreachable `slot < 0` (`rel >= 0` guaranteed by the earlier offset
  check).
  - `_apply_loaded_state`: cache `data.get("device_path")` instead of calling it twice.

  No behavior change. Existing raw-block tests pass; 

  **If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/raw_block/core.py
+++ b/lmcache/v1/storage_backend/raw_block/core.py
@@ -669,7 +669,6 @@ class RawBlockCore:
         deleted: list[bool] = []
         with self._lock:
             for encoded_key in encoded_keys:
-                existed = encoded_key in self._index or encoded_key in self._inflight
                 entry = self._index.get(encoded_key)
                 locked = self._lock_refcnt.get(encoded_key, 0) > 0
                 if entry is not None and locked and not force:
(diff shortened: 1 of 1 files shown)


lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -1234,7 +1231,8 @@ def _apply_loaded_state(self, data: dict[str, Any]) -> bool:
             return False
         if int(data.get("version", 0)) != 1:
             return False
-        if data.get("device_path") and data.get("device_path") != self.device_path:
+        checkpoint_device_path = data.get("device_path")
+        if checkpoint_device_path and checkpoint_device_path != self.device_path:
```
ApostaC, line 1235: I think this change is not necessary, but I'm okay with it.

ApostaC (approved):
Overall looks good to me. @DongDongJu Take another quick look please?

ApostaC:
Thanks for your contribution to LMCache as well!

DongDongJu (approved):
LGTM. Thanks for cleanup.

PR #3449, opened 2026-05-29
[Feat] raw_block: compress checkpoint payload to fit more entries and cut write amplification

Daejun:

The raw_block metadata checkpoint is a single full-index JSON snapshot, rewritten in full every cycle and mirrored across two fixed-size containers. Its size is dominated by repetition — the per-entry key prefix (`model@kv_rank`) and the JSON structural keys (`"offset"`, `"size"`, `"shape"`, ...) repeat for every entry. This causes two problems:

1. **Write amplification** — the whole snapshot is re-serialized and re-written every checkpoint, even for a small delta.
2. **Silent durability cliff** — once the JSON exceeds the container capacity, the checkpoint is skipped with only a warning, losing durability that cycle.

This PR optionally compresses the checkpoint payload with zlib. The on-device payload is framed with a leading magic tag (`b"LMCZ1\x00"`); recovery auto-detects the codec from the tag, so:

- the metadata header struct and `meta_version` are **unchanged**;
- an uncompressed payload is JSON and always starts with `{` (0x7B), which never collides with the tag, so **checkpoints written without compression (or by older versions) still load**;
- the CRC and the container-capacity check operate on the **stored (compressed)** bytes, so torn-write detection is preserved and more entries fit per container.

Controlled by `meta_checkpoint_compression` (`"none"` | `"zlib"`, default `"zlib"`), plumbed through the non-MP plugin (`rust_raw_block.meta_checkpoint_compression`) and the MP adapter config, and surfaced in `report_status`. zlib level 1 is used deliberately (see reviewer notes).

Compression runs on the **background checkpoint thread** and decompression runs once **at startup**, so the KV data path (put/get payload I/O) is unaffected.

**Measurements** (real raw-block device, object keys with random sha256 hashes):

_N = 20,000 entries:_

| mode | checkpoint payload | checkpoint write | recovery | recovered |
|------|-------------------:|-----------------:|---------:|----------:|
| none | 4332 KB            | 25.3 ms          | 200.5 ms | 20000     |
| zlib | 952 KB (**4.6x**)  | 38.8 ms          | 217.8 ms | 20000     |

- Recovery is dominated by `meta_verify_on_load` slot-header validation (O(N) reads); decompression is a small fraction of it (~18 ms).
- _Capacity_ (fixed ~28 KiB container payload): `none` persists ~100 entries, `zlib` persists ~550 (**~5.5x** more durable entries in the same container).
- _Cumulative write_ (10 checkpoints, index 2k→20k): `none` 23.3 MiB vs `zlib` 5.1 MiB written to the metadata region.

The ratio is bounded by the entropy floor — most bytes are random hash hex, which is incompressible; the gain comes from the repeated prefixes/structure.

- **Magic-prefix framing vs alternatives**: a header `flags` field would require changing `_META_HEADER_STRUCT` (and the hand-built test headers); bumping `meta_version` would collide with its config-knob / layout-mismatch semantics. The payload magic tag avoids both and gives automatic backward-compatible reads.
- **Level 1, not the zlib default (6)**: the payload is mostly incompressible random hash hex, so the repeated prefixes/JSON structure are already captured at level 1. Level 6 yields only ~7% smaller output for roughly double the compress CPU — a poor trade for a background, large-payload task.
- **Validation helper**: `validate_checkpoint_compression()` validates and narrows the codec to `Literal["none","zlib"]`, mirroring `normalize_raw_block_io_engine`. Invalid values are rejected eagerly in `from_dict`/`__init__`, not lazily.
- **Forward compatibility**: a compressed checkpoint cannot be read by a version predating this knob; operators set `meta_checkpoint_compression="none"` before downgrading. Documented in the design doc + user docs. The magic tag is versioned (`LMCZ1`) so future codecs can be added without breaking compatibility.
- **Test private access**: the new core tests touch a few private methods (`_select_latest_checkpoint`, `_write_checkpoint`, `_decode_checkpoint_payload`) to assert on-device byte framing and the CRC-valid-but-undecompressable path, for which there is no public surface. This matches existing patterns in that test module.

**If applicable**:

### DIFF
--- a/docs/design/v1/distributed/l2_adapters/raw_block.md
+++ b/docs/design/v1/distributed/l2_adapters/raw_block.md
@@ -96,10 +96,20 @@ Rules:
 - periodic checkpointing
 - optional checkpoint load on startup
 - optional verification on load
+- optional checkpoint payload compression (`meta_checkpoint_compression`)
 - recovery by loading the latest durable checkpoint and rebuilding the in-memory
   index
 
-The on-device format is intentionally unchanged by the MP adapter work.
+The checkpoint payload may optionally be compressed with zlib. When enabled, the
+on-device payload is framed with a leading magic tag (`b"LMCZ1\x00"`) so recovery
+auto-detects the codec without changing the metadata header struct or
+`meta_version`. An uncompressed payload is JSON and always starts with `{`, which
+never collides with the tag, so checkpoints written without compression (or by
+older versions) still load. The CRC and the container-capacity check operate on
+the stored (compressed) bytes, so torn-write detection is preserved and more
+entries fit per metadata container while fewer bytes are written each cycle.
+
+The on-device slot/header format is otherwise unchanged by the MP adapter work.
 
 Recovered keys are exposed to the shared L2 eviction policy on adapter startup,
 so reclaimed slots come from global L2 eviction or explicit `delete()` calls.
@@ -124,6 +134,7 @@ The MP adapter is configured through `--l2-adapter` JSON:
   "meta_enable_periodic": true,
   "load_checkpoint_on_init": true,
   "meta_verify_on_load": true,
+  "meta_checkpoint_compression": "zlib",
   "num_store_workers": 2,
   "num_lookup_workers": 1,
   "num_load_workers": 4
@@ -140,6 +151,9 @@ Important validation rules:
   of loading the latest on-device metadata checkpoint
 - with `use_odirect=true`, MP L1 alignment must satisfy
   `l1_align_bytes >= block_align`
+- `meta_checkpoint_compression` must be `"none"` or `"zlib"` (default `"zlib"`).
+  A compressed checkpoint cannot be read by a version that predates this knob;
+  set `"none"` before downgrading.
 
 ## Relationship to Non-MP Mode
 
--- a/docs/source/mp/l2_storage/raw_block.rst
+++ b/docs/source/mp/l2_storage/raw_block.rst
@@ -23,6 +23,13 @@ caller-provided load buffers during prefetch.
 - ``meta_checkpoint_interval_sec`` / ``meta_idle_quiet_ms`` /
   ``meta_enable_periodic`` / ``meta_verify_on_load``: Checkpoint and recovery
   controls carried over from the legacy raw-block backend.
+- ``meta_checkpoint_compression``: Checkpoint payload codec, ``"none"`` or
+  ``"zlib"`` (default ``"zlib"``). Compression shrinks the checkpoint so more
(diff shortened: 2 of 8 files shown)


ApostaC:
ping @DongDongJu

DongDongJu (commented):
Hello @Daejun,
Thanks for the detailed PR and measurements.
Can we get the actual WAF number increased by metadata? 
I want to understand that how bad it is since we are writing the metadata not that aggressively.

Daejun7Park:
@DongDongJu Yeah, it does cut the metadata write bandwidth — here are the actual numbers. Checkpoints are full-snapshot, ping-pong to one mirror, once per `meta_checkpoint_interval_sec` (default 60s) when the index is dirty. At the measured ~223 B/entry uncompressed:

| index size | snapshot / checkpoint | write rate | per day | per year |
|---|---|---|---|---|
| 10k | 2.2 MB | 0.037 MB/s | 3.2 GB | 1.2 TB |
| 100k | 22.3 MB | 0.37 MB/s | 32 GB | 11.7 TB |
| 1M | 223 MB | 3.7 MB/s | 320 GB | 117 TB |

zlib gets about 5x on that, so e.g. 100k drops from ~11.7 to ~2.3 TB/yr. You're right that we don't checkpoint aggressively, so the per-second rate is small — this is a cheap, modest win rather than a big one, and it costs nothing on the data path (compress runs on the background checkpoint thread).

A couple of smaller things it also helps with:

- Fits more entries before the durability cliff. `_write_checkpoint` skips with just a warning when the payload exceeds the container cap (`meta_total_bytes/2 - block_align`, ~128 MiB at the 256 MiB default). At ~223 B/entry that's ~600k entries; with ~5x it's ~3M before checkpoints start silently failing.
- Lower total bytes written over the device's life, which matters more on endurance-limited disks since every cycle rewrites the full snapshot regardless of how little changed.

The full-rewrite-every-cycle ratio (N/d) is the bigger structural issue, but that's a separate incremental-checkpoint change; compression is just an independent ~5x that stacks with it.

DongDongJu (approved):
Hello @Daejun, Please resolve the conflict.
Otherwise, LGTM.

Daejun7Park:
@DongDongJu Resolved the conflict.

Daejun:
@ApostaC friendly ping 🙏 — small one (+405/-11), rebased onto latest `dev` and approved by @DongDongJu, with `k3-unit-tests` green. (`unit-tests-amd` is stuck in the current repo-wide AMD-runner backlog — pending on every open PR, not specific to this PR.) Would appreciate your final review when you get a chance. Thanks!

DongDongJu:

PR #3451, opened 2026-05-29
[MP][Bugfix] Keep shared layout descriptors until final unregister (#3448)

### LINKED ISSUE #3448: [Bug] LMCache MP lookup fails after unregistering one of multiple instances with same model/world_size
## Describe the bug

When one LMCache MP server serves multiple vLLM instances with the same
`model_name` and `world_size`, unregistering one instance can break lookup for
the remaining live instances.

The server stores GPU contexts per `instance_id`, but `layout_desc_registry` is
keyed only by `(model_name, world_size)`. On `UNREGISTER_KV_CACHE`,
`GPUTransferModule.unregister_kv_cache()` removes the instance-specific GPU
context and then unconditionally unregisters the shared (model_name,
world_size) layout descriptor.

As a result, `/status` can still show live `registered_gpu_ids`, but subsequent
lookup requests fail with:

```text
LMCache ERROR: No GPU context found for model /root/models/Qwen3-4B with world size 1 during lookup!
LMCache WARNING: Session ... not found, skipping touch
```

## To Reproduce

1. Start one LMCache MP server.

```bash
lmcache server \
  --host 127.0.0.1 \
  --port 5555 \
  --http-host 127.0.0.1 \
  --http-port 8080 \
  --l1-size-gb 4 \
  --eviction-policy LRU
```

2. Start two vLLM instances using the same model and the same LMCache MP server.

```bash
CUDA_VISIBLE_DEVICES=0 vllm serve /root/models/Qwen3-4B \
  --port 8000 \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both","kv_connector_extra_config":{"lmcache.mp.host":"tcp://127.0.0.1","lmcache.mp.port":5555}}'
```

Start another equivalent vLLM instance with the same model/world size.

3. Stop one of the vLLM instances.

4. Send a request to the remaining live vLLM instance.

5. Check LMCache server logs. Lookup logs the error:

```text
No GPU context found for model /root/models/Qwen3-4B with world size 1 during lookup!
```

6. Check LMCache status:

```bash
curl -s http://127.0.0.1:8080/status
```

`registered_gpu_ids` can still contain the remaining live instance, while lookup
fails because the shared layout registry entry was removed.

## Expected behavior

Unregistering one vLLM instance should not remove the layout descriptor needed
by other live instances with the same `(model_name, world_size)`.

If multiple live instances share the same `(model_name, world_size)`,
`layout_desc_registry` should remain valid until the last matching instance is
unregistered.

## Desktop

- OS: Ubuntu 24.04
- Python: 3.12
- LMCache: 0.4.6.dev74
- vLLM: 0.21.0
- PyTorch: 2.11.0
- CUDA package stack: cu13

## Additional context

Relevant code path:

```python
# register_kv_cache
self._gpu_contexts[instance_id] = GPUContextEntry(...)
self._ctx.layout_desc_registry.register(model_name, world_size, layout_desc)

# unregister_kv_cache
entry = self._gpu_contexts.pop(instance_id, None)
self._ctx.layout_desc_registry.unregister(entry.model_name, entry.world_size)
```

Because `layout_desc_registry` is keyed by `(model_name, world_size)`, this
unregister path removes a layout descriptor that may still be needed by other
registered instances.

A simple fix is to only unregister the `(model_name, world_size)` layout when
no remaining registered context uses the same pair.

If this approach looks reasonable, I can open a PR with the change.

Javen-Ke:

Fixes #3448.

When multiple MP worker instances register the same `(model_name, world_size)`, they share one layout descriptor entry in `LayoutDescRegistry`. Previously, unregistering any one instance removed that shared entry immediately, so remaining live instances could fail lookup with "No GPU context found".

This PR tracks the number of active registrations per `(model_name, world_size)` and only removes the layout descriptor when the final matching registration is unregistered.

The regression test uses mocked GPU context construction, so it exercises the MP GPU registration/unregistration path without requiring CUDA.

Verified with:
- `pytest -q tests/v1/multiprocess/test_gpu_transfer_layout_registry.py`
- Manual MP integration test with one LMCache server and two vLLM instances sharing the same model/world size. After stopping one instance, requests to the remaining instance continued normally and the original lookup error did not reappear.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/engine_context.py
+++ b/lmcache/v1/multiprocess/engine_context.py
@@ -2,6 +2,7 @@
 """Shared context and layout descriptor registry for engine modules."""
 
 # Standard
+from dataclasses import dataclass
 import threading
 
 # First Party
@@ -21,17 +22,27 @@ from lmcache.v1.multiprocess.token_hasher import TokenHasher
 logger = init_logger(__name__)
 
 
+@dataclass
+class _LayoutDescEntry:
+    """Stored layout descriptor and its active registration count."""
+
+    layout_desc: MemoryLayoutDesc
+    ref_count: int
+
+
 class LayoutDescRegistry:
     """Thread-safe registry mapping (model_name, world_size) to MemoryLayoutDesc.
 
     Modules write to this registry when KV caches are registered.
     Consumers (e.g. LookupModule) read from it to find layout descriptors
-    for prefetch tasks.
+    for prefetch tasks. Multiple worker instances can share the same
+    ``(model_name, world_size)`` entry, so the registry keeps the descriptor
+    until the last matching registration is unregistered.
     """
 
     def __init__(self) -> None:
-        # Key: (model_name, world_size) -> MemoryLayoutDesc
-        self._registry: dict[tuple[str, int], MemoryLayoutDesc] = {}
+        # Key: (model_name, world_size) -> layout descriptor entry
+        self._registry: dict[tuple[str, int], _LayoutDescEntry] = {}
         self._lock = threading.Lock()
 
     def register(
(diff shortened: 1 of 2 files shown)


ApostaC (approved):
Thanks for contributing to LMCache! This is a clear fix!

deng451e (approved):
LGTM

sammshen (approved):
LGTM, the behavior on overwriting seems a bit off (since we will inherit the ref count of the previous layout desc). should we overwrite the ref count back to 1?

sammshen:
e.g. world size = TP * DP and we switch the TP and DP so we get different layout desc for same world_size. anyways, this bug already exists in the code so let's merge first

PR #3452, opened 2026-05-29
[MP][Bugfix] Track layout descriptor ownership across registrations

### LINKED ISSUE #3448: [Bug] LMCache MP lookup fails after unregistering one of multiple instances with same model/world_size

vmanvs:

Fixes #3448. Related to #3451.

When multiple vLLM worker instances share the same `(model_name, world_size)` on
one LMCache MP server, unregistering any single instance could remove the layout
descriptor still needed by remaining live instances. A later lookup could then fail
with:

```text
LMCache ERROR: No GPU context found for model ... with world size 1 during lookup!
```

## Root cause

`LayoutDescRegistry` stored a flat mapping:

```text
(model_name, world_size) -> MemoryLayoutDesc
```

That meant `unregister(model_name, world_size)` removed the descriptor for the
entire model/world pair, regardless of which instance owned it or whether other
instances still depended on it.

The same ownership problem applied to GPU, non-GPU, and blend registrations.

## How this PR fixes it

This PR moves ownership tracking into `LayoutDescRegistry` itself.

The registry now stores descriptors as:

```text
(model_name, world_size) -> { (source, instance_id) -> MemoryLayoutDesc }
```

Each producer registers with an explicit owner:

- `source="gpu"` for `GPUTransferModule`
- `source="non_gpu"` for `NonGPUTransferModule`
- `source="blend"` for `BlendModule`

Unregistering one owner now removes only that owner’s descriptor. The descriptor
remains discoverable through `find(model_name, world_size)` as long as any other
owner still exists for that model/world pair.

When multiple sources exist for the same key, `find()` prefers normal transfer
layouts over blend layouts:

```text
gpu / non_gpu > blend > unknown
```

This preserves generic lookup behavior while still allowing blend to maintain its
own registration.

## Why this differs from #3451

#3451 appears to address the unregister path in the GPU transfer module. This PR
fixes the issue at the registry level instead, so the same ownership behavior is
shared by GPU, non-GPU, and blend registrations.

It also avoids blend descriptors shadowing normal transfer descriptors in generic
lookup.

## Additional test fixture cleanup

This PR also adjusts the `no_shared_allocator` pytest fixture path so the marker
is checked before constructing the session-scoped 5GB shared allocator.

Previously, tests marked with `@pytest.mark.no_shared_allocator` could still cause
the shared allocator fixture to be constructed because `memory_allocator` was an
eager fixture dependency of the autouse fixture.

This keeps the default shared-allocator behavior unchanged for unmarked tests, but
makes the opt-out marker work as documented for allocator-free unit tests.

## Does this PR introduce any user-facing change?

No. This is an internal MP bugfix.

`find(model_name, world_size)` is unchanged. `register()` and `unregister()` accept
new keyword-only ownership parameters with backward-compatible defaults.

## How was this tested?

```bash
python -m pytest -q tests/v1/multiprocess/test_layout_desc_registry.py
python -m pytest -q tests/v1/multiprocess/test_non_cuda_data_transfer.py::test_unregister_non_cuda_context_preserves_sibling_layout
python -m pytest -q tests/v1/multiprocess/test_blend_server_v2.py::test_report_status_unregister_clears_cb_fields
```

**If applicable**:

- [X] this PR contains unit tests

### DIFF
--- a/lmcache/v1/multiprocess/layout_desc_registry.py
+++ b/lmcache/v1/multiprocess/layout_desc_registry.py
@@ -1,0 +1,94 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Registry for memory layout descriptors used by multiprocess lookup."""
+
+# Standard
+import threading
+
+# First Party
+from lmcache.v1.distributed.api import MemoryLayoutDesc
+
+
+class LayoutDescRegistry:
+    """Thread-safe registry mapping model/world pairs to layout descriptors.
+
+    Modules write to this registry when KV caches are registered. Consumers
+    such as ``LookupModule`` read from it to find layout descriptors for
+    prefetch tasks.
+    """
+
+    def __init__(self) -> None:
+        # Key: (model_name, world_size) -> owner -> MemoryLayoutDesc.
+        # Multiple workers can register the same model/world pair, so
+        # ownership has to survive individual unregister calls.
+        self._registry: dict[
+            tuple[str, int],
+            dict[int | None, MemoryLayoutDesc],
+        ] = {}
+        self._lock = threading.Lock()
+
+    def register(
+        self,
+        model_name: str,
+        world_size: int,
+        layout_desc: MemoryLayoutDesc,
+        *,
+        instance_id: int | None = None,
+    ) -> None:
+        """Register a layout descriptor for a model/world pair.
+
+        Args:
+            model_name: The model name.
+            world_size: The world size.
+            layout_desc: The memory layout descriptor.
+            instance_id: Optional worker instance identifier owning the layout.
+        """
+        with self._lock:
+            owners = self._registry.setdefault((model_name, world_size), {})
+            owners[instance_id] = layout_desc
+
+    def unregister(
+        self,
+        model_name: str,
+        world_size: int,
+        *,
+        instance_id: int | None = None,
+    ) -> None:
+        """Remove an owned layout descriptor for a model/world pair.
+
+        Args:
+            model_name: The model name.
+            world_size: The world size.
+            instance_id: Optional worker instance identifier used during
+                ``register``.
+        """
+        with self._lock:
+            key = (model_name, world_size)
+            owners = self._registry.get(key)
+            if owners is None:
+                return
+
+            owners.pop(instance_id, None)
+            if not owners:
+                self._registry.pop(key, None)
+
+    def find(self, model_name: str, world_size: int) -> MemoryLayoutDesc | None:
+        """Look up a layout descriptor by model/world pair.
+
+        Args:
+            model_name: The model name.
+            world_size: The world size.
+
+        Returns:
+            The layout descriptor if found, otherwise None.
+
+        Note:
+            Multiple owners for the same model/world pair are expected to have
+            compatible layout descriptors. When several owners exist, this
+            method returns the first live descriptor registered for the key.
+        """
+        with self._lock:
+            owners = self._registry.get((model_name, world_size))
+            if not owners:
+                return None
+
+            return next(iter(owners.values()))
--- a/tests/v1/multiprocess/test_blend_server_v2.py
+++ b/tests/v1/multiprocess/test_blend_server_v2.py
@@ -67,6 +67,7 @@ SERVER_PORT = (
 SERVER_URL = f"tcp://{SERVER_HOST}:{SERVER_PORT}"
 CHUNK_SIZE = 256
 CPU_BUFFER_SIZE = 5.0
+IN_PROCESS_CPU_BUFFER_SIZE = 64 * 1024**2
 DEFAULT_TIMEOUT = 10.0
 
 
@@ -2026,7 +2027,7 @@ def in_process_blend_engine() -> Generator[tuple, None, None]:
     config = StorageManagerConfig(
         l1_manager_config=L1ManagerConfig(
             memory_config=L1MemoryManagerConfig(
-                size_in_bytes=int(CPU_BUFFER_SIZE * 1024**3),
+                size_in_bytes=IN_PROCESS_CPU_BUFFER_SIZE,
                 use_lazy=True,
             ),
         ),
@@ -2066,6 +2067,7 @@ def local_kv_cache_unwrap(monkeypatch, cb_client_context: CBClientContext):
 @pytest.mark.skipif(
     not torch.cuda.is_available(), reason="report_status CB tests require CUDA"
 )
+@pytest.mark.no_shared_allocator
 def test_report_status_no_cb_registrations(
     in_process_blend_engine: tuple,
 ):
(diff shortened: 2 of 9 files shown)


lmcache/v1/multiprocess/layout_desc_registry.py
```diff
@@ -0,0 +1,114 @@
...
+"""Registry for memory layout descriptors used by multiprocess lookup."""
+
+# Standard
+from typing import Literal
+import threading
+
+# First Party
+from lmcache.v1.distributed.api import MemoryLayoutDesc
+
+RegistrySource = Literal["gpu", "non_gpu", "blend", "unknown"]
```
ApostaC: I don't think we need source. The layout desc is an attribute of model's KV cache, which should not relate to the module (gpu/non-gpu/blend).

lmcache/v1/multiprocess/layout_desc_registry.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        with self._lock:
+            owners = self._registry.get((model_name, world_size))
+            if not owners:
+                return None
+
+            _owner, layout_desc = min(
+                owners.items(),
+                key=lambda item: self._SOURCE_PRIORITY.get(item[0][0], 2),
+            )
```
ApostaC: We can directly return the first one here.

tests/v1/multiprocess/test_blend_server_v2.py
ApostaC, line 1: I didn't quite get why we make changes in this file. Is it related to this PR?

tests/v1/multiprocess/test_non_cuda_data_transfer.py
```diff
@@ -386,6 +386,49 @@ def test_server_register_and_find_non_cuda_context_layout(
     assert layout.shapes[0] == torch.Size([2, 2, 16, 16])
 
 
+@pytest.mark.no_shared_allocator
```
ApostaC, line 389: btw, why do we need `@pytest.mark.no_shared_allocator` here?

I don't think we use this in MP mode unit tests.

ApostaC (changes requested):
Thanks for your interest in LMCache! Here are my comments.
I will also take a look at #3451

lmcache/v1/multiprocess/layout_desc_registry.py
vmanvs: That makes sense. I used `source` to prevent blend and transfer descriptors from shadowing each other, but if `MemoryLayoutDesc` is expected to be an invariant of the model KV cache for a given (`model_name`, `world_size`), then module source is over-modeling. I can simplify the registry to track ownership by `instance_id` only.

lmcache/v1/multiprocess/layout_desc_registry.py
vmanvs: Agreed. This priority logic only exists because I introduced `source`. If we remove `source`, then all owners under the same `(model_name, world_size)` key should have compatible layout descriptors, so returning the first one is enough. I’ll simplify this.

tests/v1/multiprocess/test_blend_server_v2.py
vmanvs: I added this because `BlendModule` is also a producer of layout descriptors. Once the registry tracks owners, each producer should clean up its own owner entry. The test verifies that `cb_unregister_kv_cache()` removes the blend-owned descriptor instead of leaving stale registry state. I can make the PR description clearer here.

tests/v1/multiprocess/test_non_cuda_data_transfer.py
vmanvs: Correct, this MP test does not use the shared allocator. I marked it with `no_shared_allocator` precisely to avoid constructing the global 5GB allocator for a test that does not need it. The accompanying conftest.py change makes that marker work as documented by delaying `memory_allocator` construction until after the marker check.

PR #3465, opened 2026-05-30
Reuse rank0 broadcast buffer for save_only_first_rank retrieve

milesial:
## Summary

When `save_only_first_rank=True`, rank 0 already stages each retrieved memory object onto CUDA before broadcasting it to the other ranks. The receiver ranks wrap their CUDA receive buffers as temporary memory objects, but rank 0 kept using the original retrieved memory object for `batched_to_gpu()`.

If rank 0's original object is pinned CPU memory, this makes rank 0 run the paged KV load from host memory while the other ranks load from CUDA memory. This change keeps the original rank-0 memory object in `reordered_chunks` for existing cleanup/refcount behavior, but uses a temporary `TensorMemoryObj` backed by the CUDA broadcast buffer for rank 0's `batched_to_gpu()` path.

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -828,10 +828,11 @@ class LMCacheEngine:
                         **kwargs,
                     )
 
+        chunks_to_load = reordered_chunks
         if self.save_only_first_rank:
             with retrieve_stats.profile_broadcast():
                 with torch_dev.stream(self.broadcast_stream):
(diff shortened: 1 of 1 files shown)


maobaolong:
@milesial Thanks for your contribution first, but it seems duplicated to https://github.com/LMCache/LMCache/pull/3413 ?

milesial:
Ah yes thanks, I missed that one

PR #2474, opened 2026-01-22
[CI]: Correctness test specify attn backend with CLI flag

sammshen:
It looks like the `VLLM_ATTENTION_BACKEND` environment has been deprecated in the nightly builds (which the correctness suite uses).

PR #2439, opened 2026-01-19
Check failed put task count and record metrics

maobaolong:
**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #3470, opened 2026-05-31
fix(token_database): return single segment when input is shorter than separator

SuperMarioYL:

`SegmentTokenDatabase._fast_split_by_subtensor` guards the cases where the input cannot contain a separator — an empty separator (`sep_len == 0`) or a sequence shorter than the separator — by yielding the whole sequence as a single segment:

```python
if self.sep_len == 0 or len(tokens) < self.sep_len:
    yield tokens          # <-- missing return
# ... falls through ...
windows = tokens.unfold(0, self.sep_len, 1)
```

The guard was **missing a `return`**, so execution falls through to `tensor.unfold(0, sep_len, 1)`, which raises `RuntimeError: maximum size for tensor at dimension 0 is N but size is M` whenever the window is wider than the tensor.

Concretely:

| Input | Current behavior | Expected |
|-------|------------------|----------|
| sequence shorter than the separator (e.g. a 1-token request) | yields the chunk, then **crashes with `RuntimeError`** on the next iteration | one segment for the whole sequence |
| empty token sequence | **crashes with `RuntimeError`** | one (empty) segment |
| empty separator (`sep_len == 0`) | every sliding window matches vacuously → input **explodes into spurious single-token / empty segments** | one segment for the whole sequence |

So any blending/segment request whose token length is shorter than the configured separator crashes inside `process_tokens`. `sep_len == 0` can also occur when `blend_special_str` tokenizes to a single token (the leading token is dropped via `[1:]`).

The fix adds the missing `return` so the guard yields exactly one segment, honoring the guard's original intent. The common path (sequence longer than the separator) is untouched.

The only existing `SegmentTokenDatabase` test is `@pytest.mark.skipif`-gated on Hugging Face credentials (it downloads `facebook/opt-125m`), so this code path had no CI coverage. The added regression tests drive the **public** `process_tokens` interface and stub the tokenizer, so they run in CI without any network/model download. They are red on `dev` (the short-input and empty-separator cases) and green with this fix.

**If applicable**:

### DIFF
--- a/lmcache/v1/token_database.py
+++ b/lmcache/v1/token_database.py
@@ -467,10 +467,30 @@ class SegmentTokenDatabase(TokenDatabase):
         self.sep_len = len(self.sep_tokens)
 
     def _fast_split_by_subtensor(self, tokens: torch.Tensor) -> Iterable[torch.Tensor]:
-        """Match the `sep_tokens` with sliding windows"""
+        """Split ``tokens`` on every occurrence of ``self.sep_tokens``.
 
+        The separator is matched with sliding windows of width ``self.sep_len``.
+        The tokens between consecutive separators (and the leading/trailing
+        remainders) are yielded as the resulting segments; the separator tokens
+        themselves are dropped.
+
+        :param torch.Tensor tokens: The 1-D token id tensor to split.
+
+        :returns: A generator of token segments. When the sequence cannot
+            contain a separator -- either because the separator is empty
+            (``sep_len == 0``) or because the sequence is shorter than the
+            separator (``len(tokens) < sep_len``) -- the whole sequence is
+            yielded unchanged as a single segment.
+        """
+
+        # A sequence that is shorter than the separator (or an empty separator)
+        # cannot contain a match, so it forms a single segment. Returning here
+        # is required: ``tensor.unfold`` below raises a RuntimeError when the
+        # window is wider than the tensor, and an empty separator would make
+        # every window match vacuously.
         if self.sep_len == 0 or len(tokens) < self.sep_len:
             yield tokens
+            return
 
         # Unfold into sliding windows
         # shape: (num_tokens-sep_len+1, sep_len)
--- a/tests/v1/test_token_database.py
+++ b/tests/v1/test_token_database.py
@@ -8,6 +8,7 @@ import pytest
 import torch
 
 # First Party
+from lmcache.v1 import token_database
 from lmcache.v1.config import LMCacheEngineConfig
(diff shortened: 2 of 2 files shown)


ApostaC (approved):
LGTM. Although this module will be deprecated soon.

ApostaC:
@YaoJiayi can you take another quick look?

YaoJiayi (approved):

SuperMarioYL:
Gentle ping on this one — it's been approved (thanks @YaoJiayi!) and sitting green for a couple of weeks. Anything outstanding on my end before it can land? Happy to rebase onto the latest dev or squash if that helps.

SuperMarioYL:
Following up — this one is approved (thanks @ApostaC, @YaoJiayi) but still blocked by two red checks: `buildkite/k3-comprehensive-test` (build #2474) and `buildkite/k3-multiprocess-test` (build #2439). Both ran on 2026-06-03 against a base that is now 273 commits behind `dev`.

I looked into whether they are related to this change:

- Both suites pass on recent `full`-labelled PRs (#4058, #4044, #4026), so they are not broken repo-wide.
- This diff only adds a missing `return` to `SegmentTokenDatabase._fast_split_by_subtensor` (the CacheBlend segment path), plus unit tests. Neither of those two suites exercises that path.
- `tests/v1/test_token_database.py::test_segment_token_database` fails identically on a clean `dev` checkout with no changes applied, so that pre-existing failure is not from this PR either.

Could you re-run the two buildkite suites against current `dev`? If a fresh branch is easier, I am happy to rebase and force-push — just say the word.

Separately, noting @ApostaC's comment that this module is slated for deprecation: if it is going away soon and this is not worth landing, I am glad to close it. Either way works.

SuperMarioYL:
Gentle ping @YaoJiayi — this was approved, but several checks are still pending (likely the self-hosted runner). Happy to rebase or address feedback if anything is blocking the merge.

SuperMarioYL:
@ApostaC @YaoJiayi gentle ping — this has two approvals and the earlier CI failures have cleared (16 passing, rest pending). Noting the comment that this module will be deprecated soon — is it still worth merging, or would you prefer I close it out? Happy to rebase either way.

SuperMarioYL:
Closing this — with the module heading toward deprecation, I don't want to keep taking review bandwidth for a fix that may not ship. The branch stays on my fork; if token_database survives the deprecation and this short-input segmentation fix is still wanted, I'm happy to reopen or re-spin. Thanks for the reviews and for the heads-up on the deprecation!

PR #3477, opened 2026-06-01
[MP] [Metrics] Add l2_evicted_object, Add cachesalt to existing L1/L2 metrics 

aeon-x:

  Adds per-tenant `cache_salt` attribution to L1 and L2 OTel metrics so operators can slice cache usage on dashboards by tenant / isolation domain. Also
 Adds L2 eviction metrics (`l2_evicted_objects`), which were previously missing.

  **Changes:**

  - **Shared helpers** (`utils.py`): `group_by_salt(keys)` groups keys by `ObjectKey.cache_salt` into a `dict[str, int]`. `emit_salt_counts(counter,
  salt_counts)` emits one counter increment per distinct salt. Empty salt produces dimensionless increments (backward-compatible). Both L1 and L2
  subscribers use the same two functions.

  - **L1 subscriber**: All 3 counters (`l1_read`, `l1_write`, `l1_evicted`) now carry `cache_salt` attribute. L1 events already pass key lists, so the
  subscriber calls `group_by_salt` to build the salt→count dict on the drain thread.

  - **L2 emit sites**: Store, prefetch, and eviction controllers add a `key_count_per_salt` field (`Counter(k.cache_salt for k in keys)`) to event
  metadata alongside existing `key_count`/`succeeded_count`/etc. This pre-groups keys at the emit site (caller thread) so the drain thread stays O(T) per
  event instead of O(N).

  - **L2 subscriber**: 5 object counters tagged with `cache_salt` via `emit_salt_counts`: `l2_store_submitted_objects`, `l2_store_completed_objects`,
  `l2_prefetch_lookup_objects`, `l2_prefetch_load_submitted_objects`, `l2_prefetch_load_completed`. Request-level counters remain dimensionless.

  - **L2 eviction metrics** (new): `L2_KEYS_EVICTED` event published by `L2EvictionController._execute_eviction_action`. New
  `lmcache_mp.l2_evicted_objects` counter with `cache_salt` attribution in `L2MetricsSubscriber`. Logging subscriber also handles the new event.

  When ``cache_salt`` is missing or empty, the counter increments without a ``cache_salt`` attribute (dimensionless), so it behaves identically to a plain ``counter.add(len(keys))``.

  **If applicable**:


docs/design/v1/mp_observability/EVENTS.md
```diff
@@ -70,22 +70,30 @@ Producers:
...
+tenants, so per-tenant attribution requires grouping by salt.
 
 ---
 
 ## L2 Prefetch Controller Events
 
 | EventType | Metadata keys | Types |
 |---|---|---|
-| `L2_PREFETCH_LOOKUP_SUBMITTED` | `request_id`, `key_count`, `adapter_count` | `int`, `int`, `int` |
+| `L2_PREFETCH_LOOKUP_SUBMITTED` | `request_id`, `keys`, `adapter_count` | `int`, `list[ObjectKey]`, `int` |
```
ApostaC: Why do we change `key_count` to `keys`? This will greatly explode the size of the events and incur python-level overheads.

ApostaC (commented):
Just wondering, do we really need cache_salt for L1 metrics?

docs/design/v1/mp_observability/EVENTS.md
aeon-x: Need `keys` to group by the keys count in each cachesalt. 

The key list already exists at the emit site — we're just passing a reference, not copying it. So there shouldn't be any overhead now, since the Producers and Subscribers shares the same heap. 

But there is nothing wrong to keep size of events smaller, in case we want to serialize/copy them. I can try to group keys count by cachesalt before emitting.

aeon-x: Updated.

lmcache/v1/mp_observability/subscribers/metrics/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+def emit_salt_counts(counter: metrics.Counter, salt_counts: dict[str, int]) -> None:
+    """Add to *counter* once per entry in *salt_counts*.
+
+    *salt_counts* maps ``cache_salt`` values to key counts.
+    Empty salt produces a dimensionless increment (no attribute).
+    """
+    for salt, count in salt_counts.items():
+        attrs = {"cache_salt": salt} if salt else {}
+        counter.add(count, attributes=attrs)
```
royyhuang, line 28: Maybe I am wrong. Can counter have existing attributes that were shared and set when it is initialized, e.g., in subscribers' `__init__()`? If os, will this override the existing ones? Maybe worth to check this and change to appending instead of directly setting the value.

royyhuang (approved):
Otherwise LGTM.

lmcache/v1/mp_observability/subscribers/metrics/utils.py
aeon-x: I think in otel counter, it is ok to add directly. It will not override, but create new attribute set instead.

ApostaC (approved):
LGTM!

PR #3478, opened 2026-06-01
[Refactor] Consolidate ParallelStrategy construction in vllm_multi_process_adapter

Lyj1007:
## Summary

Pull the vLLM/LMCache-server matching logic out of the connector and into
the adapter, so the mapping lives in one place and is easier to extend
for future deployment topologies such as multi-node.

## Notes
- Part of the split suggested in #3248 (review). This PR contains **only** the `ParallelStrategy` refactor to prepare for multi-node .
- Pure refactor, no behavioral change.


maobaolong (commented):
@Lyj1007  Thank you for this PR. But I found that mla can be instantiated in some subclasses? In this way, there is no need to call is_mla in every API to judge, what do you think?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -224,32 +225,61 @@ class ParallelStrategy:
...
+            tp_size=pc.tensor_parallel_size,
+            pp_size=pc.pipeline_parallel_size,
+        )
+
+    @property
+    def kv_world_size(self) -> int:
+        """Number of pieces a single token chunk's KV cache is split into
+        on the LMCache server storage."""
+        if self.use_mla:
+            return self.vllm_world_size // self.tp_size
```
ApostaC, line 308: quick question: when we run dp = 8,  tp = 1, what will be the `vllm_world_size`?

Ideally, the `kv_world_size` for this case should be 1.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -498,19 +411,20 @@ def __init__(
...
         zmq_context = zmq.Context.instance()
+        parallel_strategy = ParallelStrategy.from_vllm_config(vllm_config)
+        common_adapter_kwargs: dict[str, Any] = dict(
+            server_url=server_url,
+            context=zmq_context,
+            model_name=vllm_config.model_config.model,
+            vllm_block_size=vllm_config.cache_config.block_size,
+            parallel_strategy=parallel_strategy,
+            extra_config=vllm_config.kv_transfer_config.kv_connector_extra_config,
+        )
```
ApostaC: Although this makes the code shorter, it's essentially forcing `LMCacheMPSchedulerAdapter` and `LMCacheMPWorkerAdapter` to have the same `__init__` signature. 
I would prefer to initialize them separately without having a `common_adapter_kwargs`. WDYT?

ApostaC (commented):
This one is a great refactor! Please see my comments below. Otherwise LGTM!

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -224,32 +225,61 @@ class ParallelStrategy:
...
+            The constructed ParallelStrategy.
+        """
+        pc = vllm_config.parallel_config
+        return cls(
+            use_mla=mla_enabled(vllm_config.model_config),
+            vllm_world_size=pc.world_size,
+            vllm_worker_id=pc.rank,
+            tp_size=pc.tensor_parallel_size,
+            pp_size=pc.pipeline_parallel_size,
+        )
```
ApostaC: Since this function uses `VllmConfig`, maybe move it to `lmcache_mp_connector.py` as a helper function?

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: In vLLM's `ParallelConfig` ([`vllm/config/parallel.py` L307](https://github.com/vllm-project/vllm/blob/main/vllm/config/parallel.py#L307)):

```python
        world_size: int = Field(init=False)
        """world_size is TPxPP, it affects the number of workers we create."""
```

DP is **not** included—each DP replica is a separate process with its own connector instance. vLLM  provides a distinct property for the DP-inclusive size([`vllm/config/parallel.py` L499](https://github.com/vllm-project/vllm/blob/main/vllm/config/parallel.py#L499)):
```python
    def world_size_across_dp(self) -> int:
        """world_size_across_dp is TPxPPxDP, it is the size of the world
        including data parallelism."""
        return self.world_size * self.data_parallel_size
```

So for `dp=8, tp=1, pp=1`:
- `pc.world_size = 1 × 1 = 1`
- `kv_world_size = 1 // 1 = 1` ✓

To avoid this ambiguity, I'm considering updating the field docstrings. Three options—which do you prefer?

**Option A** – trace back to the vLLM source field:
```python
vllm_world_size: int
"""From ``vllm.parallel_config.world_size``."""

vllm_worker_id: int
"""From ``vllm.parallel_config.rank``."""
```

**Option B** – emphasize the per-scheduler scope:
```python
vllm_world_size: int
"""Number of workers under one vLLM scheduler."""

vllm_worker_id: int
"""This worker's rank within its scheduler group."""
```

**Option C** – keep the current wording as-is:
```python
vllm_world_size: int
"""vLLM-side world size."""

vllm_worker_id: int
"""vLLM-side rank of the sub-process."""
```

A makes it easy to look up the upstream definition, B gives semantic context so readers won't wonder whether DP is folded in, C keeps it minimal. LMK!

Lyj1007:
> @Lyj1007 Thank you for this PR. But I found that mla can be instantiated in some subclasses? In this way, there is no need to call is_mla in every API to judge, what do you think?

Hi,@maobaolong Thanks for the suggestion! Subclassing to eliminate the repeated `if use_mla` checks makes sense.

I was thinking about how this interacts with #3248 (multi-server)—there's an additional `n_servers > 1` dimension in the dispatch logic that would still live in the adapter properties even after subclassing. So we'd end up with part of the dispatch in the subclass and part in the adapter functions.

I'm leaning toward keeping all dispatch logic in one place (adapter properties) with `ParallelStrategy` as a flat dataclass—but happy to go with subclassing if you think that's cleaner. What's your preference?

maobaolong:
@Lyj1007 OK, i'm fine to do not introduce the refactor work i motioned in this PR.

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Done—extracted `from_vllm_config` from `ParallelStrategy` into a standalone helper in `lmcache_mp_connector.py` (e30f54f). Thanks!

lmcache/integration/vllm/lmcache_mp_connector.py
Lyj1007: Agreed—restored separate initialization for the scheduler and worker adapters (7ebc247). Thanks for the suggestion!

ApostaC (approved):
LGTM!

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -711,10 +656,7 @@ def wait_for_save(self):
         """
         # In MLA scenario, only the first rank of the pipeline group
         # needs to save the KV cache.
-        if (
-            self.worker_adapter.use_mla
-            and not self.worker_adapter.is_first_rank_of_pp_group
-        ):
+        if not self.worker_adapter.is_kv_writer:
```
sammshen, line 659: this is great

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -285,33 +285,42 @@ class ParallelStrategy:
...
-    The kv worker id of the sub-process, kv_worker_id may not be equal to the 
-    actual_worker_id, in the case of mla, it will 'exclude' the effect of TP, 
-    the value is calculated by `extract_world_size_and_kv_rank` in 
-    `lmcache_mp_connector.py`.
-    """
-
-    actual_world_size: int
-    """The actual world size."""
+    vllm_world_size: int
+    """Number of workers under one vLLM scheduler."""
```
sammshen: adding to yihua's comment, let's make it explicit what falls under one vLLM scheduler (PP x TP)

lmcache/integration/vllm/vllm_multi_process_adapter.py
sammshen: Can we combine A and B and make it explicit that world size is TP x PP. I remember we had this explicitly in the old LMCache code

sammshen (changes requested):
LGTM! This refactor is nice. Just the one doc string comment

lmcache/integration/vllm/vllm_multi_process_adapter.py
Lyj1007: Thanks for the suggestion! I've updated the docstring to make it explicit:

```
vllm_world_size: int
"""Number of workers managed by one vLLM scheduler (TP × PP, excluding DP).

Mirrors ``vllm.parallel_config.world_size``.
"""
```
This should make it clear at a glance that DP is not included. PTAL and let me know if there's anything else to adjust!

sammshen (approved):
LGTM!

PR #3480, opened 2026-06-01
[Fix][RawBlock] Fix TOCTOU accounting bug in delete() by merging was_indexed into delete_many

3xdevv:

`RawBlockL2Adapter.delete()` had a TOCTOU race between reading metadata and deleting:

```python
# old — two separate operations, not atomic
metas = self._core.get_metadata_many(encoded_keys)    # step 1
deleted_bitmap = self._core.delete_many(encoded_keys) # step 2
```

Between step 1 and step 2, an inflight key could be committed to `_index`. When that happened, `meta` was `None` (key not yet indexed at step 1), so `deleted_sizes` got `0` instead of `slot_bytes`. `_notify_keys_deleted` then under-reported the freed size, leaving `total_bytes_used` permanently inflated by one slot.

The fix merges the "was it indexed?" question into `delete_many` itself. `RawBlockCore.delete_many` now returns `list[RawBlockDeleteResult]` — a frozen dataclass with `deleted` and `was_indexed` — both fields captured atomically under `_lock`. The adapter reads `result.was_indexed` instead of calling `get_metadata_many` separately.

- `RustRawBlockBackend.remove()` also calls `delete_many`; it only needs `.deleted`, updated on the same line.
- `was_indexed` is `True` iff `removed_entry is not None` — i.e., the key was in `_index` (not merely `_inflight`) when the lock was held during deletion. Inflight-only deletes cancel the write but never charged `slot_bytes` to usage in the first place, so size 0 is correct.

## Tests

```
pytest tests/v1/distributed/test_raw_block_l2_adapter.py \
  -k "test_raw_block_l2_adapter_delete_indexed_key_deducts_slot_bytes or test_raw_block_l2_adapter_delete_inflight_key_keeps_usage_at_zero" \
  -v
```

Two new tests:
- `test_raw_block_l2_adapter_delete_indexed_key_deducts_slot_bytes` — store then delete, verify usage returns to 0 (was_indexed=True path).
- `test_raw_block_l2_adapter_delete_inflight_key_keeps_usage_at_zero` — races `delete()` against a write paused at `_write_one` to reproduce the TOCTOU window and verify usage stays at 0 (was_indexed=False path).

## If applicable

- [ ] My changes require a documentation update
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
@@ -798,15 +798,16 @@ class RawBlockL2Adapter(L2AdapterInterface):
     def delete(self, keys: list[ObjectKey]) -> None:
         """Delete keys from raw-block L2 and notify listeners for removals."""
         encoded_keys = [encode_object_key(key).encoded for key in keys]
-        metas = self._core.get_metadata_many(encoded_keys)
-        deleted_bitmap = self._core.delete_many(encoded_keys, force=False)
+        delete_results = self._core.delete_many(encoded_keys, force=False)
         deleted_keys: list[ObjectKey] = []
         deleted_sizes: list[int] = []
-        for key, meta, deleted in zip(keys, metas, deleted_bitmap, strict=False):
-            if not deleted:
+        for key, result in zip(keys, delete_results, strict=False):
+            if not result.deleted:
                 continue
             deleted_keys.append(key)
-            deleted_sizes.append(0 if meta is None else int(self._core.slot_bytes))
+            deleted_sizes.append(
+                int(self._core.slot_bytes) if result.was_indexed else 0
+            )
         if deleted_keys:
             try:
                 self._notify_keys_deleted(deleted_keys, deleted_sizes)
--- a/lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
+++ b/lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
@@ -371,7 +371,7 @@ class RustRawBlockBackend(StoragePluginInterface):
     def remove(self, key: CacheEngineKey, force: bool = True) -> bool:
         spec = encode_legacy_key(key)
         with self._pin_lock:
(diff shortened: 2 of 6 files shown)


DongDongJu (changes requested):
I don’t think the new inflight test reproduces the race described in the PR. 
It cancels the inflight write before commit, so the old code would also keep usage at 0. 
Can we add a deterministic regression where delete starts while the key is inflight, then the store commits before the actual delete_many() call? 
That should fail on the old metadata-snapshot implementation and pass with was_indexed returned from delete_many().

One note here. it still has concurrency issue in put_many() but lets target later.

3xdevv:
Hello, @DongDongJu, thank you for the careful review!
I revised the test to reflect the points you mentioned. It is now only passed in changed, was_indexed version.
PTAL, thanks!

DongDongJu (approved):
LGTM

DongDongJu:
Hello @sammshen @ApostaC  Could you check this one by any chance?

lmcache/v1/distributed/l2_adapters/raw_block_l2_adapter.py
```diff
@@ -798,15 +798,16 @@ def query_load_result(self, task_id: L2TaskId) -> Bitmap | None:
     def delete(self, keys: list[ObjectKey]) -> None:
         """Delete keys from raw-block L2 and notify listeners for removals."""
         encoded_keys = [encode_object_key(key).encoded for key in keys]
-        metas = self._core.get_metadata_many(encoded_keys)
```
chunxiaozheng, line 801: @3xdevv Thanks for your fix, `get_metadata_many` is removed here, I noticed that there is no place to call this method anymore, could you help check again, if so, we can remove it.

nayeonikim:
Since this PR was opened, #3494 has merged into dev, adding RustRawBlockBackend.batched_remove(). It uses the boolean results from delete_many() to update pin bookkeeping and count removals.

With the new return type, if removed treats an object with deleted=False as truthy, and sum(results) raises TypeError. Could you update this caller to use result.deleted for both checks and counting, and verify the existing pinned/missing-key tests?

PR #3485, opened 2026-06-01
[Refactor]: break apart `utils.py` into format and engine based strategy pattern

sammshen:
Based on discussion in https://github.com/LMCache/LMCache/pull/3323 

**If applicable**:


sammshen:
this is not super urgent but we could push it in at any time

lmcache/v1/gpu_connector/kv_format/specs.py
```diff
@@ -0,0 +1,638 @@
...
+            f"{self.num_layers()} x [{self.page_buffer_size()}, 1, {self.head_size()}]"
+        )
+
+
+# ---------------------------------------------------------------------------
+# Registry: the single, flat mapping from enum -> spec class, derived from
+# each spec's own ``gpu_kv_format`` so identity lives in exactly one place.
+# ---------------------------------------------------------------------------
+_ALL_SPECS: list[type[KVFormatSpec]] = [
+    CrossLayerNhdSpec,
```
maobaolong: Putting all the spec implementations into a single file is a simple approach, but it causes the file to grow excessively large and leads to merge conflicts.

What if we split these spec files into separate files within the same folder? We could use dynamic scanning and loading to replace manual registration. This way, adding a new spec wouldn't require changing a single line of existing code, and any modifications to a spec would be contained within its own file.

WDYT?

lmcache/v1/gpu_connector/kv_format/detection.py
```diff
@@ -0,0 +1,411 @@
...
+        probe = probe[0]
+    tensor_dims = list(probe.shape)
+    dims_str = (
+        "".join(f"[{d}]" for d in list_dims) + f"[{', '.join(map(str, tensor_dims))}]"
+    )
+    logger.info("GPU KV Cache Dimensions: %s", dims_str)
+
+    if serving_engine == EngineType.TRTLLM:
+        detected_format = _detect_trtllm(list_depth, tensor_dim)
+    elif serving_engine == EngineType.VLLM:
```
maobaolong: Would it be better to use the Strategy pattern here? That way, we could replace the "if, else if" logic and implement a series of format detectors instead.

lmcache/v1/gpu_connector/kv_format/specs/base.py
```diff
@@ -0,0 +1,157 @@
...
+
+# Third Party
+import torch
+
+# First Party
+from lmcache.v1.gpu_connector.types import DiscoverableKVCache
+import lmcache.c_ops as lmc_ops
+
+
+class KVFormatSpec(ABC):
```
maobaolong, line 73: I like the base class interface you've abstracted.

lmcache/v1/gpu_connector/utils.py
```diff
@@ -296,51 +119,74 @@ def assert_layerwise_gpu_connector(gpu_connector: "GPUConnectorInterface"):
...
+    except ValueError:
+        return f"Unknown ({gpu_kv_format})"
+
+
+# Representative, diagnostic-only labels. ``GPUKVFormat`` is engine- and
+# backend-agnostic geometry; these strings name *a* backend that produces
+# each format today, for human-readable logging only. A real engine/backend
+# abstraction (if ever needed) would be a higher-level type that owns a
+# ``GPUKVFormat``, never a field on the geometry layer.
+_ATTENTION_BACKEND_LABELS: dict["lmc_ops.GPUKVFormat", str] = {
```
maobaolong: Could we put these KV-format labels into the specs implementation by adding a method like `get_label`?

lmcache/v1/gpu_connector/utils.py
```diff
@@ -34,204 +35,40 @@
...
-            )
-        inner_tight *= shape[i]
-    # Now ``inner_tight == prod(shape[1:])``; dim-0 must be >= that.
-    if stride[0] < inner_tight:
-        _fail(
-            f"dim-0 stride {stride[0]} < prod(shape[1:])={inner_tight} "
-            "(overlapping blocks)"
-        )
-    return stride[0] - inner_tight
+# Re-exported so existing ``from ...utils import DiscoverableKVCache,
```
maobaolong: Without the following `__all__` array, we can also `from lmcache.v1.gpu_connector.utils import DiscoverableKVCache`?

maobaolong (commented):
@sammshen Thanks for this great work, it make code clear and easy maintained and easy extended. It is really better than my version. 

Left some suggestions inline, please take a look and happy for discuss this anytime.

lmcache/v1/gpu_connector/kv_format/specs.py
sammshen: agree with the suggestion, I will add `lmcache/v1/gpu_connector/kv_format/specs/`

lmcache/v1/gpu_connector/kv_format/detection.py
sammshen: sure, detection by engine and format by layout makes sense to me

lmcache/v1/gpu_connector/utils.py
maobaolong: @sammshen How about this comment?

lmcache/v1/gpu_connector/kv_format/detectors/vllm.py
```diff
@@ -0,0 +1,55 @@
...
+from lmcache.v1.gpu_connector.kv_format.types import DiscoverableKVCache, LayoutHints
+import lmcache.c_ops as lmc_ops
+
+logger = init_logger(__name__)
+
+
+class VLLM_Detector(EngineDetector):
+    engine_type = EngineType.VLLM
+
+    def detect(
```
maobaolong: This detect method looks still use `if-elseif` pattern here, is it possible to define detect method within each  `EngineKVFormat` ? Thus, we will not need to modify the files out of a new defined `EngineKVFormat` file.

lmcache/v1/gpu_connector/utils.py
sammshen: agree, let me add this to `KVFormatSpec`

lmcache/v1/gpu_connector/kv_format/specs/base.py
```diff
@@ -0,0 +1,236 @@
...
+            rest record other known producers (the relationship is one format
+            to many backends).
+    """
+
+    engine_kv_format: ClassVar["lmc_ops.EngineKVFormat"]
+    is_mla: ClassVar[bool] = False
+    is_hnd: ClassVar[bool] = False
+    is_cross_layer: ClassVar[bool] = False
+    shape_desc: ClassVar[str]
+    attention_backends: ClassVar[tuple[str, ...]] = ()
```
sammshen, line 107: @maobaolong added new attribute

lmcache/v1/gpu_connector/kv_format/detectors/vllm.py
sammshen: This instinct seems correct, especially given the adoption of the Strategy pattern in the remainder of the PR. 

However, this detection happens _before_ figuring out the `EngineKVFormat` (that is what this function is doing). this means to implement strategy pattern we would have to do something like `detected_by: EngineType` and then a `matches(list_depth, tensor_dim, hints) -> bool` function, which leaks the Serving Engine back into the format

sammshen: additionally, some of the logic here really is vllm specific (like the `is_hnd` cpu safegaurd)

lmcache/v1/gpu_connector/kv_format/detectors/registry.py
```diff
@@ -0,0 +1,43 @@
...
+from pathlib import Path
+import importlib
+import pkgutil
+
+# First Party
+from lmcache.utils import EngineType
+from lmcache.v1.gpu_connector.kv_format.detectors.base import EngineDetector
+
+
+def _discover_detectors() -> dict[EngineType, type[EngineDetector]]:
```
sammshen, line 20: @maobaolong this is similar to what you did in the old in-process mode for storage connectors

lmcache/v1/gpu_connector/utils.py
```diff
@@ -810,343 +251,109 @@ def get_block_size(
...
-        # SGLang MHA 4-D inner (MP path): block_size at shape[1].
-        return kv_caches[0][0].shape[1]
-    elif engine_kv_format == lmc_ops.EngineKVFormat.NL_X_NBBS_ONE_HS:
-        raise ValueError(_ATTRIBUTE_NOT_EXIST_ERROR.format(format=engine_kv_format))
-    else:
-        raise ValueError(f"Unknown GPU KV Format: {engine_kv_format}")
+    return get_spec(kv_caches, engine_kv_format).block_size(layer_idx)
 
 
 def get_page_buffer_size(
```
ApostaC, line 266: Can we have a deprecation warning (there is an annotation called `lmcache_deprecate`) here because this is only used by non-MP mode.

lmcache/v1/gpu_connector/kv_format/specs/base.py
ApostaC, line 1: There are a lot of functions defined in the base class. It would be great if we could clarify which functions are used in MP mode, and which are used (only) in non-MP mode.

lmcache/v1/gpu_connector/kv_format/specs/base.py
```diff
@@ -0,0 +1,142 @@
...
+}
+
+
+def _render_shape(fmt: "lmc_ops.EngineKVFormat", token: Callable[[str], str]) -> str:
+    *lists, inner = fmt.name.split("_X_")
+    body = ", ".join(token(t) for t in inner.split("_"))
+    return " x ".join([token(t) for t in lists] + [f"[{body}]"])
+
+
+def shape_desc(fmt: "lmc_ops.EngineKVFormat") -> str:
```
ApostaC: nit: This name conflicts with the `PageBufferShapeDesc` (there is a lot of `shape_desc.xx` in the existing code, if I remember correctly)
Can we rename it as `describe_shape` or something similar?

lmcache/v1/gpu_connector/kv_format/specs/base.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+
+    engine_kv_format: ClassVar["lmc_ops.EngineKVFormat"]
+    is_mla: ClassVar[bool] = False
+    is_hnd: ClassVar[bool] = False
+    is_cross_layer: ClassVar[bool] = False
+    attention_backends: ClassVar[tuple[str, ...]] = ()
+
+    def __init__(self, kv_caches: DiscoverableKVCache) -> None:
+        self.kv_caches = kv_caches
```
ApostaC, line 112: It's a bit dangerous to let the KVFormatSpec object hold the raw reference to `kv_caches` because it may cause a memory leak.
We want to make sure all the references to the GPU kv tensors are removed when the serving engine gets disconnected.

ApostaC (commented):
Left some comments, PTAL.

ApostaC (approved):
LGTM

deng451e (approved):

lmcache/v1/gpu_connector/kv_format/detectors/registry.py
maobaolong: I like this pattern 👍

PR #3487, opened 2026-06-01
fix(vllm/v1): no-op KV transfer when LMCache is in degraded mode

Anai-Guo:
## Summary

When LMCache fails to initialize (for any reason), `LMCacheManager` leaves `lmcache_engine` as `None` and logs that the system will operate in **degraded / recompute mode**. The worker-side connector entry points, however, still do a bare `assert self.lmcache_engine is not None`, so the first request after startup crashes EngineCore instead of falling back to recompute.

Because `EngineDeadError` shuts the app down cleanly (exit 0), under Kubernetes the pod restarts indefinitely — any transient LMCache init failure becomes a permanent crash loop.

This guards the three worker-side methods that touch the engine so they no-op (and let vLLM recompute) when the engine is `None`:

- `start_load_kv`
- `save_kv_layer`
- `wait_for_save`

This mirrors the existing degraded-mode handling already present in `_setup_metrics` (which returns early when `lmcache_engine_metadata` is `None`). In normal operation the engine is non-`None`, so behavior is unchanged.

Reported in vllm-project/vllm#44249, where it reproduces across two image/version combinations with two different underlying init failures (ZMQ RPC lookup timeout; `cpuinfo.get_cpu_info()` JSONDecodeError) — confirming the connector-side contract violation is independent of which failure leaves `lmcache_engine = None`.

## Behavior

- **Engine healthy (normal path):** `lmcache_engine` is not `None`; guards are no-ops.
- **Engine degraded (init failed):** load/save are skipped; vLLM recomputes KV instead of crashing EngineCore on the first request.

## Test plan

- [ ] Start vLLM with the LMCache connector configured to fail init (e.g. unreachable remote backend) and confirm the first request returns a normal completion via recompute instead of killing EngineCore.
- [ ] Sanity-check that a healthy LMCache run is unaffected (cache hits/saves still occur).


ApostaC:
@Anai-Guo Thanks for contributing to LMCache!
@sammshen Can you help take a quick look at this PR, thanks!

Anai-Guo:
Extended this PR to also cover the **scheduler-side** hooks, following @Sunt-ing's matrix in vllm-project/vllm#44249. The original commit only guarded the worker-side hooks (`start_load_kv` / `save_kv_layer` / `wait_for_save`), which left "scheduler degraded only" and "LMCache fully unavailable" still crashing EngineCore with `EngineDeadError`.

The new commit no-ops the four scheduler-side hooks when `lookup_client is None` (degraded mode), mirroring the worker-side guards:

- `get_num_new_matched_tokens` → `0` (no external hit → recompute)
- `update_state_after_alloc` → no-op
- `build_connector_meta` → empty `LMCacheConnectorMetadata()`
- `request_finished` → `(False, None)`

This is the same `_init_failed` degraded-mode contract the worker guards already follow, just the scheduler-role siblings, so the full "any LMCache init failure → recompute" scope is now met in one PR rather than a follow-up.

sammshen (approved):

deng451e (approved):

Anai-Guo:
Friendly ping @ApostaC — this has two approvals (@sammshen, @deng451e) and is mergeable. Is there anything else needed before it can be merged?

Anai-Guo:
Friendly ping @maintainers — this has two approvals and has been green for a few days. Is there anything else needed before it can be merged? Happy to rebase if useful.

hlin99:
thanks for your contribution!
auto-merge is triggered. this PR will be merged automatically once all CI are passed

Anai-Guo:
Friendly ping @hlin99 — auto-merge was triggered a few days ago but the buildkite `k3-comprehensive-test` / `k3-integration-test` jobs appear to be failing (they require cluster infra, so likely not something fixable from the PR side). Could these be re-run on your infra, or is there anything else needed before this can merge? Happy to help.

Anai-Guo:
Friendly ping @sammshen @ApostaC — this has two approvals and is mergeable. Is there anything else needed before it can be merged? Happy to rebase if required.

hlin99:
hi, can you check if the CI fails are relevant to your change? thx

Anai-Guo:
Good catch — yes, the 4 failures in `tests/v1/test_v1_adapter_request_finished_scheduler_role.py` were caused by this PR, not flaky infra.

The regression came from an early-return I had added at the top of `request_finished`:

```python
if self.lmcache_engine is None or self.lookup_client is None:
    return False, None
```

That short-circuited the whole method whenever *either* backend was missing, which dropped the storage/lookup cancels that must still run when only *one* of the two is absent (e.g. Scheduler role: `lmcache_engine is None` but `lookup_client` present → the async lookup cancel was being skipped, leaking in-flight lookups).

I've removed that early return and now rely on the per-branch fail-soft logic further down, which already warns and skips *only* the unavailable backend. All 4 tests exercise exactly these combinations and should pass now. Pushed in the latest commit — mind re-running CI? Thanks!

hlin99:
there is a DCO issue in your last commit

Anai-Guo:
Thanks @hlin99 — fixed. The last commit was missing a matching `Signed-off-by` (author name didn't match the sign-off); I've re-signed it and DCO is green now.

PR #3491, opened 2026-06-02
[core] Bitmap-based prefetch result + pluggable TrimPolicy

deng451e:
## Summary

Refactors the standard L2→L1 prefetch path so the retained-key set is carried as a `Bitmap` end-to-end, and the "which found keys to keep" decision becomes a pluggable `TrimPolicy`. **Behavior-preserving** for the existing contiguous-prefix path; adds a `SEGMENTED_PREFIX` policy for non-contiguous retained sets.

This is groundwork: a follow-up PR adds sparse (scattered-key) retrieval on top of this foundation, which then only needs to add a policy + `covered_keys` rather than a parallel code path.

## What changed

**`prefetch_controller.py`**
- `TrimPolicy {PREFIX, SEGMENTED_PREFIX}` + `retained_mask(found, num_keys, policy, segments)` — the single "which found keys to retain" decision.
- `trim_load_plan_with_mask` replaces the prefix-only trim; `trim_load_plan_to_prefix` / `trim_load_plan_to_first_n_keys` removed.
- `_completed_lookups` / `_completed_results`, `query_lookup_result` / `query_prefetch_result`, and `_complete_request` / `_update_lookup_results` all carry a `Bitmap`. The load-plan trim and `_finalize_load` release the complement of `retained_mask(...)` — identical to `count_leading_ones` for `PREFIX`.
- `submit_prefetch_request` gains `policy` / `segments` (default `PREFIX`).

**`storage_manager.py`**
- `PrefetchHandle`: `l1_prefix_hit_count: int` → `l1_found_indices: tuple`, plus `l2_orig_indices: tuple` (maps the controller's local result bitmap back to original key positions).
- `query_prefetch_status` / `query_prefetch_lookup_hits` return a `Bitmap`; a `_combine_found` helper merges the L1 indices with the L2 result. Callers derive the prefix hit count via `count_leading_ones()`.

**Consumers** (`lookup`, `blend`, trace `codecs`) and the affected tests derive the count via `count_leading_ones()`.

## Why bitmap (not int)

A single int can only represent a contiguous prefix. Carrying a `Bitmap` is what makes `SEGMENTED_PREFIX` (and, later, sparse) representable, and collapses the previous `int | Bitmap` result unions.

## Testing
- All existing dense prefetch / storage-manager tests pass (behavior preservation).
- Adds a `SEGMENTED_PREFIX` controller test (retains `{0, 1, 3, 4}` across a gap, excludes the masked-out index).
- `pre-commit` clean: ruff, ruff-format, isort, codespell, mypy.


lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -47,83 +47,66 @@
...
-    {0, 1, 3}, the trimmed plan will only include key indices [0, 1]
-    and exclude index 3.
+def merge_bitmaps(bitmaps: Iterable[Bitmap], num_keys: int) -> Bitmap:
+    """Merge multiple bitmaps with a bitwise OR."""
+    if not bitmaps:
+        return Bitmap(0)
+    merged = Bitmap(num_keys)
+    for bm in bitmaps:
+        merged = merged | bm
+    return merged
```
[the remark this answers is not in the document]
deng451e: Fixed in 7c539f44 — `merge_bitmaps` now always initializes and returns a `Bitmap(num_keys)` (dropped the `if not bitmaps` / `Bitmap(0)` path), so `SEGMENTED_PREFIX`'s `found & segments` can no longer size-mismatch and empty generators are handled correctly. Added `TestMergeBitmaps` (empty list, empty generator, union) to lock it in.

lmcache/v1/distributed/storage_manager.py
```diff
@@ -502,65 +528,54 @@ def query_prefetch_lookup_hits(
...
                 total_hits,
                 handle.total_requested_keys,
-                handle.l1_prefix_hit_count,
-                l2_result,
                 elapsed_ms,
                 handle.external_request_id,
                 handle.prefetch_request_id,
             )
-        return total_hits
+        return found
```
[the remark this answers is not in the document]
deng451e: Fixed in 7c539f44 — the completion log now uses `found.popcount()` and says "retained keys" instead of "prefix hits", so it is accurate for non-contiguous policies (`SEGMENTED_PREFIX` / `SPARSE`). For the contiguous `PREFIX` case `popcount() == count_leading_ones()`, so no behavior change there. (The return value was already the full bitmap; this only affected the log line.)

lmcache/v1/mp_observability/trace/codecs.py
```diff
@@ -193,19 +193,21 @@ def _enc_prefetch_handle(h: PrefetchHandle) -> dict[str, Any]:
     return {
         "prefetch_request_id": h.prefetch_request_id,
         "external_request_id": h.external_request_id,
-        "l1_prefix_hit_count": h.l1_prefix_hit_count,
+        "l1_found_indices": list(h.l1_found_indices),
```
ApostaC, line 198: I think putting the hit count in the traces is still helpful.
Do you think putting the indices in the otel spans helps more in debugging?

lmcache/v1/distributed/api.py
```diff
@@ -171,15 +171,19 @@ class PrefetchHandle:
...
+    """Original-key indices found (read-locked) in L1 at submission time."""
 
     total_requested_keys: int
-    """Total number of keys originally requested."""
+    """Total number of keys originally requested (the result-bitmap size)."""
 
     submit_time: float
     """Monotonic timestamp when the prefetch task was submitted."""
 
+    l2_orig_indices: tuple[int, ...] = ()
```
ApostaC, line 196: Can we name it `l2_found_indices`?
I understand we need to make a change from L2 bitmap to the indices, but the caller (i.e., the "user" for the `PrefetchHandle` class) doesn't need to know this.

lmcache/v1/distributed/api.py
```diff
@@ -171,15 +171,19 @@ class PrefetchHandle:
     external_request_id: str
     """Request ID from the caller for end-to-end tracing."""
 
-    l1_prefix_hit_count: int
-    """Number of leading keys already in L1 at submission time."""
+    l1_found_indices: tuple[int, ...]
+    """Original-key indices found (read-locked) in L1 at submission time."""
```
ApostaC, line 188: Let's discuss this: whether we want to "trim" in L1.

ApostaC (commented):
Comments before we review together

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -47,83 +47,68 @@
...
+    found: Bitmap,
+    num_keys: int,
+    policy: TrimPolicy = TrimPolicy.PREFIX,
+    segments: Bitmap | None = None,
+) -> Bitmap:
+    """Subset of ``found`` to keep (load + read-lock + report); the rest is
+    released.
+
+    PREFIX = the leading contiguous run; SEGMENTED_PREFIX = the intersection
+    with ``segments``.
```
ApostaC: Need to have a better docstring for this function, including what this function logically does at a high level (not describing the implementation), the "Args" section, and the "Returns" section.

ApostaC: Also, maybe rename this function to `create_trim_mask`?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -295,18 +300,22 @@ def submit_prefetch_request(
         keys: list[ObjectKey],
         layout_desc: MemoryLayoutDesc,
         extra_count: int = 0,
+        policy: TrimPolicy = TrimPolicy.PREFIX,
+        segments: Bitmap | None = None,
```
ApostaC: No need to pass `segment` here

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -317,22 +326,27 @@ def submit_prefetch_request(
...
+                (request_id, keys, layout_desc, extra_count, policy, segments)
+            )
         self._submission_efd.notify()
         return request_id
 
-    def query_lookup_result(self, request_id: PrefetchRequestId) -> int | None:
+    def query_lookup_result(self, request_id: PrefetchRequestId) -> Bitmap | None:
         """
-        Query the number of prefix hits from the lookup phase.
+        Query the retained-key bitmap from the lookup phase.
```
ApostaC: nit: "Query the keys that are found during the lookup for a specific request".

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -702,7 +729,7 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
             if request.write_reserved_keys:
                 l1_mgr.finish_write(request.write_reserved_keys)
                 l1_mgr.delete(request.write_reserved_keys)
-            self._update_lookup_results(request.request_id, 0)
+            self._update_lookup_results(request.request_id, Bitmap(num_keys))
```
ApostaC: we can keep the old "int" for `_update_lookup_result`

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -47,83 +47,68 @@
...
-        return {}
 
-    trimmed_plan: dict[int, Bitmap] = {}
-    mask_bitmap = Bitmap(num_keys, n)
-    for adapter_idx, bitmap in load_plan.items():
-        new_bitmap = bitmap & mask_bitmap
-        if new_bitmap.popcount() == 0:
-            continue
+    PREFIX = enum.auto()
+    SEGMENTED_PREFIX = enum.auto()
```
ApostaC: nit: Let's move the TrimPolicy into `internal_api.py`

lmcache/v1/distributed/api.py
```diff
@@ -171,15 +171,19 @@ class PrefetchHandle:
...
 
     total_requested_keys: int
-    """Total number of keys originally requested."""
+    """Total number of keys originally requested (the result-bitmap size)."""
 
     submit_time: float
     """Monotonic timestamp when the prefetch task was submitted."""
 
+    l2_found_indices: tuple[int, ...] = ()
+    """Original-key index of each key submitted to L2; maps the controller's
```
ApostaC: `l2_orig_indices`

lmcache/v1/distributed/storage_manager.py
```diff
@@ -501,66 +526,65 @@ def query_prefetch_lookup_hits(
...
+            l1_hits = len(handle.l1_found_indices)
+            l2_hits = l2_r.popcount() if l2_r is not None else 0
             logger.info(
                 "Prefetch request completed (L1+L2): "
-                "%d/%d prefix hits (%d L1, %d L2) "
-                "in %.1f ms "
-                "(external_request_id=%s, "
-                "prefetch_request_id=%d)",
+                "%d/%d retained keys (%d L1, %d L2) in %.1f ms "
+                "(external_request_id=%s, prefetch_request_id=%d)",
```
ApostaC, line 578: add one more thing in the log: number of prefix hits (`found.count_leading_ones`)

Help people double-check whether the behavior is expected.

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3492, opened 2026-06-02
[core] Sparse prefetch via TrimPolicy.SPARSE + covered_keys

deng451e:
## Summary

Adds **sparse prefetch** on top of the `Bitmap` + `TrimPolicy` refactor in #3491.

`TrimPolicy.SPARSE` retains *every* found key (gap-tolerant) rather than the contiguous prefix, so a caller can coalesce scattered, independently-usable chunks (e.g. CacheBlend non-prefix retrieval) into a single high-bandwidth L2 load. There is no parallel code path — sparse is just another policy on the unified bitmap result from #3491.
 
## What changed
- **`prefetch_controller.py`**: add `TrimPolicy.SPARSE` (retain-all) + the `retained_mask` case. The controller already threads `policy` from #3491, so the load-plan trim and `_finalize_load` need no further change.
- **`storage_manager.py`**: `submit_prefetch_task` gains `policy` / `covered_keys`. The `SPARSE` branch keeps a read lock on every L1-found key (minus `covered_keys`), releases the rest, and submits all L1 misses to L2 as one sparse request (scattered `l2_orig_indices`). `covered_keys` (served elsewhere, e.g. the parent prefix) are excluded from the load and the found set and their probe locks released — so the locked set equals what the caller retrieves (no leak).

## How callers read it
`query_prefetch_status(handle)` returns the found `Bitmap`; sparse callers read `get_indices_list()` (the gap-tolerant set), dense callers `count_leading_ones()`. No separate query method.

## Testing
- Adds 3 sparse tests: retains all found across a gap, loads all found from L2, and `covered_keys` excluded + read-locks released.
- Full dense prefetch / storage / controller suites stay green.
- `pre-commit` clean (ruff, ruff-format, isort, codespell, mypy).

### DIFF
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -394,7 +394,6 @@ class StorageManager:
         extra_count: int = 0,
         external_request_id: str = "",
         policy: TrimPolicy = TrimPolicy.PREFIX,
-        covered_keys: set[ObjectKey] | None = None,
     ) -> PrefetchHandle:
         """Prefetch objects into L1 asynchronously.
 
@@ -409,10 +408,6 @@ class StorageManager:
             policy: Which retained-subset policy to apply (see
                 :class:`TrimPolicy`).  ``PREFIX`` keeps the contiguous prefix;
                 ``SPARSE`` keeps every found key (gap-tolerant).
-            covered_keys: ``SPARSE`` only. Keys already served elsewhere (e.g.
-                the parent prefix): excluded from the L2 load and the found
-                set, and any probe read-lock on them is released — so the
-                locked set equals what the caller retrieves (no leak).
 
         Returns:
             PrefetchHandle to track the task.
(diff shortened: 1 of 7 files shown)


lmcache/v1/distributed/storage_manager.py
```diff
@@ -403,6 +406,13 @@ def submit_prefetch_task(
                 key.  Total locks = 1 + extra_count.
             external_request_id: Request ID from the caller
                 for end-to-end log tracing.
+            policy: Which retained-subset policy to apply (see
+                :class:`TrimPolicy`).  ``PREFIX`` keeps the contiguous prefix;
+                ``SPARSE`` keeps every found key (gap-tolerant).
+            covered_keys: ``SPARSE`` only. Keys already served elsewhere (e.g.
+                the parent prefix): excluded from the L2 load and the found
+                set, and any probe read-lock on them is released — so the
+                locked set equals what the caller retrieves (no leak).
```
ApostaC: Why do we introduce `covered_keys`?

lmcache/v1/distributed/storage_manager.py
```diff
@@ -411,6 +421,71 @@ def submit_prefetch_task(
...
+                self._l1_manager.finish_read(keys_to_release, extra_count=extra_count)
+
+            # L1 misses (minus covered) -> L2 sparse; keep each original index.
+            found_set = set(l1_found_indices)
+            sparse_l2_indices = [
+                i
+                for i in range(len(keys))
+                if i not in found_set and keys[i] not in covered
+            ]
+            remaining_keys = [keys[i] for i in sparse_l2_indices]
```
ApostaC: Current implementation has multiple loops, which could be inefficient. 
We can rewrite the implementation here with a single loop over the `keys` list.

ApostaC (commented):
Some comments below.

ApostaC (approved):
LGTM!

royyhuang (approved):

PR #3494, opened 2026-06-02
[perf] Reduce lock acquisitions in RustRawBlockBackend.batched_remove

nayeonikim:

`RustRawBlockBackend` previously inherited the default `batched_remove`, which
loops over keys and calls single-key `remove()` per entry — acquiring `_pin_lock`
and issuing a `RawBlockCore.delete_many` call once *per key*. Under bulk eviction
this causes repeated lock acquisition and core round-trips.

This PR overrides `batched_remove` to process the batch within a single
`_pin_lock` critical section and issue a single `delete_many` call for the whole
batch, while preserving the existing contract (per-key removal semantics,
`force` handling, returned removed-count).

- Signature is identical to the base `batched_remove`, so it's a compatible
  override — no caller changes required.
- `force=False` preserves pinned/locked entries (delegated to
  `RawBlockCore.delete_many`); `force=True` removes them and clears the matching
  `_pinned_keys` entries.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
+++ b/lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py
@@ -376,6 +376,34 @@ class RustRawBlockBackend(StoragePluginInterface):
                 self._pinned_keys.discard(spec.encoded)
         return removed
 
+    def batched_remove(
+        self,
+        keys: list[CacheEngineKey],
+        force: bool = True,
+    ) -> int:
+        """Remove multiple keys in a single locked batch.
+
+        Acquires ``_pin_lock`` once and issues one ``_core.delete_many`` call
+        for the whole batch, instead of locking per key.
+
+        Args:
+            keys: Cache keys to remove.
+            force: Passed through to ``RawBlockCore.delete_many``. When false,
+                locked entries are preserved.
+
+        Returns:
+            Number of keys that were actually removed.
+        """
+        if not keys:
+            return 0
+        encoded_keys = [encode_legacy_key(key).encoded for key in keys]
+        with self._pin_lock:
+            results = self._core.delete_many(encoded_keys, force=force)
+            for encoded_key, removed in zip(encoded_keys, results, strict=True):
+                if removed:
+                    self._pinned_keys.discard(encoded_key)
+        return sum(results)
+
     def batched_submit_put_task(
         self,
         keys: Sequence[CacheEngineKey],
--- a/tests/v1/storage_backend/test_rust_raw_block_backend.py
+++ b/tests/v1/storage_backend/test_rust_raw_block_backend.py
@@ -603,6 +603,105 @@ def test_rust_raw_block_backend_pin_and_contains_are_idempotent(
 @pytest.mark.skipif(
     not _has_ext(), reason="lmcache_rust_raw_block_io extension not installed"
 )
+def test_rust_raw_block_backend_batched_remove(memory_allocator, loop_in_thread):
+    """batched_remove should drop every supplied key in a single locked batch.
+
+    Also verifies that pinned keys are preserved when ``force=False``, mirroring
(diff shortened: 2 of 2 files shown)


nayeonikim:
Hi @DongDongJu! The Buildkite k3-unit-tests failed due to a CI infrastructure issue unrelated to the code changes — `setup-lmcache-only-env.sh` failed at line 33 with a warning about leftover CUDA shared memory from a previous run. Could you please restart the Buildkite build when you get a chance? Thank you! 🙏

DongDongJu (changes requested):
Hello @nayeonikim,
Correctness test looks good. 
Since this PR is mainly a perf optimization, it may be worth adding a small regression check that batched_remove() does not fall back to per-key remove().
Do you have any perf diff number?

nayeonikim:
> Hello @nayeonikim, Correctness test looks good. Since this PR is mainly a perf optimization, it may be worth adding a small regression check that batched_remove() does not fall back to per-key remove(). Do you have any perf diff number?

Thanks for the review!

  For the regression check, I added `test_batched_remove_does_not_fall_back_to_per_key`. It spies on `remove()` and `delete_many` and asserts the whole batch is
  removed through a single `delete_many` call without ever invoking the per-key `remove()` path.

  On the perf side, this is really a lock-acquisition reduction rather than a throughput optimization: it drops `_pin_lock` and core-lock acquisitions from N to 1 per
  batch (deterministic). Single-threaded it's a small, consistent win and confirms no regression (measured on the real backend at the component level, since the
  delete path is in-memory with no I/O):

  | N | per-key | batched | locks |
  |---|---|---|---|
  | 100 | 0.24ms | 0.19ms | 100 → 1 |
  | 500 | 1.33ms | 0.97ms | 500 → 1 |
  | 1000 | 4.60ms | 3.65ms | 1000 → 1 |

Same behavior, one lock acquisition per batch instead of N.

DongDongJu:
> > Hello @nayeonikim, Correctness test looks good. Since this PR is mainly a perf optimization, it may be worth adding a small regression check that batched_remove() does not fall back to per-key remove(). Do you have any perf diff number?
> 
> Thanks for the review!
> 
> For the regression check, I added `test_batched_remove_does_not_fall_back_to_per_key`. It spies on `remove()` and `delete_many` and asserts the whole batch is removed through a single `delete_many` call without ever invoking the per-key `remove()` path.
> 
> On the perf side, this is really a lock-acquisition reduction rather than a throughput optimization: it drops `_pin_lock` and core-lock acquisitions from N to 1 per batch (deterministic). Single-threaded it's a small, consistent win and confirms no regression (measured on the real backend at the component level, since the delete path is in-memory with no I/O):
> 
> N	per-key	batched	locks
> 100	0.24ms	0.19ms	100 → 1
> 500	1.33ms	0.97ms	500 → 1
> 1000	4.60ms	3.65ms	1000 → 1
> Same behavior, one lock acquisition per batch instead of N.

Thanks for the detail and contribution!

DongDongJu (approved):
direction LGTM.

nayeonikim:
@ApostaC @sammshen  Hi! Gentle ping on this! Would you be able to take a look when you get a chance?

DongDongJu:

nayeonikim:
@sammshen @ApostaC @deng451e — this one's been approved by @DongDongJu since Jun 5 and just needs a second approving review to be mergeable. Could one of you take a look when you get a chance? It's a small one.

nayeonikim:
Hi @zhengfeihe — this PR touches `lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py`, which falls under the `storage_backend/` CODEOWNERS entry you're on. Could you self-assign and take a look when you have time? Thanks!

DongDongJu:
@maobaolong @zhengfeihe PTAL

zhengfeihe (approved):
Sorry for missing this. LGTM!

PR #3496, opened 2026-06-02
[Docs][operator] Fix CRD spec docs and policy enum gaps

3xdevv:

Cross-checked `docs/source/mp/operator.rst` and the operator CRD types against the actual Go and Python server code, and fixed three categories of issues:

**CRD enum gaps** (`lmcacheengine_types.go`):
- `eviction.policy`: added `noop` — the server supports it and the existing docs already recommend it for buffer-only mode (`storePolicy: skip_l1`), but the CRD rejected it at admission time
- `prefetchPolicy`: added `retain` — the server registers both `default` and `retain`, but the operator only allowed `default`

**Missing fields in the CRD spec reference** (`operator.rst`):
- `server.httpPort` (default `8080`) was not documented at all
- `gpuVendor` (`nvidia` / `amd`) was missing entirely from the spec reference
- `NVIDIA_DRIVER_CAPABILITIES=all` was absent from the auto-injected pod settings list
- `l2Backends` → `l2Backend` (field name was wrong; the CRD uses the singular form)

**Comment fix** (`lmcacheengine_controller.go`):
- Duplicate step numbers in reconcile comments

- `IsolatedLRU` eviction policy was intentionally **not** added to the CRD enum — it was merged 1 month ago, requires `cache_salt` to be passed or raises `ValueError`, and is designed to pair with the quota API rather than be used as a standalone option. Users can reach it via `extraArgs` if needed.
- No behaviour changes — the enum additions unlock values the server already accepts; the doc changes are documentation-only.

## If applicable

### DIFF
--- a/docs/source/mp/operator.rst
+++ b/docs/source/mp/operator.rst
@@ -265,7 +265,8 @@ Eviction
      - Description
    * - ``eviction.policy``
      - ``LRU``
-     - Only ``LRU`` is supported.
+     - ``LRU`` or ``noop``.  Use ``noop`` with ``l2Backend.storePolicy: skip_l1``
+       for buffer-only mode.
    * - ``eviction.triggerWatermark``
      - ``0.8``
      - Usage ratio (0.0--1.0] to trigger eviction.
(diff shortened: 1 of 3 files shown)


sammshen (approved):
LGTM!

deng451e (approved):
LGTM

PR #3503, opened 2026-06-03
[core] Native bulk-set: build the found bitmap via batched_set + gather

deng451e:
## Summary

`StorageManager._combine_found` built the found-key bitmap with a **per-bit Python loop** — one `Bitmap.set()` pybind crossing per found key. This moves the work into `csrc` using two **generic, reusable** bitmap primitives instead of a special-purpose helper:

- **`Bitmap.batched_set(indices)`** — set every listed bit in one native call (bounds-checked; positions `>= size` ignored). This is the reusable bulk-set primitive we'd deferred.
- **`Bitmap.gather` widened to accept any sequence** (was list-only) — so the `l2_orig_indices` *tuple* maps each L2 set bit back to its original position with no Python-side copy. Backward-compatible (existing callers pass lists).

`_combine_found` becomes a composition — seed the L1 indices, then `batched_set` the gathered L2 hits:

```python
found = Bitmap(total)
found.batched_set(handle.l1_found_indices)
if l2_local is not None:
    found.batched_set(l2_local.gather(handle.l2_orig_indices))
return found
```

**No signature change** — it still returns `Bitmap`, so `query_prefetch_status` and the prefix-hit consumers (`lookup`, `blend`) are untouched. Policy-agnostic (PREFIX / SEGMENTED_PREFIX / SPARSE); out-of-range positions are dropped (`set()` bounds-check + `gather`'s length check).

## Perf (native vs the per-bit Python loop)

| keys found / total | Python loop | native | speedup |
|---:|---:|---:|---:|
| 128 / 256 | 14.7 µs | 2.5 µs | 6.0× |
| 1024 / 2048 | 115 µs | 12.4 µs | 9.4× |
| 4096 / 8192 | 458 µs | 44.9 µs | 10.2× |

`gather` only materializes the L2 set-bit subset, so the win grows with working-set size.

## Tests
- `TestBatchedSet` (`tests/v1/native_storage_ops/test_bitmap.py`) — bounds-checking, accumulation, dedup, tuple input, a randomized Python-reference oracle, and the end-to-end `_combine_found` composition pattern; plus a `gather`-accepts-tuple case.
- `test_distributed_storage_manager` drives the real `submit_prefetch_task → query_prefetch_status → _combine_found` path (dense + `policy=SPARSE`).
- 274 passed; pre-commit (clang-format, ruff, mypy) green.

### DIFF
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -493,12 +493,11 @@ class StorageManager:
         ``handle.l2_orig_indices``.
         """
         found = Bitmap(handle.total_requested_keys)
-        for i in handle.l1_found_indices:
-            found.set(i)
+        found.batched_set(handle.l1_found_indices)
         if l2_local is not None:
-            orig = handle.l2_orig_indices
-            for local_i in l2_local.get_indices_list():
-                found.set(orig[local_i])
+            # gather maps each L2 set bit i to its original position
+            # ``l2_orig_indices[i]``; batched_set drops any position >= size.
+            found.batched_set(l2_local.gather(handle.l2_orig_indices))
         return found
 
     def query_prefetch_lookup_hits(
--- a/csrc/storage_manager/bitmap.cpp
+++ b/csrc/storage_manager/bitmap.cpp
@@ -41,6 +41,10 @@ void Bitmap::set(size_t index) {
   data_[byte_index(index)] |= static_cast<uint8_t>(1u << bit_offset(index));
 }
 
+void Bitmap::batched_set(const std::vector<size_t>& indices) {
+  for (size_t idx : indices) set(idx);
+}
+
 void Bitmap::clear(size_t index) {
   if (index >= size_) return;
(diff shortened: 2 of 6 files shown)


lmcache/v1/distributed/storage_manager.py
```diff
@@ -492,14 +492,16 @@ def _combine_found(
...
+        if l2_local is None:
+            found = Bitmap(handle.total_requested_keys)
+            for i in handle.l1_found_indices:
+                found.set(i)
+            return found
+        return l2_local.union_remapped(
+            handle.total_requested_keys,  # size
+            handle.l2_orig_indices,  # index_map
+            handle.l1_found_indices,  # seed_indices
+        )
```
ApostaC:
```python
found.batched_set(l1_found_indices)
l2_indices = l2_local_bitmap.gather(l2_orig_indices)
found.batched_set(l2_indices)
```

ApostaC (approved):
LGTM!

royyhuang (approved):

PR #3469, opened 2026-05-31
fix: Add missing completion recorder ops to python_ops_fallback

Alorun:
Fixes #3468

**What this PR does / why we need it**:

`record_completion_on_stream` / `drain_recorded_completions` exist in the native
ops but were never mirrored in `lmcache.python_ops_fallback`. On any build without
the compiled extension (where `_get_backend()` falls back to the pure-Python
module), the dispatcher in `native_completion._drain_once` calls
`_lmc_ops.drain_recorded_completions()` and crashes with:
Unlike the event recorder (`record_event_on_stream` / `drain_recorded_events`),
which is observability-only and guarded by a `hasattr` check, the completion
recorder is part of the core control flow: it drives the `finish_write` /
`finish_read_prefetched` callbacks that actually complete store/retrieve
operations. So it can't simply be skipped on the fallback path — it needs a real
implementation, otherwise those callbacks never fire and operations hang/lose
results silently.

This PR implements both functions in `python_ops_fallback` as a thread-safe,
in-process queue: enqueue immediately on `record` and atomically snapshot+clear on
`drain`, both protected by a lock. No CUDA stream ordering is needed on the
fallback path, since there is no asynchronous GPU work to order against.

**Special notes for your reviewers**:

- The lock is required: `drain`'s snapshot + clear is two operations; without it, a
  concurrent `record` could be cleared without being returned (lost completion).
  `test_concurrent_no_data_loss` is a regression guard for exactly this.
- Tests are split by scope:
  - `tests/v1/test_python_ops_fallback.py` — 5 unit/contract tests (FIFO, types,
    drain-clears, stream-ptr-ignored, concurrency). These run in the standard CI
    suite.
  - `tests/v1/multiprocess/test_completion_recorder.py` — 1 integration test
    verifying the full `submit_callback_to_stream → drain → handler` path under the
    fallback (via monkeypatching `_lmc_ops`).
- **CI gap to flag:** `tests/v1/multiprocess/` is currently excluded from CI via
  `--ignore` in `test.yml`, so the integration test added here won't run in CI
  as-is — which is also part of why this regression slipped through originally. The
  unit tests do run. I'm happy to add a dedicated CPU job for the multiprocess
  tests if you'd like; let me know the preferred approach.
- Verified locally on Linux/CUDA: full suite (`82 passed`) and the integration test
  both green.

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [x] this PR contains unit tests

Issue #3468, opened 2026-05-31
Module 'lmcache.python_ops_fallback' has no attribute 'drain_recorded_completions'


maobaolong:
**Label**
Please label your issue with "bug" and any other relevant labels so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/1882)

**Describe the bug**

When we didn't recompile the csrc, or didn't or cannot compile csrc at all, the `lmcache.python_ops_fallback` is a fallback, This bug occurs when new changes or API updates are made to the C code, but the corresponding functions are missing in python_ops_fallback.

```
[2026-05-31 09:14:11,909] LMCache ERROR: DeviceHostFuncDispatcher: drain failed (native_completion.py:116:lmcache.v1.multiprocess.native_completion)
Traceback (most recent call last):
  File "/Users/msy/projects/LMCache2/lmcache/v1/multiprocess/native_completion.py", line 114, in _drain_once
    completions = _lmc_ops.drain_recorded_completions()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'lmcache.python_ops_fallback' has no attribute 'drain_recorded_completions'
```

**To Reproduce**

python /Users/msy/projects/LMCache2/lmcache/v1/multiprocess/http_server.py --l1-size-gb=1 --eviction-policy=LRU --http-port 8080 --http-host localhost --no-l1-use-lazy 

**Expected behavior**

No error occurred.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

maobaolong:
@lyj1007 Would you like to take this issue?

Alorun:
I'd like to work on this. /assign
I've traced the root cause — record_completion_on_stream / drain_recorded_completions are missing from python_ops_fallback, so non-CUDA builds crash in _drain_once. Fix in progress (thread-safe fallback impl + tests), PR coming shortly.

PR #3507, opened 2026-06-03
[CI/CD] Add parity test between c_ops and python_ops_fallback

Lyj1007:
### Motivation

#3468 reported that two `c_ops` symbols had no fallback mirror, breaking
CPU-only deployments. #3469 fixed those two symbols, but nothing prevents
the same drift from reappearing — the two modules are kept in sync purely
by convention.

Inspired by the parity-enforcement approach originally proposed in #3082, 
this PR adds a standalone test that discovers the public surface 
of `lmcache.c_ops` at runtime and asserts every symbol has a matching fallback.

Running the test already surfaces three more `c_ops` symbols missing from
`python_ops_fallback`:

- `record_event_on_stream`
- `drain_recorded_events`
- `multi_layer_block_kv_transfer`

### What this PR does

**Parity test** — Adds `tests/v1/test_c_ops_fallback_parity.py`, which discovers the public
surface of `lmcache.c_ops` at runtime and asserts:

- every public callable / enum / descriptor class in `c_ops` exists in
  `python_ops_fallback`, enforced by three coverage tests:
    - `test_all_c_ops_callables_have_fallback` — free functions;
    - `test_all_c_ops_enums_have_fallback` — enum types;
    - `test_all_c_ops_descriptors_have_fallback` — pybind11 `py::class_<>`
      descriptor types (e.g. `PageBufferShapeDesc`);
- for shared names, parameter count, parameter names (when available),
  default values, enum members, and descriptor attributes all match.

Skips automatically on CPU-only environments where `c_ops` cannot be
imported.

**Fixes for discovered gaps** — Of the three missing symbols surfaced by the test,
`multi_layer_block_kv_transfer` is already being addressed in #3508, so this
PR fixes the other two:

- `record_event_on_stream` / `drain_recorded_events`: added pure-Python
  fallbacks using a `threading.Lock` + global buffer (same pattern as the
  completion recorder fallback from #3469), with immediate `time.time()`
  timestamping in place of CUDA host-callback ordering.

A `scenario_record_drain_event` behavioral test is also added to
`test_python_ops_fallback.py` to verify cross-backend parity.


sammshen (approved):
LGTM

ApostaC (approved):
LGTM!

ApostaC:
@Lyj1007 Actually, the new UT fails to pass. Can you try it locally?

From the buildkite output, I see it does not find the CUDA kernels used for GPU transfer. I'm actually thinking about whether we need to have a Python fallback for that? @maobaolong wdyt?

ApostaC:
Oh I see #3508

maobaolong:
@Lyj1007 Would you like to add a exclude list for this PR? Then the UT can passed before merge #3508

Lyj1007:
> @Lyj1007 Would you like to add a exclude list for this PR? Then the UT can passed before merge #3508

Thanks @baolong! Added the exclude list and verified UT passes locally. Will remove `multi_layer_block_kv_transfer` from the list once #3508 is merged. PTAL 🙏

maobaolong:
> > @Lyj1007 Would you like to add a exclude list for this PR? Then the UT can passed before merge #3508
> 
> Thanks @baolong! Added the exclude list and verified UT passes locally. Will remove `multi_layer_block_kv_transfer` from the list once #3508 is merged. PTAL 🙏

@Lyj1007 Thanks for this great work, it really help to keep python_ops_fallback works all the time.

maobaolong (approved):
LGTM

maobaolong:
@Lyj1007 The following error message was pasted from the CI logs. It appears to be related to this PR.
```python
=================================== FAILURES ===================================
______________________ test_descriptor_class_parity[Any] _______________________

cls_name = 'Any'

    @pytest.mark.skipif(not HAS_C_OPS, reason="c_ops not available (no CUDA)")
    @pytest.mark.parametrize(
        "cls_name",
        _shared_desc_names if _shared_desc_names else ["__placeholder__"],
    )
    def test_descriptor_class_parity(cls_name: str) -> None:
        """For every descriptor class shared between c_ops and the fallback,
        every public, non-callable attribute on the c_ops instance must also
        be present on the fallback instance.
        """
        if cls_name == "__placeholder__":
            pytest.skip("No shared descriptor classes found")
    
>       c_inst = _c_ops_descs[cls_name]()
                 ^^^^^^^^^^^^^^^^^^^^^^^^

tests/v1/test_c_ops_fallback_parity.py:374: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = typing.Any, args = (), kwargs = {}

    def __new__(cls, *args, **kwargs):
        if cls is Any:
>           raise TypeError("Any cannot be instantiated")
E           TypeError: Any cannot be instantiated

/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/typing.py:553: TypeError
```

Lyj1007:
> @Lyj1007 The following error message was pasted from the CI logs. It appears to be related to this PR.
> 
> ```python
> =================================== FAILURES ===================================
> ______________________ test_descriptor_class_parity[Any] _______________________
> 
> cls_name = 'Any'
> 
>     @pytest.mark.skipif(not HAS_C_OPS, reason="c_ops not available (no CUDA)")
>     @pytest.mark.parametrize(
>         "cls_name",
>         _shared_desc_names if _shared_desc_names else ["__placeholder__"],
>     )
>     def test_descriptor_class_parity(cls_name: str) -> None:
>         """For every descriptor class shared between c_ops and the fallback,
>         every public, non-callable attribute on the c_ops instance must also
>         be present on the fallback instance.
>         """
>         if cls_name == "__placeholder__":
>             pytest.skip("No shared descriptor classes found")
>     
> >       c_inst = _c_ops_descs[cls_name]()
>                  ^^^^^^^^^^^^^^^^^^^^^^^^
> 
> tests/v1/test_c_ops_fallback_parity.py:374: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> cls = typing.Any, args = (), kwargs = {}
> 
>     def __new__(cls, *args, **kwargs):
>         if cls is Any:
> >           raise TypeError("Any cannot be instantiated")
> E           TypeError: Any cannot be instantiated
> 
> /opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/typing.py:553: TypeError
> ```

Thanks for the report! This is already fixed in the latest commit — I added a `__module__` filter to exclude imported types like `typing.Any`. 🙏

hlin99:
hi @Lyj1007 one test is skipped. I'm not sure if this is expected skip. can you help confirm? thanks.

tests/v1/test_c_ops_fallback_parity.py::test_function_signature_parity[PageBufferShapeDesc] SKIPPED (Cannot inspect c_ops.PageBufferShapeDesc signature)                                                               [  2%]

Lyj1007:
> hi @Lyj1007 one test is skipped. I'm not sure if this is expected skip. can you help confirm? thanks.
> 
> tests/v1/test_c_ops_fallback_parity.py::test_function_signature_parity[PageBufferShapeDesc] SKIPPED (Cannot inspect c_ops.PageBufferShapeDesc signature) [ 2%]

Hi @hlin99, thanks for catching this — yes, that skip is **expected**.

### Why it shows up as SKIPPED

`PageBufferShapeDesc` is a pybind11 `py::class_<>` descriptor type, not a plain function. In the parity test, `_public_callables` collects anything `callable(obj) == True`, which includes classes (since `MyClass()` is callable). That's why `PageBufferShapeDesc` ends up parametrized into `test_function_signature_parity`.

Once it's in there, the test tries two ways to read its signature:

1. **Parse the pybind11 docstring** (`func(arg: type = default) -> ret`) — fails, because a `py::class_<>` docstring doesn't follow that format.
2. **Fallback to `inspect.signature(...)`** — fails, because pybind11 class constructors don't expose a usable Python signature.

Both paths fail, so the test hits the safe branch:

```python
pytest.skip(f"Cannot inspect c_ops.{func_name} signature")
```

### Why no coverage is lost

Descriptor classes have their own dedicated test:

```
test_descriptor_class_parity[PageBufferShapeDesc]
```

It compares the public **attribute set** between `c_ops.PageBufferShapeDesc` and `python_ops_fallback.PageBufferShapeDesc` — which is the right way to check parity for a class. So `PageBufferShapeDesc` is fully covered, just by a different test.

In general, any similar descriptor class that gets SKIPPED here can be safely ignored — they will all be checked later in `test_descriptor_class_parity[<ClassName>]`, which verifies parity by comparing the public attribute sets of `c_ops.<ClassName>` and `python_ops_fallback.<ClassName>`.

In short, once your changes are merged, all you need to do is revert that single commit (`2d6b143` in #3507), and `multi_layer_block_kv_transfer` will be back under full parity coverage.

PR #3508, opened 2026-06-03
feat(ops): add multi_layer_block_kv_transfer Python fallback as unified MP transfer primitive

hlin99:

Adds a pure-Python fallback implementation of `multi_layer_block_kv_transfer`, closing the gap between the Python fallback layer and the existing C ops.

This op serves as the **fundamental primitive for MP mode data transfer** — a single, unified copy operation regardless of whether the transfer happens on the worker side (data transfer path) or the server side (handle transfer path).

**Design intent:**

1. **Bridge the Python/C gap** — the C ops already ship this kernel; the Python fallback was missing it, making non-CUDA or debug environments incomplete.
2. **Single op, both paths** — `multi_layer_block_kv_transfer` is the one op that moves KV data in MP mode. It doesn't matter whether the caller is on the data-transfer path (worker-side gather/scatter) or the handle-transfer path (server-side reconstruction); both invoke the same primitive.
3. **Device contract for MP mode** — any new device/accelerator that wants to participate in MP mode only needs to implement this op. It does not need to understand which upper-layer transfer path is active.
4. **Upper-layer simplicity** — higher-level code (transfer contexts, strategies) simply calls this op to perform the copy, agnostic to the underlying path selection.


ApostaC:
Hey @hlin99 , if the goal is to make #3505 pass, we can just have a Python fallback function without a real implementation.

I think it could make sense because we should only use this function in a GPU environment, in which case we must have the c_ops compiled (i.e., the Python callback should never be called).

hlin99:
> Hey @hlin99 , if the goal is to make #3505 pass, we can just have a Python fallback function without a real implementation.
> 
> I think it could make sense because we should only use this function in a GPU environment, in which case we must have the c_ops compiled (i.e., the Python callback should never be called).

hi @ApostaC . actually i'm trying to abstract a single ops (multi_layer_block_kv_transfer) for gpu path & non_gpu path.
so this function even in fallback has to be full functional.
I will do another PR later to refactor a bit in non-gpu path to call this function in gather/scatter paths

so moving forward, for mp mode, regardless it's gpu or non-gpu path, we have fallback way to make it work, and vendors should implement their own way to make it work better. and if one day, any device becomes able to go through gpu path, they don't need to care about the ops neither. WDYT?

hlin99:
> > Hey @hlin99 , if the goal is to make #3505 pass, we can just have a Python fallback function without a real implementation.
> > I think it could make sense because we should only use this function in a GPU environment, in which case we must have the c_ops compiled (i.e., the Python callback should never be called).
> 
> hi @ApostaC . actually i'm trying to abstract a single ops (multi_layer_block_kv_transfer) for gpu path & non_gpu path. so this function even in fallback has to be full functional. I will do another PR later to refactor a bit in non-gpu path to call this function in gather/scatter paths
> 
> so moving forward, for mp mode, regardless it's gpu or non-gpu path, we have fallback way to make it work, and vendors should implement their own way to make it work better. and if one day, any device becomes able to go through gpu path, they don't need to care about the ops neither. WDYT?

underneath ops needn't aware the upper path (cuda/non-cuda), and upper layer needn't care about the underneath ops implementation (fallback or c ops)

lmcache/python_ops_fallback.py
```diff
@@ -725,6 +725,826 @@ def multi_layer_kv_transfer_unilateral(
...
+    if shape_desc is not None and shape_desc.element_size > 0:
+        dtype = _ELEMENT_SIZE_TO_DTYPE.get(shape_desc.element_size)
+        if dtype is None:
+            raise ValueError(
+                f"Unsupported element_size {shape_desc.element_size!r} in "
+                "shape_desc; cannot infer KV dtype. "
+                f"Supported sizes: {sorted(_ELEMENT_SIZE_TO_DTYPE)}"
+            )
+        return dtype
+    return torch.float16
```
maobaolong: Maybe the dtype is bfloat

hlin99: thanks for catching. fixed.

Lyj1007:
Hi @hlin99 🙌
Just a small heads-up — in #3507 I added a parity test (`tests/v1/test_c_ops_fallback_parity.py`) to make sure every function in `c_ops` has a matching implementation in `python_ops_fallback`, so we can catch C/Python mismatches early.

Since `multi_layer_block_kv_transfer` is exactly what you're implementing here, I temporarily added it to an exclude list so the UT can pass before this PR lands. Once your changes are merged, would you mind dropping that entry (introduced in commit 2d6b143 of #3507)? That'll bring the function back under full parity coverage.

Thanks you! 🙏

hlin99:
> Hi @hlin99 🙌 Just a small heads-up — in #3507 I added a parity test (`tests/v1/test_c_ops_fallback_parity.py`) to make sure every function in `c_ops` has a matching implementation in `python_ops_fallback`, so we can catch C/Python mismatches early.
> 
> Since `multi_layer_block_kv_transfer` is exactly what you're implementing here, I temporarily added it to an exclude list so the UT can pass before this PR lands. Once your changes are merged, would you mind dropping that entry (introduced in commit [2d6b143](https://github.com/LMCache/LMCache/commit/2d6b14372875574f5822acb0a89d66a8bb65b904) of #3507)? That'll bring the function back under full parity coverage.
> 
> Thanks you! 🙏

sure. No problem at all. Please go ahead to merge your PR. I can switch it on here.

maobaolong (approved):
@hlin99 The logic of multi_layer_block_kv_transfer looks extreme complex, but this is not the issue of this PR.

LGTM overall. I've tested.

maobaolong:
@chunxiaozheng @yoo-kumaneko Would you like to take a double check?

tests/v1/test_c_ops_fallback_parity.py
```diff
@@ -211,16 +211,9 @@ def _has_real_names(params):
 
 # ── Discover the intersection automatically ──
 
-# Functions intentionally excluded from parity checks.
-_EXCLUDED_FUNCS: set[str] = {
```
maobaolong, line 215: Suggest to keep this trick and just clear the content in case of future similar case.

hlin99: good suggestion! Done

hlin99:
hi @ApostaC  @sammshen  would you like to take another look? thx

sammshen (approved):

ApostaC (approved):
LGTM!

PR #3514, opened 2026-06-03
refactor: add set_shape_desc_dtype helper to avoid scattered try/except

maobaolong:
## What

Extract the repeated `try/except AttributeError` pattern around `shape_desc.dtype = ...` into a single helper `set_shape_desc_dtype()` in `python_ops_fallback.py`.

Also add a `dtype` slot to the pure-Python `PageBufferShapeDesc` fallback class.

## Why

The C++ `PageBufferShapeDesc` struct (bound via pybind11) has no `dtype` field, so assigning `shape_desc.dtype = dtype` raises `AttributeError` at runtime when the compiled extension is active. Three call sites worked around this independently with identical `try/except AttributeError: pass` blocks -- which is noisy and easy to forget when adding new call sites.

The pure-Python fallback (`PageBufferShapeDesc` in `python_ops_fallback.py`) *does* need the `dtype` field because `element_size` alone cannot distinguish `float16` from `bfloat16` (both are 2 bytes), and the CPU fallback kernel uses `dtype` to pick the right code path.

### DIFF
--- a/lmcache/python_ops_fallback.py
+++ b/lmcache/python_ops_fallback.py
@@ -318,6 +318,7 @@ class PageBufferShapeDesc:
         "hs",
         "element_size",
         "block_stride_elems",
+        "dtype",
     )
 
     def __init__(self) -> None:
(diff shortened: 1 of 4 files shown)


hlin99 (approved):
LGTM

chunxiaozheng (approved):
LGTM!

PR #3516, opened 2026-06-03
[MP][Feat] Register MP servers with the coordinator

ruizhang0101:

Adds the MP-server side of coordinator membership. An MP server can now join a coordinator — register on startup, heartbeat on a timer, deregister on shutdown — all on its own event loop and best-effort (a down coordinator never blocks the server). Opt-in via `--coordinator-url` (env fallback `LMCACHE_COORDINATOR_*`); with no URL set, behavior is unchanged.

- Register/heartbeat/deregister are one identity lifecycle in `keep_registered` (shared `assigned_id`, 404 → re-register). Transient failures keep the id; only a 404 or a failed initial register drops it.
- Config follows the MP-server convention: `CoordinatorConfig` + `--coordinator-*` flags with env fallback; heartbeat interval validated (finite, > 0) at parse time.
- **Follow-up PR (touches observability, kept out to stay focused):** when `--service-instance-id` is unset, OTel mints a local UUID but the coordinator receives `""` and assigns a *different* id → telemetry id ≠ membership id. Fix = resolve one shared identity.

**If applicable**:


ApostaC (approved):
LGTM as we discussed offline.

deng451e (approved):

PR #3519, opened 2026-06-04
[Feat][RawBlock] Add BLKDISCARD startup support

3xdevv:

Adds BLKDISCARD support for raw-block storage initialization.

This PR:
- adds `RawBlockDevice.discard(offset, length)` in the Rust raw-block binding
- splits long BLKDISCARD requests in Rust using the kernel `discard_max_bytes` limit when available
- adds `blkdiscard_on_init` to `RawBlockCore`, `RustRawBlockBackend`, and `RawBlockL2AdapterConfig`
- issues BLKDISCARD over the raw-block device during startup when configured
- rejects `blkdiscard_on_init=true` with `load_checkpoint_on_init=true`
- documents the new option in user-facing and design docs
- adds unit coverage for discard invocation, and invalid config rejection

This helps deployments start from a clean raw-block device while using an empty in-memory index.

`BLKDISCARD` is Linux-specific and device-dependent. If the underlying device does not support discard/TRIM, is not a block device, or rejects one discard chunk, the operation logs a warning and startup continues.
`blkdiscard_on_init` requires `load_checkpoint_on_init=false` because discarding the device is incompatible with loading existing on-device metadata.

**If applicable**:

### DIFF
--- a/docs/design/v1/distributed/l2_adapters/raw_block.md
+++ b/docs/design/v1/distributed/l2_adapters/raw_block.md
@@ -142,6 +142,22 @@ The on-device format is intentionally unchanged by the MP adapter work.
 Recovered keys are exposed to the shared L2 eviction policy on adapter startup,
 so reclaimed slots come from global L2 eviction or explicit `delete()` calls.
 
+## Startup Discard
+
+Raw block deployments sometimes intentionally start with an empty in-memory
+index instead of loading an on-device checkpoint. In that mode, bytes from a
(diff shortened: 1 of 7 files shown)


sammshen (approved):
LGTM, leave final reviews to @DongDongJu

3xdevv:
Hello, @DongDongJu. Could you look at this for a moment?
It's a feature addition that does not affect runtime performance.

PR #3520, opened 2026-06-04
[Bugfix] batched_nixl_desc_exists never hits on FILE (dynamic NIXL)

sihara:
The dynamic NIXL backend's batched contains calls batched_nixl_desc_exists, which only calls nixl_agent.query_memory. query_memory resolves object stores (OBJ), not POSIX files, so on a FILE backend (POSIX) it returns all-not-found -> 0 hits on every lookup. Stores succeed (files land on disk) but every request reports 'LMCache hit tokens: 0' and recomputes + re-stores, so the cache is never reused.

Add a FILE branch that delegates to nixl_desc_exists and pass the directory from the caller. The new path parameter defaults to None, so OBJ callers and the query_memory fast-path are unchanged.

Verified on a dynamic POSIX backend (nixl_pool_size: 0), APC+async off: external prefix cache hit rate 0.0% -> 99.6%, KV streamed back from storage (~56-59 GB/s) instead of recomputed on every request.

**BEFORE (bug present) — every lookup misses, nothing read back:**

```
LMCache INFO: Reqid: ...8fe76f38, Total tokens 118087, Inference Engine computed tokens: 0, LMCache hit tokens: 0, need to load: 0
LMCache INFO: Reqid: ...85b8bf6c, Total tokens 118087, Inference Engine computed tokens: 16, LMCache hit tokens: 0, need to load: 0
APIServer INFO loggers.py:259 Engine 000: ... External prefix cache hit rate: 0.0%
LMCache INFO: Reqid: ...bdb4c02f, Total tokens 118177, Inference Engine computed tokens: 118032, LMCache hit tokens: 0, need to load: 0
APIServer INFO loggers.py:259 Engine 000: ... Prefix cache hit rate: 13.6%, External prefix cache hit rate: 0.0%
```
(`External prefix cache hit rate` stays `0.0%` across the whole run; zero `Retrieved ... tokens` lines — full recompute every request.)

**AFTER (fixed) — lookups hit, KV streamed back from storage:**

```
LMCache INFO: Reqid: ...8167c1f0, Total tokens 118087, Inference Engine computed tokens: 0, LMCache hit tokens: 117760, need to load: 117760
LMCache INFO: [req_id=...8167c1f0] Retrieved 117760 out of 117760 required tokens (from 117760 total tokens). size: 17.9688 gb, cost 393.9024 ms, throughput: 45.6173 GB/s
LMCache INFO: Reqid: ...9adbe35f, Total tokens 118087, Inference Engine computed tokens: 16, LMCache hit tokens: 117760, need to load: 117744
LMCache INFO: [req_id=...8853a724] Retrieved 113664 out of 113664 required tokens (from 117760 total tokens). size: 17.3438 gb, cost 308.4387 ms, throughput: 56.2308 GB/s
LMCache INFO: [req_id=...9fe7d7b1] Retrieved 114688 out of 114688 required tokens (from 117760 total tokens). size: 17.5000 gb, cost 293.9446 ms, throughput: 59.5350 GB/s
APIServer INFO loggers.py:259 Engine 000: ... Prefix cache hit rate: 1.2%, External prefix cache hit rate: 99.6%
```
(First request after a fresh store logs `hit tokens: 0` — that's the cold store — then every reuse hits.)

### DIFF
--- a/lmcache/v1/storage_backend/nixl_storage_backend.py
+++ b/lmcache/v1/storage_backend/nixl_storage_backend.py
@@ -710,20 +710,40 @@ class NixlDynamicStorageAgent(NixlStorageAgent):
             return False
 
     def batched_nixl_desc_exists(
-        self, reg_list: List[tuple[int, int, int, str]]
+        self, reg_list: List[tuple[int, int, int, str]], path: Optional[str] = None
     ) -> int:
-        """Check if multiple descriptors exist via a single ``query_memory`` call.
+        """Check if multiple descriptors exist from the start of ``reg_list``.
 
         :param reg_list: List of tuples ``(0, 0, 0, meta_info)`` where
             *meta_info* is the formatted object-key string.
+        :param path: Directory for FILE backends (required for FILE; ignored
+            for OBJ). FILE existence is resolved on the filesystem, since NIXL
+            ``query_memory`` only answers for object stores.
         :return: Number of consecutive descriptors that exist from the
             start of the list.
-        :raises: No exceptions are raised. Errors from the underlying
-            ``query_memory`` call are caught internally and logged as
-            warnings; the method returns ``0`` in that case.
+        :raises ValueError: If ``path`` is ``None`` for a FILE backend.
+        :raises: Errors from the underlying ``query_memory`` call (OBJ
+            backends) are caught internally and logged as warnings; the
+            method returns ``0`` in that case.
         """
         if not reg_list:
             return 0
+
+        # FILE backends are not resolvable via NIXL ``query_memory`` (it only
+        # answers for object stores), so reuse the single-key
+        # ``nixl_desc_exists`` -- which already handles FILE via os.path.exists.
+        # Without this, FILE/POSIX dynamic backends always return 0 here, so
+        # every batched lookup misses and the cache is never reused.
+        if self.mem_type == "FILE":
+            if path is None:
+                raise ValueError("path must be provided for FILE backends")
+            consecutive_count = 0
+            for _, _, _, meta_info in reg_list:
+                if self.nixl_desc_exists(meta_info, path):
+                    consecutive_count += 1
+                else:
+                    break
+            return consecutive_count
 
         try:
             resp = self.nixl_agent.query_memory(
@@ -1979,7 +1999,7 @@ class NixlDynamicStorageBackend(NixlStorageBackend):
         reg_list = [(0, 0, 0, self._format_object_key(key)) for key in remaining_keys]
 
         # Use the agent's batched_nixl_desc_exists method
-        consecutive_hits = self.agent.batched_nixl_desc_exists(reg_list)
+        consecutive_hits = self.agent.batched_nixl_desc_exists(reg_list, self.path)
 
         # Update cache for the hits and return total count
         for i in range(consecutive_hits):
--- a/tests/v1/test_nixl_batched_contains.py
+++ b/tests/v1/test_nixl_batched_contains.py
@@ -61,6 +61,10 @@ def _mock_backend(**overrides) -> Mock:
         side_effect=lambda key: f"formatted_{key.chunk_hash}"
     )
     backend.agent.batched_nixl_desc_exists = Mock(return_value=0)
+    # batched_contains() reads self.path to pass it to the agent; the value
+    # is unused here (the agent is mocked), but Mock(spec=...) raises on the
+    # unset instance attribute without it.
+    backend.path = None
     for k, v in overrides.items():
         setattr(backend, k, v)
     return backend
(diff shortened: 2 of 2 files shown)


lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -688,6 +691,22 @@ def batched_nixl_desc_exists(
...
             return 0
 
+        # FILE backends are not resolvable via NIXL ``query_memory`` (it only
+        # answers for object stores), so reuse the single-key
+        # ``nixl_desc_exists`` -- which already handles FILE via os.path.exists.
+        # Without this, FILE/POSIX dynamic backends always return 0 here, so
+        # every batched lookup misses and the cache is never reused.
+        if self.mem_type == "FILE":
+            if path is None:
+                return 0
```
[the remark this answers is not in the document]
sihara: ack and revised.

yanok:
I will take another look soon. But yes, your bug description sounds quite plausible: the PRs that enabled dynamic backend with files and added batched_contains were done roughly at the same time by different people. So I guess it merged without textual conflicts, but one PR was not aware of another, as a result batched_contains has no special handling for files.

tests/v1/test_nixl_batched_contains.py
```diff
@@ -60,6 +60,7 @@ def _mock_backend(**overrides) -> Mock:
         side_effect=lambda key: f"formatted_{key.chunk_hash}"
     )
     backend.agent.batched_nixl_desc_exists = Mock(return_value=0)
+    backend.path = "/tmp/nixl_test"
```
yanok: I wonder why this is needed? Real `batched_nixl_desc_exists` won't be executed, since it's a mock, right? So it doesn't need a path. Or am I missing something?

yanok (approved):
LGTM, the fix is sound, thanks!

It can probably use more testing like:
1. With OBJ contains/batched_contains result in `queryMem` calls
2. With FILE contains/batched_contains result in os.exists calls
3. (less important) with FILE and no path set batched_contains -> ValueError

Well, you change only touches batched_contains, so you can restrict yourself to only testing it

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -673,12 +673,15 @@ def nixl_desc_exists(self, meta_info: str, path: str) -> bool:
...
+        """Check if multiple descriptors exist from the start of ``reg_list``.
 
         :param reg_list: List of tuples ``(0, 0, 0, meta_info)`` where
             *meta_info* is the formatted object-key string.
+        :param path: Directory for FILE backends (required for FILE; ignored
+            for OBJ). FILE existence is resolved on the filesystem, since NIXL
+            ``query_memory`` only answers for object stores.
         :return: Number of consecutive descriptors that exist from the
             start of the list.
         :raises: No exceptions are raised. Errors from the underlying
```
DongDongJu: Please change this docstring.

tests/v1/test_nixl_batched_contains.py
DongDongJu: I agree. please make the FILE based test.

DongDongJu (changes requested):
Hello @sihara, thanks for the clear bug report and repro numbers.
direction LGTM. I have few requests before merging it.

For dynamic FILE/POSIX NIXL, `query_memory` is the wrong existence check; reusing `nixl_desc_exists()` gives the FILE path the same `os.path.exists()` behavior as single-key `contains()`, while OBJ keeps the existing `query_memory` fast path.

Before merging, could we add regression tests for the new FILE behavior?
Also please update the docstring: it still says no exceptions are raised, 
but the new FILE path raises `ValueError` when `path` is missing.

tests/v1/test_nixl_batched_contains.py
sihara: The mocked batched_nixl_desc_exists doesn't use the path — but batched_contains (the method under test) reads self.path to pass it down. Since backend is Mock(spec=NixlDynamicStorageBackend) and path is an instance attribute (set in __init__, not on the class), Mock(spec=...) raises AttributeError on self.path unless it's set.  The value itself is unused (the agent is mocked), so backend.path = None works just as well.
I will add FILE based tests.

lmcache/v1/storage_backend/nixl_storage_backend.py
sihara: ok, will do.

sihara:
@DongDongJu I’ve updated the PR by adding comments and new test cases for the FILE backend, and I’ve pushed the changes. However, the CI tests have been failing repeatedly, and it doesn’t seem related to my changes. Could you please take a look when you have a chance? Thanks!

sihara:
@yanok @DongDongJu, could you please review the PR again? Thank you!

DongDongJu (approved):
LGTM

sihara:
@maobaolong @sammshen @chunxiaozheng Please review them. Thank you!

DongDongJu:

ApostaC (approved):
LGTM!

PR #3525, opened 2026-06-04
fix(ci): resolve AttributeError in test_execute_calls_run_http_server

hlin99:
Replace @patch decorator with pytest.importorskip + patch.object to ensure the http_server submodule is explicitly imported before patching. Skips the test gracefully when server dependencies (fastapi, uvicorn) are unavailable.

fix of  pytest tests/cli/commands/test_server.py errors.


maobaolong (approved):
@hlin99 Thanks for this fix!

yoo-kumaneko (approved):
LGTM

PR #3527, opened 2026-06-04
[Core] Restore LRU eviction for rust raw-block backend

daegyu94:

This PR restores LRU-based slot reclamation for the legacy Rust raw-block backend.
The eviction path was lost during a rebase, causing new puts to fail silently once
all raw-block slots were occupied.

When raw-block storage is full, the backend now evicts one unpinned least-recently-used
entry and retries the put. The retry is triggered only when `RawBlockCore` reports a
no-free-slot error, so non-capacity write failures do not evict unrelated cached entries.

The backend also rebuilds its LRU state after metadata recovery so that recovered
raw-block entries can participate in future eviction rounds.

`RawBlockPutManyResult` now includes a `no_free_slot_keys` field to distinguish capacity
failures from other put failures, while preserving the existing `results` and
`stored_keys` behavior.

Validated with:
```
ruff check lmcache/v1/storage_backend/raw_block/core.py \
  lmcache/v1/storage_backend/plugins/rust_raw_block_backend.py \
  tests/v1/storage_backend/test_rust_raw_block_backend.py

pytest -q tests/v1/storage_backend/test_raw_block_core.py \
  tests/v1/storage_backend/test_rust_raw_block_backend.py
```

**If applicable**:

### DIFF
--- a/docs/design/v1/distributed/l2_adapters/raw_block.md
+++ b/docs/design/v1/distributed/l2_adapters/raw_block.md
@@ -147,6 +147,30 @@ The legacy `RustRawBlockBackend` now acts as a thin facade over `RawBlockCore`.
 It preserves non-MP behavior such as prefix-oriented contains/get semantics,
 while the MP adapter uses the core's full-bitmap lookup/load API.
 
+### Backend-level LRU Eviction (Non-MP only)
+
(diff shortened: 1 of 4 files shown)


sammshen:
@DongDongJu bump for your expertise here

PR #3528, opened 2026-06-04
feat(bench): server_bench supports --mode cpu and --transfer-mode

maobaolong:
- --mode cpu: allocate POSIX-SHM-backed KV cache tensors via CpuShmTensorWrapper; the server maps the same physical pages for zero-copy STORE/RETRIEVE.
- --transfer-mode {auto,handle,data}: explicit override of the routing. auto keeps the historical gpu->handle / cpu->data dispatch. data uses REGISTER_KV_CACHE_NON_GPU_CONTEXT + PREPARE/COMMIT two-phase path with client-side self-check.
- handle mode on CPU is left as TODO for a follow-up PR.
- docs: add Bench Server section to configuration.rst.

**If applicable**:

### DIFF
--- a/lmcache/cli/commands/bench/server_bench/command.py
+++ b/lmcache/cli/commands/bench/server_bench/command.py
@@ -266,7 +266,7 @@ def run_server_bench(  # noqa: ARG001  (command kept for symmetry with siblings)
 
     # Resolve transfer mode. ``auto`` reproduces the historical
     # behaviour: gpu -> handle path, cpu -> data path.
-    transfer_mode = getattr(args, "transfer_mode", "auto")
+    transfer_mode = args.transfer_mode
     if transfer_mode == "auto":
         use_handle = use_gpu
     elif transfer_mode == "handle":
@@ -417,14 +417,14 @@ def run_server_bench(  # noqa: ARG001  (command kept for symmetry with siblings)
             client_kv_tensors = cpu_tensors
 
         # Register KV cache before any store/retrieve.
-        register_result = _send_register_kv_cache(
+        register_ok, register_response = _send_register_kv_cache(
             client,
             layout_hints=layout_hints,
             kv_caches=kv_wrappers if use_handle else None,
             use_gpu=use_gpu,
             use_handle=use_handle,
         )
-        print("REGISTER_KV_CACHE: %s" % ("OK" if register_result else "FAIL"))
+        print("REGISTER_KV_CACHE: %s" % ("OK" if register_ok else "FAIL"))
         print()
 
         # In data mode the server reply carries the SHM pool name
@@ -435,9 +435,9 @@ def run_server_bench(  # noqa: ARG001  (command kept for symmetry with siblings)
         # and unregister from the worker's resource tracker so the
         # segment is not unlinked when the bench exits -- the server
         # owns its lifetime.
-        if not use_handle and hasattr(register_result, "shm_name"):
-            shm_name = getattr(register_result, "shm_name", "")
-            pool_size = getattr(register_result, "pool_size", 0)
+        if not use_handle and register_ok and register_response is not None:
+            shm_name = register_response.shm_name
+            pool_size = register_response.pool_size
             if shm_name and pool_size > 0:
                 server_shm = shared_memory.SharedMemory(
                     name=shm_name.lstrip("/"), create=False
--- a/lmcache/cli/commands/bench/server_bench/command.py
+++ b/lmcache/cli/commands/bench/server_bench/command.py
@@ -53,8 +53,7 @@ from lmcache.cli.commands.bench.server_bench.helpers import (
     _DEFAULT_SHAPE_SPEC,
     _IMPORT_ERROR,
     DTYPE_MAP,
-    _allocate_cpu_kv_cache,
-    _allocate_gpu_kv_cache,
+    _allocate_kv_cache,
     _get_chunk_size,
     _process_request,
     _require_full_install,
@@ -304,7 +303,7 @@ def run_server_bench(  # noqa: ARG001  (command kept for symmetry with siblings)
         )
         # Paged KV demands identical ``NB`` / ``BS`` across all groups
         # (block_id -> slot maths is shared), but ``kv_size`` / ``NH`` /
-        # ``HS`` / ``dtype`` may vary per group. ``_allocate_gpu_kv_cache(
+        # ``HS`` / ``dtype`` may vary per group. ``_allocate_kv_cache(
         # groups=...)`` honours each group's own shape; ``_process_request``
         # only needs a single ``block_size`` / ``total_blocks``.
         first = layer_groups[0]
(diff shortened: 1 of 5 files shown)


lmcache/cli/commands/bench/server_bench/command.py
```diff
@@ -51,6 +53,7 @@
     _DEFAULT_SHAPE_SPEC,
     _IMPORT_ERROR,
     DTYPE_MAP,
+    _allocate_cpu_kv_cache,
```
hlin99: hi @maobaolong a quick question , is it possible to merge cpu/gpu kv_cache into single function. and inside can be diff path by torch_device_type. i guess majority of the function can be reused, and diverge is only part of the func?

lmcache/cli/commands/bench/server_bench/helpers.py
```diff
@@ -318,30 +394,235 @@ def _poll_prefetch_status(
...
 def _send_store(
     client: MessageQueueClient,
     key: IPCCacheEngineKey,
     block_offset: int = 0,
     block_size: int = 16,
     num_group_views: int = 1,
+    use_gpu: bool = True,
+    use_handle: bool | None = None,
+    client_tensors: list["torch.Tensor"] | None = None,
+    chunk_size: int = 0,
```
hlin99: can chun size be 0? or maybe 256 as default?

hlin99 (approved):
some nits, otherwise LGTM

maobaolong:
Thanks for the review @hlin99!

Addressed both nits in 79fc1035:
- merged `_allocate_cpu_kv_cache` / `_allocate_gpu_kv_cache` into a single `_allocate_kv_cache(use_gpu=...)`. The bodies were nearly identical, only diverging on whether to attach a CUDA device.
- bumped `_send_store(chunk_size=...)` default from 0 to 256 so it matches a typical server chunk size.

lmcache/cli/commands/bench/server_bench/command.py
maobaolong: done

lmcache/cli/commands/bench/server_bench/helpers.py
maobaolong: done

lmcache/cli/commands/bench/server_bench/helpers.py
```diff
@@ -242,40 +264,81 @@ def _send_register_kv_cache(
     model_name: str = _MODEL_NAME,
     world_size: int = _WORLD_SIZE,
     layout_hints: dict | None = None,
-    gpu_tensors: list[torch.Tensor] | None = None,
-) -> bool:
-    """REGISTER_KV_CACHE — register a KV cache context.
+    kv_caches: list[CudaIPCWrapper] | None = None,
+    use_gpu: bool = True,
+    use_handle: bool | None = None,
+) -> "bool | RegisterNonGpuContextResponse":
```
yoo-kumaneko: The return type `bool | RegisterNonGpuContextResponse` is a bit tricky here — in handle mode you return a plain bool, but in data mode you return the raw response object and rely on its truthiness for the "OK/FAIL" print up in `command.py`. That only works reliably if `RegisterNonGpuContextResponse` defines `__bool__` (a failure response that's non-None would still look like "OK"). Would be cleaner to always return a `(bool, Optional[RegisterNonGpuContextResponse])` tuple so the success flag is unambiguous regardless of mode.

lmcache/cli/commands/bench/server_bench/helpers.py
```diff
@@ -242,40 +264,81 @@ def _send_register_kv_cache(
...
+        dtype_str=dtype_str,
+        use_mla=False,
+    )
+    result = _call(client, RequestType.REGISTER_KV_CACHE_NON_GPU_CONTEXT, [payload])
+    if result is _TIMEOUT:
+        return False
+    # The data-mode register reply carries the server's SHM pool name
+    # and size; the bench keeps it so STORE/RETRIEVE can mmap the same
+    # pool and exchange tensor data via slot descriptors.
+    return result
```
yoo-kumaneko: The cold/warm self-check described in the docstring ("cold == warm proves the server returned the exact bytes we wrote") doesn't actually fire a comparison anywhere — both `cold_ground_truth` and the warm checksums just get printed separately. The zero-fill before RETRIEVE is a nice touch, but without a programmatic assert or at least a printed diff when they disagree, the self-check is really just two unrelated log lines. Even a simple `if cold != warm: print("CHECKSUM MISMATCH ...")` would make it useful.

lmcache/cli/commands/bench/server_bench/command.py
```diff
@@ -240,17 +259,31 @@ def run_server_bench(  # noqa: ARG001  (command kept for symmetry with siblings)
...
 
-    if not torch_dev.is_available():
+    use_gpu = args.mode == "gpu"
+    if use_gpu and not torch_dev.is_available():
         print("ERROR: --mode gpu requires CUDA")
         sys.exit(1)
 
+    # Resolve transfer mode. ``auto`` reproduces the historical
+    # behaviour: gpu -> handle path, cpu -> data path.
+    transfer_mode = getattr(args, "transfer_mode", "auto")
```
yoo-kumaneko: Nit: `getattr(args, "transfer_mode", "auto")` — `--transfer-mode` is registered unconditionally in `register_server_parser`, so argparse will always set `args.transfer_mode`. The defensive `getattr` is a leftover reflex that can quietly swallow a future rename or typo. Just `args.transfer_mode` is fine here.

yoo-kumaneko (commented):
Thanks for getting CPU mode in! A few things caught my eye during the review — nothing that blocks understanding the intent, but worth fixing before merge.

yoo-kumaneko (approved):
otherwise LGTM

maobaolong:
Thanks @yoo-kumaneko, all three are good catches. Pushed b9f1e2e1.

1. `args.transfer_mode` — fixed, dropped the getattr fallback.
2. cold vs warm comparison — the actual `if cold == warm: MATCH else: MISMATCH` (with per-chunk diff) does fire, but it lives in `command.py` around the two `_process_request` calls, not inside the helper. The helper just returns the per-pass digest list. Want me to move/duplicate that into the helper too, or is the current split OK?
3. register return type — totally fair, the dataclass response is always truthy so a failed reply would still have printed OK. Switched to `(success, Optional[response])` and treat empty `shm_name` as failure.

yoo-kumaneko:
> Thanks @yoo-kumaneko, all three are good catches. Pushed [b9f1e2e](https://github.com/LMCache/LMCache/commit/b9f1e2e18ccbf2a3bb7a0f9501c3391e4d6e9752).
> 
> 1. `args.transfer_mode` — fixed, dropped the getattr fallback.
> 2. cold vs warm comparison — the actual `if cold == warm: MATCH else: MISMATCH` (with per-chunk diff) does fire, but it lives in `command.py` around the two `_process_request` calls, not inside the helper. The helper just returns the per-pass digest list. Want me to move/duplicate that into the helper too, or is the current split OK?
> 3. register return type — totally fair, the dataclass response is always truthy so a failed reply would still have printed OK. Switched to `(success, Optional[response])` and treat empty `shm_name` as failure.

That looks great. Regarding the 2nd issue, I think the current split is fine. The PR hence looks ready to be merged.

PR #3531, opened 2026-06-04
refactor(mp): introduce create_cache_context factory

maobaolong:
This is a piece of https://github.com/LMCache/LMCache/pull/3352 to avoid the size of #3352 become too big.

So, I extract a platform-agnostic `create_cache_context` factory into `lmcache/v1/platform/cache_context.py`.

`gpu_transfer.py` now calls the factory instead of instantiating `GPUCacheContext` directly, keeping the call site decoupled from the concrete backend.

CPU and other accelerator backends will be wired in follow-up PRs.

Changes:
- Add `lmcache/v1/platform/cache_context.py` with `create_cache_context`
- Update `gpu_transfer.py` to use the factory
- Update test to monkeypatch `create_cache_context` instead of `GPUCacheContext`

### DIFF
--- a/lmcache/v1/multiprocess/modules/gpu_transfer.py
+++ b/lmcache/v1/multiprocess/modules/gpu_transfer.py
@@ -48,25 +48,27 @@ import lmcache.c_ops as lmc_ops
 logger = init_logger(__name__)
 
 
-def get_layout_desc(gpu_context: GPUCacheContext, num_tokens: int) -> MemoryLayoutDesc:
+def get_layout_desc(
+    cache_context: GPUCacheContext, num_tokens: int
+) -> MemoryLayoutDesc:
     """Get the memory layout description for a given GPU context and number of tokens.
 
     Supports multiple KV layer groups with different shapes and dtypes.
 
     Args:
-        gpu_context: The GPU cache context containing the KV cache information.
+        cache_context: The GPU cache context containing the KV cache information.
         num_tokens: The number of tokens to determine the layout for.
 
     Returns:
         MemoryLayoutDesc: The memory layout description containing shapes and dtypes.
     """
-    num_groups = gpu_context.kv_layer_groups_manager.num_groups
+    num_groups = cache_context.kv_layer_groups_manager.num_groups
     shapes = [
-        gpu_context.get_kv_buffer_shape(num_tokens, group_idx)
+        cache_context.get_kv_buffer_shape(num_tokens, group_idx)
         for group_idx in range(num_groups)
     ]
     dtypes = [
-        gpu_context.kv_layer_groups_manager.kv_layer_groups[group_idx].dtype
+        cache_context.kv_layer_groups_manager.kv_layer_groups[group_idx].dtype
         for group_idx in range(num_groups)
     ]
     return MemoryLayoutDesc(shapes=shapes, dtypes=dtypes)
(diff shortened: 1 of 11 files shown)


hlin99:
hi @maobaolong  fully understand a large work is split into small pieces for easy reviewing. thanks

is there a high level design doc available for me to understand the full picture as first step? how these small pieces can be compiled into a larger work later on. that may help a lot for reviewing. thx.

there might be a doc in some PR already, but i overlooked it. can you point it to me?

hlin99 (approved):
had an offline review with @maobaolong . LGTM

lmcache/v1/platform/cache_context.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import TYPE_CHECKING, Any
+
+# First Party
+from lmcache.utils import EngineType
+from lmcache.v1.gpu_connector.utils import LayoutHints
+from lmcache.v1.multiprocess.custom_types import KVCache
+
+if TYPE_CHECKING:
+    # First Party
+    from lmcache.v1.multiprocess.group_view import LMCacheGroupView
```
sammshen, line 28: what is this type? I cannot find it in the current codebase?

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -93,11 +94,15 @@ def batched_iteration(lst: list, batch_size: int) -> Generator[tuple, None, None
 
 @dataclass
 class GPUContextEntry:
```
sammshen: nit if we put the general cache context constructor here, do we also need to rename this to just `ContextEntry`?

lmcache/v1/platform/cache_context.py
maobaolong: This is introduced by kuntai these recent days, by #3419 , maybe you could fetch and rebase the upstream ?

lmcache/v1/multiprocess/modules/gpu_transfer.py
maobaolong: Great suggestion

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -119,7 +123,7 @@ class GPUTransferModule:
 
     def __init__(self, ctx: MPCacheEngineContext) -> None:
         self._ctx = ctx
-        self._gpu_contexts: dict[int, GPUContextEntry] = {}
+        self._gpu_contexts: dict[int, ContextEntry] = {}
```
yoo-kumaneko: Shall we still use the name "GPU_context"?

yoo-kumaneko (approved):
LGTM

lmcache/v1/multiprocess/modules/gpu_transfer.py
maobaolong: No, I will rename this together.

lmcache/v1/multiprocess/modules/gpu_transfer.py
maobaolong: @sammshen The rename operation make the changes out of this PR, would you prefer to do that? Now I've rename it. PTAL.

lmcache/v1/platform/cache_context.py
sammshen: i see it now after rebasing, thx

sammshen (approved):
LGTM!

DongDongJu (approved):
LGTM

PR #3535, opened 2026-06-04
[CI][MP] CPU-only mp mode unit test respect explicit CPU KV layout hints

DongDongJu:

Allow explicit vLLM KV layout hints to override the CPU HND fallback. 
This lets non-CUDA transfer paths handle NHD tensors correctly when the caller provides `{"kv_layout": "NHD"}`.

This keeps the existing CPU default behavior when no layout hint is provided, while fixing NHD gather/scatter handling for multiprocess non-CUDA transfer tests.

  Validated in a CPU-only/no-CUDA environment:
  - `tests/v1/multiprocess/test_non_cuda_data_transfer.py`: `27 passed`
  - `tests/v1/multiprocess`: `325 passed, 90 skipped`
  - `ruff check` and `ruff format --check` passed on changed files

  **If applicable**:

### DIFF
--- a/lmcache/v1/gpu_connector/utils.py
+++ b/lmcache/v1/gpu_connector/utils.py
@@ -587,7 +587,7 @@ def normalize_kv_and_discover_format(
         # however, get_kv_cache_layout from vllm.v1.attention.backends.utils
         # does not return the right layout for CPU attention.
         # Right fix should come from vllm side, but hardcode here as safeguard.
-        if torch_device_type == "cpu":
+        if torch_device_type == "cpu" and kv_layout is None:
             kv_layout = "HND"
             logger.info("CPU backend detected, using HND KV cache layout")
         elif kv_layout is None:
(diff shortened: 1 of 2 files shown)


ApostaC:
@hlin99 Can you take a look at this small fix? Thanks!

hlin99:
hi @DongDongJu thanks for this refactor

i just have one question..
if i recall correctly, in old version of vllm, the kv_layout is *not* None, but returns NHD which is wrong.
so after applying this PR, the old version of vllm may have problems. version after https://github.com/vllm-project/vllm/pull/42740 works fine.

so the question is: do we need to consider supporting old version of vllm? if no, it LGTM.

DongDongJu:
> hi @DongDongJu thanks for this refactor
> 
> i just have one question.. if i recall correctly, in old version of vllm, the kv_layout is _not_ None, but returns NHD which is wrong. so after applying this PR, the old version of vllm may have problems. version after [vllm-project/vllm#42740](https://github.com/vllm-project/vllm/pull/42740) works fine.
> 
> so the question is: do we need to consider supporting old version of vllm? if no, it LGTM.

I see. I always keep following the latest one and we had few release for previous versions.
I think our latest dev should follow the all others latest dev too.
But open to others opinion on it.

hlin99 (approved):
approve the PR
As vLLM now introduced fused kv for cpu, we probably can claim we support cpu with vLLM version > xxx. Forget the old version compatibility. @maobaolong WDYT? Can you double confirm? Thx

maobaolong:
@DongDongJu Would you like to resolve the conflicts first? Thank you!

PR #3537, opened 2026-06-04
[CI][Hotfix]: Pin vLLM version pre cpu layout change

sammshen:
vLLM nightly after 0.22.1rc1.dev116 standardized the CPU attention backend KV cache layout to a "blocks first" packed 4-D tensor ([num_blocks, num_kv_heads, block_size, 2*head_size]) in get_kv_cache_shape (vllm-project/vllm#44393). LMCache's normalize_kv_and_discover_format only handles the prior 5-D/3-D vLLM layouts, so the new 4-D packed tensor falls through and breaks the k3 integration tests.

Pin the shared k3 harness to vLLM 0.22.1rc1.dev116+g71df063c4 to unblock CI. Because the rolling nightly/cu130 index only keeps the latest build, point at vLLM's per-commit wheel index
(https://wheels.vllm.ai/<full-sha>/); torch still resolves from the pytorch cu130 index.

A follow-up PR adds a GPUKVFormat entry plus detection/offset logic for the packed layout and unpins this.

**If applicable**:


deng451e (approved):

DongDongJu (approved):
Thanks for the quick fix!

hlin99 (approved):
LGTM

sammshen:
will take solution from https://github.com/LMCache/LMCache/pull/3538 instead

PR #3540, opened 2026-06-04
Fixes #3490: in async-prefetch, skip GPU-resident chunks to avoid redundant prefetch

### LINKED ISSUE #3490: Async lookup/prefetch ignores `num_computed_tokens` and re-reads the full hit prefix from storage
While investigating #3473, I noticed what appears to be a separate issue related to async loading.

When async loading is enabled (`LMCACHE_ENABLE_ASYNC_LOADING=True`), the async lookup/prefetch path appears to load the **entire matched prefix** from the storage backend, including chunks that have already been computed and are already resident in vLLM's GPU prefix cache (APC).

This issue is not related to #3473. I confirmed the same behavior with `LocalDiskBackend` as well.

It appears that:

## Sync retrieve

`CacheEngine.retrieve()` is aware of `num_computed_tokens` and only loads the missing portion of the prefix. As described in the code comments:

```python
# num_computed_tokens = 288 => vLLM already has [0..287] (18 pages)
# LMCache hit_prefix_tokens = 512 => cache covers [0..511] (2 chunks)
# Skip chunk 1, retrieve chunk 2 ...
# need_to_load: 512 - 288 = 224 tokens
```

(`lmcache/v1/cache_engine.py`, retrieve path)

In other words, the synchronous retrieve path only loads the gap between what vLLM already has and what LMCache can provide.

## Async prefetch

`StorageManager.async_lookup_and_prefetch()` does not receive `num_computed_tokens` (or equivalent information). It determines the cache hit length and then loads the entire matched prefix from storage.

As a result, any overlap between:

- tokens already resident in APC / GPU memory
- tokens covered by the LMCache hit

is fetched from storage again unnecessarily.

## Impact

This results in:

- Wasted storage read bandwidth equal to the GPU-resident portion of the LMCache hit
- Unnecessary prefetch-buffer allocations
- Unnecessary memory-object allocations
- Increased backend load

This overhead occurs on every prefetch where APC and LMCache overlap, and grows over time in multi-turn workloads as more of the prefix becomes GPU-resident.

## Suggestion

When async loading is enabled, the prefetch path should ideally be aware of the number of already-computed tokens (or APC-resident tokens) and fetch only the missing portion of the prefix, similar to the synchronous retrieve path.

Otherwise, async loading may repeatedly read chunks from storage that are already available in GPU memory, increasing storage traffic without providing any benefit.

shreyasudgaonkar:
### Summary

  1. Passes num_computed_tokens from the vLLM scheduler through the full lookup-client stack into LMCacheEngine and StorageManager
  2. Chunks whose token range falls entirely within [0, num_computed_tokens) are skipped in async prefetch; the skip count is added back to retrieved_length so the scheduler response still reflects full prefix coverage
  3. Adds Prometheus counter lmcache:num_async_prefetch_skipped_tokens and an INFO log line to make the skip observable

### Local validation for the bug fix

  1. Validation was done using a local vLLM + LMCache setup with multiple prompts sharing a long prefix. When chunks were already resident in vLLM's APC, LMCache correctly skipped retrieving those from CPU storage.

  2. Test validation confirmed by the log line:
      ```
      Async prefetch: skipping 768 already-GPU-resident tokens for lookup_id=...
      need to load: 0
      ```
  3. If maintainers think so, then I can include the unit test in the PR and merge that too
   
### To validate the fix
  1. Navigate to the directory which contains the github clone with the fix 
  2. NO_NATIVE_EXT=1 pip install -e .
  3. Setup the lmcache yaml to enable async prefetch (enable_async_loading: true)
  5. export PROMETHEUS_MULTIPROC_DIR=/tmp/shared_prometheus
  6. LMCACHE_CONFIG_FILE=<your-lmcache-config.yaml> vllm serve <model> --enable-prefix-caching --kv-transfer-config '{"kv_connector":"LMCacheConnectorV1","kv_role":"kv_both"}'
  7. Run prompts which share a prefix
  8. Check for lmcache:num_async_prefetch_skipped_tokens in Prometheus or for the log line in the output of vllm to confirm that the fix is working.

### DIFF
--- a/lmcache/integration/vllm/vllm_v1_adapter.py
+++ b/lmcache/integration/vllm/vllm_v1_adapter.py
@@ -1386,6 +1386,7 @@ class LMCacheConnectorV1Impl:
                 token_ids,
                 lookup_id=req_id,
                 request_configs=request_configs,
+                num_computed_tokens=num_computed_tokens,
             )
 
         if num_external_hit_tokens is None:
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -48,6 +48,7 @@ class LMCacheStats:
     interval_lookup_hits: int
     interval_vllm_hit_tokens: int
     interval_prompt_tokens: int
+    interval_async_prefetch_skipped_tokens: int
 
     interval_num_slow_retrieval_by_time: int
     interval_num_slow_retrieval_by_speed: int
@@ -261,6 +262,7 @@ class LMCStatsMonitor:
         self.interval_lookup_hits = 0  # total hit tokens lookup
         self.interval_vllm_hit_tokens = 0  # total hit tokens in vllm
         self.interval_prompt_tokens = 0  # total prompt tokens
+        self.interval_async_prefetch_skipped_tokens = 0
         self.interval_lookup_0_hit_requests = 0
 
         self.interval_num_slow_retrieval_by_time = 0
@@ -591,6 +593,10 @@ class LMCStatsMonitor:
     def update_interval_prompt_tokens(self, delta: int):
         self.interval_prompt_tokens += delta
 
+    @thread_safe
+    def update_interval_async_prefetch_skipped_tokens(self, delta: int):
+        self.interval_async_prefetch_skipped_tokens += delta
+
     def _clear(self):
         """
         Clear all the distribution stats
@@ -606,6 +612,7 @@ class LMCStatsMonitor:
         self.interval_lookup_hits = 0
         self.interval_vllm_hit_tokens = 0
         self.interval_prompt_tokens = 0
+        self.interval_async_prefetch_skipped_tokens = 0
 
         self.interval_num_slow_retrieval_by_time = 0
         self.interval_num_slow_retrieval_by_speed = 0
@@ -822,6 +829,7 @@ class LMCStatsMonitor:
             interval_request_cache_lifespan=request_lifespan,
             interval_prompt_tokens=self.interval_prompt_tokens,
             interval_lookup_0_hit_requests=self.interval_lookup_0_hit_requests,
+            interval_async_prefetch_skipped_tokens=self.interval_async_prefetch_skipped_tokens,
         )
         self._clear()
         return ret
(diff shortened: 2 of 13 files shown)


sihara:
@shreyasudgaonkar 
I did test this patch with my workloads. there are two issues I hit.
<pre>
  File ".../vllm_v1_adapter.py", line 835, in start_load_kv
      ret_token_mask = self.lmcache_engine.retrieve(
  File ".../lmcache/v1/cache_engine.py", line 817, in retrieve
      reordered_chunks, tot_kv_size = self._async_process_tokens_internal(
  File ".../lmcache/v1/cache_engine.py", line 1596, in _async_process_tokens_internal
      future = self.event_manager.get_event_future(
  File ".../lmcache/v1/event_manager.py", line 142, in get_event_future
      assert event_id in done_events, (
  AssertionError: Event chatcmpl-af4d5ff1c01ce3a3-af16b0a9 of type EventType.LOADING is not done or not found.
  (EngineCore pid=80012) ERROR ... [core.py:1101] EngineCore encountered a fatal error.
</pre>

After I reverted below and crash was gone.
`
                self.async_lookup_server.send_response_to_scheduler(lookup_id, 0)
`

Other issue, it was not crash, but we are seeing an issue that recives HTTP 500 errors. like this
<pre>
  [21:19:41,288] storage_manager.py:716  Async prefetch: skipping 26624 already-GPU-resident tokens for lookup_id=…88a0…
  [21:19:41,343] storage_manager.py:646  Responding to scheduler … with retrieved length 28672 including num_skip_tokens 26624
  [21:19:41,477] vllm_v1_adapter.py:1431 Total tokens 29004, Inference Engine computed tokens: 18624, LMCache hit tokens: 28672, need to load: 10048
  [21:19:41,543] cache_engine.py:892     Retrieved 0 out of 10240 required tokens (from 28672 total tokens). … throughput: 0.0000 GB/s
  [21:19:41,543] vllm_v1_adapter.py:851  Request …88a0… The number of retrieved tokens is less than the expected number of tokens! This should not happen!
  [21:19:41,580] vllm_v1_adapter.py:939  Request …88a0… failed to load 10240 tokens across 640 blocks
  [21:19:42]     scheduler.py:2261        Failing 1 request(s) due to KV load failure (failure_policy=fail, 10240 tokens affected)
</pre>

These requests then return HTTP 500 to the client (failure_policy=fail).

My environment : "vLLM 0.18 + LMCache 0.4.6, NIXL POSIX backend, async loading on, ~16 concurrent requests."

shreyasudgaonkar:
@sihara apologies for the crashes. I am evaluating and will add a fix asap. Thanks for your help in uncovering the bugs.

ziruiliu:
FYI, I have tried this in sync mode in last year, as in PR #2111. However it is more complicated than I expected. The change was reverted in commit https://github.com/yoo-kumaneko/LMCache/commit/2febf17972955fc3832d78cb587db1be93ba5b70, because preemption in vllm will evict GPU cached blocks and num_computed_tokens may change when a request is rescheduled from preemption. Wish you test it thoroughly with preemption-injected cases

shreyasudgaonkar:
@ziruiliu That's exactly the problem I am running into. Thanks for confirming. I am reworking the fix and also planning tests related to preemption.

PR #3548, opened 2026-06-05
[WIP][MP][DSV4] Per-group logical block size + SWA-suffix-only offload

yoo-kumaneko:
> **Status: Work in Progress (draft).** Functional and validated end-to-end on
> DeepSeek-V4-Flash, but opened as a draft for early feedback on the approach
> while broader failure-mode coverage (APC-on, concurrency, V3.2 regression) is
> still being added. There is no "Work in Progress" label on this repo, so WIP
> is signalled via the draft status and the `[WIP]` title prefix.
>
> Supersedes the approach in #3261 (which predates the #3419 HMA redesign).
> This is a fresh PR rebuilt on top of #3419; #3261 is left open for now.

## Summary

Builds on the #3419 hybrid memory allocator (HMA) framework to make the
LMCache MP connector correct **and** efficient for DeepSeek-V4's hybrid KV
cache. It implements the two V4 items #3419 explicitly left as Non-Goals:

1. **DeepSeek-V4 logical→physical block translation** (per-group logical block
   size) — a correctness fix.
2. **Sliding-window load-plan trimming** (SWA-suffix-only offload) — an
   efficiency optimization.

No C++/CUDA changes: `block_stride_elems` and the padded-stride kernel walk
already landed upstream (#3171 / #3419).

## Problem

Under the hybrid KV cache manager, vLLM's engine groups have **different
logical block sizes** (V4-Flash: 256 for the full-attention MLA group, 64/4/8
for the SWA / compressor-state groups), while `cache_config.block_size`
collapses to their GCD. The server derived every group's `compress_ratio` /
`physical_chunk_size` from a single global
`inference_engine_logical_block_size`, and the store/retrieve loops used one
uniform `blocks_per_chunk`.

That is only correct for the group whose logical block size happens to equal
the assumed value (the MLA group). For the SWA / state groups the kernel
sliced too few block IDs per chunk and transferred only the **leading
fraction** of each chunk, leaving the rest untransferred. It went unnoticed
because the full-attention MLA group — which dominates output quality — was
the one group sized correctly.

### Per-group logical block size (correctness)

- `LayoutHints` gains `per_layer_inference_engine_logical_block_size`; a new
  `per_layer_inference_engine_logical_block_size_from_vllm` helper reads each
  `KVCacheGroupSpec.block_size` (with `UniformTypeKVCacheSpecs` inner-spec
  fallback), and the worker forwards it at registration.
- **Geometry only** — no block IDs are exposed to the server, and the hint is
  omitted for non-hybrid engines, so V3.2 / other engines keep the legacy
  scalar path unchanged (keeps #3419's minimal-server design).
- `KVLayerGroupsManager` derives `compress_ratio` / `physical_chunk_size` per
  group; the store/retrieve loops use per-group `blocks_per_chunk` for slicing,
  underflow checks, and the per-group skip-block computation.

### SWA-suffix-only offload (efficiency)

- For a sliding-window group the attention mask never reads tokens older than
  the window, so only the trailing `ceil(window / logical_block_size)` blocks
  of each chunk are attended on reload. Store/retrieve now move just that
  suffix for SWA groups (full chunks for full-attention groups). Default-on.
- A single source of truth (`storage_blocks_per_chunk` /
  `storage_slots_per_chunk` / `storage_tokens_per_chunk` /
  `chunk_suffix_offset_blocks`) drives the block-ID slice, the kernel
  `lmcache_chunk_size` arg, the tmp GPU buffer slot, the per-group buffer
  views, and the `MemoryObj` layout, so they cannot disagree. The connector
  keeps sending full-grain block IDs; the server selects the trailing window.

## Validation

DeepSeek-V4-Flash, tp=4, fp8, `--block-size 256`,
`--no-disable-hybrid-kv-cache-manager`, LMCache 256 GiB L1, APC off:

- Per-group geometry corrected: SWA/state groups `compress_ratio/physical_chunk`
  `4 / 256` → `1 / 1024`.
- Store/retrieve succeed with no shape error; ~37–42% external prefix cache
  hit rate; cold vs warm semantically identical.
- L1 usage **667 MB** (full chunks) → **38.7 MB** (suffix-only), ~17× shrink.
- `tests/v1` (`kv_layer_groups_manager`, `vllm_kv_cache_groups`,
  `vllm_mp_adapter`, `mp_mem_kernels`, `gpu_connector`): **123 passed, 2
  skipped**.

## Remaining before un-drafting

- Broader failure-mode coverage: APC-on, partial hits, concurrency /
  multi-request, V3.2 regression run.
- L2 (NIXL / FS) path on V4 HMA is not exercised here (L1 only).

### DIFF
--- a/lmcache/integration/vllm/kv_cache_groups.py
+++ b/lmcache/integration/vllm/kv_cache_groups.py
@@ -98,3 +98,104 @@ def create_group_views_from_vllm(
             per_layer_group_idx,
         )
     ]
+
+
+def spec_int_attr(spec: Any, attr: str) -> int:
+    """Read an integer ``attr`` off a vLLM ``KVCacheGroupSpec.kv_cache_spec``.
+
+    The value may live directly on the spec, or — for a
+    ``UniformTypeKVCacheSpecs`` wrapper that bundles several inner specs — on
+    its inner specs (all of which share one value), in which case the first
+    inner spec is read. Returns ``0`` when neither carries the attribute (or it
+    is ``None``), so callers treat ``0`` uniformly as "absent".
+    """
+    value = getattr(spec, attr, None)
+    if value is None:
+        inner_specs = getattr(spec, "kv_cache_specs", None)
+        if inner_specs:
+            first_inner = next(iter(inner_specs.values()))
+            value = getattr(first_inner, attr, None)
+    return int(value or 0)
+
+
+def _per_layer_spec_attr_from_vllm(
+    kv_cache_config: Any,
+    kv_caches: Mapping[str, Any],
+    attr: str,
+) -> list[int] | None:
+    """Map an integer ``KVCacheGroupSpec.kv_cache_spec`` attribute onto layers.
+
+    Reads ``attr`` off each engine group's spec (see :func:`spec_int_attr`) and
+    assigns it to every registered layer in that group, indexed by registration
+    order.
+
+    Args:
+        kv_cache_config: vLLM ``KVCacheConfig``. ``None`` or one without
+            ``kv_cache_groups`` yields ``None``.
+        kv_caches: Registered KV tensors keyed by layer name, in registration
+            order, providing the layer-name -> index mapping.
+        attr: The integer spec attribute to read (e.g. ``"block_size"`` or
+            ``"sliding_window"``).
+
+    Returns:
+        A list of length ``len(kv_caches)`` giving each layer's value, or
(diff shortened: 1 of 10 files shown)


yoo-kumaneko:
Superseded by #3550, which contains this PR's commit as its base plus a follow-up that makes the LMCache server block-only (moves SWA-suffix trimming to the connector). Continuing review there.

PR #3550, opened 2026-06-05
[MP][DSV4] Make the LMCache server block-only (connector-side SWA-suffix trim)

yoo-kumaneko:
> **Status: Ready for review.** Validated end-to-end on DeepSeek-V4-Flash (see Validation below). Supersedes #3548, #3261, and #3564 — all closed in favor of this branch.

## What & why

DeepSeek-V4's hybrid KV cache manager exposes several KV groups with **different logical block sizes** (256 for the full-attention MLA group, 64/4/8 for the SWA / compressor-state groups) while `cache_config.block_size` collapses to their GCD. Two problems followed from teaching the LMCache server about that vLLM-side geometry:

1. The server derived each group's `compress_ratio` / `physical_chunk_size` from a single (GCD-collapsed) scalar, so SWA / state groups silently stored only a fraction of each chunk.
2. To do SWA-suffix-only offload, the server had to understand sliding windows, convert token offsets to per-group blocks, and reshape/gather the trailing window on the hot path — error-prone (the "shape-invalid" landmine) and a layering violation: a KV **store** shouldn't need to model the engine's paging semantics.

This PR makes the server **block-only**. It receives, per group:
- already-**trimmed** block IDs (SWA groups carry only their trailing-window blocks), and
- one block count (`storage_blocks_per_chunk`) plus a per-group block skip.

It never reasons about sliding windows, logical block sizes, compress ratios, or token offsets on the transfer path. All of that now lives in the connector (scheduler side), in plain Python.

## Commit 1 — per-group logical block size + SWA-suffix offload (was #3548)

Fixes the correctness bug by plumbing per-group logical block sizes so the server derives each group's geometry correctly, and adds SWA-suffix-only offload (store/retrieve only the trailing window of each sliding-window group). Validated live on `DeepSeek-V4-Flash` tp=4: L1 dropped ~17× (667 MB → 38.7 MB) with cold≈warm semantically identical.

## Commit 2 — make the server block-only (this refactor)

**Connector (scheduler side):**
- `_slice_block_ids` trims each chunk to its trailing SWA window before sending.
- The retrieve APC skip is emitted as `skip_blocks_per_group` (per engine group, in each group's stored-block units) instead of a scalar token count — the connector folds in the token→block conversion and the suffix-offset adjustment, so the server applies it verbatim.
- A shared `storage_blocks_per_chunk` helper is the single source of the suffix formula for both the scheduler trim and the worker register hint (drift guard).

**Wire / hints:**
- New `per_layer_storage_blocks_per_chunk` layout hint — the one per-group block count the server uses for buffer sizing, kernel chunk-size, and the underflow check.
- `per_layer_sliding_window` removed from the wire.
- `per_layer_inference_engine_logical_block_size` kept, but now consumed **only** by the `cache_size_per_token` reporting metric, not the hot path.
- RETRIEVE op/protocol field `skip_first_n_tokens: int` → `skip_blocks_per_group: list[int]`.

**Server:**
- `KVLayerGroupInfo` collapses `blocks_per_chunk` and `storage_blocks_per_chunk` into one stored field (the distinction was an artifact of server-side trimming).
- Deletes the SWA machinery: `sliding_window`, `num_suffix_blocks_per_chunk`, `chunk_suffix_offset_blocks`, the retrieve-loop `view/reshape` suffix gather, and the token→block skip conversion with its alignment warning. Buffer/layout shapes derive directly from `storage_slots_per_chunk` physical slots.

The non-GPU CPU/SHM scatter path and the legacy single-group connectors convert the group-0 block skip back to a token prefix at the boundary, leaving the token-based scatter helper unchanged.

## Validation

**Live end-to-end** — `DeepSeek-V4-Flash` tp=4, fp8, block_size=256, chunk_size=1024, LMCache 256 GiB L1 (SHM), `--no-enable-prefix-caching`:

| 2026-06-05 (block-only path, first run) | 2026-06-06 (re-verified) |
|---|---|
| `l1_memory_usage_bytes` **58 MB** (vs 667 MB full-chunk) | store gate not wedged; L1 grows `0→19→77→97 MB` across chunks |
| `lookup_hit / requested` **3072 / 5120 = 60%** | warm lookup hits the full stored prefix (2048 / 2048 newly hit) |
| vLLM external prefix cache hit **32.2%** | **30.4%** |
| Cold ≈ warm ✅ semantically identical | ✅ semantically identical |
| No kernel errors | No `blocks_per_object * block_size == lmcache_chunk_size` crash |

The reduced L1 reflects SWA-suffix trimming working: only trailing-window blocks of each SWA/state chunk are stored.

**Unit tests** — 102 V4-relevant tests green (`tests/v1/test_kv_layer_groups_manager.py`, `test_kv_cache_groups.py`, `test_vllm_kv_cache_groups.py`, `test_mp_connector_slicing.py`, `test_vllm_mp_adapter.py`, `test_mp_mem_kernels.py`). New `tests/v1/test_mp_connector_slicing.py` covers the connector geometry / per-chunk trim / per-group skip arithmetic (it already caught a real `NameError` that would have crashed every hybrid store/retrieve). The `kv_layer_groups_manager` and `vllm_kv_cache_groups` suites are reworked around the storage-blocks hint.

## Not yet exercised
- Broader failure-mode coverage (APC-on, concurrency / multi-request, V3.2 regression).
- L2 (NIXL / FS) path on V4 HMA.

## Note on direction
This intentionally adds one server hint (`per_layer_storage_blocks_per_chunk`) while **removing** all server-side SWA / logical hot-path logic — a net-simpler server and a cleaner layering boundary (the store no longer models vLLM paging). Flagging the wire-protocol change (`skip` field; register hint) for maintainer review.

### DIFF
--- a/lmcache/cli/commands/bench/server_bench/helpers.py
+++ b/lmcache/cli/commands/bench/server_bench/helpers.py
@@ -362,7 +362,7 @@ def _send_retrieve(
         _INSTANCE_ID,
         [block_ids] * num_group_views,
         _make_event_handle(),
-        0,  # skip_first_n_tokens
+        [],  # skip_blocks_per_group
     ]
     result = _call(client, RequestType.RETRIEVE, payloads)
     if result is _TIMEOUT:
--- a/lmcache/integration/tensorrt_llm/tensorrt_mp_adapter.py
+++ b/lmcache/integration/tensorrt_llm/tensorrt_mp_adapter.py
@@ -414,7 +414,7 @@ class LMCacheMPKvConnectorWorker(KvCacheConnectorWorker):
                         self._instance_id,
                         [spec.block_ids],
                         event.ipc_handle(),
-                        0,  # skip_first_n_tokens
+                        [],  # skip_blocks_per_group
                     ],
                 ).result(timeout=self._mq_timeout)
             except Exception as e:
--- a/lmcache/integration/vllm/kv_cache_groups.py
+++ b/lmcache/integration/vllm/kv_cache_groups.py
@@ -98,3 +98,75 @@ def create_group_views_from_vllm(
             per_layer_group_idx,
         )
     ]
+
+
+def spec_int_attr(spec: Any, attr: str) -> int:
+    """Read an integer ``attr`` off a vLLM KVCacheGroupSpec spec.
+
+    Falls back to the first inner spec for ``UniformTypeKVCacheSpecs`` wrappers.
+    Returns ``0`` when the attribute is absent or ``None``.
+    """
+    value = getattr(spec, attr, None)
+    if value is None:
+        inner_specs = getattr(spec, "kv_cache_specs", None)
+        if inner_specs:
+            first_inner = next(iter(inner_specs.values()))
+            value = getattr(first_inner, attr, None)
+    return int(value or 0)
+
+
+def storage_blocks_per_chunk(
+    sliding_window: int,
+    logical_block_size: int,
+    chunk_tokens: int,
+) -> int:
+    """Blocks per chunk a group actually stores/retrieves (SWA-suffix-aware).
+
+    Returns the full chunk for full-attention groups (``sliding_window <= 0``),
+    or ``ceil(sliding_window / logical_block_size)`` capped at the full chunk
+    for sliding-window groups. Shared by the connector's scheduler-side trim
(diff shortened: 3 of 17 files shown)


yoo-kumaneko:
This feature has been implemented in other PRs.

PR #3558, opened 2026-06-05
[Misc] align MP server id with OTel service.instance.id

ruizhang0101:

The MP server's coordinator membership id and its OTel `service.instance.id` could diverge. `init_observability` minted a UUID into a local variable that was never written back to the config, so the coordinator registration read a still-`None` id, sent empty, and the coordinator assigned a *different* UUID. Metrics/traces and coordinator state then keyed on different ids and couldn't be correlated.

Now `run_cache_server` generates one identity at startup (`MPServerConfig.instance_id`), uses it as the coordinator membership key, and projects it onto `service.instance.id` unless observability set that value explicitly. The id is internal-only (no operator flag).

Single source of truth is `MPServerConfig.instance_id`. The standalone trace CLI path (no MP server) still falls back to its own UUID in `init_observability`.

**If applicable**:


ApostaC (commented):
We keep only one `--instance-id` config item. Default uuid4

`service_instance_id = instance_id`

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM

PR #3563, opened 2026-06-06
feat: add POSIX SHM infra for CPU KV-cache IPC

maobaolong:
## Summary

Add the POSIX shared-memory infrastructure that underpins the CPU KV-cache IPC path.

This is extracted from #3352 as a standalone, self-contained piece.

### DIFF
--- a/lmcache/v1/multiprocess/posix_shm.py
+++ b/lmcache/v1/multiprocess/posix_shm.py
@@ -99,12 +99,14 @@ def _open_and_mmap(name: str, nbytes: int, *, create: bool) -> tuple[_mmap.mmap,
 def _addr_of_mmap(mm: _mmap.mmap) -> int:
     """Return the base address of an ``mmap`` without leaking a buffer view.
 
-    A ctypes buffer view is created just long enough to read the
-    address, then dropped before this function returns; once it is
+    A single-byte ctypes view is created just long enough to read the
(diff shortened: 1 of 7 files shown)


yoo-kumaneko (approved):
LGTM. One thing worth a look:

In `migrate_to_shm_and_wrap`, the SHM segment is sized to `numel * elem_size` but `tensor.set_()` is called with `storage_offset` — so if that's ever nonzero the allocation is too small and both the write and the `to_tensor()` read go OOB. Fine for now since `attempt_permute_to_contiguous_view` always produces a zero-offset tensor in practice, but a quick assert would make that assumption explicit and cheap to catch later.

lmcache/v1/multiprocess/posix_shm.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    if not addr:
+        return
+    with _REGISTRY_LOCK:
+        mm = _ADDR_TO_MMAP.pop(addr, None)
+    if mm is None:
+        return
+    try:
+        mm.close()
+    except (BufferError, ValueError):
```
chunxiaozheng: @maobaolong the all catched exceptions only use `pass`, there is a minor question: should we add some warn logs?

chunxiaozheng (approved):
LGTM!

PR #3564, opened 2026-06-06
[MP][DSV4] Replace compress_ratio with per_layer_storage_blocks_per_chunk

yoo-kumaneko:

The existing `compress_ratio` / `physical_chunk_size` derivation in
`KVLayerGroupsManager` infers how many physical slots a KV group has per
LMCache chunk from `inference_engine_logical_block_size`. This works for
the current two-tier (compressed / non-compressed) split but does not
generalise to per-group storage budgets (e.g. SWA-only suffix storage
coming in a follow-up PR).

This PR replaces that indirect derivation with a direct
`storage_blocks_per_chunk` field on `KVLayerGroupInfo`, populated at
registration time from a new `per_layer_storage_blocks_per_chunk` layout
hint. The server becomes a "pure block mover": it only needs to know how
many blocks to transfer per group per chunk, and nothing about logical
block sizes or compression ratios.

Changes:
- `KVLayerGroupInfo`: drops `compress_ratio` / `physical_chunk_size` /
  `_derive_compression_metadata`; adds `storage_blocks_per_chunk` +
  `storage_slots_per_chunk`.
- `gpu_context`: new `get_storage_kv_buffer_shape()` replaces
  `get_physical_chunk_size()`; tmp-buffer sizing unified through it.
- `gpu_transfer`: store and retrieve kernels read
  `group.storage_slots_per_chunk` directly.
- `kv_cache_groups.py`: new helpers `spec_int_attr`,
  `storage_blocks_per_chunk`, `per_layer_storage_blocks_per_chunk_from_vllm`
  compute the per-layer hint from vLLM group specs at registration time.
- `vllm_multi_process_adapter`: ships the hint in
  `layout_hints["per_layer_storage_blocks_per_chunk"]`.

For non-hybrid (non-V4) models the hint is absent and every group falls
back to `lmcache_logical_chunk_size // bs`, preserving existing behaviour.

`group_compress_ratios` is kept as a backward-compatible stub (returns
all 1s) to avoid breaking any external callers that read the status dict.
It can be removed in a later cleanup PR.

A follow-up PR will use this hint mechanism to implement SWA-suffix-only
storage for DeepSeek-V4's sliding-window groups (connector pre-slices the
block ID list and ships a smaller `storage_blocks_per_chunk` for those
groups).

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/kv_cache_groups.py
+++ b/lmcache/integration/vllm/kv_cache_groups.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 # Standard
+import math
 from collections.abc import Mapping
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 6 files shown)


yoo-kumaneko:
Closing in favor of #3550 (`feat/connector-side-swa-trim`).

This branch was an attempt to reshape the layout hint from `per_layer_storage_blocks_per_chunk: list[int]` into a per-engine-group `dict[int, int]`. The data-structure half landed, but it inadvertently dropped the connector-side per-group grain-translation machinery (`_group_ratio` / `_group_geometry` / per-group block-ID slicing) that DeepSeek-V4's mixed logical block sizes require. The result slices a uniform `chunk_size // logical_bs` for every group while still passing each group's own `storage_slots_per_chunk` to the transfer kernel, which violates the kernel invariant `num_blocks_per_object * shape_desc.bs == lmcache_chunk_size` for any group whose `storage_blocks_per_chunk` differs from group 0's (live crash: `blocks_per_object * block_size (256) must equal lmcache_chunk_size (1024)`).

#3550 keeps the server-side `storage_blocks_per_chunk` simplification **and** the connector-side per-group translation — the two halves of one design. Validated end-to-end on DeepSeek-V4-Flash (tp=4, fp8); 102 targeted unit tests green.

PR #3568, opened 2026-06-06
fix(nixl): create storage directory if it doesn't exist

JinuJeong:
Ensure file-based NIXL L2 adapters create the configured `file_path` directory at initialization when it doesn't exist, matching the behavior of `fs_l2_adapter.py`.

Previously, starting an adapter with a non-existent directory would raise `FileNotFoundError` on the first store/load.

**If applicable**:


maobaolong (approved):
@JinuJeong This PR looks very clear, lgtm.

sammshen (approved):
LGTM

Issue #3562, opened 2026-06-06
[good-sec-issue] SessionManager.cleanup_expired is never called in production


maobaolong:
## Bug / Missing Feature

`SessionManager.cleanup_expired()` in `lmcache/v1/multiprocess/session.py` is defined but never called anywhere in production code (only in tests).

This means sessions that are never explicitly removed via `remove()` will accumulate in `_sessions` indefinitely, causing a slow memory leak over time.

## Expected behavior

A background thread (or periodic hook) should call `cleanup_expired()` at a configurable interval so stale sessions are reclaimed automatically.

## Suggested fix

Start a daemon thread inside `SessionManager.__init__` that periodically calls `cleanup_expired()`, similar to how other background janitors are implemented in the codebase. The interval could default to something like 60 s and be configurable via constructor argument.

And suggest to use `PeriodicThread` wrapped and call `PeriodicThreadRegistry#register` to register current thread to the `PeriodicThreadRegistry`, so that we can use `/periodic-threads` http endpoint to show all the background thread information.

zhengchenyu:
Please assign to me!

maobaolong:
@zhengchenyu Thanks you

xiaoba17:
Hi, I noticed this issue has been assigned for a while but is now stale, and I couldn't find an active PR addressing it. Is it still being worked on? If not, I'd be happy to take it over. @maobaolong

zhengfeihe:
Hi ~ @xiaoba17 

I think it's fixed by https://github.com/LMCache/LMCache/pull/3570

I will close the issue

xiaoba17:
> Hi ~ [@xiaoba17](https://github.com/xiaoba17)
> 
> I think it's fixed by [#3570](https://github.com/LMCache/LMCache/pull/3570)
> 
> I will close the issue

@zhengfeihe Thanks for letting me know! I’ll look for another issue to work on.

PR #3570, opened 2026-06-06
feat: add thread to clean up expired sessions for SessionManager.

zhengchenyu:

add thread to clean up expired sessions for SessionManager.

#3562 

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/engine_context.py
+++ b/lmcache/v1/multiprocess/engine_context.py
@@ -163,8 +163,10 @@ class MPCacheServerContext:
 
     def close(self) -> None:
         """
-        Tear down the storage manager and the process-global GDS context.
+        Tear down the session manager, storage manager, and the process-global
+        GDS context.
         """
+        self._session_manager.close()
         self._storage_manager.close()
         # Tear down the GDS cuFile context (the shared slab + its handle).
         get_gds_context().close()
--- a/lmcache/v1/multiprocess/session.py
+++ b/lmcache/v1/multiprocess/session.py
@@ -14,6 +14,13 @@ import time
 from lmcache.logging import init_logger
 from lmcache.v1.multiprocess.custom_types import IPCCacheServerKey
 from lmcache.v1.multiprocess.token_hasher import TokenHasher
+from lmcache.v1.periodic_thread import (
+    PeriodicThread,
+    PeriodicThreadRegistry,
+    ThreadLevel,
+    ThreadRunSummary,
+    create_periodic_thread,
+)
 
 logger = init_logger(__name__)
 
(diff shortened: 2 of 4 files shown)


ApostaC:
Thanks for your contribution! The PR looks good in general.

A high-level design question: do we want to have a new thread? Or do we want to check and clear the "expired" sessions during the creation of the new sessions? 

I feel like both will have some overhead, but not sure which one will be better.

Also cc @maobaolong

maobaolong:
> Thanks for your contribution! The PR looks good in general.
> 
> A high-level design question: do we want to have a new thread? Or do we want to check and clear the "expired" sessions during the creation of the new sessions?
> 
> I feel like both will have some overhead, but not sure which one will be better.
> 
> Also cc @maobaolong

@zhengchenyu How about your thoughts?

zhengchenyu:
@ApostaC I discuessed this with @maobaolong, and I think this scenario will frequently create and delete session, so it would be more appropriate to use a thread to clean up expired session.

maobaolong:
> @ApostaC I discuessed this with @maobaolong, and I think this scenario will frequently create and delete session, so it would be more appropriate to use a thread to clean up expired session.

@zhengchenyu Thanks for your share your thought, @ApostaC WDYT?

ApostaC (approved):
LGTM!

ApostaC:
A potential follow-up: can we also add a Prometheus metric to count the number of "manually cleaned sessions"?

maobaolong (approved):
LGTM. And @zhengchenyu Would you like to take the followup task ?

romain-primalabs:
Hello @ApostaC, you look like the expert for this. 

It seems that since this change was introduced on my side I noticed a lot of frequent (every 3-4 hours) prefix hit rate going to 0% following. I suspect this being related to this thread cleaning up by look at `created_at` instead of "last accessed".

https://github.com/zhengchenyu/LMCache/blob/424652228c0585ed3d49d76c3108c00bf07aca1b/lmcache/v1/multiprocess/session.py#L193

What do you think ?

maobaolong:
@romain-primalabs Good catch! I suspect that you environment is not a pd disaggregated deployment, the session will be cleaned while task's total duration exceeded 10 minutes. 

So we need a way to make ttl configurable rather than hardcode.

ApostaC:
@romain-primalabs Hey, thanks for the catch! It would be more helpful if you can share some repro. I suspect it could be some other issues. Feel free to reach out to us on slack as well!

ApostaC:
Btw, feel free to hard-code a new value in the code and see whether this problem can be fixed.

maobaolong:
> @romain-primalabs Good catch! I suspect that you environment is not a pd disaggregated deployment, the session will be cleaned while task's total duration exceeded 10 minutes.
> 
> So we need a way to make ttl configurable rather than hardcode.

https://github.com/LMCache/LMCache/pull/4091 This PR make session ttl configurable. Would you like to take a test

romain-primalabs:
Hello @maobaolong and @ApostaC , thanks for the reactivity! Really appreciate that. Let me try to join the slack to chat more about this issue.

PR #3583, opened 2026-06-08
[Bugfix] Fix L2 bench L1 alignment for raw-block O_DIRECT

3xdevv:

This PR fixes the `lmcache bench l2` data path so the benchmark uses an explicit L1 backing buffer with aligned per-object views instead of treating each object as an independent host allocation.

The main changes are:

- allocate a single aligned L1 buffer for one in-flight store wave and one in-flight load wave
- build `TensorMemoryObj` views from that buffer with aligned offsets and physical sizes so O_DIRECT-backed raw-block adapters receive valid L1 layout information
- advertise the intended L1 alignment through `L1MemoryDesc` instead of the tensor element size
- keep store/load verification semantics the same while clarifying that verification is optional and focused on byte-for-byte validation
- add unit coverage for aligned buffer creation, bounds checking, and the CLI bench runner wiring

Why this is needed:

The previous benchmark object construction could produce L1 metadata that did not match the buffer layout expected by raw-block O_DIRECT paths. That made the benchmark less representative and could trip alignment-sensitive adapters. Reusing a single aligned L1 buffer makes the benchmark data model match the adapter contract more closely.

- This PR is scoped to the CLI L2 adapter benchmark path and its tests/docs.
- The new `make_memory_objects(...)` contract now expects an L1 buffer and starting offset, and records object addresses relative to that L1 region.
- The benchmark still creates store/load object batches lazily, but the L1 capacity is reserved up front for both directions.

**If applicable**:

### DIFF
--- a/docs/source/cli/bench.rst
+++ b/docs/source/cli/bench.rst
@@ -1165,10 +1165,9 @@ round against the byte pattern that ``store`` wrote (see
    [Verify] Checking store -> load data integrity for last measured round...
    [Verify] OK
 
-Verification is **off** by default because the stricter byte pattern
-also forces every key to allocate its own ``data_size`` buffer
-(otherwise the runner is free to reuse a single shared buffer across
-keys to keep the memory footprint small).
+Verification is **off** by default so the benchmark can focus on L2
+adapter throughput unless byte-for-byte store/load validation is
+explicitly requested.
 
 
 Exit codes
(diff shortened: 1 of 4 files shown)


chunxiaozheng:
@3xdevv Thanks for your fix! I will take a look later.

chunxiaozheng:
@3xdevv Thanks for your contribution, but I saw a similar fix(https://github.com/LMCache/LMCache/pull/3603) today, which has already been merged and should have the same effect as the current PR.

PR #3584, opened 2026-06-08
[Feat][MP] Support Device-DAX as hybrid L1 cache overflow

DongDongJu:
                                                                                                                                            
This PR supports using Device-DAX memory as capacity expansion inside L1. 
In the hybrid configuration, LMCache allocates from DRAM L1 first  and then spills L1 allocations into Device-DAX before falling back to lower tiers.                                                          
                                                                                                                                            
The implementation keeps the existing Device-DAX allocator / adapter path. 
Users continue to configure explicit L1 and adapter sizes. 
When `--l1-devdax-path` matches a registered DAX adapter, 
LMCache infers that the adapter is L1 overflow capacity, consumes it into L1, and does not also register it as L2.     
                                                                                                                                            
This allows configurations such as:                                                                                                                                                                                                                                                     
- 30 GiB DRAM L1 only
- 30 GiB DRAM L1 + 30 GiB Device-DAX L1 overflow                                                                                            
- 30 GiB DRAM L1 + 30 GiB Device-DAX L2
- 30 GiB DRAM L1 + 30 GiB Device-DAX L1 overflow + 30 GiB raw-block L2 ...                                                                                             
                                                                                                                                            
The main intended scope of this PR is hybrid DRAM L1 + Device-DAX L1 overflow. DAX-only L1 remains possible when only `--l1-devdax-path` is set, and it uses the same `DevDaxMemoryAllocator` path, but the performance results below focus on the hybrid L1 use case.
                                                                                                                                            
The key behavior is that Device-DAX can now participate in the same L1 tier as DRAM. This is different from using Device-DAX as L2: L2 may retain more blocks for some workloads, but it uses the separate L2 adapter / storage path.
                                                                                                                                   
For GPU transfer, the Device-DAX mmap is registered with CUDA through `cudaHostRegister` when CUDA is available. When registration succeeds, the DAX tensor is treated as pinned host memory for GPU transfer. If registration fails, LMCache falls back to the safe staged path through pinned DRAM.                                                                                                                               
                                                                                                                                            
Capacity comparison note: hybrid L1 has 30 GiB DRAM L1 + 30 GiB Device-DAX L1 capacity. 
The pure DRAM row is a 30 GiB control. The DAX L2 and raw-block L2 rows are the closest aggregate-capacity comparison because they also use 30 GiB DRAM L1 plus a 30 GiB lower tier.

This version also addresses review feedback: 

Thanks @deng451e and @jayhpark530

- added missing type hints and config normalization docstring
- preserved `update_stats` when hybrid L1 frees primary DRAM allocations
- added cleanup through `DevDaxMemoryAllocator.__del__`
- changed the shared-memory-pool test to verify the public `MPCacheEngineContext.shm_pool_info` behavior instead of testing a private helper
 directly
- added coverage for CUDA host registration success and fallback

GPU <-> L1 host transfer bandwidth:

- GPU: NVIDIA H100 PCIe
- Torch: `2.12.0+cu130`
- Transfer size: 1 GiB
- Repeats: 7 after 2 warmups
- Device-DAX used for this microbenchmark: `/dev/daxX.Y`

```text
path                         direction  median_gbps  mean_gbps
Pinned DRAM reference        store D2H   55.54        55.54
Pinned DRAM reference        load H2D    55.62        55.62
Device-DAX registered        store D2H   47.59        47.59
Device-DAX registered        load H2D    50.05        50.05
Device-DAX staged fallback   store D2H   22.84        22.87
Device-DAX staged fallback   load H2D    25.96        25.90
```

vLLM e2e performance check:
- Benchmark: `lmcache bench engine --workload prefix-suffix-tuner`
- vLLM: `0.22.1`                                                      
- L1 DRAM size: 30 GiB
- Device-DAX / L2 size: 30 GiB
- Device-DAX used for e2e: `/dev/daxX.Y`
- Working sets tested: 20 GiB, 45 GiB, 75 GiB
- vLLM flags included `--max-model-len 8192`, `--enforce-eager`, `--no-enable-prefix-caching`, `--max-num-batched-tokens 8192`, and `--max-num-seqs 8`                                                            
- raw-block L2 baseline used file-backed `O_DIRECT` on local ext4
- `pass2_hit_pct` is based on LMCache lookup counters for the measured second pass
- 
Qwen/Qwen3-0.6B: 
```text                                                               
working_set_gib  mode                 success  input_tps  mean_ttft_ms  p90_ttft_ms  pass2_hit_pct
20               dram_l1              47/47    77975.61   54.63         55.78        81.2
20               hybrid_l1_devdax     47/47    67082.15   63.41         71.06        81.2
20               dax_l2               47/47    60777.15   69.77         73.29        81.2

45               dram_l1              107/107  64958.59   64.75         69.25        0.0
45               hybrid_l1_devdax     107/107  64457.39   66.09         72.17        12.9
45               dax_l2               107/107  39688.92   106.99        149.77       50.7
                                   
75               dram_l1              179/179  56976.59   74.03         77.19        0.0
75               hybrid_l1_devdax     179/179  65078.45   64.99         71.21        0.0
75               dax_l2               179/179  42642.43   99.32         153.37       27.2                                                   
```                                                                   
Qwen/Qwen3-14B:                                                       
                                                                      
```text                                                               
working_set_gib  mode                 success  input_tps  mean_ttft_ms  p90_ttft_ms  pass2_hit_pct
20               dram_l1              33/33    29723.59   142.65        145.77       81.2         
20               hybrid_l1_devdax     33/33    29949.11   141.85        144.36       81.2
20               dax_l2               33/33    29850.59   142.35        144.65       81.2
                                                                                                                                            
45               dram_l1              75/75    12190.39   350.29        354.96       0.0
45               hybrid_l1_devdax     75/75    13958.49   305.67        353.07       17.3
45               dax_l2               75/75    14024.69   304.35        352.92       43.3
                                                                                                                                            
75               dram_l1              125/125  12074.43   353.58        359.37       0.0
75               hybrid_l1_devdax     125/125  12219.75   349.46        352.62       0.0
75               dax_l2               125/125  13528.76   315.50        350.67       26.0
```                                                                                                                                         
                                                                      
Summary from the e2e run:

- For smaller model, hybrid L1 avoids the higher TTFT seen by the L2 paths at 45 GiB and 75 GiB working sets, but it is not faster than pure DRAM at every point.                                                                                                                                
- For larger model, hybrid L1 improves over the 30 GiB DRAM-only control at 45 GiB and 75 GiB working sets, while DAX L2 is competitive or better at larger working sets because it retains more second-pass hits.                                                               
- These results should be read as evidence that Device-DAX is useful as L1 capacity expansion, not as a claim that hybrid L1 always dominates every L2 configuration.
                                                                      
**If applicable**:                                                                                                                          
                                   

deng451e:
Overall this looks good and matches what we discussed: DAX as a capacity expansion within L1 DRAM, and the numbers are promising.

One thing to confirm: the GPU→DAX path still bounces through the CUDA context's pinned buffer since the DAX mapping isn't registered, so it's device→pinned-DRAM (GPU DMA) → DAX (CPU memcpy) rather than a direct DMA into DAX. Two questions:

1. What's the load/store bandwidth on this path alone?
2. A thought for later: should we enable GPU-direct via `cudaHostRegister`, with a safe fallback to this path?

jayhpark530:
I may be missing some context here, but I have one clarification question about the intended scope: the PR title/description focuses on Device-DAX as hybrid L1 overflow, and @deng451e also described this as “DAX as a capacity expansion within L1 DRAM.”

However, from the code path it looks like setting only `--l1-devdax-path` can still select `DevDaxMemoryAllocator` directly, resulting in a DAX-only L1 mode, without the DRAM-primary + DAX-overflow inference that happens when there is a matching DAX adapter.

Is DAX-only L1 also an intentionally supported mode in this PR, or should the supported configuration be understood as hybrid DRAM L1 + Device-DAX overflow only?

Given that DAX-only L1 could also be useful as a standalone mode in the future, one possible follow-up consideration could be its GPU transfer path. My understanding is that the current implementation exposes the DAX mapping as a CPU tensor via `mmap`/`torch.frombuffer`, so it may not have the same pinned-memory behavior as the DRAM L1 path. I wonder whether `cudaHostRegister` or a similar registration mechanism could be considered as a future optimization for this path.
This was also one of the points I had in mind from the earlier DAX backend PR and the related public issue discussion :)

DongDongJu:
Hello @deng451e, Thanks for chking in.

> Overall this looks good and matches what we discussed: DAX as a capacity expansion within L1 DRAM, and the numbers are promising.
> 
> One thing to confirm: the GPU→DAX path still bounces through the CUDA context's pinned buffer since the DAX mapping isn't registered, so it's device→pinned-DRAM (GPU DMA) → DAX (CPU memcpy) rather than a direct DMA into DAX. Two questions:
> 
> 1. What's the load/store bandwidth on this path alone?

will update the PR description including this.

> 2. A thought for later: should we enable GPU-direct via `cudaHostRegister`, with a safe fallback to this path?

I want to introduce it in next one but will update it including this one.

DongDongJu:
Hello @jayhpark530 

> I may be missing some context here, but I have one clarification question about the intended scope: the PR title/description focuses on Device-DAX as hybrid L1 overflow, and @deng451e also described this as “DAX as a capacity expansion within L1 DRAM.”
> 
> However, from the code path it looks like setting only `--l1-devdax-path` can still select `DevDaxMemoryAllocator` directly, resulting in a DAX-only L1 mode, without the DRAM-primary + DAX-overflow inference that happens when there is a matching DAX adapter.
> 
> Is DAX-only L1 also an intentionally supported mode in this PR, or should the supported configuration be understood as hybrid DRAM L1 + Device-DAX overflow only?

It intentionally support both, but without hybrid mode (support dax with other backends), it is not that meaningful. thats the reason why I didnt emphasize that much.

> 
> Given that DAX-only L1 could also be useful as a standalone mode in the future, one possible follow-up consideration could be its GPU transfer path. My understanding is that the current implementation exposes the DAX mapping as a CPU tensor via `mmap`/`torch.frombuffer`, so it may not have the same pinned-memory behavior as the DRAM L1 path. I wonder whether `cudaHostRegister` or a similar registration mechanism could be considered as a future optimization for this path. This was also one of the points I had in mind from the earlier DAX backend PR and the related public issue discussion :)

I do agree. I want to introduce it in next PR but let me just include this one into this one and update the PR description as well. Thanks for check!

DongDongJu:
I updated the PR @deng451e @ApostaC @jayhpark530 PTAL!

lmcache/v1/distributed/config.py
```diff
@@ -43,9 +95,27 @@ class L1MemoryManagerConfig:
...
+            self.devdax_path = self.devdax_path.strip()
+
+        if self.devdax_size_in_bytes < 0:
+            raise ValueError("devdax_size_in_bytes must be >= 0")
+        if self.devdax_size_in_bytes and not self.devdax_path:
+            raise ValueError("devdax_size_in_bytes requires devdax_path")
+
+        if self.devdax_path:
+            self.use_lazy = False
+            self.shm_name = ""
```
ApostaC: Instead of silently setting use_lazy = False, can we raise an error and explicitly require `--no-l1-use-lazy` to be set on the user side?

(Same for shm)

lmcache/v1/distributed/config.py
```diff
@@ -118,6 +188,41 @@ class StorageManagerConfig:
     """ Interval (ms) for the periodic event notifier heartbeat. """
 
 
+def normalize_storage_manager_config(config: StorageManagerConfig) -> None:
```
ApostaC, line 257: nit: we can call it `validate_storage_manage_config`. `normalize` sounds like it's changing the config, but seems like it's not.

And we can call it in `StorageManagerConfig.__post_init__`

lmcache/v1/distributed/memory_manager/l1_memory_manager.py
```diff
@@ -50,7 +51,37 @@ def create_memory_allocator(config: L1MemoryManagerConfig) -> MemoryAllocatorInt
...
+            "align bytes is %d bytes",
+            config.devdax_path,
+            config.size_in_bytes,
+            config.align_bytes,
+        )
+        return DevDaxMemoryAllocator(
+            config.size_in_bytes,
+            config.devdax_path,
+            align_bytes=config.align_bytes,
+        )
```
ApostaC: Here is a bit complicated. Can we have a unified DevDaxMemoryAllocator for both cases?

lmcache/v1/memory_management.py
ApostaC, line 1: Since MixedMemoryAllocator (I don't like its current name nor the implementation, btw) is widely used in a lot of places, can we avoid adding new complexities to it?

I would recommend maintaining a memory allocator instance in `DevDaxMemoryAllocator` for local memory allocations.

ApostaC (changes requested):
Did a pass. High-level concern is the change to the memory manager.

lmcache/v1/memory_management.py
DongDongJu: Sure, Let me work on the direction that having new memory allocator dealing managing both DAX and mixedmemoryallocator's behavior in same place not modifying the MixedMemoryAllocator itself.

lmcache/v1/distributed/config.py
DongDongJu: Done. I changed this to fail explicitly instead of silently overriding the config.
  Now `--l1-devdax-path` requires users to pass both:
  `--no-l1-use-lazy`
  `--no-l1-use-shm`
  This keeps the CLI behavior explicit and avoids hidden config mutation.

I will make the follow up PR for further optimization.

lmcache/v1/distributed/config.py
DongDongJu: Updated this path.

  There are now two steps called from StorageManagerConfig.__post_init__:

  `normalize_storage_manager_config(self)`
  `validate_storage_manager_config(self)`

  I kept normalize because it does mutate config in the hybrid case: when `--l1-devdax-path` matches a registered DAX adapter, it consumes that adapter into L1 overflow by setting `devdax_size_in_bytes` and removing that adapter from the L2 list. 
Validation then checks incompatible combinations.

lmcache/v1/distributed/memory_manager/l1_memory_manager.py
DongDongJu: Done. The allocation policy is now unified inside DevDaxMemoryAllocator like we discussed before.

  The L1 manager creates DevDaxMemoryAllocator for both DAX-only and hybrid cases. 

In the hybrid case, DevDaxMemoryAllocator owns a local MixedMemoryAllocator internally and tries local DRAM first, then spills to Device-DAX.

This keeps the policy in one allocator instead of splitting it in the L1 manager.

lmcache/v1/memory_management.py
DongDongJu: Done. I moved the hybrid behavior out of MixedMemoryAllocator.

  DevDaxMemoryAllocator now keeps the local DRAM allocator as an internal member and manages DRAM-first / DAX-overflow allocation itself. 
MixedMemoryAllocator is only used as the existing local allocator implementation, without carrying the Device-DAX policy.

DongDongJu:
@ApostaC fixed based on discussion. PTAL

dongkimn:
Hello @DongDongJu,
I would like to suggest the implementation of memory allocator in “memory_management.py”. Could you consider separating the class level memory allocators such as "class DevDaxMemoryAllocator" or “GPUMemoryAllocator” into module level like "DevDaxMemoryAllocator.py" or “GPUMemoryAllocator.py” ? The reason is to keep responsibilities isolated so each memory allocator remains maintainable and scalable as it grows without dependency when the new features might be added.

DongDongJu:
> Hello @DongDongJu, I would like to suggest the implementation of memory allocator in “memory_management.py”. Could you consider separating the class level memory allocators such as "class DevDaxMemoryAllocator" or “GPUMemoryAllocator” into module level like "DevDaxMemoryAllocator.py" or “GPUMemoryAllocator.py” ? The reason is to keep responsibilities isolated so each memory allocator remains maintainable and scalable as it grows without dependency when the new features might be added.

Hello @dongkimn, Sounds great idea. I will visit this issue after this one merged.

lmcache/v1/distributed/config.py
```diff
@@ -43,9 +98,37 @@ class L1MemoryManagerConfig:
...
+            raise ValueError(
+                "l1-devdax-path requires lazy allocation to be disabled. "
+                "Please set --no-l1-use-lazy."
+            )
+        if self.devdax_path and self.use_shm:
+            raise ValueError(
+                "l1-devdax-path requires SHM to be disabled. "
+                "Please set --no-l1-use-shm."
+            )
+
```
ApostaC: We don't need to introduce the new `l1_use_shm` here. We can raise when `self.devdax_path and self.shm_name`

ApostaC (commented):
Just a small comment on the CLI changes. Otherwise LGTM!

lmcache/v1/distributed/config.py
ApostaC: My bad. I previously thought shm is also a CLI arg that needs something like `--no-shm`. But I was wrong. Please see my latest comments

lmcache/v1/distributed/l1_manager.py
```diff
@@ -815,6 +819,30 @@ def get_l1_memory_desc(self):
...
+    def get_devdax_reconfigure_status(self) -> L2ReconfigureStatus | None:
+        """Return L1 Device-DAX reconfiguration status when available."""
+        get_status = getattr(
+            self._memory_manager,
+            "get_devdax_reconfigure_status",
+            None,
+        )
+        if get_status is None:
+            return None
+        return get_status()
```
ApostaC: It's weird that a method in `L1Manager` returns `L2ReconfigureStatus`.

ApostaC (commented):
I see you are trying to add reconfigure for L1 devdax as well.
From code cleanliness's perspective, I would like to avoid adding device-specific functions in the L1 memory manager and the L1 manager. Ideally, the methods they provided should be minimized and generalized. And this is why we have a new GDSL1MemoryManager instead of adding GDS-specific functions in the existing L1MemoryManager class.

That said, for reconfiguration, we can design an L1-generic one, and have that in a separate PR. WDYT?

DongDongJu:
> I see you are trying to add reconfigure for L1 devdax as well. From code cleanliness's perspective, I would like to avoid adding device-specific functions in the L1 memory manager and the L1 manager. Ideally, the methods they provided should be minimized and generalized. And this is why we have a new GDSL1MemoryManager instead of adding GDS-specific functions in the existing L1MemoryManager class.
> 
> That said, for reconfiguration, we can design an L1-generic one, and have that in a separate PR. WDYT?

sry PR was not read yet.
cuz I also felt similar feeling.
Let me change it to draft. I will ping you when its ready.

DongDongJu:
@ApostaC @deng451e Dropped the l1 reconfigure support in this PR and focused on this PR description itself. PTAL.

ApostaC (approved):
LGTM!

deng451e (approved):
LGTM!

PR #3589, opened 2026-06-08
[core] Add GDS L1 slab-file tier (cuFile DMA) for MP mode

Shaoting-Feng:
Add gds l1 which can replace DRAM l1.

**If applicable**:


lmcache/v1/gpu_connector/gds_context.py
```diff
@@ -0,0 +1,455 @@
...
+        exactly once and keep it for the context's lifetime.
+        """
+        with self._stream_lock:
+            if self._registered_stream_handle is not None:
+                return
+            current_stream = torch_dev.current_stream()
+            raw_stream = current_stream.cuda_stream
+            ca.register_stream(raw_stream)
+            self._registered_stream_handle = raw_stream
+            self._registered_stream = current_stream
```
ApostaC: What's the behavior when there are more than 2 GPUCacheContexts? (E.g. TP > 1)

lmcache/v1/gpu_connector/_cufile_async.py
```diff
@@ -0,0 +1,368 @@
...
+
+        sub = Submission(size=size, file_offset=file_offset, buf_offset=buf_offset)
+        _check(
+            libcufile.cuFileReadAsync(
+                self._handle,
+                ctypes.c_void_p(buf_base),
+                ctypes.byref(sub._size),
+                ctypes.byref(sub._file_offset),
+                ctypes.byref(sub._buf_offset),
+                ctypes.byref(sub._bytes_done),
```
ApostaC, line 331: Why do we need to use `byref` here?
If it's not `byref`, we can probably get rid of `Submission`.

lmcache/v1/gpu_connector/gds_context.py
ApostaC, line 1: Check whether the functions are noop when `initialized=False`

lmcache/v1/gpu_connector/_cufile_async.py
ApostaC, line 1: maybe move `_cufile_async.py` and `gds_context.py` into `lmcache/v1/gpu_connector/`

tests/v1/gpu_connector/test_gds_context.py
```diff
@@ -0,0 +1,147 @@
...
+class TestInitializeRounding:
+    def test_slab_size_rounded_up_to_alignment(self, monkeypatch, tmp_path):
+        ctx = GDSContext()
+        monkeypatch.setattr(ctx, "_open_and_register_slab", lambda use_direct_io: None)
+        ctx.initialize(
+            GdsL1Config(file_location=str(tmp_path), size_in_bytes=3 * 4096 + 1)
+        )
+        assert ctx.initialized is True
+        assert ctx._slab_size == 4 * 4096
+        assert ctx._slab_path.endswith("lmcache_gds_slab.bin")
```
ApostaC: Avoid assertions on private members in a class. (Or use private members in the class)

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/gpu_connector/_cufile_async.py
Shaoting-Feng: [`cuFileReadAsync`/`WriteAsync`](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufilereadasync) take `size`/`file_offset`/`buffer_offset` as pointers, read when the stream executes, not at submit. [`cuFileStreamRegister`](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufilestreamregister) flag `0x7` is supposed to make them submit-time, but a unit test shows they're still read at execution time (cuFile 1.17.0 + 1.18.0).

So the param storage must stay alive until the op completes; we can't free or reuse it at submit.

lmcache/v1/gpu_connector/gds_context.py
Shaoting-Feng: - `register_gpu_buffer` — no-op when `initialized=False`.
- `read_async` / `write_async` — not reached when off: the `gpu_ops` call site only invokes them for a `GDSMemoryObject`. If they ever were called uninitialized they raise.
- `close` — effectively a no-op when off.

Let me know if any of these guarded with an explicit `if not self.initialized: return` is better.

lmcache/v1/multiprocess/gpu_context.py
```diff
@@ -391,6 +402,14 @@ def __init__(
...
         # GPU streams
         self.cuda_stream_ = torch_dev.Stream(device=self.device_)
+
+        # Register the staging buffer with the GDS cuFile context on the
+        # context's CUDA stream.
+        with torch_dev.stream(self.cuda_stream_):
+            get_gds_context().register_gpu_buffer(
+                self._temp_buffer.buffer, self._temp_buffer.single_batch_bytes
+            )
+
```
ApostaC, line 405: This seems incorrect. We should register the buffer with its full size instead of `single_batch_bytes`

ApostaC (commented):
One quick comment

lmcache/v1/multiprocess/gpu_context.py
Shaoting-Feng: Previously, we register per-chunk because cuFile caps a single cuFileBufRegister at 16 MiB, and the full buffer exceeds that. Now refactor to register as a whole and manages the 16 MiB capacity inside the gds context.

Shaoting-Feng: The kernel launch time is also reduced if we only register once.
Screenshot 2026-06-09 at 5 22 17 PM

Shaoting-Feng:
Tested read bandwidth ~5.33 GiB/s on tmp server.

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3593, opened 2026-06-08
[MP] Add support for OBJ backend and batched operations in NIXL dynamic L2 adapter

ziruiliu:

This PR adds OBJ backend support to the NIXL dynamic L2 adapter for LMCache MP mode.

The existing s3_l2_adapter uses the S3 client library path that supports virtual-hosted style addressing only. For some on-prem object storage deployments, especially those using endpoint overrides and self-signed certificates, that path is difficult to use directly. In contrast, the in-process nixl_storage_backend already supports NIXL dynamic object storage access for put/get. This PR migrates that dynamic object-storage capability into the MP-mode nixl_store_dynamic L2 adapter so MP deployments can use NIXL OBJ as an L2 backend.

This PR also also introduced batched lookup/load/store operations to replace _execute_*_in_loop

Previously, MP-mode dynamic NIXL store/load built and submitted each object independently through _execute_*_in_the_loop() which impacts the prefill performance severely.

**What is changed**
Enabling OBJ backend
- Added OBJ as a valid backend for nixl_store_dynamic.
- Added dynamic NIXL object registration for store/load operations.
- Updated store/load paths to use:
    - file registration for file-like backends such as POSIX, GDS, GDS_MT, HF3FS
    - object registration for OBJ
- Added secondary lookup support for OBJ through NIXL object presence queries.
- Documented the difference between file semantics and object semantics in the adapter module docstring.

Enabling batched operations
 - Added DynamicStorageSpec and _DynamicStorageDesc to describe per-object/file transfer targets and temporary file metadata.
 - Added batched NIXL registration helpers:
     - _register_batched_files()
     - _register_batched_objects()
     - _release_batched_storage()

 - Added batched transfer APIs in the dynamic storage agent:
     - dynamic_store_batch()
     - dynamic_load_batch()

**Eviction and Usage**

OBJ support intentionally does not implement adapter-side usage accounting or eviction.

For file backends, the adapter can recover file size through os.stat() and can delete files with os.unlink(). That makes local usage tracking and eviction practical.

For OBJ, NIXL is primarily a transfer library. It supports object transfer operations, but it does not expose a general backend-neutral object deletion or object-size query interface through this adapter path. Because of that:
- max_capacity_bytes is set to 0 for OBJ
- supports_global_eviction is false
- delete() is a no-op for OBJ
- usage fraction remains unsupported

Object lifecycle management should be handled by the object storage system or deployment policy.

**If applicable**:


lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -60,20 +70,30 @@
...
+    """Derive a deterministic storage name from an ObjectKey.
 
     Replaces ``/`` in model names with ``--`` to avoid creating
     subdirectories (e.g. ``meta-llama/Llama-3-8B`` becomes
     ``meta-llama--Llama-3-8B``).
     """
     safe_model_name = key.model_name.replace("/", "--")
     chunk_hex = key.chunk_hash.hex()
-    return f"{safe_model_name}_{key.kv_rank:08x}_{chunk_hex}.bin"
+    return f"{safe_model_name}_{key.kv_rank:08x}_{chunk_hex}"
```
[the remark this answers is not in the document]
ziruiliu: 👍 , resolved. Add cache_salt to both object and file naming scheme

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -169,11 +192,47 @@ def _register_single_file(self, fd: int, file_size: int, page_size: int):
         )
         return reg_descs, xfer_handler
 
+    def _alloc_device_id(self) -> int:
+        """Allocate a unique OBJ device id for one register/deregister cycle."""
+        with self._device_id_lock:
+            device_id = self._device_id_counter
+            self._device_id_counter += 1
+        return device_id
```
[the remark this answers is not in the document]
ziruiliu: hardcode 1000000 is not a good choice. I have added a macro in both _alloc_device_id and _alloc_device_ids

maobaolong:
@ziruiliu Thanks for this PR, will take a look this week.

ziruiliu:
@YaoJiayi could you please also take a look? Current nixl storage adapter only supports file like backend and has to be extended to support OBJ backend

ziruiliu:
This branch has to merge changes in #3696  which brings parallel file loading. Working on it

ziruiliu:
Updated this branch on top of latest upstream dev.

This also revises the code introduced by PR #3696. That PR added concurrent per-file load using asyncio/gather and left a TODO to batch a request's files into one native NIXL transfer. This update completes that TODO by using the native batched NIXL operation interface, making the dynamic adapter behave similarly to the static adapters.

One behavior difference is intentional: NIXL currently exposes only an aggregated status for a batched transfer, not per-file/object success. Therefore, if a native batched load fails, the adapter treats the whole batch as failed rather than marking partial per-key success.

sihara:
> Updated this branch on top of latest upstream dev.
> 
> This also revises the code introduced by PR #3696. That PR added concurrent per-file load using asyncio/gather and left a TODO to batch a request's files into one native NIXL transfer. This update completes that TODO by using the native batched NIXL operation interface, making the dynamic adapter behave similarly to the static adapters.
> 
> One behavior difference is intentional: NIXL currently exposes only an aggregated status for a batched transfer, not per-file/object success. Therefore, if a native batched load fails, the adapter treats the whole batch as failed rather than marking partial per-key success.

@ziruiliu  
Tested this PR on top of `0.5.2-dev` (which already includes #3696), single GH200, Qwen3-14B, vLLM 0.24, NIXL 1.3.0.

**Good news: at `chunk_size=1024` it's a real win at mid-concurrency.**

TTFT p50 (ms):

| cons | 0.5.2-dev | +#3593 |
|---|---|---|
| 4 | 354 | **296** |
| 8 | 558 | 565 |

TTFT p99 (ms):

| cons | 0.5.2-dev | +#3593 |
|---|---|---|
| 4 | 437 | **413** |
| 8 | 809 | 819 |

At cons=4, #3593 clearly lowers TTFT. At cons=8 both are limited by storage/fabric rather than the adapter, so there's no gain (TTFT is marginally higher, within noise).

**Problem: at smaller `chunk_size` (which means more chunks per request) the native batched load breaks in the NIXL transfer.** A smaller `chunk_size` splits each request's KV into more chunks, so the batched transfer carries more files. At `chunk_size <= 512` — including the **default 256** — the single batched `postXferReq` is rejected:

```
io_uring_io_queue.cpp:153] No more free blocks available
posix_backend.cpp:230] Error preparing I/O operation: -6
nixl_agent.cpp:1103] postXferReq: backend 'POSIX' failed to post the transfer request with status NIXL_ERR_NOT_ALLOWED
  File ".../nixl_store_dynamic_l2_adapter.py", line 568, in dynamic_load_batch
    await self._post_non_blocking(handle)
nixl._bindings.nixlNotAllowedError: NIXL_ERR_NOT_ALLOWED
```

The batch flattens all of a request's chunks into one transfer; once the batch is large enough (roughly >= 59 chunks/request, i.e. `chunk_size <= 512`), NIXL's POSIX/io_uring backend reports *No more free blocks available* and `postXferReq` is refused. With the intentional whole-batch-fail semantics, the request then misses entirely (no L2 read) and recomputes on GPU. It's deterministic. Threshold: works at ~53 chunks/request, fails at ~59.

TTFT (ms), cons=4, by `chunk_size`:

| chunk_size | ~chunks/req | 0.5.2-dev p50 | +#3593 p50 | 0.5.2-dev p99 | +#3593 p99 |
|---|---|---|---|---|---|
| 1024 | 30 | 360 | **296** | 424 | **397** |
| 512 | 59 | 387 | **5933** | 496 | **6023** |
| 256 (default) | 118 | 393 | **6082** | 527 | **6187** |

At `chunk_size <= 512` #3593's TTFT jumps ~15-20x (full recompute), vs a small improvement at 1024. (The `chunk_size=1024` baseline reads 354 above and 360 here because the two tables are separate benchmark runs — run-to-run variance.)

This might need to split the descriptor list into bounded sub-batches (e.g. <= 48) per `make_prepped_xfer`, or fall back to per-chunk transfer above the limit — that would keep the cons=4 win without regressing the default `chunk_size`. Does that sound reasonable?

ziruiliu:
> 
> The batch flattens all of a request's chunks into one transfer; once the batch is large enough (roughly >= 59 chunks/request, i.e. `chunk_size <= 512`), NIXL's POSIX/io_uring backend reports _No more free blocks available_ and `postXferReq` is refused. With the intentional whole-batch-fail semantics, the request then misses entirely (no L2 read) and recomputes on GPU. It's deterministic. Threshold: works at ~53 chunks/request, fails at ~59.
> 
> TTFT (ms), cons=4, by `chunk_size`:
> 
> chunk_size	~chunks/req	0.5.2-dev p50	+#3593 p50	0.5.2-dev p99	+#3593 p99
> 1024	30	360	**296**	424	**397**
> 512	59	387	**5933**	496	**6023**
> 256 (default)	118	393	**6082**	527	**6187**
> At `chunk_size <= 512` #3593's TTFT jumps ~15-20x (full recompute), vs a small improvement at 1024. (The `chunk_size=1024` baseline reads 354 above and 360 here because the two tables are separate benchmark runs — run-to-run variance.)
> 
> This might need to split the descriptor list into bounded sub-batches (e.g. <= 48) per `make_prepped_xfer`, or fall back to per-chunk transfer above the limit — that would keep the cons=4 win without regressing the default `chunk_size`. Does that sound reasonable?

Agree. There should be a limit of number of descriptors in a NIXL transfer. I will try to add some code addressing this problem

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -586,75 +1042,99 @@ def _signal_lookup_event(self) -> None:
...
                     if key in self._memory_objects or key in self._inflight_stores:
                         continue
-                    if self._total_bytes + mem_size > self._max_capacity_bytes:
-                        logger.warning(
-                            "Storage capacity exceeded, skipping store for key %s",
-                            key,
-                        )
-                        break
+                    if self.nixl_agent.mem_type == "FILE":
+                        if self._total_bytes + mem_size > self._max_capacity_bytes:
```
maobaolong, line 1072: Why we only check over `_max_capacity_bytes` for `FILE` `mem_type`?

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -172,11 +223,47 @@ def _register_single_file(self, fd: int, file_size: int, page_size: int):
         )
         return reg_descs, xfer_handler
 
+    def _alloc_device_id(self) -> int:
+        """Allocate a bounded OBJ device id for one register/deregister cycle."""
+        with self._device_id_lock:
+            device_id = self._device_id_counter
+            self._device_id_counter = (self._device_id_counter + 1) % MAX_OBJ_DEVICE_ID
```
maobaolong, line 230: It seems that the `_device_id_counter` will go back to 0 after it over `MAX_OBJ_DEVICE_ID`, could you explain more about this to help me understand the background?

maobaolong:
@ziruiliu A high level suggestion, but it can be improved in a followup PR. I found that there are two mem_type `OBJ` and `FILE`, before this PR, it support `FILE` only, but now, this PR add `if` condition check to determine the code for `OBJ` or `FILE`, the current code seems a little complex. Could we turn File and Object into two separate strategies or sub-implementations?

This would make the upper-level Nixl storage adapter much cleaner. By treating Object and File simply as two distinct implementations of the underlying differences, we can encapsulate that logic within those two specific classes.

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
ziruiliu: Because NIXL library does not provide delete/remove interface. For files you can see current code uses os.delete() to delete, but for objects this adapter is unable to evict or remove. So for remote S3 object we'd better do not implement quota and eviction

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -261,6 +348,307 @@ def dynamic_delete_file(self, file_path: str) -> None:
...
+        page_size: int,
+    ) -> None:
+        """Write a batch of objects/files with one NIXL transfer."""
+        if not specs:
+            return
+
+        mem_indices = [idx for spec in specs for idx in spec.mem_indices]
+        storage_indices = list(range(len(mem_indices)))
+
+        if self.mem_type == "OBJ":
```
maobaolong, line 502: If we could treat Object and File as two separate subclasses of a same parent class or two policies of current class, we might be able to eliminate these if-else statements.

Correspondingly, you would redefine similar methods to implement the specific differences between Object and File, while keeping the interface layer consistent.

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -261,6 +348,307 @@ def dynamic_delete_file(self, file_path: str) -> None:
...
+        page_size: int,
+    ) -> None:
+        """Read a batch of objects/files with one NIXL transfer."""
+        if not specs:
+            return
+
+        mem_indices = [idx for spec in specs for idx in spec.mem_indices]
+        storage_indices = list(range(len(mem_indices)))
+
+        if self.mem_type == "OBJ":
```
maobaolong, line 548: Following up on the above, could we define an abstract method for the parts that differ? 

For example, we could call it registerBatch. For objects, the implementation would be _register_batched_objects. For files, it could first call _open_batched_file_descs and then _register_batched_files.

This way, the file-related code will be in the file implementation subclass, and the object-related code will be in the object implementation subclass. Meanwhile, the common parts of NixStoreAdapter will reside in the shared code file. Would that be much clearer?

lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
```diff
@@ -302,6 +739,8 @@ def cleanup_temp_files(self) -> None:
         GC; orphans don't affect correctness because they're never matched
         by the deterministic ``ObjectKey → filename`` mapping.
         """
+        if self.mem_type != "FILE":
```
maobaolong, line 742: ditto

ziruiliu:
Discussed with @maobaolong , there are indeed differences between OBJ and FILE types which makes code hard to understand. OBJ and FILE has following differences:


Type | FILE | OBJ
-- | -- | --
Store publication | temp file, then atomic rename | direct object write
Secondary lookup | os.stat | NIXL object query
Capacity accounting | known byte size | object size unavailable on lookup
Delete / eviction | supported via unlink | currently unsupported
Shutdown cleanup | can remove files | currently cannot remove objects

We have to figure out which part is public and which has to be divided into 2 sub classes

ziruiliu:
I would file another PR to enable similar path-style s3 access and self-signed CA in s3 l2 adapter first

ofer:
> Discussed with @maobaolong , there are indeed differences between OBJ and FILE types which makes code hard to understand. OBJ and FILE has following differences:
> 
> Type	FILE	OBJ
> Store publication	temp file, then atomic rename	direct object write
> Secondary lookup	os.stat	NIXL object query
> Capacity accounting	known byte size	object size unavailable on lookup
> Delete / eviction	supported via unlink	currently unsupported
> Shutdown cleanup	can remove files	currently cannot remove objects
> We have to figure out which part is public and which has to be divided into 2 sub classes

I have a branch that implements the difference as a polymorphic setup to make it a little clearer where the lines are drawn, I have not yet set it up as a PR.

https://github.com/ofer/LMCache/tree/feature/add-obj-support-to-l2-nixl-dynamic

ziruiliu:
This PR has been created so long. There are too many changes in LMCache and in NIXL, e.g. path mode in file backend. I will close this PR.

PR #3597, opened 2026-06-08
[Feat][Coordinator] Global CacheBlend fingerprint directory on the MP coordinator

ruizhang0101:

CacheBlend lookup is local to one mp-server today, so a request routed to another
replica recomputes KV a peer already holds. This adds an **opt-in, fleet-wide
CacheBlend fingerprint directory on the MP coordinator**: blend servers publish
stored ranges on STORE and query request tokens on LOOKUP; the coordinator runs
the match and returns shared-L2 storage keys to prefetch + re-RoPE.

Local and coordinator matching are **mutually exclusive by coordinator
presence**: with a coordinator configured, all fingerprint matching goes through
it (the fleet directory is a superset of the local table) and the local matcher
is skipped; with none, matching is purely local and unchanged.

- **Coordinator owns hashing + matching** (servers send raw tokens + storage
  mapping), so the algorithm can evolve coordinator-only. Same algorithm as the
  local matcher: poly-hash table keyed by `(model_scope, poly_hash)`, strided
  rolling-hash probe; scope `model_name@cache_salt`; TP resolved at retrieve.
- **REST:** `POST`/`DELETE /blend/fingerprints`, `POST /blend/match`.
  `GlobalBlendMatcher` on `app.state`; `mp_coordinator/blend_client.py` (mirrors
  `registrar.py`) bridges the blend module — best-effort publish + non-blocking
  submit/poll match, bounded by a per-lookup deadline (default 50 ms; a slow or
  absent coordinator → no fleet matches that lookup).
- Fleet matches are returned as `CBMatchResult` and **merged into
  `CBUnifiedLookupResult.non_prefix_segments`** (prefix-covered dropped,
  leftmost-greedy overlap dedup), so they ride the existing sparse-prefetch +
  retrieve + re-RoPE path with no protocol change.

Design doc: `docs/design/v1/mp_coordinator/blend_lookup.md`.

- Eviction is lazy (stale entry → wasted prefetch → recompute, never wrong KV);
  the `DELETE` route + `remove` are wired/tested for a future eager hook.
- `register` skips (logs error) a range whose chunk count ≠ `object_keys` count
  rather than truncating (a shift would mis-map KV).
- Full `cb_unified_lookup`/`store` path needs the native ext (CI/GPU); the
  coordinator, matcher, client, hashing, and schemas are unit-tested here.

**If applicable**:

- [ ] this PR contains user facing changes - docs added (design doc)

### DIFF
--- a/docs/design/v1/mp_coordinator/blend_lookup.md
+++ b/docs/design/v1/mp_coordinator/blend_lookup.md
@@ -1,0 +1,213 @@
+# Global CacheBlend Lookup (Coordinator Fingerprint Directory)
+
+A coordinator-level capability that lets a CacheBlend lookup on one mp-server
+find and reuse chunk KV cached anywhere in the fleet, fetched from a shared,
+content-addressed L2. Blend servers **publish fingerprints on STORE** and
+**query on LOOKUP**; the coordinator holds the fleet-wide directory and runs the
+match. It is **opt-in** (enabled only when a coordinator URL is configured) and
+**additive** (the existing local matcher fast path is unchanged).
+
+Code: `lmcache/v1/mp_coordinator/blend_directory.py`,
+`lmcache/v1/mp_coordinator/http_apis/blend_directory_api.py`,
+`lmcache/v1/multiprocess/modules/blend_coordinator.py`.
+
+## Why
+
+CacheBlend lookup today is **local to one mp-server**: the matcher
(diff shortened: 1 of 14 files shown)


deng451e (approved):
LGTM

ApostaC (approved):
LGTM!

PR #3567, opened 2026-06-06
[GPUKVFormat]: support vLLM CPU 2-fused KV layout

sammshen:
Under RFC: https://github.com/LMCache/LMCache/issues/3560

vLLM #44393 ([Attention][CPU] Standardize kv layout to blocks first, part of RFC #42082) changed the CPU attention backend KV cache from the 5D K/V-major [2, NB, NH, BS, HS] to the 4D blocks-first [NB, NH, BS, 2 * HS]

Previous CI Hotfix (now deprecating): https://github.com/LMCache/LMCache/commit/0fae56457ea31a6f3587330dd7b06ff808fd3362

In the long term, since vllm is movign towards this direction, we will likely need to make this kind of change for other attention types.

**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #3602, opened 2026-06-09
fix: handle NL_X_NB_NH_BS_TWO_HS in get_group_data_ptrs

maobaolong:
## Problem

`get_group_data_ptrs` raises `ValueError: Unknown GPU KV Format: 10` when the KV cache uses the `NL_X_NB_NH_BS_TWO_HS` layout (value 10), which was added in #3567 but not included in the per-layer branch of this function.

## Fix

Add `NL_X_NB_NH_BS_TWO_HS` to the per-layer list branch in `get_group_data_ptrs`. This format has the same per-layer tensor structure (`NL x [NB, NH, BS, 2, HS]`) as the other formats already handled in that branch.

### DIFF
--- a/lmcache/v1/gpu_connector/utils.py
+++ b/lmcache/v1/gpu_connector/utils.py
@@ -1274,6 +1274,7 @@ def get_group_data_ptrs(
         F.NL_X_NB_TWO_NH_BS_HS,
         F.NL_X_NB_BS_HS,
         F.NL_X_NBBS_ONE_HS,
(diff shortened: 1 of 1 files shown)


maobaolong:
@sammshen Would you like to take a look  at this PR? Thanks!

chunxiaozheng (approved):
lgtm!

hlin99 (approved):
LGTM

PR #3603, opened 2026-06-09
bench: support aligned L1 buffers for L2 adapters

Javen-Ke:

`lmcache bench l2` could not benchmark O_DIRECT L2 backends such as `raw_block` because the benchmark registered its L1 buffer with `align_bytes=1`. For `raw_block` with `use_odirect=true`, the adapter requires `l1_align_bytes >= block_align`, so adapter creation failed with:

```text
raw_block requires l1_align_bytes >= block_align when use_odirect=true
```

This PR adds `--l1-align-bytes` to `lmcache bench l2`, allocates an aligned benchmark L1 buffer, and creates benchmark `MemoryObj` views from that same registered buffer. With `--l1-align-bytes 4096`, the L2 bench can now exercise O_DIRECT backends correctly.

This only changes the L2 benchmark data path. It does not change LMCache server/runtime L2 adapter behavior.

Previously, the benchmark passed an `L1MemoryDesc` to `create_l2_adapter`, but the actual store/load `MemoryObj` tensors were allocated independently. That meant the registered L1 range was not the memory range used for payload I/O. This PR makes the benchmark payload objects slices of the registered L1 buffer, so the descriptor's address range and alignment describe the actual buffers used by store/load.

The benchmark now validates that `data-size-kb * 1024` is a multiple of `--l1-align-bytes`, because O_DIRECT paths require aligned buffer address, offset, and size.

## Verified with

```bash
.venv/bin/python -m pytest -q tests/cli/commands/bench/l2_adapter_bench/test_data.py
```

Manual raw_block O_DIRECT validation:

```bash
RAW_BLOCK_PATH=/var/tmp/lmcache-l2bench-raw-block.img
rm -f "$RAW_BLOCK_PATH"
truncate -s 512M "$RAW_BLOCK_PATH"

RAW_BLOCK_L2='{
  "type": "raw_block",
  "device_path": "'"$RAW_BLOCK_PATH"'",
  "slot_bytes": 4194304,
  "capacity_bytes": 0,
  "use_odirect": true,
  "block_align": 4096,
  "header_bytes": 4096,
  "meta_total_bytes": 67108864,
  "meta_magic": "PRTST001",
  "io_engine": "posix",
  "load_checkpoint_on_init": false,
  "num_store_workers": 1,
  "num_lookup_workers": 1,
  "num_load_workers": 1
}'

lmcache bench l2 \
  --l2-adapter "$RAW_BLOCK_L2" \
  --num-keys 1 \
  --in-flight 1 \
  --data-size-kb 1024 \
  --l1-align-bytes 4096 \
  --rounds 1 \
  --warmup-rounds 0 \
  --lookup-max-hit-rate 1.0 \
  --no-skip-verify
```

Negative validation with `--l1-align-bytes 1` still fails with the expected raw_block alignment error.

## If applicable

* [x] this PR contains user facing changes - docs added
* [x] this PR contains unit tests

### DIFF
--- a/docs/source/cli/bench.rst
+++ b/docs/source/cli/bench.rst
@@ -929,6 +929,12 @@ support a clean store -> load round-trip.
    ``"use_odirect": true``) or that talk to a remote service without
    a local cache, the default combined run is usually fine.
 
+   O_DIRECT adapters may also require the benchmark L1 buffer to
+   satisfy the adapter's block alignment. Use ``--l1-align-bytes`` to
+   set that alignment, commonly ``4096`` for local block devices. The
+   payload size (``--data-size-kb * 1024``) must be a multiple of the
+   selected alignment.
+
 
 Quick start
 ~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


sammshen (approved):
LGTM, this is also useful for NIXL

maobaolong (approved):
LGTM! Thanks for this improvement!

PR #3609, opened 2026-06-09
fix(memory): avoid valkey using a dangling pointer by using a staging buffer

### LINKED ISSUE #3610: TensorMemoryObj.byte_array returns dangling memoryview after allocator reuse
**Label**

**Describe the bug**

`TensorMemoryObj.byte_array` creates a `memoryview` via `ctypes.from_address()` that has no ownership link to the underlying tensor or the MemoryObj's `ref_count`. When the allocator frees and reuses that memory region (via `ref_count_down` → `free`), the still-live memoryview becomes a dangling pointer — reading or writing through it accesses memory that now belongs to a different allocation.

This causes heap corruption when used with valkey-glide's zero-copy buffer-GET (`valkey-glide-sync >= 2.3.0`), which writes into the memoryview from a native thread without the GIL held. Manifests as segfaults, `UnicodeDecodeError` on unrelated objects, and `TypeError` on `threading.Lock` methods in vLLM EngineCore processes.

**To Reproduce**

```python
import torch
from lmcache.v1.memory_management import TensorMemoryAllocator

allocator = TensorMemoryAllocator(torch.zeros(8192, dtype=torch.uint8))

# Allocate, get byte_array, then free
obj1 = allocator.allocate(torch.Size([4096]), torch.uint8)
obj1.raw_data.fill_(0xAA)
mv1 = obj1.byte_array
assert mv1.cast("B")[0] == 0xAA

obj1.ref_count_down()  # freed — slot returned to pool

# Allocate again — gets the same slot
obj2 = allocator.allocate(torch.Size([4096]), torch.uint8)
obj2.raw_data.fill_(0xBB)

# mv1 is dangling — sees obj2's data
assert mv1.cast("B")[0] == 0xBB  # BUG: proves dangling pointer
```

**Expected behavior**

While a `byte_array` memoryview is alive, the allocator should NOT be able to recycle the underlying memory. The memoryview should keep the MemoryObj alive via `ref_count`.

**Root Cause**

In [`TensorMemoryObj.byte_array`](https://github.com/LMCache/LMCache/blob/v0.4.3/lmcache/v1/memory_management.py#L644-L662):

```python
byte_array = (ctypes.c_ubyte * num_bytes).from_address(
    ctypes.addressof(ubyte_ptr.contents)
)
return memoryview(byte_array)
```

The ctypes array created by `from_address()` does not hold a reference to `self.raw_data` (the tensor). When [`ref_count_down`](https://github.com/LMCache/LMCache/blob/v0.4.3/lmcache/v1/memory_management.py#L539-L555) triggers `allocator.free(self)`, the address range is marked free and reused immediately, but any previously-returned memoryview still points to it.

**Suggested Fix**

Bump `ref_count` in `byte_array` and register a `weakref.finalize` on the ctypes array to decrement when collected:

```python
byte_array._tensor_ref = self.raw_data
self.ref_count_up()
weakref.finalize(byte_array, self.ref_count_down)
```

PR with fix and regression test: (link to PR once created)

- OS: Amazon Linux 2023 (EC2 g6.12xlarge)
- Python: 3.12
- LMCache: 0.4.3
- vLLM: 0.21.0
- valkey-glide-sync: 2.3.1

**Additional context**

- Reported originally at https://github.com/valkey-io/valkey-glide/issues/6215
- The bug is masked when using valkey-glide < 2.3.0 because the non-buffer GET path copies data with the GIL held, which serializes access to the buffer
- The corruption is sporadic (minutes to hours under load) because it requires the allocator to recycle a slot during the narrow window of a GIL-released FFI write

jduo:

Fixes #3610

GLIDE's buffer-GET writes with the GIL released into a caller-provided buffer. When that buffer aliases slab-allocator memory (`TensorMemoryObj.byte_array` via `ctypes.from_address()`), another thread can recycle the slot mid-write, causing heap corruption.

**Demonstration of the bug:**

```python
allocator = TensorMemoryAllocator(torch.zeros(8192, dtype=torch.uint8))

obj1 = allocator.allocate(torch.Size([4096]), torch.uint8)
obj1.raw_data.fill_(0xAA)
mv1 = obj1.byte_array  # memoryview via from_address — no ref to tensor

obj1.ref_count_down()  # slot returned to free list

obj2 = allocator.allocate(torch.Size([4096]), torch.uint8)
obj2.raw_data.fill_(0xBB)

assert mv1[0] == 0xBB  # PASSES — mv1 is dangling, sees obj2's data
```

**Fix:** Write into a per-thread reusable `bytearray` (not slab memory), then copy into the destination under the GIL. Eliminates the dangling-slot race without pinning or touching `ref_count`/`pin_count` — so the allocator's eviction loop is never blocked.

**Scope:** This fix is applied only to `ValkeyConnector` (the only connector that uses GLIDE's GIL-released buffer-GET). `RESPConnector` and `RedisConnector` are not affected by this race because they do not perform GIL-released writes into caller-provided buffers.

**Changes:**
* `ValkeyConnector._ThreadWorkerPool`: add `_get_scratch()` + stage-then-copy in `_do_get_into`
* Regression tests covering the underlying bug, staging buffer safety, single GET, and batched GET paths

Proven stable: 100 concurrent users, 64k context, 1 hour × 2, zero crashes, zero alloc failures, P99 on par with unpatched baseline (benchmarked by @omerrubi-amzn).

Related: valkey-io/valkey-glide#6215

**If applicable**:


lmcache/v1/memory_management.py
```diff
@@ -768,6 +765,12 @@ def byte_array(self) -> memoryview:
         byte_array = (ctypes.c_ubyte * num_bytes).from_address(
             ctypes.addressof(ubyte_ptr.contents)
         )
+        # Prevent GC of the tensor and prevent allocator reuse while
+        # this ctypes array (and any memoryview over it) is alive.
+        byte_array._tensor_ref = self.raw_data
+        self.ref_count_up()
+        weakref.finalize(byte_array, self.ref_count_down)
+
         return memoryview(byte_array)
```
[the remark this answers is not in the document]
jduo: This suggestion seems to be incorrect.
When LMCache's ValkeyConnector does dst = dst.cast("B"), the original memoryview dies, the finalizer fires, and the memory gets freed prematurely.

The original approach (finalizer on the ctypes array) is correct because the ctypes array stays alive as long as any memoryview (original or cast) references it.

omerrubi-amzn:
Thanks for tracking this down — I've been independently reproducing #6215 (real `valkey-glide-sync` 2.3.1 + LMCache 0.4.3), and I can confirm this fix resolves it: my connector-level repro goes from **195/195 corruptions to 0/197** with this change, and the added unit test passes. The diagnosis (dangling `byte_array` memoryview written into by GLIDE's GIL-released native thread after the slot is recycled) is spot on.

Before this lands, I want to raise a design concern about putting the `ref_count` side effect **inside the `byte_array` getter**, and propose an alternative for discussion.

### Concerns with the getter-side-effect approach

1. **Blast radius.** `byte_array` is read in ~40 places across ~20 files (every remote connector, the L2 adapters, `local_disk`/`fs`/`raw_block` backends, `audit_backend`, `cachegen_decoder`, S3 streams, …). All of them now silently acquire a ref and depend on GC to release it.
2. **Non-deterministic release.** The ref drops only when the ctypes array is collected (`weakref.finalize`). On the async connectors the view lives in a completed coroutine frame retained by the asyncio Task, so the ref lingers until GC.
3. **Eviction pressure under load.** `can_evict()` requires `ref_count == 1`, so views that haven't been collected keep slots non-evictable. Under sustained load I see this surface as `Failed to allocate memory block` floods with loads aborting before they issue remote GETs — i.e. it can throttle the hot path.
4. **CI / contract change.** `test_redis_connector` asserts `ref_count == 1` after a put; the getter's transient view leaves it at `2` until GC (`assert 2 == 1`). It's GC-timing rather than a leak (`del` of the view restores the count immediately), but it shows the getter now has an observable side effect callers must account for.
5. **Side effects on non-write paths.** `len(memory_obj.byte_array)` now allocates a ctypes array + memoryview + finalizer just to read a length (could be `get_size()`), and the `x if isinstance(byte_array, …) else …` patterns access the property twice → double ref bump.

### Proposed alternative: a scoped-pin context manager

Keep the getter side-effect free, and protect **only** the dangerous window (the in-flight buffer write) explicitly:

```python
@contextmanager
def byte_array_pinned(self):
    """Yield a 'B'-format memoryview while guaranteeing the allocator cannot
    free/recycle this slot for the duration (issue #6215). Released
    deterministically on exit — no reliance on GC/weakref."""
    self.ref_count_up()
    mv = self.byte_array          # plain getter, no side effect
    if mv.format != "B":
        mv = mv.cast("B")
    try:
        yield mv
    finally:
        mv.release()
        self.ref_count_down()
```

Receive sites opt in around the in-flight GET:

```python
with memory_obj.byte_array_pinned() as dst:
    bytes_read = await asyncio.wrap_future(
        self._pool.submit_get_into(key.to_string(), dst))
# slot is releasable again here — immediately, no GC
```

I prototyped this and validated with `gc.disable()`: inside the block, triggering the exact race (a `ref_count_down()` mid-flight) does **not** let a concurrent realloc reuse the slot; after the block exits, **without any `gc.collect()`**, the slot is released and reusable. So protection holds across the write and release is deterministic.

### Tradeoffs

| | This PR (getter side effect) | Scoped pin |
|---|---|---|
| Release timing | Non-deterministic (GC) | Deterministic (`with` exit) |
| Release overhead | same `ref_count_down`, + per-access `weakref.finalize` + GC pressure | same `ref_count_down`; on success only drops to base ref (no `free()` on GET path) |
| Throughput under memory pressure | eviction delay → alloc-failure churn | no eviction delay |
| Blast radius | ~40 `byte_array` consumers | ~2 receive sites |
| `len()` / sync sites | bump + leak a transient ref | unaffected |
| `test_redis_connector` | needs adjusting (`ref==2`) | unchanged |
| **Safety by default** | ✅ automatic — every access protected | ⚠️ opt-in — a new receive path that forgets the CM is unprotected |
| Migration cost | localized | add API + migrate receive sites |

On performance specifically: deterministic release is **not** more expensive than GC release — it's the same `ref_count_down`, and on the success path it only drops the protection ref back to the caller's base ref (no `free()` on the GET critical path). It additionally avoids the per-access `weakref.finalize` registration and the eviction-delay throttling. The one genuine advantage of this PR's approach is that it's **fail-safe by default** — protection can't be forgotten — whereas the scoped pin is fail-open.

### Suggested middle ground

Keep the getter side-effect free + add `byte_array_pinned()` as the blessed receive API, migrate the `_get`/`_batched_get` paths to it, switch `len(byte_array)` → `get_size()`, and add a lint/CI guard that flags raw `byte_array` use on async receive paths so a forgotten pin fails review rather than silently regressing. That keeps the deterministic-release / low-blast-radius benefits while closing the fail-open gap.

jduo:
> Thanks for tracking this down — I've been independently reproducing #6215 (real `valkey-glide-sync` 2.3.1 + LMCache 0.4.3), and I can confirm this fix resolves it: my connector-level repro goes from **195/195 corruptions to 0/197** with this change, and the added unit test passes. The diagnosis (dangling `byte_array` memoryview written into by GLIDE's GIL-released native thread after the slot is recycled) is spot on.
> 
> Before this lands, I want to raise a design concern about putting the `ref_count` side effect **inside the `byte_array` getter**, and propose an alternative for discussion.
> 
> ### Concerns with the getter-side-effect approach
> 1. **Blast radius.** `byte_array` is read in ~40 places across ~20 files (every remote connector, the L2 adapters, `local_disk`/`fs`/`raw_block` backends, `audit_backend`, `cachegen_decoder`, S3 streams, …). All of them now silently acquire a ref and depend on GC to release it.
> 2. **Non-deterministic release.** The ref drops only when the ctypes array is collected (`weakref.finalize`). On the async connectors the view lives in a completed coroutine frame retained by the asyncio Task, so the ref lingers until GC.
> 3. **Eviction pressure under load.** `can_evict()` requires `ref_count == 1`, so views that haven't been collected keep slots non-evictable. Under sustained load I see this surface as `Failed to allocate memory block` floods with loads aborting before they issue remote GETs — i.e. it can throttle the hot path.
> 4. **CI / contract change.** `test_redis_connector` asserts `ref_count == 1` after a put; the getter's transient view leaves it at `2` until GC (`assert 2 == 1`). It's GC-timing rather than a leak (`del` of the view restores the count immediately), but it shows the getter now has an observable side effect callers must account for.
> 5. **Side effects on non-write paths.** `len(memory_obj.byte_array)` now allocates a ctypes array + memoryview + finalizer just to read a length (could be `get_size()`), and the `x if isinstance(byte_array, …) else …` patterns access the property twice → double ref bump.
> 
> ### Proposed alternative: a scoped-pin context manager
> Keep the getter side-effect free, and protect **only** the dangerous window (the in-flight buffer write) explicitly:
> 
> ```python
> @contextmanager
> def byte_array_pinned(self):
>     """Yield a 'B'-format memoryview while guaranteeing the allocator cannot
>     free/recycle this slot for the duration (issue #6215). Released
>     deterministically on exit — no reliance on GC/weakref."""
>     self.ref_count_up()
>     mv = self.byte_array          # plain getter, no side effect
>     if mv.format != "B":
>         mv = mv.cast("B")
>     try:
>         yield mv
>     finally:
>         mv.release()
>         self.ref_count_down()
> ```
> 
> Receive sites opt in around the in-flight GET:
> 
> ```python
> with memory_obj.byte_array_pinned() as dst:
>     bytes_read = await asyncio.wrap_future(
>         self._pool.submit_get_into(key.to_string(), dst))
> # slot is releasable again here — immediately, no GC
> ```
> 
> I prototyped this and validated with `gc.disable()`: inside the block, triggering the exact race (a `ref_count_down()` mid-flight) does **not** let a concurrent realloc reuse the slot; after the block exits, **without any `gc.collect()`**, the slot is released and reusable. So protection holds across the write and release is deterministic.
> 
> ### Tradeoffs
> This PR (getter side effect)	Scoped pin
> Release timing	Non-deterministic (GC)	Deterministic (`with` exit)
> Release overhead	same `ref_count_down`, + per-access `weakref.finalize` + GC pressure	same `ref_count_down`; on success only drops to base ref (no `free()` on GET path)
> Throughput under memory pressure	eviction delay → alloc-failure churn	no eviction delay
> Blast radius	~40 `byte_array` consumers	~2 receive sites
> `len()` / sync sites	bump + leak a transient ref	unaffected
> `test_redis_connector`	needs adjusting (`ref==2`)	unchanged
> **Safety by default**	✅ automatic — every access protected	⚠️ opt-in — a new receive path that forgets the CM is unprotected
> Migration cost	localized	add API + migrate receive sites
> On performance specifically: deterministic release is **not** more expensive than GC release — it's the same `ref_count_down`, and on the success path it only drops the protection ref back to the caller's base ref (no `free()` on the GET critical path). It additionally avoids the per-access `weakref.finalize` registration and the eviction-delay throttling. The one genuine advantage of this PR's approach is that it's **fail-safe by default** — protection can't be forgotten — whereas the scoped pin is fail-open.
> 
> ### Suggested middle ground
> Keep the getter side-effect free + add `byte_array_pinned()` as the blessed receive API, migrate the `_get`/`_batched_get` paths to it, switch `len(byte_array)` → `get_size()`, and add a lint/CI guard that flags raw `byte_array` use on async receive paths so a forgotten pin fails review rather than silently regressing. That keeps the deterministic-release / low-blast-radius benefits while closing the fail-open gap.

Thanks, this is a good suggestion to avoid use cases where we don't actually need the pinning. I've implemented as above and updated relevant connectors (valkey, RESP, and redis). I've also added the linting rule suggested.

lmcache/v1/storage_backend/connector/valkey_connector.py
```diff
@@ -471,6 +466,8 @@ async def _batched_get(
                 continue
 
             memory_objs.append(mobj)
+            # Pin during native write (GIL released in FFI call).
+            mobj.ref_count_up()
```
DongDongJu: where this one ref_count_down in success path?

lmcache/v1/storage_backend/connector/valkey_connector.py
```diff
@@ -505,6 +502,11 @@ async def _batched_get(
                 if mobj is not None:
                     mobj.ref_count_down()
             raise
+        finally:
+            # Release write-pin on all slots.
+            for mobj in memory_objs:
+                if mobj is not None:
+                    mobj._write_pin.unpin()
```
DongDongJu: this one should match with 470 lines. isnt it? LMK if im wrong.

lmcache/v1/storage_backend/connector/valkey_connector.py
DongDongJu, line 530: hmm. isnt this problematic with line 470?

DongDongJu (changes requested):
in valkey_connector, we call `mobj.ref_count_up()`, but later the `finally` block calls
`mobj._write_pin.unpin()` instead of releasing the same refcount. 
I also do not see a matching `_write_pin.pin()` in this path.

This means:
- success path leaks one refcount per returned object;
- miss/exception paths only drop one ref and can leave the allocation alive;
- `_write_pin.unpin()` is called without a matching `_write_pin.pin()`.

Could we remove the manual `ref_count_up()` / `_write_pin.unpin()` here and use a scoped memory-management API for the batch buffers instead?

jduo:
@DongDongJu , we're looking at other possible approaches to solve this as some benchmarks indicated there was performance degradation.

omerrubi-amzn:
Thanks for iterating on this @jduo — the `WritePin` atomic approach is clean. It addresses the concerns I raised (getter side-effect free, deterministic release, low blast radius). The C++ `std::atomic` is the right primitive here.

### Question: deferred free under pool exhaustion

I see `can_evict` checks `pin_count` but not `_write_pin.is_pinned()` — so the evictor can select a write-pinned slot and call `ref_count_down()`, but `free()` is deferred until the pin drops. This is better than blocking eviction outright (which we proved fails catastrophically — see below).

My question: when the deferred `free()` finally fires in `byte_array_pinned()`'s `finally` block, does the slot actually return to the allocator's free list at that point? Or can the allocator enter its `sleep(0.1)` busy-loop because it doesn't "see" the pending return? Under our workload (128 concurrent GETs × ~150 ms each), there are always ~1.3 GB of transiently pinned slots — if those are invisible to `allocate()`, the same starvation pattern could surface.

### Our load-test data

We've been benchmarking #6215 fixes under sustained pressure (100 concurrent users, 64k context, Llama-3.1-70B, 7×p5.48xlarge, 50-shard ElastiCache Valkey cluster, L1 pool near-full at ~70 GB):

| Approach | Result |
|---|---|
| Original (weakref on getter) | P50 regressed 2× ; alloc failure floods |
| Scoped pin (`ref_count_up`/`down`) | **Crashed at 42 min** — 888K alloc failures/hr, evictor starved |
| Scoped pin (`pin_count`) | Same failure — any pin that blocks eviction is fatal at this occupancy |
| **Staging scratch buffer** | ✅ Full hour ×2, zero crashes, zero alloc failures, P99 on par with unpatched baseline |

The staging buffer approach writes GLIDE's buffer-GET into a per-thread reusable `bytearray` (not slab memory), then copies into the slab under the GIL. Zero allocator interaction → eliminates #6215 by construction. Cost is one memcpy per chunk — which is noise given the GIL-bound client overhead that dominates L2 GET latency (see [Tensormesh's Redis optimization writeup](https://www.tensormesh.ai/blog-posts/blog-redis-llm-kv-cache-optimization-30x-throughput) — client-side GIL serialization was the 30× bottleneck, not buffer copies)


### Recommendation

I'd suggest a two-step approach:

1. **Merge a safe fix now** — the staging scratch buffer eliminates the dangling-pointer race by construction with zero allocator interaction. It's proven stable under sustained production-grade load and has no regression. This unblocks users hitting #6215 today.

2. **Open a follow-up optimization issue** for the `WritePin` zero-copy approach. Once the `_batched_get` ref leak is fixed and the deferred-free behavior is validated under sustained pool exhaustion (our fleet can run this when GPUs are back), it can replace the staging buffer as a perf optimization — removing the memcpy for environments where every ms on the L2 GET path matters.



diff for the staging scratch buffer:
```
--- a/lmcache/v1/storage_backend/connector/valkey_connector.py
+++ b/lmcache/v1/storage_backend/connector/valkey_connector.py
@@ -205,6 +205,19 @@ class _ThreadWorkerPool:
         """SET a key (runs on a worker thread)."""
         self._get_client().set(key_str.encode(), data)
 
+    def _get_scratch(self, size: int) -> bytearray:
+        """Per-thread reusable staging buffer for buffer-GET (issue #6215).
+
+        Glide writes into this thread-private bytearray (not the L1 slab),
+        then we copy into the destination under the GIL. Eliminates the
+        dangling-slot race without pinning or touching ref_count/pin_count.
+        """
+        buf = getattr(self._local, "scratch", None)
+        if buf is None or len(buf) < size:
+            buf = bytearray(size)
+            self._local.scratch = buf
+        return buf
+
     def _do_get_into(self, key_str: str, buf: memoryview) -> int:
         """GET a key into a buffer (runs on a worker thread).
 
@@ -216,12 +229,15 @@ class _ThreadWorkerPool:
         """
         client = self._get_client()
         if self._has_buffer_get:
-            result = client.get(key_str.encode(), buffer=buf)
+            size = buf.nbytes
+            scratch = self._get_scratch(size)
+            scratch_view = memoryview(scratch)[:size]
+            result = client.get(key_str.encode(), buffer=scratch_view)
             if result is None:
                 return 0
-            # GLIDE's buffer GET writes the value into ``buf`` and returns the
-            # number of bytes written as bytes, e.g. b'1000'.
-            return int(result)
+            n = int(result)
+            buf[:n] = scratch_view[:n]
+            return n
         else:
             data = client.get(key_str.encode())
             if data is None:
```

lmcache/v1/storage_backend/connector/valkey_connector.py
jduo: Yes this is an issue. Changed the approach to use a staging buffer for simplicity.

lmcache/v1/storage_backend/connector/valkey_connector.py
jduo: Yes this is an issue. Changed the approach to use a staging buffer for simplicity.

lmcache/v1/storage_backend/connector/valkey_connector.py
jduo: Yes this is an issue. Changed the approach to use a staging buffer for simplicity.

jduo:
> Thanks for iterating on this @jduo — the `WritePin` atomic approach is clean. It addresses the concerns I raised (getter side-effect free, deterministic release, low blast radius). The C++ `std::atomic` is the right primitive here.
> 
> ### Question: deferred free under pool exhaustion
> I see `can_evict` checks `pin_count` but not `_write_pin.is_pinned()` — so the evictor can select a write-pinned slot and call `ref_count_down()`, but `free()` is deferred until the pin drops. This is better than blocking eviction outright (which we proved fails catastrophically — see below).
> 
> My question: when the deferred `free()` finally fires in `byte_array_pinned()`'s `finally` block, does the slot actually return to the allocator's free list at that point? Or can the allocator enter its `sleep(0.1)` busy-loop because it doesn't "see" the pending return? Under our workload (128 concurrent GETs × ~150 ms each), there are always ~1.3 GB of transiently pinned slots — if those are invisible to `allocate()`, the same starvation pattern could surface.
> 
> ### Our load-test data
> We've been benchmarking #6215 fixes under sustained pressure (100 concurrent users, 64k context, Llama-3.1-70B, 7×p5.48xlarge, 50-shard ElastiCache Valkey cluster, L1 pool near-full at ~70 GB):
> 
> Approach	Result
> Original (weakref on getter)	P50 regressed 2× ; alloc failure floods
> Scoped pin (`ref_count_up`/`down`)	**Crashed at 42 min** — 888K alloc failures/hr, evictor starved
> Scoped pin (`pin_count`)	Same failure — any pin that blocks eviction is fatal at this occupancy
> **Staging scratch buffer**	✅ Full hour ×2, zero crashes, zero alloc failures, P99 on par with unpatched baseline
> The staging buffer approach writes GLIDE's buffer-GET into a per-thread reusable `bytearray` (not slab memory), then copies into the slab under the GIL. Zero allocator interaction → eliminates #6215 by construction. Cost is one memcpy per chunk — which is noise given the GIL-bound client overhead that dominates L2 GET latency (see [Tensormesh's Redis optimization writeup](https://www.tensormesh.ai/blog-posts/blog-redis-llm-kv-cache-optimization-30x-throughput) — client-side GIL serialization was the 30× bottleneck, not buffer copies)
> 
> ### Recommendation
> I'd suggest a two-step approach:
> 
> 1. **Merge a safe fix now** — the staging scratch buffer eliminates the dangling-pointer race by construction with zero allocator interaction. It's proven stable under sustained production-grade load and has no regression. This unblocks users hitting #6215 today.
> 2. **Open a follow-up optimization issue** for the `WritePin` zero-copy approach. Once the `_batched_get` ref leak is fixed and the deferred-free behavior is validated under sustained pool exhaustion (our fleet can run this when GPUs are back), it can replace the staging buffer as a perf optimization — removing the memcpy for environments where every ms on the L2 GET path matters.
> 
> diff for the staging scratch buffer:
> 
> ```
> --- a/lmcache/v1/storage_backend/connector/valkey_connector.py
> +++ b/lmcache/v1/storage_backend/connector/valkey_connector.py
> @@ -205,6 +205,19 @@ class _ThreadWorkerPool:
>          """SET a key (runs on a worker thread)."""
>          self._get_client().set(key_str.encode(), data)
>  
> +    def _get_scratch(self, size: int) -> bytearray:
> +        """Per-thread reusable staging buffer for buffer-GET (issue #6215).
> +
> +        Glide writes into this thread-private bytearray (not the L1 slab),
> +        then we copy into the destination under the GIL. Eliminates the
> +        dangling-slot race without pinning or touching ref_count/pin_count.
> +        """
> +        buf = getattr(self._local, "scratch", None)
> +        if buf is None or len(buf) < size:
> +            buf = bytearray(size)
> +            self._local.scratch = buf
> +        return buf
> +
>      def _do_get_into(self, key_str: str, buf: memoryview) -> int:
>          """GET a key into a buffer (runs on a worker thread).
>  
> @@ -216,12 +229,15 @@ class _ThreadWorkerPool:
>          """
>          client = self._get_client()
>          if self._has_buffer_get:
> -            result = client.get(key_str.encode(), buffer=buf)
> +            size = buf.nbytes
> +            scratch = self._get_scratch(size)
> +            scratch_view = memoryview(scratch)[:size]
> +            result = client.get(key_str.encode(), buffer=scratch_view)
>              if result is None:
>                  return 0
> -            # GLIDE's buffer GET writes the value into ``buf`` and returns the
> -            # number of bytes written as bytes, e.g. b'1000'.
> -            return int(result)
> +            n = int(result)
> +            buf[:n] = scratch_view[:n]
> +            return n
>          else:
>              data = client.get(key_str.encode())
>              if data is None:
> ```

Thanks for the suggestion @omerrubi-amzn .

The scratch buffer solution is a fair bit easier to get right. I've added an implementation to this PR. Also covered the batched get code path, and applied the same fix to RESP connector and Redis connector.

omerrubi-amzn:
LGTM, thanks for driving the investigation and the fix!

DongDongJu (approved):
Hello @jduo,
The connector-level staging approach is acceptable. 
It avoids changing allocator semantics, avoids blocking eviction through ref/pin-counting, and keeps the hot path straightforward. 

BTW, can you change the PR name to do not reflect the full allocator fix. it seems valkey mitigation.
Otherwise, LGTM.

jduo:
> Hello @jduo, The connector-level staging approach is acceptable. It avoids changing allocator semantics, avoids blocking eviction through ref/pin-counting, and keeps the hot path straightforward.
> 
> BTW, can you change the PR name to do not reflect the full allocator fix. it seems valkey mitigation. Otherwise, LGTM.

Good point. I've updated the PR name.

sammshen (approved):
LGTM!

PR #3625, opened 2026-06-10
[cli] refactor query and trace cli

chunxiaozheng:
we should not put all sub-commands into `__init__.py`, `__init__.py` only implements some abstract method and override `register` and `execute`, the implementation of each sub-command should be placed in a separate file.


sammshen (approved):
Like this refactor a lot, LGTM!

deng451e (approved):
LGTM

PR #3631, opened 2026-06-11
[Feat] Worker-adapter liveness hygiene for MP server reaping (1/2)

royyhuang:

First of two PRs replacing #3231 (reworked per review feedback). End goal: when a vLLM worker dies silently, the MP server reaps its `GPUCacheContext` and CUDA IPC handles instead of leaking them. This PR is the adapter-side groundwork. No wire or server change — those land in PR 2.

- **uuid-based `instance_id`.** `os.getpid()` collides across containerized replicas with per-pod PID namespaces, so a new worker can silently bind to a dead worker's stale server context. Now a uuid4-derived 63-bit int: OS entropy (safe under `random.seed()`), masked to stay signed-int64-safe in msgpack.
- **Pessimistic heartbeat start.** Lazy start at first store/retrieve is unchanged. New: the create path clears `health_event` before `start()`. The first successful ping then always fires the recover callback and re-registers. The server noops if still registered, rebuilds if not — this is what makes recovery after a future reap automatic.
- **Dropped retrieves are reported.** The unhealthy early-return of `submit_retrieve_request` discarded the request id, so async loads (`load_kv_async=True`) hung forever in `WAITING_FOR_REMOTE_KVS`. Both `get_finished` branches now report each dropped id exactly once.
- **Clean shutdown.** `shutdown()` stops the heartbeat before UNREGISTER (today the thread is never stopped). Straggler guards stop an in-flight cycle or recover callback from re-registering a ghost context after UNREGISTER. Adds public `PeriodicThread.stop_requested`.
- **Startup WARNING** when `3 × heartbeat_interval > 30 s`: the future server reap floor only covers the default 10 s interval.

Includes `docs/design/v1/multiprocess/worker_liveness.md` — high-level design for the full feature.

- @maobaolong's question from #3231: the lazy heartbeat start is kept untouched. PR 2 covers the register-to-first-request window with a server-side registration grace, not by pinging through warmup.
- Tri-state ping, cold-start grace, and terminal PING-False from #3231 are all gone. Recovery uses the existing recover callback + idempotent register, with a plain bool ping.
- PR 2 scope: PING payload `[] → [int | None]`, server-side `last_seen`, reaper thread, config knobs.

Refs #3231

**If applicable**:


docs/design/v1/multiprocess/worker_liveness.md
```diff
@@ -0,0 +1,181 @@
...
+|                           |                   |   has_liveness_signal}               |
++---------------------------+                   |  _lock (leaf); pop -> cleanup        |
+                                                +--------------------------------------+
+```
+
+## 3. Protocol Change
+
+The PING payload changes in place from `[]` to `[int | None]`; the response stays
+`bool`, always `True` (`None` marks an untracked prober such as the scheduler
+adapter, which registers nothing and is never reapable). PING also moves to SYNC
```
ApostaC: We may not want to move PING to SYNC dispatch -- right now, REGISTER_KV_CACHE also uses SYNC dispatch, but it may block the server-side dispatcher thread for a while -- which will make PING unresponsive.

The good solution is to move REGISTER_KV_CACHE to BLOCKING dispatch, and also have a separate thread to answer PING. But for now, let's put PING to the BLOCKING dispatch using normal thread pool

ApostaC: Another reason to share the normal pool with other handlers is that if the normal pool cannot handle PING for more than 10 seconds, the vLLM should also enter the downgrade mode.

docs/design/v1/multiprocess/worker_liveness.md
```diff
@@ -0,0 +1,181 @@
...
+
+### 5.2 Locking
+
+Each transfer module gains one leaf `threading.Lock` guarding all of its
+per-instance state; in `NonGPUTransferModule` the context and strategy dicts
+mutate as a pair in one critical section, so a reap racing a re-register can never
+strand a fresh context without its strategy. The lock is never held across context
+construction, storage calls, or any other component, so no thread holds two locks.
+External readers use the locked accessors `get_context_entry` (get-and-refresh)
+and `context_entries_snapshot`, replacing the unlocked `cache_contexts` property.
```
ApostaC: This part is a bit unclear. What does the lock do here?

docs/design/v1/multiprocess/worker_liveness.md
```diff
@@ -0,0 +1,181 @@
...
+
+### 5.3 Reaper
+
+`ManagementModule` (the PING owner) owns the reaper thread, scanning every
+`reap_timeout / 4`. Each scan calls `reap_stale_instances` on every target: under
+the module lock, collect ids whose staleness exceeds their window and pop them;
+outside the lock, run the same cleanup as a client unregister and log a WARNING
+per instance (repeated reaps of one id signal a too-small timeout). Reaped ids fan
+out to every `InstanceReapListener.drop_instance_state(id)`; `BlendV3Module` drops
+its context mirrors there, releasing the last strong reference. Collect+pop shares
```
ApostaC: Btw, we are removing the context mirror from the BlendV3Module.

docs/design/v1/multiprocess/worker_liveness.md
```diff
@@ -0,0 +1,181 @@
...
+        self, reap_timeout_s: float, registration_grace_s: float
+    ) -> list[int]: ...
+    def tracked_instance_count(self) -> int: ...
+
+class InstanceReapListener(Protocol):
+    def drop_instance_state(self, instance_id: int) -> None: ...
+```
+
+Both transfer modules implement `InstanceLivenessTarget`; `BlendV3Module` is an
+`InstanceReapListener`; `ManagementModule` receives both by constructor injection.
```
ApostaC, line 127: I just realized one thing: previously, the transfer contexts and the ping module were handled by the same class, but now they are in two different classes. 

Right now, it's more like a tree-structure:
- MP Engine Context (owns all the sub modules and the storage manager)
  - Management Module -- which answers PING, and has a reference to the MP Engine Context
  - Transfer Module (GPU transfer or CPU transfer) -- which handles register KV cache, store, and retrieve -- it also has a reference to the MP module.

The problem is that the Management Module cannot directly see the Transfer Modules.

Therefore, maybe we need to do
Management Module calls `self._ctx.xxx` when handling ping or reap --> the MPEngineContext dispatches the function to the correct transfer module.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1110,44 +1162,67 @@ def _send_register_kv_caches_request(
...
+        if self._heartbeat_stop_requested():
+            logger.info("Heartbeat stop requested; skipping KV cache re-registration")
+            return False
+
         try:
-            self._send_register_kv_caches_request(self.kv_caches)
-            logger.warning("Finished re-registering KV caches after server recovery")
+            registered = self._send_register_kv_caches_request(
+                self.kv_caches, abort_predicate=self._heartbeat_stop_requested
+            )
```
ApostaC: I'm not so sure why we introduced the `abort_predicate` here. Let's discuss

ApostaC (commented):
Left some comments. PTAL

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1420,10 +1511,20 @@ def get_finished(
...
+        self._dropped_retrieves.clear()
+
         # Update the internal states
         ret_stores = self._process_finished_stores(
             finished_stores, finished_req_ids_from_engine
         )
+        # A dropped retrieve may already be engine-finished in this same
+        # call; it must not appear in both finished_recving and
+        # finished_sending (mirrors the unhealthy branch's guard).
+        ret_stores -= dropped_retrieves
```
ApostaC: Looks like a bit weird here

ApostaC (approved):
LGTM!

deng451e (approved):

PR #3634, opened 2026-06-11
[Core] Add bounded retry for store-side CPU memory admission

dvaldivia:
Adds a new config knob `store_admission_retry_ms` that gives a failed
store-side CPU pool allocation a bounded window to retry (re-attempting
eviction each pass) before the chunk is dropped, and a Prometheus counter
`lmcache:store_dropped_chunk_count` so drops are visible at all.

**Default `0` preserves today's behavior exactly** — a failed store-side
allocation drops immediately, no sleeping, no retry. The retry only applies
to the store path (`store()` / `store_layer()` in `cache_engine.py`); the
retrieve path and every other `busy_loop=False` allocation site (disk-load
staging, NIXL staging, EC engine) are untouched.

### What's happening today

Under concurrent store pressure, `LocalCPUBackend.allocate(eviction=True,
busy_loop=False)` fails transiently — the LRU eviction candidates are pinned
by in-flight requests — and `store()` breaks out of the chunk loop: the
failed chunk and every later chunk of that request are silently dropped and
never reach any storage backend. Because lookups truncate at the first
missing chunk (the vLLM KV-connector contract is contiguous-prefix-only),
sparse drops collapse the retrievability of entire long documents.

We hit this on a production-style benchmark (concurrency 28, ~36k-token
documents, Qwen3-30B, write-through to a storage plugin):

| CPU pool | admission failures | chunks stored | retrievals (3 reps) |
|---|---:|---:|---:|
| 128 GiB | 36,819 | ~54% | 289 |
| 256 GiB (no other change) | — | ~100% | 47,180 (98.9 GB served) |

The 128 GiB row is the storage tier being effectively write-only, and it's
nearly invisible: the only signal is per-attempt WARNING spam from the
allocator, and `lmcache:num_hit_tokens / num_requested_tokens` only count
attempted retrieves, so they read 100% the whole time data is being lost.

Right-sizing `max_local_cpu_size` is the first-order fix, but the right size
depends on workload concurrency and document length, and overshooting wastes
pinned host memory. A few tens of milliseconds of bounded retry lets an
in-flight request unpin an eviction candidate — versus losing the chunk and
later re-prefilling a whole 36k-token document because the lookup truncated
at the hole. The counter makes whatever still drops visible so operators can
size the pool from data.

Relationship to existing knobs: `extra_config.force_store_wait=True` already
busy-loops stores **unboundedly** (with a documented deadlock risk for
concurrent stores) — when set, allocation never returns `None` and this knob
is a no-op. `pd_allocation_timeout_sec` is the analogous bounded timeout but
PD-backend-only. This PR brings the same bounded idea to the default
local-CPU store path.

### Changes

- `lmcache/v1/config.py` — `store_admission_retry_ms: int = 0` in
  `_CONFIG_DEFINITIONS` (env: `LMCACHE_STORE_ADMISSION_RETRY_MS`), placed
  with the memory-management knobs.
- `lmcache/v1/cache_engine.py` — `_retry_store_allocation()` (bounded by a
  `time.monotonic()` deadline, 2 ms interval) wrapping the store-path
  allocations; `_on_store_chunk_dropped()` counts the drop and replaces the
  per-request warning with one aggregated warning rate-limited to once per
  10 s, pointing at `max_local_cpu_size` / `store_admission_retry_ms`.
- `lmcache/observability.py` — `update_store_dropped_chunk_count()` +
  `lmcache:store_dropped_chunk_count` Prometheus counter, same pattern as
  the neighboring `lmcache:local_cpu_evict_failed_count`.
- `lmcache/v1/memory_management.py` — per-attempt out-of-memory warnings in
  `AddressManager.allocate`/`batched_allocate` demoted to debug
  (`PagedTensorMemoryAllocator` already logs the identical condition at
  debug; the caller decides how to handle pressure).
- `docs/source/api_reference/configurations.rst` — knob documented.

### How to test this PR?

```bash
pytest tests/v1/test_cache_engine.py -k "admission" -v
pytest tests/test_observability.py -k "store_dropped" -v
```

- `test_store_drop_on_admission_failure` — real engine via
  `LMCacheEngineBuilder` with `MockGPUConnector` (CPU-only) and a ~5-chunk
  pool, pool held full, knob at default `0`: the store drops (lookup
  returns 0) and the counter increments.
- `test_store_admission_retry` — same setup with
  `store_admission_retry_ms=10000` and a timer freeing the pool 200 ms in:
  the store succeeds and the counter does not move.
- Both use `@pytest.mark.no_shared_allocator` so the small pool takes
  effect; no GPU required (same harness as
  `test_freeze_with_real_cache_engine`, targets the CPU CI lane).

Heads-up for reviewers: I could not run the pytest suite locally (no torch
in this environment) — `ruff check`, `ruff format --check`, `isort`,
`codespell`, and `py_compile` all pass on the touched files; the two engine
tests are written against the existing CPU-only harness and should run in
the CPU lane here.

### Checklist

- [x] Based on `dev`
- [x] DCO sign-off on the commit
- [x] Default-off: no behavior change unless the new knob is set
- [x] Type hints + docstrings on new/changed functions
- [x] New behavior covered by tests (drop path and recovery path, CPU-only)
- [x] Docs updated (`configurations.rst`)
- [x] `ruff` / `isort` / `codespell` clean on touched files
- [ ] Full `pre-commit run --all-files` + pytest (needs CI — no torch locally)

### DIFF
--- a/lmcache/observability.py
+++ b/lmcache/observability.py
@@ -70,6 +70,8 @@ class LMCacheStats:
     interval_local_cpu_evict_keys_count: int  # evict keys count
     interval_local_cpu_evict_failed_count: int  # evict failed count
 
+    interval_store_dropped_chunk_count: int  # store chunks dropped on allocation
+
     interval_forced_unpin_count: int  # forced unpin count due to timeout
 
     # Real time value measurements (will be reset after each log)
@@ -293,6 +295,8 @@ class LMCStatsMonitor:
         self.interval_local_cpu_evict_count = 0
         self.interval_local_cpu_evict_keys_count = 0
         self.interval_local_cpu_evict_failed_count = 0
+
+        self.interval_store_dropped_chunk_count = 0
 
         self.interval_forced_unpin_count = 0
 
@@ -572,6 +576,17 @@ class LMCStatsMonitor:
         self.interval_local_cpu_evict_failed_count += evict_failed_count
 
     @thread_safe
+    def update_store_dropped_chunk_count(self, dropped_count: int):
+        """
+        Record store-path chunks that were dropped because CPU memory
+        could not be allocated for them.
+
+        Args:
+            dropped_count: The number of chunks that were dropped.
+        """
+        self.interval_store_dropped_chunk_count += dropped_count
+
+    @thread_safe
     def update_forced_unpin_count(self, delta: int):
         self.interval_forced_unpin_count += delta
 
@@ -627,6 +642,8 @@ class LMCStatsMonitor:
         self.interval_local_cpu_evict_count = 0
         self.interval_local_cpu_evict_keys_count = 0
         self.interval_local_cpu_evict_failed_count = 0
+
+        self.interval_store_dropped_chunk_count = 0
 
         self.interval_forced_unpin_count = 0
 
@@ -792,6 +809,7 @@ class LMCStatsMonitor:
             interval_local_cpu_evict_count=self.interval_local_cpu_evict_count,
             interval_local_cpu_evict_keys_count=self.interval_local_cpu_evict_keys_count,
             interval_local_cpu_evict_failed_count=self.interval_local_cpu_evict_failed_count,
+            interval_store_dropped_chunk_count=self.interval_store_dropped_chunk_count,
             interval_forced_unpin_count=self.interval_forced_unpin_count,
             local_cache_usage_bytes=self.local_cache_usage_bytes,
             remote_cache_usage_bytes=self.remote_cache_usage_bytes,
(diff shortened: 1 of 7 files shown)


lmcache/observability.py
```diff
@@ -1038,6 +1056,13 @@ def __init__(
             labelnames=labelnames,
         )
 
+        self.counter_store_dropped_chunk_count = self._create_counter(
+            name="lmcache:store_dropped_chunk_count",
```
DongDongJu, line 1060: Since its saying "Total number of KV chunks dropped on the store path because CPU memory allocation failed".
Can we rename this counter to `store_drop_event_count` or `store_admission_failure_count`?

DongDongJu (changes requested):
Hello @dvaldivia, Thanks for the contribution!.
I think the retry logic is okay. 
I left one nit

PR #3638, opened 2026-06-11
[Bug] async_save_bytes_to_disk: add exception safety for write failures

### LINKED ISSUE #3533: [Bug] async_save_bytes_to_disk lacks exception safety, leading to self-deadlock
**Label**
`bug`, `memory-management`, `async-io`

---

**Describe the bug**

When `async_save_bytes_to_disk` in `LocalDiskBackend` encounters an I/O exception (disk full, alignment failure, NFS timeout, etc.), it skips the `ref_count_down()` cleanup. The MemoryObj leaks with a non-zero refcount. Python GC eventually collects it, triggering `__del__` → `parent_allocator.free(self)` → `AddressManager.free()` → acquires `self._lock`. If another thread is simultaneously holding `self._lock` inside `batched_allocate()`, the process self-deadlocks, because `AddressManager._lock` is a non-reentrant `threading.Lock`.

The root cause chain:

```
disk full / IO error
  → write_file() raises (local_disk_backend.py:542)
  → ref_count_down() skipped (local_disk_backend.py:555)
  → MemoryObj leaks: ref_count > 0
  → GC collects → __del__ → parent_allocator.free(self)
  → AddressManager.free() acquires self._lock
  → if batched_allocate() already holds _lock → self-deadlock
```

Three concrete defects in the codebase:

1. **No `try/finally` around `write_file`** — `lmcache/v1/storage_backend/local_disk_backend.py:532-559`
   - `ref_count_up()` is called at line 357 before async submission.
   - `write_file()` at line 542 can raise for many reasons.
   - `ref_count_down()` at line 555 is never reached on exception.

2. **`__del__` calls `free()` with side effects** — `lmcache/v1/memory_management.py:504-520`
   - On GC collection of a leaked MemoryObj, `__del__` warns and calls `parent_allocator.free(self)`.
   - This acquires `AddressManager._lock` from inside the GC thread.

3. **`AddressManager._lock` is `threading.Lock`, not `RLock`** — `lmcache/v1/memory_management.py:986`
   - `batched_allocate()` holds this lock during iteration over `self._explicit_list` (lines 1100-1128).
   - If GC fires during iteration, `__del__` tries to re-acquire the same lock → deadlock.

**To Reproduce**

1. Configure LMCache with `store_location: "LocalDiskBackend"`, async loading enabled.
2. Set `max_local_disk_size` larger than the actual available disk space on the L2 partition.
3. Run a multi-round benchmark workload that continuously stores KV cache to L2.
4. Wait for `write_file` to fail → observe GC WARNING logs → service hangs with no crash.

Minimal reproduction config:

```yaml
chunk_size: 256
local_cpu: true
max_local_cpu_size: 100
local_disk: "file:///path/to/small/partition/"
max_local_disk_size: 100
enable_async_loading: true
store_location: "LocalDiskBackend"
```

Observed logs:

```
LMCache WARNING: MemoryObj at 0x7f... is being garbage collected with ref_count=1, pin_count=0
```

`py-spy dump` shows MainThread stuck at:

```
PyThread_acquire_lock_timed
  → AddressManager.free
  → TensorMemoryObj.__del__
  → (GC collection)
  → AddressManager.batched_allocate
```

**Expected behavior**

`async_save_bytes_to_disk` should guarantee cleanup via `try/finally`, ensuring `ref_count_down()` (and optionally `remove_put_task`) always execute regardless of I/O success or failure. The MemoryObj should not leak, the GC should not need to invoke `__del__` → `free()` on leaked objects, and no self-deadlock should occur.

**Desktop (server environment)**

- OS: Linux (Docker, SGLang image `sglang-lmcache:05201706`)
- LMCache version: 0.4.5
- Python: 3.12
- GPU: NVIDIA L20
- CPU staging pool: 100 GB
- L2 (NVMe): local filesystem, 70 GB partition (vs 100 GB configured limit)
- Serving engine: SGLang
- Workload: multi-round QA benchmark (8 users, 5000-token history, qps=1)

**Additional context**

This is not limited to disk-full scenarios. Any exception in `write_file` can trigger the same chain:

| Trigger | Likelihood |
|---------|-----------|
| Disk space exhausted | Medium |
| O_DIRECT alignment failure under high concurrency | Medium |
| NFS timeout / remote filesystem disconnect | Low |
| Filesystem quota exceeded | Low |
| NVMe hot-unplug | Low |

Each trigger has low individual probability, but collectively the risk surface is non-trivial. The consequence (service hang requiring a restart) is severe.

### Proposed fix

**Minimal fix** — wrap `async_save_bytes_to_disk` cleanup in `try/finally`:

```python
def async_save_bytes_to_disk(self, key, memory_obj, on_complete_callback=None):
    kv_chunk = memory_obj.tensor
    assert kv_chunk is not None
    buffer = memory_obj.byte_array
    path = self._key_to_path(key)

    size = len(buffer)
    self.usage += size
    self.stats_monitor.update_local_storage_usage(self.usage)

    try:
        self.write_file(buffer, path)
    finally:
        memory_obj.ref_count_down()
        if sys.exc_info()[1] is None:
            size = memory_obj.get_physical_size()
            shape = memory_obj.metadata.shape
            dtype = memory_obj.metadata.dtype
            fmt = memory_obj.metadata.fmt
            cached_positions = memory_obj.metadata.cached_positions
            self.insert_key(key, size, shape, dtype, fmt,
                            cached_positions=cached_positions)
    self.disk_worker.remove_put_task(key)
```

**Defense-in-depth** (two additional layers):

- Change `AddressManager._lock` to `threading.RLock()` in `memory_management.py:986` to allow re-entrant acquisition from GC.
- Add a `_freed` guard flag in `__del__` to prevent double-free if `free()` is called from multiple paths.

### Affected files

| File | Lines | Issue |
|------|-------|-------|
| `lmcache/v1/storage_backend/local_disk_backend.py` | 532–559 | Missing try/finally |
| `lmcache/v1/memory_management.py` | 986 | `threading.Lock` → `RLock` |
| `lmcache/v1/memory_management.py` | 504–520 | `__del__` without re-entrancy guard |

mikenosuspend:
## Summary

Fixes #3533 — when `async_save_bytes_to_disk` encounters an I/O exception, `ref_count_down` was skipped, causing MemoryObj leak → GC → `__del__` → self-deadlock on `AddressManager._lock`.

## Changes

`lmcache/v1/storage_backend/local_disk_backend.py` — wrapped `write_file` in `try/except/finally`:
- `ref_count_down` and `remove_put_task` moved to `finally` — always execute
- `self.usage` rolled back in `except` on failure, then re-raised
- `insert_key` and `on_complete_callback` only run on success

## Before

\```python
self.write_file(buffer, path)
# ... (ref_count_down skipped on exception)
memory_obj.ref_count_down()
self.insert_key(key, ...)
self.disk_worker.remove_put_task(key)
\```

## After

\```python
try:
    self.write_file(buffer, path)
except Exception:
    self.usage -= size
    self.stats_monitor.update_local_storage_usage(self.usage)
    raise
finally:
    memory_obj.ref_count_down()
    self.disk_worker.remove_put_task(key)
# Only reached on success:
self.insert_key(key, ...)
\```

## Verification

- Reproduced GC Warning (`MemoryObj at 0x... is being garbage collected with ref_count=1`) via monkey-patched `write_file` failures
- Reproduced self-deadlock — confirmed with `py-spy dump`
- After fix: no GC warnings, no deadlock under same conditions


mikenosuspend:
@DongDongJu ready for review. Fixes #3533.

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -539,25 +539,27 @@ def async_save_bytes_to_disk(
...
-        self.write_file(buffer, path)
-
-        # ref count down here because there's a ref_count_up in
-        # `submit_put_task` above.
-        # Ref count down better be before `insert_key` for testing
-        # purposes (e.g., testing mem_leak).
-        # TODO(Jiayi): This could be problematic if the
-        # freed memory object is immediately reused.
+        try:
+            self.write_file(buffer, path)
```
DongDongJu, line 543: I would prefer the same pattern already used by the mp filesystem L2 adapter: write to a temp path, replace() atomically on success, and unlink the temp file on failure.

lmcache/v1/storage_backend/local_disk_backend.py
```diff
@@ -539,25 +539,27 @@ def async_save_bytes_to_disk(
         self.stats_monitor.update_local_storage_usage(self.usage)
```
DongDongJu, line 539: Is this need to be under lock as well? WDYT?

DongDongJu (changes requested):
Hello @mikenosuspend,
Thanks for the contribution!. Do you plan to address mp `StoreController` too? 
Looking for the next PR.
I left few comments.
Also, Please add a regression test that forces write_file() to raise and verifies refcount cleanup, put-task cleanup, no insert_key(), no callback, and rollback of usage / current_cache_size.

lmcache/v1/storage_backend/local_disk_backend.py
mikenosuspend: Good catch. The original code already does self.usage += size without a lock on the success path (line 538 of the original), so this PR deliberately keeps the same locking convention for the rollback path. Making usage counter updates thread-safe would require touching both paths, plus the remove() method which does use disk_lock — that inconsistency existed before this PR. I think it's better handled as a separate follow-up to keep this fix focused on the exception safety / deadlock issue.

lmcache/v1/storage_backend/local_disk_backend.py
mikenosuspend: Good catch on the temp-file pattern — updated write_file to follow the same approach as fs_connector.py: write to tmp_path, os.replace on success, os.unlink on failure. The try/except/finally in async_save_bytes_to_disk is retained to guarantee ref_count_down and remove_put_task on the caller side. Code is ready, I'll run tests on our GPU cluster next week and push the update shortly after.

mikenosuspend:
@DongDongJu Thanks for the thorough review! All comments should be addressed in the latest push. Here's a summary:

Changes applied:
1. Temp-file + atomic replace — write_file now follows the same pattern as fs_connector.py: writes to tmp_path first, os.replace on success, os.unlink on failure. This prevents readers from seeing partially-written files in multi-process scenarios. O_DIRECT path uses mmap alignment with buffered I/O fallback.

2. Lock on rollback path — usage and current_cache_size rollbacks are now under self.disk_lock in the except block (per your comment).

3. Regression test — added TestAsyncSaveBytesToDiskExceptionSafety in tests/v1/storage_backend/test_local_disk_backend.py. Mocks write_file to raise OSError and verifies: ref_count_down called, remove_put_task called, insert_key NOT called, callback NOT called, usage rolled back, current_cache_size rolled back.

4. Real-world testing — deployed on SGLang IP mode (NVIDIA L20, 8 users × QPS=4 × 5000-token history, 6000s benchmark). No GC warnings, no deadlock, zero .tmp file residue.

mp StoreController — happy to address in a follow-up PR.

Future follow-ups (separate PRs):
- AddressManager._lock → RLock for defense-in-depth
- _freed guard in __del__ to prevent double-free
- mp mode StoreController exception safety

Let me know if anything else needs adjustment.

mikenosuspend:
@DongDongJu gentle ping — all review comments have been addressed in the latest commits. Would appreciate a re-review when you have bandwidth. Thanks!
Also, looks like the comprehensive/integration/multiprocess Buildkite pipelines and the Code Quality check haven't been triggered — likely a fork PR restriction. Could you approve/trigger them when you get a chance?

PR #3642, opened 2026-06-11
refactor(build): extract extension build into profile-based architecture

maobaolong:

setup.py was a big mess file — adding a backend meant surgery on the same file.
This PR splits it into self-contained profiles:

setup_extensions/
├── __init__.py              # exports BuildPolicy, BuildProfile
├── common_cpp.py            # common C++ extension build
├── policy.py                # BuildPolicy — discover, resolve, build
└── build_profiles/
    ├── __init__.py           # BuildProfile ABC
    ├── cuda.py
    ├── rocm.py
    ├── sycl.py
    └── musa.py

Adding a new backend: drop a .py in build_profiles/, subclass BuildProfile, done.
No module list to update — pkgutil.iter_modules finds it.

- sys.path.insert in setup.py so pip install -e works under PEP 517 isolation
- CUDA detect now nvcc-based (not torch.cuda.is_available) so CI builds work
- Explicit BUILD_WITH_* no longer silently falls back to another backend

**If applicable**:

### DIFF
--- a/setup_extensions/build_profiles/__init__.py
+++ b/setup_extensions/build_profiles/__init__.py
@@ -2,7 +2,7 @@
 """Base classes for the platform extension build pattern.
 
 Each hardware platform (CUDA, ROCm, SYCL, MUSA, ...) implements
-:class:`PlatformStrategy`.
+:class:`BuildProfile`.
 The :class:`BuildPolicy` orchestrates auto-detection, fallback, and building.
 """
 
@@ -18,8 +18,8 @@ if TYPE_CHECKING:
     from setup_extensions.common_cpp import CommonExtSpec
 
 
-class PlatformStrategy(ABC):
-    """Strategy for building platform-specific extensions.
+class BuildProfile(ABC):
+    """Build profile for platform-specific extensions.
 
     Subclasses must define:
         name     – unique identifier string.
@@ -27,7 +27,7 @@ class PlatformStrategy(ABC):
                    selection.
 
     Subclasses must implement:
-        detect() – auto-detect if this platform's hardware/compiler is present.
+        detect() – auto-detect if this profile's hardware/compiler is present.
         build()  – return ``(ext_modules, cmdclass)`` for extensions.
 
     Subclasses may override:
@@ -92,7 +92,7 @@ class PlatformStrategy(ABC):
     # ------------------------------------------------------------------
 
     def is_explicitly_requested(self) -> bool:
-        """Return True when this platform was selected via env var."""
+        """Return True when this profile was selected via env var."""
         if not self.env_var:
             return False
         # Standard
@@ -102,12 +102,12 @@ class PlatformStrategy(ABC):
 
     @abstractmethod
     def detect(self) -> bool:
-        """Auto-detect if this platform's toolchain / hardware is available."""
+        """Auto-detect if this profile's toolchain / hardware is available."""
         ...
 
     @abstractmethod
     def build(self) -> tuple[list["Extension"], dict]:
-        """Build platform-specific extension modules.
+        """Build profile-specific extension modules.
 
         Returns:
             ``(ext_modules, cmdclass)`` tuple.
@@ -126,7 +126,7 @@ class PlatformStrategy(ABC):
     def default_cxx_flags(self) -> list[str]:
         """Default C++ flags to hand to downstream consumers (e.g. optional
         L2 storage backends) that need a representative set of flags
-        compatible with this platform.
+        compatible with this profile.
 
         Subclasses override when their default ABI differs from the empty
         baseline.
@@ -136,6 +136,6 @@ class PlatformStrategy(ABC):
     def requirements_file(self) -> Optional[str]:
         """Core requirements file name, relative to ``requirements/``.
 
-        Return ``None`` when this platform has no extra deps.
+        Return ``None`` when this profile has no extra deps.
         """
         return None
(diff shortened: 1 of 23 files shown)


maobaolong:
@ApostaC Would you like to take a look at this PR? Thanks!

hlin99 (commented):
like this device.py as strategy

setup_extensions/strategies/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        import sys
+
+        if os.environ.get("NO_CUDA_EXT", "0") == "1":
+            print(
+                "warning: NO_CUDA_EXT is deprecated; use NO_NATIVE_EXT=1 instead.",
+                file=sys.stderr,
+            )
+        return (
+            os.environ.get("NO_NATIVE_EXT", "0") == "1"
+            or os.environ.get("NO_CUDA_EXT", "0") == "1"
```
ApostaC, line 82: (Maybe I'm wrong) IIUC, `NO_NATIVE_EXT` skips all the native C++ modules (also includes non-CUDA modules). If that's true, should we drop the `or os.environ.get("NO_CUDA_EXT", "0") == "1"` condition?

setup_extensions/build_profiles/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+            or os.environ.get("NO_CUDA_EXT", "0") == "1"
+        )
+
+    @classmethod
+    def is_gpu_ext_disabled(cls) -> bool:
+        """Return True when GPU extensions are disabled."""
+        # Standard
+        import os
+
+        return os.environ.get("NO_GPU_EXT", "0") == "1"
```
ApostaC, line 98: `NO_GPU_EXT` and `NO_CUDA_EXT` are a bit confusing here. Maybe add a bit more docstrings (or documentations) to explain it?

setup_extensions/strategies/cuda.py
```diff
@@ -0,0 +1,93 @@
...
+    # Third Party
+    from setuptools.extension import Extension
+
+# First Party
+from setup_extensions.strategies import BuildStrategy
+
+ENABLE_CXX11_ABI = os.environ.get("ENABLE_CXX11_ABI", "1") == "1"
+
+
+class CudaStrategy(BuildStrategy):
```
ApostaC: Regarding the naming:
- `StorageBackendStrategy` is under the folder `storage_backends/`
- `BuildStrategy` is under the folder `strategies/`

A proposal that may improve the consistency:
- `strategies/` --> `platforms` (or `accelerators/`)
- `BuildStrategy` --> `PlatformStrategy` (or `AcceleratorStrategy`)

WDYT?

ApostaC (commented):
This is very clear! I was hoping someone could improve the `setup.py` for a long while. Thanks a lot!

A separate question: do you think it makes sense to use CMake or other building tools instead of relying on torch.cpp_extension in the future? If that's the case, can the current refactored interface still be helpful?

setup_extensions/strategies/cuda.py
maobaolong: @ApostaC How about these?
- `strategies` -> `build_profiles` 
-  `BuildStrategy` -> `BuildProfile`
- `StorageBackendStrategy` -> `StorageBackendProfile`
-  `storage_backends` -> `storage_backend_profiles`

setup_extensions/build_profiles/__init__.py
maobaolong: done

setup_extensions/strategies/__init__.py
maobaolong: yeah, done

maobaolong:
> This is very clear! I was hoping someone could improve the `setup.py` for a long while. Thanks a lot!
> 
> A separate question: do you think it makes sense to use CMake or other building tools instead of relying on torch.cpp_extension in the future? If that's the case, can the current refactored interface still be helpful?

@ApostaC Thanks for the review. About `CMake` or other better building tools, let's do some follow-up research, if we find a better approach, we can replace what we're currently using.

setup_extensions/strategies/cuda.py
ApostaC: This sounds good!

ApostaC (approved):
LGTM!

setup_extensions/build_profiles/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        Controlled by the ``NO_NATIVE_EXT`` environment variable.
+        When ``True``, all native C++ extensions are skipped — including
+        both common extensions (Redis, filesystem, storage manager)
+        and GPU backend extensions.
+        """
+        # Standard
+        import os
+        import sys
+
+        if os.environ.get("NO_CUDA_EXT", "0") == "1":
```
hlin99, line 75: there is some historical story here....

NO_CUDA_EXT is widely used by many tools including ci, and packaging.... 

so we don't want to break NO_CUDA_EXT semantic, that's why current logic NO_CUDA_EXT==NO_NATIVE_EXT, and NO_GPU_EXT is used for skipping cuda backend.

@deng451e  please correct me if i'm wrong

maobaolong:
@hlin99 Thanks for the review and help to correct the logic. I push a new commit for fix the logic of `NO_xxx_EXT`, PTAL.

hlin99 (approved):
LGTM

PR #3643, opened 2026-06-11
fix(vllm): wire MP cache reset

milesial:
## Summary
- Add scheduler-role `LMCacheMPConnector.reset_cache()` so vLLM cache reset can reach the MP connector instead of falling through the base no-op.
- Add `LMCacheMPSchedulerAdapter.reset_cache()` to clear scheduler lookup bookkeeping and send `RequestType.CLEAR` to the MP server.
- Add focused MP adapter tests for successful CLEAR and timeout/unhealthy behavior.

## Root cause
The MP vLLM connector did not expose a `reset_cache()` hook, so vLLM's reset connector path could not reliably propagate the external cache clear into the LMCache MP server. The adapter also needed to clear local scheduler lookup state around the server clear.

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -802,6 +802,20 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
     # Scheduler-side methods
     # ==============================
 
+    def reset_cache(self) -> bool | None:
+        """Reset LMCache MP cache state from the scheduler role.
+
+        Returns:
+            True when the MP server clears successfully, False on failure, and
+            None for worker-role connectors.
+        """
+        if self.role != KVConnectorRole.SCHEDULER:
+            return None
+
+        success = self.scheduler_adapter.reset_cache()
+        self.request_trackers.clear()
+        return success
+
     def get_num_new_matched_tokens(
         self,
         request: "Request",
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -758,6 +758,33 @@ class LMCacheMPSchedulerAdapter:
         self._pending_lookups.discard(request_id)
         self._finished_lookup_results.pop(request_id, None)
 
+    def reset_cache(self) -> bool:
+        """Clear server-side cache and scheduler-side lookup bookkeeping.
+
+        Returns:
+            True if the server acknowledged the clear request, False if the
+            request timed out.
+        """
+        self._pending_lookups.clear()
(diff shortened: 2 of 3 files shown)


maobaolong:
@milesial Sorry for the late reply, would you like to active work on this feature? I'll active to review this PR if you like.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -802,6 +802,20 @@ def get_kv_connector_stats(self) -> "KVConnectorStats | None":
...
+
+        Returns:
+            True when the MP server clears successfully, False on failure, and
+            None for worker-role connectors.
+        """
+        if self.role != KVConnectorRole.SCHEDULER:
+            return None
+
+        success = self.scheduler_adapter.reset_cache()
+        self.request_trackers.clear()
```
maobaolong, line 816: Do not drop live request state during a failed local reset

vLLM calls `connector.reset_cache()` whenever `reset_external=true` even when `kv_cache_manager.reset_prefix_cache()` has already returned `False` because requests still hold blocks (`Scheduler.reset_prefix_cache` combines the two results only afterwards). Clearing this map removes the state that the next `build_connector_meta()` / `update_state_after_alloc()` reaches through `_get_request_tracker()`, so a reset issued during traffic can turn a clean reset failure into an assertion failure on the next scheduler step. Please refuse/skip the connector reset while requests are active (or otherwise preserve/rebuild their tracker state), and add a regression test for `reset_external=true` with a running request.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -758,6 +758,33 @@ def cleanup_lookup_result(self, request_id: str) -> None:
...
+
+        Returns:
+            True if the server acknowledged the clear request, False if the
+            request timed out.
+        """
+        self._pending_lookups.clear()
+        self._finished_lookup_results.clear()
+
+        try:
+            send_lmcache_request(
```
maobaolong, line 772: Do not force-clear MP storage while worker transfers are in flight

This RPC reaches `ManagementModule.clear()` -> `storage_manager.clear(force=True)`. `L1Manager.clear()` explicitly documents that `force=True` frees read/write-locked objects and may corrupt in-flight store/prefetch operations. The scheduler adapter cannot establish quiescence here: pending store/retrieve futures live in the worker adapters, and the scheduler tracker is removed before an async store finishes. Thus even an empty scheduler tracker map would not make this CLEAR safe. Please coordinate a drain/reject across workers, or make the server-side clear preserve locked objects and return failure, and cover the race with a concurrency regression.

maobaolong:
As #5091 merged, this PR can be closed, thanks for the contribution @milesial

PR #3646, opened 2026-06-12
fix(operator): install cert-manager in e2e smoke suite

royyhuang:
## Problem

The operator e2e integration test (Buildkite `lmcache-operator-integration-test`) fails in `BeforeSuite`: the controller-manager Deployment never becomes `Available`.

## Root cause

The CacheBlend injection webhook made **cert-manager a hard prerequisite**. `make deploy` (`config/default`) now applies:

- a self-signed `Issuer` + `Certificate` (`cert-manager.io/v1`), and
- a `MutatingWebhookConfiguration` annotated with `cert-manager.io/inject-ca-from`,

and the controller-manager Deployment mounts the cert-manager-issued `webhook-server-cert` secret at `/tmp/k8s-webhook-server/serving-certs`.

On a fresh Kind cluster with no cert-manager, the deploy step is rejected (unknown `cert-manager.io` kinds) and the serving-cert secret never materialises — so the manager pod is stuck `ContainerCreating`, never goes `Available`, and the 3-minute `BeforeSuite` wait times out.

## Fix

Install cert-manager from the e2e `BeforeSuite` before deploying the controller:

- New helpers in `test/utils/utils.go`: `InstallCertManager` (apply upstream manifest, wait for the three cert-manager Deployments to be `Available`), `UninstallCertManager` (best-effort), `IsCertManagerCRDsInstalled` (guard).
- `BeforeSuite` installs cert-manager up front (before the slow image build/load, so its webhook endpoints are warm by deploy time). Install is **skipped** when the cluster already ships cert-manager, and `AfterSuite` only uninstalls what the suite installed — so a shared OpenShift/EKS cluster is left untouched.
- Release is overridable via `CERT_MANAGER_VERSION` (default `v1.16.3`).

Both the M1 (`test-e2e-kind`) and M2/GPU (`test-e2e-gpu-kind`) targets share the single `//go:build e2e` `BeforeSuite`, so both Kind tiers are covered.

`AGENTS.md` updated to document the cert-manager prerequisite and skip/override behaviour.

## Test plan

- `go build ./...` — clean
- `go vet -tags=e2e ./test/...` — clean
- `gofmt` — clean
- Buildkite e2e run on this PR validates the end-to-end path on Kind.

### DIFF
--- a/operator/AGENTS.md
+++ b/operator/AGENTS.md
@@ -41,7 +41,16 @@ details.
 Builds the manager image, loads it into a dedicated Kind cluster
 (`operator-test-e2e-<id>` by default), installs CRDs, deploys the controller,
 runs every `//go:build e2e` spec under `test/e2e/`, then tears the
-cluster down. No prereqs beyond Kind + Docker on `$PATH`.
+cluster down. No prereqs beyond Kind + Docker on `$PATH` (plus network
+egress to GitHub).
+
+The suite installs **cert-manager** into the cluster before deploying the
+controller — it issues the mutating webhook's serving cert and injects the
+CA bundle (see `config/certmanager`), so the controller-manager Deployment
+never reaches `Available` without it. Install is skipped when the cluster
+already ships cert-manager (e.g. an existing OpenShift/EKS cluster), and
+the suite only uninstalls what it installed. Override the release with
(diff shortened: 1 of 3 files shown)


deng451e (approved):
LGTM

ApostaC (approved):
LGTM!

PR #3647, opened 2026-06-12
fix(operator): emit --engine-type blend for CacheBlend engine

royyhuang:

The lmcache server selects the CacheBlend V3 engine via `--engine-type blend` (`lmcache/v1/multiprocess/server.py`); the V2 engine moved to `--engine-type blend_legacy`. The value `blend_v3` is no longer recognized, so the operator-generated CacheBlend DaemonSet started a server that loaded **no** blend module — CacheBlend was silently broken.

This switches the operator's hardcoded engine-type value from `blend_v3` to `blend`, realigning the `CacheBlendEngine` controller with the current server contract.

Value-only change to the `--engine-type` flag. The `blend_v3` / `BlendV3Module` naming (module name) is left untouched — only the CLI flag value changed. Sample manifest and operator docs updated to match.

**If applicable**:

### DIFF
--- a/docs/source/mp/operator.rst
+++ b/docs/source/mp/operator.rst
@@ -603,7 +603,7 @@ for the technique itself.
 
 It has two halves the operator runs together:
 
-- a GPU-resident ``blend_v3`` engine (``lmcache server --engine-type blend_v3``),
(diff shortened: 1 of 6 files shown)


deng451e (approved):
LGTM

ApostaC (approved):
LGTM!

PR #3662, opened 2026-06-12
[MP] [Coordinator] Resync and Evict L2 KVCache through LMCache L2 API

aeon-x:
### MP server — new HTTP endpoints

Three endpoints on the MP server, documented in `docs/design/v1/multiprocess/l2_apis.md` and `docs/source/mp/http_api.rst`:

- `GET /l2/adapters` — enumerate every configured L2 adapter with its `type_name` and primary flag. Operators read this to learn which value to pass as `?adapter=` on the other endpoints.
- `DELETE /l2` — delete a caller-supplied list of keys (idempotent; in-flight-locked keys are skipped). Batch cap 10 000 keys/request.
- `GET /l2/keys` — paginate keys resident in L2, optionally filtered by `model_name`. Backed by `ListObjectsV2` on `S3L2Adapter`; other adapters return 501. `SerdeL2AdapterWrapper` delegates straight to its inner adapter.

`DELETE /l2` and `GET /l2/keys` accept an optional `?adapter=<type_name>` selector; omitting it targets the primary (first-configured) adapter, preserving v1 behavior. First match wins when multiple adapters share a `type_name`.

### Coordinator — active eviction loop

`L2EvictionManager.execute_evictions` now fire-and-forgets a single `DELETE /l2` to a uniformly random registered MP server — one shared bucket means one dispatch evicts for the whole fleet. The LRU is **not** cleared on dispatch; it's cleared by the `delete` event that the MP server's `on_l2_keys_deleted` listener posts back via
`POST /l2/events`. At-least-once semantics, safe because the L2 delete is idempotent.

### Coordinator — startup resync

New `L2ResyncManager` waits for the first MP server to register, then paginates `GET /l2/keys` to seed `L2UsageManager` + `L2EvictionManager` with whatever is already resident. Best-effort: failures fall back to the ongoing usage-event stream.

### Event protocol

`POST /l2/events` now accepts `type: "delete"` alongside `store` and `lookup`. A `delete` event subtracts the key's previously-recorded bytes from per-salt totals (the wire `bytes` field is ignored; the coordinator already has the size on file from the original `store`).

### Coordinator config (env-var only)

- `LMCACHE_MP_COORDINATOR_ENABLE_STARTUP_RESYNC` (default `True`)
- `LMCACHE_MP_COORDINATOR_RESYNC_POLL_INTERVAL` (default `1`)
- `LMCACHE_MP_COORDINATOR_RESYNC_MAX_WAIT` (default `60`)
- `LMCACHE_MP_COORDINATOR_RESYNC_PAGE_SIZE` (default `1000`)

- **Event-driven, not RPC-driven.** The coordinator's usage + LRU state is only cleared when a `delete` event arrives. If dispatch fails or the MP server drops the request, the next eviction cycle re-picks the same keys.
- **No barrier on dispatch.** `execute_evictions` returns as soon as `asyncio.create_task` is scheduled; `L2EvictionManager` keeps strong references to in-flight tasks in `_in_flight_dispatches` and exposes `wait_for_in_flight_dispatches()` for the coordinator's lifespan to drain on shutdown.
- **Adapter selector defaults to primary.** Existing v1 clients (and the coordinator's own eviction/resync flows) don't pass `?adapter=` and continue to target the primary adapter unchanged. The selector is operator surface, not a behavior change for the hot path.
- The startup-resync knobs are env-var-only for now; happy to add matching CLI flags in a follow-up if reviewers prefer.

## If applicable


lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -75,6 +75,33 @@ def usage_fraction(self) -> float:
...
+    it verbatim to the next call to continue. The token is opaque — the
+    caller MUST NOT inspect or modify it.
+
+    Adapter-level: this type carries no adapter identification. The
+    storage manager attaches an ``adapter`` field when returning the
+    page to HTTP callers.
+    """
+
+    entries: tuple[KeyEntry, ...]
+    next_page_token: str | None
```
ApostaC: What does `token` mean in `next_page_token`? 

---
Update: maybe rename it to `has_next_page: bool`?

lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
```diff
@@ -81,9 +205,14 @@ def _object_key_to_string(key: ObjectKey) -> str:
...
-    """Flatten slashes and URL-encode to form a safe HTTP path."""
-    flat = key_str.replace("/", "_")
-    return "/" + url_quote(flat)
+    """URL-encode the object name to form a safe HTTP path.
+
+    ``url_quote`` keeps ``/`` (which is legal inside S3 object names)
+    and percent-encodes other reserved characters like ``@``. The
+    leading ``/`` is required by the HTTP path grammar — S3 strips it
+    when resolving to the object key.
+    """
```
ApostaC: nit: we don't need to touch the docstrings here.

lmcache/v1/distributed/storage_manager.py
```diff
@@ -774,6 +774,36 @@ def reconfigure_l2_adapter(
         result["adapter_index"] = adapter_index
         return result
 
+    def primary_l2(self) -> tuple[AdapterDescriptor, L2AdapterInterface]:
```
ApostaC: Why do we need `primary_l2`? What if the user has more than 1 L2 storage? (Like both local disk + remote S3)

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -75,6 +75,33 @@ def usage_fraction(self) -> float:
         return self.total_bytes_used / self.total_capacity_bytes
 
 
+@dataclass(frozen=True)
+class KeyEntry:
+    """One entry in a :class:`KeyListPage` — the JSON-safe
+    :class:`EncodedObjectKey` (so it serializes cleanly when returned
+    over HTTP) and the bytes the adapter has on disk for it."""
```
ApostaC: nit: docstrings too long, we only need the first sentence: "One entry in a :class:`KeyListPage` including the encoded object key and its object size"

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -75,6 +75,33 @@ def usage_fraction(self) -> float:
...
+class KeyListPage:
+    """A page of keys returned by ``L2AdapterInterface.list_l2_keys``.
+
+    ``next_page_token`` is ``None`` when the listing is exhausted; pass
+    it verbatim to the next call to continue. The token is opaque — the
+    caller MUST NOT inspect or modify it.
+
+    Adapter-level: this type carries no adapter identification. The
+    storage manager attaches an ``adapter`` field when returning the
+    page to HTTP callers.
```
ApostaC: nit: for dataclass, we usually have the member-specific docstrings under each member.

E.g.
```python
class KeyListPage:
    """A page of keys returned by ``L2AdapterInterface.list_l2_keys``.
    """

    entries: tuple[KeyEntry, ...]
    """The keys in the current page.
    """

    next_page_token: str | None
    """None means this is the last page. Otherwise, there are remaining pages.
    """
```

ApostaC (commented):
Mostly LGTM.

One problem is that the doc string is a bit too long. AI-generated docstrings tend to describe the internal implementation of the function and how it relates to other modules' implementation. We should make it shorter and focus on the expected semantics of the function instead of implementation details.

lmcache/v1/distributed/l2_adapters/base.py
aeon-x: it is the cursor to tell where to start loading the next page.

lmcache/v1/distributed/l2_adapters/base.py
aeon-x: fixed

lmcache/v1/distributed/l2_adapters/base.py
aeon-x: fixed

lmcache/v1/distributed/storage_manager.py
aeon-x: fixed

lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
aeon-x: fixed

lmcache/v1/multiprocess/http_apis/l2_api.py
```diff
@@ -0,0 +1,241 @@
...
+    return {
+        "adapters": [
+            {"index": i, "type_name": desc.type_name, "primary": i == 0}
+            for i, (desc, _) in enumerate(adapters)
+        ]
+    }
+
+
+@router.delete("/l2", response_model=None)
+async def delete_l2(
```
ApostaC: nit: can we name this to `delete_from_l2`? `delete_l2` sounds like deleting an L2 adapter.

ApostaC (approved):
Otherwise LGTM!

lmcache/v1/multiprocess/http_apis/l2_api.py
aeon-x: updated!

deng451e (approved):

PR #3663, opened 2026-06-12
fix(nixl): support dynamic layerwise file transfers

### LINKED ISSUE #3559: [RFC] : NIXL GDS_MT dynamic FILE backend behavior with use_layerwise=true
## Summary

I am trying to confirm whether the NIXL dynamic FILE/GDS_MT backend is expected to work with `use_layerwise: true`, or whether there is a layerwise sizing/keying issue in the current implementation.

In my local vLLM + LMCache benchmarking, layerwise mode with NIXL GDS_MT produced retrieval failures such as:

```text
RuntimeError: Layerwise retrieve returned 0/79 chunks for layer 51; refusing partial KV load
```

I also saw a `StopIteration` from `save_kv_layer` in one run, and CUDA OOM when trying to compensate by increasing the NIXL buffer size. I am not sure yet whether these are all the same root cause, so I wanted to ask maintainers before turning this into a PR.

If this is a real issue, I am happy to pick it up and work on a fix.

## Configuration

The relevant setup was approximately:

```yaml
use_layerwise: true
save_unfull_chunk: false
extra_config:
  enable_nixl_storage: true
  nixl_backend: GDS_MT
  nixl_path: <local nvme path>
  nixl_pool_size: 0   # dynamic storage
  nixl_backend_params:
    thread_count: 4
```

The workload was vLLM + LMCache with a long-prefill/short-output benchmark. For Qwen/Qwen3-32B, each layerwise object appeared to be around 1 MiB logically for a 256-token chunk, while a full KV chunk across all layers is much larger.

## Suspected issue 1: dynamic NIXL storage may use full KV chunk alignment in layerwise mode

In `lmcache/v1/storage_backend/nixl_storage_backend.py`, `NixlStorageConfig.from_cache_engine_config()` appears to compute alignment from the full metadata KV shape:

```python
align_bytes = get_size_bytes(
    [torch.Size(metadata.kv_shape)], [metadata.kv_dtype]
)
```

This is around lines 188-191 in my local checkout at commit `b0d1e1e74e89da1d0c9764475c5e6b35a4989af1`.

Then `NixlDynamicStorageAgent.create_batched_storage_handler()` accepts a single `page_size` and registers each FILE/object descriptor with that size:

```python
reg_list.append((0, page_size, descs[i].device_id, descs[i].meta_info))
xfer_desc.append((0, page_size, descs[i].device_id))
```

This is around lines 581-587.

Later, dynamic `storage_to_mem` / `mem_to_storage` appear to pass `self.memory_allocator.align_bytes` as that `page_size`, and the debug logging reports `page_size * len(keys)` as transfer size around lines 1479-1481 and 1607-1609.

My question: in `use_layerwise: true`, should the dynamic NIXL FILE/GDS_MT path instead register and transfer the logical per-layer `MemoryObj` size, e.g. something like `mem_obj.metadata.get_size()`, rather than the allocator alignment/full-chunk size?

If the current behavior is intentional, I may be misunderstanding how the layerwise object shape is represented in this backend.

## Suspected issue 2: presence cache keying may be too coarse for layerwise objects

The same backend seems to use `key.chunk_hash` for local presence cache updates/checks in several places, for example:

- discard on failed read around line 1473
- add on successful read around line 1477
- add on successful write around line 1614
- presence check around line 1644

In layerwise mode, it looks like multiple per-layer objects can share the same chunk hash and differ by layer identity. Should this cache use the full `CacheEngineKey`/formatted object key, including layer id, rather than only `chunk_hash`?

## Observed symptoms

The most useful failure I saw was:

```text
RuntimeError: Layerwise retrieve returned 0/79 chunks for layer 51; refusing partial KV load
```

I also saw:

```text
StopIteration
...
lmcache/integration/vllm/vllm_v1_adapter.py", line 1074, in save_kv_layer
    next(layerwise_storer)
```

And after increasing the NIXL buffer size substantially:

```text
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 496.00 MiB
```

One older debug run also had:

```text
GDS_MT: warning: buffer registration failed - will use compat mode: error=5030
```

but I only saw that in one old debug run, so I do not think that warning is the main issue here.

## Questions

1. Is NIXL dynamic FILE/GDS_MT intended to support `use_layerwise: true` today?
2. In layerwise mode, should the NIXL backend allocate/register/transfer per-layer object sizes rather than full KV chunk / allocator-aligned sizes?
3. Should the NIXL presence cache key include full layerwise object identity instead of only `chunk_hash`?
4. If this is a bug, would a minimal fix in `nixl_storage_backend.py` be acceptable? I can work on a PR.

Thanks! I wanted to check the intended design before sending a patch, since this touches NIXL storage semantics and layerwise retrieval.

maheshrbapatu:

Dynamic NIXL FILE/GDS-style storage can store one object per layer when `use_layerwise=True`, but the old dynamic path still treated those layer objects like full KV chunks in several places. That meant:

- allocator layout was derived from full-chunk metadata instead of the layerwise layout
- storage and memory transfer descriptors used full object/page sizes instead of each object's logical layer size
- presence/progress state was keyed by chunk hash, so different layer objects for the same chunk could collide
- FILE batched contains checks queried the formatted object key instead of the full cache path

This PR fixes the dynamic NIXL path by deriving the allocator layout from layerwise metadata, registering dynamic storage and memory descriptors with per-object logical transfer sizes, and keying presence/progress state by the formatted storage object key.

Fixes #3559

- The change is scoped to dynamic NIXL FILE/GDS-style transfers; static NIXL registration keeps the existing flow.
- Cleanup paths were hardened so allocated memory objects, NIXL handles, and FILE descriptors are released if transfer handle acquisition or async submission fails.
- The focused regression tests cover layerwise key collisions, FILE batched lookup paths, logical transfer sizes, and failure cleanup behavior.

**Testing**:

- `ruff check lmcache/v1/storage_backend/nixl_storage_backend.py tests/v1/test_nixl_batched_contains.py tests/v1/test_nixl_storage.py`
- `git diff --check`
- `python -m py_compile lmcache/v1/storage_backend/nixl_storage_backend.py tests/v1/test_nixl_batched_contains.py tests/v1/test_nixl_storage.py`
- `docker run --rm --entrypoint /bin/bash --user $(id -u):$(id -g) -e HOME=/tmp -e PYTHONPATH=/workspace/LMCache -v /localhome/local-mbapatu/LMCache-pr-3559:/workspace/LMCache -w /workspace/LMCache lmcache/vllm-openai:cu13-arm64-local-layernixlfix -lc "python -m pytest -q tests/v1/test_nixl_batched_contains.py tests/v1/test_nixl_storage.py"`
  - result: 33 passed, 4 skipped; skipped tests are the cuFile/GDS variants when the local test environment cannot register GDS file handles.

**If applicable**:


maheshrbapatu:
Hello Team,
Can someone help with review of this PR

lmcache/v1/storage_backend/nixl_storage_backend.py
```diff
@@ -642,13 +679,22 @@ def __init__(
...
 
-    def create_batched_storage_handler(self, descs: list[NixlDesc], page_size: int):
+    def create_batched_storage_handler(
+        self, descs: list[NixlDesc], xfer_sizes: Sequence[int] | int
+    ):
+        if isinstance(xfer_sizes, int):
+            sizes = [xfer_sizes] * len(descs)
+        else:
+            sizes = list(xfer_sizes)
+        assert len(sizes) == len(descs)
```
DongDongJu: please use an explicit `ValueError` instead of `assert` for runtime validation if you touch this again.

DongDongJu (changes requested):
Hello @maheshrbapatu,
Thanks for the contribution.
I left one nit.

lmcache/v1/storage_backend/nixl_storage_backend.py
mbapatu-spec: Thanks for the reivew @DongDongJu . I have addressed the comment.

maheshrbapatu:
Hello Team,
Can anyone from take a look at this whenever you have time,
Thanks !

maheshrbapatu:
Hi Team
Can someone please review this

DongDongJu:
Hello, @maheshrbapatu Please fix the ci first `AttributeError: 'NixlDynamicStorageAgent' object has no attribute 'buffer_device_id`.

maheshrbapatu:
Looks like approved CI checks are passing now,
Apologies for overlooking error earlier.
Can someone please help with triggering pending CI workflows

maheshrbapatu:
Hello, 
Can someone please help on this

maheshrbapatu:
@DongDongJu can we merge this change?
If this is not required at the moment, I can just close the PR too.

PR #3664, opened 2026-06-12
[Feat] Server-side worker liveness tracking and reaping (2/2)

royyhuang:

Second of two PRs (replacing the closed #3231, reworked per review). #3631 added the adapter-side groundwork; this turns on **server-side reaping**: when a vLLM worker dies silently, the MP server reaps its `GPUCacheContext` and CUDA IPC handles instead of leaking them forever.

- **PING carries `instance_id`** (`[] → [int | None]`, kept on BLOCKING/NORMAL dispatch). The server refreshes per-instance `last_seen` on each ping; the worker adapter sends its id, the scheduler sends `None`.
- **Two-tier reaping.** A ping-proven worker is reaped after `worker_reap_timeout_seconds` (120s); a worker that registered but never pinged (warmup, or death before first request) after `worker_registration_grace_seconds` (3600s). Only PING latches "ping-proven"; transfer traffic refreshes `last_seen` without latching.
- **Per-module leaf locks.** The reaper runs off the MQ handler threads, so the GPU and non-GPU registries are lock-guarded. Non-GPU pops its context + strategy as a pair, so a reap racing a re-register can never strand a context without its strategy.
- **Blend reap listener.** `BlendV3Module.drop_instance_state` drops the reaped instance's CB rope state; all blend reads route through the locked `get_context_entry` (the GPU context itself is owned by `GPUTransferModule`, no mirror — reaping the entry frees it directly).
- **Reaper wiring.** `ManagementModule` owns a `PeriodicThread` reaper (scan = timeout/4), started only when reaping is enabled, with the liveness targets / reap listeners injected at composition; it precedes the transfer modules in the module list so `close()` stops it before they clear state.
- **Freeze-gap detection.** A whole-process freeze (SIGSTOP, cgroup freezer, VM migration) past the reap window produces no ping failure and no edge; the heartbeat forces one unhealthy cycle on thaw so the worker re-registers.

**Config**: `--worker-reap-timeout-seconds` (0 disables) and `--worker-registration-grace-seconds`, validated (timeout 0 or ≥30s; grace ≥ timeout). Keep the timeout ≥ 3× the adapter heartbeat interval.

**Design doc**: `docs/design/v1/multiprocess/worker_liveness.md` (covers both PRs).

- Server-side unit tests are in `tests/v1/multiprocess/test_worker_liveness.py` (runs in the MP CI job, not the standard suite). 14 tests, no GPU needed.
- Blend `_cb_*` state stays lock-free (per-key atomic pops in `drop_instance_state`), matching existing blend_v3 style.
- `ManagementModule` receives its liveness targets / reap listeners by **constructor injection** (keeps `MPCacheEngineContext` as pure shared state) — discussed and agreed in review.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -271,14 +271,21 @@ def _raise_server_unreachable(server_url: str, timeout: float) -> NoReturn:
 def send_ping(
     mq_client: MessageQueueClient,
     timeout: float,
+    instance_id: int | None = None,
 ) -> bool:
     """Send a PING request and return the result.
+
+    Args:
+        mq_client: The message queue client.
+        timeout: Seconds to wait for the server's response.
+        instance_id: The worker's instance ID so the server can refresh its
+            liveness, or None for an untracked prober (scheduler adapter).
 
     Returns:
         True if server is healthy, False on timeout or error.
     """
     try:
-        future = send_lmcache_request(mq_client, RequestType.PING, [])
+        future = send_lmcache_request(mq_client, RequestType.PING, [instance_id])
         return future.result(timeout=timeout)
     except TimeoutError:
         return False
@@ -389,6 +396,7 @@ class HeartbeatThread(PeriodicThread):
         mq_client: MessageQueueClient,
         health_event: threading.Event,
         interval: float = DEFAULT_HEARTBEAT_INTERVAL,
+        instance_id: int | None = None,
     ):
         """
         Args:
@@ -398,6 +406,9 @@ class HeartbeatThread(PeriodicThread):
                 Adapters check this event to decide whether to proceed
                 with operations or enter degraded mode.
             interval: Seconds between heartbeat pings and ping timeout.
+            instance_id: The worker's instance ID sent with each PING so the
+                server can refresh its liveness, or None for an untracked
+                prober (the scheduler adapter).
         """
         super().__init__(
             name="lmcache-heartbeat",
@@ -407,6 +418,7 @@ class HeartbeatThread(PeriodicThread):
         self._mq_client = mq_client
         self._health_event = health_event
         self._interval = interval
+        self._instance_id = instance_id
 
         # Optional callback invoked on the unhealthy->healthy edge,
         # before the health event is set. See register_recover_callback.
@@ -445,7 +457,9 @@ class HeartbeatThread(PeriodicThread):
         UNREGISTER must not re-register a ghost context.
         """
         was_healthy = self._health_event.is_set()
-        healthy = send_ping(self._mq_client, timeout=self._interval)
+        healthy = send_ping(
+            self._mq_client, timeout=self._interval, instance_id=self._instance_id
+        )
 
         if self.stop_requested:
             return ThreadRunSummary(
@@ -1181,6 +1195,7 @@ class LMCacheMPWorkerAdapter:
                 mq_client=self.mq_client,
                 health_event=self._health_event,
                 interval=self._heartbeat_interval,
+                instance_id=self.instance_id,
             )
             heartbeat.register_recover_callback(self._reregister_kv_caches_callback)
             heartbeat.start()
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -7,6 +7,7 @@ from typing import Any, Callable, NoReturn, Protocol
 import enum
 import os
 import threading
+import time
 import uuid
 
 # Third Party
@@ -419,6 +420,11 @@ class HeartbeatThread(PeriodicThread):
         self._health_event = health_event
         self._interval = interval
         self._instance_id = instance_id
+        # Monotonic time of the last successful cycle, used for freeze-gap
+        # detection. Baselined in start() (not here): under lazy start,
+        # construction can precede the first cycle by the whole warmup, which
+        # would otherwise misfire the detector on the first ping.
+        self._last_success: float = 0.0
 
         # Optional callback invoked on the unhealthy->healthy edge,
         # before the health event is set. See register_recover_callback.
@@ -449,6 +455,15 @@ class HeartbeatThread(PeriodicThread):
         """
         self._recover_callback = callback
 
+    def start(self) -> threading.Thread | None:
+        """Baseline the freeze-gap clock, then start the thread.
+
+        Returns:
+            The started thread, or None if already running.
+        """
+        self._last_success = time.monotonic()
+        return super().start()
+
     def _execute(self) -> ThreadRunSummary:
         """Run one heartbeat cycle: ping, recover callback, event update.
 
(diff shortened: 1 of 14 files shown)


lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -444,14 +471,31 @@ def _execute(self) -> ThreadRunSummary:
...
             return ThreadRunSummary(
                 success=True,
                 message="stop requested; skipping health update",
             )
 
+        # Freeze-gap: a ping that succeeds while still "healthy" after a long
+        # no-ping gap means the whole process was frozen (SIGSTOP, cgroup
+        # freezer, VM migration) past the server's reap window, so the server
+        # may have reaped us. No edge would fire to re-register, so force one
+        # unhealthy cycle; the next success then takes the recovery edge.
```
ApostaC: Do we really need to consider cases like "SIGSTOP, cgroup freezer, and VM migration"?

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -437,10 +451,107 @@ def context(self) -> MPCacheServerContext:
         """Return the shared engine context. Exposed for testing only."""
         return self._ctx
 
-    @property
-    def cache_contexts(self) -> dict[int, ContextEntry]:
-        """Per-instance GPU context registry."""
-        return self._cache_contexts
+    def get_context_entry(self, instance_id: int) -> ContextEntry | None:
```
ApostaC: nit: maybe call it `get_and_touch_context_entry` because it will also update its last seen time?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -437,10 +451,107 @@ def context(self) -> MPCacheServerContext:
...
+        with self._lock:
+            entry = self._cache_contexts.get(instance_id)
+            if entry is not None:
+                entry.last_seen = now
+                entry.has_liveness_signal = True
+
+    def tracked_instance_count(self) -> int:
+        """Return the number of currently registered instances."""
+        with self._lock:
+            return len(self._cache_contexts)
```
ApostaC, line 503: This is useful! We can also have a Prometheus gauge metric reporting this count. See `_setup_metrics` function in `lookup.py` under the same folder.

lmcache/v1/multiprocess/modules/gpu_transfer.py
```diff
@@ -437,10 +451,107 @@ def context(self) -> MPCacheServerContext:
...
+
+    def reap_stale_instances(
+        self, reap_timeout_s: float, registration_grace_s: float
+    ) -> list[int]:
+        """Reap GPU registrations that have gone silent.
+
+        A ping-proven instance is judged against ``reap_timeout_s``; one
+        that has never pinged against the larger ``registration_grace_s``.
+        Stale entries are collected and popped under the lock, then cleaned
+        up (layout-registry unregister, ``empty_cache``) outside it.
```
ApostaC: nit: we can shorten the docstring.
```suggestion
```

lmcache/v1/multiprocess/modules/non_gpu_transfer.py
```diff
@@ -72,6 +78,13 @@ def __init__(self, ctx: MPCacheServerContext) -> None:
         self._ctx = ctx
         self._non_gpu_contexts: dict[int, NonGPUContextEntry] = {}
         self._strategies: dict[int, TransferStrategy] = {}
+        # Guards _non_gpu_contexts and _strategies together: they are
+        # inserted and popped as a pair so "strategy present iff entry
+        # present" always holds (a reap racing a re-register can never strand
+        # a context without its strategy). The reaper mutates them off the MQ
+        # main loop. Distinct from _pending_shm_lock, which is never held
+        # together with this lock.
```
ApostaC: nit: we only need to keep "Guards _non_gpu_contexts and _strategies together".

lmcache/v1/multiprocess/config.py
```diff
@@ -69,6 +69,39 @@ class MPServerConfig:
     ``run_cache_server``) so metrics, traces, and coordinator state all key on
     the same id. Set via ``--instance-id``; defaults to a random UUID v4."""
 
+    worker_reap_timeout_seconds: float = 120.0
+    """Silence budget (seconds) after which a ping-proven worker's KV cache
+    registration is reaped. 0 disables worker reaping. Keep it >= 3 x the vLLM
+    adapter's heartbeat interval so a few missed pings never reap a live
+    worker."""
```
ApostaC: nit: don't mention vLLM since the MP should be inference engine agnostic.

lmcache/v1/multiprocess/engine_module.py
```diff
@@ -63,3 +63,65 @@ def report_status(self) -> dict:
...
+        """Return the number of currently tracked instances."""
+        ...
+
+
+class InstanceReapListener(Protocol):
+    """A module notified when an instance is reaped, to drop mirrored state.
+
+    Implemented by modules holding a second reference to a reaped instance's
+    resources (e.g. blend rope mirrors of a GPU cache context).
+    """
```
ApostaC: I feel like this can be merged into `InstanceLivenessTarget`.

lmcache/v1/multiprocess/server.py
```diff
@@ -164,25 +169,29 @@ def _build_modules(
...
+        transfer_modules.append(GPUTransferModule(ctx))
+        transfer_modules.append(NonGPUTransferModule(ctx))
     else:
         raise ValueError(
             f"Unsupported supported_transfer_mode '{mp_config.supported_transfer_mode}'"
         )
 
     logger.info("Supported transfer mode: %s", mp_config.supported_transfer_mode)
 
+    blend_modules: list[EngineModule] = []
```
ApostaC: nit: the name `blend_modules` is a bit confusing because there will only be at most one blend module

ApostaC (commented):
Otherwise LGTM. Please rebase the code to fix the naming as well.

lmcache/integration/vllm/vllm_multi_process_adapter.py
royyhuang: Removed. We can add this back once we are really facing the issue related to frozen process. Most likely not.

lmcache/v1/multiprocess/modules/gpu_transfer.py
royyhuang: Fixed

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
royyhuang: I would like to defer this to another small PR, simply keeping this PR focusing on liveness tracking and reaping.

lmcache/v1/multiprocess/modules/gpu_transfer.py
royyhuang: Fixed

lmcache/v1/multiprocess/modules/non_gpu_transfer.py
royyhuang: Fixed.

lmcache/v1/multiprocess/config.py
royyhuang: Fixed

lmcache/v1/multiprocess/server.py
royyhuang: Fixed.

lmcache/v1/multiprocess/engine_module.py
royyhuang: done.

ApostaC (approved):
LGTM!

deng451e (approved):

PR #3673, opened 2026-06-14
rename GPUKVFormat -> EngineKVFormat, gpu_kv_* -> engine_kv_*

maobaolong:

GPUKVFormat does not always mean GPU (CPU/HPU/XPU...), rename to EngineKVFormat to describe the inference engine layout. Keep backward compat aliases (GPUKVFormat = EngineKVFormat) so existing callers like lmc_ops.GPUKVFormat.XXX still work.

**If applicable**:

### DIFF
--- a/csrc/mem_kernels.cu
+++ b/csrc/mem_kernels.cu
@@ -30,33 +30,34 @@ namespace lmc {
 
 // inline helper to check MLA (callable from device and host)
 __host__ __device__ __forceinline__ bool is_mla(
-    const GPUKVFormat gpu_kv_format) {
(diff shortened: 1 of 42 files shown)


maobaolong:
Thanks for the review! Pushed an update that drops the `GPUKVFormat` alias from all internal type hints, enum refs and docstrings. The backward-compat alias is now only declared in one place (next to the enum definition) so external callers still work.

sammshen (approved):
LGTM! Like this rename

hlin99 (approved):
LGTM
engine format makes more sense than gpu format

Issue #3674, opened 2026-06-14
[Refactor][Platform]: Move GPUKVFormat into a header file not coupled to accelerator


sammshen:
The `GPUKVFormat` (soon renamed to `EngineKVFormat`) and `TrasnferDirection` should live in theri own header files now that LMCache is supporting more platforms

MartinHua:
Hi I'm learning LMCache and want to onboard. Can I claim this as my first issue?

PR #3676, opened 2026-06-14
refactor: move KV-transfer enums into a backend-agnostic header

MartinHua:

`TransferDirection` and `GPUKVFormat` were duplicated byte-for-byte in
`csrc/mem_kernels.cuh` (CUDA) and `csrc/sycl/mem_kernels_sycl.h` (SYCL),
coupling these backend-agnostic descriptors to each accelerator's kernel
header. As LMCache adds more platforms, every new backend would have to
re-copy the enums. This extracts them into `csrc/kv_transfer_types.h`, a
dependency-free header (no torch/ATen/vendor includes) that both backends
include.

- Pure move — no behavioral change. Enum names, values, ordering, and the
  pybind-exposed Python API (`lmc_ops.GPUKVFormat`, `TransferDirection`)
  are unchanged, so the existing test suite is the regression guard.
- Scoped deliberately to the *move* only. The `GPUKVFormat` -> `EngineKVFormat`
  rename mentioned in the issue is intentionally left out to keep this focused
  (the rename touches ~417 Python references) and is tracked separately.
- No build-config change needed: quoted includes resolve relative to the
  including file (CUDA/ROCm/MUSA build from `csrc/`; SYCL uses `../`).
- One comment normalization: the last enum's note read "not the CUDA/SYCL
  transfer kernels" per file; merged it reads "not the device transfer kernels".
- Updated `docs/design/v1/gpu_connector/layout-invariant.md` "Adding a new
  format" step to point at the new shared header and both pybind modules.

Refs #3674

**If applicable**:


hlin99:
Thanks for the contribution. very quick action indeed.

if we can move the enum definitions in https://github.com/MartinHua/LMCache/blob/refactor/3674-kv-format-header/lmcache/python_ops_fallback.py to the single source, it would be perfect!

MartinHua:
> Thanks for the contribution. very quick action indeed.
> 
> if we can move the enum definitions in https://github.com/MartinHua/LMCache/blob/refactor/3674-kv-format-header/lmcache/python_ops_fallback.py to the single source, it would be perfect!

Thanks @hlin99! Quick clarification on what you have in mind, since "single source" can mean two different things here.

My understanding is at runtime there'll always be two enum objects — the C++/pybind one, and a Python one the fallback uses when the compiled extension isn't built. Please correct me if I'm wrong. And I can think of two ways:

1. Two definitions, kept honest by a test — with the existing test_c_ops_fallback_parity.py guarding C++↔Python value parity. Matches the repo's current convention. 
2. One authored source via codegen — a single spec generates the C++ header. But seems no codegen precedent in the repo yet.

Which are you thinking of?. 🙏

hlin99:
> > Thanks for the contribution. very quick action indeed.
> > if we can move the enum definitions in https://github.com/MartinHua/LMCache/blob/refactor/3674-kv-format-header/lmcache/python_ops_fallback.py to the single source, it would be perfect!
> 
> Thanks @hlin99! Quick clarification on what you have in mind, since "single source" can mean two different things here.
> 
> My understanding is at runtime there'll always be two enum objects — the C++/pybind one, and a Python one the fallback uses when the compiled extension isn't built. Please correct me if I'm wrong. And I can think of two ways:
> 
> 1. Two definitions, kept honest by a test — with the existing test_c_ops_fallback_parity.py guarding C++↔Python value parity. Matches the repo's current convention.
> 2. One authored source via codegen — a single spec generates the C++ header. But seems no codegen precedent in the repo yet.
> 
> Which are you thinking of?. 🙏

hi @MartinHua I don't have concrete solution right now, but opt#2 makes sense to me, for those enum definitions. and opt#1 could focus on function check. WDYT?

if opt#2 is not easy to handle in this PR, we can do in a following PR. thanks

MartinHua:
@hlin99 I agree one single source is the right call for the enums. The parity test only stays for function checks. But I would propose doing it in a tracked follow-up issue (with detailed plans for adding this capability) so #3676 stays focused and lands now.

Proposal — there's no codegen / build-time-generation in LMCache today, and the ROCm path hipifies csrc/, which makes "generate a header before compile" order-sensitive. So rather than a full codegen step, I'd propose a lighter route that still gives a true single source — an X-macro .def that lists the members once:

```
// csrc/kv_transfer_formats.def
X(NL_X_NB_BS_HS, 3)
...
```

The C++ enum and both pybind .value() blocks #include it (pure preprocessor), and the Python fallback parses the same file at import. That collapses all four current copies — the enum header, the CUDA + SYCL pybind blocks, and the Python IntEnum — into one, with no build-system or CI changes. Bonus: C++↔Python enum parity then holds by construction, so the parity test naturally narrows to function checks, exactly as you said.

If you'd rather standardize a real codegen step repo-wide instead (a script generating the header + pybind blocks + Python enum, with a freshness check), happy to go that way, but heavier since it has to integrate into different things.

I'll open a follow-up issue capturing this so we can keep this PR small. Sound good?

hlin99 (approved):
LGTM！
thanks for contributing LMCache. @MartinHua 

@maobaolong @sammshen would you like to have a second check? thx.

MartinHua:
Kindly ping @maobaolong @sammshen PTAL, thanks!

maobaolong (approved):
@MartinHua This is a good direction, thanks for this refactor work, it can reduce duplicated for existing code and future new adding code. LGTM.

I believe we can do more refactor target `csrc` similar with what we just done of `platform` python module.

PR #3678, opened 2026-06-14
[refactor] refactor CLI for multi sub commands 

chunxiaozheng:
# Summary
Refactored all CLI commands that contain subcommands to use a new `CompositeCommand` base class, replacing manual subcommand registration with automatic package-level discovery.

# Motivation
Previously, each parent command (e.g. `bench`, `query`, `quota`, `tool`, `trace`) manually imported and registered its subcommands. This led to:
1. Boilerplate registration code duplicated across commands
2. Tight coupling between parent and child commands
3. Forgetting to register new subcommands

# Changes
Added `CompositeCommand` (in `lmcache/cli/commands/base.py`): a BaseCommand subclass that auto-discovers nested `BaseCommand` implementations by scanning sibling modules in the same package via `discover_subclasses()`.
Converted all composite commands (`bench`, `query`, `quota`, `tool`, `trace`) to inherit from `CompositeCommand`, removing manual registration logic.
Each subcommand is now a standalone file/module under its parent package (e.g. `bench/engine_bench/`, `bench/server_bench/`, `quota/get_command.py`), automatically picked up at runtime.
No functional changes to CLI behavior or argument parsing — all existing commands and flags work identically.

# Key Design Decisions
1. `CompositeCommand` scans only direct children of its package (non-recursive), filtering out `_-prefixed` modules.
2. `require_defined_in_module=True` ensures only classes defined in the scanned module are registered (avoids re-importing base classes).
3. Supports N-level nesting: a subcommand can itself be a `CompositeCommand`(e.g. `tool` → `cache_simulator` → `simulate`/`sweep`/`gen_dataset`).


chunxiaozheng:
@ApostaC @KuntaiDu Could you help take a look?

KuntaiDu (approved):
LGTM as long as CI passes

maobaolong (approved):
LGTM. This is an awesome refactor. It really make it easy to extend cli commands, and This is very friendly for new contributor.

Issue #3539, opened 2026-06-04
[good-first-issue] docs: improve Chinese (zh_CN) documentation translations


glbyktjys:
**Summary**

We've built a pipeline that auto-translates LMCache docs into Chinese, and it re-runs every Monday to translate new content from the English .rst files. This gives us a strong starting point, but machine translation still needs human review for technical accuracy. For example, in an LLM-serving context, "recipe" shouldn't be the literal "食谱" (a cooking recipe); "操作手册" (practical guide) or "配置示例" (configuration example) reads far more naturally, depending on context.
If you read Chinese and are new to LMCache, this is a great first PR :)

**Details**

### **0. Before you start:**
**A weekly job (Monday 09:00 UTC) opens/updates a single PR on the branch `automation/update-chinese-docs`. The bot only translates entries that are empty or stale (marked #, fuzzy because the English .rst changed).** It will not touch an entry that already has a finished, non-fuzzy translation, so your work is safe — as long as you clear the #, fuzzy marker after editing (see step 1). Still, please:

- **Skim the open automation/update-chinese-docs PR/branch so you don't pick an entry it's mid-translating.**
- Branch off the latest dev.

### **1. What to edit:**
The Chinese translation files are located under: `docs/source/locale/zh_CN/LC_MESSAGES/**/*.po`
⚠️  Do not edit the English .rst files under `docs/source/`. The source documentation lives in the .rst files, while the Chinese translations live in the .po files.

Each .po entry is a pair:
```
#: ../../source/recipes/index.rst:4
#, fuzzy          ← delete this line (or just the word "fuzzy") after fixing msgstr
msgid "Recipes"        # English source — DO NOT CHANGE
msgstr "食谱"          # Chinese translation — fix THIS line
```

Edit only the msgstr value, leaving msgid untouched. After fixing an entry, remove any #, fuzzy marker on it. This matters twice: Sphinx ignores fuzzy entries at build time (so your fix won't appear until the flag is gone), and the weekly bot would otherwise re-translate it over your edit. 

One important exception: many .po files start with a #, fuzzy line near the top of the file header. That belongs to the file header. Please leave the header as-is and only remove #, fuzzy from the specific translation entries you update.

Good starting points (high visibility): `recipes/*.po`, `index.po`, `developer_guide/*.po`, `cli/*.po`, `mp/*.po`.

### **2. Verify locally**
```
pip install -r requirements/docs.txt
[...]

PR #3693, opened 2026-06-15
[good-first-issue] docs: fix Chinese colon in RST directive in mp/hybrid_models.po

April-Sonnet:
Refs #3539

Fixed a RST syntax error in the Chinese translation of `mp/hybrid_models.po`:
- `：doc:`quickstart`` → `:doc:`quickstart`` (Chinese colon replaced with English colon)

The `msgid` already uses the correct `:doc:` prefix; only the `msgstr` had the incorrect Chinese colon, which would break Sphinx rendering.

- [x] Only `mp/hybrid_models.po` is changed, 1 line
- [x] Changes are purely RST syntax fix

Refs #3692 (same type of fix)


sammshen (approved):

ApostaC (approved):

April-Sonnet:
This fix is no longer needed — the .po file was regenerated upstream (POT-Creation-Date: 2026-06-22) and the full-width colon issue is already resolved in the current `dev` translation. The regenerated file has a completely restructured translation that supersedes the original base.

PR #3697, opened 2026-06-15
[feat][cb] fault-inject L2 adapter + segmented-prefix retrieval path

deng451e:

Real L2 caches always return a clean contiguous-prefix-then-tail result, so the *segmented* found-set paths (segmented prefetch → scatter → attention) are never exercised by normal workloads. This PR adds a way to trigger them deterministically and wires the segmented-prefix retrieval policy end-to-end.

1. **Fault-inject L2 adapter** (`type: "fault_inject"`) — a decorator over a real inner adapter (e.g. `fs_native`) that drops a deterministic key subset *at load* to simulate partial L2 retrieve failures. Lookup passes through; the gap appears at load failure (the faithful "L2 retrieve error"), and the prefetch controller releases load-failed locks via the trim mask. Drop-set knobs: `rate`, `seed`, `gap_indices`, and `drop_chunk_hashes` (content-keyed, so the same chunk fails across prefix and sparse legs). Auto-discovered via the L2 adapter registry.

2. **Segmented-prefix retrieval** — thread `TrimPolicy` through the prefix lookup path so a mid-prefix gap no longer truncates everything after it. The dense prefetch branch forwards the policy, `LookupModule.lookup` takes a policy arg, and the CB prefix leg is gated on `CB_SEGMENTED_PREFIX=1` to request `SEGMENTED_PREFIX`. Gapped loads now leave post-gap chunks L1-resident (picked up by the sparse leg as L1 hits, the hole recomputed) instead of discarding everything after the gap.

**Special notes for reviewers**:

- The fault-inject adapter is **test/diagnostic only** (logs a warning when active) and gated behind config; no production path changes by default.
- Segmented-prefix delivery is behind the `CB_SEGMENTED_PREFIX` flag, so default prefix-leg behavior is unchanged.

**If applicable**:

- [x] User-facing changes — docs added (`docs/design/.../fault_inject.md`)
- [x] Unit tests (`test_fault_inject_l2_adapter.py`; segmented-prefix coverage in `test_prefetch_controller.py`)

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/fault_inject_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/fault_inject_l2_adapter.py
@@ -115,18 +115,15 @@ class FaultInjectL2AdapterConfig(L2AdapterConfigBase):
         if not isinstance(inner_type, str):
             raise ValueError("'inner' adapter spec must include a string 'type'")
 
-        # Build the inner adapter config via the registry (lazy-importing its
-        # module if needed), mirroring parse_args_to_l2_adapters_config.
+        # Build the inner adapter config via the public registry accessor
+        # (lazy-importing its module if needed), mirroring
+        # parse_args_to_l2_adapters_config.
         # First Party
         from lmcache.v1.distributed.l2_adapters.config import (  # noqa: PLC0415
-            _L2_ADAPTER_CONFIG_REGISTRY,
-            _ensure_config_loaded,
+            get_l2_adapter_config_class,
         )
 
-        _ensure_config_loaded(inner_type)
-        if inner_type not in _L2_ADAPTER_CONFIG_REGISTRY:
-            raise ValueError(f"unknown inner adapter type {inner_type!r}")
-        inner_cls = _L2_ADAPTER_CONFIG_REGISTRY[inner_type]
+        inner_cls = get_l2_adapter_config_class(inner_type)
         inner_config = inner_cls.from_dict(inner)
         inner_config.eviction_config = cls._parse_eviction_config(inner)
         inner_config.persist_config = cls._parse_persist_config(inner)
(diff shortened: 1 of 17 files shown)


lmcache/v1/multiprocess/modules/blend_v3.py
```diff
@@ -794,25 +808,43 @@ def cb_unified_lookup(
...
         # --- Prefix leg: poll (consume-once) until the L1+L2 prefix lands. ---
         if job.prefix_chunks is None:
-            p = self._lookup_module.query_prefetch_status(rid)
-            if p is None:
-                return None  # prefix still loading -> defer
-            job.prefix_chunks = p
+            if segmented:
+                # Segmented: also capture the gapped retained set so the
+                # post-gap chunks are delivered through the prefix leg (below).
+                seg = self._lookup_module.query_prefetch_status_segmented(rid)
```
ApostaC: Note: can we directly use storage manager instead of lookup module?

lmcache/v1/distributed/l2_adapters/fault_inject_l2_adapter.py
ApostaC, line 1: user-facing interface for this new L2 adapter

lmcache/v1/multiprocess/config.py
```diff
@@ -306,6 +312,13 @@ def add_mp_server_args(
         "pinged (model warmup or early death). Must be >= the worker reap "
         "timeout. Default is 3600.",
     )
+    mp_group.add_argument(
+        "--enable-segmented-prefix",
+        action="store_true",
+        help="CacheBlend (--engine-type blend) only: on a mid-prefix L2 "
+        "retrieve failure, retain the gapped prefix so post-gap chunks stay "
+        "L1-resident instead of truncating at the gap. No effect otherwise.",
+    )
```
ApostaC, line 321: user facing doc for this

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM

PR #3698, opened 2026-06-15
[Fix][RawBlock] Roll back put-task refs when dispatch scheduling fails

nayeonikim:

`batched_submit_put_task` raises a ref and reserves a `_put_tasks` entry for each
pending object before scheduling it on the event loop. If
`asyncio.run_coroutine_threadsafe` fails (e.g. the loop is shutting down), the
unscheduled items leaked their ref and key — the MemoryObj stayed pinned and the
key was permanently blocked from future puts. This wraps dispatch so unscheduled
items roll back their ref and `_put_tasks` entry on failure.

Already-scheduled items still release their own ref / `_put_tasks` entry in the
coroutine's `finally`; only the unscheduled tail (`pending[scheduled_count:]`) is
rolled back here. Covers both the io_uring batched and per-key paths.


DongDongJu (approved):
LGTM

DongDongJu:
@sammshen @maobaolong @ankit-sam Could you take a look once?

ankit-sam (approved):
LGTM

sammshen (approved):
LGTM!

nayeonikim:
@DongDongJu Could you merge this when you get a chance? CI ran on an earlier commit — the latest commit only has a minor fix but the checks didn't re-trigger. All reviews are approved. Thanks!

PR #3704, opened 2026-06-15
Add blocking WAIT_PREFETCH_STATUS to avoid busy-polling the lookup wait

### LINKED ISSUE #3694: SGLang + LMCache MP: replace polling for prefetch completion
### Background
In LMCache MP mode, the inference engine talks to the LMCache server through ZMQ. On the client side, the MP connector submits cache operations such as `LOOKUP` and `RETRIEVE`. On the server side, the LMCache MP server handles these requests and manages KV cache movement across storage layers.

For a load path, the client first submits a `LOOKUP` request. The server tracks the lookup/prefetch state by request_id. The client then checks the prefetch result through `QUERY_PREFETCH_STATUS`. Once the matched chunks are ready in L1, the client can submit `RETRIEVE` to copy the KV cache chunks back to GPU.

### Problem

`_wait_for_lookup` waits for the lookup/prefetch result by repeatedly sending `QUERY_PREFETCH_STATUS` in a loop with time.sleep(0.001)

This works functionally, but it has two drawbacks:
1. It creates repeated ZMQ requests while the prefetch is still in progress.
2. It adds polling delay before the client observes that the prefetch has completed.

This is not noticeable when the data is already available in L1, because `QUERY_PREFETCH_STATUS` can return quickly. However, when the lookup triggers movement from a slower storage layer into L1, this wait is on the request path and can affect request latency.

### Goal

Improve the lookup wait path so the client does not need to busy-poll `QUERY_PREFETCH_STATUS`. The desired behavior is to avoid client-side busy polling after `LOOKUP` is submitted. The client should be able to efficiently determine when the corresponding prefetch operation has completed and then continue with `RETRIEVE`, whether through a wait mechanism, a notification-based approach, or another suitable design.

### Highlevel Task (not limited to the following)
1. Review the current `LOOKUP`, `QUERY_PREFETCH_STATUS`, and `RETRIEVE` flow in MP mode.
2. Design and implement an asynchronous loading mechanism that removes the need for client-side polling.
3. Add tests or benchmarks to compare the current polling behavior with the updated wait behavior, especially when L2 is enabled.

HaozheZhang6:

The SGLang MP adapter's `_wait_for_lookup` busy-polls `QUERY_PREFETCH_STATUS` in a 1 ms sleep loop (it carries a `TODO(Shaoting)` noting a blocking variant is the real fix). That adds repeated ZMQ requests and polling latency on the load path -- noticeable when a lookup triggers movement from a slower storage layer into L1.

This adds a blocking `WAIT_PREFETCH_STATUS` request type, following that TODO:

- `PrefetchController` gets a condition variable over its existing `_prefetch_results_lock`. `_complete_request` notifies it when a result is published, and a new `wait_prefetch_result(request_id, timeout)` blocks until the result is available (or times out).
- `StorageManager.wait_prefetch_status(handle, timeout)` delegates to it; L1-only prefetches (`prefetch_request_id == -1`) return immediately.
- The `LookupModule.wait_prefetch_status` handler blocks, then reuses the existing `query_prefetch_status` path to read the chunk count.
- The client `_wait_for_lookup` now issues a single `WAIT_PREFETCH_STATUS` request instead of the poll loop.

Fixes #3694

- The new handler is on the `NORMAL` thread pool. There isn't a blocking/IO pool today, so a concurrent burst of waits could occupy `NORMAL` workers while prefetches are in flight -- happy to add a dedicated pool if you'd prefer that.
- The condition variable shares the existing `_prefetch_results_lock`, so it can't race with `query_prefetch_result` (which pops under the same lock). `wait_prefetch_result` only peeks; the caller's existing query still pops with exactly-once semantics.
- I can't run the daemon end-to-end locally (needs the built native ops + GPU MP infra), so that path is left to CI. I verified the condition-variable mechanism on CPU (blocks-until-complete / already-complete-returns-immediately / timeout / concurrent waiters) and added `TestWaitPrefetchResult` to `tests/v1/distributed/test_prefetch_controller.py`.

**If applicable**:


DongDongJu (changes requested):
Hello @HaozheZhang6,

I am OK with this after CI, but please add one MP-level test or a targeted `LookupModule` test that exercises the full `WAIT_PREFETCH_STATUS -> query_prefetch_status` path. 
The controller tests cover the condition variable, but the request handler and protocol path are where regressions are most likely.

otherwise, LGTM

HaozheZhang6:
Done — added `tests/v1/multiprocess/test_lookup_wait_prefetch.py` exercising the `wait_prefetch_status -> query_prefetch_status` handler path: success → count + END event + exactly-once job consumption, timeout → None with the job kept, and unknown request → 0. Storage manager is mocked so it runs without a GPU/native bitmap.

glbyktjys:
Thank you @DongDongJu for reviewing this.
Hi @HaozheZhang6, thank you very much for your efforts on improving LMCache MP support with SGLang. We really appreciate your interest in contributing here.

Since these changes touch the MP integration path, we do need end-to-end local testing to make sure the implementation works correctly and does not introduce regressions. For LMCache, validating that the code is correct, reproducible, and maintainable for the community is an important part of the contribution process.

**If you do not currently have access to the needed GPU environment, could you please ask in issue #3694  and collaborate with someone who has the setup to verify the change together?** Once the PR includes clear test results, such as the measured prefetch improvement from L2, along with reproduction steps and confirmation that the end-to-end flow works, it will be much easier for maintainers to review and move forward.

Thank you again for working on this, and we’d be happy to continue guiding the contribution once the testing part is addressed.

HaozheZhang6:
Marked this draft while I get the e2e numbers — bringing up the SGLang + LMCache MP harness (the `.buildkite/k3_tests/sglang` perf + correctness scripts) on a GPU box to measure the L2→L1 prefetch wait (blocking vs the current poll) and confirm the flow end-to-end. Will post results + repro steps and re-mark ready.

HaozheZhang6:
Got the SGLang + LMCache MP path running on an A100 and measured the L2→L1 prefetch wait end to end. (`REGISTER_KV_CACHE` shares the KV cache over CUDA IPC, so it needs a real Linux GPU box.)

**Setup.** `lmcache server` with a `mock` L2 adapter — `mock_bandwidth_gb` sets the prefetch latency — and `--l2-store-policy skip_l1`, so a cache hit always takes the L2→L1 prefetch path instead of being served from L1. SGLang `--enable-lmcache` serving `Qwen2.5-0.5B-Instruct`. Each trial warms a ~4k-token prefix, `/flush_cache`es the radix tree so the next hit misses on GPU and goes to LMCache, then times to first token. CPU is read from `/proc`, split between the `lmcache server` daemon and the SGLang worker. 8 trials/point, medians.

```bash
lmcache server --host 127.0.0.1 --port 5556 --l1-size-gb 2 --eviction-policy LRU \
  --l2-store-policy skip_l1 --l2-adapter '{"type":"mock","max_size_gb":10,"mock_bandwidth_gb":0.15}'

python -m sglang.launch_server --model-path Qwen/Qwen2.5-0.5B-Instruct --port 30000 \
  --page-size 32 --enable-lmcache --lmcache-config-file lmcache_config.yaml
```

**Results** — each cell is `busy-poll → WAIT (this PR)`:

| mock L2 bw | prefetch | TTFT | SGLang-worker CPU/req | daemon CPU/req |
|---|---|---|---|---|
| 0.5 GB/s  | 0.32 s | 317 → 318 ms   | 95 → 65 ms  | 340 → 350 ms |
| 0.15 GB/s | 0.94 s | 936 → 933 ms   | 185 → 65 ms | 610 → 320 ms |
| 0.05 GB/s | 2.69 s | 2694 → 2694 ms | 415 → 95 ms | 740 → 385 ms |

The end-to-end flow works and the output matches the L1-hit path. TTFT is unchanged — the prefetch dominates it. The win is CPU: the old loop sleeps 1 ms and re-sends `QUERY_PREFETCH_STATUS` each iteration, i.e. ≈ one round-trip per ms of prefetch (hundreds to thousands of them), so its CPU grows with the prefetch time on both the worker and the daemon. `WAIT_PREFETCH_STATUS` sends one request and blocks on the condition variable, so the polling CPU stays flat as the L2 gets slower — which is exactly when the prefetch wait is long enough to matter. (Table is medians; the busy-poll also had occasional CPU spikes the blocking wait didn't.)

(The earlier Code Quality red was just a missing blank line in the new test — fixed in the last commit.)

Marking ready for review. Happy to share the harness scripts.

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -64,6 +64,7 @@ class RegisterNonGpuContextResponse:
     "RETRIEVE",
     "LOOKUP",
     "QUERY_PREFETCH_STATUS",
+    "WAIT_PREFETCH_STATUS",
```
Shaoting-Feng, line 67: Can you update the doc, e.g., https://github.com/LMCache/LMCache/blob/3fcbc3b53f4979b34b1f3abd2bb5b24182266464/docs/source/mp/index.rst#zmq-protocol and mention this is for SGLang only?

Shaoting-Feng (approved):
Just one comment on doc, otherwise LGTM

lmcache/v1/multiprocess/protocols/engine.py
HaozheZhang6: Done, added in 71f19578.

DongDongJu (approved):
LGTM

PR #3708, opened 2026-06-15
[Feat] [Coordinator] Optimize coordinator global CacheBlend token matching

ruizhang0101:

The coordinator's global CacheBlend directory matched request tokens with a per-position Python `dict.get` loop holding a single global lock — O(n) per query, serializing every publish/query. This optimizes the match path:

- **Vectorized probe** — replace the Python loop with a per-scope direct-address table indexed by one numpy gather; only occupied slots reach a short full-hash verify loop. ~2–2.5x faster `match()`, flat across directory size.
- **Per-scope tables + locks** — the single global lock is gone; queries/publishes to different models no longer contend.
- **Concurrent match dispatch** — the client daemon dispatches match queries to a thread pool instead of servicing them serially, so one slow coordinator reply no longer stalls the queue (`LMCACHE_COORDINATOR_BLEND_MATCH_CONCURRENCY`, default 8).
- **Compact wire form** — match tokens ship as a base64 `uint32` buffer (`tokens_b64`) instead of a JSON int list: ~1.4x smaller, ~6x faster to decode, fed straight to the matcher as an array.
- **Scope = model only** — dropped `cache_salt` from the scope key. Salt isolation still holds because matched content hashes are expanded to `ObjectKey`s with the *requester's* salt at retrieve, so a cross-salt match misses at L2. Collapses one table per `(model, salt)` to one per model.

Combined coordinator-side per-query work (decode + match) is ~4.5x faster.

- **Never wrong KV**: a rare bucket collision (low-bit truncation, last-writer-wins) only drops the overwritten chunk from global matching — a missed reuse recomputed downstream. Same tolerance as the existing stale-entry / dropped-publish paths.
- The directory now reveals cross-salt content *existence* to trusted mp-servers (not content or KV). Fine for trusted fleet infra; revisit if salts are a hard security boundary.
- Wire change: `BlendMatchRequest.tokens` → `tokens_b64` (coordinator + client updated together; register still ships raw tokens as JSON).

**If applicable**:


deng451e (approved):
Good work!

ApostaC (approved):
LGTM!

PR #3709, opened 2026-06-15
[CI] fix flaky failures in AMD CI

ApostaC:

**If applicable**:


hlin99 (approved):
LGTM

maobaolong (approved):
@ApostaC Thanks for the fix, It looks like this issue was caused by version discrepancies in some third-party package dependencies between the AMD and NVIDIA environments.

From what I can see, your fix definitely works. However, it would be even better if you could clarify the underlying reason for the root cause in the description.

PR #3714, opened 2026-06-16
[CI][MP] Register HTTP API routers explicitly to fix AMD unit tests

sammshen:

The `unit-tests-amd` Buildkite pipeline has been red on **every** build — including docs-only and dependabot PRs — so the failure is independent of PR contents.

Root cause: the multiprocess HTTP server builds its routes by *filesystem reflection* (`pkgutil.iter_modules` over `Path(__file__).parent`). Under this pipeline's editable install (`uv pip install -e .`), the nested discovery of `internal_api_server/common` returns no modules, so `common_api.router` ends up empty and `/run_script` (plus `/env`, `/conf`, `/loglevel`, `/metrics`) never register. `test_run_script_endpoint_present` then fails on every build, and `--maxfail=1` aborts the whole suite — turning one environment-sensitive test into an always-red pipeline regardless of the PR.

This PR registers the routers via **explicit imports** instead of filesystem reflection, so registration no longer depends on the on-disk package layout (which differs between source checkouts and editable/wheel installs). It also drops `--maxfail=1` from the bare-metal pipeline so a single failing test can't red the entire matrix.

- Behavior is unchanged on a healthy install — the MP server still exposes the same 27 routes (verified locally: all expected endpoints present). Covered by the existing `test_common_api.py` and `test_http_api_registry.py` integration tests, which were the ones going red on AMD.
- `lmcache/v1/utils/router_discovery.py` is intentionally left in place; it is still used by the `mp_coordinator` and `internal_api_server` registries, which share the same latent fragility. Converting those to explicit registration is a sensible follow-up, kept out of scope here.
- Trade-off of the explicit list: adding a new `http_apis`/`common` module now requires one import line (documented in each module's docstring) instead of being auto-discovered.
- I could not reproduce the editable-install failure locally (the `uv` editable build needs the ROCm C-extension toolchain), but the explicit imports remove the failing mechanism by construction.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/http_api_registry.py
+++ b/lmcache/v1/multiprocess/http_api_registry.py
@@ -1,25 +1,45 @@
 # SPDX-License-Identifier: Apache-2.0
-# Standard
-from pathlib import Path
-
 # Third Party
 from fastapi import APIRouter, FastAPI
 
 # First Party
-from lmcache.logging import init_logger
-from lmcache.v1.utils.router_discovery import discover_api_routers
+from lmcache.v1.multiprocess.http_apis.cache_api import router as cache_router
+from lmcache.v1.multiprocess.http_apis.common_api import router as common_router
+from lmcache.v1.multiprocess.http_apis.conf_api import router as conf_router
+from lmcache.v1.multiprocess.http_apis.healthcheck_api import (
+    router as healthcheck_router,
+)
+from lmcache.v1.multiprocess.http_apis.quota_api import router as quota_router
+from lmcache.v1.multiprocess.http_apis.reconfigure_api import (
+    router as reconfigure_router,
+)
+from lmcache.v1.multiprocess.http_apis.root_api import router as root_router
+from lmcache.v1.multiprocess.http_apis.status_api import router as status_router
+from lmcache.v1.multiprocess.http_apis.version_api import router as version_router
 
-logger = init_logger(__name__)
+# Every multiprocess HTTP API router, registered in import order. Add a new
+# ``http_apis`` module's router here to expose it. Routers are imported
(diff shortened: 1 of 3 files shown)


lmcache/v1/multiprocess/http_api_registry.py
```diff
@@ -1,44 +1,54 @@
...
     def __init__(self, app: FastAPI):
         self.app = app
         self.router = APIRouter()
 
     def register_all_apis(self) -> None:
-        """
-        Discover and register all ``*_api`` modules under
-        the ``http_apis`` directory.
-        """
-        apis_path = Path(__file__).parent / "http_apis"
```
maobaolong, line 34: Regarding the current issue where the AMD unit tests are failing, we should prioritize finding the root cause. 

If we truly cannot identify the root cause, as a fallback, should we consider keeping the previous dynamic discovery method for endpoints? Meanwhile, simply adds the necessary endpoints explicitly to the route list to pass the UT. @sammshen

lmcache/v1/multiprocess/http_apis/common_api.py
```diff
@@ -1,46 +1,48 @@
...
 router = APIRouter()
-
-# Modules that depend on vLLM-specific ``app.state`` attributes and
-# therefore cannot run on the multiprocess HTTP server.
-_MP_INCOMPATIBLE_MODULES: frozenset[str] = frozenset()
-
-_common_path = Path(internal_api_server.__file__).parent / "common"
-_common_package = f"{internal_api_server.__name__}.common"
-
-for _discovered in discover_api_routers(
```
maobaolong, line 41: ditto

PR #3718, opened 2026-06-16
fix(mp): Eliminate Container PID Namespace Collisions in LMCache MP Server Worker Registrations

weizhoublue:

a follow-up for https://github.com/LMCache/LMCache/pull/3631

When worker containers connect to same shared LMCache MP  server, they query  os.getpid()  and send  1  as  instance_id.
In containerized  setups, PID namespaces isolate processes, causing PID collisions across replicas

Note: The vLLM adapter path is already covered by #3631. This PR extends the unique UUID-derived instance ID behavior to SGLang and TensorRT-LLM adapters

**If applicable**:

### DIFF
--- a/lmcache/integration/sglang/multi_process_adapter.py
+++ b/lmcache/integration/sglang/multi_process_adapter.py
@@ -2,9 +2,9 @@
 # Standard
 from dataclasses import dataclass
 from typing import Optional
-import os
 import threading
 import time
+import uuid
 
 # Third Party
 from sglang.srt.configs.model_config import ModelConfig
@@ -113,7 +113,15 @@ class LMCacheMPConnector:
         self.model_name = sgl_config.model_path
         self.num_layers = len(k_pool)
         self.tp_group = tp_group
-        self.instance_id = os.getpid()
+
+        # Instance id for GPU worker. uuid4-derived (OS entropy) rather
+        # than os.getpid() to avoid collision in containerized deployments.
+        # Masked to 63 bits to stay signed-int64-safe for any msgpack peer.
+        self.instance_id = uuid.uuid4().int & ((1 << 63) - 1)
+        logger.info(
+            "LMCache MP worker adapter created with instance_id=%d",
+            self.instance_id,
+        )
         self._mq_timeout = mq_timeout
         self._heartbeat_interval = heartbeat_interval
         self._registered = False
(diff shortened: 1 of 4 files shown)


weizhoublue:
@royyhuang  hi , I think the solution of pid issue may also apply to sglang.  please help review , thanks

weizhoublue:
the failing CI  failed at test_common_api.py , which is not related to this PR
https://buildkite.com/lmcache/unit-tests-amd/builds/7348/list?sid=019ecec9-61ad-4fe8-96d2-8be8acfa7a80&tab=output

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -113,7 +113,15 @@ def __init__(
         self.model_name = sgl_config.model_path
         self.num_layers = len(k_pool)
         self.tp_group = tp_group
-        self.instance_id = os.getpid()
+
+        # Instance id for GPU worker. uuid4-derived (OS entropy) rather
+        # than os.getpid() to avoid collision in containerized deployments.
+        # Masked to 63 bits to stay signed-int64-safe for any msgpack peer.
+        self.instance_id = uuid.uuid4().int & ((1 << 63) - 1)
```
DongDongJu, line 120: IMO, This is the right direction. 
Please mention in the PR body whether the vLLM adapter path is already covered by #3631, 
so this PR’s scope is clear.

lmcache/integration/tensorrt_llm/tensorrt_mp_adapter.py
```diff
@@ -288,7 +289,14 @@ def __init__(self, llm_args: TorchLlmArgs) -> None:
             os.environ.get("LMCACHE_MQ_TIMEOUT", DEFAULT_MQ_TIMEOUT)
         )
 
-        self._instance_id = os.getpid()
+        # Instance id for GPU worker. uuid4-derived (OS entropy) rather
+        # than os.getpid() to avoid collision in containerized deployments.
+        # Masked to 63 bits to stay signed-int64-safe for any msgpack peer.
+        self._instance_id = uuid.uuid4().int & ((1 << 63) - 1)
```
DongDongJu, line 295: Same comment as SGLang: the generated ID is safe for the wire type. 
Please keep the ID stable for the worker lifetime only; no persistence is needed.

DongDongJu (approved):
Hello @weizhoublue,
This is a good fix for containerized MP workers. 
Please just clarify in the PR body whether vLLM was already handled in #3631. 
Otherwise LGTM.

lmcache/integration/sglang/multi_process_adapter.py
weizhoublue: updated

PR #3721, opened 2026-06-16
refactor(mp): Rename EngineDrivenContext to TransferBackend

Lyj1007:

This PR renames the engine-driven transfer backend class hierarchy for clarity and fixes a redundant naming issue in the log path.

**Changes**:

1. **Rename class hierarchy** — to better reflect their role as transfer backends rather than "contexts":

       Before:
         └─ EngineDrivenTransferContext    (data path via data copying in adapter)
             └─ EngineDrivenContext (transfer_context/base.py)
                ├─ EngineDrivenContextPickle (transfer_context/pickle.py)
                └─ EngineDrivenContextShm    (transfer_context/shm.py)

       After:
         └─ EngineDrivenTransferContext    (data path via data copying in adapter)
             └─ TransferBackend (transfer_context/base.py)
                ├─ PickleTransferBackend (transfer_context/pickle.py)
                └─ ShmTransferBackend    (transfer_context/shm.py)

2. **Remove `supported_transfer_mode` redundant variable** — the old code computed `supported_transfer_mode = "SHM" if shm_name and pool_size > 0 else "pickle"` solely for a log message, which duplicated the selection logic already inside `create_transfer_backend()`. This is now replaced with `type(self._backend).__name__` in the log line, which directly reflects the actual backend type in use, eliminating both the duplicated logic and the confusing `supported_transfer_mode` naming.

Pure refactor — no behavior change. The log message now accurately reports the concrete backend class name (e.g. `ShmTransferBackend` / `PickleTransferBackend`) instead of a separately computed string that could diverge from reality.

### DIFF
--- a/lmcache/v1/multiprocess/transfer_context/worker_transfer.py
+++ b/lmcache/v1/multiprocess/transfer_context/worker_transfer.py
@@ -23,10 +23,10 @@ from lmcache.v1.multiprocess.mq import MessageQueueClient
 from lmcache.v1.multiprocess.protocol import RequestType
 from lmcache.v1.multiprocess.protocols.engine import RegisterEngineDrivenContextResponse
 from lmcache.v1.multiprocess.transfer_context.base import (
-    EngineDrivenContext,
     EngineDrivenContextMetadata,
+    TransferBackend,
     compute_kv_layout,
-    create_engine_driven_context,
+    create_transfer_backend,
     gather_paged_kv_to_cpu,
     scatter_cpu_to_paged_kv,
 )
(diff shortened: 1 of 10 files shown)


lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -403,11 +403,10 @@ def register(
             shm_name=shm_name,
             pool_size=pool_size,
         )
-        supported_transfer_mode = "SHM" if shm_name and pool_size > 0 else "pickle"
```
hlin99, line 419: hi, is there new logic to replace this line?

Lyj1007: Hi! The original `supported_transfer_mode` line was only used to produce a log message — it didn't drive any actual branching logic. In the new code:

1. The **selection logic** (SHM vs pickle) already lives inside `create_transfer_backend()`, which inspects `shm_name` / `pool_size` and returns the appropriate backend instance. This function is unchanged — no logic was moved into or out of it.

2. The old two lines — computing `supported_transfer_mode` and then logging it — are consolidated into a single `logger.info` call:

   ```python
   logger.info(
       "Worker non-GPU transfer context registered (instance_id=%d, backend=%s)",
       instance_id,
       type(self._backend).__name__,
   )
   ```

   This directly reads the concrete backend type via `type(self._backend).__name__`, so it reflects which backend was actually created without needing a separate `supported_transfer_mode` variable or duplicating any selection logic.

So there's no missing logic — the old `supported_transfer_mode` variable was redundant with what `create_transfer_backend()` already decides, and removing it eliminates the naming confusion while keeping the log output equally informative.

Issue #3553, opened 2026-06-05
Tracking: MUSA backend support roadmap


superleo:
## Goal

Track the staged plan for MUSA backend support in LMCache, including current implementation status, reviewer-facing scope, and follow-up work.

## Current Status

MUSA support is implemented through Stage 3, with Stage 4 completed the full IPC handle path. SGlang MUSA In progress.

### Stage 1: Basic vLLM MUSA Connector

Status: Completed 
- https://github.com/LMCache/LMCache/pull/3483

Scope:

- Support `NL x [2, NB, BS, NH, HS]`
- Support MLA `NL x [NB, BS, HS]`
- Handle `vllm_cached_tokens`
- Add shape documentation and guards
- Reject unsupported layouts explicitly

### Stage 2: vLLM In-process Performance MVP

Status: Completed 
- https://github.com/LMCache/LMCache/pull/3580

Scope:

- Add optional native MUSA acceleration for vLLM gather/scatter.
- Directly support LMCache/vLLM MUSA layouts without extra transpose copies.
- Keep torch `index_select` / `index_copy_` path as correctness fallback.
- Add benchmark coverage against the current torch path.
- Keep SGLang MUSA out of scope and explicitly unsupported.

### Stage 3: MP Non-handle Data-path MVP

Status: Completed 
- https://github.com/LMCache/LMCache/pull/3725 

Scope:

- Enable MUSA MP non-handle data path.
- Reuse Stage 2 worker-side gather/scatter, using native acceleration when available.
- Keep torch fallback for correctness.
- Avoid MUSA IPC handle passing and server-side MUSA tensor reconstruction in MVP.
- Validate MP correctness and basic performance

### Fix:  
- https://github.com/LMCache/LMCache/pull/3875

### Stage 4: Full MUSA IPC / Advanced Transfer Path

Status: WIP

Scope:

- MUSA IPC handle path.
- Server-side MUSA tensor reconstruction.
- MUSA IPC event integration.
- Full `multi_layer_block_kv_transfer` equivalent for MUSA.
- Hybrid KV group data path if required by production MP performance.

[Update 20260611]To keep Stage 4 reviewable, propose splitting it into three smaller pieces and aligning the exact boundaries with maintainers before implementation.

Stage 4.1: Fail-Closed Capability Gate
status:  Completed 
- https://github.com/LMCache/LMCache/pull/3756

- Add only the capability gate, platform registration, and `MusaIPCWrapper` boundary.
[...]

PR #3725, opened 2026-06-16
feat: enable MUSA MP engine-driven non-handle path

superleo:

This is Stage 3 of MUSA support. It enables the MP engine-driven, non-handle data path for MUSA workers so MP store/retrieve can move KV data through worker-side gather/scatter into CPU/SHM chunks, instead of passing MUSA IPC handles or reconstructing MUSA tensors on the server.

Main changes:
- Route MUSA MP workers through `EngineDrivenTransferContext(device_type="musa")`.
- Validate Stage 3 MUSA MP layouts and fail closed for unsupported layouts.
- Reuse the Stage 2 MUSA gather/scatter native adapter when `LMCACHE_MUSA_NATIVE_KV_TRANSFER=1`.
- Preserve the torch fallback when native transfer is disabled, unavailable, ABI-incompatible, or declines a transfer.
- Keep native MUSA transfer implementation under `lmcache.v1.platform.musa.native_kv_transfer`.
- Add regression coverage for native MUSA scatter block-aligned `skip_first_n_tokens` rounding.

Stage 3 intentionally does not add MUSA IPC handle passing or server-side MUSA tensor reconstruction. The MVP keeps device movement on the worker side and sends CPU/SHM chunks through the existing engine-driven MP path.

Goal check:
- [x] Enable the MUSA MP non-handle data path.
- [x] Reuse Stage 2 worker-side gather/scatter with optional native acceleration.
- [x] Keep torch fallback for correctness.
- [x] Avoid MUSA IPC handle passing and server-side MUSA tensor reconstruction.
- [x] Validate MP correctness and basic performance on MUSA.

Local validation:
```bash
.venv/bin/python -m pytest \
  tests/v1/multiprocess/test_engine_driven_transfer.py \
  tests/v1/test_musa_native.py \
  tests/v1/test_musa_connector.py \
  tests/v1/test_musa_support.py \
  tests/benchmarks/test_musa_inprocess_benchmark.py -q

Result:
======================= 88 passed, 409 warnings in 8.71s =======================
```

```bash
tests/v1/test_python_ops_fallback.py
tests/v1/test_c_ops_fallback_parity.py

Key items:
tests/v1/test_python_ops_fallback.py::TestScenarios::test_2_compare[multi_layer_kv_transfer] PASSED
tests/v1/test_python_ops_fallback.py::TestScenarios::test_2_compare[multi_layer_block_kv_transfer] PASSED
tests/v1/test_c_ops_fallback_parity.py::test_function_signature_parity[multi_layer_block_kv_transfer] PASSED
tests/v1/test_c_ops_fallback_parity.py::test_all_c_ops_callables_have_fallback PASSED
tests/v1/test_c_ops_fallback_parity.py::test_all_c_ops_enums_have_fallback PASSED
tests/v1/test_c_ops_fallback_parity.py::test_all_c_ops_descriptors_have_fallback PASSED

================== 53 passed, 2 skipped, 54 warnings in 2.47s ==================

Skipped items:
tests/v1/test_c_ops_fallback_parity.py::test_enum_parity[__placeholder__] SKIPPED
tests/v1/test_c_ops_fallback_parity.py::test_descriptor_class_parity[__placeholder__] SKIPPED
```

E2E Test With VLLM MUSA Pass.

If applicable:
[ ] this PR contains user facing changes - docs added
[x]this PR contains unit tests


lmcache/v1/multiprocess/transfer_context/base.py
```diff
@@ -712,3 +787,235 @@ def scatter_cpu_to_paged_kv(
...
+    block_size: int,
+    device: torch.device,
+) -> torch.Tensor:
+    """Expand paged block IDs into token slot IDs."""
+    slots = [
+        block_id * block_size + token_offset
+        for block_id in block_ids
+        for token_offset in range(block_size)
+    ]
+    return torch.tensor(slots, dtype=torch.long, device=device)
```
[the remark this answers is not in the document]
superleo: fixed

hlin99 (changes requested):
hi @superleo the design philosophy of the transfer_context is device agnostic.
lmc_ops.multi_layer_block_kv_transfer is the entry to the device layer. for musa, I think all the change under transfer context can be wrapped in this multi_layer_block_kv_transfer function. just hide all the details(any check, tensor reshape, fallback logic, etc) under your platform specific code, so when other platforms are added in, they don't need to take care of other devices' logic.

superleo:
> hi @superleo the design philosophy of the transfer_context is device agnostic. lmc_ops.multi_layer_block_kv_transfer is the entry to the device layer. for musa, I think all the change under transfer context can be wrapped in this multi_layer_block_kv_transfer function. just hide all the details(any check, tensor reshape, fallback logic, etc) under your platform specific code, so when other platforms are added in, they don't need to take care of other devices' logic.

Thank you Tony, the boundary concern makes sense. The current PR puts the optional MUSA native fast path in transfer_context to reuse Stage 2 gather/scatter quickly, but that does leak MUSA-specific checks and fallback behavior into a device-agnostic layer.

I’ll refactor it so transfer_context only prepares the generic block/object transfer arguments and calls lmc_ops.multi_layer_block_kv_transfer. The MUSA-specific native eligibility checks, slot mapping, staging, reshape handling, native-vs-fallback decision, and env/ABI checks will move under the MUSA/platform transfer implementation behind that entry point.

I’ll also move the native-specific tests out of transfer_context behavior tests or rewrite them to exercise the generic transfer entry, so future device backends do not need to account for MUSA logic.

maobaolong:
@superleo Can you explain why MUSA (which already supports`lmcache-driven` mode) is also developing an `engine-driven` approach?

superleo:
> @superleo Can you explain why MUSA (which already supports`lmcache-driven` mode) is also developing an `engine-driven` approach?

Thank you baolong,  this is a correctness-first MUSA MP path that reuses the Stage 2 worker-side gather/scatter logic. This also helps compatibility with older MUSA software stacks or devices where the handle-based path may not be available, stable, or validated. In those cases, engine-driven gives us a correctness-first non-handle path with torch fallback, while still allowing native MUSA acceleration behind `multi_layer_block_kv_transfer` when available.

And we agree transfer_context should stay device-agnostic. 
Updated this PR, moves the MUSA native checks, staging/reshape details, and fallback selection behind the platform-specific `multi_layer_block_kv_transfer` entry. The worker engine-driven path now calls the generic gather/scatter helpers; MUSA native acceleration is selected inside the MUSA platform layer when available.

superleo:
Hi @ApostaC @hlin99 @maobaolong ,This PR was intended to keep MUSA-specific logic inside the MUSA platform layer. However, CI has a common dispatch gap in the engine-driven block-transfer path, In the CI/source-build configuration, lmcache.c_ops.multi_layer_block_kv_transfer can be ptr-only. The existing common logic only routed CPU tensors through the Python fallback in that mode. For a non-CPU backend that needs a Python/platform fallback, the generic gather/scatter path could still select the compiled ptr-only path and fail before the backend-specific code had a chance to run.

that's why this PR adds a small generic platform hook mechanism:

- common code asks the platform registry whether the current device.type has a block-transfer hook;
- MUSA registers its hook inside the MUSA platform package;
- devices without a hook keep the existing behavior;
- hook failures fail closed and fall back to the generic Python path.

hlin99:
> Hi @ApostaC @hlin99 @maobaolong ,This PR was intended to keep MUSA-specific logic inside the MUSA platform layer. However, CI has a common dispatch gap in the engine-driven block-transfer path, In the CI/source-build configuration, lmcache.c_ops.multi_layer_block_kv_transfer can be ptr-only. The existing common logic only routed CPU tensors through the Python fallback in that mode. For a non-CPU backend that needs a Python/platform fallback, the generic gather/scatter path could still select the compiled ptr-only path and fail before the backend-specific code had a chance to run.
> 
> that's why this PR adds a small generic platform hook mechanism:
> 
> * common code asks the platform registry whether the current device.type has a block-transfer hook;
> * MUSA registers its hook inside the MUSA platform package;
> * devices without a hook keep the existing behavior;
> * hook failures fail closed and fall back to the generic Python path.

hi @superleo here're some comments, which may help
1. since we don't have musa device in CI, any change on musa should not affect existing cuda/cpu CI in theory
2. you need to rebuild c_ops in __init__.py. you can refer to xpu. you may have several ops backends as candidate, i.e. a) existing python fallback b) musa native? c) musa c++ (let's assume).  what you need to do during __init__ is to detect your device, environment, capability, etc to re-encapsulate the ops by selecting each ops from those backends. the assumption here is: the ops is decided one time during init, and won't be dynamic change. so the try/exception in this PR to dynamically switch ops is something unnecessary in my mind, and also have performance panelty.  so moving all the logic outside of the backend ,  not inside. eventually, musa c ops should have identical functions as python fallback, some functions are implemented by c, some by native, some by python fallbacks, etc. the point is every device has identical ops, so upper logic can all those ops without considering which device current is.
3. i still did not get the point of changing transfer_context. maybe when you fixed #2, the change in transfer_context is unnecessary.  there's also some logic in base.py to determine if backend supports tensor or ptr. as long as tensor is accepted by the backend, data path will use tensor rather than converting it to ptr. I feel current logic can fits your needs as long as you can rebuild the ops correctly.

hope i explain the design clearly. let me know if this works for you.

lmcache/v1/gpu_connector/musa_connectors.py
hlin99, line 1: mp mode does not use connectors. is this connector change for something else?

hlin99 (changes requested):

hlin99:
one thing might help you

do some modifications for musa, and on your local machine, try to get following two tests pass, which can verify if c ops is rebuild as expected.
tests/v1/test_python_ops_fallback.py
tests/v1/test_c_ops_fallback_parity.py

superleo:
> > Hi @ApostaC @hlin99 @maobaolong ,This PR was intended to keep MUSA-specific logic inside the MUSA platform layer. However, CI has a common dispatch gap in the engine-driven block-transfer path, In the CI/source-build configuration, lmcache.c_ops.multi_layer_block_kv_transfer can be ptr-only. The existing common logic only routed CPU tensors through the Python fallback in that mode. For a non-CPU backend that needs a Python/platform fallback, the generic gather/scatter path could still select the compiled ptr-only path and fail before the backend-specific code had a chance to run.
> > that's why this PR adds a small generic platform hook mechanism:
> > 
> > * common code asks the platform registry whether the current device.type has a block-transfer hook;
> > * MUSA registers its hook inside the MUSA platform package;
> > * devices without a hook keep the existing behavior;
> > * hook failures fail closed and fall back to the generic Python path.
> 
> hi @superleo here're some comments, which may help
> 
> 1. since we don't have musa device in CI, any change on musa should not affect existing cuda/cpu CI in theory
> 2. you need to rebuild c_ops in **init**.py. you can refer to xpu. you may have several ops backends as candidate, i.e. a) existing python fallback b) musa native? c) musa c++ (let's assume).  what you need to do during **init** is to detect your device, environment, capability, etc to re-encapsulate the ops by selecting each ops from those backends. the assumption here is: the ops is decided one time during init, and won't be dynamic change. so the try/exception in this PR to dynamically switch ops is something unnecessary in my mind, and also have performance panelty.  so moving all the logic outside of the backend ,  not inside. eventually, musa c ops should have identical functions as python fallback, some functions are implemented by c, some by native, some by python fallbacks, etc. the point is every device has identical ops, so upper logic can all those ops without considering which device current is.
> 3. i still did not get the point of changing transfer_context. maybe when you fixed [[Unittest] Better unit test to spin up redis and lmcache server during testing #2](https://github.com/LMCache/LMCache/pull/2), the change in transfer_context is unnecessary.  there's also some logic in base.py to determine if backend supports tensor or ptr. as long as tensor is accepted by the backend, data path will use tensor rather than converting it to ptr. I feel current logic can fits your needs as long as you can rebuild the ops correctly.
> 
> hope i explain the design clearly. let me know if this works for you.

Thanks Tony @hlin99 I think this design makes sense. The current PR tried to keep MUSA-specific logic out of MP by adding a platform hook, but that still made the common transfer path aware of backend-specific dispatch. I’ll revise it to follow the existing lmcache.c_ops backend composition model instead: add a MUSA ops backend selected in lmcache/__init__.py, keep transfer_context device-agnostic, and make the final lmcache.c_ops surface identical across devices.

hlin99:
thanks. the goal is to make each device isolated, so they dont need to aware each other. and logical layer/platform layer isolated as well - logical/platform layer can upgrade their implementation freely w/o unnecessary coupling

lmcache/v1/gpu_connector/musa_connectors.py
superleo: thanks tony, fixed. it was for stage2 backward compatibity

hlin99 (approved):
LGTM！
It's much more clear right now and aligns to the design. keeping device isolated and de-coupled is essential for easy maintain.

thanks for re-work this PR multiple times.

@ApostaC @maobaolong  do you want to double check? thx.

maobaolong (commented):
@superleo The code looks good for me, it would be really better if you could provide a user-facing document that explains how to set up an environment to enable the `musa native kv transfer`. It should also cover which third-party dependency packages need to be installed and how to install them. If these are missing, it seems like it can fallback to python ops automatically, hope we can introduce these to the user-facing document also.

maobaolong:
> @superleo The code looks good for me, it would be really better if you could provide a user-facing document that explains how to set up an environment to enable the `musa native kv transfer`. It should also cover which third-party dependency packages need to be installed and how to install them. If these are missing, it seems like it can fallback to python ops automatically, hope we can introduce these to the user-facing document also.

But if you plan to create a new followup PR to introduce related document, that works too.

superleo:
> > @superleo The code looks good for me, it would be really better if you could provide a user-facing document that explains how to set up an environment to enable the `musa native kv transfer`. It should also cover which third-party dependency packages need to be installed and how to install them. If these are missing, it seems like it can fallback to python ops automatically, hope we can introduce these to the user-facing document also.
> 
> But if you plan to create a new followup PR to introduce related document, that works too.

Thanks baolong @maobaolong , we will add related document in followup PR. Added to https://github.com/LMCache/LMCache/issues/3553.  And we also plan to migrate the musa native ops to upstream. 

And the k3-blend-test CI seems a env issue(upload files issue) [buildkite/k3-blend-test](https://buildkite.com/lmcache/k3-blend-test/builds/1191)

maobaolong (approved):

PR #3727, opened 2026-06-16
[Hotfix] raw block pre-commit warnings

DongDongJu:
Fixes two pre-existing `pre-commit run --all-files` failures on latest `dev` in the raw block code path:

- Marks the second regex fragment in `test_raw_block_uring_cmd.py` as raw to avoid Python `SyntaxWarning: invalid escape sequence`.
- Updates the Rust raw block LBA offset error formatting to use inline format args, satisfying Clippy with `-D warnings`.
These issues are not caused by PR #3718, but they currently prevent the full pre-commit hook set from passing on `dev`.

If applicable:


DongDongJu:
@ApostaC @sammshen PTAL!

DongDongJu:
it was fixed by other commit

PR #3731, opened 2026-06-17
[Feat] Expose CacheBlend coordinator knobs as CLI flags

ruizhang0101:

Adds `--blend-chunk-size` and `--blend-probe-stride` to the `lmcache coordinator`
command. These are the last two `MPCoordinatorConfig` fields that were only
reachable via `LMCACHE_MP_COORDINATOR_*` env vars; the other 7 already have flags.

`blend_chunk_size` is the global-CacheBlend match unit and **must equal the
LMCache chunk size the blend servers use**, so deployments that run a non-default
chunk size need to be able to set it explicitly rather than relying on the 256
default.

Pure CLI surface change: each flag defaults to `None` and falls back to the env
var / config default in `execute()`, matching the existing flags. No behavior
change when the flags are unset.

**If applicable**:

### DIFF
--- a/docs/source/mp/coordinator.rst
+++ b/docs/source/mp/coordinator.rst
@@ -23,8 +23,9 @@ Expected log output:
 
 The CLI accepts ``--host``, ``--port``, ``--instance-timeout``,
 ``--health-check-interval``, ``--eviction-check-interval``,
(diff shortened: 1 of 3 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3732, opened 2026-06-17
[Feat] Operator: reconcile the MP coordinator (LMCacheCoordinator CRD)

ruizhang0101:

Adds a third operator CRD, `LMCacheCoordinator`, so the operator can manage the
fleet MP coordinator instead of requiring a hand-rolled Deployment. Unlike the
engines (a per-node DaemonSet), the coordinator is a single fleet-wide service,
so it reconciles a **Deployment + ClusterIP Service** (port 9300, `/healthz`
probes), mirroring the existing `CacheBlendEngine` controller pattern (no
finalizer; owner-ref GC handles teardown).

Also wires engines to the coordinator: `LMCacheEngineSpec` and
`CacheBlendEngineSpec` gain a `coordinator` block (`ref` to an
`LMCacheCoordinator`, or an explicit `url`). The reconciler resolves a `ref` to
the coordinator's Service URL and emits `--coordinator-url` (+ advertise-IP via
the downward API, heartbeat / L2-event flags) so servers register on startup.

- ServiceMonitor support is wired but defaults disabled — the coordinator only
  exposes `/healthz`, not a Prometheus `/metrics` endpoint yet.
- Verified end-to-end on a GPU cluster: coordinator comes up, a GPU engine with
  `coordinator.ref` resolves the URL, becomes Ready, and registers in the
  coordinator's `/instances` fleet (new `coordinator_registration_smoke_test.go`,
  e2e_gpu).

**If applicable**:

- [x] this PR contains user facing changes - docs added (DESIGN.md, sample CR)


operator/api/v1alpha1/lmcacheengine_types.go
```diff
@@ -251,6 +251,44 @@ type RawL2AdapterSpec struct {
...
+
+	// url is an explicit coordinator base URL (e.g. http://coordinator:9300),
+	// used to target a coordinator the operator does not manage.
+	// +optional
+	URL *string `json:"url,omitempty"`
+
+	// advertiseIP is the IP the coordinator should reach this server at. When
+	// unset, the server's pod IP is injected via the downward API.
+	// +optional
+	AdvertiseIP *string `json:"advertiseIP,omitempty"`
```
royyhuang, line 278: Add docs indicating that users should not touch this option in almost every case. Only touch this if the user knows what he/she is really doing.

operator/internal/resources/coordinator_resolve.go
royyhuang, line 1: Can we move this `resource/` dir since I feel this file is a util file.

royyhuang (changes requested):

royyhuang (approved):
LGTM

ApostaC (approved):
LGTM!

Issue #3730, opened 2026-06-17
[good-first-issue] cleanup: Move type-only imports under TYPE_CHECKING


glbyktjys:
**Summary**
There are many places in the codebase where a class or module is imported only for use in type annotations, and not as a runtime value. These imports can be moved under a typing.TYPE_CHECKING guard so they do not run at import time.
This helps keep runtime imports cleaner and avoids unnecessary import-time dependencies.

**Details**
For example, in [lmcache/v1/distributed/api.py](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/distributed/api.py#L18) `IPCCacheServerKey` is imported, but it is used only as a type annotation in [`ipc_key_to_object_keys`](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/distributed/api.py#L210)

In cases like this, the import should be changed to:

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from lmcache.v1.multiprocess.custom_types import IPCCacheServerKey
```

**Note**
1. Only move imports that are used purely for type annotations. Imports that are used at runtime should remain as normal imports. This includes anything that is instantiated, called, subclassed, used with isinstance, or otherwise required during execution.
2. Please run all necessary testing after making the change to confirm that the refactor does not affect any runtime behavior or core logic. You could follow the testing guidance in [`AGENT.md`](https://github.com/LMCache/LMCache/blob/dev/AGENTS.md#testing)

PR #3736, opened 2026-06-17
chore: move type-only imports under TYPE_CHECKING -- l2_adapters (#3730)

HaozheZhang6:
Part of #3730. Moves type-only imports under `if TYPE_CHECKING:` for the L2 adapters (`base`, `config`, `fs`, `s3`, `mooncake_store`).

Narrowed to this batch per review -- follow-up PRs will cover the remaining modules.


glbyktjys:
Hi @HaozheZhang6, I think the current scope of this PR is enough. We can address adding `TYPE_CHECKING` guards and `from __future__ import annotations` in separate follow-up PRs if needed. Could you also make sure to address the bot nit comments? Thank you!

HaozheZhang6:
Done — ran isort over the `TYPE_CHECKING` blocks so the imports are sorted with the right `# Standard` / `# First Party` / `# Third Party` headings. Kept the scope as you suggested.

I left the `cache_context.py` and `hfbucket_connector.py` nits alone: they want `import x` ahead of `from x import y`, but `.isort.cfg` sets `from_first = true`, so isort keeps `from` first and already reports both as sorted.

ApostaC:
Hey @HaozheZhang6 , maybe let's start with a smaller set of code (i.e., 3-5 source files) at a time? This will make code reviews easier and avoid potential conflicts. Also, this leaves opportunities for other people who want to contribute.

HaozheZhang6:
Makes sense -- I'll trim this down to a small batch and follow up with the rest in separate PRs.

HaozheZhang6:
Trimmed down to the l2_adapters (base, config, fs, s3, mooncake_store) -- 5 files. I can take one more module as a follow-up; happy to leave the rest of #3730 open for others to pick up.

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

sammshen:
please fix DCO and merge conflicts thanks!

PR #3740, opened 2026-06-17
[MP][P2P] introduce P2P adapter

ApostaC:

Main changes:
- Export a new interface to pass the `layout_desc` as a hint in the l2 lookup interface -- this changes a lot of files but just a few lines in each file
- Implement the P2P adapter. Just the following 4 files.
  - lmcache/v1/distributed/l2_adapters/p2p_l2_adapter.py
  - tests/v1/distributed/l2_adapters/test_p2p_l2_adapter.py
  - tests/v1/distributed/l2_adapters/test_p2p_l2_adapter_integration.py
  - docs/design/v1/distributed/l2_adapters/p2p_l2_adapter.md

**If applicable**:


lmcache/v1/distributed/l2_adapters/p2p_l2_adapter.py
```diff
@@ -0,0 +1,392 @@
...
+        self,
+        keys: list[ObjectKey],
+        objects: list[MemoryObj],
+    ) -> L2TaskId:
+        """No-op. The store policy never routes store tasks to a P2P adapter."""
+        return _NOOP_STORE_TASK_ID
+
+    def pop_completed_store_tasks(self) -> dict[L2TaskId, L2StoreResult]:
+        """No-op. The store event fd is never signaled."""
+        return {}
```
ApostaC: Since these two functions will still be called by the store controller, we need to make sure that the _NOOP_STORE_TASK_ID is not accumulating in the store controller (which is essentially a memory leak)

lmcache/v1/distributed/l2_adapters/p2p_l2_adapter.py
```diff
@@ -0,0 +1,392 @@
...
+        self,
+        keys: list[ObjectKey],
+        objects: list[MemoryObj],
+    ) -> L2TaskId:
+        local_addresses = self._tc_context.get_transfer_channel_address(
+            [(obj.shm_offset, obj.shm_byte_length) for obj in objects]
+        )
+        with self._lock:
+            task_id = self._next_task_id_locked()
+            remote_addresses = [self._remote_addresses[key] for key in keys]
```
ApostaC: We may raise `KeyError` here. Please double-check whether the caller has exception handling. If not, we need to make sure this function does not raise.

lmcache/v1/distributed/l2_adapters/p2p_l2_adapter.py
ApostaC, line 1: I do see we used a lot of locks in this module. 
Conceptually, there is only one thread operating on this module's main functions -- which is the prefetch controller's loop thread.
The store controller will also call this module, but it's noop so it should be fine.

Given that, can we get rid of locks in this class?

ApostaC (commented):
Please see my comments.

lmcache/v1/distributed/l2_adapters/p2p_l2_adapter.py
```diff
@@ -0,0 +1,397 @@
...
+    - lookup_timeout_s: deadline for a lookup result before it counts as a miss.
+    - load_timeout_s: deadline for a load before it counts as a failure.
+    """
+
+    def __init__(
+        self,
+        peer_mq_server_url: str,
+        peer_transfer_channel_server_url: str,
+        lookup_timeout_s: float = 10.0,
+        load_timeout_s: float = 10.0,
```
ApostaC, line 81: can increase a bit

lmcache/v1/distributed/l2_adapters/p2p_l2_adapter.py
```diff
@@ -0,0 +1,397 @@
...
+    def submit_load_task(
+        self,
+        keys: list[ObjectKey],
+        objects: list[MemoryObj],
+    ) -> L2TaskId:
+        task_id = self._next_task_id
+        self._next_task_id += 1
+
+        remote_addresses = [self._remote_addresses.get(key) for key in keys]
+        if any(addr is None for addr in remote_addresses):
```
ApostaC: check addr is valid

sammshen (approved):
LGTM!

KuntaiDu (approved):
LGTM!

PR #3745, opened 2026-06-17
[Prefetch only sliding window 1/N] Add bitmap operations

KuntaiDu:

Add sliding-window-related bitmap operations, so that we can identify which ObjectKeys are needed under current lookup results and sliding window results.

**If applicable**:

### DIFF
--- a/benchmarks/microbenchmark/bitmap_ops_benchmark.py
+++ b/benchmarks/microbenchmark/bitmap_ops_benchmark.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # SPDX-License-Identifier: Apache-2.0
 """Microbenchmark for the ranked fold (:func:`fold_unfold_ranked`).
 
@@ -5,7 +6,7 @@ Compares the pure-Python reference against the native C++ implementation across
 request sizes, including a DeepSeek-scale hybrid case (1M tokens, 8 object
 groups mixing full attention and sliding window). Run with::
 
-    python -m lmcache.v1.distributed.bitmap_ops.benchmark
+    python benchmarks/microbenchmark/bitmap_ops_benchmark.py
 
 The native op scans the packed ``Bitmap`` buffer directly -- no Python per-bit
 loop and no ``Bitmap``<->tensor conversion -- so it stays sub-millisecond even
--- a/csrc/storage_manager/bitmap.cpp
+++ b/csrc/storage_manager/bitmap.cpp
@@ -145,7 +145,7 @@ size_t Bitmap::clo() const {
   return inverted.clz();
 }
 
-int64_t Bitmap::find_rightmost_one() const {
+int64_t Bitmap::highest_set_bit() const {
   // Scan bytes from the most significant down; the first non-zero byte holds
   // the highest set bit. Bits beyond size_ are never set, so the result is
   // always < size_.
--- a/csrc/storage_manager/bitmap.cpp
+++ b/csrc/storage_manager/bitmap.cpp
@@ -145,6 +145,22 @@ size_t Bitmap::clo() const {
   return inverted.clz();
 }
 
+int64_t Bitmap::find_rightmost_one() const {
+  // Scan bytes from the most significant down; the first non-zero byte holds
+  // the highest set bit. Bits beyond size_ are never set, so the result is
+  // always < size_.
+  for (size_t byte = data_.size(); byte-- > 0;) {
+    const uint8_t value = data_[byte];
+    if (value != 0) {
+      const unsigned highest =
+          31u -
+          static_cast<unsigned>(__builtin_clz(static_cast<unsigned>(value)));
+      return static_cast<int64_t>(byte * kBitsPerByte + highest);
+    }
+  }
+  return -1;
+}
+
 Bitmap Bitmap::operator&(const Bitmap& other) const {
   const size_t result_size = (size_ <= other.size_) ? size_ : other.size_;
   Bitmap result(result_size);
--- a/csrc/storage_manager/bitmap.cpp
+++ b/csrc/storage_manager/bitmap.cpp
@@ -2,6 +2,8 @@
 
 #include "bitmap.h"
 
+#include <algorithm>
+
 namespace lmcache {
 namespace storage_manager {
 
@@ -43,6 +45,34 @@ void Bitmap::set(size_t index) {
 
 void Bitmap::batched_set(const std::vector<size_t>& indices) {
   for (size_t idx : indices) set(idx);
+}
+
+void Bitmap::set_range(size_t start, size_t end) {
+  if (end > size_) end = size_;
+  if (start >= end) return;
+
+  const size_t first_byte = start / kBitsPerByte;
+  const size_t last_byte = (end - 1) / kBitsPerByte;  // inclusive
+  const unsigned head_low = bit_offset(start);
+
+  if (first_byte == last_byte) {
+    // Whole range lives in one byte: set bits [head_low, head_low + width).
+    const unsigned width = static_cast<unsigned>(end - start);  // 1..8
+    const uint8_t mask = static_cast<uint8_t>(((1u << width) - 1u) << head_low);
+    data_[first_byte] |= mask;
+    return;
+  }
+
+  // Head partial byte: bits [head_low, 8).
+  data_[first_byte] |= static_cast<uint8_t>(0xFFu << head_low);
+  // Middle full bytes.
+  std::fill(data_.begin() + static_cast<std::ptrdiff_t>(first_byte + 1),
+            data_.begin() + static_cast<std::ptrdiff_t>(last_byte),
+            static_cast<uint8_t>(0xFF));
+  // Tail partial byte: bits [0, tail_width).
+  const unsigned tail_width =
+      static_cast<unsigned>(end - last_byte * kBitsPerByte);  // 1..8
+  data_[last_byte] |= static_cast<uint8_t>((1u << tail_width) - 1u);
 }
 
 void Bitmap::clear(size_t index) {
--- a/csrc/storage_manager/bitmap.h
+++ b/csrc/storage_manager/bitmap.h
@@ -48,6 +48,18 @@ class Bitmap {
   void batched_set(const std::vector<size_t>& indices);
 
   /**
+   * @brief set every bit in the half-open range ``[start, end)`` to 1.
+   *
+   * Fills whole bytes at once, so setting a contiguous span is far cheaper than
+   * the equivalent per-bit ``set`` calls. ``end`` is clamped to the bitmap
+   * size; an empty or out-of-range range is a no-op.
+   *
+   * @param start First bit position to set (inclusive).
+   * @param end One past the last bit position to set (exclusive).
+   */
+  void set_range(size_t start, size_t end);
+
+  /**
    * @brief clear the bit at the specified index to 0.
    */
   void clear(size_t index);
--- a/csrc/storage_manager/bitmap.h
+++ b/csrc/storage_manager/bitmap.h
@@ -93,17 +93,14 @@ class Bitmap {
   size_t clo() const;
 
   /**
-   * @brief index of the highest set bit (the "right-most 1").
+   * @brief index of the highest set bit.
    *
-   * The return type is signed (``int64_t``, not ``size_t``) specifically so the
-   * empty case can be reported as ``-1``; an unsigned type has no spare value
-   * for "no bit set". Callers that feed the result back into ``set`` /
-   * ``clear`` / ``test`` (which take ``size_t``) must therefore check for
-   * ``-1`` first.
+   * The return type is signed (``int64_t``) so the empty bitmap can be reported
+   * as ``-1`` (an unsigned index has no spare value for "no bit set").
    *
    * @return the largest index whose bit is set, or ``-1`` if no bit is set.
    */
-  int64_t find_rightmost_one() const;
+  int64_t highest_set_bit() const;
 
   /**
    * @brief bitwise AND operation between two bitmaps.
--- a/csrc/storage_manager/bitmap.h
+++ b/csrc/storage_manager/bitmap.h
@@ -95,6 +95,12 @@ class Bitmap {
   /**
    * @brief index of the highest set bit (the "right-most 1").
    *
+   * The return type is signed (``int64_t``, not ``size_t``) specifically so the
+   * empty case can be reported as ``-1``; an unsigned type has no spare value
+   * for "no bit set". Callers that feed the result back into ``set`` /
+   * ``clear`` / ``test`` (which take ``size_t``) must therefore check for
+   * ``-1`` first.
+   *
    * @return the largest index whose bit is set, or ``-1`` if no bit is set.
    */
   int64_t find_rightmost_one() const;
--- a/csrc/storage_manager/bitmap.h
+++ b/csrc/storage_manager/bitmap.h
@@ -93,6 +93,13 @@ class Bitmap {
   size_t clo() const;
 
   /**
+   * @brief index of the highest set bit (the "right-most 1").
+   *
+   * @return the largest index whose bit is set, or ``-1`` if no bit is set.
+   */
+  int64_t find_rightmost_one() const;
+
+  /**
    * @brief bitwise AND operation between two bitmaps.
    *
    * @return a new Bitmap that is the result of the bitwise AND operation.
--- a/csrc/storage_manager/fold.h
+++ b/csrc/storage_manager/fold.h
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <utility>
+#include <vector>
+
+#include "bitmap.h"
+
+namespace lmcache {
+
+namespace storage_manager {
+
+/**
+ * @brief Fold/unfold over the ``group x chunk x kv_rank`` ranked layout.
+ *
+ * Computes the longest model-wide prefix that every object group can serve
+ * under its own rule (full attention or a cross-chunk sliding window), and the
+ * concrete keys each group must retain to serve it.
+ *
+ * The input ``found`` is group-major / chunk-major / rank-minor: bit
+ * ``g * (num_chunks * num_ranks) + j * num_ranks + r`` is set iff chunk ``j``
+ * of object group ``g`` is present for kv_rank ``r``. A chunk counts as present
+ * for a group only when every kv_rank shard is present.
+ *
+ * @param found Presence bitmap of length
+ *     ``group_windows.size() * num_chunks * num_ranks``.
+ * @param num_chunks Number of LMCache chunks in the request.
+ * @param num_ranks Number of kv_rank shards per chunk.
+ * @param group_windows Per-object-group cross-chunk sliding-window size in
+ *     chunks, in object-group order; ``<= 0`` means full attention.
+ *
+ * @return ``{hit_length, retain_mask}``: ``hit_length`` in chunks and a retain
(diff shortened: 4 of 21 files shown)


lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -134,6 +137,18 @@ class InFlightPrefetchRequest:
...
     """Which retained-subset policy to apply (see :class:`TrimPolicy`)."""
 
+    group_windows: tuple[int, ...] = ()
+    """Per-object-group cross-chunk sliding-window sizes (in chunks), in
+    object-group order. Non-empty turns ``PREFIX`` into the hybrid windowed
+    fold; empty keeps ``PREFIX`` as the contiguous leading-ones path. When set,
+    ``keys`` is laid out group-major / chunk-major / rank-minor:
+    ``len(keys) == len(group_windows) * num_chunks * num_ranks``."""
+
+    num_ranks: int = 1
```
KuntaiDu: Why we need this? Can we infer the `num_ranks` simply by looking at the length of keys?

lmcache/v1/distributed/api.py
```diff
@@ -269,6 +216,13 @@ class PrefetchHandle:
     """Original-key index of each key submitted to L2; maps the controller's
     local result bitmap back to original positions."""
 
+    fold_hit_length: int | None = None
```
KuntaiDu: ?????
This variable name is weird --- why the external caller needs to know that this "hit length" is from "fold"? And why does caller need to know what is fold?

Just call it prefix length --- and please examine all other variable names in the code diff. Avoid this type of naming that exposes unnecessary detail -- really confusing.

lmcache/v1/distributed/api.py
```diff
@@ -23,14 +23,25 @@
...
+    key (gap-tolerant) -- the "all servable positions" selection, for non-prefix
+    access patterns such as blend. Equivalent to the ALL selection with a
+    per-chunk (window-1) unfold. ``SEGMENTED_PREFIX`` is kept as a compatibility
+    alias of ``SPARSE`` (same behavior); external callers may pass either.
     """
 
     PREFIX = enum.auto()
-    SEGMENTED_PREFIX = enum.auto()
     SPARSE = enum.auto()
+    SEGMENTED_PREFIX = enum.auto()
```
KuntaiDu, line 32: You misunderstood what I meant. 
I want the code to literally look like:

PREFIX = enum.auto()
ALL = enum.auto()

SEGMENTED_PREFIX = ALL
SPARSE = ALL

lmcache/v1/distributed/storage_manager.py
```diff
@@ -428,11 +521,28 @@ def submit_prefetch_task(
...
+            return self._submit_fold_prefetch_l1(
+                keys,
+                l1_read_result,
+                extra_count=extra_count,
+                external_request_id=external_request_id,
+                group_windows=tuple(group_windows),
+                num_ranks=num_ranks,
+            )
+
+        if policy in (TrimPolicy.SPARSE, TrimPolicy.SEGMENTED_PREFIX):
```
KuntaiDu: if policy in `TrimPolicy.ALL`

lmcache/v1/distributed/storage_manager.py
```diff
@@ -428,11 +521,28 @@ def submit_prefetch_task(
 
         l1_read_result = self._l1_manager.reserve_read(keys, extra_count=extra_count)
 
-        if policy is TrimPolicy.SPARSE:
-            # SPARSE: retain a read lock on every L1 hit (not just the leading
-            # prefix) and send all L1 misses to L2 as one coalesced request.
-            # reserve_read locks only SUCCESS keys, so the found-set already
-            # equals the locked set -- nothing to release.
+        if policy is TrimPolicy.PREFIX and group_windows:
```
KuntaiDu: Please unify the code path.

Original code path should go through the same code path as hybrid, with the correct `group_windows` and other auxilary vars being set.

lmcache/v1/distributed/storage_controllers/bitmap_ops/fold.py
```diff
@@ -0,0 +1,207 @@
...
+        The half-open chunk range this group must keep. ``(0, 0)`` when
+        ``prefix_len <= 0``.
+    """
+    if prefix_len <= 0:
+        return (0, 0)
+    lo = 0 if window <= 0 else max(0, prefix_len - window)
+    return (lo, prefix_len)
+
+
+def fold_unfold(
```
KuntaiDu: Please use the efficient O(n) impl.

Also please implement it into
merge: group of bitmap across adapters -- group of bitmap
fold -> group of bitmap
find rightmost 1 -> group of bitmap
unfold -> group of bitmap
select: group of bitmap across adapters + unfold results ---> pruned results of groupo of bitmap across adapters.

But make sure the final testing behavior does not assert function output --- it should assert the results of the final output of TrimPolicy.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -1027,22 +918,38 @@ def retrieve(
...
+                        needed_keys
                     ) as memory_objs:
-                        if not memory_objs or len(memory_objs) != len(obj_keys):
+                        if not memory_objs or len(memory_objs) != len(needed_keys):
                             logger.error("Some keys not found during retrieve!")
                             return event.ipc_handle(), False
 
                         total_bytes += sum(mo.get_size() for mo in memory_objs)
 
+                        padded: list[MemoryObj | None] = (
```
KuntaiDu: Why padding is needed? Reply underneath this comment.

lmcache/v1/distributed/bitmap_ops/fold.py
```diff
@@ -0,0 +1,207 @@
+# SPDX-License-Identifier: Apache-2.0
```
KuntaiDu, line 1: Where is the object from different groups being separated? Reply me with the concrete line of code under this comment.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done — removed the threaded field/param. `num_ranks` is now inferred from the keys via `num_distinct_kv_ranks()` (count of distinct `kv_rank`s). This is also more correct for MLA, where all TP workers share one `kv_rank` and the count is 1 rather than `world_size`.

lmcache/v1/kv_layer_groups.py
```diff
@@ -562,34 +557,56 @@ def _detect_object_groups(
...
-        #    )
-        #    for sw_size_chunks, kernel_group_indices in sorted(
-        #        groups_by_sw_size.items(), key=lambda kv: kv[1][0]
-        #    )
-        # ]
+        return self._detect_object_groups_by_sw(engine_group_infos)
+
+    def _detect_object_groups_by_sw(
+        self, engine_group_infos: "Sequence[EngineGroupInfo]"
+    ) -> list[ObjectGroupInfo]:
```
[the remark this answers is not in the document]
KuntaiDu: Done — merged the two methods into one parameterless `_detect_object_groups`.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -906,22 +918,38 @@ def retrieve(
...
+                            [None] * lo
+                            + list(memory_objs)
+                            + [None] * (len(obj_keys) - hi)
+                        )
```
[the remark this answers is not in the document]
KuntaiDu: Done — removed the trailing `[None] * (len - hi)` pad (hi is always `len(obj_keys)`); only the leading pad remains.

lmcache/v1/distributed/api.py
KuntaiDu: Renamed `fold_hit_length` → `prefix_length` and reworded the docstring to drop the "fold" internals. Audited the rest of the diff — this was the only external-facing name exposing the mechanism.

lmcache/v1/distributed/api.py
KuntaiDu: Done — the enum is now exactly:
```python
PREFIX = enum.auto()
ALL = enum.auto()
SEGMENTED_PREFIX = ALL
SPARSE = ALL
```
so `TrimPolicy.SPARSE is TrimPolicy.ALL` and `TrimPolicy.SEGMENTED_PREFIX is TrimPolicy.ALL`.

lmcache/v1/distributed/storage_manager.py
KuntaiDu: Done — now `if policy is TrimPolicy.ALL` (covers `SPARSE`/`SEGMENTED_PREFIX` via the aliases).

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
KuntaiDu: The leading `[None] * lo` pad aligns the in-window chunks to their GPU block positions. `transfer_kv_per_object_group` maps each memory object to a block by its index in the list and skips the first `len(list) - sw_size_chunks` entries. For a sliding-window group the in-window chunks belong at indices `[lo, prefix_len)` (lo = prefix_len - w), not `[0, w)`, so the skipped head must be `None`-padded for the present tail to land at the correct offsets. (The trailing pad was redundant and is removed.) This is now documented inline at the retrieve loop.

lmcache/v1/distributed/bitmap_ops/fold.py
KuntaiDu: Concrete lines:
- Keys: `lmcache/v1/distributed/api.py:295` — `ipc_key_to_object_keys` outer loop `for object_group_id in object_group_ids` emits a distinct `ObjectKey` (distinct `object_group_id`) per group.
- Detection: `lmcache/v1/kv_layer_groups.py:551` `_detect_object_groups` buckets kernel groups into `ObjectGroupInfo`s by window size.
- Storage/transfer: `lmcache_driven_transfer.py:735` (store) and `:923` (retrieve) — per-object-group loops, each allocating/transferring its own `MemoryObj` (keys from `resolve_obj_keys(key, range(num_object_groups))`).

lmcache/v1/distributed/storage_manager.py
KuntaiDu: Partially. The non-hybrid path can run the same fold with `group_windows=(-1,)` for the L1 step, but the legacy path also performs the L2 handoff, which does not compose with the fold (the fold retain set is non-sequential, unlike the leading-prefix L2 continuation). A true single path needs the L1∪L2 fold. Quick question to pick the approach: is L2 active in the target deployment, or L1-only for now? L1-only → I drop the legacy walk and route everything through the fold; L2 active → it is the larger L1∪L2-fold change.

lmcache/v1/distributed/storage_controllers/bitmap_ops/fold.py
KuntaiDu: O(n) is done — replaced the O(N^2) scan with a running consecutive-present count (`run >= min(window, L)`), dropped the dead `_group_serves`. The staged restructure (`merge → fold → rightmost → unfold → select` over per-group bitmaps) plus moving tests to assert the `TrimPolicy` output rather than the helper output is still pending — will push that next.

lmcache/v1/distributed/storage_controllers/prefetch_policy.py
```diff
@@ -76,6 +81,57 @@ def select_l1_retentions(
...
+                    f"fold key count {num_keys} is not a multiple of "
+                    f"len(group_windows) * num_ranks = {divisor}"
+                )
+            num_chunks = num_keys // divisor
+            prefix_length, retained = fold_unfold_ranked(
+                presence, num_chunks, num_ranks, group_windows
+            )
+            return retained, prefix_length
+
+        retained = build_trim_mask(presence, num_keys, selection)
```
KuntaiDu: Rename this function maybe --- currently the name feels like it should belongs to TrimPolicy while it's not.

docs/design/v1/distributed/bitmap_ops/README.md
```diff
@@ -0,0 +1,114 @@
...
+lookup key layout: a chunk counts as present only when **all** kv_rank shards are
+present, and the retain mask re-expands over every shard. This replaces the old
+`// world_size` arithmetic with an explicit rank reduction.
+
+## Where it runs, and why
+
+LMCache prefetch is a two-stage **predict-then-execute** path: lookup (presence)
+→ plan → load → finalize (reconcile rare failures). The fold lives in
+`PrefetchPolicy.compute_retained` and is driven by the prefetch controller, not
+inside `ObjectKey` or the L2 adapters, because:
```
KuntaiDu: No need to explain the motivation in such detail. Just say what it is for and breifly describer its functionality.

lmcache/v1/distributed/bitmap_ops/fold.py
```diff
@@ -0,0 +1,317 @@
...
+    return fold_unfold_ranked(found, num_chunks, 1, group_windows)
+
+
+def _fold_python(
+    found: Bitmap,
+    num_chunks: int,
+    num_ranks: int,
+    group_windows: Sequence[int],
+) -> Bitmap:
+    """Pure-Python fallback for :func:`fold` (used if the native op is absent)."""
```
KuntaiDu: Please say this is only for testing purpose.
Also, please adjust the docstring format. Find other places in LMCache code and @CLAUDE.md as reference.

tests/v1/distributed/test_bitmap_ops.py
```diff
@@ -0,0 +1,535 @@
...
+
+# --------------------------------------------------------------------------- #
+# Native vs pure-Python fallback equivalence (the native ops are the oracle's  #
+# fast path; the fallbacks must agree bit-for-bit)                             #
+# --------------------------------------------------------------------------- #
+
+
+class TestNativeMatchesFallback:
+    def test_fold_and_unfold_random_equivalence(self):
+        rng = random.Random(1234)
```
KuntaiDu: I do think we need large-scale test. But can we construct it manually (and make it challenging, so that a wrong algorithm likely result in wrong output) instead of doing it randomly? It will make things flaky.

csrc/storage_manager/bitmap.cpp
```diff
@@ -115,6 +145,22 @@ size_t Bitmap::clo() const {
   return inverted.clz();
 }
 
+int64_t Bitmap::find_rightmost_one() const {
```
KuntaiDu: Why `int64_t` isntead of `size_t`? If it is `int64_t` it means that we cannot use it in, say `clear` or something?

lmcache/v1/distributed/bitmap_ops/fold.py
```diff
@@ -0,0 +1,336 @@
...
+            ``g * (num_chunks * num_ranks) + j * num_ranks + r`` set iff chunk
+            ``j`` of object group ``g`` is present for kv_rank ``r``.
+        num_chunks: number of LMCache chunks in the request.
+        num_ranks: number of kv_rank shards per chunk (``world_size`` at lookup).
+        group_windows: per-object-group cross-chunk sliding-window size in
+            chunks, in object-group order; ``<= 0`` means full attention.
+
+    Returns:
+        A bitmap of size ``num_chunks + 1``; bit ``L`` set iff every group can
+        serve a length-``L`` prefix. Bit 0 is always set. Feed it to
```
yoo-kumaneko: why do we need bit 0 here at all if it's always set?

lmcache/v1/distributed/bitmap_ops/fold.py
```diff
@@ -0,0 +1,336 @@
...
+    if not group_windows:
+        raise ValueError("group_windows must be non-empty")
+    if num_chunks < 0:
+        raise ValueError(f"num_chunks must be >= 0 (got {num_chunks})")
+
+    return _native_fold(found, num_chunks, num_ranks, list(group_windows))
+
+
+def find_rightmost_one(servable: Bitmap) -> int:
+    """Index of the highest set bit (the right-most 1), or -1 if none.
```
yoo-kumaneko: if bitmap[0] is always zero, then it will never return -1. Not sure if I understand it correctly.

yoo-kumaneko (approved):
had a quick skim of the code. A high-performance implementation of a set of bitmap primitives. Looks good to me.

csrc/storage_manager/bitmap.h
```diff
@@ -80,6 +92,19 @@ class Bitmap {
...
+   *
+   * The return type is signed (``int64_t``, not ``size_t``) specifically so the
+   * empty case can be reported as ``-1``; an unsigned type has no spare value
+   * for "no bit set". Callers that feed the result back into ``set`` /
+   * ``clear`` / ``test`` (which take ``size_t``) must therefore check for
+   * ``-1`` first.
+   *
+   * @return the largest index whose bit is set, or ``-1`` if no bit is set.
+   */
+  int64_t find_rightmost_one() const;
```
ApostaC: In my mental model, "rightmost" is the lowest set bit. 
For example, for a bitmap corresponding to 8 keys, if it's `0b00001111`, it means the first 4 keys are not found, and the last 4 keys are found.

In this mental model, the rightmost one is 1, while the highest bit is the first 0.

Can we have a better name to avoid the confusion here?

csrc/storage_manager/fold.h
```diff
@@ -0,0 +1,64 @@
...
+ * @param num_chunks Number of LMCache chunks in the request.
+ * @param num_ranks Number of kv_rank shards per chunk.
+ * @param group_windows Per-object-group cross-chunk sliding-window size in
+ *     chunks, in object-group order; ``<= 0`` means full attention.
+ *
+ * @return A bitmap of size ``num_chunks + 1``; bit ``L`` set iff every group
+ *     can serve a length-``L`` prefix. Bit 0 is always set.
+ */
+Bitmap fold(const Bitmap& found, size_t num_chunks, size_t num_ranks,
+            const std::vector<int64_t>& group_windows);
```
ApostaC: What's the reason we return the bitmap for `num_chunks + 1`? (I'm wondering what's the use case for bit 0)

benchmarks/microbenchmark/bitmap_ops_benchmark.py
ApostaC, line 1: Is this the best place to put the benchmarking script? How about we directly put this into `benchmarks/` folder?

ApostaC (commented):
Mostly LGTM! Left a few quick comments

lmcache/v1/distributed/bitmap_ops/fold.py
ApostaC: same question here

KuntaiDu: The goal is to use bit 0 as a sentinel to simplify bit arithmetics (e.g. `find_rightmost_one`).

Basically bit 0 set to 1 means that any request can always hit 0 chunks from LMCache.

KuntaiDu: But I agree it causes more confusion than how much it helps. I'll remove bit 0.

csrc/storage_manager/bitmap.h
KuntaiDu: renamed.

benchmarks/microbenchmark/bitmap_ops_benchmark.py
KuntaiDu: moved.

csrc/storage_manager/fold.h
KuntaiDu: removed bit 0

ApostaC (approved):
LGTM!

PR #3746, opened 2026-06-17
ci(operator): tag pre-release images as latest-rc, not latest

royyhuang:

The operator release workflow tags every published release with the floating `:latest` image tag and repoints the `operator-latest` GitHub release alias (and its `install.yaml`) at it. A pre-release such as `operator-v0.4.8rc1` therefore clobbers the stable `:latest` image and the stable install manifest that production users pull.

This change detects `rc`/`alpha`/`beta` in the release version and:
- pushes the floating tag as `:latest-rc` for pre-releases, `:latest` for stable releases;
- skips the `operator-latest` stable alias update for pre-releases.

The exact version tag (e.g. `:v0.4.8rc1`) is unchanged. Detection reads the version string rather than the GitHub "pre-release" checkbox, so it works regardless of how the release is cut.

Stable release behavior is unchanged: `:vX.Y.Z` + `:latest` + `operator-latest` alias update. Pre-release detection is a regex `(rc|alpha|beta)` on the version string.

**If applicable**:

### DIFF
--- a/.github/workflows/operator_release.yml
+++ b/.github/workflows/operator_release.yml
@@ -42,9 +42,20 @@ jobs:
       - name: Extract version from tag
         id: version
         run: |
-          # Strip "operator-" prefix from tag to get version
+          # Strip "operator-" prefix from tag to get version (e.g. operator-v0.4.8rc1 -> v0.4.8rc1).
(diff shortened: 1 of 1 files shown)


deng451e (approved):
LGTM

ApostaC (approved):

PR #3747, opened 2026-06-17
[CI] Publish rc release tags as clean pre-releases + gate CLI wheel against native ops

deng451e:
Description:

  `lmcache-cli` only ever shipped `.devN` pre-releases on PyPI. The build
  workflows strip every tag that isn't `vX.Y.Z`, so an `rc` release (e.g.
  `v0.4.8rc1`) loses its tag and setuptools-scm falls back to the previous
  stable tag, stamping a `0.4.8.devN` build.

  - Broaden the "Remove non-release tags" regex to keep PEP 440 `rc/a/b`
    pre-release tags while still stripping invalid variant tags
    (`v*-cu129`, `v*-alpha`; see #3123). `v0.4.8rc1` now ships as a clean
    `0.4.8rc1` pre-release. Applied to all three build workflows
    (cli / main / cu129) so every artifact agrees on the version.
  - Add a CLI smoke-test gate that fails the build if `import lmcache.c_ops`
    succeeds, ensuring native ops never leak into the pure-Python wheel
    (complements the existing `.so`/`.pyd` bundle check).

  `rc` releases now publish to production PyPI as pre-releases (hidden from
  default `pip install`, installable via `--pre`). A stable `lmcache-cli`
  still requires a clean `vX.Y.Z` tag with a green CLI build.

  **If applicable**:

  - [ ] this PR contains unit tests, adding 'Refs #<issue number>' will not close the issue, but help provide the reviewer

### DIFF
--- a/.github/workflows/build_cli_artifacts.yml
+++ b/.github/workflows/build_cli_artifacts.yml
@@ -46,8 +46,10 @@ jobs:
               with:
                 fetch-depth: 0
 
-            - name: Remove non-release tags
-              run: git tag -l | grep -vE '^v[0-9]+\.[0-9]+\.[0-9]+$' | xargs -r git tag -d
+            # Keep stable vX.Y.Z and rc/a/b pre-release tags; strip variant tags
+            # (v*-cu129, v*-alpha) that yield invalid PEP 440 versions (see #3123).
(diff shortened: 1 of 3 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3748, opened 2026-06-18
feat(local-disk): extend by_gpu sharding for worker-aware multi-device paths

daegyu94:

This PR extends the existing `by_gpu` sharding policy in the local disk
backend to support worker-aware multi-device path assignment.

Previously, local-disk `by_gpu` selected a single path from the physical GPU
device id. That had a few limitations:

1. extra disk paths could remain idle when `#paths > #workers`
2. physical device id is not the right key in data-parallel or device-remapped
   deployments
3. global cache management in multi-path mode could break path locality after
   eviction, causing KV cache to be re-inserted on a different path than the
   one originally mapped for that key

This change extends `by_gpu` to use local worker metadata when available:

- `#paths == #workers`: one path per worker
- `#paths > #workers`: each worker gets a contiguous subset of paths
- `#workers > #paths`: workers share paths by modulo mapping

Within an assigned subset, KV cache is routed by stable per-key hashing.

This PR also changes multi-path local-disk cache management from global to
path-local. Each assigned path now has its own quota, usage tracking, and
eviction-policy state, which keeps cache placement aligned with the intended
path mapping even after eviction.

When local worker metadata is not available, `by_gpu` keeps the legacy
single-path `device_id % num_paths` behavior.

- This extends the existing `by_gpu` policy; it does not introduce a new
  sharding mode.
- The worker-aware branch is keyed by local rank, which makes path assignment
  coherent in DP and device-remapped setups.
- Path count and local worker count must be equal or exact multiples;
  unsupported uneven layouts fail at startup.
- In multi-path mode, cache management is path-local rather than global.
- Tests were added for worker-aware assignment, legacy compatibility, subset
  routing, and multi-path local-disk backend behavior.
- Docs and the local backend example were updated accordingly.

**If applicable**:

### DIFF
--- a/docs/source/api_reference/configurations.rst
+++ b/docs/source/api_reference/configurations.rst
@@ -42,10 +42,10 @@ Basic cache settings that control the core functionality of LMCache.
      - Path (or comma-separated paths) to local disk cache directories. Format: ``"file:///path/to/cache"`` or ``"/path/a,/path/b"`` for multi-device I/O. See ``local_disk_path_sharding`` for how paths are assigned to GPUs.
    * - local_disk_path_sharding
      - LMCACHE_LOCAL_DISK_PATH_SHARDING
-     - Strategy for selecting a path when multiple paths are provided. Currently only ``"by_gpu"`` is supported, which selects paths based on GPU device ID (default: "by_gpu").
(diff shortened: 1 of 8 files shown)


DongDongJu:
Not sure this UT failed by this PR. but @daegyu94 Could you check once?

PR #3755, opened 2026-06-18
first commit on token dropping api

rannnayy:

**If applicable**:

### DIFF
--- a/docs/design/sdk/kvcache.md
+++ b/docs/design/sdk/kvcache.md
@@ -65,7 +65,7 @@ The SDK is a **separate process** from the LMCache server.
 ```
 SDK process                          LMCache server process
 -----------                          ----------------------
-LMCacheSDKContext
+LMCacheKVCacheContext
   └ MessageQueueClient ──ZMQ──▶ MQ dispatch
                                   ├ EngineDrivenTransferModule  (PREPARE_/COMMIT_ STORE/RETRIEVE)
                                   │     └ ShmTransferStrategy
@@ -83,7 +83,7 @@ LMCacheSDKContext
 
 ## Registration handshake
 
-`connect()` constructs an `LMCacheSDKContext`, whose `__init__` runs a one-time handshake.
+`connect()` constructs an `LMCacheKVCacheContext`, whose `__init__` runs a one-time handshake.
 All calls are blocking and bounded by `timeout`:
 
 1. `GET_WORLD_SIZE [model_name]` → `world_size`. **Raises `KVCacheSDKError` if 
@@ -102,11 +102,11 @@ All calls are blocking and bounded by `timeout`:
 
 ## Public API
 
-Exported from `lmcache.sdk`: `LMCacheSDKContext`, `KVCacheSDKError`.
+Exported from `lmcache.sdk`: `LMCacheKVCacheContext`, `KVCacheSDKError`.
 Operations are module-level functions in `lmcache.sdk.kvcache`: `connect`, `close`,
 `retrieve`, `store`.
 
-### `connect(url, model_name, timeout=60.0) -> LMCacheSDKContext`
+### `connect(url, model_name, timeout=60.0) -> LMCacheKVCacheContext`
 
 Opens a `MessageQueueClient` to `url` (TCP) and runs the [registration
 handshake](#registration-handshake). The returned context is context-manager capable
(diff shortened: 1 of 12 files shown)


lmcache/v1/multiprocess/modules/management.py
```diff
@@ -99,74 +58,72 @@ def get_handlers(self) -> list[HandlerSpec]:
                 self.report_block_allocations,
                 ThreadPoolType.NORMAL,
             ),
+            HandlerSpec(
```
KuntaiDu: Can we do this through HTTP? This code path is meant for handling things related to KV cache. For information-related thing HTTP endpoint is a better choice.

lmcache/sdk/kvcache.py
```diff
@@ -0,0 +1,622 @@
...
+    Args:
+        url: ZMQ endpoint URL for the LMCache message queue.
+        model_name: Model name used by the running LMCache server instance.
+        timeout: Timeout in seconds for blocking MQ calls. Defaults to 60.
+    
+    Returns:
+        An initialized LMCacheSDKContext instance.
+        Ready to be passed to close(), retrieve(), and store() functions.
+    """
+    return LMCacheSDKContext(
```
KuntaiDu: Naming-wise, maybe `LMCacheKVCacheContext`? THe name `LMCacheSDKContext` sounds like it is a general context across all future LMCache sdk.

lmcache/v1/multiprocess/protocols/engine.py
```diff
@@ -237,4 +239,15 @@ def get_protocol_definitions() -> dict[str, ProtocolDefinition]:
             response_class=bool,
             handler_type=HandlerType.BLOCKING,
         ),
+        # Register SDK transfer strategy (for non-GPU transfers initiated by the SDK)
+        # Payload:
+        #   - instance_id: int - Unique identifier for the SDK instance
+        #   - model_name: str - Name of the model associated with the SDK instance
+        #   - world_size: int - World size of the SDK instance
+        # Returns: None
+        "REGISTER_SDK_TRANSFER_STRATEGY": ProtocolDefinition(
```
KuntaiDu, line 248: Let's talk --- not sure why this API is needed

KuntaiDu (commented):
Some comments, PTAL

lmcache/v1/multiprocess/protocols/engine.py
rannnayy: All four protocols--`PREPARE/COMMIT_ STORE/RETRIEVE`--requires `instance_id` in the payload. The `instance_id` is passed to `_resolve_for_transfer` to get the entry from `_engine_driven_contexts` and the strategy from `_strategies`, which will be used to do the work. This is the reason behind the extra addition of protocol.

`register_kv_cache_engine_driven_context` almost does the same thing, but it needs the layout descriptor. Either specifying the model config during startup, or call `resolve_model_name` which gets the layout in the engine, which is done by `register_sdk_transfer_strategy`.

docs/design/sdk/kvcache.md
```diff
@@ -0,0 +1,280 @@
...
+   reservation, and lock release.
+2. **Data plane (POSIX shared memory)** — the KV bytes are read/written directly in the
+   server's L1 SHM pool; only metadata crosses the socket.
+
+```py
+import lmcache.sdk.kvcache as lmc_sdk
+
+ctx = lmc_sdk.connect(
+	url="tcp://localhost:5555", 
+	model_name="Qwen/Qwen3-8B"
```
KuntaiDu, line 21: Maybe we also need to add cache salt

rannnayy: Cache salt exists in both `retrieve()` and `store()`. Since `connect()` is a one time at first request send, `cache_salt` wasn't added there (correct me if I'm wrong).

lmcache/v1/multiprocess/modules/management.py
rannnayy: Done. Used HTTP call to /conf and /status to get world_size, shm_name, and chunk_size, rather than exposing new MQ path.

lmcache/sdk/kvcache.py
rannnayy: Done. Revised the name to LMCacheKVCacheContext instead of LMCacheSDKContext.

PR #3758, opened 2026-06-18
Fix: periodic thread restart after failure

zupengwang:

Reset `PeriodicThread`'s running state when its run loop exits.

When `_execute()` raised `IrrecoverableException`, the run loop stopped without
clearing `_running`. Although `is_running` correctly reported that the worker was no
longer alive, a subsequent `start()` checked the stale flag, treated the thread as
still running, and refused to restart it.

The run loop now clears `_running` whenever it exits. The regression test exercises
the public API by terminating the first worker with an irrecoverable exception and
verifying that a second `start()` creates and runs a new worker.

Validation on macOS:

- `pytest -q --confcutdir=tests/v1 tests/v1/test_periodic_thread.py`
- SPDX header check, torch-device guard, isort, Ruff, Ruff format, codespell,
  clang-format, and mypy pre-commit hooks

Validation on Linux x86_64 (`Ubuntu 22.04`):

- `pytest -q --confcutdir=tests/v1 tests/v1/test_periodic_thread.py`
- `cd rust/raw_block && cargo fmt --check`
- `cd rust/raw_block && cargo clippy -- -D warnings`

The Rust checks were run on Linux because the `io-uring` dependency is Linux-only.

**If applicable**:


zupengwang:
Hi maintainers, this PR is ready for review. It fixes the stale `PeriodicThread` running state after an `IrrecoverableException` and includes a regression test for restarting the worker. All required CI checks and DCO are passing, and the Linux Rust checks also pass. Thanks!

zupengwang:
Hi maintainers, could someone review this restart-state fix when you have a chance? DCO and the Buildkite checks are passing, and the PR includes a regression test for restarting `PeriodicThread` after an `IrrecoverableException`. Happy to narrow or reshape the test if there is a preferred pattern.

chunxiaozheng (approved):
@zupengwang Sorry for the delay review, LGTM, thanks for your contribution!

chunxiaozheng:
@maobaolong could you help take a look?

maobaolong:
@zupengwang Good catch, and thanks @chunxiaozheng to let this PR re-active, LGTM.

maobaolong (approved):

Issue #3768, opened 2026-06-19
 TensorMemoryAllocator.memcheck() can race with lazy allocator address-space growth


superleo:
Label
bug, memory-management, onboarding

Summary
TensorMemoryAllocator.memcheck() can report a false allocator inconsistency when lazy allocator address-space growth happens between separate size reads.

Details
TensorMemoryAllocator.memcheck() previously read free size, allocated size, and heap size through separate AddressManager calls. With lazy allocation, sbrk() may grow the address space between those reads. That can cause memcheck() to compare values from different allocator states and incorrectly fail the consistency check.

The same split-read pattern also affected L1 memory usage reporting when it computed used and total size from separate address-manager reads.

Steps / Reproduction (if applicable)

Create a TensorMemoryAllocator with a smaller initial address space than the backing buffer.
Trigger memcheck() while lazy address-space growth can occur between get_free_size() and get_heap_size() / allocated-size reads.
Observe that the size-consistency check may compare stale free size against the newer heap size.
Expected Outcome / Goal
Allocator consistency checks and memory-usage reporting should read free, allocated, and heap sizes from a single locked snapshot so all values represent the same allocator state.

Actual Outcome (if applicable)
memcheck() may incorrectly log "Memory allocator size is inconsistent" and return False, even though the allocator state is valid.

Additional Context
https://buildkite.com/lmcache/k3-unit-tests/builds/3543/list?sid=019edadb-d813-457e-9d51-1269f3173421&tab=output

superleo:
close it since https://github.com/LMCache/LMCache/pull/3799 is working on the issue.

PR #3769, opened 2026-06-19
fix: Fix ci lazy allocator memcheck race

superleo:

Fixes a race in memory allocator consistency checks by reading free, allocated, and heap sizes from a single locked `AddressManager` snapshot.

This prevents `TensorMemoryAllocator.memcheck()` and L1 memory usage reporting from mixing allocator state before and after lazy address-space growth.

Issue: https://github.com/LMCache/LMCache/issues/3768

The regression test monkeypatches the old `get_free_size()` path to grow the address space after reading free size. `memcheck()` now stays consistent because it uses the new snapshot API instead.

**If applicable**:


superleo:
close it since #3799 is working on the issue.

PR #3771, opened 2026-06-19
[Bugfix] [MP] fix missing cache_salt in free_lookup_locks call

### LINKED ISSUE #3770: Bug: `free_lookup_locks` missing `cache_salt` causes KV chunks to stay read-locked
### Problem

In `update_state_after_alloc`, `free_lookup_locks` is called without `cache_salt`:

```python
self.scheduler_adapter.free_lookup_locks(
    token_ids=list(tracker.all_token_ids),
    start=0,
    end=free_end,
    request_id=request.request_id,
    # cache_salt missing!
)
```

Earlier in maybe_submit_lookup_request, the lookup was submitted with cache_salt=tracker.cache_salt, so the server built an ObjectKey using that salt. The mismatched free_lookup_locks call (defaulting to "") constructs a different ObjectKey, so the server can't locate and release the lock.

In multi-tenant deployments where each request carries a different cache_salt, matched KV chunks remain read-locked until TTL expiry, blocking eviction.

The same bug exists in vllm's upstream fallback copy and has a pending fix there: vllm-project/vllm#44097.

Since vllm preferentially imports LMCacheMPConnector from the lmcache package at runtime (falling back to its own copy only when lmcache is not installed),

chfeng-cs:
Fix #3770. In `update_state_after_alloc`, `free_lookup_locks` was called without `cache_salt`, causing it to default to `""`. Since the earlier `maybe_submit_lookup_request` used `tracker.cache_salt` to build the `ObjectKey`, the server couldn't match and release the lock — leaving KV chunks read-locked until TTL expiry in multi-tenant deployments.

Fix: pass `cache_salt=tracker.cache_salt` to `free_lookup_locks`, consistent with the lookup submission.

The same bug exists in vllm's upstream fallback copy: vllm-project/vllm#44097.


ApostaC (approved):

sammshen (approved):
LGTM!

PR #3773, opened 2026-06-19
Fix: return boolean heartbeat running status

zupengwang:

Ensure that `HeartbeatService.status()` always returns a boolean value for
`running`.

Previously, `self.thread and self.thread.is_alive()` evaluated to `None` before
the heartbeat thread was created. As a result, the heartbeat status API returned
`"running": null` even though its documented contract specifies a boolean.

Use an explicit `None` check so the initial status is `"running": false`, and
tighten the existing unit test to enforce the boolean contract.

Validation performed on Linux x86_64:

- `pytest -q --confcutdir=tests/lmcache_frontend tests/lmcache_frontend`
  - 9 passed
  - All hooks passed, including mypy, Rust fmt, and Rust clippy

No documentation change is needed because this fix aligns the implementation
with the existing documented boolean contract.

**If applicable**:


zupengwang:
Hi maintainers, just checking whether this small heartbeat status fix is still useful. It keeps the existing documented boolean contract by returning `running: false` before the heartbeat thread exists, and the DCO plus Buildkite checks are passing. Happy to adjust if you prefer a different API/test shape.

maobaolong (approved):
@zupengwang Good catch! 

Image

zhengfeihe (approved):
Nice Catch~

PR #3775, opened 2026-06-19
[Bugfix] [CLI bench] Unregister KV cache from MP server on bench server exit

ApostaC:

`lmcache bench server` registers a KV cache context with the MP server at startup but never unregisters it on exit, leaking one context entry on the server per bench run.

This PR adds the KV cache unregistration logic to the bench's shutdown path.

**If applicable**:


KuntaiDu (approved):
lgtm

sammshen (approved):

PR #3776, opened 2026-06-19
[Prefetching sliding window 2/N] Propagate sliding-window info to the prefetch path

KuntaiDu:

Propagate the sliding window metadata from Group all the way to prefetch.

**If applicable**:


lmcache/v1/kv_layer_groups.py
```diff
@@ -24,6 +25,29 @@
 
 logger = init_logger(__name__)
 
+# Toggle for splitting kernel groups into per-window object groups. Off by
```
KuntaiDu: If there are standard ways of setting env vars in LMCache multiprocess mode (not LMCache in process mode), can you follow the convention there?

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -352,7 +369,11 @@ def query_prefetch_status(
         # 1. the world size is the same between keys
         # 2. the lookup sort the keys in prefix order and breaks at the
         #    first failure
-        found_count = found.count_leading_ones() // job.world_size
+        # Chunk-major layout packs world_size * num_object_groups keys per chunk,
+        # so a chunk is a full-attention hit only when its whole slice is set.
+        found_count = found.count_leading_ones() // (
```
KuntaiDu: Can you make this to an auxiliary function? Because when sliding window is there, this arithmetic is no longer true (unless the last group is always full attention group)

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -138,6 +134,12 @@ class InFlightPrefetchRequest:
     policy: TrimPolicy = TrimPolicy.PREFIX
     """Which retained-subset policy to apply (see :class:`TrimPolicy`)."""
 
+    group_windows: tuple[int, ...] = ()
+    """Per-object-group cross-chunk sliding-window sizes (in chunks), in
+    object-group order, propagated from the lookup for the prefetch policy.
+    Carried for the upcoming windowed fold; the current full-attention prefix
+    trim does not consume it. Empty means a single full-attention group."""
```
ApostaC: We only need half of the first sentence (`Per-object-group cross-chunk sliding-window sizes (in chunks), in object-group order`) plus some description for the values under different cases.

E.g.

```suggestion
    """Per-object-group cross-chunk sliding-window sizes (in chunks), in
    object-group order. -1 means full attention groups.
```

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -352,6 +327,10 @@ def submit_prefetch_request(
                 workers can each consume one read lock.
             policy: Which retained-subset policy to apply (see
                 :class:`TrimPolicy`).  Defaults to ``PREFIX``.
+            group_windows: Per-object-group cross-chunk sliding-window sizes (in
+                chunks), propagated to the request for the prefetch policy. The
+                current full-attention prefix trim does not consume it; empty
+                means a single full-attention group.
```
ApostaC: Quick tip: you can tell your Claude code to rewrite the new docstrings: "Do not say anything about implementation details inside the method since they should be invisible to the callers. Just describe the interface's behavior that the caller needs to be aware of."

Following the above instructions, it should remove 
```
propagated to the request for the prefetch policy. The
current full-attention prefix trim does not consume it; empty
means a single full-attention group.
```
as they are all class-internal implementation-related details.

(Also, tell it "do not touch any existing docstrings")

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -645,42 +534,42 @@ def register_kv_cache(
...
-                cache_context=cache_context,
-                model_name=model_name,
-                world_size=world_size,
-                last_seen=now,
-                has_liveness_signal=False,
-            )
+        # Publish the per-object-group cross-chunk windows so the (manager-blind)
+        # lookup path can size the chunk-major key layout and propagate the
+        # windows downstream. ``-1`` for a group means full attention; with
+        # object-group separation disabled this is a single ``(-1,)`` window.
```
ApostaC: The comments are not needed.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -67,6 +68,34 @@ def compute_extra_count(
...
+
+    The chunk-major key layout packs ``world_size * num_object_groups`` keys per
+    chunk, so a fully-present prefix of ``key_prefix_len`` keys spans
+    ``key_prefix_len // (world_size * num_object_groups)`` chunks.
+
+    NOTE: this is only correct under full attention -- every object group present
+    for every hit chunk. Once sliding-window prefetch lands (the windowed fold),
+    the model-wide hit is no longer a uniform contiguous key prefix, and the
+    chunk count must come from the fold's reported hit length rather than this
+    arithmetic. Isolated here so that switch is a one-function change.
```
ApostaC: nit: we can have a `Note:` section (just like `Args:` and `Returns:` section)

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -78,6 +107,10 @@ class _PrefetchJob:
     # or chunk_hashes is empty).  Consumed at ``MP_LOOKUP_PREFETCH_END``
     # emission time in ``query_prefetch_status``.
     requested_tokens: int
+    num_object_groups: int = 1
+    """Number of object groups in the chunk-major key layout; the per-chunk key
+    count is ``world_size * num_object_groups``. ``1`` (the default) is the
+    single full-attention group / object-group-separation-disabled case."""
```
ApostaC: nit: we don't need this.

(I actually feel like the comments above and below look annoying as well... But let's leave it there and maybe fix them in the future)

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -266,20 +299,31 @@ def lookup(
         session.set_tokens(list(key.token_ids))
         session.lookup_ipc_key = key
 
-        obj_keys = ipc_key_to_object_keys(key, chunk_hashes, [0])[0]
+        # Resolve the registered object groups and lay keys out chunk-major
+        # (chunk -> object group -> kv_rank), so each chunk's keys are contiguous
+        # and a chunk is a full-attention hit only when every group is present.
+        # An unregistered model defaults to a single full-attention group. The
+        # per-group windows are propagated to the prefetch policy; the current
+        # full-attention prefix logic does not consume them yet.
```
ApostaC: nit: simplify or remove the comment here.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -399,7 +444,11 @@ def free_lookup_locks(
         )
         if not chunk_hashes:
             return
-        obj_keys = ipc_key_to_object_keys(key, chunk_hashes, [0])[0]
+        # Release across every object group: a hybrid model locks keys in
+        # multiple groups at lookup, so releasing only group 0 would leak the
+        # rest. ``finish_read_prefetched`` is a no-op for keys that were not
+        # locked, so over-resolving is safe.
+        obj_keys = self._object_keys_all_groups(key, chunk_hashes)
```
ApostaC: Potential problem: Over-resolving is not safe because it may unlock the keys that are locked by other requests.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -437,7 +486,10 @@ def end_session(self, request_id: str) -> None:
             return
 
         chunk_hashes = [TokenHasher.hash_to_bytes(h) for h in session.get_hashes(0)]
-        obj_keys = ipc_key_to_object_keys(session.lookup_ipc_key, chunk_hashes, [0])[0]
+        # Touch every object group's keys so their LRU recency stays aligned
+        # across groups (a chunk is a model-wide hit only if all its groups are
+        # resident, so they should age together).
```
ApostaC: nit: we can remove this as well

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -448,6 +500,71 @@ def end_session(self, request_id: str) -> None:
...
+    ) -> list[ObjectKey]:
+        """Build a chunk-major flat key list across all object groups.
+
+        Resolves keys per object group (each group's list is chunk-major /
+        rank-minor) and interleaves them so the layout is
+        ``chunk -> object group -> kv_rank``: every chunk's ``num_groups * R``
+        keys are contiguous. A contiguous leading-ones prefix over the flat list
+        then equals the full-attention model-wide hit (a chunk counts only when
+        all its groups are present). With a single group this is byte-identical
+        to the single-group layout.
```
ApostaC: This part is a bit hard to understand. Maybe replace this with an illustrative example? WDYT?

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -40,6 +40,10 @@ class _LayoutDescEntry:
 
     layout_desc: MemoryLayoutDesc
     ref_count: int
+    object_group_windows: tuple[int, ...] = ()
+    """Per-object-group cross-chunk sliding-window sizes (in chunks), in
+    object-group order, used by the hybrid prefix-cache lookup. ``-1`` per entry
+    means full attention; an empty tuple means a single full-attention group."""
```
ApostaC: 1. Remove `used by the hybrid prefix-cache lookup.` since if other modules are going to use this in the future, the docstring will be inconsistent.
2. For a single full attention group, I do see some code passing `(-1, )` instead of an empty tuple. How about let's use `(-1, )` because it does not need extra if logic to check?

lmcache/v1/kv_layer_groups.py
```diff
@@ -24,6 +25,33 @@
...
+# default, which keeps every kernel group in a single full-attention object
+# group (the pre-hybrid behavior). This is an MP-mode (cache-server) feature
+# flag; it follows the ``ENV_MP_* = "LMCACHE_MP_*"`` env-var convention used
+# elsewhere in the multiprocess path (e.g. ``ENV_MP_TRANSFER_MODE``). Read at
+# detection time so it can be set per process / overridden in tests.
+ENV_MP_SEPARATE_OBJECT_GROUPS = "LMCACHE_MP_SEPARATE_OBJECT_GROUPS"
+
+
+def separate_object_groups_enabled() -> bool:
```
ApostaC: How about we use a CLI args to control the behavior? We don't use any environment variables in MP mode AFAIK.

lmcache/v1/kv_layer_groups.py
```diff
@@ -556,40 +581,50 @@ def calculate_num_blocks(self, kernel_group_idx: int, num_tokens: int) -> int:
...
+
+        When object-group separation is disabled (the default; see
+        :func:`separate_object_groups_enabled`), every kernel group shares a
+        single full-attention object group -- the pre-hybrid behavior. When
+        enabled, kernel groups are bucketed by cross-chunk sliding-window size:
+        non-sliding-window groups (``sw_size_tokens == -1``) share one
+        full-attention object group, and each distinct window size (rounded up
+        to chunks) forms its own object group, so the kernel groups inside one
+        object share a single prefix-matching rule. A non-hybrid model yields a
+        single full-attention object group either way.
```
ApostaC: Can be simplified to 
```
Put all layers into a single object group if object-group separation is disabled.
Otherwise, it will group the kernel groups based on sliding window sizes measured in number of chunks. 
```

ApostaC (commented):
One potential correctness problem + quite a few docstring nit comments. PTAL

lmcache/v1/kv_layer_groups.py
KuntaiDu: Sure. But just FYI -- MP mode already uses env var:
https://github.com/LMCache/LMCache/blob/3fcbc3b53f4979b34b1f3abd2bb5b24182266464/lmcache/v1/multiprocess/transfer_context/worker_transfer.py#L42

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -738,24 +582,17 @@ def _start_lookup_phase(
         layout_desc: MemoryLayoutDesc,
         extra_count: int = 0,
         policy: TrimPolicy = TrimPolicy.PREFIX,
+        group_windows: tuple[int, ...] = (),
```
KuntaiDu: After second thought, I prefer this one to be called 
attn_desc: AttentionDesc
Code placement wisse it should be similar to MemoryLayoutDesc

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -645,42 +534,39 @@ def register_kv_cache(
...
-            self._cache_contexts[instance_id] = ContextEntry(
-                cache_context=cache_context,
-                model_name=model_name,
-                world_size=world_size,
-                last_seen=now,
-                has_liveness_signal=False,
-            )
+        kv_groups_manager = cache_context.kv_layer_groups_manager
+        object_group_windows = tuple(
+            kv_groups_manager.get_sw_size_chunks(object_group_id)
```
KuntaiDu: kv_groups_manager.get_attn_desk(object_group_id)

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -40,6 +40,10 @@ class _LayoutDescEntry:
 
     layout_desc: MemoryLayoutDesc
     ref_count: int
+    object_group_windows: tuple[int, ...] = (-1,)
+    """Per-object-group cross-chunk sliding-window sizes (in chunks), in
+    object-group order. ``-1`` means a full-attention group; the default
+    ``(-1,)`` is a single full-attention group."""
 
 
 class LayoutDescRegistry:
```
KuntaiDu, line 50: After second thought, this should be called 
"AttentionDescRegistry"

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -62,6 +66,7 @@ def register(
         model_name: str,
         world_size: int,
         layout_desc: MemoryLayoutDesc,
+        object_group_windows: tuple[int, ...] = (-1,),
```
KuntaiDu: In Jenga's paper, there are two types of heterogeneity:
shape and dependency
so what is a good name for this one? @CLAUDE let's discuss. I guess it should be something like
sliding_window_registry or prefix_dependency_registry? Let's discuss.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -228,6 +217,7 @@ def __init__(
                 MemoryLayoutDesc,
                 int,
                 TrimPolicy,
+                tuple[AttnWindowDesc, ...],
```
KuntaiDu: ??? Why tuple[AttnWindowDesc] not just AttnWindowDesc? As I said the impl ideally should be similar to MemoryLayoutDesc

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -67,6 +68,34 @@ def compute_extra_count(
...
+def _get_prefix_hit_length(
+    found_prefix_len: int,
+    world_size: int,
+    num_object_groups: int,
+) -> int:
+    """Return the prefix hit length in chunks.
+
+    The chunk-major key layout packs ``world_size * num_object_groups`` keys per
+    chunk, so a fully-present prefix of ``found_prefix_len`` keys hits
+    ``found_prefix_len // (world_size * num_object_groups)`` chunks.
```
ApostaC, line 80: Maybe have an example here.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -399,7 +439,9 @@ def free_lookup_locks(
         )
         if not chunk_hashes:
             return
-        obj_keys = ipc_key_to_object_keys(key, chunk_hashes, [0])[0]
+        # Release across every object group, mirroring lookup, which locks keys
+        # in every group; releasing only group 0 would leak the rest.
+        obj_keys = self._object_keys_all_groups(key, chunk_hashes)
```
ApostaC: This is correct for now (only for full attention models).
For SWA models, we need to skip "unlocking" the chunks that is outside the sliding window.

Kuntai: we can fully skip the swa layers -- this needs to be double checked -- e.g., sw length = 512, LMCache hit 1024, vLLM hit 768, then chunk 512-768 may be leaked

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -448,6 +490,70 @@ def end_session(self, request_id: str) -> None:
...
+    def _object_keys_all_groups(
+        self,
+        key: IPCCacheServerKey,
+        chunk_hashes: list[bytes],
+    ) -> list[ObjectKey]:
+        """Resolve a flat key list spanning every registered object group.
+
+        Used by the lock-release and touch paths, which must cover all object
+        groups (a hybrid model locks/stores keys in multiple groups). Key order
+        is irrelevant for those callers, so groups are flattened group-major.
```
ApostaC: Keep only the first sentence here.

lmcache/v1/multiprocess/config.py
```diff
@@ -340,27 +263,13 @@ def add_mp_server_args(
...
-        "retrieve failure, retain the gapped prefix so post-gap chunks stay "
-        "L1-resident instead of truncating at the gap. No effect otherwise.",
+        "--no-separate-object-groups",
+        dest="separate_object_groups",
+        action="store_false",
+        help="Keep all kernel groups in a single full-attention object group "
+        "instead of splitting them per sliding-window size at KV-cache "
+        "registration. Object-group separation is on by default (for hybrid "
+        "models).",
     )
```
ApostaC, line 353: `action=argparse.BooleanOptionalAction`

See `--l1-use-lazy`

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -136,15 +170,19 @@ class MPCacheServerContext:
...
+        separate_object_groups: Whether to split kernel groups into one object
+            group per sliding-window size at KV-cache registration.
     """
 
     def __init__(
         self,
         storage_manager_config: StorageManagerConfig,
         chunk_size: int = 256,
         hash_algorithm: str = "blake3",
+        separate_object_groups: bool = False,
```
ApostaC: We can create a "good second issue" -- refactor this to only pass the `MPServerConfig`.

tests/v1/multiprocess/test_query_lookup_hits.py
```diff
@@ -209,3 +223,56 @@ def test_server_handler_registered():
...
+
+def test_chunk_major_object_keys_orders_chunk_then_group_then_rank():
+    """Keys are laid out chunk -> object group -> kv_rank, so each chunk's keys
+    are contiguous (the property that makes a leading-ones prefix equal to the
+    full-attention model-wide hit)."""
+    module = LookupModule(MagicMock())
+    key = _lookup_key(world_size=2)
+    chunk_hashes = [b"c0", b"c1"]
+
+    keys = module._chunk_major_object_keys(key, chunk_hashes, 2)
```
ApostaC: Avoid calling private methods in unit tests.

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -448,6 +490,70 @@ def end_session(self, request_id: str) -> None:
...
+        """Build a chunk-major flat key list across all object groups.
+
+        The keys are ordered ``chunk -> object group -> kv_rank`` so that all
+        keys belonging to one chunk are contiguous. A leading-ones prefix over
+        the flat list then maps directly to a whole-chunk hit count.
+
+        Example (2 chunks ``c0,c1``; 2 groups ``g0,g1``; 2 kv_ranks ``r0,r1``)::
+
+            [c0g0r0, c0g0r1, c0g1r0, c0g1r1,   # chunk 0: all groups, all ranks
+             c1g0r0, c1g0r1, c1g1r0, c1g1r1]   # chunk 1: ...
```
yoo-kumaneko, line 563: love this illustration. it really makes it easier to understand.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -448,6 +490,70 @@ def end_session(self, request_id: str) -> None:
...
+            for group_keys in per_group:
+                obj_keys.extend(group_keys[lo:hi])
+        return obj_keys
+
+    def _object_keys_all_groups(
+        self,
+        key: IPCCacheServerKey,
+        chunk_hashes: list[bytes],
+    ) -> list[ObjectKey]:
+        """Resolve a flat key list spanning every registered object group.
```
yoo-kumaneko: would be clearer if we add a similar illustration like the above.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -448,6 +490,70 @@ def end_session(self, request_id: str) -> None:
...
+
+        Args:
+            key: The IPC key (model/world/worker, salt).
+            chunk_hashes: Chunk hashes to resolve keys for.
+            num_groups: Number of object groups to resolve.
+
+        Returns:
+            The chunk-major flattened list of object keys.
+        """
+        per_group = ipc_key_to_object_keys(key, chunk_hashes, list(range(num_groups)))
```
yoo-kumaneko, line 575: why do we need the `num_groups` parameter? As I noticed that the method below it, `_object_keys_all_groups`, uses `attn_desc.num_object_groups` to get the groups.

lmcache/v1/multiprocess/config.py
```diff
@@ -45,11 +45,10 @@ class MPServerConfig:
     ('default' for standard prefix caching, 'blend' when cacheblend is enabled).
     """
 
-    enable_segmented_prefix: bool = False
-    """CacheBlend only (engine_type='blend'): on a mid-prefix L2 retrieve
-    failure, retain the gapped contiguous prefix so the post-gap chunks stay
-    L1-resident (served by the sparse leg as L1 hits, the hole recomputed)
-    instead of truncating the prefix at the gap. No effect for other engines."""
+    separate_object_groups: bool = True
```
yoo-kumaneko, line 48: here the default is `True` while in other places like `cache_context.py` and `kv_layer_groups.py` the default value for the parameter of the method is `False`. Is it set up this way on purpose? or shall we make it the same everywhere?

lmcache/v1/multiprocess/modules/lookup.py
yoo-kumaneko: if I understand it correctly, this method returns a list of "group major" object keys.

yoo-kumaneko (approved):
had a quick review. this PR passes the AttnWindowDesc through the modules and updates the interfaces. It also skips the out-of-window SWA chunks in the transfer. The trimming logic in prefetch is left untouched for follow-up PRs. overall LGTM.

lmcache/v1/multiprocess/config.py
KuntaiDu: The default should be always True. Let me align.

lmcache/v1/multiprocess/modules/lookup.py
KuntaiDu: Yes. And we should remove this --- I checked the code and it seems that we can directly reuse the chunk major object key helper function to replace this one without side effect.

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -478,8 +494,14 @@ def end_session(self, request_id: str) -> None:
             )
             return
 
+        ipc_key = session.lookup_ipc_key
         chunk_hashes = [TokenHasher.hash_to_bytes(h) for h in session.get_hashes(0)]
-        obj_keys = self._object_keys_all_groups(session.lookup_ipc_key, chunk_hashes)
+        attn_desc = self._ctx.layout_desc_registry.find_attn_desc(
+            ipc_key.model_name, ipc_key.world_size
+        )
+        obj_keys = self._chunk_major_object_keys(
```
KuntaiDu: Why not just `obj_keys = self._chunk_major_obj_keys(ipc_key, chunk_hashes)`?

lmcache/v1/multiprocess/modules/lookup.py
KuntaiDu: done

lmcache/v1/multiprocess/config.py
KuntaiDu: fixed.

lmcache/v1/multiprocess/modules/lookup.py
KuntaiDu: Both works, but let me align.

lmcache/v1/multiprocess/modules/lookup.py
ApostaC: 1

ApostaC (approved):
User doc + small comments.
Otherwise LGTM!

PR #3778, opened 2026-06-19
[Bugfix] Release P2P transfer-channel client when adapter is removed

ApostaC:

When a P2P L2 adapter is removed, `P2PL2Adapter.close()` tore down its MQ client and event fds but never released the transfer-channel client it had obtained from the shared `NixlTransferChannelContext`. That client stayed registered in the context, leaking its nixl handles (the prepped remote dlist and any in-flight transfer handles) on every adapter removal.

This PR adds a `remove_transfer_channel_client(peer_advertise_url)` method to the `TransferChannelContext` contract (implemented for the nixl backend) and calls it from `P2PL2Adapter.close()`, so the peer's client is closed and dropped from the context when the adapter no longer needs it.

- The context keys clients by peer URL with no refcounting, matching the existing one-reader-per-peer assumption. If two adapters ever target the same peer, the first `close()` would drop the shared client — out of scope for this fix.
- `P2PL2Adapter.close()` is now idempotent (guarded by the existing closed flag).
- Removing an unknown/already-removed peer is a no-op.

**If applicable**:

### DIFF
--- a/docs/design/v1/distributed/l2_adapters/p2p_l2_adapter.md
+++ b/docs/design/v1/distributed/l2_adapters/p2p_l2_adapter.md
@@ -101,7 +101,8 @@ storage manager owns its lifecycle. The config carries the peer's two URLs:
 ## close()
 
 Sets a closed flag (later submits are inert), unregisters the lookup/load fds
-from the periodic notifier, closes the MQ client, and closes the three event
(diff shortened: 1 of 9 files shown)


sammshen (approved):
LGTM!

deng451e (approved):
LGTM!

PR #3790, opened 2026-06-21
feat(ops): support HND layouts in Python KV transfer fallback

zupengwang:

The Python fallback for `multi_layer_kv_transfer` previously rejected the following head-major KV layouts:

- `NL_X_TWO_NB_NH_BS_HS`
- `NL_X_NB_TWO_NH_BS_HS`

These layouts are supported by the CUDA implementation, but the fallback raised `NotImplementedError`, preventing non-CUDA transfer paths from handling head-major vLLM FlashAttention and FlashInfer buffers.

This PR:

- adds vectorized H2D and D2H transfers for both HND layouts
- preserves the physical head dimension when gathering and scattering slots
- validates block size, head size, page-buffer size, and hidden-size consistency
- documents the public fallback function contract
- adds regression tests for both layouts, both transfer directions, raw pointers, and invalid dimensions

Existing NHD, MLA, and FlashInfer fallback paths remain unchanged.

Validated remotely in a Linux CPU-only environment:

- Before the implementation, all four HND transfer cases failed with `NotImplementedError`
- `pytest tests/v1/test_python_ops_fallback.py -q`: `57 passed`
- `pre-commit run --all-files`: all hooks passed

**If applicable**:


zupengwang:
Hi maintainers, could someone review whether this Python fallback HND-layout support is wanted for the dev branch? It fills the non-CUDA fallback gap for the two head-major KV layouts already handled by the CUDA path, with regression coverage for both directions and invalid dimensions. DCO and the current Buildkite checks are passing.

PR #3793, opened 2026-06-21
fix(memory): cache ctypes ubyte-array types to stop heap-type leak

### LINKED ISSUE #3767: EngineCore RssAnon grows unbounded on long timed_trace replay (Qwen3-14B + LMCache v1 embedded), leading to host OOM-kill
**Summary**

  When replaying a long timed trace (Mooncake `toolagent_trace.jsonl`, ~23K
  requests) against vLLM + LMCache v1 in embedded mode (LMCacheConnectorV1,
  `kv_role: kv_both`), the vLLM EngineCore process's `RssAnon`
  (anonymous-page RSS) grows monotonically. The host eventually OOM-kills
  EngineCore. Running the same trace with LMCache disabled shows the same
  growth pattern but at a dramatically slower rate, suggesting the LMCache
  v1 connector / embedded integration is the dominant source of the
  anonymous memory growth.

  **Details**

  Environment:

  - **vLLM version:** 0.22.1rc1.dev38+gfd9e91d7e
  - **LMCache version:** v0.4.8rc1 (gf52b174c)
  - **Mode:** LMCache v1, embedded (`KV connector: LMCacheConnectorV1`, `kv_role: kv_both`)
  - **Model:** Qwen/Qwen3-14B
  - **GPU:** 1× NVIDIA L40S (TP=1)
  - **CPU:** Intel Xeon (GNR / Granite Rapids)
  - **Host RAM:** 256 GB DDR
  - **/dev/shm:** 126 GB
  - **OS:** Ubuntu 24.04.3 LTS (Noble Numbat)
  - **Kernel:** 6.16.0-16-generic #16 SMP PREEMPT_DYNAMIC

  Workload:

  Replaying the Mooncake `toolagent_trace.jsonl` via `vllm bench serve` in
  timed-trace mode:

  - Trace: https://github.com/kvcache-ai/Mooncake/blob/main/FAST25-release/traces/toolagent_trace.jsonl
  - `--num-prompts 23608`
  - `--self-timed --timed-trace-chunk-hash-size 512 --timed-trace-sec-multiplier 0.001`
  - `--ignore-eos`

  Observation:

  The vLLM EngineCore process's `RssAnon` grows monotonically over the
  run. `Shmem` is stable (LMCache local CPU pool is capped and behaves as
  expected); the growth is anonymous memory in EngineCore. The host
  eventually OOM-kills EngineCore.

  Bisect — same trace **without LMCache**:

  With LMCache disabled (`--kv-transfer-config` removed, all other flags
  identical), RssAnon still grows but **much more slowly**: the run has
  reached ~1600 requests over 3 hours and is still progressing, vs. the
  LMCache-enabled run terminating around request 200.

  This suggests the LMCache v1 connector / embedded integration
  is the dominant source of the anonymous memory growth (likely pinned
  host buffers used for GPU↔CPU KV staging, since the leak is anon and
  not shm, and shm/Shmem growth is bounded as expected by
  `LMCACHE_MAX_LOCAL_CPU_SIZE`).

  What was already tried:

  - `LMCACHE_MAX_LOCAL_CPU_SIZE=200` and `=100` — same crash, same
    unbounded RssAnon growth pattern.
  - `--no-enable-prefix-caching` is already set in the launch command.

  **Steps / Reproduction (if applicable)**

  Launch script:

  ```bash
  export PYTHONHASHSEED=0
  export LMCACHE_CHUNK_SIZE=256
  export LMCACHE_LOCAL_CPU=True
  export LMCACHE_REMOTE_URL="lm://localhost:8100"
  export LMCACHE_REMOTE_SERDE=naive
  export LMCACHE_LOOKUP_TIMEOUT_MS=3600000
  export LMCACHE_BLOCKING_TIMEOUT_SECS=3630
  export LMCACHE_MAX_LOCAL_CPU_SIZE=100
  export LMCACHE_NUM_CONNECTIONS=1
  export LMCACHE_CACHE_POLICY=LFU

  # LMCache server
  python -m lmcache.v1.server 0.0.0.0 8100 &

  # vLLM server
  vllm serve Qwen/Qwen3-14B \
      --port 8000 \
      --kv-transfer-config '{"kv_connector": "LMCacheConnectorV1", "kv_role": "kv_both"}' \
      --no-enable-prefix-caching &

  # Trace replay
  vllm bench serve \
    --backend openai \
    --model Qwen/Qwen3-14B \
    --base-url http://localhost:8000 \
    --endpoint /v1/completions \
    --dataset-name timed_trace \
    --dataset-path /path/to/toolagent_trace.jsonl \
    --num-prompts 23608 \
    --ignore-eos \
    --self-timed \
    --timed-trace-chunk-hash-size 512 \
    --timed-trace-sec-multiplier 0.001
  ```

  1. Start the LMCache v1 server.
  2. Start vLLM with the LMCacheConnectorV1 kv-transfer config.
  3. Run vllm bench serve in timed-trace mode against the Mooncake
  toolagent_trace.jsonl with --num-prompts 23608.
  4. Observe EngineCore RssAnon (e.g. grep RssAnon /proc/<pid>/status)
  growing monotonically until the host OOM-kills the process around
  request ~200.

  **Expected Outcome / Goal**

  EngineCore-side anonymous memory usage should be bounded over the
  lifetime of a long trace replay, so that long-running workloads do not
  trigger host OOM. Specifically, pinned host buffers used by the LMCache
  v1 connector for GPU↔CPU KV staging should be allocated from a bounded
  pool rather than growing per-transfer.

  **Actual Outcome (if applicable)**

  EngineCore's RssAnon grows monotonically and without plateau during
  the run, eventually triggering a kernel OOM kill of the EngineCore
  process. Kernel OOM log from one such crash:

  Out of memory: Killed process 142685 (VLLM::EngineCor)
  total-vm:256016404kB, anon-rss:1815020kB, file-rss:107600kB,
  shmem-rss:122981084kB, UID:0
  oom_score_adj:0
  constraint=CONSTRAINT_NONE, global_oom

  The LMCache-enabled run terminates around request 200, while the same
  trace with LMCache disabled has reached ~1600 requests over 3 hours and
  is still progressing.

  **Additional Context**

  Asks:

  1. Is unbounded EngineCore-side anonymous memory growth a known issue
  with LMCacheConnectorV1 in v0.4.8rc1?
  2. Are there pinned-host-buffer allocation paths in the v1 connector
  that allocate per-transfer rather than from a bounded pool?
  3. Recommended workaround / config to bound this growth, short of
  disabling LMCache entirely?

THINKER-ONLY:

Fixes #3767.

`TensorMemoryObj.byte_array` builds a fresh `(ctypes.c_ubyte * num_bytes)`
array type on every call. `ctypes` does **not** cache these array types
(unlike pointer types, which use `_pointer_type_cache`); each `*` call goes
through `PyCArrayType_from_ctype` and produces a brand-new heap type whose
metadata is held by the type system forever. The remote-backend put/get path
accesses `byte_array` once per chunk transfer (`lm_connector.py:72,118`,
`memory_management.py:838`), so every remote put/get permanently leaks
~1-2 kB of heap-type metadata.

Over a long timed-trace replay this manifests as the symptom in #3767:
- `Shmem` stays bounded (the up-front pinned CPU pool sized by
  `LMCACHE_MAX_LOCAL_CPU_SIZE`)
- `RssAnon` grows monotonically
- The host eventually OOM-kills the EngineCore process

The fix caches the array type per length in a module-level dict and reuses
it on every access. The cached `from_address(addr)` instance never owns the
underlying buffer, only the metadata (length, item-type), so caching by `N`
is sound.

Reproduction harness (not part of this PR; available on request) drives
`LMCacheEngine` directly with `LMCACHE_CACHE_POLICY=LFU`,
`remote_url="lm://..."`, `remote_serde="naive"`, 1 GB pool, 768 tokens/iter,
2000 iters. Steady state begins around iter 600 once the CPU pool is full
and ref counts have settled.

| Run | RssAnon iter 800 → iter 1999 | steady-state slope |
|----|----|----|
| baseline (LFU, before fix)  | +14576 kB | ~12 kB / iter |
| in-tree fix (LFU, after fix) | **+1736 kB** | **~1.5 kB / iter** (~88% reduction) |

The residual ~1.5 kB / iter is unrelated background churn (transient
`threading.Condition` objects, prometheus value buffers); it does not grow
without bound and is independent of LMCache code paths.

I considered three alternatives before settling on the cache:
1. `memoryview(self.raw_data)` directly — clean but `raw_data` may carry
   alignment padding (e.g. from `batched_allocate`); the existing comment
   in `byte_array` explicitly notes the size has to be the logical
   `get_size()`, not `raw_data` length.
2. `memoryview(self.raw_data.contiguous().numpy())` — would force CPU-only
   semantics and pay a numpy dtype-view cost on the hot put/get path.
3. Cache the array type per length (this PR) — minimal, preserves existing
   semantics, ~5 lines of net change.

Two adjacent issues found while bisecting #3767, neither of which is the
root cause and both worth filing as separate PRs:

- `LRUCachePolicy.chunk_hash_to_init_timestamp` never shrinks
  (`update_on_force_evict` is `pass`); affects users running
  `LMCACHE_CACHE_POLICY=LRU`. The reporter uses `LFU`, so this is unrelated
  to #3767.
- `LFUCachePolicy.get_evict_candidates` removes keys from `key_to_freq` /
  `freq_to_keys` before the caller commits to evicting them. The
  `continue` / `break` paths in
  `LocalCPUBackend.batched_allocate` (lines 800, 813) can leave the
  `hot_cache` entry in place while LFU state is gone, orphaning the
  underlying `MemoryObj`. Only triggers on the layerwise / batched
  allocation path; the non-layerwise put path used by the issue reporter
  does not hit it.

**If applicable**:

### DIFF
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -4,7 +4,7 @@ from collections import deque
 from contextlib import nullcontext
 from dataclasses import dataclass
 from enum import Enum, auto
-from functools import wraps
+from functools import cache, wraps
 from typing import Any, List, Optional, Tuple, Union
 import abc
 import ctypes
@@ -41,11 +41,11 @@ logger = init_logger(__name__)
 #
 # Caching by length is safe: the `from_address(addr)` instance never owns the
 # underlying buffer, only the metadata (length, item-type), and that metadata
-# depends solely on `N`.
-_UBYTE_ARRAY_TYPE_CACHE: dict[int, type] = {}
-
-
-def _get_cached_ubyte_array_type(num_bytes: int) -> type:
+# depends solely on `N`. ``functools.cache`` provides a thread-safe unbounded
+# memoization primitive, so concurrent first-time accesses for the same `N`
+# cannot race to create distinct heap types.
+@cache
+def _get_cached_ubyte_array_type(num_bytes: int) -> type[ctypes.Array[ctypes.c_ubyte]]:
     """Return a cached ``ctypes.c_ubyte * num_bytes`` array type.
 
     Args:
@@ -55,11 +55,7 @@ def _get_cached_ubyte_array_type(num_bytes: int) -> type:
         The cached ``ctypes.Array`` subclass for the given length. Subsequent
         calls with the same ``num_bytes`` return the same type object.
     """
-    arr_type = _UBYTE_ARRAY_TYPE_CACHE.get(num_bytes)
-    if arr_type is None:
-        arr_type = ctypes.c_ubyte * num_bytes
-        _UBYTE_ARRAY_TYPE_CACHE[num_bytes] = arr_type
-    return arr_type
+    return ctypes.c_ubyte * num_bytes
 
 
 # Helper functions for thread safety
@@ -875,9 +871,7 @@ class TensorMemoryObj(MemoryObj):
         # https://github.com/LMCache/LMCache/issues/3767.
         arr_type = _get_cached_ubyte_array_type(num_bytes)
         ubyte_ptr = ctypes.cast(ptr, ctypes.POINTER(ctypes.c_ubyte))
-        byte_array = arr_type.from_address(
-            ctypes.addressof(ubyte_ptr.contents)
-        )
+        byte_array = arr_type.from_address(ctypes.addressof(ubyte_ptr.contents))
         return memoryview(byte_array)
 
     @property
--- a/lmcache/v1/memory_management.py
+++ b/lmcache/v1/memory_management.py
@@ -29,6 +29,39 @@ import lmcache.c_ops as lmc_ops
 logger = init_logger(__name__)
 
 
+# Cache for ctypes ubyte-array types keyed by length.
+#
+# ctypes does not cache `(c_ubyte * N)` array types -- each call to the `*`
+# operator builds a fresh heap type via PyCArrayType_from_ctype. The heap
+# type metadata stays alive forever (held by the type system), so calling
+# `(ctypes.c_ubyte * N).from_address(...)` on every TensorMemoryObj.byte_array
+# access leaks ~1-2 kB per call. Under long-running remote-backend put/get
+# workloads this is the dominant source of monotonic anonymous-memory growth
+# (see https://github.com/LMCache/LMCache/issues/3767).
+#
+# Caching by length is safe: the `from_address(addr)` instance never owns the
+# underlying buffer, only the metadata (length, item-type), and that metadata
+# depends solely on `N`.
+_UBYTE_ARRAY_TYPE_CACHE: dict[int, type] = {}
+
+
+def _get_cached_ubyte_array_type(num_bytes: int) -> type:
+    """Return a cached ``ctypes.c_ubyte * num_bytes`` array type.
+
+    Args:
+        num_bytes: The length of the array type in bytes.
+
+    Returns:
+        The cached ``ctypes.Array`` subclass for the given length. Subsequent
+        calls with the same ``num_bytes`` return the same type object.
+    """
+    arr_type = _UBYTE_ARRAY_TYPE_CACHE.get(num_bytes)
+    if arr_type is None:
+        arr_type = ctypes.c_ubyte * num_bytes
+        _UBYTE_ARRAY_TYPE_CACHE[num_bytes] = arr_type
+    return arr_type
+
+
 # Helper functions for thread safety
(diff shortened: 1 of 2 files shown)


DongDongJu (approved):
Hello @THINKER-ONLY,
I didnt know about this problem.
Thanks for the fix. LGTM

sammshen (approved):
LGTM!

PR #3799, opened 2026-06-22
Read allocator sizes from a single locked snapshot in memcheck/usage reporting

### LINKED ISSUE #3768: TensorMemoryAllocator.memcheck() can race with lazy allocator address-space growth

nevasini1:
## Summary

Closes #3768.

`TensorMemoryAllocator.memcheck()` read free size, allocated size, and heap size through **separate** `AddressManager` calls:

```python
total_free_size = self.address_manager.get_free_size()
if (total_free_size + self.address_manager.total_allocated_size
        != self.address_manager.get_heap_size()):
    logger.error("Memory allocator size is inconsistent")
```

With lazy allocation, `sbrk()` can grow the address space **between** those reads, so the check compares a stale free size against a newer heap size and incorrectly logs `Memory allocator size is inconsistent` / returns `False` on a perfectly valid allocator. The same split-read pattern affected L1 memory-usage reporting, which derived `used`/`total` from separate address-manager reads.

## Fix

Add `AddressManager.get_size_snapshot()`, which captures `(free, allocated, heap)` atomically under the allocator's existing `_lock` and returns an `AddressSizeSnapshot`. All three values then describe one coherent state, so `free + allocated == heap` always holds for a consistent allocator regardless of concurrent `sbrk()` growth.

Routed through the snapshot:
- `TensorMemoryAllocator.memcheck`
- `gds_l1_memory_manager`: `memcheck` + `get_memory_usage`
- `l1_memory_manager`: `get_memory_usage`

All four sites operate on an `AddressManager` (verified: `TensorMemoryAllocator.address_manager`, GDS L1's `_address_manager`, and both branches of L1's `get_address_manager` — `pin_allocator.address_manager` and `LazyMemoryAllocator.get_address_manager()` — are `AddressManager` instances).

## Testing

Adds `TestAddressSizeSnapshot` to `tests/v1/test_address_manager.py`:
- snapshot fields match the individual accessors on a quiescent allocator;
- `free + allocated == heap` holds across allocate / free / sbrk;
- the invariant never breaks while a background thread grows the space via `sbrk()` (4 observer threads × 20k snapshots).

`ruff check`, `ruff format --check`, and `py_compile` pass on all changed files. (The allocator module imports `torch`, so the suite runs in CI; I additionally validated the snapshot arithmetic and concurrency guarantee with a standalone replica of the size bookkeeping.)

### DIFF
--- a/lmcache/v1/distributed/memory_manager/gds_l1_memory_manager.py
+++ b/lmcache/v1/distributed/memory_manager/gds_l1_memory_manager.py
@@ -107,9 +107,10 @@ class GDSL1MemoryManager:
 
     def get_memory_usage(self) -> tuple[int, int]:
         """Return ``(used_bytes, total_bytes)`` of the slab."""
-        free_size = self._address_manager.get_free_size()
-        total_size = self._address_manager.get_heap_size()
-        return total_size - free_size, total_size
+        # Single locked snapshot so used/total can't be computed from reads
+        # taken across a concurrent sbrk() growth (see #3768).
+        snapshot = self._address_manager.get_size_snapshot()
+        return snapshot.allocated, snapshot.heap
 
     def get_l1_memory_desc(self) -> Optional[L1MemoryDesc]:
         """Return ``None``: the GDS L1 medium is the slab file, not a buffer.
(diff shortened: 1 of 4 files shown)


DongDongJu (approved):
Hello @nevasini1,
Thanks for the fix. LGTM.

chunxiaozheng:
@nevasini1 There are some conflicts, could you help fix it? And then I will take a look.

PR #3797, opened 2026-06-22
core: convert f-string log calls in system_detection.py to %-format

nevasini1:
## Summary

Closes #3794.

Converts the three f-string `logger` calls in `lmcache/v1/system_detection.py` to lazy `%`-format so interpolation only runs when the record is actually emitted — matching the convention already used elsewhere (e.g. `lmcache/v1/cache_engine.py`).

- `logger.info(f"{system} system available memory: {available_gb:.2f} GB")` → `logger.info("%s system available memory: %.2f GB", system, available_gb)`
- `logger.warning(f"Failed to get system available memory using psutil: {e}")` → `%s` for the exception
- `logger.warning(f"Failed to auto read NUMA mapping from system: {e}")` → `%s` for the exception

Type-aware specifiers per the convention in #3785: `%.2f` for the float memory value, `%s` for `system` (str) and exceptions.

## Testing

- `ruff format` / `ruff check` pass.
- `python -m py_compile` passes.
- No remaining f-string `logger` calls in the file.

Issue #3794, opened 2026-06-21
[good-first-issue] core: convert f-string log calls in lmcache/v1/system_detection.py to %-format


sat-v4:
As title suggests.

PR #3801, opened 2026-06-22
docs: codify the log format-specifier convention in coding standards

### LINKED ISSUE #3785: [good-first-issue] core: `instance_id` type is `str` in some modules, `int` in others — format specifiers can't be applied mechanically
Thanks for taking the time to read this! First contribution here, so feedback is very welcome.

## The problem

Picking `%s` vs `%d` in log calls seems simple until `instance_id` shows up. In some files it's a `str`, in others it's an `int`. When converting f-strings to %-format, each variable needs a quick type check — can't just do find-and-replace.

### `instance_id` is two different types

**As `str`** (dict key, e.g. lmcache/v1/config_base.py:258, self.lmcache_instance_id = f"{config_name.lower()}_{uuid.uuid4().hex}"):
```
lmcache/v1/cache_engine.py:2081     instance_id: str
lmcache/v1/cache_engine.py:2138     destroy(cls, instance_id: str)
lmcache/v1/cache_controller/controllers/full_sync_tracker.py:53    instance_id: str
lmcache/v1/storage_backend/connector/bigtable_config.py:11          instance_id: str
lmcache/v1/cache_controller/full_sync_sender.py:105                 def instance_id(self) -> str
```

**As `int`** (GPU index / worker number):
```
lmcache/cli/commands/bench/server_bench/helpers.py:320    instance_id: int = 0
lmcache/v1/multiprocess/modules/blend.py:460              logger.info("...instance_id %d", instance_id)
lmcache/v1/multiprocess/modules/blend_v3.py:521,534       logger.info("...instance %d", instance_id)
lmcache/v1/multiprocess/modules/engine_driven_transfer.py:408  logger.info("...instance ID %d", instance_id)
lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py:706 logger.info("...GPU ID %d", instance_id)
```

### `worker_id` also has some format drift

`lmcache/v1/metadata.py:36` says `worker_id: int`, but a few places use `%s` for it (e.g. `lmcache/v1/cache_engine.py:157` before the fix in #3424). Most of the multiprocess modules already use `%d` correctly — just a few spots drifted.

## The rule that seems to fit the codebase

Based on existing patterns across the repo (`adapter_id` → `%d`, `task_id` → `%d`, `req_id` → `%s`):

- `int` (counts, numeric IDs) → `%d`
- Everything else (str, bool, objects, exceptions) → `%s`

For bools specifically: `%d` prints `1`/`0`, so `%s` was used to get `True`/`False` — but couldn't find an existing LMCache convention for this, so please correct if wrong.

## Suggestion

Two possible directions, whichever the maintainers prefer:

**A) Lightweight**: Add a few lines to `docs/coding_standards.md` about the format specifier convention. Something like "use `%d` for `int`, `%s` for everything else — check type annotations, not neighboring log calls."

**B) Structural**: Unify `instance_id`'s type. The multiprocess modules using `instance_id` as `int` (GPU index) could maybe be renamed to `gpu_id` or similar, keeping `instance_id` as `str` everywhere.

## Questions

1. Is "`%d` for `int`, `%s` for everything else" the right convention for LMCache? If so, happy to add it to the coding standards doc.
2. Worth opening follow-up issues to fix the remaining `%s`-on-int drift? Or better to leave it and just fix cases as they come up?
3. For the `instance_id` type split — intentional by layer, or something that could be cleaned up?

---

This came up while fixing the remaining f-string log calls in `cache_engine.py` (following up on #3424, which was closed but a handful of f-strings had been missed). Pre-commit is green (ruff, isort, codespell, clang-format, mypy), tests pass (9/9, 57 skipped — no GPU in local env). Nothing urgent here, just wanted to surface the inconsistency in case it's useful.

Thanks again to the maintainers for all the work on LMCache, and for the onboarding-2026 effort — it really helped to have a clear path for a first issue (❁´ω`❁).

Refs: #3424, #3372, #3784

nevasini1:
## Summary

Closes #3785 (Direction A — the lightweight option proposed in the issue).

The issue notes that converting f-string log calls to `%`-format isn't a mechanical find-and-replace because the right specifier depends on the variable's type — and some names (`instance_id`) are `str` in most modules but `int` (GPU index) in a few multiprocess modules. It asks whether "`%d` for `int`, `%s` for everything else" is the right convention and offers to document it.

This PR documents exactly that, in `docs/coding_standards.md` §7.4 (Logging):

- Prefer lazy `%`-formatting over f-strings in log calls.
- Choose the specifier from the variable's **type annotation**, not neighboring calls: `%d` for `int`, `%s` for everything else (`str`/`bool`/objects/exceptions), `%f`/`%.2f` for floats.
- Calls out the `instance_id` str-vs-int ambiguity and the bool `%d`→`1`/`0` vs `%s`→`True`/`False` gotcha.

This gives the f-string→%-format conversions (e.g. #3794, my #3797) a single convention to follow.

## Testing

Docs-only change. `codespell --toml pyproject.toml docs/coding_standards.md` passes.

If maintainers prefer Direction B (unifying `instance_id`'s type) instead, happy to close this — but the convention is useful either way.

### DIFF
--- a/docs/coding_standards.md
+++ b/docs/coding_standards.md
@@ -253,6 +253,12 @@ C++/CUDA files use clang-format (Google style, 80-col). Rust code uses `cargo fm
 - Use the project's `init_logger(__name__)` from `lmcache.logging`, not bare `logging.getLogger()`.
 - Operational logs (store/retrieve activity, background task progress) should be at `DEBUG` level, not `INFO`.
 - Log levels must be appropriate: do not use `WARNING` for conditions that are expected during normal concurrent operation (e.g., a key not found due to a benign race condition).
+- Use lazy `%`-formatting, not f-strings, in log calls so interpolation only runs when the record is emitted: `logger.info("instance %s", instance_id)`, not `logger.info(f"instance {instance_id}")`.
+- Pick the format specifier from the variable's **type annotation**, not from neighboring log calls:
+  - `int` (counts, numeric IDs) → `%d`
+  - everything else (`str`, `bool`, objects, exceptions) → `%s`
+  - floats use `%f` / `%.2f` as appropriate
+  - Watch ambiguous names: `instance_id` is a `str` in most modules (e.g. `cache_engine.py`, `config_base.py`) but an `int` GPU index in some multiprocess modules — check the annotation before choosing `%s` vs `%d`. Bools print as `1`/`0` under `%d`, so use `%s` when you want `True`/`False`.
 
 ### 7.5 Error Handling
 


chunxiaozheng (approved):
lgtm

maobaolong (approved):
LGTM

PR #3802, opened 2026-06-22
[good-first-issue] fix(mp): add background cleanup thread to SessionManager for expired session reclamation (#3562)

### LINKED ISSUE #3562: [good-sec-issue] SessionManager.cleanup_expired is never called in production

April-Sonnet:
Closes #3562
Refs #3372

`SessionManager.cleanup_expired()` was defined but never called in production code — sessions that were never explicitly removed via `remove()` would accumulate in `_sessions` indefinitely, causing a slow memory leak.

This PR adds a `SessionCleanupThread` (subclass of `PeriodicThread`) that periodically calls `cleanup_expired()` at a configurable interval (default 60s, configurable via `cleanup_interval` constructor argument). The thread starts automatically in `SessionManager.__init__`, is registered with the global `PeriodicThreadRegistry`, and can be stopped via `SessionManager.stop_cleanup()`.

**Changes:**
- New `SessionCleanupThread` class in `session.py` — PeriodicThread subclass with LOW level
- `SessionManager.__init__` now starts the cleanup thread automatically
- `SessionManager.stop_cleanup(timeout)` for graceful teardown
- `MPCacheServerContext.close()` calls `stop_cleanup()` during shutdown
- Test fixtures updated to stop threads after each test

- The default interval (60s) and init_wait (30s) follow the pattern used by other PeriodicThreads in the codebase
- Thread level is LOW since a missed cleanup cycle is not critical — sessions will still be cleaned up on the next cycle
- Tested that existing SessionManager tests pass with the new thread lifecycle

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/engine_context.py
+++ b/lmcache/v1/multiprocess/engine_context.py
@@ -163,8 +163,10 @@ class MPCacheServerContext:
 
     def close(self) -> None:
         """
-        Tear down the storage manager and the process-global GDS context.
+        Tear down the storage manager, session manager, and the
+        process-global GDS context.
         """
+        self._session_manager.stop_cleanup()
         self._storage_manager.close()
         # Tear down the GDS cuFile context (the shared slab + its handle).
         get_gds_context().close()
--- a/lmcache/v1/multiprocess/session.py
+++ b/lmcache/v1/multiprocess/session.py
@@ -14,8 +14,51 @@ import time
 from lmcache.logging import init_logger
 from lmcache.v1.multiprocess.custom_types import IPCCacheServerKey
 from lmcache.v1.multiprocess.token_hasher import TokenHasher
+from lmcache.v1.periodic_thread import (
+    PeriodicThread,
+    PeriodicThreadRegistry,
+    ThreadLevel,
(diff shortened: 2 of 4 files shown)


maobaolong:
@April-Sonnet Thanks for this contribution, but #3570 has been merged, so we can close this PR for now.

PR #3808, opened 2026-06-22
feat(mp): async engine-driven store

hlin99:
Introduce AsyncEngineDrivenTransferContext that offloads both the gather (GPU→CPU copy) and commit (CPU→server) phases of engine-driven stores to a background ThreadPoolExecutor, unblocking the forward thread.

Only the store path is made async; retrieve remains synchronous as its performance benefit is expected to be minimal (the forward thread must wait for KV data before proceeding anyway). Async retrieve is left as a future TODO if profiling shows otherwise.


lmcache/v1/multiprocess/transfer_context/async_engine_driven.py
```diff
@@ -0,0 +1,326 @@
...
+
+    ``submit_store`` only performs lightweight preparation (prepare_store,
+    buffer allocation) on the forward thread and immediately submits all
+    GPU/copy work to the background ``commit_executor``, so the forward thread
+    is never blocked by gather kernel launch latency.
+
+    This class is only instantiated by the factory when the device is
+    async-capable, so the constructor creates async resources unconditionally;
+    there is no ``self._async_capable`` flag.
+    """
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/transfer_context/async_engine_driven.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        self._sync_gather_events(suppress_errors=False)
+
+    def close(self) -> None:
+        """Drain in-flight gather/commit work before closing the base context."""
+        with self._inflight_lock:
+            self._is_closing = True
+        self._sync_gather_events(suppress_errors=True)
+        self._commit_executor.shutdown(wait=True, cancel_futures=False)
+        super().close()
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/transfer_context/async_engine_driven.py
```diff
@@ -0,0 +1,326 @@
...
+
+        return completion
+
+    def flush_inflight_gathers(self) -> None:
+        """Synchronize all in-flight gather (GPU->CPU) events.
+
+        Called at preemption/eviction time so that vLLM cannot overwrite
+        paged KV blocks before a deferred gather has finished reading them.
+        """
+        self._sync_gather_events(suppress_errors=False)
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/transfer_context/async_engine_driven.py
```diff
@@ -0,0 +1,326 @@
...
+            # never double-release or double-resolve.
+            commit_executor.submit(_commit_after_gather)
+        except Exception:
+            logger.exception("Failed to submit async engine-driven store")
+            if staged_chunks:
+                self._release_staging(staged_chunks)
+            completion.set_result(False)
+            return completion
+
+        return completion
```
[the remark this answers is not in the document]
hlin99: fixed

lmcache/v1/multiprocess/transfer_context/async_engine_driven.py
```diff
@@ -0,0 +1,326 @@
...
+        self._commit_executor: ThreadPoolExecutor = ThreadPoolExecutor(
+            max_workers=self._commit_workers,
+            thread_name_prefix="lmcache_engine_driven_commit",
+        )
+        self._inflight_lock = threading.Lock()
+        self._inflight_gather_events: set[Any] = set()
+        self._staging_pool: dict[
+            tuple[tuple[int, ...], torch.dtype], list[torch.Tensor]
+        ] = {}
+        self._is_closing = False
```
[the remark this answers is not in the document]
hlin99: fixed

hlin99:
FYI - some perf data

1. CUDA device uses engine driven **sync** mode vs **async** mode (1st time run, no cache hit, store time matters)

> 🎯 **Conclusion:** 
> In async mode, the vLLM forward pass returns immediately, resulting in higher overall throughput and reduced execution time.

| SYNC  | ASYNC |
| :--- | :--- |
| image | image |

2. CUDA device uses async mode by LMCache driven  vs engine driven  (1st time run, no cache hit, store time matters)

> 🎯 **Conclusion:** 
> In async mode, the vLLM forward pass returns immediately whether utilizing the LMCache-driven or engine-driven path. Consequently, both approaches yield equivalent overall throughput and comparable execution times.

| LMCache Driven (IPC + c ops) | Engine Driven (Data + c ops) |
| :--- | :--- |
| image | image |


3. CUDA async mode by LMCache driven + c ops vs engine driven + py ops  (1st time run, no cache hit, store time matters)

> 🎯 **Conclusion:** 
> In async mode, although Python-based copy operations are inherently slower than C operations, they do not degrade overall throughput or execution time. This holds true provided the Python copy completes within the window between consecutive store operations, effectively hiding the latency.

| LMCache Driven (IPC + c ops) | Engine Driven (Data + py ops) |
| :--- | :--- |
| image | image |

docs/design/v1/multiprocess/engine_driven_transfer_design.md
```diff
@@ -123,14 +141,23 @@ Why `prepare → data operation → commit`:
...
-(CUDA → `LMCacheDrivenTransferContext`, otherwise → `EngineDrivenTransferContext`).
+and async capability:
+- CUDA device → `LMCacheDrivenTransferContext`
+- Non-CUDA device → `_build_engine_driven_context()`, which probes async primitives:
+  - async primitives available (stream, event with record/synchronize/wait, pin_memory) →
+    `AsyncEngineDrivenTransferContext`
+  - otherwise → `EngineDrivenTransferContext` (synchronous fallback)
+
+The routing can be overridden via the `mode` argument or `LMCACHE_MP_TRANSFER_MODE`
+environment variable (`MPTransferMode.AUTO` / `LMCACHE_DRIVEN` / `ENGINE_DRIVEN`).
```
ApostaC: A separate note: in general, I think the main functionality should be controlled only by CLI args. 
This is because the environment variables config is pretty hard to track and document.

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -217,6 +220,15 @@ def submit_retrieve(
     def close(self) -> None:
         """Release resources held by this context."""
 
+    def flush_inflight_gathers(self) -> None:
```
ApostaC: nit on naming: `gather` needs more context to understand. Maybe we can call it `flush_inflight_stores`?

ApostaC: Maybe we can also make this an abstract function and also implement this in the subclasses as a no-op?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1695,6 +1698,24 @@ def get_block_ids_with_load_errors(self) -> set[int]:
...
+
+        Args:
+            need_flush: When True, flush in-flight gather operations on the
+                transfer context. When False, this is a no-op.
+        """
+        if not need_flush:
+            return
+        if not self.is_healthy or self.transfer_ctx is None:
+            return
+        self.transfer_ctx.flush_inflight_gathers()
```
ApostaC: One potential problem for LMCache-driven transfer.
If the code is still the way I understand it, the worker adapter will hold the MessageQueueFutures returned by the LMCache-driven transfer context. The transfer_ctx itself does not know whether the transfer has finished or not. The worker adapter should query those futures and wait until they are finished.

ApostaC (commented):
Please see the comments here

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
hlin99: agree! fixed by [6c40881](https://github.com/LMCache/LMCache/pull/3808/commits/6c408811a0044482a8323efbfdf59e8ddc4066e9)

docs/design/v1/multiprocess/engine_driven_transfer_design.md
hlin99: agree, fixed by [32c097a](https://github.com/LMCache/LMCache/pull/3808/commits/32c097a004d080d147cc8c5d764c97f7bd377060)

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
hlin99: good suggestion. fixed by [384c54d](https://github.com/LMCache/LMCache/pull/3808/commits/384c54dfd431108b2e9532049604747afaff1a39)

lmcache/integration/vllm/vllm_multi_process_adapter.py
hlin99: hi @ApostaC . I revisited the flow, and realized that there might be a small window where worker hand over the buffer to server, and serve does not put it on stream timely, that there will be problem during such a window, preemption comes.

For the buffer that server already put into stream, the torch_dev.synchronize() will guarantee finish.

so for LMCache-driven transfer, the logic needs to be adjusted to properly handle preemption. It won't be too complex in my mind, but too keep this PR focusing on async tranfer, a seperate PR for LMCache-driven transfer preemption handler would be better. WDYT? I can do at follow up PR.

ApostaC:
Thanks for the update! I will take another look soon!

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -1049,6 +1073,50 @@ def build_connector_meta(
...
+          an in-flight gather against the old blocks must be flushed first.
+        - ``SchedulerOutput.preempted_req_ids``: requests preempted this step.
+          Their blocks are freed and may be reused, so the same applies.
+
+        Args:
+            scheduler_output: The vLLM scheduler output for this step.
+
+        Returns:
+            True if a flush is needed, False if the step is preemption-free.
+        """
```
ApostaC: Regarding this function: maybe we should double-check when & how vLLM calls `handle_preemption`

I vaguely remember that `handle_preemption` will always be called when there is preemption. Therefore, we may not need to have this function. (Also may not need to have `self._need_flush`)

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1695,6 +1698,24 @@ def get_block_ids_with_load_errors(self) -> set[int]:
...
+            need_flush: When True, flush in-flight gather operations on the
+                transfer context. When False, this is a no-op.
+        """
+        if not need_flush:
+            return
+        if not self.is_healthy or self.transfer_ctx is None:
+            return
+        self.transfer_ctx.flush_inflight_stores()
+        # Force device sync here, compare to preemption, perf panelty is trivial
+        torch_dev.synchronize()
```
ApostaC, line 1717: QQ on this: for lmcache driven mode, cpu <-> gpu copy happens in another process. Will `torch_dev.synchronize()` also sync the kernels that launched in a different process?

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -556,8 +576,69 @@ def create_transfer_context(
...
     if resolved_mode is MPTransferMode.ENGINE_DRIVEN:
-        return EngineDrivenTransferContext()
+        return _build_engine_driven_context()
     # AUTO: dispatch by device type (CUDA -> handle path, else -> data path).
     if device_type == "cuda":
         return LMCacheDrivenTransferContext()
+    return _build_engine_driven_context()
+
+
+def _supports_async_primitives() -> bool:
```
ApostaC: nit: let's put `_build_engine_driven_context` and `_supports_async_primitives` to the top of this file.

In general, we are following the convention of:
- imports
- (private) helper function definition and helper class definition
- main class definition
- module-level public function definition

ApostaC (commented):
Otherwise LGTM!

lmcache/integration/vllm/lmcache_mp_connector.py
hlin99: > Regarding this function: maybe we should double-check when & how vLLM calls `handle_preemption`
> 
> I vaguely remember that `handle_preemption` will always be called when there is preemption. Therefore, we may not need to have this function. (Also may not need to have `self._need_flush`)

hi @ApostaC , even though, I don't quite understand why vllm calls handle_preemptions in each step (not only when preemption happens), vllm actually calls it in every model executor without any judgement. So whether preemption happens or not relies on connector to handle it properly. that's why this function was introduced.

So i think the logic is required, but maybe I can rename the function for readability and leave some comments there.

And in addition, when I revisit mp connector code, I realized what you mentioned that for LMCache driven mode, we intentionally overlook preemption in some cases. I will try to fix it in the next PR, both for common scheduler logic and lmcache driven worker logic. sounds good?

image

image

lmcache/integration/vllm/vllm_multi_process_adapter.py
hlin99: hmmm.... very good point. I overlooked it crosses processes in LMCache driven mode, so it's not as straightforward as I thought. maybe additional MSG is required for preemption handling.

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
hlin99: thanks for the guidance! will follow the convention moving forward!

hlin99:
GSM8K: No crash & No accuracy issue, during store/retrieve with preemptions. the scores are equivalent each time

Baseline score: w/o any preemption
image


1st run w/ preemption (KV cache store)
image

2nd run w/ preemption (KV cache retrieve)
image

hlin99:
hi @ApostaC  thanks for the review and comments. addressed all of them. please take another look. thx.

lmcache/integration/vllm/lmcache_mp_connector.py
ApostaC: Awesome! Thanks for checking!

ApostaC (approved):
LGTM!

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -101,6 +101,42 @@ class SupportsHMA:  # type: ignore[no-redef]
...
+
+    Returns:
+        True if a flush is needed, False otherwise.
+    """
+    cached_reqs = getattr(scheduler_output, "scheduled_cached_reqs", None)
+
+    # Primary signal: requests resumed from preemption this step.
+    resumed_ids = getattr(cached_reqs, "resumed_req_ids", None)
+    if resumed_ids:
+        logger.warning("<preempted> by resumed requests: %s", resumed_ids)
```
sammshen, line 126: what is "< preempted > by"? can we just say resumed requests or preempted requests

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -101,6 +101,42 @@ class SupportsHMA:  # type: ignore[no-redef]
 
 
 # Helper functions
+def _is_preempted(scheduler_output: SchedulerOutput) -> bool:
```
sammshen: nit: suggest rename _has_preemption_ops() or something similar

sammshen: nit: please remove flushing related docstring comments here. the helper should not be aware of this

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -466,6 +502,11 @@ class LMCacheMPConnectorMetadata(KVConnectorMetadata):
     def __init__(self):
         super().__init__()
         self.requests: list[LMCacheMPRequestMetadata] = []
+        # True only on scheduler steps where preemption/eviction may overwrite
+        # blocks referenced by in-flight async stores. Worker-side
+        # handle_preemptions() uses this hint to flush deferred gather
+        # (device->CPU copy) work before vLLM can overwrite source KV blocks.
+        self.need_flush: bool = False
```
sammshen: can we be more verbose here? e.g. `need_flush_transfers_pre_forward_pass` (the exact name not sure)      also can we remove the comment and just rely on the variable name

ApostaC: Maybe `need_flush_before_forward`?

lmcache/integration/vllm/lmcache_mp_connector.py
hlin99: hi Samuel. thanks for the review and comments.

this <preempted> is kind of label that later on, i would like to setup a CI for preemption, and it would be easier to grep the keyword.... WIP PR https://github.com/LMCache/LMCache/pull/3981

do you think it's okay to leave such msg here, or any other suggestion?

lmcache/integration/vllm/lmcache_mp_connector.py
hlin99: good suggestion! thanks @sammshen  & @ApostaC 
fixed by [9ac1db5](https://github.com/LMCache/LMCache/pull/3808/commits/9ac1db528d3dac69d2fbf818e22180643d25effd)

lmcache/integration/vllm/lmcache_mp_connector.py
hlin99: good point!

fixed by [75dd679](https://github.com/LMCache/LMCache/pull/3808/commits/75dd679655efd8deda9ec3b980f6361f82142b2e)

hlin99:
hi @sammshen thanks for your review

I either fixed it or left comments for explanation. would you like to have another look? thx.

lmcache/integration/vllm/lmcache_mp_connector.py
sammshen: let's say <preemption req detected>

sammshen: current is fine

sammshen (approved):
LGTM!

PR #3812, opened 2026-06-22
[Core][RawBlock] Batch load_many reads with per-IO results

3xdevv:

Raw-block `load_many` submitted reads individually even when io_uring batching was available. Additionally, `wait_iouring` exposed only an aggregate failure, preventing callers from preserving successful reads when another operation in the batch failed. This PR reduces submission overhead and reports results per requested key.

- Collect eligible `load_many_into` reads and submit them in one batch.
- Batch split `uring_cmd` chunks and map their completion results back to logical reads.
- Copy aligned temporary buffers back only when every chunk for a logical read succeeds.
- Return a per-operation success bitmap from the Rust `wait_iouring` API.
- Validate completion counts and fail the expected batch when the bitmap is incomplete or oversized.
- Preserve successful reads while returning `False` or `None` for individual failures.
- Fail write operations when any submitted completion reports failure.
- Update test doubles for per-I/O results and FDP placement IDs.

- `wait_iouring` now returns `list[bool]` instead of `None`.
- For `uring_cmd`, completion counts correspond to physical chunks after splitting, while callers receive results aligned with logical reads.
- Submission and setup errors can still be raised before a batch ID is returned.

**If applicable**:


lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -1184,7 +1200,15 @@ def _write_uring_cmd_buffers(
             chunk_buffers,
             chunk_lens,
         )
-        raw_dev.wait_iouring(batch_id)
+        if not all(
+            self._wait_iouring_results(
```
DongDongJu, line 1361: I think we should return all-false on length mismatch btw returned one and original requested length.

DongDongJu (changes requested):
I left one nit.
Otherwise LGTM

lmcache/v1/storage_backend/raw_block/core.py
3xdevv: I intentionally kept _write_uring_cmd_buffers as a raise-on-failure path instead of making it return a per-input bitmap like _read_uring_cmd_buffers. This PR is focused on the read/load path, where callers need per-key success results. For writes, the current behavior remains: validate completions for the submitted chunks and raise on failure or count mismatch. When the write path is updated later, similar to the direction in #3636 , this can be resolved by changing the write helpers to return per-input results as well.

DongDongJu (approved):
Looks good overall, but before merge 
please clarify that `expected_count` is the submitted chunk-I/O count

rust/raw_block/src/lib.rs
```diff
@@ -2423,24 +2424,16 @@ impl RawBlockDevice {
...
                         }
                     }
                     // Clear stored buffer objects for this batch
                     let mut stored_objs = self.batched_buffer_objs.lock().unwrap();
                     stored_objs.remove(&batch_id);
-                    return if let Some(e) = first_error {
-                        Err(e)
-                    } else {
-                        Ok(())
-                    };
```
zhengfeihe, line 2443: Can we add a log here for the detailed error message, otherwise we are completely ignoring error infos

rust/raw_block/src/lib.rs
```diff
@@ -2405,12 +2405,13 @@ impl RawBlockDevice {
     ///     batch_id: The batch ID returned by batched_write() or batched_read().
     ///               Only completions from this batch are checked.
     ///
-    /// Returns an error if any I/O operation in this batch failed. The error message
-    /// includes details about the first failure encountered.
+    /// Returns a success bitmap aligned with the operations submitted in this
+    /// batch. Submission/setup errors are still raised before a batch id is
+    /// returned.
```
zhengfeihe: Please also update the Document (README) to match this code change

zhengfeihe (commented):
I have two nits. Otherwise LGTM~

zhengfeihe (approved):
LGTM!

PR #3818, opened 2026-06-22
[Feat][MP] Add local memory pressure status reporting

DongDongJu:

Adds a policy-aware, node-local memory-pressure status contract for MP servers.

This PR:

- Adds immutable, transport-neutral L1 and per-adapter L2 usage snapshots in `StorageManager`.
- Preserves stable adapter IDs and known used bytes when capacity is unknown.
- Collects usage and health independently, so one backend failure produces an `unknown` partial entry instead of failing the whole report.
- Classifies `high` at the configured eviction watermark and `critical` at the tier's reported logical capacity; health remains separate from capacity pressure.
- Keeps each L2 adapter distinct and respects `IsolatedLRU` per-`cache_salt` watermark semantics.
- Serializes runtime adapter removal while backend status is read, preventing an adapter from being closed during a snapshot.
- Adds `GET /status/memory-pressure` to the current typed HTTP context and offloads synchronous backend collection from the event loop.
- Documents the response, failure semantics, metrics relationship, Linux PSI relationship, and coordinator boundary.

**Coordinator scope**

The current MP coordinator owns fleet membership/heartbeats and fleet control paths such as L2 usage/quota/eviction/events, pin/delete, and prefetch. 
It does not currently ingest or persist per-instance memory-pressure snapshots. 
This patch therefore keeps status collection local to the MP server and does not put volatile pressure state into the coordinator registry. 
A future coordinator reporter or proxy can consume the same transport-neutral snapshot contract.

**Metrics scope**

This revision removes the earlier request-populated metrics subscriber, eviction-rate sampler, and standalone `/memory_pressure` route. 
Current `dev` already provides live L1 usage, L2 used-byte, and eviction time-series signals, but not L2 capacity/watermark pressure series. 
This PR deliberately defines a local point-in-time status contract first; `/status/memory-pressure` never mutates metric state. A future live metric or coordinator reporter can consume the same canonical snapshot.

This PR provides the complementary LMCache-local view through `GET /status/memory-pressure`: a point-in-time L1 and per-L2-adapter snapshot of usage, capacity, configured eviction policy and watermark, health, and derived capacity-pressure level. 

**If applicable**

- [x] this PR contains user-facing changes - docs added


docs/source/mp/http_api.rst
```diff
@@ -415,6 +419,105 @@ been initialized:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Returns a typed, point-in-time view of local capacity pressure. The MP server
+collects L1 usage and each live L2 adapter independently; it does not aggregate
+multiple L2 backends, so a hot adapter cannot be hidden by a colder one.
+``adapter_id`` is the stable runtime adapter index and is ``null`` for L1.
+
+The level follows local policy rather than fixed monitoring thresholds:
+
+- ``high`` means usage reached the tier's configured global-capacity eviction
```
sammshen, line 432: won't eviction mean we never report `high`?

docs/source/mp/http_api.rst
```diff
@@ -415,6 +419,105 @@ been initialized:
...
+  ``trigger_watermark``;
+- ``critical`` means reported usage reached physical capacity;
```
sammshen: nit: maybe remove physical here. it is the logical size we set for the backend right?

docs/source/mp/http_api.rst
sammshen: would it be a lot of work to just report a percentage?

docs/source/mp/http_api.rst
DongDongJu: you are right

docs/source/mp/http_api.rst
DongDongJu: Yes under healthy steady-state eviction, `high` should usually be brief rather than a latched alert. 

The node-local L1/L2 eviction loops sample once per second and start reclaiming at the same `trigger_watermark`; this endpoint samples independently, so it can report `high` after the threshold is crossed, while reclamation is running, or when reclamation is incomplete. 

A sparse external poll can miss that excursion after usage drops.

The local MP server remains responsible for eviction, and this level should not be used as a Kubernetes liveness/readiness decision.
If we add fleet-level control later, the coordinator can poll and aggregate the continuous ratio/history, and a Kubernetes controller or metrics adapter could apply placement, scaling, or draining policy.

DongDongJu: No additional backend collection is needed: every tier already returns `used_ratio = used_bytes / capacity_bytes`. 
For example, `0.8` means 80%; it is `null` when usage or capacity is unknown and can exceed `1.0` if reported usage exceeds reported capacity. 
Keeping it as a fraction also puts it on the same scale as `trigger_watermark` and avoids duplicate ratio/percentage fields.

A future coordinator can aggregate this value across MP servers and expose it through metrics for a Kubernetes controller or custom-metrics consumer. 
Neither integration exists today; this PR only defines the node-local status contract.

DongDongJu:
@KuntaiDu PTAL

PR #3820, opened 2026-06-22
Token dropping sdk v3

rannnayy:

This PR is a third version of #3755 and #3787 meant to implement token dropping API as a KV cache SDK so that ML engineers can do KV cache optimization inside KV connector. This is done by sending a prefill request, calling `retrieve()` API to obtain the KV cache tensors, modify it, then call the `store()` API to store back the KV cache to LMCache before sending a decode request with the modified KV cache.

Unlike the previous two versions, this version registers an `EngineDrivenTransferContext` and wraps the `EngineDrivenContext` it creates in a new `ContiguousEngineDrivenTransfer`, a thin adapter that copies contiguous chunk slices straight to/from the transport's chunk slots, skipping redundant paged scatter/gather in V2 (#3787). It masks the SHM-vs-pickle difference exposed by `prepare_store()`. The SDK can then call `ContiguousEngineDrivenTransfer.store()` / `.retrieve()` directly.

**If applicable**:


docs/design/sdk/kvcache.md
```diff
@@ -0,0 +1,282 @@
+# KV Cache SDK
```
KuntaiDu, line 1: It's a little bit long. Ideally less than 200 lines

docs/design/sdk/kvcache.md
```diff
@@ -0,0 +1,282 @@
...
+> therefore built in HND-physical order `[NB, 2, NH, BS, HS]` regardless of the engine's GPU
+> layout, so detection reads NH/BS from the right slots.
+
+## Public API
+
+Exported from `lmcache.sdk`: `LMCacheKVCacheContext`, `KVCacheSDKError`.
+Operations are module-level functions in `lmcache.sdk.kvcache`: `connect`, `close`,
+`retrieve`, `store`.
+
+### `connect(url, http_url, model_name, kv_buffer_bytes, timeout=60.0) -> LMCacheKVCacheContext`
```
KuntaiDu: Ideally we want to have just 1 url, but i'm fine with leaving it to future pr

lmcache/sdk/kvcache.py
```diff
@@ -0,0 +1,509 @@
...
+            mp_conf = response.json()["mp"]
+        except (requests.RequestException, KeyError, ValueError) as err:
+            raise KVCacheSDKError(
+                f"failed to fetch server config from {self._http_url}/conf"
+            ) from err
+        self._chunk_size: int = int(mp_conf["chunk_size"])
+        self.shm_name: str = str(mp_conf.get("shm_name", "")).lstrip("/")
+        if self.shm_name and not self.shm_name.startswith("lmcache_l1_pool_"):
+            self.shm_name = f"lmcache_l1_pool_{self.shm_name}"
+        print(
```
KuntaiDu: logger.info

lmcache/sdk/kvcache.py
```diff
@@ -0,0 +1,509 @@
...
+            ) from err
+
+        self._pending_lookups: set[str] = set()
+        self._finished_lookups: dict[str, int] = {}
+
+    def register_kv_caches(
+        self,
+    ) -> None:
+        """Register the KV cache layout for the model with the SDK context."""
+        entry = next((e for e in self._cache_context_meta_conf.values()
```
KuntaiDu: Readability

lmcache/sdk/kvcache.py
```diff
@@ -0,0 +1,509 @@
...
+            kernel_group = kernel_groups[0]
+            dtype = getattr(torch, kernel_group["dtype"].replace("torch.", ""))
+            tokens_per_block = kernel_group.get("tokens_per_block", 0)
+
+            inner = [
+                int(x) for x in
+                kernel_group["engine_kv_concrete_shape"].split("[")[1].rstrip("] ").split(",")
+            ]
+
+            # From get_num_heads (lmcache/v1/gpu_connector/utils.py)
```
KuntaiDu: Import from gpu_connector/utils

lmcache/sdk/wrapper/contiguous.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-License-Identifier: Apache-2.0
```
KuntaiDu, line 1: Maybe move to sdk/kvcache/wrapper/
and rename to ContiguousTransferWrapper

docs/design/sdk/kvcache.md
rannnayy: The two URLs are different. One is for the ZMQ, and another is for HTTP, only for learning the KV cache size. There is no endpoint using ZMQ that exposes the KV cache size, so we can't reduce it to only the ZMQ URL without adding more protocol.

examples/kvcache_sdk/e2e_kv_edit.py
```diff
@@ -0,0 +1,342 @@
+# SPDX-License-Identifier: Apache-2.0
```
KuntaiDu: Maybe change this file to jupyter notebook (and make it much shorter >.<)

examples/kvcache_sdk/run_e2e_kv_edit.sh
```diff
@@ -0,0 +1,286 @@
+#!/usr/bin/env bash
```
KuntaiDu: Same here --- move these 2 files to one unified jupyter notebook example. IIRC you can run bash command inside jupyter notebook.

KuntaiDu (approved):
Otherwise LGTM

examples/kvcache_sdk/run_e2e_kv_edit.sh
rannnayy: Yes, we can do that, but then since the LMCache and vLLM should be running while the example e2e is running, the cell won't stop.

To solve this, there are 3 options:
1. Run in Jupyter Notebook Shell script cell with `&` appended to the end of the command (put the process on the background). The issue is that, these processes will need to be killed later using the PID, which is a much complicated for an ML engineer (not MLSys).
2. Run in Jupyter Notebook Python script via subprocess.Popen (or similar functions). This will obfuscate the real command, making it a bit harder to modify (requires some thinking load).
3. Ask the user to run the command in a terminal (chosen for now).

Please let me know if this sounds good.

sammshen (approved):
LGTM!

PR #3822, opened 2026-06-22
[operator][feat] Auto-inject LMCacheEngine connection into vLLM pods via webhook

royyhuang:

Wiring a vLLM Deployment to an operator-managed `LMCacheEngine` currently
means hand-mounting the `<engine>-connection` ConfigMap and launching vLLM
through a `sh -c` wrapper that cats it into `--kv-transfer-config`. The
operator already solves this for `CacheBlendEngine` with a mutating webhook;
this PR brings the same ergonomics to the standard `LMCacheEngine` path.

A vLLM pod now opts in with a label + annotation and the webhook injects the
connection wiring at pod CREATE — no manual mount, no shell wrapper.

**User-facing changes**:

Opt a vLLM pod in via pod-template label `lmcache.ai/lmcache-inject: "true"`
plus annotation `lmcache.ai/lmcache-engine: "<engine>"` (optional
`lmcache.ai/lmcache-container` to target a non-first container). Launch
args-only through the image ENTRYPOINT. The webhook then injects, reading the
engine's `<engine>-connection` ConfigMap:

- `--kv-transfer-config <inline JSON>` (the `LMCacheMPConnector` config, inlined — no volume mount needed)
- `hostIPC: true` (CUDA IPC with the node-local LMCache server)
- `PYTHONHASHSEED=0` (deterministic prefix hashing, set only if absent)

It fails open (`failurePolicy: Ignore`) and is idempotent, mirroring the
CacheBlend webhook. Skips are diagnosable via `lmcache.ai/lmcache-skip-reason`
(`command-override`, `kv-transfer-config-present`, `engine-not-found`,
`target-container-not-found`). New sample: `config/samples/vllm_lmcache_deployment.yaml`
(the old manual-mount `vllm_deployment.yaml` sample is removed in favor of it).

- The new `LMCachePodInjector` (path `/mutate-lmcache--v1-pod`, webhook
  `mlmcachepod.lmcache.ai`) gates on the connection ConfigMap **alone** — no
  engine-CR lookup. The whole connector config lives in the ConfigMap and
  `LMCacheEngine` has no injection sub-spec, so there are **no new RBAC or CRD
  changes** (`configmaps get` is already granted).
- Inline-JSON injection (not a volume mount): the `$(cat ...)` form needs a
  `command:` override, which collides with the command-override skip gate.
  Inlining mirrors the CacheBlend injector exactly.
- The skeleton shared by both injectors (decode/opt-in gate, connection +
  target-container resolution, command-override gate, skip/stamp/patch) is
  extracted into `pod_inject_common.go` and reused, cutting the duplication.
- For disambiguation the CacheBlend handler is renamed
  `pod_injector.go` -> `cacheblend_pod_injector.go`, its struct
  `PodInjector` -> `CacheBlendPodInjector`, and its webhook
  `mpod.lmcache.ai` -> `mcacheblendpod.lmcache.ai`. No behavior change.
- Validated locally: `make test` (webhook coverage 86.7%) and
  `make test-e2e-gpu-kind` (15 passed, 0 failed) — including a new CPU-only
  injection smoke that asserts the stamp + injected wiring off a real pod.

**If applicable**:


ruizhang0101 (approved):
LGTM

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3779, opened 2026-06-20
[Model Support]: Group layers w/ EngineKVFormat for MiniMax M3

sammshen:
Detect the Engine KV format per engine group (engine + server sides) and build each kernel group's shape_desc from per-layer formats for models that mix formats across groups. Per-layer formats are transient; contexts keep a single representative format for the not-yet-per-group consumers. No behavior change for existing models.

PR #3826, opened 2026-06-23
[Refactor] Dispatch transfers per group EngineKVFormat (MiniMaxM3 2/2)

sammshen:
Persist the format per kernel group (KernelGroupInfo.engine_kv_format, set from the value PR A already computes) and switch the per-group consumers from the context's single representative to group.engine_kv_format: multi_layer_block_kv_transfer, get_group_data_ptrs, the base_cache_context shape-desc path, and is_mla where per-group. Supply the format in parse_kvcache_shape_spec. Route M3 through the MP connector (non-MP stays single-format). Adds the M3 end-to-end store/retrieve correctness test. Enables MiniMax-M3; GPU/M3-verified.

### DIFF
--- a/lmcache/integration/vllm/kv_cache_groups.py
+++ b/lmcache/integration/vllm/kv_cache_groups.py
@@ -118,7 +118,7 @@ def create_engine_group_infos_from_vllm(
     # First Party
     from lmcache.utils import EngineType
     from lmcache.v1.gpu_connector.utils import (
-        normalize_and_discover_per_group_formats,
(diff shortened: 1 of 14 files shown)


sammshen:
Folded into #3779 — the two halves (group + transfer dispatch) are one ~90-line coherent change, both CPU-tested and no-op for existing models, so they read cleaner as a single PR than a stacked pair. #3779 now covers both; the M3 enablement (MP routing + e2e correctness test) remains the follow-up.

PR #3827, opened 2026-06-23
[MP][Coordinator] Warm-prefetch KV from L2 into L1

ruizhang0101:

Adds a coordinator-controlled *warm prefetch*. A client POSTs a token sequence + a target node to the coordinator, which dispatches to that MP server's new `POST /l2/prefetch`. The server hashes the tokens into per-rank keys and loads those chunks from L2 into L1 — leaving them **resident, retained, and unlocked** — so a later request for that prefix hits L1 instead of paying the L2 fetch inline. Useful for pre-warming a node ahead of a known or shifting workload.

Key pieces:
- **`PrefetchMode.WARM`** (a warm-only prefetch intent) threaded through `StorageManager.submit_prefetch_task` → `PrefetchController`. It redefines the load to *retain (permanent)* **and acquire no read lock**: a warm has no imminent consumer to pin for, so loaded keys transition write-locked → ready via `finish_write` (not `finish_write_and_reserve_read`), and the L1-hit `reserve_read` pass is skipped entirely. The chunks are left resident, unlocked, evictable, and re-lookupable — a later real lookup takes its own lock.
- MP-server **`POST /l2/prefetch`** (token-based; key construction stays server-side) returning a `request_id`, and **`GET /l2/prefetch/{request_id}`** for status. State lives in a `WarmPrefetchJobs` handle table on `app.state`. Status is a **reactive query** — there is no server-side poll loop and nothing to release (no lock), so completion is observed on demand. It reports `found_keys`/`total_keys` (the chunks this request loaded from L2).
- Warm loads use the gap-tolerant **`TrimPolicy.SPARSE`** so an already-resident leading chunk doesn't stop the rest from loading. (The default `PREFIX` would trim everything past the first resident chunk, because the warm skips the L1-hit pass and a resident chunk reads as a gap.)
- Coordinator **`POST /l2/prefetch`** + **`GET /l2/prefetch/{instance_id}/{request_id}`** via `L2PrefetchManager`, which proxies the submit and relays status from the named instance (404 for an unknown instance, 502 if unreachable).

- **No lock by design.** A warm has no downstream reader, so it holds no read lock — nothing to release, no abandonment leak, no pinned L1. The job table is handle-only; an unpolled job lingers only as a small dict entry (no pinned memory) until polled or dropped by a future TTL sweep.
- Token-based by design: callers can't build internal content keys (chunk hash + per-rank layout bitmap), so they send `token_ids` and the server reconstructs keys via the same path as lookup.
- `found_keys`/`total_keys` count only chunks *loaded from L2* by this request; chunks already resident in L1 are skipped at `reserve_write` and not counted — deliberate, since an already-resident entry may be a transient temporary from another lookup, so claiming it as warmed could mislead. Documented in the API reference and design doc.
- `skip_l2=True` is honored for `WARM` (returns an empty handle without touching L2), with a regression test.
- Scope is single-node (one `instance_id` = one node's shards); multi-node fan-out is deferred to a future KV-cache-directory PR (called out in the design doc).
- Validated end-to-end on GPU (real coordinator + MP server over HTTP, FS L2 + GPU L1): a cold warm loads all chunks, retained and unlocked (`reserve_read` SUCCEEDS, `unsafe_read` NOT_READABLE), re-lookupable; and a partial-residency warm loads the remaining chunks under `SPARSE`.

**If applicable**:

- [x] this PR contains user facing changes — docs added (`docs/source/mp/coordinator.rst`, `docs/source/mp/http_api.rst`) plus design docs and a runnable example

### DIFF
--- a/lmcache/v1/distributed/api.py
+++ b/lmcache/v1/distributed/api.py
@@ -34,23 +34,15 @@ class TrimPolicy(enum.Enum):
 
 
 class PrefetchMode(enum.Enum):
-    """The intent of a prefetch, selecting the whole load discipline per request.
+    """The intent of a prefetch request.
 
-    The two modes are not independent knobs but one discriminant: each bundles a
-    retention decision *and* a locking decision, because both follow from whether
-    there is an imminent consumer for the loaded keys.
+    ``LOOKUP`` -- prefetch for an imminent reader: loaded keys are pinned for
+    the requesting workers, and whether they persist or are dropped after use
+    follows the configured prefetch policy.
 
-    ``LOOKUP`` -- load for an imminent reader (the lookup path). Retention defers
-    to the configured :class:`PrefetchPolicy` (the global ``prefetch_policy``
-    config decides temporary vs. retained), and loaded/already-resident keys are
-    **read-locked** to pin them for the requesting workers until they finish.
-
-    ``WARM`` -- speculative pre-warm (the coordinator-driven warm prefetch). There
-    is no imminent reader, so loaded keys are forced **permanent** and
-    transitioned write-locked -> ready via ``finish_write`` -- i.e. **no read
-    lock is acquired**. The object is immediately resident, unlocked, and
-    evictable; a warm holds nothing to release, and completion is observed by
-    querying the prefetch handle. A later real lookup takes its own lock.
+    ``WARM`` -- speculative pre-warm with no imminent reader: loaded keys are
+    retained and left unpinned (immediately resident and evictable), so a later
+    lookup can hit them.
     """
 
     LOOKUP = enum.auto()
--- a/lmcache/v1/distributed/api.py
+++ b/lmcache/v1/distributed/api.py
@@ -33,6 +33,30 @@ class TrimPolicy(enum.Enum):
     SPARSE = enum.auto()
 
 
+class PrefetchMode(enum.Enum):
+    """The intent of a prefetch, selecting the whole load discipline per request.
+
+    The two modes are not independent knobs but one discriminant: each bundles a
+    retention decision *and* a locking decision, because both follow from whether
+    there is an imminent consumer for the loaded keys.
+
+    ``LOOKUP`` -- load for an imminent reader (the lookup path). Retention defers
+    to the configured :class:`PrefetchPolicy` (the global ``prefetch_policy``
+    config decides temporary vs. retained), and loaded/already-resident keys are
+    **read-locked** to pin them for the requesting workers until they finish.
+
+    ``WARM`` -- speculative pre-warm (the coordinator-driven warm prefetch). There
+    is no imminent reader, so loaded keys are forced **permanent** and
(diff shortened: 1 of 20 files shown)


lmcache/v1/multiprocess/warm_prefetch.py
```diff
@@ -0,0 +1,107 @@
...
+        completes.
+    """
+    handle = await asyncio.to_thread(
+        storage_manager.submit_prefetch_task,
+        keys,
+        layout_desc,
+        extra_count=_WARM_PREFETCH_EXTRA_COUNT,
+        external_request_id=external_request_id,
+        retention=RetentionPolicy.RETAIN,
+    )
```
ApostaC: we don't need submit_prefetch_task because it's non-blocking

ApostaC (commented):
1. user facing doc
2. add api to query the prefetch status so we don't need to poll in the server process
3. most of functions are non blocking so we probably don't need asyncio

lmcache/v1/distributed/api.py
```diff
@@ -33,6 +33,30 @@ class TrimPolicy(enum.Enum):
...
+    to the configured :class:`PrefetchPolicy` (the global ``prefetch_policy``
+    config decides temporary vs. retained), and loaded/already-resident keys are
+    **read-locked** to pin them for the requesting workers until they finish.
+
+    ``WARM`` -- speculative pre-warm (the coordinator-driven warm prefetch). There
+    is no imminent reader, so loaded keys are forced **permanent** and
+    transitioned write-locked -> ready via ``finish_write`` -- i.e. **no read
+    lock is acquired**. The object is immediately resident, unlocked, and
+    evictable; a warm holds nothing to release, and completion is observed by
+    querying the prefetch handle. A later real lookup takes its own lock.
```
ApostaC: nit: simplify this

lmcache/v1/distributed/storage_manager.py
```diff
@@ -419,11 +421,47 @@ def submit_prefetch_task(
...
+            skip_l2: If True, do not submit to L2. For ``LOOKUP`` this returns
+                the already-resident L1 hit set only; for ``WARM`` (which does
+                not probe L1) it is a no-op that returns an empty handle.
+            mode: The prefetch intent (see :class:`PrefetchMode`).  ``WARM``
+                is the speculative pre-warm path: every key loaded from L2 is
+                retained (permanent) regardless of the configured prefetch
+                policy, and **no read lock** is taken (this method skips the
+                L1-hit ``reserve_read`` entirely).  ``LOOKUP`` (default) loads
+                for an imminent reader: retention defers to the policy and
+                found/loaded keys are read-locked.
```
ApostaC: nit: simplify this

ApostaC (commented):
Otherwise LGTM

ApostaC (approved):
LGTM!

KuntaiDu (approved):
LGTM

PR #3828, opened 2026-06-23
fix: resolve scheduler heartbeat thread startup failure

weizhoublue:

Before PR #3248,  _ensure_heartbeat_started() checks if  self._heartbeat  is  None  to decide whether to initialize heartbeat. It works to start the heartbeat.
```
self._heartbeat: HeartbeatThread | None = None
```

After PR #3248, Initial value is  {}  which is not  None .  _ensure_heartbeat_started() will return immediately and heartbeat thread never starts.
```
self._heartbeats: dict[str, HeartbeatThread] = {}
```

In Python, empty dict  {}  evaluates to  False , while non-empty dict evaluates to True.  Thus, using  if self._heartbeats  to check whether initialization completed is correct.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -682,10 +682,10 @@ class LMCacheMPSchedulerAdapter:
 
     def _ensure_heartbeat_started(self) -> None:
         """Lazily start the heartbeat thread on first use."""
-        if self._heartbeats is not None:
+        if self._heartbeats:
             return
         with self._heartbeat_lock:
-            if self._heartbeats is not None:
+            if self._heartbeats:
                 return
             for url, client in self.mq_clients.items():
                 hb = HeartbeatThread(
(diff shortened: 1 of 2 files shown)


weizhoublue:
@Lyj1007   this is a follow-up for PR #3248 
could you check wether this PR makes sense, thanks

Lyj1007:
> @Lyj1007[这是对 PR #3248](https://github.com/LMCache/LMCache/pull/3248) 的后续跟进， 请问这个 PR 是否合理？谢谢。

Thanks @weizhoublue for the careful follow-up, and apologies for the oversight on my side.

You're right — this is a regression from #3248. When I switched the scheduler-side heartbeat from a single `Optional[HeartbeatThread]` to a per-server `dict[str, HeartbeatThread]`, I forgot to update the guard in `_ensure_heartbeat_started`. Sorry for missing this when updating the type.

The added unit test is a nice regression guard — it covers both the "first call starts one `HeartbeatThread` per server URL" and "second call is a no-op" paths, which is exactly the behavior #3248 silently broke.

The fix LGTM.

Lyj1007:
The fix itself LGTM — the `{} is not None` early-return bug is real and worth merging.

One thought on the test: the new case covers a dual-server setup but lives in the single-node test file (`test_free_locks.py`). Would it be cleaner to put multi-node heartbeat tests in a separate file/class, so single-node and multi-node scenarios don't get mixed together as this grows?

@maobaolong  could you take a look at the test organization part?

PR #3829, opened 2026-06-23
refactor(platform): auto-discover DeviceIPCWrapper subclasses under platform/<device>/

maobaolong:

Move every concrete DeviceIPCWrapper next to its accelerator code under
platform/<device>/ and replace the static if/elif dispatch with runtime
subclass discovery (same pattern as cache_context.py). Adding a new
backend now only requires shipping a DeviceIPCWrapper subclass with
device_type and wrap() — no edits to the dispatcher.

**If applicable**:

### DIFF
--- a/lmcache/v1/platform/_registry.py
+++ b/lmcache/v1/platform/_registry.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 # Standard
 from typing import Any, Callable, Dict
+import threading
 
 # First Party
 from lmcache.logging import init_logger
@@ -46,8 +47,12 @@ _KV_WRAPPER_FACTORIES: Dict[str, Callable[..., Any]] = {}
 # Missing entry == always available.
 _AVAILABILITY: Dict[str, Callable[[], bool]] = {}
 
-# Guard so discovery only runs once (lazy init).
+# Guard so discovery only runs once (lazy init).  The lock plus the
+# double-checked flag below keep the first concurrent caller from
+# racing a second one through the scan and emitting duplicate
+# "multiple wrappers claim device_type=..." warnings.
 _WRAPPERS_DISCOVERED: bool = False
+_DISCOVERY_LOCK = threading.Lock()
 
 
 def _discover_wrappers_once() -> None:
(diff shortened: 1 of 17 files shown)


superleo:
hi @maobaolong would you mind to update the design docs to match this refactor? They still describe the old locations and registration model,  or follow up in a seperate PR?

maobaolong:
@superleo Thanks for your review, the relevant documents have been updated.

lmcache/v1/platform/base_ipc_wrapper.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Subclasses implement ``__init__`` (populate the interface fields from a
+    tensor) and ``to_tensor`` (reconstruct the tensor from the handle).
+    """
+
+    #: Set ``True`` on the subclass that should be used as the default
+    #: factory for a given ``device_type``.  Auto-discovery
+    #: (:func:`~lmcache.v1.platform._registry._discover_wrappers_once`)
+    #: skips subclasses where this is ``False``.
+    _is_default_wrapper: ClassVar[bool] = False
```
zxue2: @maobaolong could you pls check if   ```_is_default_wrapper: ClassVar[bool] = False``` can be removed  from ```DeviceIPCWrapper``` and keep ```_is_default_wrapper``` only on concrete subclasses.  It seems no need to have  ```_is_default_wrapper```  in base-class.  Similarly there is no ```device_type```.

 btw, the registry already treats a missing attribute as False ```getattr(cls, "_is_default_wrapper", False)```

maobaolong:
@zxue2 Good catch, dropped `_is_default_wrapper` from the base class. About `device_type`, the base class doesn't declare it either, so we're already in the same shape. Thanks!

lmcache/v1/platform/base_ipc_wrapper.py
maobaolong: done

maobaolong:
@hlin99 @chunxiaozheng Would you like to help to take another look?

lmcache/v1/platform/_registry.py
```diff
@@ -101,30 +203,60 @@ def get_kv_wrapper_factory(device_type: str) -> Callable[..., Any]:
...
     """
+    global _WRAPPERS_DISCOVERED
     _KV_WRAPPER_FACTORIES.clear()
     _KV_WRAPPER_FACTORIES.update(state.get("kv_wrapper", {}))
     _AVAILABILITY.clear()
     _AVAILABILITY.update(state.get("availability", {}))
+    _WRAPPERS_DISCOVERED = bool(state.get("discovered", False))
+
+
+def reset_for_tests() -> None:
```
hlin99, line 244: is this test function? looks no caller

hlin99 (approved):
one nit, otherwise LGTM

lmcache/v1/platform/_registry.py
maobaolong: yeah, it will used for flaky ut.

chunxiaozheng (approved):
LGTM

PR #3832, opened 2026-06-23
fix: bounds-check S3L2Adapter GET to prevent writing past MemoryObj (#3831)

### LINKED ISSUE #3831: S3L2Adapter GET callback can write past MemoryObj bounds on oversized responses
**Describe the bug**
`S3L2Adapter._get_request()` copies S3 response chunks into a caller-provided `MemoryObj` with:

```python
ctypes.memmove(data_ptr + offset, chunk, len(chunk))
```

The callback does not verify that the write range is inside the destination buffer:

```python
0 <= offset
offset + len(chunk) <= mem_obj.get_size()
```

As a result, if an S3 or S3-compatible endpoint returns an object body larger than the local LMCache layout expects, the GET callback can write past the logical `MemoryObj` bounds.

The non-MP `S3Connector.get()` path already performs the expected object-size check before downloading:

```python
if obj_size != memory_obj.get_size():
    memory_obj.ref_count_down()
    return None
```

The S3 L2 adapter lookup path caches HEAD `Content-Length`, but only treats it as an existence signal; it does not compare that length with the requested layout size or the destination `MemoryObj.get_size()` before `submit_load_task()` issues the GET.

Affected paths:

- `lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py::_get_request`
- `lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py::_execute_lookup`
- `lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py::_execute_load`
- `lmcache/v1/storage_backend/connector/s3_connector.py::get` has the safer comparison behavior for reference

**To Reproduce**
Use the following non-crashing harness from the audit workspace:

[repro_s3_l2_memmove_bounds.py](https://github.com/user-attachments/files/29235376/repro_s3_l2_memmove_bounds.py)

```bash
python ./LMCache/repro_s3_l2_memmove_bounds.py
```

The harness loads the real `s3_l2_adapter.py` file and stubs only unavailable local dependencies. It creates a fake `MemoryObj` with logical size 8 bytes, but gives it a larger backing buffer with a guard region after byte 8. It then invokes the real `on_body` callback with a 16-byte chunk.

Observed output:

```text
reported_size=8
callback_chunk_len=16
guard_before=b'GGGGGGGG'
guard_after=b'AAAAAAAA'
guard_overwritten=True
```

This avoids corrupting unrelated process memory while demonstrating that the callback writes beyond `MemoryObj.get_size()`.

**Expected behavior**
`S3L2Adapter` should reject object data whose size does not match the expected destination size, and the load bitmap should not mark the object as successfully loaded.

Recommended behavior:

- Compare remote/cached `Content-Length` with the expected layout size or `mem_obj.get_size()` before downloading.
- In `on_body`, reject negative offsets and any write where `offset + len(chunk) > mem_obj.get_size()`.
- Ensure the request future fails and `_execute_load()` leaves the corresponding bitmap bit unset on mismatch.

Example defensive check:

```python
expected = mem_obj.get_size()

def on_body(chunk, offset, **kwargs):
    end = offset + len(chunk)
    if offset < 0 or end > expected:
        raise RuntimeError(
            f"S3 GET size mismatch for {key_str}: write [{offset}, {end}) "
            f"outside destination buffer of {expected} bytes"
        )
    ctypes.memmove(data_ptr + offset, chunk, len(chunk))
```

**Screenshots**
Not applicable.

 - OS: Ubuntu 24.04.3 LTS under WSL2
 - Browser: Not applicable
 - Version: LMCache `dev` branch, commit `47517888506352d11f9c2eadd48c145f96ce6f8c`

 - Device: Not applicable
 - OS: Not applicable
 - Browser: Not applicable
 - Version: Not applicable

**Additional context**
This issue affects the S3 L2 adapter, which is a documented multiprocess L2 backend configured with:

```bash
--l2-adapter '{"type": "s3", "s3_endpoint": "s3://my-bucket", "s3_region": "us-west-2"}'
```

A practical mismatch can happen when the configured S3-compatible endpoint is compromised/misconfigured, when another writer replaces an object under the same key, or when stale objects from a different LMCache layout/configuration remain in the bucket.

The safer fix should include both a pre-download size comparison and an in-callback bounds check, so that even unexpected callback offsets or chunking behavior cannot write outside the target `MemoryObj`.

HaozheZhang6:
`S3L2Adapter._get_request`'s body callback `memmove`'d S3 chunks into the destination `MemoryObj` without checking the write stayed in bounds, so a response larger than the local layout expects could write past the buffer. The non-MP `S3Connector.get` already guards this with an object-size check; the L2 adapter cached the HEAD `Content-Length` but only treated it as an existence signal.

The fix compares the cached size against the destination `get_size()` before issuing the GET (same shape as `S3Connector.get`) and drops out-of-bounds bytes in the body callback as a backstop, with `on_done` failing the request so the load reports a miss rather than corrupting memory.

Two regression tests cover both paths -- rejected via the cached size after a lookup, and via the body-callback guard with no cached size.

Fixes #3831.

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
@@ -870,9 +870,16 @@ class S3L2Adapter(L2AdapterInterface):
     def _get_request(self, key_str: str, mem_obj: MemoryObj):
         req = self._make_request("GET", key_str)
         data_ptr = mem_obj.data_ptr
+        buf_size = mem_obj.get_size()
+        overflow = {"hit": False}
 
         def on_body(chunk, offset, **kwargs):
-            # Write chunk into the caller-provided MemoryObj buffer.
+            # Write chunk into the caller-provided MemoryObj buffer, dropping
+            # any bytes past its bounds so an oversized S3 object can't write
+            # out of bounds; on_done fails the request if that happens (#3831).
+            if offset < 0 or offset + len(chunk) > buf_size:
+                overflow["hit"] = True
+                return
             ctypes.memmove(data_ptr + offset, chunk, len(chunk))
 
         def on_done(error=None, status_code=None, **kwargs):
@@ -880,6 +887,11 @@ class S3L2Adapter(L2AdapterInterface):
             if error or not ok:
                 raise RuntimeError(
                     f"S3 GET failed for {key_str}: {error or status_code}"
+                )
+            if overflow["hit"]:
+                raise RuntimeError(
+                    f"S3 GET for {key_str} returned more data than the "
+                    f"{buf_size}-byte destination buffer"
                 )
 
         s3_req = s3.S3Request(
@@ -1201,6 +1213,15 @@ class S3L2Adapter(L2AdapterInterface):
         for i, (key, obj) in enumerate(zip(keys, objects, strict=True)):
             try:
                 key_str = _object_key_to_string(key)
+                with self._lock:
+                    cached_size = self._object_size_cache.get(key_str)
+                if cached_size is not None and cached_size != obj.get_size():
+                    logger.error(
+                        f"Size mismatch for {key_str}: S3 has {cached_size} "
+                        f"bytes, but the destination expects {obj.get_size()} "
+                        f"bytes; skipping load to avoid writing past the buffer."
+                    )
+                    continue
                 s3_req = self._get_request(key_str, obj)
                 futures.append(asyncio.wrap_future(s3_req.finished_future))
                 launched_indices.append(i)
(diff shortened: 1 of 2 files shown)


JiataiWang:
I tested this path against current `dev` and `awscrt` 0.36.2. There’s one callback-order issue worth fixing before this merges.

`S3Request._on_finish()` [completes `finished_future` before invoking `on_done`](https://github.com/awslabs/aws-crt-python/blob/main/awscrt/s3.py#L3352-L3440). In the current overflow branch, `on_body` only sets a flag and returns successfully, so CRT can complete the future successfully and `_execute_load()` can set the load bitmap bit. The later `RuntimeError` from `on_done` cannot change that future; the native bridge reports it as an unraisable callback exception instead. The test fake calls `on_done` before `set_result()`, which hides the real ordering.

Raising directly from `on_body` is the small fix: CRT captures body/header callback exceptions and completes `finished_future` exceptionally. Checking the successful GET `Content-Length` in `on_headers` also catches both undersized and oversized objects before copying, while the per-chunk range check remains the backstop.

@HaozheZhang6, are you still planning to refresh this PR? I have a guard-buffer regression for both size directions and a bad callback offset, and I’m happy to share it or help here rather than open a duplicate.

Issue #3753, opened 2026-06-18
Reduce bringup time with parallel and batched recovery slot-header validation


nayeonikim:
**Summary**
raw_block restart/recovery validates recovered slot headers one slot at a time, which dominates bringup time on large devices. Parallelize the header reads (per I/O engine) to cut bringup latency.

**Details**
On restart, RawBlockCore rebuilds its index from the checkpoint and then validates each entry by reading its on-device slot header (_validate_loaded_entries) — currently a serial, one-read-per-slot loop. We keep the validation logic and on-device format unchanged and only parallelize the reads, using each engine's native mechanism:

POSIX: a bounded pool of reader threads over contiguous entry ranges.
io_uring: a single batched_read + wait_iouring for all headers in a batch (non-uring_cmd).
The engine-specific part is confined to header reading; a single validation loop is shared across engines.

**Expected Outcome / Goal**
Lower bringup latency, with the gap over the serial path widening as entry count grows. A synthetic bringup benchmark (checkpoint-load + validation time) makes it reproducible.

**Actual Outcome**
Validation runs serially, so bringup time grows linearly with entry count and leaves device parallelism (POSIX threads / NVMe queue depth) unused.

**Additional Context**
- Work in progress as two complementary PRs (POSIX threadpool path + io_uring  `batched_read` path) plus the shared bringup benchmark;
-- Future work (out of scope here): batched header recovery for the `uring_cmd` (NVMe passthrough) path.

nayeonikim:
Being worked on by @nayeonikim  (io_uring path) and @daegyu94(POSIX path + benchmark).
cc @DongDongJu

daegyu94:
Our preliminary results were collected on a Samsung BM1743 61 TB SSD using an 8 TiB raw-block cache space with the POSIX engine:

```
threads=1:
  median=62.063691s
  mean=62.071091s
  samples=[62.184437, 62.063691, 61.965146]

threads=8:
  median=10.888474s
  mean=10.880261s
  samples=[10.929659, 10.888474, 10.822650]

Median speedup:
  5.70x
```

With eight reader threads, median restart recovery time decreased from approximately 62.1 seconds to 10.9 seconds, corresponding to a 5.70× speedup and an approximately 82.5% reduction in recovery time.

nayeonikim:
[...]

PR #3835, opened 2026-06-23
[Perf] Reduce bringup time with parallel and batched recovery slot-header validation

daegyu94:

This PR addresses #3753 by reducing RawBlockCore restart recovery time through engine-specific optimizations: multithreaded slot-header validation for the POSIX I/O engine and batched reads for the io_uring I/O engine.

During restart, recovery validates checkpoint entries against on-device slot headers. For multi-terabyte raw-block caches, this can require hundreds of thousands or millions of small reads, making serial validation a major contributor to bringup time.

*Recovery benchmark*

Performance was measured with a local synthetic recovery harness that writes checkpoint metadata and slot headers without full payloads.
The benchmark ran on a Samsung BM1743 NVMe SSD using an 8 TiB cache with 16 MiB slots, corresponding to approximately 524K recovered entries.

```bash
python benchmarks/storage_backend_io/raw_block_recovery_bringup_bench.py \
  --device-path /dev/nvmeXn1 --cache-space-gb 8192 --slot-bytes 16777216 \
  --measure --io-engine posix io_uring --threads 1 8 --repeats 5
```

| recovery path | median | speedup vs serial |
|---|---|---|
| `posix` serial (threads=1) | 70.59s | 1.00× |
| `posix` threadpool (threads=8) | 16.31s | 4.33× |
| **`io_uring` batched** | **11.55s** | **6.11×** |

Both optimizations substantially reduce recovery time: the 8-thread POSIX path improves over serial validation, while batched `io_uring` provides an additional ~1.4× speedup and reaches 6.11× over serial.

**Special notes for reviewers**

* `use_uring_cmd` (NVMe passthrough) batching is excluded because it requires separate low-level validation.
* If a batch read fails, recovery falls back to per-slot reads to identify the failed entries.

**If applicable**

* [x] This PR contains user-facing changes; documentation has been added.
* [x] This PR contains unit tests.


nayeonikim:
Hi, @DongDongJu @ApostaC, could you please take a look at this PR?
It speeds up raw-block restart recovery by parallelizing POSIX slot-header validation and batching regular io_uring header reads. Thanks!

lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -1876,29 +1882,19 @@ def _recover_checkpoint_dtype(
...
+        if not items:
+            return
+
+        offsets = [int(entry.offset) for _, entry in items]
+        headers = self._read_slot_headers(offsets)
+        to_drop = [
+            encoded_key
+            for (encoded_key, entry), slot_hdr in zip(items, headers, strict=False)
+            if self._is_stale_header(encoded_key, entry, slot_hdr)
+        ]
```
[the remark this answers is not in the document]
nayeonikim: @daegyu94, could you take a look at this suggestion and decide whether we should apply it? This line comes from the original POSIX recovery change.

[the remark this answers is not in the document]
daegyu94: Resolved!

daegyu94:
Hi @DongDongJu @ApostaC  @zhengfeihe ,
just following up on this PR. When you have a chance, could you please take a look? Thanks!

DongDongJu:
will take a look tomorrow morning. Thanks!

lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -2553,6 +2550,159 @@ def _validate_loaded_entries(self) -> None:
...
+        offsets: list[int],
+    ) -> list[Optional[tuple[int, int]]]:
+        """Return slot headers using a bounded POSIX reader thread pool."""
+        n = len(offsets)
+        max_workers = min(self._recovery_read_threads, n)
+        ranges = self._build_recovery_item_ranges(n, max_workers)
+        work_items = [(offsets, start, end) for start, end in ranges]
+        results: list[Optional[tuple[int, int]]] = [None] * n
+
+        def read_range(
```
zhengfeihe: Could we avoid the nested function here? Also, pool.map already yields results in input order, so I think we can simplify the logic here a bit.

zhengfeihe (commented):
Otherwise LGTM. (Apologize for the late review 🙏

benchmarks/storage_backend_io/raw_block_recovery_bringup_bench.py
```diff
@@ -0,0 +1,353 @@
+# SPDX-License-Identifier: Apache-2.0
```
DongDongJu: Should we need this benchfile here?
I think just for rawblock is too narrow for the purpose of folder.
either combine this into as as func in to storage_backend_io_benchmark.py or suggest to remove this.

DongDongJu (approved):
Hello @daegyu94,
LGTM functionally. I left one suggestion.
The POSIX path keeps recovery work bounded, and the io_uring path correctly handles batching, ordering, and per-read fallback.
I agree with simplifying the range helper, but please retain chunked tasks rather than creating one future per slot.

lmcache/v1/storage_backend/raw_block/core.py
daegyu94: Good point! I moved the range reader to _read_slot_header_range() and now flatten the ordered pool.map() results directly, while keeping bounded chunked work items.

benchmarks/storage_backend_io/raw_block_recovery_bringup_bench.py
daegyu94: Agreed. I removed the standalone benchmark and README section.

zhengfeihe (approved):

PR #3836, opened 2026-06-23
feat(valkey): add per-key TTL + partial-chunk validation to Valkey glide connector

omerrubi-amzn:

Two related improvements to the Valkey GLIDE connector (`valkey://`), stacked as separate commits:

1. **Optional per-key TTL (feature-flagged).** Today the connector writes keys with no expiry. A Valkey/Redis node configured with a `volatile-lru` / `volatile-lfu` eviction policy only evicts keys that have a TTL, so it never reclaims the KV-cache keys and the L2 remote cache chokes once it hits `maxmemory`. This adds an opt-in TTL so the eviction policy can reclaim memory under pressure. Two new `extra_config` keys:
   - `valkey_enable_ttl` (bool, default `false`) — feature flag.
   - `valkey_ttl_sec` (int seconds, default `86400` when enabled but unset) — per-key TTL; must be positive.

   When enabled, keys are written via `glide_sync.ExpirySet(ExpiryType.SEC, ttl)` (built lazily and cached; the immutable `ExpirySet` is safe to share across worker threads). The no-TTL path is unchanged, so existing deployments are unaffected until they opt in.

2. **Reject partial / unfull chunks.** The connector uses single-key, fixed-size storage (same scheme as `RESPConnector`) with no per-chunk metadata, so partial/unfull chunks cannot be represented or verified on read. The adapter now asserts `save_chunk_meta` and `save_unfull_chunk` are both `False`, mirroring `RESPConnectorAdapter`. Also coerces free-form `extra_config` booleans via `_to_bool` (so a string like `"false"` parses correctly — `bool("false")` is truthy), applied to `save_chunk_meta` and the existing `tls_enable` read.

- The two changes are independent commits; happy to split into separate PRs if preferred.
- TTL default behavior is unchanged (feature off). With TTL on, a key can expire between a positive `contains`/`batched_async_contains` and the subsequent `get`; this degrades gracefully to a cache miss / shorter consecutive prefix (handled in `_batched_get_non_blocking`), with no stale data.
- All commits are DCO signed-off.

**If applicable**:

- [x] this PR contains user facing changes - docs added (`docs/source/kv_cache/storage_backends/valkey.rst`)

### DIFF
--- a/docs/source/kv_cache/storage_backends/valkey.rst
+++ b/docs/source/kv_cache/storage_backends/valkey.rst
@@ -47,6 +47,12 @@ The following ``extra_config`` keys are supported:
    * - ``valkey_database``
      - None
      - Database ID (standalone mode only, ignored in cluster mode).
+   * - ``valkey_enable_ttl``
+     - ``false``
+     - Feature flag. When ``true``, every key is written with an expiry (see ``valkey_ttl_sec``) so Valkey/Redis ``volatile-*`` eviction policies can reclaim L2 cache keys once the node reaches ``maxmemory``. When ``false`` (default), keys are persisted without a TTL.
+   * - ``valkey_ttl_sec``
+     - ``86400``
+     - Key TTL in seconds, applied only when ``valkey_enable_ttl`` is ``true``. Must be a positive integer. If the flag is enabled but this key is omitted, it defaults to ``86400`` (24 hours).
    * - ``request_timeout``
      - ``5.0``
      - GLIDE request timeout in seconds. Also used as the Python-side Future timeout.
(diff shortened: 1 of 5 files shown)


sammshen (approved):
LGTM

DongDongJu (approved):
Hello @omerrubi-amzn,
Thanks for your contribution. direction for both your feature make sense to me.
general question I have for this, how we can define the TTL for each key? and its meaningful? I feel like granularity for policy it bit too fine-grain.

omerrubi-amzn:
@DongDongJu 

> how we can define the TTL for each key? and its meaningful? I feel like granularity for policy it bit too fine-grain.

Good question! To clarify — the TTL is **not** per-key in the sense of different expiry times for different keys. It's a single global value (`valkey_ttl_sec`) applied uniformly to **all** keys written by this connector instance. So the operator sets one number (e.g. `3600`) and every KV cache chunk gets that same TTL stamped at write time.

The reason we stamp each key individually (rather than relying solely on a global eviction policy) is that Valkey's `volatile-lru` / `volatile-lfu` policies **only evict keys that carry a TTL**. Without it, the L2 cache keys are never candidates for eviction and the node chokes at `maxmemory`. An `allkeys-*` policy would work but it evicts indiscriminately — problematic if the Valkey instance is shared with other data.

So the granularity from the operator's perspective is coarse (one config knob for the whole connector), but it's *implemented* as a per-key expiry because that's what Valkey's eviction semantics require.

If a future use case needs differentiated TTLs (e.g. shorter TTL for prefill chunks vs. decode chunks), we could extend this, but for now the uniform TTL is the simplest thing that makes `volatile-*` eviction work correctly.

PR #3839, opened 2026-06-23
Fix _alloc_page_aligned_pinned_view()

zxue2:
Pin the host buffer in case of an accelerator is present as to fix the issue of "RuntimeError: level_zero backend failed with error: 20 (UR_RESULT_ERROR_DEVICE_LOST)" during running "pytest tests/benchmarks/test_xpu_connector_benchmark.py".

**If applicable**:

### DIFF
--- a/lmcache/python_ops_fallback.py
+++ b/lmcache/python_ops_fallback.py
@@ -20,7 +20,8 @@ import numpy as np
 import torch
 
 # First Party
-from lmcache import torch_dev
+from lmcache import torch_dev, torch_device_type
+from lmcache.logging import init_logger
 
 # Store the tensor objects in memory so that they can be accessed
 # outside the scope of this file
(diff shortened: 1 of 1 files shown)


hlin99:
hi, the change overall makes sense. 
i'm looking at the same issue recently. I have two small questions
1. Why we only pin memory when acc is available? according to the function semantic, pin this cpu memory should be there regardless of acc status.
2. i'm not sure if on any device, pin host memory will crash. to make this fallback device agnostic as much as possible, is it good idea to always try pin memory, and if any time there's exception, we record it and moving forward set pin=False. this is kind of auto detection

zxue2:
> hi, the change overall makes sense. i'm looking at the same issue recently. I have two small questions
> 
> 1. Why we only pin memory when acc is available? according to the function semantic, pin this cpu memory should be there regardless of acc status.
> 2. i'm not sure if on any device, pin host memory will crash. to make this fallback device agnostic as much as possible, is it good idea to always try pin memory, and if any time there's exception, we record it and moving forward set pin=False. this is kind of auto detection

@hlin99 ,   one exception is that StubCPUDevice(device_type=cpu)  is_available() always returns False for now. I add this in comments.  Agree all accelerator devices should support pinned mem here otherwise it should be a bug of that device that shall be addressed first in torch dev.,
However, we can also have a fallback mechanism to cache one-shot decision for robustness, i.e.,  pin host buffers only when an accelerator is present. Probed lazily on first allocation; if a pinned allocation ever fails at runtime we flip this to False permanently and fall back to pageable memory for all subsequent allocations.

hlin99 (approved):
LGTM

zxue2:
@maobaolong  could you pls have a look? thx

maobaolong (approved):
LGTM

PR #3841, opened 2026-06-23
[Fix][RawBlock] Align uring_cmd temporary buffers

3xdevv:

Raw-block `io_uring_cmd` transfers already align device offsets and transfer lengths, but the temporary Python buffers used for padded writes and read copy-back were plain `bytearray` allocations. Those allocations do not guarantee a block-aligned userspace address, so the NVMe passthrough path can hand the kernel a transfer whose range is aligned but whose userspace buffer address is not. In that case, the kernel/DMA mapping path may need an additional copy or bounce-style handling before the device can consume the buffer. Keeping these temporary buffers aligned avoids relying on lower layers to compensate for unaligned userspace memory.

- Adds a `RawBlockCore` helper that returns a block-aligned writable memoryview backed by an overallocated bytearray.
- Uses the aligned temporary buffer for padded `io_uring_cmd` writes when the source view is shorter than the aligned transfer length.
- Uses the same aligned temporary buffer for `io_uring_cmd` reads that need a full-transfer target before copying the logical payload back to the caller.
- Adds focused raw-block core tests that verify the uring-cmd write and read temporary buffers are block-aligned while preserving payload and padding behavior.

Tested by:

```bash
pytest -xvs tests/v1/storage_backend/test_raw_block_core.py
```

**If applicable**:


DongDongJu (approved):
LGTM

3xdevv:
Hello, @zhengfeihe. This is another low-risk change for rawblock. Could you take a look when you have time? Thanks!

zhengfeihe (approved):
LGTM!

PR #3843, opened 2026-06-23
[core] Add AMD hipFile backend for the GDS L1 tier (MP mode)

andyluo7:

The GDS L1 slab-file tier added in #3589 reaches storage via NVIDIA cuFile (ctypes against `libcufile.so`). This PR adds the AMD ROCm analog, [ROCm/hipFile](https://github.com/ROCm/hipFile) ("AMD Infinity Storage"), so the GDS L1 tier works on AMD GPUs. hipFile mirrors the cuFile async surface 1:1 (every symbol the tier uses has a `hipFile*` equivalent since ROCm 7.2.0), so this is a backend mirror rather than a redesign.

Changes:
- **`_hipfile_async.py`** (new) — ctypes wrapper over `libhipfile.so` mirroring `_cufile_async`'s surface (`register_buffer`/`register_stream`/`register_handle`/`AsyncHandle`/`Submission`). hipFile ships no Python binding, so the driver lifecycle, handle/buffer/stream registration, and the `hipFileError_t` / `hipFileDescr_t` structs are bound directly via ctypes.
- **`_gds_async.py`** (new) — dispatch shim that re-exports the cuFile or hipFile backend based on `torch.version.hip`. `gds_context` imports it as `ca` and is now platform-agnostic.
- **`_cufile_async.py`** — factored `cuFileHandleRegister`/`Deregister` into `register_handle`/`deregister_handle` so both backends expose an identical surface. No behavior change on NVIDIA.
- **`gds_context.py`** — registers the slab handle through the dispatch shim instead of importing `cufile.bindings` directly.

- The cuFile path is functionally unchanged; the only refactor is moving handle (de)registration behind `register_handle`/`deregister_handle`, which the existing `AsyncHandle` now also uses.
- The `requires_gds` guard in `test_gds_context.py` was generalized to `_gds_available()` (CUDA + nvidia-fs, **or** ROCm + a loadable `libhipfile.so`). The existing roundtrip tests are backend-agnostic and exercise hipFile on AMD unchanged.
- hipFile must be built with `amdclang++` and C++20 (`std::bit_cast`); only Ubuntu 24.04 / el9 / openSUSE packages are published today.

**Validation** — built `libhipfile.so` from source and ran on two AMD nodes:

| | MI350X (gfx950, ROCm 7.2) | MI300X (gfx942) |
|---|---|---|
| `test_hipfile_async.py` | 19/19 pass | 19/19 pass |
| GDS roundtrip (write/read, two-stream, >16 MiB split) | 3/3 pass | 3/3 pass |
| hipFile mode | compat | compat |

The roundtrips run the full `GDSContext → _gds_async → _hipfile_async → libhipfile.so → GPU↔slab` path including a position-dependent pattern across the 16 MiB region boundary, confirming data integrity on both architectures. Both nodes ran in hipFile **compat mode**; the zero-copy P2PDMA fast path is pending bare-metal hardware with `CONFIG_PCI_P2PDMA` + `amdgpu-dkms >= 30.20.1` + a local ext4/xfs NVMe.

**If applicable**:

### DIFF
--- a/docs/source/mp/configuration.rst
+++ b/docs/source/mp/configuration.rst
@@ -223,10 +223,23 @@ GDS L1 Tier
 Source: ``lmcache/v1/distributed/config.py``
 
 Opt-in. Setting ``--gds-l1-path`` switches the L1 medium from pinned DRAM to
-an NVMe slab file accessed via GPUDirect Storage (cuFile DMA). The CPU
-pinned-DRAM tier is then disabled, and ``--l1-size-gb`` sizes the slab.
-Disable byte-array L2 adapters when this is on (the GDS tier exposes no L1
-memory buffer for them to register).
+an NVMe slab file accessed via GPUDirect Storage DMA. The CPU pinned-DRAM tier
+is then disabled, and ``--l1-size-gb`` sizes the slab. Disable byte-array L2
+adapters when this is on (the GDS tier exposes no L1 memory buffer for them to
+register).
+
+The DMA path is selected automatically by platform: **cuFile**
+(``libcufile.so``) on NVIDIA and **hipFile** (``libhipfile.so``,
+`ROCm/hipFile <https://github.com/ROCm/hipFile>`_) on AMD ROCm. The same
+flags apply to both; no configuration change is needed to switch vendors.
+
+.. note::
+
+   AMD hipFile requires ROCm >= 7.2.0. The zero-copy GPUDirect fast path
+   additionally needs a kernel built with ``CONFIG_PCI_P2PDMA``,
+   ``amdgpu-dkms >= 30.20.1``, and the slab on a local NVMe ext4/xfs
+   filesystem; where those are unavailable hipFile transparently falls back to
+   a host-bounce compatibility path (correct, but not zero-copy).
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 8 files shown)


andyluo7:
Thanks @gemini-code-assist — addressed in 8cde7e5:

- **Thread-safe lazy init**: added a module `_init_lock` guarding the one-time `libhipfile` dlopen and the driver open/close. Used double-checked locking so the per-DMA hot path (`_lib()` once loaded, called on every `read_async`/`write_async`) stays lock-free, and `_ensure_driver_open`/`close_driver` call `_lib()` *before* taking the lock to avoid re-entering the non-reentrant lock.
- **Docstrings**: added `Args:`/`Returns:`/`Raises:` to the public `register_*`/`deregister_*` functions.
- **`__exit__` type hints**: added.

**Declining the `Submission.get_refs()` suggestion**, for these reasons:
- The SLF (private-member-access) rule is intentionally disabled for this path — `pyproject.toml` enables SLF only under `lmcache/v1/multiprocess/**` and `lmcache/v1/distributed/**`, not `lmcache/v1/gpu_connector/**`.
- `Submission` and `AsyncHandle` live in the **same module**, and this mirrors the existing `_cufile_async.py` (its `AsyncHandle` accesses `sub._size`/`_bytes_done` the same way). The two backend wrappers are meant to stay structurally identical so they diff cleanly; adding a `get_refs()` indirection to only one would diverge them for no enforcement benefit.

Re-validated on MI350X (gfx950, ROCm 7.2) after these changes: 19/19 unit + 3/3 GDS roundtrip tests still pass.

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3853, opened 2026-06-24
[GPU connector] Allow non-zero storage_offset in dim-0-padded layout (MTP + CPU offload)

### LINKED ISSUE #3821: Crash with offload + MTP/spec-decode together: `storage_offset != 0` in `_validate_dim0_padded_layout`
## What I hit

When I enable **CPU KV offload** (`LMCacheMPConnector`) **and MTP / speculative decoding** at the same time, the engine crashes during KV-cache registration:

```
ValueError: attempt_permute_to_contiguous_view: tensor is non-contiguous and not a
  supported (dim-0 padding only) layout — storage_offset != 0 (slice/narrow view, base address shifted).
```

CPU offload on its own works, and MTP on its own works — it's only the **combination** that fails.

## Environment

- vLLM `v0.23.1rc1.dev206+g93bad1191` (nightly `93bad119…`)
- LMCache `0.4.8rc4` (this connector code looks the same in `0.4.7`)
- Model: `deepseek-ai/DeepSeek-V4-Flash` (sparse-MLA hybrid), single B300, TP=1
- `LMCacheMPConnector`, `kv_role=kv_both`, CPU L1 only

## Reproduce

Serve with offload **and** MTP enabled:

```
vllm serve <model> --tensor-parallel-size=1 --kv-cache-dtype=fp8 ... \
  --speculative-config '{"method":"mtp","num_speculative_tokens":2}' \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both"}'
```

Drop `--speculative-config` (offload only) **or** `--kv-transfer-config` (MTP only) and it boots fine.

## What seems to be going on

With MTP enabled, the KV cache ends up laid out as a **single shared pool** where each per-layer view starts at a **non-zero `storage_offset`** (the draft shares the target's buffers, so the per-layer caches become slice/`narrow` views). Without MTP, the per-layer caches are `storage_offset == 0` dim-0-padded views, which the connector already supports.

`_validate_dim0_padded_layout` (`lmcache/v1/gpu_connector/utils.py`) rejects any non-zero offset:

```python
if storage_offset != 0:
    _fail("storage_offset != 0 (slice/narrow view, base address shifted)")
```

## What fixed it for me

I changed that one check to allow a non-zero `storage_offset` (kept every other dim-0-padding invariant):

```python
# _validate_dim0_padded_layout
-    if storage_offset != 0:
-        _fail("storage_offset != 0 (slice/narrow view, base address shifted)")
```

After that, offload + MTP boot and serve together, and it's been running fine: KV offloads to CPU and loads back on hits, no store errors, and MTP keeps accepting drafted tokens normally (which made me fairly confident the restored KV is correct — the draft reads the same offloaded KV).

When I poked at why this seemed OK rather than just silencing the check:

- `CudaIPCWrapper` already captures `storage_offset` and rebuilds the view with it (`set_(storage, storage_offset, shape, stride)`).
- `get_group_data_ptrs` passes `tensor.data_ptr()` to the transfer kernel, which already includes the offset (`storage().data_ptr() + storage_offset * itemsize`).
- The CPU/SHM tier keeps its own separate `storage_offset == 0` assert, so it's still guarded.

I am curious if I am missing something. Since this check was clearly added on purpose, I'd rather ask than assume: is it safe to relax this in general, or are there transfer paths where a non-zero `storage_offset` would read/write from the wrong place? I only exercised the single-GPU MLA per-layer path — I did **not** test **GDS**, **NIXL**, the **`blend_v3`** module, or the **SGLang / TRT-LLM** cache formats.

nafis271:
## Problem
`_validate_dim0_padded_layout` rejects `storage_offset != 0`, which blocks **MTP speculative decoding together with LMCache CPU offload** on vLLM. With MTP enabled, per-layer KV are non-zero-`storage_offset` slices of a unified pool, so registration fails:
```
ValueError: attempt_permute_to_contiguous_view: ... storage_offset != 0 (slice/narrow view, base address shifted).
shape=(90211, 64, 132), stride=(1039680, 132, 1), storage_offset=8640, dtype=torch.uint8
```
Issue: https://github.com/LMCache/LMCache/issues/3821

## Why it looks safe
From what I can tell, the non-zero offset is carried through on the GPU transfer path:
- `CudaIPCWrapper` captures `storage_offset` and rebuilds the view via `set_(storage, storage_offset, shape, stride)`.
- `get_group_data_ptrs` passes `tensor.data_ptr()` (= `storage().data_ptr() + storage_offset*itemsize`) to the transfer kernel, so `block_stride_elems` is applied from the correct base.
- The CPU/SHM tier keeps its own independent `storage_offset == 0` assert, so the relaxation stays scoped to the GPU path.

## Validation (current `dev` / v0.5.0)
DeepSeek-V4-Flash + MTP + CPU offload on lmcache **0.5.0**:
- **Unpatched** → crashes at `kv_format/contiguity.py:133` with the tensor above.
- **Patched** → all 168 per-layer KV tensors register (`Wrapping 168 KV cache tensors`), the engine boots and serves, 0 store errors.

Also validated under real production mirror traffic: ~1M tokens round-tripped through CPU L1, MTP acceptance ~2.5.

cc @zhengfeihe @liuyumoye

Fixes #3821


lmcache/v1/gpu_connector/kv_format/contiguity.py
```diff
@@ -90,7 +90,9 @@ def _validate_dim0_padded_layout(tensor: torch.Tensor) -> int:
     * Every interior dim ``i`` satisfies
       ``stride[i] == prod(shape[i+1:])`` -- only dim-0 may carry
       padding, with ``stride[0] >= prod(shape[1:])``.
-    * ``storage_offset == 0`` -- no slice/narrow base shift.
+    * ``storage_offset`` may be non-zero (slice/narrow view): the IPC
+      wrapper carries it verbatim and downstream uses ``data_ptr()``, so
+      the padded layout is read from the correct base.
```
ApostaC: Since we removed the validation of the storage offset here. We should just drop this in the docstring.

lmcache/v1/gpu_connector/kv_format/contiguity.py
```diff
@@ -129,8 +131,13 @@ def _fail(reason: str) -> None:
...
         _fail("stride[-2] != shape[-1] (last-two dims not tightly packed)")
-    if storage_offset != 0:
-        _fail("storage_offset != 0 (slice/narrow view, base address shifted)")
+    # A non-zero storage_offset (slice/narrow view -- e.g. an MTP draft's shared
+    # KV, or per-layer slices of a unified KV pool) is safe here: CudaIPCWrapper
+    # round-trips storage_offset verbatim and get_group_data_ptrs hands the
+    # transfer kernel tensor.data_ptr() (offset already baked in), so the dim-0
+    # padding (block_stride_elems) is applied from the correct base. The CPU/SHM
+    # tier keeps its own storage_offset == 0 assert, so this stays GPU-path only.
+    # See https://github.com/LMCache/LMCache/issues/3821
```
ApostaC: Same as above. We can drop these comments as well.

ApostaC (commented):
Thanks for the fix! Please update the docstring.
Otherwise LGTM!

ApostaC (commented):
Also, can we have a unit test to verify whether the current CUDAIPCWrapper supports storage_offset > 0?

nafis271:
Thanks for the review @ApostaC! The latest commit should address your review!

ApostaC (approved):
LGTM!

deng451e (approved):
LGTM!

PR #3857, opened 2026-06-24
feat(prefetch): speculative (predictive) prefetch for the L2->L1 path

nevasini1:
## Summary

LMCache's `PrefetchController` is **reactive** — a caller hands it explicit keys via `submit_prefetch_request(...)` and it loads them. Nothing **predicts** which keys a future request will need, so the first request for an uncached prefix always pays full L2 load latency.

This adds the **predictive ("speculative")** half: learn the request access stream and predict the next likely keys, so they can be warmed into L1 ahead of need. It's the software-portable idea from **CXL-SpecKV: A Disaggregated FPGA Speculative KV-Cache for Datacenter LLM Serving** ([arXiv:2512.11920](https://arxiv.org/abs/2512.11920)), adapted to LMCache's tiered storage. (The paper's other two techniques already map onto LMCache: CXL disaggregation via the Maru backend; KV compression via the SERDE path.)

## What's added

- **`SpeculativePrefetcher`** (`storage_controllers/speculative_prefetcher.py`) — a **dependency-free, generic** online predictor of the next likely cache keys. It blends a **recency-decayed first-order Markov successor model** (`P(next=B | last=A)`) with a small **popularity prior** — the lightweight stand-in for the paper's LSTM sequence predictor. `observe()/observe_sequence()` learn; `predict()/predict_keys()` return ranked guesses with `[0,1]` confidence. Amortized O(1)/O(out-degree) updates; bounded memory (`max_sources` eviction + decayed-weight pruning).
- **`PrefetchController` wiring (optional, default-off)** — a `speculator` constructor param; `submit_prefetch_request` feeds the access stream to the predictor (lock-guarded, since submission is multi-threaded); `predict_prefetch_keys()` exposes ranked predictions; `report_status()` surfaces enablement + model size. **With `speculator=None` there is zero behavior change and zero overhead.**
- **Config + CLI** — `enable_speculative_prefetch` (default `False`) plus `speculative_prefetch_max_keys` / `_min_confidence` / `_decay`, with matching `--enable-speculative-prefetch` / `--speculative-prefetch-*` flags; `StorageManager` builds and injects the predictor only when enabled.
- **Design doc** — `docs/design/v1/distributed/storage_controllers/speculative_prefetch.md`.

## Design boundary (deliberate)

This implements and wires the **prediction** half end-to-end but does **not** auto-issue speculative loads inside the controller. `PrefetchController` read-locks every loaded key and expects the submitter to release it via `query_prefetch_result(...)`; an internal load with no consumer would **leak L1 read-locks**. So predictions are surfaced via `predict_prefetch_keys()` for a caller to issue through the normal lock-owned `submit_prefetch_request(...)` path.

The natural follow-up — a thin driver that calls `predict_prefetch_keys()` and submits a capped speculative prefetch loaded as *temporary/non-retained*, plus an end-to-end hit-rate vs. wasted-bandwidth benchmark — is intentionally left to a separate PR so it can be validated against a running engine. The design doc spells this out.

## Testing

- `tests/v1/distributed/test_speculative_prefetcher.py` — exhaustive predictor coverage (Markov learning, context exclusion, no self-transitions, popularity fallback, recency decay, confidence filtering, bounded memory, deterministic ordering, arbitrary hashable keys). **Dependency-free → runs without torch**; verified locally (18/18 behaviors).
- `tests/v1/distributed/test_prefetch_controller.py` — wiring test: the controller trains the predictor from submitted requests and predicts the learned successor; and is a no-op without a speculator. (CUDA-gated, like the rest of that module → runs on GPU CI.)
- `ruff check`, `ruff format`, `isort`, `codespell`, `py_compile` all pass. Commit is DCO-signed.

## Notes for reviewers

This is opt-in and inert by default, so it's safe to merge as the predictive foundation; the load-driver follows. Happy to adjust the prediction signal (e.g. key off multi-turn `cache_salt`/session or a prefix-tree) or the integration surface based on your preference.

### DIFF
--- a/docs/design/v1/distributed/storage_controllers/speculative_prefetch.md
+++ b/docs/design/v1/distributed/storage_controllers/speculative_prefetch.md
@@ -1,0 +1,107 @@
+# Speculative Prefetch
+
+Predictive ("speculative") prefetching: guess which cache keys a future request
+will need and warm them into L1 from a slower L2 tier *before* the request
+arrives. This is the software-portable idea from **CXL-SpecKV: A Disaggregated
+FPGA Speculative KV-Cache for Datacenter LLM Serving**
+([arXiv:2512.11920](https://arxiv.org/abs/2512.11920)), adapted to LMCache's
+tiered storage. The paper's other two techniques map onto existing LMCache
+features and are out of scope here: CXL memory disaggregation is already
+available through the Maru backend (`storage_backend/maru_backend.py`), and
+hardware KV compression has a software analog in the existing SERDE path.
+
+## Motivation
+
+Today's `PrefetchController` is **reactive**: a caller hands it an explicit list
+of keys via `submit_prefetch_request(...)`, and it looks them up across L2
+adapters, plans loads, reserves L1, and loads. Nothing *predicts* which keys to
+prefetch ahead of an actual request — so the first request for a not-yet-cached
+prefix always pays the full L2 load latency. Predictive prefetch hides that
+latency for access patterns that are partially predictable: multi-turn chat
+continuations, RAG documents that co-occur, and the ordered chunks of one
+prompt.
+
+## Components
+
+### `SpeculativePrefetcher` (`speculative_prefetcher.py`)
+
+The predictor — the "brain". It is intentionally **generic over a hashable key
+type and imports nothing from the rest of LMCache**, so it is unit-testable in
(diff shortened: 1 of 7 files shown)


sammshen:
One potential other way of writing a speculative prefetcher is as an L2 Adapter

PR #3862, opened 2026-06-24
feat(MP): add securityContext to crd(lmcache.lmcache.ai_lmcacheengines.yaml) and modify controller

ghkdwlgns612:
## Summary
When running the `lmcache-server` DaemonSet Pod together with a vLLM Pod for IPC communication, we encountered a `cudaErrorMapBufferObjectFailed` error.

After investigation, we found that the vLLM Pod was running as the non-root `nobody` user for security reasons, while the `lmcache-server` DaemonSet Pod was running as `root`. Because of this user mismatch, the `lmcache-server` Pod could not properly read or write the IPC-related files created by the vLLM Pod.

To address this, this PR adds support for configuring `securityContext` when deploying the `lmcacheengines` CRD, so that the LMCache server Pod can run with the appropriate user settings for IPC communication.

## Details

The issue occurred when the `lmcache-server` DaemonSet Pod and the vLLM Pod were running with different users while communicating through IPC.

The vLLM Pod was running as the non-root `nobody` user for security reasons, and IPC-related files were created under `/dev/shm` with the following ownership and permissions:

```text
-rw------- 1 nobody nogroup 81920 cuda.shm.fffe.223.1
-rw------- 1 nobody nogroup 81920 cuda.shm.fffe.224.1
-rw------- 1 nobody nogroup 81920 cuda.shm.fffe.225.1
-rw------- 1 nobody nogroup 81920 cuda.shm.fffe.226.1
-rw------- 1 nobody nogroup 64424509440 lmcache_l1_pool_1
-rw------- 1 nobody nogroup 80064 torch_547_3673712610_0
-rw------- 1 nobody nogroup 80064 torch_548_3790871826_0
-rw------- 1 nobody nogroup 80064 torch_549_88063950_0
-rw------- 1 nobody nogroup 80064 torch_550_2061898546_0
```

Although these files were visible from the `lmcache-server` Pod through the shared `hostPath` mount for `/dev/shm`, the `lmcache-server` Pod was running as `root` and could not properly access the IPC files created by the vLLM Pod. As a result, LMCache failed during `REGISTER_KV_CACHE` while trying to unwrap the CUDA IPC handle:

```text
torch.AcceleratorError: CUDA error: mapping of buffer object failed
```

To resolve this, this PR adds security context-related fields to the `LMCacheEngine` CRD. It also updates the controller logic that watches the `LMCacheEngine` resource so that the configured `securityContext` is applied when creating the `lmcache-server` DaemonSet.

With this change, users can configure the `lmcache-server` Pod to run with the appropriate user and group settings required to access IPC-related files created by the vLLM Pod.

### DIFF
--- a/operator/api/v1alpha1/lmcacheengine_types.go
+++ b/operator/api/v1alpha1/lmcacheengine_types.go
@@ -372,6 +372,14 @@ type LMCacheEngineSpec struct {
 	// +optional
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 
+	// securityContext defines the container-level security context for the
+	// LMCache container.
+	//
+	// When unset, the operator preserves the existing default behavior and runs
+	// the container with privileged: true.
+	// +optional
+	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
+
 	// podAnnotations are additional annotations added to pods.
 	// +optional
 	PodAnnotations map[string]string `json:"podAnnotations,omitempty"`
--- a/operator/api/v1alpha1/zz_generated.deepcopy.go
+++ b/operator/api/v1alpha1/zz_generated.deepcopy.go
@@ -21,7 +21,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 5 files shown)


ghkdwlgns612:
@hickeyma  @deng451e  @sammshen  @ApostaC  Hello. I'm sorry but I can't seem to change my reviewer. Could you please review it? I know you're busy, but thank you.

Issue #3720, opened 2026-06-16
[Good second issue] Extending the LMCache CLI


chunxiaozheng:
# Overview
The LMCache CLI (lmcache) is designed as a plugin-style, auto-discovery architecture that supports N-level nested subcommands. It is a relatively independent module that can be extended with new commands or subcommands without modifying any existing registration code.

**If you are new to LMCache, you can get some help from https://github.com/LMCache/LMCache/issues/3372**

# Architecture
The CLI framework is built on two core abstractions in `lmcache/cli/commands/base.py`:
depend on the refactor merged(#3678)
| Class | Role |
|  ----  | ---- |
| `BaseCommand` | Abstract base class for leaf commands (commands that perform actual work) |
| `CompositeCommand` | A BaseCommand subclass that groups sub-subcommands by overriding register() to create nested subparsers |

# Auto-discovery mechanism
The CLI uses a unified, recursive auto-discovery model powered by `pkgutil.iter_modules` (via the `discover_subclasses` utility). The same mechanism applies at every level of the command tree:
## How it works (at any level):
1. A package's `__init__.py` defines either the top-level discovery (`commands/__init__.py`) or a `CompositeCommand` subclass.
2. At startup, `pkgutil.iter_modules` scans all direct submodules (files and sub-packages) of that package.
3. Each discovered module is imported, and all concrete `BaseCommand` subclasses are collected.
4. Modules prefixed with _ (e.g. `_helpers.py`, `_utils.py`) are excluded from the scan — use this convention for internal utilities.
- Level 1 — Top-level commands:
`lmcache/cli/commands/__init__.py` calls `discover_subclasses` on its own package. Every concrete `BaseCommand` subclass found (including `CompositeCommand` subclasses) is instantiated and stored in `ALL_COMMANDS`, then registered with the root argument parser in `main.py`. Adding a new `.py` file with a `BaseCommand` subclass requires zero edits to any existing file.
- Level 2+ — Subcommands within a group (`CompositeCommand`):
`CompositeCommand` is a `BaseCommand` subclass designed for commands that only group sub-subcommands. Its `register()` method automatically scans the package where it is defined, discovers all child `BaseCommand` subclasses in sibling modules, and registers them as nested subparsers. This means adding a new subcommand to an existing group also requires zero edits — just drop a new `.py` file (or sub-package) in the group's directory.
- Level N — Arbitrary nesting:
[...]

PR #3867, opened 2026-06-25
[CLI] Add `describe engine` command

a-m-n-s:

Adds `lmcache describe engine`, the engine-side counterpart to the existing
`lmcache describe kvcache`. It gives operators a one-shot status view of the
inference engine a running LMCache instance is paired with:

    $ lmcache describe engine --url http://localhost:8000

    ================ Inference Engine ================
    Model:                  meta-llama/Llama-3.1-8B-Instruct
    Max context (tokens):   131072
    Status:                 OK
    Running requests:       3

Data is read from the engine's own HTTP surface — model identity and context
window from `/v1/models`, liveness from `/health`, and in-flight load from the
Prometheus `vllm:num_requests_running` gauge on `/metrics`. The running-requests
lookup is best-effort: if metrics are disabled or unreachable it degrades to
`N/A` rather than failing the command. Only the primary `/v1/models` fetch exits
non-zero on failure. `--url` defaults to `http://localhost:8000` (the engine
default) and supports `--format json` / `--output` like the other describe
targets.

Refs #3720

**notes**

- The new `EngineDescriber` mirrors the existing `KVCacheDescriber`
  section-builder pattern, and the engine tests mirror the kvcache tests in the
  same file, so the structure should look familiar.
- Docs updated alongside the code: the user guide (`docs/source/cli/describe.rst`)
  gains an `engine` section, and the design doc
  (`docs/design/cli/commands/describe.md`) now documents the implemented command
  (data sources, fields, JSON shape, error handling) instead of marking it as
  stubbed.

**If applicable**:

**Follow-up**

While adding this command I ended up hand-rolling another `normalize_url` +
`urlopen` try/except, the same boilerplate that's already duplicated across the
other client commands. Filed #3868 to track pulling it into a shared
`lmcache/cli/http.py`; intentionally left out of this PR to keep the scope to
`describe engine`.

### DIFF
--- a/docs/design/cli/commands/describe.md
+++ b/docs/design/cli/commands/describe.md
@@ -9,7 +9,8 @@ and entry point are all working. The next step (Phase 1 per
 [commands.md](../commands.md)) is to implement `lmcache describe kvcache`, which
 provides a rich status dashboard of a running LMCache KV cache service.
 
-`describe engine` is Phase 2 scope and stubbed but not implemented here.
+`describe engine` (Phase 2) is now implemented — see the
+[`describe engine`](#lmcache-describe-engine) section below.
 
 ---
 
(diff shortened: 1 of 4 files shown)


KuntaiDu (approved):
LGTM! Thank you for your contribution!

ApostaC (approved):
LGTM!

PR #3869, opened 2026-06-25
[Prefetch sliding window 3/3] Prefetching / load only KV cache in sliding window 

KuntaiDu:

Prefetching / load only KV cache in sliding window 

**If applicable**:


lmcache/v1/kv_layer_groups.py
```diff
@@ -502,11 +502,12 @@ def get_subchunk_sw_size_tokens(self, kernel_group_idx: int) -> int:
...
             The sub-chunk sliding window size. Will be the same as the
-            chunk size for non-slding-window models or big-sliding-
-            window models.
+            chunk size for non-sliding-window models, big-sliding-window
+            models, and Mamba (``sw_size_tokens=0``) groups whose
+            recurrent state always occupies a full chunk.
         """
         sw_size_tokens = self._kernel_groups[kernel_group_idx].sw_size_tokens
-        if sw_size_tokens == -1 or sw_size_tokens >= self._lmcache_tokens_per_chunk:
+        if sw_size_tokens <= 0 or sw_size_tokens >= self._lmcache_tokens_per_chunk:
```
KuntaiDu: Can we do 
`elif sw_size_tokens == 0` instead of folding the branch, and comment that this branch is for mamba models?

lmcache/v1/kv_layer_groups.py
```diff
@@ -579,6 +584,8 @@ def _detect_object_groups(
         for kernel_group_idx, group in enumerate(self._kernel_groups):
             if group.sw_size_tokens == -1:
                 sw_size_chunks = -1
+            elif group.sw_size_tokens == 0:
```
KuntaiDu: Can we comment that this is for mamba?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -74,26 +75,47 @@ def build_trim_mask(
...
+        length in chunks and ``retain_mask`` is the bitmap of retained key
+        indices.
     """
     if policy is TrimPolicy.PREFIX:
-        return Bitmap(num_keys, found.count_leading_ones())
-    return found
+        stride = attn_desc.num_object_groups * attn_desc.world_size
+        num_chunks = num_keys // stride if stride else 0
+        windows = attn_desc.num_chunks_in_sw
+        if attn_desc.force_retrieve_full_kv:
```
KuntaiDu: Add a comment here to briefly explain that every group is forced to sliding window, and this flag is for benchmarking only

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -74,26 +75,47 @@ def build_trim_mask(
...
+            windows = [-1] * attn_desc.num_object_groups
+        hit_length, retain = fold_unfold_ranked(
+            found,
+            num_chunks,
+            attn_desc.world_size,
+            windows,
+        )
+        return hit_length, retain
+    stride = attn_desc.num_object_groups * attn_desc.world_size
+    hit_chunks = found.count_leading_ones() // stride if stride else 0
```
KuntaiDu: why stride can be 0?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -349,6 +380,7 @@ def submit_prefetch_request(
         extra_count: int = 0,
         policy: TrimPolicy = TrimPolicy.PREFIX,
         attn_desc: AttnWindowDesc = DEFAULT_ATTN_WINDOW_DESC,
+        group_layout_descs: dict[int, MemoryLayoutDesc] | None = None,
```
KuntaiDu: Why using a int dict to MemoryLayoutDesc? I thought MemoryLayoutDesc itself should contain the information of all object ids. In this case why do we map from obj id to memory layout desc?

KuntaiDu: Good question. `MemoryLayoutDesc` describes a single group's shapes + dtypes (the `shapes: list[torch.Size]` and `dtypes: list[torch.dtype]` are for the kernel groups *within* one object group). `L1Manager.reserve_write` takes a single `MemoryLayoutDesc` to allocate one MemoryObj.

When object groups have different tensor shapes (e.g. full-attention group has 10 layers × head_dim=512, SW group has 50 layers × head_dim=256), each key's MemoryObj needs the layout of its own group. So we need a mapping `group_id → MemoryLayoutDesc`.

Making `MemoryLayoutDesc` contain all groups' layouts would require changing its interface and the `L1Manager.reserve_write` contract — which is a bigger refactor. The dict keeps the change minimal.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done. Added comment: `# Benchmarking flag: treat every group as full-attention.`

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: It can't — `stride = num_object_groups * world_size`, both of which are >= 1 (validated in AttnWindowDesc.__post_init__). Removed the `if stride else 0` guards in all three files.

lmcache/v1/distributed/api.py
```diff
@@ -272,8 +272,19 @@ class AttnWindowDesc:
     """
 
     num_chunks_in_sw: list[int]
+    """Per-object-group window sizes in chunks."""
+
+    world_size: int = 1
+    """Number of kv_rank shards per chunk (tensor-parallel world size)."""
+
+    force_retrieve_full_kv: bool = False
+    """When True, treat every object group as full-attention during prefetch."""
```
ApostaC: Can we say "this is for local benchmarking purpose"

lmcache/v1/distributed/storage_manager.py
```diff
@@ -404,29 +405,23 @@ def submit_prefetch_task(
...
-                object-group order.
-            skip_l2: If True, do not load from L2. For ``LOOKUP`` only
-                already-resident L1 keys are returned; for ``WARM`` nothing is
-                loaded and an empty handle is returned.
-            mode: The prefetch intent (see :class:`PrefetchMode`).  ``WARM``
-                retains loaded keys and pins none; ``LOOKUP`` (default) pins
-                them for an imminent reader and follows the policy.
+            extra_count: Extra read locks per key for TP workers.
+            external_request_id: Request ID for tracing.
+            policy: Retained-subset policy (``PREFIX`` or ``SPARSE``).
```
ApostaC: need to keep the details regarding the expected high level behavior of different options.

lmcache/v1/distributed/storage_manager.py
```diff
@@ -404,29 +405,23 @@ def submit_prefetch_task(
...
+            attn_desc: Per-object-group attention windows and world_size.
+            skip_l2: If True, only return already-resident L1 keys.
+            group_layout_descs: Maps object_group_id to that group's layout
+                (each group is a separate keyed allocation, possibly with
+                different tensor shapes); ``None`` when all share ``layout_desc``.
+            mode: Prefetch intent (``WARM`` or ``LOOKUP``).
```
ApostaC: keep the old docstring for `mode`

lmcache/v1/distributed/storage_manager.py
```diff
@@ -494,87 +490,117 @@ def submit_prefetch_task(
...
+                mode=mode,
+            )
 
-            err, obj = entry
-            if err != L1Error.SUCCESS:
-                break
+        raise ValueError(f"Unsupported trim policy: {policy}")
 
-            hit_count += 1
+    def _submit_prefix_fold(
```
ApostaC, line 505: nit: we can move the helper functions to the "last part" of the class.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -670,13 +670,26 @@ def register_kv_cache(
             engine_type=engine_type,
             separate_object_groups=self._ctx.separate_object_groups,
         )
+        kv_groups_manager = cache_context.kv_layer_groups_manager
+        num_groups = kv_groups_manager.num_object_groups
```
ApostaC: nit: group -> object_group

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -1031,19 +1044,39 @@ def retrieve(
...
             prefetched_keys: list[ObjectKey] = []
             total_bytes = 0
             try:
                 for obj_group_id in range(num_object_groups):
                     obj_keys = obj_keys_per_obj_group[obj_group_id]
+                    if attn_desc.is_full_attention(obj_group_id):
+                        num_skip = 0
+                    else:
+                        sw_size_chunks = attn_desc.num_chunks_in_sw[obj_group_id]
+                        num_skip = max(0, len(obj_keys) - sw_size_chunks)
```
ApostaC: This is duplicated with some codes in `transfer_kv_per_object_group`

lmcache/v1/multiprocess/config.py
```diff
@@ -56,6 +56,11 @@ class MPServerConfig:
     L1-resident (served by the sparse leg as L1 hits, the hole recomputed)
     instead of truncating the prefix at the gap. No effect for other engines."""
 
+    force_retrieve_full_kv: bool = False
+    """Retrieve the full KV cache during prefetch, ignoring per-group
+    sliding-window sizes.  When set, every prefix chunk is prefetched
+    regardless of sliding-window bounds."""
```
ApostaC: tag it as "benchmark only"

lmcache/v1/multiprocess/engine_context.py
```diff
@@ -160,29 +177,19 @@ def find_attn_desc(self, model_name: str, world_size: int) -> AttnWindowDesc:
...
-    and layout descriptor registry. Modules receive this context at init
-    and use it for shared operations.
-
-    Args:
-        storage_manager_config: Configuration for the storage manager.
-        chunk_size: Chunk size for KV cache operations.
-        hash_algorithm: Hash algorithm for token hashing.
-        separate_object_groups: Whether to split kernel groups into one object
-            group per sliding-window size at KV-cache registration. Default True.
-    """
```
ApostaC, line 175: We probably should keep this

lmcache/v1/kv_layer_groups.py
```diff
@@ -537,13 +537,14 @@ def get_subchunk_sw_size_tokens(self, kernel_group_idx: int) -> int:
             kernel_group_idx: 0-based kernel group index.
 
         Returns:
-            The sub-chunk sliding window size. Will be the same as the
-            chunk size for non-slding-window models or big-sliding-
-            window models.
```
ApostaC, line 553: this can be kept

lmcache/v1/kv_layer_groups.py
```diff
@@ -591,11 +592,7 @@ def calculate_num_blocks(self, kernel_group_idx: int, num_tokens: int) -> int:
     def _detect_object_groups(
         self, engine_group_infos: "Sequence[EngineGroupInfo]"
     ) -> list[ObjectGroupInfo]:
-        """Bucket kernel groups into object groups.
-
-        Puts all kernel groups into a single object group when object-group
-        separation is disabled (the default). Otherwise groups the kernel groups
-        by sliding-window size measured in number of chunks.
```
ApostaC, line 617: This can be kept

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -74,26 +75,36 @@ def build_trim_mask(
...
     num_keys: int,
     policy: TrimPolicy = TrimPolicy.PREFIX,
-) -> Bitmap:
-    """Subset of ``found`` to keep (load + read-lock + report); the rest is
-    released.
-
-    PREFIX trims at the first gap (leading contiguous run). The non-PREFIX
-    policies keep every set bit, gaps included, and differ only in intent:
-    SEGMENTED_PREFIX keeps the keys that loaded when an L2 hit fails to load
-    into L1 (e.g. OOM) mid-prefix; SPARSE keeps an intentionally scattered set.
```
ApostaC, line 88: we can keep this

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -349,39 +371,21 @@ def submit_prefetch_request(
...
-        Thread-safe. Can be called from any thread.
-
-        The retained subset of found keys is chosen by ``policy`` (see
-        :class:`TrimPolicy`).  With the default ``PREFIX`` policy, only the
-        **contiguous prefix** of found keys is loaded from L2: if L2 has keys
-        {0, 1, 3, 4} but not key 2, only keys {0, 1} are loaded because the gap
-        at index 2 breaks the prefix.  Keys outside the retained set are never
-        transferred, saving I/O bandwidth and L1 memory.  Use
-        :meth:`query_prefetch_result` to retrieve the retained set once the
-        request completes.
```
ApostaC: Revisit this

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -670,13 +670,26 @@ def register_kv_cache(
...
         )
-        kv_groups_manager = cache_context.kv_layer_groups_manager
+        group_layout_descs: dict[int, MemoryLayoutDesc] | None = None
+        if num_groups > 1:
+            group_layout_descs = {
+                gid: get_layout_desc(
+                    cache_context, self._ctx.chunk_size, object_group_id=gid
+                )
+                for gid in range(num_groups)
+            }
```
ApostaC: Why do we pass `None` for the single object group case? Please revisit this.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -899,16 +935,34 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
-            is_temporary=[not r for r in retentions],
-            layout_desc=request.layout_desc,
-            mode="new",
-        )
+        retention_map = dict(zip(keys_to_reserve, retentions, strict=True))
+
+        if request.group_layout_descs:
+            groups: dict[int, list[ObjectKey]] = {}
+            for key in keys_to_reserve:
+                groups.setdefault(key.object_group_id, []).append(key)
```
ApostaC: Please check `itertools.groupby`

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -917,13 +971,28 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
                     request.request_id,
                     key,
                     err,
                 )
 
+        # Step 4b: keys that already exist in L1 but couldn't be
+        # write-locked (e.g. read-locked by a concurrent request).
+        # Acquire read locks so they can be served directly from L1
+        # without reloading from L2.
+        if not_writable_keys:
```
ApostaC: Potential race condition: key is evicted between the "reserve_write" above and the "reserve_read" below.

Or handle the failure case of reserve_read
.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -938,35 +1007,83 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+            for key in request.l1_existing_keys
+            if key not in set(retained.gather(request.keys))
+        ]
+        if l1_existing_not_retained:
+            l1_mgr.finish_read(
+                l1_existing_not_retained,
+                extra_count=request.extra_count,
+            )
+            request.l1_existing_keys -= set(l1_existing_not_retained)
+
```
ApostaC: A little bit back and forth here. Maybe try to see if we can simplify the overall workflow.

ApostaC (commented):
Let's think about whether we need to separate "L2 -> L1 duplicated keys" into another PR.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -774,18 +866,98 @@ def _start_pending_requests(self) -> None:
...
+        A lock-free peek estimates the L1 hit length. The pin pass then
+        read-locks the window of that hit and every L1-resident key beyond
+        it — the final window position is unknown until L2 results arrive::
+
+            SW keys in L1:
+
+            |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
+                                          ^ L1 hit length               ^ (unknown yet)
+            |      skip      |    pin     |  pin if in L1  |pin if in L1|pin if in L1|
```
KuntaiDu: Put this figure earlier at the second line of docstring.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -845,14 +1027,22 @@ def _start_lookup_phase(
...
 
-        # Step 1: get load plan from policy. Exclude draining adapters so no
-        # new load targets them; any keys they locked during lookup fall
-        # outside the plan and get unlocked in _unlock_unneeded_keys.
+        num_keys = len(request.keys)
+
+        # Step 1 — load plan from policy. SW keys not in L1:
+        # |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
+        #                               ^ L1 hit length               ^ L1+L2 hit length
+        # |       -        |     -      |   candidate    | candidate  | candidate  |
```
KuntaiDu, line 953: This might be better:

Potential L2 load candidate:

        # |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
        #                               ^ L1 hit length               ^ L1+L2 hit length
        # |       -        |     -      |    candidate   | candidate  | candidate  |

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1053,167 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
-        # Step 2: trim the load plan to the policy's retained subset
-        num_keys = len(request.keys)
-        merged_lookup = merge_bitmaps(load_plan.values(), num_keys)
-        retained = build_trim_mask(merged_lookup, num_keys, request.policy)
+        load_plan = trim_load_plan_with_mask(load_plan, ~request.l1_bitmap)
+
+        # Step 2 — the union fold sets the L1+L2 hit length. SW keys in L1:
+        # |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
+        #                               ^ L1 hit length               ^ L1+L2 hit length
+        # |      skip      |   unpin    |     unpin      |    pin     |   unpin    |
```
KuntaiDu: Maybe this is better
```
        # |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
        #                               ^ L1 hit length               ^ L1+L2 hit length
        # |      skip      |   unpin    |     pin --> unpin      |    pin     |   unpin    |
```

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1053,167 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
-                )
+            # Nothing to load from L2: L1 alone serves the retained set (or
+            # there is no hit at all). Unlock all L2 lookup locks.
+            request.load_plan = {}
+            self._unlock_unneeded_keys(request)
+            self._complete_without_load(request, hit_length)
+            return
+
+        # Step 3 — reserve L1 write buffers for the plan keys; any
+        # reservation failure drops its key. SW keys not in L1:
```
KuntaiDu: Make sure:
1. SW keys not in L1 ---> SW keys in L2 (and not in L1)
2. Make sure this SW keys xxx description is in a new line

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -74,26 +106,54 @@ def build_trim_mask(
...
+        hit_length, retain = fold_unfold_ranked(
+            found,
+            num_chunks,
+            attn_desc.world_size,
+            windows,
+        )
+        return hit_length, retain
+    stride = attn_desc.num_object_groups * attn_desc.world_size
+    hit_chunks = found.count_leading_ones() // stride
+    return hit_chunks, found
```
KuntaiDu:
```
if policy is TrimPolicy.PREFIX:
   xxx
elif policy is TrimPolicy.SPARSE:
   xxx
else:
   raise xxx
```

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -74,26 +106,54 @@ def build_trim_mask(
...
+
+    Args:
+        keys: The request's object keys, in prefix order.
+        selected: The subset of ``keys`` whose indices should be set.
+
+    Returns:
+        A ``len(keys)``-sized bitmap.
+    """
+    bitmap = Bitmap(len(keys))
+    for i, key in enumerate(keys):
```
KuntaiDu: This seems to be slow. Is there a way to accelerate it? (e.g. by itertool or something)

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -74,26 +106,54 @@ def build_trim_mask(
...
-    """Subset of ``found`` to keep (load + read-lock + report); the rest is
-    released.
-
-    PREFIX trims at the first gap (leading contiguous run). The non-PREFIX
-    policies keep every set bit, gaps included, and differ only in intent:
-    SEGMENTED_PREFIX keeps the keys that loaded when an L2 hit fails to load
-    into L1 (e.g. OOM) mid-prefix; SPARSE keeps an intentionally scattered set.
+    attn_desc: AttnWindowDesc = DEFAULT_ATTN_WINDOW_DESC,
+) -> tuple[int, Bitmap]:
+    """Compute the retained subset of found keys and the prefix hit length.
```
KuntaiDu: The original comment (e.g. PREFIX, SPARSE)

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -845,14 +1027,23 @@ def _start_lookup_phase(
...
         request.phase = PrefetchPhase.PLAN_AND_LOAD
         self._status_lookup_phase_count -= 1
         self._status_load_phase_count += 1
 
-        # Step 1: get load plan from policy. Exclude draining adapters so no
-        # new load targets them; any keys they locked during lookup fall
-        # outside the plan and get unlocked in _unlock_unneeded_keys.
+        num_keys = len(request.keys)
+
+        # Step 1 — load plan from policy.
```
KuntaiDu: Step 1 --- generate what keys should be loaded from L2 to L1

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1054,164 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+            self._finish_request(request)
+            return
+
+        # Step 3 — reserve L1 write buffers for the plan keys; any
+        # reservation failure drops its key.
+        # SW keys in L2 (and not in L1):
+        # |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
+        #                               ^ L1 hit length               ^ L1+L2 hit length
+        # |       -        |     -      |       -        |  loading*  |     -      |
+        # (*) dropped on OOM or contention.
```
KuntaiDu: Can we create a function, say _fallback_to_l1_hit(), which is 
            request.load_plan = {}
            self._unlock_unneeded_keys(request)
            self._report_lookup_hit(request.request_id, hit_length)
            self._finish_request(request)
            return
and then when any of L2 load fails, we _fallback_to_l1_hit?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1054,164 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+        request.load_plan = trimmed_plan
+
+        # Step 6 — free L2 lookup locks for keys outside the plan.
+        # L2 lookup locks:
+        # |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
+        #                               ^ L1 hit length               ^ L1+L2 hit length
+        # |      free      |    free    |      free      | keep plan  |    free    |
+        self._unlock_unneeded_keys(request)
+
+        if not trimmed_plan:
```
KuntaiDu: Same here -- can we `_fallback_to_l1_hit`?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -349,19 +428,22 @@ def submit_prefetch_request(
...
         Thread-safe. Can be called from any thread.
 
-        The retained subset of found keys is chosen by ``policy`` (see
-        :class:`TrimPolicy`).  With the default ``PREFIX`` policy, only the
-        **contiguous prefix** of found keys is loaded from L2: if L2 has keys
-        {0, 1, 3, 4} but not key 2, only keys {0, 1} are loaded because the gap
-        at index 2 breaks the prefix.  Keys outside the retained set are never
-        transferred, saving I/O bandwidth and L1 memory.  Use
+        A key counts as found if it is already resident in L1 or any L2
+        adapter reports it; the retained subset of found keys is chosen by
```
KuntaiDu: Put
`The retained subset of found keys is chosen by ...` to a new line to reduce code diff.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -1109,14 +1399,16 @@ def _poll_load_results(
...
-    def _finalize_load(self, request: InFlightPrefetchRequest) -> None:
+    def _finish_request(self, request: InFlightPrefetchRequest) -> None:
         """
-        Finalize a completed load: build result bitmap, transition L1
-        state, release read locks outside the retained set, and report the
-        retained-key bitmap.
-
-        Partial load failures can create gaps, so a loaded key may fall
-        outside the policy's retained set; its read lock must be released.
+        Finish a request — the single completion path, with or without an
```
KuntaiDu: Docstring is a bit werid. Here is a reference

Finishes a request and free all unnecessary locks.

Workflow: 
1. check L1 hit length
2. Free all other locks that does not locate in the sliding window corresponds to L1 hit.
...

lmcache/v1/distributed/storage_manager.py
```diff
@@ -494,87 +500,117 @@ def submit_prefetch_task(
...
+        l1_read_result: dict[ObjectKey, tuple[L1Error, "MemoryObj | None"]],
+        attn_desc: AttnWindowDesc,
+        extra_count: int,
+        external_request_id: str,
+        layout_desc: MemoryLayoutDesc,
+        skip_l2: bool,
+        group_layout_descs: dict[int, MemoryLayoutDesc] | None = None,
+        mode: PrefetchMode = PrefetchMode.LOOKUP,
+    ) -> PrefetchHandle:
+        """PREFIX path: fold L1 presence, retain in-window keys, submit rest to L2."""
```
KuntaiDu: Docstring is a bit too short.
Please doc the args and returns.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -87,13 +122,33 @@ def build_trim_mask(
...
+        ValueError: If ``policy`` is not a known :class:`TrimPolicy`.
     """
+    stride = attn_desc.num_object_groups * attn_desc.world_size
     if policy is TrimPolicy.PREFIX:
-        return Bitmap(num_keys, found.count_leading_ones())
-    return found
+        num_chunks = num_keys // stride
+        windows = attn_desc.num_chunks_in_sw
+        # Benchmarking flag: treat every group as full-attention.
+        if attn_desc.force_retrieve_full_kv:
```
KuntaiDu: Can we rename all `force_retrieve_full_kv` to `force_retrieve_full_kv_benchmark_only`?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: explicit dispatch — `if PREFIX / if SEGMENTED_PREFIX or SPARSE / raise ValueError` (SEGMENTED_PREFIX kept: it is a distinct member used by the fault-injection path).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Resolved by deletion: reservation is now all-or-nothing, so the membership bitmap is no longer needed at all; the per-group reserve now uses sorted+itertools.groupby (C-speed).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Restored the original docstring text verbatim (only the new attn_desc arg / tuple return / Raises were added).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: 'Step 1 — generate what keys should be loaded from L2 to L1.'

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Superseded by the stronger version: with lock state tracked on the request (l2_locked + hit_reported), release is idempotent and every path now calls _finish_request directly — one reconciler, no separate fallback helper needed.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Same resolution: this branch is now just _finish_request(request).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: original lines restored verbatim; the new 'A key counts as found…' sentence sits on its own lines above.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Rewrote in the suggested style: one-line summary + numbered Workflow (collect results → release L2 locks → transition/delete → fold → free locks outside retained → report).

lmcache/v1/distributed/storage_manager.py
KuntaiDu: Done: full docstring with Args and Returns for _submit_prefix_fold.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Renamed repo-wide (26 sites, 9 files): force_retrieve_full_kv → force_retrieve_full_kv_benchmark_only. The user-facing CLI flag --force-retrieve-full-kvcache is unchanged.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Fixed structurally: the controller now pins L1-resident keys first (reserve_read is an atomic observe-and-lock), and any reserve_write failure just abandons the load — no lock is ever taken through a failure result. A deterministic regression test (EvictionRacingL1Manager, fires an eviction between every pair of L1 manager calls) covers exactly this race: tests/v1/distributed/test_prefetch_controller.py::TestConcurrentEvictionRace.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Simplified end to end: pin pass → union fold → all-or-nothing reserve → load, with lock state tracked on the request and a single completion path (_finish_request) that reconciles everything (no mid-flight pin/unpin churn, no per-branch cleanup). Design doc: docs/design/v1/distributed/storage_controllers/prefetch_l1_pin_pass.md.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: itertools.groupby over keys sorted by object_group_id (stable sort keeps within-group prefix order; chunk-major input interleaves group ids, so the sort is required for groupby to batch correctly).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Restored verbatim.

lmcache/v1/distributed/api.py
KuntaiDu: Done via rename: the field is now force_retrieve_full_kv_benchmark_only, and its docstring says it is a benchmarking flag.

lmcache/v1/multiprocess/config.py
KuntaiDu: Done via rename: force_retrieve_full_kv_benchmark_only.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
KuntaiDu: Fixed: the transfer path now always builds the per-group layout map (group 0 -> merged layout in the single-group case), and the controller no longer special-cases None — its request field is a plain dict (empty = every group falls back to layout_desc) and the reserve path is one groupby loop for both cases.

lmcache/v1/distributed/storage_manager.py
```diff
@@ -494,87 +500,137 @@ def submit_prefetch_task(
...
+                ``None`` when all keys share ``layout_desc``.
+            mode: The prefetch intent (see :class:`PrefetchMode`).
 
-            hit_count += 1
+        Returns:
+            A :class:`PrefetchHandle` carrying the L1 hit (retained indices
+            and hit chunks) and the pending L2 prefetch request id (``-1``
+            when nothing was submitted to L2).
+        """
+        num_groups = attn_desc.num_object_groups
```
KuntaiDu: Again `num_groups` --> `num_object_groups`

KuntaiDu: Done: num_groups → num_object_groups (also applied to the same local in lmcache_driven_transfer).

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
KuntaiDu: Done: the num_groups local is now num_object_groups.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
KuntaiDu: Agreed it duplicates the window-skip in transfer_kv_per_object_group. Deduplication isn't necessary right now — the helper already carries a TODO(sliding-window) to hoist the skip into a parameter, and the retrieve site now has a mirroring TODO pointing at it, so the two sites cross-reference until that lands.

lmcache/v1/multiprocess/config.py
```diff
@@ -374,6 +379,13 @@ def add_mp_server_args(
         "retrieve failure, retain the gapped prefix so post-gap chunks stay "
         "L1-resident instead of truncating at the gap. No effect otherwise.",
     )
+    mp_group.add_argument(
+        "--force-retrieve-full-kvcache",
```
KuntaiDu: --force-retrieve-full-kvcache-benchmark-only

lmcache/v1/distributed/storage_manager.py
```diff
@@ -494,87 +500,137 @@ def submit_prefetch_task(
...
-        if remaining_keys and self._has_l2_adapters():
+
+        if not l1_only and remaining_keys:
             prefetch_request_id = self._prefetch_controller.submit_prefetch_request(
                 remaining_keys,
                 layout_desc,
                 extra_count=extra_count,
                 attn_desc=attn_desc,
-                policy=policy,
+                policy=TrimPolicy.PREFIX,
```
KuntaiDu: why not `policy=policy`?

lmcache/v1/multiprocess/config.py
KuntaiDu: Done: CLI flag renamed to --force-retrieve-full-kvcache-benchmark-only (help text now leads with "Benchmarking only:"), args mapping updated, and the two pipeline.yml comments that referenced the old flag updated to match. Verified end-to-end: flag parses and sets force_retrieve_full_kv_benchmark_only.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -774,18 +868,105 @@ def _start_pending_requests(self) -> None:
...
 
+    def _pin_l1_keys(
+        self,
+        keys: list[ObjectKey],
+        extra_count: int,
+        policy: TrimPolicy,
+        attn_desc: AttnWindowDesc,
+        mode: PrefetchMode,
+    ) -> tuple[Bitmap, set[ObjectKey]]:
+        """Peek L1, then pin (read-lock) the servable subset (LOOKUP only)::
```
KuntaiDu: Peek L1, then pin (read-lock) keys in L1. Details in the figure below.

lmcache/v1/distributed/storage_manager.py
KuntaiDu: Done: _submit_prefix_fold now takes policy and forwards policy=policy (the dispatcher only routes PREFIX there, but no more hardcoding); also fixed the same pattern in the SPARSE branch where policy was in scope.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: summary is now 'Peek L1, then pin (read-lock) keys in L1. Details in the figure below.'

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1059,135 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
-            layout_desc=request.layout_desc,
-            mode="new",
-        )
+        retention_map = dict(zip(keys_to_reserve, retentions, strict=True))
+
+        # When per-group layouts are available, batch reserve_write by
+        # object_group_id so each group uses its own tensor shapes.
+        if request.group_layout_descs:
+            # Chunk-major order interleaves group ids, so sort before
+            # groupby (stable: within-group prefix order is preserved).
```
KuntaiDu: DOn't think these 2 lines answer any confusion. Remove.

KuntaiDu: Removed.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -170,6 +232,19 @@ class InFlightPrefetchRequest:
     # Load phase: keys that were write-reserved in L1
     write_reserved_keys: list[ObjectKey] = field(default_factory=list)
     write_reserved_objs: dict[ObjectKey, MemoryObj] = field(default_factory=dict)
+    # L1 pin pass, set when the request starts (before L2 lookup):
+    # key indices served from L1 — pinned (read-locked) in LOOKUP mode, a
+    # point-in-time peek in WARM mode. Pins outside the final retained set
+    # are released in _finish_request.
+    l1_bitmap: Bitmap = field(default_factory=lambda: Bitmap(0))
+    # Keys backing l1_bitmap with read locks (always empty in WARM mode).
+    l1_pinned_keys: set[ObjectKey] = field(default_factory=set)
```
KuntaiDu: Maybe call it `l1_readlocks`

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -157,6 +212,13 @@ class InFlightPrefetchRequest:
     pending_lookup_tasks: dict[int, L2TaskId] = field(default_factory=dict)
     # Lookup phase: adapter_idx -> bitmap (populated as results arrive)
     lookup_results: dict[int, Bitmap] = field(default_factory=dict)
+    # L2 lookup locks currently held (adapter_idx -> key indices). Mirrors
+    # lookup_results on arrival; _release_l2_locks subtracts as locks are
+    # returned, so releasing is idempotent from any state.
+    l2_locked: dict[int, Bitmap] = field(default_factory=dict)
```
KuntaiDu: Maybe call this `l2_adapter2readlocks`

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -170,6 +232,19 @@ class InFlightPrefetchRequest:
     # Load phase: keys that were write-reserved in L1
     write_reserved_keys: list[ObjectKey] = field(default_factory=list)
     write_reserved_objs: dict[ObjectKey, MemoryObj] = field(default_factory=dict)
+    # L1 pin pass, set when the request starts (before L2 lookup):
+    # key indices served from L1 — pinned (read-locked) in LOOKUP mode, a
+    # point-in-time peek in WARM mode. Pins outside the final retained set
+    # are released in _finish_request.
+    l1_bitmap: Bitmap = field(default_factory=lambda: Bitmap(0))
```
KuntaiDu: maybe `l1_lookup_result`?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -774,18 +880,106 @@ def _start_pending_requests(self) -> None:
...
+        if not keys_to_pin:
+            return Bitmap(num_keys), set()
+
+        pin_results = self._l1_manager.reserve_read(
+            keys_to_pin, extra_count=extra_count
+        )
+        # The fold input is the PIN RESULTS, never the peek: a key evicted
+        # between peek and pin simply drops out here. pin_bitmap's set
+        # indices are exactly keys_to_pin's positions, so zip them instead
+        # of re-scanning all keys.
```
KuntaiDu: I don't understand this comment. It's NOT explaining the functionality of the code. It's explaining the some details that do not matters for developer. Don't do this again.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1072,157 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+            ]
+            if to_unpin:
+                self._l1_manager.finish_read(to_unpin, extra_count=request.extra_count)
+                request.l1_pinned_keys.difference_update(to_unpin)
+
+        # Step 3 — reserve L1 write buffers for the plan keys, all or
+        # nothing: any reservation failure (OOM or contention) abandons the
+        # L2 load; finish returns every reserved buffer and serves the L1
+        # hit. The engine recomputes the rest and re-stores it within one
+        # round.
```
KuntaiDu:
```suggestion
        # Step 3 — reserve L1 write buffers for the plan keys
        # If any failure (OOM or contention or others) happens,
        # we fall back to l1-only longest hit (`l1_fallback_retain`)
```

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -1185,41 +1504,61 @@ def _finalize_load(self, request: InFlightPrefetchRequest) -> None:
...
 
-    def _unlock_all_plan_keys(self, request: InFlightPrefetchRequest) -> None:
-        """Phase 2 unlock: release L2 locks for all keys in the load plan."""
-        for adapter_idx, load_bitmap in request.load_plan.items():
-            unlock_keys = load_bitmap.gather(request.keys)
-            self._l2_adapters[adapter_idx].submit_unlock(unlock_keys)
+        Operates on the tracked lock state (``request.l2_locked``) and
+        subtracts what it releases, so any release sequence is idempotent —
+        every path can converge on ``_finish_request`` without
+        double-unlocking.
```
KuntaiDu:
```suggestions
       """Release L2 locks in ``request.l2_locked``, except for keys in ``keep``.
```

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: l1_pinned_keys → l1_readlocks.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: l2_locked → l2_adapter2readlocks.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: l1_bitmap → l1_lookup_result (all 17 sites, docstrings included).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Rewrote as a functional statement: 'Record the keys reserve_read locked; keys that failed (e.g. evicted since the peek) are left out of the L1 lookup result.' Also swept the rest of the file with the same guidance — removed the rationale-style asides (idempotent/optimization/reconciles phrasing) from inline comments.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Applied your suggestion verbatim (figure kept below it).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Applied: summary is now 'Release L2 locks in ``request.l2_adapter2readlocks``, except for keys in ``keep``.' and the idempotence paragraph was cut to one functional sentence.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -774,18 +877,84 @@ def _start_pending_requests(self) -> None:
...
+            l1_lookup_result, len(keys), policy, attn_desc
+        )
+        stride = attn_desc.num_object_groups * attn_desc.world_size
+        within_l1_hit = Bitmap(len(keys), hit_length * stride)
+        # evictable: pinned SW chunks behind the window, within the L1 prefix
+        # hit -- won't be read again, so release their locks.
+        evictable = l1_lookup_result & ~l1_window & within_l1_hit
+        if evictable.popcount() > 0:
+            to_unpin = [key for key in evictable.gather(keys) if key in l1_readlocks]
+            if to_unpin:
```
KuntaiDu: Name 1 case where `to_unpin` is empty. If you can't remove this if.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1047,155 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
             request.lookup_results,
             routing_descriptors,
         )
-
-        # Step 2: trim the load plan to the policy's retained subset
-        num_keys = len(request.keys)
-        merged_lookup = merge_bitmaps(load_plan.values(), num_keys)
-        retained = build_trim_mask(merged_lookup, num_keys, request.policy)
+        load_plan = trim_load_plan_with_mask(load_plan, ~request.l1_lookup_result)
```
KuntaiDu: Add a comment here saying we don't need to load keys that are already in L1

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1047,155 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+
+        # Step 2 — the union fold sets the L1+L2 hit length. Evictable pins
+        # are unpinned just below (no-load finishes release them at finish).
+        # Trim to the retained subset of the L1 ∪ L2 union, so
+        # L1-resident keys extend the hit even where L2 lacks them.
```
KuntaiDu: Please be simple. Explain the output don't explain the process. REMEMBER THIS.
Example: 
Step 2 -- calculate L1 + L2 maximum hit length.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -863,66 +1047,155 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
-                    },
-                )
+            # Nothing to load from L2: finish releases all unneeded locks
+            # and reports the L1-only hit.
+            self._finish_request(request)
+            return
+
+        # Unpin keys outside both the retained set and the L1-only fallback
+        # (`l1_fallback_retain`, what finish falls back to if the load
+        # aborts or fails), so the reservation below can evict them.
```
KuntaiDu: Too long. Just say unpin the keys based on the following figure.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: You're right — no such case exists: evictable ⊆ l1_readlocks by construction, so the filter could never remove anything. The `if to_unpin` and the filter are removed (this also fell out of merging l1_lookup_result and l1_readlocks into one bitmap field). The remaining `if evictable.popcount() > 0` guard stays: it is false for every full-attention model (no SW chunks to release).

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: '# Keys already in L1 don't need to be loaded from L2.'

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: 'Step 2 — calculate the L1 + L2 maximum hit length and the retained set.' Guidance saved: comments state the output, not the process.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: Done: 'Unpin the keys based on the following figure.' + the figure.

lmcache/v1/distributed/storage_manager.py
```diff
@@ -478,78 +480,111 @@ def submit_prefetch_task(
...
-        hit_count = 0
-        for key in keys:
-            entry = l1_read_result.get(key, None)
-            if entry is None:
-                break
+        # PREFIX: fold the per-(group, chunk, rank) L1 presence into the
+        # model-wide hit and the per-object-group retain set (sliding-window
+        # aware). All-full-attention reduces to the contiguous leading-ones
+        # prefix. Keys past the L1 hit are sent to L2.
+        if spec.policy is TrimPolicy.PREFIX:
```
ApostaC: if -> elif

lmcache/v1/distributed/storage_manager.py
ApostaC: nit: we can move the helper functions to the "last part" of the class.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
ApostaC, line 1: Remove the changes in `store`/`retrieve` code path (keep the group layout desc registration)

ApostaC (commented):
See the comments

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -174,6 +232,14 @@ class InFlightPrefetchRequest:
...
     write_reserved_keys: list[ObjectKey] = field(default_factory=list)
     write_reserved_objs: dict[ObjectKey, "MemoryObj"] = field(default_factory=dict)
+    # Key indices found (and read-locked) in L1 when the request starts.
+    l1_readlocks: Bitmap = field(default_factory=lambda: Bitmap(0))
+
+    group_layout_descs: dict[int, MemoryLayoutDesc] = field(default_factory=dict)
+    """Maps object_group_id to that group's layout. Each object group is a
+    separate keyed L1 allocation, so each needs its own descriptor (one
+    ``MemoryLayoutDesc`` describes a single group's MemoryObj). Groups
+    missing from the map fall back to ``layout_desc``."""
```
ApostaC: I'm considering replacing the existing `layout_desc` with `group_layout_descs`. 
Otherwise, passing both `layout_desc` and `group_layout_desc` at the same time will be confusing.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -186,13 +252,16 @@ class PrefetchController(StorageControllerInterface):
     """
     Asynchronously prefetches data from L2 adapters into L1 memory.
 
-    The controller:
+    The controller (terms as in the module-docstring interval figure):
```
ApostaC: remove "(terms as in the module-docstring interval figure):" because people don't know what it refers to.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -186,13 +252,16 @@ class PrefetchController(StorageControllerInterface):
...
     1. Accepts prefetch requests via submit_prefetch_request (thread-safe).
-    2. Runs a background thread that submits lookup_and_lock to all adapters.
-    3. Uses PrefetchPolicy to compute a load plan from lookup results.
-    4. Reserves L1 write buffers and submits load tasks to adapters.
-    5. On completion, transitions loaded keys to read-locked state.
-    6. Reports the number of prefix hits via query_prefetch_result.
+    2. Runs a background thread that pins L1-resident keys (read locks)
+       and submits lookup_and_lock to all adapters.
```
ApostaC: ~~Quick question: won't the L1 keys be pinned by the storage manager already before calling prefetch controller?~~

~~In storage manager, it will call `l1_manager.reserve_read` which will directly pin the found keys.~~

Okay I got it after discussing offline

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -186,13 +252,16 @@ class PrefetchController(StorageControllerInterface):
...
+    3. Computes the load plan from the L1 ∪ L2 union fold: keys in L2 and
+       not in L1 inside the final window; pinned keys are never transferred.
```
ApostaC: "keys in L2 and not in L1 inside the final window; pinned keys are never transferred" is confusing here.

Maybe we can write:
```
Computes the load plan based on the found keys in L1 and L2.
```

Additionally, when will we unlock the extra locked keys in L1? Please reflect it in the docstring.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -186,13 +252,16 @@ class PrefetchController(StorageControllerInterface):
...
+    4. Reserves L1 write buffers (all-or-nothing) and submits load tasks.
+    5. On completion, loaded keys become pinned for the retriever.
```
ApostaC: I think let's not mix the use of "pin" and "read-lock". We should use "lock"/"unlock" instead of "pin"/"unpin".

Read-lock is an internal mechanism where "pin" usually refers to the user-facing cache management operation.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -744,13 +826,77 @@ def _start_pending_requests(self) -> None:
     # Lookup phase
     # =========================================================================
 
+    def _pin_l1_keys(
```
ApostaC: Similar to the comments above, let's rename this to `_lock_l1_keys`.

~~Also, we should double check whether we really need this.~~

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -744,13 +826,77 @@ def _start_pending_requests(self) -> None:
...
+        """Read-lock the L1-resident keys the request will hold.
+
+        Pins every key readable in L1, then, for windowed ``PREFIX``
+        retention, releases the read locks on sliding-window chunks behind
+        the L1 hit's own window. Pinning does not refresh eviction recency;
+        the LRU signal is sent by ``_finish_request`` for the retained keys
+        only.
+
+        Args:
+            keys: The request's object keys, in prefix order.
```
ApostaC: Any extra information about the order? (Like chunk - group)?

Also, please add a short sentence to explain that StorageManager already capped away the real "L1" hits.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -816,66 +959,149 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+            |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
+                                          ^ L1 hit length               ^ L1+L2 hit
+            |       -        |     -      |       -        |  loading   |     -      |
+
+        Successful reservations are recorded on
+        ``request.write_reserved_keys`` / ``request.write_reserved_objs``.
+        Failures publish an ``L2_PREFETCH_FAILED`` event; the caller
+        abandons the L2 load if any key failed (all-or-nothing). No lock is
+        ever acquired through a failure result, so a concurrent eviction
+        between manager calls cannot invalidate the reported hit.
```
ApostaC: nit: Remove the sentence
```
        No lock is
        ever acquired through a failure result, so a concurrent eviction
        between manager calls cannot invalidate the reported hit.
```

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -816,66 +959,149 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+
+    def _reserve_load_buffers(
+        self,
+        request: InFlightPrefetchRequest,
+        keys_to_reserve: list[ObjectKey],
+    ) -> set[ObjectKey]:
+        """Reserve L1 write buffers for the keys to load from L2.
+
+        The keys sit in the loading segment — in L2, not in L1, inside the
+        final window::
```
ApostaC: For those keys in L1, when did we "lock" them?

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -890,39 +1116,35 @@ def _transition_to_load_phase(self, request: InFlightPrefetchRequest) -> None:
...
+            # promoted; a later request finds it through its L1 pin pass.
             self._event_bus.publish(
                 Event(
-                    event_type=EventType.L2_PREFETCH_LOOKUP_COMPLETED,
-                    metadata={
-                        "request_id": request.request_id,
-                        "prefix_hit_count": 0,
-                    },
+                    event_type=EventType.L2_PREFETCH_FAILED,
+                    metadata={"reason": "l1_contended", "keys": contended_keys},
```
ApostaC, line 1129: I thought we would still proceed if there are contended L1 keys. Maybe I'm wrong? But why do we emit `L2_PREFETCH_FAILED` event.

ApostaC (commented):
Please see my comments, mostly nit comments, but with a few questions.

Let's double check that every "locked" keys will be "unlocked".

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: We will giveup L2 load, since the contention is because the key that L2 wants to load is holding a l1 write lock.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
KuntaiDu: We lock them to avoid loading them again from L2.

```
            |out of L1-hit sw|in L1-hit sw|out of L2-hit sw|in L2-hit sw| remaining  |
                                          ^ L1 hit length               ^ L1+L2 hit
            |       -        |     -      |       -        |  loading   |     -      |
```
This will overlock the keys in "out of L2-hit sw" and "remaining", which will never be used. But the time duration of doing so is short (after lookup, before the loadplan being generated) so I think it's fine.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -915,7 +909,11 @@ def _start_lookup_phase(
             return
 
         for adapter_id, adapter in routing_adapters.items():
-            task_id = adapter.submit_lookup_and_lock_task(spec.keys, spec.layout_desc)
+            # The layout is an advisory hint (see L2AdapterInterface); pass
+            # group 0's.
+            task_id = adapter.submit_lookup_and_lock_task(
```
ApostaC, line 917: L2 adapters' `submit lookup and lock task` needs to be updated. This is for the P2P adapter (other adapters are not using this).
Let's do this immediately after this PR.

lmcache/v1/multiprocess/modules/blend_v3.py
```diff
@@ -841,7 +841,7 @@ def _sparse_prefetch_submit(
         handle: PrefetchHandle = self._ctx.storage_manager.submit_prefetch_task(
             PrefetchRequestSpec(
                 keys=uniq_keys,
-                layout_desc=layout_desc,
+                group_layout_descs={0: layout_desc},
```
ApostaC, line 858: Let's throw an runtime error and force cacheblend to use `--no-separate-object-groups` mode for now.

We can make `--no-separate-object-groups` as default.

ApostaC, line 858: Leave a TODO comment here for Weishu to handle in the future

lmcache/v1/multiprocess/modules/lookup.py
```diff
@@ -289,14 +289,13 @@ def lookup(
 
         group_layout_descs = self._ctx.layout_desc_registry.find_group_layout_descs(
             model_name, world_size
-        )
+        ) or {gid: layout_desc for gid in range(attn_desc.num_object_groups)}
```
ApostaC: let's log error if find is not successful and create a "0" return

lmcache/v1/multiprocess/modules/p2p_controller.py
```diff
@@ -246,7 +246,9 @@ def p2p_lookup_and_lock(
         # NOTE: skip_l2=True -- only objects already resident in L1 are locked.
         handle = self._ctx.storage_manager.submit_prefetch_task(
             PrefetchRequestSpec(
-                keys=keys, layout_desc=layout_desc, policy=TrimPolicy.SPARSE
+                keys=keys,
+                group_layout_descs={0: layout_desc},
+                policy=TrimPolicy.SPARSE,
```
ApostaC, line 255: left a TODO and update this immdiately after this PR (see the related comment above)

lmcache/v1/multiprocess/warm_prefetch.py
```diff
@@ -92,7 +92,7 @@ def submit(
         handle = storage_manager.submit_prefetch_task(
             PrefetchRequestSpec(
                 keys=keys,
-                layout_desc=layout_desc,
+                group_layout_descs={0: layout_desc},
```
ApostaC, line 95: @ruizhang0101 we need to update here.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -91,13 +128,29 @@ def build_trim_mask(
...
-    return found
+        num_chunks = num_keys // stride
+        hit_length, retain = fold_unfold_ranked(
+            found,
+            num_chunks,
+            attn_desc.world_size,
+            attn_desc.num_chunks_in_sw,
+        )
+        return hit_length, retain
+    if policy in (TrimPolicy.SEGMENTED_PREFIX, TrimPolicy.SPARSE):
```
KuntaiDu: maybe `if` -> `elif`

lmcache/v1/multiprocess/config.py
```diff
@@ -352,9 +353,10 @@ def add_mp_server_args(
     mp_group.add_argument(
         "--separate-object-groups",
         action=argparse.BooleanOptionalAction,
-        default=True,
+        default=False,
         help="Split kernel groups into one object group per sliding-window size "
-        "at KV-cache registration (for hybrid models). (Default is True)",
+        "at KV-cache registration (for hybrid models). Off by default until "
+        "the separated-group path is production ready. (Default is False)",
```
ApostaC: Let's remove "until the separated-group path is production ready. (Default is False)" -- likely we will forget to update it when it's production ready

ApostaC (approved):
Otherwise LGTM!

sammshen (approved):
LGTM!

PR #3873, opened 2026-06-25
raw-block: persist stable device identity in checkpoints

daegyu94:
Raw-block checkpoint recovery previously relied on the configured `device_path`
to identify checkpoint metadata. That is fragile because device paths can change
across reboot, and the same NVMe namespace may be opened as `/dev/nvme*` or
`/dev/ng*` depending on whether `io_uring_cmd` is used.

This PR adds a stable `device_id` to raw-block checkpoint metadata:
- NVMe namespaces use `nvme:<wwid>` from `/sys/block/<namespace>/wwid`
- regular files use `file:<st_dev>:<st_ino>:<st_size>`

Recovery now validates `device_id` instead of relying on `device_path`, so path
changes do not reject valid checkpoints. Unsupported paths fail fast when
checkpoint loading or periodic checkpointing is enabled.

For upgrade compatibility, this PR also adds an opt-in legacy migration switch:
- MP `raw_block`: `allow_legacy_checkpoint_without_device_id=true`
- non-MP `rust_raw_block`: `rust_raw_block.allow_legacy_checkpoint_without_device_id=true`

This option allows loading older checkpoints that were written before `device_id`
was recorded. It only applies when the checkpoint is missing `device_id`;
checkpoints with a mismatched `device_id` are still rejected. After the legacy
checkpoint is loaded and a new checkpoint is written, the option should be
disabled again.

`device_path` may still appear in diagnostics or legacy metadata, but recovery
validation uses stable `device_id`.

The legacy migration option is disabled by default and is intended only as a
one-time upgrade path for pre-`device_id` checkpoints.

**If applicable**:

**Tested with**:

```bash
pytest -q tests/v1/storage_backend/test_rust_raw_block_backend.py \
tests/v1/storage_backend/test_raw_block_core.py
```

### DIFF
--- a/docs/design/v1/distributed/l2_adapters/raw_block.md
+++ b/docs/design/v1/distributed/l2_adapters/raw_block.md
@@ -101,6 +101,13 @@ Rules:
 
 The on-device format is intentionally unchanged by the MP adapter work.
 
+Checkpoint payloads record a stable `device_id` rather than relying on
+`device_path`, which can change across reboot or when the same NVMe namespace is
+opened through `/dev/nvme*` versus `/dev/ng*`. Older checkpoints that do not
+contain `device_id` are accepted only when
+`allow_legacy_checkpoint_without_device_id=true`, and that option is intended
+only for one-time migration before writing a new checkpoint with `device_id`.
+
 Recovered keys are exposed to the shared L2 eviction policy on adapter startup,
 so reclaimed slots come from global L2 eviction or explicit `delete()` calls.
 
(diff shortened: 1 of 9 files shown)


DongDongJu (approved):
LGTM

daegyu94:
Rebased onto the latest `upstream/dev` to resolve the merge conflict, and fixed two tests that needed real temporary files for stable `device_id` validation.
Validation: 42 raw-block tests passed.

DongDongJu:
@daegyu94 Please resolve the conflict

daegyu94:
@DongDongJu Resolved the merge conflict by rebasing onto the latest `upstream/dev`.

DongDongJu:
@ApostaC @sammshen Could you take a look?

PR #3876, opened 2026-06-25
[p2p][bugfix] fix coordinator server disconnected error

chunxiaozheng:
When the coordinator's uvicorn keep-alive timeout (default `5s`) equals the MP server's heartbeat interval (also `5s`), a race condition causes intermittent Server disconnected without sending a response errors. The server may close the idle connection at the exact moment the client sends its next heartbeat.

Fix: Add a configurable timeout_keep_alive parameter (default `10s`) to the coordinator's uvicorn server, ensuring idle connections stay open longer than the heartbeat interval.

the error log is as follows:
Clipboard_Screenshot_1782388224


ApostaC (approved):

sammshen (approved):
LGTM!

PR #3881, opened 2026-06-25
feat(cli): add CSV output format for metrics

CicaMatt:

Adds a third CLI output format, `csv`, alongside `terminal` and `json`. It
registers a `CsvFormatter` through the existing `@register_formatter`
registry, so every metric-producing command (`bench`, `describe`, `ping`,
`query`, `quota`) supports `--format csv` with no changes at the call sites.

The format renders metrics in long/"tidy" form with a fixed
`section,index,key,label,value` schema — one row per metric. This adds a
capability the existing formats don't cover: tabular output that loads
directly into a spreadsheet or dataframe. It's most useful for comparing
`lmcache bench` runs (append each summary to one CSV, then diff/pivot) and
for slicing nested status reports with plain text tools. A per-list-group
`index` column keeps repeated sections (e.g. multiple models or L2
adapters) distinguishable, so the output is lossless; the stdlib `csv`
module handles quoting/escaping, so no new dependency is introduced.

- Scope is intentionally CSV-only. The formatter layer previously had no
  direct unit tests, so `tests/cli/test_formatter.py` is new and covers
  `CsvFormatter` plus the shared `get_formatter`/registry path. I did not
  add tests for the pre-existing `terminal`/`json` formatters — that's
  outside this change and better as a focused follow-up; happy to open an
  issue for it.
- The report title is deliberately omitted from CSV output (a flat table
  has no slot for a presentational header); the `format(title, ...)`
  parameter is kept only to satisfy the `MetricsFormatter` interface.
- Docs: updated every English CLI page that enumerates the formats plus the
  developer guide; left the `zh_CN/**.po` catalogs to the translation
  automation.

**If applicable**:

### DIFF
--- a/docs/source/cli/bench.rst
+++ b/docs/source/cli/bench.rst
@@ -779,7 +779,7 @@ Options
    * - ``--format FORMAT``
      - ``terminal``
      - Stdout output format for the final metrics summary. Available:
-       ``terminal``, ``json``.
+       ``terminal``, ``json``, ``csv``.
    * - ``--output PATH``
      - *(unset)*
      - Save the final metrics summary to a file at PATH (format chosen
--- a/docs/source/cli/describe.rst
+++ b/docs/source/cli/describe.rst
@@ -71,7 +71,7 @@ Options
    * - ``--url``
      - LMCache HTTP server URL (default: ``http://localhost:8080``).
    * - ``--format``
-     - Output format: ``terminal`` (default) or ``json``.
+     - Output format: ``terminal`` (default), ``json``, or ``csv``.
    * - ``--output PATH``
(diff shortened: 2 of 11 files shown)


CicaMatt:
Heads-up on test scope: this PR adds the first direct unit tests for the formatter layer, but they cover only CsvFormatter (plus the shared get_formatter/registry path). The pre-existing terminal and json formatters still have no dedicated unit tests — they're only exercised incidentally via command-level tests. I kept this PR focused rather than backfilling coverage for code it doesn't change.

If this gets merged, I'd suggest a follow-up issue to add symmetric unit tests for TerminalFormatter and JsonFormatter, so the whole formatter layer is covered consistently. Happy to open that issue and take it on.

PR #3882, opened 2026-06-25
[Fix][RawBlock] Cap uring_cmd auto transfer size by max_segments

3xdevv:

When `use_uring_cmd=true`, raw-block auto-detected the NVMe transfer split size from `max_hw_sectors_kb` only. The NVMe passthrough path uses one scatter-gather segment per page, so a transfer that fits the byte limit can still exceed `max_segments` and fail with `EINVAL`.

- Caps auto-detected `max_data_transfer_size` by both `max_hw_sectors_kb * 1024` and `max_segments * page_size`.
- Falls back to the existing `max_hw_sectors_kb` behavior when `max_segments` is unavailable.
- Updates raw-block MP storage docs to describe the auto-detection bounds.
- Adds tests for the uncapped case, the `max_segments` cap, and the missing-`max_segments` fallback.

- Explicitly configured positive `max_data_transfer_size` values keep the existing behavior and must still be `block_align` aligned.
- Auto-detected values are still aligned down to `block_align` after applying device queue limits.

Tested by:

```bash
pytest -xvs tests/v1/storage_backend/test_raw_block_uring_cmd.py
```

**If applicable**:


DongDongJu (approved):
LGTM.

ankit-sam (approved):
LGTM

DongDongJu:
Hello @sammshen Could you take a look by any chance?

zhengfeihe (approved):
LGTM!

zhengfeihe:
it only tightens the auto-detected cap when max_data_transfer_size is not explicitly configured, low risk change, I will merge it

PR #3884, opened 2026-06-25
Fix LMCache V2 In-Process Memory Exhaustion - Ref Count Leaks

annguy3n:

In [LMCacheEngine.retrieve()](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/cache_engine.py#L777), when chunks are retrieved and `remove_after_retrieve` is `False`, the engine was not decrementing the reference count `ref_count_down()` or unpinning unpin()` the `MemoryObj` instances once retrieval completed. As a result:

* The reference counts remained elevated (≥2), preventing the LRU cache eviction policy from reclaiming these memory pages.
* Staging memory blocks remained pinned indefinitely, leaking CPU memory blocks. Under concurrent or long-running serving workloads, the allocator pool would run out of free pages and crash the serving container.

We uncovered the bug during continuous GKE serving benchmarks under high concurrency. 

We updated the cleanup block in `LMCacheEngine.retrieve()` to unconditionally call `memory_obj.ref_count_down()` and `memory_obj.unpin()` for all successfully retrieved chunks when they are returned to the caller.

It is backward-compatible and does not alter the storage engine's interface or retrieval path logic itself.

**If applicable**:

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -929,7 +929,9 @@ class LMCacheEngine:
                 # See pd_backend.py line 605 TODO comment.
                 if self._is_sync_pd_backend():
                     memory_obj.ref_count_down()
-            elif not self.async_loading:
+            else:
+                if memory_obj.is_pinned:
+                    memory_obj.unpin()
                 memory_obj.ref_count_down()
 
         retrieved_tokens = torch.sum(ret_mask)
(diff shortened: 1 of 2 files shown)


sammshen (approved):
LGTM!

ApostaC (approved):
LGTM!

ApostaC (changes requested):
@annguy3n Please fix the CPU UT. Thanks!

ApostaC (approved):

PR #3885, opened 2026-06-25
feat: implement Cloud Bigtable L2 Adapter and Layer-Group Sharding optimizations

annguy3n:

Implements the out-of-process remote L2 cache backend for Google Cloud Bigtable `BigtableL2Adapter`, introduces Layer-Group Sharding to handle multi-megabyte prompt caches, and applies critical performance and safety optimizations to make it an enterprise-grade caching solution for large language models (LLMs).

Design Key Features:

1. Out-of-Process Remote L2 Caching (BigtableL2Adapter)
* Implements the L2AdapterInterface
* Includes `BigtableL2KeyEncoder` to generate highly distributed, collision-free row keys containing SHA-256 chunk prefixes, model fingerprints, and worker/rank identifiers to prevent partition hot-spotting.

2. Layer-Group Sharding (In-Process & Out-of-Process)
* Introduces the `BigtablePayloadSharder` which slices massive contiguous tensor byte arrays along the layer dimension into configurable column groups (e.g. layers_0_9).
* Slices are stored in distinct column qualifiers under the same row, keeping individual cells under the optimal <10 MiB limit for LSM-tree database compaction.
* Enables pipelined cache loading: vLLM can download the first layer group and begin GPU prefill execution while subsequent layer groups are loaded concurrently in the background.

3. Row-Key Collision Prevention under use_layerwise=True
When LMCache runs with use_layerwise=True (serving caches layer-by-layer), the row keys generated by both BigtableSchema (in-process) and BigtableL2KeyEncoder (L2) automatically append the @layer_{layer_id} suffix to isolate layer data and prevent collisions.

Added Documentation & Tutorials

* Tutorial 1: Serving Massive Models (FP8 Compression)
* Tutorial 2: 3-Tier Hybrid Cascade (Local CPU -> Redis -> Bigtable SSD)
* Tutorial 3: Multi-Tier Storage Cascading (SSD Hot Tier -> HDD Cost-Effective Archive): Detailing both native Google Cloud Tiered Storage (zero configuration, automatic background age-based tiering) and manual LMCache cascading.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/bigtable_key_encoder.py
+++ b/lmcache/v1/distributed/l2_adapters/bigtable_key_encoder.py
@@ -1,0 +1,53 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Bigtable key encoder for translating ObjectKeys to row keys.
+"""
+
+# First Party
+from lmcache.v1.distributed.api import ObjectKey
+
+
+class BigtableL2KeyEncoder:
+    """Encoder to translate ObjectKeys into Bigtable row keys using a template."""
+
+    def __init__(
+        self,
+        template: str = "{hash_prefix}@{model}@{rank}@{group}@{hash}@{salt}",
+        layer_group_size: int = 0,
+    ):
+        self.template = template
+        self.layer_group_size = layer_group_size
+
+    def encode_row_key(self, key: ObjectKey) -> bytes:
+        """Translates a logical ObjectKey into a physical Bigtable row key.
+
+        Formats the row key using the configured template, injecting the model
+        name, rank, group, chunk hash, and optional cache salt.
+
+        Args:
+            key: The ObjectKey to encode.
+
+        Returns:
+            The formatted row key as bytes.
+        """
+        template = self.template
+        # Strip @{salt} if salt is empty to prevent a trailing @
+        if not key.cache_salt and template.endswith("@{salt}"):
+            template = template[:-7]
+
+        model_name = key.model_name
+        if self.layer_group_size > 0:
+            model_name = f"{model_name}@lg{self.layer_group_size}"
+
+        hash_hex = key.chunk_hash.hex()
+        row_key_str = template.format(
+            hash_prefix=hash_hex[:4],
+            model=model_name,
+            rank=f"{key.kv_rank:08x}",
+            group=f"{key.object_group_id:x}",
+            hash=hash_hex,
+            salt=key.cache_salt,
+        )
+        if hasattr(key, "layer_id"):
+            row_key_str += f"@layer_{key.layer_id}"
+        return row_key_str.encode("utf-8")
--- a/lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
@@ -944,7 +944,9 @@ class BigtableL2Adapter(L2AdapterInterface):
                 any_success = False
 
                 for (idx, key, key_str), result in zip(indexed, results, strict=True):
-                    if isinstance(result, BaseException):
+                    if isinstance(result, asyncio.CancelledError):
+                        raise result
+                    elif isinstance(result, BaseException):
                         last_error = result
                         logger.error(
                             "BigtableL2Adapter lookup failed for %s: %s",
@@ -1045,7 +1047,9 @@ class BigtableL2Adapter(L2AdapterInterface):
                 for (idx, key, key_str, obj), result in zip(
                     indexed, results, strict=True
                 ):
-                    if isinstance(result, BaseException):
+                    if isinstance(result, asyncio.CancelledError):
+                        raise result
+                    elif isinstance(result, BaseException):
                         last_error = result
                         logger.error(
                             "BigtableL2Adapter load failed for %s: %s",
--- a/lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
@@ -1,0 +1,1121 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Cloud Bigtable L2 adapter.
+"""
+
+# Future
+from __future__ import annotations
+
+# Standard
+from collections import defaultdict
+from typing import TYPE_CHECKING, Any, Optional
+import asyncio
+import threading
+
+if TYPE_CHECKING:
+    from lmcache.v1.distributed.internal_api import L1MemoryDesc
+
+# Third Party
+from google.cloud.bigtable.data import (
+    BigtableDataClientAsync,
+    DeleteAllFromRow,
+    RowMutationEntry,
+    SetCell,
+    row_filters,
+)
+import google.api_core.exceptions as google_exceptions
+import google.auth
+
+# First Party
+from lmcache.logging import init_logger
+from lmcache.native_storage_ops import Bitmap
+from lmcache.utils import TTLCache
+from lmcache.v1.distributed.api import MemoryLayoutDesc, ObjectKey
+from lmcache.v1.distributed.internal_api import L2StoreResult
+from lmcache.v1.distributed.l2_adapters.base import (
+    L2AdapterInterface,
+    L2TaskId,
+)
+from lmcache.v1.distributed.l2_adapters.bigtable_key_encoder import (
+    BigtableL2KeyEncoder,
+)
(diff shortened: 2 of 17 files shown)


lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
```diff
@@ -0,0 +1,973 @@
...
+                                    operation_timeout=self._config.write_timeout_sec,
+                                    **kwargs,
+                                )
+                            )
+                        results = await asyncio.gather(*tasks, return_exceptions=True)
+                        for res in results:
+                            if isinstance(res, Exception):
+                                return res
+                        return None
+
```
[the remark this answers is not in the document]
annguy3n: Bigtable's 200 MiB single-row mutation limit. built dual-mode write where LMCache dynamically bundles mutations if the chunk size is safely under 150 MiB, but falls back to separate concurrent mutation requests if the chunk is 150 MiB or larger.

lmcache/v1/storage_backend/connector/bigtable_connector.py
```diff
@@ -350,82 +399,100 @@ async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
...
+                    *[
+                        table.mutate_row(
+                            entry.row_key,
+                            entry.mutations,
+                            operation_timeout=self.cfg.write_timeout_sec,
+                            **kwargs,
+                        )
+                        for entry in entries
+                    ]
+                )
```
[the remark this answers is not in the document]
annguy3n: Similar to the L2 adapter, this uses dual-mode writing to bypass gRPC message limits. Chunks under 150 MiB are already combined into a single mutate_row request. Chunks exceeding 150 MiB are split into parallel requests to ensure each payload fits within transport limits.

lmcache/v1/storage_backend/connector/bigtable_connector.py
```diff
@@ -563,120 +653,126 @@ def support_batched_put(self) -> bool:
...
+                    None, self.sharder.shard, data_bytes
+                )
+                for qualifier, shard_data in shards.items():
+                    entry = RowMutationEntry(
+                        row_key,
+                        [SetCell(self.cfg.family_name, qualifier, shard_data)],
+                    )
+                    current_batch.append(entry)
+                    current_batch_keys.append(key.to_string())
+                    current_batch_size += len(shard_data)
```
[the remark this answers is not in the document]
annguy3n: Combining shards for a large row (up to 240 MiB) into a single RowMutationEntry would result in a single massive entry in the batch, causing the bulk write to fail the gRPC transport limit.

The current logic splits shards into separate entries and flushes the batch incrementally when MAX_BATCH_SIZE_BYTES (30 MiB) is exceeded, keeping individual request payloads within a safe range.

lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
```diff
@@ -0,0 +1,973 @@
...
+                                and cell.qualifier == self._column_name_bytes
+                            ):
+                                val = cell.value
+                                break
+                        if val is None:
+                            logger.warning(
+                                f"Column {self._config.column_name} not "
+                                f"found in row {key_str}"
+                            )
+                            continue
```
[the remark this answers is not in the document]
annguy3n: combining mutations for large chunks (>= 150 MiB) would exceed Bigtable/gRPC single-request payload limits (200MB hard limit / 30MB safety batch limit).

lmcache/v1/storage_backend/connector/bigtable_connector.py
```diff
@@ -89,6 +146,25 @@ def __init__(
         # Use native AsyncPQExecutor to run pure coroutines directly on the event loop
         self.pq_executor = AsyncPQExecutor(loop, max_workers=self.cfg.thread_pool_size)
 
+        # Compute total layers dynamically
+        num_layers = 0
+        for shape in self.meta_shapes:
+            if len(shape) >= 2:
+                num_layers += shape[1]
```
DongDongJu, line 153: is this right assumption?

tests/v1/storage_backend/test_raw_block_uring_cmd.py
```diff
@@ -308,7 +308,7 @@ def test_uring_cmd_explicit_transfer_limit_must_be_block_aligned():
     with pytest.raises(
         ValueError,
         match=r"max_data_transfer_size \(5000\) must be a multiple of "
-        "block_align \(4096\)",
+        r"block_align \(4096\)",
```
DongDongJu, line 311: Thanks for the fix.

lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
```diff
@@ -0,0 +1,1224 @@
...
+        self._next_task_id: L2TaskId = 0
+        self._completed_store_tasks: dict[L2TaskId, L2StoreResult] = {}
+        self._completed_lookup_tasks: dict[L2TaskId, Bitmap] = {}
+        self._completed_load_tasks: dict[L2TaskId, Bitmap] = {}
+
+        # Client-side locking (refcount per ObjectKey)
+        self._locked_keys: dict[ObjectKey, int] = defaultdict(int)
+
+        # Thread-safe exists cache
+        self._exists_cache = TTLCache(
```
DongDongJu, line 339: So what is the exact role of TTLCache in mp mode?

DongDongJu (commented):
Hello @annguy3n,
Thanks for the contribution.
I left few questions.

lmcache/v1/distributed/l2_adapters/bigtable_l2_adapter.py
annguy3n: It's to reduce database lookup overhead and redundant store tasks.

When multiple GPU workers query L2 for the same prompt chunks concurrently:
- The first worker triggers a Bigtable read to check if the key exists.
- The result is cached (hits as True, misses as False—negative caching).
- Subsequent queries from all other workers are answered instantly from the local TTLCache, avoiding redundant round-trips to Bigtable.

Before starting a multi-megabyte store task, the adapter checks `self._exists_cache.get(key_str)`. If the key is already marked as stored (either from a previous local write or a successful lookup), the write is skipped. This saves considerable network bandwidth and database write capacity.

lmcache/v1/storage_backend/connector/bigtable_connector.py
annguy3n: In [metadata.get_shapes()](https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/metadata.py#L84) structures shapes such that the number of layers is always at index 1:

- Standard models: [kv_size, num_layers, chunk_size, hidden_dim]
- Sharded layer groups: [kv_size, group.num_layers, num_tokens, hidden_dim]

Summing shape[1] dynamically computes the correct total number of layers across all model layouts. We though could create a helper function in `metadata` to have `metadata.get_num_layers` or something like that.

lmcache/v1/distributed/l2_adapters/bigtable_key_encoder.py
```diff
@@ -0,0 +1,53 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Bigtable key encoder for translating ObjectKeys to row keys.
+"""
+
+# First Party
+from lmcache.v1.distributed.api import ObjectKey
+
+
+class BigtableL2KeyEncoder:
```
sammshen, line 10: LGTM, why don't we just put the hash first? 

Also dumb question, even in the previous model, lexicographical should still randomize by hash since this is the first differing point in the key right? so does the order really matter?

annguy3n: Great questions!

The order does matter. Bigtable sorts all row keys lexicographically and splits them into chunks (called tablets) across different database servers.

We use the hash_prefix (4 char salt) first instead of full hash to balance anti-hotspotting and scanability. 

- If we put the model first, Bigtable puts all files for that model on the same server. Under heavy load, all writes hit that single server, overloading it and causing timeouts (known as "hotspotting").
- If we put full hash first, row keys are scattered completely randomly across the database - this prevents hotspotting but makes it impossible to scan or clean up keys by model, rank, or group without doing a slow, expensive full-table scan.

sammshen (approved):
LGTM!

DongDongJu (approved):
LGTM

DongDongJu (approved):
LGTM

PR #3887, opened 2026-06-25
[CLI] Output the model answer in `lmcache query engine`

sammshen:

`lmcache query engine` streamed the model's completion text but then threw it away, printing only token/latency stats. So there was no way to see what the engine actually answered. This makes the command output the answer.

- `_stream` now returns the completion text (`answer`) alongside the existing metrics.
- `Request.send_request` returns a `(answer, metrics)` tuple instead of just the metric map; the answer is split out so it never leaks into the numeric metric loop.
- `EngineCommand` renders the answer as an `Answer` field, so it shows up in terminal, JSON, and `--output` file results next to the stats.

Example:
```
================= Query Engine =================
Model:                      MiniMaxAI/MiniMax-M3
Answer:                                    Paris
Input tokens:                                188
Output tokens:                                32
--------------- Latency Metrics ----------------
TTFT (ms):                                 25.32
...
```

Only the `query engine` path is touched; the metrics framework and other commands are unchanged. Verified end-to-end against a live OpenAI-compatible engine and via the unit tests.

**If applicable**:

### DIFF
--- a/lmcache/cli/commands/query/_request.py
+++ b/lmcache/cli/commands/query/_request.py
@@ -130,7 +130,8 @@ def _stream(
     chat: bool,
     max_tokens: int,
 ) -> dict[str, Any]:
-    """POST with ``stream: true``; parse SSE; return TTFT/TPOT and token metrics."""
+    """POST with ``stream: true``; parse SSE; return the completion text plus
+    TTFT/TPOT and token metrics."""
     payload = {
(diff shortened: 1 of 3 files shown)


ApostaC (approved):
LGTM!

deng451e (approved):
LGTM!

PR #3891, opened 2026-06-26
[Fix][RawBlock] Bounce unaligned buffers in io_uring uring_cmd read paths

nayeonikim:

Fixes an EINVAL bug in the Rust `RawBlockDevice` read path when
`use_uring_cmd=true` and the caller passes a non-page-aligned buffer for a
multi-page transfer. Affects both `read_uring` and `batched_read`.

NVMe passthrough (`io_uring_cmd`) builds a PRP (Physical Region Page) list
for transfers larger than one page, and every entry in that list must be
page-aligned. Today:

* `read_uring` only checks pointer alignment when `use_odirect=true`. Under
  `use_uring_cmd=true` (which forces `use_odirect=false` on char devices) an
  unaligned destination pointer is submitted as-is and the kernel returns
  `[Errno 22] io_uring I/O error`.
* `batched_read` has no bounce path at all. With `use_odirect=true` it
  rejects unaligned buffers with a `PyValueError`; with `use_uring_cmd=true`
  it submits them unchanged and hits the same EINVAL.

The fix extends the alignment check to cover `use_uring_cmd` and routes
unaligned or short buffers through the existing `AlignedBuf` bounce
machinery (`IoSubmission.bounce` + `copy_from_bounce_buffer` in
`handle_completion_result`).

* `read_uring`: broaden `ptr_aligned`'s gate from `self.use_odirect` to
  `self.use_odirect || self.use_uring_cmd`. The existing
  `use_bounce = !ptr_aligned || cap < total_len` branch and the
  aligned-bounce submission path below handle the rest with no further
  changes.
* `batched_read`: hoist `use_uring_cmd` out before releasing the GIL, then
  for each entry compute the same `ptr_aligned` / `use_bounce` decision per
  item. When bouncing, allocate an `AlignedBuf`, route the submission's
  `ptr_addr` through it, and populate `bounce` / `original_ptr` /
  `payload_len` so the completion handler copies the result back to the
  caller. The previous strict reject for `cap < total_len` and the
  `"O_DIRECT requires aligned buffers"` error are dropped (the bounce now
  covers both cases); `cap == 0` remains a hard error, and the O_DIRECT
  `offset` / `total_len` alignment checks (kernel-required for the I/O
  itself) are kept.

No Python changes. The Rust bounce is the single defense for every caller.

* Scope is read paths only. The same `use_odirect`-only alignment check
  exists on the write side (`write_uring` / `batched_write` and the shared
  `prepare_iouring_write_buffer` helper introduced by an in-flight write
  PR). The write-side fix follows the same one-line pattern and will land
  as a separate PR on top of that one.
* Verified on a Samsung PM1753 NVMe char device (`/dev/ng10n1`, set via
  `LMCACHE_TEST_CHAR_DEVICE`): 4 new tests pass and the existing tests in
  `test_raw_block_uring_cmd.py` and `test_raw_block_core.py` remain green.
  `cargo build --release` and `pre-commit run` (including `Rust Format
  Check` and `Rust Clippy Check`) all pass.

**If applicable**:


DongDongJu (approved):
Hello @nayeonikim,

Thanks for the fix. LGTM

nayeonikim:
@sammshen @ApostaC @deng451e — approved by @DongDongJu since Jul 16, and it just needs a second approving review to be mergeable. Could someone take a look when you have time?

nayeonikim:
Hi @zhengfeihe — this PR touches `rust/raw_block/src/lib.rs`, which falls under the `/rust/` CODEOWNERS entry you're on. Could you self-assign and take a look when you have time? Thanks!

zhengfeihe:
@nayeonikim 

Hi~ I will take a look shortly.

zhengfeihe (approved):
LGTM!

zhengfeihe:
Oh

I think shall be fixed by this https://github.com/LMCache/LMCache/pull/4488

Trigger the ci again

PR #3892, opened 2026-06-26
feat(s3): opt-in per-cache_salt bucket isolation for multi-tenancy

divyvasal:
## Summary

Opt-in **per-`cache_salt` S3 bucket isolation** for the S3 L2 adapter: route each tenant's KV to its **own physical S3 bucket**, keyed by `cache_salt`. On a shared multi-tenant inference box, no two tenants share a bucket — one tenant's writes never consume or evict another's cache, and there is no cross-tenant read path. Default behavior is **unchanged** (single bucket); the feature is entirely opt-in.

## Motivation

Today the S3 backend is single-bucket: every tenant's KV lands in one bucket and `cache_salt` only prefixes the object key. For multi-tenant serving that means **shared capacity** (one tenant can fill or LRU-evict another's entries) and a shared blast radius. Physical per-tenant buckets give hard isolation plus independent lifecycle/capacity per tenant.

## Design (default-safe)

Two new optional `S3L2AdapterConfig` fields, both defaulting to **exact current behavior**:

- `s3_bucket_mode`: `"single"` (default) | `"per_cache_salt"`
- `s3_bucket_template`: bucket-name format for per-salt mode; supports `{salt}` (sanitized `cache_salt`) and `{base}` (the configured bucket name), e.g. `"kv-cache-{salt}"` or `"{base}-{salt}"`.

In `per_cache_salt` mode:

- **Routing** — `_make_request` (PUT/GET/HEAD/DELETE) sets the virtual-hosted `Host` per key. The CRT `S3Client` is **region-bound, not bucket-bound**, so a single client serves all tenant buckets — only the per-request `Host` changes. Unsalted keys and `single` mode keep the configured bucket.
- **Listing / eviction** — `list_l2_keys` iterates the base bucket plus every observed tenant bucket with a composite cursor, so the eviction controller sees keys across all tenant buckets and issued deletes route back to the correct bucket.
- **Provisioning** — the first write to a tenant bucket idempotently creates it (plus a configurable lifecycle TTL) via boto3; failures are logged, not raised. Can be turned off to require buckets be pre-created by an onboarding hook.

The routing logic lives in a new **dependency-free** module `s3_bucket_router.py` (no `awscrt` / `lmcache` imports) so it is unit-testable and reviewable in isolation.

## Tests

- `tests/v1/distributed/test_s3_bucket_router.py` — 10 unit cases: sanitize, `single`-mode no-op, per-salt routing, distinct tenants → distinct buckets, unsalted → base, `{base}`/`{salt}` placeholders, cross-bucket listing, cursor roundtrip.
- **Validated end-to-end** on a multi-replica 8×GPU box (vLLM + LMCache MP, S3 L2), not just unit tests:
  - **Isolation** — two tenants → two separate buckets, `0` objects leaked into the base/other bucket.
  - **Read reuse** — identical prompt twice under one tenant → cold prefill then a full KV cache hit (the load path works through per-tenant routing).
  - **Provisioning** — tenant buckets auto-created with the lifecycle TTL on first write.
  - **Eviction** — under sustained load the L2 eviction controller crossed its watermark and evicted (deleted) from the per-tenant bucket; footprint stayed bounded by the cap.
  - `single` mode is byte-for-byte unchanged (covered by the existing `test_s3_l2_adapter.py`).

## Notes

Draft for now — happy to take feedback on the config surface and on whether auto-provisioning belongs in the adapter or in a separate onboarding hook.

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/s3_bucket_router.py
+++ b/lmcache/v1/distributed/l2_adapters/s3_bucket_router.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Pure, dependency-free bucket routing for the S3 L2 adapter (multi-tenant isolation).
+
+Kept import-free (no awscrt / lmcache) so the routing logic is unit-testable in isolation and
+trivial to review. The adapter delegates here.
+"""
+from __future__ import annotations
+
+# Standard
+import base64
+import json
+import re
+
+_SANITIZE = re.compile(r"[^a-z0-9-]+")
+
+
+def sanitize_salt(salt: str) -> str:
+    """cache_salt → an S3-bucket-name-safe fragment (lowercase [a-z0-9-], trimmed)."""
+    return _SANITIZE.sub("-", (salt or "").lower()).strip("-")
+
+
+def salt_of_key(key_str: str) -> str:
+    """Extract the cache_salt from a stored object key.
+    Key format: ``<model>@<rank>@<group>@<hash>[@<cache_salt>]`` ('@' barred in model & salt)."""
+    parts = (key_str or "").split("@")
+    return parts[4] if len(parts) >= 5 else ""
+
+
+def resolve_bucket_host(
+    key_str: str,
+    *,
+    base_endpoint: str,
+    mode: str = "single",
+    template: str = "{base}-{salt}",
+) -> str:
+    """Virtual-hosted S3 Host for a key.
+
+    - mode 'single' (default = upstream behavior): always the configured base bucket.
+    - mode 'per_cache_salt': a *salted* key routes to its own bucket named by ``template``
+      (``{salt}`` = sanitized cache_salt, ``{base}`` = base bucket name). Unsalted keys, and any
+      salt that sanitizes to empty, fall back to the base bucket.
+
+    base_endpoint is virtual-hosted: ``<bucket>.s3.<region>.amazonaws.com``.
+    """
+    if mode != "per_cache_salt":
+        return base_endpoint
+    safe = sanitize_salt(salt_of_key(key_str))
+    if not safe:
+        return base_endpoint
+    base_bucket, _, suffix = base_endpoint.partition(".")
+    if not suffix:
+        return base_endpoint
+    bucket = template.format(salt=safe, base=base_bucket)
+    return f"{bucket}.{suffix}"
+
+
+# ---- per-bucket listing / eviction support ---------------------------------
+def bucket_hosts(
+    base_endpoint: str,
+    seen_salts: set[str],
+    mode: str = "single",
+    template: str = "{base}-{salt}",
+) -> list[str]:
+    """Ordered list of bucket Hosts the adapter must list for eviction: the base bucket plus
+    one per observed tenant salt (per_cache_salt mode). 'single' mode → just the base."""
+    if mode != "per_cache_salt":
+        return [base_endpoint]
+    base_bucket, _, suffix = base_endpoint.partition(".")
+    hosts = [base_endpoint]
+    for salt in sorted(s for s in seen_salts if s):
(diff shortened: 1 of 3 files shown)


divyvasal:
Thanks for the review — addressed all points in the latest push:

- **Provisioning race (`_ensure_bucket`)** — added a dedicated `_bucket_creation_lock`; the bucket is now recorded in `_ensured_buckets` **only after** a successful `create_bucket` + lifecycle (with a double-checked `_lock` guard), so a failed create is never cached as "done" and a concurrent writer can't see a not-yet-created bucket.
- **`us-east-1` location constraint** — `create_bucket` now omits `CreateBucketConfiguration` when `region == "us-east-1"` and passes `LocationConstraint` otherwise.
- **boto3 credentials** — the management client is now built with `aws_access_key_id` / `aws_secret_access_key` from the config (falls back to the default chain when unset, matching the data-plane client).
- **`_seen_salts` mutation** — now under `self._lock`.
- **Template validation** — `S3L2AdapterConfig` validates `s3_bucket_template` at construction time (`.format(salt=…, base=…)`), raising `ValueError` on an unknown placeholder / bad spec instead of failing on the first PUT.
- **Cross-bucket pagination** — `list_l2_keys` now accumulates across tenant buckets up to the page budget (requesting only `max_keys - len(entries)` per bucket) instead of returning after the first non-empty bucket, so listing many small tenant buckets no longer costs a round-trip each.

divyvasal:
@ApostaC @maobaolong @chunxiaozheng — when you have a cycle, could one of you take a review pass on this?

Small, fully **opt-in** S3 change: routes KV objects to a per-`cache_salt` bucket so multi-tenant deployments don't share one L2 namespace (default off → no behavior change for existing users). Gemini's comments are addressed; `DCO` + `k3-unit-tests` are green. The red `unit-tests-amd` is the pre-existing ROCm-benchmark breakage currently failing on `dev` itself (`6a57f42b`) and other open PRs — unrelated to this diff.

It's an independent sibling of #3909 (both S3-L2, neither stacked on the other), so they can be reviewed in any order.

divyvasal:
Production validation update — this branch is now serving real multi-tenant traffic in our inference fleet (applied on top of 0.5.1.dev, MP server + engine_driven transport, vLLM 0.24 TP4/TP8 on H200/B200):

- Salted requests (`cache_salt=org:<id>`) store and retrieve KV against their own physical bucket (`<template>-org-<id>`); ~24MB chunk objects, verified via bucket listings and the adapter's usage gauge tracking L1 write-through 1:1.
- Unsalted traffic (health probes) falls through to the base bucket — no behavior change for deployments that never set a salt, matching the opt-in design.
- Auto-creation, cross-region template suffixing, and the eviction controller's per-bucket accounting all behaved under a multi-hour soak with model park/wake cycles.

Two adjacent issues we hit while productionizing (both pre-existing, not introduced by this PR): #4068 — skipped write reservations leak L1 write locks / publish unwritten objects; #4069 — worker rank crash when the MP server restarts under SHM transport. Fixes/details there.

@ApostaC @maobaolong @chunxiaozheng — review pass would be appreciated; we're carrying this as a wheel patch in production until it lands.

divyvasal:
@ApostaC gentle ping — CI is green and the PR is mergeable; so far only gemini-code-assist has reviewed. Would appreciate a maintainer review when you have a moment. It's fully opt-in (per-`cache_salt` bucket isolation, default off), so there's no behavior change for existing users. Happy to address any feedback.

divyvasal:
Bumping this. It's been open a few weeks with only the Gemini bot so far. It's opt-in per cache_salt, so nothing changes unless you turn it on. Any chance one of you could take a pass? @ApostaC @maobaolong

divyvasal:
Revived and rebased onto current `dev` to clear the stale state. No conflicts; the branch still includes the review-feedback changes.

PR #3894, opened 2026-06-26
[Bugfix] trtllm: cap kernel object batch at 4 to match c_ops limit (#3889)

### LINKED ISSUE #3889: TRT-LLM connector: _TRTLLM_KERNEL_BATCH_SIZE (32) exceeds c_ops kernel's 4-object cap → store fails for prompts >4 chunks
The TensorRT-LLM GPU connector batches KV transfers in groups of `_TRTLLM_KERNEL_BATCH_SIZE = 32` (`lmcache/v1/gpu_connector/gpu_connectors.py`), but the c_ops kernel `multi_layer_block_kv_transfer` accepts at most 4 memory objects per call. Any store/load with more than 4 chunks raises:

```
RuntimeError: Expected 1-4 LMCache objects, got <N>
```

after which the TRT-LLM executor loop dies.

### Where it comes from
- `gpu_connectors.py`: `_TRTLLM_KERNEL_BATCH_SIZE = 32`, used as the batch stride in `_batched_transfer` (passes up to 32 object pointers per `lmc_ops.multi_layer_block_kv_transfer` call).
- `csrc/mp_mem_kernels.cuh`: `struct MemoryObj4 { ScalarType* objects[4]; int num_objects; }`
- `csrc/mp_mem_kernels.cu`: `TORCH_CHECK(num_objects >= 1 && num_objects <= 4, "Expected 1-4 LMCache objects, got ", num_objects);`

The kernel is hard-capped at 4 by design, so a Python batch size of 32 looks inconsistent with it.

### Reproduce
TRT-LLM + `connector: lmcache`, chunk_size 256, send any prompt that produces more than 4 chunks (>~1024 tokens). Short prompts (≤4 chunks) don't trigger it, which is likely why the current E2E tests pass.

### Fix that worked for me
Set `_TRTLLM_KERNEL_BATCH_SIZE = 4` (the kernel's max) — `_batched_transfer` already loops with that stride. Verified on gpt-oss-120b (TRT-LLM 1.3, lmcache 0.5.0): stores/retrieves succeed at ~45 GB/s, no errors.

Present on `main` @ 6f444eb.

Happy to send this as a PR if useful.

nevasini1:
## Summary

Fixes #3889.

The TRT-LLM GPU connector (`TRTLLMGPUConnector._batched_transfer`) batched KV transfers in groups of `_TRTLLM_KERNEL_BATCH_SIZE = 32` object pointers per `lmc_ops.multi_layer_block_kv_transfer` call. The c_ops kernel packs pointers into a fixed `MemoryObj4` struct (`csrc/mp_mem_kernels.cuh`) and **hard-rejects any call with more than 4 objects**:

```cpp
// csrc/mp_mem_kernels.cu
TORCH_CHECK(num_objects >= 1 && num_objects <= 4,
            "Expected 1-4 LMCache objects, got ", num_objects);
```

So any store/load spanning more than 4 chunks (~>1024 tokens at `chunk_size=256`) raised `RuntimeError: Expected 1-4 LMCache objects, got <N>` and killed the TRT-LLM executor loop. Short prompts (≤4 chunks) never tripped it, which is why the existing E2E tests pass.

## Fix

- Introduce a named constant `_MULTI_LAYER_KERNEL_MAX_OBJECTS = 4` documenting the kernel ABI cap, and set the batch stride to it. `_batched_transfer` already loops with this stride, so the only effect is more (smaller) kernel launches — no semantic change. The kernel has no larger batched variant; 4 is a genuine ABI ceiling.

## Test

- Adds `TestTRTLLMBatchedTransferCap` — a **GPU-free** regression test that stubs the kernel call and drives `_batched_transfer` with 9 chunks (the failing >4-chunk case), asserting no single kernel invocation ever receives more than 4 object pointers while every chunk is still transferred exactly once.

Verified locally: `ruff check`, `ruff format --check`, `isort --check-only`, `codespell` all clean.

### DIFF
--- a/lmcache/v1/gpu_connector/gpu_connectors.py
+++ b/lmcache/v1/gpu_connector/gpu_connectors.py
@@ -1933,7 +1933,14 @@ class SGLangLayerwiseGPUConnector(GPUConnectorInterface):
         return torch.Size([num_tokens, 2, self.hidden_dim_size])
 
 
-_TRTLLM_KERNEL_BATCH_SIZE = 32
+# The ``multi_layer_block_kv_transfer`` kernel packs object pointers into a
+# fixed ``MemoryObj4`` struct (``csrc/mp_mem_kernels.cuh``) and hard-rejects any
+# call with more than 4 objects (``TORCH_CHECK(num_objects <= 4, ...)`` in
+# ``csrc/mp_mem_kernels.cu``). The Python batch stride must therefore never
+# exceed this cap, otherwise stores/loads spanning more than 4 chunks raise
+# ``RuntimeError: Expected 1-4 LMCache objects, got <N>``.
+_MULTI_LAYER_KERNEL_MAX_OBJECTS = 4
+_TRTLLM_KERNEL_BATCH_SIZE = _MULTI_LAYER_KERNEL_MAX_OBJECTS
 
 
 class TRTLLMGPUConnector(GPUConnectorInterface):
--- a/tests/v1/test_trtllm_integration.py
+++ b/tests/v1/test_trtllm_integration.py
@@ -416,34 +416,28 @@ class TestTRTLLMBatchedTransferCap:
             <= gpu_connectors._MULTI_LAYER_KERNEL_MAX_OBJECTS
         )
 
-    def test_batched_transfer_never_exceeds_kernel_cap(
+    def test_batched_to_gpu_never_exceeds_kernel_cap(
         self, monkeypatch: pytest.MonkeyPatch
     ) -> None:
         # Standard
+        from unittest.mock import MagicMock
         import contextlib
 
         # First Party
         from lmcache.v1.gpu_connector import gpu_connectors
         from lmcache.v1.gpu_connector.gpu_connectors import TRTLLMGPUConnector
-
-        # Build without __init__ to avoid creating real CUDA streams.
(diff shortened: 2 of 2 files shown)


hlin99:
thanks for the fix. can you check cq issue and comments from gemini first?

nevasini1:
Thanks @hlin99 — addressed both in e3d1257.

- **Code-quality (mypy) failure:** it was `tests/v1/test_trtllm_integration.py:474: Argument 1 to "_batched_transfer" ... incompatible type "list[_FakeMemoryObj]"; expected "list[MemoryObj]"`. Fixed by the test rewrite below (no more fake type, no private call).
- **gemini-code-assist feedback** (bypassing the constructor with `object.__new__` and testing the private `_batched_transfer`): the test now builds the connector via the public constructor, registers the pool with `register_kv_caches()`, and drives the public `batched_to_gpu()` with real `TensorMemoryObj` instances and real block ids. Only the GPU-specific bits are stubbed — `torch.cuda.Stream` / `torch.cuda.stream` (for CPU instantiation) and the `multi_layer_block_kv_transfer` kernel (to count object pointers per call). It still asserts no kernel call ever receives more than 4 objects while all 9 chunks transfer exactly once.

The production one-line fix (cap the batch stride at the kernel's 4-object limit) is unchanged.

PR #3903, opened 2026-06-26
[operator] replace privileged:true with unconfined profiles

xneng:

This PR removes `privileged: true` from operator-managed LMCache and CacheBlend engine DaemonSets. Instead, the engine container now uses `seccompProfile: Unconfined` and `appArmorProfile: Unconfined`, which is the narrower privilege surface needed for CUDA IPC receiving/opening via cudaIpcOpenMemHandle.

Kubernetes 1.30+ is required for the new securityContext.appArmorProfile field. Clusters older than 1.30 would need the legacy AppArmor annotation path, should we add support for legacy k8s version?

**If applicable**:

### DIFF
--- a/docs/source/locale/zh_CN/LC_MESSAGES/mp/operator.po
+++ b/docs/source/locale/zh_CN/LC_MESSAGES/mp/operator.po
@@ -98,8 +98,8 @@ msgid "Prerequisites"
 msgstr "先决条件"
 
 #: ../../source/mp/operator.rst:41
-msgid "Kubernetes 1.20+"
-msgstr "Kubernetes 1.20+"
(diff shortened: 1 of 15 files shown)


royyhuang:
Hi @xneng, previously we turn on the privileged mode for the pod mainly because we want the LMCache/Cacheblend daemonset to have visibility on GPUs but without requesting GPUs from GPUs device plugin (since doing so will starve the real serving engine pods' requests for GPU resource). So it is not a CUDA IPC reason to enable privileged mode. For CDUA IPC, I believe enabling HostIPC should suffice. 

I am not sure if this PR will fix the GPU visibility issue.

PR #3905, opened 2026-06-26
Route affinity thread pool on dense worker rank instead of hashed identity

YuhanLiu11:

The GPU affinity thread pool (STORE/RETRIEVE) routed each request with
`hash(zmq_identity) % num_workers`. The DEALER client never set an identity, so
the routing key was an opaque per-connection blob. Hashing N distinct clients
into N worker slots is not a bijection — by the birthday/pigeonhole effect,
distinct vLLM workers collide onto one worker thread while other threads sit
idle. In an 8-way TP run we saw the 8 ranks land on ~6 of 8 threads (two
doubled, two idle), which serializes those ranks' GPU transfers on a shared
thread (on top of the GIL) and starves some GPUs of a dedicated transfer thread.

This PR routes on a dense per-worker rank instead of a hash:

- `MessageQueueClient` gains an optional `routing_key` that stamps the worker's
  rank onto the ZMQ DEALER identity before connect.
- The server derives the affinity key via `_affinity_key_from_identity`:
  `int(identity)` when the client set a dense rank, otherwise falls back to
  `hash(identity)` — so this is fully backward compatible with any client that
  doesn't set one.
- The worker adapter (the client that issues STORE/RETRIEVE) sets its physical
  rank as the routing key. With `max_gpu_workers >= world_size`, each rank
  deterministically maps to its own worker thread (1:1), eliminating the
  cross-rank collisions.

Note: this removes the cross-rank serialization layered on top of the GIL; it
does not by itself reduce GIL contention.

- The worker adapter uses `vllm_worker_id` (physical rank, unique per worker /
  GPU), **not** `kv_worker_id`: for MLA models `kv_worker_id` collapses to 0
  across all TP ranks, which would give every worker the same DEALER identity
  and collide them on the ROUTER (only one peer keeps the identity; the rest
  never receive replies and time out).
- The scheduler adapter intentionally keeps an opaque identity — it issues only
  non-affinity requests (e.g. LOOKUP), and claiming a rank there would collide
  with the same-rank worker on the server's ROUTER socket.
- The pool logs each rank → worker-slot (thread) assignment once at INFO as a
  positive confirmation, and warns once if two distinct clients still share a
  slot (pool sized smaller than the number of clients, or non-dense keys).
- Sizing: a strict 1:1 mapping requires `--max-gpu-workers >= world_size`.

**If applicable**:


lmcache/v1/multiprocess/affinity_pool.py
```diff
@@ -41,6 +47,14 @@ def __init__(
         self._num_workers = max_workers
         self._queues: list[queue.Queue] = [queue.Queue() for _ in range(max_workers)]
         self._threads: list[threading.Thread] = []
+        # Collision tracking: maps a worker slot to the first affinity key
+        # routed there, so we can warn once when two distinct keys (clients)
+        # land on the same slot and get serialized onto one thread. Guarded by
+        # ``_collision_lock`` since submit() may be called from multiple
+        # threads.
+        self._collision_lock = threading.Lock()
```
ApostaC: Maybe we don't need this lock because this class is owned by a single-thread.

lmcache/v1/multiprocess/affinity_pool.py
```diff
@@ -41,6 +52,18 @@ def __init__(
...
         self._queues: list[queue.Queue] = [queue.Queue() for _ in range(max_workers)]
         self._threads: list[threading.Thread] = []
+        # Dynamic ``affinity_key -> worker slot`` assignment. The first time a
+        # key is submitted it is bound to the next free slot (round-robin) and
+        # cached here so all later tasks for that key reuse the same worker
+        # thread. ``_next_slot`` is the monotonically increasing arrival
+        # counter; ``slot = _next_slot % _num_workers``. Guarded by
+        # ``_assign_lock`` because submit() may be called concurrently from
+        # multiple threads; ``_overflow_warned`` makes the overflow warning
+        # fire at most once.
```
ApostaC: nit -- Simplify comment: we only need to talk about what the key and value of `self._key_to_slot` are.

lmcache/v1/multiprocess/affinity_pool.py
```diff
@@ -75,17 +102,91 @@ def _worker(q: queue.Queue) -> None:
...
+
+        The first ``max_workers`` distinct keys each get their own worker
+        thread regardless of their numeric values; further distinct keys wrap
+        around and share a worker (logged once at WARNING level). Tasks sharing
+        a key -- and tasks whose keys share a slot -- execute sequentially in
+        FIFO order on that one thread.
+
+        The first time each key is routed, its ``key -> worker slot (thread)``
+        assignment is logged once at INFO level, so a run's log positively
+        shows which worker thread each client was bound to.
```
ApostaC: nit: this can be removed.

ApostaC (commented):
Otherwise LGTM

lmcache/v1/multiprocess/affinity_pool.py
```diff
@@ -75,17 +86,73 @@ def _worker(q: queue.Queue) -> None:
...
+        """Return the worker slot bound to ``affinity_key``, assigning on first use.
+
+        The first time a key is seen it is bound to the next free slot
+        (``_next_slot % _num_workers``) and the binding is cached, so the
+        mapping is independent of the key's numeric value: any set of up to
+        ``_num_workers`` distinct keys lands on distinct worker threads. Once
+        more than ``_num_workers`` distinct keys have been seen, later keys wrap
+        around and share a slot with an earlier key; the first such overflow is
+        logged once at WARNING level.
+
```
ApostaC: nit: we don't need this. In general, the docstring should not describe the implementation details of the function (e.g., what variables it uses)

lmcache/v1/multiprocess/affinity_pool.py
```diff
@@ -28,6 +27,10 @@
 class AffinityThreadPool:
     """Thread pool that routes tasks to workers by affinity key.
 
+    Not thread-safe: ``submit()`` must be called from a single thread (the
+    ``MessageQueueServer`` main loop). The routing state is unsynchronized, so
+    dispatching submits from multiple threads would race on slot assignment.
```
ApostaC: nit: can be simplified a bit (only keep the first sentence)

```suggestion
    Not thread-safe: ``submit()`` must be called from a single thread (the
    ``MessageQueueServer`` main loop). 
```

ApostaC (commented):
Otherwise LGTM!

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3907, opened 2026-06-27
[Refactor]: Consolidate all grouping queries on EngineKVFormat (also fixes SGLang tests)

sammshen:
## What

The structural shape of a normalized `kv_caches` (cross-layer / K/V-split / per-layer-list) and the MLA modifier were classified in two drifting places: a Python `KVFormatSpec` ClassVar per spec, and hand-rolled `is_mla` helpers duplicated across `mem_kernels.cu` and `mem_kernels_sycl.cpp`. Detection branched on these via an ad-hoc per-layer slice that assumed one normalized entry per registered tensor, which broke SGLang's K/V-split `[keys, values]` layout with an `IndexError` at registration.

This moves the classification to a single source of truth on the `EngineKVFormat` itself, and the SGLang fix falls out of it.

## Changes

- New `csrc/engine_kv_format.h` holds the enum (split out of `kv_transfer_types.h`) plus `constexpr` `is_cross_layer` / `is_kv_list` / `is_layer_list` / `is_mla`, guarded `__host__ __device__` so the device kernels share them. pybind exposes them via `lmc_ops`; the duplicate `is_mla` is removed from the CUDA and SYCL kernels.
- `KVFormatSpec` and the specs drop their flag ClassVars; the `gpu_connector` facades read `lmc_ops`.
- `normalize_and_discover_per_layer_formats` front-loads on `is_layer_list_format`: a format that isn't a per-layer list is returned whole, so the K/V-split layout is never sliced — the SGLang fix is now a property of the format, not a special case. `_pick_layout_probe_tensor` reads the predicates too.

## Not in scope

`is_hnd` is intentionally left out: its live consumer is the CUDA `check_head_size` path (2 formats) and it had drifted from the Python specs (4 formats), so unifying it risks kernel behavior and needs a separate audited change. The dead Python `is_hnd` facade is removed.

## Tests

Golden classification + partition invariant over every format (via `lmc_ops`), and a per-layer regression covering the SGLang K/V-split, uniform, and mixed-format cases. Full kv_format / kv-layer-groups / kernel suites pass against a local CUDA rebuild.

### DIFF
--- a/csrc/engine_kv_format.h
+++ b/csrc/engine_kv_format.h
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+// Physical KV-cache memory layout an engine hands to LMCache, plus the
+// classification predicates over it. Vendor-header-free, so every backend and
+// the Python facade (lmc_ops) share one definition. Detection (raw layout ->
+// format) lives in lmcache/v1/gpu_connector/kv_format.
+
+/*
+Symbol Reference:
+NL: number of layers
+NB: number of blocks/pages
+BS: block/page size
+NBBS: block/page buffer size = NB * BS
+NH: number of heads
+HS: head size
+TWO: 2
+ONE: 1
+
+_ means a dimension within the same tensor
+_X_ means a dimension across a list
+
+A_X_B_X_C_D_E means:
+kv_cache: List[List[torch.Tensor]]
+len(kv_cache) = A
+len(kv_cache[0]) = B
+kv_cache[0][0].shape = (C, D, E)
+*/
+enum class EngineKVFormat : int {
+  NB_NL_TWO_BS_NH_HS = 0,
+  /*
+  used by:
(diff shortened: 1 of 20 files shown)


csrc/engine_kv_format.h
```diff
@@ -0,0 +1,152 @@
+// SPDX-License-Identifier: Apache-2.0
```
sammshen, line 1: open to suggestions on naming this file

lmcache/v1/gpu_connector/utils.py
```diff
@@ -275,6 +268,8 @@ def normalize_and_discover_per_layer_formats(
                 serving_engine,
                 layout_hints,
             )
+            if not lmc_ops.is_layer_list(fmt):
```
sammshen: this is the single line that fixes the sglang test

ApostaC (approved):
LGTM!

deng451e (approved):

PR #3908, opened 2026-06-27
Collapse object-group KV transfer into one GIL-released native call

YuhanLiu11:
Collapse object-group KV transfer into one GIL-released native call

The per-instance transfer worker threads contend on the GIL because each
staging copy and each kernel launch in an object-group transfer is a
separate native call, and every call releases and re-acquires the GIL.
For a multi-chunk batch this means dozens of GIL handoffs per transfer,
which serializes the worker threads against each other.

This collapses a whole object group's transfer into a single native call.
The Python side becomes a "planner" that resolves every argument to plain
pointers/scalars while holding the GIL, then hands the full plan (staging
copies + kernel launches across all batches) to a new
execute_object_group_transfer native entry point, which issues all of it
on the stream inside one GIL release.

Native side:
- Add lmcache_memcpy_async_batched: issue a list of lazy-allocator copies
  in one GIL-released call instead of one per copy.
- Add execute_object_group_transfer plus the plan types StagingCopy,
  LaunchVar, BatchStep, and KernelGroupSpec (mp_mem_kernels.{cu,cuh}),
  exposed through pybind.

Python side:
- gpu_ops.py: add batched H2D/D2H helpers and the staging-plan builders,
  each gated on hasattr(lmc_ops, ...) so the code still works against an
  older c_ops build (falling back to the per-op path; the GIL win only
  materializes once the extension is rebuilt).
- lmcache_driven_transfer.py: transfer_kv_per_object_group dispatches to
  the single-call fast path when the extension supports it and every
  object is lazy-allocator-backed; GDS / plain-tensor objects and older
  builds keep the per-op path.

No behavioral change to the transferred data; this is a GIL-contention /
throughput optimization for the multiprocess transfer workers.

Signed-off-by: YuhanLiu11 <yliu738@wisc.edu>


lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -52,6 +55,11 @@
 
 logger = init_logger(__name__)
 
+# Whether the native extension can execute a whole object-group transfer plan in
+# one GIL-released call. Older c_ops builds lack it; transfer_kv_per_object_group
+# then falls back to the per-op Python path.
+_CAN_PLAN_OBJECT_GROUP_TRANSFER = hasattr(lmc_ops, "execute_object_group_transfer")
```
ApostaC: We can skip this because this will always be compiled after this PR is in.

lmcache/v1/gpu_connector/gpu_ops.py
```diff
@@ -90,3 +99,259 @@ def lmcache_memcpy_async_d2h(
...
+        True if every non-None object is lazy-allocator-backed.
+    """
+    for memory_obj in memory_objs:
+        if memory_obj is None:
+            continue
+        if isinstance(memory_obj, GDSMemoryObject) or not isinstance(
+            memory_obj.parent(), LazyMemoryAllocator
+        ):
+            return False
+    return True
```
ApostaC: It should work as long as it's not GDS

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -230,6 +238,187 @@ def _recalculate_blocks_to_skip(
...
+    object_group = kv_groups_manager.object_groups[object_group_id]
+    kernel_group_ids = object_group.kernel_group_indices
+    is_h2d = direction == lmc_ops.TransferDirection.H2D
+    max_batch_size = cache_context.max_batch_size
+
+    # Keep every tensor whose data_ptr the native call dereferences referenced
+    # for the duration of this function. (The underlying storage is owned by the
+    # cache context / caller and outlives the async copies; this is belt-and-
+    # suspenders against the temporary view objects being collected early.)
+    keep_alive: list[torch.Tensor] = []
```
ApostaC: Probably we don't need `keep_alive`

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3909, opened 2026-06-27
feat(s3-l2): make LRU eviction survive process restarts (opt-in)

divyvasal:
## Summary

Make the **S3 L2 cache's LRU eviction survive process restarts**. Today the adapter's byte counter and the eviction policy's key index are **in-memory** — they reset to 0 on every restart / pod replacement / spot reclaim. So a fresh process doesn't "see" the objects already in the bucket, can't evict them, and the bucket grows across process generations until the lifecycle TTL ages it out. The configured **LRU cap stops bounding the bucket across restarts**. This adds two opt-in mechanisms to fix that. Both default **off** → no behavior change for existing deployments.

Independent of (and complementary to) the per-`cache_salt` bucket PR (#3892) — different branch, different files-of-concern.

## (a) `s3_seed_usage_on_start` (default `False`)

On startup, scan the bucket (paginated `ListObjectsV2`, progress-logged, non-fatal on error) and **seed both**:
- the adapter usage counter (`_total_bytes_used`), and
- the eviction **policy's** key index,

using S3 `LastModified` as the access-order proxy (oldest-modified = first victim). After seeding, `_check_and_evict_global` sees the true footprint and evicts down to the cap immediately — across any restart.

The load-bearing detail: seeding usage alone is useless; the policy must get the keys too. The path is `_notify_keys_stored` → `L2EvictionPolicy.on_l2_keys_stored` → `policy.on_keys_created` → `LRUEvictionPolicy._order`. `on_keys_created` reverses keys within a call, so seeding is done **one key at a time, oldest-LastModified-first**, which lands the oldest object at the front of `_order` (first eviction victim) — verified against the real `LRUEvictionPolicy`.

## (b) `s3_persist_lru_index` (default `False`, `s3_lru_index_checkpoint_interval_s` default 300)

(a) uses `LastModified` as the order proxy, which mis-ranks a **hot-but-old** prefix (created first, reused thousands of times, never rewritten) → it could be evicted despite being hot. To preserve **true access-order** across restarts:
- A daemon thread periodically serializes the access-order index (key → size + last-access tick) to a compact JSON sidecar (`_lmcache_lru_index.json`); access ticks are only recorded when the flag is on (hot path untouched otherwise).
- On startup, restore the index from the sidecar (true LRU order), reconcile against the live listing (drop vanished keys; append new-since-checkpoint keys by `LastModified` as coldest), and rebuild the tick counter. A missing/partial/corrupt/unknown-version sidecar is treated as absent → falls back to (a)'s scan. `close()` writes a final checkpoint.

## Tests

`tests/v1/distributed/test_s3_l2_adapter_seed.py` — 18 unit tests (stub awscrt/torch; pure logic runs for real, including against the real `LRUEvictionPolicy`): `LastModified` parsing + ordering, size-summing, **policy `_order` population in eviction order**, skip-already-known, sidecar serialize/deserialize roundtrip, corrupt/partial/unknown-version → treated as absent, tick recording gated by the flag.

## Notes / open questions (flagged, not guessed)
- Tick-tracking lives in the **adapter** (it already sees every store/read) rather than reaching into the policy's private `_order`. If you'd prefer the checkpoint to live with the policy, that's a design call.
- On restore, objects present in the bucket but absent from the sidecar are placed as **coldest** (evicted first). If brand-new-since-checkpoint objects should instead be treated as hot, that ordering should flip.
- Seeding calls `_notify_keys_stored` once per key to preserve exact order (one-time startup cost); a batched order-preserving insert on the policy would speed huge-bucket seeds but doesn't exist today.

Draft for now — feedback welcome on the config surface and the two ordering questions above.


divyvasal:
Thanks — addressed both in the latest push:

- **`_seed_policy_in_order` O(N) → O(1)** (high): replaced the per-key notify loop with a **single batch `_notify_keys_stored` of the reversed key list**. Since `on_keys_created` reverses internally, the reversed batch reconstructs the same oldest-first policy order with one lock acquisition + one listener notification instead of N — a 100k-object bucket no longer thrashes on startup. As you noted, tick tracking is **suppressed during the batch** and the `_access_ticks` are then **seeded explicitly in oldest→newest order**, so access-order stays correct. (The existing `test_seed_policy_in_order_populates_policy_lru_order` still passes against the real `LRUEvictionPolicy`.)
- **`_checkpoint_lru_index(timeout=60.0)`** (medium): the PUT-wait bound is now a parameter, and **`close()` passes `timeout=5.0`** so a slow/unresponsive S3 can't block shutdown past the container's termination grace period (→ SIGKILL). The background loop keeps the default.

Also folded in the earlier fix where a missing sidecar (404) raised `RuntimeError` in the CRT callback — now recorded + treated as cleanly absent (falls back to scan).

divyvasal:
**Note on the red `buildkite/unit-tests-amd` check — it's a pre-existing breakage on `dev`, not this PR.**

The failure is in an unrelated ROCm GPU benchmark:

```
tests/benchmarks/test_benchmark.py::test_store_1GB[False-disk]
  AcceleratorError: HIP error: invalid device function (hipErrorInvalidDeviceFunction)
  -> TimeoutError: Store operation haven't finished in 60 seconds
```

With `--maxfail=1`, the suite aborts at this benchmark **before** the `s3_l2_adapter` tests are even collected — so nothing in this PR's diff runs in that lane.

It's repo-wide, not PR-specific:
- `dev` HEAD (`6a57f42b`) is itself `amd=failure`.
- Other open PRs fail the same way (e.g. #3910, #3908).
- Last green AMD commit was `73bf76b3`; the break landed after it (looks tied to the unpinned `torch==2.10.0+rocm7.0` / rocm7.0 wheel the ROCm job installs).

This PR only touches `lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py` (+ its test). The s3-l2 unit tests pass locally (18/18, incl. the LRU-ordering test against the real `LRUEvictionPolicy`), and `DCO` + `k3-unit-tests` are green here. Happy to rebase once `dev`'s AMD lane is fixed. Marking ready for review on logic.

divyvasal:
@ApostaC @maobaolong @chunxiaozheng (and @YaoJiayi @hlin99 for the eviction-policy interaction) — review would be appreciated when convenient.

Opt-in change so the S3 L2 LRU index **survives process restarts**: a startup bucket-scan seeds usage, and a periodic sidecar checkpoint persists the access order (default off → unchanged behavior otherwise). Gemini's notes are addressed (batched O(1) policy seed, bounded checkpoint timeout, `close()` flush). s3-l2 unit tests pass locally (18/18, incl. ordering against the real `LRUEvictionPolicy`); `DCO` + `k3-unit-tests` green. The red `unit-tests-amd` is the pre-existing `dev`-wide ROCm benchmark failure (`test_store_1GB`, `hipErrorInvalidDeviceFunction`), not this PR — `--maxfail=1` aborts before the s3 tests even run.

Independent sibling of #3892.

ApostaC:
@divyvasal Thanks for contributing to LMCache

@aeon-x can you help take a look? Thanks!

aeon-x:
Hey @divyvasal , we have a start up resync feature in LMCache coordinator. It does exactly what you have in this PR in the coordinator level. Can you check it out? 

https://docs.lmcache.ai/mp/coordinator.html

divyvasal:
Thanks @aeon-x — I looked at the coordinator startup-resync. IIUC it backfills the coordinator's usage/eviction trackers from an MP server's `GET /cache/objects`, i.e. from a *server's* in-memory view.

The case this PR targets is when that view is itself gone. On spot-reclaim / pod-replacement the whole L2 server process is replaced, so a freshly-started server reports an **empty** `/cache/objects`, and there's no surviving peer to resync from — the coordinator would resync to empty and the bucket keeps growing past the cap. `s3_seed_usage_on_start` instead scans the **S3 bucket directly** (durable ground truth, `LastModified` as the access-order proxy) so the LRU footprint + cap survive a full server restart; `s3_persist_lru_index` checkpoints the order so a *clean* restart skips the full scan.

It also pairs with #3892 (opt-in per-`cache_salt` bucket isolation): with one physical bucket per tenant, the durable state is spread across N buckets, and a direct per-bucket scan is the natural per-tenant ground truth on restart — something a coordinator resyncing from a replaced server can't reconstruct.

So I see these as complementary to the coordinator resync rather than duplicating it — different source of truth (S3 bucket vs. server view). Happy to add a docs note cross-referencing the two so users pick the right mechanism for their topology. Does that distinction line up with your read?

divyvasal:
Small nudge on this one. CI is green (15/15) and it's opt-in, so no default behavior change. Would really appreciate a review when someone has a few minutes. @ApostaC @sammshen

divyvasal:
Revived and rebased onto current `dev` to clear the stale state. No conflicts; the branch still includes the review-feedback changes.

PR #3910, opened 2026-06-27
ci: pin vLLM nightly to the canary-verified version

maobaolong:
Pin every vLLM install in this repo to the version recorded on the
`buildkite_latest_tested_vllm` branch (file: `latest_tested_vllm.txt`).
That branch is updated by the `:pushpin: Verify & pin vLLM nightly`
canary build whenever it successfully runs `vllm_bench` against a fresh
nightly. With this PR every other build picks up that known-good
version, instead of independently resolving "latest nightly" and
sometimes hitting a freshly broken wheel.

### Changes

- New shared helper `.buildkite/k3_harness/resolve-pinned-vllm.sh`
  resolves `PINNED_VLLM_VERSION` from (in order):
  1. `PINNED_VLLM_VERSION` env var (explicit override)
  2. `latest_tested_vllm.txt` fetched from the
     `buildkite_latest_tested_vllm` branch
  3. empty -> caller falls back to nightly
  Set `USE_PINNED_VLLM=false` to skip the fetch (used by the canary
  build itself, so it can probe a fresher nightly).

- `setup-env.sh` sources the helper and installs
  `vllm[...]==<pin>` when a version is resolved; otherwise it keeps the
  existing "latest nightly" behaviour.

- `docker/example_build.sh` sources the same helper, so local docker
  builds and any CI that runs the script also pick up the pin via
  `--build-arg VLLM_VERSION=...`. The Dockerfiles already supported a
  pinned `VLLM_VERSION` build-arg, so they need no change.

### Verification

Locally I sourced the helper in three modes:

- `PINNED_VLLM_VERSION=0.6.4.dev999+gtest` -> pin honoured.
- `USE_PINNED_VLLM=false` -> empty (nightly fallback).
- No env vars -> fetched real value `0.23.1rc1.dev508+gc6dd32a81`
  from the live branch.

`bash -n` passes on all three changed scripts; `pre-commit run` on
the changed files passes (only `codespell` had files to check).

### Follow-ups not in this PR

- The canary pipeline that writes `latest_tested_vllm.txt` lives on a
  separate branch and will be PR'd separately.

### DIFF
--- a/docker/example_build.sh
+++ b/docker/example_build.sh
@@ -3,7 +3,16 @@
 # Update the following variables accordingly
 CUDA_VERSION=13.0
 DOCKERFILE_NAME='Dockerfile'
-VLLM_VERSION="nightly"
+# VLLM_VERSION resolution order:
+#   1. Pre-set VLLM_VERSION env var (overrides everything).
+#   2. PINNED_VLLM_VERSION resolved from the
+#      `buildkite_latest_tested_vllm` branch (the most recent vLLM nightly
+#      verified by the canary build). Empty when offline / first run.
+#   3. Fallback to "nightly" so behaviour matches the previous default.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
+# shellcheck source=../.buildkite/k3_harness/resolve-pinned-vllm.sh
+source "${SCRIPT_DIR}/../.buildkite/k3_harness/resolve-pinned-vllm.sh"
+VLLM_VERSION="${VLLM_VERSION:-${PINNED_VLLM_VERSION:-nightly}}"
 DOCKER_BUILD_PATH='../' # This path should point to the LMCache root for access to 'requirements' directory
 UBUNTU_VERSION=24.04
 
--- a/docker/example_build.sh
+++ b/docker/example_build.sh
@@ -4,16 +4,10 @@
 # Update the following variables accordingly
 CUDA_VERSION=13.0
 DOCKERFILE_NAME='Dockerfile'
-# VLLM_VERSION resolution order:
-#   1. Pre-set VLLM_VERSION env var (overrides everything).
-#   2. PINNED_VLLM_VERSION resolved from the
-#      `buildkite_latest_tested_vllm` branch (the most recent vLLM nightly
-#      verified by the canary build). Empty when offline / first run.
-#   3. Fallback to "nightly" so behaviour matches the previous default.
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
-# shellcheck source=../.buildkite/k3_harness/resolve-pinned-vllm.sh
-source "${SCRIPT_DIR}/../.buildkite/k3_harness/resolve-pinned-vllm.sh"
-VLLM_VERSION="${VLLM_VERSION:-${PINNED_VLLM_VERSION:-nightly}}"
+# Set VLLM_VERSION to a specific version before running this script,
+# e.g.: VLLM_VERSION=0.9.1 ./example_build.sh
+# Defaults to "nightly" if not set.
+VLLM_VERSION="${VLLM_VERSION:-nightly}"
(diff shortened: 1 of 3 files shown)


maobaolong:
Both ai suggestions above applied in 33e3540d:

- swapped `sed | head` for a single `awk` so we don't trip SIGPIPE under `pipefail` and avoid the GNU-only `\\+` regex.
- added `#!/usr/bin/env bash` to `docker/example_build.sh` so direct execution doesn't fall back to `/bin/sh`.

maobaolong:
Hit a real-world failure on https://buildkite.com/lmcache/k3-comprehensive-test/builds/3187 right after merging the canary write side:

```
no version of vllm[runai]==0.23.1rc1.dev508+gc6dd32a81
```

Root cause: `wheels.vllm.ai/nightly/cu130/` is a rolling index that only keeps the latest wheel — `dev508` was already replaced by `dev516` a few hours later, so the pin became unsatisfiable.

Fix in d5cda948: vLLM also serves every nightly at `wheels.vllm.ai/<full-sha>/cu130/` (a permanent PEP 503 simple index). The PEP 440 local version (`+g<short-sha>`) gives us the short SHA; we expand it via the public GitHub commits API and add the archived URL as an `--extra-index-url`. `GITHUB_TOKEN` is honoured (5000 req/h) but optional, with 3x retry and a warning-only fallback so a transient API hiccup never breaks an unrelated job.

Verified locally: empty-array expansion is safe under `set -u`, short-SHA extraction handles both `dev508+gc6dd32a81` and unpinned versions, and the archived URL serves the wheel that the rolling index has rolled off.

maobaolong:
Follow-up in 4b55f1e0: now that the canary writes `full_sha` and `archive_index_url` directly into `latest_tested_vllm.txt`, parse them in `resolve-pinned-vllm.sh` and consume the archive URL in `setup-env.sh` directly. The common path no longer touches the GitHub commits API at all. The live-API path stays as a fallback for old-format pin files and for cases where the canary couldn't resolve the full SHA that run.

Verified the pin file already serves the new format:
```
$ curl -s https://raw.githubusercontent.com/LMCache/LMCache/buildkite_latest_tested_vllm/latest_tested_vllm.txt
0.23.1rc1.dev519+g091d13976
short_sha=091d13976
full_sha=091d13976c1c246714bb2112dd2e208561dda6a3
archive_index_url=https://wheels.vllm.ai/091d13976c1c246714bb2112dd2e208561dda6a3/cu130
```

The first line is still the bare version, so any consumer that just `head -n1`'s the file keeps working.

docker/example_build.sh
```diff
@@ -1,9 +1,19 @@
...
-VLLM_VERSION="nightly"
+# VLLM_VERSION resolution order:
+#   1. Pre-set VLLM_VERSION env var (overrides everything).
+#   2. PINNED_VLLM_VERSION resolved from the
+#      `buildkite_latest_tested_vllm` branch (the most recent vLLM nightly
+#      verified by the canary build). Empty when offline / first run.
+#   3. Fallback to "nightly" so behaviour matches the previous default.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" && pwd)"
+# shellcheck source=../.buildkite/k3_harness/resolve-pinned-vllm.sh
+source "${SCRIPT_DIR}/../.buildkite/k3_harness/resolve-pinned-vllm.sh"
```
ApostaC: This is the example script for people to build the Docker image, we should not couple it with our CI/CD.

The vLLM version should be resolved before calling this script.

ApostaC (commented):
Otherwise LGTM

docker/example_build.sh
maobaolong: Good point. Removed the source of the CI resolve script from example_build.sh — it now just reads VLLM_VERSION from the env (defaulting to nightly), keeping it self-contained for end users.

sammshen: agree

sammshen (approved):
LGTM!

ApostaC (approved):
LGTM!

PR #3914, opened 2026-06-27
docs(coding-standards): add log format specifier convention

April-Sonnet:
Adds a new subsection (7.7) to `docs/coding_standards.md` documenting the printf-style format specifier convention for logging calls:

| Type | Specifier |
|------|-----------|
| `int` (counts, numeric IDs) | `%d` |
| `str` (string IDs, names) | `%s` |
| `bool` | `%s` |
| `float` | `%s` or `%r` |
| Objects / exceptions | `%s` |

Also clarifies that when `instance_id` is used as an `int` (GPU/worker index in the multiprocess layer), prefer more descriptive names like `gpu_instance_id` or `worker_id` to avoid confusion with the `str` `instance_id` in the engine layer.

Refs #3785

### DIFF
--- a/docs/coding_standards.md
+++ b/docs/coding_standards.md
@@ -374,3 +374,19 @@ Use this checklist before submitting a PR or during review:
 - [ ] Thread safety maintained for shared state
 - [ ] No unnecessary memory copies in hot paths
 - [ ] CUDA/GPU resources properly managed
+
+### 7.7 Log Format Specifiers
+
+Use the correct `printf`-style format specifier based on the **type annotation**, not neighboring log calls:
+
+| Type | Specifier | Example |
+|------|-----------|---------|
+| `int` (counts, numeric IDs) | `%d` | `logger.info("Retrieved %d chunks", chunk_count)` |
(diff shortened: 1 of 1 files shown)


ApostaC (approved):
LGTM!

docs/coding_standards.md
```diff
@@ -374,3 +374,19 @@ Use this checklist before submitting a PR or during review:
 - [ ] Thread safety maintained for shared state
 - [ ] No unnecessary memory copies in hot paths
 - [ ] CUDA/GPU resources properly managed
+
+### 7.7 Log Format Specifiers
```
ApostaC: Please either fix the section number or put the new section into the correct place. Thanks!

ApostaC (changes requested):
Otherwise LGTM!

April-Sonnet:
Updated: moved the 7.7 section before the `---` divider (now properly between 7.6 Resource Management and Section 8). The content is unchanged. @ApostaC please take another look when you have time.

PR #3916, opened 2026-06-27
[Fix][FS Native] Avoid O_DIRECT read-ahead splits

3xdevv:

`read_ahead_size` is a buffered-I/O optimization and should not be applied when the native filesystem connector is using `O_DIRECT`. Splitting an O_DIRECT read can create unaligned direct I/O requests and defeat the point of direct I/O benchmarking.

- Skips read-ahead splitting whenever an FS native read actually uses `O_DIRECT`.
- Centralizes O_DIRECT enablement checks for compile-time availability, transfer length alignment, and buffer address alignment.
- Keeps the existing buffered fallback for transfer lengths that are not filesystem-block aligned.
- Reports a clear runtime error for misaligned O_DIRECT buffers instead of silently measuring buffered I/O.
- Updates FS native docs for the O_DIRECT buffer-alignment behavior.
- Adds native FS connector tests for O_DIRECT with read-ahead configured and for misaligned O_DIRECT buffers.

- The O_DIRECT/read-ahead test skips when the native FS extension, `O_DIRECT`, or filesystem support is unavailable.
- Tested by:
```bash
pytest -xvs tests/v1/storage_backend/test_fs_native_connector.py
```

**If applicable**:


ApostaC (approved):

sammshen (approved):
LGTM！

3xdevv:
Hello, @ApostaC @sammshen. Could you please take a look at this PR when you have a chance?

PR #3918, opened 2026-06-27
[BugFix] lmcache-driven (TRT-LLM MP): trim trailing partial chunk before downsample

### LINKED ISSUE #3917: [Bug] lmcache-mp (TRT-LLM): store/retrieve crash on non-chunk-aligned block lists (save_unfull_chunk=False)
### Summary
With the TensorRT-LLM `lmcache-mp` connector and `save_unfull_chunk=False` (the default), any request whose prompt length isn't an exact multiple of `chunk_size` crashes the store (and retrieve), failing the request:

```
AssertionError: len(block_ids[0]) should be a multiple of total_blocks_per_chunk (8), but got 68
  File ".../lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py", line 186, in downsample_and_stage_block_ids
    assert len(old_block_ids) % total_blocks_per_chunk == 0, (
```

### Environment
- LMCache 0.5.0
- TensorRT-LLM 1.3.0rc16, connector `lmcache-mp`
- gpt-oss-120b: `tokens_per_block=32`, `chunk_size=256` → 8 blocks per chunk

### Root cause
`downsample_and_stage_block_ids` (added in #3612 for sliding-window / multi-object-group support) asserts each group's block list is an exact multiple of `total_blocks_per_chunk`. But the engine hands it a block list covering *all* prompt tokens, so it can include a trailing **partial** chunk when the prompt length isn't a multiple of `chunk_size` (e.g. 8 full chunks = 64 blocks + a 4-block tail = 68 blocks, not a multiple of 8).

`num_chunks` (derived from the keys, which already honor `save_unfull_chunk=False`) counts only the whole chunks, and the caller's own underflow check already guarantees `len(block_ids) >= num_chunks * blocks_per_chunk`. Nothing trims the upper-bound partial before the assert, so any non-aligned prompt deterministically crashes.

### Reproduce
TRT-LLM + `connector: lmcache-mp`, `chunk_size=256`, `save_unfull_chunk=False` (default); send any prompt longer than one chunk whose length isn't a multiple of `chunk_size`. A single such request reliably reproduces the assert. (Short or chunk-aligned prompts, and pure cache hits, don't trigger it — likely why the E2E tests pass.)

I'll open a PR with a fix.

nafis271:
Fixes #3917.

`downsample_and_stage_block_ids` asserts each group's block list is an exact multiple of `total_blocks_per_chunk`, but the engine passes a block list covering all prompt tokens, which can carry a trailing partial chunk when the prompt length isn't a multiple of `chunk_size`. With `save_unfull_chunk=False` (the default) this crashes every such store and retrieve (assert + repro in the issue).

This passes `num_chunks` (the chunk count from the keys) into `downsample_and_stage_block_ids` and trims each group's block list to `num_chunks * total_blocks_per_chunk` before downsampling. The callers' existing underflow check already guarantees `len >= num_chunks * blocks_per_chunk`, so this drops exactly the trailing partial chunk and keeps the whole chunks the keys cover. The strict `% == 0` assert (trivially true after the trim) is replaced with an explicit `>=` precondition that raises `ValueError`, so an underflow fails loudly (and isn't compiled out under `-O`) rather than silently truncating short.

Verified on TRT-LLM 1.3 + LMCache 0.5.0 (gpt-oss-120b, `save_unfull_chunk=False`): non-aligned prompts that previously crashed now store and retrieve as cache hits (e.g. a 2029-token prompt stores 7 whole chunks / 1792 tokens, the partial dropped); ~730 stores under mirror traffic with zero asserts; chunk-aligned stores are unchanged.

### DIFF
--- a/lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
+++ b/lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
@@ -126,6 +126,7 @@ def batched_iteration_with_skip(
 def downsample_and_stage_block_ids(
     cache_context: BaseCacheContext,
     block_ids: list[list[int]],
+    num_chunks: int,
 ) -> list[torch.Tensor]:
     """Cut the block id lists to skip the unneeded blocks in a chunk and
     stage it into GPU tensors for later use.
@@ -138,6 +139,9 @@ def downsample_and_stage_block_ids(
     Args:
         cache_context: The cache context containing the KV cache information.
         block_ids: The original block id lists, indexed by LMCache KV group index.
+        num_chunks: Number of whole chunks the keys cover. Block ids beyond
+            ``num_chunks`` whole chunks (a trailing partial chunk) are trimmed
+            off before downsampling.
 
     Returns:
         The cut block id lists, indexed by LMCache KV group index.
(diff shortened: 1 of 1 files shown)


nafis271:
Thanks for the review — addressed both: documented `num_chunks` in the docstring, and switched the precondition from an `assert` to an explicit `ValueError` so it isn't compiled out under `-O`.

ApostaC:
Hi @nafis271, thanks for the contribution!

The flag `save_unfull_chunk` was used in non-MP mode, and it's no longer used in MP mode.

In MP mode, we only save the full chunks. If TRTLLM crashes, I think we should debug in the TRTLLM integration module (under `lmcache/integration`) instead of changing the core module behavior.

PR #3920, opened 2026-06-28
[CI] Supports timer trigger test vllm and record

maobaolong:

The https://github.com/LMCache/LMCache/pull/3910 will use the record file in the https://github.com/LMCache/LMCache/blob/buildkite_latest_tested_vllm/latest_tested_vllm.txt

https://buildkite.com/lmcache/k3-multiprocess-test/builds/3163

Clipboard_Screenshot_1782607974
 

**If applicable**:

### DIFF
--- a/.buildkite/k3_tests/multiprocess/pipeline.yml
+++ b/.buildkite/k3_tests/multiprocess/pipeline.yml
@@ -16,6 +16,7 @@ x-flaky-retry: &flaky-retry
 
 steps:
   - group: ":compression: Multiprocess (2-GPU)"
+    if: build.env("TEST_ONLY") !~ /true/
     steps:
       - label: ":compression: vllm_bench"
         command: .buildkite/k3_tests/multiprocess/run.sh vllm_bench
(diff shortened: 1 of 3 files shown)


maobaolong:
Done, switched to jq. Also used `.sha // empty` so it outputs nothing instead of `null` when the field is missing.

ApostaC (commented):
LGTM in general. Should we merge #3910 first?

ApostaC (approved):
LGTM in general.

.buildkite/k3_tests/multiprocess/pipeline.yml
```diff
@@ -16,6 +16,7 @@ x-flaky-retry: &flaky-retry
 
 steps:
   - group: ":compression: Multiprocess (2-GPU)"
+    if: build.env("VERIFY_AND_PIN_VLLM") !~ /true/
```
sammshen, line 19: nice!

sammshen (approved):
LGTM!

PR #3924, opened 2026-06-28
[platform][refactor] refactor device detection

chunxiaozheng:
1. move device detection related code from `lmcache/__init__.py` to `platform/device_detection.py`
2. add auto discover for device


chunxiaozheng:
@hlin99 @maobaolong could you help take a look? Besides, I have a little question, if we add the auto discover, the order of detecting devices should be in alphabetical order, not in the previously hard coded order, is this reasonable?

chunxiaozheng:
> @hlin99 @maobaolong could you help take a look? Besides, I have a little question, if we add the auto discover, the order of detecting devices should be in alphabetical order, not in the previously hard coded order, is this reasonable?

Sorry, the cpu tests failed, and i will fix later.

lmcache/v1/platform/cuda/__init__.py
```diff
@@ -26,3 +27,33 @@ def _cuda_is_available() -> bool:
 
 register_availability("cuda", _cuda_is_available)
```
maobaolong: Would you be willing to remove the explicit register in this PR as well?

lmcache/v1/platform/cuda/__init__.py
```diff
@@ -26,3 +27,33 @@ def _cuda_is_available() -> bool:
...
 register_pin_memory_backend("cuda", CudaPinMemoryBackend)
```
maobaolong: Ditto

lmcache/v1/platform/musa/__init__.py
```diff
@@ -1,2 +1,39 @@
...
+            # Third Party
+            import torch
+
+            return hasattr(torch, "musa") and torch.musa.is_available()  # type: ignore[attr-defined]
+        except Exception:
+            return False
+
+    @property
+    def priority(self) -> int:
+        return 0
```
hlin99: hi @chunxiaozheng what does this priority 0,1,2 mean? can we have a enum definition to make it more clear to understand?

hlin99 (approved):
LGTM
if we can have a more readable priority, it would be better.

lmcache/v1/platform/musa/__init__.py
chunxiaozheng: @hlin99 thanks for your review, after discussed offline with @maobaolong , perhaps there is no need to set priority here, maintain a fair order through alphabetical order, or we can set environment variables(such as `export DEVICE_TYPE=musa`) to enforce a device.

lmcache/v1/platform/cuda/__init__.py
chunxiaozheng: done

lmcache/v1/platform/cuda/__init__.py
chunxiaozheng: done

maobaolong (approved):
As discussed offline, this LGTM.

PR #3926, opened 2026-06-28
[Core] Add TTLLock batch operations for L1 read locks

THINKER-ONLY:

This PR optimizes read-lock accounting in the distributed L1 manager by adding native count-based operations to `TTLLock`. The current L1 manager already supports reserving and releasing more than one read count for a key through `extra_count`, but it implements that behavior with Python loops that repeatedly call the C++ `TTLLock.lock()` / `TTLLock.unlock()` methods. That means each key can cross the Python/C++ boundary `1 + extra_count` times even though the underlying operation is just atomic counter arithmetic.

The new `TTLLock.lock_count(count)` and `TTLLock.unlock_count(count)` APIs move that repeated work into native code. Existing `lock()` and `unlock()` remain available and now delegate to the count-based versions with `count=1`, so existing callers keep the same behavior. The L1 manager uses the new batch operations in `reserve_read`, `finish_read`, and `finish_write_and_reserve_read`, reducing repeated native calls while preserving the same logical read-lock counts and state transitions.

The native implementation keeps the existing TTL semantics intact. If a lock has expired, `lock_count(count)` resets the stale counter to the requested count and refreshes the expiration time. If the lock is still active, it atomically increments the current counter and refreshes the TTL. `unlock_count(count)` preserves the old unlock behavior by clamping the counter at zero instead of allowing underflow. The implementation also validates non-positive counts and rejects counter overflow explicitly, which makes misuse fail clearly rather than corrupting the counter.

Reviewer map:

| Area | Key files | What to review |
| --- | --- | --- |
| Native API | `csrc/storage_manager/ttl_lock.{h,cpp}` | Count validation, overflow handling, expired-TTL reset semantics, and thread-safety around the temporary reset sentinel. |
| Python exposure | `csrc/storage_manager/pybind.cpp`, `lmcache/native_storage_ops.pyi` | Python binding names, argument handling, and documented exception behavior. |
| L1 manager hot path | `lmcache/v1/distributed/l1_manager.py` | Replacement of Python loops with one `lock_count()` / `unlock_count()` call per successful key. |
| Tests | `tests/v1/native_storage_ops/test_ttl_lock.py`, `tests/v1/distributed/test_l1_manager.py` | Coverage for native semantics, invalid inputs, overflow, TTL expiry, concurrency, and L1 state transitions. |

The practical contribution is narrow but important: the PR removes avoidable interpreter overhead from a distributed-cache hot path without changing the public L1 manager API. The old per-count behavior is still represented exactly by the new count-based operations, and the tests assert both the native lock semantics and that the L1 manager actually uses the new batch methods.

Optimization evidence:

I measured the exact operation this PR optimizes: acquiring and releasing multiple read-lock counts from Python. The old path performs `count` Python calls to `lock()` followed by `count` Python calls to `unlock()`. The new path performs one `lock_count(count)` call followed by one `unlock_count(count)` call. The benchmark uses `TTLLock` directly, runs 20,000 iterations per repeat, reports the median of 7 repeats, and measures one acquire+release cycle in microseconds.

| Read-lock count | Old Python loop median | New batch API median | Speedup |
| --- | ---: | ---: | ---: |
| 8 | 3.116 us | 0.392 us | 7.96x |
| 32 | 11.913 us | 0.388 us | 30.72x |
| 128 | 46.706 us | 0.388 us | 120.37x |

The speedup scales with `count` because the old path pays Python/C++ call overhead once per count, while the new path pays it once per key. This benchmark intentionally isolates lock accounting overhead; it does not claim an end-to-end distributed-cache throughput number.

- This is a performance PR, not an API redesign. The only new public surface is the optional `TTLLock.lock_count()` / `TTLLock.unlock_count()` native binding.
- The old single-count methods remain supported. `lock()` is equivalent to `lock_count(1)`, and `unlock()` is equivalent to `unlock_count(1)`.
- `unlock_count()` intentionally clamps at zero, matching the previous repeated-`unlock()` behavior.
- `lock_count()` raises on invalid count and counter overflow. That behavior is documented in the C++ header and Python stub.

Validation performed:

- `pytest -q tests/v1/native_storage_ops/test_ttl_lock.py tests/v1/distributed/test_l1_manager.py`
  - Result: `108 passed`
  - This covers native `TTLLock` behavior and the distributed L1 manager paths changed by this PR.
  - Passed: SPDX, direct torch device check, isort, ruff, ruff-format, codespell, clang-format, mypy.

**If applicable**:

### DIFF
--- a/csrc/storage_manager/pybind.cpp
+++ b/csrc/storage_manager/pybind.cpp
@@ -37,8 +37,13 @@ PYBIND11_MODULE(native_storage_ops, m) {
       .def("lock", &TTLLock::lock,
            "Increment the lock counter by 1 and update the TTL. "
            "If the previous TTL has expired, reset counter to 1.")
+      .def("lock_count", &TTLLock::lock_count, py::arg("count"),
+           "Increment the lock counter by count and update the TTL. "
+           "If the previous TTL has expired, reset counter to count.")
       .def("unlock", &TTLLock::unlock,
            "Decrement the lock counter by 1 (minimum 0).")
+      .def("unlock_count", &TTLLock::unlock_count, py::arg("count"),
+           "Decrement the lock counter by count (minimum 0).")
       .def("is_locked", &TTLLock::is_locked,
            "Check if the lock is held (counter > 0 and TTL not expired).")
       .def("reset", &TTLLock::reset,
--- a/csrc/storage_manager/ttl_lock.cpp
+++ b/csrc/storage_manager/ttl_lock.cpp
@@ -2,6 +2,10 @@
 
 #include "ttl_lock.h"
 
+#include <limits>
+#include <stdexcept>
+#include <thread>
+
 namespace lmcache {
 namespace storage_manager {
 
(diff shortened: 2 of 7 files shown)


ApostaC (approved):
Thanks for contributing to LMCache! LGTM

sammshen (approved):
LGTM！

PR #3927, opened 2026-06-28
[Core] Batch store memory allocation in LMCacheEngine

THINKER-ONLY:

This PR optimizes the `LMCacheEngine.store()` hot path by batching memory allocation for consecutive chunks that share the same KV shape and dtype. Before this change, `store()` asked the storage manager to allocate one `MemoryObj` per processed chunk. That was simple, but it meant a normal prefill/store request with many full-sized chunks repeatedly paid the allocator call overhead even when every chunk had the same allocation shape.

The new flow keeps a small pending-chunk buffer inside `store()`. As tokens are processed into cache chunks, the engine groups adjacent chunks whose `metadata.get_shapes(num_tokens)` and `metadata.get_dtypes()` match. When the group ends, or when token processing finishes, the engine flushes the group through `StorageManager.batched_allocate()` with the group size. The rest of the store pipeline remains the same: successfully allocated memory objects are paired with their original `(start, end, key)` metadata, copied from GPU through `batched_from_gpu()`, and submitted to storage through `batched_put()`.

This keeps the optimization focused on allocation. It does not change how tokens are segmented, how keys are generated, how KV events are emitted, how GPU extraction works, or how storage backends receive the final objects. The order of `starts`, `ends`, `keys`, and `memory_objs` is still derived from the token-processing order, and differently shaped chunks are flushed separately so tail chunks and `save_unfull_chunk` behavior stay isolated from the full-chunk fast path.

The fallback behavior is also preserved. If `batched_allocate()` cannot allocate the group and returns `None`, the helper falls back to the previous one-by-one `allocate()` behavior. That matters because the existing single-allocation path owns eviction behavior and `force_store_wait` / `busy_loop` behavior. If a backend returns a partial list with a `None` inside it, the store path keeps only the successful prefix, releases any later allocated objects, and stops at the first failed slot. That matches the old partial-store semantics while avoiding unused `MemoryObj` leaks.

Reviewer map:

| Area | Key files | What to review |
| --- | --- | --- |
| Store batching | `lmcache/v1/cache_engine.py` | Pending chunk grouping, shape/dtype boundary flushes, and preservation of token-processing order. |
| Allocation helpers | `lmcache/v1/cache_engine.py` | Batch fast path, one-by-one fallback, partial-list cleanup, and `force_store_wait` propagation. |
| Regression tests | `tests/v1/test_cache_engine.py` | Public `store()` tests for uniform batches, batch failure fallback, partial batch cleanup, and differently sized tail chunks. |

The main contribution is reducing allocator overhead for the common case without making the store pipeline less predictable. For a run of `N` compatible chunks, the allocation path changes from `N` individual `allocate()` calls to one `batched_allocate(..., batch_size=N)` call. If the fast path is not available or cannot allocate, the code returns to the existing per-chunk behavior.

Optimization evidence:

I measured the store allocation path with a lightweight public-`store()` benchmark using a mocked storage manager and mocked GPU connector. This isolates the Python store-path overhead and allocator call pattern without mixing in real GPU transfer, disk I/O, or backend-specific allocator latency. The benchmark stores 256 tokens as 64 chunks with identical shape and dtype, runs 9 repeats after warmup, and reports the median store-call time.

The `origin/dev` baseline performs one `allocate()` call per chunk. This PR performs one `batched_allocate()` call for the whole compatible chunk group.

| Branch / path | Median `store()` time | `batched_allocate()` calls | `allocate()` calls |
| --- | ---: | ---: | ---: |
| `origin/dev` one-by-one allocation | 2246.809 us | 0 | 64 |
| This PR batch fast path | 1614.174 us | 1 | 0 |

In this isolated benchmark, the batched path reduces allocator calls from 64 to 1 and lowers median mocked `store()` time by about 28.2% (`1.39x` faster). This is not an end-to-end serving benchmark; it is intended to prove that the PR removes repeated allocation-call overhead on the exact hot path it changes.

- This PR does not change the public `LMCacheEngine.store()` API.
- The optimization is intentionally scoped to allocation. GPU copy, storage put, token database processing, and KV event construction are not redesigned.
- The batch fast path uses `eviction=False` for the initial attempt. If direct batch allocation fails, the fallback path calls the existing one-by-one `allocate()` method, preserving the previous eviction behavior.
- The implementation deliberately flushes before a shape/dtype change. This prevents full chunks and tail chunks from being allocated as if they had the same layout.

Validation performed:

- New focused store tests:
  - `test_store_uses_batched_allocate_for_uniform_chunks`
  - `test_store_falls_back_to_single_allocate_when_batched_allocate_fails`
  - `test_store_keeps_successful_prefix_from_partial_batched_allocate`
  - `test_store_batches_only_chunks_with_matching_shapes`
  - Result: `4 passed`
- Store hot-path regression subset:
  - Existing `test_paged_same_retrieve_store` nodes.
  - CPU and remote `test_paged_store_offset` nodes.
  - Existing `test_paged_store_kv_tensors_mask` nodes.
  - The four new batch allocation tests.
  - Result: `12 passed`
- Local-disk offset verification:
  - `test_paged_store_offset[cpu-False-local_disk-256]`
  - `test_paged_store_offset[cpu-True-local_disk-256]`
  - Result: `2 passed` 
  - Passed: SPDX, direct torch device check, isort, ruff, ruff-format, codespell, clang-format, mypy.

**If applicable**:

### DIFF
--- a/lmcache/v1/cache_engine.py
+++ b/lmcache/v1/cache_engine.py
@@ -477,8 +477,77 @@ class LMCacheEngine:
         if request_configs is not None and len(request_configs) != 0:
             assert isinstance(request_configs, dict)
 
+        force_store_wait = self.config.get_extra_config_value("force_store_wait", False)
+
         with store_stats.profile_process_tokens():
             prev_key = 0
+            pending_chunks: List[Tuple[int, int, CacheEngineKey]] = []
+            pending_shapes: Optional[list[torch.Size]] = None
+            pending_dtypes: Optional[list[torch.dtype]] = None
+
+            def flush_pending_chunks() -> bool:
+                nonlocal prev_key, pending_shapes, pending_dtypes
+                nonlocal tot_kv_size, tot_token_num
+
+                if not pending_chunks:
+                    return False
+
+                assert pending_shapes is not None
+                assert pending_dtypes is not None
+                allocated_objs, allocation_stopped = self._allocate_store_memory_objs(
+                    pending_shapes,
+                    pending_dtypes,
+                    len(pending_chunks),
+                    busy_loop=force_store_wait,
+                )
+
+                for (start, end, key), memory_obj in zip(
+                    pending_chunks,
+                    allocated_objs,
+                    strict=False,
+                ):
+                    starts.append(start)
+                    ends.append(end)
+                    keys.append(key)
+                    memory_objs.append(memory_obj)
+                    tot_kv_size += memory_obj.get_size()
+                    tot_token_num += end - start
+
+                    # Create KV event
+                    if self.kv_events_enabled:
+                        stored_event = CacheStoreEvent(
+                            block_hashes=[key.chunk_hash],
+                            parent_block_hash=None if start == 0 else prev_key,
+                            token_ids=[],
+                            block_size=end - start,
+                            lora_id=None,
+                            medium="cpu",
+                            lora_name=None,
+                        )
+                        if tokens is not None:
+                            stored_event.token_ids = convert_tokens_to_list(
+                                tokens,
+                                start,
+                                end,
+                            )
+                            if isinstance(tokens, torch.Tensor):
+                                stored_event.medium = tokens.device
+                        elif hashes is not None:
+                            stored_event.token_ids = hashes[start : end + 1]
+                        logger.debug(
+                            (
+                                "Added kv cache event '%s' to kv cache events queue"
+                                % stored_event
+                            )
+                        )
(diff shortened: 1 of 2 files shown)


ApostaC (commented):
Hi @THINKER-ONLY , thanks for your contribution!

Your code changes land into non-MP mode. However, we are gradually deprecating the non-MP mode in Q3-Q4 this year. We recommend you to take a look at the MP mode (see https://docs.lmcache.ai/getting_started/quickstart.html)

PR #3930, opened 2026-06-29
[feat][cb] enable cb compartible with vllm hma

deng451e:
enable blend server with  vLLM's hybrid KV-cache manager (HMA) on multi-group SWA models. Fixed via three changes, all gated on `engine_type == "blend"` (standard prefix-caching unaffected):

1. **Per-group scatter** — `cb_retrieve_pre_computed` takes per-group block tables (`list[list[int]]`) and builds slot_mapping per group.
2. **Single object group** — force `separate_object_groups=False`. CB's scatter is single-group; the default split leaves group 1 unfilled.
3. **Full per-chunk SWA KV** — enable `full_sw_kv`. Sub-chunk window-cut caches are valid only at a chunk's original position, but CacheBlend reuses chunks at arbitrary positions. Threaded into `GPUCacheContext` before buffer/group sizing via `enable_full_sw_kv()` (overrides `get_subchunk_sw_size_tokens`, `get_attn_desc`).

- All three key off `engine_type == "blend"`; standard path keeps SWA windowing and object-group separation, so store/transfer volume and prefix-caching semantics are unchanged.
 

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/engine_context.py
+++ b/lmcache/v1/multiprocess/engine_context.py
@@ -180,9 +180,11 @@ class MPCacheServerContext:
         chunk_size: int = 256,
         hash_algorithm: str = "blake3",
         separate_object_groups: bool = True,
+        full_sw_kv: bool = False,
     ) -> None:
         self._chunk_size = chunk_size
         self._separate_object_groups = separate_object_groups
+        self._full_sw_kv = full_sw_kv
 
         # Initialize the process-global GDS context.
         # No-op when GDS L1 is disabled (config is None).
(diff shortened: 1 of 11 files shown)


lmcache/v1/multiprocess/engine_context.py
```diff
@@ -180,9 +180,11 @@ def __init__(
         chunk_size: int = 256,
         hash_algorithm: str = "blake3",
         separate_object_groups: bool = True,
+        full_sw_kv: bool = False,
```
ApostaC, line 183: what's the use of `full_sw_kv`?

ApostaC (commented):
One quick question

lmcache/v1/multiprocess/engine_context.py
deng451e: Sliding window groups can store only the chunk's last sw_size KV, which can be invalidated for non-prefix cache hits. full_sw_kv disables the window cut and stores the full chunk KV. Only enabled for engine_type == "blend".

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3932, opened 2026-06-29
feat(s3-l2): keep S3 KV eviction ahead of writes — hysteresis, bulk deletes, best-effort writes, coalescing, metrics

divyvasal:
## Problem

Under sustained writes at the S3 L2 cap, eviction can't keep up and the tier deadlocks: writes
outrun per-key `DeleteObject` eviction → L2 pins at the cap → the store path backpressures vLLM →
`running` drops to 0 and aggregate throughput goes to **0** (a hard stall, not graceful degradation).

## Changes (all opt-in — defaults preserve current behavior)

- **Eviction hysteresis** — a triggered sweep evicts down to a `target_watermark` (low watermark) in
  one pass instead of a fixed `eviction_ratio`, so headroom always exists and eviction can't fall
  behind a write burst. Applied to both the mp-coordinator eviction manager and the distributed L2
  eviction controller. *Default `target_watermark=0.0` → old fixed-ratio behavior.*
- **Bulk `DeleteObjects`** — replace the per-key delete loop with batched S3 `DeleteObjects`
  (≤1000 keys/call) so the evictor's delete rate matches the write rate.
- **Best-effort writes** — when `usage_fraction >= store_reject_watermark` the adapter drops the
  store (cache-miss + recompute) instead of blocking, hard-bounding the overshoot at the source.
  *Gated on `store_reject_watermark > 0.0` → off by default.*
- **Write coalescing** — single-flight on `chunk_hash`, so duplicate concurrent stores of the same
  prefix collapse to one upload.
- **Metrics** — `eviction_lag` + in-flight / rejected / coalesced store counters for observability.

## Validation (live, 8×H200, gpt-oss-120b + lmcache MP + S3 KV tier)

Under a **13× over-cap overwhelm**, the box degraded **gracefully** — kept serving (`running`≈155,
~159K tok/s, never 0) and drained back to the target watermark when load dropped; previously this
regime stalled to throughput 0. Eviction fired with a dynamic ratio that scaled with overshoot
(`evicting ratio=0.13 (target=0.70)` at usage 0.80 → `0.31` at usage 1.02). Best-effort writes
hard-bounded usage to ~1.2× cap (bounded overshoot) instead of unbounded growth.

## Tests

- `tests/v1/distributed/test_s3_l2_adapter.py` — bulk-delete request build/parse, reject-at-watermark,
  coalesce-duplicate-keys.
- `tests/v1/mp_coordinator/test_eviction_manager.py` — hysteresis eviction amounts.

## Notes

Draft: branch is a few commits behind `dev` — will rebase before marking ready-for-review; opening
as draft to run CI + gather early feedback on the approach.


divyvasal:
Thanks for the review @gemini-code-assist — addressed all three in 5aca71d0:

- **HIGH (in-flight leak):** `_execute_store` now wraps its body in `try/except/finally` so the A4 in-flight-key release + store-efd notify always run — including on `CancelledError` (shutdown/timeout) and unhandled exceptions. Keys can no longer be coalesce-skipped forever after a failed/cancelled store.
- **MEDIUM (config validation):** added `EvictionConfig.__post_init__` validating `trigger_watermark ∈ (0,1]`, `target_watermark` = 0.0 or `∈ (0, trigger_watermark)`, `eviction_ratio ∈ [0,1]`.
- **MEDIUM (type hint):** added the return type to `_delete_many_request`.

divyvasal:
Rebuilt on current dev after the coordinator restructure: hysteresis now lives in FleetEvictionController (config + --target-watermark), bulk DeleteObjects and the review fixes are unchanged. Tests ported to the controller suite. Not stale, ready for review.

PR #3943, opened 2026-06-29
feat(operator): make engine DaemonSet privileged mode optional

royyhuang:

The engine DaemonSet container previously hardcoded `privileged: true` for both
LMCacheEngine and CacheBlendEngine. In many clusters that is unnecessary — the
`runtimeClassName: nvidia` + `NVIDIA_VISIBLE_DEVICES=all` combination already
grants the GPU visibility LMCache needs for CUDA IPC without consuming a
device-plugin GPU. Requiring privileged mode everywhere broadens the pod's
security surface (and forces Pod Security Standards to allow the `privileged`
profile) on clusters that don't need it.

This adds an optional `spec.privileged` field (default `false`) to both engine
specs, threaded into the shared DaemonSet builder. Operators set it to `true`
only on clusters where the engine cannot otherwise see the GPUs.

- **Default flips from `true` to `false`** — this rewrites the DaemonSet pod
  template, so upgrading forces a rolling pod replacement. Existing CRs that
  relied on privileged for GPU visibility must set `spec.privileged: true`
  before upgrading. Called out in an upgrade note in the README.
- **AMD/ROCm requires it**: `gpuVendor: amd` has no NVIDIA-style RuntimeClass
  device injection, so privileged is the only path to `/dev/kfd`/`/dev/dri`.
  The AMD sample now sets `privileged: true` explicitly and the field godoc
  documents this.
- The regenerated CRD bases also pick up pre-existing `noop`/`retain` enum
  drift from already-committed Go source on `dev` — expected, not part of this
  change's intent.

**If applicable**:

### DIFF
--- a/docs/source/mp/operator.rst
+++ b/docs/source/mp/operator.rst
@@ -318,8 +318,8 @@ L2 Storage
      - List of L2 backends (``type`` + ``config``).
        See :doc:`l2_storage/index`.
 
-GPU Vendor
-~~~~~~~~~~
+GPU & Security
+~~~~~~~~~~~~~~
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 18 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3946, opened 2026-06-29
[CI] cu129 release image: use stable vLLM wheel instead of nightly

deng451e:

The `image-release-cu129` stage in `docker/Dockerfile` backs the **stable** Docker release: the `publish-image` job builds `lmcache/vllm-openai:latest-cu129` (and the versioned tag) from it whenever a non-RC GitHub release is published. That stage was still resolving vLLM from the nightly index (`https://wheels.vllm.ai/nightly/`), so stable cu129 images shipped on top of a moving nightly vLLM build instead of a pinned stable release.

This drops the nightly `--extra-index-url` and resolves vLLM from the stable PyPI wheel with the cu129 torch index hinted, switching `--index-strategy` from `unsafe-first-match` to `unsafe-best-match`. It mirrors the existing cu13 `image-release` stage so both stable release images install stable vLLM.

Build-stage-only change; the cu13 `image-release` stage already uses this exact pattern. No user-facing docs reference the cu129 vLLM build index, so no docs change is needed.

**If applicable**:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -126,11 +126,12 @@ WORKDIR /workspace
 ENTRYPOINT ["/opt/venv/bin/vllm", "serve"]
 
 #################### vLLM IMAGE & LMCache (Release, cu129) ######################
-# Installs nightly cu129 vLLM  
+# Installs the stable cu129 vLLM release wheel and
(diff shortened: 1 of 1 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3949, opened 2026-06-30
[RawBlock] Support padded payloads in batched io_uring put_many

nayeonikim:

Regular io_uring `RawBlockCore.put_many` batches writes through `batched_write`, but padded entries (`payload_len < total_len`) fall back to per-entry `write_uring` calls because `batched_write` has no separate payload length. This PR lets padded payloads use the same batched path.

- Add an optional `payload_lens` argument to `batched_write` after `placement_ids`. Omitting it keeps `payload_lens == total_lens`.
- Share padded write preparation between `write_uring` and `batched_write`: the source is read only within its bounds, and `[payload_len, total_len)` is always written as zeroes.
- Route all regular io_uring writes in `RawBlockCore.put_many` through `batched_write`.

- Padded `RawBlockCore` payloads are still bounced in Rust because `byte_array` exposes only the payload bytes. The gain is fewer submissions, not zero-copy.
- `batched_write` used to copy `total_len` bytes from each source, which could read past a shorter buffer. It now reads at most `payload_len` bytes and raises `ValueError` for a buffer shorter than that.
- `io_uring_cmd` writes are unchanged.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/raw_block/core.py
+++ b/lmcache/v1/storage_backend/raw_block/core.py
@@ -1291,26 +1291,16 @@ class RawBlockCore:
             )
             return
 
-        can_batch = all(
-            int(payload_len) == int(total_len)
-            for payload_len, total_len in zip(payload_lens, total_lens, strict=True)
+        # batched_write takes payload_lens and total_lens separately, so it
+        # handles O_DIRECT padding (payload_len < total_len) by bouncing and
+        # zero-filling internally. All io_uring writes go through one batch.
+        batch_id = raw_dev.batched_write(
+            [int(offset) for offset in offsets],
+            list(buffers),
+            [int(total_len) for total_len in total_lens],
+            [int(payload_len) for payload_len in payload_lens],
         )
-        # batched_write currently accepts only total_lens. When O_DIRECT padding
-        # is required, payload_len < total_len, so fall back to write_uring; it
-        # carries both lengths and lets Rust build the aligned padded transfer.
-        if can_batch:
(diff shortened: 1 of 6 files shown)


nayeonikim:
@DongDongJu This is the follow-up I mentioned in #3636 (
[comment](https://github.com/LMCache/LMCache/pull/3636#discussion_r3404195479)). It extends batched_write to carry (payload_len, total_len), allowing padded O_DIRECT put_many writes to stay on the batched path instead of falling back to the per-entry write_uring path.

Marked **draft** because the branch is stacked on top of #3636
(`perf/rawblock-put-many-iouring-batch`). I'll rebase onto `dev` once
Part A lands — no rush to review until then.

Single commit (`f184295`) vs. #3636. The shape:

- New shared helper `prepare_iouring_write_buffer` in
  `rust/raw_block/src/lib.rs` unifies padding semantics between serial
  `write_uring` and `batched_write` (zero-fill `[payload_len, total_len)`,
  keep direct/fixed-buffer fast path when the source tail is already
  zero, bounce + explicit zero-fill otherwise).
- `batched_write` gains an optional `payload_lens` arg. The 3-arg form
  stays identical, so existing callers are unaffected.
- `RawBlockCore.put_many` now stays on the batched path even when
  `payload_len < total_len`.

Correctness validation (real PM1753 NVMe + O_DIRECT smoke + full raw
block pytest) is in the PR body. The previously-implicit fallback
path is now both explicit in code and covered by tests.

Flagging when you have time — happy to wait on this.

PR #3951, opened 2026-06-30
[Patch][Docs]: allow bare host in lmcache.mp.host

sammshen:

Setting `lmcache.mp.host` to a bare host/IP (e.g. `127.0.0.1`) — exactly as the docs example showed — crashed EngineCore on startup:

```
zmq.error.ZMQError: Invalid argument (addr='127.0.0.1:5557')
```

ZeroMQ requires an explicit transport scheme. The connector builds the server URL as `f"{host}:{port}"`, so a bare host yields `127.0.0.1:5557`, which `zmq.Socket.connect` rejects. The default (`tcp://localhost`) hid the issue because the scheme was baked into the default string; any user override of the host silently broke it.

This PR normalizes all resolved server URLs through a new `_ensure_zmq_scheme()` helper that prepends `tcp://` when no `<scheme>://` prefix is present. Already-schemed URLs (`tcp://`, `ipc://`, …) pass through unchanged, so the multi-server `lmcache.mp.server_urls` path is unaffected.

Also fixes the docs that produced the bad command:
- `quickstart.rst` / `configuration.rst` examples used a bare host → now use `tcp://`.
- `quickstart.rst` makes the server `--host`/`--port` explicit so the `lmcache server` and `vllm serve` commands visibly line up, surfaces the full `kv_connector_extra_config` (previously you had to hunt for where it goes), and notes the host must carry a transport prefix.

**If applicable**:

### DIFF
--- a/docs/source/getting_started/quickstart.rst
+++ b/docs/source/getting_started/quickstart.rst
@@ -34,7 +34,9 @@ This guide helps you get LMCache running end-to-end in a couple of minutes. Use 
          .. tab-item:: MP mode (recommended)
             :sync: mp
 
-            Start the LMCache server:
+            Start the LMCache server. ``--host`` / ``--port`` set the ZMQ
+            address vLLM connects to; they are spelled out here so the two
+            commands line up (these are also the defaults):
 
             .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


sammshen:
@maobaolong @chunxiaozheng would you like to take a look?

deng451e (approved):
LGTM

KuntaiDu (approved):
LGTM!

PR #3953, opened 2026-06-30
[CLI] bench engine: add back navigation, exit, and URL shorthand to interactive setup

sammshen:

Makes the `lmcache bench engine` interactive setup revisable and easier to exit. Previously the guided TUI only moved forward — a typo in an early field meant restarting, and there was no clean way to bail out.

This PR adds three things:

- **Go back** — type `<` at any prompt (after the first) to step back to the previous question and re-answer it. One consistent key across both typed and arrow-key prompts (it was inconsistent before: typed prompts vs. Esc). The flow is now driven by a single loop that tracks a trail of committed answers and un-sets the most recent one when stepping back, so going back works uniformly across required items, the general/workload gates, and the final action menu.
- **Exit** — a new **Quit** action in the final menu, plus clean `Ctrl-C` / `Ctrl-D` handling at any prompt. The latter matters because the arrow-key selector runs in raw mode, where Ctrl-C would otherwise not interrupt.
- **URL shorthand** — for the Engine URL and LMCache URL fields, a bare port (`8000`) expands to `http://localhost:8000`, and a bare host/`host:port` gains an `http://` scheme. Fully-qualified URLs pass through unchanged.

- Scope is the interactive flow only. URL normalization is applied to prompted values; bare ports passed via `--engine-url` on the non-interactive command line are not normalized (can extend to the argparse layer in a follow-up if desired).
- `GO_BACK` is a sentinel returned by the terminal prompt primitives; the driver compares with `is`. Gate decisions are tracked outside `InteractiveState` so they never leak into exported configs.

**If applicable**:

### DIFF
--- a/lmcache/cli/commands/bench/engine_bench/interactive/__init__.py
+++ b/lmcache/cli/commands/bench/engine_bench/interactive/__init__.py
@@ -7,6 +7,7 @@ returns a complete ``argparse.Namespace`` ready for the orchestrator.
 """
 
 # Standard
+from typing import cast
 import argparse
 import sys
 
@@ -20,6 +21,7 @@ from lmcache.cli.commands.bench.engine_bench.interactive.state import (
 from lmcache.cli.commands.bench.engine_bench.interactive.terminal import (
     BOLD,
     CYAN,
+    GO_BACK,
     RESET,
     YELLOW,
     prompt_bool,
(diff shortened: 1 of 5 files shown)


sammshen:
@maobaolong @chunxiaozheng would you like to take a look?

deng451e (approved):
LGTM

KuntaiDu (approved):
LGTM!

PR #3955, opened 2026-06-30
Pipeline batched EXISTS in the Valkey connector

ksmotiv8:

`batched_contains` scans for the longest existing prefix of a key list. It submitted one `EXISTS` per key across the connector's worker pool, so N keys took roughly `ceil(N / num_workers)` round-trips. This pipelines all the checks into a single non-atomic GLIDE `Batch` run with `exec`, so the scan is one round-trip regardless of N. The prefix scan gates time-to-first-token on L2 lookups, so the round-trips add up.

- `submit_batch_exists` and `_do_batch_exists` are added to `_ThreadWorkerPool`; `_count_consecutive_exists` and `_batched_async_contains` call them instead of fanning out per key.
- `exec` keeps submission order, so the consecutive-prefix count is identical to the old path. `raise_on_error=True` only surfaces a real batch or transport error, which matches `Future.result()` raising on the per-key path.
- Measured on Valkey 9.1 with 8 workers: the prefix-scan EXISTS rate goes from about 17k to 220k ops/s at 1024 keys, and the gain grows with key count.
- No public API or behavior change, only the internal round-trip pattern.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/valkey_connector.py
+++ b/lmcache/v1/storage_backend/connector/valkey_connector.py
@@ -314,6 +314,34 @@ class _ThreadWorkerPool:
         """Check if a key exists (runs on a worker thread)."""
         return bool(self._get_client().exists([key_str.encode()]))
 
+    def _do_batch_exists(self, key_strs: List[str]) -> List[bool]:
+        """EXISTS many keys in a single pipelined round-trip (worker thread).
+
+        Builds a non-atomic GLIDE ``Batch`` of one EXISTS per key and runs it
+        with ``exec``, so N keys cost one round-trip instead of N individual
+        submissions fanned across the pool.
+
+        ``exec`` returns command results in submission order (the Redis
+        pipeline ordering guarantee), so ``results[i]`` is the EXISTS result
+        for ``key_strs[i]``.  EXISTS has no per-key logical failure mode, so
+        ``raise_on_error=True`` only surfaces a genuine batch/transport error
+        — which should propagate, matching the per-key path's
+        ``Future.result()``.
+        """
+        if not key_strs:
+            return []
+        # Third Party
+        import glide_sync  # type: ignore[import-untyped]
+
+        client = self._get_client()
+        batch = glide_sync.Batch(is_atomic=False)
+        for key_str in key_strs:
+            batch.exists([key_str.encode()])
+        results = client.exec(batch, raise_on_error=True)
+        if results is None:
+            raise RuntimeError("GLIDE Batch.exec returned no results for EXISTS")
+        return [bool(r) for r in results]
+
     def submit_set(self, key_str: str, data: bytes) -> Future:
         """Submit a SET operation."""
         return self._executor.submit(self._do_set, key_str, data)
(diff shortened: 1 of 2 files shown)


maobaolong:
@ksmotiv8 This is a failed test relevant to this PR.

```
tests/v1/storage_backend/test_valkey_connector.py::test_valkey_batch_operations FAILED                                                                   [ 78%]
--
  |  
  | =========================================================================== FAILURES ===========================================================================
  | _________________________________________________________________ test_valkey_batch_operations _________________________________________________________________
  |  
  | valkey_url = 'valkey://mock.local:0', local_backend = <lmcache.v1.storage_backend.local_cpu_backend.LocalCPUBackend object at 0x7fb7864984a0>
  | valkey_config = LMCacheEngineConfig(chunk_size=256, local_cpu=True, max_local_cpu_size=5.0, local_cpu_use_hugepages=False, reserve_loc...tate_cache=False, max_hidden_state_cpu_size=2.0, hidden_state_layers=None, hidden_states_retrieve_mode='prefix_strict')
  | autorelease_v1 = <function autorelease_v1.<locals>._factory at 0x7fb7861cdf80>
  |  
  | def test_valkey_batch_operations(
  | valkey_url, local_backend, valkey_config, autorelease_v1
  | ):
  | """Test batch operations: batched_put, batched_get, batched_async_contains."""
  | async_loop, async_thread = init_asyncio_loop()
  |  
  | try:
  | connector = autorelease_v1(
  | CreateConnector(valkey_url, async_loop, local_backend, valkey_config)
  | )
  |  
  | num_keys = 10
  | keys = [dumb_cache_engine_key(i) for i in range(num_keys)]
  |  
  | # Batch exists — all should be False initially
  | future = asyncio.run_coroutine_threadsafe(
  | connector.batched_async_contains("test_lookup", keys), async_loop
  | )
  | >           assert future.result() == 0, "No keys should exist initially"
  | ^^^^^^^^^^^^^^^
  |  
  | tests/v1/storage_backend/test_valkey_connector.py:272:
  | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  | /usr/lib/python3.12/concurrent/futures/_base.py:456: in result
  | return self.__get_result()
  | ^^^^^^^^^^^^^^^^^^^
  | /usr/lib/python3.12/concurrent/futures/_base.py:401: in __get_result
  | raise self._exception
  | lmcache/v1/storage_backend/connector/instrumented_connector.py:109: in batched_async_contains
  | return await self._connector.batched_async_contains(lookup_id, keys, pin)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | lmcache/v1/storage_backend/connector/valkey_connector.py:702: in batched_async_contains
  | return await self._pq_executor.submit_job(
  | lmcache/v1/storage_backend/job_executor/pq_executor.py:53: in submit_job
  | return await done
  | ^^^^^^^^^^
  | lmcache/v1/storage_backend/job_executor/pq_executor.py:80: in _worker
  | result = await fn(*args, **kwargs)


```

ksmotiv8:
Thanks for the catch. The failure was the test's MockThreadWorkerPool missing the new submit_batch_exists method the connector now calls. Fixed by adding submit_batch_exists/_do_batch_exists to the mock, mirroring the existing per-key path. Reproduced the failure locally, and with the fix the full test_valkey_connector.py suite passes (52/52). Pushed, and k3-unit-tests is green on the new commit.

PR #3967, opened 2026-07-01
[Core][Refactor] Split memory allocators into modules

DongDongJu:

  This PR modularizes the v1 memory allocator implementation by moving concrete allocator classes out of the monolithic `lmcache.v1.memory_management` module into `lmcache.v1.memory_allocators`.

  It moves the allocator implementations into focused modules for:

  - tensor, paged tensor, buffer, host, pin, GPU, mixed, paged CPU/GPU, GDS file, HIP file, ad-hoc, and Device-DAX allocators

  It keeps the existing public import surface intact:

  - existing imports from `lmcache.v1.memory_management` continue to work
  - `lmcache.v1.memory_allocators` exposes allocator classes lazily

  The follow-up compatibility patch avoids circular imports when users import allocator package/submodule paths before `memory_management`.

  This is intended to be a refactor/import compatibility change only. 
Allocator behavior is preserved; concrete allocator logic was moved into modules without changing the functional contract.

  The lazy re-export logic is included so the new module layout remains import-order safe while preserving backward compatibility.

  Validated with:

  - vLLM import/tests via `uv`
  - SGLang import/tests via `uv`
  - LMCache-side TRT-LLM tests

  **If applicable**:


DongDongJu:
@ApostaC @sammshen @deng451e @chunxiaozheng PTAL. pure move-like refactor include few exception control.
Newly added lines are mostly tests and documents.

lmcache/v1/memory_management.py
```diff
@@ -1611,1738 +1652,62 @@ def check_consistency(self) -> bool:
...
+    ),
+    "PagedCpuGpuMemoryAllocator": (
+        "lmcache.v1.memory_allocators.paged_cpu_gpu_memory_allocator"
+    ),
+    "PagedTensorMemoryAllocator": (
+        "lmcache.v1.memory_allocators.paged_tensor_memory_allocator"
+    ),
+    "PinMemoryAllocator": "lmcache.v1.memory_allocators.pin_memory_allocator",
+    "TensorMemoryAllocator": "lmcache.v1.memory_allocators.tensor_memory_allocator",
+}
```
ApostaC, line 1675: What's the difference between this and the code in `memory_allocators/__init__.py`?

ApostaC (commented):
Mostly LGTM! There is also `lmcache/v1/lazy_memory_allocator.py` which can be moved under `memory_allocators`

lmcache/v1/memory_management.py
DongDongJu: - `memory_allocators/__init__.py` supports the new API:
  `from lmcache.v1.memory_allocators import TensorMemoryAllocator`

- `memory_management.py` keeps the old API working:
  `from lmcache.v1.memory_management import TensorMemoryAllocator`

The `memory_management.py` mapping uses fully-qualified module paths and also keeps the legacy `__all__` entries, including core types like `MemoryObj`, `MemoryFormat`, and `AddressManager`.

I will do changes for just this import in next PR. because hard to track it.

DongDongJu:
> Mostly LGTM! There is also `lmcache/v1/lazy_memory_allocator.py` which can be moved under `memory_allocators`

Moved it. Thanks!

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3978, opened 2026-07-02
[MP][Coordinator] Add Pin/Unpin Tokens API

ruizhang0101:
Adds a token-based pin/unpin API to the coordinator for MP mode.

- **Coordinator** — `POST`/`DELETE /cache/pins` pin/unpin a token sequence, protecting its keys from the coordinator's fleet-wide, per-`cache_salt` L2 eviction plan (shared L2). Body: `model_name`, `world_size`, `token_ids`, `cache_salt`.
- The coordinator resolves `token_ids` → object keys **locally** (no MP-server round-trip) via a `TokenHasher` configured to match the fleet's `--chunk-size` / `--hash-algorithm` (new coordinator CLI flags; `chunk_size` is unified with the CacheBlend chunk size).
- Pins are reference-counted and exclude keys from `L2EvictionManager.compute_eviction_plan`; unpin restores eligibility.

Includes tests and docs (`docs/source/mp/coordinator.rst`, `docs/source/cli/coordinator.rst`).


lmcache/v1/distributed/storage_manager.py
```diff
@@ -730,6 +730,30 @@ def touch_l1_keys(self, keys: list[ObjectKey]):
...
+        """Pin the given keys in L1.
+
+        Args:
+            keys (list[ObjectKey]): List of object keys to pin.
+
+        Returns:
+            int: The number of keys successfully pinned.
+        """
+        results = self._l1_manager.pin(keys)
+        return sum(1 for err in results.values() if err == L1Error.SUCCESS)
```
ApostaC: What should we do regarding L2? Also, what are the expected behavior if there are multiple L2s?

Maybe consider having the interface like
```python
def pin(self, keys, tier):
```

lmcache/v1/mp_coordinator/cache_control/eviction_manager.py
```diff
@@ -72,12 +75,35 @@ def on_remove(self, key: ObjectKey) -> None:
...
         self._policy.on_keys_removed([key])
 
+    def pin(self, keys: list[ObjectKey]) -> None:
+        """Increment the L2 pin count of each key (excludes it from eviction).
+
+        Args:
+            keys: The object keys to pin.
+        """
+        for key in keys:
+            self._pin_counts[key] = self._pin_counts.get(key, 0) + 1
```
ApostaC, line 85: It's a bit weird that in L1, we implement it in L1 manager, but for L2, we implement in the eviction manager.

Is it possible to have a similar code path for both L1 and L2?

lmcache/v1/mp_coordinator/cache_control/pin_manager.py
```diff
@@ -0,0 +1,116 @@
...
+from lmcache.logging import init_logger
+from lmcache.v1.mp_coordinator.registry import MPInstance
+
+logger = init_logger(__name__)
+
+
+class PinManager:
+    """Submit token-based pin/unpin requests to MP servers."""
+
+    async def submit_pin(
```
ApostaC: I'm a bit confused. Why does the MP server need to call `submit_pin` to the coordinator? Should it be reversed?

ApostaC (changes requested):
did a quick pass. We can discuss offline

lmcache/v1/mp_coordinator/utils/cache_utils.py
```diff
@@ -0,0 +1,65 @@
...
+        token_ids=tuple(token_ids),
+        start=0,
+        end=len(token_ids),
+        request_id="",
+        cache_salt=cache_salt,
+    )
+    chunk_hashes = token_hasher.compute_chunk_hashes(list(token_ids))
+    if not chunk_hashes:
+        return [], 0
+    obj_keys = ipc_key_to_object_keys(ipc_key, chunk_hashes, [0])[0]
```
ApostaC, line 65: This requires `--no-separate-object-groups` when launching lmcache server.
Need to update the doc

ApostaC (approved):
Otherwise LGTM!

deng451e (approved):

PR #3980, opened 2026-07-02
ci: collapse cpu-only e2e steps into a matrix

maobaolong:
The three cpu_e2e_validation steps only differ by a couple of env vars and were dragging around three copies of the same podSpec. Folded them into a Buildkite matrix + adjustments so each mode only sets the env it actually needs.

- shm: no env override, relies on the script's default fallback (unset != empty here, so we can't just pass an empty string)
- pickle: LMCACHE_SHM_NAME=""
- server-side copy: LMCACHE_MP_TRANSFER_MODE=lmcache_driven

Behavior is identical to before, just ~40 lines lighter.

### DIFF
--- a/.buildkite/k3_tests/multiprocess/pipeline.yml
+++ b/.buildkite/k3_tests/multiprocess/pipeline.yml
@@ -206,25 +206,22 @@ steps:
       # runner script, so an empty string ("") is meaningful and differs
       # from unset. `__default__` is the sentinel meaning "shm mode".
       - label: ":compression: cpu_e2e_validation ({{matrix.mode}})"
-        command: bash .github/scripts/run-cpu-e2e-validation.sh
+        # matrix.adjustments does not support `env:` in Buildkite. Instead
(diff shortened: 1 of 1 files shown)


ApostaC (approved):
LGTM!

.buildkite/k3_tests/multiprocess/pipeline.yml
hlin99, line 1: hi @maobaolong, the structure looks good, but how does this PR manage resources like below? is it missed?

resources:
      requests:
        cpu: "8"
        memory: "256Gi"
      limits:
        cpu: "8"
        memory: "256Gi"
    volumeMounts:
      - { name: hf-cache, mountPath: /root/.cache/huggingface }
      - { name: dshm, mountPath: /dev/shm }

maobaolong: @hlin99 This might have been caused by Git diff view. You might think I missed this part, but it is actually included in the section below.

hlin99 (approved):
LGTM

PR #3985, opened 2026-07-02
[Feat]: add disable_l1_retrieval config for MP retrieval

sammshen:

Adds a `StorageManagerConfig.disable_l1_retrieval` flag (and matching
`--disable-l1-retrieval` CLI arg) for MP mode. When enabled, retrieval skips the
L1 hit-check in `submit_prefetch_task` so every key is routed to L2, regardless
of whether it is already resident in L1. A key that lives only in L1 surfaces as
a miss.

This is useful for isolating/benchmarking L2 behavior and for deployments that
want L1 to act purely as a transient staging/write buffer rather than a read
cache.

There is no existing config that makes retrieval *never* serve from L1:

- `store_policy: "skip_l1"` is a write-path policy — it deletes a key from L1
  only *after* the async L2 store finalizes (`_finalize_store`), so the key stays
  readable in L1 for the entire store window.
- The prefetch path must stage L2 hits into L1 (the caller reads from L1), so
  concurrent readers can hit those keys while they are read-locked.

So this flag provides a hard "never serve reads from L1" guarantee rather than
relying on the emergent, transient behavior above.

Implementation is minimal: no public method signatures change. The config value
is read once in `StorageManager.__init__` and turns the existing `reserve_read`
call into a conditional (empty result → all keys route to L2, flowing unchanged
through the existing miss-handling in the SPARSE/PREFIX paths). WARM mode already
never consults L1 residency, so it is unaffected.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/config.py
+++ b/lmcache/v1/distributed/config.py
@@ -244,6 +244,10 @@ class StorageManagerConfig:
     prefetch_policy: str = "default"
     """ The L2 prefetch policy name. """
 
+    disable_l1_retrieval: bool = False
+    """ When True, retrieval never serves reads from resident L1: every key is
+    routed to L2 regardless of L1 residency. """
+
     prefetch_max_in_flight: int = 8
     """ Maximum number of concurrent prefetch requests. """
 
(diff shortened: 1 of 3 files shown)


hlin99 (approved):
LGTM

maobaolong:
@sammshen Why isn't this PR active anymore?

PR #3986, opened 2026-07-02
chore(cli/metrics): move type-only imports under TYPE_CHECKING (#3730)

Anai-Guo:
### What

Part of #3730. Moves the type-only internal imports in the `lmcache/cli/metrics`
module under a `typing.TYPE_CHECKING` guard so they no longer run at import time.

| File | Moved under `TYPE_CHECKING` | Kept as runtime import |
|------|-----------------------------|------------------------|
| `formatter.py` | `Section` | `sections_to_dict` |
| `handler.py` | `MetricsFormatter`, `Section` | `JsonFormatter`, `TerminalFormatter` |
| `metrics.py` | `MetricsHandler` | `FileHandler`, `Section`, `sections_to_dict` |

### Why these are safe to move

Each moved symbol is used **only** in annotations. Symbols that are executed at
runtime are deliberately left as normal imports:

- `sections_to_dict(...)` is called.
- `JsonFormatter()` / `TerminalFormatter()` are instantiated as formatter defaults.
- `FileHandler` is used with `isinstance(...)`.
- `Section(...)` is instantiated in `metrics.py`, so it stays a runtime import there
  (only `formatter.py` / `handler.py`, where it appears purely in annotations, guard it).

Annotation references to the guarded names are quoted as forward refs
(e.g. `list["Section"]`, `Optional["MetricsFormatter"]`) so they still resolve for
type checkers. No `from __future__ import annotations` was added, to keep the change localized.

### Verification

- `ruff check` and `ruff format --check` pass on all three files.
- No runtime references to the guarded names remain outside the `TYPE_CHECKING` block.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lmcache/cli/metrics/formatter.py
+++ b/lmcache/cli/metrics/formatter.py
@@ -6,20 +6,23 @@ Formatters are attached to handlers, separating rendering from destination.
 """
 
 # Standard
-from typing import Any
+from typing import TYPE_CHECKING, Any
 import abc
 import inspect
 import json
 
 # First Party
-from lmcache.cli.metrics.section import Section, sections_to_dict
+from lmcache.cli.metrics.section import sections_to_dict
+
+if TYPE_CHECKING:
+    from lmcache.cli.metrics.section import Section
 
 
 class MetricsFormatter(abc.ABC):
     """Base class for metrics formatters."""
 
     @abc.abstractmethod
-    def format(self, title: str, sections: list[Section]) -> str:
+    def format(self, title: str, sections: list["Section"]) -> str:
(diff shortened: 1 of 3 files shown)


Anai-Guo:
Closing this one — the tracking issue #3730 was closed as completed on 2026-07-18, so this per-file batch no longer has an open task behind it and it has sat without review for three weeks.

The conversion itself is still valid if you want it (the files it touches don't use a `TYPE_CHECKING` guard on `dev` yet), so happy to reopen or resubmit if the cleanup is picked back up. Sorry for the queue noise in the meantime.

PR #3989, opened 2026-07-02
[feat] add dual-RoPE support for cb under HMA (SWA models)

deng451e:
**Description:**
Follow-up to #3930: enables CacheBlend re-RoPE for dual-RoPE SWA models (distinct local/global rope theta, e.g. Gemma 3) under the hybrid KV-cache manager.

**Changes**
- Per-engine-group rope caches: one cos/sin cache per distinct rope + group→cache mapping, carried in `CB_REGISTER_ROPE_V3` via IPC-shared caches.
- Per-group scatter: slots resolved per engine group so KV lands in the correct physical blocks under HMA.
- Fail-fast validation on rope registration/retrieve (invalid group/cache index raises instead of defaulting to group 0).

**Validation** — gemma-3-27b-it, 1×H200, HMA on:
- Correctness: `recall_match=1.0`, `output_match=1.0` (byte-identical greedy output).
- Speedup: ~3× TTFT on non-prefix reuse (2.96× @8K → 3.45× @16K), growing with context.
- Dual-RoPE path active: 2 caches, group map `[0,0,0,0,0,0,1]`.

### DIFF
--- a/lmcache/v1/multiprocess/modules/blend_v3.py
+++ b/lmcache/v1/multiprocess/modules/blend_v3.py
@@ -71,9 +71,9 @@ logger = init_logger(__name__)
 class _CBRopeState:
     """Per-instance RoPE state IPC-shared from vLLM; dangles on reallocate.
 
-    Models with per-layer-type RoPE (e.g. Gemma 3: local theta=1e4 /
-    global theta=1e6) register one cache per distinct rope and a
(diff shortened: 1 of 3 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #3990, opened 2026-07-03
[MP][Coordinator] token-based cache delete API

ruizhang0101:
## Token-based cache delete API (MP + coordinator)

Adds a coordinator-driven way to delete a token sequence's KV cache across **L1
and L2**, symmetric with the pin API (#3978).

### Coordinator — `POST /cache/delete`
Token-addressed, routed by `instance_id`. The coordinator resolves the tokens to
object keys **locally** (the same resolver the pin API uses), then issues a
single key-addressed `DELETE /cache/objects` to the target node for the
requested `tier` (`l1` / `l2` / `all`). It is pin-aware: for a non-`force` delete
that touches L2 it holds back keys protected by an L2 pin (they stay resident);
`force` deletes them and drops the pins. The response reports `requested` chunks,
`affected` keys removed, and `skipped` (L1 locks + L2 pins).

### Node — `DELETE /cache/objects` unified for L1 + L2
The existing key-addressed L2 delete now deletes **L1, L2, or both**, selected by
a `tier` field, with `force` to bypass L1 read/write locks. So a delete needs
only one endpoint and one call; the L2 adapter is resolved only when the tier
includes L2 (a pure `l1` delete works on an L1-only server).

### Supporting changes
- L1 deletion plumbing: `L1Manager.delete(keys, force)` and
  `StorageManager.delete_l1_keys`.
- A single shared key resolver `resolve_object_keys` used by pin (coordinator)
  and delete/prefetch (node); removed the duplicate
  `mp_coordinator/utils/cache_utils.py` and the node-only `resolve_l1_keys`.
- `L2EvictionManager.filter_unpinned` / `drop_pins` for pin-aware L2 delete.

Rebased on the upstream pin API (#3978). Unit tests and docs
(`docs/source/mp/http_api.rst`, `docs/source/mp/coordinator.rst`) included.


ApostaC:
Does this PR need to rebase after #3978?

ApostaC (approved):
LGTM!

deng451e (approved):

PR #3991, opened 2026-07-03
[Bugfix] Use batched get in remote backend health check if supported

ertcmm:
This PR fixes a bug in the `RemoteBackendHealthCheck` where the health check always fails when using the Mooncake storage backend. 

The Mooncake backend does not support single `get()` operations and raises `NotImplementedError` when `get()` is called. This PR updates the get step in the health check to first verify if the connector supports batched get via `connector.support_batched_get()`. If supported, it retrieves the key using `batched_get_blocking()`; otherwise, it falls back to the legacy `get_blocking()`.

None.

**If applicable**:

### DIFF
--- a/lmcache/v1/health_monitor/checks/remote_backend_check.py
+++ b/lmcache/v1/health_monitor/checks/remote_backend_check.py
@@ -289,7 +289,11 @@ class RemoteBackendHealthCheck(HealthCheck):
             future = self.backend.submit_put_task(key, put_obj)
             future.result(timeout=self._get_ping_timeout())
             # get
-            get_obj = self.backend.get_blocking(key)
+            if connector.support_batched_get():
+                get_objs = self.backend.batched_get_blocking([key])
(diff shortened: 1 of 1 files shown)


maobaolong (approved):
@ertcmm This PR LGMT overall, except for the DCO issue.

ertcmm:
> @ertcmm This PR LGMT overall, except for the DCO issue.

Thanks! I have signed off the commit and force-pushed the updates. The DCO check should pass now.

ApostaC (approved):
LGTM.
Btw, the current code is for the legacy non-MP mode, which will be gradually deprecated in Q3-Q4 this year.

To use mooncake in MP mode, please check: https://docs.lmcache.ai/mp/l2_storage/mooncake_store.html

ertcmm:
@ApostaC Got it, thanks for the confirmation and the pointers.
For the CI failure — from the logs it seems unrelated to the PR changes themselves. Would you mind re‑triggering the failed job?

Issue #2546, opened 2026-02-05
Feature Request: Add Batched Disk Write Support to LocalDiskBackend


VinsmokeSanji33:
Introduce batched disk write support in LocalDiskBackend to significantly reduce write latency, smooth disk queue depth, and improve SSD efficiency for disk-heavy KV cache workloads (e.g., vLLM + LMCache).

This feature buffers small KV eviction writes in memory and flushes them to disk in larger, sequential batches, optionally using asynchronous/background workers.

Motivation
Current Behavior

Each KV cache eviction may trigger an immediate synchronous disk write

Writes are often small and frequent

Disk I/O happens on the inference path

Observed Problems

High write latency (100 ms+)

Disk queue depth spikes under sustained load

Inference stalls due to blocking I/O

Inefficient use of SSD bandwidth (many small writes vs fewer large writes)

jayhpark530:
I completely agree. Batched/asynchronous writes seem like a promising optimization for SSD-heavy eviction workloads.
That said, we’ll need to carefully think through the extra buffering cost and compatibility with existing write semantics and on-disk formats.
I’m interested in this direction and would like to explore it further when I get a chance.

VinsmokeSanji33:
hey @jayhpark530 thanks for going through the request , i am new to the offloading study i am currently working on a disk heavy offload soution and understanding pros and cons of the workload , right now i am observing a very high read and write latency on data center grade disks , i was thinking batched async writes will help in optimizing the worload i agree with you on compatibility with exeisting system , would love to work on this further 

if possible could provide any refernce or insights to look into this feature , it will be really helpful  

thank you 
rohith behera :)

DongDongJu:
Sounds promising. If you guys dont mind, 
Please make the PR for this with vllm/sglang integration test for reference performance.
I will be happy to work on this if nobody plans to work on.

VinsmokeSanji33:
@DongDongJu @jayhpark530 

Hi, I’m still interested in working on this issue. I’ve been exploring disk-heavy KV offloading and believe batched/asynchronous disk writes can help reduce latency and improve SSD efficiency.

I’m still learning this area, so a little guidance on where to start or what parts to focus on would be really helpful.
Since the issue was marked stale, I wanted to check if this is still open for contribution. If yes, I’d be happy to work on it.

Thanks!

PR #3994, opened 2026-07-03
Fs localdisk batched write

funkpopo:

Refs #2546

This PR implements a real batched write path for `LocalDiskBackend`.

Previously, `LocalDiskBackend.batched_submit_put_task()` handled a batch by iterating over each `(key, MemoryObj)` pair and calling `submit_put_task()` one item at a time. That meant batched stores still paid per-item overhead for in-flight tracking, capacity reservation, eviction checks, refcount updates, and event-loop worker submission.

This PR changes the local disk batched put path to plan and submit the batch as a batch:

- Adds an internal `_DiskPutItem` structure to capture the key, memory object, physical size, tensor metadata, memory format, and cached-position metadata needed for disk persistence.
- Adds batched in-flight put tracking helpers to `LocalDiskWorker`:
  - `insert_put_tasks()`
  - `remove_put_tasks()`
- Reworks `LocalDiskBackend.batched_submit_put_task()` so it:
  - validates that `keys` and `memory_objs` have matching lengths;
  - deduplicates repeated keys within the same batch;
  - skips keys that already have an in-flight disk put;
  - reserves cache capacity for accepted items during one batch-planning phase;
  - evicts existing disk entries when needed before scheduling the batch;
  - increments refcounts only for accepted items;
  - submits one disk-worker job through `async_batched_save_bytes_to_disk()`.
- Adds `async_batched_save_bytes_to_disk()` to persist accepted batch items from a single worker task.
- Shares the actual single-item write/index/refcount logic between the single-key and batched paths through `_save_disk_put_item()`.
- Fixes an existing cleanup issue in the single-key `submit_put_task()` path where a capacity-reservation failure could leave the key marked as an in-flight put task.

This reduces Python scheduling overhead and lock churn for batched local-disk stores while preserving the existing public `LocalDiskBackend` API.

The batch path intentionally does not fall back to calling `submit_put_task()` for each item. The implementation keeps a real batch-planning phase and submits one worker job for the accepted items.

Batch behavior is conservative:

- duplicate keys in the same batch are skipped after the first occurrence;
- keys already present in `put_tasks` are skipped;
- items that cannot reserve disk capacity are rejected and removed from `put_tasks`;
- refcounts are incremented only after capacity reservation succeeds;
- failed write items clean up their reserved capacity, refcount, and in-flight state before surfacing an error;
- if one item fails inside the batch worker task, the remaining items are still attempted and the batch task raises a summarized error after processing the batch.

Validation run:

```bash
python -m py_compile lmcache/v1/storage_backend/local_disk_backend.py tests/v1/storage_backend/test_local_disk_backend.py

git diff --check -- lmcache/v1/storage_backend/local_disk_backend.py tests/v1/storage_backend/test_local_disk_backend.py

pytest -xvs tests/v1/storage_backend/test_local_disk_backend.py
```

**If applicable**:


ApostaC:
Hi @funkpopo , thanks for contributing to LMCache!

This PR is for non-MP mode, which will be gradually deprecated in Q3-Q4. We recommend you take a look at MP mode, where we have good support for FS local disk (implemented in C++). 

Please take a look at our docs: https://docs.lmcache.ai/mp/l2_storage/fs_native.html

PR #3997, opened 2026-07-03
[good-first-issue] chore(audit_backend): convert f-string log calls to %-format

### LINKED ISSUE #3952: [good-first-issue] core: convert f-string log calls in lmcache/v1/storage_backend/audit_backend.py to %-format
## Description

Two `logger.debug(f"...")` calls in `lmcache/v1/storage_backend/audit_backend.py` use f-string formatting. Per Python's logging docs and ruff G004, `%`-format is preferred so the message is only built when the log actually emits.

## Scope

- line 62 — `self.logger.debug(f"[AUDIT_BACKEND] Checking contains for key: {key}")`
- line 74 — `self.logger.debug(f"[AUDIT_BACKEND] Getting data for key: {key}")`

Both convert to `%s`. No behavior change; format style only.

## Fix pattern

```python
# Before
self.logger.debug(f"[AUDIT_BACKEND] Checking contains for key: {key}")
# After
self.logger.debug("[AUDIT_BACKEND] Checking contains for key: %s", key)
```

Ref #3372

kartikhans:

Closes #3952

Two `logger.debug` calls in `lmcache/v1/storage_backend/audit_backend.py`
built their message with an f-string, so the string was formatted even
when the log level was disabled. Converting both to `%`-format defers
the formatting to inside the logger, matching Python's logging-docs
recommendation and ruff G004.

- line 62 — `Checking contains for key: {key}` → `Checking contains for key: %s`
- line 74 — `Getting data for key: {key}` → `Getting data for key: %s`

No behavior change; format style only.

- Scope is exactly the two lines called out in the issue.
- `pre-commit run --all-files` on the touched file — all green.

**If applicable**:

(No user-facing change; log-format cleanup only → no docs/tests.)

### DIFF
--- a/csrc/sycl/pybind_sycl.cpp
+++ b/csrc/sycl/pybind_sycl.cpp
@@ -58,8 +58,8 @@ PYBIND11_MODULE(xpu_ops, m) {
         py::call_guard<py::gil_scoped_release>());
 
   // CacheGen / RoPE kernels (Intel XPU).  Names match the
-  // lmcache.python_ops_fallback module so lmcache._get_backend() can
-  // transparently override.
+  // lmcache.python_ops_fallback module so the backend selection in
+  // lmcache.v1.platform can transparently override.
   m.def("calculate_cdf", &calculate_cdf_xpu, py::arg("input"),
         py::arg("max_bins"));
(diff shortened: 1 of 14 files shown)


maobaolong (approved):
lgtm! Thanks for your contribution.

PR #3998, opened 2026-07-03
fix(examples/disagg): guard chat streaming loop against empty choices

### LINKED ISSUE #3475: [Bug] disagg_prefill proxy crashes on chat streaming chunks with empty choices
**Label**
Bug, PD disaggregation

**Describe the bug**
The example disaggregated prefill proxy crashes when serving /v1/chat/completions with streaming enabled.

The crash happens in:

examples/disagg_prefill/disagg_proxy_server.py

Specifically, the proxy assumes every streamed completion chunk from the decoder contains at least one choice:

completion_data["choices"][0]["text"]

However, the decoder can return a valid streaming chunk with:

{
  "choices": []
}

For example, this can happen for usage/final/metadata chunks or other non-token chunks. When that happens, the proxy raises:

IndexError: list index out of range

This causes the ASGI streaming response to fail even though the decoder request itself may have returned HTTP 200.

**To Reproduce**

1.Start a disaggregated prefill setup using the example proxy:
python /home/LMCache/examples/disagg_prefill/disagg_proxy_server.py \
  --host 0.0.0.0 \
  --port 8000 \
  --prefiller-host localhost \
  --prefiller-port 8100 \
  --num-prefillers 1 \
  --decoder-host 127.0.0.1 \
  --decoder-port 8200 \
  --decoder-init-port 8300 \
  --decoder-alloc-port 8400 \
  --num-decoders 1 \
  --proxy-host 127.0.0.1 \
  --proxy-port 8500 \
  --model /path/to/Qwen3-8B \
  --pd-buffer-size 4294967296 \
  --chunk-size 256

2.Send a streaming chat completion request to the proxy:
curl -N http://127.0.0.1:8000/v1/chat/completions \
  -d '{
    "model": "qwen3-8b",
    "messages": [
      {"role": "user", "content": "hello"}
    ],
    "max_tokens": 16,
    "temperature": 0,
    "stream": true
  }'
The proxy may crash with:
ERROR: Exception in ASGI application

3.File "/home/LMCache/examples/disagg_prefill/disagg_proxy_server.py", line 726, in generate_stream
    "content": completion_data["choices"][0]["text"]
               ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range

**Expected behavior**
The proxy should handle streaming chunks with empty choices defensively.

**Screenshots**

<img width="901" height="458" alt="Image" src="https://github.com/user-attachments/assets/a8eeec40-e000-4e4e-94ba-ef6442f6e75b" />

kartikhans:

Closes #3475

The disaggregated-prefill proxy's `/v1/chat/completions` handler converts
each decoder stream chunk into the chat-completion delta format by
indexing `completion_data["choices"][0]["text"]`. When the decoder emits
a valid non-token chunk (final `usage` chunk, metadata, keepalive, etc.)
with an empty `choices` list, that indexing raises `IndexError` — which
the surrounding `except (json.JSONDecodeError, KeyError)` clause does not
catch — and the ASGI streaming response fails even though the decoder
itself returned HTTP 200.

**Fix**: guard the branch. If `completion_data["choices"]` is empty or
missing, forward the original chunk unchanged instead of trying to
build a chat-completion delta from it. Non-delta chunks like `usage`
are already part of the OpenAI chat streaming spec, so downstream
clients know how to handle them.

- One-line behavioral guard before the existing conversion block. No
  refactor of surrounding code.
- Repro: send a chat completion request against the proxy with
  `stream=True` and `stream_options={"include_usage": true}` — the
  decoder's final usage chunk (`choices: []`) will exercise the fixed
  path.
- `pre-commit run --all-files` on the touched file — all green.

**If applicable**:

(Fixes a proxy bug in an example; behavioral scope is narrow enough
that adding a test to the examples tree seems out of proportion.
Happy to add one if a reviewer wants it.)

### DIFF
--- a/examples/disagg_prefill/disagg_proxy_server.py
+++ b/examples/disagg_prefill/disagg_proxy_server.py
@@ -714,6 +714,14 @@ async def handle_chat_completions(request: Request):
                         json_str = chunk_str[6:].strip()  # Remove 'data: ' prefix
                         if json_str:
                             completion_data = json.loads(json_str)
+                            # Decoder can emit non-token chunks (usage, final
+                            # metadata, keepalives) with an empty choices list.
+                            # Those aren't chat-completion deltas, so pass the
+                            # original chunk through unchanged instead of
+                            # indexing into an empty list.
+                            if not completion_data.get("choices"):
+                                yield chunk
+                                continue
                             chat_completion_data = {
                                 "id": completion_data["id"],
                                 "object": "chat.completion.chunk",


ApostaC (approved):
LGTM! Thanks for your contribution!

sammshen (approved):
LGTM!

PR #4000, opened 2026-07-04
feat: Add generic base discovery, starting with PinMemory

hlin99:
This PR decouples platform capabilities from individual device definitions and reduces the code changes needed to add new devices or backend capabilities.

- **Generic base discovery**
  - Registry base classes are automatically discovered from `lmcache.v1.platform.base`.
  - Adding a new base capability no longer requires manually updating registry bootstrap code.

- **Device implementation auto-discovery**
  - Device-specific implementations are scanned from platform subpackages and registered by `(base_class, device_type, impl_key)`.
  - Adding a new device backend only requires defining the implementation class with `device_type`.

- **PinMemory migration**
  - Moves `PinMemoryBackend` under `lmcache.v1.platform.base`.
  - Registers `CudaPinMemoryBackend` through the generic registry.
  - Uses `PinMemoryBackend` as the first migrated capability to validate the new discovery flow.

### DIFF
--- a/lmcache/v1/platform/_registry.py
+++ b/lmcache/v1/platform/_registry.py
@@ -1,21 +1,26 @@
 # SPDX-License-Identifier: Apache-2.0
 """Platform backend registry.
 
-Each accelerator sub-package (``platform/cuda``, ``platform/cpu``,
-future ``platform/xpu`` ...) ships a concrete
-:class:`~lmcache.v1.platform.base_ipc_wrapper.DeviceIPCWrapper`
-subclass with a ``device_type`` ClassVar and a ``wrap`` factory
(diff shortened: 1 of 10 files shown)


lmcache/v1/platform/_registry.py
ftian1, line 1: should avoid to use global variable and extract the logic to such a class?

```
class Registry:
    def register(self, base: type, impl: type, device_type: str, impl_key: str="default"): ...
    def get(self, base: type, device_type: str, impl_key: str="default") -> type: ...
    def resolve(self, base: type, device_type: str, impl_key: str="default") -> type: ...
```

hlin99:
discussed with baolong, will go in another way, so close this PR

PR #4002, opened 2026-07-05
[TPU] Add tpu integration: LMCacheStorageKVStore host-bytes backend

lokic233:
## Summary

Adds `lmcache.integration.tpu` — a host-side raw-bytes KV store (`LMCacheStorageKVStore`) that lets LMCache serve as a persistent storage tier behind the vLLM **[tpu-inference](https://github.com/vllm-project/tpu-inference)** `TPUOffloadConnector`. Companion PR: vllm-project/tpu-inference (adds the `LMCacheHostBackend` that calls this).

## Context / why this design

On TPU, KV cache lives as **JAX arrays** in an XLA mesh, and tpu-inference's `TPUOffloadConnector` already owns the HBM↔host movement, staging, and scheduler-side content hashing. Rather than port LMCache's torch/CUDA GPU connectors to JAX (LMCache has no `torch_xla` device path today, and `tpu_platform` hard-allowlists KV connectors), the connector hands LMCache a flat, host-resident **byte buffer** per KV block (bit-exact, incl. bfloat16) + a stable key, and LMCache provides the tiers below host RAM.

## What's in this PR

`lmcache/integration/tpu/lmcache_storage_kv_store.py`: `LMCacheStorageKVStore` exposes `put/get/remove/contains` over raw bytes, persisting via content-addressed `CacheEngineKey` (`model@world@worker@hash@dtype`) + disk serialization with LRU disk-usage accounting. No `torch_xla` / JAX dependency — runs on any host. Plus a README and host-only unit tests.

## Testing

`tests/v1/integration/tpu/test_lmcache_storage_kv_store.py` — 6 host-only tests (put/get/remove, overwrite, content-addressed file naming, LRU disk eviction, stable hashing, from_env). Runs standalone (`--noconftest`), no torch_xla/JAX/transformers needed.

## Real-hardware validation (tpu7x)

The paired tpu-inference `LMCacheHostBackend` + this raw-bytes path (reference `file` backend) was validated **bit-for-bit on a real tpu7x 2x2x1 slice** via tpu-inference's `offline_inference_kv_cache_verification.py`: both the default-off path and a forced-spill run produced output bit-for-bit identical to the no-connector baseline; a direct probe confirmed disk eviction+reload is bit-exact including bfloat16.

We also ran a 4-arm KV-cache benchmark (0.5B + 7B, real tpu7x): the disk tier is correct and materializes bit-exact spilled KV, and beats the native CPU-offload tier on extended-tier hit-rate. Honest caveat: on dense-reuse single-host workloads the reuse pool stays HBM-resident so the tier is rarely read back — its win is temporally-sparse reuse (large working set, low prompts/group). Follow-up: wire the full `StorageManager` so TPU serving can use the remote/distributed tiers (Redis/Mooncake/NIXL) for cross-instance sharing.

Note: this targets `dev`; I saw #3924 (device-detection refactor) landed recently — this PR adds only new files under `lmcache/integration/tpu/` so it's independent of that change.

---
*Developed with assistance from Navi (an AI coding assistant); all code was reviewed and the benchmarks were run + verified on real tpu7x hardware.*

### DIFF
--- a/lmcache/integration/tpu/README.md
+++ b/lmcache/integration/tpu/README.md
@@ -1,0 +1,57 @@
+# LMCache TPU integration
+
+Enables LMCache as a host-side KV storage tier behind the vLLM
+[tpu-inference](https://github.com/vllm-project/tpu-inference) `TPUOffloadConnector`.
+
+## Model
+
+Unlike the CUDA / XPU / HPU GPU connectors, this path does **not** move torch
+tensors out of device memory. On TPU, KV lives as JAX arrays in an XLA mesh, and
+`TPUOffloadConnector` already owns the HBM↔host movement. It hands LMCache a flat,
+host-resident **byte buffer** per KV block (serialized bit-exactly by the
+connector's value bridge) plus a stable string key.
+
+`LMCacheStorageKVStore` exposes the minimal contract the connector needs:
+
+```python
+put(key: str, data: bytes) -> None
+get(key: str) -> bytes | None
+remove(key: str) -> None
+contains(key: str) -> bool
+```
+
+and persists through LMCache machinery:
+- content-addressed `CacheEngineKey` (`model@world@worker@hash@dtype`),
+- disk serialization with LRU disk-usage accounting.
+
+This module has **no** `torch_xla` / JAX dependency and runs on any host.
+
+## Usage (from tpu-inference)
+
+```bash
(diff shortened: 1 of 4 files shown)


DongDongJu:
Hello @lokic233 Could you check the DCO and do you know the easy way to test this PR?

PR #4003, opened 2026-07-05
Explicit KV-layout support for CPU-presented host-memory backends

lokic233:
## Summary
Three independent, reviewable commits. Split-KV support is a **generic LMCache representation capability**, not a Metal-specific workaround.

### Commit A — generic split-KV registration + normalization
Support registering a layer's KV cache as a split `(key, value)` tensor pair alongside the existing fused single-tensor form. Any serving engine whose native key and value caches are physically separate (e.g. a host-memory backend exposing zero-copy views of distinct key/value buffers) can register `{layer: (key, value)}`; LMCache normalizes it into its internal per-layer structure with **no fused tensor ever allocated**, reusing the existing `TWO_X_NL_X_NB_BS_NH_HS` split format. Mismatched K/V shape or dtype is a hard `ValueError`. Fused paths unchanged. Backend-neutral: no backend import, no MLX import, no backend-name conditional.

### Commit B — explicit KV-layout resolution for CPU-presented backends
LMCache treats the `"cpu"` torch device as HND by default. Some backends present KV as CPU torch tensors whose physical layout is NHD; LMCache must be configurable to interpret those CPU-presented tensors using the backend-defined NHD layout. Authoritative precedence (canonical key `lmcache.kv_layout`, alias `kv_layout`; then vLLM runtime query; then `LMCACHE_VLLM_KV_LAYOUT`; then device default). Invalid explicit values raise before registration/transfer; layout is never silently guessed from ambiguous dimensions. Resolved layout is **connector-instance state**, never process-global.

> Ownership: LMCache owns connector-specific layout resolution, validation, format interpretation, gather and scatter. It does **not** own the backend's physical cache layout — it is configured to interpret the CPU-presented split tensors using their backend-defined NHD layout.

### Commit C — gate CUDA-only deps off Darwin
`cufile-python` and `nvtx` have no macOS wheels / fail to build on arm64 macOS. Both are imported lazily and degrade gracefully, so gate them behind `platform_system != "Darwin"`. Packaging only; no data-plane behavior change; Linux/CUDA installs unaffected.

## Tests
- `tests/v1/test_split_kv_support.py` (commit A): fused unchanged; split CPU NHD/HND normalize; split gather→scatter bit-exact; K/V never swapped; fused GPU paths unchanged; split shape/dtype mismatch raises.
- `tests/v1/test_metal_engine_driven.py` (commit B): CPU default HND; explicit NHD/HND; precedence (config > runtime > env); invalid raises; two instances don't share state; ambiguous shape no silent guess; CPU NHD gather/scatter bit-exact.
- Existing 118 fused `gpu_connector` tests pass unchanged.

DCO: all three commits signed off.

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -46,7 +46,11 @@ from lmcache.integration.vllm.kv_cache_group_edits import (
 from lmcache.integration.vllm.kv_cache_groups import (
     create_engine_group_infos_from_vllm,
 )
-from lmcache.integration.vllm.utils import mla_enabled, vllm_layout_hints
+from lmcache.integration.vllm.utils import (
+    mla_enabled,
+    read_explicit_kv_layout_from_config,
+    vllm_layout_hints,
+)
 from lmcache.utils import init_logger as lmcache_init_logger
 from lmcache.v1.multiprocess.group_view import slice_block_ids_per_group
 
@@ -581,6 +585,17 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
 
         assert vllm_config.kv_transfer_config is not None
 
+        # Resolve the explicit KV-layout override ONCE, here, and store it on
+        # the connector instance (no process-global state). Every layout
+        # consumer (register_kv_caches / KV-format detection / engine-driven
+        # gather+scatter / the worker transfer context) is passed this same
+        # value, so STORE and RETRIEVE use exactly the same decision. Reading
+        # the canonical ``lmcache.kv_layout`` (or alias ``kv_layout``) here
+        # also fails fast on an invalid value.
+        self._explicit_kv_layout = read_explicit_kv_layout_from_config(
+            vllm_config.kv_transfer_config
+        )
+
         # Multi-server: prefer lmcache.mp.server_urls (list or comma-separated
         # string) over the single-server lmcache.mp.host / lmcache.mp.port.
         server_urls_cfg = vllm_config.kv_transfer_config.get_from_extra_config(
@@ -754,7 +769,7 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
         engine_group_infos = create_engine_group_infos_from_vllm(
             kv_cache_config,
             kv_caches,
-            layout_hints=vllm_layout_hints(),
+            layout_hints=vllm_layout_hints(self._explicit_kv_layout),
         )
         self.worker_adapter.register_kv_caches(
             kv_caches, engine_group_infos=engine_group_infos
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
-from typing import TYPE_CHECKING, Literal, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Literal, Optional, Tuple
 import hashlib
 import os
 import string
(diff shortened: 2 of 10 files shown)


lokic233:
Companion vLLM-metal PR: vllm-project/vllm-metal#474

lmcache/v1/gpu_connector/cpu_connectors.py
ftian1, line 1: putting cpu_connectors.py under gpu_connector folder is a little weird, should we change the "gpu_connector" folder name to “connectors”?

lokic233:
CI note: `buildkite/k3-unit-tests` was failing on this PR **before** the refactor (first revision, which only added an in-process connector, also failed) — it appears pre-existing / infra-related, not caused by this change. Locally, the full suites this PR touches pass: `tests/v1/test_metal_engine_driven.py tests/v1/gpu_connector/ tests/v1/multiprocess/test_engine_driven_transfer.py` → **161 passed, 16 skipped**. Could a maintainer confirm whether k3-unit-tests is green on `dev` HEAD independently? Happy to rebase/adjust if it's real.

lmcache/v1/gpu_connector/cpu_connectors.py
lokic233: Good call — and it's now moot: I rewrote this PR to target the **engine-driven MP path** instead of an in-process connector, so `cpu_connectors.py` (and the `VLLMPagedMemGPUConnectorV2._stream_context` hook it needed) are gone entirely. The current diff is just 4 files: the `kv_format` detector fix, a Darwin gate in `requirements/common.txt`, an `LMCACHE_VLLM_KV_LAYOUT` env fallback in `integration/vllm/utils.py`, and a no-accelerator test. Nothing new lands under `gpu_connector/` except the one-line detector change.

On the broader point: agreed that a `connectors/` (or `kv_transfer/`) home reads better than `gpu_connector/` now that host-memory backends ride the same path — happy to do that rename as a separate cleanup PR if you'd like, since it'd touch a lot of imports and is orthogonal to this change.

— via Navi on behalf of dengcchi

lokic233:
**CI update — `k3-unit-tests` failed again on the rebased head (`d7993de`), and I have strong evidence it's pre-existing flakiness rather than this PR.** I can't confirm it from the fork side (no Buildkite retry permission and the logs are auth-gated), so I'd appreciate a maintainer re-running it or sharing the failing test name.

**Why this PR is very unlikely to be the cause:**
- The diff is 4 files: the `kv_format` detector (device-agnostic logic), a `requirements/common.txt` Darwin gate (`platform_system != "Darwin"` evaluates **true** on the Linux CI runner, so it's a no-op there), `tests/v1/test_metal_engine_driven.py` (3 CPU-only tests), and an env-var fallback in `integration/vllm/utils.py`. **Zero CUDA code paths are touched.**
- Full-suite `pytest --collect-only` on this branch: **5007 tests collected, 0 collection errors** — so it's not an import/collection abort.
- Every CPU-runnable test passes locally: `tests/v1/gpu_connector/` + `tests/v1/multiprocess/{test_engine_driven_transfer,test_lmcache_driven_layout_registry}` + the new metal test = **163 passed, 16 skipped**.

**Why I believe k3 is flaky right now:** recent `k3-unit-tests` builds are mixed across unrelated PRs — e.g. build #4899 failed on a trivial *"move type-only imports"* good-first-issue PR that can't plausibly break a real test, while #4897/#4898/#4901/#4905–4910 passed. Under `--maxfail=1` on a shared single-GPU k8s pod, an intermittent failure aborts the whole run regardless of PR content.

Could you re-run build #4911 (or point me at the failing test)? If it's a specific test my detector change touches, I'll fix it immediately. Also — `unit-tests-amd` #8931 has been stuck in *scheduled* and never started.

— via Navi on behalf of dengcchi

lokic233:
**Update: I reproduced the full `k3-unit-tests` suite on a real H100 (CUDA), and the failure is a pre-existing, PR-independent CUDA-context flake — not this change.** @ftian1 could you re-run `k3-unit-tests` on this PR when you get a chance? I don't have Buildkite retry access from the fork.

### What I ran
Exact k3 selection (`pytest --maxfail=1`, same `--ignore` set) on an H100, torch cu13, freshly built native extensions — once on this PR's head, and once on **clean `dev` with this PR reverted**.

### Result: clean `dev` (no PR) fails identically
```
= 1 failed, 1987 passed, 122 skipped, 1 xpassed in 633s =
FAILED tests/v1/gpu_connector/test_utils_shape_desc.py::test_per_layer_scalar_accessors_per_layer_list
torch.AcceleratorError: CUDA error: invalid argument   # at torch.randn(..., device="cuda"), line 115
```
The **same test fails with the same error on `dev` without any of my changes.**

### Why it's not this PR (three independent checks)
1. **Passes in isolation** — that test alone: 1 passed; its whole file: 13 passed; the entire `tests/v1/gpu_connector/` dir: 126 passed, 8 skipped. So it's not a logic/assertion bug.
2. **`dev` fails identically without this PR** (above).
3. **This PR doesn't touch it** — the diff is 4 files (`kv_format` detector, `requirements/common.txt`, `integration/vllm/utils.py`, and `tests/v1/test_metal_engine_driven.py`); none is `test_utils_shape_desc.py` or `gpu_connector/utils`.

### Nature of the failure
It's a `cudaErrorInvalidValue` raised at plain `torch.randn(device="cuda")` **tensor allocation** — before any assertion — accompanied by `CudaIPCTypes: Producer process has been terminated before all shared CUDA tensors released`. That's a **poisoned CUDA context** left by an earlier multiprocessing/CUDA-IPC test in the full-suite order (the poisoner isn't in `gpu_connector/`, which passes standalone). Under `CUDA_LAUNCH_BLOCKING=1` + `--maxfail=1`, that async error surfaces at the next CUDA call and aborts the run — so it's ordering/timing/GPU-state dependent, i.e. flaky. Consistent with the mixed k3 history on unrelated PRs.

### Also: PR-affected tests are green on CUDA
Everything this PR touches passes on the H100: `tests/v1/test_metal_engine_driven.py`, `tests/v1/gpu_connector/`, `tests/v1/multiprocess/{test_engine_driven_transfer,test_lmcache_driven_layout_registry}`, `tests/v1/test_vllm_mp_adapter.py` = **197 passed, 8 skipped**.

Happy to help harden the test isolation separately (a CUDA-context reset / `torch.cuda.ipc_collect()` between tests would likely stabilize this), but that's suite-wide and orthogonal to this PR. For now a re-run should clear it. Thanks!

— via Navi on behalf of dengcchi

lokic233:
**CI update on the reworked commit (`f90d8a3`): the layout patch's own tests are green; the one k3 red is a pre-existing, PR-independent CUDA-context flake.** Verified on a real H100.

### The test-portability issue is fixed
A previous revision had a test that assumed the KV-format detector's device type is always `cpu`. The detector actually reads a process-level `torch_device_type` (set by the platform at import: `cpu` on Apple-Metal, `cuda` on a GPU host). Fixed by pinning it to `cpu` in the CPU-scenario tests via an autouse fixture. **All 12 layout tests now pass on both a `cpu`-platform host (Apple M4) and a `cuda`-platform host (H100).**

### The remaining k3 failure is not this PR
`k3-unit-tests` #4939 (on `f90d8a3`) stops at:
```
FAILED tests/v1/gpu_connector/test_utils_shape_desc.py::test_per_layer_scalar_accessors_per_layer_list
torch.AcceleratorError: CUDA error: invalid argument   # at torch.randn(..., device="cuda"), line 115
```
This is a `cudaErrorInvalidValue` on a plain `torch.randn` **allocation** — before any assertion — accompanied by `CudaIPCTypes: Producer process has been terminated before all shared CUDA tensors released`. A poisoned CUDA context left by an earlier multiprocessing/CUDA-IPC test, surfacing at the next CUDA call under `--maxfail=1` + `CUDA_LAUNCH_BLOCKING=1`.

Three independent checks that it's not this PR:
1. **Passes in isolation** on the H100 (`1 passed`) — not a logic bug.
2. **This PR doesn't touch it** — the diff is 7 files under `lmcache/integration/vllm/` + the detector + `requirements` + one new test; `test_utils_shape_desc.py` and `gpu_connector/utils` are untouched.
3. **Clean `dev` (this PR reverted) fails identically** at the same test with the same error — reproduced on the H100.

I also tried a conftest CUDA-context-hygiene fixture (`torch.cuda.ipc_collect()` + `synchronize()` between tests); it did **not** clear it — the leak needs a more targeted fix in whichever upstream MP/IPC test drops shared CUDA tensors, which is orthogonal to this PR. Happy to open that as a separate test-infra PR if useful.

Could a maintainer re-run `k3-unit-tests` (it's ordering/timing-dependent, so a re-run may pass), or confirm it's a known flake? The layout change itself is green. Thanks!

— via Navi on behalf of dengcchi

PR #4013, opened 2026-07-06
fix(gpu_connector): accept dim-0-padded KV views at nonzero storage_offset (DeepSeek V4 mixed-compression pools)

divyvasal:
## Problem

DeepSeek V4 packs the fp8 KV cache and the fp4 Lightning-Indexer cache into **one pool buffer**, so a KV group's tensor is a base-shifted slice of that buffer — dim-0 block padding **and** `storage_offset != 0` (e.g. `shape=(112432, 64, 132), stride=(1039680, 132, 1), storage_offset=8640, dtype=uint8`).

`_validate_dim0_padded_layout` accepts the dim-0 padding (added for exactly this DeepSeek V4 integration) but hard-rejects any nonzero `storage_offset`, so `LMCacheMPConnector` fails at KV registration on both sides:
- engine side: crash at KV-cache init;
- server side: `register_kv_caches` raises the same `ValueError`, and the engine's registration then times out after 300 s ("LMCache server did not respond to register_kv_caches").

This makes the MP connector unusable for DeepSeek V4 class models.

## Why accepting the offset is safe

- Every consumer addresses tensors via `Tensor.data_ptr()`, which **already includes the storage offset** (`get_kernel_ptr` in `csrc/mem_kernels.cu`; every `KVFormatSpec.data_ptrs` implementation).
- Kernels step blocks via `PageBufferShapeDesc.block_stride_elems` — never by recomputing a tight stride from the logical shape.
- The CUDA-IPC path transports `(shape, stride, storage_offset)` verbatim and rebuilds with `Tensor.set_`.
- PyTorch validates offset + extent against the storage **at view construction** (`as_strided` / `set_` raise `setStorage: ... out of bounds`), so any tensor reaching this function is in-bounds by construction — no replacement bounds check is needed.

All other invariants (inner dims tightly packed, only dim-0 padded, no overlapping blocks) are unchanged.

## Validation

Live on 8× B200 serving **deepseek-v4-flash TP4** with `LMCacheMPConnector` → per-box `lmcache server` (S3 L2):
- KV groups with `storage_offset=45360` / `padding_per_block=251728` registered cleanly on engine and server;
- `Stored 512 tokens in 0.033 seconds`;
- repeat prompt: `Prefetch request completed (L1+L2): 2/2 retained keys`.

## Tests

- new: offset-slice acceptance (DeepSeek V4 pool-slice analog, CPU tensors — the check is pure metadata);
- new: 4-D interior-dim padding still rejected (guards the relaxation's scope);
- full `tests/v1/gpu_connector/test_utils_shape_desc.py`: **15/15 pass**.

### DIFF
--- a/lmcache/v1/gpu_connector/kv_format/contiguity.py
+++ b/lmcache/v1/gpu_connector/kv_format/contiguity.py
@@ -90,7 +90,17 @@ def _validate_dim0_padded_layout(tensor: torch.Tensor) -> int:
     * Every interior dim ``i`` satisfies
       ``stride[i] == prod(shape[i+1:])`` -- only dim-0 may carry
       padding, with ``stride[0] >= prod(shape[1:])``.
-    * ``storage_offset == 0`` -- no slice/narrow base shift.
+    * ``storage_offset`` may be nonzero (slice/narrow base shift). This
+      is safe because every consumer addresses the tensor through
+      ``Tensor.data_ptr()`` -- which already includes the storage
+      offset -- and steps blocks via
+      :class:`PageBufferShapeDesc.block_stride_elems`; the CUDA-IPC
+      path likewise transports ``(shape, stride, storage_offset)``
+      verbatim, and PyTorch validates offset + extent against the
+      storage at view construction. DeepSeek V4 hits this: the engine
+      packs the fp8 KV and the fp4 indexer caches into ONE pool
+      buffer, so a group's KV is a base-shifted slice of that buffer
+      (dim-0-padded AND ``storage_offset != 0``).
 
     Callers must pass the stride-sorted permuted view (not the original
     tensor): for tensors that are both permuted and dim-0-padded, the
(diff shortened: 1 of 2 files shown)


zhengfeihe:
Hi~

This has already been fixed in https://github.com/LMCache/LMCache/pull/3853.

You can sync your own repo with the latest `dev` branch, and the issue should be resolved. (

divyvasal:
Thanks @zhengfeihe — confirmed, #3853 covers this exactly (same storage_offset acceptance in `_validate_dim0_padded_layout` plus an IPC round-trip test). We patched against the 0.5.0 release and missed it on dev. Closing as duplicate.

For anyone else hitting this on the 0.5.0 release with DeepSeek V4 mixed-compression pools: the fix is on dev post-#3853; until a release ships it, overlaying the patched `contiguity.py` on both the engine and the MP server works (both sides validate).

PR #4015, opened 2026-07-06
fix(mp): reclaim CUDA-IPC-imported KV memory when an instance is released

### LINKED ISSUE #4014: MP server retains CUDA-IPC-mapped KV pool memory after the vLLM client dies (112 GB/GPU held until server restart)
## Symptom

Running the multiprocess flow (`LMCacheMPConnector` in vLLM → `lmcache server` on the same host): after the vLLM container is torn down (`docker rm -f`, i.e. client death without a graceful disconnect), the lmcache server process **keeps the client's CUDA-IPC-mapped KV pool memory mapped**. Observed on 8× B200 after serving deepseek-v4-flash TP4 (`--gpu-memory-utilization 0.85`):

```
nvidia-smi --query-compute-apps=pid,used_memory
200742 (lmcache server), 112252 MiB   # on each of GPUs 4-7
```

The server also logs, for later requests:

```
LMCache WARNING: No registered GPU context found for instance ID <id> (lmcache_driven_transfer.py:700)
```

— so it KNOWS the client contexts are gone, but the IPC mappings (and therefore the physical HBM) are never released. The next model launched on those GPUs then fails vLLM's startup free-memory check:

```
ValueError: Free memory on device cuda:5 (65.75/178.35 GiB) on startup is less than desired GPU memory utilization (0.85, 151.6 GiB).
```

Only restarting the lmcache server frees the memory.

## Expected

On client death / deregistration (the same condition that produces the "No registered GPU context" warning), the server should `cudaIpcCloseMemHandle` / unmap all of that instance's imported KV mappings so the HBM returns to the system.

## Environment

- lmcache 0.5.0 (server: host venv, torch 2.11.0+cu130; client: vllm/vllm-openai:0.24.0 image)
- vLLM 0.24.0, `LMCacheMPConnector` via `lmcache.mp.server_urls=tcp://127.0.0.1:6000`
- 8× B200, single host, `--ipc=host --pid=host --network host` container

Happy to test a patch — we hit this in a model-swap fleet where containers are torn down and replaced on the same GPUs many times a day.

divyvasal:
Fixes #4014.

## Problem

The MP server imports every client's KV cache via CUDA IPC. On instance release (explicit unregister, liveness reaper, server close) it drops the tensor references and calls `empty_cache()` — but **IPC-imported segments live in the caching allocator's IPC cache and are only unmapped by `torch.cuda.ipc_collect()`**. Without it the server retains the dead client's entire KV pool until process restart.

A detail that hides the bug: while the client lives, `nvidia-smi` charges the pool to the **client** pid and the server shows only its own small buffers (~0.7 GB). The moment the client dies, the pinned pool **re-charges to the server pid** — observed jump to **112,252 MiB on each of 4 GPUs** after `docker rm -f` of a vLLM (TP4, `gpu-memory-utilization 0.85`, B200) container. The next model launched on those GPUs then fails vLLM's startup free-memory check.

## Fix

A `_reclaim_device_memory()` step (`empty_cache()` + getattr-guarded `ipc_collect()`) invoked **once per release batch** on all three release paths (unregister / reaper / close).

**The load-bearing ordering:** `ipc_collect()` only frees segments whose last tensor reference is gone, so every path now drops its `ContextEntry` references *before* reclaiming — a live reference turns the collect into a silent no-op for that entry's segments. `getattr`-guarded for device modules without `ipc_collect` (xpu / musa).

## Tests

`tests/v1/multiprocess/test_ipc_memory_reclaim.py` — 9 unit tests against the real registry flow with a stubbed device module (same harness style as `test_worker_liveness.py`):
- reclaim fires on unregister / reaper / close; exactly once per batch;
- no reclaim on the warn path (unknown instance), the no-op reap scan, or an empty-registry close;
- graceful degradation when the device module lacks `ipc_collect`;
- **weakref-verified entry death before `ipc_collect` fires** on both the unregister and reaper paths (guards the ref-lifetime ordering above).

## Live validation (8× B200, vllm/vllm-openai:0.24 client TP4, single-host MP server, S3 L2)

Identical scenario, before/after:

| | before | after (this fix) |
|---|--:|--:|
| server GPU mem while client alive | ~0.7 GB/GPU | ~0.7 GB/GPU |
| client killed (`docker rm -f`, no dereg), after reaper | **112.2 GB/GPU pinned**, until server restart | **~0.69 GB/GPU**, same pid, no restart |
| reaper log | — | `Reaped GPU instance <id>: silent for ~305-329s (pinged=False)` ×4 (one per TP worker) |

Operational note for deployments doing frequent container swaps: never-pinged instances are only reaped after `worker_registration_grace_seconds` (default **3600 s**) — worth lowering (we run 300 s) so a died client's pool is pinned for minutes, not an hour.


ApostaC (changes requested):
Thanks for your contribution! At a high level this fix makes sense to me.

Just wondering if we can just call `ipc_collect` in `_release_entry`? This can simplify the code change a lot

divyvasal:
Thanks for the review @ApostaC ! You're right that the call sites carry too much ceremony. I've consolidated it, with one important caveat about where the `ipc_collect` can live.

Calling it inside `_release_entry(entry)` specifically can't work, and the reason is the whole subtlety of this bug: `ipc_collect()` only unmaps an IPC-imported segment once the last tensor reference to it is dead. Inside _release_entry, the caller's argument binding (plus the loop variable and the container being iterated at the call sites) still reference the entry — so the collect becomes a silent no-op for exactly the segments we're trying to free. In the most common failure case (a single dead client, which is the 112 GB/GPU repro from #4014) the batch has one entry, every reference is still live during the call, and the fix would reclaim nothing.

What I've done instead (keeps your simplification, keeps the ordering correct): a single `_release_entries(entries)` helper that owns the entire sequence: release each entry's resources, drop every reference (it takes the caller's container as the only remaining reference and clears it), then one `empty_cache()` + `ipc_collect()` per batch. The three call sites collapse to "build list -> call helper", and the refcount subtlety lives in exactly one documented place. Side effect worth flagging: `close()` now also goes through `_release_entry`, so it additionally unregisters layout descriptors at shutdown (previously it only closed cache contexts). Strictly more cleanup, but let me know if that's undesired.

The existing tests still pin the load-bearing behavior (weakref-verified that entry refs are dead when `ipc_collect` fires, exactly one reclaim per batch, graceful degradation on device modules without `ipc_collect`). Running the suite against a real GPU box now. Will push the revision once it's green.

royyhuang (approved):
Thanks for the PR. The fix LGTM.

royyhuang:
> What I've done instead (keeps your simplification, keeps the ordering correct): a single `_release_entries(entries)` helper that owns the entire sequence: release each entry's resources, drop every reference (it takes the caller's container as the only remaining reference and clears it), then one `empty_cache()` + `ipc_collect()` per batch. The three call sites collapse to "build list -> call helper", and the refcount subtlety lives in exactly one documented place. Side effect worth flagging: `close()` now also goes through `_release_entry`, so it additionally unregisters layout descriptors at shutdown (previously it only closed cache contexts). Strictly more cleanup, but let me know if that's undesired.

I like this refactoring. It is much cleaner now. For the side-effect on `close()`, I think it is fine? @ApostaC please confirm.

divyvasal:
Ran the tests — 9/9 green against a production install of the 0.5.0 wheel with the revised module overlaid.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -725,25 +725,79 @@ def reap_stale_instances(
...
 
     def _release_entry(self, entry: ContextEntry) -> None:
         """Release resources for a popped entry (run outside the lock).
 
+        NOTE: does NOT return device memory by itself — the caller must drop
+        every reference to *entry* and then call
+        :meth:`_reclaim_device_memory`, otherwise the CUDA-IPC-imported KV
+        segments stay mapped in this process (observed as ~112 GB/GPU
+        retained by the server after a vLLM client died, LMCache#4014).
+
```
ApostaC: Since this function only has 2 lines, and it's only called in `_release_entries`, we don't need to have a separate function to handle this.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -725,25 +725,79 @@ def reap_stale_instances(
...
+        Owns the load-bearing ordering for LMCache#4014: release each
+        entry's resources, drop EVERY reference to the entries, then run
+        one ``empty_cache()`` + ``ipc_collect()`` pass. A CUDA-IPC-imported
+        segment is only unmapped once its last tensor reference is gone, so
+        a single live reference turns the collect into a silent no-op for
+        that entry -- which is also why this takes the caller's container
+        and clears it: call sites must pass the ONLY container holding the
+        entries and must not bind them to local names (index the list for
+        any field access instead).
```
ApostaC: nit: we don't need this in the docstring. Please make sure the docstring just describes the high-level behavior that the caller needs to know, rather than describing the implementation details of the function.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -725,25 +725,79 @@ def reap_stale_instances(
...
+        last tensor referencing them is gone. Callers must therefore drop
+        all ``ContextEntry`` references BEFORE calling this; a live
+        reference turns ``ipc_collect()`` into a no-op for that entry's
+        segments. Guarded ``getattr``: non-CUDA device modules (xpu / musa)
+        do not expose ``ipc_collect``.
+        """
         torch_dev.empty_cache()
+        ipc_collect = getattr(torch_dev, "ipc_collect", None)
+        if ipc_collect is not None:
+            ipc_collect()
```
ApostaC: Similarly, we don't need to have a function for this. Put this in `_release_entries`

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -893,14 +944,23 @@ def unregister_kv_cache(self, instance_id: int) -> None:
             instance_id: The GPU instance ID (such as PID).
         """
         with self._lock:
-            entry = self._cache_contexts.pop(instance_id, None)
-        if entry is None:
+            # Comprehension keeps the popped entry OUT of any local name: `popped`
+            # must remain the only reference so _release_entries' reclaim actually
+            # unmaps the IPC segments (and mypy sees a clean list[ContextEntry]).
```
ApostaC: nit: remove those comments.

tests/v1/multiprocess/test_ipc_memory_reclaim.py
```diff
@@ -0,0 +1,232 @@
...
+    assert dev.calls == ["empty_cache", "ipc_collect"]
+    assert list(module._cache_contexts) == [3]
+
+
+def test_reaper_noop_scan_does_not_reclaim(monkeypatch) -> None:
+    """A scan that reaps nothing must not thrash the allocator."""
+    dev = _FakeTorchDev()
+    monkeypatch.setattr(gpu_mod, "torch_dev", dev)
+    module = _bare_module()
+    module._cache_contexts[1] = _entry()
```
ApostaC: Generally, we should avoid the use of private functions and private members in unit tests because they are subject to changes.

ApostaC (changes requested):
Thanks for the update. Left some nit comments.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
divyvasal: Done in b1f631a6 — `_release_entry` inlined into the `_release_entries` loop.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
divyvasal: Done — docstring reduced to the caller contract (entries must be the only remaining references; the list is cleared). The #4014 mechanism is now a one-line comment at the reclaim site.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
divyvasal: Done — `_reclaim_device_memory` inlined as well; kept the `getattr` guard since xpu/musa device modules don't expose `ipc_collect`.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
divyvasal: Removed. The ref-lifetime invariant those comments described is enforced by the weakref-based tests, so the code carries it silently.

tests/v1/multiprocess/test_ipc_memory_reclaim.py
divyvasal: Reworked — the tests now construct the module through the real `__init__`, register via the public `register_kv_cache` (with the module-level context factory stubbed), and assert only through `context_entries_snapshot()` / `reap_stale_instances()` / `unregister_kv_cache()` / `close()` plus the injected fake device module. No private members or `__new__` wiring remain.

divyvasal:
@ApostaC gentle ping when you have a cycle — review feedback was addressed and the branch has been running in our production fleet since (the reaper's CUDA-IPC reclaim path exercises on every model swap). Happy to rebase if you'd like a fresh CI run first.

ApostaC (commented):
Thanks for the update. LGTM!

ApostaC (approved):
Thanks for the update. LGTM!

PR #4016, opened 2026-07-06
[RawBlock][FDP] Add NVMe FDP discovery and placement plumbing

daegyu94:

This PR adds the raw_block FDP plumbing needed to discover NVMe Flexible Data
Placement status and carry placement identifiers through the raw-block stack.

- Adds Rust raw-block support for querying FDP RUH status via NVMe I/O
  Management Receive and exposes it to Python as
  `RawBlockDevice.fetch_fdp_status()`.
- Adds FDP placement directive support to Rust write paths, with validation that
  rejects reserved placement identifier `0` and out-of-range values.
- Exposes `RawBlockCore.fetch_fdp_status()` and validates optional placement
  identifiers for batched data writes and metadata checkpoint writes.
- Adds raw_block MP adapter configuration for `fdp_enabled`,
  `fdp_placement_ids`, `fdp_data_placement_policy`, and
  `meta_checkpoint_placement_id`, including discovery of non-zero device
  placement identifiers.
- Threads placement identifiers through the adapter/core write boundary.
- Adds cache-salt prefix based KV data placement for the MP raw_block adapter.
  When `fdp_enabled=true`, `fdp_data_placement_policy` defaults to
  `cache_salt_prefix`. Only `cache_salt` values in `bucket:suffix` form opt in
  to FDP placement. The bucket name is the case-insensitive prefix before `:`,
  so `RAG:app1`, `rag:app2`, and `rag:` share the same FDP bucket.
- Adds `fdp_data_placement_policy="cache_salt_rank"` for multi-tenant GPU nodes
  backed by FDP SSDs with many usable placement identifiers. This policy keeps
  the same `bucket:suffix` tenant bucket isolation as `cache_salt_prefix`, then
  further separates rank-local KV cache write streams within each bucket by the
  local rank encoded in `kv_rank`.
- With `cache_salt_rank`, each bucket may use up to the detected visible GPU
  count worth of placement identifiers. Placement identifiers are allocated
  lazily as local ranks appear, so applications using fewer GPUs consume fewer
  identifiers. If the per-bucket quota is zero or exceeded, or if the FDP
  placement identifier pool is exhausted, affected writes fall back to default
  NVMe placement with no FDP directive.
- Values without `:` or with an empty bucket prefix use default NVMe placement
  with no FDP directive. If cache-salt buckets exceed the available FDP
  placement identifiers, additional buckets fall back to default NVMe placement
  and the adapter emits one warning.
- Keeps metadata checkpoint placement separate from KV data placement.
  Metadata checkpoint writes can opt into a configured placement identifier;
  otherwise they use default NVMe placement.
- Documents FDP configuration behavior and hardware-gated probe commands across
  user docs, design docs, and the Rust raw-block README.

- FDP data placement is enabled by policy, not by arbitrary non-empty
  `cache_salt`. With the default `cache_salt_prefix` policy, only
  `bucket:suffix` values participate in FDP placement. Plain values such as
  `rag` preserve the existing default NVMe placement behavior.
- `cache_salt_rank` is an extension of the cache-salt bucket policy for
  multi-tenant GPU nodes. It uses the local rank encoded in `kv_rank`, not the
  full packed `kv_rank`, so different global/world-size encodings for the same
  local rank reuse the same placement identifier within a bucket.
- `cache_salt_rank` uses the detected visible GPU count as the per-bucket
  placement identifier quota. If the quota is zero, exceeded, or placement
  identifiers are exhausted, writes fall back to no FDP directive rather than
  sharing another bucket's identifiers.
- Placement identifier `0` is treated as reserved/default and is not accepted as
  an explicit directive. Use `None` to omit an FDP directive.
- `fdp_enabled` requires `io_engine="io_uring"` and `use_uring_cmd=true`,
  because FDP status and placement directives use NVMe namespace character-device
  passthrough.
- `meta_checkpoint_placement_id` must not overlap with KV data
  `fdp_placement_ids`. If `fdp_placement_ids` is omitted, the adapter uses all
  discovered non-zero identifiers except the metadata checkpoint identifier.
- Hardware probe tests are opt-in. `test_raw_block_fdp_status_probe.py` is
  status-only/read-only. Destructive raw-block tests require
  `LMCACHE_TEST_RAW_BLOCK_DESTRUCTIVE=1`.

**Testing**:

- `python -m pytest -q tests/v1/multiprocess/test_raw_block_l2_adapter.py`
  - includes `cache_salt_prefix` and `cache_salt_rank` FDP placement policy tests
- `python -m pytest -q tests/v1/storage_backend/test_raw_block_key_codec.py`
- `python -m pytest -q tests/v1/storage_backend/test_raw_block_fdp_status_probe.py`

**If applicable**:


tests/v1/storage_backend/raw_block_fdp_probe_utils.py
```diff
@@ -0,0 +1,89 @@
+# SPDX-License-Identifier: Apache-2.0
```
DongDongJu: Do we need this file?

DongDongJu (changes requested):
Hello @daegyu94, Thanks for the contribution.
I left one question about the new file.
Also, Can we add the cache_salt based auto data placement as well?

tests/v1/storage_backend/raw_block_fdp_probe_utils.py
daegyu94: > Do we need this file?

Resolved by merging it into raw_block_test_utils.py.

daegyu94: > Hello @daegyu94, Thanks for the contribution. I left one question about the new file. Also, Can we add the cache_salt based auto data placement as well?

Resolved by adding cache_salt-based automatic data placement. 
With `fdp_enabled=true`, `bucket:suffix` cache_salts are mapped to FDP placement IDs, while plain cache_salts keep default NVMe placement.

DongDongJu (changes requested):
LGTM.
Could you add the e2e real run test result to the PR description?
Can we have a better tail latency when we apply this?

daegyu94:
@DongDongJu 

We evaluated LMCache trace replay on a Samsung PM9D3a FDP-enabled SSD using the following configuration:

* Configured test capacity: 960 GiB
* Device utilization: 89%
* Benchmark write volume: 4× the configured capacity

Before each experiment, we initialized the test capacity with sequential writes using `fio` to ensure consistent starting conditions.

The two experimental configurations were:

* `no_fdp`: Trace replay without FDP placement directives
* `fdp_separated`: Trace replay with workload-class-based separation using FDP Placement Identifiers (PIDs)

### Workload

Both configurations used the same synthetic storage-level LMCache traces, with nine concurrent workers spanning 855 GiB:

* 2x chat/churn: 4 MiB objects, 90 GiB each
* 1x long-context: 32 MiB objects, 180 GiB
* 2x RAG/shared-prefix: 16 MiB objects, 90 GiB each
* 3x random-prompt/low-reuse: 2 MiB objects, 90 GiB each
* 1x metadata-heavy: 1 MiB objects, 45 GiB

For `fdp_separated`, each workload class was assigned a distinct cache salt.

### Results

| Configuration   |   WAF | Improvement |
| --------------- | ----: | ----------: |
| `no_fdp`        | 2.600 |           — |
| `fdp_separated` | 1.425 | 45.2% lower |

The residual WAF of 1.425 may be attributable to the device’s `Initially Isolated` policy and relatively large 16 GiB reclaim units, rather than to ineffective workload separation.

Latency represents the end-to-end completion time of each replayed object I/O operation.

| Metric                | `no_fdp` (ms) | `fdp_separated` (ms) | Improvement |
| --------------------- | ------------: | -------------------: | ----------: |
| Average write latency |    25,364.155 |           11,249.847 | 55.6% lower |
| Write p90 latency     |    36,748.414 |           25,960.293 | 29.4% lower |
| Average read latency  |        36.375 |               23.386 | 35.7% lower |
| Read p90 latency      |        59.809 |               46.638 | 22.0% lower |

Compared with no_fdp, fdp_separated reduced WAF by 45.2% and improved both average and p90 read/write latencies.

DongDongJu:
@daegyu94 Please resolve the conflict! result looks promising

lmcache/v1/storage_backend/raw_block/core.py
```diff
@@ -1385,6 +1513,7 @@ def _write_one(
                         payload_len,
                     ],
                     [hdr_total, total_len],
+                    [placement_id, placement_id],
```
DongDongJu, line 1529: So, in use_uring_cmd=false case, is this automatically ignored?

DongDongJu (commented):
I left one comment, otherwise LGTM.

lmcache/v1/storage_backend/raw_block/core.py
daegyu94: Yes. [None, None] is passed, so the Rust backend applies no placement directive.

daegyu94:
> @daegyu94 Please resolve the conflict! result looks promising

@DongDongJu  Resolved by rebasing onto the latest `dev`. Thanks!

DongDongJu (approved):
LGTM. @zhengfeihe Can you take a look this?

zhengfeihe:
I’ve started reviewing the PR, but since it’s relatively large, please allow me some time to review it thoroughly and try out the feature. I’ll finish the review by the end of this week.

DongDongJu:
> I’ve started reviewing the PR, but since it’s relatively large, please allow me some time to review it thoroughly and try out the feature. I’ll finish the review by the end of this week.

Thanks! see you soon.

daegyu94:
@DongDongJu @zhengfeihe 

Apologies for the late update.

I accidentally omitted placement affinity from the free-slot reuse path.
Because the RawBlock backend globally reuses slots released by cache eviction, rewrites without PID affinity can cause cross-PID pollution.

This change prefers a free slot with the same placement ID and, to preserve write availability, falls back to any available free slot instead of enforcing strict affinity. 
The observed fallback rate was below 5%.

rust/raw_block/src/lib.rs
```diff
@@ -425,6 +445,112 @@ fn nvme_identify_ns(fd: RawFd, nsid: u32) -> Result<NvmeIdNs, PyErr> {
...
+        cdw11: (data_len >> 2) - 1,
+        timeout_ms: 0,
+        result: 0,
+        ..unsafe { std::mem::zeroed() }
+    };
+
+    // SAFETY: ioctl with properly initialized command structure.
+    let rc = unsafe { libc::ioctl(fd, NVME_IOCTL_IO_CMD, &mut cmd as *mut NvmePassthruCmd) };
+
+    if rc < 0 {
```
zhengfeihe: Do we want to handle rc > 0 error case here ? 

```
enum {
	NVME_SC_SUCCESS			= 0x0,
	NVME_SC_INVALID_OPCODE		= 0x1,
	NVME_SC_INVALID_FIELD		= 0x2,
	NVME_SC_CMDID_CONFLICT		= 0x3,
	NVME_SC_DATA_XFER_ERROR		= 0x4,
	NVME_SC_POWER_LOSS		= 0x5,
	NVME_SC_INTERNAL		= 0x6,
	NVME_SC_ABORT_REQ		= 0x7,
	NVME_SC_ABORT_QUEUE		= 0x8,
	NVME_SC_FUSED_FAIL		= 0x9,
	NVME_SC_FUSED_MISSING		= 0xa,
	NVME_SC_INVALID_NS		= 0xb,
	NVME_SC_LBA_RANGE		= 0x80,
	NVME_SC_CAP_EXCEEDED		= 0x81,
	NVME_SC_NS_NOT_READY		= 0x82,
	NVME_SC_CQ_INVALID		= 0x100,
	NVME_SC_QID_INVALID		= 0x101,
	NVME_SC_QUEUE_SIZE		= 0x102,
	NVME_SC_ABORT_LIMIT		= 0x103,
	NVME_SC_ABORT_MISSING		= 0x104,
	NVME_SC_ASYNC_LIMIT		= 0x105,
	NVME_SC_FIRMWARE_SLOT		= 0x106,
	NVME_SC_FIRMWARE_IMAGE		= 0x107,
	NVME_SC_INVALID_VECTOR		= 0x108,
	NVME_SC_INVALID_LOG_PAGE	= 0x109,
	NVME_SC_INVALID_FORMAT		= 0x10a,
	NVME_SC_BAD_ATTRIBUTES		= 0x180,
	NVME_SC_WRITE_FAULT		= 0x280,
	NVME_SC_READ_ERROR		= 0x281,
	NVME_SC_GUARD_CHECK		= 0x282,
	NVME_SC_APPTAG_CHECK		= 0x283,
	NVME_SC_REFTAG_CHECK		= 0x284,
	NVME_SC_COMPARE_FAILED		= 0x285,
	NVME_SC_ACCESS_DENIED		= 0x286,
};
```

This is a reference I found on https://android.googlesource.com/kernel/mediatek/+/android-4.4.4_r3/include/linux/nvme.h

docs/source/mp/l2_storage/raw_block.rst
```diff
@@ -63,7 +84,49 @@ caller-provided load buffers during prefetch.
...
+  using different GPU subsets; 128 identifiers can cover up to 16 applications
+  with eight rank streams each, while 256 identifiers can cover up to 32.
+- Bucket-to-placement assignments are process-local. Restart recovery does not
+  need them for correctness because ``cache_salt`` is part of the object key,
+  but first-seen FDP placement assignments may change after adapter restart.
+- Slot affinity is process-local and is not stored in metadata checkpoints.
+  After recovery, free slots have no recorded affinity until they are reused.
+- Store and retrieve requests must use the same ``cache_salt`` to address the
+  same object. This is the normal LMCache key identity rule; FDP placement is a
+  write directive and is not used to locate data on reads.
```
zhengfeihe, line 136: If 16 vLLM services use the same cache_salt prefix, they all fall into one bucket: rank 0 of every service shares one stream, rank 1 shares another, and the whole bucket is capped at visible_gpu_count streams. 

The  "16 applications × 8 rank streams" only happens when each service uses its own prefix ?

(Correct me if I am wrong on this

This is still better than before, but worth clearly stated in the doc if this is it .

zhengfeihe (commented):
I have two minor comments. Learned a lot from reviewing this PR! Thanks for the tremendous work!

rust/raw_block/src/lib.rs
daegyu94: @zhengfeihe 

Good catch. A negative return value represents an ioctl/system error, while a positive return value represents an NVMe command completion status. The previous check incorrectly treated a controller-reported command error as success.

I updated the code to handle rc > 0 separately and return an error containing the NVMe status value. I also applied the same handling to the Identify admin-command path, which had the same rc < 0-only check.

docs/source/mp/l2_storage/raw_block.rst
daegyu94: @zhengfeihe 

Yes, your understanding is correct.

`cache_salt_rank` assigns placement streams using `(case-insensitive bucket prefix, local_rank)`. For example, `batch:app-a` and `batch:app-b` belong to the same `batch` bucket. Local rank 0 from both applications shares one placement identifier, local rank 1 shares another, and so on, up to `visible_gpu_count` streams for that bucket. The suffix remains part of the cache key but does not affect FDP placement.

This sharing can be intentional because the number of FDP placement identifiers/RUHs available on an SSD is limited. An operator may group applications with similar cache lifetimes under the same prefix so that they share placement streams.

To obtain 16 applications × 8 distinct rank streams, each application must use a unique bucket prefix, for example:

```text
app01:<suffix>
app02:<suffix>
...
app16:<suffix>
```

This can allocate up to 128 distinct placement identifiers, assuming enough identifiers are available and the adapter detects at least eight visible GPUs.

I've clarified this sharing behavior and the assumption behind the example in the documentation. Thanks for pointing it out.

zhengfeihe (approved):
LGTM～ （I will let Dongju press the merge button

PR #4018, opened 2026-07-06
fix(memory): defer host pinning in LazyMemoryAllocator to first allocate

### LINKED ISSUE #4019: [Bug] LazyMemoryAllocator pins the host pool in __init__, creating a CUDA context on cuda:0 at MP-server start (before any worker connects)
**Label**

bug

**Describe the bug**

In a Dynamo k8s deployment (LMCache MP mode + DGD), bringing up the LMCache MP-server pod alone is enough to make its process appear on `cuda:0` in `nvidia-smi`, holding ~616 MiB with zero workers connected. The workers may later land on GPUs that have nothing to do with `cuda:0`, so the process ends up holding memory on a GPU that no part of the deployment actually uses. This is harmless when the deployment owns the whole node (all 8 GPUs), but becomes a real problem when using fewer than 8: whatever other workload is scheduled onto `cuda:0` now shares it with this ~616 MiB ghost process.

When the multiprocess (MP) cache server starts, `LazyMemoryAllocator.__init__` immediately pins the initial chunk of the CPU KV pool via `cudaHostRegister` (`torch_dev.ext.pin_memory(..., cudaHostRegisterMapped)`). Host pinning creates a CUDA primary context on the current device, which at server start is the default device `cuda:0`. The server therefore allocates a ~616 MiB context on `cuda:0` before any worker has connected, on a GPU it may never use. With the operator's `NVIDIA_VISIBLE_DEVICES=all`, this always lands on `cuda:0` regardless of which GPU(s) the workers actually run on.

**Environment**

- LMCache: server image `lmcache/vllm-openai` (observed `nightly-2026-04-25` ... `0.5.1.dev45`); operator `lmcache/lmcache-operator:v0.5.0`
- GPU: NVIDIA B200 x8, `NVIDIA_VISIBLE_DEVICES=all` (operator default)
- Deploy: `LMCacheEngine` CR (`l1.sizeGB: 200`), MP mode (CUDA-IPC)
- K8s : Dynamo 1.2.0

**To Reproduce**

1. Apply an `LMCacheEngine` CR alone (no paired workers / DGD).
2. Wait for the manager DaemonSet pod to be `Running`.
3. Run `nvidia-smi` on the node. The MP-server python process holds ~616 MiB on GPU 0, with no worker connected:

```text
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI          PID   Type   Process name                          GPU Memory   |
|=========================================================================================|
|    0   N/A  N/A        721984    C   /opt/venv/bin/python3                     616MiB    |
+-----------------------------------------------------------------------------------------+
# LMCacheEngine CR applied ALONE (no DGD / no worker). PID 721984 = MP server.
# It sits on GPU 0 only, deterministically. Re-applying the CR always lands here.
```

**Expected behavior**

The MP server is a host-memory cache (`l1.sizeGB` lives in RAM) and does not need a GPU of its own. It should hold no CUDA context at boot. If a pinned-pool context is required at all, it should be created lazily on the GPU where KV transfers actually happen (the worker's device), not on `cuda:0` before any worker exists.

**Additional context**

- Root cause is in `lmcache/v1/lazy_memory_allocator.py::__init__`:

  ```python
  self._pin_memory_chunk(0, self._curr_size)   # cudaHostRegister -> CUDA context on current device (cuda:0)
  self._allocator = TensorMemoryAllocator(...)
  self._expand_thread.start()                   # background thread pins more, also on cuda:0
  ```

  `_pin_memory_chunk` calls `torch_dev.ext.pin_memory(ptr, size, 2)` (`cudaHostRegisterMapped`), which creates the primary context on whatever device is current: `cuda:0` at server start.
- Also independent of transfer mode: `LMCACHE_MP_TRANSFER_MODE=engine_driven` leaves the GPU-0 context unchanged, so it is the boot-time pin, not IPC/worker activity.
- Impact: ~600 MiB wasted on `cuda:0` on every MP server even when idle; an unexpected co-tenant for whatever workload is scheduled onto `cuda:0` (can contribute to OOM); the context sits on a device unrelated to where transfers happen.

changhyeonnam:
Fixes #4019 

#### Problem
 
`LazyMemoryAllocator.__init__` pins the CPU KV pool with `cudaHostRegister` (`torch_dev.ext.pin_memory(..., cudaHostRegisterMapped)`). Host pinning creates a CUDA primary context on the current device. Since `__init__` runs at MP-server start, before any worker connects, the context lands on the default device `cuda:0`.
 
```python
def __init__(self, ...):
    self._pin_memory_chunk(0, self._curr_size)   # cudaHostRegister -> CUDA context on cuda:0
    ...
    self._expand_thread.start()                  # background thread pins more, also on cuda:0
```
 
As a result, the server holds a ~616 MiB context on `cuda:0` even when idle with zero workers. The MP server is a host-memory cache and does not need a GPU of its own. On a shared node this becomes an unexpected co-tenant on `cuda:0` and can contribute to OOM. Reproducible with the `LMCacheEngine` CR alone (no workers), and independent of transfer mode, so it is the boot-time pin, not IPC/worker activity.
 
 #### Solution

Defer pinning (and the background expansion thread) out of `__init__` to a new idempotent `ensure_pinning(device)`, triggered lazily by the first `allocate()` /`batched_allocate()`. The first allocation runs inside the worker's device
context, so the pinned-pool CUDA context lands on a real worker GPU instead of`cuda:0`.
- `__init__` now only does host-side work (buffer alloc, address bookkeeping) : no CUDA calls, no context.
- The chosen device is remembered (`_pin_device`) and the background expansion thread pins subsequent chunks inside the same device context.
- `close()` takes the init lock and sets a `_closed` flag, so a shutdown racing the first allocate can neither leak the expansion thread nor let pinning start after the buffer is freed. Unpinning runs in the same device context as pinning.

#### Verification

- Unit tests (`tests/v1/test_lazy_memory_allocator.py`, 9 cases, no GPU required — the device layer is mocked and the tests observe pin/unpin calls and the device context they run in):
  - construction performs no CUDA calls (core contract)
  - first allocate pins exactly once, on the current device
  - `ensure_pinning` is idempotent and pins once under 8-thread contention
  - `close()` is safe before pinning, unpins after pinning, and blocks late pinning

- End-to-end on a multi-GPU node (patched nvidia dynamo image overlaid on `lmcache/vllm-openai:latest-nightly`):
  - booting the MP engine pod alone: no CUDA context on any GPU (previously ~616 MiB squatted on `cuda:0` at. idle)
  - after launching the nvidia dynamo DGD, each worker's pinned-pool process appears on the GPU that worker actually runs on

### Special notes for your reviewers :
- The fix relies on the first `allocate()` running inside the worker's device context. This holds for the MP-server path (allocation happens during the worker's transfer, under `with torch_dev.device(...)`) and was verified end-to-end with a dynamo DGD deployment. If a new caller ever allocates outside a device context, pinning would bind to that thread's current device; `ensure_pinning(device)` is public so such a caller can bind explicitly.
- Deferring the thread start makes `close()` racy against a concurrent first allocate, so `close()` now takes the init lock and sets `_closed`. Worth a close look at the lock protocol (`ensure_pinning` / `close`); the expansion thread never takes the lock, so joining under it cannot deadlock.
- Intentionally unchanged to keep the PR scoped: the warning-only behavior on `pin_memory` failure and the alignment asserts in `_pin_memory_chunk` are pre-existing.

### If applicable:

### DIFF
--- a/lmcache/v1/distributed/l1_manager.py
+++ b/lmcache/v1/distributed/l1_manager.py
@@ -5,7 +5,7 @@ Managing objects and memory for L1 cache
 
 # Standard
 from dataclasses import dataclass
-from typing import Literal
+from typing import TYPE_CHECKING, Literal
 import threading
 
 # First Party
@@ -27,6 +27,10 @@ from lmcache.v1.memory_management import MemoryObj
 from lmcache.v1.mp_observability.event import Event, EventType
 from lmcache.v1.mp_observability.event_bus import get_event_bus
 from lmcache.v1.mp_observability.otel_init import register_gauge
+
+if TYPE_CHECKING:
+    # Third Party
+    import torch
 
 logger = init_logger(__name__)
 
@@ -238,6 +242,18 @@ class L1Manager:
         """
         with self._lock:
             self._registered_listeners.append(listener)
+
+    def warm_up(self, device: "int | torch.device") -> None:
+        """Begin deferred L1 memory initialization for ``device``.
+
+        Forwards to the memory manager tier; a no-op for tiers without
+        deferred host pinning.
+
+        Args:
+            device: Device whose context the deferred initialization
+                should run under.
+        """
+        self._memory_manager.warm_up(device)
 
     @l1_mgr_synchronized
     def reserve_read(
--- a/lmcache/v1/distributed/memory_manager/gds_l1_memory_manager.py
+++ b/lmcache/v1/distributed/memory_manager/gds_l1_memory_manager.py
@@ -2,7 +2,11 @@
 """GDS slab-file L1 memory manager."""
 
 # Standard
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    # Third Party
+    import torch
 
 # First Party
 from lmcache.integration.vllm.utils import get_size_bytes
@@ -134,6 +138,9 @@ class GDSL1MemoryManager:
     def close(self) -> None:
         """No-op: the GDSContext owning the slab is closed at server shutdown."""
         return
+
+    def warm_up(self, device: "int | torch.device") -> None:
+        """No-op: the GDS slab tier has no deferred host pinning."""
 
     def memcheck(self) -> bool:
         """For debug purposes; logs allocator state and checks consistency.
--- a/lmcache/v1/distributed/memory_manager/l1_manager_protocol.py
+++ b/lmcache/v1/distributed/memory_manager/l1_manager_protocol.py
@@ -2,7 +2,11 @@
 """Structural interface shared by the L1 memory manager tiers."""
 
 # Standard
-from typing import Optional, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, Optional, Protocol, runtime_checkable
+
+if TYPE_CHECKING:
+    # Third Party
+    import torch
 
 # First Party
 from lmcache.v1.distributed.api import L1BackendType, MemoryLayoutDesc
@@ -42,6 +46,13 @@ class L1ManagerProtocol(Protocol):
         """Describe the underlying L1 buffer for L2-adapter registration.
 
         Returns ``None`` for tiers with no registerable L1 buffer (e.g. GDS).
+        """
+        ...
+
+    def warm_up(self, device: "int | torch.device") -> None:
+        """Start deferred initialization (e.g. host pinning) for ``device``.
+
+        No-op for tiers without deferred initialization.
         """
         ...
 
--- a/lmcache/v1/distributed/memory_manager/l1_memory_manager.py
+++ b/lmcache/v1/distributed/memory_manager/l1_memory_manager.py
@@ -3,6 +3,11 @@
 
 # Standard
 from multiprocessing import shared_memory
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # Third Party
+    import torch
 
 # First Party
 from lmcache.logging import init_logger
@@ -215,6 +220,18 @@ class L1MemoryManager:
             size=self._size_in_bytes,
             align_bytes=self._align_bytes,
         )
+
+    def warm_up(self, device: "int | torch.device") -> None:
+        """Begin deferred allocator initialization for ``device``.
+
+        Forwards to the allocator; a no-op for allocators without
+        deferred host pinning.
+
+        Args:
+            device: Device whose context the deferred initialization
+                should run under.
+        """
+        self._allocator.warm_up(device)
 
     def close(self) -> None:
         """
--- a/lmcache/v1/distributed/storage_manager.py
+++ b/lmcache/v1/distributed/storage_manager.py
@@ -6,7 +6,7 @@ Distributed multi-tier storage manager for MP mode
 # Standard
 from contextlib import contextmanager
 from dataclasses import replace
-from typing import Iterator, Literal, Optional
+from typing import TYPE_CHECKING, Iterator, Literal, Optional
 import threading
 import time
 
@@ -61,6 +61,10 @@ from lmcache.v1.mp_observability.trace.decorator import (
     publish_call_event,
 )
 from lmcache.v1.platform import HAS_EVENTFD
+
+if TYPE_CHECKING:
+    # Third Party
+    import torch
 
 logger = init_logger(__name__)
 
@@ -166,6 +170,19 @@ class StorageManager:
             ),
             self.get_l2_usages,
         )
+
+    def warm_up(self, device: "int | torch.device") -> None:
+        """Begin deferred L1 memory initialization for ``device``.
+
+        Called at worker registration so host pinning overlaps engine
+        startup instead of the first request; a no-op for tiers without
+        deferred pinning.
+
+        Args:
+            device: Device whose context the deferred initialization
+                should run under.
+        """
+        self._l1_manager.warm_up(device)
 
     # External APIs for serving engine integration code to call
     @enable_tracing()
--- a/lmcache/v1/lazy_memory_allocator.py
+++ b/lmcache/v1/lazy_memory_allocator.py
@@ -271,7 +271,8 @@ class LazyMemoryAllocator(MemoryAllocatorInterface):
         pinned-pool CUDA context lands on the worker GPU.
         """
         if not self._pinning_started:
-            self.ensure_pinning(torch_dev.current_device())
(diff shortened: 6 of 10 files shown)


ApostaC:
Hi, thanks for contributing to LCMache.

I think this will create an extra, large delay during the inference (first request), which is what we don't want.

changhyeonnam:
@ApostaC 
You're right, the first allocate pins `init_size` (default 20 GB), so the first request stalls.

I could pin at a warm-up point instead. `ensure_pinning(device)` takes the device, so the registration path can call it once the worker's device is known, before any request. That keeps the pin off the request path and on the worker GPU, not `cuda:0`. The pin cost doesn't go away, but it moves to worker init, so no request pays for it.

What do you think?

changhyeonnam:
@ApostaC 
I refined the warm-up idea from my previous comment into a concrete design. please take a look.!

To make the trade-off concrete, here are the three variants as timelines. 
The `▓▓▓` block is the foreground pin of the init chunk (`init_size`, 20 GB by default). 
It happens exactly once in every variant; the only question is **where it sits and which GPU gets the CUDA context it creates**.

**(1) original LMCache behavior: pin at server boot (the bug in #4019)**

```
 server start                                          first request
 ────●▓▓▓▓▓▓─────────────────────────────────────────────────●──────▶
     └─ pin at boot: no worker exists yet, so the CUDA context
        lands on the default device cuda:0 ✗
        That context is never used afterwards, pure waste
        (~616 MiB with zero workers connected, see #4019)
```

**(2) this PR: pin on first `allocate()`**

```
 server start                                          first request
 ────●──────────────────────────────────────────────────────●▓▓▓▓▓▓●──▶
     └─ no pin, cuda:0 stays clean ✓                         └─ pin on the
                                                                worker GPU ✓
                                                                first request
                                                                stalls ✗
```

**(3) proposal: pin at worker registration, in the background**

```
 server start         worker registers KV caches       first request
 ────●────────────────────●──────────────────────────────────●──────▶
     └─ no pin,           ├─ worker device known here ✓
        cuda:0 clean ✓    ├─ bg thread: pin ▓▓▓▓▓▓ done
                          └─ overlaps engine init, no request pays ✓
```

Why moving the pin is safe: pinned host memory is usable from every device once registered (UVA); the pin call just needs *some* CUDA context to run under. The server needs a context on the worker's GPU anyway (for IPC / transfer kernels), so pinning under that context makes the extra cost zero, while pinning at boot creates one on cuda:0 that nothing ever uses. The rest of the pool (up to `--l1-size-gb`) is pinned by the background expansion thread in all three variants.

Implementation notes:

- `register_kv_cache` receives the worker's KV tensors, so the device is known at registration, before any request.
- `REGISTER_KV_CACHE` runs on the SYNC lane, so the pin fires on a background thread instead of inline.
- `allocate()` keeps the lazy fallback: worst case equals (2), in practice never hit.

This matches the expected behavior in #4019: no CUDA context at boot, context created on the worker's device only.

If this sounds good, I'll update the PR (tests included: registration pins before first allocate; allocate during in-flight pin blocks then succeeds; repeated registrations pin once).

DongDongJu:
IMO, Its make sense approach.
but plz make sure that we should not assume that server/worker is ready state before finishing pinning memory.

Tptogiar:
goood

changhyeonnam:
@ApostaC @DongDongJu Could you take another look at this PR? I have now validated it end to end on a Kubernetes deployment, and I want to share the actual behavior at each stage.

**Setup.** A B200 node running Dynamo 1.3.0 in disaggregated mode: one vLLM prefill worker and one vLLM decode worker, one GPU each, serving Qwen3-30B-A3B-FP8. The LMCache MP server runs as a separate per node pod. The server image is `lmcache/vllm-openai:nightly-2026-08-13` (lmcache v0.5.4.dev32) with this branch's `lazy_memory_allocator.py` applied, and the workers bundle the same lmcache build.

**Stage 1, server boot with zero workers.** This is the exact reproduction condition of the original issue. With the MP server fully up (ZMQ and HTTP listening), `nvidia-smi` shows no lmcache process on any of the 8 GPUs, and the boot log contains no pinning or device initialization at all. Before this fix, the server already held about 616MiB on `cuda:0` at this point. The engine pod also restarted once during testing and reproduced the same clean boot, so the behavior is consistent.

**Stage 2, workers register.** CUDA contexts appear only at registration time, and only on the GPUs the workers actually occupy:

```
LMCache INFO: Initialized cuda stream on device cuda:0
LMCache INFO: Initialized cuda stream on device cuda:1
```

`nvidia-smi` confirms the server process holds about 756MiB on exactly those two GPUs, and the other six stay untouched. In other words, the pinned pool context now lands where a worker lives instead of defaulting to `cuda:0`.

**Stage 3, under load.** A shared prefix benchmark (1024 prompts, 32 groups) ran through the pool. The deferred expansion thread kicked in on demand (`LazyMemoryAllocator: Expanding` appears in the log once traffic starts), and the cache served real traffic:

```
lmcache_mp_lookup_requested_tokens_total 1507328
lmcache_mp_lookup_hit_tokens_total       1452032   (96.3 percent hit rate)
```

So the full lifecycle of the deferred pinning design holds in practice: no CUDA work in `__init__`, pinning triggered by the first allocation inside a worker device context, and the expansion path working normally under production style load.

One side note: while setting this up I found an unrelated limitation on the Dynamo side (its PD path rejected the LMCacheMP connector at request time) and filed ai-dynamo/dynamo#13258 for it. It does not affect this PR.

Thanks in advance for the re-review!

changhyeonnam:
@ApostaC @DongDongJu 
Status update on the CI failures.

I analyzed the failure logs of buildkite/k3-multiprocess-test and confirmed the failing step is vllm_bench, specifically its throughput gate: 12.09% slower before the warm up commit, 7.43% after, against the 5% allowance. The other steps in the pipeline pass.

My read of the residual gap: with eager pinning the server pins the 80GB pool from boot and has the whole engine startup (about 260s) to finish, while registration time warm up only has the window between KV registration and the LMCache bench (about 70s). Measured cudaHostRegister runs at 1.77GB/s on a B200 node, so 80GB takes about 45s. That fits the window on our hardware but may not on the CI machine, in which case the tail of pinning overlaps the benchmark.

Question for the maintainers: is this gate meant to also measure the pool warm up window, or should the benchmark warm up wait until pinning completes so both variants are compared in the same steady state? If the latter, I am happy to add that wait to the CI script, plus collect the server log as an artifact so the expansion timeline is visible.

PR #4022, opened 2026-07-06
xpu: correct LMCache KV offload on Intel XPU

libinta:
Makes LMCache's local_cpu KV offload work correctly. Two problems on the XPU path are addressed:
a DEVICE_LOST crash, and silent KV corruption on partial chunks. Adds benchmark coverage for the SGLang XPU
connector.
Change:
1. Correct partial-chunk KV handling on XPU
Two coupled failure modes on the KV-transfer staging guard:
- **Corruption:** staging a *partial* chunk through `gpu_buffer[:, :, :end-start, :]`
  produces a non-contiguous view, but the SYCL kernel reads the base pointer as
  tightly packed `[kv, num_layers, end-start, hidden]` — scattering every layer
  to the wrong offset and silently corrupting retrieved KV.
- **Crash:** sending a partial chunk down the direct kernel path against
  *pageable* host memory is illegal under Level Zero → `UR_RESULT_ERROR_DEVICE_LOST`,
  which kills the device for the rest of the process.
Fix: partial chunks take the **direct path** (guard `end - start != gpu_buffer.shape[2]`),
which is now safe because the host buffer is USM-pinned (change 1) — a compute
kernel is permitted to write to USM host memory, so no DEVICE_LOST. Only chunks
larger than the buffer stage through `gpu_buffer`.
Verified on Arc Pro B70: `test_vllm_paged_connector_v2_with_gpu_and_mla` fails
with the staging guard, passes with the direct-path guard, no DEVICE_LOST under
`use_gpu=True`.

2. Benchmarks
- New `tests/benchmarks/test_xpu_sglang_connector_benchmark.py`: pytest-benchmark
  suite for `SGLangXPUConnector` mirroring the vLLM `test_xpu_connector_benchmark.py`
  (same axes, workload sizes, group names) — store-v2 / retrieve-v2 / lookup-v2,
  parametrized over backend (cpu/disk), `use_gpu`, `save_unfull_chunk`.
- Adds `_skip_if_xpu_without_gpu_buffer` to the XPU benchmarks: `use_gpu=False`
  exercises an unsupported Level Zero host-pointer kernel path (DEVICE_LOST that
  cascades all subsequent tests); `use_gpu=True` passes at ~35 GB/s.
Test: On Arc Pro B70 : `use_gpu=True` store/retrieve/lookup benchmarks pass at
~35 GB/s; `test_vllm_paged_connector_v2_with_gpu_and_mla` passes across kv
formats; `use_gpu=False` cases skipped (unsupported Level Zero host-pointer path).


csrc/sycl/mem_kernels_sycl.cpp
```diff
@@ -1021,3 +1026,23 @@ void lmcache_memcpy_async(uintptr_t dest, uintptr_t src, size_t nbytes,
...
 }
+
+// ---------------------------------------------------------------------------
+// Pinned host allocation (SYCL/XPU analog of the CUDA cudaHostAlloc path in
+// csrc/mem_alloc.cpp). LMCache local_cpu backend expects a device-accessible
+// (USM host) buffer; without this the XPU build silently fell back to pageable
+// host memory, dropping D2H store throughput ~20x. sycl::malloc_host bound to
+// the current XPU device context gives true USM-pinned host memory.
+// ---------------------------------------------------------------------------
+uintptr_t alloc_pinned_ptr_xpu(size_t size, int device_index) {
```
hlin99: is the right way to replace alloc_pinned_ptr?

csrc/sycl/mem_kernels_sycl.cpp
```diff
@@ -1021,3 +1026,23 @@ void lmcache_memcpy_async(uintptr_t dest, uintptr_t src, size_t nbytes,
...
+  sycl::queue& q = c10::xpu::getCurrentXPUStream(device_index).queue();
+  void* ptr = sycl::malloc_host(size, q.get_context());
```
hlin99: qq: is passing context the mechanism that can make the buffer pinned?

lmcache/v1/memory_management.py
```diff
@@ -597,6 +536,39 @@ def _allocate_cpu_memory(
     if size == 0:
         return torch.empty(0, dtype=torch.uint8)
 
+    if torch_device_type == "xpu":
```
hlin99: device specific code is not expected to be here.

hlin99 (commented):
left some highlevel questions...

hlin99:
please also fix DCO, code quality issues

tests/benchmarks/test_xpu_sglang_connector_benchmark.py
```diff
@@ -0,0 +1,539 @@
...
+        num_layers=num_layers,
+        use_gpu=use_gpu,
+        chunk_size=chunk_size,
+        dtype=dtype,
+        device=device,
+        use_mla=use_mla,
+    )
+
+
+def _kv_device(kvcaches) -> torch.device:
```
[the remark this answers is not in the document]
libinta: Done

lmcache/v1/memory_management.py
libinta: removed

csrc/sycl/mem_kernels_sycl.cpp
libinta: sycl::malloc_host is the USM pinned-host allocator, so calling it is what makes the buffer pinned/device-accessible. The context argument is mandatory in every malloc_host overload (there's no context-less one). updated the comments also

csrc/sycl/mem_kernels_sycl.cpp
libinta: based on your comments, changed to use alloc_pinned_ptr similar to other kernels under the same name.

libinta:
> please also fix DCO, code quality issues

done. thanks for review

hlin99:
it seems this PR addresses two problems, 1 is alloc pin memory by sycl, 2 is some logic change on memory management and sglang.

to make review faster, it would be nice to split the PR into 2 PRs. and i guess the logical change part may take some time

DongDongJu:
> it seems this PR addresses two problems, 1 is alloc pin memory by sycl, 2 is some logic change on memory management and sglang.
> 
> to make review faster, it would be nice to split the PR into 2 PRs. and i guess the logical change part may take some time

I do agree.

libinta:
> > it seems this PR addresses two problems, 1 is alloc pin memory by sycl, 2 is some logic change on memory management and sglang.
> > to make review faster, it would be nice to split the PR into 2 PRs. and i guess the logical change part may take some time
> 
> I do agree.

@DongDongJu  Please check, pr is split, new pr https://github.com/LMCache/LMCache/pull/4129

libinta:
> it seems this PR addresses two problems, 1 is alloc pin memory by sycl, 2 is some logic change on memory management and sglang.
> 
> to make review faster, it would be nice to split the PR into 2 PRs. and i guess the logical change part may take some time

@hlin99 Please check, pr is split, new pr https://github.com/LMCache/LMCache/pull/4129,thanks

libinta:
@hlin99  can you review this again? thanks

PR #4023, opened 2026-07-06
feat(valkey): Tag GLIDE connections as GlidePySync(lmcache:<ver>)

omerrubi-amzn:
## Summary

Attributes LMCache's valkey-glide connections in `CLIENT INFO` / `CLIENT LIST` by setting GLIDE's `client_info_tag` client-config option, producing a library name of `GlidePySync(lmcache:<version>)`. This makes LMCache usage distinguishable on the server side (adoption tracking, observability, debugging) **without** changing GLIDE's own library identity — GLIDE-level adoption metrics keep counting these connections, with a per-framework breakdown on top.

Example resulting `lib-name`: `GlidePySync(lmcache:0.3.9)`.

## Motivation

`client_info_tag` was added to the glide Python client in [valkey-io/valkey-glide#6389](https://github.com/valkey-io/valkey-glide/pull/6389) and ships in **valkey-glide 2.5.0**. It preserves the underlying `GlidePy`/`GlidePySync` token while appending a parenthesized framework tag (following the ioredis `ioredis(<tag>)` convention), which is the recommended way for a wrapping layer like LMCache to identify itself.

## What changed

`lmcache/v1/storage_backend/connector/valkey_connector.py`:

- **`_lmcache_client_info_tag()`** — builds `lmcache:<lmcache_version>`. Falls back to a bare `lmcache` if the resolved version string contains whitespace (GLIDE's config validation rejects whitespace tags).
- **`_glide_config_supports_client_info_tag(config_cls)`** — feature-detects support from the client-config constructor signature.
- **`_ThreadWorkerPool._get_client`** — passes `client_info_tag` through in **both** the standalone (`GlideClientConfiguration`) and cluster (`GlideClusterClientConfiguration`) branches, **only** when supported.

### Version gating (important)

`client_info_tag` is not in any released wheel at the time of writing, and older glide clients raise `TypeError` on the unknown kwarg. Rather than compare version strings (brittle across custom/sync builds), the connector inspects the config constructor signature and applies the tag **only** when the `client_info_tag` parameter is present. On current glide builds, behavior is unchanged.

## Testing

New `tests/v1/storage_backend/test_valkey_client_info_tag.py` (injects a fake `glide_sync`, no server needed):

- Tag formatting (`lmcache:<version>`, no whitespace) and the whitespace fallback.
- Signature-based support detection (present / absent / uninspectable).
- `_get_client` **sets** the tag when supported and **omits** it when not — for both standalone and cluster.
- Standalone still forwards `database_id` alongside the new tag.

Local run: `ruff check` + `ruff format --check` clean; existing `test_valkey_connector.py` + the new tests pass (62 passed).


sammshen (approved):
LGTM

DongDongJu (approved):
LGTM

PR #4024, opened 2026-07-06
[Operator] CacheBlend webhook: support namespace-qualified engine refs for cross-namespace binds

aeon-x:
## Summary

The `lmcache.ai/cacheblend-engine` annotation now accepts a namespace-qualified
value `"<namespace>/<name>"`, letting a vLLM pod bind a CacheBlendEngine that
lives in another namespace. A bare name still resolves in the pod's own
namespace, so existing deployments are unaffected.

## Motivation

A single shared CacheBlendEngine (e.g. in a dedicated cache namespace) should be
able to serve vLLM pods across app namespaces. For `LMCacheEngine` this already
works without the webhook: controllers that render pod specs directly (e.g.
VLLMRuntime) read the `<engine>-connection` ConfigMap from any namespace and
inject `--kv-transfer-config` themselves.

CacheBlend has no such path: its wiring (payload init container, plugin volume,
PYTHONPATH, hostIPC, pull secrets) is applied only by the mutating webhook, so
the annotation is the *only* way to bind — and it could only name an engine in
the pod's own namespace. This PR makes the webhook itself resolve the engine's
namespace from the annotation.

## Changes

- `cacheblend_pod_injector.go`
  - New `parseEngineRef(ref, defaultNS)` splits `"<namespace>/<name>"`; a bare
    name (or blank namespace, e.g. `"/name"`) falls back to the pod's namespace.
  - The engine CR and its `<engine>-connection` ConfigMap are resolved in the
    **engine's** namespace instead of the pod's.
  - Annotation doc comment updated.
- `cacheblend_pod_injector_envtest_test.go`
  - New envtest: a pod in a different namespace binds `<engine-ns>/<engine>`
    and is fully injected — hostIPC, plugin volume, init container, and
    `--kv-transfer-config` pointing at `tcp://<engine>.<engine-ns>.svc`.

## Behavior

| Annotation value | Engine resolved in |
|---|---|
| `my-engine` | pod's namespace (unchanged) |
| `cache-system/my-engine` | `cache-system` |
| `/my-engine` | pod's namespace (fallback) |

Cross-namespace binds require no engine-side configuration. Injection remains
fail-open: a ref to a non-existent engine/namespace skips with the existing
`engine-not-found` stamp. The webhook already reads engines and ConfigMaps
cluster-wide under the manager ServiceAccount, so no RBAC changes are needed.

## Testing

- `go build ./...`, `go vet ./...` clean
- Full operator suite passes, including the webhook envtest suite with the new
  cross-namespace spec


royyhuang (commented):
LGTM

royyhuang (approved):

ApostaC (approved):
LGTM!

PR #4027, opened 2026-07-07
[MP coordinator] Chunk eviction dispatch to the delete cap; add default quota limit for unquota'd salts

aeon-x:
## Summary

Two coupled fixes to the coordinator's L2 eviction loop, both hit in real
deployments:

1. **Chunk eviction dispatch to the MP server's delete cap.** The MP endpoint
   rejects any single `DELETE /cache/objects` with more than `MAX_DELETE_BATCH`
   keys with HTTP 400. A full-salt eviction (e.g. a quota dropped to 0)
   routinely exceeds that cap, so the whole dispatch failed and the keys were
   never evicted. `execute_evictions` now splits the plan into chunks of at
   most `MAX_DELETE_BATCH` (imported from the MP server's `object_service`, so
   the two sides can't drift) and dispatches each chunk as its own request.

2. **Default quota limit — a safe-by-default signal that arms eviction of
   unquota'd data.** Quotas live in memory, so a (re)started coordinator has
   an empty quota table until the external quota controller re-syncs it.
   Previously, unregistered salts had an implicit 0-byte limit (allowlist
   semantics), so an eviction sweep in that window planned a full eviction of
   every tenant. Unregistered salts are now governed by a configurable
   **default limit** (`GET`/`PUT /quota/config`):

   - `null` (boot default) — unquota'd salts are **exempt** from eviction; a
     cold quota table cannot mass-evict unknown tenants.
   - `0` — strict allowlist: all bytes under unquota'd salts become evictable
     next cycle. The controller sets this **after** re-registering every
     per-salt quota — it is the explicit signal that arms fleet-wide eviction.
   - `> 0` — every unquota'd salt gets that byte budget.

   Explicit per-salt quotas are enforced as soon as they are registered,
   regardless of the default. The MP server's local eviction controller keeps
   the legacy `get_limit_bytes` path (unregistered ⇒ 0) and is unaffected.

The two changes motivate each other: flipping the default to `0` is exactly
the event that produces the over-cap eviction plans that chunking handles.

## ⚠️ Behavior change

Out of the box, a coordinator previously evicted unquota'd salts
automatically. After this PR, eviction of unquota'd data is **inert until
armed**: deployments without an external quota controller must issue

```bash
curl -X PUT http://<coordinator>:9300/quota/config \
    -H 'Content-Type: application/json' -d '{"default_limit_gb": 0}'
```

once after startup to restore the previous allowlist behavior.

## Controller flow (after coordinator restart)

```bash
# 1. re-register every tenant quota
PUT /quota/{salt} {"limit_gb": N}     # one per tenant
# 2. arm eviction of everything else
PUT /quota/config {"default_limit_gb": 0}
```

## Changes

- `distributed/quota_manager.py` — new `set_default_limit_bytes` /
  `get_default_limit_bytes` / `effective_limit_bytes` (explicit entry wins;
  unregistered salts resolve to the default, `None` = exempt).
  `get_limit_bytes` is untouched.
- `cache_control/eviction_manager.py` — `compute_eviction_plan` uses
  `effective_limit_bytes` and skips salts whose limit is `None`; dispatch loop
  chunks the plan to `MAX_DELETE_BATCH`, each chunk tracked in
  `_in_flight_dispatches`.
- `http_apis/quota_api.py` + `schemas.py` — `GET`/`PUT /quota/config`
  (declared before `/quota/{cache_salt}` so `config` is not captured as a
  salt).
- Docs: design doc (`l2_usage_and_eviction.md`) and coordinator guide
  (endpoint reference, controller flow).

## Tests

- `test_quota_manager.py` — default-limit matrix: unset/0/positive, explicit
  quota wins, explicit 0 while default unset, reset to `None`, negative
  rejected.
- `test_eviction_manager.py` — exempt boot state, explicit-quota-while-unset,
  arming flip (`None` → `0`), positive default as budget; chunking test
  (5 keys, cap 2 → batches 2/2/1, all within cap).
- `test_quota_api.py` — `/quota/config` round-trips (null/0/positive),
  negative rejected, route-shadowing check (`config` never becomes a salt),
  controller-flow test.
- `test_cache_api.py` — pin/unpin test arms enforcement via the real endpoint.


lmcache/v1/mp_coordinator/app.py
```diff
@@ -111,7 +111,13 @@ async def _health_loop() -> None:
...
     async def _eviction_loop(http_client: httpx.AsyncClient) -> None:
         """Periodically check usage against quotas and dispatch
-        eviction RPCs to any one registered MP server."""
+        eviction RPCs to any one registered MP server.
+
+        Waits ``eviction_startup_delay`` seconds before the first sweep so
+        the startup resync can backfill trackers first (avoids evicting
+        against cold, incomplete state right after boot)."""
+        if config.eviction_startup_delay > 0:
+            await asyncio.sleep(config.eviction_startup_delay)
```
ApostaC: Can we use an `asyncio.Event` to do the synchronization so that the eviction runs after resync?
(In this case, we don't need the `eviction_startup_delay`)

ApostaC (commented):
One quick comment. PTAL!

lmcache/v1/mp_coordinator/app.py
aeon-x: @ApostaC  @ruizhang0101  
Right now, in our inference usecase, we need to resync the per tenant quota to coordinator periodically (eg. every 10 seconds). 

Since we are not persisting the quota information in coordinator now, the `eviction_startup_delay` provides a sufficient buffer time to set quota before any eviction loop starts.

ApostaC: Yeah, but I think using events to synchronize is a cleaner and more evolvable solution. Maybe we will add more complicated logic in the future, and the `eviction_startup_delay` may need extra human effort to configure.

aeon-x: Updated logic. A external api call will be required to set the default limit to start eviction for unquota salt. 

Boot: 
default_limit_gb: None 
No eviction for unquota salt 

From outside: 
PUT /quota/salt-a {"limit": 50}
PUT /quota/salt-b {"limit": 100 }
PUT /quota/config {"default_limit_gb": 0} 

All unquota salt's kvcache will be evicted as the default_limit_gb is 0.

lmcache/v1/distributed/quota_manager.py
ApostaC, line 1: Will the lmcache server own an instance of the quota manager? If so, will there be cases that the lmcache server's quota manager conflicts with the coordinator's quota manager?

lmcache/v1/mp_coordinator/cache_control/eviction_manager.py
```diff
@@ -101,9 +104,14 @@ def compute_eviction_plan(self) -> dict[str, list[ObjectKey]]:
...
-        eviction. Pinned keys are always excluded. Pure — no network
-        calls, no state mutation.
+        their LRU keys; a quota of ``0`` means full eviction. Salts
+        without an explicit quota use the registry's default limit
+        (``QuotaManager.effective_limit_bytes``): until the external
+        quota controller sets one (``PUT /quota/config``), the default
+        is ``None`` and unquota'd salts are **exempt** — so a freshly
+        restarted coordinator with an empty quota table cannot
+        mass-evict unknown tenants. Pinned keys are always excluded.
+        Pure — no network calls, no state mutation.
```
ApostaC: 2 nits:
- remove the sentences starting from `the default is 'None' and unquota'd...`
- just say "the coordinator will not start evicting unquota'd salts" after "until the external quota controller sets...."

ApostaC (commented):
Otherwise LGTM!

lmcache/v1/distributed/quota_manager.py
aeon-x: Yes, the lmcache server's quota manager and coordinator's quota manager will have different quota data. 

Only one will be the source of truth. Updated doc to note on this.

lmcache/v1/mp_coordinator/cache_control/eviction_manager.py
aeon-x: updated

ApostaC (approved):
LGTM!

deng451e (approved):

Issue #3957, opened 2026-06-30
[Bug] lmcache_mp_connector_0201.py (vLLM 0.20.1) out of sync with refactored adapter API — ParallelStrategy/SchedulerAdapter signature mismatch


d3nb:
/kind bug

## Summary

On **vLLM 0.20.1 + LMCache 0.5.0**, `LMCacheMPConnector` cannot initialize because **`lmcache/integration/vllm/lmcache_mp_connector_0201.py` is out of sync with the refactored adapter API**. It constructs `ParallelStrategy` and `LMCacheMPSchedulerAdapter`/`LMCacheMPWorkerAdapter` with the *old* (pre-multi-server) signatures.

The generic `lmcache_mp_connector.py` *does* match the current adapter API (`server_urls`, the new `ParallelStrategy` fields), but it imports `KVCacheSpecKind` from `vllm.v1.kv_cache_interface`, which only exists in **vLLM 0.23.0+**, so it cannot be imported on 0.20.1.

Net result: **there is no working `LMCacheMPConnector` path on vLLM 0.20.1** — the version-specific `_0201` connector is stale, and the generic one requires a newer vLLM.

## Environment

- **vLLM:** v0.20.1
- **LMCache:** 0.5.0 (PyPI wheel; `c_ops` rebuilt from sdist against the image's CUDA 13 torch — note: the prebuilt cp312 wheel's `c_ops` links `libcudart.so.12` and falls back to the python backend on a CUDA 13 torch image)
- **Python:** 3.12
- **Model:** `deepseek-ai/DeepSeek-V4-Flash`, FP8 KV, DP=4 + EP, `--block-size 256`
- **Hardware:** NVIDIA B300
- Connector selected via `--kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_connector_module_path":"lmcache.integration.vllm.lmcache_mp_connector_0201","kv_role":"kv_both","kv_connector_extra_config":{"lmcache.mp.host":"tcp://...","lmcache.mp.port":18009}}'`

## Errors (two cascading)

**(1)** `lmcache_mp_connector_0201.py:156` (`create_worker_adapter`) / `:123` (`create_scheduler_adapter`):

```
TypeError: ParallelStrategy.__init__() takes 7 positional arguments but 8 were given
```

[...]

PR #4028, opened 2026-07-07
fix(integration/vllm): guard state transition for non-chosen connector

Alex-ai-future:

Syncs the state transition fix from [vLLM PR #47505](https://github.com/vllm-project/vllm/pull/47505) to LMCache's three `lmcache_mp_connector` variants.

**Bug**: In `update_state_after_alloc`, the condition `tracker.needs_retrieve()` only checks `num_lmcache_hit > num_vllm_hit` — it does not check whether this connector was the chosen one in a MultiConnector. When a non-chosen connector receives `num_external_tokens=0`:

| Consequence | Impact |
|---|---|
| State enters `WAITING_FOR_LOAD` instead of `READY` | State machine corruption |
| `free_lookup_locks` not called → read locks held until TTL (10 min) | Lock leak |
| `allocated_block_ids` populated → retrieve task created (after vLLM #46865) | Unwanted KV transfer |

**Fix** (one line per file):
```python
# Before
condition = tracker.needs_retrieve()
# After
condition = tracker.needs_retrieve() and num_external_tokens > 0
```

When `condition=False`, the existing `else` branch correctly sets state → `READY`, frees all locks, and calls `cleanup_lookup_result`.

**Files changed** (3 variants for different vLLM versions):
- `lmcache/integration/vllm/lmcache_mp_connector.py` (latest)
- `lmcache/integration/vllm/lmcache_mp_connector_0201.py` (vLLM 0.20.1)
- `lmcache/integration/vllm/lmcache_mp_connector_0180.py` (vLLM 0.18.0)

**Version compatibility verification**:

MultiConnector was introduced in vLLM PR #17564 (2025-05-14), before all three targeted versions. Verified that all three vLLM versions have identical `MultiConnector.update_state_after_alloc` behavior — non-chosen connectors receive `(empty_blocks, 0)`:

| vLLM version | Release date | MultiConnector present | Non-chosen receives `num_external_tokens=0` | `needs_retrieve()` logic |
|---|---|---|---|---|
| v0.18.0 | 2026-03-20 | ✅ | ✅ `c.update_state_after_alloc(request, empty_blocks, 0)` | `num_lmcache_hit_blocks > num_vllm_hit_blocks` |
| v0.20.1 | 2026-05-04 | ✅ | ✅ `c.update_state_after_alloc(request, empty_blocks, 0)` | `num_lmcache_hit_blocks > num_vllm_hit_blocks` |
| main | — | ✅ | ✅ `c.update_state_after_alloc(request, empty_blocks, 0)` | `num_lmcache_hit_tokens > num_vllm_hit_tokens` |

The bug exists in all three versions, and the fix applies identically to each.

- This is the LMCache-side sync of vLLM PR #47505 (already submitted, reviews the identical logic)
- LMCache maintains three connector copies for different vLLM API versions; all three have the same bug
- This fix is a prerequisite for safely merging vLLM PR #46865 (MultiConnector passes real blocks to non-chosen connectors)
- Refs vllm-project/vllm#47505, vllm-project/vllm#46865

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -1068,7 +1068,10 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
             tracker.append_block_ids(tuple(new_block_ids))
 
         # Update the state of the tracker
-        condition = tracker.needs_retrieve()
+        # Skip retrieve when this connector is not the chosen one in a
+        # MultiConnector (num_external_tokens == 0). Otherwise the state
+        # would erroneously enter WAITING_FOR_LOAD and leak lookup locks.
+        condition = tracker.needs_retrieve() and num_external_tokens > 0
         if tracker.state == LMCacheMPRequestState.PREFETCHING:
             # If need to retrieve, change to WAITING_FOR_LOAD
             # Otherwise, change to READY
--- a/lmcache/integration/vllm/lmcache_mp_connector_0180.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0180.py
@@ -805,7 +805,10 @@ class LMCacheMPConnector(KVConnectorBase_V1):
             tracker.append_block_ids(new_block_ids)
 
         # Update the state of the tracker
-        condition = tracker.needs_retrieve()
+        # Skip retrieve when this connector is not the chosen one in a
+        # MultiConnector (num_external_tokens == 0). Otherwise the state
+        # would erroneously enter WAITING_FOR_LOAD and leak lookup locks.
+        condition = tracker.needs_retrieve() and num_external_tokens > 0
         if tracker.state == LMCacheMPRequestState.PREFETCHING:
             # If need to retrieve, change to WAITING_FOR_LOAD
             # Otherwise, change to READY
--- a/lmcache/integration/vllm/lmcache_mp_connector_0180.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0180.py
@@ -852,7 +852,10 @@ class LMCacheMPConnector(KVConnectorBase_V1):
             tracker.append_block_ids(new_block_ids)
 
         # Update the state of the tracker
-        condition = tracker.needs_retrieve()
+        # Skip retrieve when this connector is not the chosen one in a
+        # MultiConnector (num_external_tokens == 0). Otherwise the state
+        # would erroneously enter WAITING_FOR_LOAD and leak lookup locks.
+        condition = tracker.needs_retrieve() and num_external_tokens > 0
         if tracker.state == LMCacheMPRequestState.PREFETCHING:
             # If need to retrieve, change to WAITING_FOR_LOAD
             # Otherwise, change to READY
--- a/lmcache/integration/vllm/lmcache_mp_connector_0201.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0201.py
@@ -846,7 +846,10 @@ class LMCacheMPConnector(KVConnectorBase_V1):
             tracker.append_block_ids(new_block_ids)
 
         # Update the state of the tracker
-        condition = tracker.needs_retrieve()
+        # Skip retrieve when this connector is not the chosen one in a
+        # MultiConnector (num_external_tokens == 0). Otherwise the state
+        # would erroneously enter WAITING_FOR_LOAD and leak lookup locks.
+        condition = tracker.needs_retrieve() and num_external_tokens > 0
         if tracker.state == LMCacheMPRequestState.PREFETCHING:
             # If need to retrieve, change to WAITING_FOR_LOAD
             # Otherwise, change to READY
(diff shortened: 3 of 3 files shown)


d3nb:
This shouldn't age out — it is the only remaining carrier for half of a fix that is already merged everywhere else, and I think it needs just a small rebase to be trivially mergeable.

**Where things stand.** #4781 landed the same guard on 2026-08-27, but only in `lmcache_mp_connector.py`:

```diff
-        condition = tracker.needs_retrieve()
+        condition = num_external_tokens > 0 and tracker.needs_retrieve()
```

This PR makes the functionally identical change in that file (plus an explanatory comment), and — the part nothing else covers — carries it into `lmcache_mp_connector_0180.py` and `lmcache_mp_connector_0201.py`.

**So the useful shape now is smaller than when this was opened.** The `lmcache_mp_connector.py` hunk is redundant with #4781 and will conflict: the branch predates it, and the hunk header has moved from line 1068 to 953. The two snapshot hunks are untouched by anything merged so far. A rebase that drops the first hunk and keeps the other two should be conflict-free and reviewable in a minute.

**What closing it costs.** `_0180` and `_0201` would be left permanently without a guard that the live connector, vLLM (#47505), and the LMCache backport all have. Anyone pinned to vLLM 0.18.x / 0.20.1 and using MultiConnector keeps hitting the non-chosen-connector state leak with no path forward. That drift — a fix landing in the live connector while the version snapshots silently fall behind — is exactly what I reported in #3957, and this PR is the concrete remedy for one instance of it.

@Alex-ai-future, would you be up for that rebase? Happy to test the result against our vLLM 0.20.1 + MultiConnector deployment, or to open a rebased version myself if you'd rather not carry it.

Alex-ai-future:
Thanks @d3nb for the detailed guidance.

I rebased this PR onto the latest `dev` and pushed the update.

Following your suggestion:
- Removed the now-redundant change in `lmcache_mp_connector.py`, which is already covered by #4781.
- Kept the guard in `lmcache_mp_connector_0180.py` and `lmcache_mp_connector_0201.py` for the legacy vLLM compatibility paths.

The branch should now be conflict-free and the PR is ready for another review. Local Ruff, Python syntax, and diff checks pass; the targeted pytest could not start because this environment is missing `numpy`.

d3nb:
That's exactly the shape I had in mind — `+8/-2` across the two snapshots, no overlap with #4781. Reviewable by inspection in a minute: it's the same one-line condition change already merged for the live connector, applied to the two legacy paths.

One thing I should be straight about, since I offered to verify it: **I can't give you a meaningful end-to-end test of the `_0201` path, and it's worth knowing why.**

`lmcache_mp_connector_0201.py` currently can't initialize at all on vLLM 0.20.1 — it constructs `ParallelStrategy` and the scheduler/worker adapters with the pre-multi-server signatures, which no longer match the refactored adapter API. That's #3957. So any run I attempt there fails before it ever reaches the state machine this guard touches. Testing it would tell us nothing about this PR.

That does not argue against merging — the guard is correct and the snapshots should not be left behind the connector they were copied from. But it does mean the two stack rather than substitute:

- **This PR** fixes the non-chosen-connector state leak in `_0180` / `_0201`.
- **#3957** is the reason `_0201` still won't start on 0.20.1 afterwards.

Worth flagging so merging this isn't read as "0.20.1 + MultiConnector works now" — it closes one of the two doors, and the other is still shut.

For what it's worth, on `_0180` I have no way to test either (we're not pinned to 0.18.x), so I'd treat both hunks as inspection-reviewed rather than waiting on anyone's runtime verification. Your missing-`numpy` problem and my `_0201` problem both point the same way: this one is small and mechanical enough that a reviewer's eyes are the right bar.

Alex-ai-future:
@deng451e Could you please take a look at the rebased version? The remaining change is limited to the two legacy connector snapshots and mirrors the guard already merged in #4781.

d3nb:
Correction to my comment above from 2026-09-06 — I stated something that was wrong, and it's relevant to how this PR gets reviewed.

I wrote that `lmcache_mp_connector_0201.py` "can't initialize at all on vLLM 0.20.1" because it still used the pre-multi-server adapter signatures (#3957). That was already out of date when I wrote it: **#4135, merged on 2026-07-20, realigned `_0201.py` with the current adapter API** (it was filed against #4134, a duplicate of #3957, which is why I missed it). Checking `dev` today, `ParallelStrategy` is built with keyword arguments including `n_servers`, the scheduler adapter gets `server_urls=[...]`, and the worker adapter's single `server_url` matches its current signature. I should have checked before asserting it. Sorry for the noise.

So please disregard the "these two stack" part of that comment. #3957 is resolved by #4135 (I'm closing it), and the guard in this PR is the only remaining difference between `_0201` and the live connector for this path.

One related thing worth clarifying before anyone acts on it: #3798 was closed on 2026-09-15 with a note that `_0201.py` had been removed. As of today, though, both `lmcache_mp_connector_0201.py` and `lmcache_mp_connector_0180.py` are still present on `dev` and on `main`. If they're about to be deleted, this PR can simply be closed along with them; if they're staying, the guard here is still needed so they don't lag behind #4781. Either answer is fine — I just didn't want this closed on the assumption that the files are already gone.

PR #4033, opened 2026-07-07
[CI/CD] add preemption check for LMCache/Engine driven paths

hlin99:
## Motivation

Closes a testing gap: Currently, our testing suite lacks end-to-end (E2E) coverage for preemption behaviors. This PR introduces a dedicated CI/CD check to resolve this blind spot.

**Comprehensive path coverage:** The new E2E tests explicitly validate preemption handling across both the LMCache-driven and Engine-driven paths within Multi-Processing (MP) mode.

## Test Methodology & Verification

- **Trigger Mechanism:** Preemption is actively triggered during an lm-eval run by intentionally tuning the GPU memory utilization parameters.
- **Success Criteria:** The test suite executes two consecutive runs under these preemptive conditions. The implementation is considered successful and stable if the evaluation score variance between both runs stays strictly within 1%.


.buildkite/k3_tests/multiprocess/pipeline.yml
```diff
@@ -124,6 +124,36 @@ steps:
...
+          esac
+          .buildkite/k3_tests/multiprocess/run.sh lm_eval_preemption
+        matrix:
+          setup:
+            mode:
+              - "lmcache_driven"
+              - "engine_driven"
+        timeout_in_minutes: 30
+        env:
+          GPU_MEMORY_UTILIZATION: "0.4"
```
maobaolong, line 149: It seems mismatch with the above comment.

.buildkite/k3_tests/multiprocess/pipeline.yml
```diff
@@ -124,6 +124,36 @@ steps:
                   - { name: dshm, emptyDir: { medium: Memory, sizeLimit: 128Gi } }
         artifact_paths: ["*.log"]
 
+      # Preemption correctness: GPU_MEMORY_UTILIZATION=0.5 forces the KV cache
```
maobaolong: Is this dependent the GPU memory size?

.buildkite/k3_tests/multiprocess/pipeline.yml
```diff
@@ -124,6 +124,36 @@ steps:
...
+        command: |
+          case "{{matrix.mode}}" in
+            lmcache_driven) ;;
+            engine_driven)
+              export LMCACHE_MP_TRANSFER_MODE="engine_driven"
+              export L1_USE_LAZY="false"
+              ;;
+          esac
+          .buildkite/k3_tests/multiprocess/run.sh lm_eval_preemption
+        matrix:
```
maobaolong, line 142: I like this trick😄

maobaolong (approved):
LGTM. Discussed with @hlin99 offline, we can do some refactor to reduce duplicated code, but it can be done in follow-up PR or this PR.

.buildkite/k3_tests/multiprocess/pipeline.yml
hlin99: yes. depend on model, gpu memory size, bs, etc. so this util value is for current environment. any change on HW/model may affect preemption. that's why I also check the log whether preemption really happens or not.

.buildkite/k3_tests/multiprocess/pipeline.yml
hlin99: good catch. I adjusted this value several times to get a balanced one between preemption frequency and case execution time. so the comment update is missed

hlin99: fixed by [176b975](https://github.com/LMCache/LMCache/pull/4033/commits/176b975b6b39c0a7427a4701c4e50c7169d86a3d)

hlin99:
> LGTM. Discussed with @hlin99 offline, we can do some refactor to reduce duplicated code, but it can be done in follow-up PR or this PR.

hi @maobaolong . good suggestion to combine bash script. already done in [a308552](https://github.com/LMCache/LMCache/pull/4033/commits/a308552d2bc4f529546067ca9fe667ecf35e0796). would u like to double confirm? thx

maobaolong:
@hlin99 Thanks for did the refactor in this PR. LGTM again!

ApostaC (approved):
LGTM!

PR #4034, opened 2026-07-07
[Fix][FS Native] Fix fs native object group key parsing

nayeonikim:

Fixes fs_native filesystem key parsing for ObjectKey values that include `object_group_id`.

The Python fs_native adapter serializes keys as:

- unsalted: `model@kv_rank@object_group_id@chunk_hash`
- salted: `model@kv_rank@object_group_id@chunk_hash@cache_salt`

The C++ `FSConnector` still parsed the older 3/4-field format, so salted fs_native stores failed before creating the expected file. This PR updates the C++ parser to accept the current 4/5-field format and preserve `object_group_id` in generated filenames, matching the Python fs_l2 filename layout.

Added a native-backed regression test that exercises `NativeConnectorL2Adapter + LMCacheFSClient` instead of only mocking the native connector. The new test verifies both unsalted and salted fs_native stores create filenames with `object_group_id`.

Tested with:

- `NO_GPU_EXT=1 python -m pip install -e . --no-build-isolation`
- `python -m pytest -xvs tests/v1/distributed/test_fs_native_connector_keys.py`
- `python -m pytest -xvs tests/v1/distributed/test_native_connector_l2_adapter.py::TestObjectKeySerialization tests/v1/distributed/test_fs_l2_adapter_keys.py`
- `python -m ruff check --no-cache tests/v1/distributed/test_fs_native_connector_keys.py`
- `python -m ruff format --check tests/v1/distributed/test_fs_native_connector_keys.py`
- `clang-format --dry-run --Werror csrc/storage_backends/fs/connector.cpp csrc/storage_backends/fs/connector.h`
- `git diff --check`

**If applicable**:

### DIFF
--- a/csrc/storage_backends/fs/connector.cpp
+++ b/csrc/storage_backends/fs/connector.cpp
@@ -27,22 +27,25 @@ std::string FSConnector::replace_all(const std::string& str,
 
 std::string FSConnector::key_to_filename(const std::string& key) {
   // Input key format (from _object_key_to_string):
-  //   Unsalted: <model_name>@<kv_rank_hex>@<chunk_hash_hex>
-  //   Salted  : <model_name>@<kv_rank_hex>@<chunk_hash_hex>@<cache_salt>
+  //   Unsalted:
+  //   <model_name>@<kv_rank_hex>@<object_group_id_hex>@<chunk_hash_hex>
+  //   Salted  :
+  //   <model_name>@<kv_rank_hex>@<object_group_id_hex>@
+  //   <chunk_hash_hex>@<cache_salt>
   //
   // Output filename (matching fs_l2_adapter.py._object_key_to_filename):
-  //   Unsalted: <model_name_safe>@0x<kv_rank_hex>@<chunk_hash_hex>.data
+  //   Unsalted:
+  //   <model_name_safe>@0x<kv_rank_hex>@<object_group_id_hex>@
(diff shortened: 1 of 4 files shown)


DongDongJu (changes requested):
Hello @nayeonikim,
Thanks for the fix!
The fix looks correct that the native parser now matches the Python ObjectKey serialization and fs_l2 filename layout, and the native-backed salted regression test covers the actual failure.

Please fix the ut issue after that LGTM.

nayeonikim:
> Hello @nayeonikim,
> Thanks for the fix!
> The fix looks correct that the native parser now matches the Python ObjectKey serialization and fs_l2 filename layout, and the native-backed salted regression test covers the actual failure.
> 
> Please fix the ut issue after that LGTM.

Thanks for the review! @DongDongJu 

The UT failure came from the FS-native O_DIRECT tests added after the original PR version. They were still using the legacy 3-field key format, so after rebasing onto the latest dev, they failed during key parsing before reaching the intended O_DIRECT check.

I’ve updated those test keys to the current 4/5-field format including object_group_id, and the CI is now green.

DongDongJu (approved):
LGTM

nayeonikim:
Closing as superseded by #4865, which fixes the same object-group key parsing issue through the shared `keys.h` helper and includes the O_DIRECT test key updates.

PR #4037, opened 2026-07-07
[CI] Install FFmpeg in CPU e2e validation to fix torchcodec load failure

deng451e:
### **What this PR does / why we need it**
vllm  nighly import now `torchcodec` unconditionally at multimodal/video.py, which requires FFmpeg (`libav*`) at import time. Since the CPU CI environment doesn't include FFmpeg, `vllm serve` fails to start—even for text-only models—with:

```text
RuntimeError: Could not load libtorchcodec
OSError: libavutil.so.60: cannot open shared object file
```

This PR installs `ffmpeg` alongside `libnuma1` in `run-cpu-e2e-validation.sh`. Installation is gated by library checks to avoid unnecessary reinstalls while preserving the existing sudo fallback and fail-soft behavior.

### **Special notes for reviewers**

- Verified on Ubuntu 24.04 by reproducing the failure and confirming that installing `ffmpeg` resolves the `torchcodec` import issue.
- The script is shared by both the Buildkite pipeline and `.github/workflows/cpu_device.yml`, so this fixes both CI paths.
- Baking `ffmpeg` into the `ci-base` Docker image would be a good follow-up optimization, but this script change unblocks CI immediately.

### **If applicable**

- [ ] This PR contains user-facing changes (docs added)
- [ ] This PR contains unit tests

### DIFF
--- a/.github/scripts/run-cpu-e2e-validation.sh
+++ b/.github/scripts/run-cpu-e2e-validation.sh
@@ -406,27 +406,41 @@ if ! wait_for_endpoint_contains "http://localhost:${LMCACHE_HTTP_PORT}/healthche
 fi
 echo "✅ LMCache server is healthy"
 
-echo "[Phase 2 / Step 4] Installing libnuma (Linux only) and starting vLLM server"
+echo "[Phase 2 / Step 4] Installing system libs (Linux only) and starting vLLM server"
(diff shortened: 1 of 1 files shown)


sammshen (approved):
LGTM!

ApostaC (approved):
LGTM!

PR #3950, opened 2026-06-30
KVCache SDK: batch and stream interface

rannnayy:
**What this PR does / why we need it**:

This PR implements a stream interface for user to do KV cache modification, such as token dropping, and batch all the stream requests together to demonstrate decoding throughput improvement.

**Special notes for your reviewers**:

**If applicable**:

- [x] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #4042, opened 2026-07-07
Get query tensor v2

rannnayy:

Token dropping enables decoding throughput improvement up to 1.5x-1.7x (merged PR #3950). To drop tokens smartly, instead of doing random token drop in #3950, we need more contexts than just the KV to determine which token to drop. Therefore, this PR implements the code to get Query tensor from vLLM to LMCache, and verifies the Query tensor through a user-interface API, a thin wrapper of the KVCache SDK.

Also included a refactor of the SDK so that the KV retrieve-store mechanism could be reused cleanly for retrieving Q from the SDK.

**If applicable**:


lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -175,7 +175,7 @@ def _start_heartbeat(self) -> None:
             mq_client=self.mq_client,
             health_event=self._health_event,
             interval=self._heartbeat_interval,
-            instance_id=self.instance_id,
+            instance_ids=[self.instance_id],
```
KuntaiDu: Wondering why this is needed.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -861,6 +1145,14 @@ def wait_for_save(self):
             ops.append(meta.op)
             cache_salts.append(meta.cache_salt)
 
+        # Consume the per-step query capture plan built during save_kv_layer.
+        q_step_state = self._q_step_state
+        self._q_step_state = None
+        self._q_step_disabled = False
+
+        if q_step_state is None:
```
KuntaiDu: `if not q_step_state:`

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -871,6 +1163,15 @@ def wait_for_save(self):
         self.worker_adapter.batched_submit_store_requests(
             request_ids, ops, event, cache_salts=cache_salts
         )
+        if q_step_state is not None:
```
KuntaiDu: ditto

lmcache/integration/sglang/multi_process_adapter.py
rannnayy: We need to register the paged QRingBuffer layout to LMCache, and to do so, we need a separate instance_id. Registering Q with the same instance_id as the existing KV's will not be recognized as new registration by REGISTER_KV in the server.

In lmcache/v1/multiprocess/modules/engine_driven_transfer.py:
```
        with self._lock:
            existing = self._engine_driven_contexts.get(payload.instance_id)
            if existing is not None:
                existing.last_seen = now
                logger.info(
                    "Instance %d already registered (non-GPU); refreshing liveness",
                    payload.instance_id,
                )
                return RegisterEngineDrivenContextResponse(
                    shm_name=shm_name, pool_size=pool_size
                )
```

To make Q transfer reuse the existing LMCache transfer machinery, we need the paged structure. The paged structure needs to be registered, hence to register the paged-Q, we need a separate instance_id. 
The heartbeat then needs to be extended to handle more than one instance_id.

What do you think?

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -3,7 +3,7 @@
 # Standard
```
KuntaiDu, line 3: LoC code change ideally should be really small (e.g. <10 LoC)

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -674,6 +679,9 @@ def __init__(
                 parallel_strategy=parallel_strategy,
                 extra_config=vllm_config.kv_transfer_config.kv_connector_extra_config,
             )
```
KuntaiDu, line 702: from lmcache.experimental.engine_side import init_dispatcher
`self.dispatcher = init_dispatcher(kwargs)`

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -759,6 +781,10 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         self.worker_adapter.register_kv_caches(
             kv_caches, engine_group_infos=engine_group_infos
         )
+        if self.q_ring_capture is not None:
```
KuntaiDu: self.dispatcher.dispatch("register_kv_caches", **kwargs)

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -826,13 +852,21 @@ def save_kv_layer(
...
+        in wait_for_save().
+
         Args:
             layer_name (str): the name of the layer.
             kv_layer (torch.Tensor): the paged KV buffer of the current
                 layer in vLLM.
             attn_metadata (AttentionMetadata): the attention metadata.
             **kwargs: additional arguments for the save operation.
         """
+        if self.q_ring_capture is not None:
```
KuntaiDu: self.dispatcher.dispatch("save_kv_layer", **kwargs)

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -784,6 +815,13 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
         self.worker_adapter.register_kv_caches(
             kv_caches, engine_group_infos=engine_group_infos
         )
+        dispatch(
```
KuntaiDu: Prefer

```
if self.dispatcher is not None:
   self.dispatcher.dispatch(xxx)
```

Current code reads like dispatcher will always be initialized and dispatch will always be called, which is not the case.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1380,6 +1416,9 @@ def _reregister_kv_caches_callback(self) -> bool:
             )
             return False
         logger.warning("Finished re-registering KV caches after server recovery")
+
+        dispatch(self.dispatcher, "reregister")
```
KuntaiDu: ditto

KuntaiDu (approved):
Please add some unit test. Otherwise LGTM.

YaoJiayi (approved):

YaoJiayi (approved):

PR #4044, opened 2026-07-08
[Bugfix] Postpone the `is_kv_writer` check after starting the worker heartbeat

### LINKED ISSUE #4038: MP mode: non-writer TP ranks can be silently and permanently deregistered under slow ramp-up (MLA?)
## Summary

In MP (multiprocess) mode with an MLA model running tensor parallelism > 1, non-writer TP ranks can be silently and permanently deregistered by the server's worker-liveness reaper under a slow QPS ramp-up, with no way to recover. Once this happens, those ranks flood `ValueError: No GPU context registered for instance ID <id>` on every subsequent store/retrieve call until the pod is restarted.

## Environment

- LMCache version: 0.5.0
- Mode: multiprocess (MP) mode, single server (`n_servers=1`)
- Model: DeepSeek-V3 (MLA), `tensor-parallel-size=8`
- vLLM connector: `lmcache.integration.vllm.lmcache_mp_connector` / `vllm_multi_process_adapter`

## Observed behavior

- `vllm-container` logs show a growing count of queued/waiting requests over time.
- `lmcache` sidecar logs flood with:
  ```
  ValueError: No GPU context registered for instance ID <id>
  ```
  for a fixed, recurring set of instance IDs (7 of 8 TP ranks, in our case).
- The affected worker processes never crash or restart — confirmed via `ps aux` inside the container, all TP worker processes were alive and running the entire time.
- The lmcache server's `/status` endpoint showed only 1 tracked instance (`tracked_instances: 1`) despite 8 TP workers being active.
- No `"Reaped GPU instance"` warning was visible in the logs available to us at investigation time, though our log retention window did not reach back far enough to capture the actual reap event (see "What we've ruled out" below).

## Reproduction conditions (as far as we've narrowed it down)

- Model uses MLA (`is_mla=True`), so LMCache's own KV-cache sharding designates only one TP rank per node as a `kv_writer` (only that rank issues STORE requests; the rest do not, by design — MLA's KV cache is TP-invariant, so only one copy needs to be stored).
- Deployment ran tensor parallelism > 1 (`tensor-parallel-size=8` in our case).
- Traffic was ramped up gradually (low QPS at first), and during the low-QPS period the workload was mostly cache misses (STORE only, few or no RETRIEVE hits).
- We have not yet reproduced this against a non-MLA model or at `tensor-parallel-size=1`; we suspect (but have not confirmed on the LMCache side) that the issue is specific to MLA + TP > 1, since non-writer ranks only exist in that configuration.

## What we've ruled out

- Not a pod-level crash or OOM: no container restarts, no k8s events, no crash in the vLLM worker logs.
- Not related to the `--worker-reap-timeout-seconds` (ping-proven worker silence) path — we found no evidence of the affected ranks ever completing a PING before the failure state began.
- Not reproduced so far in offline traffic replay testing (only observed in production so far, under the specific slow-ramp-up / mostly-cache-miss traffic pattern described above).

## Config at time of incident

- `--worker-reap-timeout-seconds`: default (120s)
- `--worker-registration-grace-seconds`: default (3600s)

## Ask

We'd like guidance from maintainers on the intended design here — specifically:
- Is a worker that registers but then goes a long time without any store or retrieve activity (and therefore never starts sending PINGs) expected to eventually become permanently unrecoverable once it's reaped, or is there a self-healing path we're missing?
- Is this failure mode expected to be specific to MLA models running with tensor parallelism > 1, given that only one TP rank is a `kv_writer` in that configuration?

Happy to share more logs/details, and happy to submit a PR once we've aligned on the right approach.

ApostaC:

Fixes #4038 

For MLA models, postpone the check to `is_kv_writer` so that every worker process will go through `_ensure_heartbeat_started` for every vllm worker. 

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -843,11 +843,6 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
 
         This prevents overwrites of paged KV buffer before saving done.
         """
-        # In MLA scenario, only the first rank of the pipeline group
-        # needs to save the KV cache.
-        if not self.worker_adapter.is_kv_writer:
-            return
-
         metadata = self._get_connector_metadata()
         assert isinstance(metadata, LMCacheMPConnectorMetadata)
 
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -1399,6 +1399,9 @@ class LMCacheMPWorkerAdapter:
         """
         self._ensure_heartbeat_started()
 
+        if not self.is_kv_writer:
+            return
+
         if not self.is_healthy:
             return
 


royyhuang (approved):
LGTM

sammshen (approved):
LGTM!

PR #2705, opened 2026-03-06
[Connector] Maru: zero-copy KV cache sharing via CXL shared memory 

jooho-XCENA:
**What this PR does / why we need it**:

This PR adds [Maru](https://xcena-dev.github.io/maru/), a CXL shared-memory-based KV cache storage engine, as a new remote backend for LMCache.
With CXL memory, instances share KV cache via direct memory access rather than network transfer, eliminating the data-copy overhead of network-based backends. Beyond raw throughput, this also reduces CPU and NIC utilization on the host, freeing system resources for inference itself.
It uses the existing plugin architecture, so no changes to core logic are required. We hope this addition serves as a meaningful step for LMCache's shared storage capabilities.

For details, see [documentation](https://xcena-dev.github.io/maru/) / [github](https://github.com/xcena-dev/maru).

**P2P KV Sharing Benchmark**:

<details>
<summary>setup</summary>

### Hardware Configuration

- **GPU**: NVIDIA RTX PRO 6000 (96GB) × 2
- **CPU**: AMD EPYC 9555 64-Core × 2 (128 threads)
- **DRAM**: 756 GB DDR5 DRAM
- **CXL Memory**: 6× CXL Type-3 Memory Expander (229 GB each, 1,374 GB total)
- **Topology**:
  - Single-node
  - 2 GPUs on separate NUMA nodes (GPU0→NUMA0, GPU1→NUMA1)
  - PCIe Gen5 x16
- **Transfer**: Cross-Node NIXL TCP (UCX_TLS=cuda_ipc,cuda_copy,tcp)

### Software

- **Model**: Meta-Llama/Llama-3.1-8B-Instruct (TP=1)
- **vLLM Version**: 0.13.0+cu128
- **LMCache Version**: 0.3.13.dev97
- **NIXL Version**: 0.10.1
- **Transfer Methods**:
  - **Maru**:
    - CXL shared memory pool (200 GB, /dev/dax)
    - Zero-copy mmap between vLLM instances via maru_meta_server
  - **LMCache P2P backend (NIXL)**:
    - UCX transport (cuda_ipc,cuda_copy,tcp) over TCP loopback
    - CPU DRAM storage (200 GB)
- **OS**: Ubuntu 24.04.3 LTS, Kernel 6.17.0-14-generic
- **CUDA**: 12.8.0, Driver 580.126.09

### Data

- **Dataset**: LV-Eval: A Balanced Long-Context Benchmark with 5 Length Levels Up to 256K
</details>

image

[...]

Issue #3654, opened 2026-06-12
[RFC][MP] L1 backend extension points


jooho-XCENA:
Hi all, following up on #3262 . We agreed there that L1 basically means memory the xPU can access directly, without staging.

This is now happening in code: the CPU path, GDS (#3589, merged), DAX (#3584, in review), Maru (in progress). What's interesting is that each backend ended up needing a different part of the stack — which I think tells us something about where the extension points should be. 

Before more backends show up, it seems like a good time to step back and agree on that picture. Longer term, what I'd really like is for L1 to host multiple backends at once — local and distributed backends side by side — and getting the extension points right now would make that a lot easier later.

I'm not proposing a specific first PR here; I'd rather agree on the overall picture first, and discuss the starting point together.

### What each backend actually needed

**DAX** (#3584 , local L1 target): only the allocation side. It adds a new allocator (DRAM primary + DAX overflow), registers the mmap with cudaHostRegister so the normal copy path just works, and touches nothing else.

**GDS** (#3589 , local L1 target): allocation + data path. The slab manager plugs in behind the memory-manager seam that the PR introduced, and gpu_ops got a cuFile branch for GDS objects.

**Maru** (upcoming, distributed L1 target): allocation + control. The backend owns the key state and lifecycle — store, pin, and retrieve are calls to its own server, not local cache-engine operations. Upstream has no L1-level seam for this today, so a backend like this would have to land as backend-specific special cases inside the L1 manager — which I'd like to avoid.

So a backend may replace up to three separate things, and a fourth shows up once you run more than one L1:

1. Allocation — where the bytes live and how slots are handed out (DRAM, NVMe slab, DAX, CXL pool). #3589 already introduced a seam here, so this level is mostly covered.

2. Data movement — the copy between the GPU and the backing memory. gpu_ops hardcodes one isinstance branch per backend today (cuFile for GDS, memcpy for DRAM, …). Dispatch is already per-object so this isn't urgent, but the object or allocator could eventually own its copy function so gpu_ops stops growing a branch per backend.

[...]

PR #4052, opened 2026-07-08
[MP][Maru] Maru CXL L1 backend for MP mode

seohui-XCENA:

This PR adds [Maru](https://github.com/xcena-dev/maru), a CXL shared-memory KV cache engine, as a **shared L1 tier** for MP mode — the shared-L1 counterpart to the in-process Maru backend from #2705. Today each MP server has a private, node-local L1 (pinned DRAM / Device-DAX / GDS). With Maru, MP servers on separate nodes `mmap` one CXL pool and read each other's entries zero-copy: a node that never stored a key still hits it if a peer did, with no network transfer.

Control stays entirely with LMCache. Maru is a sibling L1 manager (`MaruL1Manager`) behind a new structural `L1ManagerInterface` Protocol; the stock controllers, L1↔L2 tiering, eviction, and L2 adapters run unchanged on top. The L1 core (`l1_manager.py`, `l1_memory_manager.py`, `internal_api.py`) is byte-identical to `dev`; the remaining edits are additive wiring only.

Design doc: `docs/design/v1/distributed/maru_l1.md` · Maru: [docs](https://xcena-dev.github.io/maru/) / [github](https://github.com/xcena-dev/maru)

**Benchmark**:

<details>
<summary>setup</summary>

- Hardware: 1× NVIDIA RTX PRO 6000 Blackwell Server Edition (96 GB), Intel Xeon 6730P (Granite Rapids), 1 TB DRAM; CXL pool via Device-DAX (`/dev/dax9.0`, 500 GB), served by `maru-server` + maru-resource-manager
- Versions: vLLM 0.24.0 · LMCache v0.1.dev1942 · Maru 0.1.0 · model `Qwen/Qwen3-8B` (bf16 KV)
- Workload: `lmcache bench engine`, `long-doc-qa` — one exported config replayed unchanged against all three setups: 40 GB KV volume → 29 documents × 10k tokens (`tokens_per_gb_kvcache=7281`), 1 query/doc, `tile` order, 4 in-flight, 128 output tokens (`ignore_eos`), seed 42
- vLLM identical in all runs: `--gpu-memory-utilization 0.5` (GPU KV ≈ 30 GiB ≈ 220k tokens < 290k-token workload, so the GPU cache alone cannot hold the working set), default prefix caching, chunk size 256
- LMCache server (MP mode, `lmcache_driven`): **B** `--l1-size-gb 60` (DRAM L1, default lazy allocator) · **B′** same + `--no-l1-use-lazy` (preallocated `MixedMemoryAllocator`) · **C** `--maru-server-url maru://localhost:5555 --maru-pool-size-gb 60 --l1-size-gb 0` (Maru CXL L1); same 60 GB capacity, no eviction in any run
- All cached runs had identical cache behavior — warm-up pass all-miss, measured pass **100% hit** (289,536 / 289,536 lookup tokens, from Prometheus counters) — so the deltas between B/B′/C isolate the L1 medium and allocator, not hit-rate differences
</details>

| Metric | A: vLLM alone | B: MP + DRAM L1 (lazy alloc, default) | B′: MP + DRAM L1 (mixed alloc) | C: MP + Maru L1 |
|---|---|---|---|---|
| Mean TTFT (ms) | 1404.0 | 215.7 (**−84.6%**) | 178.3 (**−87.3%**) | 184.6 (**−86.9%**) |
| P50 TTFT (ms) | 1244.7 | 206.5 | 175.6 | 181.3 |
| P90 TTFT (ms) | 2022.2 | 234.0 | 197.7 | 204.8 |
| P99 TTFT (ms) | 2032.7 | 376.2 | 263.2 | 283.0 |
| Benchmark duration (s) | 30.13 | 17.73 | 16.69 | 16.58 |
| Input throughput (tok/s) | 9,639 | 16,380 | 17,395 | 17,516 |
| Output throughput (tok/s) | 123.2 | 209.4 | 222.4 | 223.9 |
| Mean decode speed (tok/s) | 53.1 | 66.3 | 66.7 | 66.6 |
| Mean request latency (ms) | 3,905 | 2,150 | 2,101 | 2,109 |

All runs: 29/29 requests successful, 290,396 input tokens. 
B′ = B with `--no-l1-use-lazy` (preallocated pinned-DRAM `MixedMemoryAllocator` instead of the default `LazyMemoryAllocator`).
- Maru CXL L1 vs DRAM L1: with identical 100% hit rates

Grafana (LMCache MP dashboard) during the runs — L1 usage filling over the warm-up pass and 100%-hit read phase:

| DRAM L1 (run B) | Maru L1 (run C) |
|---|---|
| lmcache_observability | maru_observability |

- **Control / directory split.** LMCache decides everything (store / evict / promote / pin); the passive `MaruServer` is only a shared directory (key → region/offset) plus a cross-node `pin_count`. Page reclamation is driven by LMCache eviction, so the allocator's `free` is a no-op.
- **Why a sibling, not an allocator swap.** A private tier keeps membership and locks in a local dict; a shared tier needs them authoritative in the shared directory (cross-node reads and pins). `MaruMemoryAllocator` supplies the CXL medium, `MaruL1Manager` the shared control. (The MP Coordinator is unrelated: it plans fleet-wide L2 eviction on logical keys, not L1 physical-page state.) A later `L1StateBackend` seam to shrink the sibling is possible follow-up, out of scope here.
- **Startup guards** reject unsupported combos up front: gds/devdax, `skip_l1` store policy, registered/RDMA-type L2 (copy-type only), p2p, and engine_driven/auto transfer mode.

**If applicable**:

- [x] this PR contains user facing changes - docs added (`docs/source/mp/configuration.rst` + design doc)


DongDongJu:
Hello @seohui-XCENA,
Thanks for the contribution!
I have QQ before going deeper into implementation details:

Do we want to treat Maru as the long-term shared-L1 architecture, or as one possible backend implementation of a more general shared-L1 abstraction?

I understand the positive value of Maru: it provides a shared CXL pool, shared key -> location directory, and cross-node pin accounting. 
That is useful for quickly validating shared L1. 
However, these concepts do not seem inherently Maru-specific. 
LMCache could potentially evolve a native shared-L1 path using Device-DAX plus the MP coordinator or a coordinator-adjacent L1 directory service and community moving to towards to that.

So I wonder whether we should first define a generic shared-L1 backend boundary, 
for example an `L1StateBackend` / `L1DirectoryBackend`, and then support both:


- LMCache-native (devdax/gds/p2p) + coordinator-backed shared L1
- Maru-backed shared L1

under the same abstraction.

That would avoid baking Maru into the L1 control path as a separate full `MaruL1Manager`, and would also let us compare both implementations behind the same interface. 
In that model, Maru can still be valuable as an implementation, but LMCache would keep the shared-L1 architecture generic.

Would you guys be open to collaborating on that direction, or explaining what Maru provides that would be difficult or undesirable to build with LMCache native plus the LMCache coordinator?

youngrok-XCENA:
Thanks for the quick response, @DongDongJu 

This PR gives Maru users an MP-mode path, and it's built on the generic L1 interface we proposed in https://github.com/LMCache/LMCache/issues/3654.

We're on board with your suggestion to keep the generic shared-L1 layer backend-neutral.

That said, fully implementing that direction would involve changes to the LMCache L1 core — which this PR deliberately leaves untouched — so I'd suggest we take it up in a dedicated issue or PR.

DongDongJu:
> Thanks for the quick response, @DongDongJu
> 
> This PR gives Maru users an MP-mode path, and it's built on the generic L1 interface we proposed in #3654.
> 
> We're on board with your suggestion to keep the generic shared-L1 layer backend-neutral.
> 
> That said, fully implementing that direction would involve changes to the LMCache L1 core — which this PR deliberately leaves untouched — so I'd suggest we take it up in a dedicated issue or PR.

MP coordinator is intended to become the distributed L1 control plane, then I do not think we need a full external L1 manager abstraction here.

The coordinator can own global metadata, placement, ownership, leases/pins, stale cleanup, and eviction coordination. DevDAX/CXL/GDS/Maru-like components should then be memory/data-plane backends, not separate owners of L1 lifecycle.

In that model, `MaruL1Manager` feels like it bypasses the coordinator and creates another source of truth for L1 membership and pinning. I think we should clarify this ownership model before merging a Maru-specific full L1 manager.

@ApostaC @deng451e @sammshen please leave you guys any opinion.

youngrok-XCENA:
@DongDongJu Thanks — we support the ownership model you're proposing: the coordinator owns the fleet-wide directory, placement, and leases/pins, and Maru runs under that boundary as the memory/data-plane backend for a shared CXL pool.

This PR already follows that premise. Every cache-control decision is made by the unchanged stock controllers; `MaruL1Manager` only executes those calls, and the L1 core is untouched (`l1_manager.py` is byte-identical to dev — hence the "sibling" behind `L1ManagerInterface`).

You're right that today the authority for shared-pool membership and the read-safety refcount sits in MaruServer. But that isn't bypassing the coordinator — it has no shared-L1 directory or lease path yet. And that state can't stay in each process's local `_objects` map the way stock L1 does: on a shared pool, an instance has to find keys it never wrote, and before a page is reclaimed we have to know no other instance is still reading it. Both need a shared home next to the pool.

On why a full sibling rather than a narrower seam: the existing `MemoryObj`/allocator override points are per-server seams that can't reach that shared state, and a narrower seam — the `L1StateBackend` direction you sketched — means reshaping the L1 core, which we didn't want to do unilaterally in a backend PR before the ownership model is settled. Narrowing it to the exact shared-L1 boundary is our next item — glad to shape it together.

Building Maru gave us implementation-backed requirements — shared-pool placement, cross-instance read safety, allocation ownership, recovery — that we'd like to work through together as that model takes shape; meanwhile we're happy to document the current ownership explicitly and keep the path opt-in.

seohui-XCENA:
Hello! Gentle ping on this one. Any chance someone could take another look?
We'd like to give Maru users an MP-mode path, since Maru's non-MP storage backend is deprecated.

@ApostaC @sammshen @deng451e @DongDongJu @ruizhang0101 could you let us know what this PR would need in order to move forward? If there's a shape you'd prefer, or specific changes you'd want first, we're happy to do the work.

jooho-XCENA:
The direction in #5030/#5105 reinforces the rationale for the shared-pool authority model we had been developing and [described here in July](https://github.com/LMCache/LMCache/pull/4052#issuecomment-4934280592). Maru was designed for cross-host access to fabric-attached CXL, and we are currently testing that deployment through a physical CXL switch.

Keeping pool resource management separate also matters when applications beyond LMCache share the fabric’s capacity. Maru’s Resource Manager offers a basis for that common layer, while LMCache retains its cache policy.

Could we revisit this PR in light of that common direction and clarify any remaining ownership or interface concerns? We’d be happy to contribute our implementation and fabric-testing experience toward a common backend-neutral contract.

@ApostaC, building on your [earlier guidance](https://github.com/LMCache/LMCache/issues/3654#issuecomment-4734183518) that LMCache should retain native cache logic and the authoritative view of object locations, would separating shared physical-pool allocation from LMCache’s object metadata and lifecycle be a suitable direction? It would help to understand which responsibilities you would want the common L1 interface to retain in LMCache and which it could delegate to a backend such as Maru.

PR #4062, opened 2026-07-09
[sglang] Support DSA (DeepSeek-V3.2) pools via a second kernel group …

liuyumoye:
### What this PR does / why we need it
Add DSA (DeepSeek-V3.2 sparse attention) support for SGLang multi-process LMCache integration.

SGLang DSA maintains two coupled layer buffers sharing page allocator & block-id space but different tensor layouts:
1. `kv_buffer[layer]`: Fused latent K/V `(PBS, 1, HS)`, bf16/fp8
2. `index_k_with_scale_buffer[layer]`: 2D uint8 page blob storing fp8 keys + fp32 scales

Referencing vLLM hybrid-KV design, we register two `EngineGroupInfo` with the same `engine_group_id=0` to sync store/evict of both buffers.

### Code Changes
1. `lmcache/integration/sglang/multi_process_adapter.py`
   - Add `dsa_index_buffer` param to `LMCacheMPConnector`
   - DSA registers dual engine groups; original single-group logic reserved for MHA/MLA
   - Replace hardcoded block-id duplication with `expand_engine_block_ids`

2. New `lmcache/integration/sglang/kv_cache_groups.py`
   - Factory function returns dual groups for DSA, empty list for vanilla attention
   - Minimal vLLM-aligned layout hints without extra DSA flag

3. `lmcache/v1/gpu_connector/kv_format/detectors/sglang.py`
   - Standardize MLA tensor layout to unified `NL_X_NB_BS_HS`
   - Detect 2D uint8 DSA index buffer, reshape to isolate single-page blocks
   - Support mixed-shape tensor list split detection

4. `lmcache/integration/sglang/__init__.py`
   - Remove legacy monkey-patch & DSA pool registry
   - Directly consume `k_pool/v_pool/dsa_index_buffer` from new SGLang interface
   - Identify MLA via `k_pool[i] is v_pool[i]`

### Special Notes for Reviewers
1. Dependency: Matching SGLang PR required to pass `dsa_index_buffer`：https://github.com/sgl-project/sglang/pull/30662
2. Backward Compatibility: MHA/MLA wire protocol unchanged, fully bit-compatible
3. Design: Shared `engine_group_id=0` ensures one block-id controls both KV and index pages
4. Identification: DSA index buffer uniquely matched by `ndim=2 + uint8 dtype`
5. Guard: Throw error if DSA buffer is used with non-MLA pools

### Validation
- No user-facing config/docs changes; auto-enable when `dsa_index_buffer` is not None
- No new unit tests, full end-to-end manual validation with DeepSeek-V3.2
- Warm hit output bitwise consistent with cold inference

### Checklist
- [ ] Docs updated (No user-facing changes)
- [ ] Unit tests added (Manually verified)

### DIFF
--- a/lmcache/integration/sglang/__init__.py
+++ b/lmcache/integration/sglang/__init__.py
@@ -1,2 +1,3 @@
 # SPDX-License-Identifier: Apache-2.0
-
+#
+# Package marker.
--- a/lmcache/integration/sglang/multi_process_adapter.py
+++ b/lmcache/integration/sglang/multi_process_adapter.py
@@ -16,10 +16,11 @@ import zmq
 
 # First Party
 from lmcache import torch_dev
-from lmcache.integration.sglang.sglang_adapter import (
-    LoadMetadata,
-    StoreMetadata,
+from lmcache.integration.sglang.kv_cache_groups import (
+    build_sglang_layout_hints,
+    create_engine_group_infos_from_sglang,
 )
+from lmcache.integration.sglang.sglang_adapter import LoadMetadata, StoreMetadata
 from lmcache.integration.vllm.vllm_multi_process_adapter import (
     DEFAULT_HEARTBEAT_INTERVAL,
     DEFAULT_MQ_TIMEOUT,
@@ -35,6 +36,7 @@ from lmcache.v1.multiprocess.custom_types import (
     KVCache,
 )
 from lmcache.v1.multiprocess.futures import MessagingFuture
+from lmcache.v1.multiprocess.group_view import EngineGroupInfo, expand_engine_block_ids
 from lmcache.v1.multiprocess.mq import MessageQueueClient
 from lmcache.v1.multiprocess.protocol import RequestType
 from lmcache.v1.platform.cuda.ipc_wrapper import CudaIPCWrapper
@@ -47,19 +49,18 @@ _WAIT_LOOKUP_RESPONSE_BUFFER_S = 5.0
 
 
 def _wrap_sglang_kv_caches(
-    k_pool: list[torch.Tensor],
-    v_pool: list[torch.Tensor],
+    kv_cache_pools: list[list[torch.Tensor]],
 ) -> KVCache:
-    """Flatten SGLang's depth-2 ``[K_layers, V_layers]`` KV layout into a
-    single flat ``KVCache`` so it fits upstream's wire
-    ``KVCache`` payload type. The daemon's
-    :func:`normalize_kv_and_discover_format` recognizes this shape from
-    ``EngineType.SGLANG`` plus a ``tokens_per_block`` ``LayoutHints`` field
-    and splits it back at its midpoint before format detection.
+    """Flatten SGLang's pool groups into a single flat ``KVCache`` so it
+    fits upstream's wire ``KVCache`` payload type.  Each element of
+    *kv_cache_pools* is a per-layer tensor list for one pool group
+    (K, V, DSA-index, …).  The daemon's
+    :func:`normalize_kv_and_discover_format` resolves the layout from
+    ``EngineType.SGLANG`` plus ``tokens_per_block``.
     """
     wrapped: KVCache = []
-    wrapped.extend(CudaIPCWrapper(tensor) for tensor in k_pool)
-    wrapped.extend(CudaIPCWrapper(tensor) for tensor in v_pool)
+    for pool in kv_cache_pools:
+        wrapped.extend(CudaIPCWrapper(tensor) for tensor in pool)
     return wrapped
 
 
@@ -132,8 +133,7 @@ class LMCacheMPConnector:
         page_size: int,
         host: str,
         port: int,
-        k_pool: list[torch.Tensor],
-        v_pool: list[torch.Tensor],
+        kv_cache_pools: list[list[torch.Tensor]],
         tp_group: Optional[torch.distributed.ProcessGroup] = None,
         mq_timeout: float = DEFAULT_MQ_TIMEOUT,
         heartbeat_interval: float = DEFAULT_HEARTBEAT_INTERVAL,
@@ -141,9 +141,9 @@ class LMCacheMPConnector:
         self.tp_size = tp_size
         self.worker_id = rank
         self.page_size = page_size
-        self.device = k_pool[0].device
+        self.device = kv_cache_pools[0][0].device
         self.model_name = sgl_config.model_path
-        self.num_layers = len(k_pool)
+        self.num_layers = len(kv_cache_pools[0])
         self.tp_group = tp_group
         self.instance_id = os.getpid()
         self._mq_timeout = mq_timeout
@@ -165,27 +165,29 @@ class LMCacheMPConnector:
                 f"{self._lmcache_chunk_size} and {self.page_size}"
             )
 
+        # DSA classification is delegated to ``create_engine_group_infos_from_sglang``
+        # (uses the same ``is_deepseek_dsa(hf_config)`` check as sglang's pool init).
+        layout_hints = build_sglang_layout_hints(self.page_size)
+        self.engine_group_infos: list[EngineGroupInfo] = (
+            create_engine_group_infos_from_sglang(
+                kv_cache_pools, self.page_size, sgl_config.hf_config
+            )
+        )
+
         # Upstream's REGISTER_KV_CACHE protocol takes flat positional args:
         # (instance_id, kv_cache, model_name, world_size, engine_type,
-        # layout_hints, engine_group_infos). SGLang's natural KV layout is depth-2
-        # ([K_layers, V_layers]); we flatten it on the wire to fit
-        # ``KVCache = list[DeviceIPCWrapper]``. The daemon recognizes the
-        # SGLang-MHA flat-of-2NL pattern from ``EngineType.SGLANG`` plus the
-        # ``tokens_per_block`` hint and un-flattens + reshapes per layer.
-        # SGLang is non-hybrid (a single KV cache group), so engine_group_infos is the
-        # empty list -- which the server treats as one group spanning all layers
-        # (matching the vLLM non-hybrid and TensorRT-LLM register paths).
+        # layout_hints, engine_group_infos).
         send_lmcache_request(
             self.mq_client,
             RequestType.REGISTER_KV_CACHE,
             [
                 self.instance_id,
-                _wrap_sglang_kv_caches(k_pool, v_pool),
+                _wrap_sglang_kv_caches(kv_cache_pools),
                 self.model_name,
                 self.tp_size,
                 EngineType.SGLANG,
-                {"tokens_per_block": self.page_size},
-                [],
+                layout_hints,
+                self.engine_group_infos,
             ],
         ).result(timeout=self._mq_timeout)
         self._registered = True
@@ -210,11 +212,18 @@ class LMCacheMPConnector:
         return self._lmcache_chunk_size
 
     @torch.no_grad()
-    def _global_min_tokens(self, local_tokens: int) -> int:
+    def _broadcast_lookup_result(self, local_tokens: int) -> int:
+        """Broadcast the lookup result from rank 0 to all TP ranks.
+
+        Only rank 0 performs the daemon LOOKUP + WAIT_PREFETCH_STATUS
+        flow (see :meth:`lookup_kv`). This method broadcasts the
+        result so every rank sees the same matched-token count without
+        independently querying the daemon.
+        """
         if self.tp_size == 1:
             return local_tokens
         t = torch.tensor([local_tokens], dtype=torch.int32, device=self.device)
-        dist.all_reduce(t, op=dist.ReduceOp.MIN, group=self.tp_group)
+        dist.broadcast(t, src=0, group=self.tp_group)
         return int(t.item())
 
     def _create_key(
@@ -286,6 +295,7 @@ class LMCacheMPConnector:
         end: int,
         request_id: str,
     ) -> None:
+        """Release *this rank's* read locks in ``[start, end)``."""
         if start >= end or not self.is_healthy:
             return
         send_lmcache_request(
@@ -297,7 +307,7 @@ class LMCacheMPConnector:
                     start=start,
                     end=end,
                     request_id=request_id,
-                    no_worker_id=True,
+                    no_worker_id=False,
                 ),
                 self.tp_size,
             ],
@@ -323,9 +333,8 @@ class LMCacheMPConnector:
         if not self.is_healthy or not request_id:
             return 0
 
-        # If a previous LOOKUP for this rid is still pending (e.g., a
-        # rescheduling pass or a prior partial flow), release its locks
-        # first so we don't accumulate read-lock reservations.
+        # Release any stale locks from a prior LOOKUP for the same rid
+        # (e.g., a rescheduling pass) before firing a new one.
         with self._pending_lookups_lock:
             stale = self._pending_lookups.pop(request_id, None)
         if stale is not None and stale.locks_held:
@@ -339,20 +348,25 @@ class LMCacheMPConnector:
         if aligned_end == 0:
             return 0  # too few tokens; no chunk-aligned range to LOOKUP
 
-        lookup_key = self._create_key(
-            token_ids,
-            start=0,
-            end=aligned_end,
-            request_id=request_id,
-            no_worker_id=True,
-        )
-        send_lmcache_request(
-            self.mq_client,
-            RequestType.LOOKUP,
-            [lookup_key, self.tp_size],
-        ).result(timeout=self._mq_timeout)
-        matched = self._wait_for_lookup(request_id)
-        matched = self._global_min_tokens(matched)
+        # Only rank 0 talks to the daemon; the matched-token count is
+        # broadcast to the followers.
+        if self.worker_id == 0:
+            lookup_key = self._create_key(
+                token_ids,
+                start=0,
+                end=aligned_end,
+                request_id=request_id,
+                no_worker_id=True,
+            )
+            send_lmcache_request(
+                self.mq_client,
+                RequestType.LOOKUP,
+                [lookup_key, self.tp_size],
+            ).result(timeout=self._mq_timeout)
+            matched = self._wait_for_lookup(request_id)
+        else:
+            matched = 0
+        matched = self._broadcast_lookup_result(matched)
 
         # Daemon now holds read locks for the matched chunks. Record
         # state for the eventual retrieve_kv / release_pending /
@@ -398,11 +412,19 @@ class LMCacheMPConnector:
             pending = self._pending_lookups.pop(request_id, None)
         if pending is None:
             return
+        # Every rank frees *its own* still-held read locks. LOOKUP is
+        # rank-0-only on the wire, but the daemon reserves an
+        # independent lock per TP rank (worker_id=None expansion), so a
+        # rank-0-only release would leak the other ranks' locks until
+        # read-lock TTL expiry.
         if pending.locks_held and pending.matched_token_num > 0:
             self._free_lookup_locks(
                 pending.token_ids, 0, pending.matched_token_num, request_id
             )
-        send_lmcache_request(self.mq_client, RequestType.END_SESSION, [request_id])
+        # END_SESSION targets shared daemon-side session state, so only
+        # rank 0 sends it.
+        if self.worker_id == 0:
+            send_lmcache_request(self.mq_client, RequestType.END_SESSION, [request_id])
 
     def _submit_retrieve(
         self,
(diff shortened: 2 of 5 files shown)


lmcache/integration/sglang/__init__.py
```diff
@@ -1,2 +1,18 @@
...
+# from its ``kv_buffer`` list to grab the ``index_k_with_scale_buffer``.
+#
+# Both mechanisms have been removed. sglang's
+# ``mem_cache/storage/lmcache/lmc_radix_cache.py`` now classifies the KV pool
+# explicitly (see ``_extract_kv_pools``) and passes ``k_pool``, ``v_pool`` and
+# the optional ``dsa_index_buffer`` directly into ``LMCacheMPConnector``.
+# The adapter recognizes the MLA / DSA case by ``k_pool[i] is v_pool[i]`` and
+# picks the right on-wire layout without any pool-side patching.
+#
+# The module is kept as an importable package marker; nothing needs to run at
```
Shaoting-Feng: Can you clean up the comment?

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -316,20 +323,29 @@ def lookup_kv(self, token_ids: list[int], request_id: str) -> int:
...
-            no_worker_id=True,
-        )
-        send_lmcache_request(
-            self.mq_client,
-            RequestType.LOOKUP,
-            [lookup_key, self.tp_size],
-        ).result(timeout=self._mq_timeout)
-        matched = self._wait_for_lookup(request_id)
-        matched = self._global_min_tokens(matched)
+        # Only the lead TP rank (rank 0) communicates with the daemon.
```
Shaoting-Feng: This change is irrelevant to DSA support. It should be in a separate PR.

lmcache/v1/gpu_connector/kv_format/detectors/sglang.py
```diff
@@ -55,13 +55,57 @@ def discover(
...
+                    or layer.shape[0] % block_size != 0
+                ):
+                    mla_ok = False
+                    break
+                num_blocks = layer.shape[0] // block_size
+                mla_reshaped.append(layer.view(num_blocks, block_size, layer.shape[2]))
+            if mla_ok:
+                return lmc_ops.EngineKVFormat.NL_X_NB_BS_HS, mla_reshaped
+
+        # -- MLA fallback (fused PBS): shape[1] == 1, no block hint --
```
Shaoting-Feng: Why MLA (with tokens_per_block) and MLA fallback coexist? Which is the codepath for MLA models?

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -316,20 +323,29 @@ def lookup_kv(self, token_ids: list[int], request_id: str) -> int:
...
+        # All sglang TP ranks share the same scheduler loop and the same
```
Shaoting-Feng: retrieve runs on every rank, which mat over-release daemon read locks

lmcache/integration/sglang/__init__.py
liuyumoye: Thanks for pointing it out, cleaned up.

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: Thanks for the review. To clarify, this PR only touches the LOOKUP path — RETRIEVE is completely unchanged. The two use independent lock mechanisms with no shared reference counts:

### LOOKUP (what this PR modifies)
Only rank 0 fires LOOKUP (`lookup_kv` has `if self.worker_id == 0` guard around line 296 in `multi_process_adapter.py`). The daemon's `LookupModule.lookup()` acquires read locks via `submit_prefetch_task`.

After LOOKUP, the matched-token count is broadcast to all ranks via `_broadcast_lookup_result`.

Lock release (`_free_lookup_locks` → `FREE_LOOKUP_LOCKS` → daemon's `finish_read_prefetched`) is always guarded by `self.worker_id == 0`. One acquire → one release, no over-release possible.

### RETRIEVE (untouched by this PR)
`retrieve_kv` fires on every rank — each TP rank independently calls `_submit_retrieve`, which sends a RETRIEVE request to the daemon.

The daemon's RETRIEVE handler does its own `read_prefetched_results` (acquire) → `finish_read_prefetched` (release) per rank. This is the same pattern the vLLM adapter uses and is correct — every TP rank needs to copy its own KV shard into GPU memory.

These RETRIEVE-level read locks are entirely daemon-internal and unrelated to the LOOKUP-level lock pool.

### Summary
In short: LOOKUP acquires once (rank 0), releases once (rank 0). RETRIEVE does its own 1:1 acquire/release per rank on a separate lock pool. The two paths never share reference counts, so each rank running RETRIEVE cannot cause over-release of the LOOKUP locks — there is no "over-release daemon read locks" issue here.

lmcache/v1/gpu_connector/kv_format/detectors/sglang.py
liuyumoye: For DSA models, the flat KV list mixes MLA KV tensors (3-D) with DSA index buffers (2-D uint8). `normalize_and_discover_per_layer_formats` calls discover first on the whole mixed list, which can't be precisely classified. The fallback returns a generic per-layer format to let the per-shape retry classify each uniform group independently. I've refactored the function to make the structure clearer.

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -187,11 +186,18 @@ def chunk_size(self) -> int:
         return self._lmcache_chunk_size
 
     @torch.no_grad()
-    def _global_min_tokens(self, local_tokens: int) -> int:
+    def _broadcast_lookup_result(self, local_tokens: int) -> int:
+        """Broadcast the lookup result from rank 0 to all TP ranks.
+
+        Only rank 0 performs the daemon LOOKUP + WAIT_PREFETCH_STATUS
```
DongDongJu, line 218: what happened for other parallelism? is this guaranteed?

DongDongJu (commented):
Hello @liuyumoye, Thanks for your contribution!
I left one question.

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -316,20 +323,29 @@ def lookup_kv(self, token_ids: list[int], request_id: str) -> int:
...
+            )
+            send_lmcache_request(
+                self.mq_client,
+                RequestType.LOOKUP,
+                [lookup_key, self.tp_size],
+            ).result(timeout=self._mq_timeout)
+            matched = self._wait_for_lookup(request_id)
+        else:
+            matched = 0
+        matched = self._broadcast_lookup_result(matched)
```
ApostaC: Is it correct for a non-MLA model? In these models, different TP ranks will have different keys.
Although it's unlikely, different ranks may have a different number of matches in that case.

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -455,7 +484,15 @@ def retrieve_kv(self, load_metadata: LoadMetadata) -> int:
...
-        self._free_lookup_locks(token_ids, 0, offset, request_id)
+        # Only the lead TP rank (rank 0) releases the prefix read locks.
+        # All ranks share the same LOOKUP-level lock reservation; releasing
+        # from every rank would over-release (each call via FREE_LOOKUP_LOCKS
+        # releases 1+extra_count references).  The trailing suffix locks are
+        # released inside the daemon's RETRIEVE handler via
+        # finish_read_prefetched, with extra_count correctly accounted for
+        # at LOOKUP time.
+        if self.worker_id == 0:
+            self._free_lookup_locks(token_ids, 0, offset, request_id)
```
ApostaC: The same here, if we want to keep all-rank lookup for non-MLA models, here needs to be updated as well.

ApostaC (changes requested):
Thanks for the contribution! Some concerns on non-MLA model correctness. PTAL!

Also, it would be great if you can share some benchmarking results for DSA. (no need to be too complicated)

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: Good point. From my understanding: for non-MLA setups, each TP rank maintains its own separate keys. 
On the client side, rank 0 sends LOOKUP with worker_id=None, then broadcasts one result:

```
def _create_key(..., no_worker_id: bool = False) -> IPCCacheServerKey:
    return IPCCacheServerKey(
        model_name=self.model_name,
        world_size=self.tp_size,
        worker_id=None if no_worker_id else self.worker_id,
        ...
    )
lookup_key = self._create_key(..., no_worker_id=True,)
```
Server side, worker_id=None is expanded into all TP ranks (kv_rank per worker), so the lookup checks all shards:
```
if ipc_key.worker_id is None:
    kv_ranks = [
        ObjectKey.ComputeKVRank(..., global_rank=worker_id, ..., local_rank=worker_id)
        for worker_id in range(ipc_key.world_size)
    ]
else:
    kv_ranks = [ObjectKey.ComputeKVRank(..., global_rank=ipc_key.worker_id, ...)]
```
Then hit-length is computed from leading contiguous found keys and converted to chunk hits by dividing with world_size * num_object_groups:
```
# lookup.py
found_count = _get_prefix_hit_length(
    found.count_leading_ones(), job.world_size, job.num_object_groups
)
>
def _get_prefix_hit_length(found_prefix_len, world_size, num_object_groups):
    return found_prefix_len // (world_size * num_object_groups)
```
So in non-MLA, even if rank-local availability differs, reported hit length is the common prefix that is valid across all ranks (safe/conservative, may under-report but should not over-report).
Please let me know if this matches your understanding.

ApostaC: Oh yeah that makes sense!

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: As mentioned above, this part does not require updates.

liuyumoye:
> Thanks for the contribution! Some concerns on non-MLA model correctness. PTAL!
> 
> Also, it would be great if you can share some benchmarking results for DSA. (no need to be too complicated)

Thanks for your review! I will double-check the correctness for non-MLA models and provide simple DSA benchmark results：

## TTFT Summary
| length | cold\_miss | radix\_hit | lmcache\_hit |
|--------|-----------|-----------|-------------|
| 8192   | 122.3ms |     45.0ms |       62.2ms |
| 32768  | 698.2ms |     74.8ms |      166.0ms  |

## Test Startup Commands
### 1. LMCache Daemon

``` shell
python3 -m lmcache.v1.multiprocess.server \\
  --host 127.0.0.1 \\
  --port 5556 \\
  --chunk-size 256 \\
  --l1-size-gb 128 \\
  --max-workers 8 \\
  --eviction-policy LRU \\
  --no-l1-use-lazy \\
  --prometheus-port 19100
```

### 2. SGLang Server with LMCache Enabled
```
python3 -m sglang.launch_server \
  --model-path deepseek-ai/DeepSeek-V2-Lite-Chat \
  --served-model-name ds-v2-lite \
  --host 0.0.0.0 \
  --port 8021 \
  --tp 8 \
  --trust-remote-code \
  --dtype auto \
  --context-length 82768 \
  --page-size 64 \
  --chunked-prefill-size 8192 \
  --max-running-requests 16 \
  --mem-fraction-static 0.7 \
  --enable-lmcache \
  --lmcache-config-file lmcache_config.yaml \
  --enable-metrics \
  --log-level-http warning
```
lmcache_config.yaml:
```
mp_host: 127.0.0.1
mp_port: 5556
chunk_size: 256
```

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: Great question. We only tested TP > 1 && EP > 1 cases for this PR. Without the current changes, MHA models running under multi-TP cannot store or retrieve KV cache properly across ranks. Other parallelism strategies remain untested. From my understanding, DP is likely compatible, but PP and CP may introduce new issues. If cross-support for other parallel modes is needed, I’d like to implement it in a separate PR later.

ApostaC (approved):
lgtm！

ApostaC:
@liuyumoye thanks for sharing the testing results!
@DongDongJu @Shaoting-Feng do you guys want to take another quick look?

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -187,11 +186,18 @@ def chunk_size(self) -> int:
...
+        Only rank 0 performs the daemon LOOKUP + WAIT_PREFETCH_STATUS
+        flow (see :meth:`lookup_kv`). This method broadcasts the
+        result so every rank sees the same matched-token count without
+        independently querying the daemon.
+        """
         if self.tp_size == 1:
             return local_tokens
         t = torch.tensor([local_tokens], dtype=torch.int32, device=self.device)
-        dist.all_reduce(t, op=dist.ReduceOp.MIN, group=self.tp_group)
+        dist.broadcast(t, src=0, group=self.tp_group)
```
DongDongJu, line 226: I think this rank is global. isnt it?

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -302,10 +308,11 @@ def lookup_kv(self, token_ids: list[int], request_id: str) -> int:
 
         # If a previous LOOKUP for this rid is still pending (e.g., a
         # rescheduling pass or a prior partial flow), release its locks
-        # first so we don't accumulate read-lock reservations.
+        # first so we don't accumulate read-lock reservations. Only the
+        # lead rank (worker 0) communicates with the daemon.
         with self._pending_lookups_lock:
             stale = self._pending_lookups.pop(request_id, None)
-        if stale is not None and stale.locks_held:
+        if stale is not None and stale.locks_held and self.worker_id == 0:
```
DongDongJu: But here we using local_rank. not sure this semantic works for all the cases.

DongDongJu (commented):
Hello @liuyumoye,
Thanks for the great contribution.
I just left one concern about global rank vs local rank but LMK if I was wrong.

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: Thanks for flagging this — I traced the call chain to confirm the semantics:

The rank received here is the TP-group-global rank (`ps.tp_rank`), not a node-local rank.

### Full call chain
1. `multi_process_adapter.py`: `self.worker_id = rank`
2. `registry.py`: `LMCRadixCache(..., rank=ctx.tp_rank, ...)`
3. `kv_cache_builder.py`: `tp_rank=ps.tp_rank`
4. `scheduler.py`: `ParallelState(tp_rank=tp_rank, ...)`
5. `engine.py` / `data_parallel_controller.py`: Each node owns a slice of `[0, tp_size)`, and slices from all nodes combined exactly cover the range `0..tp_size-1`.

Across the entire TP group, there is exactly one worker with `tp_rank == 0`, regardless of the number of nodes.
This rank is the same source rank `src=0` used in `dist.broadcast(..., group=self.tp_group)`.
As a result, the daemon communication leader and broadcast source are guaranteed to be the same process.
The `worker_id == 0` gate works correctly for all cases below:
- Single-node TP
- Multi-node TP (only the node holding TP-rank-0 communicates with the daemon)
- MLA (all TP ranks hold identical KV latents; fetching on rank-0 then broadcasting is inherently correct per design)

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -483,9 +520,13 @@ def retrieve_kv(self, load_metadata: LoadMetadata) -> int:
...
             if not retrieve_succeeded:
-                self._free_lookup_locks(
-                    token_ids, offset, retrieve_token_num, request_id
-                )
+                # Free the suffix read locks that the daemon's RETRIEVE
+                # handler would normally consume on success. Only the lead
+                # rank does this — all ranks share the same lock reservation.
+                if self.worker_id == 0:
+                    self._free_lookup_locks(
+                        token_ids, offset, retrieve_token_num, request_id
```
Shaoting-Feng: The failure cleanup assumes all TP ranks see the same RETRIEVE outcome, but the N RETRIEVEs are independent requests:

1. **Worker 0 fails, others succeed → double-release.** The successful workers' read locks were already consumed by the daemon's RETRIEVE handler (`finish_read_prefetched`); worker 0's `_free_lookup_locks` (worker-agnostic key, expands to all workers) then releases them again. The read-lock counter is shared across requests, so this can silently unpin a chunk a concurrent request is still reading.
2. **Worker j≠0 fails, worker 0 succeeds → leak.** Worker j is blocked by the `worker_id == 0` gate; worker 0 never enters the failure branch. Bounded by the read-lock TTL, but chunks stay pinned for up to `read_ttl_seconds`.

lmcache/integration/sglang/multi_process_adapter.py
Shaoting-Feng: I think in `read_prefetched_results`, it calls `unsafe_read` so there is no acquiring locks.

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: Thanks — you're right on both points, and they stem from the same root issue. The behavior of `unsafe_read` is exactly what breaks the original failure-cleanup logic, and this fix resolves both problems. The changes are available in the latest multi_process_adapter.py.

#### Lock Accounting Rules
SGLang setup: world_size == tp_size, non-MLA models → extra_count = 0
1. **LOOKUP**: Only rank 0 runs, no_worker_id=True
   The daemon expands `ipc_key_to_object_keys` into N independent per-rank ObjectKeys, each acquiring 1 read lock.
2. **RETRIEVE**: All ranks run independently, no_worker_id=False (bind to its own worker_id)
   `read_prefetched_results` uses `unsafe_read` (no additional lock acquisition). On success, `finish_read_prefetched(extra_count=0)` decrements only the lock counter for this rank's worker-key.
3. **FREE_LOOKUP_LOCKS (fixed)**: All ranks run independently, no_worker_id=False
   Only releases the lock bound to the current rank's worker-key.

RETRIEVE consumes exactly one pre-reserved lock allocated during LOOKUP for each rank, so failure cleanup must also operate per-rank.

#### Root Cause of Two Failure Scenarios
1. **Rank 0 fails, other ranks succeed → double-release**
Ranks 1..N-1 already release their locks via `finish_read_prefetched`. Previously rank 0 triggered cleanup with a worker-agnostic key. The daemon broadcast this release to all workers and re-released locks already at zero. This could silently unpin chunks that ongoing concurrent requests are accessing.

2. **Non-zero rank j fails, rank 0 succeeds → lock leak**
The original `if self.worker_id == 0` guard blocked cleanup from rank j, while rank 0 took the success path and skipped cleanup. The lock belonging to worker j remained held until read-lock TTL expired.

#### Fix Details
- Make `_free_lookup_locks` use `no_worker_id=False` to target per-worker keys.
- Remove all `worker_id == 0` guards along every lock-release path:
  - Failure cleanup in lookup_kv
  - release_pending
  - lock release in end_session
  - Prefix release `[0, offset)` in retrieve_kv
  - Failure suffix release `[offset, retrieve_token_num)` in retrieve_kv

Note: END_SESSION remains rank-0 only, since session state is shared; this prevents redundant "session not found" warnings on follower ranks.

#### Post-fix Invariant
LOOKUP (rank 0, worker_id=None) reserves N independent per-rank locks.
Each rank owns exactly one lock and bears full responsibility for releasing it, via either:
- Successful RETRIEVE
- Failure cleanup
- release_pending
- end_session

Both double-release and TTL-bounded lock leak hazards are eliminated.

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -286,6 +295,21 @@ def _free_lookup_locks(
         end: int,
         request_id: str,
     ) -> None:
+        """Release this rank's read locks in ``[start, end)``.
+
+        Uses a per-worker key (``no_worker_id=False``) so the daemon
```
Shaoting-Feng: nit: This part of the docstring is not needed because it describes too much about the implementation detail.

lmcache/integration/sglang/multi_process_adapter.py
```diff
@@ -339,20 +366,29 @@ def lookup_kv(self, token_ids: list[int], request_id: str) -> int:
...
-        )
-        send_lmcache_request(
-            self.mq_client,
-            RequestType.LOOKUP,
-            [lookup_key, self.tp_size],
-        ).result(timeout=self._mq_timeout)
-        matched = self._wait_for_lookup(request_id)
-        matched = self._global_min_tokens(matched)
+        # Only the lead TP rank (rank 0) communicates with the daemon.
+        # All sglang TP ranks share the same scheduler loop and the same
```
Shaoting-Feng: nit: lines 370-374 are not needed because it describes too much about the implementation detail.

Shaoting-Feng (approved):
Generally LGTM. Some minor comments on comments.

lmcache/integration/sglang/multi_process_adapter.py
liuyumoye: Thanks! I've updated the code comments accordingly.

PR #4075, opened 2026-07-10
refactor(mp):Platform refactor ipc abstraction

superleo:

This PR introduces a platform-level event IPC abstraction for the LMCache-driven multiprocess handle path.

- Adds `EventIPCBackend`, a CUDA-style default implementation, and `DeviceSpec`-based backend lookup.
- Adds a fail-closed MUSA backend using the current TorchMUSA interprocess `Event` API through `platform/musa/ipc_wrapper.py`.
- Routes server-side store/retrieve event creation, import, wait, record, and export through the platform backend.
- Routes worker-side producer-event export through the platform backend.
- Generalizes `CUDAMessagingFuture` to `DeviceMessagingFuture` while preserving the `CUDAMessagingFuture` and `to_cuda_future()` compatibility aliases.
- Ensures retrieve operations wait for the worker producer event before writing KV-cache blocks.
- Records completion events before exporting their handles.
- Preserves the existing message wire format.
- Binds default event backends to the torch module declared by the requested `DeviceSpec`, rather than whichever accelerator is globally active.

This removes backend-specific event handling from the LMCache-driven multiprocess path. Additional accelerators can provide safe event IPC implementations without adding device-specific imports or branches to generic multiprocess code.

## Special notes for reviewers

- MUSA support remains explicitly opt-in and capability-gated.
- MUSA fails closed unless handle transfer and the required TorchMUSA interprocess `Event` API are available.
- The MUSA backend follows the latest `dev` platform structure and uses `platform/musa/ipc_wrapper.py`.
- The implementation no longer depends on the removed native `musa_aiter` event helper interface or Stage4 native event helpers.
- Generic LMCache-driven multiprocess modules do not import MUSA packages or branch on the MUSA device type.
- Worker-side event construction and initial recording in integration adapters remain unchanged.
- Focused validation completed with 96 tests passing and 13 platform-specific tests skipped.
- Ruff, formatting, isort, codespell, SPDX checks, clang-format, and mypy pass.

## If applicable

- [x] This PR contains user-facing changes — docs added
- [x] This PR contains unit tests

### DIFF
--- a/lmcache/v1/multiprocess/futures.py
+++ b/lmcache/v1/multiprocess/futures.py
@@ -4,8 +4,10 @@ from typing import Any, Generic, Optional, TypeVar
 import threading
 
 # First Party
-from lmcache import torch_dev, torch_device_type
+from lmcache import torch_dev
+from lmcache.utils import lmcache_deprecate
 from lmcache.v1.mp_observability.errors import LMCacheTimeoutError
+from lmcache.v1.platform.base.event_ipc import get_event_ipc_backend
 
 T = TypeVar("T")
 
@@ -68,21 +70,45 @@ class MessagingFuture(Generic[T]):
         self.result_ = result
         self.is_done_.set()
 
+    def to_device_future(
+        self,
+        device: Any | None = None,
+    ) -> "DeviceMessagingFuture":
+        """Wrap this future in a device-aware future.
+
+        Args:
+            device: The device whose event backend orders completion. Defaults
+                to the active device.
+
+        Returns:
+            A DeviceMessagingFuture pending on both this future and the event.
+        """
+        # TODO: need extra type checking for the future type
+        return DeviceMessagingFuture.FromMessagingFuture(self, device)  # type: ignore
+
+    @lmcache_deprecate("Use to_device_future() instead")
     def to_cuda_future(
         self,
         device: Any | None = None,
-    ) -> "CUDAMessagingFuture":
-        # TODO: need extra type checking for the future type
-        return CUDAMessagingFuture.FromMessagingFuture(self, device)  # type: ignore
+    ) -> "DeviceMessagingFuture[T]":
+        """Return a device-aware future using the deprecated CUDA name.
+
+        Args:
+            device: Device on which the completion event will be imported.
+
+        Returns:
+            A device-aware future wrapping this messaging future.
+        """
+        return self.to_device_future(device)
 
 
-class CUDAMessagingFuture(MessagingFuture[T]):
+class DeviceMessagingFuture(MessagingFuture[T]):
     """
-    The future class that wraps both result and a CUDA IPC event.
-    The `query`, `wait`, and `result` methods will pend on both the
-    original future and the CUDA event.
+    The future class that wraps both a result and a device IPC event.
+    The `query`, `wait`, and `result` methods pend on both the original
+    future and the device event, ordered through the platform event backend.
     The original future should return tuple[bytes, T], where the first
-    element is the serialized CUDA event.
+    element is the serialized device event handle.
     """
 
     def __init__(
@@ -95,46 +121,36 @@ class CUDAMessagingFuture(MessagingFuture[T]):
         self.event_: Any | None = None
         self.result_: T | None = None
         self.device_ = device if device is not None else torch_dev.current_device()
+        self._event_backend = get_event_ipc_backend(self.device_)
+        self._event_backend.check_event_support(self.device_)
 
-    def _on_raw_future_complete(self):
+    def _on_raw_future_complete(self) -> None:
         """
-        Update the CUDA event and result when the raw future is complete.
+        Update the device event and result when the raw future is complete.
         """
         event_bytes, result = self.raw_future_.result()
         self.result_ = result
 
-        # Not all backends support interprocess Events (CUDA IPC specific)
-        if not hasattr(torch_dev, "Event") or not hasattr(
-            torch_dev.Event, "from_ipc_handle"
-        ):
-            raise RuntimeError(
-                f"Backend '{torch_device_type}' does not support interprocess "
-                "Events (Event.from_ipc_handle not available). "
-                "Multiprocess IPC requires CUDA."
-            )
-        self.event_ = torch_dev.Event.from_ipc_handle(self.device_, event_bytes)
+        self.event_ = self._event_backend.import_event(event_bytes, self.device_)
 
     def wait(self, timeout: Optional[float] = None) -> bool:
         """
-        Wait for the future to be done, with the CUDA stream.
+        Wait for the future to be done, ordered through the device event.
 
         Args:
             timeout (Optional[float]): Maximum time to wait for the UNDERLYING
                 RAW FUTURE in seconds. The exact timeout is not guaranteed
-                when waiting on the CUDA event. (NOTE: this could be improved
+                when waiting on the device event. (NOTE: this could be improved
                 with careful threading management)
 
         Returns:
             bool: True if the future is done, False if the timeout was reached.
 
-        Raises:
-            ValueError: if the timeout is not None.
-
         Notes:
             This function does not support waiting for a specific time.
         """
         if self.event_:
-            self.event_.synchronize()
+            self._event_backend.synchronize_event(self.event_, self.device_)
             return True
 
         flag = self.raw_future_.wait(timeout)
@@ -144,7 +160,7 @@ class CUDAMessagingFuture(MessagingFuture[T]):
         self._on_raw_future_complete()
 
         assert self.event_ is not None
-        self.event_.synchronize()
+        self._event_backend.synchronize_event(self.event_, self.device_)
 
         return True
 
@@ -155,7 +171,7 @@ class CUDAMessagingFuture(MessagingFuture[T]):
         Args:
             timeout (Optional[float]): Maximum time to wait for the UNDERLYING
                 RAW FUTURE in seconds. The exact timeout is not guaranteed
-                when waiting on the CUDA event. (NOTE: this could be improved
+                when waiting on the device event. (NOTE: this could be improved
                 with careful threading management)
 
         Returns:
@@ -167,7 +183,7 @@ class CUDAMessagingFuture(MessagingFuture[T]):
         flag = self.wait(timeout)
         if not flag:
             raise LMCacheTimeoutError(
-                "CUDAMessagingFuture result not available within timeout"
+                "DeviceMessagingFuture result not available within timeout"
             )
 
         assert self.result_ is not None
@@ -181,23 +197,27 @@ class CUDAMessagingFuture(MessagingFuture[T]):
             bool: True if the future is done, False otherwise.
         """
         if self.event_:
-            return self.event_.query()
+            return self._event_backend.query_event(self.event_)
 
         if self.raw_future_.query():
             self._on_raw_future_complete()
             assert self.event_ is not None
-            return self.event_.query()
+            return self._event_backend.query_event(self.event_)
 
         return False
 
     def set_result(self, result: T) -> None:
         raise NotImplementedError(
-            "CUDAMessagingFuture does not support set_result directly"
+            "DeviceMessagingFuture does not support set_result directly"
         )
 
     @staticmethod
     def FromMessagingFuture(
         raw_future: MessagingFuture[tuple[bytes, T]],
         device: Any | None = None,
-    ) -> "CUDAMessagingFuture[T]":
-        return CUDAMessagingFuture(raw_future, device)
+    ) -> "DeviceMessagingFuture[T]":
+        return DeviceMessagingFuture(raw_future, device)
+
+
+# Backward-compatible alias for existing imports.
+CUDAMessagingFuture = DeviceMessagingFuture
--- a/lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
+++ b/lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
@@ -12,12 +12,11 @@ import time
 import torch
 
 # First Party
-from lmcache import torch_dev, torch_device_type
+from lmcache import torch_dev
 from lmcache.logging import init_logger
 from lmcache.utils import (
     EngineType,
     _lmcache_nvtx_annotate,
-    check_interprocess_event_support,
 )
 from lmcache.v1.distributed.api import (
     MemoryLayoutDesc,
@@ -49,6 +48,10 @@ from lmcache.v1.multiprocess.native_completion import (
 )
 from lmcache.v1.multiprocess.protocols.base import RequestType
 from lmcache.v1.platform.base.cache_context import BaseCacheContext
+from lmcache.v1.platform.base.event_ipc import (
+    EventIPCBackend,
+    get_event_ipc_backend,
+)
 from lmcache.v1.platform.cache_context import create_cache_context
 import lmcache.c_ops as lmc_ops
 
@@ -594,6 +597,7 @@ class ContextEntry:
         has_liveness_signal: True once the instance has sent at least one
             PING. Selects the reap window (timeout vs registration grace).
             Latched only by PING, never by traffic.
+        event_backend: Cached event backend selected for this context's device.
     """
 
     cache_context: BaseCacheContext
@@ -601,6 +605,7 @@ class ContextEntry:
     world_size: int
     last_seen: float = 0.0
     has_liveness_signal: bool = False
+    event_backend: EventIPCBackend | None = None
 
 
 class LMCacheDrivenTransferModule(InstanceLivenessTarget):
@@ -761,7 +766,7 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
         torch_dev.empty_cache()
         ipc_collect = getattr(torch_dev, "ipc_collect", None)
         if ipc_collect is not None:
-            # Non-CUDA device modules (xpu / musa) do not expose ipc_collect.
+            # Backends without IPC collection omit this optional operation.
             ipc_collect()
 
     def get_handlers(self) -> list[HandlerSpec]:
@@ -879,6 +884,8 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
             separate_object_groups=self._ctx.separate_object_groups,
             full_sw_kv=self._ctx.full_sw_kv,
         )
+        event_backend = get_event_ipc_backend(cache_context.device)
+        event_backend.check_event_support(cache_context.device)
         layout_desc = get_layout_desc(
             cache_context, self._ctx.chunk_size, object_group_id=0
         )
@@ -895,6 +902,7 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                 world_size=world_size,
                 last_seen=now,
                 has_liveness_signal=False,
+                event_backend=event_backend,
             )
 
         logger.info(
@@ -970,6 +978,9 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
             raise ValueError(f"No GPU context registered for instance ID {instance_id}")
         cache_context = entry.cache_context
         model_name = entry.model_name
+        event_backend = entry.event_backend
+        if event_backend is None:
+            raise RuntimeError("Registered cache context has no event backend")
 
         num_object_groups = cache_context.kv_layer_groups_manager.num_object_groups
         obj_keys_per_obj_group = self._ctx.resolve_obj_keys(
@@ -991,8 +1002,7 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
             torch_dev.device(cache_context.device),
             torch_dev.stream(cache_context.stream),
         ):
-            check_interprocess_event_support()
-            event = torch_dev.Event(interprocess=True)
+            event = event_backend.create_event(cache_context.device)
 
             # Fail closed: every LMCache group must have block IDs covering all
             # chunks. A short list (e.g. a caller/protocol bug) would otherwise
@@ -1015,23 +1025,17 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                     num_chunks,
                     blocks_per_chunk,
                 )
-                event.record()
-                return event.ipc_handle(), False
+                event_backend.record_event(event, cache_context.stream)
+                return event_backend.export_event(event, cache_context.device), False
 
             block_ids_per_group_gpu = downsample_and_stage_block_ids(
                 cache_context, gpu_block_ids
             )
 
-            if not hasattr(torch_dev.Event, "from_ipc_handle"):
-                raise RuntimeError(
-                    f"Backend '{torch_device_type}' does not support IPC event "
-                    "handles (Event.from_ipc_handle not available). "
-                    "Multiprocess IPC requires CUDA."
-                )
-            vllm_event = torch_dev.Event.from_ipc_handle(
-                cache_context.device, event_ipc_handle
+            producer_event = event_backend.import_event(
+                event_ipc_handle, cache_context.device
             )
-            vllm_event.wait(stream=cache_context.stream)
+            event_backend.wait_event(producer_event, cache_context.stream)
 
             # CPU-synchronous sentinel: a GPU store is about to be enqueued.
             # Must be published via publish() (not publish_on_stream) so the
@@ -1098,9 +1102,8 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                 store_succeeded = True
             except Exception:
                 logger.exception("Cannot store keys due to exception")
-                return event.ipc_handle(), False
             finally:
-                event.record()
+                event_backend.record_event(event, cache_context.stream)
                 # Fail closed: commit the reserved objects only when every chunk
                 # copied successfully; otherwise the whole store is skipped.
                 stored_count = len(all_dict) if store_succeeded else 0
@@ -1136,7 +1139,10 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                 num_chunks * self._ctx.chunk_size,
                 ed - st,
             )
-        return event.ipc_handle(), True
+        return (
+            event_backend.export_event(event, cache_context.device),
+            store_succeeded,
+        )
 
     @_lmcache_nvtx_annotate
     def retrieve(
@@ -1168,6 +1174,7 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
 
         Raises:
             ValueError: If no GPU context is registered for the given instance ID.
+            RuntimeError: If the backend does not support IPC event handles.
         """
         st = time.perf_counter()
 
@@ -1176,6 +1183,9 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
             raise ValueError(f"No GPU context registered for instance ID {instance_id}")
         cache_context = entry.cache_context
         model_name = entry.model_name
+        event_backend = entry.event_backend
+        if event_backend is None:
+            raise RuntimeError("Registered cache context has no event backend")
 
         num_object_groups = cache_context.kv_layer_groups_manager.num_object_groups
         obj_keys_per_obj_group = self._ctx.resolve_obj_keys(
@@ -1218,8 +1228,7 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
             torch_dev.device(cache_context.device),
             torch_dev.stream(cache_context.stream),
         ):
-            check_interprocess_event_support()
-            event = torch_dev.Event(interprocess=True)
+            event = event_backend.create_event(cache_context.device)
 
             # Fail closed: a short block-id list would drive the transfer
             # kernel to write out-of-bounds GPU memory. Checked on the raw
@@ -1240,16 +1249,21 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                     num_chunks,
                     blocks_per_chunk,
                 )
-                event.record()
-                return event.ipc_handle(), False
+                event_backend.record_event(event, cache_context.stream)
+                return event_backend.export_event(event, cache_context.device), False
 
             # Cut and stage all block_ids to GPU once before the transfer
             block_ids_per_group_gpu = downsample_and_stage_block_ids(
                 cache_context, gpu_block_ids
             )
+            producer_event = event_backend.import_event(
+                event_ipc_handle, cache_context.device
+            )
+            event_backend.wait_event(producer_event, cache_context.stream)
 
             prefetched_keys: list[ObjectKey] = []
             total_bytes = 0
+            retrieve_succeeded = True
             try:
                 for obj_group_id in range(num_object_groups):
                     obj_keys = obj_keys_per_obj_group[obj_group_id]
@@ -1258,7 +1272,8 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                     ) as memory_objs:
                         if not memory_objs or len(memory_objs) != len(obj_keys):
                             logger.error("Some keys not found during retrieve!")
-                            return event.ipc_handle(), False
+                            retrieve_succeeded = False
+                            break
 
                         total_bytes += sum(mo.get_size() for mo in memory_objs)
 
@@ -1277,9 +1292,9 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                         prefetched_keys.extend(obj_keys)
             except Exception:
                 logger.exception("Cannot retrieve keys due to exception")
-                return event.ipc_handle(), False
+                retrieve_succeeded = False
             finally:
-                event.record()
+                event_backend.record_event(event, cache_context.stream)
                 if prefetched_keys:
                     submit_callback_to_stream(
                         cache_context.cupy_stream,
@@ -1307,12 +1322,16 @@ class LMCacheDrivenTransferModule(InstanceLivenessTarget):
                         },
                     ),
                 )
-        tokens_retrieved = num_chunks * self._ctx.chunk_size
-        ed = time.perf_counter()
-        logger.info(
-            "Retrieved %d tokens in %.3f seconds",
-            tokens_retrieved,
-            ed - st,
-        )
+        if retrieve_succeeded:
+            tokens_retrieved = num_chunks * self._ctx.chunk_size
+            ed = time.perf_counter()
+            logger.info(
+                "Retrieved %d tokens in %.3f seconds",
+                tokens_retrieved,
+                ed - st,
+            )
 
-        return event.ipc_handle(), True
+        return (
+            event_backend.export_event(event, cache_context.device),
+            retrieve_succeeded,
+        )
--- a/lmcache/v1/multiprocess/transfer_context/worker_transfer.py
+++ b/lmcache/v1/multiprocess/transfer_context/worker_transfer.py
@@ -31,6 +31,10 @@ from lmcache.v1.multiprocess.transfer_context.base import (
     scatter_cpu_to_paged_kv,
 )
 from lmcache.v1.platform import get_device_spec, resolve_kv_wrapper_factory
+from lmcache.v1.platform.base.event_ipc import (
+    EventIPCBackend,
+    get_event_ipc_backend,
+)
 
 logger = init_logger(__name__)
 
@@ -160,10 +164,7 @@ def _build_lmcache_driven_context(device_type: str) -> "TransferContext":
 
 
 class IPCEvent(Protocol):
-    """Protocol for IPC-capable CUDA events used by transport operations."""
-
-    def ipc_handle(self) -> object:
-        """Return an IPC handle consumable by the multiprocess server."""
+    """Protocol for device events used by transport operations."""
 
     def wait(self, stream: object | None = None) -> None:
         """Make ``stream`` wait for this event (async ordering primitive)."""
@@ -181,12 +182,29 @@ def _single_group_block_ids(block_ids: list[list[int]]) -> list[int]:
     return block_ids[0]
 
 
+def _get_kv_device(kv_caches: dict[str, torch.Tensor]) -> torch.device:
+    """Return the device shared by a non-empty KV-cache mapping.
+
+    Args:
+        kv_caches: Worker KV-cache tensors keyed by layer name.
+
+    Returns:
+        The device of the first KV-cache tensor.
+
+    Raises:
+        ValueError: If ``kv_caches`` is empty.
+    """
+    if not kv_caches:
+        raise ValueError("LMCache-driven transfer requires at least one KV cache")
+    return next(iter(kv_caches.values())).device
+
+
 class TransferContext(ABC):
     """Abstract transport layer for worker-side KV transfer.
 
     Concrete implementations encapsulate how worker-side store/retrieve
-    operations are transmitted to the multiprocess server. CUDA paths return
-    CUDA-aware futures backed by MQ requests, while CPU paths may perform
+    operations are transmitted to the multiprocess server. Device-handle paths
+    return event-aware futures backed by MQ requests, while CPU paths may perform
     gather/scatter synchronously and return already-resolved futures.
     """
 
@@ -194,7 +212,7 @@ class TransferContext(ABC):
     def register(
         self,
         instance_id: int,
-        kv_caches: dict[str, torch.Tensor],
+        _kv_caches: dict[str, torch.Tensor],
         model_name: str,
         world_size: int,
         blocks_in_chunk: int,
@@ -304,14 +322,16 @@ class TransferContext(ABC):
 class LMCacheDrivenTransferContext(TransferContext):
     """LMCache-driven IPC + MQ future transport context.
 
-    In this mode the serving engine provides device handles (IPC for CUDA,
-    SHM wrappers for CPU with CUDA-IPC-like semantics) and the LMCache
-    server performs direct device-side data transfer.
+    In this mode the serving engine provides device handles (accelerator IPC,
+    or SHM wrappers for CPU with IPC-like semantics) and the LMCache server
+    performs direct device-side data transfer.
     """
 
     def __init__(self) -> None:
         self._mq_client: MessageQueueClient | None = None
         self._send_request: SendRequest | None = None
+        self._device: torch.device | None = None
+        self._event_backend: EventIPCBackend | None = None
 
     def register(
         self,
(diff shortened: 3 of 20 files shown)


lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,185 @@
...
+        device: A ``torch.device``-like object, a device-type string, or None.
+            When None, the active ``torch_device_type`` is used.
+
+    Returns:
+        The registered ``EventIPCBackend`` for the resolved device type; the
+        CUDA-style default when no ``DeviceSpec`` is registered.
+    """
+    device_type = _resolve_device_type(device)
+    # First Party
+    from lmcache.v1.platform import get_device_spec
```
maobaolong, line 265: Why we add the import block within the body of methods, rather than the head of this file?

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,185 @@
...
+    from lmcache.v1.platform.base_device_spec import DeviceSpec
+
+    spec = get_device_spec(device_type)
+    if spec is None:
+        spec = DeviceSpec()
```
maobaolong: It looks like every time this function is called, a new `DeviceSpec` instance might be created. Is that what you intended?

lmcache/v1/platform/musa/__init__.py
```diff
@@ -39,3 +39,15 @@ def is_handle_transfer_available(self) -> bool:
         from lmcache.v1.platform.musa.ipc import is_musa_handle_transfer_available
 
         return is_musa_handle_transfer_available()
+
+    @property
+    def event_ipc_backend(self):
```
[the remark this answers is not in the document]
superleo: resolved

lmcache/v1/multiprocess/futures.py
```diff
@@ -95,46 +108,36 @@ def __init__(
         self.event_: Any | None = None
         self.result_: T | None = None
         self.device_ = device if device is not None else torch_dev.current_device()
+        self._event_ops = get_event_ipc_backend(self.device_)
 
     def _on_raw_future_complete(self):
```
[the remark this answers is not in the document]
superleo: resolved

lmcache/v1/platform/base/device_spec.py
```diff
@@ -42,6 +49,9 @@ class DeviceSpec:
     # Cached pin-memory backend instance (lazy-initialized).
     _pin_backend_cache: PinMemoryBackend | None = None
 
+    # Cached event IPC backend instance (lazy-initialized).
+    _event_backend_cache: "EventIPCBackend | None" = None
```
maobaolong: Why the `_event_backend_cache` define here looks different with the above `_pin_backend_cache`?

maobaolong:
@superleo Thanks for this great refactor, this PR looks good overall, just left some minor nits. @chunxiaozheng just for heads-up, is this the PR you were planning to work on? Superleo has already submitted this PR now, so you might want to check it as well.

lmcache/v1/platform/event_ipc.py
superleo: thanks baolong, the import intentionally remains local. Platform subclass discovery can import this module while platform.__init__ is still initializing, before get_device_spec is defined. Moving it to module scope could cause a circular-initialization failure. I added a comment explaining this.

lmcache/v1/platform/event_ipc.py
superleo: thanks baolong, and good catch. Recreating the fallback was not intended and defeated backend caching. will modified to reuses a shared fallback backend, with a regression test covering repeated lookups.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -948,6 +948,8 @@ def store(
             raise ValueError(f"No GPU context registered for instance ID {instance_id}")
         cache_context = entry.cache_context
         model_name = entry.model_name
+        event_ops = get_event_ipc_backend(cache_context.device)
```
maobaolong: Why this named `event_ops`? The return value of `get_event_ipc_backend` is an `EventIPCBackend` class.

maobaolong:
@hlin99  Do you want to take a look at this too?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: fixed, Renamed event_ops to event_backend

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,185 @@
...
+        The device-type string; the active ``torch_device_type`` when None.
+    """
+    if device is None:
+        return torch_device_type
+    device_type = getattr(device, "type", None)
+    if isinstance(device_type, str):
+        return device_type
+    if isinstance(device, str):
+        return device
+    return str(device)
```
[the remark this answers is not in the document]
superleo: Fixed integer device indices and indexed strings such as "cuda:0"/"musa:0"

lmcache/v1/platform/base/device_spec.py
superleo: thanks baolong, pinMemoryBackend is a runtime import, EventIPCBackend is type-only to avoid circular imports, so it uses a quoted forward reference. We added an inline comment documenting this.

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,219 @@
...
+    device_type: str
+
+    def check_event_support(self, device: object) -> None: ...
+    def create_event(self, device: object) -> object: ...
+    def export_event(self, event: object, device: object) -> bytes: ...
+    def import_event(self, handle: bytes, device: object) -> object: ...
+    def record_event(self, event: object, stream: object) -> None: ...
+    def wait_event(self, event: object, stream: object) -> None: ...
+    def query_event(self, event: object) -> bool: ...
+    def synchronize_event(self, event: object, device: object) -> None: ...
```
ApostaC: We need docstrings for the functions.

Also, is it possible to avoid the use of `object` type?

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,219 @@
...
+        return torch_device_type
+    device_type = getattr(device, "type", None)
+    if isinstance(device_type, str):
+        return device_type
+    if isinstance(device, str):
+        return device.split(":", maxsplit=1)[0]
+    return str(device)
+
+
+def get_event_ipc_backend(device: object | None = None) -> EventIPCBackend:
```
ApostaC: nit: is it possible to avoid having `object | None = None`.
If it makes sense to pass in a device, we should require it instead of having a None as default parameter.

(similar for other helper functions)

lmcache/v1/platform/event_ipc.py
ApostaC, line 1: nit: please reorganize the helper functions a bit.

Following the common practice, the layout of the source file should be:
- imports
- private helper functions (starting with `_`) and private helper classes
- Public classes (that will be used by other modules)
  - `__init__`
  - public functions
  - private functions and helper functions starting with `_`.
- Main functions (that will be used by other modules)

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,219 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Platform device-event IPC abstraction.
+
+The ``lmcache_driven`` multiprocess handle path orders cross-process KV-cache
+transfers with device events. Accelerators expose different event IPC ABIs
+(CUDA: ``torch.cuda.Event`` interprocess handles; MUSA: native ``musa_aiter``
+symbols). This module defines a small backend-neutral interface so generic MP
```
ApostaC: nit: we can remove the sentence: `(CUDA: ``torch.cuda.Event`` interprocess handles; MUSA: native ``musa_aiter``
symbols).`

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,219 @@
...
+            return True
+    return False
+
+
+class DefaultEventIPCBackend:
+    """CUDA-style event IPC backend over an injectable event module.
+
+    Works for any backend whose ``Event`` class supports interprocess handles
+    (CUDA, and the ``StubEvent`` used in CPU-only test environments). MUSA
+    overrides this with a native implementation.
```
ApostaC: no need to mention MUSA since it's not related here.

lmcache/v1/platform/base/device_spec.py
```diff
@@ -42,6 +49,10 @@ class DeviceSpec:
     # Cached pin-memory backend instance (lazy-initialized).
     _pin_backend_cache: PinMemoryBackend | None = None
 
+    # EventIPCBackend is imported only under TYPE_CHECKING to avoid a runtime
+    # cycle with event_ipc's platform-registry lookup, hence the forward ref.
+    _event_backend_cache: "EventIPCBackend | None" = None
```
ApostaC: Why is `_event_backend_cache` optional? Also, why cannot it to be initialized when creating the `DeviceSpec`?

lmcache/v1/multiprocess/futures.py
```diff
@@ -68,21 +69,33 @@ def set_result(self, result: T) -> None:
...
+
+        Returns:
+            A DeviceMessagingFuture pending on both this future and the event.
+        """
         # TODO: need extra type checking for the future type
-        return CUDAMessagingFuture.FromMessagingFuture(self, device)  # type: ignore
+        return DeviceMessagingFuture.FromMessagingFuture(self, device)  # type: ignore
+
+    # Backward-compatible alias (Goal 3).
+    to_cuda_future = to_device_future
```
ApostaC: To make it more clear and explicit, I suggest we keep the `def to_cuda_future` wrapped with `@lmcache_deprecate` from `lmcache.utils` module.

lmcache/v1/multiprocess/futures.py
```diff
@@ -181,23 +184,27 @@ def query(self) -> bool:
...
     def FromMessagingFuture(
         raw_future: MessagingFuture[tuple[bytes, T]],
         device: Any | None = None,
-    ) -> "CUDAMessagingFuture[T]":
-        return CUDAMessagingFuture(raw_future, device)
+    ) -> "DeviceMessagingFuture[T]":
+        return DeviceMessagingFuture(raw_future, device)
+
+
+# Backward-compatible alias for existing imports (Goal 3).
```
ApostaC: don't say `Goal 3` here because nobody will understand it 😄

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
```diff
@@ -352,43 +367,90 @@ def submit_store(
...
+            ValueError: If ``kv_caches`` is empty.
+        """
         if self._mq_client is None or self._send_request is None:
             raise RuntimeError(
                 "LMCache-driven transfer context is not registered. "
                 "Call register() before submit_store()."
             )
+        device = _get_kv_device(kv_caches)
+        event_ops = get_event_ipc_backend(device)
+        event_ops.check_event_support(device)
```
ApostaC: we should save device and event_ops in during `__init__` instead of calling it every time

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -948,6 +948,8 @@ def store(
             raise ValueError(f"No GPU context registered for instance ID {instance_id}")
         cache_context = entry.cache_context
         model_name = entry.model_name
+        event_backend = get_event_ipc_backend(cache_context.device)
+        event_backend.check_event_support(cache_context.device)
```
ApostaC: We should save the `event_backend` during start up.

Also, I do see in some classes, it's called `event_ops`. I suggest we keep a consistent naming. How about name the variables to `event_backend` in all places?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -1234,7 +1238,8 @@ def retrieve(
                     ) as memory_objs:
                         if not memory_objs or len(memory_objs) != len(obj_keys):
                             logger.error("Some keys not found during retrieve!")
-                            return event.ipc_handle(), False
+                            retrieve_succeeded = False
+                            break
```
ApostaC, line 1276: why do we introduce a `retrieve_succeed` variable here instead of following the old code (directly return) here.

ApostaC (changes requested):
Got some comments, mostly coding style related.

The design looks good to me.

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,219 @@
...
+        try:
+            signature = inspect.signature(obj)  # type: ignore[arg-type]
+        except (TypeError, ValueError):
+            continue
+        if "interprocess" in signature.parameters:
+            return True
+    return False
+
+
+class DefaultEventIPCBackend:
```
maobaolong, line 163: Do you think it would be more appropriate to put this implementation inside the cuda module?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: thanks yihua @ApostaC , if keep the old early return event.ipc_handle(), False inside the try. In Python, the return expression is evaluated before finally, so that would call export_event() before record_event().

superleo: Kept retrieve_succeeded because returning directly would export the handle before finally records the event

lmcache/v1/platform/event_ipc.py
superleo: thanks baolong @maobaolong , Kept the default implementation generic because both CUDA-style events and the CPU stub use it; MUSA retains its platform-specific implementation

lmcache/v1/platform/event_ipc.py
superleo: thanks yihua @ApostaC , added docstrings for all public EventIPCBackend methods and kept helpers private.

lmcache/v1/multiprocess/futures.py
superleo: Removed

lmcache/v1/platform/event_ipc.py
superleo: removed

lmcache/v1/platform/event_ipc.py
superleo: removed

lmcache/v1/multiprocess/futures.py
superleo: Replaced the alias with an explicitly deprecated to_cuda_future() wrapper and migrated internal callers

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
superleo: Cached the device and backend during register()

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: Cached the backend when each server context is registered, modified the variable name to event_backend throughout.

lmcache/v1/platform/base/device_spec.py
superleo: None is the lazy-initialization sentinel. Eager initialization would import and construct the event backend during platform discovery, it lead a circular import and unnecessarily initialize backends for inactive devices.

lmcache/v1/platform/event_ipc.py
superleo: thanks and good point. Reorganized event_ipc.py so private module-level helpers come first, followed by the public backend classes and the public get_event_ipc_backend() entry point

lmcache/v1/platform/event_ipc.py
superleo: thanks and agreed. The backend selection should require an explicit device. I’ll remove the None default from get_event_ipc_backend() and _resolve_device_type() and update the affected tests. Other optional parameters will remain only where the default has intentional semantics

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -1216,16 +1225,21 @@ def retrieve(
...
-                event.record()
-                return event.ipc_handle(), False
+                event_backend.record_event(event, cache_context.stream)
+                return event_backend.export_event(event, cache_context.device), False
 
             # Cut and stage all block_ids to GPU once before the transfer
             block_ids_per_group_gpu = downsample_and_stage_block_ids(
                 cache_context, gpu_block_ids
             )
+            producer_event = event_backend.import_event(
```
maobaolong, line 1259: This part doesn't seem to be functionally equivalent to how it was before the refactor. Could you explain the reasoning behind this?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -993,23 +1003,17 @@ def store(
                     num_chunks,
                     blocks_per_chunk,
                 )
-                event.record()
-                return event.ipc_handle(), False
+                event_backend.record_event(event, cache_context.stream)
```
maobaolong, line 1028: Why don't we use the event object to call this method `record` like we did before, instead of using event_backend and call the `record_event` method?

Is it possible to wrap the Event object during construction so that it functions like before? Specifically, the Stream would be a member of the Event object, and this new wrapped Event object would still expose the same APIs as the original, such as `event.record()` and `event.ipc_handle()`.

Would this approach be feasible?

lmcache/v1/multiprocess/futures.py
```diff
@@ -68,21 +70,45 @@ def set_result(self, result: T) -> None:
...
+
+        Returns:
+            A DeviceMessagingFuture pending on both this future and the event.
+        """
         # TODO: need extra type checking for the future type
-        return CUDAMessagingFuture.FromMessagingFuture(self, device)  # type: ignore
+        return DeviceMessagingFuture.FromMessagingFuture(self, device)  # type: ignore
+
+    @lmcache_deprecate("Use to_device_future() instead")
+    def to_cuda_future(
```
maobaolong, line 90: As this is a internal API, so is it possible to remove this api completely?

chunxiaozheng:
@superleo Thanks for your contribution, I have a minor question, with this PR, the IPC has a good abstraction, but the code in `platform/_registry.py`, such as `_discover_wrappers_once`, which is used to discover `ipc wrapper`, is not used in anywhere ideally? and we can delete it.

chunxiaozheng:
> @superleo Thanks for your contribution, I have a minor question, with this PR, the IPC has a good abstraction, but the code in `platform/_registry.py`, such as `_discover_wrappers_once`, which is used to discover `ipc wrapper`, is not used in anywhere ideally? and we can delete it.

Besides, in my understanding, we need refactor `ipc_wrapper` related code, but I didn't see any places where these were updated.

superleo:
> > @superleo Thanks for your contribution, I have a minor question, with this PR, the IPC has a good abstraction, but the code in `platform/_registry.py`, such as `_discover_wrappers_once`, which is used to discover `ipc wrapper`, is not used in anywhere ideally? and we can delete it.
> 
> Besides, in my understanding, we need refactor `ipc_wrapper` related code, but I didn't see any places where these were updated.

thanks chunxiao @chunxiaozheng , _registry / DeviceIPCWrapper are the memory-handle path and remain in use (vLLM adapter, worker_transfer). This PR only abstracts event IPC (EventIPCBackend).  Deleting _discover_wrappers_once would break KV handle discovery. Unifying wrappers onto DeviceSpec is a follow-up (we can see the TODO on DeviceSpec), not part of this change.

lmcache/v1/multiprocess/futures.py
superleo: thanks baolong, in-tree callers already use to_device_future() and to_cuda_future is only a deprecated alias (in design markdown doc - Goal 3 we decide to remain it in this PR).

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: We keep events + EventIPCBackend so MUSA can use native record/export without teaching MP about torch.musa.Event. Baking stream into the event also diverges from CUDA’s record(stream) model and conflicts with ADR-2 (in design doc, worker still builds raw events). Prefer keep backend-ops; wrapper can be a follow-up if desired.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: thanks baolong @maobaolong ,For store, this is a 1:1 rewrite of from_ipc_handle + wait onto EventIPCBackend (CUDA-equivalent).
For retrieve, you’re right it’s not equivalent: the old path took event_ipc_handle but never waited on it. This PR adds the missing producer wait so retrieve matches store and the design doc.

maobaolong:
> > @superleo Thanks for your contribution, I have a minor question, with this PR, the IPC has a good abstraction, but the code in `platform/_registry.py`, such as `_discover_wrappers_once`, which is used to discover `ipc wrapper`, is not used in anywhere ideally? and we can delete it.
> 
> Besides, in my understanding, we need refactor `ipc_wrapper` related code, but I didn't see any places where these were updated.

`ipc_wrapper` related code have been refactored and merged. @superleo

maobaolong:
@superleo Sorry there is another PR merged and affected your PR make it conflict, please resolve the conflict first.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: fixed, using event_backend and call record_event method.

superleo:
> @superleo Sorry there is another PR merged and affected your PR make it conflict, please resolve the conflict first.

@maobaolong  thanks baolong, the design document was updated to match the latest dev platform architecture and implementation:
- Defines a unified EventIPCBackend interface for event creation, import/export, recording, waiting, querying, and synchronization.
- Moves backend selection into DeviceSpec, keeping generic multiprocess code platform-neutral.
- Updates MUSA support to use ipc_wrapper.py and the TorchMUSA interprocess Event API instead of the removed native musa_aiter helpers.
- Documents the complete worker-server-future event flow, fail-closed behavior, compatibility guarantees, scope exclusions, and test strategy.

lmcache/v1/platform/base_device_spec.py
```diff
@@ -96,6 +101,41 @@ def is_handle_transfer_available(self) -> bool:
         # TODO(chunxiaozheng): implement on subclasses
         return True
 
+    @property
+    def event_ipc_backend(self) -> "EventIPCBackend":
```
maobaolong: As we have CpuDeviceSpec now. we should return None for this base class, and move this implementation to the `CpuDeviceSpec`, WDYT?

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    device_type = _resolve_device_type(device)
+    # Platform subclass discovery can import this module before
+    # platform.__init__ has defined get_device_spec, so this import must stay
+    # lazy to avoid a circular-initialization failure.
+    # First Party
+    from lmcache.v1.platform import get_device_spec
+
+    spec = get_device_spec(device_type)
+    if spec is None:
```
maobaolong, line 268: As we have a spec for each device, so the `spec` here should never be none, if so, we should raise an exception, we don't need to handle the fallback scenario, wdyt?

lmcache/v1/platform/base_device_spec.py
superleo: agreed, moved it.

lmcache/v1/platform/event_ipc.py
superleo: agreed

maobaolong:
@superleo Sorry https://github.com/LMCache/LMCache/pull/4184 has been merged just now, and your pr get conflict, please resolve the conflict again.

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,272 @@
+# SPDX-License-Identifier: Apache-2.0
```
maobaolong, line 1: As this is a base class, so it should be better to placed to `base` folder.

maobaolong:
@superleo It should helps if you update docs/source/developer_guide/extending_lmcache/adding_a_new_device_backend.rst in this PR.

lmcache/v1/platform/event_ipc.py
```diff
@@ -0,0 +1,272 @@
...
+    Args:
+        device: A ``torch.device``-like object (has ``.type``), a device-type
+            string, or an integer device index.
+
+    Returns:
+        The device-type string. Integer indices use the active
+        ``torch_device_type``.
+    """
+    if isinstance(device, int):
+        return torch_device_type
```
ApostaC, line 57: If we call `_resolve_device_type(2)`, what will it resolve to? Will it resolve to the same device as calling `_resolve_device_type(0)`?

lmcache/v1/platform/musa/event_ipc.py
```diff
@@ -0,0 +1,115 @@
...
+        get_torch_musa_module = getattr(self._ipc(), "get_torch_musa_module", None)
+        if not callable(get_torch_musa_module):
+            raise RuntimeError("MUSA event IPC requires the TorchMUSA module")
+        torch_musa = get_torch_musa_module()
+        if torch_musa is None:
+            raise RuntimeError("MUSA event IPC requires the TorchMUSA module")
+        return DefaultEventIPCBackend(
+            event_module=torch_musa,
+            device_type=self.device_type,
+        )
```
ApostaC, line 66: This function is on the host path. Should we consider only creating `DefaultEventIPCBackend` once during initialization and directly return it rather that creating it every time when this function is called?

lmcache/v1/platform/base/device_spec.py
```diff
@@ -56,6 +57,10 @@ class DeviceSpec:
     # Cached DeviceOps singleton instance (lazy-initialized).
     _ops_cache: DeviceOps | None = None
 
+    # EventIPCBackend is imported only under TYPE_CHECKING to avoid a runtime
+    # cycle with event_ipc's platform-registry lookup, hence the forward ref.
+    _event_backend_cache: "EventIPCBackend | None" = None
+
```
ApostaC: Since the abstract implementation for `event_ipc_backend` is None, we should move `_event_backend_cache` to all the child classes that need it (e.g., cuda and cpu). 
Defining it in the base class might be an "overkill".

lmcache/v1/multiprocess/futures.py
```diff
@@ -95,46 +121,36 @@ def __init__(
...
-        if not hasattr(torch_dev, "Event") or not hasattr(
-            torch_dev.Event, "from_ipc_handle"
-        ):
-            raise RuntimeError(
-                f"Backend '{torch_device_type}' does not support interprocess "
-                "Events (Event.from_ipc_handle not available). "
-                "Multiprocess IPC requires CUDA."
-            )
-        self.event_ = torch_dev.Event.from_ipc_handle(self.device_, event_bytes)
+        self._event_backend.check_event_support(self.device_)
```
ApostaC: Ideally, `check_event_support` should be done during initialization. We should avoid calling it in the hot path.

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
ApostaC, line 1: I see it adds a lot of new docstrings (which is a bit out of this PR's scope).

I'm not against this, but would like @hlin99 to take a look at whether the docstrings here are accurate.

ApostaC (commented):
Mostly nit comments. Otherwise LGTM!

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
ApostaC: @superleo circling back to this question. Do you have any reference showing that getting the ipc handle before `record` will lead to errors? 
AFAIK, it should be okay in CUDA. (maybe I'm wrong, happy to learn here!)

lmcache/v1/platform/event_ipc.py
superleo: agreed, will move it

lmcache/v1/platform/event_ipc.py
superleo: yes, both 0/2 resolve to same device type. Will add docstring

lmcache/v1/platform/musa/event_ipc.py
superleo: Good catch, will cache it.

lmcache/v1/platform/base/device_spec.py
superleo: ok, will move it.

lmcache/v1/multiprocess/futures.py
superleo: ok

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
superleo: @ApostaC  thanks yihua, you are right and  it's ok in cuda and recent version of musa, both support exporting ipc handle before recording the event. We kept this change because older torchmusa(opensource project) versions showed inconsistent behavior for unrecorded event handles on these early-return paths. Recording before export gives all backends the same completion-event ordering, without changing the wire format. The change only affects the failure path, where we otherwise export the handle before the finally block records the event.

lmcache/v1/platform/base/device_spec.py
superleo: fixed

lmcache/v1/platform/musa/event_ipc.py
superleo: fixed, default adapter is now lazy created once and cached.  Registration initializes it through check_event_support()

lmcache/v1/platform/event_ipc.py
superleo: added docstring

maobaolong:
@superleo Not sure the failed UT is related?

This is the detailed log.
```shell

tests/v1/test_torch_ops.py::TestScenarios::test_1_scenario[cuda_c_ops-record_drain_event-scenario_record_drain_event] FAILED                             [ 97%]
--
  |  
  | =========================================================================== FAILURES ===========================================================================
  | ___________________________________ TestScenarios.test_1_scenario[cuda_c_ops-record_drain_event-scenario_record_drain_event] ___________________________________
  |  
  | self = <tests.v1.test_torch_ops.TestScenarios object at 0x7fe8ceb3c110>, backend = ('cuda_c_ops', <module 'lmcache.c_ops'>, 'cuda'), name = 'record_drain_event'
  | fn = <function scenario_record_drain_event at 0x7fe8cebccc20>
  |  
  | @pytest.mark.parametrize("name,fn", list(SCENARIO_REGISTRY.items()))
  | def test_1_scenario(
  | self,
  | backend: tuple,
  | name: str,
  | fn: Any,
  | ) -> None:
  | """Run a single scenario with a specific backend configuration.
  |  
  | Each (scenario, backend) pair is a separate pytest test case, giving
  | per-scenario per-backend failure reporting without subprocess indirection.
  |  
  | Results are stored in the module-level _results dict for later comparison
  | by test_2_compare.
  | """
  | backend_id, ops, device = backend
  | >       result = fn(ops, device)
  | ^^^^^^^^^^^^^^^
  |  
  | tests/v1/test_torch_ops.py:2852:
  | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  |  
  | ops = <module 'lmcache.c_ops'>, device = 'cuda'
  |  
  | def scenario_record_drain_event(ops: Any, device: str) -> dict[str, torch.Tensor]:
  | """Test record_event_on_stream / drain_recorded_events contracts.
  |  
  | Verified backend-agnostic: native c_ops uses cudaLaunchHostFunc on the
  | default stream (ptr=0), which fires synchronously after device_sync; the
  | fallback enqueues immediately with time.time(). Both paths satisfy every
  | assertion below.
  | """
  | ops.drain_recorded_events()  # clear residual global state
  |  
  | assert ops.drain_recorded_events() == []
  |  
  | ops.record_event_on_stream(
  | 0, "mp.store.start", "sess-1", {"device": "cuda:0"}, {"count": 10}
  | )
  | ops.record_event_on_stream(
  | 0, "mp.store.end", "sess-1", {"device": "cuda:0"}, {"count": 10}
  | )
  | device_sync(device)
  | result = ops.drain_recorded_events()
  | >       assert len(result) == 2
  | E       assert 0 == 2
  | E        +  where 0 = len([])
  |  
  | tests/v1/test_torch_ops.py:2755: AssertionError


```

superleo:
> @superleo Not sure the failed UT is related?
> 
> This is the detailed log.
> 
> ```shell
> tests/v1/test_torch_ops.py::TestScenarios::test_1_scenario[cuda_c_ops-record_drain_event-scenario_record_drain_event] FAILED                             [ 97%]
> --
>   |  
>   | =========================================================================== FAILURES ===========================================================================
>   | ___________________________________ TestScenarios.test_1_scenario[cuda_c_ops-record_drain_event-scenario_record_drain_event] ___________________________________
>   |  
>   | self = <tests.v1.test_torch_ops.TestScenarios object at 0x7fe8ceb3c110>, backend = ('cuda_c_ops', <module 'lmcache.c_ops'>, 'cuda'), name = 'record_drain_event'
>   | fn = <function scenario_record_drain_event at 0x7fe8cebccc20>
>   |  
>   | @pytest.mark.parametrize("name,fn", list(SCENARIO_REGISTRY.items()))
>   | def test_1_scenario(
>   | self,
>   | backend: tuple,
>   | name: str,
>   | fn: Any,
>   | ) -> None:
>   | """Run a single scenario with a specific backend configuration.
>   |  
>   | Each (scenario, backend) pair is a separate pytest test case, giving
>   | per-scenario per-backend failure reporting without subprocess indirection.
>   |  
>   | Results are stored in the module-level _results dict for later comparison
>   | by test_2_compare.
>   | """
>   | backend_id, ops, device = backend
>   | >       result = fn(ops, device)
>   | ^^^^^^^^^^^^^^^
>   |  
>   | tests/v1/test_torch_ops.py:2852:
>   | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
>   |  
>   | ops = <module 'lmcache.c_ops'>, device = 'cuda'
>   |  
>   | def scenario_record_drain_event(ops: Any, device: str) -> dict[str, torch.Tensor]:
>   | """Test record_event_on_stream / drain_recorded_events contracts.
>   |  
>   | Verified backend-agnostic: native c_ops uses cudaLaunchHostFunc on the
>   | default stream (ptr=0), which fires synchronously after device_sync; the
>   | fallback enqueues immediately with time.time(). Both paths satisfy every
>   | assertion below.
>   | """
>   | ops.drain_recorded_events()  # clear residual global state
>   |  
>   | assert ops.drain_recorded_events() == []
>   |  
>   | ops.record_event_on_stream(
>   | 0, "mp.store.start", "sess-1", {"device": "cuda:0"}, {"count": 10}
>   | )
>   | ops.record_event_on_stream(
>   | 0, "mp.store.end", "sess-1", {"device": "cuda:0"}, {"count": 10}
>   | )
>   | device_sync(device)
>   | result = ops.drain_recorded_events()
>   | >       assert len(result) == 2
>   | E       assert 0 == 2
>   | E        +  where 0 = len([])
>   |  
>   | tests/v1/test_torch_ops.py:2755: AssertionError
> ```

hi baolong, this PR didn't modified native c_ops.record_event_on_stream() / drain_recorded_events() path. this issue appears unrelated to the Event IPC abstraction, likely belongs to the native CUDA callback path

maobaolong:
> > @superleo Not sure the failed UT is related?
> > This is the detailed log.
> > ```shell
> > tests/v1/test_torch_ops.py::TestScenarios::test_1_scenario[cuda_c_ops-record_drain_event-scenario_record_drain_event] FAILED                             [ 97%]
> > --
> >   |  
> >   | =========================================================================== FAILURES ===========================================================================
> >   | ___________________________________ TestScenarios.test_1_scenario[cuda_c_ops-record_drain_event-scenario_record_drain_event] ___________________________________
> >   |  
> >   | self = <tests.v1.test_torch_ops.TestScenarios object at 0x7fe8ceb3c110>, backend = ('cuda_c_ops', <module 'lmcache.c_ops'>, 'cuda'), name = 'record_drain_event'
> >   | fn = <function scenario_record_drain_event at 0x7fe8cebccc20>
> >   |  
> >   | @pytest.mark.parametrize("name,fn", list(SCENARIO_REGISTRY.items()))
> >   | def test_1_scenario(
> >   | self,
> >   | backend: tuple,
> >   | name: str,
> >   | fn: Any,
> >   | ) -> None:
> >   | """Run a single scenario with a specific backend configuration.
> >   |  
> >   | Each (scenario, backend) pair is a separate pytest test case, giving
> >   | per-scenario per-backend failure reporting without subprocess indirection.
> >   |  
> >   | Results are stored in the module-level _results dict for later comparison
> >   | by test_2_compare.
> >   | """
> >   | backend_id, ops, device = backend
> >   | >       result = fn(ops, device)
> >   | ^^^^^^^^^^^^^^^
> >   |  
> >   | tests/v1/test_torch_ops.py:2852:
> >   | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> >   |  
> >   | ops = <module 'lmcache.c_ops'>, device = 'cuda'
> >   |  
> >   | def scenario_record_drain_event(ops: Any, device: str) -> dict[str, torch.Tensor]:
> >   | """Test record_event_on_stream / drain_recorded_events contracts.
> >   |  
> >   | Verified backend-agnostic: native c_ops uses cudaLaunchHostFunc on the
> >   | default stream (ptr=0), which fires synchronously after device_sync; the
> >   | fallback enqueues immediately with time.time(). Both paths satisfy every
> >   | assertion below.
> >   | """
> >   | ops.drain_recorded_events()  # clear residual global state
> >   |  
> >   | assert ops.drain_recorded_events() == []
> >   |  
> >   | ops.record_event_on_stream(
> >   | 0, "mp.store.start", "sess-1", {"device": "cuda:0"}, {"count": 10}
> >   | )
> >   | ops.record_event_on_stream(
> >   | 0, "mp.store.end", "sess-1", {"device": "cuda:0"}, {"count": 10}
> >   | )
> >   | device_sync(device)
> >   | result = ops.drain_recorded_events()
> >   | >       assert len(result) == 2
> >   | E       assert 0 == 2
> >   | E        +  where 0 = len([])
> >   |  
> >   | tests/v1/test_torch_ops.py:2755: AssertionError
> > ```
> 
> hi baolong, this PR didn't modified native c_ops.record_event_on_stream() / drain_recorded_events() path. this issue appears unrelated to the Event IPC abstraction, likely belongs to the native CUDA callback path

@superleo Thanks for the explanation. I've re-run the UT for you and it looks like it passed. It definitely doesn't have anything to do with your PR, it seems like it was just a flaky test.

We can look into what happened later on, but for now, it's clearly unrelated to this PR.

maobaolong (approved):
@superleo Thanks for this great refactor, LGTM!

superleo:
hi @maobaolong  @ApostaC , the current k3-multiprocess-test fail seems not related with this Patch.  The error log of cufile :  23-07-2026 05:51:26:823 [pid=1533 tid=1533] NOTICE  cufio-drv:851 running in compatible mode
 23-07-2026 05:51:32:784 [pid=1533 tid=1533] ERROR  0:140 unable to load,  libnuma.so.1.0.0 

 23-07-2026 05:51:32:784 [pid=1533 tid=1533] ERROR  0:98 dlopen error libnuma.so.1.0.0: cannot open shared object file: No such file or directory
 23-07-2026 05:51:33:114 [pid=1533 tid=1686] ERROR  0:58 NUMA not available
 23-07-2026 05:51:33:114 [pid=1533 tid=1687] ERROR  0:58 NUMA not available
 23-07-2026 05:51:33:114 [pid=1533 tid=1689] ERROR  0:58 NUMA not available
 23-07-2026 05:51:33:114 [pid=1533 tid=1688] ERROR  0:58 NUMA not available
 23-07-2026 05:51:33:194 [pid=1533 tid=1533] ERROR  cufio-fs:68 mount options not specified for device in mount table /dev/dm-0
 23-07-2026 05:53:01:878 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:891 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:910 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:926 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:939 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:952 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:961 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:970 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:979 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:01:992 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:3 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:12 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:22 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:29 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:39 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:47 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:55 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:60 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:67 [pid=1533 tid=1709] ERROR  0:58 NUMA not available
 23-07-2026 05:53:02:72 [pid=1533 tid=1709] ERROR  0:58 NUMA not available

lmcache/v1/multiprocess/transfer_context/worker_transfer.py
superleo: And we can modify it in a follow up PR (musa-4-2 or musa-4-3, or a seperate fix) if lin add new comments.

maobaolong:
@superleo I retry the failed test just now.

maobaolong:
Now it passed, i also found that the error log you motioned is appeared even it passed.

ApostaC (approved):
LGTM!

PR #3038, opened 2026-04-15
feat(pd_backend): fully async PD backend

hlin99:
PDBackend with async implementations:
- Non-blocking transfer: batched_submit_put_task returns immediately (fire-and-forget enqueue) instead of blocking vLLM worker thread until remote alloc + RDMA write complete
- Event-driven flow control: replace time.sleep busy-wait polling with Condition-based notification, waking immediately when resources are freed
- Self-contained resource release: remove() internally calls ref_count_down() and decrements inflight counter, eliminating caller responsibility for manual cleanup. cache_engine.py updated with _is_sync_pd_backend() guard to prevent double-free
- Startup capacity validation: new pd_max_prefill_len config raises ValueError at init if buffer cannot hold the max prefill length, catching misconfiguration before runtime
- Configurable timeouts: pd_allocation_timeout_sec, pd_shutdown_timeout_sec, pd_condition_poll_interval_sec replace scattered hardcoded constants
- Backward compatible: split into pd_backend.py (sync) and pd_backend_async.py (async), selectable per-instance via pd_backend_mode config (default: "async"). Sync and async instances can coexist in the same cluster — e.g. sender on async while receiver on sync, or vice versa — with no wire protocol incompatibility

5%+ performance gain by async @16K prompt len, and if prompt len is longer, perf gain would be bigger.
Async:
image

Sync:
image

---

> [!NOTE]
> **Medium Risk**
> Introduces a new asyncio-based PD backend and switches default PD mode to it, changing concurrency, allocation flow control, and request lifecycle handling. Risk is mainly around correctness of admission/rollback/abort paths and potential regressions when mixing `sync`/`async` modes or enforcing `total_chunks` requirements.
> 
> **Overview**
> Adds a new `PDBackendAsync` (selectable via `pd_backend_mode`, defaulting to `"async"`) that runs PD sender/receiver logic on dedicated asyncio event loops, making `batched_submit_put_task()` fire-and-forget and replacing busy-wait polling with condition-driven backpressure.
> 
> Implements **receiver-side reservation-based admission control** (`ReservationManager`) using `total_chunks` to prevent chunked-prefill deadlocks, adds fail-fast protocol checks/rollback, and introduces an abort path (`CancelNotif`) plus `StorageManager.cancel_request()` to clean up in-flight transfers on request abort.
> 
[...]

PR #2154, opened 2025-12-03
[Bugfix] AttributeError when using store-based KV sharing without disagg_spec (missing null check)

tianlang-wq:
This change fixes a potential issue in the disaggregation logic where the KV producer role might attempt to access request.disagg_spec without proper validation, which could cause issues when disagg_spec is None.

Key changes:
- Add additional check to ensure request.disagg_spec exists before accessing num_transferred_tokens property
- Prevents potential AttributeError when disagg_spec is None
- Maintains the original logic for calculating skip_leading_tokens when disagg_spec is properly configured

This ensures more robust handling of disaggregation scenarios in the KV cache transfer process.

related: https://github.com/LMCache/LMCache/issues/2145

PR #1827, opened 2025-10-20
[bugfix]: only update skip_leading_tokens on last PP rank in wait_for…

tianlang-wq:
🧠 Background / Motivation

After reviewing the changes in [#1813](https://github.com/LMCache/LMCache/pull/1813)
, I found that the issue does not originate from the lookup logic.
The lookup function only tracks how many tokens have been processed.
For new prompts, rank0 always calls wait_for_save() to store the KV cache.

However, in the current implementation, the following part at the end of wait_for_save():

save_spec.skip_leading_tokens = len(token_ids)
if request.disagg_spec:
    request.disagg_spec.num_transferred_tokens = len(token_ids)

is executed on rank0, which prematurely marks all token_ids as already saved.
As a result, rank1 (and later stages) will skip saving due to this check:

if skip_leading_tokens == len(token_ids):
    continue  # skip this request

This causes later pipeline stages to incorrectly skip KV cache saving.

🧰 What This PR Does

This PR updates the logic to ensure the save-state update only happens on the last PP rank:

if get_pp_group().is_last_rank:
    save_spec.skip_leading_tokens = len(token_ids)
    if request.disagg_spec:
        request.disagg_spec.num_transferred_tokens = len(token_ids)

✅ Impact

Fixes incorrect KV cache saving behavior under pipeline parallel (PP) mode.

Ensures each PP stage correctly saves its respective KV cache.

No behavioral change for single-rank or non-PP configurations.

🧪 Testing

Run inference with PP enabled.

Verify that:

Only the last PP rank updates skip_leading_tokens.

All PP ranks correctly save their portion of KV cache.

Confirm behavior remains identical in non-PP mode.

🔗 Related Issue

[#1807](https://github.com/LMCache/LMCache/issues/1807)

PR #4076, opened 2026-07-10
[Bugfix] Persist PD transfer progress across chunked-prefill steps

westnight:

Under chunked prefill, connector metadata flows scheduler → worker only.
`DisaggSpec.num_transferred_tokens` is advanced on the worker after a
successful PD store, but that mutation is not written back to the
scheduler's `RequestTracker`. The next chunk rebuilds `ReqMeta` with
`num_transferred_tokens=0`, so:

```python
skip = min(save_spec.skip_leading_tokens, disagg_spec.num_transferred_tokens)
```

becomes `min(N, 0) = 0` and the already-transferred prefix is stored/PD'd
again (or relies on decoder-side key dedup).

This PR keeps a worker-local `req_id → transferred` watermark, restores it
before the skip gate, and records it after a real store (last PP rank only,
same as today's progress advance). LocalCPU `skip_leading_tokens` is never
used to seed PD progress: a full local hit with zero peer progress must
still transfer.

- Agg / non-PD paths are unchanged (`disagg_spec is None`).
- Related but non-fixing prior work: #1827 (last-PP-rank advance), #2154
  (null guard), #3038 (async PD / `total_chunks`), #2972 (bidir probe).
- Happy to open a tracking issue if preferred before merge.

**If applicable**:


ApostaC:
Hi @westnight , thanks for your contribution! Do you have logs/screenshots + a way to reproduce the error? 

Also, we are moving to the MP mode deployment. Please take a look at our new doc: https://docs.lmcache.ai/mp/disaggregated_prefill.html

westnight:
@ApostaC I can reproduce this with Qwen3-8B, TP=2 (PP=1). TP=1 often does not show it, because scheduler/worker share the same in-process DisaggSpec and worker updates to num_transferred_tokens appear to stick.

LMCache config (sender / prefiller):

chunk_size: 256
local_cpu: False
enable_pd: True
transfer_channel: "nixl"
pd_role: "sender"
pd_proxy_host: "localhost"
pd_proxy_port: 7500
pd_buffer_size: 1073741824  # 1GB
pd_buffer_device: "cpu"
nixl_backends: [UCX]


Use a prompt long enough to force chunked prefill (e.g. ~1700 tokens with max_num_batched_tokens=256).

Optional debug logs (LMCACHE_DEBUG_PD_TRANSFER=1) in lmcache/integration/vllm/vllm_v1_adapter.py — observation only, not part of the fix:

`
@@ -1187,6 +1187,22 @@ class LMCacheConnectorV1Impl:

            if self.kv_role == "kv_producer" and request.disagg_spec:
                # Restore worker-local PD progress lost across metadata rebuilds.
                remembered = self._pd_transferred_tokens.get(request.req_id, 0)
                if remembered > request.disagg_spec.num_transferred_tokens:
                    request.disagg_spec.num_transferred_tokens = remembered
                skip_leading_tokens = min(
                    skip_leading_tokens, request.disagg_spec.num_transferred_tokens
                )
                transferred = request.disagg_spec.num_transferred_tokens
                local_skip = skip_leading_tokens
                skip_leading_tokens = min(local_skip, transferred)

                if os.environ.get("LMCACHE_DEBUG_PD_TRANSFER"):
                    logger.warning(
                        "[PD-REPRO] wait_for_save req=%s token_len=%d "
                        "local_skip=%d num_transferred=%d effective_skip=%d "
                        "is_last_prefill=%s",
                        request.req_id,
                        len(token_ids),
                        local_skip,
                        transferred,
                        skip_leading_tokens,
                        request.is_last_prefill,
                    )
`

Observed without the watermark fix (TP=2): each chunked-prefill step restarts PD from offset 0; num_transferred is lost across steps even though the worker advanced it on the previous step:

w/o watermark fix
fail_skip

with watermark fix
success_skip

PR #4078, opened 2026-07-10
[Fix][RawBlock] Rebuild free slots during checkpoint recovery

3xdevv:

Raw-block stores committed key metadata in `_index`, tracks in-flight writes separately in `_inflight`, and uses `_free_slots` only as a physical reusable-slot stack. 

Raw-block checkpoints used to persist both committed index entries and the allocator free list. That made recovery depend on a snapshot of `free_slots`, even though a crash-time checkpoint can observe allocator state after a slot reservation but before the corresponding in-flight write commits to `_index`. In that case the slot is below `next_slot`, absent from `entries`, and absent from `free_slots`, so it can be lost after restart, permanently.

This PR makes checkpoint recovery derive reusable slots from the durable committed state instead.

- Stops writing `free_slots` into new raw-block checkpoint snapshots.
- Rebuilds `_free_slots` during `RawBlockCore` recovery from `entries + next_slot`.
- Keeps older checkpoints loadable even when they still contain `free_slots`; recovery ignores that field.
- Adds a regression test for an older checkpoint shape with `next_slot` advanced, one committed entry, and empty `free_slots`, verifying the missing slot is reused after recovery.

- Suggested verification:

```bash
pytest -q tests/v1/storage_backend/test_raw_block_core.py
```

**If applicable**:


ApostaC:
Hey @DongDongJu , can you take a look at this PR? Thanks!

DongDongJu (approved):
LGTM

ApostaC (approved):

PR #4082, opened 2026-07-10
Support pipeline parallelism with MLA models in multi-server MP mode

hamstergamerszhang-ops:
## Pipeline parallelism + MLA: making it actually work

I've been running Kimi K2 and GLM-4.6 through vLLM with LMCache for a while now, and kept hitting the same wall: the moment you turn on pipeline parallelism with more than one LMCache server, the connector flat-out refuses to start. The error message says "only supports tensor parallelism (TP), not pipeline parallelism (PP)" — which is fair enough as a guard, but it meant I was stuck choosing between PP (which my hardware layout needs) and LMCache (which my workload needs).

So I dug into *why* the guard was there, and it turned out to be a really subtle locking bug that's fixable without redesigning the protocol. Here's what I found and what I changed.

### The actual problem

`compute_extra_count` decides how many read locks to acquire on each KV chunk in L1. For MLA models, all TP workers within a rank-shard share the same KV object, so you need `tp_size - 1` extra locks (one per additional reader). The old code inferred whether a model was MLA by checking `tp > world_size` — if the tensor-parallel size was bigger than the world size, it assumed MLA (because vLLM divides world_size by tp for MLA).

That heuristic works fine when PP=1. But the moment PP > 1, `world_size = tp × pp`, so the heuristic collapses. Concretely: TP=4, PP=2, 2 servers gives you wire `tp_size=2` and `world_size=2`, so `2 > 2` is false → `extra_count = 0` → the server locks 1 copy but 4 workers try to read it → premature eviction → silent KV corruption. The guard was there to stop you from hitting that.

### What I changed

Rather than inferring MLA, I made it explicit. `IPCCacheServerKey` now carries a `use_mla` flag — same backward-compatible msgspec field pattern as the existing `cache_salt` (old payloads without it decode as `False`, which is safe). The adapters set it from `ParallelStrategy.use_mla`, which already gets it from vLLM's `model_config.use_mla`.

With the flag explicit, `compute_extra_count` becomes trivially correct: `tp_size - 1 if use_mla else 0`. No guessing.

But while testing that, I found two more issues in `ParallelStrategy` that the PP guard was also papering over:

**`kv_tp_size`** was `tp_size // n_servers`, which assumes TP groups split evenly across servers. They don't — the connector assigns servers by contiguous rank blocks (`server_urls[rank // ranks_per_node]`), so under PP the server blocks align with pipeline stages and *all* `tp_size` TP ranks of a stage land on one server. The correct formula is `min(tp_size, ranks_per_node)`.

**`is_kv_writer`** had the same blind spot. It used `rank % (tp_size // n_servers)`, which picked multiple writers per shard under PP (double-stores) and zero writers on some servers when PP=1 and ns>1 (missing stores). The fix is `(rank % ranks_per_node) % tp_size == 0`, which gives exactly one writer per (server, pipeline-stage) pair.

I also cleaned up the other adapters while I was at it — the TensorRT-LLM adapter had `tp_size` hardcoded to `1` in its LOOKUP calls, and the SDK was sending `world_size` where it should have sent the reader count. Both now take the value as a parameter.

### A tip for anyone debugging this kind of thing

The thing that took me the longest was realizing that the comment above the PP guard said "Non-MLA mode is not affected by this restriction" — but the actual code checked `if n_servers > 1` with no MLA gate at all. The code was stricter than the comment. If you're ever staring at a guard that seems overly conservative, check whether the comment matches the condition. Sometimes the guard was written for a bug that's since been fixed elsewhere, and nobody went back to relax it.

### Testing

I wrote a standalone test runner (`run_pp_fix_tests.py`) that works without torch/vllm installed — it stubs the heavy intermediate modules and tests the actual edited source. 60 checks covering:

- `compute_extra_count` for 9 TP×PP×ns×MLA combos
- `IPCCacheServerKey.use_mla` round-trip, identity, and msgpack wire backward-compat
- `lookup()` / `free_lookup_locks()` integration (extra_count flows correctly)
- `kv_tp_size` + `is_kv_writer` lock-balance for 12 MLA combos (one writer per (server, stage), acquire = release)
- Source-inspection checks that every adapter sets `use_mla` on keys

There's also a pytest version (`test_pipeline_parallel_fix.py`) for running in a full env.

### What I couldn't do here

I'm on Apple Silicon so I couldn't run an actual CUDA inference test. The lock-balance math is exhaustively verified, but a real multi-GPU PP run would be the final confirmation. If anyone with an H100 or A100 rig wants to try `--pipeline-parallel-size 2 --tensor-parallel-size 4` with Kimi K2 + 2 LMCache servers, that'd be the definitive test.

### Compatibility

- Wire-compatible with old servers/clients (msgspec field-default)
- `model_config.use_mla` is already set correctly for Kimi K2 (`kimi_linear`) and GLM-4.5/4.6 (`glm4_moe_lite`) in vLLM's `is_deepseek_mla()` — I verified this against vLLM main
- DP guard is retained (that's a separate limitation, not touched here)


hamstergamerszhang-ops:
Thanks for the review @gemini-code-assist — good catch on the source-string-inspection test. That was a shortcut I took to avoid pulling in vLLM as a test dependency, but you're right that it violates the repo style guide and would break on any rename.

Fixed in `b6ed1c5`:

1. **Extracted `_validate_multi_server_config`** into a standalone module (`lmcache/integration/vllm/multi_server_config.py`) that has zero vLLM/ZMQ imports. This means it can be unit-tested directly without mocking the connector's heavy dependencies.

2. **Replaced the source-inspection test** with three behavioral tests that call `_validate_multi_server_config` directly:
   - `test_multi_server_dp_still_blocked` — passes a config with `dp_size=2, n_servers=2` and asserts `ValueError` with "data parallelism"
   - `test_multi_server_config_world_size_divisibility` — passes a non-divisible world_size and asserts `AssertionError`
   - `test_multi_server_config_valid` — passes a valid PP+MLA config and asserts no exception

The connector still re-exports `_validate_multi_server_config` for backward compatibility, but the canonical implementation lives in the lightweight module.

Also took the opportunity to add explicit ROCm detection (`is_rocm` property on `CudaDeviceSpec` via `torch.version.hip`). ROCm already works through PyTorch's CUDA compat layer — this just makes the detection explicit for logging and metrics.

ApostaC:
Can you share experiment results, logs, and screenshots to show it is working? Thanks for your contribution!

hamstergamerszhang-ops:
Fair ask. Honest answer: we haven't run this against a real multi-server pipeline-parallel setup yet — reproducing the actual PP+MLA scenario (>1 LMCache server, an MLA model like Kimi K2/GLM-4.6) needs enough GPUs/machines to genuinely exercise the distributed path, and we don't have much hardware to test that at scale right now.

What we have verified: the unit tests added in `b6ed1c5` (`test_multi_server_dp_still_blocked`, `test_multi_server_config_world_size_divisibility`, `test_multi_server_config_valid`) cover the locking/config logic directly, and those pass. But that's not the same as a real end-to-end run under load, and we won't claim it is.

We're working on getting access to a proper multi-server test rig so we can post real logs/screenshots from an actual PP+MLA run rather than asserting it works. Will follow up here once we have that.

PR #3696, opened 2026-06-15
[MP] perf(nixl_store_dynamic): load a request's chunks concurrently

sihara:
The dynamic L2 adapter loaded each chunk of a request **serially**
  (`for key: await dynamic_load_file`), so a single multi-chunk request paid
  `num_chunks × (per-file register + transfer + poll + deregister)` in series —
  the dominant cost of single-request KV loads. Issue the per-chunk reads with
  `asyncio.gather` so the transfers overlap.
  
  ## Measured
  MP, Qwen3-14B-FP8, 256 docs / 1.25 TB, NIXL POSIX (DIO + io_uring), `before → after PR`:
  
  | cons | app-BW (GB/s) | TTFT p50 (ms) | TTFT p99 (ms) |
  |----:|:-------------:|:-------------:|:-------------:|
  | 1 | 4.7 → 20.0 | 1039 → 243 | 1051 → 254 |
  | 2 | 9.1 → 38.8 | 1068 → 247 | 1084 → 295 |
  | 4 | 21.2 → 58.3 | 903 → 312 | 1089 → 554 |
  | 8 | 43.7 → 70.3 | 857 → 566 | 1045 → 754 |
  
  **Static `nixl_store` adapter under identical conditions (reference target):**
  
  | cons | app-BW (GB/s) | TTFT p50 (ms) | TTFT p99 (ms) |
  |----:|:-------------:|:-------------:|:-------------:|
  | 1 | 22.9 | 212 | 224 |
  | 2 | 42.0 | 227 | 276 |
  | 4 | 62.9 | 307 | 463 |
  | 8 | 70.8 | 573 | 824 |
  
  After the fix, dynamic is **on par with static**: cons=8 saturates the storage
  fabric (~70 GB/s) and single-request **TTFT drops ~4×**.
  
  ## Notes
  Each chunk still registers/deregisters its own file. **TODO:** register a
  request's files in a single batched call (like the static adapter); the
  consideration is one large per-request registration list instead of many
  small ones.
  
  Adds a multi-key load test exercising the concurrent path.

PR #4089, opened 2026-07-12
[Misc] NIXL L2 connector, Optimized nixl l2 dynamic connector

ofer:
Changed the dynamic to use path mode and flattened the transfer request so NIXL does the whole transfer in one call instead of one per file and the os.open isn't called anymore.

The NIXL store dynamic connector uses NIXL to transfer KVs to files.  NIXL recently introduced a mode to reduce the need for os level calls for opening and closing files.  The change in addition to batching all reads together increases performance of the dynamic l2 storage path.

These changes yield significant performance gains on the NIXL dynamic storage L2 adapter:

This is the TTFT of varying context lengths using the original nixl store dynamic and the modified nixl store dynamic adapters on Qwen/Qwen3-30B-A3B-Instruct-2507-FP8 on an NVIDIA L40s (chunk size 256)
image

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
@@ -103,6 +103,7 @@ class DynamicNixlStorageAgent:
         self.backend_params = backend_params
         self.l1_align_bytes = l1_memory_desc.align_bytes
         self.file_path = backend_params["file_path"]
+        self.devId = 0
         os.makedirs(self.file_path, exist_ok=True)
         self.use_direct_io = (
             str(backend_params.get("use_direct_io", "false")).lower() == "true"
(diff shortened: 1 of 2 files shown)


ApostaC:
Hi @ofer , thanks for your contribution! 

In terms of logistics, can you fix the DCO issue? Thanks!

ApostaC:
Also, it would be great if you can elaborate how does your changes improve the performance. This will make the PR review much easier.

ziruiliu:
There was a PR #3593 including the improvement that batch operation in one call, it also include enabling OBJ backend in this connector. The path-mode improvement is new.

ziruiliu:
> There was a PR #3593 including the improvement that batch operation in one call, it also include enabling OBJ backend in this connector. The path-mode improvement is new.

Hi @ofer, hope you are not stopped by this #3593. I will refactor this old PR so you can continue on yours. Just be careful about exception I mentioned in last comment

ofer:
> Also, it would be great if you can elaborate how does your changes improve the performance. This will make the PR review much easier.

The idea is that NIXL is designed for batch processing and can split work among threads internally, especially with a larger batch.  The other advantage is that many Pybind11 calls are heavy so that reduction is also an improvement especially around the GIL.

I worked on this PR before PR #3696, after retesting with the newer PR, the batching is no longer showing improvement over the current implementation in my testing, so I put this in draft.  I have other ideas that will likely show improvement.

PR #4101, opened 2026-07-13
[Test] Require an explicit GDS-capable dir for the DMA roundtrip tests

chengy-sysu:

The roundtrip tests in `test_gds_context.py` (`test_gds_write_read_roundtrip`, `test_gds_chunk_larger_than_region_roundtrip`, `test_gds_two_stream_write_read`) place the GDS slab file in pytest's `tmp_path`. `tmp_path` lives under `/tmp`, which is typically on the OS disk rather than a data disk set up for GDS. OS disks commonly use LVM/device-mapper volumes that nvidia-fs cannot register, so on NVIDIA hosts this leads to one of two outcomes, neither desirable:

1. With cuFile compat mode disabled, `cuFileHandleRegister` returns error 5008 and the tests fail even though the code under test is correct.
2. With compat mode enabled (the default), cuFile silently falls back to POSIX IO and the tests pass without ever exercising the DMA path they claim to.

Changes:

1. On NVIDIA, a new `LMCACHE_GDS_TEST_DIR` environment variable must name a directory on a GDS-capable filesystem (e.g. ext4 on a directly attached NVMe device). When unset, the roundtrip tests are skipped with an actionable reason. `tmp_path` is deliberately not used as a fallback because compat mode can silently turn it into POSIX IO and produce false positives.
2. The named directory is probed first: a small preallocated file is opened with `O_DIRECT` and registered with the GDS driver, mirroring how `GDSContext` opens its slab. If registration fails (e.g. the variable points at the OS disk), the tests skip with the underlying error instead of failing.
3. The slab is placed in a private subdirectory created with `tempfile.TemporaryDirectory` and cleaned up after the test.
4. On ROCm, `tmp_path` remains the default when the variable is unset. Per `_gds_available`, library loadability is a sufficient gate for these correctness checks because the hipFile host-bounce fallback still round-trips correctly, so no GDS-capable filesystem is required. ROCm behavior is unchanged.

The seven pure-logic tests (mocked cuFile seam) are unaffected.

**Special notes**:

Verified on an A100 host with a directly attached NVMe (ext4), nvidia-fs loaded, and `allow_compat_mode: false`:

1. `LMCACHE_GDS_TEST_DIR` unset: 7 passed, 3 skipped, with the skip reason explaining the variable and why `tmp_path` is not used.
2. `LMCACHE_GDS_TEST_DIR=/tmp` (deliberately pointing at the OS disk): 3 skipped, with the underlying cuFile error 5008 and the directory in the skip reason.
3. `LMCACHE_GDS_TEST_DIR` pointing at the NVMe mount: 10 passed, all through the real DMA path, temporary slab directories cleaned up.

**If applicable**:

### DIFF
--- a/tests/v1/gpu_connector/test_gds_context.py
+++ b/tests/v1/gpu_connector/test_gds_context.py
@@ -33,6 +33,7 @@ from lmcache.v1.gpu_connector.gds_context import (
     get_gds_context,
     initialize_gds_context,
 )
+from lmcache.v1.memory_management import GDSMemoryObject
 
 
 def _fake_stream(handle: int):
@@ -345,6 +346,7 @@ def test_gds_write_read_roundtrip(gds_slab_dir: Path):
         )
         assert err == L1Error.SUCCESS
         mem_obj = objs[0]
+        assert isinstance(mem_obj, GDSMemoryObject)
 
         buf.fill_(0xAB)
         torch.cuda.synchronize()
@@ -383,6 +385,7 @@ def test_gds_chunk_larger_than_region_roundtrip(gds_slab_dir: Path):
         )
         assert err == L1Error.SUCCESS
         mem_obj = objs[0]
+        assert isinstance(mem_obj, GDSMemoryObject)
 
         # Position-dependent pattern: a mis-offset or swapped segment (e.g. the
         # second segment using the wrong slab offset) would corrupt the bytes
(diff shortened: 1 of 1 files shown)


sammshen (approved):
LGTM

chengy-sysu:
@hickeyma @ApostaC Can you take a look at this PR? Thanks!

ApostaC (approved):

PR #4103, opened 2026-07-13
[Python FS L2 adapter] implement delete and fire L2 listener notifications

aeon-x:
## Summary

The pure-Python `fs` L2 adapter's `delete()` was a no-op stub, and the adapter
fired no `L2AdapterListener` callbacks at all. This PR implements real
deletion and wires up the store / access / delete notifications, so the `fs`
adapter participates in L2 key management end-to-end.

## Changes

- `delete()` — unlinks each key's backing file on the adapter's event loop
  (same bridge pattern as the S3 / HFBucket adapters, 30s result timeout),
  then fires `_notify_keys_deleted` with the stat'd byte sizes. Idempotent:
  missing files are skipped; per-key failures are logged and skipped.
- `_execute_store` now fires `_notify_keys_stored` for newly written keys
  (skipped already-on-disk re-stores are not re-notified, so byte accounting
  is never double-counted).
- `_execute_load` now fires `_notify_keys_accessed` for loaded keys.
- Stale comments updated (`submit_unlock`, `get_usage`).

## Accepted concurrency trade-off

The `fs` adapter tracks no per-key in-flight locks (unlike S3's
`_locked_keys`), so `delete` is best-effort under concurrency:

- A delete racing an in-flight **load** turns that load into a miss — the
  load path already treats `FileNotFoundError` as not-found, so this degrades
  gracefully (covered by a test).
- A delete racing a **store** of the same key can be undone by the store's
  tmp-file rename landing afterwards; the key simply becomes resident again
  and a later eviction cycle re-picks it.

Per-key lock tracking (S3-style skip of in-flight keys) is a possible
follow-up if these races matter for a workload; they are benign for the
eviction use case.

## Tests

New `tests/v1/distributed/test_fs_l2_adapter_delete.py` (7 tests, public
interface only): delete removes keys + notifies + byte accounting shrinks by
the stat'd size (verified via the lookup API); missing-key no-op;
double-delete idempotency; empty key list; load-after-delete degrades to a
miss; re-store of an existing key does not re-notify; load fires accessed.


maobaolong (approved):
@aeon-x This LGTM. Thanks for your contribution!

The missing delete content for L2 fs adapter has been filled in.

deng451e (approved):
LGTM

PR #4110, opened 2026-07-14
[Docs] share_kv_cache quickstart: use a >=1-chunk prompt in centralized example

### LINKED ISSUE #4109: [Docs] share_kv_cache centralized quickstart never produces a cache hit (sub-chunk prompt)
**Label**

`bug`, `documentation` (relates to [LMCache Onboarding #3372](https://github.com/LMCache/LMCache/issues/3372))

**Describe the bug**

In the [Share KV cache across multiple LLMs](https://docs.lmcache.ai/getting_started/quickstart/share_kv_cache.html) quickstart, the **Centralized KV cache sharing** example uses a very short prompt:

```
"Explain the significance of KV cache in language models."
```

That prompt is ~12 tokens, far below the configured `chunk_size: 256`. Because `save_unfull_chunk` defaults to `False`, a prompt smaller than one full chunk is never stored. As a result, following the steps verbatim produces **no cross-instance cache hit** — the second engine logs `LMCache hit tokens: 0`, so the example does not demonstrate the KV reuse it describes.

The P2P example on the same page already avoids this by repeating the sentence (`printf '...%.0s' {1..100}`) to exceed one chunk, so the two examples are inconsistent.

**To Reproduce**

1. Follow the *Centralized KV cache sharing* section verbatim (config with `chunk_size: 256`, two `vllm serve` engines on ports 8000/8001).
2. Send the short-prompt `curl` request to the engine on port 8000.
3. Send the same request to the engine on port 8001.
4. Observe the port-8001 engine logs.

**Expected behavior**

The second engine retrieves the KV cache from the first and logs a non-zero hit, e.g. `LMCache hit tokens: <N>`, demonstrating cross-instance reuse.

**Actual behavior**

The second engine logs `LMCache hit tokens: 0` — nothing was stored on the first request because the prompt is shorter than `chunk_size` and `save_unfull_chunk` is `False` by default.

**Additional context**

Two possible fixes:
1. Make the centralized example's prompt exceed one full chunk (mirror the P2P example), and add a note explaining the sub-chunk-prompt limitation. *(preferred — consistent with the rest of the page)*
2. Set `save_unfull_chunk: true` in the example config.

I have a docs-only fix implementing option 1 ready to submit as a PR.

`docs/source/getting_started/quickstart/share_kv_cache.rst`

SeongjaeP:

Fixes #4109.
Refs #3372.

The "Centralized KV cache sharing" quickstart example used a 12-token prompt
(`"Explain the significance of KV cache in language models."`), which is far
below the configured `chunk_size: 256`. Since `save_unfull_chunk` defaults to
`False`, that sub-chunk prompt is never stored. Following the guide verbatim,
the second engine therefore logs:

```
LMCache hit tokens: 0
```

so the example never actually demonstrates the cross-instance KV reuse it
describes.

This PR makes the two `curl` requests repeat the sentence to exceed one full
chunk (using `printf '...%.0s' {1..100}`), mirroring the existing P2P example
on the same page which already does this. It also adds a `.. note::`
explaining that prompts shorter than `chunk_size` are not stored unless
`save_unfull_chunk: true` is set, so a short prompt can never produce a
cross-instance hit.

**If applicable**:


YeonwooSung:
Thanks for filing #4109 and opening this — the diagnosis is correct.

I opened https://github.com/LMCache/LMCache/pull/4792 with the same centralized-prompt change plus the matching example README and an MP-mode sharing guide (the in-process page is deprecated). If maintainers prefer to merge this PR first, I can rebase #4792 and drop the overlapping hunk.

PR #4116, opened 2026-07-15
feat(vllm): add enqueue-time prefetch to LMCacheMPConnector

chfeng-cs:
This PR ports enqueue-time KV prefetch from [vllm-project/vllm#42321](https://github.com/vllm-project/vllm/pull/42321) into LMCache's `LMCacheMPConnector`.

Previously, LMCache lookup started when a request was selected by the scheduler. For disk-backed KV cache under load, this put L2-to-CPU staging on the TTFT critical path. This change starts lookup when the request enters the waiting queue, allowing storage I/O to overlap with queue wait.

```text
Before: enqueue → queue wait → schedule → LMCache lookup → prefill
After:  enqueue → LMCache lookup ─┐
                → queue wait ─────┴→ schedule → prefill
```

KV cache loading path before and after enqueue-time prefetch

## Implementation

- Add the opt-in `lmcache.mp.eager_prefetch` configuration, disabled by default.
- Implement `LMCacheMPConnector.on_new_request()` to submit lookup at enqueue time.
- Reuse the idempotent `maybe_submit_lookup_request()`, keeping the scheduling-time call as a fallback.
- Skip resumable requests because their token IDs may be incomplete at enqueue time.
- Document the new connector option.

```json
{
  "lmcache.mp.eager_prefetch": true
}
```

## Flow boundaries

- **Changed:** Lookup submission moves from scheduling time to enqueue time, so LMCache's L2-to-CPU staging can overlap with scheduler queue wait.
- **Unchanged:** `get_num_new_matched_tokens()` still checks the asynchronous result and drives GPU block allocation.
- **Unchanged:** `start_load_kv()` still submits `RETRIEVE`; LMCache performs H2D into vLLM GPU KV blocks through CUDA IPC.
- **Unchanged:** Residual prefill and decode follow the existing path.

## Upstream evaluation

The upstream experiment used Llama 3.1 8B on an NVIDIA L20 with an 8,767-token disk-backed KV context. Five requests were submitted at 2.4-second intervals, and results were averaged over three runs.

- **No backlog:** Approximately 304–306 ms TTFT for both variants, with no measurable eager-prefetch overhead.
- **Under backlog:** Eager prefetch kept TTFT nearly flat at approximately 306 ms by starting disk I/O during queue wait.
- **Overall:** Request 5 completed about 470 ms earlier, reducing TTFT by up to 28%.

Timing instrumentation:

- https://github.com/chfeng-cs/vllm/tree/lmcache-eager-prefetch-bench
- https://github.com/chfeng-cs/LMCache/tree/eager-prefetch-timing

These results come from the upstream vLLM PR; this PR ports the mechanism into the LMCache-owned integration.

- The feature is opt-in and disabled by default.
- The SVG is a conceptual timeline and does not represent actual operation durations.

**If applicable**:


maobaolong:
@chfeng-cs Thanks for the contribution! I think this is a great idea.

Prefetching kvcache as soon as a request enters the VLLM Scheduler queue is a good idea. This effectively shifts the time spent moving data from L2 to L1 into the queuing period as much as possible, which is a big help.

Overall, the logic looks good. One small thing: we should probably fix that DCO issue first.

maobaolong:
@chunxiaozheng How about your idea for this PR?

maobaolong:
@sammshen Would you like to take a look at this interesting PR?

chfeng-cs:
I've fixed the DCO issue and updated the PR. Thanks for the review and the kind words!

ApostaC (approved):
LGTM!

ApostaC:
@chfeng-cs It would be very helpful if you could share a few performance benchmark numbers or benchmark cases showing the benefit of this eager prefetch. 

We can make a blog post about this if you are interested. Please feel free to reach out to us offline.

KuntaiDu (approved):
LGTM!

PR #4118, opened 2026-07-15
turboquant: fp8 passthrough for vLLM kv-cache-dtype fp8

### LINKED ISSUE #4064: TurboQuant serde fails on vLLM fp8 KV cache — CompilationError: cannot cast uint8 to fp8e4nv, 0 bytes offloaded to L2
With the TurboQuant serde on an L2 adapter and vLLM serving with --kv-cache-dtype fp8, every serialize task throws a Triton CompilationError and no KV is written to L2 — the SSD copy is silently empty while store requests report "completed." Serving with bf16/auto KV works, so the failure is specific to the fp8 (uint8-backed) KV path.

Environment

LMCache v0.5.1 (TurboQuant serde, MP server + fs L2 adapter)
vLLM 0.22.0 (V1), LMCacheMPConnector, kv_role=kv_both
Model: Qwen3-Coder-30B-A3B-Instruct-FP8 · GPU: RTX PRO 6000 (sm_120), CUDA 13.0
--kv-cache-dtype fp8; preset turboquant_k8v4 (also repros with 4bit_nc, 3bit_nc)
Config — L2 adapter: {"type":"fs","base_path":"…","serde":{"type":"turboquant","preset":"turboquant_k8v4","block_size":16,"max_workers":8}}

Steps: start MP server w/ turboquant L2 adapter → serve vLLM --kv-cache-dtype fp8 via LMCacheMPConnector → drive traffic past L1 → du -sb <L2 path> stays 0.

Expected: KV quantized and written to L2; restores dequantize back.
Actual: 0 bytes on L2; all serialize/store tasks fail; l2_store_completed_requests_total increments but …_objects_chunks_total/on-disk bytes = 0 (failure swallowed).

Log

LMCache ERROR: Serde task 0 (SERIALIZE) failed        (async_processor.py:196)
  File ".../serde/turboquant/turboquant.py", line 641, in serialize
    triton_turboquant_store(key, value, ...)
  File ".../serde/turboquant/store_kernel.py", line 387, in triton_turboquant_store
    _tq_fused_store_fp8[grid](k_flat, v_flat, ...)
triton.compiler.errors.CompilationError: at 45:56:
AssertionError: cannot cast uint8[constexpr[128]] to <['128'], fp8e4nv>
LMCache WARNING: Store task 0 to adapter 0 failed for keys: [...]

lcheng321:
# Problem

vLLM `--kv-cache-dtype fp8` produces KV tensors as fp8 bit patterns packed in uint8 containers. The store kernel calls `.to(tl.float8e4nv)` which performs numeric conversion instead of bitcast, causing a CompilationError. The load kernel always dequantizes to float16 with no passthrough path, making round‑trip impossible.

Closes #4064

# Fix

- store_kernel.py: Added `KEY_ALREADY_FP8` constexpr. Uses `bitcast=True` instead of numeric `.to()`. Added `tl.static_assert` to reject e4b15 (vLLM only emits e4m3/e4nv).

- decode_kernel.py: Added `KEY_PASSTHROUGH_FP8` constexpr. Skips dequantization and writes raw uint8 bytes directly.

- turboquant.py: Auto‑detects dtype on both paths (`src.dtype == uint8` in serialize, `dst.dtype == uint8` in deserialize). Allocates `k_out` and `v_out` separately so keys stay uint8 while values remain float16. No public API change.

# Evidence

## Before (upstream, no fix):
triton.compiler.errors.CompilationError: at 45:56:
AssertionError: cannot cast uint8[constexpr[128]] to <['128'], fp8e4nv>

## After (this PR), full round‑trip verification:
serialize() wrote 200704 bytes (expected 200704)
deserialize() completed
key bytes match: True (mismatches: 0 / 131072)
PASS: all key fp8 bytes match -- round‑trip verified

## Full test suite (rerun after final changes):
22 passed

### DIFF
--- a/lmcache/v1/distributed/serde/turboquant/decode_kernel.py
+++ b/lmcache/v1/distributed/serde/turboquant/decode_kernel.py
@@ -62,6 +62,7 @@ def _tq_full_dequant_kv(
     BLOCK_D: tl.constexpr,
     NORM_CORRECTION: tl.constexpr = 0,
     FP8_E4B15: tl.constexpr = 0,  # 1 = use e4b15 (Ampere/Ada), 0 = e4nv (Hopper+)
+    KEY_PASSTHROUGH_FP8: tl.constexpr = 0,  # 1 = K_out_ptr is uint8; store raw fp8 bits
 ):
     """Full dequant: reconstruct K (MSE centroids * norm or FP8) and V to fp16."""
     pos = tl.program_id(0)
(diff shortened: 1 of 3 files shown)


lcheng321:
### before

LMCache4064before

### after

after

lcheng321:
@ApostaC @chunxiaozheng @maobaolong Pushed a fix for the Ampere bitcast issue Gemini flagged. Would appreciate a review when you have a moment.

PR #4121, opened 2026-07-16
[Refactor] Group prefetch params into PrefetchRequestSpec

aeon-x:
## Summary

Groups the prefetch entry-point parameters into a single frozen dataclass,
`PrefetchRequestSpec`, in `distributed/api.py` (beside `PrefetchHandle`). It
bundles the six L2-fetch inputs: `keys`, `layout_desc`, `extra_count`,
`policy`, `attn_desc`, `mode`.

- **`PrefetchController.submit_prefetch_request(spec)`** — replaces the six
  flat params and removes the fragile positional submission-queue tuple; the
  field list + defaults now live in exactly one place.
- **`StorageManager.submit_prefetch_task(spec, external_request_id="",
  skip_l2=False)`** — takes the shared spec plus the two StorageManager-layer
  orchestration knobs as explicit args. These are deliberately kept **out** of
  the spec: the controller also consumes `PrefetchRequestSpec`, so folding
  `skip_l2`/`external_request_id` in would give it settable-but-ignored fields.
- Updates the six MP-module callers (`warm_prefetch`, `p2p_controller`,
  `blend`, `blend_v3`, `lookup`) and all affected tests.

No behavior change.

## Testing

- CI tests


maobaolong (approved):
LGTM — The dataclass is a nice cleanup, it's much cleaner now.

maobaolong:
@chunxiaozheng Would you like to take a double check?

sammshen (approved):

PR #4056, opened 2026-07-08
style(fs_connector): convert f-string log calls to %-format (G004)

orangeCatDeveloper:
## What
Converts the `logger.*(f"...")` calls in `lmcache/v1/storage_backend/connector/fs_connector.py` to lazy `%`-formatting, per Python's logging guidance and ruff rule **G004**.

This continues the logging-format cleanup campaign (ref #3372) — same treatment already applied to `token_database.py` (#3982) and `s3_connector.py` (#3977).

## Why
`%`-style logging defers string interpolation until the record is actually emitted, so filtered-out log calls cost nothing.

## Testing
- `ruff check --select G004` is clean on the file.
- `ruff check` and `ruff format --check` pass.

PR #3982, opened 2026-07-02
style(token_database): convert f-string log calls to %-format (G004)

Anai-Guo:
## What

Converts the three `logger.*(f"...")` calls in `lmcache/v1/token_database.py` to lazy `%`-formatting, per Python's logging guidance and ruff rule **G004**.

This continues the logging-format cleanup campaign (ref #3372) — same treatment already applied to `s3_connector.py` (#3977, merged) and `local_disk_backend.py` (#3973).

## Why

`%`-style logging defers string interpolation until the record is actually emitted, so `debug`/`info` calls that are filtered out at the current level do no formatting work. It also silences ruff G004 for this file.

## Changes

- `logger.info(f"Initialized NONE_HASH={NONE_HASH} ...")` → `%s` with `NONE_HASH` arg
- `logger.info(f"Loaded '{func_name}' from {module_path} ...")` → `%s`/`%s` args
- `logger.warning(f"Could not load '{hash_algorithm}' ...")` → `%s` arg (multi-line string preserved)

No behavior change — message text and arguments are identical.

## Verification

```
ruff check --select G004 lmcache/v1/token_database.py   # All checks passed!
ruff format --check lmcache/v1/token_database.py        # already formatted
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4124, opened 2026-07-16
style(s3_connector): convert f-string log calls to %-format (G004)

Smallfu666:
## What
Convert the 11 f-string logging calls in
`lmcache/v1/storage_backend/connector/s3_connector.py` to lazy %-format
so arguments are interpolated only when the log record is emitted (G004).

## Why
Part of the ongoing G004 cleanup campaign (cf. #3982 token_database,
#4056 fs_connector). No behavior change.

## How verified
`ruff check <file> --select G004` clean; `pre-commit run --files <file>` passes.

### DIFF
--- a/lmcache/v1/storage_backend/connector/s3_connector.py
+++ b/lmcache/v1/storage_backend/connector/s3_connector.py
@@ -218,8 +218,8 @@ class S3Connector(RemoteConnector):
         if got["err"] or got["status"] != 200:
             if got["status"] != 404:  # Don't warn for 404, it's expected
                 logger.warning(
-                    "Encountering error in S3 HEAD request "
-                    f"with error code: {got['status']}"
+                    "Encountering error in S3 HEAD request with error code: %s",
+                    got["status"],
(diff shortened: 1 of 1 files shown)


maobaolong (approved):
@Smallfu666 Thank you for your contribution! This PR LGTM!

Just a heads-up: starting in Q3 or Q4, we are going to focus on LMCache MultiProcess mode. 

Would you like to try the `lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py`?

Smallfu666:
Hi @maobaolong, thanks for the review! I'd be happy to look at the S3L2Adapter. I noticed #3831 is still open (the GET callback can write past MemoryObj bounds on oversized responses) — would fixing that plus test coverage for the failure paths be a good place to start, or is the MultiProcess push pointing more toward integration work (real S3/MinIO, concurrent stress)? Happy to go either way, just want to align on priority first.

sammshen (approved):

PR #4128, opened 2026-07-16
[Fix] Add kernels for vLLM-packed KV cache format

### LINKED ISSUE #4111: MP mode: RuntimeError: Unsupported EngineKVFormat: 10 (NL_X_NB_NH_BS_TWO_HS) — no native kernel + no per-format fallback
**Label:** bug

**Describe the bug**

In MP mode, storing KV cache fails with:

```
LMCache ERROR: Cannot store keys due to exception (lmcache_driven_transfer.py:1078)
  File ".../lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py", line 1066, in store
    transfer_kv_per_object_group(...)
  File ".../lmcache_driven_transfer.py", line 449, in transfer_kv_per_object_group
    _run_object_group_transfer_plan(...)
  File ".../lmcache_driven_transfer.py", line 402, in _run_object_group_transfer_plan
    lmc_ops.execute_object_group_transfer(...)
RuntimeError: Unsupported EngineKVFormat: 10
```

`EngineKVFormat` value 10 is `NL_X_NB_NH_BS_TWO_HS` — vLLM's non-MLA blocks-first
layout with K/V fused into the trailing dim (per-layer tensor
`[NB, NH, BS, 2, HS]`, registered raw as 4-D `[NB, NH, BS, 2*HS]`).

There are **two underlying defects**:

1. **Missing native kernel for format 10.** The format is fully wired for the
   Python side — enum (`csrc/engine_kv_format.h`, `python_ops_fallback.py`),
   pybind registration (`csrc/pybind.cpp`), the vLLM detector that emits it
   (`lmcache/v1/gpu_connector/kv_format/detectors/vllm.py:44`), the layout spec
   (`specs/nl_x_nb_nh_bs_two_hs.py`), and a working generic Python-fallback
   transfer. But it is **absent from both native CUDA transfer kernels**:
   - `csrc/mp_mem_kernels.cu` (MP object-group path) — its `DISPATCH_FORMAT`
     switch enumerates formats 0–9; format 10 falls through to
     `default: TORCH_CHECK(false, "Unsupported EngineKVFormat: ", ...)`
     (line ~294).
   - `csrc/mem_kernels.cu` (single-layer path) — likewise not handled
     (its switch stops earlier; same `default` throw).

   So the detector can classify a KV cache into a format that no native kernel
   implements.

2. **No per-format fallback — the native/Python decision is a single global
   boolean.** `lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py`
   computes `_HAS_NATIVE_OBJECT_GROUP_TRANSFER` once at import (line 57) as
   "does a native `execute_object_group_transfer` symbol exist?" — **not**
   "does the native impl support *this* format?". On a CUDA host the native
   symbol shadows the Python fallback for all formats, and the store path
   (line 446) commits to the native call with no `try/except` and no
   per-format capability check. So a format the native kernel can't handle
   hard-crashes instead of using the Python-fallback implementation that
   already exists and supports it.

**To Reproduce**

1. Run LMCache in MP mode with the native (CUDA) ops backend, on latest dev 
2. Serve a vLLM model (eg. GPT OSS 20B) on latest main 
3. Issue a request so LMCache stores KV.
4. The store fails with `RuntimeError: Unsupported EngineKVFormat: 10`.

**Expected behavior**

Storing KV in format 10 should succeed. Either:
- the native MP object-group kernel (`mp_mem_kernels.cu`) implements format 10
  (`[NB, NH, BS, 2, HS]` — note the K/V axis is second-to-last / interleaved,
  which no existing native case covers, so this needs new index math), and/or
- LMCache degrades gracefully per-format: when the native kernel does not
  support the detected format, fall back to the Python-fallback
  `execute_object_group_transfer` (which already handles format 10) instead of
  crashing. A native "supported formats" capability query checked alongside
  the existing `GDSMemoryObject` condition at
  `lmcache_driven_transfer.py:446` would do this cleanly.

**Additional context**

- VLLM introduced the change recently in this PR (merged to main on 7/11/2026): https://github.com/vllm-project/vllm/pull/44455 
It is part of the Standardize KV-cache Layouts RFC: https://github.com/vllm-project/vllm/issues/42082 
- Format 10 (`NL_X_NB_NH_BS_TWO_HS`) and the native object-group transfer path
  are both very recent (`#3907` 2026-06-29, `#3908` 2026-07-02), which is
  likely why this hasn't surfaced widely yet — it needs the narrow
  intersection of {recent build} × {blocks-first fused attention backend} ×
  {MP + native CUDA path}.
- Workaround (no rebuild): configure vLLM to use a standard flash-attn /
  flashinfer attention backend, which emits a layout in the supported 0–9 set
  (e.g. format 1/2/6/7) that the native MP kernel handles.
- Of the two fixes, the per-format fallback (defect #2) is the smaller, safer,
  CPU-testable change and restores correctness immediately (slower path); the
  native kernel (defect #1) is the performant long-term fix but requires a GPU
  build and careful index math for the interleaved K/V axis.

ApostaC:

Fixes #4111 

Added the following:
- Kernel support for existing format `NL_X_NB_NH_BS_TWO_HS`
- New KV format `NL_X_NB_BS_NH_TWO_HS` (NHD) with kernel support
- Fix for CPU-side fallback ops

One extra note: the kernel treats the format as normal HND/NHD but with kv_size = 1 and head_size = 2 * "real head size"

image

**If applicable**:


lmcache/python_ops_fallback.py
```diff
@@ -299,6 +299,12 @@ class EngineKVFormat(IntEnum):
     # second-to-last, recovered by splitting the fused [..., 2 * head_size].
     NL_X_NB_NH_BS_TWO_HS = 10
 
+    # used by: vLLM non-MLA blocks-first attention (NHD layout) with K/V fused
+    # into the trailing dim. Per-layer physical shape
+    # [num_blocks, block_size, num_heads, 2, head_size] -- like
+    # NL_X_NB_NH_BS_TWO_HS but tokens before heads.
+    NL_X_NB_BS_NH_TWO_HS = 11
```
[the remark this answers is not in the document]
ApostaC: added

lmcache/v1/multiprocess/transfer_context/base.py
```diff
@@ -357,7 +375,9 @@ def gather_paged_kv_to_cpu(
 
     block_size = get_block_size(normalized, engine_kv_format)
     num_layers = get_num_layers(normalized, engine_kv_format)
-    hidden_dim_size = get_hidden_dim_size(normalized, engine_kv_format)
+    token_hidden = get_num_heads(normalized, engine_kv_format) * get_head_size(
```
sammshen: what is this rename???

lmcache/v1/gpu_connector/utils.py
```diff
@@ -324,6 +324,13 @@ def get_page_buffer_size(
     return get_spec(kv_caches, engine_kv_format).page_buffer_size()
 
 
+def get_kv_size(
```
sammshen, line 327: like this new helper

lmcache/v1/multiprocess/transfer_context/base.py
ApostaC: good question to ask my claude 😠

ApostaC: Oh it's not "hidden dim" any more, since we now supports "packed KV" in the hidden dim

sammshen: nit: put a TODO for renaming to content size

sammshen (approved):
LGTM!

royyhuang (approved):

PR #4129, opened 2026-07-17
xpu: pin LMCache host buffer via sycl::malloc_host

libinta:
The local_cpu backend host buffer was never pinned on XPU: the pinned-alloc path (lmc_ops.alloc_pinned_ptr) is CUDA-only (cudaHostAlloc) and is not compiled into xpu_ops, so allocation silently fell back to pageable host memory -> D2H KV store ran at ~1.5 GB/s.

Add a SYCL analog alloc_pinned_ptr/free_pinned_ptr (sycl::malloc_host on the current XPU device context) bound under the same names as the CUDA ops, so lmcache._get_backend() overrides them by name on XPU with no device-specific branching in _allocate_cpu_memory/_free_cpu_memory. Restores true USM-pinned host memory -> D2H store ~10-30 GB/s (~20x).

sycl::malloc_host is capped at the device max_mem_alloc_size (~30 GiB on B70 = 95% of one card's VRAM), so a large max_local_cpu_size can overflow it. Add a device-agnostic pageable fallback (with a warning) so the engine degrades gracefully instead of crashing; torch pin_memory=True is avoided because torch XPU pinned buffers segfault on access above ~16 GiB. shm / NUMA / hugepage allocations are excluded from the fallback since a plain pageable buffer cannot preserve their semantics.

Split out of #4022 (pin-memory allocation only); the SGLang connector logic changes in that PR are being tracked separately.

**If applicable**:


maobaolong:
@libinta This operator was implemented for the lmcache in-process mode, right? Is there a corresponding operator for the LMCache multi-process mode?

lmcache/v1/memory_management.py
```diff
@@ -588,7 +594,21 @@ def _allocate_cpu_memory(
                     "Failed to allocate huge pages. "
                     "Please grow the 2 MiB hugepage pool."
                 )
-        raise
+            raise
+        if shm_name or numa_mapping is not None or use_hugepages:
```
DongDongJu: Can we avoid the fallback in here? its better to raise the error that unsupported pinned size.

csrc/sycl/mem_kernels_sycl.cpp
```diff
@@ -1021,3 +1021,21 @@ void lmcache_memcpy_async(uintptr_t dest, uintptr_t src, size_t nbytes,
...
+// ---------------------------------------------------------------------------
+uintptr_t alloc_pinned_ptr(size_t size, unsigned int flags) {
+  void* ptr = sycl::malloc_host(size, c10::xpu::get_device_context());
+  TORCH_CHECK(ptr != nullptr, "sycl::malloc_host failed for ", size,
+              " bytes (XPU pinned host)");
+  return reinterpret_cast<uintptr_t>(ptr);
+}
+
+void free_pinned_ptr(uintptr_t ptr) {
+  sycl::free(reinterpret_cast<void*>(ptr), c10::xpu::get_device_context());
```
DongDongJu, line 1051: Is there any race condition problem between this alloc and free func when we have multiple devices?

DongDongJu (changes requested):
Hello @libinta,
Thanks for the contributions.
I left few opinions
Especially, I don't know why we need a fallback in here.

lmcache/v1/memory_management.py
```diff
@@ -588,7 +594,21 @@ def _allocate_cpu_memory(
...
+            raise
+        logger.warning(
+            "Pinned host allocation of %d bytes via the native allocator failed "
+            "(%s); falling back to pageable host memory.",
+            size,
+            e,
+        )
+        buffer = torch.empty(size, dtype=torch.uint8, pin_memory=False)
+        buffer.fill_(0)
+        _FALLBACK_HOST_BUFFERS[buffer.data_ptr()] = buffer
```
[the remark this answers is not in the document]
libinta: done

lmcache/v1/memory_management.py
```diff
@@ -607,6 +627,11 @@ def _free_cpu_memory(
     if torch_dev.is_available():
         torch_dev.synchronize()
 
+    # Pageable fallback buffers are torch-owned (freed on GC); just drop our
+    # reference instead of routing through the native free path.
+    if _FALLBACK_HOST_BUFFERS.pop(buffer.data_ptr(), None) is not None:
+        return
```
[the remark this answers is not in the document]
libinta: done

csrc/sycl/mem_kernels_sycl.cpp
libinta: fixed to handle race condition

libinta:
> @libinta This operator was implemented for the lmcache in-process mode, right? Is there a corresponding operator for the LMCache multi-process mode?

yes, it's for in process. there is different set of apis for mp, we haven't implemented that yet as our test case is in process. MP can be next.

hlin99:
hi @libinta thanks for splitting the PR, but I would expect this PR only contains sycl kernel. the logic change in memory management should be in this other PR for broader review.

libinta:
> hi @libinta thanks for splitting the PR, but I would expect this PR only contains sycl kernel. the logic change in memory management should be in this other PR for broader review.

Thanks. Done

DongDongJu (approved):
LGTM

hlin99:
hi @libinta please fix DCO issue, and there's a build error in buildkite/unit-tests-xpu CI caused by this PR. please fix. thx

hlin99 (approved):
LGTM

PR #4135, opened 2026-07-17
fix(vllm/mp): realign the 0201 MP connector with the current adapter API (#4134)

### LINKED ISSUE #4134: [Bug] MP connector variants (_0201/_0180) drifted from vllm_multi_process_adapter — MP mode cannot start on vLLM 0.19/0.20
## Summary

The versioned vLLM MP connectors (`lmcache/integration/vllm/lmcache_mp_connector_0201.py`, `_0180.py`) are out of sync with the current `vllm_multi_process_adapter.py`, so the MP path cannot start on vLLM 0.19.x/0.20.x. Three drifts:

1. `create_scheduler_adapter` passes `server_url=server_url` (str), but `LMCacheMPSchedulerAdapter.__init__` now takes `server_urls: list[str]`. The vLLM-bundled copy of the connector passes a plain string positionally, which then gets iterated character-wise: `zmq.error.ZMQError: Invalid argument (addr='t')`.
2. Both `create_scheduler_adapter` and `create_worker_adapter` construct `ParallelStrategy` with the old 7-argument layout; the dataclass now has 6 fields (`kv_world_size` / `kv_worker_id` became derived properties): `TypeError: ParallelStrategy.__init__() takes 7 positional arguments but 8 were given`.
3. `wait_for_save` in `_0201` reads `self.worker_adapter.use_mla` / `.is_first_rank_of_pp_group`, which no longer exist on `LMCacheMPWorkerAdapter` (they moved into `parallel_strategy`): `AttributeError: 'LMCacheMPWorkerAdapter' object has no attribute 'use_mla'`.

## Environment

- LMCache main @ `668a1fd` (source build), vLLM 0.19.1 (`vllm/vllm-openai:latest-cu130-ubuntu2404`)
- `lmcache server --l1-size-gb N --eviction-policy LRU` + vLLM `--kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both"}'`

## Extra wrinkle (vLLM-side, FYI)

vLLM's `KVConnectorFactory` resolves registered names **before** consulting `kv_connector_module_path`, so a user cannot point the registered name `LMCacheMPConnector` at the (fixed) lmcache-shipped module — the vendored copy inside vLLM always wins. We worked around it by exporting an alias class name (`LMCacheMPConnectorV2 = LMCacheMPConnector` at the end of `_0201`) and passing that as `kv_connector` with `kv_connector_module_path`. Might be worth documenting, or upstreaming a rename so the lmcache-side connector is reachable.

With fixes 1–3 (plus the alias trick) the MP connector starts and serves fine on vLLM 0.19.1. Happy to PR the three one-liners.

Anai-Guo:
## Summary

Fixes #4134 (partially — see "Scope" below).

`lmcache/integration/vllm/lmcache_mp_connector_0201.py` drifted from the current `vllm_multi_process_adapter.py`, so the MP path cannot start on vLLM 0.20.1. Three drifts, all confirmed against `dev`:

1. **`server_url=` -> `server_urls=[...]`.** `create_scheduler_adapter` passes `server_url=server_url`, but `LMCacheMPSchedulerAdapter.__init__` now takes `server_urls: list[str]` (it asserts `len(server_urls) >= 1` and fans out a `MessageQueueClient` per URL).

2. **`ParallelStrategy` arity.** Both factories construct it with 7 positional args, but the dataclass now has 6 fields — `kv_world_size` / `kv_worker_id` became derived properties (`TypeError: __init__() takes 7 positional arguments but 8 were given`).

   Since the MLA adjustment now lives inside those properties (`kv_world_size` divides by `tp_size` when `use_mla`), feeding them the *pre-adjusted* values from `extract_world_size_and_kv_rank` would adjust twice. That helper is therefore superseded and removed, and the raw vLLM geometry is passed straight through — mirroring `build_parallel_strategy_from_vllm_config` in the unversioned `lmcache_mp_connector.py`. `n_servers=1` keeps the single-server behaviour this connector already had.

3. **`wait_for_save` attributes.** It reads `self.worker_adapter.use_mla` and `.is_first_rank_of_pp_group`; neither exists on `LMCacheMPWorkerAdapter` any more (`is_first_rank_of_pp_group` now appears nowhere else in the repo). Both moved onto `parallel_strategy`, surfaced via the `is_kv_writer` property — which encodes exactly the old condition: `not is_kv_writer` ⟺ `use_mla and worker_id % (tp_size // n_servers) != 0`.

## Scope

The issue reports `_0180` as affected too, but it isn't: it calls the adapters through the **legacy positional signature** (`server_url, ctx, model, kv_world_size, kv_worker_id, block_size`), which `_normalize_adapter_init_args` still explicitly supports, and it has no `use_mla` guard in `wait_for_save`. Left untouched.

The issue's vLLM-side `KVConnectorFactory` note (registered names resolving before `kv_connector_module_path`) is upstream in vLLM and out of scope here.

## Testing

No test covers the versioned connectors (importing them requires a matching vLLM), so this is verified by inspection against the adapter signatures on `dev`: `ParallelStrategy` fields, `LMCacheMPSchedulerAdapter.__init__`, and the `is_kv_writer` property. `ruff check` / `ruff format` are clean on the changed lines (the one pre-existing `torch_device_type` unused-import warning is untouched).

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector_0201.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0201.py
@@ -91,6 +91,14 @@ def build_parallel_strategy(vllm_config: VllmConfig) -> ParallelStrategy:
     ``ParallelStrategy`` derives ``kv_world_size`` / ``kv_worker_id`` from the
     raw vLLM geometry (including the MLA adjustment), so the world size and
     rank are passed through unmodified.
+
+    Args:
+        vllm_config: The vLLM configuration to read the parallel geometry and
+            MLA setting from.
+
+    Returns:
+        A ``ParallelStrategy`` describing the KV parallel layout for this
+        worker, with ``n_servers`` fixed at 1.
     """
     pc = vllm_config.parallel_config
     return ParallelStrategy(
--- a/lmcache/integration/vllm/lmcache_mp_connector_0201.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0201.py
@@ -84,28 +84,23 @@ def reformat_block_ids(block_ids: tuple[list[int], ...] | None) -> list[int]:
     return block_ids[0]
 
 
-def extract_world_size_and_kv_rank(
-    world_size: int,
-    rank: int,
-    vllm_config: VllmConfig,
-) -> tuple[int, int]:
+def build_parallel_strategy(vllm_config: VllmConfig) -> ParallelStrategy:
     """
-    Convert the rank for the MLA.
+    Build the KV parallel geometry from a vLLM config.
+
+    ``ParallelStrategy`` derives ``kv_world_size`` / ``kv_worker_id`` from the
+    raw vLLM geometry (including the MLA adjustment), so the world size and
+    rank are passed through unmodified.
     """
-    use_mla = mla_enabled(vllm_config.model_config)
-    if not use_mla:
-        return world_size, rank
-    else:
-        # Tensor parallel does not change the KV caches for MLA models.
-        # So we need to "exclude" the effect of TP on rank and world size
-        tp_size = vllm_config.parallel_config.tensor_parallel_size
-        # vLLM constructs TP groups first, and then construct other
-        # parallel groups on top of TP groups.
-        # for example, TP=4, PP=2,
-        # PP group: [0, 1, 2, 3], [4, 5, 6, 7]
-        # TP group: [0, 4], [1, 5], [2, 6], [3, 7]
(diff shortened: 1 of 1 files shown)


sammshen (approved):
LGTM

deng451e (approved):
LGTM

Anai-Guo:
Friendly ping @sammshen @deng451e — this has been approved since 07-17. Is there anything else needed before this can be merged? 🙏

PR #4136, opened 2026-07-17
[Bugfix] Handle request abort during MP retrieve

RedHeartSecretMan:
Fixes the LMCacheMP side of vllm-project/vllm#43226.

## Problem

When a request is aborted while an asynchronous retrieve is active, the worker adapter can report a nonexistent store completion for that request. The receive and abort notifications may also cross scheduler steps: a worker can remove its completed retrieve future before the abort-related `finished_req_ids` arrives.

In either case, vLLM can receive both `finished_recving` and `finished_sending` for the same aborted request. Because the scheduler processes receive completions first, it frees the request and then fails the ownership assertion while processing the send completion.

## Root cause

- The scheduler moved a submitted retrieve directly to `READY`, so later scheduling in the same loading window could create store metadata.
- The worker treated an engine-finished request with no active or completed store future as store-complete, even when that request still had a retrieve in flight.
- A retrieve can finish in one worker step and the scheduler-confirmed abort can arrive in a later step, after the retrieve future has been removed.

## Fix

- Add an explicit `LOADING` state for submitted retrieves:
  `PREFETCHING -> WAITING_FOR_LOAD -> LOADING -> READY`.
- Transition `LOADING -> READY` only after `finished_recving`.
- Do not submit store metadata for new or cached requests while they are `LOADING`.
- Do not wait for a nonexistent store when a loading request is aborted.
- Carry scheduler-confirmed aborted-retrieve IDs to workers and retain them across `get_finished()` calls.
- Suppress store completion for pending retrieves, retrieves completed in the current step, and explicitly marked cross-step aborts.
- Deduplicate handled abort IDs so repeated engine notifications cannot synthesize a later store completion.
- Preserve the current lazy-offload path without marking a nonexistent pending store.

This preserves vLLM's scheduler ownership assertions and enforces the connector invariant that an active retrieve cannot also enter the store-completion lifecycle. This is different from vllm-project/vllm#46304.

## Tests

The refreshed single commit is based on LMCache `dev` at `09bc14c`.

- `pytest -q --confcutdir=tests/v1 tests/v1/test_vllm_mp_adapter.py tests/v1/test_lazy_offload_pending_store.py tests/v1/test_lmcache_mp_connector.py tests/v1/test_mp_connector_mm_keys.py` — 57 passed, 2 skipped. The vLLM-dependent connector modules are skipped on the current macOS development host because vLLM is unavailable there.
- Repository pre-commit hooks on all five changed files — passed, including SPDX, isort, ruff, ruff-format, codespell, and mypy.
- `python -m compileall` on all changed Python files — passed.
- `git diff --check` — passed.

Coverage includes pending-retrieve aborts, same-step retrieve completion plus abort, cross-step retrieve completion then abort, repeated engine notifications, unhealthy-worker handling, new/cached request store suppression while loading, metadata marker delivery, and lazy-offload behavior.

## Production validation

The earlier implementation of this fix was validated with vLLM v0.25.1, MiniMax M3, TP=8, and LMCache MP:

- built and ran one version-consistent LMCache image for the server and all vLLM workers;
- primed a shared prefix, restarted only vLLM to clear its local prefix cache, then disconnected 32 concurrent requests during confirmed remote retrieval;
- LMCache reported 9,216-token retrieves for the abort-stress requests;
- no `AssertionError`, traceback, or EngineCore failure occurred;
- `/health` remained successful and a post-abort completion succeeded.

The current rebased commit preserves that core fix while adapting it to the latest metadata split and lazy-offload changes. It has not yet been rerun on a GPU host; the regression and static checks above were rerun after the rebase.

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -352,6 +352,7 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
                 extra_config=vllm_config.kv_transfer_config.kv_connector_extra_config,
             )
             self.request_trackers: dict[str, LMCacheMPRequestTracker] = {}
+            self._aborted_retrieve_req_ids: set[str] = set()
 
             # GPU block pool reference
             self._gpu_block_pool: "BlockPool | None" = None
@@ -520,6 +521,10 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
         """
         metadata = self._get_connector_metadata()
         assert isinstance(metadata, LMCacheMPConnectorMetadata)
+        if not self.lazy_offload:
+            self.worker_adapter.mark_aborted_retrieves(
+                metadata.aborted_retrieve_req_ids
+            )
 
         request_ids = []
         ops = []
@@ -912,6 +917,8 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
         """
         metadata = LMCacheMPConnectorMetadata()
         metadata.need_flush_before_forward = _has_preemption_reqs(scheduler_output)
+        metadata.aborted_retrieve_req_ids = self._aborted_retrieve_req_ids
+        self._aborted_retrieve_req_ids = set()
 
         self._process_retrieve_requests(metadata)
         self._process_new_requests(scheduler_output, metadata)
@@ -933,6 +940,11 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
             connector_output (KVConnectorOutput): the worker-side
                 connectors output.
         """
+        for request_id in connector_output.finished_recving or ():
+            tracker = self.request_trackers.get(request_id)
+            if tracker is not None and tracker.state == LMCacheMPRequestState.LOADING:
+                tracker.state = LMCacheMPRequestState.READY
+
         if not self.lazy_offload:
             return
         if not self._gpu_block_pool:
@@ -969,6 +981,13 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
             returned by the engine.
         """
 
+        tracker = self.request_trackers.get(request.request_id)
+        load_in_flight = (
+            tracker is not None and tracker.state == LMCacheMPRequestState.LOADING
+        )
+        if load_in_flight and not self.lazy_offload:
+            self._aborted_retrieve_req_ids.add(request.request_id)
+
         params: dict[str, Any] | None = getattr(request, "kv_transfer_params", None)
         return_params: dict[str, Any] | None = {} if params is not None else None
 
@@ -977,9 +996,9 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
             and return_params is not None
             and "cached_token_stats" in params
         ):
-            request_tracker = self._get_request_tracker(request.request_id)
-            num_vllm = request_tracker.num_vllm_hit_tokens
-            num_lmcache = request_tracker.num_lmcache_hit_tokens
+            assert tracker is not None
+            num_vllm = tracker.num_vllm_hit_tokens
+            num_lmcache = tracker.num_lmcache_hit_tokens
             return_params["cached_token_stats"] = {
                 "num_vllm_cached_tokens": num_vllm,
                 "num_lmcache_cached_tokens": num_lmcache,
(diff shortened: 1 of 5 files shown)


sammshen:
sorry if I missed it but where do we set the state to LOADING?

RedHeartSecretMan:
Thanks — `LOADING` is set in `_process_retrieve_requests()` immediately after retrieve metadata is added:

```python
if r_metadata is not None:
    metadata.add_request_metadata(r_metadata)
    request_tracker.state = LMCacheMPRequestState.LOADING
```

`build_connector_meta()` calls `_process_retrieve_requests()` before `_process_new_requests()` and `_process_cached_requests()`, so both store paths observe `LOADING` and skip store submission. The state transitions to `READY` in `update_connector_output()` only after `finished_recving` is reported.

I also added coverage for the cached-request store path and addressed the test return-type review comments in 328bf52.

RedHeartSecretMan:
Follow-up from a true external-cache retrieve reproduction: the previous revision still had a cross-step race that the original tests did not cover.

Sequence:
1. a worker reports `finished_recving` and removes its retrieve future;
2. the scheduler receives an abort before consuming/settling that completion;
3. the abort-related `finished_req_ids` reaches workers in a later step;
4. because the retrieve future is already gone, the adapter synthesized `finished_sending`, and vLLM hit the line-2500 scheduler ownership assertion.

Commit 893bd3e adds an explicit scheduler-to-worker marker for requests aborted while `LOADING`, so the marker survives across `get_finished()` calls and suppresses the nonexistent store completion. It also adds the cross-step regression test.

Validation on vLLM v0.25.1 / MiniMax M3 / TP=8:
- built and ran a full LMCache image from this branch (Python and `c_ops` ABI matched);
- primed 9,216 tokens per worker, reset only vLLM local prefix cache, then disconnected 32 concurrent requests during LMCache retrieve;
- the pre-fix revision reproduced `assert req_id in self.requests` at scheduler.py:2500;
- 893bd3e completed the same stress test, remained healthy, and accepted a post-abort completion;
- targeted tests: 31 passed; ruff: passed.

RedHeartSecretMan:
Additional validation on current head `8f11b93`:

- Targeted connector/adapter tests: **31 passed** against vLLM v0.25.1.
- K3 Buildkite check now passes after guarding the connector-only test when optional vLLM is unavailable.
- Built a version-consistent LMCache image from the PR source and used the same image for the standalone LMCache server and all TP=8 vLLM workers.
- Primed a shared prefix, restarted only vLLM to clear its local prefix cache, then disconnected 32 concurrent requests during remote retrieval.
- LMCache confirmed 9,216-token remote retrieves for the abort-stress requests on the worker ranks.
- No `AssertionError`, `AttributeError`, traceback, or EngineCore failure appeared in either service log; `/health` remained 200 and a post-abort completion succeeded.

The previous `flush_inflight_stores` AttributeError was reproduced as a partial bind-mount/version mismatch and is absent with the consistent image.

RedHeartSecretMan:
Issue #4185 independently reproduced the same abort/completion race on DeepSeek-V4-Flash with H20, and @zhengfeihe confirmed that this PR is the pending fix for that issue.

The current head is mergeable, DCO and K3 tests pass, and the cross-step race has production-style regression coverage. @sammshen, could you or another maintainer review the lifecycle approach and advise what remains before approval?

damodaryekkuluri:
Ran into the same vLLM crash using LMCache. @RedHeartSecretMan thanks for the fix. Anything pending to get this reviewed and merged?

RedHeartSecretMan:
Refreshed this PR onto current `dev` (`09bc14c`) in `b10588b`. GitHub now reports it as mergeable, and the previous conflict is resolved.

This refresh preserves the abort/retrieve lifecycle fix while adapting it to the current metadata split and lazy-offload paths.

Current-head validation:

- targeted regression suite: 57 passed, with 2 tests skipped because the optional vLLM dependency is unavailable in the local environment;
- repository pre-commit hooks passed on all five changed files, including ruff, formatting, codespell, and mypy;
- DCO passed;
- [K3 Build #7409](https://buildkite.com/lmcache/k3-unit-tests/builds/7409) passed;
- [AMD Build #11428](https://buildkite.com/lmcache/unit-tests-amd/builds/11428) passed.

The PR description contains the full validation details and distinguishes the earlier TP=8 GPU stress validation from the checks rerun on this rebased head. All existing reviewer requests remain active. There are no known code, conflict, or CI blockers remaining; this is ready for maintainer review.

damodaryekkuluri:
(reposting @RedHeartSecretMan request)

@sammshen, could you or another maintainer review the lifecycle approach and advise what remains before approval?

PR #4138, opened 2026-07-17
[Feat] Add SageMaker HyperPod L2 adapter for MP mode

### LINKED ISSUE #4114: Add SageMaker HyperPod L2 adapter for MP mode
**Label**

SageMaker HyperPod provides a node-local ai-toolkit daemon for managed tiered KV cache. LMCache currently supports it only through the in-process SageMakerHyperPodConnector.
  
When using LMCacheMPConnector with a standalone lmcache server, there is no HyperPod L2 adapter. As a result, MP mode can use its CPU-memory L1 cache but cannot store KV cache in managed kv cache(ai-toolkit).

Add a SageMakerHyperPodL2Adapter implementing L2AdapterInterface and register it as:

```
  {
    "type": "sagemaker-hyperpod"
  }
```
```  
  lmcache server \
    --port 6555 \
    --l1-size-gb 8 \
    --chunk-size 256 \
    --l2-adapter '{
      "type": "sagemaker-hyperpod",
      "url": "sagemaker-hyperpod://${NODE_IP}:9200",
      "shared_memory_name": "ai_toolkit_cache"
    }'
```

- Continue using the existing in-process connector. This does not support users who explicitly need LMCache MP mode.
- Use another MP L2 backend such as S3 or Redis. This does not use HyperPod's managed cache.

**Additional context**
The in-process mode has been deprecated, and since SageMaker HyperPod supports managed tiered KV cache, this MP connector is needed. In particular, LMCache only supports multi-group in MP mode, so it's required for using models like GLM.

bluayer:

Fixes #4114

Adds an MP-mode L2 storage adapter for the Amazon SageMaker HyperPod ai-toolkit tiered cache daemon. LMCache already has an in-process connector for this backend (`sagemaker_hyperpod_connector.py`), but the new multiprocess (MP) architecture has no HyperPod backend

How it works:

- Store is a plain HTTP `PUT` to the node-local daemon.
- Load: Instead of sending bytes back over HTTP, the daemon returns a **lease** telling where the object lives inside its `/dev/shm` arena, and the adapter copies it straight out of shared memory. `lookup_and_lock`/`unlock` map directly onto lease acquire/release.
- Leases expire after a TTL, so the adapter re-checks expiry after each copy and turns any stale read into a miss.
- The daemon owns object lifecycle (no delete API), so `eviction` config is rejected. A circuit breaker fast-fails while the daemon is unreachable; HTTP errors like 404/504 are normal miss responses and never trip it.

Config example:

```
--l2-adapter '{"type": "sagemaker-hyperpod", "url": "sagemaker-hyperpod://<node-ip>:9200"}'
```

- Validated end-to-end on a real SageMaker HyperPod EKS cluster: Qwen on g6e/L40S and a large quantized MoE model (GLM-5.2-Int4, TP=8 on p5.48xlarge/H100), including fault drills (breaker open/recovery) and concurrent load.

**If applicable**:

### DIFF
--- a/docs/source/mp/l2_storage/remote_and_distributed.rst
+++ b/docs/source/mp/l2_storage/remote_and_distributed.rst
@@ -8,6 +8,7 @@ sharing cache across nodes.
    :maxdepth: 1
 
    bigtable
+   sagemaker_hyperpod
    s3
    hfbucket
    mooncake_store
--- a/docs/source/mp/l2_storage/sagemaker_hyperpod.rst
+++ b/docs/source/mp/l2_storage/sagemaker_hyperpod.rst
@@ -1,0 +1,77 @@
+SageMaker HyperPod
+==================
+
+An L2 adapter backed by the **ai-toolkit** daemon of Amazon SageMaker
+HyperPod, which provides shared memory access for LMCache.
+
+**Required fields:**
+
+- ``url``: ai-toolkit daemon URL (``sagemaker-hyperpod://host:port``).
+
+**Optional fields:**
+
+- ``bucket`` (string, default ``"lmcache"``): ai-toolkit cache namespace.
+- ``shared_memory_name`` (string, default ``"shared_memory"``): Name of
+  shared memory segment.
+- ``max_concurrent_requests`` (int, default ``100``): Maximum concurrent
+  HTTP requests.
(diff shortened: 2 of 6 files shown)


sammshen (approved):
LGTM, thanks for adding docs

ApostaC (approved):
LGTM!

PR #4143, opened 2026-07-17
fix: scope keys_in_request per lookup via ContextVar (#4137)

### LINKED ISSUE #4137: Resident CPU-tier prefix cache entry spuriously measures as a full miss under concurrent write load (not explained by capacity or same-key contention)
## Summary

With `LMCacheConnectorV1` (local CPU backend) serving concurrent requests, a prefix that is confirmed resident (freshly warmed immediately before measurement) still frequently returns full-recompute-level TTFT if any other request is concurrently writing a different prefix to the CPU tier, even with generous capacity headroom and no contention on the same key.

## Environment

- lmcache==0.3.12, vllm==0.13.0
- `vllm serve <model> --enable-prefix-caching --kv-transfer-config '{"kv_connector": "LMCacheConnectorV1", "kv_role": "kv_both"}'`
- `LMCACHE_LOCAL_CPU=True`, `LMCACHE_MAX_LOCAL_CPU_SIZE=20` (GiB), `LMCACHE_CHUNK_SIZE=256`
- Single RTX 4090, single vLLM server process, `ThreadPoolExecutor`-dispatched HTTP client (4 workers)
- Prefix length: 28,000 real tokens (Qwen2.5-0.5B-Instruct)

## Reproduction

1. Maintain a small pool of 2 "resident" prefixes, each re-warmed with a `max_tokens=2` request immediately before I measure it.
2. Interleave concurrent "miss" requests: brand-new, globally-unique 28K-token prefixes, never reused.
3. Dispatch both kinds concurrently (Poisson arrivals, 4 worker threads, hit_ratio≈0.5) against the same server.
4. Measure TTFT for each "hit"-kind request (the one just re-warmed) and compare against the known hit/miss TTFT distributions from isolated, sequential measurement on the same hardware/model (clean, non-overlapping: hit max 95.9ms, miss min 395.7ms).

## Result

At n=150 (66 realized "hit" draws), 80.3% of hit-kind requests measured miss-level TTFT (>150ms; observed range 850-2400ms). I tested and ruled out three candidate causes:

- **Capacity**: raising `LMCACHE_MAX_LOCAL_CPU_SIZE` from 6 to 20 GiB didn't reduce the failure rate. Each request's KV footprint at this prefix length is ~343MB (12,288 bytes/token × ~27,890 real tokens); even 4 concurrent workers plus both resident identities live at once, doubled for margin, is only ~4.1GB, well under the 20GiB cap.
- **Same-key contention**: I instrumented whether another request was concurrently warming/measuring the same resident identity (`same_identity_concurrent`). No significant relationship with failure (Mann-Whitney U, p=0.072), and the trend runs the wrong direction for a same-key-race explanation (failed rows: mean 0.34 concurrent same-key accessors; ok rows: mean 0.62).
- **Sequential timing race**: repeating the exact warm-then-measure pattern 30 times fully sequentially (zero concurrency, zero other traffic) produced 0/30 failures, ruling out an intrinsic async-write-vs-read race independent of concurrency.

What does predict failure: whether any miss request (to an unrelated key) is concurrently in flight during the hit's own measurement window (`concurrent_miss_count`):

| | ok | failed | failure rate |
|---|---:|---:|---:|
| `concurrent_miss_count == 0` | 12 | 8 | 40.0% |
| `concurrent_miss_count >= 1` | 1 | 45 | 97.8% |

Fisher's exact test: odds ratio 67.5, p = 2.9e-7.

The residual 40% failure rate at `concurrent_miss_count == 0` is plausibly a measurement-granularity artifact: the counter is sampled once, immediately before the hit's final request is issued, so a miss that begins microseconds later isn't captured. I haven't confirmed this.

## Root cause

I believe this is a request-scoping bug, not a dict-corruption or capacity issue. In `lmcache/v1/storage_backend/local_cpu_backend.py` (v0.3.12), `LocalCPUBackend` keeps a single, instance-level list of looked-up keys, and its own comment documents an assumption this violates under concurrency:

```python
# to help maintain suffix -> prefix order in the dict
# assumption: only one request is looked up at a time
# (only one worker per cache engine)
self.keys_in_request: List[CacheEngineKey] = []

def contains(self, key, pin=False):
    with self.cpu_lock:
        if key not in self.hot_cache:
            return False
        if pin:
            self.hot_cache[key].pin()
            self.keys_in_request.append(key)   # shared across ALL requests
        return True

def touch_cache(self):
    with self.cpu_lock:
        for key in reversed(self.keys_in_request):
            self.cache_policy.update_on_hit(key, self.hot_cache)
        self.keys_in_request = []              # clears the WHOLE shared list
```

The same pattern recurs in `batched_async_contains` (same file, ~line 211-228), which is on the actual request-serving hot path.

`self.cpu_lock` correctly serializes raw `hot_cache` dict access, so `contains()`/`get_blocking()` should still find a resident key's entry. The corruption is in the eviction bookkeeping: one request's `touch_cache()` can clear another request's just-appended keys before that second request records its own `cache_policy.update_on_hit()` call, silently dropping the "this key was just touched" signal for an unrelated, already-resident key. That makes the key look stale to the eviction policy regardless of how much raw capacity is available, which matches what I saw in the Capacity test above: the failure rate didn't move because the cause was never capacity.

Tracing the caller in `storage_manager.py`: `StorageManager` runs a single dedicated background thread with its own asyncio event loop, and `async_lookup_and_prefetch()` (which awaits `batched_async_contains()`) runs on that loop. So all concurrent requests' lookups execute as interleaved coroutines on a single OS thread, not on separate threads. A `threading.local()` fix wouldn't be enough here since every request shares one thread; a correct fix likely needs per-task scoping (`contextvars.ContextVar`) or explicit per-request key-list threading through the call chain, spanning at least `local_cpu_backend.py` and `storage_manager.py`. `touch_cache()` is also called from `cache_engine.py`, which I haven't inspected yet.

This file has no async device-copy code (no `.to(device, non_blocking=True)`, no CUDA streams/events), so as far as I can tell this is a pure Python scoping bug in LMCache's own request handling, not an upstream PyTorch/CUDA issue.

Happy to share the instrumented repro script if useful.

vedjaw:

Fixes #4137. `LocalCPUBackend` / `LocalDiskBackend` kept a single instance-level `keys_in_request` list under the assumption that only one lookup runs at a time. Under concurrent pin=True lookups (interleaved coroutines on StorageManager's asyncio thread, or concurrent sync lookups on worker threads), one request's `touch_cache()` clears another request's keys before `cache_policy.update_on_hit()` runs. Resident prefixes then look cold to LRU and get evicted under write load, showing up as full-miss TTFT even with ample CPU-tier capacity.

This PR:
- Scopes the key list with `contextvars.ContextVar` (per asyncio task / per thread) in both local backends
- Calls `touch_cache()` from `async_lookup_and_prefetch` when `pin=True`, so the async connector path actually records hits (previously keys were appended but never touched on that path)
- Adds asyncio + threaded regression tests for isolation

- `threading.local()` is not enough here: all async lookups share one OS thread on the storage-manager loop.
- Prometheus `local_cpu_keys_in_request_count` now reads the ContextVar (typically 0 on the scrape thread between requests, same practical behavior as an empty list between lookups).

**If applicable**:

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lmcache/v1/storage_backend/local_cpu_backend.py
+++ b/lmcache/v1/storage_backend/local_cpu_backend.py
@@ -2,6 +2,7 @@
 # Standard
 from concurrent.futures import Future
 from contextlib import nullcontext
+from contextvars import ContextVar
 from typing import TYPE_CHECKING, Any, Callable, List, Optional, Sequence, Union
 import threading
 import time
@@ -37,6 +38,28 @@ if TYPE_CHECKING:
     from lmcache.v1.cache_controller.worker import LMCacheWorker
 
 logger = init_logger(__name__)
+
+# Per-task / per-thread list of keys touched during a pin=True lookup.
+# Must not be instance state: StorageManager serves concurrent lookups as
+# interleaved coroutines on one asyncio thread, so a shared list lets one
+# request's touch_cache() wipe another request's keys (see #4137).
+_keys_in_request_var: ContextVar[Optional[List[CacheEngineKey]]] = ContextVar(
+    "local_cpu_keys_in_request", default=None
+)
+
+
+def _get_keys_in_request() -> List[CacheEngineKey]:
+    keys = _keys_in_request_var.get()
+    if keys is None:
+        keys = []
+        _keys_in_request_var.set(keys)
+    return keys
+
+
+def _take_keys_in_request() -> List[CacheEngineKey]:
+    keys = _keys_in_request_var.get()
+    _keys_in_request_var.set(None)
+    return keys if keys is not None else []
 
 
 class LocalCPUBackend(AllocatorBackendInterface):
@@ -85,11 +108,6 @@ class LocalCPUBackend(AllocatorBackendInterface):
         self.config = config
         self.metadata = metadata
 
-        # to help maintain suffix -> prefix order in the dict
-        # assumption: only one request is looked up at a time
-        # (only one worker per cache engine)
-        self.keys_in_request: List[CacheEngineKey] = []
-
         # Batched message sender for controller communication
         self.batched_msg_sender: Optional[BatchedMessageSender] = None
 
@@ -118,7 +136,7 @@ class LocalCPUBackend(AllocatorBackendInterface):
             lambda: len(self.hot_cache)
         )
         prometheus_logger.local_cpu_keys_in_request_count.set_function(
-            lambda: len(self.keys_in_request)
+            lambda: len(_keys_in_request_var.get() or [])
         )
 
     def __str__(self):
@@ -131,15 +149,15 @@ class LocalCPUBackend(AllocatorBackendInterface):
             if pin:
                 self.hot_cache[key].pin()
                 # vllm lookup sets pin to True
-                self.keys_in_request.append(key)
+                _get_keys_in_request().append(key)
             return True
 
     def touch_cache(self):
         # flip the order of the keys in the request
         with self.cpu_lock:
-            for key in reversed(self.keys_in_request):
+            keys_in_request = _take_keys_in_request()
+            for key in reversed(keys_in_request):
                 self.cache_policy.update_on_hit(key, self.hot_cache)
-            self.keys_in_request = []
 
     def exists_in_put_tasks(self, key: CacheEngineKey) -> bool:
         """
@@ -251,7 +269,7 @@ class LocalCPUBackend(AllocatorBackendInterface):
                 if pin:
                     self.hot_cache[key].pin()
                     # vllm lookup sets pin to True
-                    self.keys_in_request.append(key)
+                    _get_keys_in_request().append(key)
                 num_hit_chunks += 1
         return num_hit_chunks
 
--- a/lmcache/v1/storage_backend/local_cpu_backend.py
+++ b/lmcache/v1/storage_backend/local_cpu_backend.py
@@ -157,6 +157,14 @@ class LocalCPUBackend(AllocatorBackendInterface):
         with self.cpu_lock:
             keys_in_request = _take_keys_in_request()
             for key in reversed(keys_in_request):
+                # A key can leave hot_cache between the lookup that pinned it and
+                # this touch. update_on_hit assumes it is still there -- the LRU
+                # policy calls hot_cache.move_to_end(key), which raises KeyError --
+                # so skip anything that is gone. This was previously masked: with a
+                # single shared list, the first touch_cache() consumed every
+                # request's keys, so the rest were dropped rather than touched.
+                if key not in self.hot_cache:
+                    continue
                 self.cache_policy.update_on_hit(key, self.hot_cache)
 
     def exists_in_put_tasks(self, key: CacheEngineKey) -> bool:
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -2,6 +2,7 @@
 # Standard
 from concurrent.futures import Future, ThreadPoolExecutor
 from contextlib import nullcontext
+from contextvars import ContextVar
 from typing import TYPE_CHECKING, Any, Callable, List, Optional, Sequence
 import asyncio
 import os
@@ -36,6 +37,27 @@ if TYPE_CHECKING:
 logger = init_logger(__name__)
 
 _DEFAULT_THREAD_COUNT = 4
+
+# Per-task / per-thread list of keys touched during a pin=True lookup.
+# Shared instance state is incorrect under concurrent lookups on one
+# asyncio thread (see #4137 / LocalCPUBackend).
+_keys_in_request_var: ContextVar[Optional[List[CacheEngineKey]]] = ContextVar(
+    "local_disk_keys_in_request", default=None
+)
+
+
+def _get_keys_in_request() -> List[CacheEngineKey]:
+    keys = _keys_in_request_var.get()
+    if keys is None:
+        keys = []
+        _keys_in_request_var.set(keys)
+    return keys
+
+
+def _take_keys_in_request() -> List[CacheEngineKey]:
+    keys = _keys_in_request_var.get()
+    _keys_in_request_var.set(None)
+    return keys if keys is not None else []
 
 
 # TODO(Jiayi): handle cases where cache is repetitvely prefetched.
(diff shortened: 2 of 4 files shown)


DongDongJu:
Please check the ut error
```

FAILED tests/v1/storage_backend/test_local_cpu_backend.py::TestLocalCPUBackendAllocatorAlignment::test_concurrent_pin_lookups_isolate_keys_in_request - KeyError: CacheEngineKey(model_name='test_model', world_size=3, worker_id=0, chunk_hash=-5841188751982805022, dtype=torch.bfloat16, request_configs={}, tags=None, _dtype_str='bfloat16')
--
  | !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
  | ======================================= 1 failed, 3946 passed, 181 skipped, 1 xpassed, 294 warnings in 562.79s (0:09:22) =======================================
  | buildkite-test-collector - WARNING - No BUILDKITE_ANALYTICS_TOKEN environment variable present
  | /opt/venv/lib/python3.12/site-packages/_pytest/unraisableexception.py:67: PytestUnraisableExceptionWarning: Exception ignored in: <function BaseEventLoop.__del__ at 0x7f4b5a45d1c0>
  |  
  | Traceback (most recent call last):
  | File "/usr/lib/python3.12/asyncio/base_events.py", line 728, in __del__
  | self.close()
  | File "/usr/lib/python3.12/asyncio/unix_events.py", line 68, in close
  | super().close()
  | File "/usr/lib/python3.12/asyncio/selector_events.py", line 104, in close
  | self._close_self_pipe()
  | File "/usr/lib/python3.12/asyncio/selector_events.py", line 111, in _close_self_pipe
  | self._remove_reader(self._ssock.fileno())
  | File "/usr/lib/python3.12/asyncio/selector_events.py", line 298, in _remove_reader
  | key = self._selector.get_key(fd)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/usr/lib/python3.12/selectors.py", line 190, in get_key
  | return mapping[fileobj]
  | ~~~~~~~^^^^^^^^^
  | File "/usr/lib/python3.12/selectors.py", line 71, in __getitem__
  | fd = self._selector._fileobj_lookup(fileobj)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | File "/usr/lib/python3.12/selectors.py", line 225, in _fileobj_lookup
  | return _fileobj_to_fd(fileobj)
  | ^^^^^^^^^^^^^^^^^^^^^^^
  | File "/usr/lib/python3.12/selectors.py", line 42, in _fileobj_to_fd
  | raise ValueError("Invalid file descriptor: {}".format(fd))
  | ValueError: Invalid file descriptor: -1
  |  
  | Enable tracemalloc to get traceback where the object was allocated.
  | See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
  | warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
  | [2026-07-17 14:35:07,716] LMCache WARNING: MemoryObj at 0 is being garbage collected with ref_count=1, pin_count=0. This indicates ref_count_down()/unpin() was not called properly. (memory_management.py:681:lmcache.v1.memory_management)
  | [2026-07-17 14:35:07,717] LMCache WARNING: MemoryObj at 0 is being garbage collected with ref_count=1, pin_count=0. This indicates ref_count_down()/unpin() was not called properly. (memory_management.py:681:lmcache.v1.memory_management)
  | Exception ignored in atexit callback: <bound method MeterProvider.shutdown of <opentelemetry.sdk.metrics._internal.MeterProvider object at 0x7f47c6dd4440>>
  | Traceback (most recent call last):
  | File "/opt/venv/lib/python3.12/site-packages/opentelemetry/sdk/metrics/_internal/__init__.py", line 541, in shutdown
  | raise Exception(
  | Exception: MeterProvider.shutdown failed because the following metric readers failed during shutdown:
  | PrometheusMetricReader: KeyError(<opentelemetry.exporter.prometheus._CustomCollector object at 0x7f47c6dd4380>)
  | Exception ignored in atexit callback: <bound method MeterProvider.shutdown of <opentelemetry.sdk.metrics._internal.MeterProvider object at 0x7f47c55e5a30>>
  | Traceback (most recent call last):
  | File "/opt/venv/lib/python3.12/site-packages/opentelemetry/sdk/metrics/_internal/__init__.py", line 541, in shutdown
  | raise Exception(
  | Exception: MeterProvider.shutdown failed because the following metric readers failed during shutdown:
  | PrometheusMetricReader: KeyError(<opentelemetry.exporter.prometheus._CustomCollector object at 0x7f47c55e5580>)
  | [W717 14:35:11.111908935 CudaIPCTypes.cpp:16] Producer process has been terminated before all shared CUDA tensors released. See Note [Sharing CUDA tensors]
  | ERROR: setup-lmcache-only-env.sh failed at line 33 (exit code 1)

```

vedjaw:
@DongDongJu Thanks — root-caused and fixed in 59a2662.

The `KeyError` is a latent crash this PR exposed rather than one it created. `touch_cache()` has always ended with:

```python
for key in reversed(keys_in_request):
    self.cache_policy.update_on_hit(key, self.hot_cache)
```

and the LRU policy's `update_on_hit` calls `hot_cache.move_to_end(key)`, which raises `KeyError` for a key that is no longer there. That was previously masked: with a single shared `self.keys_in_request`, whichever request called `touch_cache()` first consumed *everyone's* keys and cleared the list, so the remaining requests iterated nothing. Scoping the list per request means each request now genuinely touches its own keys — including a key that has left `hot_cache` between the lookup that pinned it and the touch, which is the one that blew up.

Fix is to skip keys that are gone, since touching an absent key has no meaning. Added `test_touch_cache_skips_keys_evicted_after_lookup`, which pins a key, drops it from `hot_cache` behind the lookup's back, and asserts `touch_cache()` does not raise.

One caveat worth stating plainly: I could not run the suite locally — it needs a torch environment I don't have on this machine — so I verified the mechanism by reading `LRUCachePolicy.update_on_hit` and the pre/post behaviour of the shared list rather than by reproducing. Please do re-run CI; if the `KeyError` persists I'd want the full traceback, since that would mean it is coming from a different `update_on_hit` path than the LRU one.

PR #4144, opened 2026-07-17
[feat]: strided fused-K re-RoPE for vLLM packed KV

deng451e:
vLLM's packed KV layout stores K and V interleaved per head as [.., num_kv_heads, 2*head_size] (lmcache formats NL_X_NB_NH_BS_TWO_HS / NL_X_NB_BS_NH_TWO_HS, kv_size=1). CacheBlend's re-RoPE only rotates the K half, but rotary_embedding_k_fused assumes a contiguous [.., num_kv_heads, head_size] key buffer.

Add rotary_embedding_k_fused_strided: the rotary device functions take an explicit per-head element stride (head_stride), so K can be rotated in place over the packed buffer by passing head_stride = 2*head_size (V, the trailing head_size, is left untouched). The original rotary_embedding_k_fused delegates with head_stride = head_size, so all existing callers are unchanged.

blend_v3._apply_cb_rope_batched detects the fused-packed formats and routes to the strided kernel instead of rejecting them as MLA.

tests/v1/test_fused_rerope.py validates the strided kernel against a copy-based reference (contiguous gather -> existing kernel -> scatter): K bit-identical, V byte-identical, and the strided path is 2.4-3.6x faster (no gather/scatter of the K half).

**If applicable**:

### DIFF
--- a/csrc/engine_kv_format.h
+++ b/csrc/engine_kv_format.h
@@ -153,3 +153,12 @@ LMC_KV_FORMAT_HD constexpr bool is_mla(EngineKVFormat f) {
   return f == EngineKVFormat::NL_X_NB_BS_HS ||   // vLLM MLA
          f == EngineKVFormat::NL_X_NBBS_ONE_HS;  // SGLang MLA
 }
+
+// vLLM fused K/V: K and V packed into the trailing dim (2 * head_size) with no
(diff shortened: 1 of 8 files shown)


ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!!

PR #4147, opened 2026-07-18
refactor(platform): Refactor IPC wrapper cls registry to DeviceSpec

maobaolong:

Refactor: remove `lmcache/v1/platform/_registry.py` and fold the KV-cache IPC
wrapper lookup into `DeviceSpec`.

Before this PR the platform package carried two parallel registries: `DeviceSpec`
(added recently) for per-accelerator metadata / capabilities, and a separate
lazy-scan registry in `_registry.py` that indexed `DeviceIPCWrapper` subclasses
by their `device_type` ClassVar. Two places to look, two places to keep in sync,
and the scan-and-index-on-first-use logic was more machinery than the problem
warranted.

This PR keeps a single source of truth:

- Each `DeviceSpec` subclass (`CudaDeviceSpec`, `CpuDeviceSpec`,
  `MusaDeviceSpec`) declares its default wrapper via a new
  `ipc_wrapper_cls` attribute.
- `lmcache.v1.platform.resolve_kv_wrapper_factory(device_type)` replaces
  `platform._registry.get_kv_wrapper_factory(...)` — same signature, same
  return contract, just looks it up off the registered spec.
- `_registry.py` and its lazy-discovery scaffolding (threading lock, `_is_default_wrapper`
  flag, `snapshot` / `restore` helpers) are deleted.
- `RawCudaIPCWrapper` stays off the spec, so TRT-LLM's direct-instantiation path
  is unchanged.

Call sites in `vllm_multi_process_adapter.py` and
`multiprocess/transfer_context/worker_transfer.py` are updated. The design doc
in `docs/design/v1/multiprocess/device_ipc_wrapper_design.md` is refreshed to
match. Existing tests that used `_registry.snapshot / restore` were migrated to
patch `DeviceSpec.ipc_wrapper_cls` directly.

- Pure refactor, no behaviour change. Adding a new accelerator still means
  dropping a sub-package under `platform/<device>/` — now with a `DeviceSpec`
  subclass that sets `ipc_wrapper_cls` instead of a wrapper class with a
  `device_type` ClassVar.
- `pre-commit run --all-files` is clean locally.
- Affected tests pass locally:
  `tests/v1/multiprocess/test_engine_driven_transfer.py`,
  `tests/v1/platform/test_cache_context_dispatch.py`,
  `tests/v1/test_musa_mp_handle.py`.

**If applicable**:


lmcache/v1/platform/cpu/__init__.py
```diff
@@ -1,8 +1,48 @@
...
+
+# First Party
+from lmcache.v1.platform.base_device_spec import DeviceSpec
+
+if TYPE_CHECKING:
+    # First Party
+    from lmcache.v1.platform.base_ipc_wrapper import DeviceIPCWrapper
+
+
+class CpuDeviceSpec(DeviceSpec):
```
chunxiaozheng, line 25: @maobaolong Thanks for your contribution, but this has a small question, the `_detect_device` will scan all sub-classes of `DeviceSpec`, and the `CpuDeviceSpec` will be in the first position, but in the original behavior, `cpu` should be in the end position.

lmcache/v1/platform/__init__.py
chunxiaozheng, line 52: @maobaolong `_DEVICE_REGISTRY` will include `CpuDeviceSpec`

lmcache/v1/platform/cpu/__init__.py
maobaolong: Good catch, thanks! Fixed in e9518283: the auto-detection loop now skips cpu so accelerators still win, and the StubCPUDevice branch keeps handling the no-accelerator case. DEVICE_TYPE=cpu remains available as an explicit opt-in.

lmcache/v1/platform/__init__.py
maobaolong: Right — that's exactly the trigger. Addressed together with the cpu comment above in e9518283 by skipping cpu inside the _detect_device loop, so having CpuDeviceSpec in _DEVICE_REGISTRY no longer changes the auto-detection order.

maobaolong:
@chunxiaozheng Thanks for your review suggestion, it really helps, now i updated my branch to address these comments, PTAL.

yoo-kumaneko (approved):
LGTM in general. 
The `continue` for `cpu` in `_detect_device()`'s auto-detection loop only stays safe because `CpuDeviceSpec` doesn't override `is_available()` (inherits the base `-> False`). Worth a one-line comment on `CpuDeviceSpec` making that invariant explicit, so a future override doesn't silently break accelerator auto-detection. Not a blocker though.

chunxiaozheng (approved):
LGTM

Issue #4025, opened 2026-07-06
LMCache 2026 Q3 Roadmap


ApostaC:
# LMCache 2026 Q3 Roadmap 

In Q3, we are evolving towards a more distributed, more performant, and smarter KV cache management stack based on the MP mode. 
In the meantime, we will start the deprecation process of the non-MP mode since Q3.

## 1. Distributed KV Cache: 

**Objective: efficiently handle KV caches that are distributed across multiple nodes**
- [ ] “Coordinator mode” with native language (c++/rust) implementation
- [ ] P2P KV cache sharing stress test and performance optimization
  - [ ] #4041 
- [ ] Native PD disaggregation in MP mode

## 2. Serving Engine Integration: 

**Objective: Ensure a performant and stable integration with popular serving engines**

- [ ] SGLang MP mode async store/retrieve support
- [ ] SGLang HiCache integration 
- [ ] Integrate with the latest vLLM interfaces
- [ ] Omni model support (vLLM-omni, SGL-omni)

## 3. KV Cache Management Interface

**Objective: enable users to control KV caches manually**

- [ ] Pin, delete, and move KV caches ( #3978 , #3990 )
- [ ] Inspect and update KV caches
- [ ] Configure the management policy (eviction, store, and prefetch policies) during runtime

## 4. Ecosystem Support

**Objective: support and integrate with the latest innovations in the ecosystem**

- [ ] Step-by-step guide for new accelerator/platform integration. ( #4004 )
- [ ] New accelerators 
  - [ ] MACA ( #3856 )
  - [ ] Ascend ( #3968 )
  - [ ] MooreThreads 
  - [ ] AWS Trainium
- [ ] Step-by-step guide for new storage integration
- [ ] New storage systems
  - [ ] MinIO
  - [ ] 3FS 
  - [ ] io_uring

## 5.  Production Deployment

**Objective: have a cloud-native way to deploy MP mode smoothly** 

- [ ] Remove the hostIPC & privileged mode requirement from the LMCache operator
- [ ] Fault tolerance and high availability features for the LMCache coordinators

## 6. KV Cache Optimization Algorithms

**Objective: make KV cache smaller, faster, and smarter**

- [ ] CacheBlend mode and deployment guide
- [ ] More KV Cache compression algorithms

## 7. Core Optimizations

**Objective: improve the LMCache internal modules**

- [ ] Profiling & benchmarking the LMCache core 
[...]

PR #4151, opened 2026-07-18
[MP][Coordinator] Commit OpenAPI contract with drift-guard test

theap06:

Freezes the MP coordinator's REST surface as a committed, reviewable contract. `docs/design/v1/mp_coordinator/openapi.json` is exported from the FastAPI app via a new regen tool (`python -m lmcache.v1.mp_coordinator.utils.export_openapi`), and a drift test fails whenever the app and the committed file disagree — so every API-surface change lands as an explicit contract diff instead of silent drift.

This is groundwork for Q3 native coordinator mode (Refs #4025): porting the coordinator to a native implementation needs a frozen compatibility baseline, and "parity" becomes a checkable statement — serving this contract. It is useful independently of that decision as a regression guard on the Python coordinator.

- No behavior change; the diff is a regen tool, the generated spec, the drift test, and a README section.
- The guard already proved itself while preparing this PR: rebasing over #3990 changed the surface (`POST /cache/delete`), the test failed, and regenerating produced the reviewable 119-line contract diff.
- The spec is generated with fastapi 0.111 / pydantic 2.13. A future FastAPI/pydantic upgrade may legitimately shift schema output and trip the test — that is intended (an upgrade is a wire-contract event); regenerate and review the diff.

**If applicable**:

### DIFF
--- a/docs/design/v1/mp_coordinator/README.md
+++ b/docs/design/v1/mp_coordinator/README.md
@@ -35,6 +35,21 @@ POSTs to that mp server's **specific** existing endpoint (e.g. `/pin`,
 `/quota`). There is no generic command channel and no per-instance connection
 state — just an HTTP call to the relevant resource.
 
+## API contract
+
+`openapi.json` (this directory) is the coordinator's committed REST contract,
+exported from the FastAPI app. `tests/v1/mp_coordinator/test_openapi_contract.py`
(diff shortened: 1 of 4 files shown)


ApostaC:
Hi @theap06 , thanks for your interest! 
I think it's a bit pre-mature to freeze the coordinator interface for now since it's still quickly evolving.

PR #4162, opened 2026-07-20
Support vLLM's packed (fused-K/V) layout `NL_X_NB_BS_NH_TWO_HS` in CacheBlend V3

shawnguyen-tensormesh:
**Follow-up to #4128** (`[Fix] Add kernels for vLLM-packed KV cache format`), which added the packed layout to the **bulk/MP transfer** path (`mp_mem_kernels.cu`, the `nl_x_nb_bs_nh_two_hs` format spec, detectors, and the Python fallback). This PR extends the *same* packed layout to the two **CacheBlend-specific** paths that #4128 did not touch: the per-token scatter kernel (`mem_kernels.cu`, the sibling of the `mp_mem_kernels.cu` #4128 fixed) and the re-RoPE step (`blend_v3.py`). No file overlap with #4128.

### Summary

Teaches CacheBlend V3 to handle the fused-K/V NHD layout (`[NB, BS, NH, 2, HS]`, `kv_size == 1`) that vLLM now produces after [vllm#44455](https://github.com/vllm-project/vllm/pull/44455). After #4128 the bulk transfer already handled this (prefix reuse / offload work), but CacheBlend V3's non-prefix path still assumed a split K/V (`kv_size == 2`) in two spots - the per-token scatter kernel and the re-RoPE step - and rejected the packed layout, so ~half of cross-engine CacheBlend retrieves failed and fell back to recompute. This PR adds packed support to both, restoring **non-prefix CacheBlend reuse** for packed-layout models (e.g. `gpt-oss-120b`) on current vLLM. Bulk transfer is untouched.

### What changed

```mermaid
flowchart LR
    subgraph before["Before (packed layout)"]
        b1["re-RoPE: shape[0]!=2<br/>→ RuntimeError 'MLA unsupported'"]
        b2["scatter: no case<br/>→ RuntimeError 'Unsupported EngineKVFormat'"]
    end
    subgraph after["After"]
        a1["re-RoPE: un-fold [.,nh,2,hs],<br/>rotate K half, write back"]
        a2["scatter: fused branch + k_or_v_size=1<br/>+ H2D/D2H dispatch"]
    end
    before --> after
```
The format spec intentionally reports the packed layout as fused (`kv_size == 1`, `head_size = 2·HS`) so the NHD-fused transfer kernels move the contiguous per-head `[K, V]` blob unchanged - right for bulk store/load. CacheBlend is the only consumer that must address K and V *separately* (re-RoPE rotates K only; the per-token scatter writes specific slots). So the fix is scoped to those two paths and leaves the fused transfer representation - and all working store/load/offload - alone. Rather than blindly overlaying the packed layout as split (which the transfer kernels can't move), the scatter reuses the existing fused/MLA addressing and the re-RoPE un-folds only where it needs the K half.

### Testing

- **Kernel round-trip** (`test_multi_layer_kernel_packed_nhd`): exact K/V match at `head_size` 64 (gpt-oss) and 128, tokens {256, 500, 1024, 8000}.
- **End-to-end** on `gpt-oss-120b`, TP=2, `--engine-type blend --enable-segmented-prefix`:
  - `lmcache_blend_retrieve_failures_total`: **~50% → 0**; non-prefix reuse fires (~80k reused tokens/burst); `MLA unsupported` / `Unsupported EngineKVFormat` gone.
  - **Correctness** (known-topic oracle): CacheBlend reuse produces answers that correctly identify the reused documents' topics - i.e. the reused KV is semantically correct, not garbage-that-doesn't-error.
  - **Concurrency** 1→32: engine stable (0 errors); reuse-ON throughput ~16× (RAG) / ~8× (prefix-shift) vs recompute at load.
  - **Tiers**: verified with reuse served from both L1 (RAM) and an L2 (SSD) backend.

### DIFF
--- a/csrc/mem_kernels.cu
+++ b/csrc/mem_kernels.cu
@@ -256,9 +256,16 @@ __device__ __forceinline__ int64_t page_buffer_offset(
            k_or_v * block_size * scalars_per_token +
            block_offset * scalars_per_token + scalar_offset;
   }
-  // MLA formats: vLLM (NL_X_NB_BS_HS) and SGLang (NL_X_NBBS_ONE_HS)
+  // MLA formats: vLLM (NL_X_NB_BS_HS) and SGLang (NL_X_NBBS_ONE_HS); plus the
+  // fused-K/V NHD packed layout NL_X_NB_BS_NH_TWO_HS ([NB, BS, NH, 2*HS]).
+  // The latter is token-contiguous with a single fused K/V axis (kv_size == 1,
+  // scalars_per_token == NH * 2 * HS), so it addresses like MLA — K and V ride
+  // together in each token's contiguous span. (The HND-packed sibling
+  // NL_X_NB_NH_BS_TWO_HS is heads-outermost and would need its own offset case;
+  // not wired here — CB v3 only routes the NHD-packed layout.)
   else if constexpr (format == EngineKVFormat::NL_X_NB_BS_HS ||
-                     format == EngineKVFormat::NL_X_NBBS_ONE_HS) {
+                     format == EngineKVFormat::NL_X_NBBS_ONE_HS ||
+                     format == EngineKVFormat::NL_X_NB_BS_NH_TWO_HS) {
     return token_idx * scalars_per_token + scalar_offset;
   }
   // vllm flash attention (HND) — physical: [2, NB, NH, BS, HS]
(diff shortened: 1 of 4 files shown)


deng451e:
Packed KV format is supported in this PR #4144, PTAL.

PR #4164, opened 2026-07-20
[RawBlock] Re-enable uring_cmd batched recovery header validation

nayeonikim:

Re-enables the `use_uring_cmd` path in batched recovery header
validation. Scope is restart recovery read path only.

Changes:

- `RawBlockCore._read_slot_headers`: drop the `not self.use_uring_cmd`
  guard so the batched path is used for both `io_uring` and
  `io_uring_cmd` when `n > 1`.
- `_read_slot_header_batch` docstring: remove the "uring_cmd
  passthrough falls back to serial reader" note.
- Unit test rename:
  `test_validate_loaded_entries_uring_cmd_keeps_sequential_reader`
  → `..._uses_batched_reader`, with the assertion inverted.
- `raw_block_recovery_bringup_bench.py`:
  - Add `io_uring_cmd` as a benchmark engine option.
  - Add `--cmd-device-path` so prepare can run on a block device while
    measurement runs on the NVMe char device.
  - Extend variant tuple to
    `(label, io_engine, use_uring_cmd, recovery_threads)`.
  - Omit `device_path` on the synthetic checkpoint payload to avoid
    block/char path mismatch in the benchmark fixture (fixture-only,
    not a production schema change).
- Docs (`docs/design/v1/distributed/l2_adapters/raw_block.md`,
  `docs/source/mp/l2_storage/raw_block.rst`): note that both
  `io_uring` and `io_uring_cmd` recovery use batched header
  validation.

Stacked on **#3891** (read-path `uring_cmd` alignment prerequisite)
and **#3835** (batched recovery header validation infra). Marked as
**draft** until both merge.

**If applicable**:

### DIFF
--- a/benchmarks/storage_backend_io/README.md
+++ b/benchmarks/storage_backend_io/README.md
@@ -170,6 +170,32 @@ print(f"rust vs local: {(rust / local - 1.0) * 100.0:+.2f}%")
 PY
 ```
 
+## Raw Block Recovery Bringup Benchmark
+
+`raw_block_recovery_bringup_bench.py` measures synthetic `RawBlockCore` restart
+recovery. The prepare step writes checkpoint metadata and per-slot headers only;
+it does not write full payload contents. Use it to compare checkpoint loading and
+slot-header validation across recovery thread counts.
+
+```bash
+python benchmarks/storage_backend_io/raw_block_recovery_bringup_bench.py \
+  --device-path /dev/nvme1n1 \
+  --cache-space-gb 250 \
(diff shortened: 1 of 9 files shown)


nayeonikim:
Hi, @DongDongJu @zhengfeihe  
this follows the recovery work in #3835 with daegyu: after the passthrough transfer-size and alignment fixes in #3882 and #3891, this PR enables batched `io_uring_cmd` recovery that was previously kept serial.
Could you take a look when you have a chance?

PR #4172, opened 2026-07-21
[Cacheblend][Coordinator] Enable global token matching in l1 using p2p

ruizhang0101:

Let a coordinator CacheBlend match be fetched directly from a peer's L1 over the P2P transfer channel (previously shared-L2 only):

- P2P serving read-locks and returns every L1-resident key (sparse, gap-tolerant), not just the contiguous prefix; requesters still apply their own trim policy — no wire change.
- `skip_l2` honored on the SPARSE prefetch branch (prevents recursive peer fetches).
- Blend fleet directory enabled by `--coordinator-url` (resolved once at config parsing), not only `LMCACHE_COORDINATOR_URL`.

Verified e2e across two hosts (gpt-oss-20b + CacheBlend connector, NIXL).

**If applicable**:


ApostaC (approved):
LGTM!

deng451e (approved):
LGTM

PR #4177, opened 2026-07-21
Refine k3 tests path filtering

VincyZhang:

This PR rewrites the k3 Buildkite path filter to be pipeline-aware instead of using a single global trigger rule.

- upload-pipeline.sh now passes the pipeline file into the filter, so the skip decision can depend on which k3 test is being uploaded.
- path-filter.sh now classifies changed files by shared runtime surface and suite-specific surfaces, so unrelated k3 test suites do not fan out unnecessarily.
- README.md documents the new routing model for `unit`, `integration`, `correctness`, `multiprocess`, `comprehensive`, `blend`, `sglang`, and `xpu`.
- The skip log/annotation text was updated to reflect pipeline-specific relevance instead of “only trivial files changed”.

- The new behavior keeps trivial docs-only changes skippable.
- `force-ci` still bypasses the skip logic.
- Shared runtime changes still fan out to the relevant suites, but suite-local `.buildkite/k3_tests/<suite>/...` changes now only affect that suite.

**If applicable**:


VincyZhang:
@hlin99  @maobaolong Please help to review.

maobaolong:
@VincyZhang Thanks for the hard working on this, will take a look these week.

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +49,151 @@ _path_filter_is_trivial() {
...
+        benchmarks/*) return 0 ;;
+        examples/disagg_prefill*/*) return 0 ;;
+        .buildkite/k3_tests/comprehensive/*) return 0 ;;
+    esac
+    return 1
+}
+
+_path_filter_is_blend_surface() {
+    case "$1" in
+        .buildkite/k3_harness/setup-blend-env.sh) return 0 ;;
```
hlin99: `.buildkite/k3_harness/*` is already included in the shared runtime surface, so these suite-specific harness entries are redundant and can never affect the result.

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +49,151 @@ _path_filter_is_trivial() {
...
+        .buildkite/k3_tests/blend/*) return 0 ;;
+    esac
+    return 1
+}
+
+_path_filter_is_sglang_surface() {
+    case "$1" in
+        .buildkite/k3_harness/setup-sglang-env.sh) return 0 ;;
```
hlin99: ditto

hlin99 (commented):
Hi @VincyZhang  The overall direction makes sense! 
I have a highlevel question about the current allowlist-based definition of the shared runtime surface may cause tests to be silently skipped.

As I understand it, changed files are currently classified into several layers:

1. trivial files that do not require k3 tests;
2. files explicitly known to be unrelated to a particular device or suite;
3. files explicitly known to be relevant to a particular pipeline; and
4. shared runtime files identified by an allowlist.

The problem with the fourth layer is that any non-trivial path that is not covered by these rules is treated as unrelated. We already have one concrete example: `setup_extensions/**` is used by `setup.py`, but it is missing from the shared runtime allowlist. Similar gaps may appear whenever a new source or build directory is added without a corresponding update to `path-filter.sh`. 
In that case, all k3 pipelines could be skipped for the new path.

Since running extra tests is preferable to missing relevant coverage, could we remove the shared-runtime allowlist and instead treat every remaining non-trivial file as shared and relevant by default? In other words, after handling trivial files, explicitly unrelated device/suite paths, and explicitly relevant suite paths, the fallback should be to run the pipeline rather than skip it. This would preserve the targeted exclusions while making the filter safer for paths that are new or not yet classified.

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +49,151 @@ _path_filter_is_trivial() {
...
+        .buildkite/k3_harness/*) return 0 ;;
+        .buildkite/k3_tests/common_scripts/*) return 0 ;;
+    esac
+    return 1
+}
+
+_path_filter_is_unit_surface() {
+    case "$1" in
+        .buildkite/k3_tests/unit/*) return 0 ;;
+        tests/v1/platform/musa/*|tests/v1/platform/rbln/*|tests/v1/platform/xpu/*)
```
maobaolong: how to handle the scenario a new added device with its new added test UTs? It would be nice if we don't need to update this file while new device added..

maobaolong:
@VincyZhang I agree with the view point of @hlin99 .

We'd better to define a black list rather than white list, because we are all not sure whether the new added file should affect xpu or not. WDYT?

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: Adopt this rule: detect trivial changes, clearly distinguish relevant / irrelevant scenarios for each test, and fallback to running all tests as a safety net—that is sufficient. This eliminates the need to maintain an allowlist while also avoiding redundancy.

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: same solution as above.

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: It can be extracted by `"$changed_file" == tests/v1/platform/*/*`, no need to list them all, and new device added will not influence this condition.
For now, the `tests/v1/platform/xxx.py` will still trigger full unit test suites, while `tests/v1/platform/device/xxx.py` will only trigger the corresponding device related tests.

VincyZhang:
> @VincyZhang I agree with the view point of @hlin99 .
> 
> We'd better to define a black list rather than white list, because we are all not sure whether the new added file should affect xpu or not. WDYT?

Agree. Actually the first three rules with a general fallback is enough to cover the requirement.

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +53,188 @@ _path_filter_is_trivial() {
...
+    # 1) trivial files are always skipped.
+    if _path_filter_is_trivial "$changed_file"; then
+        return 0
+    fi
+
+    case "$pipeline_kind" in
+        unit)
+            case "$changed_file" in
+                tests/v1/platform/*)
+                    # skip all files under second-level platform dirs
```
hlin99, line 69: is this unit for cuda unit test? if so, shall we return 1 if files under platform/cuda is changed?

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +53,188 @@ _path_filter_is_trivial() {
...
+        *k3_tests/multiprocess/pipeline.yml) echo multiprocess ;;
+        *k3_tests/blend/pipeline.yml) echo blend ;;
+        *k3_tests/sglang/pipeline.yml) echo sglang ;;
+        *k3_tests/xpu/pipeline.yml) echo xpu ;;
+        *k3_tests/comprehensive/pipeline.yml) echo comprehensive ;;
+        *) echo generic ;;
+    esac
+}
+
+_path_filter_file_affects_pipeline() {
```
hlin99: seems no use of this function?

hlin99 (commented):
otherwise LGTM

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: Right, I also updated xpu/amd filter with the same pattern

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: Good catch! Fixed.

VincyZhang:
Currently each vendor(device) will maintain it's own k3s_tests/ folder as well as it's own filter scripts. In this PR I use XPU/AMD as examples and leave placeholder for others.

.buildkite/k3_tests/amd/filter.sh
```diff
@@ -0,0 +1,30 @@
...
+    local changed_file="$1"
+
+    case "$changed_file" in
+        tests/v1/platform/rocm/*|lmcache/v1/platform/rocm/*|.buildkite/k3_tests/amd/*)
+            return 1
+            ;;
+        tests/v1/platform/*|lmcache/v1/platform/*)
+            # Keep generic platform roots as non-targets for the ROCm pipeline;
+            # only the ROCm subtree should trigger it.
+            return 0
```
andyluo7, line 16: Shared platform code is used by ROCm too. This would skip AMD CI for changes such as `lmcache/v1/platform/base/event_ipc.py` and its tests. Please skip only known vendor-specific subtrees; keep shared platform code AMD-relevant.

hlin99: right.
Any change to files directly under  lmcache/v1/platform/  or  lmcache/v1/platform/base  needs to be triggered on any device, as those changes suppose to be common

correspondingly, any change to files directly under  tests/v1/platform/  or  tests/v1/platform/base(we don't have such folder, but would expect one in near future) should do the same.

VincyZhang: Got your points. Refine xpu/amd filter in commit https://github.com/LMCache/LMCache/pull/4177/commits/60dad0a18486371fc4f224519f3f7f31869bf58d

.buildkite/k3_tests/amd/filter.sh
```diff
@@ -0,0 +1,39 @@
...
+    local changed_file="$1"
+
+    case "$changed_file" in
+        .buildkite/k3_tests/amd/*|.buildkite/k3_tests/common_scripts/*)
+            return 1
+            ;;
+        .buildkite/k3_tests/*/*)
+            return 0
+            ;;
+        tests/v1/platform/base/*|tests/v1/platform/rocm/*|lmcache/v1/platform/base/*|lmcache/v1/platform/rocm/*)
```
andyluo7: ROCm still inherits `CudaDeviceSpec`, so changes under `lmcache/v1/platform/cuda/*` can affect AMD. Please keep that subtree and its tests AMD-relevant, with a path-filter test covering this case.

andyluo7 (changes requested):
One remaining AMD coverage gap.

hlin99 (approved):
LGTM

.buildkite/k3_tests/amd/filter.sh
VincyZhang: Resolved, please review again. @andyluo7

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +53,208 @@ _path_filter_is_trivial() {
...
+        integration)
+            case "$changed_file" in
+                .buildkite/k3_tests/integration/*|.buildkite/configs/local_cpu.yaml|.buildkite/configs/local_disk.yaml)
+                    return 1
+                    ;;
+                examples/*)
+                    # Example-only changes should not trigger integration tests.
+                    return 0
+                    ;;
+                tests/*|.buildkite/k3_tests/*)
```
andyluo7, line 113: [P1] Keep shared K3 scripts relevant

This catch-all also matches .buildkite/k3_tests/common_scripts/*, so changes to helpers.sh, path-filter.sh, or upload-pipeline.sh make integration, correctness, multiprocess, comprehensive, blend, and SGLang skip. Please handle common_scripts/* as relevant before suite-local exclusions and add a table-driven routing test for every pipeline.

VincyZhang: Good catch! Fixed in https://github.com/LMCache/LMCache/pull/4177/commits/474c423364f81a3da09d088df0f0960f2517dd10

maobaolong:
@andyluo7 Would you like to take another look?

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +53,214 @@ _path_filter_is_trivial() {
...
+
+_path_filter_pipeline_kind() {
+    case "$1" in
+        *k3_tests/unit/pipeline.yml) echo unit ;;
+        *k3_tests/integration/pipeline.yml) echo integration ;;
+        *k3_tests/correctness/pipeline.yml) echo correctness ;;
+        *k3_tests/multiprocess/pipeline.yml) echo multiprocess ;;
+        *k3_tests/blend/pipeline.yml) echo blend ;;
+        *k3_tests/sglang/pipeline.yml) echo sglang ;;
+        *k3_tests/xpu/pipeline.yml) echo xpu ;;
```
andyluo7: **[P2] Match the actual XPU pipeline paths**

The XPU uploaders pass `xpu/unittests/pipeline.yml` and `xpu/multiprocess/pipeline.yml`, so both are classified as `generic` and never load `xpu/filter.sh`. Please recognize these nested paths and load the parent device filter.

.buildkite/k3_tests/common_scripts/path-filter.sh
```diff
@@ -56,6 +53,214 @@ _path_filter_is_trivial() {
...
+                    if [[ "$changed_file" == lmcache/v1/platform/*/* ]]; then
+                        return 0
+                    fi
+                    return 1
+                    ;;
+                examples/*)
+                    # Example-only changes should not trigger unit tests.
+                    return 0
+                    ;;
+                .buildkite/k3_tests/unit/*|tests/*)
```
andyluo7, line 99: **[P2] Keep other suites from triggering unit**

The unit branch has no exclusion for other `.buildkite/k3_tests/*` directories, so an integration-only change falls through as relevant and still runs unit. Please skip other suite-local paths here; the earlier `common_scripts/*` check preserves shared helpers.

maobaolong:
@VincyZhang Would you like to do a new round update?

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: Fixed. XPU pipeline detection now matches the actual nested paths. The device filter is loaded from the device suite root.

.buildkite/k3_tests/common_scripts/path-filter.sh
VincyZhang: Fixed. Unit now skips suite-local changes under .buildkite/k3_tests/* unless they are under unit/. Shared common_scripts/ remains classified as relevant before the suite-local rule, so common upload/filter changes still trigger unit.

VincyZhang:
@andyluo7 Could you please help to review again?

maobaolong (commented):
LGTM from my side. My previous concerns around the safer fallback behavior and device path routing have been addressed.

maobaolong (approved):

andyluo7 (approved):
Re-reviewed 2cc19f4f: the nested XPU routing and unit-suite exclusions are fixed. The remaining Blend failure is an external clone-auth failure. LGTM.

PR #4178, opened 2026-07-21
[Perf][FS] Batch filesystem L2 I/O operations

2xdevv:

The filesystem L2 adapter previously awaited each key’s disk
operation serially. Batched execution should schedule independent
operations concurrently while preserving positional results and
per-key failure isolation.

- Runs filesystem store, lookup, and load operations concurrently
  while retaining direct single-key paths.
- Preserves positional lookup and load bitmaps.
- Isolates ordinary per-key I/O failures while propagating
  cancellation and other base exceptions.
- Adds tests for single-key and batched round trips, concurrent
  execution, duplicate keys, and failure isolation.

**Results (speedup = orig ÷ batched, median latency for the 16-key batch, tested on BM1743)**

***Buffered (default, page cache)***

| payload | store | load | lookup |
|--:|--:|--:|--:|
| 128K | 2.1× | 2.1× | 1.9× |
| 512K | 3.1× | 3.3× | 2.4× |
| 1M | 3.6× | 4.0× | 1.9× |
| 4M | 6.6× | 7.1× | 2.0× |
| 16M | 9.8× | 9.7× | 1.8× |
| 64M | 12.2× | 11.3× | 3.0× |

***O_DIRECT (page cache bypassed, real disk I/O)***

| payload | store | load | lookup |
|--:|--:|--:|--:|
| 128K | 2.3× | 2.4× | 1.9× |
| 1M | 3.9× | 3.7× | 2.6× |
| 16M | 3.9× | 3.8× | 2.2× |
| 64M | 2.6× | 5.2× | 2.8× |

**Duplicate key handling**:

- Store writes the first object for each unique key once.
- Load reads each unique key once and copies the data into all
  matching destinations.
- Each successfully populated load position is represented in the
  positional bitmap.

**Store behavior change**:

When a store batch contains duplicate keys and the first write fails,
the adapter no longer attempts a later duplicate as an implicit retry.
The original serial implementation could process the later duplicate
after the failure. The batched implementation writes each unique key
only once. This tradeoff is reasonable because filesystem write
failures are expected to be uncommon.

- The tests exercise the standard async filesystem path with
  O_DIRECT disabled.

Tested by:

```bash
pytest -q \
  tests/v1/distributed/test_fs_l2_adapter_batched_io.py
```

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py
@@ -608,6 +608,67 @@ class FSL2Adapter(L2AdapterInterface):
 
     # ---- store ----------------------------------------------------------
 
+    async def _store_one(
+        self,
+        key: ObjectKey,
+        obj: MemoryObj,
+    ) -> tuple[bool, int]:
+        """Store a single key's bytes atomically.
+
+        Writes to a per-key temp path, then ``rename``s onto the final
+        path. Returns ``(success, bytes_written)`` where
+        ``bytes_written == 0`` on either a skipped (already-present) or
+        failed write; ``success`` is False only on I/O errors — a
+        skipped key is not a failure.
+        """
+        file_path, tmp_path = self._key_to_file_and_tmp_path(key)
+
+        if await aiofiles.os.path.exists(file_path):
+            return True, 0
+
+        buf = obj.byte_array
+        size = len(buf)
+
+        try:
+            do_odirect = self._use_odirect
+            if do_odirect:
+                aligned = self._os_disk_bs > 0 and size % self._os_disk_bs == 0
+                if not aligned:
+                    logger.warning(
+                        "Cannot use O_DIRECT for writing size %d, "
+                        "not aligned to block size %d.",
+                        size,
+                        self._os_disk_bs,
+                    )
+                    do_odirect = False
+
+            if do_odirect:
+                await self._loop.run_in_executor(
+                    None,
+                    self._write_with_odirect,
+                    tmp_path,
+                    buf,
+                )
+            else:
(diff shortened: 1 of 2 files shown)


lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py
```diff
@@ -581,65 +653,57 @@ async def _execute_store(
...
+            grouped = {}
             success = False
 
+        # Fan out one write per unique key. Keeping the first object for
+        # duplicates preserves the previous serial behavior, where the first
+        # occurrence created the final file and later occurrences skipped it.
+        unique_pairs = [
+            (key, occurrences[0][1]) for key, occurrences in grouped.items()
+        ]
+        if len(unique_pairs) == 1:
```
maobaolong, line 699: It looks like there is special handling for the case with only a single element. Does this actually lead to a significant performance boost, or is it more of a maintenance overhead?

If there is no real benefit, wouldn't introducing this extra branch negatively impact code read ability?

2xdevv: Hello, @maobaolong, thank you for the review!

I tested batch size 1 with 128 KiB and 2 MiB payloads using both buffered I/O and `O_DIRECT`. Compared with the singleton fast path, `asyncio.gather()` degraded lookup by 6.9–16.5%, load by 3.7–7.8%, and store by 4.6–7.5%.

I believe this is because `asyncio.gather()` adds task creation, event-loop scheduling, and result-collection overhead, while a single operation has no concurrency benefit to offset that cost.

Given these results, do you think this performance degradation is tolerable compared with the maintenance overhead of keeping the singleton fast paths?

PR #4181, opened 2026-07-21
[BugFix] Attach daemon shared memory via read-only mmap (SageMaker HP)

### LINKED ISSUE #4234: [Bug] Resource tracker deletes daemon-owned shared memory on client exit (SageMaker HyperPod L2 adapter)
**Label**

bug

**Describe the bug**

The SageMaker HyperPod L2 adapter's client attaches to the ai-toolkit daemon's shared-memory segment with `multiprocessing.shared_memory.SharedMemory(create=False)`. Resource tracker registers the segment **even for attach-only handles** and calls `shm_unlink()` at interpreter exit.

Because the segment is owned by the ai-toolkit daemon (not by LMCache), this cleanup deletes the daemon's node-local cache out from under the daemon and every other client on the node. Any LMCache pod restart (rollout, eviction, crash) becomes a **node-wide cache wipe** whenever the pod can see the host's `/dev/shm` (`hostIPC: true`), and the replacement pod crash-loops with "segment not found" until the daemon is manually restarted.

**To Reproduce**

On a SageMaker HyperPod EKS cluster the failure chain is:
1. Deploy LMCache (MP mode) with the `sagemaker-hyperpod` L2 adapter in a pod with `hostIPC: true`
2. Delete/replace the LMCache pod
3. `/dev/shm/ai_toolkit_cache` disappears from the **host**
4. The replacement pod crash-loops with "ai-toolkit shared-memory segment not found"; every other client on the node loses L2

Reproduced 3/3 times on a HyperPod EKS cluster (g6e.xlarge, separate LMCache/vLLM pods).

**Expected behavior**

An attach-only client must never delete a segment it does not own.

**Screenshots**

N/A — log evidence instead. Replacement pod after a client pod deletion:

```
RuntimeError: ai-toolkit shared-memory segment 'ai_toolkit_cache' not found
```

 - N/A

 - N/A

**Environment:**
 - OS: Ubuntu 22.04 (containers), Amazon EKS on SageMaker HyperPod
 - Python: 3.12
 - LMCache: `dev` (also affects 0.5.2, the first release with the sagemaker-hyperpod MP adapter)
 - Deployment: MP mode, separate LMCache/vLLM pods, `hostIPC: true`

**Additional context**

- A `subPath` bind mount of the segment masks the deletion (unlink hits a mountpoint → `EBUSY`) but only in topologies without `hostIPC`, and it pins the mount to a stale inode across daemon restarts — so it is not a fix.

bluayer:

Fixes https://github.com/LMCache/LMCache/issues/4234

The SageMaker HyperPod L2 client attaches to the ai-toolkit daemon's shared-memory segment with `multiprocessing.shared_memory.SharedMemory`. Shared Memory lib's resource tracker registers the segment even for attach-only handles and calls `shm_unlink()` at interpreter exit — the daemon-owned segment is deleted out from under the daemon and every other client on the node.

In k8s, this turns any LMCache pod restart into a node-wide cache wipe when the pod can see the host's `/dev/shm`. Reproduced failure chain on a SageMaker HyperPod EKS cluster:

1. LMCache pod is deleted/replaced (rollout, eviction, crash)
2. resource tracker unlinks `/dev/shm/ai_toolkit_cache` on the host
3. the replacement pod (and any other client) crash-loops with "segment not found"

This PR replaces the attach path with `_ReadOnlySharedMemoryMapping`, a small helper that opens `/dev/shm/<name>` and maps it with `mmap(fd, 0, PROT_READ)`

- Why not `SharedMemory(track=False)`: it requires Python 3.13+, and it still maps the segment writable. `mmap(PROT_READ)` covers both properties on every supported version.
- Verified end-to-end on a SageMaker HyperPod EKS cluster (separate LMCache / vLLM pods, `hostIPC: true`), A/B test:
  - before: client pod deletion destroyed the host arena every time
    (3/3 reproductions, replacement pod crash-loops)
  - after: arena inode unchanged across repeated client pod deletions and full teardown; replacement pods reattach immediately
  - L1/L2 serving paths unaffected (L1-evict → L2 load verified, identical outputs, no errors)

**If applicable**:


ApostaC (approved):
LGTM!

KuntaiDu (approved):
LGTM!

PR #4183, opened 2026-07-22
feat(mp): multimodal support for MP mode (no encoder cache yet)

royyhuang:

MP mode derives cache keys from raw token ids. vLLM emits identical placeholder token ids for every image, so two requests with the same text but different images collide on identical keys — lookup can serve the wrong image's KV (silent cross-image contamination). In-process mode already prevents this by overwriting placeholder spans with `mm_hash`-derived values (`apply_mm_hashes_to_token_ids`); the MP connector never adopted that scheme.

This PR ports it to the MP connector: the request tracker captures mm features at admission and materializes an mm-adjusted copy of the prompt token ids once; `get_token_ids_for_keys()` then feeds every key-derivation site (lookup, lock release, store, retrieve) and L0 telemetry, so all sites stay mutually consistent.

Validated end-to-end (Qwen2.5-VL-3B + MP server): same prompt with two different images now stores distinct chunks and answers correctly for each image; before the fix the second image reused the first image's keys.

Scope note: this covers KV-cache keying only — encoder-cache (EC) support for MP mode is a separate, larger effort and intentionally not included.

- Materialize-once vs in-process's apply-per-call: MP has four key sites invoked repeatedly per scheduler step; one materialization keeps them consistent by construction and avoids the repeated list→tensor→list roundtrips the in-process adapter carries `TODO: Optimize` markers for.
- The 16-bit mm-hash projection (`hex_hash_to_int16`) is inherited unchanged so both modes derive identical key token values.
- `lmcache_mp_connector_0180/_0201` variants (older vLLM pins) intentionally untouched; the same diff backports if desired.

**If applicable**:


lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -302,6 +315,15 @@ def num_allocated_blocks(self) -> dict[int, int]:
             for engine_group_idx, blocks in self.allocated_block_ids.items()
         }
 
+    def get_token_ids_for_keys(self) -> list[int]:
```
ApostaC: nit: maybe rename it to `get_token_ids`?

ApostaC (commented):
Left one nit comment. 

Also, it would be great if we can have some user-facing docs for a multi-modal model.

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #4188, opened 2026-07-22
ci: Update e2e validation script and CPU config to support deepseek model

maobaolong:

Currently, vLLM cpu device does not support MLA for now, we can run the model with MLA disabled first to verify the CPU device for LMCache.

**If applicable**:


.github/scripts/run-cpu-e2e-validation.sh
```diff
@@ -255,7 +302,23 @@ print(json.dumps({
...
+  # Optional flags — appended only when set, so unrelated callers stay
+  # bit-identical to the pre-refactor invocation.
+  local -a extra_serve_args=()
+  if [ -n "${VLLM_LOAD_FORMAT}" ]; then
+    extra_serve_args+=(--load-format "${VLLM_LOAD_FORMAT}")
+  fi
+  if [ -n "${VLLM_HF_OVERRIDES}" ]; then
+    extra_serve_args+=(--hf-overrides "${VLLM_HF_OVERRIDES}")
+  fi
+  # trust-remote-code is required for models with custom python modelling
```
DongDongJu, line 314: trust-remote-code is okay for this CI/CD?

DongDongJu (commented):
LGTM, I left one question.

.github/scripts/run-cpu-e2e-validation.sh
maobaolong: yeah should be fine — we use dummy weights here, no real model code is loaded. tarball is from our own release too, not pulled from HF directly. only enabled for deepseek, opt-125m doesn't use it.

hlin99 (approved):
LGTM. Checked with @maobaolong offline. 
This is the 1st step for mla model enabling on cpu. 
Now, this PR runs mla model in non mla mode

maobaolong:
@hlin99 Yep, thanks for the review!

DongDongJu (approved):
LGTM

PR #4193, opened 2026-07-22
[MP][S3 L2 adapter] support self signed certs and path style access

ziruiliu:

Adds S3 L2 adapter support for private S3-compatible object storage deployments that require:
  - Path-style S3 addressing (/<bucket>/<object-key>)
  - A custom/self-signed CA certificate bundle

This supports small on-premises object-storage clusters such as MinIO, Ceph RGW, and similar deployments where virtual-hosted DNS and publicly trusted TLS certificates are not available.

**What is changed**
  - Adds s3_addressing_style, with virtual as the backward-compatible default and path for path-style requests.
  - Adds required s3_bucket when path-style addressing is selected.
  - Adds optional s3_ca_bundle, passed to AWS CRT’s TLS trust store.
  - Updates object operations and ListObjectsV2 paths for path-style addressing.
  - Documents the new options and adds unit coverage for configuration validation and request paths.

  Example:
  {
    "type": "s3",
    "s3_endpoint": "s3.internal:9000",
    "s3_region": "us-east-1",
    "s3_addressing_style": "path",
    "s3_bucket": "lmcache",
    "s3_ca_bundle": "/etc/ssl/private-s3-ca.pem"
  }

This patch also addressed problem in #3593, which enables OBJ backend in NIXL.
NIXL library has no delete interface and has limitation in batched operations, therefore we'd better update s3 l2 adpater to support self-hosted object storage.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
@@ -1018,7 +1018,7 @@ class S3L2Adapter(L2AdapterInterface):
         # urlencode handles percent-escaping of values (continuation
         # tokens are typically base64 with ``+``/``/``/``=`` chars).
         path_prefix = (
-            "/" if self._bucket is None else f"/{url_quote(self._bucket, safe='')}/"
+            "/" if self._bucket is None else f"/{url_quote(self._bucket, safe='')}"
         )
         path = path_prefix + "?" + urlencode(params, quote_via=url_quote)
 
--- a/lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
@@ -188,9 +188,21 @@ def _object_key_to_string(key: ObjectKey) -> str:
     return base
 
 
-def _format_safe_path(key_str: str) -> str:
-    """URL-encode the object name to form a safe HTTP path."""
-    return "/" + url_quote(key_str)
+def _format_safe_path(key_str: str, bucket: Optional[str] = None) -> str:
+    """URL-encode an object name to form a virtual- or path-style HTTP path.
+
+    Args:
+        key_str: S3 object name.
+        bucket: Bucket name for path-style addressing. When omitted, produce
+            the virtual-hosted-style path.
+
+    Returns:
+        The request path for the object.
+    """
+    object_path = url_quote(key_str)
+    if bucket is None:
+        return "/" + object_path
+    return "/" + url_quote(bucket, safe="") + "/" + object_path
 
 
 def _make_credentials_provider(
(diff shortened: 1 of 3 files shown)


maobaolong:
@ziruiliu CQ issue occurred, would you like to fix it before review?

tests/v1/distributed/test_s3_l2_adapter.py
```diff
@@ -747,6 +754,48 @@ def test_help_nonempty(self):
         assert "s3_endpoint" in S3L2AdapterConfig.help()
 
 
+    def test_path_style_requires_bucket(self):
```
maobaolong, line 756: You may need to remove an extra blank line before this.

tests/v1/distributed/test_s3_l2_adapter.py
```diff
@@ -746,6 +753,47 @@ def test_help_nonempty(self):
...
+            S3L2AdapterConfig.from_dict(
+                {
+                    "s3_endpoint": "s3://bucket",
+                    "s3_region": "us-east-1",
+                    "s3_addressing_style": "invalid",
+                }
+            )
+
+
+def test_path_style_uses_service_host_bucket_prefix_and_bucket_root():
```
maobaolong: Why this test case isn't in any class?

lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
```diff
@@ -973,7 +1017,10 @@ def _list_request(
             params.append(("continuation-token", continuation_token))
         # urlencode handles percent-escaping of values (continuation
         # tokens are typically base64 with ``+``/``/``/``=`` chars).
-        path = "/?" + urlencode(params, quote_via=url_quote)
+        path_prefix = (
+            "/" if self._bucket is None else f"/{url_quote(self._bucket, safe='')}/"
+        )
+        path = path_prefix + "?" + urlencode(params, quote_via=url_quote)
```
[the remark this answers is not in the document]
maobaolong: @ziruiliu How about this issue?

tests/v1/distributed/test_s3_l2_adapter.py
ziruiliu: no special reason. I left it there while adding the path-style coverage, but grouping it under a dedicated TestPathStyleAddressing class is clearer. I’ll move it.

lmcache/v1/distributed/l2_adapters/s3_l2_adapter.py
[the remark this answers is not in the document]
ziruiliu: codex is right.
The path-style ListObjectsV2 resource should be /<bucket>?list-type=2, without the trailing slash.
My current s3 storage tolerates the extra slash so my end2end test successfully passed.
I will correct this API format

maobaolong (approved):
@ziruiliu Thanks for the update. Everything looks good so far.

Issue #4186, opened 2026-07-22
CPU-only lmcache server + CUDA vLLM worker: REGISTER_KV_CACHE fails with 'Device UUID not found' (examples/multi_process + docstring mismatch)


zhaotyer:
## Summary

When the `lmcache server` runs in a CPU-only container (no GPU visible → `StubCPUDevice`) and the vLLM worker has CUDA GPUs, the worker's `REGISTER_KV_CACHE` request fails at server side with:

```
RuntimeError: Device UUID f231cc14-0922-0ca9-5c57-efd8982e6c5e not found in the discovered devices. Please make sure the process can see all the accelerator devices
```

The `examples/multi_process/` k8s YAMLs describe exactly this topology (CPU-only DaemonSet + CUDA vLLM Deployment) but never set `lmcache.mp.mp_transfer_mode=engine_driven` on the client, so they appear to hit the same failure. Two related doc bugs make it hard to diagnose: the connector docstring describes `engine_driven` / `lmcache_driven` exactly opposite to what the factory code does.

### Server (CPU-only container)

```bash
lmcache server \
    --host 0.0.0.0 --port 8000 \
    --l1-size-gb 20 --eviction-policy LRU --chunk-size 16
```

Startup log shows the server fell back to the CPU stub:

```
LMCache INFO:  torch_dev=StubCPUDevice(device_type=cpu), torch_device_type=cpu
...
LMCache INFO: Supported transfer mode: auto
LMCache INFO: LMCache ZMQ cache server is running on tcp://0.0.0.0:8000
```

### Client (vLLM with GPU, default `mp_transfer_mode=auto`)

```
vllm serve <model> --kv-transfer-config '{"kv_connector":"LMCacheMPConnector", "kv_role":"kv_both", "kv_connector_extra_config": {"lmcache.mp.host": "tcp://<server>", "lmcache.mp.port": 8000}}'
```

### Error (server-side)

```
LMCache ERROR: Error handling request RequestType.REGISTER_KV_CACHE
...
  File ".../lmcache/v1/platform/cache_context.py", line 149, in _detect_device_type
    device_types = {w.to_tensor().device.type for w in kv_caches}
  File ".../lmcache/v1/multiprocess/custom_types.py", line 145, in to_tensor
    device_index = self._get_device_index_from_uuid(self.device_uuid)
  File ".../lmcache/v1/multiprocess/custom_types.py", line 79, in _get_device_index_from_uuid
    raise RuntimeError(
[...]

PR #4194, opened 2026-07-22
docs(mp-connector): correct inverted mp_transfer_mode descriptions (#4186)

Anai-Guo:
## What

Fixes the `lmcache.mp.mp_transfer_mode` docstring on `LMCacheMPConnector`, which described the two transfer modes **backwards** relative to the factory code in `lmcache/v1/multiprocess/transfer_context/worker_transfer.py`.

Reported in #4186 (section "Doc bugs found while debugging", item 2).

## The inversion

Current docstring (`lmcache/integration/vllm/lmcache_mp_connector_0201.py`):

```
auto (default; CUDA -> engine_driven, others -> lmcache_driven),
engine_driven (force IPC / SHM zero-copy),
lmcache_driven (force worker-side gather/scatter copy)
```

But the factory does the opposite:

```python
# worker_transfer.py
if resolved_mode is MPTransferMode.LMCACHE_DRIVEN:
    return _build_lmcache_driven_context(device_type)   # -> LMCacheDrivenTransferContext
if resolved_mode is MPTransferMode.ENGINE_DRIVEN:
    return _build_engine_driven_context()               # -> EngineDrivenTransferContext
# AUTO: dispatch by device type
if device_type == "cuda":
    return LMCacheDrivenTransferContext()
return _build_engine_driven_context()
```

And the two context classes' own docstrings confirm the semantics:

- `LMCacheDrivenTransferContext` — *"the serving engine provides device handles (IPC for CUDA, SHM wrappers for CPU) and the LMCache server performs direct device-side data transfer"* → **IPC / SHM zero-copy**.
- `EngineDrivenTransferContext` — *"the engine (worker side) owns the data movement: the worker adapter gathers/packs KV into CPU buffers"* → **worker-side gather/scatter copy**.

So the correct mapping is:

- `lmcache_driven` = IPC / SHM zero-copy
- `engine_driven` = worker-side gather/scatter copy
- `auto`: CUDA → `lmcache_driven`, others → `engine_driven`

This matters because #4186's documented escape hatch for a CPU-only lmcache server + CUDA worker is "force `engine_driven`". With the docstring inverted, a user reading it would pick the wrong mode.

## Change

Docstring-only; no behavior change.

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector_0201.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector_0201.py
@@ -451,9 +451,9 @@ class LMCacheMPConnector(KVConnectorBase_V1):
     - lmcache.mp.heartbeat_interval: interval (seconds) between server
       heartbeat pings.
     - lmcache.mp.mp_transfer_mode: routing mode for the worker -> server
-      transfer context. One of ``auto`` (default; CUDA -> engine_driven,
-      others -> lmcache_driven), ``engine_driven`` (force IPC / SHM
-      zero-copy), ``lmcache_driven`` (force worker-side gather/scatter
+      transfer context. One of ``auto`` (default; CUDA -> lmcache_driven,
+      others -> engine_driven), ``lmcache_driven`` (force IPC / SHM
+      zero-copy), ``engine_driven`` (force worker-side gather/scatter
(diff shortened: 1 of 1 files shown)


hlin99 (approved):
LGTM. thx for the fix！

maobaolong (approved):
@Anai-Guo Good catch! Thanks for the correction.

PR #4201, opened 2026-07-22
refactor(mp): decouple worker transfer context from vllm integration

maobaolong:

`LMCacheDrivenTransferContext` was importing `wrap_kv_caches` from `lmcache.integration.vllm.vllm_multi_process_adapter` and passing a hard-coded `EngineType.VLLM` to the server on register. That's a backwards dependency -- the transport layer under `lmcache.v1.multiprocess` shouldn't reverse-depend on a specific engine integration, and it made the "core" path look vLLM-only when it really isn't.

The wrap helpers themselves are engine-neutral (they just dispatch on `tensor.device.type` via `resolve_kv_wrapper_factory`), so this PR moves them to a new device-agnostic module `lmcache/v1/platform/kv_wrap.py`. The vLLM adapter now passes its own `EngineType.VLLM` through `TransferContext.register(..., engine_type=...)`, and the worker transport is free of any vLLM import.

- No behavior change on the wire -- same `REGISTER_KV_CACHE` payload with the same `EngineType.VLLM` value, just now supplied by the adapter instead of hard-coded in the transport.
- `TransferContext.register` gains an `engine_type` kwarg with `EngineType.VLLM` as the default, so any out-of-tree caller keeps working.
- The old adapter-hosted `wrap_kv_caches` / `wrap_one_kv_cache` names are gone; a couple of tests that were `monkeypatch`ing them were updated to target `kv_wrap` (their real home) and `worker_transfer` (where the bound reference lives).

**If applicable**:


ApostaC (approved):
LGTM!

zhengfeihe (approved):

hlin99 (approved):
LGTM

PR #4202, opened 2026-07-22
fix(mp): improve device IPC error when the receiver lacks accelerator access

Smallfu666:

In MP mode, when a process that receives an accelerator IPC handle has no accelerator visibility (for example an `lmcache server` started without GPU access), `DeviceIPCWrapper._get_device_index_from_uuid` failed at `REGISTER_KV_CACHE` with a generic "Device UUID ... not found in the discovered devices" error. That points operators at device visibility, when the real issue is that this process has no accelerator at all. This is the failure reported in #4186.

This PR makes the error actionable while keeping it engine- and role-agnostic: `DeviceIPCWrapper` is platform-base code, below the engine integration layer, so the message names no specific engine, process role, or config key. It points at the two supported topologies: give the receiving process visibility of the same accelerator devices, or configure the MP deployment to use `engine_driven` transfer so the sending client performs the device-side copy. Device-capable hosts are unaffected -- the discovery/lookup path is unchanged, and an available-device/unknown-UUID lookup still raises the original error.

- Scoped to the runtime diagnostic. The `examples/multi_process/` topology is intentionally out of this PR: per #2128 the example was originally a GPU-visible server using CUDA IPC, so switching it to `engine_driven` would be choosing a new topology rather than a fix. Better handled separately once the intended deployment is settled.
- Addresses the inline review: the message no longer assumes vLLM or a server role, and no longer implies a client-only change is always sufficient.

**Testing**:

```
pytest tests/v1/platform/test_ipc_wrapper.py -q
3 passed
```

Covers: a device-incapable host raises the actionable engine-agnostic error; a device-capable host with an unknown UUID still raises the original error; a known UUID resolves to its ordinal.

**If applicable**:

Addresses #4186

### DIFF
--- a/lmcache/v1/platform/base/ipc_wrapper.py
+++ b/lmcache/v1/platform/base/ipc_wrapper.py
@@ -88,7 +88,41 @@ class DeviceIPCWrapper:
 
     @classmethod
     def _get_device_index_from_uuid(cls, device_uuid: str) -> int:
-        """Get the physical device ordinal from its UUID."""
+        """Get the physical device ordinal from its UUID.
+
+        Args:
+            device_uuid: the UUID of the device to resolve, as carried in an
+                inter-process device handle received from a peer process.
+
+        Returns:
+            The physical device ordinal for ``device_uuid`` on this host.
+
+        Raises:
+            RuntimeError: if this process has no accelerator devices available
+                -- e.g. a CPU-only ``lmcache server`` receiving a CUDA worker's
+                IPC handle -- in which case the message points at the
+                ``engine_driven`` transfer mode; or if a device is available
+                but ``device_uuid`` is not among the discovered devices.
+        """
+        # A device-incapable process (e.g. a CPU-only lmcache server) can never
+        # resolve a peer's device IPC handle. Fail with actionable guidance
+        # instead of the generic "not found" below, which misleads operators
+        # into checking device visibility when the real fix is the transfer
+        # mode. See issue #4186.
+        if not torch_dev.is_available():
+            raise RuntimeError(
+                f"Received an inter-process handle for device {device_uuid}, "
+                "but this process has no accelerator devices available "
+                "(running on a CPU-only host). This usually means a CUDA "
+                "worker registered its KV caches with a CPU-only LMCache "
+                "server using the default 'auto' (lmcache_driven) transfer "
+                "mode, which requires the server to open device IPC handles. "
+                "Set 'lmcache.mp.mp_transfer_mode=engine_driven' in the vLLM "
+                "kv_connector_extra_config (or export "
+                "LMCACHE_MP_TRANSFER_MODE=engine_driven) so the worker performs "
+                "the device-side copy instead."
+            )
+
         cls._discover_devices()
 
         with DeviceIPCWrapper._device_mapping_lock:
--- a/lmcache/v1/platform/base/ipc_wrapper.py
+++ b/lmcache/v1/platform/base/ipc_wrapper.py
@@ -99,28 +99,29 @@ class DeviceIPCWrapper:
 
         Raises:
             RuntimeError: if this process has no accelerator devices available
-                -- e.g. a CPU-only ``lmcache server`` receiving a CUDA worker's
-                IPC handle -- in which case the message points at the
-                ``engine_driven`` transfer mode; or if a device is available
-                but ``device_uuid`` is not among the discovered devices.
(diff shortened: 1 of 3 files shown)


maobaolong:
@Smallfu666 Thanks for this fix. I'll take a look today.

maobaolong:
I'm just curious, what were the original use cases for this example when it was first proposed? At that time, it didn't support CPU, and there wasn't any `lmcache-driven` and `engine-driven`,  what was the initial scenario? @Smallfu666 @ningziwen

lmcache/v1/platform/base/ipc_wrapper.py
```diff
@@ -88,7 +88,41 @@ def _discover_devices(cls) -> None:
...
+        # mode. See issue #4186.
+        if not torch_dev.is_available():
+            raise RuntimeError(
+                f"Received an inter-process handle for device {device_uuid}, "
+                "but this process has no accelerator devices available "
+                "(running on a CPU-only host). This usually means a CUDA "
+                "worker registered its KV caches with a CPU-only LMCache "
+                "server using the default 'auto' (lmcache_driven) transfer "
+                "mode, which requires the server to open device IPC handles. "
+                "Set 'lmcache.mp.mp_transfer_mode=engine_driven' in the vLLM "
```
maobaolong: The error message incorrectly assumes that the engine is vLLM. However, by design, Except for the engine integration layer, lmcache other layers are engine agnostic.

Smallfu666: Good point -- this is below the engine integration layer, so the message shouldn't assume vLLM (or even that the receiver is a server). I've made it engine- and role-agnostic: it now points at the two supported topologies -- give the receiving process visibility of the same accelerator devices, or configure the MP deployment to use `engine_driven` so the sending client performs the device-side copy -- without naming any engine or config key.

Smallfu666:
I checked the original #2128 proposal. Its intended topology was a GPU-visible LMCache server using CUDA IPC -- the description explicitly said the server needed GPU access and that both containers should see the four GPUs, and that it was "not just CPU-only" despite the name.

The final DaemonSet, though, doesn't request `nvidia.com/gpu`; it relied on the runtime exposing the devices anyway. On a standard device-plugin setup that leaves the example ambiguous -- the server can instead come up without accelerator visibility, which is the case reported in #4186.

So changing the example to `engine_driven` would be selecting a newer CPU-only-server topology rather than restoring the original behavior. I've dropped the example change from this PR and kept it focused on the engine-agnostic runtime diagnostic. We can revisit the Kubernetes example separately once its intended topology is decided -- either GPU-visible server + CUDA IPC, or an explicitly CPU-only + `engine_driven` deployment (the latter would also want the deployment docs, entrypoint, health probes, and node placement updated).

@ningziwen, please correct me if I'm missing anything from the original environment.

maobaolong:
> I checked the original #2128 proposal. Its intended topology was a GPU-visible LMCache server using CUDA IPC -- the description explicitly said the server needed GPU access and that both containers should see the four GPUs, and that it was "not just CPU-only" despite the name.
> 
> The final DaemonSet, though, doesn't request `nvidia.com/gpu`; it relied on the runtime exposing the devices anyway. On a standard device-plugin setup that leaves the example ambiguous -- the server can instead come up without accelerator visibility, which is the case reported in #4186.
> 
> So changing the example to `engine_driven` would be selecting a newer CPU-only-server topology rather than restoring the original behavior. I've dropped the example change from this PR and kept it focused on the engine-agnostic runtime diagnostic. We can revisit the Kubernetes example separately once its intended topology is decided -- either GPU-visible server + CUDA IPC, or an explicitly CPU-only + `engine_driven` deployment (the latter would also want the deployment docs, entrypoint, health probes, and node placement updated).
> 
> @ningziwen, please correct me if I'm missing anything from the original environment.

Thanks for the explanation. I agree that the original example shouldn't be modified. I see the scenario you're addressing now.

The PR LGTM. Thanks for the fix!

maobaolong (approved):

Smallfu666:
Still relevant. I rechecked the current `dev` branch, and `_get_device_index_from_uuid()` still falls through to the generic device-UUID error when no accelerator is available, so this diagnostic fix has not been superseded.

The previous review thread is resolved and the PR already has an approval. Happy to rebase onto the latest `dev` if fresh CI is preferred; otherwise the change is still mergeable as-is.

Issue #4127, opened 2026-07-16
[RFC] L2 encryption as a serde extension


aeon-x:
Hi all — a proposal for encrypting KV cache **at rest in L2** (S3 / `fs` /RESP / …). In multi-tenant deployments, tenants' KV cache is stored in a shared L2 backend, so it needs encryption to protect it from anyone with access to that remote storage.

## Scope

Protects KV cache **at rest in L2** against anyone who can read the remote storage (bucket / disk / RESP). So a party with node/process access on a running server is not defended against — that's a different threat model.

## Encryption methods

The serde runs on **both** legs of the round trip — `serialize` on store (L1→L2) and `deserialize` on load (L2→L1) — so the transform must be **reversible**: That rules out one-way functions (hash/digest); this is symmetric encryption, decryptable with the key.

Default: **AES-128-GCM** (AEAD — confidentiality + integrity). Random 96-bit IV per chunk, wire frame `[1B version][12B IV][ciphertext‖16B tag]`; a tag failure (tamper / wrong key) → load miss, never silent corruption. 

AES-128-GCM is the fastest *authenticated* option on AES-NI (raw AES-CTR is faster but drops integrity). AES-128 is already computationally unbreakable,  so we default to 128 and leave key size a config knob (256 for compliance mandates). Compression isn't an alternative (doesn't encrypt), and KV is high-entropy — use quantization for size, *before* encryption.

Throughput/core on server AES-NI:

| Method | GB/s/core | Auth? | Notes |
| --- | --- | --- | --- |
| **AES-128-GCM** | ~4–8 | ✅ | chosen — already unbreakable, fewer rounds, HW-accelerated |
| AES-256-GCM | ~2–5 | ✅ | ~30% slower; opt in for compliance mandates |
| ChaCha20-Poly1305 | ~1–2 | ✅ | wins only *without* AES-NI (ARM) |
| AES-CTR (raw) | ~5–9 | ❌ | no integrity — rejected |

## Encryption Keys: `cache_salt` is the selector, not the key

`cache_salt` picks *which* key; the secret material comes from a swappable `KeyProvider` so the trust model is a config choice, not baked in:
- **`HkdfKeyProvider`** (default) — `key = HKDF(master_key, info=cache_salt)`, one mounted master secret. Per-tenant ciphertext, but any master holder can derive every tenant's key → trust model is "fleet vs. outside."
[...]

PR #4203, opened 2026-07-22
[Serde] add ObjectKey to parameter

aeon-x:
## What

Threads an optional `ObjectKey` through the distributed serde interface so a future encryption serde can key ciphertext per `cache_salt`. Every existing serde is content-agnostic and ignores the key.

Part of  #4127

## Changes

- **`base.py` / `multi.py`**: add `key` / `keys` (optional, default `None`) to `serialize`/`deserialize` and `submit_serialize`/`submit_deserialize`.
- **`async_processor.py`**: forward `keys` into `_run_task`; normalize a missing (`None`) `keys` to a per-object list of `None` so the strict-zip transform loops stay positionally aligned; pass `keys[i]` per object.
- **`serde_wrapper.py`**: store path passes the aligned key list; load path builds a selected-key list in the same `bitmap.test(i)` loop that selects the loaded `src`/`dst` objects, keeping keys aligned with the objects that actually loaded.
- **Concrete serdes** (`fp8`, `turboquant`, `asym_k16_v8`) and the single-to-multi adapters accept the key; the adapters forward it to the wrapped base serde. All are content-agnostic and ignore it.
- **`turboquant`**: rename local K/V cache tensors `key`/`value` → `k_tensor`/`v_tensor` (matches the file's existing `k_out`/`v_out` convention) to avoid shadowing the new `key` parameter.
- Update the serde design docs (`README.md`, `turboquant.md`, `serde_wrapper.md`) to document the optional key.

## Testing

- `pytest tests/v1/distributed/serde/` → 64 passed, 26 skipped (skips are CUDA-gated: fp8/turboquant/e2e).


lmcache/v1/distributed/serde/turboquant/turboquant.py
```diff
@@ -836,15 +844,15 @@ def deserialize(self, src: MemoryObj, dst: MemoryObj) -> None:
                     k_layer = torch.matmul(
                         k_layer.to(torch.float32), pi_t.T.contiguous()
                     ).to(k_out.dtype)
-                key = k_layer.contiguous().view(num_tokens, hidden_dim)
-                value = (
+                k_tensor = k_layer.contiguous().view(num_tokens, hidden_dim)
+                v_tensor = (
```
KuntaiDu: Seems to be unnecessary code diff. Maybe consider remove it?

lmcache/v1/distributed/l2_adapters/serde_wrapper.py
```diff
@@ -199,7 +199,9 @@ def submit_store_task(
         try:
             with self._lock:
                 self._store_tasks[wrapped_id] = state
-                serde_task_id = self._serde.submit_serialize(objects, temp_objs)
+                serde_task_id = self._serde.submit_serialize(
+                    objects, temp_objs, state.keys
```
KuntaiDu, line 203: Seems that `key` won't be None. In this case I would prefer the signature of 3rd parameter to be 
ObjectKey
instead of
ObjectKey | None

lmcache/v1/distributed/serde/turboquant/turboquant.py
aeon-x: removed

lmcache/v1/distributed/l2_adapters/serde_wrapper.py
aeon-x: updated

ApostaC (approved):
Did a very quick pass. LGTM!

KuntaiDu (approved):
LGTM!

PR #4206, opened 2026-07-22
[Fix] MiniMax M3 and other Mamba Model for incoming vLLM 0.26

### LINKED ISSUE #4154: Native transfer kernels (execute_object_group_transfer / multi_layer_block_kv_transfer) throw CUDA invalid argument on MiniMax-M3 multi-format KV topology (118 tensors, B200); pure-torch fallback works
Title: `lmc_ops.execute_object_group_transfer` / `multi_layer_block_kv_transfer` fail with CUDA "invalid argument" on every store for MiniMax-M3 (118 KV tensors, 3 kernel groups, mixed formats); pure-torch fallback works

## Environment
- lmcache 0.5.1 (pip wheel, cu13x build shipped in wheel; `lmcache.c_ops` backend selected, `multi_layer_block_kv_transfer mode: ptr`)
- vLLM 0.23.1rc1.dev672 (also reproduced on stock `vllm/vllm-openai` MiniMax-M3 image — MXFP8; and on an NVFP4 fork — arch-independent of quant)
- GPU: 8x B200 (sm_100), TP4; driver 580.x; CUDA_LAUNCH_BLOCKING=1 confirms the failing call
- Server: external `lmcache server` (MP mode) + LMCacheMPConnector, chunk_size 256, L1 100GB pinned, L2 fs_native

## Model topology (from server logs)
118 registered KV tensors per rank -> 3 kernel groups in 1 object group:
- group 0: 60 layers, GQA KV, shape (13524, 2, 128, 1, 128) bf16, tokens_per_block=128, slots_per_block=1 ("compressed")
- group 1: 57 layers, MSA indexer K-only, shape (13524, 128, 128) bf16, kv=1, slots_per_block=128, block_stride_elems=16384
- group 2: 1 layer (EAGLE3 drafter), shape (13524, 2, 16, 128, 128) bf16

## Symptom
Every store fails:
```
File ".../lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py", line 402, in _run_object_group_transfer_plan
    lmc_ops.execute_object_group_transfer(
torch.AcceleratorError: CUDA error: invalid argument
```
Forcing the non-planned path (`_HAS_NATIVE_OBJECT_GROUP_TRANSFER=False`) moves the same failure to
`lmc_ops.multi_layer_block_kv_transfer` (line ~556 callsite). Reproduced under CUDA_LAUNCH_BLOCKING=1 —
it is the kernel launch itself, not an async report. sm_100 cubins ARE present in the wheel (other models work).

## Workaround (proves the plumbing is fine)
2-line patch routing the callsite to the wheel's own pure-torch
`lmcache.python_ops_fallback.multi_layer_block_kv_transfer` (signature-identical): stores, L1 hits,
and fs_native L2 disk hits then all work end-to-end on M3 TP4 with EAGLE3 (all 118 tensors, 4 per-group
formats). So the group/format bookkeeping is correct — only the native kernels reject this topology.

Suspect: the unusual group-1 layout (kv=1, nh=1, bs=128, block_stride_elems=16384 — K-only indexer cache
with slots_per_block=128) and/or group-0 slots_per_block=1 "compressed" layout produce launch parameters
(grid/block dims or shared-mem size) outside kernel limits.

## Repro
- MiniMax-M3 (stock image works to repro): serve with LMCacheMPConnector + external server, send any
  >256-token prompt, watch server log on store.
- Configs attached: [boot_fork.sh, boot_lmc_server.sh]

ApostaC:

Fixes #4154 

This PR adds support for incoming vLLM 0.26 unified KV cache, with the following changes:
- Introduce a new KV cache edit module for unified Mamba cache
- Fixed logical errors during KV cache detection when the KV cache list contains multiple different formats
- Fixed the wrong head-size detection issue (#4154) caused by incorrect permutations during KV cache registration.

**If applicable**:

### DIFF
--- a/lmcache/v1/gpu_connector/kv_format/contiguity.py
+++ b/lmcache/v1/gpu_connector/kv_format/contiguity.py
@@ -17,6 +17,29 @@ from lmcache.v1.gpu_connector.kv_format.types import DiscoverableKVCache
 logger = init_logger(__name__)
 
 
+def _get_expected_shape(stride: tuple[int, ...], num_elems: int) -> tuple[int, ...]:
+    """Return the expected shape for a stride-sorted tensor view.
+
+    The returned shape is the tightest possible (contiguous) layout
+    that preserves the original storage size and stride ordering. For
+    example, a tensor with ``stride=(16, 4, 1)`` has an expected shape
+    of ``(4, 4, 4)``, which is the smallest shape that can be laid out
+    in memory with those strides.
+
+    Args:
+        stride: The stride of the tensor.
+        num_elems: The total number of elements in the tensor.
+
+    Returns:
+        The expected shape for a contiguous layout.
+    """
+    assert len(stride) > 0, "Stride must have at least one dimension"
+    expected_shape = [num_elems // stride[0]]
+    for i in range(0, len(stride) - 1):
+        expected_shape.append(stride[i] // stride[i + 1])
+    return tuple(dim for dim in expected_shape)
+
+
 def attempt_permute_to_contiguous_view(
     kv_caches: DiscoverableKVCache,
 ) -> DiscoverableKVCache:
@@ -50,13 +73,11 @@ def attempt_permute_to_contiguous_view(
     preserved.
     """
     if isinstance(kv_caches, torch.Tensor):
-        if kv_caches.is_contiguous():
-            return kv_caches
         strides = kv_caches.stride()
         perm = sorted(range(kv_caches.ndim), key=lambda i: strides[i], reverse=True)
         result = kv_caches.permute(perm)
         if result.is_contiguous():
-            return result
+            return result.view(_get_expected_shape(result.stride(), result.numel()))
         padding_per_block = _validate_dim0_padded_layout(result)
         logger.debug(
             "attempt_permute_to_contiguous_view: accepting dim-0-padded "
--- a/lmcache/integration/vllm/kv_cache_group_edits.py
+++ b/lmcache/integration/vllm/kv_cache_group_edits.py
@@ -4,8 +4,8 @@
 This is needed to mask out attention-specific details while making sure that
 LMCache can still store / load KV cache correctly.
 
-Currently there are two edits, one per :class:`KVCacheGroupEdit` subclass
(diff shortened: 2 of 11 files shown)


sammshen (approved):

lmcache/v1/gpu_connector/kv_format/contiguity.py
```diff
@@ -17,6 +17,29 @@
 logger = init_logger(__name__)
 
 
+def _get_expected_shape(stride: tuple[int, ...], num_elems: int) -> tuple[int, ...]:
```
KuntaiDu, line 20: why do we need this >.<

KuntaiDu (approved):
lgtm

2imi9:
We hit the same crash on AMD ROCm (MI308X / gfx942), not only B200: serving MiniMax-M3 (MXFP8) on vLLM 0.26 through `LMCacheMPConnector` fails at engine init with

```
RuntimeError: shape '[3554, 128, 128, 2, 128]' is invalid for input of size 58228736
```

Same root cause as #4154: M3's heterogeneous topology (60 rank-4 main GQA caches `(NB, NH, BS, 2*HS)` + 57 rank-3 lightning-indexer key caches `(NB, BS, HS)`) goes through whole-list format detection, which picks the rank-4 fused branch from layer 0 and reshapes every tensor with `fused_dim=2*HS` — the rank-3 indexer cache (numel `3554*128*128` = 58,228,736) then can't take the `[..., 2, HS]` split.

Cherry-picking #4206 onto the `v0.5.2` tag applies cleanly and fixes it: both formats now detect per-layer (`NL_X_NB_BS_NH_TWO_HS` main + `NL_X_NB_BS_HS` indexer), M3 serves with coherent output, and evict→retrieve is decode-equivalent to vLLM's own HBM prefix-cache path in our testing. The indexer format retrieves natively on gfx942 (no `python_ops` fallback).

Since `v0.5.2` is the latest release and #4206 merged the day after it was tagged, `v0.5.2` crashes on M3 + vLLM 0.26 out of the box. Is a `v0.5.3` point release carrying #4206 on the roadmap? Happy to help validate on ROCm. Thanks for the quick turnaround on the fix.

PR #4208, opened 2026-07-23
[Telemetry] report L2-related telemetry

KuntaiDu:

Collect L2-related telemetry.

**If applicable**:


lmcache/usage_telemetry/flush.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+import threading
+
+# First Party
+from lmcache.logging import init_logger
+
+logger = init_logger(__name__)
+
+
+class UsageFlushThread:
```
maobaolong: This class can simplifies implementation by leveraging the capabilities of PeriodicThread after merge https://github.com/LMCache/LMCache/pull/4214. By using PeriodicThread, it also gains improved observability, for example `lmcache/v1/internal_api_server/common/periodic_thread_api.py`.

<details>
<summary> patch after merge #4214 </summary>

diff --git a/lmcache/usage_telemetry/flush.py b/lmcache/usage_telemetry/flush.py
index 26c12ece025..0000000000 100644
--- a/lmcache/usage_telemetry/flush.py
+++ b/lmcache/usage_telemetry/flush.py
@@ -1,69 +1,68 @@
 """Flush-thread scaffolding shared by the interval usage reporters."""
 
 # Future
 from __future__ import annotations
 
 # Standard
 from typing import Callable
 import os
-import threading
 
 # First Party
 from lmcache.logging import init_logger
+from lmcache.v1.periodic_thread import (
+    PeriodicThread,
+    ThreadLevel,
+    ThreadRunSummary,
+)
 
-class UsageFlushThread:
+def _read_flush_interval() -> float:
+    """Read LMCACHE_USAGE_TRACK_INTERVAL, clamp to >= 1 s.
+
+    A malformed value falls back to the default instead of disabling
+    telemetry. Clamped to >= 1 s so the flush loop cannot busy-spin.
+    """
+    try:
+        interval = float(os.getenv("LMCACHE_USAGE_TRACK_INTERVAL", "600"))
+    except ValueError:
+        logger.debug("Invalid LMCACHE_USAGE_TRACK_INTERVAL", exc_info=True)
+        interval = 600.0
+    return max(interval, 1.0)
+
+
+class UsageFlushThread(PeriodicThread):
     """Daemon thread that runs a flush callback once per report interval.
 
     Calls *flush* every ``LMCACHE_USAGE_TRACK_INTERVAL`` seconds
     (default 600, clamped to >= 1 s; a malformed value falls back to
     the default instead of disabling telemetry). The thread starts
     immediately on construction.
+
+    Built on :class:`PeriodicThread`: ``wake()`` and ``stop()`` (which
+    also short-circuits the interval sleep) are inherited unchanged, so
+    callers keep the same API. A raising *flush* no longer kills the
+    thread — :class:`PeriodicThread` logs it and records a failed run,
+    which lines up with the telemetry package's swallow-not-throw rule.
     """
 
     def __init__(self, flush: Callable[[], None]) -> None:
         """Start the flush thread.
 
         Args:
             flush: Called on the flush thread once per interval. Must
-                not raise — an exception ends the thread (the reporters'
-                flush methods are guarded and never raise).
+                not raise in practice (the reporters' flush methods are
+                guarded); an exception is logged and counted as a failed
+                run without stopping the thread.
         """
+        super().__init__(
+            name="lmcache-usage-report",
+            interval=_read_flush_interval(),
+            level=ThreadLevel.LOW,
+        )
         self._flush = flush
-        # Clamp to >= 1 s: Event.wait(0) would turn the flush loop into
-        # a busy spin.
-        try:
-            flush_interval = float(os.getenv("LMCACHE_USAGE_TRACK_INTERVAL", "600"))
-        except ValueError:
-            logger.debug("Invalid LMCACHE_USAGE_TRACK_INTERVAL", exc_info=True)
-            flush_interval = 600.0
-        self._flush_interval: float = max(flush_interval, 1.0)
-        self._stop_event = threading.Event()
-        self._wake = threading.Event()
-        self._thread = threading.Thread(
-            target=self._run, daemon=True, name="lmcache-usage-report"
-        )
-        self._thread.start()
-
-    def wake(self) -> None:
-        """Trigger one flush now instead of at the next interval tick."""
-        self._wake.set()
-
-    def stop(self) -> None:
-        """Stop the thread; it runs no further flushes. Idempotent.
-
-        Does not flush — callers send their own final flush on shutdown.
-        """
-        self._stop_event.set()
-        self._wake.set()
+        self.start()
 
-    def _run(self) -> None:
-        while True:
-            self._wake.wait(timeout=self._flush_interval)
-            self._wake.clear()
-            if self._stop_event.is_set():
-                return
-            self._flush()
+    def _execute(self) -> ThreadRunSummary:
+        self._flush()
+        return ThreadRunSummary(success=True)


<details>

KuntaiDu: Thanks for letting me know! I will wait till #4214 being merged

lmcache/usage_telemetry/flush.py
```diff
@@ -0,0 +1,72 @@
...
+    """Start a ``LOW``-level :class:`PeriodicThread` that runs *flush*.
+
+    The thread first runs after one full flush interval and is
+    registered in the global ``PeriodicThreadRegistry``, so *name* must
+    be unique per process.
+
+    Args:
+        name: Thread name, also the registry key.
+        flush: Called on the flush thread every
+            ``LMCACHE_USAGE_TRACK_INTERVAL`` seconds. Must not raise;
```
maobaolong: flush: Called on the flush thread every LMCACHE_USAGE_TRACK_INTERVAL
    seconds. Any exception is caught by PeriodicThread and logged;
    the thread keeps running.

lmcache/usage_telemetry/mp_continuous.py
```diff
@@ -206,10 +185,15 @@ def _on_event(event: Event) -> None:
...
                     buffer.append(sample)
-                    if len(buffer) >= self._max_buffered_samples:
+                    # Only the sample that fills the buffer to exactly
+                    # max_buffered_samples sets overflow: one buffer fill
+                    # triggers one early flush.
+                    if len(buffer) == self._max_buffered_samples:
                         overflow = True
             if overflow:
-                self._wake.set()
+                threading.Thread(
```
maobaolong: Why not call self._flush_thread.wake() , but start a new short life thread to execute flush?

maobaolong (commented):
@KuntaiDu I took a look at the logic related to flash, and left a couple of comments.

lmcache/usage_telemetry/mp_continuous.py
KuntaiDu: the update was before the merge of ur pr. I just did a new commit

lmcache/usage_telemetry/flush.py
KuntaiDu: Thanks. Fixed.

sammshen (approved):
LGTM!

ApostaC:
The UT of this PR runs pretty long (I just retriggered). Is it flaky or there is some real issue behind?

maobaolong (approved):
LGTM

KuntaiDu:
> The UT of this PR runs pretty long (I just retriggered). Is it flaky or there is some real issue behind?

Seems to be outdated LMCache. I just merged latest dev and retried.

PR #4210, opened 2026-07-23
fix(pd): fail fast when a prefill exceeds the receiver buffer capacity

### LINKED ISSUE #4209: [Bug] PD receiver hangs for the full allocation timeout, then fails with a misleading "over-subscribed" error, when a prefill needs more chunks than the whole pd_buffer_size (no fail-fast)
## Description

In Prefill/Decode disaggregation, the receiver (decoder) admits an incoming KV
transfer through a reservation-based admission control loop
(`ReservationManager.async_try_admit`). A request is admitted only when the
buffer has enough **free** chunks:

```python
# lmcache/v1/storage_backend/pd_backend_async.py:151-152
available = self._total_chunks - self._total_reserved
if available >= total_chunks:
    ...
```

`available` can never exceed the buffer's **total** capacity (`self._total_chunks`).
So if a single request's declared `total_chunks` is **greater than the whole
buffer**, admission can never succeed. There is no guard for this case, so the
request:

1. blocks in the admission loop until `pd_allocation_timeout_sec` elapses, then
2. `async_try_admit` returns `False` and the receiver raises the generic
   `"Receiver reservation admission timed out or was aborted ... Buffer may be
   over-subscribed"` (`pd_backend_async.py:1311-1316`).

From the client this surfaces as the decoder dropping the connection mid-transfer
(`ServerDisconnectedError` / `Connection reset by peer` /
`Not enough data to satisfy transfer length header`). The error message points
at "over-subscription" (a transient/concurrency condition) when the real cause
is a **structural** one: `pd_buffer_size` is too small to ever hold this prefill.

The init-time `pd_max_prefill_len` check (`pd_backend_async.py:484-500`) is the
only capacity guard, and it is **opt-in** (default `pd_max_prefill_len=0` skips
it) and validates against the configured `pd_max_prefill_len`, not the actual
per-request length. So a deployment that does not set `pd_max_prefill_len`, or a
request longer than it, gets no upfront error — only the hang-then-timeout above.

## Environment

- LMCache: `0.5.1-g979719d7`
- vLLM: `0.24.0`
- Model: Qwen3-32B, `dtype=bfloat16`, `kv_cache_dtype=auto` (bf16 KV)
- Parallelism: `tensor_parallel_size=1`, `pipeline_parallel_size=1`
- GPU: H100 NVL (~94 GiB)
- PD config (receiver): `pd_backend_mode=async`, `pd_buffer_device=cuda`,
  `pd_allocation_timeout_sec=10`

## Reproduction

KV geometry for the model above: `64 layers × 8 KV heads × 128 head_dim × 2 (K+V)
× 2 B (bf16) = 262 144 B/token`; with `chunk_size=256` → **64 MiB per chunk**.

- `pd_buffer_size = 2 GiB` → `2 147 483 648 / 67 108 864 = 32` chunks.
- A 16 384-token prefill needs `16384/256 + 1 = 65` chunks.

Steps:
1. Configure a PD receiver with `pd_buffer_size=2147483648` (2 GiB),
   `pd_allocation_timeout_sec=10`, and **do not** set `pd_max_prefill_len`.
2. Send a request whose prefill is ≥ 16 384 tokens (65 chunks) through the
   prefiller to this decoder.
3. Observe: the transfer stalls ~10 s, the decoder logs the "over-subscribed"
   admission-timeout `RuntimeError`, and the client sees a connection reset.

`65 > 32`, so the request can **never** be admitted — but nothing tells the
operator that; it just times out and drops the connection.

## Root cause

`_async_allocate_and_put` validates `total_chunks == 0` but never validates
`total_chunks > buffer_total_chunks` before entering the admission wait
(`pd_backend_async.py:1302-1310`). The admission loop then waits out
`pd_allocation_timeout_sec` on a request it can never satisfy.

## Suggested fix

Fail fast at admission time when a request's declared `total_chunks` exceeds the
buffer's total capacity, with an actionable message (needs N chunks, buffer holds
M, increase `pd_buffer_size`). This turns a 10 s hang + misleading error + torn
connection into an immediate, self-explanatory error. (PR to follow.)

## Related / follow-up (separate from the fail-fast)

The **concurrency** case is distinct and is not a bug per se: at
`pd_buffer_size=8 GiB` (128 chunks) with 16 384-token prefills (65 chunks each),
the buffer admits only `floor(128/65)=1` concurrent transfer. At `CON=4`, three
requests correctly wait for a reservation, hit `pd_allocation_timeout_sec`, and
are reset — admission control working as designed, but **invisibly**. A small
follow-up would surface the concurrent-admission capacity at init (e.g. log
"buffer admits ~1 concurrent prefill of pd_max_prefill_len tokens") so operators
can size `pd_buffer_size` for their concurrency instead of discovering the wall
at load time. Filing separately to keep the fix PR focused.

ardecode:

Fixes #4209

In Prefill/Decode disaggregation, the receiver admits an incoming KV transfer
through reservation-based admission control that only succeeds when the buffer
has enough *free* chunks (`available >= total_chunks`). `available` can never
exceed the buffer's total capacity, so a request whose declared `total_chunks`
exceeds the whole buffer can never be admitted — yet there was no guard for it.
Such a request blocked for the full `pd_allocation_timeout_sec` and then raised a
misleading "over-subscribed" admission-timeout error, tearing down the transfer
mid-flight (clients saw `ServerDisconnectedError` / `Not enough data to satisfy
transfer length header`).

This adds an explicit fail-fast in `_async_allocate_and_put`, co-located with the
existing `total_chunks == 0` validation: when a request's declared `total_chunks`
exceeds the buffer's total capacity, it is rejected immediately with an
actionable error (needs N chunks, buffer holds M, increase `pd_buffer_size`)
instead of hanging until the allocation timeout.

**If applicable**:


DongDongJu (approved):
Hello @ardecode 
LGTM, Thanks for the fix.

deng451e (approved):
LGTM!

hlin99 (approved):
LGTM. Thanks for the fix

PR #4211, opened 2026-07-23
[MP][Bugfix] Split store batches into uniform-size sub-batches for native connectors

Writtic:

`NativeConnectorL2Adapter.submit_store_task` sends a whole store batch in one `submit_batch_set` call. The pybind wrapper takes the first buffer's size as `batch_chunk_num_bytes` ("all must match" in `connector_pybind_utils.h`), and the C++ connectors throw `buffer size mismatch` on any other size, failing the whole batch. It fails silently too: `do_batch_set` has no per-key logging (unlike `do_batch_get`) and the adapter reduces the completion to a bare `ok=False`, so the data is dropped with no log line and no retry.

Batches do arrive mixed. `StoreController._group_keys_by_shape` buckets keys by `(model_name, kv_rank)`, one bucket per `submit_store_task`, on the invariant its docstring states ("Each bucket shares a single `(shape, dtype)`") and asks to extend when a shape-affecting field is added to `ObjectKey`. `object_group_id` is such a field and is missing from the tuple. Under `separate_object_groups=True` (the MP default) each object group gets its own key from `ipc_key_to_object_keys` and its own memory-object size, so a hybrid-attention model puts two sizes in one bucket and the connector rejects all of it.

Observed on vLLM MP, a Gemma model with heterogeneous KV layout, RESP L2 to Redis 7.2.4: 5318 of 5326 store keys failed over a 5000-request benchmark and L2 stayed empty, while a single-key `submit_batch_set` probe to the same Redis from the same pod succeeded.

- `submit_store_task` groups keys by `memoryview.nbytes` and submits one native batch per size, aggregated under a single task id (`_PendingStoreTask`). The `L2TaskId` contract and `pop_completed_store_tasks` are unchanged, and a uniform batch still makes exactly one native call.
- The demux thread logs the native error string on a failed sub-batch instead of swallowing it.
- A native submit that raises partway through shrinks the aggregate to the sub-batches already submitted, so the task completes as failed rather than hanging.
- Adds `lmcache_mp.l2_adapter_pending_entries`, an observable gauge reporting `len()` of each pending map per live adapter, tagged by `l2_name` and `map`. Zero counts are reported so a draining backlog is visible.

1. The `StoreController` grouping tuple needs fixing too. I will send it separately since it is a different module and this PR is already under review. That one keeps buckets uniform at the source, this one makes the adapter hold the native contract on its own.

2. The store-bytes event just below the submit call assumes uniformity as well: `successful_objs[0].get_size() * len(successful_objs)`. I left it alone here; it becomes correct once the tuple is fixed.

3. @maobaolong asked for a way to confirm the pending maps are not leaking. The gauge covers `_pending_ops`, `_pending_store_sizes`, and the new `_pending_store_tasks`. Wiring the `l2_name` tag surfaced that adapter-side type names did not match the registered names other l2-tagged metrics use, so the factories now pass the registered ones (`resp`, `mooncake_store`, `native_plugin`, `fs_native`). The leak paths the gauge exposes are tracked in #4342.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 # Standard
 from collections import defaultdict
 from typing import Any
+import dataclasses
 import select
 import threading
 
@@ -81,6 +82,16 @@ def _obj_to_memoryview(
     return obj.byte_array  # type: ignore[return-value]
 
 
+@dataclasses.dataclass
+class _PendingStoreTask:
+    """Aggregation state for one store task split into uniform-size
+    sub-batches (see ``submit_store_task``)."""
+
+    remaining: int
+    ok: bool = True
+    bytes_stored: int = 0
+
+
 class NativeConnectorL2Adapter(L2AdapterInterface):
     """
     Wraps a pybind-wrapped C++ IStorageConnector to
@@ -129,6 +140,10 @@ class NativeConnectorL2Adapter(L2AdapterInterface):
             int,
             tuple[str, L2TaskId, int, list[ObjectKey] | None],
         ] = {}
+
+        # Store tasks split into uniform-size sub-batches:
+        # task_id → aggregation state across its native futures
+        self._pending_store_tasks: dict[L2TaskId, _PendingStoreTask] = {}
 
         # Completed results (same pattern as MockL2Adapter)
         self._completed_stores: dict[L2TaskId, L2StoreResult] = {}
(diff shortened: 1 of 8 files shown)


lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
```diff
@@ -195,19 +210,61 @@ def submit_store_task(
...
         # non-blocking so holding the lock is brief.
         with self._lock:
             task_id = self._get_next_task_id()
-            future_id = int(self._client.submit_batch_set(key_strings, memviews))
-            self._pending_ops[future_id] = (
-                self._OP_STORE,
-                task_id,
-                len(keys),
-                None,
+            self._pending_store_tasks[task_id] = _PendingStoreTask(
```
maobaolong, line 263: I noticed you're adding an object to this data structure here. When the current thread encounters an exception, it is popped, otherwise, another thread processes it and pops it.

I'm still concerned about potential leaks, though. Is there a mechanism in place to ensure it doesn't leak? Or at the very least, is there a way for us to monitor it to confirm that no leaks are occurring? How about add a lambda based gauge metric?

I noticed that there are similar issues for ` _pending_ops` and `_pending_store_sizes` too, we could open this up for further discussion, it doesn't necessarily have to be resolved within this PR.

Writtic: @maobaolong Thanks for the detailed review. I don't think there is any mechanism that fully prevents the leak.

On the Python side every insert is paired with a removal: the submit failure path cleans up its own entries ([L255-L267](https://github.com/LMCache/LMCache/blob/89b24c303a626774910ae09c2631647854869ad0/lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py#L255-L267)), and the demux thread pops all three maps when a completion arrives ([L494-L548](https://github.com/LMCache/LMCache/blob/89b24c303a626774910ae09c2631647854869ad0/lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py#L494-L548)). So these maps can only grow without bound in two cases: a submitted future_id never gets a completion, or the demux thread that routes completions dies.

The first case seems to be the bigger one, and it lives on the C++ side. For example, if `create_connection()` throws at worker startup, the worker exits without draining its queue ([connector_base.h L585-L589](https://github.com/LMCache/LMCache/blob/89b24c303a626774910ae09c2631647854869ad0/csrc/storage_backends/connector_base.h#L585-L589)). An exception that is not a `std::exception` also kills the worker: it escapes the per-tile catch ([L573](https://github.com/LMCache/LMCache/blob/89b24c303a626774910ae09c2631647854869ad0/csrc/storage_backends/connector_base.h#L573)) and its tile never completes. A future stuck this way also pins the L1 read locks held by the in-flight store, so the cost looks bigger than dict growth.

On adding the lambda-based gauge metric: agreed, and I'll try adding it in a follow-up PR using the existing `register_gauge` helper, following the `lmcache_mp.num_inflight_l2_stores` pattern in StoreController ([L265-L277](https://github.com/LMCache/LMCache/blob/89b24c303a626774910ae09c2631647854869ad0/lmcache/v1/distributed/storage_controllers/store_controller.py#L265-L277)).

I also agree with your point on `_pending_ops` / `_pending_store_sizes`. Looking at the code, two things bother me, and I'd like to follow up with:

1. Wrapping the demux routing and listener callbacks in try/except so a single callback exception can't kill the demux thread.
2. Strengthening the completion contract in `ConnectorBase` so every submitted future_id completes even when a worker dies. Once that holds, the Python maps are provably bounded and the gauge metric becomes a verification tool.

If you'd prefer the lambda-based gauge metric in this PR, I can add it right away.

maobaolong: Thanks for your reply. It would be nice if you could handle the other parts in a later PR. For the current one, it would be helpful to add some lambda gauge metrics, so that we can verify the leak.

Writtic: Added the gauge metric in f5051d2: `lmcache_mp.l2_adapter_pending_entries`, an observable gauge reporting `len()` of each pending map per live adapter, tagged by `l2_name` and `map`. Zero counts are reported so a backlog draining back to zero is visible. While wiring `l2_name` I found the adapter-side type names didn't match the registered names the other l2-tagged metrics use, so the factories now pass the registered names (resp, mooncake_store, native_plugin, fs_native). PTAL.

Writtic:
@maobaolong Could you please take another look when you have a chance?

lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py
Writtic: Opened #4342 for the wider discussion, with the two paths in `worker_loop_for_queue` that can leave a submitted `future_id` without a completion, and the L1 read locks a hung store task holds. The gauge added here is what makes it observable, so this PR stays as it is.

PR #4212, opened 2026-07-23
[MP] make DeviceHostFuncDispatcher a PeriodicThread

maobaolong:

Right now `DeviceHostFuncDispatcher` in `native_completion.py` runs its own hand-rolled daemon thread with a `_stop_flag` / `_wake` event pair. It does exactly what `PeriodicThread` was made for: sleep for `drain_interval_seconds`, drain once, repeat. So this PR just makes it inherit `PeriodicThread`, moves the drain into `_execute()`, and registers it into `PeriodicThreadRegistry` (level=HIGH).

Nice side effects:

- shows up in the periodic-thread HTTP API together with `PinMonitor`, session cleanup, p2p poller and the reaper
- gets `is_active` / total_runs / failed_runs / last_summary tracking for free
- stop path is a bit cleaner: `super().stop()` + unregister + one final `_drain_once()` to flush residual completions (same behavior as before)

Public API (`register` / `start` / `stop` / `dispatched_count` / `handler_exception_counts` and the `drain_interval_seconds=` ctor arg) is unchanged, so existing call sites and tests keep working.

- Renamed the inner `self._lock` to `self._registry_lock` because `PeriodicThread` already has its own `_lock: RLock` for run stats — kept them separate to avoid mypy conflicts and semantic overlap.
- Existing tests in `tests/v1/multiprocess/test_completion_recorder.py` and `tests/v1/multiprocess/test_lmcache_driven_layout_registry.py` still use the class the same way; the fake dispatcher in the layout-registry test already stubs `register/start/stop` so it's unaffected.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/native_completion.py
+++ b/lmcache/v1/multiprocess/native_completion.py
@@ -40,6 +40,12 @@ import torch  # noqa: F401 — must be imported before lmcache.c_ops
 
 # First Party
 from lmcache.logging import init_logger
+from lmcache.v1.periodic_thread import (
+    PeriodicThread,
+    PeriodicThreadRegistry,
+    ThreadLevel,
+    ThreadRunSummary,
+)
 import lmcache.c_ops as _lmc_ops
 
 logger = init_logger(__name__)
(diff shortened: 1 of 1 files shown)


maobaolong:
@elliotz-ai Would you like to take a look at this ? Thanks!

maobaolong:
@ApostaC @KuntaiDu Would you like to take a look at this PR also? Thanks!

KuntaiDu (approved):
lgtm!

ApostaC (approved):
LGTM!

PR #4214, opened 2026-07-23
[Core] add wake() to PeriodicThread

maobaolong:

Add a `wake()` method to `PeriodicThread` so subclasses can trigger a
run immediately instead of waiting for the next interval tick. As a
side effect, `stop()` no longer sits on the interval sleep.

The `UsageFlushThread` proposed in #4208 does exactly this pattern
today with its own `wake`/`stop` events. With this PR, that thread
(and any future reporter with the same shape) can be built directly on
`PeriodicThread` and drop the hand-rolled scaffolding, while still
getting the registry / status / level features for free.

**If applicable**:


KuntaiDu (approved):
LGTM!

maobaolong:
@DongDongJu Would you like to take a double check?Thanks.

ApostaC (approved):
LGTM!

PR #3608, opened 2026-06-09
[Core][MP] Introduce object_group_id into the ObjectKey

ApostaC:
# Compatibility Warning

> [!IMPORTANT]
> **This PR (and the follow-up PRs) changes the file format of the L2 KV cache when using MP mode.**
> **The L2 KV caches stored before these PR will be invalidated and no longer retrievable!**

> [!IMPORTANT]
> **This PR (and the follow-up PRs) changes the file format of the L2 KV cache when using MP mode.**
> **The L2 KV caches stored before these PR will be invalidated and no longer retrievable!**

> [!IMPORTANT]
> **This PR (and the follow-up PRs) changes the file format of the L2 KV cache when using MP mode.**
> **The L2 KV caches stored before these PR will be invalidated and no longer retrievable!**

**What this PR does / why we need it**:

To support efficient loading of hybrid models' KV cache, we need to separate the KV cache into object groups. Therefore, we need an object_group_id as a field in the ObjectKey

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [x] this PR contains unit tests

Issue #3106, opened 2026-04-22
[Feature] Support multi-group KV layouts (hybrid sliding + full attention, e.g. Gemma-4)


rarepepi:
**Labels:** `new feature`, `kv-connector`, `architecture`

**Is your feature request related to a problem? Please describe.**
LMCache is currently unusable on any model that mixes attention types — Gemma-4 being the immediate example, with 50 sliding-window layers and 10 full-attention layers having different `(num_kv_heads, head_size)` shapes. The allocator correctly produces a multi-group buffer, but every storage backend and connector (`gds`, `local_disk`, `maru`, `storage_manager`, all three `kv_connectors`, `cachegen_encoder`) reaches for `MemoryObj.tensor`, which materializes a single-shape view and fails on the mismatched group.

Concretely, on `RedHatAI/gemma-4-31B-it-FP8-block` with the V1 connector:

| Group | Layers | Shape `[chunks, 2, tokens, kv_heads, head_size]` | kv_heads × head_size |
|-------|--------|--------------------------------------------------|---------------------|
| 0 (sliding) | 50 | `[5862, 2, 16, 16, 256]` | 16 × 256 = 4096 |
| 1 (full)    | 10 | `[5862, 2, 16,  4, 512]` |  4 × 512 = 2048 |

Sum matches the allocator's `115,343,360` exactly — allocation is fine; it's every reshape downstream that breaks. Net effect: LMCache can't cache Gemma-4 traffic at all.

**Describe the solution you'd like**
Make `MemoryObj.tensor` group-aware and update the ~8 callers that currently assume a single shape:

1. Replace `MemoryObj.tensor` (single tensor) with either a list-returning API (`.views() -> List[Tensor]`) or keep `.tensor` but have it return a tuple/list when the underlying buffer is multi-group.
2. Update the single-shape callers:
   - `gds`, `local_disk`, `maru` storage backends
   - `storage_manager`
   - All three `kv_connectors` (read + write paths)
   - `cachegen_encoder`
3. Persist/serialize per-group metadata (shape, dtype, layer-group assignment) alongside the blob so offloaded KV can be rehydrated into the right group on load.

This lines up with the direction taken in SGLang (`HiSWARadixCache`, and confirms this is a broader ecosystem gap, not a Gemma-4 quirk.

**Describe alternatives you've considered**
- `--no-enable-prefix-caching` — doesn't help; LMCache intercepts earlier.
[...]

PR #4217, opened 2026-07-23
[Bugfix][MP] Trim sliding-window loads when object-group separation is disabled

d3nb:
# Sliding-window load trimming without object-group separation

## What this PR does

Fix H2D retrieve correctness for mixed-attention models when
`--no-separate-object-groups` puts full-attention and sliding-window KV groups
in one object.

## Why this is needed

With `separate_object_groups=False`, all kernel groups share one object group.
That object group is intentionally reported as full attention, so the existing
object-level sliding-window skip does not run.

For a reused prefix, the engine may already have released old sliding-window
logical blocks. Its block table can therefore contain repeated placeholder
block IDs for those out-of-window chunks. LMCache currently schedules every
cached chunk for every kernel group, causing H2D scatter to overwrite that
shared placeholder block.

The default separated-object-group path already trims the SW object as a
whole. The cache object itself is valid; the missing piece is a per-kernel
fallback for the supported no-separation mode.

## Fix

- Add `KVLayerGroupsManager.get_sw_size_chunks()` to expose each kernel group's
  effective cross-chunk window. It returns `-1` for full attention and
  `full_sw_kv`.
- Only when H2D sees a full-attention object group, calculate the leading
  out-of-window blocks independently for each kernel group. Separated SW
  object groups retain the existing object-level path.
- Combine this boundary with `skip_first_n_tokens` using `max()`. Both are
  prefixes of the same transfer range, so adding them would double-count their
  overlap.
- Apply the same logic to the native object-group planner and Python fallback.

D2H/store remains unchanged, full-attention groups remain unchanged, and
object-group separation remains the preferred/default path.

## Tests

Added regression coverage for both native and fallback transfer paths:

- mixed full-attention + sliding-window H2D trimming;
- D2H/store remains untrimmed;
- `full_sw_kv` remains untrimmed;
- separated SW objects use only the existing object-level trim;
- overlap with `skip_first_n_tokens` is not double-counted.

The regression test fails on the current `dev` base and passes with this patch:

```text
base:    4 failed, 6 passed
patched: 10 passed
```

Also added manager-level coverage for token-to-chunk window conversion and
`full_sw_kv`.

Refs #3106 and #3608.

### DIFF
--- a/lmcache/v1/kv_layer_groups.py
+++ b/lmcache/v1/kv_layer_groups.py
@@ -561,6 +561,25 @@ class KVLayerGroupsManager:
             return self._lmcache_tokens_per_chunk
         return sw_size_tokens
 
+    def get_sw_size_chunks(self, kernel_group_idx: int) -> int:
+        """Return a kernel group's cross-chunk sliding-window size.
+
+        Args:
+            kernel_group_idx: 0-based kernel group index.
+
+        Returns:
+            The sliding-window size measured in LMCache chunks, or ``-1`` when
+            the group should retain full attention across chunks.
+        """
+        if self._full_sw_kv:
+            return -1
+        sw_size_tokens = self._kernel_groups[kernel_group_idx].sw_size_tokens
+        if sw_size_tokens == -1:
+            return -1
+        return (
+            sw_size_tokens + self._lmcache_tokens_per_chunk - 1
+        ) // self._lmcache_tokens_per_chunk
+
     def get_attn_desc(self) -> AttnWindowDesc:
         """Return the cross-chunk attention windows of all object groups.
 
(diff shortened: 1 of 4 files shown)


d3nb:
@ApostaC — since you were just in this area merging #4781, would you (or whoever you'd route it to) be able to take a look at this one? It has had no human review since it was opened on 2026-07-23.

Different bug from the #4781/#47505 guard, but adjacent: this is the H2D retrieve correctness fix for mixed full/sliding-window attention when `separate_object_groups=False`. That flag is currently the only way to run DeepSeek-V4 through the MP connector at all — it is the workaround I posted on #3106, and this PR fixes what the workaround then exposes.

Short version of the failure: with separation off, all kernel groups share one object group, and that group is deliberately reported as full attention, so the object-level sliding-window skip never runs. On a reused prefix the engine may already have released the out-of-window sliding blocks, leaving repeated placeholder block IDs in the block table, and LMCache still schedules loads for those chunks.

State of the PR: 1 commit, `+376/-1` across 4 files (2 of them tests), DCO green, no conflicts. The included tests fail 4 / pass 6 on base and pass 10 with the fix. We also run it in our own DeepSeek-V4 deployment, where it survives long external-hit runs and an AIPerf end-to-end pass.

Happy to rebase, split it, or adjust the approach if you'd prefer it structured differently.

PR #4222, opened 2026-07-24
[Bugfix][Core] Preserve object-group layouts in L2 prefetch

amd-ivaganev:

Hybrid KV models can expose object groups with different tensor shapes and byte sizes. The multiprocess registry and prefetch request previously carried only one MemoryLayoutDesc, so lookup, L1 reservation, and L2 load batches reused the first object-group layout for every key. For GPT-OSS 120B, NIXL then paired objects with incompatible byte sizes and page counts, causing L2 prefetch to fail after a cold fill.

This PR:

- Adds an immutable ObjectGroupLayoutDesc that preserves every object-group layout while retaining uniform-layout compatibility.
- Validates layout cardinality at the public API and registry boundaries.
- Partitions hybrid requests into homogeneous layout shards for lookup, reservation, and load, then restores results to chunk-major order.
- Makes NIXL load completion transactional so failed or partial transfers cannot be reported as cache hits and handles are released on every path.
- Correlates observability by adapter and physical task ID so same-adapter hybrid fanout cannot collide or be undercounted.
- Adds API, registry, lookup, controller, NIXL, trace-codec, metrics, and hybrid-layout regressions, plus user and design documentation.

Validation:

- pre-commit run --all-files passed, including Rust formatting and Clippy with warnings denied.
- 101 CPU API, registry, lookup, trace, and observability tests passed.
- 50 GPU prefetch and L1/L2 metrics tests passed.
- 36 NIXL adapter tests passed; 12 existing eviction tests remain skipped for their documented file-descriptor leak.
- A focused Sphinx build with warnings treated as errors passed.
- A 60,160-token GPT-OSS 120B replay, with L1 and the local vLLM prefix cache disabled, hit all 60,160 external-cache tokens and prefetched all 470 hybrid objects without NIXL errors.

- PrefetchRequestSpec still accepts a singular MemoryLayoutDesc, which the registry expands uniformly for compatibility. LayoutDescRegistry.find remains the group-zero compatibility view; hybrid-aware paths use find_object_group_layouts.
- Please pay particular attention to layout-shard result ordering, transactional NIXL completion, and the distinction between logical adapter counts and physical task counts in metrics.
- The full documentation tree renders but retains 29 pre-existing warnings outside this change; the affected documentation passes focused Sphinx validation with warnings treated as errors.
- The broader suite reached 2,387 passes before an unrelated GDS test selected the default cufile backend on ROCm. Forced hipfile validation passed 43 of 44 backend tests and both cache-engine variants; the remaining failure is an existing stale thread-pool expectation.

**If applicable**:


ApostaC:
@amd-ivaganev Can you directly try passing `--no-separate-object-groups`?

amd-ivaganev:
> @amd-ivaganev Can you directly try passing `--no-separate-object-groups`?

Tried that and it works around the proximate problem, but at the cost of efficiency. Did you want that as an experiment or is this the recommended solution?

PR #4223, opened 2026-07-24
[bench CLI] Fix the KV cache size detection when running model with TP

ApostaC:

Previously, when a model is launched with TP, the `lmcache bench engine` will use single TP-slice's KV cache size to calculate the number of sessions.

This PR fixes this by detecting all the ranks from the same model replica and add the KV cache size together.

**If applicable**:


deng451e (approved):
LGTM!

Issue #4141, opened 2026-07-17
[good-first-issue] core: convert f-string log calls in azure_connector.py to %-format


vedjaw:
## Description

Seven `logger.*` calls in `lmcache/v1/storage_backend/connector/azure_connector.py` use f-string formatting. Per Python's logging docs and ruff G004, `%`-format is preferred so the message is only built when the log actually emits.

## Scope

- line 152 — `logger.debug(f"Blob not found: {key_str}")`
- line 155 — `logger.warning(f"Azure HEAD/properties error for {key_str}: {e}")`
- lines 233–237 — `logger.error(f"Size mismatch for {key_str}: ...")`
- line 249 — `logger.error(f"Failed to download {key_str} from Azure: {e}")`
- lines 276–280 — `logger.error(f"Cannot upload {key_str}: chunk size ...")`
- line 294 — `logger.debug(f"Uploaded {key_str} to Azure successfully")`
- line 296 — `logger.error(f"Failed to upload {key_str} to Azure: {e}")`

All convert to `%s` / `%d` as appropriate. No behavior change; format style only.

## Fix pattern

```python
# Before
logger.debug(f"Blob not found: {key_str}")
# After
logger.debug("Blob not found: %s", key_str)
```

Refs #3372

vedjaw:
PR ready: will link shortly — converts the 7 G004 f-string log calls in `azure_connector.py` to `%-format`.

vedjaw:
Opened https://github.com/LMCache/LMCache/pull/4142

Issue #3974, opened 2026-07-01
[good-first-issue] core: convert f-string log calls to %-format in storage_backend (storage_manager / local_cpu_backend / remote_backend)


waynel96:
## Description

Several `logger.*(f"...")` calls in the storage backend use f-string formatting. Per Python's logging docs and ruff G004, `%`-format is preferred so the message is only built when the log actually emits.

## Scope

Following @maobaolong's direction in #3372, this targets 3 files in `lmcache/v1/storage_backend/`:

- `storage_manager.py` — 10 calls
- `local_cpu_backend.py` — 16 calls
- `remote_backend.py` — 17 calls

All convert to `%s`/`%d`/`%.2f` (numeric fields keep their prior precision, e.g. `:.2f` → `%.2f`). No behavior change; format style only.

## Fix pattern

```python
# Before
logger.error(f"Error closing backend {name}: {e}")
# After
logger.error("Error closing backend %s: %s", name, e)
```

Ref #3372

PR #4233, opened 2026-07-24
[good-first-issue] health_monitor: convert f-string log calls in base.py to %-format

Anai-Guo:
## Summary

Follow-on to the ongoing `%`-format logging migration in this repo (#4141, #3974, #3952, #3794, #3434 …). `lmcache/v1/health_monitor/base.py` was not covered by any of those issues but still builds every log record eagerly with f-strings.

This converts all 13 f-string `logger.*` calls in that file to lazy `%`-style arguments.

## Why

With f-strings the message is interpolated before the call, so the work happens even when the level is filtered out. That matters most in `_run_health_checks()` and `_recover_from_local_cpu_fallback()`, which run on every health-monitor tick, and in the `check.name()` call sites where the argument is a method call rather than a plain attribute.

## Changes

- 13 `logger.debug/info/warning/error` calls in `lmcache/v1/health_monitor/base.py` moved from f-strings to `%`-format with deferred arguments.
- Rendered messages are byte-identical to before; multi-line implicit concatenations were re-wrapped so the format string stays one literal and the arguments follow.
- No behavioural change, no API change, no new imports.

## Verification

Run with the ruff version pinned in `.pre-commit-config.yaml` (`v0.11.7`):

```
uvx ruff@0.11.7 check lmcache/v1/health_monitor/base.py          # All checks passed!
uvx ruff@0.11.7 check --select G lmcache/v1/health_monitor/base.py  # All checks passed!
uvx ruff@0.11.7 format --check lmcache/v1/health_monitor/base.py    # already formatted
```

No test asserts on any of the touched log strings.

---

### DIFF
--- a/lmcache/v1/health_monitor/base.py
+++ b/lmcache/v1/health_monitor/base.py
@@ -243,15 +243,16 @@ class HealthMonitor(PeriodicThread):
             try:
                 instances = cls.create(self._manager)
             except Exception as e:
-                logger.warning(f"Failed to create health check {cls.__name__}: {e}")
+                logger.warning("Failed to create health check %s: %s", cls.__name__, e)
                 continue
             for instance in instances:
                 self._health_checks.append(instance)
                 # Initialize previous status as healthy
                 self._previous_check_status[instance.name()] = True
                 logger.info(
-                    f"Registered health check: {instance.name()} "
-                    f"with fallback_policy: {instance.fallback_policy}"
+                    "Registered health check: %s with fallback_policy: %s",
+                    instance.name(),
+                    instance.fallback_policy,
                 )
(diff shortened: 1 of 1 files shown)


maobaolong (approved):
@Anai-Guo LGTM. Thank you for your contributions. Moving forward, I'd like you to shift your focus to the LMCache Multiprocess Mode, as the maintainability of the current In-Process Mode will decrease in the future.

ApostaC (approved):
LGTM!

maobaolong:
@Anai-Guo Close this PR ref to https://github.com/LMCache/LMCache/pull/4290#issuecomment-5112191703

PR #4237, opened 2026-07-25
feat: Add report-host override for frontend heartbeat addr conf

maobaolong:

When running the lmcache frontend dashboard locally, the MP server registers
itself with the discovery service but the dashboard cannot reach the
node it discovered. Root cause: the heartbeat plugin builds its
`api_address` from `get_local_ip()`, which picks a non-loopback NIC
address (or a VPN address on macOS). That address is often not
routable back from the dashboard side, so `/lmcache_infos` ends up
listing an unreachable `apiAddress`.

This PR gives the operator an explicit escape hatch: a new
`plugin.frontend.report-host` config key (and matching `--report-host`
CLI flag) that, when set, is used verbatim as the host part of the
reported address, skipping auto-detection entirely.

**If applicable**:

### DIFF
--- a/docs/source/mp/frontend_dashboard.rst
+++ b/docs/source/mp/frontend_dashboard.rst
@@ -75,23 +75,66 @@ Install the extra dependencies used by the frontend and discovery service:
 
 These are not pulled in by the base ``lmcache`` install to keep it slim.
 
-Quick Start
------------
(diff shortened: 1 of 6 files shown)


yoo-kumaneko (approved):
minor changes. LGTM

ApostaC (approved):
LGTM!

PR #4238, opened 2026-07-25
[CI] pin vllm-cpu-nightly on macOS to dodge broken upstream wheel

maobaolong:

`CPU Device Tests` has been red on every PR since 2026-07-24 ~08:34 UTC on the macOS legs. The last upstream `vllm-cpu-nightly` upload (`0.25.2.dev202607240834`) declares `torch==2.13.0` in its wheel metadata but the shipped `vllm/_C.abi3.so` was compiled against an older libtorch. On macOS the loader trips over:

```
Symbol not found: __ZN3c104impl3cow23materialize_cow_storageERNS_11StorageImplE
  Referenced from: .../site-packages/vllm/_C.abi3.so
  Expected in:     .../torch/lib/libc10.dylib
```

and then `cpu_worker.py` blows up at `torch.ops._C.init_cpu_memory_env(...)` with `AttributeError: '_OpNamespace' '_C' object has no attribute 'init_cpu_memory_env'`. Linux is fine because our install adds `--extra-index-url https://download.pytorch.org/whl/cpu`, which still serves `torch-2.11.0+cpu` — that matches the wheel ABI.

This pins the macOS leg to the last known-good nightly (`dev202607230832`) via a new `VLLM_CPU_NIGHTLY_SPEC` env consumed by `install_vllm_cpu.sh`. Linux keeps rolling nightly, so we don't freeze CI coverage of upstream vLLM CPU changes on the platform that actually works.

Affected runs: e.g. https://github.com/LMCache/LMCache/actions/runs/30140482282

**Special notes for your reviewers**

Temporary pin — please revert once upstream ships a macOS-clean nightly.

**If applicable**

### DIFF
--- a/.github/scripts/install_vllm_cpu.sh
+++ b/.github/scripts/install_vllm_cpu.sh
@@ -18,6 +18,11 @@
 #   PIP_BIN="uv pip" install_vllm_cpu.sh
 #                                 # use `uv pip` and pass extra flags
 #                                 # via PIP_INSTALL_EXTRA_ARGS
+#   VLLM_CPU_NIGHTLY_SPEC="vllm-cpu-nightly==0.25.2.dev202607230832" \
+#     install_vllm_cpu.sh         # pin a specific nightly build (used
+#                                 # by the macOS CI leg to dodge a
+#                                 # broken upstream wheel; see the
+#                                 # workflow for context)
 #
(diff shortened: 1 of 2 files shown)


hlin99 (approved):
lgtm

yoo-kumaneko (approved):
hot fix. LGTM

PR #4242, opened 2026-07-25
feat(musa): add MP transfer platform adapters

superleo:

Defines the process-local pointer contract for MUSA multiprocess transfers.

Cross-process tensors are imported through Torch-MUSA's public IPC API before the transfer context is created. The existing packed paged-buffer pointers, staging-buffer pointers, and raw stream-pointer interfaces remain unchanged.

MUSA-specific pointer validation, optional non-owning Tensor views, stream ordering, and resource lifetime handling stay inside the MUSA adapter.

Adds MUSA pointer-to-Tensor reconstruction, MP cache-context lifecycle management, and stream-ordered completion/event publication.

MUSA block transfer remains disabled. This PR does not change generic multiprocess logic, CUDA behavior, or the wire format.

**If applicable**:

### DIFF
--- a/docs/design/v1/platform/musa/mp_transfer_pointer_contract.md
+++ b/docs/design/v1/platform/musa/mp_transfer_pointer_contract.md
@@ -1,0 +1,54 @@
+# MUSA Multiprocess Transfer Pointer Contract
+
+## Goal
+
+Keep the existing multiprocess pointer APIs unchanged. MUSA tensors are opened
+through TorchMUSA IPC before transfer setup, and MUSA-specific adaptation stays
+inside the MUSA platform implementation.
+
+## Process boundary
+
+Pointers are process-local. The worker exports tensors with
+`torch.musa.ipc.export_tensor()`, and the server opens them with
+`torch.musa.ipc.open_tensor()`. Only pointers obtained from the server's opened
+tensors are passed to transfer operations; raw pointers are never serialized.
+
+The server keeps each IPC owner alive until the transfer stream is synchronized
+and the cache context closes.
+
+## Platform contract
+
+The generic path keeps the existing calls:
+
+```python
+paged_ptrs = context.get_kernel_group_kv_pointers(group_idx)
+staging_ptrs = [context.get_temp_kernel_group_buffer(i, group_idx).data_ptr()]
+lmc_ops.multi_layer_block_kv_transfer(paged_ptrs, staging_ptrs, ...)
+```
+
+`MUSACacheContext` returns the same packed `int64` pointer tensor shape used by
+the current pointer path. `construct_musa_tensor_from_data_pointer()` rebuilds
+a non-owning Tensor view from a process-local pointer and explicit shape,
+stride, dtype, device, and storage-offset metadata.
+
+The supported layouts are:
+
+- `NL x [2, NB, BS, NH, HS]`
+- `NL x [NB, BS, HS]`
+
+The helper supports explicit strides so padded block layouts can be represented
+without copying. The allocation owner must outlive every reconstructed view.
+
+## Stream ordering
+
+Generic completion and event recorders continue to pass an integer stream
+pointer. `MusaDeviceOps` wraps that pointer with TorchMUSA's external-stream
+API, synchronizes it, and then publishes to the existing Python recorder queue.
+The wrapper does not own or destroy the underlying stream.
+
+## Compatibility
+
+No pointer-versus-Tensor branch is added to generic multiprocess code. CUDA
+native signatures, callbacks, transfer planning, and wire format are unchanged.
+The MUSA block-transfer capability remains disabled until the transfer path
+consumes this contract.
--- a/lmcache/v1/platform/musa/cache_context.py
+++ b/lmcache/v1/platform/musa/cache_context.py
@@ -1,0 +1,483 @@
+# SPDX-License-Identifier: Apache-2.0
+"""MUSA cache context for LMCache-driven multiprocess transfer."""
+
+# Future
+from __future__ import annotations
+
+# Standard
+from collections.abc import Sequence
+from typing import TYPE_CHECKING, Any, cast
+
+# Third Party
+import torch
+
+# First Party
+from lmcache import torch_dev
+from lmcache.logging import init_logger
+from lmcache.utils import EngineType
+from lmcache.v1.gpu_connector.kv_format.types import DiscoverableKVCache
+from lmcache.v1.gpu_connector.utils import (
+    LayoutHints,
+    get_device,
+    get_group_data_ptrs,
+    normalize_and_discover_per_layer_formats,
+)
+from lmcache.v1.kv_layer_groups import KVLayerGroupsManager
+from lmcache.v1.multiprocess.custom_types import KVCache
+from lmcache.v1.multiprocess.group_view import engine_group_layer_indices
+from lmcache.v1.platform.base.cache_context import BaseCacheContext
+from lmcache.v1.platform.ops_types import EngineKVFormat
+
+if TYPE_CHECKING:
+    # First Party
+    from lmcache.v1.multiprocess.group_view import EngineGroupInfo
+
+logger = init_logger(__name__)
+
+
+def unwrap_kv_cache_tensors(kv_caches: KVCache) -> list[torch.Tensor]:
+    """Return tensors reconstructed from MUSA IPC wrappers.
+
+    Args:
+        kv_caches: MUSA IPC wrappers received over the multiprocess wire.
+
+    Returns:
+        The reconstructed MUSA KV-cache tensors.
+    """
+    return [wrapper.to_tensor() for wrapper in kv_caches]
+
+
+class _MUSAHostCallbackStream:
+    """Small adapter for stream-ordered callback call sites.
+
+    CUDA contexts expose a CuPy stream as ``cupy_stream``. MUSA does not use
+    CuPy here, so this adapter provides ordered callback and synchronization
+    operations over a TorchMUSA stream and exposes its pointer through the
+    existing platform stream contract.
+    """
+
+    def __init__(self, stream: object) -> None:
+        self._stream = stream
+
+    def synchronize(self) -> None:
+        """Wait for all work submitted to the wrapped MUSA stream."""
+        synchronize = getattr(self._stream, "synchronize", None)
+        if not callable(synchronize):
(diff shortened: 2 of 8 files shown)


DongDongJu (approved):
It do not hurting the cuda behavior. LGTM

lmcache/v1/platform/base/cache_context.py
ApostaC, line 1: Sorry, but I did not see the point of introducing these two new functions.

IIUC, they are just directly calling the original functions, right? And the meaning of `transfer_data` in the function names is not super clear to me.

lmcache/v1/platform/base/device_ops.py
ApostaC, line 1: What will be wrong if we directly change the `record_completion_on_stream` and `record_event_on_stream` signatures?

Also, I'm not sure why we introduced these two functions either.

lmcache/v1/mp_observability/event_bus.py
```diff
@@ -164,32 +165,54 @@ def has_subscribers(self, event_type: EventType) -> bool:
...
+                    int_metadata,
+                )
+            else:
+                _lmc_ops.record_event_on_stream(
+                    stream.ptr,
+                    event.event_type.value,
+                    event.session_id,
+                    str_metadata,
+                    int_metadata,
+                )
```
ApostaC: If I understand correctly, we already automatically resolve the device for `lmc_ops` when importing it (it's a PR by @zxue2 ). In this case, we don't need to call `resolve_device_ops` again, and we don't even need to pass in the device type to this function.

ApostaC (commented):
Thanks @superleo . I just want to have more clarity before we add more complexity into the code.

lmcache/v1/platform/base/cache_context.py
hlin99: I have the same concern. can't we utilize existing ops(use base class or overwrite it) to resolve the problem?
new ops per device usually means more maintain overhead...

lmcache/v1/mp_observability/event_bus.py
superleo: thanks @ApostaC @zxue2  I’ll remove them and use the already resolved lmc_ops

lmcache/v1/platform/base/cache_context.py
superleo: Yes @ApostaC  the defaults preserve the existing cached pointer path. In Stage 4.3（PR 3979）, MusaCacheContext overrides them to return tensors because those tensors cannot be recovered after converting them to raw pointers. The naming is unclear and will rename transfer_data to transfer_operand.

superleo: thanks @hlin99 the DeviceOps select the backend implementation, but the operands have already been converted to pointers before it is called. These hooks preserve the tensor operands used by MUSA currently, other devices can use the defaults and do not need additional implementations.

lmcache/v1/platform/base/device_ops.py
superleo: thanks yihua @ApostaC , the existing record_completion_on_stream methods mirror the native c_ops interface and are bulk-bound by name, so they currently take an integer stream pointer. Chang it would either break that binding contract or require special handling in bind_native() and updates to the existing low-level callers.The new submit_methods are the higher-level platform hook, the default implementation converts the stream object to stream.ptr, while the dependent Stage 4.3 MUSA(PR 3979) implementation synchronizes the stream before recording. Other backends inherit the default and do not need additional implementations. Is it ok if I clarify this distinction in the document?

lmcache/v1/platform/base/cache_context.py
hlin99: is there a way to do something on parameters rather than creating a new interface, if it's just ptr/tensor diff on para?
c ops uses ptr, and python fallback uses tensor. there're already some examples in current python fallback. does it help?

ApostaC: Agree with @hlin99 , we don't want to introduce a new interface since it may create extra confusion for maintainers and new developers.

lmcache/v1/platform/base/device_ops.py
ApostaC: Since `record_completion_on_stream` is defined in `DeviceOps`, we can have MUSA's implementation override the existing function rather than introducing a new one -- similar to our discussion with @hlin99 above

ApostaC: I kind of feel like we are introducing some extra complexity here that is not 100% needed.

lmcache/v1/platform/base/cache_context.py
superleo: Agree with @ApostaC @hlin99  and reuse the existing cache-context APIs, removed the two transfer-data methods, keep staging buffers as tensors until the transfer path chooses pointer or tensor mode, and let MUSA override the existing paged-KV method.

lmcache/v1/platform/base/device_ops.py
superleo: agree with @ApostaC @hlin99 , reuse the existing `record_*_on_stream` APIs, keep a small python adapter so cuda forwards stream.ptr to the native recorder, while musa can override the same methods to synchronize before recording, existing integer-pointer callers will remain supported.

lmcache/v1/mp_observability/event_bus.py
```diff
@@ -165,13 +165,18 @@ def has_subscribers(self, event_type: EventType) -> bool:
...
 
     def publish_on_stream(self, stream: Any, event: Event) -> None:
-        """Schedule event recording as a CUDA host function on *stream*.
+        """Schedule event recording after earlier work on ``stream``.
 
-        Uses a C++ callback via ``cudaLaunchHostFunc`` so the callback
-        never touches the GIL, avoiding the CUDA-driver/GIL deadlock.
+        Native backends use a host callback that avoids acquiring the GIL on
+        the driver thread. Other platforms may provide equivalent ordering
+        through ``DeviceOps``.
```
ApostaC: nit: Let's remove this part of the docstring. I think it's explaining unnecessary details

lmcache/v1/platform/base/cache_context.py
```diff
@@ -96,16 +96,30 @@ def close(self) -> None:
...
     def get_temp_kernel_group_buffer(
         self, batch_idx: int, kernel_group_idx: int
     ) -> torch.Tensor:
         """Returns a typed temp-buffer view for a (batch, kernel-group)
-        pair."""
+        pair.
+
+        The multiprocess transfer path converts the returned tensor to
+        ``data_ptr()`` only for the pointer-based native path; tensor-backed
+        backends receive the tensor unchanged.
```
ApostaC: nit: drop this part of the comment. It's not related to the behavior of this function

lmcache/v1/platform/torch_ops.py
```diff
@@ -2649,16 +2649,40 @@ def get_gpu_pci_bus_id(device_id: int = 0) -> str | None:
...
     """Fallback: immediately enqueue the completion without stream ordering.
 
     Args:
-        cuda_stream_ptr: Ignored on non-CUDA path.
+        stream_or_ptr: Stream object or integer pointer. Ignored on the
+            non-CUDA fallback path after validation.
         kind: Dispatch key identifying the handler (e.g. "finish_write").
         payload: Opaque msgpack-encoded bytes forwarded to the handler.
     """
+    _coerce_stream_ptr(stream_or_ptr)
```
ApostaC: Since the argument `stream_or_ptr` is not used at all in this function. We don't need to call `_coerce_stream_ptr` here. (The same for the function below)

lmcache/v1/platform/base/device_ops.py
```diff
@@ -221,11 +261,71 @@ def drain_recorded_completions(self):
...
 
-    def record_completion_on_stream(self, *args, **kwargs):
-        return torch_ops.record_completion_on_stream(*args, **kwargs)
+    def record_completion_on_stream(
+        self,
+        stream_or_ptr: object,
+        kind: str,
+        payload: bytes,
+    ) -> None:
+        """Record a stream-ordered completion for the dispatcher.
```
ApostaC: I'm not sure why we need to rewrite the base class's behavior.

IIUC, the original design of this module (cc @zxue2 @hlin99 ) wants every device to overwrite the functions if needed, while keeping the base class's function a simple fallback. The goal here is to keep the base device ops simple and clear.

If we have any special logic for MUSA, we should overwrite the two functions in MUSA's own device ops class instead of introducing a special binding logic in the base class. 
I understand that you changed the function signature here to accept a stream object; it would be appreciated to update the existing C++ function to accept the stream object rather than a raw pointer. In this case, we can avoid changes in this file.

ApostaC (changes requested):
Now I understand the problem here. 

The current problems is that this PR still introduces quite a few if conditions and ad-hoc logics, which makes the code fragile and a bit too complicated.

Two high-level suggestions:
1. Make MUSA library be able to accept pointers (you guys can reconstruct the tensor/stream objects in your own library)
2. Refactor the existing CUDA function to make it accept tensors and stream.

I personally prefer 1 because 2 has a much bigger blast raduis.

superleo:
@ApostaC Thanks, the current musa native path already extracts data pointers before launching raw-pointer kernels, and TorchMUSA can wrap an external stream pointer. We’ll prototype a pointer-based solution with explicit layout, device and lifetime contracts.

lmcache/v1/platform/base/device_ops.py
zxue2: yes,   prefer either way below,
- MUSA record_completion_on_stream from binary .so  replacing baseline ops by bind_native  
-  Override  baseline record_completion_on_stream in musa/device_ops.py.   

Hi @superleo  is it feasible for MUSA implementation  to follow the baseline ops /CUDA style?

superleo: thanks @zxue2

maobaolong (approved):
LGTM

superleo:
Hi @ApostaC This current CI failure appears unrelated to #4242. The PR only changes MUSA platform code and tests; it does not affect the CUDA/vLLM benchmark path or generic multiprocess execution. All functional checks passed, and the only failure was a single-run throughput slowdown of 6.42% against a 5% threshold. This is likely benchmark variance, so a rerun should confirm it. The XPU event-signature failure is also outside this PR’s changed paths.

superleo:
hi Tony @hlin99 would you help to have a look?

ApostaC (approved):
Approve to unblock this PR

PR #4067, opened 2026-07-09
[GDS] always enable thread pool for disk I/O

glimchb:
The thread pool was previously gated on `use_gds`. Enable it unconditionally (default 4 threads) so POSIX-fallback paths also benefit from parallel I/O.  Users can set `disk_io_threads=0` in `extra_config` to disable it.

Also drops the deprecated `gds_io_threads` config key.

Generated with [Devin](https://devin.ai)

**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #4243, opened 2026-07-25
[fix]: disable the GDS thread pool in the no-pool batched_get test

deng451e:
#4067 made the disk I/O thread pool always-on by default, which broke test_batched_get_blocking_no_thread_pool's assumption that a default backend has the pool off. The failure only surfaces on cuFile-equipped CI nodes (TestGdsBackend is skipif-gated on GDS support), so most runs skip it. Pass disk_io_threads=0 explicitly, restoring the test's intent of exercising the no-pool path.

**If applicable**:


ApostaC (approved):

DongDongJu (approved):
LGTM

PR #4244, opened 2026-07-26
[SDK] Query tensor capture (#4042) + fix silent query corruption under continuous batching

Zefan-Cai:
## What this PR is

This is **#4042's work plus one correctness fix**, on top of current `dev`:

- commits `13633e4..23d08fc` are @rannnayy's from #4042 (authorship preserved), merged with current `dev` (no conflicts);
- one new commit fixes a silent query-corruption bug in the query-capture row mapping, with unit tests.

Opened as a superset so the fix can be reviewed against a tree where query capture actually works end-to-end. **If #4042 merges first I'll rebase this down to just the fix commit** — whichever is less work for you. Happy to split it either way.

## The bug

`QRingBufferCapture._build_q_step_state` maps query rows to ring slots **positionally**: it walks `metadata.requests` and advances a `row_cursor` by `op.end - op.start`. That assumes each request contributed exactly its store op's (chunk-aligned) token count of rows to the step's batched query tensor.

Under continuous batching a request's actual row count is its **scheduled** token count for the step. So any chunk-unaligned prompt tail — or a batch row order that differs from `metadata.requests` order — shifts every following request's rows, and the scatter writes **other tokens' query vectors** into their ring blocks. Silently: `retrieve(QUERY)` still returns a full-length, well-formed tensor, and the only signal is a `logger.warning`.

Minimal trigger, no decode mixing needed: request A with a 10,300-token prompt (store op covers `[0, 10240)`) followed by request B in the same all-STORE step. B's ring slots start at row 10,240; B's rows start at 10,300.

### Measured impact

Per-token-row fp64 fingerprints compared across two identical prefill+retrieve runs with a cache clear in between. Qwen3-8B, H200, `--enforce-eager --no-enable-prefix-caching`, `chunk_size=256`, prompt lengths 10159–16807:

| concurrent prefill streams | query rows identical across runs |
|---|---|
| 1 | 100.0% |
| 4 | 99.7% |
| 30 | **9.8%** |

KV was byte-identical (29/29 requests) in the same runs — KV transfers by block IDs and never touches this mapping. Layer-0 Q and K come from the same forward's hidden state, so floating-point nondeterminism cannot affect only one of them. The differing rows are not perturbed values: per-row relative deviation reaches the metric's theoretical maximum, i.e. they are **real query vectors belonging to other tokens**. Ruled out separately: incomplete `/cache/clear`, row permutation (`sort(x) == sort(y)` fails), cross-request contamination at matching offsets, uninitialized/zero slots (15615 of 15616 rows unique, no zeros).

Consequences: any query-driven algorithm computes on corrupted data at realistic concurrency — SnapKV's importance term is entirely query-driven, so ~90% of it was wrong at 30 streams — and the failure is invisible in single-stream testing, which is what CI covers today.

## The fix

Attribute rows to ops through ground truth instead of position:

- row `r` of the step's query writes its KV to GPU slot `attn_metadata.slot_mapping[r]`;
- the op's token `i` lives in GPU slot `block_ids[i // block_size] * block_size + i % block_size` (op block lists are already pre-sliced to `[start, end)`);
- intersecting the two identifies each op's rows regardless of batch composition or ordering. Each GPU slot is written at most once per step, so a full intersection is a bijection between an op's tokens and its rows.

`save_kv_layer` already receives the layer's `attn_metadata`; the connector now threads it into the dispatcher (one line). **The vLLM-side diff is unchanged.**

Behavioral improvements that fall out of the same change:

- **Mixed STORE/RETRIEVE steps capture normally.** The coarse `all(direction == "STORE")` bail-out is no longer needed — it was discarding whole steps and produced ~11.4k `Skip query` warnings in a single session.
- **Skips are per-request, not per-step.** An op that is only partially present in the step (chunked prefill split below a chunk boundary) or a ring-allocation failure now skips just that op instead of discarding requests that were already planned.
- **CUDA-graph padding rows** (slot `-1`, or rows beyond `slot_mapping`) drop out of the mapping naturally.

### Known residual limitation

An op whose leading tokens were computed in an *earlier* scheduler step cannot be captured in this one-shot design — those rows are gone. It is now **skipped with a warning instead of writing wrong data**. Fixing it properly means accumulating per-request Q across steps, which is the same machinery decode-step capture would need; noted in `docs/design/sdk/qringbuffer.md` and left as follow-up.

## Validation

- **13 new unit tests** (`tests/sdk/test_qringbuffer_row_mapping.py`), CPU-only: aligned baseline, chunk-unaligned tail not shifting the next request, batch row order permuted vs metadata order, mixed STORE/RETRIEVE, op partially present, ring exhaustion, `slot_mapping` padding, missing `slot_mapping`, no-store-ops, plus a randomized fuzz over batch compositions that asserts every accepted op captures exactly its own rows.
- Tests pass against a real `vllm==0.25.1` install as well as with a stubbed `vllm` on CPU.
- An end-to-end A/B (same box, same config, fixed vs stock) on an H100 is running; I'll post the numbers as a comment.

## Suggested follow-up

A CI regression at **≥8 concurrent prefill streams**: two identical runs, assert query tensors byte-identical. Concurrency 1 passes even with the bug, which is exactly why this survived.

## Docs

`docs/design/sdk/qringbuffer.md` gains a "Row-to-token attribution" section explaining the invariant, so the next reader does not reintroduce a positional cursor.


Zefan-Cai:
### Note on the DCO check

My own two commits are signed off. The DCO check still reports **2 commits incorrectly signed off**: `23d08fc` ("lazy import") and `d7a1319` ("lazy import, cov query 0 if not opt-in"), both inherited from #4042 and authored by @rannnayy. A sign-off is an attestation by the commit's author, so I can't add those on her behalf.

Worth flagging that this is **not specific to this PR** — the same two commits are the head of #4042's branch, so #4042 is blocked by the same check. Two ways out, whichever you prefer:

1. @rannnayy signs off those two commits on the #4042 branch (`git rebase --signoff 60216f3..HEAD` or a DCO remediation commit); I then rebase this PR on the updated branch.
2. #4042 merges first (after the sign-off), and I rebase this PR down to **just the fix commit**, which is already signed — that path leaves this PR with a single commit and no inherited DCO debt.

Option 2 is probably the least work for everyone, and it's what I'd suggest unless you'd rather review the fix together with the capture code.

Zefan-Cai:
### End-to-end A/B on an H100 — stock vs fixed, same box, same config

Qwen3-4B, `vllm==0.25.1` + the SDK's intermediate-tensor patch, `--enforce-eager --no-enable-prefix-caching`, `chunk_size=256`, `max_num_batched_tokens=20480`, `max_num_seqs=32`, `lmcache.q.ring_depth=2`, L1 pool 400 GB. 30 random prompts of 10,000–16,800 tokens (deliberately chunk-unaligned). Each arm runs the identical concurrent prefill workload **twice** with `POST /cache/clear` in between, retrieves every stream's KV and QUERY through the SDK, and compares per-token-row fp64 fingerprints (sum, sum of squares, sum of |x|, accumulated over all layers) with **exact equality**. The fixed arm additionally runs a third pass at concurrency 1 as an uncontended reference.

| 30 concurrent streams | stock (`23d08fc`) | fixed |
|---|---|---|
| query rows captured (of 400,384 chunk-aligned prompt rows) | 66,048 = 16.5% | 178,432 = 44.6% |
| of those, identical across two identical runs | **56,483 = 85.5%** | **178,432 = 100.00%** |
| streams with ≥1 corrupted query row | **3 of the 6 that captured anything** | **0 of 26** |
| streams with a complete query tensor | 4 / 30 | 5 / 30 |
| streams with no query tensor at all | 14 / 30 | 4 / 30 |
| KV rows identical across runs (control) | 100.00% | 99.97% |
| `Skip query` warnings | 22 | 46 |

On the fixed branch, the rows captured in both the concurrent and the serial (concurrency-1) arms also agree with the serial reference: 212,992 / 212,992 = **100.00%** — *within floating-point rounding* (rtol 1e-3 on the fp64 row fingerprints). Two precision notes so this claim is exactly as strong as the data: (a) prefill numerics are not batch-invariant, so bitwise equality across different batch compositions is not expected — the exact-equality comparison is the cross-run one above; (b) the serial arm itself captured only part of some streams (back-to-back submissions can outrun ring-block reclamation at `ring_depth=2`), so this comparison covers the 212,992-row overlap, not all captured rows. It is still the stronger claim: concurrent capture produces the *same values* as uncontended capture, not merely stable ones.

**Correctness.** The worst stock stream matched **0 of 4,096** rows across two identical runs — its entire captured query tensor was other tokens' data. Streams 2 and 8 were partially wrong (10,008/11,264 and 11,915/16,128).

**Coverage, as a side effect.** The fix captures **2.7× more rows** (178,432 vs 66,048): per-request skipping replaced whole-step discarding, so requests that previously got nothing now get at least the chunks whose ops were fully present in their step.

#### Read the stream count, not the row rate

The row-level figure here (85.5% identical) is milder than the 9.8% measured earlier on Qwen3-8B / H200, and that is not a contradiction — severity scales with how often several requests share a forward step. With `max_num_batched_tokens=20480` and 10–16.8k-token prompts, many steps here hold a single request, and single-request steps are trivially aligned, so most of what stock captured came from steps that could not be misaligned in the first place. The signal stable across both setups is the stream-level one: wherever multiple requests do share a step, the captured queries are silently wrong, and the affected fraction grows with batching pressure.

#### On the KV control

100.00% in the stock arm, 99.97% in the fixed arm. The single differing stream (29) also differs from the serial reference (15,238/15,360) and is one of the most heavily chunk-split streams: FlashAttention prefill is not batch-invariant, so a different split changes the reduction order and the last bits of the KV, and run-to-run scheduling differs. KV transfers by block IDs and never passes through the row-attribution code. `VLLM_BATCH_INVARIANT=1` (as `e2e_kv_edit` sets) removes it.

#### The residual limitation, now quantified

Correct attribution does not by itself make concurrent-prefill capture *complete*: 5/30 streams get a full query tensor, 16/30 partial, 4/30 empty — 44.6% of rows. At this concurrency vLLM chunk-splits most requests across scheduler steps, and an op whose leading tokens were computed in an earlier step has no query rows in the current one, so it is skipped. The captured *amount* therefore varies run to run with scheduling; the captured *values* never do.

The follow-up that closes this — reserving a request's ring range on first sight and filling it incrementally across steps — is the same machinery decode-step Q capture needs, since decode emits one query row per step. Happy to take it on as the next PR if it fits the roadmap.

#### Two environment prerequisites worth documenting

Both cost real debugging time and surface on the vLLM side only as a 300 s `register_kv_caches` timeout:

1. The LMCache server must be able to **see the vLLM worker's GPU** — it resolves the registered KV cache by device UUID for IPC, so restricting the server's `CUDA_VISIBLE_DEVICES` to a different device fails with `RuntimeError: Device UUID ... not found in the discovered devices`.
2. When `lmcache.c_ops` is not built (plain `pip install -e .`), `GPUCacheContext` falls back to a cupy path and fails with `ModuleNotFoundError: No module named 'cupy'`; `cupy-cuda13x` / `-cuda12x` fixes it. A startup check would save the next person the same hour.

wanke1997:
Hey @Zefan-Cai , thanks for opening a PR to address the bugs that I found yesterday. 

This fix addresses the correctness issue in query-row attribution by mapping each query row to its actual KV slot, thereby preventing silent cross-request corruption under continuous batching. 

However, it does not guarantee that all query rows are captured when one cache chunk is generated across multiple scheduler steps. In that case, the current implementation skips the incomplete chunk instead of combining query rows from multiple steps.

As a result, the captured Q data is now correct, but it may be incomplete for chunked prefill and decode workloads, which will block the implementation of the algorithms that need Q cache for decode steps.

A possible solution is to keep a small amount of per-request state across scheduler steps. Each step can fill in the query rows that are currently available, and the Q cache entry should only be published after all rows in the chunk have been captured.

Zefan-Cai:
Thanks @wanke1997 — that is exactly the right characterization. After this fix the Q cache is **fail-closed** (whatever is captured is correct) but **not complete** for chunked-prefill or decode workloads: the A/B above quantifies it — at 30 concurrent streams only 5/30 requests end up with a full-length Q, 44.6% of chunk-aligned rows are captured, and the skips are precisely the store ops whose leading tokens were computed in an earlier scheduler step.

Agreed on the follow-up shape too. Concretely, what I have in mind for the cross-step accumulation:

1. **Scheduler side** — include each scheduled request's per-step token range and block IDs in the connector metadata *before* its store op materializes. Today `metadata.requests` only carries requests that already have an op, so the worker cannot attribute rows that arrive ahead of the op — that is the structural reason accumulation is impossible in the current design, not just unimplemented.
2. **Worker side** — reserve the request's ring blocks at first sight and scatter rows incrementally as steps contribute them, with an explicit per-chunk coverage bitmap; submit `STORE_Q` only when a chunk's bitmap is complete. Publishing at chunk granularity (how the store path chunks anyway) bounds in-flight staging to ~one chunk per request and keeps ring sizing predictable.
3. **Lifecycle** — free reservations on request finish/abort; on preemption-with-recompute, invalidate the affected bitmaps (slot_mapping changes, so previously staged rows must not be published).
4. **Decode capture falls out naturally** — a decode step just contributes one row to the current chunk's bitmap. That is the piece decode-time algorithms (R-KV-style periodic eviction) need from this path, so the same machinery serves both gaps.

The "explicit completion tracking before publishing" you mention is exactly the bitmap in (2) — without it we would just trade silent corruption for silent truncation, since `retrieve()` today cannot distinguish a complete tensor from a partial one.

Happy to take this as the follow-up PR once this one (or #4042) lands — unless you or @rannnayy would rather fold it in here.

rannnayy:
Hi @Zefan-Cai and @wanke1997!

Thanks for testing the API and finding out the bug!

Since this PR fixes a bug in #4042 that is nearly merged, I have folded this PR's commit into #4042 and will have it merged by today. I hope both of you don't mind me folding in your changes. All authorship and commit preserved.
Please find the folded-in commit here: https://github.com/LMCache/LMCache/pull/4042/commits/2f247861eba7dcae7a6705dbab9b9416968c5f02.

I am working on your follow-up suggestion regarding getting the Q for the decode steps. Will make sure to tag both of you for rounds of feedbacks afterwards! Thank you!

PR #4245, opened 2026-07-26
[Feat] Time GC pause in mp server

zhengfeihe:

Adds an opt-in GC monitor to the MP server. It uses `gc.callbacks`, which are invoked by CPython on every garbage collection cycle, and logs collections that exceed a configurable threshold.

A full (generation 2) garbage collection is a stop-the-world operation that traverses the entire heap. As a result, it can occur during store/retrieve request handling and introduce tail latency that is not accounted for elsewhere in the server. This logging mechanism helps users identify and investigate unexpected latency spikes caused by GC.

Example log

```
GC gen2 143.6ms collected=243 uncollectable=0 builtins.function=60915 builtins.tuple=57130 builtins.dict=38030 builtins.cell=20575 builtins.list=17251 builtins.builtin_function_or_method=15536 weakref.ReferenceType=15145 builtins.getset_descriptor=13784
```

Follow-up work on GC freezing and additional GC optimizations will come later, as more testing is needed.

**If applicable**:


ApostaC (approved):
LGTM!

ApostaC:
@zhengfeihe remember to fix the dco issue, thanks!

DongDongJu (approved):
LGTM for functionality itself.
But, what is for? Is CPython GC causing problem?

zhengfeihe:
> @zhengfeihe remember to fix the dco issue, thanks!

Sorry, my bad. Thanks for the reminder. Fixed now.

zhengfeihe:
> LGTM for functionality itself. But, what is for? Is CPython GC causing problem?

No specific incident reported yet. This is preventive instrumentation. Long-running Python services can have meaningful GC overhead, and gc tunning is usually needed, eg: vLLM already has bunch GC-related optimizations. LMCache has not explored this area much yet, so I would rather measure gen2 pauses first. This also gives users a way to rule GC in or out when investigating unexplained latency spikes.

PR #4246, opened 2026-07-26
feat(core): Introduce gRPC transport type

maobaolong:

This PR replaces the custom ZMQ-based multiprocess message queue with gRPC. Each MP request type maps to a typed protobuf RPC, while the existing LMCache adapter APIs and legacy `tcp://` / `ipc://` configuration aliases remain supported.

**Benefits of gRPC**:

- **Explicit, typed protocol contracts**: protobuf defines request and response schemas, catching protocol drift earlier than the previous custom envelope.
- **Simpler request routing**: one gRPC method per `RequestType` removes manual request-type framing and dispatch logic.
- **Better startup and outage behavior**: `wait_for_ready`, structured status codes, and asynchronous futures preserve pending requests during server startup or temporary outages.
- **Standard transport tooling**: HTTP/2, TCP and Unix socket support, generated clients and servers, and common tracing and debugging tools reduce custom networking code.
- **Lower migration cost**: existing MP call sites remain unchanged, and legacy transport URLs are translated to their gRPC equivalents.

**RPC extensibility**:

- Python/protobuf bindings are derived automatically from the `MessageQueue` service descriptor and `ProtocolDefinition`; codecs are compiled once at import rather than using reflection on each request.
- The bounded Batch fast path uses an RPC-agnostic `method_id + payload` envelope, so adding an RPC does not require another Batch oneof arm.
- Adding an ordinary RPC requires its `RequestType` and `ProtocolDefinition`, protobuf request/response messages plus one service method, the business handler, and contract/round-trip tests. It does not require a `typed_rpc.py` registry entry or per-RPC conversion wrappers.

**Performance**:

Local loopback benchmark against `dev@9997afbc`: 20,000 blocking `PING` requests, one client, 8 handler workers, concurrency 16, median of 5 runs.

| Transport | `dev` ZMQ | This PR gRPC | Throughput change | Average latency |
| --- | ---: | ---: | ---: | ---: |
| TCP | 7,440 req/s | 23,785 req/s | 3.20x | 2.146 ms -> 0.67 ms (-68.8%) |
| Unix socket | 7,623 req/s | 24,198 req/s | 3.17x | 2.095 ms -> 0.66 ms (-68.5%) |

The client transparently combines overlapping control-plane calls into a bounded protobuf micro-batch. Single callers continue to use dedicated unary RPCs without a batching delay. The descriptor-driven refactor itself improved the gRPC-only benchmark from 20,488 to 23,785 req/s on TCP and from 21,109 to 24,198 req/s on Unix sockets.

**Special notes for reviewers**:

- This changes the MP wire protocol; LMCache MP clients and servers should be upgraded together.
- `lmcache_mq.proto` is the single source of truth. Generated protobuf/gRPC stubs are never tracked by Git; full package builds and test setup generate them with the pinned `requirements/proto.txt` toolchain. To regenerate manually: `python -m lmcache.v1.multiprocess.transport.grpc_impl._proto_gen._generate`.
- The CLI-only wheel skips gRPC stub generation because it does not provide the multiprocess runtime.
- The gRPC runtime is loaded lazily so importing protocol helpers does not initialize native gRPC libraries.

**If applicable**:


ApostaC (commented):
Quick comment: I found that there are a lot of new helper functions in `mq.py` that handles type conversions. Any chance we can modularize or better organize those codes?

lmcache/v1/multiprocess/mq.py
```diff
@@ -1,67 +1,1377 @@
 # SPDX-License-Identifier: Apache-2.0
```
KuntaiDu, line 1: This file is the major concern --- the code change is too large in this file. It seems to be some wrappers around functions that convert between python <> grpc. Can we do it in a more automatic way, like function wrapper?

KuntaiDu (commented):
Other files are short and looks good to me. See comments.

maobaolong:
Thanks for the review, will do a refactor.

lmcache/v1/multiprocess/mq.py
maobaolong: Addressed in 6a1feadd. I moved the Python/protobuf adapters and the TypedRpcSpec registry out of mq.py into grpc_impl/typed_rpc.py, so mq.py now contains only client/server transport control flow (2,755 -> 1,052 lines). I kept the adapters explicit rather than reflection-based because several RPCs require non-1:1 conversions such as IPC wrappers, optional sentinels, torch dtypes, and serialized contexts. The typed RPC, round-trip, batching, and compatibility suites pass.

maobaolong:
@ApostaC Addressed the modularization feedback in 6a1feadd: all Python/protobuf conversion helpers and the TypedRpcSpec registry now live in grpc_impl/typed_rpc.py, while mq.py is limited to transport control flow. The adapters remain explicit for the RPCs with custom IPC/domain conversions, and the full typed/compatibility test set passes.

maobaolong:
Follow-up refactor is in `0a2ea00d`. The Python/protobuf bindings are now generated automatically from the `MessageQueue` service descriptor plus `ProtocolDefinition`; the 134 per-RPC conversion helpers and manual registry were removed. Batch also uses an RPC-agnostic envelope, so a new ordinary RPC no longer needs a registry entry, conversion wrappers, or Batch oneof fields.

maobaolong:
The current PR Full Build failures do not appear to be specific to this PR.

On this PR, the red checks are the CUDA wheel build jobs inside PR Full Build (Build main / Build artifacts, Build cu129 / Build cu129 artifacts, and the arm64 variant), while the regular test lanes are green.

The same PR Full Build pattern is also happening on unrelated PRs today, for example:
- Fix MP event IPC lifetime on ROCm: https://github.com/LMCache/LMCache/actions/runs/32850532428
- fix(sglang): distinguish single-head MHA from MLA: https://github.com/LMCache/LMCache/actions/runs/32834482260

In both of those runs, the CUDA wheel build jobs failed in the same workflow, which suggests this is a shared build-lane issue rather than something introduced by the gRPC transport changes here.

.buildkite/k3_harness/setup-env.sh
```diff
@@ -261,8 +261,10 @@ echo "--- :python: Installing LMCache from source"
 # Skip setuptools_scm git describe; the repo carries non-PEP-440 tags
 # (nightly, nightly-cu13) that crash the newer vcs_versioning backend.
 export SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LMCACHE="${SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LMCACHE:-0.0.0+ci}"
+uv pip install -r requirements/proto.txt
 uv pip freeze | sort > /tmp/env-before-lmcache.txt
 uv pip install -e . --no-build-isolation
+python -m lmcache.v1.multiprocess.transport.grpc_impl._proto_gen._generate
```
ApostaC, line 267: In order for normal user to use install and use lmcache, do they need to run this command locally as well?

lmcache/v1/multiprocess/transport/grpc_impl/proto/lmcache_mq.proto
ApostaC, line 1: Is it possible to split this as what we did in `lmcache/v1/multiprocess/protocols/`?

lmcache/v1/multiprocess/transport/grpc_impl/typed_rpc.py
```diff
@@ -0,0 +1,666 @@
...
+    Any,
+    Callable,
+    Union,
+    get_args,
+    get_origin,
+    get_type_hints,
+    is_typeddict,
+)
+import enum
+import pickle
```
ApostaC, line 21: We should avoid use pickle due to security issues. Let's use msgspec for serialize & deserialize.

lmcache/v1/multiprocess/affinity_pool.py
```diff
@@ -27,8 +27,8 @@
 class AffinityThreadPool:
     """Thread pool that routes tasks to workers by affinity key.
 
-    Not thread-safe: ``submit()`` must be called from a single thread (the
-    ``MessageQueueServer`` main loop).
+    Thread-safe: concurrent transport workers may submit tasks while routing
+    for each affinity key remains stable.
```
ApostaC, line 31: We deliberately make it non-thread safe, because in the old code we know that there will only be 1 thread calling this function -- so we can do lock-free

lmcache/v1/multiprocess/mq.py
ApostaC, line 1: I feel like this is almost a rewrite of the existing code.

Can we have two new classes instead of rewriting the existing ones. 

Additionally, at a high level, can we make ZMQ and GRPC co-exist for a period of time and remove zmq once we think GRPC is stable enough?

ApostaC (changes requested):
Did another pass. I think in general this could work.

High-level concerns:
- Comparing to ZMQ, GRPC introduced a lot of new code for serialization and deserialization (under `grpc_impl`). And this part of the code seems not easy to maintain
- If we merge this PR, the old ZMQ path will be completely dropped. This means that it will be hard to maintain if we find issues in GRPC code path.

.buildkite/k3_harness/setup-env.sh
maobaolong: Normal wheel users should not need to run this command. Package builds generate the stubs and include them in the wheel. This explicit CI command is intended as a generation/health check.

lmcache/v1/multiprocess/transport/grpc_impl/proto/lmcache_mq.proto
maobaolong: Is this what you expected? https://github.com/LMCache/LMCache/pull/4819/changes#diff-3d71b1ca4e9c90e800327fe3c2545915218fca27474721d2988a327ffdedbdd1R452

maobaolong: The layout under `protocols/` can also be:
```
common.proto
engine.proto
controller.proto
blend.proto
blend_v2.proto
blend_v3.proto
observability.proto
p2p.proto
service.proto
```

lmcache/v1/multiprocess/affinity_pool.py
maobaolong: That was true for the ZMQ main-loop model, where submit() had a single caller. 
The gRPC server invokes handlers from its worker pool concurrently, so the affinity mapping now has multiple submitters. The lock only protects key-to-slot assignment, task execution remains concurrent.

lmcache/v1/multiprocess/mq.py
maobaolong: This is a hard choice, Initially, i intended to introduce an abstracted layer to keep both zmq and grpc transport types, but after several round iterators, i found i cannot focus on the new added grpc transport type, so i decided to drop the abstraction layer by https://github.com/LMCache/LMCache/pull/4246/changes/f514173dc48776a3920c71a75707a485a9037f6b .

But I totally agree to keep both of them before we merge grpc transport type, and we have to keep them both, otherwise, we cannot rollback to a stable `zmq`.  

So I'll bring the abstraction layer back after the new added grpc transport type can be a good state.

The goal of the recent is:
- Less efforts while introduce new field/rpc/service
- Easy to be used for client
- Pass all the exist CIs.

PR #4248, opened 2026-07-26
fix: apply non-deprecated config aliases instead of silently dropping them

divya0795:

`_resolve_config_aliases` (`lmcache/v1/config_base.py`) only remapped a config
alias to its current key when the alias was *also* listed in
`_DEPRECATED_CONFIGS`. Every alias that lives only in `_CONFIG_ALIASES` matched
neither the deprecated branch nor the `config_definitions` branch, so it fell
through to the `else` "Unknown configuration key" branch and was **silently
dropped** (only a warning was logged, the value never reached the current key).

In `lmcache/v1/config.py` that makes ~9 of the documented backward-compat
aliases no-ops today:

- `enable_xpyd` → `enable_pd`
- `controller_url` → `controller_pull_url`
- `lmcache_worker_port` → `lmcache_worker_ports`
- `nixl_peer_host` / `nixl_peer_init_port` / `nixl_peer_alloc_port` /
  `nixl_proxy_host` / `nixl_proxy_port` / `nixl_role` → the corresponding
  `pd_*` P/D names

Setting any of these via env var, YAML file, or `from_dict` had no effect on
the intended current key.

The fix resolves each key in this order:

1. `config_definitions` first — so a name that is both a live key and a legacy
   alias keeps its current-key meaning (e.g. `nixl_buffer_size`, which is both
   a real config key and an alias for `pd_buffer_size`);
2. then `config_aliases` for **every** alias, regardless of deprecation status;
3. then keep deprecated-only keys (e.g. `nixl_peer_port`) under their original
   name for backward compatibility.

Deprecation warnings still fire independently, and unknown keys are still
dropped with a warning. The two aliases that already worked (`plugin_locations`,
`external_backends`, which are in both `_CONFIG_ALIASES` and
`_DEPRECATED_CONFIGS`) keep working.

- The change is in the shared helper, so it applies to every public constructor
  that calls it: `from_env`, `from_file`, `from_dict`, and
  `update_config_from_env`.
- New tests in `tests/v1/test_config.py::TestConfigAliasResolution` cover the
  `from_dict` and `from_env` alias paths, the deprecated+alias path
  (`external_backends`), and the alias/definition name collision
  (`nixl_buffer_size`). The first two fail on `dev` and pass with this change;
  the rest of `test_config.py` is unchanged (79 → 83 passing locally,
  `pytest tests/v1/test_config.py`).
- No behavior change for keys that were already resolved correctly; this only
  restores the documented aliasing that was being dropped.

**If applicable**:

### DIFF
--- a/lmcache/v1/config_base.py
+++ b/lmcache/v1/config_base.py
@@ -188,24 +188,40 @@ def _resolve_config_aliases(
     config_aliases: dict,
     deprecated_configs: dict,
 ) -> dict:
-    """Resolve configuration aliases and handle deprecated configurations."""
+    """Resolve configuration aliases and handle deprecated configurations.
+
+    For each key in ``config_dict`` the resolution order is:
+
+    - If the key is listed in ``deprecated_configs``, a deprecation warning is
+      logged, independently of how the key is subsequently resolved.
+    - If the key is a current, valid key in ``config_definitions`` it is kept
+      as-is. This is checked before ``config_aliases`` so that a name which is
+      both a live key and a legacy alias keeps its current-key meaning.
+    - Otherwise, if the key is an alias in ``config_aliases``, its value is
+      remapped to the current key it points to. This happens for every alias,
+      not only those that are also listed in ``deprecated_configs``.
+    - Otherwise, if the key is deprecated but has no alias mapping, it is kept
+      under its original name for backward compatibility.
+    - Any remaining key is unknown and is dropped with a warning.
+    """
     resolved = {}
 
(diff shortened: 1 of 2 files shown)


ApostaC:
Hi @divya0795 , this PR is for non-MP mode only. We will gradually stop maintaining the code for non-MP mode and deprecate it in 2026 Q3-Q4 .

Thanks for your contributions!

divya0795:
Thanks @ApostaC for the info.

PR #4251, opened 2026-07-26
[Fix][NIXL] Close static storage agent during adapter shutdown

2xdevv:

The static NIXL L2 adapter owns a `NixlStorageAgent` whose cleanup releases prepared descriptor-list handles, deregisters NIXL memory, and closes file-backed storage descriptors. However, `NixlStoreL2Adapter.close()` previously stopped only the event loop and event notifiers without calling the agent cleanup. Normal shutdown or adapter reconfiguration therefore leaked NIXL resources and up to `pool_size` storage file descriptors per adapter.

- Close the `NixlStorageAgent` after the adapter’s event-loop thread has stopped.
- Use `try`/`finally` so the event loop and event notifiers are still closed if NIXL cleanup raises.
- Add regression coverage that verifies descriptor-list handles are released, memory registrations are removed, and every storage FD is closed with `EBADF`.
- Re-enable the eviction tests that were skipped because of the FD leak.
- Update the re-enabled usage assertions to the current `AdapterUsage` interface.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
@@ -578,7 +578,9 @@ class NixlStoreL2Adapter(L2AdapterInterface):
         with self._lock:
             return self._completed_load_tasks.pop(task_id, None)
 
-    def close(self):
+    def close(self) -> None:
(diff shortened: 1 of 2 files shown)


ApostaC (approved):
LGTM!

zhengfeihe (approved):
It's proper close fix.(  Help unblock auto-merge

PR #4252, opened 2026-07-26
[Examples] Add an R-KV token dropping and benchmarking notebook

wanke1997:

This PR adds an end-to-end example of **R-KV token dropping** through the LMCache
SDK. The notebook retrieves a request's KV cache and query states after prefill,
scores every past token, drops half of them, remaps the survivors, and continues
decoding from the compressed cache.

R-KV extends SnapKV's attention-importance term with a **redundancy** term, so a
token is kept only if it is both attended-to *and* says something new:

$$\text{score}(j) = \lambda \cdot \text{importance}(j) - (1-\lambda) \cdot \text{redundancy}(j)$$

The redundancy term compares **all pairs of keys** — $O(n^2 d)$ per layer against
$O(w n d)$ for SnapKV. At $n = 16128$ that is ~250x more arithmetic per layer, and
a literal implementation also materializes an $n \times n$ matrix per layer. A
direct port of the reference code needed **over an hour** for the 30 prompts used
here. Making R-KV practical is what the example is about.

### Layout

Per review feedback, the readable version is the notebook and the optimized ones
live in a subdirectory:

| File | Implementation |
| --- | --- |
| `examples/token_dropping/rkv_token_dropping.ipynb` | Route A, literal all-pairs formulation — the reference, and the one to read first |
| `examples/token_dropping/rkv_variants/rkv_variant_cpu_exact.ipynb` | Route A, exact and linear, on the CPU |
| `examples/token_dropping/rkv_variants/rkv_variant_gpu_exact.ipynb` | Route A, exact and linear, on the GPU |
| `examples/token_dropping/rkv_variants/rkv_variant_buffered_cpu.ipynb` | Route B, buffered / chunked, on the CPU |

Each of the four is **standalone** — same setup, same prompts, its own baseline,
its own accuracy and plot — and each documents the trade-off it is making. The
main notebook links to the other three, and they link back.

### Measured results

All four were executed end to end on 30 LongBench-v2 prompts (mean 14114 tokens)
with Qwen3-8B on one H200 and a 96-core Xeon, dropping 50% of tokens. The numbers
in the notebooks are those measurements, and the executed outputs are committed.

| | one-shot | `cpu_exact` | `gpu_exact` | `buffered_cpu` |
| --- | --- | --- | --- | --- |
| Formulation | route A, literal | route A, linear | route A, linear | route B, chunked |
| Exact? | yes (the reference) | yes, rel. err ~6e-7 | yes, rel. err ~6e-7 | no, 96.2% agreement |
| Redundancy cost | $O(n^2 d)$ | $O(n c d)$ | $O(n c d)$ | $O(n c d)$ |
| VRAM | ~9.8 GiB | none | **1.55 GiB** | none |
| Compress 30 requests | 41.6 s | 67.3 s | **37.5 s** | **29.9 s** |
| Decode throughput | 1766 tok/s | 1785 tok/s | 1774 tok/s | 1774 tok/s |
| Accuracy (baseline 10–11/30) | 13/30 | 14/30 | 14/30 | 13/30 |

Every implementation compacted all 30 requests to exactly 50%. The self-tests
pass: the tiled reference matches the literal one to 4.7e-10, both exact variants
reproduce the reference redundancy to ~6e-7 with a **bit-identical kept-token
set**, `gpu_exact` returns one distinct result over five identical calls, and the
buffered variant agrees with route A on 96.16% of kept tokens.

### Fix to `examples/token_dropping/utils.py`

That investigation found that `rerotate_k_cache` carried out the rotation — six
multiplies and two adds per element — in the cache's own dtype. In bfloat16 that
rounds at every step. Measured on keys with a standard deviation of 3:

| arithmetic | keys changed vs fp32 | rotate-away-and-back error |
| --- | --- | --- |
| bfloat16 | 50% | 2.5e-1 |
| **fp32** (this PR) | — | **6.3e-2** |

The arithmetic now runs in fp32 and is rounded back to the cache dtype exactly
once. Measured cost of the change: **+0.2 s** to compress 30 requests for three of
the four notebooks — `modify` is dominated by waiting on LMCache, so the extra
arithmetic fits in that slack — and **+2.0 s (+7%)** for `buffered_cpu`, whose
compression is short enough that the re-rotation is a visible share of it. This
does *not* remove the CPU/GPU split above, which is the final rounding back to
bfloat16.

This file is shared with the SnapKV and random-dropping notebooks, which benefit
from the same fix.

### Also

`pyproject.toml` gains a `codespell` `ignore-regex` for base64 image payloads.
Executed notebooks embed matplotlib plots, and that data will eventually contain
something that looks like a typo — it already did here (`fPr`). The existing
notebooks in `dev` have the same exposure and have simply been lucky.

- The notebooks require an LMCache server started with `--enable transfer_query`
  and a vLLM server with `lmcache.mp.transfer_intermediate_tensors` enabled. Setup
  commands are in the first cell of each notebook; the vLLM patch is described in
  [`examples/token_dropping/README.md`](../blob/dev/examples/token_dropping/README.md).
- vLLM prefix caching is disabled so the measured decode-throughput change is
  attributable to the tokens dropped through LMCache rather than to vLLM's own
  cache.
- The four implementations use the same global R-KV formulation except
  `buffered_cpu`, which evaluates redundancy within chunks and is intentionally
  approximate.
- The benchmark numbers are hardware- and workload-dependent. Please read them
  together with the "Reading these numbers" section rather than as guarantees.

**If applicable**:

### DIFF
--- a/examples/token_dropping/README.md
+++ b/examples/token_dropping/README.md
@@ -23,6 +23,19 @@ smaller model and
 | --- | --- | --- |
 | [random_token_dropping.ipynb](./random_token_dropping.ipynb) | Drops a random subset of past tokens. Uses the KV cache only. | No |
 | [snapkv_token_dropping.ipynb](./snapkv_token_dropping.ipynb) | SnapKV: keeps the first and last window, as well as the tokens the recent-window queries attend to most. Needs each request's query tensor to score importance. | Yes |
+| [rkv_token_dropping.ipynb](./rkv_token_dropping.ipynb) | R-KV: SnapKV's importance term minus a **redundancy** term, so a token is kept only if it is both attended-to and says something new. Three optimized variants live in [rkv_variants/](./rkv_variants). | Yes |
+
+R-KV's redundancy term compares all pairs of keys, which is $O(n^2)$ in both
+time and memory. [rkv_token_dropping.ipynb](./rkv_token_dropping.ipynb)
+implements it the way the paper describes it; the three notebooks in
+[rkv_variants/](./rkv_variants) are standalone alternatives that make it
+cheaper:
+
+| Notebook | Redundancy term | Exact? | VRAM |
+| --- | --- | --- | --- |
+| [rkv_variant_cpu_exact.ipynb](./rkv_variants/rkv_variant_cpu_exact.ipynb) | Rewritten as one inner product, so the $n \times n$ matrix never exists. | Yes | None |
+| [rkv_variant_gpu_exact.ipynb](./rkv_variants/rkv_variant_gpu_exact.ipynb) | The same rewrite, tiled over rows on the GPU. Fastest of the four. | Yes | ~1.55 GiB |
+| [rkv_variant_buffered_cpu.ipynb](./rkv_variants/rkv_variant_buffered_cpu.ipynb) | Keys compared only inside a chunk, making the matrix block-diagonal. | No, ~96% token agreement | None |
 
 [snapkv_colab.ipynb](./snapkv_colab.ipynb) is a demonstration done in Google
 Colab Notebook, which can also be accessed here:
--- a/examples/token_dropping/rkv_token_dropping.ipynb
+++ b/examples/token_dropping/rkv_token_dropping.ipynb
@@ -1,0 +1,1557 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7fb27b941602401d91542211134fc71a",
+   "metadata": {},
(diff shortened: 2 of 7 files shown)


ApostaC:
Thanks for your contribution! Also, feel free to reach out to Kuntai in LMCache slack workspace

@KuntaiDu Please take a look at this as well.

KuntaiDu (approved):
LGTM!

ApostaC (approved):

PR #4253, opened 2026-07-26
fix(vllm): handle the vLLM 0.26.0 fused KV cache layout in group edits

### LINKED ISSUE #4247: [Bug] LMCacheMPConnector silently corrupts output on shared-prefix requests for hybrid Mamba/GDN models under vLLM 0.26.0 (EngineKVFormat 10)
## Environment

- **LMCache:** `0.5.2` (tag `v0.5.2`); also reproduced on `0.5.3.dev35` (`dev` @ `927897db`, i.e. *with* #4206)
- **vLLM:** `0.26.0` (source build). **Not** reproducible on vLLM `0.25.1`.
- **Model:** `nvidia/Qwen3.6-27B-NVFP4` — `Qwen3_5ForConditionalGeneration`, hybrid Mamba/GDN (linear) + full attention, NVFP4 weights
- **Connector:** `LMCacheMPConnector`, `kv_role=kv_both`, separate MP server over ZMQ, `fs_native` L2
- **GPU:** NVIDIA GB10 (DGX Spark), aarch64, compute capability 12.1, CUDA 13.0, TP=1
- **Attention backend:** FlashInfer · **KV cache dtype:** fp8 · **Spec decode:** MTP, `num_speculative_tokens=2`
- **vLLM flags:** `--enable-prefix-caching --enable-chunked-prefill --max-num-batched-tokens 1600 --mamba-block-size 8` (`mamba_cache_mode=align`, attention `block_size=1600`, mamba page padded 0.88%)
- **LMCache MP server:** `--chunk-size 1600 --l1-size-gb 30 --eviction-policy LRU --eviction-trigger-watermark 0.90`

## Symptom

With `LMCacheMPConnector` enabled, **any request that hits a shared prefix returns multilingual token salad** and runs to `max_tokens`. There is **no error anywhere** — no exception, no `result=False`, no MP-server log entry. Stores and retrieves all report success.

Example output (the prompt asked a one-sentence question in Chinese):

```
mấyadoras Pieمع acqu楚 cab1igiaLK Champ Transaction山湖xid近了国际社会会把 Chauactic
MethodInfo seat浦东新区charsトム Gabin努力做到 lk珊珊ところがormenteintis假}- tato腊迅 caval
```

## Reproducer

`lmcache-repro.py` (inlined at the bottom of this issue): sends one large stable system prompt (+10 tool schemas) with slightly varied short user queries, so every turn after the first is a prefix-cache hit that diverges near the end — the pattern any agentic CLI produces.

```bash
python lmcache-repro.py --rounds 1              # shared prefix  -> reproduces
python lmcache-repro.py --rounds 1 --unique     # control: unique prefix -> clean
```

Prompt size is set by `--guidelines` (default 120 ≈ 10.5k tokens). The numbers in the table below come from an equivalent in-house script whose prompt was Traditional Chinese at ~14.1k tokens; the version below is a cleaned-up English port with the same structure, verified to exercise the same path (first turn 17.7s → subsequent 4.1s on a cache hit). Increase `--guidelines` if a larger prompt is needed to trigger it on other hardware.

## Bisect

All arms use the identical prompt, sampling params and model:

| vLLM | LMCache | connector | prefix | garbage turns |
|---|---|---|---|---|
| 0.26.0 | 0.5.2 | **on** | shared | **6/6** |
| 0.26.0 | 0.5.3.dev35 (incl. #4206) | **on** | shared | **7/8** |
| 0.26.0 | 0.5.2 | **on** | **unique** | 0/4 |
| 0.26.0 | — | **off** | shared | 0/8 |
| 0.25.1 | 0.5.2.dev50 | **on** | shared, real external hits | 0/16 |

The 0.25.1 arm was verified to actually exercise the external path (`External prefix cache hit rate` 54–71%, `Prefetch request completed (L1+L2): 8/8 retained keys`, 0 MP-server errors) — not a silent no-op.

### Ruled out

- **MTP / speculative decoding** — clean in the connector-off arm with MTP still enabled.
- **vLLM's own prefix caching** — clean in the connector-off arm with `--enable-prefix-caching` still on and hits confirmed (first turn 18.6s → subsequent 6.1s).
- **Prompt length, tool schemas, sampling params** — identical across all arms.
- **Stale on-disk cache** — reproduced from an empty L2 directory.

The only variable that flips the outcome is whether an LMCache **hit** is possible.

## Where it seems to come from

`EngineKVFormat` detection differs between the two vLLM versions. On **0.26.0 + LMCache 0.5.2**, five KV groups are detected as two different formats:

```
Engine KV Format: EngineKVFormat.NL_X_NB_TWO_BS_NH_HS  NL x [NB, 2, BS, NH, HS]   (x4)
Engine KV Format: EngineKVFormat.NL_X_NB_NH_BS_TWO_HS  NL x [NB, NH, BS, 2, HS]   (x1)
```

The 4 format-2 groups are the `mamba-page-view` synthetic tensors (`KV cache group edits applied: {'mamba-page-view': 48}`); the format-10 group is real attention, which vLLM 0.26.0 + FlashInfer emits with **K/V fused into the trailing dim**.

On **0.25.1**, every group detects as format 2 and the bug cannot trigger.

Format 10 (`NL_X_NB_NH_BS_TWO_HS`) support is new in 0.5.2 (added by #4128 for #4111) — its transfer kernels treat the layout as HND with `kv_size == 1` and `hs == 2 * head_size`. That is exactly the code path absent from every release that works here, which points at the fused-K/V transfer path being wrong for this mixed-format, hybrid-Mamba topology.

Notes:
- **#4206 does not fix it.** On `dev` @ `927897db` the detection output changes (a second pass at `detection.py:69` now also emits format 11 `NL_X_NB_BS_NH_TWO_HS`), but the corruption remains at 7/8.
- Downgrading to 0.5.2.dev50 against vLLM 0.26.0 is **not** a workaround: every store fails with `RuntimeError: Unsupported EngineKVFormat: 10`, the exception is swallowed, and LMCache becomes a silent no-op (0 chunks stored, 0.0% external hit). Output looks clean for the wrong reason — worth checking `External prefix cache hit rate` before believing any "pass" here.

## Related

- #4111 / #4128 — added the format-10 kernels this report suspects
- #4160 — CacheBlend V3 on the sibling packed layout `NL_X_NB_BS_NH_TWO_HS` (format 11)
- #3388 — degenerate output via MP connector, but that one logs `result=False`; this one is completely silent
- #4155 — silent corruption via the *Python fallback* path; this one uses the native kernels

## Current workaround

Pin vLLM `0.25.1` + LMCache `0.5.2.dev50`.

---

<details>
<summary><code>lmcache-repro.py</code></summary>

```python
#!/usr/bin/env python3
"""Reproducer: LMCacheMPConnector silently corrupts output on shared-prefix
requests for a hybrid Mamba/GDN + full-attention model under vLLM 0.26.0.

Sends the same large system prompt (+ tool schemas) with slightly different
short user queries, so every turn after the first is a prefix-cache hit that
diverges near the end -- the pattern an agentic CLI produces.

  --unique   prepend a random nonce to the system prompt so no prefix-cache
             hit is possible. This is the control arm: it stays clean even
             with LMCache enabled.

Exit code 1 if any turn produced garbage.
"""

import argparse
import json
import re
import sys
import time
import urllib.request
import uuid

GUIDE = (
    "Always follow existing project conventions. Analyze surrounding code, "
    "tests, and configuration first. Never assume a library is available; "
    "verify it is already used in the project before relying on it. Mimic "
    "the style, structure, framework choice, typing, and architectural "
    "patterns of existing code. Prefer running the project's own lint and "
    "test commands over guessing."
)

TOOLS = [
    {
        "type": "function",
        "function": {
            "name": name,
            "description": desc,
            "parameters": {
                "type": "object",
                "properties": {
                    "query": {"type": "string", "description": "the query string"},
                    "path": {"type": "string", "description": "absolute path"},
                    "limit": {"type": "integer", "description": "max results"},
                },
                "required": ["query"],
            },
        },
    }
    for name, desc in [
        ("read_file", "Read and return the contents of a file at an absolute path."),
        ("write_file", "Write content to a file, creating parent dirs as needed."),
        ("list_directory", "List entries of a directory, sorted, with file types."),
        ("search_file_content", "Regex search across files, returning matches."),
        ("glob", "Find files matching a glob pattern, newest first."),
        ("run_shell_command", "Execute a shell command and capture its output."),
        ("web_fetch", "Fetch a URL and return readable text content."),
        ("web_search", "Search the web and return ranked results."),
        ("save_memory", "Persist a fact to long-term project memory."),
        ("todo_write", "Create or update the structured task list."),
    ]
]

QUERIES = [
    "What is the maximum character level in this game?",
    "What is the maximum character level in this game? Answer briefly.",
    "What's the max character level in the game?",
    "Tell me the maximum character level.",
    "Please state the maximum character level in this game.",
    "Could you tell me the max level in this game?",
    "Game max character level?",
    "What is the maximum character level? Number only.",
]

# Scripts that must never appear in an English answer -- the corruption is a
# multilingual token salad, so these are a reliable tell.
ALIEN = re.compile(r"[Ѐ-ӿ؀-ۿऀ-ॿ฀-๿가-힯�]")


def build_system(n_guidelines: int, unique: bool) -> str:
    head = f"[session {uuid.uuid4()}] " if unique else ""
    return (
        head
        + "You are an interactive CLI agent specializing in software "
        "engineering tasks.\n\n"
        + "\n".join(f"# Guideline {i}\n{GUIDE}" for i in range(n_guidelines))
    )


def ask(url: str, model: str, system: str, query: str, max_tokens: int):
    body = json.dumps(
        {
            "model": model,
            "messages": [
                {"role": "system", "content": system},
                {"role": "user", "content": query},
            ],
            "tools": TOOLS,
            "max_tokens": max_tokens,
            "temperature": 0.7,
            "top_p": 0.95,
            "stream": False,
        }
    ).encode()
    req = urllib.request.Request(
        url + "/v1/chat/completions", data=body,
        headers={"Content-Type": "application/json"},
    )
    t0 = time.time()
    with urllib.request.urlopen(req, timeout=1800) as r:
        d = json.load(r)
    el = time.time() - t0
    m = d["choices"][0]["message"]
    text = (m.get("reasoning") or m.get("reasoning_content") or "") + " " + (
        m.get("content") or ""
    )
    return text, d["usage"], el


def main() -> int:
    p = argparse.ArgumentParser()
    p.add_argument("--url", default="http://localhost:8000")
    p.add_argument("--model", default="Qwen3.6-27B")
    p.add_argument("--rounds", type=int, default=1)
    p.add_argument("--guidelines", type=int, default=120,
                   help="controls prompt size; 120 ~= 14k tokens")
    p.add_argument("--max-tokens", type=int, default=700)
    p.add_argument("--unique", action="store_true",
                   help="control arm: unique prefix, no cache hit possible")
    a = p.parse_args()

    bad = turn = 0
    for _ in range(a.rounds):
        for q in QUERIES:
            turn += 1
            system = build_system(a.guidelines, a.unique)
            try:
                text, usage, el = ask(a.url, a.model, system, q, a.max_tokens)
            except Exception as e:  # noqa: BLE001
                print(f"turn {turn:>3}  REQUEST FAILED: {type(e).__name__}: {e}")
                continue
            alien = len(ALIEN.findall(text))
            garbage = alien >= 5
            bad += garbage
            print(
                f"turn {turn:>3}  {'GARBAGE' if garbage else 'ok':<8} "
                f"alien={alien:<4} prompt={usage['prompt_tokens']:<6} "
                f"out={usage['completion_tokens']:<5} {el:5.1f}s",
                flush=True,
            )
            if garbage:
                print(f"    sample: {text[:240]!r}", flush=True)

    print(f"\n=== {bad}/{turn} turns produced garbage ===")
    return 1 if bad else 0


if __name__ == "__main__":
    sys.exit(main())
```

</details>

thegoldenflow:
Fixes #4247

`_SubpagedAttentionViewEdit` gates on `ndim == 5`. vLLM 0.26.0 registers the non-MLA KV cache as rank 4, so the edit silently stops firing and the attention group is left paged at the backend's kernel block size instead of the scheduler's. On a hybrid-Mamba model this corrupts store/retrieve with no exception and no log line. Confirmed fixed end-to-end on affected hardware — see Validation.

## Problem

vLLM 0.26.0 changed the registered non-MLA KV cache from a rank-5 K/V-split layout to a rank-4 K/V-fused layout. All paths under `vllm/v1/attention/backends/`:

| backend | 0.25.1 — `(num_blocks, 2, block_size, num_kv_heads, head_size)` | 0.26.0 — `(num_blocks, num_kv_heads, block_size, 2 * head_size)` |
| --- | --- | --- |
| `flash_attn` | `flash_attn.py:132` | `flash_attn.py:141` |
| `flashinfer` | `flashinfer.py:392` | `flashinfer.py:408` |
| `triton_attn` | `triton_attn.py:345` | `triton_attn.py:351` |
| `flex_attention` | `flex_attention.py:133` | `flex_attention.py:138` |

ROCm AITER FA moved the same way (`rocm_aiter_fa.py:767` → `775`). The change is **not** universal: `hpc_attn.py:293` is byte-identical in both tags and still rank 5, so both ranks stay reachable and both have to be handled rather than swapped. The block dim is index 2 in both layouts, so only the rank distinguishes them.

## Effect

On 0.26.0 the `ndim == 5` gate is unsatisfiable, so the edit never fires and the attention group keeps the backend's kernel block size.

This only bites where that differs from the scheduler block size — backends advertising fixed integer sizes, where `select_common_block_size` descends to the largest advertised divisor (FlashInfer `[16, 32, 64]`, ROCm AITER FA `[16, 32]`). Backends advertising `MultipleOf(16)`, such as `flash_attn`, resolve to the logical size directly, so `shape[2] == spec.block_size` and the edit correctly never needed to fire there.

## Why it is silent rather than loud

Two guards should have caught a kernel-paged group; neither can.

At `lmcache/v1/kv_layer_groups.py:403`, `tokens_per_block` falls back to `bs` — the physical slot count read off the registered tensor — whenever the engine group info is absent or non-positive, explicitly so the group is "treated as non-compressed". The group is then misread not as compressed but as *uncompressed at the wrong block size*, so the `slots_per_block != tokens_per_block` diagnostic at `:691` never fires and nothing downstream has anything left to disagree about. (Trace credit: @MikeWang0316tw — my original read assumed the compressed variant.) And where the info is present, every check in `_validate_block_chunk_size_config` is a pure modulo test: `select_common_block_size` only returns divisors (`vllm/v1/worker/utils.py:344`), so the ratio is integral by construction, and a guard rejecting only non-integer ratios can never reject a wrong-but-integer one.

The only observable signal was the applied-edits counter missing a key — `{'mamba-page-view': N}` with `subpaged-attention-view` absent rather than zero, since it increments only on a match.

## Fix

- `matches()` — accept rank 4 or rank 5, comparing `shape[2]` against `spec.block_size` in both.
- `apply()` — handle both ranks; validate the `shape[1] == 2` invariant only where it applies (rank 5).
- `_in_memory_order()` — new helper. vLLM registers the rank-4 tensor as a `permute` view of the backend's physical layout (`get_kv_cache_stride_order`), which under NHD is the transpose `(0, 2, 1, 3)`. It is not contiguous in logical dim order, and pages only tile by byte range in memory order, so the rank-4 branch re-views in descending-stride order.

**Why patch `matches()` rather than move the edit after normalization.** The comment at `lmcache/integration/vllm/lmcache_mp_connector.py:773` states the edit must precede both group-info creation and transfer registration so they see the same edited views. `apply_kv_cache_group_edits` rebinds `kv_caches` at `:776`, and both `create_engine_group_infos_from_vllm` (`:779`) and `worker_adapter.register_kv_caches` (`:784`) read the rebound local. Moving it later would build group infos from the un-repaged tensor — precisely the bug being fixed.

## Known limitations

1. **Rank 4 under HND still resolves `block_size` to 1.** The edit machinery emits an NHD-shaped synthetic view by construction (`num_heads` is a synthetic 1), while the HND detector branch selects `NL_X_NB_TWO_NH_BS_HS`, whose `block_size()` reads a different axis. Pre-existing and orthogonal to the rank gate — it affects every edited group, Mamba included — and not reached by default, since `get_required_kvcache_layout` returns `None` and vLLM falls back to NHD. Filing separately.

2. **Rank-5 behavior is deliberately unchanged, including its `is_contiguous()` guard.** Extending the memory-order normalization to rank 5 would let a permuted tensor (vLLM ≤ 0.25.x under HND) re-view successfully and then be misread downstream as `block_size == 1` — converting a loud `ValueError` into a silent wrong block size. That is the exact failure class this PR exists to remove.

**Rollout note:** after upgrading, clear legacy L2 objects. The object key carries no layout fingerprint and the filesystem backend does not overwrite an existing object on store, so stale entries written by the unpatched code would be served as-is.

## Validation

End-to-end confirmed by @MikeWang0316tw on affected hardware (GB10, FlashInfer, hybrid Mamba/GDN, `block_size=1600`). Stock vLLM `v0.26.0`, no local patches; two arms differing by this one file on the same LMCache build; greedy decoding verified bit-reproducible across two rounds.

| | patched | control |
| --- | --- | --- |
| `KV cache group edits applied` | `{'mamba-page-view': 48, 'subpaged-attention-view': 17}` | `{'mamba-page-view': 48}` |
| `Engine KV Format` | 5x `NL_X_NB_TWO_BS_NH_HS` | 4x `NL_X_NB_TWO_BS_NH_HS` + 1x `NL_X_NB_BS_NH_TWO_HS` |
| ran to `max_tokens` | 0/16 | 12/16 |
| flagged garbage | 0/16 | 2/16 |
| external prefix cache hit rate | 40–91% | 57–60% |

Both arms genuinely consumed external tokens, so this is not a no-op passing for the wrong reason.

The format row is the strongest signal and needs no generation: on 0.25.1 all five groups detect as format 2, and with this patch on 0.26.0 they do again. The fused format reaching the transfer layer is the bug's signature, not its cause — those kernels were being handed a group that should never have reached them unedited.

## Testing

`tests/v1/test_vllm_kv_cache_group_edits.py` — 19 tests, pure CPU, no GPU or vLLM install needed (vLLM and the spec-kind helpers are stubbed, and the module is re-imported under the stubs with a binding assert, so an installed vLLM cannot leak in; runs with or without vLLM present). Reverting the source fix turns 8 of them red with `assert 64 == 1600` and a log line reading `{'mamba-page-view': 4}`.

This module had no unit coverage before — testing was GPU end-to-end only, which pins faithful retrieve but cannot gate *which rules fire* for a given vLLM version. That is exactly why a rule silently not firing was undetectable in CI. The new tests pin the matching gate per registered layout and the resulting block granularity, not internal view shapes.

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/kv_cache_group_edits.py
+++ b/lmcache/integration/vllm/kv_cache_group_edits.py
@@ -64,6 +64,22 @@ RegisteredKVCache: TypeAlias = torch.Tensor | list[torch.Tensor]
 # to fill the page.
 _SYNTHETIC_NUM_HEADS = 1
 
+# Ranks of the two attention KV layouts vLLM registers, and the dim holding
+# the paging granularity in each. Through vLLM 0.25.x every non-MLA backend
+# split K and V into their own dim:
+#
+#     (num_blocks, 2, block_size, num_heads, head_size)          -- rank 5
+#
+# vLLM 0.26.0 packs them into the content dim instead (flash_attn.py,
+# flashinfer.py, triton_attn.py, flex_attention.py, rocm_aiter_fa.py):
+#
+#     (num_blocks, num_heads, block_size, 2 * head_size)         -- rank 4
+#
+# The block dim is index 2 in both, so only the rank distinguishes them.
+_SPLIT_KV_NDIM = 5
+_FUSED_KV_NDIM = 4
+_ATTENTION_BLOCK_DIM = 2
+
 # Standard-paged (non-MLA) attention kinds eligible for the sub-paged edit.
 _SUBPAGEABLE_ATTENTION_KINDS = frozenset(
     {
@@ -162,6 +178,30 @@ def _synthetic_attention_shape(elems_per_page: int, block_size: int) -> tuple[in
             f"(2, block_size={block_size}, num_heads={_SYNTHETIC_NUM_HEADS}, head_size)"
         )
     return _SYNTHETIC_NUM_HEADS, elems_per_page // denom
+
+
+def _in_memory_order(kv_cache: torch.Tensor) -> torch.Tensor:
+    """Return ``kv_cache`` permuted so its dims run in descending stride order.
+
+    vLLM allocates the KV tensor in its backend's preferred physical layout and
+    hands back a ``permute`` view of it (``get_kv_cache_stride_order`` /
+    ``vllm/v1/worker/gpu/attn_utils.py``), so the registered tensor's *logical*
+    dim order is generally not its memory order and it need not be contiguous.
+    Re-viewing pages by byte range is only meaningful in memory order.
+
+    Used for the rank-4 fused layout only; see :meth:`
+    _SubpagedAttentionViewEdit.apply` for why rank 5 keeps a stricter guard.
+
+    A no-op for an already-contiguous tensor, whose strides are descending.
+
+    Args:
+        kv_cache: Registered attention KV tensor.
+
+    Returns:
+        A view of ``kv_cache`` whose dims are ordered outermost-first in memory.
+    """
+    order = sorted(range(kv_cache.ndim), key=kv_cache.stride, reverse=True)
+    return kv_cache.permute(*order)
 
 
 class KVCacheGroupEdit(ABC):
@@ -267,11 +307,13 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
     """Re-view a kernel-paged attention tensor as logical-block pages.
 
     For a Mamba-hybrid model vLLM inflates the attention block size to align
-    with the Mamba page (e.g. 544 for Qwen3.5-0.8B), and that size is used for
-    all prefix-caching logic at the scheduler. But at the worker the attention
-    kernel has to run at block size 32 for numerical stability (vLLM #27753,
-    working around the NaN-propagation issue
-    Dao-AILab/flash-attention#1974), so the registered tensor is paged as
+    with the Mamba page (``vllm/platforms/interface.py:_align_hybrid_block_size``;
+    e.g. 544 for Qwen3.5-0.8B), and that size is used for all prefix-caching
+    logic at the scheduler. But the backend can only page the physical tensor
+    at a *kernel* block size it actually supports: ``select_common_block_size``
+    (``vllm/v1/worker/utils.py``) returns the largest advertised size dividing
+    the logical one, so a backend advertising fixed sizes (FlashInfer's
+    ``[16, 32, 64]``, ROCm AITER FA's ``[16, 32]``) re-pages the tensor as
 
         ``[#blocks, 2, 32, #heads, head_size]``
 
@@ -281,6 +323,14 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
     block size (one logical block = its 17 contiguous kernel pages),
 
         ``[#blocks / 17, 2, 544, 1, head_size']``
+
+    Backends advertising ``MultipleOf(16)`` (FlashAttention) page at the
+    logical size directly and never need the edit.
+
+    Both attention KV layouts are handled: the rank-5 K/V-split layout of
+    vLLM <= 0.25.x and the rank-4 fused layout of vLLM >= 0.26.0 (see
+    ``_SPLIT_KV_NDIM`` / ``_FUSED_KV_NDIM``). The re-view is a byte-range
+    reinterpretation, so it is indifferent to which one it is given.
 
     Cost: before this fix ``kv_caches[:, 0]`` is just the K tensor; after, it
     interleaves K and V at kernel-page granularity. The bytes round-trip
@@ -296,12 +346,13 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
             # compression (DeepSeek) belong to other transfer paths.
             get_kv_cache_spec_kind(spec) in _SUBPAGEABLE_ATTENTION_KINDS
             and not _declares_slot_compression(spec)
-            # (num_blocks, 2, block_size, num_heads, head_size) layout whose
-            # block dim disagrees with the scheduler block-id unit -- the
-            # backend re-paged the tensor at its kernel block size.
+            # A paged attention layout -- rank 5 (K/V split, vLLM <= 0.25.x) or
+            # rank 4 (K/V fused, vLLM >= 0.26.0) -- whose block dim disagrees
+            # with the scheduler block-id unit, i.e. the backend re-paged the
+            # tensor at its kernel block size.
             and isinstance(kv_cache, torch.Tensor)
-            and kv_cache.ndim == 5
-            and kv_cache.shape[2] != spec.block_size
+            and kv_cache.ndim in (_SPLIT_KV_NDIM, _FUSED_KV_NDIM)
+            and kv_cache.shape[_ATTENTION_BLOCK_DIM] != spec.block_size
         )
 
     def apply(
@@ -312,18 +363,28 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
     ) -> torch.Tensor:
         """Re-view ``kv_cache`` at logical-block granularity.
 
-        The tensor is kernel-paged as ``(num_kernel_pages, 2,
-        kernel_block_size, num_kv_heads, head_size)``; the result is
-        ``(num_logical_blocks, 2, spec.block_size, num_heads, head_size)``
-        over the same storage.
+        The tensor is kernel-paged, either as ``(num_kernel_pages, 2,
+        kernel_block_size, num_kv_heads, head_size)`` (vLLM <= 0.25.x) or as
+        ``(num_kernel_pages, num_kv_heads, kernel_block_size, 2 * head_size)``
+        (vLLM >= 0.26.0); the result is ``(num_logical_blocks, 2,
+        spec.block_size, num_heads, head_size)`` over the same storage.
+
+        A rank-4 tensor is re-viewed in memory order, since vLLM registers it
+        as a permute view of the backend's physical layout. A rank-5 tensor
+        must already be contiguous in logical order.
 
         Raises:
-            ValueError: If the layout is not the expected kernel-paged shape,
-                the sizes do not divide evenly, or the kernel pages of one
-                logical block do not tile its page bytes exactly (which would
-                indicate an undeclared packed layout that must not be edited).
+            ValueError: If the layout is not a recognized kernel-paged shape,
+                the sizes do not divide evenly, the pages are not contiguous
+                (in memory order for rank 4, logical order for rank 5), or the
+                kernel pages of one logical block do not tile its page bytes
+                exactly (which would indicate an undeclared packed layout that
+                must not be edited).
         """
-        if not isinstance(kv_cache, torch.Tensor) or kv_cache.shape[1] != 2:
+        if not isinstance(kv_cache, torch.Tensor) or kv_cache.ndim not in (
+            _SPLIT_KV_NDIM,
+            _FUSED_KV_NDIM,
+        ):
             got = (
                 tuple(kv_cache.shape)
                 if isinstance(kv_cache, torch.Tensor)
@@ -331,10 +392,16 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
             )
             raise ValueError(
                 f"expected a (num_blocks, 2, block_size, num_heads, head_size) "
+                f"or (num_blocks, num_heads, block_size, 2 * head_size) "
                 f"attention KV tensor, got {got}"
             )
+        if kv_cache.ndim == _SPLIT_KV_NDIM and kv_cache.shape[1] != 2:
+            raise ValueError(
+                f"expected a (num_blocks, 2, block_size, num_heads, head_size) "
+                f"attention KV tensor, got {tuple(kv_cache.shape)}"
+            )
         logical_block_size = spec.block_size
-        kernel_block_size = kv_cache.shape[2]
+        kernel_block_size = kv_cache.shape[_ATTENTION_BLOCK_DIM]
         if logical_block_size % kernel_block_size != 0:
             raise ValueError(
                 f"logical block size {logical_block_size} is not a multiple of "
@@ -354,18 +421,45 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
                 f"{ratio} kernel pages ({kernel_page_bytes * ratio} bytes) do "
                 f"not tile the logical page ({spec.page_size_bytes} bytes)"
             )
-        if not kv_cache.is_contiguous():
-            raise ValueError(
-                "kernel-paged attention KV tensor must be contiguous to "
-                "re-view as logical pages"
-            )
+        if kv_cache.ndim == _FUSED_KV_NDIM:
+            # A rank-4 tensor is a permute view of the backend's physical
+            # layout (under NHD, stride order ``(0, 2, 1, 3)``), so it is
+            # contiguous in memory order rather than logical order. Pages only
+            # tile by byte range in memory order.
+            ordered = _in_memory_order(kv_cache)
+            if not ordered.is_contiguous():
+                raise ValueError(
+                    "kernel-paged attention KV tensor must be contiguous in "
+                    "memory order to re-view as logical pages (shape "
+                    f"{tuple(kv_cache.shape)}, strides {tuple(kv_cache.stride())})"
+                )
+            if ordered.shape[0] != num_kernel_pages:
+                raise ValueError(
+                    f"expected a num-blocks-first KV cache layout; outermost "
+                    f"memory dim is {ordered.shape[0]}, not the kernel page "
+                    f"count {num_kernel_pages}"
+                )
+        else:
+            # Rank 5 deliberately keeps the stricter logical-order guard.
+            # Extending the rank-4 memory-order normalization to rank 5 would
+            # let a permuted tensor (vLLM <= 0.25.x under HND) re-view
+            # successfully and then be misread downstream: the edited view is
+            # NHD-shaped, so the HND detector branch reads its block size from
+            # the synthetic ``num_heads`` axis and resolves 1. Failing loudly
+            # here is strictly better than that silent corruption.
+            ordered = kv_cache
+            if not ordered.is_contiguous():
+                raise ValueError(
+                    "kernel-paged attention KV tensor must be contiguous to "
+                    "re-view as logical pages"
+                )
 
         num_blocks = num_kernel_pages // ratio
         elems_per_page = spec.page_size_bytes // kv_cache.element_size()
         num_heads, head_size = _synthetic_attention_shape(
             elems_per_page, logical_block_size
         )
-        return kv_cache.view(num_blocks, 2, logical_block_size, num_heads, head_size)
+        return ordered.view(num_blocks, 2, logical_block_size, num_heads, head_size)
 
 
 class _MambaUnifiedViewEdit(KVCacheGroupEdit):
--- a/lmcache/integration/vllm/kv_cache_group_edits.py
+++ b/lmcache/integration/vllm/kv_cache_group_edits.py
@@ -64,6 +64,22 @@ RegisteredKVCache: TypeAlias = torch.Tensor | list[torch.Tensor]
 # to fill the page.
 _SYNTHETIC_NUM_HEADS = 1
 
+# Ranks of the two attention KV layouts vLLM registers, and the dim holding
+# the paging granularity in each. Through vLLM 0.25.x every non-MLA backend
+# split K and V into their own dim:
+#
+#     (num_blocks, 2, block_size, num_heads, head_size)          -- rank 5
+#
+# vLLM 0.26.0 packs them into the content dim instead (flash_attn.py,
+# flashinfer.py, triton_attn.py, flex_attention.py, rocm_aiter_fa.py):
+#
+#     (num_blocks, num_heads, block_size, 2 * head_size)         -- rank 4
+#
+# The block dim is index 2 in both, so only the rank distinguishes them.
+_SPLIT_KV_NDIM = 5
+_FUSED_KV_NDIM = 4
+_ATTENTION_BLOCK_DIM = 2
+
 # Standard-paged (non-MLA) attention kinds eligible for the sub-paged edit.
 _SUBPAGEABLE_ATTENTION_KINDS = frozenset(
     {
@@ -162,6 +178,30 @@ def _synthetic_attention_shape(elems_per_page: int, block_size: int) -> tuple[in
             f"(2, block_size={block_size}, num_heads={_SYNTHETIC_NUM_HEADS}, head_size)"
         )
     return _SYNTHETIC_NUM_HEADS, elems_per_page // denom
+
+
+def _in_memory_order(kv_cache: torch.Tensor) -> torch.Tensor:
+    """Return ``kv_cache`` permuted so its dims run in descending stride order.
+
+    vLLM allocates the KV tensor in its backend's preferred physical layout and
+    hands back a ``permute`` view of it (``get_kv_cache_stride_order`` /
+    ``vllm/v1/worker/gpu/attn_utils.py``), so the registered tensor's *logical*
+    dim order is generally not its memory order and it need not be contiguous.
+    Re-viewing pages by byte range is only meaningful in memory order.
+
+    Used for the rank-4 fused layout only; see :meth:`
+    _SubpagedAttentionViewEdit.apply` for why rank 5 keeps a stricter guard.
+
+    A no-op for an already-contiguous tensor, whose strides are descending.
+
+    Args:
+        kv_cache: Registered attention KV tensor.
+
+    Returns:
+        A view of ``kv_cache`` whose dims are ordered outermost-first in memory.
+    """
+    order = sorted(range(kv_cache.ndim), key=kv_cache.stride, reverse=True)
+    return kv_cache.permute(*order)
 
 
 class KVCacheGroupEdit(ABC):
@@ -267,11 +307,13 @@ class _SubpagedAttentionViewEdit(KVCacheGroupEdit):
     """Re-view a kernel-paged attention tensor as logical-block pages.
 
     For a Mamba-hybrid model vLLM inflates the attention block size to align
-    with the Mamba page (e.g. 544 for Qwen3.5-0.8B), and that size is used for
-    all prefix-caching logic at the scheduler. But at the worker the attention
-    kernel has to run at block size 32 for numerical stability (vLLM #27753,
-    working around the NaN-propagation issue
-    Dao-AILab/flash-attention#1974), so the registered tensor is paged as
+    with the Mamba page (``vllm/platforms/interface.py:_align_hybrid_block_size``;
+    e.g. 544 for Qwen3.5-0.8B), and that size is used for all prefix-caching
+    logic at the scheduler. But the backend can only page the physical tensor
+    at a *kernel* block size it actually supports: ``select_common_block_size``
+    (``vllm/v1/worker/utils.py``) returns the largest advertised size dividing
+    the logical one, so a backend advertising fixed sizes (FlashInfer's
+    ``[16, 32, 64]``, ROCm AITER FA's ``[16, 32]``) re-pages the tensor as
 
         ``[#blocks, 2, 32, #heads, head_size]``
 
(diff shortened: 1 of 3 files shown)


ApostaC:
Hi @wsyjh8 , the correct behavior should trigger the `_MambaUnifiedViewEdit` which works for vLLM 0.26.0's unified 4-rank KV cache format

thegoldenflow:
Hi @ApostaC,  Thanks for the pointer. I checked this against both the rule predicates and the reporter's vLLM 0.26.0 registration path, and I think these are two different rank-4 layouts.

`_MambaUnifiedViewEdit` matches only when the spec kind is `MAMBA`. The rank-4 tensor addressed by this PR belongs to an attention spec in `_SUBPAGEABLE_ATTENTION_KINDS`, so it cannot match that rule. For the same reason, allowing rank 4 in `_SubpagedAttentionViewEdit` does not make the two rules overlap: their spec-kind predicates are disjoint.

In the reporter's vLLM 0.26.0 GPU-model-runner path, `_reshape_kv_cache()` registers attention layers as individual tensors, while a `MambaSpec` is still expanded into a `state_tensors` list. That is also consistent with the runtime counters:

```text
control:
{'mamba-page-view': 48}

patched:
{'mamba-page-view': 48, 'subpaged-attention-view': 17}
```

`mamba-unified-view` is absent in both arms. The patch leaves the 48 Mamba-layer entries unchanged and applies the subpaged edit to 17 attention-layer entries.

At the KV-cache-group level, that changes the one affected attention group's detected format:

```text
control:
4x NL_X_NB_TWO_BS_NH_HS
1x NL_X_NB_BS_NH_TWO_HS

patched:
5x NL_X_NB_TWO_BS_NH_HS
```

The patched result also matches the five-group format observed with vLLM 0.25.1.

There may be another vLLM configuration or runner where the Mamba cache itself is registered as a unified rank-4 tensor, in which case `_MambaUnifiedViewEdit` would be the correct rule. If that is the configuration you had in mind, I would appreciate a pointer to its registration path.

If the concern is instead that rank-4 attention should have its own edit class rather than extending `_SubpagedAttentionViewEdit`, I am happy to restructure it that way.

ApostaC:
I tried LMCache latest dev branch and vLLM 0.26.0, and did not see any need for `_SubpagedAttentionViewEdit`. 
The script I used:

```bash
vllm serve Qwen/Qwen3.6-27B \
  --port 8100 \
  --max-model-len auto \
  --enable-prefix-caching \
  --trust-remote-code \
  --enforce-eager \
  --no-enable-flashinfer-autotune \
  --moe-backend triton \
  --mamba-cache-mode align \
  --max-num-batched-tokens 1500 \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both","kv_connector_extra_config":{"lmcache.mp.port":6555}}'
```

```bash
lmcache server --chunk-size 784 --max-workers 4 --port 6555 --l1-size 50 --eviction-policy LRU --http-port 7070 
```

I send the same request twice, with vLLM restarted, and I don't see any garble when using this (I'm on H200).

thegoldenflow:
Hi @ApostaC , Thanks for running it. I think that result may be expected: on H200 (SM90), vLLM tries FlashAttention ahead of FlashInfer for ordinary causal attention (`vllm/platforms/cuda.py:148`), so your run likely selected FlashAttention — though the startup log would confirm.

FlashAttention advertises `MultipleOf(16)` (`flash_attn.py:81`), so when the KV-manager block size is already supported, `select_common_block_size()` returns it directly (`vllm/v1/worker/utils.py:290`). Kernel and scheduler block sizes are then equal, and `_SubpagedAttentionViewEdit` correctly does not match — there is no sub-paging to undo.

The reporter's configuration used FlashInfer, which advertises fixed kernel block sizes `[16, 32, 64]`. With a larger logical block, that backend can split one scheduler block into multiple kernel pages — the path addressed by this PR.

This startup line would confirm which case you're in:

```text
Using ... attention backend
```

If it says `FlashAttention`, the clean result is expected and this configuration does not exercise the path fixed here. If it says `FlashInfer`, the selected attention block size or the following log line would show whether sub-paging actually occurred:

```text
KV cache group edits applied: ...
```

These are initialization-time signals, so if the original log is gone, a `--load-format dummy` startup should be enough to collect them without running the full reproducer.

SebastianGode:
any news here? Would be cool to be able to use LMCache again on Qwen models

Lewis-panda:
I reproduced this against current `dev` (`f9addd2e`) and found an additional
NHD/HND boundary in the rank-4 fix.

The rank-4-to-rank-5 synthetic view in this PR repairs the vLLM-generated NHD
case, but the resulting view no longer preserves enough layout identity for
HND: LMCache detects block size `1` instead of the logical block size `1600`.

| variant | generalized CPU contract | vLLM-generated NHD / HND detected block size |
|---|---:|---:|
| current `dev` | 5 passed / 17 failed | 64 / 64 |
| #4253 cleanly merged onto current `dev` | 12 passed / 10 failed | 1600 / 1 |
| layout-aware candidate | 22 passed / 0 failed | 1600 / 1600 |

The generator row uses vLLM commit `754e1c3d`'s own
`_reshape_attention_kv_cache`; its NHD logical view is non-contiguous while HND
is contiguous. The candidate remains zero-copy and preserves physical byte
order in both layouts. The 22 cases also cover 1/2/32 heads, legacy-contiguous
and vLLM non-contiguous NHD views, invalid page/head/content metadata,
non-dense input, unsubpaged/compressed/non-hybrid pass-through, and the legacy
rank-5 path. Current adjacent LMCache selections remain 20/20 passing.

I prepared the current-`dev` implementation and tests as review input rather
than opening a duplicate PR:

https://github.com/Lewis-panda/LMCache/commit/5e11c48759655ad2fe3b12b818d9329b4a5439d8

The production change recovers dense physical dimension order from strides,
derives the kernel/logical-page ratio from the byte contract, then emits a
layout-aware rank-4 NHD or HND view. It also cross-checks available head and
content metadata so a plausible shape cannot silently mask an inconsistent
page contract. The #4253 test stub will need the current `MLA_ATTENTION` enum
member when rebased; that is separate test-fixture drift.

Two claim limits are important:

1. I have not run this candidate end-to-end on an affected model/GPU, so this
   is contract-level evidence, not production closure, and it does not claim
   to fix #4674.
2. Persistent L2 data must be invalidated on rollout. The current object key
   has no layout/schema fingerprint, and the filesystem backend does not
   overwrite an existing stale object during normal store. For closure, I
   suggest clearing legacy L2, populating with corrected code, restarting all
   processes, and then checking an L2-only hit against an exact expected
   answer.

Please feel free to cherry-pick/adapt the commit into this PR. If maintainers
prefer a separate superseding PR, I can open one explicitly linked here.

thegoldenflow:
> I reproduced this against current `dev` (`f9addd2e`) and found an additional NHD/HND boundary in the rank-4 fix.
> 
> The rank-4-to-rank-5 synthetic view in this PR repairs the vLLM-generated NHD case, but the resulting view no longer preserves enough layout identity for HND: LMCache detects block size `1` instead of the logical block size `1600`.
> 
> variant	generalized CPU contract	vLLM-generated NHD / HND detected block size
> current `dev`	5 passed / 17 failed	64 / 64
> #4253 cleanly merged onto current `dev`	12 passed / 10 failed	1600 / 1
> layout-aware candidate	22 passed / 0 failed	1600 / 1600
> The generator row uses vLLM commit `754e1c3d`'s own `_reshape_attention_kv_cache`; its NHD logical view is non-contiguous while HND is contiguous. The candidate remains zero-copy and preserves physical byte order in both layouts. The 22 cases also cover 1/2/32 heads, legacy-contiguous and vLLM non-contiguous NHD views, invalid page/head/content metadata, non-dense input, unsubpaged/compressed/non-hybrid pass-through, and the legacy rank-5 path. Current adjacent LMCache selections remain 20/20 passing.
> 
> I prepared the current-`dev` implementation and tests as review input rather than opening a duplicate PR:
> 
> [Lewis-panda@5e11c48](https://github.com/Lewis-panda/LMCache/commit/5e11c48759655ad2fe3b12b818d9329b4a5439d8)
> 
> The production change recovers dense physical dimension order from strides, derives the kernel/logical-page ratio from the byte contract, then emits a layout-aware rank-4 NHD or HND view. It also cross-checks available head and content metadata so a plausible shape cannot silently mask an inconsistent page contract. The #4253 test stub will need the current `MLA_ATTENTION` enum member when rebased; that is separate test-fixture drift.
> 
> Two claim limits are important:
> 
> 1. I have not run this candidate end-to-end on an affected model/GPU, so this
>    is contract-level evidence, not production closure, and it does not claim
>    to fix [[Bug] After #4253, LMCacheMP retrieve still near-corrupts on multi-session prefix hits (Qwen3.x hybrid + FlashInfer) #4674](https://github.com/LMCache/LMCache/issues/4674).
> 2. Persistent L2 data must be invalidated on rollout. The current object key
>    has no layout/schema fingerprint, and the filesystem backend does not
>    overwrite an existing stale object during normal store. For closure, I
>    suggest clearing legacy L2, populating with corrected code, restarting all
>    processes, and then checking an L2-only hit against an exact expected
>    answer.
> 
> Please feel free to cherry-pick/adapt the commit into this PR. If maintainers prefer a separate superseding PR, I can open one explicitly linked here.

Thanks, this reproduces a real limitation in the current rank-4 handling.

I agree this should be incorporated into #4253 rather than opened as a duplicate PR. I’ll rebase onto current dev, adapt 5e11c487, keep the legacy rank-5 path strict, and add layout-aware NHD/HND contract tests.

I’ll also keep the remaining MP retrieve issue in #4674 separate and update the rollout notes about invalidating legacy L2 data.

zhengfeihe:
Hi @thegoldenflow, thank you for the fix!  Sorry for leaving it unattended for a while.

 @ApostaC @KuntaiDu
 
Could you take another look at this fix? Without it, vLLM + LMCache +  **FlashInfer** still silently pollutes responses.
I make a regression test: https://github.com/LMCache/LMCache/pull/4954 for quick verification.

dmmdea:
Independent verification on a different platform, in case it helps this land: Qwen3.8-27B-INT4 (`RedHatAI/Qwen3.8-27B-INT4`, Qwen3.5-arch GDN hybrid), TP=2 on 2× RTX 5060 Ti 16 GB (sm_120), vLLM 0.28.0, `--kv-cache-dtype fp8` (which selects FlashInfer on this GPU), `--mamba-cache-mode align`, LMCache 0.5.4 MP server with `--separate-object-groups`, chunk 1568, L1 only (32 GB, no L2 adapter).

A planted-needle document (~23.7k tokens) answers correctly cold and from the VRAM prefix cache. After it is flooded out of VRAM and served back from the LMCache tier (15,680 external-hit tokens, TTFT ~1.0 s) the answer is garbage on stock 0.5.4 — every counter green. With this PR's `kv_cache_group_edits.py` change applied on top of 0.5.4 (nothing else changed, same arm, same session) the same eviction phase returns the exact needle with the same 15,680 external hits and the same TTFT:

| | stock 0.5.4 | + this PR |
|---|---|---|
| evict: external-hit tokens | 15,680 | 15,680 |
| evict: TTFT | 1.03 s | 1.04 s |
| evict: answer | `addOn!!!!…` ❌ | `7392-KESTREL` ✅ |

The fp16-KV seat (FLASH_ATTN) was never affected, matching the FlashInfer-only diagnosis in #4886. Would be glad to see this merged — it unblocks fp8 KV (and roughly 2× the KV pool) on hybrids for us.

zhengfeihe:
Hi~ @thegoldenflow 

Would you like to merge master into this branch to keep it up to date and fix failing unit tests (if any) ?

thegoldenflow:
Thanks @zhengfeihe — I’ll merge the latest dev into this branch and fix the failing unit tests.

thegoldenflow:
@zhengfeihe done — rebased onto current dev (3373a4a1) rather than merging, since there were no review comments to preserve (happy to switch to a merge commit if you prefer). Zero conflicts; `git range-diff` shows the production change unchanged apart from one context line. CI is green. Thanks for #4978 — the rebased branch should now merge cleanly into dev for the faithfulness test; tell me if it needs anything from my side.

Two fixes in the test file, none in production code:
1. The July red run was a defect in this PR's own test fixture. CI installs a real vLLM, and `test_mp_connector_layout_preference.py` imports the connector first, binding `kv_cache_group_edits` to it; the fixture's `sys.modules.pop()` followed by `from lmcache.integration.vllm import kv_cache_group_edits` then returned the stale module from the parent package attribute, so every test here silently ran against the wrong module (`KV cache group edits applied: none`). The fixture now re-imports via `importlib.import_module` under the stubs, restores state in `finally`, and asserts the module is stub-bound so a leak fails loudly. Reproduced the CI signature locally under the same pollution and confirmed the fix removes it.
2. Added the `MLA_ATTENTION` stub member @Lewis-panda predicted the rebase would need (read at match time by `_SubpagedMLAAttentionViewEdit`, 1dfa0777).

Evidence: 19/19 pass, alone and inside the full CPU suite; reverting only the production module still turns exactly 8 red with `assert 64 == 1600`; full CPU suite on HEAD vs dev has identical failure sets (+19 passes).

Scope: keeping this PR to the NHD path, which is what every end-to-end validation ran — @MikeWang0316tw (GB10, FlashInfer, vLLM 0.26.0), @dmmdea (2× RTX 5060 Ti, fp8 KV → FlashInfer, vLLM 0.28.0), plus @zhengfeihe's regression test #4954. @Lewis-panda's layout-aware HND handling (5e11c48) is a real improvement but has contract-level evidence only, and HND isn't reachable by default, so I'd rather land it as a follow-up than change the validated arm — Lewis-panda, happy for you to open that once this merges; I'll review. Rollout note added to the description: legacy L2 objects must be cleared, since the object key carries no layout fingerprint.

@ApostaC your H200 run selected FlashAttention (MultipleOf(16) → no sub-paging), so it can't hit this path. To reproduce: force FlashInfer with `--attention-backend FLASHINFER` and keep `--mamba-cache-mode align` so the attention block size lands above 64, then check two startup lines — `Using ... attention backend` must say FlashInfer, and `KV cache group edits applied:` will list `subpaged-attention-view` only on the patched arm.

@zhengfeihe could you also kick the two required Buildkite checks (k3-integration-test / k3-multiprocess-test)? They've shown "Expected" since July.

zhengfeihe:
@thegoldenflow 

Thanks for the prompt response ~ I enabled tests running

lmcache/integration/vllm/kv_cache_group_edits.py
```diff
@@ -164,6 +180,30 @@ def _synthetic_attention_shape(elems_per_page: int, block_size: int) -> tuple[in
...
+    A no-op for an already-contiguous tensor, whose strides are descending.
+
+    Args:
+        kv_cache: Registered attention KV tensor.
+
+    Returns:
+        A view of ``kv_cache`` whose dims are ordered outermost-first in memory.
+    """
+    order = sorted(range(kv_cache.ndim), key=kv_cache.stride, reverse=True)
+    return kv_cache.permute(*order)
```
deng451e: can we inline this function, as it only called once.

deng451e:
Thanks for the fix! I’ve verified that it works correctly with the FlashInfer attention backend. Please address the comment and resolve the merge conflict. Otherwise, LGTM!

lmcache/integration/vllm/kv_cache_group_edits.py
```diff
@@ -64,6 +64,22 @@
...
+# vLLM 0.26.0 packs them into the content dim instead (flash_attn.py,
+# flashinfer.py, triton_attn.py, flex_attention.py, rocm_aiter_fa.py):
+#
+#     (num_blocks, num_heads, block_size, 2 * head_size)         -- rank 4
+#
+# The block dim is index 2 in both, so only the rank distinguishes them.
+_SPLIT_KV_NDIM = 5
+_FUSED_KV_NDIM = 4
+_ATTENTION_BLOCK_DIM = 2
+
```
deng451e: Could we inline these three constants and move the layout history comments into the _SubpagedAttentionViewEdit class, also making them more concise?

thegoldenflow: Thanks for verifying it! Yes — I'll inline the three constants, move and trim the layout-history comments into _SubpagedAttentionViewEdit, then rebase onto the latest dev and resolve the merge conflict. I'll update the PR once it's clean.

deng451e:
Thanks for the fix! I’ve verified that it works correctly with the FlashInfer attention backend. Please address the comment and resolve the merge conflict. Otherwise, LGTM!

lmcache/integration/vllm/kv_cache_group_edits.py
thegoldenflow: Done in `8746accf` — inlined into the rank-4 branch of `apply()` and kept the memory-order rationale as a short comment at the call site. No behavior change.

lmcache/integration/vllm/kv_cache_group_edits.py
thegoldenflow: Done in `8746accf` — inlined the three constants and moved the trimmed layout note into the `_SubpagedAttentionViewEdit` docstring.

thegoldenflow:
@deng451e Both review comments are addressed in `8746accf`, and I rebased onto current `dev` to resolve the merge conflict. It's a pure refactor with no behavior change, so your FlashInfer verification still applies; the 19 unit tests pass. Thanks again for verifying the fix!

deng451e (approved):
LGTM!

lmcache/integration/vllm/kv_cache_group_edits.py
```diff
@@ -382,18 +408,46 @@ def apply(
...
-                "kernel-paged attention KV tensor must be contiguous to "
-                "re-view as logical pages"
-            )
+        if kv_cache.ndim == 4:
+            # vLLM registers the rank-4 tensor as a permute view of the
+            # backend's physical layout (get_kv_cache_stride_order), so it need
+            # not be contiguous. Pages only tile by byte range in memory order:
+            # re-view dims outermost-first by stride (a no-op if contiguous).
+            order = sorted(range(kv_cache.ndim), key=kv_cache.stride, reverse=True)
+            ordered = kv_cache.permute(*order)
```
ApostaC: In `lmcache/v1/gpu_connector/kv_format/contiguity.py`, we have a function `attempt_permute_to_contiguous_view`. Please use that.

ApostaC (changes requested):
Please have one quick fix, otherwise LGTM!

lmcache/integration/vllm/kv_cache_group_edits.py
thegoldenflow: Done in 5affe07a — switched to `attempt_permute_to_contiguous_view`. Kept the explicit `is_contiguous()` check afterwards, since the helper accepts dim-0-padded views and this path needs pages to tile exactly.

PR #4262, opened 2026-07-27
fix: NVFP4 TP deadlock workaround — remove redundant stream ctx, add timeout to future.result()

ZKeeer:
## NVFP4 TP Deadlock Workaround

Two small LMCache-side fixes for the `trtllm_fp4_block_scale_moe` kernel hang under interprocess CUDA event interference (observed on B200, TP=8, EP=8 with GLM-5.2-NVFP4).

### Fix 1: Remove redundant stream context manager

`lmcache_mp_connector.py`: `with torch_dev.stream(torch_dev.current_stream())` was a no-op (setting current stream to itself), causing unnecessary `torch._C._cuda_setStream()` driver calls on every forward step.

### Fix 2: Add timeout to `future.result()` calls

`vllm_multi_process_adapter.py`: `MessagingFuture.result()` was called with `timeout=None` (infinite wait). Adding 60s timeout converts silent hangs into recoverable errors within vLLM's 300s `execute_model` timeout, allowing recompute fallback via `kv_load_failure_policy: recompute`.

### Root Cause

Full analysis in the companion issue doc: `trtllm_fp4_block_scale_moe` kernel hangs when PDL (Programmatic Dependent Launch) interacts with interprocess CUDA events created by LMCache's connector. The FP8 counterpart (`trtllm_fp8_block_scale_moe`) does not exhibit this issue. A flashinfer-side fix (disabling PDL for FP4 MoE kernel) is proposed separately.


ApostaC (approved):
LGTM! Thanks for your contribution!

Shaoting-Feng (approved):

Issue #4263, opened 2026-07-27
[Bug] register_kv_caches crashes under vLLM 0.26.0 V2 model runner on hybrid Gemma-4 MoE: KV format detection reshape off by exactly 2x


Writtic:
## Environment

- **LMCache:** `0.5.2` (PyPI wheel). The image also carries a local backport of #4211 onto `lmcache/v1/distributed/l2_adapters/native_connector_l2_adapter.py`; it is unrelated to this report, the crash happens in `register_kv_caches` before any store runs.
- **vLLM:** `0.26.0` (official `vllm/vllm-openai:v0.26.0` image)
- **Model:** Gemma-4-26B-A4B NVFP4 (`--quantization modelopt`), MoE. Gemma-4 family hybrid attention (sliding-window + full-attention layer groups with different KV shapes). Draft-model speculative decoding (`num_speculative_tokens=4`).
- **Connector:** `LMCacheMPConnector`, `kv_role=kv_both`, `kv_connector_module_path=lmcache.integration.vllm.lmcache_mp_connector`
- **Parallelism:** TP=1 (uniproc executor)
- **`VLLM_USE_V2_MODEL_RUNNER=1`** (this MoE model requires the V2 model runner; the default runner mis-serves it on vLLM >= 0.25.1)

<details>
<summary>Launch config (flags unrelated to the crash trimmed; local paths redacted)</summary>

```bash
# MP server sidecar; L1/L2 tiering flags omitted, the crash happens before any store
lmcache server --host 0.0.0.0 --port 5555

vllm serve <Gemma-4-26B-A4B-NVFP4 checkpoint> \
  --tensor-parallel-size 1 \
  --max-model-len 65536 \
  --enable-prefix-caching \
  --quantization modelopt \
  --speculative-config '{"model": "<draft checkpoint>", "num_speculative_tokens": 4}' \
  --kv-transfer-config '{"kv_connector": "LMCacheMPConnector", "kv_connector_module_path": "lmcache.integration.vllm.lmcache_mp_connector", "kv_role": "kv_both", "kv_connector_extra_config": {"lmcache.mp.host": "tcp://localhost", "lmcache.mp.port": 5555, "lmcache.mp.mq_timeout": 60.0}}'
```

</details>

The v0.5.2 release notes say:

> ⚠️ Requires LMCache v0.5.2 for vLLM ≥ 0.26.0 — upgrade to this release to run against vLLM 0.26.0 and later.

[...]

PR #4220, opened 2026-07-23
[Add] New engine KV format for vLLM unified KV cache layout

ApostaC:
**What this PR does / why we need it**:

- Introduce two new engine KV formats: `NL_X_NB_NH_BS_CS` (HND) and `NL_X_NB_BS_NH_CS` (NHD)
- Deprecated old engine KV format: `NL_X_NB_NH_BS_TWO_HS` and `NL_X_NB_BS_NH_TWO_HS`

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #4265, opened 2026-07-27
[Test] Add regression test for mixed-shape rank-4 fused KV detection

Writtic:

Refs #4263.

Adds a regression test for the startup crash reported in #4263: blocks-first fused rank-4 KV cache groups whose trailing dims differ (hybrid sliding + full attention under the vLLM 0.26 V2 model runner). v0.5.2's whole-set detection applied the first tensor's `fused_dim` to every layer and died in reshape; #4206 (merged after v0.5.2 was cut) routes mixed-shape sets to per-shape buckets.

`test_normalize_per_layer_formats.py` covers rank-5 + rank-3 mixes but not mixed rank-4 fused, so the #4263 shape pattern had no coverage. The new test locks in the per-bucket split: each group keeps its own `fused_dim` and every layer reports `NL_X_NB_BS_NH_TWO_HS`.

Verified: passes on `dev` (20e87af); the same input reproduces the v0.5.2 crash (`RuntimeError: shape ... is invalid for input of size ...` in `detectors/vllm.py`) when run against that tag.

Test-only change.

**If applicable**:


ApostaC (approved):

ApostaC:
@Writtic Can you please fix the DCO by `git commit --amend --signoff` then do a force push to your branch?

Writtic:
@ApostaC Thanks for the review!

If you have the bandwidth, could you also take a look at https://github.com/LMCache/LMCache/pull/4211? My understanding is that at least two approvals are required.

Shaoting-Feng (approved):

Writtic:
@ApostaC @Shaoting-Feng After this PR was opened, #4220 was merged, which caused the tests to fail. I’ve fixed the issue, so could you please take another look?

Shaoting-Feng:
> @Shaoting-Feng After this PR was opened, #4220 was merged, which caused the tests to fail. I’ve fixed the issue, so could you please take another look?

LGTM. Enabled auto-merge.

Writtic:
@Shaoting-Feng Hmm, the **buildkite/k3-multiprocess-test** CI failed, but judging from the RuntimeError below, it doesn't look related to this change. Leaving a note here for the record.

```
RuntimeError: Insufficient space in /dev/shm: 160 MiB required, 64 MiB free.
```

zhengfeihe:
Hi ~ @Writtic 

I think issue shall be resolved by this one https://github.com/LMCache/LMCache/pull/4309

I triggered again, shall be good this time (ideally

PR #4271, opened 2026-07-27
[Fix][NIXL] Fail store tasks atomically on pool exhaustion

2xdevv:

The static NIXL L2 adapter previously stopped preparing a store batch when its storage pool could not allocate enough slots. If no objects had been prepared, the task incorrectly succeeded with zero transferred bytes. If earlier objects had already received slots, the adapter transferred and published that partial batch while reporting the entire task as successful.

This change treats pool exhaustion as a task failure before any transfer occurs.

- Route empty storage allocations through the existing failure path.
- Free every provisional storage slot accumulated for the task.
- Return an unsuccessful result with zero transferred bytes.
- Avoid publishing keys, notifying listeners, or updating usage for failed batches.
- Preserve zero-byte success when every submitted key already exists.
- Add regression coverage for oversized batches, already-full pools, and existing-key-only stores.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
@@ -713,9 +713,9 @@ class NixlStoreL2Adapter(L2AdapterInterface):
 
         For each key-object pair, memory page indices are mapped to storage
         slot indices and a single batched DMA write is issued. On success the
-        key-to-storage mapping is recorded in ``_memory_objects``. On transfer
-        failure, all allocated storage slots are freed and the task is marked
-        as failed.
+        key-to-storage mapping is recorded in ``_memory_objects``. On preparation
+        or transfer failure, all allocated storage slots are freed and the
+        task is marked as failed.
 
         Args:
             keys: Keys identifying each object to store.
@@ -744,8 +744,8 @@ class NixlStoreL2Adapter(L2AdapterInterface):
                     num_objs=len(mem_indices)
                 )
 
-                if storage_indices == []:
-                    break
+                if not storage_indices:
+                    raise RuntimeError("Insufficient NIXL storage capacity")
 
                 mem_indices_flat.extend(mem_indices)
                 storage_indices_flat.extend(storage_indices)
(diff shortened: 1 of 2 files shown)


DongDongJu (approved):
LGTM. This fixes a real correctness issue where pool exhaustion could be reported as a successful zero-byte or partial store.

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
```diff
@@ -748,8 +744,8 @@ async def _execute_store_in_the_loop(
                     num_objs=len(mem_indices)
                 )
 
-                if storage_indices == []:
-                    break
+                if not storage_indices:
+                    raise RuntimeError("Insufficient NIXL storage capacity")
```
zhengfeihe, line 748: I'm a little concerned that logging the exception with a full traceback might be too heavy here. We'd likely see the same error (with an identical traceback) for every failed store, and the store controller already logs a warning for each failed task. 

How about mirroring DynamicNixlStoreL2Adapter instead: set success = False, and log a single warning, keeping logger.exception for real transfer errors?

zhengfeihe (commented):
Sorry for being late, otherwise lgtm

lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
2xdevv: Thanks for the suggestion, @zhengfeihe. The dynamic adapter keeps objects stored before capacity exhaustion. Here, setting `success = False` and breaking would similarly allow the already-prepared subset to be transferred and published, while reporting the task as failed.

Most other adapters also retain successful writes on partial failure. However, the static adapter can roll back capacity exhaustion easily: it allocates slots before starting the transfer, so rollback only needs to free the provisional slots.

Did you intend to adopt the partial-store behavior too, or preserve this PR’s atomic rollback? For the latter, I could catch a dedicated capacity-exhaustion exception and log a warning, keeping `logger.exception` for unexpected errors.

zhengfeihe: Sorry for the late reply. Thanks for the detailed response. 

I think atomic rollback is fine enough. A warning logs is what I want to add (. Sorry for blocking for the small nits

2xdevv: No worries, thanks for clarifying, @zhengfeihe. Pool exhaustion now logs a warning without a traceback, while preserving atomic rollback. Unexpected errors still use logger.exception.

zhengfeihe (approved):

PR #4273, opened 2026-07-27
[ROCm] Build and publish prebuilt gfx942/gfx950 wheel

andyluo7:
## Summary

Adds a **release path for the ROCm/HIP wheel** so AMD Instinct users can `pip install` LMCache into an upstream `vllm/vllm-openai-rocm` container instead of building from source — parity with the CUDA cu129 flow.

Today the only ROCm option is a from-source build (`BUILD_WITH_HIP=1`); CUDA users get prebuilt wheels. This closes that gap for **gfx942** (MI300X/MI325X) and **gfx950** (MI350X/MI355X).

## How it works

- Built in `rocm/dev-ubuntu-22.04:7.2.3-complete` against the **public `torch==2.11.0+rocm7.2`**, so the extension's C++ ABI matches the torch shipped in `vllm/vllm-openai-rocm` (torch 2.11, ROCm 7.2, cp312).
- `PYTORCH_ROCM_ARCH=gfx942;gfx950` bakes both archs into one fat binary (torch's `cpp_extension` injects `--offload-arch` for the hipified `.hip` sources).
- `auditwheel` excludes torch **and every ROCm/driver-coupled runtime lib**, globbing versioned SONAMEs so `libamdhip64.so.7` / `libhsa-runtime64.so.1` are **not** vendored into the wheel (an unversioned `libamdhip64.so` exclude silently misses them and couples the wheel to the build host's KFD driver). Result: a lean **7.5 MB** wheel that binds to the host image at runtime — same policy as the cu129 wheel excluding `libcudart`.
- Published to a dedicated `v{tag}-rocm` **GitHub Release** (not PyPI), mirroring cu129, so the default `pip install lmcache` stays clean.

## Changes

- `.github/workflows/build_rocm_artifacts.yml` — reusable build workflow + optional on-device smoke-test job (gated on a self-hosted AMD runner via `vars.HAS_AMD_GPU_RUNNER`).
- `.github/scripts/build_rocm_wheel.sh` — container build + `auditwheel repair`.
- `.github/scripts/smoke_test_rocm_wheel.sh` — install into the vLLM ROCm image, load `c_ops`, run the real KV-transfer kernels on-device.
- `.github/workflows/publish.yml` — `build-rocm` job + `publish-rocm-github-release`.
- `docs/.../installation.rst` — prebuilt-wheel install tab for the upstream vLLM ROCm image; from-source ROCm index bumped `rocm7.0 → rocm7.2`.

## Validation (on real MI350X / gfx950)

The dev-ubuntu wheel built against **public** torch was installed into `vllm/vllm-openai-rocm:v0.25.0` (fork torch `2.11.0+gitd0c8b1f`) and:

- ✅ `import lmcache.c_ops` loads — proves cross-torch ABI compatibility (public 2.11 ↔ fork 2.11).
- ✅ HIP runtime device query round-trips to the GPU.
- ✅ **56/56 `tests/v1/test_mem_kernels.py`** cases pass on-device (real KV-transfer kernel execution).
- ✅ `llvm-objdump --offloading` confirms both `gfx942` and `gfx950` code objects in the shipped `.so`.

gfx942 code objects are compiled-in and verified present; on-device execution was validated on gfx950 (MI350X hardware).

## Test plan

- [x] End-to-end run of the exact committed `build_rocm_wheel.sh` → lean cp312 `manylinux_2_35_x86_64` wheel with gfx942+gfx950.
- [x] End-to-end run of the exact committed `smoke_test_rocm_wheel.sh` in `vllm/vllm-openai-rocm:v0.25.0` → all kernel tests pass.
- [ ] Maintainers: register a self-hosted AMD Instinct runner (label `amd-gpu`, set repo var `HAS_AMD_GPU_RUNNER=true`) to enable the on-device smoke-test gate in CI. The build job itself needs no GPU.


andyluo7:
### On-device validation now covers **both** target archs

Ran the exact committed `smoke_test_rocm_wheel.sh` (build once on `rocm/dev-ubuntu-22.04:7.2.3-complete` against public `torch==2.11.0+rocm7.2`, then install the same lean wheel into `vllm/vllm-openai-rocm:v0.25.0`) on real hardware:

| GPU | Arch | Result |
|-----|------|--------|
| MI350X | gfx950 | `tests/v1/test_mem_kernels.py` — **56 passed** |
| MI300X | gfx942 | `tests/v1/test_mem_kernels.py` — **56 passed** |

Both load the wheel's *compiled* `c_ops` (not the torch-baseline fallback) and execute the real KV-transfer kernels on-device against the image's fork torch `2.11.0+gitd0c8b1f`.

The second commit hardens the smoke test after I found it could silently pass against the torch baseline (source tree shadowing the installed wheel) — it now asserts the compiled extension actually loaded and avoids a double-import segfault.

.github/actionlint.yaml
ApostaC, line 1: Are we going to add a self-hosted runner to LMCache?

.github/workflows/build_rocm_artifacts.yml
```diff
@@ -0,0 +1,130 @@
...
+    workflow_call:
+
+env:
+    LC_ALL: en_US.UTF-8
+    # Pinned to match the upstream vllm/vllm-openai-rocm image.
+    ROCM_IMAGE: rocm/dev-ubuntu-22.04:7.2.3-complete
+    VLLM_ROCM_IMAGE: vllm/vllm-openai-rocm:v0.25.0
+    TORCH_ROCM_SPEC: torch==2.11.0+rocm7.2
+    TORCH_ROCM_INDEX: https://download.pytorch.org/whl/rocm7.2
+    PYTORCH_ROCM_ARCH: gfx942;gfx950
```
ApostaC, line 31: IIUC, the current version is hard-coded. What if vLLM & LMCache (and also PyTorch) update their versions? What if later versions of LMCache become incompatible with vLLM v0.25?

docs/source/getting_started/installation.rst
```diff
@@ -60,6 +60,33 @@ Install LMCache
...
+                        Install directly inside an upstream vLLM ROCm container — torch and the
+                        ROCm runtime are already present, so ``--no-deps`` binds against them:
+
+                        .. code-block:: bash
+
+                            docker run -it --device /dev/kfd --device /dev/dri \
+                                --group-add video --security-opt seccomp=unconfined \
+                                --entrypoint bash vllm/vllm-openai-rocm:v0.25.0
+
+                            VERSION=0.4.3  # replace with target release
```
ApostaC: Change it to at least 0.5.3 since the next release will be 0.5.3

ApostaC (commented):
Please take a look at my comments, thanks!

.github/actionlint.yaml
andyluo7: Good call — no, LMCache shouldn't take on a self-hosted AMD runner for this. I've **removed the self-hosted `smoke-test-rocm-image` job and this `.github/actionlint.yaml`** (it only existed to declare the `amd-gpu` label). The remaining `build-rocm-artifacts` job runs on `ubuntu-latest` (compiling HIP needs no GPU). On-device validation is run **manually** on AMD hardware via `.github/scripts/smoke_test_rocm_wheel.sh` — gfx942 + gfx950 results are in the PR description.

.github/workflows/build_rocm_artifacts.yml
andyluo7: Fair concern. The pins are consolidated as env vars at the top of this workflow as the single update point, and I've expanded the comment to explain the model:

The wheel links torch and the ROCm runtime as **external, auditwheel-excluded** libraries (bound at load time), so compatibility is tied to the **torch minor version (2.11)** — *not* to a specific vLLM/SGLang image tag. It loads in any ROCm image shipping torch 2.11 (cp3x), which today includes both `vllm/vllm-openai-rocm` and `lmsysorg/sglang` ROCm images. `v0.25.0` was only the manual smoke-test target, and that reference is now gone with the self-hosted job.

This is the same versioning model the repo already uses for the CUDA wheels (cu129/cu130 each pin a CUDA+torch pair; the main wheel's own docstring notes "building from source will be required if a serving engine uses an earlier non-ABI-compatible version of torch"). When the upstream ROCm images move to a new torch/ROCm minor, bump `ROCM_IMAGE` + `TORCH_ROCM_SPEC` + `TORCH_ROCM_INDEX` together here. LMCache-vs-vLLM app-level compatibility is orthogonal — it's governed by LMCache's own vLLM version support, same as on CUDA.

docs/source/getting_started/installation.rst
andyluo7: Done — bumped the ROCm install example to `VERSION=0.5.3`.

ApostaC (commented):
Thanks for the update. Let's drop `smoke_test_rocm_wheel.sh` if it's not needed?

andyluo7:
Done — dropped `.github/scripts/smoke_test_rocm_wheel.sh` (and its `publish.yml` path-filter entry) in 0af62796. It was only used by the removed self-hosted job; the manual on-device validation steps now live inline in `build_rocm_artifacts.yml`. Thanks for the review!

ApostaC (approved):
LGTM!

deng451e (approved):
LGTM!

PR #4235, opened 2026-07-24
[Serde] AES-GCM 

aeon-x:
## Summary

Adds an `aesgcm` serde that encrypts KV cache bytes **at rest in the L2 tier**
(S3 / `fs` / RESP / …), keyed per `cache_salt` so each tenant's ciphertext is
distinct. In multi-tenant deployments the shared L2 backend holds many tenants'
KV; this protects it from anyone who can read the remote storage.

It plugs into any L2 adapter through the existing `SerdeL2AdapterWrapper` — no
adapter or controller changes. This builds on the `ObjectKey`-threading serde
interface change (already merged); it's the first serde that uses the per-key
context.

Part of  #4127

## What it does

- **AES-128-GCM** by default (AEAD — confidentiality + integrity), random 96-bit
  IV per chunk, wire frame `[1B version][12B IV][ciphertext‖16B tag]`. A tag
  mismatch (tampering or wrong key) raises `InvalidTag`, which the wrapper turns
  into a load miss — never silent corruption. `aes_bits: 256` is a config knob
  for compliance mandates.
- **`KeyProvider`** abstraction with **`HkdfKeyProvider`** (in `key_provider.py`):
  derives a per-`cache_salt` key from one master key (read from
  `master_key_path`, e.g. a mounted K8s Secret) via HKDF-SHA256, cached per salt.
- Content is treated as an opaque byte blob (exact round-trip). Encryption runs
  off the inference hot path (L1↔L2 durable, not L0↔L1 CUDA-IPC), so its cost
  hides behind storage latency.

Adds `cryptography` to runtime deps (`requirements/common.txt`).

## Scope / threat model (intentionally narrow)

- **In scope:** L2 bytes vs. a reader of the remote storage.
- **Out of scope:** L1 (host RAM) and L0 (GPU) hold plaintext — no defense
  against access to a running server process.
- The `hkdf` provider is **"fleet vs. outside"** trust: any server with the
  master key can decrypt any tenant's data. True per-tenant isolation
  (`KeyringKeyProvider` + KMS) and object-name **metadata** hardening
  (`cache_salt`/`chunk_hash` are still visible in L2 object names) are
  documented as deferred follow-ups in the design doc.

## Config

```json
{"type": "s3", "bucket": "kv",
 "serde": {"type": "aesgcm", "key_provider": "hkdf",
           "master_key_path": "/etc/lmcache/keys/master"}}
```

## Docs

- Design doc: `docs/design/v1/distributed/serde/aesgcm.md` (threat model,
[...]

PR #4274, opened 2026-07-27
[Operator] Support L2 serde with aesgcm 

aeon-x:
## Summary

Exposes the `aesgcm` L2-at-rest encryption serde (#4235) on the operator's
CRD-managed path. Today encryption is only reachable through the raw
`--l2-adapter` JSON plus a hand-wired Secret mount; this adds a typed
`spec.l2Backend.serde` field that renders the serde config and delivers
the master key to the engine pods declaratively.

The field is modeled as a serde union (exactly one type set) rather than an
encryption-specific block, so other server-side serde types (fp8,
turboquant, …) can gain typed CRD fields later without an API break. Follows
the same patterns as the existing `resp`/`raw` union and RESP `authSecretRef`
handling.

Part of #4127

## What it does

**CRD field** (`LMCacheEngine` and `CacheBlendEngine` — shared `L2BackendSpec`):

```yaml
spec:
  l2Backend:
    raw:                     # or resp: — the serde applies to either
      type: fs
      config: {base_path: /data/lmcache/l2}
    serde:
      aesgcm:                # exactly one serde type must be set
        masterKeySecretRef:
          name: lmcache-l2-master-key   # user-created; data key "master";
                                        # same-namespace only
        keyProvider: hkdf               # default (only implemented provider)
        aesBits: 128                    # default; 256 for compliance
```

**Render** — `buildL2AdapterJSON` attaches the serde sub-dict to whichever
adapter is configured:

```json
{"type": "fs", "base_path": "...",
 "serde": {"type": "aesgcm", "key_provider": "hkdf",
           "master_key_path": "/etc/lmcache/keys/master", "aes_bits": 128}}
```

**Secret contract**: the **user creates** the master-key Secret **in the
engine's namespace** and references it by name (`corev1.LocalObjectReference`
— cross-namespace is impossible by construction, so CR-create permission
cannot be leveraged to read Secrets from other namespaces; see review
discussion). The DaemonSet **mounts the Secret directly** — read-only at
`/etc/lmcache/keys/master`, mode 0400, projecting only the `master` data key
via `items`. The operator never reads or writes Secrets for this feature and
**never generates key material** — provenance and rotation stay with the
user/security team; central-namespace key stores can be projected in via
external-secrets/replicator-style tooling. A missing Secret or missing
`master` key surfaces as a pod mount event and self-heals once fixed;
repointing the ref rolls the DaemonSet, so a key change actually takes
effect (the server reads the key once at startup).

**Validation**: exactly one serde type must be set; `masterKeySecretRef.name`
required; combining `serde` with a raw adapter that already sets a `serde`
config key is rejected (one would silently overwrite the other).

## Scope notes

- Applies to both engines: CacheBlendEngine projects onto the shared spec and
  reuses `buildDaemonSetCore`, so args + mount flow through unchanged.
- The `keyring`/KMS per-tenant provider stays deferred (the CRD enum admits
  only `hkdf`), matching the serde side.
- Threat model is unchanged from #4235: at-rest protection for the L2 tier
  ("fleet vs. outside" trust with hkdf); see
  `docs/design/v1/distributed/serde/aesgcm.md`.

## Tests

- `api/v1alpha1`: validation (valid spec, empty serde union, missing secret
  name, raw+serde conflict).
- `internal/resources`: serde sub-dict emission for resp and raw adapters
  (defaults + `aesBits: 256` override), no serde sub-dict when none is
  configured, DaemonSet master-key volume/mount (references the user's
  Secret directly, projects only the `master` item, 0400, read-only), no
  mount without an aesgcm serde.

`make test` (full envtest suite), `make lint` (0 issues), `make manifests
generate` (CRDs + deepcopy regenerated) all clean.

## Docs

- User guide: `docs/source/mp/operator.rst` — serde/aesgcm rows in the CRD
  spec reference, validation rules, and an "Encrypted L2 Backend" example.
- `operator/DESIGN.md`: spec walkthrough.
- Sample CR: `operator/config/samples/lmcache_v1alpha1_lmcacheengine_l2_encrypted.yaml`
  (includes the `kubectl create secret` one-liner).


operator/internal/controller/reconcile_helpers.go
```diff
@@ -506,3 +506,91 @@ func (r *LMCacheEngineReconciler) deleteRESPAuthSecretIfExists(ctx context.Conte
...
+			sourceNS, ref.Name, resources.L2EncryptionKeyDataKey)
+	}
+
+	desired := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      localName,
+			Namespace: engine.Namespace,
+			Labels:    resources.StandardLabels(engine.Name),
+		},
+		Data: map[string][]byte{
```
DongDongJu: A user allowed to create an engine CR in namespace A can ask the operator to copy data.master from namespace B. 
This makes the operator a confused deputy.
Same-namespace-only would be the simplest safe behavior.

DongDongJu (commented):
Hello @aeon-x, Thanks for the contribution.
I left one comment.

operator/internal/controller/reconcile_helpers.go
aeon-x: Updated

ApostaC (commented):
Can we update the user-facing docs as well? Thanks! Otherwise LGTM!

ApostaC (approved):
LGTM!

KuntaiDu (approved):
lgtm

Issue #4226, opened 2026-07-24
[RFC] Extending MP coordinator as the control plane for distributed KV Cache management.


ruizhang0101:
## Summary

Extending the MP coordinator to LMCache's distributed KV-cache control plane, including two components:
- **metadata**: a key directory (key → placements across nodes and tiers), some indexes for efficient lookup. All of them will be built from events emitted by MP servers  and only guarantee **eventual consistent**. 
- **management API**: instance registry/heartbeat, quota, cache control (delete/move/prefetch), pins etc. 

Durable state are backed using **etcd** with periodical **snapshot**. 

Adding a new light L1 coordination interface for handling multi-l1 media(native DRAM, CXL, GDS, Device-DAX etc), which will take care of the memory allocation and eviction. The media should bring their own L1 coordinator (CXL coordinator, GDS coordinator, etc.) 

## Motivation

More and more use cases require a centralized service that has a “global view” of all the objects and memory allocation status, including (but not limited to) the following: 
- Manage objects in multi-tier/share L1 (CXL, GDS, etc) 
- Manage objects in a shared L2
- External management in a distributed KV cache setup (delete, move, pin, prefetch etc.) 
- Quota management for in different storage tier 
- Segmented matching (non-prefix matching)
- KV cache discovery and configuration persistence
- Intelligent KV cache load and store planning

The coordinator we have today is a simple python-based FastAPI server with a L2-only view doing:
- External management like KV cache deletion, pin/unpin, prefetch etc.
- Manage objects in a shared L2
- L2 quota management
- Cacheblend non-prefix matching
- Discover KV cache in L2. 

However, the current version fails to: 
- Handle multi-tier/share L1 like CXL and GDS because it doesn't have L1 KV cache view. 
- Have a good performance under load
- Have consistency across the crash of the system components. 
- Have intelligent KV cache load and store planning. 

Extending the current MP coordinator will unlock these things. 

## System Overview

<img width="1274" height="638" alt="Image" src="https://github.com/user-attachments/assets/871bcf4f-4678-4a84-b346-d09fd9ea1b1c" />

Ownership model:

| owner                                                     | owns                                                                                                                             | consistency                    |
| --------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------------------------ |
[...]

PR #4275, opened 2026-07-28
[MP][Coordinator] Fleet-wide key directory built from cache events

ruizhang0101:

Refs #4226 (M1). Adds the key directory to the MP coordinator: an eventually consistent, in-memory map from each `ObjectKey` to its placements across the fleet (instance, tier, backend, size), built purely from `CacheEvent` batches emitted by MP servers.

- **Event ingestion** (`POST /directory/events`): per-instance FIFO by `seq` with duplicate dropping and gap detection, and incarnation fencing (a restarted instance's first batch atomically drops every placement its previous incarnation reported).
- **Placement lookup** by key (`POST /directory/lookup`) and by token sequence (`POST /directory/lookup_tokens`, prefix-exact via the fleet `TokenHasher` + per-rank fan-out). Answers are hints under the RFC's validate-on-use rule.
- **Observability** (`GET /directory/stats`): key/placement counts and per-instance stream state.

The cache-event vocabulary lives in `schemas.py` as the two-sided contract the MP-server emitter will import (same pattern as `L2EventListener`). Event entries already carry an optional `content_hash` so the M2 content index is additive.

**If applicable**:

- [x] this PR contains unit tests (34 new; full `tests/v1/mp_coordinator/` suite: 194 passing)

### DIFF
--- a/docs/design/v1/mp_coordinator/key_directory.md
+++ b/docs/design/v1/mp_coordinator/key_directory.md
@@ -1,0 +1,113 @@
+# Key directory
+
+Module: `lmcache/v1/mp_coordinator/key_directory.py`
+HTTP surface: `lmcache/v1/mp_coordinator/http_apis/directory_api.py` (`/directory/*`)
+
+This is the first milestone (M1) of the control-plane RFC
+([issue #4226](https://github.com/LMCache/LMCache/issues/4226)): a fleet-wide
+directory mapping each `ObjectKey` to its known placements — which instance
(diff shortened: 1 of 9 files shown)


ApostaC (approved):
LGTM. Please test the performance of the http endpoint or the key directory class

ruizhang0101:
Performance testing:
image

deng451e (approved):
LGTM

PR #4288, opened 2026-07-28
[Core] validate async memcpy host ranges

DongDongJu:
Validate arguments passed to asynchronous memcpy kernels before calculating copy boundaries.

  - Rejects zero and non-power-of-two host buffer alignments.
  - Detects size_t overflow in host_buffer_offset + nbytes.
  - Computes copy sizes without overflowing aligned boundary calculations.
  - Applies the same validation to CUDA and SYCL implementations.

If applicable:


DongDongJu:
Found during memory sharing work. normally hard to hit this condition.

PR #4292, opened 2026-07-29
[MP][Coordinator] Emit cache events from MP server to the key directory

ruizhang0101:

Refs #4226 (control-plane RFC). This is the emission half of the fleet-wide key directory (M1, directory side merged in #4275): MP servers stream `STORE`/`ACCESS`/`DELETE` cache events for **both tiers** to the coordinator's `/directory/events` — by riding the existing observability event bus rather than adding parallel listener plumbing.

- **One subscriber, no extra threads**: `CacheEventSubscriber` (on the `mp_observability` EventBus) maps the storage layer's key events onto the directory vocabulary, batches them order-preservingly (pending batches: consecutive same-identity entries coalesce, an identity change starts a new batch — a store→delete→re-store can never reorder), assigns per-instance `seq`, and delivers through `CacheEventSink` on the bus's drain thread. Flushing is interval-paced (`--coordinator-event-flush-interval`), with the L1 eviction loop's tick as the flush pump so burst-ending tails (e.g. L2 store completions) deliver promptly.
- **Transport seam**: the only sink today is HTTP push; a production message-queue sink (e.g. Kafka, partition key = `instance_id`) drops in without touching producers — the directory already absorbs at-least-once delivery (seq dedup), loss (gap detection → resync flag), and restarts (incarnation fencing, `incarnation` = server start time). Bus-queue overflow degrades the same way as transport loss.
- **Producers**: L1 store/evict events now carry `meta: list[L1ObjectMeta]` (per-object size + `L1Backend` medium via `L1ManagerProtocol.get_backend`; the Device-DAX tier resolves it per object through `MemoryObj.parent()`), so a hybrid DRAM+DAX L1 reports exactly where each object landed and deletes target the same placement identity. `touch_keys` publishes the new `l1.keys.accessed`. The L2 base adapter's notify funnel publishes `l2.keys.stored/accessed/deleted` with the adapter type name stamped at build time — **runtime-added adapters emit automatically**.
- `Tier` moves to `lmcache/v1/distributed/api.py` alongside the new `L1Backend` enum (`tiers.py` removed); `l1_primary_backend` lives next to `L1ManagerConfig`.

**User-facing changes**: `--coordinator-l2-event-reporting` / `--coordinator-l2-event-flush-interval` are replaced by `--coordinator-event-reporting` / `--coordinator-event-flush-interval` (env: `LMCACHE_COORDINATOR_EVENT_REPORTING` / `LMCACHE_COORDINATOR_EVENT_FLUSH_INTERVAL`). One flag gates both the key-directory stream and the existing quota stream (re-basing the quota stream onto this emitter is a follow-up PR). Event reporting requires observability enabled — combining it with `--disable-observability` fails fast at startup.

- Design doc: `docs/design/v1/mp_coordinator/cache_events.md` (transport contract, ordering/seq semantics, per-medium attribution, bus coupling); `docs/design/v1/mp_observability/event-bus.md` §7 updated for the third event category (loss-tolerant cache-event stream vs business-logic listeners vs observability).
- Verified end-to-end with vLLM (`LMCacheMPConnector`) + MP server + coordinator: directory placements matched on-disk L2 truth byte-for-byte (9/9 keys, L1+L2), ACCESS on prefix hit, DELETE removal across both tiers, and tick-driven tail delivery.

**If applicable**:


lmcache/v1/distributed/api.py
```diff
@@ -24,6 +24,32 @@
...
+    value (``Tier.L2 == "l2"``) and serializes as that value. ``ALL`` is only
+    valid for operations that explicitly support multiple tiers.
+    """
+
+    L1 = "l1"
+    L2 = "l2"
+    ALL = "all"
+
+
+class L1Backend(str, enum.Enum):
```
ApostaC: nit: `L1BackendType`

lmcache/v1/distributed/config.py
```diff
@@ -205,6 +206,25 @@ class L1ManagerConfig:
     """ Time to live for each object's read lock. Default is 300s (5 minutes). """
 
 
+def l1_primary_backend(config: L1ManagerConfig) -> L1Backend:
```
ApostaC: We can try to separate the L1 backend name for each event -- so we don't need `l1_primary_backend`

lmcache/v1/distributed/l1_manager.py
```diff
@@ -428,7 +429,7 @@ def finish_read(
         self._event_bus.publish(
             Event(
                 event_type=EventType.L1_KEYS_EVICTED,
-                metadata={"keys": need_to_free_keys},
+                metadata={"keys": need_to_free_keys, "meta": freed_meta},
```
ApostaC, line 432: Double-check whether there will be duplicated events: L1_KEYS_EVICTED and L1_KEYS_DELETED

lmcache/v1/mp_coordinator/cache_events.py
```diff
@@ -0,0 +1,328 @@
...
+        self._pending_batches: list[_PendingBatch] = []
+
+    def get_subscriptions(self) -> dict[EventType, EventCallback]:
+        """Return the storage key events this subscriber consumes."""
+        return {
+            EventType.L1_WRITE_FINISHED: self._on_l1_store,
+            EventType.L1_WRITE_FINISHED_AND_READ_RESERVED: self._on_l1_store,
+            EventType.L1_KEYS_EVICTED: self._on_l1_delete,
+            EventType.L1_READ_FINISHED: self._on_l1_access,
+            EventType.L1_KEYS_ACCESSED: self._on_l1_access,
```
ApostaC: note: L1_READ_FINISHED and L1_KEYS_ACCESSED could be duplicated. We can use L1_KEYS_ACCESSED

lmcache/v1/mp_coordinator/cache_events.py
ApostaC, line 1: Another side note: events for L1 clear

lmcache/v1/mp_coordinator/cache_events.py
```diff
@@ -0,0 +1,328 @@
...
+        if flush_interval < 0:
+            raise ValueError(f"flush_interval must be >= 0 (got {flush_interval})")
+        self._sink = sink
+        self._instance_id = instance_id
+        self._incarnation = incarnation
+        self._access_backend = access_backend.value
+        self._flush_interval = flush_interval
+        self._last_flush = time.monotonic()
+        self._seq = 0
+        self._lock = threading.Lock()
```
ApostaC: double check whether we can do lock-free because this class will only be accessed by the event bus thread.

ApostaC (commented):
ptal

lmcache/v1/distributed/memory_manager/devdax_l1_memory_manager.py
```diff
@@ -64,6 +66,22 @@ def __init__(self, config: L1MemoryManagerConfig) -> None:
         self._size_in_bytes = config.size_in_bytes
         self._align_bytes = config.align_bytes
 
+    def get_backend(self, memory_obj: MemoryObj) -> L1BackendType:
```
ApostaC: nit: get_backend_type

ApostaC (approved):
Otherwise LGTM

deng451e (approved):
lgtm

PR #4294, opened 2026-07-29
[Fix][RawBlock] handle uring_cmd SQE build errors

nayeonikim:
Fail the operation instead of hanging wait_iouring, and reject an alignment that is not a multiple of the NVMe LBA size.

  The io_uring worker discarded the result of `build_and_submit_sqe`. A submission
  that failed to build (offset or length not LBA-aligned, zero length,
  out-of-range NLB) had already been counted as in-flight but produced neither an
  SQE nor a CQE, so `wait_iouring()` for that batch blocked forever.

  - A build failure now completes the operation with that error and releases the
    global and per-batch in-flight accounting, so the caller gets an error instead
    of hanging. The retry path is indexed by the successfully built submissions
    rather than the original batch.
  - `RawBlockDevice` now rejects an `alignment` that is not a non-zero multiple of
    the NVMe namespace LBA size when `use_uring_cmd=true`. The configured
    alignment was never checked against the device geometry, so a 512-byte
    alignment on a 4Kn namespace was accepted at open and then produced
    LBA-unaligned offsets on the normal I/O path.

  Both new tests need an NVMe namespace character device (`/dev/ngXnY`) and skip
  without one, so they were run on real hardware: on a 4Kn namespace (LBA 4096)
  both pass; on a 512-byte-LBA namespace the build-error test passes and the
  alignment test skips as designed. `test_uring_cmd_middle_build_error_does_not_hang`
  fails only the middle request of a read-only `batched_read`
  (`offsets=[0, 1, lba_size]`) and runs in a subprocess with a join timeout, so a
  regression fails the test instead of hanging the suite.

  **If applicable**:


DongDongJu (approved):
Hello @nayeonikim, Thanks for the contribution.
LGTM. This fixes a real hang in the uring_cmd error path, and the accounting/retry changes look correct. 
Thanks for testing this on both 4Kn and 512-byte-LBA devices.

DongDongJu:
@zhengfeihe PTAL!

zhengfeihe (approved):
I ran locally with your fix& tests,confirmed passing. Thanks~ LGTM!

PR #4297, opened 2026-07-29
[Fix][NIXL] Fix close() deadlock when it races event loop startup

zhengfeihe:

Fix for passing  test_close_does_not_raise in nixl unit test.This flaky ut is affecting a lot of prs. 

The adapter starts its event loop thread as the last statement of __init__, so close() could check is_running() before the thread reached run_forever() and skip the only call that stops the loop, leaving the untimed join() blocked forever

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py
@@ -593,12 +593,18 @@ class NixlStoreL2Adapter(L2AdapterInterface):
             if tasks:
                 await asyncio.gather(*tasks, return_exceptions=True)
 
-        if self._loop.is_running():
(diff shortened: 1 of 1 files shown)


zhengfeihe:
```
tests/v1/distributed/test_nixl_store_l2_adapter.py::TestCloseInterface::test_close_releases_nixl_resources PASSED                                        [ 32%]
tests/v1/distributed/test_nixl_store_l2_adapter.py::TestCloseInterface::test_close_does_not_raise PASSED                                                 [ 32%]
```

DongDongJu (approved):
Hello @zhengfeihe, LGTM. 
The race analysis looks correct.
is_running() is not a safe condition here because close() can run after the loop thread starts but before it reaches run_forever(). 
In that window, skipping loop.stop() leaves the following join() blocked.

hlin99 (approved):
LGTM

Issue #4276, opened 2026-07-28
Refine Tests Structure: Cross-Module Cleanup Plan


VincyZhang:
## Summary
This issue tracks a structured cleanup of test organization and test guard logic to improve readability, portability, and platform compatibility across the test suite.

### Primary goals:
- Standardize device-aware vs device-specific patterns.
- Rework skip logic to be explicit and consistent.
- Normalize xpu/gpu marker usage.
- Remove hard-coded device assumptions and validate XPU behavior.

### Phase 1: basic
- Normalize device checks and helper usage in basic tests.
- Replace ad hoc skip checks with consistent patterns.
- Add device/framework markers for future use.
- Related PR: https://github.com/LMCache/LMCache/pull/4257 

### Phase 2: benchmark/cli/disagg
- Apply the same guard/marker conventions to benchmark, CLI, and disagg tests.
- Remove hard-coded backend assumptions where possible.
- Ensure XPU-only tests skip correctly on non-XPU runtime.
- Related PR: https://github.com/LMCache/LMCache/pull/4259

### Phase 3: v1/distributed
- Audit distributed tests for mixed device semantics.
- Separate device-aware paths from device-specific expectations.
- Normalize skip reasons and marker combinations.
- Perform hard-code cleanup, then run XPU validation for updated paths.
- Related PR: https://github.com/LMCache/LMCache/pull/4302, https://github.com/LMCache/LMCache/pull/4303

### Phase 4: v1/gpu_connector
- Align connector tests with platform abstraction usage.
- Remove direct hard-coded torch device usage when abstraction is available.
- Validate xpu/gpu marker semantics in connector-focused tests.
- Related PR: https://github.com/LMCache/LMCache/pull/4513

### Phase 5: v1/multiprocess
- Audit multi-process tests for environment- and device-dependent assumptions.
- Make skip logic deterministic and explicit.
- Perform hard-code cleanup, then run XPU validation for updated paths.
- Related PR: https://github.com/LMCache/LMCache/pull/4298

### Phase 6: v1/mp_coordinator/observability
- Normalize device-dependent test setup in coordinator and observability tests.
- Review flaky skip conditions and unify guard style.
- Ensure marker usage reflects true runtime dependencies.
- Perform hard-code cleanup, then run XPU validation for updated paths.
- Related PR: https://github.com/LMCache/LMCache/pull/4301

### Phase 7: v1/storage_backend
- Remove hard-coded device/backend combinations where generic logic is possible.
[...]

PR #4303, opened 2026-07-29
fix(turboquant): handle non-CUDA device capabilities

VincyZhang:

- Fix TurboQuant FP8 format selection on non-CUDA devices.
- `torch_dev.get_device_capability()` returns a CUDA `(major, minor)` tuple on CUDA but a dictionary on XPU. The previous comparison caused a type error on XPU.
- Use e4nv for non-CUDA capabilities, since e4b15 is specific to NVIDIA Ampere/Ada GPUs.
- Update TurboQuant serde tests to use LMCache device abstractions and derive lazy allocation from pin-memory support.

- XPU validation: `tests/v1/distributed/serde/test_turboquant.py` passed with `22 passed`.

**If applicable**:

Highlevel task tracking: https://github.com/LMCache/LMCache/issues/4276


tests/v1/distributed/serde/test_turboquant.py
```diff
@@ -276,7 +277,7 @@ def _make_turboquant_storage_manager(preset: str) -> StorageManager:
         l1_manager_config=L1ManagerConfig(
             memory_config=L1MemoryManagerConfig(
                 size_in_bytes=256 * 1024 * 1024,
-                use_lazy=torch.cuda.is_available(),
+                use_lazy=(torch_dev.is_available()),
```
DongDongJu: xpu support pinning behavior?

DongDongJu (commented):
Hello @VincyZhang, overall make sense.
Please check my one QQ.

tests/v1/distributed/serde/test_turboquant.py
VincyZhang: Good catch. 
XPU does not currently support pining behavior. `L1MemoryManagerConfig` currently auto-disables `use_lazy` in that case, which is why the XPU serde tests still pass through the non-lazy path. 
However, `torch_dev.is_available()` alone is misleading here. I will gate lazy allocation on the pin-memory capability `current_device_spec.is_pin_supported` instead.

hlin99 (approved):
LGTM

hlin99:
hi @DongDongJu PR scope has been limited. would you like to take a look again? thx

lmcache/v1/distributed/serde/turboquant/decode_kernel.py
```diff
@@ -24,7 +24,10 @@ def _use_fp8_e4b15(device: int = 0) -> int:
     """Return 1 if device needs fp8e4b15 (Ampere/Ada, SM < 8.9), else 0."""
     if device not in _FP8_E4B15:
         cap = torch_dev.get_device_capability(device)
-        _FP8_E4B15[device] = 1 if cap < (8, 9) else 0
+        # CUDA returns (major, minor) tuple; other backends (e.g. XPU) return
+        # a dict. fp8e4b15 is NVIDIA Ampere/Ada specific, so non-tuple caps
+        # always use e4nv (0).
+        _FP8_E4B15[device] = 1 if isinstance(cap, tuple) and cap < (8, 9) else 0
```
zhengfeihe, line 36: Can you update the comment?

1. some backends, eg: MUSA will still return tuple

https://github.com/MooreThreads/torch_musa/blob/467bb873388939c646ac4525f463bcdcc3eb570e/torch_musa/core/device.py#L83

Probably limit the scope to only XPU here, also add a note "To be verified " this might depend on real device on different backends ?

2. fp8e4b15 is NVIDIA Ampere/Ada specific

How about change this to 

Triton's fp8e4nv is supported on CUDA SM89+, so older CUDA architectures use fp8e4b15.

zhengfeihe (commented):
Otherwise LGTM~

lmcache/v1/distributed/serde/turboquant/decode_kernel.py
hlin99: thanks @zhengfeihe very good point!

hlin99:
seems latest code introduced a xpu failure. updating to latest to double check

zhengfeihe (approved):
LGTM

PR #4310, opened 2026-07-29
[MP][Coordinator] Route quota accounting through the cache-event stream

ruizhang0101:

Refs #4226. With the fleet cache-event stream in place (#4292), the coordinator's quota accounting no longer needs its own event feed. This PR routes it through the same stream and deletes the legacy path:

- **`CacheEventRouter`** (`cache_control/event_router.py`): batches the key directory *applies* (past seq dedup and incarnation fencing) fan out to the L2 usage ledger and eviction LRU — `STORE` → `record_stored`/`on_store`, `ACCESS` → `on_lookup`, `DELETE` → `record_evicted`/`on_remove`. Tier-`l2` only; the router is a standalone object so a future message-queue consumer calls it the same way the HTTP handler does.
- **Deleted**: `POST /quota/events`, the `UsageEvent`/`EventType`/`ReportUsage*` schemas, and the MP-server `L2EventListener` + its flush task. The MP server now runs one event client, one stream, one wire format.

**User-facing changes**: the `/quota/events` endpoint is removed (it was only ever called by MP servers); `/quota` config/status endpoints are unchanged. No flag changes.

**If applicable**:


lmcache/v1/mp_coordinator/cache_control/resync_manager.py
```diff
@@ -101,14 +113,31 @@ async def resync_from(
...
-                self._usage_manager.record_stored(obj_key, size_bytes)
-                self._eviction_manager.on_store(obj_key)
+                batch_entries.append(
+                    CacheEventEntry(key=encoded, size_bytes=size_bytes)
+                )
                 total += 1
+            if batch_entries:
+                # Placement identity must match the live event stream's
+                # (backend = adapter type name, plus the shared flag) so
+                # later DELETE events remove the backfilled placements.
```
tusharg1993, line 142: This comment suggests that getting a `body` with no shared flag or adapter name should be impossible. (This also aligns with what storage_manager.py expects too: https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/distributed/storage_manager.py#L1070).  

Should we just log an error here if we find any such malformed events to do idempotent all or nothing processing? Essentially, storing these events with unknown is a bit more risky IMO.

lmcache/v1/mp_coordinator/cache_control/resync_manager.py
tusharg1993, line 94: Asking for my own understanding and clarifying. https://github.com/LMCache/LMCache/blob/dev/lmcache/v1/multiprocess/cache_control/object_service.py#L51 shows me that when adapter is none, then only the first adapter is picked. 

Is this okay in current flow? Shouldn't we get the list of cache objects across all the adapters?

lmcache/v1/mp_coordinator/cache_control/eviction_manager.py
```diff
@@ -68,18 +73,39 @@ def __init__(
...
+        if batch.tier != Tier.L2:
+            return
+        for entry in batch.entries:
+            key = entry.key.to_object_key()
+            if batch.event_type == CacheEventType.STORE:
+                self.on_store(key)
+            elif batch.event_type == CacheEventType.ACCESS:
+                self.on_lookup(key)
+            elif batch.event_type == CacheEventType.DELETE:
+                self.on_remove(key)
```
tusharg1993: Please correct me if I am following the code wrong. L2UsageManager  now tracks L2 bytes per placement, so deleting one private copy correctly leaves any other copies counted. However, here on_remove removes the  key from the LRU even when another L2 placement remains. Then usage can still exceed quota while the eviction policy will disagree. Should we remove the key from the LRU only after its last L2 placement is gone?

lmcache/v1/mp_coordinator/cache_control/resync_manager.py
ruizhang0101: Hi, yes you are correct. Right now it will only get the first adapter that is configured. This was the implementation before. So for this PR that main goal is trying to refactor, we didn't modify this part of the logic. However, this can be done in the following PR. Despite that, I am actually thinking whether we should have resync in the future. Maybe we can do snapshot and event replay instead of resync since each storage medium need a resync which is not that scalable.

lmcache/v1/mp_coordinator/cache_control/resync_manager.py
ruizhang0101: Fixed.

lmcache/v1/mp_coordinator/cache_control/eviction_manager.py
ruizhang0101: Fixed

lmcache/v1/distributed/l2_adapters/base.py
```diff
@@ -358,22 +359,25 @@ def register_listener(self, listener: L2AdapterListener) -> None:
         """Register a listener to receive L2 adapter events."""
         self._listeners.append(listener)
 
-    def set_backend_name(self, name: str) -> None:
-        """Set the registered adapter type name used to tag cache events.
+    def set_backend_name(self, name: str, shared: bool = False) -> None:
```
ApostaC: Maybe move the shared flag to `__init__`

lmcache/v1/mp_coordinator/cache_control/event_router.py
```diff
@@ -0,0 +1,76 @@
...
+        consumer does not care about (tiers, event types) are its own
+        job to skip.
+
+        Args:
+            batch: The applied batch.
+        """
+        ...
+
+
+class CacheEventRouter:
```
ApostaC: Maybe CacheEventDispatcher or better names? (To avoid people feeling like this one  will "route" the event based on the event type)

lmcache/v1/mp_coordinator/api.py
```diff
@@ -62,7 +62,11 @@ class CacheEventBatch:
     """A batch of same-typed cache events from one MP server.
 
     Attributes:
-        instance_id: The emitting MP server (non-empty).
+        instance_id: The emitter's stream id (non-empty). MP servers use
```
ApostaC: we can use:
```
instance_id: the emitter's unique ID (non-empty).
```

lmcache/v1/mp_coordinator/key_directory.py
```diff
@@ -77,14 +82,21 @@ class InstanceDirectoryStats:
...
-        num_keys: Number of keys with at least one placement on the
-            instance.
+        num_l1_keys: Number of keys the stream has reported L1
+            placements for (the fencing index). Approximate for shared
+            pools: a counted placement may since have been deleted or
+            re-reported by another emitter.
+        num_l2_keys: Number of keys currently holding an L2 placement
+            whose last reporter is this instance (computed from the
+            placements at read time; L2 is not fenced, so this survives
+            the stream's restarts).
```
ApostaC: Simplify the docstring (only keep the first sentence + "eventual consistency")

lmcache/v1/mp_coordinator/key_directory.py
```diff
@@ -77,14 +82,21 @@ class InstanceDirectoryStats:
...
     """
 
     incarnation: int
     last_seq: int
     gap_detected: bool
-    num_keys: int
+    num_l1_keys: int
+    num_l2_keys: int
```
ApostaC: we can have only num_l1_keys (discard num_l2_keys)

ApostaC (commented):
Otherwise LGTM

ApostaC (approved):
LGTM!

KuntaiDu (approved):
LGTM!

sammshen (approved):
LGTM!

PR #4333, opened 2026-07-30
[Fix][Native Connector] Balance batch tiles

2xdevv:

Native connector batching previously calculated a rounded-up tile size and derived every range from that fixed size. When the item count was only slightly larger than the tile count, this concentrated work in the first tiles and produced empty requests, reducing worker-pool utilization.

For example, distributing 9 items across 8 tiles previously produced:

```text
2, 2, 2, 2, 1, 0, 0, 0
```

This PR:

- Uses quotient/remainder partitioning so tile sizes differ by at most one and every tile is nonempty.
- Produces the following distribution for 9 items across 8 tiles:

```text
2, 1, 1, 1, 1, 1, 1, 1
```

- Centralizes range calculation in a shared helper used by GET, SET, EXISTS, and DELETE.
- Removes the obsolete fixed tile size from batch preparation and tile request construction.

- The default tile count remains the smaller of the operation's worker count and item count.
- Mooncake continues to use one tile through its existing tile-count override.
- This changes no public API or configuration.
- No repository unit test was added because `ConnectorBase` does not currently have a C++ unit-test harness.

**If applicable**:

### DIFF
--- a/csrc/storage_backends/connector_base.h
+++ b/csrc/storage_backends/connector_base.h
@@ -84,7 +84,7 @@ class ConnectorBase : public IStorageConnector {
     validate_batch_inputs(keys, bufs, lens);
 
     size_t num_items = keys.size();
-    auto [batch_future_id, batch_state, num_tiles, tile_size] =
+    auto [batch_future_id, batch_state, num_tiles] =
         prepare_batch_operation(num_items, Op::BATCH_TILE_GET);
 
     // pre-allocate per-key results for load error tolerance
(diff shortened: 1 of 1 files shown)


2xdevv:
Hello, @DongDongJu, @sammshen, @hlin99. Could you take a look when you get a chance? Thanks.

DongDongJu (approved):
LGTM. Thanks for the improvement.

zhengfeihe (approved):
Thanks for the improvement. LGTM!

PR #4343, opened 2026-07-30
fix(mp_observability): stop the replaced bus in init_event_bus (#4340)

### LINKED ISSUE #4340: [Bug][MP Observability] init_event_bus leaks the previous EventBus drain thread on reinitialization
**Label**

`bug`, `mem_leak`, `mp_mode`

**Describe the bug**

`init_event_bus()` replaces the process-local observability `EventBus` singleton without stopping the previous instance.

If the previous bus has already been started, its background drain thread remains alive after replacement. Repeated initialization in the same Python process therefore causes the number of live `EventBus` threads to grow linearly.

Because `EventBus.stop()` is not called, the previous bus also does not run its subscriber shutdown hooks.

The normal MP server startup path initializes observability once, so this primarily affects tests, embedded or replay use cases, and any future same-process reconfiguration.

**To Reproduce**

Tested against `origin/dev` at commit `88c482d042d4272671ce4289f76fb9c896a26436`.

```python
import gc
import threading
import weakref

from lmcache.v1.mp_observability.event_bus import (
    EventBusConfig,
    init_event_bus,
)


def live_event_bus_threads() -> int:
    return sum(
        thread.name == "EventBus" and thread.is_alive()
        for thread in threading.enumerate()
    )


baseline = live_event_bus_threads()
old_bus_refs = []

for _ in range(10):
    bus = init_event_bus(EventBusConfig(enabled=True))
    bus.start()
    old_bus_refs.append(weakref.ref(bus))

    init_event_bus(EventBusConfig(enabled=False))

del bus
gc.collect()

print("extra live EventBus threads:", live_event_bus_threads() - baseline)

# Cleanup leaked buses so the reproducer does not leave background threads.
for bus_ref in old_bus_refs:
    old_bus = bus_ref()
    if old_bus is not None:
        old_bus.stop()

print("after cleanup:", live_event_bus_threads() - baseline)
```

Observed output:

```text
extra live EventBus threads: 10
after cleanup: 0
```

**Expected behavior**

Given that `init_event_bus()` is documented to replace the global singleton, it should not silently orphan a running bus.

The previous bus should either be stopped before replacement, or repeated initialization while a bus is running should be explicitly rejected and documented.

**Environment**

- OS: macOS 26.2
- Python: 3.12.11
- LMCache commit: `88c482d042d4272671ce4289f76fb9c896a26436`

**Additional context**

The current implementation directly overwrites the singleton:

```python
_global_bus = EventBus(config)
```

It does not call `stop()` on the previous instance.

I also ran an isolated verification in five independent Python processes, with 20 replacements per process. Each process retained exactly 20 live drain threads, and all threads exited after explicitly stopping the old buses. OpenTelemetry gauge registration was disabled for that verification to isolate the EventBus thread lifecycle.

This report is specifically about the process-local EventBus drain-thread and subscriber lifecycle. It does not require CUDA to reproduce and is not claiming a CUDA resource leak or a KV-cache data-path correctness issue.

I would be happy to work on the fix and regression test if the expected reinitialization semantics are confirmed.

Anai-Guo:
## Description

Closes #4340.

`init_event_bus()` rebinds the process-local `_global_bus` without stopping the bus it replaces. When the replaced bus had already been started, its background `EventBus` drain thread stays alive for the rest of the process, so repeated initialization grows the live thread count linearly. Because `stop()` never runs, the replaced bus also skips its subscriber `shutdown()` hooks.

The fix stops the previous bus **after** the swap, so publishers move to the new bus before the old one performs its final drain. The stop is wrapped in `try/except` so a subscriber that raises inside `shutdown()` cannot leave the global pointing at a half-initialized state.

As @adolphinvx notes in the issue, the normal MP server startup path initializes observability once, so this primarily affects tests, embedded/replay use cases, and same-process reconfiguration.

## Verification

Ran the reporter's reproducer against `dev` with only `lmcache.logging` / `otel_init` stubbed:

```
before: extra live EventBus threads: 10
after:  extra live EventBus threads: 0
```

`tests/v1/mp_observability/test_event_bus.py`: **44 passed** with the fix. Three regression tests added under `TestGlobalSingleton`:

- `test_init_stops_replaced_bus` — the previous drain thread exits and the subscriber's `shutdown()` runs. **Fails on unmodified `dev`.**
- `test_repeated_init_does_not_accumulate_threads` — thread count returns to baseline after 5 init/start cycles. **Fails on unmodified `dev`.**
- `test_init_survives_a_failing_subscriber_shutdown` — a subscriber raising in `shutdown()` still leaves `get_event_bus()` pointing at the new bus.

The existing `test_init_replaces_global` and `test_init_with_none_uses_defaults` are unchanged and still pass; no test relied on a replaced bus staying alive.

`stop()` is bounded here — `_run()` waits on `self._wake` with a 0.1 s timeout and re-checks `_stop_flag`, so `join()` returns promptly. Calling `stop()` on a never-started bus (including the module-level disabled default) is already a documented no-op path.

Lint/format clean under the pinned `ruff 0.11.7` (`check` and `format --diff`) using the repo's `pyproject.toml`. `isort 6.0.1` reports no new findings — the two headings it flags reproduce identically on unmodified `dev` in an out-of-tree check.

### DIFF
--- a/lmcache/v1/mp_observability/event_bus.py
+++ b/lmcache/v1/mp_observability/event_bus.py
@@ -379,9 +379,23 @@ def get_event_bus() -> EventBus:
 def init_event_bus(config: EventBusConfig | None = None) -> EventBus:
     """Replace the global singleton with a new EventBus built from *config*.
 
+    The bus being replaced is stopped after the swap, so its drain thread
+    exits and its subscribers run their shutdown hooks rather than being left
+    running for the lifetime of the process.
+
     Returns the newly created bus.
     """
     global _global_bus, _observability_enabled
+    previous_bus = _global_bus
     _global_bus = EventBus(config)
     _observability_enabled = config.enabled if config else True
+
+    # Stop the old bus only after the swap, so publishers move to the new bus
+    # before the old one performs its final drain.
+    if previous_bus is not None:
+        try:
+            previous_bus.stop()
+        except Exception:
+            logger.exception("EventBus: error stopping the replaced event bus")
+
     return _global_bus
--- a/lmcache/v1/mp_observability/event_bus.py
+++ b/lmcache/v1/mp_observability/event_bus.py
@@ -396,6 +396,14 @@ def init_event_bus(config: EventBusConfig | None = None) -> EventBus:
     exits and its subscribers run their shutdown hooks rather than being left
     running for the lifetime of the process.
 
+    Precondition: re-initialization is only safe once the existing producers
+    have stopped, or are prepared to re-fetch the global bus. A producer that
+    cached the result of ``get_event_bus()`` before this call keeps a reference
+    to the replaced bus, which is drained and stopped here, so anything it
+    publishes afterwards is dropped. Supporting reconfiguration while producers
+    are live needs a different design (lazy bus resolution in the publishers, or
+    a swap barrier) and is out of scope for this function.
+
     Returns the newly created bus.
     """
     global _global_bus, _observability_enabled
--- a/lmcache/v1/mp_observability/event_bus.py
+++ b/lmcache/v1/mp_observability/event_bus.py
@@ -114,6 +114,7 @@ class EventBus:
         self._stop_flag = threading.Event()
         self._lock = threading.Lock()
         self._thread: threading.Thread | None = None
+        self._stopped: bool = False
         self._registered_subscribers: list[EventSubscriber] = []
         self._discard_count: int = 0
         self._last_discard_warning: float = 0.0
@@ -229,6 +230,10 @@ class EventBus:
             return
 
         self._stop_flag.clear()
+        # Re-arm stop() so a bus that is started again can run its shutdown
+        # once more; without this a stop -> start -> stop cycle would skip the
+        # second cleanup.
+        self._stopped = False
         self._thread = threading.Thread(
             target=self._run,
             daemon=True,
@@ -239,7 +244,15 @@ class EventBus:
 
     def stop(self) -> None:
         """Stop the drain thread, flush remaining events, and shut down
-        all registered subscribers.  Safe to call when not started."""
+        all registered subscribers.  Safe to call when not started, and
+        idempotent: repeated calls after the first (until the bus is started
+        again) are no-ops, so each subscriber's ``shutdown()`` runs at most
+        once per start cycle."""
+        with self._lock:
+            if self._stopped:
+                return
+            self._stopped = True
+
         self._stop_flag.set()
         self._wake.set()
         if self._thread is not None and self._thread.is_alive():
--- a/tests/v1/mp_observability/test_event_bus.py
+++ b/tests/v1/mp_observability/test_event_bus.py
@@ -43,6 +43,7 @@ class _RecordingSubscriber(EventSubscriber):
         self._event_types = event_types
         self.events: list[Event] = []
         self.shutdown_called = False
+        self.shutdown_count = 0
 
     def get_subscriptions(self):
         return {et: self._on_event for et in self._event_types}
@@ -52,6 +53,7 @@ class _RecordingSubscriber(EventSubscriber):
 
     def shutdown(self) -> None:
         self.shutdown_called = True
+        self.shutdown_count += 1
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


adolphinvx:
Thanks for fixing #4340. I verified that repeated initialization no longer accumulates `EventBus` threads, and all 44 EventBus tests pass.

I found two potential lifecycle concerns:

1. **Cached publishers keep using the stopped bus.** The documented pattern is to store `get_event_bus()` in each producer. After re-initialization, those producers still reference the old bus, so events are queued without a drain thread and eventually dropped. If re-initialization is only valid after all producers stop, that precondition should be documented; otherwise runtime reconfiguration needs a different design.

2. **Subscriber shutdown can run twice.** Calling `bus.stop()` and later replacing it with `init_event_bus()` invokes the same subscriber’s `shutdown()` twice. The hook is not required to be idempotent, so `EventBus.stop()` should guard against repeated cleanup and include a regression test.

One smaller test issue: `test_init_survives_a_failing_subscriber_shutdown` does not exercise the new outer `try/except`, because `EventBus.stop()` already catches subscriber shutdown exceptions internally.

Anai-Guo:
Thanks for the careful review, @adolphinvx — both lifecycle points are valid.

**2. Double subscriber shutdown.** Fixed. `EventBus.stop()` is now idempotent: it guards its cleanup (final drain + subscriber `shutdown()`) so it runs at most once per start cycle, and `start()` re-arms the guard so a bus that is started again can still shut down on its next `stop()`. Added `test_stop_is_idempotent` (asserts `shutdown_count == 1` after two `stop()` calls — it fails on the pre-guard code) and `test_stop_runs_again_after_restart`.

**3. Test didn't exercise the outer `try/except`.** Correct — `stop()` already catches subscriber exceptions internally, so a subscriber that raises in `shutdown()` never reached the outer guard in `init_event_bus()`. Replaced `test_init_survives_a_failing_subscriber_shutdown` with `test_init_survives_a_failing_stop`, which makes the replaced bus's `stop()` itself raise and verifies the swap still completes and the exception is swallowed.

**1. Cached publishers keep referencing the stopped bus.** Agreed this is real, but it's the broader runtime-reconfiguration problem rather than the thread-leak this PR targets — a producer that cached `get_event_bus()` before re-init keeps publishing to the old (now-drained) bus. Rather than expand scope here, I think the honest fix is to document the precondition (re-initialization is only safe once producers have stopped/re-fetched the bus) and, if runtime reconfiguration becomes a requirement, track it separately since it needs a different design (e.g. producers resolving the bus lazily, or a swap barrier). Happy to add that doc note in this PR or open a follow-up issue — whichever you prefer.

aryangorde8:
The red CI on this PR looks stale rather than a real failure. I tested the change against current `dev` (`f82f6fd3`) because I ran into the same leak from a different direction in #4491.

**The patch still applies cleanly** to `dev` — no rebase conflict.

**With it applied, `tests/v1/mp_observability/` gives 393 passed, 1 failed, 1 skipped.** The single failure is `subscribers/logging/test_extra_stats.py::TestExtraStatsLoggingSubscriber::test_store_window_logs_tokens_size_and_throughput`, and I confirmed it is **pre-existing**: after reverting the patch, that test fails identically on clean `dev`. It is a timing-dependent flake unrelated to this change — the test's synthetic event timestamps are always older than `_prune_pending`'s 60-second wall-clock deadline, so if a flush happens to fire between subscriber construction and the first event, the pending START is dropped and the throughput reads `n/a`. I will open a separate issue for it.

Scope of what I actually ran, so this is not overstated: the `mp_observability` suite only, on a CPU-only machine with no GPU. The failing checks here are `k3-unit-tests` and `unit-tests-amd`, which cover much more than that, and those runs are from 30–31 July — `dev` has moved a long way since. So this is not proof the PR is green, only that the fix is not what breaks it, and that a re-run is worth doing before anyone reads the diff as broken.

@Anai-Guo — if you push a rebase onto current `dev`, CI should re-run against a much newer base.

On the change itself, the `_stopped` flag with re-arming in `start()` reads correctly to me, and stopping the outgoing bus *after* the swap is the right order — publishers move to the new bus before the old one performs its final drain. Independently reproduced the underlying leak on `dev`: two `init_observability` calls leave two live `EventBus` threads, the first still draining while unreachable, which is exactly what #4340 describes.

Worth noting this is not only a test concern. `lmcache/v1/multiprocess/server.py:368` and `lmcache/cli/commands/trace/_driver.py:177` both call `init_observability`, so any process that initialises observability twice leaks a started bus with every subscriber still attached.

adolphinvx:
Thanks @Anai-Guo — the idempotent stop() change and the strengthened failing-stop() test address my second and third points.
For the cached-publisher concern, I’d prefer to document the precondition in this PR: re-initialization is only safe after existing producers have stopped or re-fetched the global bus. A separate follow-up issue for proper runtime reconfiguration would also make sense, since that requires a broader design.
@aryangorde8’s independent reproduction and focused test results are helpful and support the fix. Since the PR has not been updated since July 31 and the required Buildkite checks are still red, could you rebase onto the current dev branch and rerun CI?
With the documentation note and refreshed CI, I have no further concerns from my review.
Once the branch is rebased and CI is refreshed, a maintainer review would be appreciated.

Anai-Guo:
Thanks @adolphinvx — doc note pushed in 8f49ee4.

`init_event_bus`'s docstring now states the precondition explicitly: re-initialization is only safe once existing producers have stopped or are prepared to re-fetch the global bus, since a producer that cached `get_event_bus()` before the swap keeps publishing into the bus this function drains and stops. It also names why the general case needs a different design (lazy bus resolution in the publishers, or a swap barrier) rather than an extension of this fix. I'll open the follow-up issue for runtime reconfiguration so it's tracked separately.

On the rebase: I can't produce one from this fork. `Anai-Guo/LMCache:dev` is pinned at `8d3ce30` (2026-07-17) and `POST /merge-upstream` is refused with

> refusing to allow an OAuth App to create or update workflow `.github/workflows/build_cli_artifacts.yml` without `workflow` scope

My token has no `workflow` scope, and creating a ref at any newer `dev` commit hits the same content guard, so the merge-base can't move forward. What the new commit does give you is a fresh CI run: the checks re-evaluate on the PR merge ref, which is this branch merged with today's `dev` tip, and the PR is still `MERGEABLE` with the diff confined to the same two files. If the Buildkite pipeline needs a real rebased branch rather than the merge ref, a maintainer push to the branch (or a re-run with "update branch") would do it — I just can't drive that from here.

PR #4347, opened 2026-07-30
fix(memory): AdHocMemoryAllocator must not hand out uninitialized memory

mykaul:
## Summary
- `AdHocMemoryAllocator.allocate()` returned `torch.empty()` buffers -- genuinely uninitialized memory, whatever bytes happened to be there. For a floating dtype, those bytes can decode to a NaN/Inf bit pattern.
- Many tests across the suite (Scylla backend, DAX backend, and others) allocate test `MemoryObj`s through this allocator, round-trip them through a backend, and assert `torch.equal(retrieved.tensor, original.tensor)`. `torch.equal()` returns `False` for NaN vs NaN even when the tensors are otherwise bit-identical, so these assertions failed intermittently depending on whatever garbage was left in that memory at allocation time -- not a real bug in the code under test.
- Fixed by filling the allocated buffer with real (if arbitrary) finite content, per shape/dtype group, respecting each group's own dtype (`torch.randn(...).to(dtype)` for floating dtypes, zeros otherwise) instead of leaving it uninitialized.

## Test plan
- [x] `pytest tests/v1/test_memory_allocator_import_surface.py` passes
- [x] Repro'd the original flakiness against the pre-fix allocator (confirmed via `torch.equal` returning `False` on bit-identical NaN-containing tensors), and confirmed it's gone after the fix, across multiple fresh-process runs
- [x] Existing tests that consume this allocator (Scylla backend, DAX backend test suites) still pass

### DIFF
--- a/lmcache/v1/memory_allocators/ad_hoc_memory_allocator.py
+++ b/lmcache/v1/memory_allocators/ad_hoc_memory_allocator.py
@@ -8,7 +8,7 @@ import torch
 
 # First Party
 from lmcache import torch_dev
-from lmcache.utils import _lmcache_nvtx_annotate, get_size_bytes
+from lmcache.utils import _lmcache_nvtx_annotate
 from lmcache.v1.memory_management import (
     MemoryAllocatorInterface,
     MemoryFormat,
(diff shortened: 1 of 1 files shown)


hlin99 (approved):
LGTM
can you fix DCO issue? thx for the contribution

PR #4350, opened 2026-07-30
[GDS] O_DIRECT + pinned-buffer POSIX fallback for GdsBackend

xiangping-chen:

Adds an **O_DIRECT + pinned-host-buffer POSIX fallback** to `GdsBackend`, so the backend performs well on hosts without GPUDirect Storage (`use_gds: false`) and works on filesystems where the current implementation cannot run at all.

Two concrete problems with the existing `use_gds: false` path:

1. It writes via a `MAP_SHARED` writable `mmap`. Parallel/network filesystems and FUSE-backed mounts reject shared writable mappings with `EOPNOTSUPP`, so the fallback simply fails there.
2. It stages through pageable memory, forcing the driver to bounce through its own internal pinned buffer before the `cudaMemcpy`/`hipMemcpy`.

This replaces the mmap with a per-thread **pinned** host buffer plus a single `pwrite`/`preadv`, optionally with `O_DIRECT` (new `extra_config.use_direct_io`) to bypass the page cache. Pinned staging makes the host↔device copy a full-bandwidth DMA. Unaligned transfers, or a filesystem that refuses `O_DIRECT`, degrade to buffered I/O automatically. `use_gds: false` also now uses a plain `GPUMemoryAllocator`, so the path no longer depends on cuFile being loadable.

The original implementation is by **Jason Goldschmidt** (first commit; his `Signed-off-by` is preserved, with mine as relayer). It was written against an older `dev`, so I rebased it onto current `dev` and **dropped the parts upstream has since implemented independently** — the CUDA/ROCm memcpy loader (#3970, more robust than the original's `torch.version.hip` dispatch) and enabling the thread pool for both datapaths (#4067). That removed roughly a quarter of the original diff.

The second commit fixes defects found in review and during live testing:

- **Pinned-buffer tracking crash.** `_get_pinned_buffer` tracked buffers in a `list` and called `list.remove(old_buf)` on growth. `list.remove` compares with `==`, which on tensors yields a tensor and then raises `RuntimeError` from `bool()` as soon as another thread's buffer precedes it — and `RuntimeError` is not caught by the surrounding `except ValueError`. In `_load_gds` that is swallowed into `return -1`, which makes the caller evict the entry, so the symptom is **silent cache loss**. Reproduced with torch 2.11: 3 of 4 threads raise. Now keyed by thread id.
- **O_DIRECT was silently non-deterministic.** `torch.empty(pin_memory=True)` does not guarantee page alignment — the caching pinned allocator sub-allocates, so after one unaligned-sized request later pointers come back offset (observed 2048, 2560). `_direct_io_aligned` then rejects every transfer and O_DIRECT quietly degrades to buffered based on nothing but allocation history. Surfaced as a test-ordering failure. Fixed by over-allocating one alignment unit and returning an aligned view.
- **Missing save guard.** `_save_gds` had no `else` for "neither GDS module nor GPU memcpy available": it wrote nothing, then failed in `os.rename` with a `FileNotFoundError` for a temp file never created, after the caller was told the chunk was persisted. Now raises explicitly, mirroring `_load_gds`. The `pwrite` loop also gained a no-progress guard, and a failed save unlinks its partial temp file.
- **O_DIRECT refusal was under-detected.** Only `EINVAL` was latched. The filesystems this fallback targets refuse with `EOPNOTSUPP`/`ENOTSUP` or `EPERM`, which escaped the handler and propagated out — turning graceful degradation into a hard failure exactly where it was needed.
- **Allocator chosen before GDS was resolved.** `initialize_allocator` read `config.use_gds` and ran *before* the fstype-based auto-disable, so the tmpfs/overlayfs path built a `CuFileMemoryAllocator` — requiring libcufile — while the rest of the backend ran the POSIX fallback. libcufile is exactly what may be missing there.
- **Work ran on the wrong CUDA device.** PyTorch's current device is thread-local and defaults to index 0, and neither the `gds-io` pool nor asyncio's executor selected one, so a rank owning e.g. `cuda:3` created a primary CUDA context on `cuda:0` from every I/O thread. The pool now has a device-selecting initializer, and the pinned allocation and both memcpys run in a `torch_dev.device()` context. `dst_device` is resolved to a concrete index first, since `set_device` rejects the index-less `"cuda"` that is its default.

- Tested end-to-end against live vLLM (Qwen3-Coder-30B-A3B, TP=1, RTX PRO 6000 Blackwell / sm_120) with the KV store on XFS over a 4-drive NVMe md RAID-0. Full 4K–126,900-token TTFT sweep, 30 passes, re-run after each fix: 0 errors, 0 unaligned fallbacks, 0 `O_DIRECT` rejections — i.e. O_DIRECT used for every I/O. Performance flat across all iterations (126,900-token hot TTFT ~1,255–1,274 ms), so none of the fixes cost anything.
- The pre-existing `test_use_gds_false_save_load_roundtrip` uses a 2048-byte chunk, so its transfers are unaligned and every I/O took the *buffered* branch — the O_DIRECT path had **no** coverage. Added an aligned round-trip test asserting `O_DIRECT` actually reaches `os.open`, plus regression tests for the pinned-buffer crash, the errno latch (with `EINVAL` kept as a control), the fstype auto-disable allocator, and the device targeting. Each was confirmed to fail without its corresponding fix.
- **On partial O_DIRECT transfers:** an earlier review round flagged that the read/write retry loops would use unaligned offsets after a short transfer. On investigation this is not reachable — Linux requires O_DIRECT transfers to be a multiple of the device logical block size (512 here, verified empirically: `len=512` and `off=512` are accepted, `len=1000` and `off=1000` return `EINVAL`), so a partial count is necessarily block-aligned and the continuation stays valid. No change made; noting it so the next reviewer doesn't re-derive it.
- Remaining known gaps, happy to address here or as follow-ups: `close()` does not free pinned buffers held by asyncio executor threads (only pool threads exit); user docs for `use_direct_io` are not updated; and the D2H copy is not explicitly stream-ordered against the producer stream.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/gds_backend.py
+++ b/lmcache/v1/storage_backend/gds_backend.py
@@ -6,8 +6,8 @@ from typing import Any, Callable, List, Optional, Sequence, Tuple, Union
 import asyncio
 import ctypes
 import ctypes.util
+import errno
 import json
-import mmap
 import os
 import struct
 import threading
@@ -17,7 +17,6 @@ import uuid
 
 # Third Party
 import aiofile
-import numpy as np
 import torch
 
 # First Party
@@ -30,6 +29,7 @@ from lmcache.utils import (
 )
 from lmcache.v1.config import LMCacheEngineConfig
 from lmcache.v1.memory_allocators.cu_file_memory_allocator import CuFileMemoryAllocator
+from lmcache.v1.memory_allocators.gpu_memory_allocator import GPUMemoryAllocator
 from lmcache.v1.memory_allocators.hip_file_memory_allocator import (
     HipFileMemoryAllocator,
 )
@@ -51,6 +51,14 @@ _METADATA_MAX_SIZE = 4096  # reserve 4K for metadata.
 # TODO: It is possible to read this 4KB block without triggering read-ahead by
 # various means.
 _DEFAULT_THREAD_COUNT = 4
+
+# O_DIRECT (buffered-fallback POSIX path) alignment. O_DIRECT requires the file
+# offset, transfer length, and memory buffer to all be aligned to the block
+# size; 4096 satisfies every common Linux filesystem (and equals the metadata
+# header size, so KV data starts on an aligned boundary). getattr keeps this
+# importable on platforms without O_DIRECT (e.g. macOS), where it resolves to 0.
+_DIRECT_IO_ALIGN = 4096
+_O_DIRECT_FLAG = getattr(os, "O_DIRECT", 0)
 
 
 class UnsupportedMetadataVersion(Exception):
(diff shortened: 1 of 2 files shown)


zhengfeihe:
Thank you for your contribution and for thoroughly testing this feature!

However, in-process mode will be deprecated soon. If you're interested, we recommend contributing to MP mode instead.

xiangping-chen:
> Thank you for your contribution and for thoroughly testing this feature!
> 
> However, in-process mode will be deprecated soon. If you're interested, we recommend contributing to MP mode instead.

Sure. I will add the fix to MP mode. Thanks!

PR #4351, opened 2026-07-30
[core][feat] add NL_X_NB_BSV_BSS  format  for CB 

deng451e:

**If applicable**:

### DIFF
--- a/csrc/mp_mem_kernels.cu
+++ b/csrc/mp_mem_kernels.cu
@@ -49,8 +49,10 @@ __device__ inline size_t calculate_engine_global_offset(
     // Flash Infer HND: L tensors [NB, 2, NH, BS, HS]
     return engine_block_idx * shape_desc.kv_size * scalars_per_block +
            k_or_v * scalars_per_block;
-  } else if constexpr (format == EngineKVFormat::NL_X_NB_BS_HS) {
-    // MLA: L tensors [NB, BS, HS]
+  } else if constexpr (format == EngineKVFormat::NL_X_NB_BS_HS ||
+                       format == EngineKVFormat::NL_X_NB_BSV_BSS) {
+    // MLA: L tensors [NB, BS, HS]; blocked-scale shares the block base,
+    // only its within-block layout differs (handled in the transfer body).
     return engine_block_idx * scalars_per_block;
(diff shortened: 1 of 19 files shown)


csrc/mp_mem_kernels.cu
```diff
@@ -223,6 +225,29 @@ __device__ void multi_layer_block_transfer_single_block(
...
+        warp_copy<ScalarType>(eng_vals, lmc_row, val_units);
+        warp_copy<ScalarType>(eng_scale, lmc_row + val_units, scale_units);
+      } else {
+        warp_copy<ScalarType>(lmc_row, eng_vals, val_units);
+        warp_copy<ScalarType>(lmc_row + val_units, eng_scale, scale_units);
+      }
+    }
+    return;
+  }
+
```
ApostaC, line 250: We probably don't need this.

csrc/mp_mem_kernels.cu
```diff
@@ -411,6 +439,16 @@ void multi_layer_block_kv_transfer(
...
 
+  if (engine_kv_format == EngineKVFormat::NL_X_NB_BSV_BSS) {
+    // Blocked-scale indexer cache: the per-token fp32 scale must be a whole
+    // number of transfer units, so pin 4-byte units regardless of row width.
+    TORCH_CHECK(head_bytes % sizeof(uint32_t) == 0,
+                "NL_X_NB_BSV_BSS row bytes (", head_bytes,
+                ") must be divisible by 4");
+    LAUNCH_TEMPLATED(uint32_t);
+    return;
+  }
```
ApostaC, line 450: This can be dropped as well

ApostaC (approved):
LGTM!

csrc/mp_mem_kernels.cu
deng451e: This split copy is required. LMCache chunks are token-major (2LTD), with each token stored as [128B values | 4B scale], while vLLM pages store values and scales in separate regions. This branch converts between those layouts. store uses this block kernel, while CB RETRIEVE uses the token-granular path and may write to shifted offsets. A contiguous copy would store raw page data that RETRIEVE misinterprets as token records, reintroducing cache corruption.

KuntaiDu (approved):
lgtm!

PR #3968, opened 2026-07-01
[MP][Ascend] Ascend NPU multiprocess format support + generic engine-driven path

chloroethylene:
[MP][Ascend] support Ascend NPU and the per-layer (K, V) tuple KV format
Add Ascend NPU as a first-class platform and introduce the per-layer
(K, V) tuple engine KV format (NL_X_TWO_X_NB_BS_NH_HS) produced by
vLLM-Ascend, where layers are outermost and each layer carries its own
(K, V) pair of paged [num_blocks, block_size, num_heads, head_size]
tensors.

Platform:
- New NpuDeviceInfo detection entry (lmcache/v1/platform/npu) that loads
  the lmcache_ascend.c_ops backend on an NPU host; native kernels stay
  in the LMCache-Ascend plugin.
- New Ascend build profile (setup_extensions/build_profiles/ascend.py):
  detects the CANN toolchain via ASCEND_HOME_PATH and supplies the ABI
  flags; the upstream package builds only the common C++ extensions.

KV format (NL_X_TWO_X_NB_BS_NH_HS):
- Register the enum value across the C++ header, pybind bindings, and the
  Python fallback, and add the orthogonal is_kv_second_tuple modifier.
- Allow tuples in DiscoverableKVCache, list-depth probing, and device
  lookup so the native per-layer (K, V) structure is recognized end to
  end; add a vLLM detector branch and a format spec.
- Add the _transfer_per_layer_kv_tuple path, which transposes the
  structure to the SGLang MHA [[K_layers], [V_layers]] layout the kernel
  consumes.
- Route the transfer context gather/scatter through get_device /
  get_dtype / get_group_data_ptrs instead of tensors[0], so the nested
  per-layer tuple layout resolves device, dtype, and plane pointers
  correctly.

Tests:
- Extend the format-classification golden tests to pin the new format's
  structural partition and the is_kv_second_tuple modifier.
- Add compute_kv_layout and accessor/grouping tests for the tuple format.

**What this PR does / why we need it**:

**Special notes for your reviewers**:

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [x] this PR contains unit tests

PR #4362, opened 2026-07-31
Add Ascend NPU (torch_npu) support

cklxx:
## What

Add Huawei Ascend NPU (`torch_npu`) support to LMCache, so KV cache can be
offloaded from vLLM-Ascend to any LMCache backend on Ascend hardware.

## Changes

- **`platform/npu`**: an `NpuDeviceSpec`/`NpuDeviceOps` plugin so the device
  registry auto-discovers Ascend (`torch.npu`) as `"npu"`, mirroring the
  existing `hpu` plugin. No change to the detection core.
- **`CreateGPUConnector`**: dispatch `VLLMPagedMemNPUConnectorV2` for `npu`.
- **`npu_connector.py`**: an NPU paged-mem KV connector. Device-agnostic
  `.to()` copies, `torch.npu.synchronize()`, eager mode (no `mark_step`).
  Each MLA layer's latent cache arrives as a 1-tuple, unwrapped before
  `.view()`.
- **KV-format detector**: recognize the Ascend MLA layout — a depth-2 list of
  rank-4 `[num_blocks, block_size, 1, head_size]` latent caches — and
  normalize it to the canonical rank-3 `NL_X_NB_BS_HS` format.
- **metadata head_size**: vLLM-Ascend caches only the compressed MLA latent
  (`kv_lora_rank`), not `kv_lora_rank + qk_rope_head_dim`. Size the memory
  pool to the latent on NPU so it matches the on-device cache tensor. Gated
  on the `npu` device, so CUDA MLA (which caches the full dim) is unaffected.

## Testing

Verified end-to-end on a Huawei Ascend 910B host (CANN, `torch_npu`) with
vLLM-Ascend serving a GLM-style MLA model, `LMCacheConnectorV1`,
`--tensor-parallel-size 16 --enforce-eager`:

- **store**: `Stored 1536 out of total 1536` tokens to the remote backend.
- **remote read**: after a full server restart (local CPU cache empty), the
  same prompt yields `External prefix cache hit rate: 91.4%` — the hit can
  only come from the remote backend.
- no crashes across repeated store/retrieve requests.

The `head_size` change is the key fix: without it, `memory_obj.tensor.copy_`
fails with a `576 vs 512` shape mismatch (`kv_lora_rank + rope` vs the
latent-only on-device cache).

### DIFF
--- a/lmcache/integration/vllm/utils.py
+++ b/lmcache/integration/vllm/utils.py
@@ -275,6 +275,18 @@ def create_lmcache_metadata(
     chunk_size = config.chunk_size
     num_kv_head = model_cfg.get_num_kv_heads(parallel_cfg)
     head_size = model_cfg.get_head_size()
+    if use_mla:
+        from lmcache import torch_device_type
+        if torch_device_type == "npu":
+            # vLLM-Ascend caches only the compressed MLA latent
+            # (kv_lora_rank); size the pool to it so it matches the
+            # on-device KV cache tensor.
+            for _cfg in (getattr(model_cfg, "hf_text_config", None),
+                         getattr(model_cfg, "hf_config", None)):
+                _lora = getattr(_cfg, "kv_lora_rank", None)
+                if _lora:
+                    head_size = _lora
+                    break
     kv_shape = (num_layer, 1 if use_mla else 2, chunk_size, num_kv_head, head_size)
 
     # Extract engine_id and kv_connector_extra_config from vllm_config if available
--- a/lmcache/integration/vllm/vllm_service_factory.py
+++ b/lmcache/integration/vllm/vllm_service_factory.py
@@ -100,6 +100,18 @@ class VllmServiceFactory(BaseServiceFactory):
         chunk_size = self.lmcache_config.chunk_size
         num_kv_head = model_config.get_num_kv_heads(parallel_config)
         head_size = model_config.get_head_size()
+        if use_mla:
+            from lmcache import torch_device_type
+            if torch_device_type == "npu":
+                # vLLM-Ascend caches only the compressed MLA latent
+                # (kv_lora_rank); size the pool to it so it matches the
+                # on-device KV cache tensor.
+                for _cfg in (getattr(model_config, "hf_text_config", None),
+                             getattr(model_config, "hf_config", None)):
+                    _lora = getattr(_cfg, "kv_lora_rank", None)
+                    if _lora:
+                        head_size = _lora
+                        break
         kv_shape = (
             num_layer,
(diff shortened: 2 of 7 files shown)


hlin99:
hi @cklxx thanks for your contribution and would love to see Ascend is back from plugin.

very quick question: in process mode is deprecating and won't be maintained anymore. in this case, looks like adding npu connector is not quite necesary, maybe we can remove this part out of the PR and leave the other part as it is. 

WDYT?

cklxx:
Thanks @hlin99! Glad the plugin direction fits.

One constraint worth flagging: on Ascend the in-process `LMCacheConnectorV1` path is currently the only one that works end-to-end — I verified store + remote read (90%+ external hit, no crashes) on a 910B with vLLM-Ascend + GLM MLA. The MP-server path needs `native_storage_ops`, which has no NPU build yet, so dropping `npu_connector.py` would leave NPU without a working transfer path today.

Would you prefer I keep the connector until NPU MP / `device_ops` lands, or split it into a separate PR? And what's the rough timeline for MP on NPU?

hlin99:
> Thanks @hlin99! Glad the plugin direction fits.
> 
> One constraint worth flagging: on Ascend the in-process `LMCacheConnectorV1` path is currently the only one that works end-to-end — I verified store + remote read (90%+ external hit, no crashes) on a 910B with vLLM-Ascend + GLM MLA. The MP-server path needs `native_storage_ops`, which has no NPU build yet, so dropping `npu_connector.py` would leave NPU without a working transfer path today.
> 
> Would you prefer I keep the connector until NPU MP / `device_ops` lands, or split it into a separate PR? And what's the rough timeline for MP on NPU?

there's another PR that enable ascend mp mode. not sure if you're aware of it.  https://github.com/LMCache/LMCache/pull/3968

cklxx:
Thanks for the pointer — #3968 looks like the right home for this. It adds the Ascend native build (the `native_storage_ops` gap I hit), the per-layer tuple format spec, and the MP transfer path, all on the non-deprecated MP route, which supersedes the in-process approach here.

Closing this in favor of #3968. One thing worth double-checking there: MLA memory-pool sizing — on GLM-style MLA, vLLM-Ascend caches only the compressed latent (`kv_lora_rank`, e.g. 512), while `model_config.get_head_size()` returns `kv_lora_rank + qk_rope_head_dim` (576), so sizing the pool to the full dim mismatches the on-device cache. If #3968's format specs already cover that, great — otherwise happy to send it as a small follow-up. I can also help test #3968 on a 910B.

cklxx:
Superseded by #3968 (Ascend MP mode). Closing in favor of that.

Issue #4364, opened 2026-07-31
[RFC] Consolidate O_DIRECT handling: nine independent implementations, none of which handles a filesystem refusing the flag


xiangping-chen:
**Label**

`RFC`, `Refactoring`, `backend`, `mp_mode` — also touches `nixl`, `localdisk`, `gds`, `fs_connector`.

---

## Is your feature request related to a problem? Please describe.

O_DIRECT gating is reimplemented independently in **nine** places, and no two agree on behaviour. Verified against `upstream/dev`:

| # | Site | Mode |
|---|------|------|
| 1 | `lmcache/v1/storage_backend/local_disk_backend.py` | in-process |
| 2 | `lmcache/v1/storage_backend/connector/fs_connector.py` | in-process |
| 3 | `lmcache/v1/storage_backend/nixl_storage_backend.py` (static pool + dynamic = 2 copies) | in-process |
| 4 | `lmcache/v1/storage_backend/raw_block/core.py` + `rust/raw_block/src/lib.rs` | shared |
| 5 | `lmcache/v1/gpu_connector/gds_context.py` | shared (GDS L1) |
| 6 | `lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py` | **MP** |
| 7 | `lmcache/v1/distributed/l2_adapters/nixl_store_l2_adapter.py` | **MP** |
| 8 | `lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py` | **MP** |
| 9 | `csrc/storage_backends/fs/connector.cpp` (drives MP `fs_native`) | **MP** |

### 1. No site handles a filesystem that refuses O_DIRECT

Every site detects support by asking the **platform** — `hasattr(os, "O_DIRECT")`, `getattr(os, "O_DIRECT", 0)`, or `#ifdef O_DIRECT` — then opens without classifying the errno. Across all nine files:

```
grep -c 'errno\.(EINVAL|EOPNOTSUPP|ENOTSUP|EPERM)'   ->   0   in every one
```

The platform check answers a different question from the one that matters. `open()` may refuse the flag with `EINVAL` (documented), a filesystem may report `EOPNOTSUPP`/`ENOTSUP`, and `EPERM` is possible under policy. Because none of that is classified, the failure modes are inconsistent and none is a fallback:

[...]

PR #4366, opened 2026-07-31
[Fix][MP] Load an FSL2Adapter batch concurrently instead of key by key (2.2x)

### LINKED ISSUE #4365: [Bug][MP] FSL2Adapter serialises every key inside a load batch (2.2x slower on 24 MiB chunks, measured)
**Label**

`bug`, `performance`, `mp_mode`, `backend` — also relevant to `fs_connector`.

---

## Describe the bug

`FSL2Adapter` never parallelises the keys inside a `submit_load_task` batch. Each key's I/O
is awaited before the next begins, so a batch of N keys takes N sequential round trips.

`lmcache/v1/distributed/l2_adapters/fs_l2_adapter.py`, `_execute_load`:

```python
for i, key in enumerate(keys):
    ...
    if self._use_odirect:
        num_read = await self._loop.run_in_executor(   # awaited in-loop
            None, self._read_with_odirect, file_path, dst_buf,
        )
        ...
        continue
    async with aiofiles.open(file_path, "rb") as f:
        num_read = await _async_readinto_full(f, dst_buf)   # awaited in-loop
```

Dispatching to `run_in_executor` does not help, because the result is awaited immediately.
The same shape appears in the store path (`for key, obj in zip(keys, objects, strict=True)`,
which awaits `exists` → write → `replace` per key, i.e. three sequential round trips each)
and in the lookup path (`for i, key in enumerate(keys)` awaiting `_key_exists_on_disk`).

**There is no outer concurrency to compensate.** `submit_load_task` schedules exactly one
coroutine per batch, and the caller puts a whole request into one batch —
`storage_controllers/prefetch_controller.py:931`, "Step 7: submit load tasks per adapter":

```python
for adapter_idx, bitmap in trimmed_plan.items():
    per_adapter_keys = bitmap.gather(request.keys)
    ...
    task_id = self._l2_adapters[adapter_idx].submit_load_task(
        per_adapter_keys, per_adapter_objs
    )
```

So all of a request's keys for a given adapter are serialised. Concurrency exists only
*across adapters*, which does nothing for the common single-`fs`-adapter deployment.

## Measurements

`lmcache bench l2` on XFS over a 4-drive NVMe RAID-0. The experiment holds total work
constant and moves concurrency from *within* a batch to *across* batches, so the only
variable is where the parallelism sits.

**24 MiB keys** (a real KV chunk: Qwen3-Coder-30B-A3B, `chunk_size: 256`, TP=1 → 98,304
B/token × 256 = 24 MiB), 32 keys total, `use_odirect: true`:

| Layout | Lookup | Store | **Load** |
|---|---:|---:|---:|
| 32 keys, 1 batch | 2.43 ms | 82.84 ms | **68.75 ms** (11,202 MB/s) |
| 4 keys × 8 batches | 4.79 ms | 79.11 ms | **31.60 ms** (24,308 MB/s) |

**Load is 2.2× slower** when the same 32 keys arrive as one batch instead of eight.

Store is *not* materially affected at this size (1.05×) — it appears write-bandwidth-bound
on the device, so the serialisation is masked. Lookup is too small here to measure
meaningfully. **The defect is in all three code paths, but only Load demonstrably suffers
at this size.**

The mechanism is confirmed by a batch-size sweep at `--in-flight 1` (256 KB keys), where
duration scales linearly with batch size — the signature of a serial loop:

```
--num-keys      1      4      8     16     32     64    128
duration     0.34   1.87   6.63   9.02  16.00  25.91  46.16   ms
```

128× the keys, 136× the duration.

### Reproducing

```bash
# one batch of 32
lmcache bench l2 --l2-adapter '{"type":"fs","base_path":"/mnt/nvme/l2bench","use_odirect":true}' \
  --num-keys 32 --in-flight 1 --data-size-kb 24576 --l1-align-bytes 4096 --rounds 3

# eight batches of 4 -- identical total work
lmcache bench l2 --l2-adapter '{"type":"fs","base_path":"/mnt/nvme/l2bench","use_odirect":true}' \
  --num-keys 4  --in-flight 8 --data-size-kb 24576 --l1-align-bytes 4096 --rounds 3
```

Note `lmcache bench l2` needs the `native_storage_ops` extension (it imports `Bitmap`);
`NO_GPU_EXT=1 python setup.py build_ext --inplace` is enough to get it.

## Expected behavior

A batch API should exploit the batch. Loading N keys should overlap the N I/Os, as the
other adapters already do:

- `nixl_store_l2_adapter` builds one vectorised descriptor list and issues a single
  transfer;
- `raw_block_l2_adapter` uses dedicated `ThreadPoolExecutor`s (`_load_pool`, `_store_pool`,
  `_lookup_pool`);
- `s3_l2_adapter` gathers per-key futures with `asyncio.gather`.

`fs` is the odd one out, and it is the default file-backed adapter.

The fix looks small: build the per-key coroutines and `await asyncio.gather(*tasks)`
instead of awaiting in the loop, with a bound on in-flight I/O so a large batch cannot
swamp the executor. Worth noting from the data that unbounded is not automatically better
— at 24 MiB keys, `1 key × 32 batches` measured *worse* than `4 keys × 8` (65.72 ms vs
33.37 ms), so a concurrency limit belongs in the design rather than a bare `gather`.

## Caveats

- **Buffered numbers are page-cache reads, not disk.** The same comparison with
  `use_odirect: false` shows a much larger spread (121.83 ms vs 19.09 ms, 6.4×), but the
  bench writes then reads 768 MiB on a host with ample RAM, so those figures are memory
  bandwidth. The O_DIRECT row above is the one that reflects storage, and 2.2× is the
  figure I would rely on.
- 3 measurement rounds, 1 warmup, single host. Enough to establish the effect and its
  rough size, not a precise characterisation.
- The `1 key × 32` O_DIRECT anomaly noted above suggests executor or queue-depth limits at
  high concurrency; I have not investigated it.
- I have not measured the end-to-end TTFT impact in a live MP deployment — only the
  adapter in isolation.

## Environment

- LMCache `v0.5.1.dev52`, `upstream/dev`
- Ubuntu 26.04, kernel 7.0.0-28-generic
- torch 2.11.0+cu130, CUDA 13.0
- XFS on `/dev/md0`, software RAID-0 over 4 × Dell PM1735a MU 1.6 TB NVMe

## Related

Same defect class as the in-process `LocalDiskBackend`, whose `batched_get_blocking`
defaulted to a serial `for key in keys: get_blocking(key)` and was given a thread pool
(#3961, hardened further in #4348). The MP interface removed that particular footgun — it
is batch-native, with no single-key `get` to loop over and a `Bitmap` result for partial
success — but `FSL2Adapter`'s implementation reintroduces the behaviour internally.

Separately, #4364 proposes consolidating O_DIRECT handling and also concerns this file;
that is a different defect and a different fix.

xiangping-chen:
Fixes #4365.

`FSL2Adapter._execute_load` awaited each key's I/O before starting the next, so a batch of N keys cost N sequential round trips:

```python
for i, key in enumerate(keys):
    ...
    num_read = await self._loop.run_in_executor(   # awaited in-loop
        None, self._read_with_odirect, file_path, dst_buf,
    )
```

Dispatching to `run_in_executor` did not help, because the result is awaited immediately.

Nothing upstream masked it. `submit_load_task` schedules one coroutine per batch, and `prefetch_controller.py:931` puts all of a request's keys for an adapter into a single batch — so a long context serialised end to end. A 126,900-token request is ~496 chunks.

This gathers the per-key coroutines behind a bounded `asyncio.Semaphore`, using **the shape already established by `_execute_delete` in this same file** (which bounds unlink calls with `_DELETE_CONCURRENCY = 64`).

**Measurements** — `lmcache bench l2`, XFS over 4-drive NVMe RAID-0, 32-key batch of 24 MiB values (a KV chunk for a 30B model at `chunk_size: 256`, TP=1), `use_odirect: true`, `--in-flight 1`:

| Config | Duration | Throughput |
|---|---:|---:|
| `io_concurrency=1` (identical to pre-fix behaviour) | 70.03 ms | 10,978 MB/s |
| **default (`io_concurrency` omitted)** | **31.53 ms** | **24,362 MB/s** |

**2.22×**, and a single large batch now matches the same work split across eight batches (32.3 ms), which was the point.

The mechanism was confirmed independently by a batch-size sweep at `--in-flight 1`, where duration scaled linearly with batch size — 128× the keys gave 136× the duration.

**Choosing the default.** 16 comes from a sweep at the same value size rather than being picked by feel:

```
io_concurrency:   1      4      8     16     32
duration (ms): 67.3   37.0   32.2   31.7   32.0
```

The curve plateaus from ~8 onward with no measured penalty at 32, so 16 sits on the plateau with headroom. It is bounded rather than unbounded because batch size is chosen by the caller, so an unbounded fan-out would put an arbitrary number of large transfers in the executor at once. Smaller values prefer more concurrency (256 KiB values were still improving at 32-way), so the bound is exposed as the `io_concurrency` adapter config key.

- **Only the load path is changed.** The store and lookup paths have the same serial shape, but store measured as write-bandwidth-bound at this value size (82.8 ms vs 79.1 ms split across batches, 1.05×) so the serialisation is masked there, and lookup is too small to measure. I would rather not change them on the strength of a code reading alone; happy to if you would prefer the consistency, or as a follow-up once there is a case where they measurably matter.

- **The tests assert the observable contract, not wall-clock speed.** A timing assertion would be flaky in CI. They instead observe how many loads are in flight at once via a probe around `_read_with_odirect`: keys in a batch must overlap, and the overlap must not exceed `io_concurrency`. `io_concurrency=1` is pinned as serial — that case *is* the pre-fix behaviour, which is how the contrast is asserted inside the fixed code.

- One limitation worth stating: the new test module cannot run against un-fixed code, because it imports `_DEFAULT_IO_CONCURRENCY` and uses the new kwarg, so pre-fix it is a collection error rather than a clean failure. `test_io_concurrency_one_is_serial` compensates.

- Data correctness under concurrency is covered explicitly (`test_load_data_is_correct_when_concurrent` gives each key a distinct payload, so an index/buffer mix-up would surface as wrong bytes rather than a failed load), as is partial-batch reporting with keys now completing out of order.

- 57 tests pass: 10 new plus all 47 existing `fs_l2_adapter` tests (`test_fs_l2_adapter_delete.py`, `test_fs_l2_adapter_keys.py`). Note `lmcache bench l2` needs the `native_storage_ops` extension; `NO_GPU_EXT=1 python setup.py build_ext --inplace` is enough to get it.

- For context on where `fs` sits: `nixl_store` already issues one vectorised transfer per batch, `raw_block` uses dedicated `ThreadPoolExecutor`s, and `s3` gathers per-key futures. `fs` was the outlier, and it is the default file-backed adapter.

**If applicable**:


xiangping-chen:
**End-to-end TTFT numbers, since the adapter microbenchmark in the description only measured the adapter in isolation.**

Full MP-mode deployment (`lmcache server` + `LMCacheMPConnector` + vLLM 0.23.0), `fs` L2 adapter, single 96 GiB GPU, Qwen3-Coder-30B-A3B at TP=1, KV store on XFS over 4-drive NVMe RAID-0. The two arms differ **only** in `io_concurrency`; `io_concurrency=1` reproduces the pre-fix serial path exactly.

Hot TTFT, median of 5 passes (ms):

| Context (tokens) | `io_concurrency=1` (pre-fix) | default 16 (post-fix) | Speedup |
|---:|---:|---:|---:|
| 4,096 | 92 | 92 | 1.00x |
| 16,384 | 761 | **339** | **2.24x** |
| 32,768 | 2107 | **627** | **3.36x** |
| 65,536 | 2848 | **1132** | **2.52x** |
| 126,900 | 5544 | **2309** | **2.40x** |

**4,096 shows no gain, which is the expected shape.** At 16 chunks the request is dominated by everything other than storage, so serialisation is not the bottleneck. The defect only bites once the batch is large enough — matching the crossover seen in the batch-size sweep.

The end-to-end gain (2.2-3.4x) exceeds the 2.2x adapter-level figure because the microbenchmark used 32-key batches, whereas 126,900 tokens is ~496 chunks per request, so the serial penalty compounds further at realistic batch sizes.

### Validity

Both arms were verified to be measuring the intended thing:

```
token counts:    requested == actual, exactly, at all 5 sizes (tokenizer-built prompts)
L2 traffic:      arm a  L1=64  L2=4491
                 arm b  L1=64  L2=4619      -> ~99% of reads served by the fs adapter
store failures:  0 in both arms
alloc failures:  0 in both arms
error lines:     0 in both arms
data written:    23 GB / 959 files, identical in both arms
```

Hot < cold at every size, so no thrashing. An earlier attempt at `--l1-size-gb 8` was discarded because L1 exhaustion (`Failed to batched allocate 457 memory blocks of size 25165824`) made hot slower than cold at 32,768; this run uses 48 GiB and shows zero allocation failures.

### Caveats

- One run per arm. The deltas (124-236%) are far outside the ~1.7% run-to-run noise measured on this hardware, but the individual figures are not precision numbers.
- **Buffered path, not O_DIRECT.** O_DIRECT stores currently fail with `EINVAL` on this adapter (`os.write` on an O_DIRECT fd; see LMCache/LMCache#4364), so L2 never populates with `use_odirect: true` and there would be nothing to measure. The serial loop this PR fixes is identical on both paths, so the fix is what is being measured, but O_DIRECT numbers are unavailable until that separate defect is addressed.
- Not comparable with in-process-mode TTFT figures: different vLLM version, different prompt construction, and MP adds IPC. The arms are only comparable to each other.
- Arm assignment was verified from the driver script rather than from server logs, because the adapter did not log `io_concurrency`. The second commit on this branch adds it to the init line so future runs can be confirmed from logs alone.

xiangping-chen:
Fixed the `k3-unit-tests` failure. Root cause was in the test, not in `fs_l2_adapter.py` itself.

**Symptom** (Buildkite):
```
FAILED tests/v1/distributed/test_fs_l2_adapter_load_concurrency.py::test_load_batch_runs_keys_concurrently
AssertionError: every key should load
assert False
 +  where False = all([False, False, False, False, False, False, ...])
```

**Root cause**: the `_Buf` test double backed `byte_array` with a plain `bytearray`. O_DIRECT constrains the buffer *address*, not just the transfer length, and a `bytearray`'s address comes from the general-purpose allocator with no alignment guarantee — the exact defect class this whole thread (#4364/#4367) has been about, just relocated into my own test.

`_read_with_odirect` returns `0` on any exception rather than raising, by design — a real refusal degrades a key to a cache miss instead of failing the whole batch. That's why this surfaced as "every key should load" at the load-side assertion rather than as a raised error nearer the actual `EINVAL`.

It passed reliably in my own development (NVMe/XFS) because same-sized small allocations happened to land aligned there — allocator luck, not a guarantee. Buildkite's runner has a different allocator arena layout and consistently didn't.

Confirmed rather than assumed: I forced a `bytearray`-backed buffer 3 bytes past a page boundary and got the identical failure, every time, on the same host that otherwise passes reliably:
```
FAILED test_load_batch_runs_keys_concurrently - AssertionError: every key should load
FAILED test_io_concurrency_one_is_serial
FAILED test_load_concurrency_is_bounded_by_config
```
(the three tests that set `use_odirect=True`; the two that don't were unaffected either way.) Swapping `_Buf` to an anonymous `mmap`-backed buffer — which the kernel always returns page-aligned — passed every time under the same forced offset, and 8/8 clean runs on the unmodified buffer too.

Fixed in `d828eb61...42ef7cfa` (rebased onto the branch's merge-from-`dev` commit). `tests/v1/distributed/test_l1_memory_manager.py`, `test_fs_l2_adapter_delete.py` and `test_fs_l2_adapter_keys.py` are unaffected — 57 passed, 23 skipped, repeated 8x.

PR #4367, opened 2026-07-31
[Fix] Align LazyMemoryAllocator's buffer base to align_bytes

xiangping-chen:
Related to #4364.

`LazyMemoryAllocator` backs its arena with `torch.empty()`, which guarantees only 64-byte alignment. `TensorMemoryAllocator` then hands out objects at `base + k * align_bytes`, so every consumer's actual alignment is the alignment of that base — meaning `L1MemoryManager.get_l1_memory_desc()` advertised `align_bytes=4096` while delivering a pointer at `page + 64`. `--l1-use-lazy` defaults to `True`, so this was the default behavior. The non-lazy path is unaffected (`MixedMemoryAllocator` uses `cudaHostAlloc`/`mmap`, both page-aligned).

This matters because `align_bytes` is a correctness contract, not a hint:
- `raw_block_l2_adapter` validates it, but the check passes while the delivered pointer is still misaligned — it survives only because the Rust layer bounce-buffers at runtime.
- The NIXL adapters validate an *offset from the base*, so an unaligned base is invisible to them.
- O_DIRECT and GPUDirect need absolute alignment.

**The fix**: over-allocate `align_bytes - 1` and slice to an aligned offset — the same technique `_alloc_page_aligned_pinned_view` in `lmcache/v1/platform/torch_ops.py` already uses, just never applied to the lazy path. Also rejects a non-power-of-two `align_bytes` (the arithmetic assumes it), matching `DevDaxMemoryAllocator`'s existing check.

**Note on the O_DIRECT failure that led here**: this fixes the alignment contract, which stands on its own — a live MP deployment showed every O_DIRECT store fail with `EINVAL` under lazy L1 and zero failures under eager L1. I could not isolate the exact kernel-level mechanism (an isolated write with the same 64-byte misalignment and length succeeded in several separate attempts on the same host), so please read this as a contract fix that also resolves the observed failure, not a fully characterized fix for that specific error.

**Testing**: 3 new tests in `TestLazyMemoryAllocator` (base alignment across several `align_bytes` values, per-object address alignment, rejection of non-power-of-two `align_bytes`) — all fail without this change. One assertion added to `test_get_l1_memory_desc_size_matches_config`, which previously checked only the advertised value, not the delivered pointer. 86 passed / 3 skipped across `test_memory_management.py`, `test_l1_memory_manager.py`, `test_memory_allocator_import_surface.py`.

**If applicable**:


zhengfeihe:
Oh, this is memory allocation difference between PyTorch and cudaHostAlloc/mmap 

Change looks fine to me. Thanks for the fix! One ask: could you shorten/consolidate the PR description and the in-code comment block? A brief constraint comment can reduce a lot of reviewer context overhead.

xiangping-chen:
Trimmed both — PR description down from 788 to 341 words, and the in-code comment block in `lazy_memory_allocator.py` from 17 lines to 9 (`595b7037`). Kept the load-bearing why in each; dropped the restated mechanism and the duplicated wording in the error message. Re-ran the alignment tests plus the wider regression sweep (`test_memory_management.py`, `test_l1_memory_manager.py`, `test_memory_allocator_import_surface.py`) after the trim — no logic changed, 86 passed / 3 skipped.

KuntaiDu (approved):
LGTM!

KuntaiDu:
@ApostaC @sammshen @YaoJiayi PTAL

lmcache/v1/memory_allocators/lazy_memory_allocator.py
```diff
@@ -82,8 +82,22 @@ def __init__(
         Args:
             init_size (int): Initial size of the memory allocation in bytes.
             final_size (int): Final size of the memory allocation in bytes.
-            align_bytes (int, optional): Alignment in for the underlying allocations
+            align_bytes (int, optional): Alignment for the underlying allocations.
+                Must be a positive power of two. The buffer's base address is
+                aligned to this value, not merely the offsets within it, because
+                consumers reached via ``get_l1_memory_desc()`` rely on the
+                absolute alignment of the pointers they receive.
```
ApostaC: nit: these two sentences can be removed.

lmcache/v1/memory_allocators/lazy_memory_allocator.py
```diff
@@ -82,8 +82,22 @@ def __init__(
...
+
+        Raises:
+            ValueError: If ``align_bytes`` is not a positive power of two.
+            RuntimeError: If the platform does not support memory pinning, or if
+                the allocated buffer could not be aligned to ``align_bytes``.
         """
+        # Same requirement, and same wording, as DevDaxMemoryAllocator: the
+        # base-alignment arithmetic below assumes a positive power of two.
```
ApostaC: nit: please remove this comment. We should not mention other modules here.

lmcache/v1/memory_allocators/lazy_memory_allocator.py
```diff
@@ -109,8 +123,29 @@ def __init__(
             buf = arr_type.from_address(ptr)
             self._buffer = torch.frombuffer(buf, dtype=torch.uint8)
         else:
-            self._buffer = torch.empty(
-                self._final_size, dtype=torch.uint8, device="cpu", pin_memory=False
+            # torch.empty() only guarantees 64-byte alignment, but consumers of
+            # get_l1_memory_desc() (O_DIRECT, RDMA/GDS) need the buffer base
+            # itself aligned to align_bytes. Over-allocate and slice to an
+            # aligned offset, matching _alloc_page_aligned_pinned_view in
+            # lmcache/v1/platform/torch_ops.py.
```
ApostaC: nit: remove the sentence `Over-allocate and slice to an ...`

ApostaC (commented):
Thanks a lot for your contribution. Just a few nit comments. Otherwise LGTM!

lmcache/v1/memory_allocators/lazy_memory_allocator.py
xiangping-chen: Removed both sentences.

lmcache/v1/memory_allocators/lazy_memory_allocator.py
xiangping-chen: Removed -- no other modules referenced now.

lmcache/v1/memory_allocators/lazy_memory_allocator.py
xiangping-chen: Removed that sentence.

xiangping-chen:
Addressed all three nits from @ApostaC's review in `9e1b6ee4`:

- Removed the two docstring sentences about `get_l1_memory_desc()` consumers.
- Removed the comment referencing `DevDaxMemoryAllocator` by name.
- Removed the `_alloc_page_aligned_pinned_view` sentence restating the code below it.

No logic changed. Re-verified after the edit: 86 passed / 3 skipped across `test_memory_management.py`, `test_l1_memory_manager.py`, `test_memory_allocator_import_surface.py` -- same as before the trim.

ApostaC (approved):
LGTM!

PR #4370, opened 2026-08-01
feat: implement Python fallbacks for hybrid-model KV transfer operations

hlin99:
## Motivation

-- THIS IS TO BUILD FUNDAMENTAL INFRASTRUCTURE FOR HYBRID-MODEL ON ENGINE-DRIVEN PATH !

## Summary

Implement the missing Python fallbacks required for engine-driven hybrid-model support.
- Add fallback plan descriptors and executors for object-group transfers and flattened CacheBlend retrieval.
- Add fallback support for HND, fused KV, blocked-scale, and MLA KV-cache layouts.
- Add strided RoPE execution and alignment-aware async memcpy handling.
- Route CUDA through the Python fallback path for CI. (this change to be reverted before merging)

## Tests
- Add round-trip tests for split-KV and heterogeneous fused-KV object-group transfer plans.
- Add CacheBlend retrieval-plan tests, including RoPE and scatter behavior.


tests/v1/test_c_ops_parity.py
```diff
@@ -213,22 +213,13 @@ def _has_real_names(params):
 # These are native-only with no torch fallback; call sites guard with hasattr.
 # NOTE: execute_object_group_transfer previously had a signature-only stub
 # (raised NotImplementedError unconditionally) — not a real fallback.
-_EXCLUDED_FUNCS: set[str] = {
-    "execute_object_group_transfer",
-    "execute_cb_retrieve_plan_flat",
-}
+_EXCLUDED_FUNCS: set[str] = set()
```
hlin99, line 226: add some comments: this exclude list is just for temp CI w/a. python fallbacks are still needed to be implemented.

hlin99: hi @maobaolong  @chunxiaozheng  added comments to illustrate the purpose of the exclude lists as we discussed offline. please help take a look. thanks

hlin99:
all the CI using cuda + python fallback ops passed as shown. Will revert the hack way to avoid merging by accident
image

lmcache/v1/platform/torch_ops.py
```diff
@@ -846,31 +1036,24 @@ def _per_layer_paged_shape(
         A tuple representing the shape needed to reconstruct one layer's tensor
         from a raw pointer via :func:`_tensor_from_ptr`.
     """
-    fmt = int(engine_kv_format)
-    if fmt == int(EngineKVFormat.NL_X_NBBS_ONE_HS):
-        return (nb * bs, 1, hs)
-    if fmt == int(EngineKVFormat.NL_X_NB_BS_HS):
+    if is_mla(engine_kv_format):
```
maobaolong, line 1063: This looks much better than before. 😄

lmcache/v1/platform/torch_ops.py
```diff
@@ -2188,56 +2823,43 @@ def lmcache_memcpy_async(
...
         return
 
-    # 4. Pointer mode
-    if not isinstance(dest, int) or not isinstance(src, int):
-        raise TypeError(
-            "dest and src must be both int (pointer mode) "
-            "or both torch.Tensor (tensor mode)"
+    # The direction identifies the device on which a raw pointer resides.
+    # Without CUDA, both pointers are CPU addresses in the fallback path.
+    use_cuda = torch.cuda.is_available()
```
maobaolong: [P1] Do not infer a raw pointer device from CUDA availability. Adding these fallback methods makes the existing hasattr-based fast-path guards true on XPU/MUSA/HPU as well, while the production planners pass integer data_ptr values. When CUDA is unavailable, this code treats an XPU/MUSA device pointer as a CPU address and may dereference it through ctypes; when CUDA is available, it can misclassify it as CUDA. The XPU tests currently avoid this by passing Tensor descriptors, unlike production. Please retain the backend/device in the plan and dispatch staging through DeviceOps (so native XPU/MUSA memcpy overrides are honored), or make fallback planners carry Tensor objects instead of raw device pointers. Please add a test using the same integer-pointer representation as the production planner.

lmcache/v1/platform/torch_ops.py
```diff
@@ -670,6 +769,52 @@ def multi_layer_kv_transfer(
...
+            if int(direction) == int(TransferDirection.D2H):
+                key_value[:, layer_id, valid_mask_kv, :] = lmc_valid.reshape(
+                    2, -1, hidden_size
+                ).to(kv_device, non_blocking=False)
+        elif is_fused_packed_hnd or is_fused_packed_nhd:
+            # Fused K/V layouts have one chunk plane. HND stores heads
+            # before tokens within a block; NHD already has token-major
+            # ordering once its first two dimensions are flattened.
+            paged_logical = (
+                paged_tensor.permute(0, 2, 1, 3).reshape(page_buffer_size, hidden_size)
```
maobaolong: [P1] Avoid flattening the permuted fused-HND destination this way. For normal shapes with both block_size > 1 and num_heads > 1, permute(0, 2, 1, 3).reshape(...) materializes a contiguous copy. The following index_copy_ therefore updates the copy rather than paged_tensor, so H2D/CacheBlend retrieval silently leaves the real KV cache unchanged. Please scatter directly into paged_tensor using block_indices and block_offsets (analogous to the split-HND branch), and add an H2D test with multiple heads and multiple tokens per block.

lmcache/v1/platform/torch_ops.py
```diff
@@ -607,26 +643,77 @@ def multi_layer_kv_transfer(
...
+            if is_hnd_split_blocks_first
+            else (2, num_blocks, num_heads, block_size, head_size)
+        )
+    elif is_fused_packed_hnd:
+        num_blocks = page_buffer_size // block_size
+        num_heads = hidden_size // (2 * head_size)
+        layer_shape = (num_blocks, num_heads, block_size, 2 * head_size)
+    elif is_fused_packed_nhd:
+        num_blocks = page_buffer_size // block_size
+        num_heads = hidden_size // (2 * head_size)
```
maobaolong: [P1] Preserve the native/default behavior for fused NHD. head_size is a public optional argument with default 0, and the validation above only requires it for HND, but this branch divides by 2 * head_size. A fused-NHD call that relies on the documented default therefore raises ZeroDivisionError, while the native NHD path does not require head_size. Please construct/flatten the NHD view without requiring head_size (or otherwise preserve native parity) and add a regression test with head_size omitted.

maobaolong (changes requested):
Thanks for adding these Python fallbacks. I found three correctness blockers that should be addressed before merge: the production raw-pointer plan is unsafe on non-CUDA accelerators, fused-HND H2D scatter can write into a materialized copy, and fused-NHD divides by the default head_size=0. Please also add tests that exercise the production representation (integer data_ptr values) on non-CUDA backends, plus direct H2D/D2H coverage for split-HND, fused-HND/NHD, and blocked-scale layouts.

lmcache/v1/platform/torch_ops.py
hlin99: fixed by [5948c3b](https://github.com/LMCache/LMCache/pull/4370/commits/5948c3bdd6aafdcbd1f2cbdaa746ce9a25798e42)

lmcache/v1/platform/torch_ops.py
hlin99: fixed by [25e8fe1](https://github.com/LMCache/LMCache/pull/4370/commits/25e8fe102cf98494ee866ee511ffbe24bcc5d47a)

lmcache/v1/platform/torch_ops.py
hlin99: This is the known issue that I mentioned. right now, passing ptr from upper layer and do ptr->tensor later in the ops is not a safe way.
We can fundamentally resolve the issue in step#3, which is passing tensor from upper layer and do tensor->ptr as needed. So in this PR, will leave it as it is.

hlin99:
hi @maobaolong the ptr->tensor issue is a known issue that will be addressed by step#3 of https://github.com/LMCache/LMCache/issues/4465 .
the other two were fixed, and all CI showed green on cuda device with torch ops (hack way). please have a look. I'll revert the hack before merging.

lmcache/v1/platform/torch_ops.py
```diff
@@ -159,6 +164,84 @@ def _tensor_from_ptr(
...
+        if byte_tensor.data_ptr() != value:
+            raise RuntimeError(
+                "Failed to create an aliasing tensor view for the raw pointer"
+            )
+    return byte_tensor
+
+
+def _device_for_raw_pointer(ptr: int) -> torch.device:
+    """Return a CUDA device for a pointer owned by PyTorch's CUDA allocator."""
+    if not torch.cuda.is_available():
```
maobaolong: Should this kind of usage be banned?

hlin99: done

lmcache/v1/platform/cuda/device_ops.py
```diff
@@ -23,17 +23,6 @@
 class CudaDeviceOps(DeviceOps):
     device_type: ClassVar[str] = "cuda"
 
-    def ensure_native(self) -> None:
```
maobaolong, line 26: Just heads-up, revert this before merge.

ApostaC:
Hey Tony @hlin99 , just to give you some context on `StagingCopy`, `LaunchVar`, `BatchStep`.

Those classes are introduced just for the CUDA "fast path optimizations" and are subject to change. Therefore, it's more recommended to keep them as CUDA-specific classes.

Hybrid model support does not need those classes. You can take a look at `transfer_kv_per_object_group` function in `lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py` (the non-fast-path). Hopefully this information will help 🙏

hlin99:
> Hey Tony @hlin99 , just to give you some context on `StagingCopy`, `LaunchVar`, `BatchStep`.
> 
> 
> 
> Those classes are introduced just for the CUDA "fast path optimizations" and are subject to change. Therefore, it's more recommended to keep them as CUDA-specific classes.
> 
> 
> 
> Hybrid model support does not need those classes. You can take a look at `transfer_kv_per_object_group` function in `lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py` (the non-fast-path). Hopefully this information will help 🙏

sure. Thanks for sharing @ApostaC   I will take a look

PR #4375, opened 2026-08-01
[core][bugfix] fs connector: classify O_DIRECT refusals and check buffer alignment

nosey-dewdrop:

Refs #4364. This is site 9 of the nine O_DIRECT implementations listed there,
`csrc/storage_backends/fs/connector.cpp`, which drives the MP `fs_native`
adapter. The issue notes a Python helper is not applicable to this file, so
this is the C++ counterpart. It does not touch the other eight sites.

Two defects from that report are present here:

1. **No errno classification.** The connector asks the platform with
   `#ifdef O_DIRECT` and then opens. Nothing inspects `EINVAL`,
   `EOPNOTSUPP`, `ENOTSUP` or `EPERM`, so a refusal becomes a
   `std::runtime_error` and the key fails. None of those errnos means the data
   is unavailable, so each should degrade to buffered I/O.

2. **Alignment gated on length only.**

   ```cpp
   bool aligned = conn.disk_block_size > 0 && len % conn.disk_block_size == 0;
   ```

   O_DIRECT also constrains the buffer address. A length aligned host buffer at
   an unaligned address passes this gate, `open()` accepts the flag, and
   `write()` answers `EINVAL`. That is the shape reproduced on XFS in
   https://github.com/LMCache/LMCache/issues/4364#issuecomment-5146560980, where
   57 stores failed and the L2 tier held 0 files while reporting healthy.

**Changes**

- Check the buffer address as well as the transfer length before requesting
  O_DIRECT.
- Classify the four refusal errnos and fall back to buffered I/O rather than
  failing the key, both when `open()` refuses the flag and when the first
  `read()` or `write()` refuses it after a successful open.
- Latch the refusal on the connector so the other workers stop paying for an
  attempt known to fail on this path, and log it once instead of per file.
- Carry `errno` on I/O errors so a refusal can be told apart from a real fault.

- Scope is this file only. I have not touched the block size question, which is
  the third defect in #4364. This still uses `statvfs` `f_bsize` rather than the
  device logical block size. Using the larger value costs a missed O_DIRECT
  rather than an `EINVAL`, so it is a separate change.
- **The O_DIRECT path itself is not verified on my machine.** I develop on
  macOS, which has no `O_DIRECT`, so locally the new test exercises only the
  buffered round trip and says so in its output. The refusal branches are
  reasoned from the reproduction in the issue, not observed here. @xiangping-chen,
  you have the XFS over NVMe host and already reproduced the write time
  `EINVAL`, would you be willing to run this branch against it?
- The test places an 8 KiB payload in a buffer 512 bytes past a 4096 byte
  boundary, so the length gate passes and the address gate does not. It follows
  the `tests/v1/shm_allocator/csrc` pattern, CMake builds the binary and pytest
  runs it, skipping when no toolchain is present.
- I have a separate PR open on the shared connector base, #4374. The two do not
  overlap.

**If applicable**:

### DIFF
--- a/csrc/storage_backends/fs/connector.cpp
+++ b/csrc/storage_backends/fs/connector.cpp
@@ -2,12 +2,88 @@
 
 #include "connector.h"
 #include <cerrno>
+#include <cstdint>
 #include <cstdio>
 #include <stdexcept>
 #include <string>
 
 namespace lmcache {
 namespace connector {
+
+namespace {
+
+// Carries the errno alongside the message. A caller cannot classify an
+// O_DIRECT refusal from strerror text alone.
+struct IoError : std::runtime_error {
+  IoError(const std::string& what, int errnum)
+      : std::runtime_error(what), err(errnum) {}
+  int err;
+};
+
+// Errnos meaning this file or filesystem will not serve direct I/O.
+// open() is documented to answer EINVAL, a filesystem may answer
+// EOPNOTSUPP, and EPERM is reachable under policy. None of them says the
+// data is unavailable, so each should degrade to buffered I/O rather than
+// fail the request.
+bool is_direct_io_refusal(int err) {
+  if (err == EINVAL || err == EPERM) return true;
+#ifdef EOPNOTSUPP
+  if (err == EOPNOTSUPP) return true;
+#endif
+#ifdef ENOTSUP
+  if (err == ENOTSUP) return true;
+#endif
+  return false;
+}
+
+// O_DIRECT constrains the file offset, the transfer length and the buffer
+// address. Every transfer here starts at offset zero, so the other two are
+// checked. Length alone is not enough, because a pinned host allocator can
+// return a block sized region at an address that is not block aligned.
+bool odirect_alignment_ok(size_t block_size, size_t len, const void* buf) {
+  if (block_size == 0) return false;
+  if (len % block_size != 0) return false;
+  return reinterpret_cast<uintptr_t>(buf) % block_size == 0;
+}
+
+// Opens with O_DIRECT when asked, and falls back to a buffered open when the
+// flag is refused. Reports which one the caller got, since that decides
+// whether the transfer has to stay aligned.
+int open_maybe_direct(const char* path, int flags, mode_t mode,
+                      bool want_direct, bool* opened_direct) {
+  *opened_direct = false;
+#ifdef O_DIRECT
+  if (want_direct) {
+    int fd = ::open(path, flags | O_DIRECT, mode);
+    if (fd >= 0) {
+      *opened_direct = true;
+      return fd;
+    }
+    if (!is_direct_io_refusal(errno)) {
+      return -1;
+    }
+  }
+#else
+  (void)want_direct;
+#endif
+  return ::open(path, flags, mode);
+}
+
+bool odirect_already_refused(const WorkerFSConn& conn) {
+  return conn.odirect_refused != nullptr &&
+         conn.odirect_refused->load(std::memory_order_relaxed);
+}
+
+void note_odirect_refused(WorkerFSConn& conn) {
+  if (conn.odirect_refused == nullptr) return;
+  if (!conn.odirect_refused->exchange(true, std::memory_order_relaxed)) {
+    fprintf(stderr,
+            "[LMCache FS] O_DIRECT refused on this path, continuing with "
+            "buffered I/O\n");
+  }
+}
+
+}  // namespace
 
 // ---------------------------------------------------------------
 // Helpers
@@ -95,7 +171,7 @@ static void write_all(int fd, const void* data, size_t len) {
     ssize_t n = ::write(fd, ptr + written, len - written);
     if (n < 0) {
       if (errno == EINTR) continue;
-      throw std::runtime_error("write failed: " + std::string(strerror(errno)));
+      throw IoError("write failed: " + std::string(strerror(errno)), errno);
     }
     if (n == 0) {
       throw std::runtime_error("write returned 0");
@@ -111,12 +187,39 @@ static size_t read_all(int fd, void* buf, size_t len) {
     ssize_t n = ::read(fd, ptr + total, len - total);
     if (n < 0) {
       if (errno == EINTR) continue;
-      throw std::runtime_error("read failed: " + std::string(strerror(errno)));
+      throw IoError("read failed: " + std::string(strerror(errno)), errno);
     }
     if (n == 0) break;  // EOF
     total += static_cast<size_t>(n);
   }
   return total;
+}
+
+// Reads exactly len bytes, optionally issuing a small leading read first so
+// the filesystem starts reading ahead. Shared by the direct and the buffered
+// attempt so both paths behave identically.
+static void read_whole_file(int fd, void* buf, size_t len,
+                            size_t read_ahead_size,
+                            const std::string& file_path) {
+  size_t n;
+  if (read_ahead_size > 0 && len > read_ahead_size) {
+    size_t n_head = read_all(fd, buf, read_ahead_size);
+    if (n_head < read_ahead_size) {
+      n = n_head;  // short read on the head, treat as incomplete
+    } else {
+      size_t n_tail =
+          read_all(fd, static_cast<char*>(buf) + read_ahead_size,
+                   len - read_ahead_size);
+      n = n_head + n_tail;
+    }
+  } else {
+    n = read_all(fd, buf, len);
+  }
+  if (n != len) {
+    throw std::runtime_error("incomplete read for " + file_path + ": expected " +
+                             std::to_string(len) + ", got " +
+                             std::to_string(n));
+  }
 }
 
 // ---------------------------------------------------------------
@@ -163,6 +266,7 @@ WorkerFSConn FSConnector::create_connection() {
   conn.use_odirect = use_odirect_;
   conn.disk_block_size = disk_block_size_;
   conn.read_ahead_size = read_ahead_size_;
+  conn.odirect_refused = &odirect_refused_;
   return conn;
 }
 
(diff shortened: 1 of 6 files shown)


xiangping-chen:
Thanks for taking site 9 — reviewed the diff and ran it on the XFS-over-NVMe-RAID-0 host from the #4364 reproduction, since you asked.

## Summary

**Correctness: the core fix is real and I can confirm it with a genuine reproduction.** The regression test as written, however, does not exercise the fix on this hardware — it passes identically with and without the fix, for a reason worth digging into. Details below, plus one runtime-refusal-path gap that's unverified either way.

## Confirmed: the address-alignment defect is real, on real hardware

Built `fs_odirect_fallback` against this branch and against the pre-PR `connector.cpp`/`.h`, on `/mnt/lmcache` (XFS on `/dev/md0`, the exact volume from the #4364 repro). The test as shipped (`kSkew = 512`) passes both before and after this PR:

```
pre-PR  connector, kSkew=512:  PASS
post-PR connector, kSkew=512:  PASS
```

That's a vacuous pass, not a confirmation. The reason: `/sys/block/md0/queue/logical_block_size` is **512** on this device, while `conn.disk_block_size` (`statvfs` `f_bsize`) is **4096**. A buffer skewed by exactly 512 bytes is misaligned relative to the FS's reported block size but *is* aligned relative to what the kernel actually enforces for O_DIRECT (the device's logical sector size). So the write succeeds via O_DIRECT on both trees, and the new alignment check and the new refusal-and-retry logic are never reached — no `[LMCache FS] O_DIRECT refused` line in `stderr` either way.

Rerunning with `kSkew = 4` (not a multiple of 512, genuinely misaligned against the real requirement) gives the result the PR wants to demonstrate:

```
pre-PR  connector, kSkew=4:  FAIL — "store failed: write failed: Invalid argument"
post-PR connector, kSkew=4:  PASS
```

So the underlying defect (length-only gating, per #4364) is real and this PR's `odirect_alignment_ok` genuinely fixes it — I just needed a different skew to see it on this device. **Suggest changing `kSkew` from `512` to something not a multiple of the real device sector size** (4, or anything not a multiple of 512) so the test fails on `dev` and passes on this branch on real hardware, not just in principle. Worth a comment noting *why* 512 doesn't work, too, since the next person to touch this test will otherwise reintroduce it.

## Unverified: the runtime refusal-and-retry path

Neither skew value exercises `is_direct_io_refusal` / `note_odirect_refused` / the `catch (const IoError&)` branches, because `odirect_alignment_ok` is evaluated *before* `open()` and both misalignment cases I tried are already caught there. On this kernel, `tmpfs` (`/dev/shm`) also now tolerates O_DIRECT, so I couldn't force an actual `open()`-time or `write()`-time refusal to hit the retry path live.

That doesn't mean the logic is wrong — reading it, it looks correct (fd is closed before the retry, the buffered reopen uses the same `O_TRUNC` flags so a partial direct write is safely overwritten, and the shared `atomic<bool>` latch is exchange-based so concurrent first-failures only log once). But it means this specific behavior is currently reasoned-about rather than exercised, on top of not being observed on real hardware either. Given #4359 already established a pattern for this on the Python side — monkeypatch/interpose the syscall to force a specific errno rather than relying on a filesystem that happens to refuse — I'd suggest the same here via `LD_PRELOAD` interposition of `open`/`write`/`read`, so the retry path is deterministically covered independent of what the underlying FS tolerates. Happy to be told this is already covered somewhere I didn't check.

## Other things checked, no issues found

- **Build**: builds clean with `-Wall -Wextra`; the only warnings (`unused parameter 'chunk_size'`) predate this PR.
- **Thread safety**: `odirect_refused_` is a plain `std::atomic<bool>` shared via pointer across all `WorkerFSConn` instances from one `FSConnector`; `exchange` in `note_odirect_refused` means concurrent first-failures log once, not once per worker. No lifetime issue — the atomic lives in the connector, which outlives the workers.
- **Partial-write safety**: retry reopens with the same `O_CREAT | O_WRONLY | O_TRUNC`, so a partially-written direct attempt is truncated and rewritten from scratch rather than corrupted or appended-to.
- **EPERM in the refusal set**: matches the `_DIRECT_IO_UNSUPPORTED_ERRNOS` set already used on the Python side (#4359), so this is consistent with existing project precedent rather than a new judgment call.
- **`tests/v1/storage_backends/__init__.py`**: confirmed this package doesn't already exist on `dev`, so no namespace-package collision.
- Not in scope, and correctly disclosed as such: `disk_block_size_` still comes from `statvfs` `f_bsize` rather than the device logical block size. That's exactly what made the `kSkew=512` test vacuous above, so it's worth linking these two together — the alignment fix is real, but as long as the block-size source stays at `f_bsize`, any test skew that's merely "not a multiple of 4096" risks landing on a value that's still a multiple of the real device sector, and passing for the wrong reason on the next device with a different sector size. A comment to that effect near `odirect_alignment_ok` might save someone a debugging session later.

## Verdict

The C++ fix looks correct and I could reproduce the defect it addresses. Before merge I'd want:
1. `kSkew` changed to a value that isn't a multiple of common device sector sizes (or derive it from the actual alignment requirement rather than a hardcoded 512), so the test isn't hardware-dependent-vacuous.
2. Either a note that the runtime refusal/retry path is unverified on real hardware, or (preferably) a fault-injected unit test so it's covered deterministically.

Everything else — thread safety, partial-write handling, errno set, scope discipline — held up. Happy to re-run once the test is adjusted.

nosey-dewdrop:
Thank you for running it on the real host — the vacuous pass is the finding here, and it's a better one than the bug it was hiding. Both points are addressed in `326d2ef`.

**1. `kSkew`**

Changed to an odd value rather than to 4, because odd is provable rather than empirical. Every real `logical_block_size` is a power of two, and an odd offset is never a multiple of a power of two greater than one, so the address is misaligned on any device instead of on the ones we happened to try. That closes the exact hole you flagged — that "not a multiple of 4096" can still land on a legal address for the next device with a different sector size.

Checked the arithmetic against every sector size a device can report, reproducing `skewed_buffer` standalone:

```
kSkew=512  ->  sector 512   LEGAL(vacuous)  1024 misaligned  2048 misaligned  4096 misaligned
kSkew=4    ->  sector 512   misaligned      1024 misaligned  2048 misaligned  4096 misaligned
kSkew=1    ->  sector 512   misaligned      1024 misaligned  2048 misaligned  4096 misaligned
```

The first row is your host, derived rather than assumed, which is why the test passed on both trees there.

The comment above `kSkew` now carries the why, including your `f_bsize` 4096 versus `logical_block_size` 512 measurement, so the next person doesn't put 512 back.

**2. `f_bsize` versus the device sector size**

Added the note near `odirect_alignment_ok` you suggested. Worth stating in both directions: for the check itself `f_bsize` is the conservative side, since it's a multiple of the sector size on mainstream setups and anything the check accepts the kernel accepts too. The gap only bites when *choosing* a deliberately misaligned address, which is precisely how this test ended up vacuous.

**3. The runtime refusal path**

You're right that no skew reaches it — `odirect_alignment_ok` runs before `open()`, so `is_direct_io_refusal`, `note_odirect_refused` and the `IoError` branches are unreachable from this test by construction, not by accident of filesystem. I took the note option for now and wrote it into the test header, naming interposition and #4359 as the way to cover it properly, so the gap is recorded at the place someone will read it rather than only in this thread.

I'd rather not push the interposer in this PR without running it. My machine is macOS, where `O_DIRECT` isn't defined and the entire direct path compiles out under the `#ifdef`, so I can't produce a red-then-green for it and would only be shipping a second reviewed-but-unexercised test into a PR whose review was about exactly that. Happy to open it as a follow-up against Linux CI, or to add it here if you'd prefer it blocking.

For completeness on what I could and couldn't verify locally: the alignment arithmetic above is measured, and both files compile clean with `-Wall -Wextra` (the two `unused parameter 'chunk_size'` warnings are the pre-existing ones you noted). The O_DIRECT behaviour itself is not runnable on this machine, so your reproduction remains the evidence for it.

xiangping-chen:
Re-reviewed at `326d2efd`. Both items from my earlier review are addressed, and I verified the test fix on the #4364 reproduction host rather than taking it on trust. Two new findings, one of which I think blocks merge.

## Verified

`kSkew = 1` now makes this a real regression test. Same host as before (XFS on NVMe RAID-0, `f_bsize` 4096, `logical_block_size` 512), same test binary, only `connector.{cpp,h}` swapped:

```
PRE-PR  connector, kSkew=1:  FAIL — "store failed: write failed: Invalid argument"
POST-PR connector, kSkew=1:  PASS
```

Previously both trees passed with `kSkew = 512`. Also confirmed: no new warnings under `-Wall -Wextra` (only the pre-existing `unused parameter 'chunk_size'`), the pytest wrapper passes end to end, and the backing buffer is still large enough (worst case ends at +12288 of a 16384 allocation).

The reasoning you used for the value is stronger than the `4` I suggested — an odd offset can't be a multiple of any power-of-two sector size greater than one, so it's device-independent rather than tuned to the one device that was tried. The `f_bsize` vs `logical_block_size` note at `odirect_alignment_ok` is accurate too: `f_bsize` being a multiple of the sector size does make the check conservative in the safe direction.

The documented gap on the runtime refusal path is what I asked for, and the pointer to #4359's interposition approach is the right next step.

## Issues

**error — user-facing doc now misrepresents behavior**

`docs/source/mp/l2_storage/fs_native.rst:45-52` documents the pre-PR contract:

> **Memory-buffer alignment.** ... If the buffer is misaligned, the underlying `read`/`write` syscall returns `EINVAL` (this is **not** caught by the length-fallback path above and will surface as a runtime error).

After this PR that is wrong in both halves: a misaligned buffer *is* caught (by the new address check in `odirect_alignment_ok`) and it does *not* surface as a runtime error — it degrades to buffered I/O, exactly like the length case the same paragraph contrasts it against. This PR touches no docs.

`fs_native` is the adapter driven by the connector you changed, so this is in scope even though the block-size question isn't. Suggested fix: fold caveat 2 into the same shape as caveat 1 — the connector now checks the buffer address as well as the length, and falls back to a buffered open when either is unaligned, so correctness is preserved but you silently don't get true direct I/O. `--l1-align-bytes` guidance stays as is. Worth checking `fs.rst` at the same time; I didn't verify whether its wording makes the same claim for the Python adapter, which this PR doesn't change.

**warning — alignment-disabled O_DIRECT is now silent**

`csrc/storage_backends/fs/connector.cpp:289-301` and `:355-366`. `note_odirect_refused` is the only diagnostic, and it is gated on `want_odirect && !opened_direct` (plus the `IoError` retry paths at `:313` and `:381`). When `odirect_alignment_ok` returns false, `want_odirect` is false, so nothing is logged — on every operation, for the life of the process.

Net effect for an operator who sets `use_odirect: true` with a misaligned buffer: before this PR they got a loud `EINVAL`; after it they get buffered I/O forever with no indication why. That is the right correctness trade-off and matches what #4364 asked for, but losing the signal entirely is a step back in diagnosability, and it is the case most likely to be hit in practice (a stale `--l1-align-bytes` against a large-block FS).

The pre-PR code didn't log on length misalignment either, so this isn't a regression you introduced — but you're adding a second silent path, and you already have the log-once machinery. A `note_odirect_unaligned`-style one-shot next to `note_odirect_refused`, latched the same way, would cover it cheaply. Reasonable to defer if you'd rather keep this PR to one concern; if so, please say so explicitly in the PR body so it isn't lost.

## Summary table

| Severity | Count | Key items |
|----------|-------|-----------|
| error    | 1 | `fs_native.rst:45-52` states misalignment surfaces as `EINVAL`; it now falls back to buffered |
| warning  | 1 | alignment-disabled O_DIRECT logs nothing (`connector.cpp:289-301`, `:355-366`) |
| info     | 0 | — |

Checked and clean: no new compiler warnings; regression test now discriminates pre/post on real hardware; buffer sizing safe; no signature or caller impact from `326d2efd` (comments plus a test-local `constexpr`); thread safety of the `odirect_refused_` latch unchanged from my earlier pass (`exchange`-based, log-once, owned by the connector which outlives its workers); scope still one file plus its test.

Happy to re-run the pre/post comparison on that host once the doc paragraph is updated.

PR #4381, opened 2026-08-01
tests: lazily construct shared memory_allocator fixture (#4295)

### LINKED ISSUE #4295: tests/conftest.py: session-scoped 5GB memory_allocator fixture always instantiated, adds ~9GB RSS to every test run
**Label**
bug, Testing, mem_leak, performance

`tests/conftest.py` defines a session-scoped fixture that eagerly pre-allocates a **5 GB pinned CPU memory buffer**, and an `autouse=True` function-scoped fixture that pulls it in as a dependency for *every single test in the repo*:

```python
# tests/conftest.py:784
@pytest.fixture(scope="session")
def memory_allocator():
    """One MixedMemoryAllocator (5GB) for the whole test session;
    .close() is a no-op per-test."""
    _real = MixedMemoryAllocator(5 * 1024 * 1024 * 1024)  # 5GB
    ...

@pytest.fixture(autouse=True)  # function-scoped by default
def use_shared_allocator(request, monkeypatch, memory_allocator):
    """Default: patch. Opt out with @pytest.mark.no_shared_allocator."""
    if request.node.get_closest_marker("no_shared_allocator"):
        # do NOT patch for this test
        yield
        return
    ...
```

Because `memory_allocator` is declared as a parameter of the `autouse` fixture, pytest instantiates it **before** the `no_shared_allocator` marker check even runs, so every test session pays this cost regardless of whether any test actually needs the shared allocator.

Measured the real-world cost directly:

```python
import resource
def rss(): return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024
print("before", rss())                                  # ~920 MB (torch import baseline)
alloc = MixedMemoryAllocator(5 * 1024 * 1024 * 1024)     # 5 GB requested
print("after", rss())                                    # ~8900 MB (+8 GB!)
```

The pinned-memory allocation path first-touches every page (`torch.empty(size, pin_memory=True)` + zero-fill), so the full 5 GB (plus overhead, ~8 GB observed RSS delta) is committed immediately and held for the entire pytest session — even for test files that never touch `LMCacheEngineBuilder` or the shared allocator at all (e.g. `tests/v1/storage_backend/test_scylla_backend.py`, which was the file that surfaced this while investigating a separate, unrelated memory leak in that test's own fixtures).

**To Reproduce**
1. Run any single test in the repo that imports `tests/conftest.py`'s fixtures, e.g.:
   ```
   pytest -xvs tests/v1/storage_backend/test_scylla_backend.py::TestScyllaDBBackend::test_table_name
   ```
2. Observe RSS with `/usr/bin/time -v pytest ...` — resident memory jumps to ~9 GB on the very first test collected/run, even though `test_table_name` does no meaningful I/O and allocates almost nothing itself.

**Expected behavior**
Running a small, self-contained unit test file should not require ~9 GB of pinned host memory. The shared allocator is a reasonable optimization for tests that actually exercise `LMCacheEngineBuilder`-created engines at scale, but:
- it shouldn't be constructed for test sessions/files that never use it,
- `@pytest.mark.no_shared_allocator` should actually prevent the 5 GB allocation, not just skip the monkeypatch, and
- 5 GB may be larger than necessary for most unit tests (this size predates recent test additions and hasn't been revisited).

**Additional context**
Discovered while investigating https://github.com/LMCache/LMCache PR review conversation around `tests/v1/storage_backend/test_scylla_backend.py`'s `local_cpu_backend` fixture, which had its own (now-fixed) leak: each test constructed a fresh `LocalCPUBackend` with `max_local_cpu_size=1.0` (1 GB) via `return` instead of `yield`, so `.close()` was never called and ~13 GB accumulated across that file's 13 tests. After fixing that leak and right-sizing that fixture to `0.1` GB, the process still showed a ~9 GB baseline RSS — traced to this session-scoped shared allocator, which is unrelated to the Scylla backend and affects the whole test suite.

Suggested directions (not prescriptive):
- Make `memory_allocator` fixture lazy (only construct on first actual use via `request.getfixturevalue` inside `use_shared_allocator`, after the marker check).
- Right-size the shared pool (e.g. profile what the heaviest consuming test actually needs) rather than a flat 5 GB.
- Ensure `@pytest.mark.no_shared_allocator` fully opts a test out of paying the allocation cost, not just the monkeypatch.

vedjaw:
Closes #4295

`use_shared_allocator` (autouse, function-scoped) declared the session-scoped `memory_allocator` fixture as a parameter, so pytest instantiated the 5GB pinned-CPU allocator *before* the `no_shared_allocator` marker check ran — every test session paid the ~8-9GB RSS cost even when no test used the shared allocator.

This resolves the fixture lazily via `request.getfixturevalue("memory_allocator")` *after* the marker check:
- sessions running only `no_shared_allocator`-marked tests never construct the 5GB allocator;
- the first test that actually uses it triggers exactly one construction; later tests reuse the cached session instance (`getfixturevalue` caches session fixtures exactly like a declared parameter);
- behavior for tests using the shared allocator is unchanged; the 5GB size is deliberately untouched (out of scope per the issue).

- Verified fixture semantics with a structural replica (same session fixture + autouse shape, counter instead of the 5GB allocation): marked-only session → zero constructions; mixed session → single construction on first unmarked test, reused afterwards.
- `ruff format` / `ruff check` on `tests/conftest.py` — green; `python3 -m py_compile` — green.
- Could not run the full suite locally (no CUDA build profile on macOS); the change is confined to fixture wiring.
- DCO sign-off included.

**If applicable**:

- [x] this PR contains unit tests (existing suite exercises the fixture on every run)

Made with [Cursor](https://cursor.com)


mykaul (approved):
LGTM.

vedjaw:
@hickeyma @sammshen @ApostaC @deng451e pls approve if it looks good to you.

zhengfeihe (approved):
LGTM!

maobaolong (approved):

PR #4397, opened 2026-08-03
test: refine tests v1 level1 tests

VincyZhang:

1. Standardizes device gating in `tests/v1/`, aligns test runtime checks with backend capabilities。
2. Aligning backend markers and runtime gating, reducing environment-driven false failures.  
3. Refines multiple test package init guards by making `importorskip` conditions more explicit and targeted, so unsupported environments are skipped intentionally with clearer scope and less accidental over-skipping.  

This PR is limited to tests and test-execution configuration; it does not change product runtime behavior.

**If applicable**:


tests/v1/test_cache_engine.py
```diff
@@ -62,12 +68,9 @@ def get_expected_count(token_len, save_unfull_chunk, chunk_size):
 
 
 @pytest.mark.parametrize("save_unfull_chunk", [False, True])
-@pytest.mark.skipif(
-    not torch.cuda.is_available(),
-    reason="TODO: Add non-CUDA implementation to VLLMPagedMemGPUConnectorV2",
-)
+@pytest.mark.cuda
```
hlin99: pytestmark = pytest.mark.cuda                
can we skip all the test file in the very beginning?

tests/v1/mp_observability/subscribers/logging/test_extra_stats.py
```diff
@@ -294,13 +295,13 @@ def test_window_resets_while_cumulative_persists(self):
             subs[EventType.L1_EVICTION_LOOP_TICK](_tick())
 
             subs[EventType.MP_STORE_START](
-                _start(EventType.MP_STORE_START, "req-2", 200.0)
+                _start(EventType.MP_STORE_START, "req-2", now + 1.0)
```
hlin99, line 298: leave a question mark here

.github/workflows/test.yml
```diff
@@ -112,14 +112,7 @@ jobs:
 
       - name: "Run non-CUDA unit tests"
```
hlin99: more presice statement might be "CPU only unit tests" ?

tests/v1/test_trtllm_integration.py
```diff
@@ -278,7 +288,7 @@ def test_construct(self) -> None:
         # First Party
         from lmcache.v1.gpu_connector.gpu_connectors import TRTLLMGPUConnector
 
-        device = torch.device("cuda:0")
+        device = torch.device(torch_device_type)
```
hlin99: torch.device(torch_device_type, 0)  ？

tests/v1/test_trtllm_integration.py
```diff
@@ -296,7 +306,7 @@ def test_get_shape(self) -> None:
...
```
hlin99: ditto

tests/v1/test_trtllm_integration.py
```diff
@@ -313,7 +323,7 @@ def test_register_kv_caches_with_4d_tensor(self) -> None:
...
```
hlin99: ditto

tests/v1/test_torch_ops.py
```diff
@@ -11,12 +11,18 @@
...
 
 # First Party
+from lmcache import torch_device_type
 from lmcache.v1.multiprocess.native_completion import (
     DeviceHostFuncDispatcher,
     submit_callback_to_stream,
 )
 from lmcache.v1.platform import torch_ops as _py_ops
 
+pytestmark = pytest.mark.skipif(
```
hlin99: o , we don't need to add logic here. just checkout, and I'll help handle this test after some refactor. thanks.

tests/v1/test_token_database.py
hlin99, line 1: hi，is the change in this file by accident?

tests/v1/test_sglang_mp_adapter.py
hlin99, line 1: leave question mark here

tests/v1/test_pos_kernels.py
hlin99, line 1: TODO: https://github.com/LMCache/LMCache/pull/4370

tests/v1/test_musa_support.py
```diff
@@ -210,7 +212,7 @@ def test_create_gpu_connector_layerwise_rejected_on_hpu(
     Previously, ``use_layerwise=True`` on HPU silently fell through into
     ``VLLMPagedMemLayerwiseGPUConnector`` — the CUDA layerwise connector —
     which then crashed on HPU tensors when constructing
-    ``torch.cuda.Stream()``. The guard must reject this combination with a
+     ``torch.cuda.Stream()``. The guard must reject this combination with a
```
hlin99: space by accident?

tests/v1/test_musa_support.py
hlin99, line 1: file to be moved to platform/musa, but could be in seperate PR. just fyi

tests/v1/test_musa_native.py
hlin99, line 1: file to be moved to platform/musa, but could be in seperate PR. just fyi

tests/v1/test_musa_connector.py
hlin99, line 1: file to be moved to platform/musa, but could be in seperate PR. just fyi

tests/v1/test_mp_mem_kernels.py
```diff
@@ -16,11 +19,13 @@
 import lmcache.c_ops as lmc_ops
 
 # Skip all tests if cuda is unavailable
-pytestmark = pytest.mark.skipif(
-    not torch.cuda.is_available() or torch.cuda.device_count() == 0,
-    reason="No CUDA GPU present",
-)
-
+pytestmark = [
```
hlin99, line 22: all device should pass this test, but pending on https://github.com/LMCache/LMCache/pull/4370

tests/v1/test_mem_kernels.py
```diff
@@ -8,22 +8,30 @@
...
+# Local
+from .utils import (
+    check_mem_obj_equal,
+    check_paged_kv_cache_equal,
+    generate_kv_cache_paged,
+    generate_kv_cache_paged_list_tensors,
+)
+
+pytestmark = [
+    pytest.mark.cuda,
```
hlin99, line 23: is there any error running this test on xpu?

tests/v1/test_cache_engine.py
VincyZhang: I see—there is an inconsistency in this test. In fact, the test is skipped right at the start unless `torch_device_type` is 'cuda'  and available; however, only 60% of the test cases are marked with `pytest.cuda`, because the remaining 40% are device-agnostic.

tests/v1/mp_observability/subscribers/logging/test_extra_stats.py
VincyZhang: Switched to now-based timestamps to avoid flaky tests.

Fixed values like 100.0/200.0 can be treated as stale and pruned before END arrives, so START/END no longer pair and throughput becomes n/a. Using now, now+0.5, now+1.0 keeps events valid and makes the throughput assertion stable.

.github/workflows/test.yml
VincyZhang: Good catch!

tests/v1/test_trtllm_integration.py
VincyZhang: Changed to `device = torch.device(f"{torch_device_type}:0")`

tests/v1/test_token_database.py
VincyZhang: No, updated the test to use `db._hash_tokens` instead of a hand-rolled hash so expectations follow the real production hash path and stay stable across hash backend/compatibility changes.

tests/v1/test_sglang_mp_adapter.py
VincyZhang: Moved adapter imports behind `import_adapter_symbols` to defer optional `sglang-dependent` imports, so test collection still skips cleanly when sglang is absent while keeping import-order clean.

tests/v1/test_torch_ops.py
VincyZhang: Thanks for reminding, reverted this file.

tests/v1/test_musa_support.py
VincyZhang: Good catch!

tests/v1/test_mp_mem_kernels.py
VincyZhang: Got it!

tests/v1/test_mem_kernels.py
VincyZhang: This test's written around CUDA pinned-memory and CUDA transfer behavior, and calling CUDA-specific APIs like `torch.cuda.cudart()`/`cudaHostRegister`/`cudaHostUnregister` which are not supported on XPU.

hlin99: okay, i see. thanks for checking.

.buildkite/k3_tests/xpu/run.sh
```diff
@@ -81,7 +80,7 @@ fi
 log "discovered ${#XPU_TEST_FILES[@]} XPU-related test files"
 printf '  %s\n' "${XPU_TEST_FILES[@]}"
 
-PYTEST_ARGS=(-q --maxfail=1 -m "not cuda")
+PYTEST_ARGS=(-q --maxfail=1 -m "not cuda and not musa and not sglang")
```
hlin99, line 83: sglang is excluded, i guess the reason is we don't have sglang in the docker env, right?
but later on, as long as we have sglang in docker or can install in the test, sglang is testable.

tests/v1/mp_observability/subscribers/logging/test_extra_stats.py
VincyZhang: It's actually two pair of test windows, one is from `now` to `now+0.5`, the other one is 'now+1` to `now+1.5`. Both window length is 0.5 second.

hlin99: I see. thx

hlin99 (approved):
LGTM.

ApostaC (approved):
lgtm

PR #4399, opened 2026-08-03
[Operator] Make hostIPC optional

### LINKED ISSUE #4394: [Operator] Remove the hostIPC requirement
**Label**
new feature

The Q3 roadmap (#4025, §5 Production Deployment) includes *"Remove the hostIPC & privileged mode requirement from the LMCache operator"*. The privileged half was resolved by #3943 (`spec.privileged`, default `false`). The hostIPC half is still hardcoded: the operator always sets `hostIPC: true` on engine DaemonSet pods, and both injection webhooks force it onto opted-in vLLM pods.

There is also a contradiction inside the repo about whether hostIPC is actually needed:

- `operator/DESIGN.md` and `operator/README.md` state that `hostIPC: true` is required for CUDA IPC, and that without it `cudaIpcOpenMemHandle` fails with `cudaErrorMapBufferObjectFailed`.
- `examples/multi_process/lmcache-daemonset.yaml` and `vllm-deployment.yaml` (the reference manifests) use **no hostIPC at all** — only a `hostPath` mount of `/dev/shm` on both pods.

I verified on GPU that hostPath with /dev/shm works well without `hostIPC: true`(NVIDIA L40S, separate lmcache-server / vLLM pods):

Conclusion: the requirement is a **shared `/dev/shm` tmpfs**, not the IPC namespace. `CudaIPCWrapper` uses PyTorch `_share_cuda_()`, whose handle references a shared-memory ref-counter file in `/dev/shm` that the receiving process opens by name.

Mirror the #3943 pattern:

- Add `spec.hostIPC` (`*bool`, default `false`) to `LMCacheEngine` and `CacheBlendEngine`.
- Default (`false`): the DaemonSet mounts the host's `/dev/shm` via `hostPath` (volume `dev-shm`); no hostIPC. A user-supplied mount at `/dev/shm` suppresses the default.
- Opt-in (`true`): legacy behavior. Pod joins the host IPC namespace, no hostPath mount.
- Both injection webhooks mirror the engine's mode on opted-in vLLM pods.
- Fix the DESIGN.md/README claims.

This narrows the host-level grant from the whole host IPC namespace (System V IPC, POSIX MQs of every process on the node) to a single tmpfs mount, and it also makes the `/dev/shm` dependency explicit in the pod spec

N/A

**Additional context**

- Verified end-to-end with the modified operator on a GPU cluster: engine DaemonSet (hostPath `/dev/shm`, `hostIPC: false`) + webhook-injected vLLM pod complete a KV store → retrieve round-trip; injection webhook and payload-staging smokes pass.
- Related: #4025 (roadmap)

bluayer:

Fixes #4394. Part of the Q3 roadmap (#4025, §5): remove the hostIPC & privileged mode requirement from the operator (the privileged half was done in #3943).

Cross-pod CUDA IPC between the engine and vLLM needs both pods to see the same `/dev/shm` tmpfs, because the CUDA IPC handles are opened via files the driver and PyTorch create there. A shared host IPC namespace is not required. The operator previously hardcoded `hostIPC: true` on engine DaemonSet pods, and both injection webhooks forced it onto opted-in vLLM pods.

Following the #3943 pattern, this PR adds `spec.hostIPC` (`*bool`, default `false`) to `LMCacheEngine` and `CacheBlendEngine`:

- **Default (`false`)**: the engine DaemonSet mounts the host's `/dev/shm` via hostPath (volume `lmcache-dev-shm`), and the injection webhooks mirror the mount onto opted-in vLLM pods. This narrows the host-level grant from the whole host IPC namespace to a single tmpfs mount, and makes the `/dev/shm` dependency explicit in the pod spec.
- Opt-in (`true`): legacy behavior, kept for clusters that block hostPath volumes.
- A user-supplied mount at `/dev/shm`, or a volume already named `lmcache-dev-shm`, suppresses the injection.

**Verification** (NVIDIA L40S, EKS, separate lmcache-server / vLLM pods):
- With the modified operator, the e2e suite passes against a real cluster: CRD/lifecycle/field-coverage/auth smokes, the GPU vLLM integration round-trip, and both injection-webhook smokes. 

- **Upgrade impact**: the default flips, so upgrading restarts engine pods once. Running vLLM pods need no coordinated restart, since a hostIPC pod and a hostPath pod see the same host `/dev/shm`. AMD and hostPath-blocked clusters should set `spec.hostIPC: true` before upgrading.
- This does not reach PSS `baseline` compliance by itself, since hostPath volumes are also rejected by `baseline`/`restricted`, but it is a strictly narrower grant.

**If applicable**:

### DIFF
--- a/operator/internal/webhook/pod_inject_common.go
+++ b/operator/internal/webhook/pod_inject_common.go
@@ -184,7 +184,8 @@ func prepareInjection(
 // CUDA IPC requirement is that both processes see the same /dev/shm
 // tmpfs (PyTorch's CUDA IPC handles reference a ref-counter file there). When
 // the engine opts into hostIPC, the pod instead joins the host IPC namespace,
-// which exposes the host's /dev/shm without a mount. User-supplied wiring is
+// which exposes the host's /dev/shm without a mount. A pod that already has
+// hostIPC enabled also needs no /dev/shm mount. Other user-supplied wiring is
 // left untouched: an existing mount at /dev/shm or an existing volume named
 // "lmcache-dev-shm" suppresses the injection.
 //
@@ -195,6 +196,8 @@ func prepareInjection(
 func applyIPCSharing(pod *corev1.Pod, target *corev1.Container, hostIPC bool) {
 	if hostIPC {
 		pod.Spec.HostIPC = true
+	}
+	if pod.Spec.HostIPC {
 		return
 	}
 	// Leave user-supplied wiring untouched: an existing mount at /dev/shm, or
--- a/operator/internal/webhook/pod_inject_common.go
+++ b/operator/internal/webhook/pod_inject_common.go
@@ -178,6 +178,37 @@ func prepareInjection(
 	return kvJSON, idx, admission.Response{}, true
 }
 
+// applyIPCSharing wires the pod for cross-pod CUDA IPC with the node-local
+// engine, mirroring the engine's spec.hostIPC mode. By default (hostIPC=false)
+// it mounts the host's /dev/shm into the target container via hostPath — the
+// CUDA IPC requirement is that both processes see the same /dev/shm
+// tmpfs (PyTorch's CUDA IPC handles reference a ref-counter file there). When
+// the engine opts into hostIPC, the pod instead joins the host IPC namespace,
+// which exposes the host's /dev/shm without a mount. User-supplied wiring is
+// left untouched: an existing mount at /dev/shm or an existing volume named
+// "lmcache-dev-shm" suppresses the injection.
+//
+// Parameters:
+//   - pod: the decoded pod (mutated in place: hostIPC or volumes).
+//   - target: the resolved vLLM container (mutated in place: volume mount).
+//   - hostIPC: the engine's resolved spec.hostIPC value.
+func applyIPCSharing(pod *corev1.Pod, target *corev1.Container, hostIPC bool) {
+	if hostIPC {
+		pod.Spec.HostIPC = true
+		return
+	}
+	// Leave user-supplied wiring untouched: an existing mount at /dev/shm, or
+	// an existing volume named "lmcache-dev-shm" (whatever its source — mounting a
+	// volume of unknown source at /dev/shm could silently shadow the host
+	// tmpfs, and appending a same-named volume would be invalid).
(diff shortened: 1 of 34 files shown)


KuntaiDu:
@royyhuang Could you please take a look? Also cc @DongDongJu

ApostaC:
@aeon-x Please take a look at this PR as well. Thanks!

aeon-x:
Thanks a lot for the contribution. This looks very promising, let us verify in a k8s env.

operator/internal/webhook/pod_inject_common.go
```diff
@@ -158,6 +158,37 @@ func prepareInjection(
...
+// which exposes the host's /dev/shm without a mount. User-supplied wiring is
+// left untouched: an existing mount at /dev/shm or an existing volume named
+// "lmcache-dev-shm" suppresses the injection.
+//
+// Parameters:
+//   - pod: the decoded pod (mutated in place: hostIPC or volumes).
+//   - target: the resolved vLLM container (mutated in place: volume mount).
+//   - hostIPC: the engine's resolved spec.hostIPC value.
+func applyIPCSharing(pod *corev1.Pod, target *corev1.Container, hostIPC bool) {
+	if hostIPC {
```
royyhuang, line 197: IIUC, this only checks if `LMCacheEngine`'s `hostIPC` is set or not. Probably also need to check if user manually set host IPC in vllm deployment spec. If users manually set it, we also need to make sure we don't "double mount" the `/dev/shm`.

royyhuang (changes requested):
I have successfully run this PR in local kind cluster. Please see and fix the minor comment on an edge case below.

operator/internal/webhook/pod_inject_common.go
bluayer: Yes you're correct. I missed that edge case. Thank you.

bluayer: The fix has been completed. If hostIPC exists, the volume mount is skipped.

royyhuang:
Please resolve the conflicts and we can merge this PR.

bluayer:
@royyhuang, I resolved the conflicts by rebase

royyhuang (approved):
LGTM.

DongDongJu (approved):
LGTM

PR #4404, opened 2026-08-03
[MP][Coordinator] Construct token directory and its API based on cache event

ruizhang0101:

Coordinator-side features need two lookups the fleet can't answer today: given keys → their token ids, and given token content → the keys holding it (blend-style fragment matching, follow-up). This PR builds the data plane: every `STORE` cache event entry carries its chunk's `token_ids` (a deliberate wire trade — tokens ride once per rank/group/tier placement in exchange for a self-contained protocol with no cross-entry ordering), and the key directory maintains a chunk-hash-keyed token index — `chunk_hash → (token_ids, keys)` — where every record joins its chunk's binding at creation and the binding dies with the chunk's last record (delete, incarnation fencing, `drop_instance`). Bounded structurally, no new config, and no hashing anywhere: the index key is the chunk hash every entry and lookup already carries. Emission is worker-0-only from the LMCache-driven store path and costs nothing when `--coordinator-event-reporting` is off; event ingestion runs off the coordinator's event loop. Refs #4226.


docs/source/mp/coordinator.rst
```diff
@@ -628,6 +628,152 @@ Usage events arrive on the fleet cache-event stream
...
+.. code-block:: bash
+
+    # Everything on server-1's L1:
+    curl -s "http://localhost:9300/directory/keys?tier=l1&instance_id=server-1&limit=100"
+
+``POST /directory/token_ids``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Return the token ids cached chunks hold, for the given keys (e.g. keys taken
+from ``GET /directory/keys`` or ``POST /directory/lookup``).
```
ApostaC: We can unify `token_ids` -> `keys` and its reverse into the same endpoint.

ApostaC: Clarify the semantics when we provide "token ids" into the lookup -- the token ids should be the whole request instead of just a single chunk (we need the prefix)

lmcache/v1/mp_coordinator/cache_events.py
```diff
@@ -148,21 +158,32 @@ def __init__(
         instance_id: str,
         incarnation: int,
         flush_interval: float = _DEFAULT_FLUSH_INTERVAL,
+        token_binding_cache_size: int = _DEFAULT_TOKEN_BINDING_CACHE_SIZE,
```
ApostaC: we can use a module-level global const instead of letting the caller to "configure" this parameter.

lmcache/v1/mp_coordinator/cache_events.py
```diff
@@ -248,12 +270,21 @@ def _on_l2_store(self, event: Event) -> None:
...
         )
 
+    def _on_tokens_stored(self, event: Event) -> None:
+        chunk_hashes: list[bytes] = event.metadata["chunk_hashes"]
+        token_chunks: list[list[int]] = event.metadata["token_chunks"]
+        for chunk_hash, chunk in zip(chunk_hashes, token_chunks, strict=True):
+            self._token_bindings[chunk_hash] = tuple(chunk)
+            self._token_bindings.move_to_end(chunk_hash)
+        while len(self._token_bindings) > self._token_binding_cache_size:
+            self._token_bindings.popitem(last=False)
```
ApostaC: We need to add a warning/error log when LRU is triggered.

lmcache/v1/mp_observability/event.py
```diff
@@ -75,6 +75,13 @@ class EventType(Enum):
     # Chunk hash logging events
     MP_LOOKUP = "mp.lookup"
 
+    # Per-chunk token bindings for a submitted store. Metadata (parallel
+    # lists): chunk_hashes (list[bytes], the chunks' ObjectKey prefix
+    # hashes) and token_chunks (list[list[int]]). Published before the
+    # store's write-finished events so consumers can stamp token ids
+    # onto the subsequent STORE events.
+    MP_TOKENS_STORED = "mp.tokens.stored"
```
ApostaC: Simplify doc string here.
The name is not necessarily coupled with "store"

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -1157,6 +1166,53 @@ def store(
             store_succeeded,
         )
 
+    def _publish_token_bindings(
```
ApostaC: move this function to the end of this class

ApostaC (commented):
PTAL

docs/source/mp/coordinator.rst
ApostaC, line 1: Add another example for keys form request

ApostaC (approved):
LGTM!

sammshen (approved):
LGTM!

PR #4079, opened 2026-07-10
[Refactor][MP] Move the shared KV copy helpers to object_group_transfer

divyvasal:
Reworked per review (see the design discussion below with @ApostaC and @hlin99): the descriptor/planner abstraction is gone. This is now the minimal enabler for engine-driven hybrid support, done the way the review asked: **relocate, do not rewrite**.

`transfer_kv_per_object_group` and the helpers it depends on (`batched_iteration_with_skip`, `downsample_and_stage_block_ids`, `recalculate_blocks_to_skip`, `_run_object_group_transfer_plan`) move **verbatim** from `modules/lmcache_driven_transfer.py` to `lmcache/v1/multiprocess/object_group_transfer.py`. `lmcache_driven_transfer.py` imports the two helpers its class code still calls; `qstore.py` and the `batched_iteration_with_skip` test import from the new module directly.

Nothing in the moved code is specific to who drives the transfer: it reads its geometry from the cache context's `KVLayerGroupsManager`. The engine-driven path (#4410 follow-up) can now import the copy helpers without importing the LMCache-driven module, and will build its `KVLayerGroupsManager` at registration per the review's three steps.

Verified: every moved function body is byte-identical to `dev`'s (whitespace-insensitive diff per function); `tests/v1/multiprocess/` shows no new failures against `dev`; `ruff`/`isort`/`mypy` clean on the touched files.

Follow-ups from @sammshen's review on 09-15, all applied: the module is named `object_group_transfer.py` rather than a second `transfer_utils`; `recalculate_blocks_to_skip` is public because #4410 imports it; the dead re-exports are dropped; the docstring now just says what the module is.

PR #4410, opened 2026-08-04
[Feat][MP] Engine-driven multi-group transfers via KVLayerGroupsManager

divyvasal:
Rebuilt along @ApostaC's three steps, on top of #4079 (first commit is that move; it drops out when #4079 merges):

1. Registration initializes `KVLayerGroupsManager` and derives every group's geometry from it (no per-group re-detection). The manager also validates chunk and window divisibility.
2. Block IDs are parsed per group from the manager's kernel groups; sliding-window groups keep only the trailing window of each chunk.
3. Gather/scatter call the shared helpers moved in #4079, extended with `blocks_per_window`.

The chunk token count is `blocks_in_chunk * lcm(group block sizes)`, folding in the mixed-geometry fix @skaulintel found on the old branch (gemma-4 registered half of every chunk). Covers both transports: SHM reserves per-group slots, pickle sends a group-major payload. k3 lanes: kimi_linear_tp and engine-driven HMA lm-eval, both transports.

### DIFF
--- a/.buildkite/k3_tests/multiprocess/pipeline.yml
+++ b/.buildkite/k3_tests/multiprocess/pipeline.yml
@@ -200,6 +200,49 @@ steps:
         plugins: [{ kubernetes: { podSpec: *pod-1gpu } }]
         artifact_paths: ["*.log"]
 
+      # Engine-driven HMA verifies both transport backends end-to-end. Both
+      # matrix entries use the dshm pod: pickle remains lazy and ignores the
+      # mount, while SHM disables lazy L1 allocation and maps the pool.
+      - label: ":electric_plug: hma_lm_eval_gemma4_engine_driven / {{matrix.transport}}"
+        command: |
+          export LMCACHE_MP_TRANSFER_MODE="engine_driven"
+          export ENGINE_DRIVEN_TRANSPORT="{{matrix.transport}}"
+          case "{{matrix.transport}}" in
+            pickle)
+              export L1_USE_LAZY="true"
+              # Pickle sends complete hybrid KV payloads through one MQ
+              # connection. Limit this correctness-only job's request
+              # concurrency so queued payloads do not exhaust the RPC budget;
+              # the SHM matrix entry retains the regular 50-way stress load.
+              export NUM_CONCURRENT="4"
+              export LIMIT="30"
+              export STORE_DRAIN_SECONDS="180"
+              export LMCACHE_MQ_TIMEOUT="300"
+              ;;
+            shm)
+              export L1_USE_LAZY="false"
+              ;;
+          esac
+          .buildkite/k3_tests/multiprocess/run.sh hma_lm_eval_gemma4_engine_driven
+        matrix:
+          setup:
+            transport:
+              - "pickle"
+              - "shm"
+        timeout_in_minutes: 60
+        retry: *flaky-retry
+        env:
+          MODEL: "google/gemma-4-31B-it"
+          SCORE_TOLERANCE: "0.03"
+          ATTENTION_BACKEND: "auto"
+          GPU_MEMORY_UTILIZATION: "0.85"
+          MAX_WAIT_SECONDS: "600"
+          LIMIT: "100"
+          CPU_BUFFER_SIZE: "80"
+        agents: { queue: "k8s" }
+        plugins: [{ kubernetes: { podSpec: *pod-1gpu-dshm } }]
+        artifact_paths: ["*.log"]
+
       - label: ":compression: fault_tolerance"
         command: .buildkite/k3_tests/multiprocess/run.sh fault_tolerance
         timeout_in_minutes: 30
--- a/.buildkite/k3_tests/multiprocess/scripts/launch-processes.sh
+++ b/.buildkite/k3_tests/multiprocess/scripts/launch-processes.sh
@@ -14,6 +14,7 @@ vllm_port="${VLLM_PORT:-8000}"
 vllm_baseline_port="${VLLM_BASELINE_PORT:-9000}"
 CPU_BUFFER_SIZE="${CPU_BUFFER_SIZE:-80}"
 MAX_WORKERS="${MAX_WORKERS:-4}"
+LMCACHE_MQ_TIMEOUT="${LMCACHE_MQ_TIMEOUT:-10}"
 MODEL="${MODEL:-Qwen/Qwen3-14B}"
 BUILD_ID="${BUILD_ID:-local_$$}"
 
@@ -152,7 +153,7 @@ VLLM_SERVER_DEV_MODE=1 \
 VLLM_BATCH_INVARIANT=${BATCH_INVARIANT} \
 PYTHONHASHSEED=0 \
 vllm serve "$MODEL" \
-    --kv-transfer-config "{\"kv_connector\":\"LMCacheMPConnector\", \"kv_role\":\"kv_both\", \"kv_load_failure_policy\": \"recompute\", \"kv_connector_extra_config\": {\"lmcache.mp.port\": $LMCACHE_PORT, \"lmcache.mp.mq_timeout\": 10}}" \
+    --kv-transfer-config "{\"kv_connector\":\"LMCacheMPConnector\", \"kv_role\":\"kv_both\", \"kv_load_failure_policy\": \"recompute\", \"kv_connector_extra_config\": {\"lmcache.mp.port\": $LMCACHE_PORT, \"lmcache.mp.mq_timeout\": $LMCACHE_MQ_TIMEOUT}}" \
     $ATTENTION_BACKEND_ARG \
     --port "$vllm_port" \
     --no-async-scheduling \
--- a/.buildkite/k3_tests/multiprocess/scripts/run-hma-lm-eval-engine-driven.sh
+++ b/.buildkite/k3_tests/multiprocess/scripts/run-hma-lm-eval-engine-driven.sh
@@ -1,0 +1,261 @@
+#!/usr/bin/env bash
+# Engine-driven HMA (hybrid memory allocator) correctness test.
+#
+# Identical to run-hma-lm-eval.sh in flow and coverage, but forces the
+# multiprocess transfer path to Engine-driven (not the default LMCache-driven
+# / AUTO selection). This ensures the Engine-driven path is exercised on CUDA
+# workers with a real hybrid model.
+#
+# Forced via LMCACHE_MP_TRANSFER_MODE=engine_driven (environment variable
+# respected by TransferContextFactory before AUTO detection runs).  The CI
+# pipeline also sets lmcache.mp.mp_transfer_mode=engine_driven in
+# kv_connector_extra_config for belt-and-suspenders coverage.
+#
+# Models (selected by run-single-test.sh):
+#   - google/gemma-4-31B-it: sliding-window + full-attention hybrid whose full
+#     layers have a larger head_dim, exercising per-group HMA store/retrieve
+#     via Engine-driven path.
+#   - Qwen/Qwen3.5-0.8B: Mamba/GDN + full-attention hybrid; exercising the
+#     registration-time cache re-views (kv_cache_group_edits.py).
+#
+# Flow (single GPU, no baseline server):
+#   1. Sanity-check the LMCache server log to confirm Engine-driven context was
+#      created (not LMCache-driven).
+#   2. vLLM run: lm_eval (gsm8k) against vLLM+LMCache, populating LMCache.
+#   3. Reset vLLM's *local* prefix cache (APC) only, leaving LMCache intact, via
+#      the dev-mode endpoint POST /reset_prefix_cache.
+#   4. LMCache retrieve run: re-run lm_eval; vLLM APC misses → LMCache serves KV.
+#   5. Assert the two runs' gsm8k scores match.
+#   6. Assert LMCache actually served retrieves in the retrieve run (non-vacuous).
+#
+# The reset endpoint requires VLLM_SERVER_DEV_MODE=1 (set by launch-processes.sh).
+# The LMCACHE_MP_TRANSFER_MODE and ENGINE_DRIVEN_TRANSPORT must be set before
+# launching the servers (already done by the pipeline matrix command).
+set -e
+set -o pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../../../.." && pwd)"
+
+source "${REPO_ROOT}/.buildkite/k3_tests/common_scripts/helpers.sh"
+
+# Configuration
+VLLM_PORT="${VLLM_PORT:-8000}"
+MODEL="${MODEL:-google/gemma-4-31B-it}"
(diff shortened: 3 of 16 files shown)


hlin99:
https://github.com/LMCache/LMCache/pull/4412 is the CI for engine-driven hybrid model. can you apply the PR and let's see how it works? thanks.

hlin99:
hi @divyvasal [k3-multiprocess-test] shows failure. can you help check?

divyvasal:
Root cause found — it's a model/scope mismatch, not the harness:

```
worker_transfer.py:738 in register
RuntimeError: engine-driven multi-group transfer only supports uniform-coverage groups;
              group 0 is sliding-window (sw_size_tokens=1024). Run with a unified KV cache manager instead.
```

The model the engine-driven hybrid CI (#4412) launches has a **sliding-window group** (group 0, `sw_size_tokens=1024`). v1 engine-driven multi-group is **uniform full-attention only** — SW groups are deferred to the lmcache-driven path — so `register()` rejects it, and because that runs inside vLLM engine init, the whole server fails to start. Expected v1 behavior, just surfacing as a hard engine-init crash rather than a graceful skip.

Three ways forward:

1. **Point the engine-driven hybrid CI at a full-attention hybrid model** (MLA + indexer — the GLM/DeepSeek-style hybrids this multi-group path was validated on). That's exactly the v1 scope and exercises the path this PR adds.
2. **Graceful fallback** — instead of hard-raising in `register()` on a SW group, detect it at the connector and route the model through the existing single-group / lmcache-driven path so the server still starts (degrade, don't crash engine init). Small, robust change.
3. **SW support in engine-driven multi-group** — the v2 follow-up; out of scope here.

Plan: (1) for the CI now so it lands on a model this PR actually supports, plus (2) so a SW model can never crash engine init. Will push both.

hlin99:
> Root cause found — it's a model/scope mismatch, not the harness:
> 
> ```
> worker_transfer.py:738 in register
> RuntimeError: engine-driven multi-group transfer only supports uniform-coverage groups;
>               group 0 is sliding-window (sw_size_tokens=1024). Run with a unified KV cache manager instead.
> ```
> 
> The model the engine-driven hybrid CI (#4412) launches has a **sliding-window group** (group 0, `sw_size_tokens=1024`). v1 engine-driven multi-group is **uniform full-attention only** — SW groups are deferred to the lmcache-driven path — so `register()` rejects it, and because that runs inside vLLM engine init, the whole server fails to start. Expected v1 behavior, just surfacing as a hard engine-init crash rather than a graceful skip.
> 
> Three ways forward:
> 
> 1. **Point the engine-driven hybrid CI at a full-attention hybrid model** (MLA + indexer — the GLM/DeepSeek-style hybrids this multi-group path was validated on). That's exactly the v1 scope and exercises the path this PR adds.
> 2. **Graceful fallback** — instead of hard-raising in `register()` on a SW group, detect it at the connector and route the model through the existing single-group / lmcache-driven path so the server still starts (degrade, don't crash engine init). Small, robust change.
> 3. **SW support in engine-driven multi-group** — the v2 follow-up; out of scope here.
> 
> Plan: (1) for the CI now so it lands on a model this PR actually supports, plus (2) so a SW model can never crash engine init. Will push both.

hi， what does v1 & v2 refer to ? i think the problem we're tring to resolve here is the hybrid model in engine driven path(inc linear attn).....here, the CI we use cuda device, so we have full set of c ops which can be utilized even on engine-driven path. (i'm implementing a python fallback ops that other devices can take much easier for functionalites)

the hybrid model makes the diverge day-to-day, so would like to see v1 to resolve the problem totally. v2 seems to be too late

divyvasal:
Sorry, "v1/v2" was just my loose wording. v1 meant the first version that only handled full-attention groups, and v2 meant adding sliding-window. They aren't real milestones. You're right that it should be fully fixed in this PR.

Already done: pushed `ecde6e0a`. `register()` no longer rejects sliding-window groups. A sliding-window group now stores just the last blocks of each chunk (same idea as the lmcache-driven `downsample_and_stage_block_ids`), and this works on both the CUDA C-ops path and the python fallback. Full-attention groups behave exactly as before.

`k3-multiprocess-test` is running again on the hybrid model now.

KuntaiDu:
Happy to review when it's ready!

hlin99:
hi @divyvasal good to see shm path succeeded! thanks
for pickle path, is there any reason we dropped it? In theory, pickle should also work in the design, but I also found pickle is more challenging, do you have any insight?

divyvasal:
@hlin99 Nothing fundamental. It was scoped out, and `register()` fails loudly rather than silently degrading.

The multi-group path is built around the SHM slot model: `prepare_store_grouped` / `prepare_retrieve_grouped` reserve a fixed-size slot per (group, chunk) up front (all-or-nothing, group-major on the server), and the planner's copy steps resolve raw pointers into those slots. Pickle mode has neither piece. Objects are serialized whole with variable size, so there's no reservation step and no stable destination pointer for the plan executor to target.

It's doable in theory: gather each group's chunk into a staging tensor, serialize the group tensors as one object (or N keyed objects), and reverse on retrieve. But that's a different prepare/commit shape rather than a reuse of the shared plan executor, and the all-or-nothing multi-group semantics would need reimplementing on the pickle server path. Since pickle mode only arises when the server has no SHM pool (lazy L1), i.e. it's already the degraded transport, I kept the initial scope to SHM.

Pickle multi-group can be a follow-up PR later. It would pair well with your python fallback ops for non-CUDA devices.

divyvasal:
@KuntaiDu It's ready for review.

hlin99:
hi @divyvasal  please cherry-pick kimi CI from https://github.com/LMCache/LMCache/pull/4412

and please get pickle back. l'll check the error msg. thx.

divyvasal:
@hlin99 found the pickle bug. It was a use-after-free on the retrieve side, nothing wrong with the transport design.

Unpickled chunks arrive unpinned, so scatter_cpu_to_paged_kv pins them into temporary copies and launches the async H2D through raw pointers. Torch's stream tracking cannot see that use, so when the function returned, the temporaries were freed and the caching host allocator handed the same pinned memory to the next group's pin_memory() while the previous group's copies were still in flight. With six groups scattered back to back per retrieve, the staging memory got overwritten mid-transfer almost every time, but timing dependent. That is why the scores bounced around (0.03, 0.13, 0.50 across runs) and why the failure only showed on pickle: the SHM slot views stay referenced and the gather buffers are pre-pinned, so SHM never takes that branch.

Fix is 0cc4b50d: synchronize before releasing the temporaries, pinned inputs keep the fully async path. Verified on an 8xH100 box with a standalone repro (about 4000 corrupted blocks per run before, zero after) and the full hma gemma-4 pickle flow end to end: 0.8000 vs 0.8000, zero drift, test passed. Also added a CUDA regression test that scatters unpinned chunks across six groups back to back.

Earlier fixes on this branch (window-sized gather buffers in ab813492 and the missing sync before pickling in 052e06fd) were real bugs too, all three stacked. CI is rerunning now.

divyvasal:
Marking this ready for review. k3-multiprocess-test is green on the current head: the hma gemma-4 engine-driven entries pass on both shm and pickle, and the kimi entries pass as well, so the CI goal we set is met. The xpu failure is the known pre-existing regression on dev and AMD is queued behind runner availability.

One sequencing note: per the plan on #4079, the move onto the shared transfer descriptor happens after that PR lands, so this branch will get one rebase then. Review here is the multi-group registration, the server side, and the two transports. @KuntaiDu

hlin99:
hi @divyvasal can you help resolve the conflict? there're a lot of refactor recently. sorry for the inconvinience.

hlin99 (approved):

divyvasal:
@hlin99 Let me rebase #4079 and then resolve the conflicts on #4410 . Thanks for the approval.

divyvasal:
Done, both are merged with dev now. #4079 first, then this one.

The conflicts came from the c_ops removal (#4502) and the KV format enum move (#4473). Mechanical on both branches: kernels and plan constructors moved to `device_ops`, `TransferDirection` and `EngineKVFormat` to `lmcache_native`, and the native plan type hints to `ops_types`. #4079 had moved that code into `transfer_descriptor.py`, so the same migration is replayed there.

Two things worth flagging on this PR:

- I kept the plan-driven batching in gather/scatter rather than dev's inline slicing loops, since the planner is what this path shares with the lmcache-driven one.
- Picked up `all_null_chunk_masks` from dev, and fixed a stale `stub_native_storage_ops` fixture reference (renamed to `stub_lmcache_native` on dev).

I ran the multiprocess suite against a clean dev checkout as a baseline before and after. No new failures on either branch.

One note: this branch still carries `transfer_plan.py`, so it will want a second rebase onto the descriptor once #4079 lands. That was the plan from the original split, just calling it out so the churn is not a surprise.

divyvasal:
Fixed the two multi-group test failures in `00e81106`.

Both hard-coded a block size of 4, which is the NHD reading of a 5-D `[2, NB, BS, NH, HS]` tensor. The vLLM detector forces HND when `torch_device_type` is `"cpu"`, so on a CPU-only CI host the same tensor reads as `[2, NB, NH, BS, HS]` and the block size is 2. Production is unaffected, since a GPU worker resolves NHD and gets the layout vLLM reports through `layout_hints`.

The sliding-window case was worse than a wrong constant. With block size 2 the chunk is 4 tokens, so the fixed 4-token window was no longer smaller than the chunk, the group quietly became full attention, and the test stopped covering sliding windows at all.

Both now derive the block size from `compute_kv_layout` (what registration itself uses) and size the window at half a chunk. I confirmed they pass under both layouts by patching the detector to report `cuda`, and checked that the patch really does flip detection (block size 2 to 4) so the second run is not vacuous.

While I was in there I also checked the production path: `compute_kv_layout` derives block size, layer count, heads, head size and kv size from the detected format, the per-group loop calls the same helper, and `vllm_layout_hints()` forwards the real layout. The NHD assumption was only in the tests, not in the feature.

skaulintel:
@divyvasal @hlin99 — ran this branch (`00e81106`) on Intel XPU and CUDA and hit a
correctness bug worth folding into your rebase. One statement in `worker_transfer.py`;
PR against `feat/engine-driven-e2e`: <LINK>.

**Symptom.** gemma-4-31B, TP2, HMA on. Registration and stores work, but the tier stores **exactly half** the bytes per token (440.00 KiB/token vs the config's 880.00) and turn 2 is a clean external hit that decodes to garbage. Hybrid-off control on the same branch answers correctly.

**Cause.** In `EngineDrivenTransferContext.register`, `chunk_tokens = blocks_in_chunk * block_size` multiplies vLLM's *scheduling* block count by the block size of whichever layer is first in `kv_caches`. Equal under one geometry only. gemma-4 has 6 groups (5 sliding at 32-token blocks + 1 global at 64), so a 256-token chunk registers as `4 * 32 = 128`; the divisibility guard stays quiet because 128 divides both. Fix is`lcm(EngineGroupInfo.tokens_per_block)`.

**Why CI is green.** `test_engine_driven_multigroup.py` is geometry-uniform in all 10 cases — group *count*, never mixed *shape*. Confirmed on hardware: gpt-oss-20b (2 groups,one geometry) is correct at 36.00 KiB/token, gemma-4 (6 groups, two geometries) corrupts.The PR adds a 3-second GPU-free test that's red on `00e81106`; `tests/v1/multiprocess/` is594 passed / 34 skipped.

**Not XPU-specific.** Same 880.00 KiB/token on 2x L40S with the fix, byte-identical to the fixed XPU run. The line is engine-driven only, which CUDA's default IPC path never reaches.

opened a pull request with the fix here - https://github.com/divyvasal/LMCache/pull/3

divyvasal:
Rebuilt this on the reshaped #4079 following @ApostaC's steps: registration now builds KVLayerGroupsManager and parses per-group geometry from it, and the transfer paths call the moved helpers. Also folds in @skaulintel's mixed-geometry chunk fix with a test that fails without it. The descriptor/planner from the old branch is gone.

divyvasal:
CI is green here except buildkite/k3-unit-tests, which failed on test_mq_lookup. That one is a flake unrelated to this PR: the same lane passed on the previous build of this branch with no Python changed in between (only buildkite yaml and shell), all four GitHub Actions lanes run that same test on this commit and pass, and the lane fails intermittently on other contributors' branches too.

The MQ tests bind hardcoded loopback ports, so two runs sharing a runner race for the same port and the loser reports a client timeout. I opened #4944 to fix that at the source by giving each test a free port.

The engine-driven k3 multiprocess lanes (kimi_linear_tp and the HMA lm-eval, both transports) pass on this commit.

divyvasal:
Heads-up for whichever of this and #4068 merges second: `_write_chunks` here releases leftover reservations with `finish_write` followed by `delete_l1_keys`. #4068 adds `abort_write` for exactly this case, because `finish_write` fires the store listener before the delete lands, so an L2 store controller can persist the unwritten object. Once #4068 is in, this should call `abort_write` instead.

divyvasal:
Heads-up on a lane this PR adds. `kimi_linear_tp_engine_driven / pickle` is intermittently failing: across every real attempt so far it is 3 pass / 2 fail, while the `shm` arm is 3/3. It fails the byte-exact output comparison at the end of the kimi script, so the retrieve does land but the KV content diverges after an identical prefix. The lane is green here only because the step retries and the third attempt passed.

What I ruled out, in case it saves anyone time:

- not a regression from the rebase: the diff between the last green run and the failing ones is mechanical, only the `instance_id` plumbing from #5064 plus the #4079 rename
- not a vLLM change: both runs installed `0.29.1rc1.dev128+gcd10ed6f9`
- not node specific: the pass and both failures ran on the same agent
- not load: the green run had more concurrent jobs on that node than the first red one

The remaining suspect is buffer lifetime or ordering in the pickle scatter, which is the same area as #4830. I cannot reproduce it off the GPU lane because the test written for that mechanism is CUDA gated, so I am flagging it rather than letting it be discovered later.

Happy to dig further, or to gate that matrix entry until it is understood, if reviewers would rather the rest land first.

maobaolong:
Hi @divyvasal , as #4079 merged, you can rebase this PR now.

divyvasal:
Thanks @maobaolong. Rebasing onto the new dev now.

divyvasal:
@maobaolong Rebased onto current dev, so #4079's changes are out of the diff now. Please take a look.

divyvasal:
@maobaolong Rebased onto current dev. Two of dev's refactors needed real changes rather than a replay: handlers now take the argument-free `@request_handler()`, and `reserve_write` lost its mode argument, so the multi-group write path and the test fake follow the new signatures.

The k3 harness reorganisation also landed, so the engine-driven coverage is expressed the way dev does it now: a pipeline lane that selects the transfer mode through the environment and reuses the existing `kimi_linear_tp` workload. I dropped `run-hma-lm-eval-engine-driven.sh`, which only duplicated the HMA workload to set the transfer mode that `launch-processes.sh` now passes itself, and whose lane would have re-enabled `hma_lm_eval_gemma4` while dev has it disabled. Happy to add that lane back if you'd rather have it.

divyvasal:
@maobaolong One more ask: could you add the `mp` label? The multiprocess pipeline is gated on `mp` (or `full`), so the engine-driven lane this PR ports has not run on the current head. The rest is green: 21 checks pass, including code quality, the CPU device matrix and k3-unit-tests.

maobaolong:
@divyvasal Done.

PR #4411, opened 2026-08-04
[Fix][CLI] long-doc-permutator: size contexts in tokens, not words

BoJiang03:

`--ldp-context-length` and `--ldp-system-prompt-length` are documented as token
counts, but `long-doc-permutator` spent them as **word** counts: it sampled
`length` pseudo-words from a synthetic vocabulary. Each pseudo-word carries an
incrementing numeric suffix (`babazuz7151`), so it costs several BPE tokens —
and the factor is tokenizer-specific, so identical flags produce differently
sized prompts on different models.

Measured against `openai/gpt-oss-20b`:

```
lmcache bench engine --workload long-doc-permutator \
  --ldp-num-contexts 4 --ldp-context-length 4500 \
  --ldp-system-prompt-length 1000 --ldp-num-permutations 24
```

| | requested | actually sent |
|---|---|---|
| per request | 19,000 | **74,470** (3.92x) |
| whole run | 456,000 | **1,787,280** |

The expansion factor is not a constant: gpt-oss-20b sees 3.92x, Qwen3.6 sees
6.2x. So blend-vs-baseline runs are comparing at a ~4x larger operating point
than their flags describe, the comparison is not portable across models, and a
context length that looks safely inside `--max-model-len` can overflow it.

**The fix.** Build context text from words that each cost exactly one token
under the model's tokenizer, and that each begin a word. Such a token is
segmented independently of its neighbours, so N words is exactly N tokens
whichever order a permutation puts them in. The same flags now produce 4,500
tokens per context.

Candidates come from the vocabulary's *keys*, not from `decode()` output.
SentencePiece decodes `▁the` to `"the"`, dropping the very marker that
identifies a word start, so a decode-based filter finds nothing at all on
Llama-family tokenizers. Each candidate is then verified by encoding it as it
will appear in a prompt, under whichever joining convention that tokenizer
makes exact — a leading space per word, or plain separators for tokenizers that
charge a token for an explicit leading space (Phi-3). The pool carries its
convention, so the two cannot be separated, and the joined total is checked
before use.

Exercised on 18 tokenizers; 4,500 pool words encode to exactly 4,500 tokens,
and stay exact after a shuffle, on all but WordPiece:

| family | tokenizers | join | result |
|---|---|---|---|
| byte-level BPE | gpt-oss-20b, gpt2, Qwen2.5, Qwen3, Llama-3.1, falcon, gpt-neox, opt, phi-2, starcoder2, deepseek-coder | leading space | exact |
| SentencePiece | Llama-2, TinyLlama, Yi, flan-t5, xlm-roberta | leading space | exact |
| SentencePiece, space is a token | Phi-3-mini | separator | exact |
| WordPiece | bert-base-uncased | — | rejected: marks continuations (`##ing`), not word starts |

Verified end-to-end against vLLM 0.26.0 + `openai/gpt-oss-20b`, reading
`num_input_tokens` out of `bench_results.csv`:

| `-n 3 --ldp-context-length 2000 --ldp-system-prompt-length 500` | requested | `num_input_tokens` | TTFT |
|---|---|---|---|
| before | 6,500 | 24,916 (3.83x) | 1.29 s |
| after | 6,500 | **6,576** | 0.21 s |

The residual +76 is the harmony chat template (+74) plus the two `\n\n`
separators joining the three contexts into one user message — i.e. everything
the workload itself controls is now exact.

**This makes a tokenizer mandatory for `long-doc-permutator`** — the one
behaviour change a user can notice. A length expressed in tokens cannot be
honoured without a tokenizer, and neither text-level fallback works for *this*
workload: pseudo-words miss the count (that is the bug), and uniform filler
emits identical contexts, which collapses the blend hit rate the workload
exists to measure. So the workload now fails with the flag to set:

```
long-doc-permutator needs a tokenizer to size its contexts in tokens, but none
could be loaded for model 'not-a-real-model' (auto-detected from the engine
when --model is omitted). Pass --model with a HuggingFace repo ID or a local
path, e.g. --model openai/gpt-oss-20b, and make sure transformers is installed.
```

In the common case nothing changes: `--model` is auto-detected from the engine,
and the tokenizer loads.

Also in this PR:

- `try_load_tokenizer` is extracted into `engine_bench/tokenizers.py`, shared
  with `prefix-suffix-tuner`, next to the new single-token pool builder. That
  is the only change to `prefix-suffix-tuner`: its behaviour is untouched, it
  keeps its `"hi"`-filler fallback.
- Tests use an offline tokenizer stand-in rather than downloading a real one.
  The assertions that pinned the old behaviour (`len(ctx.split()) == length`)
  now count tokens, and fail against the old generator — under the stand-in a
  context of 100 requested words costs roughly ten times that many tokens.

Not addressed here, noted for a follow-up: `--tokens-per-gb-kvcache` /
`--lmcache-url` are required for every workload even though only KV-sizing
needs them.

**If applicable**:

### DIFF
--- a/docs/design/cli/commands/bench/engine_bench/bench-engine.md
+++ b/docs/design/cli/commands/bench/engine_bench/bench-engine.md
@@ -418,10 +418,10 @@ chunk-level cache lookup and eviction.
 | Field | CLI arg | Default | Description |
 |-------|---------|---------|-------------|
 | `num_contexts` | `--ldp-num-contexts` | 5 | Number of unique context documents (`N`) |
-| `context_length` | `--ldp-context-length` | 5000 | Tokens per context |
-| `system_prompt_length` | `--ldp-system-prompt-length` | 1000 | Shared system prompt tokens (`0` disables) |
+| `context_length` | `--ldp-context-length` | 5000 | Tokens per context (exact) |
+| `system_prompt_length` | `--ldp-system-prompt-length` | 1000 | Shared system prompt tokens, exact (`0` disables) |
 | `num_permutations` | `--ldp-num-permutations` | 10 | Distinct permutations to send (capped at `N!`) |
-| `vocab_size` | (none — hardcoded in factory) | 8000 | Vocabulary pool size for synthetic context generation |
+| `vocab_size` | (none — hardcoded in factory) | 8000 | Number of distinct single-token words contexts are sampled from |
 | `num_inflight_requests` | `--ldp-num-inflight-requests` | 1 | Max concurrent in-flight requests |
 
 **Stress axes** (each config field tunes one):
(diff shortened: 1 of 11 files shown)


ApostaC (approved):
LGTM!

deng451e (approved):
LGTM!

Issue #4100, opened 2026-07-13
[Feature][MP] Add io_uring fixed-buffer support for Lazy and Mixed L1 allocators


2xdevv:
**Label**

new feature

**Is your feature request related to a problem? Please describe.**

io_uring fixed-buffer registration is not implemented for the MP `raw_block` path. There was an initial attempt in [#3246](https://github.com/LMCache/LMCache/pull/3246), but MP requires a different integration from the legacy `RustRawBlockBackend`.

The legacy backend owns its CPU backend and can register allocator buffers directly during initialization. In MP mode:

- `RawBlockL2Adapter` does not own the L1 allocator.
- L1 memory is managed separately by `L1MemoryManager`.
- MP uses either `MixedMemoryAllocator` or `LazyMemoryAllocator`.
- Lazy memory is pinned and published incrementally.

Both allocators also have a boundary problem. io_uring limits each registered buffer to 1 GiB, while both allocators manage a larger contiguous address space. `AddressManager` may therefore return an allocation spanning two registered-buffer indices. One `ReadFixed` or `WriteFixed` SQE cannot handle such a request because it contains only one buffer index.

**Describe the solution you'd like**

Add an MP-specific interface for exposing fixed-buffer regions and notifying `RawBlockL2Adapter` when new regions become available.

For both allocators:

1. Divide eligible L1 memory into regions no larger than 1 GiB.
2. Assign one fixed-buffer index to each region.
3. Change Rust lookup from exact base-pointer matching to range containment.
4. Split an I/O that crosses region boundaries into multiple fixed SQEs, one per registered region.
5. Advance the buffer pointer, length, and device offset for each segment.
6. Aggregate the segment completions into one logical I/O completion.
7. Fall back to ordinary io_uring I/O only when registration is unavailable or part of the request is not covered by registered regions.

### MixedMemoryAllocator

Mixed owns its complete arena at initialization, so all eligible regions can be registered once before I/O starts.

Its `BINARY_BUFFER` allocations and unsupported backing types should continue using ordinary I/O. In particular, POSIX shared-memory backing may not be eligible for io_uring fixed-buffer registration and must fall back cleanly.

### LazyMemoryAllocator

Lazy should use the sparse-buffer APIs available since Linux 5.13:

1. Reserve a sparse table based on the final L1 size.
2. Register the initially committed regions before serving requests.
3. For each expansion:
   1. Pin the new host-memory range for CUDA.
[...]

PR #4413, opened 2026-08-04
[Feature][RawBlock] Enable fixed-buffer I/O for MP L1 memory

2xdevv:

Partially addresses #4100 by enabling fixed-buffer I/O for stable MP L1 ranges. Incremental registration as lazy memory grows and splitting I/O across registered-region boundaries remain follow-up work.

The multiprocessing raw-block adapter previously always used ordinary io_uring buffers because it had no safe description of which L1 memory range could remain registered. Registering the final arena capacity directly is unsafe when only a lazy prefix is stable and is inappropriate for POSIX shared memory.

- Extends the L1 memory descriptor with an optional stable registration size.
- Exposes the full anonymous mixed-memory arena, no range for POSIX shared memory, and a bounded snapshot of the current lazy-memory prefix.
- Registers eligible L1 memory when io_uring and zero-copy are enabled, splitting ranges into regions no larger than 1 GiB.
- Falls back to ordinary io_uring when no stable range exists or kernel registration fails.
- Resolves interior buffer views to their registered region and uses fixed read/write SQEs only when the complete request fits within one region.
- Validates registered ranges and serializes registration with device shutdown.
- Adds descriptor, adapter, core, Rust range-resolution, and device integration coverage.

- Requests crossing registered-region boundaries intentionally use ordinary io_uring SQEs.
- Fixed-buffer registration is one-shot and best-effort; failure does not make the adapter unhealthy.

**If applicable**:

### Load performance

A/B testing with `lmcache bench l2`, using a wrapper to supply the production lazy L1 allocator: **20 GiB default initial size, 100 GiB final capacity**. The benchmark’s timed loops and production I/O implementation were unchanged.

Results are median load throughput across five A/B pairs per payload. Each trial’s throughput is measured payload bytes divided by total measured I/O time.

| Payload | Ordinary io_uring (MiB/s) | Fixed buffers (MiB/s) | Improvement |
|---|---:|---:|---:|
| 256 KiB | 4725.8 | 5028.4 | +6.4% |
| 512 KiB | 5750.4 | 8676.5 | +50.9% |
| 1 MiB | 6999.9 | 8485.3 | +21.2% |
| 2 MiB | 6663.0 | 9260.4 | +39.0% |
| 4 MiB | 6186.2 | 9555.5 | +54.5% |
| 8 MiB | 7077.2 | 9317.5 | +31.7% |
| 16 MiB | 7682.4 | 9108.9 | +18.6% |
| 32 MiB | 8085.9 | 9146.9 | +13.1% |

Fixed buffers primarily benefited **loads** in this benchmark. Store throughput improved only **0.4–2.8% at 512 KiB–8 MiB**, was essentially unchanged at **16–32 MiB**, and regressed **4.5% at 256 KiB**.

**Setup and validation**

- Raw-block backend with `O_DIRECT`, regular io_uring, and `enable_zero_copy=true` in both modes.
- 32 keys per submission, eight in-flight submissions, eight load workers, and io_uring queue depth 256.
- Fixed mode used the L1 descriptor’s startup registration snapshot. Ordinary mode changed only `stable_registration_size` to `None`. Registration was verified as **20 GiB versus zero** in every run.
- Both modes completed lazy growth and CUDA registration of the full 100 GiB before warmup. Initialization, registration, growth, and buffer initialization/reset were outside I/O timing.
- Each trial ran in a fresh process. Mode and payload order alternated between pairs. All 80 trials passed data verification and completion checks; fixed-buffer loads were faster in all 40 paired comparisons.

**Scope:** Backing memory was allocated once per trial and reused across rounds. The largest combined store/load allocation was 16 GiB, entirely within the initial 20 GiB registered prefix. These results demonstrate steady-state load benefits with already-populated, CUDA-registered memory; they do not measure allocation churn, I/O outside the registered prefix, or end-to-end MP server performance.


2xdevv:
Hello, @DongDongJu, @zhengfeihe. Could you take a look at this PR? It enables io_uring fixed-buffer support to reduce I/O overhead and improve throughput in the MP raw-block path. This is also discussed in [issue #4100](https://github.com/LMCache/LMCache/issues/4100).

DongDongJu (changes requested):
Hello @2xdevv, Thanks for the continuous working on this issue.

I have two minor comments.

We should make activation observable. 
Registering the default 20 GiB Lazy prefix requires sufficient RLIMIT_MEMLOCK or CAP_IPC_LOCK; otherwise the feature falls back to ordinary io_uring. 

Please expose registered state/bytes in report_status(), document the memlock requirement, and include one MP A/B result confirming the fixed path was active.

2xdevv:
> Hello @2xdevv, Thanks for the continuous working on this issue.
> 
> I have two minor comments.
> 
> We should make activation observable. Registering the default 20 GiB Lazy prefix requires sufficient RLIMIT_MEMLOCK or CAP_IPC_LOCK; otherwise the feature falls back to ordinary io_uring.
> 
> Please expose registered state/bytes in report_status(), document the memlock requirement, and include one MP A/B result confirming the fixed path was active.

Hello, @DongDongJu. 

I’ve made the requested changes. Could you take another look?

DongDongJu (approved):
Hello @2xdevv, Thanks for the fix! LGTM

DongDongJu:
@zhengfeihe Could you take a look? it just enable existing fixed buffer for mp connection.

PR #4191, opened 2026-07-22
feat(cli): Add MLA/TP support for bench-server cli tool 

maobaolong:
**What this PR does / why we need it**:

The existing bench only understood classical K/V (`kv_size=2`) and
a single worker. Running DeepSeek-style MLA models or any TP > 1
setup against the MP server had no local reproducer; regressions
in `lmcache_driven_transfer` / vLLM detector / lookup locking would
only show up in full integration runs.

This PR let `lmcache bench server` to exercise MLA models and simulate
tensor-parallel workers, so we can validate the MP server's MLA +
TP path end-to-end without a real vLLM deployment.

**Special notes for your reviewers**:

- How to test

lmcache server
```shell
/Users/mbl/projects-test/.venv-lmcache/bin/python /Users/mbl/projects-test/LMCache/lmcache/v1/multiprocess/http_server.py --host 127.0.0.1 --port 5599 --l1-size-gb 1 --eviction-policy LRU --http-port 18080 --chunk-size 16 
[2026-07-22 16:17:42,264] LMCache INFO: torch_dev=StubCPUDevice(device_type=cpu), torch_device_type=cpu (__init__.py:249:lmcache.v1.platform)
[2026-07-22 16:17:42,815] LMCache WARNING: Device cpu is not available, using fallback ops. (__init__.py:224:lmcache.v1.platform)
[2026-07-22 16:17:44,192] LMCache INFO: multi_layer_block_kv_transfer mode: tensor (base.py:94:lmcache.v1.multiprocess.transfer_context.base)
[2026-07-22 16:17:44,373] LMCache INFO: Discovered API module: cache_api (router_discovery.py:53:lmcache.v1.utils.router_discovery)
[2026-07-22 16:17:44,406] LMCache INFO: Discovered API module: env_api (router_discovery.py:53:lmcache.v1.utils.router_discovery)
[2026-07-22 16:17:44,408] LMCache INFO: Discovered API module: loglevel_api (router_discovery.py:53:lmcache.v1.utils.router_discovery)
[2026-07-22 16:17:44,409] LMCache INFO: Discovered API module: metrics_api (router_discovery.py:53:lmcache.v1.utils.router_discovery)
[2026-07-22 16:17:44,412] LMCache INFO: Discovered API module: periodic_thread_api (router_discovery.py:53:lmcache.v1.utils.router_discovery)
[...]

PR #4417, opened 2026-08-04
fix(cli): index trace replay percentiles by nearest rank, not by rank

### LINKED ISSUE #4416: [Bug][CLI] trace replay percentiles are one sample too high at whole ranks, so p90/p99 report the max
**Label**

`bug`, `cli`

**Describe the bug**

`_percentile()` in `lmcache/cli/commands/trace/_stats.py` treats the nearest-rank rank as a 0-based index, so it returns one sample too high whenever `pct/100 * N` lands on a whole number.

The docstring on the function says the p-th percentile is the sample at index `ceil(p/100 * N) - 1`. The code drops the `- 1`:

https://github.com/LMCache/LMCache/blob/3b8093cf8860a39d05937af915adfb5db493a047/lmcache/cli/commands/trace/_stats.py#L88-L95

`ceil(p/100 * N)` is a 1-based rank. Using it directly as a list index skips to the next sample.

The practical effect is that `lmcache trace replay` reports the slowest single sample as p90 or p99 at the sample counts you are most likely to hit. With 10 replayed calls for a qualname, p90 equals max. With 100, p99 equals max. Fractional ranks are unaffected, so the numbers look plausible and only the tail is wrong.

There is a second, smaller error in the same expression. `pct / 100.0` is inexact for most integer percentiles, so `(7 / 100.0) * 100` is `7.000000000000001` and `ceil()` picks rank 8. Multiplying before dividing keeps the product exact.

**To Reproduce**

Tested against `origin/dev` at commit `3b8093cf8860a39d05937af915adfb5db493a047`.

```python
from lmcache.cli.commands.trace._stats import ReplayStatsCollector, _percentile

vals = [float(i) for i in range(1, 101)]
print(_percentile(vals, 50))   # 51.0, want 50.0
print(_percentile(vals, 99))   # 100.0, want 99.0  (this is max)
print(_percentile(vals, 7))    # 8.0,   want 7.0   (the float rounding case)

s = ReplayStatsCollector()
for latency_ms in range(1, 11):
    s.record("op.foo", latency_ms / 1000.0)
stats = s.summary()["op.foo"]
print(stats.p90_ms, stats.max_ms)   # 10.0 10.0, want 9.0 10.0
```

Output on `dev`:

```
51.0
100.0
8.0
10.0 10.0
```

**Expected behavior**

Nearest-rank as documented: rank `ceil(p/100 * N)`, 1-based, so index `rank - 1`. p50 of 1..100 is 50, p99 is 99, p90 of ten samples is the ninth.

**Additional context**

The existing unit test encodes the off-by-one rather than catching it. `tests/cli/commands/trace/test_stats.py::TestPercentile::test_p50_on_100_values` asserts `51.0` with the comment `nearest-rank with floor((50/100)*100)=50 → index 50 → value 51`, which is where the 1-based rank got read as a 0-based index. That test needs updating alongside the fix.

Scope check on the two other percentile helpers in the CLI, neither of which should change here:

- `lmcache/cli/commands/bench/engine_bench/stats.py::_percentile` uses linear interpolation on purpose and is documented as such. Not the same method, not affected.
- `lmcache/cli/commands/bench/l2_adapter_bench/result.py` uses a `+0.999999` ceiling approximation. Different code, worth a separate look, out of scope for this.

The changed values are user visible: terminal output from `lmcache trace replay`, plus the `p50_ms`/`p90_ms`/`p99_ms` columns in the CSV and JSON exports.

I have a fix with regression tests ready and will open a PR against `dev` shortly. Happy to change direction if you would rather keep the current numbers and correct the docstring instead, though the tail numbers are wrong either way.

bharadwaj-pendyala:

Fixes #4416

`_percentile()` in the trace replay stats used `ceil(pct/100 * N)` as a list index. That expression is a 1-based rank, so every percentile that lands on a whole rank returned the next sample up. The tail percentiles are the ones that hurt: with 10 recorded calls for a qualname p90 came back as the max, and with 100 calls p99 came back as the max. Fractional ranks were already right, which is why the numbers looked plausible.

The function's own docstring already specifies the correct formula, index `ceil(p/100 * N) - 1`. The code just dropped the `- 1`.

On `dev` at `3b8093c`:

```
_percentile([1.0 .. 100.0], 50)  ->  51.0   want 50.0
_percentile([1.0 .. 100.0], 99)  -> 100.0   want  99.0   (the max)
_percentile([1.0 .. 100.0],  7)  ->   8.0   want   7.0
```

The last one is a second, smaller error in the same expression. `pct / 100.0` is inexact for most integer percentiles, so `(7 / 100.0) * 100` evaluates to `7.000000000000001` and `ceil()` takes rank 8. Multiplying before dividing keeps the product exact. I checked every integer percentile 1 through 99 against `Fraction` arithmetic for `n` up to 100,000 and multiply-first matches exactly, where divide-first is wrong for 141 `(pct, n)` pairs in `n` up to 1000 alone.

`max(rank, 1)` keeps the floor at the first sample. A positive `pct` small enough that `pct * n / 100.0` underflows to `0.0` would otherwise index `[-1]` and return the max.

The existing test asserted the buggy value, so it changes here. `test_p50_on_100_values` had `assert _percentile(vals, 50) == 51.0` with the comment `nearest-rank with floor((50/100)*100)=50 → index 50 → value 51`. That comment is the bug: it reads the 1-based rank as a 0-based index. Updated to `50.0`.

This changes user-visible numbers. Terminal output from `lmcache trace replay` and the `p50_ms`/`p90_ms`/`p99_ms` columns in `trace_replay_ops.csv` and `trace_replay_summary.json` all shift by one rank at whole-rank sample counts. That is the point of the fix, but flagging it in case anyone has stored comparisons.

Deliberately not touched:

- `bench/engine_bench/stats.py::_percentile` uses linear interpolation on purpose and is documented that way. Different method, not a bug.
- `bench/l2_adapter_bench/result.py` has its own nearest-rank variant using a `+0.999999` approximation. Worth a separate look, out of scope here.

I opened #4416 and this PR together rather than waiting, since the docstring already stated the intended behaviour and there was only one way to make the code match it. If you would rather keep the current numbers and correct the docstring instead, say so and I will flip it, though `p99 == max` at N=100 is wrong under either definition.

**If applicable**:

On the docs box: the CLI docs list p50/p90/p99 as fields but never define the percentile method, so there is nothing there stating the old behaviour to correct. Happy to add a line if you want the method written down.

**Validation**

All run locally at base `3b8093c`, source-only install (`NO_NATIVE_EXT`), Python 3.12 on macOS.

Tests, with the fix reverted and the new tests in place (red):

```
FAILED TestPercentile::test_p50_on_100_values
FAILED TestPercentile::test_whole_rank_does_not_advance_to_the_next_sample
FAILED TestPercentile::test_whole_rank_survives_inexact_division
FAILED TestReplayStatsCollector::test_p90_of_ten_samples_is_the_ninth
```

With the fix:

```
pytest tests/cli/ -> 506 passed
ruff check         -> All checks passed!
ruff format --check-> 2 files already formatted
mypy               -> Success: no issues found in 1 source file
```

`tests/cli/commands/trace/test_dispatch.py` and `test_driver.py` do not collect in my environment because they import `lmcache.native_storage_ops`, which a source-only install does not build. Unrelated to this change, and CI covers them.

Edge cases checked by hand: `n=1`, `pct` underflowing to zero, `pct=99.99999999999999`, and an exhaustive sweep of `pct` 1..99 against `n` 1..3000 with zero disagreements against exact `Fraction` arithmetic.


bharadwaj-pendyala:
@yoo-kumaneko you approved #4441 earlier today. This PR is the same defect in the CLI's copy of the helper, `_percentile` in `lmcache/cli/commands/trace/_stats.py`, and it is the older of the two (#4416).

Re-checked against `dev` at 9fc6e1a, where that file has not changed since 3b8093c:

```
_percentile([1.0 .. 100.0], 50)  ->  51.0   want 50.0
_percentile([1.0 .. 100.0], 90)  ->  91.0   want 90.0
_percentile([1.0 .. 100.0], 99)  -> 100.0   want 99.0
_percentile([1.0 ..  10.0], 90)  ->  10.0   want  9.0
```

Whenever `pct * n` is a multiple of 100, the result is one sample too high. At the two sizes where that lands on the last element, p90 over 10 samples and p99 over 100, `lmcache trace stats` prints the slowest call in the trace as the tail latency for that qualname.

The existing `test_p50_on_100_values` asserts `51.0` today, with a comment spelling out the floor-versus-rank mixup. This PR flips that assertion to `50.0` and adds four cases around it. `pytest tests/cli/commands/trace/test_stats.py` gives 15 passed on the branch, and it still merges cleanly into `dev`.

32 lines added, 6 removed, one helper and its test file.

yoo-kumaneko (approved):
LGTM. The code should be corrected while the docstring be kept.

bharadwaj-pendyala:
Thanks @yoo-kumaneko. To confirm on the docstring: it stays exactly as it is. The diff never touches it.

The docstring on `dev` already describes the correct behaviour, "the p-th percentile is the sample at index `ceil(p/100 * N) - 1`". It was the code underneath that did something else. This PR makes the code match what the docstring already claimed, and the only lines removed are the two inline `# bisect_left ...` comments inside the function body, which described the buggy formula.

One thing on merging. You asked for a second approval on #4441, and this one is in the same state, approved by you and still showing `blocked`. Does it need the same second reviewer?

The pending `buildkite/unit-tests-amd` here is not the blocker. Build #10356 has sat at "scheduled" since 5 August, and the same check is still pending on #4437, #4438, #4439, #4442, #4443 and #4444, all of which merged this week. `buildkite/k3-unit-tests` passed on this branch and DCO is green.

bharadwaj-pendyala:
@yoo-kumaneko coming back to the second-approval question from 11 August.

`buildkite/unit-tests-amd` is no longer pending here. Build #10356, the one that had been sitting at "scheduled" since 5 August, started and passed on 14 August:

```
$ gh api repos/LMCache/LMCache/commits/e862b8c7/status
combined=success
  success  buildkite/k3-unit-tests   Build #6337 passed  (21m 45s)
  success  buildkite/unit-tests-amd  Build #10356 passed (19m 52s)
```

So everything on this PR is green now and the only thing missing is the second approval. You raised that ask on #4441 with sammshen, ApostaC, deng451e and hickeyma. Does this one need the same ping, or does it ride along with that request?

Nothing in the diff has moved since you approved: 32 added, 6 removed, across `lmcache/cli/commands/trace/_stats.py` and `tests/cli/commands/trace/test_stats.py`.

bharadwaj-pendyala:
@maobaolong thanks for the approval on #4441 this morning. This PR is the CLI copy of that same off-by-one, and you are the closest reviewer for it: `lmcache/cli/commands/trace/_stats.py` came in with your 3b8093cf (#4191), and you are still the only author the file has.

It has been sitting at 1 of 2 approvals since 9 August. Both checks are green on head `e862b8c7`:

```
$ gh api repos/LMCache/LMCache/commits/e862b8c7/status --jq '.state, (.statuses[] | "\(.context)  \(.state)  \(.description)")'
success
buildkite/k3-unit-tests  success  Build #6337 passed (21 minutes, 45 seconds)
buildkite/unit-tests-amd  success  Build #10356 passed (19 minutes, 52 seconds)
```

The diff has not moved since yoo-kumaneko approved it: 32 added, 6 removed, across `_stats.py` and `tests/cli/commands/trace/test_stats.py`.

Would you take a look? It is the same fix you already read on #4441, in a smaller file.

PR #4420, opened 2026-08-04
[Feat] Add optional uGDS backend for GDS L1 tier

silence-breaker:
# Integrate uGDS as an optional GDS L 1 backend
## **What this PR does / why we need it**:
1. This PR integrates [[uGDS](https://github.com/ScaleX-IO/uGDS)](https://github.com/ScaleX-IO/uGDS) as an optional backend for LMCache's existing GDS L 1 tier. uGDS is a userspace GPUDirect Storage implementation developed by ScaleX. The purpose of this change is to let LMCache use uGDS through its existing GDS L 1 abstraction; it does not change or attempt to optimize LMCache's cuFile implementation.
2. In the SSD-only high-load benchmark, uGDS delivers up to **2.68× TTFT speedup** on cache-hit requests and **2.63×the logical KV retrieval throughput** of NVIDIA cuFile GDS.
3. Why uGDS has a better performance than GDS:  User-space processes issue transfer commands directly to kernel space, eliminating filesystem translation and page cache copy overhead to reduce CPU involvement, while preserving system usability to the greatest extent possible. 
### Architecture
```
vLLM (scheduler / worker)
    │
    ├─ LMCacheMPConnector          (official vLLM KV connector)
    │      │
    │      └─ LMCache adapter      (ZMQ control + CUDA IPC)
    │              │
    ▼              ▼
vLLM paged KV   LMCache MP server
                       │  gather/scatter kernel
                       ▼
                 GPU staging buffer
                       │
                       ├─ cuFileAsync  → ext4 slab file   (GDS)
                       └─ uGDSAsync    → raw block device  (uGDS)
```
### Details: How uGDS Outperforms GDS within LMCache
We comprehensively evaluate uGDS:
#### Environment
| Component | Version |
|-----------|---------|
| GPU | NVIDIA A100-SXM4-40GB |
| SSD | Samsung 990 PRO (PCIe Gen4 x4, cross-root-port P2P) |
| Model | Qwen3-0.6B |
| vLLM | 0.20.1+cu129 |
| LMCache | 0.5.1.dev35 |
| PyTorch | 2.11.0+cu129 |
| CUDA | 12.9 |
#### SSD-only high-load test
The high-load run uses a 40 GiB SSD L1, disables vLLM APC, and requires every measured request to report a full external-cache hit. LMCache STORE and RETRIEVE token totals are also checked before a point is accepted. Context pressure spans 3,840 to 40,704 prompt tokens; concurrency spans 1 to 256 requests with 1,024 tokens per request.
lmcache_e2e_high_load
#### Cold decode throughput
Cold requests where the KV cache STORE (GPU → SSD write) runs concurrently with decode. Prompt lengths 512/1024/2048, output lengths 256 and 1024 tokens, 3 repeats per config, report p50.
lmcache_e2e_decode_throughput

#### Performance breakdown
The matched breakdown follows the difference from raw 16 MiB async reads, through the production-shaped LMCache transaction, to the request-level SSD-to-temp stage. It then changes the API and stream count to test whether the difference is a general storage-bandwidth limit or specific to LMCache's I/O shape.
lmcache_ugds_gds_breakdown
Therefore, under LMCache's current workload profile of a single load-stream and 16 MiB async reads, uGDS sustains a service rate of approximately 5.5 GiB/s with low submission thread utilization.

## **Special notes for reviewers**:
1. **This is a uGDS integration, not a cuFile optimization.** The benchmark results characterize the new backend under the documented hardware and I/O shape. 
2. **uGDS uses a raw device.** Unlike the cuFile/hipFile slab-file backends,   `/dev/ugds_drvX` maps offsets directly to disk LBAs. LMCache does not bind, format, or mount the device, but the configured L 1 range is overwritten. The backend must therefore be used only with a dedicated device whose data may be destroyed. When using the uGDS backend, do not allow multiple LMCache servers to mount the same SSD device concurrently.
3. **The existing backend remains the default.** `--gds-l1-backend=auto` continues to select cuFile on CUDA and hipFile on ROCm. No uGDS library or device is accessed unless `--gds-l1-ackend=ugds` is explicitly requested.
4. **uGDS interface stability.** LMCache currently consumes the following uGDS API surface:
   - `uGDSDriverOpen` / `uGDSDriverClose`
   - `uGDSHandleRegister` / `uGDSHandleDeregister`
   - `uGDSBufRegister` / `uGDSBufDeregister`
   - `uGDSStreamRegister` / `uGDSStreamDeregister`
   - `uGDSReadAsync` / `uGDSWriteAsync`
   As the uGDS maintainers, we will preserve these function signatures, data layouts, ownership rules, stream-ordering behavior, completion semantics, and error semantics across future uGDS releases. New functionality may be added through new APIs. If an incompatible change ever becomes unavoidable, it will use an explicitly versioned interface with a documented migration path instead of silently breaking this LMCache integration. 
1. **Test coverage.** This PR includes tests for backend selection and configuration, uGDS handle/buffer/stream lifecycle, async read/write and cleanup, GDS-context round trips, multiprocess adapter behavior, and benchmark configuration/analysis helpers. The hardware-backed tests require a CUDA GPU and a dedicated NVMe device bound to the uGDS driver.

## **if applicable**:


KuntaiDu (approved):
LGTM. @ApostaC do we have GDS CI?

DongDongJu (changes requested):
Hello @silence-breaker,
Before start the review, How can we enable ugds_drv?
Is that proprietary driver?

silence-breaker:
Thank you very much for your interest and support for uGDS ! The entire project is open-sourced under the BSD 3-Clause License. The source code repository of uGDS is available here: [[uGDS](https://github.com/ScaleX-IO/uGDS)](https://github.com/ScaleX-IO/uGDS). You can start with the official installation documentation: https://github.com/ScaleX-IO/uGDS/blob/main/docs/installation.md.

silence-breaker:
@DongDongJu @KuntaiDu Sorry to bother you . Could you plz review this PR if you're free 😀

andyluo7:
Nice results, and I like that the GDS L1 abstraction absorbed a third backend without
`gds_context.py` having to learn about it. A few things from the AMD/hipFile side, since
I wrote that backend.

**1. `file_location` changes meaning between backends, with nothing validating it.**

For cuFile/hipFile it's a directory and the slab file is created inside it. For uGDS the
same field is opened `O_RDWR` and written to directly:

```python
if self._backend == "ugds":
    self._slab_path = selected
    ...
    fd = os.open(self._slab_path, os.O_RDWR)
```

A directory fails cleanly with `EISDIR`, so the ordinary misconfiguration is safe. The one
that worries me is a valid-but-wrong device: pointing this at `/dev/nvme0n1` or any disk
holding a filesystem will silently destroy it, and there's no way back. Both docstrings say
the path should be `/dev/ugds_drv*`, but nothing enforces it.

Could `_open_and_register_slab` `stat` the path first and require a device node — ideally
one matching the uGDS driver — and fail fast otherwise? It's a few lines and the downside
it prevents is unrecoverable. Minor related nit: `_ugds_async` calls these "raw block
devices" while the config docstring says "character-device path"; worth settling which,
since it determines whether the check is `S_ISBLK` or `S_ISCHR`.

**2. Backend and platform can now be mismatched.**

`_ugds_async` binds `cudaStream_t` and has no `torch.version.hip` guard, but
`_load_backend` will return it on a ROCm build if someone sets `--gds-l1-backend ugds`.
The reverse is now reachable too: `cufile` on ROCm and `hipfile` on CUDA were impossible
before this PR, because dispatch happened at import time off `torch.version.hip`. Now
they're just config values.

A compatibility check in `_load_backend` would turn all three into a clear config error
instead of a dlopen or missing-symbol failure deep in the driver layer. If uGDS is
NVIDIA-only for now, saying so explicitly there is probably the whole fix.

**3. `select_backend` rebinds module globals at runtime.**

This is fine today — `gds_context.py` and its test are the only importers and both do
`import ... as ca`, so they see the rebind. But the shim's correctness now depends on
nobody ever writing `from _gds_async import AsyncHandle`, which would capture the
pre-`select_backend` binding and silently keep using the wrong backend. Worth a line in
the module docstring saying module-style import is required.

Also, should a second `select_backend` call with a *different* name raise? Right now it
rebinds under any live `AsyncHandle`s. The singleton comment in `GDSContext` suggests it
can't happen, in which case failing loudly costs nothing.

**4. @KuntaiDu on GDS CI** — there isn't any hardware coverage. The GDS tests are
unit-level against fakes, and this PR follows that pattern, which is why `_ugds_async`
needs a real device to exercise. That gap now covers three backends; the hipFile path has
never been run in CI either. I have MI300X and MI355X and am happy to run the hipFile GDS
L1 path regularly if that's useful — won't help uGDS, but it would stop the AMD side of
this abstraction from drifting untested.

`use_direct_io` being silently ignored for uGDS is documented in the docstring, but a
warning when it's set together with `ugds` might save someone a confused benchmark.

ApostaC:
Hi @silence-breaker , it would be great if you can also create a page under: https://docs.lmcache.ai/mp/l2_storage/supported_storages.html (docs source file is in `docs/source` folder).

We can explain the details about how to install ugds drivers in that page.

silence-breaker:
@ApostaC Thanks for the suggestion. I want to clarify the storage tier before adding the page: in this PR, uGDS is implemented as a backend of GDSL1MemoryManager and is selected with --gds-l1-backend ugds; it does not register an --l2-adapter type. Therefore, listing it under “Supported L2 Backends” may imply an unsupported configuration. So I add the installation guide link in `docs/source/mp/configuration.rst`😀

silence-breaker:
@andyluo7 Thanks for the detailed review — these were good catches. I’ve pushed a small follow-up addressing the safety and dispatch concerns:

1. Before opening an uGDS slab, LMCache now requires the path to be a character device and verifies through `/sys/dev/char/<major>:<minor>/subsystem` that it is managed by `ugds_drv`. Invalid paths fail before `os.open`, so regular
   files, block devices such as `/dev/nvme0n1`, and unrelated character devices are rejected. I also standardized the terminology to “raw character device.”

2. Backend/platform compatibility is now checked when the backend is explicitly selected:

   - cuFile requires a CUDA PyTorch build;
   - hipFile requires a ROCm PyTorch build;
   - the LMCache uGDS integration is explicitly CUDA-only for now.

   This turns mismatches into clear configuration errors instead of deferred `dlopen` or driver failures. ROCm uGDS support can be handled separately once its complete buffer-registration and async-stream paths are supported and tested.

3. The dispatch module now documents that callers must import it as a module rather than importing individual symbols. After the first explicit backend selection, selecting the same resolved backend remains idempotent, while attempting to switch to another backend raises `RuntimeError`. This prevents live handles from becoming associated with a different set of global backend functions.

4. uGDS now emits a warning when `use_direct_io=True`, since the option is ignored by that backend.

I added negative tests for invalid device nodes and incompatible platform/backend combinations, plus coverage for idempotent selection, rejected backend switching, and the direct-I/O warning. The targeted suite currently reports `46 passed, 4 skipped`; the skipped tests are the real-hardware GDS/uGDS round trips.

silence-breaker:
@andyluo7 Hello, exactly uGDS also supports AMD backend.

chengy-sysu:
Hi @andyluo7,

uGDS supports both NVIDIA and AMD backends. The PR did not reflect that, since the `flags` argument to `uGDSBufRegister` was hardcoded to `0`. Fixed in `f3f77918`.

The flag is now derived from `torch.version.hip`:

```python
def _buf_register_flags() -> int:
    return _UGDS_REGISTER_DMABUF if torch.version.hip is not None else 0
```

`_UGDS_REGISTER_DMABUF` is `UGDS_REGISTER_DMABUF (0x1)` from `ugds.h`. It selects the path uGDS uses to pin GPU memory: setting it takes the AMD path of a dma-buf export plus `dma_buf_pin()`, and clearing it takes the NVIDIA path of `nvidia_p2p_get_pages()`.

The criterion `torch.version.hip` is the same one `_validate_backend_platform()` uses for `cufile` and `hipfile` in `_gds_async.py:79-91`, so all three backends now select the platform on the same basis.

**Testing**

`tests/v1/gpu_connector/test_ugds_async.py::test_register_buffer_passes_pointer_size_and_flags` is now parametrized: it monkeypatches `torch.version.hip` to `None` and to `"6.3"` and asserts the flags passed to `uGDSBufRegister` are `0` and `0x1` respectively.

andyluo7:
Thanks for turning all four around so quickly, and for the correction on AMD — you're
right and I had it backwards. I'd read "the LMCache uGDS integration is explicitly
CUDA-only for now" as the intent, and was going to suggest tightening the guard to
`torch.version.cuda is None`. The `_ugds_async` docstring already says the opposite
(`libugds.so` "built for the active PyTorch platform: CUDA-only for NVIDIA or HIP-only for
AMD ROCm"), so the permissive check is correct as written. Disregard that.

Worth settling the wording somewhere though, since those two statements point in different
directions and the docs page @ApostaC asked for is where people will look. A reader
currently can't tell whether ROCm is supported-and-untested or not-yet-supported.

**The safety fix works.** Re-ran it against real device nodes on an MI300X box:

```
/dev/nvme0n1     -> rejected: uGDS path must be a character device
/etc/hostname    -> rejected: uGDS path must be a character device
/dev/null        -> rejected: uGDS path is not managed by ugds_drv
```

`/dev/null` is the good case — right device type, wrong subsystem, still refused.

Incidentally, on one shared MI355X node all five apparently-idle NVMe drives turned out to
be LVM members backing a mounted `/data`. Exactly the situation where the old code would
have quietly eaten a filesystem, so the check earns its keep.

**I tried to validate the ROCm path end to end and could not.** What I did verify, on
MI355X (gfx950) / ROCm 7.2.4:

- uGDS's HIP backend builds cleanly: `-DUGDS_BACKEND_HIP=ON -DUGDS_BACKEND_CUDA=OFF`.
  Two snags worth a doc line: `find_package(hip)` fails unless you pass
  `-DCMAKE_PREFIX_PATH=/opt/rocm`, and the HIP backend needs CMake >= 3.21, which none of
  the AMD hosts I touched had installed.
- The resulting `libugds.so` contains **no GPU code objects** — it's host-side code linking
  `libamdhip64.so.7`; only uGDS's own test binaries embed kernels. So the ctypes surface
  LMCache uses is architecture-independent, and `CMAKE_HIP_ARCHITECTURES` (which defaults
  to `gfx90a;gfx942`, excluding gfx950) doesn't affect the LMCache integration. Good news,
  but I only know it because I checked, so it may be worth stating.

Where I got stuck is hardware. uGDS takes exclusive ownership of an NVMe controller —
unbind from the kernel `nvme` driver, bind to `ugds_drv` — so exercising it needs a bare
metal host with a spare NVMe controller *and* `CONFIG_HSA_AMD_P2P=y`. Across five AMD
nodes I have access to (MI300X, 2x MI350X, 2x MI355X) none satisfies both: two lack the
P2P kernel config, one has it but every NVMe controller is in a live LVM group, and the
cloud MI350X has no NVMe at all (virtio-only, so nothing to rebind).

So from my side the AMD path is **built but unexercised**. I mention it because the PR
currently reads as though ROCm is supported, and I don't want it to merge looking validated
on AMD when as far as I can tell nobody has run it. What hardware did you use for the HIP
backend? If you have a working setup I'm happy to reproduce against it, and I'll keep
looking for a node with a free NVMe controller on my end.

chengy-sysu:
Hi @andyluo7,

We tested on an MI210 server:

| Item | Value |
|---|---|
| Server | PUERSAI PRI253-D12H |
| GPU | AMD Instinct MI210 (gfx90a) |
| OS / kernel | Ubuntu 22.04, 6.8.0-94-generic |
| System ROCm | 7.1.1 (`/opt/rocm-7.1.1`) |
| PyTorch | 2.9.1+rocm6.3 (pip wheel) |
| Python | 3.10.12 |
| NVMe | Samsung 990 PRO |

This host satisfies all three conditions you listed: bare metal, `CONFIG_HSA_AMD_P2P=y`, and a spare 990 PRO controller that can be unbound from the kernel `nvme` driver and dedicated to `ugds_drv`.

We ran end-to-end LMCache tests on this host, exercising the full L1 cache path this PR touches:

```
bench script -> GDSContext.transfer_async() -> _gds_async.py (backend dispatch, resolves to ugds)
-> _ugds_async.py (ctypes) -> libugds.so -> ugds_drv kernel module -> NVMe P2P DMA
```

32 MB chunks, median of 20 iterations after 3 warmups, with a readback check on every iteration:

| Mode | Depth | Per-IO | BW |
|---|---|---|---|
| Write | 1 | 4.74 ms | 6749 MB/s |
| Write | 4 | 4.68 ms | 6842 MB/s |
| Write | 16 | 4.68 ms | 6845 MB/s |
| Read | 1 | 5.48 ms | 5835 MB/s |
| Read | 4 | 5.50 ms | 5820 MB/s |
| Read | 16 | 5.47 ms | 5845 MB/s |

All passed, with data integrity verified.

andyluo7:
Hi @chengy-sysu , thanks for the data. Looks good.

silence-breaker:
@ApostaC Could you take a look please? Thank you very much.😀

silence-breaker:
@DongDongJu Have you enabled ugds_drv successfully according to the installation guide? Is there any quesions?

ApostaC (approved):
LGTM in general. Please fix the linter issue!

silence-breaker:
Sry, I fix them now. PTAL

DongDongJu (changes requested):
Hello @silence-breaker, from my understanding, it need to umount the other nvme drive with this new drive. Could you add the documentation with warning that it need dedicated SSD for launching this?
And few nits: please check L1 size is whether over device cap or not.
Otherwise LGTM.

silence-breaker:
@DongDongJu Thanks to your advise! We have added warning in docs and the check of SSD capacity in LMCache

DongDongJu:
@silence-breaker UT was failed. PTAL

DongDongJu (approved):
LGTM

Issue #3362, opened 2026-05-21
[MP] AMD/HIP-specific deadlock at TP>=4: hipEventDestroy → SyncAllStreams collides with stream-ordered Python host callback on the GIL


elliotz-ai:
## Summary

LMCache MP mode deadlocks on AMD ROCm at TP >= 4 under sustained load. The same workload runs cleanly on NVIDIA H100. The deadlock is a GIL × HIP-driver-lock AB-BA between the Python thread destructing an interprocess `torch.cuda.Event` and the HIP driver thread that runs a stream-ordered Python host callback. LMCache/LMCache#3336 (merged target) avoids the deadlock at the LMCache level by routing the host callback through C++ with `gil_scoped_release`. This issue documents the AMD-specific observations so AMD can investigate whether the destructor cost in `hipEventDestroy` for interprocess events can be reduced or bounded.

## Environment

- LMCache `v0.4.4` unpatched (issue does not reproduce with LMCache/LMCache#3336 applied)
- vLLM `v0.19.0`
- torch `2.10.0+rocm7.0.2` on ROCm 7.0.2
- MI350X (`gfx950`), 8× per node
- Mistral-Small-24B-Instruct-2501, `--load-format=dummy`, FP8 KV cache, `--enable-prefix-caching`, `--async-scheduling`
- `LMCacheMPConnector` over loopback
- L1 host pool 5 GiB, `--chunk-size 256`

## Repro

1. Start LMCache MP server: `lmcache server --host 127.0.0.1 --port 6555 --l1-size-gb 5.0 --chunk-size 256 --max-workers 4`
2. Start vLLM with `LMCacheMPConnector` over loopback and `--kv-cache-dtype=fp8 --enable-prefix-caching --async-scheduling --tensor-parallel-size N`
3. Drive 2 QPS with a 5k-token prefix per request, `max_tokens=50`

We don't yet have a HIP-only standalone repro independent of LMCache; reproducing currently requires the vLLM + LMCache MP stack as configured above. Happy to work on stripping this down to a HIP-only test if it would help.

## TP-scaling matrix

| Hardware | TP | Result |
|---|---|---|
| MI350X | 2 | clean over 5 min (600 reqs) |
| MI350X | 4 | hangs ~110 s in (180 reqs dispatched at hang) |
| MI350X | 8 | hangs ~70 s in (120 reqs dispatched at hang) |
| H100 80GB | 4 | clean over 5 min (600 reqs, 0 errors) |
[...]

PR #4422, opened 2026-08-04
fix(mp): avoid cross-process Event.from_ipc_handle deadlock on ROCm

### LINKED ISSUE #4419: [MP][ROCm] lmcache_driven deadlocks at TP=1 in Event.from_ipc_handle (import_event) — infinite block, distinct from #3362
## Summary

In the default `lmcache_driven` MP transfer mode (the mode `auto` selects on a HIP device), **vLLM EngineCore deadlocks on the very first retrieve** on AMD ROCm — at **TP=1, single GPU, with no sustained load**. The deadlock is an infinite block inside `torch.cuda.Event.from_ipc_handle` (which internally calls `torch.cuda.Stream.from_ipc_handle`), **not** the GIL × `hipEventDestroy` AB-BA documented in #3362.

This is different from #3362 (which only reproduces at TP>=4 under load, and is fixed at the LMCache level by #3336). The `import_event` path used to *import* the peer process's completion event deadlocks even at TP=1, so #3336 alone does not cover it.

A minimal 8-line ROCm-specific patch in `event_ipc.py` works around it and makes `lmcache_driven` fully functional on AMD, with a real latency win (see below).

---

## Environment

- **Host**: Ubuntu 22.04, 8× AMD MI300X (gfx942, 192 GB each), ROCm 7.0.2 host driver, Docker 29.1.3
- **vLLM image**: `rocm/vllm:rocm7.14.0_cdna_ubuntu24.04_py3.14_pytorch_2.11.0_vllm_0.23.0` (official Docker Hub)
- **Python**: 3.14 (in-image)
- **torch**: 2.11.0+rocm7.14 (in-image)
- **LMCache**: source-built from the `dev` branch (commit `3b8093c`, shallow clone depth=1)
- **GPU**: AMD MI300X (gfx942), **single GPU, TP=1**
- **Model**: `deepseek-ai/DeepSeek-V2-Lite` (MLA, 27 layers, ~16 GB weights)

---

## Step 0: Pull the official ROCm vLLM image

```bash
# On the host with MI300X GPUs
docker pull rocm/vllm:rocm7.14.0_cdna_ubuntu24.04_py3.14_pytorch_2.11.0_vllm_0.23.0
```

Verify it pulled correctly:

```bash
docker images | grep rocm/vllm
# Expected: rocm/vllm  rocm7.14.0_cdna_ubuntu24.04_py3.14_pytorch_2.11.0_vllm_0.23.0  <size>
```

---

## Step 1: Create and enter the Docker container

```bash
docker run -d --name lmcache-repro \
  --device=/dev/kfd --device=/dev/dri \
  --group-add video --cap-add SYS_PTRACE --security-opt seccomp=unconfined \
  --ipc=host --shm-size 16G \
  -v /PATH/TO/huggingface_cache:/root/.cache/huggingface \
  rocm/vllm:rocm7.14.0_cdna_ubuntu24.04_py3.14_pytorch_2.11.0_vllm_0.23.0 \
  sleep infinity
```

> **Important flags explained**:
> - `--device=/dev/kfd --device=/dev/dri --group-add video`: GPU access
> - `--cap-add SYS_PTRACE`: required for `py-spy` later
> - `--ipc=host --shm-size 16G`: shared memory for IPC
> - `-v ...:/root/.cache/huggingface`: mount a **pre-downloaded model cache** (avoids a huge download in the container); the model `deepseek-ai/DeepSeek-V2-Lite` should exist in this cache directory

Verify the container is running:

```bash
docker ps --filter name=lmcache-repro
docker exec lmcache-repro python -c "import vllm; print('vllm', vllm.__version__)"
# Expected: vllm 0.23.1.dev...  (or 0.23.x)
```

Enter the container:

```bash
docker exec -it lmcache-repro bash
# All subsequent commands run inside the container unless noted otherwise
```

---

## Step 2: Check available GPUs and pick a free one

```bash
# On the host (outside the container):
rocm-smi --showmeminfo vram | grep -E "GPU\[|Used"
```

Make sure at least one GPU has plenty of free VRAM (e.g. only 297 MB baseline usage). Pick a GPU index that is **not occupied** by other users. In this reproduction we use **GPU 1** (GPU 0 happened to be taken by another workload).

---

### 3.1 Clone the `dev` branch

```bash
cd /root
git clone --depth 1 --branch dev https://github.com/LMCache/LMCache.git
cd LMCache
git log --oneline -1  # confirm: 3b8093c "feat(cli): Add MLA/TP support..."
```

### 3.2 Check key version constraints

The official ROCm vLLM image ships with **numpy 2.4.4**. LMCache's `requirements/common.txt` has `numpy<=2.2.6`. If you run a full `uv pip install`, uv will try to downgrade numpy to 2.2.6 → triggers a from-source compilation (because 2.2.6 has no Python 3.14 wheel) → fails due to missing `meson-python` → the entire LMCache installation aborts silently.

```bash
python -c "import numpy; print(numpy.__version__)"   # expect 2.4.4
grep 'numpy' requirements/common.txt                   # expect numpy<=2.2.6
```

### 3.3 Build LMCache with `--no-deps` (retain image's numpy/torch/vllm)

```bash
export PYTORCH_ROCM_ARCH=gfx942
export TORCH_DONT_CHECK_COMPILER_ABI=1
export CXX=hipcc
export BUILD_WITH_HIP=1

uv pip install --system -e . --no-build-isolation --no-deps
```

This compiles the HIP `c_ops` extension for `gfx942` but **skips all dependency resolution**, so the image's numpy 2.4.4, torch 2.11.0+rocm7.14, and vllm 0.23.x stay untouched.

> **Why `--no-build-isolation`?** Without it, pip creates a temporary venv for the build and the compiled HIP extension can't find the image's torch/ROCm headers.
>
> **Why `--no-deps`?** Without it, uv downgrades numpy → from-source build → `meson-python` not found → installation fails. The image already has most dependencies.

### 3.4 Verify the build

```bash
python -c 'import lmcache, lmcache.c_ops; print("lmcache", lmcache.__version__, "c_ops OK")'
# Expected: lmcache 0.1.dev1 c_ops OK
# (0.1.dev1 is fine — it's from setuptools_scm on a shallow clone without tags)
```

### 3.5 Install the remaining runtime dependencies

Three packages are needed for MP mode but were skipped by `--no-deps`:

```bash
uv pip install --system aiofile aiofiles sortedcontainers pyzmq msgspec safetensors
uv pip install --system "opentelemetry-exporter-prometheus>=0.50b0"
```

> **Python 3.14 / CuPy note:** `cupy-rocm-7-0` can resolve NumPy 2.5. That version is incompatible with the Numba version used by this environment (NumPy 2.4 or lower is required), and LMCache can then fall back to CLI-only mode before failing to import `device_ops`. After installing or reinstalling ROCm CuPy, run `uv pip install --system --force-reinstall numpy<2.5` before starting the MP server.
>
> **Source-tree dependency note:** `--no-deps` intentionally skips dependencies. The MP e2e needs `aiofile`, `aiofiles`, `sortedcontainers`, `pyzmq`, `msgspec`, `safetensors`, and `opentelemetry-exporter-prometheus`; the commands above install them explicitly.

Verify all critical imports work:

```bash
python -c "import zmq, msgspec, fastapi, httpx, aiofiles, sortedcontainers, redis, safetensors, yaml; print('all deps ok')"
```

---

## Step 4: Start the LMCache MP server

> **Critical**: the MP server **must see all GPUs** so it can resolve the physical GPU UUID that vLLM sends when registering KV caches. Do **not** set `HIP_VISIBLE_DEVICES` on the server.

```bash
export LMCACHE_TRACK_USAGE=false
export LMCACHE_DISABLE_BANNER=1

lmcache server \
  --host localhost --port 5555 \
  --http-host 0.0.0.0 --http-port 8085 \
  --chunk-size 256 \
  --l1-size-gb 20 \
  --eviction-policy LRU \
  > /tmp/mp_server.log 2>&1 &
```

Wait ~15 seconds and verify:

```bash
# Check the server is listening on ZMQ + HTTP
ss -tlnp | grep -E "5555|8085"
# Or: netstat -tlnp | grep -E "5555|8085"

# Check the HTTP metrics endpoint
curl -s -o /dev/null -w "metrics=%{http_code}\n" http://localhost:8085/metrics
# Expected: metrics=200

# Confirm the log shows startup
grep -i "ZMQ cache server is running\|Supported transfer mode" /tmp/mp_server.log
# Expected: Supported transfer mode: auto
#          LMCache ZMQ cache server is running on tcp://localhost:5555
```

---

## Step 5: Start vLLM with LMCacheMPConnector

> **Critical**: vLLM must be constrained to **one free GPU** (e.g. GPU 1) with `HIP_VISIBLE_DEVICES=1`. The MP server must see all GPUs (no constraint), but vLLM should only use the free one.

```bash
export HIP_VISIBLE_DEVICES=1        # use only GPU 1 (physical)
export VLLM_USE_V1=1
export SAFETENSORS_FAST_GPU=1
export LMCACHE_TRACK_USAGE=false

vllm serve deepseek-ai/DeepSeek-V2-Lite \
  --trust-remote-code \
  --max-model-len 8192 \
  --gpu-memory-utilization 0.4 \
  --disable-hybrid-kv-cache-manager \
  --no-enable-prefix-caching \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both","kv_connector_extra_config":{"lmcache.mp.host":"tcp://localhost","lmcache.mp.port":5555}}' \
  > /tmp/vllm_mp.log 2>&1 &
```

> **Why `--disable-hybrid-kv-cache-manager` and `--no-enable-prefix-caching`?** DeepSeek-V2-Lite uses MLA attention which requires these flags. Disabling internal prefix caching also makes the external LMCache hit the *only* source of cache hits — cleaner evidence.

Wait ~60-100 seconds for model loading + CUDA graph capture:

```bash
# Check health
curl -s -o /dev/null -w "health=%{http_code}\n" --max-time 5 http://localhost:8000/health
# Expected: health=200

# Check the log for successful startup
grep -i "Application startup complete\|Registering kv caches" /tmp/vllm_mp.log
# Expected: Registering kv caches!
#          Application startup complete.

# Confirm the transfer mode is "auto" (→ lmcache_driven on HIP)
grep -i "mp_transfer_mode" /tmp/vllm_mp.log | tail -1
# Expected: lmcache.mp.mp_transfer_mode = auto

# Confirm MP server registered the KV caches via lmcache_driven
grep -i "Registered KV cache\|lmcache_driven" /tmp/mp_server.log | tail -3
# Expected: Registered KV cache for GPU ID <uuid> with 27 layers (lmcache_driven_transfer.py)
```

---

### 6.1 Prepare a request with a long shared prefix

First request (cold — this will STORE successfully):

```bash
# Inside the container:
PREFIX=$(python3 -c "
block = 'LMCache is a KV cache layer for LLM serving. In multiprocess mode, an independent LMCache server process holds the KV cache, and vLLM connects to it as a client over ZMQ, transferring KV tensors via CUDA/HIP IPC with zero copy. This decouples cache lifetime from the inference engine and enables cache sharing. '
print('You are a meticulous technical assistant specializing in GPU inference systems. Background you must remember:\n\n' + block * 85)
")

# Send cold request (A)
curl -s -X POST http://localhost:8000/v1/completions \
  -d "{\"model\":\"deepseek-ai/DeepSeek-V2-Lite\",\"prompt\":$(python3 -c "import json; print(json.dumps('$PREFIX'))"),\"max_tokens\":32,\"temperature\":0.0}" \
  -o /tmp/rA.json

python3 -c "import json; print('A usage:', json.load(open('/tmp/rA.json')).get('usage'))"
# Expected: {'prompt_tokens': 5989, 'total_tokens': 6021, 'completion_tokens': 32}

# Check MP server — store should have succeeded
grep -i "Stored\|lmcache_mp_l1_memory_usage" /tmp/mp_server.log | tail -2
curl -s http://localhost:8085/metrics | grep -E "lmcache_mp_l1_memory_usage_bytes|lmcache_mp_lookup_hit_tokens_total" | grep -v "#"
# Expected: Stored 5888 tokens in 0.065s
#          lmcache_mp_l1_memory_usage_bytes > 0
#          lmcache_mp_lookup_hit_tokens_total = 0
```

### 6.2 Send the SAME request again (retrieve — this WILL deadlock)

```bash
# Send identical request (C) — this will hang forever
timeout 120 curl -s -X POST http://localhost:8000/v1/completions \
  -d "{\"model\":\"deepseek-ai/DeepSeek-V2-Lite\",\"prompt\":$(python3 -c "import json; print(json.dumps('$PREFIX'))"),\"max_tokens\":32,\"temperature\":0.0}" \
  -o /tmp/rC.json

# The above command will time out after 120s — C never returns
```

> **Expected deadlock behavior**:
> - The curl command hangs indefinitely (the 120s timeout will kill it)
> - `rC.json` is never written
> - After ~130s, the MP server logs: `Reaped GPU instance ... silent for ~130s (pinged=True)`
> - vLLM health endpoint still returns 200 (API server is alive, but EngineCore is wedged)

---

### 7.1 Install py-spy

```bash
uv pip install --system py-spy
```

### 7.2 Find the wedged EngineCore PID and dump its stack

```bash
PID=$(ps aux | grep "VLLM::EngineCore" | grep -v grep | awk '{print $2}' | head -1)
echo "EngineCore PID: $PID"

# Check process state — should be Rl (runnable/busy, not sleeping)
ps -p $PID -o pid,stat,wchan,comm

# Dump the Python call stack
py-spy dump --pid $PID --nonblocking 2>&1
```

### 7.3 Expected py-spy output (the deadlock stack)

```
Process <PID>: VLLM::EngineCore
Python v3.14.6 (...)

Thread <PID> (active+gil)
    from_ipc_handle          (torch/cuda/streams.py:197)   <-- BLOCKS FOREVER
    import_event             (lmcache/v1/platform/base/event_ipc.py:225)
    _on_raw_future_complete  (lmcache/v1/multiprocess/futures.py:134)
    query                    (lmcache/v1/multiprocess/futures.py:203)
    get_finished             (lmcache/integration/vllm/vllm_multi_process_adapter.py:1609)
    get_finished             (lmcache/integration/vllm/lmcache_mp_connector.py:984)
    _get_kv_connector_output (vllm/v1/worker/kv_connector_model_runner_mixin.py:103)
    __exit__                 (contextlib.py:148)
    execute_model            (vllm/v1/worker/gpu_model_runner.py:4276)
    ...
```

### 7.4 Interpret the stack

1. vLLM `execute_model` finishes, enters `_get_kv_connector_output`
2. This calls `get_finished` on the retrieve future
3. The future's `_on_raw_future_complete` tries to **import the MP server's completion event** via `import_event`
4. `import_event` calls `torch.cuda.Event.from_ipc_handle(device, handle)`
5. **`from_ipc_handle` blocks forever on ROCm** — the cross-process event IPC ABI is unsupported or hangs

The code that blocks:

```python
# lmcache/v1/platform/base/event_ipc.py, line 222-225
class DefaultEventIPCBackend:
    def import_event(self, handle: bytes, device: object) -> object:
        return self._event_module.Event.from_ipc_handle(  # ← hangs here on ROCm
            device, handle
        )
```

Compare with the store path (which succeeds): store only does a local `event.record(stream)` — no cross-process import needed.

---

### 8.1 The patch

Edit `lmcache/v1/platform/base/event_ipc.py`, in the `DefaultEventIPCBackend` class, replace the `import_event` method:

```python
# BEFORE (original, deadlocks on ROCm):
def import_event(self, handle: bytes, device: object) -> object:
    """Reconstruct an event from an IPC handle on ``device``."""
    return self._event_module.Event.from_ipc_handle(  # type: ignore[union-attr]
        device, handle
    )

# AFTER (patched, works on ROCm):
def import_event(self, handle: bytes, device: object) -> object:
    """Reconstruct an event from an IPC handle on ``device``.

    ROCm workaround (PRE_HIP_PATCH): on AMD ROCm,
    torch.cuda.Event.from_ipc_handle blocks indefinitely in the
    lmcache_driven multiprocess path, so cross-process event IPC
    cannot be used. Fall back to synchronizing the current stream
    (which orders the IPC KV-copy on this device) and return a
    local already-signaled event so downstream synchronize/query
    calls stay correct.
    """
    import torch
    if getattr(torch.version, "hip", None) is not None:
        torch.cuda.current_stream(device).synchronize()
        return self._event_module.Event()
    return self._event_module.Event.from_ipc_handle(
        device, handle
    )
```

> **What the patch does**: On ROCm, instead of importing the peer's event (which hangs), it synchronizes the current stream on this device (flushing all pending GPU ops, including the IPC KV-copy). It returns a local already-signaled event so that the downstream `synchronize_event`/`query_event` calls work correctly.

### 8.2 Restart everything with the patch

```bash
# Kill old processes
pkill -9 -f vllm
pkill -9 -f "VLLM::EngineCore"
pkill -9 -f "lmcache server"
sleep 3

# Verify clean
ps aux | grep -E "vllm|lmcache server|EngineCore" | grep -v grep | grep -v defunct
# Should be empty

# Clear old logs
rm -f /tmp/vllm_mp.log /tmp/mp_server.log

# Restart MP server (same command as Step 4)
lmcache server \
  --host localhost --port 5555 \
  --http-host 0.0.0.0 --http-port 8085 \
  --chunk-size 256 --l1-size-gb 20 --eviction-policy LRU \
  > /tmp/mp_server.log 2>&1 &

# Wait for server
sleep 15
curl -s -o /dev/null -w "metrics=%{http_code}\n" http://localhost:8085/metrics
# Expected: metrics=200

# Restart vLLM (same command as Step 5)
HIP_VISIBLE_DEVICES=1 VLLM_USE_V1=1 SAFETENSORS_FAST_GPU=1 \
vllm serve deepseek-ai/DeepSeek-V2-Lite \
  --trust-remote-code --max-model-len 8192 --gpu-memory-utilization 0.4 \
  --disable-hybrid-kv-cache-manager --no-enable-prefix-caching \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both","kv_connector_extra_config":{"lmcache.mp.host":"tcp://localhost","lmcache.mp.port":5555}}' \
  > /tmp/vllm_mp.log 2>&1 &

# Wait for vLLM (60-100s)
sleep 90
curl -s -o /dev/null -w "health=%{http_code}\n" --max-time 5 http://localhost:8000/health
# Expected: health=200
```

### 8.3 Re-run the cold + repeat test

```bash
# Cold (A) — same as before, should still work
PREFIX=$(python3 -c "...")  # same as Step 6.1
curl -s -X POST http://localhost:8000/v1/completions \
  -d "{\"model\":\"deepseek-ai/DeepSeek-V2-Lite\",\"prompt\":$(python3 -c "import json; print(json.dumps('$PREFIX'))"),\"max_tokens\":32,\"temperature\":0.0}" \
  -o /tmp/rA.json

# Repeat (C) — this time it should RETURN, not hang
time curl -s -X POST http://localhost:8000/v1/completions \
  -d "{\"model\":\"deepseek-ai/DeepSeek-V2-Lite\",\"prompt\":$(python3 -c "import json; print(json.dumps('$PREFIX'))"),\"max_tokens\":32,\"temperature\":0.0}" \
  -o /tmp/rC.json

echo "C exit code: $?"
# Expected: C exit code: 0, latency ~0.36s (vs cold ~1.4s)
```

### 8.4 Verify the cache hit

```bash
# MP server — retrieve should have succeeded
grep -i "Stored\|Retrieved" /tmp/mp_server.log | tail -4
# Expected:
#   Stored 5888 tokens in 0.065s (lmcache_driven_transfer.py)
#   Retrieved 5888 tokens in 0.030s (lmcache_driven_transfer.py)

# MP metrics — should show hit
curl -s http://localhost:8085/metrics | grep -E "lmcache_mp_l1_memory_usage_bytes|lmcache_mp_lookup_hit_tokens_total" | grep -v "#"
# Expected:
#   lmcache_mp_l1_memory_usage_bytes 1.83140352e+08     (183 MB)
#   lmcache_mp_lookup_hit_tokens_total{...} 5888.0

# py-spy — EngineCore should be back to idle
PID=$(ps aux | grep "VLLM::EngineCore" | grep -v grep | awk '{print $2}' | head -1)
py-spy dump --pid $PID --nonblocking 2>&1 | head -8
# Expected: Thread <PID> (idle) — wait (queue.py:199)
```

---

## Observed behavior (before patch)

- First (cold) request: store succeeds (`Stored 5888 tokens in 0.065s`, `lmcache_driven_transfer.py`), curl returns normally
- Second (identical) request for retrieve: **never returns**; after ~130s the MP server `Reaped` the GPU instance (`silent for ~130s, pinged=True`); curl times out
- No `memory access fault` — the process is truly alive (state `Rl` runnable) but stuck in `from_ipc_handle`
- This is NOT the same as #3362 (which only reproduces at TP>=4 under load)

## Root cause

`lmcache/v1/platform/base/event_ipc.py` — `DefaultEventIPCBackend.import_event` calls `torch.cuda.Event.from_ipc_handle(device, handle)`, which internally calls `torch.cuda.Stream.from_ipc_handle`. On AMD ROCm, this cross-process event/stream IPC import blocks indefinitely.

The store path works because it only does a local `event.record(stream)` — no cross-process import is needed. The retrieve path must import the peer process's completion event to order the GPU copy, and that import never returns.

The module docstring already acknowledges this: *"The `lmcache_driven` multiprocess handle path orders cross-process KV-cache transfers with device events. Accelerators expose different event IPC ABIs."* On AMD that ABI is unsupported for `from_ipc_handle`.

## Minimal workaround

```python
def import_event(self, handle, device):
    import torch
    if getattr(torch.version, "hip", None) is not None:
        torch.cuda.current_stream(device).synchronize()
        return self._event_module.Event()
    return self._event_module.Event.from_ipc_handle(device, handle)
```

## Result after the workaround (default `lmcache_driven`, TP=1)

| Metric | Source | Cold (A) | Repeat (C) |
|---|---|---|---|
| STORE / RETRIEVE | MP server log | `Stored 5888 tokens in 0.065s` | `Retrieved 5888 tokens in 0.030s` |
| `lmcache_mp_lookup_hit_tokens_total` | MP metrics | 0 | 5888 |
| `lmcache_mp_l1_memory_usage_bytes` | MP metrics | 0 | 1.83e8 (183 MB) |
| request latency | curl | ~1.4 s | 0.36 s (~4× faster) |
| GPU memory faults | both logs | 0 | 0 |

After the patch, py-spy shows EngineCore back to `idle` (waiting on the input queue). Notably, `lmcache_driven` (zero-copy HIP IPC) yields a real latency win here (~4× vs cold), whereas the `engine_driven` fallback mode is slower than a cold request due to gather/scatter overhead.

## Additional pitfall: MP server GPU visibility

If the MP server has `HIP_VISIBLE_DEVICES` constrained (e.g. to a single GPU), it may fail to resolve the physical GPU UUID that vLLM reports during KV-cache registration. The error is:

```
Device UUID ... not found in the discovered devices
```

**Fix**: the MP server must see **all GPUs** (do not set `HIP_VISIBLE_DEVICES` on the server). Only constrain vLLM's GPU visibility. The server then iterates all visible GPUs by UUID to find the matching physical device for each IPC handle.

## Ask

1. Is `Event.from_ipc_handle` on peer events expected to work on ROCm, or is the event IPC ABI unsupported there? (cc @ROCm — this looks like the same root defect family as #3362.)
2. Should `check_event_support` / `import_event` degrade gracefully on HIP rather than hang indefinitely? A backend-neutral HIP branch in `event_ipc.py` would let AMD users run the default `lmcache_driven` mode.

## References

- #3362 — AMD/HIP deadlock at TP>=4 (`hipEventDestroy → SyncAllStreams` GIL AB-BA); the `import_event` hang here is a separate, TP=1-reproducible failure of the same HIP event IPC ABI
- LMCache/LMCache#3336 — LMCache-side fix for #3362 (does not cover `import_event`)
- Related installation issues affecting AMD users: vllm-project/vllm#50953 (same `--no-deps` / `sortedcontainers` / `opentelemetry` pain points on ROCm)

maobaolong:
## Problem

On AMD ROCm/HIP builds, `Event.from_ipc_handle` internally calls `Stream.from_ipc_handle`, which **blocks indefinitely** in cross-process contexts — even at TP=1 on a single GPU. This breaks the default `lmcache_driven` transfer path on **all ROCm deployments**.

Closes #4419.

### Root Cause

Call chain during `lmcache_driven` retrieve:

```
lmcache_driven_transfer.py: retrieve()
  -> future.result()
    -> _on_raw_future_complete()
      -> DefaultEventIPCBackend.import_event()
        -> torch.cuda.Event.from_ipc_handle(device, handle)
          -> torch.cuda.Stream.from_ipc_handle()  <- DEADLOCK on ROCm
```

On NVIDIA GPUs, `cudaIpcOpenEventHandle` works correctly. On AMD ROCm, the HIP equivalent may block when the event was created by a different process. This is distinct from #3362 (hipEventDestroy GIL deadlock at TP>=4, already fixed in #3336) — #4419 happens at TP=1.

### Fix

1. Add a `_is_rocm()` helper (cached via `@functools.cache`, checks `torch.version.hip`)
2. In `DefaultEventIPCBackend.import_event`, when running on ROCm, use a **synchronize-and-trigger-local-event fallback**:
   - `torch.cuda.current_stream(device).synchronize()` — ensures all GPU work (including the exporting process's) is visible
   - Create and record a local interprocess event
   - Synchronize it (mark as triggered)
   - Return it so downstream `wait_event` is a no-op

### Correctness Argument

On a single physical GPU shared across processes (the typical MP deployment):
- Both processes share the same GPU hardware
- `current_stream(device).synchronize()` waits for *all* work submitted to that device — including the exporting process's
- When synchronize returns, the data is guaranteed visible
- The locally-triggered event is just a formality for the code path

**Limitation**: Multi-GPU / multi-node setups should continue using `engine_driven` transfer until ROCm's `Stream.from_ipc_handle` stabilises for cross-process event import.

### Verification

Tested on AMD MI300X (gfx942), ROCm 7.14, PyTorch 2.11, vLLM 0.23, DeepSeek-V2-Lite:

| Mode | Before Fix | After Fix |
|------|-----------|-----------|
| lmcache_driven retrieve | Deadlock (GPU memory fault / 300s timeout) | 5888 tokens in 0.030s |
| Repeated request latency | N/A | 0.36s vs cold 1.40s (~4x speedup) |
| GPU memory faults | 1+ per retrieve | 0 |

The fix is minimal (net +45/-1 lines in one file) and has **zero effect on CUDA/non-ROCm paths**: the `_is_rocm()` guard ensures the fallback only activates on HIP builds.

### DIFF
--- a/lmcache/v1/platform/base/event_ipc.py
+++ b/lmcache/v1/platform/base/event_ipc.py
@@ -17,10 +17,31 @@ from __future__ import annotations
 
 # Standard
 from typing import Protocol, runtime_checkable
+import functools
 import inspect
 
 # First Party
 from lmcache import torch_dev, torch_device_type
+
+
+@functools.cache
+def _is_rocm() -> bool:
+    """Return ``True`` when the active PyTorch build targets ROCm/HIP.
+
+    When PyTorch is compiled for ROCm, ``torch.version.hip`` is set to the
+    HIP version string (e.g. ``"7.2.53211-..."``).  For CUDA builds the
+    attribute is absent or ``None``.  The result is cached so the check
+    runs at most once per process.
+
+    This helps ``DefaultEventIPCBackend`` avoid cross-process
+    ``Event.from_ipc_handle`` deadlocks on AMD hardware (see #4419).
+    """
+    try:
+        # Third Party
+        import torch
+    except ImportError:
+        return False
+    return getattr(torch.version, "hip", None) is not None
 
 
 def _accepts_interprocess(event_cls: object) -> bool:
@@ -220,7 +241,32 @@ class DefaultEventIPCBackend:
         return event.ipc_handle()  # type: ignore[attr-defined]
 
     def import_event(self, handle: bytes, device: object) -> object:
-        """Reconstruct an event from an IPC handle on ``device``."""
+        """Reconstruct an event from an IPC handle on ``device``.
+
+        On ROCm/HIP builds, ``Event.from_ipc_handle`` internally calls
+        ``Stream.from_ipc_handle``, which may block indefinitely in
+        cross-process contexts even at TP=1 (refs #4419).  The fallback
+        synchronises the current stream on the target device — which
+        guarantees that all prior GPU work (including the exporting
+        process's) is visible — and returns a locally-triggered event so
+        that downstream ``wait_event`` calls are no-ops.
+
+        This is semantically correct on a single physical GPU shared
+        across processes; multi-GPU / multi-node setups should prefer
+        ``engine_driven`` transfer until ROCm's ``Stream.from_ipc_handle``
+        stabilises for cross-process use.
+        """
+        if self.device_type == "cuda" and _is_rocm():
+            # Third Party
+            import torch
+
+            torch.cuda.current_stream(device).synchronize()
+            event = self._event_module.Event(  # type: ignore[union-attr]
+                interprocess=True
+            )
+            event.record(torch.cuda.current_stream(device))
+            event.synchronize()
+            return event
         return self._event_module.Event.from_ipc_handle(  # type: ignore[union-attr]
             device, handle
         )


andyluo7:
Took a look since I've got AMD hardware handy. I don't think the fallback preserves the ordering the event was doing.

The producer records on its compute stream and ships the handle without syncing (`lmcache_mp_connector.py:585` → `worker_transfer.py:540`), so the `wait_event` on the consumer side is the only thing stopping LMCache from reading KV before vLLM has finished writing it. The fallback hands back an already-completed event, so that wait becomes a no-op.

Wrote a small two-process repro — share the buffer once, pass only the event handle, long kernel then write. Counting fresh process-pairs that read unwritten data, in `vllm/vllm-openai-rocm:v0.26.0` (torch 2.11.0+gitd0c8b1f, ROCm 7.2.3):

| | MI300X gfx942 | MI355X gfx950 |
|---|---|---|
| no ordering at all | 12/12 | 11/12 |
| this PR's fallback | 12/12 | 10/12 |
| real `from_ipc_handle` + `wait_event` | 0/12 | 0/12 |

So the fallback measures the same as no sync. Also `current_stream()` isn't the stream LMCache actually works on — that's `cache_context.stream` (`platform/cuda/cache_context.py:414`). And retrieve has the same shape as store (`:1339`, `worker_transfer.py:615`), so both directions are affected.

Other thing: I couldn't get `from_ipc_handle` to hang at all, 24/24 clean across both cards. And in torch 2.11 it just forwards to the C++ base, there's no `Stream.from_ipc_handle` hop:

```python
@classmethod
def from_ipc_handle(cls, device, handle):
    return super().from_ipc_handle(device, handle)
```

My repro isn't the real worker/server topology though, so I could just be missing the trigger. Can you post the py-spy stack and your arch/ROCm/torch versions? Happy to repro properly and help get a fix in — I just don't think the ordering can be faked locally.

Minor: the fallback's `Event(interprocess=True)` doesn't need to be IPC-capable since nothing exports it, and destroying those is the `hipEventDestroy` stall from ROCm/rocm-systems#7520 (fixed on CLR `develop`, not in a release yet).

maobaolong:
@andyluo7 Thanks for the verification on your side, and https://github.com/LMCache/LMCache/pull/4640 this is my newer fixed PR, and there is an error stack and full reproduce steps, if convince, would you like to help to reproduce that issue in your amd node?

PR #4424, opened 2026-08-04
[Operator] Remove Hardcoded CacheBlend webhook-injected vLLM params

aeon-x:
## Summary

Stops the CacheBlend mutating webhook from injecting three vLLM flags it hardcoded as Go constants, and adds one typed CR field for a connector tunable that genuinely needs operator support.

**No longer injected:**

| Flag | Was | Why removed |
|---|---|---|
| `--attention-backend CUSTOM` | hardcoded | not the operator's to set |
| `--block-size 64` | hardcoded | not the operator's to set |
| `--no-async-scheduling` | hardcoded | the MoE-garble issue it worked around is resolved |

**Added:** `spec.blend.partialBucket` (int32, optional, min 1) — emitted as `kv_connector_extra_config["cb.partial_bucket"]` in the connection ConfigMap. Unset omits the key (padding disabled).

## Tests

- **Webhook (unit + envtest):** the injected flag set is asserted individually, so a change to it fails the suite.
- **`internal/resources`:** `blend.partialBucket` set → `cb.partial_bucket` in the ConfigMap JSON; unset → key omitted.
- **e2e smoke test:** drops the injected-args assertion and keeps the `Using AttentionBackendEnum.CUSTOM backend` log-banner check — a stronger end-to-end signal than asserting the flag we used to pass.

`go build`, `go vet`, `gofmt`, and the full `go test ./...` are clean; Sphinx builds `operator.rst` without new warnings.


docs/source/mp/operator.rst
```diff
@@ -997,10 +997,12 @@ not reach ``vllm serve``:
...
 private-image pull secret, and the required CacheBlend vLLM flags
-(``--attention-backend CUSTOM``, ``--kv-transfer-config`` from the engine's
-connection ConfigMap, ``--block-size 64``, ``--pipeline-parallel-size 1``,
-``--no-enable-chunked-prefill``, ``--no-async-scheduling``, ``--enforce-eager``).
-You supply only the model and your non-CacheBlend flags.
+(``--kv-transfer-config`` from the engine's connection ConfigMap,
+``--pipeline-parallel-size 1``, ``--no-enable-chunked-prefill``,
+``--enforce-eager``).  You supply only the model and your non-CacheBlend flags.
+
+``--attention-backend`` and ``--block-size`` are **not** injected; a value you
```
royyhuang: No need to mention this.

operator/internal/webhook/cacheblend_inject_builders.go
```diff
@@ -57,24 +57,16 @@ var cbStaging = payloadStaging{
 }
 
 // CacheBlend-required vLLM flag names and fixed values (design §7 M5). The
-// CacheBlend matcher and connector hard-require these; several fail loudly,
-// --no-async-scheduling fails silently (MoE garble).
+// CacheBlend matcher and connector hard-require these. --attention-backend and
+// --block-size are deliberately absent: the operator does not manage them.
```
royyhuang: Similarly, no need to mention this. Also check and remove similar sentences, code, and tests in all places. No need for the users or other developers to know this.

royyhuang (changes requested):
Overall LGTM. Some nits below.

docs/source/mp/operator.rst
aeon-x: removed

royyhuang (approved):
LGTM

deng451e (approved):
LGTM

PR #4429, opened 2026-08-05
[MP] Skip NIXL poll sleep on terminal state

nayeonikim:

The async completion loop in the NIXL store L2 adapters (static NixlStorageAgent.post_non_blocking and dynamic DynamicNixlStorageAgent._post_non_blocking) called await asyncio.sleep(0.01) unconditionally on every iteration, including the iteration where check_xfer_state() had already returned DONE or ERR. As a result, any transfer that started in a pending state paid one extra fixed 10ms delay after it had actually completed, before the loop exited.

This PR guards the sleep so it only runs while the transfer is still pending. When the state check reports a terminal state, the loop now exits immediately instead of yielding for another 10ms.

**If applicable**:

### DIFF
--- a/lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
+++ b/lmcache/v1/distributed/l2_adapters/nixl_store_dynamic_l2_adapter.py
@@ -33,6 +33,7 @@ import uuid
 # Third Party
 from nixl._api import nixl_agent as NixlAgent
 from nixl._api import nixl_agent_config as NixlAgentConfig
+from nixl._api import nixl_xfer_handle as NixlXferHandle
 from nixl._api import (
     nixlBind,
 )
@@ -315,7 +316,7 @@ class DynamicNixlStorageAgent:
             return os.path.join(self.file_path, _object_key_to_relpath(key))
         return os.path.join(self.file_path, _object_key_to_filename(key))
 
-    async def _post_non_blocking(self, handle):
+    async def _post_non_blocking(self, handle: NixlXferHandle) -> None:
         """Await a nixl transfer until done."""
         state = self.nixl_agent.transfer(handle)
         while state != "DONE" and state != "ERR":
@@ -323,7 +324,8 @@ class DynamicNixlStorageAgent:
                 state = self.nixl_agent.check_xfer_state(handle)
             except nixlBind.nixlBackendError:
                 raise
-            await asyncio.sleep(0.01)
+            if state != "DONE" and state != "ERR":
+                await asyncio.sleep(0.01)
         if state == "ERR":
             raise RuntimeError("NIXL transfer failed")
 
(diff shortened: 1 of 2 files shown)


DongDongJu (commented):
Hello @nayeonikim, Thanks for the catch.
Can we have any performance number for this perf patch?

nayeonikim:
> Hello @nayeonikim, Thanks for the catch. Can we have any performance number for this perf patch?

@DongDongJu I ran a local microbenchmark against a real NIXL POSIX backend; this is
not a cluster or end-to-end run. The measured interval is from transfer
submission through terminal-state observation.

The benchmark covers 480 transfers (static/dynamic x store/load x
4 KiB/64 KiB/1 MiB, 40 runs each). All transfers initially returned `PROC`,
so they exercised the polling path changed by this patch. For the `before`
control, I restored the pre-patch unconditional sleep after the terminal-state
check.

| agent / op    | payload | before p50 | after p50 | delta     |
|---------------|--------:|-----------:|----------:|----------:|
| static store  |   4 KiB |  20.212 ms | 10.125 ms | 10.087 ms |
| static load   |   4 KiB |  20.196 ms | 10.120 ms | 10.076 ms |
| dynamic store |   4 KiB |  20.275 ms | 10.210 ms | 10.065 ms |
| dynamic load  |   4 KiB |  20.252 ms | 10.169 ms | 10.083 ms |
| static store  |  64 KiB |  20.236 ms | 10.154 ms | 10.082 ms |
| static load   |  64 KiB |  20.192 ms | 10.112 ms | 10.080 ms |
| dynamic store |  64 KiB |  20.309 ms | 10.234 ms | 10.075 ms |
| dynamic load  |  64 KiB |  20.251 ms | 10.189 ms | 10.062 ms |
| static store  |   1 MiB |  20.962 ms | 10.818 ms | 10.144 ms |
| static load   |   1 MiB |  20.677 ms | 10.591 ms | 10.086 ms |
| dynamic store |   1 MiB |  20.696 ms | 10.635 ms | 10.061 ms |
| dynamic load  |   1 MiB |  20.437 ms | 10.363 ms | 10.074 ms |

The p50 reduction is 10.06-10.14 ms in every case, independent of payload
size, matching the single 10 ms polling sleep skipped after observing a
terminal state.

The approximately 49% relative reduction is specific to this low-latency
local POSIX setup. On a slower backend, the relative improvement would be
smaller, while the patch still removes approximately one polling interval.

DongDongJu (approved):
Hello @nayeonikim, Thanks for the reply.
LGTM.

DongDongJu:
@maobaolong @zhengfeihe PTAL

zhengfeihe (approved):
LGTM!

(Might cause chunxiao to rebase his pr after merge this
https://github.com/LMCache/LMCache/pull/4531

zhengfeihe:
CI shall be fixed by this https://github.com/LMCache/LMCache/pull/4724

PR #4430, opened 2026-08-05
[Fix][NIXL] Report capacity-limited dynamic stores as failed

2xdevv:

Dynamic NIXL store batches can reach capacity after earlier keys have already been written. The loop stopped at the first rejected key but left the task marked successful, allowing callers to treat skipped keys as persisted.

- Mark the store task unsuccessful when the capacity check stops a batch.
- Preserve completed writes and existing usage accounting when later keys are skipped.
- Add regression coverage for the task result, usage, files, and lookup visibility of a partially stored batch.

- This preserves the adapter's existing partial-batch behavior: completed keys remain stored when a later key is rejected.
- The result contract reports zero transferred bytes for failed tasks, while retained partial writes remain reflected in usage accounting.

**If applicable**:


DongDongJu (approved):
LGTM!

zhengfeihe (approved):
LGTM!

PR #4431, opened 2026-08-05
refactor(kv_format): move EngineKVFormat facts onto KVFormatSpec

maobaolong:
## What

Move the `EngineKVFormat` classification out of the call sites and onto the
per-format `KVFormatSpec`.

`torch_ops.py` had the same format lists written down over and over: `is_mla`
in three places, HND in three places (`_is_hnd_format`, the cross-layer
helper, the fused helper), plus a scatter of inline `fmt == SOME_FORMAT`
checks in the transfer helpers. They were already slightly out of sync with
each other, and adding a format meant grepping for all of them.

## How

Each spec now declares its own static facts as class attributes (default
`False`, so a spec only states what applies to it):

- structural shape: `is_cross_layer` / `is_kv_list` / `is_layer_list`
- modifiers: `is_mla`, `is_hnd`, `is_fused_packed`, `is_two_major`,
  `is_pbs_fused`

```python
class NL_X_NB_NH_BS_CS_Spec(KVFormatSpec):
    engine_kv_format = lmc_ops.EngineKVFormat.NL_X_NB_NH_BS_CS
    attention_backends = ("vLLM non-MLA blocks-first, fused K/V (HND)",)
    is_layer_list = True
    is_hnd = True
    is_fused_packed = True
```

Everything else reads them through `get_spec_class(fmt)`; the four public
`torch_ops` predicates (`is_cross_layer` / `is_kv_list` / `is_layer_list` /
`is_mla`) keep their names and signatures and are now one-line lookups, so the
`c_ops` parity contract is unchanged. `csrc/engine_kv_format.h` keeps its copy —
the device kernels need it — and the classification test now pins the two sides
against each other.

`get_spec_class` indexes by enum *value*, because the native pybind
`EngineKVFormat` and the pure-Python fallback one are different types with the
same members and both reach the table.

## Notes

- `_is_cross_layer_format` / `_is_sglang_mha_format` / `_is_mla_format` were
  pure aliases of the public predicates; dropped in favour of the public ones.
- Python-side `is_hnd` is now the honest fact ("heads before block tokens"), so
  it also covers the cross-layer and fused HND formats. The dispatch in
  `multi_layer_block_kv_transfer` is unaffected (cross-layer and fused are
  matched by earlier branches); the local HND flags in the cross-layer and fused
  helpers now come from the same fact instead of their own lists.
- `_per_layer_paged_shape` still switches on the format: that one renders a
  *paged buffer* shape rather than answering a yes/no fact, so it is left alone.

## Test

- `pytest tests/v1/gpu_connector tests/v1/test_torch_ops.py tests/v1/test_c_ops_parity.py tests/v1/test_mp_mem_kernels.py tests/v1/platform tests/cli` — pass
- new: `test_spec_facts_match_predicates` (spec vs `c_ops`) and
  `test_spec_only_facts_match_golden` (golden table for the spec-only facts)
- `pre-commit run --all-files` — all Python hooks pass (the Rust clippy hook
  fails on macOS because `io-uring` is Linux-only, unrelated to this change)

### DIFF
--- a/csrc/engine_kv_format.h
+++ b/csrc/engine_kv_format.h
@@ -153,15 +153,16 @@ enum class EngineKVFormat : int {
 // the two sides can never drift; the predicates below are one-line lookups.
 // Exactly one structural shape (cross_layer / kv_list / layer_list) is true per
 // format. The remaining flags are modifiers layered on top of it.
+// Every field defaults to false; the table below only sets the true ones.
 struct FormatFacts {
-  bool is_cross_layer;   // all layers in one fused tensor
-  bool is_kv_list;       // keys and values in two top-level lists
-  bool is_layer_list;    // one list entry per layer
-  bool is_mla;           // MLA: single latent KV head (no separate K/V)
-  bool is_hnd;           // heads before block tokens (HND layout)
-  bool is_fused_packed;  // K/V packed in trailing dim (kv_size == 1)
-  bool is_two_major;     // size-2 K/V axis precedes the block axis
-  bool is_pbs_fused;     // paged buffer size fused into one axis
+  bool is_cross_layer = false;   // all layers in one fused tensor
+  bool is_kv_list = false;       // keys and values in two top-level lists
+  bool is_layer_list = false;    // one list entry per layer
+  bool is_mla = false;           // MLA: single latent KV head (no separate K/V)
+  bool is_hnd = false;           // heads before block tokens (HND layout)
+  bool is_fused_packed = false;  // K/V packed in trailing dim (kv_size == 1)
+  bool is_two_major = false;     // size-2 K/V axis precedes the block axis
+  bool is_pbs_fused = false;     // paged buffer size fused into one axis
 };
 
 // clang-format off
--- a/csrc/engine_kv_format.h
+++ b/csrc/engine_kv_format.h
@@ -148,49 +148,85 @@ enum class EngineKVFormat : int {
   #define LMC_KV_FORMAT_HD
 #endif
 
-// Structural shape of the normalized kv_caches: exactly one is true per format.
+// Static layout facts for each format, declared once per format (indexed by the
+// enum value). This mirrors the facts declared on each Python KVFormatSpec so
+// the two sides can never drift; the predicates below are one-line lookups.
+// Exactly one structural shape (cross_layer / kv_list / layer_list) is true per
+// format. The remaining flags are modifiers layered on top of it.
+struct FormatFacts {
+  bool is_cross_layer;  // all layers in one fused tensor
+  bool is_kv_list;      // keys and values in two top-level lists
+  bool is_layer_list;   // one list entry per layer
+  bool is_mla;          // MLA: single latent KV head (no separate K/V)
+  bool is_hnd;          // heads before block tokens (HND layout)
+  bool is_fused_packed; // K/V packed in trailing dim (kv_size == 1)
+  bool is_two_major;    // size-2 K/V axis precedes the block axis
+  bool is_pbs_fused;    // paged buffer size fused into one axis
+};
+
+// clang-format off
+LMC_KV_FORMAT_HD constexpr FormatFacts FORMAT_FACTS[] = {
+    /*  0 NB_NL_TWO_BS_NH_HS      */ {true,  false, false, false,
+                                      false, false, false, false},
+    /*  1 NL_X_TWO_NB_BS_NH_HS    */ {false, false, true,  false,
+                                      false, false, true,  false},
+    /*  2 NL_X_NB_TWO_BS_NH_HS    */ {false, false, true,  false,
+                                      false, false, false, false},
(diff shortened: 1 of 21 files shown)


chunxiaozheng (approved):
@maobaolong This looks more clear, thanks for your contribution, LGTM!

sammshen:
as mentioned, we can clean up the helpers in: https://github.com/LMCache/LMCache/blob/dev/csrc/engine_kv_format.h

maobaolong:
@sammshen PR updated, I cleaned the `[engine_kv_format.h](https://github.com/LMCache/LMCache/blob/dev/csrc/engine_kv_format.h)` . PTAL.

csrc/engine_kv_format.h
```diff
@@ -28,116 +28,37 @@ len(kv_cache) = A
...
+  X(NL_X_TWO_NB_NH_BS_HS,   false, false, true,  false, true,  false, true,  false) \
+  X(NL_X_NB_TWO_NH_BS_HS,   false, false, true,  false, true,  false, false, false) \
+  X(NB_NL_TWO_NH_BS_HS,     true,  false, false, false, true,  false, false, false) \
+  X(TWO_X_NL_X_NB_BS_NH_HS, false, true,  false, false, false, false, false, false) \
+  X(NL_X_NB_NH_BS_TWO_HS,   false, false, true,  false, true,  true,  false, false) \
+  X(NL_X_NB_BS_NH_TWO_HS,   false, false, true,  false, false, true,  false, false) \
+  X(NL_X_NB_NH_BS_CS,       false, false, true,  false, true,  true,  false, false) \
+  X(NL_X_NB_BS_NH_CS,       false, false, true,  false, false, true,  false, false) \
+  X(NL_X_NB_BSV_BSS,        false, false, true,  true,  false, false, false, false)
+// clang-format on
```
ApostaC: This is not maintainable & readable. I prefer to have the old & separated definition of each format with an explanation of what it is for.

Regarding the `FormatFacts`, we can also have an explicit list.

ApostaC (commented):
Quick comment, PTAL

csrc/engine_kv_format.h
maobaolong: Ok, I revert the changes and use a `explicit list` for `FORMAT_FACTS`

maobaolong:
@ApostaC Thanks for the review. I've made the changes based on your comment and PTAL.

csrc/engine_kv_format.h
```diff
@@ -148,49 +148,74 @@ enum class EngineKVFormat : int {
...
+struct FormatFacts {
+  bool is_cross_layer;   // all layers in one fused tensor
+  bool is_kv_list;       // keys and values in two top-level lists
+  bool is_layer_list;    // one list entry per layer
+  bool is_mla;           // MLA: single latent KV head (no separate K/V)
+  bool is_hnd;           // heads before block tokens (HND layout)
+  bool is_fused_packed;  // K/V packed in trailing dim (kv_size == 1)
+  bool is_two_major;     // size-2 K/V axis precedes the block axis
+  bool is_pbs_fused;     // paged buffer size fused into one axis
+};
```
ApostaC, line 166: To be safer, we probably need to have a default value for every field.
In some C++ compilers, if we do not specify the default value, there might be some undefined behavior.

lmcache/v1/platform/torch_ops.py
```diff
@@ -741,82 +739,70 @@ def multi_layer_kv_transfer_unilateral(
...
+    """
+    # Imported lazily, not at module scope: the specs package reads
+    # ``lmcache.c_ops``, whose shim is installed only once this module (the
+    # torch baseline behind it) has been imported.
+    # First Party
+    from lmcache.v1.gpu_connector.kv_format.specs.registry import get_spec_class
+
+    return get_spec_class(engine_kv_format)
+
+
```
ApostaC: nit: let's put the helper function to the top part of the source file (i.e., before the public main functions)

ApostaC (approved):
I like this version! 

Just a quick follow up: I think we have a bunch of `is_xxx` defined. We probably should double check how many of them is "really" used. (Maybe they are all useful, but we can clear some of them if it's not used)

csrc/engine_kv_format.h
maobaolong: done

lmcache/v1/platform/torch_ops.py
maobaolong: done

maobaolong:
> I like this version!
> 
> Just a quick follow up: I think we have a bunch of `is_xxx` defined. We probably should double check how many of them is "really" used. (Maybe they are all useful, but we can clear some of them if it's not used)

All the nits have been done.

@ApostaC FYI: For the question you motioned, I did a research and find out the `is_two_major` and `is_pbs_fused` are only used by `KVFormatSpec` and `torch_ops.py` and test scope. For other `is_xxx` is really used external.

PR #4432, opened 2026-08-05
[Perf][P2P] Slice NIXL L1 memory registration to avoid the >4 GiB destination-MR throughput cliff

BoJiang03:

RDMA reads whose **destination** lies in a single NIXL memory registration
larger than ~4 GiB run at ~55% of line rate on our 100 GbE RoCE testbed
(mlx5, 4 KB-page registration, IOMMU on): ~6.0 GB/s instead of ~9.3 GB/s.
A production L1 buffer is tens of GB and is registered as one region, so the
MP-server P2P path always sits in the slow regime.

Controlled experiments isolating the variable (transfer channel benchmark,
zero GPU, two nodes):

| read | destination MR layout | GB/s |
|---|---|---|
| 3.62 GB | 6.5 GiB × 1 | 6.00 |
| 3.62 GB | **3.5 GiB × 2 (same total bytes)** | **9.24** |
| 5.89 GB | 5.9 GiB × 1 | 6.02 |
| 5.89 GB | **3.5 + 2.4 GiB slices** | **9.32** |
| 5.89 GB | 24 GiB × 1 | 6.04 |
| 5.89 GB | 24 GiB in 3.5 GiB slices | 9.24 |

Three properties pin down the mechanism as **per-MR translation depth** on
the requesting NIC (not total pinned memory, not read size):

- Same read size and same *total* registered bytes; only the slicing differs
  (rows 1–2).
- The penalty is a step, not a slope: a 24 GiB single MR is no slower than a
  5.9 GiB one (rows 3 vs 5).
- It is asymmetric: inflating the *source* MR is free; only the destination
  (where the NIC writes) matters.

**The fix**: register the L1 buffer as multiple page-aligned slices of at
most `mr_slice_bytes` instead of one region. Slice boundaries are rounded
down to `align_bytes` multiples so no transfer page straddles two
registrations, and NIXL/UCX transparently splits reads that span a slice
boundary (verified with the benchmark's `--verify` data check) — the
page-index math and the descriptor list are completely untouched, so the
whole change is confined to the registration call in
`NixlTransferChannelContext.__init__`.

Configuration is layered:

- Mechanism (`NixlTransferChannelContext(mr_slice_bytes=...)`): default `0`
  = single region, behavior unchanged.
- Policy (`P2PConfig.mr_slice_bytes`, CLI `--p2p-mr-slice-bytes`): default
  **3.5 GiB = slicing ON** for the MP server's P2P path.
- The transfer channel benchmark gains `--mr-slice-bytes` (accepts `3.5GB`)
  so the effect can be measured directly.

Measured impact, three levels of the same stack (all A/B with the flag as
the only delta):

| level | before | after |
|---|---|---|
| raw channel, 5.89 GB read | 6.0 GB/s | **9.3 GB/s (+55%)** |
| MP-server P2P, per request (40K-token doc: lookup incl. wire + retrieve) | ~1236 ms | **~900 ms (−27%)** |
| full stack: 2-node vLLM + `lmcache bench engine` long-doc-qa, p50 TTFT | 1545.8 ms | **1197.8 ms (−22.5%)** |

- The ~4 GiB threshold is presumably NIC/IOMMU-dependent (we measured mlx5 /
  100 GbE / IOMMU on / 4 KB pages). 3.5 GiB was the fastest measured safe
  size on this hardware; since slicing a small buffer is harmless (a <3.5 GiB
  buffer still gets one region), a conservative default seemed better than
  auto-detection. `--p2p-mr-slice-bytes 0` restores the old behavior.
- Only the requester/destination side benefits, but the default applies to
  both peers (each side registers its own L1 once and both may act as
  readers).
- Scope: this touches only the MP-server transfer channel
  (`lmcache/v1/distributed/transfer_channel/`). The legacy
  `lmcache/v1/transfer_channel/nixl_channel.py` (P/D-disagg path) has the
  same single-large-MR pattern and would need the same treatment; left out
  deliberately to keep this PR focused.
- `nixl_storage_backend.py` already registers per-page and is unaffected.

**Reproduction A — isolated channel (two nodes, no GPU needed;
`CUDA_VISIBLE_DEVICES=` works)**:

```bash
# node A (source)
python -m lmcache.tools.transfer_channel_benchmark --role server \
  --transfer-channel-type nixl --url 0.0.0.0:7600 --control-url 0.0.0.0:7610 \
  --buffer-size $((6144*1048576)) --page-size 65536 \
  --object-size $((36*1048576)) --num-objects 156 --num-source-objects 156

# node B (destination/requester) — run once per arm:
#   arm 1: --mr-slice-bytes 0      (single MR,   ~6 GB/s here)
#   arm 2: --mr-slice-bytes 3.5GB  (sliced,      ~9.3 GB/s here)
python -m lmcache.tools.transfer_channel_benchmark --role client \
  --transfer-channel-type nixl --url <nodeA>:7600 --control-url <nodeA>:7610 \
  --listen-url <nodeB>:7601 --page-size 65536 --object-size $((36*1048576)) \
  --num-objects 156 --iters 10 --warmup 2 --seed 0 --verify \
  --mr-slice-bytes 3.5GB
```

**Reproduction B — full-stack TTFT A/B (two nodes, one GPU each;
this is the −22.5% p50 TTFT experiment)**:

Design: warm node A's L1 with 8 × 40K-token documents through vLLM-A, then
run the *same* documents (same `--seed`) against vLLM-B — every lookup on B
misses locally and P2P-loads the full document (~5.9 GB) from A. P2P-loaded
keys are retained as temporary (deleted after the read), so **every query
re-pays the full wire cost**, which makes TTFT a direct read of wire
throughput. The two arms differ ONLY in server-B's `--p2p-mr-slice-bytes`;
restart server-B *and* vLLM-B between arms (fresh L1 + fresh connector),
leave A untouched.

All LMCache/vLLM processes need `PYTHONHASHSEED=0` (MP mode).

```bash
# 0) coordinator (runs anywhere; here: node B)
lmcache coordinator --host 0.0.0.0 --port 9300

# 1) node A — MP server + vLLM (will hold the warm KV cache).
#    L1 must fit all docs below the eviction trigger watermark (default 0.8):
#    8 docs x 40K tok x 147456 B/tok (Qwen3-8B) = 47.2 GB -> 64 GB L1 (69% full).
lmcache server --host 0.0.0.0 --port 5560 --http-port 8080 \
  --l1-size-gb 64 --eviction-policy LRU --chunk-size 256 \
  --p2p-advertise-url <nodeA-ip>:9400 --p2p-transfer-engine nixl \
  --coordinator-url http://<nodeB-ip>:9300 --coordinator-advertise-ip <nodeA-ip>

vllm serve Qwen/Qwen3-8B --port 8000 --max-model-len 40960 \
  --gpu-memory-utilization 0.3 \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both",
    "kv_connector_extra_config":{"lmcache.mp.host":"tcp://localhost","lmcache.mp.port":5560}}'

# 2) warm A (1 query/doc stores the docs into A's L1), then verify no eviction:
lmcache bench engine --engine-url http://<nodeA-ip>:8000 --model Qwen3-8B \
  --workload long-doc-qa --no-interactive --seed 42 \
  --tokens-per-gb-kvcache 6781 --kv-cache-volume 47.2 \
  --ldqa-document-length 40000 --ldqa-query-per-document 1 \
  --ldqa-num-inflight-requests 1 --ldqa-max-output-length 16
curl -s http://<nodeA-ip>:8080/metrics | grep -E "chunks_total|usage_ratio"
#   -> lmcache_mp_l1_write_chunks_total = 1248 (8 docs x 156 chunks),
#      lmcache_mp_l1_usage_ratio ~= 0.69, and NO lmcache_mp_l1_evicted_chunks_total
#      line (the counter is only exported after the first eviction)

# 3) node B — MP server + vLLM. Start B only AFTER the warm pass finished
#    (kv_role is kv_both, so a live B would also save the docs it prefills
#    and cross-seed the fleet). ONE arm per run; between arms stop BOTH,
#    then relaunch. Arm "off" adds `--p2p-mr-slice-bytes 0`; arm "on" uses
#    the default (3.5 GiB slices).
lmcache server --host 0.0.0.0 --port 5561 --http-port 8081 \
  --l1-size-gb 24 --eviction-policy LRU --chunk-size 256 \
  --p2p-advertise-url <nodeB-ip>:9401 --p2p-transfer-engine nixl \
  --coordinator-url http://<nodeB-ip>:9300 --coordinator-advertise-ip <nodeB-ip> \
  # --p2p-mr-slice-bytes 0        # <- arm "off" only

# --gpu-memory-utilization on B is LOAD-BEARING: vLLM-B's own KV pool must be
# SMALLER than the doc set (here ~29 GB < 47.2 GB), otherwise vLLM's built-in
# prefix cache retains the docs after the first query and later queries never
# reach LMCache/P2P at all. 0.3 on a 96 GB card ~= 29 GB; scale to your VRAM.
vllm serve Qwen/Qwen3-8B --port 8001 --max-model-len 40960 \
  --gpu-memory-utilization 0.3 \
  --kv-transfer-config '{"kv_connector":"LMCacheMPConnector","kv_role":"kv_both",
    "kv_connector_extra_config":{"lmcache.mp.host":"tcp://localhost","lmcache.mp.port":5561}}'

# 4) the measured pass: same seed => same docs => every query P2P-loads from A
lmcache bench engine --engine-url http://<nodeB-ip>:8001 --model Qwen3-8B \
  --workload long-doc-qa --no-interactive --seed 42 \
  --tokens-per-gb-kvcache 6781 --kv-cache-volume 47.2 \
  --ldqa-document-length 40000 --ldqa-query-per-document 3 \
  --ldqa-num-inflight-requests 1 --ldqa-max-output-length 16 \
  --output-dir out-<arm> --json
# TTFT (the headline metric): out-<arm>/**/bench_summary.json -> results.p50_ttft_ms
```

TTFT is the primary metric and needs nothing extra. The per-transfer wire
GB/s we quote additionally came from a small logging hook around NIXL's
`get_xfer_telemetry` in the channel client (not part of this PR); the TTFT
delta reproduces without it.

Our numbers (Qwen3-8B, 2× dual-socket nodes, 100 GbE RoCE, 24 queries/arm):
p50 TTFT 1545.8 → 1197.8 ms, wire median 4.97 → 6.79 GB/s. A second,
independent run of this recipe exactly as written reproduced it within
noise: p50 TTFT 1542.6 → 1207.1 ms (−21.8%), wire 5.02 → 6.39 GB/s.

**If applicable**:

      (`docs/design/v1/distributed/transfer_channel/overall.md`: registration
      contract + slicing rationale)
      (`test_nixl_impl.py`: cross-slice-boundary read with data verification,
      validation errors; `test_config.py`: flag parsing/default/disable)

### DIFF
--- a/tests/v1/distributed/transfer_channel/test_nixl_impl.py
+++ b/tests/v1/distributed/transfer_channel/test_nixl_impl.py
@@ -234,3 +234,65 @@ def test_connecting_registers_clients_on_both_sides(fresh_contexts):
 def test_server_is_available_from_context(shared_contexts):
     ctx_a, _, _, _ = shared_contexts
     assert ctx_a.get_transfer_channel_server() is not None
+
+
+# =========================================================
+# MR slicing (mr_slice_bytes)
+# =========================================================
+def test_sliced_registration_serves_read_across_slice_boundary():
+    """With mr_slice_bytes=1024 the 4096-byte buffer registers as 4 slices;
+    a read spanning the destination's first slice boundary must still land
+    intact (nixl splits transfers across registration slices transparently).
+    """
+    ctx_a = ctx_b = None
+    buf_a = torch.zeros(_BUF_SIZE, dtype=torch.uint8)
+    buf_b = torch.arange(0, 256, dtype=torch.uint8).repeat(_BUF_SIZE // 256)
+    desc_a = L1MemoryDesc(ptr=buf_a.data_ptr(), size=buf_a.numel(), align_bytes=_ALIGN)
+    desc_b = L1MemoryDesc(ptr=buf_b.data_ptr(), size=buf_b.numel(), align_bytes=_ALIGN)
+    url_a, url_b = _next_url(), _next_url()
+    try:
+        ctx_a = NixlTransferChannelContext(
+            desc_a, listen_url=url_a, advertise_url=url_a, mr_slice_bytes=1024
+        )
+        ctx_b = NixlTransferChannelContext(
+            desc_b, listen_url=url_b, advertise_url=url_b, mr_slice_bytes=1024
+        )
+        client = ctx_a.get_transfer_channel_client(ctx_b.advertise_url)
+
+        # [768, 1792) crosses the slice boundary at 1024 on both sides.
+        local = ctx_a.get_transfer_channel_address([(768, 1024)])
+        remote = ctx_b.get_transfer_channel_address([(768, 1024)])
+        task_id = client.submit_read(local, remote)
+
+        result = _wait_finished(client, task_id)
+        assert result.is_finished() is True
+        assert result.succeeded_mask == [True]
+        assert torch.equal(buf_a[768:1792], buf_b[768:1792])
+        # Regions outside the read stay untouched.
+        assert torch.count_nonzero(buf_a[:768]) == 0
+        assert torch.count_nonzero(buf_a[1792:]) == 0
+    finally:
+        if ctx_a is not None:
+            ctx_a.close()
+        if ctx_b is not None:
+            ctx_b.close()
+
+
+def test_mr_slice_bytes_negative_raises_value_error():
+    buf = torch.zeros(_BUF_SIZE, dtype=torch.uint8)
+    desc = L1MemoryDesc(ptr=buf.data_ptr(), size=buf.numel(), align_bytes=_ALIGN)
+    url = _next_url()
+    with pytest.raises(ValueError):
+        NixlTransferChannelContext(
+            desc, listen_url=url, advertise_url=url, mr_slice_bytes=-1
+        )
+
+
+def test_mr_slice_bytes_below_alignment_raises_value_error():
+    buf = torch.zeros(_BUF_SIZE, dtype=torch.uint8)
+    desc = L1MemoryDesc(ptr=buf.data_ptr(), size=buf.numel(), align_bytes=_ALIGN)
+    url = _next_url()
+    with pytest.raises(ValueError):
+        NixlTransferChannelContext(
+            desc, listen_url=url, advertise_url=url, mr_slice_bytes=_ALIGN - 1
+        )
--- a/tests/v1/distributed/transfer_channel/test_nixl_impl.py
+++ b/tests/v1/distributed/transfer_channel/test_nixl_impl.py
@@ -257,6 +257,12 @@ def test_sliced_registration_serves_read_across_slice_boundary():
         ctx_b = NixlTransferChannelContext(
             desc_b, listen_url=url_b, advertise_url=url_b, mr_slice_bytes=1024
         )
+        # Slicing must actually have happened -- guard against a regression
+        # that ignores mr_slice_bytes and registers one region (a single-MR
+        # read across [768, 1792) would pass the data checks below too).
+        assert ctx_a.num_mr_slices == _BUF_SIZE // 1024
+        assert ctx_b.num_mr_slices == _BUF_SIZE // 1024
+
         client = ctx_a.get_transfer_channel_client(ctx_b.advertise_url)
 
         # [768, 1792) crosses the slice boundary at 1024 on both sides.
--- a/docs/design/v1/distributed/transfer_channel/overall.md
+++ b/docs/design/v1/distributed/transfer_channel/overall.md
@@ -98,7 +98,7 @@ The normal way application code obtains a context:
 
 | Function | Purpose |
 |---|---|
-| `initialize_transfer_channel_context(transfer_channel_type, l1_memory_desc, listen_url, advertise_url, **kwargs)` | Create the process-global context via the registered factory. Raises if one already exists, or if the type is unknown. `**kwargs` are forwarded to the factory (e.g. `backends=["UCX"]` for nixl). |
+| `initialize_transfer_channel_context(transfer_channel_type, l1_memory_desc, listen_url, advertise_url, **kwargs)` | Create the process-global context via the registered factory. Raises if one already exists, or if the type is unknown. `**kwargs` are forwarded to the factory (e.g. `backends=["UCX"]` and `mr_slice_bytes` for nixl). |
 | `get_transfer_channel_context()` | Return the global context (raises if not initialized). |
 | `delete_transfer_channel_context()` | Close and clear the global context. |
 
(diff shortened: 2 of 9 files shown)


tests/v1/distributed/transfer_channel/test_nixl_impl.py
```diff
@@ -234,3 +234,65 @@ def test_connecting_registers_clients_on_both_sides(fresh_contexts):
 def test_server_is_available_from_context(shared_contexts):
     ctx_a, _, _, _ = shared_contexts
     assert ctx_a.get_transfer_channel_server() is not None
+
+
+# =========================================================
+# MR slicing (mr_slice_bytes)
+# =========================================================
+def test_sliced_registration_serves_read_across_slice_boundary():
```
DongDongJu, line 242: would still pass if mr_slice_bytes were completely ignored and the buffer were registered as one MR.

DongDongJu (changes requested):
Hello @BoJiang03, Thanks for the great perf improvement!!
I really like your problem definition / resolving way.
code changes are generally LGTM. I left the one test nit.

Also, one qq is that was this 4gib limit based on platform?
if yes, can you change it to the knob in overall this patch?
I would also consider naming it nixl_mr_slice_bytes, since this is currently a NIXL-specific workaround.

BoJiang03:
> Hello @BoJiang03, Thanks for the great perf improvement!! I really like your problem definition / resolving way. code changes are generally LGTM. I left the one test nit.
> 
> Also, one qq is that was this 4gib limit based on platform? if yes, can you change it to the knob in overall this patch? I would also consider naming it nixl_mr_slice_bytes, since this is currently a NIXL-specific workaround.

Thanks for the review! All three points are addressed in the follow-up commit.

- Test nit: you're right. The slice math is now extracted into a pure function compute_mr_slice_regions() with its own hardware-free test file, so the partition invariants actually run in CI (the nixl integration tests are skipped there by importorskip). The cross-boundary test now also asserts num_mr_slices == 4 through a new public property. I mutation-tested this: an implementation that ignores mr_slice_bytes fails at that assertion.
- Naming: renamed to P2PConfig.nixl_mr_slice_bytes / --p2p-nixl-mr-slice-bytes. The mechanism-layer parameter on NixlTransferChannelContext keeps its name since it already lives in the nixl namespace.
- Yes, the ~4 GiB limit is platform-dependent, and it is already a knob rather than a hard-coded constant: --p2p-nixl-mr-slice-bytes (default 3.5 GiB, the fastest safe size measured on our mlx5/100GbE testbed; 0 disables slicing entirely). Happy to flip the default to 0 if you'd prefer opt-in behavior on unmeasured hardware.

PR #4434, opened 2026-08-05
[MP] feat:support lazy offload

chunxiaozheng:
Support lazy offload.

the log as follows, the offload threshold is 2, and offload 1 request every time:
image
image
image
image
image

### DIFF
--- a/lmcache/integration/vllm/lazy_offload_pending_store.py
+++ b/lmcache/integration/vllm/lazy_offload_pending_store.py
@@ -126,6 +126,8 @@ class FIFOOffloadPolicy(OffloadPolicy):
                 to_offload.append(self._pending_items[req_id])
                 del self._pending_items[req_id]
                 self._finished_requests_count -= 1
+            if len(to_offload) >= count:
+                break
         return to_offload
 
 
@@ -156,12 +158,13 @@ class LazyOffloadPendingStore:
         else:
             raise ValueError(f"Unknown offload policy: {policy}")
 
-        # TODO: support more flexible select count
+        # TODO(chunxiaozheng): support more flexible select count
         self._select_count = (
             configs.get("lmcache.mp.lazy_offload_select_count", 10) if configs else 10
         )
 
-        # TODO: use gpu block pool to judge should offload and select items
+        # TODO(chunxiaozheng): use gpu block pool to judge should offload
+        #  and select items
         # GPU block pool reference
         self._gpu_block_pool: "BlockPool | None" = None
 
--- a/lmcache/integration/vllm/lazy_offload_pending_store.py
+++ b/lmcache/integration/vllm/lazy_offload_pending_store.py
@@ -140,9 +140,9 @@ class LazyOffloadPendingStore:
 
         # TODO: use gpu block pool to judge should offload and select items
         # GPU block pool reference
-        self._gpu_block_pool: BlockPool | None = None
+        self._gpu_block_pool: "BlockPool | None" = None
 
-    def bind_gpu_block_pool(self, gpu_block_pool: BlockPool) -> None:
+    def bind_gpu_block_pool(self, gpu_block_pool: "BlockPool") -> None:
         """Bind the GPU block pool to the pending store."""
         self._gpu_block_pool = gpu_block_pool
 
--- a/lmcache/integration/vllm/lazy_offload_pending_store.py
+++ b/lmcache/integration/vllm/lazy_offload_pending_store.py
@@ -1,0 +1,160 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# Standard
+from abc import ABC, abstractmethod
+from collections.abc import Iterator
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+# First Party
+from lmcache.utils import init_logger as lmcache_init_logger
+
+if TYPE_CHECKING:
+    # First Party
+    from lmcache.integration.vllm.lmcache_mp_connector import LMCacheMPRequestMetadata
+
+
+logger = lmcache_init_logger(__name__)
+
+
+@dataclass
+class PendingStoreItem:
+    """
+    Represents a pending store operation in the lazy offload queue.
+
+    Attributes:
+        metadata: The store metadata to be submitted.
+    """
+
+    metadata: "LMCacheMPRequestMetadata"
+
+
+# TODO: support more offload policies
+class OffloadPolicy(ABC):
+    """
+    Abstract base class for lazy offload policies.
+
+    Subclasses define when to trigger offload (should_offload) and
+    which items to return (select_items).
+    """
+
+    @abstractmethod
+    def add(self, item: PendingStoreItem):
+        """Add a pending store item to the queue or other data structures."""
+        ...
+
+    @abstractmethod
+    def should_offload(self) -> bool:
+        """Determine whether the queue should be drained.
+
+        Returns:
+            True if offload should be triggered.
+        """
+        ...
+
+    @abstractmethod
+    def select_items(self, count: int) -> Iterator[PendingStoreItem]:
+        """Select which items to offload from the queue.
+
+        Args:
+            count: The number of items to select.
+
+        Returns:
+            A Iterator of PendingStoreItem.
+        """
+        ...
+
+
+class FIFOOffloadPolicy(OffloadPolicy):
+    """
+    FIFO offload policy: triggers when pending count reaches threshold,
+    and returns a fixed batch_size number of items from the front.
+    """
+
+    def __init__(self, configs: dict | None = None):
+        """
+        Args:
+            configs: The configuration for the FIFO offload policy.
+        """
+        self.pending_items: list[PendingStoreItem] = []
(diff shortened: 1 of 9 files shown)


lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+
+    request_id: str
+    metadatas: list[tuple["LMCacheMPRequestMetadata", dict[int, bytes]]] = field(
+        default_factory=list
+    )
+    is_finished: bool = False
+
+
+# TODO: support more offload policies
```
maobaolong: It would be nice to find the author of the TODO if you add your id to the TODO.

chunxiaozheng: @maobaolong Thanks for your suggestion, I have updated.

lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,213 @@
...
+
+# First Party
+from lmcache.utils import init_logger as lmcache_init_logger
+
+if TYPE_CHECKING:
+    # Third Party
+    from vllm.v1.core.block_pool import BlockPool
+
+    # First Party
+    from lmcache.integration.vllm.lmcache_mp_connector import LMCacheMPRequestMetadata
```
ApostaC: Should it be imported from `lmcache.integration.vllm.lmcache_mp_metadata`?

lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Abstract base class for lazy offload policies.
+
+    Subclasses define when to trigger offload (should_offload) and
+    which items to return (select_items).
+    """
+
+    @abstractmethod
+    def add(self, meta: "LMCacheMPRequestMetadata", block_hashes: dict[int, bytes]):
+        """Add a pending store item to the pending store."""
```
ApostaC: nit: let's have the high-level description of the args in the docstring.

lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,213 @@
...
+        ...
+
+    @abstractmethod
+    def should_offload(self) -> bool:
+        """Determine whether the queue should be drained.
+
+        Returns:
+            True if offload should be triggered.
+        """
+        ...
```
ApostaC: I'm wondering about the difference between `should_offload` and `select_items`.

Is my understanding correct that `select_items` will always return a non-empty list if we check `should_offload == True`?

Additionally, can we only have `select_items`, and when it returns an empty list, we will know we should not do offload.

lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,213 @@
...
+
+    request_id: str
+    metadatas: list[tuple["LMCacheMPRequestMetadata", dict[int, bytes]]] = field(
+        default_factory=list
+    )
+    is_finished: bool = False
+
+
+# TODO(chunxiaozheng): support more offload policies
+class OffloadPolicy(ABC):
```
ApostaC: I'm wondering whether this class is called in the scheduler or the worker (I assume it's in the scheduler). Would be helpful if we clarify it in the class docstring.

lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1670,6 +1703,129 @@ def get_finished(
 
         return ret_stores, finished_retrieves
 
+    @_lmcache_nvtx_annotate
+    def get_finished_with_lazy_offload(
```
ApostaC, line 1709: Are there any duplicated codes with the existing `get_finished`? (Especially the logic to handle the finished retrieves).

If so, we can consider to break it down into multiple functions, like
- get_finished_retrieves
- get_finished_stores
- get_finished_stores_with_lazy_offload

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -1053,7 +1126,54 @@ def _process_cached_requests(
...
+    def _process_lazy_offload_store_requests(
+        self, metadata: LMCacheMPConnectorMetadata
+    ):
+        if not self.lazy_offload or not self._pending_store.should_offload():
+            return
+
+        if not self._gpu_block_pool:
+            raise ValueError("Lazy offload is enabled but no GPU block pool is bound")
+
+        for item in self._pending_store.select_items():
```
ApostaC: Let's do a quick offline review of this part of the logic.

ApostaC (commented):
LGTM in general! Please see the comment

ApostaC:
Additionally, do we plan to add end-to-end CI for this?

lmcache/integration/vllm/lazy_offload_pending_store.py
chunxiaozheng: @ApostaC Thanks for your correction!

lmcache/integration/vllm/lazy_offload_pending_store.py
chunxiaozheng: done

lmcache/integration/vllm/lazy_offload_pending_store.py
chunxiaozheng: Good idea.

lmcache/integration/vllm/lazy_offload_pending_store.py
chunxiaozheng: done.

lmcache/integration/vllm/vllm_multi_process_adapter.py
chunxiaozheng: As discussed offline, I left a `TODO` here, we will remove or optimize here in later PR.

chunxiaozheng:
> Additionally, do we plan to add end-to-end CI for this?

Yeah, we will add end-to-end test later.

chunxiaozheng:
@ApostaC I have updated, PTAL.

ApostaC (approved):
LGTM!

lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+
+    request_id: str
+    metadatas: list[tuple["LMCacheMPRequestMetadata", dict[int, bytes]]] = field(
+        default_factory=list
+    )
+    is_finished: bool = False
+
+
+# TODO(chunxiaozheng): support more offload policies
```
maobaolong: Just a suggestion, you can separate this class into a single file under a single folder, let's say it `offload_policy/base.py`, and move the `FIFOOffloadPolicy` implementation to a file named `offload_policy/fifo.py` , how about your idea?

maobaolong: Then, new policies can be placed to this folder and take no affect to the existing files.

lmcache/integration/vllm/lazy_offload_pending_store.py
```diff
@@ -0,0 +1,215 @@
...
+        configs: dict | None = None,
+    ):
+        """
+        Initialize the pending store queue.
+
+        Args:
+            configs: The configuration for the pending store.
+        """
+        policy = (
+            configs.get("lmcache.mp.lazy_offload_policy", "FIFO") if configs else "FIFO"
```
maobaolong: This can be improved to be easy to extend, just a heads-up, do not block this PR merge.

lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -864,7 +907,21 @@ def update_connector_output(self, connector_output: KVConnectorOutput):
...
+        if not isinstance(meta, LMCacheMPWorkerMetadata):
+            return
+        for req_id, count in meta.completed_store_requests.items():
+            if self.scheduler_adapter.update_pending_store_count(req_id, count):
+                gpu_block_ids = self._pending_store.get_request_gpu_block_ids(req_id)
+                self._gpu_block_pool.free_blocks(
+                    [self._gpu_block_pool.blocks[bid] for bid in gpu_block_ids]
+                )
+                self._pending_store.remove_request_gpu_block_ids(req_id)
+                self.scheduler_adapter.end_session(req_id)
```
maobaolong, line 924: This is a important method. If it isn't called, the sessions will leak. In the regular flow, it should be called, but under exceptional conditions—specifically if a block hash mismatch is detected—it can still end up reaching this point.

maobaolong: After a deep discussion, i found there are two times session create, first time is when lookup called, second time when store. 

So this is not a bug.

lmcache/integration/vllm/lazy_offload_pending_store.py
chunxiaozheng: @maobaolong Thanks for your suggestion, I have updated, LGTM.

maobaolong (approved):
LGTM. Thanks for this great feature.

PR #4437, opened 2026-08-06
[Core] Optimize store/retrieve for linear models

ApostaC:

**Summary**

- Treat the linear models as sliding window models with sliding window size = engine block size
- Skip the null blocks during store
- Only retrieve the KV cache (call H2D) within the "sliding window" for linear layers

**Note:**

Linear models are required to use `--separate-object-groups` from now on.

**If applicable**:


lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -141,39 +143,38 @@ def _has_preemption_reqs(scheduler_output: SchedulerOutput) -> bool:
...
             f"Mamba-hybrid models with LMCache require "
-            f"block_size <= max_num_batched_tokens < 2 * block_size so every "
-            f"prefill step advances exactly one block and every block boundary "
-            f"gets a state snapshot; got max_num_batched_tokens={max_batched}, "
-            f"block_size={block_size}. Set --max-num-batched-tokens "
-            f"{block_size}."
+            f"max_num_batched_tokens >= block_size so every prefill step "
+            f"advances at least one full block; got "
+            f"max_num_batched_tokens={max_batched}, block_size={block_size}. "
+            f"Set --max-num-batched-tokens to at least {block_size}."
```
KuntaiDu, line 177: Check the case where LMCache block size is NOT a multiple of vLLM block size

KuntaiDu (approved):
Otherwise LGTM

ApostaC:
Tested with Gemma and DSV4, everything should work

deng451e (approved):
LGTM!

PR #4438, opened 2026-08-06
[Coordinator][CacheBlend] serve fleet blend lookup from the key directory

ruizhang0101:

Fleet-wide CacheBlend reuse currently runs on its own fingerprint directory
(`blend_directory.py`), fed by a publish RPC that every blend store makes to
`POST /blend/fingerprints`. This replaces it with a **content index derived from
the key directory's token bindings**, so fleet blend rides the cache-event
stream MP servers already emit.

Dropping the second source of truth is what buys the rest:

| | standalone fingerprint table | content index |
| --- | --- | --- |
| fed by | its own publish RPC per blend store | the one cache-event stream |
| match | 64-bit hash, trusted | hash to discover, **tokens to verify** |
| collision | wasted prefetch | candidate skipped |
| eviction | lazy tombstone, stale entries tolerated | exact, with the binding |
| placements | unknown; prefetch is blind | known (peer L1 vs shared L2) |

Main changes:

- **`token_offset` on token bindings.** Chunk hashes are prefix-chained, so the
  coordinator cannot derive where a chunk sat in its sequence, and re-RoPE needs
  it. `mp.tokens` already computed the offsets; they were being discarded.
- **Bindings hold tokens as a read-only `uint32` array** instead of
  `tuple[int, ...]`: ~1 KB per 256-token chunk against ~10 KB of boxed ints, and
  content comparison against a query window stays vectorized.
- **New `ContentIndex`** answering the fragment lookup, served by
  `POST /directory/blend-lookup`.
- **MP server becomes query-only** — the per-store fingerprint publish and the
  eviction publish are deleted, since blend stores already reach the directory
  through `LMCacheDrivenTransfer.store`.
- **Removes** `blend_directory.py`, `blend_directory_api.py`, the `/blend/*`
  endpoints and their wire schemas.


lmcache/cli/commands/coordinator.py
ApostaC, line 1: user facing doc for this

ApostaC (approved):
Otherwise LGTM

deng451e (approved):
LGTM

PR #4442, opened 2026-08-06
[CI][MP] Add DeepSeek-V4-Flash L1 correctness test (4-GPU)

BoJiang03:
## What this PR does / why we need it:

Adds a `dsv4_flash_tp` multiprocess CI test (4-GPU group) covering LMCache's
hybrid-KV-group slot-compression path against the real DeepSeek-V4-Flash model
(sparse-MLA + fp8, tokens_per_block 256/64/8/4).
Flow: an L1-enabled LMCache MP server (`--l1-size-gb 40`, chunk 256) + TP=4 vLLM
(`fp8_ds_mla`) serves one long deterministic request cold, resets vLLM's local
prefix cache via `/reset_prefix_cache` (no second 160GB weight load needed since
DSV4 has no recurrent state), re-serves the identical request from the L1 CPU
pool, and asserts byte-identical greedy outputs plus a non-zero retrieve count
(so the comparison cannot pass vacuously via vLLM's own prefix cache).
The prompt deliberately ends mid-sentence inside the 80x-repeated paragraph:
an open-ended ending leaves the first generated token at a measured top-2
logprob gap of 0.0000, where the retrieve path's prompt-tail recompute jitter
flips the greedy argmax (~2/9 fresh starts). The mid-sentence ending pins all
128 generated tokens (measured min gap 6.75).

## Special notes for your reviewers:

Validated locally on 4x H200: 5/5 fresh-startup cycles pass (plus 42 warm
store/retrieve cycles); each TP rank retrieves 8960 tokens (35 chunks) from L1.
`--enforce-eager` is used because FULL decode CUDA-graph capture crashes in
custom all-reduce at TP=4 on the pinned nightly (same as kimi_linear_tp);
correctness tests don't need CUDA graphs. The only untested-in-CI variable is
the first cold download of the 160GB weights, covered by the 90-min timeout.

### DIFF
--- a/.buildkite/k3_tests/multiprocess/scripts/run-dsv4-flash-tp.sh
+++ b/.buildkite/k3_tests/multiprocess/scripts/run-dsv4-flash-tp.sh
@@ -40,14 +40,16 @@
 #   Delete this workaround once vLLM's pin carries SM120 again; the guard
 #   below then no-ops on its own, since the bundled DeepGEMM will work.
 #
-#   Disabling DeepGEMM instead (VLLM_USE_DEEP_GEMM=0) does not work on SM120:
-#   only the fp8 linear and MXFP4 MoE call sites are gated by it. Measured on
-#   an SM120 node, that path falls to CutlassFp8BlockScaledMMKernel + MARLIN
-#   and then dies in mhc_pre_broadcast_tilelang, which calls DeepGEMM
-#   unconditionally; patching that one exposes the next wall
-#   (torch.ops._C.cutlass_scaled_mm has no SM120 block-scaled kernel in the
-#   wheel), and behind it DSv4's per-layer _o_proj -> deep_gemm_fp8_o_proj
-#   has no non-DeepGEMM implementation at all.
+#   Disabling DeepGEMM instead (VLLM_USE_DEEP_GEMM=0) does not work on SM120.
+#   Only the fp8 linear and FP4 MoE call sites consult is_deep_gemm_supported();
+#   measured on an SM120 node, those do demote (CutlassFp8BlockScaledMMKernel,
+#   MARLIN) and the run then dies in mhc_pre_broadcast_tilelang, which calls
+#   DeepGEMM unconditionally. Patching that gate exposes the next wall --
+#   torch.ops._C.cutlass_scaled_mm has no SM120 block-scaled kernel in the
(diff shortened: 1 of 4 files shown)


hlin99:
Thanks for the contribution!
Right, the deepseek v4 KV path is actually missing in CI, and it's the right thing to do. some comments:
1. TP4 actually consumes large part of our CI resource. if it's online for each approved PR, I would imaging the slow down of CI somehow. So, is there any way to run it in TP2 by any quantization?
2. I guess getting the resource, setting up the environment for this CI already takes amount of time, in this case, shall we test something more to make the waiting time be paid off, like 100 lm eval items

The most concern is on the resource, I tried to launch TP4 case before, and I found it really took long time to get 4 GPU from a single node. Just FYI my experience.

cc @ApostaC

hlin99:
@BoJiang03 please fix the DCO issue also. thx

.buildkite/k3_tests/multiprocess/scripts/run-dsv4-flash-tp.sh
```diff
@@ -0,0 +1,399 @@
...
+# custom_all_reduce.cuh) on current nightlies; a correctness test needs no
+# CUDA graphs, and eager also skips several minutes of capture/compile.
+VLLM_SERVER_DEV_MODE=1 vllm serve "$MODEL" \
+    --tensor-parallel-size "$TENSOR_PARALLEL_SIZE" \
+    --enable-expert-parallel \
+    --kv-cache-dtype fp8_ds_mla \
+    --tokenizer-mode deepseek_v4 \
+    --trust-remote-code \
+    --enforce-eager \
+    --enable-prefix-caching \
```
hlin99, line 334: to test LMCache, no prefix caching. otherwise may cache in gpu memory

BoJiang03: There is a reset_vllm_prefix_cache function call later. To test that STORE works correctly, the prefix cache should be active. You can also refer to .buildkite/k3_tests/multiprocess/scripts/run-kimi-linear-tp.sh, which also enables prefix caching

hlin99: okay, seems this way works. no prefix caching might be more straightforward. reset prefix cache is not needed anymore

hlin99 (approved):
LGTM

ApostaC (approved):

PR #4446, opened 2026-08-07
ci: add Intel XPU wheel build and release workflow

VincyZhang:

Add an Intel XPU wheel release path, mirroring the existing ROCm/CUDA variant flow.

Main changes:
- Add reusable XPU artifact workflow to build wheel in a pinned XPU runtime image.
- Add XPU wheel build script with
- Integrate XPU into publish pipeline
- Add Intel XPU wheel installation instructions.

This is needed to provide a prebuilt XPU wheel distribution path (without publishing variant wheels to PyPI), reducing from-source build burden for Intel XPU users.

- Validated locally using:
  - `vllm/vllm-openai-xpu:v0.26.0`
- Detected/pinned from validation:
  - Python 3.12.3
  - torch `2.12.0+xpu`
  - oneAPI compiler 2025.3.3
- End-to-end local build succeeded and produced:
  - `lmcache-0.0.0.dev0-cp312-cp312-manylinux_2_35_x86_64.whl`

**Install after `xpu` wheel release**:

```bash
VERSION=0.5.3  # replace with target release
pip install lmcache==${VERSION} --no-deps \
  --find-links https://github.com/LMCache/LMCache/releases/expanded_assets/v${VERSION}-xpu
```

**If applicable**:


.github/workflows/build_xpu_artifacts.yml
```diff
@@ -0,0 +1,121 @@
...
+                    set -euxo pipefail
+                    WHEEL=$(ls /work/LMCache/dist/*.whl | head -n 1)
+                    python3 -m pip install --no-deps "$WHEEL"
+                    python3 - <<"PY"
+                      import lmcache
+                      import torch
+                      print("LMCache import OK")
+                      print("Torch:", torch.__version__)
+                      PY
+                  '
```
ApostaC: From my claude reviews: this may have an indentation issue. We can put everything in the same line separated by `;` rather than writting them in multiple lines.

.github/workflows/build_xpu_artifacts.yml
```diff
@@ -0,0 +1,121 @@
...
+# its C++ ABI matches the target Intel vLLM XPU runtime. auditwheel excludes
+# torch and oneAPI/SYCL runtime libraries, so the wheel binds to the host torch
+# and userspace drivers at runtime -- like the ROCm/CUDA variant flows.
+#
+# Runtime install guidance (manual and scripted validation):
+# - launch runtime containers with --shm-size=4g to avoid /dev/shm starvation.
+# - install LMCache wheel with --no-deps to preserve pinned torch/oneAPI stack.
+#
+# Uploaded as the `release-xpu-artifacts` GHA artifact. The XPU wheel carries a
+# +xpu local version and is published to a dedicated v{tag}-xpu GitHub Release
```
ApostaC, line 15: Can we double-check where we enforce this `+xpu local version` in the code? I didn't see it.

ApostaC (commented):
One high-level comment: we should have a new tab in the user-facing documentation talking about how to install xpu wheels. It's in `installation.rst` under `docs/` folder.

Please see other detailed comments below.

.github/workflows/build_xpu_artifacts.yml
VincyZhang: Good catch, it is not enforced today. `SETUPTOOLS_SCM_PRETEND_VERSION` is set from the base release version, so the XPU wheel currently follows the same versioning model as the ROCm wheel, e.g. `0.5.3rc4` rather than `0.5.3rc4+xpu`. I will remove the inaccurate `+xpu` local version comment and add the XPU installation documentation.
XPU wheel release is similar to ROCm here: https://github.com/LMCache/LMCache/releases?page=1#release-v0.5.3rc4-rocm

VincyZhang:
The XPU wheel is built when qualifying pushes to `dev`, `release-*`, version tags, and published GitHub Releases.

- **Temporary location:** GitHub Actions artifact storage, default retained to 90 days
- **Release location:** For a published upstream GitHub Release, the wheel is downloaded from the workflow artifact and uploaded to a dedicated GitHub Release named `v<version>-xpu`.  GitHub Release assets do not expire automatically; they remain available until the corresponding release is deleted.

.github/workflows/build_xpu_artifacts.yml
VincyZhang: Thanks for pointing out. Resolved.

andyluo7:
Heads up on wheel naming — I hit this on the ROCm side.
  
  XPU_VERSION is just ${BASE} with no local version, and the repair uses --plat manylinux_2_35_x86_64, which is the same tag the ROCm wheel gets. So once both ship:
  lmcache-0.5.3-cp312-cp312-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl   (CUDA, PyPI)
  lmcache-0.5.3-cp312-cp312-manylinux_2_35_x86_64.whl                        (ROCm)
  lmcache-0.5.3-cp312-cp312-manylinux_2_35_x86_64.whl                        (XPU)
  Last two are the same filename. Different binaries, nothing to tell them apart, and pip show won't tell you which one you've got either.
  The convention's already in this PR — TORCH_XPU_SPEC is torch==2.12.0+xpu. Same thing on the wheel gives 0.5.3+xpu. lmcache==0.5.3 still matches it since == ignores the local segment, so nothing existing breaks.
  I've got #4481 open doing this for ROCm with +rocm7.2.
  Minor: if you add it, the install snippets in publish.yml and installation.rst should pin ${VERSION}+xpu instead of plain ${VERSION}.

ApostaC:
@VincyZhang @andyluo7 So what's the conclusion on the naming? Will there need to be any changes to this PR (as well as potentially changing the AMD artifact build pipeline) to avoid the naming conflict?

andyluo7:
Conclusion from my side: yes, changes needed — and it's broader than these two PRs.

Every non-PyPI variant we publish is currently indistinguishable from the default build. This already ships today:

```
nightly (cu13.0)   lmcache-0.5.4.dev32-cp310-cp310-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
nightly-cu129      lmcache-0.5.4.dev32-cp310-cp310-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
```

Same filename, different CUDA builds. Ditto `v0.5.4rc2-cu129` versus the `0.5.4rc2` wheel on PyPI.

What I'd suggest, matching what torch itself does:

- the PyPI wheel stays bare — PyPI rejects local versions, so it has to
- anything published to a GitHub release carries a local version naming the variant: `+cu129`, `+rocm7.2`, `+xpu`

Then `pip show` tells you which build you have, and `lmcache==0.5.3+rocm7.2` is expressible. A bare `lmcache==0.5.3` still resolves any of them since `==` ignores the local segment, so nothing existing breaks.

Your two questions:

- **#4446** — one line, set the wheel version to `${BASE}+xpu`. `TORCH_XPU_SPEC` is already `torch==2.12.0+xpu`, so the convention's in the PR.
- **AMD pipeline** — already done in #4481, approved, just needs merging. Nothing further.

The cu129 and nightly cases are pre-existing and independent of both PRs. Happy to take those as a follow-up if you want the convention applied everywhere.

VincyZhang:
LMCache XPU wheel release in Github (not Pypi) follows the same pattern as ROCm, and is implemented in this PR, please review. @ApostaC @andyluo7 
For  LMCache XPU nightly build/verify/publish, and corresponding vLLM nightly version update can be discussed in this PR: https://github.com/LMCache/LMCache/pull/4616

andyluo7:
Checked it over — this does what was needed. The smoke-check that installs the wheel into
the runtime image and imports `lmcache` is a good addition; the ROCm flow has no equivalent
and probably should.

Two things, neither blocking.

The version fallback doesn't fire:

```bash
BASE=$(git tag --list 'v[0-9]*' | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -1 | sed 's/^v//' || echo "0.0.0.dev")
```

Exit status comes from `sed`, not `grep`, so on a checkout with no matching tags `BASE` ends
up empty and the version becomes `+xpu`, which isn't valid PEP 440. My fault originally —
it's copied from the ROCm script. #4482 fixes it there with `BASE="${BASE:-0.0.0.dev}"`;
the same one-liner works here.

Smaller one: the auditwheel repair auto-skips when `patchelf` is missing and copies the raw
wheel instead. `patchelf` is pip-installed a few lines above so it shouldn't trigger, but it
means a publish could quietly emit a `linux_x86_64`-tagged wheel rather than
`manylinux_2_35_x86_64` if that install ever changes. Worth failing instead of skipping
unless `SKIP_AUDITWHEEL_REPAIR` was set explicitly.

Rest looks right — `xpu_version` wired through as an output so `publish.yml` doesn't repeat
the literal, `secrets: inherit` present, exclude list covers the oneAPI/SYCL surface, docs
pin `${VERSION}+xpu`.

ApostaC:
@VincyZhang Hey Vincy, is this PR ready for review again? Feel free to let us know!

VincyZhang:
@ApostaC It's ready for review again.

ApostaC (approved):
LGTM!

hlin99 (approved):
LGTM

PR #4447, opened 2026-08-07
[Misc] Update the default value of shm name and supported transfer mode to avoid shm warnings

ApostaC:

Recently, we've seen quite a few partners and users complaining about/being confused by the following logs:

```
LMCache WARNING: Insufficient /dev/shm capacity: need 2575906635776 bytes, have 1622248488960 bytes. Disabling SHM, falling back to pickle.
```

This PR fixes this by updating the default values of the CLI.

**If applicable**:


lmcache/v1/multiprocess/config.py
```diff
@@ -340,7 +343,7 @@ def add_mp_server_args(
     mp_group.add_argument(
         "--shm-name",
         type=str,
-        default=None,
+        default="",
```
hlin99, line 347: hi @ApostaC  "" will make pickle as the selected path and ignore shm path, unless the --shm-name explicitly set by cmdline. is this the purpose of the PR?

ApostaC: Yes. I will update the documentation in this PR as well

hlin99: Sure. As this is a cmdline breaking change, the associated CI (severals in buildkite multiproces) might need to be updated as well.

hlin99:
Hi @ApostaC 

I double-checked the current logic and confirmed that any device emitting this warning is non-CUDA: under `auto` mode, CUDA always selects the LMCache-driven path and such warning won't pop up.

> **Compatibility consideration:** For non-CUDA devices, changing the default to the LMCache-driven path may prevent them from working with the same command line(vs current pickle fallback). At present, only CUDA, CPU, and MUSA support both execution paths.

If the confusing warning is the main issue being addressed here, would it be possible to make the log more actionable instead? For example, the message could explain:

- which mode was selected;
- why that mode was selected for the current device; and
- which setting users can use to explicitly override the selection.

### To facilitate your decision making, just a quick recap of the `auto` design goal

The original purpose of `auto` was to hide LMCache- versus Engine-driven implementation details and preserve a consistent command line across multi-platforms, without requiring users to change their command-line configuration for different hardware environments. This is especially valuable for large CSP deployments, where the same configuration may be used across different hardware environments.

- **Short term**
  - **CUDA:** use the LMCache-driven path.
  - **All other devices:** use the Engine-driven path.

- **Longer term**
  - Make this selection fully transparent through default allowlists for LMCache-driven and Engine-driven modes, or through preferences reported by the platform layer.

ApostaC:
@hlin99 Let me clarify.

The biggest problem is that currently the default behavior is to allocate a lot of `/dev/shm` no matter the environment. And this hidden side-effect is not expected and not well documented either.

We have found some users with CUDA environments complain that LMCache takes too much of `/dev/shm`, which impacts their internal environments. And this complaint actually confuses our forward deployment engineers a lot because they don't have any idea why LMCache is requiring a large shm.

Another example is when we are doing benchmarks on the InferenceX platform (which is almost 100% maintained by coding agents), this behavior actually confuses the Claude code. Claude code limits the `--l1-size` to the `/dev/shm` size configured on the machine and then leads to very bad end-to-end performance. It also takes a long time for us to debug.


Since, from what we have seen, Nvidia/Amd is still the major platform people use, let's simplify the default behavior without requiring a large `/dev/shm`. For other use cases, people should be very sure about what they are configuring and what will be the side effect.

hlin99:
> @hlin99 Let me clarify.
> 
> The biggest problem is that currently the default behavior is to allocate a lot of `/dev/shm` no matter the environment. And this hidden side-effect is not expected and not well documented either.
> 
> We have found some users with CUDA environments complain that LMCache takes too much of `/dev/shm`, which impacts their internal environments. And this complaint actually confuses our forward deployment engineers a lot because they don't have any idea why LMCache is requiring a large shm.
> 
> Another example is when we are doing benchmarks on the InferenceX platform (which is almost 100% maintained by coding agents), this behavior actually confuses the Claude code. Claude code limits the `--l1-size` to the `/dev/shm` size configured on the machine and then leads to very bad end-to-end performance. It also takes a long time for us to debug.
> 
> Since, from what we have seen, Nvidia/Amd is still the major platform people use, let's simplify the default behavior without requiring a large `/dev/shm`. For other use cases, people should be very sure about what they are configuring and what will be the side effect.

Okay. Now I see. thanks @ApostaC for your explanation. make sense and agree.

ApostaC:
@hlin99 Thanks a lot for your understanding 🙏

sammshen (approved):
LGTM!

KuntaiDu (approved):
lgtm!

PR #4451, opened 2026-08-07
ci(cpu): enable MLA for DeepSeek-V2-Lite CPU e2e

maobaolong:
## Summary

vLLM merged a CPU MLA backend (vllm-project/vllm#49453), so DeepSeek-V2/V3 can now route through MLA on CPU instead of falling back to MHA.

This flips the `deepseek-v2-lite` CPU e2e profile from `VLLM_MLA_DISABLE=1` to `0` so we can validate LMCache's KV format / transfer path against a real MLA attention backend on CPU.

## Changes

- `.github/workflows/cpu_device.yml`: set `mla_disable: "0"` for the `deepseek-v2-lite` matrix profile (was `"1"`).
- `.github/scripts/run-cpu-e2e-validation.sh`: update the now-stale comment that claimed the CPU backend raised for MLA.

LMCache already supports MLA end-to-end (`mla_enabled`, `validate_mla_config`, and the `NL_X_NB_BS_HS` KV-format detection in the vLLM connector), so no code changes are needed — this is purely a CI toggle to exercise that path.

## Test plan

CI runs the `deepseek-v2-lite` leg on `ubuntu-22.04` and `macos-latest` with MLA enabled; if the e2e passes we confirm DeepSeek runs through MLA on CPU with LMCache.


ApostaC (approved):

chunxiaozheng (approved):
lgtm

Issue #4445, opened 2026-08-07
[Refactor][Platform]: Rename native_storage_ops → lmache_native and bind device-independent KV-format types (EngineKVFormat) into it


maobaolong:
## Summary

This proposal extends and operationalizes #3674 ("Move GPUKVFormat into a header file not coupled to accelerator") by going one step further: **rename the portable native extension `native_storage_ops` → `lmache_native`, and bind the device-independent KV-format data structures (`EngineKVFormat`, the classification predicates, and the transfer types) into that device-independent module.** The Python fallback `torch_ops.py` would then only need to implement the *device-specific* ops in pure Python, and can consume `lmache_native.EngineKVFormat` directly instead of re-defining a duplicate `IntEnum`.

This is a natural follow-up to the `DeviceOps` abstraction introduced in #3936.

## Background — current state

- `csrc/engine_kv_format.h` already exists as a **vendor-header-free** header (its own comment: *"Vendor-header-free, so every backend and the Python facade (lmc_ops) share one definition"*). It defines the `EngineKVFormat` enum plus classification predicates (`is_cross_layer`, `is_kv_list`, `is_layer_list`, `is_mla`, shape helpers).
- `csrc/kv_transfer_types.h` holds the device-independent transfer types.
- **But** `EngineKVFormat` is currently bound into the **accelerator-coupled** module, not the portable one. In `csrc/pybind.cpp`:
  ```cpp
  PYBIND11_MODULE(c_ops, m) {            // ← device/backend-specific (CUDA/HPU/SYCL)
      py::enum_<EngineKVFormat>(m, "EngineKVFormat") ...
      m.attr("GPUKVFormat") = m.attr("EngineKVFormat");  // back-compat alias
  }
  ```
  So `EngineKVFormat` is only importable when the accelerator `c_ops` extension is built/imported.
- To compensate, `lmcache/v1/platform/ops_types.py` **duplicates** the type in Python:
  ```python
  class EngineKVFormat(IntEnum): ...     # re-defined so call sites work without the native module
  GPUKVFormat = EngineKVFormat
  ```
  It is intentionally kept `IntEnum`-based so it compares equal to the pybind11 enum member with the same integer value.
- `lmcache/v1/platform/torch_ops.py` is the pure-Python op fallback. `DeviceOps` (`lmcache/v1/platform/base/device_ops.py`) delegates to it, and accelerator subclasses bulk-bind native kernels via `bind_native`.
[...]

PR #4453, opened 2026-08-07
refactor: rename native_storage_ops to lmcache_native and relocate KV-format/transfer types

maobaolong:
## Summary

This PR implements issue #4445: renaming `native_storage_ops` to `lmcache_native` and relocating the device-independent KV-format / transfer types into the new module so they live in a single, backend-agnostic place.

### Changes
- Rename the compiled extension module `native_storage_ops` → `lmcache_native` (build spec + pybind module name).
- Move `EngineKVFormat` / `GPUKVFormat` / `TransferDirection` and the format predicates (`is_kv_list`, `is_layer_list`, `is_cross_layer`, `is_mla`) into `lmcache_native`. The C++ `engine_kv_format.h` / `kv_transfer_types.h` are now included from `csrc/storage_manager/pybind.cpp`.
- Add a top-level `lmcache_native` shim and keep a `native_storage_ops` shim for backward compatibility.
- Forward the relocated names from `lmcache.c_ops` (via `lmcache.v1.platform.ops_types`).
- `lmcache/v1/platform/ops_types.py` now exposes the pure-Python `EngineKVFormat` / `TransferDirection` enums plus the `is_*` predicates (with identical semantics to the C++ predicates), so consumers get a working fallback even when the native extension is not built (CPU-only / test envs).

### Fixes for CI
- **CPU wheel build**: use **absolute** `include_dirs` when building the common C++ extensions, so `engine_kv_format.h` resolves regardless of the build working directory used by `python -m build` (previously failed under Ninja's temp build dir).
- **Import / test failures**: make the top-level `lmcache_native` shim tolerate a missing native extension instead of failing the whole import, and forward `lmcache.c_ops` relocated names through `ops_types` (always importable, pure-Python fallback).
- Fixed the `native_storage_ops.pyi` stub (dropped it in favor of the Python shim) and corrected the `lmcache_native.pyi` enum stubs so mypy passes.

## Test plan
- [x] `pre-commit run --all-files` passes locally (ruff/ruff-format/isort/codespell/clang-format/mypy).


lmcache/__init__.py
```diff
@@ -27,21 +27,50 @@
...
+    "EngineKVFormat",
+    "GPUKVFormat",
+    "TransferDirection",
+    "is_cross_layer",
+    "is_kv_list",
+    "is_layer_list",
+    "is_mla",
+)
+
+
```
ApostaC, line 44: Why do we have this? Is this because we don't want to keep the compatibility of the reference to `lmcacache.c_ops.EngineKVFormat` (and other symbols)?

ApostaC (commented):
One quick question. Otherwise LGTM! 
This refactor makes a lot of sense.

lmcache/__init__.py
maobaolong: This shim preserves backward compatibility for lmcache.c_ops.EngineKVFormat / TransferDirection / is_* so existing call sites and downstream code don't break after the relocation.

ApostaC (approved):

chunxiaozheng (approved):
lgtm

PR #4433, opened 2026-08-05
[MP][MUSA] Add SGLang MUSA connectors and multiprocess MHA transfer

superleo:
**What this PR does / why we need it**:
based on: https://github.com/LMCache/LMCache/pull/4454
Enable LMCache on MUSA for SGLang, covering both in-process connectors and multiprocess handle-mode transfer for SGLang's split K/V MHA layout.
- Add `sglang_musa_connectors` for in-process MHA/MLA (non-layerwise) and layerwise MHA store/load via TorchMUSA ops.
- Extend MUSA `DeviceOps` / cache context for SGLang multiprocess MHA (`TWO_X_NL_X_NB_BS_NH_HS`): packed pointer operands, nested K-then-V reconstruction, and staging copy helpers.
- Add unit tests for connectors, pointer reconstruction, block transfer, and adapter routing.

**Special notes for your reviewers**:
- test passed with Musa 4.3.5. SGlang0.5.13 on 8*S5000 machine.
- Multiprocess MLA and layerwise MLA remain out of scope (fail closed).

**If applicable**:
- [x] this PR contains user facing changes - docs added
- [x] this PR contains unit tests

PR #4454, opened 2026-08-07
feat(sglang/mp): decouple SGLang MP transfer

superleo:
- need by: #4433 
Decouple the SGLang multi-process KV transfer path from a hardcoded CUDA backend so any platform that implements the device spec can participate, and fix a latent IPC-event lifetime bug in the async store/retrieve futures.

Platform/device abstraction (no MUSA-specific code):
- multi_process_adapter: replace hardcoded CudaIPCWrapper with get_device_spec() + wrap_one_kv_cache(); gate registration on is_handle_transfer_available() so a missing capability fails fast instead of crashing deep in a backend. migrate to_cuda_future -> to_device_future.
- sglang_adapter: add _model_uses_mla() and forward the actual MLA cache-row width (kv_head_dim) so MLA engines select the right path.

Fixes:
- futures: add MessagingFuture.retain_reference() so async callers keep an exported IPC event alive for the future's lifetime; replaces the private `_export_event` attribute hack on the returned future.
- base/cache_context: fix off-by-one where the per-layer format list was sized against len(kv_caches_) instead of num_layers_.

**If applicable**:


maobaolong (approved):
@superleo This PR looks take minor changes for platform and lmcache layer. It looks good for me about these parts. And it should be take a double check for the integration layer codes by other reviewers who is more familiar with `SGLang`.

tests/v1/test_sglang_mp_adapter.py
Shaoting-Feng: We can remove this now since import sglang is moved to TYPE_CHECKING.

Shaoting-Feng (approved):
LGTM

tests/v1/test_sglang_mp_adapter.py
superleo: Good point, removed pytest.importorskip("sglang") and its comment.

superleo:
hi @Shaoting-Feng , the current failed k3-mp-test case seems an env issue: "Free memory on device cuda:0 (65.48/94.97 GiB) is less than desired GPU memory utilization (0.8, 75.97 GiB)"

PR #4457, opened 2026-08-08
test(platform): cover the cross-process event ordering contract

andyluo7:
## Why

The `lmcache_driven` path's correctness rests on one invariant that nothing currently checks: the event returned by `import_event` must represent the **producer's** pending work.

The producer records an event on the stream that wrote the KV blocks and ships only the handle, without synchronizing (`worker_transfer.py`, `submit_store` / `submit_retrieve`). The consumer imports it and calls `wait_event` before reading those blocks (`lmcache_driven_transfer.py`). Nothing else orders the two processes.

If a backend returns an already-completed event, that `wait_event` becomes a no-op and the consumer can read KV blocks the producer has not finished writing. It doesn't crash or hang — it yields subtly wrong cache contents under load, so it can pass tests and benchmarks and still ship.

Existing coverage cannot catch this, by construction. `tests/v1/platform/test_event_ipc.py` validates delegation against an injected fake module, and that fake's `query()` returns `True` unconditionally — the fake itself models the failing behavior. There is currently no on-hardware test of event IPC at all. The design doc's review checklist lists "tests cover import/export and stream ordering"; import/export is covered, ordering was not.

This gets more valuable as the platform surface grows. CUDA, MUSA, XPU, HPU and CPU are in tree today and more device backends are landing. Every one of them implements `import_event`, and right now a new vendor has no executable definition of what "correct" means here — only prose.

## What

- `tests/v1/platform/test_event_ipc_ordering.py` — asserts the contract directly instead of racing for it: while the producer's write is in flight the imported event must report incomplete, and waiting on it must make the write visible.
- `docs/design/v1/platform/base/event_ipc_abstraction.md` — adds an explicit **Ordering contract** section. The doc described the mechanism but never stated the rule, including that host-side synchronization in the importing process is not a substitute, since it drains only that process's own streams.

The assertion fires on `query_event` while a ~1s GPU delay is still pending, so the margin is roughly three orders of magnitude. It is not a timing race and should not flake.

## Testing

| | MI300X (gfx942) | MI355X (gfx950) |
|---|---|---|
| this branch | PASS | PASS |
| with #4422 applied | FAIL | FAIL |

Run in `vllm/vllm-openai-rocm:v0.26.0` (torch 2.11.0+gitd0c8b1f, ROCm 7.2.3). Skips cleanly where event IPC is unavailable (`1 skipped in 0.02s`). No new dependencies, single GPU, ~5s.

This came up while reading #4422, but the test stands on its own regardless of how that one lands.


andyluo7:
Pushed a fix — CI caught a real problem with the original assertion, thanks to whoever's runner that is.

I had asserted that `query_event()` on the imported event reports incomplete while the producer's work is still pending. That holds on ROCm but not on CUDA: an imported IPC event reports complete in the importing process regardless of the producer's progress, since the recorded state is per-process bookkeeping. Cross-process ordering is carried by `wait_event` alone. I'd only validated on AMD, so the test was over-specified and correctly failed on the NVIDIA runner.

Narrowed it to the property that actually is the contract — after `wait_event`, the producer's write must be visible — and widened the producer's delay for margin. The failure mode is now one-sided: if the consumer is ever starved long enough to miss the window the test passes rather than failing, so it won't flake red.

Re-validated:

| | on `dev` | with #4422 applied |
|---|---|---|
| NVIDIA B300 | pass | n/a (fallback is ROCm-gated) |
| MI300X (gfx942) | pass | fail |
| MI355X (gfx950) | pass | fail |

Also added the missing `Signed-off-by` for DCO.

ApostaC (approved):
LGTM in general. Not sure whether this is part of #4551 so @royyhuang please take a look.

docs/design/v1/platform/base/event_ipc_abstraction.md
```diff
@@ -60,6 +60,26 @@ class EventIPCBackend(Protocol):
...
+
+`import_event` must return an event that represents the *producer's* work, on
+the producer's stream, in the producer's process. The imported event is the
+only thing ordering the two processes: the producer records it and ships the
+handle without synchronizing, and the consumer's `wait_event` before touching
+shared KV-cache memory is what keeps it from reading blocks the producer has
+not finished writing.
+
+A backend must therefore not substitute a locally created or already-completed
+event for a real imported one. Doing so makes the consumer's `wait_event` a
```
ApostaC: What do you mean by: "A backend must therefore not substitute a locally created or already-completed event for a real imported one."?

tests/v1/platform/test_event_ipc_ordering.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    device = buf.device
+    torch.cuda.set_device(device)
+    backend = get_event_ipc_backend(device)
+    stream = torch.cuda.Stream(device=device)
+    event = backend.create_event(device)
+    with torch.cuda.stream(stream):
+        torch.cuda._sleep(SLEEP_CYCLES)
+        buf.fill_(MAGIC)
+        backend.record_event(event, stream)
```
royyhuang, line 92: Since producer is a freshly spawned process, CUDA might lazily loads the kernel for `fill_()` while the device is idle, hence blocked by the sleep call (tested locally that `buf.fill_()` is blocked by 2s exactly the sleep time) and making event recorded after the buffer is filled. 

This makes the stream executed in the following order: long sleep -> load and run fill_() kernel -> writing finished -> record event -> wait event with immediate return -> read data, not the asynchronous event waiting and ordering we are testing. So probably add a tiny warmup run for `fill_()` before the actual run (e.g., at the beginning of the producer function).

royyhuang (changes requested):
Thanks for your contribution! The testing coverage makes sense to me. Please see my comment below.

royyhuang (commented):
Another thing just comes in mind. Our unit test pipeline has `CUDA_LAUNCH_BLOCKING` enabled. So no matter what the kernels will be blocking and sequentially launched. And your test might not guard against any contract violation: everything will pass?. Is it possible we remove the `CUDA_LAUNCH_BLOCKING` flag @ApostaC? If that is not possible, we can also try `monkeypatch.delenv()`

andyluo7:
@royyhuang you were right, and it's worse than the test just being weak — it was passing on
CUDA for the wrong reason entirely. Fixed in 8d9be719.

Measured the lazy load in a faithful repro (fresh spawned child, IPC-shared buffer, 2s
delay), timing how long the host blocks inside `fill_()`:

```
B300    cold 1974.72 ms   warmed 0.02 ms
MI355X  cold    7.28 ms   warmed 0.05 ms
```

So on CUDA the export happened ~2s in, after the write had already landed, exactly as you
described. On ROCm the load is cheap enough that the window survived — which is why the
test caught the #4422 fallback there but never caught anything on NVIDIA.

Added the warmup you suggested. Checked it with a deliberately broken backend whose
`import_event` returns a locally recorded and synchronized event:

```
B300   without warmup -> PASS   (vacuous, missed it)
B300   with warmup    -> caught, 4194304/4194304 elements unwritten
gfx950 with warmup    -> caught, 4194304/4194304 elements unwritten
```

Unpatched `dev` still passes on both, and the #4422 fallback is still caught on gfx950.

On `CUDA_LAUNCH_BLOCKING` — good point, same failure shape. Rather than change the pipeline
I clear it for the child via `monkeypatch.delenv`; the producer is spawned, so it reads the
environment when it builds its own context and the parent's already-initialised context is
unaffected.

@ApostaC rewrote that paragraph to say what actually happens rather than state a rule —
which event gets returned, why `wait_event` then returns immediately, and that the failure
is silent. Pointed it at #4422 as the concrete case.

royyhuang (approved):
Thanks for addressing our comments. LGTM now.

PR #4460, opened 2026-08-08
Add batch layerwise for MP LMCache driven mode

zxue2:
MP LMCache driven mode currently with per chunk mode only.   Add per N-layer batch.   N is configurable for different granularity. For more design details, please refer https://github.com/zxue2/LMCache/blob/enable_kv_load_per_layer/docs/design/v1/multiprocess/lmcache_driven_layerwise_batch_design.md 

``` --layerwise-batch``` is parsed into MPServerConfig.layerwise_batch and stored on the server context which exposes ```layerwise_batch: int — layers per GPU batch (0 = disabled)``` and ```layerwise_loading: bool — derived as layerwise_batch > 0```.  The server advertises this to each worker at registration: the REGISTER_KV_CACHE MQ response class changed from None to int, returning layerwise_batch. The worker/adapter derive their own behavior from that value.

 To make per-layer H2D a single contiguous memcpy, store D2H writes the chunk in a per-layer interleaved layout ```[L, 2, T, D] ([K0,V0, K1,V1, …])``` instead of the default grouped ```[2, L, T, D] ([K0,K1,…, V0,V1,…])```. A new ```kv_interleaved``` selects the corresponding offset formula in the scatter/gather kernel. 

 Layerwise retrieve returns per-layer IPC event handles  instead of a single handle. The vLLM connector's ```wait_for_layer_load(layer_name)``` extracts the layer index and waits on that layer's event, so the worker blocks only until the layer it is about to compute is resident. Support per N-layer batch.

 Layout is fixed per deployment: a server is uniformly per-layer or uniformly per-chunk . Store and retrieve both read the same layerwise_loading. Mixed per-layer / per-chunk deployments sharing one persistent L2 need extra discriminator in metadata. But in practice there should be no need for mixed deployment.

e.g.,

Enable per-layer , batch 8-layer:

```lmcache server --layerwise-batch 8 --l1-size-gb 20 --eviction-policy LRU --port 5555 --max-workers 2```

```
(APIServer pid=584949) INFO 08-08 08:38:45 [loggers.py:310] Engine 000: Avg prompt throughput: 5.0 tokens/s, Avg generation throughput: 31.2 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 91.3%
```

```
======== Engine Benchmark Result (long-doc-qa) =========
-------------------- Configuration ---------------------
Engine URL:                        http://localhost:8000/
Model:      /media/model-space/Qwen/Qwen2.5-32B-Instruct
Workload:                                    long-doc-qa
----------------------- Results ------------------------
Successful requests:                                   4
Failed requests:                                       0
Benchmark duration (s):                            14.57
Total input tokens:                                40136
Total output tokens:                                 512
Input throughput (tok/s):                        2755.52
Output throughput (tok/s):                         35.15
----------------- Time to First Token ------------------
Mean TTFT (ms):                                  2409.84
P50 TTFT (ms):                                    686.85
P90 TTFT (ms):                                   5613.87
P99 TTFT (ms):                                   7513.34
-------------------- Decoding Speed --------------------
Mean decode (tok/s):                               19.38
P99 decode (tok/s):                                21.20
```

per-chunk by default:

```lmcache server  --l1-size-gb 20 --eviction-policy LRU --port 5555 --max-workers 2```

```
(APIServer pid=591144) INFO 08-08 08:51:01 [loggers.py:310] Engine 000: Avg prompt throughput: 35.6 tokens/s, Avg generation throughput: 38.9 tokens/s, Running: 1 reqs, Waiting: 0 reqs, GPU KV cache usage: 26.9%, Prefix cache hit rate: 0.0%, External prefix cache hit rate: 92.4%
```

```
======== Engine Benchmark Result (long-doc-qa) =========
-------------------- Configuration ---------------------
Engine URL:                        http://localhost:8000/
Model:      /media/model-space/Qwen/Qwen2.5-32B-Instruct
Workload:                                    long-doc-qa
----------------------- Results ------------------------
Successful requests:                                   4
Failed requests:                                       0
Benchmark duration (s):                            14.45
Total input tokens:                                40136
Total output tokens:                                 512
Input throughput (tok/s):                        2778.38
Output throughput (tok/s):                         35.44
----------------- Time to First Token ------------------
Mean TTFT (ms):                                  2344.03
P50 TTFT (ms):                                    688.06
P90 TTFT (ms):                                   5589.79
P99 TTFT (ms):                                   7455.34
-------------------- Decoding Speed --------------------
Mean decode (tok/s):                               19.34
P99 decode (tok/s):                                21.20
```

**If applicable**:

### CODE REVIEW

**Comment by zxue2** [comment]
@ApostaC  could you pls have a look?  thx

File: csrc/mp_mem_kernels.cu [hunk 1]
```diff
@@ -380,23 +387,25 @@ void multi_layer_block_kv_transfer_templated(
...
   const cudaStream_t stream = at::cuda::getCurrentCUDAStream();
 
+  // --- Upload LMCache object pointers to device ---
+  // Small allocation (N * 8 bytes); caching allocator makes this O(1).
+  // The copy runs on the current stream, so it completes before the kernel.
+  auto gpu_ptrs_tensor =
+      torch::empty({num_objects},
+                   torch::TensorOptions().dtype(torch::kInt64).device(device));
+  gpu_ptrs_tensor.copy_(
+      torch::from_blob(lmcache_objects_ptrs.data(), {num_objects},
```
ApostaC: The goal of `MemoryObj4` is to avoid having this. In some cases, this copy will fall back to a sync copy and force a stream synchronization.

zxue2: If ```lmcache_objects_ptrs.data()``` is non-pinned (pageable) memory, ```cudaMemcpyAsync``` falls back to a path with one extra host-side memcpy into a driver-managed pinned staging buffer before the DMA. This blocks the host thread  but does not synchronize the GPU stream . Is this the case you're referring to, or something else?
For the legacy per-chunk path with hardcoded max_batch_size = 4, this is a 32-byte staging copy .

The change is because the legacy temp buffer pool is repurposed for per-layer mode. In per-chunk, each temp buffer holds one full chunk (all layers), so num_objects ≤ 4. In per-layer mode, max_chunks_per_pass = max_batch_size × total_layers / N, which can exceed 4 (e.g., 64 layer for Qwen-32B with N=4). MemoryObj4's hardcoded objects[4] would cap this at 4, forcing 16× more kernel launches and defeating the merged N-layer path.

I can split into two kernel paths (MemoryObj4 for per-chunk, device ptr array for per-layer) if you prefer. The trade-off is some duplicated kernel/launch code to avoid a 32-byte pageable copy whose impact is likely very limited.  Could you pls share your preference? @ApostaC

zxue2: Split scatter kernel into per-chunk (MemoryObj4) and per-layer (dynamic ptr) paths(https://github.com/LMCache/LMCache/pull/4460/changes/8dd1fe1d572d7eb0f1012234cee98d2b87b11a95)

ApostaC (changes requested):
I went through the code again. I think at a highlevel, it could work.

However, to be 100% transparent and honest, I do have bias towards layerwise transfer -- I don't want to experience the painful maintenance of such code path again without having improvement on common use cases. 
I totally understand that layerwise is important for your use case, and I really appreciate your work on this. That's why I did not directly say no to this functionality.

That said, I suggest we have a totally separate code path that does not touch any existing logic. More specifically, we will have:
- a different `lmcache_mp_connector.py` and `vllm_multi_process_adapter.py` just for layerwise transfer. This can be loaded runtime with connector path configuration passed to vLLM.
- A different transfer module just for layerwise (so that we don't touch blend_v3.py and `lmcache_driven_transfer.py`
- A different protocol definition for the new transfer module
- A new c++ source code for the layer wise kernels.

Also, please make sure that the partial results handling are hidden in the result itself instead of introducing new thing in `mq.py`.

zxue2:
> I went through the code again. I think at a highlevel, it could work.
> 
> However, to be 100% transparent and honest, I do have bias towards layerwise transfer -- I don't want to experience the painful maintenance of such code path again without having improvement on common use cases. I totally understand that layerwise is important for your use case, and I really appreciate your work on this. That's why I did not directly say no to this functionality.
> 
> That said, I suggest we have a totally separate code path that does not touch any existing logic. More specifically, we will have:
> 
> * a different `lmcache_mp_connector.py` and `vllm_multi_process_adapter.py` just for layerwise transfer. This can be loaded runtime with connector path configuration passed to vLLM.
> * A different transfer module just for layerwise (so that we don't touch blend_v3.py and `lmcache_driven_transfer.py`
> * A different protocol definition for the new transfer module
> * A new c++ source code for the layer wise kernels.
> 
> Also, please make sure that the partial results handling are hidden in the result itself instead of introducing new thing in `mq.py`.

Sure, I'll manage to decouple the source codes of per-layer and per-chunk as much as possible. It seems per-layer and per-chunk are comparable per some experiments by lmcache bench with Qwen3-32B/TP2/CUDA/long-doc-qa/4-requests per-run.   Btw could you pls share your most concerned models?   thx @ApostaC

ApostaC:
@zxue2 One potential model to look at is deepseek v4 -- very popular model, and has small KV cache size as well.

zxue2:
> @zxue2 One potential model to look at is deepseek v4 -- very popular model, and has small KV cache size as well.

@ApostaC  Sorry for late update due to a break since last week.  https://github.com/LMCache/LMCache/pull/4816 is created to address the refactor requirements.  PR 4816's logic is functionally identical to PR 4460.  No change to legacy per-chunk/per-request path.  Please help check below points especially.

- lmcache_driven_transfer.py has a minor change set , i.e. , _transfer_object_group() hook to accommodate per-layer path  for code reuse.  
- A general streaming partial responses from a server to a client over a single request still needs to touch mq.py.   It fits the  specific per-layer use case.  

Will update validation results with Qwen and deepseek v4 later.

PR #3033, opened 2026-04-15
 [Observability] Per-request root OTel span and SpanRegistry for MP server tracing

deng451e:
Description:

Summary
 
  - Adds a root "request" OTel span per session wrapping all child spans (mp.lookup_prefetch, mp.retrieve, mp.store)  
  - Introduces 4 CPU-synchronous sentinels: MP_REQUEST_START, MP_STORE_SUBMITTED, MP_RETRIEVE_SUBMITTED, MP_SESSION_END; root span close deferred until
  in-flight GPU callbacks complete
  - Introduces SpanRegistry — shared span context dict enabling multi-subscriber, multi-level tracing (e.g. request → mp.retrieve → l2.disk_load) without
  coupling to MPServerTracingSubscriber; new subscribers plug in with two lines in config.py
- Adds a Docker Compose example for viewing waterfall traces (OTel Collector → Tempo → Grafana) with a pre-provisioned dashboard
 - Design doc covering all request trace scenarios and a worked example for registering sub-spans
  
Screenshot 2026-04-14 at 4 55 14 PM

---

> [!NOTE]
> **Medium Risk**
> Changes MP server event emission and tracing lifecycle handling around async GPU operations; incorrect ordering/counter logic could lead to leaked spans or misleading trace boundaries, though the change is well-covered by new tests.
> 
> **Overview**
> Adds a per-request root OTel span (`"request"`) for MP server tracing and ensures `mp.lookup_prefetch`, `mp.retrieve`, and `mp.store` spans are properly parented under it, avoiding orphaned spans in trace backends.
> 
> Introduces four new CPU-synchronous lifecycle events (`MP_REQUEST_START`, `MP_STORE_SUBMITTED`, `MP_RETRIEVE_SUBMITTED`, `MP_SESSION_END`) and updates the MP server to emit them so the tracing subscriber can defer closing the root span until any in-flight GPU store/retrieve callbacks complete (while stamping the root end time at `MP_SESSION_END`).
> 
> Adds a shared `SpanRegistry` for cross-subscriber context propagation, wires it into observability config, expands tracing tests to cover deferral/race scenarios, and includes a new `examples/observability` Docker Compose stack (Collector/Tempo/Prometheus/Grafana) plus a design doc to demonstrate and validate the new trace hierarchy.
> 
[...]

PR #4461, opened 2026-08-08
[MP][Observability] Decompose transfer timing: gather/DMA phases, reserve time

zhengfeihe:

Storing the KV cache from the serving engine to CPU memory, and loading it back from CPU memory to GPU memory, are among the most performance-critical paths in LMCache. Today, there is no easy way to measure them.

## Scope

This PR covers the LMCache-driven path only. Engine-driven measurement is out of scope.

## Why not Nsight?

`nsys` gives a much more detailed picture, but running it across LMCache releases × models × hardware is impractical.

It belongs at the end of the loop, as the tool a developer reaches for to confirm a specific optimization—not as the thing you run to notice a regression.

We need always-on, low-overhead observability for both:

* the DMA staging section
* the gather/scatter kernel section

## Design

This PR introduces an RAII `PhaseTimer::Section`.

The constructor creates a CUDA event pair and records the start event on the transfer stream. The destructor records the end event, so scope exit closes the interval.

Impact on the existing transfer paths is kept minimal:

* With timing disabled, or for a zero-byte section, the `Section` stays inert: no events and no queue traffic.
* If event creation fails, it degrades to inert. The transfer itself is unaffected.
* Completed records are pushed in batches under a single lock into a bounded queue, with the oldest entries evicted when full.

Python drains the queue from the transfer-completion path and publishes the samples onto the event bus.

Draining is decoupled from any single request: it collects sections that completed earlier, so no request waits on the timing path.

## Example

I made a local Grafana example with running data

Screenshot From 2026-08-18 02-04-57

**If applicable**:


zhengfeihe:
Sorry... Please hold off on reviewing the PR for now. I rechecked it today and found a few areas that could still be refined, so I've relabeled it as a draft PR. :bow:

csrc/cuda/mp_mem_kernels.cu
```diff
@@ -489,66 +494,107 @@ void execute_object_group_transfer(
...
+      const int64_t block_bytes = static_cast<int64_t>(desc.kv_size) * desc.nl *
+                                  desc.bs * desc.nh * desc.hs *
+                                  desc.element_size;
+      const int64_t moved_blocks = static_cast<int64_t>(launch.total_blocks) -
+                                   launch.skip_prefix_n_blocks;
+      if (moved_blocks > 0) {
+        kernel_bytes += block_bytes * moved_blocks;
+      }
+    }
+
```
zhengfeihe, line 552: This computes the exact byte count for the kernel section. The extra arithmetic is a few integer ops per launch, so the cost is negligible; the alternative (reusing the staged payload) is simpler but overestimates kernel bytes,  noticeably so for sliding-window / hybrid models where skip_prefix_n_blocks is common. 🤔

zhengfeihe:
Hi~@royyhuang @ruizhang0101  I refactored the code and rewrote the PR description.
 
 I think it should be much cleaner than the first version. Could you help take a look when you get a chance? Thanks!

csrc/cuda/mp_mem_kernels.cu
```diff
@@ -489,66 +494,107 @@ void execute_object_group_transfer(
...
+      const PageBufferShapeDesc& desc =
+          kernel_group_specs[launch.group_idx].shape_desc;
+      const int64_t block_bytes = static_cast<int64_t>(desc.kv_size) * desc.nl *
+                                  desc.bs * desc.nh * desc.hs *
+                                  desc.element_size;
+      const int64_t moved_blocks = static_cast<int64_t>(launch.total_blocks) -
+                                   launch.skip_prefix_n_blocks;
+      if (moved_blocks > 0) {
+        kernel_bytes += block_bytes * moved_blocks;
+      }
```
ApostaC: nit: can we have a helper function to calculate the kernel bytes and step bytes so that the main function looks cleaner?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -60,6 +64,7 @@
 _HAS_NATIVE_OBJECT_GROUP_TRANSFER: bool = hasattr(
     device_ops, "execute_object_group_transfer"
 )
+_HAS_PHASE_TIMING_POP: bool = hasattr(device_ops, "pop_completed_phase_timings")
```
ApostaC: nit: maybe a more general flag name: `_HAS_PHASE_TIMING` -- it would be clearer when reading the code that uses this flag.

ApostaC (commented):
Otherwise LGTM. Please fix the conflicts as well, thanks!

csrc/cuda/mp_mem_kernels.cu
zhengfeihe: extracted, thanks for the advice~

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
zhengfeihe: called it `_HAS_TRANSFER_PHASE_TIMING` now :thinking:

deng451e:
This is very useful feature! I wonder if we can add these event sub-spans to the request’s root span, as visualized in #3033?

ApostaC (approved):
LGTM!

ApostaC:
Actually one more thing: please update the user-facing docs about the new observability metrics: https://docs.lmcache.ai/mp/observability/metrics.html

(in `docs/source`)

zhengfeihe:
Thanks for the review and advice!

I will update the user facing docs and try add these event sub-spans to the request’s root span! Will ping you again when I finished ~

lmcache/v1/mp_observability/subscribers/transfer_phase_sampler.py
```diff
@@ -0,0 +1,67 @@
...
+    def _on_transfer_end(self, event: Event) -> None:
+        """Pop every finished sample and publish them in one event.
+
+        No-op when the native op is unavailable or nothing has finished.
+        """
+        if not _HAS_TRANSFER_PHASE_TIMING:
+            return
+        samples = _device_ops.pop_completed_phase_timings()
+        if not samples:
+            return
```
andyluo7, line 61: When a transfer produces no samples, END marks its tracing entry complete but this return sends no signal that can retire it, so it remains until TTL. Please emit completion for every END and add an empty-sample cleanup test.

zhengfeihe: Sorry I missed this comment. 

Thanks for your review and catching this! Fixed. Every END now publishes a keyed samples event (empty pops included), which retires the transfer immediately

zhengfeihe:
Sample trace pictures:
Store
Screenshot From 2026-08-31 16-28-24
Retrieve 
Screenshot From 2026-08-31 16-27-00

these phase timings are stream intervals, so they include the wait for SMs (kernel) and for the copy engine (staging), not just pure execution time, waiting the earlier per-request numbers silently ignored, yet it is genuinely part of the transfer's critical path. 

I dropped the kernel throughput metric because calculating throughput as bytes divided by the entire interval does not accurately reflect the actual transfer speed when most of that interval is spent waiting for contended SM resources. The wait time is still visible in the tracing span.

deng451e (approved):
This is great contribution! LGTM.

PR #4466, opened 2026-08-09
[MP] Batch store transfers via contiguous non-skipped runs

zhengfeihe:

Split D2H batches at None entries (skipped chunks) instead of dropping any batch containing one, so store can use the full max_batch_size like retrieve already does. Store previously pinned batch_size=1 to survive skips one object at a time.

## Performance

Measured on DeepSeek-V2-Lite-Chat (MLA), 1x H100 80GB, TP=1, 128 concurrent
lanes, 15-minute production-trace replay via aiperf, vLLM 0.26.0. Identical
parameters across both runs; the only variable is this patch. Raw data is the
final /metrics scrape of each run; all numbers below derive from the
histogram `_sum` / `_count` pairs quoted here.

### Raw histogram data (store direction, device cuda:0 / d2h)

| prometheus histogram | before `_count` | before `_sum` | after `_count` | after `_sum` |
|---|---:|---:|---:|---:|
| `lmcache_mp_l0_l1_store_throughput_GB_per_second` | 6,213 | 83,626.50 | 6,288 | 169,535.60 |
| `lmcache_mp_transfer_kernel_throughput_GB_per_second{d2h}` | 43,669 | 6,175,433.41 | 11,688 | 7,394,114.53 |
| `lmcache_mp_transfer_staging_throughput_GB_per_second{d2h}` | 43,669 | 1,570,133.40 | 11,688 | 422,107.23 |
| `lmcache_mp_l0_l1_store_reserve_time_seconds` | 6,330 | 2.5799 | 6,369 | 1.3167 |
| `lmcache_mp_l0_l1_load_throughput_GB_per_second` (control) | 344 | 9,485.30 | 330 | 9,106.73 |

### Derived: mean = sum / count

| metric | before | after | delta |
|---|---:|---:|---:|
| store throughput (GB/s) | 13.46 | **26.96** | **+100.3%** |
| kernel D2H sections issued | 43,669 | 11,688 | **-73%** |
| kernel D2H per-section (GB/s) | 141.41 | 632.62 | +347% |
| staging DMA D2H (GB/s) | 35.96 | 36.11 | +0.4% |
| reserve_write mean (ms) | 0.408 | 0.207 | -49% |
| load throughput (GB/s, control) | 27.57 | 27.60 | +0.1% |

Before this patch batch_size is pinned to 1, so one kernel section = one
chunk: the run stored 43,669 chunks across 6,213 store requests (~7.0
chunks/request). After, the same workload (6,288 requests, +1.2%) needs
only 11,688 sections — an effective batch of ~3.7.

### Per-chunk time decomposition

One chunk = 27 layers x 576 head dim x 2 bytes x 1024 tokens = 31.85 MB
= 0.03185 GB. Component time per chunk = 0.03185 / mean GB/s; reserve per
chunk = reserve mean / (chunks per request); "other" is the remainder
against the composite.

| per chunk | before | after | derivation |
|---|---:|---:|---|
| total (composite) | 2.37 ms | 1.18 ms | 0.03185 / store GB/s |
| staging (DMA) | 0.89 ms (37%) | 0.88 ms (75%) | 0.03185 / staging GB/s |
| gather kernel | 0.23 ms (10%) | 0.05 ms (4%) | 0.03185 / kernel GB/s |
| reserve_write | 0.06 ms (2%) | 0.03 ms (2%) | reserve mean / 7.0 |
| other (metadata/IPC/queueing) | **1.20 ms (51%)** | **0.22 ms (19%)** | total - the abov

No kernel or C++ change is involved. Of the 1.19 ms saved per chunk, 0.98 ms
is per-operation overhead that disappears when ~3.7x fewer operations are
issued, and 0.17 ms is the kernel side effect of larger launches (launch cost
amortized, grid no longer starved at MLA's nh=1). DMA throughput is unchanged,
as expected: the same bytes cross PCIe in fewer, larger operations. After this
patch staging owns 75% of the store path, which is where follow-up work should
go. Retrieve already used the full batch size; the load row is a no-change
control.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
+++ b/lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
@@ -173,6 +173,41 @@ def all_null_chunk_masks(
     return masks
 
 
+def contiguous_runs(
+    memory_objs: Sequence["MemoryObj | None"],
+) -> list[tuple[int, list["MemoryObj"]]]:
+    """Split a batch into maximal contiguous runs of non-``None`` objects.
+
+    ``None`` entries (chunks the storage manager skipped) act as
+    delimiters and appear in no run, so every returned run can be
+    transferred as one kernel batch with consecutive object positions.
+
+    Args:
+        memory_objs: Batch entries in chunk order; ``None`` marks a
+            skipped chunk.
+
+    Returns:
+        ``(offset, objects)`` pairs in order, where ``offset`` is the
+        run's start position within *memory_objs* and ``objects`` is the
+        non-empty list of consecutive non-``None`` entries.
+    """
+    runs: list[tuple[int, list[MemoryObj]]] = []
+    current: list[MemoryObj] = []
+    current_start = 0
+    for idx, memory_obj in enumerate(memory_objs):
+        if memory_obj is None:
+            if current:
+                runs.append((current_start, current))
+                current = []
+            continue
+        if not current:
+            current_start = idx
+        current.append(memory_obj)
+    if current:
+        runs.append((current_start, current))
+    return runs
+
+
 def downsample_and_stage_block_ids(
     cache_context: BaseCacheContext,
     block_ids: list[list[int]],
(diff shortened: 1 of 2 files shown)


zhengfeihe:
Discussed offline, this shall be replaced by an alternative double buffer pr (coming soon)

PR #4467, opened 2026-08-09
[Bugfix][GPU Connector] Fix silent KV corruption for vLLM fused/packed (CS) layout in the in-process path

### LINKED ISSUE #4463: [Bug][GPU Connector] Silent KV cache corruption on vLLM 0.26 fused/packed KV layout
**Label**

`bug`

**Describe the bug**

On vLLM 0.26, the in-process connector (`LMCacheConnectorV1`) **silently returns wrong
KV data** for models whose KV cache uses vLLM's fused/packed ("unified") layout, where
K and V share the trailing per-head dimension. A request served from the LMCache CPU
tier produces different output than the same request computed from scratch.

There is no exception and no warning — generations are simply wrong.

Setting `use_gpu_connector_v3=True` is not a workaround; it fails differently, with a
hard crash on the first store:

```
File "lmcache/v1/cache_engine.py", line 558, in store
File "lmcache/v1/gpu_connector/gpu_connectors.py", line 619, in from_gpu
    memory_obj_tensor.copy_(tmp_gpu_buffer, non_blocking=True)
RuntimeError: The size of tensor a (1024) must match the size of tensor b (2048) at
              non-singleton dimension 3
```

So on this stack both in-process code paths are unusable. `LMCacheMPConnector` is
**not** affected and works as a workaround.

Severity: silent data corruption in the default configuration.

**To Reproduce**

Save as `repro_lmcache_fused_kv.py`. Requires only vLLM + LMCache — no dataset, no
server (except for the optional `lmcache-mp` control). Runs in ~2 minutes per mode and
exits non-zero when the bug is present.

vLLM's own prefix cache is disabled, so the second identical request **must** be served
by the offload tier. Under greedy decoding both generations must be byte-identical.

The script also **asserts that a transfer actually happened** before comparing outputs,
by hooking the exact call each connector makes to pull KV back from its CPU tier — so a
`MATCH: True` cannot be a silent no-op, and a `lmcache` failure cannot be "the cache was
skipped". It ships with three controls (`baseline`, `vllm-offload`, `lmcache-mp`) which
are what make a failure attributable.

```python
#!/usr/bin/env python3
"""Repro: LMCache in-process connector corrupts KV on vLLM's fused/packed KV layout.

  python repro_lmcache_fused_kv.py lmcache        # BUG: MATCH False
  python repro_lmcache_fused_kv.py baseline       # control: MATCH True
  python repro_lmcache_fused_kv.py vllm-offload   # control: MATCH True
  python repro_lmcache_fused_kv.py lmcache-mp     # control: MATCH True (needs MP server)

vLLM's own prefix cache is disabled, so the 2nd identical request cannot be served
by vLLM and must come from the offload tier. Greedy decoding => both generations
must be byte-identical.

The script also ASSERTS that a transfer really happened, so "MATCH: True" cannot be
a silent no-op (see check_offload_happened).
"""

import os
import sys

MODES = ("lmcache", "lmcache-mp", "vllm-offload", "baseline")
MODE = sys.argv[1] if len(sys.argv) > 1 else "lmcache"
if MODE not in MODES:
    sys.exit(f"unknown mode {MODE!r}; choose one of {MODES}")

os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"   # engine in-process
if MODE.startswith("lmcache"):
    os.environ["LMCACHE_CHUNK_SIZE"] = "256"
    os.environ["LMCACHE_LOCAL_CPU"] = "True"
    os.environ["LMCACHE_MAX_LOCAL_CPU_SIZE"] = "8"
    os.environ.setdefault("PYTHONHASHSEED", "0")     # required by the MP connector

from vllm import LLM, SamplingParams          # noqa: E402
from vllm.config import KVTransferConfig      # noqa: E402

ktc = None
if MODE == "lmcache":
    ktc = KVTransferConfig(kv_connector="LMCacheConnectorV1", kv_role="kv_both")
elif MODE == "lmcache-mp":
    ktc = KVTransferConfig(
        kv_connector="LMCacheMPConnector", kv_role="kv_both",
        kv_connector_extra_config={"lmcache.mp.host": "tcp://127.0.0.1",
                                   "lmcache.mp.port": 6555})
elif MODE == "vllm-offload":
    ktc = KVTransferConfig(
        kv_connector="OffloadingConnector", kv_role="kv_both",
        kv_connector_extra_config={"cpu_bytes_to_use": 8 << 30, "block_size": 256})

# ---------------------------------------------------------------------------
# Offload verification hooks.
#
# Each hook wraps the one call the connector makes to actually pull KV back from
# its CPU tier. Counting those calls proves the 2nd request was SERVED FROM CACHE
# rather than silently recomputed -- without this, a no-op connector would "pass".
# ---------------------------------------------------------------------------
LOADS: list[int] = []

if MODE == "lmcache":
    from lmcache.v1.gpu_connector import gpu_connectors as _gc

    for _cls in ("VLLMPagedMemGPUConnectorV2", "VLLMPagedMemGPUConnectorV3"):
        _k = getattr(_gc, _cls, None)
        if _k is None:
            continue

        def _wrap(klass):
            _orig = klass.to_gpu

            def _counting_to_gpu(self, memory_obj, start, end, **kw):
                LOADS.append(memory_obj.get_size())
                return _orig(self, memory_obj, start, end, **kw)

            klass.to_gpu = _counting_to_gpu

        _wrap(_k)

elif MODE == "lmcache-mp":
    from lmcache.integration.vllm import vllm_multi_process_adapter as _mpa

    _MPW = _mpa.LMCacheMPWorkerAdapter
    _orig_submit = _MPW.batched_submit_retrieve_requests

    def _counting_submit(self, request_ids, ops, event, **kw):
        LOADS.append(len(request_ids))
        return _orig_submit(self, request_ids, ops, event, **kw)

    _MPW.batched_submit_retrieve_requests = _counting_submit

elif MODE == "vllm-offload":
    from vllm.v1.kv_offload.cpu.manager import CPUOffloadingManager

    _orig_prepare_load = CPUOffloadingManager.prepare_load

    def _counting_prepare_load(self, keys, req_context):
        LOADS.append(len(list(keys)))
        return _orig_prepare_load(self, keys, req_context)

    CPUOffloadingManager.prepare_load = _counting_prepare_load


def check_offload_happened() -> str:
    """Prove request #2 was served from the offload tier rather than recomputed.

    Returns:
        A human-readable evidence string.

    Raises:
        SystemExit: If no KV was actually loaded from the offload tier.
    """
    if MODE == "baseline":
        return "n/a - baseline has no offload tier by design"
    if not LOADS:
        sys.exit(
            f"VERIFY FAILED [{MODE}]: the offload tier was never asked to load. "
            "vLLM served the request itself, so this run proves nothing. Check "
            "that enable_prefix_caching=False."
        )
    if MODE == "lmcache":
        return f"LMCache loaded {sum(LOADS) / 1e6:.1f} MB in {len(LOADS)} chunks"
    if MODE == "lmcache-mp":
        return f"LMCache MP issued {sum(LOADS)} retrieve request(s)"
    return f"vLLM loaded {sum(LOADS)} chunks from its CPU tier"


llm = LLM(model="Qwen/Qwen3-8B", kv_transfer_config=ktc, max_model_len=8192,
          gpu_memory_utilization=0.85,
          # MUST stay False. With prefix caching on, vLLM's GPU cache serves the
          # 2nd request itself, the offload tier is never consulted and the bug is
          # masked -- verified: "need to load: 0", MATCH True on an affected build.
          enable_prefix_caching=False)

# See "prompt sensitivity" note in the issue -- deliberately semi-repetitive.
prompt = ("Reference dossier 4815162342.\n\n"
          + "The quick brown fox jumps over the lazy dog near the riverbank. " * 260
          + "\n\nQuestion: Summarize the passage in one sentence.\nAnswer:")

sp = SamplingParams(max_tokens=32, temperature=0.0)          # greedy => deterministic
first = llm.generate([prompt], sp)[0].outputs[0].text        # full compute
LOADS.clear()  # only inspect request #2
second = llm.generate([prompt], sp)[0].outputs[0].text       # served from the KV cache

evidence = check_offload_happened()

print(f"\nmode: {MODE}")
print(f"offload verified: {evidence}")
print(f"1st (compute): {first!r}")
print(f"2nd (cached) : {second!r}")
print(f"MATCH: {first == second}")
os._exit(0 if first == second else 1)
```

Run:

```bash
python repro_lmcache_fused_kv.py lmcache        # exit 1  <-- the bug
python repro_lmcache_fused_kv.py baseline       # exit 0  (control)
python repro_lmcache_fused_kv.py vllm-offload   # exit 0  (control)
python repro_lmcache_fused_kv.py lmcache-mp     # exit 0  (control, needs MP server)
```

The `lmcache-mp` control needs a server:

```bash
python -m lmcache.v1.multiprocess.server --host 0.0.0.0 --port 6555 \
    --l1-size-gb 20 --eviction-policy LRU --max-workers 4 --chunk-size 256
```

All four modes on the affected build:

```
### lmcache       exit=1  offload verified: LMCache loaded 528.5 MB in 14 chunks   MATCH: False
### vllm-offload  exit=0  offload verified: vLLM loaded 14 chunks from its CPU tier MATCH: True
### lmcache-mp    exit=0  offload verified: LMCache MP issued 1 retrieve request(s) MATCH: True
### baseline      exit=0  offload verified: n/a - baseline has no offload tier      MATCH: True
```

**Two things that will stop you reproducing it**

1. **`enable_prefix_caching` must stay `False`.** With it on, vLLM's own GPU prefix
   cache serves the second request, the offload tier is never consulted, and the bug
   is masked (verified: `need to load: 0`, `MATCH: True` on an affected build). The
   bug does still occur with prefix caching on, but only once the GPU cache evicts,
   which needs many documents and a capped KV cache — a far more fragile repro.
2. **The prompt matters.** Too repetitive and the bug is masked (a false negative we
   hit); too random and greedy output is degenerate and flips on any numerical noise
   (`vllm-offload`, a known-correct connector, also reported `MATCH: False`). The
   prompt above was validated to discriminate across all four modes. **If a control
   fails, distrust the prompt, not the connector.**

Observed:

```
mode: lmcache
1st (compute): ' The passage consists of multiple repetitions of the sentence "The quick brown
                 fox jumps over the lazy dog near the riverbank." ...'
2nd (cached) : ' The passage is a series of the same sentence about a quick brown fox jumps
                 over the lazy dog near the riverbank." \n\nQuestion: What is the main idea'
MATCH: False
```

Deterministic — identical wrong output across runs. All three controls pass, so
decoding determinism and the harness itself are sound.

**Expected behavior**

`MATCH: True`. A cache hit must be numerically identical to a full recompute.

**Environment**

| | |
|---|---|
| LMCache | `dev` @ `8932ed07` |
| vLLM | 0.26.0 |
| Model | Qwen/Qwen3-8B (36 layers, 8 KV heads, head_dim 128, bf16) |
| Hardware | 1x H100 80GB, TP=1 |
| OS / Python | Linux, Python 3.12, torch 2.11.0+cu130 |
| Config | `chunk_size=256`, `local_cpu=True`, non-layerwise, `use_gpu_connector_v3=False` (default) |

thegoldenflow:
<html>
<body>
<html><head></head><body><p>Fixes #4463</p><h2>Root cause</h2><p>vLLM 0.26 uses a fused/packed KV layout (<code inline="">NL_X_NB_BS_NH_CS</code> / <code inline="">NL_X_NB_NH_BS_CS</code>) where each head stores <code inline="">[K | V]</code> in a content dimension <code inline="">CS = 2 * HS</code>.</p><p>The in-process transfer path had an implicit LMCache-side layout contract, which caused three related failures:</p><ol><li><p><strong>V2 silent corruption.</strong><br><code inline="">VLLMPagedMemGPUConnectorV2</code> uses true split-KV <code inline="">KV_2LTD</code> memory objects (<code inline="">[2, L, T, NH*HS]</code>), while <code inline="">multi_layer_kv_transfer</code> treated fused buffers as packed (<code inline="">[1, L, T, NH*CS]</code>). The engine slot stride was therefore derived from half the real packed row width, causing slot aliasing and preventing the V plane from round-tripping correctly.</p></li><li><p><strong>Fused HND <code inline="">head_size</code> mismatch.</strong><br>The HND fused branch interpreted <code inline="">head_size</code> as logical <code inline="">HS</code> and multiplied it by two, while callers already pass the format spec's packed <code inline="">CS = 2*HS</code>. This doubled the packed per-head stride again.</p></li><li><p><strong>V3 first-store shape mismatch.</strong><br><code inline="">cache_engine.store()</code> allocates memory objects before V3's lazy layout discovery registers the KV-layer-groups manager. The first allocation therefore used <code inline="">[2, L, T, NH*HS]</code>, while the later group-derived temporary buffer used <code inline="">[1, L, T, NH*CS]</code>, producing the reported first-store <code inline="">copy_</code> shape mismatch.</p></li></ol><p>The MP connector is unaffected because its fused page geometry is already explicit (<code inline="">kv_size=1</code>, <code inline="">hs=CS</code>) and it copies packed rows directly.</p><h2>Fix</h2><p>Make the LMCache-side layout an explicit transfer contract instead of inferring it from caller identity or tensor shape:</p><ul><li><p>Add <code inline="">MemObjKVLayout</code> with:</p><ul><li><p><code inline="">SPLIT_KV_2LTD</code>: <code inline="">[2, L, T, NH*HS]</code></p></li><li><p><code inline="">FUSED_PACKED</code>: <code inline="">[1, L, T, NH*CS]</code></p></li></ul></li><li><p>Require fused transfer callers to declare the layout explicitly; shapes validate the declared contract rather than define it.</p></li><li><p>Add split-KV addressing for fused CS/TWO_HS, NHD/HND formats, mapping each head's K and V planes onto the leading/trailing halves of the engine's packed <code inline="">CS</code>.</p></li><li><p>Standardize fused <code inline="">head_size</code> to mean the format spec's packed per-head width (<code inline="">CS</code>); fix the HND addressing and update CacheBlend to pass <code inline="">shape_desc.hs</code>.</p></li><li><p>Select vectorized transfer width from the per-head run rather than the full row, so transfers never cross fused head boundaries.</p></li><li><p>Make <code inline="">metadata.get_shapes()</code> return logical split-KV shapes for fused in-process kernel groups. V2 and V3 therefore use the same persistent <code inline="">KV_2LTD</code> representation, while CacheBlend/MP staging remains packed.</p></li><li><p>Update the torch fallback with the same split/packed semantics and validation. SYCL receives signature parity only; fused XPU transfer remains out of scope.</p></li></ul><p>For packed vLLM rows, K occupies the leading <code inline="">HS</code> and V the trailing <code inline="">HS</code>. This matches vLLM 0.26's packed cache split in <code inline="">vllm/v1/attention/backends/flash_attn.py</code> and LMCache's existing fused K-only re-RoPE behavior.</p><h2>Tests</h2><p>Added <code inline="">tests/v1/test_fused_kv_transfer.py</code> covering:</p><ul><li><p>Independent D2H and H2D semantic oracles against a pure-PyTorch reference, rather than relying only on round-trips.</p></li><li><p>CS and deprecated TWO_HS spellings, both NHD and HND.</p></li><li><p>Split and packed LMCache-side layouts.</p></li><li><p>Shuffled slot mappings, partial transfers, <code inline="">skip_prefix_n_tokens</code>, bf16/fp16, and vectorization edge cases.</p></li><li><p>Contract validation for missing/unknown layouts and shape/layout mismatches.</p></li></ul><p>Added <code inline="">tests/v1/test_metadata_shapes.py</code> to verify that fused V3 shapes remain <code inline="">[2, L, T, NH*HS]</code> before and after layout discovery, while MLA remains single-plane.</p><p>Existing GPU-connector, torch-ops, multiprocess, cache-engine, and memory-management suites show no new regressions.</p><h3>End-to-end repro</h3><p>The original CUDA/H100 repro is still pending on this branch; the development machine used for the fix has no CUDA GPU.</p>
mode | before | after
-- | -- | --
LMCache V2 | MATCH: False | pending
LMCache V3 | first-store crash | pending
baseline / vLLM offload / LMCache MP | pass | pending (expect unchanged)

<h2>Backward compatibility</h2><ul><li><p>Existing fused in-process cache entries should be flushed: V2 entries were written with the corrupted addressing, and pre-fix V3 fused paths did not produce reliable compatible entries.</p></li><li><p>CacheBlend and MP keep their packed staging representation.</p></li><li><p>Non-fused formats keep their existing behavior; the new layout parameter defaults to <code inline="">UNSPECIFIED</code>.</p></li><li><p>Deprecated TWO_HS spellings share the corrected fused kernel paths but are no longer produced by current layout detection.</p></li></ul><h2>Out of scope</h2><ul><li><p>XPU fused transfer semantics.</p></li><li><p>HND end-to-end validation; HND is covered by the kernel/reference tests.</p></li></ul></body></html>
</body>
</html>


Vivo50E:
We hit this bug in vLLM-Omni and reproduced it end to end on an A100. This PR fixes it. The symptom on `dev` has also changed since this was opened.

Interest declared: vLLM-Omni's KV offload (vllm-project/vllm-omni#1330) is blocked on this. Every released LMCache silently corrupts KV on vLLM 0.26+, so there's no version we can ship against.

vLLM 0.27.0, `FLASH_ATTN`, A100, TP=1, Qwen2.5-Omni-3B thinker (36 layers, `num_kv_heads=2`, `head_size=128`, dense). The backend registers `(97871, 2, 16, 256)` stride `(8192, 256, 512, 1)`; `attempt_permute_to_contiguous_view` sorts that to `(97871, 16, 2, 256)` and the rank-4 branch reads it as `NL_X_NB_BS_NH_CS`.

Worth flagging separately: at `num_kv_heads == 2` the two readings are shape-identical. `[NB, BS, KV=2, NH*HS=256]` and `[NB, BS, NH=2, CS=2*HS=256]` give the same 4-tuple, block size and head size — nothing about the tensor distinguishes them. At any other head count a `num_kv_heads` hint would catch it.

Measured with two probes in the vLLM-Omni runner reading the engine's own `self.kv_caches` (same `data_ptr`/shape/stride as what LMCache gets): one fingerprints the KV behind a request's first N tokens, the other diffs per-block sums so a restore names the blocks it wrote to. One engine answers the same prompt twice, cold then served, so the hit is the only variable; the no-LMCache control passes.

The round-2 request holds blocks 37–68 (slots 592–1103):

| LMCache | blocks written | `[0,512)` non-zero | text |
|---|---|---|---|
| 0.5.4 (PyPI) | 18–34 | 0 / 9,437,184 | diverges |
| `dev` @ `11b389a9` | 37–68 | 4,718,563 / 9,437,184 (exactly half) | diverges |
| `dev` @ `7cc0a2cc` (today) | not re-probed | not re-probed | still diverges |
| this PR (`74c0a121`) | 37–68 | 9,437,124 / 9,437,184 | identical |

All of them report `Retrieved 512 out of 512 required tokens ... 0.0176 gb`, which is the right byte count. Nothing raises.

On 0.5.4 the per-token stride is 256 elements where the layout needs 512, so slot 592 lands at block 18.5 instead of 37 and slot 1104 at 34.5 instead of 69 — 17 blocks, and the probe reports exactly `changed_blocks=17 first=18 last=34`. The blocks the model reads stay at literal zero; the ones written belonged to another request.

At `11b389a9` the destination is right but only half the elements come back, matching the V-plane root cause here. On `dev` at `7cc0a2cc` today it still fails, and the cold round is wrong too — three requests share a prefix, so requests 2 and 3 are already served from LMCache in round 1 and come back mangled. I didn't re-probe that one, so I can't say which root cause is live, only that `dev` doesn't have it fixed.

With this PR the restored prefix is bit-identical to the cold computation (`sum=31047.3946`, `nonzero=9437124/9437184`, both planes full), the text matches word for word, and the test passes.

The branch is 117 commits behind `dev` and the symptom has changed underneath it, so a rebase would show which root causes are still live. Happy to re-run all of this on one.

thegoldenflow:
@Vivo50E, Thanks for the detailed A100/vLLM-Omni reproduction. This is very helpful and confirms that the issue is still present on the current dev branch.

I’ll rebase #4467 onto the latest dev, rerun the targeted tests, and update this thread with the post-rebase results. I’ll keep 74c0a121 as the pre-rebase comparison point.

andyluo7 (commented):
The layout contract and reference tests look sound. Please run the AMD lane on this head; it changes the HIP-built transfer kernel, and there is no AMD result attached yet.

thegoldenflow:
@andyluo7 Thanks — I understand that adding the `amd` label admits the AMD lane. I don't have AMD hardware locally, so I'll rely on that lane for HIP/ROCm validation.

At the moment, no AMD status has appeared for head `0eaaf28`; only `k3-unit-tests` and `unit-tests-xpu` have reported. Could you check whether the AMD build is queued, or re-trigger it if needed? I'll keep this head frozen (no force-pushes) until all relevant lanes have reported.

Two required checks are also still missing on this head: `k3-multiprocess-test` is gated on the `mp`/`full` label, and `k3-integration-test` (which should run on every PR) hasn't reported. Could you or another maintainer add `mp` and re-trigger the integration lane, so all relevant results land on the same SHA?

Vivo50E:
@thegoldenflow -- this branch now conflicts with dev. Could you rebase and push?

@andyluo7 happy to help get this one over the line if there is anything useful I can take on.

PR #4474, opened 2026-08-10
[MP][Feat][good-first-issue] Add Phoenix L2 Adapter — GPU-Direct Storage Backend for LMCache MP Mode

MonthFall:
### 1. LMCache Disk L2 Adapter: Current State and Motivation

In LMCache's MP (multi-process) mode, the existing disk-based L2 adapters (e.g., `fs_native`, `fs`) retrieve KV cache data via **POSIX `read()`**, which forces data through host CPU memory. The retrieve path is `NVMe →(POSIX read)→ L1 CPU MemoryObj →(H2D scatter via CUDA IPC)→ vLLM GPU paged KV cache` — **two PCIe traversals** where CPU DRAM acts as an intermediary. 
Overly long access paths limit the performance of LMcache when using SSDs as L2 storage.

Although the legacy **In Process** mode includes a GDS backend implementation, however, MP mode has **no GPU-direct storage L2 adapter** currently — all disk-based adapters go through the CPU. We need a new GPU-direct storage L2 adapter that truly eliminates the CPU from the critical retrieve path, enabling NVMe-to-GPU DMA directly.

### 2. The Phoenix Project

[**Phoenix**](https://github.com/xPU-IO/phoenix) (published at SC '25) is a refactored I/O stack for GPU Direct Storage that offers extremely low software overhead. It is developed by [**xPU.IO**](https://github.com/xPU-IO), an open-source community co-initiated by **Tencent DCPD** (Tencent Data Computing Platform Department) and **NICELab** (Shanghai Jiaotong University and Xiamen University) , in collaboration with leading industry partners such as **MetaX**, focusing on xPU I/O for AI. The project aims to provide high-performance, low-overhead, and vendor-neutral GPU-direct storage I/O for AI training and inference workloads.

#### 2.1 Phoenix vs. NVIDIA GDS

Phoenix targets the same storage→GPU direct-I/O problem as NVIDIA GPUDirect Storage (cuFile), but it using CPU MMU to do the GPU virtual address to phycial address translation. It is a fit when you value control, portability, and deployment breadth over binding to a proprietary stack.

| Advantage | What it means in practice |
| --- | --- |
| **Fully open source, top to bottom** | The entire path — `phoenixfs` kernel module and `libphoenix` library— is open and auditable. When something breaks you can trace and fix it yourself, down to the kernel P2P mapping, instead of filing a ticket against a closed `libcufile.so`. Essential for self-controlled / compliance-audited environments. |
| **Vendor-agnostic by construction** | The DMA backend is selected at build time (CUDA / HIP / NPU) behind a `connector` function-pointer table; core code carries no vendor headers. The same upper-layer and adapter code runs on NVIDIA today, and on AMD / Huawei NPU once their backends land. cuFile is NVIDIA-only. | 
| **No cuFile dependency, no filesystem gate** | Phoenix works at the VFS / block layer and only needs a valid `fd`. Local ext4 / xfs or parallel filesystems (NFS/BeeGFS / Lustre) are transparent — no cuFile support-list, no `nvidia-fs`, no GDS-specific driver requirement. `fd` points somewhere, the DMA reads from there.  
| **`io_uring`-accelerated NVMe** | Pheonix removed the `phony buffer` used the cufile, so the GPU Pheonix removed the `phony buffer` used the cufile, so GPU IO can be accelerated by mature kernel I/O stack, such as iouring or libaio.  Thus batch / async engine (`phxfs_*_batch`, `phxfs_batch_submit_*`) can drives parallel NVMe arrays at full bandwidth with minimal syscall and queueing overhead.  In real-world testing on a bare interface, a 4x PCIe 4.0 NVMe drive delivered read performance of 24 GB/s and write performance of 15 GB/s under 512 concurrent requests.。
| **Environmental Compatibility** | Phoenix offers better compatibility. Older versions of network passthrough solutions, such as GPU Direct RDMA—which are commonly used in multi-node deployment environments—rely on `nvidia-peermem` for address registration. This might creates some conflicts with exsiting Phoenix mechnism. However, Phoenix provides a GPU staging buffer mode，which only register a 256M buffer in GPU address space，therefore ensuring compatibility with other older kernels.

In addition, we have also taken note of other [GPU I/O acceacceleration soleration solutioutions](https://github.com/ScaleX-IO/LMCache/commits/ugds-backend/)  that utilize an SPDK-like approach to fully utilize the bandwidth of NVMe devices in user mode. However, these solutions face compatibility challenges when integrated with various file systems in production environments.

#### 2.2 What Phoenix Brings to LMCache

By integrating Phoenix as an L2 adapter, LMCache gains 2 key advantages:

**1. Direct NVMe-to-GPU DMA, bypassing CPU DRAM entirely:**

```
Current:  NVMe →(POSIX read)→ CPU MemoryObj →(H2D scatter)→ GPU paged KV    [2× PCIe]
Phoenix:  NVMe →(phxfs DMA)→ GPU MemoryObj →(D2D scatter)→ GPU paged KV     [1× PCIe + 1× GPU-internal]
```

This halves PCIe traffic on the retrieve path, frees CPU memory bandwidth for other workloads, and reduces retrieve latency — especially impactful for long-context workloads where KV cache chunks are large and frequently evicted to disk.

**2. Reduced L1 cache dependency — cheaper to operate:**

Unlike `fs_native` and other POSIX-based adapters that heavily rely on a large L1 CPU DRAM cache to absorb retrieve traffic, Phoenix's DMA path is fast enough that **LMCache can operate with a dramatically smaller L1 cache while maintaining near-peak performance**. In our benchmarks (see Section 4), Phoenix with only **61 GB L1** achieves TTFT comparable to what `fs_native` needs **500 GB+ L1** to reach. This means deployments can save hundreds of GB of pinned CPU DRAM — a significant cost reduction, especially for multi-instance serving where CPU memory is a shared, contested resource.

#### 3.1 Architecture Compatibility

LMCache's MP (multi-process) mode follows an **L1-centric + multi-L2-adapter-parallel** architecture:

- **L1** = CPU pinned DRAM (hot cache, managed by `L1Manager`)
- **L2 Adapters** = pluggable backends (fs, S3, Redis, etc.), each running a background worker thread with event-fd-driven store/lookup/load queues
- **PrefetchController** orchestrates L2→L1 data flow: LOOKUP phase (broadcast to all adapters) → LOAD phase (policy-based routing, data fills L1 CPU MemoryObjs)
- **Retrieve handler** reads prefetched results from L1 and does H2D scatter to vLLM GPU paged KV cache via CUDA IPC

The core challenge: **L1 is the first-class citizen in lmcache MP mode, it serves as the unified intermediary hub for multiple L2 adapters within lmcache. The kvcache is loaded into L1 uniformly from each L2 adapter, and then distributed to the vLLM workers via a unified H2D scatter**.

To match this architecture，the previous GDS backend implementation (in the now-deprecated in-process mode) would DMA to a GPU buffer and then D2H back to the CPU obj, followed by the standard H2D scatter to the vLLM worker. 

This completely negated the performance advantage of GPU-direct storage — the data still traversed PCIe twice (NVMe→GPU via DMA, then GPU→CPU via D2H), plus an extra GPU→CPU copy that wouldn't exist in a plain POSIX read path.

We solve this with an **asymmetric load and store** design and a **Device-Resident MemoryObj** flow.

#### 3.2 Asymmetric PHX Design

| Operation | Data Path | I/O Mechanism | Rationale |
|---|---|---|---|
| **Store** | CPU MemoryObj → POSIX write → NVMe | `phxfs_write_batch` / POSIX | the KV cache must first be written to L1 in a unified manner for parallel L2 adapters |
| **Load** | NVMe → `phxfs_read_batch` DMA → GPU MemoryObj | Phoenix DMA | **Performance-critical path** — bypasses CPU entirely |

The key insight: PHX value is only realized if the DMA'd data **stays on GPU** and flows directly to the retrieve handler's scatter kernel via D2D (device-to-device copy). If data is D2H'd back to CPU, the DMA benefit is negated.

#### 3.3 Device-Resident MemoryObj Flow

To let GPU-resident data flow from adapter to retrieve without going through CPU, we extended the LMCache framework with a **device-obj injection mechanism**:

```
LOOKUP phase (PrefetchController):
  1. Broadcast lookup to all adapters
  2. Collect bitmaps

LOAD phase (PrefetchController):
  3. submit_load_task → PhxL2Adapter
  4. PhxL2Adapter._process_load:
     a. Allocate GPU MemoryObj from PhxDeviceMemoryAllocator (phx pool)
     b. phxfs_read_batch DMA: NVMe → GPU MemoryObj (batched, grouped by device)
     c. Store GPU objs in _load_device_objs[task_id]
  5. PrefetchController._poll_load_results:
     a. pop_loaded_device_objs(task_id) → gets GPU MemoryObjs
     b. L1Manager.replace_memory_obj(key, gpu_obj) — swaps CPU obj for GPU obj in L1

RETRIEVE phase (retrieve handler):
  6. read_prefetched_results → gets GPU MemoryObj from L1
  7. transfer_kv_per_object_group(H2D):
     - PyTorch copy_ auto-detects device: GPU src → D2D (no PCIe!)
     - multi_layer_kv_transfer kernel scatters to vLLM paged KV
  8. release_device_objs(keys) — recycle DMA buffer (or D2H write-back to CPU obj)
```

This design preserves LMCache's existing architecture: the PrefetchController still manages the load lifecycle, L1 still serves as the cache layer, and the retrieve handler is unchanged except for the `copy_` device auto-detection.

#### 3.4 Multi-Device Support

The adapter supports multiple GPUs via `device_ids` configuration. Keys are routed to devices by `kv_rank` (extracted from the `ObjectKey`), with one `PhxCache` + `PhxDeviceMemoryAllocator` per device. Batch reads are grouped per device for concurrent I/O.

#### 3.5 Graceful Fallback

When Phoenix is unavailable (e.g., `phxcache` not installed, device init failed, or pool exhausted), the adapter transparently falls back to POSIX read, filling the controller-provided CPU MemoryObj. This ensures the PR introduces no hard dependency on Phoenix hardware.

### 4. End-to-End Performance Results

We conducted evaluations on an 8×NVIDIA H20（96GB HBM each）and 4×NVMe Gen4 SSDs server. We configured these 4 SSDs in a RAID 0 array, achieving a total read/write bandwidth of 26/16 GB/s.

We evaluated the Phoenix L2 adapter with two end-to-end workloads, including an **open-source real-world agent workload traces**  and  a **production-scale internal workload traces**, validating the Phoenix L2 adapter across different workload scales.

---

#### Experiment 1: Open-Source Agent Trace (codex_swebenchpro_traces)

**Workload:**
- **Trace dataset**: [codex_swebenchpro_traces](https://huggingface.co/datasets/Inferact/codex_swebenchpro_traces) — a typical agent-mode long-context trace dataset derived from SWE-bench Pro
- **Model**: [Open-Source LLM Model](https://huggingface.co/tencent/Hy3)  (TP=8, ~600 GB GPU memory across 8×H20)
- **Concurrency**: 50 traces replayed simultaneously, representing **50 concurrent users** in a round-robin interleaved fashion
- **Context scale**: Max prompt context = **165,915 tokens**; Max per-user KV cache = **~50 GB**; Total KV cache across all 50 users = **~1.5 TB**
- **L1 cache (eval)**: 61 GB / 500 GB (deliberately small to force L2 lookups)

**Results:**

| System | Mean TTFT (s) | Speedup vs. PHX |
|---|---|---|
| No offloading (recompute) | 5.593 | PHX is **6.12×** faster |
| Prefill | 1.509 | PHX is **1.77×** faster |
| SSD via `fs_native` (L1=500 GB) | 1.105 | PHX is **1.24×** faster |
| SSD via `fs_native` (L1=61 GB) | 1.271 | PHX is **1.43×** faster |
| **PHX (Phoenix DMA, L1=61 GB)** | **0.887** | — |

Key takeaways:
- Phoenix with only **61 GB L1** outperforms `fs_native` with **500 GB L1** by 1.24× — demonstrating that Phoenix's fast DMA path dramatically reduces the need for large CPU DRAM caches, lowering deployment cost.
- Phoenix is **1.43× faster** than `fs_native` under identical L1 capacity (61 GB), purely from eliminating the CPU intermediary on the retrieve path.
- Compared to no cache offloading (full recomputation), Phoenix delivers **6.12×** TTFT reduction.

---
#### Experiment 2: Large-Scale Concurrency Sweep (Internal Trace)

**Workload:**
- **Trace dataset**: An internal production trace dataset, containing **11,596 requests** with contexts up to 256K tokens
- **Model**:  [Open-Source LLM Model](https://huggingface.co/tencent/Hy3-FP8)  (TP=8, ~300 GB GPU memory across 8×H20)
- **Concurrency sweep**: 1 → 100 concurrent users, each concurrency level processing ~1,000 requests
- **L1 cache**: 1 TB DRAM (same across all three systems)
- **Systems compared**:
  - **dram_only** — no L2 offloading (L1 only)
  - **ssd+dram** — L2 via `fs_native` (POSIX read through CPU)
  - **phx+dram** — L2 via Phoenix (GPU-direct DMA)

**Results — Mean TTFT by concurrency bucket:**

| Concurrency | dram_only | ssd+dram | phx+dram |
|---|---|---|---|
| Low (1–20) | 8.386 s | 3.217 s | 3.276 s |
| Mid (21–50) | 27.245 s | 7.810 s | **6.154 s**|
| High (51–100) | 38.924 s † | 15.506 s | **11.048 s** |

Key takeaways:
1. **L2 offloading is essential for high concurrency.** Without L2 (`dram_only`), the system saturates at CC=55 with mean TTFT exploding to 38.9 s. Both L2-backed systems sustain CC=100 — an **82% increase** in max concurrency.
2. **Phoenix delivers the best mean TTFT, especially at high concurrency.** In the 51–100 range, `phx+dram` achieves **11.05 s** mean TTFT vs `ssd+dram`'s 15.51 s — an **27% improvement**. 
3. **At low concurrency (1–20), all systems perform similarly** — the bottleneck is GPU prefill compute, not I/O, so the L2 backend choice has minimal impact.

---

## **Special notes for your reviewers**

- This is my first PR to LMCache 🙇 please be gentle.
- I ran `pre-commit run --all-files` locally — all green.
- I ran `pytest -xvs tests/...` — all green.

### 1. Modification Summary

This PR touches the LMCache MP-mode distributed layer. The changes fall into three categories:

#### New Files (3 files, ~1,370 lines)

| File | Description |
|---|---|
| `lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py` | Main `PhxL2Adapter` implementation: asymmetric PHX (POSIX store + DMA load), batch read, fd cache, multi-device routing, perf instrumentation |
| `lmcache/v1/memory_allocators/phx_device_memory_allocator.py` | `PhxDeviceMemoryAllocator`: GPU memory pool with `phxfs_regmem` registration for DMA, vendor-neutral alignment (queries `page_size` at runtime) |
| `lmcache/v1/distributed/memory_manager/phx_l1_memory_manager.py` | `PhxL1MemoryManager`: extends `L1MemoryManager`, overrides `free()` to dispatch device-resident objs to their parent allocator (instead of CPU slab) |

#### Modified Files (7 files, ~140 lines)

| File | Change |
|---|---|
| `lmcache/v1/distributed/config.py` | Add `phx_l2_enabled` flag (auto-detected from L2 adapter config) |
| `lmcache/v1/distributed/l1_manager.py` | Add `replace_memory_obj()` (swap CPU obj → GPU obj in L1) + `release_device_objs()` (recycle DMA buffer after D2D, optional write-back) |
| `lmcache/v1/distributed/memory_manager/__init__.py` | Export `PhxL1MemoryManager` |
| `lmcache/v1/memory_allocators/__init__.py` | Export `PhxDeviceMemoryAllocator` | 
| `lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py` | Add specific checks for device memory objects to prevent entering the fast path reserved for CPU memory objects only. | 
| `tests/v1/distributed/test_phx_l2_adapter.py` | unittests | 
| `docs/design/v1/distributed/l2_adapters/phx_l2_adapter.md` | design doc | 

#### External Package (not part of this repo)

The `phxcache` pybind11 C++ extension (wrapping the Phoenix `phxfs_*` C API) is maintained in the [Phoenix repository](https://github.com/xPU-IO/phoenix) under **`adapters/lmcache/phxcache/`**. It provides `PhxCache`, `PhxFile` classes with `read`/`read_batch`/`write`/`write_batch`/`regmem`/`deregmem` methods. 

The adapter imports it lazily (`import phxcache` inside `_init_devices`), so LMCache works without it installed (POSIX fallback).

#### Prerequisites

1. **Phoenix kernel module** (`phoenixfs`) + **user library** (`libphoenix.so`): Build and install from the [Phoenix source](https://github.com/xPU-IO/phoenix).
```bash
   # Clone the Phoenix repository
   git clone https://github.com/xPU-IO/phoenix.git
   cd phoenix

   # Build the kernel module, user library, and tests
   mkdir -p build && cd build
   cmake ../
   make -j

   # Install to system paths (libphoenix.so → /usr/local/lib,
   # headers → /usr/local/include, kernel module → /lib/modules/$(uname -r)/extra)
   sudo make install

   # Load the NVIDIA driver first, then insert the phoenixfs kernel module
   nvidia-smi  # ensures NVIDIA driver is loaded
   sudo make insmod   # loads phoenixfs.ko (STAGING mode by default)

   # Verify the module is loaded and device nodes exist
   lsmod | grep phoenixfs
   ls /dev/phxfs_dev*   # should show /dev/phxfs_dev0, /dev/phxfs_dev1, ...
   ```
   > **Note**: Phoenix is self-contained — it does **not** require GPUDirect Storage (`nvidia-fs`), a specific filesystem, or any RDMA stack. It issues `O_DIRECT` I/O on whatever fd the application opens. See the [Phoenix install guide](https://github.com/xPU-IO/phoenix/blob/main/doc/install.md) for details on BAR mapping modes (STAGING vs FULL), vendor backend selection, and troubleshooting.

2. **`phxcache` Python package**: pybind11 extension wrapping `libphoenix.so`, located at `adapters/lmcache/phxcache/` in the Phoenix repository.
```bash
   cd phoenix/adapters/lmcache/phxcache
   bash install.sh   # auto-detects conda env, sets CUDA_HOME/LIBRARY_PATH/etc.
   ```

3. **Hardware**: NVIDIA GPUs + NVMe SSDs.

#### Configuration

Add a `phx` L2 adapter to your LMCache MP server config:

```json
--l2-adapter '{"type": "phx", "base_path": "/mnt/nvme/kv_cache", "device_ids": [4, 5, 6, 7], "buffer_size_mb": 2048}'
```

| Parameter | Type | Description |
|---|---|---|
| `base_path` | str | Root directory for KV cache files |
| `device_ids` | list[int] | CUDA GPU IDs for phxfs DMA (one buffer pool per device) |
| `buffer_size_mb` | int | GPU buffer pool size per device (default: 2048 MiB) |
| `use_direct_io` | bool | Use `O_DIRECT` for I/O (default: true) |
| `max_capacity_bytes` | int | Max storage, 0=unlimited (default: 0) |
| `perf_log_dir` | str | Optional perf log directory (hit rate + per-phase timing) |

The `phx_l2_enabled` flag is auto-detected — when any L2 adapter has `type: "phx"`, L1 automatically uses `PhxL1MemoryManager` for device-obj dispatch. No extra config needed.

---

**If applicable:**

### References

1. Jianqin Yan, Shi Qiu, et al. "Phoenix: A Refactored I/O Stack for GPU Direct Storage without Phony Buffers." *Proceedings of the International Conference for High Performance Computing, Networking, Storage and Analysis (SC '25)*, November 2025. 
2. Phoenix project GitHub: [https://github.com/xPU-IO/phoenix](https://github.com/xPU-IO/phoenix) — `phoenixfs` kernel module, `libphoenix` userspace library, and `phxcache` lmcache adapter.
3. xPU.IO: [https://github.com/xPU-IO](https://github.com/xPU-IO) --An open-source community co-initiated by NICELab (Xiamen University) and Tencent DCPD, focusing on xPU I/O for AI.
4. codex_swebenchpro_traces dataset: [https://huggingface.co/datasets/Inferact/codex_swebenchpro_traces](https://huggingface.co/datasets/Inferact/codex_swebenchpro_traces)

### DIFF
--- a/docs/design/v1/distributed/l2_adapters/phx_l2_adapter.md
+++ b/docs/design/v1/distributed/l2_adapters/phx_l2_adapter.md
@@ -1,0 +1,151 @@
+# Phoenix L2 Adapter
+
+`lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py`
+
+A disk-based L2 adapter that uses [Phoenix](https://github.com/xPU-IO/phoenix)
+GPU-direct storage I/O to load KV cache data directly from NVMe to GPU memory
+via DMA, bypassing the host CPU entirely on the retrieve path.
+
+## Motivation
+
+Existing disk-based L2 adapters (`fs_native`, `fs`) retrieve data via POSIX
+`read()`, which forces data through host CPU memory:
+
+```
+NVMe →(POSIX read)→ L1 CPU MemoryObj →(H2D scatter)→ GPU paged KV    [2× PCIe]
+```
+
+Phoenix eliminates the CPU intermediary on the load path:
+
+```
+NVMe →(phxfs DMA)→ GPU MemoryObj →(D2D scatter)→ GPU paged KV    [1× PCIe + 1× GPU-internal]
+```
+
+This halves PCIe traffic on retrieve and reduces latency, especially for
+long-context workloads with large KV cache chunks.
+
+## Asymmetric Design
+
+The adapter uses **asymmetric store/load** because LMCache's MP-mode
+architecture always hands adapters CPU `MemoryObj` buffers (L1 is CPU DRAM):
+
+| Operation | Data Path | I/O Mechanism |
+|---|---|---|
+| **Store** | CPU MemoryObj → POSIX write → NVMe | `phxfs_write_batch` or POSIX `write` |
+| **Load** | NVMe → `phxfs_read_batch` DMA → GPU MemoryObj | Phoenix DMA (preferred) or POSIX `read` (fallback) |
+
+Store goes through CPU because the data is already in CPU memory (written by
+the StoreController from L1). Trying to DMA from CPU to NVMe via Phoenix would
+add no benefit over a standard POSIX write.
+
+Load is the performance-critical path. The key insight: PHX value is only
+realized if the DMA'd data **stays on GPU** and flows to the retrieve handler
+via D2D (device-to-device copy). If data is D2H'd back to CPU, the DMA benefit
+is negated.
+
+## Device-Resident MemoryObj Flow
+
+To let GPU-resident data flow from adapter to retrieve without going through
+CPU, the adapter extends the framework with a **device-obj injection mechanism**:
+
+```
+LOOKUP phase (PrefetchController):
+  1. Broadcast lookup to all adapters
+  2. Collect bitmaps
+
+LOAD phase (PrefetchController):
+  3. submit_load_task → PhxL2Adapter
+  4. PhxL2Adapter._process_load:
+     a. Allocate GPU MemoryObj from PhxDeviceMemoryAllocator (phx pool)
+     b. phxfs_read_batch DMA: NVMe → GPU MemoryObj (batched, grouped by device)
+     c. Store GPU objs in _load_device_objs[task_id]
+  5. PrefetchController._poll_load_results:
+     a. pop_loaded_device_objs(task_id) → gets GPU MemoryObjs
+     b. L1Manager.replace_memory_obj(key, gpu_obj) — swaps CPU obj for GPU obj in L1
+
+RETRIEVE phase (retrieve handler):
+  6. read_prefetched_results → gets GPU MemoryObj from L1
+  7. transfer_kv_per_object_group:
+     - PyTorch copy_ auto-detects device: GPU src → D2D (no PCIe!)
+     - multi_layer_kv_transfer kernel scatters to vLLM paged KV
+  8. release_device_objs(keys) — recycle DMA buffer (or D2H write-back to CPU obj)
+```
+
+This preserves LMCache's existing architecture: the PrefetchController still
+manages the load lifecycle, L1 still serves as the cache layer, and the
+retrieve handler is unchanged except for `copy_` device auto-detection.
+
+`pop_loaded_device_objs()` returns an empty dict for keys that fell back to
+POSIX read (those filled the controller-provided CPU obj directly) or for
+tasks that have already been popped. The base class `L2AdapterInterface`
+provides a no-op default so other adapters are unaffected.
+
+## Fallback
+
+When Phoenix is unavailable, the adapter transparently falls back to POSIX
+read/write. This happens when:
+
+- `phxcache` Python package is not installed (`ImportError` in `_init_devices`)
+- `device_ids` is not configured (no DMA devices initialized)
+- Device initialization fails (e.g., no Phoenix kernel module)
+- DMA buffer pool is exhausted at load time (per-key fallback)
+
+In fallback mode, `is_phx_available()` returns `False`, load fills the
+controller-provided CPU `MemoryObj` directly via POSIX `read()`, and
+`pop_loaded_device_objs()` returns an empty dict. The adapter introduces **no
+hard dependency** on Phoenix hardware.
+
+## Multi-Device Support
+
+The adapter supports multiple GPUs via `device_ids` configuration. Keys are
+routed to devices by `kv_rank` (extracted from `ObjectKey`):
+
+```python
+def _kv_rank_to_device(self, kv_rank: int) -> int:
+    # kv_rank = (world_size << 24) | (global_rank << 16)
+    #         | (local_world_size << 8) | local_rank
+    return (kv_rank >> 16) & 0xFF  # global_rank → CUDA device id
+```
+
+One `PhxCache` + `PhxDeviceMemoryAllocator` is created per device. Batch reads
+are grouped per device for concurrent I/O.
+
+## Configuration
+
+```json
+{
+  "type": "phx",
+  "base_path": "/mnt/nvme/kv_cache",
+  "device_ids": [4, 5, 6, 7],
+  "buffer_size_mb": 2048,
+  "use_direct_io": true,
+  "max_capacity_bytes": 0,
+  "perf_log_dir": null
+}
+```
+
+| Parameter | Type | Default | Description |
+|---|---|---|---|
+| `base_path` | str | (required) | Root directory for KV cache files |
+| `device_ids` | list[int] | `None` | CUDA GPU IDs for phxfs DMA (one buffer pool per device). `None` → POSIX-only mode |
+| `buffer_size_mb` | int | 2048 | GPU buffer pool size per device (MiB) |
+| `use_direct_io` | bool | true | Use `O_DIRECT` for I/O |
+| `max_capacity_bytes` | int | 0 | Max storage, 0=unlimited |
+| `perf_log_dir` | str | `None` | When set, writes perf log (hit rate + per-phase timing) to this dir |
+
+When `device_ids` is omitted or `phxcache` is not installed, the adapter
+operates in POSIX fallback mode — no GPU or Phoenix hardware is required.
+
+## External Dependency
+
+The `phxcache` pybind11 C++ extension (wrapping the Phoenix `phxfs_*` C API)
+is maintained in the [Phoenix repository](https://github.com/xPU-IO/phoenix)
+under `adapters/lmcache/phxcache/`. It is imported lazily inside
+`_init_devices()`, so LMCache works without it installed (POSIX fallback).
+
+## close()
+
+Stops the background worker thread, closes all cached read file descriptors,
+releases any device objs never popped by the controller (to avoid leaking phx
+pool memory), frees device allocators, closes PhxCache instances, and closes
+the three event notifiers. `close()` is idempotent.
--- a/lmcache/v1/distributed/config.py
+++ b/lmcache/v1/distributed/config.py
@@ -211,12 +211,6 @@ class L1ManagerConfig:
     read_ttl_seconds: int = field(default=300)
     """ Time to live for each object's read lock. Default is 300s (5 minutes). """
 
-    enable_write_back: bool = False
-    """ When True, device-resident MemoryObjs in L1 are D2H-copied to CPU
-    objs after D2D transfer, preserving the L1 cache entry for future H2D
-    hits.  When False (default), device objs are deleted after D2D,
-    freeing the DMA buffer immediately but losing the L1 cache entry. """
-
 
 @dataclass
 class EvictionConfig:
@@ -465,14 +459,6 @@ def add_storage_manager_args(
         default=300,
         help="Time to live for each object's read lock. Default is 300s.",
     )
-    ttl_group.add_argument(
-        "--enable-write-back",
-        action="store_true",
-        default=False,
-        help="D2H-copy device-resident MemoryObjs to CPU after D2D transfer, "
-        "preserving L1 cache entries for future H2D hits. Default is False "
-        "(delete device objs after D2D to immediately free DMA buffer).",
-    )
 
     # Eviction Config
     eviction_group = parser.add_argument_group(
@@ -625,7 +611,6 @@ def parse_args_to_config(
         phx_l2_enabled=phx_l2_enabled,
         write_ttl_seconds=args.l1_write_ttl_seconds,
         read_ttl_seconds=args.l1_read_ttl_seconds,
-        enable_write_back=getattr(args, "enable_write_back", False),
     )
 
     eviction_config = EvictionConfig(
--- a/lmcache/v1/distributed/config.py
+++ b/lmcache/v1/distributed/config.py
@@ -198,11 +198,24 @@ class L1ManagerConfig:
     """ Optional GDS L1 tier. When set, the GDS slab is the L1 medium
     (mutually exclusive with the pinned-DRAM tier in ``memory_config``). """
 
+    phx_l2_enabled: bool = False
+    """ When True, L1 uses :class:`PhxL1MemoryManager` whose ``free()``
+    dispatches device-resident MemoryObjs (injected by PhxL2Adapter via
+    :meth:`L1Manager.replace_memory_obj`) to their parent allocator, while
+    CPU objs still go through the CPU pinned-DRAM allocator.  Mutually
+    exclusive with GDS L1 and Device-DAX L1. """
+
     write_ttl_seconds: int = field(default=600)
     """ Time to live for each object's write lock. Default is 600s (10 minutes). """
 
     read_ttl_seconds: int = field(default=300)
     """ Time to live for each object's read lock. Default is 300s (5 minutes). """
+
+    enable_write_back: bool = False
+    """ When True, device-resident MemoryObjs in L1 are D2H-copied to CPU
+    objs after D2D transfer, preserving the L1 cache entry for future H2D
+    hits.  When False (default), device objs are deleted after D2D,
+    freeing the DMA buffer immediately but losing the L1 cache entry. """
 
 
 @dataclass
@@ -301,6 +314,15 @@ def validate_storage_manager_config(config: StorageManagerConfig) -> None:
         and config.l1_manager_config.memory_config.devdax_path
     ):
         raise ValueError("gds-l1-path cannot be used with l1-devdax-path")
+
+    if config.l1_manager_config.phx_l2_enabled:
+        if config.l1_manager_config.gds_l1_config is not None:
+            raise ValueError(
+                "phx L2 adapter cannot be used with GDS L1 "
+                "(GDS L1 exposes no registerable memory region for DMA)"
+            )
+        if config.l1_manager_config.memory_config.devdax_path:
+            raise ValueError("phx L2 adapter cannot be used with Device-DAX L1")
 
     memory_config = config.l1_manager_config.memory_config
     if not (memory_config.devdax_path and memory_config.devdax_size_in_bytes):
@@ -443,6 +465,14 @@ def add_storage_manager_args(
         default=300,
         help="Time to live for each object's read lock. Default is 300s.",
     )
+    ttl_group.add_argument(
+        "--enable-write-back",
+        action="store_true",
+        default=False,
+        help="D2H-copy device-resident MemoryObjs to CPU after D2D transfer, "
+        "preserving L1 cache entries for future H2D hits. Default is False "
+        "(delete device objs after D2D to immediately free DMA buffer).",
+    )
 
     # Eviction Config
     eviction_group = parser.add_argument_group(
@@ -583,11 +613,19 @@ def parse_args_to_config(
             use_direct_io=args.gds_l1_use_direct_io,
         )
 
+    l2_adapter_config = parse_args_to_l2_adapters_config(args)
+
+    phx_l2_enabled = any(
+        get_type_name_for_config(cfg) == "phx" for cfg in l2_adapter_config.adapters
+    )
+
     l1_manager_config = L1ManagerConfig(
         memory_config=memory_config,
         gds_l1_config=gds_l1_config,
+        phx_l2_enabled=phx_l2_enabled,
         write_ttl_seconds=args.l1_write_ttl_seconds,
         read_ttl_seconds=args.l1_read_ttl_seconds,
+        enable_write_back=getattr(args, "enable_write_back", False),
     )
 
     eviction_config = EvictionConfig(
@@ -597,8 +635,6 @@ def parse_args_to_config(
         extra_logging_enabled=getattr(args, "enable_extra_logging", False),
         extra_logging_interval=getattr(args, "extra_logging_interval", 10.0),
     )
-
-    l2_adapter_config = parse_args_to_l2_adapters_config(args)
 
     config = StorageManagerConfig(
         l1_manager_config=l1_manager_config,
--- a/lmcache/v1/distributed/config.py
+++ b/lmcache/v1/distributed/config.py
@@ -186,6 +186,24 @@ class GdsL1Config:
 
 
 @dataclass
+class PhxL1Config:
+    """Configuration for the Phoenix L1 tier.
+
+    When present on :class:`L1ManagerConfig`, L1 uses
+    :class:`PhxL1MemoryManager` whose ``free()`` dispatches device-resident
+    MemoryObjs (injected by PhxL2Adapter via
+    :meth:`L1Manager.replace_memory_obj`) to their parent allocator, while
+    CPU objs still go through the CPU pinned-DRAM allocator.  Mutually
+    exclusive with GDS L1 and Device-DAX L1.
+
+    The presence of this config is inferred automatically at parse time from
+    the L2 adapter config (any adapter with ``type == "phx"``); users do not
+    set it directly.  Fields are reserved for future PhxL1MemoryManager
+    tuning (e.g. device-obj pool sizing).
+    """
+
+
+@dataclass
 class L1ManagerConfig:
     """
     Special config for the L1 Object/Key manager
@@ -198,9 +216,10 @@ class L1ManagerConfig:
     """ Optional GDS L1 tier. When set, the GDS slab is the L1 medium
     (mutually exclusive with the pinned-DRAM tier in ``memory_config``). """
 
-    phx_l2_enabled: bool = False
-    """ When True, L1 uses :class:`PhxL1MemoryManager` whose ``free()``
-    dispatches device-resident MemoryObjs (injected by PhxL2Adapter via
+    phx_l1_config: "PhxL1Config | None" = None
+    """ Optional Phoenix L1 tier. When set, L1 uses
+    :class:`PhxL1MemoryManager` whose ``free()`` dispatches
+    device-resident MemoryObjs (injected by PhxL2Adapter via
     :meth:`L1Manager.replace_memory_obj`) to their parent allocator, while
     CPU objs still go through the CPU pinned-DRAM allocator.  Mutually
     exclusive with GDS L1 and Device-DAX L1. """
@@ -309,7 +328,7 @@ def validate_storage_manager_config(config: StorageManagerConfig) -> None:
     ):
         raise ValueError("gds-l1-path cannot be used with l1-devdax-path")
 
-    if config.l1_manager_config.phx_l2_enabled:
+    if config.l1_manager_config.phx_l1_config is not None:
         if config.l1_manager_config.gds_l1_config is not None:
             raise ValueError(
                 "phx L2 adapter cannot be used with GDS L1 "
@@ -601,14 +620,19 @@ def parse_args_to_config(
 
     l2_adapter_config = parse_args_to_l2_adapters_config(args)
 
-    phx_l2_enabled = any(
+    # Infer PhxL1Config presence from the L2 adapter config: any "phx"
+    # adapter requires PhxL1MemoryManager so device-resident objects injected
+    # by the adapter are dispatched correctly on free().
+    phx_l1_config: PhxL1Config | None = None
+    if any(
         get_type_name_for_config(cfg) == "phx" for cfg in l2_adapter_config.adapters
-    )
+    ):
+        phx_l1_config = PhxL1Config()
 
     l1_manager_config = L1ManagerConfig(
         memory_config=memory_config,
         gds_l1_config=gds_l1_config,
-        phx_l2_enabled=phx_l2_enabled,
+        phx_l1_config=phx_l1_config,
         write_ttl_seconds=args.l1_write_ttl_seconds,
         read_ttl_seconds=args.l1_read_ttl_seconds,
     )
--- a/lmcache/v1/distributed/l1_manager.py
+++ b/lmcache/v1/distributed/l1_manager.py
@@ -19,6 +19,7 @@ from lmcache.v1.distributed.memory_manager import (
     GDSL1MemoryManager,
     L1ManagerProtocol,
     L1MemoryManager,
+    PhxL1MemoryManager,
 )
 from lmcache.v1.distributed.memory_manager.devdax_l1_memory_manager import (
     DevDaxL1MemoryManager,
@@ -200,6 +201,11 @@ class L1Manager:
         elif config.memory_config.devdax_path:
             self._memory_manager = DevDaxL1MemoryManager(config.memory_config)
             logger.info("L1Manager: Device-DAX L1 tier enabled; CPU-only L1 disabled")
+        elif config.phx_l2_enabled:
+            self._memory_manager = PhxL1MemoryManager(config.memory_config)
+            logger.info(
+                "L1Manager: PhxL1MemoryManager enabled (CPU DRAM + device-obj dispatch)"
+            )
         else:
             self._memory_manager = L1MemoryManager(config.memory_config)
 
@@ -229,6 +235,39 @@ class L1Manager:
                 "L1 used/total ratio (0.0–1.0)",
                 lambda: _l1_usage_ratio_or_zero(L1Manager._gauge_target),
             )
+
+    @l1_mgr_synchronized
+    def replace_memory_obj(self, key: ObjectKey, new_obj: MemoryObj) -> L1Error:
+        """Replace the MemoryObj stored for ``key`` with ``new_obj``.
+
+        Used by the prefetch controller when an L2 adapter (e.g.
+        ``PhxL2Adapter``) produces a device-resident MemoryObj that should be
+        served from L1 instead of the CPU buffer pre-allocated by
+        ``reserve_write``. The old CPU obj is freed via the L1 memory
+        manager; the new obj's lifecycle is owned by its own allocator
+        (freed via the memory manager's ``free`` dispatch on subsequent
+        eviction/delete/read-finish).
+
+        The key must currently exist in L1 (typically write-locked, between
+        ``reserve_write`` and ``finish_write``). Object state (locks,
+        temporary flag) is preserved; only the underlying buffer reference
+        is swapped.
+
+        Args:
+            key: The object key whose MemoryObj should be replaced.
+            new_obj: The new MemoryObj to store (may be device-resident).
+
+        Returns:
+            L1Error.SUCCESS on success, L1Error.KEY_NOT_EXIST otherwise.
+        """
+        entry = self._objects.get(key, None)
+        if entry is None:
+            return L1Error.KEY_NOT_EXIST
+        old_obj = entry.memory_obj
+        entry.memory_obj = new_obj
+        if old_obj is not None:
+            self._memory_manager.free([old_obj])
+        return L1Error.SUCCESS
 
     def register_listener(self, listener: L1ManagerListener) -> None:
         """Register a listener for L1Manager events.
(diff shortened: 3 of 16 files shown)


lmcache/v1/memory_allocators/phx_device_memory_allocator.py
```diff
@@ -0,0 +1,87 @@
...
+
+    :param size: Size of the device memory pool in bytes.
+    :param device: Torch device string (e.g. ``"cuda:4"``).  Must match the
+        device that ``phx_cache`` is bound to.
+    :param phx_cache: A ``phxcache.PhxCache`` instance with an open Phoenix
+        device connection.  Required.
+    """
+
+    def __init__(self, size: int, device=None, phx_cache=None) -> None:
+        assert phx_cache is not None, (
```
maobaolong, line 59: `phx_cache` I guess this should be imported outside of this repo, so it would be easy to use if the user-facing document could show how to build and install the dependency from `phoenix` or even give a link can helps.

MonthFall: Thank you for your valuable feedback! 

We are currently writing a PhxL2Adapter user handbook to help users quickly and easily experience the benefits of Phoenix. 
This handbook can be considered an expanded version of  **Deployment Guide: Phoenix + LMCache.** 
In addition to instructions on how to build and install the Phoenix kernel module and the phxcache Python package, we will also provide some examples of startup scripts.

We will submit this handbook to the [lmcache documentation](https://docs.lmcache.ai/#) soon.

maobaolong: @MonthFall Thanks for the feedback, as deploy Phoenix need install kernel module, this would be highlight to heads-up potential users to aware this restriction.

maobaolong:
@MonthFall Another feedback from @ApostaC , he is considering slight refactoring for the storage manager and prefetch controller. Regarding the Phoenix PR, please minimize your changes on them.

MonthFall:
> @MonthFall Another feedback from @ApostaC , he is considering slight refactoring for the storage manager and prefetch controller. Regarding the Phoenix PR, please minimize your changes on them.

Hi @maobaolong , thanks for relaying @ApostaC's feedback! 

We deeply agreed that this is an excellent suggestion: decoupling from the core components of lmcache will effectively improve the compatibility of Phxl2adapter.

Our latest implementation ([adce30f](https://github.com/LMCache/LMCache/pull/4474/commits/adce30f64a0cf326d8ab4791c899010180a62947)) already achieves zero diff on both **storage_manager.py** and **prefetch_controller.py** (as well as **l2_adapters/base.py**). These three files are now byte-identical to the upstream dev branch.

We did this by leveraging LMCache's existing is_temporary mechanism: instead of having the prefetch controller and storage manager manually call **release_device_objs** to recycle DMA buffers, the Phoenix adapter now self-admits device-resident MemoryObjs into L1 via **replace_memory_obj(key, obj, mark_temporary=True)** inside its own **query_load_result**. The framework's built-in finish_read → free() path automatically handles buffer recycling when read locks reach zero — no framework code needed.

All Phoenix-specific changes are now confined to:

- **l1_manager.py**: one new optional parameter (mark_temporary) on replace_memory_obj + a small module-level registry (weakref) so the adapter can obtain the L1Manager reference
- **phx_l2_adapter.py**: self-admission logic in query_load_result
- **config.py**: phx_l2_enabled flag (already present)

Happy to walk through the design in detail if helpful.

maobaolong:
@MonthFall Would you like to fix the DCO issue first? It is easy to fix, just sign-off your commit.

lmcache/v1/distributed/storage_controllers/prefetch_controller.py
```diff
@@ -1339,6 +1339,7 @@ def _finish_request(self, request: InFlightPrefetchRequest) -> None:
         # |       -        |     -      |       -        |load→locked |     -      |
         # SW keys in L1:
         # |     unlock     |   unlock   |     unlock     |   locked   |   unlock   |
+        # Transition loaded keys out of write-locked state.
```
maobaolong: Is this comment necessary?

lmcache/v1/distributed/storage_manager.py
```diff
@@ -386,6 +386,7 @@ def finish_read_prefetched(
         finish_result = self._l1_manager.finish_read(keys, extra_count=extra_count)
         successful_keys = [k for k, e in finish_result.items() if e == L1Error.SUCCESS]
         failed_keys = [k for k, e in finish_result.items() if e != L1Error.SUCCESS]
+
```
maobaolong: @MonthFall How about revert this extra blank line?

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -486,8 +486,15 @@ def transfer_kv_per_object_group(
         This function expects the caller to stage the block ids (list[list[int]])
         into GPU tensors and pass them in as `block_ids_gpu`.
     """
-    if _HAS_NATIVE_OBJECT_GROUP_TRANSFER and not any(
-        isinstance(mo, GDSMemoryObject) for mo in memory_objs
+    if (
```
maobaolong: Can you add comment to explain the changes?

lmcache/v1/distributed/l1_manager.py
```diff
@@ -200,6 +223,11 @@ def __init__(self, config: L1ManagerConfig):
         elif config.memory_config.devdax_path:
             self._memory_manager = DevDaxL1MemoryManager(config.memory_config)
             logger.info("L1Manager: Device-DAX L1 tier enabled; CPU-only L1 disabled")
+        elif config.phx_l2_enabled:
+            self._memory_manager = PhxL1MemoryManager(config.memory_config)
```
maobaolong, line 232: If we explicitly specify that the L2 is the Phoenix L2 adapter through the L2 adapter parameter, we can automatically infer here that we must use the `PhxL1MemoryManager`. So, do we really need to provide an additional, redundant configuration item to the user?

maobaolong (commented):
@MonthFall Left a couple of comment.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
MonthFall: Thanks for the suggestion! We've added an inline comment above the fast-path guard explaining the rationale. Here's a summary of what the comment covers:

The fast path (_run_object_group_transfer_plan) is a CPU-MO-only path — it assumes every memory object is backed by host memory so the native object-group transfer can issue DMA straight from CPU buffers. Entering it with a device-resident memory object (DMO, e.g. the phx GPU-resident objects) would DMA from a wrong address space and corrupt data.

Three guards keep us in the fast path only when it is safe:

- _HAS_NATIVE_OBJECT_GROUP_TRANSFER — the native fast path exists.
- No GDSMemoryObject — historically this was sufficient to guarantee "all MOs are CPU MOs," because GDSMemoryObject was the only non-CPU object type.
- No MO has a non-CPU raw_tensor — added for phx device-resident objects (DMOs), which are not GDSMemoryObject and therefore slip past guard 2.

The comment also notes that guard 3 is strictly broader than guard 2 and would suffice on its own; guard 2 is kept to preserve the original GDS-specific semantics and to short-circuit the more expensive tensor inspection for GDS paths.

lmcache/v1/distributed/l1_manager.py
MonthFall: Great point — you're absolutely right that the boolean was redundant. We've removed phx_l2_enabled entirely and replaced it with a PhxL1Config dataclass, mirroring the existing GdsL1Config pattern.

Here's what changed:

New PhxL1Config (in config.py, right after GdsL1Config): an intentionally minimal config class — currently empty, with fields reserved for future PhxL1MemoryManager tuning (e.g. device-obj pool sizing).

L1ManagerConfig.phx_l1_config: PhxL1Config | None = None: replaces the old phx_l2_enabled: bool field. This is fully symmetric with gds_l1_config: GdsL1Config | None = None.

Automatic inference at parse time: PhxL1Config() is constructed when any L2 adapter has type == "phx"; users never set it directly. The inference lives in one place (parse_args_to_config), not duplicated across the config surface.

L1Manager.__init__: elif config.phx_l1_config is not None: — identical in structure to if config.gds_l1_config is not None:.

validate_storage_manager_config: checks phx_l1_config is not None for the mutual-exclusion rules, rather than re-scanning the L2 adapter list.

This way L1ManagerConfig no longer carries a reverse dependency on L2AdaptersConfig, L1Manager doesn't need to import any L2 symbols, and all three L1 tier selections (GDS / Device-DAX / Phoenix) follow the same pattern.

lmcache/v1/distributed/config.py
```diff
@@ -185,6 +185,24 @@ class GdsL1Config:
     """Allocation alignment; cuFile/hipFile and O_DIRECT require 4 KiB."""
 
 
+@dataclass
+class PhxL1Config:
```
maobaolong, line 189: @MonthFall This is an empty data class for now, do you mean you keep this for future extensible? If not, a bool variable can be easy and clear to Indicate whether use phoenix?

MonthFall: That's absolutely correct. I realize that in the future, we may need to configure the management of phx's GPU staging buffer (currently functioning as a gpuhbm L1).

More specifically, since the PHX L2 adapter currently handles the allocation of the GPU staging buffer, a single “replace memory obj” operation is required. However, if we can move the allocation of the GPU staging buffer to a more appropriate location (the PHX L1 memory manager), then the prefetch controller could just pass device MO rather than CPU MOs to the PHX L2 adapter when submiting a load task, eliminating the need for any subsequent “replace memory obj” operations.

This might prevent the L2 adapter from having to manage L1 memory.

To achieve this, we could have the PHX L1 Memory Manager override the `allocate` interface to support allocating different MOs based on the L2 adapter type (e.g., for fsnative or s3, the PHX L1MM allocates CPU MOs from dram; for phx, the PHX L1MM allocates GPU MOs from GPU staging buffer).

lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py
```diff
@@ -0,0 +1,1384 @@
...
+from dataclasses import dataclass
+from typing import Any, Optional
+import os
+import threading
+import time
+
+# First Party
+from lmcache import torch_device_type
+from lmcache.logging import init_logger
+from lmcache.native_storage_ops import Bitmap
```
maobaolong: This is not correct after recent refactor, 

```
tests/v1/distributed/test_phx_l2_adapter.py:27: in <module>
    from lmcache.v1.distributed.l2_adapters.phx_l2_adapter import (
lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py:27: in <module>
    from lmcache.native_storage_ops import Bitmap
E   ModuleNotFoundError: No module named 'lmcache.native_storage_ops'
```

maobaolong:
```suggestion
from lmcache.lmcache_native import Bitmap
```

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -487,8 +487,33 @@ def transfer_kv_per_object_group(
...
+    #      semantics for callers that still rely on the GDS-specific check and
+    #      to short-circuit the more expensive tensor inspection for GDS paths.
+    if (
+        _HAS_NATIVE_OBJECT_GROUP_TRANSFER
+        and not any(isinstance(mo, GDSMemoryObject) for mo in memory_objs)
+        and not any(
+            mo is not None
+            and mo.raw_tensor is not None
+            and mo.raw_tensor.device.type != "cpu"
+            for mo in memory_objs
```
ApostaC: can we have a small helper function like:
```
def _eligible_for_fast_path(obj: MemoryObj):
    """ returns true if this memory object supports fast path """
```
And then have the condition become `all(_eligible_for_fast_path(obj) for obj in memory_objs)`.

This will make it much simpler to read.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -487,8 +487,33 @@ def transfer_kv_per_object_group(
         This function expects the caller to stage the block ids (list[list[int]])
         into GPU tensors and pass them in as `block_ids_gpu`.
     """
-    if _HAS_NATIVE_OBJECT_GROUP_TRANSFER and not any(
-        isinstance(mo, GDSMemoryObject) for mo in memory_objs
+    # The fast path below (`_run_object_group_transfer_plan`) is a CPU-MO-only
```
ApostaC: nit: we don't use `MO` anywhere in the code. Please avoid inventing new terminologies. 
Also, we can drastically simplify the comment here. Just say the fast path is only for CPU-GPU transfer will be enough.

docs/design/v1/distributed/l2_adapters/phx_l2_adapter.md
ApostaC, line 1: The design doc is good. We also need to have a user-facing doc so that people will know how to use it.

Please see https://docs.lmcache.ai/mp/l2_storage/supported_storages.html and we can create a sub-page for Phoenix. The source code is under `docs/source`.

lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py
```diff
@@ -0,0 +1,1384 @@
...
+        dt = time.perf_counter() - self.t0
+        bw = total_bytes / dt / 1024 / 1024 if dt > 0 and total_bytes > 0 else 0.0
+        _perf_write(
+            f"STORE task={task_id} keys={n_keys} ok={n_ok} "
+            f"{total_bytes / 1024 / 1024:.1f}MB in {dt * 1000:.1f}ms "
+            f"({bw:.0f}MB/s) mode={mode}"
+        )
+
+
+class _LoadPerf:
```
ApostaC, line 138: Regarding the perf log module, LMCache itself has an infra that allows modules to emit "events" and then have an asynchronous consumer process the events and write to file or emit to prometheus/opentelemetry systems. 
You can take a look at the "mp_observability" module. 

No need to change anything in this PR. But you can do it in future PRs if you want.

lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py
```diff
@@ -0,0 +1,1384 @@
...
+
+
+def _perf_write(line: str) -> None:
+    """Write a line to the perf log file (thread-safe)."""
+    f = _lmc_perf_log_file
+    if f is None:
+        return
+    with _lmc_perf_log_lock:
+        f.write(line + "\n")
+        f.flush()
```
ApostaC: Every time, we only write a single line, which may include a lot of small file IOs on the hot path. 

Maybe we can consider implementing a "buffered write" to reduce the IOs for perf log.

lmcache/v1/distributed/l1_manager.py
```diff
@@ -230,6 +263,59 @@ def __init__(self, config: L1ManagerConfig):
                 lambda: _l1_usage_ratio_or_zero(L1Manager._gauge_target),
             )
 
+        # Register self in the module-level registry so L2 adapters
+        # constructed after this point can fetch the live L1Manager.
+        global _CURRENT
+        _CURRENT = weakref.ref(self)
+
+    @l1_mgr_synchronized
+    def replace_memory_obj(
```
ApostaC, line 272: This function is not a general L1 memory operation, and this may lead to extra maintenance complexity. We should not have it here.

To be clearer, the L1 manager only manages the state of each object (i.e., the cooresponding key, the read/write lock state). It should not have any visibility into the CPU or GPU device.

In the meantime, another **very important invariant** is that we do not want anyone else to be able to create/free memory objects in the system explicitly; otherwise, nobody will be able to maintain the lifecycle management logic and fix memory leak problems. This was a big issue in the old "in-process" mode. The caller should only use reserve_read/finish_read/reserve_write/finish_write.

ApostaC (changes requested):
Please take a look at my comments. The biggest problem is `replace_memory_obj`.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
```diff
@@ -487,8 +487,33 @@ def transfer_kv_per_object_group(
...
+    #      past guard 2. This check is strictly broader than guard 2 and would
+    #      suffice on its own; guard 2 is kept to preserve the original
+    #      semantics for callers that still rely on the GDS-specific check and
+    #      to short-circuit the more expensive tensor inspection for GDS paths.
+    if (
+        _HAS_NATIVE_OBJECT_GROUP_TRANSFER
+        and not any(isinstance(mo, GDSMemoryObject) for mo in memory_objs)
+        and not any(
+            mo is not None
+            and mo.raw_tensor is not None
```
ApostaC: side note for here: originally, the fast path is compatible with a `None` memory object. Please make sure that it's still compatible after the update.

lmcache/v1/distributed/l1_manager.py
MonthFall: This is a very insightful suggestion. We also discussed this issue with @maobaolong  [above](https://github.com/LMCache/LMCache/pull/4474#discussion_r3772657613) .
 Our consensus is: 
- First, the L2 adapter should not be allowed to manipulate the L1 manager; 
- second, the L2 adapter should not be allowed to allocate L1 memory (our current GPU staging buffer serves as the GPU L1). 

Therefore, the update we are currently pursuing is to leverage lmcache’s existing extensible L1 memory manager mechanism to implement the PHX L1 memory manager (PHX L1MM), just as the GDS L1MM and DEVDAX L1MM do.

The difference is that the PHX L1MM inherits rather than reimplements the L1MM, retaining DRAM L1 to maintain compatibility with other L2 adapters. For PHX path, the SSD remains L2, regarding the GPU DMA buffer as L1; for other adapters, they continue to treat DRAM as L1. DRAM L1 and the GPU DMA buffer L1 will coexist in parallel, and they be completely isolated from one another, no longer be mixed with “replace memory obj”.

Next, we will streamline the PHX L2 adapter so that it retains only the I/O logic. We will also move the management of the GPU DMA buffer L1 to the PHX L1MM and rewrite `_memory_manager.allocate`.

The caller—such as the prefetch controller—will automatically call `self._memory_manager.allocate` when calling `reserve_write` to allocate both the GPU L1(keys hit in PHX adapter) and DRAM L1 (keys hit in other L2 adapters like redi, s3).

Then, the prefetch controller can simply pass device memory objects rather than CPU memory objects to the PHX L2 adapter when submitting a load task, eliminating the need for any subsequent “replace memory obj” operations.

maobaolong: @MonthFall Sounds good. Regarding the size of this PR is over 2k, i guess the changes you motioned above would increase the PR size again, so I suggest to create another PR to implement the above changes first, after it merge, we can rebase this PR, WDYT?

And, we'd better to focus on building a more general L1 memory manager that isn't tied to Phoenix.

MonthFall: @maobaolong That sounds great. In my next commit, I’ll first address all the other minor issues (such as the user guide, comments, and readability improvements) to ensure that the current PR is in good shape. 

Then I’ll create a new PR to implement the design described above to resolve the “replace memory obj” issue. This will effectively keep the PR’s scope manageable and ensure maintainability. Thank you very much for your suggestion!

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
MonthFall: Thank you for your suggestion！ We have already update this in the latest commit. Helper func Done. Added **_eligible_for_fast_path** and the guard is now all(_eligible_for_fast_path(mo) for mo in memory_objs).

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
MonthFall: Simplify comment Simplified — the comment now just says the fast path is only for CPU-GPU transfer, and dropped the "MO"/"DMO" terminology.

lmcache/v1/multiprocess/modules/lmcache_driven_transfer.py
MonthFall: None handling Note: memory_objs can contain None (store path skips failed entries), so the helper takes **MemoryObj | None** and treats None as eligible — equivalent to the original not any(mo is not None and ...) semantics.

lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py
MonthFall: Thanks for the pointer to mp_observability. That event-based infra is definitely a cleaner fit than our ad-hoc perf log — decoupled emit/consume + async aggregation beats hand-rolled file logging. I'll plan to migrate to it in a follow-up PR.

lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py
MonthFall: This is a very, very good suggestion, we had previously overlooked the potential load that perf logs might place on I/O. 

We have switched to a 64 KiB block buffer and dropped the per-line flush, so the hot path now amortizes many small writes into one. Further migration to the mp_observability event infra is planned for a follow-up PR.

docs/design/v1/distributed/l2_adapters/phx_l2_adapter.md
MonthFall: Thank you! I've added a user-facing page for the Phoenix adapter at docs/source/mp/l2_storage/phx.rst, linked from the supported storages page (under File & Block). 

It covers prerequisites (OS/kernel/CUDA/build tools), installation of the Phoenix kernel module and phxcache, the --l2-adapter config with all parameters, the automatic L1 integration, and the POSIX fallback behavior.

lmcache/v1/memory_allocators/phx_device_memory_allocator.py
MonthFall: Hi, I've added a user-facing page for the Phoenix adapter at docs/source/mp/l2_storage/phx.rst, linked from the supported storages page (under File & Block). 
It covers prerequisites (OS/kernel/CUDA/build tools), installation of the Phoenix kernel module and phxcache, the --l2-adapter config with all parameters, the automatic L1 integration, and the POSIX fallback behavior.

lmcache/v1/distributed/l2_adapters/phx_l2_adapter.py
MonthFall: We have resolved this issue , thanks!

maobaolong:
@MonthFall Would you like to fetch and rebase to dev branch? I guess the failed amd issue has been resolved by another merged PR.

PR #4477, opened 2026-08-10
[good-first-issue] storage: convert f-string log calls in infinistore_adapter.py to %-format (#4402)

m0g3r:
**What this PR does / why we need it**:

Closes #4402
Refs #3372

Convert the `logger.info(f"...")` call in
`lmcache/v1/storage_backend/connector/infinistore_adapter.py` to lazy `%-format`
so the message is interpolated only when the log record is actually emitted
(ruff G004). No behavior change; format style only.

```diff
-        logger.info(f"Creating Infinistore connector for URL: {context.url}")
+        logger.info("Creating Infinistore connector for URL: %s", context.url)
```

**Special notes for your reviewers**:

- Same lazy-logging pattern as #4319 for `redis_adapter.py`; this is the only
  `logger.*` call in the file, so it completes the conversion here.
- `ruff check --select G004` flags the original line and passes after the change.
  Note `G` is currently commented out in the `[tool.ruff.lint] select` list in
  `pyproject.toml`, so this is not yet CI-enforced — happy to fold in enabling
  `G` repo-wide as a follow-up if maintainers want it.
- `ruff format --check`, `ruff check`, `isort --check-only` and `codespell` are
  all clean on the touched file.
- The two remaining f-strings in the file are `raise ValueError(...)` messages,
  which G004 does not cover; left unchanged.
- Commit includes DCO sign-off (`-s`).

@linear3735 commented `/claim` on #4402 on 2026-08-03 but the issue was never
assigned and no PR followed, so I picked it up per the 7-day guidance in #3372.
Happy to close this if they are still working on it.

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

_Disclosure: written with AI assistance; the lint commands above were run locally
against `dev` @ 9fc6e1af._

PR #4331, opened 2026-07-29
[good-first-issue] storage: use lazy logging in audit_connector.py (#4305)

HermanZeng:
**What this PR does / why we need it**:

Closes #4305
Refs #3372

This PR converts all 15 f-string `logger.*` calls in `lmcache/v1/storage_backend/connector/audit_connector.py` to lazy `%`-style logging. It also updates the audit connector tests to inspect `LogRecord.getMessage()`, so they continue to validate the rendered audit messages after the format string and arguments are stored separately.

The final audit-message text remains unchanged for LMCache's current connector and key types, including six-decimal cost formatting, eight-character checksum prefixes, spacing, and the full-width separator in the initialization message. The change does not alter connector I/O, checksum behavior, return values, exception propagation, logger hierarchy, or timing boundaries.

An earlier PR, #4250, addressed the same cleanup and received review, but it was closed when the maintainers chose to reserve these small onboarding tasks for first-time contributors. As this is my first LMCache contribution, I am respectfully picking up that still-unmerged work. I rebased the change on the current `dev` branch and independently reviewed and tested the resulting diff.

**Special notes for your reviewers**:

- The scope is limited to the audit connector and its existing test file.
- Raw `LogRecord.msg` and `LogRecord.args` now contain the lazy template and arguments as expected; assertions use `getMessage()` to check the final text without mutating the record.
- The existing `str(e)` expressions are intentionally retained, preserving the previous exception-stringification timing while deferring the surrounding log-message interpolation.
- A differential runtime check exercised every converted call site against the version on `dev`, including async and sync success/failure, PUT failure, GET miss/mismatch/failure, checksum handling, excluded commands, and `NotAudit`; the rendered messages matched.
- AI-assisted tooling was used during implementation and adversarial review. I personally reviewed the final diff and ran every verification command listed below.

Verification completed locally:

```text
pytest -xvs tests/v1/storage_backend/test_audit_connector.py
# 15 passed

pytest --confcutdir=tests/v1 -q tests/v1/test_impl_completeness.py
# 4 passed

ruff check --select G004 lmcache/v1/storage_backend/connector/audit_connector.py
# passed

ruff check lmcache/v1/storage_backend/connector/audit_connector.py tests/v1/storage_backend/test_audit_connector.py
# passed

ruff format --check lmcache/v1/storage_backend/connector/audit_connector.py tests/v1/storage_backend/test_audit_connector.py
# passed

pre-commit run --all-files
# passed
```

[...]

PR #4319, opened 2026-07-29
[good-first-issue] storage: convert f-string log calls in redis_adapter.py to %-format (#4311)

vedjaw:
**What this PR does / why we need it**:
Closes #4311
Refs #3372

Convert `logger.*(f\"...\")` calls in `lmcache/v1/storage_backend/connector/redis_adapter.py` to lazy `%-format` so arguments are interpolated only when the log record is emitted (ruff G004). No behavior change.

**Special notes for your reviewers**:
- Follow-up to the same lazy-logging pattern as #4142.
- Ran `ruff check --select G004` on the touched file — all green.
- Ran `ruff format` / `ruff check` on the touched file — all green.
- Commit includes DCO sign-off (`-s`).

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

Made with [Cursor](https://cursor.com)

PR #4478, opened 2026-08-10
[good-first-issue] health_monitor: convert f-string log calls in base.py to %-format

m0g3r:

Refs #3372

All 13 `logger.*` calls in `lmcache/v1/health_monitor/base.py` used f-strings, so each
message was formatted even when the record was filtered out. Convert them to lazy
`%-format` (ruff G004). No behavior change; format style only.

Continues the same cleanup as #4319 (`redis_adapter.py`), #4331 (`audit_connector.py`)
and #4477 (`infinistore_adapter.py`).

- `ruff check --select G004` on this file: **13 violations before, 0 after**.
- **Message text is unchanged.** Since this hand-edits 13 message strings, I verified
  it mechanically rather than by eye: I parsed the pre- and post-change file, replaced
  every f-string `{expr}` and every `%s`/`%d` placeholder with a sentinel, and diffed
  the resulting literals. All 13 match exactly.
- Placeholder count equals argument count for all 13 calls (checked via AST).
- Two calls needed care: one mixed an f-string with an adjacent plain string, and one
  carried an `f` prefix with no placeholders at all — both are now plain strings.
- `ruff format --check`, `ruff check`, `isort --check-only`, `codespell`: all clean.
- No test in the repo asserts on `record.msg`, so this does not require the test
  adjustment that #4331 needed.
- Commit includes DCO sign-off (`-s`).

There are ~400 more G004 hits elsewhere in `lmcache/v1/` (largest: `eic_connector.py`,
`bigtable_connector.py`, `sagemaker_hyperpod_connector.py`, `hf3fs_connector.py`).
Happy to keep going file-by-file if this shape is what you want — I'd rather confirm
the approach on one reviewable PR than open a pile of them at once.

**If applicable**:

_Disclosure: written with AI assistance; all commands above were run locally against
`dev` @ 9fc6e1af._

### DIFF
--- a/lmcache/v1/health_monitor/base.py
+++ b/lmcache/v1/health_monitor/base.py
@@ -243,15 +243,16 @@ class HealthMonitor(PeriodicThread):
             try:
                 instances = cls.create(self._manager)
             except Exception as e:
-                logger.warning(f"Failed to create health check {cls.__name__}: {e}")
+                logger.warning("Failed to create health check %s: %s", cls.__name__, e)
                 continue
             for instance in instances:
                 self._health_checks.append(instance)
                 # Initialize previous status as healthy
                 self._previous_check_status[instance.name()] = True
                 logger.info(
-                    f"Registered health check: {instance.name()} "
-                    f"with fallback_policy: {instance.fallback_policy}"
+                    "Registered health check: %s with fallback_policy: %s",
+                    instance.name(),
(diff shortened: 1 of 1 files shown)


chunxiaozheng (approved):
LGTM

deng451e (approved):
LGTM

PR #4479, opened 2026-08-10
[good-first-issue] storage: convert f-string log calls in bigtable_connector.py to %-format

m0g3r:

Refs #3372

All 43 `logger.*` calls in `lmcache/v1/storage_backend/connector/bigtable_connector.py`
used f-strings, so each message was formatted even when the record was filtered out.
Convert them to lazy `%-format` (ruff G004). No behavior change; format style only.

Continues #4319 (`redis_adapter.py`), #4331 (`audit_connector.py`), #4477
(`infinistore_adapter.py`) and #4478 (`health_monitor/base.py`).

- `ruff check --select G004` on this file: **43 violations before, 0 after**.
- **This conversion was scripted, not hand-typed** — 43 call sites is past the point
  where manual edits are trustworthy. The transform reads the AST, rebuilds each
  template, and splices it back using the original source spans, so the argument
  expressions keep the author's exact text (e.g. `1e6` is not rewritten to
  `1000000.0`).
- Verified mechanically, both before and after `ruff format`:
  - **Message text unchanged** — every f-string `{expr}` and every `%s`/`%.2f`
    placeholder replaced with a sentinel, then the literals diffed: all 43 match.
  - **Placeholder count == argument count** for all 43 calls.
  - `exc_info=True` on the `_batched_put_internal` flush error is preserved.
- The three `{value:.2f}` format specs became `%.2f`, so the rendered numbers are
  identical.
- Where merging the f-string fragments pushed a template past 88 columns, the string
  is split across adjacent literals (concatenation, so the value is unchanged) rather
  than truncated or reworded.
- `ruff check`, `ruff format --check`, `isort --check-only`, `codespell`: all clean.
- No test asserts on `record.msg`, so this needs no test adjustment.
- Commit includes DCO sign-off (`-s`).

I can't run `tests/v1/` locally (needs the compiled `lmcache.c_ops` extension), so
this leans on CI for the runtime check — the verification above is static.

**If applicable**:

_Disclosure: written with AI assistance; all commands above were run locally against
`dev` @ 9fc6e1af._

### DIFF
--- a/lmcache/v1/storage_backend/connector/bigtable_connector.py
+++ b/lmcache/v1/storage_backend/connector/bigtable_connector.py
@@ -87,8 +87,12 @@ def _extract_shards_from_row(row: Any, family_name: str) -> dict[str, bytes]:
             shards[qual] = val
         else:
             logger.warning(
-                f"Discarding stale shard {qual} with timestamp {ts} "
-                f"(max timestamp is {max_ts}, diff is {(max_ts - ts) / 1e6:.2f}s)"
+                "Discarding stale shard %s with timestamp %s (max timestamp is %s, "
+                "diff is %.2fs)",
+                qual,
+                ts,
+                max_ts,
+                (max_ts - ts) / 1e6,
             )
 
     return shards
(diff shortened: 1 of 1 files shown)


deng451e (approved):
LGTM

maobaolong (approved):
lgtm

Issue #4029, opened 2026-07-07
[ROCm] stable pypi and nightly wheels of ROCm LMCache


functionstackx:
Hi @sammshen  @andyluo7 

is there any active workstreams for ROCm pypi lmcache wheels and nightly wheels of ROCm LMCache

Currently on the LMCache docs https://docs.lmcache.ai/getting_started/installation.html i am just seeing CUDA LMCache for stable release pypi wheels & nightly wheels.

I do see from installs for building from source tho and see @seungrok has it working tho ideally there can be an prebuilt ROCm wheel to have an first class ROCm experience & to speed up on installation
https://github.com/SemiAnalysisAI/InferenceX/blob/58839b7f782d427ad7f7f98dab97f251a3b96a0f/benchmarks/single_node/agentic/dsv4_fp4_mi355x_vllm.sh#L283-L290

<img width="746" height="414" alt="Image" src="https://github.com/user-attachments/assets/6b163a50-2040-451d-b1e2-3ab26b9f93e8" />

<img width="818" height="465" alt="Image" src="https://github.com/user-attachments/assets/fba28fda-5ce9-4c88-9547-73d2f30d2396" />

<img width="761" height="613" alt="Image" src="https://github.com/user-attachments/assets/4d6cd202-4968-499b-a360-998047bbaef7" />

fengxiaohu:
# Add ROCm prebuilt wheels for stable and nightly LMCache releases

## Summary

ROCm users can currently build LMCache from source with `BUILD_WITH_HIP=1`, but there is no first-class prebuilt ROCm wheel experience 

Related: #4029

## Motivation

The current installation docs provide stable and nightly wheel instructions for CUDA, while ROCm users must build from source. This makes installation slower and less reliable for AMD/ROCm environments.

Providing ROCm wheels would improve the first-time install experience and make ROCm a first-class supported deployment path.

## Proposed Scope

- Build ROCm 7.0 wheels with:
  - `BUILD_WITH_HIP=1`
  - `CXX=hipcc`
  - `TORCH_DONT_CHECK_COMPILER_ABI=1`
  - `PYTORCH_ROCM_ARCH="gfx942,gfx950"`
  - `PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/rocm7.0`

- Publish stable ROCm wheels to a dedicated GitHub Release, for example:
  - `vX.Y.Z-rocm70`

[...]

PR #4481, opened 2026-08-10
ci(rocm): tag the ROCm wheel with a +rocm7.2 local version

andyluo7:
## Why

`build_rocm_artifacts.yml` states in its own header that the ROCm wheel "carries a `+rocm7.2` local version". The build never applied one. The published asset on `v0.5.3-rocm` is:

```
lmcache-0.5.3-cp312-cp312-manylinux_2_35_x86_64.whl
```

versus the CUDA wheel on PyPI:

```
lmcache-0.5.3-cp312-cp312-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
```

Same distribution name, same version, same interpreter tag, differing only in the manylinux minor. Nothing in the artifact identifies it as a ROCm build, so an installed wheel can't be told apart after the fact and the ROCm build can't be pinned explicitly. My fault — I wrote that flow in #4273.

This applies the local version the comment already promises, matching how PyTorch tags its own ROCm builds (`torch-2.11.0+rocm7.2`).

**What this does not do:** it doesn't change which wheel a resolver prefers. `0.5.3+rocm7.2` sorts *above* `0.5.3`, so where both are visible the ROCm wheel already won. The win is that `pip show lmcache` reports the build in use, and that the ROCm wheel can be pinned as `lmcache==0.5.3+rocm7.2`.

Existing instructions keep working — a bare `==0.5.3` matches a local version under PEP 440, so nothing that pins the plain version breaks.

## What

- `build_rocm_artifacts.yml`: add `ROCM_LOCAL_VERSION: rocm7.2` beside the existing ABI pins and apply it to the wheel version.
- Expose the resolved version as a `workflow_call` output, and consume it in the release notes in `publish.yml` rather than repeating the literal. A comment drifting from behaviour is what hid this in the first place; copying the string into a second workflow would set up the same failure.
- `installation.rst`: pin explicitly and note what the local version buys.

## Testing

Ran the real build script in `rocm/dev-ubuntu-22.04:7.2.3-complete` on MI300X, same command CI issues:

```
dist/lmcache-0.5.3+rocm7.2-cp312-cp312-manylinux_2_35_x86_64.whl
```

Installed into the upstream image (`vllm/vllm-openai-rocm:v0.26.0`, torch 2.11.0, ROCm 7.2.3, cp312):

| check | result |
|---|---|
| `pip install ... --no-deps` | clean |
| `pip show lmcache` | `Version: 0.5.3+rocm7.2` |
| `import lmcache` | OK |
| `llvm-objdump --offloading` on `c_ops*.so` | `gfx942`, `gfx950` |

Arch code objects are unchanged, so the version change doesn't disturb the fat binary.

Version resolution checked both ways: non-tag builds produce `0.5.3+rocm7.2`, tag builds `0.5.4+rocm7.2`.

Follow-up: a rolling `nightly-rocm` release (the actual parity gap raised in #4029) is next, and wants this landed first so nightly artifacts aren't ambiguous too.

### DIFF
--- a/.github/workflows/build_rocm_artifacts.yml
+++ b/.github/workflows/build_rocm_artifacts.yml
@@ -15,6 +15,14 @@ name: Build ROCm Artifacts
 
 on:
     workflow_call:
+        outputs:
+            rocm_version:
+                description: >-
+                    Full wheel version including the ROCm local segment
+                    (e.g. 0.5.3+rocm7.2). Consume this rather than
+                    reconstructing the version, so the local segment has a
+                    single source of truth.
+                value: ${{ jobs.build-rocm-artifacts.outputs.rocm_version }}
 
 env:
     LC_ALL: en_US.UTF-8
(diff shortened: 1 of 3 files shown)


deng451e (approved):
LGTM

ApostaC (approved):
LGTM!

PR #4482, opened 2026-08-10
ci(rocm): publish a rolling nightly-rocm wheel release

andyluo7:
## Why

`nightly_build.yml` publishes rolling `nightly` (cu13.0) and `nightly-cu129` releases. There is no ROCm equivalent, so ROCm users can't test against `dev` without building from source. That's the parity gap raised in #4029 — and unlike PyPI publication, this one is actually fixable.

Rebased onto `dev` now that #4481 has landed, so ROCm nightlies inherit the `+rocm7.2` local version and are distinguishable from the CUDA nightlies on sight.

## What

- `nightly-rocm-wheel` job calling the existing `build_rocm_artifacts.yml` reusable workflow, so nightly and release wheels are built by exactly the same path.
- `publish-nightly-rocm` job creating a rolling `nightly-rocm` release, mirroring the `nightly-cu129` pattern.
- Docs: a ROCm tab under Nightly.

### The versioning wrinkle

On non-tag refs the reusable workflow pretends to be the newest release tag. A nightly built from `dev` would therefore have been versioned `0.5.3` — colliding with the published release, and sorting as though it *were* that release.

So this adds an opt-in `dev_version` input that reproduces setuptools_scm's guess-next-dev: bump the patch of the newest tag, append the commit count. That's what the CUDA nightlies already get from cibuildwheel, so ROCm nightlies end up versioned consistently with them.

The input defaults to `false`, so the release pipeline is untouched.

## Testing

Version resolution, extracted programmatically from the workflow rather than retyped:

| ref | `dev_version` | result |
|---|---|---|
| tag `v0.5.4` | false | `0.5.4+rocm7.2` |
| `dev` | **true** | `0.5.4.dev43+rocm7.2` |
| `dev` | false | `0.5.3+rocm7.2` (unchanged) |

Cross-checked against the CUDA nightly published today, `lmcache-0.5.4.dev40-cp310-...whl` — same guess-next-dev computation, the difference being commits landed on `dev` since that build ran.

PEP 440 semantics confirmed: `0.5.4.dev43+rocm7.2` is a pre-release (so `--pre` is required, matching the CUDA nightly instructions), sorts above `0.5.3`, and below the eventual `0.5.4`.

The wheel build itself is unchanged from #4481, where I verified on MI300X that the produced wheel installs into `vllm/vllm-openai-rocm:v0.26.0` and retains gfx942 + gfx950 code objects.

## Note for reviewers

The cu13.0 and cu12.9 nightly wheels currently have *identical* filenames (`lmcache-0.5.4.dev40-cp310-cp310-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl` in both releases), so they're indistinguishable once downloaded — the same class of problem #4481 fixed for ROCm. Out of scope here; happy to follow up separately, as discussed on #4446.

### DIFF
--- a/docs/source/getting_started/installation.rst
+++ b/docs/source/getting_started/installation.rst
@@ -124,6 +124,23 @@ Install LMCache
                                 --find-links https://github.com/LMCache/LMCache/releases/expanded_assets/nightly-cu129 \
                                 --index-strategy unsafe-best-match
 
+                    .. tab-item:: ROCm 7.2
+
+                        Run inside an upstream vLLM ROCm container so torch and the ROCm
+                        runtime are already present, then install with ``--no-deps``:
+
+                        .. code-block:: bash
+
+                            docker run -it --device /dev/kfd --device /dev/dri \
+                                --group-add video --security-opt seccomp=unconfined \
+                                --entrypoint bash vllm/vllm-openai-rocm:v0.26.0
+
+                            pip install lmcache --pre --no-deps \
+                                --find-links https://github.com/LMCache/LMCache/releases/expanded_assets/nightly-rocm
+
+                        Nightly ROCm wheels are versioned like the CUDA nightlies with the
+                        ROCm local segment appended, e.g. ``0.5.4.dev15+rocm7.2``.
+
             .. tab-item:: From Source
 
                 ``--no-build-isolation`` ensures the kernels are compiled against the same torch
--- a/docs/source/getting_started/installation.rst
+++ b/docs/source/getting_started/installation.rst
@@ -135,11 +135,22 @@ Install LMCache
                                 --group-add video --security-opt seccomp=unconfined \
                                 --entrypoint bash vllm/vllm-openai-rocm:v0.26.0
 
-                            pip install lmcache --pre --no-deps \
+                            pip install lmcache --pre --no-deps --no-index \
                                 --find-links https://github.com/LMCache/LMCache/releases/expanded_assets/nightly-rocm
 
                         Nightly ROCm wheels are versioned like the CUDA nightlies with the
                         ROCm local segment appended, e.g. ``0.5.4.dev15+rocm7.2``.
+
+                        .. note::
+
+                            ``--no-index`` is required here. ``--find-links`` only *adds* a
+                            source, so without it pip also considers PyPI — and under PEP 440 a
+                            pre-release such as ``0.5.4rc4`` outranks ``0.5.4.dev15+rocm7.2``, so
+                            ``--pre`` would install the CUDA wheel instead. The stable tab does
+                            not need it because ``lmcache==${VERSION}+rocm7.2`` is an exact pin
+                            that only the ROCm release can satisfy. ``--no-deps`` is what makes
+                            ``--no-index`` safe here: torch and the ROCm runtime come from the
+                            container, so nothing else needs resolving.
 
             .. tab-item:: From Source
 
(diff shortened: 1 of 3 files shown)


ApostaC:
@deng451e can you help review this? Thanks!

ApostaC (commented):
LGTM in general. Thanks for the contribution.

One thing that needs to be fixed:
In our current ROCm installation documentation, the `VERSION` is tagged as `v0.5.3`. But what we should do something like `VERSION=v0.5.3+rocm7.2`, otherwise it will still resolve to CUDA wheels.

Please see the screenshot below:
Image

docs/source/getting_started/installation.rst
```diff
@@ -124,6 +124,23 @@ Install LMCache
...
+                        runtime are already present, then install with ``--no-deps``:
+
+                        .. code-block:: bash
+
+                            docker run -it --device /dev/kfd --device /dev/dri \
+                                --group-add video --security-opt seccomp=unconfined \
+                                --entrypoint bash vllm/vllm-openai-rocm:v0.26.0
+
+                            pip install lmcache --pre --no-deps \
+                                --find-links https://github.com/LMCache/LMCache/releases/expanded_assets/nightly-rocm
```
deng451e: `--find-links` only adds a source, so pip install will likely resolve to the PyPI CUDA rc wheel. Adding `--no-index` pins it to the ROCm wheel from the GitHub Release.

```diff
-pip install lmcache --pre --no-deps \
+pip install lmcache --pre --no-deps --no-index \
     --find-links .../releases/expanded_assets/nightly-rocm
```

andyluo7:
Good catch @deng451e — fixed in 2e8cc71b.

It's worse than "likely": PyPI currently carries `0.5.4rc4`, and under PEP 440 a
pre-release outranks a dev release, so `0.5.4rc4 > 0.5.4.dev43+rocm7.2`. With `--pre` and
PyPI still in the index, pip would take the CUDA rc every time. Added `--no-index`, which is
safe here only because `--no-deps` means torch and the ROCm runtime come from the container
and nothing else needs resolving.

@ApostaC on the stable tab — I think that one is already handled, it just doesn't look the
way you described. On `dev` after #4481 it reads:

```bash
VERSION=0.5.3  # replace with target release
pip install lmcache==${VERSION}+rocm7.2 --no-deps \
    --find-links .../releases/expanded_assets/v${VERSION}-rocm
```

so the `+rocm7.2` is on the pin rather than in the `VERSION` variable. Being an exact local
version, PyPI has nothing that can satisfy it. Your screenshot is probably the published
site, which hasn't rebuilt since #4481 merged this morning. Happy to restructure it to
`VERSION=0.5.3+rocm7.2` if you'd rather it read that way.

Worth noting the CUDA nightly tabs have the same shape without `--no-index`, but they can't
take it — no `--no-deps` there, so they need PyPI to resolve dependencies. Different problem,
not touching it here.

deng451e (approved):
lgtm!

andyluo7:
Thanks @deng451e.

@ApostaC — this is only waiting on your stable-tab point. My read is that it's already
handled on `dev` after #4481: the command is `pip install lmcache==${VERSION}+rocm7.2`, and
since no such version exists on PyPI the exact pin can only be satisfied by the ROCm release.
Your screenshot looked like the published docs site, which hadn't rebuilt at the time.

Happy to restructure it to `VERSION=0.5.3+rocm7.2` if you'd rather it read that way — say
which and I'll push it. Otherwise this is ready.

ApostaC (approved):

Issue #3376, opened 2026-05-23
S3 backend cannot upload into minio that I am sure minio works fine


gaowayne:
**Label**
Please label your issue so that it can easily be easily categorized under [LMCache Onboarding](https://github.com/LMCache/LMCache/issues/1882)

**Summary**
S3 backend failed to upload to minio

**Details**
```
(EngineCore pid=79465) [2026-05-23 15:24:54,249] LMCache ERROR: Failed to upload deepseek-ai/DeepSeek-R1-Distill-Qwen-32B@1@0@-41257af1dd85c366@bfloat16 to S3: AWS_ERROR_S3_INVALID_RESPONSE_STATUS: Invalid response status from request (s3_connector.py:578:lmcache.v1.storage_backend.connector.s3_connector)
(EngineCore pid=79465)   File "/mnt/tlc/wayne/vllmlmcache/lib/python3.12/site-packages/awscrt/s3.py", line 857, in _on_finish
(EngineCore pid=79465)     self._on_done_cb((EngineCore pid=79465)   File "/mnt/tlc/wayne/vllmlmcache/LMCache/lmcache/v1/storage_backend/connector/s3_connector.py", line 525, in on_done
(EngineCore pid=79465)     raise RuntimeError(f"Upload failed in S3Connector: {done}")
(EngineCore pid=79465) RuntimeError: Upload failed in S3Connector: {'err': S3ResponseError(name='AWS_ERROR_S3_INVALID_RESPONSE_STATUS', message='Invalid response statusfrom request', code=14343), 'status': 400}
(EngineCore pid=79465) Exception ignored in: <awscrt.s3._S3RequestCore object at 0x726b3b742bd0>(EngineCore pid=79465) Traceback (most recent call last):
(EngineCore pid=79465)   File "/mnt/tlc/wayne/vllmlmcache/lib/python3.12/site-packages/awscrt/s3.py", line 857, in _on_finish
(EngineCore pid=79465) [2026-05-23 15:24:54,250] LMCache ERROR: Failed to upload deepseek-ai/DeepSeek-R1-Distill-Qwen-32B@1@0@-59d48422a2544c33@bfloat16 to S3: AWS_ERROR_S3_INVALID_RESPONSE_STATUS: Invalid response status from request (s3_connector.py:578:lmcache.v1.storage_backend.connector.s3_connector)(EngineCore pid=79465)     self._on_done_cb(
[...]

PR #2624, opened 2026-02-20
[Misc] S3 path style access

S4AR-inc:
**What this PR does / why we need it**:
This PR allows configuring path style bucket access for s3 connector.
Fixes #2621 Fixes #3376

**If applicable**:

- [X] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #4473, opened 2026-08-10
refactor: switch KV format enums/helpers to lmcache_native

maobaolong:
This PR switches internal callsites of EngineKVFormat, TransferDirection, is_kv_list, is_layer_list, is_cross_layer, and is_mla from c_ops/ops_types to lmcache_native.

It also removes the redundant definitions from lmcache/v1/platform/ops_types.py, updates torch_ops to import from lmcache_native, and adjusts tests and CI so lmcache_native is built in non-CUDA test jobs as well.

Validation:
- So files after this refactor
```
-rwxr-xr-x 1 root    root    37739136 Aug 11 06:38 c_ops.cpython-312-x86_64-linux-gnu.so*
-rwxr-xr-x 1 root    root     7367464 Aug 11 06:38 lmcache_fs.cpython-312-x86_64-linux-gnu.so*
-rwxr-xr-x 1 root    root     7828520 Aug 11 06:38 lmcache_native.cpython-312-x86_64-linux-gnu.so*
-rwxr-xr-x 1 root    root     7213696 Aug 11 06:38 lmcache_redis.cpython-312-x86_64-linux-gnu.so
```

PR #4484, opened 2026-08-10
[Bug][S3] propagate connector errors off CRT threads and trip breaker on HTTP failures

### LINKED ISSUE #4483: [Bug][S3] Connector errors raised on CRT threads are swallowed; circuit breaker ignores HTTP failures
**Describe the bug**

`lmcache/v1/storage_backend/connector/s3_connector.py` has two independent
error-handling defects. Both reproduce against any failing remote backend, and
both are visible in the log attached to #3376.

**Defect 1 — the PUT and GET completion callbacks raise on aws-crt threads**

`_s3_upload.on_done` and `_s3_download.on_done` raise a `RuntimeError` when the
request completes with an error or a non-success status. Those callbacks run on
aws-crt threads, as the `NOTE(Jiayi)` comment above `_s3_download.on_done`
already points out, and `awscrt.s3._S3RequestCore._on_finish` resolves
`finished_future` *before* it invokes `on_done`. The raise therefore cannot
reach the coroutine awaiting the future. It is handed to Python's unraisable
hook, which prints:

```
Exception ignored in: <awscrt.s3._S3RequestCore object at 0x...>
```

Two consequences:

* When aws-crt does report an error code, the future rejects and `_put` / `get`
  do enter their `except` branch — but with the crt error, while the message
  that carried the HTTP status is lost as untraceable noise. This is the
  "hundreds of identical `Exception ignored in:` lines" in #3376.
* When aws-crt reports no error code but the status is not a success — exactly
  the case these callbacks test for — the failure is lost completely. The
  `await` returns normally, so `_put` records the chunk in `object_size_cache`
  as if the upload had landed, and `get` hands back a `MemoryObj` that was
  never written. A later lookup then treats the key as a hit that S3 cannot
  serve.

**The file already contains the correct pattern.** `_get_object_size.on_done`
and `_get_object_size_async.on_done` only record into a `got` dict and let the
caller inspect it once the future has resolved. The HEAD path is right; the PUT
and GET paths are not.

**Defect 2 — the circuit breaker cannot trip on HTTP errors**

`_update_connection_failures(error_msg: str) -> bool` decides whether a failure
counts toward the breaker by substring-matching the message against
`CONNECTION_REFUSED`, `SOCKET`, `DNS` and `TIMEOUT`.

A persistent HTTP 400/401/403 — that is, any misconfiguration — matches none of
them. `connection_failures` never increments, `connection_disabled` never flips,
and LMCache reissues a guaranteed-doomed PUT for every single chunk,
indefinitely. That is why the log in #3376 contains hundreds of identical errors
within seconds. The breaker exists precisely to stop this, and it is blind to
the most common cause.

**To reproduce**

No object store, credentials, GPU or network are needed — stub the
`s3.S3Request` boundary and drive `on_done` directly:

1. Replace `awscrt.s3.S3Request` with a stub that resolves `finished_future`
   and *then* calls `on_done(error=None, status_code=400)`, mirroring
   `_S3RequestCore._on_finish`.
2. `await connector.put(key, memory_obj)`.
3. The `RuntimeError` raised inside `on_done` never reaches the coroutine;
   `connector.object_size_cache` contains the key as though the upload
   succeeded, and `connector.connection_failures` is still `0`.

The same stub with `status_code=403` on a GET makes `get()` return a
`MemoryObj` that was never filled.

**Expected behavior**

* A PUT or GET that completes with an error or a non-success status surfaces on
  the asyncio side, where `_put` / `get` / `batched_get` already have `try` /
  `except` blocks that log and feed the breaker.
* A repeated, non-retryable HTTP status trips the breaker, so LMCache stops
  hammering a backend that is guaranteed to reject every chunk.

**Additional context**

This is independent of the bucket-addressing question that caused the 400 in
#3376; that is being handled separately in #2624. These two defects are about
what the connector does with *any* failure, whatever produced it.

I have a fix with unit tests and will open a PR referencing this issue.

aryangorde8:
Fixes #4483

Two independent error-handling defects in `S3Connector`. Both reproduce against
any failing remote backend; they are independent of the bucket-addressing
question in #3376 / #2624.

*Errors raised on aws-crt threads never reached the coroutine.*
`_s3_upload.on_done` and `_s3_download.on_done` raised a `RuntimeError` on
failure, but `awscrt.s3._S3RequestCore._on_finish` resolves `finished_future`
before it invokes `on_done`, and does so on an aws-crt thread. The raise was
swallowed by the unraisable hook (`Exception ignored in: <awscrt.s3._S3RequestCore object at 0x...>`).
Where aws-crt set an error code the await still failed, but with the crt error
rather than the status that explained it; where aws-crt set no error code and
only the HTTP status was bad — the case these callbacks explicitly tested for —
the failure was lost entirely: `_put` recorded the chunk in `object_size_cache`
as if the upload had landed, and `get` returned a `MemoryObj` that was never
written.

Both callbacks now only record into a `done` dict, the way
`_get_object_size.on_done` already does, and `_s3_upload` / `_s3_download`
return that dict alongside the request. `_put`, `get` and `batched_get`
inspect it after awaiting the future, inside the `try` / `except` blocks that
already log and feed the circuit breaker. The GET success condition
`(status_code in (200, 206)) or (status_code is None)` is unchanged, and so is
the PUT condition `status_code in (200, 201)`.

*The circuit breaker could not trip on HTTP errors.*
`_update_connection_failures` only matched `CONNECTION_REFUSED`, `SOCKET`,
`DNS` and `TIMEOUT` in the error message, so a persistent 400/401/403 never
incremented `connection_failures`, `connection_disabled` never flipped, and
LMCache reissued a doomed request for every chunk indefinitely — the hundreds
of identical errors within seconds seen in #3376. It now takes an optional
`status_code` and counts the non-retryable ones, with a distinct log message so
a breaker trip caused by an HTTP status reads differently from one caused by a
connection error.

Two things I would like a decision on, both deliberately left as implemented
rather than open in the code:

1. **Which statuses count as non-retryable.** The initial set is
   `NON_RETRYABLE_STATUS_CODES = frozenset({400, 401, 403})` — the statuses
   that mean "this request is refused for what it is", so every subsequent
   chunk hits the same wall. **404 is deliberately excluded**, because
   `_get_object_size` treats it as an expected cache miss; there is a
   regression test for that. Whether 405, 409 or 5xx belong here is a
   judgement call I am happy to change.

2. **The attribute names.** `connection_failures` and `connection_disabled`
   now also count HTTP rejections, so their names are slightly narrower than
   their meaning. I kept them as they are to keep the diff small and reviewable
   — renaming them is a separate, mechanical change if you want it.

Smaller notes:

* `_update_connection_failures` still returns `is_connection_error`, so the
  existing callers keep logging non-connection failures exactly as before. The
  docstring spells out that a `False` return no longer means "ignored".
* `_failed_status_code` prefers the `status_code` attribute that aws-crt puts
  on `S3ResponseError` and falls back to what the callback recorded. The
  callback runs after the future resolves, so on a crt-reported failure the
  recorded status can race the awaiting coroutine; reading it off the exception
  removes that race for the path that matters. (The same race is latent in the
  existing HEAD path; I have not touched it here.)
* `batched_get` was updated too — its `asyncio.gather(..., return_exceptions=True)`
  loop now sees a bad status on a future that resolved successfully.
* The new tests stub the `s3.S3Request` boundary only; the stub reproduces
  aws-crt's ordering (future first, then `on_done`, on another thread). No
  object store, credentials, GPU or network involved. All 12 fail-to-pass and
  regression cases run in a few seconds on CPU.
* Unrelated to this PR, and left alone on purpose: the tests drain the job
  executor directly instead of calling `S3Connector.close()`, because `close()`
  awaits `AsyncPQExecutor.shutdown`, which is synchronous and blocks on a
  coroutine scheduled on the very loop it is awaited on. Happy to file that
  separately.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/s3_connector.py
+++ b/lmcache/v1/storage_backend/connector/s3_connector.py
@@ -21,6 +21,13 @@ from lmcache.v1.storage_backend.local_cpu_backend import LocalCPUBackend
 
 logger = init_logger(__name__)
 
+# HTTP statuses that mark a request as permanently rejected rather than
+# transiently failed: the request is refused for what it is, so retrying it
+# for every subsequent chunk hits the same wall. They therefore count toward
+# the connection circuit breaker. 404 is deliberately excluded: it is an
+# expected cache miss, not a broken connection.
+NON_RETRYABLE_STATUS_CODES = frozenset({400, 401, 403})
+
 
 class Priorities(IntEnum):
     PEEK = auto()
@@ -298,9 +305,15 @@ class S3Connector(RemoteConnector):
         self,
         key_str: str,
         mem_obj: MemoryObj,
-    ) -> "s3.S3Request":
+    ) -> tuple["s3.S3Request", dict]:
         """
         Download a file from S3.
+
+        Returns:
+            The in-flight request and the dict its completion callback
+            records into. The dict holds the aws-crt error under "err" and
+            the HTTP status under "status", and is only meaningful once
+            `finished_future` has resolved.
         """
         headers = HttpHeaders()
         headers.add("Host", self.s3_endpoint)
@@ -315,14 +328,17 @@ class S3Connector(RemoteConnector):
             # Directly write chunk to the memory object at the correct offset
             self._write_mem_obj(mem_obj, chunk, offset)
 
+        done = {"err": None, "status": None}
+
         # NOTE(Jiayi): Run in crt threads (not this thread) with GIL
         # See https://github.com/awslabs/aws-crt-python/blob/4250709624119de1af3ca86816e1a154fcac7cc8/source/common.c#L51
+        # Only record here: aws-crt resolves `finished_future` before running
+        # this callback, so raising would not reach the awaiting coroutine.
+        # The caller inspects `done` after the await instead, the same way
+        # `_get_object_size` inspects `got`.
         def on_done(error=None, status_code=None, **kwargs):
-            ok = (status_code in (200, 206)) or (status_code is None)
-            if error or not ok:
-                raise RuntimeError(
-                    f"Failed to download {key_str} from S3: {error or status_code}"
-                )
+            done["err"] = error
+            done["status"] = status_code
 
         # TODO(Jiayi): Need to support offset to enable zero-copy
         # More concretely, we need to get the shared memory offset.
@@ -336,7 +352,27 @@ class S3Connector(RemoteConnector):
             on_done=on_done,
         )
 
-        return s3_req
+        return s3_req, done
+
+    @staticmethod
+    def _download_error(key_str: str, done: dict) -> Optional[str]:
+        """
+        Check a completed download for failure.
+
+        Args:
+            key_str: The key that was downloaded, used in the message.
+            done: The dict recorded by `_s3_download`'s completion callback,
+                holding the aws-crt error under "err" and the HTTP status
+                under "status".
+
+        Returns:
+            An error message if the download failed, None otherwise.
+        """
+        status_code = done["status"]
+        ok = (status_code in (200, 206)) or (status_code is None)
+        if done["err"] or not ok:
+            return f"Failed to download {key_str} from S3: {done['err'] or status_code}"
+        return None
 
     async def get(self, key: CacheEngineKey) -> Optional[MemoryObj]:
         # Circuit breaker: if connection is disabled, return None immediately
@@ -380,7 +416,7 @@ class S3Connector(RemoteConnector):
             memory_obj.ref_count_down()
             return None
 
-        s3_req = self._s3_download(
+        s3_req, done = self._s3_download(
             key_str=key_str,
             mem_obj=memory_obj,
         )
@@ -389,15 +425,22 @@ class S3Connector(RemoteConnector):
             # use blocking_timeout_sec in config to control the timeout
             await asyncio.wrap_future(s3_req.finished_future)
 
+            download_error = self._download_error(key_str, done)
+            if download_error is not None:
+                raise RuntimeError(download_error)
+
             # Reset failure counter on success
             self._reset_connection_failures()
 
             return memory_obj
         except Exception as e:
             error_msg = str(e)
+            status_code = self._failed_status_code(e, done)
 
             # Update connection failures and check if it's a connection error
-            is_connection_error = self._update_connection_failures(error_msg)
+            is_connection_error = self._update_connection_failures(
+                error_msg, status_code
+            )
 
             if not is_connection_error:
                 # Log non-connection errors
@@ -419,6 +462,7 @@ class S3Connector(RemoteConnector):
 
         memory_objs: List[Optional[MemoryObj]] = []
         futures = []
+        dones = []
         future_to_memobj_idx = []
 
         for idx, key in enumerate(keys):
@@ -460,12 +504,13 @@ class S3Connector(RemoteConnector):
 
             memory_objs.append(memory_obj)
 
-            s3_req = self._s3_download(
+            s3_req, done = self._s3_download(
                 key_str=key_str,
                 mem_obj=memory_obj,
             )
             fut = asyncio.wrap_future(s3_req.finished_future)
             futures.append(fut)
+            dones.append(done)
             future_to_memobj_idx.append(len(memory_objs) - 1)
 
         # Use return_exceptions to prevent one failure from stopping all downloads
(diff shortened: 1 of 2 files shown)


aryangorde8:
@ApostaC — this pull request has been open since Monday and CI is green (`k3-unit-tests` passing; `unit-tests-amd` is queued, as it currently is across the open pull requests). Would you be able to review it, or direct me to whoever maintains the S3 connector?

Two points in the description are the only ones I believe require a maintainer's decision rather than mine:

1. `NON_RETRYABLE_STATUS_CODES = frozenset({400, 401, 403})`. 404 is excluded deliberately, as `_get_object_size` treats it as an expected cache miss and there is a regression test covering that. Whether 405, 409 or 5xx belong in the set is a judgement call, and I am willing to change it.
2. `connection_failures` and `connection_disabled` now also count HTTP rejections, so their names are narrower than their meaning. I have left them unchanged to keep the diff reviewable; renaming is mechanical and could follow separately.

The remainder is self-contained. The change is confined to `s3_connector.py`, and the tests stub the `s3.S3Request` boundary only: no bucket, credentials, network or GPU are involved, and they complete in a few seconds on CPU.

One update since I opened this pull request. The final bullet noted that `close()` awaits `AsyncPQExecutor.shutdown`, which blocks on a coroutine scheduled on the loop it is awaited on, and stated that I would file it separately. That is now #4490.

If it would make review easier, I would be glad to split this into two pull requests: the `on_done` callback change and the circuit-breaker change are independent of one another.

aryangorde8:
`buildkite/unit-tests-amd` went red on this PR (build #10714). The failure is not from
this change, and this PR's tests did not run.

The build stopped at 37%:

```
FAILED tests/v1/gpu_connector/test_kv_format_classification.py::test_classification_matches_golden
AssertionError: EngineKVFormat.NB_NL_TWO_BS_NH_HS: got (False, False, False, False), expected (True, False, False, False)
1 failed, 2125 passed, 79 skipped, 1 xpassed in 326.06s
```

The pipeline runs `pytest --maxfail=1` over the whole suite, so it aborted in
`tests/v1/gpu_connector/` and never reached `tests/v1/storage_backend/`. The tests added
here were not executed.

On the commit that was built, the two sides of that assertion agree in source:
`csrc/engine_kv_format.h` gives `NB_NL_TWO_BS_NH_HS` (index 0) `.is_cross_layer = true`,
and `EXPECTED` in the test says `(True, False, False, False)`. So the module loaded on the
ROCm agent disagreed with the source it was built from. The golden values are unchanged on
current `dev` — #4473 only renamed `lmcache.c_ops` to `lmcache.lmcache_native` — so `dev`
should reproduce this.

Two things that may matter to whoever picks it up:

- **This pipeline is not running.** Of its 61 most recent builds, 32 are still `scheduled`
  and 28 `skipped`; none completed. Every recent `dev` build is skipped, so there is no
  green baseline to compare against. This build was queued on 2026-08-11 and only ran
  today.
- **A failing run never cleans up its workspace.** In `.buildkite/pipeline.yml` the
  cleanup (`cd /; sudo rm -rf "$TARGET"`) is after `pytest` in the same command block, so
  the next build on that agent inherits the previous checkout. With the
  `c_ops` -> `lmcache_native` rename in flight across several branches, an inherited
  workspace is one way an installed module ends up disagreeing with the source it was
  built from.

I re-ran this PR's tests locally on Python 3.12 (the version the pipeline uses): 12 passed.

Filed the AMD failure separately as #4572 so it does not have to live in this thread.

aryangorde8:
Rebased onto current `dev` and force-pushed. The change itself is unchanged: +582/-27
across the same two files.

That also resolves the `unit-tests-amd` failure I described above, and I was wrong about
its cause. It was not a bug on `dev` — the branch was simply based before #4453, which
moved `EngineKVFormat` and the `is_*` predicates out of the device module into
`lmcache_native`. On a `hipcc`-compiled build those predicates returned all-`False`, so
`test_classification_matches_golden` failed at 37% and `--maxfail=1` stopped the suite
there, which is why this PR's own tests never ran.

#4490's build, on a post-#4453 base, passes that test on the same agent. I have corrected
#4572 accordingly.

Tests re-run on the rebased tree under Python 3.12 (the version the pipeline uses):
12 passed.

aryangorde8:
Correcting myself: the rebase did not clear this, and my explanation above was wrong.

Build #11142 ran against the rebased head and failed on the same assertion. #4453 did not
fix the native predicates — in the window between #4453 and #4473 the test happened to
read them through the `lmcache.c_ops` compat shim, which forwards to the pure-Python
`lmcache.v1.platform.ops_types`, so it was not exercising the native path at all. #4473
pointed it back at `lmcache_native` and it fails again. Full evidence in #4572, which I
have corrected.

The practical consequence for this PR: `unit-tests-amd` cannot go green here until that
ROCm bug is fixed, since any branch on current `dev` hits it at 37% and the suite stops
before reaching `tests/v1/storage_backend/`. `k3-unit-tests` passes on the rebased head,
and the tests added here pass locally on Python 3.12.

The rebase itself is fine and I have left it in place — the branch is now current with
`dev` and the diff is unchanged.

Issue #2923, opened 2026-03-31
[Roadmap] LMCache Roadmap for 2026 Q2


sammshen:
# LMCache Q2 Roadmap

With the influx of AI generated code, LMCache has adopted two strategies: 
- the introduction of cursor bugbot which will serve as a first pass on PRs, which will ultimately end up still requiring two maintainer approvals.
- PRs that are overly large in scope or obviously generated by AI without any effort put into keeping the code up to standard and without a human generated PR description may end up not being reviewed. Stricter reviews are **highly recommended** as code becomes easier to generate.

## 1. LMCache MP and CLI Production Readiness

Q1 marked the development of LMCache MP mode, which is the future of LMCache, along with the LMCache Operator. Much of the documentation, observability, CI/CD, and fault tolerance that will make LMCache production ready is still missing.

- [ ] Advanced L2 store/prefetch policies 
- [ ] Dynamic prefetch admission controller
- [x] Non GPU wheels for only CLI
- [x] AMD support for MP mode
- [x] Comprehensive documentation for non-GPU CLI usage (like kubectl)
- [x] LMCache recipes & tutorials for popular open models and HW
- [x] Production deployment best practices scaling, tuning, monitoring recommendations

### Fault Tolerance:

- [x] Stress test high when requests fail, serving engines go down, LMCache goes down, high concurrency, high preemption, etc.
- [x] Add reproducible fault tolerance CI/CD

### Observability:

- [ ] Worker affinity & health check stabilization
- [x] K8s operator
- [x] Observability setup recipes like OTel collector + Grafana configuration

## 2. Core Features and KV Cache Optimizations (lead by @KuntaiDu)

LMCache aims to be the habitat of all of the latest KV Cache related research such as KV compaction, compression, codec, blending etc. providing an interface to plug in these optimizations without hacking into serving engines like vllm or hf transformers. A clean and modular interface along with many of the long awaited LMCache features like CacheBlend will finally hit the Open Source Community in the MP mode in Q2!

- [ ] Release MP mode CacheBlend support
- [x] Interface to support token dropping in KV cache (sparse KV cache optimizations)
- [x] Interface to support KV cache quantization (e.g., TurboQuant)
- [x] Interface to support KV cache encoding and decoding (i.e., KV cache <-> bytestreams e.g. CacheGen)

## 3. Northbound Tree

[...]

PR #4486, opened 2026-08-10
feat(mp): add TOML config-file loader for the multiprocess server

SuperMarioYL:

Adds a TOML config-file loader for the multiprocess (MP) cache server, addressing the Q2 2026 roadmap item **"MP toml configuration"** (#2923 §3, currently unchecked/unassigned).

Today the MP server is argparse-only (`add_mp_server_args` in `lmcache/v1/multiprocess/config.py`): every fleet deployment must pass ~17 CLI flags, and there is no way to ship a reproducible, reviewable config file. This PR introduces:

- `LMCACHE_MP_CONFIG` env var + `--config-file` CLI flag pointing at a TOML file.
- `load_mp_config_from_toml(path)` — reads the TOML (stdlib `tomllib` on Python ≥3.11, `tomli` fallback on 3.10), normalizes keys to `MPServerConfig` field names (e.g. `max-gpu-workers` → `max_gpu_workers`), and returns `{}` on missing/invalid/non-mapping files with a warning (never raises).
- `merge_mp_config_file_into_args(args)` — resolves `--config-file` > `LMCACHE_MP_CONFIG`, then merges the file into the argparse `Namespace` **before** `parse_args_to_mp_server_config`: the file supplies defaults, and a CLI flag wins only when it differs from the argparse default (so `--port 6000` beats `port = 5555` in the file, but an absent flag lets the file win over the default).
- Wirings at all three CLI entries (`server.py`, `http_server.py`, `cli/commands/server.py`) so the merge runs before config construction.
- `tomli; python_version < "3.11"` conditional dependency (in `requirements/common.txt` — the repo loads deps dynamically from `requirements/*.txt` via `setup.py`, so this is the correct location rather than a static `pyproject.toml` slot).
- `docs/source/mp/configuration.rst` — a "Configuration File (TOML)" section with a sample `mp.toml`.

- **Why TOML and not reuse the in-process `LMCACHE_CONFIG_FILE` YAML path?** The Q2 roadmap (#2923 §3) explicitly names TOML for MP config; the Q3 roadmap (#4025) is silent on the format and does not supersede. So this introduces TOML per the roadmap mandate rather than reusing YAML. (This pre-empts the natural "reuse YAML" question.)
- **Backward compatible**: with no `--config-file` and no `LMCACHE_MP_CONFIG`, behavior is byte-for-byte identical to today (argparse-only). The three wirings are each one import + one line, before `parse_args_to_mp_server_config`.
- **Graceful degradation**: missing / invalid / non-mapping TOML → warn and fall back to argparse defaults. The server never aborts startup on a config-file problem.
- **Precedence**: `--config-file` flag > `LMCACHE_MP_CONFIG` env var; CLI flags > file values (only when the flag differs from the argparse default); `config-file`/`LMCACHE_MP_CONFIG` themselves are reserved keys (rejected if present in the TOML).
- **Scope**: deliberately additive — no `MPServerConfig` field add/remove/rename, no change to the in-process `config.py` YAML loader, no change to `parse_args_to_p2p_config`.

**If applicable**:

- [x] this PR contains user facing changes - docs added (`docs/source/mp/configuration.rst`)
- [x] this PR contains unit tests (`tests/v1/multiprocess/test_mp_config.py` — 18 tests covering: TOML load → MPServerConfig fields, key normalization, CLI-overrides-file precedence, flag-equal-to-default-lets-file-win, `runtime_plugin_config` table→JSON, `LMCACHE_MP_CONFIG` env resolution, flag-over-env, backward-compat no-file, missing/invalid/non-mapping graceful fallback, and the 3.10 `tomli` fallback path via a shim on 3.11+)

Refs #2923

### DIFF
--- a/docs/source/mp/configuration.rst
+++ b/docs/source/mp/configuration.rst
@@ -55,7 +55,7 @@ unchanged (argparse-only).
 
 Point the server at a file with ``--config-file`` or the
 ``LMCACHE_MP_CONFIG`` environment variable (the flag wins over the env var).
-A missing or unparseable file logs a warning and falls back to argparse-only
+A missing or unparsable file logs a warning and falls back to argparse-only
 behavior — it never aborts startup.
 
 Keys mirror the ``MPServerConfig`` field names; both ``max-gpu-workers`` and
--- a/docs/source/mp/configuration.rst
+++ b/docs/source/mp/configuration.rst
@@ -1,5 +1,4 @@
 Configuration Reference
-=======================
 
 This page documents every CLI argument accepted by the LMCache multiprocess
 server.  Arguments are grouped by the config module that defines them.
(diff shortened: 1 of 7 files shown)


SuperMarioYL:
Resolved the base drift: `dev` had moved 167 commits ahead and the only conflicting file was `docs/source/mp/configuration.rst` (upstream's new "Per-request LMCache configuration" section landed at the same spot as our TOML section). Rebased onto the latest `dev` and kept both sections — upstream's first, then the TOML config-file section. All Python changes merged cleanly. The TOML loader test suite passes locally (18/18) and the branch is mergeable again.

SuperMarioYL:
Resolved the base drift: `dev` had moved 167 commits ahead and the only conflicting file was `docs/source/mp/configuration.rst` (upstream's new "Per-request LMCache configuration" section landed at the same spot as our TOML section). Rebased onto the latest `dev` and kept both sections — upstream's first, then the TOML config-file section. All Python changes merged cleanly. The TOML loader test suite passes locally (18/18) and the branch is mergeable again.

PR #4488, opened 2026-08-11
CI: fix gemma4 HMA test by enabling --separate-object-groups explicitly

maobaolong:
The hma_lm_eval (gemma-4) multiprocess test was added relying on the separate_object_groups config default of True (hybrid per-group HMA path). Commits #3869 and #4437 flipped that default to False, so gemma-4's sliding-window + full-attention groups collapse into a single full-attention object group, breaking per-group store/retrieve and the bit-exact score check.

Set SEPARATE_OBJECT_GROUPS=1 for the test (plumbed through launch-processes.sh) so it no longer depends on the server default.

**If applicable**:

### DIFF
--- a/.buildkite/k3_tests/multiprocess/pipeline.yml
+++ b/.buildkite/k3_tests/multiprocess/pipeline.yml
@@ -220,30 +220,7 @@ steps:
         plugins: [{ kubernetes: { podSpec: *pod-1gpu-dshm } }]
         artifact_paths: ["*.log"]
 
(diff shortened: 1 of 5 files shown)


ApostaC (commented):
can add to both deepseek v4 and kimi linear test

ApostaC:
also disable gemma 4 k3 hma test

ApostaC (approved):

chunxiaozheng (approved):
lgtm

PR #4490, opened 2026-08-11
[Bug][Executor] drain the queue before cancelling workers on AsyncPQExecutor shutdown

### LINKED ISSUE #4489: [Bug] AsyncPQExecutor shutdown deadlocks: close() awaits a sync shutdown, and shutdown_async cancels workers before draining the queue
**Describe the bug**

`AsyncPQExecutor` in `lmcache/v1/storage_backend/job_executor/pq_executor.py` cannot
be shut down reliably. There are two separate problems, and the correct pattern for
the second one already exists in the same file.

**Problem 1 — `S3Connector.close()` cannot complete**

```python
async def close(self):
    await self.pq_executor.shutdown(wait=True)
```

`AsyncPQExecutor.shutdown` is **synchronous** (it returns `None`), so `await`ing it is
already wrong. More importantly it is implemented as:

```python
def shutdown(self, wait: bool = True) -> None:
    future = asyncio.run_coroutine_threadsafe(self.shutdown_async(wait), self.loop)
    if wait:
        future.result()
```

`close()` is a coroutine that runs *on* `self.loop`, so `future.result()` blocks that
loop while waiting for a coroutine that needs the same loop to make progress. It
deadlocks.

Reproduced with a standalone probe: scheduling `await executor.shutdown(wait=True)`
onto the executor's own loop never returns, and the call has to be abandoned by
timeout.

**Problem 2 — `shutdown_async(wait=True)` can hang on `queue.join()`**

In `AsyncPQExecutor.shutdown_async`, the workers are cancelled *before* the queue is
drained:

```python
        # Cancel all worker futures to ensure they stop
        # Even when wait=False, we should cancel workers to prevent hanging coroutines
        for fut in self._workers:
            if not fut.done():
                fut.cancel()          # <-- line 114

        if wait:
            # Let workers drain tasks and exit gracefully
            await self._queue.join()  # <-- line 118
```

A sentinel is pushed per worker just above this, and `_worker` calls
`self._queue.task_done()` when it consumes one. If a worker is cancelled before it
consumes its sentinel, `task_done()` is never called for that item, the queue's unfinished
count never reaches zero, and `await self._queue.join()` waits forever.

**`AsyncPQThreadPoolExecutor.shutdown_async`, in the same file, already does this in the
right order** — it awaits `self._queue.join()` first and only then cancels the worker
futures, with a separate `else` branch that cancels immediately when `wait=False`. The
subclass is correct; the base class is not.

**To reproduce**

Timing dependent, so it is interpreter-sensitive. A test fixture that tears down with
`await executor.shutdown_async(wait=True)` on the executor's loop passes consistently on
Python 3.10, 3.11 and 3.12, and hangs on **every** test on Python 3.13 (3.13.15), where
it surfaces as a `TimeoutError` from whatever bounds the wait. CI's `Test` matrix covers
3.13, so this is reachable from CI.

**Expected behavior**

`shutdown_async(wait=True)` drains and returns, and `close()` on a connector that owns an
`AsyncPQExecutor` completes rather than deadlocking.

**Suggested fix**

For problem 2, match the ordering the subclass already uses: `join()` first, cancel
after. For problem 1, `close()` should await `shutdown_async` directly rather than the
synchronous `shutdown` wrapper.

**Additional context**

Found while adding tests in #4484 — the fixture there works around this by using
`wait=False`, which cancels the workers without waiting on a drain a cancelled worker can
no longer complete. That is a workaround in test code, not a fix for the executor. I have
not touched `pq_executor.py` in that PR, to keep its scope to the S3 connector.

I am happy to send a PR for this if the suggested direction looks right.

aryangorde8:
Fixes #4489

`AsyncPQExecutor.shutdown_async` cancelled the worker futures and *then* awaited
`queue.join()`. A sentinel is pushed per worker just above that, and `_worker`
calls `task_done()` when it consumes one — so a worker cancelled before it
reaches its sentinel leaves the queue's unfinished count above zero and the
join waits forever.

`AsyncPQThreadPoolExecutor.shutdown_async`, in the same file, already does this
in the correct order: drain first, cancel after, with a separate `else` branch
that cancels immediately when `wait=False`. This PR makes the base class match
its own subclass.

`S3Connector.close()` separately awaited the **synchronous** `shutdown`, which
calls `run_coroutine_threadsafe(...).result()` — blocking the very loop the
coroutine needs in order to make progress. It now awaits `shutdown_async`.

@zhengfeihe — kept small as asked: +16/−8 across two source files, plus a test
file. No behaviour change beyond the ordering.

The hang is timing dependent, which is worth knowing when reading the test.
Against the unfixed executor the two `wait=True` tests pass on Python 3.10,
3.11 and 3.12 and **fail on 3.13** (10s timeout, 5/5 runs), where cancellation
lands before the workers drain. The CI matrix covers 3.13, so this is reachable
from CI — it is how the bug was found, via a fixture teardown in #4484.

Note this touches `s3_connector.py`, which #4484 also modifies, in a different
region. They should merge cleanly; whichever lands second I will rebase if
needed.

#4484's test fixture currently works around this by shutting down with
`wait=False`. I left that as is rather than have that PR depend on this one.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/s3_connector.py
+++ b/lmcache/v1/storage_backend/connector/s3_connector.py
@@ -710,4 +710,6 @@ class S3Connector(RemoteConnector):
             self.connection_failures = 0
 
     async def close(self):
-        await self.pq_executor.shutdown(wait=True)
+        # `shutdown` is synchronous and blocks on a coroutine scheduled on this
+        # same loop, so awaiting it here would deadlock. Await the async form.
+        await self.pq_executor.shutdown_async(wait=True)
--- a/lmcache/v1/storage_backend/job_executor/pq_executor.py
+++ b/lmcache/v1/storage_backend/job_executor/pq_executor.py
@@ -107,20 +107,26 @@ class AsyncPQExecutor(BaseJobExecutor):
                 (sentinel_priority, next(self._counter), _SENTINEL, None, {}, None)
             )
 
-        # Cancel all worker futures to ensure they stop
-        # Even when wait=False, we should cancel workers to prevent hanging coroutines
-        for fut in self._workers:
-            if not fut.done():
-                fut.cancel()
-
         if wait:
-            # Let workers drain tasks and exit gracefully
+            # Let workers drain tasks and exit gracefully. Cancelling before
+            # draining would stop a worker before it consumes its sentinel, so
+            # `task_done` would never be called for that item and this join
+            # would wait forever.
             await self._queue.join()
+            # Cancel all worker futures after queue is drained
+            for fut in self._workers:
+                if not fut.done():
+                    fut.cancel()
             # Wait for all workers to complete (with cancellation handling)
             await asyncio.gather(
                 *[asyncio.wrap_future(fut, loop=self.loop) for fut in self._workers],
                 return_exceptions=True,
             )
+        else:
+            # When wait=False, just cancel workers immediately
+            for fut in self._workers:
+                if not fut.done():
+                    fut.cancel()
 
     def shutdown(self, wait: bool = True) -> None:
         """Sync shutdown — blocks the calling thread until shutdown is complete."""
--- a/tests/v1/storage_backend/test_pq_executor.py
+++ b/tests/v1/storage_backend/test_pq_executor.py
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Unit tests for `AsyncPQExecutor` shutdown."""
+
+# Standard
+import asyncio
+import threading
+
+# Third Party
+import pytest
+
+# First Party
+from lmcache.v1.storage_backend.job_executor.pq_executor import AsyncPQExecutor
+
+
+@pytest.fixture
+def loop_in_thread():
+    loop = asyncio.new_event_loop()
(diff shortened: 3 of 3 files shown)


aryangorde8:
`buildkite/unit-tests-amd` went red on this PR (build #10748). The failure is not from this
change, and this PR's tests did not run.

The pipeline runs `pytest --maxfail=1` over the whole suite. It stopped at 62%:

```
FAILED tests/v1/multiprocess/test_blocked_scale_format.py::test_blocked_roundtrip_value_exact_across_alignments
AssertionError: misaligned scatter must land every token's values AND scale at the blocked-layout positions of the destination slots
1 failed, 3546 passed, 88 skipped, 1 xpassed in 404.00s
```

`tests/v1/storage_backend/` sorts after `tests/v1/multiprocess/`, so
`tests/v1/storage_backend/test_pq_executor.py` was never reached.

The failing assertion is a misaligned H2D scatter for `EngineKVFormat.NL_X_NB_BSV_BSS`:
the gather half of the test passes and the scatter returns whole rows zeroed. That is a
device-transfer path with nothing in common with the executor shutdown change here.

Reported with the rest of the AMD CI state in #4572. That test landed on 2026-07-31 with
#4351 and this pipeline has not completed a build since, so it may have been failing on
ROCm since then.

PR #4491, opened 2026-08-11
[CI][FIX] Stop the global EventBus leaked by init_observability tests

zhengfeihe:

Found unit test scenario_record_drain_event failed in https://github.com/LMCache/LMCache/pull/4466 's pipeline,
error said 

```
        # Multiple records on the default stream (ptr=0) must all enqueue.
        for i in range(5):
            ops.record_event_on_stream(0, f"mp.test.{i}", f"sess-{i}", {}, {"idx": i})
        device_sync(device)
        events = ops.drain_recorded_events()
>       assert len(events) == 5
E       assert 0 == 5
E        +  where 0 = len([])
```

This is caused by extra event bus started by previous tests not stopped properly 

event recorder is static , so any other threads can drain it 

```
EventRecorder& EventRecorder::instance() {
  static EventRecorder recorder;
  return recorder;
}
```

**If applicable**:


aryangorde8:
I hit a similar test-pollution problem this week, so I read this one closely and reproduced the underlying leak locally. The diagnosis is right and the fix does resolve the reported failure.

**The root cause is `init_event_bus`, not these tests.** `init_observability` ends with `bus.start()` (`config.py:494`), so it returns a bus with a live drain thread. `init_event_bus` then replaces the global without stopping the outgoing one:

```python
global _global_bus, _observability_enabled
_global_bus = EventBus(config)
```

That is why the thread survives. Reproduced against `dev` at `f82f6fd3`, with `init_otel_metrics` patched exactly as these tests do:

```
after 1st init_observability : 1 EventBus thread
after 2nd init_observability : 2 EventBus threads
first bus thread still alive : True
global is now the second bus : True
```

The first bus is unreachable and still draining the static recorder — the `assert 0 == 5` in #4466. Stopping both brings it back to 0.

This reaches production, not just tests: `lmcache/v1/multiprocess/server.py:368` and `lmcache/cli/commands/trace/_driver.py:177` both call `init_observability`, so any process that initialises observability twice orphans a started bus with every subscriber still attached. Having `init_event_bus` stop the outgoing `_global_bus` before replacing it would fix this once for every caller, including tests not yet written.

That is clearly a separate change and I would not hold this PR for it — the CI failure is real and this unblocks it. Flagging so it does not get lost.

**One suggestion on the cleanup, on maintainability grounds rather than correctness.** The two cleanup lines run unguarded, so I checked whether an exception could leak a bus here. It cannot: `bus.start()` is the last thing `init_observability` does, so a raise inside the `with patch(...)` block happens before any thread exists. Worth stating so nobody else spends time on it. (It does leave `_global_bus` pointing at an enabled-but-unstarted bus, which pollutes differently, but that would not reproduce this failure.)

The reason I would still suggest a fixture is that the next test added to this file has to remember both lines. The suite already has two patterns for this:

- `tests/v1/mp_observability/test_config.py:69-89` — `try` / `finally` with `init_event_bus(EventBusConfig(enabled=False))` in the `finally`.
- `tests/v1/distributed/test_distributed_storage_manager.py:906-923` — a fixture doing `try: yield ... finally: bus.stop(); init_event_bus(...)`.

This file already has an autouse fixture (`_mock_otel_provider`), so the shape is there. Entirely your call — the current form is correct for the two tests as written.

For the record, I confirmed these two are the complete set in this file: the other three tests call `init_otel_metrics` directly and never construct a bus.

PR #4497, opened 2026-08-11
[Operator] Add optional initContainers to LMCacheEngine

elliotz-ai:

The `raw_block` L2 adapter's `RawBlockDevice` opens `device_path` expecting the file to already exist at the configured size — it never creates or grows it. On a fresh node, the engine's DaemonSet pod crash-loops on startup with:

```
File "lmcache/v1/storage_backend/raw_block/core.py", line 417, in _rawdev
    self._raw = RawBlockDevice(...)
FileNotFoundError: [Errno 2] open failed
```

Today there is no operator-native way to prepare that file before the `lmcache` container starts — `buildDaemonSetCore` hardcodes a single container with a fixed `Command`/`Args`, and `LMCacheEngineSpec` has no init-container-style hook. Anyone hitting this has to either build a custom image with a startup wrapper, or run an out-of-band job outside the operator's reconciliation to pre-`fallocate`/`truncate` the file on every node before the DaemonSet schedules there.

**What this PR does**:

- Adds `spec.initContainers []corev1.Container` to `LMCacheEngineSpec`, following the same pass-through pattern as the existing `spec.volumes`/`spec.volumeMounts` fields.
- Wires it straight into the DaemonSet's pod template (`buildDaemonSetCore`) as `InitContainers`, run before the main `lmcache` container as Kubernetes' init-container semantics already guarantee.
- No default behavior changes: the field is optional and empty by default, so existing engines render an identical DaemonSet.

This unblocks the common case of pre-sizing a `raw_block` device path via a small init container (e.g. `busybox` running `fallocate`/`truncate`) targeting the same node-local volume mount as the main container, without any custom image or out-of-band tooling.

- Regenerated `zz_generated.deepcopy.go` and `config/crd/bases/lmcache.lmcache.ai_lmcacheengines.yaml` via `make generate manifests`. The CRD diff is large (~1500 lines) purely because `controller-gen` inlines the full `corev1.Container` OpenAPI schema, matching what already happens for the existing `volumes`/`volumeMounts` fields in this same file.
- Verified `go build ./...`, `go test ./...`, and `make lint` all pass. The two pre-existing `envtest`-based failures (`internal/webhook`, `internal/controller`) reproduce identically on an unmodified `dev` checkout (missing local envtest binaries) and are unrelated to this change.

**If applicable**:

- [x] this PR contains user facing changes - CRD field addition (documented via field comment/OpenAPI description)

### DIFF
--- a/docs/source/mp/operator.rst
+++ b/docs/source/mp/operator.rst
@@ -614,6 +614,12 @@ Overrides & Extras
    * - ``volumeMounts``
      - --
      - Extra volume mounts.
+   * - ``initContainers``
+     - --
+     - Additional init containers run before the lmcache container starts,
+       in order.  Use to prepare state the engine can't create itself, e.g.
+       pre-sizing a ``raw_block`` L2 adapter's ``device_path`` file (see
+       :ref:`mp-operator-preexisting-l2-state` below).
    * - ``podAnnotations``
      - --
      - Extra pod annotations.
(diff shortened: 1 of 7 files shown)


royyhuang (changes requested):
Overall LGTM. Please add user side documentation in `docs/source/mp/operator.rst`. Also please update the sample config in `operator/config/samples/lmcache_v1alpha1_lmcacheengine.yaml` following the file's convention (e.g., optional specs are listed in comments for illustration purpose).

elliotz-ai:
Added both — initContainers documented in the CRD Spec Reference table plus a worked example (pre-sizing a raw_block L2 device) in operator.rst, and a matching commented block in lmcache_v1alpha1_lmcacheengine.yaml following the file's convention.

> Overall LGTM. Please add user side documentation in `docs/source/mp/operator.rst`. Also please update the sample config in `operator/config/samples/lmcache_v1alpha1_lmcacheengine.yaml` following the file's convention (e.g., optional specs are listed in comments for illustration purpose).

royyhuang (approved):
LGTM!

ApostaC (approved):

PR #4500, opened 2026-08-12
[Example] Tune R-KV token-dropping default config and add Qwen3-1.7B Colab demo

wanke1997:

Adopt `lambda=0.25`, `sim_threshold=0.5`, `window=16`, `kernel=7` as the R-KV
default across the token-dropping notebooks. This was selected by a grid search
on Qwen3-8B (30 LongBench-V2 prompts, 50% KV drop) and re-validated with
repeats: it lands ~16/30 mean (vs ~15.5/30 for the previous 0.4 threshold and
10–11/30 no-drop baseline). Refresh the baseline/drop accuracy numbers in every
notebook's summary tables to match re-run outputs; throughput and compression
time are unchanged within measurement noise so they are kept.

Also add `rkv_colab.ipynb`, a self-contained Qwen3-1.7B token-dropping demo for
a single Colab T4.

Example/notebook-only change; no library code is touched. Accuracy on this
workload carries ~1–2/30 run-to-run noise from decode non-determinism (see the
"Reading these numbers" section), so the four 8B implementations land at
16/16/15/14 for the same config — all within that band, none systematically
better. The exact variants still assert their redundancy against the literal
reference in their self-tests, so they select identical tokens; residual
differences are 1–2 bfloat16 ULP in the RoPE re-rotation (CPU vs GPU).

**If applicable**:


KuntaiDu (approved):
LGTM! Can you fix the precommit?

wanke1997:
> LGTM! Can you fix the precommit?

Fixed, thanks for reminder!

ApostaC (approved):
Thanks!

PR #4502, opened 2026-08-12
refactor: remove legacy c_ops surface

maobaolong:

This PR completes the native-ops refactor so LMCache no longer exposes a generic compiled-module compatibility surface:

- Remove the legacy generic ops module and its synthetic compatibility shim entirely.
- Export the resolved per-platform singleton as `lmcache.device_ops` for device-agnostic callers.
- Name compiled extensions by platform: `lmcache.cuda_ops` for CUDA/ROCm and `lmcache.xpu_ops` for XPU.
- Move all CUDA implementation files and private headers from `csrc/` into `csrc/cuda/`; keep only device-independent native code at the root.
- Update CUDA/ROCm build profiles, hipify paths, wheel validation, CI, tests, and documentation.
- Make shared Python types and optional platform-native symbols explicit on the `DeviceOps` boundary.

The resulting source layout separates shared native infrastructure from accelerator implementations and allows new devices to provide their own ops without inheriting CUDA naming.

**Validation**:

- Changed-file pre-commit suite, including ruff, mypy, codespell, and clang-format.
- 217 tests passed, 15 skipped across platform dispatch, torch fallback, connectors, multiprocess, DAX, and parity coverage.
- Sphinx HTML build completed successfully; its existing unrelated warnings remain unchanged.
- Repository source, tests, CI, and docs contain no references to the removed generic ops name.

After this PR, the generated so files on cuda test machine are:
- cuda_ops.cpython-312-x86_64-linux-gnu.so 21716128 bytes
- lmcache_fs.cpython-312-x86_64-linux-gnu.so 7367432 bytes
- lmcache_native.cpython-312-x86_64-linux-gnu.so 7828464 bytes
- lmcache_redis.cpython-312-x86_64-linux-gnu.so 7213672 bytes

**If applicable**:


setup_extensions/build_profiles/rocm.py
```diff
@@ -83,25 +86,26 @@ def build(self) -> tuple[list["Extension"], dict]:
...
+                "cal_cdf.cu",
+                "ac_enc.cu",
+                "ac_dec.cu",
+                "pos_kernels.cu",
+                "mem_alloc.cpp",
+                "utils.cpp",
+                "event_recorder.cpp",
+                "completion_recorder.cpp",
+            ]
+        )
```
ApostaC, line 104: Why is this not `csrc/cuda/xxx.hip`?

ApostaC (commented):
Only one quick comment regarding the AMD build. PTAL

setup_extensions/build_profiles/rocm.py
maobaolong: Because this list names the canonical CUDA sources to hipify, not checked-in HIP files, the actual HIP outputs are generated at build time under ./csrc_hip/cuda`, so `./csrc/cuda/*.hip` would be misleading here.`

ApostaC: oh I got it

ApostaC (approved):

chunxiaozheng (approved):
@maobaolong Thanks for your great work, LGTM!

PR #4509, opened 2026-08-12
[Bugfix][MP] Break the host-func closure cycle pinning dead workers' KV pools

zhengfeihe:

I can still observe the KV pool not being released locally after #4015.

Did some digging, and found the cause is a reference cycle involving `GPUCacheContext`.

At first glance this lambda looks like it captures `self.device_`:

```
    f = lambda logger: logger.info("... %s", str(self.device_))
```

However, due to how Python closures work, it captures `self` entirely: `self.device_` is an attribute lookup deferred to call time (late binding), so the closure cell has to hold the whole object.

This closes a reference cycle:

```
    GPUCacheContext -> cupy stream -> host-func callback -> closure cell -> GPUCacheContext
```

The release path (#4015) relies on `torch.cuda.ipc_collect()`, which only unmaps segments whose tensors are already unreferenced -- and silently  skips the rest. With the cycle, the context's refcount never reaches zero when the server drops its reference, so the dead worker's entire IPC-imported KV pool stays mapped.

Python's cyclic GC (a gen-2 pass) does clean such cycles eventually, which I believe is why the memory appeared to be released in earlier validation: on a busy server enough allocations happen to trigger a pass. But gen-2 collection is driven by allocation counts, not time -- on an idle server it may never run, and the pool stays pinned indefinitely. We should not bet the unmap on it.

**If applicable**:


zhengfeihe:
sorry false alarm :cry: 

Not a real problem

PR #4510, opened 2026-08-12
fix: eager-init kv_layer_groups_manager in V3 connector

Anduin-Lothar:

Override `initialize_kvcaches_ptr()` in `VLLMPagedMemGPUConnectorV3` to build `kv_layer_groups_manager` when kvcaches are first registered, not lazily on the first `to_gpu()`/`from_gpu()` call.

**Problem**: `cache_engine.store()` calls `metadata.get_shapes()` to allocate memory objects **before** `from_gpu()` runs. Without eager init, `kv_layer_groups_manager` is `None` at that point, so `get_shapes()` falls back to the legacy single-group path `[kv_shape[1], kv_shape[0], num_tokens, kv_shape[3]*kv_shape[4]]`. For models with heterogeneous layer groups (e.g. DeepSeek V4 with mixed compressed + dense groups), the fallback returns a single shape instead of per-group shapes, causing `memory_obj.group_prefix_sum` to be computed wrong. When `from_gpu()` later slices `memory_obj.get_tensor(i)` by `group_prefix_sum`, it reads out-of-bounds or misaligned data.

Also fix `memory_obj.tensor` → `memory_obj.raw_tensor` in `eic_connector` (3 sites). EIC needs the raw flat buffer for RDMA registration and cudaMemcpy; the `.tensor` property returns a reshape view that may be non-contiguous or offset, giving wrong `data_ptr()` under multi-group layouts.

- The V3 connector currently initializes lazily in `to_gpu()`/`from_gpu()`. This PR adds eager initialization at `register_kv_caches` time. The lazy path in `to_gpu`/`from_gpu` is preserved as a safety net (no-op if already init via `self.init` flag).
- The `eic_connector.py` change (`tensor` → `raw_tensor`) aligns with how other connectors access memory object data — `raw_tensor` is the underlying physical tensor, while `tensor` may return a permuted view.

**If applicable**:


Anduin-Lothar:
Hi @ApostaC @sammshen @maobaolong — gentle ping.

The CI workflows (Test / CPU Device Tests / PR Full Build / CodeQL / Code Quality) are still in "awaiting maintainer approval" 5 days after opening — they haven't run yet. Could a maintainer approve-and-run the workflows so we can get test results, and start review?

Thanks!

PR #4511, opened 2026-08-12
fix: prevent busy_loop deadlock in EIC connector retrieve path

Anduin-Lothar:

`EICConnector._batched_get` uses `asyncio.gather` to run N concurrent `_get` calls, each calling `get_data()` → `memory_allocator.allocate()` with the default `busy_loop=True`.

When the CPU staging buffer is full, each concurrent retrieve enters the busy_loop in `LocalCPUBackend.allocate()`, sleeping and retrying indefinitely. The loop assumes "stores will clear" to free space, but all retrieve coroutines are blocked in the busy_loop — no store is running, so space is never freed, causing a **deadlock**.

This contradicts the `LocalCPUBackend` docstring which states `busy_loop` should only be used for single-retrieve scenarios, not concurrent ones:

> 1. synchronous case: one retrieve at a time (okay to busy loop because stores will clear)
> 2. asynchronous case: many retrieves happen concurrently (if they busy_loop, deadlock happens)

Fix: pass `busy_loop=False` to `allocate()` in `get_data()`. On allocation failure, `get_data()` returns `None` and the caller (`_get`) filters it out, degrading to a cache miss instead of deadlocking.

- This is a 1-line fix. The EIC connector's `_batched_get` bypasses the async serializer (unlike the storage manager's retrieve path), so concurrent retrieves can deadlock on CPU staging buffer contention.
- Related: #2942 (deadlock in LocalCPUBackend when `use_hot=False`) and PR #3804 (not yet merged) address the same class of bug at the backend level, but EIC's concurrent retrieve path needs this fix at the connector level regardless.

**If applicable**:


Anduin-Lothar:
Hi @sammshen @maobaolong @chunxiaozheng — gentle ping.

Same situation as #4510: CI workflows still awaiting maintainer approval 5 days in, haven't run. Could someone approve-and-run + start review?

1-line fix: pass `busy_loop=False` to `allocate()` in `EICConnector.get_data()`. `_batched_get` runs N concurrent `_get` coroutines via `asyncio.gather`; when the CPU staging buffer is full, all of them enter `LocalCPUBackend.allocate()`'s busy_loop simultaneously, no store is running to free space → deadlock. After the fix, allocation failure degrades to a cache miss.

Thanks!

PR #4512, opened 2026-08-12
fix: unpin KV caches for aborted requests in sync mode

Anduin-Lothar:

In sync mode, `lookup(pin=True)` pins KV caches in the Worker's `LocalCPUBackend`. Normally unpin happens in `wait_for_save()`, but **aborted requests may skip `wait_for_save()` entirely**, leaving the pin count unbalanced and leaking pinned memory.

The Scheduler-side `request_finished()` calls `storage_manager.cancel_request()`, but `LocalCPUBackend` does not override `cancel_request()` (it is a no-op in the base class), so the Scheduler cannot release Worker-side pins across processes.

Fix: call `lookup_unpin()` for all `finished_req_ids` in `get_finished()` (Worker-side API) when `async_loading` is disabled.

**Why guard with `not self.async_loading`?**

In async mode, abort cleanup is already handled via `lookup_client.cancel_lookup()` on the Scheduler side, which triggers Worker-side `cleanup_memory_objs()`. Adding `lookup_unpin` here for async mode would risk double-cleanup.

**Is this safe for normally-completed requests?**

Yes. `lookup_unpin()` uses `dict.pop()` to remove the `lookup_id` from `lookup_pins`. If `wait_for_save()` already unpinned, the second call is a safe no-op (key not in dict).

- Related: PR #2786 (pin count balancing for `kv_consumer`), PR #3340 (graceful skip on Scheduler-role abort). These fix Scheduler-side abort cleanup, but do not address the Worker-side pin leak in sync mode.
- Related: PR #2045 (async abort handling) covers the async mode path.

**If applicable**:


Anduin-Lothar:
Hi @sammshen @maobaolong @YaoJiayi — gentle ping.

Same situation: CI workflows awaiting maintainer approval 5 days in, haven't run. Could someone approve-and-run + start review?

Quick recap of the fix: in sync mode, call `lookup_unpin()` for all `finished_req_ids` in `get_finished()` (Worker-side) when `async_loading` is disabled. Aborted requests can skip `wait_for_save()` entirely, leaving pin counts unbalanced and leaking pinned memory. `LocalCPUBackend` doesn't override `cancel_request()`, so Scheduler-side `storage_manager.cancel_request()` can't release Worker-side pins across processes. The `not self.async_loading` guard avoids double-cleanup with the async path (#2045).

`lookup_unpin()` uses `dict.pop()`, so calling it twice for already-unpinned req_ids is a safe no-op.

Thanks!

PR #4513, opened 2026-08-12
tests: refine gpu_connector tests

VincyZhang:
- Unifies test device handling to runtime-aware abstractions (`torch_dev` / `torch_device_type`) in GPU connector-related tests.
- Makes tests/disagg/test_nixl_channel.py pytest-collectable/executable by adding real test cases.
- Updates XPU CI selector in .buildkite/k3_tests/xpu/run.sh to stop globally ignoring tests/v1/gpu_connector.

- Scope is test/CI behavior only; no production logic changes.
- Goal is to reduce CUDA hard-coding and make non-CUDA test behavior deterministic.

**If applicable**:


hlin99 (approved):
LGTM

tests/v1/gpu_connector/test_kv_format_detection.py
```diff
@@ -37,21 +38,18 @@ def test_vllm_cross_layer():
 # The CPU-HND safeguard forces HND regardless of hint when running on a CPU
 # host; bypass it so the hint-driven NHD/HND branch is exercised on any host.
 _VLLM_DEV = "lmcache.v1.gpu_connector.kv_format.detectors.vllm.torch_device_type"
+_MOCK_DEVICE_TYPE = "cuda" if torch_device_type == "cpu" else torch_device_type
 
 
-def test_vllm_flash_attn_nhd_vs_hnd(monkeypatch):
-    monkeypatch.setattr(_VLLM_DEV, "cuda")
+def test_vllm_flash_attn_hnd(monkeypatch):
```
zhengfeihe, line 44: I am a little concerned why we drop nhd here?

The remaining case now feeds an NHD tensor ([2, NB, BS, NH, HS]) with an HND hint and asserts the HND format, ( I know right now we only detect by hint

zhengfeihe (commented):
Thanks for the refinement!

Since there’s no production impact as described, this should be fine. I left one comment, please kindly check it~

tests/v1/gpu_connector/test_kv_format_detection.py
VincyZhang: It's a long story. When enabling the CPU device in LMCache (https://github.com/LMCache/LMCache/pull/3280), we found a bug in vLLM: the attention backend used for CPU employs the HND KV layout but didn't report this hint, causing the KV connector to process using the NHD layout instead. This vLLM bug is fixed by https://github.com/vllm-project/vllm/pull/42740. However, considering the compatibility, we hardcoded the HND layout in LMCache —ignoring the hint—whenever the torch device was set to CPU; this is why only HND is currently handled on the CPU. 
It has been three months since the vLLM fix was implemented, so it 's time to remove this hardcoded setting. We can handle this in a separate PR, as the change involves related testing and verification that fall outside the scope of the current PR.

zhengfeihe: I see~ Thanks for the explanation.

zhengfeihe (approved):
LGTM!

PR #4517, opened 2026-08-12
fix: bound multiprocess retrieve completion

### LINKED ISSUE #4516: [MP] Bound retrieve completion when a CUDA IPC event remains unsignaled
## Summary

`DeviceMessagingFuture.query()` can wait indefinitely after importing a CUDA IPC event if the event never becomes ready. The multiprocess vLLM connector has no deadline for this completion phase, so an affected request can remain in `WAITING_FOR_REMOTE_KVS` indefinitely and block scheduler progress.

## Code path

The future currently has three phases:

1. wait for the raw response/event handle;
2. import the IPC event;
3. poll `event.query()`.

Existing MQ timeout configuration bounds message transport, not phase 3. A synthetic future whose imported event always returns `False` from `query()` reproduces the unbounded wait.

## Safety constraint

A simple timeout followed by `finished_recving()`, invalidation, and recompute is unsafe for LMCache-driven zero-copy retrieval. The original H2D operation may still target the same paged-KV block IDs; releasing or reusing them before transfer terminality can allow late DMA to overwrite valid KV.

Transparent recovery therefore requires one of:

- cancel + positive terminal acknowledgement before block reuse;
- quarantine of timed-out block IDs until the event completes/cancellation is acknowledged; or
- a fail-stop policy that tears down the affected worker/context.

## Requested change

Please add a configurable MP retrieve-completion deadline (disabled by default for compatibility), plus:

- pending retrieve count and oldest pending age;
- timeout/failure counters labeled by phase (MQ response, IPC import, event completion);
- a safe timeout policy. An initial fail-stop policy is acceptable; transparent recompute must not release target blocks until terminality is established.

## Related but not equivalent

- vllm-project/vllm#49556: retrieve failures are not propagated/invalidated in the fallback connector
- LMCache#4419: ROCm deadlock while importing an IPC event
- LMCache#3336: AMD/HIP host-callback deadlock fix
- LMCache#2973: retrieve coordination improvements

Searches across open and closed LMCache/vLLM issues and PRs did not find an existing report matching an NVIDIA IPC event that remains imported but unsignaled indefinitely.

malaiwah:
## Summary

- add opt-in `lmcache.mp.retrieve_timeout` for MP retrieve completion
- track each retrieve from submission through device-event readiness
- raise `LMCacheTimeoutError` when the deadline expires
- fail closed without reporting target blocks reusable
- keep the default at `0` (disabled) for compatibility
- document the deadline and its late-DMA safety constraint

Fixes #4516.

## Safety

This intentionally does **not** call `finished_recving()`, invalidate blocks, or transparently recompute on expiry. LMCache currently has no transfer-cancellation acknowledgement or block-quarantine contract. Reusing paged-KV block IDs while an old H2D may still target them could allow late DMA to corrupt valid KV.

The initial deadline is therefore fail-stop: the error propagates out of the connector and aborts the affected worker. Transparent recovery remains follow-up work requiring cancellation + terminal acknowledgement, or allocation quarantine.

## Tests

- 6 focused CPU/static regression cases passed:
  - disabled-by-default behavior
  - expiry raises
  - expired transfers do not expose block IDs as load errors/reusable
  - a ready event wins at the deadline boundary
  - negative, infinite, and NaN values are rejected
- `ruff check` passed for changed Python files
- `ruff format --check` passed
- `py_compile` and `git diff --check` passed

A broader 32-test file run under a temporary no-native-extension test harness produced 31 passes; the one unrelated CPU-layout test requires the native `EngineKVFormat` extension and fails under that stub.

### DIFF
--- a/docs/source/mp/configuration.rst
+++ b/docs/source/mp/configuration.rst
@@ -565,6 +565,13 @@ All connector-level options are passed through
      - ``10.0``
      - Interval (seconds) between periodic heartbeat pings sent from the
        connector to the server.
+   * - ``lmcache.mp.retrieve_timeout``
+     - ``0.0``
+     - Maximum time (seconds) from retrieve submission to device-event
+       completion. ``0`` disables the deadline. When enabled, expiry raises
+       a timeout and aborts the worker rather than releasing target KV blocks:
+       the underlying transfer currently has no cancellation acknowledgement,
+       so reusing those blocks could allow a late DMA write to corrupt KV.
    * - ``lmcache.mp.mp_transfer_mode``
      - ``auto``
      - Routing mode for the worker -> server transfer context. One of
--- a/lmcache/integration/vllm/vllm_multi_process_adapter.py
+++ b/lmcache/integration/vllm/vllm_multi_process_adapter.py
@@ -5,8 +5,10 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Callable, NoReturn, Protocol
 import enum
+import math
 import os
 import threading
+import time
 import uuid
 
 # Third Party
@@ -27,6 +29,7 @@ from lmcache.v1.multiprocess.group_view import (
     EngineGroupInfo,
     expand_engine_block_ids,
 )
+from lmcache.v1.mp_observability.errors import LMCacheTimeoutError
 from lmcache.v1.multiprocess.mq import MessageQueueClient, MessagingFuture
 from lmcache.v1.multiprocess.protocol import RequestType, get_response_class
 from lmcache.v1.multiprocess.transfer_context import (
@@ -57,6 +60,11 @@ class ExtraConfigDefault(enum.Enum):
     # Interval (seconds) between periodic heartbeat pings
     # to the server.
     heartbeat_interval = 10.0
+    # Maximum time (seconds) from retrieve submission to device-event
+    # completion. Zero disables the deadline. A deadline raises instead of
+    # releasing target blocks because the underlying transfer cannot yet be
+    # cancelled or quarantined safely.
+    retrieve_timeout = 0.0
     # Routing mode for ``create_transfer_context``: ``auto`` keeps the
     # historical CUDA -> lmcache_driven / others -> engine_driven dispatch;
     # ``lmcache_driven`` forces the IPC / SHM zero-copy path where the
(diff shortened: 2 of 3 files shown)


malaiwah:
## Production reproduction and exact-version validation update

This PR now has a production confirmation from a GLM-5.2 / TP=4 LMCache MP service under sustained concurrent coding-agent load. Full details are also recorded on #4516.

### Incident

- Stall began at `2026-08-14 23:11:01Z`.
- `running=0`, `deferred=4`, useful prompt/generation/MTP counters frozen.
- Logical GPU KV use was `64.6%`, so this was not block-pool exhaustion.
- All four GPUs were at 0% utilization while `/health` and `/v1/models` remained 200.
- LMCache remained healthy: controller queues/inflight zero, `active_sessions=0`, and failure counters did not grow.
- No CUDA OOM, allocation failure, traceback, EngineCore/worker death, Xid, thermal throttle, or container restart occurred.
- A 30-second sample showed the native prefix-query counter growing at ~109.27M tokens/s while useful inference counters remained frozen, consistent with the scheduler polling an unresolved external retrieve.

The service self-recovered without restart after **88m40s**. Four deferred requests drained to one as two long-held chat requests ended; the last deferred request disappeared three seconds after a final chat request ended. Throughput and 94-100% GPU utilization resumed immediately. This strongly ties scheduler progress to orphaned `WAITING_FOR_REMOTE_KVS` entries being removed by request completion/cancellation rather than by normal connector completion.

Available metrics still cannot distinguish raw-response wait, IPC event import, and imported-event query. The controller-empty/session-zero state plus the nonblocking poll loop are consistent with the event-query phase, but phase telemetry is needed to prove that sub-phase.

### Exact deployed adapter test

In addition to the PR's existing focused tests, I rebased the timeout logic mechanically onto the exact adapter source deployed in the incident image and ran a separate compatibility suite inside a disposable, network-disabled CPU container made from that same image:

```text
......                                                                   [100%]
6 passed in 0.01s
```

Covered:

- default deadline remains disabled;
- an expired pending retrieve raises `LMCacheTimeoutError`;
- expired target blocks are not exposed as load errors/reusable;
- a ready retrieve wins at the deadline boundary;
- negative, infinite, and NaN values are rejected.

Current branch checks were repeated:

```text
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

The fail-stop policy remains intentional. vLLM calls `get_finished()` from the worker's connector-finalization path; the exception becomes a failed worker RPC and terminates EngineCore execution. It does not ACK `finished_recving` or release target blocks, avoiding late-DMA reuse corruption until LMCache has a cancellation/terminal-acknowledgement contract.

No production process, image, power cap, or configuration was changed during capture or validation.

PR #4520, opened 2026-08-12
[Bugfix][MP] Make lookup accounting retry-safe

### LINKED ISSUE #4508: [Bug][MP] num_stored_tokens double counted when the scheduler retries get_num_new_matched_tokens, later stores unreachable
## Environment

- LMCache dev @ 00f763c1
- vLLM 0.23.0

## Bug

`LMCacheMPConnector.get_num_new_matched_tokens` advances the tracker's stored-token watermark on every call (`lmcache/integration/vllm/lmcache_mp_connector.py:733`):

```python
ret = self.scheduler_adapter.check_lookup_result(request.request_id)
...
# Update num stored tokens for the tracker
tracker.increase_num_stored_tokens(ret)
```

The scheduler can call this more than once for the same waiting request. When `allocate_slots` returns `None` (vllm `v1/core/sched/scheduler.py:774`, "The request cannot be scheduled"), the request stays in the waiting queue with `num_computed_tokens == 0` and the next scheduling step queries the connector again. The cached lookup result is only cleared by `cleanup_lookup_result` inside `update_state_after_alloc`, which has not run at that point, so `check_lookup_result` returns the same cached hit and the watermark compounds: after n attempts, `num_stored_tokens == n * ret`.

## Consequence

Store metadata is produced from the inflated watermark (`GetStoreMetadata` starts at `num_stored_tokens`), so the token range `[ret, n * ret)` is never stored. Retrieval stops at the first missing chunk, so every chunk the request stores past the hole is unreachable. The trigger is a lookup hit combined with an allocation failure, which is exactly the memory-pressure regime.

## Repro

Two tests against dev @ 00f763c1, run with vLLM 0.23.0 installed. Both assert the correct behavior and fail:

```
E  AssertionError: belief corrupted: num_stored_tokens 1024 != server truth 512
E  AssertionError: store starts at 1024, leaving [512, 1024) never stored: all later chunks are unreachable
```

<details>
<summary>test_issue_num_stored_tokens_double_count.py</summary>

```python
"""Repro for the num_stored_tokens double count.

Both tests assert the correct behavior; a failure demonstrates the bug.
"""

# Standard
from types import SimpleNamespace

# Third Party
import pytest

pytest.importorskip("vllm", reason="MP connector imports vLLM at module top")

# Third Party
from vllm.v1.request import RequestStatus  # noqa: E402

# First Party
from lmcache.integration.vllm.lmcache_mp_connector import (  # noqa: E402
    LMCacheMPConnector,
)
from lmcache.integration.vllm.lmcache_mp_metadata import (  # noqa: E402
    LMCacheMPRequestMetadata,
)

CHUNK = 256
TOKENS_PER_BLOCK = 16


class _FakeLookupAdapter:
    """Mimics LMCacheMPSchedulerAdapter lookup caching semantics.

    check_lookup_result returns the cached aggregate on every call after the
    first: production caches results in _finished_lookup_results and only
    cleanup_lookup_result (called from update_state_after_alloc) clears them.
    """

    lmcache_tokens_per_chunk = CHUNK

    def __init__(self, hit_tokens: int) -> None:
        self._hit = hit_tokens

    def maybe_submit_lookup_request(
        self, request_id: str, token_ids: list[int], cache_salt: str = ""
    ) -> None:
        pass

    def check_lookup_result(self, request_id: str) -> int | None:
        return self._hit


def _make_connector(hit_tokens: int) -> LMCacheMPConnector:
    connector = LMCacheMPConnector.__new__(LMCacheMPConnector)
    connector.request_trackers = {}
    connector.scheduler_adapter = _FakeLookupAdapter(hit_tokens)  # type: ignore
    connector._hit_alignment_tokens = TOKENS_PER_BLOCK
    return connector


def _make_request(num_tokens: int, request_id: str = "req-0") -> SimpleNamespace:
    return SimpleNamespace(
        request_id=request_id,
        cache_salt=None,
        all_token_ids=list(range(num_tokens)),
        status=RequestStatus.WAITING,
    )


def test_repeated_matched_tokens_calls_keep_stored_belief_consistent() -> None:
    hit = 2 * CHUNK  # 512-token LMCache hit
    connector = _make_connector(hit)
    request = _make_request(num_tokens=600)

    # First scheduling attempt: lookup hit, then allocate_slots fails.
    need, is_async = connector.get_num_new_matched_tokens(request, 0)
    assert need == hit and is_async

    tracker = connector.request_trackers["req-0"]
    assert tracker.num_stored_tokens == hit

    # Retry on the next step (same request, still num_computed_tokens == 0).
    need2, _ = connector.get_num_new_matched_tokens(request, 0)
    assert need2 == hit

    assert tracker.num_stored_tokens == hit, (
        "belief corrupted: num_stored_tokens "
        f"{tracker.num_stored_tokens} != server truth {hit}"
    )


def test_store_after_retry_starts_at_true_watermark() -> None:
    """Consequence: the first store op produced after the request finally
    schedules must start at the server-stored watermark (512). Starting
    beyond it leaves [512, 1024) never stored, and the prefix lookup stops
    there on retrieval, so every chunk stored after the hole is unreachable.
    """
    hit = 2 * CHUNK
    connector = _make_connector(hit)
    request = _make_request(num_tokens=1600)

    connector.get_num_new_matched_tokens(request, 0)
    connector.get_num_new_matched_tokens(request, 0)  # allocate_slots retry

    tracker = connector.request_trackers["req-0"]
    # Allocation finally succeeds; the whole prompt is scheduled.
    tracker.append_block_ids(([*range(1, 101)],))  # 100 blocks = 1600 tokens
    tracker.increase_num_scheduled_tokens(1600 - hit)
    tracker.num_vllm_hit_tokens = 0
    tracker.num_lmcache_hit_tokens = hit

    meta = LMCacheMPRequestMetadata.GetStoreMetadata(
        tracker, CHUNK, [TOKENS_PER_BLOCK]
    )
    assert meta is not None, "no store metadata produced at all (silent hole)"
    assert meta.op.start == hit, (
        f"store starts at {meta.op.start}, leaving [{hit}, {meta.op.start}) "
        "never stored: all later chunks are unreachable"
    )
```

</details>

## Suggested fix

Apply the lookup accounting exactly once per tracker lifetime: either move `increase_num_stored_tokens` into `update_state_after_alloc` (the commit point that already calls `cleanup_lookup_result`), or keep it in `get_num_new_matched_tokens` behind a once-per-tracker flag.

rob-9:
## Purpose

Fixes #4508.

When vLLM cannot allocate slots for a waiting request, the scheduler may call `get_num_new_matched_tokens` again before `update_state_after_alloc` clears the cached lookup result. The MP connector previously added that same result to `num_stored_tokens` on every call. This inflated the stored-token watermark and could make later store metadata skip an unstored token range, leaving subsequent chunks unreachable through prefix lookup.

This PR makes lookup accounting idempotent for the lifetime of a request tracker:

- adds `account_lookup_result` to apply the cached lookup result only once;
- uses that method when the MP connector processes lookup results; and
- adds regression coverage for repeated scheduler polling, repeated `update_state_after_alloc` calls, one-time cleanup of the cached lookup result, and the resulting token range in store metadata.

## Notes

The `_lookup_result_accounted` flag belongs to each request tracker and is initialized to `False` whenever a new tracker is created. It applies only to the initial cached lookup result; updates for newly stored chunks continue to use `increase_num_stored_tokens` normally.

The regression test uses a typed scheduler-adapter double with the production method signatures, so it exercises the connector retry and allocation flow without contacting a cache server.

Validated in a Linux Codespace with vLLM 0.23.0 against the issue's reported commit, `00f763c`. The file `tests/v1/test_issue_num_stored_tokens_double_count.py` below is the repro script from #4508 and is not committed in this PR.

Before the fix:

```bash
python -m pytest --confcutdir=tests/v1 -q \
  tests/v1/test_issue_num_stored_tokens_double_count.py
```

```text
FAILED test_repeated_matched_tokens_calls_keep_stored_belief_consistent
E   AssertionError: belief corrupted: num_stored_tokens 1024 != server truth 512

FAILED test_store_after_retry_starts_at_true_watermark
E   AssertionError: store starts at 1024, leaving [512, 1024) never stored:
E   all later chunks are unreachable

2 failed, 58 warnings in 17.34s
```

After cherry-picking this fix onto `00f763c`, both the issue reproduction and the regression tests added by this PR pass:

```bash
python -m pytest --confcutdir=tests/v1 -q \
  tests/v1/test_issue_num_stored_tokens_double_count.py \
  tests/v1/test_mp_connector_lookup_retry.py
```

```text
tests/v1/test_issue_num_stored_tokens_double_count.py::test_repeated_matched_tokens_calls_keep_stored_belief_consistent PASSED
tests/v1/test_issue_num_stored_tokens_double_count.py::test_store_after_retry_starts_at_true_watermark PASSED
tests/v1/test_mp_connector_lookup_retry.py::test_lookup_retry_accounts_for_stored_watermark_once PASSED
tests/v1/test_mp_connector_lookup_retry.py::test_store_after_lookup_retry_starts_at_true_watermark PASSED

4 passed, 58 warnings in 9.26s
```

## Checklist

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -707,7 +707,8 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
             tokens for which KV cache is actually available at the time of the
             call. If the cache cannot be loaded for some tokens (e.g., due to
             connectivity issues or eviction), those tokens must not be taken
-            into account.
+            into account. This method may be called repeatedly before block
+            allocation, so stored-token accounting must be idempotent.
         """
         tracker = self._get_or_create_request_tracker(request)
         # TODO: support loading KV for preempted requests in the future
@@ -729,8 +730,10 @@ class LMCacheMPConnector(KVConnectorBase_V1, SupportsHMA):
 
         assert ret % self.scheduler_adapter.lmcache_tokens_per_chunk == 0
 
-        # Update num stored tokens for the tracker
-        tracker.increase_num_stored_tokens(ret)
+        # The scheduler may retry this method while allocation is blocked and
+        # the adapter will return the same cached lookup result. Account for
+        # that result only once during the lifetime of this tracker.
+        tracker.account_lookup_result(ret)
 
         # Save the vllm and lmcache hit tokens. The vLLM hit count is
         # rounded down to a boundary aligned for every engine group (e.g.
--- a/lmcache/integration/vllm/lmcache_mp_metadata.py
+++ b/lmcache/integration/vllm/lmcache_mp_metadata.py
@@ -60,6 +60,9 @@ class LMCacheMPRequestTracker:
     # requests.
     num_stored_tokens: int = 0
 
+    # Whether the cached lookup result has been included in num_stored_tokens.
+    _lookup_result_accounted: bool = False
+
     # Staging load operation -- save vllm and lmcache hit tokens during lookup
     num_vllm_hit_tokens: int = 0
     num_lmcache_hit_tokens: int = 0
@@ -77,6 +80,7 @@ class LMCacheMPRequestTracker:
         self.all_token_ids = request.all_token_ids
         self.allocated_block_ids = {}
         self.num_stored_tokens = 0
+        self._lookup_result_accounted = False
         self.num_vllm_hit_tokens = 0
         self.num_lmcache_hit_tokens = 0
         self.state = LMCacheMPRequestState.PREFETCHING
@@ -111,6 +115,24 @@ class LMCacheMPRequestTracker:
     ####
     def increase_num_scheduled_tokens(self, num_new_tokens: int):
         self.num_scheduled_tokens += num_new_tokens
+
+    def account_lookup_result(self, num_stored_tokens: int) -> None:
+        """Include an external lookup result in the stored-token watermark.
+
+        Args:
+            num_stored_tokens: Number of prefix tokens found by the lookup.
+
+        Returns:
+            None.
+
+        Notes:
+            Only the first call affects the watermark. Later calls are ignored
+            because the scheduler may poll the same cached result repeatedly
+            before it can allocate request blocks.
+        """
+        if not self._lookup_result_accounted:
+            self.increase_num_stored_tokens(num_stored_tokens)
+            self._lookup_result_accounted = True
 
     def increase_num_stored_tokens(self, num_new_tokens: int):
         """Increase the number of stored tokens for the current request
--- a/tests/v1/test_mp_connector_lookup_retry.py
+++ b/tests/v1/test_mp_connector_lookup_retry.py
@@ -1,0 +1,126 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for retry-safe MP connector lookup accounting."""
+
+# Standard
+from types import SimpleNamespace
+
+# Third Party
+import pytest
+
+pytest.importorskip("vllm", reason="MP connector imports vLLM at module scope")
+
+# Third Party
+from vllm.v1.request import RequestStatus  # noqa: E402
+
+# First Party
+from lmcache.integration.vllm.lmcache_mp_connector import (  # noqa: E402
+    LMCacheMPConnector,
+)
+from lmcache.integration.vllm.lmcache_mp_metadata import (  # noqa: E402
+    LMCacheMPRequestMetadata,
+)
+
+CHUNK_SIZE = 256
+TOKENS_PER_BLOCK = 16
+
+
+class _LookupAdapter:
+    """Test adapter exposing the scheduler lookup contract."""
+
+    lmcache_tokens_per_chunk = CHUNK_SIZE
+
+    def __init__(self, hit_tokens: int) -> None:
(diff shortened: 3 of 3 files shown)


DongDongJu (changes requested):
Hello @rob-9, Thanks for the patch.
I encounter same issue during dynamo kvcache aware routing work.

However, lookup results are not guaranteed to remain identical: after a positive cached result, check_lookup_result() can return 0 when server health drops. 

In that case the previous watermark remains accounted and a locally computed request can still STORE after a hole. 

Please make the accounting reflect the latest non-None result, including a downgrade to zero, and add a hit -> 0 before allocation regression test. 

Also, the rebased test fixture needs updating for request_configs and _connector_stats. After those changes and green CI, this should be in good shape.

Please let me know If I understand incorrectly.

rob-9:
@DongDongJu thanks for the review.  I've updated the accounting to use latest lookup result, including when a hit drops to zero. That also clears the stale hit counts, so requests that fall back to local computation won’t skip unstored ranges. Also updated the fixtures for `request_configs` and `_connector_stats`.

DongDongJu (approved):
LGTM

DongDongJu:
@maobaolong @ApostaC PTAL. its accounting bug for mp

PR #4521, opened 2026-08-12
Refactor memory allocator package exports

maobaolong:
## Summary
- replace the handwritten allocator export map with AST-based discovery over `*_allocator.py`
- keep lazy package exports and add a regression test that new allocator modules are exported automatically
- document that full `pre-commit` runs require the Rust toolchain on PATH

## Testing
- `/Users/mbl/projects/LMCache/.venv/bin/pytest -xvs tests/v1/test_memory_allocator_import_surface.py`
- `/Users/mbl/projects/LMCache/.venv/bin/pre-commit run --all-files` *(fails in this environment only because `cargo`/`rustfmt` are not installed; Python hooks passed and commit-time hooks passed for changed files)*

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,6 +84,11 @@ mypy --config-file=pyproject.toml   # Type checking
 codespell --toml pyproject.toml     # Spell checking
 ```
 
+Before running `pre-commit run --all-files`, verify that `cargo`, `rustfmt`, and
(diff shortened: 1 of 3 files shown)


maobaolong:
Closing this PR to resubmit the same change on a normally named branch and without the AGENTS.md update.

PR #4523, opened 2026-08-12
[MP][Bugfix] free_lookup_locks: release only the keys the prefetch actually locked

YuhanLiu11:

`free_lookup_locks` releases the read locks that a lookup's prefetch acquired but
the engine will never retrieve (e.g. chunks already covered by vLLM's local prefix
cache). It rebuilt the key set as the full cross product of
`chunks-in-range × all object groups` — but the prefetch (via `unfold`) locks
sliding-window object groups only on the trailing `[max(0, H - window), H)` chunks
of the model-wide hit `H`, not on the whole prefix.

On hybrid-attention models with `--separate-object-groups` this made every warm
request call `finish_read` on keys the prefetch never locked. Observed with
Kimi-K3 (windows `[1, -1]`, TP=8, chunk size 768): a single 167-chunk warm request
emitted 1328 warnings —

    L1Manager: finish read on non-existing key ... (1200×)
    L1Manager: finish read on non-read-locked key ... (128×)

Beyond the log noise, releasing a lock you don't hold is unsafe under concurrency:
if another in-flight request legitimately holds a read lock on one of those keys
(overlapping-prefix workloads), the spurious `finish_read` strips that reader's
lock and its data can be evicted mid-read.

Fix — make the release side mirror `unfold`'s locking contract:

- `Session` gains `prefetch_hit_chunks`, recorded by `query_prefetch_status` when
  the prefetch result is consumed (which all connector free paths do first).
- `free_lookup_locks` now releases, per object group, exactly
  `locked(group) ∩ [start, end)`:
  - full attention: `[0, H)`
  - sliding window: `[max(0, H - window), H)`
- Full-attention releases are additionally capped at `H` (keys past the hit were
  never locked).
- Defensive fallback when no hit was recorded (free before any status query):
  full-attention groups keep the legacy full-range release; sliding-window groups
  release nothing — a leaked read lock expires with its TTL, an over-release could
  corrupt a concurrent reader.

This also resolves the pre-existing NOTE in `free_lookup_locks` about SWA
over-release (`window=512, LMCache hit 1024, vLLM hit 768`): the window group now
releases `[512, 768)` while the retrieve releases `[768, 1024)` — no leak, no
double release.

Verified end-to-end with Kimi-K3 (TP=8) + the long-doc-qa benchmark: warm request
retrieved 122k tokens/rank with zero `L1Manager` finish-read warnings (previously
1328 per request).

- The locked-range formula intentionally duplicates `unfold`'s semantics
  (`lmcache/v1/distributed/bitmap_ops/fold.py`) analytically instead of storing
  the retain bitmap; the hit length is the only state carried over, on the
  already-per-request `Session`.
- `query_prefetch_status` is the single point where the hit is recorded;
  `wait_prefetch_status` delegates to it, so both polling paths are covered.

**If applicable**:

### DIFF
--- a/lmcache/v1/multiprocess/modules/lookup.py
+++ b/lmcache/v1/multiprocess/modules/lookup.py
@@ -406,6 +406,14 @@ class LookupModule:
             job.attn_desc.num_chunks_in_sw,
         )
 
+        # Record the model-wide hit length on the session so a later
+        # free_lookup_locks can reconstruct which keys the prefetch
+        # read-locked (see ``unfold``: full-attention groups lock the whole
+        # hit prefix, sliding-window groups only its in-window suffix).
+        self._ctx.session_manager.get_or_create(
+            job.request_id
+        ).prefetch_hit_chunks = found_count
+
         self._ctx.event_bus.publish(
             Event(
                 event_type=EventType.MP_LOOKUP_PREFETCH_END,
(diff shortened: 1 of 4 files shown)


ApostaC (approved):
LGTM

deng451e (approved):
LGTM

PR #4526, opened 2026-08-13
[Coordinator][Refactor]: cache-event ingest layer and fleet controllers

ruizhang0101:

Three seams in the mp coordinator were folded together; this splits them.

- The key directory was both the placement store *and* the fencing/dedup gate every cache event had to pass through. It becomes an ordinary `CacheEventConsumer`; admission moves to a new `ingest/` layer (`EventGate` + `CacheEventBroadcaster`).
- `cache_control/` + the eviction manager become `controllers/`. `FleetEvictionController` owns the quota registry, the usage view, and its own control loop — previously a `while True` in `create_app`'s lifespan.
- `POST /directory/events` → `POST /events`, in its own `events_api.py`: the stream feeds every consumer of it, not just the directory.

Also drops the startup L2 resync, so `POST /events` is the coordinator's single ingest path.

Two behavior changes, both deliberate:

1. **Endpoint rename is a breaking wire change.** Both sides move together in-tree (`HttpCacheEventSink` posts the new path), but a fleet upgraded piecemeal will silently stop reporting — old MP servers get a 404 and the sink drops the batch.

2. **Startup resync is removed**, along with `LMCACHE_MP_COORDINATOR_ENABLE_STARTUP_RESYNC`, `RESYNC_POLL_INTERVAL`, `RESYNC_MAX_WAIT`, and `RESYNC_PAGE_SIZE`. A restarted coordinator now rebuilds usage purely from the event stream, so per-salt totals start at zero and climb rather than being seeded from a live server's `GET /cache/objects`. Unquota'd salts stay exempt until the quota controller arms a default limit, so a cold coordinator still cannot mass-evict — but it will under-evict until the stream catches up. This removes the gate's second, cursor-free `reconcile()` door; event replay from a durable transport remains the intended answer for gap recovery.

`GET /directory/stats` keeps its exact response shape; the per-instance row is now joined at the API layer from the gate's stream cursor and the directory's L1 key count.

**If applicable**:


lmcache/v1/mp_coordinator/controllers/eviction_controller.py
```diff
@@ -1,11 +1,16 @@
...
-server. LRU bookkeeping is updated when the corresponding ``delete``
-event arrives back on the fleet cache-event stream
-(``POST /directory/events``).
+"""Fleet-wide per-``cache_salt`` L2 eviction control loop.
+
+Holds a target (:class:`QuotaManager` budgets), observes the value
+(:class:`L2UsageManager` byte totals), and dispatches
+``DELETE /cache/objects`` for the LRU's victims to close the gap. Both
+halves are maintained from the cache-event stream the ingest layer
+admits, so accounting and dispatch cannot diverge.
```
ruizhang0101: remove this

lmcache/v1/mp_coordinator/controllers/eviction_controller.py
```diff
@@ -206,24 +210,46 @@ def compute_eviction_plan(self) -> dict[str, list[ObjectKey]]:
...
+        registry: InstanceRegistry,
+        http_client: httpx.AsyncClient,
+        check_interval: float,
+    ) -> None:
+        """Run the control loop until cancelled.
+
+        Sleeps first, so construction never races the first check. Safe
+        to start immediately: unquota'd salts are exempt until the quota
+        controller arms a default limit, so a cold table cannot
+        mass-evict.
```
ruizhang0101: remove this

lmcache/v1/mp_coordinator/http_apis/directory_api.py
```diff
@@ -214,11 +190,33 @@ def _scan() -> DirectoryListResponse:
...
     """Return a point-in-time summary of directory contents.
 
+    Joins the two halves of per-instance state: the ingest gate's stream
+    cursor and the directory's L1 key count.
+
     Returns:
         Key/placement counts plus per-instance stream state (incarnation,
-        last applied seq, gap flag), keyed by ``instance_id``.
+        last applied seq, gap flag, L1 key count), keyed by
+        ``instance_id`` — an instance appears if either half knows it.
```
ruizhang0101: the stats doesn't need to return stream cursor for now, just directory stats. In this case, can the _NO_STREAM be removed?

lmcache/v1/mp_coordinator/http_apis/events_api.py
```diff
@@ -0,0 +1,47 @@
...
+    emission order. Duplicates and stale incarnations are dropped and
+    counted, not errors.
+
+    Args:
+        body: The event batches to ingest.
+
+    Returns:
+        Counts of applied and dropped batches.
+    """
+    gate = get_context(request).event_gate
```
ruizhang0101: use event_gate instead of gate

lmcache/v1/mp_coordinator/http_apis/quota_api.py
```diff
@@ -187,8 +194,8 @@ async def list_status(request: Request, tier: Tier = Tier.L2) -> StatusListRespo
     """
     _require_supported_tier(tier)
     ctx = get_context(request)
-    usage_view = ctx.usage_manager
-    store = ctx.quota_manager
+    usage_view = ctx.eviction_controller.usage
+    store = ctx.eviction_controller.quota
```
ruizhang0101: store is ambiguous.

ruizhang0101 (commented):
PTAL

lmcache/v1/mp_coordinator/app.py
```diff
@@ -129,65 +116,42 @@ async def _health_loop() -> None:
...
             health_task = asyncio.create_task(_health_loop())
         if config.eviction_check_interval > 0:
-            eviction_task = asyncio.create_task(_eviction_loop(outbound_client))
-        if config.enable_startup_resync:
-            resync_task = asyncio.create_task(_startup_resync(outbound_client))
+            eviction_task = asyncio.create_task(
+                eviction_controller.run(
+                    registry, outbound_client, config.eviction_check_interval
+                )
+            )
```
ApostaC, line 136: Please correct me if I'm wrong. `outbound_client` here is for sending actions to LMCache servers, right?

In the future, we may also want to decouple this into an "action execution" module. Where the actions are standardized, and the controller does not need to care about how to send the action and which LMCache instance to send it to.

ApostaC (commented):
Like this refactor very much. Just one comment. Otherwise LGTM

ApostaC (approved):

sammshen (approved):
LGTM!

PR #4528, opened 2026-08-13
refactor: drop memory_allocators package exports

maobaolong:
## Summary
- remove the package-level export surface from `lmcache.v1.memory_allocators`
- keep the package importable for submodule imports by retaining only the package marker file
- trim import-surface tests so they cover the remaining submodule and `memory_management` contracts

## Testing
- `/Users/mbl/projects/LMCache/.venv/bin/pytest -xvs tests/v1/test_memory_allocator_import_surface.py`
- commit-time `pre-commit` hooks via `git commit`


ApostaC (approved):
I think this looks good to me.

@DongDongJu IIRC, maybe it's introduced by you (I could be wrong)? If so, please take a quick look at this as well. Thanks!

DongDongJu:
Yes I did and made the PR for this https://github.com/LMCache/LMCache/pull/4598 yesterday too.

DongDongJu (approved):
It looks very similar what I made.
either works for me.
LGTM

PR #4540, opened 2026-08-13
[Bugfix] Fix logging calls that drop their message at runtime, and guard against regressions

### LINKED ISSUE #4539: [Bug] Logging calls pass more args than their format string consumes, so the message is dropped at runtime
### Describe the bug

```python
except Exception as e:
```

```python
logger.warning(
)
```

```python
logger.warning(
)
```

### To Reproduce

```python
import logging, sys

```

```
--- Logging error ---
  ...
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
```

```
Found 2 errors.
```

### Environment

- LMCache `dev` @ `0373a573`

alooshxl:

Fixes #4539

Three `logger.*` calls passed more positional arguments than their format string consumes. `logging` renders a record lazily as `msg % args`, so each of these raises `TypeError: not all arguments converted during string formatting` *inside* the logging machinery. Python's logging swallows that and prints a `--- Logging error ---` traceback to stderr instead — **the intended message is never emitted**. All three sit on failure paths, which is exactly where losing the message hurts most.

**1. `p2p_backend.py`** — two arguments, one placeholder. The caught exception had no placeholder to receive it, so the reason a peer socket failed to close was discarded entirely:

```python
 except Exception as e:
-    logger.error("Failed to close peer %s lookup socket", self.peer_init_url, e)
+    logger.error(
+        "Failed to close peer %s lookup socket: %s", self.peer_init_url, e
+    )
```

**2 & 3. `mock_connector.py`** — a stray comma where implicit string concatenation was intended. This both truncated the message ("Failed to allocate memory even with") and turned its tail into an argument. Rewritten as `%`-format, matching the convention in #3372:

```python
 logger.warning(
-    "Failed to allocate memory even with",
-    f" busy loop on {i} out of {len(mock_objs)} objects",
+    "Failed to allocate memory even with busy loop on %d out of "
+    "%d objects",
+    i,
+    len(mock_objs),
 )
```

**Preventing regressions.** `PLE1205`/`PLE1206` were not in the ruff `select` list, so nothing in CI catches this. Enabling just those two rules reports **exactly the two static cases fixed here and nothing else repo-wide**, so they turn on with no extra cleanup:

```
All checks passed!     # after this PR; 2 errors before it
```

I deliberately left `G` (flake8-logging-format) commented out — it currently reports 511 findings, mostly the f-string logging already being migrated file-by-file.

Ruff cannot see through an f-string, so case 2 above stays invisible to `PLE1205`. `tests/test_logging_format.py` closes that gap with a small AST walk over `lmcache/`. It is stdlib-only, runs in <1s, and reports every offending call in one assertion:

```
AssertionError: Logging calls would raise TypeError at runtime:
  lmcache/v1/storage_backend/connector/mock_connector.py:266 expects 0 arg(s) but 1 were supplied: 'Mock object is None on '
  lmcache/v1/storage_backend/connector/mock_connector.py:278 expects 0 arg(s) but 1 were supplied: 'Failed to allocate memory even with'
  lmcache/v1/storage_backend/p2p_backend.py:155 expects 1 arg(s) but 2 were supplied: 'Failed to close peer %s lookup socket'
```

- Verified the test actually catches the bugs: it fails with all three findings above when the two source fixes are reverted, and passes with them applied.
- The checker is intentionally conservative to avoid false positives on unrelated code — it skips `*args` splats, `%(name)s` mapping style, and `*` width/precision, and only treats a receiver whose name contains "log" as a logger.
- Ran locally on `dev` @ `0373a573`: `ruff check` (clean repo-wide with the new rules), `ruff format --check` (1177 files already formatted), `isort`, `codespell`, the SPDX header hook, and `mypy` on the new file — all clean.
- Happy to split the `pyproject.toml` rule change into its own PR if you would rather keep the fix and the lint change separate.

**If applicable**:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,10 @@ select = [
     # "I",
     # flake8-logging-format
     #"G",
+    # pylint: logging call passes too many / too few args for its format string.
+    # Such a call raises TypeError inside logging and the message is dropped.
+    "PLE1205",
+    "PLE1206",
     # flake8-self (private member access)
     "SLF",
 ]
--- a/tests/test_logging_format.py
+++ b/tests/test_logging_format.py
@@ -1,0 +1,145 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Guards logging calls in ``lmcache/`` against format-string/argument mismatches.
+
+``logging`` renders a record lazily as ``msg % args``. When the two disagree the
+formatting raises ``TypeError`` *inside* the logging machinery, so the intended
+message is replaced by a ``--- Logging error ---`` traceback on stderr. These
+calls sit on failure paths, which is exactly where losing the message hurts most.
+
+Ruff's ``PLE1205``/``PLE1206`` cover the static-string case. They cannot see
+through an f-string, so ``logger.warning(f"a {x}", " and b")`` -- a stray comma
+where implicit string concatenation was intended -- stays invisible to the
+linter. This test closes that gap while the codebase migrates from f-string
+logging to %-format.
+"""
+
+# Standard
+from pathlib import Path
+from typing import List, Optional, Tuple
+import ast
+import re
+
+LMCACHE_ROOT = Path(__file__).resolve().parents[1] / "lmcache"
+
+LOG_METHODS = frozenset(
+    {"debug", "info", "warning", "warn", "error", "exception", "critical", "fatal"}
+)
+
+# Excluded from linting in pyproject.toml; generated vLLM connector shims.
+EXCLUDED = ("lmcache_mp_connector_",)
+
+# %[(key)][flags][width][.precision][length]conversion
+PERCENT_SPEC = re.compile(
+    r"%(?:\((?P<key>[^)]*)\))?"
+    r"[#0\- +]*"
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -87,6 +87,10 @@ select = [
     # "I",
     # flake8-logging-format
     #"G",
+    # pylint: logging call passes too many / too few args for its format string.
+    # Such a call raises TypeError inside logging and the message is dropped.
+    "PLE1205",
```
zhengfeihe, line 92: This is kinda too specific, why not just 

"PLE"

tests/test_logging_format.py
```diff
@@ -0,0 +1,145 @@
+# SPDX-License-Identifier: Apache-2.0
```
zhengfeihe, line 1: Thanks for writing the tests! But I don't think we need a extra tests for this. It's kinda over-testing from my perspective, can you remove it ? CI is already pretty heavy now I think

zhengfeihe (commented):
Thanks for the fix~ Please check comments.

PR #4547, opened 2026-08-13
[Bugfix] disk: release accounted disk space on external removals

### LINKED ISSUE #4546: [Bug] LocalDiskBackend leaks disk-space accounting on every external remove, eventually dropping all disk puts silently
**Label**

bug

**Describe the bug**

`LocalDiskBackend` tracks how much of its disk budget is in use with
`self.current_cache_size`, but only its *internal* eviction loop ever decrements
that counter. Every other removal path shrinks `self.usage` and drops the entry
from `self.dict` while leaving `current_cache_size` untouched, so the counter
only ever grows. Once the accumulated phantom usage exceeds
`max_local_disk_size`, `submit_put_task` silently drops every subsequent disk
put — even though the disk is empty.

On `dev` (0373a573) `current_cache_size` has exactly four references, all in
`lmcache/v1/storage_backend/local_disk_backend.py`:

| Line | Code | Note |
|---|---|---|
| 175 | `self.current_cache_size = 0.0` | init |
| 346 | `while self.current_cache_size + required_size > self.max_cache_size:` | the capacity gate in `submit_put_task` |
| 358 | `self.current_cache_size -= self.dict[evict_key].size` | decrement — internal eviction only |
| 364 | `self.current_cache_size += required_size` | increment — every accepted put |

`remove()` (lines 252-286) pops the entry, decrements `self.usage`, unlinks the
file and updates the cache policy, but never touches `current_cache_size`:

```python
def remove(self, key: CacheEngineKey, force: bool = True) -> bool:
    lock_context = self.disk_lock if force else nullcontext()
    with lock_context:
        if not (meta := self.dict.pop(key, None)):
            return False
        path = meta.path
        size = meta.size
        self.usage -= size                                   # <-- usage shrinks
        self.stats_monitor.update_local_storage_usage(self.usage)
        os.remove(path)
        if force:
            self.cache_policy.update_on_force_evict(key)
    ...
    return True
```

The internal eviction loop gets away with this because it deducts the chunk
itself (line 358) before calling `self.batched_remove(evict_keys, force=False)`.
Every removal that originates *outside* the backend uses the default
`force=True` and is never accounted for.

**Trigger paths**

`StorageManager.remove` (`storage_manager.py:1069`) and
`StorageManager.batched_remove` (`storage_manager.py:1095`) both call into the
backend with the default `force=True`, so all of these leak:

- `LMCacheEngine.clear()` → `_clear()` → `storage_manager.remove(...)`
  (`cache_engine.py:1599`)
- `LMCacheEngine.move()` → `storage_manager.batched_remove(keys,
  locations=[old_position])` (`cache_engine.py:1313`) — moving a chunk off disk
  leaks the space it used to occupy
- `LMCacheEngine.compress()` / `decompress()` → `storage_manager.batched_remove(
  keys, locations=[location])` (`cache_engine.py:1472`, `:1533`)
- the remove-after-retrieve path used by the PD-disaggregation receiver role
  (`cache_engine.py:928` and `:1805`), when the disk backend is among
  `retrieve_locations`

**Consequence**

The drift is monotonic and never recovers within the process lifetime. Once
`current_cache_size` passes `max_local_disk_size`, the `while` loop in
`submit_put_task` asks the cache policy for eviction candidates, gets none back
(the dict really is empty, or holds far less than the counter claims), logs

```
No eviction candidates found. Disk space under pressure.
```

at WARNING level, and returns without storing anything. The disk tier then
degrades to a silent no-op for the rest of the process: puts stop landing, hit
rate collapses, and the only signal is a repeated warning that points at disk
pressure while the disk is in fact nearly empty. The reported `usage` metric
stays correct throughout, so monitoring shows an empty disk cache that refuses
to accept data.

**Expected behavior**

Removing a chunk should return its space to the budget, so that puts keep being
accepted for as long as the disk actually has room.

**Reproduction**

A standalone model of the four counter sites above (no torch required) makes the
drift visible. Chunk = 128 KiB, budget = 2.5 chunks:

```
A) store 2, remove both externally, store 1 more
   pre-fix: disk holds 0 chunk(s), usage=0 B, current_cache_size=262144 B, next put DROPPED
   fixed  : disk holds 1 chunk(s), usage=131072 B, current_cache_size=131072 B, next put ACCEPTED

B) repeated store/remove cycles on an empty disk
   pre-fix: first dropped put at cycle 2, current_cache_size=262144 B while usage=0 B
   fixed  : first dropped put at never, current_cache_size=0 B while usage=0 B
```

Scenario B is the important one: the disk is empty on every cycle
(`usage=0`), yet after two store/remove rounds the backend refuses all
further puts.

**Suggested fix**

Deduct the chunk in `remove()` when `force=True`, mirroring the existing
`if force:` guard that already distinguishes external removals from the
internal eviction path (which has already done the deduction itself).

**Related, not part of this report**

`submit_put_task` does not skip keys that are already in `self.dict` (unlike
`LocalCPUBackend`), so re-putting an existing key charges the budget twice while
`insert_key` only refreshes recency. That is a second, independent source of
upward drift and should be fixed separately.

alooshxl:

Fixes #4546

`LocalDiskBackend` reserves disk budget in `submit_put_task` and charges it to
`self.current_cache_size`, but only its own eviction loop ever gave that space
back. On `dev` the counter has exactly four references, all in
`local_disk_backend.py`: it is zeroed at init (175), read by the capacity gate
(346), decremented by the internal eviction loop (358), and incremented by every
accepted put (364). `remove()` is not among them — it shrank `usage` and popped
the entry from `self.dict`, but left `current_cache_size` inflated forever.

The internal eviction loop gets away with that because it deducts the chunk
itself before delegating:

```python
for evict_key in evict_keys:
    self.current_cache_size -= self.dict[evict_key].size

self.batched_remove(evict_keys, force=False)
```

Every removal that originates *outside* the backend takes the default
`force=True` and was never accounted for. `StorageManager.remove` and
`StorageManager.batched_remove` both call the backend with that default, so
`LMCacheEngine.clear()`, `move()`, `compress()`, `decompress()` and the
remove-after-retrieve path all leak budget on each call.

Because the drift is monotonic, `current_cache_size` eventually exceeds
`max_local_disk_size`. From then on the `while` loop in `submit_put_task` asks
the cache policy for a victim, gets nothing back (the dict genuinely is empty,
or holds far less than the counter claims), logs

```
No eviction candidates found. Disk space under pressure.
```

and returns without storing. **The disk tier silently degrades to a no-op for
the rest of the process** — puts stop landing, hit rate collapses, and the only
signal is a warning blaming disk pressure while the disk is nearly empty. The
`usage` metric stays accurate the whole time, so monitoring shows an empty disk
cache that refuses to accept data.

**The fix** — deduct the chunk in `remove()` when `force=True`:

```python
             self.usage -= size
             self.stats_monitor.update_local_storage_usage(self.usage)

+            # Release the space this chunk reserved in `submit_put_task`.
+            # Deducted next to `usage` and before the fallible `os.remove` so
+            # both counters stay consistent with `dict` on every exit path.
+            # Skipped when force=False: the internal eviction path already
+            # deducted the chunk before calling us.
+            if force:
+                self.current_cache_size -= size
```

I considered the alternative of dropping the pre-deduction in `submit_put_task`
and letting `remove()` own the decrement unconditionally. I went with the
`if force:` guard instead because it is strictly additive — the eviction loop is
untouched, so its interleaving of "deduct, remove, re-check the gate" inside the
held `disk_lock` cannot change. It also reuses the guard the function already
has for `update_on_force_evict`, which encodes exactly the same distinction
(external removal vs. internal eviction), so there is one consistent meaning of
`force` in the method rather than two.

On the error path: `os.remove(path)` can raise after the entry has already been
popped and `usage` already decremented. Putting the new deduction next to the
`usage` update, ahead of that call, means a failing unlink leaves `dict`,
`usage` and `current_cache_size` mutually consistent rather than re-introducing
the same drift on a rarer path. It also needs no `try`/`except`, so the change
stays a single guarded statement. (The orphaned file such a failure would leave
behind is a pre-existing, separate concern and is not touched here.)

I also added a docstring to `remove()` documenting the `force` contract, since
"the caller has already adjusted the accounting" is the invariant the fix hangs
on and nothing stated it before.

**I could not execute the test suite locally, and I want to be upfront about
that.** My machine is Windows, and `LocalDiskBackend.__init__` calls
`os.statvfs`, which is POSIX-only — so every test in
`tests/v1/storage_backend/test_local_disk_backend.py`, existing ones included,
is unrunnable here regardless of anything else. (`torch` is also absent: the
interpreter is Python 3.14 and the project supports `>=3.10,<3.14`, so
collection stops at `import torch` before it even gets that far.) Everything
below is what I *did* run, and I have not claimed a passing test run.

Static checks, all on the two changed files, run with the CI-pinned ruff 0.11.7:

| Command | Result |
|---|---|
| `python -m ruff check` | `All checks passed!` |
| `python -m ruff format --check` | `2 files already formatted` |
| `python -m isort --check-only` | clean, no output |
| `python -m mypy --ignore-missing-imports --follow-imports=silent --explicit-package-bases` | `Success: no issues found in 2 source files` |
| `python -m codespell_lib --toml pyproject.toml` | clean, no output |
| `python tools/check_spdx_header.py` | clean |

Since I could not run the new tests, I validated the underlying arithmetic with
a standalone pure-Python model of the four `current_cache_size` sites (written
outside the repo, not committed). Chunk = 128 KiB, budget = 2.5 chunks:

```
A) store 2, remove both externally, store 1 more
   pre-fix: disk holds 0 chunk(s), usage=0 B, current_cache_size=262144 B, next put DROPPED
   fixed  : disk holds 1 chunk(s), usage=131072 B, current_cache_size=131072 B, next put ACCEPTED

B) repeated store/remove cycles on an empty disk
   pre-fix: first dropped put at cycle 2, current_cache_size=262144 B while usage=0 B
   fixed  : first dropped put at never, current_cache_size=0 B while usage=0 B

C) internal eviction only (4 puts into a budget of 2)
   pre-fix: resident=['k2', 'k3'], current_cache_size=262144 B
   fixed  : resident=['k2', 'k3'], current_cache_size=262144 B
```

B is the failure users would hit: the disk is empty every cycle (`usage=0`),
yet after two store/remove rounds the backend refuses all further puts. C is
the control — the internal eviction path behaves identically before and after,
confirming the `force=False` case is not double-counted.

**The two new tests** go in the existing
`tests/v1/storage_backend/test_local_disk_backend.py` and assert behaviour
through the public API only — no reads of `current_cache_size`:

- `test_put_accepted_after_stored_keys_are_removed` fills the budget, removes
  everything via `batched_remove`, then stores a fresh chunk and asserts
  `contains()` reports it. Pre-fix that put is dropped, so the assertion fails.
- `test_eviction_keeps_only_the_newest_chunks` stores four chunks into a budget
  of two and asserts the two oldest are gone and the two newest are resident.
  This is the guard against the opposite error — releasing space twice would let
  the cache grow past its budget.

Both size the backend from the real `get_physical_size()` of an allocation, so
the budget stays exact whatever the allocator's alignment is, and both drive
`submit_put_task`'s `on_complete_callback` through a `threading.Event` rather
than sleeping: the callback only fires for puts the backend accepted, which is
precisely the signal this bug destroys.

**Follow-up, deliberately not in this PR**: unlike `LocalCPUBackend`, whose
`submit_put_task` returns early when the key is already in `hot_cache`,
`LocalDiskBackend.submit_put_task` does not skip keys already present in
`self.dict`. Re-putting an existing key charges the budget a second time while
`insert_key` only refreshes recency, which is an independent source of the same
upward drift. I kept this PR to the remove-path accounting; happy to file that
separately.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/local_disk_backend.py
+++ b/lmcache/v1/storage_backend/local_disk_backend.py
@@ -254,6 +254,20 @@ class LocalDiskBackend(StorageBackendInterface):
         key: CacheEngineKey,
         force: bool = True,
     ) -> bool:
+        """
+        Remove a cached chunk from disk and release its accounted space.
+
+        :param key: The cache key identifying the chunk to remove.
+        :param force: ``True`` for removals initiated outside the backend
+            (the caller does not hold ``disk_lock`` and has not adjusted the
+            disk space accounting). ``False`` is reserved for the internal
+            eviction path in ``submit_put_task``, which already holds
+            ``disk_lock`` and has already deducted the chunk from
+            ``current_cache_size`` before calling this method.
+
+        :return: ``True`` if the key was present and removed, ``False`` if
+            it was not cached.
+        """
         lock_context = self.disk_lock if force else nullcontext()
         with lock_context:
             if not (meta := self.dict.pop(key, None)):
@@ -263,6 +277,14 @@ class LocalDiskBackend(StorageBackendInterface):
             size = meta.size
             self.usage -= size
             self.stats_monitor.update_local_storage_usage(self.usage)
+
+            # Release the space this chunk reserved in `submit_put_task`.
+            # Deducted next to `usage` and before the fallible `os.remove` so
+            # both counters stay consistent with `dict` on every exit path.
+            # Skipped when force=False: the internal eviction path already
+            # deducted the chunk before calling us.
+            if force:
+                self.current_cache_size -= size
 
             # NOTE: The following code will cause deadlock
             # res = asyncio.run_coroutine_threadsafe(
--- a/tests/v1/storage_backend/test_local_disk_backend.py
+++ b/tests/v1/storage_backend/test_local_disk_backend.py
@@ -610,3 +610,184 @@ class TestBatchedGetBlocking:
         results = local_disk_backend.batched_get_blocking([])
         assert results == []
         local_disk_backend.local_cpu_backend.memory_allocator.close()
+
+
+class TestDiskSpaceAccountingOnRemove:
+    """Tests for how the disk budget is accounted across removals.
+
+    ``submit_put_task`` charges every chunk against ``max_local_disk_size``
+    and drops the put when the budget is exhausted and nothing can be
+    evicted.  Removing a chunk therefore has to give its space back, or the
+    backend ends up refusing puts while the disk is actually empty.
+    """
+
+    _SHAPE = torch.Size([2, 2, 16, 8, 128])
+    _DTYPE = torch.bfloat16
+
+    @pytest.fixture
+    def running_loop(self):
+        """Create an asyncio event loop running in a background thread.
+
+        ``submit_put_task`` hands the disk write to the backend's event loop,
+        so the loop has to be running for a put to ever reach the disk.
+        """
+        loop = asyncio.new_event_loop()
+        ready = threading.Event()
+
+        def run_loop() -> None:
+            asyncio.set_event_loop(loop)
+            ready.set()
+            loop.run_forever()
+
+        thread = threading.Thread(target=run_loop, daemon=True)
+        thread.start()
+        if not ready.wait(timeout=5.0):
+            raise RuntimeError("Event loop thread failed to start within 5 seconds")
+
+        yield loop
+
+        loop.call_soon_threadsafe(loop.stop)
+        thread.join(timeout=5)
+        loop.close()
+
+    def test_put_accepted_after_stored_keys_are_removed(
+        self,
+        temp_disk_path: str,
+        running_loop: asyncio.AbstractEventLoop,
+        local_cpu_backend: LocalCPUBackend,
+    ) -> None:
+        """Space freed by ``batched_remove`` is available to later puts.
+
+        Fills the disk budget, drops every key through the public
+        ``batched_remove`` API, then stores a new chunk.  The new chunk has
+        to land on disk -- the disk is empty at that point, so there is room
+        for it.
+        """
+        backend = self._make_backend(
+            temp_disk_path, running_loop, local_cpu_backend, chunk_budget=2.5
+        )
+        try:
+            stored_keys = [create_test_key(i) for i in range(300, 302)]
+            for key in stored_keys:
+                assert self._put_and_wait(backend, key, local_cpu_backend), (
+                    f"initial put for {key} was dropped"
+                )
+                assert backend.contains(key)
+
+            assert backend.batched_remove(stored_keys) == len(stored_keys)
+            for key in stored_keys:
+                assert not backend.contains(key)
+
+            fresh_key = create_test_key(302)
+            assert self._put_and_wait(backend, fresh_key, local_cpu_backend), (
+                "put was dropped even though every stored key had been "
+                "removed and the disk was empty"
+            )
+            assert backend.contains(fresh_key)
+        finally:
+            backend.close()
+            local_cpu_backend.memory_allocator.close()
+
+    def test_eviction_keeps_only_the_newest_chunks(
+        self,
+        temp_disk_path: str,
+        running_loop: asyncio.AbstractEventLoop,
+        local_cpu_backend: LocalCPUBackend,
+    ) -> None:
+        """Storing past the budget evicts the oldest chunks -- and no more.
+
+        Four chunks are stored into a budget that holds two, so the two
+        oldest must be gone and the two newest must remain.  Releasing more
+        space than a chunk occupies would let the cache grow past its
+        budget instead.
+        """
+        backend = self._make_backend(
+            temp_disk_path, running_loop, local_cpu_backend, chunk_budget=2.5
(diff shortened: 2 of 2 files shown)


alooshxl:
Following up on the caveat above about not being able to run the tests locally: CI has now run them. `buildkite/k3-unit-tests` **passed** in 17m22s ([build #6962](https://buildkite.com/lmcache/k3-unit-tests/builds/6962)).

That pipeline runs pytest across the whole `tests/` tree with `--maxfail=1` (ignoring only `tests/disagg`, `test_pos_kernels.py`, `tests/skipped` and `test_eic.py`), so `tests/v1/storage_backend/test_local_disk_backend.py` is in scope and both new tests executed and passed.

`buildkite/unit-tests-amd` is still queued — it is showing `scheduled` on every open PR right now, so that looks like agent-queue backlog rather than anything to do with this change.

PR #4550, opened 2026-08-13
platform/cuda: add shared-flag event IPC backend (event IPC without hostIPC)

royyhuang:
## Motivation

Deploying MP mode on Kubernetes currently requires `hostIPC: true` on both the vLLM pod and the LMCache DaemonSet. We traced the actual dependency: CUDA interprocess **event** handles (`cudaIpcGetEventHandle`/`cudaIpcOpenEventHandle`, used by `DefaultEventIPCBackend`) only resolve when both processes share a `/dev/shm` tmpfs — that is the hostIPC side effect the `lmcache_driven` path depends on for every STORE/RETRIEVE. CUDA IPC **memory** handles have no such dependency: they rendezvous inside the kernel driver and work across fully isolated containers.

## What this adds

`SharedFlagEventIPCBackend` — an `EventIPCBackend` implementation built on memory handles alone:

- Each process lazily allocates a 32 KiB flag buffer per GPU (4096 × 8-byte slots) and exports it once with `cudaIpcGetMemHandle`.
- An event is a `(slot, sequence)` pair: `record_event` = `cuStreamWriteValue64(slot, seq)` on the recording stream; `wait_event` = `cuStreamWaitValue64(slot, seq, GEQ)`; query/synchronize = 8-byte device read.
- Exported handles are self-contained (mem handle + slot offset + sequence, 81 bytes), so they ride the existing opaque-bytes wire field with no registration side channel.
- Slots are per recording stream, keeping slot values monotonic and the GEQ wait race-free.

Also consolidates shared CUDA plumbing into `platform/cuda/utils.py` (lazy cuda-python bindings accessor + small helpers) and switches `ipc_wrapper.py` to it, removing duplicated try/except import blocks (first commit).

## Verification

- 15 tests in `tests/v1/platform/test_shared_flag_event_ipc.py`, parity-parametrized against `DefaultEventIPCBackend` where CUDA events also work (never-recorded semantics, cross-process blocking-until-record-point), plus shared-flag-only coverage (same-process import ordering, stream independence, stale-slot/sequence discipline, malformed/version/offset handle rejection).
- Verified end-to-end across two fully isolated docker containers (separate IPC + PID namespaces, no shared /dev/shm): imported event queried incomplete while the producer stream was busy, a gated read blocked until the record point and observed only the post-record data.
- `pre-commit` (ruff, mypy, isort, codespell) clean; full `tests/v1/platform/` suite passes (134 passed, 6 skipped).

## Scope / status

Standalone: **not yet bound** to `CudaDeviceSpec.event_ipc_backend`. The design doc (`docs/design/v1/platform/cuda/shared_flag_event_ipc.md`) records the empirically load-bearing details (driver 580 / CUDA 13), the deliberate semantic differences vs CUDA events, and the integration constraints that must be resolved before spec binding (connector call sites constructing `torch_dev.Event(interprocess=True)` directly must route through the backend). Selection/config plumbing is a follow-up alongside the raw KV-wrapper work for hostIPC-free deployment.

### DIFF
--- a/docs/design/v1/platform/cuda/shared_flag_event_ipc.md
+++ b/docs/design/v1/platform/cuda/shared_flag_event_ipc.md
@@ -1,0 +1,121 @@
+# Shared-Flag Event IPC Backend
+
+## Motivation
+
+The `lmcache_driven` handle path orders cross-process KV transfers with
+device events (`event_ipc_abstraction.md`). The CUDA implementation behind
+`DefaultEventIPCBackend` uses interprocess event handles — and empirically
+(driver 580, CUDA 13) those only resolve when both processes share a
+`/dev/shm` tmpfs. CUDA IPC *memory* handles have no such dependency: they
+rendezvous in the kernel driver and work across fully isolated containers.
+On Kubernetes, the event handles are what force `hostIPC: true` (or a
+shared hostPath tmpfs) onto the vLLM pod and the LMCache DaemonSet for
+every STORE/RETRIEVE. `SharedFlagEventIPCBackend` removes that requirement
+by implementing event semantics on memory handles alone.
+
+## Mechanism
+
+Each process lazily allocates one flag buffer per GPU (4096 × 8-byte
+slots) and exports it once with `cudaIpcGetMemHandle`. An event is a
+`(slot, sequence)` event object:
+
+| protocol method | implementation |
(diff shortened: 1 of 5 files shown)


royyhuang:
Recreating with the project PR template.

PR #4551, opened 2026-08-13
[MP][Platform]: add timeline-semaphore event IPC backend (no hostIPC)

royyhuang:

Part 1 of removing the `hostIPC: true` requirement for deploying MP mode in containers (vLLM pod + LMCache DaemonSet).

We traced what actually needs hostIPC: CUDA interprocess **event** handles (`cudaIpcGetEventHandle`/`cudaIpcOpenEventHandle`, used by `DefaultEventIPCBackend` on every `lmcache_driven` STORE/RETRIEVE) only resolve when both processes share a `/dev/shm` tmpfs — the hostIPC side effect. CUDA IPC **memory** handles have no such dependency: they rendezvous inside the kernel driver and work across fully isolated containers (verified empirically; only the PID *values* of the two processes must differ).

This PR adds `TimelineSemaphoreEventIPCBackend`, an `EventIPCBackend` implementation built on memory handles alone, using timeline semaphores (the Vulkan/D3D12 concept: a monotonically increasing 64-bit value — signaling writes a target value, waiting blocks until the value reaches it). Each process lazily allocates a small buffer of semaphore slots (4096 × 8 bytes, exported once via `cudaIpcGetMemHandle`); an event is a `(slot, sequence)` pair recorded with `cuStreamWriteValue64` on the recording stream and awaited with `cuStreamWaitValue64` (wait for "greater than and equal"). Exported handles are self-contained 81-byte payloads, so they ride the existing opaque-`bytes` wire field with no registration side channel. Design doc: `docs/design/v1/platform/cuda/timeline_semaphore_event_ipc.md` (mechanism, empirically load-bearing details on driver 580/CUDA 13, semantics vs CUDA events, integration constraints).

The first commit consolidates shared CUDA plumbing into `platform/cuda/utils.py` (lazy cuda-python bindings accessor + helpers) and switches `ipc_wrapper.py` to it, removing duplicated try/except imports.

**Series context** — upcoming sub-PRs for the full hostIPC removal:
1. **(this PR)** event ordering without hostIPC: timeline-semaphore event backend, standalone.
2. Wiring: bind this backend via `CudaDeviceSpec.event_ipc_backend` behind a config knob, route the connectors' direct `torch_dev.Event(interprocess=True)` call sites through `backend.create_event`.
3. Raw `cudaIpcGetMemHandle`-based KV-cache registration (extend `RawCudaIPCWrapper` with `cuMemGetAddressRange` offset handling for caching-allocator tensors + wrapper selection).

- Standalone: nothing selects this backend yet; no behavior change for any existing path. The design doc's "Integration constraints" section lists exactly what PR 2 must resolve before spec binding.
- Verified across two fully isolated docker containers (separate IPC + PID namespaces, no shared /dev/shm), re-run against this exact revision: imported event queried incomplete while the producer stream was busy; a gated read blocked until the record point and observed only post-record data.
- Tests are parity-parametrized against `DefaultEventIPCBackend` where CUDA events also work; timeline-semaphore-only cases cover same-process import and sequence discipline.

**If applicable**:


lmcache/v1/platform/cuda/timeline_semaphore_event_ipc.py
```diff
@@ -0,0 +1,657 @@
...
+            through the write enqueue (see there).
+        slot_by_stream: Raw ``cudaStream_t`` -> slot index.
+        next_seq_by_slot: Slot index -> last assigned sequence.
+        next_free_slot: Next never-used slot. Slots are not reclaimed.
+    """
+
+    device_index: int
+    base_ptr: int
+    handle_bytes: bytes
+    host_ops_stream: object
```
ApostaC: Dumb question: is this stream on the host? Or is it on a specific device?

Update: does this mean the stream that "hosts" the buffer?

lmcache/v1/platform/cuda/timeline_semaphore_event_ipc.py
```diff
@@ -0,0 +1,657 @@
...
+            self.slot_by_stream[raw_stream] = slot
+        seq: int = self.next_seq_by_slot.get(slot, _UNRECORDED_SEQ) + 1
+        self.next_seq_by_slot[slot] = seq
+        return slot * _SLOT_BYTES, seq
+
+
+# Semaphore-buffer base pointers created by this process, by handle bytes:
+# import_event resolves same-process handles locally, since
+# cudaIpcOpenMemHandle cannot open a handle in the exporting process.
+_LOCAL_BUFFERS_BY_HANDLE: dict[bytes, int] = {}
```
ApostaC, line 183: How many local buffers will we have? Is it the same as the number of devices?

lmcache/v1/platform/cuda/timeline_semaphore_event_ipc.py
```diff
@@ -0,0 +1,657 @@
...
+    if event.seq == _UNRECORDED_SEQ:
+        return
+    raw: cudaStream_t = _raw_stream_handle(stream, event.device_index)
+    with torch.cuda.device(event.device_index):
+        result: tuple[object, ...] = _cuda.driver.cuStreamWaitValue64(
+            _cuda.driver.CUstream(raw),
+            _cuda.driver.CUdeviceptr(event.base_ptr + event.slot_offset),
+            event.seq,
+            _cuda.driver.CUstreamWaitValue_flags.CU_STREAM_WAIT_VALUE_GEQ,
+        )
```
ApostaC, line 208: Please educate me on the semantics here: does it mean all the kernels enqueued after this function will be executed after the condition is true on the "value"?

I'm wondering how CUDA driver implements it. Does it require some kind of on-device spin check?

lmcache/v1/platform/cuda/timeline_semaphore_event_ipc.py
```diff
@@ -0,0 +1,657 @@
...
+                    _err, handle = handle_result
+                except Exception:
+                    if host_stream is not None:
+                        _cuda.runtime.cudaStreamDestroy(host_stream)
+                    _cuda.runtime.cudaFree(base)
+                    raise
+            buffer = _TimelineSemaphoreBuffer(
+                device_index=device_index,
+                base_ptr=int(base),
+                handle_bytes=bytes(handle.reserved),
```
ApostaC, line 615: What's the type of `handle`? Asking this because I'm wondering what's there in `handle.reserved` -- and to make sure that it will always be the same if we have the same `base`.

ApostaC (approved):
Mostly LGTM. Just have a few questions.

lmcache/v1/platform/cuda/timeline_semaphore_event_ipc.py
royyhuang: Added in unit tests

sammshen (approved):
LGTM!

ruizhang0101 (approved):
LGTM

ApostaC:
Let's not merge this PR until fixing AMD tests

royyhuang:
AMD tests passed. I will enable auto merge here.

PR #4552, opened 2026-08-14
[CLI] Add rag-qa-quality workload to `lmcache bench engine`

deng451e:
DESCRIPTION
===========
Every document is reused at a position it was never cached at — the RAG serving pattern. The measured request is one a real deployment would receive, so answer-quality changes are attributable to the cache rather than to a perturbation introduced to provoke it.

Usage:

```bash
lmcache bench engine --engine-url http://localhost:8000 \
    --workload rag-qa-quality --tokens-per-gb-kvcache 6000 \
    --rag-dataset musique --rag-num-samples 50 --rag-output run.json
```

Datasets aren't vendored: `musique` and `hotpotqa` download from the HF Hub on first use; any other value is a local path. Four record schemas are recognized (MuSiQue, HotpotQA in both shapes, CacheBlend-style `ctxs`).

Three shared-module changes, all small:

- `BaseWorkload.extra_metric_sections()` — default-empty hook letting a workload add its own report sections. `StatsCollector` covers what every workload shares; this covers what one measures alone.
- `RequestSender(extra_body=...)` — merged with the existing `ignore_eos` injection, for options the OpenAI client has no parameter for (`chat_template_kwargs`).
- `_REQUIRED_WORKLOAD_ARGS` in `command.py` — routes `--rag-dataset` through the existing missing-argument path so `--no-interactive`, the guided TUI, and `--config` replay handle it consistently.

Two reporting decisions worth a look:

- Unparsed answers are written as `f1: null`, not `0.0`; `f1_mean` covers parsed samples only and is reported beside `parse_rate`. Scoring unparsed samples as zero makes a broken arm look like a quality result — not hypothetical during bring-up, where one stack returned no parseable answers at all and `parse_rate` is what made that visible instead of a deflated F1.
- A `run_fingerprint` covers everything determining the prompts. The workload reports one arm; comparing stacks means running it twice and diffing by sample id, and the fingerprint is what makes the two files comparable.

Known limitation: documents are padded to a multiple of 256 tokens (LMCache's default chunk size) so each occupies whole chunks at the same phase in both the prefill and the composite. The alignment unit is hardcoded rather than read from the server, because the baseline stack has no LMCache server to ask and differently-padded runs would no longer share prompts. Deployments on a different chunk size get partial reuse, visible in the reported hit rate.

## If applicable

- [x] this PR contains user facing changes — docs added (`docs/design/cli/commands/bench/engine_bench/bench-engine.md` §4.6)


KuntaiDu (approved):
lgtm!

lmcache/cli/commands/bench/engine_bench/workloads/rag_qa_quality.py
ApostaC, line 1: Make "document length alignment" configurable.

ApostaC (approved):
Otherwise LGTm

lmcache/cli/commands/bench/engine_bench/workloads/rag_qa_quality.py
deng451e: addressed

PR #4553, opened 2026-08-14
[CLI] make max output length tunable for the long-doc-permutator workload

deng451e:
DESCRIPTION
===========

The `long-doc-permutator` workload had no way to set its generation length. It called `RequestSender.send_request()` without `max_tokens`, so every permutation request silently decoded 128 tokens — the sender's signature default, not a deliberate choice — and `validate_max_output_length_supported()` rejected the workload outright if a user tried to set one.

That makes a prefill-only measurement impossible. The workload exists to stress blended KV reuse across reorderings of the same documents, which is a prefill concern; the 128 decode tokens per request add noise to `elapsed_time`, `output_throughput`, and the decode-speed percentiles without exercising anything the workload is about.

This adds `--ldp-max-output-length`, following the existing per-workload flag convention (`--ldqa-max-output-length`, `--mrc-output-length`):

- `max_output_length` field on `LongDocPermutatorConfig`, validated `>= 1`, threaded through `resolve()` into `send_request(max_tokens=...)`.
- `--ldp-max-output-length` registered in the existing ldp argument group.
- `long-doc-permutator` added to `_WORKLOADS_WITH_MAX_OUTPUT_LENGTH` so the validator no longer reports it as lacking the parameter.
- Matching `ConfigItem` in the interactive schema, which also gives it `--export-config` / `--config` round-tripping.
- Flag tables in `docs/design/.../bench-engine.md` and `docs/source/cli/bench.rst`.

The default stays 128 — the value the workload was already getting — so existing runs are unchanged. Use `--ldp-max-output-length 1` for a pure prefill/TTFT measurement, or pair a larger value with `--ignore-eos` for a reproducible decode phase. Warmup is untouched and still uses the sender's 1-token default.

Pre-existing behavior left alone, flagged in case a reviewer prefers otherwise: the guard in `command.py` fires on `--ldqa-max-output-length` but then only asks whether the *workload* has some max-output-length parameter. So `--ldqa-max-output-length N --workload long-doc-permutator` is accepted and ignored — the same looseness that already applied to `multi-round-chat` (which reads `--mrc-output-length`). Tightening the flag/workload pairing would change multi-round-chat too, so it felt out of scope.

Verified: `test_long_doc_permutator.py` + `test_create_workload.py` pass (70). The full `engine_bench` suite shows 45 failures in `test_prefix_suffix_tuner.py` and `test_random_prefill.py` — identical on a clean tree, untouched by this change. Also smoke-tested the real CLI end to end via `--export-config`.

## If applicable

- [x] this PR contains user facing changes — docs added

### DIFF
--- a/docs/design/cli/commands/bench/engine_bench/bench-engine.md
+++ b/docs/design/cli/commands/bench/engine_bench/bench-engine.md
@@ -423,6 +423,7 @@ chunk-level cache lookup and eviction.
 | `num_permutations` | `--ldp-num-permutations` | 10 | Distinct permutations to send (capped at `N!`) |
 | `vocab_size` | (none — hardcoded in factory) | 8000 | Number of distinct single-token words contexts are sampled from |
 | `num_inflight_requests` | `--ldp-num-inflight-requests` | 1 | Max concurrent in-flight requests |
+| `max_output_length` | `--ldp-max-output-length` | 128 | Max tokens generated per request; `1` measures prefill alone |
 
 **Stress axes** (each config field tunes one):
 
--- a/docs/source/cli/bench.rst
+++ b/docs/source/cli/bench.rst
@@ -325,6 +325,11 @@ dispatched with semaphore-controlled concurrency.
    * - ``--ldp-num-inflight-requests``
      - 1
      - Maximum concurrent in-flight requests.
+   * - ``--ldp-max-output-length``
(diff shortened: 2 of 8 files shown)


ApostaC (approved):
LGTM!

ruizhang0101 (approved):
LGTM

PR #4555, opened 2026-08-14
[Bugfix][CLI][L2 Bench] Use L1-relative memory object addresses

manojgop:

The L2 benchmark creates memory objects as views into a shared, registered L1 buffer. It previously stored each object's absolute virtual address in `MemoryObjMetadata.address`.

NIXL adapters interpret this field as a byte offset within the registered L1 buffer. Using an absolute address can therefore produce invalid page indices and cause NIXL store or load operations to fail.

This PR:

- Stores each object's buffer-relative byte offset instead of its absolute virtual address.
- Adds regression assertions verifying the expected metadata offsets.

This change is limited to L2 benchmark data construction. The NIXL adapters are unchanged. Production LMCache allocators already use buffer-relative addresses, so this makes the benchmark follow the existing address contract.

Validation:

```bash
pytest -xvs tests/cli/commands/bench/l2_adapter_bench/test_data.py
```
Result: `4 passed`.

All applicable pre-commit checks passed for the changed files.

**If applicable**:

- [x] this PR contains unit tests (`tests/cli/commands/bench/l2_adapter_bench/test_data.py`)

### DIFF
--- a/lmcache/cli/commands/bench/l2_adapter_bench/data.py
+++ b/lmcache/cli/commands/bench/l2_adapter_bench/data.py
@@ -128,7 +128,7 @@ def make_memory_objects(
         metadata = MemoryObjMetadata(
             shape=torch.Size([data_size]),
             dtype=torch.uint8,
-            address=raw_tensor.data_ptr(),
+            address=start,
             phy_size=data_size * raw_tensor.element_size(),
             fmt=MemoryFormat.KV_2LTD,
             ref_count=1,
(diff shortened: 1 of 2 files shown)


zhengfeihe (approved):
Thanks for catching this! 

Yeah, I think it should be a relative offset, not an absolute address here

Quick Reference: 

```
    @property
    def shm_offset(self) -> int:
        """Return the byte offset of this object inside the SHM pool."""
        return self.meta.address
````

manojgop:
> 

Is there anything else that needs to be done for this PR before it can be merged ?

Issue #3936, opened 2026-06-29
[feat]: Proposal about Unified Device Ops via `DeviceOps` Abstraction


zxue2:
# Design: Unified Device Ops via `DeviceOps` Abstraction

---

## 1. Goal

Unify the per-device **ops** (callable ops plus shared types, currently
exposed as `lmcache.c_ops`) behind a single `DeviceOps` abstraction in
`lmcache/v1/platform/`, alongside the existing device abstractions
(`DeviceIPCWrapper`, `PinMemoryBackend`, `BaseCacheContext`).

**The torch reference implementation moves *into* `DeviceOps`.** The base class
*is* the CPU/torch backend — `python_ops_fallback.py` is
**deprecated entirely**, its logic migrated into the platform package. Every device
(CUDA, HIP, XPU, MUSA, HPU) is a `DeviceOps` subclass that overrides only what it
accelerates and inherits the torch baseline for everything else.

---

## 2. What exists today

| Concern | Mechanism | Location |
|---------|-----------|----------|
| Compiled CUDA ops | `PYBIND11_MODULE(c_ops)` — 31 ops + 3 types (+`GPUKVFormat` alias) | `csrc/pybind.cpp` |
| Compiled SYCL ops | `PYBIND11_MODULE(xpu_ops)` — 12 ops + 2 enums (+`GPUKVFormat`); **19 ops fall back to torch** | `csrc/sycl/pybind_sycl.cpp` |
| Torch/CPU reference | `python_ops_fallback.py` — 31 ops + 3 types | `lmcache/python_ops_fallback.py` |
| MUSA ops | Python adapter: import `py_ops`, override 1 fn, optional native | `lmcache/v1/platform/musa/ops.py` |
| HPU ops | None — uses `python_ops_fallback` entirely | (via `_get_backend`) |
| Runtime selection | `_get_backend()`: `py_ops` base + first-available backend, `__dict__.update` merge | `lmcache/__init__.py` |
| Build selection | `BuildProfile` subclasses auto-discovered | `setup_extensions/build_profiles/` |
| Device services registry | `DeviceIPCWrapper`/`PinMemoryBackend`/`BaseCacheContext` auto-discovered by `device_type` | `lmcache/v1/platform/` |

### 2.1 Current Problems

1. **Two parallel selection mechanisms.** `_get_backend()` uses a hand-maintained
   `backend_candidates` list + `__dict__.update` merge — separate from the clean
[...]

PR #4556, opened 2026-08-14
feat: discover out-of-tree device specs

maobaolong:
## Summary
- discover `DeviceSpec` backends from installed `lmcache.v1.device_specs` entry points in addition to in-tree `lmcache.v1.platform` packages
- reuse the shared platform registry for device detection and backend resolution so external wheels behave like built-in platform backends
- document out-of-tree backend packaging, add registry tests, and record the local verification/toolchain pitfalls found during validation

## Testing
- `/Users/mbl/projects/LMCache/.venv/bin/python -m pytest tests/v1/platform/test_device_detect.py tests/v1/platform/test_device_ops.py`
- `/Users/mbl/projects/LMCache/.venv/bin/pre-commit run --all-files` (fails only because local `cargo` is not installed; Python hooks passed)
- `PATH="/Users/mbl/projects/LMCache/.venv/bin:$PATH" make clean html` from `docs/` (build completed, but the repo still has pre-existing Sphinx warnings/errors outside this change)

## Notes
- `tests/v1/platform/test_event_ipc.py tests/v1/platform/test_cache_context_dispatch.py` mostly pass, but `test_cuda_device_spec_exposes_cached_default_event_backend` still fails on this CPU-only host because it expects `torch_device_type == "cuda"`; this was not touched by the change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,13 @@ uv pip install torch               # pre-requisite for CUDA extensions
 uv pip install -e . --no-build-isolation
 ```
 
+On this checkout, do not assume ``python`` is available on ``PATH`` or
+that ``python3`` points at the project interpreter: the local system
+``python3`` may still be Python 3.9 while LMCache development here uses
+``.venv/bin/python`` (Python 3.12). Run tests, docs, and pre-commit via
+``.venv/bin/python`` / ``.venv/bin/<tool>`` or ``uv run`` to avoid
+accidentally validating against the wrong interpreter.
+
 ## Build & Install
 
 ```bash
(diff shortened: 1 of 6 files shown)


andyluo7:
Making the collision explicit instead of letting the dict comprehension silently overwrite is a clear improvement — the old `{spec.device_type: spec for ...}` gave last-wins with no signal at all.

One thing worth thinking about before this locks in: `_register_device_spec` resolves collisions by registration order, and since the in-tree loop runs before entry points, the first spec to claim a `device_type` always wins and everything after it is dropped with a warning.

That's the right default for an accidental clash. It also forecloses the deliberate case, which is live in #3936 right now: ROCm can't have its own `DeviceSpec` because torch built for ROCm reports `cuda`, so a ROCm spec — in-tree under `platform/rocm/` or shipped in an external wheel — collides with `CudaDeviceSpec` and would now be skipped rather than selected. `device_type` can't just return `"rocm"` either; it's interpolated into `torch.device(f"{torch_device_type}:{idx}")` in ~20 places, and `torch.device("rocm:0")` raises on ROCm torch.

So the two threads collide: @zxue2 is asking for a `platform/rocm` package on #3936, and this PR would make such a package inert.

Might be worth making the tiebreak explicit rather than positional — prefer the most-derived spec that reports `is_available()`, or let a spec declare a priority. Both keep the accidental-clash warning while leaving room for a vendor to specialise an existing device type.

Not blocking the entry-point mechanism itself, which looks fine.

maobaolong:
@andyluo7 Thanks for the valuable suggestion from your side! I'll think about it and do some update on #4561 , as this is an early draft version PR, and it duplicated to #4561 , so close it now. Please feel free to continue discussion on #4561

PR #4561, opened 2026-08-14
feat: support external device plugins

maobaolong:

- Discovers external `DeviceSpec` classes from the `lmcache.device_plugins` Python entry-point group, allowing hardware vendors to ship independently versioned device-support wheels.
- Preserves the existing in-tree model under `lmcache/v1/platform/<device>/`; both integration models use the same interfaces and process-wide registry instances.
- Separates torch-facing `device_type` from the unique LMCache `backend_name`, so multiple concrete backends can intentionally share one torch device category.
- Resolves `LMCACHE_DEVICE_BACKEND` explicitly when set. Otherwise, LMCache automatically selects the only available backend for a `device_type` and raises an actionable error if multiple backends report available.
- Rejects duplicate `backend_name` registrations deterministically while allowing distinct backends to share a `device_type`.
- Adds a dedicated `RocmDeviceSpec`. CUDA and ROCm both use torch's `device_type="cuda"`, but mutually exclusive CUDA/ROCm runtime checks select the correct backend without requiring an environment variable.
- Validates plugin contracts, isolates entry-point load failures, and covers discovery, invalid plugins, duplicate backend names, shared device types, explicit selection, and runtime detection with unit tests.
- Documents the in-tree and external-wheel workflows, selection semantics, IPC capabilities, and ownership tradeoffs.

**Reviewer follow-up**:

- The CUDA/ROCm automatic-selection behavior was validated on MI355X with ROCm 7.2 and B300 with CUDA 13.0: each environment resolves exactly one backend without `LMCACHE_DEVICE_BACKEND`.
- `CudaDeviceSpec` checks the CUDA runtime through `torch.version.cuda`; HIP-specific detection remains isolated in `RocmDeviceSpec` through `torch.version.hip`.

**Validation**:

- `44 passed, 1 skipped` in the focused device plugin, DeviceOps, and RBLN platform suite.
- All Python, policy, formatting, codespell, clang-format, mypy, and Rust formatting pre-commit hooks passed. Rust clippy is Linux-only because `raw_block` depends on `io-uring`, so GPU/Linux CI is authoritative for that hook.
- Sphinx generated all HTML successfully. The full documentation baseline emitted 29 existing warnings; the changed device-backend guide emitted no warning.

**External XPU Validation**:

- Follow-up validation PR: `#4624` (`test: verify external native XPU device wheel in CI`), opened on top of this branch and targeting `dev`.
- Validation goal: remove in-tree XPU platform ownership from the validation branch, install LMCache plus an out-of-tree XPU device wheel, and prove that `lmcache.device_plugins` can restore working XPU support through the external plugin path introduced here.
- External plugin repo: `opendataio/lmcache_xpu_device_with_native`
- Published wheel used by CI: `v0.1.4`
- XPU CI result: `buildkite/unit-tests-xpu` passed in Build `#855`

Key evidence from Build `#855`:

- The job downloaded the external plugin source from `https://github.com/opendataio/lmcache_xpu_device_with_native/archive/refs/tags/v0.1.4.tar.gz`
- It built the native extension from the external repo as `lmcache_xpu_device_with_native.xpu_ops`
- The compile command used `-I/workspace/build/buildkite/csrc`, so shared LMCache headers were consumed from LMCache `csrc` through an include path rather than copied into the plugin repo
- The wheel build completed successfully:
  - `Successfully built lmcache_xpu_device_with_native-0.1.4-cp312-cp312-linux_x86_64.whl`
- The runtime validation completed successfully:
  - `external native xpu device plugin resolved successfully`
- The XPU shard then ran the XPU-related pytest suite and passed native-kernel coverage such as:
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_cdf_sycl[256]`
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_encode_sycl[64]`
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_decode_sycl[256]`
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_rope_sycl[4096]`

This validates that the external plugin-discovery mechanism in this PR is sufficient for a full XPU externalization path:

- XPU `DeviceSpec` / `DeviceOps` registration is resolved through `lmcache.device_plugins`
- the native SYCL extension is built and installed from the out-of-tree plugin package
- and the XPU CI path continues to work with XPU supplied externally rather than by LMCache main repo ownership

Related links:

- Validation PR: https://github.com/LMCache/LMCache/pull/4624
- External plugin repo: https://github.com/opendataio/lmcache_xpu_device_with_native
- External wheel release used in CI: https://github.com/opendataio/lmcache_xpu_device_with_native/releases/tag/v0.1.4
- XPU Buildkite build: https://buildkite.com/lmcache/unit-tests-xpu/builds/855

**If applicable**:

- [x] this PR contains user-facing changes - docs added

### DIFF
--- a/lmcache/v1/platform/__init__.py
+++ b/lmcache/v1/platform/__init__.py
@@ -155,6 +155,12 @@ def _resolve_device_spec(device_type: str) -> DeviceSpec:
         ]
         if len(compatible_candidates) == 1:
             return compatible_candidates[0]
+
+        default_candidates = [
+            spec for spec in candidates if spec.backend_name == device_type
+        ]
+        if len(default_candidates) == 1:
+            return default_candidates[0]
     if device_type in ("", "cpu"):
         return _FALLBACK_CPU_SPEC
     raise RuntimeError(
--- a/lmcache/v1/platform/__init__.py
+++ b/lmcache/v1/platform/__init__.py
@@ -15,7 +15,8 @@ sub-packages so each can evolve independently:
 
 Third-party accelerators can ship a :class:`DeviceSpec` subclass in a
 separate wheel and register it through the ``lmcache.device_plugins`` Python
-entry-point group. No LMCache source changes are required.
+entry-point group. This complements the built-in integration model for
+backends maintained directly in the LMCache repository.
 """
 
 # Future
--- a/lmcache/v1/platform/__init__.py
+++ b/lmcache/v1/platform/__init__.py
@@ -43,6 +43,7 @@ from typing import TYPE_CHECKING, Any
 import os
 
 # First Party
+from lmcache.logging import init_logger
 from lmcache.v1.platform._device_detect import (
     DEVICE_BACKEND_ENV_VAR,
     _build_backend_registry,
@@ -69,6 +70,8 @@ from lmcache.v1.platform.event_notifier import consume_fd as consume_fd
 from lmcache.v1.platform.event_notifier import (
     create_event_notifier as create_event_notifier,
 )
+
+logger = init_logger(__name__)
 
 # ---------------------------------------------------------------------------
 # Resolve device ops
@@ -131,6 +134,9 @@ def _resolve_device_spec(device_type: str) -> DeviceSpec:
         return candidates[0]
 
     if len(candidates) > 1:
+        candidate_backend_names = ", ".join(
+            sorted(spec.backend_name for spec in candidates)
+        )
         explicit_backend = os.environ.get(DEVICE_BACKEND_ENV_VAR, "").strip().lower()
         if explicit_backend:
             dev_spec = _DEVICE_REGISTRY.get(explicit_backend)
@@ -139,6 +145,13 @@ def _resolve_device_spec(device_type: str) -> DeviceSpec:
 
         available_candidates = [spec for spec in candidates if spec.is_available()]
         if len(available_candidates) == 1:
+            logger.info(
+                "Auto-selected backend [%s] for accelerator %r from candidate "
+                "backends [%s].",
+                available_candidates[0].backend_name,
+                device_type,
+                candidate_backend_names,
+            )
             return available_candidates[0]
         if len(available_candidates) > 1:
             backend_names = ", ".join(
--- a/lmcache/v1/platform/__init__.py
+++ b/lmcache/v1/platform/__init__.py
@@ -150,12 +150,6 @@ def _resolve_device_spec(device_type: str) -> DeviceSpec:
                 f"{DEVICE_BACKEND_ENV_VAR}=<backend_name> to choose one explicitly."
             )
 
-        compatible_candidates = [
-            spec for spec in candidates if spec.is_runtime_compatible()
-        ]
-        if len(compatible_candidates) == 1:
-            return compatible_candidates[0]
-
         default_candidates = [
             spec for spec in candidates if spec.backend_name == device_type
         ]
--- a/lmcache/v1/platform/__init__.py
+++ b/lmcache/v1/platform/__init__.py
@@ -40,9 +40,12 @@ __all__ = [
 
 # Standard
 from typing import TYPE_CHECKING, Any
+import os
 
 # First Party
 from lmcache.v1.platform._device_detect import (
+    DEVICE_BACKEND_ENV_VAR,
+    _build_backend_registry,
     _build_device_registry,
 )
 from lmcache.v1.platform._device_detect import (
@@ -72,10 +75,10 @@ from lmcache.v1.platform.event_notifier import (
 # ---------------------------------------------------------------------------
 
 
-# Keep the historical private name as an alias for existing internal tests and
-# downstream diagnostics. Both detection and resolution now share the same
-# DeviceSpec instances, including external plugins.
-_DEVICE_REGISTRY: dict[str, DeviceSpec] = _build_device_registry()
+# Keep the historical private name for the backend-name registry so existing
+# tests and downstream diagnostics can still inspect the resolved spec table.
+_DEVICE_REGISTRY: dict[str, DeviceSpec] = _build_backend_registry()
+_DEVICE_TYPE_REGISTRY: dict[str, tuple[DeviceSpec, ...]] = _build_device_registry()
 
 
 # ---------------------------------------------------------------------------
@@ -99,7 +102,7 @@ def resolve_kv_wrapper_factory(device_type: str) -> Any:
     Raises:
         ValueError: If no spec / wrapper is registered for *device_type*.
     """
-    spec = _DEVICE_REGISTRY.get(device_type)
+    spec = _resolve_device_spec(device_type)
     wrapper_cls = spec.ipc_wrapper_cls if spec is not None else None
     if wrapper_cls is None:
         raise ValueError(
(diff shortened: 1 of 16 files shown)


maobaolong:
> https://github.com/LMCache/LMCache/pull/4556#issuecomment-5297818021

@andyluo7  Thanks for the suggestion. I updated the PR to separate the torch-facing device type from the LMCache backend selector.

The new behavior is:
- `device_type` is still the torch-facing category and may be shared by multiple backends
- `backend_name` is a unique LMCache-side identifier for one concrete backend
- `LMCACHE_DEVICE_BACKEND` can be used to select a specific backend explicitly
- if multiple backends share the same `device_type`, `DEVICE_TYPE` alone is no longer enough to disambiguate them

This keeps `DEVICE_TYPE` aligned with torch semantics, while still allowing vendor-specific specializations for the same device type.

andyluo7:
That split is exactly what was needed — thanks for turning it around so quickly.

One gap left for the ROCm case: auto-selection keys on `is_available()`, but `CudaDeviceSpec.is_available()` is just `torch.cuda.is_available()`, which returns `True` on a ROCm build (checked on torch 2.11.0+rocm7.2). So once a `RocmDeviceSpec` exists both candidates report available and `_resolve_device_spec` raises `Multiple DeviceSpec backends are available for accelerator 'cuda'` — meaning every ROCm user would have to set `LMCACHE_DEVICE_BACKEND` just to start.

Having `CudaDeviceSpec.is_available()` return `torch.cuda.is_available() and torch.version.hip is None` would make the automatic path work. This PR doesn't touch `platform/cuda/__init__.py`, so it could be a follow-up — but worth settling before a ROCm backend lands on top of it.

maobaolong:
@andyluo7  Thanks for calling this out. I fixed it by making `CudaDeviceSpec.is_available()` return `False` on ROCm builds (`torch.version.hip is not None`) and added a regression test for the shared-`cuda` backend case. This should keep auto-selection working without requiring `LMCACHE_DEVICE_BACKEND` for ROCm users.

Also, is AMD planning to add a dedicated ROCm backend module on top of this plugin model?

andyluo7:
Validated the fix on both — auto-selection resolves correctly with no env var on either side.

**MI355X (gfx950), `vllm/vllm-openai-rocm:v0.26.0`, torch 2.11.0+gitd0c8b1f, hip 7.2.53211, cuda None:**
```
CudaDeviceSpec: device_type=cuda backend=cuda available=False
RocmDeviceSpec: device_type=cuda backend=rocm available=True
RESOLVED -> RocmDeviceSpec  backend_name=rocm  device_type=cuda
```

**B300, `vllm/vllm-openai:v0.26.0`, torch 2.11.0+cu130, hip None, cuda 13.0:**
```
CudaDeviceSpec available=True
RocmDeviceSpec available=False
RESOLVED -> CudaDeviceSpec  backend_name=cuda
```

The `is_runtime_compatible` split works because `torch.version.cuda` and `torch.version.hip` are mutually exclusive, so exactly one candidate is available per build and nobody needs `LMCACHE_DEVICE_BACKEND`. `RocmDeviceSpec` inheriting `device_type = "cuda"` is also the right call — that string is interpolated into `torch.device(f"{torch_device_type}:{idx}")` in ~20 places, and per the HIP docs @zxue2 linked, `cuda` is the only valid value.

On your question: yes. With this landed there's a place for the ROCm divergences to live, which there wasn't before. The obvious first ones are the three duplicated `_is_rocm()` predicates — `compute/attention/utils.py:21`, the `torch.version.hip` check in `_gds_async.py`, and the one #4422 proposes to add in `platform/base/event_ipc.py` — all of which become `RocmDeviceSpec` overrides once this exists. That's a natural follow-up rather than something to fold in here.

maobaolong:
@andyluo7 Thanks again for validating this on both ROCm and CUDA.  
For clarification, is_runtime_compatible() was folded into each backend’s is_available() in 90f9c86e. CUDA checks torch.version.cuda, while ROCm checks torch.version.hip, so the hardware auto-selection behavior you validated remains unchanged. The only intentional difference is that no-hardware lookups keep CUDA as the default backend unless LMCACHE_DEVICE_BACKEND is set explicitly.  
No runtime selection code has changed since your validation.

maobaolong:
Supplemental XPU validation note:

A follow-up validation PR was created on top of this branch: #4620.

That PR tested the XPU path with:
- in-tree `lmcache/v1/platform/xpu` removed,
- an external `lmcache_xpu_device` wheel installed at runtime, and
- an explicit assertion that `lmcache.device_plugins` resolves the `xpu` backend before XPU tests run. The external wheel used there was published from `opendataio/lmcache_xpu_device` (`v0.1.0`). ([github.com](https://github.com/LMCache/LMCache/pull/4620))

This gives useful evidence that the **XPU platform Python layer** can be externalized on top of the plugin-discovery mechanism introduced in this PR.

However, this is **not yet a full proof of complete XPU externalization**.

Current limitation:
- the external repo does **not** yet include `csrc/sycl/`
- `XpuDeviceOps` in the external wheel still imports `lmcache.xpu_ops`
- and `lmcache.xpu_ops` is still built from LMCache main repo sources

So the current result supports:
- externalization of `XpuDeviceSpec` / `XpuDeviceOps` registration and resolution
- but **not yet full externalization of the XPU native SYCL implementation**

TODO:
- move `csrc/sycl/` into the external XPU device repo
- build the XPU native extension there
- stop relying on LMCache main repo to provide `lmcache.xpu_ops`
- rerun the same XPU validation after that

maobaolong:
Follow-up to my earlier TODO note:

https://github.com/LMCache/LMCache/pull/4561#issuecomment-5325455551

This TODO is now resolved by #4624.

Validation artifacts:

- Validation PR: https://github.com/LMCache/LMCache/pull/4624
- External plugin repo: https://github.com/opendataio/lmcache_xpu_device_with_native
- External wheel release used by CI: https://github.com/opendataio/lmcache_xpu_device_with_native/releases/tag/v0.1.4
- XPU CI build: https://buildkite.com/lmcache/unit-tests-xpu/builds/855

What changed relative to the earlier TODO:

- [x] `csrc/sycl/` lives in the external XPU device repo
- [x] the native XPU extension is built in that external repo and packaged into the wheel release above
- [x] shared LMCache public headers are consumed from LMCache `csrc` through `LMCACHE_CSRC_INCLUDE_DIR`; they are not copied into the plugin repo
- [x] the plugin no longer relies on LMCache main repo to build `lmcache.xpu_ops`; the native module is now provided by the external package as `lmcache_xpu_device_with_native.xpu_ops`
- [x] the XPU validation was rerun on top of this branch and `buildkite/unit-tests-xpu` passed in Build `#855`

Key evidence from Build `#855`:

- downloaded source: `https://github.com/opendataio/lmcache_xpu_device_with_native/archive/refs/tags/v0.1.4.tar.gz`
- built native module: `building 'lmcache_xpu_device_with_native.xpu_ops' extension`
- compile command included: `-I/workspace/build/buildkite/csrc`
- wheel build succeeded: `Successfully built lmcache_xpu_device_with_native-0.1.4-cp312-cp312-linux_x86_64.whl`
- runtime plugin resolution succeeded: `external native xpu device plugin resolved successfully`
- XPU native-kernel tests then passed, including:
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_cdf_sycl[256]`
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_encode_sycl[64]`
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_decode_sycl[256]`
  - `tests/benchmarks/test_xpu_kernels_microbench.py::test_bench_rope_sycl[4096]`

Conclusion:

This PR's external-device-plugin mechanism has now been validated end to end for XPU, including the native SYCL path, not just the Python registration layer.

andyluo7:
Re-ran it on `ff20e19f` since the `is_runtime_compatible()` fold landed after my last check. Same result on both:

MI355X (gfx950), torch 2.11.0+gitd0c8b1f, hip 7.2.53211, cuda None:
```
Cuda available=False | Rocm available=True
RESOLVED -> RocmDeviceSpec  backend_name=rocm  device_type=cuda
```

B300, torch 2.11.0+cu130, hip None, cuda 13.0:
```
Cuda available=True | Rocm available=False
RESOLVED -> CudaDeviceSpec  backend_name=cuda
```

So the inline `torch.version.cuda` / `torch.version.hip` checks behave the same as the separate `is_runtime_compatible()` did — exactly one candidate is available per build and neither needs `LMCACHE_DEVICE_BACKEND`.

lmcache/v1/platform/__init__.py
```diff
@@ -213,16 +126,42 @@ def _resolve_device_spec(device_type: str) -> DeviceSpec:
     Raises:
         RuntimeError: If an accelerator device has no registered spec.
     """
-    dev_spec = _DEVICE_REGISTRY.get(device_type)
-    if dev_spec is not None:
-        return dev_spec
+    candidates = _DEVICE_TYPE_REGISTRY.get(device_type, ())
```
KuntaiDu, line 132: Nit: maybe we can log all candidates and specify which one did LMCache choose?

KuntaiDu (approved):
LGTM! Left one small nit.

lmcache/v1/platform/__init__.py
maobaolong: Quite a good suggestion!

ApostaC (approved):
LGTM

PR #4563, opened 2026-08-14
[MP] Make L2 LRU eviction byte-aware

positive666:
## Summary

L2 adapters already report the byte size of each stored object, but the eviction
policy discarded it and interpreted `eviction_ratio` as a fraction of key count.
That works for uniform objects, but mixed-size KV objects can leave the cache
above its watermark or evict substantially more data than requested.

This change forwards L2 object sizes to the LRU policies. Sized, heterogeneous
workloads evict the smallest LRU prefix that reaches the byte target. Unsized
callers and uniform-size workloads retain the existing count-based behavior and
rounding. A bucket also stays in count mode while any object size is unknown.

The same accounting is maintained independently for each `cache_salt`. Size
refreshes and removals update the totals in constant time. Durable isolated-LRU
state includes size weights only for complete buckets; older order-only
checkpoints restore in count mode.

## Failed-delete progress

A persistent failure on an oversized LRU head could otherwise select the same
key on every controller pass and block smaller objects behind it. The controller
now moves attempted keys to the MRU end before calling the adapter. Successful
deletes still remove the keys through the normal adapter notification; a raised
or skipped delete cannot monopolize the next pass. Adapter exceptions are logged
per state so the background eviction thread continues running. The move is not
recorded as a cache access and does not change access metrics.

With a permanently failing 320 MiB head followed by 100 deletable 1 MiB objects
and `eviction_ratio=0.2`, five passes produced:

| Policy | Bytes freed |
| --- | ---: |
| Previous PR revision | 0 MiB |
| Current `dev` count policy | 64 MiB |
| Updated byte-aware policy | 100 MiB |

## End-to-end reproduction

Using a 256 MiB mock L2 cache containing four 33 MiB objects and one 66 MiB
object, with a 0.6 high watermark and a 0.25 eviction ratio:

- `dev` selected one 33 MiB object and needed a second controller pass.
- This change selected two 33 MiB objects and completed in one pass.

Both runs ended with the same three objects and 132 MiB stored. The workload
used Qwen3.5-4B BF16 with vLLM `0.23.1rc1.dev753+g4875b4456.cu129`.

## Scope

The eligible-key byte-target case raised in review was tested separately. No
current production caller combines real byte weights with an eligibility
filter: local L2 supplies sizes without a filter, while the filtered paths use
unit weights. An exact eligible-byte pre-pass fixed the synthetic case but made
the 1M-key reviewer workload 12.8% slower (731.4 ms to 825.0 ms median), so it is
not included here.

Related open PRs were checked. #3859 changes FS capacity reporting and #3909
restores S3 LRU state; neither changes victim selection from key count to bytes.

## Validation

```text
python -m pytest \
  tests/v1/distributed/test_lru_eviction_policy.py \
  tests/v1/distributed/test_isolated_lru_eviction_policy.py \
  tests/v1/distributed/test_cache_salt_l2_eviction.py \
  tests/v1/mp_coordinator/persistence/test_durable_components.py \
  tests/v1/mp_coordinator/persistence/test_restore_without_replay.py -q
79 passed

pre-commit: SPDX, isort, ruff, ruff-format, codespell, mypy
passed

git diff --check
passed
```

- [ ] This PR contains user-facing changes requiring documentation.
- [x] This PR contains unit tests.

### DIFF
--- a/docs/design/v1/mp_coordinator/persistence/README.md
+++ b/docs/design/v1/mp_coordinator/persistence/README.md
@@ -15,7 +15,7 @@ contract below and restores itself from its own section:
 | `KeyDirectory` | `key_directory` | A restarted MP server re-announces its L1, but nothing re-announces bytes already resting in L2 |
 | `CacheUsageManager` | `cache_usage` | Byte accounting for those same placements; quota enforcement is blind without it |
 | `EventGate` | `stream_cursors` | Fencing compares an arriving batch against a prior incarnation; with no cursor there is nothing to compare, and a restarted server's stale L1 slice is advertised forever |
-| `IsolatedLRUEvictionPolicy` | `lru_order` | Recency is position, not a timestamp |
+| `IsolatedLRUEvictionPolicy` | `lru_order` | Recency is positional; complete byte weights drive eviction |
 | `FleetEvictionController` | `pins` | Operator intent |
 | `QuotaManager` | `quotas` | Operator intent |
 
--- a/lmcache/v1/distributed/eviction.py
+++ b/lmcache/v1/distributed/eviction.py
@@ -100,6 +100,22 @@ class EvictionPolicy:
         """
         pass
 
+    def on_keys_created_with_sizes(
+        self,
+        keys: list[ObjectKey],
+        sizes: list[int],
+    ) -> None:
+        """Notify the policy that sized keys have been created.
+
+        Policies that do not use object sizes inherit the compatibility
+        fallback, which delegates to :meth:`on_keys_created`.
+
+        Args:
+            keys: Keys that have been created.
+            sizes: Size in bytes for each key.
+        """
+        self.on_keys_created(keys)
+
     @abstractmethod
     def on_keys_touched(self, keys: list[ObjectKey]):
         """
@@ -120,6 +136,17 @@ class EvictionPolicy:
         """
         pass
 
+    def on_eviction_attempted(self, keys: list[ObjectKey]) -> None:
+        """Notify the policy that keys were submitted for eviction.
+
+        Policies may use this signal to keep an unremovable LRU head from
+        blocking other candidates. The default implementation preserves
+        compatibility for policies that do not track retry order.
+
+        Args:
+            keys: Keys submitted to an eviction destination.
+        """
+
     @abstractmethod
     def get_eviction_actions(
         self,
@@ -132,10 +159,10 @@ class EvictionPolicy:
 
         Args:
             expected_ratio (float): A hint indicating approximately what
-                fraction of tracked keys should be evicted. Value should be
-                in range [0.0, 1.0]. For example, 0.1 means roughly 10% of
-                keys should be evicted. This is a hint and the policy may
-                return more or fewer keys.
+                fraction of tracked eviction weight should be evicted. Policies
+                may use bytes when sizes are available and unit weight otherwise.
+                Value should be in range [0.0, 1.0]. This is a hint and the
+                policy may return more or fewer keys.
             key_eligible_filter: An optional callable that takes an ObjectKey
                 and returns True if the key is eligible for eviction. When
                 provided, keys for which the filter returns False will be
@@ -222,8 +249,12 @@ class L2EvictionPolicy(L2AdapterListener):
     def policy(self) -> EvictionPolicy:
         return self._policy
 
-    def on_l2_keys_stored(self, keys: list[ObjectKey], sizes: list[int]):
-        self._policy.on_keys_created(keys)
+    def on_l2_keys_stored(
+        self,
+        keys: list[ObjectKey],
+        sizes: list[int],
+    ) -> None:
+        self._policy.on_keys_created_with_sizes(keys, sizes)
 
     def on_l2_keys_accessed(self, keys: list[ObjectKey]):
         self._policy.on_keys_touched(keys)
--- a/lmcache/v1/distributed/eviction_policy/isolated_lru.py
+++ b/lmcache/v1/distributed/eviction_policy/isolated_lru.py
@@ -54,8 +54,11 @@ class IsolatedLRUEvictionPolicy(EvictionPolicy):
         default_destination: EvictionDestination = EvictionDestination.DISCARD,
     ):
         self._lock = threading.Lock()
-        # cache_salt -> ordered {ObjectKey: None} (oldest first).
-        self._per_salt_order: dict[str, OrderedDict[ObjectKey, None]] = {}
+        # cache_salt -> ordered {ObjectKey: size} (oldest first).
+        self._per_salt_order: dict[str, OrderedDict[ObjectKey, int | None]] = {}
+        self._per_salt_total_size: dict[str, int] = {}
+        self._per_salt_total_size_squared: dict[str, int] = {}
+        self._per_salt_unknown_size_count: dict[str, int] = {}
         # Registered destinations (first one wins if any are registered,
         # matching LRUEvictionPolicy semantics).
         self._destinations: list[EvictionDestination] = []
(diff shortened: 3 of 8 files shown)


pengw00:
From a prod safety perspective regarding `Victim selection`, my primary concern is the potential tail-latency overhead (O(N) linear scan) when the cache pool is cluttered with millions of tiny, fragmented metadata pieces. Under heavy runtime inference pressure, an unconstrained linear walk might introduce noticeable CPU spikes and jitter the critical path (like TTFT).

To mitigate this worst-case performance degradation while keeping your single-walk logic intact, what do you think about introducing a defensive fallback threshold? For example, adding a `max_keys_to_scan` parameter (e.g., capping the lookahead at 100 keys). If the target isn't met after the threshold, it could gracefully break the loop or trigger a batch eviction pass.

pengw00:
@positive666  You mentioned that: "Global LRU and per-cache_salt isolated LRU use the same rule." 

While applying the same ratio-to-byte calculation to an isolated tenant pool without dynamic tenant-byte caps introduces a Noisy Neighbor risk in shared cloud deployments. 

In a multi-tenant cluster, if a heavy tenant (under a specific `cache_salt`) generates an massive surge of long-context heterogeneous KV objects, the calculated byte-target for that specific cache_salt will expand exponentially. When the walk triggers, it risks aggressively exhausting the cache survival lifetime of adjacent lighter tenants sharing the physical hardware pool, leading to catastrophic cache thrashing. 

Why did we decide to replicate the global eviction rule natively inside the isolated tenant LRU instead of introducing hierarchical token-bucket limits or independent tenant-byte quota caps? 

How does this design protect lighter tenants from suffering significant TTFT tail-latency spikes when a neighboring tenant experiences a flash crowd of long-context inferences? Would love to know if there would be a quota-guard layer to enforce independent isolation ceilings here!

positive666:
Thanks for raising both points. I ran an exact base/head comparison with 100K and 1M entries, plus a controller-level two-tenant test.

For isolation, the controller still selects only the over-quota `cache_salt`. In the test, the heavy tenant dropped from 72 KiB to 32 KiB in one pass, while the light tenant stayed at 128 KiB with both keys retained. The byte target is linear and local to that bucket.

For scan cost, uniform 1M-entry global LRU was 43.96 ms on the baseline and 43.65 ms with this change. IsolatedLRU improved from 379.09 ms to 43.60 ms because the full-bucket list copy is gone. A deliberately adversarial oldest-small/newest-large layout did expose a longer lock hold: 76.6 ms at a 2x size ratio and 563 ms at 64x. So the extreme case is real, but a hard 100-key cap would miss the byte target and force repeated controller passes. If production traces show this pattern, a better follow-up would be chunking within the same controller cycle while preserving cumulative byte progress and releasing the lock between chunks.

Thanks — this was useful for defining the boundary of the change.

pengw00:
Ran the patch locally. Two things I hit

  1. A delete that keeps failing can stop eviction. If the LRU head object already meets the byte target, the action holds only that key. Your repro would do this if the 66 MiB object had been oldest, since the target is 49.5 MiB. The policy never learns the delete failed: `FSL2Adapter.delete` catches the exception, logs and returns without calling `_notify_keys_deleted` (S3 and Bigtable the same), and `_execute_eviction_action` doesn't check the result. The key stays at the head and is picked again next pass. With a made up 320 MiB + 100x1 MiB at ratio 0.2 and that delete always failing, 5 passes freed 0 MiB; `dev` freed 64 MiB, since there the bad key is 1 of 20 candidates. Fix could be to notify failed keys via `_notify_keys_accessed`, which lands on `on_keys_touched` and moves them off the head. Better not to skip them with `key_eligible_filter`, since then point 2 applies.

  2. The target counts bytes the walk can't touch. `target_size` comes from `self._total_size`, which includes keys rejected by `key_eligible_filter`, but those never add to `selected_size`. When they hold most of the bytes the target can't be reached and the walk goes to the end of the list. With 1000 pinned 1 GiB plus 1M x 4 KiB at ratio 0.2, `dev` evicts 200,200 keys and this evicts all 1M, lock held 43 ms vs 323 ms. Could the target come from the keys that pass the filter, falling back to `self._total_size` when there is none? No L2 caller passes a filter today, but `FleetEvictionController.compute_eviction_plan` does.

positive666:
Thanks for the repro, @pengw00. I could reproduce the first issue through the
L2 controller. With a failing 320 MiB oldest key followed by 100 x 1 MiB keys,
the previous patch freed 0 MiB after five passes. `dev` freed 64 MiB and the
updated patch freed all 100 MiB of deletable data.

I added a policy hook after the adapter delete call. Keys already removed by the
adapter notification are gone; attempted keys that are still present move to
the MRU end, so a bad head cannot block the next pass. This is not recorded as a
cache access.

For the filtered case, I checked the current call sites: L2 supplies sizes but
does not use a filter, while filtered callers currently use unit weights. I also
tried the eligible-byte pre-pass on your workload. It selected the expected
200K keys, but the median went from 731.4 ms to 825.0 ms (+12.8%), so I left it
out of this PR.

I also kept the byte weights across durable LRU restore. The targeted tests pass
on the remote environment: 73 passed.

pengw00:
`on_eviction_attempted` runs after `adapter.delete`, and nothing catches an exception from it.

  In `_check_and_evict_global` and `_check_and_evict_by_cache_salt` the hook is called after `_execute_eviction_action`, and there is no try anywhere in `eviction_controller.py`.

  Skips don't raise: nixl_store skipping pinned keys and dax / raw_block refusing under force=False both return normally, so the hook runs and those retry as designed. Five adapters can raise, though — nixl_store
  (batched_free), nixl_store_dynamic, dax (entry.core.delete_many), native_connector (submit_batch_delete; the 30 s timeout is handled, a raise from the client is not), and mock. When one does, the hook is skipped
  — the very case it was added for — and the exception leaves _eviction_loop, the thread function created in L2EvictionController.__init__, so eviction stops for every adapter in that process. stderr gets a
  traceback; nothing logs it and nothing restarts the thread.

  Two fixes, I think both are needed:
  - Move the on_eviction_attempted(...) call above _execute_eviction_action(...). Success behaviour is unchanged, since a successful delete drops keys by key via on_keys_removed regardless of position; on a raise,
    the key now gets moved to the back.
  - Wrap the _eviction_loop pass body in try/except Exception and log. Otherwise the move-to-back is recorded and never read, because the thread is gone a second later. That part predates this PR, but a failed
    delete didn't matter before and now the design relies on it.

  `test_failed_oversized_delete_rotates_to_smaller_keys` swaps in a delete that returns normally, so the raising path is untested.

pengw00:
Regarding the `1` as a default weight. Two distinct weights make `_sizes_are_uniform()` false, so a mixed bucket always takes the byte path with fake numbers in it — at ratio 0.2 I see 3.2x over-eviction when the size-less keys sit near the LRU head, and 101 of 101 keys selected on the `restore` path. `capture` also writes those 1s out (`persistence_type` is CHECKPOINT), so a later `restore` sums them as real sizes. Let me know if I'm misreading any of this.

Could we keep the whole bucket in count mode while any size is unknown — leave `_per_salt_total_size` unset for a salt restored without `"sizes"`, and stop `_store_sized_key` from overwriting a real weight with the default? If you'd rather keep a per-key number, the bucket's largest known size is the cheapest way to be wrong: it errs in one direction only, toward under-eviction, which costs one more 1s control cycle instead of a re-prefill. correct me if I am wrong

positive666:
You're right. Calling the hook after `adapter.delete()` still left the raising
path uncovered.

I moved `on_eviction_attempted()` before the adapter call. Successful deletes
still remove the keys through the normal listener callback, while a raised
delete now leaves the failed key at the MRU end. I also catch and log failures
per adapter state in the loop, so one backend cannot terminate the eviction
thread or prevent the other states from running.

The new regression test makes the first delete raise. That pass returns through
the loop, and the next pass reaches the 100 smaller keys behind the failed head.
The updated remote test set passes: 79 passed.

positive666:
it was mixing an unknown value into the byte
target and then persisting it as if it were real.

I changed the policies to track whether a bucket still has unknown sizes. Any
unknown keeps the whole bucket in count mode. An unsized refresh preserves a
known weight, and `capture()` omits `sizes` for an incomplete bucket, so an old
order-only checkpoint also stays in count mode after restore. Byte mode starts
only after the last unknown size is filled in.

I added your 100 unknown + 1 known key case as a regression test. At ratio 0.2
it now selects 20 keys before and after capture/restore, instead of all 101.

pengw00:
Nits:

The `on_eviction_attempted` docstrings still describe the old ordering — the ABC says
"keys were submitted for eviction" and both policies say "Keys submitted to an eviction
destination, in LRU order", but after the move they haven't been submitted yet. Someone
implementing this hook on a new policy could read that as "the delete already happened"
and get the failure path wrong, which is the one path it exists for. "in LRU order" also
isn't true on the isolated path, since `_check_and_evict_by_cache_salt` merges keys from
several salts into one `pending[destination]` list first.

`logger.exception` in `_run_eviction_pass` runs once per second per broken adapter,
because `_eviction_loop` sleeps 1 s. An adapter that stays broken — expired credentials,
say — is the expected case for that `except`, and it'll print a full traceback every
second until someone notices. `L1EvictionController._maybe_log_memory_usage` gates on
`extra_logging_enabled` / `extra_logging_interval` and could be followed here.

pengw00:
Two things suggested as follow-ups after this PR:

First: `target_size` is computed over `self._total_size` (all tracked keys) while only filter-passing keys add to `selected_size`, so when ineligible keys hold most of the weight the target is unreachable and the walk runs to the end of the list. The count branch has the same shape (`target_count` over `len(self._order)`), so it's pre-existing
— what changes is the bound: the ineligible share of the key count vs. of the bytes, which a few large pinned objects push near 100%. On 1000 pinned 1 GiB + 1M × 4 KiB at ratio 0.2, count evicts 200,200 (20% of the eligible tail, 49 ms under the lock) and byte evicts all 1,000,000 (372 ms).

Second: the byte branch doesn't run on either MP path today, since FleetEvictionController.on_store notifies without sizes and RawBlockL2Adapter reports a fixed slot_bytes — so this lands inert, which seems like the right order. Actually enabling it is small (that controller already reads self._usage_manager.get_key_bytes(Tier.L2, key)) but it's also the one carrying the pin-count filter, so it wants to come after the target fix above, not before. Independently, _execute_eviction_action publishes only key_count / key_count_per_salt while the trigger is bytes, so a pass that hit its target, one that walked the whole list, and one that freed nothing all look identical from outside.

Issue #4566, opened 2026-08-14
[Bug][MP Observability] extra_stats throughput tests are flaky: synthetic timestamps are always past the pending-prune deadline


aryangorde8:
**Label**

`bug`, `test`, `flaky-test`, `mp_observability`

**Describe the bug**

`tests/v1/mp_observability/subscribers/logging/test_extra_stats.py::TestExtraStatsLoggingSubscriber::test_store_window_logs_tokens_size_and_throughput` is timing-dependent. It passes when the process is fast and fails when it is not, reporting `avg_copy=n/a` where it expects `avg_copy=10.00GB/s`:

```
AssertionError: assert 'store ops=1 tokens=24576 size=5.00GB avg_copy=10.00GB/s' in
  'L0<->L1 stats [cpu:0] last 0.1s: store ops=1 tokens=24576 size=5.00GB avg_copy=n/a
   | retrieve ops=0 tokens=0 size=0.00GB avg_copy=n/a'
```

This is a test-only problem. Production behaviour is correct, and the cause is that the test's synthetic timestamps interact badly with the pending-prune deadline.

**Root cause**

The test builds events with fabricated wall-clock timestamps — `100.0` for the START and `100.5` for the END — chosen so `5e9 bytes / 0.5 s` renders as `10.00GB/s`.

`_prune_pending` drops any pending START older than 60 seconds, measured against the real wall clock:

```python
def _prune_pending(self) -> None:
    deadline = time.time() - _PENDING_MAX_AGE_SECONDS
```

`deadline` is therefore around `1.79e9`, and the synthetic timestamp `100.0` is always below it. Any pending START in this test is stale the instant a prune runs.

A prune runs on every flush, and `_on_store_start` calls `_maybe_flush()` *after* recording the pending START:

```python
def _on_store_start(self, event: Event) -> None:
    key = self._correlation_key(event)
    if key is not None:
        self._pending_store[key] = event.timestamp
    self._maybe_flush()
```

[...]

PR #4568, opened 2026-08-14
[Bug][Test] anchor extra_stats throughput tests to the wall clock

aryangorde8:
Refs #4566.

`test_store_window_logs_tokens_size_and_throughput` and `test_retrieve_window_logs_tokens_size_and_throughput` built their events from fabricated absolute timestamps — `100.0`/`100.5` and `200.0`/`200.25`, chosen so the arithmetic renders as `10.00GB/s` and `8.00GB/s`.

`_prune_pending` drops a pending START older than `_PENDING_MAX_AGE_SECONDS` measured against the wall clock:

```python
def _prune_pending(self) -> None:
    deadline = time.time() - _PENDING_MAX_AGE_SECONDS
```

`deadline` is around `1.79e9`, so a constant like `100.0` is always below it. A prune runs on every flush, and `_on_store_start` calls `_maybe_flush()` **after** recording the pending START. `_INTERVAL` in these tests is `0.05`, so whenever more than 50 ms passes between constructing the subscriber and delivering the first event, the flush fires, the START is pruned, and the later END finds nothing to correlate with. `dt` falls back to `0.0`, `timed` is `False`, and `_format_side` renders `n/a`.

The result is machine-speed dependent: it passes on a fast box and fails on a loaded one. The store case fails reliably on my machine on current `dev`:

```
AssertionError: assert 'store ops=1 tokens=24576 size=5.00GB avg_copy=10.00GB/s' in
  'L0<->L1 stats [cpu:0] last 0.1s: store ops=1 tokens=24576 size=5.00GB avg_copy=n/a | ...'
```

The retrieve test carries the identical pattern and passes today only because its timing happens not to trip the prune.

This anchors both to `time.time()`. The deltas — and therefore the asserted rates — are unchanged; the pending entry is simply recent enough to survive a prune.

**Production is not affected, and this deliberately does not touch it.** `EventBus.publish` stamps `Event.timestamp` from `time.time()`, so a real pending START is milliseconds old and never within 60 seconds of the deadline. The prune is doing exactly its job; only fabricated timestamps trip it.

There is a second possible fix — having `_on_store_start` / `_on_retrieve_start` call `_maybe_flush()` *before* recording the pending entry, so a flush can never prune a START recorded in the same call. That is arguably tidier on its own terms, but it changes production code to fix a test, so I did not take it. Happy to switch if you would rather fix it at that end.

Scope note: four other tests in this file use the same fabricated constants (`test_groups_by_device`, `test_no_flush_before_interval`, `test_failed_op_counts_zero_tokens`, `test_end_to_end_via_event_bus`). None of them asserts a throughput value, so their outcomes do not depend on the correlation succeeding, and I left them alone to keep the diff reviewable.

Verification. I checked the fix rather than assuming it, by forcing the condition the flake depends on — a 60 ms stall between constructing the subscriber and the first event:

```
OLD (fabricated 100.0 / 100.5)
  -> store ops=1 tokens=24576 size=5.00GB avg_copy=n/a          <- flake reproduced
NEW (time.time() anchored)
  -> store ops=1 tokens=24576 size=5.00GB avg_copy=10.00GB/s    <- fixed

With no stall, both report 10.00GB/s — which is why this hides on a fast machine.
```

`tests/v1/mp_observability/` is 389 passed, 1 skipped, 0 failed on this branch; on `dev` the store test fails. `pre-commit` passes on the changed file (ruff, ruff-format, isort, mypy, codespell, SPDX).

**If applicable**:

### DIFF
--- a/tests/v1/mp_observability/subscribers/logging/test_extra_stats.py
+++ b/tests/v1/mp_observability/subscribers/logging/test_extra_stats.py
@@ -112,15 +112,22 @@ class TestExtraStatsLoggingSubscriber:
 
     def test_store_window_logs_tokens_size_and_throughput(self):
         subs = ExtraStatsLoggingSubscriber(_INTERVAL).get_subscriptions()
+        # Anchor to the wall clock rather than a fabricated absolute value.
+        # ``_prune_pending`` drops pending starts older than
+        # ``_PENDING_MAX_AGE_SECONDS`` measured against ``time.time()``, so a
+        # constant like 100.0 is always stale and is discarded by the first
+        # flush that fires between the start and the end -- leaving the end
+        # uncorrelated and the throughput reported as ``n/a``.
+        t0 = time.time()
         with _capture_logs() as handler:
             subs[EventType.MP_STORE_START](
-                _start(EventType.MP_STORE_START, "req-1", 100.0)
+                _start(EventType.MP_STORE_START, "req-1", t0)
             )
             subs[EventType.MP_STORE_END](
                 _end(
                     EventType.MP_STORE_END,
                     "req-1",
-                    100.5,
+                    t0 + 0.5,
                     total_bytes=5_000_000_000,
                     num_tokens=24576,
                 )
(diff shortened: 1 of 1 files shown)


rishabhsinha17 (approved):
@aryangorde8 verified the mechanism against `extra_stats.py`, `_on_store_start` records the pending START and then calls `_maybe_flush()`, and `_prune_pending` compares against `time.time() - 60`, so a fabricated `100.0` epoch START really is dropped by any flush that fires in between, leaving the END uncorrelated. The relative offsets keep the 0.5s and 0.25s durations, so the asserted `10.00GB/s` and `8.00GB/s` strings are unchanged, and neither file has moved on dev since your base. LGTM.

aryangorde8:
@ApostaC @sammshen @hickeyma @deng451e friendly ping on this one.

It is a test-only fix for the extra_stats throughput flake in #4566. Production code is unchanged, CI is green, and @rishabhsinha17 already approved after checking the prune/flush path.

Could one of you review so this can merge? There is also a later PR covering the same issue (#4707); happy to close this as a duplicate if you would rather take that one.

Thanks.

PR #4503, opened 2026-08-12
cleanup: move type-only imports under TYPE_CHECKING in v1/platform

neurawn:
**What this PR does / why we need it**:

Moves type-only imports in `lmcache/v1/platform/` under `typing.TYPE_CHECKING`
so they are not executed at import time. The affected names are used only in
annotations, and every touched module already has
`from __future__ import annotations`, so the annotations stay deferred and
runtime behavior is unchanged.

Files touched:
- `lmcache/v1/platform/cache_context.py` -- `Sequence`, `LayoutHints`, `KVCache`,
  `BaseCacheContext`
- `lmcache/v1/platform/base/cache_context.py` -- `KVLayerGroupsManager`
- `lmcache/v1/platform/cpu/cache_context.py` -- `Sequence`, `KVCache`

Refs #3730

**Special notes for your reviewers**:

Candidates were identified by running (`ruff check --select TC001,TC002,TC003`).

`lmcache/v1/platform/` has further candidates that are deliberately left out of this PR to keep it small.

Verified with `python -c "import lmcache.v1.platform.cpu.cache_context"` (imports
cleanly) and `pre-commit run --all-files` (isort/ruff/ruff-format/mypy all pass;
the only failure is the Rust clippy hook, which cannot build `io-uring` on macOS
and is unrelated to this change).

Other validation run locally:
ruff check --select TC001,TC002,TC003,TC005 lmcache/v1/platform/cpu/cache_context.py lmcache/v1/platform/base/cache_context.py lmcache/v1/platform/cache_context.py

ruff check lmcache/v1/platform/cpu/cache_context.py lmcache/v1/platform/base/cache_context.py lmcache/v1/platform/cache_context.py

pytest -q tests/v1/platform/test_gpu_cache_context.py                                

All passed

**If applicable**:

- [ ] this PR contains user facing changes - docs added
- [ ] this PR contains unit tests

PR #4569, opened 2026-08-14
[good-first-issue] platform: move type-only imports to TYPE_CHECKING

neurawn:

Several imports in the `lmcache/v1/platform/` cache-context modules are used
only in type annotations, but were still executed at import time. This moves
them under a `typing.TYPE_CHECKING` guard so they no longer run at runtime,
trimming import-time dependencies for the platform layer.

All three touched modules already carry `from __future__ import annotations`,
so annotations are never evaluated at runtime and the moved names resolve for
type checkers only.

Refs #3730

- No signature, return type, or behavior changes; this is import placement only.
- #3730 is already closed as completed, but these `platform/` modules were not
  covered by the earlier passes — hence `Refs` rather than `Fixes`.

- Candidates were identified by running (`ruff check --select TC001,TC002,TC003`). `lmcache/v1/platform/` has further candidates that are deliberately left out of this PR to keep it small.

- Verified with `python -c "import lmcache.v1.platform.cpu.cache_context"` (imports
cleanly) and `pre-commit run --all-files` (isort/ruff/ruff-format/mypy all pass;
the only failure is the Rust clippy hook, which cannot build `io-uring` on macOS
and is unrelated to this change).

- Other validation run locally:

`ruff check --select TC001,TC002,TC003,TC005 lmcache/v1/platform/cpu/cache_context.py lmcache/v1/platform/base/cache_context.py lmcache/v1/platform/cache_context.py`

`ruff check lmcache/v1/platform/cpu/cache_context.py lmcache/v1/platform/base/cache_context.py lmcache/v1/platform/cache_context.py`

All passed.

**If applicable**:


neurawn:
Hi @hlin99 and @maobaolong, this pull request is a re-opened PR for my previous one: https://github.com/LMCache/LMCache/pull/4503. In #4503 , I accidentally messed up my workspace while trying to resolve conflicts, so I closed #4503 and open a new one (this current PR) instead. No significant changes compared to the last one.

hlin99 (approved):

maobaolong (approved):
LGTM

PR #4571, opened 2026-08-14
[Bugfix] storage: restore fs_native cache state after restart

### LINKED ISSUE #4564: [Bug] fs_native does not restore existing SSD files on restart, causing trigger_watermark accounting to reset to zero
**Describe the bug**
  When using `fs_native` as the L2 SSD cache, LMCache does not restore the usage/index state from existing files on startup after a process crash or restart. As a result, the
  eviction waterline is effectively reset and starts from zero again, even though the SSD directory already contains previously cached files.

  In production, this means LMCache may underestimate the real SSD usage after restart and fail to evict in time.

  Note: the waterline here is controlled by `trigger_watermark`, not `max_capacity_gb`.
  - `trigger_watermark` comes from `EvictionConfig.trigger_watermark` and is compared against `usage_fraction` in the eviction controller.
  - `max_capacity_gb` is only the capacity limit used as the denominator when computing `usage_fraction`.
  - Relevant code paths:
    - `lmcache/v1/distributed/config.py:214`
    - `lmcache/v1/distributed/storage_controllers/eviction_controller.py:324`

  **To Reproduce**

  1. Start LMCache with `fs_native` as the L2 adapter and a non-zero `max_capacity_gb`.

  ```bash
  lmcache server \
    --host 0.0.0.0 \
    --port 5555 \
    --l1-size-gb 32 \
    --eviction-policy LRU \
    --eviction-trigger-watermark 0.8 \
    --eviction-ratio 0.2 \
    --l2-adapter '{
      "type": "fs_native",
      "base_path": "/data/lmcache/l2",
      "num_workers": 32,
      "use_odirect": false,
      "max_capacity_gb": 500
    }'
  ```
  2. Write KV cache files into the SSD path, e.g. /data/lmcache/l2.
  3. Crash or stop LMCache unexpectedly.
  4. Restart LMCache with the same base_path.
  5. Observe that LMCache does not scan the existing files under /data/lmcache/l2 to rebuild usage/index state.
  6. See that the eviction waterline starts from zero again instead of reflecting the existing SSD contents.

  Expected behavior
  In production, fs_native should scan base_path on startup, recover the usage/index of files that already exist on disk, and preserve the current waterline state instead of
  resetting it to zero on every restart.

  Alternatively, if the intention is to start from a clean state, there should be an explicit cleanup mode that removes the contents of /data/lmcache/l2 before starting, rather
  than relying on restart to implicitly reset the state.

  Screenshots
  Not applicable.

  Desktop (please complete the following information):

  - OS: Linux
  - Browser: Not applicable
  - Version: Not applicable

  Smartphone (please complete the following information):

  - Device: Not applicable
  - OS: Not applicable
  - Browser: Not applicable
  - Version: Not applicable

  Additional context
  Relevant code paths:

  - fs_native creates LMCacheFSClient but does not restore existing SSD state on startup.
  - FSConnector only creates directories, initializes worker threads, and performs per-key file operations; it does not scan the directory to rebuild usage/index state.
  - The eviction controller uses trigger_watermark to compare against usage_fraction, so restarting with empty in-memory accounting can cause the controller to underestimate the
    real SSD usage.

unconsolable:

Fixes #4564.

`fs_native` previously started with empty in-memory usage and eviction
state even when its SSD directory contained cache files from an earlier
process. This caused `usage_fraction` to reset to zero and prevented
watermark-driven eviction after restart.

This PR:

- scans existing `.data` files when the fs_native adapter starts;
- restores aggregate and per-cache-salt byte accounting;
- seeds LRU/IsolatedLRU listeners with recovered keys;
- uses file mtime as the best-effort restart LRU order;
- prevents recovered keys from being double-counted on a repeated store;
- preserves correct accounting when recovered keys are deleted;
- reports recovered key/byte/skipped-file counts in adapter status;
- shares the filesystem filename codec between fs and fs_native.

The exact pre-restart access order is not persisted. File mtime is used
as a deterministic approximation until new accesses update the live LRU
order.

Unrecognized `.data` files and non-regular entries are skipped and
reported. Failure to inspect a recognized cache file fails adapter
startup instead of silently undercounting SSD usage.

**If applicable**:


DongDongJu (changes requested):
Hello @unconsolable, Thanks for heads up.
I am surprised that we still had this issue.
Could you just scan the file in the directory for recover it which not making the .data types meta data?

unconsolable:
> Hello @unconsolable, Thanks for heads up. I am surprised that we still had this issue. Could you just scan the file in the directory for recover it which not making the .data types meta data?

Yes, this implementation already follows that approach. On startup, _scan_existing_cache_entries() scans the existing *.data files directly, reconstructs the ObjectKey from each filename, and uses stat().st_size / st_mtime_ns to restore usage and approximate LRU order. It does not create or persist any additional metadata files, nor does it read the cache-file contents.

PR #4573, opened 2026-08-14
[MP][XPU] Use device-aware event for engine-driven multiprocess transfer

libinta:
The multiprocess connector (LMCacheMPConnector) unconditionally created `torch_dev.Event(interprocess=True)` in start_load_kv and save_kv_layer. `interprocess=True` is a CUDA-only capability: the XPU SYCL-backed event class does not accept the parameter and its ipc_handle() raises NotImplementedError, so the connector crashed on the first KV op on XPU.

The interprocess (exportable) event is only required by the `lmcache_driven` transfer path, which shares the event handle across processes. On XPU, create_transfer_context() already routes to the `engine_driven` path (only device_type == "cuda" selects lmcache_driven). The engine-driven path performs the copy inside the worker and orders it with torch_dev.synchronize(); the event object is not exported and is effectively ignored.

Add a device-aware factory, create_worker_ordering_event(), that returns an interprocess event on backends exposing the `interprocess` parameter (CUDA) and a plain device event otherwise (XPU). This unblocks the engine-driven multiprocess path on XPU without changing CUDA behavior.

Verified end-to-end on Intel XPU (Llama-3.1-8B, TP1): vLLM serve with LMCacheMPConnector reaches readiness with no interprocess-event error, auto-selects engine_driven, and a shared-prefix request produces an external KV cache hit (server retrieves stored tokens).

**If applicable**:

### DIFF
--- a/lmcache/integration/vllm/lmcache_mp_connector.py
+++ b/lmcache/integration/vllm/lmcache_mp_connector.py
@@ -35,6 +35,7 @@ import zmq
 
 # First Party
 from lmcache import torch_dev
+from lmcache import torch_device_type
 from lmcache.banner import print_banner_once
 from lmcache.integration.vllm.experimental import dispatch
 from lmcache.integration.vllm.kv_cache_group_edits import (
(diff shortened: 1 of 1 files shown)


lmcache/integration/vllm/lmcache_mp_connector.py
```diff
@@ -503,6 +504,36 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]):
...
+        """
+        # Lazy import: platform subclass discovery can import the connector
+        # before platform.__init__ finishes, so keep this out of module scope.
+        # First Party
+        from lmcache.v1.platform import get_device_spec
+
+        spec = get_device_spec(torch_device_type)
+        if spec is not None and spec.event_ipc_backend is not None:
+            return torch_dev.Event(interprocess=True)
+        return torch_dev.Event()
```
rebel-jinhwan, line 535: The event requirement seems to depend on which transfer context is selected, not only on the device.

- interprocess=True is needed only by lmcache-driven mp mode (it exports the handle).
- A plain `torch_dev.Event()` is used only by `AsyncEngineDrivenTransferContext.submit_store (_event.wait(stream=self._copy_stream))`.
- The synchronous `EngineDrivenTransferContext` ignores `_event` entirely and orders with `torch_dev.synchronize()`. so we don't need make a event

How about returning an interprocess event / a plain event / None depending on transfer context not only on the device?

maobaolong:
@libinta Thanks for fixing and validating the XPU issue. 

However, this PR selects the event based on device IPC capability rather than the active transfer context.

The required behavior depends on the transfer mode: LMCache-driven needs an IPC event, async engine-driven needs a local event, and synchronous engine-driven needs no event. #4854 implements this distinction directly in each TransferContext and also covers the compatibility connectors.

Therefore, I would prefer #4854 as the more complete and extensible implementation, while crediting this PR for identifying and validating the XPU issue.

maobaolong:
@libinta As #4854 has been merged so i close this PR for now, and i add you to the co-author list, a big thank for this PR.

PR #4579, opened 2026-08-15
[MP][Platform] wire timeline-semaphore event backend behind isolated_ipc config

royyhuang:

Part 2 of the series removing the `hostIPC: true` / shared-`/dev/shm` requirement from containerized MP deployments (part 1: #4551, the standalone timeline-semaphore event IPC backend).

This PR wires that backend into the MP code path behind the boolean knob, **`isolated_ipc`**: it declares that the vLLM workers and the LMCache server may run in containers that share nothing --- when no host IPC namespace and no common `/dev/shm` can be enabled. 
When enabled, `CudaDeviceSpec` resolves the timeline-semaphore backend instead of CUDA interprocess event handles, removing the shared-`/dev/shm` dependency from the event leg of MP STORE/RETRIEVE. The setting is a process-global switch (`lmcache/v1/platform/isolated_ipc.py`) applied at process init, so all three resolution sites (worker export, worker futures import, server) follow it consistently.

Configuration (must be set on both sides; the two backends' handles are mutually unparsable, so a mismatch fails loudly at event import):

- vLLM connector: `lmcache.mp.isolated_ipc` in `kv_connector_extra_config`
- LMCache server: `--isolated-ipc` / `--no-isolated-ipc`

**Defaults to `false`**: the SGLang, TensorRT-LLM, CacheBlend, and qstore integrations still create raw CUDA interprocess events either via torch cuda event or raw CUDA API call. The vLLM MP connector is migrated by this PR (producer events now created via `LMCacheMPWorkerAdapter.create_recorded_event`).

- `extra_config` gains bool-aware value coercion: the existing `type(default)(raw)` would turn the string `"false"` into `True`.
- `cuda-python` is now declared in the CUDA requirement files (the backend needs it when the switch is on; it was an undeclared import).
- Test-harness change worth knowing: exported event handles are only importable while the exporting process is alive (both backends), so the futures tests keep their helper producer process alive until the importer finishes, and the cache-server e2e harness routes event export through the resolved backend — it exercises whichever backend the switch selects.
- Remaining series work: raw `cudaIpcGetMemHandle` KV registration (removes the memory-leg sharing assumptions), migrating the remaining integrations, then flipping the default and dropping `hostIPC` from the operator.

**If applicable**:


lmcache/integration/vllm/vllm_multi_process_adapter.py
```diff
@@ -1395,6 +1422,27 @@ def _reregister_kv_caches_callback(self) -> bool:
...
+        Raises:
+            RuntimeError: If called before ``register_kv_caches()``.
+        """
+        if not self.kv_caches:
+            raise RuntimeError(
+                "KV caches are not registered. Call register_kv_caches() "
+                "before creating transfer events."
+            )
+        device = next(iter(self.kv_caches.values())).device
+        event_ipc_backend = get_event_ipc_backend(device)
```
ApostaC: Maybe double-check whether we should not to put `get_event_ipc_backend` on the hot path (i.e., we do cache it during init time)

requirements/cuda12_core.txt
```diff
@@ -2,3 +2,6 @@
 # LMCACHE_CUDA_MAJOR=12. The cu13 default lives in cuda13_core.txt.
 
 cupy-cuda12x
+# cuda.bindings, used by the timeline-semaphore event IPC backend
+# (selected under isolated IPC) and the raw CUDA IPC wrapper.
+cuda-python>=12,<13
```
ApostaC, line 7: In which filed we used cuda.bindings?

ApostaC (commented):
Otherwise LGTM.
Do we also combine this with #4640 to keep all the events "alive" when doing IPC?

ApostaC (approved):
LGTM!

sammshen (approved):

PR #4580, opened 2026-08-15
feat(operator): add runtimeClassName override to LMCacheEngine

brokedba:
The operator forces `runtimeClassName: nvidia` on the engine pods whenever `gpuVendor` is `nvidia` (the default). That `nvidia` RuntimeClass object is created by the **NVIDIA GPU Operator** , so on NVIDIA clusters that serve GPUs **without the GPU Operator** (device plugin on containerd's default runtime), the object doesn't exist and every engine pod is **admission-rejected**, leaving the DaemonSet stuck at 0.

This PR adds an optional `spec.runtimeClassName` override: an empty string clears it so pods run on the **default container runtime**, while `gpuVendor` stays `nvidia` (keeps `NVIDIA_VISIBLE_DEVICES=all`, etc.). Unset = today's behavior, unchanged.

**New Behavior:**

| `runtimeClassName` set to | Result |
|---|---|
| *(unset / nil)* | vendor default — `nvidia` ⇒ `"nvidia"` class, `amd` ⇒ none (unchanged) |
| `""` (empty) | no runtimeClassName → default runtime (the Nebius fix) |
| `"foo"` | uses `foo` verbatim (works with any vendor) |

## Failure Example
On a Nebius 8xB200 cluster (containerd `default_runtime_name` is already `nvidia`, GPUs via the device plugin, **no GPU Operator**), the engine DaemonSet can't create any pods:

```nginx
Warning  FailedCreate  daemonset/tensormesh-engine
Error creating: pods "tensormesh-engine-" is forbidden: pod rejected: RuntimeClass "nvidia" not found

# daemonset.apps/tensormesh-engine   DESIRED 2   CURRENT 0   READY 0
# lmcacheengine ... phase: Pending   (ConfigValid=True, but 0/2 instances ready)
```

Add an optional override field and apply it after the vendor-derived default (nil = derive, "" = clear, "x" = use x). No behavior change unless the field is set.
```GO
operator/api/v1alpha1/lmcacheengine_types.go  (new field on LMCacheEngineSpec)

+	RuntimeClassName *string `json:"runtimeClassName,omitempty"`

operator/internal/resources/daemonset.go  (buildDaemonSetCore)

#### BEFORE
	gpuVendor := derefString(spec.GPUVendor, lmcachev1alpha1.GPUVendorNvidia)
	var runtimeClassName *string
	if gpuVendor == lmcachev1alpha1.GPUVendorNvidia {
		rc := nvidiaRuntimeClass          // always "nvidia" — no way to opt out
		runtimeClassName = &rc
	}

#### AFTER  (+ explicit override)
	if gpuVendor == lmcachev1alpha1.GPUVendorNvidia {
		rc := nvidiaRuntimeClass
		runtimeClassName = &rc
	}
+	// explicit override wins; "" clears it -> default container runtime.
+	if spec.RuntimeClassName != nil {
+		if *spec.RuntimeClassName == "" {
+			runtimeClassName = nil
+		} else {
+			rc := *spec.RuntimeClassName
+			runtimeClassName = &rc
+		}
+	}
````
CRD + deepcopy regenerated via make manifests / make generate.

**If applicable**:
- [ x] this PR contains unit tests

### DIFF
--- a/operator/api/v1alpha1/lmcacheengine_types.go
+++ b/operator/api/v1alpha1/lmcacheengine_types.go
@@ -423,6 +423,16 @@ type LMCacheEngineSpec struct {
 	// +kubebuilder:validation:Enum=nvidia;amd
 	GPUVendor *string `json:"gpuVendor,omitempty"`
 
+	// runtimeClassName overrides the RuntimeClass for the engine pods. When unset,
+	// it is derived from gpuVendor: "nvidia" uses the NVIDIA GPU Operator's
+	// "nvidia" RuntimeClass, "amd" uses none (the default container runtime). Set
+	// it explicitly to run on a cluster whose GPU runtime is already the default
+	// (e.g. NVIDIA nodes without the GPU Operator, where no "nvidia" RuntimeClass
+	// object exists); an empty string omits runtimeClassName so pods use the
+	// default container runtime.
+	// +optional
+	RuntimeClassName *string `json:"runtimeClassName,omitempty"`
+
 	// image defines the container image to use.
 	// +optional
 	Image *ImageSpec `json:"image,omitempty"`
--- a/operator/api/v1alpha1/zz_generated.deepcopy.go
+++ b/operator/api/v1alpha1/zz_generated.deepcopy.go
@@ -841,6 +841,11 @@ func (in *LMCacheEngineSpec) DeepCopyInto(out *LMCacheEngineSpec) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.RuntimeClassName != nil {
+		in, out := &in.RuntimeClassName, &out.RuntimeClassName
+		*out = new(string)
+		**out = **in
+	}
 	if in.Image != nil {
 		in, out := &in.Image, &out.Image
 		*out = new(ImageSpec)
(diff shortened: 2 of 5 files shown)


ApostaC (commented):
QQ: will this conflict with current `gpuVendor` field in the CRD?

Also, let's update the user-facing docs as well in the `docs/source` folder since this introduces new field in the CRD.

brokedba:
@ApostaC 
> QQ: will this conflict with current `gpuVendor` field in the CRD?
 
There is no conflict with gpuvendor if you look at the code it only works when gpuvendor is set to nvida as a precondition. 
see New Behavior:
> optional override field and apply it after the vendor-derived default (nil = derive, "" = clear, "x" = use x). No behavior change unless the field is set.

okhs0712:
Hi @brokedba,

I encountered the same `RuntimeClass "nvidia" not found` issue on a GPU Operator CDI/NRI cluster and opened #5057 to extend this work with NRI-specific documentation and samples, `CacheBlendEngine` support, and cluster validation.

I rewrote #5057 so that your original `runtimeClassName` implementation is preserved as the first commit with its original authorship, `Signed-off-by` trailer, and a reference to the original commit. My follow-up commit contains the additional NRI/CDI and CacheBlend work.

If you or the maintainers would prefer these additions to be contributed directly through #4580 instead, I am happy to reorganize or close #5057.

brokedba:
@okhs0712  , https://github.com/LMCache/LMCache/pull/5057 was merged , thanks for your contribution.

Issue #4522, opened 2026-08-12
[RFC][MP Coordinator] Refactor the mp coordinator into ingest, key directory, controllers and persistence


ruizhang0101:
## Summary

The coordinator has grown one capability at a time. Each capability added its own wiring, so today the key directory, the usage view, the eviction LRU, the pin table, the dispatcher and the resync pass are tangled in ways that make the next capability harder than the last.

This RFC proposes four layers:

- **Ingest** — HTTP or a message queue delivers events; one **event gate** fences, dedups, handles gaps and owns the stream position; one **broadcaster** fans applied batches out.
- **Key directory** — the placement index and the token bindings. It is the one per-key store, and it has to exist anyway because the read APIs are served from it.
- **Controllers** — each builds exactly the state it needs from the event stream, plans against the directory, and dispatches actions to MP servers. The eviction controller is the first one; others follow the same shape.
- **Persistence** — a **checkpointer** that periodically snapshots the directory plus opted-in controller state, and a **metadata persister** that writes operator intent when it changes.

Startup resync goes away. Recovery is checkpoint plus replay from the durable transport.

This is not one PR. The goal is to agree on the shape, then land it in small steps, each useful on its own.

## Motivation

How the pieces connect today:

```text
  POST /directory/events ─┐
  startup resync ─────────┼─►  KeyDirectory  ─►  broadcaster ─►  L2UsageManager
  checkpoint restore ─────┘    fencing                            L2EvictionManager
                               dedup / gap                          LRU order
                               placements                           pin table
                               token bindings                       eviction dispatch
                                                                    (reads the usage view
                                                                     mid-batch)

  outbound:  eviction → DELETE /cache/objects
             prefetch → POST /cache/prefetches      three separate HTTP callers
             resync   → GET  /cache/objects
```

Three ways in, three ways out, and one class holding five unrelated jobs.

Concrete problems in the code today, not hypotheticals:

**1. Consumers read each other.** `L2EvictionManager.consume` asks `L2UsageManager` whether a key still has bytes, in the middle of the same batch. It works only because `create_app` registers the usage view first, which is documented in a comment. Any new consumer has to know this rule, and the rule is invisible at the call site.

[...]

PR #4583, opened 2026-08-16
[Coordinator] Add cache-event source abstraction

tusharg1993:

Introduces a transport-neutral cache-event source seam for the MP coordinator so the current HTTP path and future replayable transports share the same admission path.

- Adds a common cache-event ingestor above `EventGate` and a lifecycle/status source contract.
- Models `POST /events` as an explicitly non-durable HTTP source without changing its wire behavior.
- Clarifies that HTTP sequence gaps are detectable but not replayable, and that producer-side loss needs separate retry/spooling.
- Refs #4226 and #4522.

- This intentionally does not add Kafka/MQ, snapshot positions, readiness gating, configuration, or endpoint changes.
- Pre-commit hooks passed, including mypy, Ruff, isort, and codespell.
- All 290 MP coordinator tests passed, including real-uvicorn HTTP round-trip and sequence-gap integration tests.

**If applicable**:

### DIFF
--- a/lmcache/v1/mp_coordinator/http_apis/events_api.py
+++ b/lmcache/v1/mp_coordinator/http_apis/events_api.py
@@ -1,9 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Cache-event ingest endpoint on the coordinator (fleet-level).
+"""HTTP cache-event source endpoint on the coordinator (fleet-level).
 
-The ``/events`` surface, thin over the :class:`EventGate`. Top-level
-rather than under ``/directory`` because the stream feeds every consumer
-of it, not one of them. See
+The ``/events`` surface is thin over :class:`HttpCacheEventSource`, which
+passes request-ordered batches through the common ingestor to
+:class:`EventGate`. It is top-level rather than under ``/directory``
+because the stream feeds every consumer of it, not one of them. See
 ``docs/design/v1/mp_coordinator/ingest.md``.
 """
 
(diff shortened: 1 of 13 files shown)


lmcache/v1/mp_coordinator/http_apis/events_api.py
```diff
@@ -30,18 +30,14 @@ async def report_cache_events(
...
-    for batch in body.batches:
-        result = event_gate.ingest(batch)
-        if result == IngestResult.ADMITTED:
-            response.applied += 1
-        elif result == IngestResult.DUPLICATE:
-            response.duplicates += 1
-        else:
-            response.stale += 1
-    return response
+    summary = get_context(request).event_source.ingest(body.batches)
```
ruizhang0101: I feel like the dependency can be vice-versa, the event source can be a component of event gate, the event gate still is the entrance for the ingest just it will have different implementation depending the actual event source.

lmcache/v1/mp_coordinator/ingest/event_source.py
```diff
@@ -0,0 +1,122 @@
...
+        duplicates: Batches dropped because their sequence was already seen.
+        stale: Batches dropped because their incarnation was outdated.
+    """
+
+    applied: int = 0
+    duplicates: int = 0
+    stale: int = 0
+
+
+class CacheEventIngestor:
```
ruizhang0101: Can we get rid of this class and just make ingest batches a method of event gate

ruizhang0101 (commented):
Just a side question, otherwise LGTM.

Also, don't forget to signoff the commit.

lmcache/v1/mp_coordinator/ingest/event_source.py
tusharg1993: Done. Now, there is no CacheEventIngestor and the event gate exposes ingest batches returning a summary of ingested/skipped/etc.

ruizhang0101 (approved):
LGTM

sammshen (approved):
LGTM!

ruizhang0101:
Could you resolve the conflicts?

PR #4586, opened 2026-08-16
feat: Add DSA (Deepseek Sparse Attention) indexer support for SGLang

hekhong-png:
## Summary

This PR adds support for DSA (Deepseek Sparse Attention) indexer KV cache buffers in the SGLang integration path. DSA models such as GLM-5.2 use a dual-buffer KV cache: a standard MLA latent buffer (`kv_buffer`) plus a separate indexer buffer (`index_k_with_scale_buffer`). Both layer groups must be registered with LMCache for store/retrieve to work correctly.

This is the LMCache-side companion to SGLang PR sgl-project/sglang#35230.

### 1. New EngineKVFormat: `NL_X_NB_BSV_BSS = 14`

Added a new `EngineKVFormat` enum value for the DSA indexer format. The DSA indexer tensor is 3-D `(NB, BS, 132)` — logically identical to per-layer MLA with `nh=1`, so it reuses the same paged shape case as `NL_X_NB_BS_HS`.

### 2. SGLang format detector

`kv_format/detectors/sglang.py` now detects the DSA indexer buffer (`index_k_with_scale_buffer`) and assigns `NL_X_NB_BSV_BSS`.

### 3. GPU connector: dual-layer-group registration

`gpu_connectors.py` registers both the MLA latent group and the DSA indexer group. The connector now handles mixed-format layer groups within a single model.

### 4. SGLang adapter: `kv_caches` / `use_mla` / `kv_cache_dim` API

`sglang_adapter.py` and `multi_process_adapter.py` now accept `kv_caches` (flat list of per-layer tensors), `use_mla` (bool), and `kv_cache_dim` (int or None) instead of the old `k_pool` / `v_pool` pair. This matches the MHA / MLA / DSA abstraction in the SGLang connector.

### 5. MP adapter: `_daemon_rid` per-worker uniqueness

All TP workers share the same external `request_id` (from `req.rid`). The daemon's `_prefetch_jobs` dict is keyed by `request_id`, so concurrent LOOKUPs from different workers for the same request overwrote each other, causing prefetch results to be lost.

Fix: `_daemon_rid()` appends `#worker_id` to create a unique daemon-side `request_id` per worker. All daemon-facing wire calls (LOOKUP, WAIT_PREFETCH, RETRIEVE, STORE, FREE_LOOKUP_LOCKS, END_SESSION) now use `_daemon_rid(request_id)`. The client-side `_pending_lookups` dict remains keyed by the external `request_id` (no collision there — each worker has its own connector instance).

### 6. `_per_layer_paged_shape`: handle `NL_X_NB_BSV_BSS`

The DSA indexer format was not listed in `_per_layer_paged_shape`, falling through to the default 5-D shape `(nb, 2, bs, nh, hs)`. The DSA indexer tensor is 3-D `(NB, BS, 132)`. The store path worked by accident (`index_select` with `out=` doesn't check dimensionality), but the retrieve path crashed: `index_copy_(): dimensionality must match. Source 3, dest 5`.

Fix: add `NL_X_NB_BSV_BSS` to the same case as `NL_X_NB_BS_HS`, returning `(nb, bs, hs)`. Both are per-layer MLA-like formats with `nh=1`.

### 7. Storage backends: O_DIRECT alignment for non-4096 page sizes

`gds_backend.py` and `local_disk_backend.py` now round the bounce slot size up to an OS-page multiple. Pools whose page size is not a 4096 multiple (e.g. the DSA indexer pool) would otherwise place slots 1+ at misaligned addresses, and every `aio_write` past the first slot fails with `EINVAL`.

### 8. Tests

Added `tests/v1/gpu_connector/test_dsa_support.py` with unit tests for the DSA indexer format detection, dual-layer-group registration, and store/retrieve round-trip. Updated existing tests for the new `kv_caches` / `use_mla` / `kv_cache_dim` API.

## Testing

- [x] Unit tests pass (`pytest tests/v1/gpu_connector/test_dsa_support.py`)
- [x] Unit tests pass (`pytest tests/v1/test_sglang_adapter.py tests/v1/test_sglang_mp_adapter.py`)
- [ ] Integration: GLM-5.2-FP8 on 8×MI308X, TP8, MP mode
- [ ] Integration: DSA dual-buffer store/retrieve round-trip
- [ ] Integration: L2 cache hit on same-prefix requests

## Related

- SGLang PR: sgl-project/sglang#35230 (LMCache connector DSA support)
- SGLang PR: sgl-project/sglang#35231 (NIXL HiCache DSA fixes)


Shaoting-Feng:
Thanks for your PR. Could you fix the CI tests?

hekhong-png:
@Shaoting-Feng Thanks for the review. I've pushed a fix commit (2bb4608e) that addresses the CI failures:

- **Ruff E501** (line too long): fixed in 6 files
- **Mypy union-attr**: added `assert klg_manager is not None` after `kv_layer_groups_manager` access
- **Mypy assignment**: explicit `Optional` type annotation for the `info` variable in `kv_layer_groups.py`
- **Ruff B905** (`zip()` without `strict=`): added `strict=True`
- **Ruff F401** (unused imports): removed `get_num_layers` and `get_page_buffer_size`
- **DCO**: rebased with `Signed-off-by`

CI should be green now once it re-runs.

hekhong-png:
Recreated as #4996 with a clean rebase against current dev (19 files, no spurious deletions). The old branch was deleted and GitHub does not allow reopening a PR whose branch was recreated, so a new PR was opened.

PR #4587, opened 2026-08-16
fix: correct log and raise calls that drop their message at runtime (#4539)

### LINKED ISSUE #4539: [Bug] Logging calls pass more args than their format string consumes, so the message is dropped at runtime

Anai-Guo:

Fixes #4539.

Three `logger.*` calls pass more positional arguments than their format string consumes. `logging` renders a record lazily as `msg % args`, so each raises `TypeError: not all arguments converted during string formatting` **inside** the logging machinery. Python's logging swallows that and prints a `--- Logging error ---` traceback to stderr instead — **the intended message is never emitted**. All three sit on failure paths, which is exactly where losing the message hurts most.

**1. `lmcache/v1/storage_backend/p2p_backend.py`** — one `%s` placeholder, two args. The caught exception `e` had no placeholder and was silently dropped:

```python
-logger.error("Failed to close peer %s lookup socket", self.peer_init_url, e)
+logger.error(
+    "Failed to close peer %s lookup socket: %s", self.peer_init_url, e
+)
```

**2 & 3. `lmcache/v1/storage_backend/connector/mock_connector.py`** — a stray comma where implicit string concatenation was intended. This truncated the message and turned its tail into an argument. Both rewritten as `%`-format.

## Preventing regressions

`PLE1205` / `PLE1206` (logging too-many / too-few args) were not in the ruff `select` list. Enabling **only** those two rules reports **exactly the two statically-visible cases and nothing else repo-wide**, so they turn on with no extra cleanup:

```
All checks passed!     # after this PR; 2 errors before it
```

Ruff cannot see through the f-string in the second `mock_connector.py` case, so a small regression test (`tests/v1/distributed/test_mock_connector_logging.py`) renders every emitted record and asserts the message survives — `record.getMessage()` raises `TypeError` before the fix.

`G` (flake8-logging-format) is deliberately left commented out; it reports hundreds of findings from the f-string logging still being migrated, and is out of scope here.

## Verification

- `ruff check --select PLE1205,PLE1206 .` → clean (pinned ruff 0.11.7).
- `ruff format --check` → clean.
- Extracted the real `_batched_get` (old vs new) and drove it against fakes: the old version's `getMessage()` raises `TypeError` (message dropped); the new version renders `Mock object is None on 0 out of 1 objects`.

---

## Update: two `raise` sites with the same defect

While confirming the ruff rules cover the file, two `ValueError` raises turned up with the
identical comma-instead-of-concatenation mistake. `PLE1205`/`PLE1206` only inspect logging
calls, so neither rule catches them:

`lmcache/v1/storage_backend/connector/mock_adapter.py`

```python
raise ValueError(
    f"Invalid capacity '{parsed.netloc}' for",
    " mock connector; must be an integer (GB).",
) from e
```

`lmcache/v1/storage_backend/connector/mock_connector.py` (`AsyncLRU.put`)

```python
raise ValueError(
    f"Allocation size {alloc_size} is",
    " greater than capacity {self.capacity}",
)
```

The comma makes each half a separate exception argument, so the rendered message is a tuple
repr; the second half of the `AsyncLRU.put` message is also missing its `f` prefix, so the
capacity is never interpolated:

```
before  ('Allocation size 5 is', ' greater than capacity {self.capacity}')
after   Allocation size 5 is greater than capacity 3
```

Both are fixed by dropping the comma (and adding the missing `f`), which is a pure message
change. `ruff check` and `ruff format --check` stay clean on both files with the pinned
0.11.7.

### DIFF
--- a/lmcache/v1/storage_backend/connector/mock_adapter.py
+++ b/lmcache/v1/storage_backend/connector/mock_adapter.py
@@ -36,8 +36,8 @@ class MockConnectorAdapter(ConnectorAdapter):
             capacity_gb = int(parsed.netloc)
         except ValueError as e:
             raise ValueError(
-                f"Invalid capacity '{parsed.netloc}' for",
-                " mock connector; must be an integer (GB).",
+                f"Invalid capacity '{parsed.netloc}' for"
+                " mock connector; must be an integer (GB)."
             ) from e
 
         params = parse_qs(parsed.query) if parsed.query else {}
--- a/lmcache/v1/storage_backend/connector/mock_connector.py
+++ b/lmcache/v1/storage_backend/connector/mock_connector.py
@@ -264,8 +264,9 @@ class MockConnector(RemoteConnector):
         for i, mock_obj in enumerate(mock_objs):
             if mock_obj is None:
                 logger.warning(
-                    f"Mock object is None on {i}",
-                    f" out of {len(mock_objs)} objects",
+                    "Mock object is None on %d out of %d objects",
+                    i,
+                    len(mock_objs),
                 )
                 break
             metadata = mock_obj.metadata
@@ -276,8 +277,10 @@ class MockConnector(RemoteConnector):
             )
             if memory_obj is None:
                 logger.warning(
-                    "Failed to allocate memory even with",
-                    f" busy loop on {i} out of {len(mock_objs)} objects",
+                    "Failed to allocate memory even with busy loop on %d out of "
+                    "%d objects",
+                    i,
+                    len(mock_objs),
                 )
                 break
             memory_objs.append(memory_obj)
--- a/lmcache/v1/storage_backend/connector/mock_connector.py
+++ b/lmcache/v1/storage_backend/connector/mock_connector.py
@@ -78,8 +78,8 @@ class AsyncLRU:
             alloc_size = mock_obj.num_bytes
             if alloc_size > self.capacity:
                 raise ValueError(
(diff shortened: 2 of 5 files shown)


Anai-Guo:
Pushed one more commit: two `raise ValueError(...)` sites in the same connector pair carry the identical comma-instead-of-concatenation mistake, and `PLE1205`/`PLE1206` do not see them because those rules only inspect logging calls.

- `mock_adapter.py` — `f"Invalid capacity '{parsed.netloc}' for", " mock connector; must be an integer (GB)."`
- `mock_connector.py` (`AsyncLRU.put`) — `f"Allocation size {alloc_size} is", " greater than capacity {self.capacity}"`, whose second half is also missing its `f` prefix.

Rendered:

```
before  ('Allocation size 5 is', ' greater than capacity {self.capacity}')
after   Allocation size 5 is greater than capacity 3
```

Message-only change; `ruff check` / `ruff format --check` remain clean on both files with the pinned 0.11.7. Happy to split this into its own PR if you would rather keep this one to the logging call sites.

PR #4588, opened 2026-08-16
fix(logging): pass the arguments these log calls actually render (#4539)

### LINKED ISSUE #4539: [Bug] Logging calls pass more args than their format string consumes, so the message is dropped at runtime

Anai-Guo:

Fixes #4539

Three `logger.*` calls in `lmcache/` pass more positional arguments than their format string consumes. `logging` renders a record lazily as `msg % args`, so each of these raises `TypeError: not all arguments converted during string formatting` *inside* the logging machinery. Python's logging catches that and prints `--- Logging error ---` plus a traceback to stderr instead — **the intended message is never emitted**. All three sit on failure paths, which is exactly where losing the message hurts most.

**1. `lmcache/v1/storage_backend/p2p_backend.py:155`** — two arguments, one placeholder, so the caught exception was silently dropped:

```python
except Exception as e:
```

The exception is now rendered by its own placeholder.

**2 & 3. `lmcache/v1/storage_backend/connector/mock_connector.py:266,278`** — a stray comma where implicit string concatenation was intended, so each message was truncated and its tail silently became an argument:

```python
logger.warning(
)
```

Both are rejoined into one message and converted to `%`-format, matching the deferred-formatting convention the repo is migrating to.

Two layers guard against a regression:

- **`PLE1205`/`PLE1206` added to `[tool.ruff.lint] select`.** These were not previously selected. With them enabled, `ruff check .` (v0.11.7, the `.pre-commit-config.yaml` pin) is clean repo-wide — the only two findings were the plain-string cases fixed here, so no additional cleanup rides along. The full `G` ruleset stays out of scope: it still reports ~500 findings from the f-string logging migration that is in progress.

- **`tests/test_logging_format.py`.** Ruff cannot count placeholders inside an f-string, so case 2 above is invisible to `PLE1205`. A small AST walk over `lmcache/` compares each logging call's placeholder count against its positional-argument count, which covers the f-string variant too. Two unit tests pin the guard itself (it flags the stray-comma shape, and leaves a well-formed call alone).

Verified against the tree at `e8f9381`:

| | `test_logging_arguments_match_format_string` |
|---|---|
| before the source fix | **FAILS** — reports all 3 call sites, including the 2 f-string ones ruff misses |
| after the source fix | **PASSES** |

`ruff check`, `ruff format --diff` and `isort --diff` are clean on all four files.

**If applicable**:

### DIFF
--- a/lmcache/v1/storage_backend/connector/mock_connector.py
+++ b/lmcache/v1/storage_backend/connector/mock_connector.py
@@ -264,8 +264,9 @@ class MockConnector(RemoteConnector):
         for i, mock_obj in enumerate(mock_objs):
             if mock_obj is None:
                 logger.warning(
-                    f"Mock object is None on {i}",
-                    f" out of {len(mock_objs)} objects",
+                    "Mock object is None on %d out of %d objects",
+                    i,
+                    len(mock_objs),
                 )
                 break
             metadata = mock_obj.metadata
@@ -276,8 +277,10 @@ class MockConnector(RemoteConnector):
             )
             if memory_obj is None:
                 logger.warning(
-                    "Failed to allocate memory even with",
-                    f" busy loop on {i} out of {len(mock_objs)} objects",
+                    "Failed to allocate memory even with busy loop on %d out of "
+                    "%d objects",
+                    i,
+                    len(mock_objs),
                 )
                 break
             memory_objs.append(memory_obj)
--- a/lmcache/v1/storage_backend/p2p_backend.py
+++ b/lmcache/v1/storage_backend/p2p_backend.py
@@ -152,7 +152,9 @@ class PeerInfo:
         try:
             self.lookup_socket.close(linger=0)
         except Exception as e:
-            logger.error("Failed to close peer %s lookup socket", self.peer_init_url, e)
+            logger.error(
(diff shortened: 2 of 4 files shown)


Anai-Guo:
Closing as a duplicate — I filed #4587 for the same issue five minutes earlier, with the same three source fixes and the same `PLE1205`/`PLE1206` addition. Apologies for the noise; please review #4587 instead.

The only thing that differs here is the regression test. #4587 exercises the `mock_connector` call sites directly; this branch (`Anai-Guo:fix/logging-format-arg-count`) instead adds an AST guard that walks all of `lmcache/` and compares each logging call's placeholder count to its positional-argument count, so it also covers the f-string form that `PLE1205` cannot see anywhere in the tree, not just at these two call sites. Happy to move that over to #4587 if you'd prefer the broader guard.

Issue #4572, opened 2026-08-14
[Bug][CI] unit-tests-amd is not completing builds, and the one that did fails test_classification_matches_golden


aryangorde8:
**Label**

`bug`, `ci`

**Describe the bug**

Three related problems in the `unit-tests-amd` Buildkite pipeline. The third is why the
first two went unnoticed.

### 1. `test_classification_matches_golden` fails on the ROCm agent

[Build #10714](https://buildkite.com/lmcache/unit-tests-amd/builds/10714) (ran
2026-08-14 17:40–17:51 UTC) stopped at 37%:

```
FAILED tests/v1/gpu_connector/test_kv_format_classification.py::test_classification_matches_golden
AssertionError: EngineKVFormat.NB_NL_TWO_BS_NH_HS: got (False, False, False, False), expected (True, False, False, False)
1 failed, 2125 passed, 79 skipped, 1 xpassed in 326.06s
```

On the commit that was built, the two sides of that assertion agree in source:

- `csrc/engine_kv_format.h` gives `NB_NL_TWO_BS_NH_HS` (enum value 0)
  `{.is_cross_layer = true}` in `format_facts`
- `EXPECTED` in `tests/v1/gpu_connector/test_kv_format_classification.py` says
  `(True, False, False, False)`

So the extension module loaded on the agent disagreed with the source it was built from.
Everything the golden table expects has exactly one `True`, and the observed value was
all-`False`, which is what a default-initialised `format_facts` entry would give.

The build did compile from source — `Built lmcache @ file:///var/lib/buildkite-agent/builds/ENC1-CLS01-SVR13-1/lmcache/unit-tests-amd`,
`Prepared 1 package in 4m 28s` — so this is not a missing build step.

I have not reproduced this off the ROCm agent; I do not have one. Two candidates worth
checking, in the order I would check them:

- a workspace left behind by an earlier build (see 3 below), so the editable install
  picked up objects compiled from different source
- `hipcc` and the `constexpr` designated-initialiser table in `format_facts` — the
  predicates are `LMC_KV_FORMAT_HD constexpr`, and an all-`False` table is consistent with
  that initialisation not surviving the HIP compile

### 2. The pipeline has not completed a build

Of the 61 most recent builds in `unit-tests-amd`, **none** finished: 32 are still
`scheduled` and 28 are `skipped`. Every recent `dev` build is `skipped`, so there is no
green baseline for this pipeline at all.

[...]

PR #4441, opened 2026-08-06
Report cache simulator percentiles at the nearest rank instead of one sample above it

bharadwaj-pendyala:
## Problem

`_percentiles` in `lmcache/tools/cache_simulator/simulator.py` read the sorted samples at `int(p / 100 * n)`. That expression is the nearest **rank**, which is 1-based, used as a 0-based index. Every percentile landing on a whole rank came back one sample high.

With the samples 1.0 through 100.0:

| | p25 | p50 | p75 | p90 | p99 |
|---|---|---|---|---|---|
| before | 26.0 | 51.0 | 76.0 | 91.0 | **100.0** |
| after | 25.0 | 50.0 | 75.0 | 90.0 | 99.0 |

p99 of a hundred samples was the single worst sample, so the reported tail was the max.

`print_statistics` runs this helper over all five distributions it reports: per-request token hit rates, hit prefix lengths, chunk reuse counts, global spans, and cache positions at hit. Every one of them was off whenever `p * n` was divisible by 100.

## Solution

Compute the nearest rank as `(p * n + 99) // 100` and index at `rank - 1`.

Integer arithmetic rather than `math.ceil(p / 100 * n)` on purpose: `7 / 100.0` is not exact, so `(7 / 100.0) * 100` is `7.000000000000001` and a float `ceil()` pushes a whole rank to the next one. Keeping it in integers removes that case instead of commenting around it.

Same defect and same shape of fix as `_percentile` in `lmcache/cli/commands/trace/_stats.py` (#4416).

## Scope

One helper. No caller changes, no new arguments, no input validation added: `print_statistics` is the only caller and it always passes `[25, 50, 75, 90, 99]`. The charts in `plot_statistics` build their own histograms and are untouched.

This does change the numbers in the text report for anyone comparing against an older run. That is the point of the fix, but worth calling out.

## Validation

- `pytest tests/tools/test_cache_simulator.py` — 29 passed.
- `TestPercentiles::test_whole_ranks_do_not_advance_to_the_next_sample` and `TestPrintStatistics::test_reported_p99_is_the_ninety_ninth_sample_not_the_max` both fail on `dev` and pass here. Checked by reverting `simulator.py` to `origin/dev` with the new tests in place: 2 failed, 1 passed.
- The report-level test drives the public `print_statistics` and asserts the printed Stat 2 p99 line reads `99`, not `100`.
[...]

PR #4590, opened 2026-08-16
fix(cache-simulator): correct percentile calculation

### LINKED ISSUE #4440: cache simulator reports every whole-rank percentile one sample too high (p99 of 100 samples is the max)
## What happens

`_percentiles` in `lmcache/tools/cache_simulator/simulator.py:324` reads the sorted sample list at index `int(p / 100 * n)`. That expression is the nearest **rank**, which is 1-based, and it is used as a 0-based index. Every percentile that lands on a whole rank reports the next sample up.

```python
def _percentiles(values: list[float], pcts: list[int]) -> dict[str, float]:
    if not values:
        return {}
    s = sorted(values)
    n = len(s)
    result = {}
    for p in pcts:
        idx = min(int(p / 100 * n), n - 1)
        result[f"p{p}"] = s[idx]
    return result
```

With the values 1.0 through 100.0, the five percentiles the simulator reports come out one sample high across the board:

| | p25 | p50 | p75 | p90 | p99 |
|---|---|---|---|---|---|
| current | 26.0 | 51.0 | 76.0 | 91.0 | **100.0** |
| nearest rank | 25.0 | 50.0 | 75.0 | 90.0 | 99.0 |

p99 of a hundred samples is the single worst sample, so the reported p99 is the max. With ten samples, p90 has the same problem: reported 10.0, nearest rank 9.0.

## Reproduce

```python
from lmcache.tools.cache_simulator.simulator import _percentiles
print(_percentiles([float(i) for i in range(1, 101)], [25, 50, 75, 90, 99]))
# {'p25': 26.0, 'p50': 51.0, 'p75': 76.0, 'p90': 91.0, 'p99': 100.0}
```

## Why it matters

`print_statistics` runs `_percentiles` over five distributions: per-request hit rates, request lengths, reuse distances, reuse spans, and first-hit positions (`simulator.py:362`, `:381`, `:392`, `:422`, `:435`). Those numbers are what someone reads when sizing a cache from a trace. A p99 that is really the max makes the tail look worse than it is, and a p50 that is really the 51st sample skews the middle.

## Second defect in the same line

`p / 100 * n` divides before multiplying. `7 / 100.0` is not exact, so `(7 / 100.0) * 100` is `7.000000000000001`. That does not change the current result because `int()` truncates, but it does bite any fix built on `math.ceil`. Computing `p * n / 100.0` keeps whole ranks whole.

## Fix

Same shape as the fix for `_percentile` in `lmcache/cli/commands/trace/_stats.py` (#4416):

```python
rank = math.ceil(p * n / 100.0)
result[f"p{p}"] = s[max(rank, 1) - 1]
```

`math` is already imported in this module. Happy to send the patch with unit tests covering the whole-rank, fractional-rank, and inexact-division cases.

## Environment

`dev` @ 3d7a413, Python 3.12.

Masih-Roughani:
## Summary

Fix the cache simulator percentile calculation to use 1-based nearest-rank semantics correctly before indexing the sorted 0-based sample list.

The calculation now uses `math.ceil(p * n / 100.0)` and subtracts one when indexing. Multiplication is performed before division to avoid floating-point rounding turning an exact whole rank into the next rank.

Adds regression coverage for:
- whole-rank percentiles
- fractional ranks
- the floating-point inexact-division case

Fixes #4440

## Validation

- `python -m pytest tests/tools/test_cache_simulator.py -k percentiles -v --confcutdir=tests/tools` — 3 passed
- `python -m pytest tests/tools/test_cache_simulator.py -q --confcutdir=tests/tools` — 29 passed
- `python -m ruff check lmcache/tools/cache_simulator/simulator.py tests/tools/test_cache_simulator.py` — all checks passed
- `python -m ruff format --check lmcache/tools/cache_simulator/simulator.py tests/tools/test_cache_simulator.py` — both files already formatted
- Full `tests/tools/test_cache_simulator.py` run on Windows: 25 passed, 4 existing `TestLoadLookupEvents` tests fail because they hard-code `/tmp`, which resolves to missing `C:\tmp` on Windows. The percentile tests all pass.
- [AMD Buildkite #11203](https://buildkite.com/lmcache/unit-tests-amd/builds/11203) fails in `tests/v1/gpu_connector/test_kv_format_classification.py::test_classification_matches_golden`. This is the known, branch-independent AMD CI failure tracked in #4572; this PR does not touch GPU connector or native format-classification code.

### DIFF
--- a/lmcache/tools/cache_simulator/simulator.py
+++ b/lmcache/tools/cache_simulator/simulator.py
@@ -328,8 +328,8 @@ def _percentiles(values: list[float], pcts: list[int]) -> dict[str, float]:
     n = len(s)
     result = {}
     for p in pcts:
-        idx = min(int(p / 100 * n), n - 1)
-        result[f"p{p}"] = s[idx]
+        rank = math.ceil(p * n / 100.0)
+        result[f"p{p}"] = s[max(rank, 1) - 1]
     return result
 
 
--- a/tests/tools/test_cache_simulator.py
+++ b/tests/tools/test_cache_simulator.py
@@ -12,10 +12,12 @@ import pytest
 # First Party
 from lmcache.tools.cache_simulator.lru_cache import LRUCache, LRUCacheFast
 from lmcache.tools.cache_simulator.simulator import (
+    _percentiles,
     compute_kv_bytes_per_chunk,
     load_lookup_events,
     simulate,
 )
+
 
 # ---------------------------------------------------------------------------
 # LRUCacheFast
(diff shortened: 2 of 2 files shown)


yoo-kumaneko:
It seems that there has been a PR working on this. Not sure if it's the same issue.

https://github.com/LMCache/LMCache/pull/4441

PR #4593, opened 2026-08-16
fix(mp): harden worker heartbeat liveness

myshytf:
## Summary
- add an opt-in heartbeat response timeout independent of heartbeat cadence
- start the worker heartbeat immediately after successful KV-cache registration
- preserve interval-coupled timeout behavior by default and idempotent store/retrieve start calls
- document the response-timeout/reap-floor relationship and post-registration lifecycle

## Why
Using the heartbeat interval as the MQ response deadline couples two independent controls. A short cadence can then mark a healthy but busy NORMAL pool unhealthy. Separately, vLLM registers the cache before model warmup and CUDA graph capture; waiting for the first store/retrieve leaves that registered context without liveness pings throughout startup and any later idle period.

## Validation
- new eager-start regression was observed failing before the production change and passing afterward
- `tests/v1/test_vllm_mp_adapter.py`: 31 passed after rebasing onto current `upstream/dev`
- focused `ruff check`, `ruff format --check`, and `git diff --check` pass
- existing timeout qualification covered a busy NORMAL pool, idle periods longer than 120 seconds, a real outage, and recovery

## Compatibility
- `lmcache.mp.heartbeat_timeout <= 0` inherits the interval, preserving legacy behavior
- heartbeat starts only after successful cache registration
- idempotent start calls remain on store/retrieve for compatibility
- scheduler-adapter heartbeat behavior is unchanged

### DIFF
--- a/docs/design/v1/multiprocess/worker_liveness.md
+++ b/docs/design/v1/multiprocess/worker_liveness.md
@@ -150,6 +150,16 @@ re-registers only on a genuine recovery edge (Section 6.2). A retrieve dropped
 while the server is unhealthy is still reported via `get_finished` so async loads
 cannot hang.
 
+The worker adapter accepts `lmcache.mp.heartbeat_interval` (default `10.0` seconds)
+for the PING cadence and `lmcache.mp.heartbeat_timeout` (default `0.0`) for the
+response deadline. A non-positive timeout inherits the interval, preserving
+legacy behavior. Deployments whose NORMAL pool can run longer than the desired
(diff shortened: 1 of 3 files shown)


myshytf:
Updated the head after review:

- added the required `HeartbeatThread.__init__ -> None` annotation
- corrected the lifecycle documentation to reflect that vLLM registers KV caches before warmup/CUDA-graph capture
- rebased onto current `dev`

Focused source-matched adapter suite: 31 passed; DCO is green.

PR #4594, opened 2026-08-16
[Bugfix] fix amd ut test failed

chunxiaozheng:
Fix AMD ROCm classification failures for EngineKVFormat.

The KV-format facts were stored in a global __host__ __device__ table initialized with C++20 designated initializers. Under the AMD/HIP C++17 build path, this could produce incorrect classification results—for example, NB_NL_TWO_BS_NH_HS was reported as not cross-layer.

This change replaces the global table lookup with C++17-compatible constexpr format facts, preserving the existing classification behavior for all supported formats. Unsupported formats now fail explicitly instead of being silently classified as all false.

### DIFF
--- a/csrc/engine_kv_format.h
+++ b/csrc/engine_kv_format.h
@@ -2,6 +2,11 @@
 
 #pragma once
 
+#if !defined(__CUDA_ARCH__) && !defined(__HIP_DEVICE_COMPILE__) && \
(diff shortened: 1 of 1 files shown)


Shaoting-Feng (approved):
LGTM

maobaolong (approved):
LGTM

PR #4599, opened 2026-08-16
[Bugfix][CLI] Don't count timed-out rounds in L2 bench stats

jeojdi1:

`bench_store`, `bench_lookup` and `bench_load` all record a round that timed out as `float("inf")` seconds. That causes two problems.

`std_duration` runs `statistics.stdev` over those durations, and `inf` makes it raise. So the benchmark crashes at the end, when it prints the results.

The other one doesn't error at all. Throughput is `(bytes_per_round / _MB) / d`. With `d = inf` that gives exactly `0.0`, and the `d <= 0` check doesn't catch it. So a round that timed out gets counted as a round that ran at 0 MB/s, and the average comes out lower than it should be.

For example: durations `[0.10, 0.11, 0.09, timeout, 0.12]` at 64 MiB per round. It reports 493.25 MB/s. The four rounds that finished averaged 616.57. `min_throughput_mbps` reports `0.0`.

Now timeouts go into a `timed_out_rounds` counter instead. `attempted_rounds` keeps `total_keys` and `actual_hit_rate` at the same values as before, so only the duration and throughput numbers change.

`rounds` now counts every round that started, and I added a `rounds_timed_out` line next to it so the timeouts are visible. Happy to drop that if you'd rather leave the output as it was.

`result.py` didn't have any tests, so I added three.

**If applicable**:

### DIFF
--- a/docs/source/cli/bench.rst
+++ b/docs/source/cli/bench.rst
@@ -1380,6 +1380,7 @@ Final summary (one section per exercised operation):
    --------------------------- Store -----------------------
    Operation:                        Store
    Rounds:                           1
+   Rounds timed out:                 0
    Keys / round:                     96
    Total keys:                       96
    Total success:                    96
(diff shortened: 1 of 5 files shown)


chunxiaozheng (approved):
@jeojdi1 Thanks for your fix, LGTM!

maobaolong (approved):
LGTM

maobaolong (approved):
LGTM

PR #4601, opened 2026-08-17
[Examples] token_dropping: pass --supported-transfer-mode auto to LMCache server

wanke1997:
## Summary

Fixes the `examples/token_dropping/` SDK notebooks so they run again out of the
box. Two independent breakages are addressed:

1. **`--supported-transfer-mode auto`** is now passed to every LMCache server launch.
2. **`model_name`** uses the portable `Qwen/Qwen3-8B` repo id instead of a
   machine-local absolute path.

### Motivation
PR #4447 ("Update the default value of shm name and supported transfer mode to
avoid shm warnings") changed the LMCache server default:

| config | before | after (#4447) |
| --- | --- | --- |
| `supported_transfer_mode` | `auto` | `lmcache_driven` |

With `lmcache_driven`, the server builds only `LMCacheDrivenTransferModule` and
does **not** register the engine-driven handlers. The token_dropping SDK
examples register **CPU (engine-driven)** KV and query caches, so
`ctx.register_caches()` now sends `REGISTER_KV_CACHE_ENGINE_DRIVEN_CONTEXT`,
which the server rejects:

```
LMCache ERROR: No handler registered for request type
RequestType.REGISTER_KV_CACHE_ENGINE_DRIVEN_CONTEXT
```

and the SDK then times out:

```
LMCacheTimeoutError: Future result not available within timeout
```

These examples worked before #4447 because the old default was `auto` (loads
both the lmcache-driven and engine-driven modules). #4447 updated its own
scripts/docs to pass the flag explicitly but missed these notebooks.

### Change
Added `--supported-transfer-mode auto` to every LMCache server launch (both the
documented shell commands and the Colab `subprocess` launch lists). `auto` is
used rather than `engine_driven` so a single server keeps serving both the SDK's
engine-driven CPU path and vLLM's lmcache-driven GPU path.

### Motivation
Four notebooks hardcoded a machine-local path as `model_name`:

```python
model_name = "/data/models/Qwen3-8B"
```

`model_name` is sent as the vLLM `model` id by `make_post_completion`, but the
documented `vllm serve` uses `--served-model-name Qwen/Qwen3-8B`. The mismatch
makes every completion request 404 (`model does not exist`). It also leaks a
local filesystem path into an open-source example.

### Change
Use the portable repo id `model_name = "Qwen/Qwen3-8B"`, which matches the
served model name and loads the tokenizer/config from the HF hub cache.

## Files changed

| Notebook | transfer-mode | model_name |
| --- | :---: | :---: |
| `random_token_dropping.ipynb` | ✅ | — (already portable) |
| `rkv_token_dropping.ipynb` | ✅ | ✅ |
| `snapkv_token_dropping.ipynb` | ✅ | — |
| `rkv_variants/rkv_variant_buffered_cpu.ipynb` | ✅ | ✅ |
| `rkv_variants/rkv_variant_cpu_exact.ipynb` | ✅ | ✅ |
| `rkv_variants/rkv_variant_gpu_exact.ipynb` | ✅ | ✅ |
| `rkv_colab.ipynb` | ✅ (markdown + `subprocess`) | — |
| `snapkv_colab.ipynb` | ✅ (markdown + `subprocess`) | — |

Also fixed a couple of pre-existing missing line-continuation backslashes in the
`snapkv_colab.ipynb` documented command while touching that block.

## Testing

- Started an LMCache server with `--supported-transfer-mode auto` and re-ran
  `rkv_token_dropping.ipynb` end-to-end; `register_caches()` succeeds and the
  timeout is gone.
- With vLLM served as `Qwen/Qwen3-8B`, completion requests using the portable
  `model_name` return `200` (previously `404`).
- All eight notebooks remain valid JSON; every code cell compiles.


KuntaiDu (approved):
LGTM!

deng451e (approved):
LGTM!

